Issue #6204, opened 2016-08-18
Tooltip for exported variables


Kazuo256:
**Operating system or device - Godot version:** doesn't apply

**Issue description** (what happened, and what was expected):

I couldn't find one, so I wanted to open an issue for this entry in the site's Q&A:
https://godotengine.org/qa/313/tooltip-documentation-custom-exported-variables-gdscript

Basically, would it be possible to add a feature for specifying custom tooltip texts in exported variables?
The main motivation is to improve communication between team members inside a game project, particularly between those that produce development tools and those that use them.

**Steps to reproduce:** doesn't apply

**Link to minimal example project** (optional but very welcome): doesn't apply?

---

Thanks in advance!

Zylann:
Could we use docstrings for this?

vnen:
> Could we use docstrings for this?

Well, if you're willing to implement such thing in GDScript... Not that's a bad thing, but it'll require some work.

Zylann:
Ooops, I thought GDScript supported docstrings? I've seen the code editor color patterns like `"""test"""` in the same color as comments, and I found some in Akien's Logger plugin. But if it's not... well yeah, it's a bit of work.

RebelliousX:
It would be great to take a comment (1 line or multiple) that precedes the exported variable as a tooltip. The same way Visual Studio is handling that. In fact, it would be nice to have tooltips for all variables that have comments.

Zylann:
It can be nasty if the comment is actually not intented for the variable (TODO item or commented code before the variable). That's why I recommend docstrings, and also ignoring those that are more than 1 empty space above the documented member.

Sslaxx:
This sounds more like it's leaning more towards a general code documentation issue.

RebelliousX:
@Zylann Not really, the script editor should ignore empty lines above variable declarations. Besides, comments directly above variables (without empty lines) that don't relate to them are bad anyways.
It wouldn't require much work for Godot to implement that.

Or if that really not feasible, the comment can be on the same line after the declaration, but would be harder to do multiline tooltips that way.
`export var h_scroll_spd = 10    # Horizontal scroll speed`

Zylann:
If we can't distinguish between doc comments and simple comments, there will be cases where the editor will show both. Sometimes, I have this in my code:
```gdscript
# TODO Change this to a string after the next release
# Does things if you set it to 43
export var test = 42
```
And I don't want the TODO to be in the tooltip.

Also, consider this:
[...]

PR #85142, opened 2023-11-20
Improve usability of zooming in the animation editor

CookieBadger:
This PR seeks to improve the zoom usability in the AnimationPlayer editor, I would be happy to hear your feedback! Especially the changes I had to make to the header file is something that a more experienced pair of eyes should be laid on.

I added some code in the AnimationTrackEditor, such that zooming with the mouse zooms in on the current mouse position, whereas the zoom panner always zooms in on the center of the currently visible timeline, instead of zooming on the beginning in both cases.
This implementation comes as a result of the proposal [#6204](https://github.com/godotengine/godot-proposals/issues/6204)

I have never contributed before, so please bear with me :)

### DIFF
--- a/editor/animation_track_editor.cpp
+++ b/editor/animation_track_editor.cpp
@@ -1239,6 +1239,58 @@ void AnimationMultiTrackKeyEdit::set_use_fps(bool p_enable) {
 }
 
 void AnimationTimelineEdit::_zoom_changed(double) {
+	double zoom_pivot = 0; // Point on timeline to stay fixed.
+	double zoom_pivot_delta = 0; // Delta seconds from left-most point on timeline to zoom pivot.
+
+	int timeline_width_pixels = get_size().width - get_buttons_width() - get_name_limit();
+	double timeline_width_seconds = timeline_width_pixels / last_zoom_scale; // Length (in seconds) of visible part of timeline before zoom.
+	double updated_timeline_width_seconds = timeline_width_pixels / get_zoom_scale(); // Length after zoom.
+	double updated_timeline_half_width = updated_timeline_width_seconds / 2.0;
+	bool zooming = updated_timeline_width_seconds < timeline_width_seconds;
+
+	double timeline_left = get_value();
+	double timeline_right = timeline_left + timeline_width_seconds;
+	double timeline_center = timeline_left + timeline_width_seconds / 2.0;
+
+	if (zoom_callback_occured) { // Zooming with scroll wheel will focus on the position of the mouse.
+		double zoom_scroll_origin_norm = (zoom_scroll_origin.x - get_name_limit()) / timeline_width_pixels;
+		zoom_scroll_origin_norm = MAX(zoom_scroll_origin_norm, 0);
+		zoom_pivot = timeline_left + timeline_width_seconds * zoom_scroll_origin_norm;
+		zoom_pivot_delta = updated_timeline_width_seconds * zoom_scroll_origin_norm;
+		zoom_callback_occured = false;
+	} else { // Zooming with slider will depend on the current play position.
+		// If the play position is not in range, or exactly in the center, zoom in on the center.
+		if (get_play_position() < timeline_left || get_play_position() > timeline_left + timeline_width_seconds || get_play_position() == timeline_center) {
+			zoom_pivot = timeline_center;
+			zoom_pivot_delta = updated_timeline_half_width;
+		}
+		// Zoom from right if play position is right of center,
+		// and shrink from right if play position is left of center.
+		else if ((get_play_position() > timeline_center) == zooming) {
+			// If play position crosses to other side of center, center it.
+			bool center_passed = (get_play_position() < timeline_right - updated_timeline_half_width) == zooming;
+			zoom_pivot = center_passed ? get_play_position() : timeline_right;
+			double center_offset = CMP_EPSILON * (zooming ? 1 : -1); // Small offset to prevent crossover.
+			zoom_pivot_delta = center_passed ? updated_timeline_half_width + center_offset : updated_timeline_width_seconds;
+		}
+		// Zoom from left if play position is left of center,
+		// and shrink from left if play position is right of center.
+		else if ((get_play_position() <= timeline_center) == zooming) {
+			// If play position crosses to other side of center, center it.
+			bool center_passed = (get_play_position() > timeline_left + updated_timeline_half_width) == zooming;
+			zoom_pivot = center_passed ? get_play_position() : timeline_left;
+			double center_offset = CMP_EPSILON * (zooming ? -1 : 1); // Small offset to prevent crossover.
+			zoom_pivot_delta = center_passed ? updated_timeline_half_width + center_offset : 0;
+		}
+	}
+
+	double hscroll_pos = zoom_pivot - zoom_pivot_delta;
+	hscroll_pos = CLAMP(hscroll_pos, hscroll->get_min(), hscroll->get_max());
+
+	hscroll->set_value(hscroll_pos);
+	hscroll_on_zoom_buffer = hscroll_pos; // In case of page update.
+	last_zoom_scale = get_zoom_scale();
+
 	queue_redraw();
 	play_position->queue_redraw();
 	emit_signal(SNAME("zoom_changed"));
@@ -1419,6 +1471,11 @@ void AnimationTimelineEdit::_notification(int p_what) {
 			}
 
 			set_page(zoomw / scale);
+
+			if (hscroll->is_visible() && hscroll_on_zoom_buffer >= 0) {
+				hscroll->set_value(hscroll_on_zoom_buffer);
+				hscroll_on_zoom_buffer = -1.0;
+			}
 
 			int end_px = (l - get_value()) * scale;
 			int begin_px = -get_value() * scale;
@@ -1725,7 +1782,9 @@ void AnimationTimelineEdit::_pan_callback(Vector2 p_scroll_vec, Ref<InputEvent> 
 
 void AnimationTimelineEdit::_zoom_callback(float p_zoom_factor, Vector2 p_origin, Ref<InputEvent> p_event) {
 	double current_zoom_value = get_zoom()->get_value();
-	get_zoom()->set_value(MAX(0.01, current_zoom_value * p_zoom_factor));
+	zoom_scroll_origin = p_origin;
+	zoom_callback_occured = true;
+	get_zoom()->set_value(MAX(0.01, current_zoom_value - (1.0 - p_zoom_factor)));
 }
 
 void AnimationTimelineEdit::set_use_fps(bool p_use_fps) {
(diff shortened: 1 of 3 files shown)


CookieBadger:
Instead of always zooming in on the left side of the timeline, this is the way zooming behaves now:
AnimationPlayerZoom

Also works in the Bezier editor:
AnimationBezierZoom

editor/animation_track_editor.h
```diff
@@ -184,6 +188,7 @@ class AnimationTimelineEdit : public Range {
 	void set_track_edit(AnimationTrackEdit *p_track_edit);
 	void set_zoom(Range *p_zoom);
 	Range *get_zoom() const { return zoom; }
+	void _zoom_callback(float p_zoom_factor, Vector2 p_origin, Ref<InputEvent> p_event);
```
CookieBadger, line 191: Making the callback public is probably not a good idea, should I instead wrap it in a separate method like `on_zoom()` that is called in `_zoom_callback()`, that the outer classes then call?

AThousandShips: Either that or use `friend` for the classes that need it, the callback should remain private

fire:
I like it! @TokageItLab What do you think?

editor/animation_track_editor.h
CookieBadger: good point, thanks!

Calinou (commented):
Tested locally, there are some issues:

- The zoom is slightly offset from the mouse cursor position (or the center when using the zoom bar). This seems to vary depending on the key column's size and the editor window width.
- Editor scales above 100% further offset the zoom (see the second video). Perhaps some multiplying/dividing by `EDSCALE` needs to be done somewhere.

This has the potential to be a big productivity boost, so it's encouraging to see this being worked on nonetheless :slightly_smiling_face: 

### 100% editor scale

https://github.com/godotengine/godot/assets/180032/49230ffd-cbec-4f86-b15c-c58b79ddb16c

### 200% editor scale

https://github.com/godotengine/godot/assets/180032/681dae9c-062d-4a2f-bc3e-18cfafe955a6

TokageItLab:
I think it is a good way for improvement. I will test it as soon as some of the points already pointed out are fixed.

CookieBadger:
Thank you for pointing the issues out, the zoomed timeline length was not calculated correctly. Should be fixed now!

Calinou (approved):
Tested locally (rebased on top of `master` d76c1d0e516fedc535a2e394ab780cac79203477), it works as expected (tested at 100% and 150% scale factors).

TokageItLab (changes requested):
When using the mouse wheel, it makes sense and works correctly based on the current cursor position. Nice work.

However, when using the slider in the lower right corner, the zoom should be based on the current position (e.g. Affter Effects bahave that). Especially when changing the zoom factor from small to large, I believe it is an odd behavior to have the end of the animation as the reference.

Current this PR:

https://github.com/godotengine/godot/assets/61938263/7bd825f4-0aa2-498b-9795-d96788092fd3

FYI Affter Effects:

https://github.com/godotengine/godot/assets/61938263/5d0adb05-4ca5-4452-a2f6-0fbb0985ae4e

CookieBadger:
Thanks for the suggestion @TokageItLab , makes sense. Something that I don't like about AfterEffects' approach though is that the timeline view switches position such that the cursor is centered upon zooming (0:06 in the video). I think this is quite distracting/disturbing, and it would be more appropriate if the zoom focused on the cursor, but didn't make the timeline jump around, like this:
AnimationPlayerZoomFocusOnPosition

However then, an edge case that makes that approach problematic, is what to do if the cursor position is outside the currently displayed range. It might again be distracting or disturbing if it automatically switches to the middle, but with the current approach it is unclear what the zoom is actually focusing on.
AnimationPlayerZoomFocusOnPositionOutside

Alternatively, I could test for this case, and then just zoom on the middle of the timeline as before, but I'd like to know what you think.

TokageItLab:
BTW, the first video in your reply is a little different from the behavior of AffterEffects: it should zooms based on the right or left edge of the window until the current position is centered. Then, after the current position is centered, it zooms based on the current position. The jump in window position is done when the window is shrunk (I also feel it is a bit odd in AffterEffects).

If the current position is not on the window, it is a matter of preference.

If the current position is not on the window and is based on the center of the screen as Yuri suggests, you may need to switch the behavior after the current position appears on the screen.

Personally, I don't care much about the timeline view switching positions, so if it is easiest to change the timeline view position, that is fine.

CookieBadger:
I now implemented the zooming similar to your suggestion, the way AfterEffects does it, such that the timeline stretches from the sides until the play position is in the middle.

https://github.com/godotengine/godot/assets/53947784/a37d4289-8688-4fae-b442-3f683902f56e

In the edge case of the cursor being outside, it now zooms into the middle, or zooms out from the middle until the play position comes into view, then again zooms out on the cursor

https://github.com/godotengine/godot/assets/53947784/50326234-8c5e-4449-846a-c0401758588c

I hope every case is properly addressed now.

CookieBadger:
@TokageItLab would you like to see any other changes?

TokageItLab (approved):
Now it seems to be working fine overall.

One concern is that the center appears to shake when zooming with the slider, so it would be nice if this could be removed, but it can be followed up later as another PR.

https://github.com/godotengine/godot/assets/61938263/1d46b00d-4e71-4308-b2e3-337a6d0efcd2

editor/animation_track_editor.cpp
```diff
@@ -1239,6 +1239,56 @@ void AnimationMultiTrackKeyEdit::set_use_fps(bool p_enable) {
 }
 
 void AnimationTimelineEdit::_zoom_changed(double) {
+	double zoom_pivot = 0; // point on timeline to stay fixed
```
AThousandShips, line 1242:
```suggestion
	double zoom_pivot = 0; // Point on timeline to stay fixed.
```
Per the [comment style](https://docs.godotengine.org/en/latest/contributing/development/code_style_guidelines.html#comment-style-guide). Same for all your code

CookieBadger:
Fixed comment style, added #define constant, and added a tiny offset to prevent the wiggling that @TokageItLab mentioned. I think the PR is ready for merge now.

https://github.com/godotengine/godot/assets/53947784/3ed16940-a9c6-4ed8-a512-2cee2199d0b9

editor/animation_track_editor.cpp
```diff
@@ -6479,6 +6539,7 @@ AnimationTrackEditor::AnimationTrackEditor() {
 	timeline->connect("length_changed", callable_mp(this, &AnimationTrackEditor::_update_length));
 
 	panner.instantiate();
+	panner->set_scroll_zoom_factor(1.02);
```
TokageItLab, line 6542:
```suggestion
	panner->set_scroll_zoom_factor(SCROLL_ZOOM_FACTOR);
```
Replace 1.02 in BezierEditor as well.

CookieBadger: woops, I missed that one, sorry. Resolved.

CookieBadger:
any other requests? I'd be happy to finish this by the end of the year 😄

editor/animation_track_editor.h
```diff
@@ -41,6 +41,8 @@
 #include "scene/gui/tree.h"
 #include "scene/resources/animation.h"
 
+#define SCROLL_ZOOM_FACTOR 1.02 // Zoom factor per mouse scroll in the animation editor. The closer to 1.0, the finer the control.
```
KoBeWi, line 44: This should be defined as `static constexpr float`, preferably inside the class.

editor/animation_track_editor.h
```diff
@@ -128,6 +130,8 @@ class AnimationMultiTrackKeyEdit : public Object {
 
 class AnimationTimelineEdit : public Range {
 	GDCLASS(AnimationTimelineEdit, Range);
+	friend class AnimationBezierTrackEdit;
+	friend class AnimationTrackEditor;
```
KoBeWi, line 134:
```suggestion
	GDCLASS(AnimationTimelineEdit, Range);

	friend class AnimationBezierTrackEdit;
	friend class AnimationTrackEditor;
```

editor/animation_track_editor.cpp
```diff
@@ -1239,6 +1239,59 @@ void AnimationMultiTrackKeyEdit::set_use_fps(bool p_enable) {
 }
 
 void AnimationTimelineEdit::_zoom_changed(double) {
```
KoBeWi, line 1241: Shouldn't the variables here be `float`? Most methods called here use float.

editor/animation_track_editor.cpp
```diff
@@ -1239,6 +1239,59 @@ void AnimationMultiTrackKeyEdit::set_use_fps(bool p_enable) {
...
+
+	int timeline_width_pixels = get_size().width - get_buttons_width() - get_name_limit();
+	double timeline_width_seconds = timeline_width_pixels / last_zoom_scale; // Length (in seconds) of visible part of timeline before zoom.
+	double updated_timeline_width_seconds = timeline_width_pixels / get_zoom_scale(); // Length after zoom.
+	double updated_timeline_half_width = updated_timeline_width_seconds / 2.0;
+	bool zooming = updated_timeline_width_seconds < timeline_width_seconds;
+
+	double timeline_left = get_value();
+	double timeline_right = get_value() + timeline_width_seconds;
+	double timeline_center = get_value() + timeline_width_seconds / 2.0;
```
KoBeWi, line 1253:
```suggestion
	double timeline_left = get_value();
	double timeline_right = timeline_left + timeline_width_seconds;
	double timeline_center = timeline_left + timeline_width_seconds / 2.0;
```

editor/animation_track_editor.cpp
```diff
@@ -1239,6 +1239,59 @@ void AnimationMultiTrackKeyEdit::set_use_fps(bool p_enable) {
...
+
+	if (zoom_callback_occured) { // Zooming with scroll wheel will focus on the position of the mouse.
+		double zoom_scroll_origin_norm = (zoom_scroll_origin.x - get_name_limit()) / timeline_width_pixels;
+		zoom_scroll_origin_norm = MAX(zoom_scroll_origin_norm, 0);
+		zoom_pivot = timeline_left + timeline_width_seconds * zoom_scroll_origin_norm;
+		zoom_pivot_delta = updated_timeline_width_seconds * zoom_scroll_origin_norm;
+		zoom_callback_occured = false;
+	} else { // Zooming with panner will depend on the current play position.
```
KoBeWi, line 1261:
```suggestion
	} else { // Zooming with slider will depend on the current play position.
```
Typo?

editor/animation_track_editor.cpp
```diff
@@ -1239,6 +1239,59 @@ void AnimationMultiTrackKeyEdit::set_use_fps(bool p_enable) {
...
+		const double epsilon = 0.00001;
```
KoBeWi, line 1262: You can use `CMP_EPSILON`.

editor/animation_track_editor.cpp
CookieBadger: I am not sure. `get_value()` returns a double, and `hscroll` (derives `Range`) expects a double, so I thought I'd just do all calculations with doubles. Should I change everything to be a float instead?

editor/animation_track_editor.h
CookieBadger: moved it down to AnimationTimelineEdit, I think that's where it belongs the most

KoBeWi (approved):

akien-mga:

CookieBadger:
Thank you!! 😊

PR #85150, opened 2023-11-20
Add Save As... option to EditorResourcePicker

### LINKED ISSUE #85045: Saving Resources Externally Requires Make Unique After Parent scene have been saved.
### Godot version

v4.2.rc1.official [ad72de508]

### System information

Windows 10

### Issue description

When creating a new Resource in a scene through the Inspector and then saving the scene it is in.  The Save button no longer prompts you with a dialogue to save the Resource externally.  

The workflow I'm used to in past versions is that this would pop up a "Save As" dialog equivalent and let me save it externally.  But it seems like saving now defaults to saving locally after the scene has been saved unless the resource has been made unique.  That's how it worked in 4.2 beta6 but the rc1 seems to have changed it.

### Steps to reproduce

- Add a Node with a Resource Property in the Inspector (I used Label with the Theme property)
- Set that Property to a New Instance of the Resource.
- Save the Scene
- Try to save the Resource through the Dropdown

[Resource Save Dialog.zip](https://github.com/godotengine/godot/files/13399460/Resource.Save.Dialog.zip)

KoBeWi:
Closes #85045
image


YuriSizov (approved):
LGTM

akien-mga:
Thanks!

PR #85156, opened 2023-11-20
CI: Install Android NDK r23c explicitly, with caching

akien-mga:
Related to https://github.com/godotengine/godot-cpp/pull/1314, where we don't rely on scons to download the NDK.

I wanted to enable the caching feature from https://github.com/nttld/setup-ndk, but it doesn't seem to work together with `link-sdk`. So the gain from this change is just some clarity, it may not speed things up.

*Edit:* Keeping as draft for now, blocked by https://github.com/nttld/setup-ndk/issues/518. If we can get the cache to work, then this becomes useful.

### DIFF
--- a/.github/workflows/android_builds.yml
+++ b/.github/workflows/android_builds.yml
@@ -26,6 +26,13 @@ jobs:
           distribution: temurin
           java-version: 17
 
+      - name: Install Android NDK r23c
+        uses: nttld/setup-ndk@v1
+        with:
+          ndk-version: r23c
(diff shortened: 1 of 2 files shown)


akien-mga:
> Keeping as draft for now, blocked by https://github.com/nttld/setup-ndk/issues/518. If we can get the cache to work, then this becomes useful.

The maintainer on setup-ndk seems busy and I don't want to add undue pressure. There aren't too much benefits from this PR as is so shelving it for now.

Issue #79545, opened 2023-07-16
Scene is invalid/corrupt according to the Godot editor, but works normally ingame


MewPurPur:
### Godot version

4.1

### System information

Ubuntu 22.04.2 LTS 22.04 - Vulkan (Forward+) - integrated Intel(R) HD Graphics 4400 (HSW GT2) () - Intel(R) Core(TM) i7-4600U CPU @ 2.10GHz (4 Threads)

### Issue description

This is the tscn file:

```
[gd_scene load_steps=2 format=3 uid="uid://bgrpmpmsi1ae2"]

[ext_resource type="Script" path="res://src/UI/LevelEditor/UserLevelsMenu.gd" id="1_0uvtn"]

[node name="UserLevels" type="Control" node_paths=PackedStringArray("initial_focus")]
layout_mode = 3
anchors_preset = 15
anchor_right = 1.0
anchor_bottom = 1.0
grow_horizontal = 2
grow_vertical = 2
script = ExtResource("1_0uvtn")
initial_focus = NodePath("MainContainer/Create")

[node name="MainContainer" type="VBoxContainer" parent="."]
layout_mode = 0
anchor_left = 0.5
anchor_top = 0.5
anchor_right = 0.5
anchor_bottom = 0.5
offset_left = -40.0
offset_top = -20.0
offset_right = 40.0
offset_bottom = 20.0
grow_horizontal = 2
grow_vertical = 2

[node name="Create" type="Button" parent="MainContainer"]
unique_name_in_owner = true
layout_mode = 2
button_mask = 0
text = "#create_new"

[node name="LevelContainer" type="VBoxContainer" parent="MainContainer"]
unique_name_in_owner = true
layout_mode = 2
theme_override_constants/separation = 9

[connection signal="pressed" from="MainContainer/Create" to="." method="_on_Create_pressed"]
```

It results in this:
![image](https://github.com/godotengine/godot/assets/85438892/41005428-bbd0-4e1e-a88b-5e8ca99d0438)

And the following error: "scene/resources/packed_scene.cpp:92 - Condition "nc == 0" is true. Returning: nullptr"

It also creates adds an empty scene in the scenes tabbar, and when I click on it, it automatically closes, throwing the following two errors:
- editor/editor_data.cpp:683 - Index p_idx = 2 is out of bounds (edited_scene.size() = 2).
- editor/editor_data.cpp:745 - Index p_idx = 2 is out of bounds (edited_scene.size() = 2).

[...]

PR #85159, opened 2023-11-20
Don't abort loading when `ext_resource` is missing

### LINKED ISSUE #85152: Can't open scene missing tres external resource
### Godot version

4.2 fa1fb2a

### System information

### Issue description

I have a scene that has a dependency on a tres resource (doesn't matter type). You can't open it, there is parse error. In this case:
```
Cannot open file 'res://new_gradient_texture_2d.tres'.
Failed loading resource: res://new_gradient_texture_2d.tres. Make sure resources have been imported by opening the project in the editor at least once.
C:\godot_source\scene/resources/resource_format_text.cpp:284 - res://s.tscn:6 - Parse Error: 
Failed loading resource: res://s.tscn. Make sure resources have been imported by opening the project in the editor at least once.
C:\godot_source\editor/editor_data.cpp:626 - Index p_idx = 1 is out of bounds (edited_scene.size() = 1).
```
A missing dependencies prompt shows normally, but pressing "Open Anyway" will just print the same errors and fail again. If an importable resource is missing (e.g. icon.svg), the scene will open normally.

### Steps to reproduce

1. Create scene
2. Add any text-based resource to it (GradientTexture for example)
3. Close the scene
4. Delete resource
5. Try to open the scene

[s.tscn.txt](https://github.com/godotengine/godot/files/13419197/s.tscn.txt)

### LINKED ISSUE #86154: project loading failing with failed loading resource / failed to instantiate scene  with cyclic preload
### Tested versions

- Reproducible with 4.2.1 stable

### System information

Godot v4.2.1.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3080 Laptop GPU (NVIDIA; 31.0.15.4617) - AMD Ryzen 9 5900HX with Radeon Graphics (16 Threads)

### Issue description

When opening an affected project it fails to load, for example: 

 ```
  scene/resources/resource_format_text.cpp:284 - res://a.tscn:6 - Parse Error: 
  Failed loading resource: res://a.tscn. Make sure resources have been imported by opening the project in the editor at least once.
 ```

when trying to load the scene from the FileSystem view:

 ```
  Failed to instantiate scene state of "res://a.tscn", node count is 0. Make sure the PackedScene resource is valid.
  editor/editor_data.cpp:626 - Index p_idx = 1 is out of bounds (edited_scene.size() = 1).
 ```

Only workaround I know: break the cycle in the script. In a larger project where I first found the issue it also helped to close all open scenes and just reload.

### Steps to reproduce

minimal steps:

1. Create a 2D scene A and a 2D scene B
2. Attach a script a.gd to A's root and a script b.gd to B'
3. in a add a variable b like this: var b = preload("res://b.tscn")
4. in b add a variable a like this: var a = preload("res://a.tscn"), which creates a cycle.
5. safe all scenes (ctrl + s)
6. from the menu execute Project -> Reload Current Project

[project_loading_bug_test.zip](https://github.com/godotengine/godot/files/13673175/project_loading_bug_test.zip)

KoBeWi:
Fixes #85152
*Bugsquad edit:* Fixes #86154
Resource loaders have inconsistent error values in case of missing files. If a loader returns error that indicates missing file, it's safe to ignore it; the resource will be replaced by null when you Open Anyway.

There are more errors related to files, not sure if ignore them all (the premise is technically the same).

### DIFF
--- a/core/variant/variant_parser.cpp
+++ b/core/variant/variant_parser.cpp
@@ -1026,7 +1026,10 @@ Error VariantParser::parse_value(Token &token, Variant &value, Stream *p_stream,
 				Ref<Resource> res;
 				Error err = p_res_parser->ext_func(p_res_parser->userdata, p_stream, res, line, r_err_str);
 				if (err) {
-					return err;
+					// If the file is missing, the error can be ignored.
+					if (err != ERR_FILE_NOT_FOUND && err != ERR_CANT_OPEN) {
+						return err;
+					}
 				}
 
 				value = res;


RandomShaper:
I think that _Open anyway_ should set (and later unset) `ResourceLoader::set_abort_on_missing_resources()`. To me, that's already existing functionality to be able to load despite external resources missing. However, the branch when that's disabled, which calls `ResourceLoader::notify_dependency_error()` should maybe set `err` (the one that will be returned) to `OK`.

KoBeWi:
Changed to use `set_abort_on_missing_resources()`. The name feels backwards, unless I'm misunderstanding it.

RandomShaper:
Quoting current usages:
https://github.com/godotengine/godot/blob/c2f8fb301537a5d688d201178985963282b4f9c3/scene/resources/resource_format_text.cpp#L160-L167
https://github.com/godotengine/godot/blob/c2f8fb301537a5d688d201178985963282b4f9c3/scene/resources/resource_format_text.cpp#L469-L476
https://github.com/godotengine/godot/blob/c2f8fb301537a5d688d201178985963282b4f9c3/core/io/resource_format_binary.cpp#L455-L460
https://github.com/godotengine/godot/blob/c2f8fb301537a5d688d201178985963282b4f9c3/core/io/resource_format_binary.cpp#L688-L693

So, _abort on missing resources_, if `true`, means reporting an error to the caller. If `false`, it means send the broken dep error and otherwise ignore.

---

Now, `EditorNode::load_resource` and `EditorNode::load_scene` already provide the ignore-broken-deps functionality (`bool p_ignore_broken_deps`). I'd say that it's the responsibility of those two functions to use `ResourceLoader::set_abort_on_missing_resources()` depending on the value of that parameter:
```
if (p_ignore_broken_deps) {
    ResourceLoader::set_abort_on_missing_resources(false);
}

// Load, etc.

if (!p_ignore_broken_deps && dependency_errors.has(p_resource)) {
    // Show dependency dialog. There will be dep errors, but we won't enter this branch.
}

if (p_ignore_broken_deps) {
    ResourceLoader::set_abort_on_missing_resources(true); // Restore.
}
```

What do you think?

KoBeWi:
Makes sense, changed.

editor/editor_node.cpp
```diff
@@ -1218,7 +1218,15 @@ Error EditorNode::load_resource(const String &p_resource, bool p_ignore_broken_d
 
 	Ref<Resource> res;
 	if (ResourceLoader::exists(p_resource, "")) {
+		if (p_ignore_broken_deps) {
+			ResourceLoader::set_abort_on_missing_resources(true);
```
RandomShaper, line 1222: Are you sure this shouldn't be called with `false`?

`false` would cause `ResourceLoader::notify_dependency_error()` to be called instead of aborting and returning an error code. And by `p_ignore_broken_deps == true`, the branch below to open the deps dialog won't be run.

KoBeWi: `false` is the default value.
I tested and `true` is the correct one to use. That's why I said it's backwards 🙃

RandomShaper: Then please help me follow the logic, because I think I'm misunderstanding it at some point.

1. `EditorNode::load_resource()` is called with `p_ignore_broken_deps = true`. So, what you want is the thing to be loaded with missing external resources as `null` and no dependency error dialog.
1. `ResourceLoader::set_abort_on_missing_resources(true);`
1. `ResourceLoaderText` fails to load an external resource and checks `ResourceLoader::get_abort_on_missing_resources()`.
1. Since it's `true`, it returns `ERR_FILE_MISSING_DEPENDENCIES` and cancels further loading. That's not what we wanted, is it?

Alternatively:
1. Same as before.
1. `ResourceLoader::set_abort_on_missing_resources(false);`
1. `ResourceLoaderText` fails to load an external resource and checks `ResourceLoader::get_abort_on_missing_resources()`.
1. Since it's `false`, it reports the broken dep via `notify_dependency_error()`. `EditorNode::load_resource()` will ignore `dependency_errors` because `p_ignore_broken_deps == true` and won't show the broken deps dialog.

What am I missing here?

KoBeWi: https://github.com/godotengine/godot/blob/7022271291a3d2a9cbd6a223d22a29fd775dfc5d/scene/resources/resource_format_text.cpp#L278-L290
When the parser returns any other error than ERR_FILE_EOF or ERR_FILE_MISSING_DEPENDENCIES (the latter appears with abort = true), the loader will return empty scene.

Not sure about binary format, the issue still persists for binary scenes it seems.

KoBeWi:
Reverted to my original fix. The only call for `set_abort_on_missing_resources()` (not counting some web code) is this:
https://github.com/godotengine/godot/blob/1faf2f5bfe97edcb606d82b81094cb35e2b8e18c/editor/editor_node.cpp#L6919
The the abort is already disabled in the editor, and according to your comments it is working as intended, so the problem is text format code not handling it properly. Also the new fix was breaking binary format, so it was bad anyway.

RandomShaper:
`ResourceLoaderBinary` will error with `ERR_FILE_MISSING_DEPENDENCIES` in case an external resource can't be loaded, regardless the reason. It may make sense that loading of text resources was more permissive because of certain specific needs in the editor. However, for consistency I'd still model that with some kind of global setting that both can obey. If abort-on-missing-resources is too limiting by being a boolean, could it maybe be a three-item enum, with a new value that would behave like the current `false`, only that it won't either abort nor report broken deps in cases like a missing file?

In any case, I'm willing to attempt to add such a refactor myself in another PR. If this PR is already enough to fix a known issue, I won't oppose to merging it and we can keep the discussion over actual code showing what I'm trying to tell.

PS: I've realized why abort-on-missing-resources looks reversed. It depends on what is aborted. The editor sets to `false` so the resource loader _doesn't_ abort, but the editor itself _will_ by showing the broken deps dialog instead of the scene.

RandomShaper (approved):

akien-mga:
Thanks!

tlobig:
fixes #86154

neruthes:
I had a similar problem where no preload is involved. My guess is that strong-type variables are causing cyclic GDScript parsing. Consider some code like this:

```
# hive.gd
extends Node2D
class_name MyHive
var bee : MyBee = null
```

```
# bee.gd
extends Node2D
class_name MyBee
var hive : MyHive = null
```

I wish that this PR also fixes any problem that may arise in this hypothetical situation. I will report back if the problem persists in 4.3.

KoBeWi:
GDScript does support cyclic dependencies since 4.0. Also it shouldn't be affecting scenes...

neruthes:
> GDScript does support cyclic dependencies since 4.0. Also it shouldn't be affecting scenes...

That is a surprise to me. I will try making a minimal working example.

KoBeWi:
Though by cyclic dependencies I mean scripts only. Preloading other resources, especially scenes, in a cyclic way won't work and will make the scenes non-openable until you remove the preloads.

neruthes:
> Though by cyclic dependencies I mean scripts only. Preloading other resources, especially scenes, in a cyclic way won't work and will make the scenes non-openable until you remove the preloads.

Thanks for the clarification. I had another look into my scripts and figured out that a script contained a preload which I failed to notice.

I found the correct way to reproduce the problem and you may [have a look](https://github.com/neruthes/godot-bug-mwe-1).

In summary:
- Hive.gd (`class_name Hive`) contains preload Bee.tscn.
- Bee.tscn uses Bee.gd (`class_name Bee`).
- Bee.gd has a varaible typed Hive (`var master : Hive = null`).
- Cannot open Bee.tscn in editor, but running game with F5 is ok.

image

KoBeWi:
Thanks, there's already an existing issue about this bug, I attached your project there: https://github.com/godotengine/godot/issues/79545#issuecomment-1902264615

neruthes:
> Thanks, there's already an existing issue about this bug, I attached your project there: [#79545 (comment)](https://github.com/godotengine/godot/issues/79545#issuecomment-1902264615)

Thank you. You have done great help in this PR.

PR #85162, opened 2023-11-21
Update joint after GrooveJoint2D's length is set

### LINKED ISSUE #85144: set_length of GrooveJoint2D doesn't change its length
### Godot version

v4.1.3.stable.official [f06b6836a]

### System information

Apple Ventura 13.5, Forward+

## Setting the `length` of a `GrooveJoint2D` doesn't work. The simulated length doesn't actually get updated. It keeps the initially set length.

---

## My guess
It seems the joint's parameters [are not actually changed](https://github.com/godotengine/godot/blob/ac29e8e3fed8a8c8d74a267a60eee75b979bb994/scene/2d/joint_2d.cpp#L436), unlike in e.g. [PinJoint2D::set_softness](https://github.com/godotengine/godot/blob/ac29e8e3fed8a8c8d74a267a60eee75b979bb994/scene/2d/joint_2d.cpp#L294).

[There is](https://github.com/godotengine/godot/blob/master/servers/physics_2d/godot_physics_server_2d.cpp) a `pin_joint_set_param` and a `damped_spring_joint_set_param`, but no `groove_joint_set_param`. If I can find the time I will try to add this function and integrate it into `GrooveJoint2D::set_length`.

## Test setup:
- The left rigid body is connected to a groove joint with a length of 200px ...
- ... and the right one to a groove joint with the length of 50px **but** a script sets the length also to 200px

```
func _ready():
	$Right/GrooveJoint.set_length(200)
```

Screenshot 2023-11-20 at 17 50 44

### Expected result
Both rigid bodies should behave the exact same.

### Result
The visualisation of the right groove joint is updated correctly (yellow line), but it doesn't fall lower than the initially specified 50px.

Screenshot 2023-11-20 at 17 58 00

---

PS: I seem to become the guy that opens all the 2D physics bugs 😅 

### Steps to reproduce

1. Create a `GrooveJoint2D` connected to two bodies
2. Create a script that changes the `length` of the GrooveJoint (neither `length` nor `set_length` does work)
3. Watch the simulation play

[groove-joint-bug.zip](https://github.com/godotengine/godot/files/13417593/groove-joint-bug.zip)

jsjtxietian:
Fixes #85144

### DIFF
--- a/doc/classes/GrooveJoint2D.xml
+++ b/doc/classes/GrooveJoint2D.xml
@@ -11,9 +11,11 @@
 	<members>
 		<member name="initial_offset" type="float" setter="set_initial_offset" getter="get_initial_offset" default="25.0">
 			The body B's initial anchor position defined by the joint's origin and a local offset [member initial_offset] along the joint's Y axis (along the groove).
+			[b]Note:[/b] Setting initial offset will cause the joint to be reconfigured.
 		</member>
 		<member name="length" type="float" setter="set_length" getter="get_length" default="50.0">
 			The groove's length. The groove is from the joint's origin towards [member length] along the joint's local Y axis.
+			[b]Note:[/b] Setting the groove's length will cause the joint to be reconfigured.
 		</member>
 	</members>
 </class>
--- a/scene/2d/joint_2d.cpp
+++ b/scene/2d/joint_2d.cpp
@@ -436,6 +436,10 @@ void GrooveJoint2D::_configure_joint(RID p_joint, PhysicsBody2D *body_a, Physics
 void GrooveJoint2D::set_length(real_t p_length) {
 	length = p_length;
 	queue_redraw();
+	if (is_configured()) {
+		_disconnect_signals();
+		_update_joint(false);
+	}
 }
 
 real_t GrooveJoint2D::get_length() const {
@@ -445,6 +449,10 @@ real_t GrooveJoint2D::get_length() const {
 void GrooveJoint2D::set_initial_offset(real_t p_initial_offset) {
 	initial_offset = p_initial_offset;
 	queue_redraw();
+	if (is_configured()) {
+		_disconnect_signals();
+		_update_joint(false);
+	}
 }
 
 real_t GrooveJoint2D::get_initial_offset() const {


doc/classes/GrooveJoint2D.xml
```diff
@@ -14,6 +14,7 @@
 		</member>
 		<member name="length" type="float" setter="set_length" getter="get_length" default="50.0">
 			The groove's length. The groove is from the joint's origin towards [member length] along the joint's local Y axis.
+			[b]Note:[/b] Set the groove's length will cause the joint to be reconfigured.
```
akien-mga, line 17:
```suggestion
			[b]Note:[/b] Setting the groove's length will cause the joint to be reconfigured.
```

scene/2d/joint_2d.cpp
```diff
@@ -435,6 +435,8 @@ void GrooveJoint2D::_configure_joint(RID p_joint, PhysicsBody2D *body_a, Physics
 
 void GrooveJoint2D::set_length(real_t p_length) {
 	length = p_length;
+	_disconnect_signals();
```
AThousandShips, line 438: You should add this to `set_initial_offset` as well

jsjtxietian: Yes, also seems other joint2d are fine.

stephanbogner:
The PR has the [known issues](https://github.com/godotengine/godot/issues/85144#issuecomment-1825504765) when changing parameters during runtime ... but other than it looks like it's working well and would be a great improvement.

**I tested some more cases (see file below):**
- What if the joint is rotated?
- What if it is upside down?
- What if it gets shortened lower than the initial offset? (Answer: Object gets pushed up)
- → **They all worked as expected 🎉** 

---

[groove-joint-test.zip](https://github.com/godotengine/godot/files/13459140/groove-joint-test.zip)

jsjtxietian:
> * They all worked as expected 🎉

Thanks for your help!

At least we can update the doc to mention this behaviour. Might worth investigating more about the wired behaviour.

Calinou:
Should the same be done in 3D?

jsjtxietian:
> Should the same be done in 3D?

There's no groove joint in 3D.

Issue #85114, opened 2023-11-20
4.2rc1 - New project - Editor hangs when opening on Mac M1 Pro


an0maly33:
### Godot version

4.2rc1

### System information

MacOS 13.6

### Issue description

I just downloaded the latest version - 4.2rc1 and I can't get the editor to start.  The project manager comes up, I create a new project, but when I try to edit, the project manager closes and the editor tries to start but it just hangs.  Previously used 4.1.x with no issues.

### Steps to reproduce

Create a new project, try to edit it.

### Minimal reproduction project

N/A

akien-mga:
What renderer did you choose for the new project? Do you have the same problem with all 3 options?

If it only happens with the `GL Compatibility` one, can you try to edit the project from the command line with:
```
path/to/Godot --rendering-driver opengl3
```
(to force overriding the default which should be `opengl3_angle` on macOS, and may be the source of your issue)

an0maly33:
I tried Mobile and Forward.  Seems fine on my M2 mac with OS 14.2 though.

bruvzg:
I can't reproduce it (M1, macOS 14.1.1, with any renderer).

akien-mga:
Can you confirm that 4.1.3-stable works fine on that macOS 13.6 machine?

If so, can you test intermediate dev snapshots/betas between 4.1-stable and 4.2-rc1 to pinpoint when it started failing? E.g. you can test 4.2-beta1 and based on the result test earlier or later builds (i.e. bisect the potential regression using official builds).

akien-mga:
You can find all past dev and beta snapshots on latest pages of https://godotengine.org/blog, but there's also the archive page which lists them all: https://godotengine.org/download/archive/

an0maly33:
4.2b1 Starts up fine with forward and mobile renderers but hangs when quitting.  Any version past that hangs when the editor tries to launch.

bruvzg:
> Apparently there was a dialog that was hidden somewhere to allow xcode to install some stuff. 

Godot doesn't need Xcode for most of the stuff, so it's probably should be considered as a bug anyway. Most likely caused by iOS one-click display trying to get running simulator list from Xcode.

1992leiting:
> im on m2 pro and i cant even open on 4.2 godot. it just hangs there.
> 
> im able to run 4.1.3 but not 4.2 stable.
> 
> my OS: 13.4 and im using m2 pro 12 cpu
> 
[...]

PR #85168, opened 2023-11-21
[iOS] Check if Xcode is installed in one-click deploy code.

bruvzg:
Adds a Xcode install check to avoid unnecessary popups asking to install Xcode command line tools while enumerating running iOS simulator devices.

See https://github.com/godotengine/godot/issues/85114#issuecomment-1819981211


akien-mga (approved):
Looks good to me.

akien-mga:
Thanks!

PR #85178, opened 2023-11-21
Add unit test runner for autocompletion

HolonProduction:
To prevent GDScript changes from breaking auto completion capabilities.

> This PR adds the possibility to write Auto Completion tests. However it does not add any tests. I want to have the runner merged before working on test cases.


HolonProduction:
Just noticed that there is no Github CI that runs tests with mono enabled. Would this be possible? I see no reliable way to test auto completion with `SCRIPT` types without using another script language besides GDScript. Because any way of obtaining a SCRIPT typed reference to a GDScript might break if we decide to intelligently upgrade the type in the future.

AThousandShips:
> Just noticed that there is no Github CI that runs tests with mono enabled

There is, `GHA / 🐧 Linux / Editor w/ Mono (target=editor)`

AThousandShips:
This is because of issues, see the workflow file comment:
> Disabled due freeze caused by mix Mono build and CI

AThousandShips:
I've tested running a few times with the tests enabled and don't see any freezing, but I think this should be enabled and tested extensively in a separate PR to avoid any problems

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,201 @@
...
+				}
+			}
+		}
+		next = dir->get_next();
+	}
+}
+
+TEST_SUITE("[Modules][GDScript][Completion]") {
+	TEST_CASE("[Editor] Check suggestion list") {
+		// Set all editor settings that code completion relies on
```
AThousandShips, line 189:
```suggestion
		// Set all editor settings that code completion relies on.
```

HolonProduction:
This should be ready for a first round of review now, since I added the features which I wanted to add.

What still remains to do:
- squash the commits (will do at the end, but while in review having the history might be helpful)
- remove dummy tests (will do at the end, but they are needed for testing)

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+	if (expected.get("display", got.display) != got.display) {
+		return false;
+	}
+	if (expected.get("insert_text", got.insert_text) != got.insert_text) {
+		return false;
+	}
+	if (expected.get("kind", got.kind) != Variant(got.kind)) {
+		return false;
+	}
+	if (expected.get("location", got.location) != Variant(got.location)) {
```
AThousandShips, line 64:
```suggestion
static bool match_option(const Dictionary &p_expected, ScriptLanguage::CodeCompletionOption p_got) {
	if (p_expected.get("display", p_got.display) != p_got.display) {
		return false;
	}
	if (p_expected.get("insert_text", p_got.insert_text) != p_got.insert_text) {
		return false;
	}
	if (p_expected.get("kind", p_got.kind) != Variant(p_got.kind)) {
		return false;
	}
	if (p_expected.get("location", p_got.location) != Variant(p_got.location)) {
```

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+		return false;
+	}
+	return true;
+}
+
+void to_dict_list(Variant p_variant, List<Dictionary> &p_list) {
+	if (p_variant.is_array()) {
+		Array arr = static_cast<Array>(p_variant);
```
AThousandShips, line 72:
```suggestion
		Array arr = p_variant;
```
Unnecessary, and very unusual

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+		while (arr.size() > 0) {
+			Variant E = arr.pop_front();
+			if (E.get_type() == Variant::DICTIONARY) {
+				p_list.push_back(static_cast<Dictionary>(E));
```
AThousandShips, line 76:
```suggestion
				p_list.push_back(E);
```
Should be sufficient, or
```suggestion
				p_list.push_back(E.operator Dictionary());
```

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+			}
+		}
+	}
+}
+
+void test_directory(const String &p_dir) {
+	Error err = OK;
+	Ref<DirAccess> dir(DirAccess::open(p_dir, &err));
```
AThousandShips, line 84:
```suggestion
	Ref<DirAccess> dir = DirAccess::open(p_dir, &err);
```
Cleaner IMO, and most commonly used

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+				if (dir->file_exists(next.get_basename() + ".tscn")) {
+					String project_path = "res://completion";
+					Ref<PackedScene> scene = ResourceLoader::load(project_path.path_join(next.get_basename() + ".tscn"), "PackedScene");
+					if (scene.is_valid()) {
+						owner = scene->instantiate();
+					}
+				}
+
+				GDScriptLanguage::get_singleton()->complete_code(code, path.path_join(next), owner, &options, forced, call_hint);
+				String contains_excluded = "";
```
AThousandShips, line 150:
```suggestion
				String contains_excluded;
```
Unnecessary default declaration

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+				for (ScriptLanguage::CodeCompletionOption &option : options) {
+					for (Dictionary &E : exclude) {
```
AThousandShips, line 152:
```suggestion
					for (const Dictionary &E : exclude) {
```

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+						if (match_option(E, option)) {
+							contains_excluded = option.display;
+							break;
+						}
+					}
+					if (!contains_excluded.is_empty()) {
+						break;
+					}
+
+					for (Dictionary &E : include) {
```
AThousandShips, line 162:
```suggestion
					for (const Dictionary &E : include) {
```

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+	if (expected.get("kind", got.kind) != Variant(got.kind)) {
+		return false;
+	}
+	if (expected.get("location", got.location) != Variant(got.location)) {
+		return false;
+	}
+	return true;
+}
+
+void to_dict_list(Variant p_variant, List<Dictionary> &p_list) {
```
AThousandShips, line 70:
```suggestion
static void to_dict_list(Variant p_variant, List<Dictionary> &p_list) {
```

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+		while (arr.size() > 0) {
+			Variant E = arr.pop_front();
+			if (E.get_type() == Variant::DICTIONARY) {
+				p_list.push_back(static_cast<Dictionary>(E));
+			}
+		}
+	}
+}
+
+void test_directory(const String &p_dir) {
```
AThousandShips, line 82:
```suggestion
static void test_directory(const String &p_dir) {
```

AThousandShips (approved):
LGTM, seems to pass tests and the tests makes sense to me

modules/gdscript/tests/scripts/completion/foo.notest.gd
```diff
@@ -0,0 +1,6 @@
+extends Node
```
vnen, line 1: I don't like this file called `notest` when it is an actual test. If I understand it correctly this is so the regular test runner ignore this file, but in this case it's better to hard code the `completion` folder to be excluded. There might be completion tests where the `notest` is required (for instance, when making an auxiliary script to see if its properties are listed) so it would be better if the `notest` tag is respected for completion tests as well.

modules/gdscript/tests/scripts/completion/foo.notest.gd
```diff
@@ -0,0 +1,6 @@
...
+
+var test = 1
+
+func _ready():
+    $N￿
```
vnen, line 6: This `0xFFFF` character might be annoying to add as there is no an easy way to insert it besides copying and pasting, and it's inconsistent how it's shown across editors. Though I can't think of a better alternative for this at the moment.

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+		return false;
+	}
+	return true;
+}
+
+static void to_dict_list(Variant p_variant, List<Dictionary> &p_list) {
+	if (p_variant.is_array()) {
+		Array arr = p_variant;
+		while (arr.size() > 0) {
+			Variant E = arr.pop_front();
```
vnen, line 74: Why not use a regular `for` loop here? Doing `pop_front()` is resizing the array at every iteration, which is much more costly than just reading the value. This array isn't used afterwards anyway, so there's no need to be emptied.

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+		return false;
+	}
+	if (p_expected.get("location", p_got.location) != Variant(p_got.location)) {
+		return false;
+	}
+	return true;
+}
+
+static void to_dict_list(Variant p_variant, List<Dictionary> &p_list) {
+	if (p_variant.is_array()) {
```
vnen, line 71: Maybe this should be an `ERR_FAIL_COND` since AFAICT this should never be called with anything other than an array and we would probably want to know if that happens.

vnen (changes requested):
The idea itself looks good to me, I just found some small details.

I also wonder if we can have an easy way to create the `cfg` files since writing them by hand might be tricky and prone to typos, especially if they get big (though tests should probably be small anyway).

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,200 @@
...
+				CHECK(include.is_empty());
+
+				String expected_call_hint = conf.get_value("result", "call_hint", call_hint);
+				bool expected_forced = conf.get_value("result", "forced", forced);
+
+				CHECK(expected_call_hint == call_hint);
+				CHECK(expected_forced == forced);
+
+				if (owner) {
+					owner->queue_free();
```
vnen, line 179: This should be just `free()`. There's no queue running during the tests and the node won't be used again anyway, so it needs to freed right away.

vnen:
Thanks for working on this, BTW. It is something on my mind for a while so I'm glad to have tests for completion.

modules/gdscript/tests/scripts/completion/foo.notest.gd
HolonProduction: Yeah I also thought about this but in the end decided against it since the same approach was choose for the LSP tests. I wonder whether it would make sense to add some naming patter for folders like this to prevent hard coding.

Maybe `completion.norun` and `lsp.norun`. But hard coding could be fine in the context of tests.

modules/gdscript/tests/scripts/completion/foo.notest.gd
HolonProduction: We could choose some UTF8 emoji and replace it with 0xFFFF. The use of this emoji in the test would just be banned. Maybe ➡️ or 🖱️. But I'm not sure whether there could be problems because those emojis are longer then normal characters.

modules/gdscript/tests/test_completion.h
HolonProduction: Not sure what the usage guidelines for `ERR_FAIL_COND` are. A malformed .cfg file could result in this not being an array. Not sure whether a malformed test config should trigger ERR_FAIL_COND. But failing silently is bad anyway

modules/gdscript/tests/test_completion.h
AThousandShips: Most tests just use `memdelete` in fact, I suspect it should be used here as well

HolonProduction:
> I also wonder if we can have an easy way to create the `cfg` files since writing them by hand might be tricky and prone to typos, especially if they get big (though tests should probably be small anyway).

Generating them through CLI like the other GDScript test would be quite inconvenient I think. It sounds more like an entry for the tools menu which gives you some UI that allows to configure the test and spits out the cfg and maybe also the script with the inserted sentinel char. But since it would be dev only, having it in the menu seems wrong.

Maybe I will write a little editor plugin for it, but that is something to think about when actually writing test cases. I can also imagine that copy and pasting an existing cfg will be quite sufficient for most cases.

modules/gdscript/tests/test_completion.h
```diff
@@ -0,0 +1,201 @@
...
+
+	while (!next.is_empty()) {
+		if (dir->current_is_dir()) {
+			if (next == "." || next == "..") {
+				next = dir->get_next();
+				continue;
+			}
+			test_directory(path.path_join(next));
+		} else {
+			if (next.ends_with(".gd")) {
```
AThousandShips, line 103: This can be merged, as it is the additional nesting hurts readability IMO

modules/gdscript/tests/scripts/completion/foo.notest.gd
HolonProduction: Hard coded the test runner to ignore `completion` and also did the same for `lsp` to keep it consistent.

vnen (approved):
LGTM

akien-mga:
Thanks!

PR #85180, opened 2023-11-21
[Core] Prevent copying of `SelfList` and `SelfList::List`

AThousandShips:
Copying of these types is unsafe and should be detected

Also removed unnecessary constructors for `TileMap` `DebugQuadrant` and `RenderingQuadrant` which used copying of `SelfList::List`

Bug identified while researching:
* https://github.com/godotengine/godot/issues/84892

Unable to reliably replicate this crash so can't confirm this is a fix, will continue to investigate that fix

### DIFF
--- a/core/templates/self_list.h
+++ b/core/templates/self_list.h
@@ -159,6 +159,9 @@ public:
 		_FORCE_INLINE_ SelfList<T> *first() { return _first; }
 		_FORCE_INLINE_ const SelfList<T> *first() const { return _first; }
 
+		// Forbid copying, which has broken behavior.
+		void operator=(const List &) = delete;
+
 		_FORCE_INLINE_ List() {}
 		_FORCE_INLINE_ ~List() {
 			// A self list must be empty on destruction.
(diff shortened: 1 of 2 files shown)


AThousandShips:
The `TileMap` data types should not be a problem with this as they are `RefCounted` and therefore never invoke their copy constructor

core/templates/self_list.h
```diff
@@ -41,6 +41,9 @@ class SelfList {
 		SelfList<T> *_first = nullptr;
 		SelfList<T> *_last = nullptr;
 
+		// Forbid copying, which has broken behavior.
+		void operator=(const List &p_other) {}
```
RandomShaper, line 45: What do you think about keeping if `public` but `deleted`? More idiomatic, perhaps.

AThousandShips: I was intending to but matched the class template, I'm okay with either way and I prefer the `delete` syntax myself, more informative

Will update

RandomShaper (approved):

akien-mga:
Thanks!

AThousandShips:
Thank you!

groud:
> The TileMap data types should not be a problem with this as they are RefCounted and therefore never invoke their copy constructor

Did you verify that ? Because, as far as I remember, those constructors were needed to avoid several issues. Though I guess forbidding the copy as you did would probably trigger a compilation error :thinking: .

Anyway, that was just to mention this is probably something we need to keep an eye on.


 
I am not sure why

AThousandShips:
I did both by compilation and by that there's no such semantics in the engine, they're created and shared, not copied, were these data elements always ref counted? Looks like a remnant of a version not using that

groud:
>  were these data elements always ref counted? Looks like a remnant of a version not using that

Ah that's a good point, that might be it! Good call.

AThousandShips:
If the two quadrant types were built from the old `TileMapQuadrant` then that's probably why, it was non-ref counted and had a copy assignment with the same comment 🙂

PR #85181, opened 2023-11-21
Restored Control properties when you undo a parenting of a Control to a Container

### LINKED ISSUE #44629: Undo fails to fully restore parenting control to container
**Godot version:**
3.2.1 stable

Windows 10

**Issue description:**
Undoing an operation that reparents a control to a container undoes the parenting but doesn't undo the modifications that the container applied to the child control's anchors, margins, etc.

**Steps to reproduce:**
- Have some controls set up.
- Add a a container node.
- Reparent control to container.
- Undo -> the re-parented control finds its original place jn the hierarchy but keeps the anchors, margins, position and size the container assigned to it.
- Hope you remember your setup, or reload the scene.

Trivial.

TheSofox:
Fixes  #44629.

Previously if you added a Control to a Container that did automatic layout (Such as Button to VBoxLayout), then if you hit "Undo", the parenting would be undone, but the anchors and the size of the control would remain whatever the Container had set them to. This update fixes that.

### DIFF
--- a/editor/scene_tree_dock.cpp
+++ b/editor/scene_tree_dock.cpp
@@ -2176,6 +2176,7 @@ void SceneTreeDock::_do_reparent(Node *p_new_parent, int p_position_in_parent, V
 		}
 
 		int child_pos = node->get_index(false);
+		bool reparented_to_container = Object::cast_to<Container>(new_parent) && Object::cast_to<Control>(node);
 
 		undo_redo->add_undo_method(node->get_parent(), "add_child", node, true);
 		undo_redo->add_undo_method(node->get_parent(), "move_child", node, child_pos);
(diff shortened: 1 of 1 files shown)


YuriSizov:
Controls, in a rather hackish manner, have dedicated editor-only methods that can give you the entire state of the control which you can use to restore it later. I think this is what you should use here.

TheSofox:
It was that easy the whole time? Well, I've made the change, pushed, and on my end it works even better than before.

editor/scene_tree_dock.cpp
```diff
@@ -2195,6 +2195,10 @@ void SceneTreeDock::_do_reparent(Node *p_new_parent, int p_position_in_parent, V
 				undo_redo->add_undo_method(node, "set_position", Object::cast_to<Control>(node)->get_position());
 			}
 		}
+
+		if (Object::cast_to<Container>(new_parent) && Object::cast_to<Control>(node)) {
+			undo_redo->add_undo_method(node, "_edit_set_state", Object::cast_to<Control>(node)->_edit_get_state());
```
KoBeWi, line 2200: This a bit conflicts with the Control operation above (`set_position`). When adding node to container, its position is overriden by the set state, so setting it is redundant.

KoBeWi (approved):

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.1.

PR #85183, opened 2023-11-21
Remove clear renderpass for CSM, and change each shadow split renderpass to `loadop_clear` with the correct render area.

oconnorkg:
Removed clear renderpass for CSM, and changed each shadow split renderpass to loadop_clear with the correct renderarea.
Reduces average frame time by approximately 500μs on the default view of "godot_4_mobile_test" on a Mali G78.

### DIFF
--- a/drivers/vulkan/rendering_device_vulkan.cpp
+++ b/drivers/vulkan/rendering_device_vulkan.cpp
@@ -6847,18 +6847,10 @@ Error RenderingDeviceVulkan::_draw_list_render_pass_begin(Framebuffer *framebuff
 	render_pass_begin.pNext = nullptr;
 	render_pass_begin.renderPass = render_pass;
 	render_pass_begin.framebuffer = vkframebuffer;
-	/*
-	 * Given how API works, it makes sense to always fully operate on the whole framebuffer.
-	 * This allows better continue operations for operations like shadowmapping.
 	render_pass_begin.renderArea.extent.width = viewport_size.width;
 	render_pass_begin.renderArea.extent.height = viewport_size.height;
 	render_pass_begin.renderArea.offset.x = viewport_offset.x;
 	render_pass_begin.renderArea.offset.y = viewport_offset.y;
-	*/
-	render_pass_begin.renderArea.extent.width = framebuffer->size.width;
-	render_pass_begin.renderArea.extent.height = framebuffer->size.height;
-	render_pass_begin.renderArea.offset.x = 0;
-	render_pass_begin.renderArea.offset.y = 0;
 
 	Vector<VkClearValue> clear_values;
 	clear_values.resize(framebuffer->texture_ids.size());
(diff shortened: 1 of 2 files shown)


drivers/vulkan/rendering_device_vulkan.cpp
```diff
@@ -6847,18 +6847,10 @@ Error RenderingDeviceVulkan::_draw_list_render_pass_begin(Framebuffer *framebuff
 	render_pass_begin.pNext = nullptr;
 	render_pass_begin.renderPass = render_pass;
 	render_pass_begin.framebuffer = vkframebuffer;
-	/*
-	 * Given how API works, it makes sense to always fully operate on the whole framebuffer.
-	 * This allows better continue operations for operations like shadowmapping.
-	render_pass_begin.renderArea.extent.width = viewport_size.width;
+    render_pass_begin.renderArea.extent.width = viewport_size.width;
```
AThousandShips, line 6850:
```suggestion
	render_pass_begin.renderArea.extent.width = viewport_size.width;
```
You need to have tabs for indentation in your editor

drivers/vulkan/rendering_device_vulkan.cpp
```diff
@@ -6847,18 +6847,10 @@ Error RenderingDeviceVulkan::_draw_list_render_pass_begin(Framebuffer *framebuff
 	render_pass_begin.pNext = nullptr;
 	render_pass_begin.renderPass = render_pass;
 	render_pass_begin.framebuffer = vkframebuffer;
-	/*
-	 * Given how API works, it makes sense to always fully operate on the whole framebuffer.
-	 * This allows better continue operations for operations like shadowmapping.
 	render_pass_begin.renderArea.extent.width = viewport_size.width;
```
BastiaanOlij, line 6850: So the problem here is that, according to @reduz's experimentation, there is a performance penalty for this on desktop. We need to make this switchable so we only apply this with the Vulkan mobile renderer.

BastiaanOlij (dismissed):
Needs to be a bit more testing, I did a bunch more work around this with cascaded updating of directional shadow maps (but that's currently on hold) that would suggest this should indeed be an improvement.

We'll also need to look into making similar changes for positional lights but when I last tried I didn't get any clear measurements to see if it improved things.

drivers/vulkan/rendering_device_vulkan.cpp
Zireael07: "Desktop"? The issue is talking about Mali chips?

AThousandShips: Yes, that's why separating it and only applying it for mobile builds was suggested? It *helping* with those chips, but has a performance penalty on desktop.

Calinou: Remember that the Mobile rendering method can be used on desktop platforms too, so this may be worth gating based on vendor name rather than the rendering method used.

PR #85186, opened 2023-11-21
Fix looped animation seek end.

### LINKED ISSUE #37343: Looped animation uses value from the beginning when it shouldn't
**Godot version:**
3.2.1

**Steps to reproduce:**
1. Create an animation, e.g. for scale
2. Set length to 0.9 and snapping to 0.1
3. Add a key at the beginning and end
4. Enable looping
5. When you move cursor to the end of animation, it gets the initial value
mCaKp9Loxu

It might be because of some rounding error, but it happens even if Loop Wrap Mode is set to Clamp. That's obviously a bug.

[ReproductionProject.zip](https://github.com/godotengine/godot/files/4390567/ReproductionProject.zip)

rakkarage:
Closes: https://github.com/godotengine/godot/issues/37343

Thanks.

### DIFF
--- a/scene/animation/animation_player.cpp
+++ b/scene/animation/animation_player.cpp
@@ -183,7 +183,9 @@ void AnimationPlayer::_process_playback_data(PlaybackData &cd, double p_delta, f
 			if (next_pos > len && cd.pos <= len) {
 				looped_flag = Animation::LOOPED_FLAG_END;
 			}
-			next_pos = Math::fposmod(next_pos, (double)len);
+			if (!p_seeked) {
+				next_pos = Math::fposmod(next_pos, (double)len);
+			}
 		} break;
 
 		case Animation::LOOP_PINGPONG: {


TokageItLab (changes requested):
This fix is not good way since this should apply to only editor. Also this causes a regression where the loop does not occur when the script seeks beyond the animation time.

The correct way should be to fix the time when the timeline seek to the end in the editor, or using TOOL_ENABLED with some of the checking for editor seeking.

rakkarage:
This is not a good response.
It works from code and editor.

If you seek to a specific time from code or Editor u wanna skip modulus I think. Otherwise seekToEnd = seekToStart. You could never seek to end from code?
The animation also plays fine after seeking past end from code. But i suspect it not really go past, like editor...

Editor does not allow seek past end, anyway.
What would u like to happen when seek past end in your code? Cannot replicate regression.

YuriSizov:
Superseded by the linked PR. Thanks for your contribution nonetheless! If there is still a runtime issue worth addressing, I guess it would be good to document it with a new issue report and discuss the solution further.

Issue #33969, opened 2019-11-28
Single frame spikes even in an empty project


ghost:
**Godot version:**
3.1.1, 3.2b2

**OS/device including version:**
Windows 10 Pro 1909 x64

**Issue description:**
Every ~10 seconds i get a one-frame spike followed by 1-3 really short frames. At first i thought it was my fault, but problem persists even in an empty project.

Exported project exhibits same behavior.

My specs are i5 7300, 16gb DDR4, GTX 1050

Profiler screenshots:

Frame before the spike:
![Annotation 2019-11-28 204133](https://user-images.githubusercontent.com/58285267/69817786-915b6f80-1225-11ea-8773-c617d3af907f.jpg)

Spike:
![Annotation 2019-11-28 204205](https://user-images.githubusercontent.com/58285267/69817817-9f10f500-1225-11ea-8fab-70a32af61f01.jpg)

Frame after the spike:
![Annotation 2019-11-28 204229](https://user-images.githubusercontent.com/58285267/69817832-a801c680-1225-11ea-900f-8d5b5f048448.jpg)

**Steps to reproduce:**
1) Attach a script with empty "_process" method anywhere in the scene, so that Editor may pause the scene (only for a profiler screenshot)
2) Run project and wait for ~10 sec

**EDIT:**
So, in trying to narrow down the issue(at least as far as my abilities go), i tested this on some other machines

i7 - 9750H / 16 Gb DDR4 / RTX-2060  w/ Windows 10 - same thing,
i3 - 3240 / 4 Gb RAM (didn't check at the time, i believe DDR3) / GeForce 8600GT w/ Windows 8.1 - same thing

I tried to check the first machine (i5 7300, 16gb DDR4, GTX 1050) with Manjaro 18.1.5 and didn't notice any of similar problems (although i couldn't get Vsync to work). Project, exported on Manjaro to Windows exhibits same behavior

3.2 rc4 didn't help.

**EDIT 2: The Editing**
I decided to try Vulkan branch just to see if something different happens.
Spikes aren't there, but i noticed really strange things happen with everything that is normalized for time (with delta), i haven't noticed this before because of short lag, but in Vulkan that short lag isn't there, and after some investigating, i found that sometimes delta is negative(with same frequency as spikes in stable version). Going back to 3.2, negative delta does happen at the same time as spikes, but not everytime.

[...]

PR #81251, opened 2023-09-02
[Windows] Fix not applying NVIDIA profile to new executables

aitorciki:
An NVIDIA profile is applied to the current executable to disable threaded OpenGL optimizations on Windows (see #71472). But because the application is only added to the profile upon the profile creation, newer executables won't be added to the profile (e.g. if the profile is created on first launch of `Godot_v4.1-stable_win64.exe`, when users update the editor and launch `Godot_v4.2-stable_win64.exe`, the profile will never be applied to this new executable).

This patch fixes that scenario by splitting creating the profile (if it doesn't exist) and adding the application (if it doesn't have a profile applied) into two separate steps.

Applications that have been manually added to a different profile aren't overriden to avoid confusing users who know what they're doing.

Issue #85111, opened 2023-11-19
NVIDIA Instant Replay status indicator is everywhere now


nanodeath:
### Godot version

v4.2.rc1

### System information

Godot v4.2.rc1.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1050 Ti with Max-Q Design (NVIDIA; 31.0.15.3640) - Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz (12 Threads)

### Issue description

In Godot 4.2, the Instant Replay 🔃 icon shows up in the corner of...well, basically every conceivable window in Godot, including tooltips. This wasn't present in Godot 4.1.3.

Project Selector
![image](https://github.com/godotengine/godot/assets/10625/6fa391b2-2918-403f-97a0-7b8770386493)

Main Editor
![image](https://github.com/godotengine/godot/assets/10625/d78e5fa0-5248-4b63-8390-8149c32d751c)

Build Project tooltip
![image](https://github.com/godotengine/godot/assets/10625/901e7507-c753-418b-82e9-36f1f28faa5f)

Project menu
![image](https://github.com/godotengine/godot/assets/10625/484ccdb3-f800-4550-8952-70ab04868c32)

Filesystem Context menu
![image](https://github.com/godotengine/godot/assets/10625/64a699fa-4943-4b5c-b4dc-61a97613e8dd)

It usually doesn't impact functionality, except...when it covers up the text, which it does sometimes.

### Steps to reproduce

1. Be on Windows, have an NVIDIA graphics card, have Instant Replay enabled (which I think is the default?) _and_ don't have the status indicator HUD disabled.
2. Use Godot Editor.

---

# Workarounds:

## Turn Off Instant Replay

Just hit Alt-Z and turn off Instant Replay; however, this is not per application, and some people use it.
![image](https://github.com/godotengine/godot/assets/10625/eb24c879-16e5-464d-874f-e4a064fa198b)

## Turn Off Indicator

Also under the Alt-Z menu, you can go to settings -> HUD layout -> status indicator.

![image](https://github.com/godotengine/godot/assets/10625/4604b517-908c-42a1-afbe-fc8b1c431754)

### Minimal reproduction project

N/A

sonic2kk:
> This wasn't present in Godot 4.1.3.

[...]

PR #71472, opened 2023-01-15
Disable NVIDIA's threaded OpenGL optimization on Windows

EIREXE:
Disables NVIDIA's threaded optimization which causes random stuttering on Windows with NVIDIA GPUs, virtually all performant OpenGL programs (like ryujinx) disable this to prevent stuttering and it's a well documented issue.

Should fix #33969

- *Production edit: This closes https://github.com/godotengine/godot/issues/45660.*

PR #85188, opened 2023-11-21
[Windows] Fix NVIDIA app profile creation

aitorciki:
_Note: hopefully to be considered for inclusion in 4.2, as the existing bug potentially results in an ever-expanding NVIDIA app profile with every execution of the engine._

When adding an executable to the NVIDIA profile created to disable threaded optimization, wrong values were passed to the `launcher` and `fileInFolder` params: `memcpy` was reading a single byte off a multibyte string for each, potentially resulting in gargabe being read into these values. Non-empty values in any of these fields resulted in a corrupted profile, which wouldn't be applied to the executable at launch (and wasn't even editable from NVIDIA Control Panel, only from [NVIDIA Profile Inspector](https://github.com/Orbmu2k/nvidiaProfileInspector)).

On top of that, since #81251, the invalid profile resulted in NVIDIA API's `findApplicationByName` to always fail (as it relies on the driver's capacity to apply a profile to the target executable), and in turn the executable being re-added to the profile on every engine launch. In summary, an ever-expanding list of apps added to the profile, that wasn't ever properly applied.

This PR fixes that condition by taking three steps:
* Fixing the `launcher` and `fileInFolder` settings to receive a "full" empty mb string (I'm a total C++ noob, so would love some review to my approach to fixing `memcpy`).
* Deleting the old, potentially corrupt profile if it exists (and taking advantage of the situation to rename it to something less verbose, more in line with the rest of applications, dropping the `Nvidia Profile` from its name).
* Using the safer `GetApplicationInfo` method to check if an executable already exists inside a profile, as it doesn't rely on the driver being able to apply the profile, just on the profile metadata itself.

Some other params of the profile app are dropped, as they seem superfluous in my tests:
* `userFriendlyName` is better left to default to the executable name, as this makes its distinguishable from other executables covered by the same profile in NVIDIA Control Panel (and is what most applications default to).
* `isMetro` and `isCommandLine` seem to be ignored by the creation API no matter what value is passed (they're always read as 0 in my tests), probably unused values in the current version kept for backwards compatibility?

Since this bug results in garbage being quietly accumulated in NVIDIA profiles as users launch games or the editor, I think it'd be really interesting to consider the fix for inclusion in 4.2, to avoid that leaky behaviour to spread when the stable release sees wide adoption in a few weeks.

### DIFF
--- a/platform/windows/gl_manager_windows_native.cpp
+++ b/platform/windows/gl_manager_windows_native.cpp
@@ -88,7 +88,8 @@ typedef int(__cdecl *NvAPI_DRS_CreateApplication_t)(NvDRSSessionHandle, NvDRSPro
 typedef int(__cdecl *NvAPI_DRS_SaveSettings_t)(NvDRSSessionHandle);
 typedef int(__cdecl *NvAPI_DRS_SetSetting_t)(NvDRSSessionHandle, NvDRSProfileHandle, NVDRS_SETTING *);
 typedef int(__cdecl *NvAPI_DRS_FindProfileByName_t)(NvDRSSessionHandle, NvAPI_UnicodeString, NvDRSProfileHandle *);
-typedef int(__cdecl *NvAPI_DRS_FindApplicationByName_t)(NvDRSSessionHandle, NvAPI_UnicodeString, NvDRSProfileHandle *, NVDRS_APPLICATION *);
+typedef int(__cdecl *NvAPI_DRS_GetApplicationInfo_t)(NvDRSSessionHandle, NvDRSProfileHandle, NvAPI_UnicodeString, NVDRS_APPLICATION *);
+typedef int(__cdecl *NvAPI_DRS_DeleteProfile_t)(NvDRSSessionHandle, NvDRSProfileHandle);
 NvAPI_GetErrorMessage_t NvAPI_GetErrorMessage__;
 
 static bool nvapi_err_check(const char *msg, int status) {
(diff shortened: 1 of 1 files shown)


mhilbrunner:
Hey, thanks for contributing and sorry this has been sitting for a while! One reason may be that this is a very specific part of a already specific subset of the engine and it's user base (only affecting Windows, only on NVIDIA, and needs intersecting knowledge of rendering, engine core, Windows/NVIDIA, etc., or at least willingness to dive in).

What would've helped me review this is more context, as everything is scattered through past PRs and issues and I had to dig a bit.

### Context

First, some context: originally introduced in PR https://github.com/godotengine/godot/pull/71472 to fix https://github.com/godotengine/godot/issues/33969 and https://github.com/godotengine/godot/issues/45660, later changed by PR https://github.com/godotengine/godot/pull/81251 (which may have broken things), semi-related issue: https://github.com/godotengine/godot/issues/85111. This work has since been cherry-picked to 4.1, and is still slated to maybe be cherrypicked to 3.x, so we should do the same with this PR.

We disable NVIDIA's OpenGL threading optimization for the OpenGL backend by creating a custom NVIDIA Profile per Godot project name, which is applied to executables by executable name (i.e., matching something like `godot.windows.editor.x86_64.exe`).

This code is called via `DisplayServerWindows::DisplayServerWindows()` -> `rendering_driver == "opengl3"?` -> `GLManagerNative_Windows::initialize()` -> `GLManagerNative_Windows::_nvapi_disable_threaded_optimization()`.

This means this is run whenever the project manager opens, because it is always running in OpenGL/Compat. Whether we should always create a NVIDIA Profile, and only for specific backends, and without much dev input is IMO worthy of discussion, but outside of the scope of this PR, as we already do that. IMO it should be a project setting.

### This PR

This PR makes the following changes:

> Fixing the launcher and fileInFolder settings to receive a "full" empty mb string (I'm a total C++ noob, so would love some review to my approach to fixing memcpy).

✅ The changes look correct to me, and seem to work fine after building and testing.

> Deleting the old, potentially corrupt profile if it exists

✅ This makes sense, needs to be done and works correctly. I could verify that it successfully deletes the old profile if it exists and creates a new one, and also runs without issue if no pre-existing profile, well, exists.

> (and taking advantage of the situation to rename it to something less verbose, more in line with the rest of applications, dropping the Nvidia Profile from its name).

✅ This makes sense, as the "NVIDIDA Profile" part of the name stood out compared to all others in my list. Wouldn't bother making this change for the sake of it, but as we are already touching this, may as well.

> Using the safer GetApplicationInfo method to check if an executable already exists inside a profile, as it doesn't rely on the driver being able to apply the profile, just on the profile metadata itself.

✅ Makes sense. Works as it should.

> Using defaults/leaving off userFriendlyName, isMetro and isCommandLine 

✅ Makes sense and didn't have any negative effects in my testing. Indeed, `isMetro` and `isCommandLine` seem to be ignored anyway and never showed up in the first place before.

mhilbrunner (approved):

EIREXE:
lgtm

akien-mga:
Thanks!

PR #85192, opened 2023-11-21
SCons: Update mypy

Repiteo:
Implements a handful of minor adjustments all revolving around mypy
- Added a new section to the `.gitignore` to explicitly handle the mypy cache after running the mypy test script (or mypy in general)
- Added a new value to the mypy config: `no_implicit_optional = True`. This is because mypy made this their default value in later versions[^1], so adding a rule one way or the other clears up ambiguity for mypy installed from other sources (eg: vscode extension)
- Fixes two files that weren't conforming to `no_implicit_optional = True`. While they could've been left alone if the value was changed to `false`, all cases of changed arguments were right beside another argument that was *already* using an explicit optional wrapper; as such, the changes were sensible in-context & just so happened to be the only offending areas in the entire repo to begin with

[^1]: https://github.com/hauntsaninja/no_implicit_optional#whats-going-on

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,6 +153,9 @@ gmon.out
 # Kdevelop
 *.kdev4
 
+# Mypy
+.mypy_cache
+
 # Qt Creator
 *.config
 *.creator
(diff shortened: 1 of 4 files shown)


akien-mga (approved):

akien-mga (approved):

akien-mga:
Thanks!

PR #85197, opened 2023-11-21
Run `Clean` in Windows SCsub to clean up debug files

### LINKED ISSUE #75570: Build system: cleaning Visual Studio solution does not erase .pdb, .ilk, .lib files
### Godot version

master

### System information

Windows 11, VS2022

### Issue description

I use the following command to generate a Visual Studio 2022 solution.

`scons vsproj=true dev_mode=true dev_build=true target=editor tests=true`

All my compilation is done through VS2022. I semi-regularly clean out the solution and rebuild from scratch:

image

This process removes all the `.obj` files, but unfortunately does not remove the `.pdb.`, `.ilk`, and `.lib` files. Particularly, this causes `.pdb` files to bloat. I've had them as large as 2.5gb.

image

By comparison, this is the same folder after manually deleting the files above and recompiling from scratch:

image

Arguably the files generated by having debug symbols should be cleaned out as well when cleaning/rebuilding solution. I'm sure this is platform and build-toolchain dependent, but each toolchain/platform combo can be addressed in separate PRs.

### Steps to reproduce

See above.

N/A

brno32:
Addresses only the Windows platform in https://github.com/godotengine/godot/issues/75570

The original issue mentions it's a problem with the "Clean Solution" button in Visual Studio, but the problem is with the call to `scons --clean` itself, which Visual Studio is using under the hood.

`scons --clean` only drops the `.exe` files from the `bin` folder, but it leaves everything else that's in there.

I couldn't find a way to get SCons to tell me the list of files in the output directory it made from the build, so deriving it from the program name seemed like the next best thing. I don't think we want to just clear the `bin` folder as that could delete files from targets other than the one being cleaned.

*Bugsquad edit:*
- Fixes #75570

### DIFF
--- a/platform/windows/SCsub
+++ b/platform/windows/SCsub
@@ -3,6 +3,7 @@
 Import("env")
 
 import os
+from pathlib import Path
 from platform_methods import run_in_subprocess
 import platform_windows_builders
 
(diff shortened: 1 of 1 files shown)


YuriSizov:
> I don't think we want to just clear the bin folder as that could delete files from targets other than the one being cleaned.

This is correct, but I think the implementation with a wildcard still has this problem, because different build configurations would only differ by suffixes. So the wildcard mask can overlap.

YuriSizov:
> @YuriSizov would you mind taking another look now that I've refactored the approach?

The approach looks fine to me now. But I'm not fluent in SCons to review the implementation for correctness. So we'll need to wait a bit for someone else to take a look.

akien-mga (approved):
I don't use Windows frequently so I haven't tested, but the rationale makes sense and the implementation seems ok.

akien-mga:
Thanks!

PR #84488, opened 2023-11-05
Restore losed focus for some pop-up dialogs in visual shader editor

Chaosus:
Fix dialog behavior which may loses the required focus (for example, when the user right-click on graph if members dialog is opened).

Before:
focus

After:
focus2

PR #85205, opened 2023-11-22
Set some dialogs in the VisualShader editor to be exclusive

### LINKED ISSUE #83805: Some times can't add node in visualshader editor
### Godot version

v4.2.beta2.official [f8818f85e]

### System information

Windows 10, 64bit

### Issue description

Some times can't add node in visualshader editor

### Steps to reproduce

1. Open a VisualShader file for edit.
2. Click the `Add Node...` button, it will popup menu, it is OK.
3. Direct close the menu.
4. Clikd the `Add Node...` again, the menu can't popup.
GIF

/

jsjtxietian:
Should fix #83805

### DIFF
--- a/editor/plugins/visual_shader_editor_plugin.cpp
+++ b/editor/plugins/visual_shader_editor_plugin.cpp
@@ -4080,6 +4080,12 @@ void VisualShaderEditor::_show_members_dialog(bool at_mouse_pos, VisualShaderNod
 		saved_node_pos_dirty = false;
 		members_dialog->set_position(graph->get_screen_position() + Point2(5 * EDSCALE, 65 * EDSCALE));
 	}
+
+	if (members_dialog->is_visible()) {
+		members_dialog->grab_focus();
+		return;
+	}
+
 	members_dialog->popup();
 
 	// Keep dialog within window bounds.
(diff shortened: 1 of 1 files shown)


akien-mga:
That should work, but I'm not sure it's the best solution.

I would expect either of those behaviors:
- Either the popup should be exclusive so it can't lose focus until you close it,
- Or clicking outside the popup should close it, not just lose focus.

jsjtxietian:
I think either is ok really, I sightly perfer the first solution where it grabs focus (more intuitive) , the second one may cause some confusion.

jsjtxietian:
I have updated the code, along with some other popup window with the same problem in visual shader editor. Tested on my windows it works fine.

Chaosus:
Seems to be a good replacement for https://github.com/godotengine/godot/pull/84488.

editor/plugins/visual_shader_editor_plugin.cpp
```diff
@@ -4080,6 +4080,12 @@ void VisualShaderEditor::_show_members_dialog(bool at_mouse_pos, VisualShaderNod
 		saved_node_pos_dirty = false;
 		members_dialog->set_position(graph->get_screen_position() + Point2(5 * EDSCALE, 65 * EDSCALE));
 	}
+
+	if (members_dialog->is_visible()) {
+		members_dialog->grab_focus();
+		return;
+	}
+
```
akien-mga, line 4088: Since this dialog is now exclusive, I guess this situation should no longer happen?

May not hurt to keep it for safety.

akien-mga (approved):
Looks good to me overall.

editor/plugins/visual_shader_editor_plugin.cpp
jsjtxietian: Yes this should be removed, if there is any related bug in the future I will fix it.

PR #81968, opened 2023-09-20
[macOS] Workaround Xcode 15 linker bug.

bruvzg:
Workaround for https://github.com/godotengine/godot/issues/81948, this seems to be a linker bug in Xcode 15, and I'm not sure what exactly is causing it.

* *Bugsquad edit, closes: #81948*

Issue #81948, opened 2023-09-19
Compiling ARM64 error after installing Xcode15


MusouCrow:
### Godot version

4.2.dev [e3e2528ba]

### System information

Godot v4.2.dev (e3e2528ba) - macOS 14.0.0 - Vulkan (Forward+) - integrated Apple M1 - Apple M1 (8 Threads)

### Issue description

When I wanted to compile Godot for Apple Silicon today, I have encountered such an error:

```shell
scons: Reading SConscript files ...
Auto-detected 8 CPU cores available for build parallelism. Using 7 cores by default. You can override it with the -j argument.
Building for macOS 11.0+.
MoltenVK found at: ../MoltenVK/MoltenVK/MoltenVK.xcframework/macos-arm64_x86_64/
Building for platform "macos", architecture "arm64", target "editor".
Checking for C header file mntent.h... (cached) no
scons: done reading SConscript files.
scons: Building targets ...
[ 96%] Linking Program bin/godot.macos.editor.arm64 ...
[100%] progress_finish(["progress_finish"], [])
[100%] 0  0x10069f648  __assert_rtn + 72
1  0x1005c7c5c  ld::Fixup::applyFixup(ld::Atom const*, ld::LayoutLinkedImage const&, unsigned char*) const + 8268
2  0x10065a7d8  ___ZN2ld16LayoutExecutable27writeContentWithoutLinkEditENSt3__14spanIhLm18446744073709551615EEEy_block_invoke + 332
3  0x18ed51950  _dispatch_client_callout2 + 20
4  0x18ed64ba0  _dispatch_apply_invoke + 176
5  0x18ed51910  _dispatch_client_callout + 20
6  0x18ed633cc  _dispatch_root_queue_drain + 864
7  0x18ed63a04  _dispatch_worker_thread2 + 156
8  0x18eefb0d8  _pthread_wqthread + 228
ld: Assertion failed: (extras.otherInstrOffset != 0 && "Kind::arm64_adrp_ldr missing extra info"), function applyFixup, file Fixup.cpp, line 793.
clang: error: linker command failed with exit code 1 (use -v to see invocation)
scons: *** [bin/godot.macos.editor.arm64] Error 1
scons: building terminated because of errors.
```

### Steps to reproduce

It's the command I used: 

```shell
scons platform=macos arch=arm64 target=editor vulkan_sdk_path='../MoltenVK'
```

I'm sure there was no problem until I updated XCode15, but `x86_64` is done:

```shell
[...]

PR #85206, opened 2023-11-22
[macOS] Check Apple specific version instead of generic clang version.

bruvzg:
Follow up to #81968 (see https://github.com/godotengine/godot/issues/81948#issuecomment-1822205076). Should ensure that legacy linker is only used on Xcode 15.0.0 and 15.0.1, but not 15.1.x.


akien-mga (approved):

akien-mga:
Thanks!

PR #26043, opened 2019-02-19
Add some documentation for methods that return Copy-on-Write arrays.

JosephCatrambone:
"Fixes" #18573.  Realistically, adds some warnings about potential foot-guns.  Some classes have a .polygon attribute which actually returns a new array, not a reference to the old one.  This may lead some inexperienced developers to attempt things that appear reasonable, like `obj.polygon.resize(10)`, but do not actually change the underlying polygon.

PR #85207, opened 2023-11-22
Add further details on properties returning `Packed*Array`

### LINKED ISSUE #85198: Direct manipulation of Polygon2D polygon array elements possible in GDScript but not in CSharp.
### Godot version

v4.1.3.stable.mono.official [f06b6836a]

### System information

Godot v4.1.3.stable.mono - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3060 (NVIDIA; 31.0.15.4601) - 11th Gen Intel(R) Core(TM) i5-11400F @ 2.60GHz (12 Threads)

### Issue description

When creating a Polygon2D Node (with 4 points in the polygon array) and manipulating the points via code it seems that this is only possible in GDScript directly. In C# it seems only to be possible by replacing the whole Vector2 array (named Polygon).

### Steps to reproduce

See minimal reproduction project for a full example with GDScript & C#
- create a Polygon2D node
- Set some points for the polygon 
- attach a script to the node which manipulates one or more points in code
- if manipulating points directly in the polygon array it will only work in GDScript but not via C# 

See attached a Minimal reproduction project.

It contains two Polygon2D Nodes. The left one has a GDScript attached. 
The right one has a C# Script attached. The code should be "equal".

Both have two exposed variables: directUpdate and debugPointsToConsole which can be used to control the program flow.

The top left point (point 0 ) of both polygons will move up and down.

By switching directUpdate On/Off (True/False) the script will use the UpdateDirect() or UpdateIndirect() functions to update one points in the polygon. UpdateDirect manipulates the array directly, In UpdateIndirect the change is made to a copy of the array which is then copied back.

As you will see in C# the direct manipulation of the Polygon array seems not to work. 

I am aware of the note on the https://docs.godotengine.org/en/stable/classes/class_polygon2d.html page:
`Note: This returns a copy of the [PackedVector2Array](https://docs.godotengine.org/en/stable/classes/class_packedvector2array.html#class-packedvector2array) rather than a reference.`

If my way to trying to manipulate a point directly in C# is wrong, is there any way to do it there besides replacing the whole array? 

Thanks for any responses.
[Polygon2DTest.zip](https://github.com/godotengine/godot/files/13431809/Polygon2DTest.zip)

Edit: replaced archive. Removed unnecessary files.

AThousandShips:
Limited it to properties with setters and existing documentation, might be excessive but without some central point to point to explaining this I feel this is a relevant point to clarify generally (also easier to access the information from the built-in documentation)

Left the documentation for `Gradient` as is as it has more specific details, could possibly add specific details to some of these if they have specific setters for details, but left it simple for now

* Closes: https://github.com/godotengine/godot/issues/85198

### DIFF
--- a/doc/classes/CollisionPolygon3D.xml
+++ b/doc/classes/CollisionPolygon3D.xml
@@ -21,7 +21,6 @@
 		</member>
 		<member name="polygon" type="PackedVector2Array" setter="set_polygon" getter="get_polygon" default="PackedVector2Array()">
 			Array of vertices which define the 2D polygon in the local XY plane.
-			[b]Note:[/b] The returned value is a copy of the original. Methods which mutate the size or properties of the return value will not impact the original polygon. To change properties of the polygon, assign it to a temporary variable and make changes before reassigning the class property.
 		</member>
 	</members>
 </class>
--- a/doc/classes/Gradient.xml
+++ b/doc/classes/Gradient.xml
@@ -78,8 +78,8 @@
 	</methods>
 	<members>
 		<member name="colors" type="PackedColorArray" setter="set_colors" getter="get_colors" default="PackedColorArray(0, 0, 0, 1, 1, 1, 1, 1)">
-			Gradient's colors returned as a [PackedColorArray].
-			[b]Note:[/b] This property returns a copy, modifying the return value does not update the gradient. To update the gradient use [method set_color] method (for updating colors individually) or assign to this property directly (for bulk-updating all colors at once).
+			Gradient's colors as a [PackedColorArray].
+			[b]Note:[/b] Setting this property updates all colors at once. To update any color individually use [method set_color].
 		</member>
 		<member name="interpolation_color_space" type="int" setter="set_interpolation_color_space" getter="get_interpolation_color_space" enum="Gradient.ColorSpace" default="0">
 			The color space used to interpolate between points of the gradient. It does not affect the returned colors, which will always be in sRGB space. See [enum ColorSpace] for available modes.
@@ -89,8 +89,8 @@
 			The algorithm used to interpolate between points of the gradient. See [enum InterpolationMode] for available modes.
 		</member>
 		<member name="offsets" type="PackedFloat32Array" setter="set_offsets" getter="get_offsets" default="PackedFloat32Array(0, 1)">
-			Gradient's offsets returned as a [PackedFloat32Array].
-			[b]Note:[/b] This property returns a copy, modifying the return value does not update the gradient. To update the gradient use [method set_offset] method (for updating offsets individually) or assign to this property directly (for bulk-updating all offsets at once).
+			Gradient's offsets as a [PackedFloat32Array].
+			[b]Note:[/b] Setting this property updates all offsets at once. To update any offset individually use [method set_offset].
 		</member>
 	</members>
 	<constants>
--- a/doc/classes/PackedByteArray.xml
+++ b/doc/classes/PackedByteArray.xml
@@ -6,6 +6,7 @@
 	<description>
 		An array specifically designed to hold bytes. Packs data tightly, so it saves memory for large array sizes.
 		[PackedByteArray] also provides methods to encode/decode various types to/from bytes. The way values are encoded is an implementation detail and shouldn't be relied upon when interacting with external apps.
+		[b]Note:[/b] Packed arrays are always passed by reference. To get a copy of an array that can be modified independently of the original array, use [method duplicate]. This is [i]not[/i] the case for built-in properties and methods. The returned packed array of these are a copies, and changing it will [i]not[/i] affect the original value. To update a built-in property you need to modify the returned array, and then assign it to the property again.
 	</description>
 	<tutorials>
 	</tutorials>
--- a/editor/editor_help.cpp
+++ b/editor/editor_help.cpp
@@ -96,6 +96,18 @@ const Vector<String> classes_with_csharp_differences = {
 };
 #endif
 
+const Vector<String> packed_array_types = {
+	"PackedByteArray",
+	"PackedColorArray",
+	"PackedFloat32Array",
+	"PackedFloat64Array",
+	"PackedInt32Array",
+	"PackedInt64Array",
+	"PackedStringArray",
+	"PackedVector2Array",
+	"PackedVector3Array",
+};
+
 // TODO: this is sometimes used directly as doc->something, other times as EditorHelp::get_doc_data(), which is thread-safe.
 // Might this be a problem?
 DocTools *EditorHelp::doc = nullptr;
@@ -2193,6 +2205,12 @@ void EditorHelp::_update_doc() {
 				}
 				has_prev_text = true;
 				_add_text(descr);
+				// Add copy note to built-in properties returning Packed*Array.
+				if (!cd.is_script_doc && packed_array_types.has(prop.type)) {
+					class_desc->add_newline();
+					class_desc->add_newline();
+					_add_text(vformat(TTR("[b]Note:[/b] The returned array is [i]copied[/i] and any changes to it will not update the original property value. See [%s] for more details."), prop.type));
+				}
 			} else if (!has_prev_text) {
 				class_desc->add_image(get_editor_theme_icon(SNAME("Error")));
 				class_desc->add_text(" ");
--- a/doc/classes/CollisionPolygon2D.xml
+++ b/doc/classes/CollisionPolygon2D.xml
@@ -26,7 +26,6 @@
 		<member name="polygon" type="PackedVector2Array" setter="set_polygon" getter="get_polygon" default="PackedVector2Array()">
 			The polygon's list of vertices. Each point will be connected to the next, and the final point will be connected to the first.
 			[b]Note:[/b] The returned vertices are in the local coordinate space of the given [CollisionPolygon2D].
-			[b]Warning:[/b] The returned value is a clone of the [PackedVector2Array], not a reference.
 		</member>
 	</members>
 	<constants>
--- a/doc/classes/OccluderPolygon2D.xml
+++ b/doc/classes/OccluderPolygon2D.xml
@@ -17,7 +17,6 @@
 		</member>
 		<member name="polygon" type="PackedVector2Array" setter="set_polygon" getter="get_polygon" default="PackedVector2Array()">
 			A [Vector2] array with the index for polygon's vertices positions.
-			[b]Note:[/b] The returned value is a copy of the underlying array, rather than a reference.
 		</member>
 	</members>
 	<constants>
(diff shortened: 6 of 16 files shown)


doc/classes/CollisionPolygon3D.xml
```diff
@@ -21,7 +21,7 @@
 		</member>
 		<member name="polygon" type="PackedVector2Array" setter="set_polygon" getter="get_polygon" default="PackedVector2Array()">
 			Array of vertices which define the 2D polygon in the local XY plane.
-			[b]Note:[/b] The returned value is a copy of the original. Methods which mutate the size or properties of the return value will not impact the original polygon. To change properties of the polygon, assign it to a temporary variable and make changes before reassigning the class property.
+			[b]Note:[/b] The returned value is a copy of the original. Methods which mutate the size or properties of the return value will not impact the original array. To change properties of the array, assign it to a temporary variable and make changes before reassigning the class property.
```
AThousandShips, line 24: Updated the message to not be specific, I believe this makes each copy directly translatable

Mickeon:
I don't know how to feel about this. The note of this PR talks about "returned values" and "methods", but the subject is always a property. The meaning of the note can only be inferred if you know what setters and getters are, and is otherwise still a bit puzzling.

Also consider that both GDScript and C# discourage using setters/getters for built-in classes whenever possible, in favour of properties.

AThousandShips:
This is an existing note applied to more places, I didn't write this just updated one and copied it around, this was done in:
* https://github.com/godotengine/godot/pull/26043

Mickeon:
I'm aware it does exist elsewhere, I am just a fan of the way it is worded when applied to an Object's properties.

doc/classes/ArrayOccluder3D.xml
```diff
@@ -24,10 +24,12 @@
 		<member name="indices" type="PackedInt32Array" setter="set_indices" getter="get_indices" default="PackedInt32Array()">
 			The occluder's index position. Indices determine which points from the [member vertices] array should be drawn, and in which order.
 			[b]Note:[/b] The occluder is always updated after setting this value. If creating occluders procedurally, consider using [method set_arrays] instead to avoid updating the occluder twice when it's created.
+			[b]Note:[/b] The returned value is a copy of the original. Methods which mutate the size or properties of the return value will not impact the original array. To change properties of the array, assign it to a temporary variable and make changes before reassigning the class property.
```
Mickeon, line 27: One more thing that contributes to the issue. "_Properties_ of the array" in this context can potentially be ambiguous. What it is referring to is the PackedInt32Array of `indices` (specifically its elements), but `indices` **is** a _property_ itself.

AThousandShips:
Can work on the verbosity, or add a note to the packed arrays themselves and link to it instead, but this thing has caused a lot of confusion and I think just putting it in the manual isn't enough (and it isn't there yet either)

AThousandShips:
Could have an automatic addition, can look into that, but I think important to have self contained in the built in docs

AThousandShips:
Thinking an addition to all the `Packed*Array` classes on the lines of:
Note: When a Packed*Array is returned from a property or method from built-in methods it is *not* shared like an `Array` is, and modifying the returned array will *not* modify the property, instead make your modifications to the returned array and then assign it to the property again.

Or something, and then maybe an automatic addition on properties returning a packed array type, that might be intrusive but will see, for translation it should be in the file id say

AThousandShips:
Added notes to `Packed*Array` and will look at adding automatic details to properties next

AThousandShips:
Added an automatic note in the built-in documentation, will see how to implement the same for the rst generation etc., if this is the way to go

Any suggestions on either wording is welcome

doc/classes/Gradient.xml
```diff
@@ -78,8 +78,7 @@
 	</methods>
 	<members>
 		<member name="colors" type="PackedColorArray" setter="set_colors" getter="get_colors" default="PackedColorArray(0, 0, 0, 1, 1, 1, 1, 1)">
-			Gradient's colors returned as a [PackedColorArray].
-			[b]Note:[/b] This property returns a copy, modifying the return value does not update the gradient. To update the gradient use [method set_color] method (for updating colors individually) or assign to this property directly (for bulk-updating all colors at once).
+			Gradient's colors returned as a [PackedColorArray]. To update the gradient use [method set_color] method (for updating colors individually) or assign to this property directly (for bulk-updating all colors at once).
```
AThousandShips, line 81: Unsure if I should just remove these, but keeping them for now as they provide specific information

editor/editor_help.cpp
```diff
@@ -1831,6 +1831,15 @@ void EditorHelp::_update_doc() {
 			class_desc->push_indent(1);
 			if (!cd.properties[i].description.strip_edges().is_empty()) {
 				_add_text(DTR(cd.properties[i].description));
+				// Add copy note to built-in properties returning Packed*Array.
+				if (!cd.is_script_doc &&
```
AThousandShips, line 1835: Restricting to built-in classes at the moment, as from what I can tell at least for GDScript the data isn't copied like this but remains shared, but will investigate further as well

AThousandShips:
Not extremely familiar with the doc generation system but attempted to add the same feature to the rst side, any hints would be very welcome

doc/classes/PackedByteArray.xml
```diff
@@ -6,6 +6,7 @@
 	<description>
 		An array specifically designed to hold bytes. Packs data tightly, so it saves memory for large array sizes.
 		[PackedByteArray] also provides methods to encode/decode various types to/from bytes. The way values are encoded is an implementation detail and shouldn't be relied upon when interacting with external apps.
+		[b]Note:[/b] Arrays are always passed by reference. To get a copy of an array that can be modified independently of the original array, use [method duplicate]. This does [i]not[/i] apply to properties on built-in classes or return values from built-in methods, these are passed by copy, and updating the returned value will [i]not[/i] change the original value. To update a built-in property you will need to modify the returned array, and then assign it to the property again.
```
Mickeon, line 9: Massive rewording suggestion, applies to every other Packed*Array.
```suggestion
		[b]Note:[/b] Packed arrays are always passed by reference. To get a copy of an array that can be modified independently of the original array, use [method duplicate]. This is [i]not[/i] the case for built-in properties and methods. The returned packed array of these is a copy, and changing it will [i]not[/i] affect the original value. To update a built-in property you need to modify the returned array, and then assign it to the property again.
```
I get you tried to be as explicit as possible before, but the proper note in the class reference does a much better job at this, now.

editor/editor_help.cpp
```diff
@@ -1831,6 +1831,15 @@ void EditorHelp::_update_doc() {
...
+						(cd.properties[i].type == "PackedByteArray" || cd.properties[i].type == "PackedColorArray" || cd.properties[i].type == "PackedFloat32Array" ||
+								cd.properties[i].type == "PackedFloat64Array" || cd.properties[i].type == "PackedInt32Array" || cd.properties[i].type == "PackedInt64Array" ||
+								cd.properties[i].type == "PackedStringArray" || cd.properties[i].type == "PackedVector2Array" || cd.properties[i].type == "PackedVector3Array")) {
```
Mickeon, line 1838: This is the part where an opposite of Variant's `get_type_name` would come extremely useful, let's take notes.

doc/classes/PackedByteArray.xml
AThousandShips: Thank you will take a look 🙂

editor/editor_help.cpp
AThousandShips: Indeed, will probably use a vector of map here like already done with types that differ in C#, but did a simple solution for now, and will see how I finally apply it

doc/classes/Gradient.xml
Mickeon, line 81: While you're here, perhaps:
```suggestion
			Gradient's colors as a [PackedColorArray].
			[b]Note:[/b] Setting this property updates all colors at once. To update any color individually use [method set_color].
```

doc/classes/Gradient.xml
```diff
@@ -89,8 +88,7 @@
 			The algorithm used to interpolate between points of the gradient. See [enum InterpolationMode] for available modes.
 		</member>
 		<member name="offsets" type="PackedFloat32Array" setter="set_offsets" getter="get_offsets" default="PackedFloat32Array(0, 1)">
-			Gradient's offsets returned as a [PackedFloat32Array].
-			[b]Note:[/b] This property returns a copy, modifying the return value does not update the gradient. To update the gradient use [method set_offset] method (for updating offsets individually) or assign to this property directly (for bulk-updating all offsets at once).
+			Gradient's offsets returned as a [PackedFloat32Array]. To update the gradient use [method set_offset] method (for updating offsets individually) or assign to this property directly (for bulk-updating all offsets at once).
```
Mickeon, line 91:
```suggestion
			Gradient's offsets as a [PackedFloat32Array].
			[b]Note:[/b] Setting this property updates all offsets at once. To update any offset individually use [method set_offset].
```

doc/classes/PackedByteArray.xml
Mickeon: Not sure if this is desired or not but I'm going to comment on it as a reminder

editor/editor_help.cpp
Mickeon: There seems to be some interest in https://github.com/godotengine/godot/pull/85474 so that may be worth keeping in mind here.

doc/classes/PackedByteArray.xml
AThousandShips: Will see about adding this when I get back to looking at the specifics, just resolved some conflicts yesterday

Mickeon (approved):
Looks great and has been needed for a while.

AThousandShips:
Someone well versed in the rst generation and style would be appreciated to take a look at that side, it looked like it works correctly but I haven't worked with that side properly

Calinou (commented):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/29b3d9e9e538f0aa8effc8ad8bf19a2915292a89), it works as expected.

image

editor/editor_help.cpp
```diff
@@ -2193,6 +2205,12 @@ void EditorHelp::_update_doc() {
 				}
 				has_prev_text = true;
 				_add_text(descr);
+				// Add copy note to built-in properties returning Packed*Array.
+				if (!cd.is_script_doc && packed_array_types.has(prop.type)) {
+					class_desc->add_newline();
+					class_desc->add_newline();
+					_add_text("[b]" + TTR("Note:") + "[/b] " + vformat(TTR("The returned array is [i]copied[/i] and any changes to it will not update the original property value. See [%s] for more details."), prop.type));
```
mhilbrunner, line 2212:
```suggestion
					_add_text(vformat(TTR("[b]Note:[/b] The returned array is [i]copied[/i] and any changes to it will not update the original property value. See [%s] for more details."), prop.type));
```
Is there a reason to translate `Note:` separately? IME, concatenating translated strings should be avoided due to different grammars etc.

AThousandShips: Copied from elsewhere, forget where

mhilbrunner:
Commits need to be squashed, besides that and the small note above this LGTM.

AThousandShips:
Good, had separate commits for specific things, will fixup and squash

mhilbrunner (approved):

akien-mga:
Needs rebase.

akien-mga:
Thanks!

AThousandShips:
Thank *you*! Great to get this merged, learned entirely new areas of the system

PR #85208, opened 2023-11-22
Avoid conflict between mingw-std-threads and Clang's own

RandomShaper:
Since mingw-std-threads would inject its symbols into `std` if it thinks there's no other implementation around while using a MinGW compiler (`#if (__cplusplus < 201703L) || (defined(__MINGW32__ ) && !defined(_GLIBCXX_HAS_GTHREADS))`, it will break LLVM-MinGW builds. The reason is that LLVM-MinGW provides implementations but doesn't define `_GLIBCXX_HAS_GTHREADS`.

This PR rewrites the patch to change the check to ~`#if defined(__MINGW32__ ) && !(defined(_GLIBCXX_HAS_GTHREADS) && !defined(__clang__))`~ (`#if defined(__MINGW32__ ) && !defined(_GLIBCXX_HAS_GTHREADS) && !defined(__clang__)`).

Another option would have been just not to use mingw-std-threads at all on LLVM-MinGW, but I see value in using the same implemention in both MinGW-style compilers. Other opinions welcome.

### DIFF
--- a/thirdparty/README.md
+++ b/thirdparty/README.md
@@ -530,7 +530,8 @@ Files extracted from upstream repository:
 - `mingw.shared_mutex.h`
 - `mingw.thread.h`
 
-Once copied, apply `no_except.patch` (needed because Godot is built without exceptions).
+Once copied, apply `godot.patch` (needed because Godot is built without exceptions
+and to avoid std:: replacements leak in Clang builds).
 
 
 ## minimp3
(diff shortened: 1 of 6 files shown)


akien-mga (approved):
Looks good to me.

I agree with using the same mingw-std-threads implementation on all toolchains.

RandomShaper:
I think I've messed up the condition. Will force-push in a minute.

akien-mga:
The new condition makes sense to me, but the patch file seems out of sync.

RandomShaper:
Please let me know how it looks now. I can't trust my eyes today anymore.

akien-mga (approved):
Looks good!

bitsawer:
Tested on llvm-clang on Windows and seems to now build and run after some quick test, fixes the build error I mentioned on RocketChat.

One extra change is needed to fix building with `dev_mode` enabled. Clang warns about an unused variable as the changes in the patch file now make `errnum` variable unused. Clang seems to be a bit stricter compared to other compilers. Line 146 in `godot.patch` can probably removed completely to fix it:

```cpp
int errnum = errno;
```

RandomShaper:
@bitsawer Thanks lots for testing! Suggestion applied.

akien-mga:
Thanks!

Issue #78454, opened 2023-06-20
Godot 4 Issue - Intellisense issue with variables inside Autoload Scripts


SoldierCantKill:
### Godot version

4.0.3.stable.official [5222a99f5]

### System information

Windows 10 - Godot 4.0.3.stable.official

### Issue description

Godot 4's intellsense is very buggy, but I think Godot 4's issue with autoloads makes it majorly more difficult to code in lots of scenarios.

First, let's analyze the Godot 4 autoload to make sure there is nothing wrong with it:
![image](https://github.com/godotengine/godot/assets/75461200/86040295-65e7-471c-a399-a4ff72e0190c)

As shown in the image, all (most) variables have been statically typed, so we should have no issues accessing them.

In Godot 4, variables inside of autoloads had no problem getting picked up by intellisense, as seen here:
![image](https://github.com/godotengine/godot/assets/75461200/5f1ea310-7e66-42b6-8566-219d9214054c)

As shown in the image, Godot 4 had no issues with loading the variables, it's just that trying to access the variables inside of these objects have issues.
![image](https://github.com/godotengine/godot/assets/75461200/dfbbfedc-6832-4cce-87d3-df4d51bc1060)

Here's a temporary solution, but very hacky and kind of ruins the clear (I am an autoload) that autoload variables give:
![image](https://github.com/godotengine/godot/assets/75461200/c09da7b5-163d-4402-9058-29c99fe68d89)

### Steps to reproduce

1. Create a script that has a variable, give it a class name
2. Create an autoload script, store a variable inside the autoload script that is statically casted to the script you created in step one
4. Create another script, try to access the autoload's variable's variable, you will see that the variable inside the autoload is picked up by intellisense, but the variable inside of the variable is not picked up by intellisense.

### Minimal reproduction project

[autoload_intellisense_issue.zip](https://github.com/godotengine/godot/files/11795196/autoload_intellisense_issue.zip)

AThousandShips:
Think I saw something similar somewhere, will see if I can find it

Issue #77643, opened 2023-05-30
Autoload global variables behaving inconsistently, properties not having types inferred


nazjun:
### Godot version

v4.0.3.stable.mono.official [5222a99f5]

### System information

- OS: Windows 10
- Godot Version: 4.0.3.stable.mono
- Rendering Driver: Any
- Rendering Method: Any

### Issue description

I've noticed some strange behaviors arising from the use of static typing (along with type casting in general) with autoload global variables.

Here are my findings so far:
- Creating an autoload from a .tscn file will cause the singleton node referenced by the global variable to be unconvertible to the type of the script specified after `class_name`, if it exists.
- Creating an autoload from a .gd file, however, _does_ allow the variable to be converted to the custom type.
- Properties of custom reference types in an autoload will _not_ be inferred with the `:=` operator if the autoload is .tscn-based, but _will_ if it is .gd-based.
- Properties of custom reference types in an autoload class will not have their own custom properties appear in the auto-completion menu, if the autoload is accessed with its global variable. The auto-completion menu will only reflect the built-in classes they're derived from. Creating a variable of the singleton's custom type (which is only possible with .gd autoloads) and then accessing the reference property that way is the only way I can find that allows the auto-completion to accurately reflect the current type of the reference.

The best way I can work around this while maintaining type safety is by only using .gd-based autoloads, but unless I store an explicit reference to the singleton as its proper type, I won't be able to see the full classes of its custom properties.

### Steps to reproduce

In the main scene's script, `main.gd`:
```gdscript
class_name Main extends Node

var number: int = 7

func _enter_tree() -> void:
    Global.main = self
    Global2.main = self

func _ready() -> void:
    # Does not autofill number property, will also fail to infer type
    var num_0 := Global.main.number
    
    # Autofills number property, but casting fails when using a .tscn autoload
    var global_1 := Global as _Global
    var num_1 := global_1.main.number
    
    # Autofills number property, and casting succeeds when using a .gd autoload
    var global_2 := Global2 as _Global2
    var num_2 := global_2.main.number
```

In the first singleton script, `global.gd`, loaded as a **.tscn**:
```gdscript
class_name _Global extends Node

var main: Main
```

In the first singleton script, `global_2.gd`, loaded as a **.gd**:
```gdscript
class_name _Global2 extends Node

var main: Main
```

The autoload configuration:
![image](https://github.com/godotengine/godot/assets/31663251/749add78-46fc-439e-acf3-da8e2e126edb)

[...]

Issue #71296, opened 2023-01-12
Auto Complete/Intellisense for references in inner autoload variables can't be resolved 


kuhnchris:
### Godot version

4.0.beta10, master-branch (v4.0.beta.custom_build.2ac2db8de)

### System information

Win 11

### Issue description

It seems like there is an issue with inner-class resolving in Godot 4, when you try to access the members of a instance of a auto-load object, it apparently only results in the base Node members being displayed in the auto complete.
It *does* work to use it even if auto-complete does not suggest it, so I assume it's just an error related to the resolving within the LSP (probably, but that's just a guess). Sadly I do not really have any idea where I could investigate this - this happens in VSCode and Godot. It also sometimes fixes itself after a couple of restarts, sometimes it doesn't, deleting the .godot folder doesn't change anything.

Here's also a video showing the entire process: https://youtu.be/Q-6sxvcQq84

Edit: Just as a note: If you "redefine" the node in the current script, simply by adding it as a variable at the top of the script, all the resolves are working.

### Steps to reproduce

* Create Project
* Create Class to use as Auto-Load
* Create Class with a function
* Instance "Class with a function" in "Class to use as Auto-Load"
* Create a script in a scene
* Try to use autocomplete on the Auto-Load object
* Try to use autocomplete on the "Class with a function" instance in the Auto-Load object
* Fail to see the function name(s) / objects / variables / pretty much everything

### Minimal reproduction project

[WeirdExampleProjectForIssuesWithGDScriptReload.zip](https://github.com/godotengine/godot/files/10406457/WeirdExampleProjectForIssuesWithGDScriptReload.zip)

kuhnchris:
Also, another weird thing I noticed:
If you declare the variable outside the function (as a "property" of the calling class), you *cannot* access the function auto-complete via "variable.functionName()", *but* you can access it with "self.variable.functionName()"???
Wasn't Godot 4 suppose to remove those pesky "self." references, or are those now required instead of optional?

Video: https://youtu.be/DwnQ7yc0Xlw

Calinou:
- Related to https://github.com/godotengine/godot/issues/22824 (which is the Godot 3.x counterpart of this issue, with its different GDScript implementation).

Nukiloco:
Currently having this issue as well when I have a singleton that controls my game. This only happens if you access it by the global node.

Interesting enough assigning global to a variable in your script will bypass this issue
```
func _ready():
[...]

PR #85215, opened 2023-11-22
Handle global classes when resolving type from `PropertyInfo`

### LINKED ISSUE #74888: Code completion (type inference?) not working for custom types
### Godot version

4.0

### System information

MacOS

### Issue description

When calling a function that has a custom class, the return type is disregarded unless I explicitly cast the result to the type in the editor (breaking code completion)...

I have this function defined in a global class:
image

When I try to use the value returned by it, I get incorrect code completion:
image

Creating a new variable of the same type works (in the same file)
image

And if I cast the type to itself it also works...
image

### Steps to reproduce

Create a custom class with class_name.
Create a function in a global scene that returns an instance of that class type.
Try to reference the return value and see code completion is broken.

I've provided specific screenshots describing the issue.

HolonProduction:
Fixes partially #74888
Fixes partially #71296 
Fixes partially #78454

Fixes partially #77643

Fixes partially godotengine/godot-proposals#8545 (Should be closed if #85239 is merged)

In theory we could also resolve to `CLASS` type if the file is GDScript but we would need to pass the completion context through the whole type resolve chain to add the depending parser. Not quite sure if this is worth the trouble.

**Edit:**
Sorting doesn't work since `_get_property_location` doesn't handle global classes. But I am a fan of smaller self contained PR's.

### DIFF
--- a/modules/gdscript/gdscript_analyzer.cpp
+++ b/modules/gdscript/gdscript_analyzer.cpp
@@ -4842,8 +4842,19 @@ GDScriptParser::DataType GDScriptAnalyzer::type_from_property(const PropertyInfo
 	}
 	result.builtin_type = p_property.type;
 	if (p_property.type == Variant::OBJECT) {
-		result.kind = GDScriptParser::DataType::NATIVE;
-		result.native_type = p_property.class_name == StringName() ? SNAME("Object") : p_property.class_name;
+		if (ScriptServer::is_global_class(p_property.class_name)) {
+			result.kind = GDScriptParser::DataType::SCRIPT;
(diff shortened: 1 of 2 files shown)


HolonProduction:
Just noticed we have [the same behavior in the Analyzer](https://github.com/godotengine/godot/blob/a34814b0b6767d1e894e0d7ef8fa7076b4eb20ce/modules/gdscript/gdscript_analyzer.cpp#L4844-L4847). Should probably be changed there as well.

vnen (approved):
LGTM. Do you want to make the change to the analyzer as well?

HolonProduction:
Did the analyzer change as well

akien-mga:
Thanks!

PR #85216, opened 2023-11-22
[iOS] Fix GDExtension init callback array reallocation.

bruvzg:
It was always allocating 32 items, instead of adding 32.


akien-mga (approved):

akien-mga:
Thanks!

PR #85219, opened 2023-11-22
Remove GI methods in parentheses from light baking options

dsnopek:
Presently, on lights, the "Bake Mode" property has 3 options:

- Disabled
- Static (VoxelGI/SDFGI/LightmapGI)
- Dynamic (VoxelGI/SDFGI)

I think this is potentially confusing (it confused me :-)) because it makes it sound like lights marked "Dynamic (VoxelGI/SDFGI)" won't be baked into `LightmapGI` lightmaps.

However, they _are_ baked into `LightmapGI` lightmaps, but only the bounced light, not the direct light.

So, in this case, I think it'd be best to just remove the GI methods from the option names, since all the options do apply to all GI methods.

~~There are a number of other properties with similar options, for example, `gi_mode` on `VisualInstance3D`, but in that case I think it's OK because it's true (if marked as "Dynamic (VoxelGI only)" that visual instance won't receive light from `LightmapGI`). The same goes for the `meshes/light_baking` option from `ResourceImporterScene`.~~

**UPDATE**: Per @Calinou's suggestion, this PR now removes the GI method names from all light baking options.

### DIFF
--- a/editor/import/resource_importer_scene.cpp
+++ b/editor/import/resource_importer_scene.cpp
@@ -1933,7 +1933,7 @@ void ResourceImporterScene::get_import_options(const String &p_path, List<Import
 	r_options->push_back(ImportOption(PropertyInfo(Variant::BOOL, "meshes/ensure_tangents"), true));
 	r_options->push_back(ImportOption(PropertyInfo(Variant::BOOL, "meshes/generate_lods"), true));
 	r_options->push_back(ImportOption(PropertyInfo(Variant::BOOL, "meshes/create_shadow_meshes"), true));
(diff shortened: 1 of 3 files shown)


Calinou:
> (if marked as "Dynamic (VoxelGI only)" that visual instance won't receive light from LightmapGI). 

It'll receive indirect light from LightmapGI's dynamic probe system.

dsnopek:
Alright, I've updated the PR to remove the GI method names from all the light baking options

Calinou (approved):

YuriSizov:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #78929, opened 2023-07-02
Fix autocompletion not working for return values of some methods

heppocogne:
This commit fixes #78820.
This issue is because the return values of some methods are always inferred as `Variant` by script editor.
So, I updated `_guess_identifier_type` to use the annotated/inferred type if available.

PR #85224, opened 2023-11-22
Prefer identifiers annotated type if assigned type is incompatible to it

### LINKED ISSUE #74888: Code completion (type inference?) not working for custom types

### LINKED ISSUE #78820: Autocompletion does not work for Dictionary.get() return value
### Godot version

4.0.3-stable

### System information

Windows 11, NVIDIA GeForce GTX 1650

### Issue description

When I assigned Dictionary.get() return value to a typed variable, autocompletion dialog did not show up.

### Steps to reproduce

Please try the following script. Autocompletion dialog is not shown up for `my_class_get` even if typing ".my_" or something.
```gdscript
extends Node

class MyClass:
	var my_var


	func my_func():
		pass


var dic:={
	"a":MyClass.new(),
}


func _ready():
	var my_class_get:MyClass=dic.get("a",MyClass.new())
	my_class_get	# autocompletion does not work
	var my_class_bracket:MyClass=dic["a"]
	my_class_bracket	# autocompletion works
```

N/A

### LINKED ISSUE #78860: Intellisense on my custom class don't work when I assign a node to a variable of this class
### Godot version

4.0.3

### System information

Windows 10

### Issue description

So the issue is that when i assign a node to variable of class i created the intellisense is gone

b
it should show the init and input function as on the screenshot before
c
this bug also appears when i assign node in ready function like this
d

### Steps to reproduce

1. Create a class with a class_name and function or variable inside
2. In any other script (that is attached to any node) create variable of class from first step (for example "var x: MyClassName" )
3. Create child node and attach to it script created in first step
4. In any function of parent script write "x." (and write at list first letter of the function inside this class) and it should show function or variable in a list
5. Now after "var x: MyClassName" add "= $node_name"
6. Now when you write "x." in script it won't show a function or variable you created

N/A

### LINKED ISSUE #86513: Autocomplete doesn't correctly work on implicitly casted variables
### Tested versions

Experienced in 4.1 and 4.2 stable versions.

### System information

Godot v4.2.1.stable - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4080 (NVIDIA; 31.0.15.4601) - AMD Ryzen 9 7900X 12-Core Processor (24 Threads)

### Issue description

**This issue is related to the autocomplete features of using static types in GDScript, and not the actual execution behaviour. The actual execution behaviour works as expected.**

In GDscript, there are two ways to assign variables with a specific type, they are the following, with an example:
1. `gdscript var result := foo() as Array` (explicit cast using the `as` keyword)
2. `var result: Array = foo()` (implicit cast using static typing `: <TypeName>`operator)

One benefit of static typing is that it provides useful autocomplete suggestions when interacting with the variable. For example, the above `result` is of type `Array`, so autocomplete correctly shows methods associated with the array class:
image

This works all and well until the assignment is the value of a function call that returns a `Variant` type such as `.get` on a dictionary or any user-defined function like this simple example:

```gdscript
func arr_or_dict(x: int) -> Variant:
    if x == 0:
        return []
    return {}
```

The above example returns a `Variant` type of either `Array` or `Dictionary`. When casted using method 1 (`as` keyword`), the autocomplete works correctly, displaying the attributes associated with the casted type:
```gdscript
func explicit_cast():
    var result := arr_or_dict(0) as Array
```
image

The problem occurs when using method 2 to implicitly cast the return value - It appears that the autocomplete suggestions are incorrectly showing the methods associated with `Dictionary`, despite it being statically declared as type `Array`:
```gdscript
func implict_cast():
    var result: Array = arr_or_dict(0)
	
```
image

After testing it a bit more, it seems like the editor will always show options associated with the type of the variable of the _last_ return statement of the function and ignore the declared type. In the example given, it is `Dictionary` from the last line `return {}`. If I switch `[]` and `{}` and cast `result` into `Dictionary`, it will show array methods because the type of the last return statement will be `Array`.

The expected behaviour should be that the autocomplete suggestions always show the methods declared by the static type `Array`, exactly like the behaviour of explicitly using `as` keyword to cast.

Despite the autocomplete showing incorrect options, the execution behaviour is still correct, `result` is still of type `Array`, and calling dictionary methods given by the faulty autocomplete will still throw an error.

### Steps to reproduce

1. Attach empty script to any node in any scene
2. Paste in the MRP
3. Uncomment the `result.` lines
4. In the Godot editor, trigger the autocomplete associated with the above line and observe the incorrect suggestions of `implict_cast` compared to the correct `explicit_cast`

(Code is below, will also include a zipped project on Godot 4.2.1: [godot-autocomplete-bug-mrp.zip](https://github.com/godotengine/godot/files/13767500/godot-autocomplete-bug-mrp.zip))
```gdscript
extends Node2D

func arr_or_dict(x: int) -> Variant:
	if x == 0:
		return []
	return {}

# explicit cast (correct)
func explicit_cast():
	var result := arr_or_dict(0) as Array
	# autocomplete works correctly when using `result.`
	# result.

# implicit cast (incorrect)
func implict_cast():
	var result: Array = arr_or_dict(0)
	# when using `result.`, autocomplete shows suggestions associated with the type of the LAST return value (dict in this case)
	# result.
```

### LINKED ISSUE #86871: Autocomplete suggests invalid methods, ignoring static typing
### Tested versions

Godot v4.2.1.stable (Steam)

### System information

Godot v4.2.1.stable - Windows 10.0.19045 - GLES3 (Compatibility) - NVIDIA GeForce RTX 2060 SUPER (NVIDIA; 31.0.15.3713) - AMD Ryzen 5 3500 6-Core Processor (6 Threads)

### Issue description

I am currently working on a project allowing user to access files on the computer. I note suitable files in PackedStringArray to use less memory and fast-up the application. When it's time to not use these files anymore I want to erase them from PackedStringArray. The syntaxis hint showed me that the method I want can be used, but then for some reason debugger complains that such method does not exist in PackedStringArray. Later on, I checked up the embedded documentation and it indeed shows that 'erase' method does not exist.
I do not know if it was intended for the method to be absent, but I hope it was not as it would helped me a lot in my project.
Report_1
Report_2

### Steps to reproduce

Create a PackedStringArray and after that try to type ".erase()". The hint appears, but debugger complains.

N/A

HolonProduction:
Clashes with the PR #78929

Fixes partially #74888
Fixes partially #71296
Fixes partially #78454

Fixes #78860
Fixes #78820

Fixes partially #86172

Fixes #86513
Fixes #86871

A small self contained example demonstrating the problem:
```
func _ready():
    var a: String = 1
    a. # no string methods are suggested
```

Obviously the example isn't valid GDScript, but in more complex situations like #74888 the incompatible assignment might result from an issue with interfering the type of the assignment which might still be valid at runtime. In such cases we should respect the type that the user specified.

Implementation detail:
- If there are any ideas to prevent changing the analyzer without copying the over hundred line long function I'm all in for it.

TODO:
- ~it currently only works with local variables but as #86172 shows it can also happen with class properties~ Done

### DIFF
--- a/modules/gdscript/gdscript_analyzer.cpp
+++ b/modules/gdscript/gdscript_analyzer.cpp
@@ -5315,8 +5315,21 @@ GDScriptParser::DataType GDScriptAnalyzer::get_operation_type(Variant::Operator 
 	return result;
 }
 
-// TODO: Add safe/unsafe return variable (for variant cases)
 bool GDScriptAnalyzer::is_type_compatible(const GDScriptParser::DataType &p_target, const GDScriptParser::DataType &p_source, bool p_allow_implicit_conversion, const GDScriptParser::Node *p_source_node) {
+#ifdef DEBUG_ENABLED
+	if (p_source_node) {
+		if (p_target.kind == GDScriptParser::DataType::ENUM) {
+			if (p_source.kind == GDScriptParser::DataType::BUILTIN && p_source.builtin_type == Variant::INT) {
+				parser->push_warning(p_source_node, GDScriptWarning::INT_AS_ENUM_WITHOUT_CAST);
+			}
+		}
+	}
+#endif
+	return check_type_compatibility(p_target, p_source, p_allow_implicit_conversion, p_source_node);
+}
+
+// TODO: Add safe/unsafe return variable (for variant cases)
+bool GDScriptAnalyzer::check_type_compatibility(const GDScriptParser::DataType &p_target, const GDScriptParser::DataType &p_source, bool p_allow_implicit_conversion, const GDScriptParser::Node *p_source_node) {
 	// These return "true" so it doesn't affect users negatively.
 	ERR_FAIL_COND_V_MSG(!p_target.is_set(), true, "Parser bug (please report): Trying to check compatibility of unset target type");
 	ERR_FAIL_COND_V_MSG(!p_source.is_set(), true, "Parser bug (please report): Trying to check compatibility of unset value type");
@@ -5351,11 +5364,6 @@ bool GDScriptAnalyzer::is_type_compatible(const GDScriptParser::DataType &p_targ
 
 	if (p_target.kind == GDScriptParser::DataType::ENUM) {
 		if (p_source.kind == GDScriptParser::DataType::BUILTIN && p_source.builtin_type == Variant::INT) {
-#ifdef DEBUG_ENABLED
-			if (p_source_node) {
-				parser->push_warning(p_source_node, GDScriptWarning::INT_AS_ENUM_WITHOUT_CAST);
-			}
-#endif
 			return true;
 		}
 		if (p_source.kind == GDScriptParser::DataType::ENUM) {
--- a/modules/gdscript/gdscript_analyzer.h
+++ b/modules/gdscript/gdscript_analyzer.h
@@ -147,6 +147,7 @@ public:
 
 	Variant make_variable_default_value(GDScriptParser::VariableNode *p_variable);
 	const HashMap<String, Ref<GDScriptParserRef>> &get_depended_parsers();
+	static bool check_type_compatibility(const GDScriptParser::DataType &p_target, const GDScriptParser::DataType &p_source, bool p_allow_implicit_conversion = false, const GDScriptParser::Node *p_source_node = nullptr);
 
 	GDScriptAnalyzer(GDScriptParser *p_parser);
 };
--- a/modules/gdscript/gdscript_editor.cpp
+++ b/modules/gdscript/gdscript_editor.cpp
@@ -2009,6 +2009,21 @@ static bool _guess_identifier_type(GDScriptParser::CompletionContext &p_context,
 			default:
 				break;
 		}
+	} else {
+		if (p_context.current_class) {
+			GDScriptCompletionIdentifier base_identifier;
+
+			GDScriptCompletionIdentifier base;
+			base.value = p_context.base;
+			base.type.type_source = GDScriptParser::DataType::ANNOTATED_EXPLICIT;
+			base.type.kind = GDScriptParser::DataType::CLASS;
+			base.type.class_type = p_context.current_class;
+			base.type.is_meta_type = p_context.current_function && p_context.current_function->is_static;
+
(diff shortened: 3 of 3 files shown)


vnen (approved):
LGTM

modules/gdscript/gdscript_analyzer.cpp
```diff
@@ -5249,8 +5249,21 @@ GDScriptParser::DataType GDScriptAnalyzer::get_operation_type(Variant::Operator
...
-// TODO: Add safe/unsafe return variable (for variant cases)
 bool GDScriptAnalyzer::is_type_compatible(const GDScriptParser::DataType &p_target, const GDScriptParser::DataType &p_source, bool p_allow_implicit_conversion, const GDScriptParser::Node *p_source_node) {
+#ifdef DEBUG_ENABLED
+	if (p_target.kind == GDScriptParser::DataType::ENUM) {
+		if (p_source.kind == GDScriptParser::DataType::BUILTIN && p_source.builtin_type == Variant::INT) {
+			if (p_source_node) {
+				parser->push_warning(p_source_node, GDScriptWarning::INT_AS_ENUM_WITHOUT_CAST);
+			}
+		}
+	}
```
vnen, line 5260: Actually, since this is only relevant if there is a source node, I think the `p_source_node` check should be the outermost one.

vnen (changes requested):

vnen (approved):
LGTM. Sorry for the delay in reviewing it

YuriSizov:
Should we consider https://github.com/godotengine/godot/pull/78929 superseded?

YuriSizov:
Thanks!

PR #85227, opened 2023-11-22
Fix crash caused by conflicting menu option IDs

KoBeWi:
Regression from #85150
Trying to make new Script or new Shader was interpreted as Show in Filesystem (which shifted to 10) and resulted in `get_path()` call on null resource.


akien-mga (approved):

YuriSizov:
I should've spotted it earlier, but technically any change to this enum could create conflicts with user projects if they extend and customize the resource picker. We expose hooks to add custom options to this menu and users can provide any IDs they want. We should probably suggest they use some safe range with a high base number in the documentation.

akien-mga:
Thanks!

PR #85228, opened 2023-11-22
Don't insert newline while saving

KoBeWi:
Follow-up to #85154

Changes the new check into an error; saving while saving is now invalid behavior. I tweaked Script Editor to avoid the new error. There might be some cases when the newline is not applied correctly, but it's probably rare and harmless. Saving script in Script Editor does call `apply_changes()`.


akien-mga (approved):
Seems good. Might be worth rebasing as it's been 7 months.

akien-mga:
Thanks!

PR #85026, opened 2023-11-17
Make `screen_get_refresh_rate()` respect iOS Low Power Mode

ztc0611:
Fixes #84634.

iOS/iPadOS ramps down from 120hz to 60hz when on low power mode, and DisplayServer.screen_get_refresh_rate() didn't account for this.

PR #85230, opened 2023-11-22
[3.x] [iOS] Make `OSIPhone::get_screen_refresh_rate` respect iOS Low Power Mode

ztc0611:
Backport of #85026 for 3.x.

### DIFF
--- a/platform/iphone/os_iphone.mm
+++ b/platform/iphone/os_iphone.mm
@@ -653,7 +653,11 @@ int OSIPhone::get_screen_dpi(int p_screen) const {
 }
 
 float OSIPhone::get_screen_refresh_rate(int p_screen) const {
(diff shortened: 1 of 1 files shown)


AThousandShips:
Unsure if the changed file name is an issue but otherwise this is the kind of change that would be cherry picked as part of a batch, not individually, we only do this kind of manual cherry picking if they are needed

ztc0611:
I thought the change from OSIPhone in 3.x to OS_IOS in 4.x would cause an issue. My bad!

akien-mga:
It's fine to make dedicated PRs for 3.x, we don't usually do a lot of cherrypicking anymore so it's good if users help in the process.

akien-mga (approved):

akien-mga:
Thanks!

PR #81939, opened 2023-09-19
Replace flat buttons with flat-styled buttons with a visible pressed state

YuriSizov:
`ToolButton` is back, baby!.. Wait, no, that sounds wrong. But we do have a usability problem which `ToolButton` could've solved.

### The problem

We use flat buttons throughout the editor UI, and in a couple of `Control` nodes (`GraphEdit` and `FileDialog`). The problem with flat buttons is that they by design don't render any styleboxes except for the focus outline. So no visible pressed state for them, except for slightly tinted icon and/or text. In the editor, by default, it's a light blue color, and in the default theme it's a slightly different shade of white (which makes the pressed state completely invisible in the `GraphEdit` toolbar).

image
_Which of these buttons is toggled on and which are toggled off? This is not a trick question!_

Within the editor users have expressed trouble noticing when some tools are enabled. This can depend a lot on the exact theme and theme colors used, but the default look definitely has this issue where lightly colored blue text or blue icon is getting lost on top of a bluish background.

https://youtu.be/nu6g942bAIU?t=846

### The solution

One way to solve this problem is to make flat buttons render the pressed state as well, but that would break compatibility and expectation compared to what we have right now.

Thankfully, we have theme type variations. So I propose we use them here and introduce the `FlatButton` type variation (and also `FlatMenuButton` for `MenuButton`), which would have empty styles for everything except for the pressed state. This would make them look like exactly like flat buttons, but with a proper style for when these buttons are pressed.

https://github.com/godotengine/godot/assets/11782833/5a660532-e4b1-4c4a-a4bc-8abcab571309

With this PR these variations are present in the default theme and used in `GraphEdit` and `FileDialog` as well:

godot windows editor dev x86_64_2023-09-19_17-58-29

_It's going to look even better with https://github.com/godotengine/godot/pull/81582._

### The implementation

I went through all toolbars and toolbar-alikes in the editor that use flat buttons and updated them to use this approach. Here are a few examples:

<details>
<summary>Videos and pictures!</summary>

https://github.com/godotengine/godot/assets/11782833/bea7a697-ba62-46fd-85e4-ec88d3c7ee70

https://github.com/godotengine/godot/assets/11782833/62395a3c-f1e7-4fe2-ab34-c2307e6f4105

godot windows editor dev x86_64_2023-09-19_17-58-56
godot windows editor dev x86_64_2023-09-19_18-00-46
godot windows editor dev x86_64_2023-09-19_18-01-13
</details>

[...]

PR #85236, opened 2023-11-22
Use the flat button style on switches in scene and node docks

YuriSizov:
A follow-up to https://github.com/godotengine/godot/pull/81939.

So I have always found these two buttons to be awkwardly designed. A lot of empty space, no panels, just accented text to show which is active. Never arrived to a better design myself, but now that we have the `FlatButton` style, thanks to https://github.com/godotengine/godot/pull/81939, I think it would be appropriate to use it here.

It still feels a bit disconnected, but I think this is still better and it uses a consistent style from other parts of the editor. So should be a safe choice for now.

NVIDIA_Share_2023-11-22_19-40-00

https://github.com/godotengine/godot/assets/11782833/a53e2d47-7a1f-4195-af63-60a440d8ab77

### DIFF
--- a/editor/node_dock.cpp
+++ b/editor/node_dock.cpp
@@ -104,7 +104,7 @@ NodeDock::NodeDock() {
 	mode_hb->hide();
 
 	connections_button = memnew(Button);
-	connections_button->set_flat(true);
+	connections_button->set_theme_type_variation("FlatButton");
 	connections_button->set_text(TTR("Signals"));
 	connections_button->set_toggle_mode(true);
 	connections_button->set_pressed(true);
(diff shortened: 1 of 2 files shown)


KoBeWi:
Should Node dock be changed too? It has icons, but the buttons are flat.

Calinou:
Should these buttons be converted to use TabBar instead?

YuriSizov:
> Should Node dock be changed too? It has icons, but the buttons are flat.

Sure!

image

> Should these buttons be converted to use TabBar instead?

That was my initial idea back then, but I'm not sure it would work well inside of a dock which itself is a part of some tabs.

AThousandShips (approved):
I like it

groud (approved):

YuriSizov:
Thanks for reviews!

PR #85244, opened 2023-11-22
Give scripts access to RigidBody2D collision details

LunaticInAHat:
This PR adds some plumbing to `RigidBody2D`, such that more information about contacts is (readily) available to scripts, if contact-monitoring has been enabled.

Let's say you're making a racing game (I am). Cars are `RigidBody2D`. If a car lightly brushes a guardrail, maybe there should be a little shower of sparks. If a car smashes head-on into that guardrail, at full speed, it should take some damage.

In order for the game logic to figure out how to react to a collision, it needs to know about where the point of collision was, how "direct" the collision was, and how much energy got exchanged between the bodies. Right now, there's no really direct way for a script on a `RigidBody2D` to know that; it can receive a signal when bodies come into contact with one another (`body_entered`), and it can query the list of all the bodies it's in contact with (`get_colliding_bodies()`), but neither of those convey any qualitative information about the collision. `CharacterBody2D` can already report information of a similar flavor, through its `get_slide_collision()` method, but that doesn't help users of `RigidBody2D`.

The physics server knows quite a bit more about the collisions, and that information is accessible through `PhysicsDirectBodyState2D`, however:
  - This is not intuitive, and not documented anywhere in `RigidBody2D` itself (or any of its parent classes), so it's hard to find
  - It's annoying to have to make direct state accesses in order to get the information
  - It doesn't seem like there's anything that ensures any kind of coherency between the information that the `PhysicsDirectBodyState2D` would report, versus what `RigidBody2D::get_colliding_bodies()` would say

Thus, I have expanded the book-keeping that was already being done inside of `RigidBody2D` for contact monitoring, to preserve information about the collisions, and make it readily accessible to scripts. I tried to follow the pattern laid out by `CharacterBody2D` and `KinematicCollision2D` wherever possible.


LunaticInAHat:
Since I imply the usecase of determining energy-of-collision in my overall PR description, I do want to note that this particular usecase still will not quite work with this PR: I've added the relevant APIs to be able to retrieve the information about contacts, but unfortunately the physics server always reports an impulse of 0.0 at the moment of contact (this happens in `GodotBodyPair2D::pre_solve()`). Since `RigidBody2D` only processes `body_entered` (and related book-keeping) at the first moment of contact, this means that `RigidCollision2D::get_impulse()` always reports `Vector2.ZERO`

From the data I'm currently seeing, it appears that the 2D physics engine may need two physics frames to respond to collisions -- it appears to detect the collision on one frame (and reports it with an impulse of 0.0), and then on the next frame it actually applies an impulse to the objects involved (reporting the contact again, with an impulse != 0.0)

It's not clear to me yet whether this is a bug, or expected behavior (maybe someone with knowledge of the 2D physics engine could comment?).

If it is expected behavior, then I suppose something slightly different needs to be done in `RigidBody2D`. I believe we could easily update the impulse value with each collision frame that we remain in contact with the other body. This would help users who call `RigidBody2D::get_body_collision()` in their `_physics_process()`, but it is sub-optimal, as it means that `body_entered` handlers won't have valid impulse information available to them (the signal goes out as soon as the contact occurs, so the handler will just see the spurious 0.0 value)

If it is a bug, then we probably want to see if we can fix it in the physics server, so that signal handlers can have valid impulse information available to them.

scene/2d/physics_body_2d.cpp
```diff
@@ -1103,6 +1158,97 @@ void RigidBody2D::_reload_physics_characteristics() {
...
+Vector2 RigidCollision2D::get_local_position() const {
+	return local_pos;
+}
+
+Vector2 RigidCollision2D::get_local_normal() const {
+	return local_normal;
+}
+
+Object *RigidCollision2D::get_local_shape(int p_pair_idx) const {
+	ERR_FAIL_COND_V(p_pair_idx >= local_shapes.size(), nullptr);
```
AThousandShips, line 1174:
```suggestion
	ERR_FAIL_INDEX_V(p_pair_idx, local_shapes.size(), nullptr);
```
The argument is unsigned, so needs to check for negative values, this macro does this and gives a clearer message, same for all these

scene/2d/physics_body_2d.h
```diff
@@ -191,11 +192,32 @@ class RigidBody2D : public PhysicsBody2D {
...
+		int shape = 0;
+		int local_shape = 0;
+		Vector2 local_pos;
+		Vector2 local_normal;
+		Vector2 local_vel_at_pos;
+		Vector2 collider_pos;
+		Vector2 collider_vel_at_pos;
+		Vector2 impulse;
+
+		RigidBody2D_AddAction() = default;
```
AThousandShips, line 207:
```suggestion
		RigidBody2D_AddAction() {}
```
We use this syntax elsewhere in this file, and I don't think we use the `default` syntax anywhere

scene/2d/physics_body_2d.cpp
```diff
@@ -874,6 +890,43 @@ TypedArray<Node2D> RigidBody2D::get_colliding_bodies() const {
 	return ret;
 }
 
+Ref<RigidCollision2D> RigidBody2D::get_body_collision(int p_collider_idx) const {
+	Ref<RigidCollision2D> ret;
+	ERR_FAIL_NULL_V(contact_monitor, ret);
+	ERR_FAIL_COND_V(p_collider_idx >= (int)contact_monitor->body_map.size(), ret);
```
AThousandShips, line 896:
```suggestion
	ERR_FAIL_INDEX_V(p_collider_idx, (int)contact_monitor->body_map.size(), ret);
```
Same here

scene/2d/physics_body_2d.cpp
```diff
@@ -475,36 +493,29 @@ void RigidBody2D::_body_state_changed(PhysicsDirectBodyState2D *p_state) {
 			}
 		}
 
-		_RigidBody2DInOut *toadd = (_RigidBody2DInOut *)alloca(p_state->get_contact_count() * sizeof(_RigidBody2DInOut));
+		RigidBody2D_AddAction *toadd = (RigidBody2D_AddAction *)alloca(p_state->get_contact_count() * sizeof(*toadd));
```
AThousandShips, line 496:
```suggestion
		RigidBody2D_AddAction *toadd = (RigidBody2D_AddAction *)alloca(p_state->get_contact_count() * sizeof(RigidBody2D_AddAction));
```

scene/2d/physics_body_2d.h
LunaticInAHat: It is used in various other parts the engine (e.g., the Vulkan renderer, the GDScript module), but does appear to be in the minority.

scene/2d/physics_body_2d.cpp
LunaticInAHat: Just want to note that while I have implemented the requested change, it makes it easier to screw up this code in the future. The compiler knows the type that `toadd` points to; it will always allocate the correct amount of space. Hard-coding the type adds an opportunity for someone to forget to update their code (should the type of `toadd` ever change), and end up allocating the wrong size.

AThousandShips: It is more confusing IMO, and we don't do this anywhere else AFAIK, it should be the correct type to make it clear

scene/2d/physics_body_2d.h
AThousandShips: You didn't format it as specified 🙂

AThousandShips: Yes, see the timestamp, I commented *before* you fixed this

scene/2d/physics_body_2d.cpp
LunaticInAHat: There is a light sprinkling of other instances of this method, through the renderer, the GDScript module, and some of the Windows platform code. It's pretty much everywhere in various thirdparty modules, because enough developers have been bitten in the ass by changing the type in one place, and forgetting to change it in the other place, that it is a well-ingrained habit for long-time C developers. C++ tends not to want it as often, because there are more powerful ways to manage memory. That said, this particular line is a lot more "C" than "C++", which is why we have the ugliness of having to repeat our type three times, in the first place. I suspect the core engine probably provides better ways of doing this allocation, but I don't know what they are.

Regardless, I have made the change you requested.

scene/2d/physics_body_2d.h
LunaticInAHat: Ah, fair enough. Apologies for the double-push :)

scene/2d/physics_body_2d.cpp
LunaticInAHat: Well, like, I said, I updated it as you suggested.

AThousandShips: I know? You were stating your points, and I mine, I'm not sure what you're trying to say. I just said that it was my opinion and not objective 🙂

LunaticInAHat: Oh, no, not at all. I shared your interpretation, that we were discussing possible approaches, and rationale for them. I just noticed that you hadn't marked the conversation as resolved, as you had done with the others, so I wasn't sure if you were waiting for some further action from my side.

AThousandShips: Only left it open to continue discussing 🙂 I'm sorry for the confusion

PR #85222, opened 2023-11-22
Rework the surface upgrade tool to inform users without blocking

YuriSizov:
This removes the immediate confirmation dialog and insteads prints the message to the editor log (and it also appears as a toast). The immediate dialog is a devil's plaything, and it cannot be used in this scenario (if it can be used anywhere at all). The condition that triggers the SUT can happen during any attempt by the rendering server to read a mesh. This means it will conflict with a number of editor processes, like loading, importing, preview generation, export, CLI mode, etc.

So while this is less on the nose as far as informing users goes, it's also our best option to use the log and the toaster.

Closes https://github.com/godotengine/godot/issues/85202.
Closes https://github.com/godotengine/godot/issues/85201.
Likely closes https://github.com/godotengine/godot/issues/84732.

I will test a bunch of previously affected projects to see if there are no regressions. In the meantime, I'd appreciate if others double-checked what they can.

PR #85249, opened 2023-11-22
Suppress surface upgrade warnings when showing SurfaceUpgradeTool warning

clayjohn:
This fixes the warning spam reported here: https://github.com/godotengine/godot/pull/85222#issuecomment-1823451782

The problem comes from the fact that the SurfaceUpgradeTool callback was called for each out of date surface and the old surface-specific warning was still shown to the user.

This PR ensures that, if the surface update callback is handled, the only warning shown to the user is the own pointing them at the SurfaceUpgradeTool. We don't need the old per-surface warning in this case as the user can fix all meshes by running the tool. 

The RenderingServer still has the warning for cases where the scene is run outside the editor. 

Screenshot from 2023-11-22 15-10-26

This also updates the EditorProgress dialogue that runs after the upgrade tool is run. Previously it would start at 100%, then run through re-saving all the files (locked at 100%), then would appear to freeze while deleting the .import files. Now it properly progresses through re-saving all the files, and shows progress while deleting files.

### DIFF
--- a/editor/surface_upgrade_tool.cpp
+++ b/editor/surface_upgrade_tool.cpp
@@ -76,6 +76,7 @@ void SurfaceUpgradeTool::_try_show_popup() {
 	} else {
 		singleton->_show_popup();
 	}
+	RS::get_singleton()->set_warn_on_surface_upgrade(false);
 }
 
 void SurfaceUpgradeTool::_show_popup() {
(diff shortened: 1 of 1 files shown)


YuriSizov (approved):
LGTM, don't see any immediate problems from testing a couple of projects.

akien-mga (approved):
Also tested successfully on a couple projects.

akien-mga:
Thanks!

PR #83488, opened 2023-10-17
Fixed multiple particle issues: division by zero, color ramp override, scale dependent on amount ratio

QbieShay:
Closes  https://github.com/godotengine/godot/issues/83258
Fixes scale and amount ratio being tied together
Fixes initial ramp overriding color

The last two issues were reported to me in DM

PR #85252, opened 2023-11-22
Fix radial inwards velocity clamping incorrectly (regression from #83488)

QbieShay:
When i changed the particle code to avoid division by zero, i didn't notice i left out a part in rdial velocity, which now causes radial velocity around the center to just be a fixed speed. This correctly fixes it.

*Bugsquad edit:*
- Regression fix after #83488

### DIFF
--- a/scene/resources/particle_process_material.cpp
+++ b/scene/resources/particle_process_material.cpp
@@ -762,8 +762,11 @@ void ParticleProcessMaterial::_update_shader() {
 
 	code += "}\n";
 
-	code += "vec3 process_radial_displacement(DynamicsParameters param, float lifetime, inout uint alt_seed, mat4 transform, mat4 emission_transform){\n";
+	code += "vec3 process_radial_displacement(DynamicsParameters param, float lifetime, inout uint alt_seed, mat4 transform, mat4 emission_transform, float delta){\n";
 	code += "	vec3 radial_displacement = vec3(0.0);\n";
+	code += "	if (delta < 0.001){\n";
+	code += "		return radial_displacement;\n";
+	code += "	}\n";
 	code += "	float radial_displacement_multiplier = 1.0;\n";
 	if (tex_parameters[PARAM_RADIAL_VELOCITY].is_valid()) {
 		code += "   radial_displacement_multiplier = texture(radial_velocity_curve, vec2(lifetime)).r;\n";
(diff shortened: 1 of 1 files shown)


scene/resources/particle_process_material.cpp
```diff
@@ -774,7 +777,7 @@ void ParticleProcessMaterial::_update_shader() {
 	code += "	}else{radial_displacement = get_random_direction_from_spread(alt_seed, 360.0)* param.radial_velocity;} \n";
 	code += "	if (radial_displacement_multiplier * param.radial_velocity < 0.0){\n // Prevent inwards velocity to flicker once the point is reached.";
 	code += "		if (length(radial_displacement) > 0.01){\n";
-	code += "		radial_displacement = normalize(radial_displacement) * min(abs((radial_displacement_multiplier * param.radial_velocity)), length(transform[3].xyz - global_pivot));\n";
+	code += "		radial_displacement = normalize(radial_displacement) * min(abs((radial_displacement_multiplier * param.radial_velocity)), length(transform[3].xyz - global_pivot)/delta);\n";
```
akien-mga, line 780:
```suggestion
	code += "		radial_displacement = normalize(radial_displacement) * min(abs((radial_displacement_multiplier * param.radial_velocity)), length(transform[3].xyz - global_pivot) / delta);\n";
```

scene/resources/particle_process_material.cpp
```diff
@@ -774,7 +777,7 @@ void ParticleProcessMaterial::_update_shader() {
 	code += "	}else{radial_displacement = get_random_direction_from_spread(alt_seed, 360.0)* param.radial_velocity;} \n";
 	code += "	if (radial_displacement_multiplier * param.radial_velocity < 0.0){\n // Prevent inwards velocity to flicker once the point is reached.";
 	code += "		if (length(radial_displacement) > 0.01){\n";
-	code += "		radial_displacement = normalize(radial_displacement) * min(abs((radial_displacement_multiplier * param.radial_velocity)), length(transform[3].xyz - global_pivot));\n";
+	code += "		radial_displacement = normalize(radial_displacement) * min(abs((radial_displacement_multiplier * param.radial_velocity)), length(transform[3].xyz - global_pivot) / delta);\n";
```
clayjohn, line 780: I'm a bit skeptical of this change. Prior to https://github.com/godotengine/godot/pull/83488, the `delta` was multiplied by `radial_displacement`, then it was removed, now `radial_displacement` is divided by `delta`. Which is reversing the behaviour from what it was previously. 

Can you provide a little bit more context why dividing is now needed instead of multiplying?

QbieShay: I tried multiplying first and it didn't work so I assumed my math assumptions were wrong

akien-mga: This calls for dimensional analysis :)
But that won't be easy with all these variable names which are unclear about whether they're distances, velocities, speeds, distance deltas, or accelerations :/

QbieShay:
Alright so!

This fix is the correct fix, because the variables that are named "displacement" are actually velocities. They used to be raw displacement, but in order to fix some division by zero i did a rework of the functions to instead calculate velocities, but unfortunately i didn't rename the parameters/functions to correctly represent it.

Doing a bunch of renames at this stage of RC is not ideal, so what I suggest it's the following:
1. merge this fix (otherwise inwards radial velocity is unusable)
2. for 4.3, fix the naming ( I opened and self assigned an issue for it https://github.com/godotengine/godot/issues/85310 )

clayjohn (approved):
Thank you for the explanation. Indeed this is very confusing because the naming consistently refers to a displacement, but it is calculating a velocity. 

On review I can see that all the other components that contribute to `radial_displacement` (and `controlled_displacement`) are velocities. So it makes sense to convert this value into a velocity as well.

akien-mga:
Thanks!

PR #85253, opened 2023-11-23
[Navigation] Do not use travel cost for minimum when re-selecting end point

### LINKED ISSUE #85238: Use min distance to target to re-select end point when the original end point is not reachable in navigation's path finding
### Godot version

4.2 beta 3

### System information

Godot v4.2.beta3 - Windows 10.0.22631 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3060 Laptop GPU (NVIDIA; 31.0.15.3699) - AMD Ryzen 9 6900HS with Radeon Graphics (16 Threads)

### Issue description

The existing pathfinding algorithm for NavAgents, which calculates paths to targets, initially selects the polygon closest to the target. It then attempts to chart a path to this polygon. If the polygon is unreachable from the starting point, the algorithm reevaluates and selects a new target polygon from those it can reach. However, the current method has a flaw: When searching for the reachable polygon closest to target, it multiplies the distance from the polygon to the target by the travel cost of that polygon's region. This approach is problematic because the travel cost of the region should not influence the selection of the polygon. Instead, the algorithm should prioritize the reachable polygon nearest to the target, without multiplying in the region's travel cost.

Here is the line where the multiplication takes place, the fix is just to delete the multiplication

https://github.com/godotengine/godot/blob/5ee983188de97ae027f9b9c1443438063f708a7e/modules/navigation/nav_map.cpp#L375C10-L375C10

### Steps to reproduce

Open the Scene number 6 of the [attached project](https://github.com/godotengine/godot-proposals/files/13443814/nav-issues.zip):

Notice how it stops on the second Nav Region from the top, instead of on the third because it takes into consideration the travel cost (it should stop on the X instead):
image

[attached project](https://github.com/godotengine/godot-proposals/files/13443814/nav-issues.zip), you can switch scenes by pressing the numbers, for this bug you should look at scene 6

sandygk:
This one liner PR fixes the bug, explained here https://github.com/godotengine/godot/issues/85238

Here is a screenshot of the same scene mentioned in the bug, but with the fix applied, please read the [attached bug](https://github.com/godotengine/godot/issues/85238) for an explanation of the issue and for a minimal project to test.

image

*Bugsquad edit:*
- Fixes #85238

### DIFF
--- a/modules/navigation/nav_map.cpp
+++ b/modules/navigation/nav_map.cpp
@@ -372,7 +372,7 @@ Vector<Vector3> NavMap::get_path(Vector3 p_origin, Vector3 p_destination, bool p
 
 		// Stores the further reachable end polygon, in case our goal is not reachable.
 		if (is_reachable) {
-			real_t d = navigation_polys[least_cost_id].entry.distance_to(p_destination) * navigation_polys[least_cost_id].poly->owner->get_travel_cost();
+			real_t d = navigation_polys[least_cost_id].entry.distance_to(p_destination);
 			if (reachable_d > d) {
 				reachable_d = d;
 				reachable_end = navigation_polys[least_cost_id].poly;


smix8 (approved):

AThousandShips:
Could you update your commit message (using `git commit --amend` and then pushing with `git push --force`) to be more descriptive, the minimum was always used, the change is in the value

YuriSizov:

PR #85258, opened 2023-11-23
Fix NavigationRegion2D transform update

### LINKED ISSUE #85242: NavRegion2d merge indicators are only generated when the scene is loaded
### Godot version

4.2 beta 3

### System information

Godot v4.2.beta3 - Windows 10.0.22631 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3060 Laptop GPU (NVIDIA; 31.0.15.3699) - AMD Ryzen 9 6900HS with Radeon Graphics (16 Threads)

### Issue description

The connection indicators between two NavRegion2D  (the pink circles) are not redrawn when they are moved around, instead each nav region move its indicators with it. When the scene is reloaded the indicators are drawn correctly, but then if you move the regions you have to reload the scene to get the correct indicators.

 Here is a gif to illustrate the issue.

GIF 11-22-2023 3-45-46 PM

### Steps to reproduce

This happens with any 2d scene that have 2 or more connected nav regions 2d.

But to simplify things, you can go to the [attached project](https://github.com/godotengine/godot/files/13444254/nav-issues.zip), open the Scene3 and move one of the nav regions in the editor to notice the issue

[nav-issues.zip](https://github.com/godotengine/godot/files/13444262/nav-issues.zip)

smix8:
Fixes NavigationRegion2D transform update.

Fixes https://github.com/godotengine/godot/issues/85242

That line slipped on a refactor.

### DIFF
--- a/scene/2d/navigation_region_2d.cpp
+++ b/scene/2d/navigation_region_2d.cpp
@@ -165,6 +165,7 @@ void NavigationRegion2D::_notification(int p_what) {
 
 		case NOTIFICATION_INTERNAL_PHYSICS_PROCESS: {
 			set_physics_process_internal(false);
+			_region_update_transform();
 		} break;
 
 		case NOTIFICATION_DRAW: {


scene/2d/navigation_region_2d.cpp
```diff
@@ -165,6 +165,7 @@ void NavigationRegion2D::_notification(int p_what) {
 
 		case NOTIFICATION_INTERNAL_PHYSICS_PROCESS: {
 			set_physics_process_internal(false);
+			_region_update_transform();
```
akien-mga, line 168: The way internal physics process is used here as a one shot seems like a convoluted way to do a deferred call. I assume call deferred wouldn't be enough for this method in `NOTOFICATION_TRANSFORM_CHANGED`?

akien-mga (approved):

sandygk:
I tested the fix locally, and I can confirm it Fixes https://github.com/godotengine/godot/issues/85242

GIF 11-23-2023 7-12-06 AM

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.1.

PR #85260, opened 2023-11-23
Fix gravity properties translation in uk.po

vmikhav:
Project properties translation fix


Chaosus:
You need to change it on https://hosted.weblate.org/projects/godot-engine/godot-docs/uk/ rather than modify it directly, after than Akien will sync it with .po files (which is done regularly).

PR #85271, opened 2023-11-23
Correctly check scripts that must inherit `EditorPlugin`

### LINKED ISSUE #85268: A subclass of an GDExtension subclass of EditorPlugin cannot be used as an editor plugin
### Godot version

4.1.3

### System information

Linux

### Issue description

I'm trying to make an editor plugin on Rust:

```rust
#[derive(GodotClass)]
#[class(tool, editor_plugin, init, base=EditorPlugin)]
pub struct MyEditorPluginBase {}
```

But due to some limitations, I still have to use GDScript for some specific things.

So I tried to make a subclass of this Rust class:

```gdscript
@tool
class_name MyEditorPlugin
extends MyEditorPluginBase
```

This would work if `MyEditorPluginBase` was a GDScript class, but since it's a GDExtension class in my case, it doesn't work:

> Unable to load addon script from path: 'res://addons/my_addon/my_editor_plugin.gd' Base type is not EditorPlugin.

But it obviously **IS** EditorPlugin!

godot-editorplugin-docs

godot-editorplugin

If I understand correctly, this behavior is caused by this function, which checks some mysterious "instance base type" instead of real inheritance:

https://github.com/godotengine/godot/blob/fa4a65387ea41506ecc519fc61a06da110dce51c/editor/editor_node.cpp#L3456-L3458

https://github.com/godotengine/godot/blob/fa4a65387ea41506ecc519fc61a06da110dce51c/modules/gdscript/gdscript.cpp#L260-L263

Since the [documentation](https://docs.godotengine.org/en/stable/tutorials/plugins/editor/making_plugins.html#the-script-file) doesn't say that the script must be a **direct** subclass of EditorPlugin (and an indirect subclass works fine in GDScript), I interpret this behavior as a Godot bug.

### Steps to reproduce

See above

[gdext_editor_plugin_test.zip](https://github.com/godotengine/godot/files/13451014/gdext_editor_plugin_test.zip)

YuriSizov:
Closes https://github.com/godotengine/godot/issues/85268. There is a similar case with the visual shader editor plugin and `VisualShaderNodeCustom`, but I'm not familiar enough with this part to make a call if these need to be similarly updated.

I also noticed some related error messages were incorrect, poorly phrased, or too terse. So I updated some of them across the codebase.

### DIFF
--- a/core/io/resource_loader.cpp
+++ b/core/io/resource_loader.cpp
@@ -1113,11 +1113,10 @@ bool ResourceLoader::add_custom_resource_format_loader(String script_path) {
 	Ref<Script> s = res;
 	StringName ibt = s->get_instance_base_type();
 	bool valid_type = ClassDB::is_parent_class(ibt, "ResourceFormatLoader");
-	ERR_FAIL_COND_V_MSG(!valid_type, false, "Script does not inherit a CustomResourceLoader: " + script_path + ".");
+	ERR_FAIL_COND_V_MSG(!valid_type, false, vformat("Failed to add a custom resource loader, script '%s' does not inherit 'ResourceFormatLoader'.", script_path));
 
 	Object *obj = ClassDB::instantiate(ibt);
-
-	ERR_FAIL_NULL_V_MSG(obj, false, "Cannot instance script as custom resource loader, expected 'ResourceFormatLoader' inheritance, got: " + String(ibt) + ".");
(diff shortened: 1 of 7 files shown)


dsnopek (approved):
Thanks! These changes look good to me :-)

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.1.

PR #85275, opened 2023-11-23
Correctly restore Control position on undoing an Anchors Preset

### LINKED ISSUE #85245: Undoing the change of the Anchors Preset of a Control will not restore it's offset/position
### Godot version

4.2rc

### System information

Godot v4.2.rc (fa4a65387) - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1070 (NVIDIA; 31.0.15.3623) - Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz (8 Threads)

### Issue description

If you change the Anchors Preset of a Control, and then hit Undo, the previous Control will have its previous Preset restored, but it won't restore its previous position unless it was the default position of that preset

Video:

https://github.com/godotengine/godot/assets/3089525/74500461-81b5-4eb4-bccc-47384d32dbad

### Steps to reproduce

1. Create a Container and put a Control in it. For example, create a Panel and put a Button in it.
2. Access the Button's Layout, ensure Layout Mode is set to Anchors, and then change Anchor Preset to Bottom Right.
3. Observe the Button is in the bottom right of the Panel
4. Move the Button from its default position. Observe this new position.
5. Set the Anchor Preset to Top Left (or any other). Note the button will change is position.
6. Hit Undo (Ctrl+Z)

**Expected:**
The Button should return to it's previous position, as at the end of step 4.

**Actual:**
The button is instead returned to default position for Bottom Right, the position it was at at the end of Step 3. 

N/A

TheSofox:
Fixes #85245

Ensures that the user changing Anchors Preset or Layout Mode on a Control will have correct Undo behaviour (ie. It'll return to its previous position as expected).

### DIFF
--- a/editor/editor_inspector.cpp
+++ b/editor/editor_inspector.cpp
@@ -3747,7 +3747,11 @@ void EditorInspector::_edit_set(const String &p_name, const Variant &p_value, bo
 		bool valid = false;
 		Variant value = object->get(p_name, &valid);
 		if (valid) {
-			undo_redo->add_undo_property(object, p_name, value);
+			if (Object::cast_to<Control>(object) && (p_name == "anchors_preset" || p_name == "layout_mode")) {
+				undo_redo->add_undo_method(object, "_edit_set_state", Object::cast_to<Control>(object)->_edit_get_state());
+			} else {
+				undo_redo->add_undo_property(object, p_name, value);
+			}
 		}
 
 		List<StringName> linked_properties;


TheSofox:
Rebased PR. Did some quick testing to confirm it still resolves the bug.

KoBeWi:
Anchor presets are set via internal `_set_anchors_layout_preset` method. Instead of hard-coding a specific property in the inspector, you can modify the setter to use `_edit_set_state()` when editor hint is active.

TheSofox:
I understand not wanting to put stuff in `editor_inspector.cpp`, but I'm not sure where else to put it. The setter `_set_anchors_layout_preset()` method doesn't know whether it's an `undo` or `redo` or regular operation. The problem is that changing the anchor immediately resets attributes such as `position` and `size`, so those values need to be stored somewhere so they can be restored on an `undo`, but I don't know how to do this from the `_set_anchors_layout_preset()` when I can't seem to access the `UndoRedo` queue or even whether I should be accessing it from there.

KoBeWi (approved):
A proper solution could be adding `_edit_get_undo()` method to Node, which would allow override the undo property in the class itself, but this PR will do for now.

Repiteo:
Thanks!

PR #85284, opened 2023-11-23
Fix crash when hiding a Control during mouse-entering

### LINKED ISSUE #85277: Crash when hiding a panel from a `NOTIFICATION_MOUSE_ENTER` notification
### Godot version

v4.2.rc.custom_build [702227129]

### System information

Godot v4.2.rc (702227129) - Debian GNU/Linux trixie/sid trixie - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1060 6GB (nvidia) - AMD Ryzen 5 1600 Six-Core Processor (12 Threads)

### Issue description

Found a consistent crash when calling `hide()` from `_notification()`.
It does not happen if the call is deferred.

### Steps to reproduce

The `hide_test.gd` script has a boolean `cause_crash`. If it is true, hovering the mouse over the label crashes without an error in the Godot Editor Console. If it's false it hides the label as expected.

[hide_crash.zip](https://github.com/godotengine/godot/files/13452818/hide_crash.zip)

Sauermann:
`gui.mouse_over` can be set to `nullptr` in the `NOTIFICATION_MOUSE_ENTER`- user-callback a few lines above. This case was previously not handled.

resolve #85277

### DIFF
--- a/scene/main/viewport.cpp
+++ b/scene/main/viewport.cpp
@@ -3207,7 +3207,9 @@ void Viewport::_update_mouse_over(Vector2 p_pos) {
 			}
 
 			// Send Mouse Enter Self notification.
-			gui.mouse_over->notification(Control::NOTIFICATION_MOUSE_ENTER_SELF);
+			if (gui.mouse_over) {
+				gui.mouse_over->notification(Control::NOTIFICATION_MOUSE_ENTER_SELF);
+			}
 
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks for the quick fix!

Issue #85237, opened 2023-11-22
Freeze on ResourceSaver.save in custom importer when using multiple threads


thunderk:
### Godot version

4.2.rc1

### System information

Godot v4.2.rc1 - Manjaro Linux #1 SMP PREEMPT_DYNAMIC Thu Nov  9 04:27:49 UTC 2023 - X11 - Vulkan (Forward+) - integrated AMD Radeon Vega 8 Graphics (RADV RAVEN) () - AMD Ryzen 5 3550H with Radeon Vega Mobile Gfx (8 Threads)

### Issue description

Using a custom importer on a lot of files, with import threads enabled, sometimes causes Godot to fully freeze (including the UI), with no error in stdout/stderr.

Importing each file individually never freezes (and they all import correctly). Disabling importer threads seems to fully solve the problem.

Even setting a minimal project with only one custom importer and 10 files, it happens to freeze during import.

In this minimal project, it does not always happen, but randomly under certain conditions:

- `.godot/imported` is empty
- `*.TXTR.import` files do not exist yet
- The importer plugin is enabled in Project Settings for the first time (interestingly it happens less if it's already enabled in `project.godot`)

The UI freezes fully like this:

![image](https://github.com/godotengine/godot/assets/4595021/494ee2b4-b382-4ba5-b824-a41b9de8b44f)

Here is the minimal project:

[bug.tar.gz](https://github.com/godotengine/godot/files/13374254/bug.tar.gz)

The output before freeze (from print statements in the plugin):

```
Godot Engine v4.2.beta6.official.64150060f - https://godotengine.org
Vulkan API 1.3.257 - Forward+ - Using Vulkan Device #0: AMD - AMD Radeon Vega 8 Graphics
 
WARNING: Blend file import is enabled in the project settings, but no Blender path is configured in the editor settings. Blend files will not be imported.
     at: _editor_init (modules/gltf/register_types.cpp:63)
res://data/M000.TXTR -> res://.godot/imported/M000.TXTR-4bb0fc982e0e8da4bb882b98bceb0aad.res
res://data/M001.TXTR -> res://.godot/imported/M001.TXTR-fe74a5ef0315d8fdb6f9bdc09f3f6428.res
res://data/M003.TXTR -> res://.godot/imported/M003.TXTR-967cf8977babfd44b3b1c3d40e57708c.res
res://data/M005.TXTR -> res://.godot/imported/M005.TXTR-06dad9b17abe199d2512f9c3387587af.res
[...]

PR #85295, opened 2023-11-24
Use `SafeNumeric` to protect `max_index` in ImportThreadData

jsjtxietian:
When trying to debug #85237 , @TheDuriel and I found a issue in threaded reimport function as the title says. It still did not solve the problem because it's not related to the "correctness" of importing, it affects the progress bar UI. 

The real cause of #85237 still needs more investigation.

### DIFF
--- a/editor/editor_file_system.cpp
+++ b/editor/editor_file_system.cpp
@@ -2211,8 +2211,9 @@ void EditorFileSystem::reimport_file_with_custom_parameters(const String &p_file
 }
 
 void EditorFileSystem::_reimport_thread(uint32_t p_index, ImportThreadData *p_import_data) {
-	p_import_data->max_index = MAX(p_import_data->reimport_from + int(p_index), p_import_data->max_index);
-	_reimport_file(p_import_data->reimport_files[p_import_data->reimport_from + p_index].path);
+	int current_max = p_import_data->reimport_from + int(p_index);
+	p_import_data->max_index.exchange_if_greater(current_max);
+	_reimport_file(p_import_data->reimport_files[current_max].path);
 }
 
 void EditorFileSystem::reimport_files(const Vector<String> &p_files) {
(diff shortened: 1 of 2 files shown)


editor/editor_file_system.cpp
```diff
@@ -2211,8 +2211,11 @@ void EditorFileSystem::reimport_file_with_custom_parameters(const String &p_file
 }
 
 void EditorFileSystem::_reimport_thread(uint32_t p_index, ImportThreadData *p_import_data) {
-	p_import_data->max_index = MAX(p_import_data->reimport_from + int(p_index), p_import_data->max_index);
-	_reimport_file(p_import_data->reimport_files[p_import_data->reimport_from + p_index].path);
+	int current_max = p_import_data->reimport_from + int(p_index);
+	p_import_data->max_index_lock.lock();
+	p_import_data->max_index = MAX(current_max, p_import_data->max_index);
+	p_import_data->max_index_lock.unlock();
```
AThousandShips, line 2217:
```suggestion
	{
		MutexLock lock(p_import_data->max_index_lock);
		p_import_data->max_index = MAX(current_max, p_import_data->max_index);
	}
```
I'd suggest using a mutex lock, it's cleaner and also more secure

editor/editor_file_system.h
```diff
@@ -283,6 +283,7 @@ class EditorFileSystem : public Node {
 		const ImportFile *reimport_files;
 		int reimport_from;
 		int max_index = 0;
+		Mutex max_index_lock;
```
AThousandShips, line 286:
```suggestion
		Mutex max_index_mutex;
```
I'd name it `mutex` to avoid any confusion

editor/editor_file_system.cpp
jsjtxietian: Ah, at the first place I was looking for something smimilar to lock guard, did not realize Godot has one! Many thanks!

AThousandShips (approved):
The thread safety procedure itself looks good to me, not familiar with the larger structure so will need some more eyes on it

RandomShaper (dismissed):
I think this is a good use case for `SafeNumeric` and its `exchange_if_greater()` function.

API-Beast:
@RandomShaper
^ OP pushed the requested change.

lyuma (approved):
Looks good to me. This avoids a data race in the progress bar.

(and generally speaking, data races can lead to a wide range of undefined behavior which could be far outside the scope of just the progress bar, so it is good to fix this)

lyuma:
I would rather merge rather than the omnibus PR #88122 because it avoids getting bogged down by finding all the people to review bugfixes in three completely separate parts of the engine... and the StringName one risks a serious performance impact.
@RandomShaper if you would like to review the SafeNumeric part, but I think it's pretty straightforward

RandomShaper:
Being here again, after quite a long time, I realize this could be more accurate (and simpler), by, instead of keeping track of the highest index processed, just having an atomic counter and incrementing it by one from each thread.

The current approach is over-optimistic in that, if item 10 has started processing, 0-9 are assumed to have started, despite scheduling may have caused 10 to start before, say, 5.

Also, shouldn't the counter, either in this PR or in my suggestion, be incremented _after_ processing the item?

This PR already improves things, but giving this as food for thought and possible incremental enhancement.

akien-mga:
Thanks!

PR #85280, opened 2023-11-23
Prevent read-after-free in the queued CallableCustomStaticMethodPointer, fixes `slot >= slot_max` errors in release templates

bruvzg:
Fixes https://github.com/godotengine/godot/issues/85263 and fixes https://github.com/godotengine/godot/issues/70910

PR #85279, opened 2023-11-23
Enhance `SceneTree.change_scene*()` methods' docs

RandomShaper:
Fixes #85251.

Supersedes #85298.

PR #85298, opened 2023-11-24
Revert new scene change algorithm

### LINKED ISSUE #85251: get_tree() returning null immediately after calling change_scene_to_file() in _process or _physics_process()
### Godot version

v4.2.beta4.official [93cdacbb0]

### System information

Windows 10

### Issue description

`get_tree()` returns null when called immediately after calling `change_scene_to_file()` in `_process()` or `_physics_process()`. The error printed in debugger: `scene1.gd:14 @ _process(): Parameter "data.tree" is null.` Regardless of this the scene gets properly changed.

Doing the same from `_ready()` works as expected.

### Steps to reproduce

Make a scene with a node and another test scene to change to. Attach the following script to the node in the first scene:
~~~
extends Node
func _process(delta):
	if Input.is_action_just_pressed("ui_accept"):
		print(get_tree())
		get_tree().change_scene_to_file("res://scene2.tscn")
		print(get_tree())
~~~
Press enter and the output will be:
~~~
<SceneTree#25987908681>
<Object#null>
~~~

N/A

RandomShaper:
The more I think about it, the more it seems these changes didn't really improve anything by themselves. Much of the issues were of the kind solved now by #85280.

This reverts commit 8bebabda1868bb6d0b4f1b144f78c0d1cff8653c.
This reverts commit 4354cd8d41e044fba2dc5b63324388e01a786049.

Fixes #85251.
Supersedes #85279.

### DIFF
--- a/scene/main/scene_tree.cpp
+++ b/scene/main/scene_tree.cpp
@@ -515,10 +515,6 @@ bool SceneTree::process(double p_time) {
 
 	_flush_delete_queue();
 
-	if (unlikely(pending_new_scene)) {
-		_flush_scene_change();
-	}
-
 	process_timers(p_time, false); //go through timers
 
 	process_tweens(p_time, false);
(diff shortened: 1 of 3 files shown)


scene/main/scene_tree.cpp
```diff
@@ -1411,22 +1418,7 @@ Error SceneTree::change_scene_to_packed(const Ref<PackedScene> &p_scene) {
...
-
-	if (current_scene) {
-		// Let as many side effects as possible happen or be queued now,
-		// so they are run before the scene is actually deleted.
-		root->remove_child(current_scene);
-	}
-	DEV_ASSERT(!current_scene);
-
-	pending_new_scene = new_scene;
+	call_deferred(SNAME("_change_scene"), new_scene);
```
AThousandShips, line 1421: Does this have to be a named call?

RandomShaper:
After deeper discussion, the new algorithm also has strengths, so until we can find the best of both worlds, the new one is kept. There are very easy ways to emulate the former and the new one will be perfectly documented.

PR #85303, opened 2023-11-24
Detect proxy settings from environment variables

timothyqiu:
- Adds `load_default_proxies()` method to `HTTPClient` and `HTTPRequest`.
- Adds `auto_detect_proxy` property to `HTTPRequest`.
- Adds `network/http_proxy/auto_detect` setting to the editor.

### DIFF
--- a/core/io/http_client.cpp
+++ b/core/io/http_client.cpp
@@ -57,6 +57,10 @@ void HTTPClient::set_https_proxy(const String &p_host, int p_port) {
 	WARN_PRINT("HTTPS proxy feature is not available");
 }
 
+void HTTPClient::load_default_proxies() {
+	// Do nothing if not implemented.
+}
+
 Error HTTPClient::_request_raw(Method p_method, const String &p_url, const Vector<String> &p_headers, const Vector<uint8_t> &p_body) {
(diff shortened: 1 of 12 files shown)


timothyqiu:
These are de facto standard environment variables for proxies. For example:

* https://docs.python-requests.org/en/latest/user/advanced/#proxies
* https://github.com/curl/curl/blob/03cb1ff4d629d6110dab787c75c187626d58323d/lib/url.c#L2073-L2158

I think this could be a good start point of proxy detection. The API does not need to change if someone wants to detect Windows system proxy settings via a dedicated API.

Faless:
> These are de facto standard environment variables for proxies.

The problem with "de facto standards" is that everyone implements them in their own way, making them actually **not standard**. See here for some comparison across different programs, framework, languages: https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/

Faless:
Anyway even if we decide to implement reading the env variables, I would not expose the `load_default_proxies` method, and probably have the parsing in `IP`.

I'll think about it a bit and make a proposal for a slightly different API.

PR #85308, opened 2023-11-24
Fix a crash when trying to restore uncopyable animation tracks

### LINKED ISSUE #85307: Saving project crashes if animation mixer contains METHOD or ANIMATION tracks
### Godot version

v4.2.rc.custom_build [d6a1db2b0]

### System information

Godot v4.2.rc (d6a1db2b0) - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3090 Ti (NVIDIA; 31.0.15.4592) - AMD Ryzen 9 7950X 16-Core Processor (32 Threads)

### Issue description

`AnimatedValuesBackup::get_data()` can crash as the internal data map can have nullptr values if any of the tracks were METHOD or ANIMATION types, as `AnimatedValuesBackup::get_cache_copy()` defines that these types of tracks do not get valid copies.

The resulting crash is therefore a null pointer exception.

### Steps to reproduce

1. Open a project with these type of animation tracks.
2. Save the project.

https://github.com/godotengine/godot-demo-projects/tree/master/3d/platformer

YuriSizov:
Fixes https://github.com/godotengine/godot/issues/85307. A regression from https://github.com/godotengine/godot/pull/85266.

We don't really care about those tracks anyway, so we don't need to keep them in the backup. I made getting them from the backup an error, though, because that shouldn't be possible in the first place.

### DIFF
--- a/scene/animation/animation_mixer.cpp
+++ b/scene/animation/animation_mixer.cpp
@@ -2143,14 +2143,22 @@ void AnimatedValuesBackup::set_data(const HashMap<NodePath, AnimationMixer::Trac
 	clear_data();
 
 	for (const KeyValue<NodePath, AnimationMixer::TrackCache *> &E : p_data) {
-		data.insert(E.key, get_cache_copy(E.value));
+		AnimationMixer::TrackCache *track = get_cache_copy(E.value);
+		if (!track) {
+			continue; // Some types of tracks do not get a copy and must be ignored.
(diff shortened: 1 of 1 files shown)


akien-mga (approved):
Looks straightforward.

akien-mga:
Thanks!

PR #85311, opened 2023-11-24
Add property change guards to Sprite2D, Sprite3D and AnimatedSprite2D

### LINKED ISSUE #8510: Add property change guards to Sprite2D

My projects are 2D games.

A game developer told in a reddit thread (https://www.reddit.com/r/godot/comments/182gdgi/checking_equality_before_assignment_optimization/) how the frame rate increased considerably when a property change guards for Sprite2D were added to the games C# code.

I made a test own my own in GDScript (v4.2.rc1.mono.official [ad72de508]). The test creates 10000 sprites. Then I tested 4 different cases, where the sprites do different things in `_process`.

|   | debug fps  | release fps  |
|---|---|---|
| sprite doesn't do anything  |  73 |  105 |
| `flip_h = true` in every frame  |  35 |  48 |
| `frame = 0` in every frame  |  50 |  95 |
| `flip_h = true` and `frame = 0` in every frame  |  31 |  47 |

The cost of setting properties even without actually changing them seems to be huge.

My proposal is to add change guards to Sprite2D's properties. I am not very familiar to Godot's source code, but at least some Control use change guards extensively.

For example:

```gdscript
void Sprite2D::set_flip_v(bool p_flip) {
	vflip = p_flip;
	queue_redraw();
}
```
would be changed to 
```gdscript
void Sprite2D::set_flip_v(bool p_flip) {
	if (vflip == p_flip) {
		return
	}
	vflip = p_flip;
	queue_redraw();
}
```

Some properties, like region and frame, already check if the property actually changed, but they still do things like emitting signals:

```gdscript
void Sprite2D::set_frame(int p_frame) {
	ERR_FAIL_INDEX(p_frame, vframes * hframes);

	if (frame != p_frame) {
		item_rect_changed();
	}

	frame = p_frame;

	emit_signal(SceneStringNames::get_singleton()->frame_changed);
}
```
This would be changed to

```gdscript
void Sprite2D::set_frame(int p_frame) {
	ERR_FAIL_INDEX(p_frame, vframes * hframes);

	if (frame == p_frame) {
		return;		
	}

	item_rect_changed();
	frame = p_frame;
	emit_signal(SceneStringNames::get_singleton()->frame_changed);
}
```

If this proposal is approved, I can implement this.

This can be worked around with a few lines of code, but it will not. Just watch any Godot 2D tutorial video from YouTube and there will some variation of this:

```gdscript
sprite.flip_h = velocity.x > 0
```

I'm also _assuming_ that the performance cost of guarding a property change is higher when made with GDScript than when made in Sprite2D's setter.

So this enhancement would be beneficial in very many 2D games. Also, in my opinion it should be Node's responsibility to not to do unnecessary work if properties do not actually change.

Cannot be done with an add-on.

miv391:
Implements this proposal: https://github.com/godotengine/godot-proposals/issues/8510

If a property's setter is called with the property's current value, return immediately instead of doing anything costly.

My test creates 10000 Sprite2Ds.

```gdscript
func _ready():
	for y in 100:
		for x in 100:
			var s = SPRITE.instantiate()
			s.position = Vector2(x* 10, y* 10)
			add_child(s)
```

Each sprite has a script which does some combination of these simple setters (or just `pass`).

```gdscript
func _process(delta):
	pass
	flip_h = true
	frame = 0
```

My test version was built with Visual Studio and it's performance seems to be a bit lower than the official version. "% compared to pass" numbers are calculated from release version fps numbers.

|                                 | master, debug fps | master, release fps | % compared to pass | PR, debug fps | PR, release fps | % compared to pass |
|---------------------------------|------------------:|--------------------:|-------------------:|--------------:|----------------:|-------------------:|
| `pass`                          |                73 |                 105 |               100% |            74 |              98 |               100% |
| `flip_h = true`                 |                35 |                  48 |                46% |            32 |              92 |                94% |
| `frame = 0`                     |                50 |                  95 |                91% |            60 |              91 |                93% |
| `flip_h = true` and `frame = 0` |                31 |                  47 |                45% |            30 |              89 |                91% |

It may seem odd to just to set `flip_h = true`, but it is a very common practice to set sprite's direction in every frame like this:
```gdscript
	flip_h = velocity.x > 0
```
So there are a lot of 2D games which are setting `flip_h` in every frame without actually changing it.

_Production edit: Closes https://github.com/godotengine/godot-proposals/issues/8510_

### DIFF
--- a/scene/2d/sprite_2d.cpp
+++ b/scene/2d/sprite_2d.cpp
@@ -155,6 +155,10 @@ Ref<Texture2D> Sprite2D::get_texture() const {
 }
 
 void Sprite2D::set_centered(bool p_center) {
+	if (centered == p_center) {
+		return;
+	}
+
 	centered = p_center;
 	queue_redraw();
 	item_rect_changed();
@@ -165,6 +169,10 @@ bool Sprite2D::is_centered() const {
 }
 
 void Sprite2D::set_offset(const Point2 &p_offset) {
+	if (offset == p_offset) {
+		return;
+	}
+
 	offset = p_offset;
 	queue_redraw();
 	item_rect_changed();
@@ -175,6 +183,10 @@ Point2 Sprite2D::get_offset() const {
 }
 
 void Sprite2D::set_flip_h(bool p_flip) {
+	if (hflip == p_flip) {
+		return;
+	}
+
 	hflip = p_flip;
 	queue_redraw();
 }
@@ -184,6 +196,10 @@ bool Sprite2D::is_flipped_h() const {
 }
 
 void Sprite2D::set_flip_v(bool p_flip) {
+	if (vflip == p_flip) {
+		return;
+	}
+
 	vflip = p_flip;
 	queue_redraw();
 }
@@ -193,7 +209,7 @@ bool Sprite2D::is_flipped_v() const {
 }
 
 void Sprite2D::set_region_enabled(bool p_region_enabled) {
-	if (p_region_enabled == region_enabled) {
+	if (region_enabled == p_region_enabled) {
 		return;
 	}
 
@@ -223,6 +239,10 @@ Rect2 Sprite2D::get_region_rect() const {
 }
 
 void Sprite2D::set_region_filter_clip_enabled(bool p_region_filter_clip_enabled) {
+	if (region_filter_clip_enabled == p_region_filter_clip_enabled) {
+		return;
+	}
+
 	region_filter_clip_enabled = p_region_filter_clip_enabled;
 	queue_redraw();
 }
@@ -234,12 +254,12 @@ bool Sprite2D::is_region_filter_clip_enabled() const {
 void Sprite2D::set_frame(int p_frame) {
 	ERR_FAIL_INDEX(p_frame, vframes * hframes);
 
-	if (frame != p_frame) {
-		item_rect_changed();
+	if (frame == p_frame) {
+		return;
 	}
 
 	frame = p_frame;
-
+	item_rect_changed();
 	emit_signal(SceneStringNames::get_singleton()->frame_changed);
 }
 
@@ -260,6 +280,11 @@ Vector2i Sprite2D::get_frame_coords() const {
 
 void Sprite2D::set_vframes(int p_amount) {
 	ERR_FAIL_COND_MSG(p_amount < 1, "Amount of vframes cannot be smaller than 1.");
+
+	if (vframes == p_amount) {
+		return;
+	}
+
 	vframes = p_amount;
 	if (frame >= vframes * hframes) {
 		frame = 0;
@@ -275,6 +300,11 @@ int Sprite2D::get_vframes() const {
 
 void Sprite2D::set_hframes(int p_amount) {
 	ERR_FAIL_COND_MSG(p_amount < 1, "Amount of hframes cannot be smaller than 1.");
+
+	if (hframes == p_amount) {
+		return;
+	}
+
 	if (vframes > 1) {
 		// Adjust the frame to fit new sheet dimensions.
 		int original_column = frame % hframes;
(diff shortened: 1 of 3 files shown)


arkology:
I have one really stupid style question. Suggested code follows pattern `if (class_variable == func_arg)`, so it checks if current sprite property is same as newly set. How about to check opposite, if newly set is same as sprite property - so we focus on function argument, not sprite property? Of course this is absolutely the same in terms of functionality but I think it has slight logical difference.
Is there any style recommendations about it in Godot codebase or it doesn't actually make any sense to care about? :)

(And of course I support this guards addition, without it some users could have redraw spam without even knowing about it.)

KoBeWi:
I checked some classes and we aren't ordering this consistently.

miv391:
I used the same ordering I have previously seen in some Control nodes. At least I tried to use the same ordering in every function :)

akien-mga:
Yeah the order used here is the most common in my experience, and the most readable IMO.

kleonc:
Could be also added to `SpriteBase3D::set_centered` etc. (in `sprite_3d.cpp`).

AThousandShips:
Also unify code in `Sprite3D::set_region_rect`

miv391:
> Could be also added to `SpriteBase3D::set_centered` etc. (in `sprite_3d.cpp`).

Done.

> Also unify code in `Sprite3D::set_region_rect`, and the different syntax in `SpriteBase3D::set_alpha_hash_scale` etc.

Done.

I _didn't_ add a change guard to `set_modulate`: https://github.com/godotengine/godot/blob/84692c6252438393999c427bf7f5d35b9c43926c/scene/3d/sprite_3d.cpp#L324-L328 

That function propagates the color to child nodes and I have no idea what would happen if you add new child nodes between setting modulation color like this:
```gdscript
sprite3d.modulate = Color.RED
sprite3d.add_child(some_node)
sprite3d.modulate = Color.RED
```
At least now without the change guard the color will be propagated to the child node.

YuriSizov:
> That function propagates the color to child nodes and I have no idea what would happen if you add new child nodes between setting modulation color like this:

It doesn't actually propagate any values, it just triggers a redraw (which makes `queue_redraw` in this method redundant btw). It's safe to add a guard there, newly added children would of course be drawn on their own.

scene/2d/sprite_2d.cpp
```diff
@@ -234,12 +254,12 @@ bool Sprite2D::is_region_filter_clip_enabled() const {
...
 	ERR_FAIL_INDEX(p_frame, vframes * hframes);
 
-	if (frame != p_frame) {
-		item_rect_changed();
+	if (frame == p_frame) {
+		return;
 	}
 
+	item_rect_changed();
 	frame = p_frame;
```
YuriSizov, line 262: It probably makes more sense to trigger it after the assignment. Old code here was weird.

scene/3d/sprite_3d.cpp
```diff
@@ -767,11 +822,12 @@ bool Sprite3D::is_region_enabled() const {
...
-	bool changed = region_rect != p_region_rect;
-	region_rect = p_region_rect;
-	if (region && changed) {
-		_queue_redraw();
+	if (region_rect == p_region_rect) {
+		return;
 	}
+
+	region_rect = p_region_rect;
+	_queue_redraw();
```
YuriSizov, line 830: We can still keep the optimization and avoid redrawing if `region` is not enabled.

YuriSizov (approved):
Other than the above, seems good to me. Feel free to do the adjustments, and then it can be merged.

miv391:
Adjustments done.

YuriSizov:
Good work, needs a rebase after https://github.com/godotengine/godot/pull/85317. I merged it first since it fixes a problem and is generally more involved.

YuriSizov:
Thanks!

PR #80409, opened 2023-08-08
[DisplayServer] Add method to estimate window title bar size.

bruvzg:
Adds a `DisplayServer.window_get_title_size(String title, WindowID window)` method to estimate window title size (size of the window control buttons and title string using system font).

Implemented on macOS and Windows.

Fixes https://github.com/godotengine/godot/issues/80310

PR #85312, opened 2023-11-24
Make sure `Window`'s title is respected before we compute the size

### LINKED ISSUE #85118: Install Project popup's child controls reset their layout after changing project path, but the popup's size doesn't
### Godot version

v4.2.rc1.official [ad72de508]

### System information

Godot v4.2.rc1 - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2080 Ti (NVIDIA; 30.0.14.7280) - Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz (12 Threads)

### Issue description

When installing anything with a long name from the Asset Library Projects, the Install Project popup is wider than normal to fit the name comfortably in the popup. 

Godot-bugreport-1

However, when you browse into a different folder, it resets to the 4.1.x layout while keeping the larger window size.

Godot-bugreport-2

This results in ghosts of the wider design still being visible to the right of the actual buttons. This is only a visual glitch and does not break the functionality of the popup, and the ghosts of the buttons to the right does not react to mouse input.

### Steps to reproduce

1. Open up Godot 4.2-rc1
2. Switch to Asset Library Projects
3. Search for "Shape Manipulation"
4. Click the asset "Shape Manipulation Basics - Line2D…"
5. Click download
6. Wait for download
7. Note the wide layout of the UI
8. Click Browse
9. Click Select Current Folder, you do not need to change folder to test this

Godot-bugreport-2

Note that the layout is now narrower, but has the ghosts of the wider layout still showing.
The ghost elements are just a visual glitch and cannot be interacted with.
Besides the visual glitch the popup still functions as normal.

This was confirmed to not happen in version v4.1.1.stable.mono.official [bd6af8e0e], in this version the layout is never widened when picking an asset with a long name.

None possible, the bug is in the Project Manager

YuriSizov:
Fixes https://github.com/godotengine/godot/issues/85118.

Like the title says, we used to first apply the size limit, and then we would update said limit without it affecting the actual size. So the window would be displayed with one size, and window's contents would refer to another size. Good old DisplayServer/Window de-sync.

It may not be Windows specific, as far as OSes go, highly depends on how window management is implemented on each platform. But since we are only reproducing it on Windows for now, I'm marking it as such. Also I guess we can say it's a regression because it's been introduced with the feature to keep the title visible in 4.2: https://github.com/godotengine/godot/pull/80409.

----

Also removes some suspicious and outdated code that forced this particular dialog to change size when the warning message changed.

### DIFF
--- a/editor/project_manager.cpp
+++ b/editor/project_manager.cpp
@@ -101,12 +101,6 @@ void ProjectDialog::_set_message(const String &p_msg, MessageType p_type, InputT
 	} else if (current_install_icon != new_icon && input_type == INSTALL_PATH) {
 		install_status_rect->set_texture(new_icon);
 	}
-
-	Size2i window_size = get_size();
-	Size2 contents_min_size = get_contents_minimum_size();
-	if (window_size.x < contents_min_size.x || window_size.y < contents_min_size.y) {
-		set_size(window_size.max(contents_min_size));
-	}
 }
 
 String ProjectDialog::_test_path() {
@@ -868,6 +862,7 @@ ProjectDialog::ProjectDialog() {
 
 	msg = memnew(Label);
 	msg->set_horizontal_alignment(HORIZONTAL_ALIGNMENT_CENTER);
+	msg->set_custom_minimum_size(Size2(200, 0) * EDSCALE);
 	vb->add_child(msg);
 
 	// Renderer selection.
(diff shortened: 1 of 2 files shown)


bruvzg (approved):
Title update change seems fine.

akien-mga:
Thanks!

PR #85313, opened 2023-11-24
Fix crash on hiding grandparent Control on mouse exit

### LINKED ISSUE #85277: Crash when hiding a panel from a `NOTIFICATION_MOUSE_ENTER` notification

kitbdev:
- fixes https://github.com/godotengine/godot/issues/85277#issuecomment-1825016577
- related https://github.com/godotengine/godot/pull/85284

fixes crash when changing top level or mouse filter to stop during a mouse notification (https://github.com/godotengine/godot/issues/85277#issuecomment-1825803638).

This works by deferring sending new notifications if we are already sending notifications.
Mouse enter and exit notifications are now handled when changing visibility, top level, and mouse filter during the notification callbacks. Previously it would sending duplicate or out of order notifications.

~~However, with this there is a way to make an infinite loop:
On mouse_entered set the child to mouse filter stop, and on mouse_exited set the child to mouse filter pass. Then mouse over the child. The parent will continuously receive mouse enter and exit notifications.
Previously, it would just send mouse notifications in the wrong order and then stop. This also won't happen with hide() and show() since show() doesn't update the mouse over immediately.~~

### DIFF
--- a/scene/main/viewport.cpp
+++ b/scene/main/viewport.cpp
@@ -2451,6 +2451,14 @@ void Viewport::_gui_update_mouse_over() {
 		return;
 	}
 
+	if (gui.sending_mouse_enter_exit_notifications) {
+		// If notifications are already being sent, delay call to next frame.
+		if (get_tree() && !get_tree()->is_connected(SNAME("process_frame"), callable_mp(this, &Viewport::_gui_update_mouse_over))) {
+			get_tree()->connect(SNAME("process_frame"), callable_mp(this, &Viewport::_gui_update_mouse_over), CONNECT_ONE_SHOT);
+		}
+		return;
+	}
+
 	// Rebuild the mouse over hierarchy.
 	LocalVector<Control *> new_mouse_over_hierarchy;
 	LocalVector<Control *> needs_enter;
@@ -2507,6 +2515,8 @@ void Viewport::_gui_update_mouse_over() {
 		return;
 	}
 
+	gui.sending_mouse_enter_exit_notifications = true;
+
 	// Send Mouse Exit Self notification.
 	if (gui.mouse_over && !needs_exit.is_empty() && needs_exit[0] == (int)gui.mouse_over_hierarchy.size() - 1) {
 		gui.mouse_over->notification(Control::NOTIFICATION_MOUSE_EXIT_SELF);
@@ -2528,6 +2538,8 @@ void Viewport::_gui_update_mouse_over() {
 	for (int i = needs_enter.size() - 1; i >= 0; i--) {
 		needs_enter[i]->notification(Control::NOTIFICATION_MOUSE_ENTER);
 	}
+
+	gui.sending_mouse_enter_exit_notifications = false;
 }
 
 Window *Viewport::get_base_window() const {
@@ -3200,16 +3212,20 @@ void Viewport::_update_mouse_over(Vector2 p_pos) {
 			gui.mouse_over = over;
 			gui.mouse_over_hierarchy.reserve(gui.mouse_over_hierarchy.size() + over_ancestors.size());
 
+			gui.sending_mouse_enter_exit_notifications = true;
+
 			// Send Mouse Enter notifications to parents first.
 			for (int i = over_ancestors.size() - 1; i >= 0; i--) {
-				over_ancestors[i]->notification(Control::NOTIFICATION_MOUSE_ENTER);
 				gui.mouse_over_hierarchy.push_back(over_ancestors[i]);
+				over_ancestors[i]->notification(Control::NOTIFICATION_MOUSE_ENTER);
 			}
 
 			// Send Mouse Enter Self notification.
 			if (gui.mouse_over) {
 				gui.mouse_over->notification(Control::NOTIFICATION_MOUSE_ENTER_SELF);
 			}
+
+			gui.sending_mouse_enter_exit_notifications = false;
 
 			notify_embedded_viewports = true;
 		}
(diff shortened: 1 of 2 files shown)


scene/main/viewport.cpp
```diff
@@ -2507,6 +2513,8 @@ void Viewport::_gui_update_mouse_over() {
 		return;
 	}
 
+	gui.sending_mouse_enter_exit_notifications = true;
```
Sauermann, line 2516: In the implementation you keep track in every Viewport separately, if `sending_mouse_enter_exit_notifications` is active or not.
Have you considered the idea to keep track of the state in a global manner (for example only in the root-Viewport)? I'm not entirely sure, if there can be edge cases, where these notifications in different Viewports could interfere with each other.

However, for now it looks like a reasonable approach.

scene/main/viewport.cpp
```diff
@@ -2451,6 +2451,12 @@ void Viewport::_gui_update_mouse_over() {
 		return;
 	}
 
+	if (gui.sending_mouse_enter_exit_notifications) {
+		// If notifications are already being sent, defer call.
+		callable_mp(this, &Viewport::_gui_update_mouse_over).call_deferred();
```
Sauermann, line 2456: In the linked issue, you mention two approaches:
- prevent duplicate update
- defer duplicate update

Are there specific use-cases, that benefit from the deferred call? If there are none, then I would suggest to prevent it.
I understand, that deferring it, might be more robust.

Sauermann (commented):
I have tested and can confirm, that the PR fixes the MRP in https://github.com/godotengine/godot/issues/85277#issuecomment-1825016577 and https://github.com/godotengine/godot/issues/85277#issuecomment-1825803638.

scene/main/viewport.cpp
Sauermann: So according to your comment, preventing the duplicated update is not a valid approach. Thanks!

scene/main/viewport.cpp
```diff
@@ -2451,6 +2451,14 @@ void Viewport::_gui_update_mouse_over() {
 		return;
 	}
 
+	if (gui.sending_mouse_enter_exit_notifications) {
+		// If notifications are already being sent, delay call to next frame.
+		if (get_tree() && !get_tree()->is_connected("process_frame", callable_mp(this, &Viewport::_gui_update_mouse_over))) {
+			get_tree()->connect("process_frame", callable_mp(this, &Viewport::_gui_update_mouse_over), CONNECT_ONE_SHOT);
```
Sauermann, line 2457: Is is possible, that in a single Viewport this function could be connected to `process_frame` multiple times, so that the function is called more than once in the next frame?

kitbdev: For this, I mostly copied what `editor/plugins/animation_player_editor_plugin.cpp:1707` did to connect the signal.
I check it's not already connected to make sure it won't be connected multiple times.
If the user makes multiple changes to top level or mouse filters, then this function would only need to be called once afterward.

Sauermann: Sorry, I missed the check.

Sauermann (approved):
I have tested successfully the MRPs. Code is looking good to me and I don't see any obvious problems.

scene/main/viewport.cpp
```diff
@@ -2451,6 +2451,14 @@ void Viewport::_gui_update_mouse_over() {
 		return;
 	}
 
+	if (gui.sending_mouse_enter_exit_notifications) {
+		// If notifications are already being sent, delay call to next frame.
+		if (get_tree() && !get_tree()->is_connected("process_frame", callable_mp(this, &Viewport::_gui_update_mouse_over))) {
```
AThousandShips, line 2456: This looks like it's updated relatively often, I'd suggest using `SNAME` to avoid bottleneck with `StringName` construction, like so:
```suggestion
		if (get_tree() && !get_tree()->is_connected(SNAME("process_frame"), callable_mp(this, &Viewport::_gui_update_mouse_over))) {
```

scene/main/viewport.cpp
AThousandShips, line 2457:
```suggestion
			get_tree()->connect(SNAME("process_frame"), callable_mp(this, &Viewport::_gui_update_mouse_over), CONNECT_ONE_SHOT);
```

scene/main/viewport.cpp
```diff
@@ -3252,6 +3268,12 @@ void Viewport::_mouse_leave_viewport() {
 }
 
 void Viewport::_drop_mouse_over(Control *p_until_control) {
+	if (gui.sending_mouse_enter_exit_notifications) {
+		// If notifications are already being sent, defer call.
+		callable_mp(this, &Viewport::_drop_mouse_over).call_deferred(p_until_control);
+		return;
+	}
```
akien-mga, line 3275: I'm not familiar with this code, but slightly worried at the possibility of a deferred infinite recursion here. Are we absolutely sure that this won't end up deferring itself over and over again without ever reaching a condition that would make `gui.sending_mouse_enter_exit_notifications` false?

kitbdev: Yes, since every time `gui.sending_mouse_enter_exit_notifications` is set to true, it is immediately set to false in the same scope after it is done sending a fixed amount of notifications.
I'm not sure if there is a way to tell if this call was already deferred, but I could add a check just in case something changes and it is set to true?

Also this method only removes Controls from a list and cannot add any, so it cannot keep calling itself infinitely from user code, like was previously possible in `_gui_update_mouse_over`.

YuriSizov:
Cherry-picked for 4.2.2.

PR #85317, opened 2023-11-24
Fix invalid `frame` index when Sprite2D's `hframes` or `vframes` change

miv391:
This is my test sprite sheet:

kuva

In master this code causes an error as expected:

```gdscript
	sprite_using_sheet.hframes = 2
	sprite_using_sheet.vframes = 2

	sprite_using_sheet.frame = 4 
# --> main.gd:15 @ test2(): Index p_frame = 4 is out of bounds (vframes * hframes = 4).
```

But this is fine:

```gdscript
	sprite_using_sheet.hframes = 2
	sprite_using_sheet.vframes = 2 # now there are frames 0-3
	
	sprite_using_sheet.frame = 3
	
	sprite_using_sheet.hframes = 1 # now there are frames 0-1, but frame is still 3 
```

Now the sprite looks very odd. The size and shape match to the spritesheet's current frame size and shape, but the content is garbage.

kuva

If the same is done in editor's property Inspector, the result is more or less the same. Editor however changes the `frame` property to 1, but still the sprite looks wrong.

With this PR the sprite looks like this when the code above is run:

kuva

## How `vframes` and `hframes` changes can affect `frame`

What cases this PR handles and how. The goal is to make sure that the frame value keeps valid (must be smaller than `hframes * vframes`) and if possible, to keep `frame` referencing the same row and column in the texture sheet even if `vframes` or `hframes` changes.

### `vframes` changes 

```
         (a)       (b)

OOO  ->  OOO  or  OOO
OOO      OOO      OOO
OOO      OOO
         OOO
```

(a) `frame`'s validity or value cannot change.
(b) `frame` can become invalid, if its row is dropped. `frame`'s value cannot change.

### `hframes` changes

Special case where `vframes` is 1:

```
          (c)          (d)

OOOO  ->  OOOOOOO  or  OO
```

(c) `frame`'s validity or value cannot change.
(d) `frame` can become invalid, if its column is dropped. `frame`'s value cannot change.

When `vframes` > 1:
```
          (e)          (f)

OOOO  ->  OOOOOOO  or  OO
OOOO      OOOOOOO      OO 
OOOO      OOOOOOO      OO
```

(e) `frame` cannot become invalid, but its value can change.
(f) `frame` can become invalid if its column is dropped, also `frame`'s value can change.
      

- Cases (a) and (c) require no handling.
- Cases (b) and (d) are handled simply by checking `if frame <= hframes * vframes`. If not, `frame` is reset to 0.
- Only the cases (e) and (f) need more special handling. If the frame was located in a column that was dropped, reset `frame` to 0. Otherwise calculate original row and column and then calculate new frame value using new `hframes` values. Finally the `if frame <= hframes * vframes` check is done.

### DIFF
--- a/doc/classes/Sprite3D.xml
+++ b/doc/classes/Sprite3D.xml
@@ -10,13 +10,13 @@
 	</tutorials>
 	<members>
 		<member name="frame" type="int" setter="set_frame" getter="get_frame" default="0">
-			Current frame to display from sprite sheet. [member hframes] or [member vframes] must be greater than 1.
+			Current frame to display from sprite sheet. [member hframes] or [member vframes] must be greater than 1. This property is automatically adjusted when [member hframes] or [member vframes] are changed to keep pointing to the same visual frame (same column and row). If that's impossible, this value is reset to [code]0[/code].
 		</member>
 		<member name="frame_coords" type="Vector2i" setter="set_frame_coords" getter="get_frame_coords" default="Vector2i(0, 0)">
 			Coordinates of the frame to display from sprite sheet. This is as an alias for the [member frame] property. [member hframes] or [member vframes] must be greater than 1.
 		</member>
 		<member name="hframes" type="int" setter="set_hframes" getter="get_hframes" default="1">
-			The number of columns in the sprite sheet.
+			The number of columns in the sprite sheet. When this property is changed, [member frame] is adjusted so that the same visual frame is maintained (same row and column). If that's impossible, [member frame] is reset to [code]0[/code].
 		</member>
 		<member name="region_enabled" type="bool" setter="set_region_enabled" getter="is_region_enabled" default="false">
 			If [code]true[/code], the sprite will use [member region_rect] and display only the specified part of its texture.
@@ -28,7 +28,7 @@
 			[Texture2D] object to draw. If [member GeometryInstance3D.material_override] is used, this will be overridden. The size information is still used.
 		</member>
 		<member name="vframes" type="int" setter="set_vframes" getter="get_vframes" default="1">
-			The number of rows in the sprite sheet.
+			The number of rows in the sprite sheet. When this property is changed, [member frame] is adjusted so that the same visual frame is maintained (same row and column). If that's impossible, [member frame] is reset to [code]0[/code].
 		</member>
 	</members>
 	<signals>
--- a/scene/2d/sprite_2d.cpp
+++ b/scene/2d/sprite_2d.cpp
@@ -261,6 +261,9 @@ Vector2i Sprite2D::get_frame_coords() const {
 void Sprite2D::set_vframes(int p_amount) {
 	ERR_FAIL_COND_MSG(p_amount < 1, "Amount of vframes cannot be smaller than 1.");
 	vframes = p_amount;
+	if (frame >= vframes * hframes) {
+		frame = 0;
+	}
 	queue_redraw();
 	item_rect_changed();
 	notify_property_list_changed();
@@ -272,7 +275,21 @@ int Sprite2D::get_vframes() const {
 
 void Sprite2D::set_hframes(int p_amount) {
 	ERR_FAIL_COND_MSG(p_amount < 1, "Amount of hframes cannot be smaller than 1.");
+	if (vframes > 1) {
+		// Adjust the frame to fit new sheet dimensions.
+		int original_column = frame % hframes;
+		if (original_column >= p_amount) {
+			// Frame's column was dropped, reset.
+			frame = 0;
+		} else {
+			int original_row = frame / hframes;
+			frame = original_row * p_amount + original_column;
+		}
+	}
 	hframes = p_amount;
+	if (frame >= vframes * hframes) {
+		frame = 0;
+	}
 	queue_redraw();
 	item_rect_changed();
 	notify_property_list_changed();
--- a/scene/3d/sprite_3d.cpp
+++ b/scene/3d/sprite_3d.cpp
@@ -804,8 +804,11 @@ Vector2i Sprite3D::get_frame_coords() const {
 }
 
 void Sprite3D::set_vframes(int p_amount) {
-	ERR_FAIL_COND(p_amount < 1);
+	ERR_FAIL_COND_MSG(p_amount < 1, "Amount of vframes cannot be smaller than 1.");
 	vframes = p_amount;
+	if (frame >= vframes * hframes) {
+		frame = 0;
+	}
 	_queue_redraw();
 	notify_property_list_changed();
 }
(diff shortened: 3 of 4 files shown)


groud:
I think the fix is needed, but I would probably set it to the highest value instead.
Something like this: 
`frame = MIN(frame, (vframes * hframes) - 1)`

miv391:
> I think the fix is needed, but I would probably set it to the highest value instead.

I though about that too. But I'm guesssing that in most use cases it just doesn't matter which new valid value is chosen, it is not exactly right for the user anyway. Let's say the user has a sprite sheet of 4 rows x 6 columns. Sprite currently uses the last frame (23). Then user decides that the last row is no more needed and changes the sheet to 3 rows x 6 columns. Most likely the new last frame (17) isn't any better choice for the sprite than the first frame. In my opinion the first frame is just a more intuitive default value when such a value is needed. My other guess is that most often this bug happens when user is editing sprites and sprite sheets in editor so the user immediately sees what's happening. Frame just resetting to 0 is maybe less startling than frame jumping from 23 to 17.

akien-mga:
Makes sense to me to handle this case indeed.

Regarding what `frame` should be set to, it's a difficult question. I agree with @miv391's example that in some circumstances just clamping to the last frame may not yield any useful/predictable result.

But that made me realize that this is a problem any time `hframes` or `vframes` is changed, even if the change doesn't make `frame` go out of bounds. Changing the h and v frame count will change the meaning of `frame` anyway.

So maybe we should always reset `frame` to 0 when either `hframes` or `vframes` is changed? And this should be documented in the description of `hframes` and `vframes`.

miv391:
> So maybe we should always reset `frame` to 0 when either `hframes` or `vframes` is changed? And this should be documented in the description of `hframes` and `vframes`.

I think always resetting `frame` to 0 would be overkill. There are afterall cases where changes do not affect `frame`.  For example, you could just add more rows to the texture. `vframes` changes, but `frame` still references the same rect as before. Or you have all your frames in a long row and you add one more frame. Now `hframes` changes, but `frame` still references the same rect as before. In the pictures below I'm adding new rows/columns, but removing rows/columns would work in the same way, `frame` could become invalid, but if it stays valid, its rect stays the same.

(Red box represents the direction to where the user is expanding the texture.)

kuva

Only changing `hframes` when `vframes` > 1 would cause problems:

kuva

> Alternatively, we could try to calculate the would be `frame` number for a given `(x, y)` frame before and after the `hframes`/`vframes` change, to change `frame` so that it still points at the same rect - provided it's still in the region, otherwise it would indeed have to be set to 0 or clamped.

This would likely be the most elegant solution.

miv391:
I changed how `set_vframes` works and also updated the PR's description.

kleonc:
- `Sprite3D`'s logic needs to be kept in sync.
- Guards for early return could be added to the setters, like `if (vframes == p_amount) { return; }` (including in `set_frame`).

miv391:
> * `Sprite3D`'s logic needs to be kept in sync.

I added same logic to `Sprite3D`.

> * Guards for early return could be added to the setters, like `if (vframes == p_amount) { return; }` (including in `set_frame`).

A separate PR exists for adding property change guards: https://github.com/godotengine/godot/pull/85311

doc/classes/Sprite3D.xml
```diff
@@ -10,13 +10,13 @@
 	</tutorials>
 	<members>
 		<member name="frame" type="int" setter="set_frame" getter="get_frame" default="0">
-			Current frame to display from sprite sheet. [member hframes] or [member vframes] must be greater than 1.
+			Current frame to display from sprite sheet. [member hframes] or [member vframes] must be greater than 1. This can be automatically adjusted, when [member hframes] or [member vframes] are changed.
```
YuriSizov, line 13:
```suggestion
			Current frame to display from sprite sheet. [member hframes] or [member vframes] must be greater than 1. This property is automatically adjusted when [member hframes] or [member vframes] are changed to keep pointing to the same visual frame (same column and row). If that's impossible, this value is reset to [code]0[/code].
```

doc/classes/Sprite3D.xml
```diff
@@ -10,13 +10,13 @@
...
 		</member>
 		<member name="frame_coords" type="Vector2i" setter="set_frame_coords" getter="get_frame_coords" default="Vector2i(0, 0)">
 			Coordinates of the frame to display from sprite sheet. This is as an alias for the [member frame] property. [member hframes] or [member vframes] must be greater than 1.
 		</member>
 		<member name="hframes" type="int" setter="set_hframes" getter="get_hframes" default="1">
-			The number of columns in the sprite sheet.
+			The number of columns in the sprite sheet. When this is changed, [member frame] is recalculated so that the frame's row and column are maintained. If the original row and column do not exist anymore, [member frame] is reset to 0.
```
YuriSizov, line 19:
```suggestion
			The number of columns in the sprite sheet. When this property is changed, [member frame] is adjusted so that the same visual frame is maintained (same row and column). If that's impossible, [member frame] is reset to [code]0[/code].
```

doc/classes/Sprite3D.xml
```diff
@@ -28,7 +28,7 @@
 			[Texture2D] object to draw. If [member GeometryInstance3D.material_override] is used, this will be overridden. The size information is still used.
 		</member>
 		<member name="vframes" type="int" setter="set_vframes" getter="get_vframes" default="1">
-			The number of rows in the sprite sheet.
+			The number of rows in the sprite sheet. When this is changed, [member frame] is reset to 0, if the frame's row does not exist anymore.
```
YuriSizov, line 31:
```suggestion
			The number of rows in the sprite sheet. When this property is changed, [member frame] is adjusted so that the same visual frame is maintained (same row and column). If that's impossible, [member frame] is reset to [code]0[/code].
```

scene/2d/sprite_2d.cpp
```diff
@@ -272,7 +275,21 @@ int Sprite2D::get_vframes() const {
 
 void Sprite2D::set_hframes(int p_amount) {
 	ERR_FAIL_COND_MSG(p_amount < 1, "Amount of hframes cannot be smaller than 1.");
+	if (vframes > 1) {
+		// recalculate frame
```
YuriSizov, line 279:
```suggestion
		// Adjust the frame to fit new sheet dimensions.
```

The rephrasing is secondary, but comment needs to start with a capital letter and end with a period.

scene/2d/sprite_2d.cpp
```diff
@@ -272,7 +275,21 @@ int Sprite2D::get_vframes() const {
...
+		int original_column = frame % hframes;
+		if (original_column >= p_amount) {
+			// frame's column was dropped
```
YuriSizov, line 282:
```suggestion
			// Frame's column was dropped, reset.
```

scene/3d/sprite_3d.cpp
```diff
@@ -815,8 +818,22 @@ int Sprite3D::get_vframes() const {
 }
 
 void Sprite3D::set_hframes(int p_amount) {
-	ERR_FAIL_COND(p_amount < 1);
+	ERR_FAIL_COND_MSG(p_amount < 1, "Amount of hframes cannot be smaller than 1.");
+	if (vframes > 1) {
+		// recalculate frame
+		int original_column = frame % hframes;
+		if (original_column >= p_amount) {
+			// frame's column was dropped
```
YuriSizov, line 826: Same here.

YuriSizov (approved):
Seems good to me, but needs adjustments to the docs and comments.

miv391:
> Seems good to me, but needs adjustments to the docs and comments.

Adjustments done.

YuriSizov:
Thanks!

PR #85331, opened 2023-11-24
Add full customization of 3D navigation controls

### LINKED ISSUE #3646: Allow two modifier keys for mouse navigation in the 3D editor viewport

Porting a game from Unity to Godot and needing to use the same viewport navigation because muscle memory and easy of use.

Unity let's me pan the viewport with ALT+CMD+LEFT MOUSE
Godot is limited to only allow one Modifier key in the option panel

FC5IxhTXEAsqd5B

When I enable 3 button mouse it already pans when I press ALT and move the mouse ( without holding left mouse ) I would like it to Rotate with left mouse pressed, and pan when ALT+CMD+Left mouse is pressed. And do nothig with just ALT pressed when moving the mouse.

Allow two modifier keys to be selected and used in the project setting menu

Allowing two modifier keys to be selected in the options

two

It will be used by everyone who likes the Unity style navigation and want to use Godot without having to relearn navigation

As this solves a viewport navigation limitation this will benefit a lot of users and is not limited to a single project use case.

### LINKED ISSUE #6406: Allow full customization of 3D navigation controls (rotate, pan, zoom)

Every project.

Godot doesn't provide a way to reproduce the desired 3d navigation control scheme.

Ability to define which key/mouse/key+mouse bindings are associated with each 3d navigation control (rotate,pan,zoom).
By allowing this, users will be able to reproduce any control scheme they need that godot currently doesn't offer. Godot only offer a few presets which is limited and doesn't offer ones that users might need. So being able to customize it would allow users to achieve the control schemes they need.

For each type of navigation (rotate/pan/zoom), select whether it is a single key, multiple keys, key+mouse button, or just mouse button.
Allow choice of keys, and allow choice of mouse button (mouse buttons 4 and 5 included).

There can still be presets to choose from(for popular schemes), but choosing a preset would just auto-fill these inputs.

No- currently, the only way is to modify godot's source code and build godot from source (I'm currently doing that to implement my control scheme).

Improving the useability of the program should be a priority or at least an important issue and having it in the core would be a positive change for the program.

RobProductions:
Implements this proposal: godotengine/godot-proposals#6406 which in turn satisfies godotengine/godot-proposals#3646 as well.

*Bugsquad edit:*
- Closes https://github.com/godotengine/godot-proposals/issues/6406
- Closes https://github.com/godotengine/godot-proposals/issues/3646

The goal is to allow users to freely customize which mouse buttons/keys navigate the 3D viewport in the editor. Previously, the spatial editor controls were limited to hardcoded button checks based on the `navigation_scheme` setting. With this change, `navigation_scheme` instead controls 3 new mouse button settings and has a new `CUSTOM` setting which allows users to adjust the mouse buttons however they want:

https://github.com/godotengine/godot/assets/11141862/283d48c0-98c8-4698-8ba5-0a7cdb2e7e13

In order to properly set up user settings (in case they have an old version of Godot with "Modo" or "Maya" selected), I figured it was best to inject the preset detection call into `editor_node.cpp` so that it gets checked when the editor opens. It also protects against users manually editing their `editor_settings-4.tres` file to remove mouse settings for example, as that would be confusing to view in the settings dialog. If we don't care about these cases/have a different way to approach it feel free to let me know and I can adjust it. 

To tackle modifiers, the previous orbit, pan, and zoom settings were moved into `Shortcuts` instead because of the existing flexibility with that system to add multiple keys/any key desired:

image

I wanted to use just one modifier for each operation, but due to a bug (or intended behavior?) in InputEventKey modifiers, pressing a modifier with it's "main" key and then releasing the modifier will still count the button as being pressed. [This Reddit post](https://www.reddit.com/r/godot/comments/xf8354/problem_with_using_modifiers_with_inputeventkey/) explains it best I think. Another benefit of 2 modifiers is being able to set 2 different "standard" keys like Z/X as opposed to being limited to CTRL, ALT, etc. Like the mouse settings, these will be set by the `navigation_scheme` preset and overriding them will swap the scheme to Custom.

With these new settings in place, the user retains all the same navigation options as before and can easily access them through the preset, but they also acquire a bunch of new ways to customize the controls to their liking:

https://github.com/godotengine/godot/assets/11141862/606d8937-c666-49d0-98b8-177bfa62c76f

I've also run the doc generation tool and filled out the descriptions for the new mouse settings, as well as updating the `navigation_scheme` description to clearly state the new behavior. It seems like there aren't any unit tests for the editor yet but I made sure to run them locally just in case. I've checked tablet navigation on my Wacom as well, and ensured the `Emulate 3-Button Mouse` setting still works when the shortcuts are set properly.

Due to code ordering complications, I made the assumption that there are less orbit modifiers than pan modifiers for any usage of the same mouse button, otherwise one of the modes could be skipped over. This is true for all of the presets previously available, so it will not cause any regression; however if the user adjusts orbit modifiers in such a way that they should be detected first, it might get skipped. In the interest of creating a smaller PR to review, I've left it at that for now hoping the ordering will be suitable for the majority of use cases. In the future I can look into creating another PR that checks the modes in a more correct order based on how many shortcut keys are set for each op, though it will require a few more helper functions.

Since this makes the navigation code more generic, it becomes a lot easier to add in more presets. If this is approved, I'd love to see/work on more PRs that add in Blender/Unreal presets, or add in the ability to use mouse buttons 4 and 5. For now though, I wanted to make sure this way of doing it was functionally complete and looks good to everyone. This is my first time submitting code to Godot so please let me know if there's anything else I missed or if I can clarify anything further :) thanks!

### DIFF
--- a/editor/editor_settings_dialog.cpp
+++ b/editor/editor_settings_dialog.cpp
@@ -42,6 +42,7 @@
 #include "editor/editor_undo_redo_manager.h"
 #include "editor/event_listener_line_edit.h"
 #include "editor/input_event_configuration_dialog.h"
+#include "editor/plugins/node_3d_editor_plugin.h"
 #include "editor/themes/editor_scale.h"
 #include "editor/themes/editor_theme_manager.h"
 #include "scene/gui/panel_container.h"
@@ -74,7 +75,82 @@ void EditorSettingsDialog::_settings_property_edited(const String &p_name) {
 		EditorSettings::get_singleton()->set_manually("text_editor/theme/color_theme", "Custom");
 	} else if (full_name.begins_with("editors/visual_editors/connection_colors") || full_name.begins_with("editors/visual_editors/category_colors")) {
 		EditorSettings::get_singleton()->set_manually("editors/visual_editors/color_theme", "Custom");
+	} else if (full_name == "editors/3d/navigation/orbit_mouse_button" || full_name == "editors/3d/navigation/pan_mouse_button" || full_name == "editors/3d/navigation/zoom_mouse_button") {
+		EditorSettings::get_singleton()->set_manually("editors/3d/navigation/navigation_scheme", (int)Node3DEditorViewport::NAVIGATION_CUSTOM);
+	} else if (full_name == "editors/3d/navigation/navigation_scheme") {
+		update_navigation_preset();
 	}
+}
+
+void EditorSettingsDialog::update_navigation_preset() {
+	Node3DEditorViewport::NavigationScheme nav_scheme = (Node3DEditorViewport::NavigationScheme)EDITOR_GET("editors/3d/navigation/navigation_scheme").operator int();
+	Node3DEditorViewport::ViewportNavMouseButton set_orbit_mouse_button = Node3DEditorViewport::NAVIGATION_LEFT_MOUSE;
+	Node3DEditorViewport::ViewportNavMouseButton set_pan_mouse_button = Node3DEditorViewport::NAVIGATION_LEFT_MOUSE;
+	Node3DEditorViewport::ViewportNavMouseButton set_zoom_mouse_button = Node3DEditorViewport::NAVIGATION_LEFT_MOUSE;
+	Ref<InputEventKey> orbit_mod_key_1;
+	Ref<InputEventKey> orbit_mod_key_2;
+	Ref<InputEventKey> pan_mod_key_1;
+	Ref<InputEventKey> pan_mod_key_2;
+	Ref<InputEventKey> zoom_mod_key_1;
+	Ref<InputEventKey> zoom_mod_key_2;
+	bool set_preset = false;
+
+	if (nav_scheme == Node3DEditorViewport::NAVIGATION_GODOT) {
+		set_preset = true;
+		set_orbit_mouse_button = Node3DEditorViewport::NAVIGATION_MIDDLE_MOUSE;
+		set_pan_mouse_button = Node3DEditorViewport::NAVIGATION_MIDDLE_MOUSE;
+		set_zoom_mouse_button = Node3DEditorViewport::NAVIGATION_MIDDLE_MOUSE;
+		orbit_mod_key_1 = InputEventKey::create_reference(Key::NONE);
+		orbit_mod_key_2 = InputEventKey::create_reference(Key::NONE);
+		pan_mod_key_1 = InputEventKey::create_reference(Key::SHIFT);
+		pan_mod_key_2 = InputEventKey::create_reference(Key::NONE);
+		zoom_mod_key_1 = InputEventKey::create_reference(Key::SHIFT);
+		zoom_mod_key_2 = InputEventKey::create_reference(Key::CTRL);
+	} else if (nav_scheme == Node3DEditorViewport::NAVIGATION_MAYA) {
+		set_preset = true;
+		set_orbit_mouse_button = Node3DEditorViewport::NAVIGATION_LEFT_MOUSE;
+		set_pan_mouse_button = Node3DEditorViewport::NAVIGATION_MIDDLE_MOUSE;
+		set_zoom_mouse_button = Node3DEditorViewport::NAVIGATION_RIGHT_MOUSE;
+		orbit_mod_key_1 = InputEventKey::create_reference(Key::ALT);
+		orbit_mod_key_2 = InputEventKey::create_reference(Key::NONE);
+		pan_mod_key_1 = InputEventKey::create_reference(Key::NONE);
+		pan_mod_key_2 = InputEventKey::create_reference(Key::NONE);
+		zoom_mod_key_1 = InputEventKey::create_reference(Key::ALT);
+		zoom_mod_key_2 = InputEventKey::create_reference(Key::NONE);
+	} else if (nav_scheme == Node3DEditorViewport::NAVIGATION_MODO) {
+		set_preset = true;
+		set_orbit_mouse_button = Node3DEditorViewport::NAVIGATION_LEFT_MOUSE;
+		set_pan_mouse_button = Node3DEditorViewport::NAVIGATION_LEFT_MOUSE;
+		set_zoom_mouse_button = Node3DEditorViewport::NAVIGATION_LEFT_MOUSE;
+		orbit_mod_key_1 = InputEventKey::create_reference(Key::ALT);
+		orbit_mod_key_2 = InputEventKey::create_reference(Key::NONE);
+		pan_mod_key_1 = InputEventKey::create_reference(Key::SHIFT);
+		pan_mod_key_2 = InputEventKey::create_reference(Key::ALT);
+		zoom_mod_key_1 = InputEventKey::create_reference(Key::ALT);
+		zoom_mod_key_2 = InputEventKey::create_reference(Key::CTRL);
+	}
+	// Set settings to the desired preset values.
+	if (set_preset) {
+		EditorSettings::get_singleton()->set_manually("editors/3d/navigation/orbit_mouse_button", (int)set_orbit_mouse_button);
+		EditorSettings::get_singleton()->set_manually("editors/3d/navigation/pan_mouse_button", (int)set_pan_mouse_button);
+		EditorSettings::get_singleton()->set_manually("editors/3d/navigation/zoom_mouse_button", (int)set_zoom_mouse_button);
+		_set_shortcut_input("spatial_editor/viewport_orbit_modifier_1", orbit_mod_key_1);
+		_set_shortcut_input("spatial_editor/viewport_orbit_modifier_2", orbit_mod_key_2);
+		_set_shortcut_input("spatial_editor/viewport_pan_modifier_1", pan_mod_key_1);
+		_set_shortcut_input("spatial_editor/viewport_pan_modifier_2", pan_mod_key_2);
+		_set_shortcut_input("spatial_editor/viewport_zoom_modifier_1", zoom_mod_key_1);
+		_set_shortcut_input("spatial_editor/viewport_zoom_modifier_2", zoom_mod_key_2);
+	}
+}
+
+void EditorSettingsDialog::_set_shortcut_input(const String &p_name, Ref<InputEventKey> &p_event) {
+	Array sc_events;
+	if (p_event->get_keycode() != Key::NONE) {
+		sc_events.push_back((Variant)p_event);
+	}
+
+	Ref<Shortcut> sc = EditorSettings::get_singleton()->get_shortcut(p_name);
+	sc->set_events(sc_events);
 }
 
 void EditorSettingsDialog::_settings_save() {
@@ -96,6 +172,8 @@ void EditorSettingsDialog::popup_edit_settings() {
 	}
 
 	EditorSettings::get_singleton()->list_text_editor_themes(); // make sure we have an up to date list of themes
+
+	_update_dynamic_property_hints();
 
 	inspector->edit(EditorSettings::get_singleton());
 	inspector->get_inspector()->update_tree();
@@ -158,6 +236,12 @@ void EditorSettingsDialog::_notification(int p_what) {
 					EditorSettings::get_singleton()->check_changed_settings_in_group("builtin_action_overrides");
 			if (update_shortcuts_tab) {
 				_update_shortcuts();
+			}
+
+			if (EditorSettings::get_singleton()->check_changed_settings_in_group("editors/3d/navigation")) {
+				// Shortcuts may have changed, so dynamic hint values must update.
+				_update_dynamic_property_hints();
+				inspector->get_inspector()->update_tree();
 			}
 
 			if (EditorSettings::get_singleton()->check_changed_settings_in_group("interface/editor/localize_settings")) {
@@ -267,6 +351,13 @@ void EditorSettingsDialog::_update_shortcut_events(const String &p_path, const A
 	undo_redo->add_do_method(this, "_settings_changed");
 	undo_redo->add_undo_method(this, "_settings_changed");
 	undo_redo->commit_action();
+
+	bool path_is_orbit_mod = p_path == "spatial_editor/viewport_orbit_modifier_1" || p_path == "spatial_editor/viewport_orbit_modifier_2";
+	bool path_is_pan_mod = p_path == "spatial_editor/viewport_pan_modifier_1" || p_path == "spatial_editor/viewport_pan_modifier_2";
+	bool path_is_zoom_mod = p_path == "spatial_editor/viewport_zoom_modifier_1" || p_path == "spatial_editor/viewport_zoom_modifier_2";
+	if (path_is_orbit_mod || path_is_pan_mod || path_is_zoom_mod) {
+		EditorSettings::get_singleton()->set_manually("editors/3d/navigation/navigation_scheme", (int)Node3DEditorViewport::NAVIGATION_CUSTOM);
+	}
 }
 
 Array EditorSettingsDialog::_event_list_to_array_helper(const List<Ref<InputEvent>> &p_events) {
@@ -672,6 +763,40 @@ void EditorSettingsDialog::drop_data_fw(const Point2 &p_point, const Variant &p_
 
 void EditorSettingsDialog::_tabs_tab_changed(int p_tab) {
 	_focus_current_search_box();
+
+	// When tab has switched, shortcuts may have changed.
+	_update_dynamic_property_hints();
+	inspector->get_inspector()->update_tree();
+}
+
+void EditorSettingsDialog::_update_dynamic_property_hints() {
+	// Calling add_property_hint overrides the existing hint.
+	EditorSettings *settings = EditorSettings::get_singleton();
+	settings->add_property_hint(_create_mouse_shortcut_property_info("editors/3d/navigation/orbit_mouse_button", "spatial_editor/viewport_orbit_modifier_1", "spatial_editor/viewport_orbit_modifier_2"));
+	settings->add_property_hint(_create_mouse_shortcut_property_info("editors/3d/navigation/pan_mouse_button", "spatial_editor/viewport_pan_modifier_1", "spatial_editor/viewport_pan_modifier_2"));
+	settings->add_property_hint(_create_mouse_shortcut_property_info("editors/3d/navigation/zoom_mouse_button", "spatial_editor/viewport_zoom_modifier_1", "spatial_editor/viewport_zoom_modifier_2"));
+}
+
+PropertyInfo EditorSettingsDialog::_create_mouse_shortcut_property_info(const String &p_property_name, const String &p_shortcut_1_name, const String &p_shortcut_2_name) {
+	String hint_string;
+	hint_string += _get_shortcut_button_string(p_shortcut_1_name) + _get_shortcut_button_string(p_shortcut_2_name);
+	hint_string += "Left Mouse,";
+	hint_string += _get_shortcut_button_string(p_shortcut_1_name) + _get_shortcut_button_string(p_shortcut_2_name);
+	hint_string += "Middle Mouse,";
+	hint_string += _get_shortcut_button_string(p_shortcut_1_name) + _get_shortcut_button_string(p_shortcut_2_name);
+	hint_string += "Right Mouse";
+
+	return PropertyInfo(Variant::INT, p_property_name, PROPERTY_HINT_ENUM, hint_string);
+}
+
+String EditorSettingsDialog::_get_shortcut_button_string(const String &p_shortcut_name) {
+	String button_string;
+	Ref<Shortcut> shortcut_ref = EditorSettings::get_singleton()->get_shortcut(p_shortcut_name);
+	Array events = shortcut_ref->get_events();
+	for (Ref<InputEvent> input_event : events) {
+		button_string += input_event->as_text() + " + ";
+	}
+	return button_string;
 }
 
 void EditorSettingsDialog::_focus_current_search_box() {
--- a/editor/editor_settings_dialog.h
+++ b/editor/editor_settings_dialog.h
@@ -100,6 +100,10 @@ class EditorSettingsDialog : public AcceptDialog {
 	void _tabs_tab_changed(int p_tab);
 	void _focus_current_search_box();
 
+	void _update_dynamic_property_hints();
+	PropertyInfo _create_mouse_shortcut_property_info(const String &p_property_name, const String &p_shortcut_1_name, const String &p_shortcut_2_name);
+	String _get_shortcut_button_string(const String &p_shortcut_name);
+
 	void _filter_shortcuts(const String &p_filter);
 	void _filter_shortcuts_by_event(const Ref<InputEvent> &p_event);
 	bool _should_display_shortcut(const String &p_name, const Array &p_events) const;
@@ -107,6 +111,7 @@ class EditorSettingsDialog : public AcceptDialog {
 	void _update_shortcuts();
 	void _shortcut_button_pressed(Object *p_item, int p_column, int p_idx, MouseButton p_button = MouseButton::LEFT);
 	void _shortcut_cell_double_clicked();
+	static void _set_shortcut_input(const String &p_name, Ref<InputEventKey> &p_event);
 
 	static void _undo_redo_callback(void *p_self, const String &p_name);
 
@@ -124,6 +129,7 @@ protected:
 
 public:
 	void popup_edit_settings();
+	static void update_navigation_preset();
 
 	EditorSettingsDialog();
 	~EditorSettingsDialog();
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -1690,6 +1690,10 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 	if (b.is_valid()) {
 		emit_signal(SNAME("clicked"), this);
 
+		ViewportNavMouseButton orbit_mouse_preference = (ViewportNavMouseButton)EDITOR_GET("editors/3d/navigation/orbit_mouse_button").operator int();
+		ViewportNavMouseButton pan_mouse_preference = (ViewportNavMouseButton)EDITOR_GET("editors/3d/navigation/pan_mouse_button").operator int();
+		ViewportNavMouseButton zoom_mouse_preference = (ViewportNavMouseButton)EDITOR_GET("editors/3d/navigation/zoom_mouse_button").operator int();
+
 		const real_t zoom_factor = 1 + (ZOOM_FREELOOK_MULTIPLIER - 1) * b->get_factor();
 		switch (b->get_button_index()) {
 			case MouseButton::WHEEL_UP: {
@@ -1707,8 +1711,6 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 				}
 			} break;
 			case MouseButton::RIGHT: {
-				NavigationScheme nav_scheme = (NavigationScheme)EDITOR_GET("editors/3d/navigation/navigation_scheme").operator int();
-
 				if (b->is_pressed() && _edit.gizmo.is_valid()) {
 					//restore
 					_edit.gizmo->commit_handle(_edit.gizmo_handle, _edit.gizmo_handle_secondary, _edit.gizmo_initial_value, true);
@@ -1716,11 +1718,15 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 				}
 
 				if (_edit.mode == TRANSFORM_NONE && b->is_pressed()) {
-					if (b->is_alt_pressed()) {
-						if (nav_scheme == NAVIGATION_MAYA) {
-							break;
-						}
+					if (orbit_mouse_preference == NAVIGATION_RIGHT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2")) {
+						break;
+					} else if (pan_mouse_preference == NAVIGATION_RIGHT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2")) {
+						break;
+					} else if (zoom_mouse_preference == NAVIGATION_RIGHT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2")) {
+						break;
+					}
 
+					if (b->is_alt_pressed()) {
 						_list_select(b);
 						return;
 					}
@@ -1751,6 +1757,14 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 			} break;
 			case MouseButton::MIDDLE: {
 				if (b->is_pressed() && _edit.mode != TRANSFORM_NONE) {
+					if (orbit_mouse_preference == NAVIGATION_MIDDLE_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2")) {
+						break;
+					} else if (pan_mouse_preference == NAVIGATION_MIDDLE_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2")) {
+						break;
+					} else if (zoom_mouse_preference == NAVIGATION_MIDDLE_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2")) {
+						break;
+					}
+
 					switch (_edit.plane) {
 						case TRANSFORM_VIEW: {
 							_edit.plane = TRANSFORM_X_AXIS;
@@ -1789,8 +1803,11 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 						commit_transform();
 						break; // just commit the edit, stop processing the event so we don't deselect the object
 					}
-					NavigationScheme nav_scheme = (NavigationScheme)EDITOR_GET("editors/3d/navigation/navigation_scheme").operator int();
-					if ((nav_scheme == NAVIGATION_MAYA || nav_scheme == NAVIGATION_MODO) && b->is_alt_pressed()) {
+					if (orbit_mouse_preference == NAVIGATION_LEFT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2")) {
+						break;
+					} else if (pan_mouse_preference == NAVIGATION_LEFT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2")) {
+						break;
+					} else if (zoom_mouse_preference == NAVIGATION_LEFT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2")) {
 						break;
 					}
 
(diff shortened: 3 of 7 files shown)


YeldhamDev:
You need to run `--doctool` to update the documentation with the new settings: https://docs.godotengine.org/en/stable/contributing/documentation/updating_the_class_reference.html#updating-class-reference-when-working-on-the-engine

YeldhamDev:
There was another style error. You can set up a git hook to check for those locally: https://docs.godotengine.org/en/stable/contributing/development/code_style_guidelines.html#using-clang-format-locally

RobProductions:
> You need to run `--doctool` to update the documentation with the new settings

Actually I did but the doctool doesn't seem remove references to deleted members in other descriptions, oops. My bad for not catching it before though! And thanks for the info on the local styling tests, I'll check that out. Appreciate your patience with me setting this stuff up for the first time :)

RobProductions:
Rebased to handle merge conflict

editor/editor_settings_dialog.cpp
```diff
@@ -261,6 +337,13 @@ void EditorSettingsDialog::_update_shortcut_events(const String &p_path, const A
 	undo_redo->add_do_method(this, "_settings_changed");
 	undo_redo->add_undo_method(this, "_settings_changed");
 	undo_redo->commit_action();
+
+	bool path_is_orbit_mod = (p_path == "spatial_editor/viewport_orbit_modifier_1" || p_path == "spatial_editor/viewport_orbit_modifier_2");
+	bool path_is_pan_mod = (p_path == "spatial_editor/viewport_pan_modifier_1" || p_path == "spatial_editor/viewport_pan_modifier_2");
+	bool path_is_zoom_mod = (p_path == "spatial_editor/viewport_zoom_modifier_1" || p_path == "spatial_editor/viewport_zoom_modifier_2");
```
AThousandShips, line 343:
```suggestion
	bool path_is_orbit_mod = p_path == "spatial_editor/viewport_orbit_modifier_1" || p_path == "spatial_editor/viewport_orbit_modifier_2";
	bool path_is_pan_mod = p_path == "spatial_editor/viewport_pan_modifier_1" || p_path == "spatial_editor/viewport_pan_modifier_2";
	bool path_is_zoom_mod = p_path == "spatial_editor/viewport_zoom_modifier_1" || p_path == "spatial_editor/viewport_zoom_modifier_2";
```
Unnecessary parenthesis, hurts readability

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -1651,20 +1655,22 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
...
 				if (_edit.mode == TRANSFORM_NONE && b->is_pressed()) {
-					if (b->is_alt_pressed()) {
-						if (nav_scheme == NAVIGATION_MAYA) {
-							break;
-						}
+					if (orbit_mouse_preference == NAVIGATION_RIGHT_MOUSE && (_is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2"))) {
+						break;
+					} else if (pan_mouse_preference == NAVIGATION_RIGHT_MOUSE && (_is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2"))) {
+						break;
+					} else if (zoom_mouse_preference == NAVIGATION_RIGHT_MOUSE && (_is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2"))) {
```
AThousandShips, line 1669:
```suggestion
					if (orbit_mouse_preference == NAVIGATION_RIGHT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2")) {
						break;
					} else if (pan_mouse_preference == NAVIGATION_RIGHT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2")) {
						break;
					} else if (zoom_mouse_preference == NAVIGATION_RIGHT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2")) {
```

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -1695,6 +1701,14 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 			} break;
 			case MouseButton::MIDDLE: {
 				if (b->is_pressed() && _edit.mode != TRANSFORM_NONE) {
+					if (orbit_mouse_preference == NAVIGATION_MIDDLE_MOUSE && (_is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2"))) {
+						break;
+					} else if (pan_mouse_preference == NAVIGATION_MIDDLE_MOUSE && (_is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2"))) {
+						break;
+					} else if (zoom_mouse_preference == NAVIGATION_MIDDLE_MOUSE && (_is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2"))) {
```
AThousandShips, line 1708:
```suggestion
					if (orbit_mouse_preference == NAVIGATION_MIDDLE_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2")) {
						break;
					} else if (pan_mouse_preference == NAVIGATION_MIDDLE_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2")) {
						break;
					} else if (zoom_mouse_preference == NAVIGATION_MIDDLE_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2")) {
```

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -1733,8 +1747,11 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 						commit_transform();
 						break; // just commit the edit, stop processing the event so we don't deselect the object
 					}
-					NavigationScheme nav_scheme = (NavigationScheme)EDITOR_GET("editors/3d/navigation/navigation_scheme").operator int();
-					if ((nav_scheme == NAVIGATION_MAYA || nav_scheme == NAVIGATION_MODO) && b->is_alt_pressed()) {
+					if (orbit_mouse_preference == NAVIGATION_LEFT_MOUSE && (_is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2"))) {
+						break;
+					} else if (pan_mouse_preference == NAVIGATION_LEFT_MOUSE && (_is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2"))) {
+						break;
+					} else if (zoom_mouse_preference == NAVIGATION_LEFT_MOUSE && (_is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2"))) {
```
AThousandShips, line 1754:
```suggestion
					if (orbit_mouse_preference == NAVIGATION_LEFT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2")) {
						break;
					} else if (pan_mouse_preference == NAVIGATION_LEFT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2")) {
						break;
					} else if (zoom_mouse_preference == NAVIGATION_LEFT_MOUSE && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2")) {
```

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -1963,7 +1980,12 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 			_transform_gizmo_select(_edit.mouse_pos, true);
 		}
 
-		NavigationScheme nav_scheme = (NavigationScheme)EDITOR_GET("editors/3d/navigation/navigation_scheme").operator int();
+		ViewportNavMouseButton orbit_mouse_preference = (ViewportNavMouseButton)EDITOR_GET("editors/3d/navigation/orbit_mouse_button").operator int();
+		ViewportNavMouseButton pan_mouse_preference = (ViewportNavMouseButton)EDITOR_GET("editors/3d/navigation/pan_mouse_button").operator int();
+		ViewportNavMouseButton zoom_mouse_preference = (ViewportNavMouseButton)EDITOR_GET("editors/3d/navigation/zoom_mouse_button").operator int();
+		bool orbit_mod_pressed = (_is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2"));
+		bool pan_mod_pressed = (_is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2"));
+		bool zoom_mod_pressed = (_is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2"));
```
AThousandShips, line 1988:
```suggestion
		bool orbit_mod_pressed = _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2");
		bool pan_mod_pressed = _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2");
		bool zoom_mod_pressed = _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2");
```

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -2012,43 +2032,35 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
...
-					nav_mode = NAVIGATION_ZOOM;
-				} else if (mod == Key::ALT || mod == _get_key_modifier_setting("editors/3d/navigation/orbit_modifier")) {
-					// Always allow Alt as a modifier to better support graphic tablets.
-					nav_mode = NAVIGATION_ORBIT;
-				}
+			if ((!_is_shortcut_empty("spatial_editor/viewport_zoom_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_zoom_modifier_2")) && zoom_mod_pressed) {
+				nav_mode = NAVIGATION_ZOOM;
+			} else if ((!_is_shortcut_empty("spatial_editor/viewport_pan_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_pan_modifier_2")) && pan_mod_pressed) {
+				nav_mode = NAVIGATION_PAN;
+			} else if ((!_is_shortcut_empty("spatial_editor/viewport_orbit_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_orbit_modifier_2")) && orbit_mod_pressed) {
```
AThousandShips, line 2062:
```suggestion
			if (zoom_mod_pressed && (!_is_shortcut_empty("spatial_editor/viewport_zoom_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_zoom_modifier_2"))) {
				nav_mode = NAVIGATION_ZOOM;
			} else if (pan_mod_pressed && (!_is_shortcut_empty("spatial_editor/viewport_pan_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_pan_modifier_2"))) {
				nav_mode = NAVIGATION_PAN;
			} else if (orbit_mod_pressed && (!_is_shortcut_empty("spatial_editor/viewport_orbit_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_orbit_modifier_2"))) {
```
Cheaper condition first

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -2089,25 +2101,17 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 
 	Ref<InputEventPanGesture> pan_gesture = p_event;
 	if (pan_gesture.is_valid()) {
-		NavigationScheme nav_scheme = (NavigationScheme)EDITOR_GET("editors/3d/navigation/navigation_scheme").operator int();
+		bool orbit_mod_pressed = (_is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2"));
+		bool pan_mod_pressed = (_is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2"));
+		bool zoom_mod_pressed = (_is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2"));
```
AThousandShips, line 2106:
```suggestion
		bool orbit_mod_pressed = _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_orbit_modifier_2");
		bool pan_mod_pressed = _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_pan_modifier_2");
		bool zoom_mod_pressed = _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_1") && _is_nav_modifier_pressed("spatial_editor/viewport_zoom_modifier_2");
```

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -2089,25 +2101,17 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
...
-
-		} else if (nav_scheme == NAVIGATION_MAYA) {
-			if (pan_gesture->is_alt_pressed()) {
-				nav_mode = NAVIGATION_PAN;
-			}
+		if ((!_is_shortcut_empty("spatial_editor/viewport_zoom_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_zoom_modifier_2")) && zoom_mod_pressed) {
+			nav_mode = NAVIGATION_ZOOM;
+		} else if ((!_is_shortcut_empty("spatial_editor/viewport_pan_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_pan_modifier_2")) && pan_mod_pressed) {
+			nav_mode = NAVIGATION_PAN;
+		} else if ((!_is_shortcut_empty("spatial_editor/viewport_orbit_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_orbit_modifier_2")) && orbit_mod_pressed) {
```
AThousandShips, line 2113:
```suggestion
		if (zoom_mod_pressed && (!_is_shortcut_empty("spatial_editor/viewport_zoom_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_zoom_modifier_2"))) {
			nav_mode = NAVIGATION_ZOOM;
		} else if (pan_mod_pressed && (!_is_shortcut_empty("spatial_editor/viewport_pan_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_pan_modifier_2"))) {
			nav_mode = NAVIGATION_PAN;
		} else if (orbit_mod_pressed && (!_is_shortcut_empty("spatial_editor/viewport_orbit_modifier_1") || !_is_shortcut_empty("spatial_editor/viewport_orbit_modifier_2"))) {
```

editor/plugins/node_3d_editor_plugin.h
```diff
@@ -291,6 +298,9 @@ class Node3DEditorViewport : public Control {
 	void _nav_orbit(Ref<InputEventWithModifiers> p_event, const Vector2 &p_relative);
 	void _nav_look(Ref<InputEventWithModifiers> p_event, const Vector2 &p_relative);
 
+	bool _is_shortcut_empty(const String &pname);
+	bool _is_nav_modifier_pressed(const String &pname);
```
AThousandShips, line 302:
```suggestion
	bool _is_shortcut_empty(const String &p_name);
	bool _is_nav_modifier_pressed(const String &p_name);
```
Prfix with `p_`

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -2571,6 +2575,17 @@ void Node3DEditorViewport::scale_freelook_speed(real_t scale) {
 	surface->queue_redraw();
 }
 
+bool Node3DEditorViewport::_is_nav_modifier_pressed(const String &pname) {
+	Input *inp = Input::get_singleton();
+	return (_is_shortcut_empty(pname) || inp->is_action_pressed(pname));
+}
+
+bool Node3DEditorViewport::_is_shortcut_empty(const String &pname) {
+	Ref<Shortcut> orbit_shortcut = ED_GET_SHORTCUT(pname);
```
AThousandShips, line 2584:
```suggestion
bool Node3DEditorViewport::_is_nav_modifier_pressed(const String &p_name) {
	return _is_shortcut_empty(p_name) || Input::get_singleton()->is_action_pressed(p_name);
}

bool Node3DEditorViewport::_is_shortcut_empty(const String &p_name) {
	Ref<Shortcut> orbit_shortcut = ED_GET_SHORTCUT(p_name);
```

RobProductions:
Done, thank you! Feel free to check out the updated code :)

AThousandShips:
Looks good! Haven't tested and this isn't my area but reads a lot better 🙂

SamuelBucher:
This pull request seems to be exactly what I'm looking for. But I'm new to Godot, and the Pull request workflow in general. How do I get this pull request in my local copy of the source code?

Broken1334:
This is very nice addition!
But I'm not sure about control separation (mouse buttons in Editor Settings, modifier keys in Shortcuts), it might break UX consistency.
Maybe better to move all navigation controls to Editor Settings (Marvelous Designer) or Shortcuts (Blender)?
For example, Marvelous Designer's View Controls:
KPpcyCSiSQ

RobProductions:
> This is very nice addition! But I'm not sure about control separation (mouse buttons in Editor Settings, modifier keys in Shortcuts), it might break UX consistency. Maybe better to move all navigation controls to Editor Settings (Marvelous Designer) or Shortcuts (Blender)? 

Thanks! Yeah, that would be ideal but I don't know of an easy way to bring shortcut controls into the regular settings dialog or the other way around. The way the code is currently built, it'll automatically populate the shortcut menu with all the controls once they're registered so there's no quick way I can think of to insert a dropdown in between them. This would be a great addition for a future PR but would have to come in multiple steps, i.e. adding support for displaying shortcuts in the general settings tab in a specific order and then implementing that for these controls. 

For now it's at least functional for those that need it and the presets should be easy enough to use for anyone that just needs the same behavior as the current build :) 

> If I understand it right, that way is used in Unity, Blender and Maya. Unreal works like Godot. But, yeah, if current behavior is preferred, it should be fixed.

It is a bit jarring to work with as it forces you to do more to work around it (hence the multiple shortcuts you see here) so I would like to see it addressed, though I don't know if it'll need to be an option to keep the original behavior for compatibility purposes. Either way, that's out of my scope for now but if it is fixed I can come back to this and simplify these controls.

I just rebased on latest master to handle the merge conflict and tested to make sure it's still working, let me know if there's anything else I should do here!

RobProductions:
Rebased! Any chance this can get a review before it gets too stale?

Calinou (commented):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/811ce36c6090013f1a48676c0388892bf1621288), it works as expected.

It's nice that this adds a visual preview of what buttons the users should press after switching navigation schemes. However, the modifiers are not displayed, so some users may be confused when they try to use the left mouse button without <kbd>Alt</kbd> and seeing nothing happen when using the Modo navigation scheme:

image

I suggest displaying the modifiers that were selected in the Shortcuts tab in the enum option names, like on this mockup:

image

It should be possible to dynamically adjust enum option names by modifying the generated property hint. Values will remain compatible as long as the order is kept intact and the enum is integer-based (and not string-based like a few project settings are).

RobProductions:
@Calinou Gotcha, I definitely see how it's useful to preview the shortcuts alongside the mouse button though personally it's a little confusing seeing them in the "value" section of what's supposed to be a mouse button setting. Maybe in the future we can add a separate label generated by the different values to keep the mouse button selection clean? 

Anyways, your idea sounds good for now and I think I've figured it out, take a look:

image

And when you update the preset or the shortcuts, the hint values for the mouse buttons are also updated automatically. It worked like you described because the enum is int based:

image

The only caveat was finding a good time to refresh the tree so there can be a small delay after changing the setting for it to update the hint (it relies on a notification), but it's not worse than the usual delays you see in the settings screen so should be functional for now! Finally, while I was at it I also fixed the code ordering issue that I explained in the original post in order to reduce work for later PRs. Now you can have any amount of shortcuts on orbit, pan, and zoom, and the action with the least amount of shortcuts is always checked first so nothing should ever get skipped. 

Feel free to check out the updated code and let me know what you think :) Thanks!

Calinou (approved):
This looks great now! This helps a lot with discoverability, especially for the Zoom key with the Godot navigation scheme (<kbd>Ctrl + Shift + Middle Mouse</kbd>) which was pretty hard to discover until now.

editor/plugins/node_3d_editor_plugin.h
```diff
@@ -388,6 +399,28 @@ class Node3DEditorViewport : public Control {
...
+		_ShortcutCheckSet(bool p_mod_pressed, bool p_shortcut_not_empty, int p_input_count, const ViewportNavMouseButton &p_mouse_preference, const NavigationMode &p_result_nav_mode) :
+				mod_pressed(p_mod_pressed), shortcut_not_empty(p_shortcut_not_empty), input_count(p_input_count), mouse_preference(p_mouse_preference), result_nav_mode(p_result_nav_mode) {
+		}
+	};
+
+	struct _ShortcutCheckSetComparator {
+		_FORCE_INLINE_ bool operator()(const _ShortcutCheckSet &A, const _ShortcutCheckSet &B) const {
+			return A.input_count > B.input_count;
+		}
+	};
```
Calinou, line 420: I don't recall seeing private structs starting with an underscore in Godot. We just name the like public structs as far as I know.

RobProductions: Ah okay, I was following the `_RayResult` struct in the same file but you're right, it seems more common the other way elsewhere. I've renamed it now :) let me know if there's anything else I should adjust!

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -2433,6 +2431,32 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
...
+	Ref<Shortcut> check_shortcut = ED_GET_SHORTCUT(p_name);
+	if (!check_shortcut->get_events().is_empty()) {
+		return check_shortcut->get_events().size();
+	}
+	return 0;
+}
+
+Node3DEditorViewport::NavigationMode Node3DEditorViewport::_get_nav_mode_from_shortcut_check(ViewportNavMouseButton p_mouse_button, Vector<ShortcutCheckSet> p_shortcut_check_sets, bool p_use_not_empty) {
+	if (p_use_not_empty) {
+		for (int i = 0; i < p_shortcut_check_sets.size(); i++) {
```
AThousandShips, line 2444:
```suggestion
		for (const ShortcutCheckSet &shortcut_check_set : p_shortcut_check_sets) {
```
Use for-each here

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -2433,6 +2431,32 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
...
+			if (p_shortcut_check_sets[i].mod_pressed && p_shortcut_check_sets[i].shortcut_not_empty) {
+				return p_shortcut_check_sets[i].result_nav_mode;
+			}
+		}
+	} else {
+		for (int i = 0; i < p_shortcut_check_sets.size(); i++) {
```
AThousandShips, line 2450: Same here

editor/editor_settings_dialog.h
```diff
@@ -100,13 +100,18 @@ class EditorSettingsDialog : public AcceptDialog {
 	void _tabs_tab_changed(int p_tab);
 	void _focus_current_search_box();
 
+	void _update_dynamic_property_hints();
+	PropertyInfo _create_mouse_shortcut_property_info(String p_property_name, String p_shortcut_1_name, String p_shortcut_2_name);
```
AThousandShips, line 104:
```suggestion
	PropertyInfo _create_mouse_shortcut_property_info(const String &p_property_name, const String &p_shortcut_1_name, const String &p_shortcut_2_name);
```

editor/editor_settings_dialog.h
```diff
@@ -100,13 +100,18 @@ class EditorSettingsDialog : public AcceptDialog {
...
+	String _get_shortcut_button_string(String p_shortcut_name);
```
AThousandShips, line 105:
```suggestion
	String _get_shortcut_button_string(const String &p_shortcut_name);
```

editor/editor_settings_dialog.cpp
```diff
@@ -672,6 +763,43 @@ void EditorSettingsDialog::drop_data_fw(const Point2 &p_point, const Variant &p_
...
+	hint_string += _get_shortcut_button_string(p_shortcut_1_name) + _get_shortcut_button_string(p_shortcut_2_name);
+	hint_string += "Middle Mouse,";
+	hint_string += _get_shortcut_button_string(p_shortcut_1_name) + _get_shortcut_button_string(p_shortcut_2_name);
+	hint_string += "Right Mouse";
+
+	return PropertyInfo(Variant::INT, p_property_name, PROPERTY_HINT_ENUM, hint_string);
+}
+
+String EditorSettingsDialog::_get_shortcut_button_string(String p_shortcut_name) {
+	String button_string = "";
```
AThousandShips, line 793:
```suggestion
	String button_string;
```

editor/editor_settings_dialog.cpp
```diff
@@ -672,6 +763,43 @@ void EditorSettingsDialog::drop_data_fw(const Point2 &p_point, const Variant &p_
...
+	Ref<Shortcut> shortcut_ref = EditorSettings::get_singleton()->get_shortcut(p_shortcut_name);
+	Array events = shortcut_ref->get_events();
+	if (!events.is_empty()) {
+		for (int i = 0; i < events.size(); i++) {
```
AThousandShips, line 797:
```suggestion
		for (Ref<InputEvent> input_event : events) {
```
Note no `&`

editor/editor_settings_dialog.cpp
```diff
@@ -672,6 +763,43 @@ void EditorSettingsDialog::drop_data_fw(const Point2 &p_point, const Variant &p_
...
+void EditorSettingsDialog::_update_dynamic_property_hints() {
+	// Calling add_property_hint overrides the existing hint
+	EditorSettings *settings = EditorSettings::get_singleton();
+	settings->add_property_hint(_create_mouse_shortcut_property_info("editors/3d/navigation/orbit_mouse_button", "spatial_editor/viewport_orbit_modifier_1", "spatial_editor/viewport_orbit_modifier_2"));
+	settings->add_property_hint(_create_mouse_shortcut_property_info("editors/3d/navigation/pan_mouse_button", "spatial_editor/viewport_pan_modifier_1", "spatial_editor/viewport_pan_modifier_2"));
+	settings->add_property_hint(_create_mouse_shortcut_property_info("editors/3d/navigation/zoom_mouse_button", "spatial_editor/viewport_zoom_modifier_1", "spatial_editor/viewport_zoom_modifier_2"));
+}
+
+PropertyInfo EditorSettingsDialog::_create_mouse_shortcut_property_info(String p_property_name, String p_shortcut_1_name, String p_shortcut_2_name) {
+	String hint_string = "";
```
AThousandShips, line 781:
```suggestion
	String hint_string;
```

editor/editor_settings_dialog.cpp
```diff
@@ -74,7 +75,82 @@ void EditorSettingsDialog::_settings_property_edited(const String &p_name) {
...
+		_set_shortcut_input("spatial_editor/viewport_pan_modifier_2", pan_mod_key_2);
+		_set_shortcut_input("spatial_editor/viewport_zoom_modifier_1", zoom_mod_key_1);
+		_set_shortcut_input("spatial_editor/viewport_zoom_modifier_2", zoom_mod_key_2);
+	}
+}
+
+void EditorSettingsDialog::_set_shortcut_input(const String &p_name, Ref<InputEventKey> &p_event) {
+	Ref<Shortcut> sc = EditorSettings::get_singleton()->get_shortcut(p_name);
+	Array sc_events = sc->get_events();
+	sc_events.clear();
```
AThousandShips, line 149: Why the clear and then reassignment?

editor/editor_settings_dialog.cpp
```diff
@@ -160,6 +238,12 @@ void EditorSettingsDialog::_notification(int p_what) {
 				_update_shortcuts();
 			}
 
+			if (EditorSettings::get_singleton()->check_changed_settings_in_group("editors/3d/navigation")) {
+				// Shortcuts may have changed, so hint values must update
```
AThousandShips, line 242:
```suggestion
				// Shortcuts may have changed, so hint values must update.
```

editor/editor_settings_dialog.cpp
```diff
@@ -672,6 +763,43 @@ void EditorSettingsDialog::drop_data_fw(const Point2 &p_point, const Variant &p_
 
 void EditorSettingsDialog::_tabs_tab_changed(int p_tab) {
 	_focus_current_search_box();
+
+	// When tab has switched, shortcuts may have changed
```
AThousandShips, line 767:
```suggestion
	// When tab has switched, shortcuts may have changed.
```

editor/editor_settings_dialog.cpp
```diff
@@ -672,6 +763,43 @@ void EditorSettingsDialog::drop_data_fw(const Point2 &p_point, const Variant &p_
...
+	_update_dynamic_property_hints();
+	inspector->get_inspector()->update_tree();
+}
+
+void EditorSettingsDialog::_update_dynamic_property_hints() {
+	// Calling add_property_hint overrides the existing hint
```
AThousandShips, line 773:
```suggestion
	// Calling add_property_hint overrides the existing hint.
```

editor/plugins/node_3d_editor_plugin.h
```diff
@@ -388,6 +399,28 @@ class Node3DEditorViewport : public Control {
...
 	void scale_cursor_distance(real_t scale);
 
+	struct ShortcutCheckSet {
+		bool mod_pressed = false;
+		bool shortcut_not_empty = true;
+		int input_count = 0;
+		ViewportNavMouseButton mouse_preference = NAVIGATION_LEFT_MOUSE;
+		NavigationMode result_nav_mode = NAVIGATION_NONE;
+
+		ShortcutCheckSet(){};
```
AThousandShips, line 409:
```suggestion
		ShortcutCheckSet() {}
```

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -2433,6 +2431,32 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 	}
 }
 
+int Node3DEditorViewport::_get_shortcut_input_count(const String &p_name) {
+	Ref<Shortcut> check_shortcut = ED_GET_SHORTCUT(p_name);
+	if (!check_shortcut->get_events().is_empty()) {
+		return check_shortcut->get_events().size();
+	}
+	return 0;
```
AThousandShips, line 2439: This can just be replaced with:
```cpp
return check_shortcut->get_events().size();
```
Should there be a null check here instead? Is `ED_GET_SHORTCUT` guaranteed to return a valid ref here?

editor/editor_settings_dialog.cpp
```diff
@@ -672,6 +763,43 @@ void EditorSettingsDialog::drop_data_fw(const Point2 &p_point, const Variant &p_
...
+	hint_string += "Right Mouse";
+
+	return PropertyInfo(Variant::INT, p_property_name, PROPERTY_HINT_ENUM, hint_string);
+}
+
+String EditorSettingsDialog::_get_shortcut_button_string(String p_shortcut_name) {
+	String button_string = "";
+	Ref<Shortcut> shortcut_ref = EditorSettings::get_singleton()->get_shortcut(p_shortcut_name);
+	Array events = shortcut_ref->get_events();
+	if (!events.is_empty()) {
```
AThousandShips, line 796: Unnecessary if we're doing a loop

AThousandShips (commented):
Looks good but just some questions before I hit approve

editor/editor_settings_dialog.cpp
RobProductions: It seems `set_events()` is necessary for the shortcut to run `emit_changed()` and breaks otherwise. But good catch since we don't need to do the get_events() step, I can just make it a new Array. A question though, do I have to do anything to free up the memory used by the old array when it's replaced or will Godot take care of that? I think I did it this way originally to avoid any complications with that

AThousandShips: It will, it's fully safe, that's not a problem

editor/plugins/node_3d_editor_plugin.h
RobProductions: Unfortunately clang format removes the space when I try to add it... and I get a warning when I leave out the default constructor. Any ideas?

AThousandShips: It doesn't if you follow it properly, you didn't remove the `;`

editor/plugins/node_3d_editor_plugin.cpp
RobProductions: Ah you're right, `ED_GET_SHORTCUT` does error if the shortcut is nullptr. I'll simplify this as you suggested

AThousandShips (approved):

RobProductions:
@AThousandShips Thanks for the review and guidance! I've updated the code with all your suggestions, let me know if this looks better now

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -2632,6 +2653,15 @@ void Node3DEditorViewport::scale_freelook_speed(real_t scale) {
 	surface->queue_redraw();
 }
 
+bool Node3DEditorViewport::_is_nav_modifier_pressed(const String &p_name) {
+	return _is_shortcut_empty(p_name) || Input::get_singleton()->is_action_pressed(p_name);
+}
+
+bool Node3DEditorViewport::_is_shortcut_empty(const String &p_name) {
+	Ref<Shortcut> check_shortcut = ED_GET_SHORTCUT(p_name);
+	return check_shortcut->get_events().is_empty();
```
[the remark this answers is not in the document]
RobProductions: It can error but this isn't a dynamic check, p_name will only ever be what is put in node_3d_editor_plugin source code. Since that's the case I thought we wanted it to error, right? Since it would signify a typo on the dev side. Should I catch it and print some other error message instead?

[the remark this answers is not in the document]
AThousandShips: I'd say yes, as that would prevent a segmentation fault here, so instead use some:
```cpp
ERR_FAIL_COND_V_MSG(check_shortcut.is_null(), true, "...");
```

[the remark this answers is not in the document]
RobProductions: Done, hopefully this works

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -2632,6 +2656,18 @@ void Node3DEditorViewport::scale_freelook_speed(real_t scale) {
...
 }
 
+bool Node3DEditorViewport::_is_nav_modifier_pressed(const String &p_name) {
+	return _is_shortcut_empty(p_name) || Input::get_singleton()->is_action_pressed(p_name);
+}
+
+bool Node3DEditorViewport::_is_shortcut_empty(const String &p_name) {
+	Ref<Shortcut> check_shortcut = ED_GET_SHORTCUT(p_name);
+	if (check_shortcut.is_null()) {
+		ERR_FAIL_COND_V_MSG(check_shortcut.is_null(), true, "The Shortcut was null, possible name mismatch.");
```
AThousandShips, line 2666: Just the check, no if statement

RobProductions: Whoops, apologies. Fixed it now

akien-mga:
Thanks!

Nodragem:
Hello!
I am testing this in 4.4dev2. 
I am using a pen tablet so I don't have a middle click and use `Emulate 3 Button Mouse`.

Would it be possible to keep the default for `Emulate 3 Button Mouse` to ALT for orbit, SHIFT for pan and CTRL for zoom? Otherwise, it can be very confusing when we open 4.4 for the first time: the view is rotating without us touching anything.

Let me know if I should let my feedback somewhere else.

RobProductions:
Hi @Nodragem , in order to maximize compatibility with these new customization features, I think its best if the `Emulate 3 Button Mouse` setting does exactly what it says in the docs:

> If true, enables 3-button mouse emulation mode. This is useful on laptops when using a trackpad. When 3-button mouse emulation mode is enabled, the pan, zoom and orbit modifiers can always be used in the 3D editor viewport, even when not holding down any mouse button.

Otherwise people who want to customize orbit, pan, and zoom keys for tablet would be forced into ALT, SHIFT, and CTRL even if they have something different selected in their settings. Unfortunately the view rotating in your case is a side effect of the default navigation scheme being `Godot` which uses no keyboard press for orbit, only mouse input, so when 3 button mouse mode comes along and eliminates the need for mouse input, it will always rotate. To fix this you can change the preset to `Modo` or customize the `Viewport Orbit Modifier` in the Shortcuts tab so that a key is used. 

The only way I could see the "default" improving without eliminating the customization is if we had a different setting track the "3 button mouse"-specific modifier keys separately from standard keys, but that seems a little excessive. In my opinion, if you're changing 3 button mouse from the default state of "off" it's hopefully fine to simply change the preset as well; to me that's preferable over losing 3 button mouse key customization.

Alternatively, I could add a `Tablet` preset with your requested hotkeys to sit alongside `Godot` and the others. Though it'll still require an extra setting switch in addition to 3 button mouse, it should make things more clear to Tablet users that they should be using it instead of `Godot`. Does that sound like a good solution?

Zireael07:
Default rotating by itself with 3 button emulation is really bad UX. Can Modo be made default?

Zireael07:
I am on trackpad and you're right, in previous versions I needed to press ALT to emulate middle mouse and I didn't have the camera rotate iirc

RobProductions:
> Default rotating by itself with 3 button emulation is really bad UX. Can Modo be made default?

I don't mind going that route but I'm not sure if Godot users/maintainers will be okay with the `Godot` preset no longer being the default for non-trackpad users when you open the engine for the first time.

> I am on trackpad and you're right, in previous versions I needed to press ALT to emulate middle mouse and I didn't have the camera rotate iirc

Yeah, that's because it was hardcoded before - when 3 Button Mouse was on, it made assumptions about the scheme and just manually checked for ALT, or something like that, while ignoring mouse input. To adhere to the doc original description, "even when not holding down any mouse button" it now checks for your custom keybind while ignoring mouse input, like before except now the default scheme of `Godot` uses "no keybind" for orbit. If we change it back to ALT then 3 button mouse users lose nav customization. imo keeping the customization is preferable, but maybe we can do something where it warns the users that no key is currently set to Orbit while this button is on, prompting them to set a key or change scheme to `Tablet` as suggested above.

Perhaps "Emulate 3 Button Mouse" is better renamed to "Simulate mouse input" because that's what it was always doing? Maybe it was named that before because it kind of seemed like it was emulating just the middle mouse, but in reality it made assumptions and forced you to use the "global orbit modifier" even if your "navigation scheme" ignored it. 

> Thinking of it, your idea of adding a Tablet preset is better. Just need to automatically activate it when the user click on Emulate 3 Button mouse. Might need to rename the preset to Laptop/Pen Tablet.

Okay, thanks for the input! However I am weary to automatically set preset when clicking 3 button mouse to on - it seems a little intuitive for it to change based on that, especially if a 3 button mouse user wants Maya or Modo setting instead, we're now forcing them to go through redundant clicks. What if it worked the other way around? When you set preset to `Tablet` (or whatever name we land on) it also activates 3 button mouse, so it just requires one click as before but you simply change preset instead of that toggle. If we go this route, I could also tie in 3 button mouse to the preset detection so that changing it from `off` to `on` while using Godot, Maya, or Modo will swap preset to "custom" and make it more clear that you're doing something unexpected. Likewise, changing it to `off` while in Tablet will also make the preset switch to "custom". Does this still sound good?

prominentdetail:
Is it just me or is this no longer working in 4.4 dev 7 ?? I can't get a custom viewport orbit modifier to work.

PR #85346, opened 2023-11-25
Added Project Setting to disable cutting entire line when there is no text selected.

### LINKED ISSUE #59441: `cut` command (`ctrl+x`) deletes whole line even if nothing selected
### Godot version

v3.5.beta2.official [7a4f9dfb1]

### System information

Mac OSX (Intel 2019)

### Issue description

When in the script editor, press `Ctrl+x` (`Cmd+x` on mac).

The whole current line will be deleted. This often leads to annoyances if you are working in `Emacs` as `Ctrl+x` is such a common keybinding. 

I often find myself accidentally deleting one line and then trying to run the game.

### Steps to reproduce

When in the script editor, press `Ctrl+x` (`Cmd+x` on mac).

_No response_

TheSofox:
Fixes #59441

By default, the Godot text/code editor will behave exactly as it has before. This default and continuing behaviour is that when the user issues a cut command while no text is selected, the entire line the caret is on will be cut (VSCode, for example, has such behaviour).

However, there is now a setting (Project Settings -> GUI -> Common -> Cut Line on Empty Selection) which when disabled (true by default) will prevent this from happening. Cutting with no text selected will simply do nothing.

### DIFF
--- a/doc/classes/EditorSettings.xml
+++ b/doc/classes/EditorSettings.xml
@@ -830,6 +830,9 @@
 			The indentation style to use (tabs or spaces).
 			[b]Note:[/b] The [url=$DOCS_URL/tutorials/scripting/gdscript/gdscript_styleguide.html]GDScript style guide[/url] recommends using tabs for indentation. It is advised to change this setting only if you need to work on a project that currently uses spaces for indentation.
 		</member>
+		<member name="text_editor/behavior/navigation/cut_line_on_empty_selection" type="bool" setter="" getter="">
+			If [code]true[/code], cutting while no text is selected will cut the entire line the caret is on.
+		</member>
 		<member name="text_editor/behavior/navigation/drag_and_drop_selection" type="bool" setter="" getter="">
 			If [code]true[/code], allows drag-and-dropping text in the script editor to move text. Disable this if you find yourself accidentally drag-and-dropping text in the script editor.
 		</member>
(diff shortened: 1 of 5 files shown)


Chubercik:
Wouldn't it make more sense for this to be in `Editor Settings` instead of `Project Settings`?

I feel like this option concerns a developer's preference, not something that ought to be bound to a given project.

TheSofox:
The setting affects TextEdit as a whole, not just the Code Editor (which inherits from TextEdit). Any project you export with a TextEdit field will have use the behaviour you set for it.

It also affects any other TextEdit field in the Editor (for instance, when you're editing the Text field of a Label).

So we're dealing with three elements here:

1. Code Editor in Editor
2. Regular TextEdit boxes throughout the Editor
3. TextEdit boxes in the exported project

Before this push, all supported cutting a line on empty selection. This PR request has them all affected by a Project Setting.

TheSofox:
Okay, thinking this over, it's clear the desire is that:

1. This setting is in the Editor Settings
2. It only affects the Code Editor.

I can do this by adding a variable to TextEdit that controls this and having the CodeEditor set it to whatever the Editor setting dictates. By default, the TextEdit variable will be set to allow cutting on empty, since that's existing behaviour.

Chubercik (approved):

TheSofox:
I've been thinking over this and I'm wondering whether I made the best call by putting it in Navigation category. It's definitely the closest category, and there are similar attributes in this category such as Drag and Drop Selection, but I'm wondering whether I should have created a whole new category for this.

I could name this new category "Editing". Put it in the same location as the other categories (Text Editor -> Behavior). Thoughts?

Chubercik:
I think such addition would warrant moving certain options to this new category (such as the mentioned Drag and Drop Selection), so it would be best done (if at all) in the form of another PR.

KoBeWi:
Superseded by #90743
Thanks for the contribution nonetheless.

PR #85362, opened 2023-11-25
Make copy & paste params skip `resource_path`

### LINKED ISSUE #42415: Copy paste parameters not persistent (on material)
**Godot version:**

3.2.3.stable.mono.official

Windows 10 x64

**Issue description:**

Copy params -> paste params onto another material only keeps the pasted values as long as the material is inspected. Even saving doesn't retain them and they're lost upon reselecting the asset.

**Steps to reproduce:**
See the this demonstration:
https://streamable.com/mpziu8

Edit: There's an error though, i missed:
"Another resource is loaded from path 'res://Shared/Maps/Environments/LandingCave/Rocks.material' (possible cyclic resource inclusion)."

Flavelius:
Resource path should not be attempted to be taken over, as that's not intended for copy-paste and fails anyway, but this results in the whole paste operation failing as well.
This change skips it in the copy part, but there could be an argument for skipping it in the paste operation.

fixes https://github.com/godotengine/godot/issues/42415

### DIFF
--- a/editor/editor_data.cpp
+++ b/editor/editor_data.cpp
@@ -283,7 +283,7 @@ void EditorData::copy_object_params(Object *p_object) {
 	p_object->get_property_list(&pinfo);
 
 	for (const PropertyInfo &E : pinfo) {
-		if (!(E.usage & PROPERTY_USAGE_EDITOR) || E.name == "script" || E.name == "scripts") {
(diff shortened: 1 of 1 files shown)


KoBeWi (approved):
Instead of hard-coding every non-copiable property I think we should add something like `PROPERTY_USAGE_SKIP_COPY`.

This solution is ok for now though.

YuriSizov:
Thanks, and congrats on your first merged Godot contribution! (Was it really the first one for you? Quite unexpected 🙂 )

PR #85364, opened 2023-11-25
Instantiate type before getting argument options for autocompletion

### LINKED ISSUE #76515: autocomplete on call argumetn sometimes fail when no object is
### Godot version

master (240ed28291)

### System information

windows 11

### Issue description

when trying to autocomplete the argument of a function call and it can't find an associated object it won't autocomplete.
I found two case where he might not found an object first it can be if the scene is not opened or if referencing a node from a node path

In both cases i tested to autocomplete the first parameter of `emit_signal`. Might be only related to it.

```gdscript
extends Node2D

@export var path : NodePath
@onready var target : Node2D = get_node(path)

func _ready():
	target.emit_signal()

```

### Steps to reproduce

- copy paste the script above
- set the `NodePath `to some `Node2D`
- try to autocomplete something it will work
- try to autocomplete the first argument of `emit__signal` it won't work

N/A

### LINKED ISSUE #103191: Auto completamento não funciona para caminhos de propriedades ao serem usadas @onready vars ou @export para referenciar nós
### Tested versions

Reproducible on godot 4.3 stable

### System information

Windows 10 - godot 4.3 stable

### Issue description

When trying to access node property parameters in functions like set() or get(), auto complete does not show the options if the node is being referenced by @onready vars or @export.

### Steps to reproduce

Node referenced by @export vars:

Image

Node referenced by @onready vars:

Image

Node referenced by "$" + node_name:

Image

This is not very interesting and slows down development a lot, as the developer needs to go to the inspector and copy the property path or guess the property path.

### LINKED ISSUE #103243: list of global groups not always autocompleted when using group related node methods
### Tested versions

- Reproducible in Godot 4.4 RC 1

### System information

Godot v4.4.rc1 - Windows 11 (build 26100) - Multi-window, 1 monitor - OpenGL 3 (Compatibility) - NVIDIA GeForce RTX 4080 Laptop GPU (NVIDIA; 32.0.15.6117) - 13th Gen Intel(R) Core(TM) i9-13900HX (32 threads)

### Issue description

I was wondering if the situation below is intended behavior for the Godot script editor. The list of available global groups only appears in any of the group related methods for a node which is referenced using `$` or `%` notation.

This problem seems related to godotengine/godot#99277 and this summary (https://github.com/godotengine/godot/issues/86172#issuecomment-1860646328) of autocomplete issues in Godot. 

I can understand not wanting to give incorrect options for an unknown node in the case of node.get_node() or animations in the case of animation_player.play(). However, no matter what node (given that it has been typed or casted as a Node) you are calling `is_in_group()` on, providing the list of global groups shouldn't do any harm.

### Steps to reproduce

I've provided several examples of where the autocomplete works in the Godot Script editor, and where it does not.

```gdscript
extends Node2D

@onready var test_node: Node = Node.new() as Node
@onready var control: Control = $Control
@onready var control_1: Control = %Control

func _ready() -> void:
	
	is_in_group("globalTestGroup") # Gives list of options
	get_tree().get_nodes_in_group("globalTestGroup") # Gives list of options
	control.is_in_group("globalTestGroup") # Gives list of options
	control_1.is_in_group("globalTestGroup") # Gives list of options
	
	var this_node: Node = self as Node
	
	self.is_in_group("globalTestGroup") # Doesn't give list of options
	this_node.is_in_group("globalTestGroup") # Doesn't give list of options
	Node.new().is_in_group("globalTestGroup") # Doesn't give list of options
	test_node.is_in_group("globalTestGroup") # Doesn't give list of options

func test_func(passed_node: Node):
	passed_node.is_in_group("globalTestGroup") # Doesn't give list of options
	
	if passed_node is Node:
		passed_node.is_in_group("globalTestGroup") # Doesn't give list of options

```

N/A

HolonProduction:
Fixes #76515
Fixes #103243
Fixes #103191

### DIFF
--- a/modules/gdscript/gdscript_editor.cpp
+++ b/modules/gdscript/gdscript_editor.cpp
@@ -2779,6 +2779,10 @@ static void _find_call_arguments(GDScriptParser::CompletionContext &p_context, c
 					break;
 				}
 
+				if (base.get_type() == Variant::NIL && ClassDB::can_instantiate(class_name)) {
+					base = ClassDB::instantiate(class_name);
+				}
+
 				MethodInfo info;
 				int method_args = 0;
 
@@ -2813,6 +2817,12 @@ static void _find_call_arguments(GDScriptParser::CompletionContext &p_context, c
 					}
 
 					r_arghint = _make_arguments_hint(info, p_argidx);
+				}
+
+				if (p_base.value.get_type() == Variant::NIL) {
+					if (!base.is_ref_counted()) {
+						memdelete_notnull(base.get_validated_object());
+					}
 				}
 
 				if (p_argidx == 1 && p_context.node && p_context.node->type == GDScriptParser::Node::CALL && ClassDB::is_parent_class(class_name, SNAME("Tween")) && p_method == SNAME("tween_property")) {
--- a/modules/gdscript/tests/scripts/completion/common/argument_options_without_value.cfg
+++ b/modules/gdscript/tests/scripts/completion/common/argument_options_without_value.cfg
@@ -1,0 +1,4 @@
+[output]
+include=[
+    {"display": "\"pressed\""},
+]
--- a/modules/gdscript/tests/scripts/completion/common/argument_options_without_value.gd
+++ b/modules/gdscript/tests/scripts/completion/common/argument_options_without_value.gd
@@ -1,0 +1,4 @@
+var button: Button
+
+func _ready():
+    button.emit_signal("➡")


modules/gdscript/gdscript_editor.cpp
```diff
@@ -2668,6 +2668,10 @@ static void _find_call_arguments(GDScriptParser::CompletionContext &p_context, c
 					break;
 				}
 
+				if (base.get_type() == Variant::NIL && ClassDB::can_instantiate(class_name)) {
+					base = ClassDB::instantiate(class_name);
```
Mickeon, line 2672: This dummy `base` would have to be freed then, wouldn't it?

modules/gdscript/gdscript_editor.cpp
```diff
@@ -2808,6 +2812,15 @@ static void _find_call_arguments(GDScriptParser::CompletionContext &p_context, c
 					r_arghint = _make_arguments_hint(info, p_argidx);
 				}
 
+				if (p_base.value.get_type() == Variant::NIL) {
+					if (!base.is_ref_counted()) {
+						Object *obj = base.get_validated_object();
+						if (obj != nullptr) {
+							memfree(obj);
```
AThousandShips, line 2819: I'd say to use `memdelete_notnull` instead and not use the check against nullptr

HolonProduction: Is there maybe also some builtin, that handles refcounted correctly?

HolonProduction:
Rebased and added a test

Mickeon:
Moving to 4.4 to complement existing autocompletion improvements.

HolonProduction:
Looking at this two years later, I don't stand behind the approach anymore. As discussed in https://github.com/godotengine/godot/pull/109297 calling into arbitrary code from autocompletion is risky, the same obviously applies to constructing instances.

I'll be closing this PR for this reason. I hope we can replace the current system with a better one that isn't tied to instances in the future.

PR #85373, opened 2023-11-26
Fix updating cached singletons when reloading GDScripts

### LINKED ISSUE #85004: GDScript does not update cached singleton references when GDExtension is reloaded
### Godot version

v4.2.rc.mono.custom_build [ad72de508]

### System information

Windows 10

### Issue description

As I pointed out several times earlier in [#80284](https://github.com/godotengine/godot/pull/80284#issuecomment-1722120503), the references to singletons in `GDScript` are still not updated after the `GDExtension` reload.

https://github.com/godotengine/godot/assets/7782218/f90052a5-0de2-4310-8235-f0b9e2f74b3a

Singletons themselves are updated and work correctly, but references to these singletons are not updated in `@tool` scripts. That is, `Engine.get_singleton` returns the actual working reference, but using the singleton directly through its name causes errors.

### Steps to reproduce

* Clone `4.2.reload_singleton` branch from the MRP repository https://github.com/DmitriySalnikov/gde_test_cases/tree/4.2.reload_singleton
* Build as usual
* Open the project from the `project/` folder
* Add one character to the source code and rebuild
* Focus on the editor to reload extension
* Watch for error spam when calling by singleton name

https://github.com/DmitriySalnikov/gde_test_cases/tree/4.2.reload_singleton

DmitriySalnikov:
Fixes #85004

Since `GDScriptLanguage::reload_all_scripts()` is always called when the `GDExtension` reload is completed, I decided to update the singleton cache there.
I used the [original singleton caching](https://github.com/godotengine/godot/blob/5df98679672c12d96b4ac4d96ee17f1559207401/modules/gdscript/gdscript.cpp#L2149-L2155) code, added a key existence check, and just overwrite the values.

Without `globals.has(E.name)` singletons that were registered via `GDScript` itself or for example `GDScriptLanguageProtocol` can be added to the cache after the `GDExtension` reload and used as the rest of the singletons in `GDScript`.

#52162 could be fixed if calling `reload_all_scripts` somewhere after loading the editor ([there?](https://github.com/godotengine/godot/blob/5df98679672c12d96b4ac4d96ee17f1559207401/main/main.cpp#L3287-L3289)) and remove `globals.has(E.name)`. I tried updating scripts in `Engine::add_singleton`, but it is advisable to do it deferred, and `Engine` is a regular class that cannot call methods via `callable_mp` (although it was possible to add a static method and a static boolean variable to prevent multiple calls in one frame). This can be tested in a separate PR, but I'm not really interested in it.

It would be cool to merge this before the 4.2 release.

### DIFF
--- a/modules/gdscript/gdscript.cpp
+++ b/modules/gdscript/gdscript.cpp
@@ -2323,6 +2323,19 @@ void GDScriptLanguage::reload_all_scripts() {
 			}
 			elem = elem->next();
 		}
+
+#ifdef TOOLS_ENABLED
+		if (Engine::get_singleton()->is_editor_hint()) {
+			// Reload all pointers to existing singletons so that tool scripts can work with the reloaded extensions.
+			List<Engine::Singleton> singletons;
+			Engine::get_singleton()->get_singletons(&singletons);
+			for (const Engine::Singleton &E : singletons) {
+				if (globals.has(E.name)) {
+					_add_global(E.name, E.ptr);
+				}
+			}
+		}
+#endif
 	}
 
 	//as scripts are going to be reloaded, must proceed without locking here


DmitriySalnikov:
I simplified this commit. Now everything happens in `reload_all_scripts`. Singletons are still updated only in the editor. But I'm not sure if it's necessary to place this code inside a block with a lock?

Right now there are only [two cases](https://github.com/search?q=repo%3Agodotengine%2Fgodot+-%3Ereload_all_scripts&type=code) when a `reload_all_scripts` is called: [RemoteDebugger in a running game](https://github.com/godotengine/godot/blob/5df98679672c12d96b4ac4d96ee17f1559207401/core/debugger/remote_debugger.cpp#L596), after saving scripts in the editor, [GDExtensionManager in the editor](https://github.com/godotengine/godot/blob/5df98679672c12d96b4ac4d96ee17f1559207401/core/extension/gdextension_manager.cpp#L223) when the libraries are reloaded. Therefore, only the second case will be affected here, while nothing is added or deleted, only pointers are updated, which without this change do not allow using singletons with GDExtension hot reload.

Btw 4.2 will be the first release with hot reload support for GDExtension. 4.2 RC2 isn't the last release candidate, is it?

akien-mga:
> Btw 4.2 will be the first release with hot reload support for GDExtension. 4.2 RC2 isn't the last release candidate, is it?

A release candidate should always expect to be the last, otherwise it's not a release candidate ;)

This doesn't seem critical (doesn't prevent shipping games, only impacts hot reload during extension development), so it doesn't seem urgent to merge mere days before the stable release. Like other bug fixes, it can be merged afterwards and cherry-picked for 4.2.1.

ValorZard:
I'd like to offer a counterpoint. Hot Reloading for GDExtension seems like it will be a very popular feature for 4.2, which means making sure its bug free and ready to be used is probably important. Although, I'm not sure how annoying this bug is, so YMMV.

dsnopek (approved):
Thanks!

This fixes the issue with the MRP from https://github.com/godotengine/godot/issues/85004 in my testing.

The code makes sense to me based on looking at the references from the PR description, but it should probably be reviewed by someone on the GDScript team.

AThousandShips (approved):

vnen (approved):
Makes sense to me. Refreshing the pointers to the singletons are necessary if the extensions are reloaded.

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.1.

PR #85386, opened 2023-11-26
Improve SceneTreeEditor usability

### LINKED ISSUE #5480: SceneTreeDialog: better navigation for large scene trees

Game with large scene tree

A note on terminology: In the following sections, "originating node" refers to the node from which the NodePaths are relative to. It is equivalent to the node that you opened the selection dialog from: 
image

---

We use a lot of `@export var x: Node` etc. in our code. Our levels are large scene trees with a lot of nodes in them.

When opening the "Select a Node" dialog, it's hard to orient yourself, since it starts scrolled to the top and the originating node is not highlighted in any way:

image

- Scroll the tree view so that the originating node is centered in the view.
- Highlight the originating node, so it is easier to e.g. select a child or parent node.

Here I used the "Override" icon with 50% opacity to highlight the originating node. The dialog also starts with the scroll bar at the bottom.

image

No, this is a UX improvement.

Can't be add-on.

RedMser:
Closes https://github.com/godotengine/godot-proposals/issues/5480

A mix of usability improvements and bug fixes for the SceneTreeEditor and SceneTreeDialog. Here's an overview of the changes:

- Mark contextually relevant node using alternate text color
- Set/Clear initial selection as appropriate
- Scroll to selected node (or if none is selected, scroll to marked node)

An example of how the dialog now starts out when assigning an `@export var x: Node` variable:

image

### DIFF
--- a/editor/animation_track_editor.cpp
+++ b/editor/animation_track_editor.cpp
@@ -5006,12 +5006,17 @@ void AnimationTrackEditor::_new_track_node_selected(NodePath p_path) {
 }
 
 void AnimationTrackEditor::_add_track(int p_type) {
-	if (!root) {
+	AnimationPlayer *ap = AnimationPlayerEditor::get_singleton()->get_player();
+	if (!ap) {
+		ERR_FAIL_EDMSG("No AnimationPlayer is currently being edited.");
+	}
+	Node *root_node = ap->get_node_or_null(ap->get_root_node());
+	if (!root_node) {
 		EditorNode::get_singleton()->show_warning(TTR("Not possible to add a new track without a root"));
 		return;
 	}
 	adding_track_type = p_type;
-	pick_track->popup_scenetree_dialog();
+	pick_track->popup_scenetree_dialog(nullptr, root_node);
 	pick_track->get_filter_line_edit()->clear();
 	pick_track->get_filter_line_edit()->grab_focus();
 }
--- a/editor/editor_properties.cpp
+++ b/editor/editor_properties.cpp
@@ -2751,7 +2751,15 @@ void EditorPropertyNodePath::_node_assign() {
 		add_child(scene_tree);
 		scene_tree->connect("selected", callable_mp(this, &EditorPropertyNodePath::_node_selected));
 	}
-	scene_tree->popup_scenetree_dialog();
+
+	Variant val = get_edited_property_value();
+	Node *n = nullptr;
+	if (val.get_type() == Variant::Type::NODE_PATH) {
+		n = get_base_node()->get_node_or_null(val);
+	} else {
+		n = Object::cast_to<Node>(val);
+	}
+	scene_tree->popup_scenetree_dialog(n, get_base_node());
 }
 
 void EditorPropertyNodePath::_update_menu() {
@@ -3184,7 +3192,6 @@ void EditorPropertyResource::_resource_changed(const Ref<Resource> &p_resource) 
 			add_child(scene_tree);
 			scene_tree->connect("selected", callable_mp(this, &EditorPropertyResource::_viewport_selected));
 		}
-
 		scene_tree->popup_scenetree_dialog();
 	}
 }
--- a/editor/gui/scene_tree_editor.cpp
+++ b/editor/gui/scene_tree_editor.cpp
@@ -925,6 +925,27 @@ void SceneTreeEditor::_notification(int p_what) {
 
 			_update_tree();
 		} break;
+
+		case NOTIFICATION_VISIBILITY_CHANGED: {
+			if (is_visible()) {
+				TreeItem *item = nullptr;
+				if (selected) {
+					// Scroll to selected node.
+					item = _find(tree->get_root(), selected->get_path());
+				} else if (marked.size() == 1) {
+					// Scroll to a single marked node.
+					Node *marked_node = *marked.begin();
+					if (marked_node) {
+						item = _find(tree->get_root(), marked_node->get_path());
+					}
+				}
+
+				if (item) {
+					// Must wait until tree is properly sized before scrolling.
+					callable_mp(tree, &Tree::scroll_to_item).call_deferred(item, true);
+				}
+			}
+		} break;
 	}
 }
 
@@ -1567,7 +1588,9 @@ SceneTreeEditor::~SceneTreeEditor() {
 
 /******** DIALOG *********/
 
-void SceneTreeDialog::popup_scenetree_dialog() {
+void SceneTreeDialog::popup_scenetree_dialog(Node *p_selected_node, Node *p_marked_node, bool p_marked_node_selectable, bool p_marked_node_children_selectable) {
+	get_scene_tree()->set_marked(p_marked_node, p_marked_node_selectable, p_marked_node_children_selectable);
+	get_scene_tree()->set_selected(p_selected_node);
 	popup_centered_clamped(Size2(350, 700) * EDSCALE);
 }
 
--- a/editor/gui/scene_tree_editor.h
+++ b/editor/gui/scene_tree_editor.h
@@ -157,8 +157,8 @@ public:
 	void set_as_scene_tree_dock();
 	void set_display_foreign_nodes(bool p_display);
 
-	void set_marked(const HashSet<Node *> &p_marked, bool p_selectable = false, bool p_children_selectable = true);
-	void set_marked(Node *p_marked, bool p_selectable = false, bool p_children_selectable = true);
+	void set_marked(const HashSet<Node *> &p_marked, bool p_selectable = true, bool p_children_selectable = true);
+	void set_marked(Node *p_marked, bool p_selectable = true, bool p_children_selectable = true);
 	void set_selected(Node *p_node, bool p_emit_selected = true);
 	Node *get_selected();
 	void set_can_rename(bool p_can_rename) { can_rename = p_can_rename; }
@@ -201,7 +201,7 @@ protected:
 	static void _bind_methods();
 
 public:
-	void popup_scenetree_dialog();
+	void popup_scenetree_dialog(Node *p_selected_node = nullptr, Node *p_marked_node = nullptr, bool p_marked_node_selectable = true, bool p_marked_node_children_selectable = true);
 	void set_valid_types(const Vector<StringName> &p_valid);
 
 	SceneTreeEditor *get_scene_tree() { return tree; }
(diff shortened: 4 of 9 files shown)


Mickeon:
What's the "marked" Node in this context?

RedMser:
> What's the "marked" Node in this context?

All nodes that use the theme's accent color as their text color. It's not the same as the selected node.

For example, if you have a script on node X that contains `@export var foo: Node`, then you try assigning a node to `foo`, the marked node in the dialog is X.

Similar goes for e.g. adding a new track in AnimationPlayer. The marked node there is the animation root.

editor/plugins/cpu_particles_3d_editor_plugin.cpp
```diff
@@ -55,6 +55,8 @@ void CPUParticles3DEditor::_notification(int p_notification) {
 void CPUParticles3DEditor::_menu_option(int p_option) {
 	switch (p_option) {
 		case MENU_OPTION_CREATE_EMISSION_VOLUME_FROM_NODE: {
+			emission_tree_dialog->get_scene_tree()->set_marked(node, true);
+			emission_tree_dialog->get_scene_tree()->set_selected(nullptr);
 			emission_tree_dialog->popup_scenetree_dialog();
```
Mickeon, line 60: With these three methods all being required for the desired behavior in the PR, I wonder if it's worth clumping them up into a single function.

Perhaps by passing what they need as parameters in `popup_scenetree_dialog()`? If getting the `get_scene_tree()` is even possible from there.

editor/gui/scene_tree_editor.cpp
```diff
@@ -925,6 +925,24 @@ void SceneTreeEditor::_notification(int p_what) {
...
+		case NOTIFICATION_VISIBILITY_CHANGED: {
+			if (is_visible()) {
+				TreeItem *item = nullptr;
+				if (selected) {
+					// Scroll to selected node.
+					item = _find(tree->get_root(), selected->get_path());
+				} else if (marked.size() == 1) {
+					// Scroll to a single marked node.
+					Node *marked_node = *marked.begin();
+					item = marked_node ? _find(tree->get_root(), marked_node->get_path()) : nullptr;
```
Mickeon, line 938: `item` is intialized as `nullptr` so this seems superfluous.
```suggestion
					if (marked_node) {
						item = _find(tree->get_root(), marked_node->get_path());
					}			
```

editor/editor_properties.cpp
```diff
@@ -2751,6 +2751,18 @@ void EditorPropertyNodePath::_node_assign() {
...
+
+	scene_tree->get_scene_tree()->set_marked(get_base_node(), true, true);
+	Variant val = get_edited_property_value();
+	Node *n = nullptr;
+	if (val.get_type() == Variant::Type::NODE_PATH) {
+		NodePath path = val;
+		n = get_base_node()->get_node_or_null(path);
+	} else {
+		n = Object::cast_to<Node>(val);
+	}
```
Mickeon, line 2763: I assume the `path` here was for an explicit cast but I believe it may not be necessary for Variant?
```suggestion
	if (val.get_type() == Variant::Type::NODE_PATH) {
		n = get_base_node()->get_node_or_null(val);
	} else {
		n = Object::cast_to<Node>(val);
	}
```

editor/animation_track_editor.cpp
```diff
@@ -5011,7 +5011,7 @@ void AnimationTrackEditor::_add_track(int p_type) {
 		return;
 	}
 	adding_track_type = p_type;
-	pick_track->popup_scenetree_dialog();
+	pick_track->popup_scenetree_dialog(nullptr, root, true);
```
KoBeWi, line 5014: Something is wrong with the `root` variable. When you change the `root_node` property, you need to reselect the AnimationPlayer to update it.

editor/gui/scene_tree_editor.h
```diff
@@ -201,7 +201,7 @@ class SceneTreeDialog : public ConfirmationDialog {
 	static void _bind_methods();
 
 public:
-	void popup_scenetree_dialog();
+	void popup_scenetree_dialog(Node *p_selected_node = nullptr, Node *p_marked_node = nullptr, bool p_marked_node_selectable = false, bool p_marked_node_children_selectable = true);
```
KoBeWi, line 204: I find the default value of `p_marked_node_selectable` odd. It's `false`, but you pass `true` in almost every case, or not pass any marked node. Thus the default is effectively used only once, explicitly.

RedMser: Agreed. I took the same defaults as the `set_marked` method's defaults.
Did not want to deviate from that, but also did not want to change set_marked's defaults since that's a breaking change I imagine.

editor/gui/scene_tree_editor.cpp
```diff
@@ -925,6 +925,26 @@ void SceneTreeEditor::_notification(int p_what) {
...
+				} else if (marked.size() == 1) {
+					// Scroll to a single marked node.
+					Node *marked_node = *marked.begin();
+					if (marked_node) {
+						item = _find(tree->get_root(), marked_node->get_path());
+					}
+				}
+
+				if (item) {
+					tree->scroll_to_item(item, true);
```
KoBeWi, line 944: This doesn't seem to work reliably. The item is not always centered when you open the dialog multiple times for different nodes. Might require deferring or waiting until the dialog content is drawn.

editor/plugins/cpu_particles_3d_editor_plugin.cpp
```diff
@@ -55,8 +55,7 @@ void CPUParticles3DEditor::_notification(int p_notification) {
 void CPUParticles3DEditor::_menu_option(int p_option) {
 	switch (p_option) {
 		case MENU_OPTION_CREATE_EMISSION_VOLUME_FROM_NODE: {
-			emission_tree_dialog->popup_scenetree_dialog();
-
+			emission_tree_dialog->popup_scenetree_dialog(nullptr, node, true);
```
KoBeWi, line 58: This dialog only allows selecting MeshInstance3D, so the particles node is not marked nor selected.

editor/plugins/multimesh_editor_plugin.cpp
```diff
@@ -255,13 +255,15 @@ void MultiMeshEditor::edit(MultiMeshInstance3D *p_multimesh) {
...
-	std->popup_scenetree_dialog();
+	Node *browsed_node = nullptr;
 	if (p_source) {
+		browsed_node = node->get_node_or_null(mesh_source->get_text());
 		std->set_title(TTR("Select a Source Mesh:"));
 	} else {
+		browsed_node = node->get_node_or_null(surface_source->get_text());
 		std->set_title(TTR("Select a Target Surface:"));
 	}
+	std->popup_scenetree_dialog(browsed_node, node, false);
```
KoBeWi, line 266: Same as in particles.

editor/gui/scene_tree_editor.h
KoBeWi: SceneTreeEditor is not exposed, so you can freely change it.

editor/animation_track_editor.cpp
RedMser: Yep, seems to be updated too rarely. This also causes the warning above to not always show.

Other places in the AnimationTrackEditor appear to query the AnimationPlayer via a singleton, so I'll just follow that logic to get the up-to-date root node.

editor/plugins/multimesh_editor_plugin.cpp
RedMser: I guess there's no need to specify the marked node for these cases.
Even if you disable the node type filter, the marked node just appears greyed out here...

editor/gui/scene_tree_editor.cpp
RedMser: Oh you're right! I was able to reproduce it when picking Root Node for AnimationPlayer.
Adding a `call_deferred` was indeed enough to make it consistently work.

editor/gui/scene_tree_editor.cpp
```diff
@@ -925,6 +925,27 @@ void SceneTreeEditor::_notification(int p_what) {
...
+					// Scroll to a single marked node.
+					Node *marked_node = *marked.begin();
+					if (marked_node) {
+						item = _find(tree->get_root(), marked_node->get_path());
+					}
+				}
+
+				if (item) {
+					// Must wait until tree is properly sized before scrolling.
+					tree->call_deferred("scroll_to_item", item, true);
```
KoBeWi, line 945:
```suggestion
					callable_mp(tree, &Tree::scroll_to_item).call_deferred(item, true);
```

KoBeWi (approved):

editor/gui/scene_tree_editor.cpp
RedMser: Good call, thanks!

editor/gui/scene_tree_editor.cpp
```diff
@@ -925,6 +925,27 @@ void SceneTreeEditor::_notification(int p_what) {
...
+					// Scroll to a single marked node.
+					Node *marked_node = *marked.begin();
+					if (marked_node) {
+						item = _find(tree->get_root(), marked_node->get_path());
+					}
+				}
+
+				if (item) {
+					// Must wait until tree is properly sized before scrolling.
+					callable_mp(tree, &Tree::scroll_to_item).call_deferred(item, true);
```
bruvzg, line 945: This PR seems to be causing crashes. Probably deferred call should use `ObjectID` instead of raw pointer, otherwise the tree item might be removed before callback execution and pointer will become invalid.

RedMser: Can you show me other code locations that do something like this? Also if there's a reliable way to reproduce the crash would help.

bruvzg: > Can you show me other code locations that do something like this?

Not sure if there's exactly same situations, buy you can do something like:

Add a new callback:
```c++
void Tree::_scroll_to_item_id(ObjectID p_item_id, bool p_center_on_item) {
	TreeItem *item = Object::cast_to<TreeItem *>(ObjectDB::get_instance(p_item_id));
	if (item) {
		scroll_to_item(item, p_center_on_item);
	}
}
```

And call it like this
```c++
if (item) {
	ObjectID item_id = item->get_instance_id();
	callable_mp(tree, &Tree::_scroll_to_item_id).call_deferred(item_id, true);
}
```

PR #85387, opened 2023-11-26
Use render method from OS instead of project settings in compositor RD

### LINKED ISSUE #85143: `--rendering-method` CLI arguments are not effective on the actual scene rendering, only on warning messages
### Godot version

4.0.stable, 4.1.3.stable, 4.2.rc1

### System information

Fedora 38, GeForce RTX 4090 (NVIDIA 535.129.03)

### Issue description

`--rendering-method mobile` doesn't *actually* make a Forward+ project use the Mobile rendering method, even though the command line output says so.

### No CLI argument (use Forward+)

Looks like Forward+.

```
❯ bin/godot.linuxbsd.editor.x86_64 --path ~/Documents/Godot/test_rendering_method_cli_arguments 
Godot Engine v4.2.rc.custom_build.368e87aa0 - https://godotengine.org
Vulkan API 1.3.242 - Forward+ - Using Vulkan Device #0: NVIDIA - NVIDIA GeForce RTX 4090
```

### `--rendering-method mobile`

Looks like Forward+.

```
❯ bin/godot.linuxbsd.editor.x86_64 --path ~/Documents/Godot/test_rendering_method_cli_arguments --rendering-method mobile          
Godot Engine v4.2.rc.custom_build.368e87aa0 - https://godotengine.org
Vulkan API 1.3.242 - Forward Mobile - Using Vulkan Device #0: NVIDIA - NVIDIA GeForce RTX 4090
 
WARNING: SDFGI can only be enabled when using the Forward+ rendering backend.
     at: environment_set_sdfgi (./servers/rendering/storage/environment_storage.cpp:652)
```

#### `--rendering-method gl_compatibility` (or `--rendering-driver opengl3`)

Looks like Compatibility.

```
❯ bin/godot.linuxbsd.editor.x86_64 --path ~/Documents/Godot/test_rendering_method_cli_arguments --rendering-method gl_compatibility
Godot Engine v4.2.rc.custom_build.368e87aa0 - https://godotengine.org
OpenGL API 3.3.0 NVIDIA 535.129.03 - Compatibility - Using Device: NVIDIA - NVIDIA GeForce RTX 4090
 
WARNING: SDFGI can only be enabled when using the Forward+ rendering backend.
     at: environment_set_sdfgi (./servers/rendering/storage/environment_storage.cpp:652)
WARNING: Glow is not supported when using the GL Compatibility backend yet. Support will be added in a future release.
     at: environment_set_glow (./servers/rendering/storage/environment_storage.cpp:387)
```

### No CLI argument (use Mobile)

Looks like Mobile.

```
❯ bin/godot.linuxbsd.editor.x86_64 --path ~/Documents/Godot/test_rendering_method_cli_arguments 
Godot Engine v4.2.rc.custom_build.368e87aa0 - https://godotengine.org
Vulkan API 1.3.242 - Forward Mobile - Using Vulkan Device #0: NVIDIA - NVIDIA GeForce RTX 4090
 
WARNING: SDFGI can only be enabled when using the Forward+ rendering backend.
     at: environment_set_sdfgi (./servers/rendering/storage/environment_storage.cpp:652)
```

### `--rendering-method forward_plus`

Looks like Mobile.

```
❯ bin/godot.linuxbsd.editor.x86_64 --path ~/Documents/Godot/test_rendering_method_cli_arguments --rendering-method forward_plus    
Godot Engine v4.2.rc.custom_build.368e87aa0 - https://godotengine.org
Vulkan API 1.3.242 - Forward+ - Using Vulkan Device #0: NVIDIA - NVIDIA GeForce RTX 4090
```

#### `--rendering-method gl_compatibility` (or `--rendering-driver opengl3`)

Looks like Compatibility.

```
❯ bin/godot.linuxbsd.editor.x86_64 --path ~/Documents/Godot/test_rendering_method_cli_arguments --rendering-driver opengl3
Godot Engine v4.2.rc.custom_build.368e87aa0 - https://godotengine.org
OpenGL API 3.3.0 NVIDIA 535.129.03 - Compatibility - Using Device: NVIDIA - NVIDIA GeForce RTX 4090
 
WARNING: SDFGI can only be enabled when using the Forward+ rendering backend.
     at: environment_set_sdfgi (./servers/rendering/storage/environment_storage.cpp:652)
WARNING: Glow is not supported when using the GL Compatibility backend yet. Support will be added in a future release.
     at: environment_set_glow (./servers/rendering/storage/environment_storage.cpp:387)
```

MRP visuals for reference:

Forward+ | Mobile | Compatibility
-|-|-
Screenshot_20231120_181128 | Screenshot_20231120_181143 | Screenshot_20231120_181155

### Steps to reproduce

- Use `--rendering-method mobile` to [run a project from the command line](https://docs.godotengine.org/en/latest/tutorials/editor/command_line_tutorial.html) that uses Forward+ in its project settings.
- Alternatively, use `--rendering-method forward_plus` on a project that uses the Mobile rendering method.

[test_rendering_method_cli_arguments.zip](https://github.com/godotengine/godot/files/13417650/test_rendering_method_cli_arguments.zip)

### LINKED ISSUE #85376: Crash when passing `--rendering-method forward_plus` to a project configured for `gl_compatibility`
### Godot version

4.2.rc2 (1ba920fada9efc8c4476ded50fe673b8db541366)

### System information

Mageia 9 - Vulkan (Forward+) - dedicated AMD Radeon RX Vega M GL Graphics (RADV VEGAM) () - Intel(R) Core(TM) i7-8705G CPU @ 3.10GHz (8 Threads)

### Issue description

Passing `--rendering-method forward_plus` on the command line to a project configured with `gl_compatibility` as rendering method leads to a crash:
```
ERROR: Cannot instantiate RenderingDevice-based renderer with renderer type gl_compatibility
   at: RendererCompositorRD (./servers/rendering/renderer_rd/renderer_compositor_rd.cpp:321)
ERROR: Parameter "scene" is null.
   at: get_gi (./servers/rendering/renderer_rd/renderer_compositor_rd.h:112)

Thread 1 "godot-git" received signal SIGSEGV, Segmentation fault.
RendererSceneCull::set_scene_render (this=0xb65a410, p_scene_render=0x0) at ./servers/rendering/renderer_scene_cull.cpp:4140
4140            geometry_instance_pair_mask = scene_render->geometry_instance_get_pair_mask();
(gdb) bt
#0  RendererSceneCull::set_scene_render (this=0xb65a410, p_scene_render=0x0) at ./servers/rendering/renderer_scene_cull.cpp:4140
#1  0x00000000091d5a07 in RenderingServerDefault::RenderingServerDefault (this=0xb0e5450, p_create_thread=false) at ./servers/rendering/rendering_server_default.cpp:421
#2  0x000000000579194e in Main::setup2 () at main/main.cpp:2455
#3  0x00000000057900da in Main::setup (execpath=0x7fffffffda27 "/home/akien/.local/bin/godot-git", argc=2, argv=0x7fffffffd570, p_second_phase=true) at main/main.cpp:2184
#4  0x000000000570c3a5 in main (argc=3, argv=0x7fffffffd568) at platform/linuxbsd/godot_linuxbsd.cpp:62
```

This also seems reproducible in 4.1.3.

Likewise, passing `--rendering-driver vulkan`, or both (`rendering-method forward_plus --rendering-driver vulkan`) crashes similarly.

The issue also happens with `--rendering-method mobile`.

On the other hand, passing `--rendering-method gl_compatibility` and/or `--rendering-driver opengl3` to a RD-based project (`forward_plus` or `mobile`) works fine.

### Steps to reproduce

- Create new GL Compatibility project
- Run it from command line with `--rendering-method forward_plus`

Happens in a brand new GL Compatibility project.

jsjtxietian:
Fixes #85376

Not sure why Godot doesn't use the rendering method stored in OS in the first place, it's the correct one after all ( as it is computed in main.cpp ) . Feel free to correct me.

- *Production edit: This closes https://github.com/godotengine/godot/issues/85143.*

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -6572,6 +6572,23 @@ void EditorNode::_renderer_selected(int p_which) {
 	_update_renderer_color();
 }
 
+void EditorNode::_add_renderer_entry(const String &p_renderer_name, bool p_mark_overridden) {
+	String item_text;
+	if (p_renderer_name == "forward_plus") {
+		item_text = TTR("Forward+");
+	}
+	if (p_renderer_name == "mobile") {
+		item_text = TTR("Mobile");
+	}
+	if (p_renderer_name == "gl_compatibility") {
+		item_text = TTR("Compatibility");
+	}
+	if (p_mark_overridden) {
+		item_text += " " + TTR("(Overridden)");
+	}
+	renderer->add_item(item_text);
+}
+
 void EditorNode::_resource_saved(Ref<Resource> p_resource, const String &p_path) {
 	if (singleton->saving_resources_in_path.has(p_resource)) {
 		// This is going to be handled by save_resource_in_path when the time is right.
@@ -7496,7 +7513,6 @@ EditorNode::EditorNode() {
 	renderer->set_flat(true);
 	renderer->set_fit_to_longest_item(false);
 	renderer->set_focus_mode(Control::FOCUS_NONE);
-	renderer->connect("item_selected", callable_mp(this, &EditorNode::_renderer_selected));
 	renderer->add_theme_font_override("font", theme->get_font(SNAME("bold"), EditorStringName(EditorFonts)));
 	renderer->add_theme_font_size_override("font_size", theme->get_font_size(SNAME("bold_size"), EditorStringName(EditorFonts)));
 	renderer->set_tooltip_text(TTR("Choose a renderer."));
@@ -7510,36 +7526,33 @@ EditorNode::EditorNode() {
 		title_bar->add_child(right_menu_spacer);
 	}
 
-	String current_renderer = GLOBAL_GET("rendering/renderer/rendering_method");
+	String current_renderer_ps = GLOBAL_GET("rendering/renderer/rendering_method");
+	current_renderer_ps = current_renderer_ps.to_lower();
+	String current_renderer_os = OS::get_singleton()->get_current_rendering_method().to_lower();
 
-	PackedStringArray renderers = ProjectSettings::get_singleton()->get_custom_property_info().get(StringName("rendering/renderer/rendering_method")).hint_string.split(",", false);
-
-	// As we are doing string comparisons, keep in standard case to prevent problems with capitals
-	// "vulkan" in particular uses lowercase "v" in the code, and uppercase in the UI.
-	current_renderer = current_renderer.to_lower();
-
-	for (int i = 0; i < renderers.size(); i++) {
-		String rendering_method = renderers[i];
-
-		// Add the renderers name to the UI.
-		if (rendering_method == "forward_plus") {
-			renderer->add_item(TTR("Forward+"));
+	// Add the renderers name to the UI.
+	if (current_renderer_ps == current_renderer_os) {
+		renderer->connect("item_selected", callable_mp(this, &EditorNode::_renderer_selected));
+		// As we are doing string comparisons, keep in standard case to prevent problems with capitals
+		// "vulkan" in particular uses lowercase "v" in the code, and uppercase in the UI.
+		PackedStringArray renderers = ProjectSettings::get_singleton()->get_custom_property_info().get(StringName("rendering/renderer/rendering_method")).hint_string.split(",", false);
+		for (int i = 0; i < renderers.size(); i++) {
+			String rendering_method = renderers[i];
+			_add_renderer_entry(rendering_method, false);
+			renderer->set_item_metadata(i, rendering_method);
+			// Lowercase for standard comparison.
+			rendering_method = rendering_method.to_lower();
+			if (current_renderer_ps == rendering_method) {
+				renderer->select(i);
+				renderer_current = i;
+			}
 		}
-		if (rendering_method == "mobile") {
-			renderer->add_item(TTR("Mobile"));
-		}
-		if (rendering_method == "gl_compatibility") {
-			renderer->add_item(TTR("Compatibility"));
-		}
-		renderer->set_item_metadata(i, rendering_method);
-
-		// Lowercase for standard comparison.
-		rendering_method = rendering_method.to_lower();
-
-		if (current_renderer == rendering_method) {
-			renderer->select(i);
-			renderer_current = i;
-		}
+	} else {
+		// It's an CLI-overridden rendering method.
+		_add_renderer_entry(current_renderer_os, true);
+		renderer->set_item_metadata(0, current_renderer_os);
+		renderer->select(0);
+		renderer_current = 0;
 	}
 	_update_renderer_color();
 
--- a/editor/editor_node.h
+++ b/editor/editor_node.h
@@ -613,6 +613,7 @@ private:
 
 	void _renderer_selected(int);
 	void _update_renderer_color();
+	void _add_renderer_entry(const String &p_renderer_name, bool p_mark_overridden);
 
 	void _exit_editor(int p_exit_code);
 
--- a/servers/rendering/renderer_rd/renderer_compositor_rd.cpp
+++ b/servers/rendering/renderer_rd/renderer_compositor_rd.cpp
@@ -306,7 +306,7 @@ RendererCompositorRD::RendererCompositorRD() {
 	fog = memnew(RendererRD::Fog);
 	canvas = memnew(RendererCanvasRenderRD());
 
-	String rendering_method = GLOBAL_GET("rendering/renderer/rendering_method");
+	String rendering_method = OS::get_singleton()->get_current_rendering_method();
 	uint64_t textures_per_stage = RD::get_singleton()->limit_get(RD::LIMIT_MAX_TEXTURES_PER_SHADER_STAGE);
 
 	if (rendering_method == "mobile" || textures_per_stage < 48) {


RandomShaper:
LGTM. There are other points in the codebase doing the same, though. I think all of them warrant revisiting. `EditorNode` may need some extra love, though, as it needs to handle the difference between an CLI-overridden and project-setting rendering method. For instance, iIt may want to sufix _(overridden)_ to the renderer dropdown's caption, not allowing it to be further changed as it would then change the project setting.

jsjtxietian:
> I think all of them warrant revisiting.

yes and I think I can do it in this pr to prevent further crashes.

> `EditorNode` may need some extra love, though, as it needs to handle the difference between an CLI-overridden and project-setting rendering method.

That sounds like a usability improvement instead of fixing crash, and maybe can use a seperat pr.

Calinou (dismissed):
Tested locally, it works as expected. This also fixes https://github.com/godotengine/godot/issues/85143 :slightly_smiling_face: 

Forward+ project started with `--rendering-method mobile`:

Before (is actually using Forward+) | After (correct appearance for Mobile)
-|-
image | image

It also works the other way around (`--rendering-method forward_plus` on a Mobile project). `--rendering-method gl_compatibility` on a Forward+/Mobile project and `--rendering-method <forward_plus|mobile>` on a Comptaibility project also work.

PS: Should `OS.get_current_rendering_method()` also be used in the rendering dropdown in the top-right corner of the editor? Right now, the dropdown's current option will be incorrect if you use `--rendering-method` when opening the editor.

clayjohn (dismissed):

jsjtxietian:
@RandomShaper I push a new commit to do what you said, if the rendering method is overriden then it will only show current one with a suffix `overridden`.

editor/editor_node.cpp
```diff
@@ -6631,13 +6631,13 @@ void EditorNode::_bottom_panel_raise_toggled(bool p_pressed) {
 void EditorNode::_update_renderer_color() {
 	String rendering_method = renderer->get_selected_metadata();
 
-	if (rendering_method == "forward_plus") {
+	if (rendering_method.begins_with("forward_plus")) {
```
RandomShaper, line 6634: Why `begins_with`?

editor/editor_node.cpp
```diff
@@ -7591,36 +7590,52 @@ EditorNode::EditorNode() {
...
-
-		// Lowercase for standard comparison.
-		rendering_method = rendering_method.to_lower();
-
-		if (current_renderer == rendering_method) {
-			renderer->select(i);
-			renderer_current = i;
+		if (current_renderer_os == "gl_compatibility") {
+			renderer->add_item(TTR("Compatibility overridden"));
 		}
```
RandomShaper, line 7635: Above the `if (current_renderer_ps == current_renderer_os) {` you could add this (not checked for successful compile; feel free to fix) and use it in both branches of the `if (current_renderer_ps == current_renderer_os) {`, with `p_overridden` being `false`/`true` accordingly:
```C++
auto _add_renderer_entry = [renderer](const String& p_renderer_name, bool p_mark_overridden) {
		String item_text;
		if (p_renderer_name == "forward_plus") {
			item_text = TTR("Forward+");
		} else if (p_renderer_name == "mobile") {
			item_text = TTR("Mobile");
		} else if (p_renderer_name == "gl_compatibility") {
			item_text = TTR("Compatibility");
		} else {
			item_text = p_renderer_name; // Fallback to add unbeautified.
		}
		if (p_overridden) {
			item_text += " " + TTR("Overridden");
		}
		renderer->add_item(item_text);
};
```

Please ask me for any clarification.

RandomShaper (changes requested):
Good job. I've only added a comment about a refactor that would make the code more compact.

editor/editor_node.cpp
jsjtxietian: Oh sorry I forget to delete that, at first I set the metadata with a suffix too.

editor/editor_node.cpp
jsjtxietian: I like the lambda approach, but I though godot forbids lambda?

RandomShaper: They are allowed if the situation well justifies it. I've used them in cases like this, where they avoid code duplication, with the extra benefits of avoiding having to add a new member function, which also helps readability, as the lambda is in the scope where it's used.

DanielSnd:
I have confirmed it works (and I'm using it in my game to allow for launching in compatibility through steam launch options with arguments)

editor/editor_node.cpp
```diff
@@ -7591,36 +7590,50 @@ EditorNode::EditorNode() {
...
-		String rendering_method = renderers[i];
+	String current_renderer_ps = GLOBAL_GET("rendering/renderer/rendering_method");
+	current_renderer_ps = current_renderer_ps.to_lower();
+	String current_renderer_os = OS::get_singleton()->get_current_rendering_method().to_lower();
 
-		// Add the renderers name to the UI.
-		if (rendering_method == "forward_plus") {
-			renderer->add_item(TTR("Forward+"));
+	// Add the renderers name to the UI.
+	auto _add_renderer_entry = [this](const String &p_renderer_name, bool p_mark_overridden) {
```
clayjohn, line 7598: Is a lambda really necessary here? AFAIK our policy is to only allow them where necessary, or where they are so beneficial that it makes sense. This case doesn't look like it is necessary or adds anything over using a regular function. 

CC @RandomShaper since it was his suggestion

RandomShaper: Necessary I wouldn't say. Beneficial I would. I have already used lambdas in my PRs in this fashion, where they avoid code duplication without adding the need for a new member function.

clayjohn: Okay, our docs are less strict than what I thought:

> Lambdas should be used conservatively when they make code effectively faster or simpler, and do not impede readability. Please ask before using lambdas in a pull request.

(https://docs.godotengine.org/en/latest/contributing/development/cpp_usage_guidelines.html#lambdas)

So I'd agree its fine to use here since you support doing so. 

My personal opinion would be to never use lambda's unless necessary. But don't let me stand in the way of progress :stuck_out_tongue:

clayjohn (approved):
Everything looks good to me. 

I personally don't like introducing more lambdas into the codebase, but this one comes with Pedro's seal of approval :)

editor/editor_node.cpp
akien-mga: Not too fond of that lambda either, for me the policy means that they should only be used where they are absolutely a net positive (i.e. they enable doing something that a regular member function wouldn't).

Especially here as it's impacting a member variable (`renderer`) and not self-contained, I think this would be preferable as a member function.

RandomShaper: I hope we can discuss guidelines on lambdas at some point. As I see it, the benefit is great in cases like this: no need to touch the header (less clutter, faster iteration) and better readability because of avoiding out-of-context code ("Why does this class need that function?"). For the same reason, we could forbid using other C++ features, as there's always another way, and trim it down to C (even without classes!). I'm leaving these thoughts here so they install and start doing their job in your subconscious minds. 😃 For now, I understand if the current rule explicitly favors conservatism, my suggestion would be breaking it and this is not the right place to re-discuss it.

jsjtxietian:
Updated the code to use member function instead of lambda, squashed the commit.

clayjohn (approved):

YuriSizov (approved):
Editor parts look fine to me, but we'll probably clean them up in the future.

YuriSizov:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

PR #85390, opened 2023-11-26
UndoRedo: Fix `MERGE_ALL` commit from repeating actions

### LINKED ISSUE #26118: Possible bug with UndoRedo's MERGE_ALL mode
**Godot version:**
ced77db

**Issue description:**
It seems that there's a bug with `UndoRedo`'s `MERGE_ALL` mode, in that it repeats instructions if actions are committed fast enough. In the example bellow, if pressing "ui_accept" (enter or space) with long intervals, it works fine. Short ones will make "pressed" be printed twice.

```
var undo := UndoRedo.new()

func _input(event: InputEvent) -> void:
	if not event.is_action_pressed("ui_accept"):
		return
	
	undo.create_action("action", UndoRedo.MERGE_ALL)
	undo.add_do_method(self, "test")
	undo.commit_action()

func test() -> void:
	print("pressed")
```

TheSofox:
Fixes #26118

Committing an UNDO action will run the action. This is fine if it's set to not merge or only merge the ends, but with MERGE_ALL it meant that the action could be merged into previous actions, then the operations from ALL the merged actions would be run, instead of from the most recent action.

This pull request fixes this by keeping a record (merge_total) of the amount of operations that are in the current action that the new action is being merged into. Then, when the action is committed. This variable is used to skip over operations until we get to the ones that only the newest action added.

### DIFF
--- a/core/object/undo_redo.cpp
+++ b/core/object/undo_redo.cpp
@@ -71,7 +71,14 @@ bool UndoRedo::_redo(bool p_execute) {
 	}
 
 	current_action++;
-	_process_operation_list(actions.write[current_action].do_ops.front(), p_execute);
+
+	List<Operation>::Element *start_doops_element = actions.write[current_action].do_ops.front();
+	while (merge_total > 0 && start_doops_element) {
+		start_doops_element = start_doops_element->next();
+		merge_total--;
+	}
+
+	_process_operation_list(start_doops_element, p_execute);
 	version++;
 	emit_signal(SNAME("version_changed"));
 
@@ -104,6 +111,12 @@ void UndoRedo::create_action(const String &p_name, MergeMode p_mode, bool p_back
 				}
 			}
 
+			if (p_mode == MERGE_ALL) {
+				merge_total = actions.write[current_action + 1].do_ops.size();
+			} else {
+				merge_total = 0;
+			}
+
 			actions.write[actions.size() - 1].last_tick = ticks;
 
 			// Revert reverse from previous commit.
@@ -121,6 +134,7 @@ void UndoRedo::create_action(const String &p_name, MergeMode p_mode, bool p_back
 			actions.push_back(new_action);
 
 			merge_mode = MERGE_DISABLE;
+			merge_total = 0;
 		}
 	}
 
--- a/tests/core/object/test_undo_redo.h
+++ b/tests/core/object/test_undo_redo.h
@@ -1,0 +1,202 @@
+/**************************************************************************/
+/*  test_undo_redo.h                                                      */
+/**************************************************************************/
+/*                         This file is part of:                          */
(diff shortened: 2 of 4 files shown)


core/object/undo_redo.cpp
```diff
@@ -71,7 +71,13 @@ bool UndoRedo::_redo(bool p_execute) {
 	}
 
 	current_action++;
-	_process_operation_list(actions.write[current_action].do_ops.front(), p_execute);
+	
+	List<Operation>::Element *start_doops_element = actions.write[current_action].do_ops.front();
+	while (merge_total-- > 0 && start_doops_element) {
```
AThousandShips, line 76: We prefer not having decrement or increment in a statement like this, it hurts readability, please separate it

TheSofox: Done

TheSofox:
I've added Unit Tests that prove the bug has been fixed and test one or two other parts of `UndoRedo` for good measure.

tests/core/object/test_undo_redo.h
```diff
@@ -0,0 +1,211 @@
...
+
+void increment_property_action(UndoRedo *undo_redo, const String &name, _TestUndoRedoObject *test_object, int value, UndoRedo::MergeMode merge_mode = UndoRedo::MERGE_DISABLE) {
+	undo_redo->create_action(name, merge_mode);
+
+	Variant val = Variant(value);
+	const Variant *argptrs[1];
+	argptrs[0] = &val;
+
+	undo_redo->add_do_method(Callable(test_object, "add_to_property").bindp(argptrs, 1));
+	undo_redo->add_undo_method(Callable(test_object, "subtract_from_property").bindp(argptrs, 1));
```
KoBeWi, line 78: You can use `callable_mp()` here. There is no reason to bind these methods.

tests/core/object/test_undo_redo.h
```diff
@@ -0,0 +1,211 @@
...
+	undo_redo->commit_action();
+}
+
+TEST_CASE("[UndoRedo] Simple Property UndoRedo") {
+	GDREGISTER_CLASS(_TestUndoRedoObject);
+	UndoRedo *undo_redo = memnew(UndoRedo());
+
+	_TestUndoRedoObject *test_object = memnew(_TestUndoRedoObject());
+
+	CHECK(test_object->get_property() == 0);
```
KoBeWi, line 89: This check looks redundant. It's a test object which was just created, so obviously the value will be correct.

KoBeWi (approved):

tests/core/object/test_undo_redo.h
TheSofox: Done. Both good suggestions.

akien-mga:
Thanks!

PR #85393, opened 2023-11-26
Fix `PhysicsShapeQueryParameters3D.motion` type wrongly set to Vector2

SomeRanDev:
Maybe I'm missing something, but according to the [getter/setter](https://github.com/godotengine/godot/blob/master/servers/physics_server_3d.h#L903) and [documentation](https://docs.godotengine.org/en/stable/classes/class_physicsshapequeryparameters3d.html), this should be a Vector3.

### DIFF
--- a/misc/extension_api_validation/4.2-stable.expected
+++ b/misc/extension_api_validation/4.2-stable.expected
@@ -14,3 +14,10 @@ Validate extension JSON: Error: Field 'classes/TileData/methods/get_navigation_p
 Validate extension JSON: Error: Field 'classes/TileData/methods/get_occluder/arguments': size changed value in new API, from 1 to 4.
 
 Added optional argument to get_navigation_polygon and get_occluder to specify a polygon transform.
+
+
+GH-85393
+--------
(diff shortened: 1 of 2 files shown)


rburing (approved):

AThousandShips:
Need to add compatibility notice for this change, see `misc/extension_api_validation
/4.1-stable.expected`

SomeRanDev:
Ahhh, thank you for the thorough explanation! Sorry for the trouble. 😅 

> And once 4.2 has been released a 4.2 version of that will be added and you can add it to that

So I shouldn't add it to the `4.1-stable.expected` right now?

AThousandShips:
You can go get the CI finished, and then move it, or wait until after the new one has been added 🙂

YuriSizov:
You should be able to update your PR now, the required file has been added. 🙃

akien-mga (approved):
Rebased myself to add the compatibility statement.

Technically, this breaks compatibility AFAIK, but the previous binding was unusable.

akien-mga:
Thanks!

SomeRanDev:
Oh wow! Sorry, it's been a busy couple weeks and this completely slipped my mind. 🙇🙇
Thank you so much for putting together the final touches!!

PR #85399, opened 2023-11-26
TAA Improvements

### LINKED ISSUE #81709: TAA Bad Quality
### Godot version

4.1.1 Stable

### System information

Windows 11- 4.1.1 Stable - intel 6200U nvidia 940M - Vulkan Forward +

### Issue description

I don't know why, but I couldn't get good results from TAA like other engines (unreal and unity)
The movement completely destroys TAA !

If I only use TAA (not using MSAA as an aid) the anti-aliasing disappears completely during camera movement.
This picture was taken while I was moving the camera, you can clearly see that anti-aliasing is not working well.

image

### Steps to reproduce

image

https://github.com/LegionGames/FirstPersonController

mrjustaguy:
Resolves https://github.com/godotengine/godot/issues/81709#issuecomment-1722183283 points 2,3 and 5, very slightly 4. and point 1 is significantly improved

The video is too big for github, so there's a video on rocket chat that showcases me playing with the Improved TAA:

https://chat.godotengine.org/channel/rendering?msg=jLP3FxSt5joaNdvNq

The old Disocclusion Logic was totally removed as it was just broken. For instance, a Pixel cannot be half way disoccluded. it either is or isn't, something the logic doesn't realize. Further more, the get closest velocity gave straight up broken results, and the deviation calculation was trying to be smart, giving very hard to debug results, without actually benefiting the image.

In the end I've figured out a much better way of reducing ghosting that doesn't need to even take the disocclusion information into account for ghostless results, as the history clipping deals with that, though at the cost of having to choose between a slight jitter or ghosting. the default shouldn't have any noticable ghosting, and jitter only on highly aliased samples.

### DIFF
--- a/servers/rendering/renderer_rd/shaders/effects/taa_resolve.glsl
+++ b/servers/rendering/renderer_rd/shaders/effects/taa_resolve.glsl
@@ -164,25 +164,6 @@ void depth_test_min(uvec2 pos, inout float min_depth, inout uvec2 min_pos) {
 	}
 }
 
-// Returns velocity with closest depth (3x3 neighborhood)
-void get_closest_pixel_velocity_3x3(in uvec2 group_pos, uvec2 group_top_left, out vec2 velocity) {
-	float min_depth = 1.0;
-	uvec2 min_pos = group_pos;
-
-	depth_test_min(group_pos + kOffsets3x3[0], min_depth, min_pos);
-	depth_test_min(group_pos + kOffsets3x3[1], min_depth, min_pos);
-	depth_test_min(group_pos + kOffsets3x3[2], min_depth, min_pos);
-	depth_test_min(group_pos + kOffsets3x3[3], min_depth, min_pos);
-	depth_test_min(group_pos + kOffsets3x3[4], min_depth, min_pos);
-	depth_test_min(group_pos + kOffsets3x3[5], min_depth, min_pos);
-	depth_test_min(group_pos + kOffsets3x3[6], min_depth, min_pos);
-	depth_test_min(group_pos + kOffsets3x3[7], min_depth, min_pos);
-	depth_test_min(group_pos + kOffsets3x3[8], min_depth, min_pos);
-
-	// Velocity out
-	velocity = imageLoad(velocity_buffer, ivec2(group_top_left + min_pos)).xy;
-}
-
 /*------------------------------------------------------------------------------
 							  HISTORY SAMPLING
 ------------------------------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


Alex2782:
Godot v4.2.rc (17677a577) - macOS 13.6.1 - Vulkan (Forward+) - integrated Apple M1 - Apple M1 (8 Threads)

 **Demo:** [3d/antialiasing](https://github.com/godotengine/godot-demo-projects/tree/master/3d/antialiasing)


**master branch**

https://github.com/godotengine/godot/assets/41921395/404d7ddb-c072-4e8f-b55d-f6cb4abae57d

**this PR** (it's too blurry, and the ropes flicker on the ship )

https://github.com/godotengine/godot/assets/41921395/fdf4c675-f32f-4f9b-b48c-c009cb7e962c

mrjustaguy:
This is expected, The old implementation worked great when everything is static, but the moment you start moving, all hell breaks loose....

AThousandShips:
Even areas far from things that are moving? As far as I can see from the video the ropes aren't moving (there's nothing animating them)

mrjustaguy:
In the 1st one you can see clear aliasing on the cubes in motion, which isn't there with this PR, though this PR is slightly more jittery and blurrier, but again this is to be expected.. The results are more akin to FSR2 in terms of AA quality, as that jitters a bit too, though it isn't as blurry due to sharpening that it does.

Alex2782:
oh ok, it's about the performance. There is a significantly longer delay in the master branch as soon as "taa" is activated. (approx. 5 seconds). This PR is significantly faster under 1 second.

mrjustaguy:
In the meantime during testing while trying to further improve stability by reworking the disocclusion logic (that has been removed for this PR due to how broken it currently is and is the reason behind half the issues TAA currently suffers from) I've found that this implementation suffers from super ghosts in high contrast areas, very much like FSR2, just much, MUCH worse, and I've been unable to work out a solution for fixing them... 

Everything resulted in one of two outcomes:
1) Super Ghosting Issue persisted
2) ALL of the Old issues persisted to a high degree

Given that when this PR works fine it looks like a slightly worse FSR2 implementation, at half the cost, but has cases where it breaks much more significantly and doesn't have the benefit of being an upscaler, and I'm out of ideas how to fix the introduced breakages, I'm closing this PR as the current TAA behavior atleast acts significantly differently to FSR2 visually, and half a dozen minor issues are better vs a couple of moderate visual issues, as they stand out more when they occur, IMO.

PR #85420, opened 2023-11-27
Automatically detect screen sub-pixel layout and allow overriding it for viewports/windows.

bruvzg:
- Adds support for per-viewport sub-pixel layout override (sub-pixel AA only usable in non-scaled and non-rotated viewports).
- Adds automatically screen layout detect (on Linux/X11 and Windows using ClearType settings).
- Implements `screen_get_orientation` on desktop platforms.


Calinou:
> Is the sub-pixel rendering mode dynamic? That is, can we change it at runtime? If that's the case, perhaps we might enjoy from some event like the DPI change one to adjust depending on the current position.

Both Windows and Linux allow you to change the font rendering mode without restarting, but there's no guarantee that apps will follow until restarted.

bruvzg:
> What happens if a window is between two differently laid out displays?

Usually it's considered to be on display where most of the window is. I have not seen anything that's handling it as a special case, so I do not think we should bother.

> Is the sub-pixel rendering mode dynamic? That is, can we change it at runtime?

It is dynamic, and can be changed for each draw call. When proper DPI scaling is implemented it would make sense to adjust it on change (but currently it's not).

Calinou (commented):
Tested locally on Linux, it sometimes works as expected.

**Testing project (updated 2025-02-12):** [test_pr_85420.zip](https://github.com/user-attachments/files/18769595/test_pr_85420.zip)

With the following font configuration in KDE, LCD pixel antialiasing is applied, even though it shouldn't (subpixel optimization is set to **None** instead of the default **RGB**):

Screenshot_20231206_235202

Screenshot_20231206_235212

What's interesting is that if you set it from **RGB** to **None** in KDE and click **Apply**, it will work the next time you run the project, but only once. Subsequent runs will revert to RGB being mistakenly used.

akien-mga:
Judging by the changes, this should also happen to fix this warning we get when running the test suite (which initializes ThemeDB without the project settings part):
```
WARNING: Property not found: 'gui/theme/lcd_subpixel_layout'.
     at: get_setting_with_override (core/config/project_settings.cpp:380)
```

Calinou (commented):
Tested locally on Windows 11 24H2 (rebased on top of `master` https://github.com/godotengine/godot/commit/f418603522ebda0e00dd1e39646ceb42b8b62055), it sometimes works as expected.

Changing screen orientation then restarting the project will sometimes use the correct subpixel antialiasing mode, but I could only get VRGB to work (and only sporadically). Usually, it will just use HRGB regardless of screen orientation.

ClearType setting changes are not detected by Godot (although I didn't try logging out and back in), even though apps like Firefox update instantly when toggling ClearType (even while you are still in its configuration wizard).

doc/classes/DisplayServer.xml
```diff
@@ -1161,6 +1161,14 @@
 				Returns the screen's size in pixels. See also [method screen_get_position] and [method screen_get_usable_rect].
 			</description>
 		</method>
+		<method name="screen_get_subpixel_layout" qualifiers="const">
+			<return type="int" enum="DisplayServer.ScreenSubpixelLayout" />
+			<param index="0" name="screen" type="int" default="-1" />
+			<description>
+				Returns sub-pixel structure of the screen.
```
Calinou, line 1168: Does the value returned here depend on screen orientation? I'd mention this in the documentation.

clayjohn (commented):
Why is this going through the RenderingServer? Right now its doing an unnecessary sync every time we request the subpixel layout. Since the RenderServer doesn't care about the layout, the data shouldn't be stored in the RenderingServer

PR #85430, opened 2023-11-27
Expose RenderingServer methods to get rendering driver and method name

### LINKED ISSUE #10609: Expose methods to get the current rendering driver/method in use

A streaming overlay app

The OS class has a RenderingDriver enum that is not used anywhere. I'm looking for a get_rendering_driver() method in the docs, but it's either undocumented or not in Godot.

URL to the documentation page: https://docs.godotengine.org/en/stable/classes/class_os.html#enum-os-renderingdriver

Original issue in documentation: https://github.com/godotengine/godot-docs/issues/9851

Adding a function to detect the current rendering driver so it returns the RenderingDriver enum. This is especially useful when using `fallback_to_vulkan` or `fallback_to_d3d12` in Project Settings, so the function checks if the rendering driver fell back.

Im not familiar with C++, but I think there's a variable that holds the current renderer string name. It should be a matter of exposing the variable to a Godot class or something.

It also has to be included in the documentation.

Unlikely, since it is not an accessible value for Godot.

Has to be core, it is an unused enum.

Calinou:
This is useful for troubleshooting purposes and [debug menus](https://github.com/godot-extended-libraries/godot-debug-menu), as it obeys automatic fallbacks and CLI arguments unlike when reading the project settings.

This also improves related documentation in ProjectSettings.

PS: `--headless` doesn't use the `dummy` driver that is listed in the list of valid rendering drivers when using `--rendering-driver list`. Is that expected? For this reason, I didn't list it in the documentation.

```
bin/godot.linuxbsd.editor.x86_64 --path /tmp/4/ --headless
Godot Engine v4.2.rc.custom_build.5df986796 - https://godotengine.org

Rendering driver: vulkan
Rendering method: forward_plus
```

- This closes https://github.com/godotengine/godot-proposals/issues/10609.
- See also https://github.com/godotengine/godot/pull/85387.

### DIFF
--- a/servers/rendering_server.cpp
+++ b/servers/rendering_server.cpp
@@ -2067,6 +2067,16 @@ void RenderingServer::_particles_set_trail_bind_poses(RID p_particles, const Typ
 	particles_set_trail_bind_poses(p_particles, tbposes);
 }
 
+String RenderingServer::get_current_rendering_driver_name() const {
+	// Needs to remain in OS, since it's actually OS that interacts with it, but it's better exposed here.
+	return ::OS::get_singleton()->get_current_rendering_driver_name();
+}
+
+String RenderingServer::get_current_rendering_method() const {
+	// Needs to remain in OS, since it's actually OS that interacts with it, but it's better exposed here.
+	return ::OS::get_singleton()->get_current_rendering_method();
+}
+
 Vector<uint8_t> _convert_surface_version_1_to_surface_version_2(uint64_t p_format, Vector<uint8_t> p_vertex_data, uint32_t p_vertex_count, uint32_t p_old_stride, uint32_t p_vertex_size, uint32_t p_normal_size, uint32_t p_position_stride, uint32_t p_normal_tangent_stride) {
 	Vector<uint8_t> new_vertex_data;
 	new_vertex_data.resize(p_vertex_data.size());
@@ -3420,6 +3430,9 @@ void RenderingServer::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_video_adapter_vendor"), &RenderingServer::get_video_adapter_vendor);
 	ClassDB::bind_method(D_METHOD("get_video_adapter_type"), &RenderingServer::get_video_adapter_type);
 	ClassDB::bind_method(D_METHOD("get_video_adapter_api_version"), &RenderingServer::get_video_adapter_api_version);
+
+	ClassDB::bind_method(D_METHOD("get_current_rendering_driver_name"), &RenderingServer::get_current_rendering_driver_name);
+	ClassDB::bind_method(D_METHOD("get_current_rendering_method"), &RenderingServer::get_current_rendering_method);
 
 	ClassDB::bind_method(D_METHOD("make_sphere_mesh", "latitudes", "longitudes", "radius"), &RenderingServer::make_sphere_mesh);
 	ClassDB::bind_method(D_METHOD("get_test_cube"), &RenderingServer::get_test_cube);
--- a/doc/classes/ProjectSettings.xml
+++ b/doc/classes/ProjectSettings.xml
@@ -2753,6 +2753,8 @@
 			[b]Forward Plus[/b]: High-end renderer designed for Desktop devices. Has a higher base overhead, but scales well with complex scenes. Not suitable for older devices or mobile.
 			[b]Mobile[/b]: Modern renderer designed for mobile devices. Has a lower base overhead than Forward Plus, but does not scale as well to large scenes with many elements.
 			[b]GL Compatibility[/b]: Low-end renderer designed for older devices. Based on the limitations of the OpenGL 3.3/ OpenGL ES 3.0 / WebGL 2 APIs.
+			This can be overridden using the [code]--rendering-method &lt;method&gt;[/code] command line argument.
+			[b]Note:[/b] The actual rendering method may be automatically changed by the engine as a result of a fallback, or a user-specified command line argument. To get the actual rendering method that is used at runtime, use [method RenderingServer.get_current_rendering_method] instead of reading this project setting's value.
 		</member>
 		<member name="rendering/renderer/rendering_method.mobile" type="String" setter="" getter="" default="&quot;mobile&quot;">
 			Override for [member rendering/renderer/rendering_method] on mobile devices.
@@ -2776,7 +2778,8 @@
 			Depending on the complexity of scenes, this value may be lowered or may need to be raised.
 		</member>
 		<member name="rendering/rendering_device/driver" type="String" setter="" getter="">
-			Sets the driver to be used by the renderer when using a RenderingDevice-based renderer like the clustered renderer or the mobile renderer. This property can not be edited directly, instead, set the driver using the platform-specific overrides.
+			Sets the driver to be used by the renderer when using a RenderingDevice-based renderer like the clustered renderer or the mobile renderer. This property can't be edited directly. Instead, set the driver using the platform-specific overrides. This can be overridden using the [code]--rendering-driver &lt;driver&gt;[/code] command line argument.
+			[b]Note:[/b] The actual rendering driver may be automatically changed by the engine as a result of a fallback, or a user-specified command line argument. To get the actual rendering driver that is used at runtime, use [method RenderingServer.get_current_rendering_driver_name] instead of reading this project setting's value.
 		</member>
 		<member name="rendering/rendering_device/driver.android" type="String" setter="" getter="">
 			Android override for [member rendering/rendering_device/driver].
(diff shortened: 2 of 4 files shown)


akien-mga:
Exposing this makes sense to me, but I wonder why we put it in `OS` originally. This isn't really OS specific, but more something that reflects the engine configuration.

Wouldn't it be more appropriate in `Engine`, or even in `ProjectSettings`?

Calinou:
> Exposing this makes sense to me, but I wonder why we put it in OS originally. This isn't really OS specific, but more something that reflects the engine configuration.

I've moved the exposed method to Engine in a separate commit (can be squashed before merging). I chose to keep the internal method in OS to avoid any risk of regressions (the driver has OS-specific interactions in particular, with ANGLE, Direct3D, etc).

SheepYhangCN:
Should `Metal` be added into the document since its support was added in #88199 ?

Calinou:
> Should `Metal` be added into the document since its support was added in #88199 ?

Done (I also added a mention of `d3d12`);

huwpascoe:
Should this be removed?
https://docs.godotengine.org/en/stable/classes/class_os.html#enum-os-renderingdriver
I can't find anything that actually references it.

SheepYhangCN:
> Should this be removed?
> https://docs.godotengine.org/en/stable/classes/class_os.html#enum-os-renderingdriver
> I can't find anything that actually references it.

This enum is useless for now, there is nothing using this.
So yeah this should be remove since the method will return string directly.

huwpascoe (commented):
Probably this enum should be deleted
https://docs.godotengine.org/en/stable/classes/class_os.html#enum-os-renderingdriver

And then it's good

adamscott (approved):
Seems fine to me!

akien-mga:
> I've moved the exposed method to Engine in a separate commit (should be squashed before merging). It still works after testing.
> 
> I chose to keep the internal method in OS to avoid any risk of regressions and preserve compatibility. The rendering driver has OS-specific interactions in particular, with ANGLE, Direct3D, etc.

The problem with that is that we now have a discrepancy between the core C++ API (method in `OS`) and the scripting API (method in `Engine`). Which is a bit confusing for people using both, but also breaks the expectation for `godot-cpp` that the C++ extension API is as close as possible to the C++ one, allowing a (relatively) easy path between C++ modules and C++ extensions.

mubinulhaque:
> So if it can't be moved from OS to Engine in core, I'd keep it exposed in OS to be consistent. But I don't see a problem with moving it from OS to Engine, rendering code can include core/config/engine.h just fine, no?

Is this the only merge blocker? Testing games is getting harder, now that the RD renderers support multiple drivers. For example, a common problem is that some GPUs come with bad Vulkan drivers, but good D3D12 drivers. Being able to know the currently used rendering driver is important for end-users to know how best to play our games.

akien-mga:
> The problem with that is that we now have a discrepancy between the core C++ API (method in `OS`) and the scripting API (method in `Engine`). Which is a bit confusing for people using both, but also breaks the expectation for `godot-cpp` that the C++ extension API is as close as possible to the C++ one, allowing a (relatively) easy path between C++ modules and C++ extensions.

These are the changes needed to keep things consistent: https://github.com/akien-mga/godot/commit/3a1117c87e0074ac6f135c8df2e5d1235a1548d9

WDYT @clayjohn?

Alternatively, we can keep it in OS and expose it in OS to scripting.

clayjohn:
We discussed in the rendering meeting. We think that we should investigate exposing it through the RenderingServer as that is the most natural place for these methods to exist. 

While the indirection is awkward, it should be much nicer from a user point of view

Calinou:
> We discussed in the rendering meeting. We think that we should investigate exposing it through the RenderingServer as that is the most natural place for these methods to exist.

Done.

While I'm thinking about it, I wonder if I should name the exposed methods as such:

- `RenderingServer.get_current_rendering_driver_name()`
- `RenderingServer.get_current_rendering_method_name()`

Right now, we have the following:

- `RenderingServer.get_current_rendering_driver_name()`
- `RenderingServer.get_current_rendering_method()`

We don't have exposed enums for the rendering drivers and methods, but we should probably add one for the rendering method and expose both enums. This would allow creating and exposing the following enum-based methods:

- `RenderingServer.get_current_rendering_driver()`
- `RenderingServer.get_current_rendering_method()`

This would allow users to make comparisons that are not string-based, therefore not being prone to typos (while also allowing for autocompletion to work).

What do you think?

akien-mga:
TIWAGOS and I'd let the rendering team decide, but yes I feel exposing an enum version of those methods might make sense. And our internal code should likely also be refactored to use the enums, the string comparisons we have throughout the rendering code are a bit awkward.

SheepYhangCN:
Is it worth to make it enum?
If we made it enum, then it should replace all of the original strings, otherwise I don't think it is worth to do
And it will be a lot of work to replace all of it

huwpascoe:
Regardless of the outcome, there's an existing enum that despite being exposed, doesn't seem to be referenced anywhere:
https://docs.godotengine.org/en/stable/classes/class_os.html#enum-os-renderingdriver

Calinou:
> Regardless of the outcome, there's an existing enum that despite being exposed, doesn't seem to be referenced anywhere: [docs.godotengine.org/en/stable/classes/class_os.html#enum-os-renderingdriver](https://docs.godotengine.org/en/stable/classes/class_os.html#enum-os-renderingdriver)

Ah, that's what I meant actually. The enum *is* exposed, it's just not used anywhere.

clayjohn:
We discussed using enums for a very long time during the initial 4.0 rewrite. I was in favour of them at the time. But the consensus was that we wanted to eventually allow people to add new rendering methods and rendering drivers through GDextensions. Having a closed enum would make that very challenging.

Chubercik:
I'd like to chime in if I may: I was researching whether or not it was possible to check the rendering method idiomatically, because I wanna use a feature unavailable in the `Compatibility` renderer on `Mobile` and `Forward+` builds, but also have a fallback to a feature that's supported, to have feature-complete web builds. Am I right to believe this isn't achievable as of `4.3` without something like supporting two scene duplicates at once, one for desktop/mobile and one for web, but with this PR it's possible to spawn different nodes depending on the platform programmatically?

Calinou:
> Am I right to believe this isn't achievable as of 4.3

This can be achieved in 4.3 with the following code:

```gdscript
if ProjectSettings.get_setting_with_override("rendering/renderer/rendering_method") == "gl_compatibility":
	print("Using Compatibility")
else:
	print("Using Forward+/Mobile")
```

The web platform always uses `gl_compatibility` since it has a `.web` override for the Rendering Method project setting.

The downside of this approach is that it doesn't take CLI argument overrides into account, nor automatic fallbacks that may happen on unsupported GPUs (but this isn't implemented in 4.3 anyway).

Chubercik (approved):

AThousandShips (approved):

servers/rendering_server.cpp
```diff
@@ -2067,6 +2067,16 @@ void RenderingServer::_particles_set_trail_bind_poses(RID p_particles, const Typ
 	particles_set_trail_bind_poses(p_particles, tbposes);
 }
 
+String RenderingServer::get_current_rendering_driver_name() const {
+	// Needs to remain in OS, since it's actually OS that interacts with it, but it's better exposed here.
+	return ::OS::get_singleton()->get_current_rendering_driver_name();
+}
```
akien-mga, line 2073: I would still challenge this comment. In what way does `OS` interact with this? It just seems like the value was parked there for ease of access, but it could as well be in `RenderingServer` a priori. It's mostly set from `DisplayServer` implementations which should have access to both singletons.

I'll have a look to confirm.

akien-mga: I've confirmed these can be moved to `RenderingServer` without problems with https://github.com/akien-mga/godot/commit/5ea9478749b8a77ce6e5042cb914cb88e6ff437e.
So I think we should do it properly (amending in my commit should do the trick).

And also moved `OS::_is_gles_over_gl` that similarly doesn't need to be in `OS`, and removed the getter for `_display_driver_id`, was never used since it was added in 2020.

akien-mga (changes requested):
I still insist we do this properly by moving the C++ members to RenderingServer too.

servers/rendering_server.cpp
akien-mga: Actually some CI tests are failing with my branch. I'll look into it further.

akien-mga (approved):
I admit defeat for now, would need to look a bit more into why the editor crashes if the C++ getter is moved to RenderingServer. I still think this could be done, but requires more work, which could be done later.

Repiteo:
Thanks!

PR #85432, opened 2023-11-27
Fix GDScript thread-exit routine assuming thread-enter was called

RandomShaper:
Just as simple as that: the engine can create threads before even GDScript has been lifted up. `GDScript::thread_exit()` will be called regardless.

Fixes #85377.

@akien-mga: This is simple enough to be safe for 4.2. For 4.3 I'll submit a more comprehensive PR.

PR #85433, opened 2023-11-27
Lazily create GDScript thread lambda bookkeeping block

RandomShaper:
This is an extension to #85432 (which is included to conflicts are pre-solved), but for 4.3, given it's slightly less safe.

This avoids every thread that does not even run GDScript or doesn't use lambdas using memory for some bookkeeping they don't need.


RandomShaper:
Rebased and improved to use a paged allocator.

RandomShaper:
Superseded by #86569.

Issue #40784, opened 2020-07-28
Lod feature missing


alexwbc:
**Godot version:**
from 3.0 and beyond

**Issue description:**
Godot don't support LoD, Godot knows it while still show the vestigial menu in [GeometryInstance > Lod]: the option is there just for troll early adopters (who are mostly backfire in bad press on how "not ready" Godot is it)

**Steps to reproduce:**
Shock violently the lod_min_distance and lod_max_distance: watch Godot's icon laugh at you as nothing happen.

**Minimal reproduction project:**
create a new MeshInstance > Mesh > New Cube Mesh > GeometryInstance > Lod

Calinou:
I thought about hiding the LOD properties from the editor, but I'm not completely sure if it's a good idea. Third-party LOD assets are already able to make use of those to better integrate LOD systems.

alexwbc:
> I thought about hiding the LOD properties from the editor, but I'm not completely sure if it's a good idea.
A not working option, it's definitely not a good idea; to have the feature is more important, it's a critical one in term of performance; and performance is tightly tied with engine adoption.

> Third-party LOD assets are already able to make use of those properties to better integrate LOD systems with the rest of Godot.
Which will return lot of frustration from unwary/early adopters before realizing that either the feature is broken somewhere or not actually implemented (their asset? their pipeline, the engine itself?)

> We don't implement features to troll people in Godot. :wink: 
Not implementing features is to be expected, but leave the vestigial option there is to troll :sob:

fire:
There are still non-working LOD properties, but auto lod is implemented.

vlad0337187:
[it's told in tutorial](https://docs.godotengine.org/en/3.0/tutorials/3d/3d_performance_and_limitations.html?highlight=lod#level-of-detail-lod):
> GeometryInstance based objects have a visibility range that can be defined

I use Godot 3.2, but can't figure out, is it related for this issue or not.

Can currently GeometryInstance objects be hidden or not ?

fire:
I recommend we remove the old properties. Since the old lods never worked. I don't know how to design a place holder for a feature that doesn't work.

Calinou:
> I recommend we remove the old properties. Since the old lods never worked. I don't know how to design a place holder for a feature that doesn't work.

[...]

PR #53778, opened 2021-10-13
[3.x] Implement LOD range in GeometryInstance

Calinou:
Salvaged version of https://github.com/godotengine/godot/pull/20489. Thanks @Windfisch for the implementation :slightly_smiling_face: 

Similar to the one present in `master`, this provides a LOD system for GeometryInstance-derived nodes such as MeshInstance and MultiMeshInstance.

This LOD system currently doesn't obey shadow rendering, but it can still provide substantial performance improvements when used well. (To prevent multi-LOD objects from casting shadows over each other, only make one of the LOD levels cast shadows by configuring other meshes' Shadow Casting property to **Off**.)

On the other hand, this LOD system *should* be able to handle split screen, unlike the one currently present in `master`.

Coupled with the occlusion culling systems (portals/rooms or occluder spheres), this makes possible to achieve much greater performance in large detailed scenes.

**Testing project:** [test_lod_3.x.zip](https://github.com/godotengine/godot/files/7340633/test_lod_3.x.zip)

## Rationale

As you may have seen in <https://github.com/godotengine/godot/pull/20489>'s comments, the PR was closed since a different approach was adopted in `master`. The closing comment refers to the *automatic LOD system*, which can't be backported to `3.x` as it would require too many backwards-incompatible changes. The automatic LOD system is a good start as it requires no manual setup, but it's not suited to every use case and can't perform HLOD (merging several distant objects into a single draw call). 

However, the [*manual* (H)LOD system](https://github.com/godotengine/godot/pull/48847) can still be backported or reimplemented in a limited fashion in `3.x`, while still providing substantial performance improvements with some manual setup.

The existing LOD properties can be read by a Godot 4.0 converter to allow LOD to work automatically when porting a project from Godot 3.x to Godot 4.0.

## Preview

*The testing project has 1,134 boxes with 16,000 triangles each (20×20×20 subdivision).*

### No LOD

No LOD

*The video is stuttery due to how I recorded it, but it's much smoother in practice.*

https://user-images.githubusercontent.com/180032/137196355-719d8266-28fa-446a-9a83-2f42e7f02264.mp4

### With LOD

With LOD

*The video is stuttery due to how I recorded it, but it's much smoother in practice.*

https://user-images.githubusercontent.com/180032/137196368-e5d6e49c-b8b0-4a78-8e86-66e7943d4ad9.mp4

## TODO

[...]

PR #57196, opened 2022-01-25
GeometryInstance: Remove unimplemented LOD properties

akien-mga:
And remove matching unimplemented VisualServer functions.

Fixes #40784.

**Breaks compat** because it removes methods from the API, so any code that was using them will now raise an error. But these methods did nothing, so there should be no logic regression in user games, the only impact is that some users mistakenly calling these methods might have to remove these no-op calls from their scripts.

PR #85437, opened 2023-11-27
[3.x] Discrete Level of Detail

lawnjelly:
Add scene side discrete level of detail.

New `LOD` node for UI. `LODManager` added within `World` for automatically updating child visibilities based on distance from cameras.

Supersedes #53778
Fixes https://github.com/godotengine/godot/issues/40784

## Demo Projects
Examine in the editor while moving viewer. Run, and use cursors to move camera.
[testlod2.zip](https://github.com/godotengine/godot/files/14194713/testlod2.zip)

Run, and use cursors to move. Contains multiple LOD nodes.
[testlod3.zip](https://github.com/godotengine/godot/files/14194852/testlod3.zip)

https://github.com/godotengine/godot/assets/21999379/e2039203-049b-4cf2-8c0b-5057005fab1e

## Explanation
After exploring some fixes to #53778 I decided to try a scene side LOD system (like @Calinou 's addon).

_LOD Node showing 3 example `MeshInstances` to be used as LODs_
lod_node

_LOD Node properties_
hysteresis

_Spatial properties_
The old unused LOD properties are removed from `GeometryInstance`. The `lod_range` determines how much of the distance range is used by this LOD. So first child with 2 range will cover distance 0-2, second child with 5 range will cover distance 2-7, etc.
Children of these `Spatials` will also automatically be shown / hidden as a consequence of the scene tree, as the LODs are automatically selected.
 
log_range

## Advantages & Disadvantages

There are advantages and disadvantages to using either scene side or server side approach:

#### Advantages
* Can change LOD per camera
#### Disadvantages
* Changing LOD per camera doesn't play nice with shadows, especially OMNI and SPOTs, so you either have to deal with self shadowing artefacts or do user side workarounds.
* Only deals with displaying the final mesh, cannot throttle down animations, sound, script
* Hidden LODs still do significant processing (BVH, culling, pairing etc)
* All LOD calcs are done every frame.
* Calculations based on AABB centre may miscalculate overlap between different LODs, resulting in either flickering gaps or overlap (and consequently more processing used, when the objective is to increase performance!).
* Has to use new `visibility_parent` system in `VisualServer` to work around inability to use the hierarchy.

#### Advantages
* Can calculate a "best compromise" LOD based on the closest camera
* Shadows can use the LOD chosen and thus no self shadowing artefacts, and shadows also scale
* Can also throttle down animations, sound, scripts etc.
* Can totally turn off display at distance from the camera.
* Is very processing friendly, hidden LODs are removed from BVH and do no processing server side.
* Processing is done in round robin fashion with queues, so LOD processing is limited per frame.
* Can use the hierarchy to determine visibility - any child that is an ancestor of the LOD children will in turn be shown or hidden as per the normal scene tree behaviour, no special changes are required. Additionally, visibility change notifications can be used as normal to drive behaviour.

#### Disadvantages
* LOD chosen does not vary per camera

## Notes
* There seems to be significant demand for this in 3.x based on recent twitter threads (https://twitter.com/TheConceptBoy/status/1727003872202539151 )
* This is fairly simple but should be very flexible, would welcome further input.
* Removes the existing stubs in `VisualServer` and `GeometryInstance` as these make no sense with this PR.
* I may have a go at progressive LOD for 3.7 but this should do for a lot of basic use cases, and covered HLOD where multiple models are replaced by a single model at distance.
* The generating of the threshold distances is pretty inefficient here, but is enough to get started and can be made a lot more efficient with #82248 which adds `children_changed` signal (so the calculation can be pushed on demand instead of pulled every time).
* After writing support for Ortho cameras, I decided for now to simply disable calculations for ortho cameras until we have a use case. The use case does not seem convincing, and these are likely to cause edge case bugs as the distance is from the plane and not the camera.
* Cameras can turn on and off `affects_lod` property. Notably this is used for editor preview windows to prevent hidden windows affecting LOD.
* I removed the `lod_enabled` property in `GeometryInstance` as I realised it wasn't necessary. If users want to have nodes unaffected by LOD, they can either parent the `LOD` node to a `Spatial` and hang the nodes from this, _or_ use `Spatial` nodes as direct children of `LOD` (as these are not `GeometryInstances` they and their children will be unaffected by the `LOD` node). 
* There are 5 priority settings available in the `LOD` node properties to choose the update rate.
* The first LOD is always chosen during saving from the editor, to prevent unnecessary diffs in files.
* Updating LODs can be turned on and off in the editor via the `view` menu in the 3d view.

## Discussion
This only took me today to do as an exploratory look. Discrete LOD *can* be done using an addon, but there seems some interest in a core solution so I put this up as an option.

Note that this also won't deal sensibly with `MultiMeshes` (as these are dealt with as a block rather than performing LOD on each instance within the `MultiMesh`). If there is demand we could maybe add functionality to manage multimeshes, perhaps for vegetation, where otherwise drawcalls would be a bottleneck. These would probably have to be dealt with quite differently internally, although the user interface could probably be similar / shared.


scene/3d/camera.cpp
```diff
@@ -523,6 +525,7 @@ void Camera::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "frustum_offset"), "set_frustum_offset", "get_frustum_offset");
 	ADD_PROPERTY(PropertyInfo(Variant::REAL, "near", PROPERTY_HINT_EXP_RANGE, "0.01,8192,0.01,or_greater"), "set_znear", "get_znear");
 	ADD_PROPERTY(PropertyInfo(Variant::REAL, "far", PROPERTY_HINT_EXP_RANGE, "0.1,8192,0.1,or_greater"), "set_zfar", "get_zfar");
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "affects_lod"), "set_affects_lod", "get_affects_lod");
```
AThousandShips, line 528:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "affect_lod"), "set_affect_lod", "get_affect_lod");
```
I believe this should be active as in "it does affect LOD", otherwise it sounds like "check if it does affect LOD"

lawnjelly: Affects seems more natural to my ears (as native English speaker)? But I'd be happy to change it, there doesn't seem a lot in it.

It probably comes down to how you think of the phrase being used, it is quite subtle:
* correct - "Does the camera affect LOD?"
* incorrect "Does the camera affects LOD?"
* correct - "Camera affects LOD."
* incorrect - "Camera affect LOD."

AThousandShips: But here it's "make the camera affect LOD" IMO, it's either way but to me settings read as imperative 🙂

lawnjelly: Have changed. :+1:

doc/classes/GeometryInstance.xml
```diff
@@ -45,21 +45,9 @@
...
-		<member name="lod_min_distance" type="float" setter="set_lod_min_distance" getter="get_lod_min_distance" default="0.0">
-			The GeometryInstance's min LOD distance.
-			[b]Note:[/b] This property currently has no effect.
-		</member>
-		<member name="lod_min_hysteresis" type="float" setter="set_lod_min_hysteresis" getter="get_lod_min_hysteresis" default="0.0">
-			The GeometryInstance's min LOD margin.
-			[b]Note:[/b] This property currently has no effect.
+		<member name="lod_range" type="float" setter="set_lod_range" getter="get_lod_range" default="10.0">
+			Determines the threshold distance at which this node will be shown or hidden when this node is parented by a [LOD] node.
+			For example, a first child with a range of 2 will be shown from distance 0 to 2. A second child with a range of 5 will be shown from distance 2 to 7, etc.
```
Mickeon, line 50: I would normally suggest to surround every single value with code tags, but we need to make an exception here because it's really annoying to read.
```suggestion
			For example, a first child with a range of [code]2[/code] will be shown from distance 0 to 2. A second child with a range of [code]5[/code] will be shown from distance 2 to 7, etc.
```

doc/classes/LOD.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="LOD" inherits="Spatial" version="3.6" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
+	<brief_description>
+		Provides discrete level of detail.
+	</brief_description>
+	<description>
+		Automatically hides and shows children depending on the distance from the closest [Camera] to the [LOD] node.
+		Children are shown in order with the first children shown when closest to the [Camera], and the latter children shown when further away.
+		The threshold distances for changes are determined by [member GeometryInstance.lod_range].
+		For example, a first child with a range of 2 will be shown from distance 0 to 2. A second child with a range of 5 will be shown from distance 2 to 7, etc.
```
Mickeon, line 10: Ditto
```suggestion
		For example, a first child with a range of [code]2[/code] will be shown from distance 0 to 2. A second child with a range of [code]5[/code] will be shown from distance 2 to 7, etc.
```

doc/classes/LOD.xml
```diff
@@ -0,0 +1,27 @@
...
+	</tutorials>
+	<methods>
+	</methods>
+	<members>
+		<member name="hysteresis" type="float" setter="set_hysteresis" getter="get_hysteresis" default="1.0">
+			To prevent flickering at borders, a hysteresis distance will be added to threshold distances when the object is moving away from the [Camera].
+		</member>
+		<member name="lod_queue" type="int" setter="set_lod_queue" getter="get_lod_queue" default="0">
+			[LOD] nodes can choose to be added to any one of five queues, with 0 being lowest priority.
+			[b]Note:[/b] Higher priorities will react quicker but may take more CPU.
```
Mickeon, line 22:
```suggestion
			[LOD] nodes can choose to be added to one of five queues, with [code]0[/code] being lowest priority.
			[b]Note:[/b] Higher priorities will react quicker but may take more CPU time.
```

doc/classes/Camera.xml
```diff
@@ -137,6 +137,11 @@
 		</method>
 	</methods>
 	<members>
+		<member name="affect_lod" type="bool" setter="set_affect_lod" getter="get_affect_lod" default="true">
+			If [code]true[/code], the camera will be used to calculate the level of detail in [LOD] nodes.
+			Objects further from the camera will select lower levels of detail than those closer to the camera.
+			[b]Note:[/b] Cameras with [constant PROJECTION_ORTHOGONAL] will not affect level of detail.
```
Mickeon, line 143: May be good to say this property does nothing with orthogonal projection.
```suggestion
			[b]Note:[/b] This property has no effect if [member projection] is set to [constant PROJECTION_ORTHOGONAL].
```
Feel free to adjust.

Mickeon (commented):
Would be nice to see the viability of this for 3.x. There's no direct equivalent in 4.0, so it's difficult to judge.

_(I wholeheartedly appreciate removing the confusing properties that never had any effect, myself)_

AThousandShips:
Even though they don't do anything I'd say removing these properties is breaking compatibility, they still set values so nominally someone could be using them for their own implementation, right?

Mickeon:
To my dismay, yes. And they're not duds, either. They do set something in the VisualServer (even if it doesn't actually do anything).

lawnjelly:
> Even though they don't do anything I'd say removing these properties is breaking compatibility, they still set values so nominally someone could be using them for their own implementation, right?

See #57196 . We planned to remove these properties, but it was only delayed due to #53778 , which is superseded by this PR.

Mickeon:
It's been so long that the likelihood of these properties being used by _someone_ out there is... still low, but not impossible. The documentation was never even changed to state that they have been removed in the future.

lawnjelly:
It is possible it was used by e.g. an addon, but it's fairly simple to update an addon, and we shouldn't let it hold up core development imo. Banditing existing non-working properties for an addon must be considered "subject to change" and _caveat emptor_.

Leaving these non-working properties / functions in place _just in case_ is likely to create confusion if there is now working LOD functionality, which doesn't use them.

Mickeon:
Just because it _breaks compatibility_ doesn't mean this shouldn't be merged in my opinion, at least.

Calinou (commented):
Tested locally, it works as expected in both GLES3 and GLES2. Other than the UI nitpicks below, it seems to work pretty well.

I'll look into designing an icon for the LOD node.

**Testing project:** [test_lod.zip](https://github.com/godotengine/godot/files/14226592/test_lod.zip)

- When setting up a new LOD system, the visibility for LOD's child nodes in the scene tree dock is only updated when you save the scene. Before I saved the scene, all nodes were visible even though their distance was already configured:

image

- Child nodes' visibility changes while in the editor and when saving, which can be distracting if your node is expanded:

https://github.com/godotengine/godot/assets/180032/c66d641f-1bc6-4fdd-90f3-7da384d547f3

Maybe we should hide the visibility icon for children of a LOD node that inherit from Spatial, as there's no point in manually toggling visibility for those in the editor (changes will instantly be reverted).

- LOD update speed seems a bit slow when you have lots of instances. This might be an issue if your camera moves fast, as low-detail LODs will remain visible for a while after the camera gets close to objects. Is there a way to speed up the update rate (at the cost of increased CPU utilization)?

lawnjelly:
> I'll look into designing an icon for the LOD node.

That would be fantastic. :+1: 

> When setting up a new LOD system, the visibility for LOD's child nodes in the scene tree dock is only updated when you save the scene. Before I saved the scene, all nodes were visible even though their distance was already configured

I'll have a look at this, it's not stop ship, but could be a quality of life improvement.

> Child nodes' visibility changes while in the editor and when saving, which can be distracting if your node is expanded:
> Maybe we should hide the visibility icon for children of a LOD node that inherit from Spatial, as there's no point in manually toggling visibility for those in the editor (changes will instantly be reverted).

I'll have a look whether this is possible, no idea where the visibility icon is handled. Can also see pros and cons to this. An alternative (which I could easily put in) is a setting in the editor to turn on and off LOD in the editor, to prevent it becoming distracting.

> LOD update speed seems a bit slow when you have lots of instances. This might be an issue if your camera moves fast, as low-detail LODs will remain visible for a while after the camera gets close to objects. Is there a way to speed up the update rate (at the cost of increased CPU utilization)?

Yes, this is exactly what the `lod_queue` was for, which exists internally but is hidden from the UI .. maybe I should reinstate it. :thinking: I'll do this.

I was worried it would be overkill for users, but it does allow you to set exactly this update rate (by choice of 5 priority queues), and if you have noticed this so quickly, then users will no doubt find it useful.

lawnjelly:
> When setting up a new LOD system, the visibility for LOD's child nodes in the scene tree dock is only updated when you save the scene. Before I saved the scene, all nodes were visible even though their distance was already configured:

Ok I've fixed it up so now in the editor, if `view_level_of_detail` is switched on, if you change visibility manually of a node it will automatically override it (it detects the situation where the number of visible children is not 1, and does a full recalculate of LOD for that node).

This is good in some ways, but it's rather frustrating when you do try and change the visibility of the children and nothing happens, so I'm not entirely sure about it. Perhaps we can improve the user experience here somehow. Maybe an error warning in the output.. :man_shrugging: 

I've also added code to completely remove `LOD` node from the queues when it is hidden to save on processing. That was an oversight on my part.

Calinou:
An attempt at an icon (it aims to represent a low-detail object, i.e. a circle turned into some kind of hexagon): [icon_l_o_d.zip](https://github.com/godotengine/godot/files/14229968/icon_l_o_d.zip)

I've made the sides thicker so it looks more different from the Spatial icon.

image

lawnjelly:
> Here's an attempt at an icon 

Fantastic! Thanks. I'll add it to the PR.

I also managed to get the visibility icons to grey out in the scene tree dock. It actually looks pretty good now and is far simpler than I thought it would be, just a few lines of code.

Calinou (approved):
I've tested the latest revision of this PR and it works as expected.

There are still a few areas of improvement, but these can be left for future PRs:

- LOD nodes with no valid children or only 1 valid child could emit a node configuration warning, so you know that at least 2 valid child nodes are needed for LOD to have an effect. (A valid child node is one that extends from Spatial.)
- When initially loading a scene, LOD is progressively applied with the selected priority, which can make it obvious in certain cases:

https://github.com/godotengine/godot/assets/180032/0248fe85-5613-4446-95a3-3fb072573632

*Watch the video at 0.5× speed by right-clicking it to make it more noticeable.*

Perhaps LOD priority should be forced to its maximum value for a few frames when the engine initially loads. That said, polished games generally don't immediately boot in a playable scene (they generally hide things with loading screens), so it may be a non-issue.

doc/classes/LOD.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="LOD" inherits="Spatial" version="3.6" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
+	<brief_description>
+		Provides discrete level of detail.
+	</brief_description>
+	<description>
+		Automatically hides and shows children depending on the distance from the closest [Camera] to the [LOD] node.
+		Children are shown in order with the first children shown when closest to the [Camera], and the latter children shown when further away.
+		The threshold distances for changes are determined by [member Spatial.lod_range].
+		For example, a first child with a range of [code]2[/code] will be shown from distance 0 to 2. A second child with a range of [code]5[/code] will be shown from distance 2 to 7, etc.
```
Calinou, line 10:
```suggestion
		Automatically hides and shows [Spatial] children depending on the distance from the closest [Camera] to the [LOD] node. Child nodes that don't inherit from [Spatial] are ignored by the LOD system.
		Children are shown in order with the first children shown when closest to the [Camera], and the latter children shown when further away.
		The threshold distances for changes are determined by [member Spatial.lod_range].
		For example, a first child with a range of [code]2[/code] will be shown from distance 0 to 2. A second child with a range of [code]5[/code] will be shown from distance 2 to 7, etc.
```

editor/editor_node.cpp
```diff
@@ -1347,7 +1347,17 @@ void EditorNode::_find_node_types(Node *p_node, int &count_2d, int &count_3d) {
...
+	Ref<World> edited_world;
+
 	if (editor_data.get_edited_scene_root() != nullptr) {
+		// Allow a generic mechanism for the engine to make changes prior, and after saving.
+		if (editor_data.get_edited_scene_root()->get_tree() && editor_data.get_edited_scene_root()->get_tree()->get_root()) {
+			edited_world = editor_data.get_edited_scene_root()->get_tree()->get_root()->get_world();
+			if (edited_world.is_valid()) {
+				edited_world->notify_saving(true);
+			}
+		}
```
Calinou, line 1359: This hook could be really useful to have in editor plugins and `@tool` scripts, also in 4.x :slightly_smiling_face: 

There are multiple instances where I've wanted to make editor-only changes that aren't serialized to scene files outside of LOD.

lawnjelly:
> LOD nodes with no valid children or only 1 valid child could emit a node configuration warning, so you know that at least 2 valid child nodes are needed for LOD to have an effect. (A valid child node is one that extends from Spatial.)
> When initially loading a scene, LOD is progressively applied with the selected priority, which can make it obvious in certain cases:

Yes, as you say, easy to do in follow up. :+1:

editor/editor_node.cpp
lawnjelly: Yes I was surprised there wasn't something like this already. Although it may have to be done carefully in some cases to avoid race conditions.

akien-mga:
Thanks!

Zireael07:
Woot woot, issue closed after 4 years!

PR #85438, opened 2023-11-27
Use f-string with build progress display

EdwardBetts:


AThousandShips:
What's the reason for this change? What issue does it solve/improvement does it provide? Please add details to the description of the PR to make this clear

Calinou:
I think if we decide to switch to f-strings, we should do this on the whole buildsystem code.

That said, this particular example isn't very convincing as it looks less readable at a glance than the current form.

PR #85448, opened 2023-11-27
Update `blender_path` behavior to require exact path to executable instead of trying to guess it

mxaddict:
This PR updates the setting the `blender3_path` setting to be called `blender_path`

As well as changes it to require a full path to the executable (I and many others have been confused by this setting as it requires a DIR that contains a `blender` or `blender.exe` or on mac a `Blender` executable to be in the provided path

This change allows users to set the value to an exact path, for example `/usr/bin/blender-4.1` if you have multiple blender versions this is a must have feature.

### DIFF
--- a/modules/gltf/editor/editor_scene_importer_blend.cpp
+++ b/modules/gltf/editor/editor_scene_importer_blend.cpp
@@ -55,20 +55,7 @@
 #endif
 
 static bool _get_blender_version(const String &p_path, int &r_major, int &r_minor, String *r_err = nullptr) {
-	String path = p_path;
-#ifdef WINDOWS_ENABLED
-	path = path.path_join("blender.exe");
-#else
-	path = path.path_join("blender");
-#endif
-
-#if defined(MACOS_ENABLED)
-	if (!FileAccess::exists(path)) {
-		path = p_path.path_join("Blender");
-	}
-#endif
-
-	if (!FileAccess::exists(path)) {
+	if (!FileAccess::exists(p_path)) {
 		if (r_err) {
 			*r_err = TTR("Path does not contain a Blender installation.");
 		}
@@ -77,7 +64,7 @@ static bool _get_blender_version(const String &p_path, int &r_major, int &r_mino
 	List<String> args;
 	args.push_back("--version");
 	String pipe;
-	Error err = OS::get_singleton()->execute(path, args, &pipe);
+	Error err = OS::get_singleton()->execute(p_path, args, &pipe);
 	if (err != OK) {
 		if (r_err) {
 			*r_err = TTR("Can't execute Blender binary.");
@@ -87,7 +74,7 @@ static bool _get_blender_version(const String &p_path, int &r_major, int &r_mino
 	int bl = pipe.find("Blender ");
 	if (bl == -1) {
 		if (r_err) {
-			*r_err = vformat(TTR("Unexpected --version output from Blender binary at: %s."), path);
+			*r_err = vformat(TTR("Unexpected --version output from Blender binary at: %s."), p_path);
 		}
 		return false;
 	}
@@ -126,7 +113,7 @@ void EditorSceneFormatImporterBlend::get_extensions(List<String> *r_extensions) 
 Node *EditorSceneFormatImporterBlend::import_scene(const String &p_path, uint32_t p_flags,
 		const HashMap<StringName, Variant> &p_options,
 		List<String> *r_missing_deps, Error *r_err) {
-	String blender_path = EDITOR_GET("filesystem/import/blender/blender3_path");
+	String blender_path = EDITOR_GET("filesystem/import/blender/blender_path");
 
 	if (blender_major_version == -1 || blender_minor_version == -1) {
 		_get_blender_version(blender_path, blender_major_version, blender_minor_version, nullptr);
@@ -369,7 +356,7 @@ static bool _test_blender_path(const String &p_path, String *r_err = nullptr) {
 bool EditorFileSystemImportFormatSupportQueryBlend::is_active() const {
 	bool blend_enabled = GLOBAL_GET("filesystem/import/blender/enabled");
 
-	if (blend_enabled && !_test_blender_path(EDITOR_GET("filesystem/import/blender/blender3_path").operator String())) {
+	if (blend_enabled && !_test_blender_path(EDITOR_GET("filesystem/import/blender/blender_path").operator String())) {
 		// Intending to import Blender, but blend not configured.
 		return true;
 	}
@@ -409,11 +396,59 @@ void EditorFileSystemImportFormatSupportQueryBlend::_validate_path(String p_path
 	}
 }
 
-bool EditorFileSystemImportFormatSupportQueryBlend::_autodetect_path(String p_path) {
-	if (_test_blender_path(p_path)) {
-		auto_detected_path = p_path;
-		return true;
+bool EditorFileSystemImportFormatSupportQueryBlend::_autodetect_path() {
+	// Autodetect
+	auto_detected_path = "";
+
+#if defined(MACOS_ENABLED)
+	Vector<String> find_paths = {
+		"/opt/homebrew/bin/blender",
+		"/opt/local/bin/blender",
+		"/usr/local/bin/blender",
+		"/usr/local/opt/blender",
+		"/Applications/Blender.app/Contents/MacOS/Blender",
+	};
+	{
+		List<String> mdfind_args;
+		mdfind_args.push_back("kMDItemCFBundleIdentifier=org.blenderfoundation.blender");
+
+		String output;
+		Error err = OS::get_singleton()->execute("mdfind", mdfind_args, &output);
+		if (err == OK) {
+			for (const String &find_path : output.split("\n")) {
+				find_paths.push_back(find_path.path_join("Contents/MacOS/Blender"));
+			}
+		}
 	}
+#elif defined(WINDOWS_ENABLED)
+	Vector<String> find_paths = {
+		"C:\\Program Files\\Blender Foundation\\blender.exe",
+		"C:\\Program Files (x86)\\Blender Foundation\\blender.exe",
+	};
+	{
+		char blender_opener_path[MAX_PATH];
+		DWORD path_len = MAX_PATH;
+		HRESULT res = AssocQueryString(0, ASSOCSTR_EXECUTABLE, ".blend", "open", blender_opener_path, &path_len);
+		if (res == S_OK) {
+			find_paths.push_back(String(blender_opener_path).get_base_dir().path_join("blender.exe"));
+		}
+	}
+
+#elif defined(UNIX_ENABLED)
+	Vector<String> find_paths = {
+		"/usr/bin/blender",
+		"/usr/local/bin/blender",
+		"/opt/blender/bin/blender",
+	};
+#endif
+
+	for (const String &find_path : find_paths) {
+		if (_test_blender_path(find_path)) {
+			auto_detected_path = find_path;
+			return true;
+		}
+	}
+
 	return false;
 }
 
@@ -427,7 +462,7 @@ void EditorFileSystemImportFormatSupportQueryBlend::_select_install(String p_pat
 }
 void EditorFileSystemImportFormatSupportQueryBlend::_browse_install() {
 	if (blender_path->get_text() != String()) {
-		browse_dialog->set_current_dir(blender_path->get_text());
+		browse_dialog->set_current_file(blender_path->get_text());
 	}
 
 	browse_dialog->popup_centered_ratio();
@@ -479,76 +514,10 @@ bool EditorFileSystemImportFormatSupportQueryBlend::query() {
 		EditorNode::get_singleton()->get_gui_base()->add_child(browse_dialog);
 	}
 
-	String path = EDITOR_GET("filesystem/import/blender/blender3_path");
+	String path = EDITOR_GET("filesystem/import/blender/blender_path");
 
-	if (path == "") {
-		// Autodetect
-		auto_detected_path = "";
-
-#if defined(MACOS_ENABLED)
-
-		{
-			Vector<String> mdfind_paths;
-			{
-				List<String> mdfind_args;
-				mdfind_args.push_back("kMDItemCFBundleIdentifier=org.blenderfoundation.blender");
-
-				String output;
-				Error err = OS::get_singleton()->execute("mdfind", mdfind_args, &output);
-				if (err == OK) {
-					mdfind_paths = output.split("\n");
-				}
-			}
-
-			bool found = false;
-			for (const String &found_path : mdfind_paths) {
-				found = _autodetect_path(found_path.path_join("Contents/MacOS"));
-				if (found) {
-					break;
-				}
-			}
-			if (!found) {
-				found = _autodetect_path("/opt/homebrew/bin");
-			}
-			if (!found) {
-				found = _autodetect_path("/opt/local/bin");
-			}
-			if (!found) {
-				found = _autodetect_path("/usr/local/bin");
-			}
-			if (!found) {
-				found = _autodetect_path("/usr/local/opt");
-			}
-			if (!found) {
-				found = _autodetect_path("/Applications/Blender.app/Contents/MacOS");
-			}
-		}
-#elif defined(WINDOWS_ENABLED)
-		{
-			char blender_opener_path[MAX_PATH];
-			DWORD path_len = MAX_PATH;
-			HRESULT res = AssocQueryString(0, ASSOCSTR_EXECUTABLE, ".blend", "open", blender_opener_path, &path_len);
-			if (res == S_OK && _autodetect_path(String(blender_opener_path).get_base_dir())) {
-				// Good.
-			} else if (_autodetect_path("C:\\Program Files\\Blender Foundation")) {
-				// Good.
-			} else {
-				_autodetect_path("C:\\Program Files (x86)\\Blender Foundation");
-			}
-		}
-
-#elif defined(UNIX_ENABLED)
-		if (_autodetect_path("/usr/bin")) {
-			// Good.
-		} else if (_autodetect_path("/usr/local/bin")) {
-			// Good
-		} else {
-			_autodetect_path("/opt/blender/bin");
-		}
-#endif
-		if (auto_detected_path != "") {
-			path = auto_detected_path;
-		}
+	if (path.is_empty() && _autodetect_path()) {
+		path = auto_detected_path;
 	}
 
 	blender_path->set_text(path);
@@ -569,7 +538,7 @@ bool EditorFileSystemImportFormatSupportQueryBlend::query() {
 
 	if (confirmed) {
 		// Can only confirm a valid path.
-		EditorSettings::get_singleton()->set("filesystem/import/blender/blender3_path", blender_path->get_text());
+		EditorSettings::get_singleton()->set("filesystem/import/blender/blender_path", blender_path->get_text());
 		EditorSettings::get_singleton()->save();
 	} else {
 		// Disable Blender import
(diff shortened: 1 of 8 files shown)


mxaddict:
Tested this on both linux and windows, seems to work fine.

Someone will have to test on OSX

doc/translations/zh_CN.po
```diff
@@ -43523,7 +43523,7 @@ msgid ""
 "glTF 2.0 3D import pipeline. This importer requires Blender to be installed "
 "by the user, so that it can be used to export the scene as glTF 2.0.\n"
 "The location of the Blender binary is set via the [code]filesystem/import/"
-"blender/blender3_path[/code] editor setting.\n"
+"blender/blender_path[/code] editor setting.\n"
```
bruvzg, line 43526: Translations should not be edited here, `.po` files are generated from Weblate.

modules/gltf/editor/editor_scene_importer_blend.cpp
```diff
@@ -467,7 +455,7 @@ bool EditorFileSystemImportFormatSupportQueryBlend::query() {
 		EditorNode::get_singleton()->get_gui_base()->add_child(browse_dialog);
 	}
 
-	String path = EDITOR_GET("filesystem/import/blender/blender3_path");
+	String path = EDITOR_GET("filesystem/import/blender/blender_path");
 
 	if (path == "") {
```
bruvzg, line 460: Removal of `blender.exe`/`blender` executable name seems reasonable, but it will break all autodetection logic here (which will never include executable name), so it should be updated as well.

doc/translations/zh_CN.po
mxaddict: Thanks for the headsup, I'll remove those changes from the PR

modules/gltf/editor/editor_scene_importer_blend.cpp
mxaddict: I'm not sure what do you mean by this, could you elaborate a bit?

This change is to make the blender_path require pointing to the actual binary instead of a directory (Since this was causing issues with picking a specific blender version)

mxaddict: I think I know what you mean now, you mean the behavior for when blender_path is set to "" as in the user did not set a path.

bruvzg: > I'm not sure what do you mean by this, could you elaborate a bit?

There's autodetect logic if path is empty, e.g, few lines below it's trying to check `/opt/homebrew/bin`, but since `blender` name in no longer appended, it should be changed to `/opt/homebrew/bin/blender` to find it (similar for the other cases).

mxaddict: Yeah, I'm updating that logic now as well.

mxaddict: I'm not familiar with the OSX .app location of blender, would it make sense to have it check: `/Applications/Blender.app/Contents/MacOS/blender` ? or `/Applications/Blender.app/Contents/MacOS/Blender` ?

mxaddict:
Instead of completely removing the logic to append the `blender` `blender.exe` or `Blender` to the end of the path, I've moved it into the `bool EditorFileSystemImportFormatSupportQueryBlend::_autodetect_path(String p_path)` method as to only append it when autodetecting.

This should preserve the autodetect logic and honor the full path if a path is set in the settings.

mxaddict:
I've also removed the changes to the translation files.

AThousandShips:
You also need to handle transition from the old setting to the new one, renaming without adding backward compatibility shouldn't be done

modules/gltf/editor/editor_scene_importer_blend.cpp
```diff
@@ -398,6 +386,18 @@ void EditorFileSystemImportFormatSupportQueryBlend::_validate_path(String p_path
 }
 
 bool EditorFileSystemImportFormatSupportQueryBlend::_autodetect_path(String p_path) {
+#ifdef WINDOWS_ENABLED
+	p_path = p_path.path_join("blender.exe");
```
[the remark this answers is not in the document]
mxaddict: I see, is there a preferred way to do this instead of directly modifying the p_path value?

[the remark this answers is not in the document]
AThousandShips: See the other method that does this 🙂

mxaddict:
> You also need to handle transition from the old setting to the new one, renaming without adding backward compatibility shouldn't be done

Gotcha, I did notice that my editor kept the old value in the settings after I changed the name, can you point me in the direction of some existing code that handles renamed preferences?

AThousandShips:
I'm not sure in this case, haven't worked with this specific case, could use `_set` in the settings and add a compatibility conversion, it's done elsewhere, when setting with the old name redirect it to the new one, will check later today but give it a try

modules/gltf/editor/editor_scene_importer_blend.cpp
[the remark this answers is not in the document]
mxaddict: I don't know how I missed that 😸 thanks!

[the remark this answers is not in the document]
mxaddict: This should now be resovled

modules/gltf/editor/editor_scene_importer_blend.cpp
```diff
@@ -398,8 +385,21 @@ void EditorFileSystemImportFormatSupportQueryBlend::_validate_path(String p_path
...
+	path = path.path_join("blender.exe");
+#else
+	path = path.path_join("blender");
+#endif
+
+#if defined(MACOS_ENABLED)
+	if (!FileAccess::exists(path)) {
+		path = p_path.path_join("Blender");
+	}
+#endif
```
bruvzg, line 399: In case of macOS it's capital `B` only if it's in bundle, but it can be installed from homebrew and have lower case `b` as well, so it's probably better to add it where the path is set, something like:

```diff
diff --git a/modules/gltf/editor/editor_scene_importer_blend.cpp b/modules/gltf/editor/editor_scene_importer_blend.cpp
index 9587604e56..4440eff2e4 100644
--- a/modules/gltf/editor/editor_scene_importer_blend.cpp
+++ b/modules/gltf/editor/editor_scene_importer_blend.cpp
@@ -490,25 +490,25 @@ bool EditorFileSystemImportFormatSupportQueryBlend::query() {
 
 			bool found = false;
 			for (const String &found_path : mdfind_paths) {
-				found = _autodetect_path(found_path.path_join("Contents/MacOS"));
+				found = _autodetect_path(found_path.path_join("Contents/MacOS/Blender"));
 				if (found) {
 					break;
 				}
 			}
 			if (!found) {
-				found = _autodetect_path("/opt/homebrew/bin");
+				found = _autodetect_path("/opt/homebrew/bin/blender");
 			}
 			if (!found) {
-				found = _autodetect_path("/opt/local/bin");
+				found = _autodetect_path("/opt/local/bin/blender");
 			}
 			if (!found) {
-				found = _autodetect_path("/usr/local/bin");
+				found = _autodetect_path("/usr/local/bin/blender");
 			}
 			if (!found) {
-				found = _autodetect_path("/usr/local/opt");
+				found = _autodetect_path("/usr/local/opt/blender");
 			}
 			if (!found) {
-				found = _autodetect_path("/Applications/Blender.app/Contents/MacOS");
+				found = _autodetect_path("/Applications/Blender.app/Contents/MacOS/Blender");
 			}
 		}
 #elif defined(WINDOWS_ENABLED)
@@ -516,22 +516,22 @@ bool EditorFileSystemImportFormatSupportQueryBlend::query() {
 			char blender_opener_path[MAX_PATH];
 			DWORD path_len = MAX_PATH;
 			HRESULT res = AssocQueryString(0, ASSOCSTR_EXECUTABLE, ".blend", "open", blender_opener_path, &path_len);
-			if (res == S_OK && _autodetect_path(String(blender_opener_path).get_base_dir())) {
+			if (res == S_OK && _autodetect_path(String(blender_opener_path).get_base_dir().path_join("blender.exe"))) {
 				// Good.
-			} else if (_autodetect_path("C:\\Program Files\\Blender Foundation")) {
+			} else if (_autodetect_path("C:\\Program Files\\Blender Foundation\\blender.exe")) {
 				// Good.
 			} else {
-				_autodetect_path("C:\\Program Files (x86)\\Blender Foundation");
+				_autodetect_path("C:\\Program Files (x86)\\Blender Foundation\\blender.exe");
 			}
 		}
 
 #elif defined(UNIX_ENABLED)
-		if (_autodetect_path("/usr/bin")) {
+		if (_autodetect_path("/usr/bin/blender")) {
 			// Good.
-		} else if (_autodetect_path("/usr/local/bin")) {
+		} else if (_autodetect_path("/usr/local/bin/blender")) {
 			// Good
 		} else {
-			_autodetect_path("/opt/blender/bin");
+			_autodetect_path("/opt/blender/bin/blender");
 		}
 #endif
 		if (auto_detected_path != "") {
```

mxaddict: I have a different idea in mind, I'll push it to the PR let me know what you think, it should make the code a little easier to read, and since we're touching this method now anyway, might as well clean it up a little.

mxaddict:
I've pushed a refactor for the autodetect logic, to make it a little more readable, let me know what you guys think.

modules/gltf/editor/editor_scene_importer_blend.cpp
```diff
@@ -467,73 +454,55 @@ bool EditorFileSystemImportFormatSupportQueryBlend::query() {
...
 			char blender_opener_path[MAX_PATH];
 			DWORD path_len = MAX_PATH;
 			HRESULT res = AssocQueryString(0, ASSOCSTR_EXECUTABLE, ".blend", "open", blender_opener_path, &path_len);
-			if (res == S_OK && _autodetect_path(String(blender_opener_path).get_base_dir())) {
-				// Good.
-			} else if (_autodetect_path("C:\\Program Files\\Blender Foundation")) {
-				// Good.
-			} else {
-				_autodetect_path("C:\\Program Files (x86)\\Blender Foundation");
+			if (res == S_OK && ) {
```
AThousandShips, line 487: You removed a condition here
```suggestion
			if (res == S_OK) {
```
Unless you need some added condition

modules/gltf/editor/editor_scene_importer_blend.cpp
```diff
@@ -467,73 +454,55 @@ bool EditorFileSystemImportFormatSupportQueryBlend::query() {
...
 			DWORD path_len = MAX_PATH;
 			HRESULT res = AssocQueryString(0, ASSOCSTR_EXECUTABLE, ".blend", "open", blender_opener_path, &path_len);
-			if (res == S_OK && _autodetect_path(String(blender_opener_path).get_base_dir())) {
-				// Good.
-			} else if (_autodetect_path("C:\\Program Files\\Blender Foundation")) {
-				// Good.
-			} else {
-				_autodetect_path("C:\\Program Files (x86)\\Blender Foundation");
+			if (res == S_OK) {
+				find_paths.push_back(String(blender_opener_path).get_base_dir().path_join("blender.exe"))))
```
AThousandShips, line 488:
```suggestion
				find_paths.push_back(String(blender_opener_path).get_base_dir().path_join("blender.exe"));
```

mxaddict:
I'm working on backward compatibility for the setting path, will push it later.

mxaddict:
Would it make sense to add some new logic to the `EDITOR_GET()` macro or the `bool _get(const StringName &p_name, Variant &r_ret) const;` function from `editor/editor_settings.cpp` to check for values when requesting the `blender_path` value?

This should work fine, but it feels a bit `"hacky"` to me.

If any of you know a better way to implement a check for `filesystem/import/blender/blender3_path` and map it to new `filesystem/import/blender/blender_path` editor setting please let me know.

From what I can see in the codebase, I would either have to hook into the MARCRO or the function inside the editor_settings singleton.

mxaddict:
> Would it make sense to add some new logic to the `EDITOR_GET()` macro or the `bool _get(const StringName &p_name, Variant &r_ret) const;` function from `editor/editor_settings.cpp` to check for values when requesting the `blender_path` value?
> 
> This should work fine, but it feels a bit `"hacky"` to me.
> 
> If any of you know a better way to implement a check for `filesystem/import/blender/blender3_path` and map it to new `filesystem/import/blender/blender_path` editor setting please let me know.
> 
> From what I can see in the codebase, I would either have to hook into the MARCRO or the function inside the editor_settings singleton.

Actually, now that I've had a little more of a think about this.

Might make more sense that it only checks for that `filesystem/import/blender/blender3_path` if no `filesystem/import/blender/blender_path` is set.

And if `filesystem/import/blender/blender3_path` was set, then update it to work (As in append either `blender.exe`, `blender`, `Blender` to it, and save it into `filesystem/import/blender/blender_path` as the new setting.

mxaddict:
It also seems that having never set the `blender3_path` setting (as in fresh intall, no config changes) will never actually run the auto detect logic due to this code:
```cpp
	bool blend_enabled = GLOBAL_GET("filesystem/import/blender/enabled");
	String blender_path = EDITOR_GET("filesystem/import/blender/blender_path");
	if (blend_enabled) {
		if (blender_path.is_empty()) {
			WARN_PRINT(TTR("Blend file import is enabled in the project settings, but no Blender path is configured in the editor settings. Blend files will not be imported."));
		} else if (!FileAccess::exists(blender_path)) {
			WARN_PRINT(TTR("Blend file import is enabled, but the Blender path doesn't point to an accessible blender installation. Blend files will not be imported."));
		} else {
			Ref<EditorSceneFormatImporterBlend> importer;
			importer.instantiate();
			ResourceImporterScene::add_scene_importer(importer);

			Ref<EditorFileSystemImportFormatSupportQueryBlend> blend_import_query;
			blend_import_query.instantiate();
			EditorFileSystem::get_singleton()->add_import_format_support_query(blend_import_query);
		}
	}

```

mxaddict:
> It also seems that having never set the `blender3_path` setting (as in fresh intall, no config changes) will never actually run the auto detect logic due to this code:
> 
> ```c++
> 	bool blend_enabled = GLOBAL_GET("filesystem/import/blender/enabled");
> 	String blender_path = EDITOR_GET("filesystem/import/blender/blender_path");
> 	if (blend_enabled) {
> 		if (blender_path.is_empty()) {
> 			WARN_PRINT(TTR("Blend file import is enabled in the project settings, but no Blender path is configured in the editor settings. Blend files will not be imported."));
> 		} else if (!FileAccess::exists(blender_path)) {
> 			WARN_PRINT(TTR("Blend file import is enabled, but the Blender path doesn't point to an accessible blender installation. Blend files will not be imported."));
> 		} else {
> 			Ref<EditorSceneFormatImporterBlend> importer;
> 			importer.instantiate();
> 			ResourceImporterScene::add_scene_importer(importer);
> 
> 			Ref<EditorFileSystemImportFormatSupportQueryBlend> blend_import_query;
> 			blend_import_query.instantiate();
> 			EditorFileSystem::get_singleton()->add_import_format_support_query(blend_import_query);
> 		}
> 	}
> ```

Specifically these lines:
```cpp
 		if (blender_path.is_empty()) {
 			WARN_PRINT(TTR("Blend file import is enabled in the project 
```

As the default value is "" and this line if statement stops the logic from ever getting here:
```cpp
	if (path == "") {
		// Autodetect
		auto_detected_path = "";
```

mxaddict:
> It also seems that having never set the `blender3_path` setting (as in fresh intall, no config changes) will never actually run the auto detect logic due to this code:
> 
> ```c++
> 	bool blend_enabled = GLOBAL_GET("filesystem/import/blender/enabled");
> 	String blender_path = EDITOR_GET("filesystem/import/blender/blender_path");
> 	if (blend_enabled) {
> 		if (blender_path.is_empty()) {
> 			WARN_PRINT(TTR("Blend file import is enabled in the project settings, but no Blender path is configured in the editor settings. Blend files will not be imported."));
> 		} else if (!FileAccess::exists(blender_path)) {
> 			WARN_PRINT(TTR("Blend file import is enabled, but the Blender path doesn't point to an accessible blender installation. Blend files will not be imported."));
> 		} else {
> 			Ref<EditorSceneFormatImporterBlend> importer;
> 			importer.instantiate();
> 			ResourceImporterScene::add_scene_importer(importer);
> 
> 			Ref<EditorFileSystemImportFormatSupportQueryBlend> blend_import_query;
> 			blend_import_query.instantiate();
> 			EditorFileSystem::get_singleton()->add_import_format_support_query(blend_import_query);
> 		}
> 	}
> ```

Changing this to:
```cpp
	if (blend_enabled) {
		if (blender_path.is_empty() || FileAccess::exists(blender_path)) {
			Ref<EditorSceneFormatImporterBlend> importer;
			importer.instantiate();
			ResourceImporterScene::add_scene_importer(importer);

			Ref<EditorFileSystemImportFormatSupportQueryBlend> blend_import_query;
			blend_import_query.instantiate();
			EditorFileSystem::get_singleton()->add_import_format_support_query(blend_import_query);
		} else {
			WARN_PRINT(TTR("Blend file import is enabled, but the Blender path doesn't point to an accessible blender installation. Blend files will not be imported."));
		}
	}
```

Would allow the autodetect to work I think, will test it out and push if it works.

mxaddict:
I've push some more changes.

In summary this patch now works as follow:

1. Changed `blender3_path` to `blender_path` since blender 4.x is out and we support that version as well
2. Allow user to set full path to file for blender_path setting both in the popup and in the settings page
3. Added checking for blender and fbx reimports (previously it would not check if you had those settings set on reimport of the assets)
4. Stopped disabling blender imports when blender setting was not set (This is now handled by the check added for item 3 in this list

mxaddict:
Should there be any unit tests that I need to add with this PR?

I've only tested this PR manually on linux and windows, I don't have access to a mac, so I've not tested it on that platform sadly.

mxaddict:
Anything else that I need to add to this PR? Happy holidays btw! 😄

mxaddict:
This PR has been open for a while now, not sure what the next step is.

editor/import_dock.cpp
```diff
@@ -572,6 +572,12 @@ void ImportDock::_reimport() {
 		Error err = config->load(params->paths[i] + ".import");
 		ERR_CONTINUE(err != OK);
 
+		// Added this to check if we have proper importer support for
+		// these files before we actually attempt to import it
+		// This will allow user to set blender_path and fbx2gltf_paths
+		// if they reimport those files and don't have the settings set
```
akien-mga, line 578:
```suggestion
		// Added this to check if we have proper importer support for
		// these files before we actually attempt to import it.
		// This will allow user to set blender_path and fbx2gltf_paths
		// if they reimport those files and don't have the settings set.
```

editor/import_dock.cpp
```diff
@@ -572,6 +572,12 @@ void ImportDock::_reimport() {
...
+		EditorFileSystem::get_singleton()->scan_import_support(params->paths);
```
akien-mga, line 579: I'm a bit wary of changes to the core reimport rountine. I feel this might be easier to review as a separate PR, keeping only the direct Blender changes in this PR.

akien-mga:
Looks good overall, but I pointed a change I think would be better moved to a separate PR to be assessed properly in context. The rest seems good to merge now.

editor/import_dock.cpp
mxaddict: I think if I don't have this code, The editor does not work as intended, I don't remember what the use case was, but it was something to do with blender and fbx import not working when those assets were in the project.

I can move this into a different PR for now though as it does not affect the main goal of the PR.

mxaddict:
I've squashed the commits and removed the change to scan_import_support

akien-mga (approved):

akien-mga:

mxaddict:

No problem, and thanks for the help!

PR #85450, opened 2023-11-28
Improve documentation for dynamic properties

### LINKED ISSUE #30440: default values for properties added by _get_property_list()
As of now (3.1.1 stable) it's not possible to assign default values to properties added via _get_property_list(). My current workaround is to use the same variable and property name and set the default value on the variable, but this doesn't work for any variable that needs custom _set() code. 

There's also the problem that you can't revert properties added by _get_property_list() back to their defaults in the inspector (the revert arrow doesn't show up).

It would be neat if it was possible to add a "default_value" field to the returned dictionary from _get_property_list().

KoBeWi:
Improves descriptions of `_get_property_list()` and `_validate_property()`, adding some more details.

Resolves #30440

### DIFF
--- a/doc/classes/Object.xml
+++ b/doc/classes/Object.xml
@@ -76,83 +76,112 @@
 		<method name="_get_property_list" qualifiers="virtual">
 			<return type="Dictionary[]" />
 			<description>
-				Override this method to customize how script properties should be handled by the engine.
+				Override this method to provide a custom list of additional properties to handle by the engine.
 				Should return a property list, as an [Array] of dictionaries. The result is added to the array of [method get_property_list], and should be formatted in the same way. Each [Dictionary] must at least contain the [code]name[/code] and [code]type[/code] entries.
-				The example below displays [code]hammer_type[/code] in the Inspector dock, only if [code]holding_hammer[/code] is [code]true[/code]:
+				You can use [method _property_can_revert] and [method _property_get_revert] to customize the default values of the properties added by this method.
+				The example below displays a list of numbers shown as words going from [code]ZERO[/code] to [code]FIVE[/code], with [code]number_count[/code] controlling the size of the list:
 				[codeblocks]
 				[gdscript]
 				@tool
-				extends Node2D
+				extends Node
 
-				@export var holding_hammer = false:
-				    set(value):
-				        holding_hammer = value
+				@export var number_count = 3:
+				    set(nc):
+				        number_count = nc
+				        numbers.resize(number_count)
 				        notify_property_list_changed()
-				var hammer_type = 0
+
+				var numbers = PackedInt32Array([0, 0, 0])
 
 				func _get_property_list():
-				    # By default, `hammer_type` is not visible in the editor.
-				    var property_usage = PROPERTY_USAGE_NO_EDITOR
-
-				    if holding_hammer:
-				        property_usage = PROPERTY_USAGE_DEFAULT
-
 				    var properties = []
-				    properties.append({
-				        "name": "hammer_type",
-				        "type": TYPE_INT,
-				        "usage": property_usage, # See above assignment.
-				        "hint": PROPERTY_HINT_ENUM,
-				        "hint_string": "Wooden,Iron,Golden,Enchanted"
-				    })
+
+				    for i in range(number_count):
+				        properties.append({
+				            "name": "number_%d" % i,
+				            "type": TYPE_INT,
+				            "hint": PROPERTY_HINT_ENUM,
+				            "hint_string": "ZERO,ONE,TWO,THREE,FOUR,FIVE",
+				        })
 
 				    return properties
+
+				func _get(property):
+				    if property.begins_with("number_"):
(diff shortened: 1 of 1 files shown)


Mickeon (approved):
I like it. It's not bloated information, and links the methods together better.

doc/classes/Object.xml
```diff
@@ -76,8 +76,9 @@
 		<method name="_get_property_list" qualifiers="virtual">
 			<return type="Dictionary[]" />
 			<description>
-				Override this method to customize how script properties should be handled by the engine.
+				Override this method to provide a custom list of additional properties to handle by the engine.
```
Mickeon, line 79: Not sure about saying "additional". It's misleading because they may also already exist, and the example below even contradicts it.

KoBeWi: If it could modify existing properties, `_validate_property()` would be pointless. I'll have to re-check, but the example is likely wrong 🙃

Mickeon: Make no mistake, existing properties **created from a script**, not built-in

KoBeWi: Well the description is valid. This is what happens when you export `hammer_type`:
image
`_get_property_list()` still adds a property, even if it's a duplicate. Same applies to built-in properties, so they behave the same in this regard. The `hammer_type` added in the example is a new property, in addition to the existing one, which is not exposed in the inspector. The original property can't be modified using `_get_property_list()`.

Mickeon: > _get_property_list() still adds a property, even if it's a duplicate. 

This was indeed the only way to go about this before `_validate_property()` was added, way after the example was written

KoBeWi: So should I clarify in the example that the other `hammer_type` is a new property? Or maybe rename one of them?
Or it's fine as is?

Mickeon: In my opinion you could move the example to `_validate_property` and repurpose something else for `_get_property_list`, maybe inspired by the way TileMap layers are displayed or Mesh materials can be overridden. Not that the original example is inaccurate. I do repeat, it was the recommended and only way to do this before `_validate_property`.

KoBeWi: Pushed the new example. Here's how it looks when used:
image
The C# example needs rewriting.

doc/classes/Object.xml
```diff
@@ -76,83 +76,48 @@
...
 				Should return a property list, as an [Array] of dictionaries. The result is added to the array of [method get_property_list], and should be formatted in the same way. Each [Dictionary] must at least contain the [code]name[/code] and [code]type[/code] entries.
+				You can use [method _property_can_revert] and [method _property_get_revert] to customize the default values of the properties added by this method.
 				The example below displays [code]hammer_type[/code] in the Inspector dock, only if [code]holding_hammer[/code] is [code]true[/code]:
```
Mickeon, line 82: This needs to go with the example changed.

KoBeWi: You mean I should add it in the example?

Mickeon: > The example below displays [code]hammer_type[/code] in the Inspector dock, only if [code]holding_hammer[/code] is [code]true[/code]:

But the example is different now, so the sentence is wrong.

KoBeWi: Ah right.

Mickeon (approved):
C# example missing asside, this is very good now.

doc/classes/Object.xml
raulsntos:
```csharp
[Tool]
public partial class MyNode : Node
{
    private int _numberCount;

    [Export]
    public int NumberCount
    {
        get => _numberCount;
        set
        {
            _numberCount = value;
            _numbers.Resize(_numberCount);
            NotifyPropertyListChanged();
        }
    }

    private List<int> _numbers = new();

    public override Godot.Collections.Array<Godot.Collections.Dictionary> _GetPropertyList()
    {
        var properties = new Godot.Collections.Array<Godot.Collections.Dictionary>();

        for (int i = 0; i < _numberCount; i++)
        {
            properties.Add(new Godot.Collections.Dictionary()
            {
                { "name", $"number_{i}" },
                { "type", (int)Variant.Type.Int },
                { "hint", (int)PropertyHint.Enum },
                { "hint_string", "Zero,One,Two,Three,Four,Five" },
            });
        }

        return properties;
    }

    public override Variant _Get(StringName property)
    {
        string propertyName = property.ToString();
        if (propertyName.StartsWith("number_"))
        {
            int index = int.Parse(propertyName.Substring("number_".Length));
            return _numbers[index];
        }
        return default;
    }

    public override bool _Set(StringName property, Variant value)
    {
        string propertyName = property.ToString();
        if (propertyName.StartsWith("number_"))
        {
            int index = int.Parse(propertyName.Substring("number_".Length));
            numbers[index] = value.As<int>();
            return true;
        }
        return false;
    }
}
```

akien-mga:
Thanks!

PR #85454, opened 2023-11-28
Fix issue causing Godot Android apps / games to freeze on close

m4gr3d:
The issue occurred because during the 'close' event, the logic was trying to terminate the native engine on the UI thread instead of doing on the render thread.

### DIFF
--- a/platform/android/java/lib/src/org/godotengine/godot/Godot.kt
+++ b/platform/android/java/lib/src/org/godotengine/godot/Godot.kt
@@ -536,8 +536,11 @@ class Godot(private val context: Context) : SensorEventListener {
 		for (plugin in pluginRegistry.allPlugins) {
 			plugin.onMainDestroy()
 		}
-		GodotLib.ondestroy()
-		forceQuit()
+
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks!

ZishAan23:
In which version is this error going to be fixed ?

YuriSizov:
@ZishAan23 This is a part of Godot 4.2, released 3 weeks ago.

ZishAan23:
@YuriSizov if the error is fixed in Godot 4.2 then why my Godot editor is still causing the same issue. One error is fixed that is whenever I want to exit from my game and return to editor I can easily tap the back button and exit. However when an error occurs the same thing happens the back button doesn't work. Is there a new setting or something that I have to change to fix the issue?

YuriSizov:
@ZishAan23 This PR is addressing a very broad problem of Android apps and games freezing. There can be many reasons for why this happens. Please file a bug report with your example and a reproduction project so it can be debugged.

PR #85455, opened 2023-11-28
Notify CSGPolygon about transform changes in Path3D

### LINKED ISSUE #82024: Moving Path3D does not trigger a CSGPolygon draw update
### Godot version

4.1.1.stable, 4.2.dev5.offical

### System information

Arch Linux

### Issue description

Update after deeper understanding:
I believe the core issue I see is that Path3D takes it's global position and applies it as an offset to it's CSGPolygon Mode:Path transform. but moving the Path3D in code or the editor does not seem to trigger a redraw of the CSGPolygon with the new offset until you force an update with other notification. 

Original Report:
> If you move a node with a child CSGPolygon and Path3D in the Editor or during ready(). The visual of the CSGPolygon will move twice as far as expected and desync its reported position with it's visual position. The child CSGPolygon moves with and further than the parent visually but reports the same position/global_position as the parent moved. Only occurs with CSGPolygon's in mode: Path and only when moving a parent to both the path and CSG.
> 
> Also seen in ready() if you move the parent to x = 4.0, the CSGPolygon will visually move double distance of what the parent moves, making the CSGPolygon appear at 8.0, the extra shift is not reported in the inspector position.
> 
> Doing the SAME MOVE in process(): appears at the expected 4.0. It keeps the offset from above bug if it was seen, but if all moves are done here it all works as expected and the visual moves with the parent
> 
> Enabling Path Local does remove this issue and the moves match the parent
> 
> The expected Behavior is the CSGPolygon visual moves with the parent and matches it's position value. Moving the parent should not cause the CSG visual to move "twice" and stay offset even when resetting all the positions involved to 0,0,0
> 
> bugreport1
> In this picture the redball and the right Block are supposed to be at 0,0,0 in the parent but visually the CSGPolygon has been shifted by the parent's move away from 0,0,0

### Steps to reproduce

Grab either NodeEdited* node3d and move them in editor. (this moves the Path3D global position)
Note the wrong position of the CSG. cause a CSG update and see the correct position
Also run the project, the code should set both parent nodes to position.x = 2.0  and keep the CSGPolygons overlapping but the code in the ready runs before the draw and the process does not trigger another draw

[CSGPolygonBug.zip](https://github.com/godotengine/godot/files/12680049/CSGPolygonBug.zip)

xiongyaohua:
Fix https://github.com/godotengine/godot/issues/82024

Path3D notify transform change to CSGPolygon.

https://github.com/godotengine/godot/assets/6020486/c5555c4f-9db3-472b-b9ca-5237a5f43a2f

### DIFF
--- a/modules/csg/csg_shape.cpp
+++ b/modules/csg/csg_shape.cpp
@@ -1826,11 +1826,13 @@ CSGBrush *CSGPolygon3D::_build_brush() {
 			if (path) {
 				path->disconnect("tree_exited", callable_mp(this, &CSGPolygon3D::_path_exited));
 				path->disconnect("curve_changed", callable_mp(this, &CSGPolygon3D::_path_changed));
+				path->set_update_callback(Callable());
 			}
 			path = current_path;
 			if (path) {
 				path->connect("tree_exited", callable_mp(this, &CSGPolygon3D::_path_exited));
 				path->connect("curve_changed", callable_mp(this, &CSGPolygon3D::_path_changed));
+				path->set_update_callback(callable_mp(this, &CSGPolygon3D::_path_changed));
 			}
 		}
 
--- a/scene/3d/path_3d.cpp
+++ b/scene/3d/path_3d.cpp
@@ -50,6 +50,10 @@ Path3D::~Path3D() {
 	}
 }
 
+void Path3D::set_update_callback(Callable p_callback) {
+	update_callback = p_callback;
+}
+
 void Path3D::_notification(int p_what) {
 	switch (p_what) {
 		case NOTIFICATION_ENTER_TREE: {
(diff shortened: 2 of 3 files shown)


Chaosus:
Please amend the commit name to the actual PR name.

fire (approved):

kleonc (commented):
Does it need adding/exposing `transform_changed` signal? :thinking:

It seems rather strange design-wise that specifically Path3D would have such signal given that `transform` is Node3D's property. I suggest production team reconsidering this API-wise before merging.

xiongyaohua:
> It seems rather strange design-wise that specifically Path3D would have such signal given that transform is Node3D's property.

According to this [comment](https://github.com/godotengine/godot-proposals/issues/4729#issuecomment-1164592096) add signal to all Node3D will strongly hit performance, so I only add it to the `Path3D` to minimize the impact. Of course maybe there is better way.

YuriSizov:
> Ideally the signal should be on `Node3D`, however according to this [comment](https://github.com/godotengine/godot-proposals/issues/4729#issuecomment-1164592096) that will strongly hit performance, so I only add it to the `Path3D` to minimize impact.

I would say that given the comment you've linked, this is not an appropriate solution. We could probably add a way to hook a function pointer from the CSG shape into the path instead. That should be as cheap as doing the notification, performance wise, and won't expose anything to the API.

AThousandShips:
Like how `ViewPanner` works, with some callback

xiongyaohua:
> Like how ViewPanner works, with some callback

done.

kleonc (approved):

akien-mga:
Thanks!

PR #85457, opened 2023-11-28
Fix an issue causing the running project window to loop-restart when closed

m4gr3d:
One of the methods used to close the running project window on the Android editor was to dispatch an [Intent](https://developer.android.com/reference/android/content/Intent) to request the process closes itself.
In certain scenarios, this creates a race condition where the process receives the intent after closing itself causing it to restart itself again.


akien-mga (approved):

akien-mga:
Thanks!

PR #85462, opened 2023-11-28
TAA Improvements (Take Two)

### LINKED ISSUE #68343: Vulkan: TAA causes snapping effect when a camera and a light source are moving together
### Godot version

4.0 beta 4

### System information

Windows 10, Vulkan Clustered, GTX 650 Ti 2GB

### Issue description

When parenting a camera and a light source (spot or omni) to a character and moving, having TAA on causes a sort of snapping effect on the light. Also, any other moving lights will have the same happen to them to a lesser extent. It doesn't happen to stationary lights.

https://user-images.githubusercontent.com/25274301/200192353-d63bd610-6fcb-4a06-953e-a49bc6f7690c.mp4

### Steps to reproduce

- Create a dark scene with a low roughness material on the floor, preferably with a normal map.
- Add a moving light and camera to the scene and turn TAA on.

[LightTAA.zip](https://github.com/godotengine/godot/files/9946709/LightTAA.zip)

### LINKED ISSUE #81709: TAA Bad Quality

mrjustaguy:
Resolves https://github.com/godotengine/godot/issues/81709#issuecomment-1722183283 points 2, 5 and partially resolves point 1, with possible changes with points 3 and 4 (with the note that point 3 is no longer really valid for Specular and Shadow ghosts)

1) Removed Velocity Closest from TAA due to it's broken state, as well as Adaptive box size which depended on it, and was also a little messy. This greatly improves AA quality and Stability in Motion. This cleanup also slightly improves transparencies as there is no more Adaptive box size that is determined by velocity, making them blend more consistently with the background, though the root of the issue is still not resolved, as the root issue seems to be that transparencies are writing to the Motion Vectors, overriding what's behind them.

2) Moved Screen disocclusion after all the blending factor modifications are applied, to ensure no history is used for out of screen pixels.

3) Added Luminance check to help prevent Shadow Ghosts and Specular Ghosts. This Probably resolves https://github.com/godotengine/godot/issues/68343

### DIFF
--- a/servers/rendering/renderer_rd/effects/taa.cpp
+++ b/servers/rendering/renderer_rd/effects/taa.cpp
@@ -63,7 +63,7 @@ void TAA::resolve(RID p_frame, RID p_temp, RID p_depth, RID p_velocity, RID p_pr
 	push_constant.resolution_width = p_resolution.width;
 	push_constant.resolution_height = p_resolution.height;
 	push_constant.disocclusion_threshold = 0.025f;
-	push_constant.disocclusion_scale = 10.0f;
+	push_constant.disocclusion_scale = 0.0001f;
 
 	RD::ComputeListID compute_list = RD::get_singleton()->compute_list_begin();
 	RD::get_singleton()->compute_list_bind_compute_pipeline(compute_list, pipeline);
--- a/servers/rendering/renderer_rd/shaders/effects/taa_resolve.glsl
+++ b/servers/rendering/renderer_rd/shaders/effects/taa_resolve.glsl
@@ -91,7 +91,7 @@ vec3 reinhard_inverse(vec3 sdr) {
 const vec3 lumCoeff = vec3(0.299f, 0.587f, 0.114f);
 
 float luminance(vec3 color) {
-	return max(dot(reinhard(color), lumCoeff), 0.0001f);
+	return max(dot(color, lumCoeff), 0.0001f);
 }
 
 float get_depth(ivec2 thread_id) {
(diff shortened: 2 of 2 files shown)


mrjustaguy:
Before:

https://github.com/godotengine/godot/assets/50302479/56cb4a73-4bea-4521-9c6d-ff231d3306d1

After:

https://github.com/godotengine/godot/assets/50302479/90b900f1-a0d1-4498-825f-25ec6e547c76

Sadly due to the Compression It's hard to tell the difference with the Anti Aliasing and Motion Stability, and actually turns out it's less significant in a direct comparison, and that the transparency issues are significantly improved instead..
If you look at the Godot faces behind the Omni Light Gizmo's Quad, before it distorts like crazy, whereas after it distorts significantly less.

mrjustaguy:
To be fair, this comparison fails to see most of the improvements done due to the relatively favorable lighting conditions of the scene to the old TAA, as here it looks OK, where in other places it just fails to do any AA with such motions, where as this does an ok job in not falling apart in those scenarios.

Sadly I cannot see a way to not go with a blurry approach as attempts to avoid it always ended up in strong ghosting or jittering or absolute failing in changing motion (the issue most obvious in the current TAA, and what the https://github.com/godotengine/godot/issues/81709 complained about, and also what makes it absolutely a no go IMO)

mrjustaguy:
Aand I've gone back to the original implementation, set the box size to a fixed 0.25, dropped the disocclusion scale a ton, added a function to blur the TAA as velocity increases (mainly to combat jitter in the distance where the velocity is higher, and thus it weights more on the input sample), and suddenly the TAA looks great.... should I replace this PR with that or close it and make a take 3?

mrjustaguy:
Closing in favor of https://github.com/godotengine/godot/pull/86809

Specular fix could be salvaged if it remains an issue, as that's the only thing that seems to universally help, unlike all the other changes attempted.

PR #85464, opened 2023-11-28
Improve message about resources in use at exit

KoBeWi:
The message says "run with --verbose" even if you are running in verbose mode. This PR fixes it.


YuriSizov (approved):
Maybe we could also put the number of resources into the message? Could be useful at a glance.

YuriSizov:
Thanks!

PR #85468, opened 2023-11-28
Clarify difference between Capture and Record audio effects

RedMser:
Updated documentation of AudioEffectCapture and AudioEffectRecord:

- Capture returns a raw audio buffer, while Record encodes it into an audio stream.
- The note about `enable_input` setting was confusing in Capture docs, since the microphone use-case is only one of the many listed.
- Additionally, the demo project listed uses the Record effect, not Capture! So moved that note and link to the Record effect instead.
- Record also had a very weird wording in its docs ("It sets, It checks, It returns"...) which I cleaned up.


fire (approved):
@lyuma I think this looks ok.

akien-mga (approved):

akien-mga:
Thanks!

PR #84668, opened 2023-11-09
Allow auto-generated node names in `PopupMenu::add_submenu_item`

YuriSizov:
Follow-up to https://github.com/godotengine/godot/pull/84617. After a bit of consideration, @akien-mga and I decided that it's best to allow the forbidden character. This is an ad-hoc way to address it, though exposing it on `String` would've been useful as well. But it's an exposed method there and I don't want to mess up things.

PR #84617, opened 2023-11-08
Fix node names of submenu items across the editor

YuriSizov:
Fixes https://github.com/godotengine/godot/issues/84582, a regression from https://github.com/godotengine/godot/pull/84183. Also removes some programmer remarks and fixes some docs.

There is still an open question whether this validation is excessive or not, whether we should allow generated names here or not. We might want that, which means we need a second version of `validate_node_name` which allows `@` specifically. But I would argue that we want the user to name their node here. And the case of the C# module was the only place where we forgot to do that.

PR #85477, opened 2023-11-28
Add methods to add submenus without using names

### LINKED ISSUE #84692: Executing `PopupMenu.activate_item_by_event` function crashes Godot
### Godot version

4.2.beta.custom_build. 9df649185

### System information

Ubuntu 22.04 CI

### Issue description

When executing (this code was automatically minimized, so it is possible, that an even more "minimal" project can be created)
```
extends Node
func _process(delta):
	for i in get_children():
		i.queue_free()
	var temp_variable29522 = CodeHighlighter.new()
	temp_variable29522.clear_member_keyword_colors()
	var temp_variable29574 = GLTFBufferView.new()
	var temp_variable29766 = PopupMenu.new()
	add_child(temp_variable29766)
	temp_variable29766.get_item_icon_max_width(-63)
	temp_variable29766.get_item_shortcut(11)
	temp_variable29766.set_item_count(49)
	temp_variable29766.set_item_submenu(11, ".")
	temp_variable29766.activate_item_by_event(InputEventAction.new(), false)
```
Godot crashes:
```
Godot Engine v4.2.beta.custom_build.9df649185 - https://godotengine.org
Vulkan API 1.2.0 - Forward+ - Using Vulkan Device #0: Unknown - SwiftShader Device (LLVM 10.0.0)
ERROR: Index p_idx = -63 is out of bounds (items.size() = 0).
   at: get_item_icon_max_width (scene/gui/popup_menu.cpp:1813)
ERROR: Index p_idx = 11 is out of bounds (items.size() = 0).
   at: get_item_shortcut (scene/gui/popup_menu.cpp:1868)
Segmentation fault (core dumped)
```

This example was found by Godot fuzzer - [Qarminer](https://github.com/qarmin/Qarminer), so it is quite unlikelly that this code could be used in real project, but still this should be handled gracefully.

Memory leaks or asan backtraces are visible when using Godot build with sanitizers support - https://github.com/qarmin/GodotBuilds/actions (linux -> linux-editor-sanitizers)

   

### Steps to reproduce

Above

Above

KoBeWi:
Inspired by #84668

This PR adds `add_direct_submenu_item()` and related methods. It takes a PopupMenu directly and does not rely on its name. Also automatically adds the menu as a child if it doesn't have a parent. Refactored PopupMenu internals to not use submenu names anywhere, they are only kept for compatibility. The old methods are now deprecated.

There is some `global_menu_add_submenu_item()` method in DisplayServer, I did not touch it. I also changed editor usages to the new methods (except one case in AnimationNodeStateMachineEditor) and removed menu names, because they are no longer needed (this technically reverts #84617).

Note that when using the old methods, they are now required to have the submenu already added as a child (it allows to call the new method and simplify the code). This was actually documented, but wasn't enforced.

The new methods might need some testing. I did check if editor works correctly and tested some internal popups, but did not test the methods extensively (especially together with the old methods).

EDIT:
Fixes #84692

### DIFF
--- a/doc/classes/PopupMenu.xml
+++ b/doc/classes/PopupMenu.xml
@@ -197,13 +197,24 @@
 				If [param allow_echo] is [code]true[/code], the shortcut can be activated with echo events.
 			</description>
 		</method>
-		<method name="add_submenu_item">
+		<method name="add_submenu_item" deprecated="Prefer using [method add_submenu_node_item] instead.">
 			<return type="void" />
 			<param index="0" name="label" type="String" />
 			<param index="1" name="submenu" type="String" />
 			<param index="2" name="id" type="int" default="-1" />
 			<description>
 				Adds an item that will act as a submenu of the parent [PopupMenu] node when clicked. The [param submenu] argument must be the name of an existing [PopupMenu] that has been added as a child to this node. This submenu will be shown when the item is clicked, hovered for long enough, or activated using the [code]ui_select[/code] or [code]ui_right[/code] input actions.
+				An [param id] can optionally be provided. If no [param id] is provided, one will be created from the index.
+			</description>
+		</method>
+		<method name="add_submenu_node_item">
+			<return type="void" />
+			<param index="0" name="label" type="String" />
+			<param index="1" name="submenu" type="PopupMenu" />
+			<param index="2" name="id" type="int" default="-1" />
+			<description>
+				Adds an item that will act as a submenu of the parent [PopupMenu] node when clicked. This submenu will be shown when the item is clicked, hovered for long enough, or activated using the [code]ui_select[/code] or [code]ui_right[/code] input actions.
+				[param submenu] must be either child of this [PopupMenu] or has no parent node (in which case it will be automatically added as a child). If the [param submenu] popup has another parent, this method will fail.
 				An [param id] can optionally be provided. If no [param id] is provided, one will be created from the index.
 			</description>
 		</method>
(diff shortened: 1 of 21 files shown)


doc/classes/PopupMenu.xml
```diff
@@ -45,6 +45,17 @@
 				[b]Note:[/b] Checkable items just display a checkmark, but don't have any built-in checking behavior and must be checked/unchecked manually. See [method set_item_checked] for more info on how to control it.
 			</description>
 		</method>
+		<method name="add_direct_submenu_item">
```
YuriSizov, line 48: Maybe `add_submenu_node_item`/`get_item_submenu_node`/`set_item_submenu_node`?

KoBeWi: idk, if you think it's fine then ok. Though "add node" might suggest that the method is always adding a node, while it can also reuse node if it's already added. I'm not sure what would be a better name.

YuriSizov: I'm just not sure what "direct" would signify here. I have no association with the word in this context. Like, I understand what you mean with it, but not sure it works for user facing API.

By the way, we don't have to make adding the node optional, since it's a new method. Don't really see when that would be needed.

YuriSizov: Mmh, true, but it also can "just work" if it's already a child of this popup menu?

KoBeWi: Ok I'll rename it.

bruvzg (commented):
> There is some `global_menu_add_submenu_item()` method in DisplayServer, I did not touch it.

This should not be changed, global menus can exist without node, so they're always using string IDs (for attached `PopupMenu`s it's `__PopupMenu#{INSTANCE_ID}` so it's not directly related to `PopupMenu` submenu string).

doc/classes/PopupMenu.xml
```diff
@@ -207,6 +207,17 @@
...
 			</description>
 		</method>
+		<method name="add_submenu_node_item">
+			<return type="void" />
+			<param index="0" name="label" type="String" />
+			<param index="1" name="submenu" type="PopupMenu" />
+			<param index="2" name="id" type="int" default="-1" />
+			<description>
+				Adds an item that will act as a submenu of the parent [PopupMenu] node when clicked. This submenu will be shown when the item is clicked, hovered for long enough, or activated using the [code]ui_select[/code] or [code]ui_right[/code] input actions.
+				[param submenu] must be either child of this [PopupMenu] or has no parent node (in which case it will be automatically added as a child). If the [param submenu] popup has another parent, this method will fail.
```
akien-mga, line 217:
```suggestion
				[param submenu] must either be a child of this [PopupMenu] or have no parent node (in which case it will be automatically added as a child). If the [param submenu] popup has another parent, this method will fail.
```

akien-mga (approved):
Looks good to me overall. Didn't test.

akien-mga:
Thanks!

PR #85481, opened 2023-11-28
GDScript: Ensure super() is called in constructors

### LINKED ISSUE #40861: GDScript: no parser error when `super()` doesn't called in derived class
**Godot version:** v4.0.dev.custom_build 5ba46ddf8

**OS/device including version:** windows 10

**Issue description:**
it should remind to call `super()` if base class doesn't have any super constructor
```gdscript
class A:
	func _init(p):
		pass

class B extends A:
	func _init():  <-- no parser error here
		pass
		
```

vnen:
This is needed so the super class is properly constructed, especially when it requires arguments.

Fix #40861

### DIFF
--- a/modules/gdscript/gdscript_analyzer.cpp
+++ b/modules/gdscript/gdscript_analyzer.cpp
@@ -1803,6 +1803,19 @@ void GDScriptAnalyzer::resolve_function_body(GDScriptParser::FunctionNode *p_fun
 		}
 	}
 
+	// If this is a constructor and the super class has a custom constructor, ensure `super()` is called.
+	if (p_function->identifier != nullptr && p_function->identifier->name == GDScriptLanguage::get_singleton()->strings._init) {
+		GDScriptParser::ClassNode *base_class = parser->current_class->base_type.class_type;
+
+		while (base_class) {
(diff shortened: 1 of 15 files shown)


dalexeev:
> The thing is the constructor **has** to be called always. The implicit constructor is, and the user-created constructor should be as well, since it is what creates the object.

If for some reason overriding a constructor can lead to the implicit constructor not being called, then this is of course a bug. But I think that it can be fixed in some other way, without resorting to being forced to always call the parent constructor.

In my opinion, this is a valid feature: the user can override the constructor as they see fit, it is the user's responsibility not to violate the expectations of the base class, just like with other methods. If we just want to reduce the chance that the user will accidentally forget to call `super()`, then there are softer ways (like warnings) and this should be done for all methods, not just the constructor.

adamscott:
> just like with other methods

Discussed during the GDScript meeting. We think that the constructor is an entire different beast considering OOP principles. We should enforce the call of the constructor, any other way seems wrong.

See Typescript for example:
> One difference from the prior example is that each derived class that contains a constructor function must call super() which will execute the constructor of the base class. What’s more, before we ever access a property on this in a constructor body, we have to call super(). This is an important rule that TypeScript will enforce.
https://www.typescriptlang.org/docs/handbook/classes.html

PR #85486, opened 2023-11-28
Fix property groups overriding real properties

paulloz:
With the following script, each rebuild reverts the exported value to its default.
```cs
public partial class Foo : Node
{
    [ExportGroup("Test")]
    [Export]
    private string Test;
}
```
The same thing happens in GDScript. With the following script, modifying the default value in the code overwrites any value defined through the inspector.
```gdscript
extends Node

@export_group("test")
@export
var test : String = "Hello World!"
```
It is triggered by the group using the same name as the actual property (changing the group names above prevents the weird behaviour).

### DIFF
--- a/core/object/script_language.cpp
+++ b/core/object/script_language.cpp
@@ -631,6 +631,10 @@ bool PlaceHolderScriptInstance::has_method(const StringName &p_method) const {
 void PlaceHolderScriptInstance::update(const List<PropertyInfo> &p_properties, const HashMap<StringName, Variant> &p_values) {
 	HashSet<StringName> new_values;
 	for (const PropertyInfo &E : p_properties) {
+		if (E.usage & (PROPERTY_USAGE_GROUP | PROPERTY_USAGE_SUBGROUP | PROPERTY_USAGE_CATEGORY)) {
+			continue;
+		}
+
 		StringName n = E.name;
 		new_values.insert(n);
 


core/object/script_language.cpp
```diff
@@ -631,6 +632,10 @@ bool PlaceHolderScriptInstance::has_method(const StringName &p_method) const {
 void PlaceHolderScriptInstance::update(const List<PropertyInfo> &p_properties, const HashMap<StringName, Variant> &p_values) {
 	HashSet<StringName> new_values;
 	for (const PropertyInfo &E : p_properties) {
+		if (E.usage & PROPERTY_USAGE_GROUP || E.usage & PROPERTY_USAGE_SUBGROUP || E.usage & PROPERTY_USAGE_CATEGORY) {
```
AThousandShips, line 635:
```suggestion
		if (E.usage & (PROPERTY_USAGE_GROUP | PROPERTY_USAGE_SUBGROUP | PROPERTY_USAGE_CATEGORY)) {
```
More compact syntax and clearer

core/object/script_language.cpp
```diff
@@ -34,6 +34,7 @@
 #include "core/core_string_names.h"
 #include "core/debugger/engine_debugger.h"
 #include "core/debugger/script_debugger.h"
+#include "core/object/object.h"
```
AThousandShips, line 37:
```suggestion
```
Not needed, `PlaceholderScriptInstance` derives from `Object` and it is included already (if you are using a "include what you use" feature in your editor please turn it off when working with Godot)

paulloz: Thanks for the tip 🙏  
I switched to clangd recently and didn't notice that was the default behaviour!

YuriSizov (approved):
Makes sense, I think.

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.1.

PR #85487, opened 2023-11-28
GDScript: Make annotations register statically in parser

vnen:
So it only register the annotations once per run instead of doing it every time a script is parsed.


dalexeev (approved):
Looks good to me.

YuriSizov:
Thanks!

PR #85488, opened 2023-11-28
Move "Expected indented block" error message's line number back to where the error is

### LINKED ISSUE #81046: Error message "expected indented block" on misleading line
### Godot version

### System information

Godot v4.1.1.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1060 6GB (NVIDIA; 31.0.15.3640) - Intel(R) Core(TM) i5-3570K CPU @ 3.40GHz (4 Threads)

### Issue description

When a function body is missing, you get the following error message:

kuva

From the user's perspective the line displayed as an error line is totally wrong. I know that that is the line where the error is detected, but for the user it is a wrong line. At Godot subreddit it is quite often that users are stuck with these errors.

Another example where line 39 would be the most logical error location for the user. This is a little bit easier for the user as the error message tells that "for" block is missing something.

kuva

In my opinion the best fix would be to move the error location to the line that the error message refers to, in these examples function declaration line and for line. A small improvement would be to display the function name in the first example: `Error at (41, 1): Expected inteded block after function declaration 'foo()'`

### Steps to reproduce

```gdscript
func foo():

func bar():
	pass
```

```gdscript
func foo():
	for i in range(2):

func bar():
	pass
```

N/A

miv391:
Fixes https://github.com/godotengine/godot/issues/81046

In master "Expected indented block ..." error messages are assigned to code lines likes this:

kuva

The error message at line 7 tells that "Expected indented block after class declaration". The class declaration mentioned in error messages in not at line 7, it is at line 4. Other errors in screenshot are similar. The most confusing (I have seen this at Reddit discussions several times) is the error at line 16 which mentions function declaration. But the problem is not the function declaration at line 16, the problematic function declaration is at line 11.

In master, the error message line is obviously the line where the missing indented block is noticed.

This PR fixes this problem and now the same errors look like this:

kuva

Now all "Expected indented block ..." error messages are pointing to lines which are actually expecting something.

Other kind of errors _seem_ to be unaffected by this change:

kuva

But I haven't tested with every different error message and I don't have deep understanding about GDScript parser, so something may have broken. If there exists somewhere a test project which contains all possible GDScript error, I would be interested about that.

It might also be worth considering about changing these "Expected indented block ..." error messages a bit.

`"Expected indented block after class declaration"`
--> 
`"Expected indented block after this class declaration"`

Adding "this" would emphasize the fact that it is exactly _this_ function that needs indented block after it. But I may be overcautious here.

### DIFF
--- a/modules/gdscript/gdscript_parser.cpp
+++ b/modules/gdscript/gdscript_parser.cpp
@@ -165,7 +165,7 @@ void GDScriptParser::push_error(const String &p_message, const Node *p_origin) {
 	panic_mode = true;
 	// TODO: Improve positional information.
 	if (p_origin == nullptr) {
-		errors.push_back({ p_message, current.start_line, current.start_column });
+		errors.push_back({ p_message, previous.start_line, previous.start_column });
 	} else {
 		errors.push_back({ p_message, p_origin->start_line, p_origin->leftmost_column });
 	}


dalexeev (approved):
Makes sense to me since the `current` token can be separated from `previous` by empty lines or comments. I rebased and tested it locally, looks correct.

It would be nice to add some tests for the parser to prevent regressions:

```
800: 74: 	if (multiline && !consume(GDScriptTokenizer::Token::INDENT, R"(Expected indented block after class declaration.)")) {
1105: 62: 		if (!consume(GDScriptTokenizer::Token::INDENT, R"(Expected indented block for property after ":".)")) {
1127: 56: 			push_error("Property with inline code must go to an indented block.");
1180: 64: 		consume(GDScriptTokenizer::Token::DEDENT, R"(Expected end of indented block for property.)");
1679: 70: 		if (!consume(GDScriptTokenizer::Token::INDENT, vformat(R"(Expected indented block after %s.)", p_context))) {
2099: 64: 	if (!consume(GDScriptTokenizer::Token::INDENT, R"(Expected an indented block after "match" statement.)")) {
2151: 59: 	consume(GDScriptTokenizer::Token::DEDENT, R"(Expected an indented block after "match" statement.)");
```

AThousandShips:
Should probably also resolve:
* https://github.com/godotengine/godot/issues/92681

dalexeev:
> Should probably also resolve:
> 
> * #92681

No, I checked. #92681 is about a runtime error. It looks like it's being handled elsewhere, I'll take a look later.

akien-mga:
Thanks!

PR #85491, opened 2023-11-29
Fix documentation for `icon_and_font_color` editor setting

BNTFryingPan:
the documentation for the `interface/theme/icon_and_font_color` incorrectly states that the "dark" option makes the icons light, suitable for dark themes, and the "light" option makes the icons dark, suitable for light themes. this pr changes the documentation to say that the "dark" option makes the icons dark, suitable for light themes, and the "light" option makes the icons light, suitable for dark themes

### DIFF
--- a/doc/classes/EditorSettings.xml
+++ b/doc/classes/EditorSettings.xml
@@ -665,8 +665,8 @@
 		<member name="interface/theme/icon_and_font_color" type="int" setter="" getter="">
 			The icon and font color scheme to use in the editor.
 			- [b]Auto[/b] determines the color scheme to use automatically based on [member interface/theme/base_color].
-			- [b]Dark[/b] makes fonts and icons light (suitable for dark themes).
-			- [b]Light[/b] makes fonts and icons dark (suitable for light themes). Icon colors are automatically converted by the editor following [url=https://github.com/godotengine/godot/blob/master/editor/editor_themes.cpp#L135]this set of rules[/url].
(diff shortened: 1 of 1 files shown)


doc/classes/EditorSettings.xml
```diff
@@ -665,8 +665,8 @@
 		<member name="interface/theme/icon_and_font_color" type="int" setter="" getter="">
 			The icon and font color scheme to use in the editor.
 			- [b]Auto[/b] determines the color scheme to use automatically based on [member interface/theme/base_color].
-			- [b]Dark[/b] makes fonts and icons light (suitable for dark themes).
-			- [b]Light[/b] makes fonts and icons dark (suitable for light themes). Icon colors are automatically converted by the editor following [url=https://github.com/godotengine/godot/blob/master/editor/editor_themes.cpp#L135]this set of rules[/url].
+			- [b]Dark[/b] makes fonts and icons dark (suitable for light themes). Icon colors are automatically converted by the editor following [url=https://github.com/godotengine/godot/blob/master/editor/editor_themes.cpp#L135]this set of rules[/url].
```
YuriSizov, line 668: I know it was like this before, but I don't think we should be linking to a specific line number, given that the link points to the branch and not a specific commit/revision. It is in fact incorrect already, as line 135 is in a middle of a function.

Linking to the file itself should be fine, but the text needs to be changed appropriately, e.g. "following the set of rules defined here".

YuriSizov (approved):
Aside from the note above, this change seems to be correct. We consider that the editor has a dark theme if `icon_font_color_setting == LIGHT_COLOR`.

AThousandShips:

BNTFryingPan:
just wanted to make sure: this is all set, right?

YuriSizov:
Yep!

akien-mga:

akien-mga:
Cherry-picked for 4.2.1.

PR #85494, opened 2023-11-29
Add multi-selection for `SpriteFramesEditor`

### LINKED ISSUE #1696: Allow Multi-Selection of Frames in SpriteFrames Animations
I'm currently using AnimatedSprites, which use SpriteFrames, to add animations to the user interface of our game project.

I have been modifying the frames of our animations in SpriteFrames, by swapping out old sprites and implementing new ones, as well as moving frames around. While doing this I noticed both moving and deleting can only be done on an individual frame basis, which can make removing or moving sprites in large animations a real chore.

image
_only one frame can be selected at a time_

I would love to see multi-selection of the frames inside the Animations, by using CTRL+LMB and SHIFT+LMB respectively for single selection (current selected frames + clicked frame)  or range selection (current selected frames to clicked frame). This is how most listed items and UI elements work throughout the engine, such as the Filesystem, Scene Tree, Scene Viewer.

This would allow the user to select all frames, and then press the 'Delete' button to remove all of them, or select a few frames, and press the 'Move' buttons to move the selected range of frames around. This would greatly speed up large changes within SpriteFrames Animations, and be friendlier on the Index Fingers of all Godot Devs.

AnimatedSpriteMultiSelectionMockup
I think it should be pretty self-explanatory, but here is a small slideshow of the proposed feature, as I see it. Keep in mind the last frame, where the animation frames get deleted would obviously not look like that (empty frames), but it best showed the 'deletion' process.

It cannot be worked around through script, but it can be worked around with manual labor, by selecting each individual frame and moving/deleting it one by one. Not a great solution, but it does work. Technically.

This would be an enhancement to the UX of an already core part of the game, and one that would be in line with other selection methods inside the editor. This likely shouldn't be left as an add-on.

Hobitus:
Enabled multi-selection in the sprite_frame_editor
This was proposal [#1696](https://github.com/godotengine/godot-proposals/issues/1696)

You can now use multi-selections with CTR to add or remove a frame. Shift to selection until selection.
With a multi-selection you can:

**Copy and Paste multiple-frame**
Since you can't put arrays of frame in the Editor clipboard, I've remove it and saved the data locally. 

**Move left or right selected frame**
For example, if you have 1 and 3 selected and press move right, you will have 2,1,4,3.

**Set frame duration to all selected frame.**
 Takes the first frame for the duration value.

**Add empty frame, after or before the selection**

**Delete multiple frame at the same time**

image

I didn't see a way to add test for this class. Anything I can do?

<i>Bugsquad edit:</i>
- Closes https://github.com/godotengine/godot-proposals/issues/1696

### DIFF
--- a/editor/plugins/sprite_frames_editor_plugin.cpp
+++ b/editor/plugins/sprite_frames_editor_plugin.cpp
@@ -282,7 +282,7 @@ void SpriteFramesEditor::_sheet_add_frames() {
 	const Size2i separation = _get_separation();
 
 	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
-	undo_redo->create_action(TTR("Add Frame"), UndoRedo::MERGE_DISABLE, EditorNode::get_singleton()->get_edited_scene());
+	undo_redo->create_action(TTR("Add Frame"), UndoRedo::MERGE_DISABLE, frames.ptr());
 	int fc = frames->get_frame_count(edited_anim);
 
 	_sheet_sort_frames();
@@ -624,7 +624,7 @@ void SpriteFramesEditor::_file_load_request(const Vector<String> &p_path, int p_
 	}
 
 	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
-	undo_redo->create_action(TTR("Add Frame"), UndoRedo::MERGE_DISABLE, EditorNode::get_singleton()->get_edited_scene());
+	undo_redo->create_action(TTR("Add Frame"), UndoRedo::MERGE_DISABLE, frames.ptr());
 	int fc = frames->get_frame_count(edited_anim);
 
 	int count = 0;
@@ -674,30 +674,57 @@ void SpriteFramesEditor::_load_pressed() {
 void SpriteFramesEditor::_paste_pressed() {
 	ERR_FAIL_COND(!frames->has_animation(edited_anim));
 
+	Ref<ClipboardSpriteFrames> clipboard_frames = EditorSettings::get_singleton()->get_resource_clipboard();
+	if (clipboard_frames.is_valid()) {
+		_paste_frame_array(clipboard_frames);
+		return;
+	}
+
+	Ref<Texture2D> texture = EditorSettings::get_singleton()->get_resource_clipboard();
+	if (texture.is_valid()) {
+		_paste_texture(texture);
+		return;
+	}
+}
+
+void SpriteFramesEditor::_paste_frame_array(const Ref<ClipboardSpriteFrames> &p_clipboard_frames) {
+	if (p_clipboard_frames->frames.is_empty()) {
+		return;
+	}
+
 	Ref<Texture2D> texture;
 	float duration = 1.0;
 
-	Ref<EditorSpriteFramesFrame> frame = EditorSettings::get_singleton()->get_resource_clipboard();
-	if (frame.is_valid()) {
-		texture = frame->texture;
-		duration = frame->duration;
-	} else {
-		texture = EditorSettings::get_singleton()->get_resource_clipboard();
+	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
+	undo_redo->create_action(TTR("Paste Frame(s)"), UndoRedo::MERGE_DISABLE, frames.ptr());
+
+	int undo_index = frames->get_frame_count(edited_anim);
+
+	for (int index = 0; index < p_clipboard_frames->frames.size(); index++) {
+		const ClipboardSpriteFrames::Frame &frame = p_clipboard_frames->frames[index];
+		texture = frame.texture;
+		duration = frame.duration;
+
+		undo_redo->add_do_method(frames.ptr(), "add_frame", edited_anim, texture, duration);
+		undo_redo->add_undo_method(frames.ptr(), "remove_frame", edited_anim, undo_index);
 	}
 
-	if (texture.is_null()) {
-		dialog->set_text(TTR("Resource clipboard is empty or not a texture!"));
-		dialog->set_title(TTR("Error!"));
-		//dialog->get_cancel()->set_text("Close");
-		dialog->set_ok_button_text(TTR("Close"));
-		dialog->popup_centered();
-		return; ///beh should show an error i guess
-	}
+	undo_redo->add_do_method(this, "_update_library");
+	undo_redo->add_undo_method(this, "_update_library");
+	undo_redo->commit_action();
+}
+
+void SpriteFramesEditor::_paste_texture(const Ref<Texture2D> &p_texture) {
+	float duration = 1.0;
 
 	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
-	undo_redo->create_action(TTR("Paste Frame"), UndoRedo::MERGE_DISABLE, EditorNode::get_singleton()->get_edited_scene());
-	undo_redo->add_do_method(frames.ptr(), "add_frame", edited_anim, texture, duration);
-	undo_redo->add_undo_method(frames.ptr(), "remove_frame", edited_anim, frames->get_frame_count(edited_anim));
+	undo_redo->create_action(TTR("Paste Texture"), UndoRedo::MERGE_DISABLE, frames.ptr());
+
+	int undo_index = frames->get_frame_count(edited_anim);
+
+	undo_redo->add_do_method(frames.ptr(), "add_frame", edited_anim, p_texture, duration);
+	undo_redo->add_undo_method(frames.ptr(), "remove_frame", edited_anim, undo_index);
+
 	undo_redo->add_do_method(this, "_update_library");
 	undo_redo->add_undo_method(this, "_update_library");
 	undo_redo->commit_action();
@@ -706,31 +733,39 @@ void SpriteFramesEditor::_paste_pressed() {
 void SpriteFramesEditor::_copy_pressed() {
 	ERR_FAIL_COND(!frames->has_animation(edited_anim));
 
-	if (frame_list->get_current() < 0) {
+	Vector<int> selected_items = frame_list->get_selected_items();
+
+	if (selected_items.is_empty()) {
 		return;
 	}
 
-	Ref<Texture2D> texture = frames->get_frame_texture(edited_anim, frame_list->get_current());
-	if (texture.is_null()) {
-		return;
+	Ref<ClipboardSpriteFrames> clipboard_frames = memnew(ClipboardSpriteFrames);
+
+	for (const int &frame_index : selected_items) {
+		Ref<Texture2D> texture = frames->get_frame_texture(edited_anim, frame_index);
+		if (texture.is_null()) {
+			continue;
+		}
+
+		ClipboardSpriteFrames::Frame frame;
+		frame.texture = texture;
+		frame.duration = frames->get_frame_duration(edited_anim, frame_index);
+
+		clipboard_frames->frames.push_back(frame);
 	}
-
-	Ref<EditorSpriteFramesFrame> frame = memnew(EditorSpriteFramesFrame);
-	frame->texture = texture;
-	frame->duration = frames->get_frame_duration(edited_anim, frame_list->get_current());
-
-	EditorSettings::get_singleton()->set_resource_clipboard(frame);
+	EditorSettings::get_singleton()->set_resource_clipboard(clipboard_frames);
 }
 
 void SpriteFramesEditor::_empty_pressed() {
 	ERR_FAIL_COND(!frames->has_animation(edited_anim));
 
 	int from = -1;
+	Vector<int> selected_items = frame_list->get_selected_items();
 
-	if (frame_list->get_current() >= 0) {
-		from = frame_list->get_current();
-		sel = from;
-
+	if (!selected_items.is_empty()) {
+		from = selected_items[0];
+		selection.clear();
+		selection.push_back(from + 1);
 	} else {
 		from = frames->get_frame_count(edited_anim);
 	}
@@ -738,7 +773,7 @@ void SpriteFramesEditor::_empty_pressed() {
 	Ref<Texture2D> texture;
 
 	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
-	undo_redo->create_action(TTR("Add Empty"), UndoRedo::MERGE_DISABLE, EditorNode::get_singleton()->get_edited_scene());
+	undo_redo->create_action(TTR("Add Empty"), UndoRedo::MERGE_DISABLE, frames.ptr());
 	undo_redo->add_do_method(frames.ptr(), "add_frame", edited_anim, texture, 1.0, from);
 	undo_redo->add_undo_method(frames.ptr(), "remove_frame", edited_anim, from);
 	undo_redo->add_do_method(this, "_update_library");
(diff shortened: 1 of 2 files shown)


Chaosus:
Static check failed, you should apply the clang-format tool to format it properly. After making the change, don't forget to squash commits (see https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#modifying-a-pull-request).

Hobitus:
I feel like the move left and move right command are over complicated. It moves all the selected animation (with gap in selection), which is great, but is it really useful?
For simplicity we could pick the first or last item in the multi-selection and move that one only.

Chaosus:
> How do I apply the clang-format tool?

https://docs.godotengine.org/en/stable/contributing/development/code_style_guidelines.html#using-clang-format-locally

editor/plugins/sprite_frames_editor_plugin.cpp
```diff
@@ -1261,12 +1320,31 @@ void SpriteFramesEditor::_update_library_impl() {
...
 
+	for (int index = 0; index < selections.size(); index++) {
+		int sel = selections[index];
+		if (sel == -1) {
+			selections.remove_at(index);
+			index--;
+		}
+		if (sel >= anim_frame_count) {
+			selections.set(index, anim_frame_count - 1);
+			// since selection are ordered, if we get a frame that is outside of the range
```
dalexeev, line 1336:
```suggestion
			// Since selection is ordered, if we get a frame that is outside of the range
```

editor/plugins/sprite_frames_editor_plugin.h
```diff
@@ -115,7 +115,8 @@ class SpriteFramesEditor : public HSplitContainer {
 	SpinBox *frame_duration = nullptr;
 	ItemList *frame_list = nullptr;
 	bool loading_scene;
-	int sel;
+	Vector<int> selections;
```
dalexeev, line 118:
```suggestion
	Vector<int> selection;
```

or

```suggestion
	Vector<int> selected_frames;
```

editor/plugins/sprite_frames_editor_plugin.cpp
```diff
@@ -1112,15 +1162,18 @@ void SpriteFramesEditor::_frame_list_gui_input(const Ref<InputEvent> &p_event) {
...
 
-void SpriteFramesEditor::_frame_list_item_selected(int p_index) {
+void SpriteFramesEditor::_frame_list_item_selected(int p_index, bool p_selected) {
 	if (updating) {
 		return;
 	}
 
-	sel = p_index;
+	selections = frame_list->get_selected_items();
+	if (selections.is_empty() || p_selected == false) {
```
AThousandShips, line 1171:
```suggestion
	if (selections.is_empty() || !p_selected) {
```

editor/plugins/sprite_frames_editor_plugin.h
Hobitus: There is already a frames_selected variable, so I'll go with selection

editor/plugins/sprite_frames_editor_plugin.h
```diff
@@ -115,7 +115,8 @@ class SpriteFramesEditor : public HSplitContainer {
...
+	Vector<Ref<EditorSpriteFramesFrame>> frame_clipboard;
```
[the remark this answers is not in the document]
Hobitus: I agree that I should use the resource_clipboard instead of my frame_clipboard. We can only set resource in a resource_clipboard, I'll have to go with the EditorSpriteFramesArray suggestion.

In the code I see only 4 place where we set_resource_clipboard. In the inspector dock to copy the current resource. could be anything, in the editor resource picker(can also be anything), and in the animation_library_editor (it saves an anim resource).

editor/plugins/sprite_frames_editor_plugin.cpp
```diff
@@ -674,30 +674,29 @@ void SpriteFramesEditor::_load_pressed() {
...
+		return;
+	}
+
 	Ref<Texture2D> texture;
 	float duration = 1.0;
 
-	Ref<EditorSpriteFramesFrame> frame = EditorSettings::get_singleton()->get_resource_clipboard();
-	if (frame.is_valid()) {
+	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
+	undo_redo->create_action(TTR("Paste Frame"), UndoRedo::MERGE_DISABLE, EditorNode::get_singleton()->get_edited_scene());
```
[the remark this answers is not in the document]
Hobitus: Added the (s) here and for the Copy and Paste command too.

editor/plugins/sprite_frames_editor_plugin.h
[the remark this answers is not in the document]
Hobitus: Rename the EditorSpriteFramesFrame to EditorSpriteFrame (I feel it makes more sense), and created a EditorSpriteFrameArray, that inherits resources to put it in the EditorSettings clipboard.

editor/plugins/sprite_frames_editor_plugin.cpp
Hobitus: done

editor/plugins/sprite_frames_editor_plugin.cpp
Hobitus: done

editor/plugins/sprite_frames_editor_plugin.h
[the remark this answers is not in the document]
Hobitus: I've also put back the check if we have a Texture in the clipboard, and create a frame from it.

editor/plugins/sprite_frames_editor_plugin.h
```diff
@@ -49,14 +49,27 @@
...
 
 public:
 	Ref<Texture2D> texture;
 	float duration;
 };
 
+class EditorSpriteFrameArray : public Resource {
+	GDCLASS(EditorSpriteFrameArray, Resource);
+
+	Vector<Ref<EditorSpriteFrame>> frames;
```
dalexeev, line 63: Sorry, I worded it poorly. There is no point in having both `EditorSpriteFrame` and `EditorSpriteFrameArray` since they are not exposed. I meant something like:

```c++
class EditorSpriteFrameArray : public Resource {
    GDCLASS(EditorSpriteFrameArray, Resource);

public:
    struct Frame {
        Ref<Texture2D> texture;
        float duration;
    }
    Vector<Frame> frames;
};
```

Hobitus: The main class is SpriteFramesEditor, so I didn't use EditorSpriteFrames, would be confusing.
I picked ClipboardSpriteFrames instead.

KoBeWi:
Deleting frames should also remove selection, otherwise it looks confusing.

editor/plugins/sprite_frames_editor_plugin.cpp
```diff
@@ -674,30 +674,58 @@ void SpriteFramesEditor::_load_pressed() {
...
+		return;
+	}
+
 	Ref<Texture2D> texture;
 	float duration = 1.0;
 
-	Ref<EditorSpriteFramesFrame> frame = EditorSettings::get_singleton()->get_resource_clipboard();
-	if (frame.is_valid()) {
+	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
+	undo_redo->create_action(TTR("Paste Frame(s)"), UndoRedo::MERGE_DISABLE, EditorNode::get_singleton()->get_edited_scene());
```
KoBeWi, line 700:
```suggestion
	undo_redo->create_action(TTR("Paste Frame(s)"), UndoRedo::MERGE_DISABLE, frames.ptr());
```
If frames are external resource, they would be wrongly associated with the opened scene.

editor/plugins/sprite_frames_editor_plugin.cpp
```diff
@@ -674,30 +674,58 @@ void SpriteFramesEditor::_load_pressed() {
...
+}
+
+void SpriteFramesEditor::_paste_texture(const Ref<Texture2D> &p_texture) {
+	float duration = 1.0;
+
 	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
-	undo_redo->create_action(TTR("Paste Frame"), UndoRedo::MERGE_DISABLE, EditorNode::get_singleton()->get_edited_scene());
-	undo_redo->add_do_method(frames.ptr(), "add_frame", edited_anim, texture, duration);
-	undo_redo->add_undo_method(frames.ptr(), "remove_frame", edited_anim, frames->get_frame_count(edited_anim));
+	undo_redo->create_action(TTR("Paste Texture"), UndoRedo::MERGE_DISABLE, EditorNode::get_singleton()->get_edited_scene());
```
KoBeWi, line 722:
```suggestion
	undo_redo->create_action(TTR("Paste Texture"), UndoRedo::MERGE_DISABLE, frames.ptr());
```
Same.
I see originally it was different, but that was wrong.

editor/plugins/sprite_frames_editor_plugin.cpp
```diff
@@ -706,31 +734,41 @@ void SpriteFramesEditor::_paste_pressed() {
...
 	}
 
-	Ref<EditorSpriteFramesFrame> frame = memnew(EditorSpriteFramesFrame);
-	frame->texture = texture;
-	frame->duration = frames->get_frame_duration(edited_anim, frame_list->get_current());
+	Ref<ClipboardSpriteFrames> clipboard_frames = memnew(ClipboardSpriteFrames);
 
-	EditorSettings::get_singleton()->set_resource_clipboard(frame);
+	for (int index = 0; index < selected_items.size(); index++) {
+		int frame_index = selected_items[index];
```
KoBeWi, line 746:
```suggestion
	for (const int &frame_index : selected_items) {
```

editor/plugins/sprite_frames_editor_plugin.cpp
```diff
@@ -1129,18 +1213,24 @@ void SpriteFramesEditor::_frame_duration_changed(double p_value) {
...
-	Ref<Texture2D> texture = frames->get_frame_texture(edited_anim, index);
-	float old_duration = frames->get_frame_duration(edited_anim, index);
-
 	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
 	undo_redo->create_action(TTR("Set Frame Duration"), UndoRedo::MERGE_ENDS, EditorNode::get_singleton()->get_edited_scene());
-	undo_redo->add_do_method(frames.ptr(), "set_frame", edited_anim, index, texture, p_value);
-	undo_redo->add_undo_method(frames.ptr(), "set_frame", edited_anim, index, texture, old_duration);
+
+	for (int selected_index = 0; selected_index < nb_selected_items; selected_index++) {
+		int index = selection[selected_index];
```
KoBeWi, line 1225:
```suggestion
	for (const int &index  : selection) {
		int index = selection[selected_index];
```

editor/plugins/sprite_frames_editor_plugin.cpp
```diff
@@ -1129,18 +1213,24 @@ void SpriteFramesEditor::_frame_duration_changed(double p_value) {
 		return;
 	}
 
-	int index = sel;
-	if (index < 0) {
+	int nb_selected_items = selection.size();
+	if (nb_selected_items == 0) {
```
KoBeWi, line 1217:
```suggestion
	if (selection.is_empty()) {
```

Hobitus:
@KoBeWi 
> Deleting frames should also remove selection, otherwise it looks confusing.

From what I tested with the current build of godot (4.2.1). it works like what you see with my PR.
-Pasting new frame keeps the old selection. 
-undoing doesn't change selection. It keeps what you currently have.
-Deleting frame keep the current selection ( 3 frames with the 1 selected, press delete) and its still the 1 that is selected.

You want me to change the current behavior?

KoBeWi:
Looks like the current behavior is not that great :/
You could leave it as is then, it can be improved in a later PR.

editor/plugins/sprite_frames_editor_plugin.cpp
```diff
@@ -674,30 +674,58 @@ void SpriteFramesEditor::_load_pressed() {
 void SpriteFramesEditor::_paste_pressed() {
 	ERR_FAIL_COND(!frames->has_animation(edited_anim));
 
+	Ref<ClipboardSpriteFrames> clipboard_frames = EditorSettings::get_singleton()->get_resource_clipboard();
+
```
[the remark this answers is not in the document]
Hobitus: removed the space

editor/plugins/sprite_frames_editor_plugin.cpp
Hobitus: There was many create action in the files. changed them all to use frames.ptr()

editor/plugins/sprite_frames_editor_plugin.cpp
```diff
@@ -674,30 +674,58 @@ void SpriteFramesEditor::_load_pressed() {
...
-	if (frame.is_valid()) {
+	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
+	undo_redo->create_action(TTR("Paste Frame(s)"), UndoRedo::MERGE_DISABLE, EditorNode::get_singleton()->get_edited_scene());
+
+	int undo_index = frames->get_frame_count(edited_anim);
+
+	for (int index = 0; index < p_clipboard_frames->frames.size(); index++) {
+		const ClipboardSpriteFrames::Frame *frame = &p_clipboard_frames->frames[index];
 		texture = frame->texture;
 		duration = frame->duration;
```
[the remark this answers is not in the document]
Hobitus: done

editor/plugins/sprite_frames_editor_plugin.cpp
Hobitus: done

editor/plugins/sprite_frames_editor_plugin.cpp
Hobitus: done

editor/plugins/sprite_frames_editor_plugin.cpp
Hobitus: done

Hobitus:
I'll leave the selection behavior like it is now, but I agree with you it's not great.

KoBeWi (approved):

akien-mga:

hsandt:
Nice, I'll have to  try this one!

While I agree that deleting selected entries should clear selection in general, I see a point, on single delete, in selecting the new sprite replacing the deleted sprite at the same slot (or the one before if deleting the last sprite), so user can chain delete if they want to. This would have to be combined with https://github.com/godotengine/godot-proposals/issues/9387 so user keeps focus and can chain usage of the Delete key, or do some other operations like changing the Frame Duration multiplier (it really shines when doing many of these operations - like "merging" identical sprite frames together by removing all the duplicates and setting the Frame Duration multiplier).

Currently, clicking the bin button actually does that, so you can chain delete by clicking on the bin button multiple times. Now it's a technique I learned *because batch selection was not implemented* and it becomes much less useful now that we can delete multiple sprite frames at once. However it's worth considering the current behavior of single delete, whether with Delete key or bin button, to decide whether or not to maintain selection.

That said, in the case of multi-selection delete, I agree it's confusing. In fact, in an app like Ubuntu's Image Viewer / Eye of Gnome, when the Image Gallery is opened at the bottom, if you multi-select images and Delete them, the selection will fallback back to the image right after the last selected deleted entry (or before if last entry was deleted). So still a unique selection.

To sum up, we may want to preserve the single delete behavior of falling back to the next available entry, but when deleting multiple frames, it may be less confusing to fall back selection to just the next available entry after the last selected frame (whether "last" means "last selected" or "last in chronological order" would remain to be decided).

PR #85499, opened 2023-11-29
GDAP: Allow for custom maven repos to be private, accepting a set of credentials for each url

rhighs:
This PR allows for the definition of private custom maven dependency repositories in the context of android plugins. This needed a way to pass a set of credentials for a given url, had a look at the code and found out the parser used for `.gdap` config files supports the parsing of Dictionaries. With this in mind I have changed the way the `custom_maven_repos` property is represented in android plugin config files and made it an array of dictionaries instead of strings. If credentials are required for a repo url, there should be a "credentials" entry otherwise it is acknowledged as a public repo.

example gdap:
```
[config]
name="SimplePlugin"
binary_type="local"
binary="app-debug.aar"

[dependencies]
local=[]
remote=["com.simple.plugin:legacy-impl:1.1.0"]
custom_maven_repos=[{
  "url": "https://public-repo.dev/public-aars"  # <-- public repo
},
{
  "url": "https://private-repo.dev/private-aars"
  "credentials": {
    "username": "godot",
    "password": "duckduckgodot"
  }
}]
```

I have made a [proposal](https://github.com/godotengine/godot-proposals/issues/8443) already, but having the change ready and available at hand I thought of opening a PR just for overviewing what changes the implementation would bring. Turns out it's quite simple in practice. Only thing that leaves me wondering tho, is how great is this in terms of retro-compatibility.

### DIFF
--- a/platform/android/export/export_plugin.cpp
+++ b/platform/android/export/export_plugin.cpp
@@ -2932,7 +2932,7 @@ Error EditorExportPlatformAndroid::export_project_helper(const Ref<EditorExportP
 
 		Vector<String> android_libraries;
 		Vector<String> android_dependencies;
-		Vector<String> android_dependencies_maven_repos;
+		Vector<Dictionary> android_dependencies_maven_repos;
 
 #ifndef DISABLE_DEPRECATED
 		Vector<PluginConfigAndroid> enabled_plugins = get_enabled_plugins(p_preset);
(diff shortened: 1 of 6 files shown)


platform/android/export/export_plugin.cpp
```diff
@@ -3407,6 +3411,26 @@ void EditorExportPlatformAndroid::get_platform_features(List<String> *r_features
 void EditorExportPlatformAndroid::resolve_platform_feature_priorities(const Ref<EditorExportPreset> &p_preset, HashSet<String> &p_features) {
 }
 
+String EditorExportPlatformAndroid::_serialize_android_dependencies_maven_repos(const Vector<Dictionary> &p_maven_repos_definitions) const {
+	// serialize a custom maven repo into: "url=<url>,username=<username>,password=<password>"
+	Vector<String> serialized_android_custom_maves_repos;
+	for (const auto &android_dependencies_maven_repo : p_maven_repos_definitions) {
```
bruvzg, line 3417:
```suggestion
	for (const String &android_dependencies_maven_repo : p_maven_repos_definitions) {
```

We do not use `auto`, see https://docs.godotengine.org/en/stable/contributing/development/cpp_usage_guidelines.html#auto-keyword.

m4gr3d:
@rhighs Starting with Godot 4.2, the `gdap` packaging format has been [deprecated](https://docs.godotengine.org/en/latest/tutorials/platform/android/android_plugin.html#android-plugin) and its logic is mostly kept around for compatibility purpose, as such no changes (except bug fixes) are being accepted for it.

This proposal should instead be applied to the [new v2 packaging format](https://docs.godotengine.org/en/latest/tutorials/platform/android/android_plugin.html#v2-packaging-format).
Based on your description, it sounds like we'd want to update the [`EditorExportPlugin#_get_android_dependencies_maven_repos(...)`](https://docs.godotengine.org/en/latest/classes/class_editorexportplugin.html#class-editorexportplugin-private-method-get-android-dependencies-maven-repos) api to allow returning a `Dictionary` of maven repos and their credentials.

rhighs:
@m4gr3d sounds good, thanks for the info 👍
I'll open another PR :)

Issue #85081, opened 2023-11-19
[TRACKER] Issues with preload cache


KoBeWi:
### Godot version

4.2 rc1

### System information

Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1060 (NVIDIA; 30.0.15.1403) - Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz (8 Threads)

### Issue description

When you `preload()` a script or a scene (or possibly anything else), that resource will stay in memory in its original form and it's difficult to remove or update it. This leads to all sorts of issues, particularly when the initial loading fails the resource will be broken until editor restart or in worst case until the user performs some magic to fix the cached version (when restarting doesn't work).

Here's the list of issues caused by this behavior:

- [x] https://github.com/godotengine/godot/issues/84981
- [x] https://github.com/godotengine/godot/issues/79840
- [x] https://github.com/godotengine/godot/issues/80324
- [x] https://github.com/godotengine/godot/issues/81498
- [x] https://github.com/godotengine/godot/issues/81615
- [x] https://github.com/godotengine/godot/issues/79545

These two issues are solved, but they were related, so adding for context:
- [x] https://github.com/godotengine/godot/issues/72854
- [x] https://github.com/godotengine/godot/issues/80812

Related PRs (they don't fix the problem, but band-aid it partially):
- https://github.com/godotengine/godot/pull/84167 - helps reloading broken scenes
- https://github.com/godotengine/godot/pull/85037 - prevents corruption issues when moving class files

### Steps to reproduce

1. Use `preload()`
2. Pray it won't cause problems (it will 😉)

### Minimal reproduction project

N/A

akien-mga:
Potential candidates to add to the tracker, or close as duplicates of other issues:
- #79545 
- #85902
- #85907

Swarkin:
Changing all `preload`s to `load` seems to be a viable workaround for now.
My project doesn't seem to be corrupting the scenes anymore with that change.

ndbn:
It seems very relative to this.
Simple scene with only Line2D root node and attached script:
```
extends Line2D
class_name FishingLineScene

const FISHING_LINE: PackedScene = preload("res://scenes/player/fishing/fishing_line.tscn")

static func createScene() -> FishingLineScene:
  var scene: FishingLineScene = FISHING_LINE.instantiate()
  return scene

func add_point_from_global(global_point: Vector2) -> void:
[...]

Issue #70985, opened 2023-01-06
Circular dependencies of scenes (by preload) cause a script can not be loaded


jinyangcruise:
### Godot version

v4.0.beta10.official [d0398f62f]

### System information

Vulkan API 1.3.205 - Using Vulkan Device #0: NVIDIA - NVIDIA GeForce RTX 3060 Ti

### Issue description

My project contains two scenes "cell.tscn"（attached with a script "cell.gd") and "swap_pad.tscn" (attached with a script "swap_pad.gd"). Everytime I open this project, the "cell.tscn" always lose its script "cell.gd" with an error log : 
```
ERROR: res://src/swapPad/cell.tscn:9 - Parse Error: [ext_resource] referenced non-loaded resource at: res://src/swapPad/cell.gd
   at: _parse_ext_resource (scene/resources/resource_format_text.cpp:178)
```
I need to re-attach this script to the scene to fix it temporarily but next time this problem happen again.
I finally figured out this problem happens if two scripts have circular dependencies. I don't know if this problem is or not a bug of Godot. If not, why can I fix it temporarily by re-attach the script to the scene. 

Content of "cell.gd":
```
extends CenterContainer

var _swapPad : SwapPad = null
```

Content of "swap_pad.gd"
```
extends Control
class_name SwapPad

func _ready():
	var cellClass := preload("res://src/swapPad/cell.tscn")
```

### Steps to reproduce

0.download minimal reproduction project below
1. open project
2. open cell.tscn（src/swapPad）in godot and you will find this scene have no script
3. attach cell.gd to cell.tscn
4. save and close project
5. open project again
6. open cell.tscn（src/swapPad）in godot and you will find this scene still have no script

### Minimal reproduction project

[test1.zip](https://github.com/godotengine/godot/files/10358858/test1.zip)

Chaosus:
cc @adamscott

adamscott:
I confirm this issue. This does not seem to be a regression. I'm investigating.

dalexeev:
See also:

* #74253
* #77436

adamscott:
For more info on this issue and why it's currently difficult to fix it, see the comments of my closed PR https://github.com/godotengine/godot/pull/71004.

keir:
I put up https://github.com/godotengine/godot/pull/80854 to help make the user experience of this engine limitation easier to debug/deal with.

Muhammad-Fiqri:
[...]

PR #67714, opened 2022-10-21
Fix cyclic references in GDScript 2.0

adamscott:
# TL;DR
This PR makes possible to use cyclic references in GDScript 2.0, i.e. preload class (or use by class name) `A` in `B` and `B` in `A`, preload cyclic scenes, use typing in addons.

# Summary
This PR tweaks the engine to accept and support cyclic references.

# What it tries to solve
The current GDScript engine blocks cyclic references as this would normally lead to memory leaks. Why? Because cyclic references artificially increase the refcount of gdscripts.

## Standard case
Let's say `Main` loads `A` that loads `B`.

Here, `Main` reference count is 1 (referenced by the engine), `A` is 1 (referenced by `Main`) and `B` is 1 (referenced by `A`).

When the engine unloads, `Main` reference count will be decremented by 1 and will be deleted, because its reference count will hit 0. `A` and `B` will also be deleted because `A` lost his reference from `Main` and `B` will lose its reference from `A` later.

## Current cyclic reference issue
Currently, the engine makes it so that using cyclic references is impossible. Why, you say?

Let's bring back our example, but with a cyclic reference. Let's say `Main` loads `A` that loads `B` that loads back `A`.

Here, `Main` reference count is 1 (referenced by the engine), `A` is 2 (referenced by `Main` and `B`) and `B` is 1 (referenced by `A`).

Do you see the issue? When `Main` will be deleted, there will be no way to delete `A` and `B`. `A` will lose its reference to `Main` and will decrement, but it will not be deleted as `A` reference count is now 1. `B` is not deleted because it didn't lose its reference to `A`.

So, the answer is: the engine makes cyclic references impossible due to memory leaks concerns. That explains errors thrown by the parser or the analyzer like `Constant value uses script from "res://b.gd" which is loaded but not compiled`.

# Proposed solution
## Cyclic dependency detection
There is no way to use the reference count as a mean to know if a cyclic `Ref<RefCounted>` needs to be disposed of. If `A` and `B` refers to themselves a number of times, the reference count will skyrocket and it would not be easy to track down the exact number of times the count has been incremented.

If, someday or someone, comes up with a reliable way to detect these, it would be the ideal solution.

Meanwhile, to fix this issue, I propose to use the existing `SelfList<GDScript> script_list` found in the `GDScript` class. This list holds each `GDScript` instance.

Take this table as basis for my following example:

| Script list | Script constants |
|-|-|
| `Main` | `A`, `B`, `X` |
| `A` | |
| `B` | `C` |
[...]

PR #85501, opened 2023-11-29
Stop caching packed scenes in GDScript cache (on preload)

### LINKED ISSUE #79545: Scene is invalid/corrupt according to the Godot editor, but works normally ingame

### LINKED ISSUE #79840: Godot fails to open a scene, claiming a script is missing, but the script is there!
### Godot version

### System information

Godot v4.1.1.stable - Arch Linux #1 SMP PREEMPT_DYNAMIC Thu, 06 Jul 2023 18:35:54 +0000 - Vulkan (Forward+) - dedicated AMD Radeon RX 6800 XT (RADV NAVI21) () - AMD Ryzen 9 3900X 12-Core Processor (24 Threads)

### Issue description

When I was working on my game, changing stuff in scripts, at some point the game started crashing during normal play, I've found that it throws an error... that makes absolutely no sense. It says a property doesn't exist on a scene that uses a script that has that property.

Then opening a scene that uses this script fails claiming the script is missing.
It's not missing though - it's there, and other scenes using it open fine.
What's even weirder is that all of these problems aren't there if I go back one commit in my game's git repository, even though the faulty commit didn't affect the feral script or the troublesome scene.

Here's a thorough video demonstration of the issue. I tried a few versions of Godot and had the same results. I've recorded the video using an Arch-Linux provided binary for 4.1.1 stable, but I tested again on an official build from Godotengine.org and had the exact same problem.

I tried clearing .godot to force a full project reimport, I tired running the editor with `_sc_` (self-contained mode) and renaming existing `editor_data` folder to have a clean slate - no change.

Please watch the video for a deeper context and more detail:
https://youtu.be/gmFhS-m0AUM

### Steps to reproduce

The only reproduction steps I can provide is to open my game's project and try it on your system:

1. Follow the guide here: https://codeberg.org/Liblast/Liblast#_how_to_edit_the_game TL;DR: clone the repository with git-LFS, or it won't work.
2. Test the `main` branch and see if the game works (it should as it does in my video).
3. Then checkout this commit and see if it fails like it does in my video: https://codeberg.org/Liblast/Liblast/commit/915cbc735f2a416bc2985e02ea0c2bc838c8e012

N/A

### LINKED ISSUE #84981: Moving or renaming scenes in the editor or changing properties in custom resources causes scene corruption
### Godot version

4.2.beta4

### System information

Windows 11, AMD Ryzen 5 3600, RTX 2060

### Issue description

The Godot engine will lose references to scenes UID cache, causing scenes to register as corrupted. Animations will still work and overall functionality will be retained, but the user loses the ability to open the corrupted scenes. The console will show the below error code. Because scene functionality will be preserved, I believe this has to do with how Godot handles the UID caching of scenes with custom resources

Screenshot 2023-11-06 003745

### Steps to reproduce

This can happen when removing/renaming files whether inside or outside the Godot editor (more common outside the editor). It can also happen if you change, rename or delete properties from custom resources.

- Create a custom resource, add some properties
- Create a script that exports a variable of type custom resource
- Create a scene that uses that script. Assign an instance of the custom resource to the script
- Delete a property from custom resource script
- Move the scene in either the Godot editor or Explorer
- OR -
- Save the script, not the scene

N/A

Jordyfel:
Adresses #85081

This code was added in #67714, with the goal of resolving problems with cyclic references is `preload()`.

However, cyclic scene references still don't work, as illustrated by #70985 and other preload related issues.

Being resources, scenes don't support cyclic references, and caching seems to just make the problem appear a little bit later, and make it harder to debug (unlike with scripts, where this should be fine, because `get_shallow_script()` is designed to deal with the cyclic reference problem).

IMO this problem cannot be solved from outside core resource functionality and the GDScript module should not try. Pinging @adamscott as the original implementer, please do correct me if I have misunderstood the intent of this code.

Most posted issues regarding preload have no reproduction steps and just include already corrupted scenes, which makes it very hard to tell which this change fixes, however it should improve recovery and consistency of behavior as scenes will for sure be reloaded when the script is recompiled.

*Bugsquad edit:* Fixes #79545 Fixes #84981 Fixes #79840

### DIFF
--- a/modules/gdscript/gdscript_analyzer.cpp
+++ b/modules/gdscript/gdscript_analyzer.cpp
@@ -4144,21 +4144,14 @@ void GDScriptAnalyzer::reduce_preload(GDScriptParser::PreloadNode *p_preload) {
 		} else {
 			// TODO: Don't load if validating: use completion cache.
 
-			// Must load GDScript and PackedScenes separately to permit cyclic references
-			// as ResourceLoader::load() detect and reject those.
+			// Must load GDScript separately to permit cyclic references
+			// as ResourceLoader::load() detects and rejects those.
 			if (ResourceLoader::get_resource_type(p_preload->resolved_path) == "GDScript") {
 				Error err = OK;
 				Ref<GDScript> res = GDScriptCache::get_shallow_script(p_preload->resolved_path, err, parser->script_path);
 				p_preload->resource = res;
 				if (err != OK) {
 					push_error(vformat(R"(Could not preload resource script "%s".)", p_preload->resolved_path), p_preload->path);
-				}
-			} else if (ResourceLoader::get_resource_type(p_preload->resolved_path) == "PackedScene") {
-				Error err = OK;
-				Ref<PackedScene> res = GDScriptCache::get_packed_scene(p_preload->resolved_path, err, parser->script_path);
-				p_preload->resource = res;
-				if (err != OK) {
-					push_error(vformat(R"(Could not preload resource scene "%s".)", p_preload->resolved_path), p_preload->path);
 				}
 			} else {
 				p_preload->resource = ResourceLoader::load(p_preload->resolved_path);
--- a/modules/gdscript/gdscript_cache.cpp
+++ b/modules/gdscript/gdscript_cache.cpp
@@ -37,7 +37,6 @@
 
 #include "core/io/file_access.h"
 #include "core/templates/vector.h"
-#include "scene/resources/packed_scene.h"
 
 bool GDScriptParserRef::is_valid() const {
 	return parser != nullptr;
@@ -139,13 +138,6 @@ void GDScriptCache::move_script(const String &p_from, const String &p_to) {
 		return;
 	}
 
-	for (KeyValue<String, HashSet<String>> &E : singleton->packed_scene_dependencies) {
-		if (E.value.has(p_from)) {
-			E.value.insert(p_to);
-			E.value.erase(p_from);
-		}
-	}
-
 	if (singleton->parser_map.has(p_from) && !p_from.is_empty()) {
 		singleton->parser_map[p_to] = singleton->parser_map[p_from];
 	}
@@ -172,15 +164,6 @@ void GDScriptCache::remove_script(const String &p_path) {
 	if (singleton->cleared) {
 		return;
 	}
-
-	for (KeyValue<String, HashSet<String>> &E : singleton->packed_scene_dependencies) {
-		if (!E.value.has(p_path)) {
-			continue;
-		}
-		E.value.erase(p_path);
-	}
-
-	GDScriptCache::clear_unreferenced_packed_scenes();
 
 	if (singleton->parser_map.has(p_path)) {
 		singleton->parser_map[p_path]->clear();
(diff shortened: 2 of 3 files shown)


adamscott:
Oh, sorry, I did not see your ping! I'll take the time to investigate.

akien-mga:
I just pinged you 3 min ago so you're not too late :P

adamscott (approved):
To be honest, if I remember correctly, this was legacy code from when I tried to enable `PackedScene`s cyclic dependencies. Unfortunately, this idea was shut down, but the code stayed.

The all the tests pass and I even tested a small project. Everything seems to still work.

Great job! Thanks!

akien-mga:
Thanks!

Issue #83460, opened 2023-10-16
Manipulating the scene tree gets progressively slower with more nodes added to the tree in the editor.


viksl:
### Godot version

4.2 beta 1

### System information

Windows 11 - Vulkan - Nvidia RTX 4070 - intel i5 13600KF

### Issue description

The more nodes there are in the tree the slower working with the tree gets. I don't even have that many and moving one node in my tree in my game inside editor takes seconds (editor freezes for several seconds until it moves the node in the tree (not even moving it in space).

Link video: https://youtu.be/SAilvoE_ACc
(check how I release the mouse and move it away and how long it takes to make nodes move)

Use the MRP, there are two scenes, main and test_no_mesh_instances, feel free to use either just understand the main has an extreme amount of mesh instances in one places, the latter uses only Node3D hiearchy so it should be a lot less intensive on HW.
I've also purposely added a lot of nodes just to make the issue clear, in my game I don't have even nearly as many nodes but the freeze times are even longer than in MRP.

Please note my HW specs, it's not the highest end but certainly no slug either.

### Steps to reproduce

Steps:
1. Open the MRP
2. Open one of the scenes
3. Add a Node3D, Camera3D, WorldENvironment or any node into the sceen tree
4. Move the tree from last position to the very top of the tree for example, then move it elsewhere if you want
5. Observer the editor freezing as it's moving your node in the sceen tree.
(6. if you have Frame Time open it will show yo ua spike for GPUT TIme and maybe even CPU Time in red after it moves the nodei n the tree)

### Minimal reproduction project

[ManyNodesSlowTreeMRP.zip](https://github.com/godotengine/godot/files/12921177/ManyNodesSlowTreeMRP.zip)

EDIT: Not just moving but also simply deleting or renaming a node takes forever too

EDIT2: This has been discussed in the dev chat (rocket chat) and the issue is due to how the referencing works. When you do any of the operations mentioned above the Editor will loop through all nodes recursively (meaning all their children and children of children and so on) and loops through all their properties which has some expensive operations (casting StringName or something like that and such) which eventually starts to take a lot of time. In a more advanced test the freeze took up to 17 seconds on the above mentioned CPU which has frequency up to 5.1 GHz wtih RAM 6000 MHz so not a particularly slow machine to begin with), currently couple suggestions were mentioned but were brought down due to potentiallyy causing issues/slowdowns in other areas so it seems no on right now knows how to solve this if I understood it correctly?

[...]

PR #85502, opened 2023-11-29
Update NodePaths only in built-in resources

### LINKED ISSUE #84910: Reparenting nodes is extremely slow in certain situations
### Godot version

4.2 beta6

### System information

### Issue description

https://github.com/godotengine/godot/assets/2223172/8f2c5b00-f440-4403-9a62-07316a63cc10

The culprit is one of the recursive check PRs. I profiled the editor and in the case above most of the time is spent in `get_property_list()` method of TileAtlasSource, called by `_check_node_path_recursive()` in SceneTreeDock. Apparently this tile method is very slow and here it's called tens of times.

Not sure what the fix would be, maybe the editor should skip external resources (is it safe to assume they don't point to anything on the scene?). Or we could have a list of exceptions that are known to be slow and don't have NodePaths.

### Steps to reproduce

1. Open the MRP
2. Move Node2D under TileMap

[LagSet.zip](https://github.com/godotengine/godot/files/13355862/LagSet.zip)

KoBeWi:
Fixes #84910

### DIFF
--- a/editor/scene_tree_dock.cpp
+++ b/editor/scene_tree_dock.cpp
@@ -1812,8 +1812,6 @@ bool SceneTreeDock::_check_node_path_recursive(Node *p_root_node, Variant &r_var
 			}
 		} break;
 
-// FIXME: This approach causes a significant performance regression, see GH-84910.
-#if 0
 		case Variant::OBJECT: {
 			Resource *resource = Object::cast_to<Resource>(r_variant);
 			if (!resource) {
@@ -1822,6 +1820,11 @@ bool SceneTreeDock::_check_node_path_recursive(Node *p_root_node, Variant &r_var
 
 			if (Object::cast_to<Animation>(resource)) {
 				// Animation resources are handled by animation editor.
+				break;
+			}
+
+			if (!resource->is_built_in()) {
+				// For performance reasons, assume that scene paths are no concern for external resources.
 				break;
 			}
 
@@ -1841,9 +1844,7 @@ bool SceneTreeDock::_check_node_path_recursive(Node *p_root_node, Variant &r_var
 					undo_redo->add_undo_property(resource, propertyname, old_variant);
 				}
 			}
-			break;
-		};
-#endif
+		} break;
 
 		default: {
 		}


Mickeon:
It's been a while, could use another rebase?

KoBeWi:
Why? There are no conflicts and the code is unlikely to have stopped working.

Mickeon (approved):
🤷 I myself think this is a good idea then.

KoBeWi:
Note #85517, which "fixes" the same issue by disabling functionality.

Also relevant discussion from chat.
image
(pasting as image, because it's from private channel)

Mickeon:
My opinion is, I think the functionality should be restored but limited as this PR does. Checking only for built-in resources' NodePaths is not just performant, but much, much safer as it pretty much ensures the updated NodePath points to a valid node in the scene. If it also affected external Resources there would be no way of guaranteeing it.

KoBeWi:
Rebased and added revert.

editor/scene_tree_dock.cpp
```diff
@@ -1843,7 +1846,6 @@ bool SceneTreeDock::_check_node_path_recursive(Node *p_root_node, Variant &r_var
 			}
 			break;
 		};
```
akien-mga, line 1848: Just a consistency nitpick.
```suggestion
		} break;
```

akien-mga (approved):
Looks good to me, I think we can squash the commits.

akien-mga:
Thanks!

ShadyShlowpoke:
I'm on Godot 4.3 (beta1) and while in the editor adding/moving/removing nodes is still painfully slow when many nodes are in the scene. Is this still being worked on?

KoBeWi:
The issue linked to this pull request is closed. If the problem still exists in some form, see if it was already reported and open a new issue if it wasn't.

ShadyShlowpoke:
This may be related to #83460, which is still open. Editor freezes when adding, moving, or removing nodes in a scene with many nodes, even if they are just built-in nodes like Node2Ds. The more nodes, the longer the freeze.

PR #85503, opened 2023-11-29
Fix the documentation of `Bone2D::apply_rest`

20kdc:
`apply_rest`'s current description is highly misleading. `Stores the node's current transforms in [member rest].` implies `rest = transform` when it in fact performs `transform = rest`.

This change is compatible with, and should be backported to, `3.x` along with to any relevant engine branches.

### DIFF
--- a/doc/classes/Bone2D.xml
+++ b/doc/classes/Bone2D.xml
@@ -15,7 +15,7 @@
 		<method name="apply_rest">
 			<return type="void" />
 			<description>
-				Stores the node's current transforms in [member rest].
+				Resets the bone to the rest pose. This is equivalent to setting [member Node2D.transform] to [member rest].
 			</description>
 		</method>
 		<method name="get_autocalculate_length_and_angle" qualifiers="const">


doc/classes/Bone2D.xml
```diff
@@ -15,7 +15,7 @@
 		<method name="apply_rest">
 			<return type="void" />
 			<description>
-				Stores the node's current transforms in [member rest].
+				Resets the bone to the rest pose. Exactly equivalent to [code]transform = rest[/code].
```
Mickeon, line 18: We could be more explicit without needing a code sample.
```suggestion
				Resets the bone to the rest pose. This is equivalent to setting [member Node2D.transform] to [member rest].
```

Mickeon (approved):
This is correct and has been correct for a while. I'm surprised no one noticed.

doc/classes/Bone2D.xml
20kdc: Seems fine, committed suggestion

Mickeon:
Just one more step. Squash the additional commit (https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase)

20kdc:
> Just one more step. Squash the additional commit (https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase)

Done.

Mickeon (approved):
The more accurate documentation the merrier 👍

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

PR #82113, opened 2023-09-22
Allow Editor to reload external changes of scripts

nongvantinh:
Allow Editor to reload external changes of scripts

Fixes: https://github.com/godotengine/godot/issues/50163

PR #40598, opened 2020-07-22
GDScript 2.0 (again)

vnen:
This one includes and complements the other PR: #39093. Most changes are outlined on that description.

All the things marked as "broken" there are now implemented. This includes type checks, code completion, warnings, language server (although I could not test this), and the mentioned optimizations (constant folding and avoiding allocating an array when `range` is used as a `for` loop iterator).

`setget` is now replaced by properties, implemented as outlined in the proposal: https://github.com/godotengine/godot-proposals/issues/844.

I've tried to do a bunch of tests to find potential issues. I solved a lot of crashes and incorrect behavior along the way and this should be mostly stable. However I won't be surprised if someone can find other crash cases, or infinite loops (especially on completion), or wrong behavior. If you do, please report an issue and I'll solve it when possible.

There are still a lot to improve but this should be at the same level of what it was before (although I know completion is not super great yet, I intend to revisit the code and overall improve the detection and listing).

PR #85504, opened 2023-11-29
Fix C# script loader does not work reliably

nongvantinh:
Fixes C# script loader does not work reliably, paving the way for #82113

### DIFF
--- a/modules/mono/csharp_script.cpp
+++ b/modules/mono/csharp_script.cpp
@@ -2867,8 +2867,11 @@ Ref<Resource> ResourceFormatLoaderCSharpScript::load(const String &p_path, const
 	ERR_FAIL_COND_V_MSG(err != OK, Ref<Resource>(), "Cannot load C# script file '" + p_path + "'.");
 #endif
 
-	scr->set_path(p_original_path);
+	// Only one instance of a C# script is allowed to exist.
+	ERR_FAIL_COND_V_MSG(!scr->get_path().is_empty() && scr->get_path() != p_original_path, Ref<Resource>(),
+			"The C# script path is different from the path it was registered in the C# dictionary.");
 
+	scr->set_path(p_original_path, true);
 	scr->reload();
 
 	if (r_error) {


modules/mono/csharp_script.cpp
```diff
@@ -2867,7 +2867,8 @@ Ref<Resource> ResourceFormatLoaderCSharpScript::load(const String &p_path, const
 	ERR_FAIL_COND_V_MSG(err != OK, Ref<Resource>(), "Cannot load C# script file '" + p_path + "'.");
 #endif
 
-	scr->set_path(p_original_path);
+	bool take_over = p_path == p_original_path;
+	scr->set_path(p_original_path, take_over);
```
YuriSizov, line 2871: GDScript scripts [seem to always take over the path when loading anew](https://github.com/godotengine/godot/blob/f8a2a9193662b2e8c1d04d65e647399dee94f31e/modules/gdscript/gdscript_cache.cpp#L256). Any particular reason why we don't want to do that here?

raulsntos: It seems when this code was added GDScript was doing the same thing in their loader:

https://github.com/godotengine/godot/blob/e36fb95c50ce0cd0ab9621afe668332895712c2e/modules/gdscript/gd_script.cpp#L1835-L1843

It was changed in https://github.com/godotengine/godot/pull/40598.

As I understand it, there should only be one instance of `CSharpScript` for each script. When a new C# script is created it's stored in a dictionary on C#'s side so it can be retrieved later by its path and avoid recreating it.

So, I'm not sure that taking over the path is necessary. We shouldn't be recreating C# scripts so if that's happening there may be an issue somewhere else that should be fixed, and changing `take_over` here would only hide that issue.

---

There's more context in https://github.com/godotengine/godot/pull/82113#discussion_r1376887952.

nongvantinh: This `take_over` fix aims to avoid the problem when a C# script already exists in the ResourceCache. Without it, the `ERR_FAIL_MSG` in `Resource::set_path` will raise.

Before this PR, I added a new method for `ResourceCache` to erase the cache and avoid the above problem. However, it seems like that is not acceptable either.

Since there is only one instance for one C# script, both the `path` and the `instance` should already be the same for all instances. I don't see why this PR would create a new instance in any way.

As far as I know, the purpose of `ResourceCache` is to cache the resource to avoid it being loaded multiple times. Since the C# script's instance is already loaded, this PR simply removes and adds the same instance to workaround the `ERR_FAIL_MSG` for `Resource::set_path`.

raulsntos: I'm not saying this PR creates a new instance, just that if there's a pre-existing bug or a bug is introduced in the future that creates multiple instances this could hide that bug.

If the script `path` is already the same then why are we setting it? Should we add an assert to make sure the path is the same? Should we avoid setting the path unless it's empty?

If I understand correctly, the path should only be set if this is a new script (i.e.: the script is being created), otherwise there should be no need to set the path, but we're always setting it anyway because if it's the same path then there's no harm in that.

I'm worried that there may be a case where the script already exists but the path is not the same for some reason, this shouldn't happen but if it did there should be an error to avoid hiding a potential issue. However, I see now that since `take_over` is only true when `p_path == p_original_path` then this shouldn't hide the error in this case, because in this case `take_over` would be false.

So I think your change makes sense, but let me know if I got any of this wrong.

nongvantinh: You're right; if the path is already the same, why would we need to set it? Why not just skip this line?

When I made this change, I just wanted to preserve the old behavior without introducing any breaking changes. However, if `set_path` is too risky and only needs to be called when a script is created, why don't we set it only once and ignore any subsequent attempts to set it. That would be clear and concise.

nongvantinh: I know the reason now. We need `take_over` instead of checking for `empty` and skipping this line because we require a mechanism to trigger an error when the `path` differs from the `original path`. This ensures that whoever changes the path knows that this operation is invalid, which is why we always call `scr->set_path`. It's not innocent; it's necessary.

YuriSizov: Both path and original path refer to the same file. The original path is just unmodified from the load request, and path can be a remapped path. But I don't think there is a case where they point to different files. They are just different representations.

So IMO we should just always take over, there should be no harm in that. And as mentioned, that's what GDScript does as well.

raulsntos: I think we should at least check to make sure the path is not changed, what I'm worried about is the current `script->get_path()` being different from the new `p_original_path` (not just different representations).

Something like this:

```c++
ERR_FAIL_COND_V(script->get_path().is_empty() || script->get_path() != p_original_path, Ref<Resource>());
```

> So IMO we should just always take over, there should be no harm in that. And as mentioned, that's what GDScript does as well.

That may work for GDScript, but I think C# must ensure that the script path matches the path it was registered with in the C# dictionary, otherwise it will create a new `CSharpScript` instance.

Unlike GDScript, C# supports defining a type across multiple files with the `partial` keyword so the same script could have multiple paths but Godot only supports one so we must ensure we always use the same one.

https://github.com/godotengine/godot/blob/f8a2a9193662b2e8c1d04d65e647399dee94f31e/modules/mono/glue/GodotSharp/GodotSharp/Core/Bridge/ScriptManagerBridge.cs#L441-L452

The path used in the dictionary comes from the `ScriptPathAttributeGenerator` and should be a `res://` path relative to the project root.

YuriSizov: > I think we should at least check to make sure the path is not changed, what I'm worried about is the current `script->get_path()` being different from the new `p_original_path` (not just different representations).
> 
> Something like this

Doing it this way also makes sense to me 👍 

> That may work for GDScript, but I think C# must ensure that the script path matches the path it was registered with in the C# dictionary, otherwise it will create a new `CSharpScript` instance.

To be clear, GDScript only does this when we can't find the script in the GDScript's own cache (unrelated to `ResourceCache`). So I would say that the general logic is pretty similar. Making sure this is not a known script/class before taking over would align C# and GDScript's behavior even more, so I'm all for it.

modules/mono/csharp_script.cpp
```diff
@@ -2867,7 +2867,11 @@ Ref<Resource> ResourceFormatLoaderCSharpScript::load(const String &p_path, const
 	ERR_FAIL_COND_V_MSG(err != OK, Ref<Resource>(), "Cannot load C# script file '" + p_path + "'.");
 #endif
 
-	scr->set_path(p_original_path);
+	// Only one instance of a C# script is allowed to exist.
+	ERR_FAIL_COND_V_MSG(scr->get_path().is_empty() || scr->get_path() != p_original_path, Ref<Resource>(),
```
raulsntos, line 2871: Sorry, I think I got the condition wrong in my previous comment, it should probably be:

```suggestion
	ERR_FAIL_COND_V_MSG(!scr->get_path().is_empty() && scr->get_path() != p_original_path, Ref<Resource>(),
```

Because I think if the path is empty it's the first time we are loading the script so it's fine to take over. We only want the error to happen when the script already has a path and it's different from the registered canonical path.

modules/mono/csharp_script.cpp
```diff
@@ -2867,7 +2867,11 @@ Ref<Resource> ResourceFormatLoaderCSharpScript::load(const String &p_path, const
...
+			"The C# script path is different from the path it was registered in the C# dictionary.");
+	bool take_over = p_path == p_original_path;
+	scr->set_path(p_original_path, take_over);
```
raulsntos, line 2874: With the added `ERR_FAIL_COND_V` we can probably always take over.

```suggestion
	scr->set_path(p_original_path, true);
```

raulsntos (approved):
The code looks good to me, and nothing seems to break in my testing. But more testing/reviews are always appreciated.

I would probably hold on cherry-picking until this has gotten enough testing in master.

akien-mga:
Thanks!

PR #74935, opened 2023-03-15
[3.x] Allow rendering of 3d content at lower resolution

Ansraer:
This makes it possible to dynamically adjust the rendering resolution of 3D content. 2D content and UI stuff are still rendered at full resolution.

The user can adjust the target resolution with the `Rendering/Quality/3D/Lower Resolution` project setting. 

The editor displaying a scene at 1/4 the resolution
image

---
This PR was sponsored by Ramatak with 💚

PR #85508, opened 2023-11-29
[3.x] Add `max_render_size` ProjectSettings

BimDav:
The size at which a game is rendered plays a huge part in the game's performance. However, before this there was no way to set a maximum render size: setting the stretch mode to `Viewport`would work to set the render size to the viewport size, but there are often cases where the viewport's internal size is large but the render size has to be smaller, e.g. a 4k game giving a high performance option for players who have a large screen size.

This adds a Project Setting "max_render_size" which is applied when the stretch mode is 2D. This will help the performance of a lot of games with high viewport size.

### DIFF
--- a/core/project_settings.cpp
+++ b/core/project_settings.cpp
@@ -1085,6 +1085,10 @@ ProjectSettings::ProjectSettings() {
 	ProjectSettings::get_singleton()->set_custom_property_info("display/window/size/test_width", PropertyInfo(Variant::INT, "display/window/size/test_width", PROPERTY_HINT_RANGE, "0,7680,1,or_greater")); // 8K resolution
 	GLOBAL_DEF("display/window/size/test_height", 0);
 	ProjectSettings::get_singleton()->set_custom_property_info("display/window/size/test_height", PropertyInfo(Variant::INT, "display/window/size/test_height", PROPERTY_HINT_RANGE, "0,4320,1,or_greater")); // 8K resolution
+	GLOBAL_DEF("display/window/size/max_render_width", 0);
+	ProjectSettings::get_singleton()->set_custom_property_info("display/window/size/max_render_width", PropertyInfo(Variant::INT, "display/window/size/max_render_width", PROPERTY_HINT_RANGE, "0,7680,1,or_greater")); // 8K resolution
+	GLOBAL_DEF("display/window/size/max_render_height", 0);
+	ProjectSettings::get_singleton()->set_custom_property_info("display/window/size/max_render_height", PropertyInfo(Variant::INT, "display/window/size/max_render_height", PROPERTY_HINT_RANGE, "0,4320,1,or_greater")); // 8K resolution
 
 	GLOBAL_DEF("audio/default_bus_layout", "res://default_bus_layout.tres");
 	custom_prop_info["audio/default_bus_layout"] = PropertyInfo(Variant::STRING, "audio/default_bus_layout", PROPERTY_HINT_FILE, "*.tres");
--- a/doc/classes/ProjectSettings.xml
+++ b/doc/classes/ProjectSettings.xml
@@ -525,6 +525,12 @@
 		<member name="display/window/size/height" type="int" setter="" getter="" default="600">
 			Sets the game's main viewport height. On desktop platforms, this is the default window size. Stretch mode settings also use this as a reference when enabled.
 		</member>
+		<member name="display/window/size/max_render_height" type="int" setter="" getter="" default="0">
+			Sets the maximum height of the game's main viewport, when the stretch mode is "2D". If set to 0, the maximum height is the screen height.
+		</member>
+		<member name="display/window/size/max_render_width" type="int" setter="" getter="" default="0">
+			Sets the maximum width of the game's main viewport, when the stretch mode is "2D". If set to 0, the maximum width is the screen width.
+		</member>
 		<member name="display/window/size/resizable" type="bool" setter="" getter="" default="true">
 			Allows the window to be resizable by default.
 			[b]Note:[/b] This setting is ignored on iOS.
--- a/doc/classes/SceneTree.xml
+++ b/doc/classes/SceneTree.xml
@@ -212,6 +212,13 @@
 				Marks the most recent [InputEvent] as handled.
 			</description>
 		</method>
+		<method name="set_max_render_size">
+			<return type="void" />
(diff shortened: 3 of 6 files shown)


core/project_settings.cpp
```diff
@@ -1085,6 +1085,10 @@ ProjectSettings::ProjectSettings() {
 	ProjectSettings::get_singleton()->set_custom_property_info("display/window/size/test_width", PropertyInfo(Variant::INT, "display/window/size/test_width", PROPERTY_HINT_RANGE, "0,7680,1,or_greater")); // 8K resolution
 	GLOBAL_DEF("display/window/size/test_height", 0);
 	ProjectSettings::get_singleton()->set_custom_property_info("display/window/size/test_height", PropertyInfo(Variant::INT, "display/window/size/test_height", PROPERTY_HINT_RANGE, "0,4320,1,or_greater")); // 8K resolution
+	GLOBAL_DEF("display/window/size/max_render_width", 0);
+	ProjectSettings::get_singleton()->set_custom_property_info("display/window/size/width", PropertyInfo(Variant::INT, "display/window/size/width", PROPERTY_HINT_RANGE, "0,7680,1,or_greater")); // 8K resolution
+	GLOBAL_DEF("display/window/size/max_render_height", 0);
+	ProjectSettings::get_singleton()->set_custom_property_info("display/window/size/height", PropertyInfo(Variant::INT, "display/window/size/height", PROPERTY_HINT_RANGE, "0,4320,1,or_greater")); // 8K resolution
```
AThousandShips, line 1091:
```suggestion
	GLOBAL_DEF("display/window/size/max_render_width", 0);
	ProjectSettings::get_singleton()->set_custom_property_info("display/window/size/max_render_width", PropertyInfo(Variant::INT, "display/window/size/max_render_width", PROPERTY_HINT_RANGE, "0,7680,1,or_greater")); // 8K resolution
	GLOBAL_DEF("display/window/size/max_render_height", 0);
	ProjectSettings::get_singleton()->set_custom_property_info("display/window/size/max_render_height", PropertyInfo(Variant::INT, "display/window/size/max_render_height", PROPERTY_HINT_RANGE, "0,4320,1,or_greater")); // 8K resolution
```

doc/classes/ProjectSettings.xml
```diff
@@ -525,6 +525,12 @@
 		<member name="display/window/size/height" type="int" setter="" getter="" default="600">
 			Sets the game's main viewport height. On desktop platforms, this is the default window size. Stretch mode settings also use this as a reference when enabled.
 		</member>
+		<member name="display/window/size/max_render_height" type="int" setter="" getter="" default="0">
+			Sets the maximum height at which the game is rendered, when the stretch mode is "2D". If set to 0, the render height is the screen height.
```
AThousandShips, line 529:
```suggestion
			Sets the maximum height of the game's main viewport, when the stretch mode is "2D". If set to 0, the maximum height is the screen height.
```

doc/classes/ProjectSettings.xml
```diff
@@ -525,6 +525,12 @@
...
+		</member>
+		<member name="display/window/size/max_render_width" type="int" setter="" getter="" default="0">
+			Sets the maximum width at which the game is rendered, when the stretch mode is "2D". If set to (0, 0), the render width is the screen width.
```
AThousandShips, line 532:
```suggestion
			Sets the maximum width of the game's main viewport, when the stretch mode is "2D". If set to 0, the maximum width is the screen width.
```

BimDav:
> Is this change not relevant for 4.x?

It is, and I will make the PR for Godot 4. I had to make the fix for my game which is in production and made in Godot 3, so I did it in 3 first.

Calinou:
I'm not sure about the way this is implemented for two reasons:

- If no bilinear filtering is applied, then the result will feature uneven pixel scaling
- It encourages hardcoding a maximum resolution, which is bad for future-proofing games. I don't want to be prevented from rendering a game internally at 8K if I want to :slightly_smiling_face: 

Instead, a resolution scale multiplier should be preferred. 4.x already exposes this with the **Scaling 3D** project settings, which only affects 3D rendering without affecting 2D. This already has a 3.x implementation in https://github.com/godotengine/godot/pull/74935.

BimDav:
>     * If no bilinear filtering is applied (nearest-neighbor is the default when using the `viewport` stretch mode), then the result will feature uneven pixel scaling unless `max_render_size` is exactly half the window size (or any other integer divisor).

The maximum render size is only applied in the `2d` stretch mode, so I'm not sure this is relevant? I may have missed your point.

>     * It encourages hardcoding a maximum resolution, which is bad for future-proofing games. I've seen too many old games have hard caps on the resolution they can internally render at, [and it still happens today sometimes](https://github.com/Lyall/MGSHDFix). I don't want to be prevented from rendering a game internally at 8K if I want to[1](#user-content-fn-1-e2de57130e6216d951f63ca554014a65). It may not be viable now, but it may be viable in 5 years from now 🙂

This is true. We could make it only accessible from code, and not from the project settings so that it is clearer it is intended for quality/performance options. But it could make it too hidden to be found.

I don't really get how a scale multiplier changes your second point? I am open to a scale multiplier instead of a fixed resolution, but if the intention is to limit the render resolution below the screen resolution (it is at least my intention), then it will limit the resolution. The game dev will have to add a "high quality" option that sets max_render_size to (0, 0) in order to be future proof. That is how I used it in my game.

Calinou:
> The maximum render size is only applied in the 2d stretch mode, so I'm not sure this is relevant? I may have missed your point.

My point is more generally that by default, Godot uses nearest-neighbor filtering when the viewport size doesn't match the window size. I believe this issue trickles down to your implementation as well, but I haven't tested it locally.

> I don't really get how a scale multiplier brings a solution to your second point? I am open to a scale multiplier instead of a fixed resolution, but if the intention is to limit the render resolution below the screen resolution (it is at least my intention), then it will limit the resolution. The game dev will have to add a "high quality" option that sets max_render_size to (0, 0) in order to be future proof. That is how I used it in my game.

There's also the concern that low-end devices may need an even lower resolution to run smoothly either way. A multiplier handles this automatically, assuming low-end mobile devices have 720p displays, mid-range devices have 1080p displays and flagships have 1440p displays. On the contrary, if you set a fixed maximum resolution of 720p, the low-end devices will keep rendering at native resolution which is probably too much for them (480p would be more suited).

PR #85513, opened 2023-11-29
Save external data even without scene

KoBeWi:
When using Ctrl+S without scene opened, it saves all modified resources. However it doesn't save external data of editor plugins that don't modify Resources.

This PR ensures that unsaved external data (based on `get_unsaved_status()`) is still saved even when there is no scene. This is mostly relevant for user plugins, as none of Godot's plugins modify non-Resource data AFAIK.

Probably a better solution would be modifying EditorData's `save_editor_external_data()` to work like `_save_external_resources()`, i.e. only save modified external data and return how many plugins were saved, but `get_unsaved_status()` is a recent addition and almost no plugin implements it.

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -1693,7 +1693,7 @@ bool EditorNode::_validate_scene_recursive(const String &p_filename, Node *p_nod
 	return false;
 }
 
-int EditorNode::_save_external_resources() {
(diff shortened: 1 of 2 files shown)


akien-mga (approved):
Let's give it a try :)

akien-mga:
Thanks!

PR #85515, opened 2023-11-29
Expose `Node.is_part_of_edited_scene()`

### LINKED ISSUE #8216: Add `Node.is_being_edited()` as an equivalent of `Engine.is_editor_hint()` for editor plugins

A bespoke editor plugin for my game.

Plugin code needs to be marked with `@tool` in order to run in the editor.

However, this also causes the code to be run while editing a scene belonging to the plugin. This is often undesirable, because it makes changes to the scene which then get saved in the `.tscn` file.

For non-plugin scripts, `Engine.is_editor_hint()` is used to work around this. But in a plugin script, this function always returns true, even while the plugin is running "as a plugin" and is not being edited.

Since the node will be part of the same engine instance and scene tree, whether it's running as a plugin or running in the scene editor, a global function like `Engine.is_editor_hint()` won't be possible.

Instead, maybe a function on `Node` could be added, like `is_being_edited()`.

`is_being_edited()` could look at the node's ancestors to find out whether the node is a descendant of the scene editor. I think that would be the `CanvasItemEditor` class in the current implementation, but this is not exposed in the API.

Yes, I'm currently using something along these lines as a workaround, but it only works for main screen plugins:

```cs
    public static bool IsRunningAsPlugin(Node node)
    {
        foreach (var mainScreenChild in EditorInterface.Singleton.GetEditorMainScreen().GetChildren())
        {
            if (mainScreenChild is MyPluginMainPanel && mainScreenChild.IsAncestorOf(node))
            {
                return true;
            }
        }
        return false;
    }
```

Seems like many plugins that use scenes can benefit from this.

KoBeWi:
This PR does 3 things:
- Remove `SceneTree::is_node_being_edited()`. It's function is the same as the newly added `is_part_of_edited_scene()`, but it requires a check whether node is inside scene tree
- `is_part_of_edited_scene()` is now available outside editor, where it will always return `false`
- Exposed `is_part_of_edited_scene()` to scripting

Closes https://github.com/godotengine/godot-proposals/issues/8216
Closes (?) https://github.com/godotengine/godot-proposals/issues/900

### DIFF
--- a/doc/classes/Node.xml
+++ b/doc/classes/Node.xml
@@ -619,6 +619,12 @@
 				[method request_ready] resets it back to [code]false[/code].
 			</description>
 		</method>
+		<method name="is_part_of_edited_scene" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if the node is part of the scene currently opened in the editor.
+			</description>
+		</method>
 		<method name="is_physics_interpolated" qualifiers="const">
 			<return type="bool" />
 			<description>
--- a/scene/2d/audio_listener_2d.cpp
+++ b/scene/2d/audio_listener_2d.cpp
@@ -45,7 +45,7 @@ bool AudioListener2D::_set(const StringName &p_name, const Variant &p_value) {
 
 bool AudioListener2D::_get(const StringName &p_name, Variant &r_ret) const {
 	if (p_name == "current") {
-		if (is_inside_tree() && get_tree()->is_node_being_edited(this)) {
+		if (is_part_of_edited_scene()) {
 			r_ret = current;
 		} else {
 			r_ret = is_current();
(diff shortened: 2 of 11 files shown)


doc/classes/Node.xml
```diff
@@ -556,6 +556,12 @@
 				[method request_ready] resets it back to [code]false[/code].
 			</description>
 		</method>
+		<method name="is_part_of_edited_scene" qualifiers="const">
```
Mickeon, line 559: Why `is_part_of_edited_scene` and not `is_inside_edited_scene`? I mean, I kind of get why, but it does not quite roll off the tongue.

KoBeWi: idk, IMO this name is fine.
Also the method was changed a bit to include siblings of the scene, so `is_inside_edited_scene()` would be less accurate.

akien-mga (approved):
Looks good.

akien-mga:
Thanks!

PR #83934, opened 2023-10-25
Fix animation track paths updated by scene dock

KoBeWi:
Fixes #83908
There is still some error when moving AnimationPlayer, but looks harmless and unrelated.

PR #83263, opened 2023-10-13
Don't try updating wrong NodePaths in resources

KoBeWi:
Fixes #83209

PR #80721, opened 2023-08-17
Recurse into resources to check for changed node paths

tom95:
Fixes #74155 by recursing into resources when checking for changed `NodePath`s. Also verified to work with custom resources.

The added code follows the same pattern as [the code that handles `Node` subclasses](https://github.com/godotengine/godot/blob/0511f9d9a7d56c742d87fafdcea8785d40ad14b3/editor/scene_tree_dock.cpp#L1716). For `Node` subclasses, however, the existing code does some extra work we don't need for resources, such as recursing into the Node's children and scanning for related AnimationPlayers.

PR #85517, opened 2023-11-29
Disable a prohibitively slow code branch when reparenting nodes

YuriSizov:
Addresses (but doesn't fix) https://github.com/godotengine/godot/issues/84910 by temporarily disabling enhancements added in https://github.com/godotengine/godot/pull/80721 and its two follow-ups (https://github.com/godotengine/godot/pull/83934, https://github.com/godotengine/godot/pull/83263).

This temporarily reopens https://github.com/godotengine/godot/issues/74155.

----

We still want this functionality, but some improvements to the approach are necessary. We discussed https://github.com/godotengine/godot/pull/85502 as well, but it only addresses the issue by ignoring a handful of potentially valid cases. So for the time being we have to disable the feature, until a better solution is available.

### DIFF
--- a/editor/scene_tree_dock.cpp
+++ b/editor/scene_tree_dock.cpp
@@ -1762,6 +1762,8 @@ bool SceneTreeDock::_check_node_path_recursive(Node *p_root_node, Variant &r_var
 			}
 		} break;
 
+// FIXME: This approach causes a significant performance regression, see GH-84910.
+#if 0
 		case Variant::OBJECT: {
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

KoBeWi (approved):

akien-mga:
Thanks!

atirut-w:
I assume this will be reverted once the fixes in #85502 gets finished up?

akien-mga:
Yes, or any other follow up that improves performance with introducing new bugs.

PR #85519, opened 2023-11-29
Added proper timeout for blender rpc connection

### LINKED ISSUE #85446: Blender imports not working in blender 4.1 (with or without RPC)
### Godot version

4.2.rc.custom_build.eda44bfe1

### System information

Linux 6.6.1-arch1-1 Vulkan API 1.3.255 - Forward+

### Issue description

Importing .blend files not working when using blender version 4.1
When using blender version Blender 4.0.1 it is working, but without RPC (so it's very slow)

### Steps to reproduce

Create a simple .blend file (default cube scene will do)
Enable blender import for the project
Set blender port and path in editor settings

N/A

mxaddict:
Ran into a bug where blender rpc server would not start fast enough and cause Godot editor to think that it was not running, more details here: in #85446

This simply adds a 1 second delay between starting the rpc server and initiating the connection via HTTP.

I thought about checking the STDOUT of the rpc command, but `OS::create_process` does not have a way to read the STDOUT that I can see in the codebase.

_Production edit: Closes https://github.com/godotengine/godot/issues/85446_

### DIFF
--- a/modules/gltf/editor/editor_import_blend_runner.cpp
+++ b/modules/gltf/editor/editor_import_blend_runner.cpp
@@ -198,6 +198,40 @@ Error EditorImportBlendRunner::do_import(const Dictionary &p_options) {
 	}
 }
 
+HTTPClient::Status EditorImportBlendRunner::connect_blender_rpc(const Ref<HTTPClient> &p_client, int p_timeout_usecs) {
+	p_client->connect_to_host("127.0.0.1", rpc_port);
+	HTTPClient::Status status = p_client->get_status();
+
+	int attempts = 1;
+	int wait_usecs = 1000;
+
+	bool done = false;
+	while (!done) {
+		OS::get_singleton()->delay_usec(wait_usecs);
+		status = p_client->get_status();
+		switch (status) {
+			case HTTPClient::STATUS_RESOLVING:
+			case HTTPClient::STATUS_CONNECTING: {
+				p_client->poll();
+				break;
+			}
+			case HTTPClient::STATUS_CONNECTED: {
+				done = true;
+				break;
+			}
+			default: {
+				if (attempts * wait_usecs < p_timeout_usecs) {
+					p_client->connect_to_host("127.0.0.1", rpc_port);
+				} else {
+					return status;
+				}
+			}
+		}
+	}
+
+	return status;
+}
+
 Error EditorImportBlendRunner::do_import_rpc(const Dictionary &p_options) {
 	kill_timer->stop();
 
@@ -217,25 +251,9 @@ Error EditorImportBlendRunner::do_import_rpc(const Dictionary &p_options) {
 
 	// Connect to RPC server.
 	Ref<HTTPClient> client = HTTPClient::create();
-	client->connect_to_host("127.0.0.1", rpc_port);
-
-	bool done = false;
-	while (!done) {
-		HTTPClient::Status status = client->get_status();
-		switch (status) {
-			case HTTPClient::STATUS_RESOLVING:
-			case HTTPClient::STATUS_CONNECTING: {
-				client->poll();
-				break;
-			}
-			case HTTPClient::STATUS_CONNECTED: {
-				done = true;
-				break;
-			}
-			default: {
-				ERR_FAIL_V_MSG(ERR_CONNECTION_ERROR, vformat("Unexpected status during RPC connection: %d", status));
-			}
-		}
+	HTTPClient::Status status = connect_blender_rpc(client, 1000000);
+	if (status != HTTPClient::STATUS_CONNECTED) {
+		ERR_FAIL_V_MSG(ERR_CONNECTION_ERROR, vformat("Unexpected status during RPC connection: %d", status));
 	}
 
 	// Send XML request.
@@ -246,9 +264,9 @@ Error EditorImportBlendRunner::do_import_rpc(const Dictionary &p_options) {
 	}
 
 	// Wait for response.
-	done = false;
+	bool done = false;
 	while (!done) {
-		HTTPClient::Status status = client->get_status();
+		status = client->get_status();
 		switch (status) {
 			case HTTPClient::STATUS_REQUESTING: {
 				client->poll();
--- a/modules/gltf/editor/editor_import_blend_runner.h
+++ b/modules/gltf/editor/editor_import_blend_runner.h
@@ -33,6 +33,7 @@
 
 #ifdef TOOLS_ENABLED
 
+#include "core/io/http_client.h"
 #include "core/os/os.h"
 #include "scene/main/node.h"
 #include "scene/main/timer.h"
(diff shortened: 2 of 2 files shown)


RedMser:
Alternative approach which might be less hacky: in the switch case that checks for the connection status, add a new case for `STATUS_CANT_CONNECT` (which is equivalent to the `4` that got printed in the issue you linked), delay for ~100ms in there and count a local "retry" counter up by 1. If it hits 10, error as before. This can make the delay shorter, as it'll poll for the RPC status more frequently while starting.

mxaddict:
> Alternative approach which might be less hacky: in the switch case that checks for the connection status, add a new case for `STATUS_CANT_CONNECT` (which is equivalent to the `4` that got printed in the issue you linked), delay for ~100ms in there and count a local "retry" counter up by 1. If it hits 10, error as before. This can make the delay shorter, as it'll poll for the RPC status more frequently while starting.

I thought about this as well, but got a little side tracked looking for a sleep function 🤣 

I can update the PR to apply that kind of logic, the `STATUS_CANT_CONNECT` branch in the switch statement will have to rerun the `client->connect_to_host("127.0.0.1", rpc_port);` code a presume.

RedMser:
> the `STATUS_CANT_CONNECT` branch in the switch statement will have to rerun the `client->connect_to_host("127.0.0.1", rpc_port);` code I presume.

Yeah I believe so as well.

mxaddict:
> > the `STATUS_CANT_CONNECT` branch in the switch statement will have to rerun the `client->connect_to_host("127.0.0.1", rpc_port);` code I presume.
> 
> Yeah I believe so as well.

Since it will be waiting for max ~100ms if the RPC server start immediately, might make sense to increase the max wait time to like ~5s or ~10 seconds incase it's super slow on older machines?

It should still be faster on newer machines since it will break out of the loop when connected, but allow for slower devices to wait longer for the RPC server to spool up?

fire:
Sometimes other implementation use exponential backoff. Do you think that's required here?

fire:
Like the wikipedia page https://en.wikipedia.org/wiki/Exponential_backoff Truncated exponential backoff

mxaddict:
> Like the wikipedia page https://en.wikipedia.org/wiki/Exponential_backoff Truncated exponential backoff

I've pushed a change that allows this to return as early as 1000us instead of waiting for a full 1 second.

Latest push should also be rebased on master branch

mxaddict:
Got busy with other stuff, sorry took me a while to get this done 🤣

modules/gltf/editor/editor_import_blend_runner.h
```diff
@@ -60,6 +61,7 @@ class EditorImportBlendRunner : public Node {
 	bool is_running() { return blender_pid != 0 && OS::get_singleton()->is_process_running(blender_pid); }
 	bool is_using_rpc() { return rpc_port != 0; }
 	Error do_import(const Dictionary &p_options);
+	HTTPClient::Status connect_blender_rpc(Ref<HTTPClient>, const int timeout_usecs);
```
AThousandShips, line 64:
```suggestion
	HTTPClient::Status connect_blender_rpc(Ref<HTTPClient> p_client, int p_timeout_usecs);
```
Match arguments, and use `p_`, also no `const` for simple types

modules/gltf/editor/editor_import_blend_runner.cpp
```diff
@@ -198,30 +198,19 @@ Error EditorImportBlendRunner::do_import(const Dictionary &p_options) {
...
-	// Start Blender if not already running.
-	if (!is_running()) {
-		// Start an XML RPC server on the given port.
-		String python = vformat(PYTHON_SCRIPT_RPC, rpc_port);
-		Error err = start_blender(python, false);
-		if (err != OK || blender_pid == 0) {
-			return FAILED;
-		}
-	}
+HTTPClient::Status EditorImportBlendRunner::connect_blender_rpc(Ref<HTTPClient> client, const int timeout_usecs) {
```
AThousandShips, line 202:
```suggestion
HTTPClient::Status EditorImportBlendRunner::connect_blender_rpc(Ref<HTTPClient> p_client, int p_timeout_usecs) {
```

mxaddict:
Thanks @AThousandShips, I missed those :)

mxaddict:
Nice, build passed finally! :)

mxaddict:
Anything else that I need to add to this PR? Happy holidays btw! 😄

mxaddict:
This PR has been open for a while now, not sure what the next step is.

modules/gltf/editor/editor_import_blend_runner.h
```diff
@@ -60,6 +61,7 @@ class EditorImportBlendRunner : public Node {
 	bool is_running() { return blender_pid != 0 && OS::get_singleton()->is_process_running(blender_pid); }
 	bool is_using_rpc() { return rpc_port != 0; }
 	Error do_import(const Dictionary &p_options);
+	HTTPClient::Status connect_blender_rpc(Ref<HTTPClient> p_client, int p_timeout_usecs);
```
akien-mga, line 64:
```suggestion
	HTTPClient::Status connect_blender_rpc(const Ref<HTTPClient> &p_client, int p_timeout_usecs);
```

modules/gltf/editor/editor_import_blend_runner.cpp
```diff
@@ -198,6 +198,40 @@ Error EditorImportBlendRunner::do_import(const Dictionary &p_options) {
 	}
 }
 
+HTTPClient::Status EditorImportBlendRunner::connect_blender_rpc(Ref<HTTPClient> p_client, int p_timeout_usecs) {
```
akien-mga, line 201:
```suggestion
HTTPClient::Status EditorImportBlendRunner::connect_blender_rpc(const Ref<HTTPClient> &p_client, int p_timeout_usecs) {
```

akien-mga (commented):
Looks good to me overall.

mxaddict:
> Looks good to me overall.
> 
> Could you squash the commits? See [PR workflow](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#the-interactive-rebase) for instructions.

Sure, I'll squash the commits

mxaddict:
I've pushed the squashed commit

mxaddict:
Sorry, I missed the commit for the `const` changes, pushing now.

modules/gltf/editor/editor_import_blend_runner.cpp
```diff
@@ -198,6 +198,40 @@ Error EditorImportBlendRunner::do_import(const Dictionary &p_options) {
 	}
 }
 
+HTTPClient::Status EditorImportBlendRunner::connect_blender_rpc(const Ref<HTTPClient> p_client, int p_timeout_usecs) {
```
akien-mga, line 201: It should be with a `&`:
```suggestion
HTTPClient::Status EditorImportBlendRunner::connect_blender_rpc(const Ref<HTTPClient> &p_client, int p_timeout_usecs) {
```
Same in the header.

mxaddict:
Changed to `&p_client`

akien-mga (approved):

mxaddict:
I can't see the failure on the test.

Screenshot 2024-01-27 at 1 36 51 PM

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #85526, opened 2023-11-30
Add unit tests for `SortArray` + do some conservative refactoring

ershn:
I'm working towards tackling https://github.com/godotengine/godot-proposals/issues/8496 but for that I need a heap implementation. There is one in `SortArray` but right now it's looks more like an implementation detail of the sorting algorithm than a generic heap implementation. This PR makes it more clear what you can and cannot use in that class by doing some conservative refactoring, adding documentation, and adding unit tests.

Details on the refactoring:
- only make public what can easily be used by consumers of `SortArray` (make implementation details private)
- remove unnecessary `inline` keywords since member functions defined inside a class declaration are implicitly inline
- add a `push_heap` overload that can more easily be used by consumers of `SortArray`

Apart from the new overload the implementation of `SortArray` itself is untouched.

### DIFF
--- a/core/templates/sort_array.h
+++ b/core/templates/sort_array.h
@@ -60,7 +60,49 @@ class SortArray {
 public:
 	Comparator compare;
 
-	inline const T &median_of_3(const T &a, const T &b, const T &c) const {
+	void push_heap(int p_first, int p_last, T *p_array) const {
+		_push_heap(p_first, p_last - 1, p_first, p_array[p_last - 1], p_array);
+	}
+
+	void pop_heap(int p_first, int p_last, T *p_array) const {
+		_pop_heap(p_first, p_last - 1, p_last - 1, p_array[p_last - 1], p_array);
+	}
+
+	/**
+	 * Sort using the introsort algorithm (in-place, not stable).
+	 */
+	void sort_range(int p_first, int p_last, T *p_array) const {
+		if (p_first != p_last) {
+			_introsort(p_first, p_last, p_array, _bitlog(p_last - p_first) * 2);
+			_final_insertion_sort(p_first, p_last, p_array);
+		}
+	}
+
+	void sort(T *p_array, int p_len) const {
+		sort_range(0, p_len, p_array);
+	}
+
+	/**
+	 * Partially sort the array so that the element at the `p_nth` index has the value it would have
+	 * if the array were sorted.
+	 */
+	void nth_element(int p_first, int p_last, int p_nth, T *p_array) const {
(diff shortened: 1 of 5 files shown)


core/templates/sort_array.h
```diff
@@ -60,7 +60,78 @@ class SortArray {
 public:
 	Comparator compare;
 
-	inline const T &median_of_3(const T &a, const T &b, const T &c) const {
+	/**
```
AThousandShips, line 63: We don't normally add doc comments like this, so I'd say to remove them, especially without having them generally

tests/core/templates/test_sort_array.h
```diff
@@ -0,0 +1,239 @@
...
+#include "core/templates/sort_array.h"
+#include "core/templates/vector.h"
+
+#include "tests/test_macros.h"
+
+namespace TestSortArray {
+
+template <class T>
+struct GreaterThan {
+	bool operator()(const T &a, const T &b) const { return a > b; }
```
AThousandShips, line 43:
```suggestion
	bool operator()(const T &p_a, const T &p_b) const { return p_a > p_b; }
```
Prefix with `p_`

tests/core/templates/test_sort_array.h
```diff
@@ -0,0 +1,239 @@
...
+		Vector<int> vector = { 5, 6 };
+		SortArray<int> sorter;
+
+		sorter.nth_element(0, 0, 0, vector.ptrw());
+		CHECK(vector == Vector{ 5, 6 });
+	}
+}
+
+template <class T, class Comp = _DefaultComparator<T>>
+static Vector<T> push_pop_vector(const Vector<T> &elements) {
```
AThousandShips, line 181:
```suggestion
static Vector<T> push_pop_vector(const Vector<T> &p_elements) {
```

core/templates/sort_array.h
ershn: Do you mean I should use `//` instead of `/**` or that I should remove the comments altogether ?

If it's the later I don't mind but IMO `nth_element` at least needs some kind of explanation because its usage is really not obvious (unless you know STL has a similarly named function).
I think it would also be beneficial to keep the explanation about the sorting algorithm not being stable.

AThousandShips: Remove them entirely, you can put some details there but not a full doc comment explaining it in detail

ershn: Ok, will do.

ershn: Fixed all the review comments.
As for the method docs, I only left what I thought was strictly necessary.

PR #85532, opened 2023-11-30
Apply some low-hanging fruit optimizations to Vulkan RD

RandomShaper:
Multiple commits:

1. **Remove redundant explicit clears in the Vulkan RD**

`VK_ATTACHMENT_LOAD_OP_CLEAR`, used in the Vulkan renderer, already specifies that the render target (full or the specified region) will be cleared by the driver. There's no need for adding extra commands for clearing.

**UPDATE:** It turned out only a subset of those cases could be avoided; namely, only if both color and depth are to be rendered to a region that will be cleared. So, the improvement is not as profound as initially expected, but it can still save a few manual clears in cases like shadow mapping.

2. **Remove superfluous locking in RID owners in Vulkan RD**

The `RenderingDevice` API is fully thread-safe, in the lock-per-function fashion, so there's no need for the RID owners to be thread-safe. This change avoids a lot of spinlocks.

---

This PR would ideally be merged soon in 4.3, since it will not be very long-lived there, as the new `RenderingDeviceDriver` model will **UPDATE:** already contain the fixes ~change it all (spoiler: it will restore the explicit clearing of buffers, but will only do as an assistance to driver implementations, if the driver is flagged as needing the manual clear)~.

The most important thing of this PR is its cherry-picking into 4.2 (and older?), as it can't but improve performance.


akien-mga:
4.0 is practically EOL, I would not cherry-pick enhancements there, only absolutely critical bugfixes.

RandomShaper:
@sakrel, thank you very much for bug reporting. It turns out there were fewer cases that could be optimized reliably. I've updated the code and the PR description.

BastiaanOlij (approved):
Cool, I wanted the constrain to region in there for awhile, especially the way we update shadow maps on mobile it should make a difference once we finish off the cascaded split logic.

BastiaanOlij:
> This patch appears to fix the issue, but the code mentions to "always fully operate on the whole framebuffer", so I'm not sure if this patch breaks anything else or kills perf:

I don't know about 2D, but in 3D there are definitely assumptions in the way shadows are updated that the region is cleared as a whole, but seeing the new functionality here is governed by `p_constrained_to_region` most places should still use the old approach.

From what I understood from Reduz, this behaves better on desktop GPUs, from what I've understood about mobile, constraining to regions works better **especially** if we ensure that we respect tile boundaries in our atlasses.

So I wager we need to do follow up PRs to make the most out of these changes. I have a bunch of old shadow PRs that really need to be revisited and polished up.

YuriSizov:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

YuriSizov:
Cherry-picked for 4.1.4.

PR #85544, opened 2023-11-30
Add show_error param to script reload

samsface:
A pr to resolve the issues in this proposal: https://github.com/godotengine/godot-proposals/issues/8548.

We're adding an extra optional param to the `GDScript.reload()` function. 
Now we can pass (like the Expression class) a param `show_error:bool`.

This is useful for parsing scripts in game or in editor tools were we don't want the error output in the console or the debugger to trigger if a script fails to parse.

### DIFF
--- a/core/object/script_language.cpp
+++ b/core/object/script_language.cpp
@@ -134,7 +134,7 @@ void Script::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("has_source_code"), &Script::has_source_code);
 	ClassDB::bind_method(D_METHOD("get_source_code"), &Script::get_source_code);
 	ClassDB::bind_method(D_METHOD("set_source_code", "source"), &Script::set_source_code);
-	ClassDB::bind_method(D_METHOD("reload", "keep_state"), &Script::reload, DEFVAL(false));
+	ClassDB::bind_method(D_METHOD("reload", "keep_state", "show_error"), &Script::reload, DEFVAL(false), DEFVAL(true));
 	ClassDB::bind_method(D_METHOD("get_base_script"), &Script::get_base_script);
 	ClassDB::bind_method(D_METHOD("get_instance_base_type"), &Script::get_instance_base_type);
 
--- a/core/object/script_language.h
+++ b/core/object/script_language.h
@@ -138,7 +138,7 @@ public:
 	virtual bool has_source_code() const = 0;
 	virtual String get_source_code() const = 0;
 	virtual void set_source_code(const String &p_code) = 0;
-	virtual Error reload(bool p_keep_state = false) = 0;
+	virtual Error reload(bool p_keep_state = false, bool p_show_error = true) = 0;
 
 #ifdef TOOLS_ENABLED
 	virtual Vector<DocData::ClassDoc> get_documentation() const = 0;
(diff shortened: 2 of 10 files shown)


passivestar:
Hey, seems like I still get the error:

image

Screenshot 2023-11-30 at 20 04 24

```gdscript
static func execute_expression_as_script(expression_string : String) -> void:
	var script := GDScript.new()

	var template = """
@tool
extends EditorScript

static func run():
	_RETURN__CODE_
"""

	var code = template.replace("_CODE_", expression_string)
	var code_with_return = code.replace("_RETURN_", "return ")
	var code_without_return = code.replace("_RETURN_", "")

	# Try with return first
	script.source_code = code_with_return
	var error := script.reload(false, false)
	if error != OK:
		print("Parsing with return failed")
		# If it fails, try without return
		script.source_code = code_without_return
		var error_no_return := script.reload(false, false)
		if error_no_return != OK:
			print("Parsing failed")
			return
	
	var result = script.run()
	print(result)
```

passivestar:
from the error message looks like its at the compilation step from here:
https://github.com/godotengine/godot/pull/85544/files#diff-2d47041c61b0e94d891650482b0cf9398f22729abdc811f94c2b7da902ed392fR780

samsface:
> from the error message looks like its at the compilation step from here: https://github.com/godotengine/godot/pull/85544/files#diff-2d47041c61b0e94d891650482b0cf9398f22729abdc811f94c2b7da902ed392fR780

Ah. I was only checking when parsing. Will wrap the analyzer and compiler parts too.

passivestar:
Checked the updated build, works as expected in my case 🎉

@samsface Thanks 🙏

samsface:
> Checked the updated build, works as expected in my case 🎉
> 
> @samsface Thanks 🙏

Btw, I discovered there's a property on the Engine class called  "Engine.print_error_messages = false". This stops the red text being printed out.
I think combined with a flag to suppress the debugger in the same way might be a better solution.

passivestar:
> I think combined with a flag to suppress the debugger

You mean something like this ?
```gdscript
Engine.print_error_messages = false
Engine.debugger_stop_on_errors = false
script.run()
Engine.print_error_messages = true
Engine.debugger_stop_on_errors = true
```

This would be nice to have too, yeah

samsface:
> > I think combined with a flag to suppress the debugger
> 
> You mean something like this ?
> 
> ```gdscript
> Engine.print_error_messages = false
> Engine.debugger_stop_on_errors = false
> script.run()
> Engine.print_error_messages = true
> Engine.debugger_stop_on_errors = true
> ```
> 
> This would be nice to have too, yeah
> 
> Edit: although then you'd need an api to get the text of error messages from those scripts, like in Expressions. Because you still want to know if something went wrong, you just don't want the debugger to deal with it

Exactly that. I think the deep issue is the debugger breaking should be configurable by code somehow.

Yeah I think this PR needs an ability to get last error message too if goes forward.

CsloudX:
Need this feature, can be merged 4.4?

AThousandShips:
@CsloudX Please don't bump without contributing significant new information. Use the :+1: reaction button on the first post instead.

AThousandShips:
For CI to pass you will need to register compatibility methods, see [here](https://docs.godotengine.org/en/latest/contributing/development/handling_compatibility_breakages.html)

CsloudX:
@Calinou Sorry for this and thank you response.
for me, i need 2 features:
1. don't break in editor.
2. can get is ok for custom gds.

PR #85554, opened 2023-11-30
Issue template: Change version field to "Tested versions", asking to test earlier versions

akien-mga:
We very often end up asking users to test different versions to pinpoint if it's a regression, or need to test ourselves. Let's ask explicitly upfront.

Here's how it looks like:

*Edit:* Older version moved at the end in *Details*.

image

---

Some things to discuss:
- Is "Reproducibility" the best name for this?
- Should it be mandatory to fill? If so, should we suggest what to write if users can't test other versions for any reason?
- Where should it be located? It's related to the "Godot version" field, but I felt that putting it up there is jarring and will be a problem for users already familiar with our template.

<details>

First version:

image

</details>

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -5,19 +5,25 @@ body:
 - type: markdown
   attributes:
     value: |
-      - When reporting bugs, you'll make our life simpler (and the fix will come sooner) if you follow the guidelines in this template.
+      When reporting bugs, please follow the guidelines in this template. This helps identify the problem precisely and thus enables contributors to fix it faster.
       - Write a descriptive issue title above.
       - The golden rule is to **always open *one* issue for *one* bug**. If you notice several bugs and want to report them, make sure to create one new issue for each of them.
       - Search [open](https://github.com/godotengine/godot/issues) and [closed](https://github.com/godotengine/godot/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported. If you don't find a relevant match or if you're unsure, don't hesitate to **open a new issue**. The bugsquad will handle it from there if it's a duplicate.
-      - Verify that you are using a [supported Godot version](https://docs.godotengine.org/en/stable/about/release_policy.html).
+      - Verify that you are using a [supported Godot version](https://docs.godotengine.org/en/stable/about/release_policy.html). Please always check if your issue is reproducible in the latest version – it may already have been fixed!
+      - If you use a custom build, please test if your issue is reproducible in official builds too. Likewise if you use any C++ modules, GDExtensions, or editor plugins, you should check if the bug is reproducible in a project without these.
 
-- type: input
+- type: textarea
   attributes:
-    label: Godot version
-    description: >
-      Specify the Godot version, including the Git commit hash if using a development or non-official build. The exact Godot version (including the commit hash) can be copied by clicking the version shown in the editor (bottom bar) or in the project manager (top bar).
-      If you use a custom build, please test if your issue is reproducible in official builds too.
-    placeholder: 3.5.stable, 4.0.dev [3041becc6]
+    label: Tested versions
+    description: |
+      To properly fix a bug, we need to identify if the bug was recently introduced in the engine, or if it was always present.
+      - Please specify the Godot version you found the issue in, including the **Git commit hash** if using a development or non-official build. The exact Godot version (including the commit hash) can be copied by clicking the version shown in the editor (bottom bar) or in the project manager (top bar).
+      - If you can, **please test earlier Godot versions** (previous stable branch, and development snapshots of the current feature release) and, if applicable, newer versions (development snapshots for the next feature release). Mention whether the bug is reproducible or not in the versions you tested. You can find all Godot releases in our [download archive](https://godotengine.org/download/archive/).
+      - The aim is for us to identify whether a bug is a **regression**, i.e. an issue that didn't exist in a previous version, but was introduced later on, breaking existing functionality. For example, if a bug is reproducible in 4.2.stable but not in 4.1.stable, we would like you to test intermediate 4.2 dev and beta snapshots to find which snapshot is the first one where the issue can be reproduced.
+    placeholder: |
+
+      - Reproducible in: 4.3.dev [d76c1d0e5], 4.2.stable, 4.2.dev5 and later 4.2 snapshots.
+      - Not reproducible in: 4.1.3.stable, 4.2.dev4 and earlier 4.2 snapshots.
   validations:
     required: true
 
@@ -54,12 +60,12 @@ body:
 
 - type: textarea
   attributes:
-    label: Minimal reproduction project
+    label: Minimal reproduction project (MRP)
     description: |
       - A small Godot project which reproduces the issue, with no unnecessary files included. Be sure to not include the `.godot` folder in the archive (but keep `project.godot`).
-      - Required, unless the reproduction steps are trivial and don't require any project files to be followed. In this case, write "N/A" in the field.
-      - Drag and drop a ZIP archive to upload it. **Do not select another field until the project is done uploading.**
+      - Having an MRP is very important for contributors to be able to reproduce the bug in the same way that you are experiencing it. When testing a potential fix for the issue, contributors will use the MRP to validate that the fix is working as intended.
+      - If the reproduction steps are not project dependent (e.g. the bug is visible in a brand new project), you can write "N/A" in the field.
+      - Drag and drop a ZIP archive to upload it (max 10 MB). **Do not select another field until the project is done uploading.**
       - **Note for C# users:** If your issue is *not* C#-specific, please upload a minimal reproduction project written in GDScript. This will make it easier for contributors to reproduce the issue locally as not everyone has a .NET setup available.
-      - **If you've been asked by a maintainer to upload a minimal reproduction project, you *must* do so within 7 days.** Otherwise, your bug report will be closed as it'll be considered too difficult to diagnose.
   validations:
     required: true


AThousandShips (approved):
I think this sounds good, and the addition of the suggestion to test recent versions, can't think of anything else to add

Calinou:
Should this be a single-line field? This way, autocompletion from history would be available.

lawnjelly:
Agree that `reproducibility` doesn't sound ideal. `Reproducibility` (to my ears) is suggestive of some kind of random bug that appears only some of the time.

What we are really talking about here is more like `prevalence`, or similar word, to indicate how widespread the bug is in different versions.

Thesaurus gives:
```
commonness
currency
widespread presence
generality
pervasiveness
universality
extensiveness
ubiquity
ubiquitousness
rampancy
rifeness
frequency
regularity
familiarity
acceptance
popularity
fashionableness
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -52,6 +52,15 @@ body:
   validations:
     required: true
 
+- type: textarea
+  attributes:
+    label: Reproducibility
+    description: |
+      Please indicate which Godot version(s) you confirmed this issue on. If you found this issue while testing a pre-release snapshot, it's very useful for us to know if that issue was introduced recently in the development version (what call a regression), or if it already existed in previous stable releases.
```
paddy-exe, line 59:
```suggestion
      Please indicate which Godot version(s) you confirmed this issue on. If you found this issue while testing a pre-release snapshot, it's very useful for us to know if that issue was introduced recently in the development version (which describes a regression), or if it already existed in previous stable releases.
```

paddy-exe: A little grammar / wording nitpick

AThousandShips:
```suggestion
      Please indicate which Godot version(s) you confirmed this issue on. If you found this issue while testing a pre-release snapshot, it's very useful for us to know if that issue was introduced recently in the development version (what's called a regression), or if it already existed in previous stable releases.
```
Alternatively

paddy-exe:
I agree with @lawnjelly that something like `prevalence` describes this better.

akien-mga:
> Should this be a single-line field instead of multiline? This way, autocompletion from history would be available.

I don't think so. Ideally people should test multiple versions so I would expect good usage of this field to be multiline, and issue specific, so it shouldn't be autocompleted. An example could be:

- Reproducible in 4.2-beta4 and later, including 4.2-stable
- Not reproducible in 4.1-stable, 4.1.3-stable, or 4.2-dev1 to 4.2-beta3

> Agree that `reproducibility` doesn't sound ideal. `Reproducibility` (to my ears) is suggestive of some kind of random bug that appears only some of the time.
> 
> What we are really talking about here is more like `prevalence`, or similar word, to indicate how widespread the bug is in different versions.

I agree, though `prevalence` is probably not a word that a lot of non-native speakers would be familiar with (I guess even native speakers might struggle with it :P). The Thesaurus alternatives don't seem much better.

Maybe instead of a dedicated field for this, we can example the "Godot version" input to ask explicitly that users test intermediate versions for regressions. But it might be harder to spot since it will just be described in an already lengthy fine print description.

We could optionally add a placeholder / boilerplate structure so that users know what to include.

AThousandShips:
What about "tested/confirmed versions"?

fire:
In my view "Tested Versions" is more understandable than "Affected versions".

Assisted by AI.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -5,19 +5,25 @@ body:
...
+- type: textarea
   attributes:
-    label: Godot version
-    description: >
-      Specify the Godot version, including the Git commit hash if using a development or non-official build. The exact Godot version (including the commit hash) can be copied by clicking the version shown in the editor (bottom bar) or in the project manager (top bar).
-      If you use a custom build, please test if your issue is reproducible in official builds too.
-    placeholder: 3.5.stable, 4.0.dev [3041becc6]
+    label: Tested versions
+    description: |
+      To properly fix a bug, we need to identify when the bug was introduced in the engine, or if it was always present.
```
YuriSizov, line 19:
```suggestion
      To properly fix a bug, we need to identify if the bug was recently introduced in the engine, or if it was always present.
```
I feel like these two parts are disconnected otherwise.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -5,19 +5,25 @@ body:
...
+      - If you can, **please test earlier Godot versions** (previous stable branch, and development snapshots of the current feature release) and, if applicable, newer versions (development snapshots for the next feature release). Mention whether the bug is reproducible or not in the versions you tested. You can find all Godot releases on our [download archive](https://godotengine.org/download/archive/).
+      - The aim is for us to identify whether a bug is a **regression**, i.e. an issue that didn't exist in a previous version, but was introduced later on, breaking existing functionality. For example, if a bug is reproducible in 4.2.stable but not in 4.1.stable, we would like you to test intermediate 4.2 dev and beta snapshots to find which snapshot is the first one where the issue can be reproduced.
+    placeholder: |
+
+      - Reproducible in: 4.3.dev [d76c1d0e5], 4.2.stable, 4.2-dev5 and later 4.2 snapshots.
+      - Not reproducible in: 4.1.3-stable, 4.2-dev4 and earlier 4.2 snapshots.
```
akien-mga, line 26: That empty first line in the placeholder was needed because for `textarea` fields, GitHub automatically adds a "Leave a comment" placeholder when you click the input field... and it overlaps the placeholder configured in the template :man_facepalming: 

See the screenshot in https://github.com/godotengine/godot/pull/85554#issuecomment-1838872518

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -5,19 +5,25 @@ body:
...
+      - Please specify the Godot version you found the issue in, including the **Git commit hash** if using a development or non-official build. The exact Godot version (including the commit hash) can be copied by clicking the version shown in the editor (bottom bar) or in the project manager (top bar).
+      - If you can, **please test earlier Godot versions** (previous stable branch, and development snapshots of the current feature release) and, if applicable, newer versions (development snapshots for the next feature release). Mention whether the bug is reproducible or not in the versions you tested. You can find all Godot releases on our [download archive](https://godotengine.org/download/archive/).
+      - The aim is for us to identify whether a bug is a **regression**, i.e. an issue that didn't exist in a previous version, but was introduced later on, breaking existing functionality. For example, if a bug is reproducible in 4.2.stable but not in 4.1.stable, we would like you to test intermediate 4.2 dev and beta snapshots to find which snapshot is the first one where the issue can be reproduced.
+    placeholder: |
+
```
YuriSizov, line 24: Excessive empty line, I think
```suggestion
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -5,19 +5,25 @@ body:
...
       - The golden rule is to **always open *one* issue for *one* bug**. If you notice several bugs and want to report them, make sure to create one new issue for each of them.
       - Search [open](https://github.com/godotengine/godot/issues) and [closed](https://github.com/godotengine/godot/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported. If you don't find a relevant match or if you're unsure, don't hesitate to **open a new issue**. The bugsquad will handle it from there if it's a duplicate.
-      - Verify that you are using a [supported Godot version](https://docs.godotengine.org/en/stable/about/release_policy.html).
+      - Verify that you are using a [supported Godot version](https://docs.godotengine.org/en/stable/about/release_policy.html). Please always check if your issue is reproducible in the latest version – it may already have been fixed!
+      - If you use a custom build, please test if your issue is reproducible in official builds too. Likewise if you use any C++ modules, GDExtensions, or editor plugins, you should check if the bug is reproducible in a project without these.
 
-- type: input
+- type: textarea
```
akien-mga, line 15: Making this a `textarea` instead of `input` means we lose browser history, so if people want to report multiple issues against the same Godot version, they won't benefit from being able to select it easily.

But for cases where users really test old versions (or for us to backfill after doing our own testing), this really needs to be a multiline field IMO.

YuriSizov (approved):
I generally like this approach. I feel like there is too much text in the form in general, not sure if the message we try to send gets lost in the process.

But I don't have a better idea yet. So unless someone else has, I would go for it.

.github/ISSUE_TEMPLATE/bug_report.yml
YuriSizov: Nevermind.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -60,6 +66,5 @@ body:
       - Required, unless the reproduction steps are trivial and don't require any project files to be followed. In this case, write "N/A" in the field.
       - Drag and drop a ZIP archive to upload it. **Do not select another field until the project is done uploading.**
       - **Note for C# users:** If your issue is *not* C#-specific, please upload a minimal reproduction project written in GDScript. This will make it easier for contributors to reproduce the issue locally as not everyone has a .NET setup available.
-      - **If you've been asked by a maintainer to upload a minimal reproduction project, you *must* do so within 7 days.** Otherwise, your bug report will be closed as it'll be considered too difficult to diagnose.
```
akien-mga, line 63: We discussed removing this line a few weeks ago, as it sounds much stricter than we actually are. MRPs are a best effort and the lack thereof isn't reason for closing an issue, unless we estimate that there's really nothing of value in the bug report.

AThousandShips: Might word it more like "If the steps are not trivial, or others are unable to reproduce your bug from the reproduction steps, not providing a minimal reproduction project might lead to the issue being closed, as nothing can be done to confirm the bug. It also helps confirming that a bug has been solved."

Actually I think making a point about how the MRP isn't *just* about confirming the bug is important, making it clear it's also for confirming a fix is helpful as some don't realize it and think it's just to "prove" their bug exists.

akien-mga: Added some stuff, WDYT?

AThousandShips: Looks great!

AThousandShips (approved):
Looks great! Puts a lot more of the information at your fingertips and should help some confusion from people reporting bugs

PR #85556, opened 2023-11-30
Fix importing WAV files with odd chunk sizes

### LINKED ISSUE #85466: WAV file that works everywhere else is silent in Godot
### Godot version

4.2.rc2,3.5.1

### System information

Godot v4.2.rc (fc7920185) - NixOS 23.11 (Tapir) - Wayland - Vulkan (Mobile) - integrated Intel(R) Graphics (ADL GT2) - 12th Gen Intel(R) Core(TM) i5-1240P (16 Threads)

### Issue description

This short 1.5second WAV file came from a site with free sounds (None of the other sounds from the same source worked either).  

Godot imports this sound file with a duration of 0, so it doesn't play anything.  There are no error messages that I can see.  I have attempted to re-import it with various different settings, and have tried it with an online inspection tool that didn't throw up any clues.  This file plays properly in Audacity and VLC.

This problem is the same for Godot3.5.1, and probably applies to all versions.

### Steps to reproduce

Unzip and preview Cloth_02.wav in Godot.  It doesn't play.

The second file Cloth_02_savedfromaudacity.wav is what happens if I load and save it from Audacity.  Then Godot loads and plays it as it should.

[Cloth_02.zip](https://github.com/godotengine/godot/files/13488036/Cloth_02.zip)

N/A

bs-mwoerner:
Fixes #85466.

The RIFF specification requires a padding byte to be written when a chunk size is not even. The current WAV importer did not include that byte when moving to the next chunk, so the import would fail (read a zero-length file) when the file contained a chunk with an odd size before the actual `data` chunk.

### DIFF
--- a/editor/import/resource_importer_wav.cpp
+++ b/editor/import/resource_importer_wav.cpp
@@ -292,7 +292,9 @@ Error ResourceImporterWAV::import(const String &p_source_file, const String &p_s
 				loop_end = file->get_32();
 			}
 		}
-		file->seek(file_pos + chunksize);
+		// Move to the start of the next chunk. Note that RIFF requires a padding byte for odd
+		// chunk sizes.
+		file->seek(file_pos + chunksize + (chunksize & 1));
 	}
 
 	// STEP 2, APPLY CONVERSIONS


lawnjelly (approved):
I'm no expert in RIFF files but it does seem like chunks should be padded like this:
https://www.daubnet.com/en/file-format-riff

> unused 1 byte present, if size is odd

akien-mga:
Same email/account attribution issue on this PR :)

bs-mwoerner:
Merci. :)

YuriSizov:
Cherry-picked for 4.2.1.

PR #85585, opened 2023-12-01
Fix SnapGrid is almost invisible in light theme

### LINKED ISSUE #85574: "Use Grid Snap" icon is hardly visible with Light theme
### Godot version

### System information

Linux (Fedora Silverblue), v4.2.stable.official [46dc27791], AMD RX570, Forward+

### Issue description

I have hard time looking at Grid Snap icon in light theme:

зображення

Also, when this button active, it have cyan color, while other such buttons become blue:
зображення
зображення

### Steps to reproduce

1. Open Godot
2. Open any project
3. Editor -> Editor Settings...
4. Interface -> Theme
5. Select "Light" in "Preset"

Not specific to any project.

jsjtxietian:
Fixes #85574

cc @MewPurPur Any suggestions?


MewPurPur:
This was my mistake, I mixed this up with work on icons for my own app (which uses #def as its icon color). Godot's icon color is e0e0e0, so please use that.

jsjtxietian:
> Godot's icon color is e0e0e0, so please use that.

Thanks! I first use that too, but then find the other snap icon (Snap.svg)  is #fff so I used that, should I change that too ?

MewPurPur:
Those icons are likely slightly broken in light theme too. So yes.

akien-mga:
There seems to be a lot of icons using `fill="#fff"` (and some use `#fff` for `stroke` too.
If this should be changed to the remappable `#e0e0e0`, this might best be done in a dedicated PR. But I assume some of them use `#fff` intentionally to keep it white even on light theme (e.g. eyes of Godot logo).

```
$ rg -l 'fill="#fff"' | wc -l
87
```

MewPurPur:
Let me look into it.

MewPurPur:
I don't find anything else that needs to be changed. I should mention that #fefefe is the "forced white" color, white #fff becomes a dark gray when it's in light theme. And #e0e0e0 is the icon color, it's a very light gray, and in light theme it's a dark gray (but less dark than the one #fff becomes.)

A lot of the occurrences of #fff are outside editor/icons, which is correct and shouldn't be changed. In editor/icons, I also didn't find any occurrences of it being used for icons. It's commonly used for GUI elements (I consider it not icons, also opacity is used to determine how much those stick out) and for contrast, mainly for icons like Transform2D, which is just "T2D" as an icon and it uses a transparent white to add extra contrast to the middle letter, so as to unstick them and make the icon more readable. Also for icons like the overbright indicator, trying to have as much contrast as possible between white and black. These shouldn't be changed either. Actually, they could use #fefefe and get removed from the list of exceptions. Then there is the Godot logos using it for their eyes, obviously shouldn't be changed.

In short, I didn't find any other icons that need their fill color tweaked from #fff to #e0e0e0.

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.1.

PR #85594, opened 2023-12-01
Prevent `Camera3D.current` from being set to `true` automatically in the editor.

### LINKED ISSUE #85593: `Camera3D.current` is set to true when setting `Camera3D.current` to false on another node
### Godot version

Godot v4.2.stable 

### System information

Windows 10.0.22631 - Vulkan (Forward+) - dedicated AMD Radeon RX 7900 XTX (Advanced Micro Devices, Inc.; 31.0.22023.1014) - AMD Ryzen 7 7800X3D 8-Core Processor (16 Threads)

### Issue description

When you disable `current` on a Camera3D, the first available Camera3D in the scene will have it's `current` property set to true.

### Steps to reproduce

Add 2 Camera3D nodes to a scene, note both have `current` disabled.
Enable and disable `current` on one Camera3D.
After `current` is disabled on one Camera3D, the other Camera3D has it's `current` property set to true

[minimal.zip](https://github.com/godotengine/godot/files/13522846/minimal.zip)

addmix:
Fixes #85593

### DIFF
--- a/scene/3d/camera_3d.cpp
+++ b/scene/3d/camera_3d.cpp
@@ -272,7 +272,7 @@ void Camera3D::clear_current(bool p_enable_next) {
 	if (get_viewport()->get_camera_3d() == this) {
 		get_viewport()->_camera_3d_set(nullptr);
 
-		if (p_enable_next) {
+		if (p_enable_next && !Engine::get_singleton()->is_editor_hint()) {
 			get_viewport()->_camera_3d_make_next_current(this);
 		}
 	}


RedMser:
Would the same fix need to be done for Camera2D? It also has "assign next camera as current" logic, looking at its code.

addmix:
> Would the same fix need to be done for Camera2D? It also has "assign next camera as current" logic, looking at its code.

I gave Camera2D a quick look, and I don't believe it needs the same fix, as Camera2D doesn't have the `current` property.

KoBeWi (approved):
Yeah, Camera2D doesn't need it. The logic is a bit different.

YuriSizov:
Thanks!

PR #85603, opened 2023-12-01
Speed up `GDScript::get_must_clear_dependencies()`

### LINKED ISSUE #85435: Exponentially slow quitting time with dependencies between (gdscript) scripts
### Godot version

Godot v4.2.rc2

### System information

Godot v4.2.rc2 - Ubuntu 22.04.3 LTS 22.04 - X11 - Vulkan (Forward+) - integrated Intel(R) UHD Graphics (CML GT2) () - Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz (12 Threads)

### Issue description

When there is a lot of dependencies between types defined with class_name or class, the quitting time for game (and editor) grows exponentially.

With a "real" project (nb of scripts: 97, nb lignes of code: 11638, nb dependencies between script: 336) the quitting time is ~8.0s.
I could manage to reduce it to ~5.0s just by moving an inner class from one file to another.

On a "toy" project, with N scripts doing nothing except depending on each others, the quitting time seems exponential:

- 40 scripts : 1.5s
- 50 scripts : 2.5s
- 60 scripts : 4.5s
- 70 scripts : 8.4s
- 80 scripts : 15.7s
- 90 scripts : 27.1s
- 100 scripts : 46.8s

Adding the dependencies as a type check (`if my_var is MyType`) triggers the issue.
Adding the dependencies as a type creation (`MyType.new()`) triggers the issue.
Adding the dependencies as type hinting (`var my_var : MyType`) doesn't triggers the issue.

Adding inner (unused) classes inside these script increases a lot the time.

The dependencies being cyclical or not doesn't seems to change the behavior.

Same issue with 4.1.2-stable.

### Steps to reproduce

1) Create N scripts, each depending on each others

for example for script0.gd:

```
class_name Script0

static func some_func():
    print(Script0.new())
    print(Script1.new())
    print(Script2.new())
    print(Script3.new())
    ...

```

2) in the script of the main scene, add a dependency to one of the N inter-depending scripts, for example with `print(null is Script0)`

3) then quit directly the scene, with `get_tree().quit()`

4) launch the main scene

If N is high enough (>50), it will start to quit directly, but the quitting time will be long.

5) quit the editor.

The quitting time will also be quite long.

note: a python script has been provided in the minimal project to create all the N inter-depending gdscript scripts of step 1.

[quitting_time_dependencies.zip](https://github.com/godotengine/godot/files/13476840/quitting_time_dependencies.zip)

eldidou:
get_must_clear_dependencies() has a N^3*log(N) time complexity, and this can very quickly slow down the quitting process as more gdscripts are added in a project. This change improves it to N^2*log(N).
Instead of using all the inverted dependencies, we do the same with all (non-inverted) dependencies, which is N times faster.

Fixes #85435

### DIFF
--- a/modules/gdscript/gdscript.cpp
+++ b/modules/gdscript/gdscript.cpp
@@ -1156,8 +1156,8 @@ RBSet<GDScript *> GDScript::get_dependencies() {
 	return dependencies;
 }
 
-RBSet<GDScript *> GDScript::get_inverted_dependencies() {
-	RBSet<GDScript *> inverted_dependencies;
+HashMap<GDScript *, RBSet<GDScript *>> GDScript::get_all_dependencies() {
+	HashMap<GDScript *, RBSet<GDScript *>> all_dependencies;
 
 	List<GDScript *> scripts;
 	{
@@ -1171,51 +1171,42 @@ RBSet<GDScript *> GDScript::get_inverted_dependencies() {
 	}
 
 	for (GDScript *scr : scripts) {
-		if (scr == nullptr || scr == this || scr->destructing) {
+		if (scr == nullptr || scr->destructing) {
 			continue;
 		}
-
-		RBSet<GDScript *> scr_dependencies = scr->get_dependencies();
-		if (scr_dependencies.has(this)) {
-			inverted_dependencies.insert(scr);
-		}
+		all_dependencies.insert(scr, scr->get_dependencies());
 	}
 
-	return inverted_dependencies;
+	return all_dependencies;
 }
 
 RBSet<GDScript *> GDScript::get_must_clear_dependencies() {
 	RBSet<GDScript *> dependencies = get_dependencies();
 	RBSet<GDScript *> must_clear_dependencies;
-	HashMap<GDScript *, RBSet<GDScript *>> inverted_dependencies;
-
-	for (GDScript *E : dependencies) {
-		inverted_dependencies.insert(E, E->get_inverted_dependencies());
-	}
+	HashMap<GDScript *, RBSet<GDScript *>> all_dependencies = get_all_dependencies();
 
 	RBSet<GDScript *> cant_clear;
-	for (KeyValue<GDScript *, RBSet<GDScript *>> &E : inverted_dependencies) {
+	for (KeyValue<GDScript *, RBSet<GDScript *>> &E : all_dependencies) {
+		if (dependencies.has(E.key)) {
+			continue;
+		}
 		for (GDScript *F : E.value) {
-			if (!dependencies.has(F)) {
-				cant_clear.insert(E.key);
-				for (GDScript *G : E.key->get_dependencies()) {
-					cant_clear.insert(G);
-				}
-				break;
+			if (dependencies.has(F)) {
+				cant_clear.insert(F);
 			}
 		}
 	}
 
-	for (KeyValue<GDScript *, RBSet<GDScript *>> &E : inverted_dependencies) {
-		if (cant_clear.has(E.key) || ScriptServer::is_global_class(E.key->get_fully_qualified_name())) {
+	for (GDScript *E : dependencies) {
+		if (cant_clear.has(E) || ScriptServer::is_global_class(E->get_fully_qualified_name())) {
 			continue;
 		}
-		must_clear_dependencies.insert(E.key);
+		must_clear_dependencies.insert(E);
 	}
 
 	cant_clear.clear();
 	dependencies.clear();
-	inverted_dependencies.clear();
+	all_dependencies.clear();
 	return must_clear_dependencies;
 }
 
--- a/modules/gdscript/gdscript.h
+++ b/modules/gdscript/gdscript.h
@@ -254,7 +254,7 @@ public:
 	const Ref<GDScriptNativeClass> &get_native() const { return native; }
 
 	RBSet<GDScript *> get_dependencies();
-	RBSet<GDScript *> get_inverted_dependencies();
+	HashMap<GDScript *, RBSet<GDScript *>> get_all_dependencies();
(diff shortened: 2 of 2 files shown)


akien-mga:
For the record, your commit seems not to be linked to your GitHub account. See: [Why are my commits linked to the wrong user?](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user) for more info.

kleonc:
> get_must_clear_dependencies() has a N^3_log(N) time complexity, and this can very quickly slow down the quitting process as more gdscripts are added in a project. This change improves it to N^2_log(N).

This is an improvement indeed but I think `GDScriptLanguage::finish()` shouldn't lead to any checks for dependencies like that? :thinking: On finalization all scripts are to-be-cleared anyway, so there's no point to check/clear the full chain of dependencies for every script. Iterating over every script and clearing only its own refs to any other script should be sufficient to get rid of all cross dependencies. Meaning the finalization should be doable linearly (O(N)).

Note that I'm not really familiar with the GDScript internals so maybe I'm missing something and it's not that simple?

Anyway, could potentially be done in another PR, this is already an improvement.

kleonc:
> But this isn't about `GDScriptLanguage::finish()`, isn't it?

It is and it isn't, depends how you look at it. :upside_down_face:

#85435 which this PR aims to fix is specifically about slow quiting (both editor/runtime) when there are many cross referencing/dependant scripts. This is caused by the slow execution of `GDScriptLanguage::finish()`. And why is `GDScriptLanguage::finish()` so slow? Because it's calling `GDScript:clear()` many times and `GDScript::clear()` is slow with the current implementation.

So this PR makes `GDScript:clear()` faster which is good on its own, regardless of what calls `GDScript::clear()`. :+1: This of course makes also `GDScriptLanguage::finish()` faster as a result, which improves the quitting time.

What I've said in https://github.com/godotengine/godot/pull/85603#issuecomment-1836988340 is that `GDScriptLanguage::finish()` can potentially be made faster as it shouldn't need to call `GDScript:clear()` in its current form at all, because on finalizing all scripts clearing non-direct script dependencies shouldn't be needed (and that's what's done as a part of `GDScript:clear()`). Clearing only direct dependencies for all scripts would also result in no cross dependencies in the end and would be faster.

TLDR: this PR is good on its own but `GDScriptLanguage::finish()` could still be made faster.

---

BTW Yeah, would be nice if some benchmark times would be provided for before/after. In the end even if it helps with #85435 (it does) it still might need be considered as not fixing it if the timings are still not satisfying.

adamscott (commented):
@eldidou Thanks for your PR! For a first PR, that's a nice one!

I'm trying to see if there's things that could break with your implementation, but the logic seems fine. The idea to skip the dependencies if they don't include the current script is a good idea. It does what I thought to do when I implemented inverse dependency checks.

I asked @vnen and @dalexeev to review it, but so far, I like it.

modules/gdscript/gdscript.cpp
```diff
@@ -1171,51 +1171,41 @@ RBSet<GDScript *> GDScript::get_inverted_dependencies() {
...
 
 	RBSet<GDScript *> cant_clear;
-	for (KeyValue<GDScript *, RBSet<GDScript *>> &E : inverted_dependencies) {
-		for (GDScript *F : E.value) {
-			if (!dependencies.has(F)) {
-				cant_clear.insert(E.key);
-				for (GDScript *G : E.key->get_dependencies()) {
-					cant_clear.insert(G);
+	for (KeyValue<GDScript *, RBSet<GDScript *>> &E : all_dependencies) {
+		if (!dependencies.has(E.key)) {
```
adamscott, line 1190: Super small nitpick, if you could reverse the condition with a continue, it would be easier to read.
```suggestion
		if (dependencies.has(E.key)) {
			continue;
		}
```

RevoluPowered:
> Can confirm that with The Mirror we see a hugely expensive shutdown time because of this function.
> 
> Benchmarked using superluminal profiler and it showed this function was consuming 100% CPU for over 90 seconds during shutdown.
> 
> I'll test this fix and report back the app close time :)

4 seconds to close the editor (previously multiple minutes)
4 seconds to close the app (previously 2 minutes or more)

This makes it much more deterministic for me.

Tested on windows.

modules/gdscript/gdscript.cpp
eldidou: done

AThousandShips: Did you forget to push this?

eldidou: Indeed...
Done. For real this time.

adamscott (approved):
Seems fine to me. It improved my logic and I don't see how the new logic would not cover my old one.

Especially with all the tests passing, I'm OK merging this very soon.

YuriSizov:
Thanks! And congratulations on your first merged Godot contribution! That's a very cool way to enter the space 🙂

eldidou:
Thanks! Glad to help 🙂

adamscott:
@eldidou Don't hesitate to join our [developer's chat](https://chat.godotengine.org). We have a GDScript public meeting tomorrow.

YuriSizov:
Cherry-picked for 4.2.2.

mihe:
Seeing as how this slowdown still remains a problem when you go into the hundreds of scripts, I've gone ahead and made the change that was talked about earlier in this thread, by simply skipping the `must_clear_dependencies` work altogether. You'll find the change in #94505.

It seemed like a straight-forward enough of a change, and passes all the unit tests and whatnot, but I'd appreciate some more eyes on it.

Issue #48064, opened 2021-04-21
When packing a PackedScene into a scene, signals and scripts of that PackedScene are duplicated into the parent scene


Synzorasize:
**Godot version:**
3.2.3

**OS/device including version:**
Windows 10

**Issue description:**
I was working on a level editor for my game, and I decided to use PackedScenes for the tested levels since scenes are the default format for my original levels.
I also have PackedScenes within those scenes, though. Those are the tiles, hazards, orbs, player, etc. They all obviously have scripts attached to them (except for the regular tiles). And the hazards and orbs' body_entered are connected to themselves (in their own scene) to detect if the player touched them or not.
When I play the level, it shows this error:
```
E 0:01:28.051   connect: Signal 'body_shape_entered' is already connected to given method '_on_Hazard_body_shape_entered' in that object.
  <C++ Error>   Method failed. Returning: ERR_INVALID_PARAMETER
  <C++ Source>  core/object.cpp:1515 @ connect() # I am not even using the connect() method, haha...
  <Stack Trace> CustomLevel.gd:75 @ _on_PlayButton_pressed()
```
It appears to be blaming this line in the CustomLevel node's (The main parent node for all the elements in the level editor) script:
`	get_tree().change_scene_to(GlobalVariables.level) # The variable in a singleton which the CustomLevel node packs into.`

Of course, it isn't crashing or breaking my game, but it is annoying to see this error for _each_ node pop up every time I go to this scene or reload it. The scene would probably also take slightly more time to load due to this error. I would also want to focus on some real issues, too.

This is the code in the CustomLevel script to pack everything (since the CustomLevel node isn't the root node, all the children won't be packed if I don't set the owner to itself):
```
func saveLevel():
	position = Vector2(0,0)
	descendants = get_children() + [UI] # the variable "UI" stands for the UI node of the game which is a child of a CanvasLayer.
	for i in descendants:
		i.owner = self
	GlobalVariables.level.pack(self)

func _on_PlayButton_pressed():
	UI.show()
	saveLevel()
#	ResourceSaver.save("glitchedLevel.tscn", GlobalVariables.level) # I wanted to test what it looked like as a .tscn file
	get_tree().change_scene_to(GlobalVariables.level)
	GlobalVariables.editor = false
	GlobalVariables.editor_playing = true
```
[...]

PR #85606, opened 2023-12-01
Fix unable to determine which scene a connection belongs to when packing outside the editor

### LINKED ISSUE #85372: instantiate packed scene with editable children and editor connected signal cause error "signal already connected"
### Godot version

4.2.rc2

### System information

win11

### Issue description

instantiate packed scene with editable children and editor connected signal cause error "signal already connected"

image
image

image
image
image

### Steps to reproduce

1. make a scene with a signal connection via editor(from child to parent). 
2. make another scene, add the previous scene as a child, and turn on "editable children". 
3. make yet another scene, then add the new scene in step 2 as a child
4. in the third scene, use PackedScene.pack() to pack the child
5. instantiate the packed scene
6. observe the error "Signal '****' is already connected to given callable"

[BugSignalBug.zip](https://github.com/godotengine/godot/files/13466207/BugSignalBug.zip)

Rindbee:
Previously, the lack of scene inherited/instance state outside the editor made it impossible to determine which scene a connection in a descendant scene with editable children enabled belonged to. This may result in connections already recorded in a descendant scene being recorded again into a new `PackedScene` when packed outside the editor.
    

Exclude two cases to speed up parse:
1. A scene root that does not have editable children enabled does not have to parse its child nodes.
2. It is not necessary to parse descendant nodes where `p_owner` is not in the descendant node's owner chain.

Fix #85372, fix partially #48064.

### DIFF
--- a/scene/main/node.cpp
+++ b/scene/main/node.cpp
@@ -1998,7 +1998,7 @@ void Node::_clean_up_owner() {
 	data.OW = nullptr;
 }
 
-Node *Node::find_common_parent_with(const Node *p_node) const {
+Node *Node::find_common_owner_with(const Node *p_node) const {
 	if (this == p_node) {
 		return const_cast<Node *>(p_node);
 	}
@@ -2009,23 +2009,23 @@ Node *Node::find_common_parent_with(const Node *p_node) const {
 
 	while (n) {
 		visited.insert(n);
-		n = n->data.parent;
+		n = n->data.owner;
 	}
 
-	const Node *common_parent = p_node;
+	const Node *common_owner = p_node;
 
-	while (common_parent) {
-		if (visited.has(common_parent)) {
+	while (common_owner) {
+		if (visited.has(common_owner)) {
 			break;
 		}
-		common_parent = common_parent->data.parent;
+		common_owner = common_owner->data.owner;
 	}
 
-	if (!common_parent) {
+	if (!common_owner) {
 		return nullptr;
 	}
 
-	return const_cast<Node *>(common_parent);
+	return const_cast<Node *>(common_owner);
 }
 
 NodePath Node::get_path_to(const Node *p_node, bool p_use_unique_path) const {
--- a/scene/resources/packed_scene.cpp
+++ b/scene/resources/packed_scene.cpp
@@ -873,10 +873,6 @@ Error SceneState::_parse_node(Node *p_owner, Node *p_node, int p_parent_idx, Has
 }
 
 Error SceneState::_parse_connections(Node *p_owner, Node *p_node, HashMap<StringName, int> &name_map, HashMap<Variant, int, VariantHasher, VariantComparator> &variant_map, HashMap<Node *, int> &node_map, HashMap<Node *, int> &nodepath_map) {
-	if (p_node != p_owner && p_node->get_owner() && p_node->get_owner() != p_owner && !p_owner->is_editable_instance(p_node->get_owner())) {
-		return OK;
-	}
-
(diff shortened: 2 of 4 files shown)


KoBeWi (dismissed):
This does fix the issues and is *probably* alright, but I don't really understand what GEN_EDIT_STATE_DISABLED is for. The PR makes it technically unused (the only "use" is that you can't use another EDIT STATE in non-editor builds).

Rindbee:
Squeezed the commit.

YuriSizov:
> I don't really understand what GEN_EDIT_STATE_DISABLED is for. The PR makes it technically unused (the only "use" is that you can't use another EDIT STATE in non-editor builds).

Seems like it always has been a thing. Let's try asking @reduz, maybe he can remember the use case for this.

reduz:
This PR does not make a lot of sense honestly, you don't want to keep the instance state around when running outside the editor so keeping it just to fix a very corner case does not seem like the right to solve it.

reduz:
My feeling is that you probably want to instantiate in the right mode yourself when running outside the editor (by forcefully generating the instance state). If you don't, then packing will not work as expected and I understand this is on purpose.

Rindbee:
> This PR does not make a lot of sense honestly, you don't want to keep the instance state around when running outside the editor (this has a memory/performance impact) so keeping it just to fix a very corner case does not seem like the right way to solve it.

In this case, it appears that the behavior described in #48064 will be an expected behavior.

scene/main/node.cpp
```diff
@@ -1998,7 +1998,7 @@ void Node::_clean_up_owner() {
 	data.OW = nullptr;
 }
 
-Node *Node::find_common_parent_with(const Node *p_node) const {
+Node *Node::find_common_owner_with(const Node *p_node) const {
```
Rindbee, line 2001: The original method `find_common_parent_with()` is used as a pre-step to find the common owners.

But looking up the common owner along the ownership chain seems to be more efficient, especially when `editable_children` is enabled.

scene/resources/packed_scene.cpp
```diff
@@ -927,95 +923,59 @@ Error SceneState::_parse_connections(Node *p_owner, Node *p_node, HashMap<String
...
-						}
-						nl = nl->get_owner();
-					}
-				}
-
-				if (exists2) {
-					continue;
-				}
-			}
-
```
Rindbee, line 1018: This code seems to have the same effect as the previous check along the ownership chain. Doesn't seem to catch some different cases.

scene/resources/packed_scene.cpp
```diff
@@ -1307,39 +1300,6 @@ void SceneState::set_disable_placeholders(bool p_disable) {
...
-			} else if (names[i] == p_to_method) {
-				method_idx = i;
-			}
-		}
-
-		if (signal_idx >= 0 && method_idx >= 0) {
-			//signal and method strings are stored..
-
-			for (int i = 0; i < connections.size(); i++) {
-				if (connections[i].from == p_node && connections[i].to == p_to_node && connections[i].signal == signal_idx && connections[i].method == method_idx) {
```
Rindbee, line 1329: `has_connection()` has the same purpose as `is_connection()`, but `connections[i].from` may also be an index in `names`, so the results of `has_connection()` may be more correct.

cixil:
Hi I may have a simpler solution to the signal duplication bugs: https://github.com/godotengine/godot/pull/97303

I just noticed this PR after submitting mine and I'm not sure what the state of it is, but I wanted to link mine here because having the same reviewers might be helpful.

PR #68548, opened 2022-11-12
Add premul alpha blending to 3D (spatial) shaders

jitspoe:
This is the Godot Engine 4 version of PR https://github.com/godotengine/godot/pull/36747 to add alpha premul blending to the 3D spatial shaders and materials (already supported for 2D CanvasItem shaders).

_Production edit: Closes https://github.com/godotengine/godot-proposals/issues/3431_

PR #36747, opened 2020-03-03
[3.x] Add blend_premul_alpha support to scene shaders.

jitspoe:
This was previously only supported by 2D shaders, but can be pretty useful in 3D as well.  For example, I'm using it for lasers where I don't want to use pure additive blending, because they lose visibility with bright backgrounds, but a normal alpha blend doesn't look good, either, so I make them mostly additive with some slight alpha to ensure visibility on brighter areas.

I'm making this pull request directly into 3.2 because the gles3 renderer was removed and I can't make the pull request in master and have it merged back to 3.2.  (Also, can't compile master due to some internal compiler error, so I'll have to figure that out and make a different pull request for master to support vulkan after I figure that out, if this goes through).

*Bugsquad edit: This addresses https://github.com/godotengine/godot-proposals/issues/3431 for `3.x`.*

PR #85609, opened 2023-12-01
Add premult alpha blending to 3D (spatial) shaders

QbieShay:
Follow-up to https://github.com/godotengine/godot/pull/68548
Implements https://github.com/godotengine/godot-proposals/issues/3431#issuecomment-1757999180


OhiraKyou:
Detecting the use of a `PREMULT_ALPHA` built-in to enable a specific blend mode is inconsistent with other blend modes and doesn't always make sense.

Consider the following common use case: a user is intending to use a texture whose RGB is already premultiplied. This would require the user to assign a dummy `PREMULT_ALPHA` value of `1` to enable the blend mode without redundant multiplication.

I suggest the following:

1. Merge @jitspoe's [existing blend mode PR](https://github.com/godotengine/godot/pull/36747) to allow the blend mode to be used on its own.
2. Add a general-purpose `DELAYED_RGB_MULTIPLIER` `vec3` built-in as a separate feature, instead of a blend mode specific `PREMULT_ALPHA`. This multiplier would be useful any time the source blend mode is `one`, including both premultiplied alpha and additive blending.

The "DELAYED" part of the name is important, as it communicates that it's the _timing_ of the multiplier's application (after the sRGB to linear conversion) that justifies its usage. And, the `vec3` type, instead of `float`, enables each of the RGB channels to be modified independently.

QbieShay:
@OhiraKyou could you give me an example usecase of this multiplier being a vec3? what kind of effect would it enable?

OhiraKyou:
First, `vec3 * float` multiplication is per-component anyway. So, it's just good form to expose each component's multiplier to the user by making the multiplier a `vec3` as well. That said, here are a couple of specific use cases.

## Assumptions

The use cases described below assume the following:

- The multiplier is generally available rather than coupled to the premultiplied alpha blend mode—which it really doesn't need to be coupled to.
- The multiplier is applied after the sRGB to linear conversion and after built-in lighting.

## Use cases

### Custom channel mixing

Custom, per-channel post-processing of shader output colors—that includes lighting—may be applied in linear space by assigning a color multiplier to the delayed multiplier.

This includes processing of per-channel, non-color data for custom buffer rendering.

### Custom colored lighting

Custom colored lighting (with no actual light objects in a scene) may be implemented in linear space in otherwise sRGB compatibility renderer shaders by ignoring the existing lighting system (i.e., using `unshaded` shaders) and assigning custom light output to the delayed multiplier.

QbieShay:
We've discussed this PR in the rendering meeting and the design/thinking process around this is the following:

Like everything in Godot, we want to first talk about problems, then solutions.
The usecases you have provided are more along the lines of "if this was there, I could use it like this" rather than "I have this problem and I can't solve it without this".

Light multiplication is something I personally really want to do. I am hoping to be able to eventually have a post-light function (this is a wish, not a promise). But we need to respond with solution when there's a problem to begin with. But this is a whole different problem than what we're trying to solve here.
The problem to begin with here is having premultiplied alpha in the engine, so we should focus of solving that problem.

Additionally, concerns have been brought up about leaving doors open for usecases that are not explicitely supported: this may be a problem in the future when optimizing shaders or refactoring them. I understand that we could say "any use outside of premul alpha is unsupported and you're on your own if you do it" but this historically has never worked: people depend on it, and are disappointed when we "break" things that were never supposed to work to begin with.

I will bring this PR to completion in this dev cycle (4.3), unless something comes up, so that we can finally have premul alpha in 3D ^^

scene/resources/material.cpp
```diff
@@ -2731,7 +2739,7 @@ void BaseMaterial3D::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "alpha_hash_scale", PROPERTY_HINT_RANGE, "0,2,0.01"), "set_alpha_hash_scale", "get_alpha_hash_scale");
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "alpha_antialiasing_mode", PROPERTY_HINT_ENUM, "Disabled,Alpha Edge Blend,Alpha Edge Clip"), "set_alpha_antialiasing", "get_alpha_antialiasing");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "alpha_antialiasing_edge", PROPERTY_HINT_RANGE, "0,1,0.01"), "set_alpha_antialiasing_edge", "get_alpha_antialiasing_edge");
-	ADD_PROPERTY(PropertyInfo(Variant::INT, "blend_mode", PROPERTY_HINT_ENUM, "Mix,Add,Subtract,Multiply"), "set_blend_mode", "get_blend_mode");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "blend_mode", PROPERTY_HINT_ENUM, "Mix,Add,Subtract,Multiply,Premult Alpha"), "set_blend_mode", "get_blend_mode");
```
AThousandShips, line 2742:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::INT, "blend_mode", PROPERTY_HINT_ENUM, "Mix,Add,Subtract,Multiply,Premultiplied Alpha"), "set_blend_mode", "get_blend_mode");
```
As per https://github.com/godotengine/godot/pull/68548#discussion_r1353657033

doc/classes/BaseMaterial3D.xml
```diff
@@ -608,6 +608,9 @@
 		<constant name="BLEND_MODE_MUL" value="3" enum="BlendMode">
 			The color of the object is multiplied by the background.
 		</constant>
+		<constant name="BLEND_MODE_PREMULT_ALPHA" value="4" enum="BlendMode">
+			The color of the object is added to the background and the alpha channel is used to mask out the background. This is effectively a hybrid of blend mix and add mode, useful for FX like fire where you want the flame to add but the smoke to mix. By default, this works with unshaded materials using premultiplied textures. This may not behave as expected with a shaded material, and if you write your own shader, you will need to manually premultiply the albedo with the alpha.
```
clayjohn, line 612: This needs to be updated. I think we should highlight two workflows:
1. Unshaded materials: can either use `PREMUL_ALPHA_FACTOR` or can use a premultiplied texture in `ALBEDO`
2. shaded materials: Need to use `PREMUL_ALPHA_FACTOR`

drivers/gles3/rasterizer_scene_gles3.cpp
```diff
@@ -2690,6 +2690,11 @@ void RasterizerSceneGLES3::_render_list_template(RenderListParameters *p_params,
 								glBlendFuncSeparate(GL_DST_COLOR, GL_ZERO, GL_ZERO, GL_ONE);
 							}
 
+						} break;
+						case GLES3::SceneShaderData::BLEND_MODE_PREMUL_ALPHA: {
```
clayjohn, line 2694:
```suggestion
						case GLES3::SceneShaderData::BLEND_MODE_PREMULT_ALPHA: {
```
I prefer PREMUL (without the "t"), but CanvasItems use PREMULT, so we should use that everywhere :/

scene/resources/material.cpp
```diff
@@ -1511,6 +1514,11 @@ void BaseMaterial3D::_update_shader() {
 			case BLEND_MODE_MUL: {
 				code += "	vec3 detail = mix(ALBEDO.rgb,ALBEDO.rgb*detail_tex.rgb,detail_tex.a);\n";
 			} break;
+			case BLEND_MODE_LT_ALPHA: {
```
clayjohn, line 1517:
```suggestion
			case BLEND_MODE_PREMULT_ALPHA: {
```

servers/rendering/renderer_rd/forward_clustered/scene_shader_forward_clustered.cpp
```diff
@@ -90,6 +90,7 @@ void SceneShaderForwardClustered::ShaderData::set_code(const String &p_code) {
 	actions.render_mode_values["blend_mix"] = Pair<int *, int>(&blend_mode, BLEND_MODE_MIX);
 	actions.render_mode_values["blend_sub"] = Pair<int *, int>(&blend_mode, BLEND_MODE_SUB);
 	actions.render_mode_values["blend_mul"] = Pair<int *, int>(&blend_mode, BLEND_MODE_MUL);
+	actions.render_mode_values["blend_premul_alpha"] = Pair<int *, int>(&blend_mode, BLEND_MODE_PREMUL_ALPHA);
```
clayjohn, line 93:
```suggestion
	actions.render_mode_values["blend_premul_alpha"] = Pair<int *, int>(&blend_mode, BLEND_MODE_PREMULT_ALPHA);
```

drivers/gles3/storage/material_storage.cpp
```diff
@@ -2896,6 +2898,7 @@ void SceneShaderData::set_code(const String &p_code) {
 	actions.render_mode_values["blend_mix"] = Pair<int *, int>(&blend_modei, BLEND_MODE_MIX);
 	actions.render_mode_values["blend_sub"] = Pair<int *, int>(&blend_modei, BLEND_MODE_SUB);
 	actions.render_mode_values["blend_mul"] = Pair<int *, int>(&blend_modei, BLEND_MODE_MUL);
+	actions.render_mode_values["blend_premul_alpha"] = Pair<int *, int>(&blend_modei, BLEND_MODE_PREMUL_ALPHA);
```
clayjohn, line 2901:
```suggestion
	actions.render_mode_values["blend_premul_alpha"] = Pair<int *, int>(&blend_modei, BLEND_MODE_PREMULT_ALPHA);
```

drivers/gles3/storage/material_storage.h
```diff
@@ -245,6 +245,7 @@ struct SceneShaderData : public ShaderData {
 		BLEND_MODE_ADD,
 		BLEND_MODE_SUB,
 		BLEND_MODE_MUL,
+		BLEND_MODE_PREMUL_ALPHA,
```
clayjohn, line 248:
```suggestion
		BLEND_MODE_PREMULT_ALPHA,
```

servers/rendering/renderer_rd/forward_clustered/scene_shader_forward_clustered.h
```diff
@@ -106,7 +106,8 @@ class SceneShaderForwardClustered {
 			BLEND_MODE_ADD,
 			BLEND_MODE_SUB,
 			BLEND_MODE_MUL,
-			BLEND_MODE_ALPHA_TO_COVERAGE
+			BLEND_MODE_ALPHA_TO_COVERAGE,
+			BLEND_MODE_PREMUL_ALPHA,
```
clayjohn, line 110:
```suggestion
			BLEND_MODE_PREMULT_ALPHA,
```

servers/rendering/renderer_rd/forward_clustered/scene_shader_forward_clustered.cpp
```diff
@@ -244,7 +245,17 @@ void SceneShaderForwardClustered::ShaderData::set_code(const String &p_code) {
 			blend_attachment.dst_color_blend_factor = RD::BLEND_FACTOR_ONE_MINUS_SRC_ALPHA;
 			blend_attachment.src_alpha_blend_factor = RD::BLEND_FACTOR_ONE;
 			blend_attachment.dst_alpha_blend_factor = RD::BLEND_FACTOR_ZERO;
-		}
+		} break;
+		case BLEND_MODE_PREMUL_ALPHA: {
```
clayjohn, line 249:
```suggestion
		case BLEND_MODE_PREMULT_ALPHA: {
```

servers/rendering/shader_compiler.cpp
```diff
@@ -965,6 +965,10 @@ String ShaderCompiler::_dump_node_code(const SL::Node *p_node, int p_level, Gene
 				}
 			}
 
+			if (vnode->name == premul_alpha_name) {
+				r_gen_code.uses_premul_alpha = true;
+			}
+
```
clayjohn, line 971: This can be deleted as `uses_premul_alpha` never gets used

servers/rendering/shader_compiler.cpp
```diff
@@ -1564,6 +1568,7 @@ void ShaderCompiler::initialize(DefaultIdentifierActions p_actions) {
 	actions = p_actions;
 
 	time_name = "TIME";
+	premul_alpha_name = "PREMUL_ALPHA";
```
clayjohn, line 1571: Can be removed

servers/rendering/shader_compiler.h
```diff
@@ -84,6 +84,7 @@ class ShaderCompiler {
 		bool uses_screen_texture;
 		bool uses_depth_texture;
 		bool uses_normal_roughness_texture;
+		bool uses_premul_alpha;
```
clayjohn, line 87: can be removed

servers/rendering/shader_compiler.h
```diff
@@ -115,6 +116,7 @@ class ShaderCompiler {
 	const ShaderLanguage::FunctionNode *function = nullptr;
 	StringName current_func_name;
 	StringName time_name;
+	StringName premul_alpha_name;
```
clayjohn, line 119: Can be removed

clayjohn (commented):
Great work! I left a few comments to help clean this up. Most of them are just enforcing a consistent naming of `BLEND_MODE_PREMULT_ALPHA`

drivers/gles3/rasterizer_scene_gles3.cpp
QbieShay: Yeah :/ i updated it but i missed some.

mikejohnstn:
Any idea if `PREMUL_ALPHA` support is still targeting 4.3? I'm building a tool that depends on it, so it would help to know if I can aim to target a particular release.

clayjohn (dismissed):
Looks great!

AThousandShips (approved):

akien-mga:
This PR uses a mix of `PREMULT_ALPHA` and `premul_alpha` spelling (note the presence or absence of T). This should be unified to use only one spelling.

akien-mga:
Thanks!

scene/resources/material.cpp
```diff
@@ -688,6 +688,9 @@ void BaseMaterial3D::_update_shader() {
 		case BLEND_MODE_MUL:
 			code += "blend_mul";
 			break;
+		case BLEND_MODE_PREMULT_ALPHA:
+			code += "blend_premul_alpha";
```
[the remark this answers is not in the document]
akien-mga: So it seems like we had this `BLEND_MODE_PREMULT_ALPHA` but `blend_premul_alpha` inconsistency in CanvasItem shaders already.

Should we fix it in Spatial shaders and accept that we have an inconsistency in 2D `blend_premul_alpha` / 3D `blend_premult_alpha`, or should we do like in 2D?

doc/classes/BaseMaterial3D.xml
```diff
@@ -609,6 +609,9 @@
 		<constant name="BLEND_MODE_MUL" value="3" enum="BlendMode">
 			The color of the object is multiplied by the background.
 		</constant>
+		<constant name="BLEND_MODE_PREMULT_ALPHA" value="4" enum="BlendMode">
+			The color of the object is added to the background and the alpha channel is used to mask out the background. This is effectively a hybrid of the blend mix and add modes, useful for FX like fire where you want the flame to add but the smoke to mix. By default, this works with unshaded materials using premultiplied textures. For shaded materials, use the PREMUL_ALPHA_FACTOR built-in so that lighting can be modulated as well.
```
akien-mga, line 613: Should be `[code]PREMULT_ALPHA_FACTOR[/code]`.

scene/resources/material.cpp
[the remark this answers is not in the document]
QbieShay: I noticed that too, but I'd rather not start touching things elsewhere for now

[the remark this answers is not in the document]
akien-mga: I mean this one is a bug, it's writing `blend_premul_alpha` in Spatial shaders but what's expected is `blend_premult_alpha`.

So we have two options:
- Fix this one up to be `blend_premult_alpha`, and accept that 3D shaders will use a different keyword from 2D shaders
- Go back to the previous version of your PR before my review (I didn't know you just replicated the inconsistency of existing 2D keywords) and use `blend_premul_alpha` in 3D too.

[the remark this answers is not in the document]
QbieShay: I didn't remember either, and frankly I wonder if I did it on purpose or not. I'm not sure if we want to keep consistent with 2d and generally inconsistent, or if we should have correct 3d and weird 2D. Is there any precedent we can reference to take the decision?

jitspoe:
Does this PR address fog making stuff light up that should be transparent?  That was one thing I never got around to fixing on mine.

scene/resources/material.cpp
```diff
@@ -1819,6 +1822,11 @@ void fragment() {)";
 	vec3 detail = mix(ALBEDO.rgb, ALBEDO.rgb * detail_tex.rgb, detail_tex.a);
 )";
 			} break;
+			case BLEND_MODE_PREMULT_ALPHA: {
+				// This is unlikely to ever be used for detail textures, and in order for it to function in the editor, another bit must be used in MaterialKey,
+				// but there are only 5 bits left, so I'm going to leave this disabled unless it's actually requested.
+				//code += "\tvec3 detail = (1.0-detail_tex.a)*ALBEDO.rgb+detail_tex.rgb;\n";
```
Macksaur, line 1828: @QbieShay I don't understand why this is commented out. Before the nuance is lost to the sands of time can you expand a little more? 5 bits left where?

https://github.com/godotengine/godot/blob/f91db3dc58f1d6a0cb63d591515183b5a45cd3ba/scene/resources/material.h#L326

The bits as far as I can tell are already spent allocating `BLEND_MODE_PREMULT_ALPHA`. Commenting it out just seems a little strange.

jitspoe: Ah, that was probably taken from my original PR which didn't have the bits taken.  I was hesitant to waste a bit for this, but if the bit is already taken, this should be uncommented.

jitspoe:
Ok, just cherry-picked this and tested and verified that this does NOT properly address fog, at least independently.  I did notice there were some other fog related things in the shader file that changed, but I don't think they're related.  Guess that should still be addressed.

To better explain, I'm using flame sprites with premult alpha, and if I have fog enabled, the whole quad starts becoming visible, (additively blended with the fog color).  I thought maybe the premult_alpha_factor would somehow get used, but looking closer, I'm not really sure what premult_alpha_factor even does, as it seems to just be renamed to premult_alpha, which is a bit confusing.

QbieShay:
@jitspoe it's multiplied at the very end of the shader. Do you mind supplying a test scene with your setup (via DM is fine too) so that u can try and fix the dog issue?

clayjohn:
> Ok, just cherry-picked this and tested and verified that this does NOT properly address fog, at least independently. I did notice there were some other fog related things in the shader file that changed, but I don't think they're related. Guess that should still be addressed.
> 
> To better explain, I'm using flame sprites with premult alpha, and if I have fog enabled, the whole quad starts becoming visible, (additively blended with the fog color). I thought maybe the premult_alpha_factor would somehow get used, but looking closer, I'm not really sure what premult_alpha_factor even does, as it seems to just be renamed to premult_alpha, which is a bit confusing.

Can you share your shader? Are you writing to `PREMUL_ALPHA_FACTOR`?

akien-mga:
Make sure to also cherry-pick #91399 if testing on a branch different from upstream `master`, as this PR wouldn't be functional without that fixup.

QbieShay:
I havent had a chance to boot my dev PC yet but I think I understand the issue, I'll debug it later. If it is what I imagine, I need to multiply fog by alpha if render mode pmul alpha is on.

clayjohn:
> I havent had a chance to boot my dev PC yet but I think I understand the issue, I'll debug it later. If it is what I imagine, I need to multiply fog by alpha if render mode pmul alpha is on.

Pmul alpha is already multiplied after fog though. I don't see why you would need to multiply it twice

QbieShay:
Because the fog should be applied only to the non transparent parts of a quad. Alpha zero, factor non zero = additive.

QbieShay:
@jitspoe Please test against latest master, I can't reproduce. Don't forget to use the PREMUL_ALPHA_FACTOR

jitspoe:
So it seems like PREMUL_ALPHA_FACTOR just scales the albedo color value, effectively turning it into a regular alpha, kind of defeating the purpose of the premultiplied aspect.  Fixes the fog, but the whole point is to be able to have additive things like flames mixed in with alpha masked stuff like smoke.  Unless I'm just not understanding how this is intended to be used?  Still trying to understand the reasoning behind PREMUL_ALPHA_FACTOR.  How is it different from just scaling ALBEDO?

Building latest main now to do further tests.

jitspoe:
I'm moving the discussion to https://github.com/godotengine/godot-proposals/issues/3431#issuecomment-2116564276 so it's in one cohesive location, since there is a lot of other discussion in there as well.

xa8et67:
太棒了，多年来终于解决了这个问题。

PR #85624, opened 2023-12-01
Add `TextControl` to unify code of `TextEdit` and `LineEdit`

TheSofox:
Fixes #85511

Added TextControl as the parent of TextEdit and TextLine to consolidate code all the code between them, starting with the parts in listed in #85511 

There's a lot more code that could be consolidated, but started with the sections listed in the bug report, just to test whether this is a viable solution.

### DIFF
--- a/scene/gui/text_control.h
+++ b/scene/gui/text_control.h
@@ -1,0 +1,92 @@
+/**************************************************************************/
+/*  text_control.h                                                        */
+/**************************************************************************/
+/*                         This file is part of:                          */
+/*                             GODOT ENGINE                               */
+/*                        https://godotengine.org                         */
+/**************************************************************************/
+/* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+/* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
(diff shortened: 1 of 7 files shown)


AThousandShips:
I'm not sure about the usefulness of changing the inheritance for a single method, which as far as I know breaks compatibility as it rearranges the inheritance, but more importantly moves the enum which breaks compatibility as well

I think a less disruptive change is warranted here unless more shared code is unified

TheSofox:
> I'm not sure about the usefulness of changing the inheritance for a single method,
There are other methods, such as menu_option(int p_option), various selection methods (eg is_selecting_enabled()). and a lot of other methods/variables. where similar things are done but worded differently. Reducing this repetition would need to be an ongoing process with a bit being done at a time. 

>but more importantly moves the enum which breaks compatibility as well
Conceptually, the bug requires this break to be resolved. The entire issue is that there are two identical enums in separate classes. Either one is going to be removed or both are going to be merged in a separate place. Either way, at least one enum gets moved.

 
> I think a less disruptive change is warranted here unless more shared code is unified
I'm open to suggestions. 

> Further the new class should be abstract as you shouldn't be able to instance it as it's incomplete
Makes sense.

KoBeWi:
tbh this is the least breaking change regarding the enums, because you can still use them directly in both classes and they retain their values.

scene/gui/text_control.h
```diff
@@ -0,0 +1,92 @@
...
+public:
+	// Context menu.
+	PopupMenu *menu = nullptr;
+	PopupMenu *menu_dir = nullptr;
+	PopupMenu *menu_ctl = nullptr;
+
+	virtual void menu_option(int p_option);
+};
+
+VARIANT_ENUM_CAST(TextControl::MenuItems);
```
AThousandShips, line 90: You need to move registering the enum to this class, with `_bind_methods` and extracting the binding from the two other classes

AThousandShips (changes requested):

TheSofox:
> It presumes that the change is needed in this form without adding all the shared code, if all the shared code is added in one PR that change makes sense, but otherwise it's assuming we need this compatibility breakage here instead of doing it some other way yet locking us into this change

I get what you're saying. You want the PR to do it all in one go so it can be evaluated on its own merits.

I'm gonna have to give this more thought. In the meantime I don't mind if anyone else wants to try their own approach or work off of what I've done.

AThousandShips:
For the menu one option would be to create a dedicated popup menu to share between these two, and provide it a method to call, similar to how `ViewPanner` works

TheSofox:
That's actually the approach I'm looking into right now, I'm glad you think it would be a good solution to. Also thanks for mentioning ViewPanner, I didn't know about it before and it'll be very useful having a comparison.

YuriSizov:
The linked issue already covers the question of whether it would be rational to try and unify these controls. So I think going into this direction would be counter-productive. At the very least, it's a controversial and contested idea, and so doing a breaking change over it would be hard to justify.

There are shared bits of code, and I believe most of that duplication can be reduced without affecting user-facing facilities. I would suggest efforts to be put towards this solution.

MewPurPur:
I've noticed that LineEdit does a lot of things poorly compared to TextEdit and thought a lot about this, but I think these two classes are super different in everything aside from the part where you edit text. I've thought about having some kind of unexposed API for text editing operations that all these classes call to for basic things like text selections, drawing carets, and undo/redo. I'm not sure that's a good idea either, because it would make LineEdit a 1D wrapper dealing with unnecessarily complicated logic for 2D text, but I don't think it's a bad idea either as it would largely actually guarantee that there are no differences between the classes.

PR #50404, opened 2021-07-13
Clean convex hull decomposition code

pouleyKetchoupp:
Following-up from https://github.com/godotengine/godot/pull/50262#issuecomment-878507055 (CC @lawnjelly)

Remove unnecessary conversion between triangle data and vertex data whenever possible.

After testing performance, it doesn't seem to make much difference because the time spent in preparing and finalizing data is negligible compared to the process in VHACD, but the code generally seems cleaner this way.

PR #85631, opened 2023-12-01
Fix memory corruption and assert failures in convex decomposition

### LINKED ISSUE #85439: Crash when configuring the import of a gltf file and enabling Physics in a MeshInstance3D and switching body and shape types
### Godot version

v4.2.rc2.official [1ba920fad]

### System information

Windows 11, Intel Core i5 10300H, nVidia GTX 1650

### Issue description

When configuring the import of a gltf file, I'm trying to configure a mesh to use physics. When switching the body and shape type of this mesh, the engine will sometimes crash.

I couldn't detect the specific action that triggered the crash. Switching possible shape types appears to make the crash happen more often, but sometimes it will crash on the first shape type I choose and sometimes I can switch the shape type several times before it crashes.

The bug also happens in 4.2_RC1, but it's WAY easier to reproduce there. Sometimes merely checking Physics on will already trigger the crash. At most, doing one switch of body type (not even shape type) will also trigger the crash.

### Steps to reproduce

- Open the sample project
- Double click `room_model.glb`
- Click the `ComputerScreen` mesh instance
- Check `Physics` on
- Change body type to Area
- Change shape types

It can take a few changes to shape type to trigger the bug in RC2, but in RC1 sometimes merely checking Physics on can trigger it.

Project to reproduce the bug.

[bug_project.zip](https://github.com/godotengine/godot/files/13477636/bug_project.zip)

Video of bug happening on my machine (I was "lucky" and it happened right away, sometimes it takes  a few switches to shape type before it happens).

https://github.com/godotengine/godot/assets/1204296/797262fd-a881-4b8d-a6f8-2ed6aa593995

Malcolmnixon:
This PR fixes [issue 85439](https://github.com/godotengine/godot/issues/85439) by changing the index pre-increment to post-increment so index 0 is not skipped and preventing a potential overrun past the end of the vertices vector.

Testing involved constructing a GLB file consisting of a single triangle:
image

Before this change the code allocates vertices and indices arrays of size 3 and writes the following:
| Vertices (size=3) | Indices (size=3) |
| ---- | ---- |
| 0 = unused | 1 |
| 1 = [-1,-1,-1] | 2 |
| 2= [1,-1,-1] | 3 |
| 3 = [0,1,-1] (BUFFER OVERRUN) | unused |

After this change the code writes the following:
| Vertices (size=3) | Indices (size=3) |
| ---- | ---- |
| 0 = [-1,-1,-1] | 0 |
| 1 = [1,-1,-1] | 1 |
| 2= [0,1,-1] | 2 |

The collision shape with this new data appears correct:
image

[offset_triangle.zip](https://github.com/godotengine/godot/files/13532122/offset_triangle.zip)

### DIFF
--- a/scene/resources/importer_mesh.cpp
+++ b/scene/resources/importer_mesh.cpp
@@ -969,7 +969,7 @@ Vector<Ref<Shape3D>> ImporterMesh::convex_decompose(const Ref<MeshConvexDecompos
 				if (found_vertex) {
 					index = found_vertex->value;
 				} else {
-					index = ++vertex_count;
+					index = vertex_count++;
 					vertex_map[vertex] = index;
 					vertex_w[index] = vertex;
 				}


BastiaanOlij (approved):
Makes sense to me, would like the original author to weigh in if there was a reason for this. This is not something people tent to do by accident, most likely there was other code here before that makes this make sense..

Malcolmnixon:
> Makes sense to me, would like the original author to weigh in if there was a reason for this. This is not something people tent to do by accident, most likely there was other code here before that makes this make sense..

It looks to have been introduced with https://github.com/godotengine/godot/commit/2ca94e51e4d4477709b83b487a4f5b938f24f56b and no code before it seems to implement the vertex-deduplication map trick. The older code uses duplicated vertices, but the loop doesn't seem to suffer from N+1 overruns.

lawnjelly (approved):
Good spot, must have missed this when we looked at #50404 .

lawnjelly:
> Makes sense to me, would like the original author to weigh in if there was a reason for this. This is not something people tent to do by accident, most likely there was other code here before that makes this make sense..

Pouley is probably unlikely to reply as he's moved on to other work, but this was implementing something I suggested about removing unnecessary conversion between longhand vertex format (with duplicated verts) and vertex + index format, if I remember right. Nothing super complex going on here.

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.1.

YuriSizov:
@yythlj Please open an issue with more details about your case.

YuriSizov:
Cherry-picked for 4.1.4.

PR #77284, opened 2023-05-20
LightmapGI: Implement shadowmask for DirectionalLight

DearthDev:
Implements shadowmasking as proposed in godotengine/godot-proposals#2354.
See also #68702, #51330

Shadowmasking can be toggled on and off in the lightmap settings and is disabled by default.
Directional lights with their bake mode set to dynamic will bake their shadows into the alpha channel of the lightmap. Then, dynamic shadows will be used up close and baked shadows will be use outside the light's shadow max distance. If more than one directional light with its baked mode set to dynamic exists, then the shadowmask will only be present in areas where their shadows overlap. If no directional light with its baked mode set to dynamic exists, shadowmasking will be disabled and a warning printed saying such.

Note: #55867 has made testing with the mobile backend cumbersome.

- *Bugsquad edit: This closes godotengine/godot-proposals#2354.*

Issue #84846, opened 2023-11-13
Vulkan Mobile: LightmapGI dynamic object lighting results in black objects


SpockBauru:
### Godot version

v4.1.3.stable.official [f06b6836a] and v4.2.beta5.official [4c96e9676]

### System information

Godot v4.1.3.stable and Godot v4.2.beta5 - Windows 10.0.22621 - Vulkan (Mobile) - dedicated NVIDIA GeForce RTX 3080 (NVIDIA; 31.0.15.2849) - 12th Gen Intel(R) Core(TM) i7-12700KF (20 Threads)

### Issue description

Hey guys, Seems like Lightmap**Probes** are not working on **Mobile** renderer. They work well on **Forward+** but **not in Mobile**. 

https://github.com/godotengine/godot/assets/67930504/b3c681fa-061c-4e2e-bd4d-8b5b22cc4264

I've already:
- Tested on both Godot 4.1.3 and 4.2 beta 5.
  - Light**maps** don't work on Godot 4.1.2 on Mobile Renderer, but works in 4.2 beta 5. Light Probes don't work with Mobile Renderer on both.
- Set up lightmap UVs.
- Marked static and dynamic objects and lights.
- Set up all kinds of World Environment settings, even tried to delete and disable it.
  - In this node, the Background mode `Sky` have issues on Mobile, so I set up to `Custom Color` (seems related to https://github.com/godotengine/godot/issues/55868#issue-1077887433)
- Baked LightmapGI on both Forward+ and Mobile:
- Tried many combinations on the LightmapGI node.
- Tried a brand new project, now even the Lightmaps stopped working on Godot 4.1.2 (Still working on 4.2 beta 5)...
- Yes, I restarted my PC.
- Deleted all Godot config files in `AppData\Local` and `AppData\Roaming`

### Steps to reproduce

Bake a scene with LightmapGI and an object that transition between different lighting conditions. The Probes will work on Forward+ but not on Mobile.

### Minimal reproduction project

[LightmapProbes_Test.zip](https://github.com/godotengine/godot/files/13335324/LightmapProbes_Test.zip)

Edit: MRP for Godot 4.3.dev6
[Lightmap_Test_4.3.dev6.zip](https://github.com/godotengine/godot/files/15354314/Lightmap_Test_4.3.dev6.zip)

KarthikHebbarMN:
+1

Lasuch69:
[...]

PR #77912, opened 2023-06-06
Reimplement support for bicubic sampling in LightmapGI

Calinou:
This brings it to feature parity with 3.x's lightmapping implementation.

Only supported in Forward+, not Forward Mobile as the push constant for instances is already full (128 bytes).

I won't have much time to work on this, so please look into continuing this in your own branch if possible. I don't even know if the way of passing the lightmap texture size is correct – as I understand it, it needs to be per-instance as multiple LightmapGIs can be used in the same scene at the same time. `3.x` used a uniform here, but we don't seem to use uniforms anymore in the 4.x scene shader.

- This closes https://github.com/godotengine/godot/issues/49935. See also https://github.com/godotengine/godot/pull/77284, which benefits from bicubic sampling.

## TODO

- [ ] Make it work correctly. Currently, it breaks scene rendering even though the shader compiles successfully:

image

- [ ] Add a project setting to disable bicubic sampling for low-end systems.

Issue #63610, opened 2022-07-29
Angular distance breaks shadow with long distance or large shadow casters


atirut-w:
### Godot version

v4.0.alpha12.official [2c11e6d9e]

### System information

Fedora Linux 36 (Workstation Edition), Vulkan Clustered, GTX 1660

### Issue description

When soft shadows are being cast by objects that are thin and have lots of distance, they seem to break.

Tested with [House and Rungholt mesh](https://casual-effects.com/g3d/data10/index.html#mesh27) from McGuire CG Archive.

https://user-images.githubusercontent.com/25323231/181693286-1eadffa1-d775-455c-a86a-885315d7a620.mp4

Also with a simple stretched cube

https://user-images.githubusercontent.com/25323231/181693586-45b2e42e-d69d-42f1-ba2e-0e8057400122.mp4

### Steps to reproduce

1. Shadow caster must have sufficiently thin shadows and far away from where the shadow lands
2. DirectionalLight angular distance > 0

BraqueDown:
Related/Duplicate: #60391

atirut-w:
This seems related indeed, but might not be a duplicate. The models I used from the archive are heavily tessellated.

![Screenshot from 2022-07-29 14-34-57](https://user-images.githubusercontent.com/25323231/181708577-c38ce7c0-1d74-4f1f-a2a6-e92d565e29c3.png)
![Screenshot from 2022-07-29 14-33-41](https://user-images.githubusercontent.com/25323231/181708588-964a912c-af1b-400f-81f6-e7f190e71706.png)

BraqueDown:
I just downloaded the scene myself, and yes that boat is very well tessellated. Definitely related though.

Either way this issue can be reigned in by increasing **Pancake Size**:
![image](https://user-images.githubusercontent.com/43602802/181708834-1cfaa615-6063-4ca0-9909-362e061b7ba8.png)
![image](https://user-images.githubusercontent.com/43602802/181708922-a0ab01ef-1a36-408c-9a2c-8cfe4016be97.png)

BraqueDown:
Playing the Pancake Size slider has interesting results. Increasing causes the penumbra to apparently turn itself inside out:

https://user-images.githubusercontent.com/43602802/181711707-d4c66693-af6a-4668-bf16-147c9d166a07.mp4

Jamsers:
[...]

PR #85653, opened 2023-12-02
Implement LightmapGI shadowmasks

### LINKED ISSUE #2354: Implement support for subtractive shadowmapping or shadowmasking in DirectionalLight (for use with BakedLightmap)
*See also discussion in https://github.com/godotengine/godot/issues/46332.*

**Edit:** Going for shadowmasking instead of subtractive shadowmapping may be a better idea in the end, as it's easier to use and more flexible. See https://github.com/godotengine/godot-proposals/issues/2354#issuecomment-891389090.

The Godot editor 🙂 

When targeting low-end desktop hardware or mobile hardware, baked lightmaps are often used but combining them with real-time shadows is difficult.

While it's possible to use a DirectionalLight with the bake mode set to **Indirect** and point lights' bake mode set to **All**, there are several downsides to this:

- Distant objects won't cast any shadows, which can look bad with large structures. You can increase the DirectionalLight shadow **Max Distance** property to help with this, but this will make nearby shadows blurrier (including those from dynamic objects).
- Performance will be lower than the proposed subtractive mode, since the whole lightmapped level will cast shadows. This increases the number of vertices and draw calls in the DirectionalLight shadow cascades.

The indirect DirectionalLight approach described above is suited for many types of games, but when you want to display relatively large worlds on mid-range hardware, you may need a more oldschool solution as proposed below.

Mixing baked and real-time lighting is fundamentally a difficult problem to solve with no perfect solutions. One way to handle this is to use *subtractive* shadow maps that darken the underlying surfaces to a predefined color.

Subtractive shadowmapping isn't very realistic, but it's cheap enough to be usable on mobile platforms nowadays. On desktop platforms, its low cost will also allow players to enable more effects on low-end/mid-range hardware compared to other lighting solutions. On top of that, since subtractive shadowmapping still uses fully-baked shadows for lightmapped meshes, shadows in the distance will *never* fade away or exhibit seams between splits. Only dynamic objects will lose their shadow in the distance, but this is far less problematic than having both static and dynamic objects lose their shadows.

A subtractive shadowmapping option means that the DirectionalLight won't emit any light and will only cast a shadow. As you might expect, this isn't going to work with standard light equations 🙂 

The blending formula needs to be made toggleable with a DirectionalLight property, and will require modifying the shadow blending equations in [`drivers/gles3/shaders/scene.glsl`](https://github.com/godotengine/godot/blob/3.2/drivers/gles3/shaders/scene.glsl). The shadow color can be defined as a project setting, since we don't want to introduce per-light shadow colors in 4.0 for performance reasons. It should be configured by the user to *roughly* match the color of baked shadows.

For proper appearance, lightmapped meshes will also have to have their GeometryInstance shadow mode set to **Off** after baking the lightmap. When using subtractive shadowmapping, only dynamic objects should cast shadows since static shadows are already handled by the lightmap itself. Currently, this can be done *after* baking lightmaps but you need to revert it to **On** every time you wish the bake the lightmap, which is inconvenient. We'll have to figure out how to do this automatically.

For reference, [Unity provides a subtractive lighting mode](https://docs.unity3d.com/Manual/LightMode-Mixed-Subtractive.html). Here's a [video](https://www.youtube.com/watch?v=sQPa8sLtICs) comparing all of Unity's baked lighting modes. Unity also provides a [shadowmask](https://docs.unity3d.com/Manual/LightMode-Mixed-Shadowmask.html) mode in its HDRP rendering pipeline, but it's more demanding on the hardware and significantly more complex to implement since it requires two sets of lightmaps and light probes. Therefore, I consider implementing shadowmask to be outside the scope of this proposal.
**Edit:** There's a tutorial on implementing shadow masking in Unity SRP which is a good read: https://catlikecoding.com/unity/tutorials/custom-srp/shadow-masks/

I started looking at implementing subtractive lighting in DirectionalLight but haven't succeeded yet: https://github.com/Calinou/godot/tree/add-directionallight-subtractive-lighting

In practice, subtractive shadowmapping could look something like this: https://www.youtube.com/watch?v=lQMCy328avE
Notice how the character's real-time shadow is configured to roughly match the color of the baked shadows, yet it never darkens surfaces that are already darkened by baked shadows. To improve the appearance of dynamic objects in interiors, a subtle blob shadow can be used on top of the dynamic shadow (since the dynamic shadow only appears in exteriors).

Note that this shadowmapping mode only really makes sense for DirectionalLight, since its behavior is most likely not very suited to OmniLights and SpotLights. Also, this proposal refers to GLES3 but it's likely implementable in GLES2 and Vulkan as well.

No, as shadowmapping appearance is handled by the core renderer.

See above.

BlueCube3310:
Salvages #77284
Closes: https://github.com/godotengine/godot-proposals/issues/2354

Adds support for baked shadowmasks to LightmapGI. The shadowmasks are saved as compressed images and applied on top of the dynamic shadowmaps. 

bb

Progress:
- [x] Support baking shadowmasks,
- [x] Support Forward+ renderer,
- [x] Support Mobile renderer,
- [x] Support Compatibility renderer,
- [x] (Bugfix) Allow lightmaps without shadowmasks,
- [x] Dilate,
- [x] Fix blending artifacts, Edit: Caused by DXT1 Compression,
- [x] Document everything properly,
- [x] Clean up and optimize the code,

To consider:
- Consider using specialization constants,
- Consider blending seams.

To do in the future (important):
- ~~Compress grayscale textures as RGTC_R,~~ Edit: done
- ~~Implement and use bicubic sampling (See #89919)~~ Edit: done,
- Further optimizations and cleanup,
- Add a `shadowmask_resolution_ratio` property to scale shadowmasks independently from lightmaps.
- Remind the user when lighting needs to be rebuilt,


Calinou (commented):
Tested locally, it mostly works as expected (I tested Forward+ and Mobile).

**Testing project:** [test_lightmapgi_mobile.zip](https://github.com/godotengine/godot/files/13591806/test_lightmapgi_mobile.zip)

The way shadowmask interacts with real-time shadows looks strange. I don't recall this happening with previous PRs, but maybe I started noticing this issue just now. This is most noticeable with **Directional Soft Shadow Filter Quality** set to **Ultra**, but it occurs with any setting:

image

I'd expect the final shadow color to be `min(shadowmask, real_time_shadowmap)`. Right now, it seems to be `real_time_shadowmap > 0.0 ? real_time_shadowmap : shadowmask`. This issue affects both Forward+ and Mobile.

If you set **Directional Soft Shadow Filter Quality** to **Hard**, notice how shadowed area transitions look much sharper than intended:

image


Regarding fixing seams, I'd say it's worth the effort as right now, seams can be quite noticeable in the distance:

Real-time shadow | Shadowmask
-|-
Screenshot_20231206_222718 | Screenshot_20231206_222726

modules/lightmapper_rd/lm_compute.glsl
```diff
@@ -347,7 +349,7 @@ float get_omni_attenuation(float distance, float inv_range, float decay) {
 	return nd * pow(max(distance, 0.0001), -decay);
 }
 
-void trace_direct_light(vec3 p_position, vec3 p_normal, uint p_light_index, bool p_soft_shadowing, out vec3 r_light, out vec3 r_light_dir, inout uint r_noise) {
+void trace_direct_light(vec3 p_position, vec3 p_normal, uint p_light_index, bool p_soft_shadowing, out vec3 r_light, out vec3 r_light_dir, out float r_shadow, inout uint r_noise) {
```
bitsawer, line 352: `r_shadow` is an `out` parameter, but the function can return early without writing to it, in which case its value will be undefined when used later in the calling function. It should have a valid value in all code paths, for example `r_light` initializes its default value at the start of the function to make sure it always has a valid value.

YuriSizov:

BlueCube3310:
> The way shadowmask interacts with real-time shadows looks strange. I don't recall this happening with previous PRs, but maybe I started noticing this issue just now. This is most noticeable with **Directional Soft Shadow Filter Quality** set to **Ultra**, but it occurs with any setting:
> 
> Regarding fixing seams, I'd say it's worth the effort as right now, seams can be quite noticeable in the distance:

Both of these issues should be fixed now.

Calinou:
> Both of these issues should be fixed now.

It works great now, thanks :slightly_smiling_face:

atirut-w:
Having combined realtime and static shadows probably isn't a very good idea especially with automatically wrapped UV2s:

<details><summary>Image</summary>
<p>

image

</p>
</details> 

The quality becomes even worse if your light have it's size or angular dist set to more than 0:

<details><summary>Image</summary>
<p>

image

</p>
</details> 

On another note, I have a version of GDQuest's TPS demo with lightmaps and this PR breaks shadow on everything for some reason. I yet to encounter this on other projects.

atirut-w:
Also, could this method be adapted to static lights to give them specular reflections?

Calinou:
> Having combined realtime and static shadows probably isn't a very good idea especially with automatically wrapped UV2s. It should definitely be opt-in.

Once [bicubic sampling is reimplemented](https://github.com/godotengine/godot/pull/77912), this will be much less of an issue as the shadowmask will be much smoother.

atirut-w:
> Storing light direction requires much more memory than shadow information, which fits in a R8 texture.

A sort of semi-realtime mode, maybe? Baked shadows with realtime specular reflections? It sounds nice to have, but I'm not sure how practical that would be.

atirut-w:
Feedback/issue: light size/angular distance is not being taken into account

Also, if done right, this PR could help make #63610 less impacting.

Calinou:
> Feedback/issue: light size/angular distance is not being taken into account

The GPU lightmapper takes these parameters into account since https://github.com/godotengine/godot/pull/62054. When I tested this PR, it worked correctly on shadowmasks too.

atirut-w:
This PR apparently doesn't use the same code as #62054. This is what a directional light with angular dist of 5 looks like:

image

BlueCube3310:
> Old version
> 
> Oops, yeah. My testing scenarios didn't try to exaggerate the soft shadows enough. By the way, are shadowmaps not denoised?
> 
> 
> Addendum: will this PR get support for light types other than directional?

Rebased, now shadows are denoised + other light types can be baked, although that feature is very experimantal.

Calinou:
The denoising seems to work well from my testing.

Baking omnilights into a shadowmask seems broken currently:

image

Compare this to when you get closer and real-time shadows appear:

image

Remember that baking more than one light in a shadowmask requires switching the texture to RGBA8 (supports 4 intersecting lights at once), which will require more memory and storage. That's why I generally consider positional light shadowmasks to be not worth the trouble, especially since positional light shadows have infinite distance by default (unless you opt into distance fade).

WickedInsignia:
This feature works especially well when shadowmasks are never drawn over shadowmaps but instead replace them at a certain distance. This allows for more efficient shadowmap settings with a much lower max distance than would usually be necessary, while also providing rudimentary shadowing to faraway scenery. This is present as Distance Shadowmask in Unity and also exists in Unreal.
Insurgency Sandstorm uses this technique very effectively.
I'm fairly certain Counter Strike 2 overlaps its shadowmasks with shadowmaps, but use a fairly high-resolution bake to avoid too much of a discrepancy in quality.

Optimally users would have the option to either:
- Replace shadowmaps with shadowmask at a specific distance
- Overlap shadowmaps and shadowmasks together
- Use only shadowmasks if desired

BlueCube3310:
The different shadowmask blending modes are now available: None, Replace, Overlay and Only. Currently these properties can be tweaked by individual lightmap nodes (and therefore the instances they affect) in the scene, but this may change in the future.

SpockBauru:
Tested the windows artifact. Results are quite good!

Tested on an roughly 1km² map with lightmaps at 4096 resolution (about 4 samples per meter).
To amplify the effect, the shadow distance was reduced to only 20 meters.

 Here the test scene:
[Shadowmap_Test.zip](https://github.com/user-attachments/files/15539914/Shadowmap_Test.zip)
 
Shadowmap modes are working as expected:
Shadowmask_None
Shadowmask_Replace
Shadowmask_Overlay
Shadowmask_Only

Here a video comparison, since GitHub only allows up to 10MB the quality is reduced:


https://github.com/godotengine/godot/assets/67930504/2e811203-a220-4a47-a9de-b6330f6659e0

Finally, my classic tunnel comparison:

https://github.com/godotengine/godot/assets/67930504/ab2c0473-44b9-4aeb-8921-eadaf79505c0

Overall, everything seems to be working as expected. Excellent job everyone! \o/

atirut-w:
That's an awful amount of duped texts in the enum docs you suggested, imo.

BlueCube3310:
Added the docs, Compatibility should also work now.

WickedInsignia:
Works well on directional lights in my brief checks. A couple considerations:
- Spotlights and Omni lights are unaffected. It can be useful to replace these with Shadowmask entirely for performance reasons when shadows are needed for many static sources. Maybe a new bake mode for these called "Shadowmask and Indirect" could be added? Then whatever is set in the Lightmap node only affects Directional. Not sure about UX consistency with that so might require more thought/discussion.
- Shadowmask data is saved to a separate PNG correct? We might want to allow for separate texel density between Lightmap/Shadowmask (and Directional if possible). Bakery allows the user to define a fraction for each of these (so Directional maps can use 1/2 the texel density or less of the main Lightmap to save space). Adjusting Shadowmask resolution independent of the main Lightmap is important depending on how it's used (Only will require more resolution than Replace, for example).

SpockBauru:
Found some interesting things on compatibility render. This PR seems to be working correctly but there are additional issues that I believe are unrelated.

## This PR
Tested the windows artifact. Docs are a bit verbose but they are working:
image

Tested the 4 Shadowmask Modes (None, Replace, Overlay, Only). Every one seems to be working as expected. Good job!

## Additional Issues on Compatibility Renderer (Probably unrelated)
Found some additional issues regarding LightmapGI and Shadowmask on the Compatibility Renderer. I think they are not related to this PR but will let here for reference. Maybe they need their own issue but I'm not sure.

### (Docs) LightmapGI
On the Compatibility Renderer, LightmapGI Node shows a warning saying that Lightmaps cannot be baked on devices that not support **RD Backends**, which is confusing because most people (including me) have no idea what this means and since is a warning it leads to understand that the compatibility render will not bake at all. A bit of clarification is needed, or even better: Detect if these whatever backends are supported and not show a warning at all. (warnings are scary)
image

### (Bug) Shadowmaps
Distant hills that project shadows near the camera on other renderers does not show on compatibility renderer. Take a look on the mountain shadow at right:

Shadowmap_Bug


This seems to be just a detail on the picture, but while moving is pretty distracting.

https://github.com/godotengine/godot/assets/67930504/3b276ae8-1fd5-44ed-92ee-43e61c0f3c99

The same scene on Forward+ shows the entire shadow, and clips at the desired distance (20m):

image

Calinou:
> A bit of clarification is needed, or even better: Detect if these whatever backends are supported and not show a warning at all.

Unfortunately, performing this check is slow so we can't do it every time `_get_configuration_warnings()` is shown.

Personally, I'd get rid of the warning entirely, except on the web platform as it only supports Compatibility.

BlueCube3310:
Bicubic sampling and aliasing fixes are now supported after #89919 and #95828.

SpockBauru:
Made a quick test on the Windows fragment and overall got almost the same results from [my previews test](https://github.com/godotengine/godot/pull/85653#issuecomment-2146194388). The only visual difference is that edges are better filtered :D
shadowmask_replace

Also tested on Forward+, Mobile and Compatibility renderers. There are some issues, but they are not related to this PR and should have their own issue (some [already do](https://github.com/godotengine/godot/issues/84846)).

I believe that this PR is in a good shape and **can be merged as it is now**. Future improvements can be added later.

Excellent job everyone!

BlueCube3310:
I think this is mostly done, it just needs some optimizations + testing on different hardware (especially VR - multiview is probably broken on Compatibility, though I can't test it locally).

Calinou:
> especially VR - multiview is probably broken on Compatibility, though I can't test it locally

There's a Viewport property you can use to force XR-style rendering regardless of whether you have an HMD connected. Its variable rate shading property can also be set to XR for a similar effect (the VRS project setting affects the root viewport).

BlueCube3310:
The documentation is now slightly shorter, a configuration warning is shown when attempting to use a shadowmask without one existing, and some rendering fixes for Compatibility.

BlueCube3310:
The only big issue left is how shadowmasks should interact with directional lights. Currently when several lights are present, the shadowmask's opacity or visibility will depend on the last light's properties.

Calinou:
> The only big issue left is how shadowmasks should interact with directional lights. Currently when several lights are present, the shadowmask's opacity or visibility will depend on the last rendered light's properties.

I'd print a warning on bake if multiple directional lights are detected when shadowmasking is enabled, and print the node name of the light that was taken into account for the shadowmask.

BlueCube3310:
> > The only big issue left is how shadowmasks should interact with directional lights. Currently when several lights are present, the shadowmask's opacity or visibility will depend on the last rendered light's properties.
> 
> I'd print a warning on bake if multiple baked directional lights are detected when shadowmasking is enabled, and print the node name of the light that was taken into account for the shadowmask.
> 
> You could also try to pick the one that has the brightest effective energy (`(color * energy).get_v()`), since its shadows will be the most visible.

The first suitable light found is now used for the shadowmask.

BlueCube3310:
OIDN works correctly now, I think this PR is finally ready for review

Calinou (commented):

Some remaining issues I noticed:

- When you set the shadowmask mode to **None** and bake lightmaps again, the shadowmask source texture isn't removed from the filesystem (unlike what the description implies).

- In Compatibility, DirectionalLight3D's **Blend Splits** doesn't work correctly with the last split when the LightmapGI shadowmask mode is **Overlay**:

image

It works well in Forward+ and Mobile.

- With all rendering methods, dynamic object shadows are not rendered (even up close) when using the **Only** shadowmask mode, even though I'd expect them to appear:

Replace | Only
-|-
image | image

In comparison, this works with the **Replace** and **Overlay** shadowmask modes.

- During a recent rendering meeting, it was suggested that the shadowmask texel scale should be adjustable separately from the lightmap texel scale, so that your shadowmask can be more detailed than the lightmap. This can be important if you only bake indirect light, as indirect light data is low-frequency but the shadowmask can be higher-frequency, especially if using the **Only** shadowmask mode.
  - I'd probably add **Shadowmask Texel Scale** property below **Shadowmask Mode** that is only shown when **Shadowmask Mode** is set to a value other than **None**. This value would be relative to **Texel Scale**, with `1.0` being the same resolution, `2.0` being double resolution on each axis and so on.

BlueCube3310:
> * When you set the shadowmask mdoe to **None** and bake lightmaps again, the shadowmask source texture isn't removed from the filesystem (unlike what the description implies).

I'm not sure whether this should be the correct behavior, I intended the docs to imply that the reference to the texture would be removed so it doesn't get loaded. Though it may be better from the user's perspective to delete it, since it will get overwritten anyway on subsequent bakes.

>* With all rendering methods, dynamic object shadows are not rendered (even up close) when using the **Only** shadowmask mode, even though I'd expect them to appear:
> 

I'm not certain how to approach this, currently all real-time shadows are ignored in the shader when an object has an 'Only' shadowmask. One way would be to only disable shadow casting for static objects with a shadowmask, but that would mean they wouldn't also cast shadows onto dynamic ones. The shadows of static and dynamic objects are combined, so it's not possible to tell which ones are cast by the dynamic ones.

SpockBauru:
> * With all rendering methods, dynamic object shadows are not rendered (even up close) when using the **Only** shadowmask mode, even though I'd expect them to appear:

This is not what I would expect for a mode called "Only". I think it can be added as a new mode after the merge, maybe something like "static + dynamic".

This is a big PR and is in a good shape, there’s no need to keep overthinking it. New features can always be added later. Please, consider merge before we miss the window of opportunity ^_^

pineapplemachine:
Seconding @SpockBauru. Not having shadow masks in any form currently makes Godot's baked shadows impractical for a lot of projects, forcing those projects to accept the performance cost of real-time shadows for static geometry only in order to have acceptable shadows for dynamic geometry. This feature is currently a major blocker for optimizing performance in 3D games and supporting lower-end platforms with Godot without sacrificing shadows entirely. In my opinion, getting this merged in at least a basically usable form ought to be a very high priority.

mrjustaguy:
Eh, correct me if I'm wrong, but wouldn't you still need to render static objects in the shadow map for dynamic objects to be able to get statically cast shadows?

atirut-w:
> Eh, correct me if I'm wrong, but wouldn't you still need to render static objects in the shadow map for dynamic objects to be able to get statically cast shadows?

I've also initially thought about this, but we can't separate the two because they render into the same shadow map AFAIK. This means that if static objects both receive shadows and use static shadowmap, they could potentially look *wrong* due to getting both dynamic and static shadows.

This is why most engines have the **Only** mode not cast dyanmic shadows onto static objects, and **Replace** mode for when you need that to work where static shadows are only rendered outside of dynamic shadow's range.

SpockBauru:
Updated my MRP to the latest windows fragment:
[shadowmask_test.zip](https://github.com/user-attachments/files/17575152/shadowmask_test.zip)

I'm getting flickering issues while in movement in Forward+ rendering. This does not happens on Mobile or Compatibility:


https://github.com/user-attachments/assets/839283b3-63bb-4b28-832d-a2e110380da5

marcinn:
I see no flickering on my setup.
I'm noticing that opacity is always 1.0 for shadowmask=overlay (for replace it works fine). I don't know if it's ok.

image


----

This PR is a huge improvement in terms of quality and performance. I can switch to short distance (i.e. 50m) and cheap  orthogonal shadows, and still the entire scene is properly lit.  Thank you!

BlueCube3310:
Both issues should now be resolved

marcinn:
Thanks. Opacity works fine now!

----

One more question. How shadowmasking should work with fog/volumetric fog?
With volumetric fog enabled darkest parts are drawn in color of the fog:

shadowmask_volumetricfog

Without volumetric fog you can see pure darkness:
image

The goal is to achieve lightshafts from volumetric fog. I can achieve this by setting global density=0 and local fog volumes for each light. 

image

Is this a proper way to achieve this?

marcinn:
Can I exclude some meshes from baking their shadows or receiving static shadows?

I have moving platforms like this one.  In overlay mode two shadows are visible:
image

This can be partially workarounded by "replace" mode, but after shadow max distance the baked shadows are getting visible:
image

Shadows of other objects were also baked on these platforms:
image

As a workaround I'm thinking about hiding all moving platforms during baking lights, but I think that would be nice to have shadowmask layers (similar to light_cull_mask) to exclude object layers from baking their shadows.

As far I can see there is a Caster Mask, so I've excluded layer 2:
image

I've also set platform's mesh to layer 2:
image

But this changed nothing. Platform's shadows are baked. I'm not sure is this a bug or not.

SpockBauru:
@BlueCube3310 
> Both issues should now be resolved

Thank you, now Forward+ works and Mobile seems to be working. But I'm facing a weird issue on the Compatibility renderer, it took a while to find the issue but here it is: 

On Compatibility renderer when I set a Spotlight to "Dynamic" the shadowmask vanishes and is replaced by the shadowmap, even when Shadowmask is set to "Only". When I look away from the light the shadowmask reapears. Rebaking doesn't help the issue.

https://github.com/user-attachments/assets/617b6861-aace-43b4-aa82-2bbad83ce645



@marcinn 
> Can I exclude some meshes from baking their shadows or receiving static shadows?

Not sure If I understand the issue, maybe you are referring to setting the `Geometry Instance` -> `Global Illumination` > `Mode` to `Dynamic`:

image

In LightmapGI, when the mesh GI Mode is set to Dynamic it will be lit by the [Lightmap Probes (that little ballls)](https://docs.godotengine.org/en/stable/tutorials/3d/global_illumination/using_lightmap_gi.html#dynamic-objects). Those are like "little lights" that illuminates dynamic objects. The `gen probes` setting controls how many of them are in the screen and you can put more with nodes `LightmapProbe` if needed. 

Shadowmasks are not baked on Dynamic objects.

marcinn:
>  maybe you are referring to setting the Geometry Instance -> Global Illumination > Mode to Dynamic
> Shadowmasks are not baked on Dynamic objects.

Thanks! That was the cause. I completely forgot about GI mode. My apologies!

Calinou:
> As far I can see there is a Caster Mask, so I've excluded layer 2:

Note that this option doesn't affect lightmap baking due to https://github.com/godotengine/godot/issues/98684.

marcinn:
> Note that this option doesn't affect shadowmask baking due to #98684.

Oh, thanks for pointing this out!

BlueCube3310:
> Thanks. Opacity works fine now!
> 
> One more question. How shadowmasking should work with fog/volumetric fog? With volumetric fog enabled darkest parts are drawn in color of the fog:
> 
> shadowmask_volumetricfog
> 
> Without volumetric fog you can see pure darkness: image
> 
> The goal is to achieve lightshafts from volumetric fog. I can achieve this by setting global density=0 and local fog volumes for each light.
> 
> image
> 
> Is this a proper way to achieve this?

Volumetric fog and lightshafts depend on the real-time shadow projection since they contain more information (like depth), which is required for certain effects to work correctly. The shadowmasks are simply textures applied to a model's UV2 channel, they only store the shadow intensity.

BlueCube3310:
Shadowmasks with vertex shading now work correctly.

There are two remaining issues which I don't know how to fix:
- with the Mobile renderer, lightmaps don't render at all on certain scenes. This is unrelated to this PR, however,
- the aforementioned issue with spot lights on Compatibility. For some reason, it looks like the shaders don't process the shadowmask logic when an object has additive lighting

BlueCube3310:
Point/omni lights and shadowmasks should now work correctly on Compatibility. Therefore, I think this PR is now complete

SpockBauru:
Tested my MRP on Forward+, Mobile and Compatibility renderers. Shadowmasks is working as intended without any visual issues. Also baked the LightmapGI on each renderer and everything just works \o/

I didn't find any issue related to this PR and believe that is ready do merge :D

https://github.com/user-attachments/assets/44431be1-2253-4aa9-954c-1df46a0b2a27


### After merge

After merging, here are some thoughts about what I believe should be prioritized for LightmapGI overall:

**After merge LightmapGI Blocking bug**
Lightmap Probes are completely broken on both Mobile and Compatibility renderers, making LightmapGI unusable on the devices that would benefit most from it.

I believe this bug should be prioritized.

**After merge todo feature**
As discussed before, including an option for Shadowmask resolution would be a nice to have, but only after the merge!

**After merge to be discussed**
Currently the "Shadowmask Only" option seems to have different opinions on what it should do. I believe that it should disable all shadowmaps for good letting only shadowmasks and nothing else for both static and dynamic objects. We may have other option that project shadowmaps from dynamic objects on purely static objects (not possible right now). But these needs to be carefully discussed after the merge.

clayjohn (changes requested):
We discussed this in the Rendering meeting today and the consensus is that we should move forward and merge this as soon as possible. However, we are a bit concerned about the current `ShadowmaskMode` options. They are a bit confusing and they seem to cater to very niche use-cases that may not be widely used. 

- Replace seems to be what most people expect from this feature. It allows for having baked shadows at a distance and high quality shadows when up close when baking indirect lighting
- Overlay seems interesting, but problematic. It will break if you have any animated objects. I suspect that there will be few cases where users want to have the low resolution lightmap fully overlaid on dynamic shadows. I suspect that it would be better to have an option to blend between the shadowmask and shadow map at the edges of the shadow range instead
- Only appears to be an inferior version of using the "Static" lightmapping mode on Light3Ds. It forces using the shadowmask (which is good for performance), but still uses real time direct lighting, so it is worse for performance than static lights. Ultimately, it presents a clunky workflow to do something that is already supported by the engine. I suspect that this option was more useful before we implemented bicubic sampling because users were unhappy with the quality of baked shadows before. 

Accordingly, we agreed that two things are needed before merging:
1. Un-expose the `SHADOWMASK_MODE_ONLY` enum. Its fine to leave the code in internally for now, but we don't want to expose it to users unless we need to.
2. Mark the `shadowmask_mode` as experimental. (We also need to call it out in the release notes as something that will likely change)

Personally, I suspect that we can get away with making the shadowmask a binary selection "SHADOW_MASK_ENABLED", "SHADOW_MASK_DISABLED" with a boolean that pops up when enabled to turn on/off blending. But I would like to get some more user feedback about what they need before we go down the rabbit hole of trying to decide on users problems for them.

SpockBauru:
I believe that the Shadowmask Only will be beneficial when the option to increase shadowmask resolution is implemented, so we can have high frequency shadowmask and low frequency lightmap. But this feature can wait.

For now I fully agree with @clayjohn: we should un-expose the shadowmask modes, keep only the binary selection and merge as soon as possible. Just keep the code for future improvements.

SpockBauru:
Tested [my MRP](https://github.com/godotengine/godot/pull/85653#issuecomment-2447529657) on Forward+, Mobile and Compatibility renderers. Shadowmasks is working as intended without any visual issues. Baked the LightmapGI on each renderer and everything is working.

Also made a quick test on my (halted) personal project. Out of the box, it works as expected without any issues. Far shadows are now present:

| Shadowmaps None| Shadowmaps Replace|
| ----- | ---- |
| shadowmap_none | shadowmap_replace |

With this developers can make an option to reduce shadowmaps resolution and distance, improving the performance on weak systems.

I didn't find any issue related to this PR and believe that is ready to merge.

clayjohn (approved):
Let's go ahead and merge this!

It looks great and while I think there is room for iteration, I think it's time to get this in the hands of users so we can get some feedback

akien-mga:
Thanks!

Calinou:
@BlueCube3310 I noticed shadowmasks are saved as PNGs, but saving them as WebPs can make them much smaller on disk:

```
321k test_shadow.png
151k test_shadow.webp
```

This does not affect memory usage, but it'll reduce file sizes in version control, which is helpful as lightmap data is often iterated upon.

The downside is that we can only do that for images that are 16383×16383 or smaller, so shadowmasks that are larger would have to be resized down. In practice, only 16384×16384 can be an issue here, so we could downsize it to 16383×16383 upon saving (Godot's VRAM compression will pad the last pixel automatically).

michaelharmonart:
Not sure if this is the place to mention it, but after some testing it seems that this doesn't take into account whether shadow casting is on per mesh. Would be nice to have that supported as well to be able to disable shadow casters for baking as well.

clayjohn:
> Not sure if this is the place to mention it, but after some testing it seems that this doesn't take into account whether shadow casting is on per mesh. Would be nice to have that supported as well to be able to disable shadow casters for baking as well.

That's an existing issue in LightmapGI. https://github.com/godotengine/godot/issues/98684

michaelharmonart:
ahh ok. l never noticed since there was never a way to blend between real-time and baked lighting like this before now

I noticed now only since there was a clear transition line between realtime and baked shadows only since there was a mismatch in which objects had shadows.

IsaacMarovitz:
In 4.4.dev7 despite, having shadowmask mode set to Replace, baking does not seem to build the shadowmasks. 

image

I can't tell if I'm using it wrong, or if this is a bug.

SpockBauru:
> baking does not seem to build the shadowmasks.

@IsaacMarovitz It's working here, did you remembered to set shadowmasks to replace **before** baking?

IsaacMarovitz:
> It's working here, did you remembered to set shadowmasks to replace before baking?

I did. I'll test on another device to see if maybe this is a Mac thing.

IsaacMarovitz:
Nvm, I was using it wrong, my direction light was set to static and I missed the warn in the console. The DirectionLight Bake Mode documentation needs to be updated to reflect this change.

unfa:
Hey, where can I find a good explanation of this feature? I was looking around and couldn't really find anything.

SpockBauru:
Officially there's an explanation on the "latest" version of the docs, not the "stable" one since 4.4 stable was not released yet. Be aware that it's an experimental feature so there's no much explanation yet: 

- https://docs.godotengine.org/en/latest/classes/class_lightmapgi.html#class-lightmapgi-property-shadowmask-mode 
- https://docs.godotengine.org/en/latest/classes/class_lightmapgidata.html#enum-lightmapgidata-shadowmaskmode



I'm also answering people on this Reddit thread: https://www.reddit.com/r/godot/comments/1hcr73g

Calinou:
> Hey, where can I find a good explanation of this feature? I was looking around and couldn't really find anything.

I've opened a PR to document shadowmasks: https://github.com/godotengine/godot-docs/pull/10482

GustJc:
Not sure if something changed. 
But using the latest master (v4.4.dev.gh [24d74510e]) and trying to bake the lightmapGI using the testproject from (https://github.com/godotengine/godot/pull/85653#issuecomment-2447529657) crashes Godot.
With a lot of "Couldn't map PC to fn name" errors.

But testing it using a simple scene on a new projected or using the test_lightmapgi_mobile project (https://github.com/godotengine/godot/pull/85653#pullrequestreview-1768657637) worked fine. 



https://github.com/user-attachments/assets/852e65f9-1f42-476e-928f-f004ba76c92c

Here is the start of the crash:
(On Windows 10, Foward+, Nvidea 1070)


shadowmask

SpockBauru:
Compiled locally and can confirm that the demo is not working anymore. First time I baked, it got my PC unresponsive for a while and I had to make a hard reset. Second time it crashed Godot.

My system: `Godot v4.4.dev (24d74510e) - Windows 11 (build 22631) - Multi-window, 1 monitor - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3080 (NVIDIA; 32.0.15.6614) - 12th Gen Intel(R) Core(TM) i7-12700KF (20 threads)`

@GustJc Would you kindly open a new issue?

SpockBauru:
> Compiled the master branch locally and can confirm that the demo is not working anymore. First time I baked, it got my PC unresponsive for a while and I had to make a hard reset. Second time it crashed Godot.
> 
> My system: `Godot v4.4.dev (24d74510e) - Windows 11 (build 22631) - Multi-window, 1 monitor - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3080 (NVIDIA; 32.0.15.6614) - 12th Gen Intel(R) Core(TM) i7-12700KF (20 threads)`
> 
> @GustJc Would you kindly open a new issue?

FeralBytes:
@clayjohn 
> Personally, I suspect that we can get away with making the shadowmask a binary selection "SHADOW_MASK_ENABLED", "SHADOW_MASK_DISABLED" with a boolean that pops up when enabled to turn on/off blending. But I would like to get some more user feedback about what they need before we go down the rabbit hole of trying to decide on users problems for them.

I would like to provide some user feedback. In my limited testing overlay seemed to work better for me on the web than replace, because the default dynamic shadow distance was so low in compatibility mode on the web, the shadow would begin to hide itself before it could be replaced for objects that were tall, such as a hill nearby. So for the web, which really needs such a performant feature, I think overlay may be a better option than replace.

That also makes me wonder if shadow mask ONLY would do even better. But it does really come do to the needs of the game and the artist designing the game. I personally am in favor of keeping all of the modes exposed, because Godot is a general solution game engine. People have many potential use cases and we cannot know them all right now, so early in the release of the feature. 

Thank you for your time.

PR #85662, opened 2023-12-02
Add support for stopping `SceneTreeTimer`

### LINKED ISSUE #8577: Include a method to stop a SceneTreeTimer

SceneTreeTimers can be generated but there is no mechanism to stop them. If it is necessary to interrupt the timer for any particular reason one must wait for the timer to finish or introduce some mechanism on methods triggered by the timeout signal that they should not run anymore. This is a problem when trying to re-use resources that have SceneTreeTimers since there is no way to stop those timers when the resource is not in use.

A PR has been made with a possible implementation of this proposal:
https://github.com/godotengine/godot/pull/85662

As described above, being unable to safely stop SceneTreeTimers, prevents their use in reusable resources. This is important for console games, where reusing nodes is a major way to overcome the console limitations. It is also important for the sake of consistency. Tree-generated tweens can be stopped so it should come naturally that tree-generated timers should also be stoppable.

Include a method in the SceneTreeTimer to stop the timer.

Referring to https://github.com/godotengine/godot/pull/85662

There is little to no control of SceneTreeTimers. Of course, there are ways to avoid the impact of not stopping a timer. But this change would greatly simplify their use when not following the "await" use case.

I'm not entirely sure that SceneTreeTimers can be controlled by a plugin.

bitbrain:
The `SceneTreeTween` (3.5+) / `Tween` (4.x+) exposes a `stop()` [method already](hhttps://docs.godotengine.org/en/4.2/classes/class_tween.html#class-tween-method-stop), however, `SceneTreeTimer` does not currently. This pull request aims to fix this inconsistency and give users the option to stop scene tree timers.

**EDIT** this pull request is the result from suggestions of people in my community as they were unable to stop SceneTreeTimer and it caused confusion why e.g. `Timer` and `Tween` can be stopped, but not `SceneTreeTimer`. A proposal to this addition will be linked up separately to facilitate discussion on this change.

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/8577*

### DIFF
--- a/doc/classes/SceneTreeTimer.xml
+++ b/doc/classes/SceneTreeTimer.xml
@@ -27,6 +27,20 @@
 	</description>
 	<tutorials>
 	</tutorials>
+	<methods>
+		<method name="is_stopped" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns true in case this timer has been stopped.
+			</description>
+		</method>
+		<method name="stop">
+			<return type="void" />
+			<description>
+				Stops this timer at the next time this timer would be processed. No further signals will be emitted.
+			</description>
+		</method>
+	</methods>
 	<members>
 		<member name="time_left" type="float" setter="set_time_left" getter="get_time_left">
 			The time remaining (in seconds).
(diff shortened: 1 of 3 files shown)


AThousandShips:
What is the need to be able to do this? Consistency between different, unrelated types in itself isn't justification for adding this, what would this be needed for?

I'd suggest opening a [proposal](https://github.com/godotengine/godot-proposals) to gauge the need for this

bitbrain:
@AThousandShips people in my community asked me why they are unable to stop `SceneTreeTimer` (as they require it to do so). My hope was that this change is so tiny that it may not require a proposal to begin with. I am fine with this PR to be closed and we will provide a proposal first instead.

AThousandShips:
You don't need to close it, but please do open a proposal so this can be discussed better (and you didn't mention the fact people had asked for it in the PR so it's not clear from the start 🙂)

Issue #84258, opened 2023-10-31
Mouse input events not being detected from Area2D inside a SubViewport unless a key is pressed


lazaromenezes:
### Godot version

v4.2.beta4.official [93cdacbb0]

### System information

Godot v4.2.beta4 - Windows 10.0.23570 - Vulkan (Mobile) - dedicated NVIDIA GeForce RTX 3060 (NVIDIA; 31.0.15.5009) - AMD Ryzen 5 5500 (12 Threads)

### Issue description

Testing my 4.1.2 project with 4.2-beta 4 I noticed this behavior that might be related with the changes within subviewports and event handling. I'm not sure if any of the issues mentioned by #83276 is really similar to this so decided to open a new issue.

It seems that mouse events are not being sent to an Area2D that's a child from a sub view port, unless some other event is fired up into the main viewport before.

In the screen recording bellow, 'D' is a direct child from my root scene node while 'S' is inside a Subviewport

https://github.com/godotengine/godot/assets/8123916/3992481a-0890-44bb-990f-393786b369b2

This is the setup:

![image](https://github.com/godotengine/godot/assets/8123916/6ac78a92-0b53-4a2d-b9a7-aaa4f9a15368)

The mouse_entered and mouse_exited events are captured properly by the direct child scene "D" while the events only has effects in "S" if other event is fired before. Per exemple, start a click in "D" and drag until "S" or hold any keyboard key pressed and hover over "S", then the mouse_entered and mouse_exited work.

### Steps to reproduce

Using the given MRP:

1) Load the main scene
2) Hover over the tile "D", it should become green;
3) Exit mouse from the tile "D", it should be back to its original color;
4) Hover over the tile "S", it was supposed to become green, but it is not;
  4.1) Hold any keyboard key and hover the tile "S", it becomes green.

### Minimal reproduction project

[subviewport_mrp.zip](https://github.com/godotengine/godot/files/13219169/subviewport_mrp.zip)

AThousandShips:
To confirm, this behaviour is not present in 4.1.2?

lazaromenezes:
It is not. It works fine for both the MRP and my full project. I can say it was also not working in "4.2-beta 2" but I couldn't investigate further at the time so I don't know if the missbehavior is exactly the same as I'm reporting now.

lazaromenezes:
I was going to test the build for #83276 but it's expired :(

Sauermann:
[...]

PR #77926, opened 2023-06-06
Deprecate `Viewport::handle_input_locally`

Sauermann:
Deprecate `Viewport::handle_input_locally` in favor of:
- an automatism in `SubViewportContainer`, that propagates the event handled state to the parent viewport
- a parameter in `SubViewportContainer`, that decides, if input events should be sent to additional child SubViewports, when the event is set to handled in one of the SubViewports.

The use-case, that `SubViewport.handle_input_locally = true` achieves, is to allow a single input event to be handled in multiple SubViewports of a `SubViewportContainer`. However, this functionality is very inaccessible.
This PR implements a different method to allow this functionality by adding `SubViewportContainer.send_input_to_all_viewports`.

This change solves the problem of the automatic change of `SubViewport.handle_input_locally` (resolve #56502 (#46982))
By doing this, it makes the mentioned functionality more reliable and accessible (resolve #89073)

revert #77730 and make sure, that #76439 is not reintroduced
revert #87385 and make sure, that #84258 is not reintroduced
takes ideas of #46994

Updated 2023-09-02: Changes to documentation.
Updated 2024-01-24: Fix merge conflict with #77730
Updated 2024-01-25: Fix merge conflict with #87385
Updated 2024-03-07: Rebase after pre-commit-hook changes, Implement a replacement for the `handle_input_locally` use-case.
Updated 2024-09-15: Fix merge conflict
Updated 2024-12-01: Fix merge conflict

PR #37317, opened 2020-03-26
Separate DisplayServer from OS and add multiple windows support

akien-mga:
Same content as the `display-server` branch that @reduz worked on (with help from @kuruk-mm and @bruvzg), but rebased on top of the current `master` branch (only took ~5 hours!).

- `OS` is split into `OS` and `DisplayServer`.
- `X11` OS is renamed to `LinuxBSD` (Linux, FreeBSD and OpenBSD "desktop" systems as per Freedesktop), `X11` is now its first `DisplayServer` implementation (`DisplayServerWayland` can be added eventually).
- `Input` refactor (the singleton is now `InputFilter`).
- Adds `Window` node which is the new scene root.
- Multiple window support.
- Dialogs, Popups and Inspector docks now use Window.
- Lots of bugs, let's bring back instability to the `master` branch!

I forgot to use `--preserve-merges` when rebasing, so we lost the references to PRs #36936 and #37029 which were merged in `display-server`.

PR #85665, opened 2023-12-02
[4.2] Fix SubViewport physics picking

Sauermann:
Apply the logic of `handle_input_locally` for physics picking (as done in #77730).

resolve in 4.2 issue #84258

For master I consider #77926 the better approach, but in order to not break compatibility in 4.2, I consider this a viable option to fix this in 4.2.

### DIFF
--- a/scene/main/viewport.cpp
+++ b/scene/main/viewport.cpp
@@ -739,6 +739,14 @@ void Viewport::_process_picking() {
 
 	while (physics_picking_events.size()) {
 		local_input_handled = false;
+		if (!handle_input_locally) {
+			Viewport *vp = this;
+			while (!Object::cast_to<Window>(vp) && vp->get_parent()) {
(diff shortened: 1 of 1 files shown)


AThousandShips:
Also considering this copies the code from #77730 I'd say a co-author is required

Sauermann:
Initially this code-section was introduced by reduz in #37317:
https://github.com/godotengine/godot/blame/d76c1d0e516fedc535a2e394ab780cac79203477/scene/main/viewport.cpp#L3657-L3673
Also the style with the multiple `break;` conditionals comes from that PR.

Please let me know, if I should add reduz as co-author. I'm not yet very familiar with when to use co-author.

AThousandShips:
Then no need for co-author IMO, thought the other PR introduced this specific code

I'd suggest compressing the code regardless of the other blocks but it's either way 🙂

akien-mga (approved):

PR #85666, opened 2023-12-02
Add more C++ unit tests for String to number conversions

miv391:
This PR adds more C++ units tests to the following string to number conversion functions:
- `String::to_int()`
- `String::hex_to_int()`
- `String::bin_to_int()`
- `String::to_float()`

There are inconsistencies with the conversion functions, but the goal of this PR is not to fix anything, the goal is to document how the conversions currently work using the unit tests.

I'm planning to implement this https://github.com/godotengine/godot-proposals/issues/8546, but I'm not touching conversion functions (even a little bit) without first adding more unit tests.

### DIFF
--- a/tests/core/string/test_string.h
+++ b/tests/core/string/test_string.h
@@ -456,30 +456,93 @@ TEST_CASE("[String] Number to string") {
 }
 
 TEST_CASE("[String] String to integer") {
-	static const char *nums[4] = { "1237461283", "- 22", "0", " - 1123412" };
-	static const int num[4] = { 1237461283, -22, 0, -1123412 };
+	static const char *nums[14] = { "1237461283", "- 22", "0", " - 1123412", "", "10_000_000", "-1_2_3_4", "10__000", "  1  2  34 ", "-0", "007", "--45", "---46", "-7-2" };
+	static const int num[14] = { 1237461283, -22, 0, -1123412, 0, 10000000, -1234, 10000, 1234, 0, 7, 45, -46, -72 };
 
-	for (int i = 0; i < 4; i++) {
+	for (int i = 0; i < 14; i++) {
 		CHECK(String(nums[i]).to_int() == num[i]);
 	}
+	CHECK(String("0b1011").to_int() == 1011); // Looks like a binary number, but to_int() handles this as a base-10 number, "b" is just ignored.
+	CHECK(String("0x1012").to_int() == 1012); // Looks like a hexadecimal number, but to_int() handles this as a base-10 number, "x" is just ignored.
+
+	ERR_PRINT_OFF
+	CHECK(String("999999999999999999999999999999999999999999999999999999999").to_int() == INT64_MAX); // Too large, largest possible is returned.
+	CHECK(String("-999999999999999999999999999999999999999999999999999999999").to_int() == INT64_MIN); // Too small, smallest possible is returned.
+	ERR_PRINT_ON
 }
 
 TEST_CASE("[String] Hex to integer") {
-	static const char *nums[4] = { "0xFFAE", "22", "0", "AADDAD" };
-	static const int64_t num[4] = { 0xFFAE, 0x22, 0, 0xAADDAD };
+	static const char *nums[12] = { "0xFFAE", "22", "0", "AADDAD", "0x7FFFFFFFFFFFFFFF", "-0xf", "", "000", "000f", "0xaA", "-ff", "-" };
+	static const int64_t num[12] = { 0xFFAE, 0x22, 0, 0xAADDAD, 0x7FFFFFFFFFFFFFFF, -0xf, 0, 0, 0xf, 0xaa, -0xff, 0x0 };
 
-	for (int i = 0; i < 4; i++) {
+	for (int i = 0; i < 12; i++) {
 		CHECK(String(nums[i]).hex_to_int() == num[i]);
 	}
+
+	// Invalid hex strings should return 0.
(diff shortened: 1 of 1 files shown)


tests/core/string/test_string.h
```diff
@@ -456,30 +456,84 @@ TEST_CASE("[String] Number to string") {
 }
 
 TEST_CASE("[String] String to integer") {
-	static const char *nums[4] = { "1237461283", "- 22", "0", " - 1123412" };
-	static const int num[4] = { 1237461283, -22, 0, -1123412 };
+	static const char *nums[13] = { "1237461283", "- 22", "0", " - 1123412", "", "10_000_000", "-1_2_3_4", "10__000", "  1  2  34 ", "-0", "007", "0b1011", "0x1012" };
+	static const int num[13] = { 1237461283, -22, 0, -1123412, 0, 10000000, -1234, 10000, 1234, 0, 7, 1011, 1012 };
```
AThousandShips, line 460:
```suggestion
	static const int num[13] = { 1237461283, -22, 0, -1123412, 0, 10000000, -1234, 10000, 1234, 0, 7, 0b1011, 0x1012 };
```
These values are not aligned, `1011` in decimal is not the same as `1011` in binary, same for hexadecimal `1012` and decimal `1012`

miv391: `"0b1011"` and `"0x1012"` just look like a binary and a hexadecimal number, but the `to_int()` doesn't handle them as such. `to_int()` ignores any non digits, so what it sees is `"01011"` and `"01012"` and converts them as base-10.

AThousandShips: My bad, good point, that should be noted I think in the tests to clarify it's correct behavior

tests/core/string/test_string.h
```diff
@@ -456,30 +456,79 @@ TEST_CASE("[String] Number to string") {
...
 
-	for (int i = 0; i < 4; i++) {
+	for (int i = 0; i < 12; i++) {
 		CHECK(String(nums[i]).hex_to_int() == num[i]);
 	}
+
+	// Invalid hex strings should return 0.
+	static const char *invalid_nums[15] = { "qwerty", "QWERTY", "0xqwerty", "0x00qwerty", "qwerty00", "0x", "0x__", "__", "x12", "+", " ff", "ff ", "f f", "+ff", "--0x78" };
+
+	ERR_PRINT_OFF
```
AThousandShips, line 485: Is it required to put this here? The conversion only throws an error on too large values as far as I can see from the code

tests/core/string/test_string.h
```diff
@@ -456,30 +456,79 @@ TEST_CASE("[String] Number to string") {
...
 
-	for (int i = 0; i < 4; i++) {
+	for (int i = 0; i < 12; i++) {
 		CHECK(!(ABS(String(nums[i]).to_float() - num[i]) > 0.00001));
 	}
+
+	// Invalid float strings should return 0.
+	static const char *invalid_nums[6] = { "qwerty", "qwerty123", "0xffff", "0b1010", "--3.13", "__345" };
+
+	ERR_PRINT_OFF
```
AThousandShips, line 527: (my bad was meant here)
Is this really required? From what I can see there's no errors on this thrown

miv391: You are correct, only `hex_to_int()` prints errors when the string contains illegal characters and 0 is returned. Fixed this test and "Bin to integer" test. Also added the "too large exponent" tests.

miv391: Now the unnecessary `ERR_PRINT_OFF`'s are removed and "too high exponent" tests are added.

AThousandShips: Note that this will now print a warning cluttering the test output, see once it's run

AThousandShips: The too high exponent case isn't triggered here as it fires with exponent of 512, unsure why it's still infinity here though

tests/core/string/test_string.h
```diff
@@ -456,30 +456,93 @@ TEST_CASE("[String] Number to string") {
...
+
+	// Invalid float strings should return 0.
+	static const char *invalid_nums[6] = { "qwerty", "qwerty123", "0xffff", "0b1010", "--3.13", "__345" };
+
+	for (int i = 0; i < 6; i++) {
+		CHECK(String(invalid_nums[i]).to_float() == 0);
+	}
+
+	// Very large exponents.
+	CHECK(String("1e308").to_float() == 100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000.0);
```
AThousandShips, line 532:
```suggestion
	CHECK(String("1e308").to_float() == 1e308);
```
Unless there's some reason that won't work, or it will be very cluttered and error prone

tests/core/string/test_string.h
miv391: Fixed this. There seems to be different levels of "too high exponents". Floats between 1e309 - 1e511 will become `INFINITY`, but warning is not printed (`hex_to_int` prints errors, btw). When the float gets even larger, 1e512, warning is finally printed. Don't ask me why floats work like this :)

tests/core/string/test_string.h
miv391: Fixed.

Calinou (approved):
Code looks good on my end. Thanks for adding more tests :slightly_smiling_face:

AThousandShips (approved):

PR #63026, opened 2022-07-15
Some improvements to the SceneTreeTimer

Xwdit:
Replaced #55990 to be compatible with the latest master branch.
Implemented https://github.com/godotengine/godot-proposals/issues/4319

This PR brings some improvements to the SceneTreeTimer:

- Added a new parameter process_in_physics to SceneTree.create_timer() to update the SceneTreeTimer during the physics frame instead of the process frame.

- Added a new parameter ignore_time_scale to SceneTree.create_timer() to ignore Engine.time_scale and update the SceneTreeTimer with the actual frame delta.

- Code cleanup, moved the code that handles SceneTreeTimer to process_timers()

This should not break any compatibility, so maybe it can be cherrypick to 3.x?

PR #85674, opened 2023-12-02
Flush delete queue after process frame timers

KoBeWi:
After timers are processed in process frame, the delete queue is not flushed, which makes queued nodes exist until next frame, leading to unwanted behaviors. This is unlike physics frame, where the queue does flush after timers.

This is surprisingly easy to run into, because SceneTreeTimer defaults to process frame. Consider this code:
```GDScript
await get_tree().create_timer(0.1).timeout
queue_free()
```
The node will be freed one frame after it was queued, instead of the same frame as usually happens.

I made a reproduction project for a case where this is relevant:
[NodeStayer.zip](https://github.com/godotengine/godot/files/13536934/NodeStayer.zip)
Run `bug.tscn` and `nobug.tscn` and compare the results.

### DIFF
--- a/scene/main/scene_tree.cpp
+++ b/scene/main/scene_tree.cpp
@@ -471,12 +471,12 @@ bool SceneTree::physics_process(double p_time) {
 	MessageQueue::get_singleton()->flush(); //small little hack
 
 	process_timers(p_time, true); //go through timers
-
 	process_tweens(p_time, true);
 
 	flush_transform_notifications();
 	root_lock--;
 
+	// This should happen last because any processing that deletes something beforehand might expect the object to be removed in the same frame.
 	_flush_delete_queue();
 	_call_idle_callbacks();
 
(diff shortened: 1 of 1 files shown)


lawnjelly:
This makes sense to me although could do with more eyes to check.

A look at the blame history for this file suggests the relative order may be rather haphazard, with things inserted without having the reasoning available for the current order (this is where comments could have really helped, with comments indicating the reason for the current ordering).

e.g. In #63026 I have no idea why the timers were placed after flushing the delete queue, as timers might decide to delete something... Ah the existing timer code was there already, but the reasoning is lost to time / github history :grin: .

For instance if shifting the flush delete queue to the end, maybe add a comment that "this should happen last because any processing that deletes something beforehand might expect the object to be removed in this frame", otherwise other contributors will do the same and add things after it because they are not sure where to add. You could possibly flush the delete queue after the transform notifications, but in theory it might be better before because something might be deleted that doesn't require a transform notification.

KoBeWi:
Added comments and moved the delete flush after transform notifications. It's more consistent this way.

akien-mga (approved):
A bit late in the 4.3 cycle for that, should we give this a go early on in the 4.4 one?

Adding an approval to keep track of it, if anyone disagrees with the change, please voice it :)

KoBeWi:
Replaced by #93871, because GitHub decided to ignore pushed changes.

PR #35806, opened 2020-01-31
Add a tooltip delay setting for editor

Fermats-Fish:
Resolves #27879 and resolves #9030.

Adds an editor setting which changes the tooltip delay for tooltips inside the editor.

A restart is required for this change to be applied, however the code could be extended so that this is not required.

This is my first time contributing to a major open source project, and my first time using C++, so let me know if I should have done anything differently.

PR #85678, opened 2023-12-02
Add separate feature tags for editor runtime

### LINKED ISSUE #9030: Changing the tooltip delay in project settings affects the editor
**Operating system or device - Godot version:**
Windows 10 64bit - Godot 2.1.3 and Godot 3.0

**Issue description:**
Changing the tooltip delay in project settings will affect the editor's tooltips (after closing the project and reopening it). It should only affect the tooltips in-game.

**Steps to reproduce:**
Open any project, go to project settings and set tooltip delay to 0, then close the project and reopen it. Now all tooltips in the editor appear instantly.

### LINKED ISSUE #22073: Boot splash image is also used when Godot editor starts
**Godot version:**
Master

**Issue:**
Issue #4739 has been closed, but in master, if you choose an image for the boot splash, it's also used by Godot editor when it opens your project. If this is intended, I don't think it is a reasonable behavior.

**Steps to reproduce:**
Set a custom splash image in your project. Now it is used when your game boots.
Notice that it's also used by the Godot editor when it opens your project: some may like this, others may not.
This is especially important when the user replaces the boot splash with a blank image because he's using a custom splash screen scene. As an unexpected side effect, he will get a black boot splash when he opens his game project with the Godot editor.

**Example project:**
[boot_splash.zip](https://github.com/godotengine/godot/files/2384781/boot_splash.zip)

### LINKED ISSUE #27879: Allow changing of Editor tooltip delay
**Godot version:**

3.1

For tooltips such as found in the inspector. Currently there is no editor setting to change this. The project setting only affects the project, and would be cumbersome to use for the editor anyways.

I prefer no, or near instantaneous delay. Needing to wait nearly a full second for vital tooltip information is quite inconvenient.

KoBeWi:
This PR adds 2 new feature tags: `editor_hint` and `editor_runtime`. The former is used in the editor, the latter in running project. Both are only available in editor builds.
`editor` feature tag stays as is, for compatibility reasons and for consistency with `template`. [Documentation](https://docs.godotengine.org/en/stable/tutorials/export/feature_tags.html) will need updating after this is merged.

EDIT:
- Fixes #22073 You can override splash image for `editor_hint`.
- Fixes #9030 
- Fixes #27879
- Supersedes #35806

### DIFF
--- a/core/config/project_settings.cpp
+++ b/core/config/project_settings.cpp
@@ -1473,7 +1473,9 @@ ProjectSettings::ProjectSettings() {
 	GLOBAL_DEF(PropertyInfo(Variant::INT, "display/window/size/window_height_override", PROPERTY_HINT_RANGE, "0,4320,1,or_greater"), 0); // 8K resolution
 
 	GLOBAL_DEF("display/window/energy_saving/keep_screen_on", true);
-	GLOBAL_DEF("display/window/energy_saving/keep_screen_on.editor", false);
+#ifdef TOOLS_ENABLED
+	GLOBAL_DEF("display/window/energy_saving/keep_screen_on.editor_hint", false);
+#endif
 
 	GLOBAL_DEF("animation/warnings/check_invalid_track_paths", true);
 	GLOBAL_DEF("animation/warnings/check_angle_interpolation_type_conflicting", true);
@@ -1531,6 +1533,10 @@ ProjectSettings::ProjectSettings() {
 	GLOBAL_DEF_BASIC("internationalization/rendering/root_node_auto_translate", true);
 
 	GLOBAL_DEF(PropertyInfo(Variant::INT, "gui/timers/incremental_search_max_interval_msec", PROPERTY_HINT_RANGE, "0,10000,1,or_greater"), 2000);
+	GLOBAL_DEF(PropertyInfo(Variant::FLOAT, "gui/timers/tooltip_delay_sec", PROPERTY_HINT_RANGE, "0,5,0.01,or_greater"), 0.5);
+#ifdef TOOLS_ENABLED
+	GLOBAL_DEF("gui/timers/tooltip_delay_sec.editor_hint", 0.5);
+#endif
 
 	GLOBAL_DEF_BASIC("gui/common/snap_controls_to_pixels", true);
 	GLOBAL_DEF_BASIC("gui/fonts/dynamic_fonts/use_oversampling", true);
--- a/doc/classes/ProjectSettings.xml
+++ b/doc/classes/ProjectSettings.xml
@@ -254,6 +254,7 @@
 			Path to an image used as the boot splash. If left empty, the default Godot Engine splash will be displayed instead.
 			[b]Note:[/b] Only effective if [member application/boot_splash/show_image] is [code]true[/code].
 			[b]Note:[/b] The only supported format is PNG. Using another image format will result in an error.
+			[b]Note:[/b] The image will also show when opening the project in the editor. If you want to display the default splash image in the editor, add an empty override for [code]editor_hint[/code] feature.
 		</member>
 		<member name="application/boot_splash/minimum_display_time" type="int" setter="" getter="" default="0">
 			Minimum boot splash display time (in milliseconds). It is not recommended to set too high values for this setting.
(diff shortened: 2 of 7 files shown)


doc/classes/ProjectSettings.xml
```diff
@@ -756,8 +757,7 @@
 		<member name="display/window/energy_saving/keep_screen_on" type="bool" setter="" getter="" default="true">
 			If [code]true[/code], keeps the screen on (even in case of inactivity), so the screensaver does not take over. Works on desktop and mobile platforms.
 		</member>
-		<member name="display/window/energy_saving/keep_screen_on.editor" type="bool" setter="" getter="" default="false">
-			Editor-only override for [member display/window/energy_saving/keep_screen_on]. Does not affect exported projects in debug or release mode.
+		<member name="display/window/energy_saving/keep_screen_on.editor_main" type="bool" setter="" getter="" default="false">
```
AThousandShips, line 760:
```suggestion
		<member name="display/window/energy_saving/keep_screen_on.editor_main" type="bool" setter="" getter="" default="false">
			Editor-only override for [member display/window/energy_saving/keep_screen_on]. Does not affect exported projects in debug or release mode.
```
Or appropriate wording

RandomShaper:
Per the Godot jargon, it would have to be `tools*_`, but maybe that's not very user-friendly. Moreover, shouldn't `editor` already be used only for the editor itself vs. tools-enabled builds?

core/config/project_settings.cpp
```diff
@@ -1337,7 +1337,7 @@ ProjectSettings::ProjectSettings() {
 	GLOBAL_DEF(PropertyInfo(Variant::INT, "display/window/size/window_height_override", PROPERTY_HINT_RANGE, "0,4320,1,or_greater"), 0); // 8K resolution
 
 	GLOBAL_DEF("display/window/energy_saving/keep_screen_on", true);
-	GLOBAL_DEF("display/window/energy_saving/keep_screen_on.editor", false);
+	GLOBAL_DEF("display/window/energy_saving/keep_screen_on.editor_main", false);
```
Calinou, line 1340: This changes behavior a little since the project running from the editor will now keep the screen on (this may or may not be desired, especially on a laptop on battery). Just noting this for the future.

KoBeWi: I think the previous behavior was not really intended; it came from a limitation.

editor/project_settings_editor.cpp
```diff
@@ -287,6 +287,8 @@ void ProjectSettingsEditor::_add_feature_overrides() {
 	presets.insert("s3tc");
 	presets.insert("etc2");
 	presets.insert("editor");
+	presets.insert("editor_hint");
+	presets.insert("editor_runtime");
```
akien-mga, line 291: I wonder if we shouldn't have just `editor` and `editor_runtime` (so keep using `editor` for what is now `editor_hint` in this PR).

This would slightly break compat on the behavior of `editor`, but IMO the current behavior is confusing, and the two new feature tags make it worse as it's not obvious to users what's the difference between `editor` and `editor_hint`/`editor_runtime`.

core/config/project_settings.cpp
```diff
@@ -1473,7 +1473,7 @@ ProjectSettings::ProjectSettings() {
 	GLOBAL_DEF(PropertyInfo(Variant::INT, "display/window/size/window_height_override", PROPERTY_HINT_RANGE, "0,4320,1,or_greater"), 0); // 8K resolution
 
 	GLOBAL_DEF("display/window/energy_saving/keep_screen_on", true);
-	GLOBAL_DEF("display/window/energy_saving/keep_screen_on.editor", false);
+	GLOBAL_DEF("display/window/energy_saving/keep_screen_on.editor_hint", false);
```
akien-mga, line 1476: Should this also be enclosed in `#ifdef TOOLS_ENABLED`?

KoBeWi: It could.

editor/project_settings_editor.cpp
KoBeWi: `editor` is useful for any code that's supposed to run only in editor build, so that the code does not accidentally slip to exported project, thus making it closer to `editor_runtime`. From my experience at least, idk how other people use it.

akien-mga (approved):

akien-mga:
Thanks!

editor/project_settings_editor.cpp
KoBeWi: > I guess one could use debug for that?

Unless they ship game using debug build (which can happen sometimes ;)

PR #85681, opened 2023-12-02
Fix Tween loop initial value

### LINKED ISSUE #85644: Tweening modulate property doesn't behave right on 4.2
### Godot version

4.2-stable_mono

### System information

Windows 10, NVidiaRTX 2070

### Issue description

After updating to 4.2 I've realized that all my tweens working with the property "modulate" or "self_modulate" were behaving strangely... it happens when you use loops..  for instance if you have:

```
tween.SetLoop();
tween.TweenProperty(godotObject, "modulate", Colors.Red, 0.5f)
tween.TweenProperty(godotObject, "modulate", Colors.Blue, 0.5f);
```

that object should change its color from red to blue and then from blue to red and then loop, right? (basically, because its final value should be blue)

well, after 4.2 .. the color resets to its default color before starting a new loop, this didn't happen on older versions.
If you don't use Loops its final value persists, so the problem is only when using Loops.

so, for example if you had a modulate value of White before starting the loop.. the tween will change its modulate from red to blue, then resets to white and then start over again... 

* I haven't tested on GDScript

### Steps to reproduce

1- Create a CanvasItem such as a white ColorRect, so that you can tween its modulate property.
2- Attach a script that runs the following code (C#):

```
  tweener = this.CreateTween();
  tweener.SetLoops(0);
  tweener.SetEase(Tween.EaseType.Out);
  tweener.TweenProperty(godotObject, "modulate", Colors.Red, 0.5f);
  tweener.TweenProperty(godotObject, "modulate", Colors.Blue, 0.5f);
```

The modulate will change its value to default each time before starting a new loop

N/A

KoBeWi:
Fixes #85644

### DIFF
--- a/scene/animation/tween.cpp
+++ b/scene/animation/tween.cpp
@@ -549,9 +549,12 @@ void PropertyTweener::start() {
 		return;
 	}
 
-	if (do_continue && Math::is_zero_approx(delay)) {
-		initial_val = target_instance->get_indexed(property);
-		do_continue = false;
+	if (do_continue) {
+		if (Math::is_zero_approx(delay)) {
+			initial_val = target_instance->get_indexed(property);
+		} else {
+			do_continue_delayed = true;
+		}
 	}
 
 	if (relative) {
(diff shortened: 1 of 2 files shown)


akien-mga (approved):

YuriSizov:
Cherry-picked for 4.2.1.

PR #85686, opened 2023-12-02
Hide Node dock successfully on undo/redo and deletion 

### LINKED ISSUE #84530: Node Dock is not cleared when deleting selected node
### Godot version

4.2 beta4

### System information

### Issue description

When you select a node, the Node dock's signals and groups will point to that node. When you unselect it, the dock is properly cleared. If you instead delete the node, the dock content will stay and you can still make changes, which will go into the void (there is no way to restore the node without undoing the changes). What's worse, if you make a new scene and do the same to root node, it will result in editor crash.

Related is that the inspector does not fully clear either. The node header stays, the content is empty:
image

### Steps to reproduce

1. Add node
2. Select it
3. Go to Node dock's Groups
4. Press Delete to delete the node (tested with the "no confirm" action, IIRC Shift+Delete by default?)
5. The node is removed, but dock stays
6. Add groups

N/A

kevmorg:
This commit solves this specific issue: https://github.com/godotengine/godot/issues/84530

My partner and I located the source from this issue was in the _delete_confirm function in scene_tree_dock.cpp on line 2310.

The solution was to add NodeDock::get_singleton()->set_node(nullptr); to the function to make sure that the node dock isn't displaying signals and groups of the node after its deletion.

This is our first contribution to the project, please let us know if there is anything that we are doing wrong and how we can work to fix it.

*Bugsquad edit:*
- Fixes #84530

### DIFF
--- a/editor/node_dock.cpp
+++ b/editor/node_dock.cpp
@@ -67,14 +67,23 @@ void NodeDock::update_lists() {
 	connections->update_tree();
 }
 
+void NodeDock::_on_node_tree_exited() {
+	set_node(nullptr);
+}
+
 void NodeDock::set_node(Node *p_node) {
-	connections->set_node(p_node);
-	groups->set_current(p_node);
-	if (p_node) {
-		last_valid_node = p_node;
+	if (last_valid_node) {
+		last_valid_node->disconnect("tree_exited", callable_mp(this, &NodeDock::_on_node_tree_exited));
+		last_valid_node = nullptr;
 	}
 
+	connections->set_node(p_node);
+	groups->set_current(p_node);
+
 	if (p_node) {
+		last_valid_node = p_node;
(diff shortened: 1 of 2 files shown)


KoBeWi:
This does not cover all cases. The node can also be deleted as a result of undo/redo operation.
I think the most reliable way would be connecting the Node's `tree_exited` signal in node dock.

KoBeWi (approved):

akien-mga:

akien-mga:
There was still two commits after your rebase it seems, so I went ahead and squashed them myself, rewording the commit message to focus on the actual solution being implemented in the end instead of the original attempt that got reverted in the second commit.

akien-mga (approved):

akien-mga:

YuriSizov:
Cherry-picked for 4.2.2.

PR #85703, opened 2023-12-03
Make GDScriptAnalyzer aware of properties from other languages

TitanNano:
GDScript's auto-completion is aware of properties from scripts in other languages (integrated via GDExtensions) but the GDScript analyzer will report the property as unknown. 

The analyzer only ever checks the GDScript class and the native-type of an object for properties. The script type is currently ignored.

(this fix should also be compatible with at least 4.2 and 4.1)

### DIFF
--- a/modules/gdscript/gdscript_analyzer.cpp
+++ b/modules/gdscript/gdscript_analyzer.cpp
@@ -3769,6 +3769,60 @@ void GDScriptAnalyzer::reduce_identifier_from_base(GDScriptParser::IdentifierNod
 		}
 	}
 
+	// Check non-GDScript scripts.
+	Ref<Script> script_type = base.script_type;
+
+	if (base_class == nullptr && script_type.is_valid()) {
+		List<PropertyInfo> property_list;
+		script_type->get_script_property_list(&property_list);
+
+		for (const PropertyInfo &property_info : property_list) {
+			if (property_info.name != p_identifier->name) {
+				continue;
+			}
+
+			const GDScriptParser::DataType property_type = GDScriptAnalyzer::type_from_property(property_info, false, false);
+
+			p_identifier->set_datatype(property_type);
+			p_identifier->source = GDScriptParser::IdentifierNode::MEMBER_VARIABLE;
+			return;
+		}
+
+		MethodInfo method_info = script_type->get_method_info(p_identifier->name);
+
+		if (method_info.name == p_identifier->name) {
+			p_identifier->set_datatype(make_callable_type(method_info));
+			p_identifier->source = GDScriptParser::IdentifierNode::MEMBER_FUNCTION;
+			return;
+		}
+
+		List<MethodInfo> signal_list;
+		script_type->get_script_signal_list(&signal_list);
+
+		for (const MethodInfo &signal_info : signal_list) {
+			if (signal_info.name != p_identifier->name) {
+				continue;
+			}
+
+			const GDScriptParser::DataType signal_type = make_signal_type(signal_info);
+
+			p_identifier->set_datatype(signal_type);
+			p_identifier->source = GDScriptParser::IdentifierNode::MEMBER_SIGNAL;
+			return;
+		}
+
+		HashMap<StringName, Variant> constant_map;
+		script_type->get_constants(&constant_map);
+
+		if (constant_map.has(p_identifier->name)) {
+			Variant constant = constant_map.get(p_identifier->name);
+
+			p_identifier->set_datatype(make_builtin_meta_type(constant.get_type()));
+			p_identifier->source = GDScriptParser::IdentifierNode::MEMBER_CONSTANT;
+			return;
+		}
+	}
+
 	// Check native members. No need for native class recursion because Node exposes all Object's properties.
 	const StringName &native = base.native_type;
 


modules/gdscript/gdscript_analyzer.cpp
```diff
@@ -3769,6 +3769,27 @@ void GDScriptAnalyzer::reduce_identifier_from_base(GDScriptParser::IdentifierNod
 		}
 	}
 
+	// check non GDScript Scripts
```
fire, line 3772:
```suggestion
	// Check non-GDScript scripts.
```
We use sentence case for comments.

Sentence case is like the above comment.

TitanNano: Thanks for the feedback.

modules/gdscript/gdscript_analyzer.cpp
```diff
@@ -3769,6 +3769,27 @@ void GDScriptAnalyzer::reduce_identifier_from_base(GDScriptParser::IdentifierNod
...
+	const Ref<Script> script_type = base.script_type;
+
+	if (script_type.is_valid()) {
+		List<PropertyInfo> property_list;
+
+		script_type->get_script_property_list(&property_list);
+
+		for (PropertyInfo property_info : property_list) {
```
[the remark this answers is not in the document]
TitanNano: Thank you for the feedback.

modules/gdscript/gdscript_analyzer.cpp
```diff
@@ -3769,6 +3769,27 @@ void GDScriptAnalyzer::reduce_identifier_from_base(GDScriptParser::IdentifierNod
 		}
 	}
 
+	// Check non-GDScript scripts.
+	const Ref<Script> script_type = base.script_type;
```
AThousandShips, line 3773:
```suggestion
	Ref<Script> script_type = base.script_type;
```
Unnecessary here and potentially misleading, all it does is prevent reassigning it with a different value, `Ref` allows you to operate on it regardless

vnen:
Looks good, but I think it should also look for methods and signals.

TitanNano:
> Looks good, but I think it should also look for methods and signals.

@vnen Methods are already handled somewhere else: https://github.com/godotengine/godot/blob/d76c1d0e516fedc535a2e394ab780cac79203477/modules/gdscript/gdscript_analyzer.cpp#L5006

I can add signals, though. Should I also add constants here?

adamscott:
@TitanNano Hi and welcome around! Thanks for your PR, by the way.

Could you join a simple test project? I'm wondering how we could test this PR before merging it.

vnen:
I think the easiest way to test is with C#, since it's already integrated in Godot.

modules/gdscript/gdscript_analyzer.cpp
```diff
@@ -5005,7 +5063,7 @@ bool GDScriptAnalyzer::get_function_signature(GDScriptParser::Node *p_source, bo
 	while (base_script.is_valid() && base_script->has_method(function_name)) {
 		MethodInfo info = base_script->get_method_info(function_name);
 
-		if (!(info == MethodInfo())) {
+		if (!info.name.is_empty()) {
```
TitanNano, line 5066: @vreon @AThousandShips `info == MethodInfo()` is always `true` because internally, it compares the `id` field which is rarely changed from the default `0`. E.g. `MethodInfo::from_dict` which is used by `get_method_info` here never sets the `MethodInfo` `id`. 

I don't know what the preferred way of fixing this is here.

TitanNano:
> @TitanNano Hi and welcome around! Thanks for your PR, by the way.
> 
> Could you join a simple test project? I'm wondering how we could test this PR before merging it.

Do you have any pointers for me on how to get started with testing this? Especially with the mix of C# and GDScript.

modules/gdscript/gdscript_analyzer.cpp
```diff
@@ -3769,6 +3769,64 @@ void GDScriptAnalyzer::reduce_identifier_from_base(GDScriptParser::IdentifierNod
...
+			p_identifier->set_datatype(signal_type);
+			p_identifier->source = GDScriptParser::IdentifierNode::MEMBER_SIGNAL;
+			return;
+		}
+
+		HashMap<StringName, Variant> constant_map;
+
+		script_type->get_constants(&constant_map);
+
+
```
AThousandShips, line 3820:
```suggestion
```
Will break CI

modules/gdscript/gdscript_analyzer.cpp
vnen: I would prefer `if (info != MethodInfo())`, since it's clearer. `info.name.is_empty()` is not the most obvious for checking an empty MethodInfo.

TitanNano: I have now changed the equality operator of `MethodInfo` to also include the `name` field. I think this matches the implementation of the less-than operator.

core/object/object.h
```diff
@@ -235,7 +235,7 @@ struct MethodInfo {
 		return arguments_metadata.size() > p_arg ? arguments_metadata[p_arg] : 0;
 	}
 
-	inline bool operator==(const MethodInfo &p_method) const { return id == p_method.id; }
+	inline bool operator==(const MethodInfo &p_method) const { return id == p_method.id && name == p_method.name; }
```
vnen, line 238: I get why you changed this given the discussion, but you're not actually using this operator in the PR, so I think it's best to not include this change.

vnen:
Looks good to me. I did test this with a C# project and it seems to work well.

core/object/object.h
TitanNano: I understand. I can factor this change out into a separated PR.

TitanNano: I removed this change and will submit it as a separate PR later.

vnen (approved):
LGTM. Thanks for working on this!

modules/gdscript/gdscript_analyzer.cpp
```diff
@@ -3769,6 +3769,63 @@ void GDScriptAnalyzer::reduce_identifier_from_base(GDScriptParser::IdentifierNod
...
+			const GDScriptParser::DataType signal_type = make_signal_type(signal_info);
+
+			p_identifier->set_datatype(signal_type);
+			p_identifier->source = GDScriptParser::IdentifierNode::MEMBER_SIGNAL;
+			return;
+		}
+
+		HashMap<StringName, Variant> constant_map;
+
```
AThousandShips, line 3817:
```suggestion
```
Keep these grouped, whitespace here is confusing 🙂

modules/gdscript/gdscript_analyzer.cpp
```diff
@@ -3769,6 +3769,63 @@ void GDScriptAnalyzer::reduce_identifier_from_base(GDScriptParser::IdentifierNod
...
+		MethodInfo method_info = script_type->get_method_info(p_identifier->name);
+
+		if (method_info.name == p_identifier->name) {
+			p_identifier->set_datatype(make_callable_type(method_info));
+			p_identifier->source = GDScriptParser::IdentifierNode::MEMBER_FUNCTION;
+			return;
+		}
+
+		List<MethodInfo> signal_list;
+
```
AThousandShips, line 3801:
```suggestion
```

modules/gdscript/gdscript_analyzer.cpp
```diff
@@ -3769,6 +3769,63 @@ void GDScriptAnalyzer::reduce_identifier_from_base(GDScriptParser::IdentifierNod
 		}
 	}
 
+	// Check non-GDScript scripts.
+	Ref<Script> script_type = base.script_type;
+
+	if (base_class == nullptr && script_type.is_valid()) {
+		List<PropertyInfo> property_list;
+
```
AThousandShips, line 3777:
```suggestion
```

AThousandShips (approved):
Just some minor style nits, looks good otherwise

TitanNano:
@AThousandShips I applied your suggestions. Perhaps this can get merged then?

YuriSizov:

TitanNano:
@YuriSizov thank you for the message. How do I backport / cherry-pick this for 4.2 and 4.1? Do I just open pull requests against these branches with a reference to this PR? I wasn't able to find specific documentation.

YuriSizov:
@TitanNano You don't need to do anything if it's cherry-pickable without conflicts, we'll handle it. That said, the team hasn't expressed yet if this should be cherry-picked.

DavidVonDerau:
I'm not sure if this is related to this PR or not, but I was going through the commits that went into 4.3 and this seemed like a possible source of the following regression.

---

In `4.2.stable`, the following works w/ type checking disabled:

image

where `Multiply` is just an example method defined as

image

If I enable type checking in `4.2.stable`, I get the same warning that prompted this PR:

image

But if I use the `4.3.dev1` release containing this PR, I see the following error:

image

This is a parse error, so I can't just disable warnings.

I can get the code to parse correctly by using `call`, like so:

image

But this doesn't match the documentation on how cross-script method calls should work, and this is not backwards compatible with how it worked in 4.2 -- https://docs.godotengine.org/en/stable/tutorials/scripting/cross_language_scripting.html

TitanNano:
@DavidVonDerau there is #86259 which should fix this problem.

akien-mga:
Cherry-picked for 4.2.2.

Issue #85358, opened 2023-11-25
GPUParticles3d.emit_particle() EMIT_FLAG_COLOR does not work


thmasn:
### Godot version

4.2 RC1 and 4.2 RC2

### System information

macOs 13.5.2

### Issue description

when spawning particles on a GPUParticle3D Node via gdscript by calling emit_particle, different properties of the particle can be set.
the EMIT_FLAG_COLOR does not seem to register, and the spawned particle is not colored in the specified color.

according to the [docs](https://docs.godotengine.org/en/stable/classes/class_gpuparticles3d.html#class-gpuparticles3d-method-emit-particle), the [flag](https://docs.godotengine.org/en/stable/classes/class_gpuparticles3d.html#enum-gpuparticles3d-emitflags) 8 should set the color, but it doesnt respond. the flags for position (1) and velocity (4) work.

### Steps to reproduce

- create GPUParticle3D
- set up material to use vertex colors
- spawn a particle via code and set the color and flag
- the particle does not have the specified color

### Minimal reproduction project

minimum project with a GPUParticle3D and a script spawning particles with color red:
https://github.com/thmasn/ParticleTest

clayjohn:
It looks like the color is getting set correctly in `start()`. But in the `process()` function it is getting overridden:
https://github.com/godotengine/godot/blob/fa4a65387ea41506ecc519fc61a06da110dce51c/scene/resources/particle_process_material.cpp#L1053\

This will require a little bit of discussion. In 4.1.3, the emitted color is never used. That's just a design of the particles_process_material. You could use it in a custom shader and it would work fine. 

In 4.2-rc2 the color is correctly reset (or not) depending on the `EMIT_FLAG_COLOR` flag, but then the color value is ignored in the process function. So the end behaviour is the same. 

I think it makes sense to incorporate the emission color into the particles_process_material. But we need to discuss how we would do so. In the particles_process_material color is a dynamic thing that can change from frame to frame, so its not clear how the emitted color should interact with the dynamic color (should it multiply? should we have a new mode where color is not updated after emission?)

CC @QbieShay

QbieShay:
Huh. I don't know honestly how to make it work or how it is supposed to work. Did it work before?

clayjohn:
@QbieShay this is essentially a new behaviour request. It's not that it "worked" or "didn't work". The particles_process_material has always ignored the color of the emitted particle. But the OP is (rightfully IMO) surprised the the emission color is ignored.

[...]

PR #85717, opened 2023-12-03
Explain parameter usage in GPUParticles3D and GPUParticles2D

thmasn:
extended the description to explain how the passed parameters are used by the default ParticleProcessMaterial.
should set correct expectations for issues like [85358](https://github.com/godotengine/godot/issues/85358)

### DIFF
--- a/doc/classes/GPUParticles3D.xml
+++ b/doc/classes/GPUParticles3D.xml
@@ -35,6 +35,7 @@
 			<param index="4" name="flags" type="int" />
 			<description>
 				Emits a single particle. Whether [param xform], [param velocity], [param color] and [param custom] are applied depends on the value of [param flags]. See [enum EmitFlags].
+				The default ParticleProcessMaterial will overwrite [param color] and use the contents of [param custom] as [code](rotation, age, animation, lifetime)[/code].
 			</description>
 		</method>
 		<method name="get_draw_pass_mesh" qualifiers="const">
--- a/doc/classes/GPUParticles2D.xml
+++ b/doc/classes/GPUParticles2D.xml
@@ -37,6 +37,7 @@
 			<param index="4" name="flags" type="int" />
 			<description>
 				Emits a single particle. Whether [param xform], [param velocity], [param color] and [param custom] are applied depends on the value of [param flags]. See [enum EmitFlags].
+				The default ParticleProcessMaterial will overwrite [param color] and use the contents of [param custom] as [code](rotation, age, animation, lifetime)[/code].
 			</description>
 		</method>
 		<method name="restart">


thmasn:
how can this simple text change fail a check?
the brackets are used in other descriptions too, there are no special characters.
i assume the failed check for linux w/ mono has an external reason?

akien-mga:
@thmasn See the review comment by @AThousandShips above. There are spaces at the end of the line that should be removed (and she suggested adding a word too, though that part is not what causes the CI error, the trailing spaces are).

*Note:* Please make modifications by _amending_ the commit, and *force pushing* the changes, so that it stays as a single commit (see [PR workflow](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#modifying-a-pull-request)). While doing this, you should also amend the commit message to be more explicit (e.g. like the PR title).

thmasn:
i think i fixed it:
- trailing spaces are removed
- it is reduced to one commit with the correct title

doc/classes/GPUParticles3D.xml
```diff
@@ -34,7 +34,7 @@
 			<param index="3" name="custom" type="Color" />
 			<param index="4" name="flags" type="int" />
 			<description>
-				Emits a single particle. Whether [param xform], [param velocity], [param color] and [param custom] are applied depends on the value of [param flags]. See [enum EmitFlags].
+				Emits a single particle. Whether [param xform], [param velocity], [param color] and [param custom] are applied depends on the value of [param flags]. See [enum EmitFlags]. The default ParticleProcessMaterial will overwrite [param color] and use [param custom] as (rotation, age, animation, and lifetime).
```
akien-mga, line 37: Sorry I didn't notice this earlier (was on phone when I commented), but upon reading now I actually disagree with the earlier suggestion to add "and". The parenthesis is meant to represent what scalar each coordinate of the `Color` type will map to. I suggest making it clearer by using code markup.
```suggestion
				Emits a single particle. Whether [param xform], [param velocity], [param color] and [param custom] are applied depends on the value of [param flags]. See [enum EmitFlags]. The default ParticleProcessMaterial will overwrite [param color] and map the coordinates of [param custom] to [code](rotation, age, animation, lifetime)[/code].
```

thmasn:
i modified the text to also use the code-tag. i would favor the "use" word over "map", as the properties stay inside the vector4, and are maintained in this form.
and true, the way i understand it, GPUParticles2D uses the same ProcessMaterial, so the comment is valid for both.

i again amended the commit to match these changes.

QbieShay:
It will overwrite both custom and color

thmasn:
> It will overwrite both custom and color
yes, both are updated each frame in the processMaterial.

but, it seems to me like the intial parameters passed in custom still matter though: p.e. when you pass a lifetime of 0.0, or an age greater than the lifetime, the particle dies instantly.
on the other hand, the color that is beeing passed is instantly overriden, and therefore does not have any visible effect with the default processmaterial.

QbieShay:
Oooh good catch!

clayjohn (approved):

PR #85720, opened 2023-12-03
Added TextEditMenu for handling menu for TextEdit and LineEdit

TheSofox:
Fixes #85511

This PR creates a new class TextEditMenu which contains code that was previously duplicated in TextEdit and LineEdit. Specifically, it's for handling the context menu that pops up when you (for example) right click on a text field.

### DIFF
--- a/scene/gui/line_edit.cpp
+++ b/scene/gui/line_edit.cpp
@@ -252,9 +252,7 @@ void LineEdit::gui_input(const Ref<InputEvent> &p_event) {
 		}
 		if (b->is_pressed() && b->get_button_index() == MouseButton::RIGHT && context_menu_enabled) {
 			_update_context_menu();
-			menu->set_position(get_screen_position() + get_local_mouse_position());
-			menu->reset_size();
-			menu->popup();
+			context_menu.popup_at(get_screen_position() + get_local_mouse_position());
 			grab_focus();
 			accept_event();
 			return;
@@ -464,10 +462,7 @@ void LineEdit::gui_input(const Ref<InputEvent> &p_event) {
 			if (k->is_action("ui_menu", true)) {
 				_update_context_menu();
 				Point2 pos = Point2(get_caret_pixel_pos().x, (get_size().y + theme_cache.font->get_height(theme_cache.font_size)) / 2);
-				menu->set_position(get_screen_position() + pos);
-				menu->reset_size();
-				menu->popup();
-				menu->grab_focus();
+				context_menu.popup_at(get_screen_position() + pos, true);
 
 				accept_event();
 				return;
@@ -1472,7 +1467,7 @@ void LineEdit::_validate_caret_can_draw() {
 		draw_caret = true;
 		caret_blink_timer = 0.0;
 	}
-	caret_can_draw = editable && (window_has_focus || (menu && menu->has_focus())) && (has_focus() || caret_force_displayed);
+	caret_can_draw = editable && (window_has_focus || context_menu.has_focus()) && (has_focus() || caret_force_displayed);
 }
 
 void LineEdit::delete_char() {
@@ -1540,12 +1535,10 @@ void LineEdit::set_text_direction(Control::TextDirection p_text_direction) {
 		}
 		_shape();
 
-		if (menu_dir) {
-			menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_INHERITED), text_direction == TEXT_DIRECTION_INHERITED);
-			menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_AUTO), text_direction == TEXT_DIRECTION_AUTO);
-			menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_LTR), text_direction == TEXT_DIRECTION_LTR);
-			menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_RTL), text_direction == TEXT_DIRECTION_RTL);
+		if (context_menu.is_initialized()) {
+			context_menu.set_text_direction(text_direction);
 		}
+
 		queue_redraw();
 	}
 }
@@ -1569,9 +1562,7 @@ String LineEdit::get_language() const {
 void LineEdit::set_draw_control_chars(bool p_draw_control_chars) {
 	if (draw_control_chars != p_draw_control_chars) {
 		draw_control_chars = p_draw_control_chars;
-		if (menu && menu->get_item_index(MENU_DISPLAY_UCC) >= 0) {
-			menu->set_item_checked(menu->get_item_index(MENU_DISPLAY_UCC), draw_control_chars);
-		}
+		context_menu.set_draw_control_chars(draw_control_chars);
 		_shape();
 		queue_redraw();
 	}
@@ -2105,14 +2096,14 @@ bool LineEdit::is_context_menu_enabled() {
 }
 
 bool LineEdit::is_menu_visible() const {
-	return menu && menu->is_visible();
+	return context_menu.is_visible();
 }
 
 PopupMenu *LineEdit::get_menu() const {
-	if (!menu) {
+	if (!context_menu.is_initialized()) {
 		const_cast<LineEdit *>(this)->_generate_context_menu();
 	}
-	return menu;
+	return context_menu.get_main_menu();
 }
 
 void LineEdit::_editor_settings_changed() {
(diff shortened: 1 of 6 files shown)


AThousandShips:
You are removing the bound enum values, this breaks compatibility, they should be left as they are and extracted from the menu instead

KoBeWi:
> So simply copy the enum into this by:

Also they should be marked as deprecated.

AThousandShips:
Assuming we don't want them available any more, unsure if we do but without that we can't control the menu options by code

Note that the menu item helper class isn't exposed so we can't access these things through that

TheSofox:
Restored enums into TextEdit and LineEdit, including their bindings, and added the copying of their values from TextEditMenu as suggested by @AThousandShips.

scene/gui/text_edit.cpp
```diff
@@ -3742,132 +3733,132 @@ void TextEdit::paste_primary_clipboard(int p_caret) {
...
 
 bool TextEdit::is_menu_visible() const {
-	return menu && menu->is_visible();
+	return context_menu.is_visible();
 }
 
 void TextEdit::menu_option(int p_option) {
 	switch (p_option) {
-		case MENU_CUT: {
+		case TextEditMenu::MENU_CUT: {
```
AThousandShips, line 3748:
```suggestion
		case MENU_CUT: {
```
For all these, to keep it clean

scene/gui/line_edit.cpp
```diff
@@ -1967,128 +1956,128 @@ bool LineEdit::is_text_field() const {
 
 void LineEdit::menu_option(int p_option) {
 	switch (p_option) {
-		case MENU_CUT: {
+		case TextEditMenu::MENU_CUT: {
```
AThousandShips, line 1959:
```suggestion
		case MENU_CUT: {
```
Same here

AThousandShips (commented):
Haven't tested but looks good to me, will look further when I have the time

scene/gui/text_edit_menu.h
```diff
@@ -0,0 +1,95 @@
...
+	void popup_at(const Point2 &p_point, bool grab_focus = false);
+	bool has_focus();
+	void set_text_direction(Control::TextDirection p_text_direction);
+	void set_draw_control_chars(bool p_draw_control_chars);
+	bool is_visible() const;
+	bool is_initialized() const;
+	PopupMenu *get_main_menu() const;
+};
+
+#endif // TEXT_EDIT_MENU_H
```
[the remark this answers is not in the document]
TheSofox: Done

scene/gui/text_edit_menu.h
```diff
@@ -0,0 +1,95 @@
...
+	PopupMenu *menu = nullptr;
+	PopupMenu *menu_dir = nullptr;
+	PopupMenu *menu_ctl = nullptr;
+
+	void generate_menu(Control *control);
+	void add_option_selected_callback(Callable p_option_selected_callback);
+	void add_focus_changed_callback(Callable p_focus_changed_callback);
+	void update(Control *control, bool editable, bool selecting_enabled, bool shortcut_keys_enabled, Control::TextDirection text_direction, bool draw_control_chars, bool has_undo, bool has_redo);
+
+	void popup_at(const Point2 &p_point, bool grab_focus = false);
```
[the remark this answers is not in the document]
TheSofox: Done this for all (also removed Control argument from update function after realising it wasn't needed)

scene/gui/text_edit_menu.cpp
```diff
@@ -0,0 +1,213 @@
...
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "text_edit_menu.h"
+#include "core/input/input_map.h"
```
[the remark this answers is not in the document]
TheSofox: Separated.

scene/gui/text_edit.cpp
TheSofox: Done.

scene/gui/line_edit.cpp
TheSofox: Done

scene/gui/text_edit_menu.cpp
```diff
@@ -0,0 +1,213 @@
...
+	return menu && menu->is_visible();
+}
+
+bool TextEditMenu::is_initialized() const {
+	return menu;
+}
+
+PopupMenu *TextEditMenu::get_main_menu() const {
+	return menu;
+}
```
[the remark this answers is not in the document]
TheSofox: Done

scene/gui/text_edit_menu.h
```diff
@@ -0,0 +1,95 @@
...
+
+#ifndef TEXT_EDIT_MENU_H
+#define TEXT_EDIT_MENU_H
+
+#include "scene/gui/popup_menu.h"
+
+class TextEditMenu {
+private:
+	//void option_selected(int p_option);
+	//void focus_changed();
```
KoBeWi, line 39: What's this for?

scene/gui/text_edit_menu.h
```diff
@@ -0,0 +1,95 @@
...
+		MENU_INSERT_SHY,
+		MENU_MAX
+	};
+
+	PopupMenu *menu = nullptr;
+	PopupMenu *menu_dir = nullptr;
+	PopupMenu *menu_ctl = nullptr;
+
+	void generate_menu(Control *p_control);
+	void add_option_selected_callback(Callable p_option_selected_callback);
```
KoBeWi, line 82:
```suggestion
	void add_option_selected_callback(const Callable &p_option_selected_callback);
```
Same below.

scene/gui/text_edit_menu.cpp
```diff
@@ -0,0 +1,214 @@
...
+		}
+	}
+}
+
+bool TextEditMenu::has_focus() {
+	return menu && menu->has_focus();
+}
+
+void TextEditMenu::set_text_direction(Control::TextDirection p_text_direction) {
+	if (menu_dir) {
```
KoBeWi, line 190:
```suggestion
	if (!menu_dir) {
		return;
	}
```
Prefer early returns instead of nesting, when possible.

scene/gui/text_edit_menu.cpp
```diff
@@ -0,0 +1,214 @@
...
+	menu->add_separator();
+	menu->add_submenu_item(RTR("Text Writing Direction"), "DirMenu", MENU_SUBMENU_TEXT_DIR);
+	menu->add_separator();
+	menu->add_check_item(RTR("Display Control Characters"), MENU_DISPLAY_UCC);
+	menu->add_submenu_item(RTR("Insert Control Character"), "CTLMenu", MENU_SUBMENU_INSERT_UCC);
+}
+
+void TextEditMenu::add_option_selected_callback(Callable p_option_selected_callback) {
+	if (menu && menu_dir && menu_ctl) {
+		menu->connect("id_pressed", p_option_selected_callback);
```
KoBeWi, line 112: This should use SNAME too.

scene/gui/text_edit_menu.cpp
```diff
@@ -0,0 +1,214 @@
...
+	if (menu_dir) {
+		menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_INHERITED), p_text_direction == Control::TEXT_DIRECTION_INHERITED);
+		menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_AUTO), p_text_direction == Control::TEXT_DIRECTION_AUTO);
+		menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_LTR), p_text_direction == Control::TEXT_DIRECTION_LTR);
+		menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_RTL), p_text_direction == Control::TEXT_DIRECTION_RTL);
+	}
+}
+
+void TextEditMenu::set_draw_control_chars(bool p_draw_control_chars) {
+	if (menu && menu->get_item_index(MENU_DISPLAY_UCC) >= 0) {
```
KoBeWi, line 199: This was here originally, but I wonder what the second check is for. That option exists unconditionally.

scene/gui/text_edit_menu.h
TheSofox: Old code, will remove.

scene/gui/text_edit_menu.h
TheSofox: Done, will be in next push.

scene/gui/text_edit_menu.cpp
TheSofox: `ERR_FAIL_NULL();`  added to: `set_text_direction`, `popup_at`, `set_draw_control_chars`, `add_option_selected_callback` and `add_focus_changed_callback`. Was already in `update`
Will be in next push.

scene/gui/text_edit_menu.cpp
TheSofox: Added

scene/gui/text_edit_menu.cpp
TheSofox: Not sure about 2nd check, I assume just being extra cautious (replaced 1st check with ERR_FAIL_ like you suggested). I left it in there for safety.

TheSofox: Unless there's something you want me to do, I'll leave it for you to resolve. My best guess is the code is there in case someone removes that entry from the Popup menu (the menu is open for people to add entries to it, like filter option in the Filter LineEdit, so I assume people would theoretically be able to remove entries from it too).

scene/gui/text_edit_menu.cpp
TheSofox: Turns out adding it to `set_text_direction` gave errors because of the CodeEditor. I edited `TextEdit` so that it now generates the menu if it hasn't already before calling that function.

scene/gui/text_edit.cpp
```diff
@@ -3213,12 +3208,11 @@ void TextEdit::set_text_direction(Control::TextDirection p_text_direction) {
 		text.invalidate_font();
 		_update_placeholder();
 
-		if (menu_dir) {
-			menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_INHERITED), text_direction == TEXT_DIRECTION_INHERITED);
-			menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_AUTO), text_direction == TEXT_DIRECTION_AUTO);
-			menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_LTR), text_direction == TEXT_DIRECTION_LTR);
-			menu_dir->set_item_checked(menu_dir->get_item_index(MENU_DIR_RTL), text_direction == TEXT_DIRECTION_RTL);
+		if (!context_menu.is_initialized()) {
+			_generate_context_menu();
```
KoBeWi, line 3212: Instead of generating menu, you can skip calling `set_text_direction()` instead.

TheSofox: done

KoBeWi (approved):
Looks fine overall.

scene/gui/text_edit_menu.h
```diff
@@ -0,0 +1,93 @@
...
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef TEXT_EDIT_MENU_H
+#define TEXT_EDIT_MENU_H
+
+#include "scene/gui/popup_menu.h"
```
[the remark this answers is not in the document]
TheSofox: Trying to forward declare it gave me errors in `audio_stream_import_settings.h` and `node_3d_editor_plugin.h` so I think it's best to leave it as it is for now.

AThousandShips (approved):

scene/gui/text_edit_menu.h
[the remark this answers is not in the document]
AThousandShips: You need to restore the include in `LineEdit`, then you can safely forward declare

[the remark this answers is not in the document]
TheSofox: Tried that, then added forward-declaration for TextDirection to text_edit_menu.h. Now I'm getting errors like: '_get_menu_action_accelerator': unknown override specifier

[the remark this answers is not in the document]
TheSofox: Okay, found a way around it. Had to forward declare an enum, and then include `control.h` because I couldn't find a way to forward decleare a scoped enum (`TextDirection` specifically)

scene/gui/line_edit.h
```diff
@@ -32,46 +32,45 @@
...
+		MENU_INSERT_RLI = TextEditMenu::MENU_INSERT_RLI,
+		MENU_INSERT_FSI = TextEditMenu::MENU_INSERT_FSI,
+		MENU_INSERT_PDI = TextEditMenu::MENU_INSERT_PDI,
+		MENU_INSERT_ZWJ = TextEditMenu::MENU_INSERT_ZWJ,
+		MENU_INSERT_ZWNJ = TextEditMenu::MENU_INSERT_ZWNJ,
+		MENU_INSERT_WJ = TextEditMenu::MENU_INSERT_WJ,
+		MENU_INSERT_SHY = TextEditMenu::MENU_INSERT_SHY,
+		MENU_MAX = TextEditMenu::MENU_MAX
 	};
-
```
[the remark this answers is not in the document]
TheSofox: Restored.

Paulb23 (changes requested):
I don't entirely agree with this change, code can be too DRY and we need to be careful of that. As it introduces extra abstractions and hidden coupling.

This goes to #85624 as well.

As long as these two classes have existed there has always been a discussion about combining them in some way. Although they certainly look similar on the surface they do serve completely different use cases (Along with `RichTextLabel`). 

Before starting to make any changes in this direction, I'd suggest creating a detailed proposal of the direction / end goal here.

AThousandShips:
You can't expect someone to notice immediately and be around, everyone has their own schedule and might not notice, it's not Paulb23's fault

TheSofox:
When I posted this PR, I expected it to be evaluated on a conceptual level before we moved onto "discuss, fix and tweak a lot of various parts of it." There's over 10 reviews I had to respond to, for instance, all of which turned out to be pointless because the PR wasn't going to be accepted no matter how many tweaks I made to it.

I'm not saying its Paul's or anybody's fault, I recognise you all have your things to focus on, but meanwhile I followed a process that led to me needlessly wasting time and effort in a way that seemed totally avoidable.

KoBeWi:
Eh, I opened the issue after I noticed the code duplication while working on #85477. It was meant for discussing the solution, but after you opened a PR I just proceeded to review it normally, sorry 🙃

I still think extracting the menu has its merits and IMO could be done without a proposal. I never suggested merging TextEdit and LineEdit.

TheSofox:
Hey, after having some time to cool down, I recognise that while I was raising a genuine concern, I may not have expressed it in the most professional way. I also took Paul's comment as a flat rejection rather than "more work needed" which I apologise for.

I do agree that reducing repetition between `TextEdit` and `LineEdit` is a good idea. A quick diff compare of the two header files is very instructive. However, various solutions have their own complexities.

Issue #84609, opened 2023-11-08
ViewportTexture doesn't work in VisualShader


ilyabelow:
### Godot version

4.1.3

### System information

Godot v4.1.3.stable - Manjaro Linux #1 SMP PREEMPT_DYNAMIC Wed Oct 25 13:06:34 UTC 2023 - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3060 Laptop GPU () - 12th Gen Intel(R) Core(TM) i5-12500H (16 Threads)

### Issue description

When I try adding new ViewportTexture to a Texture2D node in visual shader editor, I get this message:

```
Can't create a ViewportTexture on this resource because it's not set as local to scene.
Please switch on the 'local to scene' property on it (and all resources containing it up to a node).
```

When I first found this problem, the shader was attached to a MeshInstance3D, which was set (along with all resources in the hierarchy) as "local to scene". Ordinary shaders, with ViewportTexture as a shader parameter, works as expected with the same setup. And if you create VisualShader just as a resource without attaching it to a mesh and checking "Local to Scene" checkbox, the result is the same message

Googling only gave [this](https://ask.godotengine.org/59479/how-do-you-use-viewporttexture-in-visual-shader) somewhat old report, so it seems that this is an old bug

P. S. Just in case I've tried this on master branch (3e7f638d7b, 08.11.23, same issue)

### Steps to reproduce

1. Create new VisualShader resource
2. Check "Local to Scene" (makes no difference)
3. Add "Texture2D" node to the shader
4. Choose "Texture" type and in source choose "New ViewportTexture"

### Minimal reproduction project

Minimal project contains only new_shader.tres with one node

[visualshader_issue.zip](https://github.com/godotengine/godot/files/13294843/visualshader_issue.zip)

clayjohn:
The MRP isn't enough to reproduce the reported issue. 

In your scene have you ensured that you checked local to scene on all resources containing the VisualShader? That means it needs to be checked on the VisualShader, the ShaderMaterial, and the Mesh. It is not enough to just check it on the VisualShader

ilyabelow:
> The MRP isn't enough to reproduce the reported issue.

What do you mean? It cannot be reproduced or the use case in MRP is invalid? If latter, than I guess the error message should be clearer I guess... and if the former, that's weird

[...]

PR #85723, opened 2023-12-03
Add a specific error when creating a ViewportTexture in a Texture2D node

ilyabelow:
Now, the "Can't create a ViewportTexture on this resource because it's not set as local to scene" error message is shown when you try to set a ViewportTexture as a texture in a Texture2D node in Visual shader

This error message is wrong - the user should not create such a texture because it would be saved to a file, which would not work and is already forbidden (and also because nodes are resources themselves which are not possible to set as local to scene, as discussed in the corresponding issue #84609)

In the PR I've added a check on parent resource type and a verbose error message which tells the user what the correct way to add a ViewportTexture is (with a Texture2DParameter node)

### DIFF
--- a/editor/editor_properties.cpp
+++ b/editor/editor_properties.cpp
@@ -56,6 +56,7 @@
 #include "scene/resources/font.h"
 #include "scene/resources/mesh.h"
 #include "scene/resources/packed_scene.h"
+#include "scene/resources/visual_shader_nodes.h"
 
 ///////////////////// Nil /////////////////////////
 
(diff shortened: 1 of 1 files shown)


editor/editor_properties.cpp
```diff
@@ -3196,6 +3197,13 @@ void EditorPropertyResource::_resource_changed(const Ref<Resource> &p_resource)
 	Ref<ViewportTexture> vpt = p_resource;
 	if (vpt.is_valid()) {
 		r = Object::cast_to<Resource>(get_edited_object());
+		if (r && r->get_class() == VisualShaderNodeTexture::get_class_static()) {
```
cvs0, line 3200: I would recommend you use `Object::cast_to` for safer type-checking.

editor/editor_properties.cpp
```diff
@@ -3196,6 +3197,13 @@ void EditorPropertyResource::_resource_changed(const Ref<Resource> &p_resource)
 	Ref<ViewportTexture> vpt = p_resource;
 	if (vpt.is_valid()) {
 		r = Object::cast_to<Resource>(get_edited_object());
+		if (r && Object::cast_to<VisualShaderNodeTexture>(r)) {
+			EditorNode::get_singleton()->show_warning(TTR("Can't create a ViewportTexture in a Texture2D node because the texture will not be bound to a scene.\nUse a Texture2DParameter node instead and set the texture in a \"Shader Parameters\" tab."));
```
AThousandShips, line 3201:
```suggestion
			EditorNode::get_singleton()->show_warning(TTR("Can't create a ViewportTexture in a Texture2D node because the texture will not be bound to a scene.\nUse a Texture2DParameter node instead and set the texture in the \"Shader Parameters\" tab."));
```
Unless there are multiple such tabs?

ilyabelow: Fair, there's only one such tab, fixed

Calinou (approved):

### Before

Screenshot_20231206_230848

### After

Screenshot_20231206_230814

Issue #84764, opened 2023-11-11
Can't select nodes behind a locked node in editor


nanodeath:
### Godot version

v4.1.3.stable.mono.official [f06b6836a]

### System information

Godot v4.1.3.stable.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1050 Ti with Max-Q Design (NVIDIA; 31.0.15.3640) - Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz (12 Threads)

### Issue description

When I lock a node in the editor, I expect to be able to click and select things behind it. Instead, clicks are ignored.

I have a game where your typical view (as the player and when editing using Godot) is an isometric camera and the player is inside a building. There's a ceiling, represented by a plane, with the "front" of the ceiling mesh is pointed down towards the player, leaving the backface to be culled/invisible. The player is able to carry a camera that renders into a viewport in the UI, but this handheld camera (and Camera3D) render from the first person perspective -- that is, below the ceiling, which is why it's there in the first place. That is, the ceiling should be invisible to the bird's eye camera, but not to the first-person camera.

Basically in the editor there's this giant invisible plane covering the entire visible map that, even when locked, blocks my attempts to select anything else in the scene. My workaround is to hide the plane and wire the `ready` event to the `show()` method as a one-shot, but this is crude.

### Steps to reproduce

1. Create any two meshes, with one mesh interposed between the other and the editor camera. In my case, one of the meshes is a large plane that sits between the rest of the scene and the editor camera.
2. Mark the mesh closer to the camera as Locked.
3. Try to click the further mesh.

In the repro project attached, notice that it's quite hard to actually select the "Player" node.

### Minimal reproduction project

[LockBlockRepro.zip](https://github.com/godotengine/godot/files/13326210/LockBlockRepro.zip) -- does not require the C# build.

kleonc:
Confirmed, same in v4.2.beta5.official [4c96e9676].

Note that as a workaround you can drag/box select, or <kbd>Alt</kbd> + right click to choose object to select from the menu:
![wtfLd3tpmf](https://github.com/godotengine/godot/assets/9283098/437dfa94-65cf-4b95-88f6-ae141d2fdbfa)

For comparison in 2D there's just clicking works fine:
![lFlDsTagrn](https://github.com/godotengine/godot/assets/9283098/218ebbec-d389-4bbc-9d8b-c5c3b349ee18)

[...]

PR #85740, opened 2023-12-04
Fix 3D Select Mode not moving clicked node

### LINKED ISSUE #85736: Click and drag on any node in 3D incorrectly moves selected node
### Godot version

3.5 stable, all 4.x stable versions

### System information

Godot v4.2.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated AMD Radeon RX 6700 XT (Advanced Micro Devices, Inc.; 31.0.21023.2010) - AMD Ryzen 7 5800X 8-Core Processor (16 Threads)

### Issue description

So this seems to be an old one, but got worse in 3.5 and 4.0, I couldn't find any other issues about this.

Basically when in Select Mode in the 3D editor, with a node selected, clicking and dragging moves the currently selected node as if you had clicked on it, regardless of which node is under the mouse cursor. Prior to 3.5 this only happened with locked nodes under the mouse specifically.

I believe the correct behavior would be to move the node under the cursor, unless it is locked in which case a box selection would be started. This would be consistent with the 2D editor and would avoid the situation in the 3D editor where you cannot perform box selection because the background is obscured by locked nodes.

Current behavior:
* No node selected, click anything and drag -> box selection
* Node selected, click on any node and drag -> selected node moves
* Node selected, click on background and drag -> box selection

GIF demonstration, the sphere is locked:
3_5_behavior

Prior to 3.5 the behavior worked like this:
* No node selected, click background and drag -> box selection
* No node selected, click unlocked or locked node and drag -> node moves if unlocked, no box selection
* Node selected, click on any **unlocked** node and drag -> node under mouse is selected and moves
* Node selected, click on any **locked** node and drag -> selected node moves
* Node selected, click on background and drag -> box selection
<details> <summary>GIF of 3.2 - 3.4 behavior (the sphere is locked)</summary>

3_2_behavior
</details>

It seems prior to 3.2 you could click and select locked nodes in the 3D editor and that seems to have prevented the bug from appearing at all.
<details><summary>GIF of original behavior</summary>

behavior_up_to_3_1_2
</details>

### Steps to reproduce

* Open the included Godot 4 project or create a blank project and add multiple objects into a 3d scene.
* Select one node
* While staying in Select mode, click the mouse in the 3D view over a non-selected node and drag
* (prior to 3.5) click the mouse over a locked node and drag

[TestClickDrag.zip](https://github.com/godotengine/godot/files/13541268/TestClickDrag.zip)

Cammymoop:
Fixes #85736 

Fixed:
Clicking and dragging a node when no nodes are selected selects and starts to translate it, rather than starting a box select.
With a node selected, clicking and dragging another node in select mode both selects and starts to translate the clicked node rather than the currently selected node.
Clicking and dragging on a locked node now starts a box select rather than translating the currently selected node (just as clicking and dragging on the background.

Clicking and dragging on (one of) the selected node(s) works as before.

3D Select Mode now works more or less like 2D Select Mode, and 3D Select Mode prior to 3.2
With the added bonus of not selecting locked nodes like was possible in 3.1

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -1859,8 +1859,14 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 						clicked = _select_ray(b->get_position());
 						selection_in_progress = true;
 
-						if (clicked.is_null()) {
-							//default to regionselect
+						bool can_select_clicked = false;
+						if (clicked.is_valid()) {
+							Node3D *clicked_node3d = Object::cast_to<Node3D>(ObjectDB::get_instance(clicked));
+							can_select_clicked = clicked_node3d && !_is_node_locked(clicked_node3d);
+						}
+
+						if (!can_select_clicked) {
+							// default to regionselect if nothing selectable was clicked
 							cursor.region_select = true;
 							cursor.region_begin = b->get_position();
 							cursor.region_end = b->get_position();
@@ -1984,13 +1990,11 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 			} else {
 				const bool movement_threshold_passed = _edit.original_mouse_pos.distance_to(_edit.mouse_pos) > 8 * EDSCALE;
 
-				// enable region-select if nothing has been selected yet or multi-select (shift key) is active
-				if (selection_in_progress && movement_threshold_passed) {
-					if (get_selected_count() == 0 || clicked_wants_append) {
-						cursor.region_select = true;
-						cursor.region_begin = _edit.original_mouse_pos;
-						clicked = ObjectID();
-					}
+				// enable region-select if multi-select (shift key) is active
+				if (selection_in_progress && movement_threshold_passed && clicked_wants_append) {
+					cursor.region_select = true;
+					cursor.region_begin = _edit.original_mouse_pos;
+					clicked = ObjectID();
 				}
 
 				if (cursor.region_select) {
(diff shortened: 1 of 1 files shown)


ryevdokimov (changes requested):
All tools can select now since: https://github.com/godotengine/godot/pull/86804

So, when you select->drag, in other modes the behavior does not match. 

https://github.com/godotengine/godot/assets/105675984/3c0998a0-560e-4216-bfab-9308bea043b4

ryevdokimov:
Also, I'm not sure if this is desired behavior. I could see situations where you want to box select, with unlocked nodes in the background where the select would be initiated. I think this makes too strong of an assumption of what the user wants, where a two-step process will prevent accidents. I would argue 2D is the one that needs to be changed.

Cammymoop:
Thanks for pointing that out, I'll fix that up.

> Also, I'm not sure if this is desired behavior. I could see situations where you want to box select, with unlocked nodes in the background where the select would be initiated. I think this makes too strong of an assumption of what the user wants, where a two-step process will prevent accidents. I would argue 2D is the one that needs to be changed.

This seems to be a regression from 3.1 and a worse regression from 3.2 onward, see the mentioned issue. I restored the old, presumably intended behavior, I think a case could be made to change the behavior (including the 2d editor) but that is separate from fixing this bug I think.

For what it's worth I support an (optional) setting to have select mode start a box select when dragging over another node instead of selecting then dragging as I would find that useful, not to the exclusion of the current behavior though, especially in 2d where I use the select mode for moving things around a lot. I just see that as a feature addition and primarily wanted to fix the bug.


As it currently stands it is simply intuitive and unhelpful and doesn't allow for starting a box select in that situation or several other more unambiguous situations. A pretty common example is where a large environment mesh exists in the scene but is locked for obvious reasons, and you want to use the select tool make box selections and move things around (along with axis constraining shortcuts this is pretty usable). 

The regression since 3.2 makes it even worse because in addition to having to hide the environment or use shift to make the box selections you also can't move an individual node without two clicks, and clicking to select it might put a gizmo under your cursor so you have to search for a neutral spot after the first click.

Cammymoop:
Translation now only starts on first click select + drag with Select or Move tool active, as in 2D editor.

ryevdokimov:
I can confirm that this PR does what it says it does and matches 2D behavior, without any issues I can find. I'll be interested to hear what the engine team thinks. @Calinou?

Cammymoop:
There is one difference from the 2D editor in that you can click on 2D nodes that are behind a locked node, but you cannot in 3D. (see also #84764 and my comment on the attached PR)

Repiteo:
Superseded by #87948

PR #85741, opened 2023-12-04
[WIP] Implement FFI for GDScript.

bruvzg:
Implement simple FFI for the GDScript using `libffi`.

- Adds `FFILibrary` class.
- Adds functions to get writable pointer of the `PackedByteArray` and copy data from a pointer to a `PackedByteArray`.
- C-struct argument/return types are encoded/decoded to the GDScript `Array`.

<details>
  <summary>Simple demo - Click to expand</summary>

```c++
#if defined(_MSC_VER)
    //  Microsoft 
    #define EXPORT __declspec(dllexport)
#elif defined(__GNUC__)
    //  GCC
    #define EXPORT __attribute__((visibility("default")))
#else
    #define EXPORT
#endif

#include <stdio.h>
#include <stdlib.h>
#include <string.h>

struct TestStruct {
    int a;
    long b;
    float c;
};

EXPORT struct TestStruct function_a(int p_a, float p_b) {
    struct TestStruct s;
    s.a = p_a;
    s.b = p_a * 10;
    s.c = p_b / 2.0;

    return s;
}

EXPORT float function_b(struct TestStruct p_s, int p_a) {
    float a = p_s.c + p_s.b * 10 + p_s.a * 100 - p_a;
    printf("==== %d %d %f %d = %f ====\n", p_s.a, p_s.b, p_s.c, p_a, a);
    return a;
}

EXPORT void function_c(int p_a, float p_b, char *p_str) {
    printf("==== %d %f %s ====\n", p_a, p_b, p_str);
    memcpy(p_str, "output test", 11);
}
```

```gdscript
func _test(path):
    # Open library.
    var lib = FFILibrary.new()
    lib.open(path, FFILibrary.ABI_DEFAULT)

    # Bind C-struct type.
    # args: name, element types...
    lib.bind_struct("TestStruct", FFILibrary.TYPE_SINT, FFILibrary.TYPE_SLONG, FFILibrary.TYPE_FLOAT)

    # Bind C-functions.
    # args: return type, name, argument types...
    var function_a = lib.bind_method("TestStruct", "function_a", FFILibrary.TYPE_SINT, FFILibrary.TYPE_FLOAT)
    var function_b = lib.bind_method(FFILibrary.TYPE_FLOAT, "function_b", "TestStruct", FFILibrary.TYPE_SINT)
    var function_c = lib.bind_method(FFILibrary.TYPE_VOID, "function_c", FFILibrary.TYPE_SINT, FFILibrary.TYPE_FLOAT, FFILibrary.TYPE_POINTER)

    # Call function A.
    print(function_a.call(44, 89.6))
    print("should be: ", [44, 440, 44.8])

    # Call function B.
    print(function_b.call([10, 20, 30.5], 233))
    print("should be: ", 997.5)

    # Call function C.
    var s = "Test String".to_utf8_buffer()
    s.push_back(0) # Add terminating zero.
    var ptr = s.unsafe_get_ptr() # Get writable pointer.
    function_c.call(10, 20.5, ptr)
    print(s.get_string_from_utf8())
    print("should be: ", "output test")
```

</details>

Tested (with the aforementioned demo only) on:
- macOS x86_64 and ARM64
- Windows (MSVC) x86_64 and x86_32
- Linux x86_64 and ARM64

TODO:
   - [ ] Documentation.
   - [ ] Test all primitive types and nested structures.
   - [ ] Build flags need to be carefully evaluated and tested on each platform and architecture, since it involves pre-processing and building assembly sources.
   - [ ] Evaluate thread safety.

### DIFF
--- a/modules/ffi/doc_classes/FFILibrary.xml
+++ b/modules/ffi/doc_classes/FFILibrary.xml
@@ -1,0 +1,161 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="FFILibrary" inherits="RefCounted" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../../doc/class.xsd">
+	<brief_description>
+	</brief_description>
+	<description>
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="bind_method" qualifiers="vararg">
+			<return type="Variant" />
+			<description>
+				Registers native function and returns [Callable].
+				Argument [code skip-lint]0[/code] should contain either [enum FFIType] value or registered structure [StringName], describing function return type.
+				Argument [code skip-lint]1[/code] should contain [StringName] with the function name.
+				Arguments [code skip-lint]2..n[/code] should contain either [enum FFIType] values or registered structure [StringName]s, describing function argument types.
+			</description>
+		</method>
+		<method name="bind_struct" qualifiers="vararg">
+			<return type="Variant" />
+			<description>
+				Registers native structure type and return [code]true[/code] if successful.
+				Argument [code skip-lint]0[/code] should contain [StringName] with the structure name.
+				Arguments [code skip-lint]1..n[/code] should contain either [enum FFIType] values or registered structure [StringName]s, describing structure element types.
+			</description>
+		</method>
+		<method name="get_method" qualifiers="const">
+			<return type="Callable" />
+			<param index="0" name="name" type="StringName" />
+			<description>
+				Returns [Callable] for the previously registered native function.
+			</description>
+		</method>
+		<method name="get_struct_offsets" qualifiers="const">
+			<return type="PackedInt64Array" />
+			<param index="0" name="name" type="StringName" />
+			<description>
+				Returns structure element offsets.
+			</description>
+		</method>
+		<method name="get_struct_size" qualifiers="const">
+			<return type="int" />
+			<param index="0" name="name" type="StringName" />
+			<description>
+				Returns structure size.
+			</description>
+		</method>
+		<method name="is_opened" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code], if library is open.
+			</description>
+		</method>
+		<method name="open">
+			<return type="int" enum="Error" />
+			<param index="0" name="path" type="String" />
+			<param index="1" name="abi" type="int" enum="FFILibrary.FFIABI" default="0" />
+			<description>
+				Opens dynamic library.
+			</description>
+		</method>
+	</methods>
+	<constants>
+		<constant name="ABI_DEFAULT" value="0" enum="FFIABI">
+			Default system ABI, based on current architecture and platform.
+		</constant>
+		<constant name="ABI_SYSV" value="1" enum="FFIABI">
+			System-V ABI.
+		</constant>
+		<constant name="ABI_VFP" value="2" enum="FFIABI">
+			ARM32 VFP ABI.
+		</constant>
+		<constant name="ABI_WIN64" value="3" enum="FFIABI">
+			x86_64 Windows MSVC ABI.
+		</constant>
+		<constant name="ABI_GNUW64" value="4" enum="FFIABI">
+			x86_64 Windows GNU ABI.
+		</constant>
+		<constant name="ABI_UNIX64" value="5" enum="FFIABI">
+			x86_64 Unix ABI.
+		</constant>
+		<constant name="ABI_STDCALL" value="6" enum="FFIABI">
+			x86_32 stdcall ABI.
+		</constant>
+		<constant name="ABI_THISCALL" value="7" enum="FFIABI">
+			x86_32 thiscall ABI.
+		</constant>
+		<constant name="ABI_FASTCALL" value="8" enum="FFIABI">
+			x86_32 fastcall ABI.
+		</constant>
+		<constant name="ABI_MS_CDECL" value="9" enum="FFIABI">
+			x86_32 cdecl ABI.
+		</constant>
+		<constant name="ABI_PASCAL" value="10" enum="FFIABI">
+			x86_32 Borland Pascal ABI.
+		</constant>
+		<constant name="ABI_REGISTER" value="11" enum="FFIABI">
+			x86_32 Delphi and Free Pascal ABI.
+		</constant>
+		<constant name="TYPE_VOID" value="0" enum="FFIType">
+			This cannot be used for argument types, only for return values.
+		</constant>
+		<constant name="TYPE_UINT8" value="1" enum="FFIType">
+			An unsigned, 8-bit integer type.
+		</constant>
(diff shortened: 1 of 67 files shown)


vnen:
This is interesting. I had an idea like this for quite some time but never got around to implementing something.

MarioLiebisch:
Neat, missed this so far and will have a look later.

IMO important concern: It might be reasonable to disable this functionality by default in the editor (and editor started projects; similar to how you have to enable/opt-in addons), simply because it could allow malicious code to do all kind of nasty stuff default tool scripts can't do.

Maybe decoupling this here, could be worth having a "Do you trust this source?" kind of banner once certain function calls or classes are used in a project?

bruvzg:
> IMO important concern: It might be reasonable to disable this functionality by default in the editor (and editor started projects; similar to how you have to enable/opt-in addons), simply because it could allow malicious code to do all kind of nasty stuff default tool scripts can't do.

GDScript can download and run executable or load GDExtension (which can have arbitrary native code) so FFI won't make it more insecure than it already is.

MarioLiebisch:
Hm, true.

Calinou:
> Maybe decoupling this here, could be worth having a "Do you trust this source?" kind of banner once certain function calls or classes are used in a project?

See https://github.com/godotengine/godot-proposals/issues/5010.

MarioLiebisch:
What's the status on this PR? Would love to potentially try to rebase and get a bit more into this feature over the Holidays, as there's sadly not too much activity here.

bruvzg:
> What's the status on this PR?

Needs testing on various platforms and architectures, it works fine on arm64 and x86-64, but extremely buggy on 32-bit x86, and not tested on any other architectures.

RedMser:
Seems like this would close https://github.com/godotengine/godot-proposals/issues/1917 (which also proposes a very similar looking API).

CsloudX:
> > What's the status on this PR?
> 
> Needs testing on various platforms and architectures, it works fine on arm64 and x86-64, but extremely buggy on 32-bit x86, and not tested on any other architectures.

So need and like this feature. can merge it. also it only works fine on arm64 and x86-64. in my work, most of dlls was 64bit. so it was very very useful. we can document it only works fine on arm64 and x86-64 now. full feature maybe fruther.

CsloudX:
@bruvzg hi, my friend. 
can this feature plan merge to 4.5 or 4.6. can't waitting!

AThousandShips:
We're in feature freeze, and this is a draft PR, so it will not be in 4.5

CsloudX:
I'm found a nice ffi plugin: https://github.com/gilzoide/cffi-gdextension and compare syntax:
current: 1. I like `lib.open(path, FFILibrary.ABI_DEFAULT)`, it let verything in my GDS. cffi plugin need add a ffilibrary resource.
cffi plugin: 
1. I like it define struct:
```
static var ExampleStruct = native_plugin_dll.define_struct("ExampleStruct", {
    "a": "int",
    "b": "int",
})

var example_struct = ExampleStruct.alloc()
    example_struct.a = 42
    var a = get_a.invoke(example_struct)
```
IMO, it more make sence for struct.

MarioLiebisch:
Usually too many things to do. 😔

As for syntax: I'd still prefer some decorators akin to .net's interop stuff.
I'm not a real fan of having to define everything twice, which seems to be the case with that cffi plugin?

modules/ffi/ffi_library.cpp
```diff
@@ -0,0 +1,669 @@
...
+
+	PackedInt64Array arr;
+	for (int i = 0; i < info->offsets.size(); i++) {
+		arr.push_back(info->offsets[i]);
+	}
+	return arr;
+}
+
+Callable FFILibrary::get_method(const StringName &p_name) const {
+	ERR_FAIL_COND_V_MSG(!library_handle, Callable(), "Library is not opened.");
```
AThousandShips:
```suggestion
	ERR_FAIL_NULL_V_MSG(library_handle, Callable(), "Library is not opened.");
```

modules/ffi/doc_classes/FFILibrary.xml
```diff
@@ -0,0 +1,161 @@
...
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="bind_method" qualifiers="vararg">
+			<return type="Variant" />
+			<description>
+				Registers native function and returns [Callable].
+				Argument [code skip-lint]0[/code] should contain either [enum FFIType] value or registered structure [StringName], describing function return type.
+				Argument [code skip-lint]1[/code] should contain [StringName] with the function name.
```
AThousandShips:
```suggestion
				Argument [code]1[/code] should contain a [StringName] with the function name.
```

modules/ffi/doc_classes/FFILibrary.xml
```diff
@@ -0,0 +1,161 @@
...
+				Arguments [code skip-lint]2..n[/code] should contain either [enum FFIType] values or registered structure [StringName]s, describing function argument types.
```
AThousandShips:
```suggestion
				Arguments [code]2..n[/code] should contain either [enum FFIType] values or registered structure [StringName]s, describing the function argument types.
```

modules/ffi/doc_classes/FFILibrary.xml
```diff
@@ -0,0 +1,161 @@
...
+			</description>
+		</method>
+		<method name="bind_struct" qualifiers="vararg">
+			<return type="Variant" />
+			<description>
+				Registers native structure type and return [code]true[/code] if successful.
```
AThousandShips:
```suggestion
				Registers a native structure type and returns [code]true[/code] if successful.
```

modules/ffi/doc_classes/FFILibrary.xml
```diff
@@ -0,0 +1,161 @@
...
+			<description>
+				Returns [code]true[/code], if library is open.
+			</description>
+		</method>
+		<method name="open">
+			<return type="int" enum="Error" />
+			<param index="0" name="path" type="String" />
+			<param index="1" name="abi" type="int" enum="FFILibrary.FFIABI" default="0" />
+			<description>
+				Opens dynamic library.
```
AThousandShips:
```suggestion
				Opens a dynamic library.
```

modules/ffi/doc_classes/FFILibrary.xml
```diff
@@ -0,0 +1,161 @@
...
+			An unsigned, 64-bit integer type.
+		</constant>
+		<constant name="TYPE_SINT64" value="8" enum="FFIType">
+			A signed, 64-bit integer type.
+		</constant>
+		<constant name="TYPE_FLOAT" value="9" enum="FFIType">
+			The C [code skip-lint]float[/code] type.
+		</constant>
+		<constant name="TYPE_DOUBLE" value="10" enum="FFIType">
+			The C [code skip-lint]double[/code] type.
```
AThousandShips:
```suggestion
			The C [code]double[/code] type.
```
For all except `float` and `int`

CsloudX:
Hi, friends, i'm coming again. need this feature very much. Any update of this? Can this merge to 4.7?

PR #83786, opened 2023-10-22
Fix ColorPicker committing extra action to history

aXu-AP:
Fixes #83642 Fixes #45186
ColorPicker committed value on both on `color_change` and `popup_closed` signals, so same value got submitted twice to undoredo.

Addionally, ColorPicker changed set value upon exit due to HTML value getting submitted. Since HTML color codes are 32-bit color, but Godot uses higher precision colors internally, this meant that color from HTML value differed from currently selected value, thus it was submitted resulting less precision on colors chosen via picker.
Fixed by checking if color is different with precision offered by 32-bit color.

EDIT: Combined #83787 into this pr since they are both needed to fix the problem.

PR #85749, opened 2023-12-04
Fix unnecessarily quantizing current color in color picker

### LINKED ISSUE #85743: Color changes when opening color picker
### Godot version

v4.3.dev.custom_build [d76c1d0e5]

### System information

Linux

### Issue description

When opening the color picker by clicking on a color property the color changes slightly.

### Steps to reproduce

* Click on a property with value `Color(.5, .5, .5) ` in the inspector
* Click away
* Open the color picker again
* See the color change to Color(0.502, 0.502, 0.502) in the "RAW" tab

[ColorChangeBug.zip](https://github.com/godotengine/godot/files/13545089/ColorChangeBug.zip)

bs-mwoerner:
Fixes #85743.

The color picker would update the current color to match the HTML string whenever the parsed value did not match the current color. Since Godot Colors have a 16 bit resolution, this would quantize the current value to the nearest HTML color even when no change was made. And since this happened whenever the HTML input lost focus, just opening and closing the color picker was enough to change a 16-bit value to the nearest 8-bit value.

With this PR, the HTML input only updates the current color when the HTML representation is different from the current color's HTML representation (and not whenever the HTML representation represents a different color than the current one).

### DIFF
--- a/scene/gui/color_picker.cpp
+++ b/scene/gui/color_picker.cpp
@@ -556,16 +556,17 @@ void ColorPicker::_html_submitted(const String &p_html) {
 		return;
 	}
 
-	const Color previous_color = color;
-	color = Color::from_string(p_html.strip_edges(), previous_color);
+	Color new_color = Color::from_string(p_html.strip_edges(), color);
 
 	if (!is_editing_alpha()) {
-		color.a = previous_color.a;
+		new_color.a = color.a;
 	}
 
-	if (color == previous_color) {
+	if (new_color.to_argb32() == color.to_argb32()) {
 		return;
 	}
+	color = new_color;
+
 	if (!is_inside_tree()) {
 		return;
 	}


akien-mga:
For the record, your commit seems not to be linked to your GitHub account. See: [Why are my commits linked to the wrong user?](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user) for more info.

KoBeWi:
#83786 seems to be also addressing this problem, in a simpler way.
CC @aXu-AP

bs-mwoerner:
> #83786 seems to be also addressing this problem, in a simpler way. CC @aXu-AP

Ah, yes, I missed that PR. `to_abgr32()` is probably the better way, although I think that other solution still requires a slight modification to actually work. I'll add my comments there.

aXu-AP:
Good work catching the error in my pr. I think this pr could be better fit to be modified, and I don't currently have time to work on it, so feel free to borrow from my solution.

bs-mwoerner:
> Good work catching the error in my pr. I think this pr could be better fit to be modified, and I don't currently have time to work on it, so feel free to borrow from my solution.

I will, thanks! And I only caught that error because I spent quite some time pulling my hair out when I was working on my own PR wondering how on Earth it still managed to get the wrong color value through to the caller despite me having breakpoints everywhere. 😅

scene/gui/color_picker.cpp
```diff
@@ -556,16 +556,16 @@ void ColorPicker::_html_submitted(const String &p_html) {
...
-		color.a = previous_color.a;
+		new_color.a = color.a;
 	}
 
-	if (color == previous_color) {
+	if (new_color.to_argb32() == color.to_argb32()) {
 		return;
 	}
+	color = new_color;
 	if (!is_inside_tree()) {
```
KoBeWi, line 569:
```suggestion
	color = new_color;

	if (!is_inside_tree()) {
```

KoBeWi (approved):

YuriSizov:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

PR #85761, opened 2023-12-04
Expand `error_macros.h` functionality

Repiteo:
A collection of minor improvements/additions to the error macro functions and defines:
- Implemented a dedicated index function for unsigned integers. Previously, it was feasible that the compared values wouldn't be properly transcribed if either value was a sufficently large `uint64_t`.
- Added explicit `int64_t`/`uint64_t` casts to the error index function calls.
- Ensure return values are printed with error, if provided. Previously this was only available for a portion of the macros that specified a return value.
- The comment above `DEV_ENABLED` is now applied to the defines, instead of the `#ifdef`. This mirrors the docstring of `GENERATE_TRAP`, allowing the comment to double as a docstring.

### DIFF
--- a/core/error/error_macros.cpp
+++ b/core/error/error_macros.cpp
@@ -115,14 +115,26 @@ void _err_print_error(const char *p_function, const char *p_file, int p_line, co
 }
 
 // Index errors. (All combinations of p_message as String or char*.)
-void _err_print_index_error(const char *p_function, const char *p_file, int p_line, int64_t p_index, int64_t p_size, const char *p_index_str, const char *p_size_str, const char *p_message, bool p_editor_notify, bool p_fatal) {
+void _err_print_index_error(const char *p_function, const char *p_file, int p_line, int64_t p_index, int64_t p_size, const char *p_index_str, const char *p_size_str, const char *p_message, bool p_editor_notify, bool p_fatal, const char *p_return) {
 	String fstr(p_fatal ? "FATAL: " : "");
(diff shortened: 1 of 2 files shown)


AThousandShips:
> Added explicit int64_t/uint64_t casts to the error index functions.

I'm not convinced this is a good idea, ensuring the user provides the correct types to the method feels like an important consideration to me

Repiteo:
> ensuring the user provides the correct types to the method feels like an important consideration to me

Hmm, that's a good point. I was thinking more in the context of ensuring the macros actually functioned in the first place, but the user might need that added agency. The value that'd benefit most from adding the explicit casts would be class enums, which would need both inputs externally cast to `int` or some equivalent to have `(m_index) < 0` function as expected

AThousandShips:
They are currently cast manually, so one could argue you save that effort, but at the same time you could get in trouble if you cast in the wrong context I'd say, the cast should be safe for smaller types but these checks aren't bound to specifically int64 either, unsure

Repiteo:
For now I'll revert it; I wouldn't want the other changes bogged down by it

Repiteo:
Oop, uh, nevermind. Turns out the uint function now makes things ambiguous. So either the issue of a `uint64_t` being improperly represented will just have to remain, or some kind of explicit casting is needed

AThousandShips:
I think you can do the casting when calling the method without any problems, as it could cast from some other type

PR #85762, opened 2023-12-04
Add `change_scene_to_node()`

### LINKED ISSUE #921: Rework the way of changing current scene
Games made with Godot.

This is actually an issue I used to have, but now it impacts me less. Still, I've seen some users having this problem and it was generally agreed that the system should change.

Right now we have 2 methods to change scene: `change_scene` and `change_scene_to`. Their naming is already confusing for some people and there are a few problems with changing scenes that these methods don't solve.

First, sometimes you want to setup the new scene. Right now it's impossible, because you don't have access to the newly changed scene in any way (`current_scene` doesn't point to the new scene immediately). So if you want to pass some data from old scene to the new one, you need to use singleton.

Second, sometimes you want to keep the old scene, e.g. when you open in-game menu and you want to restore game state later. Currently `change_scene` will automatically free the old scene.

The current workaround for both is changing the scenes manually, i.e. accessing scene tree, replacing root and setting the `current_scene`. I noticed that I actually do this quite frequently in my projects, which makes regular `change_scene` quite pointless, because it's not flexible enough. It might be alright for some prototypes or learning projects, but there aren't many games for which the current workflow is enough. Also official docs propose that you have a "main node" and keep the game state as a child of your node. This makes SceneTree root and `change_scene` methods even more useless xd

The system that was proposed originally (I mean after my original issue and then discussion about the now-closed PR) is to have three methods:
```
change_scene_path(path, keep_old = false)
change_scene_packed(packed_scene, keep_old = false)
change_scene_node(node, keep_old = false)
```
The first one takes path to the scene resource, the second one takes a PackedScene and instances it and the third one uses an already existing node. Each of them has additional argument `keep_old`, which if true, makes the old scene not being freed automatically. That scene would be returned from the method. And if `keep_old` is false, the method would free the old scene and return null.

*Alternatively, the methods could return the new scene. It has the advantage that all the methods would give you its reference (for third method you have the reference anyways, because you provide it as argument), but the downside is that you'd need to keep the old scene yourself.*

What I said above, but in C++ bruh
There was an old PR that added something similar: https://github.com/godotengine/godot/pull/26833

Well, it can be worked around quite easily:
```
var menu = preload("res://Menu.tscn").instance()
menu.game = self
get_node("/root").add_child(menu)
get_tree().current_scene = menu 
get_node("/root").remove_child(self)
```
But this code is not obvious to figure out for beginners in case they want to do either 1) or 2). Also this isn't really rarely used IMO.

Because it would replace the current `change_scene/to` methods.

### LINKED ISSUE #9243: Add change_scene_to_node() method

A 2D survivor game at mobile platform using c#

Godot has change_scene_to_file and change_scene_to_packed methods, change_scene_to_packed. Whyt not add method change_scene_to_node(const Ref<Node> &p_node). 
Now I use sub thread load and instance scene, and add_child, set_current_scene. but the replacement is raised error at Android Platform which too many
```
Condition 'p_elem->_root != this' is true
```
I guess the godot source code has another logic while change scene.

I can use the new method to change_scene

No

No

No

KoBeWi:
This PR adds a new method: `change_scene_to_node()`, which changes the scene to node. The code already existed, this change just extracts part of it into a new method. Thus calling `change_scene_to_file()` will call `change_scene_to_packed()`, which now calls `change_scene_to_node()`.
The new method is useful when you want to do some setup to the new scene before changing.
Also corrected the docs, which claimed that the new scene is instantiated after the old one is removed, which wasn't true.

Closes https://github.com/godotengine/godot-proposals/issues/921
Closes https://github.com/godotengine/godot-proposals/issues/9243

This does not implement the part where the old scene can be left intact, because it's not very useful. My current approach to the problem mentioned in the proposal is adding the menu scene on top of game and pausing the game; changing scenes temporarily is not a good practice.

### DIFF
--- a/doc/classes/SceneTree.xml
+++ b/doc/classes/SceneTree.xml
@@ -45,7 +45,21 @@
 			<description>
 				Changes the running scene to the one at the given [param path], after loading it into a [PackedScene] and creating a new instance.
 				Returns [constant OK] on success, [constant ERR_CANT_OPEN] if the [param path] cannot be loaded into a [PackedScene], or [constant ERR_CANT_CREATE] if that scene cannot be instantiated.
-				[b]Note:[/b] See [method change_scene_to_packed] for details on the order of operations.
+				[b]Note:[/b] See [method change_scene_to_node] for details on the order of operations.
+			</description>
+		</method>
+		<method name="change_scene_to_node">
+			<return type="int" enum="Error" />
+			<param index="0" name="node" type="Node" />
+			<description>
+				Changes the running scene to the provided [Node]. Useful when you want to set up the new scene before changing.
+				Returns [constant OK] on success, [constant ERR_INVALID_PARAMETER] if the [param node] is [code]null[/code], or [constant ERR_UNCONFIGURED] if the [param node] is already inside the scene tree.
+				[b]Note:[/b] Operations happen in the following order when [method change_scene_to_node] is called:
+				1. The current scene node is immediately removed from the tree. From that point, [method Node.get_tree] called on the current (outgoing) scene will return [code]null[/code]. [member current_scene] will be [code]null[/code] too, because the new scene is not available yet.
+				2. At the end of the frame, the formerly current scene, already removed from the tree, will be deleted (freed from memory) and then the new scene node will be added to the tree. [method Node.get_tree] and [member current_scene] will be back to working as usual.
+				This ensures that both scenes aren't running at the same time, while still freeing the previous scene in a safe way similar to [method Node.queue_free].
+				If you want to reliably access the new scene, await the [signal scene_changed] signal.
+				[b]Warning:[/b] After using this method, the [SceneTree] will take ownership of the node and will free it automatically when changing scene again. Any references you had to that node will become invalid.
 			</description>
 		</method>
 		<method name="change_scene_to_packed">
@@ -54,11 +68,7 @@
 			<description>
 				Changes the running scene to a new instance of the given [PackedScene] (which must be valid).
 				Returns [constant OK] on success, [constant ERR_CANT_CREATE] if the scene cannot be instantiated, or [constant ERR_INVALID_PARAMETER] if the scene is invalid.
-				[b]Note:[/b] Operations happen in the following order when [method change_scene_to_packed] is called:
-				1. The current scene node is immediately removed from the tree. From that point, [method Node.get_tree] called on the current (outgoing) scene will return [code]null[/code]. [member current_scene] will be [code]null[/code], too, because the new scene is not available yet.
-				2. At the end of the frame, the formerly current scene, already removed from the tree, will be deleted (freed from memory) and then the new scene will be instantiated and added to the tree. [method Node.get_tree] and [member current_scene] will be back to working as usual.
-				This ensures that both scenes aren't running at the same time, while still freeing the previous scene in a safe way similar to [method Node.queue_free].
-				If you want to reliably access the new scene, await the [signal scene_changed] signal.
+				[b]Note:[/b] See [method change_scene_to_node] for details on the order of operations.
 			</description>
 		</method>
 		<method name="create_timer">
--- a/scene/main/scene_tree.cpp
+++ b/scene/main/scene_tree.cpp
@@ -1690,6 +1690,13 @@ Error SceneTree::change_scene_to_packed(const Ref<PackedScene> &p_scene) {
 	Node *new_scene = p_scene->instantiate();
 	ERR_FAIL_NULL_V(new_scene, ERR_CANT_CREATE);
 
+	return change_scene_to_node(new_scene);
+}
+
+Error SceneTree::change_scene_to_node(Node *p_node) {
+	ERR_FAIL_NULL_V_MSG(p_node, ERR_INVALID_PARAMETER, "Can't change to a null node. Use unload_current_scene() if you wish to unload it.");
+	ERR_FAIL_COND_V_MSG(p_node->is_inside_tree(), ERR_UNCONFIGURED, "The new scene node can't already be inside scene tree.");
+
 	// If called again while a change is pending.
 	if (pending_new_scene_id.is_valid()) {
 		Node *pending_new_scene = ObjectDB::get_instance<Node>(pending_new_scene_id);
@@ -1707,7 +1714,7 @@ Error SceneTree::change_scene_to_packed(const Ref<PackedScene> &p_scene) {
 	}
 	DEV_ASSERT(!current_scene);
 
-	pending_new_scene_id = new_scene->get_instance_id();
+	pending_new_scene_id = p_node->get_instance_id();
 	return OK;
 }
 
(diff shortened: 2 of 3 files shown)


doc/classes/SceneTree.xml
```diff
@@ -42,7 +42,19 @@
...
+				[b]Note:[/b] See [method change_scene_to_node] for details on the order of operations.
+			</description>
+		</method>
+		<method name="change_scene_to_node">
+			<return type="int" enum="Error" />
+			<param index="0" name="node" type="Node" />
+			<description>
+				Changes the running scene to the provided [Node]. Useful when you want to setup the new scene before changing.
+				Returns [constant OK] on success, [constant ERR_INVALID_PARAMETER] if the [param node] is [code]null[/code], or [constant ERR_UNCONFIGURED] if the [param node] is already inside scene tree.
+				[b]Note:[/b] Operations happen in the following order when [method change_scene_to_packed] is called:
```
Summersay415, line 54: Here should be change_scene_to_node, not change_scene_to_packed

scene/main/scene_tree.cpp
```diff
@@ -1411,6 +1411,13 @@ Error SceneTree::change_scene_to_packed(const Ref<PackedScene> &p_scene) {
 	Node *new_scene = p_scene->instantiate();
 	ERR_FAIL_NULL_V(new_scene, ERR_CANT_CREATE);
 
+	return change_scene_to_node(new_scene);
+}
+
+Error SceneTree::change_scene_to_node(Node *p_node) {
+	ERR_FAIL_NULL_V_MSG(p_node, ERR_INVALID_PARAMETER, "Can't change to a null node. Use unload_current_scene() if you wish to unload it.");
+	ERR_FAIL_COND_V_MSG(p_node->is_inside_tree(), ERR_UNCONFIGURED, "The new scene node can't be already inside scene tree.");
```
AThousandShips:
```suggestion
	ERR_FAIL_COND_V_MSG(p_node->is_inside_tree(), ERR_UNCONFIGURED, "The new scene node can't be inside scene tree already.");
```
Or
```suggestion
	ERR_FAIL_COND_V_MSG(p_node->is_inside_tree(), ERR_UNCONFIGURED, "The new scene node can't already be inside scene tree.");
```

RadiantUwU:
Any activity?

radiantgurl (approved):
Looks good to me!

Mahi:
Any news on this? Seems like such a trivial yet useful change.

Mahi:
Any news on this? Seems like a very trivial yet useful feature.

Meanwhile for anyone else with this issue, use this workaround:
```
var tree = get_tree()
tree.current_scene.queue_free()
tree.root.add_child(my_scene)
tree.current_scene = my_scene
```

But it's 1. cumbersome to use in multiple places, 2. very beginner unfriendly, and 3. inconsistent with other `change_scene_to_*` methods.

kleonc:
Consider the user doing something like (simplified):
```gdscript
var node_a := Node.new()
var node_b := Node.new()

get_tree().change_scene_to_node(node_a)
get_tree().change_scene_to_node(node_b)
```
It potentially might be surprising for the user that `node_a` was freed as a result. The docs should probably explicitly note/warn that by passing a node to `change_scene_to_node` you kinda give up the ownership and it will be auto freed even in a scenario like above (`node_a` never became the current scene, it was a pending new scene only temporarily).

For comparison `change_scene_to_packed` is fine without such note/warn as it's not the user who instantiates the scene.

Besides that code LGTM, the addition is trivial.

KoBeWi:
Something like this?
`[b]Warning:[/b] After using this method, the [SceneTree] will take ownership of the node and will free it automatically when changing scene again. Any references you had to that node may become invalid.`

kleonc:
> Something like this? `[b]Warning:[/b] After using this method, the [SceneTree] will take ownership of the node and will free it automatically when changing scene again. Any references you had to that node may become invalid.`

Yeah, LGTM, sends the message. 👍

kleonc (approved):
LGTM, the addition makes sense, and looking at the reactions there's a demand for it.

doc/classes/SceneTree.xml
```diff
@@ -45,7 +45,21 @@
...
 				Returns [constant OK] on success, [constant ERR_CANT_OPEN] if the [param path] cannot be loaded into a [PackedScene], or [constant ERR_CANT_CREATE] if that scene cannot be instantiated.
-				[b]Note:[/b] See [method change_scene_to_packed] for details on the order of operations.
+				[b]Note:[/b] See [method change_scene_to_node] for details on the order of operations.
+			</description>
+		</method>
+		<method name="change_scene_to_node">
+			<return type="int" enum="Error" />
+			<param index="0" name="node" type="Node" />
+			<description>
+				Changes the running scene to the provided [Node]. Useful when you want to setup the new scene before changing.
```
AThousandShips:
```suggestion
				Changes the running scene to the provided [Node]. Useful when you want to set up the new scene before changing.
```
"Setup" is a noun

doc/classes/SceneTree.xml
```diff
@@ -45,7 +45,21 @@
...
+				Returns [constant OK] on success, [constant ERR_INVALID_PARAMETER] if the [param node] is [code]null[/code], or [constant ERR_UNCONFIGURED] if the [param node] is already inside scene tree.
```
AThousandShips:
```suggestion
				Returns [constant OK] on success, [constant ERR_INVALID_PARAMETER] if the [param node] is [code]null[/code], or [constant ERR_UNCONFIGURED] if the [param node] is already inside the scene tree.
```

doc/classes/SceneTree.xml
```diff
@@ -45,7 +45,21 @@
...
+				[b]Note:[/b] Operations happen in the following order when [method change_scene_to_node] is called:
+				1. The current scene node is immediately removed from the tree. From that point, [method Node.get_tree] called on the current (outgoing) scene will return [code]null[/code]. [member current_scene] will be [code]null[/code], too, because the new scene is not available yet.
```
AThousandShips:
```suggestion
				1. The current scene node is immediately removed from the tree. From that point, [method Node.get_tree] called on the current (outgoing) scene will return [code]null[/code]. [member current_scene] will be [code]null[/code] too, because the new scene is not available yet.
```

AThousandShips (approved):

doc/classes/SceneTree.xml
```diff
@@ -45,7 +45,21 @@
...
+				[b]Note:[/b] Operations happen in the following order when [method change_scene_to_node] is called:
+				1. The current scene node is immediately removed from the tree. From that point, [method Node.get_tree] called on the current (outgoing) scene will return [code]null[/code]. [member current_scene] will be [code]null[/code] too, because the new scene is not available yet.
+				2. At the end of the frame, the formerly current scene, already removed from the tree, will be deleted (freed from memory) and then the new scene node will be added to the tree. [method Node.get_tree] and [member current_scene] will be back to working as usual.
+				This ensures that both scenes aren't running at the same time, while still freeing the previous scene in a safe way similar to [method Node.queue_free].
+				If you want to reliably access the new scene, await the [signal scene_changed] signal.
+				[b]Warning:[/b] After using this method, the [SceneTree] will take ownership of the node and will free it automatically when changing scene again. Any references you had to that node may become invalid.
```
Ivorforce:
```suggestion
				[b]Warning:[/b] After using this method, the [SceneTree] will take ownership of the node and will free it automatically when changing scene again. Any references you had to that node will become invalid.
```

Ivorforce (approved):
According to the principle of least astonishment, I would have expected the interface to be designed in this way anyway — and there seems to interest in it too.

Code looks good to me.

Repiteo:
Thanks!

os-architect:
Wow. I just literally ran into a situation where I want something like this. Glad to know it's been merged. Talk about timing 😅

Thanks!!

PR #81654, opened 2023-09-14
Fix GodotPhysicsDirectBodyState2D get_contact_impulse to return current physics step impulse

KlugeNico:
Issue:
GodotPhysicsDirectBodyState2D get_contact_impulse() function returns wrong value. Initial it returns just a Vector2.ZERO.
The problem is, that not the final impulse sum is used, but the value generated in pre_solve().

Solution:
The only solution I found was to add a function that's called after solve(). It's called post_solve and is optional for stuff that must be done in a GodotConstraint2D after being solved. Here we can add the collision contact info with all required information, instead of adding it in pre_solve(), where some information (the impulse sum) is missing.

Same issue for 3D:
https://github.com/godotengine/godot/pull/73569

PR #85764, opened 2023-12-04
Fix 3D physics zero impulse contact reporting

### LINKED ISSUE #73541: PhysicsDirectBodyState3D.get_contact_impulse() returns zero vector on first frame of collision
### Godot version

v4.0.rc2.official [d2699dc7a]

### System information

EndeavourOS, Vulkan

### Issue description

Calling the `get_contact_impulse()` method on the PhysicsDirectState3D object provided in the `_integrate_forces()` method of `RigidBody3D` returns `Vector3.ZERO` for contacts that were created in the current frame. I would expect this method to return the impulse applied to the rigidbody by the contact, even during the first frame.
This becomes an issue particularly with bouncy objects as they often only have one frame of contact which results in no impulse being reported at all.

I dug a bit in the physics code and it seems that the contacts of a rigidbody are not updated after `pre_solve` in `GodotBodyPair3D` ran which seems suspicious.

### Steps to reproduce

1. set up a scene with a rigidbody with a collider and something it can collide with (I tested both rigidbodies and staticbodies for this)
2. Turn on `contact_monitor` and set `max_contacts_reported` to a reasonable value
3. Add a script to the rigidbody that implements `_integrate_forces()` and call `state.get_contact_impulse()`, for example print all impulses for all contacts
4. Observe that during the first frame after a new contact is created a zero vector is returned. This is more obvious when a physics material with bounciness 1.0 is added to the rigidbody.

[impulse_test.zip](https://github.com/godotengine/godot/files/10772643/impulse_test.zip)

darthLeviN:
There is a bug where collisions go reported with zero impulse when a contact starts.
This is because the impulse is calculated AFTER the data gets submited.
I moved the submission to the solve step. **I don't know if this interferes with multithreading. please test.**

Fixes https://github.com/godotengine/godot/issues/73541

My Minimal reproduction project for the bug :
[collision_reporting_1.zip](https://github.com/godotengine/godot/files/13552058/collision_reporting_1.zip)

Never mind the late assert. the only important one is for first contact.

Relates to https://github.com/godotengine/godot/pull/81654

Should i follow the same approach as this pull request instead? Does my method create duplicate collisions?

### DIFF
--- a/servers/physics_3d/godot_body_pair_3d.cpp
+++ b/servers/physics_3d/godot_body_pair_3d.cpp
@@ -409,20 +409,9 @@ bool GodotBodyPair3D::pre_solve(real_t p_step) {
 
 		c.acc_impulse -= j_vec;
 
-		// contact query reporting...
-
-		if (A->can_report_contacts() || B->can_report_contacts()) {
-			Vector3 crB = B->get_angular_velocity().cross(c.rB) + B->get_linear_velocity();
-			Vector3 crA = A->get_angular_velocity().cross(c.rA) + A->get_linear_velocity();
-
-			if (A->can_report_contacts()) {
-				A->add_contact(global_A + offset_A, -c.normal, depth, shape_A, crA, global_B + offset_A, shape_B, B->get_instance_id(), B->get_self(), crB, c.acc_impulse);
-			}
-
-			if (B->can_report_contacts()) {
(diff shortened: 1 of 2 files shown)


Repiteo:
Superseded by #87948

Repiteo:
Whoops, wrong PR

PR #85767, opened 2023-12-04
[iOS] Add changing of OS UI settings at runtime

### LINKED ISSUE #8556: Allow iOS UI settings to be updated at runtime

An iOS mobile game or app.

`display/window/iOS/hide_home_indicator`, `display/window/iOS/hide_status_bar`, and `display/window/iOS/suppress_ui_gesture` cannot be changed at runtime.

The reason this currently fails is that [setNeedsUpdateOfHomeIndicatorAutoHidden()](https://developer.apple.com/documentation/uikit/uiviewcontroller/2887509-setneedsupdateofhomeindicatoraut), [setNeedsStatusBarAppearanceUpdate()](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621354-setneedsstatusbarappearanceupdat), and [setNeedsUpdateOfScreenEdgesDeferringSystemGestures()](https://developer.apple.com/documentation/uikit/uiviewcontroller/2887507-setneedsupdateofscreenedgesdefer) need to be called for iOS to update their state, which never occurs if you change them via `ProjectSettings.set_setting()`. A function in DisplayServer could be added to rectify this.

For the sake of not clogging the docs with three iOS exclusive functions, there are a couple of options.

Currently, in ProjectSettings the settings are laid out like this.
`display/window/iOS/hide_home_indicator` - true/false (makes the home bar fade away and become invisible, but the gesture still operates the same way)
`display/window/iOS/hide_status_bar` - true/false (hides the clock, wifi indicator, battery, etc)
`display/window/iOS/suppress_ui_gesture` - true/false (makes swiping away from the app require two swipes up on the nav bar instead of just one)

However, `hide_home_indicator` and `suppress_ui_gesture` are mutually exclusive. If you enable both, `hide_home_indicator` overrides `suppress_ui_gesture`. It is possible that an enum might be better due to this. I have also added some missing niche cases that might be desired, that were previously impossible.

This would result in the following...

### enum IOSUIGestureState:
  
  **IOS_DEFAULT** = 0
  
  Shows home indicator and a single swipe closes the app. Control Center and Notification Center only require a single swipe to open.
  
  **SUPPRESS_GESTURE_ALL** = 1
  
  Shows the home indicator and two swipes are required to close to app. Control Center and Notification Center also require two swipes.

  **INDICATOR_HIDE** = 2
  
  Hides the home indicator and a single swipe closes the app. Control Center and Notification Center only requires a single swipe to open.
  
  **SUPPRESS_GESTURE_HOME_ONLY** = 3
  
  Shows the home indicator and two swipes are required to close to app. Control Center and Notification Center only require a single swipe to open.

  **SUPPRESS_GESTURE_STATUS_ONLY** = 4
  
  Shows the home indicator and a single swipe closes the app. Control Center and Notification Center require two swipes.

  **SUPPRESS_GESTURE_STATUS_ONLY_INDICATOR_HIDE** = 5
  
  Hides the home indicator and a single swipe closes the app. Control Center and Notification Center require two swipes.
  

### Function:
DisplayServer.ios_set_ui_state(**IOSUIGestureState** ios_gesture_state, **bool** hide_status_bar, **float** status_bar_fade_time)

--

The enum and function titles probably need adjustment but that's what I could think of off the top of my head.

Impossible, as `ProjectSettings.set_setting()` does not trigger the needed iOS system functions.

Impossible.

ztc0611:
Closes https://github.com/godotengine/godot-proposals/issues/8556.

Allows changing the state of gesture suppression and visibility of the status bar/home indicator at runtime. The use of this could be to only show or hide these functions in certain contexts, such as making only one swipe be required to leave a main menu or pause screen, but while the game is running it requires a double swipe. This is currently impossible as it requires signals to be sent to the OS that were not implemented before.

**2025-03-21 Edit**: An example I posed in the thread for the potential use of this is Roblox, which allows standard iOS gestures while in the game selection part of the app, but once you enter a game it enables the mode that requires a double swipe.

Another example includes a shareable code. While on the screen for sharing, you can swipe out with a single swipe. Once the user dismisses that screen, it goes back to needing a double swipe.

### DIFF
--- a/doc/classes/DisplayServer.xml
+++ b/doc/classes/DisplayServer.xml
@@ -853,6 +853,13 @@
 				[b]Note:[/b] Currently only implemented on Android and iOS. On other platforms, [code]screen_get_usable_rect(SCREEN_OF_MAIN_WINDOW)[/code] will be returned as a fallback. See also [method screen_get_usable_rect].
 			</description>
 		</method>
+		<method name="get_gesture_state" qualifiers="const">
+			<return type="int" enum="DisplayServer.UIGestureState" />
+			<description>
+				Returns the state of the UI gestures currently in effect.
+				[b]Note:[/b] This method is only implemented on iOS.
+			</description>
+		</method>
 		<method name="get_keyboard_focus_screen" qualifiers="const">
 			<return type="int" />
 			<description>
@@ -883,6 +890,13 @@
 			<param index="0" name="rect" type="Rect2" />
 			<description>
 				Returns the index of the screen that overlaps the most with the given rectangle. Returns [code]-1[/code] if the rectangle doesn't overlap with any screen or has no area.
+			</description>
+		</method>
+		<method name="get_status_bar_appearance" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns if the status bar is currently hidden. This can be used to modify [code]display/window/ios/hide_status_bar[/code] at runtime.
+				[b]Note:[/b] This method is only implemented on iOS.
 			</description>
 		</method>
 		<method name="get_swap_cancel_ok">
@@ -1828,6 +1842,14 @@
 				[b]Note:[/b] On iOS, this method has no effect if [member ProjectSettings.display/window/handheld/orientation] is not set to [constant SCREEN_SENSOR].
 			</description>
 		</method>
+		<method name="set_gesture_state">
+			<return type="void" />
+			<param index="0" name="ios_gesture_set" type="int" enum="DisplayServer.UIGestureState" />
+			<description>
+				Sets the UI gesture state. This can be used to modify [code]display/window/ios/ui_gesture_state[/code] at runtime.
+				[b]Note:[/b] This method is only implemented on iOS.
+			</description>
+		</method>
 		<method name="set_hardware_keyboard_connection_change_callback">
 			<return type="void" />
 			<param index="0" name="callable" type="Callable" />
@@ -1850,6 +1872,15 @@
 			<description>
 				Sets the window icon (usually displayed in the top-left corner) in the operating system's [i]native[/i] format. The file at [param filename] must be in [code].ico[/code] format on Windows or [code].icns[/code] on macOS. By using specially crafted [code].ico[/code] or [code].icns[/code] icons, [method set_native_icon] allows specifying different icons depending on the size the icon is displayed at. This size is determined by the operating system and user preferences (including the display scale factor). To use icons in other formats, use [method set_icon] instead.
 				[b]Note:[/b] Requires support for [constant FEATURE_NATIVE_ICON].
+			</description>
+		</method>
+		<method name="set_status_bar_appearance">
+			<return type="void" />
+			<param index="0" name="hide_status_bar" type="bool" default="false" />
+			<param index="1" name="status_bar_fade_time" type="float" default="0.2" />
+			<description>
+				Sets the appearance of the status bar. [param status_bar_fade_time] allows changing speed at which it appears or disappears, [code]0[/code] is instant.
+				[b]Note:[/b] This method is only implemented on iOS.
 			</description>
 		</method>
 		<method name="set_system_theme_change_callback">
@@ -2906,6 +2937,24 @@
 		</constant>
 		<constant name="INVALID_INDICATOR_ID" value="-1">
 			The ID that refers to a nonexistent application status indicator.
+		</constant>
+		<constant name="IOS_DEFAULT_GESTURE_STATE" value="0" enum="UIGestureState">
+			Shows home indicator and a single swipe closes the app. Control Center and Notification Center only require a single swipe to open.
+		</constant>
+		<constant name="IOS_SUPPRESS_GESTURE_ALL" value="1" enum="UIGestureState">
+			Shows the home indicator and two swipes are required to close to app. Control Center and Notification Center also require two swipes.
+		</constant>
+		<constant name="IOS_HIDE_GESTURE_INDICATOR" value="2" enum="UIGestureState">
+			Hides the home indicator and a single swipe closes the app. Control Center and Notification Center only requires a single swipe to open.
+		</constant>
+		<constant name="IOS_SUPPRESS_GESTURE_HOME_ONLY" value="3" enum="UIGestureState">
+			Shows the home indicator and two swipes are required to close to app. Control Center and Notification Center only require a single swipe to open.
+		</constant>
+		<constant name="IOS_SUPPRESS_GESTURE_STATUS_ONLY" value="4" enum="UIGestureState">
+			Shows the home indicator and a single swipe closes the app. Control Center and Notification Center require two swipes.
+		</constant>
+		<constant name="IOS_SUPPRESS_GESTURE_STATUS_ONLY_HIDE_GESTURE_INDICATOR" value="5" enum="UIGestureState">
+			Hides the home indicator and a single swipe closes the app. Control Center and Notification Center require two swipes.
 		</constant>
 		<constant name="SCREEN_LANDSCAPE" value="0" enum="ScreenOrientation">
 			Default landscape orientation.
--- a/doc/classes/ProjectSettings.xml
+++ b/doc/classes/ProjectSettings.xml
@@ -926,15 +926,13 @@
 		<member name="display/window/ios/allow_high_refresh_rate" type="bool" setter="" getter="" default="true">
 			If [code]true[/code], iOS devices that support high refresh rate/"ProMotion" will be allowed to render at up to 120 frames per second.
 		</member>
-		<member name="display/window/ios/hide_home_indicator" type="bool" setter="" getter="" default="true">
-			If [code]true[/code], the home indicator is hidden automatically. This only affects iOS devices without a physical home button.
-		</member>
 		<member name="display/window/ios/hide_status_bar" type="bool" setter="" getter="" default="true">
-			If [code]true[/code], the status bar is hidden while the app is running.
+			If [code]true[/code], the status bar is hidden while the app is running. This can be changed at runtime with [method DisplayServer.set_status_bar_appearance].
 		</member>
-		<member name="display/window/ios/suppress_ui_gesture" type="bool" setter="" getter="" default="true">
-			If [code]true[/code], it will require two swipes to access iOS UI that uses gestures.
-			[b]Note:[/b] This setting has no effect on the home indicator if [code]hide_home_indicator[/code] is [code]true[/code].
+		<member name="display/window/ios/ui_gesture_state" type="int" setter="" getter="" default="1">
+			Sets which gestures require a swipe, double swipe, or hiding home indicator on iOS.
+			See [enum DisplayServer.UIGestureState] for possible values and how they affect the behavior of your application.
+			This can be changed at runtime with [method DisplayServer.set_gesture_state].
 		</member>
 		<member name="display/window/per_pixel_transparency/allowed" type="bool" setter="" getter="" default="false">
 			If [code]true[/code], allows per-pixel transparency for the window background. This affects performance, so leave it on [code]false[/code] unless you need it. See also [member display/window/size/transparent] and [member rendering/viewport/transparent_background].
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -2732,9 +2732,8 @@ Error Main::setup(const char *execpath, int argc, char *argv[], bool p_second_ph
 	}
 
 	GLOBAL_DEF("display/window/ios/allow_high_refresh_rate", true);
-	GLOBAL_DEF("display/window/ios/hide_home_indicator", true);
-	GLOBAL_DEF("display/window/ios/hide_status_bar", true);
-	GLOBAL_DEF("display/window/ios/suppress_ui_gesture", true);
+	GLOBAL_DEF_BASIC(PropertyInfo(Variant::INT, "display/window/ios/ui_gesture_state", PROPERTY_HINT_ENUM, "Allow All,Suppress All,Hide Home Indicator,Suppress Home Only,Suppress Status Only,Suppress Status Only and Hide Home"), 1);
+	GLOBAL_DEF_BASIC("display/window/ios/hide_status_bar", true);
 
 #ifndef _3D_DISABLED
 	// XR project settings.
--- a/platform/ios/display_server_ios.h
+++ b/platform/ios/display_server_ios.h
@@ -81,6 +81,9 @@ class DisplayServerIOS : public DisplayServer {
 
 	int virtual_keyboard_height = 0;
 
+	UIGestureState ui_gesture_state = IOS_DEFAULT_GESTURE_STATE;
+	bool status_bar_hidden = false;
+
 	void perform_event(const Ref<InputEvent> &p_event);
 
 	void initialize_tts() const;
@@ -162,6 +165,13 @@ public:
 	virtual Rect2i screen_get_usable_rect(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
 	virtual int screen_get_dpi(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
 	virtual float screen_get_scale(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
+
+	virtual UIGestureState get_gesture_state() const override;
+	virtual void set_gesture_state(UIGestureState p_ios_gesture_set) override;
+
+	virtual bool get_status_bar_appearance() const override;
+	virtual void set_status_bar_appearance(bool p_hide_status_bar = false, float p_status_bar_fade_time = 0.2) override;
+
 	virtual float screen_get_refresh_rate(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
 
 	virtual Vector<DisplayServer::WindowID> get_window_list() const override;
--- a/platform/ios/display_server_ios.mm
+++ b/platform/ios/display_server_ios.mm
@@ -65,6 +65,9 @@ DisplayServerIOS::DisplayServerIOS(const String &p_rendering_driver, WindowMode 
 	}
 	native_menu = memnew(NativeMenu);
 
+	status_bar_hidden = GLOBAL_GET("display/window/ios/hide_status_bar");
+	ui_gesture_state = static_cast<UIGestureState>((int)GLOBAL_GET("display/window/ios/ui_gesture_state"));
+
 #if defined(RD_ENABLED)
 	rendering_context = nullptr;
 	rendering_device = nullptr;
@@ -551,6 +554,35 @@ int DisplayServerIOS::screen_get_dpi(int p_screen) const {
 		}
 		default:
 			return 72;
+	}
+}
+
+DisplayServerIOS::UIGestureState DisplayServerIOS::get_gesture_state() const {
+	return ui_gesture_state;
+}
+
+void DisplayServerIOS::set_gesture_state(UIGestureState p_ios_gesture_set) {
+	if (ui_gesture_state == p_ios_gesture_set) {
+		return;
+	}
+	ui_gesture_state = p_ios_gesture_set;
+	[AppDelegate.viewController setNeedsUpdateOfHomeIndicatorAutoHidden];
+	[AppDelegate.viewController setNeedsUpdateOfScreenEdgesDeferringSystemGestures];
+}
+
+bool DisplayServerIOS::get_status_bar_appearance() const {
+	return status_bar_hidden;
+}
+
+void DisplayServerIOS::set_status_bar_appearance(bool p_hide_status_bar, float p_status_bar_fade_time) {
+	status_bar_hidden = p_hide_status_bar;
+	if (p_status_bar_fade_time > 0) {
+		[UIView animateWithDuration:p_status_bar_fade_time
+						 animations:^{
+							 [AppDelegate.viewController setNeedsStatusBarAppearanceUpdate];
+						 }];
+	} else {
+		[AppDelegate.viewController setNeedsStatusBarAppearanceUpdate];
 	}
 }
 
--- a/platform/ios/view_controller.mm
+++ b/platform/ios/view_controller.mm
@@ -166,6 +166,7 @@
 	[self displayLoadingOverlay];
 
 	[self setNeedsUpdateOfScreenEdgesDeferringSystemGestures];
+	[self setNeedsStatusBarAppearanceUpdate];
 }
 
 - (void)observeKeyboard {
@@ -224,15 +225,31 @@
 	[[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
-// MARK: Orientation
-
 - (UIRectEdge)preferredScreenEdgesDeferringSystemGestures {
-	if (GLOBAL_GET("display/window/ios/suppress_ui_gesture")) {
-		return UIRectEdgeAll;
+	int ui_gesture_state;
+	if (!DisplayServerIOS::get_singleton()) {
+		ui_gesture_state = GLOBAL_GET("display/window/ios/ui_gesture_state");
 	} else {
-		return UIRectEdgeNone;
+		ui_gesture_state = DisplayServerIOS::get_singleton()->get_gesture_state();
 	}
+
+	switch (ui_gesture_state) {
+		case DisplayServer::IOS_DEFAULT_GESTURE_STATE:
+		case DisplayServer::IOS_HIDE_GESTURE_INDICATOR:
+			return UIRectEdgeNone;
+		case DisplayServer::IOS_SUPPRESS_GESTURE_ALL:
+			return UIRectEdgeAll;
+		case DisplayServer::IOS_SUPPRESS_GESTURE_HOME_ONLY:
+			return UIRectEdgeBottom;
+		case DisplayServer::IOS_SUPPRESS_GESTURE_STATUS_ONLY:
+		case DisplayServer::IOS_SUPPRESS_GESTURE_STATUS_ONLY_HIDE_GESTURE_INDICATOR:
+			return UIRectEdgeTop;
+	}
+
+	return UIRectEdgeNone;
 }
+
+// MARK: Orientation
 
 - (BOOL)shouldAutorotate {
 	if (!DisplayServerIOS::get_singleton()) {
@@ -281,7 +298,14 @@
 }
 
 - (BOOL)prefersStatusBarHidden {
-	if (GLOBAL_GET("display/window/ios/hide_status_bar")) {
+	bool prefers_hidden;
+	if (!DisplayServerIOS::get_singleton()) {
+		prefers_hidden = GLOBAL_GET("display/window/ios/hide_status_bar");
+	} else {
+		prefers_hidden = DisplayServerIOS::get_singleton()->get_status_bar_appearance();
+	}
+
+	if (prefers_hidden) {
 		return YES;
 	} else {
 		return NO;
@@ -289,10 +313,19 @@
 }
 
 - (BOOL)prefersHomeIndicatorAutoHidden {
-	if (GLOBAL_GET("display/window/ios/hide_home_indicator")) {
-		return YES;
+	int ui_gesture_state;
+	if (!DisplayServerIOS::get_singleton()) {
+		ui_gesture_state = GLOBAL_GET("display/window/ios/ui_gesture_state");
 	} else {
-		return NO;
+		ui_gesture_state = DisplayServerIOS::get_singleton()->get_gesture_state();
+	}
+
+	switch (ui_gesture_state) {
+		case DisplayServer::IOS_HIDE_GESTURE_INDICATOR:
+		case DisplayServer::IOS_SUPPRESS_GESTURE_STATUS_ONLY_HIDE_GESTURE_INDICATOR:
+			return YES;
+		default:
+			return NO;
 	}
 }
 
--- a/servers/display_server.cpp
+++ b/servers/display_server.cpp
@@ -571,6 +571,24 @@ float DisplayServer::screen_get_scale(int p_screen) const {
 	return 1.0f;
 }
 
+DisplayServer::UIGestureState DisplayServer::get_gesture_state() const {
+	WARN_PRINT("iOS UI not supported by this display server.");
+	return IOS_DEFAULT_GESTURE_STATE;
+}
+
+void DisplayServer::set_gesture_state(UIGestureState p_ios_gesture_set) {
+	WARN_PRINT("iOS UI not supported by this display server.");
+}
+
+bool DisplayServer::get_status_bar_appearance() const {
+	WARN_PRINT("iOS UI not supported by this display server.");
+	return false;
+}
+
+void DisplayServer::set_status_bar_appearance(bool p_hide_status_bar, float p_status_bar_fade_time) {
+	WARN_PRINT("iOS UI not supported by this display server.");
+}
+
 bool DisplayServer::is_touchscreen_available() const {
 	return Input::get_singleton() && Input::get_singleton()->is_emulating_touch_from_mouse();
 }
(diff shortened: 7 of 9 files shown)


servers/display_server.h
```diff
@@ -306,6 +306,9 @@ class DisplayServer : public Object {
 		}
 		return scale;
 	}
+
+	virtual void ios_set_ui_state(bool hide_home_indicator, bool suppress_bottom_ui_gesture, bool suppress_top_ui_gesture, bool hide_status_bar = true, float status_bar_fade_time = 0.2);
```
AThousandShips, line 310:
```suggestion
	virtual void ios_set_ui_state(bool p_hide_home_indicator, bool p_suppress_bottom_ui_gesture, bool p_suppress_top_ui_gesture, bool p_hide_status_bar = true, float p_status_bar_fade_time = 0.2);
```
Prefix arguments with `p_`

platform/ios/display_server_ios.h
```diff
@@ -153,6 +153,9 @@ class DisplayServerIOS : public DisplayServer {
 	virtual Rect2i screen_get_usable_rect(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
 	virtual int screen_get_dpi(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
 	virtual float screen_get_scale(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
+
+	virtual void ios_set_ui_state(bool hide_home_indicator, bool suppress_bottom_ui_gesture, bool suppress_top_ui_gesture, bool hide_status_bar = true, float status_bar_fade_time = 0.2) override;
```
AThousandShips, line 157:
```suggestion
	virtual void ios_set_ui_state(bool p_hide_home_indicator, bool p_suppress_bottom_ui_gesture, bool p_suppress_top_ui_gesture, bool p_hide_status_bar = true, float p_status_bar_fade_time = 0.2) override;
```

servers/display_server.cpp
```diff
@@ -411,6 +411,10 @@ float DisplayServer::screen_get_scale(int p_screen) const {
 	return 1.0f;
 };
 
+void DisplayServer::ios_set_ui_state(bool hide_home_indicator, bool suppress_bottom_ui_gesture, bool suppress_top_ui_gesture, bool hide_status_bar, float status_bar_fade_time) {
```
AThousandShips, line 414:
```suggestion
void DisplayServer::ios_set_ui_state(bool p_hide_home_indicator, bool p_suppress_bottom_ui_gesture, bool p_suppress_top_ui_gesture, bool p_hide_status_bar, float p_status_bar_fade_time) {
```

platform/ios/display_server_ios.mm
```diff
@@ -442,6 +442,25 @@
 	}
 }
 
+void DisplayServerIOS::ios_set_ui_state(bool hide_home_indicator, bool suppress_bottom_ui_gesture, bool suppress_top_ui_gesture, bool hide_status_bar, float status_bar_fade_time) {
+
```
AThousandShips, line 446:
```suggestion
```
No space at the beginning of blocks

platform/ios/view_controller.mm
```diff
@@ -220,16 +221,26 @@ - (void)dealloc {
 	[[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
-// MARK: Orientation
-
 - (UIRectEdge)preferredScreenEdgesDeferringSystemGestures {
-	if (GLOBAL_GET("display/window/ios/suppress_ui_gesture")) {
-		return UIRectEdgeAll;
-	} else {
+
```
AThousandShips, line 225:
```suggestion
```

main/main.cpp
```diff
@@ -2144,9 +2144,10 @@ Error Main::setup(const char *execpath, int argc, char *argv[], bool p_second_ph
 	}
 
 	GLOBAL_DEF("display/window/ios/allow_high_refresh_rate", true);
-	GLOBAL_DEF("display/window/ios/hide_home_indicator", true);
-	GLOBAL_DEF("display/window/ios/hide_status_bar", true);
-	GLOBAL_DEF("display/window/ios/suppress_ui_gesture", true);
+	GLOBAL_DEF_BASIC("display/window/ios/hide_home_indicator", false);
```
AThousandShips, line 2147: This default value is changed, this does affect compatibility, what's the reason for his change?

ztc0611: Hiding the home bar indicator and suppressing it are mutually exclusive, it was either a misunderstanding or my inexperience that caused the defaults to be this way in the first place. This weirdness of mutually exclusive options being possible to enable is fixed by Option 2.

platform/ios/display_server_ios.mm
```diff
@@ -442,6 +442,24 @@
 	}
 }
 
+void DisplayServerIOS::ios_set_ui_state(bool p_hide_home_indicator, bool p_suppress_bottom_ui_gesture, bool p_suppress_top_ui_gesture, bool p_hide_status_bar, float p_status_bar_fade_time) {
+	ProjectSettings::get_singleton()->set_setting("display/window/ios/hide_home_indicator", hide_home_indicator);
```
bruvzg, line 446: No platform is setting project settings, so it probably should read project setting values to the DisplayServerIOS variables on startup and only update them.

servers/display_server.cpp
```diff
@@ -698,6 +702,7 @@ void DisplayServer::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("screen_get_scale", "screen"), &DisplayServer::screen_get_scale, DEFVAL(SCREEN_OF_MAIN_WINDOW));
 	ClassDB::bind_method(D_METHOD("is_touchscreen_available"), &DisplayServer::is_touchscreen_available, DEFVAL(SCREEN_OF_MAIN_WINDOW));
 	ClassDB::bind_method(D_METHOD("screen_get_max_scale"), &DisplayServer::screen_get_max_scale);
+	ClassDB::bind_method(D_METHOD("ios_set_ui_state", "hide_home_indicator", "suppress_bottom_ui_gesture", "suppress_top_ui_gesture", "hide_status_bar", "status_bar_fade_time"), &DisplayServer::ios_set_ui_state, DEFVAL(true), DEFVAL(0.2));
```
bruvzg, line 705: Having a separate `get/set` methods for each value is probably better.

bruvzg: Also, documentation should be updated after adding new methods, see [Updating class reference when working on the engine](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html#updating-class-reference-when-working-on-the-engine).

bruvzg (commented):
This is fine in general, but it should not update ProjectSettings, this is something only editor do. Use getters/setters for the values instead.

ztc0611:
So you prefer option 3?

ztc0611:
@bruvzg  I'm trying to get this adjusted to follow your reviews and replace with `set/get`, but Im not sure which is the better way to go for this (2 or 3). Thanks!

ztc0611:
Refactored to look more like Option 2, because Option 3 looked extremely messy with `set/get`.

Here is the updated test project for the current version.
[Test.zip](https://github.com/godotengine/godot/files/13665534/Test.zip)

Here is the commit that existed prior to my most recent pushes.
https://github.com/ztc0611/godot/tree/mobile-setui-option3

servers/display_server.cpp
```diff
@@ -849,6 +873,13 @@ void DisplayServer::_bind_methods() {
 	BIND_CONSTANT(MAIN_WINDOW_ID);
 	BIND_CONSTANT(INVALID_WINDOW_ID);
 
+	BIND_CONSTANT(IOS_DEFAULT);
+	BIND_CONSTANT(IOS_SUPPRESS_GESTURE_ALL);
+	BIND_CONSTANT(IOS_INDICATOR_HIDE);
+	BIND_CONSTANT(IOS_SUPPRESS_GESTURE_HOME_ONLY);
+	BIND_CONSTANT(IOS_SUPPRESS_GESTURE_STATUS_ONLY);
+	BIND_CONSTANT(IOS_SUPPRESS_GESTURE_STATUS_ONLY_INDICATOR_HIDE);
```
bruvzg, line 881:
```suggestion
	BIND_ENUM_CONSTANT(IOS_DEFAULT);
	BIND_ENUM_CONSTANT(IOS_SUPPRESS_GESTURE_ALL);
	BIND_ENUM_CONSTANT(IOS_INDICATOR_HIDE);
	BIND_ENUM_CONSTANT(IOS_SUPPRESS_GESTURE_HOME_ONLY);
	BIND_ENUM_CONSTANT(IOS_SUPPRESS_GESTURE_STATUS_ONLY);
	BIND_ENUM_CONSTANT(IOS_SUPPRESS_GESTURE_STATUS_ONLY_INDICATOR_HIDE);
```

servers/display_server.h
```diff
@@ -308,6 +308,22 @@ class DisplayServer : public Object {
...
+		IOS_DEFAULT,
+		IOS_SUPPRESS_GESTURE_ALL,
+		IOS_INDICATOR_HIDE,
+		IOS_SUPPRESS_GESTURE_HOME_ONLY,
+		IOS_SUPPRESS_GESTURE_STATUS_ONLY,
+		IOS_SUPPRESS_GESTURE_STATUS_ONLY_INDICATOR_HIDE,
+	};
+
+	virtual IOSUIGestureState ios_get_gesture_state() const;
+	virtual void ios_set_gesture_state(IOSUIGestureState ios_gesture_set);
```
AThousandShips, line 322:
```suggestion
	virtual void ios_set_gesture_state(IOSUIGestureState p_ios_gesture_set);
```
Prefix arguments with `p_`

servers/display_server.h
```diff
@@ -308,6 +308,22 @@ class DisplayServer : public Object {
...
+
+	virtual bool ios_get_status_bar_appearance() const;
+	virtual void ios_set_status_bar_appearance(bool hide_status_bar = false, float status_bar_fade_time = 0.2);
```
AThousandShips, line 325:
```suggestion
	virtual void ios_set_status_bar_appearance(bool p_hide_status_bar = false, float p_status_bar_fade_time = 0.2);
```

servers/display_server.cpp
```diff
@@ -416,6 +416,24 @@ float DisplayServer::screen_get_scale(int p_screen) const {
 	return 1.0f;
 };
 
+DisplayServer::IOSUIGestureState DisplayServer::ios_get_gesture_state() const {
+	WARN_PRINT("iOS UI not supported by this display server.");
+	return DisplayServer::IOS_DEFAULT;
+}
+
+void DisplayServer::ios_set_gesture_state(DisplayServer::IOSUIGestureState ios_gesture_set) {
```
AThousandShips, line 424:
```suggestion
void DisplayServer::ios_set_gesture_state(IOSUIGestureState p_ios_gesture_set) {
```
Not needed

servers/display_server.cpp
```diff
@@ -416,6 +416,24 @@ float DisplayServer::screen_get_scale(int p_screen) const {
...
+	WARN_PRINT("iOS UI not supported by this display server.");
+}
+
+bool DisplayServer::ios_get_status_bar_appearance() const {
+	WARN_PRINT("iOS UI not supported by this display server.");
+	return false;
+}
+
+void DisplayServer::ios_set_status_bar_appearance(bool hide_status_bar, float status_bar_fade_time) {
```
AThousandShips, line 433:
```suggestion
void DisplayServer::ios_set_status_bar_appearance(bool p_hide_status_bar, float p_status_bar_fade_time) {
```

platform/ios/display_server_ios.h
```diff
@@ -155,6 +158,13 @@ class DisplayServerIOS : public DisplayServer {
 	virtual Rect2i screen_get_usable_rect(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
 	virtual int screen_get_dpi(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
 	virtual float screen_get_scale(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
+
+	virtual DisplayServer::IOSUIGestureState ios_get_gesture_state() const override;
+	virtual void ios_set_gesture_state(DisplayServer::IOSUIGestureState ios_gesture_set) override;
```
AThousandShips, line 163:
```suggestion
	virtual void ios_set_gesture_state(IOSUIGestureState p_ios_gesture_set) override;
```

platform/ios/display_server_ios.h
```diff
@@ -155,6 +158,13 @@ class DisplayServerIOS : public DisplayServer {
...
+
+	virtual bool ios_get_status_bar_appearance() const override;
+	virtual void ios_set_status_bar_appearance(bool hide_status_bar = false, float status_bar_fade_time = 0.2) override;
```
AThousandShips, line 166:
```suggestion
	virtual void ios_set_status_bar_appearance(bool p_hide_status_bar = false, float p_status_bar_fade_time = 0.2) override;
```

platform/ios/display_server_ios.mm
```diff
@@ -454,6 +454,32 @@
 	}
 }
 
+DisplayServerIOS::IOSUIGestureState DisplayServerIOS::ios_get_gesture_state() const {
+	return ui_gesture_state;
+}
+
+void DisplayServerIOS::ios_set_gesture_state(DisplayServer::IOSUIGestureState ios_gesture_set) {
```
AThousandShips, line 461:
```suggestion
void DisplayServerIOS::ios_set_gesture_state(IOSUIGestureState p_ios_gesture_set) {
```

platform/ios/display_server_ios.mm
```diff
@@ -454,6 +454,32 @@
...
+	ui_gesture_state = ios_gesture_set;
+	[AppDelegate.viewController setNeedsUpdateOfHomeIndicatorAutoHidden];
+	[AppDelegate.viewController setNeedsUpdateOfScreenEdgesDeferringSystemGestures];
+}
+
+bool DisplayServerIOS::ios_get_status_bar_appearance() const {
+	return status_bar_hidden;
+}
+
+void DisplayServerIOS::ios_set_status_bar_appearance(bool hide_status_bar, float status_bar_fade_time) {
```
AThousandShips, line 471:
```suggestion
void DisplayServerIOS::ios_set_status_bar_appearance(bool p_hide_status_bar, float p_status_bar_fade_time) {
```

servers/display_server.cpp
```diff
@@ -709,6 +727,12 @@ void DisplayServer::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("screen_get_scale", "screen"), &DisplayServer::screen_get_scale, DEFVAL(SCREEN_OF_MAIN_WINDOW));
 	ClassDB::bind_method(D_METHOD("is_touchscreen_available"), &DisplayServer::is_touchscreen_available, DEFVAL(SCREEN_OF_MAIN_WINDOW));
 	ClassDB::bind_method(D_METHOD("screen_get_max_scale"), &DisplayServer::screen_get_max_scale);
+
+	ClassDB::bind_method(D_METHOD("ios_set_gesture_state", "ios_gesture_set"), &DisplayServer::ios_set_gesture_state);
+	ClassDB::bind_method(D_METHOD("ios_get_gesture_state"), &DisplayServer::ios_get_gesture_state);
+	ClassDB::bind_method(D_METHOD("ios_set_status_bar_appearance", "hide_status_bar", "status_bar_fade_time"), &DisplayServer::ios_set_status_bar_appearance, DEFVAL(true), DEFVAL(0.2));
```
AThousandShips, line 733:
```suggestion
	ClassDB::bind_method(D_METHOD("ios_set_status_bar_appearance", "hide_status_bar", "status_bar_fade_time"), &DisplayServer::ios_set_status_bar_appearance, DEFVAL(false), DEFVAL(0.2));
```
Match defaults

doc/classes/DisplayServer.xml
```diff
@@ -849,6 +849,37 @@
...
+			<description>
+				Returns if the status bar is currently hidden. This can be used to modify `display/window/ios/hide_status_bar` at runtime.
+				[b]Note:[/b] This method is only implemented on iOS.
+			</description>
+		</method>
+		<method name="ios_set_gesture_state">
+			<return type="void" />
+			<param index="0" name="ios_gesture_set" type="int" enum="DisplayServer.IOSUIGestureState" />
+			<description>
+				Sets the UI gesture state. This can be used to modify `display/window/ios/ui_gesture_state` at runtime.
```
AThousandShips, line 870:
```suggestion
				Sets the UI gesture state. This can be used to modify [code]display/window/ios/ui_gesture_state[/code] at runtime.
```

doc/classes/DisplayServer.xml
```diff
@@ -849,6 +849,37 @@
...
+				[b]Note:[/b] This method is only implemented on iOS.
+			</description>
+		</method>
+		<method name="ios_set_status_bar_appearance">
+			<return type="void" />
+			<param index="0" name="hide_status_bar" type="bool" default="true" />
+			<param index="1" name="status_bar_fade_time" type="float" default="0.2" />
+			<description>
+				Sets the appearance of the status bar. `status_bar_fade_time` allows changing speed at which it appears or disappears, `0` is instant.
```
AThousandShips, line 879:
```suggestion
				Sets the appearance of the status bar. [param status_bar_fade_time] allows changing speed at which it appears or disappears, [code]0[/code] is instant.
```

doc/classes/DisplayServer.xml
```diff
@@ -849,6 +849,37 @@
...
+			<return type="int" enum="DisplayServer.IOSUIGestureState" />
+			<description>
+				Returns the state of the UI gestures currently in effect.
+				[b]Note:[/b] This method is only implemented on iOS.
+			</description>
+		</method>
+		<method name="ios_get_status_bar_appearance" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns if the status bar is currently hidden. This can be used to modify `display/window/ios/hide_status_bar` at runtime.
```
AThousandShips, line 862:
```suggestion
				Returns if the status bar is currently hidden. This can be used to modify [code]display/window/ios/hide_status_bar[/code] at runtime.
```

doc/classes/ProjectSettings.xml
```diff
@@ -766,15 +766,13 @@
...
-			If [code]true[/code], the status bar is hidden while the app is running.
+			If [code]true[/code], the status bar is hidden while the app is running. This can be changed at runtime with `DisplayServer.ios_set_status_bar_appearance`.
 		</member>
-		<member name="display/window/ios/suppress_ui_gesture" type="bool" setter="" getter="" default="true">
-			If [code]true[/code], it will require two swipes to access iOS UI that uses gestures.
-			[b]Note:[/b] This setting has no effect on the home indicator if [code]hide_home_indicator[/code] is [code]true[/code].
+		<member name="display/window/ios/ui_gesture_state" type="int" setter="" getter="" default="0">
+			Sets which gestures require a swipe, double swipe, or hiding home indicator on iOS.
+			See [enum DisplayServer.IOSUIGestureState] for possible values and how they affect the behavior of your application.
+			This can be changed at runtime with `DisplayServer.ios_set_gesture_state`.
```
AThousandShips, line 775:
```suggestion
			This can be changed at runtime with [method DisplayServer.ios_set_gesture_state].
```

doc/classes/ProjectSettings.xml
```diff
@@ -766,15 +766,13 @@
 		<member name="display/window/ios/allow_high_refresh_rate" type="bool" setter="" getter="" default="true">
 			If [code]true[/code], iOS devices that support high refresh rate/"ProMotion" will be allowed to render at up to 120 frames per second.
 		</member>
-		<member name="display/window/ios/hide_home_indicator" type="bool" setter="" getter="" default="true">
-			If [code]true[/code], the home indicator is hidden automatically. This only affects iOS devices without a physical home button.
-		</member>
 		<member name="display/window/ios/hide_status_bar" type="bool" setter="" getter="" default="true">
-			If [code]true[/code], the status bar is hidden while the app is running.
+			If [code]true[/code], the status bar is hidden while the app is running. This can be changed at runtime with `DisplayServer.ios_set_status_bar_appearance`.
```
AThousandShips, line 770:
```suggestion
			If [code]true[/code], the status bar is hidden while the app is running. This can be changed at runtime with [method DisplayServer.ios_set_status_bar_appearance].
```

servers/display_server.cpp
```diff
@@ -416,6 +416,24 @@ float DisplayServer::screen_get_scale(int p_screen) const {
 	return 1.0f;
 };
 
+DisplayServer::IOSUIGestureState DisplayServer::ios_get_gesture_state() const {
+	WARN_PRINT("iOS UI not supported by this display server.");
+	return DisplayServer::IOS_DEFAULT;
```
AThousandShips, line 421:
```suggestion
	return IOS_DEFAULT;
```

platform/ios/display_server_ios.h
```diff
@@ -155,6 +158,13 @@ class DisplayServerIOS : public DisplayServer {
 	virtual Rect2i screen_get_usable_rect(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
 	virtual int screen_get_dpi(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
 	virtual float screen_get_scale(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
+
+	virtual DisplayServer::IOSUIGestureState ios_get_gesture_state() const override;
```
AThousandShips, line 162:
```suggestion
	virtual IOSUIGestureState ios_get_gesture_state() const override;
```

platform/ios/display_server_ios.h
```diff
@@ -79,6 +79,9 @@ class DisplayServerIOS : public DisplayServer {
 
 	int virtual_keyboard_height = 0;
 
+	DisplayServerIOS::IOSUIGestureState ui_gesture_state = static_cast<DisplayServerIOS::IOSUIGestureState>((int)GLOBAL_GET("display/window/ios/ui_gesture_state"));
+	bool status_bar_hidden = GLOBAL_GET("display/window/ios/hide_status_bar");
```
AThousandShips, line 83: You can't declare this in the header with `GLOBAL_GET`

ztc0611: What would be the preferred method? The default state on launch needs to be the same as the project setting or it could cause buggy behavior.

AThousandShips: In the constructor

AThousandShips: Like is done for tts

ztc0611:
> Please use the batch feature instead of adding each one

Sorry 🤦‍♂️ I looked around for one but didn't see anything, and then didn't see this notif before I had done all of them manually. Will use the correct method next time.

platform/ios/display_server_ios.h
ztc0611: On it

platform/ios/display_server_ios.mm
```diff
@@ -454,6 +454,32 @@
 	}
 }
 
+DisplayServerIOS::IOSUIGestureState DisplayServerIOS::ios_get_gesture_state() const {
+	return ui_gesture_state;
+}
+
+void DisplayServerIOS::ios_set_gesture_state(IOSUIGestureState p_ios_gesture_set) {
+	ui_gesture_state = ios_gesture_set;
```
bruvzg, line 462:
```suggestion
	ui_gesture_state = p_ios_gesture_set;
```

platform/ios/display_server_ios.mm
```diff
@@ -454,6 +454,32 @@
...
+}
+
+bool DisplayServerIOS::ios_get_status_bar_appearance() const {
+	return status_bar_hidden;
+}
+
+void DisplayServerIOS::ios_set_status_bar_appearance(bool p_hide_status_bar, float p_status_bar_fade_time) {
+	status_bar_hidden = hide_status_bar;
+	if (status_bar_fade_time > 0) {
+		[UIView animateWithDuration:status_bar_fade_time
```
bruvzg, line 474:
```suggestion
	status_bar_hidden = p_hide_status_bar;
	if (p_status_bar_fade_time > 0) {
		[UIView animateWithDuration:p_status_bar_fade_time
```

platform/ios/display_server_ios.mm
bruvzg: Also, probably should quit early if it's not changed:
```c++
if (ui_gesture_state == p_ios_gesture_set) {
    return;
}
ui_gesture_state = p_ios_gesture_set;
```

platform/ios/display_server_ios.mm
bruvzg: Same here with early return.

servers/display_server.cpp
```diff
@@ -709,6 +727,12 @@ void DisplayServer::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("screen_get_scale", "screen"), &DisplayServer::screen_get_scale, DEFVAL(SCREEN_OF_MAIN_WINDOW));
 	ClassDB::bind_method(D_METHOD("is_touchscreen_available"), &DisplayServer::is_touchscreen_available, DEFVAL(SCREEN_OF_MAIN_WINDOW));
 	ClassDB::bind_method(D_METHOD("screen_get_max_scale"), &DisplayServer::screen_get_max_scale);
+
+	ClassDB::bind_method(D_METHOD("ios_set_gesture_state", "ios_gesture_set"), &DisplayServer::ios_set_gesture_state);
+	ClassDB::bind_method(D_METHOD("ios_get_gesture_state"), &DisplayServer::ios_get_gesture_state);
+	ClassDB::bind_method(D_METHOD("ios_set_status_bar_appearance", "hide_status_bar", "status_bar_fade_time"), &DisplayServer::ios_set_status_bar_appearance, DEFVAL(false), DEFVAL(0.2));
+	ClassDB::bind_method(D_METHOD("ios_get_status_bar_appearance"), &DisplayServer::ios_get_status_bar_appearance);
```
bruvzg, line 734: I'm not sure about functions and constants with `ios` in the name, are there equivalents for Android? Even if it's only relevant for iOS now, it's probably better to keep API OS-agnostic.

ztc0611: As far as I am aware, the only equivalent android has is Immersive Mode which is just a true/false setting that hides both and requires a double swipe all in one.

doc/classes/DisplayServer.xml
```diff
@@ -849,6 +849,37 @@
 				[b]Note:[/b] This method is implemented only on macOS.
 			</description>
 		</method>
+		<method name="ios_get_gesture_state" qualifiers="const">
+			<return type="int" enum="DisplayServer.IOSUIGestureState" />
+			<description>
+				Returns the state of the UI gestures currently in effect.
+				[b]Note:[/b] This method is only implemented on iOS.
```
AThousandShips, line 856: This note isn't necessary if it's named `ios_*`, but I agree that the naming is probably not necessarily right for all these at least

ztc0611:
Is there anything further I should do to improve this?

ztc0611:
Any chance this one could still make it into 4.2?

AThousandShips:
Since this is a feature it's unlikely to be backported to 4.2

ztc0611:
I meant to say 4.3. 🤦‍♂️ It's been a long week.

AThousandShips:
Not likely, we're in feature freeze and aiming to release soon, so will have to wait for 4.4

ztc0611:
I noticed some dev releases of 4.4 are already coming, is it possible this can get merged?

ztc0611:
@bruvzg do I need to do any other adjustments to this?

xcalibur839:
Currently working on a game that has interactive nodes near the bottom center. Without this feature implemented, the game is practically unplayable on iOS. Every other iOS game can both hide the home bar and require a swipe from the bottom to unhide it. The fact that these options are mutually exclusive in Godot is a massive problem that appears to be exclusive to Godot. Please fix/merge ASAP!! @bruvzg @AThousandShips

ztc0611:
@xcalibur839 As far as I was ever aware, hiding the gesture bar at the bottom of the screen and requiring a double swipe to exit are for whatever reason mutually exclusive on iOS. When I tested this, it appeared to be the case. I will experiment with this again. (Including in a native app in Xcode) If it is possible to combine the two, I actually need to modify this PR because it codified this limitation in the enum.

Relevant documentation:
https://developer.apple.com/documentation/uikit/uiviewcontroller/prefershomeindicatorautohidden
https://developer.apple.com/documentation/uikit/uiviewcontroller/preferredscreenedgesdeferringsystemgestures

Neither of these mention mutual exclusivity, but when trying to combine the two one would always override the other. (I no longer remember which would take priority.) All it says is "The system takes your preference into account, but returning true is no guarantee that the indicator will be hidden."

This PR's usefulness would be for contextual situations like title screens, sharable codes, etc. (Ex: You have some kind of data that can be shared from the game to other users or the Internet, while the share panel is open the app goes from requiring a double swipe to exit, to a single swipe. Another is the title screen has standard iOS behavior, but the game hides the bar or requires a double swipe.) In iOS apps a common use of this behavior is to hide the gesture bar when going into full screen to watch a video, but having it appear in the rest of the interface.

ztc0611:
I tested it in Xcode, and [prefershomeindicatorautohidden](https://developer.apple.com/documentation/uikit/uiviewcontroller/prefershomeindicatorautohidden) always overrides [preferredscreenedgesdeferringsystemgestures](https://developer.apple.com/documentation/uikit/uiviewcontroller/preferredscreenedgesdeferringsystemgestures) in my testing. I do not see any APIs that could cause alternative behavior. The gesture bar does turn translucent when being deferred, though.

I tested a handful of games and the majority of them had the defer behavior enabled. Roblox is an example of behavior this PR allows but is not currently possible, when browsing games you can swipe out no problem, but defer enables when you enter a game and requires the double swipe.

jasharpe:
I fully support this PR and actually just filed an issue on this exact thing (https://github.com/godotengine/godot/issues/104411) because I didn't see your issue in proposals.

I concur that:

1. The settings are mutually exclusive. This is confirmed by lots of chatter on the internet, as well as examining a number of existing apps where the bar is visible when the "double swipe" behavior is enabled.

2.  "Suppress UI Gesture" is more important and so should be the default.

3. Changing it at runtime may be useful.

Hope to see some more movement on this.

ztc0611:
> 2\. "Suppress UI Gesture" is more important and so should be the default.

This is true. I will amend that. The majority of developers that don't want to fiddle with settings would expect this behavior for a game, I'm not really sure what I was thinking by making the current default be the setting.

jasharpe:
Just to be totally clear, I'm just some random guy who uses Godot, not a Godot maintainer, so I can't approve this. I just appreciate this PR and want it to get in to save other people the hassle of figuring out the broken defaults. :)

ztc0611:
> Just to be totally clear, I'm just some random guy who uses Godot, not a Godot maintainer, so I can't approve this. I just appreciate this PR and want it to get in to save other people the hassle of figuring out the broken defaults. :)

I know haha, I just badly phrased my message. But maintainer or not, you're correct that making the default be requiring double swipe to exit app makes the most sense. That's the most common setting for games on iOS.

Calinou:
I agree the defaults should be tailored for games, with non-game settings being something you should set manually. (In the long term, we should provide project templates that would set the "correct" settings automatically when creating a project, and also when creating new export template presets in said projects.)

ztc0611:
Rebased and addressed merge conflict.

danijmn:
I agree that deferring gestures is probably the best default for most games. But certain games (e.g. puzzle games) can be tailored to fit entirely within the safe area, and in that case, forcing the user to swipe twice is probably pointless and defies user expectations. Some developers may be unaware of this and just go with the default, so IMO this should be mentioned in the documentation (both for the project and the runtime setting). Something like:

`If your game properly accounts for the device safe area and does not rely on swipe gestures near the screen edges, consider disabling this setting for a better user experience.`

Issue #85581, opened 2023-12-01
.NET Godot 4.2 crashes when upgrading project from 4.1.3


Dirt410:
### Godot version

4.2.stable.mono

### System information

Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3060 Ti (NVIDIA; 31.0.15.4617) - Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz (12 Threads)

### Issue description

Attempting to open a project in 4.2 that was last edited in 4.1.3 causes the editor to crash on the splash screen.

### Steps to reproduce

![image](https://github.com/godotengine/godot/assets/42548886/a6261449-c424-40d5-9eff-92d39d561075)
- Open Godot 4.2 and select project
- "This project was last edited in 4.1..." warning pops up and click OK
- A new Godot window with splash screen opens for a few seconds, then crashes.

### Minimal reproduction project

Don't know what causes it and not sure what to provide. Can provide entire C# project if needed.

akien-mga:
Are you using godot-jolt by any chance? If so you need to upgrade it or remove it to open in 4.2, the old version is not compatible and crashes.

Muini:
@akien-mga I had the same issue on macOS m1, upgrade godot-jolt fixed the issue ✨

Dirt410:
Deleting Jolt from my project worked. Thanks.

Issue #85580, opened 2023-12-01
Godot 4.2 crashes when attempting to open project from Godot 4.1.3


amarraff:
### Godot version

v4.2.stable.official [46dc27791]

### System information

Godot v4.2.stable - Kubuntu 23.10 - X11 - Vulkan (Mobile) - dedicated NVIDIA GeForce RTX 3080 (nvidia; 535.129.03) - AMD Ryzen 9 5900X 12-Core Processor (24 Threads)

### Issue description

When I open Godot 4.2 it can see my 4.1.3 project and tells me that opening it will turn it into a 4.2 project. When I hit okay, Godot opens and then crashes.

### Steps to reproduce

1. Create project in Godot 4.1.3
2. Open Godot 4.2
3. Attempt to open the project with Godot 4.2

### Minimal reproduction project

N/A

brcontainer:
@amarraff related https://github.com/godotengine/godot/issues/85416#issuecomment-1834429851?

amarraff:
> @amarraff related [#85416 (comment)](https://github.com/godotengine/godot/issues/85416#issuecomment-1834429851)?

Hi! I'm not quite sure if that's a related issue, my issue happens before I even get into the editor.

brcontainer:
Is your project FowardPlus? Try switching to Mobile mode and see if the error persists.

akien-mga:
Are you using godot-jolt by any chance? If so you need to upgrade it or remove it to open in 4.2, the old version is not compatible and crashes.

amarraff:
> Are you using godot-jolt by any chance? If so you need to upgrade it or remove it to open in 4.2, the old version is not compatible and crashes.

I am indeed using jolt, I have scenes with jolt components in them such as hinges and cone twist joints. I tried upgrading it by redownloading it in the asset library, but that doesn't seem to have fixed the problem, unless that's not how you upgrade it. I also tried changing the physics engine to the default Godot one and then opening in 4.2, but same problem.

OhiraKyou:
Jolt 0.10.0-stable seems to work in Godot 4.2. However, the asset library's installer doesn't fully reinstall add-ons; it seem to ignore existing files rather than overwriting them.

My steps to a successful launch:

1. Opened my project in Godot 4.1.3.
2. Deleted the `res://addons/godot-jolt/` directory using Godot's file explorer, selecting the "continue" option when prompted to continue or restart the editor.
3. Reinstalled Jolt from the asset library.
[...]

Issue #85576, opened 2023-11-30
4.2 Crash on startup when loading 4.1 project ("Failed to bind socket. Error: 3") 


sebastianrueckerai:
### Godot version

4.2.stable.mono.official

### System information

Godot v4.2.stable.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2080 Ti (NVIDIA; 31.0.15.4592) - AMD Ryzen 9 3950X 16-Core Processor (32 Threads)

### Issue description

I have a project that starts up fine in Godot 4.1 but crashes on startup in 4.2.
I attached the console output below (I ran it with "--verbose").

- I am able to open other project without crashes
- I tried closing open complex open scenes in the crashing project in 4.1 before reopening it in 4.2
- I am using jolt physics in my project. I switched that back to default GodotPhysics3D

I am including the complete logs further down.
There is this error, but it pops up before loading the project:
```
ERROR: GENERAL - Message Id Number: 0 | Message Id Name: Loader Message
        loader_get_json: Failed to open JSON file C:\Program Files\NVIDIA Corporation\Nsight Systems 2019.5.2\Target-Windows\x86_64\VkLayers\VkLayer_nsight-sys_windows.json
        Objects - 1
                Object[0] - VK_OBJECT_TYPE_INSTANCE, Handle 2522255807872 
```

The output right after opening a project is this:
```
WARNING: Blend file import is enabled in the project settings, but no Blender path is configured in the editor settings. Blend files will not be imported.
     at: _editor_init (modules/gltf/register_types.cpp:63)
Found .NET Sdk version '6.0.200': C:\Program Files\dotnet\sdk\6.0.200
Loaded system CA certificates
Creating VMA small objects pool for memory type index 3
Failed to bind socket. Error: 3
```

Complete output:

```
C:\Godot\Godot_v4.2-stable_mono_win64>Godot_v4.2-stable_mono_win64_console.exe
Godot Engine v4.2.stable.mono.official.46dc27791 - https://godotengine.org
OpenGL API 3.3.0 NVIDIA 545.92 - Compatibility - Using Device: NVIDIA - NVIDIA GeForce RTX 2080 Ti

Editing project: C:/Users/Sebastian/Dropbox/Godot 4 Projects/HoppBits
Godot Engine v4.2.stable.mono.official.46dc27791 - https://godotengine.org
[...]

PR #85779, opened 2023-12-05
[4.2] Fix crash when using incompatible versions of Godot Jolt

mihe:
Related to #85576, #85580 and #85581.

As per [the discussion](https://chat.godotengine.org/channel/gdextension?msg=jCpK5kNQeEgsgseWG) on Rocket Chat, This introduces a somewhat hacky check to see if the GDExtension library being loaded has its filename beginning with `godot_jolt_` and is also targeting a `compatibility_minimum` of `4.1`, which indicates that it's a version that was never officially supported to work with Godot 4.2.

Note that this PR targets the `4.2` branch, since it seemed like this would/should be a temporary hack for 4.2 specifically.

Also note that this will technically reject Godot Jolt version 0.9.0 and 0.10.0, despite those technically being (somewhat) compatible with Godot 4.2.

Let me know what you think of the error message.

CC: @akien-mga @YuriSizov @dsnopek

### DIFF
--- a/core/extension/gdextension.cpp
+++ b/core/extension/gdextension.cpp
@@ -912,6 +912,11 @@ Error GDExtensionResourceLoader::load_gdextension_resource(const String &p_path,
 		return ERR_FILE_NOT_FOUND;
 	}
 
+	if (library_path.get_file().begins_with("godot-jolt_") && compatibility_minimum[0] == 4 && compatibility_minimum[1] == 1) {
+		ERR_PRINT("Godot Jolt failed to load, as the currently installed version is incompatible with Godot 4.2. You can update it to a compatible version by deleting it and installing it again.");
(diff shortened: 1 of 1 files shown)


YuriSizov (approved):
Message looks good to me and I approve this conceptually. I'll let @dsnopek to double-check the GDExtension side of it.

akien-mga (approved):
Looks good to me. Thanks for the quick patch :)

mihe:
Seems to work fine on all three major platforms, as far as I can tell.

dsnopek (approved):
Thanks! This looks good to me :-)

If we can get the "safe mode" and crash handler stuff we've talked about implemented for 4.3, then this hack shouldn't be needed there.

YuriSizov:
Thanks!

PR #84563, opened 2023-11-07
[macOS] Add support for ANGLE over native OpenGL (as a fallback for ANGLE over Metal).

bruvzg:
- Adds support for ANGLE over native GL for macOS.
- ANGLE over Metal is used on macOS 10.15+ with Metal 2 level of hardware, older Macs default to ANGLE over native GL.
- ANGLE over GL is also used if ANGLE over Metal fails.

Depends on https://github.com/godotengine/godot-angle-static/pull/2

PR #85785, opened 2023-12-05
[macOS] Switch ANGLE backend to ANGLE over OpenGL, switch default compatibility renderer back to native.

### LINKED ISSUE #84591: TileMap performance degradation on MacOS using ANGLE
### Godot version

### System information

Godot v4.2.beta5 - macOS 14.0.0 - GLES3 (Compatibility) - ANGLE (Apple, ANGLE Metal Renderer: Apple M1, Version 14.0 (Build 23A344)) - Apple M1 (8 Threads)

### Issue description

MacOS seems to have very bad TileMap performance on very small (8 tiles) tile sets. I opened same project on my lower end Windows 10 machine and it ran perfectly. Once running game seems to run fine, it's only in the editor this issue seems to occur and I've noticed it in all the Betas I worked on this project in which was Beta 2. If I use the Stable 4.1.3 that was just released the performance issue doesn't exist. This is only since Version 4.2. I know some changes came with the system and may have greatly affected performance on the MacOS systems, but not Windows. Linux is unknown. 

### Steps to reproduce

Create a project with a TileMap and then attach any size tile set. Select the TileMap tab at the bottom and watch the editor feel like it's gonna come to a halt. 

N/A

### LINKED ISSUE #85619: MacBook Pro quickly maxes out its memory when running a project on Compatibility rendering method with ANGLE
### Godot version

4.2.stable (.NET version)

### System information

macOS Ventura 13.6.1, MacBook 2017 Intel Core i5, Iris Plus 650, 16 GB ram

### Issue description

My friend was testing my Godot project (which has the Compatibility rendering method configured) and he said that his MacBook quickly ran out of memory. The game was allocating 15 GB of RAM after a minute or two. More interestingly, the same happened with the 2D platformer demo project from https://github.com/godotengine/godot-demo-projects, when running in Compatibility mode. My Windows machine doesn't suffer from this at all.

Fortunately, we figured out that switching to Forward+ mode fixed the issue. The downside is that I have now lost a potential audience of devices that need Compatibility rendering.

Here's what he saw in the profiler:

WhatsApp Image 2023-12-01 at 17 00 58_40cb5e91

By the way, when my friend closes the editor, he gets the following which might be related (I have no idea but who knows):

image

### Steps to reproduce

Run the 2D platformer demo project from https://github.com/godotengine/godot-demo-projects and set it to the Compatibility rendering method (or possibly any other game that can run in this mode) on a MacBook (maybe only one that matches the provided system specs or maybe any Mac, I don't know).

2D/platformer demo from https://github.com/godotengine/godot-demo-projects in Compatibility rendering mode

bruvzg:
I think it's pretty clear that using ANGLE over Metal on macOS was a total failure. Unlike ANGLE over DX11 on Windows which seems to be doing a good job for supporting older hardware, it's causing a lot more issues than fixing.

Superseded https://github.com/godotengine/godot/pull/84563
Fixes https://github.com/godotengine/godot/issues/85619
Fixes https://github.com/godotengine/godot/issues/84591

### DIFF
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -1758,7 +1758,6 @@ Error Main::setup(const char *execpath, int argc, char *argv[], bool p_second_ph
 #endif
 
 		String default_driver = driver_hints.get_slice(",", 0);
-		String default_driver_macos = driver_hints_angle.get_slice(",", 1);
 
 		GLOBAL_DEF_RST_NOVAL("rendering/gl_compatibility/driver", default_driver);
 		GLOBAL_DEF_RST_NOVAL(PropertyInfo(Variant::STRING, "rendering/gl_compatibility/driver.windows", PROPERTY_HINT_ENUM, driver_hints_angle), default_driver);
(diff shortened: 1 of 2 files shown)


akien-mga:
Is there much value left in using ANGLE over OpenGL as opposed to just native OpenGL? The main rationale for using ANGLE on macOS was the fear that bitrotting OpenGL drivers would have too many issues, but if we're still using them via ANGLE there might not be much benefit in having ANGLE at all?

bruvzg:
> Is there much value left in using ANGLE over OpenGL as opposed to just native OpenGL? 

The only issues with ANGLE over OpenGL is slow shader compilation (first start will take minute or two, some WebGL issue we have on macOS), and it's not fixing GPU particles issues. It might be fixing some issues, but I'm not sure. So we probably should not have it built-in, but keeping support for dynamically linked versions cost noting (most of the code is shared with Windows, and EGL on Linux), so I do not see many reasons to completely remove macOS specific part of it.

> The main rationale for using ANGLE on macOS was the fear that bitrotting OpenGL drivers would have too many issues, but if we're still using them via ANGLE there might not be much benefit in having ANGLE at all?

Unlikely it will get less functional, and when OpenGL is removed from macOS, I would expect Google to fix all issues in the ANGLE over Metal (unless WebGPU will supersede it by the time) for the use in Chromium.

akien-mga (approved):

PR #85789, opened 2023-12-05
Make `FileDialog` filtering case insensitive

### LINKED ISSUE #85788: Allow FileDialog filter checks to not be case sensitive
### Tested versions

- Reproducible: 4.3.dev (654132cb9), 4.2.stable, 4.1.stable

### System information

Godot v4.3.dev (654132cb9) - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3080 Ti (NVIDIA; 31.0.15.3667) - AMD Ryzen 9 7900X 12-Core Processor (24 Threads)

### Issue description

When using FileDialog to save files on non-case-sensitive file systems (Windows and often macOS), a file may have an uppercase file extension ( `file.TXT` ) but the filter is set as lowercase ( `*.txt ; Text Files` ). The FileDialog will list the file but selecting it to overwrite will not detect that the file matches the filter and append the lowercase extension after the existing uppercase file extension and would result in the selected file being `file.TXT.txt`

As most other operating systems can be case sensitive, my suggestion is to add a `CaseSensitivity` enum with the values of `True`, `False` and `OS Default` to the FileDialog for toggling between using `f.match(str)` and `f.matchn(str)` in the checks which would allow it to either be explicit based on the developer intent or implicit based on the standard of the operating system.

filedialog-case-sensitivity

### Steps to reproduce

- Add a text file called `Test.TXT` to the project, make sure the extension is uppercase.
- Add a FileDialog node to a scene with `File Mode` set to `Save` and a filter set to `*.txt ; Text Files`
- Add a Button with its `pressed` signal hooked up to the FileDialog's `show()` method
- Add a Label and hook the FileDialog's `file_selected` signal up to Label's the `set_text` method
- Run the project
- Click the button to open the FileDialog
- Double click the `Text.TXT` file to select it
- The label will display `res://Test.TXT.txt`

[FileDialogCaseSensitivityRepro.zip](https://github.com/godotengine/godot/files/13561701/FileDialogCaseSensitivityRepro.zip)

DevilboxGames:
When using FileDialog to save files on non-case-sensitive file systems (Windows and often macOS), a file may have an uppercase file extension ( `file.TXT` ) but the filter is set as lowercase ( `*.txt ; Text Files` ). The FileDialog will list the file but selecting it to overwrite will not detect that the file matches the filter and append the lowercase extension after the existing uppercase file extension and would result in the selected file being `file.TXT.txt`

As most other operating systems can be case sensitive, my suggestion is to add a `CaseSensitivity` enum with the values of `True`, `False` and `OS Default` to the FileDialog for toggling between using `f.match(str)` and `f.matchn(str)` in the checks which would allow it to either be explicit based on the developer intent or implicit based on the standard of the operating system.

filedialog-case-sensitivity

Currently if `OS Default` is used then the code uses a case-sensitive check if the OS is not Windows or macOS. Is there a better approach for this? Are there any other supported OS's are not case-sensitive by default?

Attached is an MRP similar to the one provided in the issue, with the addition of an OptionButton for changing between the CaseSensitivity setting when running the project.
[FileDialogCaseSensitivity.zip](https://github.com/godotengine/godot/files/13561697/FileDialogCaseSensitivity.zip)

Fixes #85788

### DIFF
--- a/scene/gui/file_dialog.cpp
+++ b/scene/gui/file_dialog.cpp
@@ -138,7 +138,7 @@ void FileDialog::_native_dialog_cb(bool p_ok, const Vector<String> &p_files, int
 				int filter_slice_count = flt.get_slice_count(",");
 				for (int j = 0; j < filter_slice_count; j++) {
 					String str = (flt.get_slice(",", j).strip_edges());
-					if (f.match(str)) {
+					if (f.matchn(str)) {
 						valid = true;
 						break;
 					}
@@ -473,7 +473,7 @@ void FileDialog::_action_pressed() {
 				String flt = filters[i].get_slice(";", 0);
 				for (int j = 0; j < flt.get_slice_count(","); j++) {
 					String str = flt.get_slice(",", j).strip_edges();
-					if (f.match(str)) {
+					if (f.matchn(str)) {
 						valid = true;
 						break;
 					}
@@ -492,7 +492,7 @@ void FileDialog::_action_pressed() {
 				int filterSliceCount = flt.get_slice_count(",");
 				for (int j = 0; j < filterSliceCount; j++) {
 					String str = (flt.get_slice(",", j).strip_edges());
-					if (f.match(str)) {
+					if (f.matchn(str)) {
 						valid = true;
 						break;
 					}
--- a/editor/gui/editor_file_dialog.cpp
+++ b/editor/gui/editor_file_dialog.cpp
@@ -127,7 +127,7 @@ void EditorFileDialog::_native_dialog_cb(bool p_ok, const Vector<String> &p_file
 				int filter_slice_count = flt.get_slice_count(",");
 				for (int j = 0; j < filter_slice_count; j++) {
 					String str = (flt.get_slice(",", j).strip_edges());
-					if (f.match(str)) {
+					if (f.matchn(str)) {
 						valid = true;
 						break;
 					}
@@ -565,7 +565,7 @@ void EditorFileDialog::_action_pressed() {
 				String flt = filters[i].get_slice(";", 0);
 				for (int j = 0; j < flt.get_slice_count(","); j++) {
 					String str = flt.get_slice(",", j).strip_edges();
-					if (f.match(str)) {
+					if (f.matchn(str)) {
 						valid = true;
 						break;
 					}
@@ -584,7 +584,7 @@ void EditorFileDialog::_action_pressed() {
 				int filterSliceCount = flt.get_slice_count(",");
 				for (int j = 0; j < filterSliceCount; j++) {
 					String str = (flt.get_slice(",", j).strip_edges());
-					if (f.match(str)) {
+					if (f.matchn(str)) {
 						valid = true;
 						break;
 					}


scene/gui/file_dialog.cpp
```diff
@@ -355,6 +355,8 @@ void FileDialog::_action_pressed() {
 
 	if (mode == FILE_MODE_SAVE_FILE) {
 		bool valid = false;
+		String osName = OS::get_singleton()->get_name();
+		bool use_case_sensitive_filter = case_sensitive_filter == CASE_SENSITIVITY_TRUE || case_sensitive_filter == CASE_SENSITIVITY_OS_DEFAULT && osName != "Windows" && osName != "macOS";
```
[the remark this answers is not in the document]
DevilboxGames: Aha, thanks for the heads up. I expected there was a more appropriate approach but completely missed that. I've amended that now.

scene/gui/file_dialog.cpp
```diff
@@ -356,6 +356,8 @@ void FileDialog::_action_pressed() {
 	if (mode == FILE_MODE_SAVE_FILE) {
 		bool valid = false;
 
+		bool use_case_sensitive_filter = case_sensitive_filter == CASE_SENSITIVITY_TRUE || case_sensitive_filter == CASE_SENSITIVITY_OS_DEFAULT && dir_access->is_case_sensitive(f);
```
AThousandShips, line 359:
```suggestion
		bool use_case_sensitive_filter = case_sensitive_filter == CASE_SENSITIVITY_TRUE || (case_sensitive_filter == CASE_SENSITIVITY_OS_DEFAULT && dir_access->is_case_sensitive(f));
```
Or
```suggestion
		bool use_case_sensitive_filter = (case_sensitive_filter == CASE_SENSITIVITY_TRUE || case_sensitive_filter == CASE_SENSITIVITY_OS_DEFAULT) && dir_access->is_case_sensitive(f);
```
Depending on what you indented, it's ambiguous

scene/gui/file_dialog.cpp
```diff
@@ -364,7 +366,7 @@ void FileDialog::_action_pressed() {
 				String flt = filters[i].get_slice(";", 0);
 				for (int j = 0; j < flt.get_slice_count(","); j++) {
 					String str = flt.get_slice(",", j).strip_edges();
-					if (f.match(str)) {
+					if (case_sensitive_filter && f.match(str) || !case_sensitive_filter && f.matchn(str)) {
```
AThousandShips, line 369:
```suggestion
					if ((case_sensitive_filter && f.match(str)) || (!case_sensitive_filter && f.matchn(str))) {
```

scene/gui/file_dialog.cpp
```diff
@@ -383,7 +385,7 @@ void FileDialog::_action_pressed() {
 				int filterSliceCount = flt.get_slice_count(",");
 				for (int j = 0; j < filterSliceCount; j++) {
 					String str = (flt.get_slice(",", j).strip_edges());
-					if (f.match(str)) {
+					if (use_case_sensitive_filter && f.match(str) || !use_case_sensitive_filter && f.matchn(str)) {
```
AThousandShips, line 388:
```suggestion
					if ((use_case_sensitive_filter && f.match(str)) || (!use_case_sensitive_filter && f.matchn(str))) {
```

AThousandShips:
You also need to register these enums in `_bind_methods`, see the other ones

AThousandShips:
You also need to document these methods, use `--doctool`, see [here](https://docs.godotengine.org/en/stable/contributing/documentation/updating_the_class_reference.html#updating-class-reference-when-working-on-the-engine)

DevilboxGames:
> You also need to document these methods, use `--doctool`, see [here](https://docs.godotengine.org/en/stable/contributing/documentation/updating_the_class_reference.html#updating-class-reference-when-working-on-the-engine)

I've updated the doc xml now, is this correct? The `--doctool` tool didn't generate any additional entries in it for the enum or methods.

 Documentation isn't my forte so the wording might need tweaking.

doc/classes/FileDialog.xml
```diff
@@ -133,6 +136,15 @@
 		<constant name="ACCESS_FILESYSTEM" value="2" enum="Access">
 			The dialog allows accessing files on the whole file system.
 		</constant>
+		<constant name="CASE_SENSITIVITY_TRUE" value="0" enum="CaseSensitivity">
+			The dialog will use a case sensitive check against the filters when saving.
+		</constant>
+		<constant name="CASE_SENSITIVITY_FALSE" value="1" enum="CaseSensitivity">
+			The dialog will use not a case sensitive check against the filters when saving.
+		</constant>
```
YuriSizov, line 144: Generally speaking these are pretty weird names for constants. It's better to be explicit, e.g. `CASE_SENSITIVITY_SENSITIVE`/`CASE_SENSITIVITY_INSENSITIVE`.

scene/gui/file_dialog.cpp
```diff
@@ -1044,6 +1048,7 @@ void FileDialog::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::STRING, "root_subfolder"), "set_root_subfolder", "get_root_subfolder");
 	ADD_PROPERTY(PropertyInfo(Variant::PACKED_STRING_ARRAY, "filters"), "set_filters", "get_filters");
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "show_hidden_files"), "set_show_hidden_files", "is_showing_hidden_files");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "case_sensitive_filter", PROPERTY_HINT_ENUM, "True,False,OS Default"), "set_case_sensitive_filter", "get_case_sensitive_filter");
```
YuriSizov, line 1051: Same for these strings.

doc/classes/FileDialog.xml
```diff
@@ -133,6 +136,15 @@
 		<constant name="ACCESS_FILESYSTEM" value="2" enum="Access">
 			The dialog allows accessing files on the whole file system.
 		</constant>
+		<constant name="CASE_SENSITIVITY_SENSITIVE" value="0" enum="CaseSensitivity">
+			The dialog will use a case sensitive check against the filters when saving.
```
AThousandShips, line 140:
```suggestion
			The dialog will use a case sensitive check against the filters when saving.
```
Stray whitespace

scene/gui/file_dialog.cpp
```diff
@@ -1043,6 +1047,7 @@ void FileDialog::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "access", PROPERTY_HINT_ENUM, "Resources,User Data,File System"), "set_access", "get_access");
 	ADD_PROPERTY(PropertyInfo(Variant::STRING, "root_subfolder"), "set_root_subfolder", "get_root_subfolder");
 	ADD_PROPERTY(PropertyInfo(Variant::PACKED_STRING_ARRAY, "filters"), "set_filters", "get_filters");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "case_sensitive_filter", PROPERTY_HINT_ENUM, "Sensitive,Insensitive,OS Default"), "set_case_sensitive_filter", "get_case_sensitive_filter");
```
AThousandShips, line 1050:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::INT, "case_sensitive_filter", PROPERTY_HINT_ENUM, "Case Sensitive,Case Insensitive,OS Default"), "set_case_sensitive_filter", "get_case_sensitive_filter");
```
I'd say this is clearer

AThousandShips: Or, by some standards:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::INT, "case_sensitive_filter", PROPERTY_HINT_ENUM, "Case-Sensitive,Case-Insensitive,OS Default"), "set_case_sensitive_filter", "get_case_sensitive_filter");
```

doc/classes/FileDialog.xml
```diff
@@ -57,6 +57,9 @@
 			The file system access scope. See [enum Access] constants.
 			[b]Warning:[/b] Currently, in sandboxed environments such as Web builds or sandboxed macOS apps, FileDialog cannot access the host file system. See [url=https://github.com/godotengine/godot-proposals/issues/1123]godot-proposals#1123[/url].
 		</member>
+		<member name="case_sensitive_filter" type="int" setter="set_case_sensitive_filter" getter="get_case_sensitive_filter" enum="FileDialog.CaseSensitivity" default="2">
+			Whether the file type filters are case sensitive. See [enum CaseSensitivity].
```
AThousandShips, line 61:
```suggestion
			Whether the file type filters are case-sensitive. See [enum CaseSensitivity].
```
Seems to be the standard spelling, same below

AThousandShips: Though we don't use it for `DirAccess` so let's keep it in the same style

AThousandShips: We can look into standardising it later 🙂

DevilboxGames:
I noticed a merge conflict, so I've rebased and resolved it.

DevilboxGames:
It looks like one of the tests failed due to a ubuntu keyserver issue. Could a maintainer request a rerun of the failed test?

I appologise if the ping is innapropriate, but Yuri also requested requested @bruvzg to re-review the PR in December. Is there anything else which needs to be done before it can be approved for merging?

Calinou:
Tested locally on Linux (on a case-sensitive filesystem), I'm a bit confused about the behavior. The only behavior that seems desirable to me here is the one where case sensitive is set to **False**:

https://github.com/godotengine/godot/assets/180032/78d912ea-1b03-4c26-bcb4-f364bfacbac9

Otherwise, you get a double extension appended.

Also, the file filter is always case-insensitive regardless of this option, but I guess that's expected.

KoBeWi:
IMO this makes sense. The change only affects extension checking and I don't see why would you care about extension casing and make a double extension, like currently happens on mismatch.

The same change should be done in EditorFileDialog.

DevilboxGames:
> IMO this makes sense. The change only affects extension checking and I don't see why would you care about extension casing and make a double extension, like currently happens on mismatch.
> 
> The same change should be done in EditorFileDialog.

Good shout, I've added the change to EditorFileDialog too, I also noticed `FileDialog::_native_dialog_cb` needed the change too so I've dropped it in there.

KoBeWi (approved):

KoBeWi:
Needs to be squashed into 1 commit.

DevilboxGames:
Commits squashed 👍

AThousandShips (approved):

akien-mga:

PR #84660, opened 2023-11-09
Move tile transforms handling cache to TileData

groud:
This PR does multiple things:
- moves the transformed polygons data cache to the TileData object. Since it's part of the TileSet resource, that should avoid further errors like https://github.com/godotengine/godot/issues/83825. It reverts https://github.com/godotengine/godot/pull/84261 and fix the issue in a cleaner way.
- Fix some errors that would be printed when using occluders. We were storing invalid occlusion polygons in the TileSet which caused errors to be printed.
- Fixed baking NavigationRegions with rotated tiles (an island is collision shapes, the other in navigation polygons, both work):  
Screenshot_20231109_123615

*Bugsquad edit:* Fixes #83825

PR #85791, opened 2023-12-05
Move TileMapLayer to its own files

groud:
As  I mentioned in [this proposal](https://github.com/godotengine/godot-proposals/issues/6415), I'd like to move TileMap layers to their own node. This fits the 

I prefer to go there step by step. This first PR moves TileMap layer to another file. 

The next steps would be:
- Change TileMap layers to be used as an internal nodes for the TileMap node (and make sure everything works).
- Implement a TileMapGroup node, that would constraint its child TileMapLayer nodes to use a given TileSet, and to keep their transform. It should have a very simple API. Keep TileMapGroup node private for now.
- Make TileMap expand the TileMapGroup node API. 
- To ensure compatibility, implement a tool to transform the TileMap node into the (TileMapGroup + Children TileMap layers).
- At the same time: 
	- expose the TileMapLayer node.
	- expose the TileMapGroup node.
	- Change the editor to edit TileMapLayer directly (and, for simplicity of the implementation, probably disables it for the TileMap node. That kind of forces people to move the to new workflow by using the auto-conversion, but it's probably acceptable IMO)

Basically, the TileMap node could become the equivalent of (TileMapGroup + Children TileMap layers), with layers being internal nodes instead of being explicit ones.


scene/2d/tile_map.h
```diff
@@ -31,409 +31,14 @@
 #ifndef TILE_MAP_H
 #define TILE_MAP_H
 
+#include "scene/2d/tile_map_layer.h"
+class TileMapLayer;
+class TerrainConstraint;
```
AThousandShips, line 36: Aren't these classes included just before?

AThousandShips:
I'm a bit concerned by the circular inclusion, can this not be handled in any other way?

groud:
> I'm a bit concerned by the circular inclusion, can this not be handled in any other way?

I thought there was not, but I managed to solve it.

scene/2d/tile_map_layer.h
```diff
@@ -0,0 +1,431 @@
...
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef TILE_MAP_LAYER_H
+#define TILE_MAP_LAYER_H
+
+#include "scene/2d/node_2d.h"
+#include "scene/2d/tile_map.h"
+#include "scene/gui/control.h"
```
KoBeWi, line 36: What is `node_2d` and `control` for?

scene/2d/tile_map.h
```diff
@@ -35,403 +35,14 @@
 #include "scene/gui/control.h"
```
KoBeWi, line 35: Unrelated, but this can be forward-declared.

akien-mga: Forward declaring in headers helps reduce the amount of files that need to be rebuilt on incremental builds. If a commit changes `control.h`, `tile_map.h` and all the files that include it would need to be recompiled.

If it's forward declared, only `tile_map.cpp` needs to be recompiled.

groud: Oh ok, that makes sense! I'll update it then, thanks.

KoBeWi (approved):

groud:
> Needs a rebase and we can merge it soon if it unblocks your further work.

I'll rebase once https://github.com/godotengine/godot/pull/84660 is merged, to avoid conflicts.

YuriSizov:
@groud You should be able to rebase now!

akien-mga:
New Year's rebase poke ;)

groud:
> New Year's rebase poke ;)

Done!

akien-mga:
Thanks!

PR #85799, opened 2023-12-05
Allow dragging selection when selecting whole words in `RichTextLabel`

HolonProduction:
With this PR the following is possible:

[Screencast](https://github.com/godotengine/godot/assets/99511741/f353f9d0-7c2e-4ef5-a6fe-a8932f38f3e5)

Before it was not possible to drag the selection after double clicking.


bruvzg (approved):
Seems to be working fine, since the same is supported by `LineEdit` and `TextEdit` it should be in RTL as well.

akien-mga:
Thanks!

PR #85801, opened 2023-12-05
Allow dragging selection when selecting whole words in `LineEdit`

HolonProduction:
Allows to drag out the selection after a double click. This is already possible in `TextEdit`.

See also #85799 which does the same for `RichTextLabel`.


KoBeWi (approved):
Can't say if the code is 100% correct, but it works correctly.

akien-mga:
Thanks!

PR #85807, opened 2023-12-05
Add toaster notifications for import warnings.

### LINKED ISSUE #23117: Using a JPEG-Image as both the albedo texture and the normal texture turns it yellowish
**Godot version:** 5804efc6371fbfaec8fca6f237bd8478f2081086

**Issue description:**

When reusing an image for both albedo and normals (let alone if that makes sense or not), when using the image as normal map, it changes its texture import settings under _compression_ to "normal map" mode, loading the texture as RG only.

**Steps to reproduce:**
1. New project, add a MeshInstance with a CubeMesh and a SpatialMaterial
2. drag-and-drop a JPEG file from the "files" tab into the "albedo texture" slot of the SpatialMaterial
3. The mesh will be textured correctly
4. drag-and-drop the very same file onto the "normal texture" slot
5. The mesh will turn green/yellow-ish. The Texture's import settings will change as described above

This happens invisibly and is highly unintuitive. Accidentially doing this renders your texture yellow-ish forever (also when clearing the normal texture again), until you find the texture import settings tab and find the right setting to revert.
New users might well stumble upon this, as the fix is highly unobvious.

There probably should be a warning when an already-imported texture's settings are changed in such a way.

rakkarage:
Enhances visibility of texture import warnings by showing `EditorToaster` notifications, in addition to the existing `print_line` if `TOOLS_ENABLED`.

Closes: https://github.com/godotengine/godot/issues/23117

Thanks.

### DIFF
--- a/editor/import/resource_importer_texture.cpp
+++ b/editor/import/resource_importer_texture.cpp
@@ -38,6 +38,7 @@
 #include "editor/editor_node.h"
 #include "editor/editor_scale.h"
 #include "editor/editor_settings.h"
+#include "editor/gui/editor_toaster.h"
 #include "editor/import/resource_importer_texture_settings.h"
 #include "scene/resources/compressed_texture.h"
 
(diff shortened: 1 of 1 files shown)


Calinou (approved):

image

Perhaps the normal map detection messages could be compacted together when possible, but this can be left to a future PR.

akien-mga (approved):

akien-mga:
Thanks!

Issue #74060, opened 2023-02-27
the total draw call monitor always show 0 in 2d scene


roycarser:
### Godot version

Godot Engine v4.0.rc6.official.0cd148313

### System information

Windows10，Vulkan，NVIDIA GeForce RTX 2060

### Issue description

![Untitled](https://user-images.githubusercontent.com/24681803/221614670-0fbf1ade-cadf-4ae1-8cc5-f978b5933324.png)

the total draw call monitor always show 0 in 2d scene, is that only work in 3d scene ?

### Steps to reproduce

create an 2D scene with some 2d resources，then run it

### Minimal reproduction project

N/A

Calinou:
The 2D performance monitors aren't reimplemented yet.

PR #85811, opened 2023-12-05
Implement render info counters for the 2D renderer

### LINKED ISSUE #8373: Performance monitors dont report any values for 2D projects
___

***Bugsquad note:** This issue has been confirmed several times already. No need to confirm it further.*

___

**Operating system or device - Godot version:** Linux, 3.0master

**Issue description:**
Performance monitors show all 0 all the time, except for FPS which seems to be working.
Both in profiler and via code.

clayjohn:
This provides information for the debug monitors for 2D objects.

For now all the draw data is compiled together for the performance monitors to maintain consistency with previous versions, but it can be reported separately if the information is pulled from ``viewport_get_render_info()``.

Fixes: https://github.com/godotengine/godot/issues/8373 (6 year old bug report :eyes:)

_RD renderer (note draw calls and objects drawn are the same):_
Screenshot from 2023-12-05 16-46-45

_Compatibility renderer:_
Screenshot from 2023-12-05 16-40-40


Zireael07:
Finally!

lawnjelly (approved):
Looks good. :+1: 

In long run would be nice to report separately from 3D as in the amazing work in 3.x.

Calinou (approved):
Tested locally with all rendering methods, it works as expected.

image

It's interesting to note two things about 2D rendering (these apply to all rendering methods):

- A Label with a single character thas has no font outline or shadow results in 6 primitives drawn. I'd expect 2 primitives to be drawn per character instead. An empty Label doesn't result in any primitives or draw calls as expected.
- Invisible characters such as spaces add the same number of draw calls/primitives as visible characters. I'd expect these to not have to be drawn, though bitmap fonts could *in theory* make these visible (I'm not aware of any such fonts).

clayjohn:
> * A Label with a single character thas has no font outline or shadow results in 6 primitives drawn. I'd expect 2 primitives to be drawn per character instead. An empty Label doesn't result in any primitives or draw calls as expected.

That was an oversight on my part. I forgot to add the conversion from Vertex/Index to "primitive"

> * Invisible characters such as spaces add the same number of draw calls/primitives as visible characters. I'd expect these to not have to be drawn, though bitmap fonts could _in theory_ make these visible (I'm not aware of any such fonts).

That seems odd, but it must be coming from the font system as the renderer doesn't detect if the section of a texture being used is fully transparent

akien-mga:
> > That seems odd, but it must be coming from the font system as the renderer doesn't detect if the section of a texture being used is fully transparent
> 
> @bruvzg Any ideas about this?

I suggest opening a bug report about this, as I'll merge this PR now :)

akien-mga:
Thanks!

Ostreicher:
not working for me. i get zero objects drawn like here https://github.com/godotengine/godot/issues/74060#issue-1601480720
	print(get_viewport().get_render_info(Viewport.RENDER_INFO_TYPE_VISIBLE,Viewport.RENDER_INFO_DRAW_CALLS_IN_FRAME)) 
0

AThousandShips:
What version are you on? If you're on 4.2 or 4.1 this isn't available, this is 4.3 only

PR #85819, opened 2023-12-06
Add option to install android build template for export

Malcolmnixon:
This PR adds a new "--export-android-build-template" command-line option which causes the android build template to be installed before exporting the project.

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -974,6 +974,9 @@ void EditorNode::_fs_changed() {
 				} else { // Normal project export.
 					String config_error;
 					bool missing_templates;
+					if (export_defer.android_build_template) {
+						export_template_manager->install_android_template();
+					}
 					if (!platform->can_export(export_preset, config_error, missing_templates, export_defer.debug)) {
 						ERR_PRINT(vformat("Cannot export project with preset \"%s\" due to configuration errors:\n%s", preset_name, config_error));
 						err = missing_templates ? ERR_FILE_NOT_FOUND : ERR_UNCONFIGURED;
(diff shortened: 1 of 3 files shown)


BastiaanOlij:
That's pretty cool, definitely helps when people download projects or templates that require this to be installed.

dsnopek (approved):
This is a great idea!

I tested this by removing the `android/build` directory in a Quest project, and then running `godot --headless --export-debug "Quest" --export-android-build-template` and it correctly recreated the `android/build` directory before rebuilding the APK file. (I also tried the same test but omitting the `--export-android-build-template` and it fails with the expected message about the build templates not being installed in the project.)

I also skimmed the code, and it looks good to me!

akien-mga:
The feature sounds good. I'm not fully sold on the name, as it starts with `--export` but it actually performs a setup/install. Maybe `--setup-android-build-template` would be better? The description of the option should also clarify that it has to be used together with one of the actual `--export-*` options, otherwise it's a no-op from what I can tell (didn't test, but from reading the code).

Architecturally though, I'm not fully satisfied with the implementation, as it breaks platform encapsulation, adding Android specific code in `main.cpp` and `editor_node.cpp`, notably in the generic `EditorNode::export_preset`. Other platforms in the future may also need pre-export steps to be performed and it would be better to have a system for this so that any export preset can specify what steps should be ensured before an export, either automatically or optionally with a command line argument.

That being said, the issue pre-exists this PR, since `install_android_template()` is a method of the generic `ExportTemplateManager`, and we expose a dedicated option in the editor to install the Android template. So that's not something that needs to be fixed in this PR, but I do think we need to rethink this stuff in the future.

akien-mga (approved):
Looks good to me.

I still have architectural concerns with the way we do this as pointed above, but this predates this PR.

PR #84976, opened 2023-11-16
Acyclic Command Graph for Rendering Device

DarioSamo:
### Background

@reduz proposed the idea a while ago of refactoring RenderingDevice to automatically build a graph out of the commands submitted to the class (outlined [here](https://gist.github.com/reduz/980b9b2547d57e6a915b2bb7e1e76e08)). The first stepping stone towards implementing this was @RandomShaper's PR (#83452) that splits the commands into easily serializable parameters. Therefore, merging this PR will require merging that PR first (and if you wish to review this, only look at my individual commit after Pedro's changes).

### Improvements

This PR makes the following improvements towards implementing @reduz's idea.

- **RenderingDevice**'s complexity has been drastically reduced as it no longer needs to solve pipeline barriers or layout transitions for pretty much all its functionality. This responsibility has been delegated to a new class called **RenderingDeviceGraph**.
- Overall the total amount of `vkCmdPipelineBarrier` calls has been reduced immensely. On average I've noticed a reduction of about 60-80% of the total amount of barrier calls in a frame when compared to `master`.
- **RenderingDeviceGraph** is capable of reordering commands based on the dependency between the resources used to push the submitted commands to be processed as early as possible. This gives the driver much better chances of parallelizing the work effectively.
- **RenderingDeviceGraph** will group as many possible barriers as possible in **'levels'** depending on the usage of the resources. These barriers are submitted before the commands of the level are processed to perform any layout or synchronization barriers that are required.
- **RenderingDevice**'s API has been simplified as some parameters are no longer required.
  - Barrier bitmasks are gone. They no longer serve any purpose.
  - Draw list and compute list overlapping no longer needs to be specified.
  - 'Split draw lists' are gone as they can be automatically recorded by the graph instead, and has been shown to be viable already (although disabled behind an experimental macro for now).
  - Draw lists no longer need to specify their initial and final actions in excruciating detail. The operations are much simpler now. Load, clear or discard for initial. Store and discard for final. The detail behind the original action no longer serves any purpose, as the graph will automatically skip any transitions that are not required if commands that use the same image layout are chained together (e.g. render passes).
  - Draw lists no longer need to specify storage textures as it's not required at all.
- Both Forward+ and Mobile have been adapted to use the new API and have overall resulted in a net removal of code complexity that is no longer required due to the graph automatically solving what it was doing already.
- A lot of existing Vulkan synchronization errors caused by the current barrier system have been solved automatically by the graph.

### Implementation
[...]

PR #85830, opened 2023-12-06
Add the option to make textures immutable.

reduz:
By making them immutable, the upcoming acyclic render graph (#84976) can simply not have them in consideration and greatly improve performance.

**NOTE:**: This PR is slightly compat breaking. If you have imported textures from disk and later attempt to modify them via texture_update, this will throw an error and you will need to reimport textures by switching the immutable flag off (the error will instruct you to do this). I am pretty sure nobody is doing this, but just in case this is good to mention.

### DIFF
--- a/doc/classes/RenderingServer.xml
+++ b/doc/classes/RenderingServer.xml
@@ -3225,9 +3225,11 @@
 		<method name="texture_2d_create">
 			<return type="RID" />
 			<param index="0" name="image" type="Image" />
+			<param index="1" name="immutable" type="bool" default="false" />
 			<description>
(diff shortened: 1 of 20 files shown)


akien-mga:
Looks like your clang-format hooks no longer work :)

reduz:
@akien-mga ah, guess its ignoring the compat.inc files

DarioSamo:
Worth noting merging this PR will break Godot without the required changes to RD to play nicely with the missing bit, so I'm mostly using this as a reference to apply and cherry-pick for now.

clayjohn:
Superseded by https://github.com/godotengine/godot/pull/84976. We ended up figuring out a way to make textures immutable automatically so we can get the performance benefit without users having to manually flag a texture as immutable

PR #85842, opened 2023-12-06
Improve normal map VRAM Compression with RGTC

### LINKED ISSUE #57981: Vulkan: Normal map mipmaps suffer from artifacts when VRAM compression is used
### Godot version

v4.0.alpha2.official [79077e6c1]

### System information

Windows 10

### Issue description

Material seems to very quickly and aggressively switch to lover res version of roughness texture, as a result you get pixelate roughness   

https://user-images.githubusercontent.com/9384421/153638408-afc4a1f0-4545-4283-ba84-2f5438458de2.mp4

In the video, I move the camera from super close up to 30 to 50 meters away from mesh plane.

### Steps to reproduce

Create ORMMaterail3D or StandardMaterial3D and put texture with packed AO, roughness and metalness in appropriate slots. Assign that material to mesh instance and up close it looks ok, but as soon as you move slightly away from mesh instance, the lover res version kicks in and pixelated effect starts to appear. 

_No response_

BlueCube3310:
Fixes #57981

This PR extends `etcpak` with an RGTC compression module, as well as changes the compression type used by normal maps to RGTC_RG, instead of DXT5. This fixes artifacts caused by the R channel being quantized. 
Also sets the compression type to RGTC_R for textures which only use the R channel.

Comparison:
| master | PR |
|--------|----|
| aa | bb |

### DIFF
--- a/thirdparty/etcpak/ProcessRgtc.cpp
+++ b/thirdparty/etcpak/ProcessRgtc.cpp
@@ -1,0 +1,106 @@
+// -- GODOT start --
+
+#include "ForceInline.hpp"
+#include "ProcessRgtc.hpp"
+
+#include <assert.h>
+#include <string.h>
+
+static const uint8_t AlphaIndexTable[8] = { 1, 7, 6, 5, 4, 3, 2, 0 };
+
+static etcpak_force_inline uint64_t ProcessAlpha( const uint8_t* src )
+{
+    uint8_t solid8 = *src;
+    uint16_t solid16 = uint16_t( solid8 ) | ( uint16_t( solid8 ) << 8 );
+    uint32_t solid32 = uint32_t( solid16 ) | ( uint32_t( solid16 ) << 16 );
+    uint64_t solid64 = uint64_t( solid32 ) | ( uint64_t( solid32 ) << 32 );
+    if( memcmp( src, &solid64, 8 ) == 0 && memcmp( src+8, &solid64, 8 ) == 0 )
+    {
+        return solid8;
+    }
+
+    uint8_t min = src[0];
+    uint8_t max = min;
+    for( int i=1; i<16; i++ )
+    {
+        const auto v = src[i];
+        if( v > max ) max = v;
+        else if( v < min ) min = v;
+    }
+
+    uint32_t range = ( 8 << 13 ) / ( 1 + max - min );
+    uint64_t data = 0;
+    for( int i=0; i<16; i++ )
+    {
+        uint8_t a = src[i] - min;
+        uint64_t idx = AlphaIndexTable[( a * range ) >> 13];
+        data |= idx << (i*3);
+    }
+
+    return max | ( min << 8 ) | ( data << 16 );
+}
+
+void CompressRgtcR(const uint32_t *src, uint64_t *dst, uint32_t blocks, size_t width) 
+{
+	int i = 0;
+	auto ptr = dst;
+	do 
+	{
+		uint32_t rgba[4 * 4];
+		uint8_t r[4 * 4];
+
+		auto tmp = (char *)rgba;
+		memcpy(tmp, src + width * 0, 4 * 4);
+		memcpy(tmp + 4 * 4, src + width * 1, 4 * 4);
+		memcpy(tmp + 8 * 4, src + width * 2, 4 * 4);
+		memcpy(tmp + 12 * 4, src + width * 3, 4 * 4);
+		src += 4;
+		if (++i == width / 4) 
+		{
+			src += width * 3;
+			i = 0;
+		}
+
+		for (int i = 0; i < 16; i++) 
+		{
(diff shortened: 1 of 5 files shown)


thirdparty/etcpak/ProcessRgtc.cpp
```diff
@@ -0,0 +1,107 @@
+/* Godot start */
```
Calinou, line 1:
```suggestion
// -- GODOT start --
```

thirdparty/etcpak/ProcessRgtc.hpp
```diff
@@ -0,0 +1,14 @@
+/* Godot start */
```
Calinou, line 1:
```suggestion
// -- GODOT start --
```

thirdparty/etcpak/ProcessRgtc.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+#include <stddef.h>
+#include <stdint.h>
+
+void CompressRgtcR(const uint32_t *src, uint64_t *dst, uint32_t blocks, size_t width);
+void CompressRgtcRG(const uint32_t *src, uint64_t *dst, uint32_t blocks, size_t width);
+
+#endif
+
+/* Godot end */
```
Calinou, line 14:
```suggestion
// -- GODOT end --
```

thirdparty/etcpak/ProcessRgtc.cpp
```diff
@@ -0,0 +1,107 @@
...
+			r[i] = rgba[i] & 0x000000FF;
+			g[i] = (rgba[i] & 0x0000FF00) >> 8;
+		}
+		*ptr++ = ProcessAlpha(r);
+		*ptr++ = ProcessAlpha(g);
+	} 
+	while (--blocks);
+}
+
+/* Godot end */
```
Calinou, line 107:
```suggestion
// -- GODOT end --
```

thirdparty/etcpak/ProcessRgtc.cpp
```diff
@@ -0,0 +1,107 @@
+// -- GODOT start --
+
+#include "ForceInline.hpp"
+#include "ProcessRgtc.hpp"
+
+#include <assert.h>
+#include <stdint.h>
```
akien-mga, line 7: Already in the header, so shouldn't be needed here.

akien-mga:
Have you considered proposing these changes upstream to https://github.com/wolfpld/etcpak/?

Calinou (approved):
Tested locally (rebased on top of `master` 2f73a059cefadcd944b6874f2557ec82e46a562d), it works as expected. Normal maps in https://github.com/godotengine/tps-demo no longer exhibit artifacts when viewed from a distance or at oblique angles.

Before | After *(this PR)*
-|-
Screenshot_20231206_152310 | Screenshot_20231206_152149

BlueCube3310:
> Have you considered proposing these changes upstream to https://github.com/wolfpld/etcpak/?

I did, but first I would like to implement SSE4 variants of the code for faster compression when available. I may do it sometime in the near future.

Saul2022:
Sorry if i it is  a dumb  question, but will this be enabled by default on vram compresion? Anyways great job, i’vbeen waiting for this.

clayjohn (approved):
Looks good to me. 

As mentioned above, it would be really nice to have SSE4 versions of the code, and for this to be upstreamed. But I think we should merge what we have in the meantime as the quality improvement is really important

BlueCube3310:
> As mentioned above, it would be really nice to have SSE4 versions of the code, and for this to be upstreamed. But I think we should merge what we have in the meantime as the quality improvement is really important

Perhaps the Godot-specific code from this and the modified `squish` decompressor could be moved to its own module one day, since RGTC isn't really considered to be part of the S3TC family?

YuriSizov:
Thanks!

As far as I understand the added files are not technically thirdparty and are fully original? Not sure if this is ideal, but I guess it lets us keep everything nicely organized (and yeah, maybe it can be upstreamed at some point). And since they are completely original, we don't need to mention them in `/thirdparty/README`.

BlueCube3310:
> As far as I understand the added files are not technically thirdparty and are fully original? Not sure if this is ideal, but I guess it lets us keep everything nicely organized (and yeah, maybe it can be upstreamed at some point). And since they are completely original, we don't need to mention them in `/thirdparty/README`.

The files aren't completely original as they use the code from the library's DXT5 alpha compressor, so maybe that should be mentioned?

YuriSizov:
In that case, I guess yes. After the "Files extracted from upstream source:" list you could add something like

Two files (`ProcessRgtc.{cpp,hpp}`) have been added to provide RGTC compression implementation, based on library's `ProcessDxtc.{cpp,hpp}`.

PR #85845, opened 2023-12-06
Ensure more export errors are reported to users

### LINKED ISSUE #85814: Export to web fails with "Attempting to make child window exclusive" error
### Tested versions

Godot Engine v4.2.stable.official.46dc27791

### System information

Windows 10

### Issue description

When exporting to web, Godot fails with this error in the console:

> scene/main/window.cpp:884 - Attempting to make child window exclusive, but the parent window already has another exclusive child. This window: /root/@EditorNode@17120/@Panel@13/@ProjectExportDialog@559/@VBoxContainer@137/@AcceptDialog@411, parent window: /root/@EditorNode@17120/@Panel@13/@ProjectExportDialog@559, current exclusive child window: /root/@EditorNode@17120/@Panel@13/@ProjectExportDialog@559/@ProgressDialog@17444

a

This happens for newly created empty projects.

### Steps to reproduce

1. Open the attached project (or create a new project).
2. Project->Export->Export All
3. Observe the console error.  No output files are generated.

[export_test.zip](https://github.com/godotengine/godot/files/13574810/export_test.zip)

YuriSizov:
So apparently we don't have a good habit of adding export status messages when errors happen. This can lead to the export process failing without any concrete errors in the export log. This PR tries to address it to the best of my ability. In case some messages are still underreported, I've added a fallback error message to tell the users _something_.

godot windows editor dev x86_64_2023-12-06_13-47-48

It also fixes the timing issue when exporting all presets at the same time, where the error report would try to appear while the progress dialog was still visible. This leads to reports like https://github.com/godotengine/godot/issues/85814 focusing their attention on an irrelevant message.

----

Closes https://github.com/godotengine/godot/issues/85814, although I don't know what specifically blocks the users. But now, hopefully, they'll get a proper error message. It's probably cherry-pickable, but some edited code is messy so I would like a good review and some time for it to be tested before we consider that.

### DIFF
--- a/platform/android/export/export_plugin.cpp
+++ b/platform/android/export/export_plugin.cpp
@@ -2782,6 +2782,12 @@ Error EditorExportPlatformAndroid::export_project(const Ref<EditorExportPreset> 
 Error EditorExportPlatformAndroid::export_project_helper(const Ref<EditorExportPreset> &p_preset, bool p_debug, const String &p_path, int export_format, bool should_sign, int p_flags) {
 	ExportNotifier notifier(*this, p_preset, p_debug, p_path, p_flags);
 
+	const String base_dir = p_path.get_base_dir();
+	if (!DirAccess::exists(base_dir)) {
+		add_message(EXPORT_MESSAGE_ERROR, TTR("Export"), vformat(TTR("Target folder does not exist or is inaccessible: \"%s\""), base_dir));
+		return ERR_FILE_BAD_PATH;
+	}
+
 	String src_apk;
 	Error err;
 
@@ -2856,7 +2862,10 @@ Error EditorExportPlatformAndroid::export_project_helper(const Ref<EditorExportP
 		}
 		const String assets_directory = get_assets_directory(p_preset, export_format);
 		String sdk_path = EDITOR_GET("export/android/android_sdk_path");
-		ERR_FAIL_COND_V_MSG(sdk_path.is_empty(), ERR_UNCONFIGURED, "Android SDK path must be configured in Editor Settings at 'export/android/android_sdk_path'.");
+		if (sdk_path.is_empty()) {
+			add_message(EXPORT_MESSAGE_ERROR, TTR("Export"), TTR("Android SDK path must be configured in Editor Settings at 'export/android/android_sdk_path'."));
+			return ERR_UNCONFIGURED;
+		}
 		print_verbose("Android sdk path: " + sdk_path);
 
 		// TODO: should we use "package/name" or "application/config/name"?
@@ -3106,10 +3115,6 @@ Error EditorExportPlatformAndroid::export_project_helper(const Ref<EditorExportP
 		}
 	}
 
-	if (!DirAccess::exists(p_path.get_base_dir())) {
-		return ERR_FILE_BAD_PATH;
-	}
-
 	Ref<FileAccess> io_fa;
 	zlib_filefunc_def io = zipio_create_io(&io_fa);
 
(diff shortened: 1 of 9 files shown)


platform/android/export/export_plugin.cpp
```diff
@@ -2856,7 +2862,10 @@ Error EditorExportPlatformAndroid::export_project_helper(const Ref<EditorExportP
 		}
 		const String assets_directory = get_assets_directory(p_preset, export_format);
 		String sdk_path = EDITOR_GET("export/android/android_sdk_path");
-		ERR_FAIL_COND_V_MSG(sdk_path.is_empty(), ERR_UNCONFIGURED, "Android SDK path must be configured in Editor Settings at 'export/android/android_sdk_path'.");
+		if (sdk_path.is_empty()) {
+			add_message(EXPORT_MESSAGE_ERROR, TTR("Export"), TTR("Android SDK path must be configured in Editor Settings at 'export/android/android_sdk_path'."));
+			return ERR_UNCONFIGURED;
+		}
```
YuriSizov, line 2868: I'm not sure we needed to use error macros in such places to begin with, but adding proper log messages will call `ERR_PRINT` for us anyway, so they can be removed.

platform/android/export/export_plugin.cpp
```diff
@@ -3302,10 +3307,6 @@ Error EditorExportPlatformAndroid::export_project_helper(const Ref<EditorExportP
 	zipClose(unaligned_apk, nullptr);
 	unzClose(pkg);
 
-	if (err != OK) {
-		CLEANUP_AND_RETURN(err);
-	}
```
YuriSizov, line 3307: Doesn't seem to be necessary, we never reassign `err` since the last check above.

platform/windows/export/export_plugin.cpp
```diff
@@ -242,16 +243,22 @@ Error EditorExportPlatformWindows::export_project(const Ref<EditorExportPreset>
 	if (embedded) {
 		pck_path = pck_path.get_basename() + ".tmp";
 	}
+
 	Error err = EditorExportPlatformPC::export_project(p_preset, p_debug, pck_path, p_flags);
-	if (p_preset->get("codesign/enable") && err == OK) {
+	if (err != OK) {
+		// Message is supplied by the subroutine method.
+		return err;
+	}
```
YuriSizov, line 251: Don't see a point in trying to export ZIP if we failed here already. Don't do it on Linux either.

m4gr3d (approved):
The Android section looks good!

platform/ios/export/export_plugin.cpp
```diff
@@ -1617,8 +1634,14 @@ Error EditorExportPlatformIOS::_export_project_helper(const Ref<EditorExportPres
...
 	}
 
-	err = _export_ios_plugins(p_preset, config_data, dest_dir + binary_name, assets, p_debug);
-	ERR_FAIL_COND_V(err, err);
+	err = _export_ios_plugins(p_preset, config_data, binary_dir, assets, p_debug);
+	if (err != OK) {
+		// TODO: Improve error reporting by using `add_message` throughout all methods called via `_export_ios_plugins`.
+		// For now a generic top level message would be fine, but we're ought to use proper reporting here instead of
+		// just fail macros and non-descriptive error return values.
+		add_message(EXPORT_MESSAGE_ERROR, TTR("iOS Plugins"), vformat(TTR("Failed to export iOS plugins with code %d. Please check the output log."), err));
```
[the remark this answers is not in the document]
bruvzg: All the error conditions in the `_export_ios_plugins` seems to be forwarding error from `_copy_asset` (which have two error returns), all of them probably should not be ERR_ macros (since it's just adding unnecessary and non-informative messages) and should be changed to `if (err != OK) return err;`.

bruvzg (approved):

YuriSizov:
Thanks for reviews!

PR #85846, opened 2023-12-06
Fix `find_children` doesn't work with class names

### LINKED ISSUE #85690: Find Childern Does Not Work In C#
### Godot version

4.2

### System information

Windows 11 4.2

### Issue description

When running 
Godot.Collections.Array<Node> t = FindChildren("*");
GD.Print(t);
I receve an empty array even when i have nodes under a base node
image

Expected output:
A list of children nodes in this case 7 of them.

### Steps to reproduce

create a node structure with a base node and some children nodes
image
put 
`
public override void _Ready()
	{
		Godot.Collections.Array<Node> t = FindChildren("*");
		GD.Print(t);
	}
`
in the script and run the project

[Minimum project.zip](https://github.com/godotengine/godot/files/13537497/Minimum.project.zip)

nongvantinh:
`Script::get_name()` and `Script::get_path()` of inner classes return empty values. I'm unsure whether this is a bug with scripting languages or by design. This PR only addresses the bug for the outer class.

Fix: #85690

### DIFF
--- a/scene/main/node.cpp
+++ b/scene/main/node.cpp
@@ -45,6 +45,7 @@
 #include "scene/scene_string_names.h"
 #include "viewport.h"
 
+#include <modules/gdscript/gdscript.h>
 #include <stdint.h>
 
 int Node::orphan_node_count = 0;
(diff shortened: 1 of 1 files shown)


scene/main/node.cpp
```diff
@@ -1696,7 +1696,7 @@ TypedArray<Node> Node::find_children(const String &p_pattern, const String &p_ty
 			} else if (cptr[i]->get_script_instance()) {
 				Ref<Script> scr = cptr[i]->get_script_instance()->get_script();
 				while (scr.is_valid()) {
-					if ((ScriptServer::is_global_class(p_type) && ScriptServer::get_global_class_path(p_type) == scr->get_path()) || p_type == scr->get_path()) {
+					if ((ScriptServer::is_global_class(p_type) && ScriptServer::get_global_class_path(p_type) == scr->get_path()) || scr->get_path().contains(p_type)) {
```
YuriSizov, line 1699: I don't think this is safe to do under any circumstances. You can match any unrelated script this way, just because it has a similar substring.

YuriSizov: Yes, that was a part of my point. It's not a corner case. It's a fairly major case, IMO.

vnen (changes requested):

PR #85856, opened 2023-12-06
Make 2D navigation mesh baking parse all TileMapLayers

### LINKED ISSUE #85798: NavigationRegion2D doesn't take into account tiles with collision on other layers besides the first one
### Tested versions

Reproducible in 4.2.stable

### System information

Windows 11 - Godot v4.2.stable - Vulkan(Forward+/Mobile)

### Issue description

When using the NavigationRegion2D, the NavigationPolygon that is baked is incorrect.

### Steps to reproduce

1. Create a 2D scene
2. Add NavigationRegion2D
3. Add NavigationPolygon to NavigationRegion2D
4. Add TileMap (add a tileset, any will do as long as it has some tiles with PhysicsLayer painted in) child node to NavigationRegion2D
5. Add a second layer to the TileMap
6. Paint on first layer a region with tiles that have collision
7. Paint on the second layer a region with tiles that have collision
8. Bake NavigationPolygon

[TileMapNavTest.zip](https://github.com/godotengine/godot/files/13568676/TileMapNavTest.zip)

Just select the NavigationRegion2D and press Bake NavigationPolygon

smix8:
Makes 2D navigation mesh baking parse all TileMapLayers.

This lifts the limitation of the 2D navigation mesh baking only parsing the first TileMapLayer, due to popular demand.

Fixes https://github.com/godotengine/godot/issues/85798 and all the other billion linked issues of the same kind.

Note that this change is intended to enable workflows where users have their parsed polygons on another TileMapLayer than the first. The 2D navigation mesh baking still treats the Tilemap data as if it is done with a single TileMapLayer, it only now parses data from all TileMapLayers, picks the first valid data for a cell found, and combines them all together.

The parsing process will try to block the most common user errors by ignoring all TileMap cell keys that already had a navigation mesh or collision shape that fits the collision mask parsed and print a warning instead.
- Having navigation mesh for the same cell on 2 or more TileMapLayers is an error.
- Having collision polygons with the parsed collision mask for the same cell on 2 or more TileMapLayers is an error.

For navigation polygons the build and placement rules are no surface overlap at all is allowed on the same navigation map. The only thing that can "overlap" between 2 navigation meshes are unique pairs of edges in order to merge them.

For collision polygons the build and placement rules are no overlap if it has a physics layer bit that is parsed by the NavigationPolygon collision mask. This means if users have 2 TileMapLayers with collision polygons placed in the same cell, 1 intended  for the navigation mesh parsing and 1 intended for e.g. gameplay the gameplay TileMapLayer should not have a matching bit on the collision layer with the NavigationPolygon parse collision mask, else it will be treated as duplicates with errors and rejected cell data.

### DIFF
--- a/modules/navigation/2d/nav_mesh_generator_2d.cpp
+++ b/modules/navigation/2d/nav_mesh_generator_2d.cpp
@@ -575,8 +575,6 @@ void NavMeshGenerator2D::generator_parse_tilemap_node(const Ref<NavigationPolygo
 		return;
 	}
 
-	int tilemap_layer = 0; // only main tile map layer is supported
-
 	Ref<TileSet> tile_set = tilemap->get_tileset();
 	if (!tile_set.is_valid()) {
 		return;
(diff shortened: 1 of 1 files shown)


InfiniteCactusDev:
Just out of interest. Why aren't you first combining (union in Clipper) all the layer-polygons before using them for navmeshbaking?

modules/navigation/2d/nav_mesh_generator_2d.cpp
```diff
@@ -587,74 +585,96 @@ void NavMeshGenerator2D::generator_parse_tilemap_node(const Ref<NavigationPolygo
...
+			tile_transform.set_origin(tilemap->map_to_local(cell));
+
+			const Transform2D tile_transform_offset = tilemap_xform * tile_transform;
+
+			if (navigation_layers_count > 0) {
+				Ref<NavigationPolygon> navigation_polygon = tile_data->get_navigation_polygon(tilemap_layer, flip_h, flip_v, transpose);
+				if (navigation_polygon.is_valid()) {
+					if (cells_with_navigation_polygon.has(cell)) {
+#ifdef DEBUG_ENABLED
+						WARN_PRINT(vformat("TileMap navigation mesh baking error. The TileMap cell key Vector2i(%s, %s) has navigation mesh from 2 or more different TileMap layers assigned. This can cause unexpected navigation mesh baking results. The duplicated cell data was ignored.", cell.x, cell.y));
```
Scony, line 620: wouldn't it spam potentially

smix8: Added a max print count.

Scony (approved):

akien-mga:
Thanks!

874808039:
Thanks, Bug this  new  TilemapLayer node is  not working(https://github.com/godotengine/godot/pull/89179),The new TilemapLayer node cannot yet cooperate with NavigationRegion2D and cannot bake navigation polygons.Can this function be synchronized on the new node?

xynanlee:
Does this PR solves the long standing TileMap navigation meshes "cutout" issue? I tried TileMapLayer a bit and it seems not the case.

874808039:
> Does this PR solves the long standing TileMap navigation meshes "cutout" issue? I tried TileMapLayer a bit and it seems not the case.

If you click to bake, will it be successful?
image

xynanlee:
> > Does this PR solves the long standing TileMap navigation meshes "cutout" issue? I tried TileMapLayer a bit and it seems not the case.
> 
> If you click to bake, will it be successful? image

I'm not using `NavigationRegion2D` at all. There has been a long standing issue where tilemap navigation layers unable to be "cut out" by physics layers in other tilemap layers. Looking at this PR, I'm getting the impression of 1 TileMapLayer with navigation layer can parse multiple TileMapLayers with physic layers to do the "cutout". I tried it on myside and it doesn't seemed to be the case.

LeifintheWind:
Does this include changes for tilemap Physics Layers?

At the moment in 4.2 it appears as though no matter the collision mask parsed by the NavigationPolygon, only the first collision mask is used. 

As an example, I currently have two collision layers used in my tilemap: one for preventing ground navigation, one for preventing flying. My first Physics Layer has just the ground nav collision, while the second has ground and flying, and the third has just flying. When attempting to bake the nav polygon for walking it seems to only bake according to that first Physics Layer, and ignores the second one.

RuinRui:
I checked many navigation issues in the official forums, and the conclusion is that the 2D navigation layer is just a series of polygons, without the concept of layers, including the physics layer。
so even if it is solved, it is only by cutting off all merged physical layers before merging the navigation？
This fix only fixes the problem that only the first navigation layer on multiple tilemap layers is effective, and it is achieved by merging the navigation polygons in all navigation layers on the tilemap. This is my understanding.

PR #85861, opened 2023-12-06
Add sorting script previews by modified date

### LINKED ISSUE #8602: Sort script list by the scripts' modification date

Any game or plugin with some amount of scripting involved.

I often end up with a lot of open scripts populating the side scripts preview window next to the code editor. I don't want to close them all the time, but as the number of items in the small window grows, it's hard to stay organized. When working with multiple files simultaneously (like developing a class hierarchy), it can be hard to navigate to the correct files in the preview.

I'd like to add a new type of sorting to the script previews - sort by modified date. This would keep the most recently modified files at the top of the preview list.

When a script is saved, it jumps to the top of the preview list.

https://github.com/godotengine/godot-proposals/assets/49688205/72455a67-c13c-410a-a260-8a9675bac9d5

I actually put the horse before the cart by mistake here and implemented the feature in [this pull request](https://github.com/godotengine/godot/pull/85861). That's also where the video is from.

This cannot be done with an external script.

I don't think this is possible with an add-on.

VojtaStruhar:
Adds an option for sorting the scripts preview by modified time (descending), so that the latest modified files appear on top.

The default sorting by name works well, but often I have a lot of scripts open and I want to quickly jump between the last few I'm working on without having to remember which one is which. This pull request adds this functionality. In the video I modify one of the scripts, press `Cmd + S` and it jumps to the top of the preview.

https://github.com/godotengine/godot/assets/49688205/7474c0c7-1644-4b10-bfd2-b9e1953f371c

## Consideration: The scripts are sorted on save

As you can see in the code changes, the _actual sorting_ of the script items inside the preview now happens every time the script is saved (`ScriptEditor::_res_saved_callback`). For the other sorting methods this is absolutely **not** needed, but is performed anyway. I think this is not a performance bottleneck (there will likely not be _thousands_ of scripts in the preview at any one time and they are saved/sorted on demand with a user action). Since this is an editor thing, it doesn't affect games at all.

One solution to this would be to check if we are in the _SORT_BY_MODIFIED_ mode and only perform the sorting then, but that seems dirty to me. Stray if conditions like that are not very maintainable imo.

---

**This is my first real contribution**, so please double check if the changes are sane. If naming is OK, if I'm using things for what they are supposed to be used, etc.

- *Production edit: This closes https://github.com/godotengine/godot-proposals/issues/8602.*

### DIFF
--- a/editor/plugins/script_editor_plugin.cpp
+++ b/editor/plugins/script_editor_plugin.cpp
@@ -2065,7 +2065,7 @@ void ScriptEditor::_update_script_names() {
 					sd.sort_key = path;
 				} break;
 				case SORT_BY_MODIFIED: {
-					// XOR with UINT64_MAX results in descending order
+					// XOR with UINT64_MAX results in descending order.
 					uint64_t modified_time_desc = se->get_edited_resource()->get_last_modified_time() ^ UINT64_MAX;
 					sd.sort_key = String::num_uint64(modified_time_desc);
 				} break;
--- a/editor/plugins/script_editor_plugin.cpp
+++ b/editor/plugins/script_editor_plugin.cpp
@@ -1000,6 +1000,7 @@ void ScriptEditor::_res_saved_callback(const Ref<Resource> &p_res) {
 		}
 	}
 
+	_sort_list_on_update = true;
 	_update_script_names();
 	trigger_live_script_reload();
 }
@@ -2062,6 +2063,11 @@ void ScriptEditor::_update_script_names() {
 				} break;
 				case SORT_BY_PATH: {
 					sd.sort_key = path;
+				} break;
+				case SORT_BY_MODIFIED: {
+					// XOR with UINT64_MAX results in descending order
+					uint64_t modified_time_desc = se->get_edited_resource()->get_last_modified_time() ^ UINT64_MAX;
+					sd.sort_key = String::num_uint64(modified_time_desc);
 				} break;
 				case SORT_BY_NONE: {
 					sd.sort_key = "";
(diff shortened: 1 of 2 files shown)


Calinou:
Thanks for opening a pull request :slightly_smiling_face: 

Feature pull requests should be associated to a *feature proposal* to justify the need for implementing the feature in Godot core. Please open a proposal on the [Godot proposals](https://github.com/godotengine/godot-proposals) repository (and link to this pull request in the proposal's body).

VojtaStruhar:
Sorry about that! Will do

editor/plugins/script_editor_plugin.cpp
```diff
@@ -2063,6 +2064,11 @@ void ScriptEditor::_update_script_names() {
 				case SORT_BY_PATH: {
 					sd.sort_key = path;
 				} break;
+				case SORT_BY_MODIFIED: {
+					// XOR with UINT64_MAX results in descending order
```
AThousandShips, line 2068:
```suggestion
					// XOR with UINT64_MAX results in descending order.
```

editor/plugins/script_editor_plugin.h
```diff
@@ -255,6 +255,7 @@ class ScriptEditor : public PanelContainer {
 	enum ScriptSortBy {
 		SORT_BY_NAME,
 		SORT_BY_PATH,
+		SORT_BY_MODIFIED,
```
AThousandShips, line 258:
```suggestion
		SORT_BY_MODIFIED_TIME,
```
To match `FilesystemDock`

editor/plugins/script_editor_plugin.cpp
```diff
@@ -4373,7 +4379,7 @@ ScriptEditorPlugin::ScriptEditorPlugin() {
 	EDITOR_DEF("text_editor/script_list/script_temperature_enabled", true);
 	EDITOR_DEF("text_editor/script_list/script_temperature_history_size", 15);
 	EDITOR_DEF("text_editor/script_list/group_help_pages", true);
-	EditorSettings::get_singleton()->add_property_hint(PropertyInfo(Variant::INT, "text_editor/script_list/sort_scripts_by", PROPERTY_HINT_ENUM, "Name,Path,None"));
+	EditorSettings::get_singleton()->add_property_hint(PropertyInfo(Variant::INT, "text_editor/script_list/sort_scripts_by", PROPERTY_HINT_ENUM, "Name,Path,Modified,None"));
```
AThousandShips, line 4382:
```suggestion
	EditorSettings::get_singleton()->add_property_hint(PropertyInfo(Variant::INT, "text_editor/script_list/sort_scripts_by", PROPERTY_HINT_ENUM, "Name,Path,Last Modified Time,None"));
```
Same, and more clear

VojtaStruhar:
@AThousandShips thanks for the feedback!!!

editor/plugins/script_editor_plugin.cpp
```diff
@@ -2063,6 +2064,11 @@ void ScriptEditor::_update_script_names() {
 				case SORT_BY_PATH: {
 					sd.sort_key = path;
 				} break;
+				case SORT_BY_MODIFIED: {
```
AThousandShips, line 2067:
```suggestion
				case SORT_BY_MODIFIED_TIME: {
```
You forgot to update this one

VojtaStruhar: Ooops, missed one. Thanks.

VojtaStruhar:
I just realized, that the fact it's sorting on _every script save_ will interfere with the `Alt + Shift + Up/Down` keybind that moves the current script up or down in the script preview window. This is definitely not what we want for other sorting styles. Any suggestions for an elegant solution to this?
- Adding another settings checkbox to _sort on script update_ is overkill
- Stray standalone if check for this sorting mode doesn't feel right to me :/

PR #13025, opened 2017-11-18
Fix #12220: Add Decompress Bc5 to Squish

gad-o:
This Commit fixes the corrupted file preview described in #12220.
Added DecompressColourBc5 function to squish.

PR #85863, opened 2023-12-06
Fix squish RGTC_R decompression corruption

BlueCube3310:
The custom Godot code for `squish` only allows for decompressing BC5 (RGTC_RG) textures. This means attempting to decompress BC4-compressed images (for the minature texture views, for example) would result in corrupted images:

| master | PR |
|--------|----|
| aa |bb |

An MRP for testing the changes: [RGTCSquishError.zip](https://github.com/godotengine/godot/files/13590517/RGTCSquishError.zip)
Steps to reproduce:
1. Load the project and open Main.tscn
2. Select the mesh instance and look at the Albedo texture preview in the material inspector.

### DIFF
--- a/thirdparty/squish/colourblock.cpp
+++ b/thirdparty/squish/colourblock.cpp
@@ -24,9 +24,9 @@
    -------------------------------------------------------------------------- */
 
 #include "colourblock.h"
-// -- Godot start --
+// -- GODOT start --
 #include "alpha.h"
-// -- Godot end --
+// -- GODOT end --
 
 namespace squish {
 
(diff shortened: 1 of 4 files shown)


clayjohn:
Oh wow, doing this was brought up 6 years ago... https://github.com/godotengine/godot/pull/13025#issuecomment-345481407

clayjohn (approved):
Looks good to me. Tested locally and it works. 

Screenshot from 2023-12-06 13-35-19

One thing to note is that the generated preview for the file system doesn't update (even after deleting the .godot folder). So some users may be surprised to see the corruption in the file system preview

Screenshot from 2023-12-06 13-35-24

akien-mga:
> One thing to note is that the generated preview for the file system doesn't update (even after deleting the .godot folder). So some users may be surprised to see the corruption in the file system preview

File thumbnails are stored outside the project, in the user cache folder. On Linux it's in `~/.cache/godot/`.

akien-mga:
You missed the first chunk of my diff changing the style of the "Godot changes" comment at the top of `thirdparty/squish/colourblock.cpp`.

akien-mga (approved):

YuriSizov:
Thanks!

PR #85868, opened 2023-12-06
Add descriptions for tile properties

KoBeWi:
Addresses https://github.com/godotengine/godot-proposals/issues/7177#issuecomment-1681204726
image

The PR adds a new method to DocTools called `add_custom_property_description()`, which allows adding a description for arbitrary class/property. This is done outside docs, because these properties are not supposed to be documented (in some cases it's not even possible, like with TileSetAtlasSourceProxyObject), but it's still nice if they had a description in tooltip. This is now possible and likely can be expanded onto dynamic properties (e.g. ItemList's items). While adding documentation descriptions inside editor code is controversial, VisualShader already does it for its various nodes ([example](https://github.com/godotengine/godot/blob/2f73a059cefadcd944b6874f2557ec82e46a562d/editor/plugins/visual_shader_editor_plugin.cpp#L5643)).

The PR uses the new functionality in TileSet editor, for various properties. Putting as draft until I fill in all descriptions.
Also I tweaked ID property to be real EditorProperty, otherwise tooltip was not available.
image

_Production edit: closes godotengine/godot-roadmap#80_

### DIFF
--- a/editor/plugins/tiles/tile_set_atlas_source_editor.cpp
+++ b/editor/plugins/tiles/tile_set_atlas_source_editor.cpp
@@ -2406,7 +2406,36 @@ void TileSetAtlasSourceEditor::_auto_remove_tiles() {
 
 void TileSetAtlasSourceEditor::_notification(int p_what) {
 	switch (p_what) {
-		case NOTIFICATION_ENTER_TREE:
+		case NOTIFICATION_READY: {
+			atlas_source_inspector->edit(atlas_source_proxy_object);
+			atlas_source_inspector->add_custom_property_description("TileSetAtlasSourceProxyObject", "id", TTR("The tile's unique identifier within this TileSet. Each tile stores its source ID, so changing one may make tiles invalid."));
+			atlas_source_inspector->add_custom_property_description("TileSetAtlasSourceProxyObject", "name", TTR("The human-readable name for the atlas. Use a descriptive name here for organizational purposes (such as \"terrain\", \"decoration\", etc.)."));
+			atlas_source_inspector->add_custom_property_description("TileSetAtlasSourceProxyObject", "texture", TTR("The image from which the tiles will be created."));
+			atlas_source_inspector->add_custom_property_description("TileSetAtlasSourceProxyObject", "margins", TTR("The margins on the image's edges that should not be selectable as tiles (in pixels). Increasing this can be useful if you download a tilesheet image that has margins on the edges (e.g. for attribution)."));
+			atlas_source_inspector->add_custom_property_description("TileSetAtlasSourceProxyObject", "separation", TTR("The separation between each tile on the atlas in pixels. Increasing this can be useful if the tilesheet image you're using contains guides (such as outlines between every tile)."));
+			atlas_source_inspector->add_custom_property_description("TileSetAtlasSourceProxyObject", "texture_region_size", TTR("The size of each tile on the atlas in pixels. In most cases, this should match the tile size defined in the TileMap property (although this is not strictly necessary)."));
+			atlas_source_inspector->add_custom_property_description("TileSetAtlasSourceProxyObject", "use_texture_padding", TTR("If checked, adds a 1-pixel transparent edge around each tile to prevent texture bleeding when filtering is enabled. It's recommended to leave this enabled unless you're running into rendering issues due to texture padding."));
+
+			tile_inspector->edit(tile_proxy_object);
+			tile_inspector->add_custom_property_description("AtlasTileProxyObject", "atlas_coords", TTR("The position of the tile's top-left corner in the atlas. The position and size must be within the atlas and can't overlap another tile.\nEach painted tile has associated atlas coords, so changing this property may cause your TileMaps to not display properly."));
+			tile_inspector->add_custom_property_description("AtlasTileProxyObject", "size_in_atlas", TTR("The unit size of the tile."));
+			tile_inspector->add_custom_property_description("AtlasTileProxyObject", "animation_columns", TTR("Number of columns for the animation grid. If number of columns is lower than number of frames, the animation will automatically adjust row count."));
+			tile_inspector->add_custom_property_description("AtlasTileProxyObject", "animation_separation", TTR("The space (in tiles) between each frame of the animation."));
+			tile_inspector->add_custom_property_description("AtlasTileProxyObject", "animation_speed", TTR("Animation speed in frames per second."));
+			tile_inspector->add_custom_property_description("AtlasTileProxyObject", "animation_mode", TTR("Determines how animation will start. In \"Default\" mode all tiles start animating at the same frame. In \"Random Start Times\" mode, each tile starts animation with a random offset."));
+			tile_inspector->add_custom_property_description("AtlasTileProxyObject", "flip_h", TTR("If [code]true[/code], the tile is horizontally flipped."));
+			tile_inspector->add_custom_property_description("AtlasTileProxyObject", "flip_v", TTR("If [code]true[/code], the tile is vertically flipped."));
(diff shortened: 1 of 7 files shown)


editor/plugins/tiles/tile_set_editor.cpp
```diff
@@ -987,7 +987,7 @@ void TileSourceInspectorPlugin::_show_id_edit_dialog(Object *p_for_source) {
 
 void TileSourceInspectorPlugin::_confirm_change_id() {
 	edited_source->set("id", id_input->get_value());
-	id_label->set_text(vformat(TTR("ID: %d"), edited_source->get("id"))); // Use get(), because the provided ID might've been invalid.
+	id_label->set_text(itos(edited_source->get("id"))) ; // Use get(), because the provided ID might've been invalid.
```
akien-mga, line 990:
```suggestion
	id_label->set_text(itos(edited_source->get("id"))); // Use get(), because the provided ID might've been invalid.
```

editor/plugins/tiles/tile_set_atlas_source_editor.cpp
```diff
@@ -2406,7 +2407,35 @@ void TileSetAtlasSourceEditor::_auto_remove_tiles() {
 
 void TileSetAtlasSourceEditor::_notification(int p_what) {
 	switch (p_what) {
-		case NOTIFICATION_ENTER_TREE:
+		case NOTIFICATION_READY: {
+			atlas_source_inspector->edit(atlas_source_proxy_object);
+
+			DocTools *dd = EditorHelp::get_doc_data();
+			dd->add_custom_property_description("TileSetAtlasSourceProxyObject", "id", TTR("The identifier unique within this TileSet. Each tile stores its source ID, so changing one may make tiles invalid."));
```
akien-mga, line 2414:
```suggestion
			dd->add_custom_property_description("TileSetAtlasSourceProxyObject", "id", TTR("The tile's unique identifier within this TileSet. Each tile stores its source ID, so changing one may make tiles invalid."));
```

editor/plugins/tiles/tile_set_atlas_source_editor.cpp
```diff
@@ -2406,7 +2407,35 @@ void TileSetAtlasSourceEditor::_auto_remove_tiles() {
...
+			dd->add_custom_property_description("TileSetAtlasSourceProxyObject", "separation", TTR("The separation between each tile on the atlas in pixels. Increasing this can be useful if the tilesheet image you're using contains guides (such as outlines between every tile)."));
+			dd->add_custom_property_description("TileSetAtlasSourceProxyObject", "texture_region_size", TTR("The size of each tile on the atlas in pixels. In most cases, this should match the tile size defined in the TileMap property (although this is not strictly necessary)."));
+			dd->add_custom_property_description("TileSetAtlasSourceProxyObject", "use_texture_padding", TTR("If checked, adds a 1-pixel transparent edge around each tile to prevent texture bleeding when filtering is enabled. It's recommended to leave this enabled unless you're running into rendering issues due to texture padding."));
+
+			tile_inspector->edit(tile_proxy_object);
+			dd->add_custom_property_description("AtlasTileProxyObject", "atlas_coords", TTR(""));
```
akien-mga, line 2423: We shouldn't have `TTR("")` in the final version, but I assume those are placeholders to fill with more docs?

editor/plugins/tiles/tile_set_scenes_collection_source_editor.cpp
```diff
@@ -357,6 +358,19 @@ void TileSetScenesCollectionSourceEditor::_update_scenes_list() {
 
 void TileSetScenesCollectionSourceEditor::_notification(int p_what) {
 	switch (p_what) {
+		case NOTIFICATION_READY: {
+			scenes_collection_source_inspector->edit(scenes_collection_source_proxy_object);
+
+			DocTools *dd = EditorHelp::get_doc_data();
+			dd->add_custom_property_description("TileSetScenesCollectionProxyObject", "id", TTR("The identifier unique within this TileSet. Each tile stores its source ID, so changing one may make tiles invalid."));
+			dd->add_custom_property_description("TileSetScenesCollectionProxyObject", "name", TTR("The human-readable name for the scene collection. Use a descriptive name here for organizational purposes (such as \"obstacles\", \"decoration\", etc)."));
```
akien-mga, line 366:
```suggestion
			dd->add_custom_property_description("TileSetScenesCollectionProxyObject", "name", TTR("The human-readable name for the scene collection. Use a descriptive name here for organizational purposes (such as \"obstacles\", \"decoration\", etc.)."));
```

akien-mga:
> The PR adds a new method to DocTools called `add_custom_property_description()`, which allows adding a description for arbitrary class/property. This is done outside docs, because these properties are not supposed to be documented (in some cases it's not even possible, like with TileSetAtlasSourceProxyObject), but it's still nice if they had a description in tooltip. This is now possible and likely can be expanded onto dynamic properties (e.g. ItemList's items). While adding documentation descriptions inside editor code is controversial, VisualShader already does it for its various nodes ([example](https://github.com/godotengine/godot/blob/2f73a059cefadcd944b6874f2557ec82e46a562d/editor/plugins/visual_shader_editor_plugin.cpp#L5643)).

That's a bit awkward indeed to have these added manually in the source code, as opposed to extracted for documentation like we do for other class properties. But indeed the SourceProxyObject seems a bit special, and it's still relatively self-contained so it may not be too bad.

groud:
I wonder if it wouldn't be better to be able to expose those internal classes to DocTool somehow, but marking them as not fitting for display in the built documentation. That would be a bit cleaner IMO, as editing this documentation would better be in the XML.

Though, to be fair, I don't know if it deserves the additional work. I don't think there's a lot of places where we use that kind of proxy object. However, I think there's a lot of value in reusing the EditorInspector code (it could probably be done in other places), so maybe that would be a good future-proofing anyway ?

groud:
>  It also means we cannot expose this documentation online, so it cannot be referenced by users or used naturally in tutorials.

I don't think this is doable though. The TileDataProxyObject isn't supposed to be exposed, and I think there would be no sense in presenting it in the reference of a class that cannot be instantiated. How to present it online or even in the built-in documentation seems out of the scope of this PR to me. If we can at least store the data in the XML files, that's would be already good enough IMHO.

YuriSizov:
@groud If we can store the data in an XML file, this means we already have it available in online docs at no cost :)

Being non-instantiable is not an issue, we document many classes which are not supposed to be used directly by users. It's still useful.

groud:
> Being non-instantiable is not an issue, we document many classes which are not supposed to be used directly by users. It's still useful.

But I mean, this makes little sense. The proxy object are just internal stuff, to be able to have the convenience to use the EditorInspector instead of an hand-made GUI (EditorInspector needs an edited object). Exposing this documentation would basically be the equivalent to, for example, document the 2D editor toolbar buttons in the reference. That would be very confusing.

Here, those properties are most of them defined in the proxied object anyway. Like TileSetAtlasSourceProxyObject is basically a TileSetAtlasSource with some added properties (I think there's only its ID in the corresponding TileSet). And same for the TileDataProxyObject. So this documentation kind of already exists in other places.

groud:
> To me it makes little sense to document them but only in the editor. It limits the accessibility of the documentation, so why even bother? You can't look up the information, can't link others to it, can't copy and paste it.

It's just a nice to have as tooltips, that's pretty much it. And I don't think we have any tootips of the editor in the documentation, beside the ones that you can interact with (via code or via the editor). Here you can't access it at all.  TBH, that make me reconsider the fact we should move it to the XML if we have to expose it in the end. I'd rather go for TTR(...) strings in the editor, that would be consistent with other tooltips.

> At any rate, you say that we should put this information in XMLs. I agree. If we do this, it will be available online anyway. So while we don't see this from the same angle, the end result will be the same and the docs will be accessible. Let's focus on "how".

I mean, I don't mind putting it online. But what I don't want is it to appear as a random internal object in the reference. 
So I don't know, if you want a custom page for editor-internal objects (not accessible at all, that should be stated somewhere), we could, but that seems a bit weird IMHO.

YuriSizov:
> I'd rather go for TTR(...) strings in the editor, that would be consistent with other tooltips.

Then I don't think it should be mixed with DocData and displayed as other tooltips related to accessible properties.

KoBeWi:
Ok I pushed solution that doesn't involve DocData.

akien-mga (approved):
Looks good to me with this more ad hoc solution.

Would be good for @groud to also confirm you like this approach.

groud (approved):
Looks perfect to me!

akien-mga:
Thanks!

PR #85869, opened 2023-12-06
.NET: Add a warning in the inspector when properties might be out of sync

paulloz:
Add a small warning in the inspector when C# scripts are out of sync with data we gathered during compilation. The warning is displayed if any C# script in the class hierarchy was modified since the last compilation, and disappears as soon as possible.  

Sadly, the inspector is not refreshed ATM when the window is refocused (e.g. if you tab back from your external code editor). From my tests, this behaviour is not language-dependent, so it'd probably warrant a fix higher up (therefore, I didn't look how to do so in this PR).

I gladly take any suggestion for UI and wording. E.g. someone suggested to directly add a button/link to build directly from the warning. But I'm not sure how I feel about that.  

godot windows editor dev x86_64 mono_ElbJvjZ2Lp

### DIFF
--- a/modules/mono/csharp_script.cpp
+++ b/modules/mono/csharp_script.cpp
@@ -2246,6 +2246,17 @@ bool CSharpScript::_update_exports(PlaceHolderScriptInstance *p_instance_to_upda
 			} else {
 				p_instance_to_update->update(propnames, values);
 			}
+		} else if (placeholders.size()) {
+			uint64_t script_modified_time = FileAccess::get_modified_time(get_path());
+			uint64_t last_valid_build_time = GDMono::get_singleton()->get_project_assembly_modified_time();
+			if (script_modified_time > last_valid_build_time) {
+				for (PlaceHolderScriptInstance *instance : placeholders) {
+					Object *owner = instance->get_owner();
+					if (owner->get_script_instance() == instance) {
+						owner->notify_property_list_changed();
+					}
+				}
+			}
 		}
 	}
 #endif
--- a/modules/mono/editor/GodotTools/GodotTools/GodotSharpEditor.cs
+++ b/modules/mono/editor/GodotTools/GodotTools/GodotSharpEditor.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using GodotTools.Build;
 using GodotTools.Ides;
 using GodotTools.Ides.Rider;
+using GodotTools.Inspector;
 using GodotTools.Internals;
 using GodotTools.ProjectEditor;
 using JetBrains.Annotations;
@@ -45,12 +46,15 @@ namespace GodotTools
 
         // TODO Use WeakReference once we have proper serialization.
         private WeakRef _exportPluginWeak;
+        private WeakRef _inspectorPluginWeak;
 
         public GodotIdeManager GodotIdeManager { get; private set; }
 
         public MSBuildPanel MSBuildPanel { get; private set; }
 
         public bool SkipBuildBeforePlaying { get; set; } = false;
+
+        public DateTime LastValidBuildDateTime { get; private set; }
 
         [UsedImplicitly]
         private bool CreateProjectSolutionIfNeeded()
@@ -437,6 +441,21 @@ namespace GodotTools
             }
         }
 
+        private void UpdateLastValidBuildDateTime()
+        {
+            var dllName = $"{GodotSharpDirs.ProjectAssemblyName}.dll";
+            var path = Path.Combine(GodotSharpDirs.ProjectBaseOutputPath, "Debug", dllName);
+            LastValidBuildDateTime = File.GetLastWriteTime(path);
+        }
+
+        private void BuildFinished(BuildResult buildResult)
+        {
+            if (buildResult == BuildResult.Success)
+            {
+                UpdateLastValidBuildDateTime();
+            }
+        }
+
         private void BuildStateChanged()
         {
             if (_bottomPanelBtn != null)
@@ -446,6 +465,8 @@ namespace GodotTools
         public override void _EnablePlugin()
         {
             base._EnablePlugin();
+
+            UpdateLastValidBuildDateTime();
 
             ProjectSettings.SettingsChanged += GodotSharpDirs.DetermineProjectLocation;
 
@@ -615,6 +636,12 @@ namespace GodotTools
             AddExportPlugin(exportPlugin);
             _exportPluginWeak = WeakRef(exportPlugin);
 
+            // Inspector plugin
+            var inspectorPlugin = new InspectorPlugin();
+            AddInspectorPlugin(inspectorPlugin);
+            _inspectorPluginWeak = WeakRef(inspectorPlugin);
+            BuildManager.BuildFinished += BuildFinished;
+
             BuildManager.Initialize();
             RiderPathManager.Initialize();
 
@@ -627,6 +654,10 @@ namespace GodotTools
             base._DisablePlugin();
 
             _editorSettings.SettingsChanged -= OnSettingsChanged;
+
+            // Custom signals aren't automatically disconnected currently.
+            MSBuildPanel.BuildStateChanged -= BuildStateChanged;
+            BuildManager.BuildFinished -= BuildFinished;
         }
 
         public override void _ExitTree()
@@ -659,6 +690,13 @@ namespace GodotTools
                     (_exportPluginWeak.GetRef().AsGodotObject() as ExportPlugin)?.Dispose();
 
                     _exportPluginWeak.Dispose();
+                }
+
+                if (IsInstanceValid(_inspectorPluginWeak))
+                {
+                    (_inspectorPluginWeak.GetRef().AsGodotObject() as InspectorPlugin)?.Dispose();
+
+                    _inspectorPluginWeak.Dispose();
                 }
 
                 GodotIdeManager?.Dispose();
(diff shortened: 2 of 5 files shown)


modules/mono/editor/GodotTools/GodotTools/GodotSharpEditor.cs
```diff
@@ -437,6 +441,21 @@ private void ApplyNecessaryChangesToSolution()
             }
         }
 
+        private void UpdateLastValidBuildDateTime()
+        {
+            var dllName = $"{GodotSharpDirs.ProjectAssemblyName}.dll";
+            var path = GodotSharpDirs.ProjectBaseOutputPath.PathJoin("Debug").PathJoin(dllName);
```
raulsntos, line 447: Since this is a normal path, we should probably use BCL methods to join them.

```suggestion
            string dllName = $"{GodotSharpDirs.ProjectAssemblyName}.dll";
            string path = Path.Combine(GodotSharpDirs.ProjectBaseOutputPath, "Debug", dllName);
```

modules/mono/editor/GodotTools/GodotTools/Inspector/InspectorPlugin.cs
```diff
@@ -0,0 +1,45 @@
...
+                {
+                    AddCustomControl(new InspectorOutOfSyncWarning());
+                    break;
+                }
+            }
+        }
+
+        private static IEnumerable<Script> EnumerateScripts(GodotObject godotObject)
+        {
+            var script = godotObject.GetScript().AsGodotObject() as Script;
```
raulsntos, line 37: I imagine the result of `GetScript` will always be a `Script`. I think it only returns `Variant` because it can be null, but in that case C# won't throw an exception.

```suggestion
            var script = godotObject.GetScript().As<Script>();
```

modules/mono/.editorconfig
```diff
@@ -14,6 +14,8 @@ max_line_length = 120
 csharp_indent_case_contents_when_block = false
 
 [*.cs]
+# IDE0005: Remove unnecessary using directives
+dotnet_diagnostic.IDE0005.severity = suggestion
```
raulsntos, line 18: This change seems unrelated.

raulsntos (commented):
Thank you so much for working on this.

modules/mono/editor/GodotTools/GodotTools/GodotSharpEditor.cs
```diff
@@ -661,7 +688,18 @@ protected override void Dispose(bool disposing)
...
+                if (IsInstanceValid(_inspectorPluginWeak))
+                {
+                    (_inspectorPluginWeak.GetRef().AsGodotObject() as InspectorPlugin)?.Dispose();
+
+                    _inspectorPluginWeak.Dispose();
+                }
+
                 GodotIdeManager?.Dispose();
+
+                // Custom signals aren't automatically currently.
```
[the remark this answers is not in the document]
paulloz: It's probably not a big deal, no. Now that I think about it, if we disconnect, it'd make more sense in `_DisablePlugin` since they're connected in `_EnablePlugin`.

modules/mono/.editorconfig
paulloz: Oops, didn't see I staged this in the commit. I'll remove it.

modules/mono/editor/GodotTools/GodotTools/GodotSharpEditor.cs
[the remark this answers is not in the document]
raulsntos: That makes sense to me. Although in practice the plugin is never disabled.

raulsntos:
> Sadly, the inspector is not refreshed ATM when the window is refocused

That's really unfortunate. When modifying a GDScript in an external editor the inspector updates even without refocusing the Godot editor, but I couldn't find any calls to `InspectorDock::get_inspector_singleton()->update_tree();` in the GDScript module.

paulloz:
> When modifying a GDScript in an external editor the inspector updates when refocusing the Godot editor

Uuh, you're right. I'm positive I tried it out and didn't see it refresh 🤔 Looking at it now, it comes from `GDScript::update_exports()` called by `ScriptEditor::_update_modified_scripts_for_external_editor()`.

raulsntos:
It'd be nice to notify only if the file last modified date is newer than the last valid build time, so maybe something like this:

```diff
diff --git a/modules/mono/csharp_script.cpp b/modules/mono/csharp_script.cpp
index 8e1587997b..9759db5a0a 100644
--- a/modules/mono/csharp_script.cpp
+++ b/modules/mono/csharp_script.cpp
@@ -2246,6 +2246,18 @@ bool CSharpScript::_update_exports(PlaceHolderScriptInstance *p_instance_to_upda
 			} else {
 				p_instance_to_update->update(propnames, values);
 			}
+		} else if (placeholders.size()) {
+			uint64_t script_modified_time = FileAccess::get_modified_time(get_path());
+			uint64_t last_valid_build_time = FileAccess::get_modified_time(GodotSharpDirs::get_res_temp_assemblies_dir().path_join(path::get_csharp_project_name() + ".dll"));
+
+			if (script_modified_time > last_valid_build_time) {
+				for (PlaceHolderScriptInstance *instance : placeholders) {
+					Object *owner = instance->get_owner();
+					if (owner->get_script_instance() == instance) {
+						owner->notify_property_list_changed();
+					}
+				}
+			}
 		}
 	}
 #endif

```

I'm not sure if `notify_property_list_change()` is really the right thing to use here though. Maybe instead of iterating the placeholders, we could just call `InspectorDock::get_inspector_singleton()->update_tree();` like the editor does when a GDExtension is reloaded:

https://github.com/godotengine/godot/blob/2d0ee20ff30461b6b10f6fdfba87511a0ebc6642/editor/editor_node.cpp#L443-L446

We also do that when reloading C# assemblies:

https://github.com/godotengine/godot/blob/2d0ee20ff30461b6b10f6fdfba87511a0ebc6642/modules/mono/csharp_script.cpp#L1144-L1148

paulloz:
Yeah, I was pondering the call to `update_tree()` directly because I don't really know if there's a risk we end up calling that a bunch? Vs. notifying the change where it only dirties a flag, and the inspector will update only once.

YuriSizov:
As a rule of thumb you should never call `update_tree` manually. It should be done as little as possible.

paulloz:
> It's probably triggered somewhere in `GDScript::_update_exports`, maybe we could do the same in `CSharpScript::_update_exports`.

Humpf, actually the issue is this is only triggered when `text_editor/external/use_external_editor` is set, but we use `dotnet/editor/external_editor` for C#.

paulloz:
Yeah, not sure what the best way to handle this is. Should we have our own version of what's done in `ScriptEditor::_update_modified_scripts_for_external_editor` to update even when this setting is unset?

raulsntos:
For me I have both settings unset and it seems to work, I guess it's because I have the C# script open in the Godot editor. But you are right that if I have `text_editor/external/use_external_editor` unset and `dotnet/editor/external_editor` set it won't update the inspector on refocusing the Godot window, every other combination seems to work.

> Should we have our own version of what's done in `ScriptEditor::_update_modified_scripts_for_external_editor` to update even when this setting is unset?

Maybe we can just change `ScriptEditor::_update_modified_scripts_for_external_editor` to also consider `scr->get_language()->overrides_external_editor()` (like `ScriptEditor::edit` does).

modules/mono/csharp_script.cpp
```diff
@@ -2246,6 +2246,17 @@ bool CSharpScript::_update_exports(PlaceHolderScriptInstance *p_instance_to_upda
 			} else {
 				p_instance_to_update->update(propnames, values);
 			}
+		} else if (placeholders.size()) {
+			uint64_t script_modified_time = FileAccess::get_modified_time(get_path());
+			uint64_t last_valid_build_time = CSharpLanguage::get_singleton()->gdmono->get_project_assembly_modified_time();
```
raulsntos, line 2251: You could use `GDMono::get_singleton` to obtain it directly.

```suggestion
			uint64_t last_valid_build_time = GDMono::get_singleton()->get_project_assembly_modified_time();
```

raulsntos (approved):
Looks good to me and works as expected. Thanks again for contributing to the usability of the .NET module.

paulloz:
Last push is simply squashing all commits, and using `TTR()` on the string used in UI.

raulsntos:
> using `TTR()` on the string used in UI.

That shouldn't be necessary, `Label` auto translates the `Text` property.

https://github.com/godotengine/godot/blob/13a0d6e9b253654f5cc2a44f3d0b3cae10440443/scene/gui/label.cpp#L889

paulloz:
Urh, sorry, I thought it was necessary for the key to be extracted in the first place.

raulsntos:
I don't think extracting works for C# either way, so all the other strings that we have with `TTR()` are probably not translated either. For example, I couldn't find the string `Generating solution...` in [Weblate](https://hosted.weblate.org/projects/godot-engine/godot/en/?q=Generating%20solution...#search).

Maybe these strings need to be added to https://github.com/godotengine/godot-editor-l10n manually or implement C# support in [extract_editor.py](https://github.com/godotengine/godot-editor-l10n/blob/main/scripts/extract_editor.py).

paulloz:
So I guess having the `TTR()` call here is kinda future-proofing for when we add C# support to the extraction script 😅

akien-mga:
Thanks!

molingyu:
Can an editor option be provided so that the project can automatically perform C# compilation after the file is changed?

paulloz:
Hello! There are already other issues discussing this
- godotengine/godot-proposals#849
- godotengine/godot-proposals#2231
- godotengine/godot-proposals#2920
- #93431

PR #85872, opened 2023-12-06
Fix `mouse_in_window` state with confined mouse

Sauermann:
`_update_mouse_over` is called whenever the mouse is over the Window. So `mouse_in_window` needs to be set in all cases.

resolve #85713
regression from #80334


akien-mga (dismissed):

Sauermann:
I put the PR into draft status, because during my tests I found a bug. A more elaborate solution would be necessary.

PR #85877, opened 2023-12-07
Fix missing heading in translated online class reference

timothyqiu:
The RST format has a quirk that requires the section header's marker to be at least the same _visual_ length as the header's text. We can't use the text's length to generate markers because, for example, east Asian wide and fullwidth characters take twice the visual length.

Currently, the "Resources" section is missing in the index of zh_CN class reference:

ksnip_20231207-111049

Because the name becomes two characters after translation, but requires four `=`s.

```
资源
====
```

We made a dedicated `make_heading()` function that simply doubles the marker's length to solve this problem :stuck_out_tongue_closed_eyes:

### DIFF
--- a/doc/tools/make_rst.py
+++ b/doc/tools/make_rst.py
@@ -1552,16 +1552,11 @@ def make_rst_index(grouped_classes: Dict[str, List[str]], dry_run: bool, output_
 
     f.write(".. _doc_class_reference:\n\n")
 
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.1.

PR #42035, opened 2020-09-13
Fix audio bus effects not deselecting after switching focus in the editor

Vinealius:
Make the appropriate selection of the effect on the audio bus.

> - Select an effect on more than one Audio bus the other ones remains selected, so if you press the Delete key you can delete one effect selected at a time of each diferent Audio bus.

> - Create a new node in the scene and create and select a Audio bus effect, then select the new scene node and press the Delete key to delete the node, the effect will be deleted instead.

- Fixes godotengine/godot#11355

PR #85879, opened 2023-12-07
Fix permanently selected audio bus effects

### LINKED ISSUE #85478: Unable to deselect Audio bus effects (including selecting nodes or files, switching tabs, etc.)
### Godot version

3.5.3.stable.official [6c814135b]

### System information

OS: Windows 11 - CPU: Intel Core i7-8700 @ 3.20GHz - GPU: NVIDIA GeForce GTX 1060 3GB

### Issue description

When I create an effect on an Audio Bus, selecting it is permanent unless I select another effect.
I've had this issue for a while and, when first discovering thought to myself "Eh, somebody else probably found it, it's already fixed". That was about a year ago now (_maybe longer_), and it's still the exact same.

The way I'd expect it to work is it would deselect if I clicked **anywhere** outside of the effect itself or the Inspector window (_or if I switch to the Output, Debugger, etc._).

### Demonstration video
https://github.com/godotengine/godot/assets/42562727/42ee9699-0420-4b97-9644-8de62204c78e

### Steps to reproduce

1. Open a Godot project (_happens in anyone for me_) or create a new one
2. Open the Audio tab
3. Create an effect and select it or select an existing one
4. Attempting to click outside the box will never deselect it. Attempting to delete a node or clicking DEL will only delete the effect (_even when it's not visible_).
### Temporary Fix
Create a deselection effect on the same bus and delete it. While it is an easy fix, this can get in the way of many developers, as it did with me (_it took me months to realize this_).

N/A

berrybus:
This allows the audio bus effects to be deselected if the focus changes to fix [this bug](https://github.com/godotengine/godot/issues/85478). Here is a recording of the new behavior:

https://github.com/godotengine/godot/assets/10054174/54ced10a-6fb5-45a4-b9fd-d284821f90c6

This is my first PR, and I wasn't sure if this was the best way to do it. I'm open to any feedback. Thanks for taking a look!

*Bugsquad edit:*
- Fixes #85478

### DIFF
--- a/editor/editor_audio_buses.cpp
+++ b/editor/editor_audio_buses.cpp
@@ -912,6 +912,7 @@ EditorAudioBus::EditorAudioBus(EditorAudioBuses *p_buses, bool p_is_master) {
 	vb->add_child(effects);
 	effects->connect("item_edited", callable_mp(this, &EditorAudioBus::_effect_edited));
 	effects->connect("cell_selected", callable_mp(this, &EditorAudioBus::_effect_selected));
+	effects->connect("focus_exited", callable_mp(effects, &Tree::deselect_all));
 	effects->set_edit_checkbox_cell_only_when_checkbox_is_pressed(true);
 	SET_DRAG_FORWARDING_GCD(effects, EditorAudioBus);
 	effects->connect("item_mouse_selected", callable_mp(this, &EditorAudioBus::_effect_rmb));


editor/editor_audio_buses.cpp
```diff
@@ -500,6 +500,10 @@ void EditorAudioBus::_effect_selected() {
 	updating_bus = false;
 }
 
+void EditorAudioBus::_effects_deselected() {
+	effects->deselect_all();
```
berrybus, line 504: This works for some cases when focus is lost, but it's kind of annoying when your mouse is just hovering over another node like the scene editor. Is there a way to check "is left mouse button down" at this part?

KoBeWi: No. Other docks don't deselect on focus lost (e.g. file dock), rather they do it when you click away. Though then there is inspector, which doesn't allow deselecting properties 🤷‍♂️

You could use "empty_clicked" signal for deselecting. This means you'd need to click the empty space under effects, but they'd be deselectable in a predictable way at least. The empty space in bus layout (between buses) could be used too.

editor/editor_audio_buses.h
```diff
@@ -108,6 +108,7 @@ class EditorAudioBus : public PanelContainer {
 	void _effect_edited();
 	void _effect_add(int p_which);
 	void _effect_selected();
+	void _effects_deselected();
```
KoBeWi, line 111: It's a callback for focus lost, so it needs a more accurate name.

KoBeWi (approved):
The solution could use some tweaks, as discussed in the comments, but it's already acceptable I think.

editor/editor_audio_buses.cpp
berrybus: Oh okay, the empty click event makes more sense. I'll try that, thanks! I couldn't find a generic one for "clicked somewhere else /outside/ of the current panel," so I guess that doesn't exist?

KoBeWi: Yep, clicking another control at most can steal focus.

editor/editor_audio_buses.h
berrybus: I tried `effects->connect("empty_clicked", callable_mp(effects, &Tree:deselect_all)`, and got an error `Method expected 0 arguments, but called with 2`. I'm very rusty on cpp, so I don't know if there's a good way to make like... a closure or something to take in more arguments, so I'll just go ahead and rename 😅

akien-mga:
Is this equivalent to https://github.com/godotengine/godot/pull/42035?

KoBeWi:
Yes, although this solution is better.

berrybus:
I was planning to use `empty_clicked` on the effects `Tree` and some "click empty space" method on the audio buses pane, but I could not find an easy way to check an "click empty space" event on the `BoxContainer` or `ScrollContainer` of the audio buses , so I've just gone ahead and renamed the focus_exited function.

And ah, yes it looks like someone else already made this change 3 years ago haha. I suppose y'all can decide which PR to merge!

editor/editor_audio_buses.h
AThousandShips: Did you specifically type that? If so you missed a colon: `effects->connect("empty_clicked", callable_mp(effects, &Tree::deselect_all));`

berrybus: I had the two colons in the actual code, I just retyped it out like an idiot.

But also my bad, `focus_exited` takes no arguments, I'll remove the extra function 🤦

AThousandShips: You can do `effects->connect("empty_clicked", callable_mp(effects, &Tree::deselect_all).unbind(2));` if you want to use this signal

AThousandShips (approved):

editor/editor_audio_buses.h
berrybus: Ohh, got it thanks!

KoBeWi:
> I could not find an easy way to check an "click empty space" event on the BoxContainer or ScrollContainer of the audio buses

For stuff like generic containers you could use `gui_input` signal and check for mouse click, but it's not worth it.

YuriSizov:

YuriSizov:
Cherry-picked for 4.2.2.

PR #85887, opened 2023-12-07
Correctly enforce minimum window size in editor

YuriSizov:
Calling `Window`'s size methods this early doesn't sync the values with the DisplayServer because the main window doesn't have a valid ID yet. So we have to do the sync manually, otherwise the minimum size is not actually enforced.

As an extra usability improvement, and the reason why I noticed this issue, I added the upper cap on the language picker popup in the project manager. By default it just takes as much space as possible, which doesn't seem very usable to me. I made it to cap the size at half the screen resolution, but no less than the minimum window size. The result:

image

----

I refactored the PM's code related to this a bit and removed some bits which I think we no longer need.

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -6852,7 +6852,9 @@ EditorNode::EditorNode() {
 	// Define a minimum window size to prevent UI elements from overlapping or being cut off.
 	Window *w = Object::cast_to<Window>(SceneTree::get_singleton()->get_root());
 	if (w) {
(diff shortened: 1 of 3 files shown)


YuriSizov:
Also, this PR makes the project manager a bit bigger by default, matching the minimum size of the editor (so it should not make things worse for smaller screens). But the minimum allowed size of the project manager is still smaller.

bruvzg (approved):

YuriSizov:
Thanks for reviews!

YuriSizov:
Cherry-picked for 4.2.2.

PR #85899, opened 2023-12-07
Add summary on project exit when using `--print-fps`

Calinou:
- Related to https://github.com/godotengine/godot-proposals/issues/5459.

When combined with the `--quit-after` and `--disable-vsync` command line arguments, this can be used to reliably benchmark pull requests without having to add bespoke code to every project being tested.

Metrics displayed were inspired by [CapFrameX](https://www.capframex.com/blog/post/The%20challenge%20of%20displaying%20performance%20metrics%20as%20FPS).

Example output:

image

    Project FPS: 455 (2.19 mspf)
    Project FPS: 340 (2.94 mspf)
    Project FPS: 326 (3.06 mspf)
    [...]

    3000 frames rendered in 5.53 seconds.
    Frametime statistics over the duration of the run:

    Average: 553 FPS (1.81 mspf)

    Maximum: 1506 FPS (0.66 mspf)
        P99: 1466 FPS (0.68 mspf)
     Median: 783 FPS (1.28 mspf)
         P1: 214 FPS (4.67 mspf)
       P0.1: 161 FPS (6.20 mspf)
    Minimum: 92 FPS (10.84 mspf)

     Frametimes compliant with a 30 FPS cap: 100.0%
     Frametimes compliant with a 60 FPS cap: 100.0%
    Frametimes compliant with a 120 FPS cap: 99.9%

### DIFF
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -4295,6 +4295,17 @@ int Main::start() {
 uint64_t Main::last_ticks = 0;
 uint32_t Main::frames = 0;
 uint32_t Main::hide_print_fps_attempts = 3;
+
+// The first rendered frames' frametimes are significantly inflated regardless
+// of V-Sync mode, so discard them in the FPS summary (but count them in the total number
+// of rendered frames).
+const int PRINT_FPS_DISCARDED_FRAMES = 10;
+
+Vector<uint32_t> print_fps_frametimes;
+// Used to determine the duration over which frames were rendered.
+uint32_t print_fps_first_frame_ticks = 0;
+uint32_t print_fps_last_frame_ticks = 0;
+
 uint32_t Main::frame = 0;
 bool Main::force_redraw_requested = false;
 int Main::iterating = 0;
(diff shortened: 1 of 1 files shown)


main/main.cpp
```diff
@@ -3660,6 +3671,15 @@ bool Main::iteration() {
...
 	uint64_t frame_time = OS::get_singleton()->get_ticks_usec() - ticks;
 
+	if (print_fps_first_frame_ticks == 0) {
+		print_fps_first_frame_ticks = Time::get_singleton()->get_ticks_usec();
+	}
+	print_fps_last_frame_ticks = Time::get_singleton()->get_ticks_usec();
+
+	if (print_fps && Engine::get_singleton()->frames_drawn > PRINT_FPS_DISCARDED_FRAMES) {
+		print_fps_frametimes.push_back(frame_time);
+	}
```
akien-mga, line 3681: All this should be if `if (print_fps)` so that it doesn't impact iteration performance when not printing fps.

Calinou:
Rebased and tested again, it works as expected.

Calinou:
> Hopefully not just feature creep but IMO it would be useful if the tracking + reporting portion of this was usable on its own for a running project so that multiple runs could be automated with various configurations without needing to restart the project.

This would make the code a lot more complex, so it would be better done in its own PR.

Ansraer:
I can see the usecase for this, but imo this also needs to output the data in a machine readable format to be really useful.
Right now the user needs to manually extract the values from the console, which is arguably worse than something custom that does write the results to a file.
Just creating a csv file (with column headers!!!) On exit next to the binary should do the trick.

PR #85900, opened 2023-12-07
Localization: Fix pt_BR strings for Debug/Release User and Password

ricardobusta:
The old strings contained
"To Debug User/Password"
and
"Free the User/Password"

image

Replaced them with the strings with the correct meaning in Portuguese from Brazil, based in the regular "Debug" and "Release" strings, so they all match.


AThousandShips:
This is not where these files are edited, instead you edit them on the webplate

Issue #77513, opened 2023-05-26
`GDScript` resources are not unloaded due to an extra reference


dalexeev:
### Godot version

v4.0.3.stable.official [5222a99f5], master

### System information

Kubuntu 23.04

### Issue description

It looks like we have an extra reference to the script somewhere, which prevents it from being freed at the right time. 

See also [the comment](https://github.com/godotengine/godot/pull/77129#discussion_r1205894765).

CC @adamscott

### Steps to reproduce

```gdscript
extends Node

const RES_PATH = "res://a.gd"
#const RES_PATH = "res://icon.svg"

var i := 0
var wr: WeakRef

func _on_timer_timeout() -> void:
    print("===")
    if i % 2:
        print("wr = ", wr.get_ref())
    else:
        var res = load(RES_PATH)
        wr = weakref(res)
        print("ref count = ", res.get_reference_count())
        print("wr = ", wr.get_ref())
    i += 1
```

```
===
ref count = 2
wr = <GDScript#-9223372008484502375>
===
wr = <GDScript#-9223372008484502375>
===
ref count = 2
wr = <GDScript#-9223372008484502375>
===
wr = <GDScript#-9223372008484502375>
```

```
===
ref count = 1
wr = <CompressedTexture2D#-9223372010162223986>
===
wr = <null>
===
ref count = 1
wr = <CompressedTexture2D#-9223372009843456882>
===
wr = <null>
```

### Minimal reproduction project

[bug-script-ref-count.zip](https://github.com/godotengine/godot/files/11575495/bug-script-ref-count.zip)

adamscott:
> It looks like we have an extra reference to the script somewhere, which prevents it from being freed at the right time.

It's by design to enable cyclic references. Each GDScript loaded has a Ref instance living here:
https://github.com/godotengine/godot/blob/e188d619227990001667821dac8bc8940076d4a9/modules/gdscript/gdscript_cache.h#L79-L81

But the garbage collection part is not yet implemented, unfortunately.

dalexeev:
Does this mean that `@unload_static` doesn't really matter now: scripts aren't unloaded anyway?

vnen:
> It's by design to enable cyclic references. Each GDScript loaded has a Ref instance living [in the cache]

[...]

PR #85904, opened 2023-12-07
Fix for PCK file replacement being thwarted by caching

ogapo:
PCK loading supports a replacement feature such that mounting a PCK which contains a file (`foo/bar.gd` for instance) should replace the corresponding `foo/bar.gd` that shipped with the game (superseding any prior mounted PCKs, including the main PCK). This works great if you mount everything at startup, but for downloaded PCK files (DLC) they are sometimes mounted after the underlying asset has loaded. If this asset is a resource then the load-once policy makes it so that, although the new file is mounted and available, it cannot be used by gameplay without explicitly calling load (and passing the cache ignore flags). This is not really how most loading is done, and doesn't work well for indirect loads (often the case when you are picking up a new version of an asset from the DLC).

Initially I identified that the ResourceCache needed to be cleared, however clearing the whole cache when mounting any new PCK is inefficient and some content models (with many small DLC) would not want to do that. Instead, a better solve is simply when mounting a PCK file we should evict any files in the PCK from any caches. If the files are new then eviction has no effect, and if not, the eviction allows the new version of the file to be loaded as expected the next time the resource is requested.

While implementing this, I noticed there's a second-order cache for GDScripts specifically (`GDScriptCache`) and since this exists in a module it's not possible to purge it as a result of mounting a PCK (which occurs in core). Rather than trying to manually enumerate all cases (and to account for the fact that core can't know about them all anyway) I added a callback mechanism so that systems can self-register for notification of evictions and handle second-order caches accordingly. 

I wasn't sure if there's a more "godot" way to do delegates/lambdas so I just kept it simple and used a function+context pointer pair. Open to PR feedback on this area in particular :-).

It's worth noting that this isn't expected to change any instances of the old asset presently loaded (RefCounted things would still hang out until their references are all gone) but it should allow the new version of the file to be loaded after mounting. My particular use case uses a scene change to get rid of old references so I'm not sure if there's any potential gotchas in the case where multiple versions of a resource exist, but if there are then these would exist in any system that allows both immutable references and hot reloading so I don't think they'd be particularly surprising to anyone trying to dynamically mount PCKs already.


core/io/resource.h
```diff
@@ -149,11 +149,18 @@ class Resource : public RefCounted {
...
+	void (*listener)(void *p_context, const String &p_path);
+	void *context;
+};
+
 class ResourceCache {
 	friend class Resource;
 	friend class ResourceLoader; //need the lock
 	static Mutex lock;
 	static HashMap<String, Resource *> resources;
+	static Mutex listener_lock;
```
AThousandShips, line 162:
```suggestion
	static Mutex listener_mutex;
```
Better naming for `Mutex` IMO, calling a `Mutex` a lock is not entirely correct (and I'd sugest renaming the mutex above)

core/io/resource.cpp
```diff
@@ -522,6 +525,22 @@ bool ResourceCache::has(const String &p_path) {
...
+	lock.lock();
+	bool was_present = resources.erase(p_path);
+	lock.unlock();
+
+	if (was_present) {
+		listener_lock.lock();
+		for (const EvictionListenRecord &rec : eviction_listeners) {
+			rec.listener(rec.context, p_path);
+		}
+		listener_lock.unlock();
```
AThousandShips, line 538:
```suggestion
		MutexLock locked(listener_mutex);
		for (const EvictionListenRecord &rec : eviction_listeners) {
			rec.listener(rec.context, p_path);
		}
```

core/io/resource.cpp
```diff
@@ -576,3 +595,30 @@ int ResourceCache::get_cached_resource_count() {
...
+	ResourceCache::unlisten_for_eviction(p_context);
+
+	listener_lock.lock();
+
+	EvictionListenRecord rec;
+	rec.context = p_context;
+	rec.listener = p_listener;
+	eviction_listeners.push_back(rec);
+
+	listener_lock.unlock();
```
AThousandShips, line 609:
```suggestion
	MutexLock locked(listener_mutex);

	EvictionListenRecord rec;
	rec.context = p_context;
	rec.listener = p_listener;
	eviction_listeners.push_back(rec);
```

core/io/resource.cpp
```diff
@@ -576,3 +595,30 @@ int ResourceCache::get_cached_resource_count() {
...
+
+	for (int i = 0; i < eviction_listeners.size();) {
+		if (eviction_listeners[i].context == p_context) {
+			eviction_listeners.remove_at(i);
+			continue;
+		}
+		++i;
+	}
+
+	listener_lock.unlock();
```
AThousandShips, line 623:
```suggestion
	MutexLock locked(listener_mutex);

	for (int i = 0; i < eviction_listeners.size();) {
		if (eviction_listeners[i].context == p_context) {
			eviction_listeners.remove_at(i);
			continue;
		}
		++i;
	}
```

core/io/resource.cpp
```diff
@@ -522,6 +525,22 @@ bool ResourceCache::has(const String &p_path) {
 	return true;
 }
 
+bool ResourceCache::evict(const String &p_path) {
+	lock.lock();
+	bool was_present = resources.erase(p_path);
+	lock.unlock();
```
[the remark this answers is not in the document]
ogapo: Good callout. I was a bit reluctant to not just follow the local pattern but as you point out it looks like MutexLock is the newer style. Adjusted to use that (and renamed `lock` into `resources_lock` and applied the same refactor)

core/io/resource.cpp
```diff
@@ -56,32 +56,32 @@ void Resource::set_path(const String &p_path, bool p_take_over) {
 		p_take_over = false; // Can't take over an empty path
 	}
 
-	ResourceCache::lock.lock();
+	{
+		MutexLock locked(ResourceCache::resources_mutex);
```
AThousandShips, line 60:
```suggestion
		MutexLock lock(ResourceCache::resources_mutex);
```
Same for the rest, looks better (used `locked` because `lock` was taken, but now not needed)

core/io/resource_loader.cpp
```diff
@@ -981,17 +981,17 @@ String ResourceLoader::path_remap(const String &p_path) {
...
-
 	List<Resource *> to_reload;
-	SelfList<Resource> *E = remapped_list.first();
 
-	while (E) {
-		to_reload.push_back(E->self());
-		E = E->next();
-	}
+	{
+		MutexLock locked(ResourceCache::resources_mutex);
```
ogapo, line 987: @AThousandShips I've preserved the previous behavior here (only using MutexLock instead of lock/unlock), but I don't immediately see why this code needs to lock the resources_mutex at all. `remapped_list` is not under that guard anywhere else. ResourceCache explicitly friend's ResourceLoader so it can have access to internals. Wondering if this is vestigial cruft that ought to be removed?

ogapo:
@AThousandShips I've applied the rename you suggested and went ahead and squashed the commits in prep for merging if you approve.

AThousandShips:
Can't confirm the solution itself, can test later but best if someone more knowledgeable on the area of GDScript can review, and then I can approve the threading safety aspects of it (which look good to me)

adamscott:
Discussed during the GDScript Team meeting. I, for one, find this PR nicely done, at least for the GDScript part. 

But others asked if this was not a workaround for the actual problem of scripts being kept in memory when there's a cyclic reference. See https://github.com/godotengine/godot/issues/77513.

Will have to investigate.

ogapo:
It seems that the need (or not) for a cache of the compiled gdscript files is kinda separate from the idea of supporting secondary caches at all. If we accept that there are any good cases for post load processing and caching of resources, then allowing them to be aware of resource eviction/replacement is pretty much required for correctness.

Ideally whatever is happening in the script cache could be done at import time, but I'm sure that's easier said than done. The fact that the current implementation of the script cache needs the eviction callback feels ok to me, even though it might be refactored to no longer need it in the future.

adamscott:
[As asked by @reduz in the developers' chat](https://chat.godotengine.org/channel/core?msg=6ZQ5oA4t4ZNnHaQQM), what happens if the eviction is done during the loading of the same file?

Technically, when the loading is done, the cache of that file will be set by ResourceLoader.

ogapo:
Is there a mechanism I'm not aware of that could trigger a PCK to mount midway through loading an existing resource? I am only aware of ways to initiate that manually. That's the only location that calls ResourceCache::evict right now.

ogapo:
@adamscott @reduz @AThousandShips  Any thoughts on this PR? Guessing it's unlikely to make 4.5 at this point :-). I'm happy to explore other ways of accomplishing the same thing if we don't like this approach.

The issue at hand is there's currently no way to purge any caches when a new file is mounted via a PCK (which supports replacing files). This is not ideal but mostly works if all PCKs are mounted at startup. However, as soon as you start trying to do any dynamic mounting, the lack of cache control becomes problematic. Individual asset load requests can already ask to bypass the cache with a flag, but there's no way to purge the cache for the next request which is what we need in the case of a PCK file replacement. 

PCK doesn't have any concept of unload, so the best we can do is mount and replace. I actually like this approach, but it simply doesn't work unless I can guarantee that the caches are purged :-\

HolonProduction:
I don't see how purging the `ResourceCache` would solve any issues. `ResourceCache` does not own anything. The lifetime of resources isn't tied to it. If a resource is not referenced anymore it will be dropped from the cache:

```gdscript
func _run() -> void:
	print(ResourceLoader.has_cached("res://new_resource.tres"))
	var a = ResourceLoader.load("res://new_resource.tres")
	print(ResourceLoader.has_cached("res://new_resource.tres"))
	a = null
	print(ResourceLoader.has_cached("res://new_resource.tres"))
```
Prints
```
false
true
false
```


If an old resource is cached it is referenced somewhere else, and simply purging it from the cache won't fix the outdated referenced instance.

This really seems like an over-complicated workaround for the fact, that GDScripts stay referenced until engine shutdown (https://github.com/godotengine/godot/issues/77513).

ogapo:
> I don't see how purging the `ResourceCache` would solve any issues. `ResourceCache` does not own anything. The lifetime of resources isn't tied to it. If a resource is not referenced anymore it will be dropped from the cache:

That's true but if the resource is still being referenced and a PCK is mounted with a newer version, the cache causes all new content to incorrectly reference the old version. This isn't about somehow purging existing references. The cache simply is stale but isn't being purged and not doing so makes it impossible to reference the newly mounted version of the file path or UID.

PR #85914, opened 2023-12-08
CanvasItem: Expose get_effective_z_index to GDScript.

20kdc:
Recently had a fellow developer run into a need for this function (as it is, it was reimplemented in GDScript) due to picking logic being rather unusual (see attached test project). While it would be nice for this logic to be more consistent, being able to sort items properly is important anyway.

In addition this sort of thing seems to be wanted generally, i.e. https://forum.godotengine.org/t/getting-the-absolute-z-index-of-a-node/23709/2

With this in mind, I decided it would be good to simply expose this function and point out what it doesn't handle in the documentation. This allows GDScript-side code to collate and sort the objects. Again, even if picking behaviour changes to be more consistent, this should still have usecases.

Attached test project shows how the function works. It also shows how picking logic appears to act, in case that helps explain the rationale.

[godot-dev-test-effectivez.zip](https://github.com/godotengine/godot/files/13613488/godot-dev-test-effectivez.zip)

An alternative might be a "does this CanvasItem draw above this other CanvasItem" function. While this would be useful and more versatile, to get it to really work *right* would be difficult, particularly given Y-sorting became integrated into RenderingServer. It's the sort of thing that would probably need some kind of unit test that could introspect the RenderingServer's draw order to compare.

### DIFF
--- a/doc/classes/CanvasItem.xml
+++ b/doc/classes/CanvasItem.xml
@@ -417,6 +417,13 @@
 				Returns the transform from the coordinate system of the canvas, this item is in, to the [Viewport]s coordinate system.
 			</description>
 		</method>
+		<method name="get_effective_z_index" qualifiers="const">
+			<return type="int" />
+			<description>
+				Returns the total Z index, accounting for [member z_as_relative].
+				This does not account for other effects such as [member top_level], [member y_sort_enabled], or sorting by hierarchy (see [method Node.get_index]).
+			</description>
+		</method>
 		<method name="get_global_mouse_position" qualifiers="const">
 			<return type="Vector2" />
 			<description>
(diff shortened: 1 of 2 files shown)


doc/classes/CanvasItem.xml
```diff
@@ -417,6 +417,13 @@
 				Returns the transform from the coordinate system of the canvas, this item is in, to the [Viewport]s coordinate system.
 			</description>
 		</method>
+		<method name="get_effective_z_index" qualifiers="const">
+			<return type="int" />
+			<description>
+				Returns the total Z index, accounting for [member z_as_relative].
+				This does not account for other effects such as [member top_level], [member y_sort_enabled], or sorting by hierarchy (see [member get_index].)
```
AThousandShips, line 424:
```suggestion
				This does not account for other effects such as [member top_level], [member y_sort_enabled], or sorting by hierarchy (see [member get_index]).
```

20kdc: This was intended to refer to `Node.get\_index`, unsure how to fix that.

AThousandShips:
```suggestion
				This does not account for other effects such as [member top_level], [member y_sort_enabled], or sorting by hierarchy (see [method Node.get_index]).
```

20kdc: Tested and applied.

20kdc: Should this conversation be marked as resolved?

PR #85918, opened 2023-12-08
GDScript: Adjust `STATIC_CALLED_ON_INSTANCE` warning to not force native type

### LINKED ISSUE #69282: Calling static function from within a class generates a warning
### Godot version

4.0.beta6

### System information

Windows 10

### Issue description

Hi all,

Calling a static function from within the class that owns it generates a warning. I would expect that calling the function from within the class shouldn't generate the warning (particularly if that class doesn't have a class name).

Thanks!

### Steps to reproduce

Add the below to a script and note the Error (warning) it generates:

```
func _ready() -> void:
    a_static_func()

static func a_static_func() -> void:
    pass
```

N/A

### LINKED ISSUE #74397: Fix suggested by STATIC_CALLED_ON_INSTANCE warning is incorrect
### Godot version

### System information

Steam OS Holo, Steam Deck

### Issue description

The STATIC_CALLED_ON_INSTANCE warning includes a suggested fix, which only applies when the static function is defined in the base class.  Attempting to follow the suggestion blindly (which novice programmers may do) causes an error.

**Test.gd**:
```gdscript
extends Node

static func foo():
	pass

func _process(delta):
	foo()
```

The warning printed by Godot in the case above is as follows, even though `Node.foo()` doesn't exist:
```
W 0:00:00:0454   The function 'foo()' is a static function but was called from an instance. Instead, it should be directly called from the type: 'Node.foo()'.
  <GDScript Error>STATIC_CALLED_ON_INSTANCE
  <GDScript Source>Test.gd:7
```

It's not clear to me what the correct fix is in this case, because the script doesn't provide a class name.  But it's worth noting that the STATIC_CALLED_ON_INSTANCE warning is unchanged by defining a `class_name`.

I suspect this may be a case where it's hard to print an accurate and reliable suggestion.  Maybe it would be better to phrase it as an example, instead, and not try to detect the base class?
```
W 0:00:00:0454   The function 'foo()' is a static function but was called from an instance. Instead, it should be directly called from the type, for example: 'Type.foo()'.
  <GDScript Error>STATIC_CALLED_ON_INSTANCE
  <GDScript Source>Test.gd:7
```

### Steps to reproduce

N/A

N/A

20kdc:
Defaulting to the native type is less than useful, as:

* There are very few native types that are extensible and have static methods.
* Defaulting to the native type does not account for a method being script-defined.

While the "real fix" would be to carefully track the source of the method, the get_function_signature method is already complicated enough.

This will at least ensure the resulting code should always be valid.

This PR was written after seeing the following warning:

```
*  W 0:00:00:0542   The function "get_absolute_z_index()" is a static function but was called from an instance. Instead, it should be directly called from the type: "Node2D.get_absolute_z_index()".
```

Where `get_absolute_z_index()` was a static function declared in a script (and thus did not exist in Node2D, and thus the given code did not work)

* _Bugsquad edit:_ Fixes #69282.
* _Bugsquad edit:_ Fixes #74397.

### DIFF
--- a/modules/gdscript/tests/scripts/parser/warnings/static_called_on_instance.gd
+++ b/modules/gdscript/tests/scripts/parser/warnings/static_called_on_instance.gd
@@ -1,11 +1,23 @@
-class Player:
-	var x = 3
+class_name TestStaticCalledOnInstance
+
+class Inner:
+	static func static_func():
+		pass
+
+static func static_func():
+	pass
 
 func test():
-	# These should not emit a warning.
-	var _player = Player.new()
-	print(String.num_uint64(8589934592)) # 2 ^ 33
+	print(String.num_uint64(8589934592))
+	var some_string := String()
+	print(some_string.num_uint64(8589934592)) # Warning.
 
-	# This should emit a warning.
-	var some_string = String()
-	print(some_string.num_uint64(8589934592)) # 2 ^ 33
+	TestStaticCalledOnInstance.static_func()
+	static_func()
+	self.static_func()
+	var other := TestStaticCalledOnInstance.new()
+	other.static_func() # Warning.
+
+	Inner.static_func()
+	var inner := Inner.new()
+	inner.static_func() # Warning.
--- a/modules/gdscript/tests/scripts/parser/warnings/static_called_on_instance.out
+++ b/modules/gdscript/tests/scripts/parser/warnings/static_called_on_instance.out
@@ -1,7 +1,15 @@
 GDTEST_OK
 >> WARNING
->> Line: 11
+>> Line: 13
 >> STATIC_CALLED_ON_INSTANCE
 >> The function "num_uint64()" is a static function but was called from an instance. Instead, it should be directly called from the type: "String.num_uint64()".
+>> WARNING
+>> Line: 19
+>> STATIC_CALLED_ON_INSTANCE
+>> The function "static_func()" is a static function but was called from an instance. Instead, it should be directly called from the type: "TestStaticCalledOnInstance.static_func()".
+>> WARNING
+>> Line: 23
+>> STATIC_CALLED_ON_INSTANCE
+>> The function "static_func()" is a static function but was called from an instance. Instead, it should be directly called from the type: "Inner.static_func()".
 8589934592
 8589934592
--- a/modules/gdscript/gdscript_analyzer.cpp
+++ b/modules/gdscript/gdscript_analyzer.cpp
@@ -3360,12 +3360,8 @@ void GDScriptAnalyzer::reduce_call(GDScriptParser::CallNode *p_call, bool p_is_a
 			parser->push_warning(p_call, GDScriptWarning::RETURN_VALUE_DISCARDED, p_call->function_name);
 		}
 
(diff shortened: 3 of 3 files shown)


vnen (approved):

dalexeev:
Note that for non-global classes the warning will suggest invalid code.

20kdc:
> Note that for non-global classes the warning will suggest invalid code.

Example, please? I'm not sure what constitutes a "non-global class" in this case.

dalexeev:
> Example, please? I'm not sure what constitutes a "non-global class" in this case.

vnen:
Note that this would be wrong on the current master also, it would show `EditorScript.static_func()` which isn't a valid way to call it.

We can either accept this as small improvement or work on a proper fix that finds the origin of method.

PS: I wonder if we should even emit this warning with a direct call for a function in the same class. If it does not have `class_name` it is quite cumbersome to avoid this warning.

20kdc:
> 

I've amended my commit on the basis of that other languages don't warn for these kinds of self-calls to static functions.

There are more examples and counter-examples one could add involving `self` and `var x := self` and so forth.

modules/gdscript/tests/scripts/parser/warnings/static_called_on_instance.gd
dalexeev, line 1: <details>
<summary>Suggestion</summary>

```gdscript
class_name TestStaticCalledOnInstance

class Inner:
    static func static_func():
        pass

static func static_func():
    pass

func test():
    print(String.num_uint64(8589934592))
    var some_string := String()
    print(some_string.num_uint64(8589934592)) # Warning.

    TestStaticCalledOnInstance.static_func()
    static_func()
    self.static_func()
    var other := TestStaticCalledOnInstance.new()
    other.static_func() # Warning.

    Inner.static_func()
    var inner := Inner.new()
    inner.static_func() # Warning.
```

</details>

Note that `self.static_func()` does **not** trigger the warning. I would prefer that there be no warning only without an explicit `self`.

modules/gdscript/tests/scripts/parser/warnings/static_called_on_instance.out
dalexeev, line 1: <details>
<summary>Suggestion</summary>

```
GDTEST_OK
>> WARNING
>> Line: 16
>> STATIC_CALLED_ON_INSTANCE
>> The function "num_uint64()" is a static function but was called from an instance. Instead, it should be directly called from the type: "String.num_uint64()".
>> WARNING
>> Line: 22
>> STATIC_CALLED_ON_INSTANCE
>> The function "static_func()" is a static function but was called from an instance. Instead, it should be directly called from the type: "TestStaticCalledOnInstance.static_func()".
>> WARNING
>> Line: 26
>> STATIC_CALLED_ON_INSTANCE
>> The function "static_func()" is a static function but was called from an instance. Instead, it should be directly called from the type: "Inner.static_func()".
8589934592
8589934592
```

</details>

dalexeev (commented):
Makes sense to me, although the point about non-global classes still applies.

dalexeev (approved):

akien-mga:
Thanks!

Issue #85850, opened 2023-12-06
MeshInstance3D's Surface Material Override gets reset when reimporting mesh


Lippanon:
### Tested versions

v4.3.dev.custom_build [d76c1d0e5]

### System information

Windows 10, Vulkan

### Issue description

Not sure if this is a documentation issue or a MI3D bug. In any case, it's an annoying workflow issue, it's common to make changes to models, and when they have multiple surfaces/slots with different materials, one has to keep setting them. 
GeometryInstance3D's ```material_override``` is much better to work with but not suitable for every case.

A mesh with materials set in the Surface Material Override):
![image](https://github.com/godotengine/godot/assets/115422160/79e661b3-a066-4825-93d0-b81632c2aada)

The associated functions ```set_surface_override_material / get_surface_override_material``` belong to MeshInstance3D.
The docs state "Sets the override material for the specified surface of the Mesh resource. This material is associated with this MeshInstance3D rather than with mesh."
This leads one to think, by being associated with the MI3D node and not the mesh, they should not be reset by simply reimporting the mesh (after changing a setting for example) in the importer:
![image](https://github.com/godotengine/godot/assets/115422160/ffa10661-2c60-4da8-8e21-2b9a95dd2fa0)

However, they do get reset when reimporting.
I understand that changing the Mesh in the MeshInstance3D might change the number of surface slots, even so you could keep the set materials that fit the "new" amount of slots, At least it would be nice if, when the path of the Mesh remains the same, it attempts to keep the materials.

### Steps to reproduce

1. Open the MRP, notice the MeshInstance3D with a cube Mesh and 2 materials set in the override slots;
2. Double click 'a_blender_cube.glb';
3. Select Reimport, notice the Cube is still associated with the MeshInstance3D node, but the override slots have been cleared;

### Minimal reproduction project (MRP)

[SMO_cleared_bug.zip](https://github.com/godotengine/godot/files/13587883/SMO_cleared_bug.zip)

OhiraKyou:
I've been encountering this since updating from 4.1.3-stable to v4.2.stable.official [46dc27791]

KoBeWi:
Duplicate of #79779

Issue #83575, opened 2023-10-18
Duplicating a scene also duplicates child nodes' UIDs.


Ratslayer:
### Godot version

4.1.2.stable.mono

### System information

Windows 11

### Issue description

Duplicating a scene duplicates UIDs of all sub-nodes, such as MeshInstance3D.

### Steps to reproduce

There is a single script, PrintChildUID, that exposes a MeshInstance3D node and prints its UID in the _Ready method.
The scene print_node.tscn is a simple node3D scene with the PrintChildUID script and a single MeshInstance3D child
Repro steps:
Open main_scene.tscn
Duplicate any node in the scene
Launch the game
Observe that UIDs of duplicated nodes are all the same in the output, while they should be different.

### Minimal reproduction project

[uid-bug-repro.zip](https://github.com/godotengine/godot/files/13032880/uid-bug-repro.zip)

Rindbee:
When copying a node, the exported variable `_mesh` on the new node will reference the original value. This doesn't seem to be a bug.
```C#
GD.Print($"_mesh UID: {_mesh.GetInstanceId()} ", $"child UID: {GetNode("MeshInstance3D").GetInstanceId()}");
```
Although the `_mesh` is exported as node type `MeshInstance3D`, it is recorded as `NodePath` in tscn.
```tscn
[node name="Node3" parent="." instance=ExtResource("1_vwf8e")]

[node name="Node4" parent="." instance=ExtResource("1_vwf8e")]

[node name="Node5" parent="." node_paths=PackedStringArray("_mesh") instance=ExtResource("1_vwf8e")]
_mesh = NodePath("../Node4/MeshInstance3D")
```

Rindbee:
The problem is that you use `_mesh` to cache child node. Then you seem to be weird that the `_mesh` of the copied node has the same uid (refers to the same node) .

Rindbee:
There are two ways to think about it:
1. Node instance: Saving the same instance is expected;
2. NodePath: Keeping the value (usually a relative path) is also expected;

Consider a case where the exported variable cache is a node in the main scene. If you copy the node, and then paste it in a different node tree depth, there may be problems with the relative path, keep the same instance may be better.

In my opinion: Node instance is suitable for the above case; and `NodePath` is used for cases where the relative path does not change, such as your case.

Rindbee:
[...]

Issue #83200, opened 2023-10-12
(Re)importing Assets breaks apart on GLB/GLTF materials at clean project import because of an CSV file


wp2000x:
### Godot version

4.1.2-stable

### System information

Windows 11

### Issue description

The auto (Re)Importing of assets breaks apart, when a CSV file is re-imported in a project?

I noticed that, when a clean git checkout of an project is made (without the .godot folder), the auto (re)import breaks apart at GLB/GLTF textures/materials. After some time i could narrow down the suspect to an CSV file in my project and the corresponding autogenerated `.csv.import` file. When the `.csv.import file` does not exist in the repo/project it gets successfuly imported, the file gets created and everything is fine. The problem starts when it gets commited and someone else checks it out and tries to rebuild the .godot folder. 

This will result in following error (MRP example red texture cube model):
`WARNING: glTF: Image index '0' couldn't be loaded with the name: Image_0. Skipping it.`
Screenshot of it:
![2023-10-12_17-33-58](https://github.com/godotengine/godot/assets/70962730/a503698d-701b-49ff-8c0d-3c4fab4b6f32)

The material is broken (at this state, only a manual reimport by hand can fix this):
![2023-10-12_17-45-12](https://github.com/godotengine/godot/assets/70962730/b38c3d70-f76d-46ad-9d55-b6d31abbfff2)
And here how the model looks when the csv files does not exist:
![2023-10-12_17-45-44](https://github.com/godotengine/godot/assets/70962730/cc0445c8-4b73-4cc7-a06d-866f17cd800d)

### Steps to reproduce

* Run both MRP projects
* MeshImportBug_OK should run fine and show a red textured cube at the dummy scene and the MeshImportBug_Broken one not.

### Minimal reproduction project

[MeshImportBug_OK.zip](https://github.com/godotengine/godot/files/12884441/MeshImportBug_OK.zip)
[MeshImportBug_Broken.zip](https://github.com/godotengine/godot/files/12884444/MeshImportBug_Broken.zip)

fire:
I have no idea. Maybe there is a text comparison that treats .gltf.import and .gltf.csv.import the same

bitsawer:
Can confirm, tested on current master (37ee293be82e625e68704a78477b841d0ba17b8f). The broken project starts working after deleting the .csv file, its generated .translation file and the .godot folder and letting Godot reimport everything.

[...]

PR #85922, opened 2023-12-08
Prevent overriding file info of another file when reimport creates extra files

### LINKED ISSUE #75525: Glb import error
### Godot version

### System information

Windows 10, 64bit

### Issue description

First, please run the reproduction project.
When I run the test scene, It report error like follow and app exited. 
image
And I don't know where was error.

### Steps to reproduce

1. Open the reproduction project.
2. Open scene 'user:\\xd\XActors\Weapon\Gun\Rifle\XRifle.tscn'
3. Assign the scene's Meshinstance's mesh with 'user:\\xd\XActors\Weapon\Gun\Rifle\XRifle01.res'
4. Run the scene
Opps, the scene exited.
GIF

[GlbTest.zip](https://github.com/godotengine/godot/files/11120181/GlbTest.zip)

NOTE: The project that I provided was picked up from my owner project. Sorry I can't provide a minimal reproduction project because I don't know which step cause this issue. I was test like this:
Create a project, import the glb file, create meshinstance and assign mesh, but there was not error!

### LINKED ISSUE #77978: Binary mesh resource disappears from MeshInstance3D upon project restart
### Godot version

v4.1.beta.custom_build [72b59325c]

### System information

Windows 10, Vulkan

### Issue description

This has been plaguing me for a while in my main project. I can only partially reproduce it in a MRP:
When placing a binary mesh resource imported from a .glb file into a MeshInstance3D, it will be gone upon restart. It doesn't seem to happen to every resource, but I've had MIs that were fine and then start doing it out of nowhere.
I thought it was going to be related to materials due to things like https://github.com/godotengine/godot/issues/77007 but I can reliably trigger it with the steps listed below.
I have done limited testing on .tres files because I usually use .res for size/performance, but in the MRP .tres doesn't seem to trigger it, so perhaps it doesn't affect those.

Note on the Steps to reproduce: you have to delete the test_cube.res and redo the steps to trigger the bug again, but in my main project the affected resources always disappear.

### Steps to reproduce

1. Double click test_cube.glb
2. Click the mesh (test_cube_Cube_097), set save to file path as: res://test_cube.res
3. Make sure to click the little folder icon and then Save (this seems to make it consistent to proc)
4. Press Reimport
5. Drag the test_cube.res to the Mesh slot of the MeshInstance3D
6. Ctrl+s to save
7. Project -> Reload current project
8. Mesh is no longer in the MeshInstance3D

[binary res bug.zip](https://github.com/godotengine/godot/files/11681009/binary.res.bug.zip)

### LINKED ISSUE #80132: MeshInstance3D loses reference to mesh resource
### Godot version

4.1

### System information

Godot v4.1.1.stable.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2070 SUPER (NVIDIA; 31.0.15.2802) - AMD Ryzen 9 3900X 12-Core Processor (24 Threads)

### Issue description

Saving a MeshInstance3D with a mesh resource imported from a blender file will lose the reference to the mesh resource when reopening de project.

When testing I have found that the uid referencing the mesh saved in the .tscn is the same as the one int the .blend.import and not an unique one for the specific mesh
uid in .tscn: image
uid in .blend.import: image
I don't know if that is the cause of the error but I have found a way to force it to generate a unique uid:

1. select the mesh in the MeshInstance3D and "Make Unique"
2. reassign the mesh resource to the MeshInstance3D

now the .tscn has a unique uid for the mesh and remains referenced when reopening the project (this happens randomly sometimes too when working on the scene)
image

Related: [ #57053](https://github.com/godotengine/godot/issues/57053)

### Steps to reproduce

1. import a .blend file with a mesh (if the .blend has multiple mesh the error doesn't seems reproduce)
2. open the blender importer, set mesh paths and reimport
3. create a new MeshInstance3D node and assign the created .res as a mesh resource
4. save the scene and restart Godot.

[mesh_test.zip](https://github.com/godotengine/godot/files/12232647/mesh_test.zip)

### LINKED ISSUE #83154: glTF import won't update extracted texture
### Godot version

4.1.2 stable

### System information

Godot v4.1.2.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated Quadro P620 () - 11th Gen Intel(R) Core(TM) i9-11900K @ 3.50GHz (16 Threads)

### Issue description

The textures packed into glb exports won't always update when the inherited glb is updated from blender.  It might use an old texture, appear textureless, or sometimes just work and I simply cannot figure out what i am doing wrong.

### Steps to reproduce

- open project in godot and browse empty testcube folder
- copy testcube-1.glb from zip root into godot_project\testcube as testcube.glb
- switch back to godot and import glb as new inherited scene and save tscn in testcube folder. glb may appear with red "x", but reimporting fixes it?
- copy testcube-2.glb from zip root into godot_project\testcube as testcube.glb
- switch back to godot, image texture should change with some errors in output log
- copy testcube-1.glb from zip root into godot_project\testcube as testcube.glb
- cube might appear with changed texture or appear white/textureless. it has been difficult to get this to predictably happen

white-cube-errors

errors in output log appear: 

```
  Can't find file 'uid://w87lq3421wmi::::res://testcube/testcube_godette.png'.
  Attempt to disconnect a nonexistent connection from 'testcube:<Node3D#2317655010421>'. Signal: 'replacing_by', callable: 'EditorNode::set_edited_scene'.
  Can't find file 'uid://c5oarbyq0oygi::::res://testcube/testcube_godot.png'.
  Attempt to disconnect a nonexistent connection from 'testcube:<Node3D#2486517690152>'. Signal: 'replacing_by', callable: 'EditorNode::set_edited_scene'.
```

blend file notes: Blender 3.6.0, this is the default cube and default material with an image texture of a png file, the only difference between the two glb exports is which files is used. glbs are exported "limited to selected objects" with only the cube selected.

[glb-import-issue.zip](https://github.com/godotengine/godot/files/12872041/glb-import-issue.zip)

Listwon:
Fixes #80132, fixes #75525, fixes #77978, fixes #83154

Special thanks to @MetRiko and @gonzelvis for helping me figure out the exact repro steps. They had issues with `*.glb` files and repro turned out to be exactly the same as in #80132.

`EditorFileSystem::_reimport_file` calls `_find_file(p_file, &fs, cpos)` to get the file index in `filesystem` (`fs`). The problem is - if reimport creates extra fiiles, the `filesystem` gets invalidated and the next usage of `cpos` is incorrect - `_reimport_file` updates wrong file info leading to duplicated uid, dependencies etc. The uid inside created file is still correct at this point, but the data in `fs->files[cpos]` is invalid, so saving the scene after assigning the affected resource causes the resource to save with wrong uid.

Adding `_find_file(p_file, &fs, cpos)` just before the next  line  with `cpos` refreshes the `cpos` index so now it points to a proper file in `fs->files`.

### DIFF
--- a/editor/editor_file_system.cpp
+++ b/editor/editor_file_system.cpp
@@ -2154,6 +2154,9 @@ Error EditorFileSystem::_reimport_file(const String &p_file, const HashMap<Strin
 		}
 	}
 
+	// Update cpos, newly created files could've changed the index of the reimported p_file.
+	_find_file(p_file, &fs, cpos);
+
 	//update modified times, to avoid reimport
 	fs->files[cpos]->modified_time = FileAccess::get_modified_time(p_file);
 	fs->files[cpos]->import_modified_time = FileAccess::get_modified_time(p_file + ".import");


KoBeWi (approved):
The issue is confusing to reproduce and the fix is not obvious. I'd add a comment next to the new `_find_file()` explaining why is it called.

Other than that looks ok.

MetRiko:
@KoBeWi 
Here is some my thoughts about this PR to show it might be little more important than it looks like.

Every time I was trying to do something with Godot 4.x (personal projects, game jams etc.), I was always reaching the point where random scenes were getting corrupted. It was a constant thing for me especially in the 3D projects.

I was trying to fix scenes with the "Fix dependency" there never was any error.
I also looked into these "corrupted" .tscn with the text editor but then again.. I wasn't able to spot anything wrong. 

After long investigation of this issue (with the huge help of @gonzalveis) we were able to found the exact source of those problems and 100% working repro steps that could lead into corrupted scenes (and more). 

TL;DR 
This PR fixes problem which indirectly could cause:
- disappearing set references to resources inside nodes,
- invisible meshes on the scene (but the instance of such .tscn has visible mesh),
- scene is invalid/corrupt but works in the runtime,
- duplicated UIDs for two different resources,
- resource UID is changed on the saving project,
- restarting projects breaks resource references,
- and many more issues related to importing/creating resources (especially the extracted resources from .glb files).

I don't have time to examine it further, but after what I learnt when I was investigating this bug I think this probably fixes the following issues (you can see most of those issues are related to glTF and/or MeshInstance3D):
#80132, #68672, #77978, #75970, #63513, #83575, #79545, #75525, #83200, #83154, #82275, #80536, #79779, #81671, #77978, #79275, #85850, #73916, #54774

Those above would be worth to check first but there might be the chance this also can fix some of those issues:
#84981, #82652, #82422, #82702, #76236, #77007, #76691

And for testing purposes, the easiest repro steps for the bug, that this PR is fixing, are the following:
0. Prepare any .glb. If you don't have any then exported default cube from Blender will work for the repro too:
- Open Blender, select default cube, export it as .glb (with enabled option "Limit to: Selected Objects" in the export settings).
1. Create empty project with Godot 4.2, create scene with any type of node and save it.
2. Put .glb inside the project -> open it with double click to open glTF importing window
3. Export meshes from this .glb in any way you want, preferable is: 
- Press "Actions..." -> "Set Mesh Save Paths" -> "Select Current Folder" (.tres/.res setting doesn't matter) -> "Set Paths" -> "Reimport"
4. Double click on the any created meshes (exported from .glb) in the filesystem to refresh editor (DON'T SAVE ANYTHING!)
- Now you can see one of those meshes has the SCENE icon instead of MESH icon (what is weird)
5. Check UID of this mesh with wrong icon ("RMB-> Copy UID" and paste it somewhere)
6. Add this mesh to the scene (for example by drag and dropping it on the scene)
7. Save the scene with 3D node (just press ctrl+s)
- You can see icon for this mesh resource has been changed to the valid mesh icon
8. Check UID of this mesh resource again and compare with previous one: 
- **UID for this resource has been changed!**
- **This UID is exactly the same as the .glb file has!**

Now using such corrupted mesh with duplicated UID anywhere will lead to many different problems like corrupting scenes.

KoBeWi:
@MetRiko I checked these issues and was able to confirm 2 more that this PR fixes (already added them in the OP). I also closed 2 issues as already resolved/duplicate.
Other issues seem unrelated, though in some cases I think this PR does not fix things automatically; it's more to prevent the bugs from happening, which makes it more difficult to test as a fix.

akien-mga:
Thanks! Great work tracking this down :1st_place_medal:

MetRiko:
@KoBeWi 
> Other issues seem unrelated, though in some cases I think this PR does not fix things automatically

That's exactly what I meant by:
> this PR fixes problem which indirectly could cause

and by 
> Now using such corrupted mesh with duplicated UID anywhere will lead to many different problems like corrupting scenes.

This will not fix directly problems caused by duplicated UIDs but will prevent UID's duplication. There is no reason in "fixing" bugs caused by using resources with duplicated UIDs because engine is not even suppose to have duplicated UIDs anywhere at all. Each resource should always have one unique UID right? So there is no point in fixing those anomalies, however it would be nice to have a good readable error/message that will point out duplicated UIDs and how to fix it.

lyuma:
Wow!! Thanks for fixing this! I think you and I may have discovered the problem independently [around the same time, Wednesday last week](https://twitter.com/Lyuma2d/status/1732284046519214225). If I had submitted my discovery as a pull request, it might have saved you the time of doing this. Based on my diagnosis/understanding, this fix is correct.

The issue only affected files beginning with the letters "r" through "z" due to another problem in EditorFileSystem where it tries to keep directories sorted, but compares all strings against the full path starting with "res://", so most files were simply appended to the end of the directory, which would then not affect the `cpos`. (My workaround was to extract all embedded resources into a subdirectory, so as to not affect `cpos`.)

> because engine is not even suppose to have duplicated UIDs anywhere at all

What happens if the user manually copies and pastes a file elsewhere in the project, and this file has a particular UID embedded in its .res header?

MetRiko:
@lyuma 
> What happens if the user manually copies and pastes a file elsewhere in the project, and this file has a particular UID embedded in its .res header?

Let's say you have file Entity.tscn and you will create EntityCopy.tscn by duplicating it outside Godot editor (in OS file explorer or with commands). In that case Entity.tscn and EntityCopy.tscn will have the same UID.

Now if you will:
- remove UID part from EntityCopy.tscn with text editor,
- open EntityCopy.tscn inside Godot and press ctrl+s,
then EntityCopy.tscn will get the same UID as before (still duplicated).

However if you will:
 - remove UID part from EntityCopy.tscn with text editor,
 - press "Project->Reload Current Project",
 - open EntityCopy.tscn inside Godot and press ctrl+s,
then new UID for EntityCopy.tscn will be generated.

For the references.. let's say you used EntityCopy.tscn inside World.tscn when EntityCopy has still duplicated UID.
If you will open World.tscn and save it (ctrl+s), Godot will try to update `ext_resource` header automatically (you don't have to reload project). Depend on the situation, the header will be updated differently. There are 5 cases worth to mention:
- `uid=*, path=*` - nothing will be updated.
- `---, path=*` - uid will be taken from the file pointed by path.
- `uid="", path=*` - uid will be taken from the file pointed by path.
- `uid=*, ---` - SCENE CORRUPTED! You cannot open such scene within the Godot.
- `uid=*, path=""` - path will be determined based on the uid.

(where `uid=*` means property with proper path, `uid=""`, property with empty path and `---`, property removed entirely).

I think this should be raised as a separated issue/proposal.

YuriSizov:
Cherry-picked for 4.2.2.

YuriSizov:
Cherry-picked for 4.1.4.

PR #85934, opened 2023-12-08
Separate Variant enums (Type, Operator) to standalone file

Repiteo:
As part of going through and trying to split some of the bulkier classes into more managable wholes, and trying to get a feel for overall dependency across files, the single biggest roadblock I've struggled with is how many files/functions rely on `Variant::Type` that otherwise have nothing to do with `Variant`. This has led to scenarios where the enum would regularly be substituted with `int` or `uint32_t`, with `Variant::Type` being implicit or specified in documentation, but rarely enforced

This PR aims to solve this by separating that enum (alongside the much less ubiquitous `Variant::Operator`) into a dedicated file: `variant_enums.h`. As a **lot** of files utilized this enum, this PR covers a pretty staggering amount of ground. But due to the nature of the change, I couldn't imagine any easy way to split it up, so it's being proposed as one, big chunk. The overwhelming majority of changes was just a matter of converting `Variant::Type` to `VariantType`, but some more nuanced adjustments were made:
- **Converting the enums to enum classes.** Because they're no longer tied to a class, the enum names would've overlapped with several existing keywords & outright conflicted with `AABB` and `RID` across all files. While it'd be *feasible* to apply the double-colon prefix for the latter, the former would've required *dramatic* changes across the board. Instead making them into enum classes, comparable to the Input enums, means that we introduce simplified names in exchange for a mandatory prefix. This is most obvious with `VariantOperator` not needing the `OP_` prefix anymore (minus `OP_IN`, which otherwise fails to compile even in an enum class), with `VariantType` only having one non-simplified name already (`VARIANT_MAX` → `MAX`).

- **`AABB()` & `RID()` no longer need double-colon prefixes.** The prefixing was primarily (only?) needed in classes/functions nested within Variant, as the non-class enum `Type` would take precedence over the identically named classes. As this implicit nature is no longer possible with an enum class, the double-colon prefixing workaround has been discarded.

- **Array now uses `VariantType` instead of `uint32_t` where appropriate.** While other files/functions exist that can be changed to use `VariantType` instead of `int`, godot's array was *by far* the most obvious one. It was impossible previously because Array had no real means of using an enum from a class it couldn't inherit from, but now it can safely include `variant_enums.h` without issue.

- **C#: Rename property `Variant.VariantType` to `Variant.Type`.** This is perhaps the compatability-breaking change, and one I'm willing to revert if needed. However, there was a long-standing `TODO` in the C# `Variant` file to rename "Variant.Type" to "VariantType" and "VariantType" to "Type". As this PR is already doing the former, it seemed like the right time to implement the latter.


Repiteo:
My thought process was to gradually adjujst inclusion chains with a new enum system in place, but if there's already precedent for treating enums as ints in the codebase then my initial direction might've been misguided. I would still hope for proper enum assignment support in areas where dependency overlap might be a concern, but it's entirely possible there's a much less destructive means of achieving that

Repiteo:
Closing for now. Couldn't think of a wholly justified reason for making this change when exceptions in the implementation are already intentionally in place

PR #85944, opened 2023-12-08
Fix OpenSLES cleanup on Android

Alex2782:
Fixes the following error output: 
Bildschirmfoto 2023-12-08 um 23 19 02

I'm not sure if this is critical. This happens when the app is terminated via the "BACK" button. On Android, however, the app is not completely terminated, the app is still in the app list in the background and can be restarted.

### DIFF
--- a/platform/android/audio_driver_opensl.cpp
+++ b/platform/android/audio_driver_opensl.cpp
@@ -313,6 +313,15 @@ void AudioDriverOpenSL::unlock() {
 }
 
 void AudioDriverOpenSL::finish() {
+	(*playItf)->SetPlayState(playItf, SL_PLAYSTATE_STOPPED);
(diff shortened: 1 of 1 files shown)


AThousandShips:
See also:
* https://github.com/godotengine/godot/pull/85955

Alex2782:
I think it can be closed here, the extensions from @PierceLBrooks look more extensive.

[destroy](https://developer.android.com/ndk/guides/audio/opensl/opensl-prog-notes#destroy)

> We recommend that you explicitly set both the primary object interface and all associated interfaces to NULL as part of your object destruction sequence, which prevents the accidental misuse of a stale interface handle.

Pierce did that too

Alex2782:
--> https://github.com/godotengine/godot/pull/85955

PR #85964, opened 2023-12-09
Enable tests for .NET CI

HolonProduction:
Based on the discussion in #85178

During some (admittedly not very thorough) tests I couldn't reproduce the CI freeze. @AThousandShips wasn't able to reproduce either. It is possible, that in the two years since it was disabled this was fixed somehow. But further testing is necessary.

Why?
Having .NET during tests would allow to test whether GDScript autocompletion correctly works for scripts in other languages.

### DIFF
--- a/.github/workflows/linux_builds.yml
+++ b/.github/workflows/linux_builds.yml
@@ -23,13 +23,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


HolonProduction:
Closing due to seemingly no interest. And I'm not really up to date with our current CI config anymore anyway.

PR #83138, opened 2023-10-11
Fix AABB Ray Intersection

Kimau:
The old function was incorrect returning a per axis t Tested this and it works correctly returning same result as intersects_segment

PR #85966, opened 2023-12-09
Expose a method to get hovered Control in Viewport

Kimau:
Found myself needing this to resolve some UI stuff in a clean way.

### DIFF
--- a/doc/classes/Viewport.xml
+++ b/doc/classes/Viewport.xml
@@ -128,6 +128,13 @@
 				Returns the [Control] having the focus within this viewport. If no [Control] has the focus, returns null.
 			</description>
 		</method>
+		<method name="gui_get_hovered_control" qualifiers="const">
+			<return type="Control" />
+			<description>
+				Returns the [Control] that the mouse is currently hovering over in this viewport. If no [Control] has the cursor, returns null.
+				Typically the leaf [Control] node or deepest level of the subtree which claims hover. This is very useful when used together with [method Node.is_ancestor_of] to find if the mouse is within a control tree.
+			</description>
+		</method>
 		<method name="gui_is_drag_successful" qualifiers="const">
 			<return type="bool" />
 			<description>
--- a/scene/main/viewport.cpp
+++ b/scene/main/viewport.cpp
@@ -3518,6 +3518,11 @@ Control *Viewport::gui_get_focus_owner() const {
 	return gui.key_focus;
 }
 
+Control *Viewport::gui_get_hovered_control() const {
+	ERR_READ_THREAD_GUARD_V(nullptr);
+	return gui.mouse_over;
+}
+
 void Viewport::set_msaa_2d(MSAA p_msaa) {
 	ERR_MAIN_THREAD_GUARD;
 	ERR_FAIL_INDEX(p_msaa, MSAA_MAX);
@@ -4557,6 +4562,7 @@ void Viewport::_bind_methods() {
 
 	ClassDB::bind_method(D_METHOD("gui_release_focus"), &Viewport::gui_release_focus);
 	ClassDB::bind_method(D_METHOD("gui_get_focus_owner"), &Viewport::gui_get_focus_owner);
+	ClassDB::bind_method(D_METHOD("gui_get_hovered_control"), &Viewport::gui_get_hovered_control);
 
 	ClassDB::bind_method(D_METHOD("set_disable_input", "disable"), &Viewport::set_disable_input);
 	ClassDB::bind_method(D_METHOD("is_input_disabled"), &Viewport::is_input_disabled);
--- a/scene/main/viewport.h
+++ b/scene/main/viewport.h
@@ -610,6 +610,7 @@ public:
 
 	void gui_release_focus();
 	Control *gui_get_focus_owner() const;
+	Control *gui_get_hovered_control() const;
 
 	PackedStringArray get_configuration_warnings() const override;
 


scene/main/viewport.cpp
```diff
@@ -3518,6 +3518,11 @@ Control *Viewport::gui_get_focus_owner() const {
 	return gui.key_focus;
 }
 
+Control* Viewport::gui_get_mouse_over() const {
```
AThousandShips, line 3521:
```suggestion
Control *Viewport::gui_get_mouse_over() const {
```

AThousandShips:
You need to add documentation for this new method as well, see [here](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html#updating-class-reference-when-working-on-the-engine)

doc/classes/Viewport.xml
```diff
@@ -128,6 +128,12 @@
 				Returns the [Control] having the focus within this viewport. If no [Control] has the focus, returns null.
 			</description>
 		</method>
+		<method name="gui_get_mouse_over" qualifiers="const">
+			<return type="Control" />
+			<description>
+				Returns the [Control] which is the leaf node of mouse over behaviour within this viewport. If no [Control] has the cursor, returns null.
```
AThousandShips, line 134: What does this mean? I think this needs a more clear description as to the average user the meaning of "leaf node" isn't clear.

And in either way, the documentation and code uses US English 🙂 
```suggestion
				Returns the [Control] which is the leaf node of mouse over behavior within this viewport. If no [Control] has the cursor, returns null.
```

Kimau:
I was trying to keep the naming convention as close to the underlying code as possible.
I thought the fact the return type was Control was self documenting enough to show the behaviour.

I have tried to make it clear in the docs.

AThousandShips:
The return type isn't always obvious from the name in for example autocomplete, but I think the name works, we'll see what others think 🙂

Kimau:
Tried to elaborate more in the docs but I did still use the term leaf node and the likely function a user would want to use with it.

doc/classes/Viewport.xml
```diff
@@ -128,6 +128,16 @@
...
 			</description>
 		</method>
+		<method name="gui_get_mouse_over" qualifiers="const">
+			<return type="Control" />
+			<description>
+				Returns the [Control] that the mouse is currently hovering over in this viewport. 
+				If no [Control] has the cursor, returns null.
+
+				This value is always [Control] which is the has the mouse, typically the leaf node or deepest level of the tree which claims focus.
+				This is very useful when used together with [method Node.is_ancestor_of] to find if the mouse is within a control group.
```
AThousandShips, line 138:
```suggestion
				Returns the [Control] that the mouse is currently hovering over in this viewport. If no [Control] has the cursor, returns null.
				This value is always [Control] which is the has the mouse, typically the leaf node or deepest level of the tree which claims focus. This is very useful when used together with [method Node.is_ancestor_of] to find if the mouse is within a control group.
```
Each separate line is a new paragraph, this is much more compact, otherwise it would be very spread out

I'm not entirely sure about the wording of "group" as it might be confused with the existing concept of groups (which it isn't here) but can't think of a different wording

AThousandShips (approved):
LGTM, someone with detail knowledge of the mouse over system should look at the docs to check the wording and claims

Kimau:
Maybe control tree or stack is better than group 🤷

AThousandShips:
"Subtree" or "hierarchy" might be best, as they are pretty established

AThousandShips:
CC @Sauermann If I'm not mistaken you've worked with the viewport input area

scene/main/viewport.cpp
```diff
@@ -4557,6 +4562,7 @@ void Viewport::_bind_methods() {
 
 	ClassDB::bind_method(D_METHOD("gui_release_focus"), &Viewport::gui_release_focus);
 	ClassDB::bind_method(D_METHOD("gui_get_focus_owner"), &Viewport::gui_get_focus_owner);
+	ClassDB::bind_method(D_METHOD("gui_get_mouse_over"), &Viewport::gui_get_mouse_over);
```
AThousandShips, line 4565:
```suggestion
	ClassDB::bind_method(D_METHOD("gui_get_mouse_over"), &Viewport::gui_get_mouse_over);
```
Whitespace

AThousandShips:
Updated the title as the binding isn't exclusive to GDScript

doc/classes/Viewport.xml
```diff
@@ -128,6 +128,13 @@
 				Returns the [Control] having the focus within this viewport. If no [Control] has the focus, returns null.
 			</description>
 		</method>
+		<method name="gui_get_mouse_over" qualifiers="const">
+			<return type="Control" />
+			<description>
+				Returns the [Control] that the mouse is currently hovering over in this viewport. If no [Control] has the cursor, returns null.
+				This value is always [Control] which is the has the mouse, typically the leaf node or deepest level of the tree which claims focus. This is very useful when used together with [method Node.is_ancestor_of] to find if the mouse is within a control tree.
```
Sauermann, line 135: > which is the has the mouse

this doesn't sound like a proper English sentence.

Sauermann, line 135: > which claims focus

In regard to Control nodes, "focus" has the meaning, that is described for example on on this docs page: https://docs.godotengine.org/en/latest/tutorials/ui/gui_navigation.html

The Godot concept of Control node focus is unrelated to the mouse-over state. In this context, I would consider this wording ambiguous and would prefer, if "focus" is not mentioned here.

Sauermann (commented):
Implementation looks correct, didn't test it however.

doc/classes/Viewport.xml
Kimau: Tried reworking the doc string, is that better?

YuriSizov:
> I was trying to keep the naming convention as close to the underlying code as possible.
> I thought the fact the return type was Control was self documenting enough to show the behaviour.

I would say that in this case our internal naming convention is not great and should be adjusted alongside your changes if we seek consistency, or can be left as is to avoid making this PR bigger than it needs to be. But the public method needs to have a subject in its name. Internal properties and methods don't always have good names for public API because they are never considered from this perspective.

So something like `gui_get_hovered_control` would be ideal for a public method name.

elfajriharsuda (approved):

YuriSizov:
Updated and rebased on your behalf.

YuriSizov (approved):

Kimau:
> Thanks, and congrats on your first merged Godot contribution!

Thanks for making the changes and merging work for busy. Yeah funny this one beat #83138 😂

But yeah this was actually simple 👍

pc9098:
Maybe an array of all the Control nodes that are hovered would be better? Or is there a way to do that?

PR #85968, opened 2023-12-09
Bump minimum SCons version to 4.0

Repiteo:
Now matches the minimum version used by godot-cpp. In addition, adjusted a handful of sections that needed workarounds for non-4.0 versions; namely: `compile_db` is now just natively available & `methods.dump` is replaced with a built-in alternative. GitHub Actions already have a minimum version of 4.4 specified, so no changes are needed there (unless we wanted to deliberately downgrade).

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-EnsureSConsVersion(3, 0, 0)
+EnsureSConsVersion(4, 0)
 EnsurePythonVersion(3, 6)
 
 # System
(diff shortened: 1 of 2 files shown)


Calinou:
Note that Ubuntu 20.04 only has SCons 3.1.2 packaged, but Ubuntu 22.04 has SCons 4.0.1: https://packages.ubuntu.com/search?keywords=scons

akien-mga:
Partially superseded by #92043. We can still consider moving to 4.0+ in the near future though.

Repiteo:
Superseded by #99134

PR #85969, opened 2023-12-09
Pin addons folder to the top

miv391:
Implements this proposal: https://github.com/godotengine/godot-proposals/issues/8615

- Addons folder is now always the first folder in the FileSystem tree, no matter how the files are sorted. 
- Addons folder has now a new icon of its own, a mix of a folder and a puzzle piece.
- Addons folder is detected by its name, "addons", case sensitively.
- This special treatment is done only to the top level addons folder. If any subfolder has a folder named "addons", it is treated as a normal folder.

The icon's SVG is somewhat different than the other icons. My SVG skills are unfortunately a bit lacking. 

Now FileSystem looks like this: 

kuva

4x zoom:
kuva

- *bugsquad edit: implements godotengine/godot-proposals#8615*


AThousandShips:
CC @MewPurPur any ideas for design?

MewPurPur:
Not something I'm too familiar with, but the design looks perfect to me.

MarioLiebisch:
Yeah, think this looks good for the icon size.

Not really related: This might look even more clean with the topmost `res://` being removed. Afaik there's no second root (like `user://`) shown in any case, right?=

miv391:
> Not something I'm too familiar with, but the design looks perfect to me.

Thanks! The SVG code however is quite messy compared to the existing icons. Are there somewhere any instructions how to clean and optimize the SVG to the same level as the existing icons?

KoBeWi:
> Not really related: This might look even more clean with the topmost res:// being removed.

`res://` is useful for accessing/opening project's root folder.

MewPurPur:
> Thanks! The SVG code however is quite messy compared to the existing icons. Are there somewhere any instructions how to clean and optimize the SVG to the same level as the existing icons?

I'll clean it up later on along with others, dw about it.

KoBeWi:
The implementation looks alright at a glance (did not test yet), but it introduces lots of String comparisons for every folder. Maybe it can be implemented in a more efficient way, e.g. by taking advantage of the fact that `res://` is always the root of the file tree?

miv391:
> The implementation looks alright at a glance (did not test yet), but it introduces lots of String comparisons for every folder. Maybe it can be implemented in a more efficient way, e.g. by taking advantage of the fact that `res://` is always the root of the file tree?

I made some changes:

- Every top level (level = 1) folder makes a string comparison to check which icon should be used. I shortened that comparision a little bit by using `dir_level` and `dname` instead of `lpath`.
- Root level (level = 0) needs to iterate its sub folder names once to find the possible addons folder. Subdir loop does now only an int comparison instead of 1-2 string comparisons.

I _could_ remove the string comparison for the icon by adding one more parameter to `_create_tree` function that would tell if the folder is addons folder. But I feel that that would be too much optimizing.

KoBeWi (dismissed):

YuriSizov:
I like the idea of having a dedicated icon/styling for it, but I don't think we should be messing with the order of folders. At the very most it should be an option, which I would keep disabled by default. Alphabetical order is superior and more natural to follow, and I have never seen important folders being pinned like that in any other project-based tool. All such tools stop at giving such folders a different styling, and I think we should too.

YuriSizov (changes requested):
See above.

MarioLiebisch:
Sticking: I think that might be an interesting option in general, being able to make folders sticky. Or maybe just "pin" them while scrolling to show the current depth? VS Code added the latter recently (for the code editor as well as the file selection).
VS Code's sticky text

Icon: Maybe just go with a full puzzle piece instead? A second "sub icon" feels too tiny for those dimensions, even if you're on a High DPI screen.

KoBeWi:
> At the very most it should be an option, which I would keep disabled by default.

Disabled by default usually means most people won't even know about it.

> Sticking: I think that might be an interesting option in general, being able to make folders sticky.

That's what favorites are for tbh (except you don't see contents of favorite folders).

YuriSizov:
> Disabled by default usually means most people won't even know about it.

Yes, I know. It can be a display option in the dock itself or something like that. That should help with the discoverability a bit. But I'm not convinced this is going to be helpful for the majority of users to have it enabled by default. I believe most would prefer the natural, alphanumerical order. Especially if they have more folders above `addons`.

I think having a custom icon would already solve the problem with locating it for most people, without disturbing the order.

YuriSizov:
> Icon: Maybe just go with a full puzzle piece instead? A second "sub icon" feels too tiny for those dimensions, even if you're on a High DPI screen.

It doesn't have to be well defined or "anatomically" correct. It just needs to be distinct enough. It's a pretty common practice in the UI, and I think a mockup in the proposal is plenty readable for that. We also shouldn't assume there are going to be many custom icons in this context. Because having too many would reduce their usefulness, would make them stand out less. So for a one-off the exact shape is not so important.

elfajriharsuda (approved):

miv391:
If addons folder is not wanted to be pinned on top and a icon looking more like the one in the proposal is wanted, then this PR has very little to offer. My graphics designer skills are not enough to produce a small nice looking icon.

MewPurPur:
`<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path d="M2 1a1 1 0 00-1 1v10a1 1 0 001 1h4.5v-4.5a2 2 0 012-2h4a2 2 0 012 2h.5V5a1 1 0 00-1-1h-4a1 1 0 01-1-1V2a1 1 0 00-1-1z" fill="#e0e0e0"/><path d="M8.5 8a.5.5 0 00-.5.5v4a.5.5 0 00.5.5h1.5v.25a1 1 0 101 0v-.25h1.5a.5.5 0 00.5-.5v-1.5h.25a1 1 0 100-1h-.25v-1.5a.5.5 0 00-.5-.5z" fill="#e0e0e0"/></svg>`

A bit of a touchy moment here: I used the e0e0e0 fill for both icons, but you need to consider:

- The remapping between light and dark theme in editor_themes.cpp
- The ability to color folders

I've not investigated any of them, I don't have time, I'm just giving a SVG that's similar to what was proposed, since it seems like you were stuck on this.

editor/filesystem_dock.h
```diff
@@ -244,7 +244,7 @@ class FileSystemDock : public VBoxContainer {
 	void _reselect_items_selected_on_drag_begin(bool reset = false);
 
 	Ref<Texture2D> _get_tree_item_icon(bool p_is_valid, String p_file_type);
-	bool _create_tree(TreeItem *p_parent, EditorFileSystemDirectory *p_dir, Vector<String> &uncollapsed_paths, bool p_select_in_favorites, bool p_unfold_path = false);
+	bool _create_tree(TreeItem *p_parent, EditorFileSystemDirectory *p_dir, Vector<String> &uncollapsed_paths, bool p_select_in_favorites, int dir_level, bool p_unfold_path = false);
```
AThousandShips:
```suggestion
	bool _create_tree(TreeItem *p_parent, EditorFileSystemDirectory *p_dir, Vector<String> &uncollapsed_paths, bool p_select_in_favorites, int p_dir_level, bool p_unfold_path = false);
```

Repiteo:
If you still wish to integrate this PR with MewPurPur's work, it'll need a rebase

editor/filesystem_dock.h
AThousandShips, line 247:
```suggestion
	bool _create_tree(TreeItem *p_parent, EditorFileSystemDirectory *p_dir, Vector<String> &uncollapsed_paths, bool p_select_in_favorites, int p_dir_level, bool p_unfold_path = false);
```

editor/filesystem_dock.cpp
```diff
@@ -179,7 +179,7 @@ Ref<Texture2D> FileSystemDock::_get_tree_item_icon(bool p_is_valid, String p_fil
 	return file_icon;
 }
 
-bool FileSystemDock::_create_tree(TreeItem *p_parent, EditorFileSystemDirectory *p_dir, Vector<String> &uncollapsed_paths, bool p_select_in_favorites, bool p_unfold_path) {
+bool FileSystemDock::_create_tree(TreeItem *p_parent, EditorFileSystemDirectory *p_dir, Vector<String> &uncollapsed_paths, bool p_select_in_favorites, int dir_level, bool p_unfold_path) {
```
AThousandShips, line 182:
```suggestion
bool FileSystemDock::_create_tree(TreeItem *p_parent, EditorFileSystemDirectory *p_dir, Vector<String> &uncollapsed_paths, bool p_select_in_favorites, int p_dir_level, bool p_unfold_path) {
```

editor/filesystem_dock.cpp
```diff
@@ -214,7 +214,7 @@ bool FileSystemDock::_create_tree(TreeItem *p_parent, EditorFileSystemDirectory
 
 	subdirectory_item->set_text(0, dname);
 	subdirectory_item->set_structured_text_bidi_override(0, TextServer::STRUCTURED_TEXT_FILE);
-	subdirectory_item->set_icon(0, get_editor_theme_icon(SNAME("Folder")));
+	subdirectory_item->set_icon(0, dir_level == 1 && dname == "addons" ? get_editor_theme_icon(SNAME("AddonsFolder")) : get_editor_theme_icon(SNAME("Folder")));
```
AThousandShips, line 217:
```suggestion
	subdirectory_item->set_icon(0, (p_dir_level == 1 && dname == "addons") ? get_editor_theme_icon(SNAME("AddonsFolder")) : get_editor_theme_icon(SNAME("Folder")));
```

AThousandShips (dismissed):

PR #85973, opened 2023-12-09
Allow ZIPPacker and ZIPReader to use passwords

MarioLiebisch:
These changes allow Godot to both read and write files inside ZIP archives that are password protected.

Randomly stumbled over [this discussion](https://forum.godotengine.org/t/is-there-an-option-to-use-a-zip-with-password/34590) and I think this addition makes sense.

While Godot provides encrypted files for secure file storage/exchange, password protected zip files are way more accessible (e.g. with non-custom tools), while still somewhat preventing trivial access to information.

### DIFF
--- a/modules/zip/doc_classes/ZIPPacker.xml
+++ b/modules/zip/doc_classes/ZIPPacker.xml
@@ -47,6 +47,7 @@
 		<method name="start_file">
 			<return type="int" enum="Error" />
 			<param index="0" name="path" type="String" />
+			<param index="1" name="password" type="String" default="&quot;&quot;" />
 			<description>
 				Starts writing to a file within the archive. Only one file can be written at the same time.
 				Must be called after [method open].
--- a/modules/zip/zip_packer.h
+++ b/modules/zip/zip_packer.h
@@ -41,9 +41,18 @@ class ZIPPacker : public RefCounted {
 
 	Ref<FileAccess> fa;
 	zipFile zf = nullptr;
+	Vector<uint8_t> write_buffer;
+	String current_password;
+	String current_path;
 
 protected:
 	static void _bind_methods();
+
+#ifndef DISABLE_DEPRECATED
+	Error _start_file_bind_compat_85973(const String &p_path);
+
+	static void _bind_compatibility_methods();
+#endif // DISABLE_DEPRECATED
 
 public:
 	enum ZipAppend {
@@ -55,7 +64,7 @@ public:
 	Error open(const String &p_path, ZipAppend p_append);
 	Error close();
 
-	Error start_file(const String &p_path);
+	Error start_file(const String &p_path, const String &p_password);
 	Error write_file(const Vector<uint8_t> &p_data);
 	Error close_file();
 
--- a/thirdparty/minizip/unzip.c
+++ b/thirdparty/minizip/unzip.c
@@ -68,10 +68,6 @@
 #include <stdlib.h>
 #include <string.h>
 
-#ifndef NOUNCRYPT
-        #define NOUNCRYPT
-#endif
-
 #include "zlib.h"
 #include "unzip.h"
 
--- a/misc/extension_api_validation/4.2-stable.expected
+++ b/misc/extension_api_validation/4.2-stable.expected
@@ -7,3 +7,10 @@ should instead be used to justify these changes and describe how users should wo
 Add new entries at the end of the file.
 
 ## Changes between 4.2-stable and 4.3-stable
+
+GH-85973
+--------
+Validate extension JSON: Error: Field 'classes/ZIPPacker/methods/start_file/arguments': size changed value in new API, from 1 to 2.
+Validate extension JSON: Error: Field 'classes/ZIPReader/methods/read_file/arguments': size changed value in new API, from 2 to 3.
+
+Added a new argument to each and provided compatibility methods as compatibility fallback.
--- a/misc/extension_api_validation/4.2-stable.expected
+++ b/misc/extension_api_validation/4.2-stable.expected
@@ -13,4 +13,4 @@ GH-85973
 Validate extension JSON: Error: Field 'classes/ZIPPacker/methods/start_file/arguments': size changed value in new API, from 1 to 2.
 Validate extension JSON: Error: Field 'classes/ZIPReader/methods/read_file/arguments': size changed value in new API, from 2 to 3.
 
-Added a new argument to each and provided compatibility methods as compatibility fallback.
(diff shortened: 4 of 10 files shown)


modules/zip/zip_packer.h
```diff
@@ -41,10 +41,19 @@ class ZIPPacker : public RefCounted {
...
 	zipFile zf = nullptr;
+	Vector<uint8_t> write_buffer;
+	String current_password;
+	String current_path;
 
 protected:
 	static void _bind_methods();
 
+#ifndef DISABLE_DEPRECATED
+	Error _start_file_compat_nopw(const String &p_path);
```
AThousandShips, line 52:
```suggestion
	Error _start_file_compat_85973(const String &p_path);
```
This is the established format

modules/zip/zip_reader.h
```diff
@@ -45,12 +45,18 @@ class ZIPReader : public RefCounted {
 protected:
 	static void _bind_methods();
 
+#ifndef DISABLE_DEPRECATED
+	PackedByteArray _read_file_compat_nopw(const String &p_path, bool p_case_sensitive);
```
AThousandShips, line 49:
```suggestion
	PackedByteArray _read_file_compat_85973(const String &p_path, bool p_case_sensitive);
```

MarioLiebisch:
Branch is updated/squashed, sorry for the build spam, totally forgot forks run actions, too.😉

modules/zip/zip_packer.h
akien-mga: Actually the established format seems to be `_start_file_bind_compat_85973`. We have a few `compat_#####` without `bind_` but it seems to be the exception (and an oversight).

thirdparty/minizip/unzip.c
```diff
@@ -68,10 +68,6 @@
 #include <stdlib.h>
 #include <string.h>
 
-#ifndef NOUNCRYPT
-        #define NOUNCRYPT
-#endif
-
```
MarioLiebisch, line 74: This change felt really strange: Why was `NOUNCRYPT` added(?), yet the matching `NOCRYPT` is nowhere to be found. Without this change it's possible to encrypt files, but not decrypt them. If there was a specific reason for the addition/removal, it should probably be commented or maybe the definition be added to the Scons files.

modules/zip/zip_packer.h
AThousandShips: You're correct! Wasn't sure (and main point was the number and not a generic description)

modules/zip/doc_classes/ZIPPacker.xml
```diff
@@ -47,6 +47,7 @@
 		<method name="start_file">
 			<return type="int" enum="Error" />
 			<param index="0" name="path" type="String" />
+			<param index="1" name="password" type="String" default="&quot;&quot;" />
```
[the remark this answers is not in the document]
MarioLiebisch: What do you have in mind? Considering a password is a password. I thought it should be obvious they're on a per file basis.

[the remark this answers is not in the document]
AThousandShips: I'm not sure, like what it requires and how it works, unsure exactly how the details on it works, if there's any specifics, like what happens if you don't use a password if required and vice versa, including if you write to a file (unsure if you can amend a file?)

It was mostly *if* there were any details worth mentioning 🙂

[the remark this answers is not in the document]
MarioLiebisch: As far as I'm aware, passworded files work just like any other files in an archive, except you can't write them part by part, but this difference is hidden/abstracted away by the changes. I don't know about the implementation details in minizip, but I think you'll explicitly have to close both the file and archive either way (password or not) to ensure everything is written properly.

MarioLiebisch:
Yep, I remember articles about that, although I don't think anyone considers ZIP encryption flawless or perfectly secure. Would probably be worth at least a footnote in the docs. One could obviously switch to minizip-ng, too, which includes the newer AES encryption. What do the others here think about that?

Calinou:
> Yep, I remember articles about that, although I don't think anyone considers ZIP encryption flawless or perfectly secure. Would probably be worth at least a footnote in the docs. One could obviously switch to minizip-ng, too, which includes the newer AES encryption. What do the others here think about that?

Supporting a broken compression algorithm only makes this PR useful in very specific scenarios, like downloading a passworded ZIP archive you can't recreate and redistribute yourself. This is pretty uncommon in my experience.

AThousandShips:
> Supporting a broken compression algorithm only makes this PR useful in very specific scenarios, like extracting a passworded ZIP archive you can't recreate and redistribute yourself.

This would IMO suggest that if we do support it it'd be read-only, partially to prevent using a broken one for writing

MBybee:
> > Supporting a broken compression algorithm only makes this PR useful in very specific scenarios, like extracting a passworded ZIP archive you can't recreate and redistribute yourself.
> 
> This would IMO suggest that if we do support it it'd be read-only, partially to prevent using a broken one for writing

As long as it can be created in the first place, this would probably meet the original criteria well enough, though minizip-ng sounds like a good idea.

SuborbitalPigeon:
Just to clarify, when I said that ZipCrypto is broken, I meant that the encryption is effectively useless, as it can be easily circumvented in the matter of seconds. There would be nothing wrong with the data.

MarioLiebisch:
Yep, but at least it would be something that should be mentioned.

orosbogdan:
> Just to clarify, when I said that ZipCrypto is broken, I meant that the encryption is effectively useless, as it can be easily circumvented in the matter of seconds. There would be nothing wrong with the data.

Does this still apply ? 

Is there any option to load assets at runtime from a pck encrypted with a custom encryption ? The default load methods use a path (string) instead of a Stream, so it's more limitaing in what you can do...

the-key-point:
> > Just to clarify, when I said that ZipCrypto is broken, I meant that the encryption is effectively useless, as it can be easily circumvented in the matter of seconds. There would be nothing wrong with the data.
> 
> Does this still apply ?
> 
> Is there any option to load assets at runtime from a pck encrypted with a custom encryption ? The default load methods use a path (string) instead of a Stream, so it's more limitaing in what you can do...
I couldnt find any option like this one. I was looking for that feature. Exactly you are right using always only path is limiting. If it could be something like getting encrypted data(scene,resource) to memory without using path(or like reading binary bytes etc.) and decrypt it in memory on-the-fly(maybe with simple commands in script) could be sufficient.

PR #85975, opened 2023-12-09
C#: Implement function to throw on null pointers

Repiteo:
Specifically: checks if a passed `IntPtr` is equal to `IntPtr.Zero`. Utilizes the same syntax as `ArgumentNullException.ThrowIfNull`; only the argument needs to be passed, with the argument name being automatically pulled via the `CallerArgumentExpression` attribute[^1]. In addition to implementing this on the generated NativeCalls files, minor changes were also made to `bindings_generator.cpp` in order to give the resulting file proper formatting.

[^1]: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callerargumentexpressionattribute

### DIFF
--- a/modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/ExceptionUtils.cs
+++ b/modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/ExceptionUtils.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 using System.Text;
 
 #nullable enable
@@ -238,6 +239,14 @@ namespace Godot.NativeInterop
             }
 
             return variant->Type.ToString();
+        }
+
+        internal static void ThrowIfNullPtr(IntPtr ptr, [CallerArgumentExpression(nameof(ptr))] string? paramName = null)
+        {
+            if (ptr == IntPtr.Zero)
+            {
+                throw new ArgumentNullException(paramName);
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/ExceptionUtils.cs
```diff
@@ -239,5 +240,11 @@ private unsafe static string GetVariantTypeName(godot_variant* variant)
 
             return variant->Type.ToString();
         }
+
+        internal static void ThrowIfNullPtr(IntPtr ptr, [CallerArgumentExpression("ptr")] string? paramName = null)
```
raulsntos, line 244:
```suggestion
        internal static void ThrowIfNullPtr(IntPtr ptr, [CallerArgumentExpression(nameof(ptr))] string? paramName = null)
```

modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/ExceptionUtils.cs
```diff
@@ -239,5 +240,11 @@ private unsafe static string GetVariantTypeName(godot_variant* variant)
...
+        {
+            if (ptr == IntPtr.Zero)
+                throw new ArgumentNullException(paramName);
```
raulsntos, line 247:
```suggestion
            if (ptr == IntPtr.Zero)
            {
                throw new ArgumentNullException(paramName);
            }
```

raulsntos (approved):
Thanks, a few minor nit-picks but looks good to me otherwise.

YuriSizov:
Thanks!

YuriSizov:
Builds for the .NET version on the build server fail with this error, which I believe is related to this PR:

```
modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/ExceptionUtils.cs(244,90): error CS0103: The name 'ptr' does not exist in the current context [modules/mono/glue/GodotSharp/GodotSharp/GodotSharp.csproj]
```

See also https://github.com/godotengine/godot/pull/82983.

PR #85977, opened 2023-12-09
Fix changing import type of nested resources

### LINKED ISSUE #85892: Changing the 'Import As' of a PNG file to 'AtlasTexture' not working cleanly post 4.2
### Tested versions

Reproducible in 4.2

### System information

Windows 10 - Godot 4.2

### Issue description

In 4.1, changing the 'Import as' of a Texture to an AtlasTexture, would restart the editor, and everything would work post restart. 

Now, doing this does not restart the editor, and the new atlas file does not show up in the filesystem until a forced restart. 

Post restart, the atlas file shows up, but any place the Texture was used pre change of 'Import as' gets corrupted, and you have to remake the animations. 

### Steps to reproduce

1) Import few png images into a fresh project
2) Create an AnimatedSprite node, and add these png images into an animation
3) Select all the png images, and change the 'Import As' to 'AtlasTexture', and follow the remaining steps to do so
4) The images will no longer be visible with the node, and post editor restart, the sprite frames resource will have no images in it. 

It will only reproduce once. Difficult to create an MRP.

KoBeWi:
Fixes #85892

...except it doesn't work for some reason. At first the fix failed when trying to find old resource (I had 2 resources that were the same, but had different object id), but after that error magically fixed itself, it now fails to replace the resource with the new one, because loading from the old path fails (without any error ;). I suspect atlas importer does something wrong with the resource. Maybe waiting one frame™ would help, but I'm unable to determine what exactly goes wrong here. There are also errors that the newly created atlas can't be found (which I think existed before).
Putting as draft until this is resolved .-.

On a side note, I had deja vu when writing this code. We do the same exact thing in at least one other place (I mean the `_check_property_for_resource()` method).

Here's minimal project for the issue, for easier testing:
[AtlasTest.zip](https://github.com/godotengine/godot/files/13625635/AtlasTest.zip)
Select both icon files and change import type to TextureAtlas, then save the scene.

### DIFF
--- a/editor/import_dock.cpp
+++ b/editor/import_dock.cpp
@@ -531,6 +531,7 @@ void ImportDock::_reimport_and_cleanup() {
 
 	EditorResourcePreview::get_singleton()->stop(); // Don't try to re-create previews after import.
 	_reimport();
+	EditorFileSystem::get_singleton()->scan(); // In case reimport created new resources.
 
 	if (need_cleanup.is_empty()) {
 		return;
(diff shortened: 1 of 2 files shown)


tokengamedev:
Request to have this fix for 4.2 also. In some cases, it crashed the editor, but it is hard to reproduce. It may be some corner case.

PR #85991, opened 2023-12-10
Fix `FileAccessPack::get_buffer` updating position past the length of file

### LINKED ISSUE #85921: FileAccessPack::get_buffer moves the cursor past the end of file
### Tested versions

4.2.stable (exported only)

### System information

Linux

### Issue description

This line is the bug:

https://github.com/godotengine/godot/blob/a311a4b162364d032b03ddf2a0e603ba40615ad7/core/io/file_access_pack.cpp#L330

`p_length` is the buffer length provided by the user, but this line does not check if the result is bigger than the length of the file.

As a result, when user tries to get a buffer that is bigger than the file (something like `file.get_buffer(99999999)`), `file.get_position()` may be bigger than `file.get_length()`.

It doesn't break Godot directly, but breaks any calculations based on `get_position()`. For example, it breaks relative seeking.

There is another variable `t_read` which is checked and should probably be used instead:

https://github.com/godotengine/godot/blob/a311a4b162364d032b03ddf2a0e603ba40615ad7/core/io/file_access_pack.cpp#L324-L328

### Steps to reproduce

Since this is a bug in `FileAccessPack`, it is only reproducible in **exported** projects (something like `FileAccessUnix` doesn't have this bug, so it can't be reproduced in the editor).

To reproduce, try to `file.get_buffer(99999999)` and then use `file.get_position()` for something.

I made a simple script that attempts to make a relative seeking and breaks because of this bug:

[FileAccessPackTest.zip](https://github.com/godotengine/godot/files/13615040/FileAccessPackTest.zip)

jsjtxietian:
Fixes #85921

### DIFF
--- a/core/io/file_access_pack.cpp
+++ b/core/io/file_access_pack.cpp
@@ -327,7 +327,7 @@ uint64_t FileAccessPack::get_buffer(uint8_t *p_dst, uint64_t p_length) const {
 		to_read = (int64_t)pf.size - (int64_t)pos;
 	}
 
-	pos += p_length;
+	pos += to_read;
 
 	if (to_read <= 0) {
 		return 0;


akien-mga (approved):

YuriSizov:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

PR #80452, opened 2023-08-09
Fix OGG audio loop offset pop

ghost:
This fixes #75762, fixes #59348 and parts of #64775

Fixed the mixing function so it doesn't skip the last ogg packet of the file, and also made it so the burning algorithm only burns the needed samples from the last packet, and not the whole packet.

There are still some issues, if you download the example project from the issue, and you try the audio-pop-no-intro.ogg file, you'll see that there's still a pop, this is because for some reason, the engine is missing 38 samples, and not at the end of the file like the error originally was.

Another problem that became aparent is the decimal precision of the loop offset variable, in a song of mine which has a loop point at 31.46, the loop offset becomes 31.459991 or something like that, which introduces a tiny pop (this does not happen in Godot 3). If I manually change that value to 31.4600010, the pop disappears and the song plays like it does in Godot 3. So that's another problem somewhere else.

THIS DOES NOT FIX THE WAV POP.

However, it seems to have fixed the mp3 pop I was getting, so I guess it could have been a simple floating point precision error, but more testing is probably needed.

I'll keep looking into it.

If you know why the 38 samples from the 0 seconds offset loop could be going missing please let me know.

Thanks.

PR #85996, opened 2023-12-10
Fix `OggPacketSequencePlayback::next_ogg_packet()` never returning false

nikitalita:
This fixes a bug introduced in #80452 (see https://github.com/godotengine/godot/pull/80452#pullrequestreview-1773999033). The check that was here caused this function to never return false.

I've tested this with the example projects provided in the issues that the PR purported to fix, and it causes no issues.

### DIFF
--- a/modules/ogg/ogg_packet_sequence.cpp
+++ b/modules/ogg/ogg_packet_sequence.cpp
@@ -159,9 +159,7 @@ bool OggPacketSequencePlayback::next_ogg_packet(ogg_packet **p_packet) const {
 
 	*p_packet = packet;
 
-	if (!packet->e_o_s) { // Added this so it doesn't try to go to the next packet if it's the last packet of the file.
-		packet_cursor++;
-	}
+	packet_cursor++;
 
 	return true;
 }


nikitalita:
To clarify the issue here:

Previous to #80452, this function acted as originally intended: It would keep serving up the next packet and returning `true` until it reached the end of the stream, when it would return `false`. It also did so silently, without reporting errors, indicating that this was intended to be called to read from the stream until it returned false.

That PR changed that behavior such that, when it reaches the end of the stream, it'll keep serving up the last packet and returning true forever; it won't **ever** return false. Besides being intuitive, this is just semantically incorrect behavior, as there's no point to have a boolean value as a return value if there's no `false` case (the user can't set the page to an incorrect value either, as `set_page_number` is bounds checked).

While none of its _current_ usages in Godot use this to detect the EOS, this is a pretty obvious footgun that can result in infinite loops and could easily become a hazard in the future.

Since none of the Godot usages were using this to detect the EOS (they checked the packet for the `e_o_s` flag), this change shouldn't affect ogg playback at all. Indeed, I have tested all the projects included with the issues that the previous PR purported to fix, and they all playback fine.

ellenhp:
@bs-mwoerner I saw you responded in the comment linked in the description here. Does this change seem like it could have unintended consequences? The rationale given seems sound I just want to check.

bs-mwoerner:
I don't see any problems with this change. When we call `next_ogg_packet` and it returns false, we consider this an abnormal situation and print a warning. That should still be valid, because we intend to stop reading after seeing an `e_o_s` packet, so if we ever get `false`, then something's indeed not quite right and the warning would be appreciated. I tried playing a file with and without looping and with this change in place and the false-paths were never taken. 👍

I say go ahead with the change. There's even still a comment in the header that says it returns "false on error or if there is no next packet".

ellenhp:
Thanks for weighing in!

ellenhp (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #59342, opened 2022-03-20
Fix `MODEL_NORMAL_MATRIX` in fragment function of spatial shader

Chaosus:

PR #86000, opened 2023-12-10
Add `model_normal_matrix` for fragment shader

### LINKED ISSUE #85586: MODEL_NORMAL_MATRIX documented for fragment shaders but not present in GLSL
### Godot version

4.2

### System information

Godot v4.2.stable.mono - Windows 10.0.22621 - Vulkan (Forward+) - dedicated AMD Radeon RX 6600 XT (Advanced Micro Devices, Inc.; 31.0.22011.4008) - AMD Ryzen 9 5900 12-Core Processor (24 Threads)

### Issue description

I wrote a 3D fragment shader using MODEL_NORMAL_MATRIX, which is documented as one of the builtins available in fragment shaders. 

(https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/spatial_shader.html#fragment-built-ins)

The shader does not show any error within godot, however when I tried to run the project, I see this:

```
E 0:00:01:0437   _compile_variant: Failed parse:
WARNING: 0:81: '' : all default precisions are highp; use precision statements to quiet warning, e.g.:
         "precision mediump int; precision highp float;" 
ERROR: 0:2342: 'model_normal_matrix' : undeclared identifier 
ERROR: 0:2342: '' : compilation terminated 
ERROR: 2 compilation errors.  No code generated.

  <C++ Source>   servers/rendering/renderer_rd/shader_rd.cpp:294 @ _compile_variant()

```

(As an aside, the value I actually wanted was MODELVIEW_NORMAL_MATRIX which is not documented to be supported in fragment shaders at all.)

### Steps to reproduce

Make a 3D fragment shader that uses MODEL_NORMAL_MATRIX and try to run the project that uses the shader. 

Actual: Godot gives the above error. Observe that the shader editor does not show any error. Also observe that the documentation says that this should work.

Expected: The feedback from the shader editor, the documentation and the actual behavior should all match.

[model-normal-matrix-demo-project.zip](https://github.com/godotengine/godot/files/13520619/model-normal-matrix-demo-project.zip)

jsjtxietian:
Fixes #85586

But I'm not sure whether this should be exposed to fragment shader so feel free to correct me, if not then we should update the doc instead.

### DIFF
--- a/drivers/gles3/shaders/scene.glsl
+++ b/drivers/gles3/shaders/scene.glsl
@@ -583,6 +583,8 @@ void main() {
 
 #define SHADER_IS_SRGB true
 
+#define FLAGS_NON_UNIFORM_SCALE (1 << 4)
+
 /* Varyings */
 
 #if defined(COLOR_USED)
@@ -955,6 +957,7 @@ ivec2 multiview_uv(ivec2 uv) {
 
 uniform highp mat4 world_transform;
 uniform mediump float opaque_prepass_threshold;
+uniform highp uint model_flags;
 
 #if defined(RENDER_MATERIAL)
 layout(location = 0) out vec4 albedo_output_buffer;
@@ -1520,6 +1523,13 @@ void main() {
 #ifdef LIGHT_VERTEX_USED
 	vec3 light_vertex = vertex;
 #endif //LIGHT_VERTEX_USED
+
+	highp mat3 model_normal_matrix;
+	if (bool(model_flags & uint(FLAGS_NON_UNIFORM_SCALE))) {
+		model_normal_matrix = transpose(inverse(mat3(model_matrix)));
+	} else {
+		model_normal_matrix = mat3(model_matrix);
+	}
 
 	{
 #CODE : FRAGMENT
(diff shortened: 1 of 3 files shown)


clayjohn:
Duplicate of https://github.com/godotengine/godot/pull/59342

An underlying issue comes from the fact that the ``WORLD_MATRIX`` can be altered by the user. If the user alters the`WORLD_MATRIX` in the vertex shader, then those edits won't carry through to the fragment shader using this approach.

clayjohn:
Looking at this again, I think this is the correct fix. `MODEL_NORMAL_MATRIX` is constant in the vertex shader which means the user should expect it to be the same in the vertex shader and the fragment shader.

jsjtxietian:
LOL, shall I reopen it ?

clayjohn:
> LOL, shall I reopen it ?

Yes please!

clayjohn (approved):
Thanks for re-opening this!

jsjtxietian:
Conflict resolved.

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.3.1.

PR #86001, opened 2023-12-10
[Windows] Always use ANGLE in ARM builds.

### LINKED ISSUE #85961: Editor rendering issue on Windows on ARM device since 4.2
### Tested versions

4.2.stable

### System information

Godot v4.2.stable - Windows 10.0.22621 - GLES3 (Compatibility) - D3D12 (Qualcomm(R) Adreno(TM) 685 GPU) () - Snapdragon (TM) 8cx Gen 2 @ 3.0 GHz (8 Threads)

### Issue description

well~ i have full interests in using godot engine and now i am a Chinese graduate which has more free time to learn it. but it seems that there is a bug when i use Godot 4.2 stable version - although the workspace might run well, i feel a bit confused when i **create a new project or add one node**, the places of settings will **show incorrectly** when i try to **maximize the setting window and restore it** (luckily, only show incorrectly, the settings can still run correctly).
i am using Windows 11 on ARM laptop to run Godot engine 3/4 - it can runs well in using OpenGL compatibility Pack in Microsoft Store (awesome!!! but of course, only GLES2/GLES3). but this bug **does not exist** in **Godot 4.1.3 stable version** or before (iMPORTANT...), and i have tried to restart many times, the bug didn't disappear. i will upload the main screenshot as well.
now i have rollbacked in Godot 4.1.3 stable. so i hope this bug can be solved in Godot 4.2.1 or as quickly as possible, or... i'd like to find everyone's help? (tears of joy~)

### Steps to reproduce

1. i try to create new project, or add one node.
2. i maximize the setting window.
3. i restore the setting window.
4. the places of settings show incorrectly.

it can appear when i try to create new project so there is no nessesary to upload a project.godot... maybe?

bruvzg:
There's no native OpenGL drivers on Windows for ARM (only software compatibility layer), so ANGLE is a more viable option for compatibility renderer.

Fixes https://github.com/godotengine/godot/issues/85961

### DIFF
--- a/platform/windows/display_server_windows.cpp
+++ b/platform/windows/display_server_windows.cpp
@@ -4647,6 +4647,12 @@ DisplayServerWindows::DisplayServerWindows(const String &p_rendering_driver, Win
 	// Init context and rendering device
 #if defined(GLES3_ENABLED)
 
+#if defined(__arm__) || defined(__aarch64__) || defined(_M_ARM) || defined(_M_ARM64)
+	// There's no native OpenGL drivers on Windows for ARM, switch to ANGLE over DX.
+	if (rendering_driver == "opengl3") {
+		rendering_driver = "opengl3_angle";
+	}
+#else
 	bool fallback = GLOBAL_GET("rendering/gl_compatibility/fallback_to_angle");
 	if (fallback && (rendering_driver == "opengl3")) {
 		Dictionary gl_info = detect_wgl();
@@ -4667,6 +4673,7 @@ DisplayServerWindows::DisplayServerWindows(const String &p_rendering_driver, Win
 			rendering_driver = "opengl3_angle";
 		}
 	}
+#endif
 
 	if (rendering_driver == "opengl3") {
 		gl_manager_native = memnew(GLManagerNative_Windows);


Lenbot-QC:
nice work! as an WOA user, i have already made a minimal project on a **_Windows on ARM laptop_**, then test in using **_opengl3 ANGLE_** and **_opengl API 4.1_**, and then... here are the results.
**_PS: [i use three GPUParticle2D controllers to run.](https://github.com/godotengine/godot/issues/86009) Running in vsync=off, screen_res=(2560,1600,fullscreen), max_fps here sets 0 (the project what i upload sets 60.)_**

_**opengl API 4.1**: (about **72**±6 fps, max_usage=**66%**, having a **opengl->d3d12** translation layer)_
Screenshot 2023-12-11 032254

_**opengl3 ANGLE**: (about **379**±13 fps, sometimes more than 400 fps, max_usage=**100%**)_
Screenshot 2023-12-11 031540

**_72 fps_ vs. _379 fps_**, oh my... that's an awesome enhancement!!
PS: i advise to turn off the vsync (might look unsmoothly when enabled) and set max fps as **_screen refresh rate_** (e.g. 59, 60, 144, etc.) when choose ANGLE, this test is just for fun.

of course, the more **single** CPU frequency, the "more" fps (vsync=off), so cpu within low base/turbo frequency is not friendly. **then** actually watching on the performance of GPU is also **worthy**. Mine is 3.0 GHz and finally goes about 2.8 GHz because of 100% GPU usage. (but that's still awesome!! thanks to this rendering method, WOA device can really create an amazing game in different genres as well.)

and of course i will only recommend on gpu graphic which **needs translation layer**, that's because...
eeeh... we can view these two pics (look at fps). here's running on a x86_64 laptop within ATI Radeon HD 7700M.

this is _**opengl3 ANGLE**_, the single CPU frequency goes maximum, the GPU-3D usage is less than 50%. con seems more than pro.
Screenshot 2023-12-11 004503

and this, is _**opengl API 4.1**_. (tears of joy)
Screenshot 2023-12-11 004655

so if have a gpu which has enough performance and needn't translation layer (more important), needn't to use ANGLE. :D

clayjohn (approved):
Makes sense!

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #84685, opened 2023-11-09
Rework input actions to be reliable

KoBeWi:
#80859 introduced surprisingly many regressions (I mean, who would've thought that press and release count for an action don't necessarily match...). This PR reworks the input actions again. This time `pressed` is a bitmask. When an event is pressed, it sets a bit in the `pressed` integer, release resets it. Pressed state is still determined by `pressed` being greater than 0, but since every event corresponds to a bit being either 0 or 1, there is no way to "overpress" or "overrelease" the event.

Action strengths broke again after the rework, so I reworked that too. Since I am using event index anyway, I decided to just keep a vector of all event strengths. Now the action strength is the max value of all event's strengths. It makes action strength more reliable than ever.

Some considerations:
- since the pressed state is stored using bitwise operations, the event count per action is now effectively limited to 32. We should probably enforce that somehow
- `Input.press_action()` will press the action, but then it can be released by any event. `Input.release_action()` releases all events. This is actually how it was before the previous rework

Fixes all input action regressions, probably.
I tested with some projects on Windows and it appears to be working correctly overall.

*Bugsquad edit:*
- Fixes #82262 (https://github.com/godotengine/godot/pull/84685#issuecomment-1805879525)
- Fixes #82732 (https://github.com/godotengine/godot/pull/84685#issuecomment-1805879525)

PR #86010, opened 2023-12-10
Add `event_index` to InputEventAction

### LINKED ISSUE #85984: Simulating keypresses 4.1->4.2 regression
### Tested versions

- Reproducable in 4.2, 4.2.1-rc1 
- Not reproducable in 4.1.3

### System information

macOS

### Issue description

`Input.parse_input_event` and/or `Input.flush_buffered_events` not working as expected (or at least as previous)

### Steps to reproduce

Create empty project, 2D scene, drag godot icon into scene to act as a Sprite2D node and add script:

```gdscript
extends Sprite2D

func _ready():
	setup_inputmap()
	press_left(true)

func _process(delta):
	var velocity = Vector2()

	if Input.is_action_pressed("left"):
		velocity.x -= 1

	velocity = velocity.normalized() * 200
	position += velocity * delta

func press_left(pressed: bool):
	var input_event = InputEventAction.new()
	input_event.action = "left"
	input_event.pressed = pressed
	Input.parse_input_event(input_event)
	Input.flush_buffered_events()

func setup_inputmap():
	var action_name = "left"
	InputMap.add_action(action_name)
	var event_key = InputEventKey.new()
	event_key.keycode = KEY_A
	InputMap.action_add_event(action_name, event_key)

```

In 4.1 the sprite will move left
In 4.2 the sprite will not move

(attached is same as described in steps to reproduce)
[playground2.zip](https://github.com/godotengine/godot/files/13626105/playground2.zip)

KoBeWi:
Fixes #85984

Since #84685 actions rely on event id to press/release action. However InputEventAction does not correspond to any event, so it needed special handling. We could hard-code an event ID, but I thought it would be more useful to allow specifying which device the virtual event corresponds to.

### DIFF
--- a/core/input/input.cpp
+++ b/core/input/input.cpp
@@ -1244,7 +1244,7 @@ void Input::_update_action_cache(const StringName &p_action_name, ActionState &r
 	r_action_state.cache.strength = 0.0;
 	r_action_state.cache.raw_strength = 0.0;
 
-	int max_event = InputMap::get_singleton()->action_get_events(p_action_name)->size();
+	int max_event = InputMap::get_singleton()->action_get_events(p_action_name)->size() + 1; // +1 comes from InputEventAction.
 	for (const KeyValue<int, ActionState::DeviceState> &kv : r_action_state.device_states) {
 		const ActionState::DeviceState &device_state = kv.value;
 		for (int i = 0; i < max_event; i++) {
--- a/core/input/input_event.cpp
+++ b/core/input/input_event.cpp
@@ -1585,6 +1585,14 @@ float InputEventAction::get_strength() const {
 	return strength;
 }
 
+void InputEventAction::set_event_index(int p_index) {
+	event_index = p_index;
+}
+
+int InputEventAction::get_event_index() const {
+	return event_index;
+}
+
 bool InputEventAction::is_match(const Ref<InputEvent> &p_event, bool p_exact_match) const {
 	if (p_event.is_null()) {
 		return false;
@@ -1649,9 +1657,13 @@ void InputEventAction::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_strength", "strength"), &InputEventAction::set_strength);
 	ClassDB::bind_method(D_METHOD("get_strength"), &InputEventAction::get_strength);
 
+	ClassDB::bind_method(D_METHOD("set_event_index", "index"), &InputEventAction::set_event_index);
+	ClassDB::bind_method(D_METHOD("get_event_index"), &InputEventAction::get_event_index);
+
 	ADD_PROPERTY(PropertyInfo(Variant::STRING_NAME, "action"), "set_action", "get_action");
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "pressed"), "set_pressed", "is_pressed");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "strength", PROPERTY_HINT_RANGE, "0,1,0.01"), "set_strength", "get_strength");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "event_index", PROPERTY_HINT_RANGE, "-1,31,1"), "set_event_index", "get_event_index"); // The max value equals to Input::MAX_EVENT - 1.
 }
 
 ///////////////////////////////////
--- a/core/input/input_event.h
+++ b/core/input/input_event.h
@@ -453,6 +453,7 @@ class InputEventAction : public InputEvent {
 
 	StringName action;
 	float strength = 1.0f;
+	int event_index = -1;
 
 protected:
 	static void _bind_methods();
@@ -465,6 +466,9 @@ public:
 
 	void set_strength(float p_strength);
 	float get_strength() const;
+
+	void set_event_index(int p_index);
+	int get_event_index() const;
 
 	virtual bool is_action(const StringName &p_action) const;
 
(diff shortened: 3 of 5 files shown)


core/input/input.cpp
```diff
@@ -718,6 +718,7 @@ void Input::_parse_input_event_impl(const Ref<InputEvent> &p_event, bool p_is_em
 		}
 		ERR_FAIL_COND_MSG(event_index >= (int)MAX_EVENT, vformat("Input singleton does not support more than %d events assigned to an action.", MAX_EVENT));
 
+		print_line(p_event);
```
KoBeWi, line 721: It does not make any sense, but this line is required for some reason:

https://github.com/godotengine/godot/assets/2223172/c7f0b130-b95e-4892-9692-ecc8adcf9d69

I tested it multiple times, always getting the same result. This is the MRP from the issue.

core/input/input_event.h
```diff
@@ -445,6 +446,9 @@ class InputEventAction : public InputEvent {
 	void set_strength(float p_strength);
 	float get_strength() const;
 
+	void set_event_id(float p_id);
```
AThousandShips, line 449:
```suggestion
	void set_event_id(int p_id);
```
I assume from the property

core/input/input.cpp
Sauermann: I can confirm, that for me on Linux, with the print_line the sprite moves and without the print_line the sprite doesn't move.

kitbdev: I think its because in `InputMap::event_get_action_status` pressed_state isn't set, so `InputEvent::is_action_pressed` uses unset garbage memory, which was affected by the print line.

KoBeWi: Ok fixed. Apparently InputEventAction was already handled below my code, so I just integrated the event index into it.

core/input/input_event.cpp
```diff
@@ -1549,9 +1557,13 @@ void InputEventAction::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_strength", "strength"), &InputEventAction::set_strength);
 	ClassDB::bind_method(D_METHOD("get_strength"), &InputEventAction::get_strength);
 
+	ClassDB::bind_method(D_METHOD("set_event_id", "id"), &InputEventAction::set_event_id);
+	ClassDB::bind_method(D_METHOD("get_event_id"), &InputEventAction::get_event_id);
+
 	ADD_PROPERTY(PropertyInfo(Variant::STRING_NAME, "action"), "set_action", "get_action");
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "pressed"), "set_pressed", "is_pressed");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "strength", PROPERTY_HINT_RANGE, "0,1,0.01"), "set_strength", "get_strength");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "event_id", PROPERTY_HINT_RANGE, vformat("-1,%d,1", int(Input::MAX_EVENT) - 1)), "set_event_id", "get_event_id");
```
akien-mga, line 1566: Missing include for `Input` apparently, but you probably can't include it as that would be a circular dependency, since `input.h` already includes `input_event.h`. Might be easiest to hardcode.

doc/classes/InputEventAction.xml
```diff
@@ -16,6 +16,9 @@
 		<member name="action" type="StringName" setter="set_action" getter="get_action" default="&amp;&quot;&quot;">
 			The action's name. Actions are accessed via this [String].
 		</member>
+		<member name="event_id" type="int" setter="set_event_id" getter="get_event_id" default="-1">
+			The real event index in action this event corresponds to (from events defined for this action in the [InputMap]). If [code]-1[/code], a unique ID will be used and actions pressed with this ID will need to be released with another [InputEventAction].
```
AThousandShips, line 20: Does the ID matter for any checks like pressing/releasing tracking, and if so does it have to match in some special way? If so details on possible issues with using `-1` should be detailed

AThousandShips: But if you then pass another event manually with the same ID, is it released?

KoBeWi: It depends if the event has `pressed` enabled. Multiple pressed events will not release action, you need a release event.

AThousandShips: So if you pass a release event? I said "passed" not "pressed", assumed it to be `pressed == false` as normal conditions, the difference being the id

KoBeWi: Action pressed state is tracked by event index. If the action is pressed, it needs a release event with the same ID as the pressed event to get released. Action remains pressed as long as at least one ID/index was not released.

AThousandShips: I think the wording then needs to be clarified as from the current wording it sounds like it's sufficient to pass another event with `-1`, which I assume won't be the case here as it's based on the size?

KoBeWi: Well yeah, if between press and release you add a new event in the InputMap, the action won't be released using `-1`. But it's such an unrealistic corner case that I'm not sure if it's even worth mentioning 🤔

AThousandShips: Well the description says "need to be released with another [InputEventAction]." in the same sentence as talking about the id, so I think it'll cause confusion, but idk

akien-mga:
Could use a rebase for good measure.

If/when merged, how confident are we in cherry-picking this for 4.2.3 to fix the regression? Or should we keep it for 4.3 only?

groud (approved):

akien-mga:
Thanks!

PR #50504, opened 2021-07-15
[3.x] Implement constructors and operator= for std::initializer_list<T> in …

MarianoGnu:
…one dimensional data containers. As a side effect append_array() also works with initializer_list

**TODO:** reimplement for master branch

**Test case:**
Paste the following snippet on any runnable piece of code and put a breakpoint to follow the execution
```
	Array test = {1, 2, "String"};
	Vector<int> v1 = {1, 2, 3, 4, 5};
	Vector<int> v2({1, 2, 3, 4, 5});
	v1.append_array({6,7,8});
	LocalVector<int, uint32_t, true> v3 = {7,8,9,10,11};
	LocalVector<int, uint32_t, true> v4({12,13,14,15});
	LocalVector<int, uint32_t, false> v5 = {7,8,9,10,11};
	LocalVector<int, uint32_t, false> v6({12,13,14,15});
	PoolStringArray v7 = {"AB", "CD", "EF"};
	PoolStringArray v8({"AB", "CD", "EF"});
	v8.append_array({"GH", "IJ"});
	PoolVector2Array v9 = {Vector2(0, 1), Vector2(2, 3)};
	v9.append_array({Vector2(0, 1), Vector2(2, 3)});
	Variant varArray = {1, 2, "String", false, this, {1, 2}};
```

PR #50493, opened 2021-07-15
Add list initialization support for Vector & LocalVector

timothyqiu:
Adds list initialzation support for `LocalVector` in order to get rid of `std::vector` in `VulkanContext`.

* `std::initializer_list` is used because it's the required parameter type when doing list initialization.
* ranged-for loop is used because I think it's allowed since #50284 was merged and simplifies code.

Also added list initialzation support for `Vector`. No code were harmed during production, only added a unit test.

PR #86015, opened 2023-12-10
Add list initialization to Array, Variant, and TypedArray

### LINKED ISSUE #1357: Implement `std::initializer_list<T>` assignment for Vector<T> and other container types
The engine itself.

Not being able to do
```cpp
Vector<String> _colors =  {...};
```

Implement `std::initializer_list` for `Vector<T>` (and possibly other container types).

https://en.cppreference.com/w/cpp/utility/initializer_list

Multiple calls of `insert`, or `append` - `append_all` only works for other types of Godot containers.
None of these work outside a function.
One could use `std::vector` or other containers but these are not trivially convertible to Godot containers.

This is literally a proposal for modifying core collection types.

kitbdev:
- closes https://github.com/godotengine/godot-proposals/issues/1357 (unless any other types need it)
- salvages #50504
- related #50493

Arrays use Vector list initialization to set its internal Vector.
Variant becomes an array type.
TypedArrays<> must have a type to use the list initialization.

Added initializer list tests for Array, TypedArray<int>, Variant, and PackedInt32Array.
Packed Arrays are typedefs for Vector, so they had this functionality before.

I didn't change any code to use initializer lists in this PR.
In another PR, I think `build_array()` can be replaced with initializer lists in a few test files.

### DIFF
--- a/core/variant/typed_array.h
+++ b/core/variant/typed_array.h
@@ -77,6 +77,9 @@ struct VariantInternalAccessor<const TypedArray<T> &> {
 			ERR_FAIL_COND_MSG(!is_same_typed(p_array), "Cannot assign an array with a different element type."); \
 			_ref(p_array);                                                                                       \
 		}                                                                                                        \
+		_FORCE_INLINE_ TypedArray(std::initializer_list<Variant> p_init) :                                       \
+				Array(Array(p_init), m_variant_type, StringName(), Variant()) {                                  \
+		}                                                                                                        \
 		_FORCE_INLINE_ TypedArray(const Variant &p_variant) :                                                    \
 				Array(Array(p_variant), m_variant_type, StringName(), Variant()) {                               \
 		}                                                                                                        \
--- a/core/variant/array.cpp
+++ b/core/variant/array.cpp
@@ -830,6 +830,12 @@ Array::Array(const Array &p_from) {
 	_ref(p_from);
 }
 
+Array::Array(std::initializer_list<Variant> p_init) {
+	_p = memnew(ArrayPrivate);
+	_p->refcount.init();
+	_p->array = Vector<Variant>(p_init);
+}
+
 Array::Array() {
 	_p = memnew(ArrayPrivate);
 	_p->refcount.init();
--- a/core/variant/array.h
+++ b/core/variant/array.h
@@ -34,6 +34,7 @@
 #include "core/typedefs.h"
 
 #include <climits>
+#include <initializer_list>
 
 class Variant;
 class ArrayPrivate;
@@ -195,6 +196,7 @@ public:
 
 	Array(const Array &p_base, uint32_t p_type, const StringName &p_class_name, const Variant &p_script);
 	Array(const Array &p_from);
+	Array(std::initializer_list<Variant> p_init);
 	Array();
 	~Array();
 };
--- a/core/variant/variant.cpp
+++ b/core/variant/variant.cpp
@@ -2585,6 +2585,11 @@ Variant::Variant(const Dictionary &p_dictionary) :
 	memnew_placement(_data._mem, Dictionary(p_dictionary));
 }
 
+Variant::Variant(std::initializer_list<Variant> p_init) :
+		type(ARRAY) {
+	memnew_placement(_data._mem, Array(p_init));
+}
+
 Variant::Variant(const Array &p_array) :
 		type(ARRAY) {
 	memnew_placement(_data._mem, Array(p_array));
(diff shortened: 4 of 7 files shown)


nlupugla:
Thanks for working on this @kitbev. I think this would make working with Arrays in the engine code much nicer. My only comment so far is whether we would rather have the argument be an initializer list or a `Vector<Variant>`. Note that having both would be an issue I think because then an expression like `Array({1, 2})` becomes ambiguous about whether it invokes the initializer list constructor or the vector constructor. I would argue in favor of the `Vector<Variant>` route (unless there are any performance reasons I'm not considering) because there are other places in the codebase where it would be useful to directly convert a vector to an array.

RandomShaper:
> [...] an expression like `Array({1, 2})` becomes ambiguous about whether it invokes the initializer list constructor or the vector constructor [...]
Would making the `Vector` constructor `explicit` avoid it being a candidate in such an expression?

Aside, I like this kind of changes to take Godot's C++ to the 21st century. I'm not sure about adding this kind of constructor to `Variant`. What are your thoughts in keeping initializing `Array`s in `Variant` more explicit (`Variant v = Variant(Array({...});`)?

kitbdev:
Rebased to fix merge conflict.

> vector constructor

Variant already has constructors for Packed*Array and a few other Vectors including `Vector<Variant>`, so array doesn't need its own constructor for it.

>  What are your thoughts in keeping initializing Arrays in Variant more explicit

Personally, I don't think it is needed and I like the ability to nest initializer lists for array and variant without it. The variant initializer list was made just for this purpose.
If we decide it would be better, I could remove it and make it explicit.

```gdscript 
Array nested = { { 0, 1 }, 'a' };
Array nested2 = { Array({ 0, 1 }), 'a' };
```

RandomShaper:
I'm not sure about the status of this conern:
> [...] an expression like `Array({1, 2})` becomes ambiguous about whether it invokes the initializer list constructor or the vector constructor [...]

nlupugla:
Personally, I'd prefer the explicitness, but it might be an odd outlier if that's already how it works for some of the other types.

kitbdev:
>I'm not sure about the status of this conern:
    [...] an expression like Array({1, 2}) becomes ambiguous about whether it invokes the initializer list constructor or the vector constructor [...]

Array doesn't have a vector constructor, so there is no issue. And I'm pretty sure it would use the initializer list in that case if it did.

As for Variant, where there is an initializer list constructor and vector constructors (including Packed*Arrays since they're vector typedefs):
```cpp
Variant({ 0.0, 1.0 }); // Variant uses initializer_list constructor.
Variant(Vector({ 0.0, 1.0 })); // Variant uses PackedFloat64Array constructor.

Variant({ 0.0, "b" }); // Variant uses initializer_list constructor.
Variant(Vector<Variant>({ 0.0, "b" })); // Variant uses Vector<Variant> constructor.
Variant(Array({ 0.0, 1.0 })); // Variant uses Array constructor.
Variant(PackedFloat32Array({ 0.0, 1.0 })); // Variant uses PackedFloat32Array constructor.
```

The first two are actually not the same since they have different Variant types, and they won't be equal. The first is type Array and the second is type PackedFloat64Array.
So maybe it should be explicit? Or some more logic should be done in the Variant initializer list constructor to convert to a packed type? Or maybe it is fine as is?

tomfull123 (approved):
Looks good to me!

akien-mga (approved):

KoBeWi:
> I didn't change any code to use initializer lists in this PR.

Probably time to rebase #89782 after this is merged.

core/variant/typed_array.h
```diff
@@ -77,6 +77,9 @@ struct VariantInternalAccessor<const TypedArray<T> &> {
 			ERR_FAIL_COND_MSG(!is_same_typed(p_array), "Cannot assign an array with a different element type."); \
 			_ref(p_array);                                                                                       \
 		}                                                                                                        \
+		_FORCE_INLINE_ TypedArray(std::initializer_list<Variant> p_init) :                                       \
+				Array(Array(p_init), m_variant_type, StringName(), Variant()) {                                  \
```
Ivorforce, line 81: Are you sure this is typesafe? I think this will allow instantiation like:
```c++
TypedArray<int> array = { "string" };
```
Because the array `initializer_list` constructor doesn't do type checks.

Ivorforce (changes requested):

Ivorforce (approved):

Repiteo:
Thanks!

PR #86017, opened 2023-12-11
`OS_Windows::execute` - Expand variables in arguments

drewler:
When attempting to Export to a Linux target from a Windows host, I tried to use `%LOCALAPPDATA%` from "Extra Args SSH", so I could specify some paths in a shorter way.

This didn't work, since `ssh` and its arguments are passed to `OS_Windows::execute` and used for [`CreateProcessW` ](https://github.com/godotengine/godot/blob/5d924373effbca5a96b458c737ec39358ffe481f/platform/windows/os_windows.cpp#L774), so it's run directly, without `Cmd.exe` being involved. Because of this, the variables are not expanded.

I think it can be useful to expand variables in arguments as it would also make it easier to eg. reuse the same extra args for multiple machines.

### DIFF
--- a/platform/windows/os_windows.cpp
+++ b/platform/windows/os_windows.cpp
@@ -729,7 +729,14 @@ Error OS_Windows::execute(const String &p_path, const List<String> &p_arguments,
 	String path = p_path.replace("/", "\\");
 	String command = _quote_command_line_argument(path);
 	for (const String &E : p_arguments) {
-		command += " " + _quote_command_line_argument(E);
+		String argument = E;
+		DWORD bufferSize = ExpandEnvironmentStringsW((LPWSTR)argument.utf16().ptrw(), NULL, 0);
+		if (bufferSize > 0) {
+			std::wstring expanded(bufferSize, L'\0');
+			ExpandEnvironmentStringsW((LPWSTR)argument.utf16().ptrw(), &expanded[0], bufferSize);
+			argument = String(expanded.c_str());
+		}
+		command += " " + _quote_command_line_argument(argument);
 	}
 
 	ProcessInfo pi;
--- a/platform/windows/os_windows.cpp
+++ b/platform/windows/os_windows.cpp
@@ -730,11 +730,12 @@ Error OS_Windows::execute(const String &p_path, const List<String> &p_arguments,
 	String command = _quote_command_line_argument(path);
 	for (const String &E : p_arguments) {
 		String argument = E;
-		DWORD bufferSize = ExpandEnvironmentStringsW((LPWSTR)argument.utf16().ptrw(), NULL, 0);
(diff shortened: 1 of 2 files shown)


platform/windows/os_windows.cpp
```diff
@@ -729,7 +729,14 @@ Error OS_Windows::execute(const String &p_path, const List<String> &p_arguments,
...
 	String command = _quote_command_line_argument(path);
 	for (const String &E : p_arguments) {
-		command += " " + _quote_command_line_argument(E);
+		String argument = E;
+		DWORD bufferSize = ExpandEnvironmentStringsW((LPWSTR)argument.utf16().ptrw(), NULL, 0);
+		if (bufferSize > 0) {
+			std::wstring expanded(bufferSize, L'\0');
+			ExpandEnvironmentStringsW((LPWSTR)argument.utf16().ptrw(), &expanded[0], bufferSize);
+			argument = String(expanded.c_str());
+		}
```
bruvzg, line 738:
```suggestion
		DWORD bufferSize = ExpandEnvironmentStringsW((LPCWSTR)argument.utf16().ptr(), NULL, 0);
		if (bufferSize > 0) {
			Char16String expanded;
			expanded.resize(bufferSize + 1);
			ExpandEnvironmentStringsW((LPCWSTR)argument.utf16().ptr(), (LPWSTR)expanded.ptrw(), bufferSize);
			argument = String::utf16(expanded.get_data());
		}
```

We do not use `std::string`, also it better to explicitly convert it back from UTF-16.

bruvzg (commented):
What's the current behavior of execute on Linux/macOS? It might require the similar change.

platform/windows/os_windows.cpp
AThousandShips:
```suggestion
		DWORD buffer_size = ExpandEnvironmentStringsW((LPCWSTR)argument.utf16().ptr(), NULL, 0);
		if (buffer_size > 0) {
			Char16String expanded;
			expanded.resize(buffer_size + 1);
			ExpandEnvironmentStringsW((LPCWSTR)argument.utf16().ptr(), (LPWSTR)expanded.ptrw(), buffer_size);
			argument = String::utf16(expanded.get_data());
		}
```
Use `camel_case` as well

AThousandShips: My bad confused them here lol

drewler:
> What's the current behavior of execute on Linux/macOS? It might require the similar change.

That's true. `OS_Unix::execute` uses `popen` and that won't expand environment variables either.

In that case, maybe we should have a generic "expand environment variables" method (`expand_envvars`?) that has specific implementations for `OS_Windows` and `OS_Unix`? Should that be implemented in this PR or is it fine to do it in a followup?

PR #77496, opened 2023-05-26
Implement 3D shadows in the GL Compatibility renderer

clayjohn:
Fixes: https://github.com/godotengine/godot/issues/66596

Implementation of 3D shadows for the Compatibility renderer

Implementation based on https://github.com/godotengine/godot-proposals/issues/3959

### Main approach

Particularly this uses a multipass approach to lighting so that we can benefit from prefetch when using hard shadows. In practice this means:
1. All lights without shadows are rendered during base pass as well as one light with shadows
2. Each additional light with a shadow requires drawing the object again with additive blending
3. Lights with shadows are blended in sRGB space instead of linear, so the colors blow out very quickly (i.e. lights with shadows look like they do in the Godot 3.x GLES2 renderer) (Lights without shadows look normal)

### Comments

There is still one outstanding bug where shadow atlas textures are getting out of date with multiple shadowed lights in the scene.

Pictures:
<details>
  <summary>Linear vs sRGB blending</summary>

_Lighting in linear space (no shadows)_
Screenshot from 2023-05-25 20-49-27

_Lighting in sRGB space_
Screenshot from 2023-05-25 20-51-26

_Lighting in sRGB space with reduced brightness and attenuation_
Screenshot from 2023-05-25 20-51-08

</details>

<details>
  <summary>Compatibility vs Forward+</summary>

_Compatibility_
Screenshot from 2023-09-27 16-59-51

_Forward+_
Screenshot from 2023-09-27 16-25-52

</details>

PR #86018, opened 2023-12-11
Fix radiance for sky in GLES stereo rendering

### LINKED ISSUE #84598: OpenGL: Sky contribution fails in VR
### Godot version

4.2.dev 5ee983188de97ae027f9b9c1443438063f708a7e

### System information

Windows 10 - Compatibility renderer 

### Issue description

When running a VR project sky contribution is not working in the compatibility renderer:

image

It is working fine in normal mode. Can be worked around by setting sky contribution to 0 and providing an ambient color.

### Steps to reproduce

Try any project with a proper sky setup in XR, just switch to compatibility renderer.

I'll try and create an MVP soon that can be used without XR hardware, but for now this can be used:
https://github.com/BastiaanOlij/godot_openxr_demo

BastiaanOlij:
This took me waaaaaaaaaaaaaaaaaaaaaaay too long to figure out.

In OpenGL everything is output upside down compared to Vulkan, that means that our culling order is reversed, see 
```
	if (!flip_y) {
		// If we're rendering right-side up, then we need to change the winding order.
		glFrontFace(GL_CW);
	}
```
in `RasterizerSceneGLES3::render_scene`.

However, when we output to OpenXR (or webXR), the compositor expects us to deliver images with the same orientation and thus we change things up and render things "upside down", and only flip the image on output to screen. So we keep to the same culling order.

The problem then is that our sky radiance map update code runs after we do this optional flip. This logic renders a full screen triangle to perform a full screen effect. The order of the vertices of this triangle is defined on the assumption we are doing our flip.

In normal mode, we render our triangle and thus our radiance map is rendered.
In XR mode we cull our triangle and thus our radiance map remains black.

The simple answer is to disable culling as we don't care for culling in this case.

This PR cleans up a few related issues and code that assumes state is retained when it's not.

Fixes #84598

### DIFF
--- a/drivers/gles3/rasterizer_scene_gles3.cpp
+++ b/drivers/gles3/rasterizer_scene_gles3.cpp
@@ -976,10 +976,10 @@ void RasterizerSceneGLES3::_update_sky_radiance(RID p_env, const Projection &p_p
 		glDisable(GL_BLEND);
 		glDepthMask(GL_FALSE);
 		glDisable(GL_DEPTH_TEST);
+		scene_state.current_depth_test = GLES3::SceneShaderData::DEPTH_TEST_DISABLED;
 		glDisable(GL_SCISSOR_TEST);
-		glCullFace(GL_BACK);
-		glEnable(GL_CULL_FACE);
-		scene_state.cull_mode = GLES3::SceneShaderData::CULL_BACK;
+		glDisable(GL_CULL_FACE);
+		scene_state.cull_mode = GLES3::SceneShaderData::CULL_DISABLED;
 
 		for (int i = 0; i < 6; i++) {
 			Basis local_view = Basis::looking_at(view_normals[i], view_up[i]);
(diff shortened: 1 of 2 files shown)


decacis:
It's working fine now!

| Master | This PR |
|--------|---------|
|Killifish-Ripe-Composed|Gossamerwingedbutterfly-Unwilling-Distant|

dsnopek (approved):
Great catch! The code changes look good to me, but probably should be reviewed by someone who knows rendering a little better than me :-)

drivers/gles3/rasterizer_scene_gles3.cpp
```diff
@@ -1054,11 +1062,13 @@ void RasterizerSceneGLES3::_filter_sky_radiance(Sky *p_sky, int p_base_layer) {
 
 	CubemapFilterShaderGLES3::ShaderVariant mode = CubemapFilterShaderGLES3::MODE_DEFAULT;
 
-	if (p_base_layer == 0) {
+#ifdef GL_API_ENABLED
+	if (RasterizerGLES3::is_gles_over_gl() && p_base_layer == 0) {
```
clayjohn, line 1066: I don't think this is correct. We need the mipmaps generated because we want to use the mipmapped version of the base layer for calculating the radiance map. This is the "pre-filtering" part of pre-filtered importance sampling

BastiaanOlij: Ah my bad, we do this check in `_update_dirty_skys` probably because `glTexImage2D` isn't supported under GLES. It is interesting here that for GLES `glGenerateMipmap` is also skipped (that might be a bug in those places as it won't allocate space for the mipmaps)

I made the same change here to test if it would fix the issue and fall back on the old approach of filtering, I didn't change it back.

Would be good to have a discussion on why this is working the way it does in `_update_dirty_skys`

BastiaanOlij: Definitely works fine on my Quest with this returned the way it was.

clayjohn (approved):
Looks great!

YuriSizov:
Cherry-picked for 4.2.2.

YuriSizov:
Can't be cherry-picked for 4.1 because it depends on changes from https://github.com/godotengine/godot/pull/77496.

Issue #83990, opened 2023-10-26
Volumetric fog stops taking lights into account when changing scene under specific circumstances


ArtyIF:
### Godot version

v4.2.beta2.mono.official [f8818f85e]

### System information

Godot v4.2.beta2.mono - Fedora Linux 39 (Workstation Edition) - Wayland - Vulkan (Forward+) - integrated Intel(R) Graphics (ADL GT2) () - 12th Gen Intel(R) Core(TM) i3-1215U (8 Threads)

### Issue description

When changing from a scene with a WorldEnvironment with volumetric fog enabled to a scene with the same WorldEnvironment (not sure if it should actually be the same), volumetric fog stops taking lights (at least `DirectionalLight3D`s) into account. This is fixed if the first scene has a `DirectionalLight3D` with shadows on.

Expected:

![image](https://github.com/godotengine/godot/assets/8017858/ee9406ca-6f1f-4a9b-99fb-cdf1050a3f0b)

Result:

![image](https://github.com/godotengine/godot/assets/8017858/0e362442-750c-4d4b-8ed8-1507155a0393)

Might be related to #83685

### Steps to reproduce

1. Open the MRP
2. Run it normally
3. Hit the Load Scene button
4. See that the fog doesn't account for the directional light
5. Open `scene1b.tscn` and run it
6. Hit the Load Scene button
7. See that the fog works as expected

### Minimal reproduction project

[VolumetricFogBugMRP.zip](https://github.com/godotengine/godot/files/13175669/VolumetricFogBugMRP.zip)

interpol-kun:
I also ran into similar problem: volumetric fog doesn't take in account the lights which were disabled at the start of the scene, but were enabled later by game logic.

Calinou:
Thanks for the report! Consolidating in https://github.com/godotengine/godot/issues/83685, as this is due to the same cause. (The issue goes away in the MRP if you resize the window or change the 3D resolution scale factor, which causes buffers to be recreated.)

PR #77703, opened 2023-05-31
Fix management of life cycle of volumetric fog related uniform sets

RandomShaper:
Fixes #77515.

PR #76550, opened 2023-04-28
Fix voxel GI issues (2)

RandomShaper:
Follow up of #76437, in the spirit of what I suggested here: https://github.com/godotengine/godot/pull/76437#discussion_r1177534748.

Fixes additional manifestations of #76243 when other effects are toggled (I've experienced, and confirmed no longer an issue with this PR, with SSAO). It may also fix other still not caught bugs.

PR #76437, opened 2023-04-25
Fix voxel GI issues

RandomShaper:
**UPDATE:** To avoid multiple 'update' notices throughout, just know that the full text has been revised after enhancing the fixes.

---

Changes kept in two commits because, even if related, they address different issues:

- Commit 1: _Fix breakages of volumetric fog on voxel GI changes_

   Fixes the book-keeping of related uniforms. Without this change, some of them could fail to be created when needed and some others could be pointing to outdated voxel GI buffers. For the uniform sets related to GI instances, a minor sets group idiom is introduced, that helps tracking validity and saves some checks, while also assisting developers in detecting bugs.

   Fixes #69565.
   Fixes #76243.

- Commit 2: _Fix unsupported sampler filter used for voxel GI_

   Fixes a validation error* about an R8G8 texture being sampled linearly when the driver reports no support for that.
   
*:
```
drivers\vulkan\vulkan_context.cpp:267 - VALIDATION - Message Id Number: 188609398 | Message Id Name: VUID-vkCmdDispatch-magFilter-04553
[...]

PR #86023, opened 2023-12-11
Fix Volumetric Fog VoxelGI updates

### LINKED ISSUE #85765: Volumetric fog with VoxelGI weird behavior depending on camera position
### Tested versions

- Reproducible in: 4.2.stable

### System information

Windows 10 - Godot v4.2.stable - Vulkan (Forward+) - Dedicated

### Issue description

In a scene with global volumetric fog and 2 VoxelGI nodes. When running the scene, if camera start out looking at one VoxelGI node but not another, the unseen node will get overwritten by the seen node. If camera start out looking at neither, both will become white. If camera can see both, both will look normal.
屏幕截图(473)
屏幕截图(474)
屏幕截图(475)

issue brought up by @Rytelier in https://github.com/godotengine/godot/issues/85642#issuecomment-1837609425_

### Steps to reproduce

In the minimal reproduction project, hold right mouse button to control camera. By default the camera is looking downward, not seeing any VoxelGI node.

1. Run light.tscn and rotate camera upward then around to see both VoxelGI nodes are foggy and white.
2. In editor, reset camera rotation, run scene again. The red emissive is normal, but turning the camera around, the other VoxelGI node is also red.
3. While looking at one VoxelGI node, maximize the window to trigger a (redraw?). The currently seeing node will update to normal. Turn the camera around, the other VoxelGI color is now overwritten.

[voxelgi_volfog_bug2.zip](https://github.com/godotengine/godot/files/13552147/voxelgi_volfog_bug2.zip)

bitsawer:
* Fixes https://github.com/godotengine/godot/issues/85765

The problem was that when VoxelGI volumes became visible in the camera or leave it, the active VoxelGI culling results were not updated to volumetric fog uniform sets. The updates only happened when toggling the enviroment off and on or resizing the windows, which forces a full update. More specifically, this volumetric fog uniform update code needs to be re-run every time when VoxelGI culling results change:

https://github.com/godotengine/godot/blob/b94eb58d35b3dd8a9f522bc90df0db73862ef326/servers/rendering/renderer_rd/environment/fog.cpp#L857-L864

There are few similar volumetric fog update issues that I think might be caused by similar missing uniform (or some other) update, for example https://github.com/godotengine/godot/issues/83990 (the same workaround also works there, turning environment off and on or resizing the window).

The Volumetric Fog and VoxelGI interactions and uniform set handling received some work in https://github.com/godotengine/godot/pull/76437, https://github.com/godotengine/godot/pull/76550 and https://github.com/godotengine/godot/pull/77703, so it could be useful to get some feedback from @RandomShaper here. There might be a more elegant or efficient way to implement the update notifying logic, especially if there are other cases like https://github.com/godotengine/godot/issues/83990 that need to be also handled in the future, but this should fix the immediate problem.

### DIFF
--- a/servers/rendering/renderer_rd/environment/fog.cpp
+++ b/servers/rendering/renderer_rd/environment/fog.cpp
@@ -931,9 +931,10 @@ void Fog::volumetric_fog_update(const VolumetricFogSettings &p_settings, const P
 			uniforms.push_back(u);
 		}
 
-		if (fog->copy_uniform_set.is_null()) {
-			fog->copy_uniform_set = RD::get_singleton()->uniform_set_create(copy_uniforms, volumetric_fog.process_shader.version_get_shader(volumetric_fog.process_shader_version, VolumetricFogShader::VOLUMETRIC_FOG_PROCESS_SHADER_COPY), 0);
+		if (fog->copy_uniform_set.is_valid() && RD::get_singleton()->uniform_set_is_valid(fog->copy_uniform_set)) {
+			RD::get_singleton()->free(fog->copy_uniform_set);
 		}
+		fog->copy_uniform_set = RD::get_singleton()->uniform_set_create(copy_uniforms, volumetric_fog.process_shader.version_get_shader(volumetric_fog.process_shader_version, VolumetricFogShader::VOLUMETRIC_FOG_PROCESS_SHADER_COPY), 0);
 
 		if (!gi_dependent_sets_valid) {
 			fog->gi_dependent_sets.process_uniform_set = RD::get_singleton()->uniform_set_create(uniforms, volumetric_fog.process_shader.version_get_shader(volumetric_fog.process_shader_version, VolumetricFogShader::VOLUMETRIC_FOG_PROCESS_SHADER_FOG), 0);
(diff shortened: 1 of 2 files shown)


Calinou (approved):
Tested locally, it works as expected. I've also tested a larger scene with a single VoxelGI node and it still renders correctly.

### Before

https://github.com/godotengine/godot/assets/180032/1aabf80f-0f39-470e-a85a-6d73923e4c9b

### After *(this PR)*

https://github.com/godotengine/godot/assets/180032/c37fd463-1b71-48c2-9f24-d5d1a9ccb9a8

BastiaanOlij (approved):
The heart of the fix seems correct, right now we detect whether the uniform is null or dependencies have been updated, but then only create the uniform set if it's null. 
I'm guessing at some point in time we would have freed the uniform set maybe?
But freeing the uniform set before (re)creating it makes sense.

Setting dependencies as changed while we update those seems sound to me as well.

akien-mga:
@bitsawer Could you rebase this PR just to make sure it still passes CI a few months later?

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #86028, opened 2023-12-11
Prevent crash when calling `set_text()` on a removed TreeItem

### LINKED ISSUE #85650: calling TreeItem set_text() on a TreeItem that was removed via remove_child() causes debugging to end without any error message
### Godot version

4.2 Stable

### System information

Linux

### Issue description

Calling TreeItem set_text() on a TreeItem that was removed via remove_child() causes debugging to end and the application to close without any error message.

### Steps to reproduce

Run the included example to get the debugger to crash. Comment out "treeItem.remove_child(subTreeItem)" for normal behaviour. 

[treeItemCrash.zip](https://github.com/godotengine/godot/files/13534005/treeItemCrash.zip)

jsjtxietian:
Fixes #85650

Also, should I use add more description or just use `ERR_FAIL_NULL` is sufficent ?

### DIFF
--- a/scene/gui/tree.cpp
+++ b/scene/gui/tree.cpp
@@ -70,19 +70,27 @@ void TreeItem::Cell::draw_icon(const RID &p_where, const Point2 &p_pos, const Si
 }
 
 void TreeItem::_changed_notify(int p_cell) {
-	tree->item_changed(p_cell, this);
+	if (tree) {
+		tree->item_changed(p_cell, this);
+	}
 }
 
 void TreeItem::_changed_notify() {
-	tree->item_changed(-1, this);
+	if (tree) {
+		tree->item_changed(-1, this);
+	}
 }
 
 void TreeItem::_cell_selected(int p_cell) {
-	tree->item_selected(p_cell, this);
+	if (tree) {
+		tree->item_selected(p_cell, this);
+	}
 }
 
 void TreeItem::_cell_deselected(int p_cell) {
-	tree->item_deselected(p_cell, this);
+	if (tree) {
+		tree->item_deselected(p_cell, this);
+	}
 }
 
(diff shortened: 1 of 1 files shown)


AThousandShips:
This should be before updating the text IMO, in the relevant methods calling these methods

Unless we can confirm it's valid and doesn't have strange sideeffects to update the text off the tree

YuriSizov (changes requested):
This would just raise a pointless error. It should be a normal if-check, not an error macro.

> Also what about other methods like `TreeItem::_cell_selected`?

And indeed, other similar methods should also be updated with such checks.

jsjtxietian:
> This should be before updating the text IMO, in the relevant methods calling these methods

Yes but there are too many places that might call `_changed_notify`, since ` the previous assumption tree items were always in a tree`

AThousandShips:
> Yes but there are too many places that might call

I don't think that's a valid argument against adding it there, but if you confirm that it's safe and doesn't cause any side effects to manipulate the text etc. not on the tree we can add it just to the places where `tree` is accessed, otherwise I'd say we shouldn't allow the manipulation if it's not safe

jsjtxietian:
> we'd need to verify this before we make this change

Indeed, you are right. Does anyone familiar with this part of code and can give me some suggestions?

AThousandShips:
A basic test would be to try your own code and see if there's any noticeable difference, if there isn't studying the code in more detail should help

YuriSizov (approved):
I think this is good as is right now. I don't particularly see a reason to check places that call these methods. Changing items should be fine out of `Tree`.

AThousandShips:
I'll trust that the manipulation of these is safe off the tree, despite the aspects mentioned above 🙂, but won't approve myself as I can't confirm it

akien-mga:
Thanks!

PR #66260, opened 2022-09-22
Fix array-bounds warning in BVH

lawnjelly:
Provides a workaround to prevent tripping a compiler warning.

Fixes #66252

## Description
A new compiler warning for `-Werror=array-bounds` was occurring in GCC 12.2.1. Investigation suggests this was an understandable false flag caused by some `DEV_ENABLED` only debugging code.

This fix alters the error condition code slightly to prevent tripping the warning.

There is no evidence of this being a genuine bug at runtime, because earlier code in `_logic_choose_item_add_node()` prevents this condition occurring (`request_item()` failing).

## Notes
* It would probably go horribly wrong shortly after such an error condition occurred, so I did consider making it a `CRASH_COND`, however, there's some chance the error message might print before things got that bad.
* We could add this to 3.x too, depending if we are upgrading GCC there at any point in the immediate future, as the BVH code is mostly shared.

Issue #66252, opened 2022-09-22
`-Werror=array-bounds` warnings raised in `core/math/bvh_structs.inc`


akien-mga:
### Godot version

4.0.beta (8e14f9ba21725a9445f3979742f2fc87c8a7b463)

### System information

Mageia 9 Linux, x86_64 | GCC 12.2.1 (20220917)

### Issue description

When compiling on Linux with GCC 12.2.1 (snapshot from 20220917) with warnings enabled (and `werror=yes` to treat them as errors), some `-Werror=array-bounds` warnings are raised in BVH structs. The warnings aren't raised by GCC 10.2.0 - I didn't test with GCC 11.x recently but I remember seeing those warnings for a while so probably any GCC 12.x version might raise them.

```
[...]

PR #86040, opened 2023-12-11
[3.x] Workaround GCC warning in `rasterizer_canvas_batcher`

lawnjelly:
`-Werror=array-bounds` flags when creating a new batch, possibly due to the possibility of the malloc failing (out of memory).
This PR adds an explicit `CRASHNOW` in the hope the compiler will recognise this case is not intended to be recoverable.

## Notes
* In Godot we don't (by choice) deal with the situation where heap allocations fail. This is notoriously difficult to recover from in any sensible fashion, and most game software doesn't bother.
* Unfortunately some versions of GCC (specifically 13.2) `Werror=array-bounds` seems to flag on some methods we use to deal with allocation failure. I'm not clear on whether it is choking on the possibility of malloc returning NULL, or inability to see inside the `grow` clauses. 
* Installing from source latest versions of GCC is a pain so I'm attempting to guess what is triggering the warning and see if @akien-mga can test. This may or may not fix the warning. :grin: 
* `LocalVector` should be susceptible to the same warning but does not flag, which is interesting. It calls `CRASH_NOW` in the case of allocation failure, so I'm attempting to just do the same here in the hopes the compiler can understand this pattern and prevent the warning. Previously the batcher would assert only in `DEV` builds, release builds would have crashed (without message). This is a hot path, but hopefully the grow branch is less hot so an if check should be okay.
* This is very similar case to #66252 `Werror=array-bounds` again has problems dealing with code 
See specifically https://github.com/godotengine/godot/issues/66252#issuecomment-1255149471 and #66260 .

## Discussion
If this doesn't cure the warning then maybe we need some `if` `else` clauses to guide the compiler warning logic, or alternatively turn off the warning locally. The code looks to be fine (unless I'm missing something obvious), so the problem seems to be the false warning. Google suggests this is one of a few warnings that are notorious for false flags.

### DIFF
--- a/drivers/gles_common/rasterizer_canvas_batcher.h
+++ b/drivers/gles_common/rasterizer_canvas_batcher.h
@@ -682,7 +682,7 @@ public:
 
 			// this should always succeed after growing
 			batch = bdata.batches.request();
-			RAST_DEBUG_ASSERT(batch);
+			CRASH_COND_MSG(!batch, "Out of memory");
 		}
 
 		if (p_blank) {


akien-mga (approved):
Looks good. I haven't tested this PR specifically but I tested a similar change (`CRASH_COND(!batch);`) when reporting the issue, and that solved it.

akien-mga:
Thanks!

PR #86041, opened 2023-12-11
Add type when converting a PackedArray to Array with constructor

mashumafi:
Converting `PackedArray` to `Array` doesn't have any type information. This PR enhances the `Array` constructor to call `set_typed` depending on the `PackedArray` that is supplied.

This is useful because some API will return `PackedArray` such a `String::split()` but it is useful to convert to an `Array` to do further processing with `map`, `reduce`, `filter`.

### DIFF
--- a/core/variant/variant_construct.cpp
+++ b/core/variant/variant_construct.cpp
@@ -202,15 +202,15 @@ void Variant::_register_variant_constructors() {
 	add_constructor<VariantConstructNoArgs<Array>>(sarray());
 	add_constructor<VariantConstructor<Array, Array>>(sarray("from"));
 	add_constructor<VariantConstructorTypedArray>(sarray("base", "type", "class_name", "script"));
-	add_constructor<VariantConstructorToArray<PackedByteArray>>(sarray("from"));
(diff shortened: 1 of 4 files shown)


AThousandShips:
Note that this breaks compatibility as you can no longer do:
```gdscript
var arr = Array(PackedVector2Array())
arr.push_back("foo")
```

AThousandShips:
I'd say this best served by implementing typed constructors, which is a goal, so you can do `Array[int](other)`, which also hels with the mentioned methods, but I'd say it should be opt-in

mashumafi:
> Note that this breaks compatibility 

> I'd say this best served by implementing typed constructors, which is a goal, so you can do `Array[int](other)`, which also help with the mentioned methods, but I'd say it should be opt-in

This is a good way to prevent breaking compatibility, opt-in sounds like the better option. Looks like some work needs to be done to support typed constructors. Any idea if this could be done without breaking compatibility though? If so, perhaps I can take a look if there isn't already an existing PR for TypeArray constructor.

AThousandShips:
It's being investigated by the team AFAIK, but it's a big task

mashumafi:
Closing due to incompat and other work. The correct solution depends on TypeArray constructor implementation.

PR #86044, opened 2023-12-11
Disable Add button when theme item name is empty

KoBeWi:
Theme editor has Add button that does nothing if name is empty. This PR disables it when not available.

https://github.com/godotengine/godot/assets/2223172/6b3818e4-4cec-46df-8b43-086561ec57ca


YuriSizov (approved):

editor/plugins/theme_editor_plugin.cpp
```diff
@@ -2850,6 +2853,11 @@ void ThemeTypeEditor::_add_default_type_items() {
 	ur->commit_action();
 }
 
+void ThemeTypeEditor::_update_add_button(const String &p_text, LineEdit *p_for_edit) {
+	Button *button = Object::cast_to<Button>(p_for_edit->get_meta("button"));
+	button->set_disabled(p_text.strip_edges().is_empty());
```
AThousandShips, line 2858: Since this logic of `strip_edges().is_empty()` is used relatively often it makes me wonder if it'd be worth it to add a convenience function that just checks if a string is non-empty after that operation, in this case if it has any codepoints `> 32`, will see about it later

AThousandShips (approved):

YuriSizov:
Thanks!

Issue #84471, opened 2023-11-05
AssetLib is not updating size and stuck


kosartofiq:
### Godot version

v4.1.1.stable.official [bd6af8e0e]

### System information

Godot v4.1.1.stable - Windows 10.0.22621 - Vulkan (Forward+) - integrated Intel(R) UHD Graphics 620 () - Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz (8 Threads)

### Issue description

in Assetlib tab will be so big size until can't see write panel and buttons.
![Screenshot 2023-11-05 045923](https://github.com/godotengine/godot/assets/67668868/1123f503-a9ca-4d7b-991c-c063de71f560)
![Screenshot 2023-11-05 050845](https://github.com/godotengine/godot/assets/67668868/1ba6475b-f329-43e5-bce9-00c5033e8737)

### Steps to reproduce

after open a project, click in asssitlib tab, for a short time will show tab in perfect size, but after one second will update itself, and hide a part of inspector(screenshot 1), also some of button for test scene and play to test scene also will be hidden, normally it resize all window to bigger than screen.
but when clicking 2d, 3d and script , the window comes back to normal size(maximized)
when restore it and drag right edge of the screen to show all contents of the window, (I have 2 screen) it needs wide of the 2 screen to show all the contents correctly, 
but when maximized the window again. half of the window hidden , then when click others tabs , will come back to correct windows(screenshot 2)
I restarted many times. result is the same

### Minimal reproduction project

[KurdishCasino.zip](https://github.com/godotengine/godot/files/13258617/KurdishCasino.zip)

Chaosus:
I think It's fixed in https://github.com/godotengine/godot/pull/80555, that PR may be back ported to 4.1.x in order to fix this.

AThousandShips:
It was, please try 4.1.3, 4.1.1 is no longer supported

kosartofiq:
I updated to 4.1.3, sorry for that I didn't know there is update. It is one problem of this program, it doesn't notify user about new update. anyway. there is still some bug left. when open window for first time. it show correctly(not updating itself for bigger size like in v4.1.1). and change between tabs have no problems. until when restore window, then I drag right side to make wider windows (I have 2 screen), then maximize it, the problem show again. 
![Screenshot 2023-11-05 133513](https://github.com/godotengine/godot/assets/67668868/fa57d4ae-5caa-43b1-8c1a-f3720aceac2d)

Calinou:
[...]

PR #80555, opened 2023-08-12
Assetlib: Fix long plugin names breaking the UI

GrammAcc:
Edit:
I think I have this fixed now.

Assets should have their titles truncated with an ellipsis to within the column width now, and I added a tooltip with the full asset title to the ones that get truncated, so that the user can still see the full title without opening the asset page.

I also changed the pagination from having 10 fixed page buttons to using the `EDSCALE` to dynamically choose a number of nav buttons to display so that it won't cause the horizontal overflow issue on large scales but will still show more page buttons on smaller scales.

It works on my machine at all scales up to 200%. At custom scales above 200% it starts to overflow again, but this happens regardless of the title and nav buttons. I tested this by truncating all titles to three characters and removing the nav buttons completely, and it still overflows at over 200% scale, so I think that is a separate issue and probably affects other parts of the engine UI as well.

I haven't written a unit test for this yet, but I can add one if needed. It will likely be next Saturday before I am able to get to it though.

Here are the updated screenshots:

<details>
<summary>
75% scale:
</summary>

75percent

</details>

<details>
<summary>
100% scale:
</summary>

100percent

</details>

<details>
<summary>
125% scale:
</summary>

125percent_top

125percent_longtitle

</details>

<details>
<summary>
150% scale:
</summary>

150percent_top

150percent_longtitle

</details>

<details>
<summary>
175% scale:
</summary>

175percent_top

175percent_longtitle

</details>

<details>
<summary>
200% scale:
</summary>

200percent_top

200percent_longtitle

</details>

Original PR:
I removed the line that disables horizontal scrolling in the AssetLib plugin. This fixes the screen overflow issue on my machine, but I am not sure if this is an appropriate solution since there was probably a reason why scroll was disabled in the first place, and I was unable to reproduce the column alignment issues seen in the screenshots posted in #80507 on either the current master or 4.1.1 stable, so I think there may be something platform specific going on here.

My system:
- Linux version 6.4.8-arch1-1 (linux@archlinux) (gcc (GCC) 13.2.1 20230801, GNU ld (GNU Binutils) 2.41.0) Thu, 03 Aug 2023 16:02:01 +0000

Please advise if enabling horizontal scroll in the AssetLib is a problem, and I will look for another solution.

[...]

PR #86051, opened 2023-12-11
Editor: Fix AssetLib is not updating size and stuck

GrammAcc:
This is an enhancement to the fix originally provided in #80555.

When I submitted the original PR I did not account for resizing the editor window at runtime, so the number of columns for asset lib search results was recalculating on resize, but the asset title truncation threshold was not being recalculated, so columns were not spacing correctly after resizing the editor.

I added a simple for loop to the NOTIFICATION_RESIZE event of the EditorAssetLibrary class that loops through the currently displayed assets and recalculates their title truncation length. I just used a for loop over the `asset_items->get_child_count` and fetched each EditorAssetLibraryItem with `asset_items->get_child` since I wanted to avoid explicit memory allocation inside the notification handler.

Please let me know if this is not a good solution or if the use of `static_cast` is discouraged.

I tested this solution locally, and it fixes the problems I mentioned in https://github.com/godotengine/godot/issues/84471#issuecomment-1832182508.

I think this will fix the scaling problem as reported in https://github.com/godotengine/godot/issues/84471#issuecomment-1793725120 since it forces recalculation of the asset titles, but I am unable to confirm for sure since I can't test this on Windows.

This PR doesn't resolve the TODO:

https://github.com/godotengine/godot/blob/15a03ed98e63590c85dac62405d785e1b7ca7fed/editor/plugins/asset_library_editor_plugin.cpp#L67C1-L67C53

I haven't figured out exactly how to get the font size with TextServer yet, so I will submit a separate PR for that small refactor eventually. I wanted to submit this fix now though since this fixes an actual flaw in the original.

### DIFF
--- a/editor/plugins/asset_library_editor_plugin.cpp
+++ b/editor/plugins/asset_library_editor_plugin.cpp
@@ -33,6 +33,7 @@
 #include "core/input/input.h"
 #include "core/io/json.h"
 #include "core/io/stream_peer_tls.h"
+#include "core/math/vector2.h"
 #include "core/os/keyboard.h"
 #include "core/version.h"
 #include "editor/editor_node.h"
@@ -56,6 +57,7 @@ static inline void setup_http_request(HTTPRequest *request) {
 
 void EditorAssetLibraryItem::configure(const String &p_title, int p_asset_id, const String &p_category, int p_category_id, const String &p_author, int p_author_id, const String &p_cost) {
 	title->set_text(p_title);
+	title->set_tooltip_text(p_title);
 	asset_id = p_asset_id;
 	category->set_text(p_category);
 	category_id = p_category_id;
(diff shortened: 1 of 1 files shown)


GrammAcc:
Forgot to run clang-format. :man_facepalming: Just amended with formatting fixed.

GrammAcc:
Okay, this should be ready for review now. The problem was that the `clamp_width` method was originally written with the assumption that it would be called on object initialization, so it was overriding the tooltip with the truncated version of the title text and wasn't handling the case of resizing to a larger size and un-truncating. I got this fixed and amended my commit.

I also updated the commit message with the details of the edits I made to the `clamp_width` method.

Please let me know if you have any questions about the implementation. Thanks!

ckdarby:
@GrammAcc, It looks like it'll fix my issue on Linux using a tile manager. I'll keep you posted when it gets merged and 4.3 comes out. Thanks for putting in the work.

GrammAcc:
I realized this morning that there is a better way to ensure that we don't overwrite the tooltip text with the truncated text. I currently have an if check for it in the `clamp_width` method, but I think it would be best to just set it in the `configure` method where the title text is set and then not worry about it, so I will update this PR over my lunch break today with a better design.

Also, I am working through the Accelerated C++ book in order to improve my contribution efficiency and quality, so hopefully my future PRs will be easier to review. :)

GrammAcc:
I updated the PR with a better implementation. Details are in the commit message. I removed the full_title_text private member of the EditorAssetLibraryItem and used the tooltip text as the full text instead. By moving the `set_tooltip_text` call into the `configure` method, I don't have to worry about the tooltip accidentally getting updated during the `clamp_width` call, so this works fine as an invariant. Let me know if there are any problems with this approach. Thanks!

GrammAcc:
CI is failing on the MacOS template_export job for some reason. I have no idea why, but I'm working on this.

GrammAcc:
I'm closing this since #88761 was merged and seems to solve the same problem.

PR #86052, opened 2023-12-12
Fix `RegEx.sub` truncating string when `end` is used

### LINKED ISSUE #55067: End parameter in RegEx.sub removes the end of the input string from the result
### Godot version

v3.4.stable.official [206ba70f4], v4.0.dev.custom_build [081c1ceda]

### System information

Windows 10

### Issue description

Not sure if this is an actual bug or just intended behaviour that should be better documented, but in any case I think it's confusing.

Currently, if you pass an `end` parameter to [`RegEx.sub`](https://docs.godotengine.org/en/latest/classes/class_regex.html#class-regex-method-sub) it will remove the remaining part of the string after `end` from the result:

```
var rgx = RegEx.new()
rgx.compile("a")
print(rgx.sub("aaaa", "b", true, 0, 2))
# bb
```

This is specially confusing because the same logic does not apply to `offset`, that is, the initial part of the string is never removed:

```
var rgx = RegEx.new()
rgx.compile("a")
print(rgx.sub("aaaa", "b", true, 2, 4))
# aabb
```

My initial expectation was that neither the initial or final parts of the string should be removed, but in any case I think the same logic should apply to both arguments. And if this is actually the intended behavior, then the function documentation should reflect that clearly.

### Steps to reproduce

Execute the following lines of GDScript code:

```
var rgx = RegEx.new()
rgx.compile("a")
print(rgx.sub("aaaa", "b", true, 0, 2))
print(rgx.sub("aaaa", "b", true, 2, 4))
```

The result is:

```
bb
aabb
```

While the expectation was:

```
bbaa
aabb
```

Or, as a second possibility:

```
bb
bb
```

_No response_

TheSofox:
Fixes #55067

When the `end` argument of `RegEx.sub` was used, it would truncate the Subject String before even doing the substitution. This fix ensures the end of the subject string is reattached, so that the function works as expected.

### DIFF
--- a/modules/regex/regex.cpp
+++ b/modules/regex/regex.cpp
@@ -334,7 +334,7 @@ String RegEx::sub(const String &p_subject, const String &p_replacement, bool p_a
 		return String();
 	}
 
-	return String(output.ptr(), olength);
+	return String(output.ptr(), olength) + p_subject.substr(length);
 }
 
 bool RegEx::is_valid() const {
--- a/modules/regex/tests/test_regex.h
+++ b/modules/regex/tests/test_regex.h
@@ -133,6 +133,18 @@ TEST_CASE("[RegEx] Substitution") {
 	RegEx re4("(a)(b){0}(c)");
 	REQUIRE(re4.is_valid());
 	CHECK(re4.sub(s4, "${1}.${3}.", true) == "a.c.a.c.a.c.");
+
+	const String s5 = "aaaa";
+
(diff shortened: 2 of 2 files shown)


TheSofox:
I've added some Unit Tests for this fix.

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #43018, opened 2020-10-22
Improve file sorting

groud:
Follow up of #42654.

Modifies the way the type sorting is done by also using the class icons in the comparison function. This specifically group similar resources together.

Also adds the possibility to sort according to the last modification date and in reserve order for each mode.
output

PR #86064, opened 2023-12-12
Fix "Class name cannot be empty" error when sorting no import files sort by type

### LINKED ISSUE #85698: Editor spams "Class name cannot be empty." error when sorting by type.
### Godot version

v4.2.stable.mono.official [46dc27791]

### System information

Godot v4.2.stable.mono - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3060 (NVIDIA; 31.0.15.2698) - AMD Ryzen 5 5600X 6-Core Processor (12 Threads)

### Issue description

Recently it started to happen that whenever I save my project in the editor, or other similar actions, the editor hangs for a second and I get hundreds of error messages in the log, saying "Class name cannot be empty".

When opening Godot from the terminal, the error message elaborates slightly that it comes from this line: https://github.com/godotengine/godot/blob/master/editor/editor_node.cpp#L4441

One thing I noticed is that it only seems to happen when I sort by type in the filesystem window. When sorting by name (default), I have not seen it happen yet.

I have no idea where in my project this error comes from, because the message doesn't say. It seems to be from the editor trying to find an icon for custom classes? But which class, and what would be wrong with said class? I don't know if it's user error on my part but I'm reporting as a bug to be sure.

I was able to delete a lot of my assets from a test project and only keep some textures mostly, and it still happens (although with fewer messages). When I delete the textures folder also, the issue stops entirely. So we at least know that some of the textures are some of the culprits (but not all). You probably need a .NET version of Godot to open the project, but maybe not since you don't need to actually compile to see the bug.

### Steps to reproduce

1. Open the project
2. In the filesystem window, set sort to "Sort By Type" (either ascending or descending is fine).
3. Observe 108 error messages in the output window.

https://mega.nz/file/gQ5DRRbB#NaxDQN5Tkk72C4w0j-ZCadYWQfY2aFfw0Kd40DN-RPg

jsjtxietian:
Fixes #85698

### DIFF
--- a/editor/filesystem_dock.cpp
+++ b/editor/filesystem_dock.cpp
@@ -865,18 +865,7 @@ void FileSystemDock::_search(EditorFileSystemDirectory *p_path, List<FileInfo> *
 
 struct FileSystemDock::FileInfoTypeComparator {
 	bool operator()(const FileInfo &p_a, const FileInfo &p_b) const {
-		// Uses the extension, then the icon name to distinguish file types.
-		String icon_path_a = "";
-		String icon_path_b = "";
-		Ref<Texture2D> icon_a = EditorNode::get_singleton()->get_class_icon(p_a.type);
-		if (icon_a.is_valid()) {
-			icon_path_a = icon_a->get_name();
-		}
-		Ref<Texture2D> icon_b = EditorNode::get_singleton()->get_class_icon(p_b.type);
-		if (icon_b.is_valid()) {
-			icon_path_b = icon_b->get_name();
-		}
-		return NaturalNoCaseComparator()(p_a.name.get_extension() + icon_path_a + p_a.name.get_basename(), p_b.name.get_extension() + icon_path_b + p_b.name.get_basename());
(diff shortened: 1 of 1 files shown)


KoBeWi (approved):
The fix looks alright.

Unrelated, but I wonder why the comparator uses icon path. Especially when it has the type available (and multiple types can share an icon if they don't use custom one).

YuriSizov:
> Unrelated, but I wonder why the comparator uses icon path. Especially when it has the type available (and multiple types can share an icon if they don't use custom one).

Apparently, this groups similar resources together: https://github.com/godotengine/godot/pull/43018

akien-mga:
@jsjtxietian Are you able to update this PR following the above feedback? It sounds like the icon sorting could be removed.

jsjtxietian:
Yes absolutely.

YuriSizov (approved):

YuriSizov:
Thanks!

PR #86065, opened 2023-12-12
[TextServer] Do not draw non-visual characters.

bruvzg:
Adds some checks to skip drawing non-visual glyphs and glyphs with zero size texture.

See https://github.com/godotengine/godot/pull/85811#pullrequestreview-1768611978

### DIFF
--- a/scene/3d/label_3d.cpp
+++ b/scene/3d/label_3d.cpp
@@ -326,28 +326,38 @@ Ref<TriangleMesh> Label3D::generate_triangle_mesh() const {
 }
 
 void Label3D::_generate_glyph_surfaces(const Glyph &p_glyph, Vector2 &r_offset, const Color &p_modulate, int p_priority, int p_outline_size) {
-	for (int j = 0; j < p_glyph.repeat; j++) {
-		Vector2 gl_of;
-		Vector2 gl_sz;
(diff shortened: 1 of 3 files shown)


scene/3d/label_3d.cpp
```diff
@@ -326,28 +326,38 @@ Ref<TriangleMesh> Label3D::generate_triangle_mesh() const {
...
+		return;
+	}
+
+	Vector2 gl_of;
+	Vector2 gl_sz;
+	Rect2 gl_uv;
+	Size2 texs;
+	RID tex;
+
+	if (p_glyph.font_rid != RID()) {
```
AThousandShips, line 340:
```suggestion
	if (p_glyph.font_rid.is_valid()) {
```
Same for all these, as we're rearranging here

Calinou (approved):
Tested locally, it works as expected with both Label and Label3D. This text with 36 visible characters results in 36 draw calls in Forward+/Mobile:

image

```
Test Test  Test
Test	Test		Test
TestTestTest
```

`Test Test  Test` in `master` results in 60 primitives in Label3D with the default settings (which includes a font outline), and 48 with this PR.

YuriSizov:
Thanks!

PR #86067, opened 2023-12-12
Remove non-existing global class from cache

### LINKED ISSUE #81867: "FIle not found" errors after deleting folder containing GDScript file with custom class_name
### Godot version

4.1.1

### System information

Windows 11

### Issue description

When a GDScript file with a custom class_name is deleted, the class is also removed from the editor.  However, if the folder is deleted (which contained the same GDScript file) the class is seemingly not removed from the editor and when you search for nodes, it tries to open the missing file and reports errors.

### Steps to reproduce

1. Create a new folder in the resource tree (eg example).
2. Inside this new folder, create a new GDScript file to Extend Node3D (although can be anything, I think)
3. Inside the new GDScript file, add the following line under the Extends Node3D line:
    class_name someclass
4. Save the GDScript file as myscript.gd.
5. Create a new Scene, select Other Node as the Root Node and search for your new class (someclass, in this example)
6. The option to extend your new scene from this class and GDScript file is available.
7. Cancel the Create New Node dialog
8. Delete the folder created in step 1
9. Attempt step 5 again - you will get errors in the console as follows:

Attempt to open script 'res://example/myscript.gd' resulted in error 'File not found'.
  Failed loading resource: res://example/myscript.gd. Make sure resources have been imported by opening the project in the editor at least once.
  editor/create_dialog.cpp:241 - Condition "scr.is_null()" is true.

Closing and reopening Godot has no effect.  The only way to fix this, is to create the same GDScript file again and delete the file rather that the folder it is in.  Deleting the file itself in step 8 results in no errors being reported.

N/A

miv391:
Fixes problems with deleting script files containing global classes.
- When a script file containg a global class is externally deleted, Godot editor doesn't handle this gracefully. Editor keeps spamming errors, but doesn't fix the situation.
- When a directory is removed in Godot editor, directory's files are more or less ignored and global classes from those files are not properly removed. Master assumes wrongly that EditorFileSystem's scan will handle the situation, but it doesn't. 
 
Fixes https://github.com/godotengine/godot/issues/81867

## Test cases

Test project contains two new classes, `my_good_sprite_2d.gd` located in root and another in `bad/my_bad_sprite_2d.gd`.

[g42_cache_test.zip](https://github.com/godotengine/godot/files/13646625/g42_cache_test.zip)

1. Open Godot and test project
2. Start adding new node by clicking + button in Scene tree.
--> SUCCESS: MyBadSprite2D is possible node type, no errors 

3. Delete directory "bad"
4. Start adding new node by clicking + button in Scene tree.
--> FAIL: You get errors 

kuva

1. Close Godot and delete `my_good_sprite_2d.gd` from file system
2. Open Godot and test project
3. Start adding new node by clicking + button in Scene tree.
--> FAIL: Now you errors also from MyGoodSprite2D 

kuva

The only situation where removing classes from cache file works, is when only files, not directories, are deleted in Godot editor.

## Fixes

The reason for the errors is that the global classes made by the user are not properly removed from `.godot/global_script_class_cache.cfg`. 

- Changes in `create_dialog.cpp` fix the situation when files are deleted externally. When a non-existing global class is detected, it is removed from the cache. This prevents further errors.

kuva

- Changes in `dependency_editor.cpp` fix the situation when a folder containing a class file is deleted in Godot editor. When directories are deleted, master doesn't process each deleted file the same way as single deleted files are processed. Instead it is assumed that EditorFileSystem's scan will handle this, but it doesn't. Scan doesn't notice deleted directories at all. Now for every deleted file, EditorFileSystem's `update_file` is called and the global class (if file contains one) is deleted from cache file.

### DIFF
--- a/editor/create_dialog.cpp
+++ b/editor/create_dialog.cpp
@@ -33,6 +33,7 @@
 #include "core/object/class_db.h"
 #include "core/os/keyboard.h"
 #include "editor/editor_feature_profile.h"
+#include "editor/editor_file_system.h"
 #include "editor/editor_node.h"
 #include "editor/editor_paths.h"
 #include "editor/editor_scale.h"
@@ -243,7 +244,13 @@ void CreateDialog::_add_type(const String &p_type, const TypeCategory p_type_cat
 	} else {
 		if (ScriptServer::is_global_class(p_type)) {
 			Ref<Script> scr = EditorNode::get_editor_data().script_class_load_script(p_type);
-			ERR_FAIL_COND(scr.is_null());
+			if (scr.is_null()) {
+				// Remove non-existing global class from class cache.
+				ScriptServer::remove_global_class(p_type);
+				ScriptServer::save_global_classes();
+				print_verbose("Removed class " + p_type + " from global classes.");
+				return;
+			}
 
 			Ref<Script> base = scr->get_base_script();
 			if (base.is_null()) {
--- a/editor/dependency_editor.cpp
+++ b/editor/dependency_editor.cpp
@@ -574,52 +574,47 @@ void DependencyRemoveDialog::ok_pressed() {
 			emit_signal(SNAME("resource_removed"), res);
 			res->set_path("");
 		}
-	}
 
-	for (int i = 0; i < files_to_delete.size(); ++i) {
 		// If the file we are deleting for e.g. the main scene, default environment,
 		// or audio bus layout, we must clear its definition in Project Settings.
-		if (files_to_delete[i] == String(GLOBAL_GET("application/config/icon"))) {
+		if (file == String(GLOBAL_GET("application/config/icon"))) {
 			ProjectSettings::get_singleton()->set("application/config/icon", "");
 		}
-		if (files_to_delete[i] == String(GLOBAL_GET("application/run/main_scene"))) {
+		if (file == String(GLOBAL_GET("application/run/main_scene"))) {
 			ProjectSettings::get_singleton()->set("application/run/main_scene", "");
 		}
-		if (files_to_delete[i] == String(GLOBAL_GET("application/boot_splash/image"))) {
+		if (file == String(GLOBAL_GET("application/boot_splash/image"))) {
 			ProjectSettings::get_singleton()->set("application/boot_splash/image", "");
 		}
-		if (files_to_delete[i] == String(GLOBAL_GET("rendering/environment/defaults/default_environment"))) {
+		if (file == String(GLOBAL_GET("rendering/environment/defaults/default_environment"))) {
 			ProjectSettings::get_singleton()->set("rendering/environment/defaults/default_environment", "");
 		}
-		if (files_to_delete[i] == String(GLOBAL_GET("display/mouse_cursor/custom_image"))) {
+		if (file == String(GLOBAL_GET("display/mouse_cursor/custom_image"))) {
 			ProjectSettings::get_singleton()->set("display/mouse_cursor/custom_image", "");
 		}
-		if (files_to_delete[i] == String(GLOBAL_GET("gui/theme/custom"))) {
+		if (file == String(GLOBAL_GET("gui/theme/custom"))) {
 			ProjectSettings::get_singleton()->set("gui/theme/custom", "");
 		}
-		if (files_to_delete[i] == String(GLOBAL_GET("gui/theme/custom_font"))) {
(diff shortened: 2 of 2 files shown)


editor/create_dialog.cpp
```diff
@@ -243,7 +244,13 @@ void CreateDialog::_add_type(const String &p_type, const TypeCategory p_type_cat
 	} else {
 		if (ScriptServer::is_global_class(p_type)) {
 			Ref<Script> scr = EditorNode::get_editor_data().script_class_load_script(p_type);
-			ERR_FAIL_COND(scr.is_null());
+			if (scr.is_null()) {
+				// Remove non-existing global class from class cache.
+				ScriptServer::remove_global_class(p_type);
+				ScriptServer::save_global_classes();
+				print_line("Removed class " + p_type + " from global classes.");
```
AThousandShips, line 251:
```suggestion
				print_verbose("Removed class " + p_type + " from global classes.");
```
Or no printing here at all IMO

miv391: In my opinion, the best solution would be totally silent and just to remove the class from cache file. User doesn't care. But suppressing those error messages wouldn't be easy. So I added that print to show the user that _something_ was done after those errors. I fear that `print_verbose` would mean that next to nobody would see the message. Then removing the whole message would be as good.

AThousandShips: I don't agree, the message matters for verbose printing to debug things, but removing the printing all together is fine, but it has some value, but printing it always isn't good

editor/create_dialog.cpp
```diff
@@ -243,7 +244,13 @@ void CreateDialog::_add_type(const String &p_type, const TypeCategory p_type_cat
...
 		if (ScriptServer::is_global_class(p_type)) {
 			Ref<Script> scr = EditorNode::get_editor_data().script_class_load_script(p_type);
-			ERR_FAIL_COND(scr.is_null());
+			if (scr.is_null()) {
+				// Remove non-existing global class from class cache.
+				ScriptServer::remove_global_class(p_type);
+				ScriptServer::save_global_classes();
+				print_verbose("Removed class " + p_type + " from global classes.");
+				return;
+			}
```
KoBeWi, line 253: Not sure if CreateDialog should be responsible for cleaning classes..

editor/dependency_editor.cpp
```diff
@@ -574,52 +574,47 @@ void DependencyRemoveDialog::ok_pressed() {
...
 		}
-		if (files_to_delete[i] == String(GLOBAL_GET("audio/buses/default_bus_layout"))) {
+		if (file == String(GLOBAL_GET("audio/buses/default_bus_layout"))) {
 			ProjectSettings::get_singleton()->set("audio/buses/default_bus_layout", "");
 		}
 
-		String path = OS::get_singleton()->get_resource_dir() + files_to_delete[i].replace_first("res://", "/");
+		String path = OS::get_singleton()->get_resource_dir() + file.replace_first("res://", "/");
 		print_verbose("Moving to trash: " + path);
 		Error err = OS::get_singleton()->move_to_trash(path);
```
KoBeWi, line 607: This is wrong. It makes every file deleted separately when deleting directories.
You should still use `files_to_delete` for this action.

editor/dependency_editor.cpp
```diff
@@ -574,52 +574,47 @@ void DependencyRemoveDialog::ok_pressed() {
...
 		if (err != OK) {
-			EditorNode::get_singleton()->add_io_error(TTR("Cannot remove:") + "\n" + files_to_delete[i] + "\n");
+			EditorNode::get_singleton()->add_io_error(TTR("Cannot remove:") + "\n" + file + "\n");
 		} else {
-			emit_signal(SNAME("file_removed"), files_to_delete[i]);
+			emit_signal(SNAME("file_removed"), file);
 		}
+
+		EditorFileSystem::get_singleton()->update_file(file);
```
KoBeWi, line 614: Is this necessary for something? When you delete directory, this should be handled by rescan (like in the old code).

miv391: Then maybe the rescan should be fixed? Currently it doesn't handle delete directories. There is some code to handle deleted directories, but it is never called (at least not when directories are deleted from FileSystem).

KoBeWi: Well I was more asking whether this line fixes anything. If there is a case not handled by the above lines then this one could stay.

akien-mga:
Superseded by #90186. Thanks for the contribution!

PR #50196, opened 2021-07-06
Add reflection mask to reflection probes

BastiaanOlij:
The `cull_mask` property on reflection probes had a double function that did not make sense.

1. It determined which objects were rendered to the probe
2. It determined which objects would get reflections using this probe

Rarely do you want to have an object reflect onto itself so a likely use case is where you'd want to exclude the object that gets reflections from being rendered to the probe. But due to this combined function that was impossible.

The documentation stated that it should be doing 1 and never mentioned 2 however seeing that for all other objects that have a `cull_mask` (i.e. lights and decals) `cull_mask` behaves as defined at 2 I've chosen to split off 1 into a separate property.

This PR introduces a new property called `reflection_mask` that determines which layers are rendered to the probe.

Using the material tester demo I've added a reflection probe to the mirror ball:
image

Note that the mirror ball itself has been placed into layer 2

The result is proper reflections onto the mirror ball:
image

Fixes #50195

PR #86073, opened 2023-12-12
Add a reflection mask to the reflection probes

m4rr5:
This is a rebase of PR #50196. Co-authored-by: Bastiaan Olij <mux213@gmail.com>

The use case for this PR is to render an object with a reflective material, in this case a car, in a scene that should be reflected on its body. For that you place a reflection probe in the center of the car. The problem you now run into is that you do not want the (interior of the) car rendered in this probe. The `cull_mask` that each reflection probe has, does not allow you to solve this. If you put the car on a separate layer and disable that layer, not only does it no longer show up in the reflected map, but it automatically also no longer uses the same map to render the car.

This PR adds a `reflection_mask` property to reflection probes that allows you to specify which objects are rendered to the probe. The `cull_mask` still determines which objects are using this reflection probe.

To demonstrate the issue, the image below shows a car in a simple scene. In the center of the car is a reflection probe and to visualize the issue, directly above the car a reflective sphere was added that also shows what gets rendered in the reflection probe. You can clearly see how it renders the interior of the car, making the reflections on the car body look wrong.

car_1

The second image is taken with the same setup, but with layer 2 of the `reflection_mask` disabled. Layer 1 contains everything around the car. Layer 2 is the car. The reflections now look as they should on the car body. The `cull_mask` can still be used to determine what objects to target with this probe (not done explicitly in this sample screenshot).

car_2

### DIFF
--- a/doc/classes/ReflectionProbe.xml
+++ b/doc/classes/ReflectionProbe.xml
@@ -28,7 +28,8 @@
 			[b]Note:[/b] To better fit rectangle-shaped rooms that are not aligned to the grid, you can rotate the [ReflectionProbe] node.
 		</member>
 		<member name="cull_mask" type="int" setter="set_cull_mask" getter="get_cull_mask" default="1048575">
-			Sets the cull mask which determines what objects are drawn by this probe. Every [VisualInstance3D] with a layer included in this cull mask will be rendered by the probe. To improve performance, it is best to only include large objects which are likely to take up a lot of space in the reflection.
+			Sets the cull mask which determines what objects are drawn by this probe. Every [VisualInstance3D] with a layer included in this cull mask will be rendered by the probe. It is best to only include large objects which are likely to take up a lot of space in the reflection in order to save on rendering cost.
+			This can also be used to prevent an object from reflecting upon itself (for instance, a [ReflectionProbe] centered on a vehicle).
 		</member>
 		<member name="enable_shadows" type="bool" setter="set_enable_shadows" getter="are_shadows_enabled" default="false">
 			If [code]true[/code], computes shadows in the reflection probe. This makes the reflection probe slower to render; you may want to disable this if using the [constant UPDATE_ALWAYS] [member update_mode].
@@ -49,6 +50,9 @@
 		</member>
 		<member name="origin_offset" type="Vector3" setter="set_origin_offset" getter="get_origin_offset" default="Vector3(0, 0, 0)">
 			Sets the origin offset to be used when this [ReflectionProbe] is in [member box_projection] mode. This can be set to a non-zero value to ensure a reflection fits a rectangle-shaped room, while reducing the number of objects that "get in the way" of the reflection.
+		</member>
+		<member name="reflection_mask" type="int" setter="set_reflection_mask" getter="get_reflection_mask" default="1048575">
+			Sets the reflection mask which determines what objects have reflections applied from this probe. Every [VisualInstance3D] with a layer included in this reflection mask will have reflections applied from this probe. See also [member cull_mask], which can be used to exclude objects from appearing in the reflection while still making them affected by the [ReflectionProbe].
 		</member>
 		<member name="size" type="Vector3" setter="set_size" getter="get_size" default="Vector3(20, 20, 20)">
 			The size of the reflection probe. The larger the size, the more space covered by the probe, which will lower the perceived resolution. It is best to keep the size only as large as you need it.
--- a/doc/classes/RenderingServer.xml
+++ b/doc/classes/RenderingServer.xml
@@ -2894,7 +2894,7 @@
 			<param index="0" name="probe" type="RID" />
 			<param index="1" name="layers" type="int" />
 			<description>
-				Sets the render cull mask for this reflection probe. Only instances with a matching cull mask will be rendered by this probe. Equivalent to [member ReflectionProbe.cull_mask].
+				Sets the render cull mask for this reflection probe. Only instances with a matching layer will be reflected by this probe. Equivalent to [member ReflectionProbe.cull_mask].
 			</description>
 		</method>
 		<method name="reflection_probe_set_enable_box_projection">
@@ -2943,6 +2943,14 @@
 			<param index="1" name="offset" type="Vector3" />
 			<description>
 				Sets the origin offset to be used when this reflection probe is in box project mode. Equivalent to [member ReflectionProbe.origin_offset].
+			</description>
+		</method>
+		<method name="reflection_probe_set_reflection_mask">
+			<return type="void" />
+			<param index="0" name="probe" type="RID" />
+			<param index="1" name="layers" type="int" />
+			<description>
+				Sets the render reflection mask for this reflection probe. Only instances with a matching layer will have reflections applied from this probe. Equivalent to [member ReflectionProbe.reflection_mask].
 			</description>
 		</method>
 		<method name="reflection_probe_set_resolution">
--- a/drivers/gles3/storage/light_storage.cpp
+++ b/drivers/gles3/storage/light_storage.cpp
@@ -468,6 +468,9 @@ void LightStorage::reflection_probe_set_enable_shadows(RID p_probe, bool p_enabl
 void LightStorage::reflection_probe_set_cull_mask(RID p_probe, uint32_t p_layers) {
 }
 
+void LightStorage::reflection_probe_set_reflection_mask(RID p_probe, uint32_t p_layers) {
+}
+
 void LightStorage::reflection_probe_set_resolution(RID p_probe, int p_resolution) {
 }
 
@@ -480,6 +483,10 @@ RS::ReflectionProbeUpdateMode LightStorage::reflection_probe_get_update_mode(RID
 }
 
 uint32_t LightStorage::reflection_probe_get_cull_mask(RID p_probe) const {
+	return 0;
+}
+
+uint32_t LightStorage::reflection_probe_get_reflection_mask(RID p_probe) const {
 	return 0;
 }
 
--- a/drivers/gles3/storage/light_storage.h
+++ b/drivers/gles3/storage/light_storage.h
@@ -575,6 +575,7 @@ public:
 	virtual void reflection_probe_set_enable_box_projection(RID p_probe, bool p_enable) override;
 	virtual void reflection_probe_set_enable_shadows(RID p_probe, bool p_enable) override;
 	virtual void reflection_probe_set_cull_mask(RID p_probe, uint32_t p_layers) override;
+	virtual void reflection_probe_set_reflection_mask(RID p_probe, uint32_t p_layers) override;
 	virtual void reflection_probe_set_resolution(RID p_probe, int p_resolution) override;
 	virtual void reflection_probe_set_mesh_lod_threshold(RID p_probe, float p_ratio) override;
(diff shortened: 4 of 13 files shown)


Calinou (approved):
Tested locally, it works as expected in Forward+ and Mobile.

**Testing project:** [test_pr_86073.zip](https://github.com/godotengine/godot/files/13649165/test_pr_86073.zip)

Screenshot_20231212_151624

doc/classes/ReflectionProbe.xml
```diff
@@ -28,7 +28,10 @@
 			[b]Note:[/b] To better fit rectangle-shaped rooms that are not aligned to the grid, you can rotate the [ReflectionProbe] node.
 		</member>
 		<member name="cull_mask" type="int" setter="set_cull_mask" getter="get_cull_mask" default="1048575">
-			Sets the cull mask which determines what objects are drawn by this probe. Every [VisualInstance3D] with a layer included in this cull mask will be rendered by the probe. To improve performance, it is best to only include large objects which are likely to take up a lot of space in the reflection.
+			Sets the cull mask which determines what objects have reflections applied from this probe. Every [VisualInstance3D] with a layer included in this cull mask will have reflections applied from this probe.
```
Calinou, line 31:
```suggestion
			Sets the cull mask which determines what objects have reflections applied from this probe. Every [VisualInstance3D] with a layer included in this cull mask will have reflections applied from this probe. See also [method reflection_mask], which can be used to exclude objects from the reflection while still making them affected by the [ReflectionProbe].
```

doc/classes/ReflectionProbe.xml
```diff
@@ -28,7 +28,10 @@
...
+		</member>
+		<member name="reflection_mask" type="int" setter="set_reflection_mask" getter="get_reflection_mask" default="1048575">
+			Sets the reflection mask which determines what objects are drawn by this probe. Every [VisualInstance3D] with a layer included in this reflection mask will be rendered by the probe. It is best to only include large objects which are likely to take up a lot of space in the reflection in order to save on rendering cost.
```
Calinou, line 34:
```suggestion
			Sets the reflection mask which determines what objects are drawn by this probe. Every [VisualInstance3D] with a layer included in this reflection mask will be rendered by the probe. It is best to only include large objects which are likely to take up a lot of space in the reflection in order to save on rendering cost.
			This can also be used to prevent objects from reflecting upon themselves (for instance, a [ReflectionProbe] centered on a vehicle).
```

doc/classes/ReflectionProbe.xml
```diff
@@ -28,7 +28,11 @@
 			[b]Note:[/b] To better fit rectangle-shaped rooms that are not aligned to the grid, you can rotate the [ReflectionProbe] node.
 		</member>
 		<member name="cull_mask" type="int" setter="set_cull_mask" getter="get_cull_mask" default="1048575">
-			Sets the cull mask which determines what objects are drawn by this probe. Every [VisualInstance3D] with a layer included in this cull mask will be rendered by the probe. To improve performance, it is best to only include large objects which are likely to take up a lot of space in the reflection.
+			Sets the cull mask which determines what objects have reflections applied from this probe. Every [VisualInstance3D] with a layer included in this cull mask will have reflections applied from this probe. See also [method reflection_mask], which can be used to exclude objects from appearing in the reflection while still making them affected by the [ReflectionProbe].
```
AThousandShips, line 31:
```suggestion
			Sets the cull mask which determines what objects have reflections applied from this probe. Every [VisualInstance3D] with a layer included in this cull mask will have reflections applied from this probe. See also [member reflection_mask], which can be used to exclude objects from appearing in the reflection while still making them affected by the [ReflectionProbe].
```

AThousandShips (approved):

AThousandShips:

AThousandShips:
Please update your commit message to include the co aurhor properly, like so (exactly, the lines are important):
```
Add a reflection mask to the reflection probes.


Co-authored-by: Bastiaan Olij <mux213@gmail.com>
```

BastiaanOlij:
Looking good @m4rr5  !

BastiaanOlij:
Just some more background info on this fix and the discussion that may need to be had here.

My original PR became stale because at the time we had no clear use case to test the enhancement against, it was based on a simple experiment as I was converting the material demo to Godot 4 at the time. Seeing Godot 4 was years away from release and the rendering team had their hands full, this wasn't exactly and issue that deserved to be at the top of the priority list, and thus became largely forgotten :P

So currently, without this PR, reflection probes have two masks:
- The `VisualInstance3D` `Layers` mask, which is the mask that is applied when culling for our camera (the one we render the scene with), what can our camera see?
- The `ReflectionProbe` `Cull Mask`, which is the mask that determine which objects the reflection probe is applied to (similar to cull masks on lights, decals, etc).

What is missing was a mask that determines what is rendered on the reflection probe itself, a camera mask for the reflection probe as it were. My old PR, and by extension this reworked PR, solves this by introducing a new mask for this, the `Reflection Mask`, e.g. "What is reflected in this probe".

There was some debate whether this problem existed in Godot 3 and how Godot 3 solved it, and whether we should ensure Godot 4 works the same. To the best of my memory, if it works in Godot 3 it is because the `Layers` mask is also used as the `Reflection Mask` and that would in theory work.

The problem I ran into at the time is that our rendering architecture hides this information where we need it, and there was no real way to access this information. The camera culling system applies the `Layers` mask and builds a list of objects (including reflection probes) that need to be used to render the viewport, it promptly forgets about the `Layers` mask at this point in time as it has been applied and is no longer needed. When we get to rendering the reflection probe, all we have access to is the `Cull Mask`.

There is an equal argument that having this as two separate masks makes sense as you tie two unrelated things together.

This may have changed, this was over 2 years ago that I last worked on it.

servers/rendering/renderer_scene_cull.cpp
```diff
@@ -3465,7 +3465,7 @@ bool RendererSceneCull::_render_reflection_probe_step(Instance *p_instance, int
 		camera_data.set_camera(xform, cm, false, false);
 
 		Ref<RenderSceneBuffers> render_buffers = RSG::light_storage->reflection_probe_atlas_get_render_buffers(scenario->reflection_atlas);
-		_render_scene(&camera_data, render_buffers, environment, RID(), RSG::light_storage->reflection_probe_get_cull_mask(p_instance->base), p_instance->scenario->self, RID(), shadow_atlas, reflection_probe->instance, p_step, mesh_lod_threshold, use_shadows);
+		_render_scene(&camera_data, render_buffers, environment, RID(), RSG::light_storage->reflection_probe_get_reflection_mask(p_instance->base), p_instance->scenario->self, RID(), shadow_atlas, reflection_probe->instance, p_step, mesh_lod_threshold, use_shadows);
```
BastiaanOlij, line 3468: Just to complete my comment, if we want to go down the route of how I think it works in Godot 3, we would have to pass the `Layers` mask here.

We just don't have access to the `Layers` mask, we have no data with which to retrieve it.

IPlayKindred:
@clayjohn  dude check this out man

AThousandShips:
@IPlayKindred clayjohn has already been requested for a review and will review this when he has the time, no need to hurry anyone up, it's the holidays, please have patience

IPlayKindred:
oh yeah didnt think of that at all, I am not christian so it didnt cross my mind at all, happy holidays!

clayjohn:
I think this looks fine. But I want to take a moment to record the progression of this issue so we can discuss with other contributors and decide if this is what we want. 

### Godot 3.x
In Godot 3.x the cull_mask controlled which layers were rendered by the reflection probe. Items in the matching layer are shown in the reflection. 

Reflections appear on all objects that are visible and are inside the reflection probe's bounds.

### Godot 4.2

In Godot 4.2 cull_mask controls both which layers are rendered by the reflection probe and which layers use the reflection from the reflection probe.

### This PR

I think this PR makes sense. I am a bit apprehensive at adding yet another mask to ReflectionProbes, but in this case I think it is necessary. Contrast this with Lights. Lights also have a cull_mask which controls which layers they affect. But ReflectionProbes need to both affect and be affected by layers. They are both cameras and a light source so it makes sense to let users control both. 

I think the only major decision point we need to pause and think about is whether cull_mask should match the behaviour of 3.x (i.e. operate as a camera cull_mask) or whether it should match the behaviour of other visual instances like lights and decals.

jcostello:
`cull_mask` should control with meshes are render intro the reflection

the new layer control should determine what meshes are affected by the reflection

that simple should be really effective

clayjohn (commented):
Let's wait for @Bastiaan to weigh in, but my preference is also for ``cull_mask`` to retain its meaning and control what objects are rendered by the reflection probe (i.e. like a camera's cull mask) and for ``reflection_mask`` to control which objects receive reflections from the probe. 

To do that you can revert the change in renderer_scene_cull.cpp so that ``cull_mask`` is used when rendering the reflection probe.

Then just use ``reflection_mask`` instead of ``cull_mask`` when creating reflection probe instances internally here:
https://github.com/godotengine/godot/blob/179dfdc8d78b5bd5377dd115af026df58308bdaf/servers/rendering/renderer_rd/storage_rd/light_storage.cpp#L1684-L1696

BastiaanOlij:
@clayjohn ah, I see my confusion. I was mixing up `cull mask` on the reflection probe and the `layers` property on the Visual Instance (which is the one that we loose access to during culling). 

Swapping cull mask and reflection layer is indeed easy and makes sense.

BastiaanOlij:
> I think the only major decision point we need to pause and think about is whether cull_mask should match the behaviour of 3.x (i.e. operate as a camera cull_mask) or whether it should match the behaviour of other visual instances like lights and decals.

This change is applied through https://github.com/m4rr5/godot/pull/1
@m4rr5 will have to merge that into his branch for this PR to update.

Looking at this more closely, I agree swapping these two masks around. It makes much more sense.

servers/rendering/renderer_scene_cull.cpp
clayjohn, line 3468:
```suggestion
		_render_scene(&camera_data, render_buffers, environment, RID(), RSG::light_storage->reflection_probe_get_cull_mask(p_instance->base), p_instance->scenario->self, RID(), shadow_atlas, reflection_probe->instance, p_step, mesh_lod_threshold, use_shadows);
```
This change needs to be reverted. Right now in this PR reflection mask is used to cull rendering and reflections.

m4rr5:
Thanks @clayjohn for the review and fix. I've applied it and retested the build on my machine and it works fine now. The PR is updated and squashed.

servers/rendering/renderer_scene_cull.cpp
BastiaanOlij: That was the flip wasn't it? cull mask used for what is rendered to the reflection mask. Reflection mask determines on which objects the reflections are rendered.

With the changing this back, reflection mask is no longer used, and cull mask is now used for both.

BastiaanOlij: Never mind, I was being cross eyed...

BastiaanOlij (approved):

clayjohn (approved):
Looks great!

YuriSizov:
Thanks for picking it up and carrying it across the finish line! And congrats on your first merged Godot contribution!

PR #86079, opened 2023-12-12
Add `RequiredParam<T>` and `RequiredResult<T>` to mark `Object *` arguments and return values as required

### LINKED ISSUE #2241: Add method hints for non-nullable return types and nullable arguments
*See also https://github.com/godotengine/godot-proposals/issues/162.*

The Godot editor :slightly_smiling_face: 

Currently, GDNative bindings must assume that all non-primitive return types (such as Node) may be `null`. This is problematic for safety-oriented languages such as Rust and Swift, as it leads to uglier and unsafe code.

- Add method hints to ClassDB to signal that a return type will *never* be `null`.
- Add a way to hint specific method arguments as being nullable. In other words, it means that the method will behave correctly if the specified arguments are `null`.

This information can also be displayed in the generated class reference to improve documentation.

- Non-null return type hinting could be implemented as a ClassDB method flag such as `METHOD_FLAG_RETURN_NON_NULL`.
- Nullable argument hinting could be implemented using argument metadata feature (`GodotTypeInfo::Metadata`).

No, as this is core engine functionality.

This functionality is part of the engine itself.

dsnopek:
Implements proposal https://github.com/godotengine/godot-proposals/issues/2241

The core problem that this is solving is that in some languages, different language constructs/features are used with nullable and non-nullable values, where the non-nullable ones are usually friendlier for the developer to use. This has come up numerous times specifically in the context of Rust, Swift and C#, but would also apply to Kotlin and probably others.

The goal of this PR is to mark various arguments and return values in `extension_api.json` as "required", which means:

1. Passing null for a required parameter is an error, which will prevent the method from doing its intended function, and so the GDExtension binding can attempt to prevent passing null into that argument using language features. (But this isn't required.)
2. Returning null for a required return value is an error, and the GDExtension language binding can use that as a signal to do whatever error behavior makes the most sense for that language, for example, throwing an exception (ex in Python, C#, etc) or panicking (ex Rust). (But this isn't required.)

How this is accomplished:

- Instead of using `Object *` or `const Ref<T> &`  in method arguments, you can use `RequiredParam<T>`, which will get the parameter marked as "required" in `ClassDB`.
- Some new error macros are added, like `EXTRACT_REQUIRED_PARAM_OR_FAIL(m_name, m_param)` which acts just like `ERR_FAILED_NULL(m_param)`, except it also assigns a local variable with the object, if it wasn't null. Using the macro is _required_ in order to get the underlying value out of the parameter, which has two advantages:
  - It will help prevent contributors from using `RequiredParam<T>` when they don't really mean it, so (hopefully) these markers are less likely to become out-of-date
  - This will help contributors to remember to check for null (like we do currently do with the `ERR_*` macros) which is beneficial to engine development, even outside the context of GDExtension
- On return values, instead of using `RequiredParam<T>` there is `RequiredResult<T>`. This separate object is used because return values are more of a "soft" requirement and don't need a special process in order to extract the value
- In `ClassDB`, the meta for any of required parameters or return values will be `METADATA_OBJECT_IS_REQUIRED`, which will look like this in the `extension_api.json`:
  ```
  "arguments": [
    {
      "name": "node",
      "type": "Node",
      "meta": "required"
    }
  ]
  ```
- Just to test, this PR has a 2nd commit which marks a bunch of parameters and return values as required
- In the parameters, I've used a new naming convention of `rp_*` (rather than `p_*`) in order to minimize the diff when switching to `RequiredParam<T>`. This makes it so each change is just +/- 2 lines in the diff, and it's easy to review and see that there were no other functional changes
- When compiling a release build, the goal is for `RequiredParam<T>` and `RequiredResult<T>` to be optimized away, and generate basically the same binary as using a plain `Object *` or `Ref<T>`.

**UPDATE  (2024-12-10):** The original version of this had this whole auto-vivifying thing that was terrible, and was replaced with an excellent suggestion from @Bromeon in [this comment](https://github.com/godotengine/godot/pull/86079#issuecomment-2327313456).

**UPDATE (2025-10-17):** The semantics of parameters and return values turned out to be different enough that this now has two templates (`RequiredParam<T>` and `RequiredResult<T>` instead of the single `RequiredPtr<T>`).

### DIFF
--- a/core/object/object.h
+++ b/core/object/object.h
@@ -39,6 +39,7 @@
 #include "core/templates/hash_set.h"
 #include "core/templates/list.h"
 #include "core/templates/safe_refcount.h"
+#include "core/variant/required_ptr.h"
 #include "core/variant/variant.h"
 
 template <typename T>
--- a/core/variant/method_ptrcall.h
+++ b/core/variant/method_ptrcall.h
@@ -269,6 +269,42 @@ struct PtrToArg<const T *> {
 	}
 };
 
+// This is for RequiredParam.
+
+template <class T>
+struct PtrToArg<RequiredParam<T>> {
+	typedef typename RequiredParam<T>::ptr_type EncodeT;
+
+	_FORCE_INLINE_ static RequiredParam<T> convert(const void *p_ptr) {
+		if (p_ptr == nullptr) {
+			return RequiredParam<T>::_err_return_dont_use();
+		}
+		return RequiredParam<T>(*reinterpret_cast<T *const *>(p_ptr));
+	}
+
+	_FORCE_INLINE_ static void encode(const RequiredParam<T> &p_var, void *p_ptr) {
+		*((typename RequiredParam<T>::ptr_type *)p_ptr) = p_var._internal_ptr_dont_use();
+	}
+};
+
+// This is for RequiredResult.
+
+template <class T>
+struct PtrToArg<RequiredResult<T>> {
+	typedef typename RequiredResult<T>::ptr_type EncodeT;
+
+	_FORCE_INLINE_ static RequiredResult<T> convert(const void *p_ptr) {
+		if (p_ptr == nullptr) {
+			return RequiredResult<T>::_err_return_dont_use();
+		}
+		return RequiredResult<T>(*reinterpret_cast<T *const *>(p_ptr));
+	}
+
+	_FORCE_INLINE_ static void encode(const RequiredResult<T> &p_var, void *p_ptr) {
+		*((typename RequiredResult<T>::ptr_type *)p_ptr) = p_var._internal_ptr_dont_use();
+	}
+};
+
 // This is for ObjectID.
 
 template <>
--- a/core/variant/required_ptr.h
+++ b/core/variant/required_ptr.h
@@ -1,0 +1,246 @@
+/**************************************************************************/
+/*  required_ptr.h                                                        */
+/**************************************************************************/
+/*                         This file is part of:                          */
+/*                             GODOT ENGINE                               */
+/*                        https://godotengine.org                         */
+/**************************************************************************/
+/* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+/* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#pragma once
+
+#include "core/variant/variant.h"
+
+template <typename T>
+class RequiredResult {
+	static_assert(!is_fully_defined_v<T> || std::is_base_of_v<Object, T>, "T must be an Object subtype");
+
+public:
+	using element_type = T;
+	using ptr_type = std::conditional_t<std::is_base_of_v<RefCounted, T>, Ref<T>, T *>;
+
+private:
+	ptr_type _value = ptr_type();
+
+	_FORCE_INLINE_ RequiredResult() = default;
+
+public:
+	RequiredResult(const RequiredResult &p_other) = default;
+	RequiredResult(RequiredResult &&p_other) = default;
+	RequiredResult &operator=(const RequiredResult &p_other) = default;
+	RequiredResult &operator=(RequiredResult &&p_other) = default;
+
+	_FORCE_INLINE_ RequiredResult(std::nullptr_t) :
+			RequiredResult() {}
+	_FORCE_INLINE_ RequiredResult &operator=(std::nullptr_t) { _value = nullptr; }
+
+	// These functions should not be called directly, they are only for internal use.
+	_FORCE_INLINE_ ptr_type _internal_ptr_dont_use() const { return _value; }
+	_FORCE_INLINE_ static RequiredResult<T> _err_return_dont_use() { return RequiredResult<T>(); }
+
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredResult(const RequiredResult<T_Other> &p_other) :
+			_value(p_other._value) {}
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredResult &operator=(const RequiredResult<T_Other> &p_other) {
+		_value = p_other._value;
+		return *this;
+	}
+
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredResult(T_Other *p_ptr) :
+			_value(p_ptr) {}
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredResult &operator=(T_Other *p_ptr) {
+		_value = p_ptr;
+		return *this;
+	}
+
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredResult(const Ref<T_Other> &p_ref) :
+			_value(p_ref) {}
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredResult &operator=(const Ref<T_Other> &p_ref) {
+		_value = p_ref;
+		return *this;
+	}
+
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredResult(Ref<T_Other> &&p_ref) :
+			_value(std::move(p_ref)) {}
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredResult &operator=(Ref<T_Other> &&p_ref) {
+		_value = std::move(p_ref);
+		return &this;
+	}
+
+	template <typename U = T, std::enable_if_t<std::is_base_of_v<RefCounted, U>, int> = 0>
+	_FORCE_INLINE_ RequiredResult(const Variant &p_variant) :
+			_value(static_cast<T *>(p_variant.get_validated_object())) {}
+	template <typename U = T, std::enable_if_t<std::is_base_of_v<RefCounted, U>, int> = 0>
+	_FORCE_INLINE_ RequiredResult &operator=(const Variant &p_variant) {
+		_value = static_cast<T *>(p_variant.get_validated_object());
+		return *this;
+	}
+
+	template <typename U = T, std::enable_if_t<!std::is_base_of_v<RefCounted, U>, int> = 0>
+	_FORCE_INLINE_ RequiredResult(const Variant &p_variant) :
+			_value(static_cast<T *>(p_variant.operator Object *())) {}
+	template <typename U = T, std::enable_if_t<!std::is_base_of_v<RefCounted, U>, int> = 0>
+	_FORCE_INLINE_ RequiredResult &operator=(const Variant &p_variant) {
+		_value = static_cast<T *>(p_variant.operator Object *());
+		return *this;
+	}
+
+	template <typename U = T, std::enable_if_t<std::is_base_of_v<RefCounted, U>, int> = 0>
+	_FORCE_INLINE_ element_type *ptr() const {
+		return *_value;
+	}
+
+	template <typename U = T, std::enable_if_t<!std::is_base_of_v<RefCounted, U>, int> = 0>
+	_FORCE_INLINE_ element_type *ptr() const {
+		return _value;
+	}
+
+	_FORCE_INLINE_ operator ptr_type() {
+		return _value;
+	}
+
+	_FORCE_INLINE_ operator Variant() const {
+		return Variant(_value);
+	}
+
+	_FORCE_INLINE_ element_type *operator*() const {
+		return ptr();
+	}
+
+	_FORCE_INLINE_ element_type *operator->() const {
+		return ptr();
+	}
+};
+
+template <typename T>
+class RequiredParam {
+	static_assert(!is_fully_defined_v<T> || std::is_base_of_v<Object, T>, "T must be an Object subtype");
+
+public:
+	using element_type = T;
+	using ptr_type = std::conditional_t<std::is_base_of_v<RefCounted, T>, Ref<T>, T *>;
+
+private:
+	ptr_type _value = ptr_type();
+
+	_FORCE_INLINE_ RequiredParam() = default;
+
+public:
+	// These functions should not be called directly, they are only for internal use.
+	_FORCE_INLINE_ ptr_type _internal_ptr_dont_use() const { return _value; }
+	_FORCE_INLINE_ bool _is_null_dont_use() const {
+		if constexpr (std::is_base_of_v<RefCounted, T>) {
+			return _value.is_null();
+		} else {
+			return _value == nullptr;
+		}
+	}
+	_FORCE_INLINE_ static RequiredParam<T> _err_return_dont_use() { return RequiredParam<T>(); }
+
+	// Prevent erroneously assigning null values by explicitly removing nullptr constructor/assignment.
+	RequiredParam(std::nullptr_t) = delete;
+	RequiredParam &operator=(std::nullptr_t) = delete;
+
+	RequiredParam(const RequiredParam &p_other) = default;
+	RequiredParam(RequiredParam &&p_other) = default;
+	RequiredParam &operator=(const RequiredParam &p_other) = default;
+	RequiredParam &operator=(RequiredParam &&p_other) = default;
+
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredParam(const RequiredParam<T_Other> &p_other) :
+			_value(p_other._internal_ptr_dont_use()) {}
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredParam &operator=(const RequiredParam<T_Other> &p_other) {
+		_value = p_other._internal_ptr_dont_use();
+		return *this;
+	}
+
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredParam(T_Other *p_ptr) :
+			_value(p_ptr) {}
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredParam &operator=(T_Other *p_ptr) {
+		_value = p_ptr;
+		return *this;
+	}
+
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredParam(const Ref<T_Other> &p_ref) :
+			_value(p_ref) {}
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredParam &operator=(const Ref<T_Other> &p_ref) {
+		_value = p_ref;
+		return *this;
+	}
+
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredParam(Ref<T_Other> &&p_ref) :
+			_value(std::move(p_ref)) {}
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredParam &operator=(Ref<T_Other> &&p_ref) {
+		_value = std::move(p_ref);
+		return &this;
+	}
+
+	template <typename U = T, std::enable_if_t<std::is_base_of_v<RefCounted, U>, int> = 0>
+	_FORCE_INLINE_ RequiredParam(const Variant &p_variant) :
+			_value(static_cast<T *>(p_variant.get_validated_object())) {}
+	template <typename U = T, std::enable_if_t<std::is_base_of_v<RefCounted, U>, int> = 0>
+	_FORCE_INLINE_ RequiredParam &operator=(const Variant &p_variant) {
+		_value = static_cast<T *>(p_variant.get_validated_object());
+		return *this;
+	}
+
+	template <typename U = T, std::enable_if_t<!std::is_base_of_v<RefCounted, U>, int> = 0>
+	_FORCE_INLINE_ RequiredParam(const Variant &p_variant) :
+			_value(static_cast<T *>(p_variant.operator Object *())) {}
+	template <typename U = T, std::enable_if_t<!std::is_base_of_v<RefCounted, U>, int> = 0>
+	_FORCE_INLINE_ RequiredParam &operator=(const Variant &p_variant) {
+		_value = static_cast<T *>(p_variant.operator Object *());
+		return *this;
+	}
+};
+
+#define TMPL_EXTRACT_PARAM_OR_FAIL(m_name, m_param, m_retval, m_msg, m_editor)                                                 \
+	if (unlikely(m_param._is_null_dont_use())) {                                                                               \
+		_err_print_error(FUNCTION_STR, __FILE__, __LINE__, "Required object \"" _STR(m_param) "\" is null.", m_msg, m_editor); \
+		return m_retval;                                                                                                       \
+	}                                                                                                                          \
+	typename std::decay_t<decltype(m_param)>::ptr_type m_name = m_param._internal_ptr_dont_use();                              \
+	static_assert(true)
+
+// These macros are equivalent to the ERR_FAIL_NULL*() family of macros, only for RequiredParam<T> instead of raw pointers.
+#define EXTRACT_PARAM_OR_FAIL(m_name, m_param) TMPL_EXTRACT_PARAM_OR_FAIL(m_name, m_param, void(), "", false)
+#define EXTRACT_PARAM_OR_FAIL_MSG(m_name, m_param, m_msg) TMPL_EXTRACT_PARAM_OR_FAIL(m_name, m_param, void(), m_msg, false)
+#define EXTRACT_PARAM_OR_FAIL_EDMSG(m_name, m_param, m_msg) TMPL_EXTRACT_PARAM_OR_FAIL(m_name, m_param, void(), m_msg, true)
+#define EXTRACT_PARAM_OR_FAIL_V(m_name, m_param, m_retval) TMPL_EXTRACT_PARAM_OR_FAIL(m_name, m_param, m_retval, "", false)
+#define EXTRACT_PARAM_OR_FAIL_V_MSG(m_name, m_param, m_retval, m_msg) TMPL_EXTRACT_PARAM_OR_FAIL(m_name, m_param, m_retval, m_msg, false)
+#define EXTRACT_PARAM_OR_FAIL_V_EDMSG(m_name, m_param, m_retval, m_msg) TMPL_EXTRACT_PARAM_OR_FAIL(m_name, m_param, m_retval, m_msg, true)
--- a/core/variant/variant_internal.h
+++ b/core/variant/variant_internal.h
@@ -608,6 +608,30 @@ struct VariantInternalAccessor<Object *> {
 	static _FORCE_INLINE_ void set(Variant *v, const Object *p_value) { VariantInternal::object_assign(v, p_value); }
 };
 
+template <class T>
+struct VariantInternalAccessor<RequiredParam<T>> {
+	static _FORCE_INLINE_ RequiredParam<T> get(const Variant *v) { return RequiredParam<T>(Object::cast_to<T>(const_cast<Object *>(*VariantInternal::get_object(v)))); }
+	static _FORCE_INLINE_ void set(Variant *v, const RequiredParam<T> &p_value) { VariantInternal::object_assign(v, p_value.ptr()); }
+};
+
+template <class T>
+struct VariantInternalAccessor<const RequiredParam<T> &> {
+	static _FORCE_INLINE_ RequiredParam<T> get(const Variant *v) { return RequiredParam<T>(Object::cast_to<T>(*VariantInternal::get_object(v))); }
+	static _FORCE_INLINE_ void set(Variant *v, const RequiredParam<T> &p_value) { VariantInternal::object_assign(v, p_value.ptr()); }
+};
+
+template <class T>
+struct VariantInternalAccessor<RequiredResult<T>> {
+	static _FORCE_INLINE_ RequiredResult<T> get(const Variant *v) { return RequiredResult<T>(Object::cast_to<T>(const_cast<Object *>(*VariantInternal::get_object(v)))); }
+	static _FORCE_INLINE_ void set(Variant *v, const RequiredResult<T> &p_value) { VariantInternal::object_assign(v, p_value.ptr()); }
+};
+
+template <class T>
+struct VariantInternalAccessor<const RequiredResult<T> &> {
+	static _FORCE_INLINE_ RequiredResult<T> get(const Variant *v) { return RequiredResult<T>(Object::cast_to<T>(*VariantInternal::get_object(v))); }
+	static _FORCE_INLINE_ void set(Variant *v, const RequiredResult<T> &p_value) { VariantInternal::object_assign(v, p_value.ptr()); }
+};
+
 template <>
 struct VariantInternalAccessor<Variant> {
 	static _FORCE_INLINE_ Variant &get(Variant *v) { return *v; }
--- a/scene/main/node.cpp
+++ b/scene/main/node.cpp
@@ -1696,11 +1696,11 @@ void Node::_add_child_nocheck(Node *p_child, const StringName &p_name, InternalM
 	emit_signal(SNAME("child_order_changed"));
 }
 
-void Node::add_child(Node *p_child, bool p_force_readable_name, InternalMode p_internal) {
+void Node::add_child(RequiredParam<Node> rp_child, bool p_force_readable_name, InternalMode p_internal) {
 	ERR_FAIL_COND_MSG(data.tree && !Thread::is_main_thread(), "Adding children to a node inside the SceneTree is only allowed from the main thread. Use call_deferred(\"add_child\",node).");
 
 	ERR_THREAD_GUARD
-	ERR_FAIL_NULL(p_child);
+	EXTRACT_PARAM_OR_FAIL(p_child, rp_child);
 	ERR_FAIL_COND_MSG(p_child == this, vformat("Can't add child '%s' to itself.", p_child->get_name())); // adding to itself!
 	ERR_FAIL_COND_MSG(p_child->data.parent, vformat("Can't add child '%s' to '%s', already has a parent '%s'.", p_child->get_name(), get_name(), p_child->data.parent->get_name())); //Fail if node has a parent
 #ifdef DEBUG_ENABLED
@@ -2635,7 +2635,7 @@ void Node::_propagate_replace_owner(Node *p_owner, Node *p_by_owner) {
 	data.blocked--;
 }
 
-Ref<Tween> Node::create_tween() {
+RequiredResult<Tween> Node::create_tween() {
 	ERR_THREAD_GUARD_V(Ref<Tween>());
 
 	SceneTree *tree = data.tree;
--- a/core/core_bind.cpp
+++ b/core/core_bind.cpp
@@ -165,7 +165,7 @@ void ResourceLoader::_bind_methods() {
 
 ////// ResourceSaver //////
 
-Error ResourceSaver::save(const Ref<Resource> &p_resource, const String &p_path, BitField<SaverFlags> p_flags) {
+Error ResourceSaver::save(RequiredParam<Resource> p_resource, const String &p_path, BitField<SaverFlags> p_flags) {
 	return ::ResourceSaver::save(p_resource, p_path, p_flags);
 }
 
--- a/core/core_bind.h
+++ b/core/core_bind.h
@@ -109,7 +109,7 @@ public:
 
 	static ResourceSaver *get_singleton() { return singleton; }
 
-	Error save(const Ref<Resource> &p_resource, const String &p_path, BitField<SaverFlags> p_flags);
+	Error save(RequiredParam<Resource> p_resource, const String &p_path, BitField<SaverFlags> p_flags);
 	Error set_uid(const String &p_path, ResourceUID::ID p_uid);
 	Vector<String> get_recognized_extensions(const Ref<Resource> &p_resource);
 	void add_resource_format_saver(Ref<ResourceFormatSaver> p_format_saver, bool p_at_front);
--- a/core/extension/extension_api_dump.cpp
+++ b/core/extension/extension_api_dump.cpp
@@ -88,7 +88,7 @@ static String get_property_info_type_name(const PropertyInfo &p_info) {
 }
 
 static String get_type_meta_name(const GodotTypeInfo::Metadata metadata) {
-	static const char *argmeta[13] = { "none", "int8", "int16", "int32", "int64", "uint8", "uint16", "uint32", "uint64", "float", "double", "char16", "char32" };
+	static const char *argmeta[14] = { "none", "int8", "int16", "int32", "int64", "uint8", "uint16", "uint32", "uint64", "float", "double", "char16", "char32", "required" };
 	return argmeta[metadata];
 }
 
--- a/core/extension/gdextension_interface.json
+++ b/core/extension/gdextension_interface.json
@@ -1942,6 +1942,10 @@
                 {
                     "name": "GDEXTENSION_METHOD_ARGUMENT_METADATA_INT_IS_CHAR32",
                     "value": 12
+                },
+                {
+                    "name": "GDEXTENSION_METHOD_ARGUMENT_METADATA_OBJECT_IS_REQUIRED",
+                    "value": 13
                 }
             ]
         },
--- a/core/io/resource_saver.cpp
+++ b/core/io/resource_saver.cpp
@@ -97,8 +97,8 @@ void ResourceFormatSaver::_bind_methods() {
 	GDVIRTUAL_BIND(_recognize_path, "resource", "path");
 }
 
-Error ResourceSaver::save(const Ref<Resource> &p_resource, const String &p_path, uint32_t p_flags) {
-	ERR_FAIL_COND_V_MSG(p_resource.is_null(), ERR_INVALID_PARAMETER, vformat("Can't save empty resource to path '%s'.", p_path));
+Error ResourceSaver::save(RequiredParam<Resource> rp_resource, const String &p_path, uint32_t p_flags) {
+	EXTRACT_PARAM_OR_FAIL_V_MSG(p_resource, rp_resource, ERR_INVALID_PARAMETER, vformat("Can't save empty resource to path '%s'.", p_path));
 	String path = p_path;
 	if (path.is_empty()) {
 		path = p_resource->get_path();
--- a/core/io/resource_saver.h
+++ b/core/io/resource_saver.h
@@ -83,7 +83,7 @@ public:
 		FLAG_REPLACE_SUBRESOURCE_PATHS = 64,
 	};
 
-	static Error save(const Ref<Resource> &p_resource, const String &p_path = "", uint32_t p_flags = (uint32_t)FLAG_NONE);
+	static Error save(RequiredParam<Resource> p_resource, const String &p_path = "", uint32_t p_flags = (uint32_t)FLAG_NONE);
 	static void get_recognized_extensions(const Ref<Resource> &p_resource, List<String> *p_extensions);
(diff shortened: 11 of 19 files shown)


AThousandShips:
One option would be to have this requirement be strict and crash on failure to fulfil it, I feel this is the area where that'd be appropriate, as if it's not so strict an error print should be used

Bromeon:
Very cool, and that was fast! 😎 

---


> Presently, it's marked by using `METADATA_OBJECT_IS_REQUIRED` in the same metadata field that holds things like `METADATA_INT_IS_UINT32`, which means it'll end up in the `extension_api.json` like this, for example:
>  ```
>  "arguments": [
>    {
>      "name": "node",
>      "type": "Node",
>      "meta": "required"
>    }
>  ]
>  ```

Could these metadata flags potentially overlap (maybe not now with the ints, but as we add more over time)? 
If yes, how would they be represented in the JSON value for the `meta` key?


---

> * If `nullptr` is passed into a `RequiredPtr<T>`, then it'll automatically create a new instance of the object in order to satisfy the "contract" (aka "auto-vivifying"), and print an error. I'm not sure this is the right thing to do. Since we don't have exceptions, we can't really do anything smarter at runtime. And code that works with `RequiredPtr<T>` will still need to use the error macros like `ERR_FAIL_NULL()`, because those will be able to make an error message with the file and line number of where the problem is. The error message when creating a `RequiredPtr<T>` has no idea what called it, again, because we don't have exceptions so we can't unwind the stack.

I'm probably missing something here, but if the expected usage is to always abort the function (i.e. print error + return null) after a "required" precondition fails, do we actually need to create a default instance?

Not too worried about performance in the error case, but some custom classes might have side effects in their constructors?

dsnopek:
@Bromeon:

> Could these metadata flags potentially overlap (maybe not now with the ints, but as we add more over time)?
> If yes, how would they be represented in the JSON value for the `meta` key?

I really have no idea what metadata could be added in the future. But like I said in the description, something more extensible like the the options you described in https://github.com/godotengine/godot-proposals/issues/2550 would probably be better.

> I'm probably missing something here, but if the expected usage is to always abort the function (i.e. print error + return null) after a "required" precondition fails, do we actually need to create a default instance?

Nope, you're not missing anything, we don't _need_ to create a default instance. I added that because I was worried that engine contributors could mistakenly believe that `RequiredPtr<T>` somehow guarantees that the value is not null, and this would prevent crashes in that case. We could just trust that contributors will know they need to still check for null? Or perhaps (as @AThousandShips writes above) a crash is actually what we want is this case?

I'm really not sure what the right behavior is, so that's what I'm hoping to discuss here :-)

AThousandShips:
I'd say it depends on how we interpret the argument limitation, is it *shouldn't* or *can't* be null?

From the proposal:
> Add a way to hint specific method arguments as being nullable. In other words, it means that the method will behave correctly if the specified arguments are null. If this hint is not present, the argument is assumed to be non-nullable.

One use for this would be to *remove* checks, at least in non-debug builds, the annotation states that "the responsibility of ensuring this is non-null is on the caller" and no checks needs to be done on the callee side

In the same sense the check can be used *before* calling any extension methods or before calling anything *from* an extension, and exiting there, acting as a barrier

I'd prefer it to be a guarantee for the method implementer, a "I can safely skip doing any null checks in this method, someone else handles it for me", or "this can safely be optimized in some sense", akin to the LLVM `nonnull`, as an optimization hint

Bromeon:
> I really have no idea what metadata could be added in the future. But like I said in the description, something more extensible like the the options you described in [godotengine/godot-proposals#2550](https://github.com/godotengine/godot-proposals/issues/2550) would probably be better.

It's hard to plan ahead here. Most flexible would probably be a dictionary (JSON object), but it might also be overkill?
```json5
{
    // ...
    "metadata": {
        "exact_type": "float64",
        "angle": "radian",
        "nullable": true
    }
}
```
Also, we will need to keep the `"meta"` field around anyway, for backwards compatibility. If nullability and other meta types don't overlap, we could technically reuse that key. But are we sure that nullability isn't something we might at some point consider for non-object parameters, as well?

---

What I'm a bit worried is that (in a few years), this might devolve into:
```json5
{
    // ...
    "meta": "float64|angle=radian|nullable"
}
```
That is, we have a "protocol inside a protocol" -- which would also be breaking btw, since existing code likely checks for the entire value string and would not expect any separators.

---

In other words, we have multiple options:
1) reuse the `meta` field and either
   - hope potential values will stay exclusive, even in the future
   - keep using it only for those values which are exclusive, add new fields like `angle` (just an example) for specific metadata
2) keep using `meta` only for the existing meaning.
   add a new key `nullable` (or maybe rather `required`, since it's new) on the same level
3) design something more future-proof like a `metadata` object or array, meaning
   - we have to provide both representations due to backwards compatibility
   - if we choose object instead of array, technically it's similar to today's `meta` key, but with 1 extra level of nesting

I tend to favor 1) or 2), but maybe you have more thoughts on this?

Bromeon:
> One use for this would be to _remove_ checks, at least in non-debug builds, the annotation states that "the responsibility of ensuring this is non-null is on the caller" and no checks needs to be done on the callee side

There are many examples of languages that didn't have a concept of nullability/optionality and retrofitted it late. Some I know:

1. Java: everything was nullable by default
    - `@NonNull` annotations to give stronger guarantees, `@Nullable` to confirm it's indeed nullable
    - `Optional<T>` which is a competing mechanism, more FP but semantically the same as the annotations
    - annoyingly, a simple `T` can still mean "possibly nullable or not, depending on API design"
2. PHP 7.1 [introduced `?Type` syntax](https://www.php.net/manual/en/migration71.new-features.php)
    - The good thing is that `Type` was strictly non-nullable before, so it was an easy addition
3. C++ with `std::optional`
    - This is probably the worst example... Because it came so late, everyone already had their own workarounds (null pointers, bool flags, hand-written optionals, unions, ...)
4. C# 8.0 introduced `string?` in addition to `string` to denote explicit nullability
   - Since existing code has used `string` in the sense "possibly null", this needs an opt-in, allowing compatibility
   - Not being a heavy C# user myself, this looks quite elegant from the outside. It allows modern APIs to become more expressive without breaking code.

Bromeon:
An idea I mentioned during today's GDExtension meeting is the **type-state pattern**, meaning that multiple _types_ are used to represent different _states_ of an object.

Concretely, we could split `RequiredPtr<T>` into two parts (names TBD):
- `RequiredPtr<T>` as the parameter type in the function signature
    - not yet verified if non-null
    - does not allow access to inner pointer/reference
- `VerifiedPtr<T>` inside the function
    - converted from `RequiredPtr<T>` via new macro similar to `ERR_FAIL_NULL`
    - can only be brought into existence if that macro succeeds
    - allows access to pointer and is always valid to dereference (no UB, no more checks)

The core idea here is that you cannot obtain `VerifiedPtr`, without going through one of the macros that verifies the non-null property. The macro returns from the current function if the check fails, making subsequent code safe.

dsnopek:
@Bromeon I finally got around to reworking this PR based on [your suggestion](https://github.com/godotengine/godot/pull/86079#issuecomment-2327313456) of using the "type-state pattern" from several months ago

This requires using new error macros to unpack the underlying pointer, which I think would also lead to encouraging more consistent use of the error macros for pointer parameters, which I think is something @AThousandShips would like per [her earlier comments](https://github.com/godotengine/godot/pull/86079#issuecomment-1852469711).

Please let me know what you think!

core/variant/method_ptrcall.h
```diff
@@ -179,6 +179,38 @@ struct PtrToArg<const T *> {
...
+template <class T>
+struct PtrToArg<RequiredPtr<T>> {
+	typedef T *EncodeT;
+
+	_FORCE_INLINE_ static T *convert(const void *p_ptr) {
+		if (p_ptr == nullptr) {
+			// Should we show an error?
+			return RequiredPtr<T>(nullptr);
+		}
+		return RequiredPtr<T>(const_cast<T *>(*reinterpret_cast<T *const *>(p_ptr)));
```
Bromeon, line 193: Is the `const_cast` needed here?

The `reinterpret_cast` returns a type `T *const *`, which is then dereferenced into `T* const`.
That is a top-level CV qualifier and can as such be directly cast into `T*`.

It's _not_ `T const*` (== `const T*`), which _would_ need a `const_cast`.

Bromeon (commented):
Thanks so much for this! ❤️ 

Added some first feedback.

It's also convenient that `required` meta is mutually exclusive with other metas, since it's an object and others are numbers 🙂

core/variant/method_ptrcall.h
dsnopek: Ah, I think you're right - I've removed the `const_cast` in my latest push

core/variant/required_ptr.h
```diff
@@ -0,0 +1,68 @@
...
+template <class T>
+class RequiredPtr {
+	T *value = nullptr;
+
+public:
+	typedef T *type;
+
+	_FORCE_INLINE_ RequiredPtr(T *p_value) {
+		value = p_value;
+	}
```
Repiteo, line 45: It'd be nice if there was a way to print an error on null *assignment*, as that would be much clearer to users/maintainers when an improper value is passed. That could be done with a wrapper macro of sorts for construction, but a more convenient option will be added in C++20 via `source_location`. Maybe that's outside the scope of this implementation, but I felt it worth mentioning all the same

Repiteo (commented):
A fantastic foundation for this paradigm, but one that still needs some refinement. You've already mentioned the lack of `Ref<T>` support, but there's still a few tricky edge-cases to be considered

scene/main/node.cpp
```diff
@@ -1644,11 +1644,11 @@ void Node::_add_child_nocheck(Node *p_child, const StringName &p_name, InternalM
 	emit_signal(SNAME("child_order_changed"));
 }
 
-void Node::add_child(Node *p_child, bool p_force_readable_name, InternalMode p_internal) {
+void Node::add_child(const RequiredPtr<Node> &rp_child, bool p_force_readable_name, InternalMode p_internal) {
```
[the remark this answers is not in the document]
Bromeon: As far as I understood, the idea here was to minimize the risk of causing any unintended changes. It's much easier to review just one line of code and have the confidence that nothing breaks.

Maybe the `rp_` -> `p_` change could still be done in follow-up PRs? These PRs would then become "pure" refactorings and would be completely isolated from semantic changes, which could make it easier to pinpoint the issue in case of regressions.

dsnopek:
I've been doing some testing in order to try and finish this up, which has included testing `RequiredPtr` as a return value, but it kind of breaks down.

Technically, it does work as a return value (I tried it with `Image::create_empty()`), and you can use it like:

```c++
EXTRACT_REQUIRED_PTR_OR_FAIL_V(new_image, Image::create_empty(width, height, false, Image::FORMAT_RGB8), Ref<Image>());
```

And that's all good, if the function that code is in returns `Ref<Image>`.

But if that function also returns `RequiredPtr<Image>` then that `EXTRACT_*()` line makes no sense, because it will make it so we now have a case that can return null, but the whole thing is silly to begin with because `Image::create_empty()` should be "guaranteed" to return a non-null image anyway.

Also, it's extremely difficult to find functions in Godot that are guaranteed to return non-null pointers or `Ref<T>`. And the ones I found (like `Image::create_empty()`), it's hard to say if that will always be true. Like, at the moment, `Image::create_empty()` has no branch that would return a null reference, but is that a purposeful contract that we want to guaranty? Or, did it just happen to turn out that way, and in the future we may want to return null under certain conditions (like a format that's not supported or width/height that are less than zero)?

@Bromeon Do you have Godot functions on your list where it makes sense to say the return value is required? It would be easier to reason about this with some concrete examples

Anyway, I'm thinking that perhaps we should limit this to parameters for now, and maybe rename it to `RequiredParam<T>` (which is a name that I think @Repiteo or someone else used on an earlier PR)?

If we want to guaranty some return values, I think it might need to be a different class, so that we could make sure its semantics made sense for return values. Like, in that case, maybe we error when creating the object, rather than when extracting the object? The use case is inherently different, because parameters can come from developers (including via GDScript), whereas with return values, it would 100% be an engine bug if we return `nullptr`

Bromeon:
> @Bromeon Do you have Godot functions on your list where it makes sense to say the return value is required? It would be easier to reason about this with some concrete examples

Yes, there are quite a few. I don't have a concrete list, but some examples that come to mind. A lot are factory methods or infallible getters

```js
Node::duplicate
Node::create_tween
Tween::tween_*

PhysicsDirectBodyState2D::get_space_state
PhysicsDirectBodyState3D::get_space_state
```

Not sure about methods that _could_ technically fail but shouldn't, like `Node::get_node` (as there's a fallible `get_node_or_null` already).

dsnopek:
@Bromeon Thanks!

However, looking into most of those cases, they actually can return null in certain situations:

- `Node::duplicate()`: It can return null if called from the wrong thread, and there's a bunch of weird cases I don't really understand where the node can fail to duplicate. Take a look at [`Node::_duplicate`](https://github.com/godotengine/godot/blob/0400b703e4c95e7de0ee98688267b98a5fb5e358/scene/main/node.cpp#L2830)
- `Node::create_tween()`: It can return null if called from the wrong thread, or if there's no `SceneTree` (which I've personally most commonly seen in automated tests using a custom main loop via the `--main-loop` argument)
- `Tween::tween_property()`: It can return null when running on a debug build and the property given by the `NodePath` doesn't exist on the given object.
- `Tween::tween_*()`: On all of them, they can return null if the Tween is no longer valid (like it finished) or it's already started. There's also some cases where if you pass in null arguments, they can return null, but I'm not counting those because I'm imagining those arguments would be marked as required and would be protected that way

For `PhysicsDirectBodyState2D::get_space_state()` and `PhysicsDirectBodyState3D::get_space_state()`, those _probably_ won't return null, because they are straight getters in the implementations that exist within Godot. However, if a physics server is added via GDExtension, there's nothing that would prevent it from returning null. So, I could see an argument for not marking those as required, because it's not a contract we really know if we can keep.

So, I'm still at a little bit of a loss for cases where it's _clearly_ guaranteed that a method won't return null... We could certainly choose to take a stand on certain methods and mark them a required, then, I guess, crash instead of returning null? Or, maybe we still return null but then the GDExtension binding could crash?

Bromeon:
Good to know, thanks for elaborating these! There seems to be a pattern, null is only returned for those methods in (rare?) error scenarios. Which is in line with Godot's overall behavior for failing variant-calls (e.g. due to wrong arguments, hitting an error macro, etc.), and the NIL variant is returned.

I wonder if it could still be beneficial to annotate those return types as required.

This would mean _required_ in return type position would not be a hard guarantee, but it would still provide useful information in the majority of cases. A binding can then generate code that is more type-safe/ergonomic for all reasonable usages -- however, it must somehow deal with the occasional null being returned. Valid error handling strategies might then be:
- exceptions (C++, C#, Kotlin, ...)
- panics (Rust)
- return null (C++, Go, ...) i.e. just ignore the _required_ annotation
- UB for very perf-critical applications or certain Release builds

dsnopek:
Yeah, that could make sense!

That would also mean there are definitely semantic differences between how we handle arguments and return values, where for arguments it's a "hard" requirement and you need to use the extraction macros, but for return values it's a "soft" requirement.

This further makes me think we may need separate `RequiredParam<T>` and `RequiredReturn<T>` (or `RequiredValue<T>`? naming is hard) classes...

Bromeon:
> If we want to guaranty some return values, I think it might need to be a different class, so that we could make sure its semantics made sense for return values. Like, in that case, maybe we error when creating the object, rather than when extracting the object? The use case is inherently different, because parameters can come from developers (including via GDScript), whereas with return values, it would 100% be an engine bug if we return `nullptr`

> This further makes me think we may need separate `RequiredParam<T>` and `RequiredReturn<T>` (or `RequiredValue<T>`? naming is hard) classes...

If you think there's a benefit to having separate classes for parameters and return values, I'd definitely be OK with that! Question is whether there would be a lot of code duplication, or quite a bit could be shared.

So you'd suggest to first get parameters working, and then we can look into return types? Do you think the latter are still realistic for the 4.6 cycle? I'm asking because a lot of the APIs like tweens are very unergonomic to use with explicit null checks, to the point that people have written [custom libraries](https://github.com/Houtamelo/spire_tween?tab=readme-ov-file#main-reason-ergonomics) to make it easier 🙂 (not just for nulls, but the extra `unwrap`s are indeed an issue).

dsnopek:
> If you think there's a benefit to having separate classes for parameters and return values, I'd definitely be OK with that! Question is whether there would be a lot of code duplication, or quite a bit could be shared.

There would be some duplication, but they will act differently enough that I think it's justified to have them separate.

In my latest push, I've taken a stab at making a `RequiredValue` for return values. I still need to do some more testing with it, though.

> So you'd suggest to first get parameters working, and then we can look into return types? Do you think the latter are still realistic for the 4.6 cycle?

Well, I suggested it because I feel very confident about the parameters part. But if we're making return values a "soft" requirement like discussed above, they should be pretty simple, and we can try and tackle that in this same PR.

dsnopek:
Alright, I think I've got these changes where I want them to be! I'd still like to do some more testing by marking more APIs with required and making sure everything still works, but I think this can come out of DRAFT and get some wider review :-)

Regarding the two CI failures:
- One is unrelated (it's in `master` too) and will be fixed by https://github.com/godotengine/godot/pull/111771
- The other is because godot-cpp needs to be updated for this change, which will be fixed by https://github.com/godotengine/godot-cpp/pull/1868

@Bromeon Can you do some testing with this using the Rust bindings?

dsnopek:
CI is passing now!

Bromeon:
Thanks a lot for the great work so far!
So, I implemented this in Rust: https://github.com/godot-rust/gdext/pull/1383

Codegen works nicely! :+1: 

---

When testing the invocation, I ran into UB when calling `Node::create_tween()`:
```c
Engine version: Godot Engine v4.6.dev.custom_build (9ccb329effa3aaac6d0e0ab3b157b6f02b099926)
Dumping the backtrace. Please include this when reporting the bug on: https://github.com/godotengine/godot/issues
[1] /lib64/libc.so.6(+0x1a070) [0x7f88079ef070] (??:0)
[2] List<Ref<Tweener>, DefaultAllocator>::front() (/.../godot/core/templates/list.h:269)
[3] List<Ref<Tweener>, DefaultAllocator>::clear() (/.../godot/core/templates/list.h:468 (discriminator 1))
[4] List<Ref<Tweener>, DefaultAllocator>::~List() (/.../godot/core/templates/list.h:702)
[5] void LocalVector<List<Ref<Tweener>, DefaultAllocator>, unsigned int, false, false>::_resize<true>(unsigned int) (/.../godot/core/templates/local_vector.h:57 (discriminator 3))
[6] LocalVector<List<Ref<Tweener>, DefaultAllocator>, unsigned int, false, false>::resize(unsigned int) (/.../godot/core/templates/local_vector.h:191)
[7] LocalVector<List<Ref<Tweener>, DefaultAllocator>, unsigned int, false, false>::clear() (/.../godot/core/templates/local_vector.h:153)
[8] Tween::clear() (/.../godot/scene/animation/tween.cpp:222)
[9] SceneTree::finalize() (/.../godot/scene/main/scene_tree.cpp:867 (discriminator 8))
[10] OS_LinuxBSD::run() (/.../godot/platform/linuxbsd/os_linuxbsd.cpp:996)
[11] ...
```
I could also reproduce this when calling godot-rust `master` with this revision, so it's not related to my changes.
There, I additionally get this error from Godot:
```c
ERROR: Trying to unreference a SafeRefCount which is already zero is wrong and a symptom of it being misused.
Upon a SafeRefCount reaching zero any object whose lifetime is tied to it, as well as the ref count itself, must be destroyed.
Moreover, to guarantee that, no multiple threads should be racing to do the final unreferencing to zero.
   at: _check_unref_safety (./core/templates/safe_refcount.h:186)

handle_crash: Program crashed with signal 4
Engine version: Godot Engine v4.6.dev.custom_build (9ccb329effa3aaac6d0e0ab3b157b6f02b099926)
Dumping the backtrace. Please include this when reporting the bug on: https://github.com/godotengine/godot/issues
[1] /lib64/libc.so.6(+0x1a070) [0x7f8e42ff3070] (??:0)
[2] RefCounted::unreference() (/.../godot/core/templates/safe_refcount.h:186 (discriminator 5))
[3] Ref<Tween>::unref() (/.../godot/core/object/ref_counted.h:207 (discriminator 1))
[4] Ref<Tween>::~Ref() (/.../godot/core/object/ref_counted.h:225)
[5] List<Ref<Tween>, DefaultAllocator>::Element::~Element() (/.../godot/core/templates/list.h:52)
[6] void memdelete_allocator<List<Ref<Tween>, DefaultAllocator>::Element, DefaultAllocator>(List<Ref<Tween>, DefaultAllocator>::Element*) (/.../godot/core/os/memory.h:154)
[7] List<Ref<Tween>, DefaultAllocator>::_Data::erase(List<Ref<Tween>, DefaultAllocator>::Element*) (/.../godot/core/templates/list.h:249)
[8] List<Ref<Tween>, DefaultAllocator>::erase(List<Ref<Tween>, DefaultAllocator>::Element*) (/.../godot/core/templates/list.h:436)
[9] SceneTree::process_tweens(double, bool) (/.../godot/scene/main/scene_tree.cpp:835)
[10] SceneTree::process(double) (/.../godot/scene/main/scene_tree.cpp:719)
[11] Main::iteration() (/.../godot/main/main.cpp:4748 (discriminator 3))
[12] OS_LinuxBSD::run() (/.../godot/platform/linuxbsd/os_linuxbsd.cpp:991 (discriminator 1))
[13] ...
```

`RequiredValue<T>` seems to behave differently than `Ref<T>` -- does it increment the refcount before returning the object to the caller? The above looks like UAF...

---

> This further makes me think we may need separate `RequiredParam<T>` and `RequiredReturn<T>` (or `RequiredValue<T>`? naming is hard) classes...

Back to this: I would vote for calling it `RequiredReturn`, simply because it's quite possible that contributors -- especially when new to the "required" concept -- would declare parameters as `RequiredValue<SomeType>`. On the other hand, `RequiredReturn` in parameter position looks more like a mistake.

dsnopek:
> When testing the invocation, I ran into UB when calling `Node::create_tween()`

Thanks for testing and finding this issue! I'm not entirely sure how I didn't notice it in my testing originally :-/

I just pushed some changes that I think should fix it - please let me know if it works for you.

> Back to this: I would vote for calling it `RequiredReturn`, simply because it's quite possible that contributors -- especially when new to the "required" concept -- would declare parameters as `RequiredValue<SomeType>`. On the other hand, `RequiredReturn` in parameter position looks more like a mistake.

Maybe we could call it `RequiredReturnValue`? Although, that's maybe getting a little long...

Personally, I just don't like `RequiredReturn` from a English language grammatical perspective. The word "return" in this usage isn't a noun, it's a verb - you don't "return a return", you "return a value", or to get extra specific, you "return a return value". The word "return" _can_ be used as a noun (ex. "I'm awaiting his return") but that's a little old-timey and I don't know that I've ever seen it used this way in the context of programming; the noun that I've always seen used is "return value"

But, yeah, this sort of thing is always super subjective - I wouldn't be surprised to find another native English speaker from somewhere else that may feel differently about this :-) so perhaps it doesn't matter too much

Bromeon:
> Thanks for testing and finding this issue! I'm not entirely sure how I didn't notice it in my testing originally :-/

I think it's easy to miss, but I wonder why CI didn't catch it. Maybe it makes sense to add some smoke tests for known APIs? Either in core code (if that can trigger this sort of issue) or the godot-cpp CI.

In Rust I simply used this:
```rs
let mut parent = Node::new_alloc();
let child = Node::new_alloc();
parent.add_child(&child);

let tween: Gd<Tween> = parent.create_tween();
assert!(tween.is_instance_valid());
assert!(tween.to_string().contains("Tween"));

parent.free();
```

---

> Personally, I just don't like `RequiredReturn` from a English language grammatical perspective.

`RequiredResult` maybe?

dsnopek:
> Maybe it makes sense to add some smoke tests for known APIs? Either in core code (if that can trigger this sort of issue) or the godot-cpp CI.

Makes sense to me! I've made a PR to add such a test to godot-cpp's tests:

https://github.com/godotengine/godot-cpp/pull/1875

> `RequiredResult` maybe?

That could work! Let's discuss at the GDExtension team meeting before I update this PR, though

dsnopek:
Per our discussion at today's GDExtension meeting, I've renamed `RequiredValue` to `RequiredResult`

Bromeon (approved):
✅ Re-tested 17028c64d4d8e5c303cbbf768547686c12b3c2a2 with godot-rust, all our tests pass.

This looks good from my side!

core/variant/required_ptr.h
```diff
@@ -0,0 +1,258 @@
...
+	}                                                                                                                          \
+	std::conditional_t<                                                                                                        \
+			std::is_base_of_v<RefCounted, std::decay_t<decltype(m_param)>::element_type>,                                      \
+			Ref<std::decay_t<decltype(m_param)>::element_type>,                                                                \
+			std::decay_t<decltype(m_param)>::element_type *>                                                                   \
+			m_name = m_param._internal_ptr();                                                                                  \
+	GODOT_DEPRECATED_END                                                                                                       \
+	static_assert(true)
+
+#define EXTRACT_REQUIRED_PARAM_OR_FAIL(m_name, m_param) TMPL_EXTRACT_REQUIRED_PARAM_OR_FAIL(m_name, m_param, void(), "", false)
```
Ivorforce: If we wanted to make this shorter, here's two ideas:

- Use `EXTRACT_PARAM_OR_FAIL`, probably still gets the same meaning across?
- Only pass in `m_name`, `m_param` is auto-identified as `r##m_name`

The latter comes with the downside that the `m_param` (e.g. `rp_param`) is not explicitly mentioned in the macro call, which could be confusing. But it would reduce the complexity / number of params of all calls, and "enforce" the naming scheme.
 Not sure if I would do it.

core/error/error_macros.h
```diff
@@ -175,7 +175,9 @@ void _physics_interpolation_warning(const char *p_function, const char *p_file,
 #define ERR_FAIL_INDEX_V(m_index, m_size, m_retval)                                                             \
 	if (unlikely((m_index) < 0 || (m_index) >= (m_size))) {                                                     \
 		_err_print_index_error(FUNCTION_STR, __FILE__, __LINE__, m_index, m_size, _STR(m_index), _STR(m_size)); \
+		GODOT_DEPRECATED_BEGIN                                                                                  \
```
Ivorforce: All the calls `of GODOT_DEPRECATED_*` are a bit... unfortunate, but I get it. I'm not sure I can think of a good way to avoid it.

core/variant/method_ptrcall.h
```diff
@@ -269,6 +269,51 @@ struct PtrToArg<const T *> {
 	}
 };
 
+// This is for RequiredParam.
+
+template <class T>
+struct PtrToArg<RequiredParam<T>> {
+	typedef typename RequiredParam<T>::value_type EncodeT;
```
Ivorforce: I prefer using `using` for C++ reasons (it's compatible with more contexts and I think less error prone too).

```suggestion
	using EncodeT = RequiredParam<T>::value_type;
```

core/variant/method_ptrcall.h
```diff
@@ -269,6 +269,51 @@ struct PtrToArg<const T *> {
...
+
+	_FORCE_INLINE_ static RequiredParam<T> convert(const void *p_ptr) {
+		if (p_ptr == nullptr) {
+			// Should we show an error?
```
Ivorforce: Not sure I get the comment, should we?

core/variant/method_ptrcall.h
```diff
@@ -269,6 +269,51 @@ struct PtrToArg<const T *> {
...
+			GODOT_DEPRECATED_BEGIN
+			return RequiredParam<T>::err_return();
+			GODOT_DEPRECATED_END
+		}
+		return RequiredParam<T>(*reinterpret_cast<T *const *>(p_ptr));
```
Ivorforce, line 282: `reinterpret_cast` can be error prone in this context (mostly academical, since I think it only applies to virtual inheritance?), prefer using `const_cast` and `static_cast`.
You should be able to just cast like this I think:

```suggestion
		return RequiredParam<T>(*static_cast<const T *const *>(p_ptr));
```

(which should convert to `const T * const &`, which auto-copies to `T *`, I think?)

core/variant/method_ptrcall.h
```diff
@@ -269,6 +269,51 @@ struct PtrToArg<const T *> {
...
+struct PtrToArg<RequiredResult<T>> {
+	typedef typename RequiredResult<T>::value_type EncodeT;
+
+	_FORCE_INLINE_ static RequiredResult<T> convert(const void *p_ptr) {
+		if (p_ptr == nullptr) {
+			GODOT_DEPRECATED_BEGIN
+			return RequiredResult<T>::err_return();
+			GODOT_DEPRECATED_END
+		}
+		return RequiredResult<T>(*reinterpret_cast<T *const *>(p_ptr));
```
Ivorforce, line 300: Same here

core/variant/required_ptr.h
```diff
@@ -0,0 +1,258 @@
...
+
+#pragma once
+
+#include "core/variant/variant.h"
+
+template <typename T>
+class RequiredResult {
+	// We can't explicitly check if the type is an Object here, as it might've been forward-declared.
+	//  However, we can at least assert that it's an unqualified, non-builtin type.
+	static_assert(!std::is_scalar_v<T>, "T may not be a scalar type");
```
Ivorforce: Try this instead, it's a bit unusual but should address the problem better i think, since it will be fully defined sooner or later when used:


```suggestion
	static_assert(!is_fully_defined_v<T> || std::is_base_of_v<Object, T>, "T must be an Object subtype");

File: core/variant/required_ptr.h [hunk 11]
```diff
@@ -0,0 +1,258 @@
...
+	static_assert(!std::is_reference_v<T>, "T may not be a reference type");
+	static_assert(!std::is_const_v<T> && !std::is_volatile_v<T>, "T may not be cv-qualified");
+
+public:
+	using element_type = T;
+	using value_type = std::conditional_t<std::is_base_of_v<RefCounted, T>, Ref<T>, T *>;
```
Ivorforce: "Value" is ambiguous here I think, more than `ptr` anyway

```suggestion
	using ptr_type = std::conditional_t<std::is_base_of_v<RefCounted, T>, Ref<T>, T *>;
```

core/variant/required_ptr.h
```diff
@@ -0,0 +1,258 @@
...
+
+private:
+	value_type _value;
+
+	_FORCE_INLINE_ RequiredResult() {
+		if constexpr (!std::is_base_of_v<RefCounted, T>) {
+			_value = nullptr;
+		}
```
Ivorforce: This should work too:

```suggestion
		// Initializes to nullptr
		_value = value_type();
```

if value_type is `T *`, calling its constructor initializes to `nullptr`. I've been using this "trick" in `CowData` and similar, to avoid `if constexpr` complexity.

core/variant/required_ptr.h
```diff
@@ -0,0 +1,258 @@
...
+
+public:
+	RequiredResult(const RequiredResult &p_other) = default;
+	RequiredResult(RequiredResult &&p_other) = default;
+	RequiredResult &operator=(const RequiredResult &p_other) = default;
+	RequiredResult &operator=(RequiredResult &&p_other) = default;
+
+	_FORCE_INLINE_ RequiredResult(std::nullptr_t) :
+			RequiredResult() {}
+	_FORCE_INLINE_ RequiredResult &operator=(std::nullptr_t) { _value = nullptr; }
```
Ivorforce, line 56: Doesn't this fail compiling if `_value` is `Ref`? I thought the code in there forbids `nullptr` assignments.

core/variant/required_ptr.h
```diff
@@ -0,0 +1,258 @@
...
+
+	[[nodiscard, deprecated("Should not be called directly; only for internal use.")]]
+	_FORCE_INLINE_ value_type _internal_ptr() const { return _value; }
+	[[nodiscard, deprecated("Should not be called directly; only for internal use.")]]
+	_FORCE_INLINE_ static RequiredResult<T> err_return() { return RequiredResult<T>(); }
```
Ivorforce: Could we make it non-deprecated, and just call it `_internal_ptr_dont_call` or something? That should avoid all the `GODOT_DEPRECATED_*` complexity above. If we add a comment _why_ it shouldn't be called, that would be the dumber (simpler) solution I think.

core/variant/required_ptr.h
```diff
@@ -0,0 +1,258 @@
...
+			_value(p_other._value) {}
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredResult &operator=(const RequiredResult<T_Other> &p_other) {
+		_value = p_other._value;
+		return *this;
+	}
+
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredResult(const T_Other *p_ptr) :
+			_value(const_cast<std::remove_const_t<T_Other> *>(p_ptr)) {}
```
Ivorforce: Why is the `remove_const` needed here? 

Also, is this constructor sane? Can't you upgrade from const to non-const using this constructor?

I understand that it's annoying to have to use `RequiredResult<const T>` but I'm not sure if we can sanely avoid it.

core/variant/required_ptr.h
```diff
@@ -0,0 +1,258 @@
...
+	_FORCE_INLINE_ RequiredParam(Ref<T_Other> &&p_ref) :
+			_value(std::move(p_ref)) {}
+	template <typename T_Other, std::enable_if_t<std::is_base_of_v<T, T_Other>, int> = 0>
+	_FORCE_INLINE_ RequiredParam &operator=(Ref<T_Other> &&p_ref) {
+		_value = std::move(p_ref);
+		return &this;
+	}
+
+	_FORCE_INLINE_ RequiredParam(const Variant &p_variant) :
+			_value(static_cast<T *>(p_variant.get_validated_object())) {}
```
Ivorforce, line 215: Using `get_validated_object` sounds like a sane default, but it's less fast than the usual `Variant` conversion to `T *`, which is unchecked. 
I'd be in favour of this decision, but it _is_ a decision.

core/variant/required_ptr.h
Ivorforce: This could use a comment that these are equivalent to the `ERR_FAIL` macros, just with required params.

core/variant/required_ptr.h
```diff
@@ -0,0 +1,258 @@
...
+	static_assert(true)
+
+#define EXTRACT_REQUIRED_PARAM_OR_FAIL(m_name, m_param) TMPL_EXTRACT_REQUIRED_PARAM_OR_FAIL(m_name, m_param, void(), "", false)
+#define EXTRACT_REQUIRED_PARAM_OR_FAIL_MSG(m_name, m_param, m_msg) TMPL_EXTRACT_REQUIRED_PARAM_OR_FAIL(m_name, m_param, void(), m_msg, false)
+#define EXTRACT_REQUIRED_PARAM_OR_FAIL_EDMSG(m_name, m_param, m_msg) TMPL_EXTRACT_REQUIRED_PARAM_OR_FAIL(m_name, m_param, void(), m_msg, true)
+#define EXTRACT_REQUIRED_PARAM_OR_FAIL_V(m_name, m_param, m_retval) TMPL_EXTRACT_REQUIRED_PARAM_OR_FAIL(m_name, m_param, m_retval, "", false)
+#define EXTRACT_REQUIRED_PARAM_OR_FAIL_V_MSG(m_name, m_param, m_retval, m_msg) TMPL_EXTRACT_REQUIRED_PARAM_OR_FAIL(m_name, m_param, m_retval, m_msg, false)
+#define EXTRACT_REQUIRED_PARAM_OR_FAIL_V_EDMSG(m_name, m_param, m_retval, m_msg) TMPL_EXTRACT_REQUIRED_PARAM_OR_FAIL(m_name, m_param, m_retval, m_msg, true)
+
+// Technically zero-constructible, but not recommended.
```
Ivorforce: Why not recommended? This can't really be communicated through the struct.

core/variant/variant_internal.h
```diff
@@ -590,24 +590,36 @@ struct VariantInternalAccessor<IPAddress> {
...
-};
-
 template <>
 struct VariantInternalAccessor<Object *> {
 	static _FORCE_INLINE_ Object *get(const Variant *v) { return const_cast<Object *>(*VariantInternal::get_object(v)); }
 	static _FORCE_INLINE_ void set(Variant *v, const Object *p_value) { VariantInternal::object_assign(v, p_value); }
 };
 
+template <class T>
+struct VariantInternalAccessor<RequiredParam<T>> {
```
Ivorforce, line 612: Why do these need internal accessors? For `PtrToArg`?

Ivorforce (commented):
This sure is high in C++ magic complexity 😄

After all your feedback and discussion, I agree it's the right way forward. And I think it's close to mergable too. Here's a small batch of feedback.

core/variant/method_ptrcall.h
dsnopek: Hm, all the other `PtrToArg` structs use `typedef` - maybe this could be cleaned up all at once in another PR?

core/variant/method_ptrcall.h
dsnopek: I agree that `static_cast` is better than `reinterpret_cast`! However, all of the other `PtrToArg` structs use `reinterpret_cast` (and this specific cast is the same as the one for converting `Object *`) - maybe this could be cleaned up all at once in another PR?

core/variant/required_ptr.h
dsnopek: It definitely doesn't fail compiling - this would apply to all the `Tween` conversions, which build fine

core/variant/method_ptrcall.h
dsnopek: Me neither - I'll remove it :-)

core/variant/method_ptrcall.h
Ivorforce: Makes sense.

core/variant/required_ptr.h
dsnopek: We could, I think I even had something like that in an earlier version of the PR. @Repiteo did all the deprecated stuff - that came out of a review from him back in the spring, I think?

@Repiteo how do you feel about keeping/removing all the deprecated stuff?

core/variant/required_ptr.h
dsnopek: Hm, I don't remember adding this. I would personally probably lean to the other side and do the unvalidated conversion to try and make this as "no overhead" as possible

core/variant/required_ptr.h
dsnopek: Yeah, I guess `EXTRACT_PARAM_OR_FAIL()` could work! The type itself already says it's required. We should maybe get a little more input before switching it, though

But I'm not crazy about the automatic variable name idea - it's a little more magical than I personally like. I know there was a request for more magic :-) but my preference is to be more mundane

core/variant/required_ptr.h
dsnopek: I don't know, I think this may have been another bit from @Repiteo?

core/variant/variant_internal.h
dsnopek: It's used in [`binder_common.h`](https://github.com/dsnopek/godot/blob/required-args/core/variant/binder_common.h) which is part of creating method binds. For example:

```
template <typename T, typename... P, size_t... Is>
void call_with_validated_variant_args_helper(T *p_instance, void (T::*p_method)(P...), const Variant **p_args, IndexSequence<Is...>) {
	(p_instance->*p_method)((VariantInternalAccessor<GetSimpleTypeT<P>>::get(p_args[Is]))...);
}
```

core/variant/required_ptr.h
Repiteo: This was back when there was only a single struct: `RequiredPtr`. It wasn't recommended because this being zero-constructed would bypass the non-nullable initialization safeguards, so it should be done with care

dsnopek: Should we just remove this whole bit? Do we really gain anything from it? I'm actually not sure that `RequiredParam` or `RequiredReturn` are zero-constructible in the case that they hold a `RefCounted`, but we could probably limit this to the non-`RefCounted` ones if needed

core/variant/required_ptr.h
dsnopek: It looks like we don't need it here - seems to be copy-pasted from `RequiredParam`, which does need it for some reason (compilation fails without it). I'll see if I can figure out what exactly that is about

This is the problem with such a long-running PR: I don't really remember the reasons behind everything :-)

core/variant/required_ptr.h
Repiteo: I generally really dislike the "please don't use!!!" syntax of exposed methods, and a deprecated annotation felt like the best way of enforcing that in the least-ugly way. But it's also wholly stylistic, and not a hill I'll die on

core/variant/required_ptr.h
Repiteo: It's in case the `T_Other` is inherently `const`, which doesn't have the same guaranteed safeguards as the struct itself

And it's not sane, but the very unfortunate reality is that neither is the overwhelming majority of the codebase when it comes to `Object *`. There's extremely few safeguards for `const` pointers of objects, almost always defaulting to a non-const version (this too is `Variant`'s fault). Trying to enforce it even slightly in smaller applications has very annoying and obstructing ramifications that wasn't worth bikeshedding the rest of the PR

core/variant/required_ptr.h
Repiteo: This was pulled from how `Ref` handles construction from `Variant`

core/variant/required_ptr.h
dsnopek: Hm, so should I have saved that on `RequiredResult`? I already removed it because it compiled without it - but I can bring it back if it's really needed

core/variant/variant_internal.h
Repiteo: A good chunk of `VariantInternalAccessor` has become redundant with native `Variant` conversions, but unfortunately that doesn't handle the `Object *` cases yet. It's very much required for now

core/variant/required_ptr.h
dsnopek: Actually, maybe I messed up when testing with `RequiredParam` earlier - it seems to be working without the `remove_const` now!

I'm just going to remove it on `RequiredParam` too for now, but I can bring it later if needed

core/variant/required_ptr.h
Repiteo: Back when this was just `RequiredPtr`, these were explicitly deleted. They don't serve much of a purpose on their own

Repiteo (dismissed):
I just realized that the structs *themselves* conditionally check for their internal value. As much as I wish we could do that, that would completely prevent the use of this struct for forward-declared types or declarations in functions declaring that very type. This is why the static assertions for the structs are generic

I'm really sorry I didn't check the altered PR earlier, this could've been caught much sooner

core/variant/required_ptr.h
Repiteo: As long as this contains `Ref` directly, it can be removed

dsnopek: Removed!

core/variant/required_ptr.h
Repiteo: Magicing the name would mean locking us into the `rp_*` syntax, which I'm gonna hard-pass on.

`EXTRACT_PARAM_OR_FAIL()` however, I can get behind

core/variant/required_ptr.h
Repiteo: This has a niche downside of locking in this particular `is_fully_defined_v<T>` value now, which could provide a false positive if using the assertion macro with this type later. But that's not anywhere close to an immediate concern

core/variant/required_ptr.h
Repiteo: Equivalent *except* that they define a non-scoped value, meaning they don't have a full wrapper safety net. Not that it's very relevant in their case, as nobody would be making this call without intending to use the value

core/variant/required_ptr.h
dsnopek: > They don't serve much of a purpose on their own

Actually, I think we do need these explicit conversions from `nullptr_t`, because all of the other conversions come with some `is_base_of_v<T>` which will never match when `nullptr` is used. And, even though the return value is "required", we've defined that to mean `nullptr` is an error, so we do still need some way to convert from `nullptr`.

core/variant/required_ptr.h
dsnopek: Ah, ok, that does make sense!

So, if we didn't do this, then we'd actually be _removing_ the validated conversion when we replace a `Ref<T>` with `RequiredParam<T>`. But it does mean we are adding a check and potential overhead when converting a `Object *` to `RequiredParam<T>`

But we can solve that with more `std::is_base_of_v()`! I've changed this to only do the validated conversion for `RefCounted`, and skip it otherwise

core/variant/method_ptrcall.h
Ivorforce: That's fair. Let's keep it like this then.

core/variant/required_ptr.h
Ivorforce: As clever as the `[[deprecated]]` trick is, i dislike the complexity it adds just for people not to call it (e.g. all the `GODOT_DEPRECATED_*` stuff). Beyond the mental overhead, it might also affect compile time and make our builds more error prone (because we're relying on more tricks).
So i'd prefer the simpler solution.

core/variant/variant_internal.h
Ivorforce: Alright. It's a little weird to view a `RequiredPtr` as a means to introspect into a `Variant`, but I can accept it being here if we need it.

core/variant/required_ptr.h
Ivorforce: Yea, it's definitely not fool proof, but I think it should be good enough.

core/variant/required_ptr.h
Repiteo: I'll happily concede to the simpiler use-case then!

@dsnopek When removing the deprecated wrappers, go ahead and remove the `GODOT_DEPRECATED_BEGIN`/`GODOT_DEPRECATED_END` macros from `typedefs.h` entirely; they no longer serve a use

core/variant/required_ptr.h
dsnopek: Alright, I've decided to go for it! In my latest push, I've switched to the somewhat shorted `EXTRACT_PARAM_OR_FAIL()`

core/variant/required_ptr.h
dsnopek: I've removed the deprecated stuff in my latest push

dsnopek:
@Ivorforce All your previous feedback should be addressed at this point!

core/variant/required_ptr.h
```diff
@@ -0,0 +1,258 @@
...
+	if (unlikely(m_param._is_null_dont_use())) {                                                                               \
+		_err_print_error(FUNCTION_STR, __FILE__, __LINE__, "Required object \"" _STR(m_param) "\" is null.", m_msg, m_editor); \
+		return m_retval;                                                                                                       \
+	}                                                                                                                          \
+	std::conditional_t<                                                                                                        \
+			std::is_base_of_v<RefCounted, std::decay_t<decltype(m_param)>::element_type>,                                      \
+			Ref<std::decay_t<decltype(m_param)>::element_type>,                                                                \
+			std::decay_t<decltype(m_param)>::element_type *>                                                                   \
+			m_name = m_param._internal_ptr_dont_use();                                                                         \
```
Ivorforce: Might also be able to piggy-back off of `RequiredParam`

```suggestion
	typename decltype(m_param)::ptr_type m_name = m_param._internal_ptr_dont_use();                                                                         \
```

Ivorforce (approved):
Looks good to me.

Besides the Godot gotcha this adds, I have no concerns (correctness, performance, ambiguity) over this PR. After our discussions at the GDExtension meetings, I agree we should move forward with this.

Great work!

core/object/object.h
```diff
@@ -39,6 +39,7 @@
 #include "core/templates/hash_set.h"
 #include "core/templates/list.h"
 #include "core/templates/safe_refcount.h"
+#include "core/variant/required_ptr.h"
```
AThousandShips, line 42: Is this strictly necessary to include in `object.h`, it'll make this header cause mass rebuilds when changed

core/variant/required_ptr.h
dsnopek: Ah, good idea! I've updated this in my latest push

core/object/object.h
dsnopek: I don't think it's _strictly_ necessary (we'll probably never use `RequiredParam`/`RequiredResult` in `object.h` itself), but I do think it makes sense.

It's likely that we'll use those in lots and lots of places (including in `node.h`, where it's used already in one of the "example changes"), and this prevents us having to add `required_ptr.h` all over the place.

Also, I don't think we'll change `required_ptr.h` very much - maybe only in this initial introduction period if we got something wrong, but then it should remain largely untouched.

Repiteo:
Thanks!

PR #86076, opened 2023-12-12
Add bindings for setting and getting emulation mouse from touch and touch from mouse in input

CaTaTo23:
This will allow to set emulation of this input events in runtime, for example enabling it in game menu on touchscreen only device to allow use Control based buttons

PR #86081, opened 2023-12-12
Add bindings for setting and getting emulation mouse from touch and touch from mouse in input

CaTaTo23:
This will allow to set emulation of this input events in runtime, for example enabling it in game menu on touchscreen only device to allow use Control based buttons.

Remake of #86076

### DIFF
--- a/doc/classes/Input.xml
+++ b/doc/classes/Input.xml
@@ -408,6 +408,12 @@
 		</method>
 	</methods>
 	<members>
+		<member name="emulate_mouse_from_touch" type="bool" setter="set_emulate_mouse_from_touch" getter="is_emulating_mouse_from_touch">
+			If [code]true[/code], sends mouse input events when tapping or swiping on the touchscreen. See also [member ProjectSettings.input_devices/pointing/emulate_mouse_from_touch].
+		</member>
+		<member name="emulate_touch_from_mouse" type="bool" setter="set_emulate_touch_from_mouse" getter="is_emulating_touch_from_mouse">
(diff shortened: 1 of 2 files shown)


doc/classes/Input.xml
```diff
@@ -408,6 +408,12 @@
 		</method>
 	</methods>
 	<members>
+		<member name="emulate_mouse_from_touch" type="bool" setter="set_emulate_mouse_from_touch" getter="is_emulating_mouse_from_touch">
+			If [code]true[/code], sends mouse input events when tapping or swiping on the touchscreen.
```
Calinou, line 412:
```suggestion
			If [code]true[/code], sends mouse input events when tapping or swiping on the touchscreen. See also [member ProjectSettings.input_devices/pointing/emulate_mouse_from_touch].
```

doc/classes/Input.xml
```diff
@@ -408,6 +408,12 @@
...
+		</member>
+		<member name="emulate_touch_from_mouse" type="bool" setter="set_emulate_touch_from_mouse" getter="is_emulating_touch_from_mouse">
+			If [code]true[/code], sends touch input events when clicking or dragging the mouse.
```
Calinou, line 415:
```suggestion
			If [code]true[/code], sends touch input events when clicking or dragging the mouse. See also [member ProjectSettings.input_devices/pointing/emulate_touch_from_mouse].
```

KoTeYkA23: Done

akien-mga (approved):
Looks good to me.

akien-mga:
Thanks!

PR #86082, opened 2023-12-12
Replace Windows window-like build emoji with real window emoji

adamscott:
🏁 ➡️ 🪟
👏👥👏👥👏👥👏👥👏👥👏👥👏👥👏👥👏👥👏👥👏👥👏👥👏👥👏👥👏👥👏

### DIFF
--- a/.github/workflows/runner.yml
+++ b/.github/workflows/runner.yml
@@ -41,7 +41,7 @@ jobs:
 
   windows-build:
     if: ${{ vars.DISABLE_GODOT_CI == '' }}
-    name: 🏁 Windows
+    name: 🪟 Windows
     needs: static-checks
     uses: ./.github/workflows/windows_builds.yml
 
--- a/.github/workflows/windows_builds.yml
+++ b/.github/workflows/windows_builds.yml
@@ -1,4 +1,4 @@
-name: 🏁 Windows Builds
+name: 🪟 Windows Builds
 on:
   workflow_call:
 


.github/workflows/runner.yml
```diff
@@ -41,7 +41,7 @@ jobs:
 
   windows-build:
     if: ${{ vars.DISABLE_GODOT_CI == '' }}
-    name: 🏁 Windows
+    name: 🪟 Windows
```
aaronfranke, line 44: I prefer the flag personally.
```suggestion
    name: 🏁 Windows
```

adamscott:
Here's what it looks like.
image

aaronfranke:
The cross does not show up on Firefox on macOS, it just has a wood frame with a glass pane:
Screenshot 2023-12-12 at 10 45 48 AM

JoNax97:
I can't see the new emoji, I see the default "missing character" white rectangle.

YuriSizov:
I think we didn't go for the obvious one because it's a relatively new addition and thus not supported on all platforms. So we picked from a very conservative list.

That said, I think there is a good reason to keep things as they are. Workflows tend to [linger in the list](https://github.com/godotengine/godot/actions), with old names never going away. So a small rename like that could end extra noise to the sidebar on that page. Which is not the end of the world, especially now that we have a single entry point for all workflows, but I would like to avoid frivolous renames like that nevertheless 🙃

.github/workflows/runner.yml
adamscott: I didn't know about that! Thanks, Aaron!

PR #86083, opened 2023-12-12
SCons: Fix pytest asserts on Windows

Repiteo:
Despite the script normally only running in the static checks on Ubuntu 22.04, it should still be possible to run the script on Windows if desired. Previously, this would always fail the assertion checks, as it tries to compare against different path separators. This PR fixes that by changing the parsed path separators to a Unix format. Additionally, this adds a new ignore for `.pytest_cache`, to exclude the cached data folder explicitly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -156,6 +156,9 @@ gmon.out
 # Mypy
 .mypy_cache
 
+# Pytest
+.pytest_cache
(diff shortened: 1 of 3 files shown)


Repiteo:
Superseded by #91840

PR #86085, opened 2023-12-12
Cleanup in undo in `TextEdit` and `LineEdit`

TheSofox:
In both `TextEdit` and `LineEdit`, the Undo and Redo functions seemed to check for themselves whether there was an Undo/Redo available rather than using the existing `has_undo` or `has_redo` functions. I decided to make them use them to remove repetition.

Only exception was the `TextEdit` `Undo` function. I noticed its check for an Undo was slightly different from the `has_undo` function. I decided to leave it alone.

### DIFF
--- a/scene/gui/text_edit.cpp
+++ b/scene/gui/text_edit.cpp
@@ -4024,7 +4024,7 @@ void TextEdit::redo() {
 	}
 	_push_current_op();
 
-	if (undo_stack_pos == nullptr) {
+	if (!has_redo()) {
 		return; // Nothing to do.
 	}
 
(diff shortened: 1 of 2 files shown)


scene/gui/text_edit.cpp
```diff
@@ -4024,7 +4024,7 @@ void TextEdit::redo() {
 	}
 	_push_current_op();
 
-	if (undo_stack_pos == nullptr) {
+	if (!has_redo()) {
```
KoBeWi, line 4027: Should the same change be done in `undo()`?

TheSofox: Like I said, there's a bit more complexity with the function and the bit of code it's intended to replace, so I decided to lave it be.

KoBeWi (approved):

Repiteo:
Thanks!

PR #86089, opened 2023-12-12
Add PWA option to ensure cross-origin isolation headers on web export

adamscott:
Sister PR to #85939

image

When the PWA option is enabled, the installed PWA will make sure to enable COEP/COOP request headers if the option is selected.

### DIFF
--- a/misc/dist/html/full-size.html
+++ b/misc/dist/html/full-size.html
@@ -218,8 +218,29 @@ const engine = new Engine(GODOT_CONFIG);
 		threads: GODOT_THREADS_ENABLED,
 	});
 	if (missing.length !== 0) {
-		const missingMsg = 'Error\nThe following features required to run Godot projects on the Web are missing:\n';
-		displayFailureNotice(missingMsg + missing.join('\n'));
+		if (GODOT_CONFIG['serviceWorker'] && GODOT_CONFIG['ensureCrossOriginIsolationHeaders'] && 'serviceWorker' in navigator) {
+			// There's a chance that installing the service worker would fix the issue
+			Promise.race([
+				navigator.serviceWorker.getRegistration().then((registration) => {
+					if (registration != null) {
+						return Promise.reject(new Error('Service worker already exists.'));
+					}
+					return registration;
+				}).then(() => engine.installServiceWorker()),
+				// For some reason, `getRegistration()` can stall
+				new Promise((resolve) => {
+					setTimeout(() => resolve(), 2000);
+				}),
+			]).catch((err) => {
+				console.error('Error while registering service worker:', err);
+			}).then(() => {
+				window.location.reload();
+			});
+		} else {
+			// Display the message as usual
+			const missingMsg = 'Error\nThe following features required to run Godot projects on the Web are missing:\n';
+			displayFailureNotice(missingMsg + missing.join('\n'));
+		}
 	} else {
 		setStatusMode('indeterminate');
 		engine.startGame({
--- a/platform/web/doc_classes/EditorExportPlatformWeb.xml
+++ b/platform/web/doc_classes/EditorExportPlatformWeb.xml
@@ -35,6 +35,10 @@
 		</member>
 		<member name="progressive_web_app/enabled" type="bool" setter="" getter="">
 		</member>
+		<member name="progressive_web_app/ensure_cross_origin_isolation_headers" type="bool" setter="" getter="">
+			When enabled, the progressive web app will make sure that each request has cross-origin isolation headers (COEP/COOP).
+			This can simplify the setup to serve the exported game.
+		</member>
 		<member name="progressive_web_app/icon_144x144" type="String" setter="" getter="">
 		</member>
 		<member name="progressive_web_app/icon_180x180" type="String" setter="" getter="">
--- a/platform/web/export/export_plugin.cpp
+++ b/platform/web/export/export_plugin.cpp
@@ -150,6 +150,7 @@ void EditorExportPlatformWeb::_fix_html(Vector<uint8_t> &p_html, const Ref<Edito
 	config["executable"] = p_name;
 	config["args"] = args;
 	config["fileSizes"] = p_file_sizes;
+	config["ensureCrossOriginIsolationHeaders"] = (bool)p_preset->get("progressive_web_app/ensure_cross_origin_isolation_headers");
 
 	String head_include;
 	if (p_preset->get("html/export_icon")) {
@@ -222,10 +223,12 @@ Error EditorExportPlatformWeb::_build_pwa(const Ref<EditorExportPreset> &p_prese
 	const String dir = p_path.get_base_dir();
 	const String name = p_path.get_file().get_basename();
 	bool extensions = (bool)p_preset->get("variant/extensions_support");
+	bool ensure_crossorigin_isolation_headers = (bool)p_preset->get("progressive_web_app/ensure_cross_origin_isolation_headers");
 	HashMap<String, String> replaces;
 	replaces["___GODOT_VERSION___"] = String::num_int64(OS::get_singleton()->get_unix_time()) + "|" + String::num_int64(OS::get_singleton()->get_ticks_usec());
 	replaces["___GODOT_NAME___"] = proj_name.substr(0, 16);
 	replaces["___GODOT_OFFLINE_PAGE___"] = name + ".offline.html";
+	replaces["___GODOT_ENSURE_CROSSORIGIN_ISOLATION_HEADERS___"] = ensure_crossorigin_isolation_headers ? "true" : "false";
 
 	// Files cached during worker install.
 	Array cache_files;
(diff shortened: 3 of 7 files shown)


platform/web/export/export_plugin.cpp
```diff
@@ -345,6 +347,7 @@ void EditorExportPlatformWeb::get_export_options(List<ExportOption> *r_options)
 	r_options->push_back(ExportOption(PropertyInfo(Variant::BOOL, "html/focus_canvas_on_start"), true));
 	r_options->push_back(ExportOption(PropertyInfo(Variant::BOOL, "html/experimental_virtual_keyboard"), false));
 	r_options->push_back(ExportOption(PropertyInfo(Variant::BOOL, "progressive_web_app/enabled"), false));
+	r_options->push_back(ExportOption(PropertyInfo(Variant::BOOL, "progressive_web_app/ensure_cross-origin_isolation_headers"), true));
```
AThousandShips, line 350: Do we want `-` in a property name? Is it used anywhere else? It feels like it breaks the style generally

adamscott: "cross-origin" is a word by itself. It does not fit with the _ as a space rule.
https://web.dev/articles/coop-coep

akien-mga: I think it would still be less surprising if it's properly snake_case, where `-` would be invalid.

adamscott: I renamed to `progressive_web_app/ensure_cross_origin_isolation_headers`

adamscott: Oh, though. The generated label text for the option will have it's hyphen missing.

platform/web/doc_classes/EditorExportPlatformWeb.xml
```diff
@@ -35,6 +35,10 @@
 		</member>
 		<member name="progressive_web_app/enabled" type="bool" setter="" getter="">
 		</member>
+		<member name="progressive_web_app/ensure_cross_origin_isolation_headers" type="bool" setter="" getter="">
```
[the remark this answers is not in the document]
AThousandShips: What is the typo?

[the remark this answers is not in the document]
akien-mga: I don't think the lack of a hyphen would be a typo. It still conveys the right meaning. For example "open source" and "open-source" are two widespread spellings for the concept, there's no loss of meaning IMO.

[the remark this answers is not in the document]
adamscott: Is there an English translation? Or we're stuck with the automatic version?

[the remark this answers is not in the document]
AThousandShips: Could add an editor name processor entry to improve the formatting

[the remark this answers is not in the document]
akien-mga: I checked but EditorPropertyNameProcessor splits by `_`, so it can only convert single words.

I don't think it's that big a deal myself, but if it's Absolutely Unacceptable™ not to have a hyphen, I guess we can make an exception for the property name. It's just weird to have a property name that's not a valid identifier.

adamscott:
Converted to draft as this doesn't seem to work, yet.

Faless (commented):
Thanks for looking into this :blue_heart: .
It's looking good overall :rocket: (see my comment on moving some code into `engine.js`).
I like the `s/@/__` rename, I wish it was done in its own PR though (EDIT: We can probably refactor this further into something similar to `GODOT_CONFIG` in a separate PR).
There are some "style" changes (e.g. using the [loose equality operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness)) which are [not great in JS](https://github.com/denysdovhan/wtfjs) .
I'll look into normalizing the service worker using eslint like we do for other bits of the javascript.

Faless:
Attached the 2 patches (001 adds the service worker linting and its config, 002 is just a linter pass).
[patch-001.diff.txt](https://github.com/godotengine/godot/files/13685328/patch-001.txt)
[patch-002.diff.txt](https://github.com/godotengine/godot/files/13685329/patch-002.txt)

adamscott:
> There are some "style" changes (e.g. using the [loose equality operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness)) which are [not great in JS](https://github.com/denysdovhan/wtfjs) .

Actually, this was done precisely with that in mind. `null` is pretty much the exception when using the loose equality operator. (see https://eslint.org/docs/latest/rules/eqeqeq#smart)

You actually want to compare a variable loosely to null. That's because Javascript considers `null` and `undefined` nullish (hence both values pass the loose equality test with `null`), but not the falsy values (eg. `""`, `0`, which do not pass the loose equality test with `null`). Whence why, instead, we should replace all the `!value` tests by `value != null`. (but that's for another PR)

```javascript
undefined == null; // `true`
"" == null; // `false`
```

adamscott:
> Attached the 2 patches (001 adds the service worker linting and its config, 002 is just a linter pass). [patch-001.diff.txt](https://github.com/godotengine/godot/files/13685328/patch-001.txt) [patch-002.diff.txt](https://github.com/godotengine/godot/files/13685329/patch-002.txt)
> 
> EDIT: Remember to `git add platform/web/.eslintrc.sw.js`

Done!

platform/web/package.json
```diff
@@ -5,18 +5,20 @@
   "description": "Development and linting setup for Godot's Web platform code",
   "scripts": {
     "docs": "jsdoc --template js/jsdoc2rst/ js/engine/engine.js js/engine/config.js js/engine/features.js --destination ''",
-    "lint": "npm run lint:engine && npm run lint:libs && npm run lint:modules && npm run lint:tools && npm run lint:html",
-    "lint:engine": "eslint \"js/engine/*.js\" --no-eslintrc -c .eslintrc.engine.js",
+    "lint": "npm run lint:engine && npm run lint:libs && npm run lint:modules && npm run lint:tools && lint:sw && npm run lint:html",
+    "lint:engine": "eslint \"../../misc/dist/html/*.js\" \"js/engine/*.js\" --no-eslintrc -c .eslintrc.engine.js",
```
Faless, line 9:
```suggestion
    "lint:engine": "eslint \"../../misc/dist/html/*.js\" \"js/engine/*.js\" --no-eslintrc -c .eslintrc.engine.js",
    "lint:engine": "eslint \"js/engine/*.js\" --no-eslintrc -c .eslintrc.engine.js",
```

My bad, my patch incorrectly included this change

platform/web/package.json
```diff
@@ -5,18 +5,21 @@
   "description": "Development and linting setup for Godot's Web platform code",
   "scripts": {
     "docs": "jsdoc --template js/jsdoc2rst/ js/engine/engine.js js/engine/config.js js/engine/features.js --destination ''",
-    "lint": "npm run lint:engine && npm run lint:libs && npm run lint:modules && npm run lint:tools && npm run lint:html",
+    "lint": "npm run lint:engine && npm run lint:libs && npm run lint:modules && npm run lint:tools && lint:sw && npm run lint:html",
```
Faless, line 8:
```suggestion
    "lint": "npm run lint:engine && npm run lint:libs && npm run lint:modules && npm run lint:tools && npm run lint:sw && npm run lint:html",
```

platform/web/package.json
```diff
@@ -5,18 +5,21 @@
...
+    "lint:engine": "eslint \"../../misc/dist/html/*.js\" \"js/engine/*.js\" --no-eslintrc -c .eslintrc.engine.js",
     "lint:engine": "eslint \"js/engine/*.js\" --no-eslintrc -c .eslintrc.engine.js",
+    "lint:sw": "eslint \"../../misc/dist/html/service-worker.js\" --no-eslintrc -c .eslintrc.sw.js",
     "lint:libs": "eslint \"js/libs/*.js\" --no-eslintrc -c .eslintrc.libs.js",
     "lint:modules": "eslint \"../../modules/**/*.js\" --no-eslintrc -c .eslintrc.libs.js",
     "lint:tools": "eslint \"js/jsdoc2rst/**/*.js\" --no-eslintrc -c .eslintrc.engine.js",
     "lint:html": "eslint \"../../misc/dist/html/*.html\" --no-eslintrc -c .eslintrc.html.js",
-    "format": "npm run format:engine && npm run format:libs && npm run format:modules && npm run format:tools && npm run format:html",
+    "format": "npm run format:engine && npm run format:libs && npm run format:modules && npm run format:tools && format:sw && npm run format:html",
```
Faless, line 16:
```suggestion
    "format": "npm run format:engine && npm run format:libs && npm run format:modules && npm run format:tools && npm run format:sw && npm run format:html",
```

Faless (approved):
Needs a rebase (tested here: https://github.com/Faless/godot/tree/temp/86089) and squash, then I think this is good to go.
Great job :+1:

misc/dist/html/full-size.html
```diff
@@ -215,8 +215,29 @@
...
+			]).catch((err) => {
+				console.error('Error while registering service worker:', err);
+			}).then(() => {
+				window.location.reload();
+			});
+		} else {
+			// Display the message as usual
+			const missingMsg = 'Error\nThe following features required to run Godot projects on the Web are missing:\n';
+			displayFailureNotice(missingMsg + missing.join('\n'));
+		}
```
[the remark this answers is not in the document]
adamscott: Oops, will do it. I forgot.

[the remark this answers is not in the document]
adamscott: I'll keep the code as is, finally. We would need to pass in parameters to that function the config and the `engine` variable, which seems a little bit excessive.

adamscott:
Rebase done!

akien-mga:
Thanks! Great feature to have :)

PR #86093, opened 2023-12-12
[D3D12] Automatically copy runtime DLLs during export.

bruvzg:
Automatically copy runtime DLLs for the D3D 12 rendered during export.

Note: since export template for the windows do not have folder structure and consist of loose files, expected names for the libraries in the export template is `name.arch.dll` (e.g., `dxil.x86_64.dll`).

### DIFF
--- a/platform/windows/doc_classes/EditorExportPlatformWindows.xml
+++ b/platform/windows/doc_classes/EditorExportPlatformWindows.xml
@@ -18,8 +18,14 @@
 		<member name="application/copyright" type="String" setter="" getter="">
 			Copyright notice for the bundle visible to the user. Optional. See [url=https://learn.microsoft.com/en-us/windows/win32/menurc/stringfileinfo-block]StringFileInfo[/url].
 		</member>
+		<member name="application/d3d12_agility_sdk_multiarch" type="bool" setter="" getter="">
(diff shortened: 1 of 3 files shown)


platform/windows/doc_classes/EditorExportPlatformWindows.xml
```diff
@@ -18,9 +18,15 @@
 		<member name="application/copyright" type="String" setter="" getter="">
 			Copyright notice for the bundle visible to the user. Optional. See [url=https://learn.microsoft.com/en-us/windows/win32/menurc/stringfileinfo-block]StringFileInfo[/url].
 		</member>
+		<member name="application/d3d12_agility_sdk_multiarch" type="bool" setter="" getter="">
+			If [code]true[/code], and [member application/export_d3d12] is set. the Agility SDK DLLs will be stored in arch-specific subdirectories.
```
dsnopek, line 22: I think this was meant to be a comma:

```suggestion
			If [code]true[/code], and [member application/export_d3d12] is set, the Agility SDK DLLs will be stored in arch-specific subdirectories.
```

akien-mga (approved):

akien-mga:
Thanks!

PR #86094, opened 2023-12-12
Fix returned value in Resource Importer when `post_import` fails.

alesliehughes:
err was set to OK and returned on an error path.

This appears to be the correct value to return.

### DIFF
--- a/editor/import/3d/resource_importer_scene.cpp
+++ b/editor/import/3d/resource_importer_scene.cpp
@@ -3255,7 +3255,6 @@ Error ResourceImporterScene::import(ResourceUID::ID p_source_id, const String &p
 			}
 		}
 	}
-	err = OK;
 
 	progress.step(TTR("Running Custom Script..."), 2);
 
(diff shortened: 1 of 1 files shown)


Calinou:
I'd test valid and invalid import scripts with this change first, as it's possible that this seemingly erroneous code is being relied upon somewhere.

akien-mga:
Do you have a test project you used to reproduce the bug / validate the fix? This would help confirm that the fix is correct.

alesliehughes:
I don't have a project. This was picked up via coverity checks.

akien-mga:
Can you share the details on the Coverity report for that bug?

alesliehughes:
Sorry this wasn't a Coverity reported bug directly, but something I noticed during other bugs fixes.

We start with, setting this to a successful.
Error err = OK;      
..
```
if (!post_import_script->get_script_instance()) {
				EditorNode::add_io_error(TTR("Invalid/broken script for post-import (check console):") + " " + post_import_script_path);
				post_import_script.unref();
				return ERR_CANT_CREATE;
			}
```
If we cannot get an instance, report an error.

Little bit down.
...

err  is OK  at this point.
```
if (post_import_script.is_valid()) {
		post_import_script->init(p_source_file);
		scene = post_import_script->post_import(scene);
		if (!scene) {
			EditorNode::add_io_error(
					TTR("Error running post-import script:") + " " + post_import_script_path + "\n" +
					TTR("Did you return a Node-derived object in the `_post_import()` method?"));
			return err;
		}
	}
```
So the post_import script failed for some reason but now returns success.

There is other places, that return non OK further down.

Issue #68305, opened 2022-11-05
Godot Editor and Project Manager steals focus on a window manager on Linux.


nargacu83:
### Godot version

v4.0.beta4.official [e6751549c] (any 4.0 version)

### System information

Arch Linux with kernel 6.0.7-arch1-1 on AwesomeWM, X11

### Issue description

The project manager and the editor steals the focus on Window Managers on Linux with X11 when opening other programs or when changing layouts. It doesn't happen on Godot 3.x.

I tested it on [AwesomeWM](), [Hypr](https://github.com/hyprwm/Hypr), [LeftWM](https://github.com/leftwm/leftwm) and [DWM](https://dwm.suckless.org/). Only LeftWM and DWM doesn't have the issue but i'm pretty sure they don't "implement" [ConfigureNotify](https://tronche.com/gui/x/xlib/events/window-state-change/configure.html) but i don't know much in the display server area.

However i managed to "fix" the issue by commenting out this if statement.
https://github.com/godotengine/godot/blob/6882890a34528e056f13020798f4a473046990bb/platform/linuxbsd/x11/display_server_x11.cpp#L3855-L3857

### Steps to reproduce

1. Login with AwesomeWM
2. Be on the `awful.layout.suit.tile` layout
3. Launch any version of Godot 4.0
4. Launch another window

### Minimal reproduction project

You can use the default configuration of AwesomeWM.
No project needed.

fjfnaranjo:
I can also confirm this bug for Fluxbox on Debian 10.

LXDE on Debian 11 works as expected.

The fastest way to reproduce it is opening Godot with an empty project list. The project list window and the "open assets store" popup fight for the focus.

lihop:
I can confirm this issue for Godot 4.0-beta10 with XMonad 0.17.1 on NixOS 2.11.

Expected and actual behavior is shown below. I have configured XMonad to draw a bright green border around the currently focused window.

| Expected Behavior (Godot 3.4-stable) | Actual Behavior (Godot 4.0-beta10) |
| -------------------------------------------------- | ----------------------------------------------- |
| ![focus-right](https://user-images.githubusercontent.com/3696783/209721010-501d3fd6-0609-47d4-b887-816d6433294e.gif) | ![focus-wrong](https://user-images.githubusercontent.com/3696783/209723230-98ab84b1-be87-48e5-a7ce-c812862fd391.gif) |

[...]

Issue #74378, opened 2023-03-04
godot xfce4 focus steal


Mequam:
### Godot version

4.0.stable.official.92bee43ad

### System information

x86_64 Linux 6.2.2-arch1-1, Intel Core i5-7200U, xfce4 

### Issue description

When changing workspace on xfce4 godot (both editor and exported games) grabs input focus from any previously opened programs.

https://user-images.githubusercontent.com/49560528/222922924-af92e9bc-3d29-4ec0-865b-a6f6d2afc189.mp4

Interestingly I was unable to reproduce the bug on gnome, so whatever the problem is I suspect it is specific to xfce4.

From my testing this appears to be happening because xfce4 issues an extra MapNotify event which the engine uses as an indicator to grab focus.

I have a fix that works for my setup: https://github.com/Mequam/funky-godot/tree/mequam/feat/xfce4_focus_grab but I figured I should make sure that other people are having the same issue before submitting a PR.

### Steps to reproduce

Open up godot and some other program. 
Focus the other program.
Leave the workspace and return to the workspace.

### Minimal reproduction project

N/A, any project works.

akien-mga:
CC @bruvzg

compese:
related? https://github.com/godotengine/godot/issues/74587#issuecomment-1466538062

compese:
I've naively removed all XSetInputFocus calls from 4.0-stable and it solves the problem for me for the time being (seeing this issue in XMonad ). It seems like sometimes  XSetInputFocus  is getting called inappropriately i.e. when another application  has focus. I'm not an X11 programmer but my guess is Godot is getting faked out by tiling window manager abilities like resizing a window without focusing on it. 

e.g.
I have a terminal and Godot open. They're tiled and the terminal is focused. Then I expand the terminal (which concurrently shrinks Godot) while keeping focus on the terminal. AFAICT, in this scenario Godot thinks it should then have focus because it's been resized even though the user has not actually focused Godot. i.e. Godot mistakes resizing as a focus request.

Similar behavior happens if I swap the terminal and Godot while keeping focus on the terminal. Godot incorrectly interprets window re-positioning as a focus request.

[...]

Issue #41578, opened 2020-08-28
Editor Menus and Text Entry Boxes Appear at Origin


duane-r:
**Godot version:**
master v4.0.dev.custom_build.1ff139cb2

**OS/device including version:**
OS: Archlinux, AMD ryzen integrated graphics, mesa driver
Window manager: Openbox (only)

**Issue description:**

All of the editor's popups appear at origin (upper left), though they tile, to make room for more. The first screenshot shows tabbing through the rectangle entries in a textrectangle. The second just shows the help menu opening in the wrong place. Submenus will open up below the menu.

![2020-08-28-051924_1920x1080_scrot](https://user-images.githubusercontent.com/14078065/91551299-8c7c3700-e8ef-11ea-989f-44862703a1d5.jpg)

![2020-08-28-052000_1920x1080_scrot](https://user-images.githubusercontent.com/14078065/91551371-a61d7e80-e8ef-11ea-8edc-8d109482a8aa.jpg)

![2020-08-28-053518_1920x1080_scrot](https://user-images.githubusercontent.com/14078065/91551794-63a87180-e8f0-11ea-9e51-cf7579e11a3b.jpg)

**Steps to reproduce:**

**Minimal reproduction project:**

Edited for form.

Calinou:
I can't reproduce this on commit https://github.com/godotengine/godot/commit/118a7bc53f32b8303b70886eee7ea090c3d327f0 (Fedora 31 KDE, GeForce GTX 1080).

duane-r:
Still doing it on a25818072. I'm using openbox window manager. I can try putting kde on to see if anything changes. I can't do anything about my integrated graphics though. : )

mohammedzero43:
Are u using dwm or i3 windows manager? 
It will be funny if the wm treats every menu as a new window thus it pins it to the top

duane-r:
> Are u using dwm or i3 windows manager?

Just openbox, but that might be the problem.

Ok, it works perfectly in kde, so presumably the bug is limited to lxde and openbox users. However, Godot 3.2 doesn't have this issue at all.

Calinou:
> However, Godot 3.2 doesn't have this issue at all.

This is a regression from the DisplayServer refactor which allows Godot to display multiple windows.

duane-r:
As a workaround, you can set placement/policy/UnderMouse in openbox's configuration file, which at least puts the popups close to where your mouse is.

Issue #41574, opened 2020-08-28
Freeze when a window or popup is open (Linux with i3) (workaround: Use `--single-window`)


pouleyKetchoupp:
**Godot version:**
Master b0ca859501186b6fa088a3d3542f7dd4d72ae305

**OS/device including version:**
Ubuntu 20.04 / WM: i3

**Issue description:**
This issue is specific to i3 (tiling window manager).

In multi-windows mode, the whole editor freezes for a few seconds each time a new window is opened, or when the focus switches from the editor or back to the editor.

It doesn't seem to be related to a specific change in Godot but didn't happen before, so it might be due to an update in i3 or one of its dependencies.

More details here: https://github.com/godotengine/godot/pull/41456#issuecomment-681299904

**Steps to reproduce:**
Can be reproduced easily by starting the editor with any project, using i3 window manager.

**Minimal reproduction project:**
n/a

DavidNexuss:
Can confirm same thing happens with Arch Linux, I've test so far the following WMs and their last versions at the moment of writing this:

- bspwm 0.9.10-1
- gnome (gnome-desktop 1:3.36.5-1 the x11 version)
- kde plasma (plasma-desktop 5.19.4-1)

The whole editor freezes each time a popup or a window is created, the error seem more related to the latest versions of xorg packages as to me its seems more WM agnostic, at least on Arch Linux. I also tried to replicate the issue with dwm though in this one godot just crashed instead of freezing the editor.

I made the tests with a fresh build of godot in release_debug in the latest commit on the master branch #

Riteo:
If I reproduce this issue while editing a project, the editor freezes completely.

Riteo:
After a lot of ideas, tests and weird behaviour, I found a workaround which is getting us closer and closer to the exact cause of the freezing.
Since this freeze never happened before, and could be replicated even on old builds, I tried to rollback all possibly related packages with godot.
I found out eventually that by downgrading NVIDIA's drivers from version 450.66 to 450.57 the freezing doesn't happen anymore.

nathanfranke:
This can be worked around by enabling **Single Window Mode**.

Since you can't really open editor settings with this bug, edit `~/.config/godot/editor_settings-4.tres` and under `[resource]` add `interface/editor/single_window_mode = true`

Calinou:
@nathanfranke You can also pass the `--single-window` command-line argument to Godot to force single window mode.

Riteo:
[...]

PR #86101, opened 2023-12-13
[X11] Partial fix for Editor and Project Manager stealing focus on some window managers

GrammAcc:
This is a workaround for the most critical portion of the WM focus bug described in #68305. On some specific X11 WM configurations, the editor's main window and any popups it creates will fight for focus, which causes a total system lockup due to mouse and keyboard input being stolen as well. Getting out of this infinite loop requires force restarting the system.

It can be tested with the following shell script:

```bash
!#/bin/sh

godot4 &
sleep 30
pkill -x godot4
```

Here is a gif of the bug in action on my machine:

68305-bug

The gif above actually doesn't show the exact behavior. The constant focus swap is actually too fast for my screen recorder's framerate to pick up, so what you are seeing is about 24 out of 60 focus swaps. :)

Here is the same with this patch applied:

68305-fixed

My system information:
- Linux version: 6.6.6-arch1-1 (linux@archlinux) (gcc (GCC) 13.2.1 20230801, GNU ld (GNU Binutils) 2.41.0)
- WM: Fluxbox-only launched via Xinit.

The workaround identified by @nargacu83 in #68305 is to remove the call to XSetInputFocus in the ConfigureNotify event handler, so I have removed the conditional block that calls this as well as the setup code above it since there is no need to allocate the memory for the variables if they won't be used in that call anymore.

This is just a hack and is not a complete fix for #68305. Multiple developers are collaborating on a proper fix in the discussion in that issue, but time is a valuable resource that no one has enough of, so I am submitting this workaround as a stop-gap to prevent the most critical problem while we work on a full solution for the underlying cause.

This bug affects a very small number of users, and the general consensus in the related issue is that we don't want to submit a PR for a hack if we can get a proper fix. Particularly since we aren't completely sure why this workaround fixes the problem, and I agree with this as well, but personally, I am concerned that with this being a platform bug, it could affect exported games and not just developers, and in that case, it probably wouldn't show up on the developer's machine, but a player with the affected configuration could encounter the bug when running an exported Godot game, and although the number of people playing games on custom WM-only Linux machines is probably very low, it only takes one very vocal gamer on Twitter to ruin an indie studio, so I wanted to submit this workaround now instead of sitting on it since no one in the related issue has reported any regressions with this workaround, and we don't know when a proper fix will be ready.

Please feel free to close this PR if you feel that we should wait until we have a better idea of the problem and exactly *why* this workaround works. Thanks!

### DIFF
--- a/platform/linuxbsd/x11/display_server_x11.cpp
+++ b/platform/linuxbsd/x11/display_server_x11.cpp
@@ -4718,19 +4718,6 @@ void DisplayServerX11::process_events() {
 					break;
 				}
 
-				const WindowData &wd = windows[window_id];
-
-				XWindowAttributes xwa;
-				XSync(x11_display, False);
-				XGetWindowAttributes(x11_display, wd.x11_window, &xwa);
-
-				// Set focus when menu window is re-used.
-				// RevertToPointerRoot is used to make sure we don't lose all focus in case
-				// a subwindow and its parent are both destroyed.
-				if ((xwa.map_state == IsViewable) && !wd.no_focus && !wd.is_popup && _window_focus_check()) {
-					_set_input_focus(wd.x11_window, RevertToPointerRoot);
-				}
-
 				_window_changed(&event);
 			} break;
 


bruvzg:
For the reference, it was added in https://github.com/godotengine/godot/pull/41456, and issues with freezing on some WMs were reported soon after https://github.com/godotengine/godot/issues/41574, but it's marked as resolved (probably only fixed with the i3 specifically).

bruvzg:
Popup handling was changed a lot, so I'm not sure if this code is necessary for anything. It should be tested to see if it's reintroducing https://github.com/godotengine/godot/issues/37674 or https://github.com/godotengine/godot/issues/41578 on non tiling WMs.

akien-mga:
> Popup handling was changed a lot, so I'm not sure if this code is necessary for anything. It should be tested to see if it's reintroducing #37674 or #41578 on non tiling WMs.

Tested briefly on KDE Plasma on X11, it doesn't reintroduce those issues for me. From a quick test I didn't spot any weird change in behavior.

Would be good to have it tested on more DEs/WMs and on Xwayland.

CC @Riteo @groud @Calinou

dsnopek:
As another random data point, I use i3 and Godot's "window grabbing focus behavior" is super disruptive. However, in my case, it doesn't lead to multiple windows fighting for focus and things getting locked up, it's just very annoying. :-) When I have a chance I'll try with this PR and see if it helps in my case. But it would be kind of nice to have an editor setting that I could enable that would just prevent any window in the editor from grabbing focus.

GrammAcc:
> As another random data point, I use i3 and Godot's "window grabbing focus behavior" is super disruptive. However, in my case, it doesn't lead to multiple windows fighting for focus and things getting locked up, it's just very annoying. :-) When I have a chance I'll try with this PR and see if it helps in my case. But it would be kind of nice to have an editor setting that I could enable that would just prevent any window in the editor from grabbing focus.

In case it helps, #74378 is related to #68305, but has a broader scope, and users have reported odd focus behavior between workspaces on i3. I generally only use one workspace, so I haven't looked into this, but one user confirmed that this workaround does help with focus on i3 but doesn't fix the issue between workspaces. See https://github.com/godotengine/godot/issues/68305#issuecomment-1762851375 for details.

Confirmation from another i3 user would be really helpful though. :)

dsnopek:
I just tested this PR with i3, and things don't seem to get any worse, but it doesn't seem to help my problems either.

To be clear, my issues mostly revolve around switching between the Godot editor and VS Code, where Godot will grab (or refuse to give up) input focus, and despite a new window visually appearing to have focus (like, the title bar is highlighted in the way that usually indicates it has focus), all the keyboard input still goes into the Godot editor. My "work around" is that I have to toggle back and forth between focusing on Godot and focusing on the thing I actually want to get my keyboard inputs, and eventually they'll go where I want.

I'll try to see if my problem is represented in one of the existing issues, and if not, make a new issue.

GrammAcc:
> I just tested this PR with i3, and things don't seem to get any worse, but it doesn't seem to help my problems either.
> 
> To be clear, my issues mostly revolve around switching between the Godot editor and VS Code, where Godot will grab (or refuse to give up) input focus, and despite a new window visually appearing to have focus (like, the title bar is highlighted in the way that usually indicates it has focus), all the keyboard input still goes into the Godot editor. My "work around" is that I have to toggle back and forth between focusing on Godot and focusing on the thing I actually want to get my keyboard inputs, and eventually they'll go where I want.
> 
> I'll try to see if my problem is represented in one of the existing issues, and if not, make a new issue.

Yeah, there are a lot of problems with the focus behavior on WMs in the X11 backend. This PR is only intended to address the system lock up on very specific configurations since it seems like a fix for the other focus issues will require a significant overhaul of the event handlers, and a few devs have been working on this off and on for months without a lot of progress (mostly due to lack of time). I think some of the comments in #74378 mentioned similar issues with i3, and the user who tested this workaround on i3 previously did mention something similar about focus problems with an external text editor.

Thanks for testing it out!

GrammAcc:
Okay, I was able to test this on the following WMs/DEs: Fluxbox with picom (Xwayland), Openbox X11, Openbox Xwayland, Budgie, Cinnamon, Enlightenment, lxde, lxqt over Fluxbox, lxqt over Openbox, Mate, and xfce4.

I just did a quick test of launching the editor and opening a popup, so I have not tested for regressions related to #37674 and #41578 yet, but I wanted to make sure there wasn't anything weird going on with popups before regression testing on each DE.

I will check for #37674 and #41578 on each DE eventually this week, but it probably won't be tonight. :)

Here are the gifs of launching the editor and opening a popup window in each environment:

Fluxbox Xwayland:

Xwayland-fluxbox

Openbox X11:

X11-openbox

Openbox Xwayland:

Xwayland-openbox

Budgie:

budgie

Cinnamon:

cinnamon

Enlightenment:
File size too large, but the result is the same as all the others.

lxde:

lxde

lxqt-fluxbox:

lxqt-fluxbox

lxqt-openbox:

lxqt-openbox

Mate:

mate

xfce4:

xfce4

This has really reminded me why I don't use full DEs. Thank $deity for XBindKeys :rofl:

GrammAcc:
Okay, scrolling through 10 gifs is a little much. Lmk if you want me to remove them.

GrammAcc:
Okay, I took a look at #37674 and #41578 and they were much simpler to test than I thought. :man_facepalming: 

I tested both on all of these environments: Fluxbox X11, Fluxbox Xwayland, Openbox X11, Openbox Xwayland, Budgie, Cinnamon, Enlightenment, lxde, lxqt over Fluxbox, lxqt over Openbox, Mate, and xfce4.

The following is a gif of the test on Fluxbox X11. All of the other environments gave exactly the same behavior, so I am only uploading the single gif:

regression-test-fluxbox

I believe this is everything I can test as far as different DEs goes. Please let me know if you have any questions for me. Thanks!

Calinou (commented):
Tested locally (Fedora 39 KDE with X11, rebased against https://github.com/godotengine/godot/commit/96296e476ffda2d4c36680b2271e9eeff0c3b9f8), it doesn't improve the situation with regards to focus stealing on floating window managers. The issue described in https://github.com/godotengine/godot/pull/86441#pullrequestreview-1827983083 still occurs with this PR, unfortunately.

That said, with this PR cherry-picked on top of https://github.com/godotengine/godot/pull/86441, the issue reported in the other PR remains fixed. I tested the editor quickly and couldn't notice any regressions in editor popup behavior. As a result, we can still choose to merge this if it helps improve the situation on tiling WMs.

akien-mga:
This needs a rebase against the current `master` branch to resolve conflicts. And then I guess we should just give it a spin and see if any X11 users report any regression.

akien-mga (approved):
Let's try it out!

akien-mga:
Thanks!

PR #85120, opened 2023-11-20
OpenGL: Implement rendering of lightmaps

dsnopek:
This PR aims to add rendering of lightmaps (via `LightmapGI`) when using the GL Compatibility renderer.

**Note: This doesn't add lightmap baking when using the GL Compatibility renderer. Lightmaps still need to be baked using Vulkan.**

~~This marked as a draft because while it's "working", the result looks worse than I'd expect. I want to see if I can get to the bottom of that first.~~

~~**UPDATE:** Eh, it looks good enough to get some review :-)~~

**UPDATE 2:** In [a comment below](https://github.com/godotengine/godot/pull/85120#issuecomment-1819250042), @Calinou (who knows rendering much better than me) says it looks as good as he'd expect given the way the Compatibility renderer works, so I guess this just needs code review now!

PR #50601, opened 2021-07-19
Implement indirect specular occlusion in StandardMaterial3D

Calinou:
`master` version of https://github.com/godotengine/godot/pull/50603.

Both sources of AO (SSAO and AO texture maps) are used for this effect. The difference is most noticeable when SSAO is enabled (especially with height maps), but this effect also helps when SSAO is disabled.

It is an approximate effect with a very low performance cost, but makes specular reflections more believable overall.

**Testing project:** [test_ao_specular_occlusion_master.zip](https://github.com/godotengine/godot/files/8018158/test_ao_specular_occlusion_master.zip)

## Preview

*SSAO intensity was doubled to make the difference more noticeable. In real world scenarios, the difference will be more subtle.*
*Light Affect is set to 0 (default) both in the SSAO and material settings.*

*Left: partially metallic material (0.5), right: fully metallic material (1.0)*

#### Forward+

Before | After
-|-
Screenshot_20230526_234722 webp | Screenshot_20230526_234703 webp

#### Mobile[^1]

[^1]: This currently uses a faster, but *supposedly* more approximate approach to indirect specular occlusion.

Before | After
-|-
Screenshot_20230526_234747 webp | Screenshot_20230526_234735 webp

#### Forward+

Before | After
-|-
Screenshot_20230526_234445 webp | Screenshot_20230526_234424 webp

#### Mobile[^1]

Before | After
-|-
Screenshot_20230526_234602 webp | Screenshot_20230526_234549 webp

[^1]: This currently uses a faster, but *supposedly* more approximate approach to indirect specular occlusion.

PR #86102, opened 2023-12-13
Add Specular Occlusion to meshes under influence of LightmapGI

### LINKED ISSUE #8637: Add a cheap specular occlusion for LightmapGI

I'm working on a procedural and UGC game where users can create their own maps from blocks and custom models.

Some surfaces can have metallic or reflective materials that still receive reflections (from environment sky or reflection probes) even though they should be darkened/occluded from lightmaps.  

I can't use directional lightmaps as they take more size and takes longer to compute (they're also buggy at the moment).

The goal would be to add specular occlusion to lightmapped meshes.  
To make it as cheap as possible, its value would be taken from the brightness of the current targeted LM pixel and then decrease its specular intensity. 

It should work out of the box without the user needing to do anything.  
There should be an option (in the project settings or material) to disable (or enable) specular occlusion as it can be intended for some games.  

I made a proof of concept by editing the existing code, and complementing it with my other PR (https://github.com/godotengine/godot/pull/86039) :
|No Lightmap|Lightmap on 4.2|With specular occlusion|Specular Occlusion + PR (LM+SDFGI)|
:---:|:---:|:---:|:---:|
no lightmaps | godot 4.2 |lightmap & specular occlusion | lightmap & specular occlusion + sdfgi

I don't think it's possible unless you're able to get the lightmap data from shader.

LightmapGI is core

guerro323:
*Closes https://github.com/godotengine/godot-proposals/issues/8637*
##
image
##
This pull request implement specular occlusion which reduce the specular light of objects under a lightmap.  

It offers two methods and an option to disable it in the project settings:
- **Reduce**: Reduce specular light by the brightness of the current lightmap pixel.
- **Conservative** (Default): Closest to ground truth, it keeps more energy on highly reflective materials (such as `metallic=1.0 && roughness=0.0`).
- **Disabled**: Existing behavior before this PR.  

Which one to choose?
- Use `Reduce` at first, while it doesn't give a correct result especially with high reflective materials it's gonna be better than having it reflect the sky while fully occluded.  
- Use  `Conservative` when your materials are well setup (such as having correct PBR textures) and a good scene that has good overall reflection sources (either from reflection probes or with GI reflections).
- Use `Disabled` if you want your metallic and shiny materials to act the same as before or if you want a specific art style.

Specular occlusions mixes well and improves:
- Reflection probes, giving them more fidelity.
- SSR as specular occlusion will hide most of the unwanted artifacts.
- GI reflections such as SDFGI (require https://github.com/godotengine/godot/pull/86135) removing unwanted artifacts and improving resolution.

It support both Forward+, Mobile and Compatibility renderers.  

To get started using it, you just need to open an existing scene with my changes and it should work out of the box. Of course if you used materials that act on the specular light (metallic, roughness, specular...) you may need to make some small modifications on them.

##
<details><summary><h3>Screenshots and Comparisons<h3></summary>  
  
## Custom Scene  
A scene I'll provide later once I'll fix some meshes issues (and improving the textures)
|No Lightmaps|Current 4.2|Specular Occlusion `Reduce`|Specular Occlusion `Conservative`|
:---:|:---:|:---:|:---:
image | image | image | image
## Japanese Street  
Link: https://sketchfab.com/3d-models/japanese-street-at-night-fb1bdcd71a5544d699379d2d13dd1171
On this one, `Conservative` has less impact since a majority of the textures aren't metallic and reflective enough.
|Current 4.2|Specular Occlusion `Reduce`|Specular Occlusion `Conservative`|
:---:|:---:|:---:
image | image | image

## Box
Note: A fully reflective material on `Conservative` can still be 100% darkened, it's not happening here because it's not dark enough
|`metallic=0`|`metallic=1`|`metallic=1 roughness=0 reduce`|`metallic=1 roughness=0 conservative`|
:---:|:---:|:---:|:---:
image | image | image | image

## Influences by lightmap probes
Characters from: https://github.com/gdquest-demos/godot-4-3D-Characters
The scene used here isn't a good one to compare
### `metallic=1 roughness=0`
|Godot 4.2|Specular Occlusion `Reduce`|Specular Occlusion `Conservative`|
:---:|:---:|:---:
image | image | image

### `metallic=1 roughness=1`
|Godot 4.2|Specular Occlusion `Reduce`|Specular Occlusion `Conservative`|
:---:|:---:|:---:
image | image | image

Dynamic objects on the first scene:
image
*(note: probes seems to have issue outside of this PR in Physical Lights mode, I disabled it here which is why it may look different)*

</details>

# Problems
<details><summary>Dynamic lights and dynamic objects will not have a correct specular occlusion as it will be a bit darker.  
It can be boosted in the code but I just need to find the correct value so that it look good enough.   
Otherwise you can either reduce metallic or use the `Conservative` method to boost it.   
Reason is that the lightmapper seems to darken the lightmap a little bit when in dynamic mode.</summary>   

|No Lightmaps|Lightmapped Non-Static Light|Lightmapped Non-Static Light + Conservative|Lightmapped Static Light|
:---:|:---:|:---:|:---:
image | image | image| image

</details>


jcostello:
Awesome results. I will test it out.

image

In that screenshot, where the reflections comes from?

jcostello:
I test it. Works incredible well. The results are more realistic and works awesome with reflection probes

guerro323:
> Awesome results. I will test it out.
> 
> In that screenshot, where the reflections comes from?

They come from the middle of the bottom cube!

scene/resources/material.h
```diff
@@ -275,6 +275,13 @@ class BaseMaterial3D : public Material {
 		SPECULAR_MAX
 	};
 
+	enum SpecularOcclusionMode {
+		SPECULAR_OCCLUSION,
```
AThousandShips, line 279:
```suggestion
		SPECULAR_OCCLUSION_DEFAULT,
```
I'd name this more descriptively (also this would break in C# I think, where prefix is stripped)

dsnopek:
Is this a feature that makes sense to add to the compatibility renderer too? Now that PR https://github.com/godotengine/godot/pull/85120 is merged, `LightmapGI` will be rendered in the compatibility renderer too.

guerro323:
> Is this a feature that makes sense to add to the compatibility renderer too? Now that PR #85120 is merged, `LightmapGI` will be rendered in the compatibility renderer too.

Compatibility support has been added :slightly_smiling_face:
|Before|After `Default`|After `Conservative`|
:---:|:---:|:---:|
image | image | image

clayjohn (commented):
I think this looks amazing! Great work. 

My only concern is over whether we need two render modes for this. We try to keep options simple as possible. 

Looking at the code, the conservative mode only adds a couple of instructions and it appears to be the more "correct" looking option. Would it make sense to only expose one render mode for specular occlusion and have it apply the current "conservative" option?

Basically, is there a situation where the non-conservative option is preferred other than just being slightly less costly?

jcostello:
> I think this looks amazing! Great work.
> 
> My only concern is over whether we need two render modes for this. We try to keep options simple as possible.
> 
> Looking at the code, the conservative mode only adds a couple of instructions and it appears to be the more "correct" looking option. Would it make sense to only expose one render mode for specular occlusion and have it apply the current "conservative" option?
> 
> Basically, is there a situation where the non-conservative option is preferred other than just being slightly less costly?

I didn't get what you suggest. Also I found the `default` mode more pleasant for indoors and some other scenes that I have (This is an artistic choise)

guerro323:
@clayjohn 
I agree that having two modes for it complicate the whole thing :/
There are two important factors where "conservative" might not be solution:
- The project's art style
- The reflection sources: If there is only the sky for reflections you'll get a too much blown out material, it's the most accurate but may not suit the art style or context of the object.

But "conservative" is the best one when there are enough source of reflections in darker areas.
To illustrate the problem better, here is an example:
|Lightmap Only + Reflection Probe in the center (capture mostly the sky) and "Conservative"|LM + "Default"|LM + SDFGI + "Conservative"|LM + SDFGI + "Default"|
:---:|:---:|:---:|:---:
image | image | image | image  

So as shown in the table, if you choose conservative and don't have any possibility for another reflection then it will look too much blown out. But enabling SDFGI here make conservative the best choice.  

I'm not really sure what to do if we're going for one setting as it would mostly force realistic art styles, maybe going for a mix of both? Or a way to keep both but make it simpler but how?  
Maybe a slider would be better and it would lerp between the two option?

Calinou:
Can we expose the Default/Conservative option as a project setting rather than a render mode? You probably won't need to use both modes at the same time in the same project, and probably don't need to change it at runtime either.

jcostello:
> Can we expose the Default/Conservative/Disabled option as a project setting rather than a render mode? As I understand it, you probably won't need to use both modes at the same time in the same project, and probably don't need to change it at runtime either.
> 
> This should allow having only a single render mode and avoiding an increase in the number of shader permutations or VGPR usage.

Could be a Lightmap property (this is the first place I look at it)

guerro323:
It has been moved to a project settings.  

For now modifying the value require a restart.  
I'm not sure if it's possible to do a shader recompile while the editor/game is running as it become useful to swap between `Reduce/Conservative` mode when `SDFGI/VoxelGI` is disabled/enabled when combined with my other PR (https://github.com/godotengine/godot/pull/86135); it only become a problem with big projects that takes a long time to load.

s-ilent:
A few things...
1. Would it make sense to have the intensity of specular AO from lightmap be a user-adjustable value? What intensity in the lightmap is "dark" depends a lot on the scene. It seems like you're suggesting the simple/complex mode switch should be used for determining whether the user want more or less specular AO, but that seems like it makes what the specular AO is doing less explicit. Plus a project-wide setting is harder to control. I can see the specular AO is attenuated by the exposure value, but I'm worried that wouldn't cover for all scenarios. Having a tweakable value (i.e. scaling the occlusion value before clamping) would also help for different art styles. 
2. Maybe it would make sense to apply a specular occlusion attenuation function like the Frostbite one Filament uses here? That might be a better solution than the current flat adjustment based on the roughness/metallic values. <https://github.com/google/filament/blob/main/shaders/src/ambient_occlusion.fs#L85>

Calinou:
> Maybe it would make sense to apply a specular occlusion attenuation function like the Frostbite one Filament uses here? That might be a better solution than the current flat adjustment based on the roughness/metallic values. 

See https://github.com/godotengine/godot/pull/50601.

guerro323:
> Would it make sense to have the intensity of specular AO from lightmap be a user-adjustable value? What intensity in the lightmap is "dark" depends a lot on the scene.  

To what it's considered dark depends on what the camera see on the lightmapped object so it takes the current exposure in account.  
The way it work, is that if a fully rough and non metallic material pixel is dark from the camera then it should be the same for a metallic and reflective material pixel in `Reduce` mode; it can be brighter when using the `Conservative` mode and if the pixel isn't totally black, but if you have good reflection sources then it should be approximately correct).  

Having it as an user-adjustable value would complicate it even further and may introduce a slight cost of performance (and this effect is intended to be cheap and used on mobile and low-end PCs, **if you need a correct look you need to use directional lightmaps**).

> It seems like you're suggesting the simple/complex mode switch should be used for determining whether the user want more or less specular AO, but that seems like it makes what the specular AO is doing less explicit.   

For the conservative part, I'm not really sure how it's less explicit as it's an approximation of what Blender Cycles does, you can see a quick comparison between blender and godot here:  
(*top is Blender, and bottom is with conservative specular occlusion and some reflection probes*)
|Full Diffuse|Roughness=0|Metallic=1|Roughness=0 And Metallic=1|
:---:|:---:|:---:|:---:
image | image | image | image | 
image | image | image | image

|Full Diffuse|Roughness=0|Metallic=1|Roughness=0 And Metallic=1|
:---:|:---:|:---:|:---:
image | image | image | image |
image | image | image | image

WickedInsignia:
In my brief tests this works spectacularly well. Would be interesting to see a slider to adjust how much specular occlusion is present, or to have it tied to an individual material rather than the whole scene (if you've ever used Silent's Filamented Shader in Unity, it's a really great example of offering artistic options for this technique). I understand there may be some limitations to those options though.
Regardless, great work on this. It adds a lot of quality to lightmapping for reflective surfaces.

atirut-w:
(Moved from #86135)

After playing with this PR + #86135 for some time, I don't think doing the occlusion based on indirect lighting lightmap. There's the obvious issue with differences between static & dynamic lights, but also scenarios where this won't look accurate.

One of those scenarios would be a dimly lit room with window(s) to the outside:
image

The reflections here are not correct. Here is how it looks with only SDFGI:
image

Maybe we should bake a sort of "occlusion map" instead?

WickedInsignia:
@atirut-w Specular occlusion as it's implemented here doesn't seem to handle high dynamic range very well, so having areas of great contrast will mean that too little or too much is being occluded. It would be best to keep scene lighting at a fairly constant brightness when using this technique.
To my knowledge there's no setting present to reduce the light range that is being occluded, but I've seen this done in some shaders for Unity (such as Silent's Filamented).

This technique is fine and needed. It's not very physically accurate but lends much improvement to lightmapped reflections at little cost. Baking out a scene-wide ambient occlusion map might help in instances that a higher dynamic range is needed at the cost of memory footprint, but I don't think it should be used instead.

A slider to adjust what Godot considers dark enough to occlude, PLUS the ability to use a scene-wide AO bake instead of the normal lightmap if desired would probably be best.
It would also be worth keeping in mind that #86135 is available if high dynamic range lighting is needed since it is using DynamicGI for reflections instead, at a higher performance overhead than specular occlusion.

guerro323:
> A slider to adjust what Godot considers dark enough to occlude, PLUS the ability to use a scene-wide AO bake instead of the normal lightmap if desired would probably be best.  

For a greater control, I think it would be better to have a better support for it in custom shaders, like how the `light()` function works:
```glsl
void lightmap() {
    AMBIANT_LIGHT += LIGHTMAP_COLOR * LIGHTMAP_EXPOSURE;
    INDIRECT_LIGHT *= length(LIGHTMAP_COLOR) * LIGHTMAP_EXPOSURE;
}
```
The problem with that is that there needs to be a lot of stuff exposed for it; and it also needs to work for directional lightmaps, probes and should work with #86135 ...
___
For a simple slider, while it can be exposed in the LightmapGI node I'm not sure if I could instead expose it in the material shader (eg: `SPECULAR_OCCLUSION_INTENSITY`) as there were already issues when the two occlusion modes were a render mode (https://github.com/godotengine/godot/pull/86102#issuecomment-1854949786).

WickedInsignia:
> I think it would be better to have a better support for it in custom shaders

Having it as a per-material setting that's exposed in StandardMaterial and custom shaders (whether they be node or code-based) would be ideal, definitely. Sometimes you want to control how each independent material handles specular occlusion, and it also allows for a different setting on duplicate materials that are used in darker areas to overcome high dynamic range issues.

Calinou (approved):
Tested locally (rebased against `master` https://github.com/godotengine/godot/commit/17e7f85c06366b427e5068c5b3e2940e27ff5f1d), it works as expected. This is a great visual improvement :slightly_smiling_face: 

Example with a reflection probe:

### Conservative (default)

Screenshot_20240127_202948

### Reduce

Screenshot_20240127_203006

### Disabled

Screenshot_20240127_203017

Geometror:
Needs a rebase.

drivers/gles3/shaders/scene.glsl
```diff
@@ -1466,17 +1466,32 @@ void main() {
...
+						 2.0 * c1 * lightmap_captures[4].rgb * wnormal.x * wnormal.y +
+						 2.0 * c1 * lightmap_captures[7].rgb * wnormal.x * wnormal.z +
+						 2.0 * c1 * lightmap_captures[5].rgb * wnormal.y * wnormal.z +
+						 2.0 * c2 * lightmap_captures[3].rgb * wnormal.x +
+						 2.0 * c2 * lightmap_captures[1].rgb * wnormal.y +
+						 2.0 * c2 * lightmap_captures[2].rgb * wnormal.z) *
 				scene_data.emissive_exposure_normalization;
+
+#if defined(SPECULAR_OCCLUSION) || defined(SPECULAR_OCCLUSION_CONSERVATIVE)
+		float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // brightness from lightmap color
```
AThousandShips, line 1482:
```suggestion
		float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // Brightness from lightmap color.
```

doc/classes/ProjectSettings.xml
```diff
@@ -2484,6 +2484,13 @@
 		<member name="rendering/lightmapping/probe_capture/update_speed" type="float" setter="" getter="" default="15">
 			The framerate-independent update speed when representing dynamic object lighting from [LightmapProbe]s. Higher values make dynamic object lighting update faster. Higher values can prevent fast-moving objects from having "outdated" indirect lighting displayed on them, at the cost of possible flickering when an object moves from a bright area to a shaded area.
 		</member>
+		<member name="rendering/lightmapping/specular_occlusion/mode" type="int" setter="" getter="" default="2">
+			The specular occlusion method for object lighting from [LightmapGI]s.
+			- Disabled: No specular occlusion is used
+			- Reduce: Reduce the energy by the current [LightmapGI] pixel brightness
+			- Conservative: Conserve most of the energy by mixing metallic and roughness values, resulting in more realistic surface
```
AThousandShips, line 2491:
```suggestion
			- Conservative: Conserve most of the energy by mixing metallic and roughness values, resulting in a more realistic surface
```

drivers/gles3/shaders/scene.glsl
```diff
@@ -1505,7 +1520,24 @@ void main() {
 			specular_light += lm_light_l1p1 * 0.32573 * r.x * lightmap_exposure_normalization;
 		}
 #else
+#if defined(SPECULAR_OCCLUSION) || defined(SPECULAR_OCCLUSION_CONSERVATIVE)
+		vec3 c = textureLod(lightmap_textures, uvw, 0.0).rgb * lightmap_exposure_normalization;
+
+		float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // brightness from lightmap color
```
AThousandShips, line 1526:
```suggestion
		float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // Brightness from lightmap color.
```

drivers/gles3/storage/material_storage.cpp
```diff
@@ -1361,6 +1361,19 @@ MaterialStorage::MaterialStorage() {
 		actions.render_mode_defines["unshaded"] = "#define MODE_UNSHADED\n";
 		actions.render_mode_defines["fog_disabled"] = "#define FOG_DISABLED\n";
 
+		int specular_occlusion_mode = GLOBAL_GET("rendering/lightmapping/specular_occlusion/mode");
+
+		switch (specular_occlusion_mode) {
+			case 0:
+				break; // disabled, skip
```
AThousandShips, line 1368:
```suggestion
				break; // Disabled, skip.
```

servers/rendering/renderer_rd/forward_clustered/scene_shader_forward_clustered.cpp
```diff
@@ -719,6 +719,19 @@ void SceneShaderForwardClustered::init(const String p_defines) {
 		actions.render_mode_defines["debug_shadow_splits"] = "#define DEBUG_DRAW_PSSM_SPLITS\n";
 		actions.render_mode_defines["fog_disabled"] = "#define FOG_DISABLED\n";
 
+		int specular_occlusion_mode = GLOBAL_GET("rendering/lightmapping/specular_occlusion/mode");
+
+		switch (specular_occlusion_mode) {
+			case 0:
+				break; // disabled, skip
```
AThousandShips, line 726:
```suggestion
				break; // Disabled, skip.
```

servers/rendering/renderer_rd/shaders/forward_clustered/scene_forward_clustered.glsl
```diff
@@ -1379,18 +1379,34 @@ void fragment_shader(in SceneData scene_data) {
...
+						 2.0 * c1 * lightmap_captures.data[index].sh[4].rgb * wnormal.x * wnormal.y +
+						 2.0 * c1 * lightmap_captures.data[index].sh[7].rgb * wnormal.x * wnormal.z +
+						 2.0 * c1 * lightmap_captures.data[index].sh[5].rgb * wnormal.y * wnormal.z +
+						 2.0 * c2 * lightmap_captures.data[index].sh[3].rgb * wnormal.x +
+						 2.0 * c2 * lightmap_captures.data[index].sh[1].rgb * wnormal.y +
+						 2.0 * c2 * lightmap_captures.data[index].sh[2].rgb * wnormal.z) *
 				scene_data.emissive_exposure_normalization;
 
+#if defined(SPECULAR_OCCLUSION) || defined(SPECULAR_OCCLUSION_CONSERVATIVE)
+		float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // brightness from lightmap color
```
AThousandShips, line 1396:
```suggestion
		float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // Brightness from lightmap color.
```

servers/rendering/renderer_rd/shaders/forward_clustered/scene_forward_clustered.glsl
```diff
@@ -1421,7 +1437,24 @@ void fragment_shader(in SceneData scene_data) {
 			}
 
 		} else {
+#if defined(SPECULAR_OCCLUSION) || defined(SPECULAR_OCCLUSION_CONSERVATIVE)
+			vec3 c = textureLod(sampler2DArray(lightmap_textures[ofs], SAMPLER_LINEAR_CLAMP), uvw, 0.0).rgb * lightmaps.data[ofs].exposure_normalization;
+
+			float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // brightness from lightmap color
```
AThousandShips, line 1443:
```suggestion
			float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // Brightness from lightmap color.
```

servers/rendering/renderer_rd/shaders/forward_mobile/scene_forward_mobile.glsl
```diff
@@ -1138,18 +1138,33 @@ void main() {
...
+						 2.0 * c1 * lightmap_captures.data[index].sh[4].rgb * wnormal.x * wnormal.y +
+						 2.0 * c1 * lightmap_captures.data[index].sh[7].rgb * wnormal.x * wnormal.z +
+						 2.0 * c1 * lightmap_captures.data[index].sh[5].rgb * wnormal.y * wnormal.z +
+						 2.0 * c2 * lightmap_captures.data[index].sh[3].rgb * wnormal.x +
+						 2.0 * c2 * lightmap_captures.data[index].sh[1].rgb * wnormal.y +
+						 2.0 * c2 * lightmap_captures.data[index].sh[2].rgb * wnormal.z) *
 				scene_data.emissive_exposure_normalization;
 
+#if defined(SPECULAR_OCCLUSION) || defined(SPECULAR_OCCLUSION_CONSERVATIVE)
+		float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // brightness from lightmap color
```
AThousandShips, line 1154:
```suggestion
		float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // Brightness from lightmap color.
```

servers/rendering/renderer_rd/shaders/forward_mobile/scene_forward_mobile.glsl
```diff
@@ -1180,7 +1195,24 @@ void main() {
 			}
 
 		} else {
+#if defined(SPECULAR_OCCLUSION) || defined(SPECULAR_OCCLUSION_CONSERVATIVE)
+			vec3 c = textureLod(sampler2DArray(lightmap_textures[ofs], SAMPLER_LINEAR_CLAMP), uvw, 0.0).rgb * lightmaps.data[ofs].exposure_normalization;
+
+			float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // brightness from lightmap color
```
AThousandShips, line 1201:
```suggestion
			float specular_occlusion = (c.r * 0.3 + c.g * 0.59 + c.b * 0.11) * 2.0; // Brightness from lightmap color.
```

AThousandShips (approved):

passivestar:
Tested, couldn't make it work with a reflection probe, lightmap has no effect on reflections.

This is with/without a reflection probe:

gif

Tried both Reduce and Conservative, different lightmap settings, different probe settings, different material settings, different godot versions (PR's original 4.3 and 4.5 master rebase). Not sure what I could be missing

Haven't tested on other things such as sky or sdfgi

jcostello:
@passivestar can you share the test scene? Last time I tried it worked for me

lander-vr:
Just tested it as well and it doesn't seem to be working on my end either, which is curious since it doesn't look like there has been an update to this PR since Calinous last test (https://github.com/godotengine/godot/pull/86102#pullrequestreview-1847206655).

With vs w/o ref probe:
image
image

PR #86103, opened 2023-12-13
Fix volumetric fog artifacts when inside the fog

### LINKED ISSUE #74790: Volumetric Fog shows artifacts at camera edge at certain angles inside Fog Volumes
### Godot version

### System information

Windows 10, Vulkan

### Issue description

Closer to the edges of Fog Volumes (but still at reasonable distance) certain camera angles will display very noticeable artifacts that cover part of the camera. None of the project settings related to Volumetric Fog or Environment settings seem to alleviate the problem. Camera's near plane value also doesn't seem to help.

Keep in mind, the camera is still clearly inside the fog volume. The problem occurs in all axes.

### Steps to reproduce

It's visible in the editor in the MRP, position the camera near the edge of the FogVolume and look around as shown.

[volfog bug.zip](https://github.com/godotengine/godot/files/10949332/volfog.bug.zip)

Posting video at bottom because anything after seems to get eaten by github for some reason.

<video src="https://user-images.githubusercontent.com/115422160/224508971-761a49d5-b2a5-4c54-8adc-85a2dfa93b71.webm">

ecmjohnson:
Resolves #74790 by forcing the froxel volume to the full frame whenever the camera could have its near plane inside the fog volume. This uses the AABB of the fog volume as a proxy for the fog volume itself. Depth can still be culled based on the farthest corner in froxel space.

testing_fix

The min/max corners of the fog volume are tracked while mapping each corner into froxel space. This constructs an AABB around the transformed fog volume. This AABB is then expanded by the distance from the camera position to the corner of the near plane. If the camera is inside this AABB, then only the z depth froxel coordinate can be safely culled. Otherwise, I think considering just the corners is correct.

I renamed the min/max variables to froxel_min/froxel_max to be avoid confusion with the minimum/maximum world-space corners of the fog volume. I also renamed the points variable to froxels, again to be clear these points are in froxel space.

### DIFF
--- a/servers/rendering/renderer_rd/environment/fog.cpp
+++ b/servers/rendering/renderer_rd/environment/fog.cpp
@@ -622,6 +622,7 @@ void Fog::volumetric_fog_update(const VolumetricFogSettings &p_settings, const P
 
 		RD::ComputeListID compute_list = RD::get_singleton()->compute_list_begin();
 		bool any_uses_time = false;
+		Vector3 cam_position = p_cam_transform.get_origin();
 
 		for (int i = 0; i < (int)p_fog_volumes.size(); i++) {
 			FogVolumeInstance *fog_volume_instance = fog_volume_instance_owner.get_or_null(p_fog_volumes[i]);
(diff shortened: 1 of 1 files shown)


servers/rendering/renderer_rd/environment/fog.cpp
```diff
@@ -654,41 +655,68 @@ void Fog::volumetric_fog_update(const VolumetricFogSettings &p_settings, const P
...
+					froxel_min.z = 0;
+					froxel_max.x = int32_t(fog->width);
+					froxel_max.y = int32_t(fog->height);
+					for (int j = 0; j < 8; j++) {
+						froxel_max.z = MAX(froxel_max.z, froxels[j].z);
+					}
+				} else {
+					// Camera is guaranteed to be outside the fog volume
+					for (int j = 0; j < 8; j++) {
+						froxel_min = Vector3i(MIN(froxel_min.x, froxels[j].x), MIN(froxel_min.y, froxels[j].y), MIN(froxel_min.z, froxels[j].z));
```
AThousandShips, line 710:
```suggestion
						froxel_min = froxel_min.min(froxels[j]);
```
Use dedicated method here as well

servers/rendering/renderer_rd/environment/fog.cpp
```diff
@@ -654,41 +655,68 @@ void Fog::volumetric_fog_update(const VolumetricFogSettings &p_settings, const P
...
+						froxel_max = Vector3i(MAX(froxel_max.x, froxels[j].x), MAX(froxel_max.y, froxels[j].y), MAX(froxel_max.z, froxels[j].z));
```
AThousandShips, line 711:
```suggestion
						froxel_max = froxel_max.max(froxels[j]);
```

servers/rendering/renderer_rd/environment/fog.cpp
ecmjohnson: Oh yeah, of course!

clayjohn (approved):
The code looks great to me! Sorry this slipped through the cracks for so long

Repiteo:
Thanks!

idchlife:
@ecmjohnson thank you!

Anzo1992:
I downloaded the Godot 4.4-dev7 version and saw that the fix was implemented in Godot 4.4-dev4 version. 

I tested a scene and it's still showing some cutting/clipping when moving the camera around. 

This is how the fog looks like once i load the scene.
Normal(standing)

This is how it looks when I move the camera around:
Moving_camera

PR #86105, opened 2023-12-13
Fix SpotLight and OmniLight cull masks not working in mobile renderer

### LINKED ISSUE #82263: Mobile/Compatibility: SpotLight and OmniLight cull masks don't cull moving meshes
### Godot version

v4.1.1.stable.official[bd6af8e0e]

### System information

Windows 10 - Godot 4.1.1 Stable - Mobile

### Issue description

Both Omni lights and spotlights will light up a mesh that has been transformed, despite there being no overlapping layers. This does not appear to be a problem with the Forward+ Renderer nor with directional lights on the mobile renderer.

In this example I have two lights, one is set to affect all layers, the light is periodically toggled on and off,
another light is on all of the time, but it's cull mask is only set to affect layer 2, which neither of the Mesh Instances are on.
The expected behaviour is that when the first light is toggled, the cubes will no longer be visible,

Expected Behaviour

as works on the Forward+ renderer. However when using the mobile renderer, Meshes will reappear again if they are moved (via script) and will not disappear even if their movement is stopped.

Mobile Behaviour

I first noticed this behaviour when I was attempting to do an effect where all but one object would fade to black because it was the only mesh on a specific layer (which the spotlight cull was also set to) however other objects still appeared despite not being on that layer because they had been or were being moved.

### Steps to reproduce

Start a project with the mobile renderer

Create a 3D scene with a mesh that can be moved.
image
image
image

Create a spot light or omni light near the mesh so that it is noticeably lit.
image

Set the Cull Mask of the light to only affect layers that do not overlap with the cube's VisualInstance3D layers
image

The cube will correctly not be lit by the spotlight upon first playing, but once the cube has been moved, it will suddenly be lit up.
GIF 2023-09-24 1-28-41 PM

[Light Cull Mask Mini Project.zip](https://github.com/godotengine/godot/files/12709615/Light.Cull.Mask.Mini.Project.zip)

*Production edit (keywords for easier searching): OpenGL*

jsjtxietian:
Fixes #82263

I'm not sure whether there's any unexpected impacts that prevents us from adding it at the first place, feel free to correct me.

### DIFF
--- a/servers/rendering/renderer_rd/shaders/forward_mobile/scene_forward_mobile.glsl
+++ b/servers/rendering/renderer_rd/shaders/forward_mobile/scene_forward_mobile.glsl
@@ -1646,6 +1646,10 @@ void main() {
 				break;
 			}
 
+			if (!bool(omni_lights.data[light_index].mask & instances.data[draw_call.instance_index].layer_mask)) {
+				continue; //not masked
+			}
+
 			float shadow = light_process_omni_shadow(light_index, vertex, normal);
 
 			shadow = blur_shadow(shadow);
@@ -1689,6 +1693,10 @@ void main() {
 
 			if (light_index == 0xFF) {
 				break;
+			}
+
+			if (!bool(spot_lights.data[light_index].mask & instances.data[draw_call.instance_index].layer_mask)) {
+				continue; //not masked
 			}
 
 			float shadow = light_process_spot_shadow(light_index, vertex, normal);


BastiaanOlij:
Hmm, this seems like a stopgap measure. In the mobile renderer the layer mask should already be applied during the culling stage as lights are paired with the objects they illuminate. A light should never end up being paired to an object if the layer mask doesn't match. Double checking it in the shader should yield no difference, only introduce more overhead.

If there are object lit by lights that shouldn't light it, then there is an issue in the culling system.

jsjtxietian:
> In the mobile renderer the layer mask should already be applied during the culling stage as lights are paired with the objects they illuminate.

Interesting, so the real bug is lying in the cpu code ? I will take a look.

Repiteo:
Superseded by #98266

PR #86118, opened 2023-12-13
Fix so undoing complex operations in `TextEdit` will restore selections

### LINKED ISSUE #72769: Undoing TextEdit complex operation doesn't revert all selections
### Godot version

v4.0.beta.custom_build [13f0158]

### System information

Windows 10

### Issue description

When doing operations (`begin_complex_operation` and `end_complex_operation`) with multiple selections in TextEdit, some of these selections don't revert upon undo. This behavior is affects some of script editor's operations, like duplicate selection and delete lines.

### Steps to reproduce

Open script editor, make multiple selections. Keep in mind what you have selected. Do duplicate selection (ctrl+shift+D). Undo (ctrl+Z). Compare the result to what you remember.

N/A

TheSofox:
Fixes #72769

When doing a complex operation in TextEdit (using `begin_complex_operation` and `end_complex_operation` to start and end them), undoing one of these operations should have restored the original text selections, however in some cases (especially with multicarat selections), the selections wouldn't be properly restored.

It seems that there was an effort to code this in before, but it was either not finished or more likely broken by subsequent code changes. This PR fixes this using the existing system.

### DIFF
--- a/scene/gui/text_edit.cpp
+++ b/scene/gui/text_edit.cpp
@@ -7690,7 +7690,11 @@ void TextEdit::_insert_text(int p_line, int p_char, const String &p_text, int *r
 	op.version = ++version;
 	op.chain_forward = false;
 	op.chain_backward = false;
-	op.start_carets = carets;
+	if (next_operation_is_complex) {
+		op.start_carets = current_op.start_carets;
+	} else {
+		op.start_carets = carets;
+	}
 	op.end_carets = carets;
 
 	// See if it should just be set as current op.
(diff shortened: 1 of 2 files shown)


elfajriharsuda (approved):

Paulb23 (commented):
LGTM.

Nice spot, looks like it was missed out as we set `current_op.start_carets;` in `begin_complex_operation` but as it's `TextOperation::TYPE_NONE;` it gets overwritten by the new `op` thus we lose the original caret state.

Could you add a unit test to capture this behaviour?

TheSofox:
Amended commit with Unit Test. I confirmed it failed on master but passed on this branch. It tests for the exact behaviour that this PR fixes.

Paulb23 (approved):

YuriSizov:
Thanks!

Issue #53288, opened 2021-10-01
Raster occlusion incorrectly culls objects through gaps


lawnjelly:
### Godot version

4.0 dev df57aa670

### System information

Linux Mint 20, Intel HD graphics 630

### Issue description

Objects are being culled when they should be visible, in cases where there are small gaps between occluders.

Setup:
![Screenshot from 2021-10-01 08-45-37](https://user-images.githubusercontent.com/21999379/135585233-039786eb-37f8-4f7a-8a8d-46dd109947f8.png)
What should be seen:
![Screenshot from 2021-10-01 08-45-23](https://user-images.githubusercontent.com/21999379/135585250-5e2be9ae-8c9e-4e34-8489-42ccdbbd1690.png)
What is seen:
![Screenshot from 2021-10-01 08-45-09](https://user-images.githubusercontent.com/21999379/135585255-760fc5fa-c407-4634-bb2f-a0a2181b5a13.png)
Occlusion buffer:
![Screenshot from 2021-10-01 08-46-23](https://user-images.githubusercontent.com/21999379/135585455-97febc76-f351-4a0f-8863-a83a06d1a449.png)

### Steps to reproduce

Create a couple of boxes as occluders and place them with a small gap, then attempt to view a 3rd object through the gap. Vary the viewing angle / distance.

### Minimal reproduction project

[TestOcclusion.zip](https://github.com/godotengine/godot/files/7265271/TestOcclusion.zip)

### Discussion

I'm just having a look at some situations likely to trip up the raster occlusion, especially edges / joins between objects. I'm guessing this is caused by the rendering of the occlusion buffer not being conservative (or rather the opposite of conservative rasterization :smile: ), i.e. gaps between objects are being joined. This makes occlusion more effective, but produces visual artifacts.

This one can be quite disturbing as objects flicker in and out of view according to the viewing angle, and whether or not a gap is rendered in the occlusion buffer.

tinmanjuggernaut:
On Alpha13 if the gap is wide enough it works fine.

https://user-images.githubusercontent.com/632766/182120007-b8321dd2-6f48-4851-bb69-9249adb6a7e9.mp4

To trigger it requires two objects occluding something, it doesn't work with one. And the effective gap between them has to be narrow. Here the objects have a wide gap, but because of the angle, the effective gap has made the red object disappear. The difference between left and right is just camera angle, as the red box is flickering when I move the camera. Interesting how different the occlusion mask is.

[...]

PR #86121, opened 2023-12-13
Jitter raster occlusion camera to reduce false positives.

lawnjelly:
Due to the low resolution of the occlusion buffer, small gaps between occluders can be closed and incorrectly occlude instances which should show through the gaps. To ameliorate this problem, this PR jitters the occlusion buffer over time, making it more likely an instance will be seen through a gap. This is used in conjunction with an occlusion timer per instance, to prevent instances flickering on and off rapidly.

Helps address #53288

Actually this PR is _very effective_ at combating what is a difficult problem, with the only cost being a slight increase in margins for instances falsely shown when they should be occluded.

https://github.com/godotengine/godot/assets/21999379/b724686f-921d-4c63-ad1c-03c0b1688650

This can be switched on and off with project setting, which defaults to true:
`rendering/occlusion_culling/jitter_projection`
When off, the jittering and temporal "stickiness" will be turned off to give the same as legacy behaviour.

## Discussion
While I was working on software rasterizer for 3.x, I was pondering whether this could be used for raster occlusion, which reminded me of #53288 which is the Achilles heel of raster occlusion at low resolution.

I hadn't thought of this originally when filing the bug, but it occurred to me that one way of dealing with the problem is to simply jitter the occlusion camera. Providing the jitter is approximately on the scale of pixels of the buffer, then it increases the likelihood that instances will be seen through narrow gaps, which is exactly what we want. There is almost no cost to the technique (jittering is almost free, and the timers are cheap).

On the downside, this will make instances visible slightly further around occluded objects (as this is essentially how it works), however in most cases objects are expected to be occluded behind walls etc, so the cost may not be much in reality. Additionally, some cost is justifiable in order to prevent the artifacts in #53288, which may currently make raster occlusion unusable in some situations.

## Jitter scale
Jitter scale is currently such that the jitter is over approximately a pixel in the occlusion buffer. This approximately effectively increases the resolution of the occlusion buffer. The smaller the intervals within, the higher the resolution, but we would need to jitter over a larger number of frames.

Another alternative is to randomize the jitter, but this makes it difficult to calculate a deterministic instance timeout that prevents the possibility of flashing with a non-moving camera, so this option has been rejected for now.

## Notes
* It will not deal with _all_ the associated problems. Especially new objects appearing for the first time in gaps may still be hidden for a couple of frames. This is partly why the "cycle" length is kept low (9 frames) to ensure that if something is going to appear, it will appear soon.
* Cosmetic - when you view the occlusion buffer in the editor, you can see the jittering.
* Another alternative is identifying gaps, and increasing the occlusion buffer resolution in gaps. This is likely considerably more complex to implement though.
* Another alternative is to jitter _per ray_, which could make it more effective, at the cost of complexity.

### DIFF
--- a/doc/classes/ProjectSettings.xml
+++ b/doc/classes/ProjectSettings.xml
@@ -2643,6 +2643,9 @@
 			The [url=https://en.wikipedia.org/wiki/Bounding_volume_hierarchy]Bounding Volume Hierarchy[/url] quality to use when rendering the occlusion culling buffer. Higher values will result in more accurate occlusion culling, at the cost of higher CPU usage. See also [member rendering/occlusion_culling/occlusion_rays_per_thread].
 			[b]Note:[/b] This property is only read when the project starts. To adjust the BVH build quality at runtime, use [method RenderingServer.viewport_set_occlusion_culling_build_quality].
 		</member>
+		<member name="rendering/occlusion_culling/jitter_projection" type="bool" setter="" getter="" default="true">
+			If [code]true[/code], the projection used for rendering the occlusion buffer will be jittered. This can help prevent objects being incorrectly culled when visible through small gaps.
+		</member>
 		<member name="rendering/occlusion_culling/occlusion_rays_per_thread" type="int" setter="" getter="" default="512">
 			The number of occlusion rays traced per CPU thread. Higher values will result in more accurate occlusion culling, at the cost of higher CPU usage. The occlusion culling buffer's pixel count is roughly equal to [code]occlusion_rays_per_thread * number_of_logical_cpu_cores[/code], so it will depend on the system's CPU. Therefore, CPUs with fewer cores will use a lower resolution to attempt keeping performance costs even across devices. See also [member rendering/occlusion_culling/bvh_build_quality].
 			[b]Note:[/b] This property is only read when the project starts. To adjust the number of occlusion rays traced per thread at runtime, use [method RenderingServer.viewport_set_occlusion_rays_per_thread].
--- a/servers/rendering/renderer_scene_occlusion_cull.cpp
+++ b/servers/rendering/renderer_scene_occlusion_cull.cpp
@@ -43,6 +43,8 @@ const Vector3 RendererSceneOcclusionCull::HZBuffer::corners[8] = {
 	Vector3(1, 1, 1)
 };
 
+bool RendererSceneOcclusionCull::HZBuffer::occlusion_jitter_enabled = false;
+
 bool RendererSceneOcclusionCull::HZBuffer::is_empty() const {
 	return sizes.is_empty();
 }
@@ -66,6 +68,8 @@ void RendererSceneOcclusionCull::HZBuffer::clear() {
 }
 
 void RendererSceneOcclusionCull::HZBuffer::resize(const Size2i &p_size) {
+	occlusion_buffer_size = p_size;
+
 	if (p_size == Size2i()) {
 		clear();
 		return;
@@ -124,6 +128,9 @@ void RendererSceneOcclusionCull::HZBuffer::resize(const Size2i &p_size) {
 }
 
 void RendererSceneOcclusionCull::HZBuffer::update_mips() {
+	// Keep this up to date as a local to be used for occlusion timers.
+	occlusion_frame = Engine::get_singleton()->get_frames_drawn();
+
 	if (sizes.is_empty()) {
 		return;
 	}
--- a/core/config/project_settings.cpp
+++ b/core/config/project_settings.cpp
@@ -1523,6 +1523,8 @@ ProjectSettings::ProjectSettings() {
 	GLOBAL_DEF("debug/settings/crash_handler/message.editor",
 			String("Please include this when reporting the bug on: https://github.com/godotengine/godot/issues"));
 	GLOBAL_DEF_RST(PropertyInfo(Variant::INT, "rendering/occlusion_culling/bvh_build_quality", PROPERTY_HINT_ENUM, "Low,Medium,High"), 2);
+	GLOBAL_DEF_RST("rendering/occlusion_culling/jitter_projection", true);
+
 	GLOBAL_DEF_RST("internationalization/rendering/force_right_to_left_layout_direction", false);
 	GLOBAL_DEF_BASIC(PropertyInfo(Variant::INT, "internationalization/rendering/root_node_layout_direction", PROPERTY_HINT_ENUM, "Based on Application Locale,Left-to-Right,Right-to-Left,Based on System Locale"), 0);
 
--- a/modules/raycast/raycast_occlusion_cull.cpp
+++ b/modules/raycast/raycast_occlusion_cull.cpp
@@ -536,6 +536,64 @@ void RaycastOcclusionCull::buffer_set_size(RID p_buffer, const Vector2i &p_size)
 	buffers[p_buffer].resize(p_size);
 }
 
+Projection RaycastOcclusionCull::_jitter_projection(const Projection &p_cam_projection, const Size2i &p_viewport_size) {
+	if (!_jitter_enabled) {
+		return p_cam_projection;
+	}
+
+	// Prevent divide by zero when using NULL viewport.
+	if ((p_viewport_size.x <= 0) || (p_viewport_size.y <= 0)) {
+		return p_cam_projection;
+	}
+
+	Projection p = p_cam_projection;
+
+	int32_t frame = Engine::get_singleton()->get_frames_drawn();
+	frame %= 9;
+
+	Vector2 jitter;
+
+	switch (frame) {
+		default:
+			break;
+		case 1: {
+			jitter = Vector2(-1, -1);
+		} break;
+		case 2: {
+			jitter = Vector2(1, -1);
+		} break;
+		case 3: {
+			jitter = Vector2(-1, 1);
+		} break;
+		case 4: {
+			jitter = Vector2(1, 1);
+		} break;
+		case 5: {
+			jitter = Vector2(-0.5f, -0.5f);
+		} break;
+		case 6: {
+			jitter = Vector2(0.5f, -0.5f);
+		} break;
+		case 7: {
+			jitter = Vector2(-0.5f, 0.5f);
+		} break;
+		case 8: {
+			jitter = Vector2(0.5f, 0.5f);
+		} break;
+	}
+
+	// The multiplier here determines the divergence from center,
+	// and is to some extent a balancing act.
+	// Higher divergence gives fewer false hidden, but more false shown.
+	// False hidden is obvious to viewer, false shown is not.
+	// False shown can lower percentage that are occluded, and therefore performance.
+	jitter *= Vector2(1 / (float)p_viewport_size.x, 1 / (float)p_viewport_size.y) * 0.05f;
+
+	p.add_jitter_offset(jitter);
+
+	return p;
+}
+
 void RaycastOcclusionCull::buffer_update(RID p_buffer, const Transform3D &p_cam_transform, const Projection &p_cam_projection, bool p_cam_orthogonal) {
 	if (!buffers.has(p_buffer)) {
 		return;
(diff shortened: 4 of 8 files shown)


passivestar:
Jitter off:

https://github.com/godotengine/godot/assets/60579014/3239595d-5ab4-4f76-b0a0-e5a29458a839

Jitter on:

https://github.com/godotengine/godot/assets/60579014/e8be6b5b-193c-44e6-a525-774d89e49e8b

Tried different multipliers for the jitter (0.5f, 1.0f), not much difference there

Also tried using 12 frames instead of 8 but not much luck with that either. Not sure how to make this work

lawnjelly:
> Tried different multipliers for the jitter (0.5f, 1.0f), not much difference there

If you have example scene I can try this out and figure why not working.

passivestar:
>If you have example scene I can try this out and figure why not working

Here you go [MRP.zip](https://github.com/godotengine/godot/files/14637284/MRP.zip)

lawnjelly:
Ah I think it could just a small problem with my timer logic (wrongly assuming `is_occluded()` is called once per occlusion buffer render), will fix.

passivestar:
Jitter 0.25f:

https://github.com/godotengine/godot/assets/60579014/da93997f-1fa6-47c7-b11e-972ae27b89b5

Jitter 5.0f:

https://github.com/godotengine/godot/assets/60579014/26933394-9d1a-4a62-8f2f-a115101d1c9e

Still possible to spot culling even at 5.0f but it can probably be worked around with level design. And it's much better than what is currently in godot

It's probably best to expose the jitter value to the editor settings and get rid of the checkbox (let 0 mean "disabled")

lawnjelly:
> Jitter 0.25f:

I can't seem to get it to behave like that in mine, which makes me wonder if there is a screen size dependency that isn't working correctly (the 0.25f jitter should work the same whatever the size of the occlusion buffer if it works correctly). :thinking: 

Are you taking the videos by moving the viewer position in the editor?

I managed to get some flickering by looking at one of the mountains through the corridor, which was how I fixed the above bug. It's much better for me now than before, but note that jittering won't solve all possible cases, notably:

* Instances coming into view you may still get a flicker until the jittered render "finds" them.
* The jitters are not random, and can in some situations just miss the small gap required to see an object. This is something that could possibly be refined later.

passivestar:
> which makes me wonder if there is a screen size dependency that isn't working correctly

I think that's the case. At 0.25f, here's the biggest gap I can catch at native resolution:

Screenshot 2024-03-18 at 21 34 42<br>

And when I switch the viewport to half resolution this is the biggest gap (about twice as small):

Screenshot 2024-03-18 at 21 34 29<br>

> What screen resolution are you using for the window?

I'm on a laptop with a 3456x2234 screen

> Are you taking the videos by moving the viewer position in the editor?

Yeah I hold RMB, scroll all the way down to 0.2m/s and look for artifacts. Both position and camera angle contribute to them

> The jitters are not random, and can in some situations just miss the small gap required to see an object. This is something that could possibly be refined later.

Yeah I figured that much that's why I was experimenting with increasing the amount of frames with different offsets, but in the end of the day the value of 5.0f seems to be working fine even with 8 frames, and the only problem that's noticable with 5.0f in my case is the first one that you described (it just needs a couple of frames)

> Also, for most users, they will prefer something that just works (TM)

Yeah that's what I would prefer as well. Would be best if it culled in the safest way possible (only things that we're sure aren't visible), after all even when I set jitter very high and look at the overdraw it still looks near pixel perfect, so it just doesn't seems like culling too conservatively is ever gonna become a problem with godot's current culling approach

lawnjelly:
Yeah it seems likely the calculation of how much to jitter by resolution could be improved, and that's why you are needing to vary the parameter. I'll take a look. :+1:

passivestar:
Works really well with the latest changes. Biggest (persistent) gap that I could catch now is 4px wide on my screen (would be 2px in fullhd)

I believe if we combine this with an option to shrink occluders by normals at baking it will eliminate the remaining flashes and be good enough for 99.99% of the games

lawnjelly:
I've done some further tests to verify the settings, changing the occlusion buffer size by changing `occlusion_rays_per_thread` in project settings, and the jitter calculations do seem to be working correctly now, amazingly my 0.05 multiplier first guess seems approximately good (eyeballed) for around (or just over) a pixel whatever the occlusion buffer resolution, and it scales as expected with occlusion buffer size.

I've realized that your runtime occlusion buffer size is determined by how many cores your CPU has (as well as the resolution of the viewport), quite a complex calculation, so it's quite difficult to examine "like for like" without debugging or debug printing your occlusion buffer size or forcing it to a particular size.

_(If you place a breakpoint in `RaycastOcclusionCull::_jitter_projection()` you can examine `p_viewport_size` to find occlusion buffer size.)_

As far as I can see, the multiplier can be adjusted to prevent glancing false occlusion from one side of an occluder, but it probably can't guarantee to always hit tiny gaps _between_ occluders (but should usually work).

unfa:
I'm probably talking out of ignorance here but...
I wonder, why is the occlusion buffer rendered on the CPU?
We have compute shaders! Why not use that?
I'm sure we could run this raycasting thing on full res buffers and have zero mistakes. Or maybe downsample a full-sized occlusion buffer with a custom MIP filter that'd always round "down" so that any possible gaps between occluders are magnified, rather than hidden?

Calinou:
> I wonder, why is the occlusion buffer rendered on the CPU?
> We have compute shaders! Why not use that?

GPU-based solutions have a frame delay, which means you need to cater for it when camera cuts occur (e.g. in cutscenes). Many games introduce a 1-frame stutter on camera cuts because of this.

doc/classes/ProjectSettings.xml
```diff
@@ -2628,6 +2628,10 @@
 			The [url=https://en.wikipedia.org/wiki/Bounding_volume_hierarchy]Bounding Volume Hierarchy[/url] quality to use when rendering the occlusion culling buffer. Higher values will result in more accurate occlusion culling, at the cost of higher CPU usage. See also [member rendering/occlusion_culling/occlusion_rays_per_thread].
 			[b]Note:[/b] This property is only read when the project starts. To adjust the BVH build quality at runtime, use [method RenderingServer.viewport_set_occlusion_culling_build_quality].
 		</member>
+		<member name="rendering/occlusion_culling/jitter_projection" type="bool" setter="" getter="" default="true">
+			If [code]true[/code], the projection used for rendering the occlusion buffer will be jittered.
+			This can help prevent objects being incorrectly culled when visible through small gaps.
```
Calinou, line 2633: The description is short, so it can be on a single paragraph (every newline is a new paragraph):

```suggestion
			If [code]true[/code], the projection used for rendering the occlusion buffer will be jittered. This can help prevent objects being incorrectly culled when visible through small gaps.
```

Calinou (approved):

Code looks good to me.

unfa:
> . In contrast, the current CPU-based approach works on any rendering method, and could even work on a headless server with no GPU in theory (e.g. for server-side wallhack prevention).
Oh, that makes a lot of sense. Can we query the occlusion system to get stuff like this? It could also help not updating peers about stuff that's behind a mountain or in the room next door :D

Calinou:
> Oh, that makes a lot of sense. Can we query the occlusion system to get stuff like this?

Not currently, and I don't know if it'll be feasible without a frame delay. We try to minimize the amount of exposed RenderingServer methods because RenderingServer is designed to be able to run on multiple threads, so calling a getter method would require a slow synchronization process.

Chubercik:
Tested using the [`occlusion_culling_mesh_lod`](https://github.com/godotengine/godot-demo-projects/tree/master/3d/occlusion_culling_mesh_lod) example, works like a charm ^^

servers/rendering/renderer_scene_occlusion_cull.cpp
```diff
@@ -124,6 +128,11 @@ void RendererSceneOcclusionCull::HZBuffer::resize(const Size2i &p_size) {
 }
 
 void RendererSceneOcclusionCull::HZBuffer::update_mips() {
+	// Keep this up to date as used for occlusion timers.
+	// Note this will wrap around with extremely long uptime and high FPS,
+	// so we must deal with wraparound in is_occluded().
+	occlusion_frame = Engine::get_singleton()->get_frames_drawn();
```
clayjohn, line 134: `get_frames_drawn` returns a uint64_t, should we use the same for occlusion_frame so we can avoid the wraparound?

clayjohn (commented):
The idea seems pretty good and the implementation is really nice. I worry a little bit about under-occlusion though. 9 frames is a long time to leave something visible when it is fully occluded. 

My gut tells me we could get the same impact without the delay by shrinking occluder meshes. But, at the same time, shrinking occluder meshes is not as foolproof with respect to different buffer sizes and screen-space projected sizes. So maybe both are needed anyway

servers/rendering/renderer_scene_occlusion_cull.cpp
lawnjelly: The wraparound is dealt with in `renderer_scene_occlusion_cull.h`, line 196, it is fairly easy to detect and deal with:

```
// Detect wraparound.
// This can be considered somewhat optional,
// in case it is a bottleneck.
// The worst that will happen is that after wraparound
// an instance will not be occluded until it has been
// seen again (and thus the timeout is reset).
else if ((r_occlusion_timeout - occlusion_frame) >= 64) {
	r_occlusion_timeout = 0;
}
```

I think I did it as 32 bit originally to save on size in `InstanceData`, but maybe we won't have as many of these as e.g. 2D items, so maybe this is overkill, and it might be easier for readers.

lawnjelly:
From my point of view - I'm not aware of any obvious perfect solution here, with the occlusion buffer smaller than the window (as others have pointed out in #53288). Afaik we basically need some hacks to reduce the occlusion enough to prevent the problem in some way, to make up for taking the "small occlusion buffer" short cut.

> I worry a little bit about under-occlusion though. 9 frames is a long time to leave something visible when it is fully occluded.

It isn't ideal, but realistically the occlusion efficiency should still likely be >95% in most cases. In a lot of cases this might be e.g. a character still being rendered for 1/6 second after going through doorway. For users, the alternative at the moment is very broken occlusion (which may be unusable for some game releases).

> My gut tells me we could get the same impact without the delay by shrinking occluder meshes. But, at the same time, shrinking occluder meshes is not as foolproof with respect to different buffer sizes and screen-space projected sizes. So maybe both are needed anyway

Shrinking occluder meshes was mentioned in the issue, but I do suspect it could turn out to be a whole barrel of worms, given that it depends on the angle of view, distance from camera etc as you say:
https://github.com/godotengine/godot/issues/53288#issuecomment-1202718938

 It would be possible to pre-process and fix one scenario, but many others will remain broken. So shrinking might have to be done at runtime (equivalent of vertex shader, but in software), and there are a lot of potential bugs. Particularly you have to deal with shrinking adjacent walls potentially creating gaps and drastically reducing any occlusion, which probably requires pre-processing and marking edges.

Post-processing the occlusion buffer on the other hand could well be a simpler and better solution than shrinking (less bug inducing hopefully), and is well worth investigating, even if it adds some runtime cost. :thinking: This could later remove the need for jittering. I didn't have time to investigate this, but maybe another contributor will in the future.

Either way, I suspect jittering can likely do the job as a stop-gap until someone comes up with something better (bear in mind this problem has existed for > 3 years :fearful: ). It is a small change and relatively simple to add (and remove if we eventually come up with a better approach).

Chubercik (approved):
Code looks good, I'd love for this to get shipped alongside the Embree upgrade.

clayjohn (approved):
Thanks for your response. I agree, it makes sense to move forward with this. 

Shrinking meshes can be done quite safely/effectively when we generate them. But indeed, shrinking manually created meshes comes with tradeoffs and is worth investigation.

akien-mga:
Thanks!

ettiSurreal:
> I'm probably talking out of ignorance here but... I wonder, why is the occlusion buffer rendered on the CPU? We have compute shaders! Why not use that?

A bit late to this, but I imagine the true answer is that it could be optionally supported if someone were to actually implement it.

clayjohn:
> > I'm probably talking out of ignorance here but... I wonder, why is the occlusion buffer rendered on the CPU? We have compute shaders! Why not use that?
> 
> A bit late to this, but I imagine the true answer is that it could be _optionally_ supported on F+/mobile if someone were to actually implement it :)

See Calinou's [response](https://github.com/godotengine/godot/pull/86121#issuecomment-2027741820) above. 

Doing the occlusions on the GPU comes with a bunch of tradeoffs we don't want to accept. The main one being, it requires a 2 frame delay. The CPU version is super fast and works really well, so there is no point in having a GPU version that doesn't work as well. 

Eventually we will add a GPU-driven renderer (where almost all rendering takes place on the GPU). For that renderer, occlusion will be done purely on the GPU.

unfa:
I wonder however how the jittering will affect latency of the culling. Can't that also fail to immediately detect something is no longer occluded, resulting in random pop-in delay?

lawnjelly:
> I wonder however how the jittering will affect latency of the culling. Can't that also fail to immediately detect something is no longer occluded, resulting in random pop-in delay?

You could theoretically get e.g. a frame of delay if an AABB was a tiny fraction in view and it happened to be jittered in the other direction, but I've not seen this showing up in practice.

This is why the jitter pattern goes to opposite corners BTW, to prevent this occurring on more than a single frame (if it occurs at all).

PR #83452, opened 2023-10-16
Split `RenderingDevice` into API-agnostic and `RenderingDeviceDriver` parts

RandomShaper:
A huge chunk of `RenderingDeviceVulkan` is API-agnostic bookkeeping, validation, etc. Such code is also shared with others, like `RenderingDeviceD3D12`. This PR keeps a single, non-abstract `RenderingDevice` with all those API-agnostic elements and moves all the API-specific parts to a thin wrapper around the driver, which is `RenderingDeviceDriver` for Vulkan. This makes much easier to write additional rendering drivers, by allowing one to focus on the API-specific stuff. In some cases that will require modifying the interface between `RD` and `RDD`, which is not exposed to the user.

Other relevant points:
- A common context interface has been extracted from `VulkanContext` (`APIContextRD`) with the bare minimum set of services `RenderingDevice` and direct usage needs. (The specific `RenderingDeviceDriver`s have access to the specific context implementation, which open doors for them to a much richer, but non-standard, API.) **UPDATE:** Now Direct3D 12 has been merged, it gets similar love in this PR.
- There are important notes on design choice in the code. Future maintainers, please follow or discuss, but don't break consistency.
- Every struct involved now have default initializers, both in `RenderingDevice` and `RenderingDeviceDriver`.
- Some of the wiring between context and `RD` has been enhanced so now you can instantiate a `RenderingDevice` even when Godot is running in compatibility mode (this will allow to have a `RD`-based lightmapper available in a GL compatibility project). This piece of code shows how to ensure you have a usable local `RenderingDevice` in any circumstances:
   ```C++
   RD *orig_rd_singleton = RD::get_singleton();

   ApiContextRD *vc = nullptr;
   if (GLOBAL_GET("rendering/renderer/rendering_method") == "gl_compatibility") {
   #ifdef VULKAN_ENABLED
   	if (GLOBAL_GET("rendering/rendering_device/driver") == "vulkan") {
   		vc = memnew(VulkanContext);
	}
   #endif
   	CRASH_COND_MSG(!vc, "No RD driver available.");
   	vc->initialize();
   } else {
   	vc = RD::get_singleton()->get_context();
   }

   RenderingDevice *rd = memnew(RenderingDevice);
   if (GLOBAL_GET("rendering/renderer/rendering_method") == "gl_compatibility") {
   	DEV_ASSERT(RD::get_singleton() == rd);
   }
   rd->initialize(vc, true);
   // Do stuff with RD here.
   memdelete(rd);

[...]

PR #86123, opened 2023-12-13
Add a script to install Direct3D 12 SDK components

Calinou:
- Define a default path that uses the locations from the script.

This makes it much faster to get started with Direct3D 12 builds, as you only need to run `misc/scripts/install_d3d12_sdk_windows.ps1` then run `scons d3d12=yes`.

This installs DirectX Shader Compiler, Mesa NIR, WinPixEventRuntime and DirectX 12 Agility SDK.

See also https://github.com/godotengine/godot-docs/pull/8610.

## Preview

```
$ .\misc\scripts\install_d3d12_sdk_windows.ps1
Downloading DirectX Shader Compiler dxc_2023_08_14.zip ...
Extracting DirectX Shader Compiler dxc_2023_08_14.zip ...
DirectX Shader Compiler dxc_2023_08_14.zip installed successfully.

Downloading Mesa NIR godot-nir-23.1.0-1-devel.zip ...
Extracting Mesa NIR godot-nir-23.1.0-1-devel.zip ...
Mesa NIR godot-nir-23.1.0-1-devel.zip installed successfully.

Downloading WinPixEventRuntime 1.0.231030001 ...
Extracting WinPixEventRuntime 1.0.231030001 ...
WinPixEventRuntime 1.0.231030001 installed successfully.

Downloading DirectX 12 Agility SDK 1.611.2 ...
Extracting DirectX 12 Agility SDK 1.611.2 ...
DirectX 12 Agility SDK 1.611.2 installed successfully.

All Direct3D 12 SDK components were installed successfully!
You can now build Godot with Direct3D 12 support enabled by running "scons d3d12=yes" with MSVC installed (MinGW builds are currently not supported).
```


platform/windows/detect.py
```diff
@@ -188,11 +188,11 @@ def get_opts():
...
         # Direct3D 12 support.
-        ("mesa_libs", "Path to the MESA/NIR static libraries (required for D3D12)", ""),
-        ("dxc_path", "Path to the DirectX Shader Compiler distribution (required for D3D12)", ""),
-        ("agility_sdk_path", "Path to the Agility SDK distribution (optional for D3D12)", ""),
+        ("mesa_libs", "Path to the MESA/NIR static libraries (required for D3D12)", f"{os.getenv("LOCALAPPDATA")}/mesa"),
+        ("dxc_path", "Path to the DirectX Shader Compiler distribution (required for D3D12)", f"{os.getenv("LOCALAPPDATA")}/dxc"),
+        ("agility_sdk_path", "Path to the Agility SDK distribution (optional for D3D12)", f"{os.getenv("LOCALAPPDATA")}/agility_sdk"),
         ("agility_sdk_multiarch", "Whether the Agility SDK DLLs will be stored in arch-specific subdirectories", False),
-        ("pix_path", "Path to the PIX runtime distribution (optional for D3D12)", ""),
+        ("pix_path", "Path to the PIX runtime distribution (optional for D3D12)", f"{os.getenv("LOCALAPPDATA")}/pix"),
```
akien-mga, line 195: I would install those in something like `%LOCALAPPDATA%\Godot\build_deps\`. Which can can be factored out as a variable so you only need to `getenv` once.

misc/scripts/install_d3d12_sdk_windows.ps1
```diff
@@ -0,0 +1,82 @@
...
+
+Write-Output "Extracting DirectX 12 Agility SDK $agility_sdk_version ..."
+Expand-Archive "$env:TEMP\$agility_sdk_version" -DestinationPath "$env:LOCALAPPDATA\agility_sdk"
+
+# Remove the original archive as it's not required anymore.
+Remove-Item -Recurse -Force "$env:TEMP\$agility_sdk_version"
+
+Write-Output "DirectX 12 Agility SDK $agility_sdk_version installed successfully.`n"
+
+Write-Output "All Direct3D 12 SDK components were installed successfully!"
```
akien-mga, line 81: I would output where the dependencies have been installed.

misc/scripts/install_d3d12_sdk_windows.ps1
```diff
@@ -0,0 +1,82 @@
+# DirectX Shader Compiler
+
+$dxc_filename="dxc_2023_08_14.zip"
+Write-Output "Downloading DirectX Shader Compiler $dxc_filename ..."
+Invoke-WebRequest -Uri https://github.com/microsoft/DirectXShaderCompiler/releases/download/v1.7.2308/$dxc_filename -OutFile "$env:TEMP\$dxc_filename"
```
akien-mga, line 5: Can we extract all version strings as variables at the top of the script, so it's easy to update?

.gitignore
```diff
@@ -35,6 +35,9 @@ bin
 compile_commands.json
 platform/windows/godot_res.res
 
+# Downloaded by `misc/scripts/install_dxcompiler_windows.ps1`
+platform/windows/dxc/
+
```
akien-mga, line 40: Seems to be from an earlier attempt.

akien-mga (commented):
Seems useful :+1: 

I'm a bit concerned about the impact of the Agility and PIX SDKs which are optional dependencies, but now would likely be installed by most users if we document this as the easiest way to set things up.

What's the impact from those optional deps on production builds?

Another concern is that we're installing silently to `%LOCALAPPDATA%`, so it might not be easy for users to clean up old dependencies in their user folder. At least printing the path at install would help, but maybe it should even be printed when building? Or there could be an uninstall script.

MrBBBaiXue:
Is downloading these SDKs to %localappdata% a must?
I think most people will download it only for building godot, is it possible to store these sdks to the godot build folder?

Calinou:
> Is downloading these SDKs to %localappdata% a must? I think most people will download it only for building godot, is it possible to store these sdks to the godot build folder?

It could be done, but my rationale for downloading them to `%LOCALAPPDATA%` is that if you remove the Godot repository folder [and clone it again](https://xkcd.com/1597/), you won't need to redownload the SDK components. This saves bandwidth and time for returning cloners :slightly_smiling_face: 

This is a user-specific location so you don't need any special permissions to write to it.

MrBBBaiXue:
> It could be done, but my rationale for downloading them to `%LOCALAPPDATA%` is that if you remove the Godot repository folder [and clone it again](https://xkcd.com/1597/), you won't need to redownload the SDK components. This saves bandwidth and time for returning cloners 🙂.

Oh that's what I missed. I agree that we can store these in %APPDATA%, but should give a clear reminder (or providing an uninstall script).

LinuxUserGD:
https://github.com/godotengine/godot-docs/pull/8610
> Currently, we don't support building with Direct3D 12 enabled when using MinGW. Support will be added in the future if possible.

A similar Python script could still be useful for cross-compiling Godot with Direct3D on Linux using Wine/MSVC I think.

bruvzg:
> > Currently, we don't support building with Direct3D 12 enabled when using MinGW. Support will be added in the future if possible.
> 
> A similar Python script could still be useful for cross-compiling Godot with Direct3D on Linux using Wine/MSVC I think.

MinGW support will be added, either via https://github.com/godotengine/godot/pull/86150 or https://github.com/godotengine/godot/pull/83452 (Mesa static libs repo is already updated). So adding support for the Wine/MSVC is excessive (is it event possible with the current MSVC? Even if it is, it's probably complicated and unstable).

YuriSizov:
There is also a downside to using PS1 scripts. If they are not signed, you cannot run them without lowering security levels locally. Doing so is a huge PITA, so unless you have an environment ready for this, it will be very challenging to runs a PS1 script.

So moving to Python is probably worthwhile even now.

Calinou:
- Superseded by https://github.com/godotengine/godot/pull/86717.

PR #86129, opened 2023-12-13
Remember last `color_mode` and `picker_shape` in `ColorPicker`s in the editor

### LINKED ISSUE #86063: No picker mode is recorded after the next switchover
### Tested versions

### System information

Windows 11- Godot v4.2.stable.official [46dc27791] - Vulkan (Forward+)

### Issue description

When I switch picker mode the next time I open it should be the pick mode I selected last time but it's not

https://github.com/godotengine/godot/assets/46812028/522c8b55-14cf-4781-963f-a465a03c9432

### Steps to reproduce

Toggle a picker mode off and on again

N/A

TheSofox:
Fixes #86063 

Now when using `ColorPicker` in the Editor, it will always use the last `color_mode` (RGB/HSV/etc) and `picker_shape` (Rectangle/Wheel/Circle/etc.) that the user chose in ANY `ColorPicker`. (eg. If you chose HSV while in the Picker for a Sprite's modulate colour, then HSV will be chosen if you then open the Picker for the Color of a ColorRect)

This is a Project specific setting. The first time the Project is started, it will use EditorSettings for the default of these two values. Any changes will persist across all Pickers in the project and across closing and opening the project again (values saved to hard drive). It will NOT affect Pickers in exported running projects, this is Editor only.

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -3701,8 +3701,9 @@ void EditorNode::_set_current_scene_nocheck(int p_idx) {
 
 void EditorNode::setup_color_picker(ColorPicker *p_picker) {
 	p_picker->set_editor_settings(EditorSettings::get_singleton());
-	int default_color_mode = EDITOR_GET("interface/inspector/default_color_picker_mode");
-	int picker_shape = EDITOR_GET("interface/inspector/default_color_picker_shape");
+	int default_color_mode = EditorSettings::get_singleton()->get_project_metadata("color_picker", "color_mode", EDITOR_GET("interface/inspector/default_color_picker_mode"));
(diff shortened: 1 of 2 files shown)


YuriSizov:
Note that our preferred commit message format is to start with a verb and then follow up with the body of the change. Please amend your commit message to be similar to what the title of the PR is right now.

TheSofox:
Amended commit message. I'll try to remember to start commit messages with a verb from now on.

KoBeWi (approved):
I see 2 concerns here:
- Users might get confused that changing editor setting has no effect. This can be solved with documentation probably, and I doubt someone will change this more than once anyway
- When opening color picker, the setting is immediately written to the file, because editor calls `set_color_mode()` and `set_picker_shape()`. This could be solved with an intermediate method called when changing the modes manually (i.e. not from code)

The implementation looks good though, so the above remarks can be addressed later.

YuriSizov:
Thanks!

PR #86135, opened 2023-12-13
Implement GI Reflections (SDFGI, VoxelGI) for LightmapGI

guerro323:
*This complete a part of https://github.com/godotengine/godot-proposals/issues/3012 (it doesn't address disabling VoxelGI reflections for SDFGI).*
*Depends on https://github.com/godotengine/godot/pull/86102 for Specular Occlusion.*
##
image
*(greatly exaggerated for showing the reflections, look in screenshots for the normal ones)*
*(scene from https://sketchfab.com/3d-models/japanese-street-at-night-fb1bdcd71a5544d699379d2d13dd1171)*  

This add support for reflections from GI such as SDFGI or VoxelGI onto lightmapped objects.  

### Benefits and constraints
- It help level designers dealing with less reflection probes, especially in complex scenes (such as rounded walls, complex corridors that aren't straight, etc...)
- Reflection probes can still be used and blends well with the GI Reflections and should be used for cube shaped interiors or when you need a mirror-like reflection which can be ruined by SDFGI/VoxelGI voxels look. 
- There is a cost of using `LightmapGI + SDFGI`, but it's less than `SDFGI only` as we only apply the reflection buffer (further optimizations could be possible).
- `VoxelGI + LightmapGI` has a smaller cost than `VoxelGI Only` as we only trace for reflection.
- Without specular occlusion it doesn't work well with simple shapes and no textures if you crank the roughness to 0 and metallic to 100.
- Specular occlusion will help making the GI reflections look more right and occluded when it should be.

##

<details><summary><h3>Screenshots and Comparisons</h3></summary>  

### 'Reflection' scene
Source: https://github.com/Calinou/godot-reflection
|Lightmap 4.2|SDFGI|Lightmap + Specular Occlusion|Lightmap + SO + SDFGI|Lightmap + SO + VoxelGI|
:---:|:---:|:---:|:---:|:---:
image | image | image | image | image

### Cube scene
|Lightmap 4.2|SDFGI|Lightmap + Specular Occlusion|Lightmap + SO + SDFGI|Lightmap + SO + VoxelGI|
:---:|:---:|:---:|:---:|:---:|
image | image | image | image | image

### Japanese Street
Source: https://sketchfab.com/3d-models/japanese-street-at-night-fb1bdcd71a5544d699379d2d13dd1171
|Lightmap 4.2|Lightmap + Specular Occlusion|Lightmap + SO + SDFGI|Lightmap + SO + VoxelGI|
:---:|:---:|:---:|:---:
image | image | image | image

### Simple scene:
- The left part of the floor is a standard default material, the other part is a metallic with roughness set to 0.  
- The black cubes are material with roughness set to 0 and black color.  
- The grey cube is a 100% metallic white material.  
- The white cube is a 100% specular white material.  

|Lightmap Only|Lightmap + SDFGI|
:---:|:---:
image | image |  
image | image

</details>

##

### Caveats:
- You may need to modify the values of SDFGI or VoxelGI as they may be darker when lightmaps are on. A good way to check if everything is right is to have a fully reflective surface and check if the GI reflections match the same brightness/color as your scene.

### Open Questions:
- [ ] Optimizations? It is possible to optimize SDFGI to only care about reflection data? 
- [ ] An option to disable this effect? Would there be an use case to only have GI reflections only on non-lightmapped objects?

___
¹ I posted some more screenshots (without specular occlusion) and benchmarks here https://github.com/godotengine/godot-proposals/issues/3012#issuecomment-1850197439.


clayjohn:
Great work so far!

I've added this to the agenda for the next rendering meeting. This idea has popped up a few times, but I think it will require some discussion and a lot more work. Particularly we need to discuss the workflow issues here and how to optimize out the calculation of diffuse lighting from VoxelGI/SDFGI.

atirut-w:
Testing this out on my project, SDFGI reflections are abnormally dark on lightmapped meshes.

No lightmap on sphere:
image

Lightmap on sphere:
image

This issue is also apparent in a newly created project:
image

atirut-w:
> you can also disable specular occlusion

Apparently, turning it off also darken reflections?

atirut-w:
After playing with this PR for some time, I don't think doing the occlusion based on indirect lighting lightmap. There's the obvious issue with differences between static & dynamic lights, but also scenarios where this won't look accurate.

One of those scenarios would be a dimly lit room with window(s) to the outside:
image

The reflections here are not correct. Here is how it looks with only SDFGI:
image

Maybe we should bake a sort of "occlusion map" instead?

guerro323:
> Apparently, turning it off also darken reflections?
I accidentally included an old commit of the specular occlusion in it which made disabling it buggy.  
It should be fine now.

atirut-w:
> It should be fine now.

Well, about that... disabling specular occlusion now turn reflections black...
image

guerro323:
Would it be possible to have a simple scene that reproduce this? I'm not able to get black reflections :/

atirut-w:
I managed to reproduce a slightly different behavior where everything except the sky is black: [Lightmap.zip](https://github.com/godotengine/godot/files/14072261/Lightmap.zip)

guerro323:
It is fine on my side.  
I'm not sure why it doesn't work for you, maybe it is related to the shader cache? 
image

atirut-w:
Weird. Clearing the shader cache doesn't solve it either.

Here's the system information anyway, so that it may help: `Godot v4.3.dev (1bc749129) - Windows 10.0.22621 - Vulkan (Forward+) - dedicated AMD Radeon RX 6700 XT (Advanced Micro Devices, Inc.; 31.0.23013.1023) - AMD Ryzen 7 5700X 8-Core Processor (16 Threads)`

Calinou (approved):
Tested locally (rebased against `master` https://github.com/godotengine/godot/commit/17e7f85c06366b427e5068c5b3e2940e27ff5f1d), it works as expected.

### LightmapGI only

Screenshot_20240127_204434 webp

### LightmapGI + SDFGI

*Non-reflective materials appear darker because the material isn't fully rough (it's the fallback 3D material).*

Screenshot_20240127_204457 webp

### SDFGI only

Screenshot_20240127_204516 webp

### No GI

Screenshot_20240127_204531 webp

atirut-w:
> It is fine on my side. I'm not sure why it doesn't work for you, maybe it is related to the shader cache?

I narrowed it down to *something* about AMD GPUs. I tested the MRP on my laptop which has Intel Xe graphics and the issure was not occuring there. I think this needs more testing on AMD GPUs like mine (RX 6700 XT) because it's a pretty breaking bug.

One silver lining is that it only appears when specular occlusion is disabled, I guess?

takdew:
Are there any plans for this to get merged or improved? There already is a working implementation of specular occlusion.

PR #75949, opened 2023-04-11
Split editor-specific import metadata for textures

YuriSizov:
The issue right now is that we need to store current values of editor settings for comparison, so we can invalidate the imported texture and reimport it. I used the `*.import` file, where we already store some metadata and for which we have the necessary scaffolding to read and write.

Unfortunately, this was a bad idea, as the values are of course different for different users and systems, and storing them like I did introduces unnecessary changes to the `*.import` file, which should be system-independent, and which is included into version control.

So this PR addresses the issue by splitting these editor-specific metadata properties from the rest and storing them in the `.godot/imported` folder, alongside the texture itself. The file is created and loaded only if you import a texture for editor use. Existing projects will receive an error, urging them to reimport the file. I could've added some fallback logic to read from the old place just in case, but I think it's better to just reimport the texture and be done with it.

Code_2023-04-11_20-24-45

PR #86137, opened 2023-12-14
Do not print error when editor meta was not found as it will be reimported anyway after

### LINKED ISSUE #85859: `Missing required editor-specific import metadata for a texture ...` when checking out a new project/plugin
### Tested versions

Godot 4.2
But the version doesn't matter. 

### System information

Godot v4.2.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated AMD Radeon RX 6900 XT (Advanced Micro Devices, Inc.; 31.0.22023.1014) - AMD Ryzen 9 5900X 12-Core Processor (24 Threads)

### Issue description

When you checkout a new project or plugin (e.g. via Git, AssetLib) that uses icons where the import setting were modified, an error is spamed in the log for every icon.

Example:  
- `icon.svg` with import settings modified
- `Missing required editor-specific import metadata for a texture (please reimport it using the 'Import' tab): 'res://.godot/imported/icon.svg-218a8f2b3041327d8a5756f3a245f83b.editor.meta'`

### Steps to reproduce

1. Modify an icon (e.g. the default `icon.svg`)
image
2. Change an import setting (e.g. `Scale With Editor Scale`)
image
3. Close Godot and delete the `.godot` folder (To simulate a new project, e.g. when someone is checking out the project it will not exist yet)
4. Open Godot, see that there will be an error printed to the console

_Details:_
When changing some setting the `has_editor_variant` flag is set. In this case Godot expects that there must be a file in the `.godot` folder, and prints an error when not.
Content of the `icon.svg.import`
image

_Note:_
The icons will work fine from my testing, e.g. `Scale With Editor Scale` works even though the error was printed. The import file is also created after the error was printed, so this is fine as well.

[Icon.zip](https://github.com/godotengine/godot/files/13589632/Icon.zip)

Maran23:
Fixes: https://github.com/godotengine/godot/issues/85859

The error is printed when the editor metadata files for a particular image were not found,
which can happen when you check out a new project or clean up your `.godot` folder.

But this is not really necessary, as Godot will reimport the file afterwards anyway, 
because it detects that the metadata are not up to date (in this case it does not even exist).

The following is now changed:
- Deleted all 4 files (`ctex, editor.ctex, editor.meta, md5`) (happens for a newly checked out project)
    - Before: Error
    - Now: Will be recreated, no error
- Deleted `editor.meta`
    - Same as before (see also #75949):
        - Error: Missing required editor-specific import metadata for a texture (please reimport it using the 'Import' tab): 'res://.godot/imported/icon.svg-218a8f2b3041327d8a5756f3a245f83b.editor.meta'
- Deleted `editor.ctex`
    - Before: Error
    - After: Will be recreated, no error
- Deleted `ctex`
    - Same as before (recreate)
- Deleted `md5`
    - Same as before (recreate)

cc @YuriSizov I would be interested in your opinion as you have added this error line in https://github.com/godotengine/godot/pull/75949, just in case I miss something here. 🙂

### DIFF
--- a/editor/import/resource_importer_texture.cpp
+++ b/editor/import/resource_importer_texture.cpp
@@ -745,6 +745,10 @@ bool ResourceImporterTexture::are_import_settings_valid(const String &p_path) co
 
 	if (meta.has("has_editor_variant")) {
 		String imported_path = ResourceFormatImporter::get_singleton()->get_internal_resource_path(p_path);
+		if (!FileAccess::exists(imported_path)) {
+			return false;
+		}
+
 		String editor_meta_path = imported_path.replace(".editor.ctex", ".editor.meta");
 		Dictionary editor_meta = _load_editor_meta(editor_meta_path);
 


YuriSizov:
> But this is not really necessary, as Godot will reimport the file afterwards anyway

That's only true because you've removed all the contents of the `.godot` folder. However, if only the metadata file is missing then we won't reimport anything, unless there are other reasons to do so. Check the logic of `ResourceImporterTexture::are_import_settings_valid`. We return an empty dictionary, which means we will never check if import settings are valid because we have nothing to compare against.

The error prints an actionable message for the user in this case. Which was especially important when the change was introduced, as existing projects would have imported textures but not metadata files.

If the message appears under some circumstances which are valid and don't require user action, then the change should be to handle those situations gracefully rather than remove the message completely.

Maran23:
Changed the logic to check if the file exist.
Not sure about the "there is meta but it's empty" case. 
Before and right now, nothings happens if the file exists but there is nothing in it. 
If I got you right, we also want to handle the case where the dictionary is completely empty?
-> If yes, I can change it.

The new logic right now works like this:
- Deleted all 4 files (`ctex, editor.ctex, editor.meta, md5`): Will be created, no error
- Deleted `editor.meta`: Will be created, no error
- Deleted `editor.ctex`: Errors:
  Unable to open file: res://.godot/imported/icon.svg-218a8f2b3041327d8a5756f3a245f83b.editor.ctex.
  Failed loading resource: res://.godot/imported/icon.svg-218a8f2b3041327d8a5756f3a245f83b.editor.ctex. Make sure resources have been imported by opening the project in the editor at least once.
  Failed loading resource: res://icon.svg. Make sure resources have been imported by opening the project in the editor at least once.
- Deleted `ctex`: Will be created, no error
- Deleted `md5`: Will be created, no error

Although now the `Missing required editor-specific import metadata ...` will not be printed anymore as we do check the case of an non-existing file before.

YuriSizov:
You did the exact opposite of what I was describing 🙃

Maran23:
> You did the exact opposite of what I was describing 🙃 The only case where we don't want an error is when `ctex` is missing, because we will regenerate the file anyway. In other cases there should be an error.

I think you are mixing this up. I did not change the `editor.ctex` or `ctex` handling here (also not `md5`). 
And regardless, it looks fine as in this case we really need to reimport the texture to fix the scene(s) using the corresponding texture.

The only thing that is changed is when there is no `editor.meta` file (or even all files are missing, like for new projects).
In this case we can return false and let Godot regenerate the file.

Let us elaborate on this. What should happen for what case?
- All (editor) import files missing
- editor.meta missing
- editor.ctex missing
- ctex missing (not editor specific, so probably does not matter for this PR)
- md5 missing (not editor specific, so probably does not matter for this PR)

YuriSizov:
> I think you are mixing this up. I did not change the `editor.ctex` or `ctex` handling here (also not `md5`).

That's not what I was suggesting, but I did misread what kind of error we show in your bulletpoint list above. Point still stands, though, as I suggested we should show the error when the meta file is missing and we should reimport the icon without any issues if the ctex file is missing.

If non-editor ctex and editor ctex are handled differently by the engine right now, and the latter errs where the former doesn't, it's a problem that needs to be addressed (but not necessarily right here and right now).

----

The reason why we err when the meta file is missing is, as I explained, compatibility. Existing projects that predate the meta file can already have correctly imported textures which would work just fine without a reimport. The meta file is only needed to check for a reimport, it's not needed for the runtime. But we still want to tell users that they need to reimport affected textures to update them to the latest format of the metadata.

If we decide that we don't care about this anymore, then sure, we can go with the way this PR is right now and just treat missing meta files as a reason to force a reimport.

However, if we want to preserve this logic and address your issue at the same time, then we should check if the `editor.ctex` file is missing and if it is, avoid throwing the message because the reimport will happen regardless. I'm fine with whatever option, but I hope you understand the logic better now.

Maran23:
> However, if we want to preserve this logic and address your issue at the same time, then we should check if the `editor.ctex` file is missing and if it is, avoid throwing the message because the reimport will happen regardless. I'm fine with whatever option, but I hope you understand the logic better now.

Ahh, yes, I got you now. Thanks for explaining the details here. I did not fully understand the 'dependency' between the `editor.ctex` and `editor.meta` file I'm afraid.

Behaviour now:
- Deleted all 4 files (`ctex, editor.ctex, editor.meta, md5`): Will be recreated, no error
- Deleted `editor.meta`: 
Error: Missing required editor-specific import metadata for a texture (please reimport it using the 'Import' tab): 'res://.godot/imported/icon.svg-218a8f2b3041327d8a5756f3a245f83b.editor.meta'
- Deleted `editor.ctex`: Will be recreated, no error
- Deleted `ctex`: Will be recreated, no error
- Deleted `md5`: Will be recreated, no error

Looks much better. :)

YuriSizov (approved):

YuriSizov:
Thanks!

PR #86141, opened 2023-12-14
Fix Popup hover and height calculation are off by some pixels

### LINKED ISSUE #86140: `Popup` hover and height calculation are off by some pixels
### Tested versions

Godot 4.1.3
Godot 4.2.1
master (https://github.com/godotengine/godot/commit/aa5b6ed13e4644633baf2a8a1384c82e91c533a1)

### System information

Godot v4.2.1.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated AMD Radeon RX 6900 XT (Advanced Micro Devices, Inc.; 31.0.23013.1023) - AMD Ryzen 9 5900X 12-Core Processor (24 Threads)

### Issue description

In Godot, `Popup`s height and mouse hover calculation are off by some pixels.
- The wrong height calculation results in the last item is cutoff for big `Popup`s (with a `ScrollBar`).
- The wrong mouse hover calculation results in the very first item receives the hover effect when the mouse cursor is already some pixels at/in the item

### Steps to reproduce

Open a `Popup`, e.g. `Texture` on a `TextureRect`, or just `@export` a `Resource` to get a very big `Popup`.

https://github.com/godotengine/godot/assets/66004280/af397de2-3cbb-40d6-94bc-0c7c278340d6

Probably not needed, as it can be reproduced easily in the editor, but still I attached one below.
For the big `Resource` `Popup`, just create a script with the following line:
`@export var abc: Resource`

This is also included in the test project:
[Icon+.zip](https://github.com/godotengine/godot/files/13667181/Icon%2B.zip)

Maran23:
Fixes: https://github.com/godotengine/godot/issues/86140

This was caused by an incorrect calculation of the `v_separation` where it was not needed.

To better understand the issue, this is how the `v_separation` is used:
image
Every item has a top separation of 5, and a bottom separation of 5 (`v_separation` = 10).
Therefore the following statement is true: 1x item has 1x `v_separation`, which is also the fix for both calculation issues.

### DIFF
--- a/scene/gui/popup_menu.cpp
+++ b/scene/gui/popup_menu.cpp
@@ -237,8 +237,7 @@ int PopupMenu::_get_items_total_height() const {
 		items_total_height += _get_item_height(i) + theme_cache.v_separation;
 	}
 
-	// Subtract a separator which is not needed for the last item.
-	return items_total_height - theme_cache.v_separation;
+	return items_total_height;
 }
 
 int PopupMenu::_get_mouse_over(const Point2 &p_over) const {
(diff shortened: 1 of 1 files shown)


KoBeWi (approved):
Well I can confirm that PopupMenu is quite glitchy before this fix (e.g. try shrinking PopupMenu with 2 items and 0 separation. It will show the scrollbar...). Though the old code does not look obviously wrong to me, so I can't say how correct is the fix. It does not seem to break anything at least.

YuriSizov (approved):
Makes sense. It looks like modified code treated separation as an external element, something added in between items. But we clearly use the separation as a part of each item. This may be inconsistent with other GUI elements, but that's another topic. For fixing the bugs this all makes sense and does work.

akien-mga:
Thanks!

PR #86143, opened 2023-12-14
Stop using `RTR()` inside the "editor" folder

YeldhamDev:
~`RTR()` is supposed to be only used in strings that are translatable in user projects, so strings that didn't follow this rule now use `TTR()` instead.~

~This PR is necessary for my work to add built-in strings in the POT generation.~

Alright, unless we want to have loads of `#ifdef`s, this is not a good idea. The PR has now been repurposed to just switch cases of `RTR()` to `TTR()` inside the "editor" folder.

### DIFF
--- a/editor/plugins/font_config_plugin.cpp
+++ b/editor/plugins/font_config_plugin.cpp
@@ -724,7 +724,7 @@ void EditorPropertyOTFeatures::update_property() {
 		}
 		for (int i = 0; i < FGRP_MAX; i++) {
 			if (have_sub[i]) {
-				menu->add_submenu_node_item(RTR(group_names[i]), menu_sub[i]);
+				menu->add_submenu_node_item(TTRGET(group_names[i]), menu_sub[i]);
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


core/input/input_event.cpp
```diff
@@ -444,9 +444,9 @@ String InputEventKey::as_text() const {
 	} else if (keycode != Key::NONE) {
 		kc = keycode_get_string(keycode);
 	} else if (physical_keycode != Key::NONE) {
-		kc = keycode_get_string(physical_keycode) + " (" + RTR("Physical") + ")";
+		kc = keycode_get_string(physical_keycode) + " (" + TTR("Physical") + ")";
```
timothyqiu, line 447: These public API changes are technically compat-breaking and needs to be documented.

YeldhamDev: Hmm, I think those should stay as `RTR()`, actually.

editor/plugins/font_config_plugin.cpp
```diff
@@ -723,7 +723,7 @@ void EditorPropertyOTFeatures::update_property() {
 		}
 		for (int i = 0; i < FGRP_MAX; i++) {
 			if (have_sub[i]) {
-				menu->add_submenu_item(RTR(group_names[i]), "FTRMenu_" + itos(i));
+				menu->add_submenu_item(TTR(group_names[i]), "FTRMenu_" + itos(i));
```
bruvzg, line 726: This one probably should be:
```suggestion
				menu->add_submenu_item(TTRGET(group_names[i]), "FTRMenu_" + itos(i));
```

And have actual names wrapped with `TTRC`:

```diff
diff --git a/editor/plugins/font_config_plugin.cpp b/editor/plugins/font_config_plugin.cpp
index a0500bdb48..9873236422 100644
--- a/editor/plugins/font_config_plugin.cpp
+++ b/editor/plugins/font_config_plugin.cpp
@@ -856,15 +856,15 @@ EditorPropertyOTFeatures::EditorPropertyOTFeatures() {
                menu_sub[i]->connect("id_pressed", callable_mp(this, &EditorPropertyOTFeatures::_add_feature));
        }

-       group_names[FGRP_STYLISTIC_SET] = "Stylistic Sets";
-       group_names[FGRP_CHARACTER_VARIANT] = "Character Variants";
-       group_names[FGRP_CAPITLS] = "Capitals";
-       group_names[FGRP_LIGATURES] = "Ligatures";
-       group_names[FGRP_ALTERNATES] = "Alternates";
-       group_names[FGRP_EAL] = "East Asian Language";
-       group_names[FGRP_EAW] = "East Asian Widths";
-       group_names[FGRP_NUMAL] = "Numeral Alignment";
-       group_names[FGRP_CUSTOM] = "Custom";
+       group_names[FGRP_STYLISTIC_SET] = TTRC("Stylistic Sets");
+       group_names[FGRP_CHARACTER_VARIANT] = TTRC("Character Variants");
+       group_names[FGRP_CAPITLS] = TTRC("Capitals");
+       group_names[FGRP_LIGATURES] = TTRC("Ligatures");
+       group_names[FGRP_ALTERNATES] = TTRC("Alternates");
+       group_names[FGRP_EAL] = TTRC("East Asian Language");
+       group_names[FGRP_EAW] = TTRC("East Asian Widths");
+       group_names[FGRP_NUMAL] = TTRC("Numeral Alignment");
+       group_names[FGRP_CUSTOM] = TTRC("Custom");
 }
```

YeldhamDev: Done.

akien-mga (approved):

PR #86144, opened 2023-12-14
Disable mesh compression if vertex `position.z` is always 0

### LINKED ISSUE #85963: Importing a OBJ mesh into a MeshInstance2D completely messes its transforms
### Tested versions

Reproductible in v4.2.stable.mono.official [46dc27791]
Not Reproductible in v4.1.3.stable.official [f06b6836a]

### System information

Godot v4.2.stable.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1080 (NVIDIA; 30.0.15.1165) - Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz (8 Threads)

### Issue description

I'm doing a 2D game in a very polygonal/vector style, and I was planning on using 2D Meshes.
However, upon re-importing my meshes in 4.2, I noticed that all of them were incorrectly scaled in the 2D Viewport, despite their actual (Node2D) scale not changing.

## The situation in Godot `v4.2.stable.mono.official [46dc27791]`
What I expected the mesh to look like (screenshot from Blender, the mesh is built on the XZ axis, and exported to use Y as up, so it should be on XY in Godot)
image

The mesh in Godot's 2D viewport, it was clearly offset to the bottom left, and squished on one axis. Its visual scale is also immensely smaller than the export scale.
image

The MeshInstance's full hitbox (the model isn't even visible anymore)
image

## Double-checking
Now, to Check that I wasn't just going Insane or doing something wrong, I tried to create a completely new project AND a completely blank blend file with this mesh that I quickly cobbled together.
image

However, when I imported it to Godot, I was met with this, and to double check that it was also doing what I described above...
image
... I superposed the blender viewport and Godot viewport, and this mesh seems to also be squished, with no intervention on my part 
image

## Checking in Godot 4.1
To be fully sure, I imported the same OBJ meshes in Godot `v4.1.3.stable.official [f06b6836a]`.
Which seems to work as expected
image
image

### Steps to reproduce

- Create a flat mesh in Blender (or any software of your choice)
- Export that mesh as a OBJ, so that it should be aligned with the XY plane
- Import it in a fresh Godot project
- Create a 2D scene, and add a MeshInstance2D to it
- Assign your mesh to the MeshInstance2D's mesh field
- Verify that its Hitbox matches its bounds, and that it wasn't deformed or moved around by the importer

4.2 and 4.1 Projects used in this report linked below
[godot_4.1_bug_hunt.zip](https://github.com/godotengine/godot/files/13624429/godot_4.1_bug_hunt.zip)
[godot_4.2_bug_hunt.zip](https://github.com/godotengine/godot/files/13624431/godot_4.2_bug_hunt.zip)

jsjtxietian:
Fixes #85963

As https://github.com/godotengine/godot/issues/85963#issuecomment-1848615844 mentioned, "We could detect if the z channel is used in the vertex positions, if z is always 0, we could assume the mesh is 2D and disable compression" , this pr is a brute-force way of implementing it. I'm not sure if it's the right place to insert this check, but it seems the suitable place. 

Feel free to correct me!

### DIFF
--- a/editor/import/3d/editor_import_collada.cpp
+++ b/editor/import/3d/editor_import_collada.cpp
@@ -996,7 +996,16 @@ Error ColladaImport::_create_mesh_surfaces(bool p_optimize, Ref<ImporterMesh> &p
 				surftool->generate_tangents();
 			}
 
-			if (p_mesh->get_blend_shape_count() != 0 || p_skin_controller) {
+			// Disable compression if all z equals 0 (the mesh is 2D).
+			bool is_mesh_2d = true;
+			for (int k = 0; k < vertex_array.size(); k++) {
+				if (!Math::is_zero_approx(vertex_array[k].vertex.z)) {
+					is_mesh_2d = false;
+					break;
+				}
+			};
+
+			if (p_mesh->get_blend_shape_count() != 0 || p_skin_controller || is_mesh_2d) {
 				// Can't compress if attributes missing or if using vertex weights.
 				mesh_flags &= ~RS::ARRAY_FLAG_COMPRESS_ATTRIBUTES;
 			}
--- a/editor/import/3d/resource_importer_obj.cpp
+++ b/editor/import/3d/resource_importer_obj.cpp
@@ -384,7 +384,22 @@ static Error _parse_obj(const String &p_path, List<Ref<ImporterMesh>> &r_meshes,
 
 			if (p_disable_compression) {
 				mesh_flags = 0;
+			} else {
+				bool is_mesh_2d = true;
+
+				// Disable compression if all z equals 0 (the mesh is 2D).
+				for (int i = 0; i < vertices.size(); i++) {
+					if (!Math::is_zero_approx(vertices[i].z)) {
+						is_mesh_2d = false;
+						break;
+					}
+				}
+
+				if (is_mesh_2d) {
+					mesh_flags = 0;
+				}
 			}
+
 			//groups are too annoying
 			if (surf_tool->get_vertex_array().size()) {
 				//another group going on, commit it
--- a/modules/gltf/gltf_document.cpp
+++ b/modules/gltf/gltf_document.cpp
@@ -3059,7 +3059,17 @@ Error GLTFDocument::_parse_meshes(Ref<GLTFState> p_state) {
 				array[Mesh::ARRAY_TANGENT] = tangents;
 			}
 
-			if (p_state->force_disable_compression || !a.has("POSITION") || !a.has("NORMAL") || p.has("targets") || (a.has("JOINTS_0") || a.has("JOINTS_1"))) {
+			// Disable compression if all z equals 0 (the mesh is 2D).
+			const Vector<Vector3> &vertices = array[Mesh::ARRAY_VERTEX];
+			bool is_mesh_2d = true;
+			for (int k = 0; k < vertices.size(); k++) {
+				if (!Math::is_zero_approx(vertices[k].z)) {
+					is_mesh_2d = false;
+					break;
+				}
+			}
+
+			if (p_state->force_disable_compression || is_mesh_2d || !a.has("POSITION") || !a.has("NORMAL") || p.has("targets") || (a.has("JOINTS_0") || a.has("JOINTS_1"))) {
 				flags &= ~RS::ARRAY_FLAG_COMPRESS_ATTRIBUTES;
 			}
 


scene/resources/surface_tool.cpp
```diff
@@ -729,6 +729,20 @@ Ref<ArrayMesh> SurfaceTool::commit(const Ref<ArrayMesh> &p_existing, uint64_t p_
 
 	Array a = commit_to_arrays();
 
+	// Disable compression if all z euqals 0 (the mesh is 2D)
+	Vector<Vector3> mesh_vertices = a[RS::ARRAY_VERTEX];
+	bool is_mesh_2d = true;
+
+	for (int i = 0; i < mesh_vertices.size(); i++) {
+		if (!Math::is_zero_approx(mesh_vertices[i].z)) {
+			is_mesh_2d = false;
```
RedMser, line 738:
```suggestion
			is_mesh_2d = false;
			break;
```
Early cancel out of loop.

fire:
Is it true that a lot of the code assumes mesh compression is on? I am not that familiar.

scene/resources/surface_tool.cpp
jsjtxietian: Thanks!

scene/resources/surface_tool.cpp
```diff
@@ -729,6 +729,21 @@ Ref<ArrayMesh> SurfaceTool::commit(const Ref<ArrayMesh> &p_existing, uint64_t p_
...
+			break;
+		}
+	}
+
+	if (is_mesh_2d) {
+		p_compress_flags = 0;
```
lyuma, line 744: Should this code path set `ARRAY_FLAG_USE_2D_VERTICES` ?
Also, what if there are 3D normals or tangents whose z value is nonzero? There are cases in 3D where a mesh will have zero z, for example exporting a plane from Blender for use as a ground surface.

jsjtxietian: > There are cases in 3D where a mesh will have zero z

Yes I do agree there should be more discussions about how to check whether the mesh is 2D or not, this pr is just a hijack.  But on the other hand, IMHO disable mesh compression for the case you mentioned is not a severe problem.

scene/resources/surface_tool.cpp
```diff
@@ -729,6 +729,21 @@ Ref<ArrayMesh> SurfaceTool::commit(const Ref<ArrayMesh> &p_existing, uint64_t p_
 
 	Array a = commit_to_arrays();
 
+	// Disable compression if all z euqals 0 (the mesh is 2D)
```
akien-mga, line 732:
```suggestion
	// Disable compression if all z equals 0 (the mesh is 2D).
```

jsjtxietian: Thanks!

clayjohn (commented):
The code looks good, but it should be in the respective importers, not in the SurfaceTool. 

https://github.com/godotengine/godot/blob/2e7fc81315bfa8d0a15f60adff2a12b6f3104236/editor/import/3d/resource_importer_obj.cpp#L386-L387

https://github.com/godotengine/godot/blob/16d61427cab3a8e43f0a9a8ee724fc176b6433c6/modules/gltf/gltf_document.cpp#L2825-L2827

https://github.com/godotengine/godot/blob/16d61427cab3a8e43f0a9a8ee724fc176b6433c6/editor/import/3d/editor_import_collada.cpp#L999-L1002

editor/import/3d/editor_import_collada.cpp
```diff
@@ -996,7 +996,16 @@ Error ColladaImport::_create_mesh_surfaces(bool p_optimize, Ref<ImporterMesh> &p
 				surftool->generate_tangents();
 			}
 
-			if (p_mesh->get_blend_shape_count() != 0 || p_skin_controller) {
+			// Disable compression if all z euqals 0 (the mesh is 2D).
```
clayjohn, line 999:
```suggestion
			// Disable compression if all z equals 0 (the mesh is 2D).
```

clayjohn (commented):
Looks great! There is just a small typo in your comment. Once that is fixed, we can merge this for 4.3!

editor/import/3d/editor_import_collada.cpp
jsjtxietian: Thanks!

clayjohn (approved):
Looks great! Thank you

Issue #80218, opened 2023-08-03
Script Create Dialog is enormous


KoBeWi:
### Godot version

4.2 0606ba7

### System information

Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1060 (NVIDIA; 30.0.15.1403) - Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz (8 Threads)

### Issue description

When you create a new script, the dialog has a height of a few screens. Likely caused by autowrapped Labels. The bug happens even when a new Label appears in already shrunk dialog:

https://github.com/godotengine/godot/assets/2223172/f25fb153-9e24-4aaa-beef-6e7c8554cb39

### Steps to reproduce

1. Create script

### Minimal reproduction project

N/A

bruvzg:
> Seems to be a regression from https://github.com/godotengine/godot/pull/77280 according to @bruvzg, CC @Rindbee.

The #77280 seems to be fine by itself. The issue is most likely in the container control fitting code and the way how it's handling min. size (assuming it's constant, while it's never true, and min. height always depends on width or vice versa).

bruvzg:
So the root of the issue: container is calculating its minimum size and sorting controls BEFORE setting control size, and using `get_combined_minimum_size` of the child controls for this purpose. At this point, child controls have zero size, and the calculated minimum size of the child controls will have huge height (since text is broken into a single character per line to fit control zero width). But I'm not sure what would be the correct order of operation here.

https://github.com/godotengine/godot/blob/725beaa7afce294f38d4aca1fa287249021b6bf8/scene/gui/box_container.cpp#L266

Rindbee:
Seems to be the same case I'm worried about in https://github.com/godotengine/godot/issues/74052#issuecomment-1587372009.

Rindbee:
This issue is due to the window with `wrap_controls` enabled, which is similar to a container, but it does not correct itself, because its size can only be automatically expanded, not automatically shrunk.

https://github.com/godotengine/godot/blob/16a93563bfd3b02ca0a8f6df2026f3a3217f5571/scene/gui/control.cpp#L1610-L1614

The code removed in #77280 seems to prevent scaling up too quickly. But it causes #77268.

[MRP.zip](https://github.com/godotengine/godot/files/12266886/MRP.zip)

YuriSizov:
[...]

PR #77280, opened 2023-05-20
Fix unnecessary break when calculating the height of visible lines

Rindbee:
This break causes the minsize to be smaller than expected, and then the size keeps increasing by one line to cover all visible lines. This can cause performance issues when there are many visible lines.

Fix  #77268.

PR #86145, opened 2023-12-14
Set an appropriate minimum size for labels in windows that display incorrectly

### LINKED ISSUE #85551: [4.2] Export Window Sized outside of screen, cannot be resized when export template is selected
### Godot version

### System information

Godot v4.2.stable - Windows 10.0.19045 - GLES3 (Compatibility) - NVIDIA GeForce RTX 3080 (NVIDIA; 31.0.15.4617) - AMD Ryzen 5 5600X 6-Core Processor (12 Threads)

### Issue description

{73902E40-48C8-4573-A9D8-F4075EDDB875}
{4337DBCF-2184-445A-9A41-D82DE05FB55E}
{D6BE09FD-C644-49CB-894F-71F550415ABD}

### Steps to reproduce

Just converted my 4.1 project to 4.2, wanted to export immediately as the given project is just a gateway for my server.

n/a

### LINKED ISSUE #86136: Export dialog becomes extremely tall if using dedicated server export mode
### Tested versions

- Reproducible in: 4.2.stable, 4.2.1.stable
- Not reproducible in: 4.1.3.stable

### System information

Godot v4.3.dev (aa5b6ed13) - Fedora Linux 39 (KDE Plasma) - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4090 (nvidia; 545.29.06) - 13th Gen Intel(R) Core(TM) i9-13900K (32 Threads)

### Issue description

When you open the Export dialog, it initially looks correct:

Screenshot_20231214_004727

As soon as you select an export preset that uses dedicated server export mode (configured on the **Resources** tab), the window becomes extremely tall:

Screenshot_20231214_004738

This makes it impossible to reach the buttons at the bottom of the Export dialog, even on a large display.

This is presumably due to the Tree node used within the Resources tab. I bisected the regression to e5bebbc9ffda8b4fa5a786842ea24f0a3ea0763a. cc @Rindbee 

### Steps to reproduce

- Create a project with many folders and files.
- Open the Export dialog and create an export preset.
- Switch to the Resources tab and set the export mode to **Export as dedicated server**.
- Notice the window becoming extremely tall.

[test_dedicated_server_export_window_height.zip](https://github.com/godotengine/godot/files/13666766/test_dedicated_server_export_window_height.zip)

Rindbee:
When the label's `autowrap_mode` is `AUTOWRAP_WORD_SMART` and the initial `text` is set at the same time, it may have a higher height than expected. See https://github.com/godotengine/godot/issues/80218#issuecomment-1669527304 for more details.

Set an appropriate minimum size for labels in windows that display incorrectly so that these windows display properly.

As a workaround for display issues in the editor caused by #77280.

Close #86136.
Closes https://github.com/godotengine/godot/issues/85551.

### DIFF
--- a/editor/export/project_export.cpp
+++ b/editor/export/project_export.cpp
@@ -1277,6 +1277,7 @@ ProjectExportDialog::ProjectExportDialog() {
 	server_strip_message = memnew(Label);
 	server_strip_message->set_visible(false);
 	server_strip_message->set_autowrap_mode(TextServer::AUTOWRAP_WORD_SMART);
+	server_strip_message->set_custom_minimum_size(Size2(300 * EDSCALE, 1));
 	resources_vb->add_child(server_strip_message);
 
 	{
--- a/editor/import/dynamic_font_import_settings.cpp
+++ b/editor/import/dynamic_font_import_settings.cpp
@@ -1345,6 +1345,7 @@ DynamicFontImportSettingsDialog::DynamicFontImportSettingsDialog() {
 	page2_description->set_text(TTR("Add font size, and variation coordinates, and select glyphs to pre-render:"));
 	page2_description->set_h_size_flags(Control::SIZE_EXPAND_FILL);
 	page2_description->set_autowrap_mode(TextServer::AUTOWRAP_WORD_SMART);
+	page2_description->set_custom_minimum_size(Size2(300 * EDSCALE, 1));
 	page2_vb->add_child(page2_description);
 
 	HSplitContainer *page2_hb = memnew(HSplitContainer);
@@ -1418,6 +1419,7 @@ DynamicFontImportSettingsDialog::DynamicFontImportSettingsDialog() {
 	page2_0_description->set_text(TTR("Select translations to add all required glyphs to pre-render list:"));
 	page2_0_description->set_h_size_flags(Control::SIZE_EXPAND_FILL);
 	page2_0_description->set_autowrap_mode(TextServer::AUTOWRAP_WORD_SMART);
+	page2_0_description->set_custom_minimum_size(Size2(300 * EDSCALE, 1));
 	page2_0_vb->add_child(page2_0_description);
 
 	locale_tree = memnew(Tree);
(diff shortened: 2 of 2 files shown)


YuriSizov (dismissed):

editor/export/project_export.cpp
```diff
@@ -1276,6 +1276,7 @@ ProjectExportDialog::ProjectExportDialog() {
 	server_strip_message = memnew(Label);
 	server_strip_message->set_visible(false);
 	server_strip_message->set_autowrap_mode(TextServer::AUTOWRAP_WORD_SMART);
+	server_strip_message->set_size(Size2(500, 1) * EDSCALE);
```
YuriSizov, line 1279: We should use `set_custom_minimum_size` and force the minimum size, not set the default size, this is unreliable.

editor/import/dynamic_font_import_settings.cpp
```diff
@@ -1345,6 +1345,7 @@ DynamicFontImportSettingsDialog::DynamicFontImportSettingsDialog() {
 	page2_description->set_text(TTR("Add font size, and variation coordinates, and select glyphs to pre-render:"));
 	page2_description->set_h_size_flags(Control::SIZE_EXPAND_FILL);
 	page2_description->set_autowrap_mode(TextServer::AUTOWRAP_WORD_SMART);
+	page2_description->set_size(Size2(500, 1) * EDSCALE);
```
YuriSizov, line 1348: I don't see any fixed size for this dialog, so it's worth to double-check if 500 is not too much, This value should be rather conservative (applies to all 3 labels).

YuriSizov (changes requested):
See above.

editor/export/project_export.cpp
Rindbee: One characteristic of this series of issues is that during first `Label::_shape()`, the size assigned to the label by the container has not been updated, causing the label's initial width to be 0, so a relatively large minimum height is calculated. 

https://github.com/godotengine/godot/blob/26b1fd0d842fa3c2f090ead47e8ea7cd2d6515e1/scene/gui/label.cpp#L104-L106

If the initial width is large enough, this won't happen. The minimum size is calculated as a single line height.

YuriSizov: Enforcing a minimum size is still more reliable because it doesn't depend on the order of events. You just need to pick a size which will be within the dialog's own minimum size, so there is no visible difference for the end user.

YuriSizov (approved):

editor/export/project_export.cpp
```diff
@@ -1277,6 +1277,7 @@ ProjectExportDialog::ProjectExportDialog() {
 	server_strip_message = memnew(Label);
 	server_strip_message->set_visible(false);
 	server_strip_message->set_autowrap_mode(TextServer::AUTOWRAP_WORD_SMART);
+	server_strip_message->set_custom_minimum_size(Size2(300, 1) * EDSCALE);
```
AThousandShips, line 1280:
```suggestion
	server_strip_message->set_custom_minimum_size(Size2(300 * EDSCALE, 1));
```
Same for all, IMO it doesn't make sense to scale `1`, and might cause weirdness

AThousandShips (approved):

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

PR #86150, opened 2023-12-14
[D3D12] Add support for building with MinGW.

bruvzg:
Some `godot-nir-static` changes required, see https://github.com/godotengine/godot-nir-static/pull/2 (for MinGW only, MSVC build should work without it).

Tested building and running on:
   - x86_64 with MinGW/GCC (cross build from macOS and native MSYS2 build).
   - x86_64 with MinGW/LLVM (cross build from macOS).

Compiling docs update (https://github.com/godotengine/godot-docs/pull/8624)

Note: PIX do not have MinGW import libs, but it's possible to generate it using following commands (from PIX runtime folder):
```
gendef ./bin/x64/WinPixEventRuntime.dll
dlltool --machine i386:x86-64 --no-leading-underscore -d WinPixEventRuntime.def -D WinPixEventRuntime.dll -l ./bin/x64/libWinPixEventRuntime.a
```


bruvzg:
Superseded by https://github.com/godotengine/godot/pull/83452

PR #83508, opened 2023-10-17
GraphEdit connection rework (API & optimizations)

Geometror:
GE Connections

**Detailed changes:**
- Add `get_closest_connection_at_point` which return the closest connection at a given point in GraphEdit's local screen space (with the given maximum search distance). See the docs for a usage example.
- Optimize connection drawing/handling
	- Introduce connections HashMap (node -> its connections)
		- The old logic iterated over all connections for every GraphNode, now only the relevant connections are checked where possible.
	 - Each connection now caches the line, the start and end positions (in screen space) as well as an AABB. This allows for updating/drawing only those connections visible on the screen (this had a huge performance impact for large graphs). It's also used to accelerate fetching the connection from a screen position which is introduced in this PR.
- Adjust theming/visual appearance
	- hovering a connection now highlights it (theme item `connection_hover_tint_color`)
	- the highlight color when dragging a connection onto a valid port is now a theme item (theme item `connection_valid_target_tint_color`)
- Change signature of `get_connection_list` (now returns the list instead of using a return parameter) [only used internally]

For a demonstration of these new features see #83510.

PR #86158, opened 2023-12-14
Rework `GraphEdit` connections (drawing, API, optimizations)

### LINKED ISSUE #36716: Deleting GraphNodes when handling `delete_nodes_request` causes connection visuals to leak
Godot 3.2
Windows 10 64 bits

I followed the way Godot's graph editors implement the `delete_nodes_request`:

```gdscript
func _on_GraphEdit_delete_nodes_request():
	var to_delete = []
	
	for i in _graph_edit.get_child_count():
		var node = _graph_edit.get_child(i)
		if node is GraphNode:
			if node.selected:
				to_delete.append(node)
	
	for node in to_delete:
		node.free() # queue_free also works
```

It properly removes the node, however it causes connections to remain visible, until the graph gets panned around or redrawn (which will then make them disappear):

image

When redrawing, an error is logged once:
```
ERROR: Node not found: B.
   At: scene/main/node.cpp:1381
```

Note: I think this never happens in Godot's graph-based editors, because those actually rebuild the entire graph each time a node is deleted... which is very far from the way I implement mine.
This happens even if I use `queue_free()`.

Test project: 
[GraphEditDeleteRequest.zip](https://github.com/godotengine/godot/files/4272569/GraphEditDeleteRequest.zip)

If it's required to remove those connections manually, it would be really nice to have a function to do so.
Edit: indeed, it must be done manually.
```cpp
void remove_connections_from_and_to(GraphEdit *graph_edit, StringName node_name) {
	// Get copy of connection list
	List<GraphEdit::Connection> connections;
	graph_edit->get_connection_list(&connections);

	for (List<GraphEdit::Connection>::Element *E = connections.front(); E; E = E->next()) {
		const GraphEdit::Connection &con = E->get();
		if (con.from == node_name || con.to == node_name) {
			graph_edit->disconnect_node(con.from, con.from_port, con.to, con.to_port);
		}
	}
}
```

Geometror:
Supersedes #83508.

Peek 2023-12-14 15-45

# Detailed changes
## API
- Add method `get_closest_connection_at_point` which return the closest connection at a given point in GraphEdit's local screen space (with the given maximum search distance). See the docs for a usage example.
- Add method `get_connections_intersecting_with_rect` 
- Add method `reset_all_connection_activity`

## Visual
- Adjust theming/visual appearance of connection lines
	- Add a tiny rim so that they can be distinguished better when they cross/overlay each other (stripped from #85017)
		- For easy adjustments a `connection_rim_color` theme property is added
	- Hovering a connection now highlights it (theme item `connection_hover_tint_color`)
	- The highlight color when dragging a connection onto a valid port is now a theme item (theme item `connection_valid_target_tint_color`)
- Dragging connection line improvements
	- Show the correct color gradient when hovering over a valid port (it was just a solid color before)
dragged_connection_demo
	- Fix instability while zooming 
	When zooming/panning fast this could happen before: 
Pasted image 20231213162506

## Optimizations, Fixes & Refactoring
- Optimize connection drawing/handling
	- Introduce connections HashMap (node -> its connections)
		- The old logic iterated over all connections for every GraphNode, now only the relevant connections are checked where possible.
	 - Each connection now caches the line, the start and end positions (in screen space) as well as an AABB. This allows for updating/drawing only those connections visible on the screen (this had a huge performance impact for large graphs). It's also used to accelerate fetching the connection from a screen position/rect intersection which is introduced in this PR.
	 - Further improved performance (especially when panning and moving nodes)
- Connection lines are now drawn using Line2D nodes instead of using `draw_polyline` calls in only one `CanvasItem` 
	- The performance concerns I had turned out to be unfounded, this method even allows for some nice optimizations
	- The reason for this was flexibility: e.g. drawing e.g. type-dependent connection line styles (double/triple/quadruple lines, dashed) with a shader requires passing per-line uniforms to the shaders
- Connection line shader
	- Connection lines use a dedicated canvas item shader now
	- The default shader can be overridden to allow for application specific shaders (e.g. to implement connection type dependent styles)
		- The method is not yet exposed
	- Parameters/uniforms currently set by GraphEdit:
		- int from_type
		- int to_type
		- vec4 rim_color
		- float line_width
	- This allows for type-dependent connection line styles, like you can see here:
	*Keep in mind that the following is just a quick/arbitrary POC using the shader override and not included in this PR*
	Clipboard - 8  Dezember 2023 21 06

- Drawing the lines for the minimap is now completely separate, drawing the lines via a shader is not sensible here (this also paves the way for loosening up the coupling between the minimap and GraphEdit itself and allows for optimizations since the minimap is currently a bottleneck)
- Change signature of `get_connection_list` (now returns the list instead of using a return parameter) [only used internally]

- Fixes #36716.

# TODO
- [ ] Decide what to do with the anti aliased property since AA is basically free now
- [ ] Expose the connection line shader override (maybe in a followup PR?)

### DIFF
--- a/core/math/geometry_2d.h
+++ b/core/math/geometry_2d.h
@@ -119,6 +119,10 @@ public:
 		}
 	}
 
+	static real_t get_distance_to_segment(const Vector2 &p_point, const Vector2 *p_segment) {
+		return p_point.distance_to(get_closest_point_to_segment(p_point, p_segment));
+	}
+
 	static bool is_point_in_triangle(const Vector2 &s, const Vector2 &a, const Vector2 &b, const Vector2 &c) {
 		Vector2 an = a - s;
 		Vector2 bn = b - s;
@@ -247,6 +251,28 @@ public:
 			return res2;
 		}
 		return -1;
+	}
+
+	static bool segment_intersects_rect(const Vector2 &p_from, const Vector2 &p_to, const Rect2 &p_rect) {
+		if (p_rect.has_point(p_from) || p_rect.has_point(p_to)) {
+			return true;
+		}
+
+		const Vector2 rect_points[4] = {
+			p_rect.position,
+			p_rect.position + Vector2(p_rect.size.x, 0),
+			p_rect.position + p_rect.size,
+			p_rect.position + Vector2(0, p_rect.size.y)
+		};
+
+		// Check if any of the rect's edges intersect the segment.
+		for (int i = 0; i < 4; i++) {
+			if (segment_intersects_segment(p_from, p_to, rect_points[i], rect_points[(i + 1) % 4], nullptr)) {
+				return true;
+			}
+		}
+
+		return false;
 	}
 
 	enum PolyBooleanOperation {
--- a/doc/classes/GraphEdit.xml
+++ b/doc/classes/GraphEdit.xml
@@ -143,7 +143,22 @@
 				[b]Note:[/b] This method suppresses any other connection request signals apart from [signal connection_drag_ended].
 			</description>
 		</method>
-		<method name="get_connection_line">
+		<method name="get_closest_connection_at_point" qualifiers="const">
+			<return type="Dictionary" />
+			<param index="0" name="point" type="Vector2" />
+			<param index="1" name="max_distance" type="float" default="4.0" />
+			<description>
+				Returns the closest connection to the given point in screen space. If no connection is found within [param max_distance] pixels, an empty [Dictionary] is returned.
+				A connection consists in a structure of the form [code]{ from_port: 0, from_node: "GraphNode name 0", to_port: 1, to_node: "GraphNode name 1" }[/code].
+				For example, getting a connection at a given mouse position can be achieved like this:
+				[codeblocks]
+				[gdscript]
+				var connection = get_closest_connection_at_point(mouse_event.get_position())
+				[/gdscript]
+				[/codeblocks]
+			</description>
+		</method>
+		<method name="get_connection_line" qualifiers="const">
 			<return type="PackedVector2Array" />
 			<param index="0" name="from_node" type="Vector2" />
 			<param index="1" name="to_node" type="Vector2" />
@@ -154,7 +169,14 @@
 		<method name="get_connection_list" qualifiers="const">
 			<return type="Dictionary[]" />
 			<description>
-				Returns an Array containing the list of connections. A connection consists in a structure of the form [code]{ from_port: 0, from_node: "GraphNode name 0", to_port: 1, to_node: "GraphNode name 1" }[/code].
+				Returns an [Array] containing the list of connections. A connection consists in a structure of the form [code]{ from_port: 0, from_node: "GraphNode name 0", to_port: 1, to_node: "GraphNode name 1" }[/code].
+			</description>
+		</method>
+		<method name="get_connections_intersecting_with_rect" qualifiers="const">
+			<return type="Dictionary[]" />
+			<param index="0" name="rect" type="Rect2" />
+			<description>
+				Returns an [Array] containing the list of connections that intersect with the given [Rect2]. A connection consists in a structure of the form [code]{ from_port: 0, from_node: "GraphNode name 0", to_port: 1, to_node: "GraphNode name 1" }[/code].
 			</description>
 		</method>
 		<method name="get_menu_hbox">
@@ -233,7 +255,7 @@
 		<member name="connection_lines_curvature" type="float" setter="set_connection_lines_curvature" getter="get_connection_lines_curvature" default="0.5">
 			The curvature of the lines between the nodes. 0 results in straight lines.
 		</member>
-		<member name="connection_lines_thickness" type="float" setter="set_connection_lines_thickness" getter="get_connection_lines_thickness" default="2.0">
+		<member name="connection_lines_thickness" type="float" setter="set_connection_lines_thickness" getter="get_connection_lines_thickness" default="4.0">
 			The thickness of the lines between the nodes.
 		</member>
 		<member name="focus_mode" type="int" setter="set_focus_mode" getter="get_focus_mode" overrides="Control" enum="Control.FocusMode" default="2" />
@@ -417,7 +439,16 @@
 	</constants>
 	<theme_items>
 		<theme_item name="activity" data_type="color" type="Color" default="Color(1, 1, 1, 1)">
-			Color of the connection's activity (see [method set_connection_activity]).
+			Color the connection line is interpolated to based on the activity value of a connection (see [method set_connection_activity]).
+		</theme_item>
+		<theme_item name="connection_hover_tint_color" data_type="color" type="Color" default="Color(0, 0, 0, 0.3)">
+			Color which is blended with the connection line when the mouse is hovering over it.
+		</theme_item>
+		<theme_item name="connection_rim_color" data_type="color" type="Color" default="Color(0.1, 0.1, 0.1, 0.6)">
+			Color of the rim around each connection line used for making intersecting lines more distinguishable.
+		</theme_item>
+		<theme_item name="connection_valid_target_tint_color" data_type="color" type="Color" default="Color(1, 1, 1, 0.4)">
+			Color which is blended with the connection line when the currently dragged connection is hovering over a valid target port.
 		</theme_item>
 		<theme_item name="grid_major" data_type="color" type="Color" default="Color(1, 1, 1, 0.2)">
 			Color of major grid lines/dots.
--- a/misc/extension_api_validation/4.2-stable.expected
+++ b/misc/extension_api_validation/4.2-stable.expected
@@ -59,3 +59,10 @@ Validate extension JSON: Error: Field 'classes/TileMap/methods/get_collision_vis
 Validate extension JSON: Error: Field 'classes/TileMap/methods/get_navigation_visibility_mode': is_const changed value in new API, from false to true.
 
 Two TileMap getters were made const. No adjustments should be necessary.
+
+
+GH-86158
+--------
+Validate extension JSON: Error: Field 'classes/GraphEdit/methods/get_connection_line': is_const changed value in new API, from false to true.
+
+get_connection_line was made const.
--- a/scene/gui/graph_edit.compat.inc
+++ b/scene/gui/graph_edit.compat.inc
@@ -38,9 +38,14 @@ void GraphEdit::_set_arrange_nodes_button_hidden_bind_compat_81582(bool p_enable
 	set_show_arrange_button(!p_enable);
 }
 
+PackedVector2Array GraphEdit::_get_connection_line_bind_compat_86158(const Vector2 &p_from, const Vector2 &p_to) {
+	return get_connection_line(p_from, p_to);
+}
+
 void GraphEdit::_bind_compatibility_methods() {
 	ClassDB::bind_compatibility_method(D_METHOD("is_arrange_nodes_button_hidden"), &GraphEdit::_is_arrange_nodes_button_hidden_bind_compat_81582);
 	ClassDB::bind_compatibility_method(D_METHOD("set_arrange_nodes_button_hidden", "enable"), &GraphEdit::_set_arrange_nodes_button_hidden_bind_compat_81582);
+	ClassDB::bind_compatibility_method(D_METHOD("get_connection_line", "from_node", "to_node"), &GraphEdit::_get_connection_line_bind_compat_86158);
 }
 
 #endif
--- a/scene/gui/graph_edit.cpp
+++ b/scene/gui/graph_edit.cpp
@@ -32,8 +32,10 @@
 #include "graph_edit.compat.inc"
 
 #include "core/input/input.h"
+#include "core/math/geometry_2d.h"
 #include "core/math/math_funcs.h"
 #include "core/os/keyboard.h"
+#include "scene/2d/line_2d.h"
 #include "scene/gui/box_container.h"
 #include "scene/gui/button.h"
 #include "scene/gui/graph_edit_arranger.h"
@@ -52,7 +54,6 @@ constexpr int MAX_CONNECTION_LINE_CURVE_TESSELATION_STAGES = 5;
 constexpr int GRID_MINOR_STEPS_PER_MAJOR_LINE = 10;
 constexpr int GRID_MIN_SNAPPING_DISTANCE = 2;
 constexpr int GRID_MAX_SNAPPING_DISTANCE = 100;
-constexpr float CONNECTING_TARGET_LINE_COLOR_BRIGHTENING = 0.4;
 
 bool GraphEditFilter::has_point(const Point2 &p_point) const {
 	return ge->_filter_input(p_point);
@@ -212,6 +213,36 @@ GraphEditMinimap::GraphEditMinimap(GraphEdit *p_edit) {
 	minimap_offset = minimap_padding + _convert_from_graph_position(graph_padding);
 }
 
+Ref<Shader> GraphEdit::default_connections_shader;
+
+void GraphEdit::init_shaders() {
+	default_connections_shader.instantiate();
+	default_connections_shader->set_code(R"(
+// Connection lines shader.
+shader_type canvas_item;
+render_mode blend_mix;
+
+uniform vec4 rim_color : source_color;
+uniform int from_type;
+uniform int to_type;
+uniform float line_width;
+
+void fragment(){
+	float fake_aa_width = 1.5/line_width;
+	float rim_width = 1.5/line_width;
+
+	float dist = abs(UV.y - 0.5);
+	float alpha = smoothstep(0.5, 0.5-fake_aa_width, dist);
+	vec4 final_color = mix(rim_color, COLOR, smoothstep(0.5-rim_width, 0.5-fake_aa_width-rim_width, dist));
+	COLOR = vec4(final_color.rgb, final_color.a*alpha);
+}
+)");
+}
+
+void GraphEdit::finish_shaders() {
+	default_connections_shader.unref();
+}
+
 Control::CursorShape GraphEdit::get_cursor_shape(const Point2 &p_pos) const {
 	if (moving_selection) {
 		return CURSOR_MOVE;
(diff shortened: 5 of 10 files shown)


nicolasgustafsson:
The active connection line looks a little ugly when dragging in the gif you posted, is it possible to put it lower than the port in the z hierarchy? 
image

YuriSizov:
Out of curiosity, why did you decide to supersede your other PR instead of pushing an update to it? Also what exactly breaks compatibility here?

Calinou:
> Decide what to do with the anti aliased property since AA is basically free now

I'd probably leave it in, as I remember people using GraphNode/GraphEdit in games with a pixel art style.

Geometror:
@nicolasgustafsson There is definitely some room for improvement there, but since the whole line needs to be on top we can't draw it behind the port. However, I have some ideas, e.g. you could fade the rim (which would require passing the line length as an uniform).

@YuriSizov Well, although this PR includes the features of the other one, the approach for updating/rendering the connections is quite different. Since that was a big part of the other PR and @ATS already approved it, I thought it would be better to open a separate PR instead of updating the existing one because a lot has changed.
Regarding the breaks-compat tag: it's just minor things like the following, so I added it just in case: 
- I'm still fiddling with the connection lines thickness property there (the shader and the width of the generated polyline need to be coordinated/tuned); currently it has a different scale (which should better resemble the width in pixels) and therefore an adjusted default value to compensate this
- the unexposed, but public method `get_top_layer` (which seems to be unused as I just have noticed) returns the top layer which no longer draws the dragged connection/handles and filters the input for it
- the anti-aliasing property currently only anti-aliases the connections in the minimap (of course this point depends on our decision regarding that property)
- well, some of the improvements (like the rim or the applied gradient on hovering a port) change the functionality and aren't really optional (although you can turn the rim off by specifying a transparent rim color)

@Calinou Well in that case a connection line with AA turned off might be still too smooth, for something like this you're able to override the default connection line shader and achieve a true pixelated look. You can do basically everything you like as you can see in the example below ;)

ge-supercharged

YuriSizov:
> I thought it would be better to open a separate PR instead of updating the existing one because a lot has changed.

From the production point of view, you can just dismiss a stale review to indicate that your work needs to be reviewed again (and of course I would assume you leave a comment with an explanation of updates anyway). So for everyone else's benefit this wasn't necessary. But if it helps you, then opening a new PR is fine. Feel free to close the old one if you're replacing it.

AThousandShips:
As the old PR didn't have review stuff on it I'd be fine but otherwise I'd suggest avoiding superseding PRs unless most of the feedback on it is significantly outdated, to not lose a lot of progress in the feedback there

doc/classes/GraphEdit.xml
```diff
@@ -154,7 +169,14 @@
 		<method name="get_connection_list" qualifiers="const">
 			<return type="Dictionary[]" />
 			<description>
-				Returns an Array containing the list of connections. A connection consists in a structure of the form [code]{ from_port: 0, from: "GraphNode name 0", to_port: 1, to: "GraphNode name 1" }[/code].
+				Returns an [Array] containing the list of connections. A connection consists in a structure of the form [code]{ from_port: 0, from: "GraphNode name 0", to_port: 1, to: "GraphNode name 1" }[/code].
```
AThousandShips, line 172:
```suggestion
				Returns an [Array] containing the list of connections. A connection consists in a structure of the form [code]{ from_port: 0, from_node: "GraphNode name 0", to_port: 1, to_node: "GraphNode name 1" }[/code].
```
By the code

doc/classes/GraphEdit.xml
```diff
@@ -154,7 +169,14 @@
...
+			</description>
+		</method>
+		<method name="get_connections_intersecting_with_rect">
+			<return type="Dictionary[]" />
+			<param index="0" name="rect" type="Rect2" />
+			<description>
+				Returns an [Array] containing the list of connections that intersect with the given [Rect2]. A connection consists in a structure of the form [code]{ from_port: 0, from: "GraphNode name 0", to_port: 1, to: "GraphNode name 1" }[/code].
```
AThousandShips, line 179:
```suggestion
				Returns an [Array] containing the list of connections that intersect with the given [Rect2]. A connection consists in a structure of the form [code]{ from_port: 0, from_node: "GraphNode name 0", to_port: 1, to_node: "GraphNode name 1" }[/code].
```

doc/classes/GraphEdit.xml
YuriSizov: Note that this will be fixed in master by https://github.com/godotengine/godot/pull/86161, because it's an existing problem.

Geometror: Approved the other PR, so I think I can mark this as resolved.

QbieShay:
> There is definitely some room for improvement there, but since the whole line needs to be on top we can't draw it behind the port. However, I have some ideas, e.g. you could fade the rim (which would require passing the line length as an uniform).

Why does it need to draw on top of the nodes?

Geometror:
Well, in a dense graph you would only see tiny segments of your dragged connection line which makes it quite unfun/a bit awkward to use. (Every graph editor I tried so far draws it over everything else while dragging, so Godot isn't an exception here)

EnlightenedOne:
Cant you just redraw the connecting circle along with the line over the top of everything in order to have both? Slightly more faff granted but that is what Blender does:
image

Geometror:
Rebased after the editor theme generation refactoring.
Regarding the dragging connection line: I'd rather like to find a solution in a followup PR, since I noticed that it's actually not trivial to find one that has no drawbacks (e.g. the port "circle" is actually a texture that can be custom). And I like to get this merged as soon as possible, because it needs to be tested well and some other stuff depends on it.

Geometror:
Yes, that's almost how it works right now:
- the connection layer containing all Line2Ds is drawn behind all graph nodes
- the top connection layer contains one Line2D which is used for the dragged line (the point array is reset when no connection is dragged), when dropping, a Line2D is created as a child of the connection layer

Since the dragged connection line isn't updated in a draw function we can't just use the `CanvasItem` draw methods to draw two circles in `_update_top_connection_layer()`. Using round end caps helps a tiny bit, but doesn't work well with the line shader (to be precise the line rim) because the UVs aren't generated in a way that would allow using them as a distance field.
Tbh honest, I don't think this looks that ugly, especially at zoom levels <100% - and when quickly dragging connections you won't even notice it :) That's another reason why I wanted to solve this in another PR.

YuriSizov:
I think the only solution there is to fetch the icon used for the source/target port and draw it on top of the line. But indeed, this can be fixed later.

Could you squash your commits and give them a more conventional message, following our contribution guidelines? Like the current title of the PR, perhaps.

core/math/geometry_2d.h
```diff
@@ -249,6 +253,28 @@ class Geometry2D {
...
+	static bool segment_intersects_rect(const Vector2 &p_from, const Vector2 &p_to, const Rect2 &p_rect) {
+		if (p_rect.has_point(p_from) || p_rect.has_point(p_to)) {
+			return true;
+		}
+
+		const Vector2 rect_points[4] = {
+			p_rect.position,
+			p_rect.position + Vector2(p_rect.size.x, 0),
+			p_rect.position + p_rect.size,
+			p_rect.position + Vector2(0, p_rect.size.y)
```
YuriSizov, line 265: You could alternatively use `position`/`end` and their individual `x` and `y`, but I'm not sure if it would help readability. This is probably better.

doc/classes/GraphEdit.xml
```diff
@@ -143,6 +143,21 @@
 				[b]Note:[/b] This method suppresses any other connection request signals apart from [signal connection_drag_ended].
 			</description>
 		</method>
+		<method name="get_closest_connection_at_point">
+			<return type="Dictionary" />
+			<param index="0" name="point" type="Vector2" />
+			<param index="1" name="max_distance" type="float" default="4.0" />
+			<description>
+				Returns the closest connection to the given point in screen space. If no connection is found within [param max_distance] pixels, an empty [Dictionary] is returned.
+				A connection consists in a structure of the form [code]{ from_port: 0, from: "GraphNode name 0", to_port: 1, to: "GraphNode name 1" }[/code].
```
YuriSizov, line 152:
```suggestion
				A connection consists in a structure of the form [code]{ from_port: 0, from_node: "GraphNode name 0", to_port: 1, to_node: "GraphNode name 1" }[/code].
```

I think?

scene/gui/graph_edit_arranger.cpp
```diff
@@ -437,63 +435,61 @@ float GraphEditArranger::_calculate_threshold(const StringName &p_v, const Strin
...
 		int min_order = MAX_ORDER;
-		GraphEdit::Connection incoming;
-		List<GraphEdit::Connection> connection_list;
-		graph_edit->get_connection_list(&connection_list);
-		for (List<GraphEdit::Connection>::Element *E = connection_list.front(); E; E = E->next()) {
-			if (E->get().to_node == p_w) {
-				ORDER(E->get().from_node, r_layers);
+		Ref<GraphEdit::Connection> incoming;
+		List<Ref<GraphEdit::Connection>> connection_list = graph_edit->get_connection_list();
+		for (List<Ref<GraphEdit::Connection>>::Element *E = connection_list.front(); E; E = E->next()) {
```
YuriSizov, line 440: We can iterate lists with something like

```suggestion
		for (cosnt Ref<GraphEdit::Connection> &connection : connection_list) {
```

scene/gui/graph_edit_arranger.cpp
```diff
@@ -437,63 +435,61 @@ float GraphEditArranger::_calculate_threshold(const StringName &p_v, const Strin
...
 		int min_order = MAX_ORDER;
-		GraphEdit::Connection outgoing;
-		List<GraphEdit::Connection> connection_list;
-		graph_edit->get_connection_list(&connection_list);
-		for (List<GraphEdit::Connection>::Element *E = connection_list.front(); E; E = E->next()) {
-			if (E->get().from_node == p_w) {
-				ORDER(E->get().to_node, r_layers);
+		Ref<GraphEdit::Connection> outgoing;
+		List<Ref<GraphEdit::Connection>> connection_list = graph_edit->get_connection_list();
+		for (List<Ref<GraphEdit::Connection>>::Element *E = connection_list.front(); E; E = E->next()) {
```
YuriSizov, line 471: Same here.

scene/gui/graph_edit_arranger.cpp
```diff
@@ -77,15 +74,16 @@ void GraphEditArranger::arrange_nodes() {
 		if (graph_element->is_selected() || arrange_entire_graph) {
 			selected_nodes.insert(graph_element->get_name());
 			HashSet<StringName> s;
-			for (List<GraphEdit::Connection>::Element *E = connection_list.front(); E; E = E->next()) {
-				GraphNode *p_from = Object::cast_to<GraphNode>(node_names[E->get().from_node]);
-				if (E->get().to_node == graph_element->get_name() && (p_from->is_selected() || arrange_entire_graph) && E->get().to_node != E->get().from_node) {
+			List<Ref<GraphEdit::Connection>> connection_list = graph_edit->get_connection_list();
+			for (List<Ref<GraphEdit::Connection>>::Element *E = connection_list.front(); E; E = E->next()) {
```
YuriSizov, line 78: Same here as well.

scene/gui/graph_edit_arranger.cpp
```diff
@@ -77,15 +74,16 @@ void GraphEditArranger::arrange_nodes() {
 		if (graph_element->is_selected() || arrange_entire_graph) {
 			selected_nodes.insert(graph_element->get_name());
 			HashSet<StringName> s;
-			for (List<GraphEdit::Connection>::Element *E = connection_list.front(); E; E = E->next()) {
-				GraphNode *p_from = Object::cast_to<GraphNode>(node_names[E->get().from_node]);
-				if (E->get().to_node == graph_element->get_name() && (p_from->is_selected() || arrange_entire_graph) && E->get().to_node != E->get().from_node) {
+			List<Ref<GraphEdit::Connection>> connection_list = graph_edit->get_connection_list();
```
YuriSizov, line 77: I'm confused, why would we want to fetch it for every item again and again?

scene/gui/graph_edit.h
```diff
@@ -344,12 +378,20 @@ class GraphEdit : public Control {
...
 	void clear_connections();
+
+	const List<Ref<Connection>> &get_connection_list() const;
 	void force_connection_drag_end();
 
 	virtual PackedVector2Array get_connection_line(const Vector2 &p_from, const Vector2 &p_to);
+	Ref<Connection> get_closest_connection_at_point(const Vector2 &p_point, float p_max_distance = 4.0);
+	Dictionary _get_closest_connection_at_point(const Vector2 &p_point, float p_max_distance = 4.0);
+	List<Ref<Connection>> get_connections_intersecting_with_rect(const Rect2 &p_rect);
+	TypedArray<Dictionary> _get_connections_intersecting_with_rect(const Rect2 &p_rect);
```
YuriSizov, line 389: Why are these public but underscored?

scene/gui/graph_edit.cpp
```diff
@@ -1622,6 +1904,16 @@ void GraphEdit::_update_zoom_label() {
...
 }
 
+void GraphEdit::_invalidate_connection_line_cache() {
+	for (Ref<Connection> &c : connections) {
+		c->_cache.dirty = true;
+	}
+}
+
+float GraphEdit::_get_shader_line_width() {
+	return MAX(0, lines_thickness * theme_cache.base_scale + 4.0);
```
YuriSizov, line 1914: Is there a case when thickness or scale can be negative? Otherwise, `MAX` is not really needed, as `+4` guarantees positivity.

scene/gui/graph_edit.cpp
```diff
@@ -534,6 +611,7 @@ void GraphEdit::_notification(int p_what) {
 			v_scrollbar->set_anchor_and_offset(SIDE_RIGHT, ANCHOR_END, 0);
 			v_scrollbar->set_anchor_and_offset(SIDE_TOP, ANCHOR_BEGIN, 0);
 			v_scrollbar->set_anchor_and_offset(SIDE_BOTTOM, ANCHOR_END, 0);
+
```
YuriSizov, line 614: Seems unnecessary :)

YuriSizov:
Overall code seems fine, good job! Functionality-wise it all seems to work. So once the feedback is addressed, this should be good to go.

----

I also noticed an issue with visual shaders as I was testing. Reported here https://github.com/godotengine/godot/issues/87337

scene/gui/graph_edit_arranger.cpp
Geometror: I'm confused as well, it's a mistake :)

scene/gui/graph_edit.cpp
Geometror: Well, the inspector currently allows setting negative values which makes no sense, so I just added a property hint: `PROPERTY_HINT_RANGE, "0,100,0.1,suffix:px"` and a check to the setter
`ERR_FAIL_COND_MSG(p_thickness < 0, "Connection lines thickness must be greater than or equal to 0.");`.
Now `MAX` is definitely not needed anymore.

core/math/geometry_2d.h
Geometror: I think I'll leave it this way.

Geometror:
I think I addressed all the review comments :)

YuriSizov:
Builds failing because constness change needs to be added to the extension compatibility file

scene/gui/graph_edit.cpp
Geometror: Uhm, that's what I meant with :)
> and a check to the setter
ERR_FAIL_COND_MSG(p_thickness < 0, "Connection lines thickness must be greater than or equal to 0.");

YuriSizov: Sorry, I'm blind <3 All good!

YuriSizov (approved):
Aside from the compatibility check, it's all good 👍

AThousandShips (approved):

Geometror:
I also forgot to adjust the commit message. Both should be fixed now.

AThousandShips (changes requested):
You need to add a compatibility method

misc/extension_api_validation/4.2-stable.expected
```diff
@@ -59,3 +59,9 @@ Validate extension JSON: Error: Field 'classes/TileMap/methods/get_collision_vis
 Validate extension JSON: Error: Field 'classes/TileMap/methods/get_navigation_visibility_mode': is_const changed value in new API, from false to true.
 
 Two TileMap getters were made const. No adjustments should be necessary.
+
+GH-86158
```
AThousandShips, line 63:
```suggestion

GH-86158
```

scene/gui/graph_edit.h
```diff
@@ -313,6 +346,7 @@ class GraphEdit : public Control {
 #ifndef DISABLE_DEPRECATED
 	bool _is_arrange_nodes_button_hidden_bind_compat_81582() const;
 	void _set_arrange_nodes_button_hidden_bind_compat_81582(bool p_enable);
+	void _get_connection_line_bind_compat_86158(const Vector2 &p_from, const Vector2 &p_to);
```
AThousandShips, line 349:
```suggestion
	PackedVector2Array _get_connection_line_bind_compat_86158(const Vector2 &p_from, const Vector2 &p_to);
```

scene/gui/graph_edit.compat.inc
```diff
@@ -38,9 +38,14 @@ void GraphEdit::_set_arrange_nodes_button_hidden_bind_compat_81582(bool p_enable
...
 }
 
+PackedVector2Array GraphEdit::_get_connection_line_bind_compat_86158(const Vector2 &p_from, const Vector2 &p_to) {
+	return get_connection_line(p_from, p_to);
+}
+
 void GraphEdit::_bind_compatibility_methods() {
 	ClassDB::bind_compatibility_method(D_METHOD("is_arrange_nodes_button_hidden"), &GraphEdit::_is_arrange_nodes_button_hidden_bind_compat_81582);
 	ClassDB::bind_compatibility_method(D_METHOD("set_arrange_nodes_button_hidden", "enable"), &GraphEdit::_set_arrange_nodes_button_hidden_bind_compat_81582);
+	ClassDB::bind_compatibility_method(D_METHOD("get_connection_line"), &GraphEdit::_get_connection_line_bind_compat_86158);
```
AThousandShips, line 48:
```suggestion
	ClassDB::bind_compatibility_method(D_METHOD("get_connection_line", "from_node", "to_node"), &GraphEdit::_get_connection_line_bind_compat_86158);
```

scene/gui/graph_edit.h
Geometror: 🤦‍♂️

AThousandShips (approved):

YuriSizov:
Thanks!

MewPurPur:
I found this in git blame as I remembered that it upsets Aaron's attempt to implement a disable_2d build option, forcing it to make an exception for Line2D. Would it make sense to use draw_polyline on a Control with the same shader instead?

> Connection lines are now drawn using Line2D nodes instead of using draw_polyline calls in only one CanvasItem
>> The performance concerns I had turned out to be unfounded, this method even allows for some nice optimizations
>> The reason for this was flexibility: e.g. drawing e.g. type-dependent connection line styles (double/triple/quadruple lines, dashed) with a shader requires passing per-line uniforms to the shaders

Geometror:
draw_polyline is not sufficient (line caps and UVs are needed), but since Line2D uses another abstraction (LineBuilder) we could do something like that (basically by using the same drawing code as in Line2D).

PR #74626, opened 2023-03-08
Select the newly duplicated file

KoBeWi:
When you duplicate a file, the old one retains the selection. This is especially confusing when you duplicate a file to some faraway folder and then try to edit it, only to discover that you are editing the old file.

This PR makes the newly duplicated file selected.

https://user-images.githubusercontent.com/2223172/223872333-8f51548c-c6f5-4298-acf6-455cb640f3dd.mp4

PR #86164, opened 2023-12-14
Properly select the newly duplicated file

KoBeWi:
Follow-up to #74626

The previous selecting method wasn't reliable enough. The worst part was that Import dock was not updated to the new file, which caused me to change import type of a wrong file a couple of times. This PR ensures that the path is selected properly.

### DIFF
--- a/editor/filesystem_dock.cpp
+++ b/editor/filesystem_dock.cpp
@@ -1296,6 +1296,13 @@ void FileSystemDock::_fs_changed() {
 		_update_file_list(true);
 	}
 
+	if (!select_after_scan.is_empty()) {
+		_navigate_to_path(select_after_scan);
+		select_after_scan.clear();
+		import_dock_needs_update = true;
+		_update_import_dock();
+	}
+
 	set_process(false);
 }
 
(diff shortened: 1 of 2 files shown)


editor/filesystem_dock.cpp
```diff
@@ -1296,6 +1296,13 @@ void FileSystemDock::_fs_changed() {
 		_update_file_list(true);
 	}
 
+	if (!select_after_scan.is_empty()) {
+		navigate_to_path(select_after_scan);
+		select_after_scan = String();
+		import_dock_needs_update = true;
+		_update_import_dock();
```
YuriSizov, line 1303: Do you know why we call it deferred everywhere else?

editor/filesystem_dock.cpp
```diff
@@ -1296,6 +1296,13 @@ void FileSystemDock::_fs_changed() {
 		_update_file_list(true);
 	}
 
+	if (!select_after_scan.is_empty()) {
+		navigate_to_path(select_after_scan);
```
YuriSizov, line 1300: Looks to me that we can use the private method directly here, `_navigate_to_path`. Seems to be what we call consistently throughout this class and the public one doesn't do anything useful to this use case.

YuriSizov (approved):
Haven't tested, but looks sensible overall.

PS. I also reclassified it as a bug since not behaving reliably is pretty much a bug.

editor/filesystem_dock.cpp
KoBeWi: Probably to avoid multiple updates in a single frame. There is no such risk here.

editor/filesystem_dock.cpp
```diff
@@ -1296,6 +1296,13 @@ void FileSystemDock::_fs_changed() {
...
+		select_after_scan = String();
```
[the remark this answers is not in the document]
KoBeWi: Doesn't seem widely used, but why not.

YuriSizov:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

PR #86169, opened 2023-12-14
Replace `GDVIRTUAL_CALL` with `GDVIRTUAL_REQUIRED_CALL` where applicable

Repiteo:
A handful of `GDVIRTUAL_CALL` implementations were used as booleans for a check exclusively to see if they're implemented, throwing an error if they weren't. This functionality is already present in `GDVIRTUAL_REQUIRED_CALL`, though it was added later on compared to the bulk of these implementations, so this PR updates them accordingly.

### DIFF
--- a/editor/editor_script.cpp
+++ b/editor/editor_script.cpp
@@ -62,9 +62,7 @@ EditorInterface *EditorScript::get_editor_interface() const {
 }
 
 void EditorScript::run() {
-	if (!GDVIRTUAL_CALL(_run)) {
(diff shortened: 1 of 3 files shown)


dsnopek (approved):
Thanks!

I didn't test this, but the code changes look correct: the result should be same for each of the changed instances. This will be good clean-up :-)

YuriSizov (approved):

YuriSizov:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #81725, opened 2023-09-16
Fix folder moving in file system dock

Rindbee:
Include files in the folder that will be moved/renamed when looking up the file owner. 

Fixed issue when moving/renaming directories, see https://github.com/godotengine/godot/issues/77289#issuecomment-1722027314.

Rename the function/variable names to make more sense.

PR #86177, opened 2023-12-14
Fix file disappearing when renaming dependencies

### LINKED ISSUE #86174: Renaming a .gd file deletes the previously linked .tscn
### Tested versions

- Reproducible in 4.2.stable, 4.2.1.stable
- Not reproducible in 4.1.stable

### System information

Godot v4.2.1.stable - Ubuntu 22.04.3 LTS 22.04 - X11 - GLES3 (Compatibility) - NVIDIA GeForce GTX 1080 Ti (nvidia; 535.129.03) - 12th Gen Intel(R) Core(TM) i3-12100F (8 Threads)

### Issue description

Godot deletes the `.tscn` file when renaming a previously linked `.gd` file. The deleted file can't be recovered.

### Steps to reproduce

1. Create a new project.
2. Create a new scene of type Control, and save it
3. Right click the root node "Control", and select "Attach Script..."
4. Right click the root node "Control", and select "Detach Script..."
5. On the FileSystem tab right click "control.gd", and select "Rename..."
6. Move the cursor just before the dot and write "_old", the new name should be "control_old.gd"
7. Press enter

Test Project for Godot 4.2.1
You can start from step **4**.
[RenamingBugTest.zip](https://github.com/godotengine/godot/files/13677960/RenamingBugTest.zip)

KoBeWi:
Fixes #86174

When renaming a file, the editor will try to fix dependencies, but it will save affected scenes before that. However as a result of saving, the dependency might get removed thus a rename is performed on a dependency that no longer exists.

Then there is rename method that creates a `.depren` file after renaming dependency. However it treats "no rename" as success
https://github.com/godotengine/godot/blob/f8a2a9193662b2e8c1d04d65e647399dee94f31e/scene/resources/resource_format_text.cpp#L921-L924
meaning the file is not created, but the code continues normally. This results in a disaster.

The fix checks whether `.depren` file actually exists before removing the old file. Another option would be making the rename method return an error if the dependency can't be found, but I'm not sure if it's always expected to exist (in case of the fixed issue it didn't, but maybe it's expected to fail, idk).
I modified the binary format too, but I didn't test if the same bug occurs. Better to be safe anyway.

This was probably regression from #81725 or #81657

### DIFF
--- a/core/io/resource_format_binary.cpp
+++ b/core/io/resource_format_binary.cpp
@@ -1454,8 +1454,10 @@ Error ResourceFormatLoaderBinary::rename_dependencies(const String &p_path, cons
 	fw.unref();
 
 	Ref<DirAccess> da = DirAccess::create(DirAccess::ACCESS_RESOURCES);
-	da->remove(p_path);
-	da->rename(p_path + ".depren", p_path);
+	if (da->exists(p_path + ".depren")) {
+		da->remove(p_path);
+		da->rename(p_path + ".depren", p_path);
+	}
 	return OK;
 }
 
--- a/scene/resources/resource_format_text.cpp
+++ b/scene/resources/resource_format_text.cpp
@@ -1790,8 +1790,8 @@ Error ResourceFormatLoaderText::rename_dependencies(const String &p_path, const 
 		err = loader.rename_dependencies(f, p_path, p_map);
 	}
 
-	if (err == OK) {
-		Ref<DirAccess> da = DirAccess::create(DirAccess::ACCESS_RESOURCES);
+	Ref<DirAccess> da = DirAccess::create(DirAccess::ACCESS_RESOURCES);
+	if (err == OK && da->file_exists(p_path + ".depren")) {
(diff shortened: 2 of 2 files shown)


adamscott (approved):
@YuriSizov This should be a patch for 4.2.2

Jordyfel:
Since saving scenes can change dependencies, what do you think about swapping these calls:
https://github.com/godotengine/godot/blob/f8a2a9193662b2e8c1d04d65e647399dee94f31e/editor/filesystem_dock.cpp#L2004-L2009

I also feel like trying to rename a missing dependency should be an error.

KoBeWi:
`save_scene_list()` uses `r_file_owners()`, which is filled by `_find_file_owners()`, so they can't be swapped. Maybe an alternative is simply saving every scene, but changing behavior would require more testing. The current solution is a safer option.

YuriSizov:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

PR #57025, opened 2022-01-21
[WIP] Implement a Wayland DisplayServer

deralmas:
Superceded by #86180.

Resolves godotengine/godot-proposals#990.

**Multi windowing *won't* be implemented in this PR due to some incompatibilities with Wayland's and Godot windowing API which have to be sorted out. Trust me, I tried. Doing otherwise would just stall everything. I'm sorry.**

The aim of this PR is to eventually implement a fully working `DisplayServer` compatible with the [Wayland protocol](https://wayland.freedesktop.org/) on Godot 4.x, which had architectural changes that led to the closing of #27463.

## Status

It is currently **beta quality**. It's definitely usable and most features are implemented but there's still lots of testing to do. 

As there's a growing need of this backend due to poor XWayland compatibility, the main focus of this PR is polishing and documenting the current feature set as best as we can so that we can get this merged and stable by 4.3. Everything that's not implemented yet will be implemented post-merge.

For ease of testing this PR will track, whenever possible, the latest (pre)release closer to `master`.

Current version tracked: **4.2-stable**

For a list of what's missing and of known issues see below.

## Testing

Thanks to GitHub Actions, every commit is built and the result can be downloaded up to 90 days since the last change. See [Testing pull requests](https://docs.godotengine.org/en/stable/contributing/workflow/testing_pull_requests.html) for info on how to get a build.

**Note: for users without an account, the steps outlined in the documentation are broken. Instead of putting a link to the workflow file in nightly.link, go to checks, click on a random job on the left bar and copy/paste the URL up to `runs/NUMBERS`. You'll be able then to download the latest build.**

This means that usually there's no compilation required as it's already done for you.

At startup the engine will try initializing a Wayland display. If it fails, it prints an error and tries with X11. There's no need to have both libraries installed to run the executable as they're dynamically loaded at runtime if available.

Note that it's always possible to force the X11 backend to start first (triggering XWayland if available) by appending the `--display-driver x11` argument.

This PR is also completely HiDPI aware and includes fractional scaling (except for the cursor). Note that there are still some known issues with the HiDPI implementation in the cases of fractional scaling.

## Building

If for whatever reason there's the need of building this branch from scratch, a plain `scons` invocation should work fine, building an hybrid dynamically loaded X11/Wayland executable.

Every new header and dynamic library wrapper is embedded in the source, so there are no extra build dependencies to download, with the current exception of `wayland-scanner`, required for generating protocol wrappers at build time.

### Important methods missing
[...]

PR #80231, opened 2023-08-03
Support detecting and mapping Ctrl/Alt/Shift/Meta by their left/right physical location

romlok:
This PR is intended to solve https://github.com/godotengine/godot-proposals/issues/1387.  It comes in two steps/commits for now, in case the groundwork is found acceptable while the extension may be contentious.

## 1. Add `InputEventKey.location`

Adds a new enum `KeyLocation` (`UNKNOWN`, `LEFT`, `RIGHT`) and associated property `InputEventKey.location`, which indicates the location of keys which have both left and right versions, eg. shift, ctrl. Eg:

```gdscript
func _input(event: InputEvent):
    if event is InputEventKey:
        match event.location:
            KEY_LOCATION_LEFT:
                $Left.flip_v = event.is_pressed()
            KEY_LOCATION_RIGHT:
                $Right.flip_v = event.is_pressed()
            KEY_LOCATION_UNKNOWN:
                $Unknown.flip_v = event.is_pressed()
```

On its own, this part should have no impact or implication on backward compatibility, since no existing functionality is altered.

I've implemented this for (AIUI) all the supported platforms, but have only been able to test Linux and Web builds, so I need help verifying that left/right detection works correctly on the other platforms:

* [x] Linux
* [x] Web
* [x] Windows
* [x] MacOS
* [x] iOS
* [ ] Android

## 2. Allow mapping key events by location

The second commit addresses the changes needed to configure and match events against their `location`.  I'm not that familiar with the codebase (or C++ for that matter), so I don't know that I've updated everywhere that should be updated.

Screenshot_20230805_130332-1

Test/demo project: [key-location-demo-2.zip](https://github.com/godotengine/godot/files/12267587/key-location-demo-2.zip)

Screenshot_20230805_012629

### Notes:

I had to make some decisions as to the behaviour, based on my own best guess as to user expectations:

* All `InputEventKey`s store a `location` value, but only those which are mapped "physically" use it in comparisons.
* An event with an `UNKNOWN` location matches also both `LEFT` and `RIGHT` versions, but not vice versa.
* By the above, `InputEventKey.is_match` is now non-commutative (if it wasn't before) - eg. `a->is_match(b)` does not imply `b->is_match(a)`.
* The mapping UI defaults to `UNKNOWN` ("Either"), which persists through input detection until manually changed.
* The "Physical location" section of the UI only appears for the four keys that have twins (which are hard-coded right now).
[...]

Issue #77337, opened 2023-05-22
Window flickering


d0n3val:
### Godot version

4.0.3-stable, I've tested it also on 4.0.2 and 4.0.0

### System information

Windows 10, NVidia 4070 latest drivers 

### Issue description

Only by clicking on the window I get a weird flickering on the screen. It happens even before selecting any project, in the launcher, but it is more apparent opening an empty project. It seems to react to just clicking with the mouse on the screen, but in the end it happens by itself.

https://github.com/godotengine/godot/assets/627698/694dfe93-6423-46ce-a69d-45815731da8d

### Steps to reproduce

Just open the launcher and click on it, you will start seeing black frames
Open an empty project, click in the editor also will give black frames.
Run an empty project, the window will start showing (not responding) in the title as you just click on it

### Minimal reproduction project

No need, it already happens in the launcher. It is more apparent with a full empty project running (check video)

Calinou:
Can you reproduce this after switching to the Mobile or Compatibility rendering methods in the top-right corner?

Can you reproduce this after enabling **Single Window Mode** in the Editor Settings, then restarting the editor?

d0n3val:
I can confirm it still happens in Mobile and Compatibility rendering method.
I tried also switching Single Window Mode in and out and it does not affect the outcome.

Feels to me like a 4070 driver bug. My old 2070 was working fine.

How can I be of more help ?

Calinou:
> Feels to me like a 4070 driver bug. My old 2070 was working fine.

It works just fine for me on a RTX 4090 and a recent NVIDIA driver downloaded from nvidia.com (not sure which one exactly, I'm not on Windows right now). Make sure to use a driver from nvidia.com, not Windows Update.

d0n3val:
I have tried the nvidia drivers 531.79 and the two previous ones.
I am right now on nvidia studio driver 531.61.

The problem persist in all those.

I have a software to handle multiple desktops called SylphyhornPlus - I've tried to remove but problem still persist. A part from that is a recently installed window 10, not much more software on it.

My problem seems similar but unrelated with #32990 since the flickering is not when using pop-ups, just by clicking on the screen anywhere.

d0n3val:
I have tried the last three nvidia drivers, right now testing the studio drivers but no luck, it still happens. I've also uninstall th nvidia drivers and I still have it. Perhaps it is not tied to graphics drivers after all. It seems connected to just receiving input.

[...]

PR #86180, opened 2023-12-15
Add Wayland support

### LINKED ISSUE #990: Add support for the Wayland display server
Godot's platform ports.

None directly as X11 still works decently on recent Linux distros, but we need to add support for the more modern Wayland display server eventually.

Wayland is a newer, more modern display server that can be used on Linux and likely some BSD-derived OSes too (now or in the future).

Many Linux distros / DEs now run on Wayland by default, and Godot only works thanks to the `xwayland` compatibility layer.

Having builtin Wayland support will give us access to a much cleaner API as well as more advanced features for window management, etc.

Supersedes https://github.com/godotengine/godot/issues/576 - see discussion there for relevant technical details.

I'm not an expert on Wayland so I'll let others fill in and explain what are the gains we can expect from having built-in Wayland support.

From a technical implementation point of view, the following points are important:
- Implement Wayland support was intentionally kept on hold until we would split the OS and DisplayServer components, so that we can have a Linux/*BSD platform port with more than one DisplayServer implementations. This has now been done in the `master` branch, and thus Wayland support can be implemented for Godot 4.0. There's no intention to backport it to earlier releases.
- There are many issues on all platforms following the DisplayServer split and implementation of multi-window support (https://github.com/godotengine/godot/issues/37734). It is recommended to wait for most of those issues to be fixed before starting implementing a new DisplayServer (though help is very welcome from Wayland aficionados to first fix the outstanding X11 issues after this refactoring).
- Wayland and X11 support should be implemented in a way that works with a single Godot binary, i.e. the relevant libraries should probably be `dlopen`-ed so that the same binary can be used on systems with only X11, only Wayland, or both installed.

No.

This can only be done in core.

deralmas:
Squashed version of #57025 laid out for ease of reviewing.

Resolves godotengine/godot-proposals#990.

**See #57025 for further information.**

Not everything is yet implemented, either for Godot or personal limitations (I don't have all hardware in the world). A brief list of the most important issues follows:

- Single-window only: the `DisplayServer` API doesn't expose enough information for properly creating XDG shell windows.

- Very dumb rendering loop: this is very complicated, just know that the low consumption mode is forced to 2000 Hz and some clever hacks are in place to overcome a specific Wayland limitation. This will be improved to the extent possible both downstream and upstream.
**EDIT: better approach available for testing over at #87750**

- Features to implement yet: IME, touch input, native file dialog, drawing tablet (commented out due to a refactor), screen recording.

- Mouse passthrough can't be implement through a poly API, we need a rect-based one.

- The cursor doesn't yet support fractional scaling.

- Auto scale is rounded up when using fractional scaling as we don't have a per-window scale query API (basically we need `DisplayServer::window_get_scale`).

- Building with `x11=no wayland=yes opengl=yes openxr=yes` fails.

This also adds a new project property and editor setting for selecting the default DisplayServer to start, to allow this backend to start first in exported projects (X11 is still the default for now). The editor setting always overrides the project setting.

Special thanks to Drew Devault, toger5, Sebastian Krzyszkowiak, Leandro Benedet Garcia, Subhransu, Yury Zhuravlev and Mara Huldra.

## How to review this PR

Don't let the diff size scare you, it's smaller than you think!

"Wayland" is mostly an IPC; the actual interfaces are defined in XML "protocol" files. It has a very lean core protocol designed to be extended. Each of those XML files is used to generate the C wrappers that the backend uses through an utility named `wayland-scanner` which must be available on the building system (`libwayland-bin` package on Debian). We generate them this way because the result changes between Wayland versions and so it's good practice for all Wayland programs.

The core protocol is present in `thirdparty/wayland` (+3087) and the various extensions are present in `thirdparty/wayland-protocols` (+4303), copied verbatim as in `thirdparty/README.md` (+43).

Talking about `thirdparty`, various dependencies got dynwrapped into `platform/linuxbsd/wayland/dynwrappers` (+1698) and their library headers got put into `thirdparty/linuxbsd_headers/` (+7929), according to `thirdparty/linuxbsd_headers/README.md` (+14).

Note that the dynwrappers are still the handpatched kind, although the latest version of dynload-wrapper, 0.4, should create working equivalents.

Regarding the dependencies, we pull in `wayland-client` (self-explanatory), `wayland-cursor` (cursor handling), `wayland-egl-core` (EGL support) and `libdecor-0` (client-side decorations)

If you're wondering about `libdecor-0`: Wayland allows compositors to not provide SSDs (and thus require CSDs). We don't have yet support for them so I implemented (optional) support for this library that draws CSDs for us. I think that we should eventually implement our own solution (as it's being slowly being done AFAIK) to reap all the benefits of them and, well, to get rid of a dependency and all its handling code :P

As you can see, a good part of the diff (+17048) is actually data or autogenerated stuff. There's not much review that has to be done, outside of a quick glance to verify those claims. I swear I didn't put any malware in there :P

Now we talk about the actually juicy stuff.

Wayland moves a lot of responsibility to the client. Because of that, the logic is actually split between a `DisplayServerWayland` (`platform/linuxbsd/wayland/display_server_wayland.*`, +1615) and a `WaylandThread` (`platform/linuxbsd/wayland/wayland_thread.*`, +4929).

As you can see, most of the Wayland logic is in the thread, which is also probably the hardest to review. It offers various methods (which _don't_ set the internal mutex manually, this has to be done by the caller) and an `InputEvent`-style queue, which gives various notifications to the main thread, like input and rect changes.

The advantage of this design is that we can keep the thing as responsive as possible and that it takes care to translate the information to something that the engine can understand in the same place that it's handled (e.g. HiDPI scaling, `InputEvent` generation). That also moves, well... The main Wayland logic, into one place.

You should be able to understand `DisplayServerWayland` pretty well without any knowledge about the Wayland library, perhaps with the exception of `window_can_draw`. In short, the compositor wants to tell us when to draw, so we do this dirty hack where we set it to `true` only when we get a frame event. This is good enough for now but it will be improved.

Of note there's `platform/linuxbsd/wayland/detect_prime_egl.*` (+296) which is mostly similar to the X11 counterpart, with the exception that it probes up to 4 devices, allowing a bit more of leeway for people with an IGPU, DGPU and a SW rasterizer, as sometimes the DGPU would not be probed at all. That code could be shared somehow but that's a topic for another time I think.

Other than that, there's some plumbing, project/editor setting handling, doc changes, driver implementations (Vulkan, EGL) and licensing information but it should be already familiar to most of us I hope.

That's it really. I hope that this helps with reviewing :)

_Production edit: closes godotengine/godot-roadmap#10_


platform/linuxbsd/wayland/display_server_wayland.h
```diff
@@ -0,0 +1,285 @@
...
+	TTS_Linux *tts = nullptr;
+#endif
+
+#if DBUS_ENABLED
+	FreeDesktopPortalDesktop *portal_desktop = nullptr;
+
+	FreeDesktopScreenSaver *screensaver = nullptr;
+	bool screensaver_inhibited = false;
+#endif
+	static String _get_app_id_from_context(Context context);
```
AThousandShips, line 139:
```suggestion
	static String _get_app_id_from_context(Context p_context);
```

platform/linuxbsd/wayland/display_server_wayland.h
```diff
@@ -0,0 +1,285 @@
...
+
+	void _send_window_event(WindowEvent p_event);
+
+	static void dispatch_input_events(const Ref<InputEvent> &p_event);
+	void _dispatch_input_event(const Ref<InputEvent> &p_event);
+
+	void _resize_window(Size2i size);
```
AThousandShips, line 146:
```suggestion
	void _resize_window(const Size2i &p_size);
```

platform/linuxbsd/wayland/display_server_wayland.h
```diff
@@ -0,0 +1,285 @@
...
+
+	virtual void clipboard_set(const String &p_text) override;
+	virtual String clipboard_get() const override;
+	virtual Ref<Image> clipboard_get_image() const override;
+	virtual void clipboard_set_primary(const String &p_text) override;
+	virtual String clipboard_get_primary() const override;
+
+	virtual int get_screen_count() const override;
+	virtual int get_primary_screen() const override;
+	virtual Point2i screen_get_position(int p_screen = SCREEN_OF_MAIN_WINDOW) const override;
```
AThousandShips, line 186:
```suggestion
	virtual Point2i screen_get_position(int p_screen) const override;
```
Default arguments on overrides are unnecessary (and not recommended AFAIK) and clutters here IMO

platform/linuxbsd/wayland/display_server_wayland.cpp
```diff
@@ -0,0 +1,1330 @@
...
+#ifdef VULKAN_ENABLED
+#include "servers/rendering/renderer_rd/renderer_compositor_rd.h"
+#endif
+
+#ifdef GLES3_ENABLED
+#include "detect_prime_egl.h"
+#include "drivers/gles3/rasterizer_gles3.h"
+#endif
+
+String DisplayServerWayland::_get_app_id_from_context(Context context) {
```
AThousandShips, line 51:
```suggestion
String DisplayServerWayland::_get_app_id_from_context(Context p_context) {
```

platform/linuxbsd/wayland/display_server_wayland.cpp
```diff
@@ -0,0 +1,1330 @@
...
+}
+
+void DisplayServerWayland::_dispatch_input_event(const Ref<InputEvent> &p_event) {
+	Callable callable = main_window.input_event_callback;
+	if (callable.is_valid()) {
+		callable.call(p_event);
+	}
+}
+
+void DisplayServerWayland::_resize_window(Size2i size) {
```
AThousandShips, line 97:
```suggestion
void DisplayServerWayland::_resize_window(const Size2i &p_size) {
```

platform/linuxbsd/wayland/key_mapping_xkb.h
```diff
@@ -0,0 +1,63 @@
...
+#ifdef SOWRAP_ENABLED
+#include "xkbcommon-so_wrap.h"
+#else
+#include <xkbcommon/xkbcommon.h>
+#endif // SOWRAP_ENABLED
+
+class KeyMappingXKB {
+	struct HashMapHasherKeys {
+		static _FORCE_INLINE_ uint32_t hash(const Key p_key) { return hash_fmix32(static_cast<uint32_t>(p_key)); }
+		static _FORCE_INLINE_ uint32_t hash(const unsigned p_key) { return hash_fmix32(p_key); }
```
AThousandShips, line 46:
```suggestion
		static _FORCE_INLINE_ uint32_t hash(Key p_key) { return hash_fmix32(static_cast<uint32_t>(p_key)); }
		static _FORCE_INLINE_ uint32_t hash(unsigned p_key) { return hash_fmix32(p_key); }
```
No `const` for simple types

platform/linuxbsd/wayland/wayland_thread.h
```diff
@@ -0,0 +1,930 @@
...
+
+	static void seat_state_update_cursor(SeatState *p_ss);
+
+	void seat_state_echo_keys(SeatState *p_ss);
+
+	static int window_state_get_preferred_buffer_scale(WindowState *p_ws);
+	static double window_state_get_scale_factor(WindowState *p_ws);
+	static void window_state_update_size(WindowState *p_ws, int p_width, int p_height);
+
+	static Vector2i scale_vector2i(Vector2i p_vector, double p_amount);
```
AThousandShips, line 855:
```suggestion
	static Vector2i scale_vector2i(const Vector2i &p_vector, double p_amount);
```

platform/linuxbsd/wayland/wayland_thread.h
```diff
@@ -0,0 +1,930 @@
...
+	void push_message(Ref<Message> message);
+	bool has_message();
+	Ref<Message> pop_message();
+
+	void window_create(DisplayServer::WindowID p_window_id, int p_width, int p_height);
+
+	struct wl_surface *window_get_wl_surface(DisplayServer::WindowID p_window_id) const;
+
+	void window_set_max_size(DisplayServer::WindowID p_window_id, Size2i p_size);
+	void window_set_min_size(DisplayServer::WindowID p_window_id, Size2i p_size);
```
AThousandShips, line 866:
```suggestion
	void window_set_max_size(DisplayServer::WindowID p_window_id, const Size2i &p_size);
	void window_set_min_size(DisplayServer::WindowID p_window_id, const Size2i &p_size);
```

platform/linuxbsd/wayland/wayland_thread.h
```diff
@@ -0,0 +1,930 @@
...
+
+	bool window_can_set_mode(DisplayServer::WindowID p_window_id, DisplayServer::WindowMode p_window_mode) const;
+	void window_try_set_mode(DisplayServer::WindowID p_window_id, DisplayServer::WindowMode p_window_mode);
+	DisplayServer::WindowMode window_get_mode(DisplayServer::WindowID p_window_id) const;
+
+	void window_set_borderless(DisplayServer::WindowID p_window_id, bool p_borderless);
+	void window_set_title(DisplayServer::WindowID p_window_id, String p_title);
+	void window_set_app_id(DisplayServer::WindowID p_window_id, String p_app_id);
```
AThousandShips, line 874:
```suggestion
	void window_set_title(DisplayServer::WindowID p_window_id, const String &p_title);
	void window_set_app_id(DisplayServer::WindowID p_window_id, const String &p_app_id);
```

platform/linuxbsd/wayland/wayland_thread.h
```diff
@@ -0,0 +1,930 @@
...
+
+	// Optional - require idle_inhibit_unstable_v1
+	void window_set_idle_inhibition(DisplayServer::WindowID p_window_id, bool p_enable);
+	bool window_get_idle_inhibition(DisplayServer::WindowID p_window_id) const;
+
+	ScreenData screen_get_data(int p_screen) const;
+	int get_screen_count() const;
+
+	void pointer_set_constraint(PointerConstraint p_constraint);
+	void pointer_set_hint(Point2i p_hint);
```
AThousandShips, line 889:
```suggestion
	void pointer_set_hint(const Point2i &p_hint);
```

platform/linuxbsd/wayland/wayland_thread.h
```diff
@@ -0,0 +1,930 @@
...
+	PointerConstraint pointer_get_constraint() const;
+	DisplayServer::WindowID pointer_get_pointed_window_id() const;
+	BitField<MouseButtonMask> pointer_get_button_mask() const;
+
+	void cursor_hide();
+	void cursor_set_shape(DisplayServer::CursorShape p_cursor_shape);
+
+	void cursor_set_custom_shape(DisplayServer::CursorShape p_cursor_shape);
+	void cursor_shape_set_custom_image(DisplayServer::CursorShape p_cursor_shape, Ref<Image> p_image, Point2i p_hotspot);
```
AThousandShips, line 898:
```suggestion
	void cursor_shape_set_custom_image(DisplayServer::CursorShape p_cursor_shape, Ref<Image> p_image, const Point2i &p_hotspot);
```

platform/linuxbsd/wayland/wayland_thread.h
```diff
@@ -0,0 +1,930 @@
...
+	bool selection_has_mime(String p_mime) const;
+	Vector<uint8_t> selection_get_mime(String p_mime) const;
+
+	void selection_set_text(String p_text);
+
+	// Optional primary support - requires wp_primary_selection_unstable_v1
+	bool primary_has_mime(String p_mime) const;
+	Vector<uint8_t> primary_get_mime(String p_mime) const;
+
+	void primary_set_text(String p_text);
```
AThousandShips, line 919:
```suggestion
	bool selection_has_mime(const String &p_mime) const;
	Vector<uint8_t> selection_get_mime(const String &p_mime) const;

	void selection_set_text(const String &p_text);

	// Optional primary support - requires wp_primary_selection_unstable_v1
	bool primary_has_mime(const String &p_mime) const;
	Vector<uint8_t> primary_get_mime(const String &p_mime) const;

	void primary_set_text(const String &p_text);
```

platform/linuxbsd/wayland/display_server_wayland.h
AThousandShips: That's very strange, gotcha

AThousandShips: Looked further into it and not a weirdness just got confused about overrides and calling them

platform/linuxbsd/wayland/display_server_wayland.cpp
```diff
@@ -0,0 +1,1330 @@
...
+			ERR_FAIL_COND_MSG(err == ERR_CANT_CREATE, "Can't show a Vulkan window.");
+		}
+#endif
+
+#ifdef GLES3_ENABLED
+		if (egl_manager) {
+			struct wl_surface *wl_surface = wayland_thread.window_get_wl_surface(wd.id);
+			wd.wl_egl_window = wl_egl_window_create(wl_surface, wd.rect.size.width, wd.rect.size.height);
+
+			Error err = egl_manager->window_create(MAIN_WINDOW_ID, wayland_thread.get_wl_display(), wd.wl_egl_window, wd.rect.size.width, wd.rect.size.height);
```
v01dXYZ, line 155: It seems `EGLManager::window_create` expects a `display*` and a `window*` as the names of the variables suggest:

https://github.com/godotengine/godot/blob/3ce73e5d419011d1642ed30a3417362b82c41898/drivers/egl/egl_manager.cpp#L200

v01dXYZ: The manager doesn't send the values directly (for EGL != 1.5). It dereferences them before: 

https://github.com/godotengine/godot/blob/3ce73e5d419011d1642ed30a3417362b82c41898/drivers/egl/egl_manager.cpp#L61

https://github.com/godotengine/godot/blob/3ce73e5d419011d1642ed30a3417362b82c41898/drivers/egl/egl_manager.cpp#L219

Maybe that's not a bug related to your MR.

ansipunk:
Hello Riteo! Thank you for your incredible contributions.
I can't help but be curious, have you tested if Wayland support fixes black window flickering on high refresh screens on KDE with VRR enabled?

vmedea:
> Should I keep the protocol definitions vendored or should I depend on wayland-protocols?

I guess using the installed wayland-protocols gives a better idea of what wayland protocols are actually used on the computer it's built on.

One drawback would be that you'd have to take into account different versions of the protocol files that may be installed on the devr's computer (which can cover a very wide range). We've seen this with the `wl_pointer.axis_v120` event. Some experimental ones like the tablet support may be missing completely.

There may be reasons to deviate from it here specifically but in general I like the self-containedness of godot's build system, it's very convenient. Also, the computer it's built on is not the one it's going to be run on if you're distributing a game, it's not the same as usual FOSS assumptions.

geowarin:
@ansipunk Coul this be [this bug](https://gitlab.freedesktop.org/drm/amd/-/issues/2967), reported in upstream KDE.

Nobara recommends the [following workaround](https://feed.nobaraproject.org/en/flicker-issues-with-kde-wayland-at-frame-rates-100-fps).

ansipunk:
@geowarin

This problem, from my experience, is exclusive to Godot — all the other programs and games work just fine.  However, it also appears on Windows as well. From what I understand, it will happen on any platform with any configuration as long as refresh rate is higher than 60.

I asked my question because Wayland, unlike X, supports VRR and KDE is the first and only DE I know of that supports VRR, so I imagined adding Wayland support might fix this problem in serendipity.

There are many issues in Godot project that address the flickering problem like #77337 and the only workaround for it, as of now, is to turn VRR off.

Calinou:
> I asked my question because Wayland, unlike X, supports VRR and KDE is the first and only DE I know of that supports VRR, so I imagined adding Wayland support might fix this problem in serendipity.

X11 has VRR support in single-monitor configurations – I'm using one right now on KDE + NVIDIA.

iMonZ:
Finally Godot becomes usable…
Can’t wait to try Godot the first time with 4.3
I have actually waited for 4 years already…

akien-mga:
@iMonZ You can communicate your support by giving the PR a thumbsup, but please refrain from commenting if you're not participating in reviewing the PR.

akien-mga:
Needs rebase to build after the RenderingDeviceDriver rework.
```
platform/linuxbsd/wayland/display_server_wayland.h:41:10: fatal error: drivers/vulkan/rendering_device_vulkan.h: No such file or directory
   41 | #include "drivers/vulkan/rendering_device_vulkan.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

adamscott (changes requested):
## Must have before merge
- [ ] Default to `xwayland` first.
- [ ] Add an editor setting option to select which backend to use on Wayland (`native` or `xwayland`).
- [ ] Add a command line parameter to force select a Wayland backend (`native` or `xwayland`).

### Default to `xwayland` first
I think that, considering some shortcomings with a native Wayland implementation, users updating to 4.3 that use Wayland and that don't care about the backend should not see their experience change.

Decodetalkers:
> > Add a command line parameter to force select a Wayland backend (native or xwayland).
> 
> BTW, regarding this one, it's already there. It's just the default `--display-server` CLI argument.

Can there be an environment variable? Like Firefox use moz_wayland to enable Wayland support

TuxTheAstronaut:
@Riteo thats good to hear, one by concerns was that after this was merged users would update and wonder why their shaders broke but since xwayland will default most probably wont notice. 

However for people who are testing this PR if you haven't also added the fix in yourself to fix the shader bug then a quick fix is to instead cap the frame rate at your monitors refresh rate instead of using vsync to fix shaders. tho note setting the fps cap much higerer than the refresh rate will cause shaders to act up again at least from my testing.

platform/linuxbsd/wayland/display_server_wayland.cpp
```diff
@@ -0,0 +1,1353 @@
...
+
+Vector<String> DisplayServerWayland::get_rendering_drivers_func() {
+	Vector<String> drivers;
+
+#ifdef VULKAN_ENABLED
+	drivers.push_back("vulkan");
+#endif
+
+#ifdef GLES3_ENABLED
+	drivers.push_back("opengl3");
```
vmedea, line 1139: not a big deal i think but FWIW, `display_server_x11.cpp` calls this same driver `opengl3_es`, `opengl3` is the desktop GL

vmedea:
When launching godot editor using `--display-driver wayland`, any projects will still launch in X11 by default. Not sure this is intentiona but it was slightly confusing to me. Changing the project setting worked.

platform/linuxbsd/wayland/display_server_wayland.cpp
vmedea: Oh, thought wayland used EGL+GLES, that's fine then!

TuxTheAstronaut:
testing the new editor setting it seems to break the `--display-driver` command line argument. if the setting is set to prefer wayland then `--display-driver x11` will still use wayland instead.

TuxTheAstronaut:
@Riteo its working now 👍

modules/openxr/extensions/platform/openxr_opengl_extension.cpp
technobaboo, line 1: Consider replacing the X11 OpenGL graphics bindings with XrGraphicsBindingEGLMNDX for Linux (maybe in another PR), as both Monado and SteamVR, X11 and Wayland support them + it does not require a desktop window being open.

I'd be willing to help with this, as in my experience the X11/Wayland-specific graphics bindings are more trouble than they're worth and I fought to get the EGL extension supported everywhere.

technobaboo (changes requested):

modules/openxr/extensions/platform/openxr_opengl_extension.cpp
technobaboo: though if better, I could also just add wayland graphics bindings support (but why not unify them?)

technobaboo: the reason I suggested delaying the PR was just to fix wayland, whether that's via the wayland or egl graphics bindings.

technobaboo: i asked in the XR chat what they'd allow in terms of graphics bindings (because i can fix it up once i know that) and hopefully this won't be delayed for long!

btw i suggested delaying this PR because you mentioned it wouldn't even compile with openxr enabled, is that still true or will it just error?

technobaboo: https://github.com/Riteo/godot/pull/4

fixed it, worth double-checking it's up to par with the other code but functionally it works great

technobaboo: we agreed this is better done in a separate PR with different code

Domeee:
Running the editor with `run/platforms/linuxbsd/prefer_wayland` set to `true` works. Running the project from this same editor instance launches the game in X11. See the outputs of `swaymsg -t get_tree`:

```
#25: output "DP-2"
    #17: workspace "8"
      #19: con "prototype_03.tscn - awakening - Godot Engine" (xdg_shell, pid: 4072, app_id: "org.godotengine.Editor")
      #36: con "awakening (DEBUG)" (xwayland, pid: 16610, instance: "Godot_Engine", class: "awakening", X11 window: 0x800003)
```

I expect the game to run with Wayland too.

Running the game directly via command line, e.g. `/home/dome/code/godot/bin/godot.linuxbsd.editor.x86_64 --path /home/dome/code/awakening --display-driver wayland` runs the game in Wayland. Running the editor (`--editor`) via command line and then run the game from within this editor instance launches the game in X11 again.

Domeee:
In my opinion the settings from `run/platforms/linuxbsd/prefer_wayland` should be carried over to a project that is run from within the editor (not exported). When developing inside the editor I regard the running game as an extension of the editor and therefor don't expect other settings regarding the display server. This is my personal expected behavior.

Regarding exported projects I would expect an option in the projects settings, e.g. `application/run/linuxbsd/prefer_wayland`.

Domeee:
Sorry, I was not aware of the new project property section "Display Server". Still, I think for a game launched in the editor the editor settings should be carried over. From a users (game dev) point of view, working in the editor, which includes running games via the editor, should be configured via editor settings. This allows me to set the display server for the export in the project settings (an probably leaving it be for the lifetime of the project) while changing the editor settings according to my needs as developer (which may even differ across dev envs [in my case wayland on my desktop, while i run x11 on my notebook]). The current implementation may lead to a lot of builds with the wrong display server configured (just because the dev used it on his machine and forget to reset the setting).

If you don't agree, I wish the description in the editor settings will be extended with an info (e.g. set display server for the player in project settings). Thanks for your consideration! Really looking forward for wayland support to land in the next release 😍

Domeee:
The editor settings should **always** be carried over to the game when launched via editor. The editor settings should not affect the exported game. The project setting should only affect the exported game. This is how I as a Godot user would expect things to work. But I am not the reference ;-)

In the current state I consider this a minor inconvenience, which can be resolved by extending the description in the editor settings to make things clear. This way there are no more larger changes and the PR gets merged faster. This way it's also possible to consult with the community regarding the project vs. editor settings to see if my expectation matches with the community :-D

akien-mga (approved):
Did a code review on GitHub, this is excellent work! :tada: :1st_place_medal: 

I admit I didn't review WaylandThread in depth yet, but the overall architecture seems very solid and I didn't spot anything out of place in the areas I'm familiar with.

I tested briefly on Linux X11 and it works as expected ;)

I'll aim to test on Wayland in coming days, just need to setup a proper Wayland environment and get familiar with whether quirks are from this PR or from Wayland itself :)

Note this recently merged PR which impacts the keymapping code, should likely be reflected here: #80231.

AlexRamallo:
For what it's worth, I just quickly tested this on my system (Fedora Kinoite 39 + RTX 4090 + Wayland) and it seems to be working very well. Godot's editor has always been borderline unusable for me without X11 because of tons of display and input bugs, but none of those are present in this branch. Good job! :D

akien-mga:
I've been testing on KDE on Wayland (Plasma 5.27.5), and it seems to work well for the most part.
Editing a project with Wayland from the command line with `godot-git --display-driver wayland -e --path /home/akien/Projects/godot/projects/godot-4-3d-third-person-controller/ ` works fine.

As mentioned above, the parameter isn't passed further to instances created by the editor when running the project, but that's currently the way Godot works for most things. So while a subpar experience, it's not related to this PR and not something to change here.

---

A few times I was confused as to whether I was actually running on Wayland or X11. I had to use `xprop` to figure it out.
I would suggest looking into whether we could print the display server in the output, maybe on the same line as the rendering info, which currently is like:
```
OpenGL API 4.6 (Core Profile) Mesa 23.3.2 - Compatibility - Using Device: Intel - Mesa Intel(R) HD Graphics 630 (KBL GT2)
```

It could maybe become:
```
Linux (Wayland) - OpenGL API 4.6 (Core Profile) Mesa 23.3.2 - Compatibility - Using Device: Intel - Mesa Intel(R) HD Graphics 630 (KBL GT2)
```
(And likewise for other OSes, leaving out the display driver if it's the same as the OS name maybe so we don't get "Windows (Windows)", but still get "Windows (Headless)".)

Either way, that's also something for a follow-up PR IMO, doesn't need to be here.

---

One significant issue I ran into is that the project manager seems to crash my Mesa drivers when running with Wayland. Strangely enough, running a GL Compatibility project from the command line with `--display-driver wayland` does *not* crash the same way. So there's something specific to the PM. Maybe related to low processor usage mode?

It crashes on my AMD discrete GPU (Radeon RX Vega M) when auto-detected by `prime_detect` or selected with `DRI_PRIME=1`, and doesn't crash with the integrated Intel HD Graphics 630 (when forced with `DRI_PRIME=0`).

Here's the stacktrace from `DRI_PRIME=1 godot-git --display-driver wayland`:
```
OpenGL API 4.6 (Core Profile) Mesa 23.3.2 - Compatibility - Using Device: AMD - AMD Radeon RX Vega M GL Graphics (radeonsi, vegam, LLVM 15.0.6, DRM 3.54, 6.5.13-desktop-6.mga9)
(gdb) bt
#0  0x00007ffff7518580 in wl_proxy_get_version () from /lib64/libwayland-client.so.0
#1  0x00007ffff00aba8d in create_wl_buffer () from /lib64/libEGL_mesa.so.0
#2  0x00007ffff00ac1e8 in dri2_wl_swap_buffers_with_damage () from /lib64/libEGL_mesa.so.0
#3  0x00007ffff00a13c3 in dri2_swap_buffers () from /lib64/libEGL_mesa.so.0
#4  0x00007ffff0095665 in eglSwapBuffers () from /lib64/libEGL_mesa.so.0
#5  0x0000000006ed0714 in EGLManager::swap_buffers (this=0xb2c1120) at drivers/egl/egl_manager.cpp:290
#6  0x0000000005868d13 in DisplayServerWayland::swap_buffers (this=0xb2a9840) at platform/linuxbsd/wayland/display_server_wayland.cpp:1115
#7  0x0000000006e25812 in RasterizerGLES3::end_viewport (this=0xb39ef80, p_swap_buffers=true) at drivers/gles3/rasterizer_gles3.cpp:112
#8  0x0000000006e273e0 in RasterizerGLES3::set_boot_image (this=0xb39ef80, p_image=..., p_color=..., p_scale=false, p_use_filter=true) at drivers/gles3/rasterizer_gles3.cpp:475
#9  0x00000000093c63b7 in RenderingServerDefault::set_boot_image (this=0xb4ff7e0, p_image=..., p_color=..., p_scale=false, p_use_filter=true) at ./servers/rendering/rendering_server_default.cpp:279
#10 0x00000000058add13 in Main::setup2 () at main/main.cpp:2732
#11 0x00000000058aac0c in Main::setup (execpath=0x7fffffffd9e4 "/home/akien/.local/bin/godot-git", argc=2, argv=0x7fffffffd510, p_second_phase=true) at main/main.cpp:2251
#12 0x00000000057fb7e5 in main (argc=3, argv=0x7fffffffd508) at platform/linuxbsd/godot_linuxbsd.cpp:62
```

---

Related to the above, I get a bunch of warnings when running without `DRI_PRIME`, i.e. letting our `prime_detect` figure it out:
```
$ godot-git --display-driver wayland
Godot Engine v4.3.dev.custom_build.7790b2be9 - https://godotengine.org
libdecor-0.so.0: cannot open shared object file: No such file or directory
Invalid value (0) for DRI_PRIME. Should be > 0
Inconsistent value (2) for DRI_PRIME. Should be < 2 (GPU devices count). Using: 1
Inconsistent value (3) for DRI_PRIME. Should be < 2 (GPU devices count). Using: 1
Found discrete GPU, setting DRI_PRIME=1 to use it.
Note: Set DRI_PRIME=0 in the environment to disable Godot from using the discrete GPU.
OpenGL API 4.6 (Core Profile) Mesa 23.3.2 - Compatibility - Using Device: AMD - AMD Radeon RX Vega M GL Graphics (radeonsi, vegam, LLVM 15.0.6, DRM 3.54, 6.5.13-desktop-6.mga9)
```
(and then the crash as described above)

---

My Linux distro (Mageia) doesn't seem to provide libdecor-0 at all :exploding_head: 
I'll have to package it. But we should assess whether it's widely available on more mainstream distros (I expect it is and Mageia is just lagging behind in its Wayland integration).

main/main.cpp
```diff
@@ -2457,9 +2469,35 @@ Error Main::setup2() {
 	/* Initialize Display Server */
 
 	{
+		if (display_driver.is_empty()) {
+			display_driver = GLOBAL_GET("display/display_server/driver");
+		}
+
```
YuriSizov, line 2475: As a codestyle nitpick, benchmarking calls must be wrapping the scope, so this addition needs to be moved below `OS::get_singleton()->benchmark_begin_measure("Servers", "Display");`

akien-mga:
> _Edit:_ Actually even `DRI_PRIME=0` raises a warning it seems, despite us suggesting it as an option to select the integrated GPU. It seems like it should be `DRI_PRIME=` without `0`.
> The warning `Inconsistent value (2) for DRI_PRIME. Should be < 2 (GPU devices count). Using: 1` does sound like it could support passing numbers above 2 if you have more than 2 GPUs so maybe your interpretation is correct? But at least on my system, a software renderer isn't registered as a GPU.

So I guess let's keep it as is for now and see how to improve the logic later.

adamscott (approved):
Congrats, @Riteo for your hard and long commited work!

main/main.cpp
```diff
@@ -2382,7 +2371,26 @@ Error Main::setup2() {
...
+						// Skip.
+						screen_found = true;
+					}
+
+					if (!display_driver.is_empty()) {
+						// Skip.
+						prefer_wayland_found = true;
+					}
+
+					while (!screen_found || !prefer_wayland_found) {
```
YuriSizov, line 2393: Shouldn't this be `&&`? I assume we want to find both properties.

YuriSizov: Ah wait, that's correct.

YuriSizov: Yes, the negatives tricked me.

YuriSizov (approved):
Changes outside of the Linux platform code seem low risk, so this should be good to merge to facilitate further testing.

akien-mga:
A huge thankyou and congratulations for the amazing work done here @Riteo, and every one involved testing, proposing changes, and reviewing the code. :tada: :1st_place_medal: 

This is a major update for Linux users and an impressive achievement!

And just the start of the journey... now let's polish, add missing features, and see how to make good use of Wayland to improve things ever further!

oschl-git:
This might slip by unnoticed by most people in the community, but it's a really huge deal, especially for the future. Congratulations and thanks so much for working on this, @Riteo.

agentx3:
This is fantastic! I recently got into Godot, and was experiencing frustrations with its xwayland buggy-ness. What a fortunate and blessed contribution to the FOSS community.

platform/linuxbsd/detect.py
```diff
@@ -47,6 +47,8 @@ def get_opts():
         BoolVariable("fontconfig", "Use fontconfig for system fonts support", True),
         BoolVariable("udev", "Use udev for gamepad connection callbacks", True),
         BoolVariable("x11", "Enable X11 display", True),
+        BoolVariable("wayland", "Enable Wayland display", True),
```
Rindbee, line 50: These two are conflicting options, so leave one as `True`.

sim1cent:
Great work for the wayland backend, though i think the availability of a SDL backend is necessary for true linux long term compatibility, here's a old comment i posted on a related merge request.

_Many commercial engines use SDL, though statically linked (it has a neat mechanism to override itself even when statically linked), only on Linux, even if it might be used as a cross-platform abstraction.
This proves is that the de-facto windowing API for Linux is SDL, they are the only ones that seem committed to backward compatibility, and it's proven judging from their effort of maintaining a stable ABI and things like sdl12-compat (and soon sdl2-compat).
I can run 2003 builds of Linux games using modern display and audio protocols thanks to it.
It's not that Godot developers aren't good enough to make a decent Linux port, it's that SDL is the necessary overlay for long term compatibility.
This is especially critical right now when this engine is being adopted more and more, and game developers don't certainly have to maintain the Linux builds for the years to come._

Enderteck:
Is Wayland support now pretty much ready, or are there important issues/missing features ?

PR #86189, opened 2023-12-15
Fix crash when renaming input maps in single window mode

### LINKED ISSUE #86165: 4.2 crashes when renaming input maps in single window mode
### Tested versions

- Reproducible in 4.0+
- Non reproducible in older versions

### System information

Godot v4.2.1.stable - Windows 10.0.19045 - Vulkan (Mobile) - dedicated AMD Radeon RX 580 2048SP (Advanced Micro Devices, Inc.; 31.0.21905.1001) - AMD Ryzen 5 2400G with Radeon Vega Graphics (8 Threads)

### Issue description

In the input map menu, if you rename an action, and instead of pressing enter, you click on another action, the engine crashes. This only happens if you have Single Window Mode enabled, if you click anywhere else other than an action, it works, and pressing enter obviously works as intended. I did some quick tests and this problem only seems to happen in 4.x versions.

### Steps to reproduce

1. Enable Single Window Mode (Editor Settings: interface/editor/single_window_mode) and restart the editor
2. Open the Project Settings and go to the Input Map tab
3. Create 2 actions
4. Click on one of them and change it's name
5. Click on the other one

[crashing_issue.zip](https://github.com/godotengine/godot/files/13676443/crashing_issue.zip)

jsjtxietian:
Fixes #86165

It already uses `call_deferred` in another branch, so IMO it's appropriate to add it here.

### DIFF
--- a/editor/action_map_editor.cpp
+++ b/editor/action_map_editor.cpp
@@ -137,7 +137,7 @@ void ActionMapEditor::_action_edited() {
 			return;
 		}
 
-		emit_signal(SNAME("action_renamed"), old_name, new_name);
+		call_deferred(SNAME("emit_signal"), "action_renamed", old_name, new_name);
 	} else if (action_tree->get_selected_column() == 1) {
 		// Deadzone Edited
 		String name = ti->get_meta("__name");


KoBeWi:
Superseded by #89346
Although that PR made the other deferred call irrelevant, so you could change it to regular call. But that's no longer a bugfix.

jsjtxietian:
Sounds like I should close this one since it's no longer crash.

PR #81220, opened 2023-09-01
Support loading more DDS formats

BlueCube3310:
Depends on #80900

Adds support for loading the following DDS formats:
- RGB10A2,
- BGRA4,
- BC6U / BPTC_UF,
- BC6S / BPTC_SF,
- BC7 / BPTC_RGBA,
- R16F,
- RG16F,
- RGBA16F,
- R32F,
- RG32F,
- RGBA32F
- RGB9E5

An MRP containing DDS files encoded in these formats: [DDSFormatsExtended.zip](https://github.com/godotengine/godot/files/12496516/DDSFormatsExtended.zip)
Godot 4.2-dev3 is unable to load any of them correctly.

PR #86204, opened 2023-12-15
Add support for loading less common DDS formats

BlueCube3310:
Follow-up to #81220

Implements support for loading some less common DDS formats, such as:
- A8,
- DXT2,
- DXT4,
- B2GR3,
- B2GR3A8,
- LA4,
- RGB32F,
- SRGB variants of RGBA8, BC1, BC2, BC3 and BC7

~~Also changes the loader to now require the `DDPF_LUMINANCE` flag for grayscale formats, though there is a chance this may break files created by writers which don't set it.~~

~~I've marked it as draft in case there are more unsupported formats.~~

MRP for testing the changes: [DDSFormatsUncommon.zip](https://github.com/user-attachments/files/15858140/DDSFormatsUncommon.zip)

### DIFF
--- a/modules/dds/texture_loader_dds.cpp
+++ b/modules/dds/texture_loader_dds.cpp
@@ -42,14 +42,18 @@ enum {
 	DDSD_PITCH = 0x00000008,
 	DDSD_LINEARSIZE = 0x00080000,
 	DDSD_MIPMAPCOUNT = 0x00020000,
-	DDPF_FOURCC = 0x00000004,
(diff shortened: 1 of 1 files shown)


fire (approved):
If we say we support DDS, it makes sense that we should support all dds formats.

I don't have production samples to test. There is a test sample. Did not technical review.

clayjohn (approved):
Looks fine to me. I didn't test locally, but I trust that bluecube3310 has done a good job testing

akien-mga:
Thanks!

PR #86208, opened 2023-12-15
Improve visualization of WorldBoundary2D editor

### LINKED ISSUE #817: Improve editor gizmos for WorldBoundaryShape2D
image

From my experience being new to this component it does feel very unintuitive to use the handles. First of all for all other shapes the blue area represents the collision area so this make it look like the collision is a T shape which I guess it's not. Secondly you can move the line away from the origin and rotate it but there seams to be no way to change the length of it as the handles normalize the normal resulting in you having to use the inspector to change the line length. So what do I propose?

1. Use something else than a blue area for the normal to make it clear what's part of the collision shape and what's not.
2. Make it possible to use the rotation dot to also change the length of the line.

Let me know your thoughts! And depending on what you say I'll send a pr.

MarcusElg:
Fixes https://github.com/godotengine/godot-proposals/issues/817

Changes the worldboundary 2d shape editor to be more clear. Changes the normal to not be the same colour as the line as it isn't actually part of the collision shape. Also changes the visualisation of the line to be a gradient to visualise that it's an infinite line.

**Old**
image

**New**
image

Let me know what you think :smile:

### DIFF
--- a/scene/resources/2d/world_boundary_shape_2d.cpp
+++ b/scene/resources/2d/world_boundary_shape_2d.cpp
@@ -86,10 +86,41 @@ void WorldBoundaryShape2D::draw(const RID &p_to_rid, const Color &p_color) {
 	Vector2 point = distance * normal;
 	real_t line_width = 3.0;
 
-	Vector2 l1[2] = { point - normal.orthogonal() * 100, point + normal.orthogonal() * 100 };
-	RS::get_singleton()->canvas_item_add_line(p_to_rid, l1[0], l1[1], p_color, line_width);
-	Vector2 l2[2] = { point + normal.normalized() * (0.5 * line_width), point + normal * 30 };
-	RS::get_singleton()->canvas_item_add_line(p_to_rid, l2[0], l2[1], p_color, line_width);
+	// Draw collision shape line.
+	PackedVector2Array line_points = {
+		point - normal.orthogonal() * 100,
+		point - normal.orthogonal() * 60,
+		point + normal.orthogonal() * 60,
+		point + normal.orthogonal() * 100
+	};
+
+	Color transparent_color = Color(p_color, 0);
+	PackedColorArray line_colors = {
+		transparent_color,
+		p_color,
+		p_color,
+		transparent_color
+	};
+
+	RS::get_singleton()->canvas_item_add_polyline(p_to_rid, line_points, line_colors, line_width);
+
+	// Draw arrow.
+	Color arrow_color = p_color.inverted();
+
+	Transform2D xf;
+	xf.rotate(normal.angle());
+
+	Vector<Vector2> arrow_points = {
+		xf.xform(Vector2(distance + line_width / 2, -2.5)),
+		xf.xform(Vector2(distance + 20, -2.5)),
+		xf.xform(Vector2(distance + 20, -10)),
+		xf.xform(Vector2(distance + 40, 0)),
+		xf.xform(Vector2(distance + 20, 10)),
+		xf.xform(Vector2(distance + 20, 2.5)),
+		xf.xform(Vector2(distance + line_width / 2, 2.5)),
+	};
+
+	RS::get_singleton()->canvas_item_add_polyline(p_to_rid, arrow_points, { arrow_color }, line_width / 2);
 }
 
 Rect2 WorldBoundaryShape2D::get_rect() const {


scene/resources/world_boundary_shape_2d.cpp
```diff
@@ -86,10 +87,27 @@ void WorldBoundaryShape2D::draw(const RID &p_to_rid, const Color &p_color) {
...
+		p_color,
+		p_color,
+		transparent_colour
+	};
+
+	RS::get_singleton()->canvas_item_add_polyline(p_to_rid, points, colours, line_width);
+
+	// Draw normal line.
+	Vector2 l2[2] = { Vector2(0, 0), point + normal * 30 };
+	RS::get_singleton()->canvas_item_add_line(p_to_rid, l2[0], l2[1], Color::named("BLACK"), 1);
```
AThousandShips, line 110:
```suggestion
	RS::get_singleton()->canvas_item_add_line(p_to_rid, l2[0], l2[1], Color(), 1);
```
Equivalent, and using named colour access is not recommended for performance unless necessary

scene/resources/world_boundary_shape_2d.cpp
```diff
@@ -86,10 +87,27 @@ void WorldBoundaryShape2D::draw(const RID &p_to_rid, const Color &p_color) {
...
-	RS::get_singleton()->canvas_item_add_line(p_to_rid, l2[0], l2[1], p_color, line_width);
+	// Draw collision shape line.
+	PackedVector2Array points = {
+		point - normal.orthogonal() * 100,
+		point - normal.orthogonal() * 60,
+		point + normal.orthogonal() * 60,
+		point + normal.orthogonal() * 100
+	};
+
+	Color transparent_colour = Color(p_color.r, p_color.g, p_color.b, 0);
```
AThousandShips, line 98:
```suggestion
	Color transparent_colour = Color(p_color, 0);
```

scene/resources/world_boundary_shape_2d.cpp
```diff
@@ -86,10 +87,27 @@ void WorldBoundaryShape2D::draw(const RID &p_to_rid, const Color &p_color) {
...
+	PackedColorArray colours = {
```
AThousandShips, line 99:
```suggestion
	PackedColorArray colors = {
```

scene/resources/world_boundary_shape_2d.cpp
```diff
@@ -86,10 +87,27 @@ void WorldBoundaryShape2D::draw(const RID &p_to_rid, const Color &p_color) {
...
+		transparent_colour,
+		p_color,
+		p_color,
+		transparent_colour
+	};
+
+	RS::get_singleton()->canvas_item_add_polyline(p_to_rid, points, colours, line_width);
```
AThousandShips, line 106:
```suggestion
	RS::get_singleton()->canvas_item_add_polyline(p_to_rid, points, colors, line_width);
```

scene/resources/world_boundary_shape_2d.cpp
```diff
@@ -30,6 +30,7 @@
 
 #include "world_boundary_shape_2d.h"
 
+#include "core/math/color.h"
```
AThousandShips, line 33:
```suggestion
```
Unnecessary

MewPurPur:
I like the gradient, but we definitely want to keep the handle and in my opinion also make it an arrow. WorldBoundaryShape2D

MarcusElg:
> I like the gradient, but we definitely want to keep the handle with the normal debug colors and in my opinion also make it an arrow. WorldBoundaryShape2D

Which are the normal debug colours?

MewPurPur:
You know, the blue translucent.

MarcusElg:
> You know, the blue translucent.

image
Like this more? Still feel like it would be nice to have some sort of difference between the arrow and the actual collision shape. Looking at the other collisionshapes it's only the ray seperation shape that has extra visual details, all other shapes only show the actual collision shape.

KoBeWi:
Maybe it can use the same color as one-way arrow?
image

MarcusElg:
> Maybe it can use the same color as one-way arrow? image

What node is that?

KoBeWi:
CollisionShape2D with `one_way_collision` enabled.

MarcusElg:
> Maybe it can use the same color as one-way arrow? image

image
It does make sense as the worldboundaryshape always is one sided

KoBeWi:
Ok so the remaining issue would be that you can enable one-way on boundary shape, ending with two ambiguous arrows 🤔
image
I'm not sure how to solve this one, other than maybe making the boundary arrow ever more prominent or using a different arrow shape.

MarcusElg:
How would I move forward with this?

KoBeWi:
You can join the [developer chat](https://chat.godotengine.org/) to ask for feedback and discuss.

MarcusElg:
@KoBeWi what do you think of this?

image

I feel like it works quite well

MewPurPur:
I like this.

KoBeWi (approved):
The arrow could be bigger, but otherwise looks fine IMO.

MarcusElg:
> The arrow could be bigger, but otherwise looks fine IMO.

Increased the size, this is at 100% zoom:
image

MewPurPur:
Wobblyness is likely you being project with some sort of pixel snap? Pixel snap also affects debug drawing in the editor.

MarcusElg:
@Calinou what do you think of the current version of this pr? Would be nice to get it merged soon.

akien-mga:
Thanks!

PR #49524, opened 2021-06-12
Document editor import options in the class reference

Calinou:
Tooltips are displayed when hovering import options, both in the Import dock and in the import defaults editor (which is in the Project Settings).

This closes https://github.com/godotengine/godot-proposals/issues/2041. See also https://github.com/godotengine/godot/pull/48548 and #49525 (can be merged independently).

## Preview

*Disregard the black text – it's not a bug related to this PR.*

### Import dock

Import dock

### Import defaults editor

Import defaults editor

## TODO

- [x] Only one importer is present in the documentation right now. Add the rest of the importers.
- [ ] Write documentation for all importers.

PR #86209, opened 2023-12-15
Correctly register editor-only module classes with the API

### LINKED ISSUE #86206: Some GDExtension methods are unavailable in Release builds
### Tested versions

4.2.0-stable
4.2.1-stable

### System information

Windows 10

### Issue description

Godot release builds (aka export templates) do not provide the `ResourceImporterOggVorbis` methods `load_from_buffer()` and `load_from_file()`. There might be more (now or in the future).

In godot-rust, we noticed it because we check for method availability at startup time. Downstream issue: https://github.com/godot-rust/gdext/issues/489.
**However, this affects all bindings**, including godot-cpp. If someone tries to call the methods, it will fail for exported games. What makes this delicate is that it works in Debug mode but breaks in Release, so it can go undetected.

I think it's OK that not all methods are available, but this information should be exposed via `extension_api.json`, so that bindings can take appropriate steps, e.g. to remove the methods or warn upon their usage.

### Steps to reproduce

In a GDExtension binding, try to call either of the following methods:
- `ResourceImporterOggVorbis::load_from_buffer` 
- `ResourceImporterOggVorbis::load_from_file`

N/A

YuriSizov:
Fixes https://github.com/godotengine/godot/issues/86206. We can call it a regression from https://github.com/godotengine/godot/pull/49524. Probably worth cherry-picking, though technically this changes the exposed API.

This pattern is used in the GLTF module, btw.

### DIFF
--- a/modules/minimp3/register_types.cpp
+++ b/modules/minimp3/register_types.cpp
@@ -52,8 +52,13 @@ void initialize_minimp3_module(ModuleInitializationLevel p_level) {
 		ResourceFormatImporter::get_singleton()->add_importer(mp3_import);
 	}
 
+	ClassDB::APIType prev_api = ClassDB::get_current_api();
+	ClassDB::set_current_api(ClassDB::API_EDITOR);
+
 	// Required to document import options in the class reference.
 	GDREGISTER_CLASS(ResourceImporterMP3);
+
+	ClassDB::set_current_api(prev_api);
 #endif
 
 	GDREGISTER_CLASS(AudioStreamMP3);
--- a/modules/vorbis/register_types.cpp
+++ b/modules/vorbis/register_types.cpp
@@ -48,8 +48,13 @@ void initialize_vorbis_module(ModuleInitializationLevel p_level) {
 		ResourceFormatImporter::get_singleton()->add_importer(ogg_vorbis_importer);
 	}
 
+	ClassDB::APIType prev_api = ClassDB::get_current_api();
+	ClassDB::set_current_api(ClassDB::API_EDITOR);
+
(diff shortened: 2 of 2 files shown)


Bromeon (approved):
Just recompiled Godot release using commit 0f8c955c1ad19bef9d4d615681c647cbc5a495e1.
The class is now correctly exposed to the JSON as:
```json
		{
			"name": "ResourceImporterOggVorbis",
			"is_refcounted": true,
			"is_instantiable": true,
			"inherits": "ResourceImporter",
			"api_type": "editor",
			"methods": [ <load_from_buffer, load_from_file> ]
		},
```
I'm not very familiar with the codebase, but assuming that `GDREGISTER_CLASS` cannot return, it looks correct. A small test of our dodge-the-creeps example actually shows it working in Release.

Thanks a lot for this super-fast fix! 🙂

YuriSizov:
Thanks for testing!

YuriSizov:
Cherry-picked for 4.2.2.

aaronfranke:
Why was this approach used instead of moving to `if (p_level == MODULE_INITIALIZATION_LEVEL_EDITOR) { ... }`?

YuriSizov:
> Why was this approach used instead of moving to `if (p_level == MODULE_INITIALIZATION_LEVEL_EDITOR) { ... }`?

I simply mimicked existing code from the gltf module which solved the same issue. Although I guess the code there should've been changed in https://github.com/godotengine/godot/commit/de0ca3b999819a99a5dcab99f0083a760277b95e, which introduced `MODULE_INITIALIZATION_LEVEL_EDITOR`, but didn't refactor the old hack.

PR #86211, opened 2023-12-15
Fix duplicating multiple nodes at different depths in `SceneTreeDock`

### LINKED ISSUE #84333: Incorrect UndoRedo when duplicating nodes at different depths.
### Godot version

Godot v4.2.beta3 - Pop!_OS 22.04 LTS - X11 - Vulkan (Forward+) - integrated Intel(R) Graphics (ADL GT2) () - 12th Gen Intel(R) Core(TM) i5-1235U (12 Threads)

### System information

^

### Issue description

image

When duplicating these nodes, I get:

image

When I undo, it deletes the new nodes one by one. The second undo lands me here:

image

Note that the node is now Node3, rather than Node2.

Quite a few unexpected behaviors are going on. I don't know what the functionality here should be, but since undo is failing to land me at my initial place, something is definitely wrong here.

I also had this crash on me while working on a project, but couldn't reproduce it while testing for the bug report.

### Steps to reproduce

Clarified above

N/A

TheSofox:
Fixes #84333.

Duplicating multiple nodes at different depths seemed to put all the new nodes on the same level, which was counterintuitive. Also, undoing this operation was bugged.

I've fixed it so that each duplicated node appears next to the node it was duplicated from at whatever depth it was at. This also fixed the Undo functionality.

### DIFF
--- a/editor/scene_tree_dock.cpp
+++ b/editor/scene_tree_dock.cpp
@@ -713,7 +713,14 @@ void SceneTreeDock::_tool_selected(int p_tool, bool p_confirm_override) {
 
 			selection.sort_custom<Node::Comparator>();
 
-			Node *add_below_node = selection.back()->get();
+			HashMap<const Node *, Node *> add_below_map;
+
+			for (List<Node *>::Element *E = selection.back(); E; E = E->prev()) {
+				Node *node = E->get();
+				if (!add_below_map.has(node->get_parent())) {
+					add_below_map.insert(node->get_parent(), node);
+				}
+			}
 
 			for (Node *node : selection) {
 				Node *parent = node->get_parent();
@@ -740,7 +747,7 @@ void SceneTreeDock::_tool_selected(int p_tool, bool p_confirm_override) {
 
 				dup->set_name(parent->validate_child_name(dup));
 
-				undo_redo->add_do_method(add_below_node, "add_sibling", dup, true);
+				undo_redo->add_do_method(add_below_map[parent], "add_sibling", dup, true);
 
 				for (Node *F : owned) {
 					if (!duplimap.has(F)) {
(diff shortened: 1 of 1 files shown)


KoBeWi:
This changes how nodes on the same level are duplicated:

https://github.com/godotengine/godot/assets/2223172/5b747f8f-1713-4c65-b9ef-c8772e4e0181

vs old

https://github.com/godotengine/godot/assets/2223172/f9afe1cd-e95f-4cff-a220-af12e430297b

I don't think that's expected behavior.

TheSofox:
Question: You duplicate (both nodes, same level):

- TypeA1
- TypeB1

Is the result:

- TypeA1
- TypeA2
- TypeB1
- TypeB2

or:

- TypeA1
- TypeB1
- TypeA2
- TypeB2

KoBeWi:
I'd say the latter.

In case of
- TypeA1
- TypeA2
Expected result is
- TypeA1
- TypeA2
- TypeA3
- TypeA4

In your example, the latter result is consistent with the above.

TheSofox:
Understood. Another question about desired behaviour:

- AudioNode
- Player Node
- AnimationNode
- Particle Node

You select `AudioNode` and `AnimationNode` (I'm assuming order you select them in doesn't matter, let me know if it does) and duplicate. This obviously creates two more nodes (`AudioNode2` and `AnimationNode`) but where are they positioned?

KoBeWi:
Currently you'd end up with
- AudioNode
- PlayerNode
- AnimationNode
- AudioNode2
- AnimationNode2
- ParticleNode

I think you should keep the current behavior for nodes on the same level and only fix duplication with different parents.

TheSofox:
Understood. Thank you for explaining.

TheSofox:
Amended PR and pushed. Like you asked, it sticks with current behavior for nodes on the same level, but still handles having nodes with different parents.

KoBeWi (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #86212, opened 2023-12-15
Expose normal_weight as an advanced LOD generation parameter

zeux:
We had this hard coded to 2 which seems fairly high, and seems like something that you might want to control for now. This was definitely useful when experimenting with the new simplifier.

I'm used to the normal weight being less than 1, so I'm not sure if the range and defaults here are the best, but let's first expose this and worry about defaults and ranges later.

### DIFF
--- a/doc/classes/ImporterMesh.xml
+++ b/doc/classes/ImporterMesh.xml
@@ -46,10 +46,12 @@
 			<return type="void" />
 			<param index="0" name="normal_merge_angle" type="float" />
 			<param index="1" name="normal_split_angle" type="float" />
-			<param index="2" name="bone_transform_array" type="Array" />
+			<param index="2" name="normal_weight" type="float" />
+			<param index="3" name="bone_transform_array" type="Array" />
 			<description>
 				Generates all lods for this ImporterMesh.
 				[param normal_merge_angle] and [param normal_split_angle] are in degrees and used in the same way as the importer settings in [code]lods[/code]. As a good default, use 25 and 60 respectively.
+				[param normal_weight] determines the importance of normals (relative to geometry) during simplification.
 				The number of generated lods can be accessed using [method get_surface_lod_count], and each LOD is available in [method get_surface_lod_size] and [method get_surface_lod_indices].
 				[param bone_transform_array] is an [Array] which can be either empty or contain [Transform3D]s which, for each of the mesh's bone IDs, will apply mesh skinning when generating the LOD mesh variations. This is usually used to account for discrepancies in scale between the mesh itself and its skinning data.
 			</description>
--- a/doc/classes/ImporterMesh.xml
+++ b/doc/classes/ImporterMesh.xml
@@ -46,8 +46,8 @@
 			<return type="void" />
 			<param index="0" name="normal_merge_angle" type="float" />
 			<param index="1" name="normal_split_angle" type="float" />
-			<param index="2" name="normal_weight" type="float" />
-			<param index="3" name="bone_transform_array" type="Array" />
+			<param index="2" name="bone_transform_array" type="Array" />
+			<param index="3" name="normal_weight" type="float" default="2.0" />
 			<description>
 				Generates all lods for this ImporterMesh.
 				[param normal_merge_angle] and [param normal_split_angle] are in degrees and used in the same way as the importer settings in [code]lods[/code]. As a good default, use 25 and 60 respectively.
--- a/doc/classes/ImporterMesh.xml
+++ b/doc/classes/ImporterMesh.xml
@@ -47,11 +47,9 @@
 			<param index="0" name="normal_merge_angle" type="float" />
 			<param index="1" name="normal_split_angle" type="float" />
 			<param index="2" name="bone_transform_array" type="Array" />
-			<param index="3" name="normal_weight" type="float" default="2.0" />
 			<description>
 				Generates all lods for this ImporterMesh.
 				[param normal_merge_angle] and [param normal_split_angle] are in degrees and used in the same way as the importer settings in [code]lods[/code]. As a good default, use 25 and 60 respectively.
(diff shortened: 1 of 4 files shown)


doc/classes/ImporterMesh.xml
```diff
@@ -46,10 +46,12 @@
 			<return type="void" />
 			<param index="0" name="normal_merge_angle" type="float" />
 			<param index="1" name="normal_split_angle" type="float" />
-			<param index="2" name="bone_transform_array" type="Array" />
+			<param index="2" name="normal_weight" type="float" />
+			<param index="3" name="bone_transform_array" type="Array" />
```
AThousandShips, line 50: You need to bind this in the CPP file as well, also you should put this at the end to not break compatibility, and it should have a default value

zeux: Thanks - somehow I didn't realize this is script-bound, not just internal documentation. Fixed.

zeux: Hmm - that doesn't seem to fix .NET bindings (see https://github.com/godotengine/godot/actions/runs/7226635682/job/19692661857?pr=86212). Not sure what the protocol is for cases like this. One option is I can keep the parameter out of the script API for now (e.g. add `generate_lods_impl`, and keep `generate_lods` as is)?

zeux: I've removed the extra parameter from the scripting API for now to avoid compatibility issues altogether; the extra benefit is that the signature with normal weight at the end is fairly weird, and if we add more parameters here in the future, we might want to adjust the script API to take a dictionary of parameters anyway...

scene/resources/importer_mesh.h
```diff
@@ -112,7 +112,9 @@ class ImporterMesh : public Resource {
 
 	void set_surface_material(int p_surface, const Ref<Material> &p_material);
 
-	void generate_lods(float p_normal_merge_angle, float p_normal_split_angle, Array p_skin_pose_transform_array);
+	void generate_lods(float p_normal_merge_angle, float p_normal_split_angle, float p_normal_weight, Array p_skin_pose_transform_array);
+
+	void _generate_lods_script(float p_normal_merge_angle, float p_normal_split_angle, Array p_skin_pose_transform_array);
```
AThousandShips, line 117: This should be moved to the private area and called:
```suggestion
	void _generate_lods_bind(float p_normal_merge_angle, float p_normal_split_angle, Array p_skin_pose_transform_array);
```

doc/classes/ImporterMesh.xml
AThousandShips: You'd need to add compatibility binds, but I think we can go for this, will see what reviewers say

scene/resources/importer_mesh.h
zeux: Moved & renamed

scene/resources/importer_mesh.cpp
```diff
@@ -1364,7 +1367,7 @@ void ImporterMesh::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_surface_name", "surface_idx", "name"), &ImporterMesh::set_surface_name);
 	ClassDB::bind_method(D_METHOD("set_surface_material", "surface_idx", "material"), &ImporterMesh::set_surface_material);
 
-	ClassDB::bind_method(D_METHOD("generate_lods", "normal_merge_angle", "normal_split_angle", "bone_transform_array"), &ImporterMesh::generate_lods);
+	ClassDB::bind_method(D_METHOD("generate_lods", "normal_merge_angle", "normal_split_angle", "bone_transform_array"), &ImporterMesh::_generate_lods_bind);
```
clayjohn, line 1370: If backwards compatibility wasn't a problem would you still use this approach?

We have a system for providing compatibility versions of methods (for GDextensions and C#) to maintain backwards compatibility. 

So if you think that exposing normal_weight to scripting makes sense, then we should do it and just provide the compatibility method as necessary

zeux: I feel like it's not a question as to whether normal_weight should be exposed, but how it should be exposed. If the goal is to maintain compatibility and simultaneously allow future extensions - imagine 5 more floating point parameters here added over the next 5 years - I'd probably favor an interface where an object with the set of named parameters is passed, whether it's a dictionary or a custom type. If we were to extend `generate_lods` as it is with normal weight, we'd need to add it to the end of the list for GDScript compatibility (C# supports function overloading but GDScript doesn't from what I can tell); this is certainly possible, but this ordering feels unnatural due to the existing bone array argument. This is why I thought it would be better to, for now, limit this change to just the manual importer usage; ideally, another function (with a different name I'd guess...) could be added later that accepts a string->float dictionary or something along these lines?

I'm open to adding this as an optional fourth argument, it's just that the process of wrestling with the binding layer made me question whether that's a good idea so keeping the interface as is for now seemed better.

fire: There's been a repeated design where the engine exposes a dictionary of options.

lyuma: Is there a reason it can't be added to the end of the argument list with a default value? Is the whole compat issue because it was added to the middle of the argument list?

If we think it is complex or likely to change enough to require named parameters, in my opinion, the right way to do this is with a RefCounted object which can contain input (and extra output data) and can be reused between invocations (zero allocations)

zeux: I'm open for either route here; it seemed reasonable to exclude this addition from the script binding to begin with , as then the extra parameter can be exposed to scripts separately in a way that people feel is most future proof, but if the consensus is that it's better to just add the extra argument to the end with a C# specific second fixed length overload, I can certainly do that.

zeux:
Going to close this for now in the interest of focusing on improvements to the fundamental algorithms over exposing/tuning custom import parameters.

PR #86213, opened 2023-12-15
SCons: Integrate intellisense hinting for `SConstruct`/`SCsub` files

Repiteo:
Adds `scons_hints.py` to the repo, which adds intellisense logic to all* SCons files in the repo. No logic was changed across any of the SCons scripts, as this only adds the existing type-checking context for an environment. While it's not a 1-1 representation of what SCons will be calling to behind the scenes (it will frequently utilize proxies for that purpose), it *does* allow users to reliably trace back where the calls are being sent to/from in the first place. In practice, this is a simplified version of the `SCons.Script` initalization—excluding private files, extraneous includes, and workaround linking methods—alongside typing hints for `env` and `env_module` so they're immediately recognized as `SConsEnvironment` for files that open with `Import()` (aka: pretty much every single `SCsub` file)

\* Note: Two SCons files were *not* updated. Those being the standalone `SConstruct` files utilized by the text modules, as they're inherently isolated

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from scons_hints import *
+
 EnsureSConsVersion(3, 0, 0)
 EnsurePythonVersion(3, 6)
 
--- a/core/SCsub
+++ b/core/SCsub
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from scons_hints import *
+
 Import("env")
 
 import core_builders
--- a/core/config/SCsub
+++ b/core/config/SCsub
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from scons_hints import *
+
 Import("env")
 
 env_config = env.Clone()
--- a/core/crypto/SCsub
+++ b/core/crypto/SCsub
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from scons_hints import *
+
 Import("env")
 
 env_crypto = env.Clone()
--- a/core/debugger/SCsub
+++ b/core/debugger/SCsub
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from scons_hints import *
+
 Import("env")
 
 env.add_source_files(env.core_sources, "*.cpp")
--- a/core/error/SCsub
+++ b/core/error/SCsub
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from scons_hints import *
+
 Import("env")
 
 env_error = env.Clone()
--- a/core/extension/SCsub
+++ b/core/extension/SCsub
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from scons_hints import *
+
 Import("env")
 
 import make_wrappers
--- a/core/input/SCsub
+++ b/core/input/SCsub
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from scons_hints import *
+
 Import("env")
 
 import input_builders
--- a/core/io/SCsub
+++ b/core/io/SCsub
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from scons_hints import *
+
 Import("env")
 
 env.add_source_files(env.core_sources, "*.cpp")
--- a/core/math/SCsub
+++ b/core/math/SCsub
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from scons_hints import *
+
 Import("env")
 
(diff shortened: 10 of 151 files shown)


yedpodtrzitko:
I'm not convinced this is the right way how to do this, for a few reasons:

- pure "stubs" (type hints, if you will) should be generally located inside `.pyi` files instead `.py` files. Import on these files still works but without the need to have `if TYPE_CHECKING:` everywhere

- even then, these stubs should be ideally generated via `stubgen` (part of `mypy`) instead of creating them manually. 

- having the stubs included in the repo will eventually cause problems with versioning - ie. there is no strict version pin of SCons, so if a newer version of SCons will have different signature of some method, the type hints will be wrong 

- thus the SCons stubs should be ideally contributed into https://github.com/python/typeshed repo instead, so every supported editor can use them in any project using any version of SCons instead of adding ad-hoc import into every single Python file in this repo.

Repiteo:
I'm not positive if that's completely applicable here. SCons files needing type hints is less about setting up specific typing info on a per-file basis, and more about assigning inherent typing data to the filetype itself. It's more comparable to a seperate file extension/type altogether. Not to mention that `SConstruct` and `SCsub` don't utilize `.py` in the first place, so I'm not positive if it even *could* link up with `.pyi` in the first place

I'll absolutely admit to my ignorance on integrating stubs, so I apologize if this is all fully supported after all, but in some initial testing I wasn't able to make it work like I was hoping. If there *is* a proper process, it's currently beyond me

adamscott:
> I dont think this is the right way how to achieve this.
> 
>     * stubs (type hints, if you will) for various tools should be ideally sourced via  `.pyi` files rather than emulating them via `.py` file which needs to be imported everywhere
> 
>     * `.pyi` stubs can/should be generated via `stubgen` (part of `mypy`), and these stubs can be then installed as a package, your ide/editor then recognize and use them
> 
>     * the standard way how to provide the stubs is via Typeshed repo https://github.com/python/typeshed , which then provides a new package (presumably `types-scons`) which can be added/installed as a dependency by anyone who is interested in having the stubs for SCons.

I tried, but it didn't work. I don't think .pyi files are the solution here. I even tried to put stuff into `__main__.pyi`, but it doesn't seem to apply typing on the SConstruct file.



> I'll absolutely admit to my ignorance on integrating stubs, so I apologize if this is all fully supported after all, but in some initial testing I wasn't able to make it work like I was hoping. If there _is_ a proper process, it's currently beyond me

Same.

yedpodtrzitko:
> I tried, but it didn't work. I don't think .pyi files are the solution here. I even tried to put stuff into __main__.pyi, but it doesn't seem to apply typing on the SConstruct file.

so it's not as straightforward as I thought. But I did the following, and it works with VS Code:

since neither `SConstruct` nor `SCons` are executed directly, there are no `import` hinting where the code comes from, so to make it work the stubs need to be actually placed into `__builtins__.pyi`:

```
cd godot # go to godot folder
mkdir typings # default dir where VSCode search for stubs
# create a random stub
cat <<EOT >> typings/__builtins__.pyi
from _typeshed import Incomplete

EnsureSConsVersion: Incomplete
EOT
````

this should make `EnsureSConsVersion` recognized without any import, and it does indeed:


Screenshot 2023-12-23 at 16 11 19

Repiteo:
Commented out the contents of `scons_hints.py` while trying out this method, and it indeed seems to function! However, it's not an air-tight solution out of the box, because there's two issues persisting. The first is the `mypy` extension itself, as it will now throw a fit within `SConstruct`/`SCsub` files, despite linking up the type:

23-12-23 10-34-58 Code

The second, more pressing concern is that these hints are now no longer limited to `SConstruct`/`SCsub` files. This will lead to an inverse issue, causing false positives in all the `.py` files that aren't instrinsically part of an SCons environment:

23-12-23 10-46-11 Code

yedpodtrzitko:
regarding mypy, `MYPYPATH` needs to be adjusted to take the stubs into account, see https://mypy.readthedocs.io/en/stable/stubs.html

regarding the second point - I dont understand what's on the screenshot and how does it relate to the issue you are describing 🤔

Repiteo:
The second screenshot showcases `EnsureSConsVersion` appearing as hinted/valid within a `methods.py` function. These files are not part of the SCons environment, so these should *not* register as valid in this context. The terminal below demonstrates the function failing because it's trying to call a function it believes is builtin, but isn't

yedpodtrzitko:
ah gotcha. I dont have solution for that.

Repiteo:
Superseded by #93058

PR #86218, opened 2023-12-15
Fix `Node` selection when navigating `Tree` with Arrow keys

### LINKED ISSUE #12544: Arrow keys doesn't select objects on the scene hierarchy
**Operating system or device, Godot version, GPU Model and driver (if graphics related):**
Godot 3.0 Alpha 2, Windows 10 64 bit

**Issue description:**
Arrow keys doesn't select objects on the scene hierarchy, please see this video as reference:
https://youtu.be/lRbefcsYDi4?t=99

Notice how by using the arrow keys doesn't really select the items, only changes the property window information but doesn't have any effect on the viewport.

TheSofox:
Fixes  #12544

Arrow keys change actual selection (not just cursor) when navigating `Tree` and `SceneTreeDock`.

### DIFF
--- a/scene/gui/tree.cpp
+++ b/scene/gui/tree.cpp
@@ -3233,8 +3233,8 @@ void Tree::_go_up() {
 		if (!prev) {
 			return;
 		}
-		selected_item = prev;
-		emit_signal(SNAME("cell_selected"));
+
+		select_single_item(prev, get_root(), selected_col);
 		queue_redraw();
 	} else {
 		int col = selected_col < 0 ? 0 : selected_col;
@@ -3266,8 +3266,7 @@ void Tree::_go_down() {
 			return;
 		}
 
-		selected_item = next;
-		emit_signal(SNAME("cell_selected"));
+		select_single_item(next, get_root(), selected_col);
 		queue_redraw();
 	} else {
 		int col = selected_col < 0 ? 0 : selected_col;


scene/gui/tree.cpp
```diff
@@ -3233,7 +3233,8 @@ void Tree::_go_up() {
 		if (!prev) {
 			return;
 		}
-		selected_item = prev;
+
+		select_single_item(prev, get_root(), selected_col);
 		emit_signal(SNAME("cell_selected"));
```
AThousandShips, line 3238: This will now be emitted twice

TheSofox: Removed `emit_signal` in the two places it appeared under `select_single_item`

KoBeWi (approved):
`select_single_item()` looks more expensive than what was done previously, but I tested with a very big scene and it's not noticeable.

Also using that method might change what signals are emitted in different selection modes, but this could be desirable (and maybe fix other existing issues).

YuriSizov:
Thanks!

scene/gui/tree.cpp
```diff
@@ -3233,8 +3233,8 @@ void Tree::_go_up() {
 		if (!prev) {
 			return;
 		}
-		selected_item = prev;
-		emit_signal(SNAME("cell_selected"));
+
+		select_single_item(prev, get_root(), selected_col);
```
aaronfranke, line 3237: Why is `select_single_item` used for `SELECT_MULTI`?

TheSofox: Arrow key overrides it. It's exactly what happens in other systems, such as in Windows file explorer.

aaronfranke: Can you take a look at #92959? The code currently crashes, and when I fixed the crash, @KoBeWi reported incorrect behavior, that it selects items which are not visible.

PR #86220, opened 2023-12-16
Fix changing `gui_embed_subwindows` while a child window is displayed

### LINKED ISSUE #86203: Unexpected crash when updating gui_embed_subwindows to false with extra windows active
### Tested versions

Reproducible in 4.1.3, 4.2 stable

### System information

Godot v4.2.stable - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3070 (NVIDIA; 31.0.15.3699) - Intel(R) Core(TM) i7-10700F CPU @ 2.90GHz (16 Threads)

### Issue description

NB: Tested on the root viewport.

gui_embed_subwindows being updated during runtime with windows active causes strange results. 

I see that this makes sense if you take it as given that this value represents that the root viewport should or should not have external windows, and we have some and are setting it to not be able to have any, so presumably there are some underlying functions that won't interact with the external windows as expected in this state. (And vice versa for embedded)

On the off chance someone else is having trouble with this: I have been able to workaround the issues by ensuring that any external windows are freed prior to updating the value when external windows are present. I assume alternatively you could reparent them to another viewport in the interim if your use case for some reason requires they be retained. 

```func _process(_delta):
	if Input.is_action_just_pressed("debug"):
		if get_child_count() > 0:
			get_child(0).queue_free()
			await get_tree().process_frame
		get_tree().root.get_viewport().gui_embed_subwindows = !get_tree().root.get_viewport().gui_embed_subwindows
	elif Input.is_action_just_pressed("debug_2"):
		if get_child_count() > 0:
			get_child(0).queue_free()
		var new_window = window_scene.instantiate()
		add_child(new_window)
```

Alternatively I feel that having multiple viewports and not relying on changing the gui_embed_subwindows value to manage external Vs. embedded windows in your game is probably the **sanest** approach.

### Steps to reproduce

See MRP

"1" - Spawns a window, freeing the previous
"2" - Change the gui_embed_subwindows value
"3" - Try to hide the first window in the list

Pressing 1, 2, and then 3 results in a crash as we start by spawning an external window, changing gui_embed_subwindows to true, then attempt to interact with the external window by hiding hiding it

Pressing 1, 2, and then 1 results in an orphaned window which godot doesn't have in the remote tree  This changes visually to being entirely full of black once it is refreshed by e.g. resizing it.

Starting by pressing 2 once and then any combination of 1 and 3 will seemingly not result in the same issues, nor will you run into an issue after pressing 2 again regardless of the embedded window state e.g. 2 - 1 - 2 - 1 or 2 - 1 - 3 - 1. 

Edit: With 2 - 1 - 2 - 3 - 3 - 3 we end up with our previously embedded window as an external window, while still retaining the embedded window. When toggling its visibility it switches between a broken embedded window and hidden external window, and a working embedded window with a working external window. 

So it seems like the first hide after switching from embedded windows to external toggles a new active external window representation of the node, and upsets the logic behind hiding the original embedded representation 

[Min_crash_repro.zip](https://github.com/godotengine/godot/files/13687858/Min_crash_repro.zip)

Sauermann:
Changing `gui_embed_subwindows` while a child window is displayed can lead to crashes or other unintended behavior.
This PR adds checks to prevent the change while children windows are displayed.

resolve #86203

### DIFF
--- a/scene/main/viewport.cpp
+++ b/scene/main/viewport.cpp
@@ -3848,7 +3848,66 @@ Viewport *Viewport::get_parent_viewport() const {
 
 void Viewport::set_embedding_subwindows(bool p_embed) {
 	ERR_THREAD_GUARD;
-	gui.embed_subwindows_hint = p_embed;
+	if (gui.embed_subwindows_hint == p_embed) {
+		return;
+	}
+
+	bool allow_change = true;
+
+	if (!is_inside_tree()) {
+		// Change can happen since no child window is displayed.
+	} else if (gui.embed_subwindows_hint) {
+		if (!gui.sub_windows.is_empty()) {
+			// Prevent change when this viewport has embedded windows.
+			allow_change = false;
+		}
+	} else {
+		Viewport *vp = this;
+		while (true) {
+			if (!vp->get_parent()) {
+				// Root window reached.
+				break;
+			}
+			vp = vp->get_parent()->get_viewport();
+			if (vp->is_embedding_subwindows()) {
+				for (int i = 0; i < vp->gui.sub_windows.size(); i++) {
+					if (is_ancestor_of(vp->gui.sub_windows[i].window)) {
+						// Prevent change when this viewport has child windows that are displayed in an ancestor viewport.
+						allow_change = false;
+						break;
+					}
+				}
+			}
+		}
+
+		if (allow_change) {
+			Vector<int> wl = DisplayServer::get_singleton()->get_window_list();
+			for (int index = 0; index < wl.size(); index++) {
+				DisplayServer::WindowID wid = wl[index];
+				if (wid == DisplayServer::INVALID_WINDOW_ID) {
+					continue;
+				}
+
+				ObjectID woid = DisplayServer::get_singleton()->window_get_attached_instance_id(wid);
+				if (woid.is_null()) {
+					continue;
+				}
+
+				Window *w = Object::cast_to<Window>(ObjectDB::get_instance(woid));
+				if (w && is_ancestor_of(w)) {
+					// Prevent change when this viewport has child windows that are displayed as native windows.
+					allow_change = false;
+					break;
+				}
+			}
+		}
+	}
+
+	if (allow_change) {
+		gui.embed_subwindows_hint = p_embed;
+	} else {
+		WARN_PRINT("Can't change \"gui_embed_subwindows\" while a child window is displayed. Consider hiding all child windows before changing this value.");
+	}
 }
 
 bool Viewport::is_embedding_subwindows() const {


scene/main/viewport.cpp
```diff
@@ -3848,7 +3848,67 @@ Viewport *Viewport::get_parent_viewport() const {
...
+		Viewport *vp = this;
+		while (true) {
+			if (!vp->get_parent()) {
+				break;
+			}
+			vp = vp->get_parent()->get_viewport();
+			if (!vp) {
+				// Necessary in cases when function is called on viewports that are not in the scene tree.
+				break;
+			}
```
YuriSizov, line 3872: We should be able to check that before the loop. There is no situation when our current viewport is in tree, but its parents somehow aren't.

scene/main/viewport.cpp
```diff
@@ -3848,7 +3848,67 @@ Viewport *Viewport::get_parent_viewport() const {
...
+			if (vp->is_embedding_subwindows()) {
+				for (int i = 0; i < vp->gui.sub_windows.size(); i++) {
+					if (this->is_ancestor_of(vp->gui.sub_windows[i].window)) {
+						// Prevent change when this viewport has children windows, that are displayed in an ancestor viewport.
+						allow_change = false;
+						break;
+					}
+				}
```
YuriSizov, line 3880: Hmm, when would that be possible? Does viewport embedding not follow the tree hierarchy and order?

YuriSizov (approved):
Seems pretty sensible, aside from a couple of notes that I've left. This will make things slower, naturally, but we shouldn't expect such property change to be done continuously.

scene/main/viewport.cpp
Sauermann: Good idea thanks, done.

scene/main/viewport.cpp
Sauermann: Take the following scene tree:
```
- Root Window (embed windows)
  - SubViewportContainer
    - SubViewport (don't embed windows)
      - Window
```
In this case the Window is embedded within the Root Window and not in the SubViewport.
So it is possible that Windows get embedded further up the scene tree and get not embedded in the nearest Viewport.

The above code was necessary to prevent changing `gui_embed_subwindows` of the SubViewport while the Window is displayed in embedded form in the Root Window.

scene/main/viewport.cpp
```diff
@@ -3848,7 +3848,67 @@ Viewport *Viewport::get_parent_viewport() const {
...
+		if (allow_change) {
+			Vector<int> wl = DisplayServer::get_singleton()->get_window_list();
+			for (int index = 0; index < wl.size(); index++) {
+				DisplayServer::WindowID wid = wl[index];
+				if (wid == DisplayServer::INVALID_WINDOW_ID) {
+					continue;
+				}
+
+				ObjectID woid = DisplayServer::get_singleton()->window_get_attached_instance_id(wid);
+				if (woid == ObjectID()) {
```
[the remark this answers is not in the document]
Sauermann: Thanks for your improvements. I have included them.

scene/main/viewport.cpp
```diff
@@ -3848,7 +3848,66 @@ Viewport *Viewport::get_parent_viewport() const {
...
+					break;
+				}
+			}
+		}
+	}
+
+	if (allow_change) {
+		gui.embed_subwindows_hint = p_embed;
+	} else {
+		WARN_PRINT("Can't change gui_embed_subwindows, while a child window is displayed. Consider hiding all child windows before changing this value.");
```
akien-mga, line 3909:
```suggestion
		WARN_PRINT("Can't change \"gui_embed_subwindows\" while a child window is displayed. Consider hiding all child windows before changing this value.");
```

scene/main/viewport.cpp
```diff
@@ -3848,7 +3848,66 @@ Viewport *Viewport::get_parent_viewport() const {
...
+				}
+
+				ObjectID woid = DisplayServer::get_singleton()->window_get_attached_instance_id(wid);
+				if (woid.is_null()) {
+					continue;
+				}
+
+				Window *w = Object::cast_to<Window>(ObjectDB::get_instance(woid));
+				if (w && this->is_ancestor_of(w)) {
+					// Prevent change when this viewport has children windows, that are displayed as native windows.
```
akien-mga, line 3898:
```suggestion
					// Prevent change when this viewport has children windows that are displayed as native windows.
```

scene/main/viewport.cpp
```diff
@@ -3848,7 +3848,66 @@ Viewport *Viewport::get_parent_viewport() const {
...
+		while (true) {
+			if (!vp->get_parent()) {
+				// Root window reached.
+				break;
+			}
+			vp = vp->get_parent()->get_viewport();
+			if (vp->is_embedding_subwindows()) {
+				for (int i = 0; i < vp->gui.sub_windows.size(); i++) {
+					if (this->is_ancestor_of(vp->gui.sub_windows[i].window)) {
+						// Prevent change when this viewport has children windows, that are displayed in an ancestor viewport.
```
akien-mga, line 3875:
```suggestion
						// Prevent change when this viewport has children windows that are displayed in an ancestor viewport.
```

scene/main/viewport.cpp
```diff
@@ -3848,7 +3848,66 @@ Viewport *Viewport::get_parent_viewport() const {
...
 	ERR_THREAD_GUARD;
-	gui.embed_subwindows_hint = p_embed;
+	if (gui.embed_subwindows_hint == p_embed) {
+		return;
+	}
+
+	bool allow_change = true;
+
+	if (!is_inside_tree()) {
+		// Change can happen, since no child window is displayed.
```
akien-mga, line 3858:
```suggestion
		// Change can happen since no child window is displayed.
```

scene/main/viewport.cpp
```diff
@@ -3848,7 +3848,66 @@ Viewport *Viewport::get_parent_viewport() const {
...
+		Viewport *vp = this;
+		while (true) {
+			if (!vp->get_parent()) {
+				// Root window reached.
+				break;
+			}
+			vp = vp->get_parent()->get_viewport();
+			if (vp->is_embedding_subwindows()) {
+				for (int i = 0; i < vp->gui.sub_windows.size(); i++) {
+					if (this->is_ancestor_of(vp->gui.sub_windows[i].window)) {
```
AThousandShips, line 3874:
```suggestion
					if (is_ancestor_of(vp->gui.sub_windows[i].window)) {
```
Unnecessary `this`

scene/main/viewport.cpp
```diff
@@ -3848,7 +3848,66 @@ Viewport *Viewport::get_parent_viewport() const {
...
+					continue;
+				}
+
+				ObjectID woid = DisplayServer::get_singleton()->window_get_attached_instance_id(wid);
+				if (woid.is_null()) {
+					continue;
+				}
+
+				Window *w = Object::cast_to<Window>(ObjectDB::get_instance(woid));
+				if (w && this->is_ancestor_of(w)) {
```
AThousandShips, line 3897:
```suggestion
				if (w && is_ancestor_of(w)) {
```

scene/main/viewport.cpp
Sauermann: Right, it is unnecessary. Thanks.

scene/main/viewport.cpp
Sauermann: Oh my. I have to get better with these. Thanks!

akien-mga (approved):

akien-mga:
Thanks!

scene/main/viewport.cpp
YuriSizov: I see, though this is surprising. I would expect `Window` to only be affected by its closes Viewport parent. Was this discussed at any point, or was it like this already when you started looking into viewport related issues?

Sauermann: As far as I can tell, this is the original behavior since Godot 4 Alpha.
I'm not aware, that it was discussed anywhere.

PR #81180, opened 2023-08-30
Better respect auto-translation of static strings

KoBeWi:
I noticed that GUI nodes make heavy use of RTR. It has 2 problems:
- minor one is that this makes the strings translated in scene nodes (mostly affects FileDialog)
- major one is that it forces the string to be translated even if `auto_translate` is disabled

I changed RTR to TTRC where internal auto-translation can be used (Buttons and PopupMenus; since #75012 it works even in the editor). For dynamic strings I added a new `ATR()` function, which is the same as already existing `atr()`, but it's used for strings that should be fetched for translation.

Partially addresses #80549 If we make `auto_translate` propagated, this PR is required to make it actually work.

Note that we need to add ATR to the translation fetch list (I think it's done on the translation repo?)

PR #86222, opened 2023-12-16
Add option to add built-in strings in the POT generation

### LINKED ISSUE #3827: Add a way to get list of strings used in GUI nodes

A game with level editor that uses various UI nodes.

I use FileDialog in my game and noticed that its texts are not translated (the game has localization support). Seems like the only way to have localized FileDialog is to look for any untranslated text and add it manually to your translation file. This is very error-prone, as some texts might be hidden in tooltips etc, so you can easily miss something.

It would be nice to have some option to fetch a list of all strings used internally by GUI nodes. FileDialog, ConfirmationDialog, anything that has its own text, should all dump it somewhere, so the translators could access it and translate. Or even better - since all these texts are already translated in the editor, it would be useful to fetch these strings along with available translations for the selected languages.

Add an option somewhere "Extract GUI strings" or maybe they should be automatically included when generating POT file.

Not really, but you can solve it by looking at untranslated strings 👀 and typing them manually

It's an important missing feature in otherwise great localization support.

YeldhamDev:
This PR adds the option to add the built-in strings that can appear inside user projects (such as `FileDialog` node and the menu from `LineEdit`) into the POT generation. This is made possible with the new `ETR/ETRN()` functions, which from now on should be used on those occasions. Here's the sister PR on the "l10n" repo: https://github.com/godotengine/godot-editor-l10n/pull/13

As a bonus, this PR also makes so that the POT generator stops putting location comments into strings that don't have one.

*Bugsquad edit:* Closes https://github.com/godotengine/godot-proposals/issues/3827 Supersedes #81180
*Production edit:* Closes godotengine/godot-roadmap#33


Calinou:
Does this pave the way to https://github.com/godotengine/godot-proposals/issues/7334?

YeldhamDev:
@Calinou It's a first step. For, it just populates the POT, but the plan is to add translations to the PO files in the future.

KoBeWi:
Generate POT is disabled when you have only built-in strings included (no scenes). Not sure how useful is that, but technically it could work.

YeldhamDev:
@KoBeWi As stated above, this PR won't work until its sister PR is merged (https://github.com/godotengine/godot-editor-l10n/pull/13).

YeldhamDev:
Alright, I made an important change on how `ETR/ETRN()` works: now, they won't return translated strings at all, but instead the exact string it was given to them. They will now work purely to tag strings to be extract for translation.

This change was made so they could work better with `auto_translate`, because now the original string is preserved, and the controls themselves can handle a translation change with `atr()` internally. I think this technically supersedes #81180 for what it was intended to do (WDYT @KoBeWi?).

`auto_translated` isn't working as best as it could however, since it only affects the control itself, and not its children, which is bad for how some UI nodes in Godot work. I plan to open a proposal on how to tackle that.

KoBeWi:
ETR won't work for formatted text. Auto-translation only has effect on exact strings.
My PR adds ATR, which is like RTR, but respects `auto_translate`.

YeldhamDev:
> My PR adds ATR, which is like RTR, but respects auto_translate.

I'm aware, but that method has a flaw: it overrides the original string if the locale is not set to "en" at the start. That means if a control has a lot of other controls (e. g. `FileDialog`), it would need to updated all of them manually if `auto_translated` is changed. But if the original string is preserved, it can let the controls themselves update the string.

> I see you do atr(ETR( in some places, but you missed some strings.

I see. I will update those them.

KoBeWi:
I only used ATR for dynamic texts. It's basically the same as your `atr(ETR(`.

YeldhamDev:
@KoBeWi I've added more cases, give me a heads-up if I forgot anything else.

KoBeWi:
This change is missing if you were to supersede my PR:
https://github.com/godotengine/godot/pull/81180/files#diff-f7fb6c2c4c5d4fce35e389297048f54be3d4b08621edfc55a5586b437031ff10R51-R59
You already added TRANSLATION_CHANGED to ColorPicker, so you could handle this as well.

core/object/object.cpp
```diff
@@ -1473,10 +1473,14 @@ String Object::tr(const StringName &p_message, const StringName &p_context) cons
 	}
 
 	if (Engine::get_singleton()->is_editor_hint()) {
-		return TranslationServer::get_singleton()->tool_translate(p_message, p_context);
-	} else {
-		return TranslationServer::get_singleton()->translate(p_message, p_context);
+		String tr_msg = TranslationServer::get_singleton()->extractable_translate(p_message, p_context);
+		if (tr_msg.is_empty() || tr_msg == p_message) {
```
KoBeWi, line 1477: This condition will cause lots of long string comparisons. Any way to avoid it?

YeldhamDev: I think I can drop the string comparison, albeit the cases where the first condition wouldn't be triggered first would be small.

scene/main/viewport.cpp
```diff
@@ -1429,7 +1429,7 @@ String Viewport::_gui_get_tooltip(Control *p_control, const Vector2 &p_pos, Cont
 	String tooltip;
 
 	while (p_control) {
-		tooltip = p_control->get_tooltip(pos);
+		tooltip = p_control->atr(p_control->get_tooltip(pos));
```
YeldhamDev, line 1432: > This change is missing if you were to supersede my PR:
https://github.com/godotengine/godot/pull/81180/files#diff-f7fb6c2c4c5d4fce35e389297048f54be3d4b08621edfc55a5586b437031ff10R51-R59
You already added TRANSLATION_CHANGED to ColorPicker, so you could handle this as well.

Not necessary thanks to this change.

core/object/object.cpp
YeldhamDev: Done.

KoBeWi (approved):
Looks fine on GUI side.

YeldhamDev:
Here's the proposal I was talking about: https://github.com/godotengine/godot-proposals/issues/8897

Kiisu-Master:
Would this also enable translating built in strings with CSV or is that not possible?
I'm translating my app with CSV and have been searching for ways to also translate some built-in context menus, but I haven't found any info. Even for Gettext.

YeldhamDev:
@Kiisu-Master Built-in nodes should already be using translations when available.

Chubercik (approved):

core/object/object.cpp
```diff
@@ -1483,10 +1483,13 @@ String Object::tr(const StringName &p_message, const StringName &p_context) cons
 	}
 
 	if (Engine::get_singleton()->is_editor_hint()) {
-		return TranslationServer::get_singleton()->tool_translate(p_message, p_context);
-	} else {
-		return TranslationServer::get_singleton()->translate(p_message, p_context);
+		String tr_msg = TranslationServer::get_singleton()->extractable_translate(p_message, p_context);
```
akien-mga, line 1486: To be clear, is it intentional that `tr_msg` isn't returned?

editor/editor_translation.h
```diff
@@ -32,11 +32,14 @@
...
 #include "core/string/ustring.h"
+#include "core/templates/list.h"
 #include "core/templates/vector.h"
 
 Vector<String> get_editor_locales();
 void load_editor_translations(const String &p_locale);
-void load_doc_translations(const String &p_locale);
 void load_property_translations(const String &p_locale);
+void load_doc_translations(const String &p_locale);
+void load_extractable_translations(const String &p_locale);
```
akien-mga, line 42: This is starting to look like something that could be refactored in a single method, using an enum parameter to define whether it's editor, property, doc, or extractable.

But that can be for future rework.

scene/gui/color_picker.cpp
```diff
@@ -689,7 +705,7 @@ void ColorPicker::_text_type_toggled() {
 		text_type->set_icon(nullptr);
 
 		c_text->set_editable(true);
-		c_text->set_tooltip_text(RTR("Enter a hex code (\"#ff0000\") or named color (\"red\")."));
+		c_text->set_tooltip_text(ETR("Enter a hex code (\"#ff0000\") or named color (\"red\")."));
```
akien-mga, line 708: Is it intentional that some of the tooltips use `atr(ETR())` and others just `ETR()`?

akien-mga (commented):
Looks good overall, though I'm a bit confused about when we should use `atr(ETR(...))` and when it should only be `ETR(...)`.

core/object/object.cpp
YeldhamDev: Oops! Fixed.

scene/gui/color_picker.cpp
YeldhamDev: Tooltips obey auto translation, so if they are a static string, they don't need `atr()`. However, if the string has a dynamic component, it needs to be translated directly.

akien-mga (approved):

akien-mga:
Thanks!

PR #82412, opened 2023-09-26
Replace String.substr() with simpler functions where possible

MewPurPur:
`substr(0, len)` can be simplified to `left(len)` if `len >= 0`. In fact, this alternative is 10% faster.

While this is mostly about code style, I expect the performance improvement to be subtle, but non-negligible. Since String also used `substr()` in a few places when it could use `left()`, this should optimize other String functions too, and those are used all over the codebase, too.

And a couple of other code simplifications.

PR #80824, opened 2023-08-20
Optimize `String.left()` and `String.right()`

MewPurPur:
Makes these functions ~15% faster. The functions are already extremely fast, but this optimization doesn't do anything super complex, it just bypasses conditions from being unnecessarily checked again in substr() and directly copies data into the new string instead.

PR #86235, opened 2023-12-16
Optimize string operations throughout the codebase

MewPurPur:
A more manageable version of #82412 that does more types of optimizations, but only applies them to user-exposed classes (so the total volume of them is a lot less). No benchmarking done.

For reference, most of them are like:

- `s.substr(0, x)` -> `s.left(x)`
- `s.substr(0, s.length() - x)` -> `s.left(-x)`
- `s.substr(x, s.length())` -> `s.right(-x)`
- `s.substr(x, s.length() - x)` -> `s.right(-x)`

I think all of these are equivalent 100% of the time, but they are ~15% faster (benchmarking from my PR that optimized left and right).

### DIFF
--- a/core/io/file_access_compressed.cpp
+++ b/core/io/file_access_compressed.cpp
@@ -34,7 +34,7 @@
 
 void FileAccessCompressed::configure(const String &p_magic, Compression::Mode p_mode, uint32_t p_block_size) {
 	magic = p_magic.ascii().get_data();
-	magic = (magic + "    ").substr(0, 4);
+	magic = magic.rpad(4, " ");
 
 	cmode = p_mode;
 	block_size = p_block_size;
--- a/core/io/http_client.cpp
+++ b/core/io/http_client.cpp
@@ -94,7 +94,7 @@ String HTTPClient::query_string_from_dict(const Dictionary &p_dict) {
 			}
 		}
 	}
-	return query.substr(1);
+	return query.right(-1);
 }
 
 Error HTTPClient::verify_headers(const Vector<String> &p_headers) {
@@ -117,8 +117,8 @@ Dictionary HTTPClient::_get_response_headers_as_dictionary() {
 		if (sp == -1) {
 			continue;
 		}
-		String key = s.substr(0, sp).strip_edges();
-		String value = s.substr(sp + 1, s.length()).strip_edges();
+		String key = s.left(sp).strip_edges();
+		String value = s.substr(sp + 1).strip_edges();
 		ret[key] = value;
 	}
 
--- a/core/string/translation_po.cpp
+++ b/core/string/translation_po.cpp
@@ -172,10 +172,10 @@ void TranslationPO::_cache_plural_tests(const String &p_plural_rule) {
 		return;
 	}
 
-	String equi_test = p_plural_rule.substr(0, first_ques_mark).strip_edges();
+	String equi_test = p_plural_rule.left(first_ques_mark).strip_edges();
 	equi_tests.push_back(equi_test);
 
-	String after_colon = p_plural_rule.substr(p_plural_rule.find(":") + 1, p_plural_rule.length());
+	String after_colon = p_plural_rule.substr(p_plural_rule.find(":") + 1);
 	_cache_plural_tests(after_colon);
 }
 
--- a/core/string/ustring.cpp
+++ b/core/string/ustring.cpp
@@ -231,20 +231,20 @@ Error String::parse_url(String &r_scheme, String &r_host, int &r_port, String &r
 	int pos = base.find("://");
 	// Scheme
 	if (pos != -1) {
-		r_scheme = base.substr(0, pos + 3).to_lower();
-		base = base.substr(pos + 3, base.length() - pos - 3);
+		r_scheme = base.left(pos + 3).to_lower();
+		base = base.substr(pos + 3);
 	}
 	pos = base.find("/");
 	// Path
 	if (pos != -1) {
-		r_path = base.substr(pos, base.length() - pos);
-		base = base.substr(0, pos);
+		r_path = base.right(-pos);
+		base = base.left(pos);
 	}
 	// Host
 	pos = base.find("@");
 	if (pos != -1) {
-		// Strip credentials
-		base = base.substr(pos + 1, base.length() - pos - 1);
+		// Strip credentials.
+		base = base.substr(pos + 1);
 	}
 	if (base.begins_with("[")) {
 		// Literal IPv6
@@ -253,7 +253,7 @@ Error String::parse_url(String &r_scheme, String &r_host, int &r_port, String &r
 			return ERR_INVALID_PARAMETER;
 		}
 		r_host = base.substr(1, pos - 1);
-		base = base.substr(pos + 1, base.length() - pos - 1);
+		base = base.substr(pos + 1);
 	} else {
 		// Anything else
 		if (base.get_slice_count(":") > 2) {
@@ -264,8 +264,8 @@ Error String::parse_url(String &r_scheme, String &r_host, int &r_port, String &r
 			r_host = base;
 			base = "";
 		} else {
-			r_host = base.substr(0, pos);
-			base = base.substr(pos, base.length() - pos);
+			r_host = base.left(pos);
+			base = base.right(-pos);
 		}
 	}
 	if (r_host.is_empty()) {
@@ -1297,9 +1297,9 @@ Vector<String> String::rsplit(const String &p_splitter, bool p_allow_empty, int 
 
 	while (true) {
 		if (remaining_len < p_splitter.length() || (p_maxsplit > 0 && p_maxsplit == ret.size())) {
-			// no room for another splitter or hit max splits, push what's left and we're done
+			// No room for another splitter or hit max splits, push what's left and we're done.
 			if (p_allow_empty || remaining_len > 0) {
-				ret.push_back(substr(0, remaining_len));
+				ret.push_back(left(remaining_len));
 			}
 			break;
 		}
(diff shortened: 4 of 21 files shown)


core/io/file_access_compressed.cpp
```diff
@@ -33,8 +33,7 @@
 #include "core/string/print_string.h"
 
 void FileAccessCompressed::configure(const String &p_magic, Compression::Mode p_mode, uint32_t p_block_size) {
-	magic = p_magic.ascii().get_data();
-	magic = (magic + "    ").substr(0, 4);
+	magic = p_magic.ascii().get_data().rpad(4, " ");
```
AThousandShips, line 36: This calls `rpad` on `comst char *`

core/io/http_client.cpp
```diff
@@ -117,8 +117,8 @@ Dictionary HTTPClient::_get_response_headers_as_dictionary() {
 		if (sp == -1) {
 			continue;
 		}
-		String key = s.substr(0, sp).strip_edges();
-		String value = s.substr(sp + 1, s.length()).strip_edges();
+		String key = s.left(sp).strip_edges();
+		String value = s.right(-(sp + 1)).strip_edges();
```
AThousandShips, line 121: This feels very hard to read, I don't think it's an improvement readability wise

core/string/translation_po.cpp
```diff
@@ -172,10 +172,10 @@ void TranslationPO::_cache_plural_tests(const String &p_plural_rule) {
 		return;
 	}
 
-	String equi_test = p_plural_rule.substr(0, first_ques_mark).strip_edges();
+	String equi_test = p_plural_rule.left(first_ques_mark).strip_edges();
 	equi_tests.push_back(equi_test);
 
-	String after_colon = p_plural_rule.substr(p_plural_rule.find(":") + 1, p_plural_rule.length());
+	String after_colon = p_plural_rule.right(-(p_plural_rule.find(":") + 1));
```
AThousandShips, line 178: Ditto

core/string/ustring.cpp
```diff
@@ -231,20 +231,20 @@ Error String::parse_url(String &r_scheme, String &r_host, int &r_port, String &r
 	int pos = base.find("://");
 	// Scheme
 	if (pos != -1) {
-		r_scheme = base.substr(0, pos + 3).to_lower();
-		base = base.substr(pos + 3, base.length() - pos - 3);
+		r_scheme = base.left(pos + 3).to_lower();
+		base = base.right(-pos - 3);
```
AThousandShips, line 235: Ditto

core/string/ustring.cpp
```diff
@@ -231,20 +231,20 @@ Error String::parse_url(String &r_scheme, String &r_host, int &r_port, String &r
...
+		r_path = base.right(-pos);
+		base = base.left(pos);
 	}
 	// Host
 	pos = base.find("@");
 	if (pos != -1) {
-		// Strip credentials
-		base = base.substr(pos + 1, base.length() - pos - 1);
+		// Strip credentials.
+		base = base.right(-pos - 1);
```
AThousandShips, line 247: Ditto

core/string/ustring.cpp
```diff
@@ -253,7 +253,7 @@ Error String::parse_url(String &r_scheme, String &r_host, int &r_port, String &r
 			return ERR_INVALID_PARAMETER;
 		}
 		r_host = base.substr(1, pos - 1);
-		base = base.substr(pos + 1, base.length() - pos - 1);
+		base = base.right(-pos - 1);
```
AThousandShips, line 256: Ditto

scene/gui/file_dialog.cpp
```diff
@@ -800,8 +800,8 @@ void FileDialog::set_current_path(const String &p_path) {
 	if (pos == -1) {
 		set_current_file(p_path);
 	} else {
-		String path_dir = p_path.substr(0, pos);
-		String path_file = p_path.substr(pos + 1, p_path.length());
+		String path_dir = p_path.left(pos);
+		String path_file = p_path.right(-(pos + 1));
```
AThousandShips, line 804: Ditto

scene/gui/rich_text_label.cpp
```diff
@@ -4059,7 +4059,7 @@ void RichTextLabel::append_text(const String &p_bbcode) {
 				const String &expr = split_tag_block[i];
 				int value_pos = expr.find("=");
 				if (value_pos > -1) {
-					bbcode_options[expr.substr(0, value_pos)] = expr.substr(value_pos + 1).unquote();
+					bbcode_options[expr.left(value_pos)] = expr.right(-value_pos - 1).unquote();
```
AThousandShips, line 4062: Ditto

scene/gui/rich_text_label.cpp
```diff
@@ -4638,8 +4638,8 @@ void RichTextLabel::append_text(const String &p_bbcode) {
 					if (sep == -1) {
 						width = bbcode_value.to_int();
 					} else {
-						width = bbcode_value.substr(0, sep).to_int();
-						height = bbcode_value.substr(sep + 1).to_int();
+						width = bbcode_value.left(sep).to_int();
+						height = bbcode_value.right(-sep - 1).to_int();
```
AThousandShips, line 4642: Ditto

scene/gui/rich_text_label.cpp
```diff
@@ -4675,28 +4675,28 @@ void RichTextLabel::append_text(const String &p_bbcode) {
...
-			int fnt_size = tag.substr(10, tag.length()).to_int();
+			int fnt_size = tag.right(-10).to_int();
 			push_font_size(fnt_size);
 			pos = brk_end + 1;
 			tag_stack.push_front("font_size");
 
 		} else if (tag.begins_with("opentype_features=") || tag.begins_with("otf=")) {
 			int value_pos = tag.find("=");
-			String fnt_ftr = tag.substr(value_pos + 1);
+			String fnt_ftr = tag.right(-value_pos - 1);
```
AThousandShips, line 4699: Ditto

AThousandShips:
It appears the changes breaks some code, see the broken doc generation

core/io/http_client.cpp
MewPurPur: I tend to agree. I think `s.substr(pos)` is the most readable, so I'm checking now if I can optimize substr(pos) by just calling right(-pos) if a second argument isn't passed.

AThousandShips:
Unit tests show their worth, let's see what's broken with your implementation of the string methods, will do some looking into it tomorrow if you haven't solved 🙂

MewPurPur:
I can't figure it out :disappointed:

kleonc:
> I think all of these are equivalent 100% of the time


These are not the same for `x < 0`:
>* `s.substr(0, x)` -> `s.left(x)` 
>* `s.substr(x, s.length())` -> `s.right(-x)`

These are not the same for `x <= 0`:
>* `s.substr(0, s.length() - x)` -> `s.left(-x)`
>* `s.substr(x, s.length() - x)` -> `s.right(-x)`

---

> but they are ~15% faster (benchmarking from my PR that optimized left and right).

The mentioned PR: #80824 (please link next time you refer to some other PR).
Its description is a little unclear to me but I understand it as the benchmarked speedup is for old `left/right` vs new `left/right`. In such case it doesn't say anything about the performance of `left/right` vs `substr`, and this would need to be benchmarked.
But I'd rather focus on trying to optimize `substr` in the first place, e.g. are the `length()` calls optimized out by the compiler and done once? If not then these should cached/ called once manually (and used in check instead of `is_empty()` etc.).
https://github.com/godotengine/godot/blob/2d0ee20ff30461b6b10f6fdfba87511a0ebc6642/core/string/ustring.cpp#L2975-L2995
Same for other methods (like `left/right`). Note it's already like that e.g. in `find`:
https://github.com/godotengine/godot/blob/2d0ee20ff30461b6b10f6fdfba87511a0ebc6642/core/string/ustring.cpp#L3002-L3004

---

Regarding code simplification these can be done for sure:
* `s.substr(x, s.length())` -> `s.substr(x)`
* `s.substr(x, s.length() - x)` -> `s.substr(x)`

I find these better readability-wise, plus it gets rid of redundant `s.length()` calls on the caller side (`length()` is called in `substr` anyway).

AThousandShips:
I'd say that unless things are in a very performance critical area readability trumps performance, I'm in favor of the obvious ones like `substr(x, length)`, but generally to stay on the safe side as `left/right` can be confusing at times, and need to ensure properly the correct conversions

I think this should be done in two PRs, one to improve the performance, and one to replace relevant cases

MewPurPur:
Hmpf, so my old PR only did ones with `left()`, I thought `right()` would be about as easy but it's not.

I know that `substr()` is confusing for a lot of people in a way similar to ternaries, where you constantly have to check if the second argument was the final position, or the substring's length (i.e. does `"ABCDEF".substr(1, 3)` give BCD or BC?) so I always prefer the dedicated methods when the substring isn't in the middle of the string. But while `left()` tends to always be readable, `right()` just sucks for readability when you flip the sign IMO.

I think I'll abandon the idea because of that.

> Its description is a little unclear to me but I understand it as the benchmarked speedup is for old left/right vs new left/right. In such case it doesn't say anything about the performance of left/right vs substr, and this would need to be benchmarked.

The old left/right were calling substr() directly, so it seems fair to me. Unless there is a notable penalty for going through an extra function that I don't know of.

AThousandShips:
I'd say to remove the potentially ambiguous cases and see if that resolves the CI, that being the `right` cases with negative indices at least

The `substr` cases with trivial length should be good IMO, as are the trivial `left` cases, i.e. `substr(0, x)`, and maybe do `right` in a future PR with more discussion and evaluation

MewPurPur:
What the heck

MewPurPur:
I'll leave this idea at peace for now.

PR #86245, opened 2023-12-16
Check skeleton RID when using cached AABB

### LINKED ISSUE #73521: Ragdoll visibility is inconsistent due to culling box problems
### Godot version

4.0 RC2

### System information

Ubuntu 22.04, Vulkan

### Issue description

I'm working on demos for Godot 4.0, and we're having problems with a flickering mesh on a ragdoll. The physical behavior of the ragdoll looks fine, but after rolling the ragdoll around a bit, the mesh suddenly disappears on certain camera angles.

The solution I found was the following:

1- throw "extra cull margin" to the maximum value (maybe max is not needed though, but it must be high enough)
2- disabling the mesh LOD so the extra cull margin won't trigger a lower resolution LOD.

After that, the ragdoll skeleton mesh stopped flickering.

### Steps to reproduce

Rolling the Ragdolls around (before the workaround) would consistently trigger this problem, as shown in the video.

[bug2.webm](https://user-images.githubusercontent.com/7095429/219772275-da4be5b9-2de5-47b3-9cd6-c185cd8daa19.webm)

https://github.com/gdquest-demos/godot-4.0-new-features

The specific demo in question is 3D Physics Nodes.

ecmjohnson:
Resolves #73521 by ensuring that the RID of the provided skeleton matches that used to compute the AABB of the mesh.

testing_fix2

[This comment](https://github.com/godotengine/godot/issues/73521#issuecomment-1858944828) presents the root cause of the issue.

### DIFF
--- a/servers/rendering/renderer_rd/storage_rd/mesh_storage.cpp
+++ b/servers/rendering/renderer_rd/storage_rd/mesh_storage.cpp
@@ -657,7 +657,8 @@ AABB MeshStorage::mesh_get_aabb(RID p_mesh, RID p_skeleton) {
 
 	Skeleton *skeleton = skeleton_owner.get_or_null(p_skeleton);
 
-	if (!skeleton || skeleton->size == 0 || mesh->skeleton_aabb_version == skeleton->version) {
+	// A mesh can be shared by multiple skeletons and we need to avoid using the AABB from a different skeleton.
+	if (!skeleton || skeleton->size == 0 || (mesh->skeleton_aabb_version == skeleton->version && mesh->skeleton_aabb_rid == p_skeleton)) {
 		return mesh->aabb;
 	}
 
@@ -763,6 +764,7 @@ AABB MeshStorage::mesh_get_aabb(RID p_mesh, RID p_skeleton) {
 	mesh->aabb = aabb;
 
 	mesh->skeleton_aabb_version = skeleton->version;
+	mesh->skeleton_aabb_rid = p_skeleton;
 	return aabb;
 }
 
--- a/servers/rendering/renderer_rd/storage_rd/mesh_storage.h
+++ b/servers/rendering/renderer_rd/storage_rd/mesh_storage.h
@@ -153,6 +153,7 @@ private:
 		AABB aabb;
 		AABB custom_aabb;
 		uint64_t skeleton_aabb_version = 0;
+		RID skeleton_aabb_rid;
 
 		Vector<RID> material_cache;
 


servers/rendering/renderer_rd/storage_rd/mesh_storage.cpp
```diff
@@ -657,7 +657,12 @@ AABB MeshStorage::mesh_get_aabb(RID p_mesh, RID p_skeleton) {
 
 	Skeleton *skeleton = skeleton_owner.get_or_null(p_skeleton);
 
-	if (!skeleton || skeleton->size == 0 || mesh->skeleton_aabb_version == skeleton->version) {
+	if (!skeleton || skeleton->size == 0) {
+		return mesh->aabb;
+	}
+
+	// A mesh can be shared by multiple skeletons and we need to avoid using the AABB from a different skeleton
+	if (mesh->skeleton_aabb_version == skeleton->version && mesh->skeleton_aabb_rid == p_skeleton) {
```
AThousandShips, line 665:
```suggestion
	// A mesh can be shared by multiple skeletons and we need to avoid using the AABB from a different skeleton.
	if (!skeleton || skeleton->size == 0 || (mesh->skeleton_aabb_version == skeleton->version && mesh->skeleton_aabb_rid == p_skeleton)) {
```
I'd say this doesn't require to different if statements doing the same thing

ecmjohnson: I feel two separate if statements is easier to read: it logically separates between meshes that aren't connected to a skeleton and the re-use of a previously computed skeleton AABB. This is more of a style thing though and if combining them in a single statement fits better with existing code I can change it

AThousandShips: I don't see it conveying anything specific, it returns the same value and separating it into different blocks makes me confused as to why the second block is separate, when it returns the same value, and makes me confused as to if for example the value is modified

ecmjohnson: Okay, I'll combine them as you show above and add the missing period at the end of the comment

clayjohn (commented):
This makes sense, but I am a bit worried about the performance result. With this change we could end up recomputing the AABB multiple times per frame as it is never cached for a skeleton/mesh pair. 

We should almost be caching on the MeshInstance level so we can guarantee that each mesh has a single skeleton. After all, this code is getting called in `_update_instance_aabb()`

clayjohn (approved):
Actually, thinking through it some more, this is probably fine for now. We only call this function when the Instance knows it needs to update the AABB because something has changed. So the assumption is that, if this is called, it will actually do an update. 

Since the AABB is cached by the Instance already, the extra level of caching here doesn't help much.

PR #86257, opened 2023-12-17
[MP] Fix `complete_auth` notifying the wrong peer

Faless:
The SceneMultiplayer `complete_auth` method was not configuring the multiplayer peer correctly, causing it to potentially send the notification to the wrong peer, on the wrong channel, and/or with an incorrect transfer mode.

The original issue was reported in the contributors chat by @TestSubject06


dsnopek (approved):
I haven't tested this but the code changes make sense to me: the target peer, channel and mode need to be set before `_send()` calls `put_packet()`, to ensure the message goes to the right place. Without this change, whatever values were previously set would be used.

YuriSizov:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

YuriSizov:
Cherry-picked for 4.1.4.

Issue #86256, opened 2023-12-17
Can't call C# static function


CsloudX:
### Tested versions

v4.2.1.stable.mono.official [b09f793f5]

### System information

Godot v4.2.1.stable.mono - Windows 10.0.17763 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2060 (NVIDIA; 31.0.15.2849) - AMD Ryzen 9 5950X 16-Core Processor (32 Threads)

### Issue description

![image](https://github.com/godotengine/godot/assets/35988767/bddbbf8c-c7ab-4fb0-80ed-b70bb08d338e)

### Steps to reproduce

.

### Minimal reproduction project (MRP)

[CSStaticTest.zip](https://github.com/godotengine/godot/files/13709935/CSStaticTest.zip)

zaevi:
This is a weird error when parsing GDScript, but the static function do works:

```GDScript
var klass = CsMain
var a = klass.F1()
print(a) # prints 123
```

YuriSizov:
@CsloudX You didn't finish uploading the MRP. Please do not click away from the template field until it is done uploading next time you submit an issue.

For this one, could you please re-upload it?

CsloudX:
Sorry about that, and I'm upload a new MRP.

zaevi:
This is caused when lookup MethodInfo from CSharpScript:

https://github.com/godotengine/godot/blob/bf8dd73e9d88d6ab146506699f43230b12c6a2b0/modules/gdscript/gdscript_analyzer.cpp#L5016-L5018

https://github.com/godotengine/godot/blob/bf8dd73e9d88d6ab146506699f43230b12c6a2b0/core/object/object.h#L238

The `id` field is never used (at lease in C#) and is always 0. So the `info` always equals to empty `MethodInfo()`

vnen:
It this fixed by https://github.com/godotengine/godot/pull/86259?

zaevi:
> It this fixed by #86259?

![image](https://github.com/godotengine/godot/assets/12966814/575cdf24-4706-4175-8bcf-c5e14280104a)



It recognizes the static function now but gives an incorrect return type.

```csharp
[GlobalClass]
public partial class Main : Node
{
	public static int Add(int a, int b) => a + b; 

	// ...
}


```

PR #86259, opened 2023-12-17
Include `name` field in MethodInfo operator ==

TitanNano:
The GDScript analyzer currently does not recognize Methods from other scripting languages as callable. This is caused by this section of the analyzer:

https://github.com/godotengine/godot/blob/2d0ee20ff30461b6b10f6fdfba87511a0ebc6642/modules/gdscript/gdscript_analyzer.cpp#L5014-L5016

Which uses either `ScriptExtension::get_method_info` which calls [`MethodInfo::from_dict`](https://github.com/godotengine/godot/blob/2d0ee20ff30461b6b10f6fdfba87511a0ebc6642/core/object/object.cpp#L133) that never sets a `MethodInfo.id`, 

https://github.com/godotengine/godot/blob/2d0ee20ff30461b6b10f6fdfba87511a0ebc6642/core/object/script_language_extension.h#L105-L109

or `CSharpScript::get_method_info` which doesn't set an ID either.

https://github.com/godotengine/godot/blob/2d0ee20ff30461b6b10f6fdfba87511a0ebc6642/modules/mono/csharp_script.cpp#L2349-L2374

Following the implementation of `operator<` I have therefore extended the equality operator to compare the `name` field as well.

(this fix is also compatible with at least 4.2 and 4.1)

### DIFF
--- a/core/object/object.h
+++ b/core/object/object.h
@@ -235,7 +235,7 @@ struct MethodInfo {
 		return arguments_metadata.size() > p_arg ? arguments_metadata[p_arg] : 0;
 	}
 
-	inline bool operator==(const MethodInfo &p_method) const { return id == p_method.id; }
+	inline bool operator==(const MethodInfo &p_method) const { return id == p_method.id && name == p_method.name; }
 	inline bool operator<(const MethodInfo &p_method) const { return id == p_method.id ? (name < p_method.name) : (id < p_method.id); }
 
 	operator Dictionary() const;


zaevi:
Related to #86256

vnen (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

raulsntos:
@akien-mga Should likely cherry-pick https://github.com/godotengine/godot/pull/86972 as well, since it's a follow-up. For context, it fixes https://github.com/godotengine/godot/issues/86256#issuecomment-1874179213.

paulloz:
This change is breaking interop for overloaded methods. I'm opening a dedicated issue.

PR #86260, opened 2023-12-17
[MP] Fix auth not waiting for confirmation in some cases

Faless:
The auth implementation was treating any received packet as a remote confirmation after the peer was confirmed locally.
    
It now correctly awaits for the remote confirmation packet before admitting new peers.


mhilbrunner (approved):
Still haven't found time to test this one, but the changes LGTM.

akien-mga:
Cherry-picked for 4.2.2.

PR #86261, opened 2023-12-17
Docs: Add details on `CanvasItem.clip_children`

bitbutter:
Replaced terms of art with more accessible language to describe 'clip children' feature.

Included the information that only direct children are affected, and mentioned the requirement that child nodes be at the same Z Index as the clipping parent for the effect to occur

### DIFF
--- a/doc/classes/CanvasItem.xml
+++ b/doc/classes/CanvasItem.xml
@@ -562,7 +562,7 @@
 	</methods>
 	<members>
 		<member name="clip_children" type="int" setter="set_clip_children_mode" getter="get_clip_children_mode" enum="CanvasItem.ClipChildrenMode" default="0">
-			Allows the current node to clip children nodes, essentially acting as a mask.
+			If one of the Clip Children modes is active, any parts of direct child nodes that extend outside this parent's visible region are hidden. The child nodes must be at the same Z Index as the parent in order for the clipping effect to occur.
 		</member>
 		<member name="light_mask" type="int" setter="set_light_mask" getter="get_light_mask" default="1">
 			The rendering layers in which this [CanvasItem] responds to [Light2D] nodes.


doc/classes/CanvasItem.xml
```diff
@@ -562,7 +562,7 @@
 	</methods>
 	<members>
 		<member name="clip_children" type="int" setter="set_clip_children_mode" getter="get_clip_children_mode" enum="CanvasItem.ClipChildrenMode" default="0">
-			Allows the current node to clip children nodes, essentially acting as a mask.
+			If one of the Clip Children modes is active, any parts of direct child nodes that extend outside this parent's visible region are hidden. The child nodes must be at the same Z Index as the parent in order for the clipping effect to occur.
```
Mickeon, line 565: Readjusting this a bit to match the rest of the class reference.
```suggestion
			The mode in which this node clips its children. If not [constant CLIP_CHILDREN_DISABLED], any parts of direct child nodes that extend outside this parent's visible region are hidden. The child nodes must be at the same Z index as the parent in order for the clipping to occur.
```

Mickeon (changes requested):

doc/classes/CanvasItem.xml
```diff
@@ -656,7 +656,7 @@
 	</methods>
 	<members>
 		<member name="clip_children" type="int" setter="set_clip_children_mode" getter="get_clip_children_mode" enum="CanvasItem.ClipChildrenMode" default="0">
-			The mode in which this node clips its children, acting as a mask.
+			The mode in which this node clips its children. If not [constant CLIP_CHILDREN_DISABLED], any parts of direct child nodes that extend outside this parent's visible region are hidden. The child nodes must be at the same Z index as the parent in order for the clipping to occur.
```
AThousandShips: "this parent" is unclear, if "this node" then that should be used, otherwise it needs to be clarified

bitbutter: yes "this node's" is better

PR #86266, opened 2023-12-17
Use `set_value_no_signal` in editor property code

aaronfranke:
In the current master, the editor property code uses a flag `bool setting` to ignore value changed signals. However, this means there were lots of useless signals being emitted. This PR replaces `set_value` with `set_value_no_signal` so there are no longer superfluous signals to guard against.

Note 1: This may slightly change behavior for `EditorPropertyQuaternion::_custom_value_changed`, which previously did not have a `bool setting` guard. However, this was likely a bug before, as you would not want multiple signals to be emitted in the middle of this method (if one is needed, it should be at the end).

~~Note 2: There is one property still using `bool setting`: `EditorPropertyEasing`. This is because it uses a LineEdit instead of a SpinBox, which has a method `set_text` but not `set_text_no_signal`.~~

Note 3: The `EditorPropertyVectorN` code is already using `set_value_no_signal`.

Note 4: I have only tested this PR for a few minutes, I don't think it needs more testing.

### DIFF
--- a/editor/editor_properties.cpp
+++ b/editor/editor_properties.cpp
@@ -1312,17 +1312,12 @@ void EditorPropertyInteger::_set_read_only(bool p_read_only) {
 }
 
 void EditorPropertyInteger::_value_changed(int64_t val) {
-	if (setting) {
-		return;
-	}
 	emit_changed(get_edited_property(), val);
 }
 
 void EditorPropertyInteger::update_property() {
 	int64_t val = get_edited_property_value();
-	setting = true;
-	spin->set_value(val);
-	setting = false;
+	spin->set_value_no_signal(val);
 #ifdef DEBUG_ENABLED
 	// If spin (currently EditorSplinSlider : Range) is changed so that it can use int64_t, then the below warning wouldn't be a problem.
 	if (val != (int64_t)(double)(val)) {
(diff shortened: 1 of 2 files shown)


YuriSizov (approved):
Makes sense.

YuriSizov:
> There is one property still using `bool setting`: `EditorPropertyEasing`. This is because it uses a LineEdit instead of a SpinBox, which has a method `set_text` but not `set_text_no_signal`.

This can be addressed by using `set_block_signal()`.

aaronfranke:
@YuriSizov No such API exists. You may be thinking of `Object::set_block_signals(bool p_block)` which isn't what we want.

But anyway, I realized that we actually don't need this guard at all, because setting a LineEdit's text does not emit a signal:

Screenshot 2023-12-18 at 11 43 21 AM

YuriSizov:
> You may be thinking of `Object::set_block_signals(bool p_block)` 

Yes, I just missed one letter.

> which isn't what we want.

Why not? It prevents the emission of signals, exactly what `*_no_signal` methods do as well. But yes, you are correct, this method doesn't emit anything anyway.

aaronfranke:
> Why not?

It would seem wrong to me to block a signal from emitting when we can instead not emit a signal at all. But anyway yeah it's not important since there is no signal anyway.

KoBeWi (approved):

YuriSizov:
Thanks!

Issue #43078, opened 2020-10-25
property_can_revert and property_get_revert are not documented features


willnationsdev:
**Godot version:**
3.2 stable

**OS/device including version:**
N/A

**Issue description:**
I noticed that this [property_can_revert](https://github.com/godotengine/godot/blob/3.2/editor/editor_inspector.cpp#L471) method exists to forcibly override the Inspector's behavior in regards to an Object's revertability of a given property. You can then see it used with [property_get_revert](https://github.com/godotengine/godot/blob/3.2/editor/editor_inspector.cpp#L679) to actually set the value upon reverting. These are not documented features and should be added to the API docs somewhere.

N/A**Steps to reproduce:**

**Minimal reproduction project:**
N/A

Xrayez:
Is this relevant for built-in classes or scripting as well?

willnationsdev:
@Xrayez 
I haven't tested it yet, but there's someone on Reddit who was encountering an issue and I suggested he give it a shot. I have a feeling it should be accessible from scripting cause it uses `has_method` and `call` on the `Object` which should defer to the script first before checking the C++ instance.

akien-mga:
For the reference, it's used (and documented) in those three classes:
```
doc/classes/ProjectSettings.xml
115:            <method name="property_can_revert">

doc/classes/EditorSettings.xml
119:            <method name="property_can_revert">

doc/classes/ShaderMaterial.xml
22:             <method name="property_can_revert">
```

It's used with `call` because it can't use this as a method since only 3 classes implement it. I don't think it was meant to be used in scripts or other types of nodes.

It *might* be usable in scripts as a consequence of implementation, but it doesn't look like it was designed to be so. We can consider changing that but it's a feature request more than missing documentation. It would imply making this a method of `Object` instead of having it implemented only in the few classes where it's currently used.

willnationsdev:
@akien-mga Ahh, that makes more sense. It'd have to change how it is used in the Inspector then, cause right now it assumes the existence of the method is, itself, an indication that it should override everything else. I suppose the logic would need to separately check if there is a script and if the script instance has the method and can call it, etc.

AnidemDex:
For now, they are useful when you work with custom classes, see https://github.com/godotengine/godot/issues/30440#issuecomment-980517507

akien-mga:
This should be done for 4.0 IMO.

[...]

PR #86268, opened 2023-12-17
Allow node visibility to work with custom user-provided node types

aaronfranke:
In the current master, the scene tree editor's visibility system is hard-coded for a few types, namely these 4: CanvasItem, CanvasLayer, Node3D, and Window. However, this is a weird restriction that hard-codes a lot of types and requires a lot of calls to `p_node->is_class()` instead of just checking if the node has what we need.

This PR changes the code to be more generic, so now users can provide their own types that have visibility:
Screenshot 2023-12-17 at 10 40 47 AM

As a bonus, this code is a lot shorter now, since we don't have to repeat code so many times for each type.

In the current master, the only node types with `is_visible` methods are CanvasItem, CanvasLayer, Node3D, and Window, so this PR does not affect the behavior of any engine-provided node types.

### DIFF
--- a/editor/gui/scene_tree_editor.cpp
+++ b/editor/gui/scene_tree_editor.cpp
@@ -103,15 +103,12 @@ void SceneTreeEditor::_cell_button_pressed(Object *p_item, int p_column, int p_i
 		undo_redo->commit_action();
 	} else if (p_id == BUTTON_LOCK) {
 		undo_redo->create_action(TTR("Unlock Node"));
-
-		if (n->is_class("CanvasItem") || n->is_class("Node3D")) {
-			undo_redo->add_do_method(n, "remove_meta", "_edit_lock_");
-			undo_redo->add_undo_method(n, "set_meta", "_edit_lock_", true);
-			undo_redo->add_do_method(this, "_update_tree");
-			undo_redo->add_undo_method(this, "_update_tree");
-			undo_redo->add_do_method(this, "emit_signal", "node_changed");
-			undo_redo->add_undo_method(this, "emit_signal", "node_changed");
-		}
+		undo_redo->add_do_method(n, "remove_meta", "_edit_lock_");
+		undo_redo->add_undo_method(n, "set_meta", "_edit_lock_", true);
+		undo_redo->add_do_method(this, "_update_tree");
+		undo_redo->add_undo_method(this, "_update_tree");
+		undo_redo->add_do_method(this, "emit_signal", "node_changed");
+		undo_redo->add_undo_method(this, "emit_signal", "node_changed");
 		undo_redo->commit_action();
 	} else if (p_id == BUTTON_PIN) {
 		if (n->is_class("AnimationMixer")) {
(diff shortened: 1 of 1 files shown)


editor/gui/scene_tree_editor.cpp
```diff
@@ -412,59 +412,37 @@ void SceneTreeEditor::_add_nodes(Node *p_node, TreeItem *p_parent) {
...
+			if (is_visible) {
 				item->add_button(0, get_editor_theme_icon(SNAME("GuiVisibilityVisible")), BUTTON_VISIBILITY, false, TTR("Toggle Visibility"));
 			} else {
 				item->add_button(0, get_editor_theme_icon(SNAME("GuiVisibilityHidden")), BUTTON_VISIBILITY, false, TTR("Toggle Visibility"));
 			}
-
-			if (!p_node->is_connected("visibility_changed", callable_mp(this, &SceneTreeEditor::_node_visibility_changed))) {
-				p_node->connect("visibility_changed", callable_mp(this, &SceneTreeEditor::_node_visibility_changed).bind(p_node));
+			if (p_node->has_signal("visibility_changed")) {
+				if (!p_node->is_connected("visibility_changed", callable_mp(this, &SceneTreeEditor::_node_visibility_changed))) {
```
YuriSizov, line 423: Nesting can be reduced.

editor/gui/scene_tree_editor.cpp
```diff
@@ -412,59 +412,37 @@ void SceneTreeEditor::_add_nodes(Node *p_node, TreeItem *p_parent) {
...
-			} else {
-				item->add_button(0, get_editor_theme_icon(SNAME("GuiVisibilityHidden")), BUTTON_VISIBILITY, false, TTR("Toggle Visibility"));
-			}
+		}
 
-			if (!p_node->is_connected("visibility_changed", callable_mp(this, &SceneTreeEditor::_node_visibility_changed))) {
-				p_node->connect("visibility_changed", callable_mp(this, &SceneTreeEditor::_node_visibility_changed).bind(p_node));
-			}
-		} else if (p_node->is_class("Node3D")) {
+		if (p_node->is_class("CanvasItem")) {
```
YuriSizov, line 430: Seems like it can be grouped with `p_node->is_class("Node3D")` to reduce the lines further. In theory, we can also just check meta itself, which would be on brand for this PR 🙃

YuriSizov (commented):
Not particularly against this change, but duck typing introduces some issues.

First of all, we can't be sure if the entire interface is implemented, so we may end up in a weird state where some parts of the API exist and do their thing, and others don't and the feature is not fully functional.

Second, these names are pretty generic. Right now the feature is only limited to a handful of classes which implement the methods and thus their descendants also implement them. This PR, for the most part, removes class checks, which means it may now affect custom classes which are not intended to be used with this feature but have methods with same names.

editor/gui/scene_tree_editor.cpp
aaronfranke: If I combine those into one `if` statement, the line gets really long and overall harder to read. However, I have an even better way to improve this code that avoids constructing multiple Callables.

aaronfranke:
> First of all, we can't be sure if the entire interface is implemented, so we may end up in a weird state where some parts of the API exist and do their thing, and others don't and the feature is not fully functional.

What parts of the interface am I missing that we should check for?

> Second, these names are pretty generic.

Indeed, visible is a fairly generic word that users could potentially use for other things. If users are using visible in their own projects for some other meaning they should probably pick a different name. However, really, showing an eyeball in the scene hierarchy is... just not a problem, it doesn't break their projects etc. Users will likely be happy to see it there, and if they don't want it there, they can just rename their method to something other than `is_visible`.

YuriSizov:
> What parts of the interface am I missing that we should check for?

You aren't missing anything, but users might. They may implement some of the methods but not the signal or other methods. You don't check for everything all the time, which would be very verbose and unacceptable, but that also means that users may implement only parts of the logic and thus the entire thing won't function properly.

And this is where my second point loops. They may have these methods or signals implemented for entirely different reasons with entirely unexpected side effects. Not to mention that discoverability of this feature would be very poor for people to even be aware of it and its repercussions.

aaronfranke:
@YuriSizov The repercussions are that "an icon shows up in the editor" which "calls the method the user provides". I don't think this is likely to be a bad thing. As for discoverability, it seems pretty obvious to me, but we can write some documentation to explain this.

Do you have an alternative suggestion for how to provide this feature to user code?

YuriSizov:
> The repercussions are that "an icon shows up in the editor" which "calls the method the user provides". I don't think this is likely to be a bad thing.

Yes, an icon would show up that calls some method defined by the user, whether the developer wanted it or not. Which may by accident make something that should be treated with care to become trivially accessible. Without any indication to the developer that there is some feature their code would hook into.

> As for discoverability, it seems pretty obvious to me, but we can write some documentation to explain this.

It is absolutely not self-explanatory that adding some methods would enable an editor hook for a feature. There is no documentation, no UI hints, and it doesn't even follow the familiar pattern of "virtual" methods that need to be implemented consciously to enable the feature. Without reading the code, or finding this PR, you wouldn't even know that there is a feature that you can use, let alone how to use it.

So documenting this would be an absolute must, especially if we go with the duck typing approach.

> Do you have an alternative suggestion for how to provide this feature to user code?

Well, that's a question for a proposal. There were a few which discussed the idea of providing way for users to add functionality/buttons to the scene dock. We should probably establish a pattern with how this functionality can be added, and decide how extensible it makes sense to make it.

I'm not ready to offer a thought-out solution for something like this because we don't really have a good system for a feature that we want to be implementable by any class extending, say, Object, without bloating Object itself. What we did previously is exactly what you're doing. And the outcome of it was people didn't know that there is something that they can use as it was never documented, not a part of the class reference or any other documentation, and very easy to break because there is no API contract. E.g. https://github.com/godotengine/godot/issues/43078

editor/gui/scene_tree_editor.cpp
```diff
@@ -412,60 +409,30 @@ void SceneTreeEditor::_add_nodes(Node *p_node, TreeItem *p_parent) {
...
+			bool is_visible = p_node->call("is_visible");
+			if (is_visible) {
 				item->add_button(0, get_editor_theme_icon(SNAME("GuiVisibilityVisible")), BUTTON_VISIBILITY, false, TTR("Toggle Visibility"));
 			} else {
 				item->add_button(0, get_editor_theme_icon(SNAME("GuiVisibilityHidden")), BUTTON_VISIBILITY, false, TTR("Toggle Visibility"));
 			}
-
-			if (!p_node->is_connected("visibility_changed", callable_mp(this, &SceneTreeEditor::_node_visibility_changed))) {
-				p_node->connect("visibility_changed", callable_mp(this, &SceneTreeEditor::_node_visibility_changed).bind(p_node));
+			if (p_node->has_signal("visibility_changed")) {
```
aaronfranke, line 419: @YuriSizov Would it be better if I move this check to the first `if`, and also add a check for `is_visible_in_tree`? That way the buttons should only show up if the user implements "the whole API". It should be extremely unlikely that the user implements all of `is_visible`, `set_visible`, `visibility_changed`, and `is_visible_in_tree` by accident.

YuriSizov: That would make it less likely to happen by accident, yes. So in that sense it would be better.

KoBeWi:
> In the current master, the only node types with is_visible methods are CanvasItem, CanvasLayer, Node3D, and Window, so this PR does not affect the behavior of any engine-provided node types.

We have StatusIndicator now, but it's missing `visibility_changed` signal.
Meantime, MultiplayerSynchronizer does have `visibility_changed` only and it's needlessly connected (not really a problem though).

> However, really, showing an eyeball in the scene hierarchy is... just not a problem, it doesn't break their projects etc.

It results in errors when used in non-tool script. Harmless, but still.

KoBeWi (approved):
The code itself looks fine, though not sure why that Callable change is here.

There are proper checks for all methods when adding the buttons, so the system can't break due to being implemented partially etc. (with the exception of non-tool scripts, but that's not really worth handling). I mentioned some classes that could be affected unintentionally, but it doesn't happen, because they don't implement full set of methods.

One potential improvement would be introducing some constants for these methods/signals, so they are not hard-coded like that.

aaronfranke:
> though not sure why that Callable change is here.

I split that into its own PR here https://github.com/godotengine/godot/pull/87778 but I hadn't rebased this PR since that was merged. I just rebased so now it's gone.

> It results in errors when used in non-tool script. Harmless, but still.

That actually might be a helpful warning to inform users that their custom node type should be a tool script in order to handle visibility. If they make the visibility code on a non-tool script, this way there is a relevant error message instead of the visibility system just not working in the editor.

PR #85885, opened 2023-12-07
Improve engine startup/shutdown benchmarks

YuriSizov:
So I was working on the project manager and wanted to learn why does it take a whopping 3 seconds to load, with most time spent outside of PM code and in other `main.cpp` activities instead. As I did before, I tried to use our existing benchmarking system that is supposed to track startup and shutdown processes and report on time taken.

I have already discovered previously that it is not a very useful system. Many things are not tracked correctly, with begin or end points of measurement being misplaced; many other things are not tracked at all and missing. Those which are tracked are reported in a disorganized manner and are hard to make use of. For example, most of the time of the PM startup is taken by "servers", and that's pretty much all that I can tell from our current benchmark:

```
BENCHMARK:
	- register_core_types :  0.014811  sec.
	- servers :  2.620844  sec.
	- register_editor_types :  0.003172  sec.
	- scene :  0.209155  sec.
	- editor_register_and_generate_icons_all :  0.254266  sec.
	- editor_register_and_generate_icons_with_only_thumbs :  0.003362  sec.
	- editor_register_fonts :  0.005924  sec.
	- create_editor_theme :  0.606796  sec.
	- create_custom_theme :  0.606828  sec.
	- project_manager :  0.782145  sec.
	- startup_begin :  3.649693  sec.
	- unregister_editor_types :  0.000009  sec.
	- unregister_core_types :  0.02886  sec.
	- Main::cleanup :  0.118794  sec.
```

_Measured on battery with a build of 2f73a059cefadcd944b6874f2557ec82e46a562d_

So I set to try and improve all of that. First of all, I added grouping to the measurements. This makes the data structure a bit more complex, which I guess will increase the impact of using the benchmark on the benchmarked code. But what are you going to do, the observer is a part of the observed process.

I went with a Pair key for intermediate records instead of using nested HashMaps, because I think it would be slightly faster. But I didn't... benchmark it. I also replaced the Dictionary with a HashMap. I think it should be more efficient too, with no Variant marshalling required. We only needed a Dictionary when creating the JSON file, but we can fabricate one when we do the dump, which is one part of the process where it's okay to be slow.

I then went through benchmarked areas and adjusted/added measure points to make sure they track sensible data. There shouldn't be anything controversial about these changes here.

[...]

PR #5993, opened 2016-07-31
Improved drad&drop files on project manager

profelis:
Improved dnd in project manager. Dragged any file from project folder or single project folder autoimport it, otherwise confirm scan all dragger folders.

Regarding #5981, @akien-mga please review.

Issue #5953, opened 2016-07-29
Project manager scan should run in a thread to keep responsive UI


Biarity:
**Operating system or device - Godot version:**
Windows 10, Godot 2.1rc1

**Issue description** (what happened, and what was expected):
Using the scan function at the project manager crashes Godot when I click 'open'.

**Steps to reproduce:**
Open Godot, click Scan, select a large folder, click Open. Shows the Windows Close/Debug error box.

vnen:
I cannot reproduce this. The project manager stops responding while it is searching for files but if I leave it running it manages to fill the list.

Maybe moving the search to a separate thread might be good so the editor keeps responsive while searching.

Qudware:
Try one of the folders in C:\Users on Windows. Other folders on my Desktop seem to work fine.

vnen:
For me it just hangs for a long time scanning the folders. On the Downloads folder (which has just a few subfolders) it finishes pretty much instantly.

bojidar-bg:
Any chance this is related to stack overflows?

vnen:
AFAIK it does not "crash" it just stops responding. If you click on the window then Windows will show the problem and asks if you want to force close it, which can be mistaken by a crash.

I still think it's just a matter of moving the scan to a new thread to show some progress and keep the editor responsive.

akien-mga:
> I still think it's just a matter of moving the scan to a new thread to show some progress and keep the editor responsive.

Edited issue title accordingly.

Calinou:
Note that https://github.com/godotengine/godot/pull/30735 doesn't implement this.

If we want a simple solution to this, we can dim the window while projects are being scanned (like when quitting the editor or project manager). This will at least provide some kind of visual feedback without complexifying the code too much.

KoBeWi:
This doesn't need to run in thread actually. We could just utilize `_process()` method and e.g. scan 20 directories each frame. Easy to do and even allows cancelling, so it's more UX problem.

Zylann:
If this is done in process, then it should be staggered (i.e keep scanning until 10 ms elapsed then continue next frame) because filesystem access can vary wildly across computers, hardware and OSes. On the other hand, with sliiightly more work, using a thread may actually be a more efficient approach, and it can be cancelled even just with a boolean.
I did the staggered approach for Find in Files as well but I'd like to switch to a thread one day for larger projects.

akien-mga:
[...]

PR #86271, opened 2023-12-17
Optimize scanning routines in the project manager

YuriSizov:
I previously worked on improving built-in benchmarking capabilities in https://github.com/godotengine/godot/pull/85885 in hopes to get a better picture of the startup times in the project manager. Unfortunately, most of the time is eaten by DisplayServer and other main setup, and not by the project manager itself.

I couldn't find much that could be immediately improved, but I did notice that we weren't using some of the method related to scanning/project list updates efficiently. In fact the method that is specifically commented with a performance notice could be called significantly more often than it was needed. So this PR looks into such cases and tries to improve them a bit.

Two things I've identified:

- During the PM initialization we could trigger 2 expensive updates of the project list because we first would refresh it normally, and then refresh it again if we have the auto-scan folder configured.
- Scanning multiple folders, e.g. when you drag'n'drop them into the editor, was following a suboptimal path that triggered the full expensive refresh for every dropped folder.

Both of these cases are addressed now, and in both of them we now only do one expensive refresh. The refresh itself should likely be optimized as well, but that's a bit out of scope of what I'm trying to achieve here right now.

I also moved the relevant methods from `ProjectManager` to `ProjectList` because, like many other methods currently placed in `ProjectManager`, they were nothing but shells for calling the project list APIs. So moving them there makes for better organization (something I hope to tackle more generally soon). Some things were adjusted to allow for this change, but the overall logic should be exactly the same as before.

-----

The only change in logic that I decided to do relates to the confirmation dialog when you drop multiple folders for scanning. I removed the checks and the dialog completely, and my reasoning here is that the check itself wasn't really saving the user any time. It was added in https://github.com/godotengine/godot/pull/5993, long long time ago, and from the discussion the intent seems vague to me. The way it worked was as follows:

- If there is only one dropped folder and that folder contains a Godot project (identified by the presence of `project.godot`), don't show a confirmation dialog.
- In any other case show the dialog, and let the user to opt out of doing the scanning and importing.
- If scanning is performed, the steps are exactly the same and all the dropped folders and their subfolders are recursively checked for Godot projects.

So not showing the dialog when we drop one folder and it's a folder with a Godot project makes sense, we want this to work quickly. However, we don't do it quickly in practice, because the entire folder structure is scanned anyway. Which can take time for a bigger existing project. If the idea would be to immediately import the project instead of scanning, then this is not how it works in practice.

On the other hand, showing a confirmation dialog ahead of scanning makes sense to warn the user about it being slow. But that is not really dependent on how many folders we have selected and dropped. One folder with an expansive internal structure can take way longer than 10 folders which are shallow. We also don't show the dialog when you select the folder for scanning by other means.

So given all that I just removed the confirmation. We just scan all the dropped folders immediately. And scanning is a bit faster now thanks to the rest of this PR, so it's less of an issue. Here's a comparison, btw, of me dragging and dropping 2 folders full of Godot projects before and after this PR (the confirmation dialog times not included, of course):

```
BEFORE:

	[Project Manager]
		- Multiple Scans: 538.942 msec.
	[Project Manager]
		- Multiple Scans: 521.683 msec.
	[Project Manager]
		- Multiple Scans: 521.654 msec.

AFTER:

	[Project Manager]
		- Multiple Scans: 378.665 msec.
	[Project Manager]
		- Multiple Scans: 407.250 msec.
	[Project Manager]
		- Multiple Scans: 387.383 msec.
```

This is on a fast drive and on a fast laptop that is plugged in. And that's about 80 imported projects with all their subfolders checked. I think it's representative enough to say that scans such as this can be 25% faster now. Although not sure how often people do that in practice.

-----

I am marking this as cherry-pickable. I specifically tried to tackle this before doing more extensive changes, and I think this may be desirable for current stable versions.

### DIFF
--- a/editor/project_manager.cpp
+++ b/editor/project_manager.cpp
@@ -1249,6 +1249,7 @@ struct ProjectListComparator {
 	}
 };
 
+const char *ProjectList::SIGNAL_LIST_CHANGED = "list_changed";
 const char *ProjectList::SIGNAL_SELECTION_CHANGED = "selection_changed";
 const char *ProjectList::SIGNAL_PROJECT_ASK_OPEN = "project_ask_open";
 
@@ -1356,7 +1357,7 @@ ProjectList::Item ProjectList::load_project_data(const String &p_path, bool p_fa
 	return Item(project_name, description, tags, p_path, icon, main_scene, unsupported_features, last_edited, p_favorite, grayed, missing, config_version);
 }
 
-void ProjectList::migrate_config() {
+void ProjectList::_migrate_config() {
 	// Proposal #1637 moved the project list from editor settings to a separate config file
 	// If the new config file doesn't exist, populate it from EditorSettings
 	if (FileAccess::exists(_config_path)) {
@@ -1388,9 +1389,10 @@ void ProjectList::migrate_config() {
 	save_config();
 }
 
-void ProjectList::load_projects() {
+void ProjectList::update_project_list() {
 	// This is a full, hard reload of the list. Don't call this unless really required, it's expensive.
 	// If you have 150 projects, it may read through 150 files on your disk at once + load 150 icons.
+	// FIXME: Does it really have to be a full, hard reload? Runtime updates should be made much cheaper.
 
 	// Clear whole list
 	for (int i = 0; i < _projects.size(); ++i) {
@@ -1421,6 +1423,7 @@ void ProjectList::load_projects() {
 	update_dock_menu();
 
 	set_v_scroll(0);
+	emit_signal(SNAME(SIGNAL_LIST_CHANGED));
 }
 
 void ProjectList::update_dock_menu() {
@@ -1678,6 +1681,48 @@ void ProjectList::erase_missing_projects() {
 	save_config();
 }
 
+void ProjectList::_scan_folder_recursive(const String &p_path, List<String> *r_projects) {
+	Ref<DirAccess> da = DirAccess::create(DirAccess::ACCESS_FILESYSTEM);
+	Error error = da->change_dir(p_path);
+	ERR_FAIL_COND_MSG(error != OK, vformat("Failed to open the path \"%s\" for scanning (code %d).", p_path, error));
+
+	da->list_dir_begin();
+	String n = da->get_next();
+	while (!n.is_empty()) {
+		if (da->current_is_dir() && n[0] != '.') {
+			_scan_folder_recursive(da->get_current_dir().path_join(n), r_projects);
+		} else if (n == "project.godot") {
+			r_projects->push_back(da->get_current_dir());
+		}
+		n = da->get_next();
+	}
+	da->list_dir_end();
+}
+
+void ProjectList::find_projects(const String &p_path) {
+	PackedStringArray paths = { p_path };
+	find_projects_multiple(paths);
+}
+
+void ProjectList::find_projects_multiple(const PackedStringArray &p_paths) {
+	List<String> projects;
+
+	for (int i = 0; i < p_paths.size(); i++) {
+		const String &base_path = p_paths.get(i);
+		print_verbose(vformat("Scanning for projects in \"%s\".", base_path));
+
+		_scan_folder_recursive(base_path, &projects);
+		print_verbose(vformat("Found %d project(s).", projects.size()));
+	}
+
+	for (const String &E : projects) {
+		add_project(E, false);
+	}
+
+	save_config();
+	update_project_list();
+}
+
 int ProjectList::refresh_project(const String &dir_path) {
 	// Reloads information about a specific project.
 	// If it wasn't loaded and should be in the list, it is added (i.e new project).
@@ -1916,6 +1961,7 @@ void ProjectList::_show_project(const String &p_path) {
 }
 
 void ProjectList::_bind_methods() {
+	ADD_SIGNAL(MethodInfo(SIGNAL_LIST_CHANGED));
 	ADD_SIGNAL(MethodInfo(SIGNAL_SELECTION_CHANGED));
 	ADD_SIGNAL(MethodInfo(SIGNAL_PROJECT_ASK_OPEN));
 }
@@ -1926,6 +1972,7 @@ ProjectList::ProjectList() {
 	add_child(_scroll_children);
 
 	_config_path = EditorPaths::get_singleton()->get_data_dir().path_join("projects.cfg");
+	_migrate_config();
 }
 
 /// Project Manager.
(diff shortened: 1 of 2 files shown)


editor/project_manager.cpp
```diff
@@ -2183,15 +2230,6 @@ void ProjectManager::shortcut_input(const Ref<InputEvent> &p_ev) {
 	}
 }
 
-void ProjectManager::_load_recent_projects() {
-	_project_list->set_search_term(search_box->get_text().strip_edges());
```
YuriSizov, line 2187: I'm not really sure why we used to do this before updating the project list. Changing the filter already sets the search term, and during the first scan we don't have any search term anyway, it's all empty.

So this bit I simply removed even before removing the method itself. I tested and didn't see issues. The search term was still applied after scanning and importing dropped folders.

KoBeWi:
The only problem with scanning big folders is that Project Manager does not show any progres whatsoever (#5953). User might pick something very big for scan and ends up with unresponsive window.
Though scanning times are fine from my testing. 100+ are founds within seconds in a dev build, which is acceptable. But I did not test scanning whole drive 🙃

editor/project_manager.cpp
```diff
@@ -1249,6 +1249,7 @@ struct ProjectListComparator {
 	}
 };
 
+const char *ProjectList::SIGNAL_LIST_CHANGED = "list_changed";
 const char *ProjectList::SIGNAL_SELECTION_CHANGED = "selection_changed";
 const char *ProjectList::SIGNAL_PROJECT_ASK_OPEN = "project_ask_open";
```
KoBeWi, line 1254: Unrelated, but these should be StringNames (probably not static then).

editor/project_manager.h
```diff
@@ -288,12 +291,15 @@ class ProjectList : public ScrollContainer {
...
+	static const char *SIGNAL_LIST_CHANGED;
 	static const char *SIGNAL_SELECTION_CHANGED;
 	static const char *SIGNAL_PROJECT_ASK_OPEN;
 
-	void load_projects();
+	void update_project_list();
 	int get_project_count() const;
 
+	void find_projects(const String &p_path);
+	void find_projects_multiple(PackedStringArray p_paths);
```
KoBeWi, line 302:
```suggestion
	void find_projects_multiple(const PackedStringArray &p_paths);
```

editor/project_manager.cpp
```diff
@@ -1678,6 +1681,48 @@ void ProjectList::erase_missing_projects() {
...
+	PackedStringArray paths = { p_path };
+	find_projects_multiple(paths);
+}
+
+void ProjectList::find_projects_multiple(PackedStringArray p_paths) {
+	List<String> projects;
+
+	for (int i = 0; i < p_paths.size(); i++) {
+		const String base_path = p_paths.get(i);
+		print_line(vformat("Scanning for projects in \"%s\".", base_path));
```
KoBeWi, line 1712: Maybe this should be verbose only?

editor/project_manager.cpp
```diff
@@ -1678,6 +1681,48 @@ void ProjectList::erase_missing_projects() {
...
+void ProjectList::find_projects(const String &p_path) {
+	PackedStringArray paths = { p_path };
+	find_projects_multiple(paths);
+}
+
+void ProjectList::find_projects_multiple(PackedStringArray p_paths) {
+	List<String> projects;
+
+	for (int i = 0; i < p_paths.size(); i++) {
+		const String base_path = p_paths.get(i);
```
KoBeWi, line 1711:
```suggestion
		const String &base_path = p_paths.get(i);
```

editor/project_manager.cpp
```diff
@@ -1678,6 +1681,48 @@ void ProjectList::erase_missing_projects() {
 	save_config();
 }
 
+void ProjectList::_scan_folder_recursive(const String &p_path, List<String> *r_projects) {
+	Ref<DirAccess> da = DirAccess::create(DirAccess::ACCESS_FILESYSTEM);
+	Error error = da->change_dir(p_path);
+	ERR_FAIL_COND_MSG(error != OK, vformat("Failed to open the path \"%s\" for scanning with code %d.", p_path, error));
```
KoBeWi, line 1687:
```suggestion
	ERR_FAIL_COND_MSG(error != OK, vformat("Failed to open the path \"%s\" for scanning. Error code %d.", p_path, error));
```
"scanning with code" sounds ambiguous.

editor/project_manager.cpp
```diff
@@ -2866,8 +2847,9 @@ ProjectManager::ProjectManager() {
 	vb->add_child(center_box);
 
 	tabs = memnew(TabContainer);
-	center_box->add_child(tabs);
 	tabs->set_anchors_and_offsets_preset(Control::PRESET_FULL_RECT);
+	tabs->set_current_tab(0);
```
KoBeWi, line 2851: What's this line for?

KoBeWi (approved):
Left some style comments, but the overall logic is ok.

editor/project_manager.cpp
YuriSizov: Yeah, I tried making them static StringNames, but it didn't quite work. I didn't want to spend time adding the usual framework to make them accessible outside of the class with this PR, so I just followed the existing pattern. But I will rework this in a future PR. We probably can have a "project manager string names" file, like we do with other contexts.

editor/project_manager.cpp
YuriSizov: I thought about it as well. Will change.

editor/project_manager.cpp
YuriSizov: It's "Failed ... with code", we use this structure elsewhere (though, admittedly, it was I who added it elsewhere, probably)

editor/project_manager.cpp
YuriSizov: I don't know, it was a part of the old `_load_recent_projects` method, but it didn't make sense to me there either. The only context where I thought it might be applicable is the initialization, so I moved it here. But I can remove it.

editor/project_manager.cpp
YuriSizov: Rephrased it a bit.

editor/project_manager.cpp
```diff
@@ -1678,6 +1681,48 @@ void ProjectList::erase_missing_projects() {
...
 
+void ProjectList::_scan_folder_recursive(const String &p_path, List<String> *r_projects) {
+	Ref<DirAccess> da = DirAccess::create(DirAccess::ACCESS_FILESYSTEM);
+	Error error = da->change_dir(p_path);
+	ERR_FAIL_COND_MSG(error != OK, vformat("Failed to open the path \"%s\" for scanning with code %d.", p_path, error));
+
+	da->list_dir_begin();
+	String n = da->get_next();
+	while (!n.is_empty()) {
+		if (da->current_is_dir() && !n.begins_with(".")) {
```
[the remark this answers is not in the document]
YuriSizov: There are two more places that use this, but they are in `ProjectDialog`, and I don't want to create unnecessary conflicts with Nathan's PR.

YuriSizov:
Thanks for reviews!

YuriSizov:
Cherry-picked for 4.2.2.

Issue #86249, opened 2023-12-17
String::sha256_text() shows up as a hotspot during game launch


pcbeard:
### Tested versions

Reproducible in:  Godot v4.2.stable - macOS 14.3.0 - Vulkan (Forward+) - integrated Apple M1 Pro - Apple M1 Pro (10 Threads)

### System information

Godot v4.2.stable - macOS 14.3.0 - Vulkan (Forward+) - integrated Apple M1 Pro - Apple M1 Pro (10 Threads)

### Issue description

I profiled my app's startup performance and noticed a hot spot in this call sequence:

8	String::utf8() const	
5	String::sha256_text() const	
5	ShaderRD::setup(char const*, char const*, char const*, char const*)	
2	Fsr2AccumulatePassShaderRD::Fsr2AccumulatePassShaderRD()	
2	Fsr2AccumulatePassShaderRD::Fsr2AccumulatePassShaderRD()	
2	RendererRD::FSR2Effect::'unnamed'::()	
2	RendererRD::FSR2Effect::'unnamed'::()	
2	RendererRD::FSR2Effect::FSR2Effect()	
2	RendererRD::FSR2Effect::FSR2Effect()	
2	RendererSceneRenderImplementation::RenderForwardClustered::RenderForwardClustered()	
2	RendererSceneRenderImplementation::RenderForwardClustered::RenderForwardClustered()	
2	RendererCompositorRD::RendererCompositorRD()	
2	RendererCompositorRD::RendererCompositorRD()	
2	RendererCompositorRD::_create_current()	
2	RendererCompositor::create()	
2	RenderingServerDefault::RenderingServerDefault(bool)	
2	RenderingServerDefault::RenderingServerDefault(bool)	
2	Main::setup2()	
2	Main::setup(char const*, int, char**, bool)	

Inspecting `String::sha256_text()` we see it's converting it's character data to utf8 before computing the sha256 value:

```c++
String String::sha256_text() const {
	CharString cs = utf8();
	unsigned char hash[32];
	CryptoCore::sha256((unsigned char *)cs.ptr(), cs.length(), hash);
	return String::hex_encode_buffer(hash, 32);
}
```

[...]

PR #86274, opened 2023-12-17
Change hashing functions to operate on char32_t data directly

pcbeard:
This avoids converting character data to utf8, which costs memory and time.

This addresses https://github.com/godotengine/godot/issues/86249.

### DIFF
--- a/core/string/ustring.cpp
+++ b/core/string/ustring.cpp
@@ -2884,30 +2884,26 @@ uint64_t String::hash64() const {
 }
 
 String String::md5_text() const {
-	CharString cs = utf8();
 	unsigned char hash[16];
-	CryptoCore::md5((unsigned char *)cs.ptr(), cs.length(), hash);
+	CryptoCore::md5((unsigned char *)ptr(), length() * sizeof(char32_t), hash);
 	return String::hex_encode_buffer(hash, 16);
 }
 
 String String::sha1_text() const {
-	CharString cs = utf8();
 	unsigned char hash[20];
-	CryptoCore::sha1((unsigned char *)cs.ptr(), cs.length(), hash);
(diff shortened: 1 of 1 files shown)


AThousandShips (commented):
While we don't technically guarantee hash stability between versions we try to avoid breaking it, and this will *definitely* break the hash function stability as hashing the utf32 and utf8 values are *not* equivalent, I don't think this is a good solution to this

Edit: In fact we *do* have unit tests for hashes expecting a particular result, so I'd say this isn't an acceptable solution, having unit tests expecting a particular value implies that we very much would like to have stable hashes between versions, unless absolutely necessary

This will, for example, wipe out this data in a project when applying this PR:
* Any imported file will be re-imported as the md5 hash is used to check the file for updates
* Any shader will be recompiled

AThousandShips (changes requested):
As per the above

pcbeard:
Yay, tests!

AThousandShips:
One solution, which might be decently performant, would be to do the hashing in steps, with a temporary buffer, like so:
```cpp
CryptoCore::MD5Context ctx;
ctx.start();
unsigned char buf[4 * n];
unsigned char hash[16];
for (int i = 0; i < length(); i += n) {
    // Do utf32 to utf8 conversion on n characters, into buf
    ctx.update(buf, 16);
}
ctx.finish(hash);
return String::hex_encode_buffer(hash, 16);
```
Or something similar

pcbeard:
Thanks for looking into this.

AThousandShips:
We do encode up to 6 bytes of utf8 so it'd have to support that, the overlong sequences, but will try and take a look at the md5 and other hashers and see if they behave nicely with being fed very short input, I suspect that they just build up a buffer and process when they have sufficient input so it should be okay

PR #86277, opened 2023-12-17
Always call `set_animation` in `set_sprite_frames`

### LINKED ISSUE #86058: The frame duration setting of the first frame of AnimatedSprite2D is ignored in the first loop
### Tested versions

4.2.stable

### System information

Windows 11

### Issue description

If you set the frame duration to a number greater than 1 on the first frame of an AnimatedSprite2D's "default" animation, it will ignore the frame duration setting on the first loop when the animation plays.

This only happens for the animation named "default".

### Steps to reproduce

1. Add AnimatedSprite2D
2. Add SpriteFrames
3. Add frames from sprite sheet
4. Set "Autoplay on load"
5. Set "Frame duration" to a number greater than 1 on the first frame of the "default" animation
6. Save the scene and play

[AnimatedSprite2DBug.zip](https://github.com/godotengine/godot/files/13642745/AnimatedSprite2DBug.zip)

rileylyman:
Fixes #86058.

Before this patch, say that an `AnimatedSprite2D` called `a` currently has the active animation of `"foo"`. If we update `a`'s sprite frames and the new sprite frames also have an animation called `"foo"`, then `a` will not call `set_animation("foo")` and so will continue to execute the new `"foo"` animation as if it was the original `"foo"` animation. It will continue playing from the same frame with the same speed, even if the new `"foo"` has fewer frames or should last for a different duration.

This manifested in #86058 with the `"default"` animation. The first frame of the `"default"` animation in this case was supposed to last for a duration of `10`, but only lasted for `1` since we never called `set_animation("default")` with the new sprite frames and so the `AnimatedSprite2D` just used its default values until the next frame.

I also think we could get the engine to error by writing some sort of script that swapped between two sets of sprite frames with the same animation name but different frame lengths, but I didn't try it. This change would fix that as well.

### DIFF
--- a/scene/2d/animated_sprite_2d.cpp
+++ b/scene/2d/animated_sprite_2d.cpp
@@ -299,16 +299,13 @@ void AnimatedSprite2D::set_sprite_frames(const Ref<SpriteFrames> &p_frames) {
 
 		List<StringName> al;
 		frames->get_animation_list(&al);
-		if (al.size() == 0) {
-			set_animation(StringName());
+
+		animation = StringName();
+		if (!al.is_empty()) {
+			set_animation(frames->has_animation(animation) ? animation : al[0]);
+		}
+		if (!frames->has_animation(autoplay)) {
 			autoplay = String();
-		} else {
-			if (!frames->has_animation(animation)) {
-				set_animation(al[0]);
-			}
-			if (!frames->has_animation(autoplay)) {
-				autoplay = String();
-			}
 		}
 	}
 
--- a/scene/3d/sprite_3d.cpp
+++ b/scene/3d/sprite_3d.cpp
@@ -1112,16 +1112,13 @@ void AnimatedSprite3D::set_sprite_frames(const Ref<SpriteFrames> &p_frames) {
 
 		List<StringName> al;
 		frames->get_animation_list(&al);
-		if (al.size() == 0) {
-			set_animation(StringName());
+
+		animation = StringName();
+		if (!al.is_empty()) {
+			set_animation(frames->has_animation(animation) ? animation : al[0]);
+		}
+		if (!frames->has_animation(autoplay)) {
 			autoplay = String();
-		} else {
-			if (!frames->has_animation(animation)) {
-				set_animation(al[0]);
-			}
-			if (!frames->has_animation(autoplay)) {
-				autoplay = String();
-			}
 		}
 	}
 


scene/2d/animated_sprite_2d.cpp
```diff
@@ -299,16 +299,13 @@ void AnimatedSprite2D::set_sprite_frames(const Ref<SpriteFrames> &p_frames) {
 
 		List<StringName> al;
 		frames->get_animation_list(&al);
-		if (al.size() == 0) {
-			set_animation(StringName());
+
+		animation = StringName();
+		if (al.size() > 0) {
```
dalexeev, line 304:
```suggestion
		if (!al.is_empty()) {
```

dalexeev:
The changes look good to me. Please make the same changes to `AnimatedSprite3D::set_sprite_frames()` and correct the commit message (it must start with a capital letter, see [CONTRIBUTING.md](https://github.com/godotengine/godot/blob/master/CONTRIBUTING.md)).

https://github.com/godotengine/godot/blob/ed4b82c665221b6fd3c8629d563dfb22be02fa19/scene/3d/sprite_3d.cpp#L1113-L1125

scene/2d/animated_sprite_2d.cpp
```diff
@@ -299,16 +299,13 @@ void AnimatedSprite2D::set_sprite_frames(const Ref<SpriteFrames> &p_frames) {
 
 		List<StringName> al;
 		frames->get_animation_list(&al);
-		if (al.size() == 0) {
-			set_animation(StringName());
+
+		animation = StringName();
+		if (!al.is_empty()) {
+			set_animation(frames->has_animation(animation) ? animation : al[0]);
```
AThousandShips, line 305:
```suggestion
			set_animation(frames->has_animation(animation) ? animation : al.front());
```

scene/3d/sprite_3d.cpp
```diff
@@ -1112,16 +1112,13 @@ void AnimatedSprite3D::set_sprite_frames(const Ref<SpriteFrames> &p_frames) {
 
 		List<StringName> al;
 		frames->get_animation_list(&al);
-		if (al.size() == 0) {
-			set_animation(StringName());
+
+		animation = StringName();
+		if (!al.is_empty()) {
+			set_animation(frames->has_animation(animation) ? animation : al[0]);
```
AThousandShips, line 1118:
```suggestion
			set_animation(frames->has_animation(animation) ? animation : al.front());
```

AThousandShips (commented):
Cleaner and avoids using index on a list (even though it's equivalent it's better to avoid it)

scene/3d/sprite_3d.cpp
rileylyman: I tried it, it turned out I needed `al.front()->get()` instead, which I find a lot less clear and the same amount of scary since it requires a ptr deref, so i would prefer to keep `al[0]`

AThousandShips: You can instead do:
```suggestion
			set_animation(frames->has_animation(animation) ? animation : *al.front());
```
Which should be clearer, but it's either way 🙂

PR #86285, opened 2023-12-18
Avoid rescanning when copying pure-ASCII code points

pcbeard:
This refactors `String::utf8()` to call  static function _utf8_length() to compute the decoded utf8 string, and if that length is the same as .length(), assumes the code points are all ASCII, and take a simpler/faster code path to copy the characters.

This addresses https://github.com/godotengine/godot/issues/86249 by reducing the cost of `String::utf8()` when the text is pure ASCII.

### DIFF
--- a/core/string/ustring.cpp
+++ b/core/string/ustring.cpp
@@ -1976,14 +1976,9 @@ Error String::parse_utf8(const char *p_utf8, int p_len, bool p_skip_cr) {
 	}
 }
 
-CharString String::utf8() const {
-	int l = length();
-	if (!l) {
-		return CharString();
-	}
-
-	const char32_t *d = &operator[](0);
-	int fl = 0;
+static int _utf8_length(const String &s) {
+	int fl = 0, l = s.length();
+	const char32_t *d = s.ptr();
 	for (int i = 0; i < l; i++) {
 		uint32_t c = d[i];
 		if (c <= 0x7f) { // 7 bits.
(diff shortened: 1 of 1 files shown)


core/string/ustring.cpp
```diff
@@ -1996,24 +1991,44 @@ CharString String::utf8() const {
...
 	uint8_t *cdst = (uint8_t *)utf8s.get_data();
 
+	if (is_ascii) {
+		while (l--) {
+			*cdst++ = (uint8_t)*d++;
+		}
+		*cdst = '\0';
+		return utf8s;
+	}
+
```
pcbeard, line 2031: I wonder if there are SIMD optimized solutions for this too.

core/string/ustring.cpp
```diff
@@ -1976,14 +1976,9 @@ Error String::parse_utf8(const char *p_utf8, int p_len, bool p_skip_cr) {
...
 
-CharString String::utf8() const {
-	int l = length();
-	if (!l) {
-		return CharString();
-	}
-
-	const char32_t *d = &operator[](0);
-	int fl = 0;
+static int _utf8_length(const String &s) {
```
bruvzg, line 1979:
```suggestion
static _FORCE_INLINE_  int _utf8_length(const String &s) {
```

I'm not sure if I see any reason to split it to the separate function, it's used only one.

bruvzg (commented):
Would be nice to see real performance measurements.

If it's faster, the same probably can be done with `utf16` and `parse_utf*` (all use the same double pass process).

pcbeard:
I did profile with this change, and `String::utf8()` is no longer the hot spot during all this hashing. The crypto functions become the hot spot.

PR #86290, opened 2023-12-18
[macOS] Fix updating editor tab titles in the dock menu.

bruvzg:
Fixes errors cause by `update_tab_titles` trying to update non-existing global menu items.

### DIFF
--- a/editor/gui/editor_scene_tabs.cpp
+++ b/editor/gui/editor_scene_tabs.cpp
@@ -191,12 +191,14 @@ void EditorSceneTabs::_disable_menu_option_if(int p_option, bool p_condition) {
 }
 
 void EditorSceneTabs::update_scene_tabs() {
+	static bool menu_initialized = false;
 	tab_preview_panel->hide();
 
-	if (scene_tabs->get_tab_count() == EditorNode::get_editor_data().get_edited_scene_count()) {
+	if (menu_initialized && scene_tabs->get_tab_count() == EditorNode::get_editor_data().get_edited_scene_count()) {
 		_update_tab_titles();
 		return;
 	}
+	menu_initialized = true;
 
 	if (DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_GLOBAL_MENU)) {
 		DisplayServer::get_singleton()->global_menu_clear("_dock");


YuriSizov:
I'm not entirely sure what's going on here. Where do these non-existent items come from? I think we should fix that instead.

Also, if I understand the hack correctly, adding and removing 1 would be sufficient, it doesn't have to be 100? We just need to differentiate between actual index 0 and nothing being cast to 0?

bruvzg:
> Where do these non-existent items come from? I think we should fix that instead.

`update_tab_titles` can be called without full update (which is adding menu items, now both function can add items).

> Also, if I understand the hack correctly, adding and removing 1 would be sufficient, it doesn't have to be 100?

Yes, it's only to detect empty meta (which is cast to zero).

YuriSizov:
> `update_tab_titles` can be called without full update (which is adding menu items, now both function can add items).

How do we get into this situation? The point of separating update methods was to avoid calling some of the routines if the number of tabs didn't change. If we call `update_tab_titles` when we actually need to do a full update, that seems to be a bug.

kitbdev:
`_update_tab_titles()` is only called when the number of tabs is equal to the number of scenes. The number of global menu items should always be the same as the number of tabs, since it is updated in the same place, in `update_scene_tabs()`.

The only scenario I can think of that would cause the counts to not be the same is if `DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_GLOBAL_MENU)` can return different values between calls. I don't think this is possible though. 
Or is it possible that `global_menu_add_item` fails?
I don't have a mac to test, so I might be missing something.

bruvzg:
Actually the fix is much simpler, the core reason: first tab was added before connecting signals and most of the editor initialized, so first update is always called when it's already 1 == 1.

editor/gui/editor_scene_tabs.cpp
```diff
@@ -191,9 +191,11 @@ void EditorSceneTabs::_disable_menu_option_if(int p_option, bool p_condition) {
 }
 
 void EditorSceneTabs::update_scene_tabs() {
+	static bool menu_initialized = false;
 	tab_preview_panel->hide();
 
-	if (scene_tabs->get_tab_count() == EditorNode::get_editor_data().get_edited_scene_count()) {
+	if (menu_initialized && scene_tabs->get_tab_count() == EditorNode::get_editor_data().get_edited_scene_count()) {
+		menu_initialized = true;
```
YuriSizov, line 198: Shouldn't this be set after the check? Otherwise this condition can never be reached.

YuriSizov (approved):
Yeah, that makes sense now. Thanks for looking into this!

YuriSizov:
Thanks!

PR #83648, opened 2023-10-19
Add optimization for generating tangents in mikktspace.

fire:
genTangSpace was taking 70% of the time, we want to reduce it to 30% of the time.

## Test Plan

Test a project with three copies of the same load. Compare before and after. Check correctness.

_Production edit: closes godotengine/godot-roadmap#41_

PR #86293, opened 2023-12-18
Unify tangents in CSGShape

fire:
Part of https://github.com/godotengine/godot-proposals/issues/8601.

Part of https://github.com/godotengine/godot/pull/83648.

Ready for review.

----

1. ~~Enforces indexing before generating tangents for efficiency.~~
2. Changes CSGShape to use SurfaceTool::generate_tangents();
3. Needs performance testing and some validation

### DIFF
--- a/modules/csg/csg_shape.cpp
+++ b/modules/csg/csg_shape.cpp
@@ -31,6 +31,7 @@
 #include "csg_shape.h"
 
 #include "core/math/geometry_2d.h"
+#include "scene/resources/surface_tool.h"
 
 void CSGShape3D::set_use_collision(bool p_enable) {
 	if (use_collision == p_enable) {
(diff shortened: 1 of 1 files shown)


modules/csg/csg_shape.h
```diff
@@ -89,12 +89,14 @@ class CSGShape3D : public GeometryInstance3D {
...
 		Vector<Vector2> uvs;
 		Vector<real_t> tans;
+		Vector<int> indices;
 		Ref<Material> material;
 		int last_added = 0;
 
 		Vector3 *verticesw = nullptr;
 		Vector3 *normalsw = nullptr;
 		Vector2 *uvsw = nullptr;
+		int *indexsw = nullptr;
```
AThousandShips, line 99:
```suggestion
		int *indicesw = nullptr;
```
To match the vector, the other pointers match their vectors, I'd also move it down one to match the order

modules/csg/csg_shape.cpp
```diff
@@ -412,38 +416,25 @@ void CSGShape3D::_update_shape() {
...
-		array[Mesh::ARRAY_TEX_UV] = surfaces[i].uvs;
-		if (have_tangents) {
-			array[Mesh::ARRAY_TANGENT] = surfaces[i].tans;
+		for (int j = 0; j < surfaces[i].last_added; j++) {
+			surface_tool->set_normal(surfaces[i].normalsw[j]);
+			surface_tool->set_uv(surfaces[i].uvsw[j]);
+			surface_tool->add_vertex(surfaces[i].verticesw[j]);
 		}
+		for (int index_i = 0; index_i < surfaces[i].indices.size(); index_i++) {
+			surface_tool->add_index(surfaces[i].indices[index_i]);
```
AThousandShips, line 433: Above you use the pointer, but the vector here, should it be used here as well?

fire: The structure ShapeUpdateSurface is not surface tool? What do you mean?

YuriSizov:
> Ready for review.

Why is it a draft?

modules/csg/csg_shape.cpp
AThousandShips: I mean that above you use the ptrw, like so:
```cpp
surface_tool->add_vertex(surfaces[i].verticesw[j]);
```
But here the vector, should it be:
```suggestion
			surface_tool->add_index(surfaces[i].indicesw[index_i]);
```

fire: Removed vertex indices of this.

fire:
There is some feedback that this could be slow, but we can't both unify tangents and have multiple callsites..

adamscott (approved):
Great use of `SurfaceTool`.

fire:
Currently testing the obj scene from https://github.com/godotengine/godot/pull/83648#pullrequestreview-1745556457

reduz:
I don´t think doing this is a good idea. SurfaceTool is a lot slower to do this and CSG requires performance. We don´t want to unify tangent generation at all in Godot, all we care is unification in the importers.

adamscott (changes requested):
Requesting changes due to @reduz's comments.

reduz:
I just don't think its a good idea in general. SurfaceTool should not be used for any of these things, period. I know its an easy to use class, but its extremely inefficient, its not meant for internal engine usage, only for users who want something easy to use.

fire:
The profile data is saying that mikktspace tangent generation is taking the profile time, so it's not surface tool, but I understand how this is a design choice.

PR #67730, opened 2022-10-21
Add call_deferred() method to Callable

KoBeWi:
For whatever reason the C++ Callable never got `call_deferred()` method and to defer a call you had to `call_deferredp(nullptr, 0);`.

PR #86301, opened 2023-12-18
Update deferred calls to use Callables

KoBeWi:
Somewhat follow-up to #67730

This PR changes `call_deferred()` calls to use Callables, removing many unnecessary method binds as a result. I also changed some usages of MessageQueue to use Callables too.

There is still some usage of old `call_deferred()` left, namely when a method is vararg or private. I also left some MessageQueue calls in SceneTree, as I'm not sure about performance requirements in there.

EDIT:
btw this might fix  some bugs, as there were some methods called that weren't bound, resulting in silent fail.

### DIFF
--- a/core/variant/callable.cpp
+++ b/core/variant/callable.cpp
@@ -31,13 +31,12 @@
 #include "callable.h"
 
 #include "callable_bind.h"
-#include "core/object/message_queue.h"
 #include "core/object/object.h"
 #include "core/object/ref_counted.h"
 #include "core/object/script_language.h"
 
 void Callable::call_deferredp(const Variant **p_arguments, int p_argcount) const {
-	MessageQueue::get_singleton()->push_callablep(*this, p_arguments, p_argcount);
+	MessageQueue::get_singleton()->push_callablep(*this, p_arguments, p_argcount, true);
 }
 
 void Callable::callp(const Variant **p_arguments, int p_argcount, Variant &r_return_value, CallError &r_call_error) const {
--- a/editor/code_editor.cpp
+++ b/editor/code_editor.cpp
@@ -336,7 +336,7 @@ void FindReplaceBar::_replace_all() {
 	matches_label->add_theme_color_override("font_color", rc > 0 ? get_theme_color(SNAME("font_color"), SNAME("Label")) : get_theme_color(SNAME("error_color"), EditorStringName(Editor)));
 	matches_label->set_text(vformat(TTR("%d replaced."), rc));
 
-	text_editor->call_deferred(SNAME("connect"), "text_changed", callable_mp(this, &FindReplaceBar::_editor_text_changed));
+	callable_mp((Object *)text_editor, &Object::connect).call_deferred("text_changed", callable_mp(this, &FindReplaceBar::_editor_text_changed), 0U);
 	results_count = -1;
 	results_count_to_current = -1;
 	needs_to_count_results = true;
@@ -517,10 +517,10 @@ void FindReplaceBar::_show_search(bool p_focus_replace, bool p_show_only) {
 
 	if (p_focus_replace) {
 		search_text->deselect();
-		replace_text->call_deferred(SNAME("grab_focus"));
+		callable_mp((Control *)replace_text, &Control::grab_focus).call_deferred();
 	} else {
 		replace_text->deselect();
-		search_text->call_deferred(SNAME("grab_focus"));
+		callable_mp((Control *)search_text, &Control::grab_focus).call_deferred();
 	}
 
 	if (text_editor->has_selection(0) && !is_selection_only()) {
@@ -1572,20 +1572,20 @@ void CodeTextEditor::goto_line(int p_line) {
 	text_editor->remove_secondary_carets();
 	text_editor->deselect();
 	text_editor->unfold_line(p_line);
-	text_editor->call_deferred(SNAME("set_caret_line"), p_line);
+	callable_mp((TextEdit *)text_editor, &TextEdit::set_caret_line).call_deferred(p_line, true, true, 0, 0);
 }
 
 void CodeTextEditor::goto_line_selection(int p_line, int p_begin, int p_end) {
 	text_editor->remove_secondary_carets();
 	text_editor->unfold_line(p_line);
-	text_editor->call_deferred(SNAME("set_caret_line"), p_line);
-	text_editor->call_deferred(SNAME("set_caret_column"), p_begin);
+	callable_mp((TextEdit *)text_editor, &TextEdit::set_caret_line).call_deferred(p_line, true, true, 0, 0);
+	callable_mp((TextEdit *)text_editor, &TextEdit::set_caret_column).call_deferred(p_begin, true, 0);
 	text_editor->select(p_line, p_begin, p_line, p_end);
 }
 
 void CodeTextEditor::goto_line_centered(int p_line) {
 	goto_line(p_line);
-	text_editor->call_deferred(SNAME("center_viewport_to_caret"));
+	callable_mp((TextEdit *)text_editor, &TextEdit::center_viewport_to_caret).call_deferred(0);
 }
 
 void CodeTextEditor::set_executing_line(int p_line) {
--- a/scene/gui/spin_box.cpp
+++ b/scene/gui/spin_box.cpp
@@ -280,8 +280,8 @@ void SpinBox::_notification(int p_what) {
 		} break;
 
 		case NOTIFICATION_THEME_CHANGED: {
-			call_deferred(SNAME("update_minimum_size"));
-			get_line_edit()->call_deferred(SNAME("update_minimum_size"));
+			callable_mp((Control *)this, &Control::update_minimum_size).call_deferred();
+			callable_mp((Control *)get_line_edit(), &Control::update_minimum_size).call_deferred();
 		} break;
 
 		case NOTIFICATION_LAYOUT_DIRECTION_CHANGED: {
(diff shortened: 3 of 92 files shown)


YuriSizov:
Ah brilliant, I wanted to do this for so long :) You have an extra empty line, it seems, which fails static checks.

scene/gui/spin_box.cpp
```diff
@@ -280,8 +280,8 @@ void SpinBox::_notification(int p_what) {
 		} break;
 
 		case NOTIFICATION_THEME_CHANGED: {
-			call_deferred(SNAME("update_minimum_size"));
-			get_line_edit()->call_deferred(SNAME("update_minimum_size"));
+			callable_mp((Control *)this, &Control::update_minimum_size).call_deferred();
+			callable_mp((Control *)get_line_edit(), &Control::update_minimum_size).call_deferred();
```
KoBeWi, line 284: This is already called automatically when `NOTIFICATION_THEME_CHANGED` is received, the only difference is that here it's deferred for both nodes. I wonder if it could be removed.

scene/gui/text_edit.cpp
```diff
@@ -6290,9 +6283,6 @@ void TextEdit::_bind_methods() {
 	BIND_ENUM_CONSTANT(LINE_WRAPPING_NONE);
 	BIND_ENUM_CONSTANT(LINE_WRAPPING_BOUNDARY);
 
-	// Internal.
-	ClassDB::bind_method(D_METHOD("_update_wrap_at_column", "force"), &TextEdit::_update_wrap_at_column, DEFVAL(false));
```
AThousandShips, line 6294: This seems used in the converter, unsure if we should respect it as it's internal (even marked as such)

scene/gui/text_edit.cpp
```diff
@@ -428,10 +428,10 @@ void TextEdit::_notification(int p_what) {
 		case NOTIFICATION_ENTER_TREE: {
 			_update_caches();
 			if (caret_pos_dirty) {
-				MessageQueue::get_singleton()->push_call(this, "_emit_caret_changed");
+				callable_mp(this, &TextEdit::_emit_caret_changed).call_deferred();
```
AThousandShips, line 431: Since this used this syntax should it be `MessageQueue::get_singleton()->push_callable(callable_mp(this, &TextEdit::_emit_caret_changed));`?

scene/gui/text_edit.cpp
KoBeWi: This method is technically not exposed, so it shouldn't even be in the converter.

scene/gui/text_edit.cpp
KoBeWi: No, I changed MessageQueue calls to Callables to.

scene/gui/text_edit.cpp
AThousandShips: Strange, see:
```
ERROR: Missing GDScript function in pair (_update_wrap_at - ===> _update_wrap_at_column <===)
```

scene/gui/text_edit.cpp
AThousandShips: Unit tests fail for this class worth checking if using the message queue for these solves that

scene/gui/text_edit.cpp
AThousandShips: https://github.com/godotengine/godot/blob/2d0ee20ff30461b6b10f6fdfba87511a0ebc6642/editor/renames_map_3_to_4.cpp#L217

scene/gui/text_edit.cpp
YuriSizov: If needed, MessageQueue can still be used with callables, `MessageQueue::get_singleton()->push_callable` is a thing :)

AThousandShips: Did not resolve this, unsure what got broken here

scene/gui/text_edit.cpp
KoBeWi: I mean it was a wrong rename. I removed it.

scene/gui/text_edit.cpp
KoBeWi: I need to double-check all removed binds. `callable_mp()` does not support default arguments 😓

core/variant/callable.cpp
```diff
@@ -37,7 +37,7 @@
 #include "core/object/script_language.h"
 
 void Callable::call_deferredp(const Variant **p_arguments, int p_argcount) const {
-	MessageQueue::get_singleton()->push_callablep(*this, p_arguments, p_argcount);
+	MessageQueue::get_singleton()->push_callablep(*this, p_arguments, p_argcount, true);
```
KoBeWi, line 40: This makes the deferred calls print error when they fail, instead of doing so silently.

Using `callable_mp()` can be very error prone, so this helps finding wrong calls. I just wonder if it has unforeseen side effects.

kitbdev:
I think we can remove the `#include "core/object/message_queue.h"` in text_edit.cpp and other files where MessageQueue was removed.

editor/code_editor.cpp
```diff
@@ -1572,20 +1572,20 @@ void CodeTextEditor::goto_line(int p_line) {
 	text_editor->remove_secondary_carets();
 	text_editor->deselect();
 	text_editor->unfold_line(p_line);
-	text_editor->call_deferred(SNAME("set_caret_line"), p_line);
+	callable_mp((TextEdit *)text_editor, &TextEdit::set_caret_line).call_deferred(p_line);
```
kitbdev, line 1575: When using the Script editor's member overview to go to a method, it doesn't work and I get `core\object\message_queue.cpp:223 - Error calling deferred method: 'CodeEdit::TextEdit::set_caret_line': Method expected 5 arguments, but called with 1.`
Do all the arguments need to be explicitly passed now?
In that case this should be ...`call_deferred(p_line, true, true, 0, 0);`.

AThousandShips: It always has to, see above, this is a missed case of default arguments

AThousandShips:
```suggestion
	callable_mp((TextEdit *)text_editor, &TextEdit::set_caret_line).call_deferred(p_line, true, true, 0, 0);
```

editor/code_editor.cpp
```diff
@@ -1572,20 +1572,20 @@ void CodeTextEditor::goto_line(int p_line) {
...
 }
 
 void CodeTextEditor::goto_line_selection(int p_line, int p_begin, int p_end) {
 	text_editor->remove_secondary_carets();
 	text_editor->unfold_line(p_line);
-	text_editor->call_deferred(SNAME("set_caret_line"), p_line);
-	text_editor->call_deferred(SNAME("set_caret_column"), p_begin);
+	callable_mp((TextEdit *)text_editor, &TextEdit::set_caret_line).call_deferred(p_line);
+	callable_mp((TextEdit *)text_editor, &TextEdit::set_caret_column).call_deferred(p_begin);
```
AThousandShips, line 1582:
```suggestion
	callable_mp((TextEdit *)text_editor, &TextEdit::set_caret_column).call_deferred(p_begin);
```
```suggestion
	callable_mp((TextEdit *)text_editor, &TextEdit::set_caret_column).call_deferred(p_begin, true, 0);
```

editor/code_editor.cpp
```diff
@@ -517,10 +517,10 @@ void FindReplaceBar::_show_search(bool p_focus_replace, bool p_show_only) {
 
 	if (p_focus_replace) {
 		search_text->deselect();
-		replace_text->call_deferred(SNAME("grab_focus"));
+		callable_mp((Control *)replace_text, &Control::grab_focus);
```
AThousandShips, line 520:
```suggestion
		callable_mp((Control *)replace_text, &Control::grab_focus).call_deferred();
```
Nullopt otherwise

editor/code_editor.cpp
```diff
@@ -517,10 +517,10 @@ void FindReplaceBar::_show_search(bool p_focus_replace, bool p_show_only) {
...
 	} else {
 		replace_text->deselect();
-		search_text->call_deferred(SNAME("grab_focus"));
+		callable_mp((Control *)search_text, &Control::grab_focus);
```
AThousandShips, line 523:
```suggestion
		callable_mp((Control *)search_text, &Control::grab_focus).call_deferred();
```

KoBeWi:
I triple-checked everything, but still missed some cases. Thanks for looking into it.

akien-mga (approved):
Didn't review in depth, but makes sense to me overall. Could be good to merge after 4.3-dev2 is released, giving us some time to spot potential regressions before dev3.

> btw this might fix some bugs, as there were some methods called that weren't bound, resulting in silent fail.

This could actually introduce bugs, if the silent failures are now code that's not supposed to be working, and it working would introduce unwanted changes. So these might be worth flagging for relevant maintainers to look at more closely.

KoBeWi:
> This could actually introduce bugs, if the silent failures are now code that's not supposed to be working, and it working would introduce unwanted changes. So these might be worth flagging for relevant maintainers to look at more closely.

Well, one was in RichTextLabel, I don't remember the other one.

akien-mga:
> > This could actually introduce bugs, if the silent failures are now code that's not supposed to be working, and it working would introduce unwanted changes. So these might be worth flagging for relevant maintainers to look at more closely.
> 
> Well, one was in RichTextLabel, I don't remember the other one.

I see one with `_emit_navigation_debug_changed_signal` in NavigationServer3D, CC @smix8.

akien-mga:
Thanks!

PR #78472, opened 2023-06-20
Remove some `DisplayServer::mouse_get_position` calls

Sauermann:
These functions can get called in situations, where there is no reasonable DisplayServer available (e.g. XR or Gui-in-3D Demo). This PR introduces `Viewport::get_embedder_mouse_position()` which verifies if `DisplayServer::mouse_get_position` can get called.

part of #77189
resolves no longer #86240 (this got fixed in #86304)

- [X] ColorPicker (no change required)
- [X] EditorTitleBar (use `get_embedder_mouse_position`)
- [X] MenuBar (this got fixed in #86304)
- [X] MenuButton (this got fixed in #86304)
- [X] PopupMenu::_notification (use `get_embedder_mouse_position`)
- [X] PopupMenu::_parent_focused (use `get_embedder_mouse_position`)

PR #63950, opened 2022-08-05
Implement MenuBar control to wrap Popup menus or native menu, use native menu for editor.

bruvzg:
- Implements `MenuBar` control, that can use either collection of `PopupMenu` or OS native global menu.

### TODO
- [X] Implement `MenuBar` control.
- [X] Convert main editor menu to `MenuBar`.
- [X] ~Convert all sub-editor menus to `MenuBar`~ Most sub-editor menus are composed of singular `MenuButton`s created by different plugins, and since multiple can be opened at once, it's not really convenient to cram everything into a global menu. So, keeping it for main menu only.
- [x] Test on all platforms.
- [x] Add documentation.

Supersede #38044
Implements https://github.com/godotengine/godot-proposals/issues/736

Default:

Screenshot 2022-08-08 at 13 54 48

With "Interface → Editor → Use Embedded Menu" setting enabled:

Screenshot 2022-08-08 at 13 59 45

PR #50619, opened 2021-07-19
Make `MenuButton`'s `switch_on_hover` work again

YeldhamDev:
Since `Popup`s were made into full-fledged windows, the `switch_on_hover` feature (which allowed to switch from the popup of one `MenuButton` to another) broke completely. This PR fixes that.

Things that needed to be done to make this PR work:
- Make the code reside in `MenuButton` instead of `Viewport`, which since this is the only node that cares about it, is a good thing.
- Make `Viewport::_gui_find_control()` a public function, so it could be used outside of it.
- Make the code work via internal process to detect mouse input, as the mouse position can't be detected if it's outside the window in OSes like Windows. So when the popup is visible, `set_process_internal()` is enabled, and the position verified with `DisplayServer::mouse_get_position()`. Not the prettiest solution, but it's the one that works.
- Removed `MenuButton` as a friend class of `PopupMenu`. Not really related to the PR itself, but doesn't seem to be doing anything useful, and likely a leftover of something in the past.

PR #86304, opened 2023-12-18
Fix `MenuBar` and `MenuButton` hover position scaling properly with the scale factor multiplier

### LINKED ISSUE #86240: `MenuButton` hover position does not scale with the scale factor multiplier
### Tested versions

Reproducible in 4.2.1

### System information

Godot v4.2.1.stable - openSUSE Tumbleweed 20231210 - X11 - GLES3 (Compatibility) - NVIDIA GeForce GTX 1060 6GB (nvidia) - AMD Ryzen 5 1600 Six-Core Processor (12 Threads)

### Issue description

MenuBar and MenuButton's `switch_on_hover` property is not working properly when the scale factor multiplier is other than 1. The position the cursor needs to be in order for a PopupMenu to appear seems to be the one where the scale factor multiplier 1. In other words, the hover position is not scaling with the scale factor multiplier and remains constant. This results in a weird behavior, as seen in the below video.

https://github.com/godotengine/godot/assets/35376950/aa7cb8d6-fbe5-43b9-9b40-330ff65fa2c6

(Scale multiplier is set to 1.5 in the video)

The bigger the distance from 1, the weirder the behavior gets. Note that this also occurs when the scale is less than 1.

Interestingly, this doesn't seem to happen to the Godot editor itself.

### Steps to reproduce

Create a project with a MenuBar node with PopupMenu items, or multiple MenuButtons as siblings and enable "Switch on Hover". Go to the project settings, General tab, Display, Window and set the Scale to something other than 1. Run the project, click on a menu button and move your cursor around to switch on a sibling menu button on hover. You can see that the position is wrong.

[MenuButtonHoverBug.zip](https://github.com/godotengine/godot/files/13693910/MenuButtonHoverBug.zip)

OverloadedOrama:
Fixes #86240. The solution was basically to just remove `mouse_pos_adjusted` and the code related to it, and simply get the position from `get_local_mouse_position()`, and it now works properly. I don't think this breaks anything, but feel free to correct me if I'm wrong. From what I see, the code is there since the beginning of `MenuBar` (#63950) and it seems to be copy-pasted from the `MenuButton` code, introduced in #50619.

~~`MenuButton` seems to be a bit more difficult to fix. Simply using `get_local_mouse_position()` is not working properly, but getting the local mouse position from the parent is working properly. Problem is, this function is only implemented in `CanvasItem` nodes. If the `MenuButtons` have a 3D node as a parent, this solution is not going to work. Which is why, I am assuming, the `mouse_pos_adjusted` code is in `MenuButton` in the first place. Any ideas on how to handle this would be appreciated!~~
EDIT: `MenuButton` has also been fixed, thanks to #78472.

If this PR gets accepted, I would greatly appreciate it if also got cherry-picked to 4.2.x.

### DIFF
--- a/scene/gui/menu_bar.cpp
+++ b/scene/gui/menu_bar.cpp
@@ -181,24 +181,7 @@ void MenuBar::_popup_visibility_changed(bool p_visible) {
 	}
 
 	if (switch_on_hover) {
-		Window *wnd = Object::cast_to<Window>(get_viewport());
-		if (wnd) {
-			mouse_pos_adjusted = wnd->get_position();
-
-			if (wnd->is_embedded()) {
-				Window *wnd_parent = Object::cast_to<Window>(wnd->get_parent()->get_viewport());
-				while (wnd_parent) {
-					if (!wnd_parent->is_embedded()) {
-						mouse_pos_adjusted += wnd_parent->get_position();
-						break;
-					}
-
-					wnd_parent = Object::cast_to<Window>(wnd_parent->get_parent()->get_viewport());
-				}
-			}
-
-			set_process_internal(true);
-		}
+		set_process_internal(true);
 	}
 }
 
@@ -334,8 +317,7 @@ void MenuBar::_notification(int p_what) {
 				// Handled by OS.
 				return;
 			}
-
-			Vector2 pos = DisplayServer::get_singleton()->mouse_get_position() - mouse_pos_adjusted - get_global_position();
+			Vector2 pos = get_local_mouse_position();
 			if (pos == old_mouse_pos) {
 				return;
 			}
--- a/scene/gui/menu_bar.h
+++ b/scene/gui/menu_bar.h
@@ -71,7 +71,6 @@ class MenuBar : public Control {
 	int selected_menu = -1;
 	int active_menu = -1;
 
-	Vector2i mouse_pos_adjusted;
 	Vector2i old_mouse_pos;
 	ObjectID shortcut_context;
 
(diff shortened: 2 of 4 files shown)


Sauermann:
During my investigation into how to fix this, i came to a nearly identical conclusion. Please have a look at #78472.

OverloadedOrama:
Oh nice, I didn't know of this PR. It seems that our code is identical when it comes to MenuBar, but yours addresses even more issues. I guess this PR should be closed then, and hopefully yours will get merged soon and ideally cherry-picked for 4.2.2. Thanks for bringing it into my attention!

Sauermann:
Lets's leave your PR open, since it is easier to review.

OverloadedOrama:
> > It seems that our code is identical when it comes to MenuBar
> 
> Our approaches differ: in your code `set_process_internal(true);` gets called only when the MenuBar is within a `Window` while in my code it gets called additinally when it is within a `SubViewport`.

Right, I didn't notice that. I only changed the part related to the scale factor multiplier issue, I left the rest of the code as-is. But it sounds to me that your approach is even better, as it makes MenuBar hovering work while inside `SubViewport`s too. Both approaches fix #86240 though.

OverloadedOrama:
I updated the PR to also fix `MenuButton`'s scaling issue, and I updated the `MenuBar` code changes to be the same as #78472, so our PRs should be identical now for MenuButton and MenuBar, in case the maintainers decide they want to merge this one first.

Sauermann (approved):
Regarding MenuBar: I came to the same conclusion of how to fix the problem.
Regarding MenuButton: Additionally someone else needs to verify the changes, since they are identical to my suggested approach in #78472.

akien-mga (approved):
Looks good to me.

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #86305, opened 2023-12-18
Workaround on GodotSharp duplicate key exception when adding generic entries to ScriptTypeBiMap

### LINKED ISSUE #79519: C# Script Registration that use Generics can error out in the ScriptManagerBridge
### Godot version

4.1.stable.mono

### System information

Godot v4.1.stable.mono - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3080 (NVIDIA; 31.0.15.3640) - AMD Ryzen 7 5800X 8-Core Processor (16 Threads)

### Issue description

This issue occurs when two scripts are derived from a generic script that uses the same type.

Say you have a script, `GenericScript<T>`:

```cs
using Godot;
using System.Collections.Generic;

public partial class GenericScript<T> : Node3D
{
  protected List<T> someList = new List<T>();
}
```

And we have two scripts that derive from that class with the same type parameter:

```cs
public partial class TestScript : GenericScript<float>
{
	public override void _Ready()
	{
		someList.Add(1);
	}
}
```

```cs
public partial class OtherTestScript : GenericScript<float>
{

}
```

Building the mono project while a scene is loaded will present an error. See the reproduction steps below.

### Steps to reproduce

# Step 1
Create a scene that has two nodes. One node has `TestScript` attached, the other node has `OtherTestScript` attached (see above for definition).

# Step 2
Rebuild Mono from the MSBuild tab in the bottom right.

You will see the following error in the console:

```
modules/mono/glue/runtime_interop.cpp:1324 - System.ArgumentException: An item with the same key has already been added. Key: GenericScript`1[System.Single]
     at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
     at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
     at Godot.Bridge.ScriptManagerBridge.ScriptTypeBiMap.Add(IntPtr scriptPtr, Type scriptType) in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Core/Bridge/ScriptManagerBridge.types.cs:line 23
     at Godot.Bridge.ScriptManagerBridge.TryReloadRegisteredScriptWithClass(IntPtr scriptPtr) in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Core/Bridge/ScriptManagerBridge.cs:line 579
```

# Step 3A
If you Rebuild Mono again, you will see the following error and Godot will no longer be in a good state.

```
modules/mono/mono_gd/gd_mono.cpp:513 - .NET: Failed to unload assemblies. Please check https://github.com/godotengine/godot/issues/78513 for more information. (User)
```

# Step 3B
If you instead close the scene so that no scenes are loaded and Rebuild Mono, the rebuild will appear successful and no errors will be reported.

A repro project can be found in [this repository](https://github.com/taylorhadden/godot-csharp-generic-assembly-reload-error). At the time of writing this ticket, the latest version was `5f6983c`.

The repro contains the scripts defined above, a test scene, and a `README.md` file with steps essentially the same as the reproduction steps above.

Delsin-Yu:
A workaround for #79519, this issue is a blocking issue for most .Net users.
Which is one of the causes of the Mega #78513
<s>While #83204 resolved this issue by simply rejecting the duplicating addition, it still left the issue on the Native end and caused error message prints. </s>
<s>We believe this PR(86305) solves the core issue of this whole chain.</s>
Found by @Smallantsa

Validated using [This Repo](https://github.com/taylorhadden/godot-csharp-generic-assembly-reload-error)

**_It turns out that this PR is yet another workaround on the issue, while it somehow resolves the symptoms, further investigation on `Generic Derivatives holding empty Script Path` is required to fully resolve this issue._**

By returning null on `ReflectionUtils.FindTypeInLoadedAssemblies`, this PR is a workaround for issues related to the `System.ArgumentException: An item with the same key has already been added.` in `ScriptTypeBiMap`.
**While side effects are still to be validated.**

<s>It looks like the old way of obtaining a generic type will return a different instance for the type, resulting in a duplicate key exception in `ScriptTypeBiMap.Add(IntPtr scriptPtr, Type scriptType)`
```csharp
AppDomain.CurrentDomain.GetAssemblies()
            .FirstOrDefault(a => a.GetName().Name == assemblyName)?
            .GetType(typeFullName)
```

This new approach appears to return the same instance, so it should fix all issues related to:
```
modules/mono/glue/runtime_interop.cpp:1324 - System.ArgumentException: An item with the same key has already been added. Key: GenericScript`1[System.Single]
     at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
     at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
     at Godot.Bridge.ScriptManagerBridge.ScriptTypeBiMap.Add(IntPtr scriptPtr, Type scriptType) in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Core/Bridge/ScriptManagerBridge.types.cs:line 23
     whatever  other calls
```
</s>

### DIFF
--- a/modules/mono/glue/GodotSharp/GodotSharp/Core/ReflectionUtils.cs
+++ b/modules/mono/glue/GodotSharp/GodotSharp/Core/ReflectionUtils.cs
@@ -9,6 +9,7 @@ internal class ReflectionUtils
 {
     public static Type? FindTypeInLoadedAssemblies(string assemblyName, string typeFullName)
     {
-        return Type.GetType($"{assemblyName}.{typeFullName}");
+        // TODO: Validate the side effects of this workaround
+        return null;
     }
 }
--- a/modules/mono/glue/GodotSharp/GodotSharp/Core/ReflectionUtils.cs
+++ b/modules/mono/glue/GodotSharp/GodotSharp/Core/ReflectionUtils.cs
@@ -9,8 +9,6 @@ internal class ReflectionUtils
 {
     public static Type? FindTypeInLoadedAssemblies(string assemblyName, string typeFullName)
     {
-        return AppDomain.CurrentDomain.GetAssemblies()
-            .FirstOrDefault(a => a.GetName().Name == assemblyName)?
-            .GetType(typeFullName);
+        return Type.GetType($"{assemblyName}.{typeFullName}");
     }
 }


AThousandShips:
> While #83204 resolved this issue by simply rejecting the duplicating addition, it still left the issue on the Native end and caused error message prints.

To confirm, you tested this PR and confirmed it with the changed code? This PR has not been merged yet so it hasn't solved anything yet

Delsin-Yu:
> > While #83204 resolved this issue by simply rejecting the duplicating addition, it still left the issue on the Native end and caused error message prints.
> 
> To confirm, you tested this PR and confirmed it with the changed code? This PR has not been merged yet so it hasn't solved anything yet

I tested 83204  with the changed code and confirmed it does **NOT** solve the core issue.

CantyCanadian:
I don't know if I'm missing something, but building latest engine with your changes in, I get the following errors which are very similar. I was using 83204 with 4.1 but dropped the change with 4.2. Now that the error is back, I tried your fix hoping that it was a permanent solution.
image

Delsin-Yu:
> I don't know if I'm missing something, but building latest engine with your changes in, I get the following errors which are very similar. I was using 83204 with 4.1 but dropped the change with 4.2. Now that the error is back, I tried your fix hoping that it was a permanent solution.

Is it possible for you to share a minimal reproduction for your case?

CantyCanadian:
I really would, however, it's not 100% reproductible and I am not sure what is the smoking gun in my case. The only points I found that could've caused it to happen were:
- A class with a generic inheriting from a Resource type, then that generic class being inherited by various other resources, setting the generic to a value.
- One or many resources being created in the project using that non-generic top-level class.
- Those resources being passed by Export to a node in the scene.

In that case, if I build the code with it being dirty (not just building with no changes), there's a chance it breaks.

Delsin-Yu:
> I really would, however, it's not 100% reproductible and I am not sure what is the smoking gun in my case. If I had that, I could make a barebones project, but right now, all I could do is send my whole project. The only points I found that could've caused it to happen were:
> 
> * A class with a generic inheriting from a Resource type, then that generic class being inherited by various other resources, setting the generic to a value.
> * One or many resources being created in the project using that non-generic top-level class.
> * Those resources being passed by Export to a node in the scene.
> 
> In that case, if I build the code with it being dirty (not just building with no changes), there's a chance it breaks. It will break after a while of trying, but I can't figure what exactly causes it.

The issue this PR is focusing on is related to the opening scene from my tests, that is, you need to have a specific scene opened while rebuilding your C# solution; hope this info will help you identify the issue in your project. It's okay for me if you send the whole project.

dsh0416:
From our usages, we can also found this issue 100% reproducible when generic entries being re-built in the editor. We believe that this PR has solved our problem in a patched version of GodotSharp.

paulloz:
I'm wondering if it is a good thing to essentially silence the error while not solving the main issue (there are duplicate entries in the bimap). Isn't it useful for us to have an error telling us there is an issue to fix?

Delsin-Yu:
> I'm wondering if it is a good thing to essentially silence the error while not solving the main issue (there are duplicate entries in the bimap / we are inserting generics in the bimap). Isn't it useful for us to have an error telling us there is an issue to fix?

The issue we are discussing here is not user-related, it is an editor bug that requires fixing.
There is no duplicate entry created on the user side, and using generic is a standard practice in C#.

The issue here is caused by GodotSharp creating duplicate generic entries out of nowhere (under certain circumstances, one user script that derives from a generic class can cause GodotSharp to generate duplicate System.Type instances) and causing trouble.

paulloz:
Yes, I know.

What I’m saying is: as discussed in the other linked PR, generics should not hit the bimap at all. IINW, what we do here is working around so that, when they do hit it, no duplicate key is raised. Essentially removing the errors, but leaving the generics in the map. Or am I misunderstanding the fix?

>>

Smallantsa:
@raulsntos Can you help me take a look？

Delsin-Yu:
> Yes, I know. What I’m saying is: as discussed in the other linked PR, generics should not hit the bimap at all. IINW, what we do here is working around so that, when they do hit it, no duplicate key is raised. Essentially removing the errors, but leaving the generics in the map. Or am I misunderstanding the fix?

Previously, there was a bug that caused generics to hit the Bi-Map.
83204 is the "working around" you mentioned.
This PR discovered that bug (causing generics to hit the Bi-Map) and go it fixed, so no other "workaround" is needed.

paulloz:
My apologies, I misunderstood what the intent behind the patch was. You're right, this would prevent the insertion into the bimap in `ScriptManagerBridge.TryReloadRegisteredScriptWithClass` by returning early on the null check above. 

Though, I'm now a bit concerned (pardon me if this is annoying) by the fact `$"{assemblyName}.{typeFullName}"` is not what an [assembly qualified name](https://learn.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname?view=net-6.0#system-type-assemblyqualifiedname) looks like. Isn't this always returning `null`, since the types we're reloading are always in a different assembly from the executing one?

Delsin-Yu:
> My apologies, I misunderstood what the intent behind the patch was. You're right, this would prevent the insertion into the bimap in `ScriptManagerBridge.TryReloadRegisteredScriptWithClass` by returning early on the null check above.
> 
> Though, I'm now a bit concerned (pardon me if this is annoying) by the fact `$"{assemblyName}.{typeFullName}"` is not what an [assembly qualified name](https://learn.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname?view=net-6.0#system-type-assemblyqualifiedname) looks like. Isn't this always returning `null`, since the types we're reloading are always in a different assembly from the executing one?

That's a good point, I think I do need to talk to @Smallantsa for more detail.

My testing (in SharpLab) indicates that this API does not work when passing `type.FullName`, this made me think the `typeFullName` godot provides is in another format, if this actually returns null then there should be more to fix.

Delsin-Yu:
From the testing project I have here, when unloading assemblies, the `FindTypeInLoadedAssemblies` API is only called when dealing with generic types.

paulloz:
It is basically when our script path is empty when iterating through the scripts to reload ([here](https://github.com/godotengine/godot/blob/master/modules/mono/csharp_script.cpp#L988)). I'm unsure of other possible cases where the path would be empty here (essentially: part of the class hierarchy for something, but no `[ScriptPathAttribute]`).

zaevi:
> It is basically when our script path is empty when iterating through the scripts to reload ([here](https://github.com/godotengine/godot/blob/master/modules/mono/csharp_script.cpp#L988)). I'm unsure of other possible cases where the path would be empty here (essentially: part of the class hierarchy for something, but no `[ScriptPathAttribute]`).

After discussion with @Delsin-Yu and @Smallantsa:

It's ok to ignore ScriptPath for **generic types**. ` (like Component<T>) `

However, **constructed generic types** ` (like Component<int>, Component<string>) ` need the path, and I don't think we can give `[ScriptPathAttribute]` to them. Therefore, the engine will re-create `CSharpScript` for constructed generic types and re-register to `_scriptTypeBiMap` and throw exceptions.

It's necessary to prevent re-instance `CSharpScript` for constructed-generic types.

cc @raulsntos

Delsin-Yu:
Superseded by #87550

PR #86312, opened 2023-12-18
[iOS] Remove deprecated LaunchImages support and switch to Storyboard by default.

### LINKED ISSUE #82730: All rendering has jagged edges on iOS while recording or taking screenshots
### Godot version

### System information

iPhone 13 Pro Max – iOS 17.0.2

### Issue description

All rendering on screen, 2D or 3D, has jagged edges on iPhone screenshots or recordings, but not when no recording is taking place.

The issue is present on:

1. **Screenshots of a Godot game** (using the screenshot key combinations on the iPhone)
2. **Recordings of a Godot game** (recorded using the iPhone native record-screen action found at the control center). _Both the output video and the screen, while recording, show jagged edges on a Godot game. Once recording stops, jagged edges disappear, but of course remain on the recorded file._

Just to be clear, the image _seems_ to be fine when no recording is taking place, or when not looking at a screenshot. I don't know if the rendering scale of Godot changes when a recording or screenshot take place, **or whether there is an upscaling smoothing that goes away when these recordings or screenshots take place.**

---

### Could it be stretching? I don't think so.

At first I thought this was related to stretching or HiDPI settings, so I played a lot with that, to no avail. Here is an example of how it looks on the MRP:

Screenshot 2023-10-03 at 3 45 58 PM

But of course, **the issue only exists while recording or on screenshots,** so it wouldn't make sense that is related to this. Why aren't edges always jagged?

The jagginess we see in screenshots/recordings is exactly you'd see if you are indeed using **Stretch** for the window and have the aspect expand.

---

### Visual comparisons

It's easier to see in person. I've made a comparison between a screenshot and a photo taken with another phone. Of course this is not ideal, as the camera introduces it's own distortions and blurriness, but I think it helps to showcase the difference quite well.

1. Look at the arrows.
2. Look at the eyes of our little Godot buddy.

**ZOOM IN A LOT**

Jagged Edges

---

### Conclusion

The main issue with this is that any kind of marketing asset you want to make out of the game might require recording outside the target device, which is not ideal. Or anyone wanting to share a screenshot of the game will make it look ugly, which again is pretty bad presentation-wise.

I have not noticed any other app or game I have do the same on my iPhone.

If you need any more information, please let me know and I'll do my best to provide it.

### Steps to reproduce

1. New project with any 3D objects, some UI elements as well (or use included MRP).
2. Export to iOS, install on iPhone.
3. Start a recording and see how all edges become jagged, real time. Video will include jagged edges as well. Alternatively, take a screenshot and zoom in to see jaggies better. Otherwise edges will look fine.

Any fresh project will have this issue, but I am sharing one that I also tested on iPhone.

[Godot Bugs Playground.zip](https://github.com/godotengine/godot/files/12793073/Godot.Bugs.Playground.zip)

bruvzg:
Seems like use of deprecated LaunchImages is causing app to be unable to use native resolution and iPhone 6+ type UI scaling is applied.

Fixes https://github.com/godotengine/godot/issues/82730


bruvzg:
Note: storyboards are supported in iOS 8 (and later), so there's no need to keep LaunchImages for compatibility.

Calinou (approved):
This makes sense on my end (although I don't have a device to test this). Code looks good to me at a glance.

akien-mga (approved):

akien-mga:
Thanks!

owlnewworlds:
Hello everyone,

Does this issue is fixed in Godot 4.3 beta 3 ? I'm trying my first iOS exports and my game isn't fullscreen/scaled properly.

PR #86313, opened 2023-12-18
Implement `ETC2_R` and `ETC2_RG` compression to etcpak

BlueCube3310:
Follow-up to #85842

Adds an ETC2_R and ETC2_RG implementation to `etcpak`, which greatly improves normal map quality on mobile devices. Also adds SIMD-optimized variants of the RGTC compressor from the aforementioned PR.

If everything works correctly, I will attempt to upstream the changes after cleaning up the code.

I've done some basic tests and everything seems to function properly, though a lot of testing is still required.

Results:
| ETC2-RA-As-RG | ETC2-RG |
|----------|----------|
|aa|bb

A basic MRP for testing the changes: [etc_compression.zip](https://github.com/godotengine/godot/files/13708207/etc_compression.zip)


AThousandShips:
I'd say to generate a patch file for the new additions to facilitate updates etc., as is normally done, would also help upstreaming I think 🙂

fire:
> Is there a how-to generate the patch on third party code mikktspace?
>
> Commit unmodified upstream code.
> Make changes.
> git diff > patches/name.patch
> Commit changes + patch (possibly as amend of the first commit if the unmodified upstream code doesn't make sense as an isolated commit).

BlueCube3310:
So it turns out that the EAC compression is <a href=https://github.com/darksylinc/betsy/blob/cc723dcae9a6783ae572f64d12a90d60ef8d631a/bin/Data/eac.glsl#L5C14-L5C14>slightly different</a> when used on the r/rg ETC2 formats. I'll mark this PR as draft until I ensure that the algorithm works accurately.

fire (approved):
Going by the photo it looks ok, but I didn't review the code technically.

Needs testing on a mobile device.

clayjohn (approved):
Looks good. Tested on a Pixel 7 and the results look great. Good job!

akien-mga:
Would you consider submitting this new feature upstream? https://github.com/wolfpld/etcpak/

It would be best if we don't have to maintain our own downstream patches in the long run, and the etcpak maintainer might have an extra qualified review to add to ensure the functionality is good also for our use case.

BlueCube3310:
> Would you consider submitting this new feature upstream? https://github.com/wolfpld/etcpak/
> 
> It would be best if we don't have to maintain our own downstream patches in the long run, and the etcpak maintainer might have an extra qualified review to add to ensure the functionality is good also for our use case.

PR for the RGTC part is ready: https://github.com/wolfpld/etcpak/pull/36
I'll work on the ETC part soon.

fire:
The changes to RGTC for etcpak is merged.

PR #86267, opened 2023-12-17
Upgrade SDFGI to HDDAGI

reduz:
Supersedes #86007

This is a new global illumination system that is a full upgrade to SDFGI Key advantages are:

* Significantly faster base frame time.
* An order of magnitude faster updates (SDFGI drops frames very significantly when moving, as a consequence of having to redo its entire SDF even for a small motion).
* Higher quality (less light leaks, properly energy conservation, much better reflections, others)
* Occlusion remains the same as in SDFGI. I experimented with other systems (the one in DDGI) but everything has worse tradeoffs.

It is meant as a drop-in replacement, so games using SDFGI should get HDDAGI transparently.

TODO:

- [x] For some reason gets DEVICE LOST on Intel GPUs. Can´t figure out why.
- [x] Waiting for a PR fixing the normal buffer precision in order to fix some issues. https://github.com/godotengine/godot/pull/86316
- [x] Volumetric fog GI injection appears broken.

TODO AFTER MERGING:

- [ ] High density probes support.
- [ ] Dynamic objects support.

### Motion performance:

By far the **biggest improvement is the performance when moving the camera**. This is what makes SDFGI unusable on lower end hardware and can even push down the FPS very strongly on higher end hardware.  This is because SDFGI needs to **regenerate the entire SDF cascade** using a jump flood even if it moved just a bit. HDDAGI uses HDDA so it only does local updates to the cascade and no scrolling.

Here is what happens in SDFGI when the camera moves fast:

https://github.com/godotengine/godot/assets/6265307/2c43db90-f519-4ca2-b23c-bec9897ac722

As you can see FPS take a large dip. This really sucks for any kind of games where the camera moves fast, like a racing game, a fast shooter, etc.

In comparison, HDDAGI is unaffected by camera motion:

https://github.com/godotengine/godot/assets/6265307/7d21c3f4-80f3-431a-99d4-b055d6bb99ac

### Static performance:

When rendering static (camera not moving) HDDAGI is also considerably faster than SDFGI:

This is SDFGI, the first set of dispatches is the light update, the second is the probe raytracing. On a Geforce 1650 it takes 2.43ms:

image

In HDDAGI, the same task takes 0.93ms. While not a great difference on a 1650, this makes a much larger difference on IGP:
image

## Quality of Indirect Light

SDFGI does not properly do light conservation, while HDDAGI does. Here is comparison screens by @Jamsers:

SDFGI: Notice light is more uneven in general, and there are some color issues due to the use of spherical harmonics.

image
image

[...]

PR #86316, opened 2023-12-18
Use best fit normals for storing screen space normals

### LINKED ISSUE #74633: Vulkan: SDFGI reflections appear to jitter when the camera rotates slowly
### Godot version

4.0 Stable

### System information

Windows 11, Nvidia RTX4070ti, AMD Ryzen7700x

### Issue description

SDFGI reflections will visibly jitter with camera rotation, especially on materials with a metallic property of 1 and roughness property below 0.5. Will still occur on non-metallic materials.
This is visible at runtime as well as in the editor. It is not visible when panning/moving the camera laterally.

### Steps to reproduce

Apply a reflective material to any object and observe the effect while the camera is being moved to "look around" with the mouse.
Alternatively, load MPR and observe reflective cubes while moving camera around.

[SDFGI Reflection Jitter.zip](https://github.com/godotengine/godot/files/10927493/SDFGI.Reflection.Jitter.zip)

### LINKED ISSUE #84184: Vulkan: VoxelGI reflections appear to jitter when the camera rotates
### Godot version

4.1.2-stable

### System information

Windows 11 - Godot v4.1.2.stable - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 3090 (nvidia, 545.92) - AMD Ryzen 9 5900X 12-Core Processor 3.70 GHz

### Issue description

https://github.com/godotengine/godot/assets/623463/e88aea3e-3db2-4422-8e12-1a4e893f0ff0

Using voxelGI or reflectionprobe results in the reflected image stuttering when the camera rotates.

### Steps to reproduce

- Run a scene with voxelgi or reflection probe enabled
- Rotate the camera

[godot-reflection-stutter-bug.zip](https://github.com/godotengine/godot/files/13204944/godot-reflection-stutter-bug.zip)

clayjohn:
Fixes: https://github.com/godotengine/godot/issues/74633
Fixes: https://github.com/godotengine/godot/issues/84184

Needed for: https://github.com/godotengine/godot/pull/86267

This uses an approach described in https://advances.realtimerendering.com/s2010/Kaplanyan-CryEngine3(SIGGRAPH%202010%20Advanced%20RealTime%20Rendering%20Course).pdf to store normals in a way that minimizes error. Essentially, we pre-generate an LUT that describes the length of a vector that (when compressed, uncompressed, then normalized) most closely resembles the original vector. 

Thanks to Anton Kaplanyan for pioneering the technique, sharing it, and ultimately helping me fine-tune it for use in Godot. 

Before | After
-|-
Screenshot from 2023-12-21 14-36-29 | Screenshot from 2023-12-21 14-36-45

_Error visualization, darker is better_
Before | After
-|-
Screenshot from 2023-12-21 14-37-34 | Screenshot from 2023-12-21 14-37-48

_Bonus section. Using full float normals:_
Full Float | This PR
-|-
Screenshot from 2023-12-21 14-36-17 | Screenshot from 2023-12-21 14-36-45

_Bonus section. Using octahedral instead https://github.com/clayjohn/godot/commit/62d234fd6d9b86106793687cc0c721b8ac1c5c40. (pros: better quality, better encode speed; cons: worse decode speed, harder to debug)_
Reflection | Error visualization
-|-
Screenshot from 2023-12-21 14-37-11 | Screenshot from 2023-12-21 14-37-56


jcostello:
Does this affect the look of the normal map in materials outside VoxelGI and HDDAGI?

clayjohn:
> Does this affect the look of the normal map in materials outside VoxelGI and HDDAGI?

No, right now this only improves things that use the normal roughness buffer (so SDFGI, VoxelGI, SSR). It might have a slight benefit for SSAO and SSIL, but likely isn't noticable. 

I think a similar change might help LightmapGI though, so I'll experiment with that as well.

jcostello:
Let me know about the LightmapGI

Calinou (approved):
Tested locally, it works as expected. This particularly helps when rendering smooth materials at high resolutions.

Performance is pretty much identical compared to before a on RTX 4090 at 4K.

## Preview

*Click to view at full size.*

### SDFGI

Before | After
-|-
Screenshot_20231219_142933 webp | Screenshot_20231219_142955 webp


### VoxelGI

Before | After
-|-
Screenshot_20231219_143041 webp | Screenshot_20231219_143051 webp

### Normal buffer

*Looks glitchy, but the final result looks good either way. clayjohn suggested this was due to this debug view not being normalized before display.*

Before | After
-|-
Screenshot_20231219_151152 webp | Screenshot_20231219_151137 webp

## Video with slow camera rotation

*Using VoxelGI. This was rendered at 4K and encoded in 720p to rule out aliasing from the equation.*

*Don't mind the flickering at the top – it's caused by me recording with V-Sync disabled.*

### Before

https://github.com/godotengine/godot/assets/180032/946cb4d7-19ce-4ef4-9b3a-7a795a8c409e

### After


https://github.com/godotengine/godot/assets/180032/c12c1c77-ddc0-4f92-8532-0cab109f8d48

servers/rendering/renderer_rd/shaders/forward_clustered/best_fit_normal.glsl
```diff
@@ -0,0 +1,43 @@
...
+}
+
+void main() {
+	vec2 uv = vec2(gl_GlobalInvocationID.xy) / vec2(1024.0);
+	uv.y *= uv.x;
+
+	vec3 dir = vec3(uv.x, uv.y, 1.0);
+	float max_c = max(dir.x, max(dir.y, dir.z));
+	imageStore(current_image, ivec2(gl_GlobalInvocationID.xy), vec4(find_minimum_error(dir) * max_c, 1.0, 1.0, 1.0));
+}
```
Calinou, line 43: Add missing newline at end of file:

```suggestion
}

```

servers/rendering/renderer_rd/shaders/forward_clustered/best_fit_normal.glsl
```diff
@@ -0,0 +1,43 @@
+#[compute]
```
Calinou, line 1: It might be worth adding a top-level comment to this shader to explain that this shader is only run once to generate an image on the GPU, rather than something run continuously.

clayjohn: Done!

servers/rendering/renderer_rd/shaders/forward_clustered/best_fit_normal.glsl
clayjohn: Thanks!

clayjohn:
Okay, just pushed an update doing the following:
1. Ensure that the normal is being used consistently (SSAO, SSIL, SSR, VoxelGI, SDFGI)
2. Switch to a signed buffer so we can easily encode static/dynamic in the roughness channel
3. Added a compatibility function so this change doesn't break compatibility in user shaders

This should be ready for review and merging. 

@jcostello In the end it looks like LightmapGI does its own thing with normals. So I didn't touch it.

reduz:
@clayjohn it looks good, but (sorry you are going to hate me for this) from what I am reading, in modern OpenGL using 8-bit SNORM, -128 and -127 are exactly the same (both map to -1.0, see the _Signed_ section [here](https://www.khronos.org/opengl/wiki/Normalized_Integer)) which may be a problem storing roughness. 

Sounds like we should probably return to storing UNORM, so we can properly store the roughness range better for dynamic objects. My suggestion given that min roughness (0) is important for sharp reflections, while max roughness changes nothing between say 0.99 and 1.0), is to store as follows:

```glsl
roughness = roughness * (127.0/255.0);
if (dynamic_object) {
     roughness = 1.0 - roughness;
}
normal_roughness.w = roughness;
```
This allows us to take advantage of the fact that 0.5 can't be stored on UNORM, given the odd range normalization.

Hence decoding is just:

```glsl
float roughness = normal_roughness.w;
bool dynamic_object = roughness > 0.5;
if (dynamic_object) {
    roughness = 1.0 - roughness;
}
roughness /= (127.0/255.0);
```

This way we keep roughness == 0 intact and we can safely retrieve a dynamic object flag.

clayjohn:
@reduz Updated and switched back to UNORM. Everything works great still. I actually got the error down a little bit during the update, so its slightly higher quality too.

reduz (approved):
Looks fantastic!

YuriSizov:
Thanks!

PR #86322, opened 2023-12-19
[MenuBar] Store and use global menu item indices.

bruvzg:
- Fixes error message on exit (on macOS).
- Fixes updating main menu properties, when popup for special system menu is added in the middle of menu (and child node index is not global matching menu item index).


akien-mga (approved):

akien-mga:
Thanks!

PR #86325, opened 2023-12-19
Enable global classes in subfolders of res://addons/

### LINKED ISSUE #78400: Create dialog does not show global classes from addons inside subfolders
### Godot version

4.1.beta1.mono

### System information

Godot v4.1.beta1.mono - macOS 13.3.1 - Vulkan (Forward+) - integrated Apple M1 Pro - Apple M1 Pro (8 Threads)

### Issue description

The new `GlobalClass` attribute for registering global types in C# (#72619) does not seem to work in nested addons, i.e. addons that are placed within a subfolder of the `addons` directory (`res://addons/**/<addon>`).
Placing addons inside of subfolders is supported accroding to #43734

Edit: as confirmed by @raulsntos the same problem is present for GDScript.

#### Detailed showcase of the problem

Given the following project:
image

Where the `CustomProjectNode.cs`, `CustomAddonNode.cs`, and `CustomSubfolderAddonNode.cs` are implemented like this: 
```c#
using Godot;

namespace Addon;

[GlobalClass]
public partial class CustomAddonNode : Node
{
    public override void _Ready()
    {
    }

    // Called every frame. 'delta' is the elapsed time since the previous frame.
    public override void _Process(double delta)
    {
    }
}
```

And both addons are enabled in the project settings, the `CustomSubfolderAddonNode` does not show up when adding a new node:
image

### Steps to reproduce

1. Create a new project
2. Create a new C# addon
3. Create a node with the `GlobalClass` attribute
4. Move the folder containing the addon to a subfolder inside of `addons`
5. Compile the C# project
6. Enable the addon inside the project settings
7. Try to add the custom node to a scene. The node does not show up in the search

[GlobalClassTest.zip](https://github.com/godotengine/godot/files/11781466/GlobalClassTest.zip)

emklasson:
This PR lets `GlobalClass` classes get added to the "Create New Node" dialog no matter where they are under `addons/` as long as it's not in a disabled plugin (or any subfolder of such as those are assumed to belong to the disabled plugin).

This also then lets you keep non-plugin files under `addons/` and have their `GlobalClass`es work, as seems encouraged by https://docs.godotengine.org/en/stable/tutorials/best_practices/project_organization.html saying "_In general, keep third-party resources in a top-level addons/ folder, even if they aren't editor plugins. This makes it easier to track which files are third-party._"

Fixes https://github.com/godotengine/godot/issues/78400.

The code could presumably be optimised some e.g. by not using `String::split`, but I'm guessing code clarity and brevity takes preference over squeezing out performance here?

### DIFF
--- a/editor/create_dialog.cpp
+++ b/editor/create_dialog.cpp
@@ -160,8 +160,14 @@ bool CreateDialog::_should_hide_type(const String &p_type) const {
 
 		String script_path = ScriptServer::get_global_class_path(p_type);
 		if (script_path.begins_with("res://addons/")) {
-			if (!EditorNode::get_singleton()->is_addon_plugin_enabled(script_path.get_slicec('/', 3))) {
-				return true; // Plugin is not enabled.
+			String base_path = "res://addons/";
+			Vector<String> folders = script_path.split("/");
+			for (int i = 3; i < folders.size() - 1; i++) {
+				base_path += folders[i] + "/";
+				String plugin = base_path + "plugin.cfg";
+				if (FileAccess::exists(plugin) && !EditorNode::get_singleton()->is_addon_plugin_enabled(plugin)) {
+					return true; // Plugin is not enabled.
+				}
 			}
 		}
 	}


ajreckof (approved):
LGTM. 
I don't see how to do it better then with `split` and based on what is done in `EditorPluginSettings::_get_plugins` it should be good as it is.

akien-mga:
Superseded by #91337. Thanks for the contribution!

PR #85167, opened 2023-11-21
Enhance SpinLock

RandomShaper:
As advised by https://rigtorp.se/spinlock/ and other resources, a good spinlock should do the following at least (and the following points are what this PR adds):
- Use the CPU pause instruction while looping, to inform the hardware we are spinning so it does the right thing. Depending on architectures, this may improve power efficiency, for instance (better usage of phone/laptop battery). Not only that, but hyper-threaded cores can also work better since one of the logical hardware threads knows the other isn't doing anything particularly useful while the latter is just spinning around.
- If the first attempt to lock fails, instead of blindlessly retrying, loop with relaxed memory ordering until it seems the lock is available again and, only then, retry with acquire semantics. This avoids wasteful sync operations, and may improve performanve.

**2024-10-23:**
~As a disclaimer, I haven't benchmarked this for either performance nor power usage. Maybe the Production team can lend a hand.~
Using a benchmark I've written for this (https://github.com/godotengine/godot-benchmarks/pull/95), these are the results on a Windows 10 PC, before (`a`) and after (`b`) this PR (unit is time, lower is better; note how this PR highly improves the contended cases):

**2024-11-04:** Benchmarked after going back to the CPU pause approach:
```
{
        "category": "Core > Object Db",
        "name": "12 Threads Full Contention",
        "results": {
                "time": {
                        "a": 1631,
                        "b": 1227,
                        "a_div_b": 1.3292583537082314
                }
        }
}
{
        "category": "Core > Object Db",
        "name": "12 Threads Half Contention",
        "results": {
                "time": {
                        "a": 834.1,
                        "b": 631.4,
                        "a_div_b": 1.3210326259106748
                }
        }
}
{
        "category": "Core > Object Db",
        "name": "12 Threads Little Contention",
        "results": {
                "time": {
                        "a": 186.8,
                        "b": 149.9,
                        "a_div_b": 1.246164109406271
                }
        }
}
{
        "category": "Core > Object Db",
        "name": "12 Threads Slope Contention",
        "results": {
                "time": {
                        "a": 760.3,
                        "b": 614.2,
                        "a_div_b": 1.2378704005210028
                }
        }
}
{
        "category": "Core > Object Db",
        "name": "2 Threads Full Contention",
        "results": {
                "time": {
                        "a": 57.44,
                        "b": 61.74,
                        "a_div_b": 0.9303530936183997
                }
        }
}
{
        "category": "Core > Object Db",
        "name": "2 Threads Half Contention",
        "results": {
[...]

PR #86333, opened 2023-12-19
Make `RID_Owner` lock-free for fetching.

reduz:
This PR makes RID_Owner lock free for fetching values, this should give a very significant peformance boost where used.

Some considerations:

* A maximum number of elements to alocate must be given (by default 200k).
* Merged the validators and the data into a single struct, to make it more cache access friendly.
* Uses a mutex for thread safety since nowadays spinlocks are discouraged.
* Access to the RID structure is still safe given they are independent from addition/removals.
* RID access was never really thread-safe in the sense that the contents of the data are not protected anyway. Each server needs to implement locking as it sees fit.

### DIFF
--- a/core/templates/rid_owner.h
+++ b/core/templates/rid_owner.h
@@ -32,7 +32,7 @@
 #define RID_OWNER_H
 
 #include "core/os/memory.h"
-#include "core/os/spin_lock.h"
+#include "core/os/mutex.h"
 #include "core/string/print_string.h"
 #include "core/templates/hash_set.h"
 #include "core/templates/list.h"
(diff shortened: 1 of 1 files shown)


core/templates/rid_owner.h
```diff
@@ -69,42 +69,47 @@ class RID_AllocBase {
...
 	_FORCE_INLINE_ RID _allocate_rid() {
 		if (THREAD_SAFE) {
-			spin_lock.lock();
+			mutex.lock();
 		}
 
 		if (alloc_count == max_alloc) {
 			//allocate a new chunk
 			uint32_t chunk_count = alloc_count == 0 ? 0 : (max_alloc / elements_in_chunk);
+			ERR_FAIL_COND_V(THREAD_SAFE && chunk_count == chunk_limit, RID());
```
AThousandShips, line 96: This will leave the mutex permanently locked

reduz: Oh you rock, nicely spotted, thanks!

core/templates/rid_owner.h
```diff
@@ -69,42 +69,54 @@ class RID_AllocBase {
...
 	_FORCE_INLINE_ RID _allocate_rid() {
 		if (THREAD_SAFE) {
-			spin_lock.lock();
+			mutex.lock();
 		}
 
 		if (alloc_count == max_alloc) {
 			//allocate a new chunk
 			uint32_t chunk_count = alloc_count == 0 ? 0 : (max_alloc / elements_in_chunk);
+			if (THREAD_SAFE && chunk_count == chunk_limit) {
```
RandomShaper, line 96:
```suggestion
			if (unlikely(THREAD_SAFE && chunk_count == chunk_limit)) {
```

RandomShaper (commented):
If I understand correctly, the thread-safe version has pre-allocate because otherwise `memrealloc()` can change the base pointer of the data. I'm wondering if that could be solved by using a `PagedAllocator` (**UPDATE:** or `PagedArray`, maybe is what I mean). Maybe the downside is that element access needs a bit of extra arithmetic and, more prominently, indirection to reach the item on the right page. Can you confirm?

RandomShaper:
Aside, given there's no need for thread safety on the data itself (requires external sync anyway), if we couldn't go fully lock-free, this way:
- If thread-safe, `validator` becomes an `std::atomic<uint32_t>`, with load-acquire and store-release used on it.
- If not thread-safe, it keeps being a raw `uint32_t`.

PS: For the records, there's an idea I don't want to forget and so I'm writing it here: in case we still want locking, the `SpinLock` may still be advantageous, as it's sync variable could be used even for non-locky operations; e.g., to acquire-load on them. To avoid some of the downsides of it, we'd want to polish #85167. But, again, this is not relevant at the current status of the discussion.

reduz:
@RandomShaper I am wondering if making the validator an atomic really changes anything, my fear is that load-acquire still forces the other CPUs to finalize other memory writes.

RandomShaper:
Acquire-release atomics would at least be superior to either `Mutex` or `SpinLock`. The problem is thus with the functions that are now lock-less in this PR even in the thread-safe version. If we are fine with readers potentially seeing old data, those reads could be relaxed, but they have to be atomic in any case (in most or all relevant architectures that will boil down to plain reads, but the code would be more future-proof that way).

reduz:
@RandomShaper I think in theory it should be fine if they see old data in this case, the data will not be accessible anyway after the mutex ends the lock, and will be updated when removed after the mutex also ends the lock. In the meantime, nothing should bother the data access so I think it should be ok. The data itself is not thread safe anyway, even now, so its not related to the actual RID access per se.

Calinou (approved):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/10e111477db68fe65776a1d68fb1ffccaf6520fc + https://github.com/godotengine/godot/pull/86333), it works as expected on Linux + NVIDIA.

On https://github.com/godotengine/tps-demo, it loads a full second faster than on `master` with neither PR applied.

core/templates/rid_owner.h
```diff
@@ -294,43 +287,48 @@ class RID_Alloc : public RID_AllocBase {
...
+			mutex.unlock();
 		}
 	}
 
 	void set_description(const char *p_descrption) {
 		description = p_descrption;
 	}
 
-	RID_Alloc(uint32_t p_target_chunk_byte_size = 65536) {
+	RID_Alloc(uint32_t p_target_chunk_byte_size = 65536, uint32_t p_maximum_amount_of_elements = 200000) {
```
Calinou, line 325: Out of curiosity, shouldn't the default value for `p_maximum_amount_of_elements` be a multiple of `65536`?

reduz: yeah thats a good point

clayjohn:
For context, we need to merge this before merging https://github.com/godotengine/godot/pull/87590, but we are in no rush to merge https://github.com/godotengine/godot/pull/87590 for now, so no pressure

akien-mga:
Superseded by #97465.

PR #86335, opened 2023-12-19
Add `is_colliding` method to `SpringArm3D`

### LINKED ISSUE #8691: Add is_colliding() method to SpringArm3D

A third person 3D action-adventure game

TL;DR: No current way to detect if the ray/shape cast is actually colliding with something.

Trying to implement smoother movement for the camera, I found there's no built-in way to do this in the SpringArm3D node. But since I'm trying to use as much existing functionality as possible, I still wanted to work with this node. In my script I stored the intended arm length (aka zoom level) and my intention was to set the spring_length to the result of get_hit_length() when the arm is retracted due to collision and then interpolate it towards the stored value. Problem is, if I set spring_length to hit_length, there's no other way to tell if we're no longer colliding and to start moving the retracted arm back where it should be.

I've added an is_colliding() method to SpringArm3D to have a way of detecting if arm is retracted independent of the get_hit_length() results. Especially considering that get_hit_length() always returns a value, and it's not obvious from the docs.

In parity with RayCast3d and ShapeCast3d I've added a `bool collided` property and a `bool is_colliding()` const method that returns its value. The value is set to `collided = motion_delta < 1.0` where motion_delta is set by existing code to the results of intersect_ray or cast_motion.
- See godotengine/godot/pull/86335

If not required, this method can just be ignored by the user.

SpringArm3D is an exisitng node and core part of the engine.

Illauriel:
Adds a method to getect if the arm is currently retracted in parity with Raycast3D and Shapecast3D. 

For me this method was helpful to set up a third person camera which returns smoothly to its intended length after being retracted due to collision.

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/8691*

Edit: Since first making this little hack I also used it pretty successfully in my character controller, saving me a few extra raycasts. And since Springarm IS in essence a cast, using an additional raycast to detect arm's collision defeats its purpose since you could simply place the object at the contact point using just the raycast. :)

### DIFF
--- a/doc/classes/SpringArm3D.xml
+++ b/doc/classes/SpringArm3D.xml
@@ -28,6 +28,12 @@
 				Returns the spring arm's current length.
 			</description>
 		</method>
+		<method name="is_colliding" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns whether the SpringArm3D's ray or shape is intersecting with any colliders, and thus, if the spring arm's current length is less than its maximum length.
+			</description>
+		</method>
 		<method name="remove_excluded_object">
 			<return type="bool" />
 			<param index="0" name="RID" type="RID" />
--- a/scene/3d/physics/spring_arm_3d.cpp
+++ b/scene/3d/physics/spring_arm_3d.cpp
@@ -55,6 +55,7 @@ void SpringArm3D::_notification(int p_what) {
 
 void SpringArm3D::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_hit_length"), &SpringArm3D::get_hit_length);
+	ClassDB::bind_method(D_METHOD("is_colliding"), &SpringArm3D::is_colliding);
 
 	ClassDB::bind_method(D_METHOD("set_length", "length"), &SpringArm3D::set_length);
 	ClassDB::bind_method(D_METHOD("get_length"), &SpringArm3D::get_length);
@@ -130,6 +131,10 @@ real_t SpringArm3D::get_hit_length() {
 	return current_spring_length;
 }
 
+bool SpringArm3D::is_colliding() const {
+	return collided;
+}
+
 void SpringArm3D::process_spring() {
 	// From
 	real_t motion_delta(1);
@@ -187,6 +192,8 @@ void SpringArm3D::process_spring() {
 
 		get_world_3d()->get_direct_space_state()->cast_motion(shape_params, motion_delta, motion_delta_unsafe);
 	}
+
+	collided = motion_delta < 1.0;
 
 	current_spring_length = spring_length * motion_delta;
 	Transform3D child_transform;
--- a/scene/3d/physics/spring_arm_3d.h
+++ b/scene/3d/physics/spring_arm_3d.h
@@ -36,6 +36,7 @@
 class SpringArm3D : public Node3D {
 	GDCLASS(SpringArm3D, Node3D);
 
+	bool collided = false;
 	Ref<Shape3D> shape;
 	HashSet<RID> excluded_objects;
 	real_t spring_length = 1.0;
@@ -59,6 +60,7 @@ public:
 	bool remove_excluded_object(RID p_rid);
 	void clear_excluded_objects();
 	real_t get_hit_length();
+	bool is_colliding() const;
 	void set_margin(real_t p_margin);
 	real_t get_margin();
 


AThousandShips:
Please open a [proposal](https://github.com/godotengine/godot-proposals/issues) for this new feature, a this type is not really the same as the other types you propose parity with, so this needs to be discussed 🙂

Illauriel:
> Please open a [proposal](https://github.com/godotengine/godot-proposals/issues) for this new feature, a this type is not really the same as the other types you propose parity with, so this needs to be discussed 🙂

Ah, okay, I'm on it^^

scene/3d/spring_arm_3d.cpp
```diff
@@ -187,6 +192,8 @@ void SpringArm3D::process_spring() {
 
 		get_world_3d()->get_direct_space_state()->cast_motion(shape_params, motion_delta, motion_delta_unsafe);
 	}
+
```
AThousandShips, line 195: This line has whitespace, you need to remove it

scene/3d/spring_arm_3d.cpp
```diff
@@ -187,6 +192,7 @@ void SpringArm3D::process_spring() {
 
 		get_world_3d()->get_direct_space_state()->cast_motion(shape_params, motion_delta, motion_delta_unsafe);
 	}
+	collided = motion_delta < 1.0;
```
AThousandShips, line 195:
```suggestion

	collided = motion_delta < 1.0;
```
Keep the empty line, just remove the trailing whitespace 🙂

scene/3d/spring_arm_3d.cpp
Illauriel: Sorry! Should be good now^^

scene/3d/spring_arm_3d.cpp
Illauriel: S-should be good now ><

doc/classes/SpringArm3D.xml
```diff
@@ -28,6 +28,12 @@
 				Returns the spring arm's current length.
 			</description>
 		</method>
+		<method name="is_colliding" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns whether the SpringArm3D's ray or shape is intersecting with any colliders, and thus, if the spring arm's current length is less than its maximum length.
```
MewPurPur, line 34:
```suggestion
				Returns whether the [SpringArm3D]'s ray or shape is intersecting with any colliders, and thus, if the spring arm's current length is less than its maximum length.
```

AThousandShips: Unnecessary here I'd say, it's the same class, and isn't done anywhere else in the file

doc/classes/SpringArm3D.xml
```diff
@@ -28,6 +28,12 @@
 				Returns the spring arm's current length.
 			</description>
 		</method>
+		<method name="is_colliding" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns whether the [SpringArm3D]'s ray or shape is intersecting with any colliders, and thus, if the spring arm's current length is less than its maximum length.
```
AThousandShips, line 34:
```suggestion
				Returns whether the SpringArm3D's ray or shape is intersecting with any colliders, and thus, if the spring arm's current length is less than its maximum length.
```
As I mentioned in the suggestion, this isn't done anywhere else in the file so doing it here feels out of place

Illauriel: Ah. Alright, let me roll it back^^'

doc/classes/SpringArm3D.xml
MewPurPur: Hm okay, it would be best to change these as a sweeping change then. Same class is displayed with bold letters, making it more readable.

AThousandShips: Agreed, but there are seven other cases so it feels like it'd be confusing 🙂

Illauriel: > Hm okay, it would be best to change these as a more sweeping change then. Same class is displayed with bold letters, making it more readable.

I wouldn't mind adding it to the entire file. Should I proceed?^^

aaronfranke:
@Illauriel You accidentally rebased this on top of the 4.3 branch for 4.3.x, instead of the master branch for 4.x.

Illauriel:
Oh sh... my bad ^^' Brb, rebasing to master

doc/classes/SpringArm3D.xml
AThousandShips: I'd say to leave it now and fix in a separate PR

PR #72789, opened 2023-02-06
Fixed the jumping to function definition using 'Ctrl+LMB' (reverted)

Vilcrow:
Fixes #72353.

PR #86341, opened 2023-12-19
Fix the autocomplete function for the `self` keyword.

### LINKED ISSUE #85833: `self.` completion suggesting incorrect code
### Tested versions

Tested in v4.2.stable.official [46dc27791], v4.2.1.stable.official [b09f793f5].

### System information

Godot v4.2.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated Radeon RX 580 Series (Advanced Micro Devices, Inc.; 31.0.21905.1001) - AMD Ryzen 5 3600 6-Core Processor (12 Threads)

### Issue description

The self keyword are assumed to be classes by code completion (unlike before in v4.1 and previous versions), and only display constants, static functions, and new().

Edit: After further investigation by @Vilcrow, I jumped the gun on Singletons suffering the same issue as the self keyword. Apologies to @MitrB for stealing your thunder on this one.

### Steps to reproduce

- In any project open in version 4.2, create or open a script.
- Type "self." and observe the code completion suggestions.
- You should see something like this:
Screenshot 2023-12-06 061131

[BugProject.zip](https://github.com/godotengine/godot/files/13581582/BugProject.zip)

Vilcrow:
Fixes #85833 

UPD: <del>I will try to fix the bug with singleton in the coming days. While I'm converting the PR into a draft.</del> there is no bug with singleton.

### DIFF
--- a/modules/gdscript/gdscript_editor.cpp
+++ b/modules/gdscript/gdscript_editor.cpp
@@ -1494,11 +1494,8 @@ static bool _guess_expression_type(GDScriptParser::CompletionContext &p_context,
 			} break;
 			case GDScriptParser::Node::SELF: {
 				if (p_context.current_class) {
-					if (p_context.type != GDScriptParser::COMPLETION_SUPER_METHOD) {
-						r_type.type = p_context.current_class->get_datatype();
-					} else {
-						r_type.type = p_context.current_class->base_type;
-					}
+					r_type.type = p_context.current_class->get_datatype();
+					r_type.type.is_meta_type = false;
 					found = true;
 				}
 			} break;


am45-001:
After implementing this pull request and compiling the 4.2.1 source code, it still fails to treat Singletons properly. 
```self.``` completion is fixed with this though. I haven't tried the master branch yet, so maybe there's something different there.

adamscott:
CC. @dalexeev @anvilfolk

Vilcrow:
I didn't find any problems with singleton. Everything seems to be working. Describe in more detail what it is @am45-001

am45-001:
@Vilcrow , after taking another look, I seem to have been mistaken about singletons, and RenderingServer in particular.

HolonProduction:
As stated on the original issue, I don't quite understand the reason for the `if` statement. While you are at this code again, could you maybe add an explanatory comment, since its purpose isn't directly obvious.

vnen:
> As stated on the original issue, I don't quite understand the reason for the `if` statement. While you are at this code again, could you maybe add an explanatory comment, since its purpose isn't directly obvious.

I also don't understand how this could ever be `false`. With `self` being used explicitly, there's no way to also use `super` AFAICT.

In any case, even if the `if` is necessary for some reason, the `r_type.type.is_meta_type = false` line should be done after the `if` block, as it would apply to both branches.

vnen (approved):

Vilcrow:
Initially, this code snippet was entered elsewhere(see #72789), where the `if` block made sense. Then it was moved here without the necessary adjustments(my bad).

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

PR #84909, opened 2023-11-14
Expose a way to get the topmost Control node under the mouse

cdoise-vbg:
I've exposed a way to get Viewport::gui.mouse_over and call Viewport::gui_find_control to GDScript to facilitate being able to implement the following features:
- Gamepad control of mouse where speed of mouse should slow down when the cursor is over certain types of controls
- Allowing for a generic service that can provide contextual help based on the Control under the mouse cursor

From GDScript, the following can now be done to get the control under the mouse cursor (or possibly at another location):
```
var mouse_over_control = get_viewport().gui_get_mouse_over()
var found_control = get_viewport().gui_find_control(get_global_mouse_position())
```

PR #86343, opened 2023-12-19
Expose viewport `gui_find_control` to scripting

cdoise-vbg:
Resubmitting relevant changes from this pull request to expose gui_find_control to GDScript. Another pull request had already been merged in that covered getting the mouse_over control
https://github.com/godotengine/godot/pull/84909

Usage:
```gdscript
var control_under_point = get_viewport().gui_find_control(get_global_mouse_position())
print(control_under_point)
```

### DIFF
--- a/doc/classes/Viewport.xml
+++ b/doc/classes/Viewport.xml
@@ -120,8 +120,7 @@
 			<return type="Control" />
 			<param index="0" name="point" type="Vector2" />
 			<description>
-				Returns the [Control] that would be under the mouse cursor if the mouse cursor were located at the
-				viewpoint coordinates provided by the point parameter. If no [Control] is at that point, returns null.
+				Returns the [Control] that would be under the mouse cursor if the mouse cursor were located at the viewport coordinates provided by the point parameter. Returns null if there is no [Control] at that point.
 			</description>
 		</method>
 		<method name="gui_get_drag_data" qualifiers="const">
--- a/doc/classes/Viewport.xml
+++ b/doc/classes/Viewport.xml
@@ -116,6 +116,14 @@
 				Returns the visible rectangle in global screen coordinates.
 			</description>
 		</method>
+		<method name="gui_find_control">
+			<return type="Control" />
+			<param index="0" name="point" type="Vector2" />
+			<description>
+				Returns the [Control] that would be under the mouse cursor if the mouse cursor were located at the
+				viewpoint coordinates provided by the point parameter. If no [Control] is at that point, returns null.
+			</description>
+		</method>
 		<method name="gui_get_drag_data" qualifiers="const">
 			<return type="Variant" />
 			<description>
--- a/scene/main/viewport.cpp
+++ b/scene/main/viewport.cpp
@@ -4564,6 +4564,8 @@ void Viewport::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("gui_get_focus_owner"), &Viewport::gui_get_focus_owner);
 	ClassDB::bind_method(D_METHOD("gui_get_hovered_control"), &Viewport::gui_get_hovered_control);
 
+	ClassDB::bind_method(D_METHOD("gui_find_control", "point"), &Viewport::gui_find_control);
+
 	ClassDB::bind_method(D_METHOD("set_disable_input", "disable"), &Viewport::set_disable_input);
 	ClassDB::bind_method(D_METHOD("is_input_disabled"), &Viewport::is_input_disabled);
 


AThousandShips:
The bind isn't just exposed to GDScript, it's also available to C# and GDExtension, please update your commit message accordingly

doc/classes/Viewport.xml
```diff
@@ -116,6 +116,13 @@
 				Returns the visible rectangle in global screen coordinates.
 			</description>
 		</method>
+		<method name="gui_find_control">
+			<return type="Control" />
+			<param index="0" name="point" type="Vector2" />
+			<description>
+				Returns the top-level [Control] at the given point within this viewport. If no [Control] is at the point, returns null.
```
AThousandShips, line 123: This needs some more detail IMO, what does it mean "at the given point"? Does it mean:
* The point is within its bounds?
* Does it have a visible pixel in that point? (i.e. `_has_point` returns true)
* Or something else?

Details on this would be required to know how to use this properly

Sauermann, line 123: The wording "top-level Control" is ambiguous in this case and should be avoided.
For Control nodes there is the CanvasItem property `top_level` which could be confused.
See https://docs.godotengine.org/en/latest/classes/class_canvasitem.html#class-canvasitem-property-top-level

Sauermann:
> `var control_under_point = get_viewport().gui_find_control(get_global_mouse_position())`

This example will not work when the Canvas layer has a non-default transform, since your are mixing different coordinate systems:
- `CanvasItem.get_global_mouse_position()` returns the position in canvas layer coordinates
- `Viewport.gui_find_control()` accepts the position in viewport coordinates.

Please change your description to:
``` gdscript
var control_under_point = get_viewport().gui_find_control(get_viewport().get_mouse_position())
```

cdoise-vbg:
Thank you for the feedback. I'll work on updating the description. Two
scenarios I was interested in supporting with this are:
- Second player with a custom gamepad-controlled virtual cursor being able
to click on Controls.
- Testing logic that can do basic checks for the presence of a Control at a
particular coordinate on-screen.

cdoise-vbg:
I updated the description to be more specific about which coordinates are expected based on comments above. Should I include the specific code example too in that particular comment location? (meaning: "var control_under_point = get_viewport().gui_find_control(get_viewport().get_mouse_position())").

I also need to go back and fix up the 2 commits so that they appear as 1.

doc/classes/Viewport.xml
```diff
@@ -116,6 +116,14 @@
 				Returns the visible rectangle in global screen coordinates.
 			</description>
 		</method>
+		<method name="gui_find_control">
+			<return type="Control" />
+			<param index="0" name="point" type="Vector2" />
+			<description>
+				Returns the [Control] that would be under the mouse cursor if the mouse cursor were located at the
+				viewpoint coordinates provided by the point parameter. If no [Control] is at that point, returns null.
```
Sauermann, line 124:
```suggestion
				viewport coordinates provided by the point parameter. If no [Control] is at that point, returns null.
```

AThousandShips: This should also be a single line, the current text is two separate paragraphs

AThousandShips, line 124:
```suggestion
				Returns the [Control] that would be under the mouse cursor if the mouse cursor were located at the viewport coordinates provided by the point parameter. Returns null if there is no [Control] at that point.
```
As per above, and more direct wording IMO

doc/classes/Viewport.xml
cdoise-vbg: Thanks for your help on fixing this description. I committed your suggested changes down below.

cdoise-vbg: Thanks for your help on reviewing this. I committed the suggested changes from AThousandShips down below.

cdoise-vbg:
Thank you @AThousandShips for the rewording of the documentation. I've committed that from the button to accept suggested changes in the comments above. Should I now go back and squash those two commits back down to one?

name212:
@AThousandShips  will this pull reqest be merged?

AThousandShips:
It hasn't been approved yet, only once approved will it be considered for merger, we're currently in feature freeze for 4.3 so this can't be merged until after 4.3 has been released, but again after approval

cdoise-vbg:
Do I need to do anything in particular to request approval or further comments for the latest check-in from December 31?

AThousandShips:
Not that I know, but once 4.3 has been released you can write in the [developer chat](https://chat.godotengine.org) and see if anyone can take a look, feel free to join already to keep up to date with development

badsectoracula:
I fixed the conflict and did a quick test with an editor script and this seemed to work fine and as expected.

You may want to squash the commits again to 1 and fix the conflict.

Sauermann:
Please note, that `Viewport::gui_find_control` doesn't take embedded windows into account.
So the caller of this function would need to make sure, that there is no embedded window at the specified position.
Otherwise the function could return a `Control`-node, that is hidden behind an embedded window.

So I'm not sure, if exposing this function as it is, is a good idea, because people could get unexpected results, if they are unaware of these details. This implementation detail also caused problems inside the engine, which was recently discovered: see #99849

Would the following idea be better?
Expose a function `Node *Viewport::gui_find_node_at_position(const Point2 p_position)`, which returns
- the embedded window at `p_position` or otherwise
- the Control node at `p_position` or
- `nullptr`, if neither of the above conditions fit.

PR #86353, opened 2023-12-20
Fix Instance float uniform doesn't work if assigned an integer

### LINKED ISSUE #86130: Instance float uniform doesn't work if assigned an integer, unlike regular uniforms
### Tested versions

4.2 stable

### System information

Windows 10 64 bits NVIDIA GeForce GTX 1060

### Issue description

Using `set_instance_shader_parameter` with an integer when the shader declares the parameter as `instance uniform float` acts as if it doesn't assign the value. This is confusing because regular float uniforms just work when passed an int.

### Steps to reproduce

1) Create a shader declaring a `instance uniform float u_param;` and makes use of that parameter in a visible way, like showing as green color in albedo.
2) Create a `MeshInstance3D` with a box mesh
3) Create a `ShaderMaterial` assigned to `material_override`, then assign the shader to that material
4) Create a script that assigns the mesh instance's instance shader parameter using `set_instance_shader_parameter(&"u_param", 1)`
5) Run the scene: observe the parameter has no effect (not expected)
6) Change the passed value syntax to `1.0` or `float(1)`
7) Run the scene: observe the parameter now has an effect

[InstanceUniformFloat.zip](https://github.com/godotengine/godot/files/13665492/InstanceUniformFloat.zip)

The main scene contains 4 cubes:
- With instance uniform set with an int
- With instance uniform set with a float
- With regular uniform set with an int
- With regular uniform set with a float

Each ShaderMaterial is unique (though that should not matter for the instance uniforms).

Run the scene: observe the cube that was given an instance uniform with an int doesn't turn green, while all the others do.

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/86130

In a nutshell, the problem lies in the fact that `global_shader_parameters_instance_update` (when set instance uniform) does not take in a type parameter, it just uses the type of the value passed in, while in `MaterialData::update_uniform_buffer` (when set uniform) it passes the correct type to `_fill_std140_variant_ubo_value`.

So in one hand we cound add an additional parameter to `global_shader_parameters_instance_update` and give it the correct type, but that requires a bigger change. 

So what I try to do in this pr is that in `instance_geometry_set_shader_parameter`, do some check about whether the type conversion is valid and do the conversion early (keep the variant's value but use the instance's type). Your suggestions are welcomed as I'm not familiar to the variant system...

### DIFF
--- a/servers/rendering/renderer_scene_cull.cpp
+++ b/servers/rendering/renderer_scene_cull.cpp
@@ -1488,6 +1488,8 @@ void RendererSceneCull::instance_geometry_set_shader_parameter(RID p_instance, c
 
 	ERR_FAIL_COND(p_value.get_type() == Variant::OBJECT);
 
+	_update_dirty_instance(instance);
+
 	HashMap<StringName, Instance::InstanceShaderParameter>::Iterator E = instance->instance_shader_uniforms.find(p_parameter);
 
 	if (!E) {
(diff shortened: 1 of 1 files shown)


servers/rendering/renderer_scene_cull.cpp
```diff
@@ -1497,7 +1497,15 @@ void RendererSceneCull::instance_geometry_set_shader_parameter(RID p_instance, c
 		isp.value = p_value;
 		instance->instance_shader_uniforms[p_parameter] = isp;
 	} else {
-		E->value.value = p_value;
```
jsjtxietian, line 1500: Update:

I noticed that my pr only works if an debugger attached, or be more precisely, only after the `instance->instance_shader_uniforms` is correctly inited by other code, or it will go to the `if(!E)` branch, it doesn't feels right because in that branch the type info is lost.

Sabrejack:
Would this also cover the case of assigning a Vector3i() to a instance uniform vec3?  Because I get the same behavior as described in issue #86130

jsjtxietian:
> Would this also cover the case of assigning a Vector3i() to a instance uniform vec3? Because I get the same behavior as described in issue #86130

It maybe as the code doesn't care whether it's float or vec3, but it needs some test.

PR #56931, opened 2022-01-19
Enforce mult-of-4 requirements on etcpak input.

lyuma:
Re: https://github.com/godotengine/godot/issues/49981

Adds back a variation old code from modules/etc/image_compress_etc.cpp to enforce ~~power-of-two (mipmapped) and~~ multiple of 4 ~~(non-mipmapped)~~ on compressed textures.

Here is the original code from before etcpak was added:
https://github.com/godotengine/godot/pull/47370/files#diff-d43a3b757699a828b4602b39b53ce14bc22f35556ed7ea2d3943608f90bf4d34

There was something else [added by reduz in February, 2019](https://github.com/godotengine/godot/commit/74d0ed2236b333d0a292e15b9c2a9380c2c122a1#diff-4949ecbffc5fadfd3ef6ff61bcc12d88c51ea8fd1bc17d586f624b9b56685fa6R422-R427) to enforce power-of-two, but it was only intended for GLES2 and doesn't seem to be in the right place in the code to solve this particular issue.

Intermediate mipmap levels need to be resized to Nx4 or 4xN accordingly, since etcpak is stricter about padding.

Not doing so causes the following artifact when sampling the 2x2 or 1x1 mipmap levels:
image
It looks normal when samplihg 4x4 or higher:
image
Here is what I think is happening before this PR.
```
		8x8 data...
		4x4 data...
		2x2 data...
		1x1 data...
		8888888888888888..8888844444444444444422221xxxxxxxxxxxxxxxxxx <-- current state.
last two mipmap levels sample out-of-bounds
possible idea: resize Nx2 and 2xN and lower to multiple of 4, using a temporary buffer
		8888888888888888..8888844444444444444422222222222222221111111111111111
```
I have now implemented the above.

*Bugsquad edit:*
- Fixes #49981

PR #86360, opened 2023-12-20
Generate complete mipmap chains for compressed DDS images

### LINKED ISSUE #86330: DDS with implicit first mipmap size fails to import

LunaticInAHat:
This PR adjusts the DDS image loader, to ensure that the mipmap chain of compressed images is complete. This aligns with expectations elsewhere throughout the engine, that images with mipmaps will be mipmapped all the way down to 1x1, and will not have partial mipmap chains that terminate at some higher level.

Some DDS exporters create images whose mipmap chains terminate above the 1x1 case (often 4x4 or 8x8). To create the missing mipmap levels, this PR repeatedly copies the first block of compressed data from the last mipmap level that exists in the DDS. This is not mathematically correct, however the hope is that it will create mipmaps with colors that are similar enough to correct, as to not be visually objectionable, given that these low mipmap levels will only be occupying a few pixels on the screen.

The "correct" solution would require decoding the image, resampling it, re-encoding into a compressed form, and copying the result back into the original image data. I felt it was best not to incur that performance & complexity overhead on image loading, unless and until complaints are received about last-level mipmaps of DDS textures looking wrong. The limited color palette available in DXT-compressed pixel blocks means that low-level mipmaps tend not to look much like the original image, anyways.

This PR does not adjust the handling of compressed images which did not have any mipmaps populated in the source file; those images will be left un-mipmapped, as I believe the remainder of the engine is prepared to accept them in that state.

This PR resolves #86330

### DIFF
--- a/modules/dds/texture_loader_dds.cpp
+++ b/modules/dds/texture_loader_dds.cpp
@@ -410,17 +410,49 @@ Ref<Resource> ResourceFormatDDS::load(const String &p_path, const String &p_orig
 			ERR_FAIL_COND_V_MSG(pitch != 0, Ref<Resource>(), "DDS header flags specify that no linear size will given for the top-level image, but a non-zero linear size value is present in the header.");
 		}
 
+		uint32_t last_mip_offset = 0u;
 		for (uint32_t i = 1; i < mipmaps; i++) {
 			w = MAX(1u, w >> 1);
 			h = MAX(1u, h >> 1);
 
 			uint32_t bsize = MAX(info.divisor, w) / info.divisor * MAX(info.divisor, h) / info.divisor * info.block_size;
+			last_mip_offset = size;
 			size += bsize;
 		}
 
 		src_data.resize(size);
 		uint8_t *wb = src_data.ptrw();
 		f->get_buffer(wb, size);
+
+		const uint32_t last_mip_w = w;
+		const uint32_t last_mip_h = h;
+		if ((mipmaps >= 2u) && ((last_mip_w >= 2u) || (last_mip_h >= 2u))) {
+			// The source image's mipmap chain ends above the 1x1 base-case, however the renderer expects textures to have complete mipmap chains.
+			// Thus, we generate (very crude) mipmaps for the missing levels.
+			uint32_t mip_gen_offset = size;
+			while ((w >= 2u) || (h >= 2u)) {
+				w = MAX(1u, w >> 1);
+				h = MAX(1u, h >> 1);
+
+				uint32_t bsize = MAX(info.divisor, w) / info.divisor * MAX(info.divisor, h) / info.divisor * info.block_size;
+				size += bsize;
+			}
+			src_data.resize(size);
+			w = last_mip_w;
+			h = last_mip_h;
+			while ((w >= 2u) || (h >= 2u)) {
+				w = MAX(1u, w >> 1);
+				h = MAX(1u, h >> 1);
+
+				uint32_t blocks_width = MAX(info.divisor, w) / info.divisor;
+				uint32_t blocks_height = MAX(info.divisor, h) / info.divisor;
+				for (uint32_t idx = 0u; idx < (blocks_width * blocks_height); idx++) {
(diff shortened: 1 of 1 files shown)


fire:
@lyuma and a few others debuggined this in the etcpak. I'll try to find the details

LunaticInAHat:
> See #56931

Thanks for the reference. I think I have avoided the particular pitfall that the referenced PR was fixing. If I am reading it right, it was mainly concerned with ensuring that compressed image data generated by Godot was correctly formed from complete BC/DXT blocks. The implementation being fixed in that PR generated last-level mipmaps that occupied less than a full block, which resulted in garbage when sampled by the GPU. The corrected implementation from that PR ensured that even the very small mip levels generated full DXT blocks.

Similar to the corrected implementation of that PR, my PR here also generates complete blocks of BC/DXT data for the low-level mipmaps. So, they should sample correctly. They are incorrect in the sense that the colors encoded in them are not the colors that would be present in the best possible case, however they are correctly formed.

The other PR appears to have had access to the uncompressed color data of the image, so when generating the low-level mipmaps, they "smear" those colors across the 4x4 block that then goes into image encoding and pops out as BC/DXT data. The DDS loader does not have access to the uncompressed color data; we only have the compressed representation handy. So, we can't replicate the resampling technique from the other PR, unless we wanted to actually decompress the image first.



As a concrete example of how my implementation compares with a best-possible resampling of the image data: Pretend you have an image of arbitrary size with Red in the top left, Green top right, Blue bottom left, Yellow bottom right. There is a smooth gradient across the image, blending between the colors.

The best-possible 4x4 mipmap of that image will have at least two of the right colors in the right corners. The other two corners (and most of the middle pixels) will all contain incorrect colors. Exactly how incorrect the image is will depend upon how smart the encoder is when choosing which colors to encode explicitly; there are only two explicitly-specified colors in BC/DXT blocks, and all other colors referencable by the block are linearly interpolated between them. So, you can get two corners correct, and you can probably even get a band of the gradient right, but the rest is all forced to be incorrect garbage. If you chose to encode Red and Green, then there is no interpolation that gets you Blue or (full-bright) Yellow for the other corners. The 2x2 mipmap is essentially the same story; two corners are correct, the other two are wrong. The 1x1 mipmap can at least be precisely generated -- it will be some kind of grey, because that's the average of those colors.

The incorrect 4x4 mipmap that will be generated by this PR is just a duplication of the 4x4 top-left corner of the image. So, it gets the Red corner right, but everything else is varying shades of wrong. The 2x2 mipmap will be the 2x2 top-left corner of the image, and the 1x1 mipmap will just be the top-left pixel (i.e., red)

My contention is that this is probably an acceptable compromise, because most 3D game art doesn't have strongly opposing colors across it, and BC/DXT does a bad job of dealing with data like that (at low resolution), even in the best case. If complaints are received, we can reconsider and decide it's worth decompressing the image, in order to resample it, but for now it will be a big improvement just to be able to import these images at all.

PR #86361, opened 2023-12-20
Optimize Viewport's `_gui_find_control_at_pos` for `Container` nodes

rsubtil:
*(I couldn't find a related issue, but if desired, I can open one describing the problem in our project)*

When developing an application for managing the user's gaming library, we've had a few users with thousands of games reporting extremely low performance whenever they interacted with any mouse input, such as picking and scrolling the library. This is easily reproducible by adding 10~20k `Control` nodes to some container, such as `FlowContainer`. When moving the mouse and/or scrolling around, the drop in FPS is extremely noticeable (while in this simple scenario the FPS are still good, on our application performance drops to ~20FPS and has very noticeable stutters) :

https://github.com/godotengine/godot/assets/6501975/24ee74c9-7698-4942-9ce2-76fe8e82c0ba

This slowdown was not visible in any of Godot's profilers, suggesting the issue lied somewhere in engine code. After profilling with external tools, the root cause was on how the viewport tests for valid controls under a given mouse position (`Viewport::_gui_find_control_at_pos`):

https://github.com/godotengine/godot/blob/3ce73e5d419011d1642ed30a3417362b82c41898/scene/main/viewport.cpp#L1695-L1707

This function iterates over every node's children, thus being **O(N)**, which in our scenario is extremely wasteful due to our usage of `Container`s with known positions for children nodes, and since this runs on every frame of input, it quickly becomes a noticeable bottleneck.

By taking advantage of the `Container`'s behavior of automatically repositioning nodes, it was possible to greatly optimize this node lookup process by taking into account the mouse's position, lowering the candidate nodes considerably.

# Optimization

This introduces the `Container::get_children_at_pos` function for all `Container` nodes, which should return the candidate list of nodes likely to be at the given position. The default implementation returns all children, which should keep the current behavior for any containers not overriding this function. When the viewport traverses children, it will call this new function in order to optimize the search, and fallback to the existing behavior if no valid children are found. Thus, in theory, this change should not break or change any existing behavior.

This PR overrides this function for `FlowContainer` (`HFlowContainer`/`VFlowContainer`), `BoxContainer` (`HBoxContainer`/`VBoxContainer`) and `GridContainer`. These seemed the best candidates for this optimization, but if there's more containers that could benefit from this, I can add them as well.

> [!NOTE]
> In the following diagrams, grey numbers/dotted lines represent invisible nodes.

## `FlowContainer`

Since this container can fit an arbitrary number of nodes for each row, information about each row's position and first child index is stored. When searching for children, we find the row that contains the given position through a binary search, and return all the children in that row.

image

## `GridContainer`

Same behavior as `FlowContainer`.

image

<image>

## `BoxContainer`

Being a one-dimensional container, only visible children's position and index are stored. When searching for children, we search for it's position with a binary search. This container thus only returns one child.

image

# Benchmark

[ReproUIPerformance.zip](https://github.com/godotengine/godot/files/13728287/ReproUIPerformance.zip)

This test project allows to test the performance of the viewport's node lookup. It should be opened in the editor, and before running, make one of the existing container's visible in order to let it spawn the children. `Invis` variants spawn some children as invisible. Each child is a button indicating it's index, and global (x, y) position.

image

To test this yourself, enable only one of the container nodes, run the project, click on the `Benchmark` button, and don't move the mouse for the next 500 frames. The test will warp the mouse pointer to the center of the screen, and simulate both mouse motion and wheel scrolling. When done, it shows the average frame time for the test.

> [!WARNING]
> Each enabled container spawns a lot of children nodes (50k), and take significant amounts of RAM to launch (~2GB).

https://github.com/godotengine/godot/assets/6501975/88033a4c-3367-4373-9642-ab33ad1df8ea

Here are the results I obtained from my setup (Ryzen 5 5600G, RX 6500 XT, 8GB RAM):

> [!NOTE]
> Both vanilla and optimized builds were compiled with:
> `$ scons use_llvm=yes linker=mold scu_build=yes use_static_cpp=no optimize=speed`

> [!IMPORTANT]
> Values are updated when requested changes modify speedups significantly. Previous data and comparisons remain available below.

| **Node Type** | **Frame time - Vanilla (ms)** | **Frame time - Optimized (ms)** | **Speedup (relative)** |
|---------------|-------------------------------|---------------------------------|------------------------|
| `HFlowContainer` | 44.296 | 16.540 | x2.678 |
| `HFlowContainerInvis` | 65.306 | 26.398 | x2.474 |
| `VFlowContainer` | 44.084 | 16.160 | x2.728 |
| `VFlowContainerInvis` | 63.886 | 26.120 | x2.446 |
| `HBoxContainer` | 44.974 | 10.532 | x4.270 |
| `HBoxContainerInvis` | 68.046 | 19.448 | x3.499 |
| `VBoxContainer` | 44.542 | 14.746 | x3.021 |
| `VBoxContainerInvis` | 66.014 | 24.556 | x2.688 |
| `GridContainer` | 43.754 | 16.608 | x2.635 |
| `GridContainerInvis` | 65.038 | 26.692 | x2.437 |

<details><summary>Old values</summary>

### Original results
| **Node Type** | **Frame time - Vanilla (ms)** | **Frame time - Optimized (ms)** | **Speedup (relative)** | **Improvement (%)** |
|---------------|-------------------------------|---------------------------------|------------------------|---------------------|
| `HFlowContainer` | 40.690 | 16.256 | x2.503 | **+ 6.99%** |
| `HFlowContainerInvis` | 61.696 | 27.214 | x2.267 | **+ 9.13%** |
| `VFlowContainer` | 40.470 | 16.142 | x2.507 | **+ 8.82%** |
| `VFlowContainerInvis` | 61.688 | 26.044 | x2.369 | **+ 3.25%** |
| `HBoxContainer` | 42.042 | 11.038 | x3.809 | **+ 12.10%** |
| `HBoxContainerInvis` | 65.264 | 18.890 | x3.455 | **+ 1.27%** |
| `VBoxContainer` | 42.838 | 14.172 | x3.023 | *- 0.07%* |
| `VBoxContainerInvis` | 65.764 | 26.184 | x2.512 | **+ 7.01%** |
| `GridContainer` | 41.256 | 16.180 | x2.550 | **+ 3.33%** |
| `GridContainerInvis` | 62.544 | 26.808 | x2.333 | **+ 4.46%** |

> Average improvement: **+5.629%**
</details>

## Remarks

- This optimization does not improve the scenario of switching node focus with keyboard/controllers for nodes without specific focus neighbors, which go through a different function (`Control::_window_find_focus_neighbor`). Something similar is needed to optimize that scenario as well.
- There is potential to optimize further by not falling back to the slow **O(N)** search if no valid children are found on `Container`s, which makes it even faster (for the slowest time at `VBoxContainerInvis`, time becomes **17.812** and speedup **x3.692**). This, however, assumes that `Container`'s children only have content on their "bounding boxes", which is not a guarantee as nodes could be moved beyond this, and thus becoming uninteractable. While the speedup with the fallback is already considerable nevertheless, it would be interesting to find a way to support these scenarios in order to improve even further. 

https://github.com/godotengine/godot/assets/6501975/b14e701f-ee23-4cec-a546-8b07fb3cee13

### DIFF
--- a/scene/gui/box_container.cpp
+++ b/scene/gui/box_container.cpp
@@ -200,6 +200,8 @@ void BoxContainer::_resort() {
 		delta = -1;
 	}
 
+	cached_children_nodes.clear();
+
 	for (int i = start; i != end; i += delta) {
 		Control *c = Object::cast_to<Control>(get_child(i));
 		if (!c || !c->is_visible_in_tree()) {
@@ -238,6 +240,7 @@ void BoxContainer::_resort() {
 		}
 
 		fit_child_in_rect(c, rect);
+		cached_children_nodes.push_back(c);
 
 		ofs = to;
 		idx++;
@@ -287,6 +290,35 @@ Size2 BoxContainer::get_minimum_size() const {
 	}
 
 	return minimum;
+}
+
+Vector<CanvasItem *> BoxContainer::get_children_at_pos(const Point2 &p_pos) const {
+	if (cached_children_nodes.is_empty()) {
+		return Vector<CanvasItem *>();
+	}
+
+	int coord = vertical ? p_pos.y : p_pos.x;
+	// Custom binary search; we are searching for the closest value to coord, not an exact one.
+	int bin_start = 0;
+	int bin_end = cached_children_nodes.size();
+	while (bin_start < bin_end) {
+		int mid = (bin_start + bin_end) / 2;
+		if (coord < (vertical ? cached_children_nodes[mid]->get_rect().position.y : cached_children_nodes[mid]->get_rect().position.x)) {
+			bin_end = mid;
+		} else {
+			bin_start = mid + 1;
+		}
+	}
+
+	Vector<CanvasItem *> children;
+	int idx = MAX(bin_start - 1, 0);
+	ERR_FAIL_INDEX_V(idx, cached_children_nodes.size(), children);
+
+	CanvasItem *child = Object::cast_to<CanvasItem>(cached_children_nodes[idx]);
+	if (child) {
+		children.push_back(child);
+	}
+	return children;
 }
 
 void BoxContainer::_notification(int p_what) {
--- a/scene/gui/flow_container.cpp
+++ b/scene/gui/flow_container.cpp
@@ -59,6 +59,8 @@ void FlowContainer::_resort() {
 	int current_container_size = vertical ? get_rect().size.y : get_rect().size.x;
 	int children_in_current_line = 0;
 
+	cached_children_nodes.clear();
+
 	// First pass for line wrapping and minimum size calculation.
 	for (int i = 0; i < get_child_count(); i++) {
 		Control *child = Object::cast_to<Control>(get_child(i));
@@ -140,6 +142,11 @@ void FlowContainer::_resort() {
 		}
 		Size2i child_size = children_minsize_cache[child];
 
+		if (cached_children_nodes.is_empty()) {
+			// Add the first child.
+			cached_children_nodes.push_back(child);
+		}
+
 		_LineData line_data = lines_data[current_line_idx];
 		if (child_idx_in_line >= lines_data[current_line_idx].child_count) {
 			current_line_idx++;
@@ -151,6 +158,7 @@ void FlowContainer::_resort() {
 				ofs.x = 0;
 				ofs.y += line_data.min_line_height + theme_cache.v_separation;
 			}
+			cached_children_nodes.push_back(child);
 			line_data = lines_data[current_line_idx];
 		}
 
@@ -245,6 +253,47 @@ Size2 FlowContainer::get_minimum_size() const {
 	}
 
 	return minimum;
+}
+
+Vector<CanvasItem *> FlowContainer::get_children_at_pos(const Point2 &p_pos) const {
+	if (cached_children_nodes.is_empty()) {
+		return Vector<CanvasItem *>();
+	}
+
+	int coord = vertical ? p_pos.x : p_pos.y;
+	// Custom binary search; we are searching for the closest value to coord, not an exact one.
+	int bin_start = 0;
+	int bin_end = cached_children_nodes.size();
+	while (bin_start < bin_end) {
+		int mid = (bin_start + bin_end) / 2;
+		if (coord < (vertical ? cached_children_nodes[mid]->get_rect().position.x : cached_children_nodes[mid]->get_rect().position.y)) {
+			bin_end = mid;
+		} else {
+			bin_start = mid + 1;
+		}
+	}
+
+	int start_idx = MAX(bin_start - 1, 0);
+	int end_idx = MIN(bin_end, cached_children_nodes.size() - 1);
+	ERR_FAIL_INDEX_V(start_idx, cached_children_nodes.size(), Vector<CanvasItem *>());
+	ERR_FAIL_INDEX_V(end_idx, cached_children_nodes.size(), Vector<CanvasItem *>());
+
+	int start_node_idx = cached_children_nodes[start_idx]->get_index();
+	int end_node_idx = cached_children_nodes[end_idx]->get_index();
+
+	if (start_node_idx == end_node_idx) {
+		// Last row, so include the remaining children.
+		end_node_idx = get_child_count();
+	}
+
+	Vector<CanvasItem *> children;
+	for (end_node_idx--; end_node_idx >= start_node_idx; end_node_idx--) {
+		CanvasItem *child = Object::cast_to<CanvasItem>(get_child(end_node_idx));
+		if (child && child->is_visible()) {
+			children.push_back(child);
+		}
+	}
+	return children;
 }
 
 Vector<int> FlowContainer::get_allowed_size_flags_horizontal() const {
--- a/scene/gui/flow_container.h
+++ b/scene/gui/flow_container.h
@@ -57,6 +57,8 @@ private:
 
 	void _resort();
 
+	Vector<Control *> cached_children_nodes;
+
 protected:
 	bool is_fixed = false;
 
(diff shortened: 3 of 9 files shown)


scene/main/viewport.cpp
```diff
@@ -1691,8 +1691,26 @@ Control *Viewport::_gui_find_control_at_pos(CanvasItem *p_node, const Point2 &p_
 	}
 
 	Control *c = Object::cast_to<Control>(p_node);
+	Point2 point = matrix.affine_inverse().xform(p_global);
+
+	if (!c || !c->is_clipping_contents() || c->has_point(point)) {
+		// If node is a Container, it can optimize the search at the given position.
+		Container *cnt = Object::cast_to<Container>(p_node);
```
AThousandShips, line 1698:
```suggestion
		Container *cont = Object::cast_to<Container>(p_node);
```
Let's not call it that lol 🙃

rsubtil: Yeah, good catch :sweat_smile:

scene/gui/flow_container.h
```diff
@@ -57,6 +57,10 @@ class FlowContainer : public Container {
 
 	void _resort();
 
+	Vector<int> cached_children_pos;
+	Vector<int> cached_children_idx;
+	Vector<Node *> cached_visible_children;
```
Maran23, line 62: this is not used, is it?

rsubtil: No, its a remnant from early experiments.

scene/gui/box_container.cpp
```diff
@@ -238,6 +241,8 @@ void BoxContainer::_resort() {
 		}
 
 		fit_child_in_rect(c, rect);
+		cached_children_pos.push_back(vertical ? rect.position.y : rect.position.x);
+		cached_children_idx.push_back(i);
```
MewPurPur, line 245: When a non-control is added to a container, the indices of controls may change, but no resorting is needed, so `_resort()` won't run. So the logic will break because the indices are now wrong, likewise with other containers. I think you can remedy this by putting the controls directly inside the array.

No idea how this idea would affect performance.

MewPurPur (changes requested):

scene/gui/box_container.cpp
rsubtil: That's indeed a problem:

https://github.com/godotengine/godot/assets/6501975/6ecd969d-2986-4b4d-9398-ff5780ad4121

> I think you can remedy this by putting the controls directly inside the array and modifying the rest of the logic slightly.

I think caching the nodes directly will only work for `BoxContainer`s, since they only have to return one child. For the other containers, I need to return a list of nodes, and through a node reference I don't see a good way to do this. I could use `get_children()` to find where that node exists and then iterate over it, but beyond incurring that search cost, this method performs a copy of the node data, which looks extra wasteful: https://github.com/godotengine/godot/blob/3a8524dd923e9d9a79e4979a47fd427388ea7010/scene/main/node.cpp#L1548-L1558

MewPurPur: In GridContainer, you can get the row and column from division/modulo, it should be similar to BoxContainer but done twice. For FlowContainer, you could store the number of controls in each row/column, which won't change in this situation, unlike indices.

rsubtil: Those are great ideas, I'll try them in the morning!

rsubtil: Alright, so after thinking on how to implement these, I noticed that the issue would remain: while those would fix the problem of how to obtain the next nodes after the starting one, I need a way to know where the starting node lies in the children array; therefore, its index. Turns out, though, that it is possible to do just that with `Node::get_index()`, which I had no idea it existed. 

With that in mind, it's trivial to keep the current system, since now I'll have an array of cached nodes where I can obtain their updated indices.

MewPurPur: Awesome, so you'll only need one array then and it's a code simplification if anything. I'm very glad I caught this now, instead of requiring an extensive investigation of why buttons don't work in a fringe case, years down the line.

rsubtil: Pushed new changes, now this issue should no longer occur. Changes in speedups were also negligible (averaging to 0.18% faster)

MewPurPur: Do you really need to cache the positions if you have the nodes? Just wondering

rsubtil: That's a good point, now that there's direct node access there isn't much point in keeping a separate position array. I'll try this in a bit :+1:

rsubtil: Pushed now. I still haven't tested changes in speedups on all scenarios, but from quick testing, it doesn't seem to be significantly impactful. Will have to update the whole table since my baseline seems to be different now though.

scene/gui/flow_container.cpp
```diff
@@ -140,6 +142,11 @@ void FlowContainer::_resort() {
 		}
 		Size2i child_size = children_minsize_cache[child];
 
+		if (cached_children_nodes.is_empty()) {
+			// Add the first child
```
MewPurPur, line 146: Punctuation.

scene/gui/grid_container.cpp
```diff
@@ -185,11 +185,18 @@ void GridContainer::_notification(int p_what) {
...
 			valid_controls_index = 0;
+			cached_children_nodes.clear();
+
 			for (int i = 0; i < get_child_count(); i++) {
 				Control *c = Object::cast_to<Control>(get_child(i));
 				if (!c || !c->is_visible_in_tree()) {
 					continue;
 				}
+				if (cached_children_nodes.is_empty()) {
+					// Add the first child
```
MewPurPur, line 196: Ditto

scene/gui/flow_container.cpp
rsubtil: Fixed, thanks!

MewPurPur (approved):
Looks good from a code review standpoint.

scene/gui/box_container.cpp
rsubtil: Updated benchmark data, and it seems it did improve a bit, averaging +5.629%.

virophagesp:
any updates?

rsubtil:
@virophagesp not any updates here, but this should still work properly. I deployed this in a custom Godot version for the app I'm developing, and so far I haven't seen nor received any bug reports around it.

 I do want to explore again the "unsafe" optimization I mentioned though, because the change in behavior I talked about seems to occur with this optimization as well, and if it is a reasonable assumption to make, the speedup should increase quite a bit more.

PR #86364, opened 2023-12-20
Add and expose Basis/Transform2D/3D division by float operator

aaronfranke:
I ran into a situation where I needed to use `basis * (1.0 / number)` to inversely scale a basis uniformly, but it would've been nicer to have a division operator. This PR adds that operator. The code is almost identical to the `*` operator.


AThousandShips (approved):
LGTM, and good parity, as we don't do any testing for the operators of this type there's no need to add them right now

AThousandShips:
We should probably consider doing the same for `Transform2D/3D` as well, for the same reasoning

aaronfranke:
Updated to add for Transform2D and Transform3D as well.

akien-mga (approved):
Makes sense to me.

akien-mga:
Thanks!

PR #86373, opened 2023-12-21
C#: Use string literal instead of nameof

raulsntos:
Using `nameof` expressions with method parameters requires C# 11 and we're currently on C# 10.


YuriSizov (approved):

YuriSizov:
Thanks a lot!

PR #80916, opened 2023-08-23
Fix ColorPicker deferred mode not working for sliders.

ajreckof:
Fixes #80915

PR #86377, opened 2023-12-21
Emit slider's `drag_started` signal before the first value change

### LINKED ISSUE #86370: Slider drag_ended signal triggered with click interaction passes false as changed_value
### Tested versions

Reproducible in -4.2.1 stable Windows x64 forward+

### System information

Windows 10 x64 - Godot 4.2.1 stable forward+

### Issue description

I'm using a HSlider. I am using the signal drag_ended. when I click a point on the HSlider the value_changed is false even when the value is updated. this might effect sliders in general. when dragging a point to another point the value_changed is true. here is a video of my code and and a example of this happening. 

https://github.com/godotengine/godot/assets/154477152/e07d61f5-9618-44d9-abd7-5cc12ce844b2

### Steps to reproduce

make a HSlider and make a script and 2 functions inside the script. 1 called on_drag_started(): and 1 called on_drag_end():
in your function _ready(): make 2 lines. the first line will have slider.connect("drag_started",on_drag_started) and the second line will have slider.connect("drag_ended",on_drag_end) then in your on_drag_started(): function put print(slider.value) then in your on_drag_end(): function inside the () of the function add value_changed:bool and then inside the function add a line and put print(slider.value, value_changed) then run the game and test it out. 

https://github.com/KentaPanic/DragBugGodot

timothyqiu:
Fixes #86370

#80916 made `value_changed` to emit **after** `drag_started` in order to fix ColorPicker's deferred mode. But the actual value change is still **before** `drag_started`.

It was then impossible to safely ignore `value_changed` signal when dragging, because you'll miss the `value_changed` signal when pressing down the mouse button.

This PR:

1. Moves `drag_started` before the initial value change, so users can get the correct start value when handling the signal.
2. Records the value before the initial value change, so that `drag_ended` can correctly tell that the value is changed if the user simply clicks somewhere on the slider other than the grabber.

### DIFF
--- a/doc/classes/Slider.xml
+++ b/doc/classes/Slider.xml
@@ -33,7 +33,7 @@
 		</signal>
 		<signal name="drag_started">
 			<description>
-				Emitted when dragging is started.
+				Emitted when dragging is started. This is emitted before the corresponding [signal Range.value_changed] signal.
 			</description>
 		</signal>
 	</signals>
--- a/scene/gui/slider.cpp
+++ b/scene/gui/slider.cpp
@@ -64,6 +64,8 @@ void Slider::gui_input(const Ref<InputEvent> &p_event) {
 				}
 
 				grab.pos = orientation == VERTICAL ? mb->get_position().y : mb->get_position().x;
+				grab.value_before_dragging = get_as_ratio();
+				emit_signal(SNAME("drag_started"));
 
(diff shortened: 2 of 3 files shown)


KoBeWi (approved):

YuriSizov:
Thanks!

ttloves2 (commented):
Reviewed

PR #72418, opened 2023-01-30
Make MenuButton not flat by default

TheSecondReal0:
This fixes behavior that is confusing to end users (evidenced by existence of #30768 and related issues) where MenuButtons don't work/look like other buttons by default.

~~I've only removed the default setting without changing the editor theme/appearance at all, so this PR does change how MenuButtons look in the editor.
With this PR, when hovered they go from this:~~
image
~~To this:~~
image
~~In master they go from this:~~
image
~~To this (the "View" text brightened up a little):~~
image

~~At a glance this doesn't look easy to fix just by changing the editor theme, maybe setting `flat=false` on the editor menu buttons in code is a better way to fix it?~~

Added code to manually give every MenuButton the correct `flat` value, this PR no longer changes the look of the editor in any way.

PR #86378, opened 2023-12-21
Add hover highlight to main editor buttons

RobProductions:
Implements https://github.com/godotengine/godot-proposals/issues/8450

This editor styling change aims to improve usability and accessibility for the most common menu buttons in editor_node.cpp by adding a slight background box when hovered:

https://github.com/godotengine/godot/assets/11141862/d96ded5b-3ff2-4fea-ae92-e1024f5cc5aa

~There are still a few places around the editor that I think could get the same treatment but~ this should tackle the most important buttons (namely the plugin buttons which previously had very light color changes on hover) and I can address those in a future PR if this styling looks good.

The only adjustment which might cause some confusion is that the `FlatMenuButton` theme presets now use a hover highlight background by default when previously the editor flat buttons did not have this. However, the alternative would be manually adjusting all of the icon buttons individually, so I felt this was preferable to create a smaller PR. Let me know if there's anything else I should account for or explain, thanks!

Update: this should now handle most of the buttons I can think of that share a similar style to the ones in the video. There might be exceptions here but my expectation is that they can be ironed out in future PRs; this can hopefully set a precedent for styling going forward.

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -511,8 +511,9 @@ void EditorNode::_update_theme(bool p_skip_creation) {
 
 		scene_root_parent->add_theme_style_override("panel", theme->get_stylebox(SNAME("Content"), EditorStringName(EditorStyles)));
 		bottom_panel->add_theme_style_override("panel", theme->get_stylebox(SNAME("BottomPanel"), EditorStringName(EditorStyles)));
-		main_menu->add_theme_style_override("hover", theme->get_stylebox(SNAME("MenuHover"), EditorStringName(EditorStyles)));
+		main_menu->add_theme_style_override("pressed", theme->get_stylebox(SNAME("MenuTransparent"), EditorStringName(EditorStyles)));
 		distraction_free->set_icon(theme->get_icon(SNAME("DistractionFree"), EditorStringName(EditorIcons)));
+		distraction_free->add_theme_style_override("pressed", theme->get_stylebox(SNAME("MenuTransparent"), EditorStringName(EditorStyles)));
 		bottom_panel_raise->set_icon(theme->get_icon(SNAME("ExpandBottomDock"), EditorStringName(EditorIcons)));
 
 		help_menu->set_item_icon(help_menu->get_item_index(HELP_SEARCH), theme->get_icon(SNAME("HelpSearch"), EditorStringName(EditorIcons)));
@@ -528,6 +529,11 @@ void EditorNode::_update_theme(bool p_skip_creation) {
 
 		if (EditorDebuggerNode::get_singleton()->is_visible()) {
 			bottom_panel->add_theme_style_override("panel", theme->get_stylebox(SNAME("BottomPanelDebuggerOverride"), EditorStringName(EditorStyles)));
+		}
+
+		for (int i = 0; i < bottom_panel_items.size(); i++) {
+			bottom_panel_items.write[i].button->add_theme_style_override("pressed", theme->get_stylebox(SNAME("MenuTransparent"), EditorStringName(EditorStyles)));
+			bottom_panel_items.write[i].button->add_theme_style_override("hover_pressed", theme->get_stylebox(SNAME("MenuHover"), EditorStringName(EditorStyles)));
 		}
 
 		for (int i = 0; i < main_editor_buttons.size(); i++) {
@@ -3293,7 +3299,6 @@ void EditorNode::select_editor_by_name(const String &p_name) {
 void EditorNode::add_editor_plugin(EditorPlugin *p_editor, bool p_config_changed) {
 	if (p_editor->has_main_screen()) {
 		Button *tb = memnew(Button);
-		tb->set_flat(true);
 		tb->set_toggle_mode(true);
 		tb->set_theme_type_variation("MainScreenButton");
 		tb->set_name(p_editor->get_name());
@@ -5188,7 +5193,7 @@ void EditorNode::_scene_tab_closed(int p_tab) {
 
 Button *EditorNode::add_bottom_panel_item(String p_text, Control *p_item, bool p_at_front) {
 	Button *tb = memnew(Button);
-	tb->set_flat(true);
+	tb->set_theme_type_variation("FlatMenuButton");
 	tb->connect("toggled", callable_mp(this, &EditorNode::_bottom_panel_switch_by_control).bind(p_item));
 	tb->set_drag_forwarding(Callable(), callable_mp(this, &EditorNode::_bottom_panel_drag_hover).bind(tb, p_item), Callable());
 	tb->set_text(p_text);
@@ -6661,7 +6666,7 @@ EditorNode::EditorNode() {
 	scene_tabs->connect("tab_closed", callable_mp(this, &EditorNode::_scene_tab_closed));
 
 	distraction_free = memnew(Button);
-	distraction_free->set_flat(true);
+	distraction_free->set_theme_type_variation("FlatMenuButton");
 	ED_SHORTCUT_AND_COMMAND("editor/distraction_free_mode", TTR("Distraction Free Mode"), KeyModifierMask::CTRL | KeyModifierMask::SHIFT | Key::F11);
 	ED_SHORTCUT_OVERRIDE("editor/distraction_free_mode", "macos", KeyModifierMask::META | KeyModifierMask::CTRL | Key::D);
 	distraction_free->set_shortcut(ED_GET_SHORTCUT("editor/distraction_free_mode"));
@@ -6701,9 +6706,7 @@ EditorNode::EditorNode() {
 
 	main_menu = memnew(MenuBar);
 	title_bar->add_child(main_menu);
-
-	main_menu->add_theme_style_override("hover", theme->get_stylebox(SNAME("MenuHover"), EditorStringName(EditorStyles)));
-	main_menu->set_flat(true);
+	main_menu->set_theme_type_variation("FlatMenuButton");
 	main_menu->set_start_index(0); // Main menu, add to the start of global menu.
 	main_menu->set_prefer_global_menu(global_menu);
 	main_menu->set_switch_on_hover(true);
@@ -7169,7 +7172,8 @@ EditorNode::EditorNode() {
 	bottom_panel_raise = memnew(Button);
 	bottom_panel_hb->add_child(bottom_panel_raise);
 	bottom_panel_raise->hide();
-	bottom_panel_raise->set_flat(true);
+	bottom_panel_raise->set_flat(false);
+	bottom_panel_raise->set_theme_type_variation("FlatMenuButton");
 	bottom_panel_raise->set_toggle_mode(true);
 	bottom_panel_raise->set_shortcut(ED_SHORTCUT_AND_COMMAND("editor/bottom_panel_expand", TTR("Expand Bottom Panel"), KeyModifierMask::SHIFT | Key::F12));
 	bottom_panel_raise->connect("toggled", callable_mp(this, &EditorNode::_bottom_panel_raise_toggled));
--- a/editor/gui/editor_run_bar.cpp
+++ b/editor/gui/editor_run_bar.cpp
@@ -372,7 +372,7 @@ EditorRunBar::EditorRunBar() {
 
 	play_button = memnew(Button);
 	main_hbox->add_child(play_button);
-	play_button->set_flat(true);
+	play_button->set_theme_type_variation("RunBarButton");
 	play_button->set_toggle_mode(true);
 	play_button->set_focus_mode(Control::FOCUS_NONE);
 	play_button->set_tooltip_text(TTR("Run the project's default scene."));
@@ -384,7 +384,7 @@ EditorRunBar::EditorRunBar() {
 
 	pause_button = memnew(Button);
 	main_hbox->add_child(pause_button);
-	pause_button->set_flat(true);
+	pause_button->set_theme_type_variation("RunBarButton");
 	pause_button->set_toggle_mode(true);
 	pause_button->set_focus_mode(Control::FOCUS_NONE);
 	pause_button->set_tooltip_text(TTR("Pause the running project's execution for debugging."));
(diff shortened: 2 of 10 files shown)


Mickeon:
Huh. Never even noticed they didn't exist. I see this as a net positive. How does this look in other themes?

RobProductions:
> How does this look in other themes?

Okay, this had me scratching my head for hours on end and I went down a rabbit hole to debug this, but here's what happened. I did notice an issue on one of the light themes which I thought I had resolved, but when I checked again after I read your comment there were still some *oddities*. Probably better to just address it now I thought, but no matter what I tried I couldn't get theme color overrides to work across both light and dark themes for the main "plugin" buttons (to select 2D, 3D, etc.).

That's when I discovered that `add_editor_plugin()` in `editor_node.cpp` was only ever getting run when Godot initialized, not when the user selects a new theme. Any theme style/color overrides happening there will only be updated when that function is run again, so the user will have to restart the editor for those changes to take effect. I managed to reduce the impact of that down to just one minor inconsistency: the background color for the `hover_pressed` stylebox. To reiterate, that color will not be updated until the user restarts the editor, though I believe this is a barely noticeable artifact and outside the scope of this PR to fix. Should also note that **I believe this issue exists throughout the editor code** (namely whenever a theme override is set within a constructor) and I only discovered it playing around with the colors myself, so I think a future PR should address that instead. I can make a Github issue for this once I track down the other problem cases. The reason this wasn't noticed before was because the plugin buttons didn't use theme overrides before, and now they do in order to make the hover effect consistent.

Anyways, with that out of the way I corrected/simplified the code using the above knowledge and it should now work as expected on all of the available themes: 

https://github.com/godotengine/godot/assets/11141862/094408bf-660c-46e3-84d0-5a50ca616498

And here's another example:

https://github.com/godotengine/godot/assets/11141862/db78e9a7-e0be-4dd8-99d3-63b191a537dd

Again with the only caveat being the `hover_pressed` background color not updating correctly until an editor restart due to an underlying issue with theme overrides, but if this is too inconvenient I can simply remove that part and we can live without hover highlights on toggled plugin buttons for now :) Let me know if this looks good!

RobProductions:
Alright, apologies for unexpected updates to the branch but I believe I've finally sorted everything out now. The `NOTIFICATION_THEME_CHANGED` (and `update_theme()` in `editor_node.cpp`) was hiding in plain sight the entire time and moving my styling changes there allowed me to fix the issue of colors not getting corrected when the theme switched :) For good measure, I also moved the existing styling for the plugin buttons there to avoid future confusion.

Here's a quick demo showing that the colors get updated even after a theme switch:

https://github.com/godotengine/godot/assets/11141862/b6c1df6d-ee9e-4b1c-8418-a992ac685beb

The other places where I updated button styling don't need to be updated after a theme switch because the values are constant (transparent), so I left them in the constructor to match the other areas of the editor code. I still think that a few places elsewhere in the code don't use theme_changed as expected and thus have their colors messed up so I will look into making a bug report to point those out.

As for these changes though, everything should work as expected now as far as I can tell with no issues or caveats! Feel free to let me know if I missed anything else or need to address anything further, and thank you again for checking it out!

RobProductions:
Update: Upon reflection I was thinking it might be better to have a more comprehensive styling change in one PR as opposed to leaving stuff on the to-do list for https://github.com/godotengine/godot-proposals/issues/8450 , so I altered more buttons to add the hover highlight I showed above. While there's more code changed now, this should result in a more consistent UI from a single commit and we can definitively close the proposal if it's merged since most common buttons should be covered now. I've heard that this is the best way to handle editor UI updates too so that we don't end up with a half-implementation between commits. Here's a quick video to show the new buttons that received the styling update:

https://github.com/godotengine/godot/assets/11141862/099d78cc-77ae-4697-9ffc-721b6ce2aef8

And I rebased the branch for good measure :) as before, let me know if there's anything else I should take a look at, thanks!

RobProductions:
Rebased to conform to the new editor theme file structure

editor/themes/editor_theme_manager.cpp
```diff
@@ -1736,6 +1742,8 @@ void EditorThemeManager::_populate_editor_styles(const Ref<EditorTheme> &p_theme
 				style_flat_button->set_content_margin((Side)i, p_config.button_style->get_margin((Side)i) + p_config.button_style->get_border_width((Side)i));
 			}
 
+			Ref<StyleBoxFlat> style_hover_button = p_config.button_style_hover->duplicate();
```
YuriSizov, line 1745:
```suggestion
			Ref<StyleBoxFlat> style_flat_button_hover = p_config.button_style_hover->duplicate();
```

To stay consistent with existing code.

editor/themes/editor_theme_manager.cpp
```diff
@@ -1679,7 +1680,12 @@ void EditorThemeManager::_populate_editor_styles(const Ref<EditorTheme> &p_theme
 		p_theme->set_stylebox("BottomPanel", EditorStringName(EditorStyles), style_bottom_panel);
 
 		// Main menu.
-		p_theme->set_stylebox("MenuHover", EditorStringName(EditorStyles), p_config.button_style_hover);
+		Ref<StyleBoxFlat> style_widget_transparent = p_config.button_style->duplicate();
+		style_widget_transparent->set_bg_color(Color(1, 1, 1, 0));
+		style_widget_transparent->set_border_width_all(0);
+		Ref<StyleBoxFlat> style_widget_hover = p_config.button_style_hover->duplicate();
```
YuriSizov, line 1686: You should use more specific and descriptive names, "widget" means nothing in this context. `menu_transparent_style`, `menu_hover_style`, or something similar, would be more fitting.

RobProductions:
Done. The naming scheme stemmed from the old editor_theme.cpp variables and I forgot to update it to match the new structure, thanks for catching that!

Calinou (approved):
Tested locally, it works as expected. This is a nice usability improvement :slightly_smiling_face: 

### Default

https://github.com/godotengine/godot/assets/180032/1e4582bb-6d89-4618-8e43-7ba2175df401

### Light

https://github.com/godotengine/godot/assets/180032/e19b7262-2d5a-4e86-9b18-7fcece1cbb1b

### Black (OLED)

https://github.com/godotengine/godot/assets/180032/38b20f96-f647-444a-b751-fced19359b1d

RobProductions:
Thanks for the review! I fixed the merge conflict in `editor_node.cpp` by rebasing and caught another button that was missing the hover styling (onion skinning options in the animation tab)

AThousandShips:
Needs another rebase

RobProductions:
Update: It did not last long. Rebased again to handle the conflict

editor/editor_node.cpp
```diff
@@ -521,10 +522,21 @@ void EditorNode::_update_theme(bool p_skip_creation) {
 		help_menu->set_item_icon(help_menu->get_item_index(HELP_ABOUT), theme->get_icon(SNAME("Godot"), EditorStringName(EditorIcons)));
 		help_menu->set_item_icon(help_menu->get_item_index(HELP_SUPPORT_GODOT_DEVELOPMENT), theme->get_icon(SNAME("Heart"), EditorStringName(EditorIcons)));
 
+		for (int i = 0; i < main_editor_buttons.size(); i++) {
```
KoBeWi, line 525: There is a MainScreenButton variation added recently, which is used by main editor buttons. You should modify it instead of adding overrides.
(note that it's also used in Project Manager)

editor/editor_node.cpp
```diff
@@ -6675,8 +6689,9 @@ EditorNode::EditorNode() {
 	main_menu = memnew(MenuBar);
 	title_bar->add_child(main_menu);
 
+	main_menu->add_theme_style_override("normal", theme->get_stylebox(SNAME("MenuTransparent"), EditorStringName(EditorStyles)));
+	main_menu->add_theme_style_override("pressed", theme->get_stylebox(SNAME("MenuTransparent"), EditorStringName(EditorStyles)));
```
KoBeWi, line 6693: Theme properties shouldn't be accessed inside the constructor.

editor/gui/editor_run_bar.cpp
```diff
@@ -52,8 +52,20 @@ void EditorRunBar::_notification(int p_what) {
...
+			pause_button->add_theme_style_override("pressed", get_theme_stylebox(SNAME("normal"), "FlatMenuButton"));
 			stop_button->set_icon(get_editor_theme_icon(SNAME("Stop")));
+			stop_button->add_theme_style_override("disabled", get_theme_stylebox(SNAME("normal"), "FlatMenuButton"));
+			stop_button->add_theme_style_override("pressed", get_theme_stylebox(SNAME("normal"), "FlatMenuButton"));
+			play_scene_button->add_theme_style_override("disabled", get_theme_stylebox(SNAME("normal"), "FlatMenuButton"));
+			play_scene_button->add_theme_style_override("pressed", get_theme_stylebox(SNAME("normal"), "FlatMenuButton"));
+			play_custom_scene_button->add_theme_style_override("disabled", get_theme_stylebox(SNAME("normal"), "FlatMenuButton"));
+			play_custom_scene_button->add_theme_style_override("pressed", get_theme_stylebox(SNAME("normal"), "FlatMenuButton"));
+			write_movie_button->add_theme_style_override("disabled", get_theme_stylebox(SNAME("normal"), "FlatMenuButton"));
+			write_movie_button->add_theme_style_override("pressed", get_theme_stylebox(SNAME("normal"), "FlatMenuButton"));
```
KoBeWi, line 68: That's too many overrides...
It would be better to use some theme type variation (a new one if necessary).

editor/scene_tree_dock.cpp
```diff
@@ -4140,14 +4140,16 @@ SceneTreeDock::SceneTreeDock(Node *p_scene_root, EditorSelection *p_editor_selec
 	ED_SHORTCUT("scene_tree/delete", TTR("Delete"), Key::KEY_DELETE);
 
 	button_add = memnew(Button);
-	button_add->set_flat(true);
+	button_add->set_flat(false);
```
KoBeWi, line 4143:
```suggestion
```
Button is non-flat by default, so no need to set it. (same in other Buttons)

editor/themes/editor_theme_manager.cpp
```diff
@@ -1679,7 +1680,12 @@ void EditorThemeManager::_populate_editor_styles(const Ref<EditorTheme> &p_theme
 		p_theme->set_stylebox("BottomPanel", EditorStringName(EditorStyles), style_bottom_panel);
 
 		// Main menu.
-		p_theme->set_stylebox("MenuHover", EditorStringName(EditorStyles), p_config.button_style_hover);
+		Ref<StyleBoxFlat> menu_transparent_style = p_config.button_style->duplicate();
+		menu_transparent_style->set_bg_color(Color(1, 1, 1, 0));
+		menu_transparent_style->set_border_width_all(0);
+		Ref<StyleBoxFlat> menu_hover_style = p_config.button_style_hover->duplicate();
```
KoBeWi, line 1686: Why are you duplicating this if it's the same?

editor/themes/editor_theme_manager.cpp
```diff
@@ -1736,6 +1742,8 @@ void EditorThemeManager::_populate_editor_styles(const Ref<EditorTheme> &p_theme
 				style_flat_button->set_content_margin((Side)i, p_config.button_style->get_margin((Side)i) + p_config.button_style->get_border_width((Side)i));
 			}
 
+			Ref<StyleBoxFlat> style_flat_button_hover = p_config.button_style_hover->duplicate();
```
KoBeWi, line 1745: Same as above.

editor/themes/editor_theme_manager.cpp
RobProductions: I was told that Godot is more verbose intentionally to reduce confusion so I wasn't really sure how far that went, i.e. if we needed to make sure everything used in this scope could easily be edited later with less line changes. I can remove the duplicate in my next push

editor/gui/editor_run_bar.cpp
RobProductions: Will do, I was trying my best to work with what already existed in hopes that it would be less destructive but I can see how it could be simplified with a new variation. I'll add it in my next push, thanks!

editor/scene_tree_dock.cpp
RobProductions: Ah got it, I'll address this too. For future reference, is it always better to remove redundant lines like this when possible? I was under the impression that in some cases it might be better to explicitly set stuff so the intent is clear, but maybe that's not needed for something this common

RobProductions:
Hi @KoBeWi , I implemented all of your suggestions and rebased again. Thanks for the review! 

I had a bit of confusion because I nearly forgot that Button is not flat by default whereas MenuButton is, so I had to leave the setters for those button types. (Perhaps a sign to revisit https://github.com/godotengine/godot/pull/72418 ?) Also, I noticed that `main_menu` in `editor_node.cpp` already accesses "theme properties" inside the inspector before my commit, so I changed it to use a theme variation instead.

Feel free to take a look at the updated code, appreciate it :)

KoBeWi:
I'm not sure about using transparent style for pressed menu.

https://github.com/godotengine/godot/assets/2223172/8b3e2495-c94d-41a2-a88b-3982a28782dc

Most apps seem to use more accented hover style for pressed, e.g. Notepad:

https://github.com/godotengine/godot/assets/2223172/97558878-0fbb-4718-bb38-e42dca533d57

KoBeWi (approved):
Looks fine code-wise.
Could use a check from @godotengine/usability to make sure the new style is fine and makes sense everywhere it was added (and whether it's missing somewhere).

RobProductions:
> I'm not sure about using transparent style for pressed menu.

I would have used the FlatMenuButton's default "pressed" stylebox (which you can see in action on the smaller icon buttons now) but the color blended too much into the background for those top bar items, and it didn't seem consistent with the other drop down menus to use the highlight color. The transparent BG is identical to how it works now in 4.x so hopefully not too jarring.

Could definitely add in the highlight but would have to make sure all the other drop downs match too... probably a task for a future PR then? Thanks for the review :)

akien-mga:

PR #86379, opened 2023-12-21
Android: Update the logic used to start / stop the render thread

m4gr3d:
Currently the render thread is started / stopped when the activity is respectively resumed / paused. However, according to the `GLSurfaceView` documentation, this should be done instead when the activity is started / stopped, so this change updates the start / stop logic for the render thread to match the documentation.

[3.x version](https://github.com/godotengine/godot/pull/86380)


joined72:
Any news?

akien-mga (approved):

akien-mga:
Thanks!

PR #86380, opened 2023-12-21
[3.x] Android: Update the logic used to start / stop the GL thread

m4gr3d:
Currently the GL thread is started / stopped when the activity is respectively resumed / paused. However, according to the `GLSurfaceView` documentation, this should be done instead when the activity is started / stopped, so this change updates the start / stop logic for the GL thread to match the documentation.

[4.x version](https://github.com/godotengine/godot/pull/86379)


akien-mga (approved):

akien-mga:
Thanks!

PR #86383, opened 2023-12-21
Specify the path to the Java SDK used for the Android gradle build

m4gr3d:
Introduce an editor setting to allow users the ability to specify the path to the Java SDK used for the Android gradle build.
At build time, a gradle task validates the `java` version and prints out an error if it doesn't match the target version

Screenshot 2023-12-21 at 3 25 47 AM

### DIFF
--- a/platform/android/export/export_plugin.cpp
+++ b/platform/android/export/export_plugin.cpp
@@ -2111,8 +2111,17 @@ Ref<Texture2D> EditorExportPlatformAndroid::get_run_icon() const {
 	return run_icon;
 }
 
+String EditorExportPlatformAndroid::get_java_path() {
+	String exe_ext;
+	if (OS::get_singleton()->get_name() == "Windows") {
+		exe_ext = ".exe";
+	}
+	String java_sdk_path = EDITOR_GET("export/android/java_sdk_path");
+	return java_sdk_path.path_join("bin/java" + exe_ext);
+}
+
 String EditorExportPlatformAndroid::get_adb_path() {
-	String exe_ext = "";
+	String exe_ext;
 	if (OS::get_singleton()->get_name() == "Windows") {
 		exe_ext = ".exe";
 	}
@@ -2124,13 +2133,13 @@ String EditorExportPlatformAndroid::get_apksigner_path(int p_target_sdk, bool p_
 	if (p_target_sdk == -1) {
 		p_target_sdk = DEFAULT_TARGET_SDK_VERSION;
 	}
-	String exe_ext = "";
+	String exe_ext;
 	if (OS::get_singleton()->get_name() == "Windows") {
 		exe_ext = ".bat";
 	}
 	String apksigner_command_name = "apksigner" + exe_ext;
 	String sdk_path = EDITOR_GET("export/android/android_sdk_path");
-	String apksigner_path = "";
+	String apksigner_path;
 
 	Error errn;
 	String build_tools_dir = sdk_path.path_join("build-tools");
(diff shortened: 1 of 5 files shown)


platform/android/export/export_plugin.cpp
```diff
@@ -2111,6 +2111,15 @@ Ref<Texture2D> EditorExportPlatformAndroid::get_run_icon() const {
 	return run_icon;
 }
 
+String EditorExportPlatformAndroid::get_java_path() {
+	String exe_ext = "";
```
akien-mga, line 2115: Strings are default initialized to empty.
```suggestion
	String exe_ext;
```

platform/android/export/export_plugin.cpp
```diff
@@ -2329,6 +2338,32 @@ bool EditorExportPlatformAndroid::has_valid_export_configuration(const Ref<Edito
...
+		Error errn;
+		// Check for the bin directory
+		Ref<DirAccess> da = DirAccess::open(java_sdk_path.path_join("bin"), &errn);
+		if (errn != OK) {
+			err += TTR("Invalid Java SDK path in Editor Settings.");
+			err += TTR("Missing 'bin' directory!");
+			err += "\n";
+			valid = false;
+		} else {
+			// Check for the `java` command
```
akien-mga, line 2356:
```suggestion
		// Validate the given path by checking that `java` and `javac` are present under the `bin` directory.
		Error errn;
		// Check for the bin directory.
		Ref<DirAccess> da = DirAccess::open(java_sdk_path.path_join("bin"), &errn);
		if (errn != OK) {
			err += TTR("Invalid Java SDK path in Editor Settings.");
			err += TTR("Missing 'bin' directory!");
			err += "\n";
			valid = false;
		} else {
			// Check for the `java` command.
```

platform/android/export/export_plugin.cpp
```diff
@@ -2911,6 +2953,9 @@ Error EditorExportPlatformAndroid::export_project_helper(const Ref<EditorExportP
 		print_verbose("Storing command line flags...");
 		store_file_at_path(assets_directory + "/_cl_", command_line_flags);
 
+		print_verbose("Updating JAVA_HOME environment to " + java_sdk_path);
+		OS::get_singleton()->set_environment("JAVA_HOME", java_sdk_path); //set and overwrite if required
```
akien-mga, line 2957:
```suggestion
		OS::get_singleton()->set_environment("JAVA_HOME", java_sdk_path); // Set and overwrite if required.
```
And likely the same treatment for the comment it was copied from. But maybe repeating these comments isn't so useful and they could be removed (it's not clear what the `if required` means here).

akien-mga (approved):
Looks great!

YuriSizov:
Thanks!

goatchurchprime:
I think there's still an issue here, because the values stored in the editor settings will continue to over-ride these JAVA_HOME and ANDROID_HOME environment values if they are not reset-to-default.  This is only going to help people when they make a fresh install with no preset editor settings.

The most user-friendly option might be to make these boxes read-only in the UI with a tooltip that says they've been defined by the environment variables.

image

Attached is my nix-flake that I use to make godot with some edits to the source code that force-set all these values with the following substitutions:
[flake.zip](https://github.com/godotengine/godot/files/14232353/flake.zip)

```bash
                substituteInPlace platform/android/export/export_plugin.cpp \
                    --replace 'String sdk_path = EDITOR_GET("export/android/android_sdk_path")' 'String sdk_path = std::getenv("tunnelvr_ANDROID_SDK")'

                substituteInPlace platform/android/export/export_plugin.cpp \
                    --replace 'EDITOR_GET("export/android/debug_keystore")' 'std::getenv("tunnelvr_DEBUG_KEY")'

                substituteInPlace platform/android/export/export_plugin.cpp \
                    --replace 'EDITOR_GET("export/android/java_sdk_path")' 'std::getenv("tunnelvr_JAVA_SDK_PATH")'

                substituteInPlace editor/editor_paths.cpp \
                    --replace 'return get_data_dir().path_join(export_templates_folder)' 'return std::getenv("tunnelvr_EXPORT_TEMPLATES")'
                    
                    <snip>
                    
                        wrapProgram $out/bin/godot4 \
                            --set ANDROID_HOME "${androidenv}/share/android-sdk"\
                            --set JAVA_HOME "${pkgs.jdk17}/lib/openjdk"\
                            --set tunnelvr_ANDROID_SDK "${androidenv}/share/android-sdk"\
                            --set tunnelvr_JAVA_SDK_PATH "${pkgs.jdk17}/lib/openjdk"\
                            --set tunnelvr_EXPORT_TEMPLATES "${export-templates}/templates" \
                            --set tunnelvr_DEBUG_KEY "${debugKey}" \
                            --set GODOT_BLENDER3_PATH "${pkgs.blender}/bin/" \
                            --set GRADLE_OPTS "-Dorg.gradle.project.android.aapt2FromMavenOverride=${androidenv}/share/android-sdk/build-tools/34.0.0/aapt2"

```

m4gr3d:
I'm not sure I understand the issue.. Those values are meant to override the env variables so that users with multiple Java and/or Android sdks can specify which one the Godot editor should use.

Making them read only defeats that purpose.

akien-mga:
Cherry-picked for 4.2.2.

PR #86386, opened 2023-12-21
Strip edges of editor layout names

timothyqiu:
Layout names are saved as `ConfigFile` section names. Leading and trailing spaces in section names are stripped when loaded from disk via `ConfigFile::load()`.

It's currently possible to create a layout named `"    "`. But it's impossible to remove it because it's loaded as `""`. Deleting such item is considered deleting something that does not exist and is skipped.

This PR:

- Makes `EditorLayoutsDialog` always `strip_edges()` for layout names.
    - This only happens for names typed manually. So old layout names that has leading or trailing spaces can still be overwritten/deleted.
- Removes the empty string check when deleting a layout. So existing empty layout name can be deleted.
    - The check was redundant since the dialog already does such check.
- Also removes two unnecessary includes.

### DIFF
--- a/editor/editor_layouts_dialog.cpp
+++ b/editor/editor_layouts_dialog.cpp
@@ -31,8 +31,6 @@
 #include "editor_layouts_dialog.h"
 
 #include "core/io/config_file.h"
-#include "core/object/class_db.h"
-#include "core/os/keyboard.h"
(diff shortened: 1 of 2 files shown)


akien-mga (approved):

akien-mga:
Thanks!

Issue #79750, opened 2023-07-21
"UndoRedo history mismatch: expected 0, got 1" when dragging image from FileSystem tab onto Visual Shader Editor


enxas:
### Godot version

v4.1.1.stable.mono.official [bd6af8e0e]

### System information

Godot v4.1.1.stable.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1660 Ti with Max-Q Design (NVIDIA; 31.0.15.3640) - Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (12 Threads)

### Issue description

Dragging and dropping image from FileSystem tab onto Visual Shader Editor produces error: `UndoRedo history mismatch: expected 0, got 1.`

### Steps to reproduce

1. Create MeshInstance3D.
2. Select that mesh and add Surface material Override -> New ShaderMaterial -> Visual Shader.
3. Drag and drop image from FileSystem tab onto Visual Shader Editor like shown in the video below

https://github.com/godotengine/godot/assets/8630984/c0e93cfe-2d2b-46b1-9aa0-a5f3ed095235

### Minimal reproduction project

NA

KoBeWi:
Can't reproduce in 4.2 dev.

enxas:
I have compiled Godot v4.2.dev.mono.custom_build [6588a4a29] and issue i still there. I have added MRP this time.
1. Double click `test.tres` in FileSystem to open Visual Shader Editor
2. From FileSystem click and drag `Grass002_1K_Color.jpg` onto Visual Shader Editor to create Texture2D node
3. After that error is printed `UndoRedo history mismatch: expected 0, got 3.` is printed

[MRP.zip](https://github.com/godotengine/godot/files/12135877/MRP.zip)

PR #86387, opened 2023-12-21
Fix `UndoRedo history mismatch` when creating a new tile atlas

### LINKED ISSUE #86291: `UndoRedo history mismatch` when creating a new tile atlas
### Tested versions

Godot v4.3.dev (2d0ee20ff)

### System information

Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3060 (NVIDIA; 31.0.15.3619) - 11th Gen Intel(R) Core(TM) i7-11700K @ 3.60GHz (16 Threads)

### Issue description

When debugging https://github.com/godotengine/godot/issues/86226, I found that after dragging the image to create atlas, an internal error will show like the image below:

image

The log looks like this :

```
UndoRedo history mismatch: expected 0, got 1.
Add a new atlas source
```

### Steps to reproduce

Forme, just drag a random png file to create a tile atlas

N/A

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/86291

It seems that there's no need to put `set_texture_region_size` in do method as it's a newly created resource and no pairded redo method. 

But I feel the `history mismatch` problem lies in other places, it's only a stopgap measure. Any suggestions ?

### DIFF
--- a/editor/plugins/tiles/tile_set_editor.cpp
+++ b/editor/plugins/tiles/tile_set_editor.cpp
@@ -117,11 +117,11 @@ void TileSetEditor::_load_texture_files(const Vector<String> &p_paths) {
 		// Actually create the new source.
 		Ref<TileSetAtlasSource> atlas_source = memnew(TileSetAtlasSource);
 		atlas_source->set_texture(texture);
+		atlas_source->set_texture_region_size(tile_set->get_tile_size());
 
 		EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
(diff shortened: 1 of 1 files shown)


KoBeWi (approved):
This is a correct fix.

The issue is basically the same as #79750. It only happens with TileSet saved to file. Undo history starts in global context, but the newly created atlas is recognized as a built-in resource, thus part of the scene, causing mismatch.

YuriSizov:
Thanks!

PR #86399, opened 2023-12-21
Add support for selecting multiple properties with `PropertySelector`

### LINKED ISSUE #8713: Add support for selecting multiple properties with AnimationPlayer/PropertySelect

A top down 2d vampire survivors like game.

I have to add many properties for each spritesheet (vframes,hframes,frames,texture etc.)  When I am making an animation in the animationPlayer.

The feature would allow selecting multiple properties using shift+click or ctrl + click (I didn't write this functionality). The existing functionality of selecting an individual property would still exist and remain the same.

Here is the open PR https://github.com/godotengine/godot/pull/86399 basically it uses a dictionary to store which items have been selected and allows selecting multiple using shift+click ,or ctrl + click (I didn't write this functionality).

image

It would be used all the time, at the very least by me. It can currently be worked around but its tedious having to add each property one by one.

Doesn't make sense as an add-on this is core stuff.

kyle-wannacott:
Proposal: https://github.com/godotengine/godot-proposals/issues/8713

PR allows selecting and adding multiple property tracks in animation player with either shift + click (range), or ctrl + click for (single) . I would like this feature as I spend too much time adding hframes, vframes, frames and texture properties individually for each spritesheet.

I don't know if there is an issue for this, or someone was working on it.

image

Keep in mind I don't know C++ :sweat_smile: ; so I'm probably committing some sins . There is a bug when pressing arrow keys that I think I introduced (and maybe other ones);  but the multiple selecting seems to work and adds the tracks to the animation. This is a first draft; I'll add more details and clean it up tomorrow/ investigate; I'm tired  right now though : /

_Production edit: Closes https://github.com/godotengine/godot-proposals/issues/8713_

### DIFF
--- a/editor/animation_track_editor.cpp
+++ b/editor/animation_track_editor.cpp
@@ -4970,11 +4970,13 @@ void AnimationTrackEditor::_new_track_node_selected(NodePath p_path) {
 
 	switch (adding_track_type) {
 		case Animation::TYPE_VALUE: {
+			prop_selector->set_multiselect(true);
 			adding_track_path = path_to;
 			prop_selector->set_type_filter(Vector<Variant::Type>());
 			prop_selector->select_property_from_instance(node);
 		} break;
 		case Animation::TYPE_BLEND_SHAPE: {
+			prop_selector->set_multiselect(true);
 			adding_track_path = path_to;
 			Vector<Variant::Type> filter;
 			filter.push_back(Variant::FLOAT);
@@ -4994,6 +4996,7 @@ void AnimationTrackEditor::_new_track_node_selected(NodePath p_path) {
 
 		} break;
 		case Animation::TYPE_BEZIER: {
+			prop_selector->set_multiselect(true);
 			Vector<Variant::Type> filter;
 			filter.push_back(Variant::INT);
 			filter.push_back(Variant::FLOAT);
--- a/editor/animation_track_editor.cpp
+++ b/editor/animation_track_editor.cpp
@@ -5059,7 +5059,7 @@ void AnimationTrackEditor::_add_track(int p_type) {
 	pick_track->get_filter_line_edit()->grab_focus();
 }
 
-void AnimationTrackEditor::_handle_multiple_properties_selected(Array p_multiple_properties) {
+void AnimationTrackEditor::_add_multiple_tracks(Array p_multiple_properties) {
 	for (int i = 0; i < p_multiple_properties.size(); i++) {
 		_new_track_property_selected(p_multiple_properties[i]);
 	}
@@ -7388,7 +7388,7 @@ AnimationTrackEditor::AnimationTrackEditor() {
 	prop_selector = memnew(PropertySelector);
 	add_child(prop_selector);
 	prop_selector->connect("selected", callable_mp(this, &AnimationTrackEditor::_new_track_property_selected));
-	prop_selector->connect("multiple_properties_selected", callable_mp(this, &AnimationTrackEditor::_handle_multiple_properties_selected));
+	prop_selector->connect("multiple_properties_selected", callable_mp(this, &AnimationTrackEditor::_add_multiple_tracks));
 	prop_selector->set_multiselect(true);
 
 	method_selector = memnew(PropertySelector);
--- a/editor/animation_track_editor.cpp
+++ b/editor/animation_track_editor.cpp
@@ -4970,13 +4970,11 @@ void AnimationTrackEditor::_new_track_node_selected(NodePath p_path) {
 
 	switch (adding_track_type) {
 		case Animation::TYPE_VALUE: {
-			prop_selector->set_multiselect(true);
 			adding_track_path = path_to;
 			prop_selector->set_type_filter(Vector<Variant::Type>());
 			prop_selector->select_property_from_instance(node);
 		} break;
 		case Animation::TYPE_BLEND_SHAPE: {
-			prop_selector->set_multiselect(true);
 			adding_track_path = path_to;
 			Vector<Variant::Type> filter;
 			filter.push_back(Variant::FLOAT);
@@ -4996,7 +4994,6 @@ void AnimationTrackEditor::_new_track_node_selected(NodePath p_path) {
 
 		} break;
 		case Animation::TYPE_BEZIER: {
-			prop_selector->set_multiselect(true);
 			Vector<Variant::Type> filter;
 			filter.push_back(Variant::INT);
 			filter.push_back(Variant::FLOAT);
@@ -5060,6 +5057,12 @@ void AnimationTrackEditor::_add_track(int p_type) {
 	pick_track->popup_scenetree_dialog(nullptr, root_node);
 	pick_track->get_filter_line_edit()->clear();
 	pick_track->get_filter_line_edit()->grab_focus();
+}
+
+void AnimationTrackEditor::_handle_multiple_properties_selected(Array p_multiple_properties) {
+	for (int i = 0; i < p_multiple_properties.size(); i++) {
+		_new_track_property_selected(p_multiple_properties[i]);
+	}
 }
 
 void AnimationTrackEditor::_fetch_value_track_options(const NodePath &p_path, Animation::UpdateMode *r_update_mode, Animation::InterpolationType *r_interpolation_type, bool *r_loop_wrap) {
@@ -7385,6 +7388,8 @@ AnimationTrackEditor::AnimationTrackEditor() {
 	prop_selector = memnew(PropertySelector);
 	add_child(prop_selector);
 	prop_selector->connect("selected", callable_mp(this, &AnimationTrackEditor::_new_track_property_selected));
+	prop_selector->connect("multiple_properties_selected", callable_mp(this, &AnimationTrackEditor::_handle_multiple_properties_selected));
+	prop_selector->set_multiselect(true);
 
 	method_selector = memnew(PropertySelector);
 	add_child(method_selector);
--- a/editor/property_selector.cpp
+++ b/editor/property_selector.cpp
@@ -425,10 +425,12 @@ void PropertySelector::_hide_requested() {
 
 void PropertySelector::_notification(int p_what) {
 	switch (p_what) {
+		case 30: {
+			selected_properties.clear();
+		} break;
 		case NOTIFICATION_ENTER_TREE: {
 			connect(SceneStringName(confirmed), callable_mp(this, &PropertySelector::_confirmed));
 		} break;
-
 		case NOTIFICATION_EXIT_TREE: {
 			disconnect(SceneStringName(confirmed), callable_mp(this, &PropertySelector::_confirmed));
 		} break;
@@ -590,9 +592,9 @@ PropertySelector::PropertySelector() {
 	search_box->connect(SceneStringName(gui_input), callable_mp(this, &PropertySelector::_sbox_input));
 	search_options = memnew(Tree);
 	search_options->set_auto_translate_mode(AUTO_TRANSLATE_MODE_DISABLED);
-	search_options->set_select_mode(search_options->SELECT_MULTI);
+	search_options->set_select_mode(search_options->SELECT_SINGLE);
 	vbc->add_margin_child(TTR("Matches:"), search_options, true);
-	set_ok_button_text(TTR("Open"));
+	set_ok_button_text(TTR("Select"));
 	get_ok_button()->set_disabled(true);
 	register_text_enter(search_box);
 	set_hide_on_ok(false);
--- a/editor/property_selector.cpp
+++ b/editor/property_selector.cpp
@@ -358,15 +358,20 @@ void PropertySelector::_update_search() {
 }
 
 void PropertySelector::_confirmed() {
-	Array keys = selected_properties.keys();
 	TreeItem *ti = search_options->get_selected();
 	if (!ti) {
 		return;
 	}
-	for (int i = 0; i < keys.size(); i++){
-		emit_signal(SNAME("selected"), keys[i]);
+	if (search_options->get_select_mode() == search_options->SELECT_SINGLE) {
+		emit_signal(SNAME("selected"), ti->get_metadata(0));
 	}
-	selected_properties.clear();
+	if (search_options->get_select_mode() == search_options->SELECT_MULTI) {
+		Array keys = selected_properties.keys();
+		for (int i = 0; i < keys.size(); i++) {
+			emit_signal(SNAME("selected"), keys[i]);
+		}
+		selected_properties.clear();
+	}
 	hide();
 }
 
@@ -374,7 +379,7 @@ void PropertySelector::_multi_item_selected(Object *p_item, int p_column, bool p
 	TreeItem *item = Object::cast_to<TreeItem>(p_item);
 	if (p_is_item_selected) {
 		selected_properties[item->get_metadata(0)];
-	}	else {
+	} else {
 		selected_properties.erase(item->get_metadata(0));
 	}
 }
@@ -567,6 +572,10 @@ void PropertySelector::set_type_filter(const Vector<Variant::Type> &p_type_filte
 	type_filter = p_type_filter;
 }
 
+void PropertySelector::set_multiselect(bool p_is_multiselect) {
+	search_options->set_select_mode(p_is_multiselect ? search_options->SELECT_MULTI : search_options->SELECT_SINGLE);
+}
+
 void PropertySelector::_bind_methods() {
 	ADD_SIGNAL(MethodInfo("selected", PropertyInfo(Variant::STRING, "name")));
 }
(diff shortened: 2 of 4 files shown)


AThousandShips:
Please open a [proposal](https://github.com/godotengine/godot-proposals) for this to track the demand and details for this 🙂

editor/property_selector.h
```diff
@@ -48,6 +48,7 @@ class PropertySelector : public ConfirmationDialog {
 	void _update_search();
 	void _confirmed();
 	void _item_selected();
+	void _multi_item_selected(Object *p_item, int column,bool is_item_selected);
```
AThousandShips, line 51:
```suggestion
	void _multi_item_selected(Object *p_item, int p_ column, bool p_is_item_selected);
```
Prefix arguments with `p_`, and style

editor/property_selector.cpp
```diff
@@ -343,15 +343,29 @@ void PropertySelector::_update_search() {
...
 	}
-	emit_signal(SNAME("selected"), ti->get_metadata(0));
+	for (int i = 0; i < keys.size(); i++){
+		emit_signal(SNAME("selected"), keys[i]);
+	}
+	selected_properties.clear();
 	hide();
 }
 
+void PropertySelector::_multi_item_selected(Object *p_item, int column,bool is_item_selected) {
```
AThousandShips, line 360:
```suggestion
void PropertySelector::_multi_item_selected(Object *p_item, int p_column, bool p_is_item_selected) {
```

LeeWannacott:
> Please open a [proposal](https://github.com/godotengine/godot-proposals) for this to track the demand and details for this 🙂

I think I saw a tracker issue with a bunch of improvements for animationPlayer that may have included this? I'll open a specific issue for this tomorrow.

editor/property_selector.cpp
```diff
@@ -343,15 +343,29 @@ void PropertySelector::_update_search() {
 	get_ok_button()->set_disabled(root->get_first_child() == nullptr);
 }
 
+Dictionary selected_properties;
```
AThousandShips, line 346: This should not be added outside the function, it should be in the header

AThousandShips (dismissed):

AThousandShips:
Converted to draft as your initial comment said it was a draft, in the future keep in mind you can make draft PRs

AThousandShips:
Updated the title to reflect that this doesn't just affect animation but all cases where this is used

However this needs some further work, the selector needs an option to enable multi select because in many places it's not going to be supported

Also the way it works here isn't a good solution IMO, it should have a signal that handles the multiple selects and send an array instead

LeeWannacott:
> Please open a [proposal](https://github.com/godotengine/godot-proposals) for this to track the demand and details for this 🙂

Opened proposal: https://github.com/godotengine/godot-proposals/issues/8713

editor/property_selector.h
LeeWannacott: Fixed in latest commit

editor/property_selector.cpp
LeeWannacott: Fixed in latest commit

editor/property_selector.cpp
LeeWannacott: Fixed in latest commit

LeeWannacott:
> Updated the title to reflect that this doesn't just affect animation but all cases where this is used

It looks like its just used in animationPlayer for adding property tracks, bezier curve tracks and blend shape tracks (which are all fine for multi select *I think...). Although if you, or anyone else can think of other cases where it is used...

> the selector needs an option to enable multi select 

This sounds like a horrible user experience.

> because in many places it's not going to be supported

Like where?

> Also the way it works here isn't a good solution IMO, it should have a signal that handles the multiple selects and send an array instead.

It might be interesting to see how multiple select is implemented in the scene tree because you can select multiple nodes there and they seemingly didn't create a new signal for returning an array.

> signal that handles the multiple selects and send an array instead.

Well it is using an existing signal and an array the dictionary is an intermediary step. I think an array would be more complicated.

The problem with arrays are you can end up with duplicates, you don't know where things are in the array, you can end up going out of bounds. You could probably make the case for a set/hashset ,or whatever the equivalent is in cpp, because I don't think I'm using a value with it.

LeeWannacott:
> signal that handles the multiple selects and send an array instead.

Did I misinterpret do you mean this code:
for (int i = 0; i < keys.size(); i++){
		emit_signal(SNAME("selected"), keys[i]);
	}

should have a signal for sending an array?

AThousandShips:
Indeed a second signal, in fact take a look at the code in `scene/gui/file_dialog.cpp` and check the way it handles it, it is a good reference

So some details:
* There should be an option, something like `allow_multiselect` or something, which is false by default, this should control the `Tree` selection behavior
* A new signal should be added that sends an array instead of a single object in case multiple are selected (or if multi-select is enabled)

Unable to be very active right now but will be around in a few days if nothing else to continue to assist 🙂

editor/property_selector.cpp
```diff
@@ -552,18 +574,21 @@ PropertySelector::PropertySelector() {
...
 	search_options = memnew(Tree);
+	search_options->set_select_mode(search_options->SELECT_SINGLE);
 	vbc->add_margin_child(TTR("Matches:"), search_options, true);
 	set_ok_button_text(TTR("Open"));
 	get_ok_button()->set_disabled(true);
 	register_text_enter(search_box);
 	set_hide_on_ok(false);
-	search_options->connect("item_activated", callable_mp(this, &PropertySelector::_confirmed));
 	search_options->connect("cell_selected", callable_mp(this, &PropertySelector::_item_selected));
+	search_options->connect("multi_selected", callable_mp(this, &PropertySelector::_multi_item_selected));
```
LeeWannacott, line 590: Might be able to move this up into set_multiselect if is_multiselect is true, however this signal wont trigger unless SELECT_MULTI is true anyway... :thinking:

editor/property_selector.cpp
```diff
@@ -552,18 +574,21 @@ PropertySelector::PropertySelector() {
 	VBoxContainer *vbc = memnew(VBoxContainer);
 	add_child(vbc);
 	//set_child_rect(vbc);
+
 	search_box = memnew(LineEdit);
 	vbc->add_margin_child(TTR("Search:"), search_box);
 	search_box->connect("text_changed", callable_mp(this, &PropertySelector::_text_changed));
 	search_box->connect("gui_input", callable_mp(this, &PropertySelector::_sbox_input));
 	search_options = memnew(Tree);
+	search_options->set_select_mode(search_options->SELECT_SINGLE);
```
LeeWannacott, line 583: Select single as default; this only applies when have made a new PropertySelector not when switching between tracks whilst using the same PropertySelector.

editor/property_selector.cpp
```diff
@@ -348,10 +348,28 @@ void PropertySelector::_confirmed() {
 	if (!ti) {
 		return;
 	}
-	emit_signal(SNAME("selected"), ti->get_metadata(0));
+	if (search_options->get_select_mode() == search_options->SELECT_SINGLE) {
```
LeeWannacott, line 351: I thought about using a switch statement, but I think it would just take up more space and considering there is only two/ three states if you count SELECT_ROW I'm not sure if it's worth it.

editor/animation_track_editor.cpp
```diff
@@ -4740,11 +4740,13 @@ void AnimationTrackEditor::_new_track_node_selected(NodePath p_path) {
 
 	switch (adding_track_type) {
 		case Animation::TYPE_VALUE: {
+			prop_selector->set_multiselect(true);
```
LeeWannacott, line 4743: We have to set_multiselect each time user switches between type of track to add. This is because the same PropertySelector is used for three of the types: (Property, Bezier and Blend)

The other instance of a PropertySelector is method_selector which would default to single on creation and its only used in one place.

Should we consider using different PropertySelector for each animation, or a signal that says when the type of animation has changed for that instance of PropertySelector. Or just keep this approach?

LeeWannacott:
@AThousandShips 
https://github.com/godotengine/godot/blob/master/editor/property_selector.cpp#L58-L71
Is this code actually doing anything? removing it doesn't seem to break any functionality with SELECT_SINGLE. 

Looking at:
https://github.com/godotengine/godot/blob/master/editor/editor_quick_open.cpp#L190-L204
They have this code for when SELECT_MULTI is selected.

So, I am wondering if this code was necessary for SELECT_SINGLE, or just placed here without much thought as to what it does.

Also, what menu/button does `editor_quick_open` refer to; in the editor, or is it opening the editor?

AThousandShips:
Will take a look when I find some time 🙂

LeeWannacott:
https://github.com/godotengine/godot/pull/93179 . Ahhh, I think I was encountering this crash; makes more sense now. I thought it was a me problem; but I guess its an actual bug.

TheSofox:
> We wait till this is fixed/merged...

https://github.com/godotengine/godot/issues/92958 Has been fixed and merged.

LeeWannacott:
@[TheSofox](https://github.com/TheSofox). I'm still getting a crash pressing down key with nothing in search box (only with multi-select enabled) after rebasing from master (assuming I rebased correctly)

PropertySelector has its own code relating to the search box input (is it possible the bug is within property_selector.cpp?):

Note: if I comment out `current->select(0);` the crash doesn't happen (of course neither does the blue selection).

``` cpp
void PropertySelector::_sbox_input(const Ref<InputEvent> &p_ie) {
	Ref<InputEventKey> k = p_ie;

	if (k.is_valid()) {
		switch (k->get_keycode()) {
			case Key::UP:
			case Key::DOWN:
			case Key::PAGEUP:
			case Key::PAGEDOWN: {
				search_options->gui_input(k);
				search_box->accept_event();

				TreeItem *root = search_options->get_root();
				if (!root->get_first_child()) {
					break;
				}

				TreeItem *current = search_options->get_selected();

				TreeItem *item = search_options->get_next_selected(root);
				while (item) {
					item->deselect(0);
					item = search_options->get_next_selected(item);
				}

				current->select(0);

			} break;
			default:
				break;
		}
	}
}
```


I don't know if the following error message means anything to you...:
```
Engine version: Godot Engine v4.3.beta.custom_build (cd9894558bcf1e58a3b207456d152a731e846e81)
[1] /lib/x86_64-linux-gnu/libc.so.6(+0x42520) [0x7ff321642520] (??:0)
[2] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x2e97c44) [0x5855571ccc44] (??:0)
[3] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x1ba3d59) [0x585555ed8d59] (??:0)
[4] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x1baab85) [0x585555edfb85] (??:0)
[5] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x4f524d5) [0x5855592874d5] (??:0)
[6] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x2b89795) [0x585556ebe795] (??:0)
[7] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x2a5ac0e) [0x585556d8fc0e] (??:0)
[8] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x2a71dac) [0x585556da6dac] (??:0)
[9] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x2a70eb6) [0x585556da5eb6] (??:0)
[10] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x2a71f2d) [0x585556da6f2d] (??:0)
[11] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x2ac57d5) [0x585556dfa7d5] (??:0)
[12] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x60fa59) [0x585554944a59] (??:0)
[13] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x62297c) [0x58555495797c] (??:0)
[14] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x4bd34a8) [0x585558f084a8] (??:0)
[15] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x4bd627d) [0x585558f0b27d] (??:0)
[16] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x624153) [0x585554959153] (??:0)
[17] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x5f9e00) [0x58555492ee00] (??:0)
[18] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x5c63bc) [0x5855548fb3bc] (??:0)
[19] /lib/x86_64-linux-gnu/libc.so.6(+0x29d90) [0x7ff321629d90] (??:0)
[20] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x80) [0x7ff321629e40] (??:0)
[21] /home/lee/Desktop/godot/bin/godot.linuxbsd.editor.x86_64(+0x5f7605) [0x58555492c605] (??:0)
```

editor/property_selector.cpp
```diff
@@ -554,27 +572,35 @@ void PropertySelector::set_type_filter(const Vector<Variant::Type> &p_type_filte
...
 	search_box = memnew(LineEdit);
 	vbc->add_margin_child(TTR("Search:"), search_box);
 	search_box->connect(SceneStringName(text_changed), callable_mp(this, &PropertySelector::_text_changed));
 	search_box->connect(SceneStringName(gui_input), callable_mp(this, &PropertySelector::_sbox_input));
 	search_options = memnew(Tree);
 	search_options->set_auto_translate_mode(AUTO_TRANSLATE_MODE_DISABLED);
+	search_options->set_select_mode(search_options->SELECT_SINGLE);
 	vbc->add_margin_child(TTR("Matches:"), search_options, true);
-	set_ok_button_text(TTR("Open"));
+	set_ok_button_text(TTR("Select"));
```
LeeWannacott, line 597: I think Select is more appropriate for the property selector, as we aren't `Open` ing anything :thinking:

editor/property_selector.cpp
```diff
@@ -409,10 +425,12 @@ void PropertySelector::_hide_requested() {
 
 void PropertySelector::_notification(int p_what) {
 	switch (p_what) {
+		case 30: {
```
LeeWannacott, line 428: @AThousandShips I don't know what `case 30` corresponds to https://github.com/godotengine/godot/blob/master/scene/main/node.h#L364-L413 (its not listed)? it seems to emit from canceling the PropertySelector window closing it with the x in the top right and confirming selection with the `Open` button (is it like a default for dialogue), am I allowed to name this 30 signal and if so, what to name it ?

AThousandShips: `30` is `NOTIFICATION_DRAW` in `CanvasItem`, see:
https://github.com/godotengine/godot/blob/97b8ad1af0f2b4a216f6f1263bef4fbc69e56c7b/scene/main/canvas_item.h#L188

But I don't know why this was added like this

LeeWannacott: Oh okay, I probably shouldn't be using it to clear properties then :thinking:

TheSofox:
> @[TheSofox](https://github.com/TheSofox). I'm still getting a crash pressing down key with nothing in search box (only with multi-select enabled) after rebasing from master (assuming I rebased correctly)

Could you provide more detailed repo steps? Also, could you confirm if this bug appears in master and not just this PR? I'm looking into it.

LeeWannacott:
> > @[TheSofox](https://github.com/TheSofox). I'm still getting a crash pressing down key with nothing in search box (only with multi-select enabled) after rebasing from master (assuming I rebased correctly)
> 
> Could you provide more detailed repo steps? Also, could you confirm if this bug appears in master and not just this PR? I'm looking into it.

Master doesn't have muli-select for the PropertySelector, so I can't really test it other than enabling it and it crashes "to rule out anything else Ive done from my PR". 

Steps to reproduce:

1. Add an AnimationPlayer to a scene.
2. Inside AnimationPlayer select `+ Add track`
3. Select `Property Track...` type
4. Select a node to add properties e.g `Sprite2D`
5. PropertySelector comes up with the searchbar (for me it is autofocused so I don't have to select the bar (this might be dependent on some settings e.g godot ,or the OS...))
6. Press down arrow on keyboard results in a crash.

TheSofox:
Was able to repo from your steps. From what I can tell, the bug wasn't caused by any code I changed one way or another. From what I can see, the function `current->select(0);` simply shouldn't be called when the select mode is `SELECT_MULTI`. A quick fix is to only call that function if the mode if `if (search_options->get_select_mode() != Tree::SELECT_MULTI ) {` or something, and then decide what behaviour you want if it is selected and code that in the other branch (Maybe using `select_single_item`). But yeah, as far as I can tell that's what the issue is, code just wasn't set up to deal with different select modes.

LeeWannacott:
> Was able to repo from your steps. From what I can tell, the bug wasn't caused by any code I changed one way or another. From what I can see, the function `current->select(0);` simply shouldn't be called when the select mode is `SELECT_MULTI`. A quick fix is to only call that function if the mode if `if (search_options->get_select_mode() != Tree::SELECT_MULTI ) {` or something, and then decide what behaviour you want if it is selected and code that in the other branch (Maybe using `select_single_item`). But yeah, as far as I can tell that's what the issue is, code just wasn't set up to deal with different select modes.

The editor quick open https://github.com/godotengine/godot/blob/master/editor/editor_quick_open.cpp#L216-L231 seems to do the opposite,  it `current->select(0);` if `SELECT_MULTI` is true?

Which I think is the window below (scene->quick open), although maybe it doesn't error the first item is already selected/focused  before pressing down on the key (by some other part of the code).
image

PR #86404, opened 2023-12-21
Windows export rcedit improvements

Joseph-DiGiovanni:
- Allows rcedit to be executed via system PATH without specifying an executable location in editor settings.

- When rcedit is executed through wine on non-windows systems, disable unnecessary wine debug output and the wine-mono install message on first run.

- Minor code refactoring for improved readability.


platform/windows/export/export_plugin.cpp
```diff
@@ -513,9 +497,24 @@ Error EditorExportPlatformWindows::_rcedit_add_data(const Ref<EditorExportPreset
...
+		}
+
+		if (wine_path.is_empty()) {
+			wine_path = "wine"; // try to run wine from PATH
+		}
+
+		args.push_front(rcedit_path);
+		rcedit_path = wine_path;
+		setenv("WINEDEBUG", "-all", true); // Disable wine debug messages
+		setenv("WINEDLLOVERRIDES", "mscoree=", true); // Prevent wine-mono install prompt on first run
```
Calinou, line 516: I suggest using `OS::get_singleton()->set_environment()` instead for consistency with the rest of the engine, even if this block is UNIX-specific.

Joseph-DiGiovanni: Just changed it, thanks.

Joseph-DiGiovanni:
Sorry, accidentally closed the PR renaming the branch. Reopening.

Joseph-DiGiovanni:
> Are you trying to make this PR fetch from another branch?

I was 😅 I saw there is another PR attempting to remove rcedit altogether in favor of a built in solution, so I'm not even sure it's worth resubmitting anyway.

PR #86412, opened 2023-12-22
Fix data race in PagedArray

stuartcarnie:
This PR resolves a data race in the `PagedArrayPool` and `PagedArray` types. The thread sanitiser identified the data race via the `page_pool` field of the `PagedArrayPool` class. The write occurred:

https://github.com/godotengine/godot/blob/9d1cbab1c432b6f1d66ec939445bec68b6af519e/core/templates/paged_array.h#L62

and the concurrent read here:

https://github.com/godotengine/godot/blob/9d1cbab1c432b6f1d66ec939445bec68b6af519e/core/templates/paged_array.h#L78

The actual data race of the `page_pool` pointer being access concurrently is possibly safe, as it is a word read / write, however, an additional problem is the underlying memory in the `memrealloc` from the old allocation is being immediately zero'd in debug builds on macOS, so the call the `get_page` is sometimes returning an `nullptr`. In a release build, the old allocation may be immediately reused before the call to `get_page`, returning an invalid pointer, resulting in crashes, memory corruption, etc.

### DIFF
--- a/core/templates/paged_array.h
+++ b/core/templates/paged_array.h
@@ -53,7 +53,12 @@ class PagedArrayPool {
 	SpinLock spin_lock;
 
 public:
-	uint32_t alloc_page() {
+	struct PageInfo {
+		T *page = nullptr;
+		uint32_t page_id = 0;
+	};
+
+	PageInfo alloc_page() {
 		spin_lock.lock();
 		if (unlikely(pages_available == 0)) {
 			uint32_t pages_used = pages_allocated;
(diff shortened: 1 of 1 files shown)


core/templates/paged_array.h
```diff
@@ -53,7 +53,12 @@ class PagedArrayPool {
 	SpinLock spin_lock;
 
 public:
-	uint32_t alloc_page() {
+	struct PageInfo {
+		T *page;
```
RandomShaper, line 57: I'd suggest adding initializers. That's the trend in Godot lately. I think not adding them should be the exception (for spots where perf is critical and the compiler can't be trusted to optimize out redundant assignments).

RandomShaper (changes requested):
Looks good besides the little nitpick.

core/templates/paged_array.h
stuartcarnie: Sounds good 👍🏻

RandomShaper (approved):

akien-mga:

stuartcarnie:
@RandomShaper squashed

akien-mga:
Thanks! And congrats for your first merged Godot contribution :tada: -- hopefully the start of many, given what I know you've been cooking ;)

akien-mga:
Cherry-picked for 4.2.2.

PR #2316, opened 2015-07-29
Clear output console (Button + clear on Play)

StraToN:
This PR resolves Feature #1825 
- Added a "Clear" button above output console
- Added a Setting option to automatize clearing when Play is started

PR #86430, opened 2023-12-22
Improve 3D Asset Import Dialog material editing

reduz:
Implements [7238](https://github.com/godotengine/godot-proposals/issues/7238) (for the most part).

* Ability to select nodes, meshes or materials by clicking.
* Moved mesh/material previews to the inspector area.
* Ability to override materials and edit them directly in the inspector.
* By default, shows overridden materials in the imported scene, with a button to disable them and show original materials.
* Keeps compatibility with the old override format.

**Note:** Adding preview sun/env will be left for another PR.

How it looks:

asset_import_usability

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -1284,7 +1284,7 @@ void EditorNode::save_resource(const Ref<Resource> &p_resource) {
 	if (path.is_resource_file() && !FileAccess::exists(path + ".import")) {
 		save_resource_in_path(p_resource, p_resource->get_path());
 	} else {
-		save_resource_as(p_resource);
+		save_resource_as(p_resource, "");
 	}
 }
 
--- a/core/math/triangle_mesh.cpp
+++ b/core/math/triangle_mesh.cpp
@@ -107,12 +107,12 @@ void TriangleMesh::get_indices(Vector<int> *r_triangles_indices) const {
 void TriangleMesh::create(const Vector<Vector3> &p_faces, const Vector<int32_t> &p_surface_indices) {
 	valid = false;
 
-	ERR_FAIL_COND(p_surface_indices.size() && p_surface_indices.size() != p_faces.size());
-
 	int fc = p_faces.size();
 	ERR_FAIL_COND(!fc || ((fc % 3) != 0));
 	fc /= 3;
 	triangles.resize(fc);
+
+	ERR_FAIL_COND(p_surface_indices.size() && p_surface_indices.size() != triangles.size());
 
 	bvh.resize(fc * 3); //will never be larger than this (todo make better)
 	BVH *bw = bvh.ptrw();
--- a/editor/editor_resource_picker.cpp
+++ b/editor/editor_resource_picker.cpp
@@ -398,6 +398,7 @@ void EditorResourcePicker::_edit_menu_cbk(int p_which) {
 			if (edited_resource.is_null()) {
 				return;
 			}
+
 			EditorNode::get_singleton()->save_resource(edited_resource);
 		} break;
 
@@ -405,6 +406,7 @@ void EditorResourcePicker::_edit_menu_cbk(int p_which) {
 			if (edited_resource.is_null()) {
 				return;
 			}
+
 			EditorNode::get_singleton()->save_resource_as(edited_resource);
 		} break;
 
--- a/editor/icons/MaterialSelect.svg
+++ b/editor/icons/MaterialSelect.svg
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg id="svg29" width="16" height="16" version="1.1" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
+ <path id="path27" d="m8 2 6 3v6l-6 3-6-3V5zm0 12V8l6-3M8 8 2 5" fill="none" stroke="#e0e0e0" stroke-linejoin="round" stroke-width="2"/>
+</svg>
--- a/editor/icons/MaterialSelectDisabled.svg
+++ b/editor/icons/MaterialSelectDisabled.svg
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   id="svg29"
+   width="16"
+   height="16"
+   version="1.1"
+   viewBox="0 0 16 16"
+   sodipodi:docname="MaterialSelectDisabled.svg"
+   inkscape:version="1.1.2 (0a00cf5339, 2022-02-04)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <defs
+     id="defs6" />
+  <sodipodi:namedview
+     id="namedview4"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageshadow="2"
+     inkscape:pageopacity="0.0"
+     inkscape:pagecheckerboard="0"
+     showgrid="false"
+     inkscape:zoom="54.5625"
+     inkscape:cx="4.325315"
+     inkscape:cy="8"
+     inkscape:window-width="2490"
+     inkscape:window-height="1376"
+     inkscape:window-x="70"
+     inkscape:window-y="27"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg29" />
+  <path
+     id="path27"
+     d="m8 2 6 3v6l-6 3-6-3V5zm0 12V8l6-3M8 8 2 5"
+     fill="none"
+     stroke="#e0e0e0"
+     stroke-linejoin="round"
+     stroke-width="2"
+     style="stroke:#808080;stroke-opacity:1" />
+</svg>
--- a/editor/icons/MeshSelect.svg
+++ b/editor/icons/MeshSelect.svg
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   height="16"
+   viewBox="0 0 16 16"
+   width="16"
+   version="1.1"
+   id="svg827"
+   sodipodi:docname="MeshSelect.svg"
+   inkscape:version="1.1.2 (0a00cf5339, 2022-02-04)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <defs
+     id="defs831" />
+  <sodipodi:namedview
+     id="namedview829"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageshadow="2"
+     inkscape:pageopacity="0.0"
+     inkscape:pagecheckerboard="0"
+     showgrid="false"
+     inkscape:zoom="54.5625"
+     inkscape:cx="7.9908362"
+     inkscape:cy="8"
+     inkscape:window-width="2490"
+     inkscape:window-height="1376"
+     inkscape:window-x="70"
+     inkscape:window-y="27"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg827" />
+  <path
+     d="M4.73 2A2 2 0 1 0 2 4.73v6.541A2 2 0 1 0 4.73 14h6.541A2 2 0 1 0 14 11.27V4.729A2 2 0 1 0 11.27 2Zm.683 2h5.857a2 2 0 0 0 .729.73v5.856L5.412 4zM3.999 5.414 10.584 12H4.727a2 2 0 0 0-.729-.73V5.414z"
+     fill="#808080"
+     id="path825"
+     style="fill:#e0e0e0;fill-opacity:1" />
+</svg>
--- a/editor/icons/NodeSelect.svg
+++ b/editor/icons/NodeSelect.svg
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   height="16"
+   viewBox="0 0 16 16"
+   width="16"
+   version="1.1"
+   id="svg4"
+   sodipodi:docname="NodeSelect.svg"
+   inkscape:version="1.1.2 (0a00cf5339, 2022-02-04)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <defs
+     id="defs8" />
+  <sodipodi:namedview
+     id="namedview6"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageshadow="2"
+     inkscape:pageopacity="0.0"
+     inkscape:pagecheckerboard="0"
+     showgrid="false"
+     inkscape:zoom="54.5625"
+     inkscape:cx="4.325315"
+     inkscape:cy="8"
+     inkscape:window-width="2490"
+     inkscape:window-height="1376"
+     inkscape:window-x="70"
+     inkscape:window-y="27"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg4" />
+  <circle
+     cx="8"
+     cy="8"
+     r="5"
+     fill="none"
+     stroke-width="2"
+     stroke="#e0e0e0"
+     id="circle2" />
+</svg>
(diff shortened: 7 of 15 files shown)


editor/editor_node.cpp
```diff
@@ -1324,22 +1324,31 @@ void EditorNode::save_resource_as(const Ref<Resource> &p_resource, const String
 		}
 	}
 
-	file->set_file_mode(EditorFileDialog::FILE_MODE_SAVE_FILE);
+	if (p_custom_file_dialog) {
```
YuriSizov, line 1327: If you've done all this just to work around the issue where the file dialog appears behind your current dialog, then we have a system for auto-reparenting, which would be much simpler.

The dialog needs to be set up with `set_unparent_when_invisible(true)`, and then to pop it you'd use one of the `EditorInterface::get_singleton()->popup_dialog_*` methods.

editor/import/scene_import_settings.cpp
```diff
@@ -1144,18 +1280,43 @@ void SceneImportSettingsDialog::_notification(int p_what) {
...
+			for (int i = 0; i < TOOL_MAX; i++) {
+				static const char *icons[TOOL_MAX] = {
+					"ToolSelect",
+					"ListSelect",
+					"NodeSelect",
+					"MeshSelect",
+					"MaterialSelect",
+					"StandardMaterial3D",
+				};
+				tool_buttons[i]->set_icon(get_theme_icon(icons[i], "EditorIcons"));
```
YuriSizov, line 1305:
```suggestion
				tool_buttons[i]->set_icon(get_editor_theme_icon(icons[i]));
```

We have a special method now, to safe on string names.

editor/import/scene_import_settings.cpp
```diff
@@ -1144,18 +1280,43 @@ void SceneImportSettingsDialog::_notification(int p_what) {
...
+			}
+
+			tool_buttons[TOOL_PREVIEW_MATERIALS]->add_theme_color_override("icon_normal_color", Color(1, 1, 1, 0.5));
+			tool_buttons[TOOL_PREVIEW_MATERIALS]->add_theme_color_override("icon_pressed_color", Color(1, 1, 1, 1.0));
+
+			Ref<LabelSettings> ls;
+			ls.instantiate();
+			// Needs to be high contrast over checkered background, not depend on theme
+			ls->set_font_color(Color(1, 1, 1));
+			ls->set_font(get_theme_font("bold", "EditorFonts"));
```
YuriSizov, line 1315:
```suggestion
			ls->set_font(get_theme_font("bold", EditorStringName(EditorFonts)));
```

And for everything else it's like this

YuriSizov:
You have a leak related to themes, though I'm not really sure why.

editor/plugins/script_editor_plugin.h
```diff
@@ -303,6 +303,7 @@ class ScriptEditor : public PanelContainer {
 	ScriptCreateDialog *script_create_dialog = nullptr;
 	Button *scripts_visible = nullptr;
 	FindReplaceBar *find_replace_bar = nullptr;
+	EditorFileDialog *save_file_dialog;
```
AThousandShips, line 306:
```suggestion
	EditorFileDialog *save_file_dialog = nullptr;
```

editor/import/scene_import_settings.h
```diff
@@ -74,7 +82,25 @@ class SceneImportSettingsDialog : public ConfirmationDialog {
 
 	EditorInspector *inspector = nullptr;
 
-	SubViewport *base_viewport = nullptr;
+	SubViewportContainer *vp_container;
```
AThousandShips, line 85:
```suggestion
	SubViewportContainer *vp_container = nullptr;
```

editor/import/scene_import_settings.cpp
```diff
@@ -1023,47 +1029,177 @@ void SceneImportSettingsDialog::_cleanup() {
...
+							misort.mesh_instance = mi;
+							misort.surface_index = surf_index;
+							misort.point_d = ray_dir.dot(p);
+							mesh_instances.push_back(misort);
+						}
+					}
+				}
+			}
+
+			if (mesh_instances.size() == 0) {
```
AThousandShips, line 1081:
```suggestion
			if (mesh_instances.is_empty()) {
```

editor/import/scene_import_settings.cpp
```diff
@@ -1023,47 +1029,177 @@ void SceneImportSettingsDialog::_cleanup() {
...
+					String material_id = mesh_map[nd.mesh_id].material_surface_ids[mesh_instances[i].surface_index];
+					if (!material_map.has(material_id)) {
+						continue;
+					}
+					pick_list.push_back(material_id);
+					pick_names.push_back(material_map[material_id].name);
+				}
+			}
+
+			if (pick_list.size() == 0) {
```
AThousandShips, line 1154:
```suggestion
			if (pick_list.is_empty()) {
```

editor/import/scene_import_settings.cpp
```diff
@@ -1023,47 +1029,177 @@ void SceneImportSettingsDialog::_cleanup() {
...
+			dragging = false;
+		} else if (!dragging) {
+			// Okay select something
+
+			Vector3 ray_from = camera->project_ray_origin(mb->get_position());
+			Vector3 ray_dir = camera->project_ray_normal(mb->get_position());
+
+			Vector<ObjectID> instances = RenderingServer::get_singleton()->instances_cull_ray(ray_from, ray_from + ray_dir * 10000, scene_viewport->find_world_3d()->get_scenario());
+
+			if (instances.size() == 0) {
```
AThousandShips, line 1050:
```suggestion
			if (instances.is_empty()) {
```

editor/editor_node.cpp
reduz: Unfortunately a lot of dialogs can't be popped up like that because they have custom functions to do it. I feel it may be better to just fix the issue properly. I will look into it.

editor/import/scene_import_settings.cpp
reduz: Oh, nice, will use this.

editor/import/scene_import_settings.cpp
reduz: alright

editor/editor_node.cpp
YuriSizov: We can definitely extend this system for EditorFileDialog, since it's an editor-only node, without resolving it manually across all classes that need this.

fire:
@MewPurPur Tends to optimize the svg icons after the merges.

MewPurPur:
I'll do another big rundown at some later point, I'm busy these days.

coderarnold:
An issue I have with the import workflow of 3d meshes and materials is :

when I override the materials of mesh instance nodes using "Surface Material Override" (under MeshInstance3D in the node inspector), the materials are reset / cleared for ALL mesh instances using that mesh, in ALL scenes, whenever I re-import the mesh (and it has changed somehow since the last import). **This is VERY limiting**. A mesh can be used in a variety of materials combinations, but in its current state, Godot is more suited towards meshes with a fixed set of assigned materials (because materials assigned to the mesh in the import window stay assigned on the mesh itself, and don't reset anytime the mesh is updated)

The "Material override" property (under GeometryInstance3D in the node inspector) doesn't get reset though, on mesh re-import, but the problem is it doesn't allow assigning different materials to different parts of a mesh, whereas "Surface Material Override" can.

fire:
Did the behaviour change?

As far as I know the ArrayMesh materials can change due to import, but MeshInstance3D's overrides don't change.

coderarnold:
> Did the behaviour change?
> 
> As far as I know the ArrayMesh materials can change due to import, but MeshInstance3D's overrides don't change.

Oops, you're right, I just tried and it does work indeed. Sorry.
That's great then!

Has this been fixed in 4.2?

editor/import/scene_import_settings.h
```diff
@@ -185,7 +214,20 @@ class SceneImportSettingsDialog : public ConfirmationDialog {
...
+		TOOL_SELECT,
+		TOOL_SELECT_LIST,
+		TOOL_SELECT_OBJECT,
+		TOOL_SELECT_MESH,
+		TOOL_SELECT_MATERIAL,
+		TOOL_PREVIEW_MATERIALS,
+		TOOL_MAX
+	};
+
+	Button *tool_buttons[TOOL_MAX];
```
AThousandShips, line 227:
```suggestion
	Button *tool_buttons[TOOL_MAX] = {};
```
Think this is a good way to initialize these cleanly

coderarnold:
> > Did the behaviour change?
> > As far as I know the ArrayMesh materials can change due to import, but MeshInstance3D's overrides don't change.
> 
> Oops, you're right, I just tried and it does work indeed. Sorry. That's great then!
> 
> Has this been fixed in 4.2?

No, the issue is still there in  4.2. 
If you modify the mesh then re-import it, the correct material overrides are still shown in open scenes. But close and reopen them, or open not yet opened scenes having mesh instances with material overrides, and they will be all cleared to nothing.

editor/import/scene_import_settings.h
```diff
@@ -32,6 +32,8 @@
 #define SCENE_IMPORT_SETTINGS_H
 
 #include "editor/import/resource_importer_scene.h"
+#include "editor/plugins/material_editor_plugin.h"
+#include "editor/plugins/mesh_editor_plugin.h"
```
AThousandShips, line 36: These can reasonably be forward declared instead

editor/import/scene_import_settings.cpp
```diff
@@ -169,22 +176,25 @@ void SceneImportSettingsDialog::_fill_material(Tree *p_tree, const Ref<Material>
...
 
 	Ref<Texture2D> icon = get_editor_theme_icon(SNAME("StandardMaterial3D"));
 
 	TreeItem *item = p_tree->create_item(p_parent);
-	if (p_material->get_name().is_empty()) {
-		item->set_text(0, TTR("<Unnamed Material>"));
-	} else {
+	if (!p_material->get_name().is_empty()) {
 		item->set_text(0, p_material->get_name());
+	} else if (p_surface_name != "") {
```
AThousandShips, line 186:
```suggestion
	} else if (!p_surface_name.is_empty()) {
```

reduz:
Will try to pick up this PR in the coming week, sorry for the delay!

reduz:
Alright, rebased, applied suggestions and updated.

Calinou (commented):

I noticed a few usability issues though:

- This tooltip says that it works with the right click, but it actually works by left-clicking from my testing. This makes me wonder if having two separate modes is even needed, since right-clicking doesn't seem to do anything in either mode.
  - There's also a typo in "materials" in the tooltip.

image

- We have a lot of space available in the top bar, so maybe this icon-only button should become a CheckButton with the text **Show Material Overrides**:

Screenshot_20240131_002635

smix8:
I tested and apart from what Calinou already mentioned in general the additions for the most part work fine for what they are supposed to do.

Since I regularly use high-detail models I encountered a few usability problems, primarily with the selection and info:

- On a more detailed model or with certain colors / light the node "selection aabb" is not really visible. It is also not really visible when you need to zoom in  to pick details. You cant see this outside selection at all when zoomed in and need to go back to the tree view or watch the inspector.

- The mesh and material selection has no highlight at all. Together with the gray-in-gray of the preview visuals you basically have a visual fog-soup for your selection.

Now for both issues imo a color highlight of the surface, e.g. a stripe overlay, would work better imo than this "selection aabb". It can be both oc.

Another issue is that with the information split between tree left, inspector right you now need to constantly switch your eyes between them for basic things.

- When a material is selected the inspector to the right it does not show the material name or surface slot. So to figure out what it is that you have selected you need the tree view again. This is especially jarring on a complex model where the tree view is big and needs scrolling / jumps on every selection.

The raycast for picking also has detail problems.

- On a more detailed model the raycast is hit and miss, it is really hard to pick smaller objects even while fully zoomed in. E.g. I have a lot of character models with very small accessory meshes close to other surfaces, I can not select them.

- In the other direction, models with many mesh surfaces very close overlayed regularly pick the wrong surface blocking the other surface. I can clearly see the correct surface rendered on top but the picker selects the wrong surface below that surface. This is likely happening because the trimesh ray check runs into precision issues and always picks the very first surface only that passes the intersection test.

Might need a way to click&hold with a popup that gives you all the objects in a small radius to choose like the main editor viewport has. If this is too difficult to add consider rotating through very close surfaces on every new click instead of only every picking the first surface that passes the ray test.

reduz:
@smix8 AFAIK these things are out of scope for this PR.

Outline selections (like in Blender or Unity) will most likely happen once we have a proper compositor, as right now this is not possible. There are other contributors willing to give more freedom to the camera to improve also navigation of the imported scene.

But then again, this PR should be taken as a starting point for those kind of improvements, not as a solution for everything.

QbieShay:
If there's planned improvements that we know we want, I think it's good practice to open new issues for them so that they don't get forgotten. I've done it multiple times with particle work and it's worked very well for me. 

RE: big selection trees: I think it'd be a good starting point to scroll the tree to the currently selected item?

RE:object picking: I think once stencil is in we can use an ID buffer that will be a lot more efficient. Before that is in we can also have just a secondary viewport with a special ID material, but I think this is a later improvement. Should probably be tracked though.

fire:
I'll try to get some videos reviewing the asset import dialog soon.

fire (approved):
I tested the feature and the improved 3d asset importer seems to work. Did not dig into the details of the code as much.

fire:
Will try to get some comments from the skeleton bone preview tool pull request developer

fire (changes requested):
We reviewed the pull request today in discord. Review comments incoming.

MarcusElg:
This is a huge improvement. Being able to replace a material with a custom one that has it's own shader for example feels like an important feature to have. This doesn't really seam to be possible today by using the extract materials functionality. Is there any estimating on when this can be merged?

quackmarx:
I think this is a huge step toward making the 3D import pipeline more useable and encouraging good workflows. I also think it can definitely be taken further.

I agree with @lyuma:

> There is no way to initialize the custom material with the original material in the scenario that you want to "extend" a material or make minor edits to it: it starts from scratch.

A very common reason for overriding materials is to do exactly this; adjust the albedo, increase the emission, etc. I think this might be even more common than needing to change everything about the material.

For me the main goal of this system should be to:

1. Get a clear overview of what is imported and how Godot reads this
2. Make adjustments

In other engines such as Unity, making adjustments is very barebones: you only adjust very general import settings with the ability to filter out animations. I think Godot has a huge opportunity with this Import Dialog to go much further and make it a solid part of the asset authoring pipeline. It already leans towards this by moving away from the inspector into a dedicated scene viewer. I hope in the future we can leave behind the "import settings" approach of Unity and into the "asset setup and configuration" approach where you can use this window to:

- Replace materials (this is already happening here)
- Edit material properties (3D artists love finetuning in-engine and indie-devs often use 3rd party assets directly without going through a DCC)
- View textures embedded with the model and re-map them to the desired materials
- Generate channel-packed textures such as ORM ([Add a way to combine PBR maps into an ORM texture for use in ORMMaterial3D #2316](https://github.com/godotengine/godot-proposals/issues/2316)). If channel-packing became core it would also benefit other applications such as when using Terrain 3D ([Preparing Textures for Terrain 3D](https://terrain3d.readthedocs.io/en/stable/docs/texture_prep.html).
- Convert textures of a wrong format into useable: DirectX normals to OpenGL, smoothness to roughness, etc.

While I recognize that there are workarounds for everything listed above, they often require a good amount of extra time, and time added to an asset pipeline scales linearly with complexity. Also, it often requires full control over the pipeline, which is not a good thing in my opinion. Godot should - within reason - be able to accommodate pipelines that aren't build from start to finish optimally for Godot and for different ways of working, be it rapid prototyping, using 3rd party assets or just mix and matching. Plus these changes also make everything more transparent in terms of what is happening and will, in my opinion, encourage better workflows than what I currently see used by less experienced developers/artists.

TLDR; I hope - with time - we can make the Asset Import Dialog the one-stop-shop for setting up everything about the asset, not leaving anything to the inspector afterwards and that it can reasonably support common use-cases such as remapping textures, tweaking material parameters, channel-packing and converting wrongful texture channels from other render setups into proper Godot PBR maps.

PadraigK:
Is there anything any of us can do to help get this landed?

lyuma:
> Is there anything any of us can do to help get this landed?

@PadraigK We urgently need to allow referencing ExtResource in .import files UID, and to fix the engine crash that happens if a .import file has a broken UID reference.

if anyone feels inspired to help with either of the above issues (in particular, changing the ConfigFile / .import serialization to reference ExtResource by uid://), it would be much appreciated!

Once that is fixed, I think everyone is in agreement to merge this PR.

(If we merge this PR without the .import reference fixes, it means projects are going to be broken or crashy the moment one material goes missing.)

PadraigK:
> > Is there anything any of us can do to help get this landed?
> 
> @PadraigK We urgently need to allow referencing ExtResource in .import files UID, and to fix the engine crash that happens if a .import file has a broken UID reference.
> 
> if anyone feels inspired to help with either of the above issues (in particular, changing the ConfigFile / .import serialization to reference ExtResource by uid://), it would be much appreciated!

@lyuma  Thank you for pointing out the issues that are blocking this PR! 

I am interested in investigating this further and maybe contributing a patch. I have started searching for more information in the issues and proposals to try to find more information and I've joined the development chat. Please let me know 
if there's anything you can point me to that describes these issues in more detail.

lyuma:
I was shocked to find that all of the issues blocking this PR were incorrectly closed due to a fix that **"just turns the crashes into error prints"**

I have opened a new tracker issue #99910 to track the critical .import UID issues that are blocking this PR.

lyuma:
Superseded by #107351 - closing. Feel free to move the discussion to the new PR (or discuss on the proposal itself)

PR #86431, opened 2023-12-22
Improve action name for ungroup button in Scene dock

timothyqiu:
This PR improves the name of the undo action triggered by this button:

the ungroup button

The undo action name was "Button Group", which is confusing.


YuriSizov (approved):

YuriSizov:
Thanks!

PR #86435, opened 2023-12-22
Add `RD_ENABLED` when `VULKAN_ENABLED` or `D3D12_ENABLED` is added

### LINKED ISSUE #86411: Forward plus unsupported error in 4.3 dev 1 for new projects
### Tested versions

4.3 dev 1

### System information

Windows 10

### Issue description

An error message pops up when you create a brand new project in an empty folder. 
The error is "The project uses the following features not supported by this build of Godot : Forward plus"

### Steps to reproduce

1. On the Godot project manager, select New 
2. Browse for a folder to put a new project in
3. Create a folder (right click in the browse window) and select the new folder as the project folder
4. Error message appears

N/A

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/86411

In https://github.com/godotengine/godot/pull/83452, @RandomShaper changed the following line to use `RD_ENABLED` instead of `VULKAN_ENABLED` and `D3D12_ENABLED`
https://github.com/godotengine/godot/blob/9d1cbab1c432b6f1d66ec939445bec68b6af519e/core/config/project_settings.cpp#L98-L101

But as far as I can see, `RD_ENABLED` is only defined in driver/vulkan and dx12, I'm not familiar with scon but I guess the definition there only affects the subfolder and so the editor failed to have this macro defined? Feel free to correct me.

Since I dont think revert this line is a good change as godot might support other backend for rd, this pr is a safeguard as I add `RD_ENABLED` when `VULKAN_ENABLED` or `D3D12_ENABLED` is added in detect.py, I might miss some places. Thank you for your help in advance.

### DIFF
--- a/drivers/d3d12/SCsub
+++ b/drivers/d3d12/SCsub
@@ -5,8 +5,6 @@ from pathlib import Path
 
 Import("env")
 
-env.Append(CPPDEFINES=["RD_ENABLED"])
-
 env_d3d12_rdd = env.Clone()
 
 thirdparty_obj = []
--- a/drivers/vulkan/SCsub
+++ b/drivers/vulkan/SCsub
@@ -2,8 +2,6 @@
 
 Import("env")
 
-env.Append(CPPDEFINES=["RD_ENABLED"])
-
 thirdparty_obj = []
 thirdparty_dir = "#thirdparty/vulkan"
 thirdparty_volk_dir = "#thirdparty/volk"
--- a/platform/android/detect.py
+++ b/platform/android/detect.py
@@ -200,7 +200,7 @@ def configure(env: "Environment"):
     env.Append(LIBS=["OpenSLES", "EGL", "android", "log", "z", "dl"])
 
     if env["vulkan"]:
-        env.Append(CPPDEFINES=["VULKAN_ENABLED"])
+        env.Append(CPPDEFINES=["VULKAN_ENABLED", "RD_ENABLED"])
         if not env["use_volk"]:
             env.Append(LIBS=["vulkan"])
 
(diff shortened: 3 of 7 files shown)


RandomShaper:
I wonder if all this could be managed at a more centric spot. Aside, this looks good to me. I think `env.Append(CPPDEFINES=["RD_ENABLED"])` can be removed from `drivers/vulkan/SCSub` and `drivers/d3d12/SCSub`. I think they weren't contributing anything in the end.

RandomShaper (approved):

jsjtxietian:
> I think `env.Append(CPPDEFINES=["RD_ENABLED"])` can be removed from `drivers/vulkan/SCSub` and `drivers/d3d12/SCSub`

Done.

qwofford:
This PR removed the error about Forward+ when launching a new project for me. I've read the documentation about Forward+, but it's still not clear to me exactly what it is. Is it Vulkan? A combination of tools?

I would like to test that my Forward+ is working correctly, now that this PR fixes the error message itself. I am able to load a 3D scene and interact with it, and the top right entry of my toolbar reads "Forward+". Is that sufficient evidence that things are working as intended?

jsjtxietian:
> I've read the documentation about Forward+, but it's still not clear to me exactly what it is.

https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/internal_rendering_architecture.html

I guess you can understand it as a combination of fancy technologys that applys well on normal desktop hardware

> the top right entry of my toolbar reads "Forward+"

I believe yes

akien-mga:
I'm surprised that the original code doesn't work, appending to `env` should affect the base environment everywhere. Maybe it needs an `Export("env")`?

Anyway, this makes things consistent with how `VULKAN_ENABLED` and `D3D12_ENABLED` are registered so this change seems fine.

akien-mga (approved):

akien-mga:
Thanks!

jsjtxietian:
I'm thinking about if we can do something like, if  `VULKAN_ENABLED` or `D3D12_ENABLED` enabled, add `RD_ENABLED` to `CPPDEFINES`.

PR #86441, opened 2023-12-22
[X11] Fix Godot stealing focus on alternative window managers

dsnopek:
This is taking the changes from https://github.com/Mequam/funky-godot/tree/mequam/feat/xfce4_focus_grab and turning them into a PR (as @Mequam, the original author, requested on https://github.com/godotengine/godot/issues/74378)

Unfortunately, these changes don't fix the issue I am experiencing, but since others (including the original author) wanted to see these changes as a PR, I've gone ahead and created one. :-) It's my understanding that these changes do fix some of the problems from that issue.

### DIFF
--- a/platform/linuxbsd/x11/display_server_x11.cpp
+++ b/platform/linuxbsd/x11/display_server_x11.cpp
@@ -2010,7 +2010,7 @@ void DisplayServerX11::window_set_transient(WindowID p_window, WindowID p_parent
 		// RevertToPointerRoot is used to make sure we don't lose all focus in case
 		// a subwindow and its parent are both destroyed.
 		if (!wd_window.no_focus && !wd_window.is_popup && wd_window.focused) {
-			if ((xwa.map_state == IsViewable) && !wd_parent.no_focus && !wd_window.is_popup) {
+			if ((xwa.map_state == IsViewable) && !wd_parent.no_focus && !wd_window.is_popup && _window_focus_check()) {
 				XSetInputFocus(x11_display, wd_parent.x11_window, RevertToPointerRoot, CurrentTime);
 			}
 		}
@@ -2945,7 +2945,7 @@ void DisplayServerX11::window_set_ime_active(const bool p_active, WindowID p_win
 		XWindowAttributes xwa;
 		XSync(x11_display, False);
 		XGetWindowAttributes(x11_display, wd.x11_xim_window, &xwa);
-		if (xwa.map_state == IsViewable) {
+		if (xwa.map_state == IsViewable && _window_focus_check()) {
 			XSetInputFocus(x11_display, wd.x11_xim_window, RevertToParent, CurrentTime);
 		}
 		XSetICFocus(wd.xic);
@@ -4228,6 +4228,22 @@ bool DisplayServerX11::mouse_process_popups() {
 	return closed;
 }
 
+bool DisplayServerX11::_window_focus_check() {
+	Window focused_window;
+	int focus_ret_state;
+	XGetInputFocus(x11_display, &focused_window, &focus_ret_state);
+
+	bool has_focus = false;
+	for (const KeyValue<int, DisplayServerX11::WindowData> &wid : windows) {
+		if (wid.value.x11_window == focused_window) {
+			has_focus = true;
+			break;
+		}
+	}
+
+	return has_focus;
+}
+
 void DisplayServerX11::process_events() {
 	_THREAD_SAFE_METHOD_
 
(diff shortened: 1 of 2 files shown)


platform/linuxbsd/x11/display_server_x11.cpp
```diff
@@ -4227,6 +4227,22 @@ bool DisplayServerX11::mouse_process_popups() {
...
 	return closed;
 }
+bool DisplayServerX11::_window_focus_check() {
+    
+	Window focused_window;
+    int focus_ret_state;
+    XGetInputFocus(x11_display,&focused_window,&focus_ret_state);
+    
+    bool has_focus = false;
+    for (KeyValue<int, DisplayServerX11::WindowData> wid : windows) {
```
AThousandShips, line 4237:
```suggestion
    for (const KeyValue<int, DisplayServerX11::WindowData> &wid : windows) {
```

platform/linuxbsd/x11/display_server_x11.cpp
```diff
@@ -4227,6 +4227,21 @@ bool DisplayServerX11::mouse_process_popups() {
 	}
 	return closed;
 }
+bool DisplayServerX11::_window_focus_check() {
```
AThousandShips, line 4230:
```suggestion

bool DisplayServerX11::_window_focus_check() {
```

platform/linuxbsd/x11/display_server_x11.cpp
dsnopek: Thanks, fixed!

platform/linuxbsd/x11/display_server_x11.cpp
dsnopek: This one too

platform/linuxbsd/x11/display_server_x11.cpp
```diff
@@ -4719,7 +4735,7 @@ void DisplayServerX11::process_events() {
 					// Ensure window focus on click.
 					// RevertToPointerRoot is used to make sure we don't lose all focus in case
 					// a subwindow and its parent are both destroyed.
-					if (!wd.no_focus && !wd.is_popup) {
+					if (!wd.no_focus && !wd.is_popup && _window_focus_check()) {
```
bruvzg, line 4738: This one probably should not be changed, since ButtonPress should always grab focus, the rest looks reasonable.

bruvzg (commented):
This looks reasonable to me (have not tested it so far). I guess it can be combined with #86671, to only grab re-focus if it's not already focused and some other window of the same app have focus (with exception for ButtonPress which should only check the current window focus).

bruvzg (commented):
This looks reasonable to me (have not tested it so far). I guess it can be combined with #86671, to only grab re-focus if it's not already focused and some other window of the same app have focus (with exception for ButtonPress which should only check the current window focus).

platform/linuxbsd/x11/display_server_x11.cpp
dsnopek: Thanks! Updated in latest push

dsnopek:
Thanks for the review!

> This looks reasonable to me (have not tested it so far). I guess it can be combined with https://github.com/godotengine/godot/pull/86671, to only grab re-focus if it's not already focused and some other window of the same app have focus

Yes, I don't think the two PRs should conflict functionality-wise (there may end up being a Git conflict after one of them is merged, because of the proximity of the changed lines) although logically they are working in kind of inverse directions of each other.

Personally, when I test this PR on i3, it doesn't fix any issues for me, but it also doesn't appear to break anything.

Calinou (approved):
Tested locally (Fedora 39 KDE with X11, rebased against https://github.com/godotengine/godot/commit/96296e476ffda2d4c36680b2271e9eeff0c3b9f8), it works as expected.

This fixes a longstanding issue in Godot 4 where focus is stolen while the editor loads (or when a project starts), which causes some keypresses to be released if you're currently focusing on another window. I initially thought this was due to use of `Window.request_attention()` somewhere, but it's not – it still occurs if you make `Window.request_attention()` a no-op in the engine source code.

A *perfect* example of this is when you try to run diagonally in a game while the Godot editor loads, but one of the movement keypresses will be released after the editor is done loading. This PR fixes this issue completely. Previously, I could reproduce this pretty much every time, but I've tried 10+ times with this PR and couldn't reproduce the issue once.

I tested the editor quickly and couldn't notice any regressions in editor popup behavior.

### Before

*On both videos, I'm not releasing any keys while running.*

https://github.com/godotengine/godot/assets/180032/c8dcc8c8-9126-463f-88bf-4b56ebf83d7b

### After *(this PR)*

https://github.com/godotengine/godot/assets/180032/3716708e-8148-48c6-aa19-6e6668782c6b

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

ckdarby:
@YuriSizov God's work cherry picking this commit before 4.3. Thank you!!

PR #86444, opened 2023-12-22
Add static keyword to parent function sig in error

TurqW:
If the "static" keyword is omitted on a subclass function, this is (correctly) interpreted as breaking polymorphism but the error message displayed doesn't include it, and so it can be more difficult than necessary to figure out what's going wrong.

This change fixes that.


AThousandShips:
I'm not so sure about this, static functions aren't necessarily overridden, I'm not sure this path makes sense here, should this error really appear for a static method?

TurqW:
I'm not sure if that's the case or not, but it currently *does* appear, thanks to line [1689](https://github.com/godotengine/godot/blob/9d1cbab1c432b6f1d66ec939445bec68b6af519e/modules/gdscript/gdscript_analyzer.cpp#L1689).


image

PR #76025, opened 2023-04-13
Expose dialog parent-and-popup logic to the API

YuriSizov:
This was introduced in a hacky way before to fix immediate issues with some of the editor windows (see https://github.com/godotengine/godot/pull/73365). But I think we should have this feature generally accessible, so we can address similar issues in other editor windows, and so plugin developers could use it too. In fact, this is probably useful to user projects as well, if they support multiple windows.

So this removes the old hacky class and moves the necessary functionality to other places.

- `EditorInterface` receives new `popup_dialog_*` methods to automatically reparent a window to the editor UI before popping it up.
- `Window` receives `set_unparent_when_invisible`, a new scripting-only method to enable the "unparent this window when visibility changes to not visible" logic. It's scripting-only because using it requires you to have a script, and enabling it by accident can lead to orphan nodes. `Window` also receives new `popup_exclusive_*` methods, which power `EditorInterface` methods secretly.
- `Node` receives `get_last_exclusive_window`, which tries to find the top-most exclusive child of the owner window, or returns the parent window (or nothing, if there is no window); this can be exposed as a part of `Window` as well, but being a part of `Node` makes it easier to use, as it can handle the owner window internally and return a clean result.

-----

Marking as a draft as this is based on the `EditorInterface` refactoring from https://github.com/godotengine/godot/pull/75694.

PR #86446, opened 2023-12-22
Implement a `transient_to_focused` Window mode

reduz:
This intends to be the correct way to handle non-child windows becoming covered by the current window when becoming focused. Enabling this property on select windows, they will become transient to the currently focused one when becoming visible.

This deprecates the "unparent_when_invisible" function introduced by #76025.

### DIFF
--- a/doc/classes/Window.xml
+++ b/doc/classes/Window.xml
@@ -668,6 +668,9 @@
 			If [code]true[/code], the [Window] is transient, i.e. it's considered a child of another [Window]. The transient window will be destroyed with its transient parent and will return focus to their parent when closed. The transient window is displayed on top of a non-exclusive full-screen parent window. Transient windows can't enter full-screen mode.
 			Note that behavior might be different depending on the platform.
 		</member>
+		<member name="transient_to_focused" type="bool" setter="set_transient_to_focused" getter="is_transient_to_focused" default="false">
+			If [code]true[/code], and the [Window] is [member transient], this window will (at the time of becoming visible) become transient to the currently focused window instead of the immediate parent window in the hierarchy. Note that the transient parent is assigned at the time this window becomes visible, so changing it afterwards has no effect until re-shown.
+		</member>
 		<member name="transparent" type="bool" setter="set_flag" getter="get_flag" default="false">
 			If [code]true[/code], the [Window]'s background can be transparent. This is best used with embedded windows.
 			[b]Note:[/b] Transparency support is implemented on Linux, macOS and Windows, but availability might vary depending on GPU driver, display manager, and compositor capabilities.
--- a/platform/linuxbsd/x11/display_server_x11.cpp
+++ b/platform/linuxbsd/x11/display_server_x11.cpp
@@ -2891,10 +2891,15 @@ void DisplayServerX11::window_move_to_foreground(WindowID p_window) {
 	XFlush(x11_display);
 }
 
+DisplayServerX11::WindowID DisplayServerX11::get_focused_window() const {
+	return last_focused_window;
+}
+
 bool DisplayServerX11::window_is_focused(WindowID p_window) const {
 	_THREAD_SAFE_METHOD_
 
 	ERR_FAIL_COND_V(!windows.has(p_window), false);
+
 	const WindowData &wd = windows[p_window];
 
 	return wd.focused;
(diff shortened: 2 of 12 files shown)


scene/main/window.cpp
```diff
@@ -918,7 +935,7 @@ void Window::set_transient(bool p_transient) {
 		return;
 	}
 
-	if (transient) {
+	if (transient && !transient_to_focused) {
 		_make_transient();
 	} else {
 		_clear_transient();
```
YuriSizov, line 941: This will call `_clear_transient()` when you have `transient_to_focused` set to `true` and then set `transient` to `true`, I think.

scene/main/window.cpp
```diff
@@ -864,18 +872,27 @@ void Window::_make_transient() {
...
-			break;
-		}
-		if (!vp->get_parent()) {
-			break;
-		}
 
-		vp = vp->get_parent()->get_viewport();
+	if (!is_embedded() && transient_to_focused) {
+		DisplayServer::WindowID focused_window_id = DisplayServer::get_singleton()->get_focused_window();
+		window = Object::cast_to<Window>(ObjectDB::get_instance(DisplayServer::get_singleton()->window_get_attached_instance_id(focused_window_id)));
```
YuriSizov, line 880: We should probably check for `DisplayServer::INVALID_WINDOW_ID` before trying to get the window.

scene/main/window.cpp
```diff
@@ -930,6 +947,20 @@ bool Window::is_transient() const {
 	return transient;
 }
 
+void Window::set_transient_to_focused(bool p_transient_to_focused) {
+	ERR_MAIN_THREAD_GUARD;
+	if (transient_to_focused == p_transient_to_focused) {
```
AThousandShips, line 952: Early return unnecessary unless there's something triggered by it

YuriSizov: I think it keeps it in style of other methods, so it's fine. But yes, not strictly necessary.

YuriSizov (approved):
Left a couple of comments based on code review, which need to be addressed. But in testing (rebased on master and with `clear_transient` comment fixed) everything seems to be in working order. Changes are clear and make sense.

scene/main/window.cpp
bruvzg: Not sure if it can be invalid in any situations it is called, but probably worth adding a check just in case.

scene/main/window.cpp
bruvzg: Indeed, it should call `_clear_transient` only if `transient` is unset, so should be:
```c++
if (transient) {
	if (!transient_to_focused) {
		_make_transient();
	}
} else {
	_clear_transient();
}
```

doc/classes/Window.xml
```diff
@@ -668,6 +668,9 @@
 			If [code]true[/code], the [Window] is transient, i.e. it's considered a child of another [Window]. The transient window will be destroyed with its transient parent and will return focus to their parent when closed. The transient window is displayed on top of a non-exclusive full-screen parent window. Transient windows can't enter full-screen mode.
 			Note that behavior might be different depending on the platform.
 		</member>
+		<member name="transient_to_focused" type="bool" setter="set_transient_to_focused" getter="is_transient_to_focused" default="false">
+			If [code]true[/code], and the [Window] is [member transient], this window will (at the time of becoming visible) become transient to the currently focused window instead of the immediate parent window in the hierarchy.
```
bruvzg, line 672: It's probably worth mentioning that it won't change transient parent if it's already visible.

bruvzg (approved):

scene/main/window.cpp
bruvzg: I would remove it since it's clearly not doing anything.

platform/linuxbsd/x11/display_server_x11.cpp
```diff
@@ -2891,10 +2891,15 @@ void DisplayServerX11::window_move_to_foreground(WindowID p_window) {
 	XFlush(x11_display);
 }
 
+DisplayServerX11::WindowID DisplayServerX11::get_focused_window() const {
+	return last_focused_window;
```
Sauermann, line 2895: Is it intentional, that in this case `DisplayServerX11::_get_focused_window_or_popup()` is not used?

scene/main/window.cpp
reduz: Great catch, thanks!

platform/linuxbsd/x11/display_server_x11.cpp
reduz: its kinda not useful for popups, and popups can complicate things because this code could be called from a popup that is about to be closed, hence you get the wrong window.

reduz:
Okay, implemented suggested changes.

YuriSizov (approved):

PR #86447, opened 2023-12-22
Improve performance of the 'Create New Node' dialog

### LINKED ISSUE #27333: Script classes make the CreateDialog search filter prohibitively slow
**Godot version:**
Godot 3.1

**Issue description:**
The CreateDialog is *incredibly* laggy (like, prohibitively laggy) when you have many script classes defined in your project and you try to type into the search filter.

**Steps to reproduce:**
1. Create a new project.
2. Download [godot-next](https://github.com/willnationsdev/godot-next) from GitHub, extract the .zip, and add the plugin.
3. Turn the plugin on.
4. Open the CreateDialog and try typing in the search field.

It will lag incredibly long.

**Suggested solution:**
I believe this is happening because of runtime checks that the CreateDialog has to perform to track inheritance relationships between script classes, as well as a host of additional logic that allows it to handle engine types, script classes, and custom types. Every time the `text_changed` signal of the TextEdit fires (so every add or delete of a character), it *rebuilds the entire tree*, even though none of the data is really changing.

Ideally, we should instead be able to not change any of the TreeItem structure, but toggle the display of certain ones. I'm not sure yet if that is a feature, but if not, we can add it, only update the visualization during `text_changed`, but rebuild the tree only when script classes have been updated by the EditorFileSystem. This will guarantee fast opening of the CreateDialog, and fast updating of the structure as users type text, resulting in a much more responsive tool.

*Bugsquad edit (keywords for easier searching):* named class, class_name

### LINKED ISSUE #34205: Slow Create Resource Dialog in a project with a lot of class_names
**Godot version:**
Bug Introduced in 3.1.2 stable

Windows 10

**Issue description:**
When the user try to set a resource in inspector, the editor takes ages to show the context menu, this happens in projects with a lot of class_names, so should be related to https://github.com/godotengine/godot/pull/27566
@willnationsdev 
Here is a video of the problem:
https://gfycat.com/RegalNeatIrishdraughthorse
In the video the loading takes almost a full dancing pikachu loop.

This issue was introduced in godot stable 3.1.2, in 3.1.1 only happened with the add node button.

**Steps to reproduce:**
Create a project with multiple class_names (100+)
Try to set a resource in the inspector by clicking it (drag a resource on top don't reproduce the problem)

Maran23:
Fixes: https://github.com/godotengine/godot/issues/27333
Fixes: https://github.com/godotengine/godot/issues/34205

- Merged 3 for loops into 1 and save `Vector` allocation
    - We only iterate once over the `type_list` and add the matching types to the tree
- Use `get_instance_base_type()` or `get_global_name()` instead of `get_language()->get_global_class_name()` for performance considerations
    - Behaviour is the same as before, but this is a simple lookup instead
- Use `StringName` where appropriate
    - As the dialog consist of all Godot nodes, like `Node`, `Node2D`, `Sprite2D`, ... so using `StringName` makes sense here (and was also in place before, just not consistently)
- Deduplicated the code in `_add_type`
    - The two if branches were nearly the same and could be merged easily. They benefit from the optimzation mentioned in the second point

This also fixes a bug where scripts/resources with compile errors were displayed in the dialog although they can not be used, resulting in errors (and even a crash) when searching in the dialog (`Broken.gd` is a file with compile errors):
```
Cannot get class 'Broken'.
editor\create_dialog.cpp:275 - Condition "inherits.is_empty()" is true.
```

Benchmarks:
--
Dialog with nodes
-
Tested in Debug Build
Test case: 800+ Global Classes

| What | Old | New |
| - | - | - |
| Open Dialog (3x) | 4340 ms, 1505 ms, 1521 ms | 3693 ms, 905 ms, 868 ms |
| Search (Each letter of Node) | 1589 ms, 399 ms, 402 ms, 401 ms | 1036 ms, 260 ms, 258 ms, 247 ms |
| Search (Each letter of Box) | 1196 ms, 1187 ms, 1185 ms | 801 ms, 784 ms, 762 ms |

Dialog with resources
-
Tested in Debug Build
Test case: 800+ Global Resources

| What | Old | New |
| - | - | - |
| Open Dialog (3x) | 2782 ms, 1043 ms, 1026 ms | 2465 ms, 713 ms, 711 ms |
| Search (Each letter of Resource) | 594 ms, 591 ms, 548 ms, 529 ms, 520ms, 520ms, 519ms, 522ms | 441 ms, 427 ms, 391 ms, 381 ms, 373 ms, 369 ms, 366 ms, 369 ms |
| Search (Each letter of Custom) | 578 ms, 541 ms, 525 ms, 524 ms, 523 ms, 521 ms | 426 ms, 385 ms, 367 ms, 364 ms, 365 ms, 369 ms |

Result
-
Performance improved between x1.2 and x1.6. Dialog feels more snappy.

Project:
[ManyFiles.zip](https://github.com/godotengine/godot/files/13754435/ManyFiles.zip)

Note
-
Similar improvements can be done to other places in the editor as far as I can see.

### DIFF
--- a/editor/create_dialog.cpp
+++ b/editor/create_dialog.cpp
@@ -85,7 +85,7 @@ void CreateDialog::_fill_type_list() {
 	EditorData &ed = EditorNode::get_editor_data();
 
 	for (List<StringName>::Element *I = complete_type_list.front(); I; I = I->next()) {
-		String type = I->get();
+		StringName type = I->get();
 		if (!_should_hide_type(type)) {
 			type_list.push_back(type);
 
@@ -118,12 +118,12 @@ bool CreateDialog::_is_class_disabled_by_feature_profile(const StringName &p_cla
 	return !profile.is_null() && profile->is_class_disabled(p_class);
 }
 
-bool CreateDialog::_should_hide_type(const String &p_type) const {
+bool CreateDialog::_should_hide_type(const StringName &p_type) const {
 	if (_is_class_disabled_by_feature_profile(p_type)) {
 		return true;
 	}
 
-	if (is_base_type_node && p_type.begins_with("Editor")) {
+	if (is_base_type_node && p_type.operator String().begins_with("Editor")) {
 		return true; // Do not show editor nodes.
 	}
 
@@ -146,11 +146,11 @@ bool CreateDialog::_should_hide_type(const String &p_type) const {
 			}
 		}
 	} else {
-		if (!EditorNode::get_editor_data().script_class_is_parent(p_type, base_type)) {
-			return true; // Wrong inheritance.
-		}
 		if (!ScriptServer::is_global_class(p_type)) {
 			return true;
+		}
+		if (!EditorNode::get_editor_data().script_class_is_parent(p_type, base_type)) {
+			return true; // Wrong inheritance.
 		}
 
 		StringName native_type = ScriptServer::get_global_class_native_base(p_type);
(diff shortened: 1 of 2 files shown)


editor/create_dialog.cpp
```diff
@@ -182,24 +182,30 @@ void CreateDialog::_update_search() {
 	const String search_text = search_box->get_text();
 	bool empty_search = search_text.is_empty();
 
-	// Filter all candidate results.
-	Vector<String> candidates;
+	float highest_score = 0;
```
Calinou, line 185:
```suggestion
	float highest_score = 0.0f;
```

Calinou (approved):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/ce00392fc35ce639752c710094d87a4311985fb2), it works as expected.

With an optimized editor build, I can't notice too much a difference before and after this PR.

However, with a debug editor build, the difference is much more noticeable. The Create New Node dialog opens significantly faster with this PR and it doesn't freeze for as long when typing.

The difference should be even more noticeable on slower CPUs.

<details open>

- **OS:** Linux (Fedora 39)
</details>

editor/create_dialog.cpp
Maran23: Understood and done.

editor/create_dialog.h
```diff
@@ -68,10 +68,9 @@ class CreateDialog : public ConfirmationDialog {
 	HashSet<StringName> type_blacklist;
 
 	void _update_search();
-	bool _should_hide_type(const String &p_type) const;
-	void _add_type(const String &p_type, const TypeCategory p_type_category);
-	void _configure_search_option_item(TreeItem *r_item, const String &p_type, const TypeCategory p_type_category);
-	String _top_result(const Vector<String> p_candidates, const String &p_search_text) const;
+	bool _should_hide_type(const StringName &p_type) const;
+	void _add_type(const StringName &p_type, const TypeCategory p_type_category);
+	void _configure_search_option_item(TreeItem *r_item, const StringName &p_type, const TypeCategory p_type_category);
```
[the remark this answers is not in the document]
Maran23: done!

akien-mga:
Thanks!

Issue #81615, opened 2023-09-13
Using other scripts by their class_name in an import plugin causes import to fail


tcoxon:
### Godot version

4.1.1-stable

### System information

Godot v4.1.1.stable (fabd7c89d) - Arch Linux #1 SMP PREEMPT_DYNAMIC Sat, 10 Jun 2023 00:35:35 +0000 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 970 (nvidia; 530.41.03) - AMD Ryzen 5 2600 Six-Core Processor (12 Threads)

### Issue description

I have been encountering a strange issue where builds of my project created and exported from a fresh clone of the repo are missing their translations. I've narrowed this down to a weird quirk of import plugin scripts (my project uses a custom editor plugin to import translations). It seems as though global script class names don't exist during the initial import.

The reproducer project contains a CSV file `translations/example.csv` and its corresponding .import set up to be imported using the "Translation Plus" plugin.

If the two lines at the top of `addons/translation_plus/import_plugin.gd` that shadow the global class names `TranslationSet` and `TranslationPlus` are commented out, and the project's hidden .godot directory is removed, then the next time the project is opened, importing example.csv fails.

### Steps to reproduce

1. Download the attached reproducer.
2. Open the project in Godot and notice `translation/example.csv` is imported as `Translation Plus`.
3. Open the script `addons/translation_plus/import_plugin.gd`
4. Comment the two lines (these lines should not be necessary as TranslationSet and TranslationPlus are global script class names):
```GDScript
const TranslationSet = preload("translation_set.gd")
const TranslationPlus = preload("translation_plus.gd")
```

5. Close the project
6. Erase the .godot directory so that the editor does a clean reimport next time it opens the project.
7. Open the project and observe that example.csv fails to import and instead Godot imports it as `CSV Translation`. Errors are logged:

```
res://addons/translation_plus/import_plugin.gd:84 - Parse Error: Identifier "TranslationSet" not declared in the current scope.
  res://addons/translation_plus/import_plugin.gd:84 - Parse Error: Cannot infer the type of "xset" variable because the value doesn't have a set type.
  res://addons/translation_plus/import_plugin.gd:92 - Parse Error: Identifier "TranslationPlus" not declared in the current scope.
  res://addons/translation_plus/import_plugin.gd:92 - Parse Error: Cannot infer the type of "t" variable because the value doesn't have a set type.
[...]

PR #70668, opened 2022-12-28
Retry loading addons after filesystem scan

KoBeWi:
Fixes #68615 Probably. I can't test due to #70667

PR #86453, opened 2023-12-22
Defer addons loading after the first file scan/import

### LINKED ISSUE #23937: Plugin is broken on install and requires to restart Godot due to resources failing to load
Godot 3.1 alpha2

I copied a working addon into a Godot 3.1 project, but it was completely broken and produced a lot of errors. I restarted the editor, the errors went away and the plugin worked.
The same plugin in Godot 3.0.6 worked fine without errors, even on first install.
I tried to reduce the size of the project to narrow it down, but it also changed the results and became not reproducible. It always happens if I test with the whole plugin. It's a small plugin though, with two scenes.

Reproduction steps:

1) Download this project: [AddonBrokenWhenFirstAdded.zip](https://github.com/godotengine/godot/files/2611610/AddonBrokenWhenFirstAdded.zip)

2) Open it in Godot. It's an empty project as if you had created it from scratch, and should load fine.

3) Go to the system file explorer and unzip the `addons` archive to install the addon (works too if you created an empty project from scratch).

4) Focus back the editor, see it detects and imports the addon's resources. However, it prints a lot of errors which didn't happen in Godot 3.0.6. Most notably this error, which causes a lot of others due to the script not being compiled:
```
SCRIPT ERROR: GDScript::reload: Parse Error: Can't preload resource at path: res://addons/zylann.channel_packer/icons/empty.png
          At: res://addons/zylann.channel_packer/input_texture.gd:14
```
This resource definitely exists.

5) Go to project settings and activate the plugin: it makes a new menu appear on top-right, use it, notice the dialog it opens is completely broken:
image
Note: the layout problem is caused by https://github.com/godotengine/godot/issues/17626

6) Close the project, reopen it: notice everything went back to normal.
image

### LINKED ISSUE #77037: Autoloaded singleton cannot be accessed at early phase during editor plugin start-up during editor start-up
### Godot version

v4.0.1.stable.official [cacf49999]

### System information

Windows 11

### Issue description

I have a editor plug-in. In "_enter_tree" it adds and autoload singleton + adds a Control to a dock.
The Control would come from the singleton.

The autoloaded singleton is not yet accessible just right after autoload, when:
- Godot Editor (project opening) is starting-up, and plug-in is already loaded in project settings
No matter if autoload is already in the project setting, or added fo the first time by the plug-in
(Project open = Opening project.godot directly, not from project manager, but I assume it doesn't matter)

Not observed, when:
- Editor is already running and the disabled plug-in is enabled.
Error free even if singleton is removed from the autoload before enabling the plug-in.

```
@tool
class_name DebugInfoPlugin extends EditorPlugin

const AUTOLOAD_NAME = "DebugInfo"

var singleton: DebugInfoSingleton

func _enter_tree():
	add_autoload_singleton(AUTOLOAD_NAME, "res://addons/debug_info_plugin/DebugInfoSingleton.gd")
	singleton = get_node("/root/" + AUTOLOAD_NAME)
	add_control_to_dock(DOCK_SLOT_LEFT_BR, singleton._get_control())
```
Errors in console (when singleton is already added to the autoload during start-up, i.e. in previous session):
```
Node not found: "/root/DebugInfo" (absolute path attempted from "/root/@@16580/@@16699").
res://addons/debug_info_plugin/DebugInfoPlugin.gd:19 - Attempt to call function '_get_control' in base 'null instance' on a null instance.
```
If plug-in is adding the single to autoload for the first time during start-up (i.e. singleton is not yet in autoload list in Project Settings) an additional error message is in the console BEFORE the above two:
```
Parent node is busy setting up children, `add_child()` failed. Consider using `add_child.call_deferred(child)` instead.
```
This error with `add_child` is present, even, if I only have the `add_autoload_singleton` in the `_enter_tree`.
(Should I make separate ticket for this?)

### Steps to reproduce

Create an editor plug-in similar to the one in description.
Have `add_autoload_singleton` and `get_node` in `_enter_tree`.
Make another `@tool` script, which will be the singleton to load.

Make the scenarios from description:
a) Open project.godot, with plug-in already enabled to project settings -> Error
b) Open project with disabled plug-in and -> No error
Further vary:
c) Singleton is already enabled in project setting's autoload before a or b
d) Singleton is removed from project setting's autoload before a or b

N/A

adamscott:
This fixes the issue where the active plugins that are enabled refer to not yet imported textures or files.

# What it does
It defers the load of plugin scripts after the first file scan. So, no plugin will be loaded before all the files would be imported.

What happened before is that Godot would complain if loaded addons would try to load not yet imported files, but it was only a race condition because the engine would have been done so a few seconds (if not milliseconds) later.

These errors were common for projects were the addon would be installed before opening Godot. The only fix was to restart Godot, which is unfortunate.

# Fixes

Fixes godotengine/godot#77037
Fixes #23937

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -704,24 +704,6 @@ void EditorNode::_notification(int p_what) {
 		} break;
 
 		case NOTIFICATION_READY: {
-			{
-				started_timestamp = Time::get_singleton()->get_unix_time_from_system();
-				_initializing_plugins = true;
-				Vector<String> addons;
-				if (ProjectSettings::get_singleton()->has_setting("editor_plugins/enabled")) {
-					addons = GLOBAL_GET("editor_plugins/enabled");
-				}
-
-				for (int i = 0; i < addons.size(); i++) {
-					set_addon_plugin_enabled(addons[i], true);
-				}
-				_initializing_plugins = false;
-
-				if (!pending_addons.is_empty()) {
-					EditorFileSystem::get_singleton()->connect("script_classes_updated", callable_mp(this, &EditorNode::_enable_pending_addons));
-				}
-			}
-
 			RenderingServer::get_singleton()->viewport_set_disable_2d(get_scene_root()->get_viewport_rid(), true);
 			RenderingServer::get_singleton()->viewport_set_environment_mode(get_viewport()->get_viewport_rid(), RenderingServer::VIEWPORT_ENVIRONMENT_DISABLED);
 
@@ -739,6 +721,8 @@ void EditorNode::_notification(int p_what) {
 			// Set up a theme context for the 2D preview viewport using the stored preview theme.
 			CanvasItemEditor::ThemePreviewMode theme_preview_mode = (CanvasItemEditor::ThemePreviewMode)(int)EditorSettings::get_singleton()->get_project_metadata("2d_editor", "theme_preview", CanvasItemEditor::THEME_PREVIEW_PROJECT);
 			update_preview_themes(theme_preview_mode);
+
+			started_timestamp = Time::get_singleton()->get_unix_time_from_system();
 
 			/* DO NOT LOAD SCENES HERE, WAIT FOR FILE SCANNING AND REIMPORT TO COMPLETE */
 		} break;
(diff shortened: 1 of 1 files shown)


editor/editor_node.cpp
```diff
@@ -698,24 +698,6 @@ void EditorNode::_notification(int p_what) {
...
-				for (int i = 0; i < addons.size(); i++) {
-					set_addon_plugin_enabled(addons[i], true);
-				}
-				_initializing_plugins = false;
-
-				if (!pending_addons.is_empty()) {
-					EditorFileSystem::get_singleton()->connect("script_classes_updated", callable_mp(this, &EditorNode::_enable_pending_addons));
-				}
-			}
-
```
adamscott, line 718: I'm currently thinking if we should still import the addons here when were not in the "editor". We cannot import resources outside the editor, so we should import plugins faster for games.

editor/editor_node.cpp
```diff
@@ -1042,34 +1026,54 @@ void EditorNode::_resources_reimported(const Vector<String> &p_resources) {
 }
 
 void EditorNode::_sources_changed(bool p_exist) {
-	if (waiting_for_first_scan) {
-		waiting_for_first_scan = false;
+	if (!waiting_for_first_scan) {
+		return;
+	}
```
adamscott, line 1031: Here, I just early exited instead of wrapping the whole contents of the function inside an if block.

KoBeWi:
Looks similar to what #70668 tried to do.

editor/editor_node.cpp
KoBeWi: Not sure what do you mean by "editor", but this code never executes at runtime.

editor/editor_node.cpp
```diff
@@ -1042,34 +1026,54 @@ void EditorNode::_resources_reimported(const Vector<String> &p_resources) {
...
+			set_addon_plugin_enabled(addons[i], true);
 		}
+		_initializing_plugins = false;
 
-		// Start preview thread now that it's safe.
-		if (!singleton->cmdline_export_mode) {
-			EditorResourcePreview::get_singleton()->start();
+		if (!pending_addons.is_empty()) {
+			EditorFileSystem::get_singleton()->connect("script_classes_updated", callable_mp(this, &EditorNode::_enable_pending_addons));
 		}
```
KoBeWi, line 1077: This part and related code can be removed now (basically reverting #70668), as it's no longer necessary.

KoBeWi (approved):
IIRC the original reason why plugins were loaded before scanning was that there can be EditorImportPlugin (or similar) among them, which is relevant for the scan. However I tested some custom EditorImportPlugin and it seems to import stuff as soon as it's enabled.

But that was a simple case; maybe a problem arises for more advanced import plugins that customize assets which are already handled. Needs a check from @godotengine/import

adamscott:
> IIRC the original reason why plugins were loaded before scanning was that there can be EditorImportPlugin (or similar) among them, which is relevant for the scan. However I tested some custom EditorImportPlugin and it seems to import stuff as soon as it's enabled.

Where I think it would fail is when you clear the `.godot` directory, the first load import will not include the addons `EditorImportPlugin`. But again, it would have a chance to fail to do so because the plugin could have preloaded a (yet) non-imported file beforehand.

Zylann:
What if a user clones a project with import plugins in it? It sounds like it will now fail to load properly, as importing will happen before the import plugins are added.

akien-mga (dismissed):
This is just moving code around to run later / on each source change, so it seems good to me.

*Edit:* Just saw Zylann's question, should be answered before merging.

YuriSizov:
> IIRC the original reason why plugins were loaded before scanning was that there can be EditorImportPlugin (or similar) among them, which is relevant for the scan. However I tested some custom EditorImportPlugin and it seems to import stuff as soon as it's enabled.

Well, that is still the main problem. Resources may need a plugin to be loaded/imported, and plugins may need a resource to work, just like Zylann explains.

Just moving code around is not the proper solution here. We should instead add more steps to the plugin initialization instead. Either some hard defined hooks, similar to how extensions are initialized with different "levels", or an automatic solution which can detect failed imports and retry everything that fails until all issues are resolved (which seems like a complicated task, so probably not).

On top of all, I'd wait with modifications to the editor load order until I can do the refactor. There is no point making ad hoc patches for something that is going to be reworked in the same version.

YuriSizov (changes requested):
See above.

KoBeWi:
Found some project to test: #81615
Opening the project (after you follow the steps in the OP) will cause `example.csv` fallback to CSV Translation due to missing importer. Though it happens regardless if you use this PR or not. With this PR you are at least able to reimport without restarting the editor.

noidexe:
FYI tool scripts set as autoloads exhibit the same issue. If they preload any resource that has not yet been imported parsing and execution will fail

adamscott:
> FYI tool scripts set as autoloads exhibit the same issue. If they preload any resource that has not yet been imported parsing and execution will fail

You mean that this PR fixes this issue?

KoBeWi:
Seems like #92667 fixes a very similar (if not the same) issue.

akien-mga:
@Hilderin Could you check if #92667 supersedes this, and also fixes #77037 and #23937?

Hilderin:
I tested #77037 and #23937 with the PR #92667.

For #77037: the PR does not fix the problem and I don't think the original problem was caused by the order in which the plugins are loaded. I created an alternative solution to this PR with less modifications: #94802

For #23937: the PR does fix the problem of having a preload resource in tool script when starting the editor without the .godot folder. The PR #92303 already fixes the other problems linked to global class names not being registered correctly on startup with the .godot folder missing or when adding the addon files while the editor is running.

Also, I'm worry about this PR which seems to change the order of loading the plugins. It will creates problems with custom loaders that could be registered in plugins.

In conclusion, I suggest #92667 and #94802 to fix these issues.

akien-mga:
Superseded by #92667 and #94802.

PR #86454, opened 2023-12-22
C#: Upgrade TFM for `net472` and some `netstandard2.0` projects

Repiteo:
Inspired by raulsntos's post [here](https://github.com/godotengine/godot/pull/86375#issuecomment-1867963200), this PR upgrades a handful of `.csproj` files to target `net6.0` as their framework. 3 were able to be upgraded by tweaking the project file, while one needed a minor file refactor. They all build successfully, but I'm unsure of how to do a more thorough test for possible regressions beyond what's already setup in the build system checks

### DIFF
--- a/modules/mono/editor/GodotTools/GodotTools.BuildLogger/GodotTools.BuildLogger.csproj
+++ b/modules/mono/editor/GodotTools/GodotTools.BuildLogger/GodotTools.BuildLogger.csproj
@@ -1,8 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <ProjectGuid>{6CE9A984-37B1-4F8A-8FE9-609F05F071B3}</ProjectGuid>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <LangVersion>7.2</LangVersion>
+    <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>10</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Build.Framework" Version="15.1.548" ExcludeAssets="runtime" />
--- a/modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/GodotTools.OpenVisualStudio.csproj
+++ b/modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/GodotTools.OpenVisualStudio.csproj
@@ -1,12 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
-    <PropertyGroup>
-        <ProjectGuid>{EAFFF236-FA96-4A4D-BD23-0E51EF988277}</ProjectGuid>
-        <OutputType>Exe</OutputType>
-        <TargetFramework>net472</TargetFramework>
-        <LangVersion>7.2</LangVersion>
-    </PropertyGroup>
-    <ItemGroup>
-        <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
-        <PackageReference Include="EnvDTE" Version="8.0.2" />
-    </ItemGroup>
+  <PropertyGroup>
+    <ProjectGuid>{EAFFF236-FA96-4A4D-BD23-0E51EF988277}</ProjectGuid>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0-windows</TargetFramework>
+    <LangVersion>10</LangVersion>
+  </PropertyGroup>
+  <PropertyGroup Condition="Exists('$(SolutionDir)/../../../../bin/GodotSharp/Api/Debug/GodotSharp.dll') And ('$(GodotPlatform)' == 'windows' Or ('$(GodotPlatform)' == '' And '$(OS)' == 'Windows_NT'))">
+    <OutputPath>$(SolutionDir)/../../../../bin/GodotSharp/Tools</OutputPath>
+    <AppendTargetFrameworkToOutputPath>False</AppendTargetFrameworkToOutputPath>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="EnvDTE" Version="17.8.37221" />
+  </ItemGroup>
 </Project>
--- a/modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/Program.cs
+++ b/modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/Program.cs
@@ -129,7 +129,7 @@ namespace GodotTools.OpenVisualStudio
             {
                 var mainWindow = dte.MainWindow;
                 mainWindow.Activate();
-                SetForegroundWindow(new IntPtr(mainWindow.HWnd));
+                SetForegroundWindow(mainWindow.HWnd);
 
                 MessageFilter.Revoke();
             }
--- a/modules/mono/editor/GodotTools/GodotTools.IdeMessaging.CLI/GodotTools.IdeMessaging.CLI.csproj
+++ b/modules/mono/editor/GodotTools/GodotTools.IdeMessaging.CLI/GodotTools.IdeMessaging.CLI.csproj
@@ -2,12 +2,9 @@
   <PropertyGroup>
     <ProjectGuid>{B06C2951-C8E3-4F28-80B2-717CF327EB19}</ProjectGuid>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net472</TargetFramework>
-    <LangVersion>7.2</LangVersion>
+    <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>10</LangVersion>
   </PropertyGroup>
-  <ItemGroup>
-    <Reference Include="System" />
-  </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\GodotTools.IdeMessaging\GodotTools.IdeMessaging.csproj" />
   </ItemGroup>
--- a/modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj
+++ b/modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj
@@ -49,7 +49,5 @@
     <ProjectReference Include="..\GodotTools.IdeMessaging\GodotTools.IdeMessaging.csproj" />
     <ProjectReference Include="..\GodotTools.ProjectEditor\GodotTools.ProjectEditor.csproj" />
     <ProjectReference Include="..\GodotTools.Core\GodotTools.Core.csproj" />
-    <!-- Include it if this is an SCons build targeting Windows, or if it's not an SCons build but we're on Windows -->
-    <ProjectReference Include="..\GodotTools.OpenVisualStudio\GodotTools.OpenVisualStudio.csproj" Condition=" '$(GodotPlatform)' == 'windows' Or ( '$(GodotPlatform)' == '' And '$(OS)' == 'Windows_NT' ) " />
   </ItemGroup>
 </Project>


modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/Program.cs
```diff
@@ -2,11 +2,13 @@
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.ComTypes;
+using System.Runtime.Versioning;
 using System.Text.RegularExpressions;
 using EnvDTE;
 
 namespace GodotTools.OpenVisualStudio
 {
+    [SupportedOSPlatform("windows")]
```
raulsntos, line 11: Since this project is only meant to be run on Windows, shouldn't this project target `net6.0-windows`? And would that mean we don't have to use the `SupportedOSPlatform` attribute?

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/GodotTools.IdeMessaging.csproj
```diff
@@ -1,10 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <ProjectGuid>{92600954-25F0-4291-8E11-1FEE9FC4BE20}</ProjectGuid>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <LangVersion>7.2</LangVersion>
+    <TargetFramework>net6.0</TargetFramework>
```
raulsntos, line 4: I think the JetBrains team is using this library in their [Rider extension](https://plugins.jetbrains.com/plugin/13882-godot-support), so we need to make sure this won't break them (also the other IDE plugins but those are less maintained).

For example, our VSCode extension still uses .NET Framework:

https://github.com/godotengine/godot-csharp-vscode/blob/5cb26cc7fa4059d2aaacc5fdf46a7008d6c40951/GodotDebugSession/GodotDebugSession.csproj#L12

modules/mono/editor/GodotTools/GodotTools.BuildLogger/GodotTools.BuildLogger.csproj
```diff
@@ -1,8 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <ProjectGuid>{6CE9A984-37B1-4F8A-8FE9-609F05F071B3}</ProjectGuid>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <LangVersion>7.2</LangVersion>
+    <TargetFramework>net6.0</TargetFramework>
```
raulsntos, line 4: To test the GodotTools.BuildLogger project, try building a C# project from Godot and see if the build output is collected and displayed in the MSBuild dock as expected. I don't see why this wouldn't work though.

modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/Program.cs
Repiteo: I assumed so initially, and maybe that is still the proper way to handle it. However, trying that gave me an error about how a `net6.0-windows` project couldn't be targeted properly (I'll have to double-check to see the exact wording)

modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/GodotTools.OpenVisualStudio.csproj
```diff
@@ -1,12 +1,11 @@
...
-        <PackageReference Include="EnvDTE" Version="8.0.2" />
-    </ItemGroup>
+  <PropertyGroup>
+    <ProjectGuid>{EAFFF236-FA96-4A4D-BD23-0E51EF988277}</ProjectGuid>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>10</LangVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="EnvDTE" Version="17.8.37221" />
```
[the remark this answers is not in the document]
Repiteo: I can confirm it opens scripts from the file dock & debugger dock to the correct line

modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/Program.cs
Repiteo: Alright, I've got a solution. I utilized the same conditional for importing the `OpenVisualStudio` the project in the first place to change the `GodotTools` TargetFramework to `net6.0-windows`. As predicted, this removed the need for the `SupportedOSPlatform` attribute

modules/mono/editor/GodotTools/GodotTools.BuildLogger/GodotTools.BuildLogger.csproj
Repiteo: Works as expected!

modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/Program.cs
Repiteo: Double-checked to make sure that everything still opened as expected in Visual Studio with this new format, and it does 👍

raulsntos: Wait, why are we referencing this project in GodotTools? This is an executable project, not a library. Oh, maybe it's just so that `build_assemblies.py` builds it, but it's included in the GodotTools solution which is what we use with the build command so it should be enough.

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/GodotTools.IdeMessaging.csproj
Repiteo: Looking at the VSCode Extension, it's calling to version `1.1.0`, which is already bumped up to `1.1.1` in the main repo & `1.1.2` in this PR. I'm not entirely sure where the extension is even pulling the package from, but so long as the version number is incremented I would imagine that this wouldn't cause any overlapping issues

raulsntos: Yes, the extension could just not upgrade the dependency but ideally they'd stay up to date so they can get bug fixes and new features. The VSCode extension specifically is a bit under-maintained at the moment, so I care a bit more about the needs of the JetBrains Rider extension.

If changing the TFM makes upgrading harder for consumers then that's not a good situation to be in. So, if a lot of our consumers are still on .NET Framework, I think changing the TFM would not be worth it unless it brings significant benefits which I don't think it does here.

Repiteo: Yeah, that's fair. Out of all the projects in this PR, this would be the most likely candidate to remain as `netstandard2.0`. I'd be fine with that, as it'd still mean no more `net472` projects

modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/Program.cs
Repiteo: If it's able to be handled via `build_assemblies.py` alone, then that would simplify matters. I'll double-check if that's possible

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/GodotTools.IdeMessaging.csproj
Repiteo: Reverted this project entirely. This will be better suited for a PR with IDEs in mind explicitly

modules/mono/editor/GodotTools/GodotTools.OpenVisualStudio/Program.cs
Repiteo: Alright, figured it out. The current implementation has `GodotTools.csproj` changing its output path if certain arguments are passed directly, as well as a ton of extra variables that would be a bit cumbersome to try and copy over to other project files. This is in contrast to the `GodotSharp` exports, which have a dedicated function in `build_assemblies.py` to copy over its built files. While initially outside the scope of this PR, it would make sense to update `build_assemblies.py` in order to remove the `GodotTools.csproj` hardcoded elements & executable dependency altogether

Repiteo: Actually, scratch that; `Godot.NET.Sdk` has similarly hardcoded elements. This really is out of scope for this PR after all & would be better suited for a standalone refactoring of `build_assemblies.py` instead.

I'll leave this executable dependency as-is. Even if it's not technically used by GodotTools, it *is* what allows the executable to be copied over

raulsntos: So I was only suggesting to remove the `ProjectReference`:

```diff
diff --git a/modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj b/modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj
index 56ae37b4dd..f44fa9d71a 100644
--- a/modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj
+++ b/modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj
@@ -49,7 +49,5 @@
     <ProjectReference Include="..\GodotTools.IdeMessaging\GodotTools.IdeMessaging.csproj" />
     <ProjectReference Include="..\GodotTools.ProjectEditor\GodotTools.ProjectEditor.csproj" />
     <ProjectReference Include="..\GodotTools.Core\GodotTools.Core.csproj" />
-    <!-- Include it if this is an SCons build targeting Windows, or if it's not an SCons build but we're on Windows -->
-    <ProjectReference Include="..\GodotTools.OpenVisualStudio\GodotTools.OpenVisualStudio.csproj" Condition=" '$(GodotPlatform)' == 'windows' Or ( '$(GodotPlatform)' == '' And '$(OS)' == 'Windows_NT' ) " />
   </ItemGroup>
 </Project>

```

But I guess this reference was why the build outputs the executable `GodotTools.OpenVisualStudio.exe` file to the `bin/GodotSharp/Tools` directory. Copying these properties to `GodotTools.OpenVisualStudio.csproj` should probably be enough:

https://github.com/godotengine/godot/blob/179dfdc8d78b5bd5377dd115af026df58308bdaf/modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj#L8-L11
https://github.com/godotengine/godot/blob/179dfdc8d78b5bd5377dd115af026df58308bdaf/modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj#L22-L28

Although to be honest all we really need is:

```csproj
  <PropertyGroup Condition=" Exists('$(SolutionDir)/../../../../bin/GodotSharp/Api/Debug/GodotSharp.dll') ">
    <OutputPath>$(SolutionDir)/../../../../bin/GodotSharp/Tools</OutputPath>
    <AppendTargetFrameworkToOutputPath>False</AppendTargetFrameworkToOutputPath>
  </PropertyGroup>
```

And I'm not sure we care about the `Condition`. But it does look cleaner with multiple properties like in `GodotTools.csproj`.

Are you saying this is not enough? I don't see the need to modify `build_assemblies.py` but maybe I misunderstood something.

Repiteo: I was more thinking along the lines of wanting to remove any inherently hardcoded elements, but in retrospect that was a bit silly to worry about. Your condensed proposal worked, but I appended the existing conditional to cover our bases, so that now comes out as:
```csproj
  <PropertyGroup Condition="Exists('$(SolutionDir)/../../../../bin/GodotSharp/Api/Debug/GodotSharp.dll') And ('$(GodotPlatform)' == 'windows' Or ('$(GodotPlatform)' == '' And '$(OS)' == 'Windows_NT'))">
    <OutputPath>$(SolutionDir)/../../../../bin/GodotSharp/Tools</OutputPath>
    <AppendTargetFrameworkToOutputPath>False</AppendTargetFrameworkToOutputPath>
  </PropertyGroup>
```
Did a few local tests, and it works like a charm!

raulsntos (approved):
LGTM. Tested building on Linux and nothing seems to break, I trust that you tested on Windows and found no issues either.

akien-mga:
Thanks!

PR #86458, opened 2023-12-22
Fix default NavigationPolygon source geometry group name

### LINKED ISSUE #86414: NavigationRegion2D baking with groups (explicit or recursive) doesn't work as expected at runtime (but it does in the editor)
### Tested versions

- Tested in v4.2.stable.official [46dc27791]

### System information

Godot v4.2.stable - Debian GNU/Linux 11 (bullseye) 11 - X11 - Vulkan (Forward+) - integrated Intel(R) HD Graphics 5500 (BDW GT2) () - Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz (4 Threads)

### Issue description

When using a NavigationRegion2D and attempting to rebake the NavigationPolygon at runtime, it behaves differently than it does in the editor. It will use groups in the editor, it will not use groups at runtime.

In the editor, when using the `Group with Children` or `Group Explicit` modes, it will correctly bake a navigation polygon that includes the obstacles that are part of the specified group.

Here is an example of baking in the editor:

Screenshot_2023-12-22_12-45-31

In the editor, it correctly incorporates the obstacles in this polygon.

However, when then attempting to bake this again through a script at runtime, with the same navigation polygon (and thus the same settings) it seems to ignore all of the nodes in the specified group.

Here is an example of a runtime re-baking:

Screenshot_2023-12-22_12-47-55

In that screenshot, it ignores the obstacles completely (the bug.)

I don't believe the issue is within my re-baking script because it does still work when in `Root node with children` mode:

Screenshot_2023-12-22_12-48-38

Unfortunately, the "root node with children" mode includes all polygon2d nodes, which isn't always desirable and why I have been trying to use the explicit groups instead.

## Note (possible source of the issue?): 
This may be related to an issue I've noticed where the default value for the group, `navigation_polygon_source_geometry_group`,  is sporadically replaced with the similar but different value  of `navigation_polygon_source_group` when re-opening the project or sometimes re-opening or modifying the scene.

For example, in my minimal reproduction project, I am using the default value of `navigation_polygon_source_geometry_group` but when re-opening the project it changes the group value without any user input:

Screenshot_2023-12-22_13-01-23

If it it is also changing/using the incorrect group name at runtime, that might be the root cause of the issue?

I'm happy to provide more context or information if needed. (This is my first time opening an issue so please bear with me.)

### Steps to reproduce

1. Create a NavigationRegion2D
2. Create a NavigationPolygon within that region, and give it an outline
3. Set the NavigationPolygon's `source_geometry_mode` to the `Group Explicit` mode
4. Create Polygon2D or StaticBody2D nodes that are members of the default group `navigation_polygon_source_geometry_group`
5. Bake the NavigationPolygon in the editor (that part should work)
6. Write a script to bake it at runtime, then run the program and do a runtime bake (this should fail to include the group)

### Reproducing from the minimal reproduction project

1. Open the project
2. Reset the main scene's NavigationRegion's NavigationPolygon's source_geometry_group_name to the default value (it keeps changing away from the default value by itself for some reason)
3. Bake the NavigationPolygon in the editor (it should work)
4. Run the Project
5. Press 'R' to run the rebake code at runtime (it will fail to correctly use the items in the group)
6. (Optionally) Press 'M' to change the source_geometry_mode and rebake in the different modes

[Nav2DBakingIssue.zip](https://github.com/godotengine/godot/files/13747543/Nav2DBakingIssue.zip)

(Note, because of the possible bug here, you will need to reset the `source_geometry_group_name` on the NavigationRegion2D's NavigationPolygon to the default value when opening the project - as it seems to reset itself every time to an incorrect value)

smix8:
Fixes default NavigationPolygon source geometry group name.

Fixes https://github.com/godotengine/godot/issues/86414

The default class group name did not match the bind method defaults and documentation defaults.

### DIFF
--- a/scene/resources/navigation_polygon.h
+++ b/scene/resources/navigation_polygon.h
@@ -94,7 +94,7 @@ public:
 	uint32_t parsed_collision_mask = 0xFFFFFFFF;
 
 	SourceGeometryMode source_geometry_mode = SOURCE_GEOMETRY_ROOT_NODE_CHILDREN;
-	StringName source_geometry_group_name = "navigation_polygon_source_group";
+	StringName source_geometry_group_name = "navigation_polygon_source_geometry_group";
 
 	void set_vertices(const Vector<Vector2> &p_vertices);
 	Vector<Vector2> get_vertices() const;


akien-mga (approved):

akien-mga:
Thanks!

PR #86462, opened 2023-12-22
Use ObjectID's instead of node pointers to track scene groups to prevent crash.

### LINKED ISSUE #86427: Closing a scene tab causes a crash
### Tested versions

- Reproducable in: 4.3.dev.1.mono
- Not reproducable in: 4.2.1.stable.mono

### System information

Godot v4.3.dev1.mono - Arch Linux #1 SMP PREEMPT_DYNAMIC Thu, 14 Dec 2023 03:45:42 +0000 - Wayland - Vulkan (Forward+) - dedicated AMD Radeon RX 5700 XT (RADV NAVI10) () - AMD Ryzen 5 2600 Six-Core Processor (12 Threads)

### Issue description

An attempt to close an opened scene via a scene tab close button causes the whole editor to momentarily freeze and crash.

I am using KDE Plasma 5 Wayland session on Arch Linux.

### Steps to reproduce

- Create a new project
- Create and save a new scene
- Close the scene via "X" button in the scene tab

N/A

### LINKED ISSUE #86844: Closing a non empty scene crashes
### Tested versions

-Reproducible in version 4.2.1 dev

### System information

Zorin OS 16.3 x86_64 / Intel i7-2640M / NVIDIA Quadro 2000M / 8 GB RAM

### Issue description

I was closing a scene (WHICH IS NOT EMPTY) when suddenly Godot closed, I run it in terminal and this what it threw:

```
Engine version: Godot Engine v4.3.dev1.official (9d1cbab1c432b6f1d66ec939445bec68b6af519e)
[1] /lib/x86_64-linux-gnu/libc.so.6(+0x43090) [0x7f7682ad0090] (??:0)
[2] /home/goreldeen/Desktop/Godot_v4.3-dev1_linux.x86_64() [0x42857f2] (??:0)
[3] /home/goreldeen/Desktop/Godot_v4.3-dev1_linux.x86_64() [0x1d7d8f8] (??:0)
[4] /home/goreldeen/Desktop/Godot_v4.3-dev1_linux.x86_64() [0x4068c2d] (??:0)
[5] /home/goreldeen/Desktop/Godot_v4.3-dev1_linux.x86_64() [0x3ba27fc] (??:0)
[6] /home/goreldeen/Desktop/Godot_v4.3-dev1_linux.x86_64() [0x207e573] (??:0)
[7] /home/goreldeen/Desktop/Godot_v4.3-dev1_linux.x86_64() [0x4f920e] (??:0)
[8] /home/goreldeen/Desktop/Godot_v4.3-dev1_linux.x86_64() [0x41eff0] (??:0)
[9] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3) [0x7f7682ab1083] (??:0)
[10] /home/goreldeen/Desktop/Godot_v4.3-dev1_linux.x86_64() [0x42af5a] (??:0)
-- END OF BACKTRACE --
```

### Steps to reproduce

1- Open multiple scenes
2- Add content to them
3- Close one of them

[Vampire survivors clone.zip](https://github.com/godotengine/godot/files/13845210/Vampire.survivors.clone.zip)

tdaven:
Since the node is being cleaned up, queuing this to be called deferred can result in the node already being gone by the time it is called, this results in a crash.

After node_removed is called, the node's destructor gets called and its memory deallocated(verified with breakpoints), this results in the deferred call triggering a crash later.

I'm not 100% sure this is the right fix but it seems to work and is certainly the easy fix.  Just not deferring the call in this case prevents the problem since the node is obviously still valid.  I do wonder if the groupeditor caching is doing something a bit odd though and needs more work.  I'm not familiar enough with it to do that at the moment though.

Fixes #86427
Fixes #86844

### DIFF
--- a/editor/groups_editor.cpp
+++ b/editor/groups_editor.cpp
@@ -274,20 +274,22 @@ void GroupsEditor::_update_groups_and_tree() {
 	_update_tree();
 }
 
-void GroupsEditor::_update_scene_groups(Node *p_node) {
-	if (scene_groups_cache.has(p_node)) {
-		scene_groups = scene_groups_cache[p_node];
-		scene_groups_cache.erase(p_node);
+void GroupsEditor::_update_scene_groups(const ObjectID &p_id) {
+	HashMap<ObjectID, HashMap<StringName, bool>>::Iterator I = scene_groups_cache.find(p_id);
+	if (I) {
+		scene_groups = I->value;
+		scene_groups_cache.remove(I);
 	} else {
 		scene_groups = HashMap<StringName, bool>();
 	}
 }
 
-void GroupsEditor::_cache_scene_groups(Node *p_node) {
+void GroupsEditor::_cache_scene_groups(const ObjectID &p_id) {
 	const int edited_scene_count = EditorNode::get_editor_data().get_edited_scene_count();
 	for (int i = 0; i < edited_scene_count; i++) {
-		if (p_node == EditorNode::get_editor_data().get_edited_scene_root(i)) {
-			scene_groups_cache[p_node] = scene_groups_for_caching;
+		Node *edited_scene_root = EditorNode::get_editor_data().get_edited_scene_root(i);
+		if (edited_scene_root && p_id == edited_scene_root->get_instance_id()) {
+			scene_groups_cache[p_id] = scene_groups_for_caching;
 			break;
 		}
 	}
@@ -305,7 +307,7 @@ void GroupsEditor::set_current(Node *p_node) {
 
 	if (scene_tree->get_edited_scene_root() != scene_root_node) {
 		scene_root_node = scene_tree->get_edited_scene_root();
-		_update_scene_groups(scene_root_node);
+		_update_scene_groups(scene_root_node->get_instance_id());
 		_update_groups();
 	}
 
@@ -803,7 +805,7 @@ void GroupsEditor::_bind_methods() {
 void GroupsEditor::_node_removed(Node *p_node) {
 	if (scene_root_node == p_node) {
 		scene_groups_for_caching = scene_groups;
-		callable_mp(this, &GroupsEditor::_cache_scene_groups).call_deferred(p_node);
+		callable_mp(this, &GroupsEditor::_cache_scene_groups).call_deferred(p_node->get_instance_id());
 		scene_root_node = nullptr;
 	}
 
--- a/editor/groups_editor.h
+++ b/editor/groups_editor.h
@@ -78,14 +78,14 @@ class GroupsEditor : public VBoxContainer {
 	Button *add = nullptr;
 	Tree *tree = nullptr;
 
-	HashMap<Node *, HashMap<StringName, bool>> scene_groups_cache;
+	HashMap<ObjectID, HashMap<StringName, bool>> scene_groups_cache;
 	HashMap<StringName, bool> scene_groups_for_caching;
 
 	HashMap<StringName, bool> scene_groups;
 	HashMap<StringName, String> global_groups;
 
-	void _update_scene_groups(Node *p_node);
-	void _cache_scene_groups(Node *p_node);
+	void _update_scene_groups(const ObjectID &p_id);
+	void _cache_scene_groups(const ObjectID &p_id);
 
 	void _show_add_group_dialog();
 	void _show_rename_group_dialog();


editor/groups_editor.cpp
```diff
@@ -803,7 +803,7 @@ void GroupsEditor::_bind_methods() {
 void GroupsEditor::_node_removed(Node *p_node) {
 	if (scene_root_node == p_node) {
 		scene_groups_for_caching = scene_groups;
-		callable_mp(this, &GroupsEditor::_cache_scene_groups).call_deferred(p_node);
+		callable_mp(this, &GroupsEditor::_cache_scene_groups).call(p_node);
```
DarkMessiah, line 806:
```suggestion
		_cache_scene_groups(p_node);
```

tdaven: Yeah, suppose there is no need to even go through callable.

AThousandShips:
I'm not sure this is correct, calling this deferred shouldn't cause a crash unless some check is invalid later, instead checking down the line should be done IMO

tdaven:
I'm not sure this is right either.  I'm a bit confused what this is trying to do on node removal in the first place.  Hopefully @DarkMessiah can clarify.

DarkMessiah:
As far as I remember, there was a crash during development at some iteration, and that's when I changed that part to a deferred call. I think passing `ObjectID` would be the best solution in this situation. 

@AThousandShips Is the `ObjectID` unique within a single session?

tdaven:
I went ahead and pushed a version using ObjectID.

DarkMessiah:
I think that we don't need to cast `ObjectID` to `Node`. The best solution is to use `ObjectID` instead of `Node *`. 
@AThousandShips, am I right? :) 

```
HashMap<ObjectID, HashMap<StringName, bool>> scene_groups_cache;
void _update_scene_groups(ObjectID p_id);
void _cache_scene_groups(ObjectID p_id);

void GroupsEditor::_update_scene_groups(ObjectID p_id) {
        if (scene_groups_cache.has(p_id)) {
		scene_groups = scene_groups_cache[p_id];
		scene_groups_cache.erase(p_id);
	} else {
		scene_groups = HashMap<StringName, bool>();
	}
}

void GroupsEditor::_cache_scene_groups(ObjectID p_id) {
        const int edited_scene_count = EditorNode::get_editor_data().get_edited_scene_count();
	for (int i = 0; i < edited_scene_count; i++) {
		if (p_id == EditorNode::get_editor_data().get_edited_scene_root(i)->get_instance_id()) {
			scene_groups_cache[p_id] = scene_groups_for_caching;
			break;
		}
	}
}
```

tdaven:
That can work but you still need to validate you get a valid node back from `get_edited_scene_root` since you can get `nullptr` back when you close the last scene tab.  By the time the deferred call is made there is only the default empty scene tab.

AThousandShips:
Will take a look when I have time, likely in a few days, the idea here was to use `ObjectID` and retrieve it from `ObjectDB`

tdaven:
This comment I added to the issue adds come good context about this issue I feel.
https://github.com/godotengine/godot/issues/86427#issuecomment-1871609874

KoBeWi:
Wouldn't it be enough to only change the deferred call to take ObjectID?

tdaven:
Yup.  Definitely.   I had a version that did that and used the ObjectDB to look it up here:
https://github.com/godotengine/godot/compare/master...tdaven:godot:tdaven/fix-86427-object-id

But then @DarkMessiah suggested doing it this way and dropping the use of node pointers entirely.  Then @AThousandShips mentioned using the ObjectDB.  

All methods I've pushed work(removing deferred call, use ObjectID for only the deferred call and removing the use of objectID).  There just isn't a consensus on the which way is preferred.  Just waiting for reviewers to come to agreement.

AThousandShips (approved):
This change makes sense to me, the actual object isn't required as such, and the change is only to the specific data type, not the handling in general so dangling data etc. isn't introduced by this at least

Not familiar with the wider system, but as long as the system itself worked before with the one issue being dangling references this looks great!

editor/groups_editor.cpp
```diff
@@ -274,20 +274,21 @@ void GroupsEditor::_update_groups_and_tree() {
...
 }
 
-void GroupsEditor::_update_scene_groups(Node *p_node) {
-	if (scene_groups_cache.has(p_node)) {
-		scene_groups = scene_groups_cache[p_node];
-		scene_groups_cache.erase(p_node);
+void GroupsEditor::_update_scene_groups(ObjectID p_id) {
+	if (scene_groups_cache.has(p_id)) {
+		scene_groups = scene_groups_cache[p_id];
+		scene_groups_cache.erase(p_id);
```
KoBeWi, line 280: Unrelated, but now I noticed this function is inefficient. This would be better:
```suggestion
	HashMap<Node *, HashMap<StringName, bool>>::Iterator I = scene_groups_cache.find(p_node);
	if (I) {
		scene_groups = I->value;
		scene_groups_cache.remove(I);
```

editor/groups_editor.h
```diff
@@ -78,14 +78,14 @@ class GroupsEditor : public VBoxContainer {
...
+	HashMap<ObjectID, HashMap<StringName, bool>> scene_groups_cache;
 	HashMap<StringName, bool> scene_groups_for_caching;
 
 	HashMap<StringName, bool> scene_groups;
 	HashMap<StringName, String> global_groups;
 
-	void _update_scene_groups(Node *p_node);
-	void _cache_scene_groups(Node *p_node);
+	void _update_scene_groups(ObjectID p_id);
+	void _cache_scene_groups(ObjectID p_id);
```
KoBeWi, line 88:
```suggestion
	void _update_scene_groups(const ObjectID &p_id);
	void _cache_scene_groups(const ObjectID &p_id);
```

KoBeWi:
My suggestion was that this PR could've been simpler (involve less changes, because they are unnecessary), but the current solution is ok too.

fire:
I did not review the code, but I can confirm that https://github.com/godotengine/godot/issues/86844 is fixed by this.

akien-mga:
Rebased to include KoBeWi's suggestions.

akien-mga:
Thanks!

PR #86468, opened 2023-12-23
Move script name to top

### LINKED ISSUE #8317: Move the edited script filename so it's still displayed when the scripts panel is folded

Game

The edited script filename is not displayed anywhere when the scripts panel is folded

image

So it's not folded along with the scripts panel when using 'Toggle Scripts Panel'

image

See above

Should be core

Should be core

KoBeWi:
Closes https://github.com/godotengine/godot-proposals/issues/8317
image
Fun fact: the script name and icon on top was always there, and even updated properly, but never displayed after `hide()`. The PR simply makes it visible (I also did some tweaks/cleanup).
Also removed the name on left and moved the sort icon next to filter.

### DIFF
--- a/editor/script/script_editor_plugin.cpp
+++ b/editor/script/script_editor_plugin.cpp
@@ -736,25 +736,26 @@ void ScriptEditor::_go_to_tab(int p_idx) {
 
 	c = tab_container->get_current_tab_control();
 
-	if (Object::cast_to<ScriptEditorBase>(c)) {
-		script_name_label->set_text(Object::cast_to<ScriptEditorBase>(c)->get_name());
-		script_icon->set_texture(Object::cast_to<ScriptEditorBase>(c)->get_theme_icon());
+	ScriptEditorBase *seb = Object::cast_to<ScriptEditorBase>(c);
+	if (seb) {
 		if (is_visible_in_tree()) {
-			Object::cast_to<ScriptEditorBase>(c)->ensure_focus();
+			seb->ensure_focus();
 		}
 
-		Ref<Script> scr = Object::cast_to<ScriptEditorBase>(c)->get_edited_resource();
+		Ref<Script> scr = seb->get_edited_resource();
 		if (scr.is_valid()) {
 			notify_script_changed(scr);
 		}
 
-		Object::cast_to<ScriptEditorBase>(c)->validate();
+		seb->validate();
 	}
-	if (Object::cast_to<EditorHelp>(c)) {
-		script_name_label->set_text(Object::cast_to<EditorHelp>(c)->get_class());
-		script_icon->set_texture(get_editor_theme_icon(SNAME("Help")));
+
+	EditorHelp *eh = Object::cast_to<EditorHelp>(c);
+	if (eh) {
+		script_name_label->set_text(eh->get_class());
+
 		if (is_visible_in_tree()) {
-			Object::cast_to<EditorHelp>(c)->set_focused();
+			eh->set_focused();
 		}
 	}
 
@@ -934,6 +935,7 @@ void ScriptEditor::_close_tab(int p_idx, bool p_save, bool p_history_back) {
 		} else {
 			_update_selected_editor_menu();
 			_update_online_doc();
+			script_name_label->set_text(String());
 		}
 
 		_update_history_arrows();
@@ -1836,8 +1838,6 @@ void ScriptEditor::_notification(int p_what) {
 			filter_scripts->set_right_icon(get_editor_theme_icon(SNAME("Search")));
 			filter_methods->set_right_icon(get_editor_theme_icon(SNAME("Search")));
 
-			filename->add_theme_style_override(CoreStringName(normal), get_theme_stylebox(CoreStringName(normal), SNAME("LineEdit")));
-
 			recent_scripts->reset_size();
 
 			if (is_inside_tree()) {
@@ -2013,6 +2013,7 @@ void ScriptEditor::_script_selected(int p_idx) {
 	grab_focus_block = !Input::get_singleton()->is_mouse_button_pressed(MouseButton::LEFT); //amazing hack, simply amazing
 
 	_go_to_tab(script_list->get_item_metadata(p_idx));
+	script_name_label->set_text(script_list->get_item_text(p_idx));
 	grab_focus_block = false;
 }
 
@@ -2145,11 +2146,6 @@ void ScriptEditor::_update_members_overview() {
 			members_overview->set_item_metadata(-1, line);
 		}
 	}
-
-	String path = se->get_edited_resource()->get_path();
-	bool built_in = !path.is_resource_file();
-	String name = built_in ? path.get_file() : se->get_name();
-	filename->set_text(name);
 }
 
 void ScriptEditor::_update_help_overview_visibility() {
@@ -2171,7 +2167,6 @@ void ScriptEditor::_update_help_overview_visibility() {
 		filter_methods->set_visible(false);
 		help_overview->set_visible(true);
 		overview_vbox->set_visible(true);
-		filename->set_text(se->get_name());
 	} else {
 		help_overview->set_visible(false);
 		overview_vbox->set_visible(false);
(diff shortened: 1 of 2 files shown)


YuriSizov:
I support this conceptually, but we need to make sure it works nicely on small resolutions with any script name. This bar of controls already introduces issues with content overflow in some languages which translate verbosely, like French.

KoBeWi:
I checked French editor and there is no problem in 1920, but lower resolutions might have problems.

Though not sure how to make it friendly for low resolutions. If I enable trimming, the text is shrunk to 0. I can't think of a way to have a centered, trimmed text with an icon next to it.

YeldhamDev:
What's the point of showing the script's name twice, exactly?

KoBeWi:
It's not shown when you shrink the left panel (and can still be switched when switching scenes).

YeldhamDev:
> It's not shown when you shrink the left panel

I think the people who do that does so because they don't have enough horizontal space, the situation where this feature might be a problem. :stuck_out_tongue:

YuriSizov:
> I think the people who do that does so because they don't have enough horizontal space, the situation where this feature might be a problem. 😛

As someone who had a collapsed script list and made a plugin to turn it into a dock when working on a smaller laptop, I'd say that having a name visible is a necessity in that case. Although I put it on the bottom bar in my plugin:

akien-mga:
There's quite a lot of demand in the community for having script tabs at the top, instead of scene tabs, when in the script editor. This may replace the need for such a change if we were to go this way (even detached, the script editor should still have tabs at the top to switch between scripts IMO).

Otherwise, if we don't go this route or keep it configurable, I see the need for making the current script name more prominent. But I have concerns about the usability of the approach here, the top bar is quite busy and the added name doesn't particularly stand out compared to other buttons. And indeed on smaller resolutions / long script names / some locales this could become a problem. Maybe Yuri's approach to put it in the footer would be better.

KoBeWi:
It can't be add the bottom, because that's where the errors appear.
image
There is no space.

Also, unlike the top bar, the bottom one is separate per opened script, so the logic will be more complex (it needs to update in case of script rename etc.).

YeldhamDev:
> It can't be add the bottom, because that's where the errors appear.

Couldn't the error just replace the name once it appears? I don't think it's that important to show both at the same time.

KoBeWi:
I removed the icon and enabled trimming in the label, so it doesn't enforce minimum size:

https://github.com/godotengine/godot/assets/2223172/b0274029-7b5d-4d28-a6e0-6ccf615cfb90

Since moving it to bottom is going to be a hassle, I'd prefer doing that only if the above solution is not sufficient.

editor/script/script_editor_plugin.cpp
```diff
@@ -3895,27 +3890,21 @@ ScriptEditor::ScriptEditor(WindowWrapper *p_wrapper) {
...
-	filename = memnew(Label);
-	filename->set_clip_text(true);
-	filename->set_h_size_flags(SIZE_EXPAND_FILL);
-	filename->add_theme_style_override("normal", EditorNode::get_singleton()->get_editor_theme()->get_stylebox(SNAME("normal"), SNAME("LineEdit")));
-	buttons_hbox->add_child(filename);
+	filter_methods = memnew(LineEdit);
+	filter_methods->set_placeholder(TTR("Filter Methods"));
+	filter_methods->set_clear_button_enabled(true);
+	filter_methods->set_h_size_flags(SIZE_EXPAND_FILL);
+	filter_methods->connect("text_changed", callable_mp(this, &ScriptEditor::_filter_methods_text_changed));
```
AThousandShips:
```suggestion
	filter_methods->connect(SceneStringName(text_changed), callable_mp(this, &ScriptEditor::_filter_methods_text_changed));
```

editor/plugins/script_editor_plugin.cpp
```diff
@@ -4038,14 +4027,12 @@ ScriptEditor::ScriptEditor(WindowWrapper *p_wrapper) {
 
 	menu_hb->add_spacer();
```
kitbdev: Long script names get cut off, even though it has space. I think you can remove the spacers and just rely on the label horizontal centering.

editor/script/script_editor_plugin.cpp
```diff
@@ -662,25 +662,28 @@ void ScriptEditor::_go_to_tab(int p_idx) {
 
 	c = tab_container->get_current_tab_control();
 
-	if (Object::cast_to<ScriptEditorBase>(c)) {
-		script_name_label->set_text(Object::cast_to<ScriptEditorBase>(c)->get_name());
-		script_icon->set_texture(Object::cast_to<ScriptEditorBase>(c)->get_theme_icon());
+	ScriptEditorBase *seb = Object::cast_to<ScriptEditorBase>(c);
+	if (seb) {
+		script_name_label->set_text(seb->get_name());
```
kitbdev: This uses a different name than the one in `ScriptEditor::_update_script_names()` uses, which is disambiguated and uses `text_editor/script_list/list_script_names_as`. 
The `_update_script_names` one will be automatically called for scripts after a moment which causes the name to change.
They should determine the name the same way (either ignore the setting for this label, or follow it here). 
Or maybe we can rely on `_update_script_names` and not need to set it here?

kitbdev (commented):
It makes sense to take advantage of this empty space, and adding the name helps add clarity.
And its already partially implemented so we should to do something about it.

Also I think it is fine at the top, having it at the bottom but only if there are no errors seems like it would be confusing.

editor/script/script_editor_plugin.cpp
KoBeWi: Yeah, seems like the name label is set twice. I think non-disambiguated name should be preferred.

editor/script/script_editor_plugin.cpp
```diff
@@ -4271,31 +4264,22 @@ ScriptEditor::ScriptEditor(WindowWrapper *p_wrapper) {
...
+	buttons_hbox->add_child(filter_methods);
 
 	members_overview_alphabeta_sort_button = memnew(Button);
 	members_overview_alphabeta_sort_button->set_theme_type_variation(SceneStringName(FlatButton));
 	members_overview_alphabeta_sort_button->set_tooltip_text(TTRC("Toggle alphabetical sorting of the method list."));
 	members_overview_alphabeta_sort_button->set_toggle_mode(true);
 	members_overview_alphabeta_sort_button->set_pressed(EDITOR_GET("text_editor/script_list/sort_members_outline_alphabetically"));
-	members_overview_alphabeta_sort_button->connect(SceneStringName(toggled), callable_mp(this, &ScriptEditor::_toggle_members_overview_alpha_sort));
-
+	members_overview_alphabeta_sort_button->connect("toggled", callable_mp(this, &ScriptEditor::_toggle_members_overview_alpha_sort));
```
kitbdev: `SceneStringName(toggled)` is preferred right?

editor/script/script_editor_plugin.cpp
```diff
@@ -2443,9 +2439,6 @@ void ScriptEditor::_update_script_names() {
 		if (tab_container->get_current_tab() == sedata_filtered[i].index) {
 			script_list->select(index);
 
-			script_name_label->set_text(sedata_filtered[i].name);
```
kitbdev, line 2446: There are some situations that setting the name here is needed: Using save as to change the filename, on _filesystem_changed when the filesystem renames the file. and when the history back/forward buttons are pressed.

editor/script/script_editor_plugin.cpp
KoBeWi: Looks like rebase error.

editor/script/script_editor_plugin.cpp
KoBeWi: Ok I restored it and instead changed the other line, because it was unreliable.

kitbdev (approved):
Looks good.
And if we do decide to have tabs in the future, we can reassess this.

Repiteo:
Thanks!

PR #86472, opened 2023-12-23
Update Curve3D.xml

MichaelMacha:
Added clarification on discrepancy in implementation of `sample_baked_with_rotation`. GDScript expects the offset to be normalized, while C++ apparently does not.

This is worthy of discussion before accepting the change. Is it a bug? Is it possible to fix without breaking projects further? The C++ code clearly clamps offset to [0, length), but in GDScript the value only behaves as expected when something like `PathFollow3D.progress_ratio` is used (normalized) instead of `PathFollow3D.progress` (unnormalized).

Moreover, if the difference is accepted and intentional, is there a better place to document the GDScript version? It feels essential, yet out-of-place, in the C++ docs.

### DIFF
--- a/doc/classes/Curve3D.xml
+++ b/doc/classes/Curve3D.xml
@@ -137,6 +137,7 @@
 			<param index="2" name="apply_tilt" type="bool" default="false" />
 			<description>
 				Returns a [Transform3D] with [code]origin[/code] as point position, [code]basis.x[/code] as sideway vector, [code]basis.y[/code] as up vector, [code]basis.z[/code] as forward vector. When the curve length is 0, there is no reasonable way to calculate the rotation, all vectors aligned with global space axes. See also [method sample_baked].
+				Note that in the GDScript implementation, offset is expected to be normalized, that is, between zero and one. This diverges slightly from the C++ implementation, which expects an unnormalized offset.
 			</description>
 		</method>
 		<method name="samplef" qualifiers="const">


AThousandShips:
I don't see the need to document this detail, the documentation is for the bound methods, not the c++ internals, so this discrepancy isn't relevant to this documentation, nowhere else do we clarify things like this even when they occur

doc/classes/Curve3D.xml
```diff
@@ -137,6 +137,7 @@
 			<param index="2" name="apply_tilt" type="bool" default="false" />
 			<description>
 				Returns a [Transform3D] with [code]origin[/code] as point position, [code]basis.x[/code] as sideway vector, [code]basis.y[/code] as up vector, [code]basis.z[/code] as forward vector. When the curve length is 0, there is no reasonable way to calculate the rotation, all vectors aligned with global space axes. See also [method sample_baked].
+				Note that in the GDScript implementation, offset is expected to be normalized, that is, between zero and one. This diverges slightly from the C++ implementation, which expects an unnormalized offset.
```
AThousandShips, line 140: This is false, offset is in length along the curve, not normalized

AThousandShips (changes requested):

doc/classes/Curve3D.xml
MichaelMacha: Yes, I just showed up to rescind this as I'm getting conflicting results in my own analysis code. Apologies, I'm going to work on that this morning and figure out where the issue is.

AThousandShips: I don't think there's any issues in the implementation, but in your testing, but in any case if there are any issues this PR is not the solution

MichaelMacha: Perhaps I'm simply not seeing something, but my basis seems to require `progress_ratio`, while my origin seems to want `progress`, in spite of them being from the same function.

I'm ultimately attempting to align a skeleton's bones, programmatically, along a curve. Here's the code that's doing questionable things, if I may discuss it with an appropriate amount of humility:

```
func _process(delta):
	follow.progress_ratio += delta * speed
	
	#update armature
	
	#This aligns to the up vector for the curve, but we want the forward vector.
	var sample := path.curve.sample_baked_with_rotation(follow.progress_ratio).basis
	#var sample := path.curve.sample_baked_with_rotation(follow.progress).basis
	
	skeleton.set_bone_pose_rotation(0, 
		Quaternion(Vector3.RIGHT, -PI/2.0) * Quaternion(sample) #Quaternion rotation gives us forward
		)
	print(follow.progress, " vs ", path.curve.get_baked_length())
	var dist = skeleton.get_bone_global_pose(1).origin.length()
	var dist_normalized = dist / path.curve.get_baked_length()
	
	print(dist_normalized)
	#print(dist)
	
	print("T: ", follow.progress_ratio + dist_normalized)
	aim.global_position = path.curve.sample_baked_with_rotation(follow.progress + dist).origin

```

You'll note that the initial attempt to get a Transform3D used `progress` instead of `progress_ratio`, but the behavior of the bone is entirely wrong this way. When, as an experiment, I set it to `progress_ratio`, it worked out perfectly.

However, when I attempt to set the position of aim (a MeshInstance3D blue sphere) to a progressed point along the curve, it requires `progress` to function properly.

Is it possible that `_sample_posture`, which the basis is retrieved with, is expecting a normalized coordinate instead, and is returning a value on the basis of that?

AThousandShips: I think you're doing something wrong here, otherwise `PathFollow3D` wouldn't work as it just uses non-nirmalized offset, I think there's instead something wrong with how you use this value, or you are making incorrect assumptions about the orientation

MichaelMacha: According to experience and the documentation, PathFollow3D supplies both progress, for rough progress in world units along its path, and progress_ratio, which is the normalized progress along it (the value usually referred to as 't' in a Bezier curve's equation).

To test this, I used a basic curve with a twist of 6π at one of its control points. When I attempt with progress, the behavior here is very strange:

image

But, when I set it to progress_ratio, it behaves as expected, all the way to the end:

image

(Note that the camera is a child of the pointy mesh here, so orientation will be slightly different; but the mesh points along the curve, as intended, only when I'm using `progress_ratio` instead of `progress`.)

On the other hand, my aim ball, which I'm using to determine whether I'm getting the appropriate position after calculating an offset, is locked close to the first 0-1 interval of the curve when I use `progress_ratio`, but behaves fine when I use `progress`.
image
image

This seems very much to be two different results. That leads me to believe that the C++ _sample_posture is perhaps being called with the wrong value? Are we sure that it does not expect a normalized value here?

AThousandShips: It does not use the ratio one though, and it doesn't call any normalized value, check the code yourself if you don't trust me 🙂, but I know it doesn't use the normalized one, simple as that

MichaelMacha: Oh don't worry, I'm hardly taking any of this personally; but my code is performing in an unexpected way while using a few nodes which have some spotty documentation. It seems like a worthy use of my time to figure out why it's going wrong, and try and fix that.

I'm seeing that in the C++, Transform3D is constructed from two values, an origin point and a basis, sequentially. The basis is gotten with this line:

```Basis frame = _sample_posture(interval, p_apply_tilt);```

`interval` is used for the origin, too.  I'm not entirely sure what a Curve3D::Interval is yet, as it's simply returned from `_find_interval(p_offset)` and I haven't had enough time to review it, but the GDScript I showed you is pretty much all I've written for this project. I may comment it all out and try again from the top later, but it really seems to me that I should be able to call `sample_baked_with_rotation` with a single progress value and get consistent results.

To benefit this discussion, here's the full script for this project. If you think I need to share the full project then I can, it's no big deal; but I really feel like the issue is right in front of us here.

[extends Node3D

@export var speed : float = 0.1

@onready var path : Path3D = $"../.."
@onready var follow : PathFollow3D = $".."
@onready var skeleton : Skeleton3D = $"Armature/Skeleton3D"
@onready var aim : MeshInstance3D = $Aim

func _ready():
	pass # Replace with function body.


# Called every frame. 'delta' is the elapsed time since the previous frame.
func _process(delta):
	follow.progress_ratio += delta * speed
	
	#update armature
	
	#This aligns to the up vector for the curve, but we want the forward vector.
	var sample := path.curve.sample_baked_with_rotation(follow.progress_ratio).basis
	#var sample := path.curve.sample_baked_with_rotation(follow.progress).basis
	
	# Documentation bug discovered; C++ expects unnormalized curve progress, while GDScript expects
	# normalized progress.
	
	skeleton.set_bone_pose_rotation(0, 
		Quaternion(Vector3.RIGHT, -PI/2.0) * Quaternion(sample) #Quaternion rotation gives us forward
		)
	print(follow.progress, " vs ", path.curve.get_baked_length())
	var dist = skeleton.get_bone_global_pose(1).origin.length()
	var dist_normalized = dist / path.curve.get_baked_length()
	
	print(dist_normalized)
	#print(dist)
	
	print("T: ", follow.progress_ratio + dist_normalized)
	aim.global_position = path.curve.sample_baked_with_rotation(follow.progress + dist).origin
	
](url)

AThousandShips: But all this aside, what ever this is caused by, the following fact remains:
The statement you've added here is false, and the method *only* takes a value that's in the range of the baked length

I'd suggest taking this question to the other community channels, but this change isn't valid

If you find an actual bug here after doing this please open a bug report, however I'll go ahead and close this as this is not a correct change

If you figure out that this is an issue with how to describe the return value further we can add that to the documentation

But the underlying assumption of this PR is wrong and the fix is also wrong

AThousandShips:
Closing as per the above discussion

Thank you for your contribution nonetheless

doc/classes/Curve3D.xml
MichaelMacha: One issue that I'm seeing is that I think PathFollow3D is possibly applying its own rotation here too... man it would be such a relief if this was just something I messed up.

AThousandShips: I'd suggest following the code in `PathFollow`, as it evidently works 🙂

MichaelMacha:
My mistake, but thank you for the assist!

MichaelMacha:
> Closing as per the above discussion
> 
> Thank you for your contribution nonetheless

My mistake, but thank you for the assist!

AThousandShips:
If there's some details in the documentation that can be clarified this could also be reopened with a correct setup 🙂

doc/classes/Curve3D.xml
MichaelMacha: I'm not sure of that, but I've gotten my idea to work regardless using inverse kinematics and two PathFollow3D nodes (one for the worm, one for the target). I'm a little concerned by SkeletonIK3D being deprecated at the moment, but as far as I can tell nothing explicit has replaced it, and aside from a few workflow issues it's working well.

Likely my brain just went on strike at the end of December...

At this point I'm more than content to leave this issue closed, but thank you for your concern!

PR #65797, opened 2022-09-14
Add `loop_mode` to AudioStreamPlayer & AudioStreamPlayback

Mickeon:
Implements https://github.com/godotengine/godot-proposals/issues/3120, (minus actually stripping the import)

This PR implemets `loop_mode` for  **AudioStreamPlayer**, **AudioStreamPlayer2D**, **AudioStreamPlayer3D** and  **AudioStreamPlayback**.

This allows changing the looping mode of the **AudioStream**, regardless of how it was imported, even as the stream is playing, which leads to pretty interesting results! Imagine playing a sound, keep it looping continuously, until a key is released.

https://user-images.githubusercontent.com/66727710/190255193-f79c8d16-41c2-44ed-b623-baa1ea9da6bd.mp4

By default, for the **AudioStreamPlayer**'s, this property is set to "_Detect_", which lets any generated **AudioStreamPlayback** automatically figure out the `loop_mode`, based on their respective **AudioStream** Resource. 

image

Note is that, the streams may vary a lot _(MP3, Ogg, WAV...)_. Their implementation has been mostly unchanged in this PR, so it is not guaranteed that the given `loop_mode` is supported, if at all, in fact. Whenever this is the case, a warning configuration is issued. When actually playing the AudioStream with AudioStreamPlayback, the variable is internally reassigned to its closest equivalent and a warning is sent to the console (this has been configured manually).

| Configuration Warning | Console Warning |
| --- | --- |
| image image | image |

The way **AudioStreams** are imported into the editor is left unchanged. This feature is all on top of it.

## What I would like to do:
- [x] Fix **AudioStreamPlaybackWAV** not resuming forward as expected after switching from Ping Pong/Backward to Forward (as if stuck looping backwards).
- [x] ~Hide unsupported Loop Modes from selection when the **AudioStream** of **AudioStreamPlayer** does not support them _(I believe `validate_property()` allows you to do this)_.~ PROPERTY_TYPE_ENUM does not support skipping options _(adding that to my list, I suppose)_. The closest thing is a reminder next to each option.
- [x] Move LoopMode::DETECT to the top of the enum.
- [x] Manage detecting unsupported loop modes outside of **AudioStreamPlayback**, and into **AudioStreamPlayer** (Have **AudioStream** return a list of supported modes).
- [x] Clean-up this code considerably.
- [x] Add and update the documentation.

## Optional ideas:
- Default most **AudioStreamPlayback**s created with `AudioStream.instantiate_playback()` to `LoopMode.DETECT`.
[...]

PR #64488, opened 2022-08-16
Add interactive music support

reduz:
This PR adds 3 types of audio streams used for interactive music support.

* **AudioStreamInteractive**: Allows setting several clips as sub-streams and edit the transition table between them.
* **AudioStreamPlaylist**: Allows sequential or shuffled playback of a list of streams.
* **AudioStreamSynchronized**: Allows synchronous playback of several streams, the volume of each can be controlled.

These three stream types can be combined to create complex, layered interactive music and transitions between them, similar to software such as WWise, FMOD or Elias.

As an example:
* AudioStreamPlaylist insde of AudioStreamInteractive resembles how WWise does interactive music.
* AudioStreamSynchronized inside of AudioStreamInteractive resembles how Elias does interactive music.

In Godot style, these components are presented stand alone, and can be mixed and matched by the user to achieve the desired function. This supersedes #62896 and #32769.

Screens:
Interactive music, list of clips:
image
Transitions:
image
**NOTE**: Functionality is missing in Tree so it cant be expanded and have the rightmost column always visible, hopfully this will be added/fixed later on.
Playlist:
image
Synchronizer:
image

Preview video:

https://github.com/godotengine/godot/assets/6265307/9d573824-702b-4d19-8ff2-717ccbde1c96

Sample project:

[interactive_music.zip](https://github.com/godotengine/godot/files/13986908/interactive_music.zip)

PR #86473, opened 2023-12-23
Implement audio stream playback parameters.

reduz:
Implements a way for audio stream playback to be configured via parameters directly in the edited AudioStreamPlayer[2D/3D].

Currently, configuring the playback stream is not possible (or is sometimes hacky as the user has to obtain the currently played stream, which is not always immediately available).

All you can do is configure the AudioStream itself, which affects all instances in the game. With this PR, configuring playback parameters per each stream player is possible, giving users far more flexibility and freedom to configure playback or animate parameters.

This PR only implements this new feature to control looping in stream playback instances (a commonly requested feature, which was lost in the transition from Godot 2 to Godot 3). But the idea is that it can do a lot more:

* If effects are bundled to the stream, control per playback instance parameters such as cutoff or resoance, or any other exposed effect parameter per playback instance.
* For the upcoming interactive music PR (#64488), this exposes an easy way to change the active clip, which was not possible before (so this PR is a prerequisite for that one).
* For the upcoming parametrizable audio support (https://github.com/godotengine/godot-proposals/issues/3394) this allows editing and animating audio graph parameters.

Screenshot of how it looks:

image

Supersedes #65797, implements https://github.com/godotengine/godot-proposals/issues/3120.

### DIFF
--- a/doc/classes/AudioStream.xml
+++ b/doc/classes/AudioStream.xml
@@ -28,6 +28,12 @@
 			<description>
 			</description>
 		</method>
+		<method name="_get_parameter_list" qualifiers="virtual const">
+			<return type="Dictionary[]" />
+			<description>
+				Return the controllable parameters of this stream. This array contains dictionaries with a property info description format (see [method Object.get_property_list]). Additionally, the default value for this parameter must be added tho each dictionary in "default_value" field.
+			</description>
+		</method>
 		<method name="_get_stream_name" qualifiers="virtual const">
 			<return type="String" />
 			<description>
@@ -62,4 +68,11 @@
 			</description>
 		</method>
 	</methods>
+	<signals>
+		<signal name="parameter_list_changed">
+			<description>
+				Signal to be emitted to notify when the parameter list changed.
+			</description>
+		</signal>
+	</signals>
 </class>
--- a/modules/minimp3/audio_stream_mp3.cpp
+++ b/modules/minimp3/audio_stream_mp3.cpp
@@ -46,7 +46,9 @@ int AudioStreamPlaybackMP3::_mix_internal(AudioFrame *p_buffer, int p_frames) {
 	int frames_mixed_this_step = p_frames;
 
 	int beat_length_frames = -1;
-	bool beat_loop = mp3_stream->has_loop() && mp3_stream->get_bpm() > 0 && mp3_stream->get_beat_count() > 0;
+	bool use_loop = looping_override ? looping : mp3_stream->loop;
+
+	bool beat_loop = use_loop && mp3_stream->get_bpm() > 0 && mp3_stream->get_beat_count() > 0;
 	if (beat_loop) {
 		beat_length_frames = mp3_stream->get_beat_count() * mp3_stream->sample_rate * 60 / mp3_stream->get_bpm();
 	}
@@ -82,7 +84,7 @@ int AudioStreamPlaybackMP3::_mix_internal(AudioFrame *p_buffer, int p_frames) {
 
 		else {
 			//EOF
-			if (mp3_stream->loop) {
+			if (use_loop) {
 				seek(mp3_stream->loop_offset);
 				loops++;
 			} else {
@@ -141,6 +143,25 @@ void AudioStreamPlaybackMP3::seek(double p_time) {
 
 void AudioStreamPlaybackMP3::tag_used_streams() {
 	mp3_stream->tag_used(get_playback_position());
+}
+
+void AudioStreamPlaybackMP3::set_parameter(const StringName &p_name, const Variant &p_value) {
+	if (p_name == SNAME("looping")) {
+		if (p_value == Variant()) {
+			looping_override = false;
+			looping = false;
+		} else {
+			looping_override = true;
+			looping = p_value;
+		}
+	}
+}
+
+Variant AudioStreamPlaybackMP3::get_parameter(const StringName &p_name) const {
+	if (looping_override && p_name == SNAME("looping")) {
+		return looping;
+	}
+	return Variant();
 }
 
 AudioStreamPlaybackMP3::~AudioStreamPlaybackMP3() {
@@ -230,6 +251,10 @@ double AudioStreamMP3::get_length() const {
 
 bool AudioStreamMP3::is_monophonic() const {
 	return false;
+}
+
+void AudioStreamMP3::get_parameter_list(List<Parameter> *r_parameters) {
+	r_parameters->push_back(Parameter(PropertyInfo(Variant::BOOL, "looping", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_DEFAULT | PROPERTY_USAGE_CHECKABLE), Variant()));
 }
 
 void AudioStreamMP3::set_bpm(double p_bpm) {
--- a/modules/minimp3/audio_stream_mp3.h
+++ b/modules/minimp3/audio_stream_mp3.h
@@ -47,6 +47,8 @@ class AudioStreamPlaybackMP3 : public AudioStreamPlaybackResampled {
 	AudioFrame loop_fade[FADE_SIZE];
 	int loop_fade_remaining = FADE_SIZE;
 
+	bool looping_override = false;
+	bool looping = false;
 	mp3dec_ex_t *mp3d = nullptr;
 	uint32_t frames_mixed = 0;
 	bool active = false;
@@ -71,6 +73,9 @@ public:
 	virtual void seek(double p_time) override;
 
 	virtual void tag_used_streams() override;
+
+	virtual void set_parameter(const StringName &p_name, const Variant &p_value) override;
+	virtual Variant get_parameter(const StringName &p_name) const override;
 
 	AudioStreamPlaybackMP3() {}
 	~AudioStreamPlaybackMP3();
@@ -125,6 +130,8 @@ public:
 	virtual double get_length() const override;
 
 	virtual bool is_monophonic() const override;
+
+	virtual void get_parameter_list(List<Parameter> *r_parameters) override;
 
 	AudioStreamMP3();
 	virtual ~AudioStreamMP3();
--- a/modules/vorbis/audio_stream_ogg_vorbis.cpp
+++ b/modules/vorbis/audio_stream_ogg_vorbis.cpp
@@ -46,8 +46,9 @@ int AudioStreamPlaybackOggVorbis::_mix_internal(AudioFrame *p_buffer, int p_fram
 	int todo = p_frames;
 
 	int beat_length_frames = -1;
-	bool beat_loop = vorbis_stream->has_loop();
-	if (beat_loop && vorbis_stream->get_bpm() > 0 && vorbis_stream->get_beat_count() > 0) {
+	bool use_loop = looping_override ? looping : vorbis_stream->loop;
+
+	if (use_loop && vorbis_stream->get_bpm() > 0 && vorbis_stream->get_beat_count() > 0) {
 		beat_length_frames = vorbis_stream->get_beat_count() * vorbis_data->get_sampling_rate() * 60 / vorbis_stream->get_bpm();
 	}
 
@@ -99,7 +100,7 @@ int AudioStreamPlaybackOggVorbis::_mix_internal(AudioFrame *p_buffer, int p_fram
 			} else
 			**/
 
-			if (beat_loop && beat_length_frames <= (int)frames_mixed) {
+			if (use_loop && beat_length_frames <= (int)frames_mixed) {
 				// End of file when doing beat-based looping. <= used instead of == because importer editing
 				if (!have_packets_left && !have_samples_left) {
 					//Nothing remaining, so do nothing.
@@ -125,7 +126,7 @@ int AudioStreamPlaybackOggVorbis::_mix_internal(AudioFrame *p_buffer, int p_fram
 		if (!have_packets_left && !have_samples_left) {
 			// Actual end of file!
 			bool is_not_empty = mixed > 0 || vorbis_stream->get_length() > 0;
-			if (vorbis_stream->loop && is_not_empty) {
+			if (use_loop && is_not_empty) {
 				//loop
 
 				seek(vorbis_stream->loop_offset);
@@ -255,6 +256,25 @@ double AudioStreamPlaybackOggVorbis::get_playback_position() const {
 
 void AudioStreamPlaybackOggVorbis::tag_used_streams() {
 	vorbis_stream->tag_used(get_playback_position());
+}
+
+void AudioStreamPlaybackOggVorbis::set_parameter(const StringName &p_name, const Variant &p_value) {
+	if (p_name == SNAME("looping")) {
+		if (p_value == Variant()) {
+			looping_override = false;
+			looping = false;
+		} else {
+			looping_override = true;
+			looping = p_value;
+		}
+	}
+}
+
+Variant AudioStreamPlaybackOggVorbis::get_parameter(const StringName &p_name) const {
+	if (looping_override && p_name == SNAME("looping")) {
+		return looping;
+	}
+	return Variant();
 }
 
 void AudioStreamPlaybackOggVorbis::seek(double p_time) {
@@ -491,6 +511,10 @@ int AudioStreamOggVorbis::get_bar_beats() const {
 
 bool AudioStreamOggVorbis::is_monophonic() const {
 	return false;
+}
+
+void AudioStreamOggVorbis::get_parameter_list(List<Parameter> *r_parameters) {
+	r_parameters->push_back(Parameter(PropertyInfo(Variant::BOOL, "looping", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_DEFAULT | PROPERTY_USAGE_CHECKABLE), Variant()));
 }
 
 void AudioStreamOggVorbis::_bind_methods() {
--- a/modules/vorbis/audio_stream_ogg_vorbis.h
+++ b/modules/vorbis/audio_stream_ogg_vorbis.h
@@ -44,6 +44,8 @@ class AudioStreamPlaybackOggVorbis : public AudioStreamPlaybackResampled {
 
 	uint32_t frames_mixed = 0;
 	bool active = false;
+	bool looping_override = false;
+	bool looping = false;
 	int loops = 0;
 
 	enum {
@@ -94,6 +96,9 @@ public:
 	virtual void seek(double p_time) override;
 
 	virtual void tag_used_streams() override;
+
+	virtual void set_parameter(const StringName &p_name, const Variant &p_value) override;
+	virtual Variant get_parameter(const StringName &p_name) const override;
 
 	AudioStreamPlaybackOggVorbis() {}
 	~AudioStreamPlaybackOggVorbis();
@@ -151,6 +156,8 @@ public:
 	virtual double get_length() const override; //if supported, otherwise return 0
 
 	virtual bool is_monophonic() const override;
+
+	virtual void get_parameter_list(List<Parameter> *r_parameters) override;
 
 	AudioStreamOggVorbis();
 	virtual ~AudioStreamOggVorbis();
--- a/scene/2d/audio_stream_player_2d.cpp
+++ b/scene/2d/audio_stream_player_2d.cpp
@@ -36,6 +36,8 @@
 #include "scene/main/window.h"
 #include "scene/resources/world_2d.h"
 
+#define PARAM_PREFIX "parameters/"
+
 void AudioStreamPlayer2D::_notification(int p_what) {
 	switch (p_what) {
 		case NOTIFICATION_ENTER_TREE: {
@@ -222,9 +224,36 @@ void AudioStreamPlayer2D::_update_panning() {
 	last_mix_count = AudioServer::get_singleton()->get_mix_count();
 }
 
+void AudioStreamPlayer2D::_update_stream_parameters() {
+	if (stream.is_null()) {
+		return;
+	}
+
+	List<AudioStream::Parameter> parameters;
+	stream->get_parameter_list(&parameters);
+	for (const AudioStream::Parameter &K : parameters) {
+		const PropertyInfo &pi = K.property;
+		StringName key = PARAM_PREFIX + pi.name;
+		if (!playback_parameters.has(key)) {
+			ParameterData pd;
+			pd.path = pi.name;
+			pd.value = K.default_value;
+			playback_parameters.insert(key, pd);
+		}
+	}
+}
+
 void AudioStreamPlayer2D::set_stream(Ref<AudioStream> p_stream) {
+	if (stream.is_valid()) {
+		stream->disconnect(SNAME("parameter_list_changed"), callable_mp(this, &AudioStreamPlayer2D::_update_stream_parameters));
+	}
 	stop();
 	stream = p_stream;
+	_update_stream_parameters();
+	if (stream.is_valid()) {
+		stream->connect(SNAME("parameter_list_changed"), callable_mp(this, &AudioStreamPlayer2D::_update_stream_parameters));
+	}
+	notify_property_list_changed();
 }
 
 Ref<AudioStream> AudioStreamPlayer2D::get_stream() const {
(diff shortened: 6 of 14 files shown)


modules/vorbis/audio_stream_ogg_vorbis.h
```diff
@@ -152,6 +156,8 @@ class AudioStreamOggVorbis : public AudioStream {
 
 	virtual bool is_monophonic() const override;
 
+	virtual void get_parameter_list(List<Parameter> *r_parameters);
```
AThousandShips, line 159:
```suggestion
	virtual void get_parameter_list(List<Parameter> *r_parameters) override;
```

modules/minimp3/audio_stream_mp3.h
```diff
@@ -126,6 +130,8 @@ class AudioStreamMP3 : public AudioStream {
 
 	virtual bool is_monophonic() const override;
 
+	virtual void get_parameter_list(List<Parameter> *r_parameters);
```
AThousandShips, line 133:
```suggestion
	virtual void get_parameter_list(List<Parameter> *r_parameters) override;
```

modules/minimp3/audio_stream_mp3.h
```diff
@@ -72,6 +73,9 @@ class AudioStreamPlaybackMP3 : public AudioStreamPlaybackResampled {
 
 	virtual void tag_used_streams() override;
 
+	virtual void set_parameter(const StringName &p_name, const Variant &p_value);
+	virtual Variant get_parameter(const StringName &p_name) const;
```
AThousandShips, line 77:
```suggestion
	virtual void set_parameter(const StringName &p_name, const Variant &p_value) override;
	virtual Variant get_parameter(const StringName &p_name) const override;
```

modules/vorbis/audio_stream_ogg_vorbis.h
```diff
@@ -95,6 +96,9 @@ class AudioStreamPlaybackOggVorbis : public AudioStreamPlaybackResampled {
 
 	virtual void tag_used_streams() override;
 
+	virtual void set_parameter(const StringName &p_name, const Variant &p_value);
+	virtual Variant get_parameter(const StringName &p_name) const;
```
AThousandShips, line 100:
```suggestion
	virtual void set_parameter(const StringName &p_name, const Variant &p_value) override;
	virtual Variant get_parameter(const StringName &p_name) const override;
```

KoBeWi:
Supersedes #65797 I guess.

doc/classes/AudioStream.xml
```diff
@@ -28,6 +28,12 @@
 			<description>
 			</description>
 		</method>
+		<method name="_get_parameter_list" qualifiers="virtual const">
+			<return type="Array" />
+			<description>
+				Return the amount of controllable parameters into this stream. This array must be interleaved as a dictionary with a property info description (see [method Object.get_property_list]) and the default value for it.
```
AThousandShips, line 34:
```suggestion
				Return the controllable parameters of this stream. This array must be interleaved as a dictionary with a property info description (see [method Object.get_property_list]) and the default value for it.
```
Or similar, as it should return an array and not a count

reduz:
@AThousandShips Thanks!! Merged and squashed.

reduz:
Did some changes, looping is now something you can check to override, to make it less confusing.

I think this is ready for review and merge.

aaronfranke:
I actually implemented per-player audio looping in The Mirror's fork of Godot a few months ago so that audio players can be looped by users, but I didn't know if this was desired to have upstream. So from that perspective, it's great to have this in the engine, less stuff for us to maintain :P

From the GLTF perspective, the latest iteration of the audio standard has looping as a part of the audio source (what Godot would call the audio stream), so this PR isn't useful for GLTF audio, but it doesn't hurt it either.

reduz:
@aaronfranke The problem we have in the audio side is that parameters exposed at stream level are very tricky because:
1. Streams many times are imported, so you can't edit them directly.
2. Even if you can edit them directly, they are shared across all players.

So, this makes editing or interacting with local instances very hard. This PR is created to make this easy and allow exposing parameters that can be edited locally. Additionally, as they appear at node level (similar to how animationtree works), you can animate them and edit them more freely.

servers/audio/audio_stream.h
```diff
@@ -124,6 +131,7 @@ class AudioStream : public Resource {
 	GDVIRTUAL0RC(bool, _has_loop)
 	GDVIRTUAL0RC(int, _get_bar_beats)
 	GDVIRTUAL0RC(int, _get_beat_count)
+	GDVIRTUAL0RC(Array, _get_parameter_list)
```
KoBeWi, line 134: Can this use TypedArray?

doc/classes/AudioStream.xml
```diff
@@ -28,6 +28,12 @@
 			<description>
 			</description>
 		</method>
+		<method name="_get_parameter_list" qualifiers="virtual const">
+			<return type="Array" />
+			<description>
+				Return the controllable parameters of this stream. This array must be interleaved as a dictionary with a property info description (see [method Object.get_property_list]) and the default value for it.
```
KoBeWi, line 34: I think I don't fully get this description. Is the expected format `[ Dictionary, default, Dictionary, default]`? Why the default can't be part of the Dictionary? PropertyInfo would just ignore it.

doc/classes/AudioStream.xml
```diff
@@ -62,4 +68,11 @@
 			</description>
 		</method>
 	</methods>
+	<signals>
+		<signal name="parameter_list_changed">
+			<description>
+				Signal to be emitted to notify if the parameter list changed.
```
KoBeWi, line 74:
```suggestion
				Signal to be emitted to notify when the parameter list has changed.
```

KoBeWi:
When you set a parameter in AudioStreamPlayer and then uncheck it, it will be stored in the scene as null.
image

scene/2d/audio_stream_player_2d.cpp
```diff
@@ -430,6 +461,39 @@ void AudioStreamPlayer2D::_on_bus_renamed(int p_bus_index, const StringName &p_o
 	notify_property_list_changed();
 }
 
+bool AudioStreamPlayer2D::_set(const StringName &p_name, const Variant &p_value) {
+	if (playback_parameters.has(p_name)) {
+		ParameterData &pd = playback_parameters[p_name];
+		pd.value = p_value;
```
AThousandShips, line 467: I think needs to detect if the variant is null, to handle the above issue

modules/minimp3/audio_stream_mp3.cpp
```diff
@@ -161,6 +181,7 @@ Ref<AudioStreamPlayback> AudioStreamMP3::instantiate_playback() {
 	mp3s.instantiate();
 	mp3s->mp3_stream = Ref<AudioStreamMP3>(this);
 	mp3s->mp3d = (mp3dec_ex_t *)memalloc(sizeof(mp3dec_ex_t));
+	mp3s->looping = false;
```
KoBeWi, line 184: Why is this set if it's the default value?

modules/minimp3/audio_stream_mp3.cpp
```diff
@@ -232,6 +253,10 @@ bool AudioStreamMP3::is_monophonic() const {
 	return false;
 }
 
+void AudioStreamMP3::get_parameter_list(List<Parameter> *r_parameters) {
+	r_parameters->push_back(Parameter(PropertyInfo(Variant::BOOL, "looping", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_DEFAULT | PROPERTY_USAGE_CHECKABLE), Variant()));
```
KoBeWi, line 257: And here. `Variant()` is the default value for the constructor.

modules/minimp3/audio_stream_mp3.h
```diff
@@ -47,6 +47,8 @@ class AudioStreamPlaybackMP3 : public AudioStreamPlaybackResampled {
 	AudioFrame loop_fade[FADE_SIZE];
 	int loop_fade_remaining = FADE_SIZE;
 
+	bool looping_override = false;
+	bool looping = false;
```
KoBeWi, line 51: This could be an enum. `looping_override = false` and `looping = true` is invalid state.
Though it's fine like this too.

modules/vorbis/audio_stream_ogg_vorbis.cpp
```diff
@@ -125,7 +126,7 @@ int AudioStreamPlaybackOggVorbis::_mix_internal(AudioFrame *p_buffer, int p_fram
 		if (!have_packets_left && !have_samples_left) {
 			// Actual end of file!
 			bool is_not_empty = mixed > 0 || vorbis_stream->get_length() > 0;
-			if (vorbis_stream->loop && is_not_empty) {
+			if (looping && is_not_empty) {
```
KoBeWi, line 129:
```suggestion
			if (beat_loop && is_not_empty) {
```
Though renaming the variable to `use_loop` like in MP3 would be more clear.

modules/minimp3/audio_stream_mp3.cpp
```diff
@@ -143,6 +145,24 @@ void AudioStreamPlaybackMP3::tag_used_streams() {
...
+			looping_override = false;
+			looping = false;
+		} else {
+			looping_override = true;
+			looping = p_value;
+		}
+	}
+}
+Variant AudioStreamPlaybackMP3::get_parameter(const StringName &p_name) const {
+	if (p_name == SNAME("looping") && looping_override) {
```
KoBeWi, line 160:
```suggestion
	if (looping_override && p_name == SNAME("looping")) {
```
Cheaper check first.
Same in OGG.

scene/2d/audio_stream_player_2d.cpp
```diff
@@ -36,6 +36,8 @@
 #include "scene/main/window.h"
 #include "scene/resources/world_2d.h"
 
+#define PARAM_PREFIX "parameters/"
```
KoBeWi, line 39: Can this be defined elsewhere, to avoid repeating it for every player type?
Maybe it could be `constexpr String` (or `char *` if String can't be used) in AudioStream?

scene/2d/audio_stream_player_2d.cpp
```diff
@@ -222,9 +224,34 @@ void AudioStreamPlayer2D::_update_panning() {
 	last_mix_count = AudioServer::get_singleton()->get_mix_count();
 }
 
+void AudioStreamPlayer2D::_update_stream_parameters() {
+	if (stream.is_valid()) {
```
KoBeWi, line 228:
```suggestion
	if (stream.is_null()) {
		return;
	}
```
Can return early.

scene/2d/audio_stream_player_2d.cpp
```diff
@@ -222,9 +224,34 @@ void AudioStreamPlayer2D::_update_panning() {
...
+		List<AudioStream::Parameter> parameters;
+		stream->get_parameter_list(&parameters);
+		for (const AudioStream::Parameter &K : parameters) {
+			PropertyInfo pi = K.property;
```
KoBeWi, line 232:
```suggestion
			const PropertyInfo &pi = K.property;
```

scene/2d/audio_stream_player_2d.cpp
```diff
@@ -430,6 +461,39 @@ void AudioStreamPlayer2D::_on_bus_renamed(int p_bus_index, const StringName &p_o
 	notify_property_list_changed();
 }
 
+bool AudioStreamPlayer2D::_set(const StringName &p_name, const Variant &p_value) {
+	if (playback_parameters.has(p_name)) {
+		ParameterData &pd = playback_parameters[p_name];
```
KoBeWi, line 466:
```suggestion
	HashMap<StringName, ParameterData>::Iterator I = playback_parameters.find(p_name);
	if (I) {
		ParameterData &pd = I->value;
```
Avoids double lookup.

scene/2d/audio_stream_player_2d.cpp
```diff
@@ -430,6 +461,39 @@ void AudioStreamPlayer2D::_on_bus_renamed(int p_bus_index, const StringName &p_o
...
+		pd.value = p_value;
+		for (int i = 0; i < stream_playbacks.size(); i++) {
+			stream_playbacks[i]->set_parameter(pd.path, pd.value);
```
KoBeWi, line 469:
```suggestion
		for (Ref<AudioStreamPlayback> &playback : stream_playbacks) {
			playback->set_parameter(pd.path, pd.value);
```

scene/2d/audio_stream_player_2d.cpp
```diff
@@ -430,6 +461,39 @@ void AudioStreamPlayer2D::_on_bus_renamed(int p_bus_index, const StringName &p_o
...
+		}
+		return true;
+	}
+
+	return false;
+}
+
+bool AudioStreamPlayer2D::_get(const StringName &p_name, Variant &r_ret) const {
+	if (playback_parameters.has(p_name)) {
```
KoBeWi, line 478: Same thing with iterator here.

scene/2d/audio_stream_player_2d.cpp
```diff
@@ -430,6 +461,39 @@ void AudioStreamPlayer2D::_on_bus_renamed(int p_bus_index, const StringName &p_o
...
+		r_ret = playback_parameters[p_name].value;
+		return true;
+	}
+	return false;
+}
+
+void AudioStreamPlayer2D::_get_property_list(List<PropertyInfo> *p_list) const {
+	if (stream.is_valid()) {
```
KoBeWi, line 486: This can technically also be an early return, though not sure if it fits this method.

KoBeWi:
I reviewed the PR from the editor side and it looks good (aside from the null bug mentioned above).

Though I don't like how this adds even more code duplication to AudioStreamPlayer classes. I think they can be refactored using some internal class that handles common code.
That said, remember that my AudioStreamPlayer2D comments apply to other player too.

Mickeon:
> Though I don't like how this adds even more code duplication to AudioStreamPlayer classes. I think they can be refactored using some internal class that handles common code.

Someday. Someday... They should eventually share the same interface. It would save a lot when it comes to code maintenance.
If https://github.com/godotengine/godot-proposals/issues/6416 were implemented even internally these 3 classes would probably be the ones that benefit it the most

KoBeWi:
It doesn't need traits.

Maybe I'll get onto this once this PR is merged.

doc/classes/AudioStream.xml
reduz: Thats a good point, will change to that.

reduz:
@KoBeWi I agree, I think it's time we eventually create a shared internal class for all of them, though I guess not on the scope for this PR.

modules/minimp3/audio_stream_mp3.cpp
reduz: That is okay because It's a checkable property (if Variant() it means the property is unused).

scene/2d/audio_stream_player_2d.cpp
reduz: I think we have to just fuse all 3 in a common internal class at this rate, I am not sure where I would place it.

scene/2d/audio_stream_player_2d.cpp
reduz: Variant is null depends on the parameter, not all stream parameters will be necessarily exposed as checkable.

reduz: Okay I figured it out, should not save it now.

reduz:
Okay I think this should be ready now.

modules/minimp3/audio_stream_mp3.cpp
KoBeWi: I mean Parameter constructor is defined like this
`Parameter(const PropertyInfo &p_info = PropertyInfo(), const Variant &p_default_value = Variant())`
The `variant()` you use here is part of the constructor and it's unnecessary.

KoBeWi:
Now the parameter is always null and can't be changed.

doc/classes/AudioStream.xml
```diff
@@ -28,6 +28,12 @@
 			<description>
 			</description>
 		</method>
+		<method name="_get_parameter_list" qualifiers="virtual const">
+			<return type="Array" />
```
adamscott, line 32:
```suggestion
			<return type="Dictionary[]" />
```

Based on the failing CI prompt.

ellenhp: Might make sense to also slightly tweak the description below after this change since "array contains dictionaries" will be redundant.

ellenhp (approved):
LGTM for audio. I have not built or tested so it would be good if someone else could confirm that they've played with it, but the approach seems sound and I think this'll make a lot of people really happy.

reduz:
@ellenhp Oh, very happy to see you back Ellen and thanks for the review! I still have a few bugs to fix, but should be able to get this one done soon.

modules/minimp3/audio_stream_mp3.cpp
```diff
@@ -46,7 +46,9 @@ int AudioStreamPlaybackMP3::_mix_internal(AudioFrame *p_buffer, int p_frames) {
 	int frames_mixed_this_step = p_frames;
 
 	int beat_length_frames = -1;
-	bool beat_loop = mp3_stream->has_loop() && mp3_stream->get_bpm() > 0 && mp3_stream->get_beat_count() > 0;
+	bool use_loop = looping_override ? looping : mp3_stream->loop;
```
adamscott, line 49: I don't know about this. Maybe the parameter for `looping` shouldn't be a binary, but rather an enum: DEFAULT, ON, OFF

In my mind, we should be able to let the looping parameter to DEFAULT and loop the sample if the stream itself is set to looping.

scene/2d/audio_stream_player_2d.cpp
adamscott: Maybe a .inc file for common classes?

scene/2d/audio_stream_player_2d.cpp
```diff
@@ -430,6 +463,44 @@ void AudioStreamPlayer2D::_on_bus_renamed(int p_bus_index, const StringName &p_o
...
+		for (Ref<AudioStreamPlayback> &playback : stream_playbacks) {
+			playback->set_parameter(pd.path, pd.value);
+		}
+		if (p_value.is_null()) {
+			playback_parameters.erase(p_name);
+		}
+		return true;
+	}
+
+	return false;
```
adamscott, line 480:
```suggestion
	if (!I) {
		return false;
	}
	
	ParameterData &pd = I->value;
	pd.value = p_value;
	for (Ref<AudioStreamPlayback> &playback : stream_playbacks) {
		playback->set_parameter(pd.path, pd.value);
	}
	if (p_value.is_null()) {
		playback_parameters.erase(p_name);
	}
	return true;
```

modules/minimp3/audio_stream_mp3.cpp
KoBeWi: The property itself is checkable, like theme overrides. So you can choose whether use the default or a custom value.

scene/3d/audio_stream_player_3d.cpp
```diff
@@ -818,6 +850,44 @@ void AudioStreamPlayer3D::_on_bus_renamed(int p_bus_index, const StringName &p_o
...
+		for (Ref<AudioStreamPlayback> &playback : stream_playbacks) {
+			playback->set_parameter(pd.path, pd.value);
+		}
+		if (p_value.is_null()) {
+			playback_parameters.erase(p_name);
+		}
+		return true;
+	}
+
+	return false;
```
adamscott, line 867: Same return early comment here.

scene/audio/audio_stream_player.cpp
```diff
@@ -88,9 +90,73 @@ void AudioStreamPlayer::_notification(int p_what) {
...
+		for (Ref<AudioStreamPlayback> &playback : stream_playbacks) {
+			playback->set_parameter(pd.path, pd.value);
+		}
+		if (p_value.is_null()) {
+			playback_parameters.erase(p_name);
+		}
+		return true;
+	}
+
+	return false;
```
adamscott, line 138: Same return early comment here.

modules/minimp3/audio_stream_mp3.cpp
adamscott: image

Just built the PR to make sure and you're right.

KoBeWi: Right, the original screenshot is outdated.

reduz:
> Now the parameter is always null and can't be changed.

Ah shit, I thought It worked, let me test again :laughing:

reduz:
Yeah nevermind, I was misled thinking it worked, can't do it like this. Guess the null will have to stay if unassigned, else the logic gets more hairy due to using checkable logic.

KoBeWi:
Maybe you can handle it using `_validate_property()`. The null only appears after toggling the parameter, so you get a new value that can't be removed anymore.

reduz:
@KoBeWi The problem is that this is very parameter dependant, and this null comes from a checkable parameter in the OGG/MP3 streams, other parameters may not necessarily be like this.
I could still add a hack in _validate_property and make a parameter not serializable if its null, but it feels too workaroundish and not sure if this is worth it.

scene/3d/audio_stream_player_3d.cpp
reduz: I think we will need to have a discussion about these things because I don't think we need to unify so much in style for this type of code. To me early return is not necessarily more readable always. I will do it for this PR, but I think we need to lax the review requirements a bit.

KoBeWi:
We'll get a report as soon as someone finds out a parameter in their scene that they can't remove.

scene/2d/audio_stream_player_2d.cpp
KoBeWi: #87061

Mickeon:
The fact that `null` is saved as is now needs to be addressed before this gets merged, otherwise it's not going to leave a great impression, in my opinion. The feature would feel underbaked right away. There's many cases of properties saving when they don't need to, and I wouldn't want this to contribute to the pile, either.

KoBeWi:
> I could still add a hack in _validate_property and make a parameter not serializable if its null, but it feels too workaroundish and not sure if this is worth it.

Actually this should check if the value is equal to the default value defined in the parameter.

KoBeWi (approved):
Well, aside from the null issue, looks fine.
I'll try to address it when I update #87061.

modules/minimp3/audio_stream_mp3.cpp
```diff
@@ -143,6 +145,25 @@ void AudioStreamPlaybackMP3::tag_used_streams() {
...
+	}
+}
+
+Variant AudioStreamPlaybackMP3::get_parameter(const StringName &p_name) const {
+	if (looping_override && p_name == SNAME("looping")) {
+		return looping;
+	}
+	return Variant();
+}
+
```
MJacred, line 166: In line 191, add this
```c++
mp3s->looping = loop;
```
to inherit looping setting from stream and be in sync with ogg vorbis.

MJacred (changes requested):
Aside from the suggested change, the audio side lgtm.

modules/minimp3/audio_stream_mp3.cpp
reduz: ah that line should actually be removed instead, since its ignored.

MJacred: Ah, I get it. If the overwrite is deactivated, the inspector can't read `looping` and the audio player mustn't read it.

modules/vorbis/audio_stream_ogg_vorbis.cpp
```diff
@@ -382,6 +402,7 @@ Ref<AudioStreamPlayback> AudioStreamOggVorbis::instantiate_playback() {
 	ovs->frames_mixed = 0;
 	ovs->active = false;
 	ovs->loops = 0;
+	ovs->looping = loop;
```
MJacred, line 405:
```suggestion
```

As noted in the previous review: this assignment is ignored anyway.

MJacred (changes requested):
Follow-up from previous review

reduz:
Alright, should be ready to merge once it passes CI.

akien-mga:
Thanks!

reduz:
:tada: 

Thanks all involved for giving me feedback on this!

Mickeon (commented):
This feedback has been in a **pending** state for a really, really long time because I forgot to submit it. I [mentioned it on RocketChat a while back](https://chat.godotengine.org/channel/audio?msg=CWBRzKpCZeSiYtvA6) and it was potentially also later discussed, offhandendly, in a meeting on April 7th 2025.


I thought that, while out-of-date, I should submit this review now for the sake of transparency.

Nichgon:
I noticed that `parameters/looping` is only available when stream (AudioStream) is from Ogg and MP3 but not for WAV. It would be great to have this parameter for WAV files as well.

PR #86476, opened 2023-12-23
Store horizontal and vertical split offsets separately in FileSystem dock

### LINKED ISSUE #83333: FileSystem dock should remember it's split offset between runs
### Godot version

v4.2.beta.custom_build [a574c0296]

### System information

linux

### Issue description

- The FileSystem split offset is not restored between restarts
- You can now split the FileSystem dock in horizontal mode in 4.2.
- The default split offset is unusable in horizontal mode
- It's important to remember the split offset value, because I set mine to match a certain number of item columns (otherwise there's an ugly blank space on the right). Having to set it up manually again (it's a pixel perfect setting and I really want it set in the right place) is a pain. But it's a minor pain for everyone, regardless of precision.
- It can be fixed with an addon, but it really should behave correctly by default (and even be saved in editor layouts maybe)

|before setting the offset|after setting the offset|after restarting the editor|
|-|-|-|
|Screenshot from 2023-10-14 15-50-35|Screenshot from 2023-10-14 15-50-39|Screenshot from 2023-10-14 15-50-46|

### Steps to reproduce

- set the split offset of the FileSystem
- restart the editor
- notice how the split is back to default
- try again with the FileSystem in horizontal mode, and notice that the default value is unusable

any

VedatGunel:
Fixes #83333

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -5047,7 +5047,8 @@ void EditorNode::_save_docks_to_config(Ref<ConfigFile> p_layout, const String &p
 
 	// Save FileSystemDock state.
 
-	p_layout->set_value(p_section, "dock_filesystem_split", FileSystemDock::get_singleton()->get_split_offset());
+	p_layout->set_value(p_section, "dock_filesystem_h_split_offset", FileSystemDock::get_singleton()->get_h_split_offset());
+	p_layout->set_value(p_section, "dock_filesystem_v_split_offset", FileSystemDock::get_singleton()->get_v_split_offset());
 	p_layout->set_value(p_section, "dock_filesystem_display_mode", FileSystemDock::get_singleton()->get_display_mode());
 	p_layout->set_value(p_section, "dock_filesystem_file_sort", FileSystemDock::get_singleton()->get_file_sort());
 	p_layout->set_value(p_section, "dock_filesystem_file_list_display_mode", FileSystemDock::get_singleton()->get_file_list_display_mode());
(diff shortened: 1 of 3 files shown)


editor/filesystem_dock.cpp
```diff
@@ -2588,6 +2586,10 @@ void FileSystemDock::_change_split_mode() {
 	emit_signal(SNAME("display_mode_changed"));
 }
 
+void FileSystemDock::_split_dragged(int p_offset) {
+	split_box->is_vertical() ? split_box_offset_v = p_offset : split_box_offset_h = p_offset;
```
AThousandShips, line 2590:
```suggestion
	if (split_box->is_vertical()) {
		split_box_offset_v = p_offset;
	} else {
		split_box_offset_h = p_offset;
	}
```
Don't do assignment in ternary, it's unsafe and has poor readability

editor/editor_node.cpp
```diff
@@ -5047,7 +5047,12 @@ void EditorNode::_save_docks_to_config(Ref<ConfigFile> p_layout, const String &p
 
 	// Save FileSystemDock state.
 
-	p_layout->set_value(p_section, "dock_filesystem_split", FileSystemDock::get_singleton()->get_split_offset());
+	if (unlikely(p_layout->has_section_key(p_section, "dock_filesystem_split"))) {
+		p_layout->erase_section_key(p_section, "dock_filesystem_split");
+	}
```
KoBeWi, line 5052: I think there is no reason to remove this value. It will just stay unused if it exists.

editor/filesystem_dock.cpp
```diff
@@ -3767,6 +3773,8 @@ FileSystemDock::FileSystemDock() {
 
 	split_box = memnew(SplitContainer);
 	split_box->set_v_size_flags(SIZE_EXPAND_FILL);
+	split_box->connect("dragged", callable_mp(this, &FileSystemDock::_split_dragged));
+	split_box_offset_h = 240 * EDSCALE;
```
KoBeWi, line 3777: Isn't that too big for a default value? I'd use 120.

VedatGunel: On the default layout they both look fine, but when the left panel is expanded, the folder tree stays shrinked when offset is small.
| offset = 120  | offset = 240 |
| ------------- | ------------- |
| offset120 | offset240 |

KoBeWi: Eh, maybe it's fine then.

KoBeWi (approved):

akien-mga:
Thanks!

PR #86479, opened 2023-12-24
Fix scene parser reading non-built-in scripts for POT generation

YeldhamDev:


akien-mga (approved):

akien-mga:
Wait, wasn't this intentional? It seemed useful to me that selecting a `.tscn` for POT generation would automatically parse its dependencies for used strings.

YeldhamDev:
It can cause unwanted strings to be parsed, which actually happens in my personal project. Making them need to be explicitly added seems like the better choice.

akien-mga:
Thanks!

PR #86486, opened 2023-12-24
Improve `EditorDirDialog`

timothyqiu:
* Automatically selects the newly created directory
* Automatically selects "res://" when nothing is selected
* Fixes an error when overwrite/replace dialog appears
* Changes "copy checkbox + action button" to "copy button + move button"
* Double clicking a directory (un)collapses it instead of copy/move
* Moves "Create Folder" to top row for better UX
* Uses DirectoryCreateDialog for "Create Folder"

| Before | After |
|---|---|
| ksnip_20231224-140532 | image |

### DIFF
--- a/editor/filesystem_dock.cpp
+++ b/editor/filesystem_dock.cpp
@@ -1883,10 +1883,6 @@ Vector<String> FileSystemDock::_check_existing() {
 	return conflicting_items;
 }
 
-void FileSystemDock::_move_dialog_confirm(const String &p_path) {
-	_move_operation_confirm(p_path, move_dialog->is_copy_pressed());
-}
-
 void FileSystemDock::_move_operation_confirm(const String &p_to_path, bool p_copy, Overwrite p_overwrite) {
 	if (p_overwrite == OVERWRITE_UNDECIDED) {
 		to_move_path = p_to_path;
@@ -2363,6 +2359,7 @@ void FileSystemDock::_file_option(int p_option, const Vector<String> &p_selected
 				}
 			}
 			if (to_move.size() > 0) {
+				move_dialog->config(p_selected);
 				move_dialog->popup_centered_ratio(0.4);
 			}
 		} break;
@@ -3847,7 +3844,8 @@ FileSystemDock::FileSystemDock() {
 
 	move_dialog = memnew(EditorDirDialog);
 	add_child(move_dialog);
-	move_dialog->connect("dir_selected", callable_mp(this, &FileSystemDock::_move_dialog_confirm));
+	move_dialog->connect("move_pressed", callable_mp(this, &FileSystemDock::_move_operation_confirm).bind(false, OVERWRITE_UNDECIDED));
+	move_dialog->connect("copy_pressed", callable_mp(this, &FileSystemDock::_move_operation_confirm).bind(true, OVERWRITE_UNDECIDED));
 
 	overwrite_dialog = memnew(ConfirmationDialog);
 	add_child(overwrite_dialog);
@@ -3884,7 +3882,7 @@ FileSystemDock::FileSystemDock() {
 
 	make_dir_dialog = memnew(DirectoryCreateDialog);
 	add_child(make_dir_dialog);
-	make_dir_dialog->connect("dir_created", callable_mp(this, &FileSystemDock::_rescan));
+	make_dir_dialog->connect("dir_created", callable_mp(this, &FileSystemDock::_rescan).unbind(1));
 
 	make_scene_dialog = memnew(SceneCreateDialog);
 	add_child(make_scene_dialog);
--- a/editor/gui/editor_dir_dialog.cpp
+++ b/editor/gui/editor_dir_dialog.cpp
@@ -30,22 +30,18 @@
 
 #include "editor_dir_dialog.h"
 
-#include "core/io/dir_access.h"
-#include "core/os/keyboard.h"
-#include "core/os/os.h"
+#include "editor/directory_create_dialog.h"
 #include "editor/editor_file_system.h"
-#include "editor/editor_scale.h"
-#include "editor/editor_settings.h"
-#include "scene/gui/check_box.h"
+#include "scene/gui/box_container.h"
 #include "scene/gui/tree.h"
 #include "servers/display_server.h"
 
 void EditorDirDialog::_update_dir(TreeItem *p_item, EditorFileSystemDirectory *p_dir, const String &p_select_path) {
 	updating = true;
 
-	String path = p_dir->get_path();
+	const String path = p_dir->get_path();
 
-	p_item->set_metadata(0, p_dir->get_path());
+	p_item->set_metadata(0, path);
 	p_item->set_icon(0, tree->get_editor_theme_icon(SNAME("Folder")));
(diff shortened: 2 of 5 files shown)


editor/gui/editor_dir_dialog.cpp
```diff
@@ -134,29 +129,35 @@ void EditorDirDialog::_item_collapsed(Object *p_item) {
 	}
 }
 
-void EditorDirDialog::_item_activated() {
-	_ok_pressed(); // From AcceptDialog.
+void EditorDirDialog::_item_selected() {
+	bool disable = tree->get_selected() == nullptr;
+	get_ok_button()->set_disabled(disable);
+	copy->set_disabled(disable);
+	makedir->set_disabled(disable);
```
KoBeWi, line 136: This one could default to `res://` if nothing is selected.

editor/filesystem_dock.cpp
```diff
@@ -1883,8 +1883,8 @@ Vector<String> FileSystemDock::_check_existing() {
 	return conflicting_items;
 }
 
-void FileSystemDock::_move_dialog_confirm(const String &p_path) {
-	_move_operation_confirm(p_path, move_dialog->is_copy_pressed());
+void FileSystemDock::_move_dialog_confirm(const String &p_path, bool p_copy) {
+	_move_operation_confirm(p_path, p_copy);
```
KoBeWi, line 1887: You could connect to this method directly. The intermediate one was only needed for getting the bool.

editor/gui/editor_dir_dialog.h
```diff
@@ -46,20 +46,22 @@ class EditorDirDialog : public ConfirmationDialog {
 	AcceptDialog *mkdirerr = nullptr;
 
 	Button *makedir = nullptr;
+	Button *copy = nullptr;
 	HashSet<String> opened_paths;
+	String new_dir_path;
 
 	Tree *tree = nullptr;
 	bool updating = false;
-	CheckBox *copy = nullptr;
```
KoBeWi, line 53: You should also remove CheckBox forward declaration.

editor/gui/editor_dir_dialog.cpp
```diff
@@ -185,13 +186,15 @@ void EditorDirDialog::_make_dir_confirm() {
...
+		new_dir_path = dir.path_join(stripped_dirname) + "/"; // Paths returned by scanning end with '/'.
 		EditorFileSystem::get_singleton()->scan_changes(); // We created a dir, so rescan changes.
 	}
 	makedirname->set_text(""); // reset label
 }
 
 void EditorDirDialog::_bind_methods() {
-	ADD_SIGNAL(MethodInfo("dir_selected", PropertyInfo(Variant::STRING, "dir")));
+	ADD_SIGNAL(MethodInfo("copy_selected", PropertyInfo(Variant::STRING, "dir")));
+	ADD_SIGNAL(MethodInfo("move_selected", PropertyInfo(Variant::STRING, "dir")));
```
KoBeWi, line 197:
```suggestion
	ADD_SIGNAL(MethodInfo("copy_pressed", PropertyInfo(Variant::STRING, "dir")));
	ADD_SIGNAL(MethodInfo("move_pressed", PropertyInfo(Variant::STRING, "dir")));
```

editor/gui/editor_dir_dialog.cpp
```diff
@@ -204,14 +207,14 @@ EditorDirDialog::EditorDirDialog() {
 	tree = memnew(Tree);
 	vb->add_child(tree);
 	tree->set_v_size_flags(Control::SIZE_EXPAND_FILL);
-	tree->connect("item_activated", callable_mp(this, &EditorDirDialog::_item_activated));
```
KoBeWi, line 207: This makes it impossible to accept with double click. Maybe it could default to Move? (and could copy if you hold Ctrl, idk)

editor/filesystem_dock.cpp
timothyqiu: The `connect()` does not support default argument. In order to connect to `_move_operation_confirm` directly, I have to bind `OVERRITE_UNDECIDED`. But in order to bind it, it seems I have to register the enum for binding. Since the class and enum is not exposed for scripting, I think doing this makes things more complicated.

zolrath:
This may not be important but in every other dialog and menu I can see in Godot, Cancel is the last/far right option.
However, in this menu, Cancel is a middle option.

Is there a UI/UX standard for the location of the Cancel button in Godot?

editor/filesystem_dock.cpp
timothyqiu: OK, I did the change. Note that `Overwrite` enum is now changed to public and bind.

editor/gui/editor_dir_dialog.cpp
timothyqiu: I changed double clicking to (un)collapse that directory. Consistent with the filesystem dock.

Previously, double clicking moves/copies depending on the last action performed (because "Copy Files(s)" state is kept between runs).

timothyqiu:
Moved "Create Folder" to the top row so that the Cancel button is no longer in the middle :P 

image

Also changed "Create Folder" to use `DirectoryCreateDialog`, the same as FileSystem dock.

image

editor/gui/editor_dir_dialog.cpp
```diff
@@ -59,13 +55,35 @@ void EditorDirDialog::_update_dir(TreeItem *p_item, EditorFileSystemDirectory *p
...
+	if (p_paths.size() == 1) {
+		String path = p_paths[0];
+		if (path.ends_with("/")) {
+			path = path.substr(0, path.length() - 1);
+		}
+		// TRANSLATORS: %s is the file name that will be moved or duplicated.
+		set_title(vformat(TTR("Move/Duplicate: %s"), path.get_file()));
+	} else {
+		// TRANSLATORS: %d is the number of files that will be moved or duplicated.
+		set_title(vformat(TTRN("Move/Duplicate %d Item", "Move/Duplicate %d Items", p_paths.size()), p_paths.size()));
```
KoBeWi, line 83: The single item string will be unused 🤔

timothyqiu: It's necessary in order to use `TTRN`. Using `TTRN` here for languages that have more complex plural forms.

editor/gui/editor_dir_dialog.cpp
```diff
@@ -59,13 +55,35 @@ void EditorDirDialog::_update_dir(TreeItem *p_item, EditorFileSystemDirectory *p
 		p_item->set_text(0, p_dir->get_name());
 	}
 
+	if (path == new_dir_path) {
+		p_item->select(0);
+	} else if (!p_item->get_parent()) {
```
KoBeWi, line 60:
```suggestion
	if (path == new_dir_path || !p_item->get_parent()) {
```

KoBeWi (approved):

editor/gui/editor_dir_dialog.cpp
KoBeWi: Yep, I get why TTRN is here. I wonder if there is a way to avoid useless string.

KoBeWi: Actually nevermind 🤦‍♂️

akien-mga:
Thanks!

PR #86498, opened 2023-12-25
Add `rand_bool(true_chance)` function

### LINKED ISSUE #8721: Add randb() function to return a random boolean

Add a function called "randb()" that you can call anywhere to return a random boolean. This would go hand-in-hand with randi() and randf().

I do not have a quick and easy single tab-press way to return a random boolean. This interrupts my workflow and efficiency.

This simply returns a random boolean the same way randi() returns a random integer. It is a very simple solution to a workflow interruption.

### Usage
`var random_bool = randb()`

### Implementation
I have already built and compiled a [forked branch](https://github.com/WiseNoodle/godot/tree/add-randb-function) using the same implementation as randi() or randf() like so:

`_FORCE_INLINE_ bool randb() {
		return (rand() & 0x1) == 0x1;
	}`
	

This would be used very often as it is only one key press to use however, it can also be implemented using this line of code:

`randf()>0.5`

This could be made into an add-on to keep the engine lean however, I feel it should be a core component since it completes the set of tools for random generation.

WiseNoodle:
This is the implementation for adding a function to return a random boolean as in this [proposal](https://github.com/godotengine/godot-proposals/issues/8721).

### Usage
It can be used in-line, anywhere in your scripts, however you like. It works the same as `randi()` or `randf()`.

For example the line `var random_bool = randb()` would assign the variable `random_bool` an either true or false value at random.

* _Bugsquad edit:_ Closes godotengine/godot-proposals#8721.


sairam4123:
Why not just `int(randf())`?

AThousandShips:
That would yield a very very small chance of false and almost only true, not useful

WiseNoodle:
I added the new changes and verified that it works with the standard build; I need someone to compile the C# for me and test it out. My C# SDKs are broken on my machine.

clayjohn:
This looks well done, but I'm not sure if it's something we need to have in core. I appreciate the benefit of having this auto complete when you just want a random bool, but in the end, I'm not sure it really saves anything over using `randf() < x`

At the same time, the actual implementation is quite trivial and doesn't bloat the API much. So I don't have a strong opinion one way or the other. 

If there are a few other people who really want this it could be worth merging.

core/math/random_number_generator.h
```diff
@@ -52,6 +52,7 @@ class RandomNumberGenerator : public RefCounted {
 	_FORCE_INLINE_ void randomize() { randbase.randomize(); }
 
 	_FORCE_INLINE_ uint32_t randi() { return randbase.rand(); }
+	_FORCE_INLINE_ bool randb(real_t chance = 0.5) { return randbase.randb(chance); }
```
AThousandShips, line 55:
```suggestion
	_FORCE_INLINE_ bool randb(real_t p_chance = 0.5) { return randbase.randb(p_chance); }
```
Prefix arguments with `p_`

doc/classes/RandomNumberGenerator.xml
```diff
@@ -17,6 +17,18 @@
...
 	</tutorials>
 	<methods>
+		<method name="randb">
+			<return type="bool" />
+			<param index="0" name="chance" type="float" default="0.5" />
+			<description>
+				Returns a random boolean with a bias of [param chance] to be [code]true[/code] and [code]false[/code] otherwise.
+				[codeblocks]
+				[gdscript]
+				randb(0.5) # Returns true at the param chance of 0.5
```
AThousandShips, line 27:
```suggestion
				randb(0.5) # Returns true at the param chance of 0.5.
```

doc/classes/@GlobalScope.xml
```diff
@@ -1006,6 +1006,18 @@
...
 			</description>
 		</method>
+		<method name="randb">
+			<return type="bool" />
+			<param index="0" name="chance" type="float" />
+			<description>
+				Returns a random boolean with a bias of [param chance] to be [code]true[/code] and [code]false[/code] otherwise.
+				[codeblocks]
+				[gdscript]
+				randb(0.5) # Returns true at the param chance of 0.5
```
AThousandShips, line 1016:
```suggestion
				randb(0.5) # Returns true at the param chance of 0.5.
```

core/math/random_pcg.h
```diff
@@ -87,6 +87,10 @@ class RandomPCG {
 		return pcg32_boundedrand_r(&pcg, bounds);
 	}
 
+	_FORCE_INLINE_ bool randb(double chance) {
```
AThousandShips, line 90:
```suggestion
	_FORCE_INLINE_ bool randb(double p_chance) {
```

core/math/math_funcs.h
```diff
@@ -549,6 +549,7 @@ class Math {
 	static void randomize();
 	static uint32_t rand_from_seed(uint64_t *seed);
 	static uint32_t rand();
+	static bool randb(double chance);
```
AThousandShips, line 552:
```suggestion
	static bool randb(double p_chance);
```

modules/mono/glue/runtime_interop.cpp
```diff
@@ -1250,6 +1250,10 @@ uint32_t godotsharp_randi() {
 	return Math::rand();
 }
 
+bool godotsharp_randb(double chance) {
```
AThousandShips, line 1253:
```suggestion
bool godotsharp_randb(double p_chance) {
```

AThousandShips (dismissed):
LGTM, needs more discussion but with the bias I'm in favour of this addition

core/math/random_number_generator.h
WiseNoodle: Are these changes I need to make on my side and recommit or did you make them already using GitHub magic?

AThousandShips: You need to make them 🙂, they are suggestions that can be applied but they're not complete and you need to fix the uses of the arguments and such

modules/mono/glue/GodotSharp/GodotSharp/Core/GD.cs
```diff
@@ -455,6 +455,15 @@ public static uint Randi()
             return NativeFuncs.godotsharp_randi();
         }
 
+        /// <summary>
+        /// Returns a random boolean.
+        /// </summary>
+        /// <returns>A random <see langword="bool"/>.</returns>
+        public static bool Randb(double chance)
```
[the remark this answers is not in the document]
WiseNoodle: Can you please re-review these two?

modules/mono/glue/GodotSharp/GodotSharp/Core/GD.cs
```diff
@@ -455,6 +455,15 @@ public static uint Randi()
             return NativeFuncs.godotsharp_randi();
         }
 
+        /// <summary>
+        /// Returns a random boolean with a bias of <paramref name="chance"/> to be <see langword="true"/> and <see langword="false"/> otherwise.
+        /// </summary>
+        /// <returns>A random <see langword="bool"/>.</returns>
+        public static bool Randb(double p_chance = 0.5)
```
raulsntos, line 462: We don't use `p_` prefixes in the C# API.

```suggestion
        public static bool Randb(double chance = 0.5)
```

WiseNoodle: :0 Am I supposed to just know this stuff out of the box? Did I miss something in the contributing docs? 0.o (just pushed the new commit btw)

AThousandShips: It's just something to learn 🙂, but the style of the rest of the code is usually a good reference

WiseNoodle:
I wouldn't know what else to check for however, I'll continue to fix any changes as needed.

joaoh82 (approved):

MewPurPur:
I think `rand_bool()` is a better name, as we don't really have any convention for bools, unlike ints and floats which have dozens of methods ending in "f" or "i".

WiseNoodle:
> I think `rand_bool()` is a better name, as we don't really have any convention for bools, unlike ints and floats which have dozens of methods ending in "f" or "i" (moreover they are interchangeable). This would make it possible to understand the function even if you're just looking at the code and aren't familiar with it.

You have a good point, I'm happy to make these changes if needed!

Mickeon:
I'm going to propose better documentation when I have time. It seems like the maintainers are fine with this function, after all.

AThousandShips:
This was discussed and the preferred naming landed on `rand_bool`, with the additional suggestion of calling the parameter `true_chance` :)

doc/classes/@GlobalScope.xml
```diff
@@ -995,6 +995,18 @@
...
+		<method name="randb">
+			<return type="bool" />
+			<param index="0" name="chance" type="float" />
+			<description>
+				Returns a random boolean with a bias of [param chance] to be [code]true[/code] and [code]false[/code] otherwise.
+				[codeblocks]
+				[gdscript]
+				randb(0.5) # Returns true at the param chance of 0.5.
+				[/gdscript]
+				[/codeblocks]
```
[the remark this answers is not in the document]
AThousandShips:
```suggestion
				Returns either [code]true[/code] or [code]false[/code] at random. [param true_chance] should be a number between [code]0.0[/code] and [code]1.0[/code] (inclusive). The greater [param true_chance] is, the more likely this function returns [code]true[/code].
				[codeblocks]
				[gdscript]
				rand_bool(0.0) # Returns false, all the time.
				rand_bool(0.5) # Returns true, 50% of the time.
				rand_bool(1.0) # Returns true, all the time.
				[/gdscript]
				[csharp]
				GD.RandBool(0.0); // Returns False, all the time.
				GD.RandBool(0.5); // Returns True, 50% of the time.
				GD.RandBool(1.0); // Returns True, all the time.
				[/csharp]
				[/codeblocks]
```
With the new range(s)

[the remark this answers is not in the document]
kleonc: Note that given the current implementation in this PR `randb(1.0)` can return `false`. This will happen when `randf()` returns `1.0`. It's very unlikely but possible.

For a guaranteed true one would need to call `randb(chance)` with `chance > 1.0`.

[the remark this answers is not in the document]
AThousandShips: I'd say to just change it to `randf() <= p_true_chance`, the actual difference *except* for the upper end doesn't really matter (assuming `randf` is *guaranteed* to produce strictly in [0,1])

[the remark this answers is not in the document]
MewPurPur: Doesn't seem too bad to just directly return true if the passed var is >= 1.0. While it's an extra comparison it does allow to skip the RNG routine.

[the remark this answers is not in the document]
WiseNoodle: I do agree with having a check to prevent unnecessary use of RNG. It is an easy addition.

Should I refactor with the above changes and new method/variable naming?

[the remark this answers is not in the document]
AThousandShips: Please do :) I think a warning isn't needed, there isn't one for `rand_range`

WiseNoodle:
My bad, I'm struggling to fix my side; things are kinda borked. I have all the code ready I just need to clean up the mess I made making it. I accidently closed the pr.

modules/mono/glue/GodotSharp/GodotSharp/Core/GD.cs
```diff
@@ -455,6 +455,15 @@ public static uint Randi()
             return NativeFuncs.godotsharp_randi();
         }
 
+        /// <summary>
+        /// Returns a random boolean with a bias of <paramref name="true_chance"/> to be <see langword="true"/> and <see langword="false"/> otherwise.
```
Mickeon, line 459: Would be nice for this to match `@GlobalScope.xml` now.

Mickeon (commented):
Documentation is... good enough by the way.

doc/classes/@GlobalScope.xml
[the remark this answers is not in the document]
Mickeon: If this was solved before, it got reverted now to an extent.

PR #86507, opened 2023-12-25
Fix MSVC 14.1 complaining about "attempting to reference a deleted function" in RenderPassClearValue

### LINKED ISSUE #86443: Build failing in MSVC 2017
### Tested versions

Introduced in #83452

### System information

Windows, MSVC 2017

### Issue description

This union is causing an issue:
```cpp
	union RenderPassClearValue {
		Color color = {};
		struct {
			float depth;
			uint32_t stencil;
		};
	};
```
Since `Color` does not have a default constructor, it results in the following message:
```
servers/rendering/rendering_device_driver.h(538): note: 'RenderingDeviceDriver::RenderPassClearValue::RenderPassClearValue(void)': function was implicitly deleted because 'RenderingDeviceDriver::RenderPassClearValue' has a variant data member 'RenderingDeviceDriver::RenderPassClearValue::color' with a non-trivial default constructor
```

### Steps to reproduce

1. Install MSVC 2017
2. Clone master
3. Build using SCons

N/A

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/86443

We could also add something like `new (&color) Color()` inside the ctor, but it doesn't seems very necessary.

### DIFF
--- a/servers/rendering/rendering_device_driver.h
+++ b/servers/rendering/rendering_device_driver.h
@@ -535,6 +535,8 @@ public:
 			float depth;
 			uint32_t stencil;
 		};
+
+		RenderPassClearValue() {}
 	};
 
 	struct AttachmentClear {


RandomShaper (commented):
What does it happen if instead of that change, the initializer is removed from `Color`, that, after all, zero-initializes internally?

jsjtxietian:
> zero-initializes internally

@RandomShaper  Sorry I don't quite understand,  where do we put the zero-initializes? In the constructor?

image

https://godbolt.org/z/5qGvGcof8

clayjohn:
> What does it happen if instead of that change, the initializer is removed from `Color`, that, after all, zero-initializes internally?

@RandomShaper Still doesn't work because Color doesn't have a trivial constructor (at least that's what the error message says)

clayjohn (approved):
Seems fine to me. @RandomShaper should have a chance to take another look before merging

RandomShaper (approved):
Oh, I see. The compiler deletes the constructor to be safe, so an explicit constructor is needed. Makes sense! My only (soft) suggestion now would be to use the `= default` syntax, but I'm fine with either.

jsjtxietian:
> My only (soft) suggestion now would be to use the `= default` syntax, but I'm fine with either.

lol that won't work too. see https://godbolt.org/z/TY34Ph54G

So I guess it should stay as it is now

akien-mga:
Thanks!

PR #86516, opened 2023-12-26
Fix incorrect `CAMERA_DIRECTION_WORLD` calculation

### LINKED ISSUE #86502: CAMERA_DIRECTION_WORLD shader built-ins does not yield the expected results
### Tested versions

Reproduceable in v4.2.1.stable.official [b09f793f5]. Looking at the history, it looks like this has never been correct.

### System information

Godot v4.2.1.stable - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3060 Laptop GPU (NVIDIA; 31.0.15.4633) - 12th Gen Intel(R) Core(TM) i7-12700H (20 Threads)

### Issue description

CAMERA_DIRECTION_WORLD is filled with the incorrect vector, and appears to have been so when first implemented.

	actions.renames["CAMERA_DIRECTION_WORLD"] = "scene_data.view_matrix[3].xyz";

### Steps to reproduce

Create a Camera3D at the origin facing in the negative Z direction. Make a child plane facing the camera, and give it a shader material with the following fragment shader:

```
void fragment () {
    ALBEDO = (normalize(CAMERA_DIRECTION_WORLD) + vec3(1.0f)) / 2.0f;
}
```

Translate the camera without rotating it. See that the texture changes color despite the camera's world direction not changing and the other operations being constant.

Replace the shader with the following:

```
void fragment () {
    vec3 dir = normalize(INV_VIEW_MATRIX * vec4(0.0f,0.0f,1.0f,0.0f)).xyz;
    ALBEDO = (normalize(dir) + vec3(1.0f)) / 2.0f;
}
```

Translate the camera without rotation and notice that the color does not change as expected. Rotate the camera and note that the color does change.

I believe the relevant line in godot/drivers/gles3/storage/material_storage.cpp should be:
'actions.renames["CAMERA_DIRECTION_WORLD"] = "scene_data.inv_view_matrix[2].xyz";`

[mgp.zip](https://github.com/godotengine/godot/files/13764378/mgp.zip)

WASD to move. QE to turn left and right.
Feel free to delete the noise floor, but I find it helps me be sure that the camera is actually moving/rotation when necessary.

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/86502

### DIFF
--- a/drivers/gles3/storage/material_storage.cpp
+++ b/drivers/gles3/storage/material_storage.cpp
@@ -1272,7 +1272,7 @@ MaterialStorage::MaterialStorage() {
 
 		actions.renames["NODE_POSITION_WORLD"] = "model_matrix[3].xyz";
 		actions.renames["CAMERA_POSITION_WORLD"] = "scene_data.inv_view_matrix[3].xyz";
-		actions.renames["CAMERA_DIRECTION_WORLD"] = "scene_data.view_matrix[3].xyz";
+		actions.renames["CAMERA_DIRECTION_WORLD"] = "scene_data.inv_view_matrix[2].xyz";
 		actions.renames["CAMERA_VISIBLE_LAYERS"] = "scene_data.camera_visible_layers";
 		actions.renames["NODE_POSITION_VIEW"] = "(scene_data.view_matrix * model_matrix)[3].xyz";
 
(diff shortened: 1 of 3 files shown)


acarx (approved):
Tested in my custom billboard shader, `INV_VIEW_MATRIX[2].xyz` is the right value.

clayjohn (approved):

PR #71780, opened 2023-01-21
Proposed fix for volume ramp based fade-in cutting off the beginning of sounds

ellenhp:
Proposed fix for #66452. Implements an explicit fade-in based on the same exponential method used for fade-out.

*Bugsquad edit:*
- Fixes #66452

PR #86517, opened 2023-12-26
Improve AudioStream looping

reduz:
* AudioStream (OGG/MP3) import dialog now allows to optionally set loop in beats.
* Loop points other than zero now properly fade-in to avoid clicks. 
* Stream import dialog now indicates the beat where loop occurs.
* Fixed the preview generation in the importer dialog to not show the looped part.
* Cleaned up the import dialog a bit.

Screenshot:

image

### DIFF
--- a/editor/import/audio_stream_import_settings.cpp
+++ b/editor/import/audio_stream_import_settings.cpp
@@ -87,7 +87,7 @@ void AudioStreamImportSettingsDialog::_draw_preview() {
 	Size2 rect_size = rect.size;
 	int width = rect_size.width;
 
-	Ref<AudioStreamPreview> preview = AudioStreamPreviewGenerator::get_singleton()->generate_preview(stream);
+	Ref<AudioStreamPreview> preview = AudioStreamPreviewGenerator::get_singleton()->generate_preview(stream_no_loop);
 	float preview_offset = zoom_bar->get_value();
 	float preview_len = zoom_bar->get_page();
 
@@ -104,6 +104,7 @@ void AudioStreamImportSettingsDialog::_draw_preview() {
 	float inactive_from = 1e20;
 	float beat_size = 0;
 	int last_beat = 0;
+	int loop_beat = 0;
 	if (stream->get_bpm() > 0) {
 		beat_size = 60 / float(stream->get_bpm());
 		int y_ofs = beat_font->get_height(main_size) + 4 * EDSCALE;
@@ -113,6 +114,10 @@ void AudioStreamImportSettingsDialog::_draw_preview() {
 		if (stream->get_beat_count() > 0) {
 			last_beat = stream->get_beat_count();
 			inactive_from = last_beat * beat_size;
+		}
+
+		if (loop->is_pressed() && loop_offset_type->get_selected() == 1) {
+			loop_beat = loop_offset->get_value();
 		}
 	}
 
@@ -136,6 +141,7 @@ void AudioStreamImportSettingsDialog::_draw_preview() {
 	if (beat_size) {
 		Color beat_color = Color(1, 1, 1, 1);
 		Color final_beat_color = beat_color;
+		Color loop_beat_color = beat_color * Color(1, 0.5, 0.5, 1.0);
 		Color bar_color = beat_color;
 		beat_color.a *= 0.4;
 		bar_color.a *= 0.6;
@@ -164,6 +170,8 @@ void AudioStreamImportSettingsDialog::_draw_preview() {
 					// Darken subsequent beats
 					beat_color.a *= 0.3;
 					color_active.a *= 0.3;
+				} else if (beat == loop_beat && loop_beat != 0) {
+					_preview->draw_rect(Rect2i(i, rect.position.y, 2, rect.size.height), loop_beat_color);
 				} else {
 					_preview->draw_rect(Rect2i(i, rect.position.y, 1, rect.size.height), (beat % bar_beats) == 0 ? bar_color : beat_color);
 				}
@@ -174,7 +182,7 @@ void AudioStreamImportSettingsDialog::_draw_preview() {
 }
 
 void AudioStreamImportSettingsDialog::_preview_changed(ObjectID p_which) {
-	if (stream.is_valid() && stream->get_instance_id() == p_which) {
+	if (stream_no_loop.is_valid() && stream_no_loop->get_instance_id() == p_which) {
 		_preview->queue_redraw();
 	}
 }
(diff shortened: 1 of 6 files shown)


MJacred:
@reduz: please make the PR title a little more explicit, e.g. "Improve AudioStream looping when using BPM logic", as looping in seconds stays unaffected

modules/minimp3/audio_stream_mp3.cpp
MJacred, line 1: `AudioStreamPlaybackMP3::start` should reset `loop_fade_remaining` to `FADE_SIZE`

```c++
void AudioStreamPlaybackMP3::start(double p_from_pos) {
        active = true;
        seek(p_from_pos);
        loops = 0;
        begin_resample();
}
```

to match `AudioStreamPlaybackOggVorbis`, and avoid potential fade leaking if stream is stopped during fading and then restarted

```c++
void AudioStreamPlaybackOggVorbis::start(double p_from_pos) {
        ERR_FAIL_COND(!ready);
        loop_fade_remaining = FADE_SIZE;
```

MJacred: Nevermind, `AudioStreamPlayer::play` instantiates a new playback

reduz:
@MJacred It also fixes the clipping of regular loops when position is not zero.

ellenhp:
(on mobile so I can't review fully) Does this also fade in if you call `play(nonzero)`? If so we can remove the fade in logic in the audio server and fix the open bug about beginnings of sounds being cut off.

MJacred:
@reduz: `loop_fade_remaining` is only set to a value < FADE_SIZE in an if-clause, which depends on `vorbis_stream->get_bpm() > 0` and `beat_length_frames >= 0`...  
There is no fading in ogg/mp3 on looping if BPM is not used

MJacred:
@ellenhp: on `play(any_value)`, a new AudioStreamPlayback is instantiated. At least, there is no loop fading in ogg vorbis and mp3. The audio server only ever does a fade out. So afaik no fading here

ellenhp:
Unless someone removed that logic since #71780 it's definitely still in there. Imo we should move that fade to be triggered by a seek and switch it from exponential to logistic. I'm not sure if it's been suggested but I want to preemptively say I strongly disagree with moving fade-*out* logic into the playbacks because that'll introduce pops when an audio stream player is removed from the tree without being stopped. Fade-in and loop fades (both triggered by a seek operation) can be moved to playbacks without introducing bugs I think.

MJacred:
Ok, you're right. I relied on variable naming. So if there's just some non-descriptive (for me at least) math going on, I missed it.

ellenhp:
Yeah it's not documented well which is on me.

MJacred:
There's no developer who hasn't neglected documentation before. But you can dump your knowledge here https://github.com/godotengine/godot-docs/pull/7896 (or just reference valuable PRs/proposals/issues, where I can read up) and we'll have something that sticks around :D  (_PR hijacking over and out_)

akien-mga:
The docs need to be added for the new members:
```diff
 diff --git a/modules/minimp3/doc_classes/ResourceImporterMP3.xml b/modules/minimp3/doc_classes/ResourceImporterMP3.xml
index a84c51c..e46285b 100644
--- a/modules/minimp3/doc_classes/ResourceImporterMP3.xml
+++ b/modules/minimp3/doc_classes/ResourceImporterMP3.xml
@@ -33,5 +33,7 @@
 			Only has an effect if [member loop] is [code]true[/code].
 			A more convenient editor for [member loop_offset] is provided in the [b]Advanced Import Settings[/b] dialog, as it lets you preview your changes without having to reimport the audio.
 		</member>
+		<member name="loop_offset_type" type="int" setter="" getter="" default="0">
+		</member>
 	</members>
 </class>
diff --git a/modules/vorbis/doc_classes/ResourceImporterOggVorbis.xml b/modules/vorbis/doc_classes/ResourceImporterOggVorbis.xml
index dd6c181..8f225e8 100644
--- a/modules/vorbis/doc_classes/ResourceImporterOggVorbis.xml
+++ b/modules/vorbis/doc_classes/ResourceImporterOggVorbis.xml
@@ -49,5 +49,7 @@
 			Only has an effect if [member loop] is [code]true[/code].
 			A more convenient editor for [member loop_offset] is provided in the [b]Advanced Import Settings[/b] dialog, as it lets you preview your changes without having to reimport the audio.
 		</member>
+		<member name="loop_offset_type" type="int" setter="" getter="" default="0">
+		</member>
 	</members>
 </class>
```
(and the descriptions written)

editor/import/audio_stream_import_settings.cpp
```diff
@@ -541,13 +575,19 @@ AudioStreamImportSettingsDialog::AudioStreamImportSettingsDialog() {
...
+
 	loop_offset = memnew(SpinBox);
 	loop_offset->set_max(10000);
 	loop_offset->set_step(0.001);
-	loop_offset->set_suffix("sec");
 	loop_offset->set_tooltip_text(TTR("Loop offset (from beginning). Note that if BPM is set, this setting will be ignored."));
 	loop_offset->connect("value_changed", callable_mp(this, &AudioStreamImportSettingsDialog::_settings_changed).unbind(1));
 	loop_hb->add_child(loop_offset);
+	loop_offset_type = memnew(OptionButton);
+	loop_offset_type->add_item(TTR("Sec"));
```
akien-mga, line 586: We don't seem to width constrained here so I would avoid abbreviations, which can be hard to translate.
```suggestion
	loop_offset_type->add_item(TTR("Seconds"));
```

akien-mga (dismissed):
Code and rationale seem fine to me. Didn't test.

reduz:
Will do the changes suggested ASAP.

rossunger:
I'm not sure what the progress on this is, but I would like to add a feature that allows you to set "loop_start" AND "loop_end" 

This is necessary for making a sample player, where you have a "note on" and "note off".
On note on, you play from the start of the audio until the "loop_end" at which point to you loop back to "loop_start" and continue looping until "note off", at which point you jump to "loop_end" and play until the end of the audio.

I'm ok to make a separate class for this, but I'd like to integrate with what's happening here if possible

romgerman:
> I'm not sure if it's outside the scope of this PR, but AFAIK, MP3 files tend of have a little bit of silence at the beginning of files ([more info here](https://lame.sourceforge.io/tech-FAQ.txt)). This can mess up the alignment of beats in the importer. Could this be taken into account? Maybe by marking the actual start of the audio, and marking beats from that point, instead of the beginning of the audio file. I think this would require figuring out the length of the zero padding, which I think varies.

WAV has an option to trim the silence in the import settings. Maybe MP3 can also have an option like this?
image

Akosmo:
> > I'm not sure if it's outside the scope of this PR, but AFAIK, MP3 files tend of have a little bit of silence at the beginning of files ([more info here](https://lame.sourceforge.io/tech-FAQ.txt)). This can mess up the alignment of beats in the importer. Could this be taken into account? Maybe by marking the actual start of the audio, and marking beats from that point, instead of the beginning of the audio file. I think this would require figuring out the length of the zero padding, which I think varies.
> 
> WAV has an option to trim the silence in the import settings. Maybe MP3 can also have an option like this? image

I've thought about it, and it may help, but there may be a risk of trimming quiet parts at the beginning of the audio... but to be honest, I haven't used that feature in Godot yet, and I have no idea how it's implemented.
I also wonder if trimming that part of the audio may cause problems during decoding.

PR #86518, opened 2023-12-26
[Core] Add iteration support to `Array`

AThousandShips:
Added simple iteration support for `Array`, using the same method as the subscript operator, added the read-only part to the constant one as well as well for parity with that operator

Many, *many*, possible uses across the codebase but kept this restricted to a few cases, considering doing one of my sweeping PRs to improve use of `foreach` generally not just for this new addition, but will have to wait for right now

Put some of the methods in `variant.h` as they depend on `Variant` being declared and wanted them to be inlineable

Not tagging this as `Performance` as any improvement would be minimal, but it does, theoretically, carry a minor performance improvement as it avoids doing COW checks on each access, which would most of the time be a wasted check as the `Vector` is *almost always* not shared yet the mutation check is always performed

Will squash the added cases once decided in case some additional "good to add" cases are suggested

### DIFF
--- a/core/config/project_settings.cpp
+++ b/core/config/project_settings.cpp
@@ -1330,8 +1330,8 @@ void ProjectSettings::load_scene_groups_cache() {
 		for (const String &E : scene_paths) {
 			Array scene_groups = cf->get_value(E, "groups", Array());
 			HashSet<StringName> cache;
-			for (int i = 0; i < scene_groups.size(); ++i) {
-				cache.insert(scene_groups[i]);
+			for (const Variant &scene_group : scene_groups) {
+				cache.insert(scene_group);
 			}
 			add_scene_groups_cache(E, cache);
 		}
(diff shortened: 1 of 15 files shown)


AThousandShips:
I've been investigating an equivalent addition to `Dictionary` but it's a bit more complex due to the internal data, but will open one if that one pans out as well

dsnopek:
Thanks, this looks great to me!

> Will squash the added cases once decided in case some additional "good to add" cases are suggested

I think having only these few cases is fine, in fact, reviewing those changes is trickier (at least for me) than the added iterator code, so in my opinion "less is more" here :-)

Calinou:
I think once we have added this to `List`, we'll be able to use for-range looping almost everywhere where it makes sense.

AThousandShips:
`List` already supports this 🙂

dsnopek (approved):
This looks great to me!

I attempted to really scrutinize the changes to existing code to start using the iterator, and didn't notice anything that seemed like it would change behavior.

I'm not on the core team but decided to be bold and hit "approve" - I really want to be able to use this myself :-)

akien-mga:
Thanks!

AThousandShips:
Thank *you*!

PR #86520, opened 2023-12-26
Fix `ScriptLanguageExtension::_find_function` argument names

touilleMan:
`ScriptLanguageExtension::_find_function` documentation is currently wrong when compared to it GDScript implementation: 
https://github.com/godotengine/godot/blob/13a0d6e9b253654f5cc2a44f3d0b3cae10440443/modules/gdscript/gdscript_editor.cpp#L212

### DIFF
--- a/core/object/script_language_extension.cpp
+++ b/core/object/script_language_extension.cpp
@@ -107,7 +107,7 @@ void ScriptLanguageExtension::_bind_methods() {
 	GDVIRTUAL_BIND(_supports_builtin_mode);
 	GDVIRTUAL_BIND(_supports_documentation);
 	GDVIRTUAL_BIND(_can_inherit_from_file);
-	GDVIRTUAL_BIND(_find_function, "class_name", "function_name");
+	GDVIRTUAL_BIND(_find_function, "function", "code");
 	GDVIRTUAL_BIND(_make_function, "class_name", "function_name", "function_args");
 	GDVIRTUAL_BIND(_open_in_external_editor, "script", "line", "column");
 	GDVIRTUAL_BIND(_overrides_external_editor);
--- a/doc/classes/ScriptLanguageExtension.xml
+++ b/doc/classes/ScriptLanguageExtension.xml
@@ -114,9 +114,10 @@
 		</method>
 		<method name="_find_function" qualifiers="virtual const">
 			<return type="int" />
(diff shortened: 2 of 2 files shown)


AThousandShips (changes requested):
This can't be changed like this, the binding is here:
https://github.com/godotengine/godot/blob/13a0d6e9b253654f5cc2a44f3d0b3cae10440443/core/object/script_language_extension.cpp#L110

Not in `GDScript`, which *isn't* a `ScriptLanguageExtension` but a `ScriptLanguage` in the first place

The discrepancy is an issue, but this will not be a valid fix as the documentation generation won't match

AThousandShips:
The problem here is that I don't know that this can be changed without breaking compatibility with extensions, I'm not sure if these bound arguments are compatibility relevant for other languages and binds

touilleMan:
> This can't be changed like this, the binding is here:

You're right I should have modified this part too ! (I've updated my PR)

> Not in GDScript, which isn't a ScriptLanguageExtension but a ScriptLanguage in the first place

True, I was pointing out `GDScript` given it is the standard `ScriptLanguage`, and `ScriptLanguageExtension` is just a wrapper over `ScriptLanguage`.

The keypoint is that `ScriptLanguageExtension` exposes a `_find_function` method that is used to implement `ScriptLanguage::find_function`:
https://github.com/godotengine/godot/blob/13a0d6e9b253654f5cc2a44f3d0b3cae10440443/core/object/script_language_extension.h#L374

So in the end, `ScriptLanguageExtension::_find_function` should have the same signature as `ScriptLanguage::find_function`.

> The problem here is that I don't know that this can be changed without breaking compatibility with extensions

The fix is only on the name of the parameters, are you sure this breaks compatibility ?

AThousandShips:
> The fix is only on the name of the parameters, are you sure this breaks compatibility ?

I'm not sure, that's why I said "I don't know if this can be done without breaking compatibility" 🙂

CC @dsnopek

dsnopek (approved):
> I don't know that this can be changed without breaking compatibility

Yeah, I think this is fine! It doesn't change how the arguments are used, only their names. From the perspective of GDExtension, arguments are only positional, the names aren't significant.

AThousandShips (approved):

akien-mga:
Thanks!

PR #86526, opened 2023-12-26
Fix CollisionObject2D.xml DISABLE_MODE enums

kdavis:


AThousandShips (changes requested):
Thank you for your contribution, however this is not correct, these enums are used for the `disable_mode` property, you are looking at `process_mode`, there is no `Node.DISABLE_MODE_STATIC`

PR #86534, opened 2023-12-27
EDITOR: fix spelling mistake on the spanish theme importer

guy-does-things:
Changed line 7942 of the Spanish editor translation, which said "Editor de theme" but it's an incorrect translation as it means "Theme editor" and not "Editor theme", The correct translation is "Theme del Editor"


AThousandShips:

However this is not where these files are edited, instead you edit them on the webplate, see [here](https://docs.godotengine.org/en/stable/contributing/documentation/editor_and_docs_localization.html)

Closing as this is not a valid change

PR #86535, opened 2023-12-27
Expose class_has_property on ClassDB

mashumafi:
I want to be able to check if a class has a property without having to iterate those properties on my own.

### DIFF
--- a/core/core_bind.cpp
+++ b/core/core_bind.cpp
@@ -1395,6 +1395,10 @@ TypedArray<Dictionary> ClassDB::class_get_signal_list(StringName p_class, bool p
 	return ret;
 }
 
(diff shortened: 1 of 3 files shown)


RedMser:
Why not use the `in` operator of GDScript?

mashumafi:
> Why not use the `in` operator of GDScript?

Woah, I didn't realize `in` could do that. I guess the issue would be that I would still need to check if the item is a method as well since `in` is not biased and returns `true` for both.

raulsntos:
- Related https://github.com/godotengine/godot-proposals/issues/3055

The linked proposal suggests adding a `has_property` method to `Object` instead of `ClassDB::class_has_property`.

PR #75944, opened 2023-04-11
Style: Harmonize header includes in modules

akien-mga:
This applies our existing style guide, and adds a new rule to that style guide for modular components such as platform ports and modules:

Includes from the platform port or module ("local" includes) should be listed
first in their own block using relative paths, before Godot's "core" includes
which use "absolute" (project folder relative) paths, and finally thirdparty
includes.

Includes in `#ifdef`s come after their relevant section, i.e. the overall
structure is:

- Local includes
  * Conditional local includes
- Core includes
  * Conditional core includes
- Thirdparty includes
  * Conditional thirdparty includes

Module follow-up to #75932.

PR #75932, opened 2023-04-11
Style: Harmonize header includes in platform ports

akien-mga:
This applies our [existing style guide](https://docs.godotengine.org/en/latest/contributing/development/code_style_guidelines.html#header-includes), and adds a new rule to that style guide for modular components such as platform ports and modules:

> Includes from the platform port or module should be included with relative paths (relative to the root folder of the modular component, e.g. `platform/linuxbsd/`), in their own section before Godot's "core" includes.

I'll add this to the [style guide](https://docs.godotengine.org/en/latest/contributing/development/code_style_guidelines.html#header-includes) if/when we reach consensus.

This helps differentiate between "local" code and the rest of the engine which can be treated as a library. The order of includes therefore flows from the most local (`header.h` for `header.cpp`) to the most external (`<string.h>`, `<windows.h>`, etc.). It's not perfect for platform code as we also have code in `drivers/` which has a similar kind of "locality" to headers in the platform folder, but rules can never be perfect ;)

This also helps (at least theoretically) with keeping module components modular. If we want to rename a platform or module folder, or more importantly, if people need to maintain a thirdparty platform or module, keeping the component's own files with relative includes help with the plug 'n play mechanism for `platform/` and `modules/`.

Same treatment applied to modules: #75944.

PR #86537, opened 2023-12-27
Cleanup of OpenXR module SCons config

BastiaanOlij:
This PR makes a few changes to the scons setup in the OpenXR modules:
- extensions that are included depending on platform/switches have been moved into the `extensions/platform`
- all `*.cpp` files in the `extensions` folder are now compiled and linked, they no longer need to be individually added
- `action_map`, `extensions` and `scene` now have their own `SCsub` file

### DIFF
--- a/modules/openxr/SCsub
+++ b/modules/openxr/SCsub
@@ -93,37 +93,13 @@ if env["builtin_openxr"]:
 module_obj = []
 
 env_openxr.add_source_files(module_obj, "*.cpp")
-env_openxr.add_source_files(module_obj, "action_map/*.cpp")
-env_openxr.add_source_files(module_obj, "scene/*.cpp")
-
-# We're a little more targeted with our extensions
-if env["platform"] == "android":
-    env_openxr.add_source_files(module_obj, "extensions/openxr_android_extension.cpp")
-if env["vulkan"]:
-    env_openxr.add_source_files(module_obj, "extensions/openxr_vulkan_extension.cpp")
-if env["opengl3"] and env["platform"] != "macos":
-    env_openxr.add_source_files(module_obj, "extensions/openxr_opengl_extension.cpp")
-
-env_openxr.add_source_files(module_obj, "extensions/openxr_local_floor_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_palm_pose_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_composition_layer_depth_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_eye_gaze_interaction.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_htc_controller_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_htc_vive_tracker_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_huawei_controller_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_hand_tracking_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_fb_foveation_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_fb_update_swapchain_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_fb_passthrough_extension_wrapper.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_fb_display_refresh_rate_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_pico_controller_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_wmr_controller_extension.cpp")
-env_openxr.add_source_files(module_obj, "extensions/openxr_ml2_controller_extension.cpp")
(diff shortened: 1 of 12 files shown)


BastiaanOlij:
I could use some pointers here by scons aficionados, I'm sure some of the SCsub files can be done better (cc @akien-mga ).

The main goal here was to not have to list all the extensions as we're slowly adding more.

modules/openxr/extensions/openxr_pico_controller_extension.cpp
```diff
@@ -30,7 +30,7 @@
 
 #include "openxr_pico_controller_extension.h"
 
-#include "../action_map/openxr_interaction_profile_metadata.h"
+#include "modules/openxr/action_map/openxr_interaction_profile_metadata.h"
```
dsnopek, line 33: This may be a personal preference thing, but because includes are all within the same module, I personally like the relative includes better than the absolute. Doing a quick grep through the rest of the code, it seems like other modules also prefer including other things within the same module using relative paths. Not sure this actually matters, though :-)

dsnopek (commented):
Thanks!

The scons changes look great to me and I did some minimal testing, although, I'm not a scons expert, so this probably still needs feedback from the buildsystem team

modules/openxr/extensions/openxr_pico_controller_extension.cpp
AThousandShips: The [code style](https://docs.godotengine.org/en/latest/contributing/development/code_style_guidelines.html#header-includes) *does* demand absolute paths, and I believe it additionally caused issues with SCU

dsnopek: Hm, ok. I guess it's one of those cases where existing code hasn't been updated to match the code standards yet?

Grepping for `include.*\.\.` shows lots of includes for headers inside of the same driver or module. Whereas grepping for `include.*modules/` only shows includes to modules from outside a given module. I don't see any current examples of a module including headers within the same module using an absolute path.

AThousandShips: It does seem unclear, the code style should probably be updated here

BastiaanOlij: I'm really on the fence, I agree with the absolute paths but we have started to use local paths especially with modules. It's also normal for the include file that belongs with a cpp file to not have the path.

But there doesn't seem to be much rule to it.

Here I've changed it to absolute paths because we now have extensions in the platform folder (those that are not always compiled in) and in the extensions folder (those that are always compiled), and where they include the same files I personally prefer it if the code looks similar.

When it comes to text, I'm very visual, I see patterns in what is written, so if things are different when they should be the same, it stands out like a sore thumb.

dsnopek: Is there a higher authority we can appeal to on this question?

Although, in the end, it's probably not all that important, and it's easy enough to reverse later if we decide to :-)

akien-mga: The higher authority is me :P

But I haven't had time to review now, I'll assess soon. I can just point out that

>  we now have extensions in the platform folder (those that are not always compiled in) and in the extensions folder (those that are always compiled)

May be going backward with regard to the kind of encapulsation we've been striving to get with platforms and modules. It's a bit of a chicken and egg problem since both are meant to be pluggable/modular, but so far the status quo seemed to be that modules should be the most modular part that can be taken in/out at will, so it should contain its own conditionally compiled platform code.

Modules `config.py` include information about the `env` and `platform` so they can be configured conditionally, while platform `detect.py` in theory shouldn't know much about the internals of a module. It's not always been fully respected, e.g. we had some gcc optimization flags for theora set in the Linux `detect.py`, but I would consider this a bug nowadays.

Anyway, TIWAGOS, I haven't reviewed any of the changes in this PR yet.

akien-mga: > The [code style](https://docs.godotengine.org/en/latest/contributing/development/code_style_guidelines.html#header-includes) _does_ demand absolute paths, and I believe it additionally caused issues with SCU

The code style hasn't been updated yet to take these refactorings into account:

- https://github.com/godotengine/godot/pull/75932
- https://github.com/godotengine/godot/pull/75944

These should be taken as the reference of how includes should be handled in modules and platforms. It's slightly convoluted, because the various aspects we're trying to convey are more complex than it looks :)

BastiaanOlij: I've changed it to relative paths, @akien-mga if you have time to take another look and let me know if you think this is good :)

dsnopek: Looking at the current PR, I'm still seeing lots of paths starting with "modules/openxr/..." (what we've been calling "absolute", but I guess is technically still relative, but relative to the top of the Godot source code) rather than relative to source file (which we've been calling "relative").

BastiaanOlij: ah crap, I only checked the header files..

akien-mga (approved):
Looks great to me!

dsnopek (approved):
Looks great to me!

YuriSizov:
Thanks!

PR #86538, opened 2023-12-27
Fix CSGShape3D's `_mesh_changed` to be thread safe

### LINKED ISSUE #86381: Create a Mesh for a CSGMesh3D will trigger `caller thread can't call this function in this node`
### Tested versions

Godot v4.3.dev (3ce73e5d4) 

### System information

Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3060 (NVIDIA; 31.0.15.3619) - 11th Gen Intel(R) Core(TM) i7-11700K @ 3.60GHz (16 Threads)

### Issue description

When debugging https://github.com/godotengine/godot/issues/86356, I noticed that sometimes when I use the mesh droplist of CSGMesh3d node to assign a new PrismMesh, It will give this error:

image

`Caller thread can't call this function in this node (/root/@EditorNode@17141/@Panel@13/@VBoxContainer@14/@HSplitContainer@17/@HSplitContainer@25/@HSplitContainer@33/@VBoxContainer@34/@VSplitContainer@36/@VSplitContainer@62/@VBoxContainer@63/@PanelContainer@110/MainScreen/@CanvasItemEditor@9463/@VSplitContainer@9282/@HSplitContainer@9284/@HSplitContainer@9286/@Control@9287/@SubViewportContainer@9288/@SubViewport@9289/3d/CSGMesh3D). Use call_deferred() or call_thread_group() instead.`

I know it's only printed out when `DEBUG_ENABLED`, but better needs some consideration. 

I did some quick test, it seems all mesh types that has a concrete shape will lead to this error (box, cylinder,quad,torus...), while arrayMesh or placeholderMesh will not trigger this error. Also after some operation, it seems the error will not be triggered everytime I set mesh.

### Steps to reproduce

1. Open or Create a new Scene
2. Create a new CSGMesh3D
3. Select PrismMesh as mesh type
4. Reset mesh and choose other mesh types

N/A

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/86381

If in a node safe thread, just call `update_gizmos` dierctly, otherwise use `call_deferred`

### DIFF
--- a/modules/csg/csg_shape.cpp
+++ b/modules/csg/csg_shape.cpp
@@ -934,7 +934,8 @@ CSGBrush *CSGMesh3D::_build_brush() {
 
 void CSGMesh3D::_mesh_changed() {
 	_make_dirty();
-	update_gizmos();
+
+	callable_mp((Node3D *)this, &Node3D::update_gizmos).call_deferred();
 }
 
 void CSGMesh3D::set_material(const Ref<Material> &p_material) {


modules/csg/csg_shape.cpp
```diff
@@ -929,7 +929,12 @@ CSGBrush *CSGMesh3D::_build_brush() {
 
 void CSGMesh3D::_mesh_changed() {
 	_make_dirty();
-	update_gizmos();
+
+	if (likely(is_accessible_from_caller_thread())) {
+		update_gizmos();
+	} else {
+		call_deferred(SNAME("update_gizmos"));
```
AThousandShips, line 936:
```suggestion
		callable_mp((Node3D *)this, &Node3D::update_gizmos).call_deferred();
```
Use callable syntax if possible

modules/csg/csg_shape.cpp
```diff
@@ -929,7 +929,12 @@ CSGBrush *CSGMesh3D::_build_brush() {
...
+	};
```
AThousandShips, line 937:
```suggestion
	}
```

AThousandShips (commented):
Might need a dedicated approach to queue a change, but think this is a good approach, will leave it for evaluation by others as well before I'm approving 🙂

modules/csg/csg_shape.cpp
jsjtxietian: Thanks! And Merry Christmas ~

AThousandShips: You too!

KoBeWi:
The call could always be deferred, no? I don't see why a condition would be useful.

KoBeWi (approved):

PR #81324, opened 2023-09-05
Allow directly reading audio data

Faolan-Rad:
A basic solution to [Proposal](https://github.com/godotengine/godot-proposals/issues/7601)

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/7601*

PR #75065, opened 2023-03-18
Expose some `AudioStreamPlayback` methods

AThousandShips:
Returning an array with only the size of the return of the original `mix` function to convey that value, which I thinks makes sense

Not exposing `tag_used_streams` as I'm not familiar with how it works and the relevant functionality isn't exposed elsewhere.

Examples of use case of this feature https://github.com/godotengine/godot-proposals/issues/6466, https://github.com/godotengine/godot-proposals/discussions/6385

Closes godotengine/godot-proposals#127

PR #86539, opened 2023-12-27
Expose some AudioStreamPlayback methods (namely `mix_audio()`).

stechyo:
Pretty much the same as: 
 - https://github.com/godotengine/godot/pull/75065/
 - https://github.com/godotengine/godot/pull/81324

The former is archived, and the latter seems to go against the maintainers' wishes that `start`, `stop`, and `mix` should not be exposed outward. I added some shim functions to get around this, not sure if that's fine. There are several proposals with different use cases for this:
 - https://github.com/godotengine/godot-proposals/issues/127
 - https://github.com/godotengine/godot-proposals/issues/6466
 - https://github.com/godotengine/godot-proposals/issues/7601
 -  https://github.com/godotengine/godot-proposals/issues/8609

My usecase is embedding [steam-audio](https://valvesoftware.github.io/steam-audio/) as a GDExtension, which requires me to have these functions available as bound methods.

### DIFF
--- a/doc/classes/AudioStreamPlayback.xml
+++ b/doc/classes/AudioStreamPlayback.xml
@@ -113,6 +113,13 @@
 				[b]Note:[/b] Can return fewer frames than requested, make sure to use the size of the return value.
 			</description>
 		</method>
+		<method name="seek">
+			<return type="void" />
+			<param index="0" name="time" type="float" default="0.0" />
+			<description>
+				Seeks the stream at the given [param time], in seconds.
+			</description>
+		</method>
 		<method name="set_sample_playback" experimental="">
 			<return type="void" />
 			<param index="0" name="playback_sample" type="AudioSamplePlayback" />
--- a/doc/classes/AudioStreamPlayback.xml
+++ b/doc/classes/AudioStreamPlayback.xml
@@ -79,10 +79,38 @@
 				Overridable method. Called whenever the audio stream is mixed if the playback is active and [method AudioServer.set_enable_tagging_used_audio_streams] has been set to [code]true[/code]. Editor plugins may use this method to "tag" the current position along the audio stream and display it in a preview.
 			</description>
 		</method>
+		<method name="get_loop_count" qualifiers="const">
+			<return type="int" />
+			<description>
+				Returns the number of times the stream has looped.
+			</description>
+		</method>
+		<method name="get_playback_position" qualifiers="const">
+			<return type="float" />
+			<description>
+				Returns the current position in the stream, in seconds.
+			</description>
+		</method>
 		<method name="get_sample_playback" qualifiers="const" experimental="">
 			<return type="AudioSamplePlayback" />
 			<description>
 				Returns the [AudioSamplePlayback] associated with this [AudioStreamPlayback] for playing back the audio sample of this stream.
+			</description>
+		</method>
+		<method name="is_playing" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if the stream is playing.
+			</description>
+		</method>
+		<method name="mix_audio">
+			<return type="PackedVector2Array" />
+			<param index="0" name="rate_scale" type="float" />
+			<param index="1" name="frames" type="int" />
+			<description>
+				Mixes up to [param frames] of audio from the stream from the current position, at a rate of [param rate_scale], advancing the stream.
+				Returns a [PackedVector2Array] where each element holds the left and right channel volume levels of each frame.
+				[b]Note:[/b] Can return fewer frames than requested, make sure to use the size of the return value.
 			</description>
 		</method>
 		<method name="set_sample_playback" experimental="">
@@ -90,6 +118,19 @@
 			<param index="0" name="playback_sample" type="AudioSamplePlayback" />
 			<description>
 				Associates [AudioSamplePlayback] to this [AudioStreamPlayback] for playing back the audio sample of this stream.
+			</description>
+		</method>
+		<method name="start">
+			<return type="void" />
+			<param index="0" name="from_pos" type="float" default="0.0" />
+			<description>
+				Starts the stream from the given [param from_pos], in seconds.
+			</description>
+		</method>
+		<method name="stop">
+			<return type="void" />
+			<description>
+				Stops the stream.
 			</description>
 		</method>
 	</methods>
--- a/servers/audio/audio_stream.cpp
+++ b/servers/audio/audio_stream.cpp
@@ -76,29 +76,40 @@ int AudioStreamPlayback::mix(AudioFrame *p_buffer, float p_rate_scale, int p_fra
 	return ret;
 }
 
-PackedVector2Array AudioStreamPlayback::mix_audio(float p_rate_scale, int p_frames) {
-	Vector<AudioFrame> frames_in;
-	frames_in.resize(p_frames);
-
-	int frames = mix(frames_in.ptrw(), p_rate_scale, p_frames);
+PackedVector2Array AudioStreamPlayback::_mix_audio_bind(float p_rate_scale, int p_frames) {
+	Vector<AudioFrame> frames = mix_audio(p_rate_scale, p_frames);
 
 	PackedVector2Array res;
-	res.resize(frames);
+	res.resize(frames.size());
 
 	Vector2 *res_ptrw = res.ptrw();
-	for (int i = 0; i < frames; i++) {
-		res_ptrw[i] = Vector2(frames_in[i].left, frames_in[i].right);
+	for (int i = 0; i < frames.size(); i++) {
+		res_ptrw[i] = Vector2(frames[i].left, frames[i].right);
 	}
 
 	return res;
 }
 
+Vector<AudioFrame> AudioStreamPlayback::mix_audio(float p_rate_scale, int p_frames) {
+	Vector<AudioFrame> res;
+	res.resize(p_frames);
+
+	int frames = mix(res.ptrw(), p_rate_scale, p_frames);
+	res.resize(frames);
+
+	return res;
+}
+
 void AudioStreamPlayback::start_playback(double p_from_pos) {
-	start();
+	start(p_from_pos);
 }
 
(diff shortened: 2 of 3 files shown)


servers/audio/audio_stream.cpp
```diff
@@ -89,6 +113,13 @@ void AudioStreamPlayback::_bind_methods() {
 	GDVIRTUAL_BIND(_seek, "position")
 	GDVIRTUAL_BIND(_mix, "buffer", "rate_scale", "frames");
 	GDVIRTUAL_BIND(_tag_used_streams);
+
+	ClassDB::bind_method(D_METHOD("get_raw_audio", "rate_scale", "frames"), &AudioStreamPlayback::get_raw_audio);
+	ClassDB::bind_method(D_METHOD("start", "p_from_pos"), &AudioStreamPlayback::start_playback);
```
AThousandShips, line 118:
```suggestion
	ClassDB::bind_method(D_METHOD("start", "from_pos"), &AudioStreamPlayback::start_playback, DEFVAL(0.0));
```
No prefix for bind arguments, also match the default arguments of the binding

stechyo: Done, thanks for the feedback

ellenhp:
We really should have a story here because a lot of users want this and it is currently impossible without a gdextension. In gdext you can just call mix repeatedly on a playback but that's not reasonable to make every user do who wants this. It feels like every few weeks in the discord someone asks about it.

stechyo:
> We really should have a story here because a lot of users want to get raw pcm data from a stream and it is currently impossible without a gdextension. In gdext you can just call mix repeatedly on a playback but that's not reasonable to make every user do who wants this. It feels like every few weeks in the discord someone asks about it.

Just a note, correct me if I'm wrong: unless you're referring to something other than the `_mix()` method that gets generated for AudioStreamPlayback, I don't think that works. That method is virtual and, when called in GDExt for a native audio stream, does nothing (at least in my experience, I may have missed something of course).

theraot:
I remember pushback on the lines of "GDScript is too slow for that", but we should be able to use GDScript to connect this to a mixing solution in GPU, for example with compute shaders. In fact, I was able to use the fragment shader in Godot 3.x to generate audio. I shared a proof of concept in discussions: https://github.com/godotengine/godot-proposals/discussions/6476#discussioncomment-5271279

Futhermore, Godot have integration with other lnaguages that are considered faster such as C# and rust.

I have a post by Juan Linietsky agreeing on exposing more audio internals: https://x.com/reduzio/status/1683126371013349378

Thus, I hope this does not get the "It won't work on GDScript" treatment.

stechyo:
I dislike the fact that I'm exposing this to GDScript at all but "should some things only be exposed on GDExt and not C#/GDS" is a much larger and possibly controversial discussion. This PR is not the ideal solution (I've learned this with time, see https://github.com/stechyo/godot-steam-audio/issues/39), but I'm not aware of any efforts to do something similar but in a cleaner way (again, "knowing what is being worked on" is sometimes an issue in Godot, but again, that's another discussion).

doc/classes/AudioStreamPlayback.xml
```diff
@@ -79,5 +79,46 @@
...
+				Returns the number of times the stream has looped.
+			</description>
+		</method>
+		<method name="get_playback_position" qualifiers="const">
+			<return type="float" />
+			<description>
+				Returns the current position in the stream, in seconds.
+			</description>
+		</method>
+		<method name="get_raw_audio">
```
reduz, line 94: I would just call it mix_audio()

reduz (changes requested):

reduz:
Looks good to me other than what I mentioned.

doc/classes/AudioStreamPlayback.xml
stechyo: Changed, thanks

servers/audio/audio_stream.cpp
```diff
@@ -76,6 +76,30 @@ int AudioStreamPlayback::mix(AudioFrame *p_buffer, float p_rate_scale, int p_fra
...
+	Vector<AudioFrame> frames_in;
+	frames_in.resize(p_frames);
+
+	int frames = mix(frames_in.ptrw(), p_rate_scale, p_frames);
+
+	PackedVector2Array res;
+	res.resize(frames);
+
+	for (int i = 0; i < frames; i++) {
+		res.set(i, Vector2(frames_in[i].left, frames_in[i].right));
```
AThousandShips, line 89:
```suggestion
	Vector2 *res_ptrw = res.prtw();

	for (int i = 0; i < frames; i++) {
		res_ptrw[i] = Vector2(frames_in[i].left, frames_in[i].right);
```

stechyo: Done, thanks

AThousandShips:
Since much of the code and especially the documentation is the same as my original PR I have to ask if you salvaged that one?

stechyo:
Yeah, as I said in the description it's pretty much the same as two other PRs. Is there some "salvaging etiquette" I'm not aware of?

AThousandShips:
I'd say that explicitly stating it would be good in the future as it's not clear from "essentially the same", copying code is something you need to be explicit about

I'd ask that you add me as a co-author in that case, by adding:
```

Co-authored-by: A Thousand Ships <96648715+AThousandShips@users.noreply.github.com>
```
To your commit (the empty line is important)

Thank you!

stechyo:
The XML description was certainly salvaged, so it's more than reasonable to add you as a co-author. Sorry for not having done so previously!

solitaryurt:
Is this PR at the point where it just needs approval now?

jams3223:
@reduz I really want this merged, I am testing some project with @stechyo steam-audio plugin and I want to make sure that we can squash as much bugs as possible knowing Godot 4.4 is 8-9 months away from now.

MasterDingo:
Any chances to see this PR merged into 4.3?

AThousandShips:
4.3 is in final release phase so this will be not be merged before 4.4

stechyo:
Now that 4.3 has released, when could we merge this so that it's ready for 4.4? It seems there's no issues remaining with the code.

AThousandShips:
It needs an approval by the audio team

adamscott:
Just wondering, before approving the PR.

1. Is it just to tweak the PackedVector2Array in order to playback the data afterwards?
2. What about surround systems/mixing?

stechyo:
1. My use case is for tweaking the audio frame array, yes. I believe there are also use cases for audio visualisers and the like.
2. I haven’t considered surround systems.

mk56-spn:
> 1. My use case is for tweaking the audio frame array, yes. I believe there are also use cases for audio visualisers and the like.
> 2. I haven’t considered surround systems.

To add a bit to this. It was specifically static ( so audio visualizers that show the whole songs graph ( think audacity ) that were a major pain to do beforehand. So yeah this is very much appreciated for that sort of thing.

stechyo:
@adamscott do these answers block your approval? Just wondering

adamscott:
> @adamscott do these answers block your approval? Just wondering

@stechyo I'm investigating. I'm gonna give you an answer shortly.

adamscott:
> I haven’t considered surround systems.

It seems that surround is only done at the server level when adjusting volume, so no need for a special case for surround systems. Sorry, I mixed up stuff.

servers/audio/audio_stream.cpp
```diff
@@ -108,6 +133,12 @@ void AudioStreamPlayback::_bind_methods() {
 
 	ClassDB::bind_method(D_METHOD("set_sample_playback", "playback_sample"), &AudioStreamPlayback::set_sample_playback);
 	ClassDB::bind_method(D_METHOD("get_sample_playback"), &AudioStreamPlayback::get_sample_playback);
+	ClassDB::bind_method(D_METHOD("mix_audio", "rate_scale", "frames"), &AudioStreamPlayback::mix_audio);
```
adamscott, line 136: Here, I changed the bind function name, as my stated in my previous comment.
```suggestion
	ClassDB::bind_method(D_METHOD("mix_audio", "rate_scale", "frames"), &AudioStreamPlayback::_mix_audio_bind);
```

servers/audio/audio_stream.cpp
```diff
@@ -108,6 +133,12 @@ void AudioStreamPlayback::_bind_methods() {
...
+	ClassDB::bind_method(D_METHOD("start", "from_pos"), &AudioStreamPlayback::start_playback, DEFVAL(0.0));
+	ClassDB::bind_method(D_METHOD("stop"), &AudioStreamPlayback::stop_playback);
```
adamscott, line 138: If we expose start and stop, I would expose seek as well.

adamscott (changes requested):
Small changes IMO, but your PR is close to be merged!

stechyo:
I think all changes should be addressed :) could someone approve the workflow?

stechyo:
@adamscott is anything missing to merge? Just checking, idk what the etiquette is on this repo (if the commenter needs to resolve conversations, or if it's the PR author who does so).

adamscott (approved):
Seems good to me!

stechyo:
Awesome! @reduz I think your requested change should be addressed as well, can we merge this?

doc/classes/AudioStreamPlayback.xml
lhilton: @stechyo is this PR still pending the above change?

stechyo: No, the change was made when I wrote the reply above (June 16).

stechyo:
@reduz @AThousandShips my apologies for double-pinging here, but this PR has been approved by one engine maintainer and the comment of another maintainer has been addressed for a while now. Is there anything I can do to help merge this? I would love to get this in before the 4.4 freeze begins, and it seems to be ready to merge.

adamscott:
I set the PR in the merge queue, so it should be merge as soon as possible. Thanks again for the PR!

Repiteo:
Thanks, and congratulations on your first contribution! 🎉

PR #86542, opened 2023-12-27
Improve Path2D editing

KoBeWi:
When adding or splitting a point in Path2D, it was creating 2 undo actions: for creating the point and moving it. This PR merges it into one undo action.

https://github.com/godotengine/godot/assets/2223172/4282d88c-173b-4166-9b62-4275d9f4873d

### DIFF
--- a/editor/plugins/path_2d_editor_plugin.cpp
+++ b/editor/plugins/path_2d_editor_plugin.cpp
@@ -156,16 +156,14 @@ bool Path2DEditor::forward_gui_input(const Ref<InputEvent> &p_event) {
 		// Check for point creation.
 		if (mb->is_pressed() && mb->get_button_index() == MouseButton::LEFT && ((mb->is_command_or_control_pressed() && mode == MODE_EDIT) || mode == MODE_CREATE)) {
 			Ref<Curve2D> curve = node->get_curve();
+			curve->add_point(cpoint);
 
 			EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
 			undo_redo->create_action(TTR("Add Point to Curve"));
 			undo_redo->add_do_method(curve.ptr(), "add_point", cpoint);
-			undo_redo->add_undo_method(curve.ptr(), "remove_point", curve->get_point_count());
-			undo_redo->add_do_method(canvas_item_editor, "update_viewport");
-			undo_redo->add_undo_method(canvas_item_editor, "update_viewport");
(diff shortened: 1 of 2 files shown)


ajreckof:
I'm wondering if this is also needed for Path3D

KoBeWi:
I tested Path3D and it didn't seem to have this problem.

editor/plugins/path_2d_editor_plugin.cpp
```diff
@@ -222,13 +224,18 @@ bool Path2DEditor::forward_gui_input(const Ref<InputEvent> &p_event) {
...
-					undo_redo->add_do_method(canvas_item_editor, "update_viewport");
-					undo_redo->add_undo_method(canvas_item_editor, "update_viewport");
-					undo_redo->commit_action();
+					if (action == ACTION_MOVING_POINT) {
+						undo_redo->add_undo_method(curve.ptr(), "set_point_position", action_point, moving_from);
+						undo_redo->add_do_method(canvas_item_editor, "update_viewport");
+						undo_redo->add_undo_method(canvas_item_editor, "update_viewport");
+					} else {
+						undo_redo->commit_action();
+					}
```
ajreckof, line 237: I think here the use of nested actions makes it a bit harder to read especially since we are already checking the exact condition for which we should not be creating this action so I would replace those lines with 
```
					if (action == ACTION_MOVING_POINT) {
						undo_redo->create_action(TTR("Move Point in Curve"));
						undo_redo->add_undo_method(curve.ptr(), "set_point_position", action_point, moving_from);
					}
					undo_redo->add_do_method(curve.ptr(), "set_point_position", action_point, cpoint);
					undo_redo->add_do_method(canvas_item_editor, "update_viewport");
					undo_redo->add_undo_method(canvas_item_editor, "update_viewport");
					undo_redo->commit_action(false);
```

And remove the `update_viewport` on line 163-164 and 201-202 as it is better to have them just in one place instead of three

ajreckof (approved):
Tested and works perfectly. Except for the small issue I pointed above this looks really good.

editor/plugins/path_2d_editor_plugin.cpp
KoBeWi: That code doesn't look valid. You are creating a new action inside the condition, but never commit it. A commit has to exist for every started action.

akien-mga (approved):

akien-mga:
Thanks!

Issue #86490, opened 2023-12-24
Add Direct3D 12 support to official Windows binaries


oscarbg:
### Tested versions

reproducible in Godot_v4.3-dev1_win64

### System information

win11 Godot_v4.3-dev1_win64

### Issue description

Hi,
wanted to test Godot_v4.3-dev1 new D3D12 support by downloading prebuild Windows binary but seems it doesn't come with D3D12 support prebuilt..

### Steps to reproduce

running:
Godot_v4.3-dev1_win64.exe --rendering-driver d3d12
shows:
```
Unknown rendering driver 'd3d12', aborting.
Valid options are 'vulkan', 'opengl3', 'opengl3_angle', 'dummy', .
```

### Minimal reproduction project (MRP)

no need

jsjtxietian:
In the meantime you can try this: https://docs.godotengine.org/en/latest/contributing/development/compiling/compiling_for_windows.html#compiling-with-support-for-direct3d-12

Calinou:
[For context](https://godotengine.org/article/dev-snapshot-godot-4-3-dev-1/):

> Being based on proprietary technology, D3D12 support comes as an optional feature keeping the open source spirit of the engine intact. Make sure to check the [updated documentation](https://docs.godotengine.org/en/latest/contributing/development/compiling/compiling_for_windows.html) for compiling Godot with D3D12 enabled (there is also an [update regarding cross-compilation](https://github.com/godotengine/godot-docs/pull/8624) getting ready to be merged).

https://github.com/godotengine/godot/pull/86551 makes it possible to lift this requirement (and MinGW building support was added by https://github.com/godotengine/godot/pull/86150), so it should be possible to publish official D3D12-enabled builds once both PRs are merged.

bruvzg:
> https://github.com/godotengine/godot/pull/86551 makes it possible to lift this requirement (and MinGW[1](https://github.com/godotengine/godot/issues/86490#user-content-fn-1-30d16733aac01ab00304382e3ca3b954) building support was added by https://github.com/godotengine/godot/pull/83452), so it should be possible to publish official D3D12-enabled builds once both PRs are merged.

[...]

PR #86551, opened 2023-12-27
[D3D12] Dynamically load Agility SDK.

bruvzg:
Instead of relaying on exported symbols and D3D12 loader, load SDK directly via device factory. This should allow using user specified Agility SDK version instead of hard-coded one, and fallback to the system D3D driver if it's not found (so we can make official builds with Agility support, but avoid shipping proprietary DLL that require uses to accept EULA).

Adds `rendering/rendering_device/d3d12/agility_sdk_version` property to control used SDK version.

Based on Mesa Agility loading code - https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/18305/diffs#b974333948460f1be995f150bf4db414b591b1b1

- *Production edit: This partially addresses https://github.com/godotengine/godot/issues/86490.*

### DIFF
--- a/doc/classes/ProjectSettings.xml
+++ b/doc/classes/ProjectSettings.xml
@@ -2616,6 +2616,9 @@
 		<member name="rendering/renderer/rendering_method.web" type="String" setter="" getter="" default="&quot;gl_compatibility&quot;">
 			Override for [member rendering/renderer/rendering_method] on web.
 		</member>
+		<member name="rendering/rendering_device/d3d12/agility_sdk_version" type="int" setter="" getter="" default="610">
+			Version code of the Direct3D 12 Agility SDK to use ([code]D3D12SDKVersion[/code]).
+		</member>
 		<member name="rendering/rendering_device/d3d12/max_misc_descriptors_per_frame" type="int" setter="" getter="" default="512">
 			The number of entries in the miscellaneous descriptors heap the Direct3D 12 rendering driver uses each frame, used for various operations like clearing a texture.
 			Depending on the complexity of scenes, this value may be lowered or may need to be raised.
--- a/drivers/d3d12/d3d12_context.cpp
+++ b/drivers/d3d12/d3d12_context.cpp
@@ -55,27 +55,14 @@
 #include <guiddef.h>
 
 #include <dxguids.h>
+#ifndef CLSID_D3D12DeviceFactory
+// Note: symbol is not available in MinGW import library.
+const CLSID CLSID_D3D12DeviceFactory = __uuidof(ID3D12DeviceFactory);
+#endif
 #endif
 
 extern "C" {
 char godot_nir_arch_name[32];
-
-#ifdef AGILITY_SDK_ENABLED
-__declspec(dllexport) extern const UINT D3D12SDKVersion = 610;
-#ifdef AGILITY_SDK_MULTIARCH_ENABLED
-#if defined(__aarch64__) || defined(_M_ARM64) || defined(_M_ARM64EC)
-__declspec(dllexport) extern const char *D3D12SDKPath = "\\.\\arm64";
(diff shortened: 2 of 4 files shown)


doc/classes/ProjectSettings.xml
```diff
@@ -2616,6 +2616,9 @@
 		<member name="rendering/renderer/rendering_method.web" type="String" setter="" getter="" default="&quot;gl_compatibility&quot;">
 			Override for [member rendering/renderer/rendering_method] on web.
 		</member>
+		<member name="rendering/rendering_device/d3d12/agility_sdk_version" type="int" setter="" getter="" default="610">
+			D3D12 Agility SDK version to use.
```
RandomShaper, line 2620:
```suggestion
			Version code of the Direct3D3 12 Agility SDK to use ([code]D3D12SDKVersion[/code]).
```

drivers/d3d12/d3d12_context.cpp
```diff
@@ -501,7 +493,13 @@ void D3D12Context::_dump_adapter_info(int p_index) {
 }
 
 Error D3D12Context::_create_device(DeviceBasics &r_basics) {
-	HRESULT res = D3D12CreateDevice(gpu.Get(), D3D_FEATURE_LEVEL_11_0, IID_PPV_ARGS(r_basics.device.GetAddressOf()));
+	HRESULT res;
+	if (r_basics.factory) {
+		r_basics.factory->SetFlags(D3D12_DEVICE_FACTORY_FLAG_ALLOW_RETURNING_EXISTING_DEVICE | D3D12_DEVICE_FACTORY_FLAG_ALLOW_RETURNING_INCOMPATIBLE_EXISTING_DEVICE);
```
RandomShaper, line 498: Why these flags? I'm wondering if re-using a device instance can be somehow detrimental for local devices.

bruvzg: This might be unnecessary, I'll check if it's working without it (flags are copy-pasted from Mesa).

RandomShaper (approved):

RandomShaper:
A nice enhancement.

doc/classes/ProjectSettings.xml
```diff
@@ -2616,6 +2616,9 @@
 		<member name="rendering/renderer/rendering_method.web" type="String" setter="" getter="" default="&quot;gl_compatibility&quot;">
 			Override for [member rendering/renderer/rendering_method] on web.
 		</member>
+		<member name="rendering/rendering_device/d3d12/agility_sdk_version" type="int" setter="" getter="" default="610">
+			Version code of the Direct3D3 12 Agility SDK to use ([code]D3D12SDKVersion[/code]).
```
akien-mga, line 2620:
```suggestion
			Version code of the Direct3D 12 Agility SDK to use ([code]D3D12SDKVersion[/code]).
```

akien-mga (approved):

akien-mga:
Thanks!

Calinou:
It looks like this caused a build-time regression: https://forum.godotengine.org/t/error-while-compiling-godot-for-d3d12-with-scons/41654

RandomShaper:
I've just found this approach won't work on an up-to-date Windows 10, such as mine:
image

The SDK config interface is not available until a later release (Windows 10 insider build or Windows 11, I guess), as said in https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nn-d3d12-id3d12sdkconfiguration:
image

bruvzg:
> The SDK config interface is not available until a later release (Windows 10 insider build or Windows 11, I guess)

Since it loaded dynamically, I assume it won't break DX12, but only prevent it from loading Agility libs. I guess we should keep both predefined exports and factory approach in this case (will test if it works later).

Issue #69749, opened 2022-12-08
Popup subwindows created by MenuButtons not scaling relative to parent window


MACMAN2003:
___

***Bugsquad note:** This issue has been confirmed several times already. No need to confirm it further.*

___

### Godot version

4.0.beta3.mono

### System information

Windows 10, Vulkan, Nvidia RTX 2080 TI

### Issue description

When "Embed Subwindows" is set to false in the project settings, popup windows made by MenuButtons don't get placed/scaled relative to the "parent" window

<figure>
<img src="https://user-images.githubusercontent.com/20659419/206360110-91573431-cb70-47a4-95f8-444fd319c7e3.png" alt="unscaled parent window">
<figcaption>unscaled parent window^</figcaption>
</figure>

the popup window is scaled correctly when the parent window is not scaled at all.

<figure>
<img src="https://user-images.githubusercontent.com/20659419/206361095-cc796a32-efa5-4c0f-b6db-6877997b9854.png" alt="scaled parent window">
<figcaption>scaled parent window^</figcaption>
</figure>

and when maximized, the popup window stays the same size (and even in the same place) making it hard to read on high-DPI devices

This is easily fixed by setting "Embed Subwindows" to true in the project settings, but i want to also have a file dialog window that's not embedded

### Steps to reproduce

- make a new project
- go to the project settings
- set the window width and height to 1920 and 1080
- set embed subwindows to false
- set the stretch mode to "viewport" or "canvas items"
- make a new scene
- make a Control node
- make a Container node as a child of the control node
- make a MenuButton node as a child of the container node
- add an item to the menuButton node
- run the project
- resize the window
- click on the MenuButton
or
- open the reproduction project
- run the project
- resize the window
- click on the "click for menu" button

### Minimal reproduction project

[subwindows.zip](https://github.com/godotengine/godot/files/10182339/subwindows.zip)

Sauermann:
Can confirm on non-mono-build v4.0.beta.custom_build [a7937fe54]
related to #69171 which addresses the same problem for `OptionButton`.

KoB-Kirito:
Tooltip popups are also affected.

BrunkMortel:
From what I have seen in the code, there would be a workaround by exposing the inner [PanelContainer](https://github.com/godotengine/godot/blob/0028fd625e2c5b202e204bc12828cbca043213d3/scene/gui/popup_menu.h#L156C1-L156C34).
[...]

Issue #69171, opened 2022-11-25
OptionButton popup doesn't take SubViewportContainer transformation into consideration


simon-peregrine:
___

***Bugsquad note:** This issue has been confirmed several times already. No need to confirm it further.*

___

### Godot version

4.0.beta6.mono

### System information

Windows 11

### Issue description

When embedded in a SubViewportContainer whose scale is not 1, the popup menu of OptionButton is scaled and positioned as if the scale were 1.

Using the following scene structure:
![NodeLayout](https://user-images.githubusercontent.com/21371262/204009855-229ef722-a9fa-44b6-af26-a42bf0ff5747.png)

The SubViewportContainer with scale = 1:
![OptionButtonCorrect](https://user-images.githubusercontent.com/21371262/204009719-eb9cc7e4-854d-4c0a-95fa-2a869877299f.png)

The SubViewportContainer with scale = 2:
![OptionButtonIncorrect](https://user-images.githubusercontent.com/21371262/204009936-563810e9-898c-455c-ab8d-e933b06bc07a.png)

### Steps to reproduce

Put an OptionButton inside a SubViewportContainer hierarchy and set the scale of the SubViewportContainer to 2.

### Minimal reproduction project

[OptionButtonIssue.zip](https://github.com/godotengine/godot/files/10093696/OptionButtonIssue.zip)

Sauermann:
I believe that #68627 will fix the position problem. Will test this later.

Sauermann:
The combination of #66688, #68627 and #69185 will fix the wrong position. (Didn't expect it to be this complex)

The fact that the popup-menu-scale is always 1, is a different matter. I am not sure if this should be considered a bug or intended behavior?

Calinou:
Related to https://github.com/godotengine/godot/issues/65947.

clayjohn:
Will be fixed by https://github.com/godotengine/godot/pull/69185

KoB-Kirito:
Tooltip popups are also affected.

malk82:
Problem is not solved. If you have a subviewportcontainer that has multiple nested subviewport inside it. The position of the list is on the upper left corner and not inside the viewport. I'm using Godot 4.4.2

Issue #54030, opened 2021-10-20
Popups (OptionButton, tooltips) placed and scaled wrong when using `canvas_items` stretch mode or scale factor different from `1.0` (DisplayServer regression)


grok-games:
___

***Bugsquad note:** This issue has been confirmed several times already. No need to confirm it further.*

___

### Godot version

4.0-dev.20211004_win64

### System information

Windows 10

### Issue description

When using OptionButton and choosing **canvas_items** mode in Project Settings->Display, popup of OptionButton is placed and scaled wrong (fonts too). When using "disabled" mode it is placed correctly.

![popupmenu_issue1](https://user-images.githubusercontent.com/92084190/138126327-91070372-3521-4181-a25f-d437ed54140e.png)

I originally thought that this is related to my custom size of display, but it seems only "canvas_items" does this.

![popupmenu_issue2](https://user-images.githubusercontent.com/92084190/138126336-4bbbf915-5785-4247-9719-8fd5e77d4da0.png)

I've added project, so you can test. but it's pretty obvious, I have no scaling or complex settings on nodes.

### Steps to reproduce

Set canvas_items mode and use PopupButton.

### Minimal reproduction project

[game-client_issue_popup_menu.zip](https://github.com/godotengine/godot/files/7382574/game-client_issue_popup_menu.zip)

Calinou:
This is because multiple windows are now used by default (this applies to PopupMenus so they can extend outside the project window). https://github.com/godotengine/godot/pull/50680 would resolve this by default, but in the meantime, you can enable **Embed Subwindows** in the Project Settings.

eleonore122:
I've encountered the same thing, and is seems that the clickable area for the subwindow is placed in a different spot than the subwindow itself.
here you can see the dropdown menu, hovering near the option doesn't highlight anything
![option_button](https://user-images.githubusercontent.com/40723433/153252863-812a929e-65d9-4948-9058-b69081b3a5f2.png)
moving above the box entirely though, does highlight the option. 
![option_highlighted](https://user-images.githubusercontent.com/40723433/153252882-09b3d2c6-4038-4383-b3d1-d435fe467c9d.png)

KoBeWi:
Looks like the regression part is fixed. The popup works correctly when you use embedded subwindows (default).

Calinou:
Popup drawing is still blurry when using a larger-than-default window size with the `canvas_items` stretch mode:

[...]

PR #86553, opened 2023-12-27
Fix non-embedded tooltips, popups and dialogs resizing to match Viewport content scale factors != 1.0

### LINKED ISSUE #44942: PopupPanel doesn't resize properly
**Godot version:**
3.2.4.beta4

**Issue description:**
When having a label with autowrap inside of a PopupPanel and changing the content of the label the PopupPanel doesn't resize accordingly.
Creating a new label in the script and adding it as a child node resizes the PopupPanel as expected.

For example:
Peek 2021-01-05 17-00
The PanelContainer on the top left has the same content as the PopupPanel which is created by the custom tooltip for the button in the middle.

**Steps to reproduce:**
1. Create a PopupPanel
2. Create a VBoxContainer
3. Add a label inside the vbox, enable autowrap
4. Create a script to change the text of the label in the ready function 

[testpopuppanel.zip](https://github.com/godotengine/godot/files/5771414/testpopuppanel.zip)

### LINKED ISSUE #89482: Vertical scrollbar is visible in all popup menus when using a custom display scale
### Tested versions

v4.3.dev.custom_build [da945ce62]

Wasn't happening in 4.2.1

### System information

Godot v4.3.dev (da945ce62) - macOS 14.2.1 - Vulkan (Forward+) - integrated Apple M1 Max - Apple M1 Max (10 Threads)

### Issue description

It's like this in every popup menu:

image<br>

**_EDIT: Correction, this has nothing to do with a theme, this is with no theme:_**

image

### Steps to reproduce

~~Use [this theme](https://github.com/passivestar/godot-minimal-theme) with 4.3~~

~~This might be mac-specific because I asked another person and found out that it's not happening on windows~~

Open 4.3 and set custom display scale to 1.6

N/A

Koyper:
Required for fixes to https://github.com/godotengine/godot/issues/69749, https://github.com/godotengine/godot/issues/69171, and for fixes to the issues in https://github.com/godotengine/godot/issues/54030 when `embed_subwindows` is _disabled_.
Probably fixes https://github.com/godotengine/godot/issues/44942.
*Bugsquad edit:* Fixes #89482.

This will not relate directly to a fix for the blurry subwindows issue, but might help with the eventual solution for that.

This fix causes the `PopupPanel` children to correctly resize to fit the popup when the containing `Viewport.content_scale_factor` is changed. Without this fix, the children would expand outside the popup bounds if the scale factor was greater than 1.0, or shrivel if less than 1.0.

This fix is required for several separate PR's I will post, the first of which is to make tooltips scale to match the main viewport scale factor when the project `embed_subwindows` is _disabled_.

Implementation is slightly different depending on whether the popup is embedded or not. If embedded, the popup will scale simply by changing the viewport content scale factor, if the popup is _not_ embedded, i.e., using a floating system window, the popup content scale factor should be set to match the containing viewport's content scale factor. If this is done, the rendering will be correct (non-blurry).

The fixes below show the properly scaled tooltip that will be in a separate PR.

BEFORE FIX:

https://github.com/godotengine/godot/assets/33969780/9fe32aea-61b7-406b-b244-d8d152ead21e

ALSO BEFORE FIX - Note the tooltip scale stuck at 1.0

https://github.com/godotengine/godot/assets/33969780/6dac280f-a519-4db4-b3cd-1852507fb054

AFTER FIX:

https://github.com/godotengine/godot/assets/33969780/b0a11615-84f1-4226-9283-b87cc8ada48f

https://github.com/godotengine/godot/assets/33969780/3c692787-618e-4614-9220-bcc816abc3ed

### DIFF
--- a/scene/gui/dialogs.cpp
+++ b/scene/gui/dialogs.cpp
@@ -202,7 +202,7 @@ void AcceptDialog::register_text_enter(LineEdit *p_line_edit) {
 }
 
 void AcceptDialog::_update_child_rects() {
-	Size2 dlg_size = get_size();
+	Size2 dlg_size = Vector2(get_size()) / get_content_scale_factor();
 	float h_margins = theme_cache.panel_style->get_margin(SIDE_LEFT) + theme_cache.panel_style->get_margin(SIDE_RIGHT);
 	float v_margins = theme_cache.panel_style->get_margin(SIDE_TOP) + theme_cache.panel_style->get_margin(SIDE_BOTTOM);
 
--- a/scene/gui/popup.cpp
+++ b/scene/gui/popup.cpp
@@ -232,7 +232,8 @@ Size2 PopupPanel::_get_contents_minimum_size() const {
 
 void PopupPanel::_update_child_rects() {
 	Vector2 cpos(theme_cache.panel_style->get_offset());
-	Vector2 csize(get_size() - theme_cache.panel_style->get_minimum_size());
+	Vector2 panel_size = Vector2(get_size()) / get_content_scale_factor();
+	Vector2 csize = panel_size - theme_cache.panel_style->get_minimum_size();
 
 	for (int i = 0; i < get_child_count(); i++) {
 		Control *c = Object::cast_to<Control>(get_child(i));
@@ -246,7 +247,7 @@ void PopupPanel::_update_child_rects() {
 
 		if (c == panel) {
 			c->set_position(Vector2());
-			c->set_size(get_size());
+			c->set_size(panel_size);
 		} else {
 			c->set_position(cpos);
 			c->set_size(csize);
--- a/scene/gui/tree.cpp
+++ b/scene/gui/tree.cpp
@@ -3995,12 +3995,14 @@ bool Tree::edit_selected(bool p_force_edit) {
 		return false;
 	}
 
+	float popup_scale = popup_editor->is_embedded() ? 1.0 : popup_editor->get_parent_visible_window()->get_content_scale_factor();
 	Rect2 rect;
 	if (select_mode == SELECT_ROW) {
 		rect = s->get_meta("__focus_col_" + itos(selected_col));
 	} else {
 		rect = s->get_meta("__focus_rect");
 	}
+	rect.position *= popup_scale;
 	popup_edited_item = s;
 	popup_edited_item_col = col;
 
@@ -4043,7 +4045,7 @@ bool Tree::edit_selected(bool p_force_edit) {
 		popup_rect.size = rect.size;
 
 		// Account for icon.
-		Size2 icon_size = _get_cell_icon_size(c);
+		Size2 icon_size = _get_cell_icon_size(c) * popup_scale;
 		popup_rect.position.x += icon_size.x;
 		popup_rect.size.x -= icon_size.x;
 
@@ -4070,7 +4072,10 @@ bool Tree::edit_selected(bool p_force_edit) {
 		}
 
 		popup_editor->set_position(popup_rect.position);
-		popup_editor->set_size(popup_rect.size);
+		popup_editor->set_size(popup_rect.size * popup_scale);
+		if (!popup_editor->is_embedded()) {
+			popup_editor->set_content_scale_factor(popup_scale);
+		}
 		popup_editor->popup();
 		popup_editor->child_controls_changed();
 
@@ -4086,7 +4091,10 @@ bool Tree::edit_selected(bool p_force_edit) {
 		text_editor->show();
 
 		popup_editor->set_position(get_screen_position() + rect.position);
-		popup_editor->set_size(rect.size);
+		popup_editor->set_size(rect.size * popup_scale);
+		if (!popup_editor->is_embedded()) {
+			popup_editor->set_content_scale_factor(popup_scale);
+		}
 		popup_editor->popup();
 		popup_editor->child_controls_changed();
 
--- a/scene/main/viewport.cpp
+++ b/scene/main/viewport.cpp
@@ -1461,6 +1461,8 @@ void Viewport::_gui_show_tooltip() {
 	panel->set_flag(Window::FLAG_NO_FOCUS, true);
 	panel->set_flag(Window::FLAG_POPUP, false);
 	panel->set_flag(Window::FLAG_MOUSE_PASSTHROUGH, true);
+	// A non-embedded tooltip window will only be transparent if per_pixel_transparency is allowed in the main Viewport.
+	panel->set_flag(Window::FLAG_TRANSPARENT, true);
 	panel->set_wrap_controls(true);
 	panel->add_child(base_tooltip);
 	panel->gui_parent = this;
@@ -1469,17 +1471,25 @@ void Viewport::_gui_show_tooltip() {
 
 	tooltip_owner->add_child(gui.tooltip_popup);
 
+	Window *window = Object::cast_to<Window>(gui.tooltip_popup->get_embedder());
+	if (!window) { // Not embedded.
+		window = gui.tooltip_popup->get_parent_visible_window();
+	}
+	float win_scale = window->content_scale_factor;
 	Point2 tooltip_offset = GLOBAL_GET("display/mouse_cursor/tooltip_position_offset");
+	if (!gui.tooltip_popup->is_embedded()) {
+		tooltip_offset *= win_scale;
+	}
 	Rect2 r(gui.tooltip_pos + tooltip_offset, gui.tooltip_popup->get_contents_minimum_size());
-	r.size = r.size.min(panel->get_max_size());
-
-	Window *window = gui.tooltip_popup->get_parent_visible_window();
 	Rect2i vr;
 	if (gui.tooltip_popup->is_embedded()) {
 		vr = gui.tooltip_popup->get_embedder()->get_visible_rect();
 	} else {
+		panel->content_scale_factor = win_scale;
+		r.size *= win_scale;
 		vr = window->get_usable_parent_rect();
 	}
+	r.size = r.size.min(panel->get_max_size());
 
 	if (r.size.x + r.position.x > vr.size.x + vr.position.x) {
 		// Place it in the opposite direction. If it fails, just hug the border.
--- a/scene/gui/popup_menu.cpp
+++ b/scene/gui/popup_menu.cpp
@@ -312,18 +312,17 @@ int PopupMenu::_get_items_total_height() const {
 }
 
 int PopupMenu::_get_mouse_over(const Point2 &p_over) const {
-	if (p_over.x < 0 || p_over.x >= get_size().width || p_over.y < theme_cache.panel_style->get_margin(Side::SIDE_TOP)) {
+	float win_scale = get_content_scale_factor();
+	if (p_over.x < 0 || p_over.x >= get_size().width * win_scale || p_over.y < theme_cache.panel_style->get_margin(Side::SIDE_TOP) * win_scale) {
 		return -1;
 	}
 
-	Point2 ofs;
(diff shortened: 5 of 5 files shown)


scene/gui/popup.cpp
```diff
@@ -247,7 +247,9 @@ void PopupPanel::_update_child_rects() {
 
 		if (c == panel) {
 			c->set_position(Vector2());
-			c->set_size(get_size());
+			Vector2 panel_size = Vector2(get_size()) / this->get_content_scale_factor();
```
AThousandShips, line 250:
```suggestion
			Vector2 panel_size = get_size() / this->get_content_scale_factor();
```
No need to cast `Vector2` to `Vector2`

Koyper: `get_size()` on `Window` returns `Size2i`, so the divide winds up truncating the float into an int. Is there a more elegant way to handle this?

AThousandShips: Oh yes my bad, thought it was a `Control`, this is the correct way 🙂

Koyper: Okay, good.

scene/gui/popup.cpp
```diff
@@ -233,7 +233,7 @@ Size2 PopupPanel::_get_contents_minimum_size() const {
 
 void PopupPanel::_update_child_rects() {
 	Vector2 cpos(theme_cache.panel_style->get_offset());
-	Vector2 csize(get_size() - theme_cache.panel_style->get_minimum_size());
+	Vector2 csize; // There must always be `panel`, so safe to set csize below.
```
Sauermann, line 236: Is the assumption guaranteed, that "panel" is the first Control-child?
In cases, where "panel" is not the first child, the loop below can lead to unexpected results, since the size of Control nodes can be set, without the initialization of ´csize`.

The user can change the order of child nodes and can interfere with this assumption.
I believe, that it would be more robust, to initialize `csize` before the for-loop.

Koyper: Well-reasoned. I made the assumption that "panel" would always remain as the first child, not considering that the user could change the order. Is it safe to assume "panel" cannot be removed altogether?

Sauermann: Users removing internal nodes ... sounds like a recipe for disaster.
While these cases should not be expected, if a crash can be prevented with a simple `!= null` check, I believe, that it should be included.

KoBeWi: `panel` is an internal child, so user has no access to it unless they want to purposefully break the node. It can't be done by accident.

Koyper: I made the suggested change to set `csize` correctly, regardless of `panel` child index. Per KoBeWi and that I believe it is standard practice to not protect internal nodes, I did not add a `null` or `is_instance_valid` check. There are many warnings in the docs about bad consequences of deleting internal nodes, but those are usually exposed.

Koyper:
I have a related question regarding PR best practices: I've finished modifications to the remaining classes that did not respond correctly to `content_scale_factor` values other than 1.0. Should I post a separate PR for each of these or do them as a group in a single PR? The files modified are tree.cpp, popup_menu.cpp and dialog.cpp. Thanks!

akien-mga:
If it's the same bug in different related classes, I would suggest fixing all of them in the same PR (so they could be added to this one, expanding its scope a bit).

Koyper:
> If it's the same bug in different related classes, I would suggest fixing all of them in the same PR (so they could be added to this one, expanding its scope a bit). I'd go as far as suggesting to make the changes in a single commit (so amend this PR's commit with the similar changes to other Control classes).

So what about https://github.com/godotengine/godot/pull/86556 that would otherwise be included in this new group? Should I amend this PR with those changes and then close that PR?

akien-mga:
Yeah I think that would make sense. Fix all those related issues with one fell swoop.

Koyper:
I've amended this commit to add all the related changes required to implement proper scaling of non-embedded windows when `Viewport.content_scale_factor` is not equal to 1.0.

This includes tooltips, `PopupPanel`, `PopupMenu`, `Tree`, `ConfirmationDialog` and `AcceptDialog`.

scene/main/viewport.cpp
```diff
@@ -1527,17 +1529,22 @@ void Viewport::_gui_show_tooltip() {
 
 	tooltip_owner->add_child(gui.tooltip_popup);
 
+	Window *window = gui.tooltip_popup->get_parent_visible_window();
```
Sauermann, line 1532: In the case that `gui.tooltip_popup` is embedded, it is possible, that its embedder is not the parent visible window.

Can you please try:
```suggestion
	Window *window = gui.tooltip_popup->get_embedder();
	if (window == null) {
		window = ui.tooltip_popup->get_parent_visible_window();
	}
```

Koyper: Change made - thanks!

scgm0:
How's the progress of this PR?The problem of blurry sub-windows is really annoying...

Koyper:
I'll have to take care of the conflicts with a rebase, but other than that the PR is ready to merge, pending approvals, which I've been awaiting for a while.

I've tested it for MacOS and Windows and all is working correctly for me (I've been using it continuously since the PR was posted)

I should get to the rebase within a few days...

Koyper:
Ok, ready for review after fixing some issues with `PopupMenu` created by the removal of the internal margin container by #87462.

scene/gui/dialogs.cpp
```diff
@@ -202,7 +202,7 @@ void AcceptDialog::register_text_enter(LineEdit *p_line_edit) {
 }
 
 void AcceptDialog::_update_child_rects() {
-	Size2 dlg_size = get_size();
+	Size2 dlg_size = Vector2(get_size()) / this->get_content_scale_factor(); // get_size() is Size2i hence the cast to Vector2.
```
KoBeWi, line 205:
```suggestion
	Size2 dlg_size = Vector2(get_size()) / get_content_scale_factor();
```
`this` is unnecessary and the comment is somewhat obvious.

scene/gui/popup.cpp
```diff
@@ -232,7 +232,8 @@ Size2 PopupPanel::_get_contents_minimum_size() const {
 
 void PopupPanel::_update_child_rects() {
 	Vector2 cpos(theme_cache.panel_style->get_offset());
-	Vector2 csize(get_size() - theme_cache.panel_style->get_minimum_size());
+	Vector2 panel_size = Vector2(get_size()) / this->get_content_scale_factor();
```
KoBeWi, line 235:
```suggestion
	Vector2 panel_size = Vector2(get_size()) / get_content_scale_factor();
```
Same.

scene/gui/tree.cpp
```diff
@@ -3995,12 +3995,14 @@ bool Tree::edit_selected(bool p_force_edit) {
...
 	}
 
+	float popup_scale = popup_editor->is_embedded() ? 1.0 : popup_editor->get_parent_visible_window()->get_content_scale_factor();
 	Rect2 rect;
 	if (select_mode == SELECT_ROW) {
 		rect = s->get_meta("__focus_col_" + itos(selected_col));
 	} else {
 		rect = s->get_meta("__focus_rect");
 	}
+	rect.position = rect.position * popup_scale;
```
KoBeWi, line 4005:
```suggestion
	rect.position *= popup_scale;
```

scene/main/viewport.cpp
```diff
@@ -1469,17 +1471,25 @@ void Viewport::_gui_show_tooltip() {
 
 	tooltip_owner->add_child(gui.tooltip_popup);
 
+	Window *window = Object::cast_to<Window>(gui.tooltip_popup->get_embedder());
+	if (window == NULL) { // Is NULL if the Window is NOT embedded.
```
KoBeWi, line 1475:
```suggestion
	if (window == nullptr) { // Is NULL if the Window is NOT embedded.
```

scene/main/viewport.cpp
```diff
@@ -1461,6 +1461,8 @@ void Viewport::_gui_show_tooltip() {
 	panel->set_flag(Window::FLAG_NO_FOCUS, true);
 	panel->set_flag(Window::FLAG_POPUP, false);
 	panel->set_flag(Window::FLAG_MOUSE_PASSTHROUGH, true);
+	panel->set_flag(Window::FLAG_TRANSPARENT, true);
+	// A non-embedded tooltip window will only be transparent if per_pixel_transparency is allowed in the main Viewport.
```
KoBeWi, line 1465: This comment should be above the flag change.
Is this related to what this PR fixes though?

scene/main/viewport.cpp
```diff
@@ -1469,17 +1471,25 @@ void Viewport::_gui_show_tooltip() {
...
 	tooltip_owner->add_child(gui.tooltip_popup);
 
+	Window *window = Object::cast_to<Window>(gui.tooltip_popup->get_embedder());
+	if (window == NULL) { // Is NULL if the Window is NOT embedded.
+		window = gui.tooltip_popup->get_parent_visible_window();
+	}
+	float win_scale = window->content_scale_factor;
 	Point2 tooltip_offset = GLOBAL_GET("display/mouse_cursor/tooltip_position_offset");
+	if (!gui.tooltip_popup->is_embedded()) {
+		tooltip_offset = tooltip_offset * win_scale;
```
KoBeWi, line 1481:
```suggestion
		tooltip_offset *= win_scale;
```

scene/main/viewport.cpp
```diff
@@ -1469,17 +1471,25 @@ void Viewport::_gui_show_tooltip() {
...
 	Rect2 r(gui.tooltip_pos + tooltip_offset, gui.tooltip_popup->get_contents_minimum_size());
-	r.size = r.size.min(panel->get_max_size());
-
-	Window *window = gui.tooltip_popup->get_parent_visible_window();
 	Rect2i vr;
 	if (gui.tooltip_popup->is_embedded()) {
 		vr = gui.tooltip_popup->get_embedder()->get_visible_rect();
 	} else {
+		panel->content_scale_factor = win_scale;
+		r.size = r.size * win_scale;
```
KoBeWi, line 1489:
```suggestion
		r.size *= win_scale;
```

KoBeWi (approved):
I did some testing and it appears to be working correctly.
The code looks fine (aside from the style comments I left), but someone could double-check the viewport changes.

scene/main/viewport.cpp
akien-mga: Or just
```suggestion
	if (!window ) { // Not embedded.
```

scene/main/viewport.cpp
Koyper: It's true that the PR would work fine without this, but you could not have transparent backgrounds in the non-embedded tooltips. Should I remove this and at some point do a separate PR as a bug-fix?

My sense is that because this PR makes non-embedded tooltips workable (because you couldn't use them previously for projects on HiDPI displays), no one would have come accross this bug.

Koyper:
Thanks for the comments! I've made the suggested changes, and for now left in the transparency flag fix, but can remove if recommended.

akien-mga:
Thanks!

scene/main/viewport.cpp
akien-mga: This seems to have caused a regression where editor tooltips are transparent unexpectedly: https://github.com/godotengine/godot/issues/92712#issuecomment-2261752610

bruvzg: > A non-embedded tooltip window will only be transparent if per_pixel_transparency is allowed in the main Viewport.

It's allowed in the project settings not for a Viewport, and only affect requested surface format on window creation, but nothing prevents GPU driver from always using surface with alpha channel. And seems like Windows DisplayServer do not have any additional checks (unlike X11 and macOS) and will try to enable it regardless of project settings.

PR #86555, opened 2023-12-27
Fix LightmapperRD division warning in MSVC

RandomShaper:
This one (MSVC 14.3):
```
modules\lightmapper_rd\lightmapper_rd.cpp(360) : error C2220: the following warning is treated as an error
modules\lightmapper_rd\lightmapper_rd.cpp(360) : warning C4723: potential divide by 0
```

The value can't be zero with the expected values of the variables involved, though.


clayjohn (approved):

akien-mga:
Cherry-picked for 4.2.2.

PR #86561, opened 2023-12-27
CommandQueueMT: Fuse flush methods (flush_if_pending() was not thread-safe)

RandomShaper:


RandomShaper:
It seems `flush_if_pending()` has a tacit contract to be only used for the server thread for itself, so this change is not necessary. Archiving.

PR #85248, opened 2023-11-22
Fix lambda cross-thread dynamics (take 2)

RandomShaper:
This PR has two commits:
1. Reverts recent attempts to fix a chain of issues, which was making the whole thing extremely complex.
2. Implements an alternative to #84659 (the first reverted one), much simpler, now I've been lucid enough to realize how to do it properly. I'm muuuch more satisfied with this approach. (Reviewers are advised to check the diff of just this one, to be back to the starting point.)

Supersedes #85223.
Fixes #85151.

Tested the following recent related bugs are still gone after this PR (plus no address sanitizer issues detected);
- #84046
- #84190
- #85112

Issue #85151, opened 2023-11-20
GDScript engine destruction causes deadlock when destroying lambdas


Efimero:
### Godot version

v4.2.rc.custom_build [dfd61cdac]

### System information

Debian, Vulkan (Forward+)

### Issue description

<details><summary>Stack Trace</summary>

```
#0  futex_wait (private=0, expected=2, futex_word=0x55556474d280) at ../sysdeps/nptl/futex-internal.h:146
#1  __GI___lll_lock_wait (futex=futex@entry=0x55556474d280, private=0) at ./nptl/lowlevellock.c:49
#2  0x00007ffff7d61682 in lll_mutex_lock_optimized (mutex=0x55556474d280) at ./nptl/pthread_mutex_lock.c:48
#3  ___pthread_mutex_lock (mutex=0x55556474d280) at ./nptl/pthread_mutex_lock.c:93
#4  0x0000555557f673cd in __gthread_mutex_lock (__mutex=0x55556474d280) at /usr/include/x86_64-linux-gnu/c++/13/bits/gthr-default.h:749
#5  0x0000555557f6741d in __gthread_recursive_mutex_lock (__mutex=0x55556474d280) at /usr/include/x86_64-linux-gnu/c++/13/bits/gthr-default.h:811
#6  0x0000555557f67452 in std::recursive_mutex::lock (this=0x55556474d280) at /usr/include/c++/13/mutex:120
#7  0x0000555557f67d4f in std::unique_lock<std::recursive_mutex>::lock (this=0x7fffffffcc70) at /usr/include/c++/13/bits/unique_lock.h:141
#8  0x0000555557f67c7b in std::unique_lock<std::recursive_mutex>::unique_lock (this=0x7fffffffcc70, __m=...)
    at /usr/include/c++/13/bits/unique_lock.h:71
#9  0x00005555582a1d67 in MutexLock<MutexImpl<std::recursive_mutex> >::MutexLock (p_mutex=..., this=<optimized out>) at ./core/os/mutex.h:122
#10 GDScript::_remove_func_ptr_to_update (p_func_ptr_element=0x555564749d30) at modules/gdscript/gdscript.cpp:1424
#11 0x000055555848002c in GDScriptLambdaCallable::~GDScriptLambdaCallable (this=0x5555646bd690, __in_chrg=<optimized out>)
    at modules/gdscript/gdscript_lambda_callable.cpp:155
[...]

Issue #85112, opened 2023-11-19
Lambda/Callable memory leak


gokiburikin:
### Godot version

v4.2.beta6.official [64150060f]

### System information

Godot v4.2.beta6 - Windows 10.0.19044 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 970 (NVIDIA; 31.0.15.3623) - Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz (16 Threads)

### Issue description

Was testing benchmarks on the latest RC and this snippet started leaking memory as of 4.2beta6 up to the master branch. Does not happen in any version I tested from 4.1 stable to 4.2beta5.

https://github.com/godotengine/godot/assets/3820082/86283a44-2ce9-46a0-a60b-3365c016fca8

### Steps to reproduce

```gdscript
func _process(delta: float) -> void:
	for i in 10000: (func(): randf()).call()
	print(OS.get_static_memory_usage())
```

### Minimal reproduction project

N/A

gokiburikin:
Assuming I followed the bisecting steps correctly it leads back to this https://github.com/godotengine/godot/commit/271511726b02848783904429c8dde857f6266429.

Chaosus:
cc @RandomShaper / @adamscott

Issue #84190, opened 2023-10-30
Lambda call within a non-main thread forces Godot to crash at exit


Chaosus:
### Godot version

9144457484f9d2f53990a0aac37caff1c9012e6d

### System information

Windows 11

### Issue description

Running the following code:

```gdscript
extends Node3D

var thread : Thread
var array = [3, 2, 1]

func _ready():
	thread = Thread.new()
	thread.start(do_some_work)
	thread.wait_to_finish()
	print(array)

func do_some_work() -> void:
	array.sort_custom(func(a, b): return a < b)
```

executes correctly, but when user closes the application or call get_tree().quit() the Godot will crash with the following call stack:
```
 	godot.windows.editor.x86_64.exe!GDScript::clear(GDScript::ClearData * p_clear_data) Line 1440	C++
 	godot.windows.editor.x86_64.exe!GDScript::~GDScript() Line 1518	C++
 	godot.windows.editor.x86_64.exe!GDScript::`scalar deleting destructor'(unsigned int)	C++
 	[Inline Frame] godot.windows.editor.x86_64.exe!memdelete(GDScript * p_class) Line 109	C++
 	godot.windows.editor.x86_64.exe!Ref<GDScript>::unref() Line 210	C++
 	[Inline Frame] godot.windows.editor.x86_64.exe!Ref<GDScript>::{dtor}() Line 222	C++
 	[Inline Frame] godot.windows.editor.x86_64.exe!memdelete(HashMapElement<String,Ref<GDScript>> *) Line 109	C++
 	[Inline Frame] godot.windows.editor.x86_64.exe!DefaultTypedAllocator<HashMapElement<String,Ref<GDScript>>>::delete_allocation(HashMapElement<String,Ref<GDScript>> * p_allocation) Line 206	C++
 	godot.windows.editor.x86_64.exe!HashMap<String,Ref<GDScript>,HashMapHasherDefault,HashMapComparatorDefault<String>,DefaultTypedAllocator<HashMapElement<String,Ref<GDScript>>>>::clear() Line 265	C++
 	godot.windows.editor.x86_64.exe!GDScriptCache::clear() Line 445	C++
>	godot.windows.editor.x86_64.exe!GDScriptLanguage::finish() Line 2116	C++
 	godot.windows.editor.x86_64.exe!ScriptServer::finish_languages() Line 229	C++
 	godot.windows.editor.x86_64.exe!Main::cleanup(bool p_force) Line 3763	C++
[...]

PR #86569, opened 2023-12-28
GDScript: Lambda hotswap fixes

rune-scape:
im glad https://github.com/godotengine/godot/pull/85248 got merged and fixed my mistakes from https://github.com/godotengine/godot/pull/81628
however,
it has some problems that this PR fixes:
 - `thread_enter` and `thread_exit` are inadequate for initializing the threadlocal lists, as external thread created by, [FMOD](https://github.com/alessandrofama/fmod-for-godot) for example, arent tracked at all, leading to a segfault when it tries to access the threadlocal list
 - https://github.com/godotengine/godot/pull/85248 is, no offense, over complicated, manual refcounting is unnecessary and error prone, just keeping track of the lists should be used instead of moving orphaned list elements to the main thread list

and,, another thing fixed by this:
 - my original code was fundamentally broken
 - it nulls lambdas unrelated to the currently reloading script, making hot-reloading lambdas still broken, just safer, no crashing
 - just alot safer, regular mutex and data member arrangement,, easy to wrap head around

crucially it works on my project that uses lambdas and needs hot-reloading to properly update them, where the current master commit fails,, lambdas just become invalid for all but 1 script
keeping only the relevant updateable func ptrs in each script with a mutex for each is enough to make this fast and safe in lambda creation, and script reloading and clearing.
and easier to read and modify the code
it should function as i meant it to originally in https://github.com/godotengine/godot/pull/81628 and now waste no time checking unrelated elements when reloading

the following recent related bugs are still gone after this PR:
 - https://github.com/godotengine/godot/issues/84046
 - https://github.com/godotengine/godot/issues/84190
 - https://github.com/godotengine/godot/issues/85112

but i could not reproduce https://github.com/godotengine/godot/issues/85151, if someone who can reproduce it can test that its still fixed after this PR i would be so grateful
sorry for the wait

### DIFF
--- a/modules/gdscript/gdscript.cpp
+++ b/modules/gdscript/gdscript.cpp
@@ -1381,51 +1381,43 @@ String GDScript::debug_get_script_name(const Ref<Script> &p_script) {
 }
 #endif
 
-GDScript::UpdatableFuncPtr GDScript::func_ptrs_to_update_main_thread;
-thread_local GDScript::UpdatableFuncPtr *GDScript::func_ptrs_to_update_thread_local = nullptr;
-
-GDScript::UpdatableFuncPtrElement GDScript::_add_func_ptr_to_update(GDScriptFunction **p_func_ptr_ptr) {
-	UpdatableFuncPtrElement result = {};
-
-	{
-		MutexLock lock(func_ptrs_to_update_thread_local->mutex);
-		result.element = func_ptrs_to_update_thread_local->ptrs.push_back(p_func_ptr_ptr);
-		result.func_ptr = func_ptrs_to_update_thread_local;
-
-		if (likely(func_ptrs_to_update_thread_local->initialized)) {
-			return result;
-		}
-
-		func_ptrs_to_update_thread_local->initialized = true;
+GDScript::UpdatableFuncPtr::UpdatableFuncPtr(GDScriptFunction *p_function) {
+	if (p_function == nullptr) {
+		return;
 	}
 
+	ptr = p_function;
+	script = ptr->get_script();
+	ERR_FAIL_NULL(script);
+
+	MutexLock script_lock(script->func_ptrs_to_update_mutex);
+	list_element = script->func_ptrs_to_update.push_back(this);
+}
+
+GDScript::UpdatableFuncPtr::~UpdatableFuncPtr() {
+	ERR_FAIL_NULL(script);
+
+	if (list_element) {
+		MutexLock script_lock(script->func_ptrs_to_update_mutex);
+		list_element->erase();
+		list_element = nullptr;
+	}
+}
+
+void GDScript::_recurse_replace_function_ptrs(const HashMap<GDScriptFunction *, GDScriptFunction *> &p_replacements) const {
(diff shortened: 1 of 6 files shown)


rune-scape:
trying to find out why only the linux builds are failing..

rune-scape:
```
#0  0x000000000a2119dd in GDScript::UpdatableFuncPtrOwnerListContainer::UpdatableFuncPtrOwnerListContainer (
    this=0x7fffff400d80) at modules/gdscript/gdscript.cpp:1397
1397    GDScript::UpdatableFuncPtrOwnerListContainer::UpdatableFuncPtrOwnerListContainer() {
```
oops it seems threadlocal var initializers arent re-entrant safe...

i rewrote it to be way simpler anyway,

RandomShaper:
> * #85248 is, no offense, over complicated
You should have seen its first iteration. 😆

Now seriously, I did my best to understand the intent of the original code so I could provide a fix. I had to make a few assumptions. I'm happy there's a proper patch now.

rune-scape:
@RandomShaper good job, it was pretty bad code,,, 
i should have gone with simplest option to begin with :)

modules/gdscript/gdscript_lambda_callable.h
```diff
@@ -42,10 +42,9 @@ class GDScriptFunction;
 class GDScriptInstance;
 
 class GDScriptLambdaCallable : public CallableCustom {
-	GDScriptFunction *function = nullptr;
+	GDScript::UpdateableFuncPtr updatable_fn;
```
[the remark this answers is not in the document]
rune-scape: renamed the type back to Updatable 👍

RandomShaper (approved):
LGTM. There are change requests by others, I agree with, though.

modules/gdscript/gdscript.cpp
```diff
@@ -1381,51 +1381,44 @@ String GDScript::debug_get_script_name(const Ref<Script> &p_script) {
...
+		list_element = nullptr;
+	}
+}
 
-	MutexLock lock(func_ptrs_to_update_main_thread.mutex);
-	MutexLock lock2(func_ptrs_to_update_thread_local->mutex);
+void GDScript::_recurse_replace_function_ptrs(const HashMap<GDScriptFunction *, GDScriptFunction *> &p_replacements) const {
+	MutexLock outer_lock(func_ptrs_to_update_mutex);
+	for (UpdatableFuncPtr *updatable : func_ptrs_to_update) {
+		MutexLock lock(func_ptrs_to_update_mutex);
```
AThousandShips, line 1410: This locks the same lock again, what was meant to be locked here?

AThousandShips (changes requested):

modules/gdscript/gdscript.cpp
rune-scape: oops, the inner lock became redundant, forgot to remove it

AThousandShips (approved):

akien-mga:
Thanks!

rune-scape:
this should probably be cherry-picked for 4.2 ,, ill make  a commit

PR #86573, opened 2023-12-28
Iterator improvements

AThousandShips:
* Nullability safety for increment/decrement
* Style improvements

These are minor fixes and mostly style but also ensures iteration doesn't throw segfault when iterating past the end, this matches how it is done in `HashMap` which makes sense to apply everywhere IMO

Can limit it to just null fixes if desired 🙂


RandomShaper:
Looks good. I'd only suggest using `likely()`.

PR #86576, opened 2023-12-28
Add warning for using concave shape on CharacterBody3D

### LINKED ISSUE #86574: No warnings when using ConcavePolygonShape3D with CharacterBody3D
### Tested versions

4.3dev1

### System information

Windows 10 - Godot4.3dev1

### Issue description

There are no warnings when using ConcavePolygonShape3D with CharacterBody3D. It seems there are relevant warning comments in the documentation

### Steps to reproduce

Add a CollisionShape3D node under CharacterBody3D, and use ConcavePolygonShape3D as the Shape.

N/A

bikemurt:
As per #86574 

Tested a simple scene to confirm that ConcaveShapes do not work for collisions on CharacterBody3D (same as RigidBody3D).

Added the following node config warning:
image

I didn't update the language documentation as I assume that's handled separately

*Bugsquad edit:*
- Fixes #86574

### DIFF
--- a/scene/3d/collision_shape_3d.cpp
+++ b/scene/3d/collision_shape_3d.cpp
@@ -35,6 +35,7 @@
 #include "scene/resources/concave_polygon_shape_3d.h"
 #include "scene/resources/convex_polygon_shape_3d.h"
 #include "scene/resources/world_boundary_shape_3d.h"
+#include "vehicle_body_3d.h"
 
 void CollisionShape3D::make_convex_from_siblings() {
 	Node *p = get_parent();
@@ -130,10 +131,21 @@ PackedStringArray CollisionShape3D::get_configuration_warnings() const {
 	}
 
 	if (shape.is_valid() && Object::cast_to<RigidBody3D>(col_object)) {
+		String body_type = "RigidBody3D";
+		if (Object::cast_to<VehicleBody3D>(col_object)) {
+			body_type = "VehicleBody3D";
+		}
+
 		if (Object::cast_to<ConcavePolygonShape3D>(*shape)) {
-			warnings.push_back(RTR("ConcavePolygonShape3D doesn't support RigidBody3D in another mode than static."));
+			warnings.push_back(vformat(RTR("When used for collision, ConcavePolygonShape3D is intended to work with static CollisionObject3D nodes like StaticBody3D.\nIt will likely not behave well for %ss (except when frozen and freeze_mode set to FREEZE_MODE_STATIC)."), body_type));
 		} else if (Object::cast_to<WorldBoundaryShape3D>(*shape)) {
 			warnings.push_back(RTR("WorldBoundaryShape3D doesn't support RigidBody3D in another mode than static."));
+		}
+	}
+
+	if (shape.is_valid() && Object::cast_to<CharacterBody3D>(col_object)) {
+		if (Object::cast_to<ConcavePolygonShape3D>(*shape)) {
+			warnings.push_back(RTR("When used for collision, ConcavePolygonShape3D is intended to work with static CollisionObject3D nodes like StaticBody3D.\nIt will likely not behave well for CharacterBody3Ds."));
 		}
 	}
 


AThousandShips (dismissed):
This wording isn't correct, it should be more clear, `CharacterBody3D` does support this shape, it just isn't recommended:
> When used for collision, ConcavePolygonShape3D is intended to work with static [CollisionShape3D](https://docs.godotengine.org/en/stable/classes/class_collisionshape3d.html#class-collisionshape3d) nodes like [StaticBody3D](https://docs.godotengine.org/en/stable/classes/class_staticbody3d.html#class-staticbody3d) and will likely not behave well for [CharacterBody3D](https://docs.godotengine.org/en/stable/classes/class_characterbody3d.html#class-characterbody3d)s or [RigidBody3D](https://docs.godotengine.org/en/stable/classes/class_rigidbody3d.html#class-rigidbody3d)s in a mode other than Static.

The warning should match the one for `RigidBody3D` instead

milkiq:
> ConcavePolygonShape3D doesn't support RigidBody3D in another mode than static.

RigidBody3D uses this wording, so would we change both of them?

bikemurt:
> > ConcavePolygonShape3D doesn't support RigidBody3D in another mode than static.
> 
> RigidBody3D uses this wording, so would we change both of them?

Initially I found this description to be wordy, but I see what @AThousandShips is saying - you _can_ use concave shapes, it just won't work with moving bodies.

AThousandShips:
The warning should be descriptive, so it shouldn't be changed for `RigidBody3D` IMO, but both could be updated, on second thought I'm not 100% sure what is meant by the static mode here, *or* if it applies to both types, the documentation for the shape should probably be improved to clarify this 🙂

I would suggest to update both descriptions to be clearer that it's not recommended to use these

AThousandShips:
The physics team would have to confirm but I think the documentation for the shape needs to be made more clear, something like:
> When used for collision, ConcavePolygonShape3D is intended to work with static CollisionShape3D nodes like StaticBody3D and will likely not behave well for or RigidBody3Ds (except when frozen and freeze_mode set to FREEZE_MODE_STATIC) or CharacterBody3Ds.

Or something like that

And then change all the wordings here to align with that, saying something like "not recommended"

bikemurt:
I just did a little bit of testing by using a RigidBody3D (with convex collision shape) and pushing it into a RigidBody3D and CharacterBody3D.

When the RigidBody3D is set to freeze/static, the concave shape does actually work for collisions. It behaves exactly like a StaticBody3D in that case.

I think CharacterBody3D does not work at all with a concave collision shape (as far as I can tell). It doesn't interact with a StaticBody3D (my floor) or the RigidBody3D with a convex shape.


> > When used for collision, ConcavePolygonShape3D is intended to work with static CollisionShape3D nodes like StaticBody3D and will likely not behave well for or RigidBody3Ds (except when frozen and freeze_mode set to FREEZE_MODE_STATIC) or CharacterBody3Ds.

I like this.  Could split out the description for RigidBody3D and CharacterBody3D:

RigidBody3D:

`When used for collision, ConcavePolygonShape3D is intended to work with static CollisionShape3D nodes like StaticBody3D and will likely not behave well for or RigidBody3Ds (except when frozen and freeze_mode set to FREEZE_MODE_STATIC).`

CharacterBody3D:

`When used for collision, ConcavePolygonShape3D is intended to work with static CollisionShape3D nodes like StaticBody3D and will likely not behave well for CharacterBody3Ds.`

bikemurt:
Here's what I have for now. Also added a new line so it looks a little cleaner:

image

milkiq:
> Here's what I have for now. Also added a new line so it looks a little cleaner:这是我现在所拥有的。还添加了一个新行，因此看起来更干净一些：
> 
> image

It's looks nice! Easier to understand than before

scene/3d/collision_shape_3d.cpp
```diff
@@ -131,12 +131,18 @@ PackedStringArray CollisionShape3D::get_configuration_warnings() const {
 
 	if (shape.is_valid() && Object::cast_to<RigidBody3D>(col_object)) {
 		if (Object::cast_to<ConcavePolygonShape3D>(*shape)) {
-			warnings.push_back(RTR("ConcavePolygonShape3D doesn't support RigidBody3D in another mode than static."));
+			warnings.push_back(RTR("When used for collision, ConcavePolygonShape3D is intended to work with static CollisionShape3D nodes like StaticBody3D. \nIt will likely not behave well for or RigidBody3Ds (except when frozen and freeze_mode set to FREEZE_MODE_STATIC)."));
```
AThousandShips, line 134:
```suggestion
			warnings.push_back(RTR("When used for collision, ConcavePolygonShape3D is intended to work with static CollisionObject3D nodes like StaticBody3D.\nIt will likely not behave well for RigidBody3Ds (except when frozen and freeze_mode set to FREEZE_MODE_STATIC)."));
```
The documentation uses the wrong type here, should be fixed along with updating that wording

scene/3d/collision_shape_3d.cpp
```diff
@@ -131,12 +131,18 @@ PackedStringArray CollisionShape3D::get_configuration_warnings() const {
...
 		} else if (Object::cast_to<WorldBoundaryShape3D>(*shape)) {
 			warnings.push_back(RTR("WorldBoundaryShape3D doesn't support RigidBody3D in another mode than static."));
 		}
 	}
 
+	if (shape.is_valid() && Object::cast_to<CharacterBody3D>(col_object)) {
+		if (Object::cast_to<ConcavePolygonShape3D>(*shape)) {
+			warnings.push_back(RTR("When used for collision, ConcavePolygonShape3D is intended to work with static CollisionShape3D nodes like StaticBody3D. \nIt will likely not behave well for CharacterBody3Ds."));
```
AThousandShips, line 142:
```suggestion
			warnings.push_back(RTR("When used for collision, ConcavePolygonShape3D is intended to work with static CollisionObject3D nodes like StaticBody3D.\nIt will likely not behave well for CharacterBody3Ds."));
```

Calinou (commented):
I suggest adding the same warning for VehicleBody3D and VehicleWheel3D, as these have the same constraints due to their dynamic nature.

scene/3d/collision_shape_3d.cpp
```diff
@@ -130,13 +131,22 @@ PackedStringArray CollisionShape3D::get_configuration_warnings() const {
 	}
 
 	if (shape.is_valid() && Object::cast_to<RigidBody3D>(col_object)) {
+		String body_type = "RigidBody3D";
+		if (Object::cast_to<VehicleBody3D>(col_object)) body_type = "VehicleBody3D";
```
AThousandShips, line 135:
```suggestion
		if (Object::cast_to<VehicleBody3D>(col_object)) {
			body_type = "VehicleBody3D";
		}
```

bikemurt:
> I suggest adding the same warning for VehicleBody3D and VehicleWheel3D, as these have the same constraints due to their dynamic nature.

Makes sense. I took another look, and it appears that VehicleBody3D is (currently) the only node that inherits RigidBody3D. So I updated the code to use a string substitution in the RTR - not sure if we prefer writing out the strings verbosely.

Also, to me it seems that VehicleWheel3D shouldn't have a CollisionShape3D added to it since it doesn't inherit CollisionObject3D, so I didn't add a warning for that.

Calinou (approved):

AThousandShips (approved):

akien-mga:

bikemurt:
Sorry - missed that in the documentation for amending commits. Will do it that way next time

akien-mga:

PR #86583, opened 2023-12-28
Tiny fix for lightmapper DDA

reduz:
- Ensures only one axis advances at a time
- This fixes extremely corner cases where the DDA may skip over geometry
- As a bonus, clarify to the user that splitting into more meshes may improve the texture size limit.

### DIFF
--- a/editor/plugins/lightmap_gi_editor_plugin.cpp
+++ b/editor/plugins/lightmap_gi_editor_plugin.cpp
@@ -105,7 +105,7 @@ void LightmapGIEditorPlugin::_bake_select_file(const String &p_file) {
 				EditorNode::get_singleton()->show_warning(TTR("Lightmap data is not local to the scene."));
 			} break;
 			case LightmapGI::BAKE_ERROR_TEXTURE_SIZE_TOO_SMALL: {
-				EditorNode::get_singleton()->show_warning(TTR("Maximum texture size is too small for the lightmap images."));
+				EditorNode::get_singleton()->show_warning(TTR("Maximum texture size is too small for the lightmap images.\nWhile this can be fixed by increasing the maximum texture size, it is recommended you split the scene into more objects instead."));
 			} break;
 			default: {
 			} break;
--- a/modules/lightmapper_rd/lm_compute.glsl
+++ b/modules/lightmapper_rd/lm_compute.glsl
@@ -264,7 +264,16 @@ uint trace_ray(vec3 p_from, vec3 p_to, bool p_any_hit, out float r_distance, out
 			break;
 		}
 
-		bvec3 mask = lessThanEqual(side.xyz, min(side.yzx, side.zxy));
+		// There should be only one axis updated at a time for DDA to work properly.
+		bvec3 mask = bvec3(true, false, false);
+		float m = side.x;
+		if (side.y < m) {
+			m = side.y;
+			mask = bvec3(false, true, false);
+		}
+		if (side.z < m) {
+			mask = bvec3(false, false, true);
+		}
 		side += vec3(mask) * delta;
 		icell += ivec3(vec3(mask)) * step;
 		iters++;


modules/lightmapper_rd/lm_compute.glsl
```diff
@@ -264,7 +264,17 @@ uint trace_ray(vec3 p_from, vec3 p_to, bool p_any_hit, out float r_distance, out
 			break;
 		}
 
-		bvec3 mask = lessThanEqual(side.xyz, min(side.yzx, side.zxy));
+		// There should be only one axis updated at a time for DDA to work properly.
+		// bvec3 mask = lessThanEqual(side.xyz, min(side.yzx, side.zxy));
```
AThousandShips, line 268:
```suggestion
		//bvec3 mask = lessThanEqual(side.xyz, min(side.yzx, side.zxy));
```
Commented code should be without space to differentiate

editor/plugins/lightmap_gi_editor_plugin.cpp
```diff
@@ -105,7 +105,7 @@ void LightmapGIEditorPlugin::_bake_select_file(const String &p_file) {
 				EditorNode::get_singleton()->show_warning(TTR("Lightmap data is not local to the scene."));
 			} break;
 			case LightmapGI::BAKE_ERROR_TEXTURE_SIZE_TOO_SMALL: {
-				EditorNode::get_singleton()->show_warning(TTR("Maximum texture size is too small for the lightmap images."));
+				EditorNode::get_singleton()->show_warning(TTR("Maximum texture size is too small for the lightmap images.\nWhile this can be fixed by increasing the maximum texture size, it is recommended you split the scene in more objects instead."));
```
AThousandShips, line 108:
```suggestion
				EditorNode::get_singleton()->show_warning(TTR("Maximum texture size is too small for the lightmap images.\nWhile this can be fixed by increasing the maximum texture size, it is recommended you split the scene into more objects instead."));
```

modules/lightmapper_rd/lm_compute.glsl
reduz: Thanks!

jcostello:
Error while building

```
editor/plugins/lightmap_gi_editor_plugin.cpp:116:6: error: 'LightmapGIEditormPlugin' has not been declared
  116 | void LightmapGIEditormPlugin::_bake() {
      |      ^~~~~~~~~~~~~~~~~~~~~~~
editor/plugins/lightmap_gi_editor_plugin.cpp: In function 'void _bake()':
editor/plugins/lightmap_gi_editor_plugin.cpp:117:9: error: '_bake_select_file' was not declared in this scope
  117 |         _bake_select_file("");
      |         ^~~~~~~~~~~~~~~~~
```

jcostello:
Looks they are working much better now. I tried on sponza and in the TPS demo

modules/lightmapper_rd/lm_compute.glsl
```diff
@@ -264,7 +264,17 @@ uint trace_ray(vec3 p_from, vec3 p_to, bool p_any_hit, out float r_distance, out
 			break;
 		}
 
-		bvec3 mask = lessThanEqual(side.xyz, min(side.yzx, side.zxy));
+		// There should be only one axis updated at a time for DDA to work properly.
+		//bvec3 mask = lessThanEqual(side.xyz, min(side.yzx, side.zxy));
```
clayjohn, line 268:
```suggestion
```

clayjohn (commented):
Makes sense. Using ``lessThanEqual`` could return multiple channels where the channels are equal. We can't move to ``lessThan`` as we could end up with zero channels and then we are stuck. 

You just need to remove the old line before this is ready to merge.

reduz:
This should now be ready

clayjohn (approved):

akien-mga:
Thanks!

PR #86584, opened 2023-12-28
Fix `AudioServer::start_playback_stream` does not iterate through given bus volumes

MJacred:
`AudioServer::set_playback_bus_volumes_linear` properly loops through the given `p_bus_volumes`.

This fixes an inconsistency. But it does not fix any bugs, because, currently, all users (the audio players) of this func only ever send 1 bus. So actually, right now, the `HashMap` is unnecessary in
```c++
void start_playback_stream(Ref<AudioStreamPlayback> p_playback, HashMap<StringName, Vector<AudioFrame>> p_bus_volumes, float p_start_time = 0, float p_pitch_scale = 1, float p_highshelf_gain = 0, float p_attenuation_cutoff_hz = 0);
```
A simple `StringName p_bus`+`Vector<AudioFrame> p_volume_db_vector` combination would suffice


AThousandShips (approved):
Makes sense

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #86587, opened 2023-12-28
Enhance & fix `WorkerThreadPool`

### LINKED ISSUE #78790: Error on exit about pages still in use for Thread and WorkerThreadPool::Task
### Godot version

4.1.rc (9a09ee5d0217c4d5e5668fccddfa477bf16af357)

### System information

Mageia 9, Linux

### Issue description

On some projects (exact characteristics TBD, my MRP uses a VideoStreamPlayer that autoplays a video with sound), the following errors are printed in the console when closing the project:

```
ERROR: Pages in use exist at exit in PagedAllocator: 6Thread
   at: ~PagedAllocator (./core/templates/paged_allocator.h:170)
ERROR: Pages in use exist at exit in PagedAllocator: N16WorkerThreadPool4TaskE
   at: ~PagedAllocator (./core/templates/paged_allocator.h:170)
```

It happens both when closing the window with the cross button, or Alt+F4. It doesn't happen all the times, and might be related to quitting very early on (i.e. start, then close shortly after the scene starts).

### Steps to reproduce

- Run MRP directly, quit shortly after the window initialized and the scene starts rendering / playing the video.
- It may print the above errors in the terminal. If not, try again a few times.

[VideoScene.zip](https://github.com/godotengine/godot/files/11893613/VideoScene.zip)
(MRP derived from https://github.com/godotengine/godot/issues/78788, but I've seen this in other scenes not using VideoStreamPlayer.)

RandomShaper:
A number of changes to the `WorkerThreadPool`, split into commits with specific goals:

**2023-01-05:** Heavily edited after squashing two commits together and greatly polishing some changes.

- ### WorkerThreadPool: Overhaul scheduling and synchronization

This commits rewrites the sync logic in a way that the `use_system_threads_for_low_priority_tasks` setting, which was added due to the lack of a cross-platform wait-for-multiple-objects functionality, can be removed (it's as if it was effectively hardcoded to `false`).

With the new implementation, we have the best of both worlds: threads don't have to poll, plus no bespoke threads are used.

In addition, regarding deadlock prevention, since not every possible case of wait-deadlock could be avoided, this commits removes the current best-effort avoidance mechanisms and keeps only a simple, pessimistic way of detection.

It turns out that the only current user of deadlock prevention, ResourceLoader, works fine with it and so every possible situation in resource loading is now properly handled, with no possibilities of deadlocking. There's a comment in the code with further details.

Lastly, a potential for load tasks never being awaited/disposed is cleared.

- ### WorkerThreadPool: Avoid deadlocks when `CommandQueueMT` is involved

This commit lets `CommandQueueMT` play nicely with the WorkerThreadPool to avoid non-progressable situations caused by an interdependence between both. While a command queue is being flushed, it allows the `WorkerThreadPool` to release
the lock of the former while tasks are being awaited so they can make progress in case they need in turn to post to the command queue.

- ### WorkerThreadPool: Avoid most runtime allocations

Just a little optimization.

**NOTE:**
With `RID_Owner` we could replace each pair of `PagedAllocator` and `HashMap`-of-ids-to-pointers. However, that would force us to expose `RID` as the task/group id, instead of `int`, which would break the API. Too bad. Let's wait until Godot 5.0.

---

~Includes #86555.~
Fixes #78790.

---

Testing appreciated. I'll keep testing it on different projects.

### DIFF
--- a/core/object/worker_thread_pool.cpp
+++ b/core/object/worker_thread_pool.cpp
@@ -33,6 +33,7 @@
 #include "core/object/script_language.h"
 #include "core/os/os.h"
 #include "core/os/thread_safe.h"
+#include "core/templates/command_queue_mt.h"
 
 void WorkerThreadPool::Task::free_template_userdata() {
 	ERR_FAIL_NULL(template_userdata);
@@ -42,6 +43,8 @@ void WorkerThreadPool::Task::free_template_userdata() {
 }
 
 WorkerThreadPool *WorkerThreadPool::singleton = nullptr;
+
+thread_local CommandQueueMT *WorkerThreadPool::flushing_cmd_queue = nullptr;
 
 void WorkerThreadPool::_process_task(Task *p_task) {
 	int pool_thread_index = thread_ids[Thread::get_caller_id()];
@@ -428,7 +431,15 @@ Error WorkerThreadPool::wait_for_task_completion(TaskID p_task_id) {
 
 					if (!task_to_process) {
 						caller_pool_thread->awaited_task = task;
+
+						if (flushing_cmd_queue) {
+							flushing_cmd_queue->unlock();
+						}
 						caller_pool_thread->cond_var.wait(lock);
+						if (flushing_cmd_queue) {
+							flushing_cmd_queue->lock();
+						}
+
 						DEV_ASSERT(exit_threads || caller_pool_thread->signaled || task->completed);
 						caller_pool_thread->awaited_task = nullptr;
 					}
@@ -540,7 +551,14 @@ void WorkerThreadPool::wait_for_group_task_completion(GroupID p_group) {
 
 	{
 		Group *group = *groupp;
+
+		if (flushing_cmd_queue) {
+			flushing_cmd_queue->unlock();
+		}
 		group->done_semaphore.wait();
+		if (flushing_cmd_queue) {
+			flushing_cmd_queue->lock();
+		}
 
 		uint32_t max_users = group->tasks_used + 1; // Add 1 because the thread waiting for it is also user. Read before to avoid another thread freeing task after increment.
 		uint32_t finished_users = group->finished.increment(); // fetch happens before inc, so increment later.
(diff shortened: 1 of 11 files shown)


fire:
The changes to the xml documentation are straightforward changes, but any idea why the asan and such is failing.

RandomShaper:
Marking as _Needs work_ because I'm rethinking one of the changes a bit, as I've found room for extra improvement.

RandomShaper:
Now I'm happy with this.

reduz (approved):
Looks good, I like the new approach!

core/object/worker_thread_pool.cpp
```diff
@@ -150,88 +132,163 @@ void WorkerThreadPool::_process_task(Task *p_task) {
...
-		if (post) {
-			task_available_semaphore.post();
-		}
 	}
+
+	set_current_thread_safe_for_nodes(safe_for_nodes_backup);
 }
 
 void WorkerThreadPool::_thread_function(void *p_user) {
+	ThreadData &thread_data = singleton->threads[singleton->get_thread_index()];
```
AThousandShips, line 168: This fails if the return is `-1`, when the thread id is not present

RandomShaper: This case has turned out to be very interesting! At first, I thought that, since this function only serves as the entry point of the thread, there's no possibility the id is not found. Then, I realized that, given how the ids map and the threads array are initialized, there is a race condition. Finally, I decided to directly pass the pointer to the relevant thread data structure via user data. Then, and this is the amusing thing, I found that the threads are already created with `&threads[i]` passed as user data, and that was the case when the WTP was first written, even if it wasn't used then! It's as if Juan had left it for me in the future.

AThousandShips: The adventures in code! Good detective work!

clayjohn:
I just cancelled CI, the tests that run the unit tests appeared to be totally stuck. They had been running for 5 hours when I cancelled the workflow

AThousandShips:
Sounds like a deadlock somewhere, suspect the threads aren't joined correctly

RandomShaper:
I've reproduced it locally. Thankfully the current tests stress the thing enough to make the issue apparent. When group tasks are involved in huge amounts, some low-prio tasks are kept in the queue with no threads realizing the need to claim them. Under investigation. 🔍

RandomShaper:
OK. I've found and fixed the logic error.

PR #86588, opened 2023-12-28
fix GDScript style guide violation in ConfigFile documentation

poyribn:
the third script example in the
[documentation for ConfigFile](https://docs.godotengine.org/en/4.2/classes/class_configfile.html) had a spot that violates the
[GDScript style guide](https://docs.godotengine.org/en/4.2/tutorials/scripting/gdscript/gdscript_styleguide.html) , section
[Boolean Operators](https://docs.godotengine.org/en/4.2/tutorials/scripting/gdscript/gdscript_styleguide.html#boolean-operators) :

> Prefer the plain English versions of boolean operators, as they are
> the most accessible:
> - use ``not`` instead of ``!``

```GDScript
if err != OK:
	return
```

after this commit, this spot no longer violates the GDScript style guide:

```GDScript
if err == not OK:
	return
```

— sosasees <https://github.com/sosasees>

### DIFF
--- a/doc/classes/ConfigFile.xml
+++ b/doc/classes/ConfigFile.xml
@@ -51,7 +51,7 @@
 		var err = config.load("user://scores.cfg")
 
 		# If the file didn't load, ignore it.
-		if err != OK:
+		if err == not OK:
 		    return
 
 		# Iterate over all sections.


sosasees:
i believe this change is sensible to port to Godot 3.x 🙂

doc/classes/ConfigFile.xml
```diff
@@ -51,7 +51,7 @@
 		var err = config.load("user://scores.cfg")
 
 		# If the file didn't load, ignore it.
-		if err != OK:
+		if err == not OK:
```
AThousandShips, line 54: This is not correct, the style guide says to use:
```gdscript
if not X:
```
Instead of 
```gdscript
if !X:
```

This is different

This is not a valid change

AThousandShips (changes requested):
As per the above, this is wrong

Thank you for your contribution nonetheless, this however isn't correct

AThousandShips:
Closing as the code is valid and correct, and this change is wrong, nothing needs to be changed about this code

PR #86598, opened 2023-12-29
Project upgrade tool: Better space handling in "export"

pancelor:
The regexes to convert "export" lines from godot 3 to godot 4 had two small issues: they weren't recognizing export statements with 1. spaces after "export" or 2. no space before "var".

These are both valid syntax, but the conversion tool did not recognize them. This commit fixes both cases.

test case 1:    `export(int)var x = 1`
old conversion: `@export(int)var x = 1`
new conversion: `@export var x: int = 1`

test case 2:    `export (int) var y = 2`
old conversion: `@export (int) var y = 2`
new conversion: `@export var y: int = 2`

These were tested by converting projects using the Import button in the project manager.

### DIFF
--- a/editor/project_converter_3_to_4.cpp
+++ b/editor/project_converter_3_to_4.cpp
@@ -68,10 +68,10 @@ public:
 	RegEx reg_json_parse = RegEx("([\t ]{0,})([^\n]+)parse_json\\(([^\n]+)");
 	RegEx reg_json_non_new = RegEx("([\t ]{0,})([^\n]+)JSON\\.parse\\(([^\n]+)");
 	RegEx reg_json_print = RegEx("\\bJSON\\b\\.print\\(");
-	RegEx reg_export_simple = RegEx("export\\(([a-zA-Z0-9_]+)\\)[ ]+var[ ]+([a-zA-Z0-9_]+)");
-	RegEx reg_export_typed = RegEx("export\\(([a-zA-Z0-9_]+)\\)[ ]+var[ ]+([a-zA-Z0-9_]+)[ ]*:[ ]*[a-zA-Z0-9_]+");
-	RegEx reg_export_inferred_type = RegEx("export\\([a-zA-Z0-9_]+\\)[ ]+var[ ]+([a-zA-Z0-9_]+)[ ]*:[ ]*=");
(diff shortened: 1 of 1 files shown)


editor/project_converter_3_to_4.cpp
```diff
@@ -68,10 +68,10 @@ class ProjectConverter3To4::RegExContainer {
 	RegEx reg_json_parse = RegEx("([\t ]{0,})([^\n]+)parse_json\\(([^\n]+)");
 	RegEx reg_json_non_new = RegEx("([\t ]{0,})([^\n]+)JSON\\.parse\\(([^\n]+)");
 	RegEx reg_json_print = RegEx("\\bJSON\\b\\.print\\(");
-	RegEx reg_export_simple = RegEx("export\\(([a-zA-Z0-9_]+)\\)[ ]+var[ ]+([a-zA-Z0-9_]+)");
-	RegEx reg_export_typed = RegEx("export\\(([a-zA-Z0-9_]+)\\)[ ]+var[ ]+([a-zA-Z0-9_]+)[ ]*:[ ]*[a-zA-Z0-9_]+");
-	RegEx reg_export_inferred_type = RegEx("export\\([a-zA-Z0-9_]+\\)[ ]+var[ ]+([a-zA-Z0-9_]+)[ ]*:[ ]*=");
-	RegEx reg_export_advanced = RegEx("export\\(([^)^\n]+)\\)[ ]+var[ ]+([a-zA-Z0-9_]+)([^\n]+)");
+	RegEx reg_export_simple = RegEx("export[ ]*\\(([a-zA-Z0-9_]+)\\)[ ]*var[ ]+([a-zA-Z0-9_]+)");
```
AThousandShips, line 71: I don't think this is valid, this would match things like `export (int)var` and `export (int) var` which are invalid I'd say, it's supposed to match `export(int) var`

AThousandShips, line 71: This will match `exportvar` which isn't correct

AThousandShips (approved):
LGTM, can't confirm at the moment that all the combinations are valid syntax, but it doesn't really matter in this case as the results are fixed properly

akien-mga:
Thanks!

PR #86610, opened 2023-12-29
Add option for editor to follow system theme and accent colors

Joseph-DiGiovanni:
As suggested by godotengine/godot-proposals#5678, I've implemented the following:

- Automatic editor theme switching on system theme change.
- Optional accent color matching (Windows and MacOS only).
- Individual theme preset selection for light and dark mode.

Before going into more detail, here are the results:

[Screencast from 2023-12-29 08-19-26.webm](https://github.com/godotengine/godot/assets/37851452/e0d0ec1a-97ef-45c1-83fa-371b29a73229)

Screenshot 2023-12-29 102609

I've only tested it on Linux and Windows, but it should work just fine on any OS. The editor settings now feature an option to have the editor theme follow the system dark mode settings. The light and dark themes default to 2 new presets called "Custom Light" and "Custom Dark" which can be configured individually to your liking, but any existing preset may be used (e.g., Solarized Light and Solarized Dark). You may also select a light and dark theme resource file which will override the function of the existing "Custom Theme" resource when the new presets are used.

I also took the opportunity to refactor a significant portion of the editor theme creation code for improved readability and to better facilitate adding these features.

The only questionable bit is my method of detecting the switch to and from dark mode. I'm currently storing the previous values of the dark mode and accent color check in EditorNode and checking for changes on process. I am absolutely open to any suggestions on how to better implement this.

### DIFF
--- a/doc/classes/EditorSettings.xml
+++ b/doc/classes/EditorSettings.xml
@@ -703,6 +703,9 @@
 		<member name="interface/theme/draw_extra_borders" type="bool" setter="" getter="">
 			If [code]true[/code], draws additional borders around interactive UI elements in the editor. This is automatically enabled when using the [b]Black (OLED)[/b] theme preset, as this theme preset uses a fully black background.
 		</member>
+		<member name="interface/theme/follow_system_theme" type="bool" setter="" getter="">
+			If [code]true[/code], the editor theme preset will attempt to automatically match the system theme.
+		</member>
 		<member name="interface/theme/icon_and_font_color" type="int" setter="" getter="">
 			The icon and font color scheme to use in the editor.
 			- [b]Auto[/b] determines the color scheme to use automatically based on [member interface/theme/base_color].
@@ -721,6 +724,10 @@
 		</member>
 		<member name="interface/theme/spacing_preset" type="String" setter="" getter="">
 			The editor theme spacing preset to use. See also [member interface/theme/base_spacing] and [member interface/theme/additional_spacing].
+		</member>
+		<member name="interface/theme/use_system_accent_color" type="bool" setter="" getter="">
+			If [code]true[/code], set accent color based on system settings.
+			[b]Note:[/b] This setting is only effective on Windows and MacOS.
 		</member>
 		<member name="interface/touchscreen/enable_long_press_as_right_click" type="bool" setter="" getter="">
 			If [code]true[/code], long press on touchscreen is treated as right click.
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -665,6 +665,8 @@ void EditorNode::_notification(int p_what) {
 
 			callable_mp(this, &EditorNode::_begin_first_scan).call_deferred();
 
+			DisplayServer::get_singleton()->set_system_theme_change_callback(callable_mp(this, &EditorNode::_update_theme));
+
 			/* DO NOT LOAD SCENES HERE, WAIT FOR FILE SCANNING AND REIMPORT TO COMPLETE */
 		} break;
 
@@ -767,6 +769,9 @@ void EditorNode::_notification(int p_what) {
 			FileDialog::set_default_show_hidden_files(EDITOR_GET("filesystem/file_dialog/show_hidden_files"));
 			EditorFileDialog::set_default_show_hidden_files(EDITOR_GET("filesystem/file_dialog/show_hidden_files"));
 			EditorFileDialog::set_default_display_mode((EditorFileDialog::DisplayMode)EDITOR_GET("filesystem/file_dialog/display_mode").operator int());
+
+			follow_system_theme = EDITOR_GET("interface/theme/follow_system_theme");
+			use_system_accent_color = EDITOR_GET("interface/theme/use_system_accent_color");
 
 			if (EditorThemeManager::is_generated_theme_outdated()) {
 				_update_theme();
@@ -3069,6 +3074,35 @@ void EditorNode::_save_screenshot(NodePath p_path) {
 	ERR_FAIL_COND_MSG(img.is_null(), "Cannot get an image from a viewport texture of the editor main screen.");
 	Error error = img->save_png(p_path);
 	ERR_FAIL_COND_MSG(error != OK, "Cannot save screenshot to file '" + p_path + "'.");
+}
+
+void EditorNode::_check_system_theme_changed() {
+	DisplayServer *display_server = DisplayServer::get_singleton();
+
+	bool system_theme_changed = false;
+
+	if (follow_system_theme) {
+		if (display_server->get_base_color() != last_system_base_color) {
+			system_theme_changed = true;
+			last_system_base_color = display_server->get_base_color();
+		}
+
+		if (display_server->is_dark_mode_supported() && display_server->is_dark_mode() != last_dark_mode_state) {
+			system_theme_changed = true;
+			last_dark_mode_state = display_server->is_dark_mode();
+		}
+	}
+
+	if (use_system_accent_color) {
+		if (display_server->get_accent_color() != last_system_accent_color) {
+			system_theme_changed = true;
+			last_system_accent_color = display_server->get_accent_color();
+		}
+	}
+
+	if (system_theme_changed) {
+		_update_theme();
+	}
 }
 
 void EditorNode::_tool_menu_option(int p_idx) {
@@ -7527,6 +7561,15 @@ EditorNode::EditorNode() {
 	String exec = OS::get_singleton()->get_executable_path();
 	// Save editor executable path for third-party tools.
 	EditorSettings::get_singleton()->set_project_metadata("editor_metadata", "executable_path", exec);
+
+	follow_system_theme = EDITOR_GET("interface/theme/follow_system_theme");
+	use_system_accent_color = EDITOR_GET("interface/theme/use_system_accent_color");
+	system_theme_timer = memnew(Timer);
+	system_theme_timer->set_wait_time(1.0);
+	system_theme_timer->connect("timeout", callable_mp(this, &EditorNode::_check_system_theme_changed));
+	add_child(system_theme_timer);
+	system_theme_timer->set_owner(get_owner());
+	system_theme_timer->set_autostart(true);
 }
 
 EditorNode::~EditorNode() {
--- a/editor/editor_settings.cpp
+++ b/editor/editor_settings.cpp
@@ -464,11 +464,13 @@ void EditorSettings::_load_defaults(Ref<ConfigFile> p_extra_config) {
 	EDITOR_SETTING(Variant::FLOAT, PROPERTY_HINT_RANGE, "interface/inspector/float_drag_speed", 5.0, "0.1,100,0.01")
 
 	// Theme
+	EDITOR_SETTING(Variant::BOOL, PROPERTY_HINT_ENUM, "interface/theme/follow_system_theme", false, "")
 	EDITOR_SETTING(Variant::STRING, PROPERTY_HINT_ENUM, "interface/theme/preset", "Default", "Default,Breeze Dark,Godot 2,Gray,Light,Solarized (Dark),Solarized (Light),Black (OLED),Custom")
 	EDITOR_SETTING(Variant::STRING, PROPERTY_HINT_ENUM, "interface/theme/spacing_preset", "Default", "Compact,Default,Spacious,Custom")
 	EDITOR_SETTING(Variant::INT, PROPERTY_HINT_ENUM, "interface/theme/icon_and_font_color", 0, "Auto,Dark,Light")
 	EDITOR_SETTING(Variant::COLOR, PROPERTY_HINT_NONE, "interface/theme/base_color", Color(0.2, 0.23, 0.31), "")
 	EDITOR_SETTING(Variant::COLOR, PROPERTY_HINT_NONE, "interface/theme/accent_color", Color(0.41, 0.61, 0.91), "")
+	EDITOR_SETTING(Variant::BOOL, PROPERTY_HINT_NONE, "interface/theme/use_system_accent_color", false, "")
 	EDITOR_SETTING(Variant::FLOAT, PROPERTY_HINT_RANGE, "interface/theme/contrast", 0.3, "-1,1,0.01")
 	EDITOR_SETTING(Variant::BOOL, PROPERTY_HINT_NONE, "interface/theme/draw_extra_borders", false, "")
 	EDITOR_SETTING(Variant::FLOAT, PROPERTY_HINT_RANGE, "interface/theme/icon_saturation", 1.0, "0,2,0.01")
--- a/editor/editor_node.h
+++ b/editor/editor_node.h
@@ -357,6 +357,13 @@ private:
 
 	Ref<Theme> theme;
 
+	Timer *system_theme_timer = nullptr;
+	bool follow_system_theme = false;
+	bool use_system_accent_color = false;
+	bool last_dark_mode_state = false;
+	Color last_system_base_color = Color(0, 0, 0, 0);
+	Color last_system_accent_color = Color(0, 0, 0, 0);
+
 	PopupMenu *recent_scenes = nullptr;
 	String _recent_scene;
 	List<String> previous_scenes;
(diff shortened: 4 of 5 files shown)


Calinou:
> I've only tested it on Linux and Windows, but it should work just fine on any OS. 

Note that the Android editor could also benefit from this feature, but this can be left for a future PR.

editor/editor_node.cpp
```diff
@@ -638,6 +638,32 @@ void EditorNode::_notification(int p_what) {
 					EditorFileSystem::get_singleton()->scan();
 				}
 			}
+
+			bool system_theme_changed = false;
+
+			bool follow_system_theme = EDITOR_GET("interface/theme/system_theme/follow_system_theme");
```
Calinou, line 644: Reading project settings every frame should be avoided for performance reasons. I suggest using the `editor_settings_changed` signal we have somewhere to store this value to a variable, and only read from that variable instead.

Also, I suggest not performing this check every frame, but rather every second or so to reduce resource usage. This can be done using a counter variable or a Timer node.

doc/classes/EditorSettings.xml
```diff
@@ -699,6 +699,90 @@
...
+		<member name="interface/theme/system_theme/custom_light_preset/custom_light_theme" type="String" setter="" getter="">
+			The custom theme resource to use for the editor in system light mode. Must be a Godot theme resource in [code].tres[/code] or [code].res[/code] format.
+		</member>
+		<member name="interface/theme/system_theme/custom_light_preset/draw_extra_borders" type="bool" setter="" getter="">
+			If [code]true[/code], draws additional borders around interactive UI elements in the editor. This is automatically enabled when using the [b]Black (OLED)[/b] theme preset, as this theme preset uses a fully black background.
+		</member>
+		<member name="interface/theme/system_theme/custom_light_preset/icon_and_font_color" type="int" setter="" getter="">
+			The icon and font color scheme to use in the editor.
+			- [b]Auto[/b] determines the color scheme to use automatically based on [member interface/theme/base_color].
+			- [b]Dark[/b] makes fonts and icons dark (suitable for light themes). Icon colors are automatically converted by the editor following the set of rules defined in [url=
```
Calinou, line 765: The "Dark" description ends abruptly with an unclosed `url` tag.

doc/classes/EditorSettings.xml
```diff
@@ -699,6 +699,90 @@
...
+			The editor theme preset to use in system dark mode.
+		</member>
+		<member name="interface/theme/system_theme/follow_system_theme" type="bool" setter="" getter="">
+			Set theme based on system light/dark mode.
+		</member>
+		<member name="interface/theme/system_theme/light_preset" type="String" setter="" getter="">
+			The editor theme preset to use in system light mode.
+		</member>
+		<member name="interface/theme/system_theme/use_system_accent_color" type="bool" setter="" getter="">
+			Set accent color based on system settings. (Windows and MacOS only)
```
Calinou, line 784:
```suggestion
			Set accent color based on system settings.
			[b]Note:[/b] This setting is only effective on Windows and macOS.
```

Joseph-DiGiovanni:
I used a timer node as @Calinou suggested checking only once per second.

I also changed the method of triggering the theme update from manually calling `_update_theme()` to simply changing the relevant settings as the user would, relying on the `editor_settings_changed` signal to call `_update_theme()` so it doesn't run twice. This eliminates the extra second for the theme to completely change automatically as seen in the video above.

Thanks for the feedback!

Calinou:
Tested locally, it mostly works as expected.

There are some usability quirks though:

- Changing the editor theme preset listed at the top of the editor settings has no effect if **Follow System Theme** is enabled. This is because the theme preference takes priority over the theme preset.
- With this PR, we now have **3** custom themes defined: Custom Dark, Custom Light and Custom:
- **Black (OLED)** is listed below the custom dark/light options, but it should be listed above.

image

To resolve these issues, I propose removing the existing theme preset entirely and keeping only the **Light Preset** and **Dark Preset** themes. The theme in use is set by an enum property called **Theme Mode** that has the values **Automatic** (= follow system theme), **Light** and **Dark** (default).

If you customize an existing theme, it switches the currently active theme mode to **Custom**.

Joseph-DiGiovanni:
I agree with your proposals, I didn't want to step on anyone's toes which is why I implemented them as separate presets.

> installing custom theme presets from a file is currently not really a supported use case, as it involves overwriting all your editor settings.

The option for custom theme files is there in the stable release. Are you implying it doesn't work and should be removed?

Joseph-DiGiovanni:
I'm happy to report I was able to implement it almost exactly as you described. It will also read the old preset (even custom ones!) and set the new theme options accordingly. No users should notice a difference unless they open the theme settings.

[Screencast from 2024-01-02 00-27-28.webm](https://github.com/godotengine/godot/assets/37851452/9aea9ab5-903f-4493-bfa6-ffcbebee79af)

Calinou:
> The option for custom theme files is there in the stable release. Are you implying it doesn't work and should be removed?

Syntax highlighting themes can be shared as `.tet` files, but editor theme presets can't. Only custom editor themes (which have to overwrite everything or replicate the entire editor theme in some way) are supported in the form of `.res`/`.tres` files.

doc/classes/EditorSettings.xml
```diff
@@ -677,9 +677,86 @@
...
+		</member>
+		<member name="interface/theme/custom_dark_preset/custom_theme" type="String" setter="" getter="">
+			The custom theme resource to use for the editor. Must be a Godot theme resource in [code].tres[/code] or [code].res[/code] format.
+		</member>
+		<member name="interface/theme/custom_dark_preset/draw_extra_borders" type="bool" setter="" getter="">
+			If [code]true[/code], draws additional borders around interactive UI elements in the editor. This is automatically enabled when using the [b]Black (OLED)[/b] theme preset, as this theme preset uses a fully black background.
+		</member>
+		<member name="interface/theme/custom_dark_preset/icon_and_font_color" type="int" setter="" getter="">
+			The icon and font color scheme to use in the editor.
+			- [b]Auto[/b] determines the color scheme to use automatically based on [member interface/theme/base_color].
```
Calinou, line 706:
```suggestion
			- [b]Automatic (Follow System Theme)[/b] determines the color scheme to use automatically based on [member interface/theme/base_color].
```

editor/editor_settings.cpp
```diff
@@ -459,10 +459,18 @@ void EditorSettings::_load_defaults(Ref<ConfigFile> p_extra_config) {
 	EDITOR_SETTING(Variant::FLOAT, PROPERTY_HINT_RANGE, "interface/inspector/float_drag_speed", 5.0, "0.1,100,0.01")
 
 	// Theme
-	EDITOR_SETTING(Variant::STRING, PROPERTY_HINT_ENUM, "interface/theme/preset", "Default", "Default,Breeze Dark,Godot 2,Gray,Light,Solarized (Dark),Solarized (Light),Black (OLED),Custom")
+
+	// Please keep these sorted alphabetically.
+	String dark_presets = "Black (OLED),Breeze Dark,Godot 2,Gray,Solarized (Dark)";
+	String light_presets = "Solarized (Light)";
+
+	EDITOR_SETTING(Variant::INT, PROPERTY_HINT_ENUM, "interface/theme/theme_mode", 0, "Dark,Light,Auto")
```
Calinou, line 467:
```suggestion
	EDITOR_SETTING(Variant::INT, PROPERTY_HINT_ENUM, "interface/theme/theme_mode", 0, "Dark,Light,Automatic (Follow System Theme)")
```

Calinou (approved):
It works great now, exactly as I expected. Thanks a ton for following through :slightly_smiling_face:

doc/classes/EditorSettings.xml
Joseph-DiGiovanni: This value is determined by the theme colors themselves rather than the system theme, and I've considered it as part of the theme itself as it's included in the presets.

Calinou: Apologies, I pointed to the wrong description.

editor/editor_settings.cpp
Calinou: We have plenty of space to display option values, so I don't think it's too much of an issue. We could name the option just `Follow System Theme` too.

Joseph-DiGiovanni: I don't mean to be difficult about such a minor thing, but I think consistency in the options is important as we already have "Dark, Light, Auto" as options for the icon and font color setting. I'd certainly meet in the middle and say "System" could replace "Auto" if you think this option needs more clarity. Here's some screenshots:

Screenshot from 2024-01-03 21-08-47
Screenshot from 2024-01-03 21-09-52

YuriSizov (changes requested):
This needs to be reworked following the recent editor theme generation changes.

But more importantly, I find this solution to be overly complicated. I don't see the rationale for duplicating every theme property and having one set for each of dark and light modes. I don't expect people two change their system colors often, especially to change it often between light and dark modes.

Besides, on Linux at least we will probably want to try and pick the most fitting preset from our other options, not just our default light and default dark.

So all we need to add is a new "System" preset which will automatically resolve into either the Default (dark) or the Light preset on Windows and macOS. And on Linux, if we can, it can resolve to a more specific color mode, since this has been requested by users, or default to the same logic as on Windows and macOS.

Joseph-DiGiovanni:
> I don't see the rationale for duplicating every theme property

The custom themes need to be stored somewhere, so at minimum you will always need 2 sets of theme settings, and the original settings which are considered the "active" theme settings can be changed at any time and work as the user expects. It really is the simplest possible solution for the user.

> So all we need to add is a new "System" preset which will automatically resolve into either the Default (dark) or the Light preset

This is the simplest implementation possible but you lose any ability to actually customize the theme.

YuriSizov:
> The custom themes need to be stored somewhere

What custom themes? Why? We don't do that right now, if you select a preset, all your custom settings get reset.

Joseph-DiGiovanni:
> Why? We don't do that right now

Because we never had to switch between 2 different themes before this feature. Otherwise any customization would be lost the moment the system theme changes and it changes the preset. This would be very annoying for people who switch between light and dark mode throughout the day.

> Color presets only affect 4 properties, which are expected to follow the preset. The rest you can customize.

Those 4 properties are arguably the most important part of the theme customization for the average user and should be separately configurable for dark and light mode.

YuriSizov:
I think this is excessive. If you want the editor to follow your system theme, it is expected that editor theme's color properties will be updated automatically. If you want to customize the colors to fit your own preference, then you no longer follow the system theme.

People also don't generally update their system color mode often if at all, so most of what you suggest seems like a non-issue to me. As in, if you decide to customize your theme and then change the system preference at some point, it's not going to be an issue to adjust your editor settings once again. Doing this whole setup just for that rare occasion is too much and makes things much more convoluted to maintain.

Joseph-DiGiovanni:
> If you want the editor to follow your system theme

It doesn't follow the system theme as much as simply acknowledge the dark/light mode setting. Besides the original title of the PR the settings in the current state of the PR make this very clear. As a side note, the Default theme doesn't match any major OS so Gray would be more appropriate if this was the goal.

> People also don't generally update their system color mode often

MacOS has an auto option when selecting the dark/light system preference that changes with the time of day. There are applications that add this feature to windows and linux as well.

> makes things much more convoluted to maintain.

I think I addressed this pretty well. The "preset" option has just been replaced with a dark preset and light preset. The fact that there are a few clearly labeled duplicated settings only used by the theme generator doesn't realistically add anything extra to maintain. Not having them results in the total loss of customization. I think it's a fair trade off.

YuriSizov:
> MacOS has an auto option when selecting the dark/light system preference that changes with the time of day. There are applications that add this feature to windows and linux as well.

This is an interesting use case, but it's still a niche application for tripling the amount of settings, some of which don't even need to be associated with the color mode.

We can discuss the need to customize base and accent color on top of the following dark/light polarity of the system theme in a follow-up. For the initial feature this is too much granularity.

> As a side note, the Default theme doesn't match any major OS so Gray would be more appropriate if this was the goal.

If we can read values for the base and accent colors from the OS, we can automatically set them, like in case of Windows accent color. This doesn't need to be explicitly exposed, unless someone requests it. That's the main problem. This assumes a lot of customization is needed before we even implemented the base mechanic for following the system theme.

----

PS. I also noticed right now that you automatically update the editor theme when the system theme changes (and with a hack for some reason). This shouldn't be done. Regenerating the entire theme takes a noticeable amount of time. We don't want users to experience hitching in the middle of their work.

I suggest replacing it with some notification which can be clicked to confirm the change. But we also can simply keep the theme until the user restarts the application.

Joseph-DiGiovanni:
> We can discuss the need to customize base and accent color on top of following dark/light polarity of the system theme in a follow-up.

That's probably a good idea. I'm working on more basic PR right this moment.

> If we can read values for the base and accent colors from the OS

We can only read accent color. I don't think reading any kind of "base" color of the OS is even possible.

> you automatically update the editor theme when the system theme changes
> We don't want users to experience hitching in the middle of their work.

As a user of auto dark/light mode, even I will concede every app flipping colors automatically is mildly jarring. This feature would not be enabled by default and if the user manually enabled it the second of delay once a day with an obvious reason for the hitch is frankly not an issue to me.

In reference to the hack in my implementation it was required due to a bug in the way settings changes were detected so the theme change was called twice (You can see this in the original PR video). I haven't checked but hopefully your PR has fixed that. In any case if I'm just modifying the preset value as you suggest this should thankfully not be necessary.

YuriSizov:
> In reference to the hack in my implementation it was required due to a bug in the way settings changes were detected so the theme change was called twice (You can see this in the original PR video). I haven't checked but hopefully your PR has fixed that.

Let me know if you need any help with that, here or on Rocket.Chat 👍

bruvzg:
> We can only read accent color. I don't think reading any kind of "base" color of the OS is even possible.

What's a `base` color? Control background color? It should be possible to get it the same way we are reading accent color. 

It should be possible to get callback for the theme change (at least on Windows and macOS).

YuriSizov:
> What's a `base` color? Control background color?

Let's say yes. Since we use it as a base for other colors, we may want to adjust it (because we have darker and lighter variations). But anything you can get from the OS should be useful.

Joseph-DiGiovanni:
I've tracked down the extra _update_theme() call to the theme generation code emitting settings changed signals when it really shouldn't. I talked privately with @YuriSizov about it and it seems the inspector UI elements wrongfully still rely on this signal so they don't update properly without it. Until that is fixed I'd consider it an unrelated bug.

Joseph-DiGiovanni:
I've tracked down the settings changed signal to the set_initial_value() method called many times in the editor theme manager. I talked privately with @YuriSizov about it and it seems it is unintended behavior however the inspector UI elements wrongfully still rely on this signal so they don't update properly without it. Until that is fixed I'd consider it an unrelated bug.

bruvzg:
PR for callback and control bg color  - https://github.com/godotengine/godot/pull/87384, base color might need some testing (some other color name might be better), but unfortunately immersive colors aren't documented at all.

akien-mga:
IIUC this could be redone now to make use of https://github.com/godotengine/godot/pull/87384.

Joseph-DiGiovanni:
I have implemented base color matching from #87384 and cleaned up my code a bit. I've tested it on Windows 11 and it works as expected. Thanks! Although as stated previously, without system theme change detection that triggers the callback being implemented on every platform it makes more sense to just manually check for system theme changes as was done previously.

While this probably should have started as a draft I appreciate all the help that has been given and believe this should be considered for merging. Thank you.

Calinou (approved):
Tested locally on Fedora 39 KDE X11, it works as expected.

However, the theme preset switching takes a fair amount of time as it requires two updates for some reason (notice how the Tree background color is initially not updated). This doesn't occur when manually switching theme presets as shown near the end of the video below.

https://github.com/godotengine/godot/assets/180032/24373033-9c39-4939-b117-32c5bc1216d4

I'm sad to see the ability to choose different presents go away (I'd use Light + Black (OLED) personally), but it makes sense in the interest of simplicity.

Joseph-DiGiovanni:
> the theme preset switching takes a fair amount of time as it requires two updates for some reason

This is a bug unrelated to my code where any theme update not caused by a manual settings change signal actually emits a setting change signal during theme generation, resulting in 2 theme updates. This is easily fixed by making EditorSettings::set_initial_value not emit a signal but the inspector backgrounds do not update without it (which is why they only change on the second update).

These issues didn't present themselves until now because there was previously no way for a theme change to occur without changing settings manually. I'd be happy to open an issue with these details once this gets merged as the fix goes beyond the scope of this PR.

akien-mga:

PR #86616, opened 2023-12-29
Add Light2D blend modes for add and subtract with a maximum (clamped)

### LINKED ISSUE #7026: Add Light2D Blend Mode: Min and Max

A top-view puzzle game with various light sources in a dark place

I'm adding light sources on the player character and torches in the level, but in Add blending mode, it stacks up very fast and it gets too bright as soon as character approaches a torch.

So I switched to Mix blending mode, which seemed to Max at first, but when one light really gets brighter than the other, a weaker light drags the final lightness down. According to the doc, this actually uses linear interpolation.

I'd like to preserve the Max lightning from all light sources. It doesn't need to be a perfect max, a smoothed curve that reduces contribution of further lights when there are more and more would also work (for instance, a logarithmic progression). As such, a custom formula blending mode where user inputs what they want or points to a custom script would also work.

And if we add Max, we may as well add Min although I don't have a particular need for it right now.

According to this thread:
https://ask.godotengine.org/111101/achieving-light2d-with-blending-using-maximum-value-instead

this can be accomplished using the existing GL_MAX and GL_MIN value.

Add two Blend Modes: Min and Max.

Alternatively, a custom Blend Mode where you can write your own formula by iterating on some array of light values (more complex)

**Max blending mode**

When two light masks are intersecting the max value of all light sources is used for the final light value.

Since I discovered a bug in the Mix blend that makes blending assymmetrical in the meantime, I'm gonna use that bug to take a screenshot which coincidentally gives me the result I want in some cases:

image

See how 1.7 takes over 1.3 without adding extra light.

**Min blending mode**

Same thing, but the smallest value takes over. Again, I exploited a bug in the Mix blending mode to capture this:

image

Btw I'm gonna report the bug, but I can definitely not rely on it in my game since Mix will sometimes use Min, sometimes use Max depending on the order of creation of the Light2D, and at runtime the order may change again.

No, I can write my own shader to grab to the max value of some combination of values but that won't be simple. I used stencil once to make custom masks but I'm not sure you can use multiple values of them.

While a custom shader could do, it's more convenient to reuse the existing Light2D node and since Blending Mode is an enum, internal code is the best place to add a new value.

bgie:
Add Light2D blend modes for add and substract with a maximum (clamped)

Fixes godotengine/godot-proposals#7026

The proposal mentions the word 'minimum' for sake of symmetry with the word 'maximum', yet I opted to only do a maximum, but for both the additive and subtractive lights. ('add_max' and 'sub_max'). So using maximum for both, as this applies to the maximal amount of adding or subtracting the light will do. A minimum would be something different, resulting in only the intersection of multiple lights being lit, which is something I have not seen anyone ask for explicitly. A minimum would be difficult to implement with regards to combinations with regular and maximum lights.

Notes on implementation:
- the old add/sub modes are independent of the ordering of the lights. I made the new clamped add/sub also order independent. Mix mode was never order independent, and I did not touch it.
- the clamped add/sub lights are applied separately from the regular add/sub/mix lights. This allows combining both stacking and non-stacking lights in a deterministic way. The new type of non-stacking lights are applied last. This is only relevant with regards to the mix mode, but I would not advise combining mix mode with any other mode.
- you can combine clamped additive and subtractive lights. They cancel each other out. But the maxima are respected.
- the performance cost for games that use 2D lights but not these new modes is non-zero but very small: for every pixel three vec3 variables are initialized to zeros and a single add with clamp is done with these zero'd variables. I did not consider it necessary to provide a shader variant. (Also, implementing that is currently outside my comfort zone =).

Tested this on my debian linux laptop with a NVIDIA Quadro T500 Mobile using the 3 render modes.


scene/2d/light_2d.cpp
```diff
@@ -291,7 +291,7 @@ void Light2D::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "editor_only"), "set_editor_only", "is_editor_only");
 	ADD_PROPERTY(PropertyInfo(Variant::COLOR, "color"), "set_color", "get_color");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "energy", PROPERTY_HINT_RANGE, "0,16,0.01,or_greater"), "set_energy", "get_energy");
-	ADD_PROPERTY(PropertyInfo(Variant::INT, "blend_mode", PROPERTY_HINT_ENUM, "Add,Subtract,Mix"), "set_blend_mode", "get_blend_mode");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "blend_mode", PROPERTY_HINT_ENUM, "Add,Subtract,Mix,AddMaximum,SubstractMaximum"), "set_blend_mode", "get_blend_mode");
```
AThousandShips, line 294:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::INT, "blend_mode", PROPERTY_HINT_ENUM, "Add,Subtract,Mix,AddMaximum,SubtractMaximum"), "set_blend_mode", "get_blend_mode");
```

bgie: Oops... thank you :-)

doc/classes/Light2D.xml
```diff
@@ -91,5 +91,15 @@
 		<constant name="BLEND_MODE_MIX" value="2" enum="BlendMode">
 			Mix the value of pixels corresponding to the Light2D to the values of pixels under it by linear interpolation.
 		</constant>
+		<constant name="BLEND_MODE_ADD_MAX" value="3" enum="BlendMode">
+			Adds the value of pixels corresponding to the Light2D to the values of pixels under it, limited to the maximum amount of all lights using the same blend mode.
+			This creates non-stacking regular lights.
+			All additive and substractive lights with a maximum are calculated as one group, and applied after the lights without a maximum.
+		</constant>
+		<constant name="BLEND_MODE_SUB_MAX" value="4" enum="BlendMode">
+			Substracts the value of pixels corresponding to the Light2D to the values of pixels under it, limited to the maximum amount of all lights using the same blend mode.
```
AThousandShips, line 100:
```suggestion
			Subtracts the value of pixels corresponding to the Light2D from the values of pixels under it, limited to the maximum amount of all lights using the same blend mode.
```

doc/classes/Light2D.xml
```diff
@@ -91,5 +91,15 @@
...
+			This creates non-stacking negative lights.
+			All additive and substractive lights with a maximum are calculated as one group, and applied after the lights without a maximum.
```
AThousandShips, line 102:
```suggestion
			All additive and subtractive lights with a maximum are calculated as one group, and applied after the lights without a maximum.
```

doc/classes/RenderingServer.xml
```diff
@@ -5206,6 +5206,12 @@
 		<constant name="CANVAS_LIGHT_BLEND_MODE_MIX" value="2" enum="CanvasLightBlendMode">
 			The light adds color depending on transparency.
 		</constant>
+		<constant name="CANVAS_LIGHT_BLEND_MODE_ADD_MAX" value="3" enum="CanvasLightBlendMode">
+			Adds light color to the canvas, limited to the maximum amount set by all lights using the same blend mode.
+		</constant>
+		<constant name="CANVAS_LIGHT_BLEND_MODE_SUB_MAX" value="4" enum="CanvasLightBlendMode">
+			Substracts light color from the canvas, limited to the maximum amount set by all lights using the same blend mode.
```
AThousandShips, line 5213:
```suggestion
			Subtracts light color from the canvas, limited to the maximum amount set by all lights using the same blend mode.
```

scene/2d/light_2d.cpp
```diff
@@ -291,7 +291,7 @@ void Light2D::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "editor_only"), "set_editor_only", "is_editor_only");
 	ADD_PROPERTY(PropertyInfo(Variant::COLOR, "color"), "set_color", "get_color");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "energy", PROPERTY_HINT_RANGE, "0,16,0.01,or_greater"), "set_energy", "get_energy");
-	ADD_PROPERTY(PropertyInfo(Variant::INT, "blend_mode", PROPERTY_HINT_ENUM, "Add,Subtract,Mix"), "set_blend_mode", "get_blend_mode");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "blend_mode", PROPERTY_HINT_ENUM, "Add,Subtract,Mix,AddMaximum,SubtractMaximum"), "set_blend_mode", "get_blend_mode");
```
AThousandShips, line 294:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::INT, "blend_mode", PROPERTY_HINT_ENUM, "Add,Subtract,Mix,Add Maximum,Subtract Maximum"), "set_blend_mode", "get_blend_mode");
```
Should be separate words

AThousandShips:
You also need to fix the typo in your commit at some point, but no rush with that as long as it's done before merge 🙂

doc/classes/Light2D.xml
```diff
@@ -91,5 +91,15 @@
 		<constant name="BLEND_MODE_MIX" value="2" enum="BlendMode">
 			Mix the value of pixels corresponding to the Light2D to the values of pixels under it by linear interpolation.
 		</constant>
+		<constant name="BLEND_MODE_ADD_MAX" value="3" enum="BlendMode">
+			Adds the value of pixels corresponding to the Light2D to the values of pixels under it, limited to the maximum amount of all lights using the same blend mode.
+			This creates non-stacking regular lights.
+			eAll additive and subtractive lights with a maximum are calculated as one group, and applied after the lights without a maximum.
```
hsandt, line 97:
```suggestion
			eAll additive and subtractive lights with a maximum are calculated as one group, and applied after the lights without a maximum.
```
```suggestion
			All additive and subtractive lights with a maximum are calculated as one group, and applied after the lights without a maximum.
```

Calinou (changes requested):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/29b3d9e9e538f0aa8effc8ad8bf19a2915292a89), it works as expected with the Forward+ and Mobile rendering methods.

However, with the Compatibility rendering method, I get a shader compilation error that doesn't occur in `master`:

```
ERROR: CanvasShaderGLES3: Fragment shader compilation failed:
0(409) : error C1118: "break" not in loop
0(414) : error C1118: "break" not in loop
```

**Testing project:** [test_light_2d_blend_modes.zip](https://github.com/godotengine/godot/files/14809111/test_light_2d_blend_modes.zip)

Add | Add Maximum
-|-
add | add_maximum

KoBeWi:
Tested with 2 simple circle lights and there is a weird dark line in between when they are getting merged..

https://github.com/godotengine/godot/assets/2223172/9f352691-1698-4dd4-b51a-73341a5738bf

Notice how the line moves towards the light, making the sprite darker.

daninaga:
@Calinou I believe in your printscreens we can actually see it does not work, looking at the Add_Maximum we can clearly see that the rightmost, topmost light does not correctly blend with the one directly to the left of it:
image

Also same border artifacts described by @KoBeWi can be seen:

image

Furthermore I may be wrong but the case statements implemented in the blend compute should be inside a "switch" statement (canvas.glsl line 505):

```
} else if (blend_mode == LIGHT_FLAGS_BLEND_MODE_MIX) {
		color.rgb = mix(color.rgb, light_color.rgb, light_color.a);
		case LIGHT_FLAGS_BLEND_MODE_ADD_MAX: {
			vec3 delta_light = light_color.rgb * light_color.a;
			clamp_max = max(clamp_max, delta_light);
			accumulated_light_to_clamp += delta_light;
		} break;
		case LIGHT_FLAGS_BLEND_MODE_SUB_MAX: {
			vec3 delta_light = light_color.rgb * -light_color.a;
			clamp_min = min(clamp_min, delta_light);
			accumulated_light_to_clamp += delta_light;
		} break;
```

bgie:
With regards to the screenshot about the max function supposedly not working, I quickly drew a diagram to explain what one can expect from this feature. 

lights_sum_vs_max

TL;DR: for max light blending: do not use soft lights with a gradient but lights with hard edges.

Graphs show a 1-dimensional slice of a two 2d lights, with the vertical axis being light intensity.
The top-right graph explains why there is a noticeable line in the screenshots posted by other users: the gradient makes an abrupt change. In the additive case there is a plateau and 2 softer transitions (less steep angle).
For lights with a hard edge we get the opposite: max blending gives us a nice merged plateau but additive blending gives us over-exposure (peak of 2x intensity).

Max blending, imho, is not physically realistic. It does not play well with more realistic smooth lights. It is more suited for a fog-of-war effect though.

KoBeWi:
Eh, I looked closely at my video and it seems to be optical illusion. I guess I just had wrong expectations and it doesn't solve my [old proposal](https://github.com/godotengine/godot-proposals/issues/891) (I've been using a workaround since forever though, so I don't care).

It's fine if the new option requires hard edges, but this has to be better documented then. Other users might also have wrong expectations.

daninaga:
i dont know. I would expect this feature to be more like either we have a maximum light intensity or a minimum but not both at same time.
I think most users are looking for a combination of soft-falloff sum for the minimum and hard-falloff max for the maximum.

I need this exactly for fog-of-war. I have a custom shader atm but this means all lights need to be in the same object and are not parallel objects i.e. i would not be able to dynamically spawn lights:
image

DigitalTableTops:
I am eagerly awaiting this to be added :) It would be really useful for the fog of war I am working on. I do not know enough to be able to help contribute, sorry.

jeannen:
Hey! Will this be added?

Calinou:
> Hey! Will this be added?

There are merge conflicts, so this PR needs to be rebased first before it can be reviewed.

KeyboardDanni:
Personally, I would like a way to customize the light blending (right now it seems like there isn't a good way to do this). In my case, the world is in daylight so nothing is obscured, but I want to add additional normal-mapped lighting to certain elements in the scene. When a single light gets too close, it causes the colors to get oversaturated. It would be nice if I had a way to, for example, palettize or tonemap the lighting result.

chryan:
What is the process for getting a change like this approved and integrated into the engine? It's been 2 years since this change was initially proposed.

If someone fixes the merge issues, which looks like someone did, what then? How does that get approved?

PR #86617, opened 2023-12-29
Warn about case mismatch when enabling addons

KoBeWi:
`set_plugin_enabled()` expects a folder name and uses is as a path for enabling a plugin. This can cause problems in case-insensitive filesystems if you provide folder name with case mismatch. The plugin will be enabled, but you won't be able to disable it other than by removing it manually from `project.godot`.

This PR checks whether the folder you provide for `set_plugin_enabled()` has proper casing and if not, it warns and automatically fixes the path. I had to add a new `is_case_mismatch()` function to FileAccess, which checks whether path casing matches filesystem and can return a fixed path.

### DIFF
--- a/drivers/windows/file_access_windows.cpp
+++ b/drivers/windows/file_access_windows.cpp
@@ -122,17 +122,9 @@ Error FileAccessWindows::open_internal(const String &p_path, int p_mode_flags) {
 	// a file using the wrong case (which *works* on Windows, but won't on other
 	// platforms).
 	if (p_mode_flags == READ) {
-		WIN32_FIND_DATAW d;
-		HANDLE fnd = FindFirstFileW((LPCWSTR)(path.utf16().get_data()), &d);
-		if (fnd != INVALID_HANDLE_VALUE) {
-			String fname = String::utf16((const char16_t *)(d.cFileName));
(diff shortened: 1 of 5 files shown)


drivers/windows/file_access_windows.cpp
```diff
@@ -122,17 +122,9 @@ Error FileAccessWindows::open_internal(const String &p_path, int p_mode_flags) {
...
-			if (!fname.is_empty()) {
-				String base_file = path.get_file();
-				if (base_file != fname && base_file.findn(fname) == 0) {
-					WARN_PRINT("Case mismatch opening requested file '" + base_file + "', stored as '" + fname + "' in the filesystem. This file will not open when exported to other case-sensitive platforms.");
-				}
-			}
-			FindClose(fnd);
+		String physical_path;
+		if (is_case_mismatch(false, &physical_path)) {
+			WARN_PRINT("Case mismatch opening requested file '" + path.get_file() + "', stored as '" + physical_path.get_file() + "' in the filesystem. This file will not open when exported to other case-sensitive platforms.");
```
KoBeWi, line 127: tbh I haven't tested if this warning works correctly after the changes.

PR #82157, opened 2023-09-22
Change to InputEventMIDI event to include connection_index property

ramdor:
Added connection_index to the InputEventMIDI so that each MIDI message can be associated with a specific opened MIDI device.

- *Production edit: This closes https://github.com/godotengine/godot-proposals/issues/7733.*

PR #86620, opened 2023-12-29
Add MIDI controller device index to `InputEventMIDI.device` property.

### LINKED ISSUE #7733: Provide MIDI device property in InputEventMIDI class

I am working on a digital art project where I am using Godot as visualization engine which is influenced by audio analysis and incoming MIDI events.

Some external MIDI Hardware devices cannot change their MIDI channel and cannot be distinguished from each other without device information for each incoming MIDI event.

A string attribute "input" or "device" containing the device name as listed by OS.get_connected_midi_inputs() or an int attribute pointing to the corresponding array element of OS.get_connected_midi_inputs() would solve the problem.

if input_event is InputEventMIDI and input_event.device == "my_awesome_midi_controller":
    do_something_useful()

I haven't found any workaroung except integrating an external MIDI library like portmidi or rtmidi as GDExtension

All necessary information should already be available and exposing the device name should be quiet simple (i would guess).

fbcosentino:
It is possible to query the OS for the connected MIDI controllers, but the event messages' device field was not being used. This implements controller index being sent in InputEventMIDI messages in the device property, matching the index from OS.get_connected_midi_inputs().

Based on the work done by @ramdor in [this archived PR](https://github.com/godotengine/godot/pull/82157)

Closes godotengine/godot-proposals#7733

### DIFF
--- a/core/os/midi_driver.cpp
+++ b/core/os/midi_driver.cpp
@@ -42,9 +42,10 @@ void MIDIDriver::set_singleton() {
 	singleton = this;
 }
 
-void MIDIDriver::receive_input_packet(uint64_t timestamp, uint8_t *data, uint32_t length) {
+void MIDIDriver::receive_input_packet(int device_index, uint64_t timestamp, uint8_t *data, uint32_t length) {
 	Ref<InputEventMIDI> event;
 	event.instantiate();
+	event->set_device(device_index);
 	uint32_t param_position = 1;
 
 	if (length >= 1) {
--- a/core/os/midi_driver.h
+++ b/core/os/midi_driver.h
@@ -51,7 +51,7 @@ public:
 
 	virtual PackedStringArray get_connected_inputs();
 
-	static void receive_input_packet(uint64_t timestamp, uint8_t *data, uint32_t length);
+	static void receive_input_packet(int device_index, uint64_t timestamp, uint8_t *data, uint32_t length);
 
 	MIDIDriver();
 	virtual ~MIDIDriver() {}
--- a/drivers/alsamidi/midi_driver_alsamidi.cpp
+++ b/drivers/alsamidi/midi_driver_alsamidi.cpp
@@ -82,13 +82,13 @@ size_t MIDIDriverALSAMidi::msg_expected_data(uint8_t status_byte) {
 }
 
 void MIDIDriverALSAMidi::InputConnection::parse_byte(uint8_t byte, MIDIDriverALSAMidi &driver,
-		uint64_t timestamp) {
+		uint64_t timestamp, int device_index) {
 	switch (msg_category(byte)) {
 		case MessageCategory::RealTime:
 			// Real-Time messages are single byte messages that can
 			// occur at any point.
 			// We pass them straight through.
-			driver.receive_input_packet(timestamp, &byte, 1);
+			driver.receive_input_packet(device_index, timestamp, &byte, 1);
 			break;
 
 		case MessageCategory::Data:
@@ -100,7 +100,7 @@ void MIDIDriverALSAMidi::InputConnection::parse_byte(uint8_t byte, MIDIDriverALS
 
 				// Forward a complete message and reset relevant state.
 				if (received_data == expected_data) {
-					driver.receive_input_packet(timestamp, buffer, received_data + 1);
+					driver.receive_input_packet(device_index, timestamp, buffer, received_data + 1);
 					received_data = 0;
 
 					if (msg_category(buffer[0]) != MessageCategory::Voice) {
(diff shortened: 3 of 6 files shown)


AThousandShips:
The co-author didn't apply, you need to make the part in the arrow brackets their email, check out their branch and use `git log` to get a valid author to use 🙂

fbcosentino:
> The co-author didn't apply, you need to make the part in the arrow brackets their email, check out their branch and use `git log` to get a valid author to use 🙂

Done. Seems to work now.

AThousandShips:
Your commit seems not to be linked to your GitHub account. See: [Why are my commits linked to the wrong user?](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user) for more info.

fbcosentino:
> Your commit seems not to be linked to your GitHub account. See: [Why are my commits linked to the wrong user?](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user) for more info.

Fixed. (Apparently I'm as good with github as Sheldon is good with the flute. Eventually I'll get better.)

PeterMarques:
This needs to get pushed up.

core/input/input_event.cpp
```diff
@@ -1725,22 +1725,22 @@ String InputEventMIDI::to_string() {
...
-			ret = vformat("Channel Pressure: channel=%d, pressure=%d", channel, pressure);
+			ret = vformat("Channel Pressure: channel=%d, pressure=%d, device=%d", channel, pressure, get_device());
 			break;
 		case MIDIMessage::CONTROL_CHANGE:
-			ret = vformat("Control Change: channel=%d, controller_number=%d, controller_value=%d", channel, controller_number, controller_value);
+			ret = vformat("Control Change: channel=%d, controller_number=%d, controller_value=%d, device=%d", channel, controller_number, controller_value, get_device());
 			break;
 		default:
-			ret = vformat("channel=%d, message=%d, pitch=%d, velocity=%d, pressure=%d, controller_number=%d, controller_value=%d, instrument=%d", channel, message, pitch, velocity, pressure, controller_number, controller_value, instrument);
+			ret = vformat("channel=%d, message=%d, pitch=%d, velocity=%d, pressure=%d, controller_number=%d, controller_value=%d, instrument=%d, device=%d", channel, message, pitch, velocity, pressure, controller_number, controller_value, instrument, get_device());
```
Sauermann, line 1743: None of the other `InputEvent::to_string` functions display their device id. To be consistent with the other input events, I would prefer, if it wouldn't be included.
What is your motivation to include the device id in `InputEventMIDI::to_string`?

Sauermann (commented):
I have no experience with drivers, but the changes to core are looking correct.

core/input/input_event.cpp
PeterMarques: Its a Idiosyncrasy of being a music instrument input, i guess.
Not the author, but it makes sense to have the instrument along with the notes.

But... I agree with having consistency on the implementation, and to simple use the ".device" attribute to infer the origin of message, and by match one can use the to_string.

fbcosentino: since the vast majority of input events in games come from keyboard/mouse, and (at least before I finish the other branch I'm working on - which does the same as this one, but for mouse and keyboard) they get merged by the OS and therefore not being relevant, for now this leaves only gamepads with relevant `device`. So my understanding is device wasn't included in the string because it wasn't relevant. 

If youse feel this will clutter the output I can obviously remove it.

fbcosentino: Removed!

Sauermann (approved):
Changes to core are looking correct.
Can't say anything about the change to drivers.
I can't test the PR, because I don't have MIDI-devices.

ramdor:
hi @fbcosentino . I know this has been merged now, based on the changes I was dabbling with a long time ago. However, there are some issues with the coremidi implementation. I built a quick mac test and it seems like the device_index (InputEvent.device) is garbage data.

It looks to be caused by line 68 of midi_drive_coremidi.cpp where the pointer to 'i' is being used. This should not be a pointer, but the actual value of 'i'.  Otherwise multiple read callbacks will use the same pointer. Also 'i' will go out of scope and result in junk.

We need to provide the actual value of i to MIDIPortConnectSource so that when the callback function read is called when midi events come in, src_conn_ref_con is the actual value of i that was provided when MIDIPortConnectSource was called and can be used as device_index.

I hope this makes sense.  Cheers.

fbcosentino:
> some test code that does correct type casting :
> 
> ```
> #include <iostream>
> //#include <cstdint>
> 
> void testFunction(void *src_conn_ref_con)
> {
>     intptr_t device_index = reinterpret_cast<intptr_t>(src_conn_ref_con);
>     std::cout << device_index << std::endl;
> }
> 
> int main()
> {
>     int sources = 16;
>     for (int i = 0; i < sources; i++) {
>         testFunction(reinterpret_cast<void*>(static_cast<intptr_t>(i)));
>     }
> }
> ```

Hi @ramdor !

I actually don't have a mac (nor access to one). Would you be comfortable to make these changes yourself?

PR #86627, opened 2023-12-30
Add static body creation equivalent for all collision shape sibling creations

### LINKED ISSUE #86626: Can't create static body for something else then trimesh collisions
### Tested versions

- Reproducible in : 3.5.3 4.x ( 4.1, 4.3.dev_build[13a0d6e9b253654f5cc2a44f3d0b3cae10440443], assuming all other version)

### System information

### Issue description

Tere is no available possibilities for creating a static body for something else then trimesh collisions

### Steps to reproduce

Select a MeshInstance3D 
click on the Mesh drop down button on the top of the 3D viewport 
see there is no available options for creating a static body for something else then trimesh collisions

N/A

ajreckof:
Fix #86626

Just adds an option in the popup menu and handle it the same way it was handled for trimesh collision shapes

### DIFF
--- a/editor/plugins/mesh_instance_3d_editor_plugin.cpp
+++ b/editor/plugins/mesh_instance_3d_editor_plugin.cpp
@@ -59,6 +59,148 @@ void MeshInstance3DEditor::edit(MeshInstance3D *p_mesh) {
 	node = p_mesh;
 }
 
+Vector<Ref<Shape3D>> MeshInstance3DEditor::create_shape_from_mesh(Ref<Mesh> p_mesh, int p_option, bool p_verbose) {
+	Vector<Ref<Shape3D>> shapes;
+	switch (p_option) {
+		case SHAPE_TYPE_TRIMESH: {
+			shapes.push_back(p_mesh->create_trimesh_shape());
+
+			if (p_verbose && shapes.is_empty()) {
+				err_dialog->set_text(TTR("Couldn't create a Trimesh collision shape."));
+				err_dialog->popup_centered();
+			}
+		} break;
+
+		case SHAPE_TYPE_SINGLE_CONVEX: {
+			shapes.push_back(p_mesh->create_convex_shape(true, false));
+
+			if (p_verbose && shapes.is_empty()) {
+				err_dialog->set_text(TTR("Couldn't create a single collision shape."));
+				err_dialog->popup_centered();
+			}
+		} break;
+
+		case SHAPE_TYPE_SIMPLIFIED_CONVEX: {
+			shapes.push_back(p_mesh->create_convex_shape(true, true));
+
+			if (p_verbose && shapes.is_empty()) {
+				err_dialog->set_text(TTR("Couldn't create a simplified collision shape."));
+				err_dialog->popup_centered();
+			}
+		} break;
+
+		case SHAPE_TYPE_MULTIPLE_CONVEX: {
+			Ref<MeshConvexDecompositionSettings> settings;
+			settings.instantiate();
+			settings->set_max_convex_hulls(32);
+			settings->set_max_concavity(0.001);
+
+			shapes = p_mesh->convex_decompose(settings);
+
+			if (p_verbose && shapes.is_empty()) {
+				err_dialog->set_text(TTR("Couldn't create any collision shapes."));
+				err_dialog->popup_centered();
+			}
+		} break;
+
+		default:
+			break;
+	}
+	return shapes;
+}
+
+void MeshInstance3DEditor::_create_collision_shape() {
+	int placement_option = shape_placement->get_selected();
+	int shape_type_option = shape_type->get_selected();
+
+	EditorSelection *editor_selection = EditorNode::get_singleton()->get_editor_selection();
+	EditorUndoRedoManager *ur = EditorUndoRedoManager::get_singleton();
+
+	switch (shape_type_option) {
+		case SHAPE_TYPE_TRIMESH: {
+			ur->create_action(TTR(placement_option == SHAPE_PLACEMENT_SIBLING ? "Create Trimesh Collision Shape Sibling" : "Create Trimesh Static Body"));
+		} break;
+		case SHAPE_TYPE_SINGLE_CONVEX: {
+			ur->create_action(TTR(placement_option == SHAPE_PLACEMENT_SIBLING ? "Create Single Convex Collision Shape Sibling" : "Create Single Convex Static Body"));
+		} break;
+		case SHAPE_TYPE_SIMPLIFIED_CONVEX: {
+			ur->create_action(TTR(placement_option == SHAPE_PLACEMENT_SIBLING ? "Create Simplified Convex Collision Shape Sibling" : "Create Simplified Convex Static Body"));
+		} break;
+		case SHAPE_TYPE_MULTIPLE_CONVEX: {
+			ur->create_action(TTR(placement_option == SHAPE_PLACEMENT_SIBLING ? "Create Multiple Convex Collision Shape Siblings" : "Create Multiple Convex Static Body"));
+		} break;
+		default:
+			break;
+	}
+
+	List<Node *> selection = editor_selection->get_selected_node_list();
+
+	bool verbose = false;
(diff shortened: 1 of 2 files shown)


editor/plugins/mesh_instance_3d_editor_plugin.h
```diff
@@ -69,6 +72,7 @@ class MeshInstance3DEditor : public Control {
 	Control *debug_uv = nullptr;
 	Vector<Vector2> uv_lines;
 
+	Vector<Ref<Shape3D>> create_shape_from_mesh(Ref<Mesh> mesh, int p_option, bool verbose);
```
AThousandShips, line 75:
```suggestion
	Vector<Ref<Shape3D>> create_shape_from_mesh(Ref<Mesh> p_mesh, int p_option, bool p_verbose);
```
Prefix arguments with `p_`

editor/plugins/mesh_instance_3d_editor_plugin.cpp
```diff
@@ -58,6 +58,59 @@ void MeshInstance3DEditor::edit(MeshInstance3D *p_mesh) {
...
 }
 
+Vector<Ref<Shape3D>> MeshInstance3DEditor::create_shape_from_mesh(Ref<Mesh> mesh, int p_option, bool verbose) {
+	Vector<Ref<Shape3D>> shapes = Vector<Ref<Shape3D>>();
+	switch (p_option) {
+		case MENU_OPTION_CREATE_TRIMESH_STATIC_BODY:
+		case MENU_OPTION_CREATE_TRIMESH_COLLISION_SHAPE: {
+			shapes.push_back(mesh->create_trimesh_shape());
+
+			if (verbose && !shapes.size()) {
```
AThousandShips, line 68:
```suggestion
			if (verbose && shapes.is_empty()) {
```
Same for all

editor/plugins/mesh_instance_3d_editor_plugin.cpp
```diff
@@ -58,6 +58,59 @@ void MeshInstance3DEditor::edit(MeshInstance3D *p_mesh) {
...
+				err_dialog->set_text(TTR("Couldn't create a Trimesh collision shapes."));
```
AThousandShips, line 69:
```suggestion
				err_dialog->set_text(TTR("Couldn't create a Trimesh collision shape."));
```

editor/plugins/mesh_instance_3d_editor_plugin.cpp
```diff
@@ -58,6 +58,59 @@ void MeshInstance3DEditor::edit(MeshInstance3D *p_mesh) {
...
+				err_dialog->popup_centered();
+			}
+		} break;
+
+		case MENU_OPTION_CREATE_SINGLE_CONVEX_STATIC_BODY:
+		case MENU_OPTION_CREATE_SINGLE_CONVEX_COLLISION_SHAPE: {
+			shapes.push_back(mesh->create_convex_shape(true, false));
+
+			if (verbose && !shapes.size()) {
+				err_dialog->set_text(TTR("Couldn't create a single collision shapes."));
```
AThousandShips, line 79:
```suggestion
				err_dialog->set_text(TTR("Couldn't create a single collision shape."));
```

editor/plugins/mesh_instance_3d_editor_plugin.cpp
```diff
@@ -58,6 +58,59 @@ void MeshInstance3DEditor::edit(MeshInstance3D *p_mesh) {
...
+				err_dialog->popup_centered();
+			}
+		} break;
+
+		case MENU_OPTION_CREATE_SIMPLIFIED_CONVEX_STATIC_BODY:
+		case MENU_OPTION_CREATE_SIMPLIFIED_CONVEX_COLLISION_SHAPE: {
+			shapes.push_back(mesh->create_convex_shape(true, true));
+
+			if (verbose && !shapes.size()) {
+				err_dialog->set_text(TTR("Couldn't create a simplified collision shapes."));
```
AThousandShips, line 89:
```suggestion
				err_dialog->set_text(TTR("Couldn't create a simplified collision shape."));
```

editor/plugins/mesh_instance_3d_editor_plugin.cpp
```diff
@@ -67,41 +120,38 @@ void MeshInstance3DEditor::_menu_option(int p_option) {
...
+					ur->create_action(TTR("Create Trimesh Static Body"));
+				} break;
+				case MENU_OPTION_CREATE_SINGLE_CONVEX_STATIC_BODY: {
+					ur->create_action(TTR("Create Single Convex Static Body"));
+				} break;
+				case MENU_OPTION_CREATE_SIMPLIFIED_CONVEX_STATIC_BODY: {
+					ur->create_action(TTR("Create Simplified Convex Static Body"));
+				} break;
+				case MENU_OPTION_CREATE_MULTIPLE_CONVEX_CSTATIC_BODY: {
+					ur->create_action(TTR("Create Multiple Convex SStatic Body"));
```
AThousandShips, line 141:
```suggestion
					ur->create_action(TTR("Create Multiple Convex Static Body"));
```

editor/plugins/mesh_instance_3d_editor_plugin.h
```diff
@@ -43,7 +43,10 @@ class MeshInstance3DEditor : public Control {
 	GDCLASS(MeshInstance3DEditor, Control);
 
 	enum Menu {
-		MENU_OPTION_CREATE_STATIC_TRIMESH_BODY,
+		MENU_OPTION_CREATE_TRIMESH_STATIC_BODY,
+		MENU_OPTION_CREATE_SINGLE_CONVEX_STATIC_BODY,
+		MENU_OPTION_CREATE_SIMPLIFIED_CONVEX_STATIC_BODY,
+		MENU_OPTION_CREATE_MULTIPLE_CONVEX_CSTATIC_BODY,
```
AThousandShips, line 49:
```suggestion
		MENU_OPTION_CREATE_MULTIPLE_CONVEX_STATIC_BODY,
```
Assuming this is a typo

smix8:
The duplicated menu option names need to be fixed.

menu

Imo instead of making the menu even more convoluted than it already is it should be moved to a submenu where users can change the currently hardcoded settings.

ajreckof:
I reworked it so that it is now in a submenu. On the side while reworking I ended up adding the posibility to add colision shape as sibling on a multi-selection too as it was previously only for static body child.

smix8:
I tested with a the build artifact from this pr and a simple 2x2x2x BoxMesh.

The "Static Body Child" options all work as expected and create StaticBody3D child node with a CollisionShape3D child node that has a ConvexPolygonShape3D resource.

Now what does not seem to work as expected are the "Sibling" options.

I would expect them to do the very same except for leaving out the StaticBody3D node and place the CollisionShape3D node as a sibling with the same ConvexPolygonShape3D resource as with the other options.

The node creation and placement works but not the Shape3D resource creation. Not sure why but the sibling options all created ConcavePolygonShape3D for everything no matter what convex menu option is picked. I would expect them to create the same shape as the other convex options that work.

editor/plugins/mesh_instance_3d_editor_plugin.cpp
```diff
@@ -58,200 +58,159 @@ void MeshInstance3DEditor::edit(MeshInstance3D *p_mesh) {
...
+	}
+	return shapes;
+}
 
-			CollisionShape3D *cshape = memnew(CollisionShape3D);
-			cshape->set_shape(shape);
-			cshape->set_transform(node->get_transform());
+void MeshInstance3DEditor::_create_collision_shape() {
+	int placement_option = shape_placement->get_selected();
+	int shape_type_option = shape_placement->get_selected();
```
ajreckof, line 113: this is the culprit code I used in both case `shape_placement` while the second one was supposed to be `shape_type` must have been a bad copy paste. :(

smix8 (approved):
Seem to work as expected now. I did not encounter any more issues.

AThousandShips (approved):

akien-mga:
Thanks!

PR #86635, opened 2023-12-30
Improve error message when trying to instantiate a script with `_init(...)`

### LINKED ISSUE #8488: Improve error message when Godot is unable to construct GDScriptInstance

I am building an addon to manage RPG data.

For optimal clean code, we addon developers like to use the `_init()` function as a custom constructor for objects, as it forces the user to always initialize an object:
```gdscript
class_name MyNode extends Node

var _value:int

func _init(arg:int):
   _value = arg
```
This works fine, especially for custom data structures that are not utilizing the scene tree. However, when adding a node to a scene tree and attaching the script above, Godot will spit out an error that is very confusing to people who never encountered it before:
```
_create_instance: Error constructing a GDScriptInstance.
```
While the above error is logical and makes sense (as Godot requires an `_init()` constructor to call `.new()`), it is irritating to users.

The error can potentially occur in multiple scenarios:

1. `_init()` function with arguments defined
2. script has a syntax error

This proposal aims to improve the error messages around these two scenarios.

When there are no script errors but a custom `_init(...)` function was defined, show an error like this:
```
Unable to create node /root/path/to/node/in/scene/tree: script does not define _init().
```
and in case of script errors:
```
Unable to create node /root/path/to/node/in/scene/tree: script has errors.
```

No workaround sadly, apart from spending potentially hours trying to understand the issue (in larger code bases)

As the original vague error is in core.

HolonProduction:
Fixes godotengine/godot-proposals#8488

### DIFF
--- a/modules/gdscript/gdscript.cpp
+++ b/modules/gdscript/gdscript.cpp
@@ -185,7 +185,12 @@ GDScriptInstance *GDScript::_create_instance(const Variant **p_args, int p_argco
 				MutexLock lock(GDScriptLanguage::singleton->mutex);
 				instances.erase(p_owner);
 			}
-			ERR_FAIL_V_MSG(nullptr, "Error constructing a GDScriptInstance.");
+			if (r_error.error == Callable::CallError::CALL_ERROR_TOO_FEW_ARGUMENTS) {
+				ERR_FAIL_V_MSG(nullptr, vformat("Cannot init GDScriptInstance of \"%s\", because _init(...) requires parameters. Consider adding default values.", path));
+
+			} else {
+				ERR_FAIL_V_MSG(nullptr, "Error constructing a GDScriptInstance.");
+			}
 		}
 	}
 	//@TODO make thread safe


AThousandShips:
See also:
* https://github.com/godotengine/godot/pull/76632

Mickeon:
Is it possible, and could it be nice to explicitly state _how many_ parameters are required?

HolonProduction:
Is possible. But doesn't help in the context of the proposal since no parameters can be passed. But yeah the error contains the expected amount of args so adding it to the message is possible.

AThousandShips:
I'd say that #76632 is more comprehensive as it also adds error information (ignore the accidental inclusion of unrelated code by the OP of that PR, hopefully that'll get fixer or a new branch made)

AThousandShips:
Also this solution prints this message when creating an instance with too few arguments yet not none, as it is called in other situations as well, not just resource initialization, so I think the other PR is better for this

HolonProduction:
Adding the call error is definitely right. (I could probably include the changes into this PR to resolve the rebase mess up). I am just curious whether it is helpful in the context of https://github.com/godotengine/godot-proposals/issues/8488. Or if we need to extend the error message to be helpful in that context. I'll have a look at the specific message.

HolonProduction:
Closing in favor of #76632

PR #79206, opened 2023-07-08
Add a warning about C# differences to the class reference

raulsntos:
Adds a warning to the class reference documentation for classes that have _notable_[^1] differences between GDScript and C#.

- Closes https://github.com/godotengine/godot/pull/70852 as superseded.
    - The warning added by this PR + https://github.com/godotengine/godot-docs/pull/6679 seem sufficient to supersed.
- Closes https://github.com/godotengine/godot-docs/issues/5806.

[^1]: By _notable_ I mean important enough to be mentioned explicitly in the [_C# API differences to GDScript_](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_differences.html) documentation page.

PR #86639, opened 2023-12-30
Fix several reported issues in String's documentation

### LINKED ISSUE #8043: Meaningless sentence in `String.capitalize`'s description
**Your Godot version:** 4.1.1

**Issue description:**

image

**URL to the documentation page:**
https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-capitalize

### LINKED ISSUE #8044: Display "uppercase" and "lowercase" in a consistent manner
**Your Godot version:**
4.1.1

**Issue description:**

Take a look at the descriptions:

image

while it is probably clear to everyone what uppercase and lowercase mean (and it might even not be the case, for example if someone is a begginer and not a native english speaker), I think it would be nice to display them explicitely, just like how `PascalCase` and `snake_case` are shown, in this particular context.
That is, change:
   - lowercase → `lowercase`
   - uppercase → `UPPERCASE`

in the method description of `to_lower` and `to_upper`.

**URL to the documentation page (if already existing):**
https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-to-lower

### LINKED ISSUE #86734: Inconsistency: `"1a".to_snake_case() == "1a"`, but `"1aa".to_snake_case() == "1_aa"`.
### Tested versions

Reproducible in:
`3.5.3.stable` using `capitalize`
`4.2.1.stable` using `capitalize` and `to_snake_case`.

### System information

Linux on kernel 6.6.7-arch1-1

### Issue description

```gdscript
extends Node

func _ready():
	print("1a".capitalize())
	print("1aa".capitalize())
	print("1a".to_snake_case())
	print("1aa".to_snake_case())
```

Output:

```
1a
1 Aa
1a
1_aa
```

Expected:

```
1 A
1 Aa
1_a
1_aa
```

### Steps to reproduce

https://gd.tumeo.space/?KYDwLsB2AmDOAEA5A9tYAodAzArpAxvAPoBOwAhtAJ4AUAlAFzoCQADiQJaRg0BEAjOV4A6fOVYcw5ADYcAXsHp0W7LjwHkho8ZJnzFdZW07c+gkWGRFYkcgGtgRMbANHVpjVsvXbDp+RclIA

[to_snake_case_4.2.1.zip](https://github.com/godotengine/godot/files/13815827/to_snake_case_4.2.1.zip)

Mickeon:
Closes https://github.com/godotengine/godot-docs/issues/8043
Closes https://github.com/godotengine/godot-docs/issues/8044
Closes https://github.com/godotengine/godot/issues/86734
Mitigates https://github.com/godotengine/godot-docs/issues/8190

This PR updates the String's class reference slightly.
It formats uppercase and lowercase in their respective String methods and removes a seemingly useless note in `String.capitalize()` and incorporates it with the example code, instead.
~~Finally, for StringName, it adds a note exclusive to C#, because all of the listed methods are not available there. This note is almost word-for-word taken from the bug report https://github.com/godotengine/godot-docs/issues/8190 (@raulsntos)~~ 

Oh great heavens, news flash. Let's also add some examples for https://github.com/godotengine/godot/issues/86734, why not.

Could be easily cherrypicked for prior Godot 4.x versions

### DIFF
--- a/doc/classes/String.xml
+++ b/doc/classes/String.xml
@@ -96,13 +96,14 @@
 				[gdscript]
 				"move_local_x".capitalize()   # Returns "Move Local X"
 				"sceneFile_path".capitalize() # Returns "Scene File Path"
+				"2D, FPS, PNG".capitalize()   # Returns "2d, Fps, Png"
 				[/gdscript]
 				[csharp]
 				"move_local_x".Capitalize();   // Returns "Move Local X"
 				"sceneFile_path".Capitalize(); // Returns "Scene File Path"
+				"2D, FPS, PNG".Capitalize();   // Returns "2d, Fps, Png"
 				[/csharp]
 				[/codeblocks]
-				[b]Note:[/b] This method not the same as the default appearance of properties in the Inspector dock, as it does not capitalize acronyms ([code]"2D"[/code], [code]"FPS"[/code], [code]"PNG"[/code], etc.) as you may expect.
 			</description>
 		</method>
 		<method name="casecmp_to" qualifiers="const">
@@ -949,7 +950,7 @@
 		<method name="to_lower" qualifiers="const">
 			<return type="String" />
 			<description>
-				Returns the string converted to lowercase.
+				Returns the string converted to [code]lowercase[/code].
 			</description>
 		</method>
 		<method name="to_pascal_case" qualifiers="const">
@@ -962,12 +963,25 @@
 			<return type="String" />
 			<description>
 				Returns the string converted to [code]snake_case[/code].
+				[b]Note:[/b] Numbers followed by a [i]single[/i] letter are not separated in the conversion to keep some words (such as "2D") together.
+				[codeblocks]
+				[gdscript]
+				"Node2D".to_snake_case()               # Returns "node_2d"
+				"2nd place".to_snake_case()            # Returns "2_nd_place"
+				"Texture3DAssetFolder".to_snake_case() # Returns "texture_3d_asset_folder"
+				[/gdscript]
+				[csharp]
+				"Node2D".ToSnakeCase();               // Returns "node_2d"
+				"2nd place".ToSnakeCase();            // Returns "2_nd_place"
+				"Texture3DAssetFolder".ToSnakeCase(); // Returns "texture_3d_asset_folder"
+				[/csharp]
+				[/codeblocks]
 			</description>
 		</method>
 		<method name="to_upper" qualifiers="const">
 			<return type="String" />
 			<description>
-				Returns the string converted to uppercase.
+				Returns the string converted to [code]UPPERCASE[/code].
 			</description>
 		</method>
 		<method name="to_utf8_buffer" qualifiers="const">
--- a/doc/classes/StringName.xml
+++ b/doc/classes/StringName.xml
@@ -8,6 +8,7 @@
 		You will usually just pass a [String] to methods expecting a [StringName] and it will be automatically converted, but you may occasionally want to construct a [StringName] ahead of time with the [StringName] constructor or, in GDScript, the literal syntax [code]&amp;"example"[/code].
 		See also [NodePath], which is a similar concept specifically designed to store pre-parsed scene tree paths.
 		All of [String]'s methods are available in this class too. They convert the [StringName] into a string, and they also return a string. This is highly inefficient and should only be used if the string is desired.
+		[b]Note:[/b] In C#, an explicit conversion to [code]System.String[/code] is required to use the methods listed on this page. Use the [code]ToString()[/code] method to cast a [StringName] to a string, and then use the equivalent methods in [code]System.String[/code] or [code]StringExtensions[/code].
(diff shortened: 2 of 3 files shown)


doc/classes/StringName.xml
```diff
@@ -8,6 +8,7 @@
 		You will usually just pass a [String] to methods expecting a [StringName] and it will be automatically converted, but you may occasionally want to construct a [StringName] ahead of time with the [StringName] constructor or, in GDScript, the literal syntax [code]&amp;"example"[/code].
 		See also [NodePath], which is a similar concept specifically designed to store pre-parsed scene tree paths.
 		All of [String]'s methods are available in this class too. They convert the [StringName] into a string, and they also return a string. This is highly inefficient and should only be used if the string is desired.
+		[b]Note:[/b] In C#, none of the methods listed on this page are available. Instead, use the provided [code]ToString()[/code] method to cast a [StringName] to a string, and then use the equivalent methods in [code]System.String[/code] or [code]StringExtensions[/code]. For convenience, a list of equivalent methods is provided [url=$DOCS_URL/tutorials/scripting/c_sharp/c_sharp_differences.html#string]here[/url].
```
raulsntos, line 11: Just a nit pick, but I disagree a bit with the wording of the note about these methods not being available. AFAIK GDScript is just implicitly converting the StringName to String behind your back, so using `ToString` in C# is doing the same thing but more explicitly.

We discussed in [RC](https://chat.godotengine.org/channel/dotnet?msg=hKNuwHB89usJxrKy2) a while ago that StringName should be added to the list of classes with notable differences in C#:

https://github.com/godotengine/godot/blob/179dfdc8d78b5bd5377dd115af026df58308bdaf/doc/tools/make_rst.py#L105

It was an oversight when I added this note in https://github.com/godotengine/godot/pull/79206.

Mickeon: Okiedokie. What are you suggesting the note should say then, if anything?

doc/classes/String.xml
```diff
@@ -96,13 +96,14 @@
...
 				"sceneFile_path".capitalize() # Returns "Scene File Path"
+				"2D, FPS, PNG".capitalize()   # Returns "2d, Fps, Png"
 				[/gdscript]
 				[csharp]
 				"move_local_x".Capitalize();   // Returns "Move Local X"
 				"sceneFile_path".Capitalize(); // Returns "Scene File Path"
+				"2D, FPS, PNG".Capitalize();   // Returns "2d, Fps, Png"
 				[/csharp]
 				[/codeblocks]
-				[b]Note:[/b] This method not the same as the default appearance of properties in the Inspector dock, as it does not capitalize acronyms ([code]"2D"[/code], [code]"FPS"[/code], [code]"PNG"[/code], etc.) as you may expect.
```
jsjtxietian, line 105: This method "is" not the same as the default appearance of properties in the Inspector dock ?

doc/classes/StringName.xml
raulsntos: How about something like this?

```suggestion
		[b]Note:[/b] In C#, an explicit conversion to [code]System.String[/code] is required to use the methods listed on this page. Use the [code]ToString()[/code] method to cast a [StringName] to a string, and then use the equivalent methods in [code]System.String[/code] or [code]StringExtensions[/code]. For convenience, a list of equivalent methods is provided [url=$DOCS_URL/tutorials/scripting/c_sharp/c_sharp_differences.html#string]here[/url].
```

We probably don't even need the reference to the differences documentation page since it would be in the note added when the `StringName` class is added to the `CLASSES_WITH_CSHARP_DIFFERENCES` list. For reference, the note looks like this:

image

doc/classes/String.xml
Mickeon: Red and the minus sign means that this PR removes the sentence outright. Yeah, it was missing the "is", too.

Mickeon:
Sounds good. I changed and removed part of the **StringName** C# note. I would also say that the PR no longer fully addresses https://github.com/godotengine/godot-docs/issues/8190 because it needs to be stated clearly in the C# differences page, too.

raulsntos:
I think you could make the change to `CLASSES_WITH_CSHARP_DIFFERENCES` in this PR:

```diff
diff --git a/doc/tools/make_rst.py b/doc/tools/make_rst.py
index c3a21f3d7b..df3e928538 100755
--- a/doc/tools/make_rst.py
+++ b/doc/tools/make_rst.py
@@ -105,6 +105,7 @@ EDITOR_CLASSES: List[str] = [
 CLASSES_WITH_CSHARP_DIFFERENCES: List[str] = [
     "@GlobalScope",
     "String",
+    "StringName",
     "NodePath",
     "Signal",
     "Callable",

```

Mickeon:
Oh that's neat, I did not realise part of it could have been done in the same repository. Done.

raulsntos (approved):
LGTM.

Mickeon:
Addressed the above feedback with an additional note. I didn't want to add the explanation in the same line, to keep the short summary consistent to the other similar methods.

akien-mga (approved):

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

YuriSizov:
Cherry-picked for 4.1.4.

PR #86643, opened 2023-12-30
Turn off distraction free mode when selecting default layout

ryevdokimov:
Currently the behavior when selecting the default layout while distraction free mode is enabled is a bit odd:

image

The panels are restored; however, they are blank, and distraction free is still enabled according to its respective button. 

This is resolved by turning off distraction free mode, which I believe is the desired behavior when expecting the layout to return to a true default.

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -5566,6 +5566,7 @@ void EditorNode::_layout_menu_option(int p_id) {
 			layout_dialog->popup_centered();
 		} break;
 		case SETTINGS_LAYOUT_DEFAULT: {
+			set_distraction_free_mode(false);
 			_load_docks_from_config(default_layout, "docks");
 			_save_editor_layout();
 		} break;


AThousandShips:
I'm not sure this is the correct solution, instead the incorrect application of the panels should be fixed, it is unexpected IMO that applying the layout disables distraction free mode

ryevdokimov:
> I'm not sure this is the correct solution, instead the incorrect application of the panels should be fixed, it is unexpected IMO that applying the layout disables distraction free mode.

Fixing the incorrect application of the panels results in effectively the same thing, but with the addition of distraction free mode being paradoxically enabled (since the panels are returned). The alternative would be is to have the default layout restored, but to have the panels remain hidden. Which behavior is correct I suppose depends on what the definition of "default layout" is. 

In my opinion, it should return all panels, including the scene viewport to a state user are familiar with when they first open Godot without any modifications.

AThousandShips:
I'd expect setting default layout to be equivalent to:
* Disabling distraction free mode
* Restoring layout
* Enabling distraction free mode

So that when I turn it off the panels are back at their default position, without turning off the mode

Setting the default layout doesn't change other things in the current setup but instead just applies the format, keeping what ever menus you have open

ryevdokimov:
> I'd expect setting default layout to be equivalent to:
> 
> * Disabling distraction free mode
> * Restoring layout
> * Enabling distraction free mode
> 
> So that when I turn it off the panels are back at their default position, without turning off the mode
> 
> Setting the default layout doesn't change other things in the current setup but instead just applies the format, keeping what ever menus you have open

I try to view this from the perspective of user intention. Why would a user select "Default Layout" while distraction free mode is enabled? They aren't able to see what the current layout even is. 

I could see this an option to help users who have accidently enabled Distraction Free Mode and are simply wanting the layout to look like what it did before.

ryevdokimov:
I've kind of changed my opinion on this, since I forgot you can detach panels, and this restores them from floating. So, I could probably lean either way.

ryevdokimov:
No longer relevant.

Issue #86611, opened 2023-12-29
godot 4.2 doesn't have AnimationPlayer.add_animation anymore?


chun1617:
### Tested versions

Windows 10 - Godot_v4.2-stable_win64

### System information

Windows 10 - Godot_v4.2-stable_win64

### Issue description

When I try to use add_animation in AnimationPlayer, it looks like doesn't have it anymore. Is it changed to others?

### Steps to reproduce

$AnimationPlayer.add_animation

### Minimal reproduction project (MRP)

animation_player.add_animation(anim_name, anim)

AThousandShips:
You need to use the `AnimationLibrary` interface, use `get_animation_library`

chun1617:
> You need to use the `AnimationLibrary` interface, use `get_animation_library`
> 
> Instructions for this should be added to [this page](https://docs.godotengine.org/en/stable/tutorials/migrating/upgrading_to_godot_4.html#updating-scripts-to-take-backwards-incompatible-changes-into-account), which arguably makes this an issue for the [documentation](https://github.com/godotengine/godot-docs/issues)

May I ask how could I use get_animation_library to add new animation frame?
I tried the following code but look like is wrong:
`func setup_animations(data: Dictionary) -> void:
	var anim_library = animation_player.get_animation_library("")

	for anim_name in data["animations"].keys():
		var anim = Animation.new()
		anim.length = data["animations"][anim_name]["frames"].size() * 0.05
		var track_index = anim.add_track(Animation.TYPE_VALUE)

		for frame in data["animations"][anim_name]["frames"]:
			var frame_number = frame["y"] * sprite.hframes + frame["x"]
			var frame_time = frame_number * 0.05
			anim.track_set_path(track_index, String(sprite.get_path()) + ":frame")
			anim.track_insert_key(track_index, frame_time, frame_number)

		anim_library.add_animation(anim_name, anim)

	animation_player.set_animation_library(anim_library)`

AThousandShips:
You can't set it, you just update it directly, see the [documentation](https://docs.godotengine.org/en/stable/classes/class_animationmixer.html#class-animationmixer-method-get-animation-library)

No need to call any set method (there isn't one, you can have multiple libraries)

AThousandShips:
But while the documentation for migration from 3.x should mention this I'd say that you should ask in the other [community channels](https://godotengine.org/community) for support 🙂, this is for reporting bugs

AThousandShips:
[...]

PR #86647, opened 2023-12-30
Fix invalid 3-to-4 renames of `add_animation` to `add_animation_library`

### LINKED ISSUE #86646: Upgrading from 3 to 4 changes functions to invalid functions
### Tested versions

Reproducible when upgrading from 3.5 to 4.2 (and possibly other versions, I was only upgrading from 3.5 to 4.2).

### System information

Windows 10, Godot 4.2

### Issue description

Upgrading the project with the `--convert-3to4` flag changes functions to invalid functions. For example, the SpriteFrame's `AddAnimation` method was supposedly updated to `AddAnimationLibrary`, however this results in a compile error saying:
> 'SpriteFrames' does not contain a definition for 'AddAnimationLibrary' and no accessible extension method 'AddAnimationLibrary' accepting a first argument of type 'SpriteFrames' could be found (are you missing a using directive or an assembly reference?)

When I look in the assembly for SpriteFrames (Assembly GodotSharp, Version=4.2.0.0), I see that it contains a definition for `AddAnimation` but not `AddAnimationLibrary`. There are several other upgraded methods that result in this same error.

Am I missing something? This [cheat sheet](https://godot-3to4.vercel.app/) seems to indicate that the upgrade was supposed to change that method to `AddAnimationLibrary`, but why does the assembly define the old method name and not the new one?

**URL to the documentation page:**
https://docs.godotengine.org/en/4.0/tutorials/migrating/upgrading_to_godot_4.html

### Steps to reproduce

Have a Godot 3.5 project and upgrade to 4.2

This is more a general question. The project in question is private, and I suspect this applies to any project upgrading from 3 to 4.

AThousandShips:
This rename breaks `SpriteFrames` and also isn't valid as the new method takes an `AnimationLibrary`, not an `Animation`

This change needs a more elaborate fix and should be mentioned in the migration page, see:
* https://github.com/godotengine/godot/issues/86611

* Fixes: https://github.com/godotengine/godot/issues/86646

### DIFF
--- a/editor/renames_map_3_to_4.cpp
+++ b/editor/renames_map_3_to_4.cpp
@@ -163,6 +163,7 @@ const char *RenamesMap3To4::gdscript_function_renames[][2] = {
 
 	// { "_set_name", "get_tracker_name" }, // XRPositionalTracker -- CameraFeed uses this.
 	// { "_unhandled_input", "_unhandled_key_input" }, // BaseButton, ViewportContainer -- Breaks Node, FileDialog, SubViewportContainer.
+	// { "add_animation", "add_animation_library" }, // AnimationPlayer -- Breaks SpriteFrames (and isn't a correct conversion).
 	// { "create_gizmo", "_create_gizmo" }, // EditorNode3DGizmoPlugin -- May be used.
 	// { "get_dependencies", "_get_dependencies" }, // ResourceFormatLoader -- Breaks ResourceLoader.
 	// { "get_extents", "get_size" }, // BoxShape, RectangleShape -- Breaks Decal, VoxelGI, GPUParticlesCollisionBox, GPUParticlesCollisionSDF, GPUParticlesCollisionHeightField, GPUParticlesAttractorBox, GPUParticlesAttractorVectorField, FogVolume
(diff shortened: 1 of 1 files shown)


editor/renames_map_3_to_4.cpp
```diff
@@ -644,7 +644,6 @@ const char *RenamesMap3To4::csharp_function_renames[][2] = {
 	{ "_SetPlaying", "SetPlaying" }, // AnimatedSprite3D
 	{ "_ToplevelRaiseSelf", "_TopLevelRaiseSelf" }, // CanvasItem
 	{ "_UpdateWrapAt", "_UpdateWrapAtColumn" }, // TextEdit
-	{ "AddAnimation", "AddAnimationLibrary" }, // AnimationPlayer
```
AThousandShips, line 647: As none of the commented out conversions are mentioned in the same way in the C# part I just removed it, can add it as a commented out part if desired

editor/renames_map_3_to_4.cpp
```diff
@@ -163,6 +163,7 @@ const char *RenamesMap3To4::gdscript_function_renames[][2] = {
 
 	// { "_set_name", "get_tracker_name" }, // XRPositionalTracker -- CameraFeed uses this.
 	// { "_unhandled_input", "_unhandled_key_input" }, // BaseButton, ViewportContainer -- Breaks Node, FileDialog, SubViewportContainer.
+	// { "add_animation", "add_animation_library" }, // AnimationPlayer -- Breaks SpriteFrames (and isn't a correct conversion).
```
AThousandShips, line 166: Technically this now applies to `AnimationMixer`, but wanting to allow cherry picking of this so not changing this

akien-mga (approved):

akien-mga:
Thanks!

AThousandShips:
Thank you!

YuriSizov:
Cherry-picked for 4.2.2. (Included the removal from https://github.com/godotengine/godot/pull/86301 since it was conflicting with it. Also removed the C# variant during the cherry-pick for simplicity, but here's the PR for `master` https://github.com/godotengine/godot/pull/87533)

YuriSizov:
Cherry-picked for 4.1.4. (Same caveat.)

PR #86650, opened 2023-12-30
Keep dock slots hidden when default layout is selected and distractio…

ryevdokimov:
Alternative to: #86643

Currently the behavior when selecting the default layout while distraction free mode is enabled is a bit odd:

image

The panels are restored; however, they are blank, and distraction free is still enabled according to its respective button.

This is resolved by keeping dock slots hidden if distraction free mode is enabled.


ryevdokimov:
This seems to have been fixed in a different PR.

PR #85141, opened 2023-11-20
Fix android wrapper comments

Alex2782:
Only two comments adjusted, new files:
- [GodotRenderView.java](https://github.com/godotengine/godot/blob/master/platform/android/java/lib/src/org/godotengine/godot/GodotRenderView.java)
- [Godot.kt](https://github.com/godotengine/godot/blob/master/platform/android/java/lib/src/org/godotengine/godot/Godot.kt)

PR #83792, opened 2023-10-22
Fix comment typo in `gdscript_parser.h`

ryanabx:
Very simple PR. Found a typo while looking through `gdscript_parser.h` today.

PR #86652, opened 2023-12-30
Fix typo in XRInterface comment

camren-m:
Corrects a typo (the -> they) in xr_interface.h on the XRInterface class' comment.

### DIFF
--- a/servers/xr/xr_interface.h
+++ b/servers/xr/xr_interface.h
@@ -44,7 +44,7 @@ struct BlitToScreen;
 	The idea is that we subclass this class, implement the logic, and then instantiate a singleton of each interface
 	when Godot starts. These instances do not initialize themselves but register themselves with the AR/VR server.
 
-	If the user wants to enable AR/VR the choose the interface they want to use and initialize it.
+	If the user wants to enable AR/VR, they can choose the interface they want to use and initialize it.
 
 	Note that we may make this into a fully instantiable class for GDExtension support.
 */


AThousandShips:

This is a very small change for a whole PR, I think this should instead be part of a larger style and spelling fix PR that are some times done, along with applying spell checking directly

cmumme:
> Thank you for your contribution
> 
> This is a very small change for a whole PR, I think this should instead be part of a larger style and spelling fix PR that are some times done, along with applying spell checking directly

Hi, thank you for your quick response. Similar PRs about the same size have been merged in the past. (e.g. #83792) Despite this, merging this into a larger PR as has also been done in cases like #85141 would also make sense.

servers/xr/xr_interface.h
```diff
@@ -43,7 +43,7 @@ struct BlitToScreen;
 	The idea is that we subclass this class, implement the logic, and then instantiate a singleton of each interface
 	when Godot starts. These instances do not initialize themselves but register themselves with the AR/VR server.
 
-	If the user wants to enable AR/VR the choose the interface they want to use and initialize it.
+	If the user wants to enable AR/VR they choose the interface they want to use and initialize it.
```
fire, line 46:
```suggestion
	If the user wants to enable AR/VR, they can choose the interface they want to use and initialize it.
```

dsnopek (approved):
Looks good to me!

Issue #81998, opened 2023-09-20
Scaled Sprite2D displays part of previous animation frame


jackwilsdon:
### Godot version

v4.1.1.stable.official [bd6af8e0e]

### System information

Godot v4.1.1.stable - Windows 10.0.19044 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2060 SUPER (NVIDIA; 31.0.15.3713) - AMD Ryzen 5 3600X 6-Core Processor (12 Threads)

### Issue description

When scaling a Sprite2D by an odd amount (e.g. 11), part the previous animation frame is incorrectly displayed.

For example, with this sprite sheet (scaled up 10x for GitHub):

![Spritesheet](https://github.com/godotengine/godot/assets/1843197/533b2d89-e261-4373-94fc-d80ad9b62e12)

A sprite with a scale of 11 looks like this on the second frame, both in-editor (at 100% zoom) and in-game:
![Sprite showing issue](https://github.com/godotengine/godot/assets/1843197/957c4daa-7379-41d8-9211-dfbff8cca467)

Scaling it by an even amount fixes the issue (12 in this case):
![Sprite not showing the issue](https://github.com/godotengine/godot/assets/1843197/9246d937-e130-4512-a538-30390c8d479d)

Disabling centering also fixes the issue, even with an odd scale (11 in this case):
![Sprite not showing the issue](https://github.com/godotengine/godot/assets/1843197/b8d49a2b-10a6-4395-ba6d-680730950d84)

Zooming out in the editor also fixes the issue:
![Sprite not showing the issue](https://github.com/godotengine/godot/assets/1843197/87411964-a425-4290-908a-c86fc9583d6b)

### Steps to reproduce

1. Download and import the minimal reproduction project
2. Switch to the 2D viewport
3. Note that the bug is present on the oddly-scaled Sprite2D (the left one):
![Editor screenshot showing the bug](https://github.com/godotengine/godot/assets/1843197/4077972e-600e-4ce0-a68a-866cd49f858b)
4. Run the project
5. Note that the bug is present on the oddly-scaled Sprite2D (the left one):
![In-game screenshot showing the bug](https://github.com/godotengine/godot/assets/1843197/8eca1017-83d7-41a0-a89b-453ca957e5f1)

### Minimal reproduction project

[sprite2d-scaling-bug.zip](https://github.com/godotengine/godot/files/12676222/sprite2d-scaling-bug.zip)

NibblyPig:
[...]

PR #86657, opened 2023-12-31
Add Separation to Sprite2D

### LINKED ISSUE #8755: Add Separation property to Sprite2D

A 2d local multiplayer platformer.

The problem I am having in my project is that I cannot use the same atlas texture for both AnimatedSprite2D and Sprite2D nodes.

An atlas texture is a single image that contains multiple frames for animation. When creating SpriteFrames for an AnimatedSprite2D Godot allows me to specify a **Separation** value, which is the gap between each frame in the atlas texture.

However, Sprite2D does not have this option, so it cannot display the correct frame when driven by an AnimationPlayer.

This limitation forces me to use different textures for each node, which is inefficient and inconvenient.

Add a Separation value to Sprite2D under the Animation category, allowing a user to define the separation along with the hframes and vframes.

I've created a branch that implements this proposal and submitted a PR here:

It cannot be worked around with a few lines of script.

Sprite2D is a core component of Godot.

zolrath:
### What?
This PR adds a new property called Separation to the Sprite2D node. This property allows the user to specify the gap size between each frame in an atlas texture. An atlas texture is a single image that contains multiple frames for animation.

### Why?
This feature is useful for users who want to use the same atlas texture for both AnimatedSprite2D and Sprite2D nodes. Currently, AnimatedSprite2D has a Separation property, but Sprite2D does not. This means that Sprite2D cannot display the correct frame when animated by an AnimationPlayer, unless the user uses a different texture without gaps. This limitation forces the user to use more textures than necessary, which is inefficient and inconvenient.

### How?
The implementation of this feature is similar to how SpriteFrames handle the Separation property. The Sprite2D node calculates the correct texture region based on the Hframes and Vframes properties, as well as the Separation property.

### Testing?
To test this feature, the user can create a Sprite2D node and assign an atlas texture to it. The user can then adjust the Hframes, Vframes, and Separation properties to see the effect on the sprite. When changing the Frame property the correct frame is displayed.

### Additional Benefits
As an aside, this also allows a user to avoid this issue: https://github.com/godotengine/godot/issues/81998 when working with a scaled Sprite2D. With separation between frames the issue no longer presents itself, even if the underlying issue is still present.

### Screenshot
image

### Proposal
This PR implements the following proposal: https://github.com/godotengine/godot-proposals/issues/8755

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/8755*

### DIFF
--- a/scene/2d/sprite_2d.cpp
+++ b/scene/2d/sprite_2d.cpp
@@ -86,9 +86,23 @@ void Sprite2D::_get_rects(Rect2 &r_src_rect, Rect2 &r_dst_rect, bool &r_filter_c
 		base_rect = Rect2(0, 0, texture->get_width(), texture->get_height());
 	}
 
+	if (separation.width > 0 || separation.height > 0) {
+		base_rect.size.x -= ((hframes - 1) * separation.width);
+		base_rect.size.y -= ((vframes - 1) * separation.height);
+	}
+
 	Size2 frame_size = base_rect.size / Size2(hframes, vframes);
-	Point2 frame_offset = Point2(frame % hframes, frame / hframes);
+
+	int hposition = frame % hframes;
+	int vposition = frame / hframes;
+
+	Point2 frame_offset = Point2(hposition, vposition);
 	frame_offset *= frame_size;
+
+	if (separation.width > 0 || separation.height > 0) {
+		frame_offset.x += hposition * separation.width;
+		frame_offset.y += vposition * separation.height;
+	}
 
 	r_src_rect.size = frame_size;
 	r_src_rect.position = base_rect.position + frame_offset;
@@ -329,6 +343,21 @@ int Sprite2D::get_hframes() const {
 	return hframes;
 }
 
+void Sprite2D::set_separation(const Vector2i &p_separation) {
+	if (separation == p_separation) {
+		return;
+	}
+
+	separation = Vector2i(0, 0).max(p_separation);
+
+	queue_redraw();
+	item_rect_changed();
+}
+
+Vector2i Sprite2D::get_separation() const {
+	return separation;
+}
+
 bool Sprite2D::is_pixel_opaque(const Point2 &p_point) const {
 	if (texture.is_null()) {
 		return false;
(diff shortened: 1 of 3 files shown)


scene/2d/sprite_2d.cpp
```diff
@@ -329,6 +343,30 @@ int Sprite2D::get_hframes() const {
...
+		separation.width = p_separation.width;
+	} else {
+		separation.width = 0;
+	}
+
+	if (p_separation.height >= 0) {
+		separation.height = p_separation.height;
+	} else {
+		separation.height = 0;
+	}
```
AThousandShips, line 360:
```suggestion
	
	separation = Vector2i(0, 0).max(p_separation);
```
Equivalent and compact, also clearer IMO

scene/2d/sprite_2d.cpp
```diff
@@ -489,6 +535,7 @@ void Sprite2D::_bind_methods() {
 	ADD_GROUP("Animation", "");
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "hframes", PROPERTY_HINT_RANGE, "1,16384,1"), "set_hframes", "get_hframes");
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "vframes", PROPERTY_HINT_RANGE, "1,16384,1"), "set_vframes", "get_vframes");
+	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2I, "separation", PROPERTY_HINT_NONE, "suffix:px"), "set_separation", "get_separation");
```
AThousandShips, line 538:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2I, "separation", PROPERTY_HINT_RANGE, "0,100,or_greater,suffix:px"), "set_separation", "get_separation");
```
Or some similar range value, this is supported and good to use, you still need to keep the clamping in the method though as this just limits the editor input

scene/2d/sprite_2d.cpp
zolrath: That it is, thanks!

scene/2d/sprite_2d.cpp
zolrath: I tried this but didn't see any real change in the drag behavior and ended up removing it. It still seems to drag to an obscene range with almost no mouse movement.

Readded, in any case!

AThousandShips: Try `"0,100,1,or_greater,suffix:px"`

zolrath: Done!

PR #86659, opened 2023-12-31
Improved synchronization of Transforms during live debug sessions

### LINKED ISSUE #86752: Remote editing a subscene incorrectly applies changes
### Tested versions

Reproduced in Godot 4.2.1 stable

### System information

Windows 10 - 4.2.1 stable

### Issue description

When debugging a Godot application, Godot retains knowledge of subscenes that were instanced as nodes. You can now open these subscenes and edit them. However changing the position, rotation or scale of any child nodes is incorrectly applied to the nodes in the running client.

This behaviour is not noticeable if you've just saved a branch as a subscene as in this scenario the transforms between parent and child scene match.

This behaviour is super noticeable if you instance the same scene multiple times or through code.

### Steps to reproduce

Create a scene, add some child nodes, preferably `MeshInstance3D` nodes so they become something visible and editable.

Create a second scene, instance the first scene into this scene a few times and place them around.
Add the usual stuff like environment, light and camera.

Run this second scene.

In the remote debug tree, open the subscene, now move something around. Notice how all instances move to the same location in global space instead of moving in relation to their parent as would be the expected behaviour.

Video showing this in action with the attached MRP: https://youtu.be/tsSmrvXtWcs

[TestSubscene.zip](https://github.com/godotengine/godot/files/13821737/TestSubscene.zip)

MajorMcDoom:
### What problems was this meant to address?
- Some editor transform operations (such as dragging 3D nodes) are committed to the undo/redo system with their _global_ transforms, which are in turn synced to the running game. This means that all the corresponding child nodes in all instances of the scene in the running game are moved to the same location in _world space_, which is not intuitive or useful.
- Furthermore, when you move the _root_ node of a scene (by any means), this operation is synced to the running game. This means that all instances of the edited scene in the running game are moved to the same location, which is also not intuitive or useful.

In the GIF below, you can see the current, undesirable outcome:
old_3d

### What was changed?
- Transform operations are now committed to undo/redo in local instead of global space.
- Transform edits to the root node of the current edited scene are no longer synced to the running game.

In the GIF below, you can see the new, expected behaviour:
fixed_3d

### Are there any caveats or tradeoffs?
- Committing operations in local space will not introduce any new issues, since the undo/redo chronology is maintained, so there is parity with the old behaviour when strictly considering the in-editor workflow. In the niche case of parents being moved around by tool scripts, there will not be parity with the old behaviour. However, it is more intuitive, consistent, and predictable for the operations to be undone/redone in local space in this use case anyway, for the same reasons as in the running game.
- You will no longer be able to move a scene in the running game by moving its scene root in the editor. However, this is arguably something that should never have been possible, as this should be done instead by editing the scene which _contains_ the scene you want to move. This is more consistent with the usual editor workflow for nested scenes anyway, where editing the root transform of a scene does not affect instances of that scene in other scenes.
- Root scale is no longer propagated to the running game. Unlike position and rotation, it is a possibility that the user would expect the root scale to be consistent and propagated across all instances of the scene. However, there is also a possibility that all instances already had individual scales applied (e.g., lots of trees with slight variations in scale). This is a tradeoff, and I think being unable to preview a scale change is less destructive than wiping out existing scales on all instances. Another reason is that many of the synchronized operations are all-inclusive, and teasing out the scale is adding more complexity and overhead.

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -3295,8 +3295,10 @@ void Node3DEditorViewport::_menu_option(int p_option) {
 					xform.basis.rotate_local(Vector3(1, 0, 0), Math_TAU * 0.25);
 				}
 
-				undo_redo->add_do_method(sp, "set_global_transform", xform);
-				undo_redo->add_undo_method(sp, "set_global_transform", sp->get_global_gizmo_transform());
+				Node3D *parent = sp->get_parent_node_3d();
+				Transform3D local_xform = parent ? parent->get_global_transform().inverse_xform(xform) : xform;
+				undo_redo->add_do_method(sp, "set_transform", local_xform);
+				undo_redo->add_undo_method(sp, "set_transform", sp->get_local_gizmo_transform());
 			}
 			undo_redo->commit_action();
 
@@ -4362,16 +4364,7 @@ bool Node3DEditorViewport::_create_instance(Node *parent, String &path, const Po
 
 	Node3D *node3d = Object::cast_to<Node3D>(instantiated_scene);
 	if (node3d) {
-		Transform3D gl_transform;
-		Node3D *parent_node3d = Object::cast_to<Node3D>(parent);
-		if (parent_node3d) {
-			gl_transform = parent_node3d->get_global_gizmo_transform();
-		}
-
-		gl_transform.origin = preview_node_pos;
-		gl_transform.basis *= node3d->get_transform().basis;
-
-		undo_redo->add_do_method(instantiated_scene, "set_global_transform", gl_transform);
+		undo_redo->add_do_method(instantiated_scene, "set_transform", node3d->get_transform());
 	}
 
 	return true;
@@ -4607,8 +4600,8 @@ void Node3DEditorViewport::commit_transform() {
 			continue;
 		}
 
-		undo_redo->add_do_method(sp, "set_global_transform", sp->get_global_gizmo_transform());
-		undo_redo->add_undo_method(sp, "set_global_transform", se->original);
+		undo_redo->add_do_method(sp, "set_transform", sp->get_local_gizmo_transform());
+		undo_redo->add_undo_method(sp, "set_transform", se->original_local);
 	}
 	undo_redo->commit_action();
 
@@ -6168,8 +6161,10 @@ void Node3DEditor::_xform_dialog_action() {
 			tr.origin += t.origin;
 		}
 
-		undo_redo->add_do_method(sp, "set_global_transform", tr);
-		undo_redo->add_undo_method(sp, "set_global_transform", sp->get_global_gizmo_transform());
+		Node3D *parent = sp->get_parent_node_3d();
+		Transform3D local_tr = parent ? parent->get_global_transform().inverse_xform(tr) : tr;
+		undo_redo->add_do_method(sp, "set_transform", local_tr);
+		undo_redo->add_undo_method(sp, "set_transform", sp->get_transform());
 	}
 	undo_redo->commit_action();
 }
@@ -7518,8 +7513,10 @@ void Node3DEditor::_snap_selected_nodes_to_floor() {
 					new_transform.origin.y = result.position.y;
 					new_transform.origin = new_transform.origin - position_offset;
 
-					undo_redo->add_do_method(sp, "set_global_transform", new_transform);
-					undo_redo->add_undo_method(sp, "set_global_transform", sp->get_global_transform());
+					Node3D *parent = sp->get_parent_node_3d();
+					Transform3D new_local_xform = parent ? parent->get_global_transform().inverse_xform(new_transform) : new_transform;
+					undo_redo->add_do_method(sp, "set_transform", new_local_xform);
+					undo_redo->add_undo_method(sp, "set_transform", sp->get_transform());
 				}
 			}
 
--- a/editor/plugins/canvas_item_editor_plugin.cpp
+++ b/editor/plugins/canvas_item_editor_plugin.cpp
@@ -5775,7 +5775,11 @@ void CanvasItemEditorViewport::_create_nodes(Node *parent, Node *child, String &
 
 	// there's nothing to be used as source position so snapping will work as absolute if enabled
 	target_position = canvas_item_editor->snap_point(target_position);
-	undo_redo->add_do_method(child, "set_global_position", target_position);
+
+	CanvasItem *parent_ci = Object::cast_to<CanvasItem>(parent);
+	Point2 local_target_pos = parent_ci ? parent_ci->get_global_transform().xform_inv(target_position) : target_position;
+
+	undo_redo->add_do_method(child, "set_position", local_target_pos);
 }
 
 bool CanvasItemEditorViewport::_create_instance(Node *parent, String &path, const Point2 &p_point) {
--- a/scene/debugger/scene_debugger.cpp
+++ b/scene/debugger/scene_debugger.cpp
@@ -630,7 +630,34 @@ void LiveEditor::_node_set_func(int p_id, const StringName &p_prop, const Varian
 		}
 		Node *n2 = n->get_node(np);
 
+		// Do not change transform of edited scene root, unless it's the scene being played.
+		// See GH-86659 for additional context.
+		bool keep_transform = (n2 == n) && (n2->get_parent() != scene_tree->root);
+		Variant orig_tf;
+
+		if (keep_transform) {
+			if (n2->is_class("Node3D")) {
+				orig_tf = n2->call("get_transform");
+			} else if (n2->is_class("CanvasItem")) {
+				orig_tf = n2->call("_edit_get_state");
+			}
+		}
+
 		n2->set(p_prop, p_value);
+
+		if (keep_transform) {
+			if (n2->is_class("Node3D")) {
+				Variant new_tf = n2->call("get_transform");
+				if (new_tf != orig_tf) {
+					n2->call("set_transform", orig_tf);
+				}
+			} else if (n2->is_class("CanvasItem")) {
+				Variant new_tf = n2->call("_edit_get_state");
+				if (new_tf != orig_tf) {
+					n2->call("_edit_set_state", orig_tf);
+				}
+			}
+		}
 	}
 }
 
@@ -674,8 +701,35 @@ void LiveEditor::_node_call_func(int p_id, const StringName &p_method, const Var
 		}
 		Node *n2 = n->get_node(np);
 
+		// Do not change transform of edited scene root, unless it's the scene being played.
+		// See GH-86659 for additional context.
+		bool keep_transform = (n2 == n) && (n2->get_parent() != scene_tree->root);
+		Variant orig_tf;
+
+		if (keep_transform) {
+			if (n2->is_class("Node3D")) {
+				orig_tf = n2->call("get_transform");
+			} else if (n2->is_class("CanvasItem")) {
+				orig_tf = n2->call("_edit_get_state");
+			}
+		}
+
 		Callable::CallError ce;
 		n2->callp(p_method, p_args, p_argcount, ce);
+
+		if (keep_transform) {
+			if (n2->is_class("Node3D")) {
+				Variant new_tf = n2->call("get_transform");
+				if (new_tf != orig_tf) {
+					n2->call("set_transform", orig_tf);
+				}
+			} else if (n2->is_class("CanvasItem")) {
+				Variant new_tf = n2->call("_edit_get_state");
+				if (new_tf != orig_tf) {
+					n2->call("_edit_set_state", orig_tf);
+				}
+			}
+		}
 	}
 }
 


AThousandShips:
Unsure if this breaks compatibility or not, as some possibly relevant behaviour changes, needs to be discussed

MajorMcDoom:
Might be worth mentioning some context here:

As a new Godot user a few months ago, I saw the strange behaviour of live debug transforms and assumed it was a network bug. It had me feeling very lost as to what it was supposed to achieve, but it definitely didn't help me in any way.

With the new local-to-scene synchronizations, the full workflow can now be described as de facto realtime prefab editing, which not only covers a significant portion of the realtime editing features many Unity migrants have been asking for, but actually goes far beyond, since changes affect _all_ instances, _plus_ you don't have to remember and reapply the runtime changes after you stop debugging.

Editing all other node/script properties and resources in a scene at runtime has been seamless, and I think correctly syncing child transforms is the missing piece to that workflow.

BastiaanOlij:
I agree with @AThousandShips that this needs to be discussed with a few developers that are better informed about how this was originally designed to work.

That said, I just reproduced the issue, it's not clearly enough described by @MajorMcDoom, but this is a bug and his fix may be the right fix.

From what I understand of the inner workings of the remote debug feature, as scenes are instantiated and added to the scene tree within the game this creates one large scene tree. It is this scene tree that you see in the remote debug tab and that you manipulate. As we're working with one big tree, where we are manipulating a single node in the tree, updating the global or local transform should have the same result. 

But from the description of @MajorMcDoom we seem to have added the ability to open the subscenes and edit them and sync them. As mentioned above I just tried this, I wasn't aware we had this feature, it is SUPER COOL, but it's also SUPER broken :)

When editing a subscene you can't manipulate the global transform, yes we've shot ourselves in the foot that when you right click on a node and save the node as a subscene, that we save that scene with it's transforms intact which can give the illusion things are correct. But anything more complex, say having a subscene in a subscene, or having multiple instances of a subscene, and you immediately see the problem.

I just did a very simple experiment where I created a subscene with some meshes, instanced it multiple times in a parent scene and placed them around the scene, and then tried to edit the subscene to see what happened. Any change to the subscene ruined the running scene by placing all instances in the same spot. Not what I would call expected behaviour.

This also means that the 2nd and 3rd caveats are basically non-issues. When we're dealing with a subscene the root transform isn't change-able in a meaningful way, especially if the scene is instanced multiple times, as the real transform applied is in the parent scene. While if you've just saved a branch as a subscene, you can manipulate the root transform of the subscene and it will be "applied" to the parent scene, this is the only exception to the rule (I've often believed Godot shouldn't do this, and reset the transform in the subscene). 

I need to test if the fix works as advertised, but based on the testing I just did my gut feeling says this is indeed a bug and this fix should fix it.

BastiaanOlij (approved):
Tested this fix using my MRP from #86752

Works like a charm:
https://youtu.be/MPnB2HELDWQ

editor/debugger/script_editor_debugger.cpp
```diff
@@ -1192,6 +1192,19 @@ void ScriptEditorDebugger::_method_changed(Object *p_base, const StringName &p_n
 	}
 
 	if (node) {
+		// Do not send the root transform of edited scene.
```
BastiaanOlij, line 1195: I would expand on this comment and make it clear why we're doing this. I bet someone is going to raise a bug report at some point because they right clicked on a node, saved branch to subscene, and then expect manipulating the root node when editing the subscene does something. As mentioned before, this is the exception to the rule where we could expect this to do something.

In all other scenarios, especially when scenes are instanced by code or when multiple instances are present, the root node of a subscene has no meaning, the transform in the parent scene will always overrule it.

One improvement we could do is to ensure the inspector on the root node is disabled in this condition, but my knowledge of the editor is too limited to suggest how to put this into practice.

MajorMcDoom:
> Tested this fix using my MRP from #86752
> 
> Works like a charm: https://youtu.be/MPnB2HELDWQ

That's good to hear! Just to be clear, there are multiple ways to affect changes to the transforms, like Align Rotation with Camera, and Snap to Floor, sliding the actual spinners in the inspector, dragging gizmos, etc. These all go through different properties and methods, and so my fix had to account for them all, to make sure they all sync correctly, but also maintain parity with old behaviour.

I have done that testing myself, but just noting that for any future testing that might be needed.

Another thing I want to mention is that my solution, for the sake of simplicity, simply filters out certain methods and properties under certain conditions from being sent over to the running game. It does mean that if we ever change/add features that modify transforms through other properties/methods, then those will have to be accounted for as well. I considered making a catch-all solution for transforms (either on the sending side or the receiving side), but the current way the feature is implemented made it difficult, so I opted for the current solution. If anything does come up in the future, they will be tiny edge cases that can be easily addressed.

Also, @BastiaanOlij @AThousandShips considering this is now considered a bug fix, is there any reason it can't be moved to 4.3 milestone?

editor/debugger/script_editor_debugger.cpp
MajorMcDoom: What's the best practice when it comes to mentioning additional context like this that might need a lot of text? Should I just insert it all inline, or can we actually just include pull request numbers?

BastiaanOlij:
> Also, @BastiaanOlij @AThousandShips considering this is now considered a bug fix, is there any reason it can't be moved to 4.3 milestone?

I think some others who are a little more versed in the editor should verify my findings, but if the group concurs with me that this is a bug, and if no further requirements are identified, I think this should indeed be a milestone for 4.3 with a possible cherry-pick for 4.2

AThousandShips:
Needs to be verified as:
* A proper fix
* not undesired side effects

Even if we change it to 4.3 now that's no guarantee it won't be delayed due to confirmation needed, nor is it being on 4.x going to make it go slower 🙂

editor/debugger/script_editor_debugger.cpp
MajorMcDoom: I have expanded the comments.

KoBeWi:
Something is broken with undo:

https://github.com/godotengine/godot/assets/2223172/8fd2f7b6-f7cc-48ca-a1c0-72b715d841f9

I have yet to look at the code, but functionally it looks fine. The change to local transforms seems to have already existed in 2D, so this makes 2D and 3D consistent in the more useful way.
Not synchronizing root transforms also makes sense; the only time you could possibly want otherwise is when you have a single instance of the scene, or when the instance is the scene you are running, but not sure if it can be reasonably determined.
I'm not sure about the scale thing, but the reasoning seems fine.

I think if users complain about any of these changes, we could make the behavior customizable. But for now it's not necessary.

MajorMcDoom:
@KoBeWi 

> Something is broken with undo:

Can you clarify how to repro the undo issue? You can see from the GIFs in my original post that undo does work, so you may have found an edge case that I need to address.

> The change to local transforms seems to have already existed in 2D

This isn't quite true. All the operations in 2D and 3D use different properties and methods to achieve, and some already worked in local space, but not all. I did have to change a couple things. For example, drag-dropping a new sprite into the viewport was in global space.

editor/debugger/script_editor_debugger.cpp
```diff
@@ -1192,6 +1192,23 @@ void ScriptEditorDebugger::_method_changed(Object *p_base, const StringName &p_n
 	}
 
 	if (node) {
+		// Do not send the root transform of edited scene. Without this exception, all instances of the scene
+		// in the running game will snap to the same location, which is not desirable. Even in the case where
+		// there is only a single instance in the running game, its position within its parent scene still
+		// takes priority over its root position within its own scene. The subnodes in the parent scene can be
+		// edited instead to achieve the same goal.
```
KoBeWi, line 1199: Does this comment to be so lengthy?
```suggestion	
		// Do not send the root transform of edited scene, because it affects all instances in undesirable way.
```

MajorMcDoom: That was because of this review:
https://github.com/godotengine/godot/pull/86659#discussion_r1440725295

KoBeWi: :/
You could at least avoid repeating it in 2 places. Or maybe you could leave a link to this PR for further explanation, we do that sometimes, e.g.
https://github.com/godotengine/godot/blob/89cc635c0554cb2e518c830969ca4c5eedda0f4e/editor/editor_themes.cpp#L345

MajorMcDoom: Are there best practices I can refer to somewhere about leaving comments in the codebase?

KoBeWi: There's this:
https://docs.godotengine.org/en/stable/contributing/development/code_style_guidelines.html#comment-style-guide

MajorMcDoom:
@KoBeWi I have found a repro for the undo bug you encountered. Fixing now.

editor/debugger/script_editor_debugger.cpp
MajorMcDoom: Okay, I've amended.

MajorMcDoom:
Undo edge case bug has been fixed.

MajorMcDoom:
UPDATE:
I've changed the way the root transform sync exception works by implementing it on the receiving end, instead of the sending end. The advantages:

- There no longer has to be any special detection of "transform-affecting functions and properties" on the sending end, it's just a catch-all solution, which is much easier to maintain.
- On the receiving end it also now detects if the attempted edit is to the root of the current running game. If so, it does allow the root transform change, since the edit and the debug instance are running in the same context, and there can be no other interpretation of user intent.

KoBeWi (approved):
Functionally looks fine.

I'm wondering about the debugger code though. The PR adds a dependency on Node3D and CanvasItem. It can be avoided by using `is_class()` and `call()` instead (which I think might be preferred, given the dynamic nature of the debugger).

Another thing is that the transform is preserved by remembering old transform and reverting it. I think the previous property-based approach could've been better.

But that's more for the debugger team to asses.

MajorMcDoom:
@KoBeWi 

As per your suggestion, I've updated it to work with variants, `is_class`, and `call` instead, and removed the two #includes.

As for the record-restore approach, I think it makes more sense as a catch-all. If we use the property/method check approach, if anyone in the future changes how "Align Rotation with Camera" works, or introduces a new feature that moves a 2D sprite in the editor, it could break this, and someone would have to go in and reconcile the new property/function list.

Plus it is not something that executes frequently, so performance isn't an issue.

That said, I agree, debugger team feedback on this would be great.

BastiaanOlij:
> Not synchronizing root transforms also makes sense; the only time you could possibly want otherwise is when you have a single instance of the scene, or when the instance is the scene you are running, but not sure if it can be reasonably determined. I'm not sure about the scale thing, but the reasoning seems fine.
> 
> I think if users complain about any of these changes, we could make the behaviour customizable. But for now it's not necessary.

@KoBeWi  I think we need to be really careful with this because it could lead to the user making and saving changes, that once re-running are lost because the parent scene overrides the transform. The expected behavior here would only work if the user has just saved a branch as scene. Any changes the user does from here, even if the subscene is used only once, runs the risk of the parent transform  becoming useless.

So my gut feeling says we should exclude this at all times, and preferably even disable the transform field in debug mode (if we can). Then have a good explanation in a comment as to why we've made this limitation.

Other than that, this is looking really good @MajorMcDoom  :)

MajorMcDoom:
> > Not synchronizing root transforms also makes sense; the only time you could possibly want otherwise is when you have a single instance of the scene, or when the instance is the scene you are running, but not sure if it can be reasonably determined. I'm not sure about the scale thing, but the reasoning seems fine.
> > I think if users complain about any of these changes, we could make the behaviour customizable. But for now it's not necessary.
> 
> @KoBeWi I think we need to be really careful with this because it could lead to the user making and saving changes, that once re-running are lost because the parent scene overrides the transform. The expected behavior here would only work if the user has just saved a branch as scene. Any changes the user does from here, even if the subscene is used only once, runs the risk of the parent transform becoming useless.
> 
> So my gut feeling says we should exclude this at all times, and preferably even disable the transform field in debug mode (if we can). Then have a good explanation in a comment as to why we've made this limitation.
> 
> Other than that, this is looking really good @MajorMcDoom :)

Thank you!

As for your responses to @KoBeWi concerning the exceptional cases to allow, I think we should distinguish between different use cases mentioned, because they do deserve different treatment:

- **Editing the root of a scene that is also the scene being run**. i.e. there is no parent that can move. This case I have already made an exception for in my latest changes, and I think should be allowed. The scene would, structurally and visually, form the entire scope of operation for the user in both windows. It's the most logical interpretation of the user's intent that if they move the whole scene in one, it affects the other.
- **Editing the root of a scene that is a single-instance child in the running game**. In this case, the side effects are bad, as you said. Any change you make to the root in the editor would immediately mess up the runtime position state of the child in the running game.

I'm not sure about your idea of disabling the root Transform in the editor though. First, it's not really practical because it's not just a field - it's _any operation_ that can _affect_ the transform that is being blocked. Second, we don't really want to _prevent edits_ to the project. We simply don't want them to propagate them to the running game in a way that doesn't make sense.

BastiaanOlij:
@MajorMcDoom sorry I wasn't clear, I was purely talking about that second use case (e.g. **Editing the root of a scene that is a single-instance child in the running game.**), the first situation it makes sense it is editable.

Anyway I think the way you made it work is fine, disabling the transform field would be a nice to have and indeed as you say, it would likely remain incomplete.

As long as we clearly document the behaviour and why we did what we did, it's a great fix.

BastiaanOlij:
ping @akien-mga  

Imho this is an important bug fix that should not fall through the cracks. Right now this is marked as needing consensus with no further action seeming to happen nor clarity on who could provide that consensus. This sadly is an area of the editor that doesn't see heavy enough use within core contributors to gain their feedback. Users who do make use of this, won't test this until it's merged, possibly not even until it's in stable.
Waiting on @reduz for this doesn't seem productive, he is far too busy to look into something trivial like this.
So who else could weigh in on this to push it over the line?

@Faless, it's a little hidden in the "4.3 merge queue" thingy but you're tagged there to weigh in on the debugger changes here.  If you have a moment that would be cool!

akien-mga:
Couldn't get a final review, so let's get this merged. Sorry for the delay :)

akien-mga:

MajorMcDoom:

Awesome. Thanks y'all! :)

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -4362,16 +4364,7 @@ bool Node3DEditorViewport::_create_instance(Node *parent, String &path, const Po
...
-		Transform3D gl_transform;
-		Node3D *parent_node3d = Object::cast_to<Node3D>(parent);
-		if (parent_node3d) {
-			gl_transform = parent_node3d->get_global_gizmo_transform();
-		}
-
-		gl_transform.origin = preview_node_pos;
-		gl_transform.basis *= node3d->get_transform().basis;
-
-		undo_redo->add_do_method(instantiated_scene, "set_global_transform", gl_transform);
```
ryevdokimov, line 4374: Removing this code ends up breaking the drag to instantiate scene at cursor position functionality. See mentioned issue.

MajorMcDoom: I cannot seem to replicate this.
placement

clayjohn: @MajorMcDoom you seem to be testing with dev2. Do you get the same results when testing with latest master?

ryevdokimov: Unless I'm misunderstanding something, I can see the build in that video is 4.3dev2. Your PR was merged less than a day ago so it's not going to be in any dev build. Can you test after building from the last commit in your PR?

MajorMcDoom: Well that's embarrassing, I clicked the wrong exe. 😅
I now have a reliable repro, and I will take a look soon. Thank you!

ryevdokimov: All good. I'm not seeing a replacement for the code that was deleted, at the very least I would expect that the position of the new instance gets assigned by preview node's position at some point.

MajorMcDoom: @clayjohn @ryevdokimov Sorry, noob question: Are there special protocols for modifying a PR after it's already been merged? Is it possible to work from here, or should I start a new PR?

ryevdokimov: You would just start a new PR.

PR #86664, opened 2023-12-31
Add code examples to Basis' documentation

Mickeon:
This PR attempts to add code examples to the **Basis** Variant type, which really needs some more detailing. 

More knowledgeable users may already know what a **Basis** is and it feels like this class reference is written for them. For an introduction to the concept, this page is extremely dry, and it makes things look more complicated than they are in practice. _<sub>I am also talking from personal experience here.</sub>_

The original intention was to overhaul the **Basis** class reference outright, but I was keen on splitting the PR in two. This is because I am not sure what would be a... _good enough way_ to do this, and concentrated feedback from more "savvy" users and maintainers is very welcome here.

In particular, it would be nice to find a way to represent the **Basis** matrix in the examples in a way that doesn't look out of place or overly verbose.

One problem:
Each so-called matrix "_column_" is a **Vector3**, which when written tidily, forces the code examples to display them horizontally rather than vertically. It may confusing.

### DIFF
--- a/doc/classes/Basis.xml
+++ b/doc/classes/Basis.xml
@@ -71,6 +71,12 @@
 			<param index="1" name="order" type="int" default="2" />
 			<description>
 				Constructs a pure rotation Basis matrix from Euler angles in the specified Euler rotation order. By default, use YXZ order (most common). See the [enum EulerOrder] enum for possible values.
+				[codeblock]
+				# Creates a Basis whose z axis points down.
+				var my_basis = Basis.from_euler(Vector3(TAU / 4, 0, 0))
+
+				print(my_basis.z) # Prints (0, -1, 0).
+				[/codeblock]
 			</description>
 		</method>
 		<method name="from_scale" qualifiers="static">
@@ -78,6 +84,13 @@
 			<param index="0" name="scale" type="Vector3" />
 			<description>
 				Constructs a pure scale basis matrix with no rotation or shearing. The scale values are set as the diagonal of the matrix, and the other parts of the matrix are zero.
+				[codeblock]
+				var my_basis = Basis.from_scale(Vector3(2, 4, 8))
+
+				print(my_basis.x) # Prints (2, 0, 0).
+				print(my_basis.y) # Prints (0, 4, 0).
+				print(my_basis.z) # Prints (0, 0, 8).
+				[/codeblock]
 			</description>
 		</method>
 		<method name="get_euler" qualifiers="const">
@@ -98,6 +111,18 @@
 			<return type="Vector3" />
 			<description>
 				Assuming that the matrix is the combination of a rotation and scaling, return the absolute value of scaling factors along each axis.
+				[codeblock]
+				var my_basis = Basis(
+				    Vector3(2, 0, 0),
+				    Vector3(0, 4, 0),
+				    Vector3(0, 0, 8)
+				)
+				# Rotating the Basis in any way preserves its scale.
+				my_basis = my_basis.rotated(Vector3.UP, TAU / 2)
+				my_basis = my_basis.rotated(Vector3.RIGHT, TAU / 4)
+
+				print(my_basis.get_scale()) # Prints (2, 4, 8).
+				[/codeblock]
 			</description>
 		</method>
 		<method name="inverse" qualifiers="const">
@@ -140,6 +165,14 @@
 			<return type="Basis" />
 			<description>
 				Returns the orthonormalized version of the matrix (useful to call from time to time to avoid rounding error for orthogonal matrices). This performs a Gram-Schmidt orthonormalization on the basis of the matrix.
+				[codeblock]
+				# Rotate this Node3D every frame.
+				func _process(delta):
+				    basis = basis.rotated(Vector3.UP, TAU * delta)
+				    basis = basis.rotated(Vector3.RIGHT, TAU * delta)
+
+				    basis = basis.orthonormalized()
+				[/codeblock]
 			</description>
 		</method>
 		<method name="rotated" qualifiers="const">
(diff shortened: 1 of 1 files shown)


doc/classes/Basis.xml
```diff
@@ -190,6 +243,18 @@
...
+				    Vector3(1, 2, 3),
+				    Vector3(4, 5, 6),
+				    Vector3(7, 8, 9)
+				)
+
+				var b = a.transposed()
+				print(b.x) # Prints (1, 4, 7).
+				print(b.y) # Prints (2, 5, 8).
+				print(b.z) # Prints (3, 6, 9).
+				[/codeblock]
```
fire, line 257: Mickeon for the use of basis in orientation, it's typically 1, 0, 0, 0, 1, 0, 0, 0, 1. Like normalized.

fire: Like around one, not like 9.

fire: Oh this is in the wrong section header.

Mickeon: The docs for this & similar really want to express it as working with a 3x3 matrix, no mention of basis. For this case I feel like the starting value is appropriate, since transposing is a broad concept that applies to all matrices, not just the Basis type.

doc/classes/Basis.xml
```diff
@@ -148,13 +181,33 @@
 			<param index="1" name="angle" type="float" />
 			<description>
 				Introduce an additional rotation around the given axis by [param angle] (in radians). The axis must be a normalized vector.
+				[codeblock]
+				var my_basis = Basis.IDENTITY
+				var angle = PI / 2
+
+				my_basis = my_basis.rotated(Vector3.UP, angle)      # Rotate around the up axis (yaw)
+				my_basis = my_basis.rotated(Vector3.RIGHT, angle)   # Rotate around the right axis (pitch)
+				my_basis = my_basis.rotated(Vector3.FORWARD, angle) # Rotate around the forward axis (roll)
```
aaronfranke, line 190: The +Z axis is Vector3.BACK. If you rotate around Vector3.FORWARD then the rotation would be the reverse of the usual direction people expect to rotate in a right-handed coordinate system (different from Euler angles of `(0, 0, angle)`)

```suggestion
				my_basis = my_basis.rotated(Vector3.BACK, angle) # Rotate around the back axis (roll)
```

doc/classes/Basis.xml
```diff
@@ -148,13 +181,33 @@
...
+				var my_basis = Basis(
+				    Vector3(1, 1, 1),
+				    Vector3(2, 2, 2),
+				    Vector3(3, 3, 3)
+				)
+				my_basis = basis.scaled(Vector3(0, 2, -2))
+
+				print(my_basis.x) # Prints (0, 2, -2).
+				print(my_basis.y) # Prints (0, 4, -4).
+				print(my_basis.z) # Prints (0, 6, -6).
```
aaronfranke, line 209: `basis` is not defined here, only `my_basis` is defined. But also, why use `my_basis` here but `a` and `b` below? We should pick one convention to use for all examples.

doc/classes/Basis.xml
```diff
@@ -98,6 +111,18 @@
...
 				Assuming that the matrix is the combination of a rotation and scaling, return the absolute value of scaling factors along each axis.
+				[codeblock]
+				var my_basis = Basis(
+				    Vector3(2, 0, 0),
+				    Vector3(0, 4, 0),
+				    Vector3(0, 0, 8)
+				)
+				# Rotating the Basis in any way preserves its scale.
+				my_basis = my_basis.rotated(Vector3.UP, PI)
+				my_basis = my_basis.rotated(Vector3.RIGHT, PI / 2)
```
aaronfranke, line 122: I would suggest using TAU in code examples instead of PI since it results in more readable code.

```suggestion
				my_basis = my_basis.rotated(Vector3.UP, TAU / 2)
				my_basis = my_basis.rotated(Vector3.RIGHT, TAU / 4)
```

This can be read as "perform a half turn around UP and a quarter turn around RIGHT".

doc/classes/Basis.xml
Mickeon: I was wondering the same question myself. `my_basis` is more friendly but repetitive, while using single characters would be more concise but be potentially confusing when combined with `x`, `y` and `z`.

aaronfranke: I'm good with either one, but after thinking about it I am leaning towards `my_basis`, since it shows the common pattern of `thing = thing.operation()`, as some users accidentally write `thing.operation()` and it's discarded.

doc/classes/Basis.xml
Mickeon: I love TAU myself, but I decided to use PI because it's more commonly understood. PI is a must when learning about angles. I could still change it if it would make the code examples more consistent across the board.

doc/classes/Basis.xml
```diff
@@ -190,6 +243,18 @@
 			<return type="Basis" />
 			<description>
 				Returns the transposed version of the matrix.
```
AThousandShips, line 245:
```suggestion
				Returns the transposed version of the matrix, this turns rows into columns and vice versa.
```
I think having a verbal description of what this means along with the example would be good, wording is just a suggestion

Mickeon: I was actively avoiding to touch the descriptions in this PR. I have always gone too far and it's part of the reason as too why all of the past "Overhaul Documentation" PRs have been taking so long to be merged.

AThousandShips: Then let's do that in separately 🙂

Mickeon:
Addressed the above feedback. Also replaced every PI example with TAU.

doc/classes/Basis.xml
```diff
@@ -190,6 +243,18 @@
...
+				[codeblock]
+				var my_basis = Basis(
+				    Vector3(1, 2, 3),
+				    Vector3(4, 5, 6),
+				    Vector3(7, 8, 9)
+				)
+				var my_basis = my_basis.transposed()
```
aaronfranke, line 252:
```suggestion
				my_basis = my_basis.transposed()
```

aaronfranke (approved):

fire (approved):

akien-mga:
Would be good to add the C# counterparts, CC @godotengine/dotnet.

paulloz:
> Would be good to add the C# counterparts, CC @godotengine/dotnet.

I can make a follow-up PR after this is merged.

akien-mga:
Thanks!

Mickeon:
Aww, the triple.

PR #86669, opened 2023-12-31
Stop escaping `'` on POT generation

YeldhamDev:
Escaping the `'` character is unnecessary, and makes the resulting POT incompatible with translations tools such as Weblate.

### DIFF
--- a/editor/pot_generator.cpp
+++ b/editor/pot_generator.cpp
@@ -141,7 +141,7 @@ void POTGenerator::_write_to_pot(const String &p_file) {
 
 			// Write context.
 			if (!context.is_empty()) {
-				file->store_line("msgctxt " + context.c_escape().quote());
+				file->store_line("msgctxt " + context.json_escape().quote());
 			}
 
 			// Write msgid.
@@ -183,11 +183,11 @@ void POTGenerator::_write_msgid(Ref<FileAccess> r_file, const String &p_id, bool
 	}
 
 	for (int i = 0; i < lines.size() - 1; i++) {
-		r_file->store_line((lines[i] + "\n").c_escape().quote());
+		r_file->store_line((lines[i] + "\n").json_escape().quote());
 	}
 
 	if (!last_line.is_empty()) {
-		r_file->store_line(last_line.c_escape().quote());
+		r_file->store_line(last_line.json_escape().quote());
 	}
 }
 


editor/pot_generator.cpp
```diff
@@ -141,7 +141,8 @@ void POTGenerator::_write_to_pot(const String &p_file) {
 
 			// Write context.
 			if (!context.is_empty()) {
-				file->store_line("msgctxt " + context.c_escape().quote());
+				// Escaping `'` is unnecessary, and is incompatible with translations tools like Weblate.
```
AThousandShips, line 144:
```suggestion
				// Escaping `'` is unnecessary, and is incompatible with translation tools like Weblate.
```

editor/pot_generator.cpp
```diff
@@ -182,12 +183,12 @@ void POTGenerator::_write_msgid(Ref<FileAccess> r_file, const String &p_id, bool
 		r_file->store_line("\"\"");
 	}
 
+	// Escaping `'` is unnecessary, and is incompatible with translations tools like Weblate.
```
AThousandShips, line 186:
```suggestion
	// Escaping `'` is unnecessary, and is incompatible with translation tools like Weblate.
```

AThousandShips (commented):
Is it only this escape that is incompatible? Also in what way is it incompatible, a link would be nice to avoid confusion 🙂

YeldhamDev:
@AThousandShips It will trigger errors in tools like Weblate, it happened to me with my own project. I don't know where this information is explicitly written however.

KoBeWi:
Makes me wonder whether the current implementation of `c_escape()` makes sense. You don't need to escape `'` inside `"` string and vice-versa. Maybe it could be somehow accounted for.

editor/pot_generator.cpp
```diff
@@ -182,12 +183,12 @@ void POTGenerator::_write_msgid(Ref<FileAccess> r_file, const String &p_id, bool
 		r_file->store_line("\"\"");
 	}
 
+	// Escaping `'` is unnecessary, and is incompatible with translation tools like Weblate.
```
KoBeWi, line 186: Shouldn't this be line below, for better context?

YeldhamDev: It's explaining the reason for the `replace()` function in both blocks below.

KoBeWi: It does, but currently it may look like it applies to the `for` loop, not the line in loop.

YeldhamDev: So you think it would be better to just have the comment in one of the blocks?

KoBeWi: IMO yes.

YeldhamDev: Done.

KoBeWi (approved):

editor/pot_generator.cpp
```diff
@@ -141,7 +141,8 @@ void POTGenerator::_write_to_pot(const String &p_file) {
 
 			// Write context.
 			if (!context.is_empty()) {
-				file->store_line("msgctxt " + context.c_escape().quote());
+				// Escaping `'` is unnecessary, and is incompatible with translation tools like Weblate.
+				file->store_line("msgctxt " + context.c_escape().replace("\\'", "\'").quote());
```
akien-mga, line 145: We have `json_escape` which is like `c_escape` but doesn't escape `'` and `\a`, whatever that is. That might work here?

Pheubel: > We have `json_escape` which is like `c_escape` but doesn't escape `'` and `\a`, whatever that is. That might work here?

according to [wikipedia](https://en.wikipedia.org/wiki/Bell_character) `\a` is a bell character. i am not sure what it is used for nowadays, but i don't think it will be a problem.

YeldhamDev: Done.

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #84380, opened 2023-11-02
Replace `floor()` with `round()` in pixel snap to transform

KeyboardDanni:
Fixes https://github.com/godotengine/godot/issues/56793.
Fixes https://github.com/godotengine/godot/issues/84632.

Similar to https://github.com/godotengine/godot/pull/76277 except this PR also makes the necessary changes to `renderer_viewport.cpp` to avoid regressions in camera jitter. In fact, with these changes, the camera is actually *more* stable according to my testing.

Before:

https://github.com/godotengine/godot/assets/34800072/dd708c82-9c80-4f2f-8843-7ba8ef04d0e6

After:

https://github.com/godotengine/godot/assets/34800072/8ea44763-9553-4040-90a3-1f3eb2cbe606

**NOTE:** You will still get camera jitter if you are using a script to make a camera follow an object that is moved in idle process. Camera movement is fine if the object is moved in physics process. This is an existing issue which will be addressed in a separate PR.

Test project download:

[PixelPerfectTest.zip](https://github.com/godotengine/godot/files/13244300/PixelPerfectTest.zip)

_Production edit: added fixes https://github.com/godotengine/godot/issues/84632_

PR #86670, opened 2023-12-31
Fix sprite crunch at half-pixels

### LINKED ISSUE #84632: Sprite "crunch" when placed in certain positions on screen
### Godot version

### System information

Godot v4.2.beta5 - Windows 10.0.22621 - Vulkan (Forward+) - dedicated AMD Radeon RX 7900 XTX (Advanced Micro Devices, Inc.; 31.0.22023.1014) - AMD Ryzen 9 7950X3D 16-Core Processor (32 Threads)

### Issue description

In 4.2, there is an issue mostly visible on pixel art games with small resolutions that scale up to larger screens. While the game is running, moving sprites to certain places in the screen will cause the sprites to appear uneven or "crunched", and it's usually more noticeable if the sprite is animated.

example

Window settings:
image

This seems to happen when moving the position of the sprite to X.5.
This happens in all 3 renderers.
This issue does not occur in Godot 4.1.3.
This issue occurs as early as 4.2 dev1.

Note: I'm using a 4K monitor with Windows scaled up to 175%.

### Steps to reproduce

1. Download the reproduction project
2. Open the project
3. Run the game
4. If the issue doesn't immediately show itself, slightly moving the character to the left/right a few times should trigger it.

[SpriteCrunchIssue.zip](https://github.com/godotengine/godot/files/13301032/SpriteCrunchIssue.zip)

adamscott:
Rounds the canvas item position for nearest filtered textures before rendering it.

Fixes #84632

### DIFF
--- a/drivers/gles3/rasterizer_canvas_gles3.cpp
+++ b/drivers/gles3/rasterizer_canvas_gles3.cpp
@@ -849,7 +849,22 @@ void RasterizerCanvasGLES3::_record_item_commands(const Item *p_item, RID p_rend
 
 		if (c->type != Item::Command::TYPE_MESH) {
 			// For Meshes, this gets updated below.
-			_update_transform_2d_to_mat2x3(base_transform * draw_transform, state.instance_data_array[r_index].world);
+			Transform2D final_transform = base_transform * draw_transform;
+
+			// Make sure to round nearest filtered textures to avoid sprite "crunching".
+			// We don't round linear filtered textures because it would look identical to nearest filtering.
+			switch (p_item->texture_filter) {
+				case RS::CanvasItemTextureFilter::CANVAS_ITEM_TEXTURE_FILTER_NEAREST:
+				case RS::CanvasItemTextureFilter::CANVAS_ITEM_TEXTURE_FILTER_NEAREST_WITH_MIPMAPS:
+				case RS::CanvasItemTextureFilter::CANVAS_ITEM_TEXTURE_FILTER_NEAREST_WITH_MIPMAPS_ANISOTROPIC: {
+					final_transform.set_origin(final_transform.get_origin().round());
+				} break;
+				default: {
+					// Do nothing.
+				}
+			}
+
+			_update_transform_2d_to_mat2x3(final_transform, state.instance_data_array[r_index].world);
 		}
 
 		// Zero out most fields.
--- a/servers/rendering/renderer_rd/renderer_canvas_render_rd.cpp
+++ b/servers/rendering/renderer_rd/renderer_canvas_render_rd.cpp
@@ -425,6 +425,20 @@ void RendererCanvasRenderRD::_render_item(RD::DrawListID p_draw_list, RID p_rend
 
 	PushConstant push_constant;
 	Transform2D base_transform = p_canvas_transform_inverse * p_item->final_transform;
+
+	// Make sure to round nearest filtered textures to avoid sprite "crunching".
+	// We don't round linear filtered textures because it would look identical to nearest filtering.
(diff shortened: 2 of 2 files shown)


drivers/gles3/rasterizer_canvas_gles3.cpp
```diff
@@ -849,7 +849,11 @@ void RasterizerCanvasGLES3::_record_item_commands(const Item *p_item, RID p_rend
 
 		if (c->type != Item::Command::TYPE_MESH) {
 			// For Meshes, this gets updated below.
-			_update_transform_2d_to_mat2x3(base_transform * draw_transform, state.instance_data_array[r_index].world);
+			Transform2D final_transform = base_transform * draw_transform;
+			final_transform.set_origin(
+					Vector2(
+							Math::round(final_transform.get_origin().x), Math::round(final_transform.get_origin().y)));
```
AThousandShips, line 855:
```suggestion
			final_transform.set_origin(final_transform.get_origin().round());
```
Dedicated method for this 🙂

servers/rendering/renderer_rd/renderer_canvas_render_rd.cpp
```diff
@@ -425,6 +425,9 @@ void RendererCanvasRenderRD::_render_item(RD::DrawListID p_draw_list, RID p_rend
 
 	PushConstant push_constant;
 	Transform2D base_transform = p_canvas_transform_inverse * p_item->final_transform;
+	base_transform.set_origin(Vector2(
+			Math::round(base_transform.get_origin().x),
+			Math::round(base_transform.get_origin().y)));
```
AThousandShips, line 430:
```suggestion
	base_transform.set_origin(base_transform.get_origin().round());
```

Calinou:
Out of curiosity, why is this only done when the filter mode is Linear and not Nearest?

drivers/gles3/rasterizer_canvas_gles3.cpp
```diff
@@ -849,7 +849,18 @@ void RasterizerCanvasGLES3::_record_item_commands(const Item *p_item, RID p_rend
 
 		if (c->type != Item::Command::TYPE_MESH) {
 			// For Meshes, this gets updated below.
-			_update_transform_2d_to_mat2x3(base_transform * draw_transform, state.instance_data_array[r_index].world);
+			Transform2D final_transform = base_transform * draw_transform;
+			switch (p_item->texture_filter) {
```
AThousandShips, line 853: Given Calinou's question I'd say a comment here explaining this would be helpful 🙂

adamscott:
> Out of curiosity, why is this only done when the filter mode is Linear and not Nearest?

My bad, I just found out that I mixed up Linear and Nearest. I meant to do this on Nearest only.

Nearest:
[Capture vidéo du 2024-01-04 13-54-09.webm](https://github.com/godotengine/godot/assets/270928/87c939e9-b871-4948-bd7d-a9158d8b9c90)

Linear (I don't want to round up, because it will look like nearest if I round up (without the pixelated inbetweens):
[Capture vidéo du 2024-01-04 13-54-35.webm](https://github.com/godotengine/godot/assets/270928/6f394013-a1bd-40bb-9e9b-ebdd2283eb26)

clayjohn:
I think this should be tucked behind a check for ``rendering/2d/snap/snap_2d_transforms_to_pixel``. 

For performance reasons we should cache the result of course and add a method to enable pixel snapping inside the rendering server.

adamscott:
#84380 is better suited to fix https://github.com/godotengine/godot/issues/84632.

PR #86671, opened 2023-12-31
[X11] Don't re-set input focus if the given window already has it (fixes Godot stealing input focus on i3)

### LINKED ISSUE #86442: [X11] Godot (mostly) refusing to give up input focus when switching to external editor when using i3 window manager
### Tested versions

Reproducible in Godot `master` (9d1cbab1c432b6f1d66ec939445bec68b6af519e), Godot v4.2.1 and v4.1.3

(I've had this problem on all Godot 4.x versions since switching to i3 -- I used to use notion, another tiling window manager, but I don't recall if it had the same problem)

### System information

Ubuntu 22.04, X11 with i3 window manager

### Issue description

I have Godot configured to use VS Code as an external editor. When Godot opens VS Code, most of the time (like 1% of the time this _doesn't_ happen), it refuses to give up input focus, even though VS Code's title bar is highlighted in the way that usually indicates that it has input focus (and Godot's is greyed out in the way that usually indicates it doesn't have input focus).

So, if I press any keys, including shortcuts, they are received by Godot rather than VS Code. This happens when Godot and VS Code are on the same screen, or on different screens. (When on different screens, it's very easy to see that my keyboard input is really going to Godot, because I can see my actions happening there.)

In order to get VS Code to have focus, I need to first switch focus back to Godot, and then switch again to VS Code (either using the mouse, or the i3 keyboard shortcuts that I have configured). Sometimes I have to do this more than once, although, usually once will do it.

(Note: switching to single window mode makes no difference, although, I usually have the editor in single window mode because there's other _weirdnesses_ with i3 without

This seems sort of similar to these other X11 input focus issues:

- Issue https://github.com/godotengine/godot/issues/74378 with PR https://github.com/godotengine/godot/pull/86441
- Issue https://github.com/godotengine/godot/issues/68305 with PR https://github.com/godotengine/godot/pull/86101

However, it's not exactly the same, and, unfortunately, the PRs that fix those issues for others don't seem to fix my issue. :-/ So, I've decided to make a new issue here!

When I try using Gnome instead (on the same computer with the same Godot, VS Code and configurations of them both), this problem doesn't happen. When the external editor is opened by Godot in Gnome, it immediately has input focus and receives my key presses.

### Steps to reproduce

1. Start a session with the i3 window manager
2. Open a Godot project
3. Change your editor settings to use VS Code as an external editor
4. Click the "Open Script" icon next to a node in your scene tree
5. After VS Code opens and it's title bar is highlighted blue and Godot's title bar is greyed out (indicating the VS Code should have focus and Godot should not), press some keys and see that Godot is receiving them rather than VS Code (this is easiest to see with shortcuts like Ctrl+N)

n/a - this isn't project specific

dsnopek:
Fixes https://github.com/godotengine/godot/issues/86442

Currently, `DisplayServerX11` will set the input focus to a given window (via `XSetInputFocus()`) in a bunch of places. All this change does is checks to see if the given window already has input focus, and if so, skips setting input focus.

This fixes an issue I've been having in every version of Godot 4 with the i3 window manager when Godot launches the external editor (VS Code in my case), where input focus would visually seem to be on the external editor, but Godot would continue to receive all my key presses. (See in the issue linked above for a more detailed explanation.)

I don't know X11 very well, but what I think is happening, is Godot is re-setting the input focus to itself when it gets the mouse click to open the external editor, but that doesn't happen immediately, it happens sometime after VS Code has already gotten focus. So, shortly after VS Code gets focus, it gets yanked back to Godot.

In some ways, this is kind of the inverse of PR https://github.com/godotengine/godot/pull/86441, which some users of other alternative window managers say fixes similar problems for them (but not me).

### DIFF
--- a/platform/linuxbsd/x11/display_server_x11.cpp
+++ b/platform/linuxbsd/x11/display_server_x11.cpp
@@ -2011,7 +2011,7 @@ void DisplayServerX11::window_set_transient(WindowID p_window, WindowID p_parent
 		// a subwindow and its parent are both destroyed.
 		if (!wd_window.no_focus && !wd_window.is_popup && wd_window.focused) {
 			if ((xwa.map_state == IsViewable) && !wd_parent.no_focus && !wd_window.is_popup && _window_focus_check()) {
-				XSetInputFocus(x11_display, wd_parent.x11_window, RevertToPointerRoot, CurrentTime);
+				_set_input_focus(wd_parent.x11_window, RevertToPointerRoot);
 			}
 		}
 	} else {
@@ -2951,7 +2951,7 @@ void DisplayServerX11::window_set_ime_active(const bool p_active, WindowID p_win
 		XSync(x11_display, False);
 		XGetWindowAttributes(x11_display, wd.x11_xim_window, &xwa);
 		if (xwa.map_state == IsViewable && _window_focus_check()) {
-			XSetInputFocus(x11_display, wd.x11_xim_window, RevertToParent, CurrentTime);
+			_set_input_focus(wd.x11_xim_window, RevertToParent);
 		}
 		XSetICFocus(wd.xic);
 	} else {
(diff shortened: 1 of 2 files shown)


akien-mga:
Tested briefly on KDE / X11, didn't spot any unexpected behavior.

dsnopek:
I've rebased this one now that PR https://github.com/godotengine/godot/pull/86441 is merged, and re-tested it: I still have the issue without this PR, and it's still fixed with this PR.

It would be sooooo amazing if this could maybe make it into 4.3-dev3?

akien-mga (approved):
Let's give it a spin :)

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #86672, opened 2023-12-31
Fix incorrect VehicleWheel3D Roll Influence description

### LINKED ISSUE #85805: VehicleWheel3D wheel_roll_influence documentation is wrong
### Tested versions

godot 4.1 and 4.2 stable
both integrated 3D physics, and Jolt physics

### System information

MacOs

### Issue description

the description and tooltip for `VehicleWheel3D.wheel_roll_influence` currently states:
`This value affects the roll of your vehicle. If set to 1.0 for all wheels, your vehicle will be prone to rolling over, while a value of 0.0 will resist body roll.`
but my testing seems to indicate the opposite is the case: a higher value results in less body roll.

i am unsure if:
A) the tooltip describes the desired behaviour, and the code creates a wrong result, or
B) the code is correct, and the tooltip is wrong.
the way physics work, i would assume that wheel_roll_influence is a multiplier of the gyroscopic force stabilizing a wheel when rotating, so my guess would be B). but since i cannot find the code handling this, i cannot confirm if this is it.

either way, it is a very confusing issue to have when starting to create vehicles in godot.

### Steps to reproduce

create a basic vehicle, then set the `wheel_roll_influence` to different values, and try out how the car handles.
it seems a low value makes the vehicle prone to rolling over, while a high value makes the vehicle more stable, and less likely to roll over.

see the minimal reproduction project:
[vehicleBodyTest 2.zip](https://github.com/godotengine/godot/files/13577619/vehicleBodyTest.2.zip)
[video of MRP](https://youtu.be/lP00s9nDl4Q)
- it includes a basic VehicleBody3D, with simple `arrow key` controls.
- the `wheel_roll_influence` can be changed by pressing the keys `1` and `2`.
- the `R` key will reset the car to the start.

qwofford:
Updated VehicleWheel3D Roll influence description and confirmed the tooltip has been updated in the Editor Inspection window.

I updated the `es.po` translation file, but additional translations need to be corrected. I updated english text for `fr.po`, `zh_CN.po` and `zh_TW.po` in this PR, but I need to find someone to update the alternative text for each of these files.

This PR fixes #85805.

### DIFF
--- a/doc/classes/VehicleWheel3D.xml
+++ b/doc/classes/VehicleWheel3D.xml
@@ -81,7 +81,7 @@
 			This is the distance in meters the wheel is lowered from its origin point. Don't set this to 0.0 and move the wheel into position, instead move the origin point of your wheel (the gizmo in Godot) to the position the wheel will take when bottoming out, then use the rest length to move the wheel down to the position it should be in when the car is in rest.
 		</member>
 		<member name="wheel_roll_influence" type="float" setter="set_roll_influence" getter="get_roll_influence" default="0.1">
-			This value affects the roll of your vehicle. If set to 1.0 for all wheels, your vehicle will be prone to rolling over, while a value of 0.0 will resist body roll.
+			This value affects the roll of your vehicle. If set to 1.0 for all wheels, your vehicle will resist body roll, while a value of 0.0 will be prone to rolling over.
 		</member>
 	</members>
 </class>


AThousandShips:
The translations are still present, and please squash your commits into one 🙂, see [here](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#the-interactive-rebase)

AThousandShips (commented):
Other than those changes this looks good, based on indications from testing by people on the issue report and statements there this should be correct

qwofford:
Do I understand correctly that you want me to remove all translation changes from this pull request, such that the impacted lines in `*.po` remain as they are on `master` right now?

What is the appropriate approach to update translations?

AThousandShips:
Yes, indeed, the updates to the translation is as I said in the issue, you edit them on the webplate, see [here](https://docs.godotengine.org/en/stable/contributing/documentation/editor_and_docs_localization.html)

AThousandShips (approved):
Looks good to me, as per above

~Please squash your commits into one, see [here](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#the-interactive-rebase)~

akien-mga:

YuriSizov:
Cherry-picked for 4.2.2.

YuriSizov:
Cherry-picked for 4.1.4.

PR #86673, opened 2023-12-31
Rewrite of the floating-point number parsing function

odriansu:
Addresses a bug  that led to incorrect parsing when exponent digits were not explicitly set. For instance, it resulted in 9991 instead of 99.91 when the string was "99.91e"

Moreover, it resolves the issue where an exponent less than e-308  led to a result of 0. For instance, "1e-309" was producing 0,  the correct output should be "1.00000000000000189e-309."

### DIFF
--- a/core/string/ustring.cpp
+++ b/core/string/ustring.cpp
@@ -2460,219 +2460,163 @@ bool String::is_numeric() const {
 }
 
 template <class C>
-static double built_in_strtod(
-		/* A decimal ASCII floating-point number,
-		 * optionally preceded by white space. Must
-		 * have form "-I.FE-X", where I is the integer
-		 * part of the mantissa, F is the fractional
-		 * part of the mantissa, and X is the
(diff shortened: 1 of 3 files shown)


kleonc:
Note there are some tests for this method, if you could please inspect these / add new relevant ones:
https://github.com/godotengine/godot/blob/13a0d6e9b253654f5cc2a44f3d0b3cae10440443/tests/core/string/test_string.h#L516-L546

peteroupc:
I am aware of additional floating-point number parsing tests at the following repository: https://github.com/nigeltao/parse-number-fxx-test-data

odriansu:
With the tests, another error was identified where an exponent exceeds range; int 32 bits [−2,147,483,647, +2,147,483,647]. This could lead to overflow and produce incorrect results. For instance, 1e-4147483648 results in infinity instead of zero, and 1e-4147483648 produces zero when it should be infinity.

PR #86677, opened 2024-01-01
Implement overdraw, lighting, and unshaded debug draw modes for OpenGL

### LINKED ISSUE #82485: OpenGL: Debug draw modes are not working
### Godot version

4.1.1 Macos

### System information

M1 Macbook Air big sur

### Issue description

Wireframe mode does not work in 4.1.1 WHEN COMPATIBILITY RENDERER IS ENABLED
image
this is godot 3.5 ^
image
this is godot 4.1.1 ^

### Steps to reproduce

have a 3d mesh

set perspective to display wireframe

enable the compatibility renderer

be on mac (?)

N/A

daustria:
Debug lighting and unshaded should work as expected. For unshaded specifically it seems like most of the work was already there and just needed to be toggled by setting the right variant.

For overdraw mode, using the same shader as the one used in the forward renderer results in faded colours, probably due to doing blending in srgb vs linear space. one way to compensate for this is to set a higher alpha which is done in this implementation. I attached some pictures of what this looks like, although it doest quite match the forward renderer. I chose to set ALPHA=0.2 instead of ALPHA=0.1 as in forward+ (i wanted to match the colour of forward+, setting ALPHA=0.3 makes the objects look too blue compared to forward+)

overdraw imgaes

_bugsquad edit: Fixes https://github.com/godotengine/godot/issues/82485_

### DIFF
--- a/drivers/gles3/rasterizer_scene_gles3.cpp
+++ b/drivers/gles3/rasterizer_scene_gles3.cpp
@@ -1485,7 +1485,15 @@ void RasterizerSceneGLES3::_setup_environment(const RenderDataGLES3 *p_render_da
 	//time global variables
 	scene_state.ubo.time = time;
 
-	if (is_environment(p_render_data->environment)) {
+	if (get_debug_draw_mode() == RS::VIEWPORT_DEBUG_DRAW_UNSHADED) {
+		scene_state.ubo.use_ambient_light = true;
+		scene_state.ubo.ambient_light_color_energy[0] = 1;
+		scene_state.ubo.ambient_light_color_energy[1] = 1;
+		scene_state.ubo.ambient_light_color_energy[2] = 1;
+		scene_state.ubo.ambient_light_color_energy[3] = 1.0;
+		scene_state.ubo.use_ambient_cubemap = false;
+		scene_state.ubo.use_reflection_cubemap = false;
+	} else if (is_environment(p_render_data->environment)) {
 		RS::EnvironmentBG env_bg = environment_get_background(p_render_data->environment);
 		RS::EnvironmentAmbientSource ambient_src = environment_get_ambient_source(p_render_data->environment);
 
@@ -2324,7 +2332,7 @@ void RasterizerSceneGLES3::render_scene(const Ref<RenderSceneBuffers> &p_render_
 	bool keep_color = false;
 	float sky_energy_multiplier = 1.0;
 
-	if (get_debug_draw_mode() == RS::VIEWPORT_DEBUG_DRAW_OVERDRAW) {
+	if (unlikely(get_debug_draw_mode() == RS::VIEWPORT_DEBUG_DRAW_OVERDRAW)) {
 		clear_color = Color(0, 0, 0, 1); //in overdraw mode, BG should always be black
 	} else if (render_data.environment.is_valid()) {
 		RS::EnvironmentBG bg_mode = environment_get_background(render_data.environment);
@@ -2693,6 +2701,7 @@ void RasterizerSceneGLES3::_render_list_template(RenderListParameters *p_params,
 			}
 			glBindTexture(GL_TEXTURE_CUBE_MAP, texture_to_bind);
 		}
+
 	} else if constexpr (p_pass_mode == PASS_MODE_DEPTH || p_pass_mode == PASS_MODE_SHADOW) {
 		shader_variant = SceneShaderGLES3::MODE_DEPTH;
 	}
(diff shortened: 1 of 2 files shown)


drivers/gles3/rasterizer_scene_gles3.cpp
```diff
@@ -3638,6 +3650,29 @@ void fragment() {
 		scene_globals.default_material = material_storage->material_allocate();
 		material_storage->material_initialize(scene_globals.default_material);
 		material_storage->material_set_shader(scene_globals.default_material, scene_globals.default_shader);
+		default_material_data_ptr = static_cast<GLES3::SceneMaterialData *>(GLES3::MaterialStorage::get_singleton()->material_get_data(scene_globals.default_material, RS::SHADER_SPATIAL));
+	}
+
+	{
+		// Overdraw material and shader.
+		scene_globals.overdraw_shader = material_storage->shader_allocate();
```
AThousandShips, line 3658: You need to free these in the destructor

drivers/gles3/rasterizer_scene_gles3.cpp
```diff
@@ -3752,6 +3787,10 @@ RasterizerSceneGLES3::~RasterizerSceneGLES3() {
 	RSG::material_storage->material_free(scene_globals.default_material);
 	RSG::material_storage->shader_free(scene_globals.default_shader);
 
+	// Overdraw Shader
+	RSG::material_storage->shader_free(scene_globals.overdraw_material);
```
AThousandShips, line 3791:
```suggestion
	RSG::material_storage->material_free(scene_globals.overdraw_material);
```

daustria: wow, seems like nothing slips by you and these automated checks

fire:
@lyuma do you think we should have an engine implementation of the gamma conversion hack for compatibility/opengl?

daustria:
> @lyuma do you think we should have an engine implementation of the gamma conversion hack for compatibility/opengl?
that sounds interesting, is there a link or file i can see to understand what hack you are talking about?

drivers/gles3/rasterizer_scene_gles3.cpp
```diff
@@ -2693,6 +2693,10 @@ void RasterizerSceneGLES3::_render_list_template(RenderListParameters *p_params,
 			}
 			glBindTexture(GL_TEXTURE_CUBE_MAP, texture_to_bind);
 		}
+
+		if (get_debug_draw_mode() == RS::VIEWPORT_DEBUG_DRAW_UNSHADED) {
+			shader_variant = SceneShaderGLES3::MODE_UNSHADED;
+		}
```
clayjohn, line 2699: We shouldn't be using shader variants for this. We should just do what we do in the RD renderers and avoid passing in any lighting information. We already avoid lights and reflectionprobes. So all that is left is making sure there is no ambient lighting from the sky:
https://github.com/godotengine/godot/blob/89cc635c0554cb2e518c830969ca4c5eedda0f4e/servers/rendering/renderer_rd/storage_rd/render_scene_data_rd.cpp#L116-L124

We need to avoid shader variants as they are always compiled. The change you made in scene.glsl would increase shader compilation time by 50% for all scene shaders (even if this debug draw mode is never used)

drivers/gles3/rasterizer_scene_gles3.cpp
```diff
@@ -2730,8 +2734,16 @@ void RasterizerSceneGLES3::_render_list_template(RenderListParameters *p_params,
 			material_data = surf->material_shadow;
 			mesh_surface = surf->surface_shadow;
 		} else {
-			shader = surf->shader;
-			material_data = surf->material;
+			if (get_debug_draw_mode() == RS::VIEWPORT_DEBUG_DRAW_OVERDRAW) {
+				material_data = overdraw_material_data_ptr;
+				shader = material_data->shader_data;
+			} else if (get_debug_draw_mode() == RS::VIEWPORT_DEBUG_DRAW_LIGHTING) {
```
clayjohn, line 2740:
```suggestion
			} else if (unlikely(get_debug_draw_mode() == RS::VIEWPORT_DEBUG_DRAW_LIGHTING)) {
```
Needs to be wrapped in `unlikely()` to help with performance

drivers/gles3/rasterizer_scene_gles3.cpp
```diff
@@ -2730,8 +2734,16 @@ void RasterizerSceneGLES3::_render_list_template(RenderListParameters *p_params,
 			material_data = surf->material_shadow;
 			mesh_surface = surf->surface_shadow;
 		} else {
-			shader = surf->shader;
-			material_data = surf->material;
+			if (get_debug_draw_mode() == RS::VIEWPORT_DEBUG_DRAW_OVERDRAW) {
```
clayjohn, line 2737:
```suggestion
			if (unlikely(get_debug_draw_mode() == RS::VIEWPORT_DEBUG_DRAW_OVERDRAW)) {
```

clayjohn (commented):
Looks great so far. I left a few comments that will need to be addressed. But it shouldn't be too much work. Please let me know if you have any questions!

drivers/gles3/rasterizer_scene_gles3.cpp
daustria: just wondering, what does this do? tried following the macro expansion on my IDE, and it seems like it just expands to the same expression

clayjohn: Its a hint to the compiler to tell it that the if statement is most likely to be false. This allows it to emit some special assembly that generally makes the statement faster when it fails.

Without the hint the compiler will make a guess about how often the branch succeeds. If the compiler guesses wrong then it will be slightly slower

clayjohn (approved):
Looks great to me now! Good job!

YuriSizov:
Thanks! And congrats on your first merged Godot contribution!

akien-mga:
Cherry-picked for 4.2.2.

Issue #86489, opened 2023-12-24
Race condition in GradientTexture2D.GetImage


Renari:
### Tested versions

4.2.1.stable.mono

### System information

Godot v4.2.1.stable.mono - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3080 Ti (NVIDIA; 31.0.15.4633) - Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz (16 Threads)

### Issue description

When calling `GradientTexture2D.GetImage`, it seems it can return null, this is currently not documented as being possible.

### Steps to reproduce

```csharp
using Godot;
using System.Threading;

public partial class GenerateGradient : Node2D
{
	[Export] public int MapSize = 1024;
	
	// Called when the node enters the scene tree for the first time.
	public override void _Ready()
	{
		new Thread(GenerateMap).Start();
	}
	
	private void GenerateMap()
	{
		var gradientTexture = new GradientTexture2D
		{
			Width = MapSize,
			Height = MapSize,
			Fill = GradientTexture2D.FillEnum.Radial,
			FillFrom = new Vector2(0.5f, 0.5f),
			FillTo = new Vector2(0, 0.5f),
			Gradient = new Gradient()
		};
		Image gradientImage;
		do
		{
			GD.Print("pass");
			Thread.Sleep(1);
			gradientImage = gradientTexture.GetImage();
		} while (gradientImage == null);
	}
}
```
In the above example `pass` can be printed out multiple times, based on the documentation I would expect this to only ever be printed once.

### Minimal reproduction project (MRP)

[GetTexture2DImage.zip](https://github.com/godotengine/godot/files/13761227/GetTexture2DImage.zip)

jsjtxietian:
From the code that it can return a null when the texture is invalid. Should we update the document for this?

And it doesn't seem very related to the threading (maybe it's unsafe to call `texture_2d_get` in another thread? Idk )
```
Ref<Image> GradientTexture2D::get_image() const {
	const_cast<GradientTexture2D *>(this)->update_now();
	if (!texture.is_valid()) {
		return Ref<Image>();
	}
	return RenderingServer::get_singleton()->texture_2d_get(texture);
}
```

PR #86685, opened 2024-01-01
Update `get_image` doc to mention that it will return an empty image with invalid texture

jsjtxietian:
Related to https://github.com/godotengine/godot/issues/86489

Unfortunately I can't reproduce that issue with a similiar gdscript version of that MRP, but I think it's better to mention this behaviour. 

I checked all the virtual `get_image` implementation, looks like it's appropriate to add this note (`NoiseTexture2D`'s `get_image` will directly return the image stored in member, but I think it's fine).

### DIFF
--- a/doc/classes/Texture2D.xml
+++ b/doc/classes/Texture2D.xml
@@ -123,6 +123,7 @@
 			<return type="Image" />
 			<description>
 				Returns an [Image] that is a copy of data from this [Texture2D] (a new [Image] is created each time). [Image]s can be accessed and manipulated directly.
+				[b]Note:[/b] This will return [code]null[/code] if this [Texture2D] is invalid.
 				[b]Note:[/b] This will fetch the texture data from the GPU, which might cause performance problems when overused.
 			</description>
 		</method>


doc/classes/Texture2D.xml
```diff
@@ -123,7 +123,7 @@
 			<return type="Image" />
 			<description>
 				Returns an [Image] that is a copy of data from this [Texture2D] (a new [Image] is created each time). [Image]s can be accessed and manipulated directly.
-				[b]Note:[/b] This will fetch the texture data from the GPU, which might cause performance problems when overused.
+				[b]Note:[/b] It will return an empty image if this [Texture2D] is invalid. This will fetch the texture data from the GPU, which might cause performance problems when overused.
```
AThousandShips, line 126:
```suggestion
				[b]Note:[/b] Thid will return an empty image if this [Texture2D] is invalid.
				[b]Note:[/b] This will fetch the texture data from the GPU, which might cause performance problems when overused.
```
Generally afaik we use this wording, and I'll make this two notes as they aren't related

AThousandShips (approved):
Looks good and correct otherwise! Thank you! And happy new year!

doc/classes/Texture2D.xml
jsjtxietian: Thanks!

jsjtxietian:
> And happy new year!

You too! In the new year I hope I can "bother" you with more PRs to Godot，thanks for your hard work!

doc/classes/Texture2D.xml
jsjtxietian: It returns a `Ref<Image>()`, can this be considered as null ? print it will show a `"<Object#null>"` , it does have a type lol.

AThousandShips:
Looking forward to it 🙂

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

PR #86689, opened 2024-01-01
2D shadows respect z_index

### LINKED ISSUE #64939: CanvasItem light_mask is ignored by LightOccluder2D occluder_light_mask
### Godot version

v4.0.alpha14.official [106b68050]

### System information

Ubuntu, Nvidia

### Issue description

By setting CanvasItem light_mask to a different layer than LightOccluder2D light_mask, it was possible to have shadows not being drawn on top of sprites. This does no longer work.

Godot 3:

image

Identical setup in Godot 4:

image

Especially in topdown games you generally need this to work.

### Steps to reproduce

1. Set your PointLight2D's range_item_cull_mask to cover layer 1 and 2
2. Set your Sprite2D's light_mask to 2 only
3. Add LightOccluder2D that remains on default for occluder_light_mask and light_mask (layer 1)
4. have the LightOccluder2D polygon cover (part of) the sprite
5. hit F6 and note that the shadow is rendered on top of the sprite. the shadow should not be rendered on top of the sprite as is the case in Godot 3.

projects for both godot 3 and 4 to compare:

[light-occluder-sprite-bug.zip](https://github.com/godotengine/godot/files/9435066/light-occluder-sprite-bug.zip)

### LINKED ISSUE #74265: Shadows cover sprites when set to show behind sprite
### Godot version

4.0.stable.official

### System information

Linux (KDE Neon 5.26), X11, AMD Vega64

### Issue description

Shadows via LightOccluder2D are cast over top sprites, regardless of order.

Screenshot_20230302_133948
Screenshot_20230302_134001

### Steps to reproduce

1. Create a Sprite2D with a CanvasTexture containing a diffuse layer and normal map
2. Add a LightOccluder2D as a child of the sprite, set CanvasItem > Show Behind Parent to true
3. Run with a light source nearby

Expected result: Shadow should not cover the parent sprite
Actual result: Shadow covers the parent sprite

[gd-shadow-bug.zip](https://github.com/godotengine/godot/files/10877235/gd-shadow-bug.zip)

DanielKauss:
This PR adds support to `LightOccluder2D`s for the z index, fixes #64939 and fixes #74265. All `canvasItem`s will now receive shadows only from `LightOccluder2D`s that have a higher z index. This is achieved by passing the z index as the integer part of the shadow texture, and using a different depth check. The new depth check first compares the shadow and object z index to determine if a shadow should be cast, and later performs the normal check with only the fractional depth part.

`DirectionalLight2D`s maintain their old behavior since their depth maps can have values above or below 1 and 0, which would mess up the new depth check. The z index used by occluders is absolute, since it is not known in the lightning pass. This change also only applies to the `forward+` renderer, since the size of the `canvas.glsl` shader's push constant has been increased to 144, which some low end devices don't support. The only change that affects performance is the removal of `sampler2DShadow` in favor of a custom sampler, but the effect should be minimal. 

Here is a comparison, left being old and right new. Instead of all the shadows being overlayed on top, they now respect the layering. 

image

Please let me know if this is acceptable to merge and if I should do any changes, this is my first PR here. However, I believe that this is a crucial feature, as 2D lighting is currently almost unusable for me.


servers/rendering_server.h
```diff
@@ -1480,6 +1480,7 @@ class RenderingServer : public Object {
 	virtual void canvas_light_occluder_set_as_sdf_collision(RID p_occluder, bool p_enable) = 0;
 	virtual void canvas_light_occluder_set_transform(RID p_occluder, const Transform2D &p_xform) = 0;
 	virtual void canvas_light_occluder_set_light_mask(RID p_occluder, int p_mask) = 0;
+	virtual void canvas_light_occluder_set_z_index(RID p_occluder, int z_index) = 0;
```
AThousandShips, line 1483:
```suggestion
	virtual void canvas_light_occluder_set_z_index(RID p_occluder, int p_z_index) = 0;
```
Prefix arguments with `p_`

servers/rendering/renderer_canvas_cull.h
```diff
@@ -302,6 +302,7 @@ class RendererCanvasCull {
 	void canvas_light_occluder_set_as_sdf_collision(RID p_occluder, bool p_enable);
 	void canvas_light_occluder_set_transform(RID p_occluder, const Transform2D &p_xform);
 	void canvas_light_occluder_set_light_mask(RID p_occluder, int p_mask);
+	void canvas_light_occluder_set_z_index(RID p_occluder, int z_index);
```
AThousandShips, line 305:
```suggestion
	void canvas_light_occluder_set_z_index(RID p_occluder, int p_z_index);
```

servers/rendering/renderer_canvas_cull.cpp
```diff
@@ -1923,6 +1923,12 @@ void RendererCanvasCull::canvas_light_occluder_set_light_mask(RID p_occluder, in
 
 	occluder->light_mask = p_mask;
 }
+void RendererCanvasCull::canvas_light_occluder_set_z_index(RID p_occluder, int z_index) {
```
AThousandShips, line 1926:
```suggestion

void RendererCanvasCull::canvas_light_occluder_set_z_index(RID p_occluder, int p_z_index) {
```
Leave a line between methods

scene/main/canvas_item.h
```diff
@@ -176,6 +176,7 @@ class CanvasItem : public Node {
 		NOTIFICATION_EXIT_CANVAS = 33,
 		NOTIFICATION_LOCAL_TRANSFORM_CHANGED = 35,
 		NOTIFICATION_WORLD_2D_CHANGED = 36,
+		NOTIFICATION_Z_INDEX_CHANGED = 37,
```
AThousandShips, line 179: You need to document this, and register it in the `_bind_methods` method

scene/2d/light_occluder_2d.cpp
```diff
@@ -166,6 +166,10 @@ void LightOccluder2D::_notification(int p_what) {
 			RS::get_singleton()->canvas_light_occluder_set_enabled(occluder, is_visible_in_tree());
 		} break;
 
+		case NOTIFICATION_Z_INDEX_CHANGED : {
```
AThousandShips, line 169:
```suggestion
		case NOTIFICATION_Z_INDEX_CHANGED: {
```

servers/rendering/renderer_rd/shaders/canvas.glsl
```diff
@@ -369,8 +369,71 @@ vec3 light_normal_compute(vec3 light_vec, vec3 normal, vec3 base_color, vec3 lig
...
+	}
+
+	vec4 shadow_color = unpackUnorm4x8(light_array.data[light_base].shadow_color);
+#ifdef LIGHT_CODE_USED
+	shadow_color.rgb *= shadow_modulate;
+#endif
+
+	shadow_color.a *= light_color.a; //respect light alpha
+
+
```
AThousandShips, line 429:
```suggestion
```

servers/rendering/renderer_rd/shaders/canvas.glsl
```diff
@@ -369,8 +369,71 @@ vec3 light_normal_compute(vec3 light_vec, vec3 normal, vec3 base_color, vec3 lig
...
+	//return vec4(shadow_uv.zzz / 2.0, 1.0);
+	//return vec4(shadow, shadow, fract(shadow_uv.z / 4.0), 1.0);
+	return mix(light_color, shadow_color, shadow);
+}
+
+
```
AThousandShips, line 435:
```suggestion
```

doc/classes/CanvasItem.xml
```diff
@@ -656,6 +656,9 @@
 		<constant name="NOTIFICATION_WORLD_2D_CHANGED" value="36">
 			The [CanvasItem]'s active [World2D] changed.
 		</constant>
+		<constant name="NOTIFICATION_Z_INDEX_CHANGED" value="36">
```
AThousandShips, line 659:
```suggestion
		<constant name="NOTIFICATION_Z_INDEX_CHANGED" value="37">
```

clayjohn:
I agree that this seems to be an valuable improvement the behaviour of light occluders. However, the way it is implemented right now won't work on mobile devices when using the Vulkan backend and it appears to not be implemented for the compatibility backend at all. 

Unfortunately, lack of mobile support is a huge problem for a PR like this, so we will need to figure out a way to implement the feature for all platforms in order to move forward. 

We typically handle culling of lights and shadows on the CPU side, so perhaps we can get away without doing the culling on the GPU.

DanielKauss:
> Unfortunately, lack of mobile support is a huge problem for a PR like this, so we will need to figure out a way to implement the feature for all platforms in order to move forward.

Makes sense. I'll see if I can add support for the other rendering engines as well. I didn't know if the approach I chose is even acceptable in the first place, since it touches the renderer shader directly, which is why I only added it to forward+ first. If it's fine I will implement it with the other back-ends as well, so just tell me.

> We typically handle culling of lights and shadows on the CPU side, so perhaps we can get away without doing the culling on the GPU.

The problem is the way that light is rendered. The shadow map gets calculated at the very end of the renderer pass, and then applied in the shader. This means that all the lights and shadows get applied at once at the end, which doesn't allow them to be rendered in order like all the normal objects.

So as far as I can see, this fix seems like a pretty good solution that doesn't involve changing the entire renderer. I could be completely wrong about this though, I've never really worked with rendering code before.

DanielKauss:
Sorry for leaving this PR dead for so long, I haven't really used Godot for a while. Recently I wanted to use it again and saw that the light issue still existed, so I came back to this and wrote a different PR with somewhat better code. It uses an entirely different approach and all of the code is new which is why I will close this one.

PR #84174, opened 2023-10-30
Hide CSGShape's `debug_collision_shape` when it is invisible

jsjtxietian:
Fixes #84132

Issue #84132, opened 2023-10-29
CSG collision shape editor gizmo does not hide


Wolfe2x7:
### Godot version

v4.1.3.rc1.official [f80c673cd]

### System information

Godot v4.1.3.rc1 - Linux Mint 20.2 (Uma) - X11 - Vulkan (Mobile) - dedicated NVIDIA GeForce GTX 1660 Ti (nvidia; 470.199.02) - Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (12 Threads)

### Issue description

When any CSG node has collisions enabled, the editor gizmo for its collision shape does not respect its visibility, potentially obscuring other nodes unless collisions are temporarily disabled for CSG nodes.

### Steps to reproduce

1. Enabled collisions on a CSG node
2. Toggle its visibility

### Minimal reproduction project

[min-repo-10-28-23-1.zip](https://github.com/godotengine/godot/files/13197016/min-repo-10-28-23-1.zip)

PR #86699, opened 2024-01-02
Fix `CSGShape3D` debug collision shapes being visible in editor

### LINKED ISSUE #87919: Collision wireframes of CSGBox3D doesn't hide in the scene view when gizmos are hidden
### Tested versions

Reproducible in Godot v4.2.1.stable and v4.3 custom build

### System information

Godot v4.2.1.stable - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2070 (NVIDIA; 31.0.15.4633) - AMD Ryzen 7 5800X 8-Core Processor (16 Threads)

### Issue description

When unchecking View Gizmos in the Perspective menu in the Editor 3D Scene collision wireframes for CSG nodes doesn't get hidden. This is inconsistent with for example CollisionShape3D's, which properly gets hidden.

image

**Hiding the node**
Hiding the node makes the collider disappear, but if you open another scene and then return the render mesh is hidden but the collision mesh is again showing.

image

**Why is this an issue?**
This gets visually noisy and makes tuning the look of my game difficult.

image

### Steps to reproduce

1. Create a 3D Scene
2. Add a CSGBox3D and turn Collisions On
3. Toggle "View Gizmos" in the Perspective menu in the Scene View
4. Observe how CSG Collision Wireframes don't get hidden like they would with CollisionShape3D

[csg-wireframe-issue-reproduction-project.zip](https://github.com/godotengine/godot/files/14154642/csg-wireframe-issue-reproduction-project.zip)

### LINKED ISSUE #96767: `CSGShape3D` debug collision visualization disappears when node is made invisible
### Tested versions

Reproducible in 4.3.stable

### System information

Godot v4.3.stable - Windows 10.0.22631 - GLES3 (Compatibility) - NVIDIA GeForce GTX 1660 SUPER (NVIDIA; 32.0.15.6081) - Intel(R) Core(TM) i7-10700F CPU @ 2.90GHz (16 Threads)

### Issue description

When a CSG node is made invisible while the game is running with `Visible Collision Shapes` debug feature enabled, the collision shape visualization disappears as well. Expected behaviour is for the visual mesh to disappear, but the collision visualization should remain visible.
DebugCollisionVisualizationVisibility

### Steps to reproduce

- Open `test.tscn` in attached project.
- Enable `Visible Collision Shapes`
- Run the game
- Use the visibility toggle button
- Observe that the collision geometry disappears for the CSG, but remains visible for the RigidBody cube.

[bug-report.zip](https://github.com/user-attachments/files/16935345/bug-report.zip)

MajorMcDoom:
`CSGShape3D` is improperly co-opting the debug collision visualization system to render editor gizmos. As such, the gizmos do not respect any gizmo visibility settings (see #87919 ).

The only way to control the visibility of the collision gizmos currently is to make the `CSGShape3D` itself invisible (workaround comes from #84174). This is incorrect because an invisible `CSGShape3D` can still have collision, so the gizmo should still be visible. This also affects runtime debug collision visualization (see #96767).

This PR makes the debug collision geometry only visible in the running game, essentially removing their usage as gizmos.
- Fixes #87919 
- Fixes #96767 
- Represents a small step toward a proper gizmo solution by removing broken bandaid solutions and setting a clean slate.

### DIFF
--- a/modules/csg/csg_shape.cpp
+++ b/modules/csg/csg_shape.cpp
@@ -513,7 +513,7 @@ Ref<ConcavePolygonShape3D> CSGShape3D::bake_collision_shape() {
 }
 
 bool CSGShape3D::_is_debug_collision_shape_visible() {
-	return is_inside_tree() && (get_tree()->is_debugging_collisions_hint() || Engine::get_singleton()->is_editor_hint());
+	return !Engine::get_singleton()->is_editor_hint() && is_inside_tree() && get_tree()->is_debugging_collisions_hint();
 }
 
 void CSGShape3D::_update_debug_collision_shape() {
@@ -603,11 +603,6 @@ void CSGShape3D::_notification(int p_what) {
 			if (!is_root_shape() && last_visible != is_visible()) {
 				// Update this node's parent only if its own visibility has changed, not the visibility of parent nodes
 				parent_shape->_make_dirty();
-			}
-			if (is_visible()) {
-				_update_debug_collision_shape();
-			} else {
-				_clear_debug_collision_shape();
 			}
 			last_visible = is_visible();
 		} break;


MajorMcDoom:
UPDATE:

This issue was previously reported, and was unfortunately misdiagnosed and led to an incorrect solution that introduces another problem that will have to be corrected (perhaps as part of this pull request):
https://github.com/godotengine/godot/issues/84132
https://github.com/godotengine/godot/pull/84174

The issue was misdiagnosed as being a _gizmo_ that fails to hide. This is not the case - it is runtime debug collision shape visualization, not an editor gizmo.

The subsequent merge made the debug collision shape visibility dependent on the visibility of the CSG node itself. While this masked the problem of the previously unhideable wireframe, it leads to incorrect behaviour, because as the documentation for the "use_collision" property states: making a CSG invisible _does not_ disable the collision. This means the debug collision shape visualization should remain visible _regardless_ of the visibility of the CSG.

Because this new pull request will completely remove the debug collision shapes from the editor, the visibility dependency introduced by #84174 can be reverted. I will do this in a commit amend tomorrow.

MajorMcDoom:
Aforementioned changes from #84174 have been reverted.

fire:
Let us know if https://github.com/godotengine/godot/pull/84174 fixes your bug.

MajorMcDoom:
> Let us know if #84174 fixes your bug.

See my two followup comments. That's a merge which actually misdiagnosed the problem and introduced another issue, which this merge would also fix.

BastiaanOlij:
@MajorMcDoom it's worth adding screenshots to PRs like this to make clear what the change is.

I haven't followed the previous discussions, but this is an interesting one.

Traditionally Godot has always had the visual and physics part of an object separate. Collision shapes are drawn in editor as well. You can see this clearly when you move the collision shape away from the object:
image
You can hide the collision shapes however if they are in the way, something that doesn't effect physics for all the right reasons, but does catch people off guard.

CSG breaks the mould. It's a single node that handles both the visual and the physics aspect of the object. Showing the collision shape superimposed over the mesh is in a way expected behaviour, you can tell immediately whether you've forgotten to enable collisions or if they are indeed enabled.

However, if you're not interested there is no way to hide them, and especially for complex extruded shapes, it can be a lot of visual clutter not to mention that it slows the IDE down. I use CSGPolygon a lot to create stuff like race tracks, often the track itself will be several CSGPolygon shapes all following the same path but without collisions just to bring in detail, and a separate CSGPolygon shape that is used to create the collision shape. It gets pretty gnarly at times :)

While I find it important to be able to see the collision shape and would not want that removed, I'd love to hide it once I'm happy with the composition of the scene.

MajorMcDoom:
@BastiaanOlij Thanks for the response!

I'm not sure if there's been a miscommunication somewhere, but I'm aware that CSG is a "hybrid" between visual and collision.

The problem is:

- CSG is currently using debug collision shapes (as in the feature you use to see collision shapes in the running game) as a gizmo.
- Because it isn't using gizmos, it of course breaks all the visibility rules that gizmos have, hence not being hideable in the editor by any means.
- This was previously reported in #84132, and was improperly "fixed" by #84174 by making the debug collision shapes (that are not supposed to run in the editor in the first place) hidden whenever the CSG itself is hidden.
- #84174 introduces an additional issue of CSG invisibility now misleading the user into thinking the collision has been turned off, when it fact it has not. This is explicitly mentioned in the documentation for CSGs.


The debug collision shape is still important to have for CSG *at runtime*, because the feature should work for all collision shapes. That is why this PR still leaves that intact.

Whether or not a CSG should show _collision gizmos_ is a discussion beyond the scope of this PR. The fact is, however, that they are currently _not_ showing gizmos - they are improperly using debug collision shapes _in place_ of gizmos.

BastiaanOlij:
> Whether or not a CSG should show _collision gizmos_ is a discussion beyond the scope of this PR. The fact is, however, that they are currently _not_ showing gizmos - they are improperly using debug collision shapes _in place_ of gizmos.

Ah, ok, I indeed missed that nuance. Sounds like a shortcut was taken to make something draw. I'd have to take a closer look to give more sensible feedback :)

BastiaanOlij:
Ok, as best as I can judge there are a few incorrect assumptions here which means I think a little more work needs to be done here.

When we look at normal collision shapes it is correct that the debug collision shape is drawn at runtime when this is enabled to help find issues with the collision setup in ones game.
However the same shape is also drawn in the IDE, as a transparent shape when the node is selected and as a grey wireframe when the node is not selected, to visualise the collision shapes in the scene. However you can turn this off in the view settings as this can be visually cluttering.

When we look at CSG the problem we have is that we have dual functionality here. The main purpose of CSG is the mesh that is generated and the gizmo controls focus on the manipulation of this mesh. The collision shape that can be added is "extra" but we lack control over whether we want to visualise this shape in the IDE.

This PR is born out of an understandable annoyance that this shape is always drawn, and drawn fully visible. It thus relegates this to the original functionality of just showing the shape during runtime if collision shapes are enabled in the debug menu.

There is however value in seeing the collision shape in the IDE at times and taking that away will frustrate users who rely on seeing that shape.
What is needed is similar functionality as exists on the collision shape node itself. E.g. it should be drawn less obvious when the CSG node is not selected, and you should be able to toggle this on or off.
If that can be added to this PR, I think this would be good functionality to have.

MajorMcDoom:
@BastiaanOlij is right in that there needs to be a proper gizmo solution for `CSGMesh3D` at some point, but I would like to address an inaccuracy in his assessment regarding the preservation of existing use cases:

There is actually currently *no use* at all for the `CSGMesh3D` collision visualization as a separate gizmo because:
- CSG visual geometry is identical to the collision geometry, so visualization is already possible.
- For the use case of visualizing _collision-only_ CSGs, the aforementioned improper fix (#84174) has already made it impossible, by making the collision geometry visualization dependent on CSG visibility.

In other words, the current gizmo solution is *entirely useless* because it only renders a wireframe over an *identical underlying mesh*, and *only when the underlying mesh is already visible*.

As such, this PR:
- Does not break any existing use cases (they are already broken)
- Fixes two UX bugs
- Represents a small step toward a proper gizmo solution by removing broken bandaid solutions and setting a clean slate.

BastiaanOlij (approved):
Speaking to @MajorMcDoom on private channels a couple days ago and his further findings reported here, I'm in agreement that the current state of things are as he says and that in order to move forward, fixing one problem at a time is better than endlessly flipping back and forth.

We just have to be mindful that if anyone complains that we don't go back to a more broken situation but make clear that we will be taking steps after this step.

So I'm for merging this and making sure that the current behaviour makes sense, and then in the future following it up with a proper gizmo implementation for CSG collisions.

fire (approved):

akien-mga:
Thanks!

MajorMcDoom:
For posterity, I've made a follow-up issue here: #96807

ydeltastar:
Can't this be cherrypicked for 4.3? The debug draw noise in a level full of CSG composites is absurd and it can't be disabled.

PR #86704, opened 2024-01-02
[GDScript] Add `StringName` support to `@GDScript.len`

### LINKED ISSUE #8774: Add support for StringName in len function

A weird text based game

len(&"test") throws an error, but is expected to return the length of the StringName, similar to how len("test") would show the length of the String. Strings and StringName can be used interchangeably in most other GDScript contexts however where len works for Strings, it does not work for StringNames.

Update the GDScript len function to also support returning the length of a StringName. 

This would make len behave same for both Strings and StringNames, so you could do len(&"test") and get a result

It can be worked around using the StringName.length() function which does the same thing. However using len is the most common method and unexpectedly breaks for StringNames where all other string functions are shared between the two. 

StringNames are core

AThousandShips:
* Closes: https://github.com/godotengine/godot-proposals/issues/8774

### DIFF
--- a/modules/gdscript/doc_classes/@GDScript.xml
+++ b/modules/gdscript/doc_classes/@GDScript.xml
@@ -148,7 +148,7 @@
 			<return type="int" />
 			<param index="0" name="var" type="Variant" />
 			<description>
-				Returns the length of the given Variant [param var]. The length can be the character count of a [String], the element count of any array type or the size of a [Dictionary]. For every other Variant type, a run-time error is generated and execution is stopped.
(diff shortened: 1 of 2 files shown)


akien-mga (approved):
Makes sense to me. Would be good to have this approved by someone from the GDScript team too.

dalexeev (approved):
It would be great to add tests for utility functions, but it is not required for this PR. The docs could be updated:

https://github.com/godotengine/godot/blob/13a0d6e9b253654f5cc2a44f3d0b3cae10440443/modules/gdscript/doc_classes/%40GDScript.xml#L147-L151

akien-mga:
Thanks!

AThousandShips:
Thank you! First PR of 2024!

PR #86708, opened 2024-01-02
Move MIDI's methods and constants into InputEventMIDI

Mickeon:
Superseded by https://github.com/godotengine/godot/pull/86713 and https://github.com/godotengine/godot/pull/86716.

This PR moves OS' methods about MIDI, and the global constants into the **InputEventMIDI**. The new methods are static.
The previous methods and constants still exist, but are marked as **deprecated**.
This means the following:

----
Edit: These methods are currently commented out, because they refuse to compile. Will ask for help later.

| Currently| This PR |
| --- | --- |
| `OS.open_midi_inputs()` | `InputEventMIDI.open_inputs()` |
| `OS.get_connected_midi_inputs()` | `InputEventMIDI.get_connected_inputs()` |
| `OS.close_midi_inputs()` | `InputEventMIDI.close_inputs()` |

| Currently| This PR |
| --- | --- |
| MIDI_MESSAGE_NONE | InputEventMIDI.**MESSAGE_NONE** |
| MIDI_MESSAGE_NOTE_ON | InputEventMIDI.**MESSAGE_NOTE_ON** |
| MIDI_MESSAGE_NOTE_OFF | InputEventMIDI.**MESSAGE_NOTE_OFF** |
| MIDI_MESSAGE_AFTERTOUCH | InputEventMIDI.**MESSAGE_AFTERTOUCH** |
| MIDI_MESSAGE_CONTROL_CHANGE | InputEventMIDI.**MESSAGE_CONTROL_CHANGE** |
| MIDI_MESSAGE_PROGRAM_CHANGE | InputEventMIDI.**MESSAGE_PROGRAM_CHANGE** |
| MIDI_MESSAGE_CHANNEL_PRESSURE | InputEventMIDI.**MESSAGE_CHANNEL_PRESSURE** |
| MIDI_MESSAGE_PITCH_BEND | InputEventMIDI.**MESSAGE_PITCH_BEND** |
| MIDI_MESSAGE_SYSTEM_EXCLUSIVE | InputEventMIDI.**MESSAGE_SYSTEM_EXCLUSIVE** |
| MIDI_MESSAGE_QUARTER_FRAME | InputEventMIDI.**MESSAGE_QUARTER_FRAME** |
| MIDI_MESSAGE_SONG_POSITION_POINTER | InputEventMIDI.**MESSAGE_SONG_POSITION_POINTER** |
| MIDI_MESSAGE_SONG_SELECT | InputEventMIDI.**MESSAGE_SONG_SELECT** |
| MIDI_MESSAGE_TUNE_REQUEST | InputEventMIDI.**MESSAGE_TUNE_REQUEST** |
| MIDI_MESSAGE_TIMING_CLOCK | InputEventMIDI.**MESSAGE_TIMING_CLOCK** |
| MIDI_MESSAGE_START | InputEventMIDI.**MESSAGE_START** |
| MIDI_MESSAGE_CONTINUE | InputEventMIDI.**MESSAGE_CONTINUE** |
| MIDI_MESSAGE_STOP | InputEventMIDI.**MESSAGE_STOP** |
| MIDI_MESSAGE_ACTIVE_SENSING | InputEventMIDI.**MESSAGE_ACTIVE_SENSING** |
| MIDI_MESSAGE_SYSTEM_RESET | InputEventMIDI.**MESSAGE_SYSTEM_RESET** |

### DIFF
--- a/core/input/input_event.cpp
+++ b/core/input/input_event.cpp
@@ -33,6 +33,7 @@
 #include "core/input/input_map.h"
 #include "core/input/shortcut.h"
 #include "core/os/keyboard.h"
+#include "core/os/midi_driver.h"
 #include "core/os/os.h"
 
 const int InputEvent::DEVICE_ID_EMULATION = -1;
@@ -1661,11 +1662,11 @@ int InputEventMIDI::get_channel() const {
 	return channel;
 }
 
-void InputEventMIDI::set_message(const MIDIMessage p_message) {
+void InputEventMIDI::set_message(const Message p_message) {
 	message = p_message;
 }
 
-MIDIMessage InputEventMIDI::get_message() const {
+InputEventMIDI::Message InputEventMIDI::get_message() const {
 	return message;
 }
 
@@ -1717,6 +1718,31 @@ int InputEventMIDI::get_controller_value() const {
 	return controller_value;
 }
 
+void InputEventMIDI::open_inputs() {
+	if (MIDIDriver::get_singleton()) {
+		MIDIDriver::get_singleton()->open();
+	} else {
+		ERR_PRINT(vformat("MIDI input isn't supported on %s.", OS::get_singleton()->get_name()));
+	}
+}
+
+void InputEventMIDI::close_inputs() {
+	if (MIDIDriver::get_singleton()) {
+		MIDIDriver::get_singleton()->close();
+	} else {
+		ERR_PRINT(vformat("MIDI input isn't supported on %s.", OS::get_singleton()->get_name()));
+	}
+}
+
+PackedStringArray InputEventMIDI::get_connected_inputs() {
+	if (MIDIDriver::get_singleton()) {
+		return MIDIDriver::get_singleton()->get_connected_inputs();
+	}
+
+	PackedStringArray list;
+	ERR_FAIL_V_MSG(list, vformat("MIDI input isn't supported on %s.", OS::get_singleton()->get_name()));
+}
+
 String InputEventMIDI::as_text() const {
 	return vformat(RTR("MIDI Input on Channel=%s Message=%s"), itos(channel), itos((int64_t)message));
 }
@@ -1724,19 +1750,19 @@ String InputEventMIDI::as_text() const {
 String InputEventMIDI::to_string() {
 	String ret;
 	switch (message) {
-		case MIDIMessage::NOTE_ON:
+		case Message::MESSAGE_NOTE_ON:
 			ret = vformat("Note On: channel=%d, pitch=%d, velocity=%d", channel, pitch, velocity);
 			break;
-		case MIDIMessage::NOTE_OFF:
+		case Message::MESSAGE_NOTE_OFF:
 			ret = vformat("Note Off: channel=%d, pitch=%d, velocity=%d", channel, pitch, velocity);
 			break;
-		case MIDIMessage::PITCH_BEND:
+		case Message::MESSAGE_PITCH_BEND:
 			ret = vformat("Pitch Bend: channel=%d, pitch=%d", channel, pitch);
 			break;
-		case MIDIMessage::CHANNEL_PRESSURE:
+		case Message::MESSAGE_CHANNEL_PRESSURE:
 			ret = vformat("Channel Pressure: channel=%d, pressure=%d", channel, pressure);
 			break;
-		case MIDIMessage::CONTROL_CHANGE:
+		case Message::MESSAGE_CONTROL_CHANGE:
 			ret = vformat("Control Change: channel=%d, controller_number=%d, controller_value=%d", channel, controller_number, controller_value);
 			break;
 		default:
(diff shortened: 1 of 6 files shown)


core/input/input_event.cpp
```diff
@@ -1661,11 +1662,11 @@ int InputEventMIDI::get_channel() const {
 	return channel;
 }
 
-void InputEventMIDI::set_message(const MIDIMessage p_message) {
+void InputEventMIDI::set_message(const InputEventMIDI::Message p_message) {
```
AThousandShips, line 1665:
```suggestion
void InputEventMIDI::set_message(const Message p_message) {
```
Unnecessary

core/input/input_event.h
```diff
@@ -508,10 +508,33 @@ class InputEventPanGesture : public InputEventGesture {
 };
 
 class InputEventMIDI : public InputEvent {
+public:
+	enum Message {
+		MESSAGE_NONE = 0,
```
AThousandShips, line 513: These should be referencing the enum in `input_enums` to avoid any copy paste errors IMO

core/input/input_event.h
```diff
@@ -508,10 +508,33 @@ class InputEventPanGesture : public InputEventGesture {
...
+		MESSAGE_TUNE_REQUEST = 0xF6,
+		MESSAGE_TIMING_CLOCK = 0xF8,
+		MESSAGE_START = 0xFA,
+		MESSAGE_CONTINUE = 0xFB,
+		MESSAGE_STOP = 0xFC,
+		MESSAGE_ACTIVE_SENSING = 0xFE,
+		MESSAGE_SYSTEM_RESET = 0xFF,
+	};
+private:
 	GDCLASS(InputEventMIDI, InputEvent);
```
AThousandShips, line 534: This should be at the top of the class

core/input/input_event.h
```diff
@@ -526,8 +549,8 @@ class InputEventMIDI : public InputEvent {
 	void set_channel(const int p_channel);
 	int get_channel() const;
 
-	void set_message(const MIDIMessage p_message);
-	MIDIMessage get_message() const;
+	void set_message(const Message p_message);
```
AThousandShips, line 552: Does this actually work? The enum values are the same but it's a different enum so it shouldn't match in GDScript and elsewhere, right?

Mickeon: Not sure what you mean by "work". When I excluded static methods, the engine seems to compile just fine when replacing the **MIDIMessage** enum with **Message** all throughout.

core/input/input_event.h
Mickeon: How? As in, a comment mentioning them?

AThousandShips: As in `MESSAGE_NONE = MIDI_MESSAGE_NONE`, I tgink the enum in the class should be secondary, the `input_enums` should be the one in charge as it were, otherwise we have a lot of secondary dependencies, the other input related enums aren't local to the events

Mickeon: Ok, so the **MIDIMessage** enum will always exist internally, but the assumption is someday it will no longer be exposed in the global scope. Understood.

AThousandShips: My reasoning being that on the GDScript side it makes sense to make the MIDI stuff self contained, because there isn't really anything else with it, but on the C++ side we have drivers etc. which don't really engage with the input events in the same way, so going through the event to do things feels unnecessary (also less changes to keep track of)

core/input/input_event.h
Mickeon: Just verified in my current build (without the static methods because I must reiterate, do not compile). Yes, you can get and set the event's `message` just fine.

AThousandShips: That's strange, because assigning other, matching valued, enums in GDScript like that doesn't work correctly, like doing:
```gdscript
event.messsage = TEXTURE_REPEAT_PARENT_NODE
```
Throws an error as the result type is not the correct enum

AThousandShips: That would be a breaking change btw, even if the types match I'd say

AThousandShips: I'm not sure changing the enum here preserves compatibility, we'll see what the CI says

core/input/input_event.h
```diff
@@ -547,11 +572,16 @@ class InputEventMIDI : public InputEvent {
 	void set_controller_value(const int p_controller_value);
 	int get_controller_value() const;
 
+	void open_inputs();
```
AThousandShips, line 575: These aren't static

core/os/midi_driver.cpp
```diff
@@ -51,57 +51,57 @@ void MIDIDriver::receive_input_packet(uint64_t timestamp, uint8_t *data, uint32_
...
-			event->set_message(MIDIMessage(data[0] >> 4));
+			event->set_message(InputEventMIDI::Message(data[0] >> 4));
 			param_position = 1;
 			last_received_message = data[0];
 		}
 	}
 
 	switch (event->get_message()) {
-		case MIDIMessage::AFTERTOUCH:
+		case InputEventMIDI::Message::AFTERTOUCH:
```
AThousandShips, line 70:
```suggestion
		case InputEventMIDI::Message::MESSAGE_AFTERTOUCH:
```
For all these

Mickeon:
I apologize for the many, many incoming merge pushes that will ensue.

core/input/input_event.h
AThousandShips: Fixed up and built your code and can confirm that this does indeed not work:
```gdscript
event.message = MIDI_MESSAGE_NONE
```
It has to be:
```gdscript
event.message = InputEventMIDI.MESSAGE_NONE
```

This breaks compatibility unless you change the bound method to take the original type

core/input/input_event.h
Mickeon: I don't know how I forgot that

AThousandShips:
Given the above issues with the property I'd suggest removing the enum related changes and keep just the moving of the MIDI driver management, that part I think makes a lot of sense to move out of `OS`, and is smooth to do and doesn't break anything

Mickeon:
I could split the PRs in two and keep this one for the Messages. They took me way longer than the methods to transfer over, and there's some discussion to be had for how useful some of them are.

AThousandShips:
Sounds like a good idea, but unsure how to handle that one without breaking compat at all, so wouldn't be feasible for the short term or even the medium term, but worth saving it

Mickeon:
As of currently, this does not compile specifically because of the enum referencing the other enum. So indeed, they are being problematic.

AThousandShips:
Oh yes that's because of the difference in class-ness, should be solved by just doing `MESSAGE_NONE = (int)MIDIMessage::NONE,`, but that does have some consequences

Mickeon:
This was an accidental removal of a branch but it's no big deal, we are splitting it in two anyway.

PR #75864, opened 2023-04-09
Close built-in shaders when closing scene

KoBeWi:
https://user-images.githubusercontent.com/2223172/230790953-ba957d2b-9b15-40a6-9bba-e5bea4aed3f2.mp4

PR #86710, opened 2024-01-02
Handle built-in shaders when closing scene

KoBeWi:
Resolves TODO in the shader editor. When closing scene, the editor will warn about unsaved shaders. This is relevant after #75864

Also fixed the unsaved shader list displaying wrong names.


akien-mga (approved):
Looks good.

akien-mga:
Thanks!

PR #86693, opened 2024-01-01
Improve all documentation about MIDI support

Mickeon:
### The OS singleton's methods

Basically borrowed from https://github.com/godotengine/godot/pull/80282.

Better links to the other related methods, mention **InputEventMIDI** since it's literally what you're using these methods for. An unusual newline in `get_connected_midi_inputs()` was stripped away. Simple good stuff.

### The InputEventMIDI class

Oh God.

- Several changes to modernize the descriptions of each property [You may be familiar with these](https://github.com/godotengine/godot/pulls?q=is%3Apr+author%3AMickeon+overhaul).
    - "zero" -> "`0`";
    - "MIDI signal" -> "MIDI message"/"Input event";
    - "0 to 127" -> "`0` to `127`".
    - etc.
- Some mentions of "piano" were replaced with "musical keyboard":
    - Look up "_piano_", you get an actual live-performance, perhaps ivory-coated piano.
    Look up "_musical keyboard_", you get the thing you may be actually able to plug into your computer.
- Remove `str()` from the `print()` examples:
    - Print supports variable arguments, so they're redundant.
- Strip away some of https://github.com/godotengine/godot/pull/77103 
    - See https://github.com/godotengine/godot/issues/76678;
    - I get it, this can *rarely* happen based on the device. But the PR didn't have to mention it so verbosely everywhere! A better explanation to the point, plus an example, is much better.
- Removed "For MIDI messages between 0x80 and 0xEF ..." note.
    - I don't see why the user would care about this, they literally cannot touch the original bytes.

### The MIDI_MESSAGE constants
The original intent was to actually explain what to expect from **InputEventMIDI** for any given message type. The problem with that is, putting a lot of **InputEventMIDI** into the global scope was a lot of information that virtually no common user would ever need. So. ideally, it needs to be kept to a minimum while still being more... substantial.

... What the heck, this is confusing. I feel like exposing other MIDI message constants in https://github.com/godotengine/godot/pull/56295 has been a mistake. 
Godot explicitly does **not** support MIDI output, only input. All of the MIDI messages in the PR explicitly need to be sent to other connected MIDI devices. Not only that, but as of currently, most of their data cannot be fetched from **InputEventMIDI** at all, which defeats the point of documenting them in the first place.
<sub>Overall, this has made explaining them in detail a hassle for me. And every time this kind of thing happens, there's an underlying problem in the implementation, I swear. https://github.com/godotengine/godot/pull/67082 </sub>

----------------------

Sources:
[...]

PR #86713, opened 2024-01-02
Move OS' MIDI methods into InputEventMIDI

Mickeon:
Partially addresses https://github.com/godotengine/godot-proposals/issues/8761.
Related to https://github.com/godotengine/godot/pull/86716 and https://github.com/godotengine/godot/pull/86693.

This PR moves the **OS**' methods related to MIDI into the **InputEventMIDI**, as static methods.
The previous methods are still available, but are marked as **deprecated**.

This means the following:
| Currently| This PR |
| --- | --- |
| `OS.open_midi_inputs()` | `InputEventMIDI.open_inputs()` |
| `OS.get_connected_midi_inputs()` | `InputEventMIDI.get_connected_inputs()` |
| `OS.close_midi_inputs()` | `InputEventMIDI.close_inputs()` |

-----------

The original intention was to [move the global constants as well](https://github.com/godotengine/godot/pull/86708), but that fell through for the time being, and will be done in a separate PR for further discussion.

On the other hand, the methods are comparatively safer to implement and do not break existing compatibility. See the proposal for reasoning.

### DIFF
--- a/core/input/input_event.cpp
+++ b/core/input/input_event.cpp
@@ -1841,6 +1841,18 @@ int InputEventMIDI::get_controller_value() const {
 	return controller_value;
 }
 
+void InputEventMIDI::open_inputs() {
+	OS::get_singleton()->open_midi_inputs();
+}
+
+void InputEventMIDI::close_inputs() {
+	OS::get_singleton()->close_midi_inputs();
+}
+
+PackedStringArray InputEventMIDI::get_connected_inputs() {
+	return OS::get_singleton()->get_connected_midi_inputs();
+}
+
 String InputEventMIDI::as_text() const {
 	return vformat(RTR("MIDI Input on Channel=%s Message=%s"), itos(channel), itos((int64_t)message));
 }
@@ -1886,6 +1898,10 @@ void InputEventMIDI::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_controller_number"), &InputEventMIDI::get_controller_number);
 	ClassDB::bind_method(D_METHOD("set_controller_value", "controller_value"), &InputEventMIDI::set_controller_value);
 	ClassDB::bind_method(D_METHOD("get_controller_value"), &InputEventMIDI::get_controller_value);
+
+	ClassDB::bind_static_method("InputEventMIDI", D_METHOD("open_inputs"), &InputEventMIDI::open_inputs);
+	ClassDB::bind_static_method("InputEventMIDI", D_METHOD("close_inputs"), &InputEventMIDI::close_inputs);
+	ClassDB::bind_static_method("InputEventMIDI", D_METHOD("get_connected_inputs"), &InputEventMIDI::get_connected_inputs);
 
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "channel"), "set_channel", "get_channel");
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "message"), "set_message", "get_message");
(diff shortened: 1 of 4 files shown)


AThousandShips (commented):
Looking at the code this has one issue:
The version in `OS` is virtual, this means that if this is overridden anywhere the two versions will no longer line up, this needs to be resolved

Currently no system overrides this, but I'd say the method here should call the OS one

Mickeon:
I updated the PR to have InputEventMIDI's methods call the OS ones as above. The intention for the originals to be deprecated should still be clear.

core/input/input_event.cpp
```diff
@@ -1717,6 +1717,18 @@ int InputEventMIDI::get_controller_value() const {
...
+void InputEventMIDI::open_inputs() {
+	OS::get_singleton()->open_midi_inputs();
+}
+
+void InputEventMIDI::close_inputs() {
+	OS::get_singleton()->close_midi_inputs();
+}
+
+PackedStringArray InputEventMIDI::get_connected_inputs() {
+	OS::get_singleton()->get_connected_midi_inputs();
```
AThousandShips, line 1729:
```suggestion
	return OS::get_singleton()->get_connected_midi_inputs();
```

core/input/input_event.cpp
```diff
@@ -1717,6 +1717,18 @@ int InputEventMIDI::get_controller_value() const {
 	return controller_value;
 }
 
+void InputEventMIDI::open_inputs() {
+	OS::get_singleton()->open_midi_inputs();
```
AThousandShips, line 1721:
```suggestion
	ERR_FAIL_NULL(OS::get_singleton());
	OS::get_singleton()->open_midi_inputs();
```
Or
```suggestion
	if (OS::get_singleton()) {
		OS::get_singleton()->open_midi_inputs();
	}
```
I believe that's what's breaking the test project but unsure

core/input/input_event.cpp
```diff
@@ -1717,6 +1717,18 @@ int InputEventMIDI::get_controller_value() const {
...
+}
+
+void InputEventMIDI::close_inputs() {
+	OS::get_singleton()->close_midi_inputs();
+}
+
+PackedStringArray InputEventMIDI::get_connected_inputs() {
+	return OS::get_singleton()->get_connected_midi_inputs();
```
AThousandShips, line 1729:
```suggestion
	ERR_FAIL_NULL_V(OS::get_singleton(), PackedStringArray());
	return OS::get_singleton()->get_connected_midi_inputs();
```
Or
```suggestion
	if (!OS::get_singleton()) {
		return PackedStringArray();
	}
	return OS::get_singleton()->get_connected_midi_inputs();
```

AThousandShips:
In fact this should probably be handled by updating the regression project, see:
https://github.com/godotengine/regression-test-project/blob/3.x/AutomaticBugs/BasicData.gd#L28

AThousandShips:
My solution was indeed unfortunately wrong, so the two options now are to either change the name or add to the regression project (and revert my suggestions)

Mickeon:
Should be easy if it really is just adding "open_inputs" to the exceptions. Although part of me wonders if it's a good opportunity to come up with a better name for these three methods.

AThousandShips:
It would be simple yeah but that would have to be merged before this

Mickeon:
https://github.com/godotengine/regression-test-project/pull/80

Mickeon:
In the wake of https://github.com/godotengine/godot/pull/86716 this has been rebased too.

PR #86716, opened 2024-01-02
Move MIDI's global constants into InputEventMIDI

Mickeon:
Partially addresses https://github.com/godotengine/godot-proposals/issues/8761.
Related to https://github.com/godotengine/godot/pull/86713 and https://github.com/godotengine/godot/pull/86693.

This PR moves the **MIDIMessage**'s global constants into the **InputEventMIDI**, under the new **Message** enum (prefixing MIDI would be very redudant).
The previous constants still exist, but are marked as **deprecated**.

This means the following:

| Currently| This PR |
| --- | --- |
| MIDI_MESSAGE_NONE | InputEventMIDI.**MESSAGE_NONE** |
| MIDI_MESSAGE_NOTE_ON | InputEventMIDI.**MESSAGE_NOTE_ON** |
| MIDI_MESSAGE_NOTE_OFF | InputEventMIDI.**MESSAGE_NOTE_OFF** |
| MIDI_MESSAGE_AFTERTOUCH | InputEventMIDI.**MESSAGE_AFTERTOUCH** |
| MIDI_MESSAGE_CONTROL_CHANGE | InputEventMIDI.**MESSAGE_CONTROL_CHANGE** |
| MIDI_MESSAGE_PROGRAM_CHANGE | InputEventMIDI.**MESSAGE_PROGRAM_CHANGE** |
| MIDI_MESSAGE_CHANNEL_PRESSURE | InputEventMIDI.**MESSAGE_CHANNEL_PRESSURE** |
| MIDI_MESSAGE_PITCH_BEND | InputEventMIDI.**MESSAGE_PITCH_BEND** |
| MIDI_MESSAGE_SYSTEM_EXCLUSIVE | InputEventMIDI.**MESSAGE_SYSTEM_EXCLUSIVE** |
| MIDI_MESSAGE_QUARTER_FRAME | InputEventMIDI.**MESSAGE_QUARTER_FRAME** |
| MIDI_MESSAGE_SONG_POSITION_POINTER | InputEventMIDI.**MESSAGE_SONG_POSITION_POINTER** |
| MIDI_MESSAGE_SONG_SELECT | InputEventMIDI.**MESSAGE_SONG_SELECT** |
| MIDI_MESSAGE_TUNE_REQUEST | InputEventMIDI.**MESSAGE_TUNE_REQUEST** |
| MIDI_MESSAGE_TIMING_CLOCK | InputEventMIDI.**MESSAGE_TIMING_CLOCK** |
| MIDI_MESSAGE_START | InputEventMIDI.**MESSAGE_START** |
| MIDI_MESSAGE_CONTINUE | InputEventMIDI.**MESSAGE_CONTINUE** |
| MIDI_MESSAGE_STOP | InputEventMIDI.**MESSAGE_STOP** |
| MIDI_MESSAGE_ACTIVE_SENSING | InputEventMIDI.**MESSAGE_ACTIVE_SENSING** |
| MIDI_MESSAGE_SYSTEM_RESET | InputEventMIDI.**MESSAGE_SYSTEM_RESET** |

----

This would also be a great opportunity to discuss the usefulness of these constants, and the names associated with them (see also the sources https://github.com/godotengine/godot/pull/86693):
- Godot does **not** support MIDI output. Only input. A lot of these assume that messages can be sent to other MIDI devices as well, which they can't.
- The **General MIDI** standard is never-changing. It has been established a long while ago and there is little need to keep a majority of what each message corresponds to in mind.
- Most users make games with Godot, they don't need any of these advanced use-case messages.
     - If the user needs a specific message constant for their music game... Hmm... It's complicated.
     The problem is that the **Message** enum may limit our choice. There could be 256 MIDI messages in total...
     Is a value outside of the **Message** enum accepted and returned from `InputEventMIDI.message` just fine? I am not sure right now.

With all of that said:
- **MESSAGE_TUNE_REQUEST** is basically useless. It was only ever used by [analog synthesizers](https://en.wikipedia.org/wiki/Analog_synthesizer), which are rare nowadays.
- **MIDI_MESSAGE_ACTIVE_SENSING** is rarely if ever sent by modern MIDI devices, especially through USB.
- You cannot get **MESSAGE_SONG_POSITION_POINTER**'s data, and is thus basically useless.
- You cannot get **MIDI_MESSAGE_SONG_SELECT**'s data, and is thus basically useless.
- None of the synchronization-related messages can easily be used, because of reasons stated above.
- It is impossible to send **MIDI_MESSAGE_SYSTEM_RESET** to connected MIDI Devices, only receive it for reasons stated above.

### DIFF
--- a/core/input/input_event.compat.inc
+++ b/core/input/input_event.compat.inc
@@ -1,0 +1,46 @@
+/**************************************************************************/
+/*  input_event.compat.inc                                                */
+/**************************************************************************/
+/*                         This file is part of:                          */
+/*                             GODOT ENGINE                               */
+/*                        https://godotengine.org                         */
+/**************************************************************************/
+/* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+/* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef DISABLE_DEPRECATED
+
+void InputEventMIDI::_set_message_bind_compat_86716(const MIDIMessage p_message) {
+	set_message(Message(p_message));
+}
+
+MIDIMessage InputEventMIDI::_get_message_bind_compat_86716() const {
+	return MIDIMessage(get_message());
+}
+
+void InputEventMIDI::_bind_compatibility_methods() {
+	ClassDB::bind_compatibility_method(D_METHOD("set_message", "message"), &InputEventMIDI::_set_message_bind_compat_86716);
+	ClassDB::bind_compatibility_method(D_METHOD("get_message"), &InputEventMIDI::_get_message_bind_compat_86716);
+}
+
+#endif // DISABLE_DEPRECATED
--- a/misc/extension_api_validation/4.4-stable.expected
+++ b/misc/extension_api_validation/4.4-stable.expected
@@ -24,3 +24,12 @@ Validate extension JSON: Error: Field 'classes/OpenXRAPIExtension/methods/regist
 Validate extension JSON: Error: Field 'classes/OpenXRAPIExtension/methods/unregister_projection_views_extension/arguments/0': type changed value in new API, from "OpenXRExtensionWrapperExtension" to "OpenXRExtensionWrapper".
 
 Switched from `OpenXRExtensionWrapperExtension` to parent `OpenXRExtensionWrapper`. Compatibility methods registered.
+
+
+GH-86716
+---------
+Validate extension JSON: Error: Field 'classes/InputEventMIDI/properties/message': type changed value in new API, from "enum::MIDIMessage" to "enum::InputEventMIDI.Message".
+Validate extension JSON: Error: Field 'classes/InputEventMIDI/methods/get_message/return_value': type changed value in new API, from "enum::MIDIMessage" to "enum::InputEventMIDI.Message".
+Validate extension JSON: Error: Field 'classes/InputEventMIDI/methods/set_message/arguments/0': type changed value in new API, from "enum::MIDIMessage" to "enum::InputEventMIDI.Message".
+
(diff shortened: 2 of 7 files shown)


Mickeon:
Seems like the only thing this fails on now is GDExtension, as it's no longer compatibile, indeed... Unsure of what to do about it, but this move was something that was eventually going to happen, sooner or later

AThousandShips:
You need to create compatibility methods

Mickeon:
I've had a thought, but it probably requires another proposal.

The problem here is that **Message** is an enum, but that makes the `message` property limited as it may only accept values within that enum. If that were no longer to be the case, any value from 0 to 255 will be able to be fetched from `message` (kind of like `NOTIFICATION_*` constants). This would also allow removing some questionably useful MIDI message constants, as users could define them just fine if necessary.

aaronfranke:
> This would also be a great opportunity to discuss the usefulness of these constants

They are useful. It's good for Godot to provide information about how to read MIDI input.

Mickeon:
Then there's a whole lot more that could be added (at least last time I checked). It's either more of them, or perhaps trimming the "deck" by only adding constants for MIDI messages that Godot can read the extra data for.

... Or also provide the raw data within InputDataMIDI.

Calinou:
> - Godot does not support MIDI output. Only input. A lot of these assume that messages can be sent to other MIDI devices as well, which they can't.

There's a PR adding MIDI output support: https://github.com/godotengine/godot/pull/98134

Mickeon:
I have just added bind compatibility methods @AThousandShips . I am not sure how to handle this specific case, though, and the code editor is not helping.

AThousandShips:
You need to add them to the class itself, and include the compat file, I can make a diff for you in a bit, or you can see the [documentation](https://docs.godotengine.org/en/latest/contributing/development/handling_compatibility_breakages.html)

Mickeon:
Done-so, I think this is fine?

core/input/input_event.compat.inc
```diff
@@ -0,0 +1,46 @@
...
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef DISABLE_DEPRECATED
+
+void InputEventMIDI::_set_message_bind_compat_86713(const MIDIMessage p_message) {
+	message = MIDIMessage(p_message);
```
AThousandShips, line 34:
```suggestion
	set_message(MIDIMessage(p_message));
```
Should always call the base method to ensure it captures all the behavior

core/input/input_event.compat.inc
```diff
@@ -0,0 +1,46 @@
...
+}
+
+MIDIMessage InputEventMIDI::_get_message_bind_compat_86713() const {
+	return MIDIMessage(message);
```
AThousandShips, line 38:
```suggestion
	return MIDIMessage(get_message());
```
Same here

misc/extension_api_validation/4.4-stable.expected
```diff
@@ -24,3 +24,12 @@ Validate extension JSON: Error: Field 'classes/OpenXRAPIExtension/methods/regist
 Validate extension JSON: Error: Field 'classes/OpenXRAPIExtension/methods/unregister_projection_views_extension/arguments/0': type changed value in new API, from "OpenXRExtensionWrapperExtension" to "OpenXRExtensionWrapper".
 
 Switched from `OpenXRExtensionWrapperExtension` to parent `OpenXRExtensionWrapper`. Compatibility methods registered.
+
+
+GH-86713
```
Bromeon, line 29: These GDExtension changes are specific to this PR, not #86713, right?
```suggestion
GH-86716
```

Mickeon: They are, it's a copy-paste blunder

Bromeon (dismissed):
Looks good from GDExtension perspective.

Unless I'm missing something, you also kept the same integer values for the same constants, which is definitely nice. 

Also, if some MIDI constants are no longer useful, a new enum might be an opportunity to remove them. It could even be that the engine still handles them (in case someone passes an older value), but they're no longer advertised as explicit constants. That would make it easier to phase out in the future.

clayjohn:
CC @raulsntos Do you know if this will break compatibility for C#? Or do you think that the compat bindings will work fine?

raulsntos:
Since the `get_message` and `set_message` methods are used as a property accessors, changing the type of the parameter/return will also change the type of the `message` property and this breaks compatibility in C#.

The main issue is really the getter because in C# you can't have overloads that change only on the return type, otherwise we'd be able to avoid breaking binary compatibility.

Mickeon:
Just for that alone, this may have just become 5.0 material, then.

PR #86721, opened 2024-01-02
Improve GDExtension Tools Integration with Editor Debug Tooling

### LINKED ISSUE #8777: Improve GDExtension Tooling with Editor Debugger Tools

We are working on a complete visual scripting subsystem solution tool for Godot based on GDExtension.

The Editor's debug tooling aspects are very restrictive and limited regarding GDExtension tooling.  In C++ modules, the tooling would have immediate access to things like ScriptDebugger, EditorDebuggerNode, and so on to interact with critical aspects of the debug subsystem, such as setting breakpoints, adding breakpoints to the Editor Debugger Panel, etc.

This change provides several benefits:

**Setting breakpoints**
Plug-in and tool developers will be able to use `EditorDebuggerSession::set_breakpoint` as a way to access the same code path that the `CodeEdit` widget does by interacting with `EditorDebuggerNode` to toggle breakpoints. This allows tooling to set a breakpoint within the script debugger subsystem and register the UI entry in the breakpoints panel.

**Checking Breakpoints**
Adding the `ScriptDebugger` via the `EngineDebugger` singleton, tooling can now check whether a specific line/source pair is a breakpoint and trigger a debugger break. This can also allow tooling to trigger a break within the editor in the same way that GDScript does when a specific error state or condition is raised while running in the editor's debugger.

**Setting Breakpoints**

```cpp
void ScriptNode::set_breakpoint(bool p_active) {
  // Registers breakpoint details in the plugin for this script/node pair
  // This is what gets returned by EditorPlugin::_get_breakpoints
  ScriptPlugin::get_singleton()->set_breakpoint(_script, this, p_active);
  // Now trigger breakpoint via the UI
  // This allows setting the breakpoints details in the Editor Panel and toggling breakpoints during runtime.
  ScriptDebuggerPlugin::get_singleton()->set_breakpoint(_script, this, p_active);
}

void ScriptPlugin::_get_breakpoints(const Ref<Script>& p_script, ScriptNode* p_node, bool p_active) {
  const String break_point = format("%s:%d", p_script->get_path(), p_node->get_id());
  if (p_active && !_breakpoints.has(break_point)) {
    _breakpoints.push_back(break_point);
  }
  else if (!p_active && _breakpoints.has(break_point)) {
    _breakpoints.remove(break_point);
  }
}

void ScriptDebuggerPlugin::set_breakpoint(const Ref<Script>& p_script, ScriptNode* p_node, bool p_active) {
  // Lookup the current session managed by the custom EditorDebuggerPlugin
  Ref<EditorDebuggerSession> session = _get_current_session();
  if (session.is_valid()) {
    // Session is valid, set the breakpoint
    // This toggles the breakpoint in the Editor UI Debugger Panel and will pass it to the underlying debugger 
    // if a debugger session is currently active, allowing for dynamic breakpoint registration
    session->set_breakpoint(p_script.get_path(), p_node->get_id(), p_active);
  }
}
```

**Checking breakpoints at runtime**

```cpp
Variant ScriptInstance::execute_step() {
  /* execute script step */
  if (EngineDebugger::get_singleton()->is_active()) {
    ScriptDebugger *debugger = EngineDebugger::get_singleton()->get_script_debugger();
    if (current_context->has_error()) {     
      debugger->debug(ScriptLanguage::get_singleton(), false, false);
    }
    else if (debugger->is_breakpoint(current_node_id, _script->get_path()) {
      debugger->debug(ScriptLanguage::get_singleton(), false, true);
    }
  }
}
```

No, these APIs are currently not available.

This relates to exposed Editor functionality.

Naros:
This PR aims to address several issues around GDExtension and Debug Tooling within the Editor.

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/8777*

### DIFF
--- a/doc/classes/EditorDebuggerSession.xml
+++ b/doc/classes/EditorDebuggerSession.xml
@@ -50,6 +50,15 @@
 				Sends the given [param message] to the attached remote instance, optionally passing additionally [param data]. See [EngineDebugger] for how to retrieve those messages.
 			</description>
 		</method>
+		<method name="set_breakpoint">
+			<return type="void" />
+			<param index="0" name="path" type="String" />
+			<param index="1" name="line" type="int" />
+			<param index="2" name="enabled" type="bool" />
+			<description>
+				Enables or disables a specific breakpoint based on [param enabled], updating the Editor Breakpoint Panel accordingly.
+			</description>
+		</method>
 		<method name="toggle_profiler">
 			<return type="void" />
 			<param index="0" name="profiler" type="String" />
--- a/doc/classes/EngineDebugger.xml
+++ b/doc/classes/EngineDebugger.xml
@@ -9,6 +9,32 @@
 	<tutorials>
 	</tutorials>
 	<methods>
+		<method name="clear_breakpoints">
+			<return type="void" />
+			<description>
+				Clears all breakpoints.
+			</description>
+		</method>
+		<method name="debug">
+			<return type="void" />
+			<param index="0" name="can_continue" type="bool" default="true" />
+			<param index="1" name="is_error_breakpoint" type="bool" default="false" />
+			<description>
+				Starts a debug break in script execution, optionally specifying whether the program can continue based on [param can_continue] and whether the break was due to a breakpoint.
+			</description>
+		</method>
+		<method name="get_depth" qualifiers="const" experimental="">
+			<return type="int" />
+			<description>
+				Returns the current debug depth.
+			</description>
+		</method>
+		<method name="get_lines_left" qualifiers="const" experimental="">
+			<return type="int" />
+			<description>
+				Returns the number of lines that remain.
+			</description>
+		</method>
 		<method name="has_capture">
 			<return type="bool" />
 			<param index="0" name="name" type="StringName" />
@@ -23,10 +49,26 @@
 				Returns [code]true[/code] if a profiler with the given name is present otherwise [code]false[/code].
 			</description>
 		</method>
+		<method name="insert_breakpoint">
+			<return type="void" />
+			<param index="0" name="line" type="int" />
+			<param index="1" name="source" type="StringName" />
+			<description>
+				Inserts a new breakpoint with the given [param source] and [param line].
+			</description>
+		</method>
 		<method name="is_active">
 			<return type="bool" />
 			<description>
 				Returns [code]true[/code] if the debugger is active otherwise [code]false[/code].
+			</description>
+		</method>
+		<method name="is_breakpoint" qualifiers="const">
+			<return type="bool" />
+			<param index="0" name="line" type="int" />
+			<param index="1" name="source" type="StringName" />
+			<description>
+				Returns [code]true[/code] if the given [param source] and [param line] represent an existing breakpoint.
 			</description>
 		</method>
 		<method name="is_profiling">
@@ -34,6 +76,18 @@
 			<param index="0" name="name" type="StringName" />
 			<description>
 				Returns [code]true[/code] if a profiler with the given name is present and active otherwise [code]false[/code].
+			</description>
+		</method>
+		<method name="is_skipping_breakpoints" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if the debugger is skipping breakpoints otherwise [code]false[/code].
+			</description>
+		</method>
+		<method name="line_poll">
+			<return type="void" />
+			<description>
+				Forces a processing loop of debugger events. The purpose of this method is just processing events every now and then when the script might get too busy, so that bugs like infinite loops can be caught
 			</description>
 		</method>
 		<method name="profiler_add_frame_data">
@@ -70,12 +124,43 @@
 				Registers a profiler with the given [param name]. See [EngineProfiler] for more information.
 			</description>
 		</method>
+		<method name="remove_breakpoint">
+			<return type="void" />
+			<param index="0" name="line" type="int" />
+			<param index="1" name="source" type="StringName" />
+			<description>
+				Removes a breakpoint with the given [param source] and [param line].
+			</description>
+		</method>
+		<method name="script_debug">
+			<return type="void" />
+			<param index="0" name="language" type="ScriptLanguage" />
+			<param index="1" name="can_continue" type="bool" default="true" />
+			<param index="2" name="is_error_breakpoint" type="bool" default="false" />
+			<description>
+				Starts a debug break in script execution, optionally specifying whether the program can continue based on [param can_continue] and whether the break was due to a breakpoint.
+			</description>
+		</method>
 		<method name="send_message">
 			<return type="void" />
 			<param index="0" name="message" type="String" />
 			<param index="1" name="data" type="Array" />
 			<description>
 				Sends a message with given [param message] and [param data] array.
+			</description>
+		</method>
+		<method name="set_depth" experimental="">
+			<return type="void" />
+			<param index="0" name="depth" type="int" />
+			<description>
+				Sets the current debugging depth.
+			</description>
+		</method>
+		<method name="set_lines_left" experimental="">
+			<return type="void" />
+			<param index="0" name="lines" type="int" />
+			<description>
+				Sets the current debugging lines that remain.
 			</description>
 		</method>
 		<method name="unregister_message_capture">
--- a/core/core_bind.cpp
+++ b/core/core_bind.cpp
@@ -33,6 +33,7 @@
 #include "core/config/project_settings.h"
 #include "core/crypto/crypto_core.h"
 #include "core/debugger/engine_debugger.h"
+#include "core/debugger/script_debugger.h"
 #include "core/io/file_access_compressed.h"
 #include "core/io/file_access_encrypted.h"
 #include "core/io/marshalls.h"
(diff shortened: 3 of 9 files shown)


AThousandShips:
Please open a [proposal](https://github.com/godotengine/godot-proposals) to track this and discuss the specifics of this feature

Naros:
@AThousandShips done - https://github.com/godotengine/godot-proposals/issues/8777
Will take a look at the formatting issues now.

doc/classes/ScriptDebugger.xml
AThousandShips, line 1: Please provide descriptions for this new documentation, at least a basic description of all methods and the class itself 🙂

doc/classes/EditorDebuggerSession.xml
AThousandShips, line 1: Same here

doc/classes/ScriptDebugger.xml
Naros: Thanks for the help and guidance, I think it should be set I hope.

Naros:
Hi, would someone me able to tell me whether the failure is related to my change or something that pre-exists in `main`?

core/register_core_types.cpp
```diff
@@ -290,6 +291,7 @@ void register_core_types() {
 	_classdb = memnew(core_bind::special::ClassDB);
 	_marshalls = memnew(core_bind::Marshalls);
 	_engine_debugger = memnew(core_bind::EngineDebugger);
+	_script_debugger = memnew(core_bind::ScriptDebugger);
```
AThousandShips, line 294: You need to delete this, it doesn't happen automatically, see below

dsnopek:
Thanks!

At a high-level, I think this is a great addition, since it's making something that's possible via Godot modules also be possible via GDExtension (and addons).

I'm not really familiar with the debugger API, but for the most part, the code here looks like a thin layer over `ScriptDebugger` from `core/debugger/script_debugger.h`, which seems good to me. But it should probably be reviewed by someone more familiar with this API (@Faless?) to see if the API that's being exposed really makes sense.

Naros:
I'd also add that if there are other bits from `SccriptDebugger` that make sense to include, let me know.  I basically grabbed just what I needed for the short-term but perhaps there are others that may warrant inclusion.

vnen (approved):
I'm not the most familiar with the debugger code, but this is mostly exposing what's there for extensions. Overall it looks fine to me.

AThousandShips (approved):

akien-mga:

Faless (commented):
Do we really want to expose `ScriptDebugger` as a separate class?

I feel that class exists in core mostly for historical reasons.

Does it make sense to move the exposed functions directly to `EngineDebugger`?

E.g.:
```
core_bind::EngineDebugger::insert_breakpoint
```

Naros:
> Do we really want to expose `ScriptDebugger` as a separate class?
> 
> I feel that class exists in core mostly for historical reasons.
> 
> Does it make sense to move the exposed functions directly to `EngineDebugger`?

I have to agree with David, really entirely your call, I can certainly align those there if you'd prefer.

Faless:
> Unfortunately, you may be the person most qualified to answer the questions you pose. :-)

> I have to agree with David, really entirely your call, I can certainly align those there if you'd prefer.

Given there doesn't seem to be any objection, I will say let's do that.

It's already confusing enough to have `EngineDebugger` and `EditorDebugger`, adding a `ScriptDebugger` will make it even worse.

I already had plans to move more of the debugger default features to the plugin interface (while still shipping them as part of Godot), so I will take a chance there to see if we can get rid of the whole ScriptDebugger concept internally too.

Naros:
Sounds good, then I'll amend the PR over the weekend and ping you when its ready for a review from your side on the particulars to make sure that the direction and goals are aligned to what you have in mind 👍 @Faless

Naros:
Hi @Faless, I have a few questions about this work.

I was reviewing and testing my changes in a local build to make sure I hadn't missed any additions needed for this to work well for our script language plug-in, and I came across a hack I had used to integrate with the `goto_line` signal.

```gdscript
TypedArray<Node> nodes = editor_node->find_children("*", "EditorDebuggerNode", true, false);
if (nodes.size() == 1) {
  // Wire this signal so when user selects a frame we can handle that here.
  Node* editor_debugger_node = Object::cast_to<Node>(nodes[0]);
  editor_debugger_node->connect("goto_script_line", callable_mp(this, &OrchestratorScriptView::_goto_script_line));
}

The `ScriptEditor` hooks into this directly into the engine to advance the developer position in the script view to the correct line based on the chosen breakpoint; however, as our plug-in isn't based on the `ScriptEditor` classes, we cannot take advantage of that callback.  I want to find a reasonable way to expose this signal.  

What I was considering was adding a new method on `EditorDebuggerPlugin`, i.e.:
```
void goto_script_line(const Ref<Script>& p_script, int p_line);
```

Then, inside `EditorDebuggerNode::_error_selected` and `_text_editor_stack_goto` where, we emit the `goto_script_line` signal that we iterate the registered debugger plugins and dispatch to this new method, allowing the plug-in an opportunity to react.  There may be other alternatives for this, but it seemed like the most likely candidate.

The last task I have open is to check how GDScript integrates with the Breakpoints bottom panel.  I'd like for my visual script tooling to add the breakpoints to that panel much like how the gutter breakpoints work in GDScript.  I should hopefully have that worked out tomorrow with any last questions before this is ready.

**Comment by Naros** [comment]
The build failures seem to be unrelated as `master` seems to be unstable atm.

**Comment by Naros** [comment]
@akien-mga @Faless is it possible to still get this into 4.3?  It would be super helpful for us to have debugger support in the editor and this is a blocker for that.

**Comment by AThousandShips** [comment]
It's marked for 4.3 so unless that is changed it's still planned for 4.3, wait and see

File: doc/classes/EngineDebugger.xml [hunk 2]
```diff
@@ -9,6 +9,32 @@
...
+		</method>
+		<method name="debug">
+			<return type="void" />
+			<param index="0" name="can_continue" type="bool" default="true" />
+			<param index="1" name="is_error_breakpoint" type="bool" default="false" />
+			<description>
+				Starts a debug break in script execution, optionally specifying whether the program can continue based on [param can_continue] and whether the break was due to a breakpoint.
+			</description>
+		</method>
+		<method name="get_depth" qualifiers="const">
```
Faless, line 26:
```suggestion
		<method name="get_depth" qualifiers="const" experimental="">
```

doc/classes/EngineDebugger.xml
```diff
@@ -9,6 +9,32 @@
...
+			<return type="int" />
+			<description>
+				Returns the current debug depth.
+			</description>
+		</method>
+		<method name="get_lines_left" qualifiers="const">
```
Faless, line 32:
```suggestion
		<method name="get_lines_left" qualifiers="const" experimental="">
```

doc/classes/EngineDebugger.xml
```diff
@@ -78,6 +149,20 @@
 				Sends a message with given [param message] and [param data] array.
 			</description>
 		</method>
+		<method name="set_depth">
```
Faless, line 152:
```suggestion
		<method name="set_depth" experimental="">
```

doc/classes/EngineDebugger.xml
```diff
@@ -78,6 +149,20 @@
...
+			<return type="void" />
+			<param index="0" name="depth" type="int" />
+			<description>
+				Sets the current debugging depth.
+			</description>
+		</method>
+		<method name="set_lines_left">
```
Faless, line 159:
```suggestion
		<method name="set_lines_left" experimental="">
```

Faless (approved):
Fantastic job! :trophy: , and sorry for the long delay between reviews.

I was worried of the potential misuse of `set_lines`/`set_depth` which can potentially break debugging but I couldn't figure out a better solution, and it's better to have this than not in 4.3.

So I've approved this, but also suggested marking those methods as experimental.

Thank you for your patience and the great work.

Faless:
> Is it safe then for my `EditorDebuggerPlugin` to cache the `EditorDebuggerSession` or its ID (as I think there should only be one active),

The ID is a bit of an implementation detail, but it's indeed the "tab number".
In general, it should be fine caching between then signals `EditorDebuggerSession.started` and `EditorDebuggerSession.stopped`.

> I also need a way to react to `breakpoint_set_in_tree` and `breakpoints_cleared_in_tree`, which are signals that are also emitted by the `EditorDebuggerNode`, when the user uses the context-menu on the breakpoints in the panel. Again, delegate these to the Plugin or should we consider a different way?

I think indeed those signals could be exposed via the `EditorDebuggerPlugin` given they are "global" (i.e. are relevant to all sessions).

Maybe `goto_script_line` could also be exposed this way?

Naros:
Thanks for the feedback @Faless, I've addressed your recommendations with the documentation and I've added the 3 methods as virtual methods on the `EditorDebuggerPlugin` interface.  Assuming the build succeeds, let me know if there is anything else you would like to see for this.

akien-mga:
Thanks!

Faless:
> I've added the 3 methods as virtual methods on the `EditorDebuggerPlugin` interface.

Oh no... as I mentioned in my comment, those should have been signals not methods :(

Naros:
> > I've added the 3 methods as virtual methods on the `EditorDebuggerPlugin` interface.
> 
> Oh no... as I mentioned in my comment, those should probably have been signals not virtual methods :(

No worries, I can send another PR to follow-up on this if that's acceptable.

Faless:
@Naros I'll let you know in case, it's not the end of the world indeed, just a bit weird, we usually prefer virtual methods for things that needs overriding and either are frequently called (they're faster than signals), or when we need to receive a return (which is not the case for those function).

At the same time, the plugin aims to be "self-contained", so the virtual method can make sense, and might not be worth changing.

Sorry for the noise, I think in the end we'll leave it as it is now :sweat_smile:

PR #86729, opened 2024-01-02
Fix missing autocompletion for inheriting classes

Mickeon:
This PR ensures we always call the very next inherited class when getting autocompletion results. This is more future-proof. Previously, some classes were "jumping" way ahead, potentially skipping some options. When people add new autocompletion results, they should be reminded of this, perhaps.

This also fixes **Input** and **SceneTree** not having any of **Object**'s autocompletion. You may have never noticed.


akien-mga (approved):

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

Issue #45296, opened 2021-01-18
Lightmapper Crashes on large scenes


NHodgesVFX:
**Godot version:**

Godot 3.2.4 Beta 6 Compiled with MSVC

**OS/device including version:**

GLES3 RTX2080s Windows 10 64bit

**Issue description:**

While attempting to bake the Godot tps demo, It often crashes at the fixing seems and denoising or the capture step. I can stop the first crash by setting the quality level to low or disabling denoising. The 2nd crash is fixed by turning off capture or setting the cell size to a large value like 25.

I remember reading the godot blog about how the light octree used in the capture setting is inefficient and that light probes would be better for larger levels and reduz confirmed this. The primary issue is the first crash though.

**Steps to reproduce:**
1. Download the tps demo from master branch https://github.com/godotengine/tps-demo
2.  Open it in godot and go into the level->geometry->models folder and find all the gltf files. Reimport them with lightmap baking enabled.
3. open level.tscn, delete gi probe and setup a bakedlightmap node so it fits the scene. Enable environment set to scene. set bounces to 5.
4. try baking it will most likely crash on the fixing seems and denoising. 
5. turn off denoising and set the bake quality to low, you should now be able to get past the first crash.
6. Observe it now crashes at the capture step, raising the cell size to 25 fixes it.

**Minimal reproduction project:**

This bug only shows it self on large projects like the tps demo. Most likely due to the scale and amount of meshes to bake. I managed to get it on a test scene with kitbash3d assets as well but cant share due to licensing. 

I can link to a zip with the tps demo setup but its large and the changes which I made to the scene are listed above.

qarmin:
Backtrace:
```
ERROR: operator[]: FATAL: Index p_index = 4 is out of bounds (count = 4).
   At: ./core/local_vector.h:163.
handle_crash: Program crashed with signal 4
Dumping the backtrace. Please include this when reporting the bug on https://github.com/godotengine/godot/issues
[1] /lib/x86_64-linux-gnu/libc.so.6(+0x46210) [0x7fedab9c4210] (??:0)
[2] LocalVector<int, unsigned int, false>::operator[](unsigned int) (/mnt/Miecz/godot3.2/./core/local_vector.h:163 (discriminator 1))
[3] LightmapperCPU::_compute_indirect_light(unsigned int, void*) (/mnt/Miecz/godot3.2/modules/lightmapper_cpu/lightmapper_cpu.cpp:859)
[...]

PR #77306, opened 2023-05-21
Fix crash in FileAccessCompressed and improve error handling

RedworkDE:
Fixes partially https://github.com/godotengine/godot/issues/62585 (on MSVC #74582 is also required or it will just crash in `CowData::resize` before any of this matters)

Only the change in `WRITE_FIT` is strictly required to fix the crash.
The error handling improvements are required so that `ResourceSaver::save` doesn't OK when output file will be corrupted, with them it correctly returns `ERR_CANT_CREATE`.
Without the `FileAccessCompressed::store_buffer` implementation, the file writing will instead get stuck spamming errors for each byte in the buffer.

PR #74582, opened 2023-03-08
Refactor how `CowData` manages its allocations

RedworkDE:
Context: I was looking into making CowData use 64-bit indices to solve the various crashes / issues caused by too large arrays, and found that the internal structure of CowData is fundamentally not suited for such a thing (and also really weird) so I went and reworked how the internal data storage is managed.

To the outside this PR shouldn't change any behavior, except that allocating somewhat larger arrays works now (and also on MSVC there are no longer cases where operations return OK but actually failed).

There are a few more improvements that could / should be done, but I stopped before feature creep completely overtook this.

~~Needs https://github.com/godotengine/godot/pull/74555 to pass CI.~~

- Use an explicit type for the data prefix to enable easier changes.
- Explicitly store the capacity instead of using the next power of two.
- If resizing to a significantly larger size (more 2x the current capacity), assume that the caller has just resized to their final size and don't allocate any excess capacity.
- Trim the capacity less aggressively to fix edge cases when repeatedly adding and removing elements causing extreme performance issues.
- Do not rely on overflow checked operations to determine the maximum size, to avoid crashes with some compilers when using too large sizes.
- On 64-bit system increase the the maximum size to 0x7ffffff0, generally by a factor of the size of the elements.
- On 32-bit system the maximum size remains limited by the size of the address space / memory, but in some cases it may still be possible to allocate larger arrays compared to before.
- Add a minimum capacity of 4 elements to remove some unnecessary reallocation for small arrays.
- Remove a few unnecessary potential copies when resizing, adding or removing elements.

---

What to do with the `ERR_PROPAGATE` macro?
I added this to check an `Error` return value and to return it from the current method if it is not OK, is this usefull for other places in the engine as well and should be added to the other ERR macros? Should I just inline that check everywhere? Or can it just stay as is?

---

What about alignment?
The old version of this just didn't care about alignment and just returned what it got from the allocator (which is malloc aligned + 16 bytes).
This assumes that the allocator pointer is aligned correctly and then offsets the first element by an multiple of the alignment, but I am not sure if the alignment is correct for types with large alignments, ie simd types. Should this be extended to ensure correct alignment for such types?

---

About the buffers in `max_size`:
The 64 bytes buffer to filling the entire address space with one array is arbitrary and not going to matter beyond preventing overflows when computing the size of the allocation as way more than that are going to be filled with godot, so such allocations are going to fail anyways.
The 15 indices buffer is similarly arbitrary and serves to prevent overflows when computing `size()+1` and similar things.
[...]

Issue #54679, opened 2021-11-06
Vulkan: Crash when baking lightmaps with LightmapGI if the resulting image is too large (due to CowData using 32-bit ints)


Calinou:
### Godot version

4.0.dev (4651b2ae5)

### System information

Fedora 34, GeForce GTX 1080 (NVIDIA 470.74)

### Issue description

The editor crashes when baking lightmaps with LightmapGI in a specific scene (see the MRP included).

Backtrace:

```
[1] /lib64/libc.so.6(+0x3d320) [0x7fdb5cf60320] (??:0)
[2] /lib64/libc.so.6(+0x163b81) [0x7fdb5d086b81] (??:0)
[3] Image::create(int, int, bool, Image::Format) (/home/hugo/Documents/Git/godotengine/godot/core/io/image.cpp:1996)
[4] LightmapperRD::_blit_meshes_into_atlas(int, Vector<Ref<Image> >&, Vector<Ref<Image> >&, AABB&, Vector2i&, int&, bool (*)(float, String const&, void*, bool), void*) (/home/hugo/Documents/Git/godotengine/godot/modules/lightmapper_rd/lightmapper_rd.cpp:258)
[5] LightmapperRD::bake(Lightmapper::BakeQuality, bool, int, float, int, bool, Lightmapper::GenerateProbes, Ref<Image> const&, Basis const&, bool (*)(float, String const&, void*, bool), void*) (/home/hugo/Documents/Git/godotengine/godot/modules/lightmapper_rd/lightmapper_rd.cpp:684)
[6] LightmapGI::bake(Node*, String, bool (*)(float, String const&, void*, bool), void*) (/home/hugo/Documents/Git/godotengine/godot/scene/3d/lightmap_gi.cpp:954)
[7] LightmapGIEditorPlugin::_bake_select_file(String const&) (/home/hugo/Documents/Git/godotengine/godot/editor/plugins/lightmap_gi_editor_plugin.cpp:39)
[8] void call_with_variant_args_helper<LightmapGIEditorPlugin, String const&, 0ul>(LightmapGIEditorPlugin*, void (LightmapGIEditorPlugin::*)(String const&), Variant const**, Callable::CallError&, IndexSequence<0ul>) (/home/hugo/Documents/Git/godotengine/godot/./core/variant/binder_common.h:223)
[...]

PR #86730, opened 2024-01-03
Promote CowData to 64 bits

### LINKED ISSUE #62585: Godot crash in ResourceSaver after 1073741824 bytes
### Godot version

4.0.alpha c41e4b1

### System information

Windows 10, 64-bit

### Issue description

I ran into this when saving a large scn with lots of textures, and used "Make Local" so all textures got embedded.

I tried saving this scene, and it gave an error in resize
```
ERROR: Condition "p_size < 0" is true. Returning: ERR_INVALID_PARAMETER
   at: CowData<unsigned char>::resize (S:\repo\godot-fire\core/templates/cowdata.h:261)
```
After this error, it proceeded to crash.
```
 	godot.windows.opt.tools.64.exe!FileAccessCompressed::store_8(unsigned char p_dest) Line 339	C++
>	godot.windows.opt.tools.64.exe!FileAccess::store_buffer(const unsigned char * p_src, unsigned __int64 p_length) Line 552	C++
 	godot.windows.opt.tools.64.exe!ResourceFormatSaverBinaryInstance::write_variant(Ref<FileAccess> f, const Variant & p_property, HashMap<Ref<Resource>,int,HashMapHasherDefault,HashMapComparatorDefault<Ref<Resource> >,DefaultTypedAllocator<HashMapElement<Ref<Resource>,int> > > & resource_map, HashMap<Ref<Resource>,int,HashMapHasherDefault,HashMapComparatorDefault<Ref<Resource> >,DefaultTypedAllocator<HashMapElement<Ref<Resource>,int> > > & external_resources, HashMap<StringName,int,HashMapHasherDefault,HashMapComparatorDefault<StringName>,DefaultTypedAllocator<HashMapElement<StringName,int> > > & string_map, const PropertyInfo & p_hint) Line 1657	C++
 	godot.windows.opt.tools.64.exe!ResourceFormatSaverBinaryInstance::write_variant(Ref<FileAccess> f, const Variant & p_property, HashMap<Ref<Resource>,int,HashMapHasherDefault,HashMapComparatorDefault<Ref<Resource> >,DefaultTypedAllocator<HashMapElement<Ref<Resource>,int> > > & resource_map, HashMap<Ref<Resource>,int,HashMapHasherDefault,HashMapComparatorDefault<Ref<Resource> >,DefaultTypedAllocator<HashMapElement<Ref<Resource>,int> > > & external_resources, HashMap<StringName,int,HashMapHasherDefault,HashMapComparatorDefault<StringName>,DefaultTypedAllocator<HashMapElement<StringName,int> > > & string_map, const PropertyInfo & p_hint) Line 1637	C++
 	godot.windows.opt.tools.64.exe!ResourceFormatSaverBinaryInstance::save(const String & p_path, const Ref<Resource> & p_resource, unsigned int p_flags) Line 2081	C++
 	godot.windows.opt.tools.64.exe!ResourceFormatSaverBinary::save(const String & p_path, const Ref<Resource> & p_resource, unsigned int p_flags) Line 2106	C++
 	godot.windows.opt.tools.64.exe!ResourceSaver::save(const String & p_path, const Ref<Resource> & p_resource, unsigned int p_flags) Line 110	C++
```

CowData resizes to the next power of two, which is 2147483648, and overflows a signed int.

I think everything here should be made less sloppy. Array operations should always use unsigned, not signed values. Integers should be 64-bit where possible: GDScript already uses 64-bit, so why are they downgraded to 32-bit in C++? Finally, FileAccess should check the return value of `resize()` and fail gracefully.

### Steps to reproduce

1. Import a very heavy scene from glTF, such that it has a few GB of assets.
2. Use "Make local" or do a compessed export using a script.

Sorry I don't have more specific steps. The asset I used is not redistributable, and the save was automated by a script in order to enable ZSTD compression.

N/A

### LINKED ISSUE #77256: Error when resizing PackedByteArray bigger than int32 limit
### Godot version

4.0.3

### System information

Windows 11 64-Bit

### Issue description

The PackedByteArray can't be bigger than 2147483647. Following [this](https://github.com/godotengine/godot/issues/46842#issuecomment-842216412) issue it should be fixed in Godot 4.0 RC 1, but running the code below in Godot 4.0.3 results in this error.

```
E 0:00:00:0834   Node.gd:9 @ _ready(): Condition "p_size < 0" is true. Returning: ERR_INVALID_PARAMETER
  <C++-Quelle>   ./core/templates/cowdata.h:262 @ resize()
  <Stacktrace>   Node.gd:9 @ _ready()

```

### Steps to reproduce

Run this code:
```
	var INT32_MAX = 2147483647 # 2^31 - 1
	var byte_arr = PackedByteArray()
	byte_arr.resize(INT32_MAX) # OK
	byte_arr.resize(INT32_MAX + 1) # error
```

[PackedByteArrayError.zip](https://github.com/godotengine/godot/files/11520507/PackedByteArrayError.zip)

### LINKED ISSUE #80533: Attempting to read large file as string causes segfault
### Godot version

4.1.1.stable, master [4714e9589]

### System information

Godot v4.2.dev (4714e9589) - Arch Linux #1 ZEN SMP PREEMPT_DYNAMIC Tue, 08 Aug 2023 22:13:48 +0000 - Wayland - Vulkan (Forward+) - dedicated AMD Radeon RX 5700 (RADV NAVI10) () - AMD Ryzen 7 5800X3D 8-Core Processor (16 Threads)

### Issue description

I attempted to read a large JSON file (1GB+) using `FileAccess.get_file_as_string`, but the result is a crash. Here's a backtrace from a debug build [4714e9589] and some info, not sure what to make of this

```
Loading...

Thread 1 "godot.linuxbsd." received signal SIGSEGV, Segmentation fault.
String::parse_utf8 (this=0x7fffffffa5f8, 
    p_utf8=0x7fff2ffff020 "[{\"p\":[-0.019984539598226547,0.3165023624897003,0.6493667960166931],\"c\":[0.15294925029263085,0.1603185939029318,0.11855615533237046],\"s\":[-5.519049167633057,-5.86762809753418,-4.488621234893799],\"q\":["..., p_len=1093019370, p_skip_cr=false) at core/string/ustring.cpp:1880
1880            dst[str_size] = 0;
(gdb) bt
#0  String::parse_utf8 (this=0x7fffffffa5f8, 
    p_utf8=0x7fff2ffff020 "[{\"p\":[-0.019984539598226547,0.3165023624897003,0.6493667960166931],\"c\":[0.15294925029263085,0.1603185939029318,0.11855615533237046],\"s\":[-5.519049167633057,-5.86762809753418,-4.488621234893799],\"q\":["..., p_len=1093019370, p_skip_cr=false) at core/string/ustring.cpp:1880
#1  0x000055555cae1c27 in FileAccess::get_file_as_string (p_path=..., r_error=0x0) at core/io/file_access.cpp:770
#2  0x000055555cae631e in FileAccess::_get_file_as_string (p_path=...) at core/io/file_access.h:227
#3  0x000055555caf36ae in call_with_variant_args_static_ret<String, String const&, 0ul> (p_method=0x55555cae62e7 <FileAccess::_get_file_as_string(String const&)>, p_args=0x7fffffffa720, r_ret=..., r_error=...)
    at ./core/variant/binder_common.h:766
#4  0x000055555caf0cbd in call_with_variant_args_static_ret_dv<String, String const&> (p_method=0x55555cae62e7 <FileAccess::_get_file_as_string(String const&)>, p_args=0x7fffffffa838, p_argcount=1, r_ret=..., r_error=..., 
    default_values=...) at ./core/variant/binder_common.h:989
#5  0x000055555caeec5a in MethodBindTRS<String, String const&>::call (this=0x5555616f8240, p_object=0x0, p_args=0x7fffffffa838, p_arg_count=1, r_error=...) at ./core/object/method_bind.h:719
#6  0x000055555870c3db in GDScriptFunction::call (this=0x555563c0aa10, p_instance=0x555563bfaba0, p_args=0x0, p_argcount=0, r_err=..., p_state=0x0) at modules/gdscript/gdscript_vm.cpp:1934
#7  0x000055555859a338 in GDScriptInstance::callp (this=0x555563bfaba0, p_method=..., p_args=0x0, p_argcount=0, r_error=...) at modules/gdscript/gdscript.cpp:1892
#8  0x000055555aaeec0c in Node::_gdvirtual__ready_call<false> (this=0x555563bf7930) at scene/main/node.h:322
#9  0x000055555aac7d65 in Node::_notification (this=0x555563bf7930, p_notification=13) at scene/main/node.cpp:186
#10 0x00005555582b77d4 in Node::_notificationv (this=0x555563bf7930, p_notification=13, p_reversed=false) at ./scene/main/node.h:49
#11 0x00005555582d7147 in Node3D::_notificationv (this=0x555563bf7930, p_notification=13, p_reversed=false) at ./scene/3d/node_3d.h:52
#12 0x000055555d092445 in Object::notification (this=0x555563bf7930, p_notification=13, p_reversed=false) at core/object/object.cpp:798
#13 0x000055555aac823e in Node::_propagate_ready (this=0x555563bf7930) at scene/main/node.cpp:230
#14 0x000055555aac81be in Node::_propagate_ready (this=0x555563b84780) at scene/main/node.cpp:221
#15 0x000055555aadae3e in Node::_set_tree (this=0x555563b84780, p_tree=0x555563b84280) at scene/main/node.cpp:2909
#16 0x000055555ab18704 in SceneTree::initialize (this=0x555563b84280) at scene/main/scene_tree.cpp:449
#17 0x0000555558035c2d in OS_LinuxBSD::run (this=0x7fffffffdbf0) at platform/linuxbsd/os_linuxbsd.cpp:900
#18 0x000055555802c274 in main (argc=3, argv=0x7fffffffe1b8) at platform/linuxbsd/godot_linuxbsd.cpp:74
(gdb) p dst
$1 = 0x7fff27ffe020 U""
(gdb) p str_size
$2 = 1093019370
(gdb) p dst[str_size]
Cannot access memory at address 0x80002c987bc8
(gdb) p _cowdata.size()
$3 = 1093019371
```

After some manual binary searching 0x7ffff7fff000 seems to pop up...
```
(gdb) p dst[872416247]
$35 = 0 U'\000'
(gdb) p dst[872416248]
Cannot access memory at address 0x7ffff7fff000
```

### Steps to reproduce

```
extends Node3D

func _ready():
	print("Loading...")
	var data = FileAccess.get_file_as_string("/path/to/out.json")
	print("Loaded")
	print(len(data))


# Called every frame. 'delta' is the elapsed time since the previous frame.
func _process(delta):
	pass
```

In case the specific file is needed it's available [here](https://april.sapples.net/out.json) (warning: it's big)

-

reduz:
Fixes a lot of bugs, please help me fill the list.
```
< I welcome the 64 bits overlords >
 ---------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

After this is merged, other areas of the engine can be fixed, such as Image, which also uses 32 bits offsets and causes crashes in some reasonable scenarios (such as baking lightmaps), but it depends on this one. Serialization also may need to be fixed.

*Bugsquad edit:*
- Fixes #62585
- Fixes #77256
- Fixes #80533

### DIFF
--- a/core/templates/cowdata.h
+++ b/core/templates/cowdata.h
@@ -46,7 +46,7 @@ class CharString;
 template <class T, class V>
 class VMap;
 
-SAFE_NUMERIC_TYPE_PUN_GUARANTEES(uint32_t)
+SAFE_NUMERIC_TYPE_PUN_GUARANTEES(uint64_t)
 
 // Silence a false positive warning (see GH-52119).
 #if defined(__GNUC__) && !defined(__clang__)
@@ -64,45 +64,71 @@ class CowData {
 	template <class TV, class VV>
 	friend class VMap;
 
+public:
+	typedef int64_t Size;
+	typedef uint64_t USize;
+	static constexpr USize MAX_INT = INT64_MAX;
+
 private:
+	// Function to find the next power of 2 to an integer.
+	static _FORCE_INLINE_ USize next_po2(USize x) {
+		if (x == 0) {
+			return 0;
+		}
+
+		--x;
+		x |= x >> 1;
+		x |= x >> 2;
+		x |= x >> 4;
+		x |= x >> 8;
+		x |= x >> 16;
+		if (sizeof(USize) == 8) {
+			x |= x >> 32;
+		}
+
+		return ++x;
+	}
+
+	static constexpr USize ALLOC_PAD = sizeof(USize) * 2; // For size and atomic refcount.
+
 	mutable T *_ptr = nullptr;
 
 	// internal helpers
 
-	_FORCE_INLINE_ SafeNumeric<uint32_t> *_get_refcount() const {
+	_FORCE_INLINE_ SafeNumeric<USize> *_get_refcount() const {
 		if (!_ptr) {
 			return nullptr;
 		}
 
-		return reinterpret_cast<SafeNumeric<uint32_t> *>(_ptr) - 2;
+		return reinterpret_cast<SafeNumeric<USize> *>(_ptr) - 2;
 	}
 
-	_FORCE_INLINE_ uint32_t *_get_size() const {
+	_FORCE_INLINE_ USize *_get_size() const {
 		if (!_ptr) {
 			return nullptr;
 		}
 
-		return reinterpret_cast<uint32_t *>(_ptr) - 1;
+		return reinterpret_cast<USize *>(_ptr) - 1;
 	}
 
-	_FORCE_INLINE_ size_t _get_alloc_size(size_t p_elements) const {
-		return next_power_of_2(p_elements * sizeof(T));
+	_FORCE_INLINE_ USize _get_alloc_size(USize p_elements) const {
+		return next_po2(p_elements * sizeof(T));
 	}
 
-	_FORCE_INLINE_ bool _get_alloc_size_checked(size_t p_elements, size_t *out) const {
+	_FORCE_INLINE_ bool _get_alloc_size_checked(USize p_elements, USize *out) const {
 		if (unlikely(p_elements == 0)) {
 			*out = 0;
 			return true;
 		}
-#if defined(__GNUC__)
-		size_t o;
-		size_t p;
+#if defined(__GNUC__) && defined(IS_32_BIT)
+		USize o;
+		USize p;
 		if (__builtin_mul_overflow(p_elements, sizeof(T), &o)) {
 			*out = 0;
 			return false;
 		}
-		*out = next_power_of_2(o);
-		if (__builtin_add_overflow(o, static_cast<size_t>(32), &p)) {
+		*out = next_po2(o);
+		if (__builtin_add_overflow(o, static_cast<USize>(32), &p)) {
 			return false; // No longer allocated here.
 		}
 #else
@@ -116,7 +142,7 @@ private:
 	void _unref(void *p_data);
 	void _ref(const CowData *p_from);
 	void _ref(const CowData &p_from);
-	uint32_t _copy_on_write();
+	USize _copy_on_write();
 
 public:
 	void operator=(const CowData<T> &p_from) { _ref(p_from); }
@@ -130,8 +156,8 @@ public:
 		return _ptr;
 	}
 
-	_FORCE_INLINE_ int size() const {
-		uint32_t *size = (uint32_t *)_get_size();
+	_FORCE_INLINE_ Size size() const {
+		USize *size = (USize *)_get_size();
 		if (size) {
 			return *size;
 		} else {
@@ -142,42 +168,42 @@ public:
 	_FORCE_INLINE_ void clear() { resize(0); }
 	_FORCE_INLINE_ bool is_empty() const { return _ptr == nullptr; }
 
-	_FORCE_INLINE_ void set(int p_index, const T &p_elem) {
+	_FORCE_INLINE_ void set(Size p_index, const T &p_elem) {
 		ERR_FAIL_INDEX(p_index, size());
 		_copy_on_write();
 		_ptr[p_index] = p_elem;
 	}
 
-	_FORCE_INLINE_ T &get_m(int p_index) {
+	_FORCE_INLINE_ T &get_m(Size p_index) {
 		CRASH_BAD_INDEX(p_index, size());
 		_copy_on_write();
 		return _ptr[p_index];
 	}
 
-	_FORCE_INLINE_ const T &get(int p_index) const {
+	_FORCE_INLINE_ const T &get(Size p_index) const {
 		CRASH_BAD_INDEX(p_index, size());
 
 		return _ptr[p_index];
 	}
 
 	template <bool p_ensure_zero = false>
-	Error resize(int p_size);
+	Error resize(Size p_size);
 
-	_FORCE_INLINE_ void remove_at(int p_index) {
+	_FORCE_INLINE_ void remove_at(Size p_index) {
 		ERR_FAIL_INDEX(p_index, size());
 		T *p = ptrw();
-		int len = size();
-		for (int i = p_index; i < len - 1; i++) {
+		Size len = size();
+		for (Size i = p_index; i < len - 1; i++) {
 			p[i] = p[i + 1];
 		}
 
 		resize(len - 1);
 	}
 
-	Error insert(int p_pos, const T &p_val) {
+	Error insert(Size p_pos, const T &p_val) {
 		ERR_FAIL_INDEX_V(p_pos, size() + 1, ERR_INVALID_PARAMETER);
 		resize(size() + 1);
-		for (int i = (size() - 1); i > p_pos; i--) {
+		for (Size i = (size() - 1); i > p_pos; i--) {
 			set(i, get(i - 1));
 		}
 		set(p_pos, p_val);
@@ -185,9 +211,9 @@ public:
 		return OK;
 	}
 
-	int find(const T &p_val, int p_from = 0) const;
-	int rfind(const T &p_val, int p_from = -1) const;
-	int count(const T &p_val) const;
+	Size find(const T &p_val, Size p_from = 0) const;
+	Size rfind(const T &p_val, Size p_from = -1) const;
+	Size count(const T &p_val) const;
 
 	_FORCE_INLINE_ CowData() {}
 	_FORCE_INLINE_ ~CowData();
@@ -200,7 +226,7 @@ void CowData<T>::_unref(void *p_data) {
 		return;
 	}
 
-	SafeNumeric<uint32_t> *refc = _get_refcount();
+	SafeNumeric<USize> *refc = _get_refcount();
 
 	if (refc->decrement() > 0) {
 		return; // still in use
@@ -208,35 +234,36 @@ void CowData<T>::_unref(void *p_data) {
 	// clean up
 
 	if (!std::is_trivially_destructible<T>::value) {
-		uint32_t *count = _get_size();
+		USize *count = _get_size();
 		T *data = (T *)(count + 1);
 
-		for (uint32_t i = 0; i < *count; ++i) {
+		for (USize i = 0; i < *count; ++i) {
 			// call destructors
 			data[i].~T();
 		}
 	}
 
 	// free mem
-	Memory::free_static((uint8_t *)p_data, true);
+	Memory::free_static(((uint8_t *)p_data) - ALLOC_PAD, false);
 }
 
 template <class T>
-uint32_t CowData<T>::_copy_on_write() {
+typename CowData<T>::USize CowData<T>::_copy_on_write() {
 	if (!_ptr) {
 		return 0;
 	}
 
-	SafeNumeric<uint32_t> *refc = _get_refcount();
+	SafeNumeric<USize> *refc = _get_refcount();
 
-	uint32_t rc = refc->get();
+	USize rc = refc->get();
 	if (unlikely(rc > 1)) {
 		/* in use by more than me */
-		uint32_t current_size = *_get_size();
+		USize current_size = *_get_size();
 
-		uint32_t *mem_new = (uint32_t *)Memory::alloc_static(_get_alloc_size(current_size), true);
+		USize *mem_new = (USize *)Memory::alloc_static(_get_alloc_size(current_size) + ALLOC_PAD, false);
+		mem_new += 2;
 
-		new (mem_new - 2) SafeNumeric<uint32_t>(1); //refcount
+		new (mem_new - 2) SafeNumeric<USize>(1); //refcount
 		*(mem_new - 1) = current_size; //size
 
 		T *_data = (T *)(mem_new);
@@ -246,7 +273,7 @@ uint32_t CowData<T>::_copy_on_write() {
 			memcpy(mem_new, _ptr, current_size * sizeof(T));
 
 		} else {
-			for (uint32_t i = 0; i < current_size; i++) {
+			for (USize i = 0; i < current_size; i++) {
 				memnew_placement(&_data[i], T(_ptr[i]));
 			}
 		}
@@ -261,10 +288,10 @@ uint32_t CowData<T>::_copy_on_write() {
 
 template <class T>
 template <bool p_ensure_zero>
-Error CowData<T>::resize(int p_size) {
+Error CowData<T>::resize(Size p_size) {
 	ERR_FAIL_COND_V(p_size < 0, ERR_INVALID_PARAMETER);
 
-	int current_size = size();
+	Size current_size = size();
 
 	if (p_size == current_size) {
 		return OK;
@@ -278,27 +305,29 @@ Error CowData<T>::resize(int p_size) {
 	}
 
 	// possibly changing size, copy on write
-	uint32_t rc = _copy_on_write();
+	USize rc = _copy_on_write();
 
-	size_t current_alloc_size = _get_alloc_size(current_size);
-	size_t alloc_size;
+	USize current_alloc_size = _get_alloc_size(current_size);
+	USize alloc_size;
 	ERR_FAIL_COND_V(!_get_alloc_size_checked(p_size, &alloc_size), ERR_OUT_OF_MEMORY);
 
 	if (p_size > current_size) {
 		if (alloc_size != current_alloc_size) {
 			if (current_size == 0) {
 				// alloc from scratch
-				uint32_t *ptr = (uint32_t *)Memory::alloc_static(alloc_size, true);
+				USize *ptr = (USize *)Memory::alloc_static(alloc_size + ALLOC_PAD, false);
+				ptr += 2;
 				ERR_FAIL_NULL_V(ptr, ERR_OUT_OF_MEMORY);
 				*(ptr - 1) = 0; //size, currently none
-				new (ptr - 2) SafeNumeric<uint32_t>(1); //refcount
+				new (ptr - 2) SafeNumeric<USize>(1); //refcount
 
 				_ptr = (T *)ptr;
 
 			} else {
-				uint32_t *_ptrnew = (uint32_t *)Memory::realloc_static(_ptr, alloc_size, true);
+				USize *_ptrnew = (USize *)Memory::realloc_static(((uint8_t *)_ptr) - ALLOC_PAD, alloc_size + ALLOC_PAD, false);
 				ERR_FAIL_NULL_V(_ptrnew, ERR_OUT_OF_MEMORY);
-				new (_ptrnew - 2) SafeNumeric<uint32_t>(rc); //refcount
+				_ptrnew += 2;
+				new (_ptrnew - 2) SafeNumeric<USize>(rc); //refcount
 
 				_ptr = (T *)(_ptrnew);
 			}
@@ -307,7 +336,7 @@ Error CowData<T>::resize(int p_size) {
 		// construct the newly created elements
 
 		if (!std::is_trivially_constructible<T>::value) {
-			for (int i = *_get_size(); i < p_size; i++) {
+			for (Size i = *_get_size(); i < p_size; i++) {
 				memnew_placement(&_ptr[i], T);
 			}
 		} else if (p_ensure_zero) {
@@ -319,16 +348,17 @@ Error CowData<T>::resize(int p_size) {
 	} else if (p_size < current_size) {
 		if (!std::is_trivially_destructible<T>::value) {
 			// deinitialize no longer needed elements
-			for (uint32_t i = p_size; i < *_get_size(); i++) {
+			for (USize i = p_size; i < *_get_size(); i++) {
 				T *t = &_ptr[i];
 				t->~T();
 			}
 		}
 
 		if (alloc_size != current_alloc_size) {
-			uint32_t *_ptrnew = (uint32_t *)Memory::realloc_static(_ptr, alloc_size, true);
+			USize *_ptrnew = (USize *)Memory::realloc_static(((uint8_t *)_ptr) - ALLOC_PAD, alloc_size + ALLOC_PAD, false);
 			ERR_FAIL_NULL_V(_ptrnew, ERR_OUT_OF_MEMORY);
-			new (_ptrnew - 2) SafeNumeric<uint32_t>(rc); //refcount
+			_ptrnew += 2;
+			new (_ptrnew - 2) SafeNumeric<USize>(rc); //refcount
 
 			_ptr = (T *)(_ptrnew);
 		}
@@ -340,14 +370,14 @@ Error CowData<T>::resize(int p_size) {
 }
 
 template <class T>
-int CowData<T>::find(const T &p_val, int p_from) const {
-	int ret = -1;
+typename CowData<T>::Size CowData<T>::find(const T &p_val, Size p_from) const {
+	Size ret = -1;
 
 	if (p_from < 0 || size() == 0) {
 		return ret;
 	}
 
-	for (int i = p_from; i < size(); i++) {
+	for (Size i = p_from; i < size(); i++) {
 		if (get(i) == p_val) {
 			ret = i;
 			break;
@@ -358,8 +388,8 @@ int CowData<T>::find(const T &p_val, int p_from) const {
 }
 
 template <class T>
-int CowData<T>::rfind(const T &p_val, int p_from) const {
-	const int s = size();
+typename CowData<T>::Size CowData<T>::rfind(const T &p_val, Size p_from) const {
+	const Size s = size();
 
 	if (p_from < 0) {
 		p_from = s + p_from;
@@ -368,7 +398,7 @@ int CowData<T>::rfind(const T &p_val, int p_from) const {
 		p_from = s - 1;
 	}
 
-	for (int i = p_from; i >= 0; i--) {
+	for (Size i = p_from; i >= 0; i--) {
 		if (get(i) == p_val) {
 			return i;
 		}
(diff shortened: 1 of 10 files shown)


Mickeon:
Would fix https://github.com/godotengine/godot/issues/45296 and https://github.com/godotengine/godot/issues/54679 for sure.

AThousandShips:
Note that this requires `_get_alloc_size` to be updated unless we still treat this as having a capped size in the 32 bit range, as it only handles 32 bits, `next_power_of_2` needs to be updated

Also needs a godot-cpp companion

core/templates/cowdata.h
```diff
@@ -200,43 +202,44 @@ void CowData<T>::_unref(void *p_data) {
...
 
-	uint32_t rc = refc->get();
+	uint64_t rc = refc->get();
 	if (unlikely(rc > 1)) {
 		/* in use by more than me */
-		uint32_t current_size = *_get_size();
+		uint64_t current_size = *_get_size();
 
-		uint32_t *mem_new = (uint32_t *)Memory::alloc_static(_get_alloc_size(current_size), true);
+		uint64_t *mem_new = (uint64_t *)Memory::alloc_static(_get_alloc_size(current_size) + ALLOC_PAD, false);
```
RandomShaper, line 239: `Memory::alloc_static()` adds a 16-byte pre-padding, of which the last 8 bytes are availale for "users". When `CowData` was 32-bit, those 8 bytes were enough to store the two header `uint32_t`s. With this PR, `CowData` is adding 16 more bytes, which gives a total of 24 extra bytes. Assuming the ref count has to become 64-bit as well, we're wasting 8 bytes.

Possible options:
- a) Redefine `PAD_ALIGN` in `memory.h` to `24` and remove the newly added `ALLOC_PAD` in `CowData`.
- b) Redefine `PAD_ALIGN` in `memory.h` to `8` (assumming the extra bytes aren't used elsewhere) and let `CowData` deal with all the pre-padding (i.e., keep the current `ALLOC_PAD` of `16`).

core/templates/cowdata.h
```diff
@@ -65,24 +65,26 @@ class CowData {
...
+	_FORCE_INLINE_ uint64_t *_get_size() const {
 		if (!_ptr) {
 			return nullptr;
 		}
 
-		return reinterpret_cast<uint32_t *>(_ptr) - 1;
+		return reinterpret_cast<uint64_t *>(_ptr) - 1;
 	}
 
 	_FORCE_INLINE_ size_t _get_alloc_size(size_t p_elements) const {
```
AThousandShips, line 90: I'd suggest switching to `uint64_t` here as `size_t` isn't guaranteed to be 64 bits though it usually is, better stay safe IMO

RandomShaper (commented):
Looks good overall.

I'm wondering if this is neglecting 32-bit builds (they don't need or won't support 64-bit stuff such as > 3 GiB allocations and maybe not even 64-bit atomics) and so `size_t` should be used intead.

**UPDATE:** Per @AThousandShips, `size_t` won't necessarilly be 64-bit on 64-bit targets, so maybe something else that matches `sizeof(void *)`, but you get the idea.

core/templates/cowdata.h
AThousandShips: Only on debug builds, it doesn't add the padding otherwise, as the second parameter is false

AThousandShips: Note that the 8 bytes are used already, forget where but I've seen it, will look

reduz: Yes I was going to mention this, it only adds 16 bytes on debug builds, but not on release, where it remains 16.

RandomShaper: I missed the argument has been changed to `false`. So, yeah, that's OK. Let's just find out (as a separate endeavor) if the extra pre-padding from `memory.h` is still needed somewhere.

AThousandShips: It is, in `memnew_arr` 🙂

core/templates/cowdata.h
reduz: I ended up using a typedef for the size, as suggested.

reduz:
I changed to use a typedef for the size, depending on compiler, so it remains 32 bits on 32 bit platforms.

reduz:
@AThousandShips regarding GodotCPP, I am not certain if something has to change on that side.

AThousandShips:
It does, it redeclares the type, unless we're okay with lack of parity over there

core/templates/vector.h
```diff
@@ -157,40 +158,40 @@ class Vector {
 		return ret;
 	}
 
-	Vector<T> slice(int p_begin, int p_end = INT_MAX) const {
+	Vector<T> slice(Size p_begin, Size p_end = INT64_MAX) const {
```
AThousandShips, line 161: Need a macro for the max as well, I'd suggest a `SIZE_MAX/MIN` and `USIZE_MAX/MIN` though might clash with definitions existing already

reduz:
@AThousandShips ah I completely forgot we strive to use the same syntax in GodotCPP than in Godot modules.. Will have to ask @bruvzg to lend a hand changing the code there once this one is merged.

So I guess this PR is "kind of" compat breaking.. though it likely only affects GodotCPP, not the rest of the extensions.

core/typedefs.h
```diff
@@ -139,9 +139,22 @@ inline void __swap_tmpl(T &x, T &y) {
...
+/* GCC, Clang */
+#if defined(__i386__) || defined(__ILP32__) || (defined(__SIZEOF_POINTER__) && (__SIZEOF_POINTER__ == 4))
+#define IS_32_BIT
+#endif
+#elif defined(_MSC_VER)
+/* Microsoft Visual C++ */
+#if defined(_M_IX86)
+#define IS_32_BIT
+#endif
+#endif
```
akien-mga, line 152: Sounds like a pretty old stackoverflow answer :)

Needs thorough review to do it right. I doubt we care about `__ILP32__`, but we do care about 32-bit ARM.

CC @bruvzg

core/typedefs.h
```diff
@@ -139,9 +139,22 @@ inline void __swap_tmpl(T &x, T &y) {
...
+
 /* Functions to handle powers of 2 and shifting. */
 
 // Function to find the next power of 2 to an integer.
+
```
akien-mga, line 157: Why? :) The comment directly qualifies the method under it.

core/typedefs.h
reduz: I checked it and it should be alright, basically anything GCC/LLVM uses __SIZEOF_POINTER__ anyway, so we are safe (and this also contemplates emscripten).

core/templates/cowdata.h
```diff
@@ -64,45 +64,74 @@ class CowData {
 	template <class TV, class VV>
 	friend class VMap;
 
+public:
+#ifdef IS_32_BIT
+	typedef int32_t Size;
+	typedef uint32_t USize;
+#else
+	typedef int64_t Size;
+	typedef uint64_t USize;
```
akien-mga, line 73: Might need more specific names, I'm worried about conflicts with potential Windows or macOS global defines.

reduz: It's in the local class context so it should be fine as long as it passes CI.

core/typedefs.h
reduz: oops, sorry

core/templates/cowdata.h
bruvzg: Is this extra complexity really necessary. 32-bit systems won't benefit for it, but using 64-bit ints should work fine on 32-bit systems (as long as it's not cast to `void*` directly).

TokisanGames:
Thank God. I haven't tested yet, but will link up related issues. Lack of 64-bit resources is a problem for Terrain3D users with big terrains.

So hopefully this will fix our issue, crashing when saving large terrains in a >1gb resource file. 
https://github.com/TokisanGames/Terrain3D/issues/159

And Godot issue https://github.com/godotengine/godot/issues/62585

Possibility superseding or related to PRs 
https://github.com/godotengine/godot/pull/74582
https://github.com/godotengine/godot/pull/77306

core/typedefs.h
bruvzg: 32-bit ARM is still supported for Android (not sure if we do official builds, but the option is still there) as well as for Windows (completely useless in practice).

core/templates/cowdata.h
bruvzg: And if we want to do it this way for some reason, why not use `size_t` and `ptrdiff_t` directly?

core/typedefs.h
Calinou: WebAssembly is also 32-bit as of writing :slightly_smiling_face:

AThousandShips: The CI will tell in fact!

RandomShaper: Couldn't this be solved with some compile-time magic? Some SFINAE, or a more modern approach, involving pointer size and type aliasing.

reduz: As I said, anything other than Windows is either GCC or LLVM which uses __SIZEOF_POINTER__, so this is already solved problem.

core/templates/cowdata.h
reduz: I don't mind leaving it 64 bits even for 32 bits systems.

core/templates/cowdata.h
```diff
@@ -64,45 +64,74 @@ class CowData {
...
+		}
+
+		--x;
+		x |= x >> 1;
+		x |= x >> 2;
+		x |= x >> 4;
+		x |= x >> 8;
+		x |= x >> 16;
+		if (sizeof(USize) == 8) {
+			x |= x >> 32;
```
AThousandShips, line 89: This leaves the maximum size as `2^33`, is this intended as a limit to not go too big or an oversight?

AThousandShips:
While changing the size values of `String` might be risky because it's exposed (granted so is `Vector`, indirectly), `VMap` and `VSet` uses `Vector` internally, should we unify all data structures using these sizes at least the non-exposed ones?

AThousandShips:
As long as the iteration cases aren't broken it should be safe enough

core/templates/cowdata.h
```diff
@@ -64,45 +64,71 @@ class CowData {
...
+		}
+
+		return ++x;
+	}
+
+	static constexpr USize ALLOC_PAD = 16; // Always 16 bye alignment.
```
AThousandShips, line 92:
```suggestion
	static constexpr USize ALLOC_PAD = 16; // Always 16 byte alignment.
```

AThousandShips (approved):
The code looks good to me, the sizing checks should work correctly with the new po2 method and my previous fix for the resize should still handle that just with a new range

Needs to ensure that any iteration being done isn't broken, might need a fix of any `for (int...)` loops using `Vector` so we don't get broken data etc., as part of the fixes involved in this are for cases where data does run out

Otherwise looks good, if everything else works right

reduz:
This is now ready to merge. 

It is to note that this PR will break GodotCPP GDextension compatibility, yet again.

This is an unfortunate consequence of the fact that GodotCPP accesses internals it should technically not for the sake of being compatible with modules. We should maybe consider having a different set of bindings, or do some things in GodotCPP different to avoid this scenario.

akien-mga:
There's a MSVC warning to fix:
```
Error: .\core/templates/ring_buffer.h(200): error C2220: the following warning is treated as an error
Warning: .\core/templates/ring_buffer.h(200): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)
```

---

And the "Test project converter" in the C# build fails with a C# stacktrace:
https://github.com/godotengine/godot/actions/runs/7472530025/job/20334943529?pr=86730

But I don't know if it's actually due to a C# issue, or just a Godot error that would happen on a non-dotnet build with just a different error (that project converter test is only run on the "Editor w/ Mono" Linux build).

The test is just this, so this can be attempted locally to see if the error is reproducible:
```
mkdir converter_test
cd converter_test
touch project.godot
godot --headless --validate-conversion-3to4
cd ..
rm converter_test -rf
```

adamscott:
cc. @Bromeon @dsnopek

AThousandShips:
Would suggest running with `-k` on windows if you can to catch all the cases as there's likely many

reduz:
I haven't used MSVC to compile in an eternity, will check during the week.

dsnopek:
> It is to note that this PR will break GodotCPP GDextension compatibility, yet again.
>
> This is an unfortunate consequence of the fact that GodotCPP accesses internals it should technically not for the sake of being compatible with modules. We should maybe consider having a different set of bindings, or do some things in GodotCPP different to avoid this scenario.

I don't think this is exactly true.

It will only break godot-cpp, if godot-cpp is interacting with a `String` or `Packed*Array` or similar, that has more items than a 32-bit index can access. It isn't actually accessing any internals that it shouldn't, it just uses implementations of `CowData` and `Vector` etc that are derived from the code for the same classes in Godot itself, so they suffer from the same limitations.

I think with bruvzg's godot-cpp PR https://github.com/godotengine/godot-cpp/pull/1357, we should be good going forward. And I think we can cherrypick it for the 4.2 and 4.1 versions of godot-cpp.

But if a GDExtension compiled with an older godot-cpp is used with a newer Godot, AND it's passed a `String` or `Packed*Array` with too many items, then there could be some breakage. However, this feels like a small enough edge case to probably be acceptable?

raulsntos:
To the extent that this changes the memory layout of Variant types, it has an impact on C# interop structs.

In C# we define interop structs that must match the memory layout of the C++ types so they can be reinterpreted and consumed from C#. For example, we get the size of a Godot String like this:

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/InteropStructs.cs#L477

This is basically copied from the C++ code (`CowData::get_size`):

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/core/templates/cowdata.h#L85

But this has changed in this PR since now the size is 64-bit, so it needs to be updated in C# too.

As long as the changes are kept internally, it should not break compatibility but changing the user-facing API would be a problem. And if any of these types actually has a size that doesn't fit in a 32-bit `int` then it will break C#[^1].

<details><summary>Here's a somewhat naive diff that should fix the stackoverflow you are getting on CI:</summary>

```diff
diff --git a/modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/InteropStructs.cs b/modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/InteropStructs.cs
index d5d9404ed1b..c806263edb5 100644
--- a/modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/InteropStructs.cs
+++ b/modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/InteropStructs.cs
@@ -474,7 +474,7 @@ namespace Godot.NativeInterop
         public readonly unsafe int Size
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => _ptr != IntPtr.Zero ? *((int*)_ptr - 1) : 0;
+            get => _ptr != IntPtr.Zero ? (int)(*((ulong*)_ptr - 1)) : 0;
         }
     }
 
@@ -725,7 +725,7 @@ namespace Godot.NativeInterop
             public readonly unsafe int Size
             {
                 [MethodImpl(MethodImplOptions.AggressiveInlining)]
-                get => _ptr != null ? *((int*)_ptr - 1) : 0;
+                get => _ptr != null ? (int)(*((ulong*)_ptr - 1)) : 0;
             }
         }
 
@@ -875,7 +875,7 @@ namespace Godot.NativeInterop
         public readonly unsafe int Size
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => _ptr != null ? *((int*)_ptr - 1) : 0;
+            get => _ptr != null ? (int)(*((ulong*)_ptr - 1)) : 0;
         }
     }
 
@@ -939,7 +939,7 @@ namespace Godot.NativeInterop
         public readonly unsafe int Size
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => _ptr != null ? *((int*)_ptr - 1) : 0;
+            get => _ptr != null ? (int)(*((ulong*)_ptr - 1)) : 0;
         }
     }
 
@@ -971,7 +971,7 @@ namespace Godot.NativeInterop
         public readonly unsafe int Size
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => _ptr != null ? *((int*)_ptr - 1) : 0;
+            get => _ptr != null ? (int)(*((ulong*)_ptr - 1)) : 0;
         }
     }
 
@@ -1003,7 +1003,7 @@ namespace Godot.NativeInterop
         public readonly unsafe int Size
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => _ptr != null ? *((int*)_ptr - 1) : 0;
+            get => _ptr != null ? (int)(*((ulong*)_ptr - 1)) : 0;
         }
     }
 
@@ -1035,7 +1035,7 @@ namespace Godot.NativeInterop
         public readonly unsafe int Size
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => _ptr != null ? *((int*)_ptr - 1) : 0;
+            get => _ptr != null ? (int)(*((ulong*)_ptr - 1)) : 0;
         }
     }
 
@@ -1067,7 +1067,7 @@ namespace Godot.NativeInterop
         public readonly unsafe int Size
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => _ptr != null ? *((int*)_ptr - 1) : 0;
+            get => _ptr != null ? (int)(*((ulong*)_ptr - 1)) : 0;
         }
     }
 
@@ -1099,7 +1099,7 @@ namespace Godot.NativeInterop
         public readonly unsafe int Size
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => _ptr != null ? *((int*)_ptr - 1) : 0;
+            get => _ptr != null ? (int)(*((ulong*)_ptr - 1)) : 0;
         }
     }
 
@@ -1131,7 +1131,7 @@ namespace Godot.NativeInterop
         public readonly unsafe int Size
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => _ptr != null ? *((int*)_ptr - 1) : 0;
+            get => _ptr != null ? (int)(*((ulong*)_ptr - 1)) : 0;
         }
     }
 

```

</details>

[^1]: More context on this: https://github.com/dotnet/runtime/issues/12221

reduz:
@raulsntos Yeah I guess the Mono bindings should not really be accessing these internals. We have explicit functions exposed in the extension API now to avoid these situations, so I suppose it would be just a matter to move to that eventually (when Mono is moved to extension)

raulsntos:
I don't want to get too off-topic, but I don't agree that the move to GDExtensions would help here. The API exposed to retrieve this data would have a bigger cost than accessing the internals directly as we are doing today so we wouldn't change that in the new system.

And to be honest that's not the problem, we just need to update the C# internals to match the C++ internals (similar to how godot-cpp is doing). I shared a diff that does just that and doesn't break compatibility since it doesn't change publicly exposed API.

The main issue is that C# types like `System.String` or arrays define their length as 32-bit integers so they can't possible work with data that is bigger than what can be represented with a `int32_t`[^1].

To put it in simple terms, if we have a `String` coming from Godot that has more items than what can fit in a 32-bit length then we wouldn't be able to marshal that into a C# `String`.

[^1]: https://github.com/dotnet/runtime/issues/12221

dsnopek:
> And to be honest that's not the problem, we just need to update the C# internals to match the C++ internals (similar to how godot-cpp is doing).

Just to be clear (although, it probably isn't necessary :-)) this isn't what the godot-cpp change is doing. godot-cpp _doesn't_ directly access Godot's internals, it always goes through the GDExtension interface.

The change to godot-cpp is doing two things:

1. Updating the C++ classes that exist entirely on the godot-cpp side to account for the possibility that the data on the Godot side could have a large enough number of elements that it requires a 64-bit integer to count/index them
2. Updating the equivalent `CowData` and `Vector` in godot-cpp to match Godot, so that you can use them in your GDExtension exactly the same as you would in Godot itself, removing the 32-bit limitations of the previous implementation. This doesn't affect interop with Godot, this is for where these templates are used purely on the godot-cpp side.

akien-mga:
> Fixes a lot of bugs, please help me fill the list.

Trying to review some of the ones mentioned in this PR, linked issues, or some I could find searching for references to CowData and 32/64 bits.

- #45296

`3.x` issue so it won't be solved by this (and likely won't ever be solved).

- #48399

Reproducible in `master`. This PR doesn't fix it because BitMap now has guards against using a size bigger than `INT32_MAX`. This PR should however make it possible to refactor BitMap to use `int64_t`.

- #51154

Not fixed by this PR. The stacktrace changes from `CowData<_AtlasWorkRectResult>::get(int)` to `CowData<_AtlasWorkRectResult>::get(long)` but it's still crashing. Likely another one that _could_ be fixed now as a follow-up.

- #54679

Seems related indeed, but I tested and this PR does _not_ solve the issue. It still crashes in the same place.
Relevant Image code is still using `int` for sizes so I assume it needs further work after this PR to actually fix the crash.
```
[3] Image::initialize_data(int, int, bool, Image::Format) (/home/akien/Projects/godot/godot.git/./core/io/image.cpp:2187 (discriminator 4))
```

- #62585

Tested with the MRP from https://github.com/godotengine/godot/issues/62585#issuecomment-1450900291, this fixes it.
I don't get a crash however before this PR in a 2 weeks old `masterbuild, but I do get a `p_size < 0` error which is now fixed.

- #77256

Fixed by this PR.

- #80533

Fixed by this PR.

akien-mga:
Diff that should make the PR pass CI (two more MSVC warning fixes, and @raulsntos' C# changes):
https://github.com/akien-mga/godot/commit/6eff4f655343d2e0a917f75eba1dcafee713cfd0

akien-mga (approved):
Tested locally, seems to work well!

As I pointed out, further work will be needed at least for Image and BitMap to make their size handling 64-bit too. Possibly other classes based on CowData internally.

akien-mga:
Thanks!

PR #86732, opened 2024-01-03
Make `LineEdit` scrolling show as much text as possible

### LINKED ISSUE #27420: LineEdit and TextEdit improvements when removing text
Hello, this is a suggestion for some LineEdit improvements when removing text. The main goal is to achieve a behavior that is more common across similar controls in other programs.

As a comparison, this is what the LineEdit looks like in Godot:

https://github.com/godotengine/godot/assets/180032/24cd377e-b179-4b1b-b4e2-b1676dbe6ec9

And this is the spotify searchbar:

https://github.com/godotengine/godot/assets/180032/05f6792e-4943-4475-a1a5-802ea98e38d0

But why do I think that the behavior of the spotify searchbar is better?
- presents as much information to the user as possible, without any loss
- characters aren't either drawn or not drawn, but instead can be partially drawn, which prevents the caret from jumping around when removing text

LineEdits always bothered me because of this.

This applies to TextEdit as well.

### LINKED ISSUE #27421: LineEdit bug when removing text with align=right/center
**Godot version:**
3.1-stable

**Issue description:**
When removing text from a LineEdit that has its alignment set to "Center" or "Right" and contains more text than can be shown, the caret starts acting weird. This gif shows it:

As you can see, the text is not being aligned properly and too few characters are drawn. With alignment set to "Center", it behaves like alignment set to "Left", then suddenly jumps back to the correct alignment when only one character remains.

**Steps to reproduce:**
Delete text from a full LineEdit with "Align" set to "Center" or "Right" by pressing backspace.

**Edit:
This is also happening with align set to "Center", the offset is wrong until theres only one remaining character.**

TheSofox:
Fixes #27420
*Bugsquad edit:* Fixes #27421

This PR makes sure the scrolling in `LineEdit` shows as much text as possible, such as when you delete text after entering more than the textbox can fit. Previously, it wouldn't update the scrolling on deleting text, now it does. See linked issue for video examples.

### DIFF
--- a/scene/gui/line_edit.cpp
+++ b/scene/gui/line_edit.cpp
@@ -1713,6 +1713,12 @@ void LineEdit::set_caret_column(int p_column) {
 	} else if (MAX(primary_caret_offset.x, primary_caret_offset.y) >= ofs_max) {
 		scroll_offset += ofs_max - MAX(primary_caret_offset.x, primary_caret_offset.y);
 	}
+
+	// Scroll to show as much text as possible
+	if (text_width + scroll_offset + x_ofs < ofs_max) {
+		scroll_offset = ofs_max - x_ofs - text_width;
+	}
+
 	scroll_offset = MIN(0, scroll_offset);
 
 	queue_redraw();


TheSofox:
Apparently my PR also fixes #27421. Quick test seems to indicate "Center" and "Right" alignment behave much better with this change.

Geometror (approved):
Tested with all align modes, works nicely. I can confirm that this fixes #27421 and most of #27420. (the "jumping" cursor is still present, but fixing that would require implementing clipping which is should be a toggleable feature IMO and is definitely out of scope here)

Although not absolutely necessary, you could add a comment describing what the code snippet does/for what it is needed since it's not that obvious (as part of the ongoing effort to make the code base more maintainable).

TheSofox:
Amended commit with additional comment. Thanks for the review and approval.

akien-mga:
Thanks!

scene/gui/line_edit.cpp
```diff
@@ -1713,6 +1713,12 @@ void LineEdit::set_caret_column(int p_column) {
 	} else if (MAX(primary_caret_offset.x, primary_caret_offset.y) >= ofs_max) {
 		scroll_offset += ofs_max - MAX(primary_caret_offset.x, primary_caret_offset.y);
 	}
+
+	// Scroll to show as much text as possible
```
KoBeWi, line 1717:
```suggestion
	// Scroll to show as much text as possible.
```
:P

TheSofox: Well, I don't know how to amend a commit after the PR has been merged. Not even sure I've permission.

KoBeWi: It's not possible, you'd need to open a new PR.
Though such random typo fixes are usually done in batches, so it's just something to remember for later.

TheSofox: Cool, I'll keep it in mind. Thanks for the heads up!

PR #86742, opened 2024-01-03
Add C# examples in `Basis.xml`

paulloz:
Follow up to #86664

### DIFF
--- a/doc/classes/Basis.xml
+++ b/doc/classes/Basis.xml
@@ -71,12 +71,20 @@
 			<param index="1" name="order" type="int" default="2" />
 			<description>
 				Constructs a pure rotation Basis matrix from Euler angles in the specified Euler rotation order. By default, use YXZ order (most common). See the [enum EulerOrder] enum for possible values.
-				[codeblock]
+				[codeblocks]
+				[gdscript]
 				# Creates a Basis whose z axis points down.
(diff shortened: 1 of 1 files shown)


doc/classes/Basis.xml
```diff
@@ -72,10 +72,18 @@
 			<description>
 				Constructs a pure rotation Basis matrix from Euler angles in the specified Euler rotation order. By default, use YXZ order (most common). See the [enum EulerOrder] enum for possible values.
 				[codeblock]
```
AThousandShips, line 74:
```suggestion
				[codeblocks]
```
Same everywhere

paulloz: Oops, my bad. For future reference, is running `doc/tools/doc_status.py` locally a good way to check everything is formatted correctly?

AThousandShips: Haven't tried, might give cryptic results like the CI (which doesn't tell you what's wrong in this case for example)

akien-mga: Not `doc_status.py`, but `make_rst.py` itself reports the errors. Provided you have GNU make and python locally, you can run `make rst` in the `doc` folder.

paulloz: Noted, thanks.

raulsntos (approved):
LGTM. Thanks!

akien-mga:
Thanks!

PR #86733, opened 2024-01-03
Add autocompletion options for AnimatedSprite's other play methods

Mickeon:
When **AnimatedSprite2D** & **AnimatedSprite3D**'s `play_backwards()` was added, autocompletion for it was forgotten. This PR fixes that, and also adds it for `set_animation()` and `set_autoplay()` for... completion.

PR #86743, opened 2024-01-03
Optimize comparisons for Object's `get_argument_options`

Mickeon:
You could say it's a continuation of https://github.com/godotengine/godot/pull/86729 and https://github.com/godotengine/godot/pull/86733

I noticed that some `get_argument_options()` assign their `p_function` **StringName** into a normal `pf` **String** before comparing the method's name. It turns out this is slightly better because it avoids the conversion every time, and sometimes it just... looks more readable.

This PR makes that consistent across the board and gives a purpose to a lonely, unused `pf` in **AnimationPlayer**'s method.
One could question why `get_argument_options()` wants a **StringName** if it's better to just convert it back to **String** every time, but who am I to judge?

I have a feeling this and the other previous PRs can be backported to 3.x, even(?).

### DIFF
--- a/core/object/object.cpp
+++ b/core/object/object.cpp
@@ -2086,15 +2086,17 @@ void ObjectDB::debug_objects(DebugFunc p_func) {
 	spin_lock.unlock();
 }
 
+#ifdef TOOLS_ENABLED
 void Object::get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const {
+	const String pf = p_function;
 	if (p_idx == 0) {
-		if (p_function == "connect" || p_function == "is_connected" || p_function == "disconnect" || p_function == "emit_signal" || p_function == "has_signal") {
+		if (pf == "connect" || pf == "is_connected" || pf == "disconnect" || pf == "emit_signal" || pf == "has_signal") {
 			List<MethodInfo> signals;
 			get_signal_list(&signals);
 			for (const MethodInfo &E : signals) {
 				r_options->push_back(E.name.quote());
 			}
-		} else if (p_function == "call" || p_function == "call_deferred" || p_function == "callv" || p_function == "has_method") {
+		} else if (pf == "call" || pf == "call_deferred" || pf == "callv" || pf == "has_method") {
 			List<MethodInfo> methods;
 			get_method_list(&methods);
 			for (const MethodInfo &E : methods) {
@@ -2103,7 +2105,7 @@ void Object::get_argument_options(const StringName &p_function, int p_idx, List<
 				}
 				r_options->push_back(E.name.quote());
 			}
-		} else if (p_function == "set" || p_function == "set_deferred" || p_function == "get") {
+		} else if (pf == "set" || pf == "set_deferred" || pf == "get") {
 			List<PropertyInfo> properties;
 			get_property_list(&properties);
 			for (const PropertyInfo &E : properties) {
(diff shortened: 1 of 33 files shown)


Mickeon:
As discussed in RocketChat, I'm not sure what the better approach is anymore. Doing this, having `get_argument_options` require a String, or a mix of both.

dalexeev:
I think a faster way would be to compare `StringName`s (like `p_function == CoreStringNames::get_singleton()->call`). But it is more verbose, and `SNAME` macro is not recommended to be used all the time. Could we solve this using code generation or `constexpr` evaluations in a way that doesn't negatively impact build time when adding/changing a new `StringName` to the codebase?

AThousandShips:
`constexpr` wouldn't be possible here as `StringName` isn't compile time due to hashing etc., so other than binding them centrally it would be equivalent to `SNAME` I'd say

Mickeon:
Despite this being an optimization, I am not too concerned on performance. It's an editor-exclusive method called once in a while. This is more-so done for consistency and future-proofing if and when more autocompletion options are added.

dalexeev:
Maybe an option with code generation is possible?

1. The build system reads the sources and collects all the names from `SNAME("name")` or similar macro and generates one or more `string_names.gen.h` files that are included in `.cpp`s, so that this alone will not cause a cascaded recompilation the entire engine.
2. When compiling `.cpp`, the macro is expanded to `StringNames::get_singleton()->name` or something similar.
3. Now we can always use the `SNAME` macro instead of `CoreStringNames` and similar singletons (if the name is not a variable, for this case we can keep the macro option that expands into a lambda).

The tricky part is how to split the names into multiple `.gen.h`s so that adding a new `StringName` doesn't recompile all the `.cpp` files that use `SNAME`. We can split into groups alphabetically or by area (core, scene, modules).

Mickeon:
Definitely out of the scope of this PR, but do discuss it with other people more savvy about it, because it does sound neat in theory.

KoBeWi:
I have attempted to do some "StringName optimizations" multiple times. E.g. the editor theme uses Strings everywhere for methods that take StringNames, so I though that it's inefficient and made reusable StringName variables. The result was that the code speed was the same, or even slower. It did not improve performance, at most it made binary size smaller. You can check my changes here: https://github.com/KoBeWi/godot/commit/124684afb0c43846eb0f6870c5e00b8df2d6263c
So far from all my attempts I got the same results. None of them was worth it performance-wise, though they did save a few kBs of exe size. IMO claims about StringNames being more performant are much exaggerated and optimizing their usage is a dead end not worth pursuing. I think they might matter only somewhere deep in core code.

Mickeon:
Hearing that makes my push towards **String** & **StringName** having the same methods on the outside feel all the more justified. Still, it's a bit disheartening to hear. Or, perhaps we shouldn't feel so bad, as the compiler does such a good job optimizing... strings? Who knows.

Mickeon:
I will probably update this PR to surround all of these with TOOLS_ENABLED, although the PR is good as is, too.

Mickeon:
Done what I said I'd do...


Is this really worth the review of all of those teams above?

core/object/object.cpp
```diff
@@ -2086,15 +2086,17 @@ void ObjectDB::debug_objects(DebugFunc p_func) {
 	spin_lock.unlock();
 }
 
+#ifdef TOOLS_ENABLED
 void Object::get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const {
+	String pf = p_function;
```
KoBeWi, line 2091:
```suggestion
	const String pf = p_function;
```
Same in other cases.

Mickeon: Done so

KoBeWi (approved):

Mickeon:
For your interest see also https://github.com/godotengine/godot/pull/87197

akien-mga:
Thanks!

Issue #82705, opened 2023-10-03
ATI/AMD Driver Crash while Building Shaders (3.x)


jitspoe:
### Godot version

v3.6.beta.custom_build [7d87ed371]

### System information

Windows 10, GLES3, AMD RX 6700 XT and other AMD cards driver version 23.9.3 and 23.9.1

### Issue description

In Fist of the Forgotten, I try to pre-build all the shaders, and during this process, the game crashes on pretty much every AMD card players have tested on.  I don't have one myself, so I can't directly test.  Shader compilation mode is set to synchronous.  Appears to be a driver issue.

```
 	ntdll.dll!_NtDelayExecution@8()	Unknown
 	KERNELBASE.dll!SleepEx()	Unknown
 	KERNELBASE.dll!_Sleep@4()	Unknown
>	atioglxx.dll!5d8893a4()	Unknown
 	atioglxx.dll![Frames below may be incorrect and/or missing, no symbols loaded for atioglxx.dll]	Unknown
 	atioglxx.dll!5d889a58()	Unknown
 	atioglxx.dll!5d890b8d()	Unknown
 	FistOfTheForgottenDemo.exe!VisualServerScene::_render_scene(const Transform p_cam_transform, const CameraMatrix & p_cam_projection, const int p_eye, bool p_cam_orthogonal, RID p_force_environment, RID p_scenario, RID p_shadow_atlas, RID p_reflection_probe, int p_reflection_probe_pass) Line 3325	C++
 	FistOfTheForgottenDemo.exe!VisualServerScene::render_camera(RID p_camera, RID p_scenario, Vector2 p_viewport_size, RID p_shadow_atlas) Line 2896	C++
 	FistOfTheForgottenDemo.exe!VisualServerViewport::_draw_3d(VisualServerViewport::Viewport * p_viewport, ARVRInterface::Eyes p_eye) Line 78	C++
 	FistOfTheForgottenDemo.exe!VisualServerViewport::_draw_viewport(VisualServerViewport::Viewport * p_viewport, ARVRInterface::Eyes p_eye) Line 109	C++
 	FistOfTheForgottenDemo.exe!VisualServerViewport::draw_viewports() Line 344	C++
 	FistOfTheForgottenDemo.exe!VisualServerRaster::draw(bool p_swap_buffers, double frame_step) Line 116	C++
 	FistOfTheForgottenDemo.exe!VisualServerWrapMT::draw(bool p_swap_buffers, double frame_step) Line 85	C++
[...]

PR #86744, opened 2024-01-03
[3.x] Prevent shuffling custom shader functions (shader cache requires determinism)

Listwon:
Custom shader function names are stored as `StringName` so their order in` Set` is determined by pointers comparison which change the order every run (because memory addresses change). This leads to shader compilation even with shader cache enabled and generated previously (shader text changes so it needs to be compiled as new shader in cache).

https://github.com/godotengine/godot/blob/4c4cb12e38dd862c1eedcb8ad17ca39ae9e3474e/servers/visual/shader_language.h#L605

I've replaced `Set` with `Vector` to prevent shuffling.

Related comment, not sure if it helps with the main issue https://github.com/godotengine/godot/issues/82705#issuecomment-1744108670

### DIFF
--- a/drivers/gles2/shader_compiler_gles2.cpp
+++ b/drivers/gles2/shader_compiler_gles2.cpp
@@ -229,17 +229,18 @@ void ShaderCompilerGLES2::_dump_function_deps(const SL::ShaderNode *p_node, cons
 
 	ERR_FAIL_COND(fidx == -1);
 
-	for (Set<StringName>::Element *E = p_node->functions[fidx].uses_function.front(); E; E = E->next()) {
-		if (r_added.has(E->get())) {
+	for (int ufidx = 0; ufidx < p_node->functions[fidx].uses_function.size(); ufidx++) {
+		StringName function_name = p_node->functions[fidx].uses_function[ufidx];
+		if (r_added.has(function_name)) {
 			continue;
 		}
 
-		_dump_function_deps(p_node, E->get(), p_func_code, r_to_add, r_added);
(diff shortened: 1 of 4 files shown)


drivers/gles2/shader_compiler_gles2.cpp
```diff
@@ -229,17 +229,18 @@ void ShaderCompilerGLES2::_dump_function_deps(const SL::ShaderNode *p_node, cons
 
 	ERR_FAIL_COND(fidx == -1);
 
-	for (Set<StringName>::Element *E = p_node->functions[fidx].uses_function.front(); E; E = E->next()) {
-		if (r_added.has(E->get())) {
+	for (int j = 0; j < p_node->functions[fidx].uses_function.size(); j++) {
```
AThousandShips, line 232: I'd rename the inner index to avoid confusion, I've never really seen an outer index `j` and an inner `i`, or name it more descriptively like `function_i`

Listwon: I think i'll name it more descriptively to avoid unnecessary changes (there are two inner loops).

clayjohn:
What are the steps for reproducing the issue? Just having a shader with a large number of functions? I would like to test if the issue exists in 4.x as well

Listwon:
@clayjohn yes, I remember it being hard to reproduce in simple projects (without custom shaders), but more functions can help (it's not guaranteed to assign them specific addresses in memory). Maybe naming them specifically (like some existing StringNames) could help too?

lawnjelly (approved):
I'm not super familiar with this area, but it seems to look good.

akien-mga:
Thanks!

PR #76591, opened 2023-04-29
Autocomplete properties in `tween_property`

HolonProduction:
Fixes godotengine/godot-proposals#6739

PR #86747, opened 2024-01-03
Add autocompletion for `OS.has_feature()`

### LINKED ISSUE #4240: Add autocompletion for `OS.has_feature()`'s parameter

A not-so typical Puzzle Bubble-like that has made general use of OS.has_feature("debug")

In my code I make decent use out of **Feature Tags**. Although I only use the most common ones, a great chunk of them are available by default, for developers that want to fine-tune their project when exporting.

However, there's a bit of a... moving back 'n' forth between the list of all **Feature Tags** and the **Script Editor**.
It's quite easy to be a little forgetful, misspell a Feature Tag and having to check the list once more just to make sure it's among the default ones. It definitely doesn't help that the Strings vary in length and capitalisation _(in 3.x)_.

It's a bit of an annoyance, seeing as, in the average project, **constant** Strings are passed as arguments for  `OS.has_feature()`s anyhow.

Add parameter autocompletion to `OS.has_feature()`, displaying all of the non-custom Feature Tags.

It's very similar to `Input.is_action_pressed()` and derivates, whose action autocompletion also aim to solve the same issue.

I don't have much else to say. If it could be integrated with custom Feature Tags it'd be nice too, but it doesn't look possible in any way as of now.

*(mockup)*:
image

Not really.

No add-on can implement this.

Mickeon:
Closes https://github.com/godotengine/godot-proposals/issues/4240.

image

The implementation borrows from **ProjectSettingsEditor**, putting into **EditorExport** for **OS** to access.

But this causes a dependency on the editor class...

### DIFF
--- a/core/core_bind.cpp
+++ b/core/core_bind.cpp
@@ -41,6 +41,9 @@
 #include "core/os/keyboard.h"
 #include "core/os/thread_safe.h"
 #include "core/variant/typed_array.h"
+#ifdef TOOLS_ENABLED
+#include "editor/export/editor_export.h"
+#endif // TOOLS_ENABLED
 
 namespace core_bind {
 
@@ -556,6 +559,19 @@ void OS::revoke_granted_permissions() {
 String OS::get_unique_id() const {
 	return ::OS::get_singleton()->get_unique_id();
 }
+
+#ifdef TOOLS_ENABLED
+void OS::get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const {
+	String pf = p_function;
+	if (p_idx == 0 && pf == "has_feature") {
+		HashSet<String> presets = EditorExport::get_singleton()->get_available_features();
+
+		for (const String &E : presets) {
+			r_options->push_back(E.quote());
+		}
+	}
+}
+#endif // TOOLS_ENABLED
 
 OS *OS::singleton = nullptr;
 
--- a/core/core_bind.h
+++ b/core/core_bind.h
@@ -242,6 +242,10 @@ public:
 	Vector<String> get_granted_permissions() const;
 	void revoke_granted_permissions();
 
+#ifdef TOOLS_ENABLED
+	virtual void get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const override;
+#endif // TOOLS_ENABLED
+
 	static OS *get_singleton() { return singleton; }
 
 	OS() { singleton = this; }
--- a/editor/export/editor_export.cpp
+++ b/editor/export/editor_export.cpp
@@ -167,6 +167,53 @@ Vector<Ref<EditorExportPlugin>> EditorExport::get_export_plugins() {
 	return export_plugins;
 }
 
+HashSet<String> EditorExport::get_available_features() {
+	HashSet<String> presets;
+
+	presets.insert("bptc");
+	presets.insert("s3tc");
+	presets.insert("etc");
+	presets.insert("etc2");
+	presets.insert("editor");
+	presets.insert("template_debug");
+	presets.insert("template_release");
+	presets.insert("debug");
+	presets.insert("release");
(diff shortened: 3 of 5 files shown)


core/core_bind.cpp
```diff
@@ -557,6 +557,33 @@ String OS::get_unique_id() const {
 	return ::OS::get_singleton()->get_unique_id();
 }
 
+void get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) {
```
AThousandShips, line 560:
```suggestion
void OS::get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const {
```

core/core_bind.cpp
```diff
@@ -557,6 +557,33 @@ String OS::get_unique_id() const {
...
+		presets.insert("template_debug");
+		presets.insert("template_release");
+		presets.insert("debug");
+		presets.insert("release");
+		presets.insert("template");
+		presets.insert("double");
+		presets.insert("single");
+		presets.insert("32");
+		presets.insert("64");
+		presets.insert("movie");
```
AThousandShips, line 579: I don't think this needs to be a hash map, the entries can be made unique as is and doesn't depend on data from anywhere else

Mickeon:
I was answering back but the comment was deleted.
It's like this because it's a copy-paste from **ProjectSettingsEditor**. There's no real place where these are "stored", quoting Vnen:

> This should be easy to add, but I don't think the list of features is available anywhere. So either it should be hardcoded into GDScript auto-completion or we need add a central list of possible features in the editor to harness for the completion list.

Maybe and ideally, the two need to be united somehow.

AThousandShips:
Indeed, noticed that it was just as I had commented, my bad 🙂

I'd say to either unite it with the other one, or use the list directly to avoid the step

Mickeon:
I think I came with a nice compromise, but the OS needs to know about EditorExport which is still not nice, so perhaps as a last resort I would have to hardcode it like https://github.com/godotengine/godot/pull/76591 has done, into `gdscript_editor.cpp`. I don't find it to be neatly organized, but it would be the less dependency-d... dependent?

core/core_bind.cpp
```diff
@@ -41,6 +41,7 @@
 #include "core/os/keyboard.h"
 #include "core/os/thread_safe.h"
 #include "core/variant/typed_array.h"
+#include "editor/export/editor_export.h"
```
AThousandShips, line 44:
```suggestion
#ifdef TOOLS_ENABLED
#include "editor/export/editor_export.h"
#endif
```

core/core_bind.cpp
```diff
@@ -557,6 +558,17 @@ String OS::get_unique_id() const {
 	return ::OS::get_singleton()->get_unique_id();
 }
 
+void OS::get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const {
+	String pf = p_function;
+	if (p_idx == 0 && pf == "has_feature") {
+		HashSet<String> presets = EditorExport::get_singleton()->get_available_features();
```
AThousandShips, line 564:
```suggestion
		HashSet<String> presets;
#ifdef TOOLS_ENABLED
		presets = EditorExport::get_singleton()->get_available_features();
#endif
```
Or the entire block, or method entirely

core/core_bind.h
```diff
@@ -242,6 +242,8 @@ class OS : public Object {
 	Vector<String> get_granted_permissions() const;
 	void revoke_granted_permissions();
 
+	virtual void get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const override;
```
AThousandShips, line 245:
```suggestion
#ifdef TOOLS_ENABLED
	virtual void get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const override;
#endif // TOOLS_ENABLED
```

Issue #57195, opened 2022-01-25
iOS export failing when entering provisioning profile and codesign identity


ehmprah:
### Godot version

3.4.stable

### System information

MacOS 12.1

### Issue description

When exporting my project for iOS, I entered all necessary codesigning data like team ID, codesign identity and provisioning profile UUID. Exporting fails with the following error:

error: PROJECT has conflicting provisioning settings. PROJECT is automatically signed, but provisioning profile PROVISIONING_PROFILE has been manually specified. Set the provisioning profile value to "Automatic" in the build settings editor, or switch to manual signing in the Signing & Capabilities editor. (in target 'PROJECT' from project 'PROJECT')
** ARCHIVE FAILED **

I would expect the automatic signing to be turned off once I supply all these details? Or am I somehow not understanding the error message?

### Steps to reproduce

See above. The only way the export works for me is to leave the provisioning profile UUID empty and by entering "iPhone Developer" / "iPhone Distribution" in the codesign identity fields.

nathanfranke:
This should be re-opened, as someone [on Reddit](https://www.reddit.com/r/godot/comments/w9fwpe/issue_with_building_for_release_on_ios_signing/) and I can reproduce it in Godot 3.5.

Luckily, I think I found a workaround for App Store Connect (Test Flight + App Store):
1) Before continuing, do yourself a favor and make sure you have a **JPG** selected for Icons→"App Store 1024 X 1024", otherwise XCode will complain about a transparent image. This should be improved in Godot too.
2) Keep "Code Sign Identity Debug" and "Code Sign Identity Release" blank in Godot export
3) Export
4) Open XCode, click the root item to open the main settings area, switch to the Build Settings tab and scroll all the way down to Signing
5) Change **all 4** "Code Signing Identity" values to "Apple Development". This is silly but from all my testing it was the only way to trick it into working.
6) Go to the "Signing & Capabilities" tab and uncheck/recheck "Automatically manage signing" a couple times, kinda like whack a mole, until 7)
7) Eventually there should be a single signing area, with "Automatically manage signing" **checked**. Now select your Team.
8) I did Product→Archive for my own workflow, and it seems to work. Make sure App Store Connect is selected, and later on it will tell you it is resigning everything

Overvault-64:
I'm getting this same issue when exporting to iOS with Godot 4.1.1. I've set nothing in the export settings other than team ID and package name.

scripturial:
I am encountering the same problem 4.1.3

[...]

PR #86748, opened 2024-01-03
[iOS] Set provisioning style for both `iPhone Developer` and `iPhone Distribution` to automatic

sassanh:
Related: https://github.com/godotengine/godot/issues/57195

I think it should work with development certificate by default (both for debug and release builds)
I do have a distribution certificate and yet I keep getting this error: in `Signing & Capabilities` tab of the target in Xcode:

```... is automatically signed for development, but a conflicting code signing identity iPhone Distribution has been manually specified. Set the code signing identity value to "Apple Development" in the build settings editor, or switch to manual signing in the Signing & Capabilities editor.```

If I replace these two lines in `project.pbxproj` file:
```
				CODE_SIGN_IDENTITY = "iPhone Distribution";
				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
```

to

```
				CODE_SIGN_IDENTITY = "iPhone Developer";
				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
```

things start to work, you may say you can set it in the Godot export wizard for iOS template, but the thing is as soon as I set a value for code sign identity, the `export_plugin.cpp` will set `CODE_SIGN_STYLE` to `MANUAL` instead of `AUTOMATIC` and it also causes an error in Xcode build. The only way things work is when `CODE_SIGN_STYLE` is kept `AUTOMATIC` and the above lines are changed. so there is no way I'm aware of that I (and I guess the people who commented on that issue and probably others) can get a working export out of the box using the export dialog, and the ipa file generation always fails.

If this PR is not accepted, please explain in details what's going on here, what kind of certificates the developer needs to setup on their machine, what configuration is needed to make it work without this PR merged. If it is not officially documented, it needs to at least be mentioned in a GitHub thread somewhere so that a Google Search finds something. The iOS deployment code is helpful if it works smoothly for people other than the one who implemented it, and it needs documentation for this purpose.

### DIFF
--- a/platform/ios/export/export_plugin.cpp
+++ b/platform/ios/export/export_plugin.cpp
@@ -231,8 +231,8 @@ void EditorExportPlatformIOS::_fix_config_file(const Ref<EditorExportPreset> &p_
 	};
 	String dbg_sign_id = p_preset->get("application/code_sign_identity_debug").operator String().is_empty() ? "iPhone Developer" : p_preset->get("application/code_sign_identity_debug");
 	String rel_sign_id = p_preset->get("application/code_sign_identity_release").operator String().is_empty() ? "iPhone Distribution" : p_preset->get("application/code_sign_identity_release");
-	bool dbg_manual = !p_preset->get_or_env("application/provisioning_profile_uuid_debug", ENV_IOS_PROFILE_UUID_DEBUG).operator String().is_empty() || (dbg_sign_id != "iPhone Developer");
-	bool rel_manual = !p_preset->get_or_env("application/provisioning_profile_uuid_release", ENV_IOS_PROFILE_UUID_RELEASE).operator String().is_empty() || (rel_sign_id != "iPhone Distribution");
+	bool dbg_manual = !p_preset->get_or_env("application/provisioning_profile_uuid_debug", ENV_IOS_PROFILE_UUID_DEBUG).operator String().is_empty() || (dbg_sign_id != "iPhone Developer" && dbg_sign_id != "iPhone Distribution");
+	bool rel_manual = !p_preset->get_or_env("application/provisioning_profile_uuid_release", ENV_IOS_PROFILE_UUID_RELEASE).operator String().is_empty() || (rel_sign_id != "iPhone Developer" && rel_sign_id != "iPhone Distribution");
 	String str;
 	String strnew;
 	str.parse_utf8((const char *)pfile.ptr(), pfile.size());

### CODE REVIEW

File: platform/ios/export/export_plugin.cpp [hunk 1]
```diff
@@ -171,7 +171,7 @@ void EditorExportPlatformIOS::get_export_options(List<ExportOption> *r_options)
 	r_options->push_back(ExportOption(PropertyInfo(Variant::STRING, "application/code_sign_identity_debug", PROPERTY_HINT_PLACEHOLDER_TEXT, "iPhone Developer"), ""));
 	r_options->push_back(ExportOption(PropertyInfo(Variant::INT, "application/export_method_debug", PROPERTY_HINT_ENUM, "App Store,Development,Ad-Hoc,Enterprise"), 1));
 	r_options->push_back(ExportOption(PropertyInfo(Variant::STRING, "application/provisioning_profile_uuid_release", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_DEFAULT | PROPERTY_USAGE_SECRET), ""));
-	r_options->push_back(ExportOption(PropertyInfo(Variant::STRING, "application/code_sign_identity_release", PROPERTY_HINT_PLACEHOLDER_TEXT, "iPhone Distribution"), ""));
+	r_options->push_back(ExportOption(PropertyInfo(Variant::STRING, "application/code_sign_identity_release", PROPERTY_HINT_PLACEHOLDER_TEXT, "iPhone Developer"), ""));
```
AThousandShips, line 174: Why this change? The release should reasonably be "distribution" not "developer", no?

AThousandShips:
I don't think there's anything strange with the `Developer` Vs `Disrtribution` to me it makes perfect sense to have two different identities in that sense

sassanh:
I don't think it's a good idea to expect the game developer to have a `Distribution` certificate when they start testing their app on iPhone. Even a dedicated iPhone developer may not go for a `Distribution` certificate until the day they want to actually publish their app in app store.
Besides that, I do have a Distribution certificate, but I don't know why I keep getting the above error.



> It's probably better to change these two lines to keep mode to `automatic` for both `developer` and `distribution` instead:

It's definitely an improvement, I will do as you said, what about changing the default of release to `developer`?

AThousandShips:
Then just change it? It's not a forced name, the *default* should be that you use a different identity for release, for security etc., does it not work by changing manually?

sassanh:
> Then just change it? It's not a forced name, the _default_ should be that you use a different identity for release, for security etc., does it not work by changing manually?

If I change it, then as I mentioned in the original post `CODE_SIGN_STYLE` becomes `MANUAL` and Xcode fails to build again, I need to open the project file in an editor or open the Xcode itself to select the profile because it is not selecting it automatically when `CODE_SIGN_STYLE` is set to `MANUAL`

> That's not what was said though:

That's why I'm asking "what about it?" with a question mark in the end which means other than agreeing with "what is said", I'm suggesting something new which is changing the default for "Code Sign Identity Release" filed to "iPhone Developer" instead of "iPhone Distribution".

The rational behind this suggestion is people may not have created a distribution profile and they may don't know by typing "iPhone Developer" in "Code Sign Identity Release" field.

AThousandShips:
> That's why I'm asking "what about it?"

I missed a detail and therefore removed that part of my note 🙂

I think the solution, as suggested above, should be to handle that case, not to change the default 🙂

sassanh:
> I missed a detail and therefore removed that part of my note 🙂

Sorry, I didn't notice you have edited 🙂 Btw I really appreciate your quick responses. So I will just make it keep `CODE_SIGN_STYLE` automatic for both values.

AThousandShips:
No worries, easy to miss 🙂

sassanh:
I updated this pull request, please review it when you got the time.

sassanh:
@AThousandShips @bruvzg Sorry to ping you, since last time you responded in less than 5 minutes, I thought maybe you haven't noticed yet PR is updated.

bruvzg (approved):

platform/ios/export/export_plugin.cpp
```diff
@@ -231,8 +231,8 @@ void EditorExportPlatformIOS::_fix_config_file(const Ref<EditorExportPreset> &p_
 	};
 	String dbg_sign_id = p_preset->get("application/code_sign_identity_debug").operator String().is_empty() ? "iPhone Developer" : p_preset->get("application/code_sign_identity_debug");
 	String rel_sign_id = p_preset->get("application/code_sign_identity_release").operator String().is_empty() ? "iPhone Distribution" : p_preset->get("application/code_sign_identity_release");
-	bool dbg_manual = !p_preset->get_or_env("application/provisioning_profile_uuid_debug", ENV_IOS_PROFILE_UUID_DEBUG).operator String().is_empty() || (dbg_sign_id != "iPhone Developer");
-	bool rel_manual = !p_preset->get_or_env("application/provisioning_profile_uuid_release", ENV_IOS_PROFILE_UUID_RELEASE).operator String().is_empty() || (rel_sign_id != "iPhone Distribution");
+	bool dbg_manual = !p_preset->get_or_env("application/provisioning_profile_uuid_debug", ENV_IOS_PROFILE_UUID_DEBUG).operator String().is_empty() || (rel_sign_id != "iPhone Developer" && rel_sign_id != "iPhone Distribution");
```
bruvzg, line 234:
```suggestion
	bool dbg_manual = !p_preset->get_or_env("application/provisioning_profile_uuid_debug", ENV_IOS_PROFILE_UUID_DEBUG).operator String().is_empty() || (dbg_sign_id != "iPhone Developer" && dbg_sign_id != "iPhone Distribution");
```

sassanh: @bruvzg Thanks! I fixed it.

akien-mga:
Cherry-picked for 4.2.2.

0xafbf:
@akien-mga Hello! Sorry to bother, can we get this cherry picked for 3.x too? Thanks!

PR #44112, opened 2020-12-05
make 2d constraint solving more deterministic by solving in push order

jordo:
This PR addresses https://github.com/godotengine/godot/issues/44099 as discussed on IRC.

@pouleyKetchoupp I have tested this pretty extensively tonight, and anecdotally it addresses https://github.com/godotengine/godot/issues/44099  with 100% consistency.  I now get 100% reliable re-simulations.  

The previous constraint_map was getting solved based on pointer hash, which is going to be inherently random...  This solves the constraints in order of which they are pushed to a list.  

The only potential  performance impact of moving to a list from a map, is removal of constraints in remove_constraint().  Previously it was by key only, which is a key lookup, but now it will iterate a list.  However this shouldn't be of any real significance as insertion cost will be improved.

@pouleyKetchoupp and @reduz This could also be applied to the 3d solver.

PR #86749, opened 2024-01-03
Update the transform before checking for collisions and start checking from the latest added constraint

### LINKED ISSUE #86199: Area2D body_entered signal is emitted too late
### Tested versions

Reproducible in: v4.2.1.stable.official [b09f793f5]

### System information

Windows 10 - v4.2.1.stable.official [b09f793f5]

### Issue description

Area2D body_entered signal seem to execute the callback 1 physics tick too late. Here is a visual representation, where the player is a CharacterBody2D, and the coin is an Area2D:

fd521bf49d9a2b73321ffdbcc6a45bbefbee1f46

This visual overlap is undesired, the Area2D (coin) is set to become invisible (visible = false) in the body_entered callback, however it is still visible for the first frame where it overlaps with the player.

This also happens in the official pong demo, where the ball can be seen overlap the paddles if the speed of the ball increases.

If maybe this is not considered a bug and it's just the way the engine was designed, there should be an easy alternative to make an Area2D with immediate body_entered callbacks. There's a running thread on the forums on this issue: https://forum.godotengine.org/t/why-are-area2d-collisions-triggered-1-frame-late/37354

### Steps to reproduce

Create a CharacterBody2D and another Area2D. Make the Area2D emit the body_entered signal, and on the callback set it so the Area2D becomes invisible or whatever. You'll see that the CharacterBody2D actually overlaps the Area2D for 1 frame. 

[MRCollisionExample.zip](https://github.com/godotengine/godot/files/13687068/MRCollisionExample.zip)
In this MRP just press right arrow key to move the CharacterBody2D to the right. When it reaches the Area2D, the callback is set to teleport the CharacterBody2D back to the start, however, they can be seen overlapping for a frame, like this:
ebeaf25a703e804c3bc0b49825080a7c27c5ed3d_2_690x456

This also happens in the Godot pong official demo: https://godotengine.org/asset-library/asset/121
If you increase the ball speed, it will be seen overlaping the paddles, which is a highly undesired visual glitch.

### LINKED ISSUE #86640: Body_Exited Signal incorrectly triggering between CollisionPolygon2D Segments despite build mode being "Solids"
### Tested versions

v4.1.stable.official [970459615]

### System information

Godot v4.1.stable - macOS 13.5.2 - Vulkan (Forward+) - integrated Intel(R) Iris(TM) Plus Graphics 655 - Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz (8 Threads)

### Issue description

Player is detected as leaving the area when they are still within the same CollisonPolygon2D. The exited signal for this only triggers when the player 'undos' within the top portion of this puzzle, which happens to be the only area where the polygon shape segments are pure rectangles. The player can undo in other sections of the puzzle without the exited signal triggering.

The exited signal should only trigger if the player is stepping outside of the entire collision area. 

### Steps to reproduce

Walk the player around using WASD or arrow keys. You have a move count and will need to recharge by stepping on the lightning bolt tiles. At anytime, you can undo by pressing "Q". In the first part of the puzzle, everything should work as intended, however once you get to the last stretch (as shown in this screenshot & video) undoing your player will cause the puzzle to deactivate because the body_exited signal is triggered. This only happens if the player is undone between the colored segments in this section. 

https://github.com/godotengine/godot/assets/60891368/211625a3-9cf4-46d7-9633-8b0980833cdc

Screenshot 2023-12-30 at 11 15 13 AM

Currently I am able to work around this by setting up a function to check when the player is being 'undone' and then checking when that is false within the body_exited signal before executing any other code. This should be fine for my project but it still doesn't solve the fact that the exited signal is being triggered here in general. 

[ISSUE_12302023.zip](https://github.com/godotengine/godot/files/13798612/ISSUE_12302023.zip)

Rindbee:
## Place the code that updates the transform before detecting collisions

Previously, after updating the body's transform through a script, the transform used in rendering will be updated immediately in the same frame.

In physics, it is usually necessary to call `GodotBody2D::integrate_velocities()` to update the body's transform. But in `GodotStep2D::step()`, the method is called after the code to detect collision, which will cause the collision to be detected until the next frame. The result is that collisions are "seen" immediately, but reported one frame later.

Fix #86199.

## Add the newest constraint to the front of `constraint_list`
    
Since `constraint_list` adds the newest constraint to the end of the list, and `GodotStep2D::_populate_island()` traverses `constraint_list` from the beginning. That is to say, the collision is judged from the earliest entered shape.

In the case where body `A` enters new shape `1` and exits old shape `0` within the same area/body `B`. The current order is `A` first exits shape `0`, then enters shape `1`. Shape `0` may be the last shape in which `B` collides with `A`, so exiting the last shape will cause `A` to exit from `B`.

Starting with the most recently added constraint can avoid this problem.

Fix #86640.

### DIFF
--- a/servers/physics_2d/godot_body_2d.h
+++ b/servers/physics_2d/godot_body_2d.h
@@ -202,7 +202,8 @@ public:
 	_FORCE_INLINE_ uint64_t get_island_step() const { return island_step; }
 	_FORCE_INLINE_ void set_island_step(uint64_t p_step) { island_step = p_step; }
 
-	_FORCE_INLINE_ void add_constraint(GodotConstraint2D *p_constraint, int p_pos) { constraint_list.push_back({ p_constraint, p_pos }); }
+	// The newest constraint is added to the front to prevent incorrect populating when replacing the old constraint with the new one.
+	_FORCE_INLINE_ void add_constraint(GodotConstraint2D *p_constraint, int p_pos) { constraint_list.push_front({ p_constraint, p_pos }); }
 	_FORCE_INLINE_ void remove_constraint(GodotConstraint2D *p_constraint, int p_pos) { constraint_list.erase({ p_constraint, p_pos }); }
 	const List<Pair<GodotConstraint2D *, int>> &get_constraint_list() const { return constraint_list; }
 	_FORCE_INLINE_ void clear_constraint_list() { constraint_list.clear(); }
--- a/servers/physics_2d/godot_step_2d.cpp
+++ b/servers/physics_2d/godot_step_2d.cpp
@@ -153,12 +153,29 @@ void GodotStep2D::step(GodotSpace2D *p_space, real_t p_delta) {
 
 	p_space->set_active_objects(active_count);
 
-	// Update the broadphase to register collision pairs.
+	{ //profile
+		profile_endtime = OS::get_singleton()->get_ticks_usec();
+		p_space->set_elapsed_time(GodotSpace2D::ELAPSED_TIME_INTEGRATE_FORCES, profile_endtime - profile_begtime);
+		profile_begtime = profile_endtime;
+	}
+
+	/* INTEGRATE VELOCITIES */
+
+	// Update transform before checking for collision.
+
+	b = body_list->first();
+	while (b) {
+		const SelfList<GodotBody2D> *n = b->next();
+		b->self()->integrate_velocities(p_delta);
+		b = n; // in case it shuts itself down
+	}
+
+	// Update the broadphase to register/unregister collision pairs.
 	p_space->update();
 
 	{ //profile
 		profile_endtime = OS::get_singleton()->get_ticks_usec();
-		p_space->set_elapsed_time(GodotSpace2D::ELAPSED_TIME_INTEGRATE_FORCES, profile_endtime - profile_begtime);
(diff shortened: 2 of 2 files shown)


servers/physics_2d/godot_body_2d.h
```diff
@@ -202,7 +202,8 @@ class GodotBody2D : public GodotCollisionObject2D {
 	_FORCE_INLINE_ uint64_t get_island_step() const { return island_step; }
 	_FORCE_INLINE_ void set_island_step(uint64_t p_step) { island_step = p_step; }
 
-	_FORCE_INLINE_ void add_constraint(GodotConstraint2D *p_constraint, int p_pos) { constraint_list.push_back({ p_constraint, p_pos }); }
+	// The newest constraint is added to the front to prevent incorrect populating when replacing the old constraint with the new one.
+	_FORCE_INLINE_ void add_constraint(GodotConstraint2D *p_constraint, int p_pos) { constraint_list.push_front({ p_constraint, p_pos }); }
```
Rindbee, line 206: Since only `GodotBody2D` uses `List` (implemented by #44112), structures supporting FIFO/LIFO are not yet used in `GodotArea2D` and `GodotArea3D`/`GodotBody3D`/`GodotSoftBody3D`. 

So it is still random in the corresponding `CollisionObject`.

PR #86750, opened 2024-01-03
Add signals for Area intersections to PhysicsBody nodes

### LINKED ISSUE #8784: Add area intersection signals to PhysicsBody nodes

A third person 3D action-adventure game where physical bodies enter areas a lot.

TL;DR: I'd like to add the likes of area_entered/exited (and their _shape_ variants) signals to all PhysicsBodies.

Basically it all started with me implementing character swimming. To my surprise I found out there's no way to tell if `CharacterBody3D` has entered `Area3D` (representing water) from the character itself. This seems like an omission, because it's a very common use case, but currently requires clunky workarounds. It also strays from the otherwise signals-oriented design.
(Also, on `RigidBody` type nodes we have signals for collisions but not `Area` intersections.)

`PhysicsBody` classes get 4 new signals corresponding to those on `Areas`.
`entered_area(area)`
`entered_area_shape(area_rid, area, area_shape_id, body_shape_id)`
`exited_area(area)`
`exited_area_shape(area_rid, area, area_shape_id, body_shape_id)`

This allows bodies and scripts attached to them to be aware of which areas they enter and with which shapes of those areas they intersect with minimal extra code.

`Body` enters `Area`, they shake hands, instead of `Area` unilaterally noticing it has been entered. `Area` already does all the heavy lifting, what's left is adding calls to emit signals on the entering bodies. Signals will be emitted right after the `Area` signals. Bodies don't need to keep references to all their intersecting areas by default, and don't need to check if they are still within the area, as `Area` already does this for them, thus this enhancement should not add a significant overhead. It's should not be breaking compat.

Per @aaronfranke's suggestion the naming convention of the signals is reversed for clarity (eg. `entered_area` instead of `area_entered`), and requires adding these new names to `SceneStringNames`.

Current workaround is adding a script to `Area` that would communicate with the entering body (so sort of reimplementing signals in script), but this requires adding these scripts to every area of interest in the game, which is a productivity bottleneck, and may even affect performance.
Another workaround is attaching an `Area` to the Body in question, setting its shape to the same shape the `Body` uses, and attaching this `Area`'s signals to `Body`'s script. But this is very clunky and draws an extra gizmo per body in debug view. 

Area intersection is a part of core physics.

Illauriel:
- Resolves godotengine/godot-proposals/issues/8784

Adds 4 new signals to all PhysicsBodies:
`entered_area(area)`
`entered_area_shape(area_rid, area, area_shape_id, body_shape_id)`
`exited_area(area)`
`exited_area_shape(area_rid, area, area_shape_id, body_shape_id)`

Signal emission is handled by Areas, when the body enters or exits them.

I need this only in 3D, but also added a 2D version of enhancement in a separate commit.

### DIFF
--- a/scene/2d/area_2d.cpp
+++ b/scene/2d/area_2d.cpp
@@ -143,8 +143,14 @@ void Area2D::_body_enter_tree(ObjectID p_id) {
 
 	E->value.in_tree = true;
 	emit_signal(SceneStringNames::get_singleton()->body_entered, node);
+	if (node->has_signal(SceneStringNames::get_singleton()->entered_area)) {
+		node->emit_signal(SceneStringNames::get_singleton()->entered_area, this);
+	}
 	for (int i = 0; i < E->value.shapes.size(); i++) {
 		emit_signal(SceneStringNames::get_singleton()->body_shape_entered, E->value.rid, node, E->value.shapes[i].body_shape, E->value.shapes[i].area_shape);
+		if (node->has_signal(SceneStringNames::get_singleton()->entered_area_shape)) {
+			node->emit_signal(SceneStringNames::get_singleton()->entered_area_shape, get_rid(), this, E->value.shapes[i].area_shape, E->value.shapes[i].body_shape);
+		}
 	}
 }
 
@@ -157,8 +163,14 @@ void Area2D::_body_exit_tree(ObjectID p_id) {
 	ERR_FAIL_COND(!E->value.in_tree);
 	E->value.in_tree = false;
 	emit_signal(SceneStringNames::get_singleton()->body_exited, node);
+	if (node->has_signal(SceneStringNames::get_singleton()->exited_area)) {
+		node->emit_signal(SceneStringNames::get_singleton()->exited_area, this);
+	}
 	for (int i = 0; i < E->value.shapes.size(); i++) {
 		emit_signal(SceneStringNames::get_singleton()->body_shape_exited, E->value.rid, node, E->value.shapes[i].body_shape, E->value.shapes[i].area_shape);
+		if (node->has_signal(SceneStringNames::get_singleton()->exited_area_shape)) {
+			node->emit_signal(SceneStringNames::get_singleton()->exited_area_shape, get_rid(), this, E->value.shapes[i].area_shape, E->value.shapes[i].body_shape);
+		}
 	}
 }
 
@@ -204,6 +216,9 @@ void Area2D::_body_inout(int p_status, const RID &p_body, ObjectID p_instance, i
 				node->connect(SceneStringNames::get_singleton()->tree_exiting, callable_mp(this, &Area2D::_body_exit_tree).bind(objid));
 				if (E->value.in_tree) {
 					emit_signal(SceneStringNames::get_singleton()->body_entered, node);
+					if (node->has_signal(SceneStringNames::get_singleton()->entered_area)) {
+						node->emit_signal(SceneStringNames::get_singleton()->entered_area, this);
+					}
 				}
 			}
 		}
@@ -214,6 +229,9 @@ void Area2D::_body_inout(int p_status, const RID &p_body, ObjectID p_instance, i
 
 		if (!node || E->value.in_tree) {
 			emit_signal(SceneStringNames::get_singleton()->body_shape_entered, p_body, node, p_body_shape, p_area_shape);
+			if (node && node->has_signal(SceneStringNames::get_singleton()->entered_area_shape)) {
+				node->emit_signal(SceneStringNames::get_singleton()->entered_area_shape, get_rid(), this, p_area_shape, p_body_shape);
+			}
 		}
 
 	} else {
@@ -231,11 +249,17 @@ void Area2D::_body_inout(int p_status, const RID &p_body, ObjectID p_instance, i
 				node->disconnect(SceneStringNames::get_singleton()->tree_exiting, callable_mp(this, &Area2D::_body_exit_tree));
 				if (in_tree) {
 					emit_signal(SceneStringNames::get_singleton()->body_exited, obj);
+					if (node->has_signal(SceneStringNames::get_singleton()->exited_area)) {
+						node->emit_signal(SceneStringNames::get_singleton()->exited_area, this);
+					}
 				}
 			}
 		}
 		if (!node || in_tree) {
 			emit_signal(SceneStringNames::get_singleton()->body_shape_exited, p_body, obj, p_body_shape, p_area_shape);
+			if (node && node->has_signal(SceneStringNames::get_singleton()->exited_area_shape)) {
+				node->emit_signal(SceneStringNames::get_singleton()->exited_area_shape, get_rid(), this, p_area_shape, p_body_shape);
+			}
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


scene/3d/area_3d.cpp
```diff
@@ -271,6 +276,7 @@ void Area3D::_body_inout(int p_status, const RID &p_body, ObjectID p_instance, i
 
 		if (!node || E->value.in_tree) {
 			emit_signal(SceneStringNames::get_singleton()->body_shape_entered, p_body, node, p_body_shape, p_area_shape);
+			node->emit_signal(SceneStringNames::get_singleton()->entered_area_shape, get_rid(), this, p_area_shape, p_body_shape);
```
AThousandShips, line 279: This isn't safe, check the code above, `node` can be null

scene/3d/area_3d.cpp
```diff
@@ -288,11 +294,13 @@ void Area3D::_body_inout(int p_status, const RID &p_body, ObjectID p_instance, i
 				node->disconnect(SceneStringNames::get_singleton()->tree_exiting, callable_mp(this, &Area3D::_body_exit_tree));
 				if (in_tree) {
 					emit_signal(SceneStringNames::get_singleton()->body_exited, obj);
+					node->emit_signal(SceneStringNames::get_singleton()->exited_area, this);
 				}
 			}
 		}
 		if (!node || in_tree) {
 			emit_signal(SceneStringNames::get_singleton()->body_shape_exited, p_body, obj, p_body_shape, p_area_shape);
+			node->emit_signal(SceneStringNames::get_singleton()->exited_area_shape, get_rid(), this, p_area_shape, p_body_shape);
```
AThousandShips, line 303: Same here

scene/2d/area_2d.cpp
```diff
@@ -231,11 +237,13 @@ void Area2D::_body_inout(int p_status, const RID &p_body, ObjectID p_instance, i
 				node->disconnect(SceneStringNames::get_singleton()->tree_exiting, callable_mp(this, &Area2D::_body_exit_tree));
 				if (in_tree) {
 					emit_signal(SceneStringNames::get_singleton()->body_exited, obj);
+					node->emit_signal(SceneStringNames::get_singleton()->exited_area, this);
 				}
 			}
 		}
 		if (!node || in_tree) {
 			emit_signal(SceneStringNames::get_singleton()->body_shape_exited, p_body, obj, p_body_shape, p_area_shape);
+			node->emit_signal(SceneStringNames::get_singleton()->exited_area_shape, get_rid(), this, p_area_shape, p_body_shape);
```
AThousandShips, line 246: And here

scene/2d/area_2d.cpp
```diff
@@ -214,6 +219,7 @@ void Area2D::_body_inout(int p_status, const RID &p_body, ObjectID p_instance, i
 
 		if (!node || E->value.in_tree) {
 			emit_signal(SceneStringNames::get_singleton()->body_shape_entered, p_body, node, p_body_shape, p_area_shape);
+			node->emit_signal(SceneStringNames::get_singleton()->entered_area_shape, get_rid(), this, p_area_shape, p_body_shape);
```
AThousandShips, line 222: And here

AThousandShips (dismissed):

AThousandShips:
You also need to document these new signals, easiest way is to compile your code and then run `--doctool`, see [here](https://docs.godotengine.org/en/latest/contributing/documentation/updating_the_class_reference.html#updating-class-reference-when-working-on-the-engine)

AThousandShips:
These also need to be called on `_clear_monitoring` with the other signals

scene/3d/area_3d.cpp
Illauriel: Ah, my bad! I'm on it

Illauriel: > Also it's not guaranteed that this is a physics body so the signal might cause errors if it's not available
Oh. Then I should probably also add a type check?

AThousandShips: Probably, a cast to the appropriate type `Object::cast_to<PhysicsBody2D/3D>` like above with `Node`

Illauriel: Roger /0

AThousandShips: It would be enough to do:
```cpp
			if (Object::cast_to<PhysicsBody3D>(node)) {
```
As it already handles null, and can still call it on `Node`

Illauriel: My first instinct was to do
```gdscript
Object *obj = ObjectDB::get_instance(objid);
Node *node = Object::cast_to<Node>(obj);
PhysicsBody3D *body = Object::cast_to<PhysicsBody3D>(obj);
```

and then

```gdscript
if (body) {
	body->emit_signal(SceneStringNames::get_singleton()->entered_area, this);
}
```
It doesn't have an advantage of being general, and since areas can intersect with other stuff, maybe people using it will want to add signals there too. It's kinda like your first suggestion but with caching the cast :)

AThousandShips: There's no need to cache the cast, since the call `emit_signal` doesn't change, it's the same regardless what class it's called on, but it's either way, you use your own judgement 🙂

Illauriel: I think I'll roll with solution 2 and leave our GridMap fellows the question of arguments :D

Illauriel: Holy, this was a lot of `if`s :)

Done

scene/2d/area_2d.cpp
```diff
@@ -157,8 +163,14 @@ void Area2D::_body_exit_tree(ObjectID p_id) {
 	ERR_FAIL_COND(!E->value.in_tree);
 	E->value.in_tree = false;
 	emit_signal(SceneStringNames::get_singleton()->body_exited, node);
+	if (node && node->has_signal(SceneStringNames::get_singleton()->exited_area)) {
```
AThousandShips, line 166:
```suggestion
	if (node->has_signal(SceneStringNames::get_singleton()->exited_area)) {
```
In the cases where no `if (node)` checks were added before this is unnecessary

scene/2d/area_2d.cpp
```diff
@@ -143,8 +143,14 @@ void Area2D::_body_enter_tree(ObjectID p_id) {
 
 	E->value.in_tree = true;
 	emit_signal(SceneStringNames::get_singleton()->body_entered, node);
+	if (node && node->has_signal(SceneStringNames::get_singleton()->entered_area)) {
```
AThousandShips, line 146:
```suggestion
	if (node->has_signal(SceneStringNames::get_singleton()->entered_area)) {
```
In the cases where no `if (node)` checks were added before this is unnecessary

Illauriel: Ah, alright, will do.

Illauriel: Done!
idk how I got 
```C++
if (node) {
	if (node && [...]
```
Must be losing focus, sorry ^^'

AThousandShips:
Will take an additional look later today 🙂, but looks good code wise I think, the decision will be on if we want this or not, but will take another look at the code and try it out

If you are able to it'd be great if you attached a demo project, with some automated display of this feature, like a moving body with signals 🙂 I can set one up if you want after I've tested, but would be nice if you could do it (I've got different things to review etc.)

Illauriel:
> If you are able to it'd be great if you attached a demo project, with some automated display of this feature, like a moving body with signals 🙂 

Alright, will do!

Illauriel:
[AreaSignalTest.zip](https://github.com/godotengine/godot/files/13833758/AreaSignalTest.zip)
Here's a test project, and it reveals a problem with current implementation. Signals depend on Area's own intersection detection so they won't be emitted when entering areas that are Monitorable, but not Monitoring, which is likely not an expected behavior.

I can come up with a different solution, but I guess it would be reasonable to first settle if we want area intersection signals on bodies or not in the proposal discussion.

Illauriel:
Closing this for the following reasons: 
- Signals in this implementation rely on Area's monitoring property being on, which defeats the purpose of detecting "Areas as environment", represented by Monitorable Areas
- Responsibility concerns. Area calling signals on Body.
- Performance concerns. Every Area in the game emitting 4 additional signals with associated checks on every interaction is more costly than I initially expected.

PR #86753, opened 2024-01-03
Add autocompletion for `Engine.get_singleton` & similar

Mickeon:
Related to https://github.com/godotengine/godot/pull/86747.

Unlike the former PR, it's simple as that. If a bunch of methods are interfacing a collection of Strings of some sort, it can be auto-completed. There's a pattern down the line.

Can be backported to prior versions.

### DIFF
--- a/core/core_bind.h
+++ b/core/core_bind.h
@@ -527,6 +527,8 @@ public:
 	void set_print_error_messages(bool p_enabled);
 	bool is_printing_error_messages() const;
 
+	virtual void get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const override;
+
 	Engine() { singleton = this; }
 };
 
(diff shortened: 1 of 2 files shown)


core/core_bind.h
```diff
@@ -527,6 +527,8 @@ class Engine : public Object {
 	void set_print_error_messages(bool p_enabled);
 	bool is_printing_error_messages() const;
 
+	void get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const override;
```
AThousandShips, line 530:
```suggestion
	virtual void get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const override;
```

AThousandShips (approved):
Looks good otherwise

core/core_bind.h
Mickeon: I'm going to cry

akien-mga:
Thanks!

PR #86754, opened 2024-01-03
Add autocompletion for SpriteFrames' methods

Mickeon:
Related to https://github.com/godotengine/godot/pull/86747 and https://github.com/godotengine/godot/pull/86753.

It's also why I wanted to prepare https://github.com/godotengine/godot/pull/86743.

image

<sub>I wish I could compile a build for a better showcase, but it keeps running out of memory during the compile.</sub>

This PR autocompletes the **SpriteFrames** methods with the names of every existing animation, if they can be detected.

I'm very fond of this. I could have **really** used it in the past.

### DIFF
--- a/scene/resources/sprite_frames.cpp
+++ b/scene/resources/sprite_frames.cpp
@@ -224,6 +224,23 @@ void SpriteFrames::_set_animations(const Array &p_animations) {
 	}
 }
 
+void SpriteFrames::get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const {
+	String pf = p_function;
+	if (p_idx == 0) {
+		if (pf == "has_animation" || pf == "remove_animation" || pf == "rename_animation" ||
+				pf == "set_animation_speed" || pf == "get_animation_speed" ||
+				pf == "set_animation_loop" || pf == "get_animation_loop" ||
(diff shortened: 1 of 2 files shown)


AThousandShips (approved):
Also makes sense

Mickeon:
The code style mandates all of the comparisons to be in one line, which I don't believe is ideal. I've seen some parts of the codebase do not have to do this.

scene/resources/sprite_frames.cpp
```diff
@@ -224,6 +224,23 @@ void SpriteFrames::_set_animations(const Array &p_animations) {
 	}
 }
 
+void SpriteFrames::get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const {
+	String pf = p_function;
+	if (p_idx == 0) {
+		if (pf == "has_animation" || pf == "remove_animation" || pf == "rename_animation"
```
AThousandShips, line 230: I think you need to move the `||` to the end of the previous row, that works elsewhere

Mickeon: Oh that's true. That's it. It's not in my style, hence I forgot to.

akien-mga:
Thanks!

Mickeon:
If needed this can be cherry-picked by the way

PR #86758, opened 2024-01-03
Add autocompletion for ProjectSettings' methods

Mickeon:
Related to https://github.com/godotengine/godot/pull/86747 and https://github.com/godotengine/godot/pull/86753.

This PR adds autocompletion options to all of the methods that interact with the settings themselves.

image

You may think "Didn't they already have autocompletion options?". Well, yeah, you're welcome. Unintentionally, **ProjectSettings** greatly benefitted from https://github.com/godotengine/godot/pull/66427 alone, because `set` and `get` are one good way to check them. This just fills in the rest, which peeved me for a while.

PR #66481, opened 2022-09-27
Add editor settings for autocompletion with NodePath and StringName

MewPurPur:
Closes https://github.com/godotengine/godot-proposals/issues/5496

This is built into the autocompletion algorithm – if you're in a function and autocompletion gives a string, it checks whether the editor setting is on. If it is, it adds the literals `&` or `^` as prefixes, based on if the argument is a NodePath or a StringName. Works in any situation, no adaptation is needed to specific functions.

https://user-images.githubusercontent.com/85438892/192419652-864ee866-b797-474a-9926-acee2fe7b1a4.mp4

PR #86764, opened 2024-01-03
Add autocompletion for InputMap's methods

Mickeon:
Related to https://github.com/godotengine/godot/pull/86747, https://github.com/godotengine/godot/pull/86753 and https://github.com/godotengine/godot/pull/86758.

image

**Input** already autocompleted actions, but **InputMap**, which only contains actions, never did.
... The code is copied straight from **Input** itself.

### DIFF
--- a/core/input/input_map.cpp
+++ b/core/input/input_map.cpp
@@ -85,6 +85,35 @@ String InputMap::suggest_actions(const StringName &p_action) const {
 	return error_message;
 }
 
+#ifdef TOOLS_ENABLED
+void InputMap::get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const {
+	const String pf = p_function;
+	bool first_argument_is_action = false;
+	if (p_idx == 0) {
+		first_argument_is_action = (pf == "has_action" || pf == "erase_action" ||
+				pf == "action_set_deadzone" || pf == "action_get_deadzone" ||
+				pf == "action_has_event" || pf == "action_add_event" || pf == "action_get_events" ||
+				pf == "action_erase_event" || pf == "action_erase_events");
+	}
+	if (first_argument_is_action || (p_idx == 1 && pf == "event_is_action")) {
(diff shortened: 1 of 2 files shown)


Calinou:
By the way, the autocompletion should use `&""` syntax after you accept a suggestion (and display results as StringName with the StringName color), since the parameter type is StringName.

Mickeon:
You would wish that. And https://github.com/godotengine/godot/pull/66481 (for https://github.com/godotengine/godot-proposals/issues/5496) would have it partially covered for it, across the board. It's outside the scope of this PR, and _I for one would rather see them as Strings because I feel like most users shouldn't have to think about it_,

core/input/input_map.cpp
```diff
@@ -85,6 +85,38 @@ String InputMap::suggest_actions(const StringName &p_action) const {
...
+		first_argument_is_action = (pf == "has_action" || pf == "erase_action" ||
+				pf == "action_set_deadzone" || pf == "action_get_deadzone" ||
+				pf == "action_has_event" || pf == "action_add_event" || pf == "action_get_events" ||
+				pf == "action_erase_event" || pf == "action_erase_events");
+	}
+	if (first_argument_is_action || (p_idx == 1 && pf == "event_is_action")) {
+		// Cannot be done. Gets the options in the context of the Editor (no user-defined actions).
+		//for (const StringName &E : get_actions()) {
+		//	r_options->push_back(E.operator String().quote());
+		//}
```
akien-mga, line 102:
```suggestion
		// Cannot rely on `get_actions()`, as it gets the options in the context of the Editor (no user-defined actions).
```
Avoid dead code that we'll keep wanting to remove ;)

akien-mga (approved):

akien-mga:
Might be worth a rebase, I just noticed this branch is a few thousand commits behind `master`.

akien-mga:
Thanks!

PR #84193, opened 2023-10-30
Extract `EditorDockManager` from `EditorNode`

kitbdev:
Moves most code about docks, dock management, and dock select popup from `EditorNode` to `EditorDockManager`.
There should be no behavioral changes in this pr.

Details
- Created new class `EditorDockManager` singleton to encapsulate docks and dock select logic from `EditorNode`
- `EditorNode` passes the dock slots and dock splits to `EditorDockManager`
- Created `DockSplitContainer : SplitContainer` that automatically shows if any children are visible, otherwise it hides
- `EditorNode` still adds the Scene, FileSystem, and other docks but it uses `EditorDockManager` to do it.
- Made `save_docks_to_config`, `load_docks_from_config`, `update_dock_slots_visibility` public in `EditorDockManager` to be accessible from `EditorNode`.
- Made `close_all_floating_docks` as part of the workaround for #84125. This can be removed after the proper fix, if wanted.
- Added signal `layout_changed` in `EditorDockManager` to trigger `EditorNode::_save_editor_layout()`. This could probably be changed to `EditorNode::get_singleton()->save_editor_layout_delayed()` if the delay is fine.
- `EditorPlugin` and `VersionControlEditorPlugin` now use `EditorDockManager` to add and remove docks
- Added optional name parameter to `add_control_to_dock`, since the default docks use a custom title on the tabs. This is not exposed to users.
- Changed some parameter names to be better (`ofs` -> `p_offset`, `p_which` -> `p_dock_slot`)

PR #23273, opened 2018-10-25
Fix segfault on editor close up

Chaosus:
Fix segfault in #23215

PR #86765, opened 2024-01-03
Allow to move FileSystem dock to bottom and drag resources across bottom docks

reduz:
* Allows moving the filesystem dock to the bottom
* Added ability to drag resources across bottom docks

Animation, moving filesystem to the bottom dock:
move_to_bottom

Drag and Drop across bottom panels:
move_to_bottom_dnd

- *Production edit: This partially addresses https://github.com/godotengine/godot-proposals/issues/1508.*

### DIFF
--- a/editor/editor_dock_manager.cpp
+++ b/editor/editor_dock_manager.cpp
@@ -113,6 +113,8 @@ void EditorDockManager::_dock_select_popup_theme_changed() {
 		dock_tab_move_left->set_icon(dock_select_popup->get_editor_theme_icon(SNAME("Back")));
 		dock_tab_move_right->set_icon(dock_select_popup->get_editor_theme_icon(SNAME("Forward")));
 	}
+
+	dock_to_bottom->set_icon(dock_select_popup->get_editor_theme_icon(SNAME("ControlAlignBottomWide")));
 }
 
 void EditorDockManager::_dock_popup_exit() {
@@ -122,6 +124,19 @@ void EditorDockManager::_dock_popup_exit() {
 
 void EditorDockManager::_dock_pre_popup(int p_dock_slot) {
 	dock_popup_selected_idx = p_dock_slot;
+	dock_bottom_selected_idx = -1;
+
+	if (bool(dock_slot[p_dock_slot]->get_current_tab_control()->call("_can_dock_horizontal"))) {
+		dock_to_bottom->show();
+	} else {
+		dock_to_bottom->hide();
+	}
+
+	if (dock_float) {
+		dock_float->show();
+	}
+	dock_tab_move_right->show();
+	dock_tab_move_left->show();
 }
 
 void EditorDockManager::_dock_move_left() {
@@ -179,23 +194,43 @@ void EditorDockManager::_dock_select_input(const Ref<InputEvent> &p_input) {
 
 		Ref<InputEventMouseButton> mb = me;
 
-		if (mb.is_valid() && mb->get_button_index() == MouseButton::LEFT && mb->is_pressed() && dock_popup_selected_idx != nrect) {
-			dock_slot[nrect]->move_tab_from_tab_container(dock_slot[dock_popup_selected_idx], dock_slot[dock_popup_selected_idx]->get_current_tab(), dock_slot[nrect]->get_tab_count());
+		if (mb.is_valid() && mb->get_button_index() == MouseButton::LEFT && mb->is_pressed()) {
+			if (dock_bottom_selected_idx != -1) {
+				EditorNode::get_singleton()->remove_bottom_panel_item(bottom_docks[dock_bottom_selected_idx]);
 
-			if (dock_slot[dock_popup_selected_idx]->get_tab_count() == 0) {
-				dock_slot[dock_popup_selected_idx]->hide();
-			} else {
-				dock_slot[dock_popup_selected_idx]->set_current_tab(0);
+				bottom_docks[dock_bottom_selected_idx]->call("_set_dock_horizontal", false);
+
+				dock_slot[nrect]->add_child(bottom_docks[dock_bottom_selected_idx]);
+				dock_slot[nrect]->show();
+				bottom_docks.remove_at(dock_bottom_selected_idx);
+				dock_bottom_selected_idx = -1;
+				dock_popup_selected_idx = nrect; // Move to dock popup selected.
+				dock_select->queue_redraw();
+
+				update_dock_slots_visibility(true);
+
+				_edit_current();
+				emit_signal(SNAME("layout_changed"));
 			}
 
-			dock_popup_selected_idx = nrect;
-			dock_slot[nrect]->show();
-			dock_select->queue_redraw();
+			if (dock_popup_selected_idx != nrect) {
+				dock_slot[nrect]->move_tab_from_tab_container(dock_slot[dock_popup_selected_idx], dock_slot[dock_popup_selected_idx]->get_current_tab(), dock_slot[nrect]->get_tab_count());
 
-			update_dock_slots_visibility(true);
+				if (dock_slot[dock_popup_selected_idx]->get_tab_count() == 0) {
+					dock_slot[dock_popup_selected_idx]->hide();
+				} else {
+					dock_slot[dock_popup_selected_idx]->set_current_tab(0);
+				}
 
-			_edit_current();
-			emit_signal(SNAME("layout_changed"));
+				dock_popup_selected_idx = nrect;
+				dock_slot[nrect]->show();
+				dock_select->queue_redraw();
+
+				update_dock_slots_visibility(true);
+
+				_edit_current();
+				emit_signal(SNAME("layout_changed"));
+			}
 		}
 	}
 }
@@ -327,6 +362,44 @@ void EditorDockManager::_dock_make_selected_float() {
 	_edit_current();
 }
 
+void EditorDockManager::bottom_dock_show_placement_popup(const Rect2i &p_position, Control *p_dock) {
+	dock_bottom_selected_idx = bottom_docks.find(p_dock);
+	ERR_FAIL_COND(dock_bottom_selected_idx == -1);
+	dock_popup_selected_idx = -1;
+	dock_to_bottom->hide();
+
+	Vector2 popup_pos = p_position.position;
+	popup_pos.y += p_position.size.height;
+
+	if (!EditorNode::get_singleton()->get_gui_base()->is_layout_rtl()) {
+		popup_pos.x -= dock_select_popup->get_size().width;
+		popup_pos.x += p_position.size.width;
+	}
+	dock_select_popup->set_position(popup_pos);
+	dock_select_popup->popup();
+	if (dock_float) {
+		dock_float->hide();
+	}
+	dock_tab_move_right->hide();
+	dock_tab_move_left->hide();
+}
+
+void EditorDockManager::_dock_move_selected_to_bottom() {
+	Control *dock = dock_slot[dock_popup_selected_idx]->get_current_tab_control();
+	dock_slot[dock_popup_selected_idx]->remove_child(dock);
+
+	dock->call("_set_dock_horizontal", true);
+
+	bottom_docks.push_back(dock);
+	EditorNode::get_singleton()->add_bottom_panel_item(dock->get_name(), dock, true);
+	dock_select_popup->hide();
+	update_dock_slots_visibility(true);
+	_edit_current();
+	emit_signal(SNAME("layout_changed"));
+
+	EditorNode::get_singleton()->make_bottom_panel_item_visible(dock);
+}
+
 void EditorDockManager::_dock_make_float(Control *p_dock, int p_slot_index, bool p_show_window) {
 	ERR_FAIL_NULL(p_dock);
 
@@ -434,6 +507,16 @@ void EditorDockManager::save_docks_to_config(Ref<ConfigFile> p_layout, const Str
 
 	p_layout->set_value(p_section, "dock_floating", floating_docks_dump);
 
+	Array bottom_docks_dump;
+
+	for (Control *bdock : bottom_docks) {
+		Control *dock = bdock;
+		String name = dock->get_name();
+		bottom_docks_dump.push_back(name);
+	}
+
+	p_layout->set_value(p_section, "dock_bottom", bottom_docks_dump);
+
 	for (int i = 0; i < vsplits.size(); i++) {
 		if (vsplits[i]->is_visible_in_tree()) {
 			p_layout->set_value(p_section, "dock_split_" + itos(i + 1), vsplits[i]->get_split_offset());
@@ -464,6 +547,7 @@ void EditorDockManager::load_docks_from_config(Ref<ConfigFile> p_layout, const S
 
 			// FIXME: Find it, in a horribly inefficient way.
 			int atidx = -1;
+			int bottom_idx = -1;
 			Control *node = nullptr;
 			for (int k = 0; k < DOCK_SLOT_MAX; k++) {
 				if (!dock_slot[k]->has_node(name)) {
@@ -492,6 +576,18 @@ void EditorDockManager::load_docks_from_config(Ref<ConfigFile> p_layout, const S
 					}
 				}
 			}
+
+			if (atidx == -1) {
+				// Try bottom docks.
+				for (Control *bdock : bottom_docks) {
+					if (bdock->get_name() == name) {
+						node = bdock;
+						bottom_idx = bottom_docks.find(node);
+						break;
+					}
+				}
+			}
+
 			if (!node) {
 				// Well, it's not anywhere.
 				continue;
@@ -505,6 +601,11 @@ void EditorDockManager::load_docks_from_config(Ref<ConfigFile> p_layout, const S
 				dock_slot[i]->move_tab_from_tab_container(dock_slot[atidx], dock_slot[atidx]->get_tab_idx_from_control(node), 0);
 				dock_slot[i]->set_block_signals(false);
 				dock_slot[atidx]->set_block_signals(false);
+			} else if (bottom_idx != -1) {
+				bottom_docks.erase(node);
+				EditorNode::get_singleton()->remove_bottom_panel_item(node);
+				dock_slot[i]->add_child(node);
+				node->call("_set_dock_horizontal", false);
 			}
 
 			WindowWrapper *wrapper = Object::cast_to<WindowWrapper>(node);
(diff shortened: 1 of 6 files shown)


YeldhamDev:
The "Make Floating" button should be expanded horizontally now, so its size isn't at odds with the new one.

ElvisVilla:
That is really helpful for those that like the horizontal File System, in my case what i would love to have is the vertical panel but the feature to but collapse it. That would be increidibly helpful to have more space whenever i need it. 

My work around is the NV addon just cuz i can collapse the file system to the bottom and hide it presing the tab, but i would love to be able to just collapse without send it to the bottom :)

KoBeWi:
Moving the dock breaks icons:

https://github.com/godotengine/godot/assets/2223172/7880afd8-3280-4a85-8ce0-e3f6c9740011

The minimum size is very small. This is relevant when your bottom dock is at minimum size. After moving file dock to bottom, it's awkwardly small:
image

The layout of elements needs some tweaking:
image
File path and filter do not need to be that big and the icons are misaligned. I'd just put every button/text field in the same row, there is enough space for that. The layout from the side dock does not look that good at the bottom.

Also any reason why Make Floating button does not appear when dock is at the bottom?

editor/editor_node.cpp
```diff
@@ -5677,16 +5861,36 @@ void EditorNode::_scene_tab_closed(int p_tab) {
...
-	Button *tb = memnew(Button);
+class EditorBottomDockButton : public Button {
+	GDCLASS(EditorBottomDockButton, Button)
+
+	static void _bind_methods() {
+		ADD_SIGNAL(MethodInfo("dropping"));
+	}
+
+public:
+	bool can_drop_data(const Point2 &p_point, const Variant &p_data) const {
```
Chubercik, line 5872:
```suggestion
	bool can_drop_data(const Point2 &p_point, const Variant &p_data) const override {
```

Chubercik (changes requested):

viksl:
1. Regarding the minimum size if it's only for when it's first time dropepd to the bottom then I think it's ok to make it a bit larger but I wouldn't set absolute minimum size so the dock couldn't be minimized below that since this will often be opened while working with the viewport I can already imagine people just quickly dropping the size down while they work and pulling it up when needed (many people do this in Blender with animaiton and pretty much any separated views it's quick and reasonably comfortable to do so when you need more space). I understand you can make it go all away but that will unnecessarily inconvenient people.

2. Overal I like this, it makes sense since FS needs to be accessible with pretty much every dock at the same time so this will allow people to streamline their workflow, love it.

3. One more thing I'd like to see is a hide/show shortcut especially now since this (naturally) reduces the viewport size would it be possible to add a customizable shortcut for people (which as default could even be left empty) to show/hide the FS dock when it's at the bottom panel? Some practical examples for this would be Unreal Engine where I believe it's control + spacebar (or something like that) and it really seems very useful.

4. The icons at least the red one in the corner for how the FS should appear (list vs icons) if clicked on will get the icons back properly btw.

reduz:
@KoBeWi 

> Also any reason why Make Floating button does not appear when dock is at the bottom?

It seems when something floats, it still belongs to one of the side docks, so making something from the bottom dock float would not have any reference to that. It would need to be something separate to make the bottom docks float eventually, hence out of the scope of this PR.

PLyczkowski:
IMO each band-aid like this, while lovely for current users, decreases editor intuitiveness for new users, which in this case would expect to relocate tabs via tab dragging, both between areas and between windows, and the long term goal should be aiming for a flexible tabbed approach. Relevant proposals https://github.com/godotengine/godot-proposals/issues/7233 https://github.com/godotengine/godot-proposals/issues/1508 https://github.com/godotengine/godot-proposals/issues/28

reduz:
@PLyczkowski To you it may be a band aid, but I look at it different:

* For us contributors it keeps the code manageable because it is not a lot. If everything was super flexible as you suggest, it would be considerably more and more complex code.
* If this flexibility is not really needed, then it is better to just allow what has specific demand, rather than go for a very flexible solution. This, again, keeps everything manageable. Flexibility has to be justified.

reduz:
Just to clarify, detaching bottom panels may happen at some point in the future, but there is not nearly as much demand, so it's on low priority.

editor/editor_node.cpp
AThousandShips:
```suggestion
	virtual bool can_drop_data(const Point2 &p_point, const Variant &p_data) const override {
```
In fact

jcostello:
What about a key bind to toogle on and off the file system dock?

reduz:
@jcostello I suggest opening a proposal for this that explains how it would work with the proper context. I am not very famliar with that area (how keyboard shortcuts are organized), and you can get better feedback from other people.

reduz:
Okay I updated this pull request with the issues mentioned above fixed. I also did a bit of an extra effort to allow other docks to be moved to horizontal position and made the system more generic (though they have to explicitly allow this, only the filesystem one does currently).

I am having trouble figuring out why the icons break, because they don't when the dock is moved to other parts. Maybe someone more familiar with this area can lend a hand.

YuriSizov:
I would prefer if we implemented this after https://github.com/godotengine/godot/pull/84193 is merged. We need to de-bloat `EditorNode`, not add more code to it.

reduz:
@YuriSizov That's fair, and the other one looks very nice, what's blocking it?

YuriSizov:
@reduz Nothing, just wasn't in scope for 4.2 and I didn't get a chance to review it yet. But I will!

YuriSizov:
I fixed the missing icon issue for you. The theming code in the FS dock was a mess. Also added missing docs, which failed the CI.

Chubercik (approved):

reduz:
@YuriSizov Thanks hugely!

KoBeWi:
Well I did say the bottom dock was too small, but now it takes half of the screen 🙃
image
It can't be shrunk past this size.

The icon bug is fixed, and the layout adjustments I mentioned would require some more changes, so they can be done later.

doc/classes/FileSystemDock.xml
```diff
@@ -33,6 +33,12 @@
 		</method>
 	</methods>
 	<signals>
+		<signal name="bottom_dock_change_placement">
+			<param index="0" name="global_rect" type="Rect2" />
+			<description>
+				Emitted the "Change Dock Placement" button is clicked in the bottom dock mode.
```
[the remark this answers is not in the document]
YuriSizov: I don't think Juan added it with public consumption in mind. Maybe there is another way to achieve what he needs here.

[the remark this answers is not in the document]
kitbdev: EditorDockManager is a singleton, so I think we can just call `filesystem_dock_show_placement_popup()` directly.

editor/editor_dock_manager.cpp
```diff
@@ -327,6 +362,44 @@ void EditorDockManager::_dock_make_selected_float() {
 	_edit_current();
 }
 
+void EditorDockManager::filesystem_dock_show_placement_popup(const Rect2i &p_position) {
```
KoBeWi, line 365: If the system is meant more generic, this method shouldn't specifically target FileSystemDock.

editor/filesystem_dock.cpp
```diff
@@ -494,43 +494,22 @@ void FileSystemDock::_update_display_mode(bool p_force) {
 
 void FileSystemDock::_notification(int p_what) {
 	switch (p_what) {
-		case NOTIFICATION_TRANSLATION_CHANGED:
-		case NOTIFICATION_LAYOUT_DIRECTION_CHANGED:
 		case NOTIFICATION_ENTER_TREE: {
 			if (initialized) {
 				return;
 			}
 			initialized = true;
```
KoBeWi, line 501: Since this is used purely for one-time initialization, it could just use `READY`, no?

editor/filesystem_dock.cpp
```diff
@@ -3634,6 +3619,54 @@ MenuButton *FileSystemDock::_create_file_menu_button() {
...
+
+		FileSystemDock::DisplayMode new_display_mode = FileSystemDock::DISPLAY_MODE_HSPLIT;
+		FileSystemDock::FileListDisplayMode new_file_display_mode = FileSystemDock::FILE_LIST_DISPLAY_THUMBNAILS;
+
+		if (has_meta("_bottom_display_mode")) {
+			new_display_mode = FileSystemDock::DisplayMode(int(get_meta("_bottom_display_mode")));
+		}
+		if (has_meta("_bottom_file_display_mode")) {
+			new_file_display_mode = FileSystemDock::FileListDisplayMode(int(get_meta("_bottom_file_display_mode")));
+		}
```
KoBeWi, line 3643: This can be simplified to
```suggestion
		FileSystemDock::DisplayMode new_display_mode = FileSystemDock::DisplayMode(int(get_meta("_bottom_display_mode", int(FileSystemDock::DISPLAY_MODE_HSPLIT))));
		FileSystemDock::FileListDisplayMode new_file_display_mode = FileSystemDock::FileListDisplayMode(int(get_meta("_bottom_file_display_mode", int(FileSystemDock::FILE_LIST_DISPLAY_THUMBNAILS))));
```
(some casts can be dropped probably)

KoBeWi:
I tested whether dock position and split mode are remembered between editor sessions and it seems to be working mostly correctly. The split mode is remembered for the current position (i.e. side/bottom), but not for the other position. The setting is lost when restarting the editor. It's a minor issue and can be fixed later, just wanted to note it.

editor/filesystem_dock.cpp
```diff
@@ -3796,13 +3833,21 @@ FileSystemDock::FileSystemDock() {
 	button_toggle_display_mode->set_flat(true);
 	toolbar_hbc->add_child(button_toggle_display_mode);
 
+	button_dock_placement = memnew(Button);
+	button_dock_placement->set_flat(true);
+	button_dock_placement->set_tooltip_text(TTR("Change Dock Placement"));
```
KoBeWi, line 3838: I found it odd that other placement buttons don't have tooltips 🤔

editor/editor_dock_manager.cpp
```diff
@@ -179,23 +194,43 @@ void EditorDockManager::_dock_select_input(const Ref<InputEvent> &p_input) {
...
-				dock_slot[dock_popup_selected_idx]->hide();
-			} else {
-				dock_slot[dock_popup_selected_idx]->set_current_tab(0);
+				bottom_docks[dock_bottom_selected_idx]->call("_set_dock_horizontal", false);
+
+				dock_slot[nrect]->add_child(bottom_docks[dock_bottom_selected_idx]);
+				dock_slot[nrect]->show();
+				bottom_docks.remove_at(dock_bottom_selected_idx);
+				dock_bottom_selected_idx = -1;
+				dock_popup_selected_idx = nrect; // move to dock popup selected
```
KoBeWi, line 207:
```suggestion
				dock_popup_selected_idx = nrect; // Move to dock popup selected.
```

editor/editor_dock_manager.cpp
```diff
@@ -434,6 +507,18 @@ void EditorDockManager::save_docks_to_config(Ref<ConfigFile> p_layout, const Str
...
 	p_layout->set_value(p_section, "dock_floating", floating_docks_dump);
 
+	Array bottom_docks_dump;
+
+	for (Control *bdock : bottom_docks) {
+		Control *dock = bdock;
+
+		String name = dock->get_name();
+
+		bottom_docks_dump.push_back(name);
```
KoBeWi, line 517:
```suggestion
		Control *dock = bdock;
		String name = dock->get_name();
		bottom_docks_dump.push_back(name);
```
Too much empty space.

KoBeWi:
Re: making the panel floating - Apparently it's possible to move bottom dock directly to floating dock if you change editor layout. So it's definitely possible to provide this option in the new popup.

editor/editor_dock_manager.cpp
```diff
@@ -527,6 +630,46 @@ void EditorDockManager::load_docks_from_config(Ref<ConfigFile> p_layout, const S
 		}
 	}
 
+	Array dock_bottom = p_layout->get_value(p_section, "dock_bottom", Array());
+	for (int i = 0; i < dock_bottom.size(); i++) {
+		String name = dock_bottom[i];
```
KoBeWi, line 635:
```suggestion
		const String &name = dock_bottom[i];
```

editor/editor_node.cpp
```diff
@@ -5137,16 +5137,36 @@ void EditorNode::_scene_tab_closed(int p_tab) {
...
+	static void _bind_methods() {
+		ADD_SIGNAL(MethodInfo("dropping"));
+	}
+
+public:
+	virtual bool can_drop_data(const Point2 &p_point, const Variant &p_data) const override {
+		if (!is_pressed()) {
+			const_cast<EditorBottomDockButton *>(this)->emit_signal("dropping");
+		}
+		return true;
```
KoBeWi, line 5152:
```suggestion
		return false;
```
`true` does not make sense.

editor/editor_node.cpp
```diff
@@ -5137,16 +5137,36 @@ void EditorNode::_scene_tab_closed(int p_tab) {
 	scene_tabs->update_scene_tabs();
 }
 
-Button *EditorNode::add_bottom_panel_item(String p_text, Control *p_item) {
-	Button *tb = memnew(Button);
+class EditorBottomDockButton : public Button {
```
KoBeWi, line 5140: You could use drag forwarding instead of adding a new class.

KoBeWi:
I don't know what causes this, but something is wrong with drag and drop on buttons:

https://github.com/godotengine/godot/assets/2223172/bfdd2147-440b-4f75-8756-47d048f21dac

editor/editor_node.cpp
reduz: Seemed more hacky to me, I think the new class is simpler.

editor/filesystem_dock.cpp
```diff
@@ -3634,6 +3619,54 @@ MenuButton *FileSystemDock::_create_file_menu_button() {
...
+		if (has_meta("_bottom_display_mode")) {
+			new_display_mode = FileSystemDock::DisplayMode(int(get_meta("_bottom_display_mode")));
+		}
+		if (has_meta("_bottom_file_display_mode")) {
+			new_file_display_mode = FileSystemDock::FileListDisplayMode(int(get_meta("_bottom_file_display_mode")));
+		}
+
+		set_display_mode(new_display_mode);
+		set_file_list_display_mode(new_file_display_mode);
+		set_custom_minimum_size(Size2(0, 400) * EDSCALE);
```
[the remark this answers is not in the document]
reduz: If you feel this is ok, I'm good, it seemd too squashed to me.

reduz:
@KoBeWi whoa that is super weird, I can't reproduce it

editor/editor_node.cpp
KoBeWi: How is adding a method more hacky than a mini-class defined in cpp file 🙃

KoBeWi:
I can reproduce it every time.
Windows 10
1920x1080
100% display scale
idk what could affect it. Also not sure how relevant is it to this PR; technically it's a bug in the new functionality.

editor/editor_node.cpp
reduz: I guess just personal preference

reduz:
@KoBeWi Nevermind I am blind, I also had the problem but did not notice it.

reduz:
@KoBeWi Okay, it looks the problem is because the order can change of bottom panel items and the signal is being connected by item index. I changed the code to simply use the pointer, so it seems to work fine now.

I will clean up the rest of the things you mentioned and hopefully this should be ready to go.

doc/classes/FileSystemDock.xml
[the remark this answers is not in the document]
reduz: Ah yeah it sounds like with EditorDockManager being a singleton now this can be simplified.

editor/filesystem_dock.cpp
reduz: Ah likely it ended up like this due to the refactor.

editor/filesystem_dock.cpp
reduz: This is probably part of the original code there that was just refactored? I think Its beyond the scope of the PR

reduz:
Allright, all changes done, please give a final check and feel free to approve if you believe nothing else remains to be done.

Chubercik (approved):

editor/editor_dock_manager.cpp
```diff
@@ -327,6 +362,44 @@ void EditorDockManager::_dock_make_selected_float() {
 	_edit_current();
 }
 
+void EditorDockManager::bottom_dock_show_placement_popup(const Rect2i &p_position) {
+	dock_bottom_selected_idx = bottom_docks.find(FileSystemDock::get_singleton());
```
KoBeWi, line 366: That's still hard-coded FileSystem dock 🙃
You could take the dock as argument, which would allow every dock to handle itself.

editor/filesystem_dock.cpp
```diff
@@ -3634,6 +3620,47 @@ MenuButton *FileSystemDock::_create_file_menu_button() {
...
+
+		FileSystemDock::DisplayMode new_display_mode = FileSystemDock::DISPLAY_MODE_TREE_ONLY;
+		FileSystemDock::FileListDisplayMode new_file_display_mode = FileSystemDock::FILE_LIST_DISPLAY_LIST;
+
+		if (has_meta("_dock_display_mode")) {
+			new_display_mode = FileSystemDock::DisplayMode(int(get_meta("_dock_display_mode")));
+		}
+		if (has_meta("_dock_file_display_mode")) {
+			new_file_display_mode = FileSystemDock::FileListDisplayMode(int(get_meta("_dock_file_display_mode")));
+		}
```
KoBeWi, line 3654: This is what I meant by "Same below" in https://github.com/godotengine/godot/pull/86765#discussion_r1462403615

editor/editor_dock_manager.cpp
reduz: ah shit did not pay attention to this. heh.

editor/filesystem_dock.cpp
reduz: oh oops, I missed that, sorry

KoBeWi (approved):
I left some minor comments., but functionality-wise it's fine now.

reduz:
Alright, done with the suggested changes.

YuriSizov:
Rebased after recent changes and also fixed another missing icon bug. Seems like a check was added in https://github.com/godotengine/godot/pull/23273 to avoid segfault on exit, but this should already be covered in 4.x where `NOTIFICATION_THEME_CHANGED` won't be sent on exit.

YuriSizov (approved):
Looks good, works good! The placement popup needs some rework, but it's a pre-existing problem.

Also we should extract bottom panel logic from EditorNode next, cc @kitbdev 😉

reduz:
Sounds good! I need to give a pass to all the editors in the bottom panel and make sure they can be undocked and redocked at some point, so at some point we can start thinking how to customize their placement too.

akien-mga:
Thanks!

Zireael07:
@ismailgamedev Commenting on closed PR is unlikely to be seen. This should probably be a separate issue instead

PR #86766, opened 2024-01-03
Add `Engine.get_{process,physics}_step()` to get unscaled time delta

Calinou:
This can be used for manual UI animations or code whose behavior shouldn't vary depending on `Engine.time_scale`.

This also clarifies the `physics_steps_per_frame` documentation with regards to fixed FPS behavior.

- This partially addresses https://github.com/godotengine/godot-proposals/issues/7775
  and https://github.com/godotengine/godot-proposals/issues/6809.

**Testing project:** [test_unscaled_delta.zip](https://github.com/godotengine/godot/files/13824084/test_unscaled_delta.zip)

## Preview

https://github.com/godotengine/godot/assets/180032/18e24602-7b51-49a1-9255-5e560a1fdfae


lawnjelly:
This looks great, I'm wondering whether a name like `get_physics_step_unscaled()` might be more informative?

Otherwise aside from reading the docs, I don't see any reason why the user would expect this value to be any different to the value passed to `physics_process()`.

scene/resources/resource_format_text.cpp
AThousandShips, line 1: These changes looks unrelated, accidentally included?

doc/classes/Node.xml
```diff
@@ -81,7 +81,7 @@
 			<return type="void" />
 			<param index="0" name="delta" type="float" />
 			<description>
-				Called during the processing step of the main loop. Processing happens at every frame and as fast as possible, so the [param delta] time since the previous frame is not constant. [param delta] is in seconds.
+				Called during the processing step of the main loop. Processing happens at every frame and as fast as possible, so the [param delta] time since the previous frame is not constant. [param delta] is in seconds and is multiplied by [member Engine.time_scale].To get the [param delta] that isn't scaled with [member Engine.time_scale], use [method Engine.get_process_step].
```
AThousandShips, line 84:
```suggestion
				Called during the processing step of the main loop. Processing happens at every frame and as fast as possible, so the [param delta] time since the previous frame is not constant. [param delta] is in seconds and is multiplied by [member Engine.time_scale]. To get the [param delta] that isn't scaled with [member Engine.time_scale], use [method Engine.get_process_step].
```

doc/classes/Node.xml
```diff
@@ -71,7 +71,7 @@
 			<return type="void" />
 			<param index="0" name="delta" type="float" />
 			<description>
-				Called during the physics processing step of the main loop. Physics processing means that the frame rate is synced to the physics, i.e. the [param delta] variable should be constant. [param delta] is in seconds.
+				Called during the physics processing step of the main loop. Physics processing means that the frame rate is synced to the physics, i.e. the [param delta] variable should be constant. [param delta] is in seconds and is multiplied by [member Engine.time_scale]. To get the [param delta] that isn't scaled with [member Engine.time_scale], use [method Engine.get_physics_step].
```
AThousandShips, line 74:
```suggestion
				Called during the physics processing step of the main loop. Physics processing means that the frame rate is synced to the physics, i.e. the [param delta] variable should be constant. [param delta] is in seconds and is multiplied by [member Engine.time_scale]. To get the [param delta] that isn't scaled by [member Engine.time_scale], use [method Engine.get_physics_step].
```

PR #86768, opened 2024-01-03
Add `text_changed` signal to Label and RichTextLabel

### LINKED ISSUE #8790: Add changed_text signal to Label and RichTextLabel

An RPG where the UI contains Labels that should be decorated based on the text itself and updated when the text changes.

I want to add custom underlines to some of my labels, which get updated at runtime. 

To achieve this, I place a `NinePatchRectangle` under each line of the text, of which I want to update the position and scale accordingly to decorate the displayed text, as well as add or remove new ones if necessary.

The `Label` and `RichTextLabel` classes receive a `text_changed` signal that passes over the new text value.

1. Add the signal to `_bind_methods()` of `Label` and `RichTextLabel` :
	```c++
	void CLASSNAME::_bind_methods() {
		...
		ADD_SIGNAL(MethodInfo("changed_text", PropertyInfo(Variant::STRING, "text")));
		...
	}
	```
2. Emit the signal at the end of `set_text()` of both classes:
	```c++
	void CLASSNAME::set_text(const String &p_bbcode) {
		...
		emit_signal(SNAME("changed_text"), text);
	}
	```

- Proposed implementation: https://github.com/godotengine/godot/pull/86768

1. implementing `_set()` to check for when `text` is set and ensuring the text only gets changed using `set()`.
	**Drawback**: Difficult to ensure in bigger projects; Reduces code quality.

2. Comparing the text to a stored one in `_process()` and emitting your own signal for when the text was altered.
	**Drawback**: May slow down bigger projects.

3. Add in the signal to a clone of the repository and compile the engine yourself, as its only 4 lines of code that need to be added.
	**Drawback**: You need to rebuild every time you want to use a new version.

Has to be implemented in the built-in classes.

Justin113D:
- Resolves https://github.com/godotengine/godot-proposals/issues/8790

Added new signal to `Label` and `RichTextLabel`:
`changed_text`, which gets emitted in `set_text()` of each class.

### DIFF
--- a/doc/classes/RichTextLabel.xml
+++ b/doc/classes/RichTextLabel.xml
@@ -666,15 +666,15 @@
 		</member>
 	</members>
 	<signals>
-		<signal name="finished">
-			<description>
-				Triggered when the document is fully loaded.
-			</description>
-		</signal>
 		<signal name="changed_text">
 			<param index="0" name="text" type="String" />
 			<description>
 				Emitted when the text of a label was changed. The new text is contained in the [param text] argument.
+			</description>
+		</signal>
+		<signal name="finished">
+			<description>
+				Triggered when the document is fully loaded.
 			</description>
 		</signal>
 		<signal name="meta_clicked">
--- a/doc/classes/RichTextLabel.xml
+++ b/doc/classes/RichTextLabel.xml
@@ -669,7 +669,7 @@
 		<signal name="text_changed">
 			<param index="0" name="text" type="String" />
 			<description>
-				Emitted when the text of a label was changed. The new text is contained in the [param text] argument.
+				Emitted when the value behind the text property of a RichTextLabel was changed. The new text is contained in the [param text] argument.
 			</description>
 		</signal>
 		<signal name="finished">
--- a/doc/classes/RichTextLabel.xml
+++ b/doc/classes/RichTextLabel.xml
@@ -666,15 +666,15 @@
 		</member>
 	</members>
 	<signals>
+		<signal name="finished">
+			<description>
+				Triggered when the document is fully loaded.
+			</description>
+		</signal>
 		<signal name="text_changed">
 			<param index="0" name="text" type="String" />
 			<description>
 				Emitted when the value behind the text property of a RichTextLabel was changed. The new text is contained in the [param text] argument.
-			</description>
-		</signal>
-		<signal name="finished">
-			<description>
-				Triggered when the document is fully loaded.
 			</description>
 		</signal>
 		<signal name="meta_clicked">
--- a/doc/classes/RichTextLabel.xml
+++ b/doc/classes/RichTextLabel.xml
@@ -666,7 +666,7 @@
 		</member>
 	</members>
 	<signals>
-		<signal name="changed_text">
+		<signal name="text_changed">
 			<param index="0" name="text" type="String" />
 			<description>
 				Emitted when the text of a label was changed. The new text is contained in the [param text] argument.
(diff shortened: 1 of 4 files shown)


Justin113D:
The linux editor action failed on the `--doctool` diff check, and showed the `finished` signal doc for `RichTextLabel` being incorrectly placed within the xml, so i moved it to where the action expected it to be. I hope that fixes it.

scene/gui/rich_text_label.cpp
```diff
@@ -5574,6 +5574,7 @@ void RichTextLabel::set_text(const String &p_bbcode) {
 	}
 	text = p_bbcode;
 	_apply_translation();
+	emit_signal(SNAME("changed_text"), text);
```
bruvzg, line 5577: RTL has a lot of different methods that change the text, `parse_bbcode`, `bbcode`, `add_*`, `push_*` so adding it to `set_text` only will have inconsistent behavior.

Not sure where's the best place to do it, maybe `_add_item()`/`_remove_item` and an additional one in the `add_text` (in a branch where existing text item is updated).

Justin113D: I'm unsure of how the entire bbcode system works, but as far as I can tell, the actual value behind `text` of `RichTextLabel` does not get changed anywhere outside of `set_text()`. 

If the value of `text` hasn't changed, the signal `changed_text` should not be emitted.

bruvzg: > If the value of text hasn't changed, the signal changed_text should not be emitted.

I guess this makes sense. But it should be at least mentioned in the documentation.

doc/classes/RichTextLabel.xml
```diff
@@ -666,6 +666,12 @@
 		</member>
 	</members>
 	<signals>
+		<signal name="text_changed">
```
AThousandShips, line 669: This needs to move down, they must be sorted alphabetically

scene/gui/rich_text_label.cpp
Justin113D: > I guess this makes sense. But it should be at least mentioned in the documentation.

I updated the documentation a little so that this is clear

doc/classes/RichTextLabel.xml
Justin113D: Ahhh that explains why the action failed before. Fixed it

KoBeWi:
We normally don't emit signals in setters, because this can be done manually by the user when setting the value:
```GDScript
rtl.text = "text"
text_changed.emit("text")
```
(or similar)

In RichTextLabel you can argue that the text can be changed indirectly by other methods, but this PR does not handle that. The new signals are not very useful IMO.

Justin113D:
> We normally don't emit signals in setters, because this can be done manually by the user when setting the value:
> 
> ```gdscript
> rtl.text = "text"
> text_changed.emit("text")
> ```
> 
> (or similar)
> 
> In RichTextLabel you can argue that the text can be changed indirectly by other methods, but this PR does not handle that. The new signals are not very useful IMO.

Ah, that's a bummer. I was hoping i could have the signal while keeping on using the built-in `text` variable, but if that's generally not how godot handles setters, then I'll see how i can make do without it. Thanks regardless! 

I'll leave the PR open for a reviewer to close it.

KoBeWi:
As explained in https://github.com/godotengine/godot-proposals/issues/8790#issuecomment-2866258291, I'm closing this pull request. Thanks for the contribution nonetheless.

PR #86769, opened 2024-01-04
Add logging when macOS export will fail due to disabled texture formats.

StagnationPoint:
Currently if a required texture format is disabled during macOS export, the export will fail without logging any information indicating the real source of the problem. This change adds error messaging that will direct the user toward the issue.

### DIFF
--- a/platform/macos/export/export_plugin.cpp
+++ b/platform/macos/export/export_plugin.cpp
@@ -2051,13 +2051,17 @@ bool EditorExportPlatformMacOS::has_valid_export_configuration(const Ref<EditorE
 	String architecture = p_preset->get("binary_format/architecture");
 	if (architecture == "universal" || architecture == "x86_64") {
 		if (!ResourceImporterTextureSettings::should_import_s3tc_bptc()) {
+			err += TTR("Cannot export for universal or x86_64 if S3TC BPTC texture format is disabled. Enable it in the Project Settings (Rendering > Textures > VRAM Compression > Import S3TC BPTC).") + "\n";
 			valid = false;
 		}
-	} else if (architecture == "arm64") {
+	}
+	if (architecture == "universal" || architecture == "arm64") {
 		if (!ResourceImporterTextureSettings::should_import_etc2_astc()) {
+			err += TTR("Cannot export for universal or arm64 if ETC2 ASTC texture format is disabled. Enable it in the Project Settings (Rendering > Textures > VRAM Compression > Import ETC2 ASTC).") + "\n";
 			valid = false;
 		}
-	} else {
+	}
+	if (architecture != "universal" && architecture != "x86_64" && architecture != "arm64") {
 		ERR_PRINT("Invalid architecture");
 	}
 


platform/macos/export/export_plugin.cpp
```diff
@@ -2051,10 +2051,12 @@ bool EditorExportPlatformMacOS::has_valid_export_configuration(const Ref<EditorE
 	String architecture = p_preset->get("binary_format/architecture");
 	if (architecture == "universal" || architecture == "x86_64") {
 		if (!ResourceImporterTextureSettings::should_import_s3tc_bptc()) {
+			err += TTR("Cannot export for universal or x86_64 if S3TC BPTC texture format is disabled. Enable it in project settings.") + "\n";
 			valid = false;
 		}
 	} else if (architecture == "arm64") {
 		if (!ResourceImporterTextureSettings::should_import_etc2_astc()) {
+			err += TTR("Cannot export for arm64 if ETC2 ASTC texture format is disabled. Enable it in project settings.") + "\n";
```
akien-mga, line 2059: I wonder why this doesn't take `universal` into account. It's already handled in the branch above obviously, but shouldn't it requires both formats? @bruvzg

bruvzg: Yes, universal export should include both formats, so it should be `} else if (architecture == "universal" || architecture == "arm64") {`.

akien-mga: That condition would never be reached though as it already checks `universal` in the previous branch.

Should work if it's just `if` and not `else if` on the other hand.

akien-mga: Note for OP: Please make modifications by _amending_ the commit, and *force pushing* the changes, so that it stays as a single commit (see [PR workflow](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#modifying-a-pull-request)).

platform/macos/export/export_plugin.cpp
```diff
@@ -2051,10 +2051,12 @@ bool EditorExportPlatformMacOS::has_valid_export_configuration(const Ref<EditorE
 	String architecture = p_preset->get("binary_format/architecture");
 	if (architecture == "universal" || architecture == "x86_64") {
 		if (!ResourceImporterTextureSettings::should_import_s3tc_bptc()) {
+			err += TTR("Cannot export for universal or x86_64 if S3TC BPTC texture format is disabled. Enable it in project settings.") + "\n";
```
AThousandShips, line 2054:
```suggestion
			err += TTR("Cannot export for universal or x86_64 if S3TC BPTC texture format is disabled. Enable it in the Project Settings.") + "\n";
```
Following the same style as other messages, should probably also instruct how to find them

platform/macos/export/export_plugin.cpp
```diff
@@ -2051,13 +2051,17 @@ bool EditorExportPlatformMacOS::has_valid_export_configuration(const Ref<EditorE
 	String architecture = p_preset->get("binary_format/architecture");
 	if (architecture == "universal" || architecture == "x86_64") {
 		if (!ResourceImporterTextureSettings::should_import_s3tc_bptc()) {
+			err += TTR("Cannot export for universal or x86_64 if S3TC BPTC texture format is disabled. Enable it in the Project Settings (Rendering > Textures > VRAM Compression > Import S3TC BPTC)") + "\n";
```
akien-mga, line 2054:
```suggestion
			err += TTR("Cannot export for universal or x86_64 if S3TC BPTC texture format is disabled. Enable it in the Project Settings (Rendering > Textures > VRAM Compression > Import S3TC BPTC).") + "\n";
```

platform/macos/export/export_plugin.cpp
```diff
@@ -2051,13 +2051,17 @@ bool EditorExportPlatformMacOS::has_valid_export_configuration(const Ref<EditorE
...
 			valid = false;
 		}
-	} else if (architecture == "arm64") {
+	}
+	if (architecture == "universal" || architecture == "arm64") {
 		if (!ResourceImporterTextureSettings::should_import_etc2_astc()) {
+			err += TTR("Cannot export for universal or arm64 if ETC2 ASTC texture format is disabled. Enable it in the Project Settings (Rendering > Textures > VRAM Compression > Import ETC2 ASTC)") + "\n";
```
akien-mga, line 2060:
```suggestion
			err += TTR("Cannot export for universal or arm64 if ETC2 ASTC texture format is disabled. Enable it in the Project Settings (Rendering > Textures > VRAM Compression > Import ETC2 ASTC).") + "\n";
```

akien-mga (approved):
Looks good to me. Minor style nitpick.

AThousandShips (approved):

StagnationPoint:
Thanks for the guidance, all! Much appreciated.

akien-mga:

platform/macos/export/export_plugin.cpp
```diff
@@ -2051,13 +2051,17 @@ bool EditorExportPlatformMacOS::has_valid_export_configuration(const Ref<EditorE
 	String architecture = p_preset->get("binary_format/architecture");
 	if (architecture == "universal" || architecture == "x86_64") {
 		if (!ResourceImporterTextureSettings::should_import_s3tc_bptc()) {
+			err += TTR("Cannot export for universal or x86_64 if S3TC BPTC texture format is disabled. Enable it in the Project Settings (Rendering > Textures > VRAM Compression > Import S3TC BPTC).") + "\n";
```
aaronfranke, line 2054: > Enable it in the Project Settings [...]

Instead of requiring the user to dig into the settings, we should just provide a button to fix it. I implemented this in PR #78457, but the button seems to have disappeared somewhere in-between my PR and your PR? It's still present for Android, but missing for macOS for some weird reason.

PR #86772, opened 2024-01-04
Display functions that were previously forgotten in Profiler

Rindbee:
The array data in `add()` contains these data, but some data may have been forgotten to be converted into `ServerFunctionInfo`.

This results in some information not being displayed in the Profiler.

| Before | This PR |
| :-------: | :---------: |
|  |  |

Should the display of these functions be reordered? According to the order in the same frame. :thinking:

### DIFF
--- a/servers/debugger/servers_debugger.cpp
+++ b/servers/debugger/servers_debugger.cpp
@@ -342,10 +342,12 @@ public:
 		}
 		ServerInfo &srv = server_data[name];
 
-		ServerFunctionInfo fi;
-		fi.name = p_data[1];
-		fi.time = p_data[2];
-		srv.functions.push_back(fi);
+		for (int idx = 1; idx < p_data.size() - 1; idx += 2) {
+			ServerFunctionInfo fi;
+			fi.name = p_data[idx];
+			fi.time = p_data[idx + 1];
+			srv.functions.push_back(fi);
+		}
 	}
 
 	void tick(double p_frame_time, double p_process_time, double p_physics_time, double p_physics_frame_time) {


servers/debugger/servers_debugger.cpp
```diff
@@ -342,10 +342,13 @@ class ServersDebugger::ServersProfiler : public EngineProfiler {
...
-		ServerFunctionInfo fi;
-		fi.name = p_data[1];
-		fi.time = p_data[2];
-		srv.functions.push_back(fi);
+		int i = 1;
+		while (i < p_data.size() - 1) {
+			ServerFunctionInfo fi;
+			fi.name = p_data[i++];
+			fi.time = p_data[i++];
+			srv.functions.push_back(fi);
```
akien-mga, line 350: This is valid, but we tend to make increments separate from using variables as this otherwise adds cognitive overload to triple check whether we're pre or post-incrementing.

IMO this would communicate the intent better:
```suggestion
			ServerFunctionInfo fi;
			fi.name = p_data[i];
			fi.time = p_data[i + 1];
			srv.functions.push_back(fi);
			i += 2;
```

akien-mga: Or in this case even a `for` loop with `i += 2` increment.

akien-mga:
> Should the display of these functions be reordered? According to the order in the same frame. 🤔

What would be the frame order?
Or conversely, where does the current order stem from?

Rindbee:
https://github.com/godotengine/godot/blob/fbaab3cf537a892295aabdfd02c8052e370e6669/main/main.cpp#L3748
https://github.com/godotengine/godot/blob/fbaab3cf537a892295aabdfd02c8052e370e6669/main/main.cpp#L3771

The array data is ordered and passed when calling `GodotPhysicsServer2D::flush_queries()`, but `flush_queries()` is not at the end of the physical frame, so the data in the array is **not** the same frame's data.

https://github.com/godotengine/godot/blob/fbaab3cf537a892295aabdfd02c8052e370e6669/servers/physics_2d/godot_physics_server_2d.cpp#L1350-L1353

https://github.com/godotengine/godot/blob/fbaab3cf537a892295aabdfd02c8052e370e6669/servers/physics_2d/godot_physics_server_2d.cpp#L1354-L1358

So in a sense, the current data is also sorted according to chronological order.

Faless (approved):
LGTM, great work :trophy:

YuriSizov:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

Issue #86428, opened 2023-12-22
Audio input gets muted after a while on android


leoyanis:
### Tested versions

4.1.1,4.2 (all devs and betas), 4.2, 4.2.1
mono and regular, tried them all

### System information

MacOS, Sonoma 14, M1 pro -> Problem on Android (tested on 10 - 13)

### Issue description

When recording audio with this minimal project, the audio input gets muted for no apparent reason.
This was a problem on many of my projects and i finally made a minimal project to troubleshoot and same issue.
This only happens on Android, but not on my mac.

https://github.com/godotengine/godot/assets/144027730/ee85caa2-61bf-4dc5-ba8e-e2eca3931ef7

In the video above i was talking and making noise, the bar at the top displays the volume.
You will see it drop to 0 permanently after a while, this was not from me stopping but from the app not detecting any audio input anymore.

The audio file is muted from that part on as you'd expect.

I have tried messing around with mix rate, stereo, other random settings, audio focus an threads but with no luck.

I believe there is some issue in the godot backend

### Steps to reproduce

Install android build template on the minimal project.
Install on android.
Start recording.
Watch how it just mutes permanently after a while.
The muted thing resets after restarting the audio recorder.

### Minimal reproduction project (MRP)

[AudioRecord test.zip](https://github.com/godotengine/godot/files/13751660/AudioRecord.test.zip)

ellenhp:
#73162 would almost certainly have fixed this, you can try it as a workaround if you're okay building the engine yourself. It should have no adverse effects, it's just a fix for an oversight on my part in #51296. Juan decided that we should fix this class of bug in the streams though so this will require a contributor to add underrun handling to the microphone audio stream.

bestvcboy:
I also encountered the same problem under Android, which is particularly unstable when recording. After recording for a while, I found that:
var indx = AudioServer.GetBusIndex("Record");
float volumeDb1 = AudioServer.GetBusPeakVolumeLeftDb(indx, 0);
float volumeDb2 = AudioServer.GetBusPeakVolumeRightDb(indx, 0);
volumeDb1 = -200; volumeDb2 = -200, And then it won't be recorded.

TrevorBlythe:
Pretty sure this is also causing the mic input to stop working the second time after opening an app in XR.

JackeHsu:
[...]

PR #86776, opened 2024-01-04
Replace the current audio driver with Oboe on Android

### LINKED ISSUE #2358: Integrate Oboe for audio on Android

Godot Audio on Android

Godot's audio setup on Android currently uses [OpenSL ES](https://developer.android.com/ndk/guides/audio/opensl), an older audio API supported on Android API 9 and above. This API has high device coverage, but is bulkier and less suitable for low-latency applications compared to the native [AAudio](https://developer.android.com/ndk/guides/audio/aaudio/aaudio) API introduced in Android API 26.

[Oboe](https://github.com/google/oboe) is Android's recommended modern audio API, acting as a C++ wrapper library that chooses between OpenSL ES and AAudio depending on device support and best performance. Oboe is compatible with API 16 onwards (covers 99% of Android devices). We expect this to modernize Godot's audio setup on Android, with lower audio latency and minimal loss of device coverage.

We will create a class `AudioDriverOboe` housed under `godot/platform/android`, which will take the place of `AudioDriveOpenSL` in `OS_Android`. Much of its functionality will mirror `AudioDriverOpenSL`, informed by Oboe's [migration guide](https://github.com/google/oboe/blob/master/docs/OpenSLESMigration.md).

Steps:
1. Add Oboe to `godot/thirdparty` and link with SCons build system.
2. Implement `AudioDriverOboe` with the set of methods inherited from `AudioDriver`, and modify `os_android.h` to use it. More implementation details are available if requested.
3. Develop test apps for latency measurement and potentially device coverage.

`AudioDriverOboe` Member variables:
```
Mutex *mutex
static AudioDriverOboe *s_ad (singleton)
AudioStream player
AudioStream recorder
bool active
bool pause
```

`AudioDriverOboe` Implemented API:
```
const char *get_name()
Error init()
void start()
int get_mix_rate()
SpeakerMode get_speaker_mode()
void lock()
void unlock()
void finish()
void set_pause(bool p_pause)
Error capture_start()
Error capture_stop()
Constructor AudioDriverOboe()
```

For latency testing, the simplest version is the [Larsen test](https://source.android.com/devices/audio/latency/measure#larsenTest) with two AudioStreamPlayer nodes, one that records audio input and one that plays the input continuously. By measuring the time between sounds, we capture the sum of input and output latency, with some device/engine overhead. This should serve as an adequate measure for approximate latency improvement.

For testing device coverage, we should have a test app that covers audio features on Godot. This may be using the [mic_record](https://github.com/godotengine/godot-demo-projects/tree/master/audio/mic_record) demo or some variation upon it which will use both input and output. If run on some device farm or set of devices, we may use app crashes or bad sound behavior to verify minimal loss in coverage.

N/A. This is expected to be used for all Godot games exported for Android.

This proposal changes the audio (a core element of the engine) intending to impact all games for the Android platform.

Android Games DevTech Team @ Google

kus04e4ek:
Closes: [godot-proposals#2358](https://github.com/godotengine/godot-proposals/issues/2358)

[Sample project 5](https://github.com/godotengine/godot/files/14546749/sample.zip)

**TODO:**
- [x] https://github.com/google/oboe/blob/main/docs/GettingStarted.md#obtaining-optimal-latency
- [x] ~Get the correct number of frames per callback and for output.~ Went with the different idea.
- [x] ~Get the correct number of frames per callback for input.~ Went with the different idea.
- [x] Handle audio devices switching.
- [x] Rewrite SCsub.
- [x] ~Properly update device list after connecting or disconnecting devices.~ Went with the different idea.
- [x] Delete not needed Oboe files.

**Known Issues:**
- [x] OpenSL not working. Tested on the virtual machine.
- [x] Bad audio quality on machines with support only for in16_t format. ~Couldn't reproduce it on my phone.~ (My phone supports int32_t format, I just didn't know). Tested on the virtual machine.
- [x] ~Audio not working on the physical device.~ An issue with my phone.
- [x] Microphone not working.
- [x] Connecting a new default device will be ignored if done outside the app.
- [x] Stream closes outside the app and doesn't reopen if other stream started by the other app.

See [the comment](https://github.com/godotengine/godot/pull/86776#issuecomment-1986911325) for new TODO/TODO after merge list.


AThousandShips:
You will also need to add the library to the list in `COPYRIGHT.txt`, instructions are available there

platform/android/audio_driver_oboe.h
```diff
@@ -0,0 +1,140 @@
...
+	bool pause = false;
+
+	std::shared_ptr<oboe::AudioStream> audio_output;
+	std::shared_ptr<oboe::AudioStream> audio_input;
+
+	std::shared_ptr<OutputErrorCallback> output_error_callback;
+	std::shared_ptr<InputErrorCallback> input_error_callback;
+
+	std::shared_ptr<oboe::AudioStreamDataCallback> audio_output_callback;
+	std::shared_ptr<oboe::AudioStreamDataCallback> audio_input_callback;
```
AThousandShips, line 102: We don't use the standard library normally, see [here](https://docs.godotengine.org/en/stable/contributing/development/cpp_usage_guidelines.html#standard-template-library), would need to be discussed here depending on the library using it, no other engine component uses this

kus04e4ek: Yeah, I know. Oboe uses shared_ptr and regular pointers, but methods accepting regular pointers are marked as deprecated, when I think about it again, it probably will be fine to use regular pointers, it just might create problems when closing the app (Through is this important?)

kus04e4ek:
> * Audio not working on the physical device.

Seems to be a problem with my device. Opened [an issue](https://github.com/google/oboe/issues/1953) on oboe repository.

platform/android/audio_driver_oboe.h
kus04e4ek: Will test latter if something is printed in the console, if variables are just pointers

kus04e4ek: Checked when shred_ptr was added, it was in the [google/oboe#821](https://github.com/google/oboe/pull/821) to fix an issue with the data race, without shared_ptr the data race will still occur, so I think shared_ptr should stay

doc/classes/AudioServer.xml
```diff
@@ -132,7 +132,7 @@
 		<method name="get_output_latency" qualifiers="const">
 			<return type="float" />
 			<description>
-				Returns the audio driver's effective output latency. This is based on [member ProjectSettings.audio/driver/output_latency], but the exact returned value will differ depending on the operating system and audio driver.
+				Returns the audio driver's effective output latency (in seconds). This is based on [member ProjectSettings.audio/driver/output_latency], but the exact returned value will differ depending on the operating system and audio driver.
```
kus04e4ek, line 135: I think it's important to clarify, since in settings the latency is in milliseconds, and it was not very clear for me what actually was returned

platform/android/string_android.h
```diff
@@ -58,4 +58,23 @@ static inline String jstring_to_string(jstring source, JNIEnv *env = nullptr) {
...
+ * @return Godot string instance.
+ */
+static inline String char_sequence_to_string(jobject source, JNIEnv *env = nullptr) {
+	if (source) {
+		if (!env) {
+			env = get_jni_env();
+		}
+
+		jclass cCharSequence = env->GetObjectClass(source);
+		jmethodID toString = env->GetMethodID(cCharSequence, "toString", "()Ljava/lang/String;");
```
kus04e4ek, line 74: Camel Case to be consistent with "jni_utils.cpp", e.g.
```
jmethodID getName = env->GetMethodID(cclass, "getName", "()Ljava/lang/String;");
jstring clsName = (jstring)env->CallObjectMethod(cls, getName);
```
But that code is pretty old, maybe that code should be more consistent and not the other way around.

platform/android/java/lib/src/org/godotengine/godot/Godot.kt
```diff
@@ -336,12 +337,14 @@ class Godot(private val context: Context) : SensorEventListener {
 		}
 
 		if (nativeLayerInitializeCompleted && !nativeLayerSetupCompleted) {
-			nativeLayerSetupCompleted = GodotLib.setup(commandLine.toTypedArray(), tts)
+			var audioManager: AudioManager = context.getSystemService(Context.AUDIO_SERVICE) as AudioManager
+			nativeLayerSetupCompleted = GodotLib.setup(commandLine.toTypedArray(), tts, audioManager)
```
kus04e4ek, line 341: `context.getSystemService(Context.AUDIO_SERVICE)` can be called on the C++ side, but the code would be harder to understand and longer, this is just here for readability. Will change if needed

kus04e4ek:
I think the PR ready. There are just a few issues that are worth addressing:
- Not all audio device types are supported when using set_device. I think it's fine to leave it like this for now? I would love to do a follow-up PR latter, but I could always finish it in this PR, if that's important. I'm just not sure yet what would be the right approach for every type. Some types might start working only when specific conditions are met, so marked those as not working for now.
- Not all audio device types are tested when using set_device. I just don't have them. So, if anyone wants to help me with that, then I would be happy to accept your help. Currently, tested are: built-in_earpiece (not working), built-in_speaker (working), bluetooth_sco (working?), bluetooth_a2dp (working), built-in_microphone (working), fm_tuner (working? Tested only on the emulator, and there was just a high-pitch sound, I guess it should be like that?), telephony (not working), remote_submix (not working), echo_reference (not expected to work, it's reserved for system apps only).
- SCO support is experimental? Right now, when setting the device on one of the streams, the device on the other would be set too. I did it that way, because turning on SCO for one stream would turn on SCO for the other. Also, when turning on SCO, the default device used by Android changes, so did it for the consistency on what the default device is. I don't think that any other Audio Driver changes devices for both streams on one set_device call, so I would appreciate some feedback.

Issues above should all be fine when not using set_device, I assume that the default device picked by Android always works, so maybe it would be better to disable it for now?

And some theoretically possible feature, that I could do in a follow-up, is very limited set_device for OpenSL.

Also, #88628 could be implemented for Android too. For the same reasons as the ones why is it implemented for macOS.

doc/classes/OS.xml
```diff
@@ -653,7 +653,7 @@
 			<param index="0" name="name" type="String" />
 			<description>
 				Requests permission from the OS for the given [param name]. Returns [code]true[/code] if the permission has been successfully granted.
-				[b]Note:[/b] This method is currently only implemented on Android, to specifically request permission for [code]"RECORD_AUDIO"[/code] by [code]AudioDriverOpenSL[/code].
+				[b]Note:[/b] This method is currently only implemented on Android and is used by [code]AudioDriverOboe[/code].
```
kus04e4ek, line 656: I think that the specific permissions and what they are for should be put somewhere else? It would feel better if they were in AudioServer docs somewhere, where the user would be expected to look first for any audio specific questions and issues. This PR also adds one more permission "MODIFY_AUDIO_SETTINGS" for SCO support on older devices, but I'm out of ideas where to put it.

platform/android/java/lib/src/org/godotengine/godot/utils/PermissionsUtil.java
```diff
@@ -63,6 +63,7 @@ public final class PermissionsUtil {
 	public static final int REQUEST_RECORD_AUDIO_PERMISSION = 1;
 	public static final int REQUEST_CAMERA_PERMISSION = 2;
 	public static final int REQUEST_VIBRATE_PERMISSION = 3;
+	public static final int REQUEST_MODIFY_AUDIO_SETTINGS_PERMISSION = 4;
```
kus04e4ek, line 66: I think that it's unused, but added for consistency

servers/audio_server.h
```diff
@@ -62,6 +62,7 @@ class AudioDriver {
 	unsigned int input_size = 0;
 
 	void audio_server_process(int p_frames, int32_t *p_buffer, bool p_update_mix_time = true);
+	void audio_server_init_channels_and_buffers();
```
kus04e4ek, line 65: Exposed it, since Oboe doesn't recommend to allocate memory in callbacks, which would happen through audio_server_process if device changed

doc/classes/AudioServer.xml
Mickeon: Thumbs up, I don't think the parentheses are needed though

kus04e4ek: Put parentheses just to match Godot's style? However, when I look at all the times "in seconds" is used it doesn't seem to be very consistent, sometimes it's written in parentheses, with comma or without anything, e.g. `Should return the length of this audio stream, in seconds` and `Returns the length of the audio stream in seconds` in AudioStream's doc. Will push just `Returns the audio driver's effective output latency in seconds`

thirdparty/README.md
```diff
@@ -724,6 +724,23 @@ with the provided patch.
   unnecessary code.
 
 
+## oboe
+
+- Upstream: https://github.com/google/oboe
+- Version: 1.8.2 (16d72c89be9eb8a7d617a7be531a31dba3db74f1, 2024)
```
kus04e4ek, line 730: Put version 1.8.2, although it's not released yet, just to have this fix: [google/oboe#1968](https://github.com/google/oboe/pull/1968). Without this fix, on some devices, the audio would be silent or with a lot of glitches since the code checking for it was never called

loliaudio:
> **EDIT**: If you're interested in testing types, then all of them can be found here: https://developer.android.com/reference/android/media/AudioDeviceInfo#TYPE_AUX_LINE Also, when testing SCO your device might require "MODIFY_AUDIO_SETTINGS" permission, this can be given at the export preset.

I would like to help testing, but I have no knowledge abou C++.
I have some of the devices listed.
Is there a way of doing it?

doc/classes/AudioServer.xml
Mickeon: There's no consistency to it. It depends on the rest of the sentence. I think in this case there's so few things to read, there's no need for parentheses. "_Returns the audio driver's effective output latency in seconds_" sounds good to me.

kus04e4ek:
> I would like to help testing, but I have no knowledge abou C++.
> I have some of the devices listed.
> Is there a way of doing it?

Thanks for replying!
First of all you need to install [git](https://git-scm.com) and [SCons](https://pypi.org/project/SCons).
Then follow the instructions [here](https://docs.godotengine.org/en/latest/tutorials/export/exporting_for_android.html) (everything before "Providing launcher icons" is required) and [here](https://docs.godotengine.org/en/latest/contributing/development/compiling/compiling_for_android.html).
Clone this PR's github repository: `git clone https://github.com/kus04e4ek/godot.git`.
I build with these options: `scons tests=false deprecated=no minizip=no vulkan=no modules_enabled_by_default=no module_text_server_fb_enabled=yes module_gdscript_enabled=yes module_freetype_enabled=yes module_minimp3_enabled=yes platform=android generate_apk=yes target=template_debug arch=<insert your architecture>` (to get your architecture you can use AIDA64, also godot uses arm32 instead of armeabi-v7a and arm64 instead of arm64-v8a)
[The latest dev build](https://godotengine.org/article/dev-snapshot-godot-4-3-dev-4) is compatible with the export template, so you can just use it as an editor.
Setup the export preset for android by overriding `Custom Template/Debug`, choosing the correct architecture and turning on `Modify Audio Settings` and `Record Audio` permissions.

kus04e4ek:
> Right now, when setting the device on one of the streams, the device on the other would be set too. I did it that way, because turning on SCO for one stream would turn on SCO for the other. Also, when turning on SCO, the default device used by Android changes, so did it for the consistency on what the default device is.

I misinterpreted Android docs, now output and input streams do not depend on each other. This introduced more bugs when setting output device, while input stream uses SCO. Have no ideas what the issue is, but will work on it. Default device still changes, but I don't think that's a big issue.

> Setting SCO just doesn't work seemingly at random times?

Fixed it, but input stream that uses SCO is **very** unstable due to #86428.

doc/classes/AudioServer.xml
kus04e4ek: Pushed the change, so will resolve this conversation

bestvcboy:
May I ask which version this is likely to be supported on?

kus04e4ek:
> May I ask which version this is likely to be supported on?

Can't really guarantee anything here, I'm not part of Godot's team, I just work on it when I can (right now I don't have time). There are still some issues and I want to refactor the code (I'll mark it as a draft, kinda forgot to do it before😅). I am also more focused on #90013 for now, which would simplify the code here and I think it should be merged first. Not many users request Oboe and there's a lot of code to go through when reviewing, so this PR is probably not a priority for members. As far as I'm aware there's no consensus on using Oboe in the first place (I plan to write a comment on Oboe's proposal page that would go through what are the advantages and disadvantages of using Oboe) (should probably mention that no one on the team seems to be against using Oboe though). I think there's currently only one member of the audio team and that's reduz, who seems to be very busy at the moment, so the reviewing process for this PR may take a long time. Because of all of this, I don't think it should be expected to be part of Godot any time soon if it will be.

kus04e4ek:
Oh, I renamed the source branch and this closed the PR sadly, I'll open a new one soon as I don't think GitHub allows to change the source branch

**EDIT:** Oh, sorry, it looks like I deleted Godot's master branch as well 😆
image

PR #86777, opened 2024-01-04
Add autocompletion for ClassDB & AudioServer

Mickeon:
Related to https://github.com/godotengine/godot/pull/86753, https://github.com/godotengine/godot/pull/86747, https://github.com/godotengine/godot/pull/86758, and https://github.com/godotengine/godot/pull/86764 

This PR adds autocompletion for almost all of the methods in **ClassDB**.

image

This is probably the biggest chain of methods to check for options. These can really come in handy preventing misspells while developing addons and the likes of them, although admittedly it's fairly _niche_.

At first, I was afraid I would have to cache the results for how many there are, but I'm surprised there was no particularly stutter.

... Because it was so small, I also added autocompletion for **AudioServer**'s `get_bus_index()` and `set_bus_send()`, which is just plain nice, I assure you.

### DIFF
--- a/core/core_bind.cpp
+++ b/core/core_bind.cpp
@@ -1512,6 +1512,30 @@ bool ClassDB::is_class_enabled(StringName p_class) const {
 	return ::ClassDB::is_class_enabled(p_class);
 }
 
+#ifdef TOOLS_ENABLED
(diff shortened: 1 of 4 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #64903, opened 2022-08-25
Fix external Material for invalid Material

antonWetzel:
## Closes https://github.com/godotengine/godot/issues/64853

"Use External" works for Imported Assets (Blender, GLTF) even without a Material to replace.
The Implementation requires the importer to assign a unique "filler_id" as meta data to a Material if the Material does not have a valid import id or name.

I could not find a solution without edits in Importers because the Materials have no identification to find the right external Material to replace.

_Bugsquad edit: fixes: https://github.com/godotengine/godot/issues/64970_

PR #86779, opened 2024-01-04
Implement Resource State Inheritance

### LINKED ISSUE #4089: Resource Inheritance (Prototypal Inheritance)

Examining Godot as a potential engine for a project.

I have a number of skills, abilities, and would like to base one resource on another. As it stands common fields must be re-entered in each similar resource.

This is often referred to as prototypical inheritance.

One Godot developer mentioned this feature on our forum.
https://discord.com/channels/212250894228652034/610347507222052867/944622534819733545

Highly sought-after feature in Unity and would give Godot a competitive advantage.
https://forum.unity.com/threads/suggestion-prefab-variants-for-any-object-scriptableobject-variants.585268/
https://github.com/GieziJo/ScriptableObjectVariant

We could keep a list of dirty property names (strings), a reference from instance to the parent, and a reference from parent to instance. In the editor when the parent is modified, the property could be propagated to the children, this would prevent having any cost at runtime

No.

This is because this feature could be used in more than custom resource types. For example Material Inheritances.

### LINKED ISSUE #8094: Allow defining inheriting Styleboxes using a parent Stylebox with overrides

Editor app with many buttons and panels

Configuring controls and themes often require setting StyleBoxes for multiple states of the same control. For example, normal, hover, active, disabled. In most cases these styleboxes will be identical except for one or two parameters, like a different background or border. 

But currently all properties need to be set for all styleboxes, which leads to a lot of data duplication and places where you need to change the same information if you want to change it. For example if you have your four Button state styleboxes set up and you decide to change the corner radius or border width, you'll need to set four values on four Stylebox resources. So 16 edits across four resources to change what I'd consider one property of the style. 

Add a simple inheritance mechanism in StyleBox resources where you can reference a parent StyleBox of the same type (StyleBoxFlat etc). This will then provide a default value for all parameters, and you can set only the parameters you want to override and be different in this StyleBox. Similar to how cascading style classes in CSS and other design style systems work.

With this in place I could just set a base ButtonStyleBox with all properties set, then add a HoverButtonStyleBox that references ButtonStyleBox and just overrides BGColor for example. If I then change border or corner properties in ButtonStyleBox this will automatically be inherited by HoverButtonStyleBox.

Add a new parentStylebox property on base StyleBox that can be used in all StyleBoxes (maybe not StyleBoxEmpty).
If parentStylebox is set, this resource will be used for default values for each property where the Resource doesn't have an override value set.

Currently it looks like the StyleBox editor interfaces already considers set parameters that are different from the base defaults overrides, that can be reset to the default. This means the interface, and possibly override flags for all properties are already in place at least in the editor. It's just the currently the stylebox doesn't inherit and override another stylebox, just the default values.

You could I guess create your own stylebox override code that generates copies of a base stylebox resources with some overrides. But it will be inconvenient and har to make as useful and convenient as integrating it in the base styleboxes.

StyleBoxes are part of the core UI components

reduz:
* Ability to inherit a resource and do local changes to it.
* Ability to customize in each resource how inheritance happens.

An inherited resource has a reference to the original one and shows the values of the original one. It can override these values and only those overrides are saved, otherwise the values in the original resource are used. This allows to have a base resource with a lot of parameters customized, and then inherited ones that may tweak one thing or two.

Ideally, this feature will also allow to tweak animations and animationlibraries from imported scenes (without having to do the hack of saving them to disk), although this needs other code changes that will happen in separate PRs.

How it looks:

image

This PR is intended to play together with [this one](https://github.com/godotengine/godot/pull/84168) and [this proposal](https://github.com/godotengine/godot-proposals/issues/8750), to allow much better editing and usage of imported assets.

- *Production edit: This closes https://github.com/godotengine/godot-proposals/issues/4089
and closes https://github.com/godotengine/godot-proposals/issues/8094.*


Mickeon:
Is a inherited Resource still its own thing at run-time?

That is, at run-time, modifying any property of a Resource does not affect any inherited Resource, is that correct?

reduz:
@Mickeon It is its own thing. The only thing I did not implement in this PR is, if while during the editor, changing the base resource should change the inherited ones (basically something editor-only to check if the property edited in the base one is the same value as in the inherited ones, then change the inherited ones). This should not be hard to do.

During run-time, however, the inherited ones are its own thing and only fetch the non-modified properties from the base one.

Mickeon:
That is excellent, and simple to describe, too. Users that want properties to propagate at run-time can listen for the `changed` signal of the inherited **Resource**, and/or override `_setup_inherited`. *As an exposed method, it's marginally more useful than _setup_local_to_scene* 💦.

Not sure about the current names, but this overall looks like a surprisingly simple yet powerful feature.

Ansraer:
It would be nice to have some kind of logo in the corner of inherited resources to let the user know that it depends on something else. And maybe we could highlight the parent resource if the "sub-resource-icon" is pressed?
Otherwise this could get messy quickly, especially if you are collaborating with other people and have no clue that some resources are sub resources. Trying to figure out why some of your resource's values change seemingly spontanously sounds like a nightmare to me.

Oh, it would also be nice if the inspector would some show (e.g. with a different background color) which fields the sub resource overwrites. And a way to un_overwrite a field would probably also be necessary.

reduz:
@Ansraer yeah I could put the instance logo in the inspector. Regarding the editor, anything that has the revert button is what it has been overriden.

TokisanGames:
Does this store only changes from the original or is it a unique instance that is easily revertable to the original?

The context I'm thinking about is a vertex painting tool, which is not practical to use in Godot due to the current design.

As we've discussed at length, the only safe option is to keep meshes tied to GLBs so they can be reimported on demand. But the meshes are locked, so we can't paint vertices. In order to do so, we need unique resources, but then we have disassociated ArrayMeshes and when the format inevitably changes, we have a problem. We can reimport the GLBs, but Godot will complain about the old format on the disassociated ArrayMeshes in all mesh instances we have painted.

Does this PR help solve this dilemma for vertex painting?

Calinou:
> Does this PR help solve this dilemma for vertex painting?

A different solution is needed, similar to the existing UV2 caching system used for lightmap unwraps: https://github.com/godotengine/godot-proposals/issues/6486

reduz:
Vertex painting is far more involved and I have the feeling it should need a separate implementation (not even using vertex arrays). I will open a proposal soon about this.

reduz:
This is ready for review.

YuriSizov:
So there is a limitation in the current implementation that only the same type can be inherited by a resource. But I think we should allow narrowing types. I can imagine a situation where I have a base StyleBox with some content margins configured, and then have a StyleBoxFlat and a StyleBoxEmpty inheriting it, but adding their own unique properties as well. In fact, this is definitely a use case we want to be able to handle in various multistate controls, like buttons.

reduz:
@YuriSizov The limitation is because by default it may lead to mistakes, but it definitely could be more lean (as in, if the resource class also inherits the base class its should be fine).

I can modify it for this as you suggest.

YuriSizov:
Yes, please do, I think it would be useful and expected.

reduz:
I applied the suggested changes. Also renamed the feature to "state inheritance" which makes more sense and is less confusing vs regular inheritance.

fire:
@reduz Were we expecting to queue this for merging in 4.3? You mentioned letting this bake in the last Pipeline and Animations meeting.

theraot (commented):
1. This works as expected for materials.

2. The usability of this PR is poor. It would be better to have a way to create an inherited resource from the filesystem, similar to creating an inherited scene.

3. The option to set `resource_inherits_state` is available for `PacketScene` (opening the `PacketScene` in the inspector, not editing the scene), and it does not work. I believe it should not be visible to the inspector.

4. I tested this in the context this proposal of mine: https://github.com/godotengine/godot-proposals/issues/9802 and found it lacking for animations, but I believe that could be addressed separately.

DaloLorn:
Point 2 doesn't seem like a dealbreaker - better a bad UX than no UX for this sort of thing - but on the other hand, now that it's missed the deadline for inclusion in 4.3, it does sound like something that should be addressed in time for 4.4.

theraot:
None of what I said is a deal breaker. I didn't say it was.

Mickeon (commented):
It is my realization ONLY NOW that the below comments have always been pending and Github never reminded me of this.

Calinou:
@CsloudX Please don't bump issues without contributing significant new information. Use the :+1: reaction button on the first post instead.

core/io/resource.cpp
```diff
@@ -437,6 +437,98 @@ String Resource::get_id_for_path(const String &p_path) const {
...
+		return ret;
+	}
+
+	bool default_value_valid;
+	Variant default_value = inherits_state->get(p_name, &default_value_valid);
+	if (!default_value_valid) {
+		return true; // Not a default value, so save.
+	}
+
+	if (default_value.get_type() != Variant::NIL && bool(Variant::evaluate(Variant::OP_EQUAL, p_value, default_value))) {
```
Macksaur, line 511: What is the reason for skipping `bool Variant::operator==(const Variant&)` here? I don't understand why it's not written as `p_value == default_value`. Can you elaborate?

fstxz:
Reverting property value doesn't work properly. Let's say you have a script item.gd:

```gdscript
class_name Item
extends Resource

@export var cost: int
```

Then you create a resource apple.tres, set cost to 10.
Then you create a new resource that inherits apple.tres, keep the cost at 10 and save it as new_apple.tres

Now if you look at the new_apple.tres's properties, the cost can be reverted, even though it's unchanged from the base apple.tres. Clicking revert button reverts the property's value to the script's default value, in this case to 0.

The cost is _not_ saved in the file itself if the value is 10, so this part works correctly, it's just the UI that is acting weirdly, the revert button is always visible and can be clicked multiple times.

If you go back to apple.tres and change the cost to 20 and save, then go back to new_apple.tres and revert the cost, it will change from 10 to 20, which is good, however the revert button is still visible, clicking it again will change the cost from 20 to 0, clicking again changes the cost from 0 to 20.

OhiraKyou:
Would this also work for meshes extracted from GLTF files on reimport?

GPU particle draw passes use the materials assigned to the mesh data. This is a problem, because that data is shared. So, you can't use the same mesh for multiple GPU passes. Note that, when more than one draw pass is used, the geometry instance material override is not a valid solution, because it doesn't support assigning a separate override per draw pass.

Furthermore, any materials assigned to a draw pass are lost on reimport due to the data being stored in the mesh asset that gets overwritten on extraction. And, even if one is willing to accept that material data is shared between all passes that use the same mesh, it's not even possible to assign that one set of persistent materials when exporting placeholder materials from Blender (see [PR 64903](https://github.com/godotengine/godot/pull/64903)).

Resource inheritance has the potential to offer workarounds for these problems by enabling mesh resources to be inherited with material overrides that, ideally, should survive model reimports and consequent mesh data overwrites.

Although, I would prefer if GPU particle draw passes had individual material override options in the first place. Really, draw passes should have been implemented as a new resource type. And, if a new resource type were added for them, they would need to be automatically converted in old projects.

Zephilinox:
what's needed to get this feature merged? I have a custom import workflow for 3D assets to get around some issues with the default one, but there's a particular pain point with importing the updated materials while keeping any changes made on the godot side, which this PR would solve

MajorMcDoom:
I made a little "Inherited Resources" addon as a workaround in the meantime, if anyone is waiting for this feature:
https://codeberg.org/MajorMcDoom/cozy-cube-godot-addons/

https://github.com/user-attachments/assets/b27573fd-39dd-453d-a6ad-cfdc4b646b97

fire:
I might look into salvaging this for the 4.7+ cycle.

core/io/resource.cpp
LucaVazz: I also wondered this so had a look trough the codebase... It seems to be a re-occuring pattern.

The basis lays in the declaration of `OP_EQUAL` regarding nil values, see [Line 533 to 607 in variant_op.cpp](https://github.com/godotengine/godot/blob/master/core/variant/variant_op.cpp#L533-L607) and [`OperatorEvaluatorEqualObjectNil` and `OperatorEvaluatorEqualNilObject` in variant_op.h](https://github.com/godotengine/godot/blob/master/core/variant/variant_op.h#L484-L506).\
Both use [`get_validated_object` in variant.cpp](https://github.com/godotengine/godot/blob/master/core/variant/variant.cpp#L2043-L2049) which returns `nullptr` if the object is not inside the ObjectDB, e.g. if it is an empty object.

The direct use of `==` would not cover this.

See also [Line 688 to 689 in script_language.cpp](https://github.com/godotengine/godot/blob/master/core/object/script_language.cpp#L688-L689):
> The evaluate function ensures that a NIL variant is equal to e.g. an empty Resource.
> Simply doing defval == p_value does not do this.

LucaVazz:
@fire Did you encounter any blocks when trying to salvage this PR?

If there is no blocking problem with it, I would be keen to try and help with salvaging this for 4.8

fire:
I have needed to do paid work. Feel free to salvage.

Orso2p2n:
Hi! Is anyone currently salvaging this?

LucaVazz:
Hey :)

I'm currently working on rebasing the changes and addressing the review comments.
However, as it's my first time properly working with the Godot source code, it's taking some time to understand everything.

KoBeWi:
This is not a good material for first contribution.

LucaVazz:
@KoBeWi I'm not against anyone else pitching in to help with getting this feature rebased and merged instead - I'd love to see it integrated in the best way possible

PR #75103, opened 2023-03-19
Enable dependency fixer for missing InstancePlaceholder paths

Flavelius:
Fixes https://github.com/godotengine/godot/issues/37818. Fixes https://github.com/godotengine/godot/issues/44720.

This treats placeholder paths as a dependency to invoke the regular ext_resource dependency fixer.
This also fixes file-reimport not retaining existing uids for resources that were renamed, and resource dependency renaming also not transitioning existing uids.

I chose to string-replace occurrences, as it was much more straightforward than parsing and rewriting everything and to not duplicate a lot of code in the process.

PR #86781, opened 2024-01-04
Allow opening scenes with missing scene dependency

KoBeWi:
When a scene has instances of scenes that no longer exist in the project, you won't be able to open them, with the well-known "Scene invalid/corrupt" error. This PR fixes that.

https://github.com/godotengine/godot/assets/2223172/492e087a-10b8-4002-8ca9-1e4a05dcd105

However, while it *technically* works (i.e. you can open the scene), it could use some improvements. I wanted it to at least display path of the missing scene, but I wasn't sure how to obtain it (in the video it shows as TODO). Also, contrary to what the tooltip says, saving the scene **will** result in data loss. The instance will be gone and any editable children data is lost. It would be nice to preserve at least the instance; if it's not possible/too difficult, changing the tooltip would be fine for now too.

### DIFF
--- a/doc/classes/MissingNode.xml
+++ b/doc/classes/MissingNode.xml
@@ -13,8 +13,11 @@
 		<member name="original_class" type="String" setter="set_original_class" getter="get_original_class">
 			The name of the class this node was supposed to be (see [method Object.get_class]).
 		</member>
+		<member name="original_scene" type="String" setter="set_original_scene" getter="get_original_scene">
+			Returns the path of the scene this node was instance of originally.
+		</member>
 		<member name="recording_properties" type="bool" setter="set_recording_properties" getter="is_recording_properties">
-			If set to [code]true[/code], allows new properties to be added on top of the existing ones with [method Object.set].
+			If [code]true[/code], allows new properties to be set along with existing ones. If [code]false[/code], only existing properties' values can be set, and new properties cannot be added.
 		</member>
 	</members>
 </class>
--- a/scene/main/missing_node.cpp
+++ b/scene/main/missing_node.cpp
@@ -66,6 +66,14 @@ String MissingNode::get_original_class() const {
 	return original_class;
 }
 
+void MissingNode::set_original_scene(const String &p_scene) {
+	original_scene = p_scene;
+}
+
+String MissingNode::get_original_scene() const {
+	return original_scene;
+}
+
 void MissingNode::set_recording_properties(bool p_enable) {
 	recording_properties = p_enable;
 }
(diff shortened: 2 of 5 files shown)


Zireael07:
@Mickeon Something for a followup PR I think?

Mickeon:
Yeah. For the focus of this PR, changing the tooltip to not blatantly lie may be good enough.

Mickeon:
Perhaps MissingNode could have a Dictionary containing NodePaths pointing to the lost properties and values. When loading a PackedScene, it is detected and rebuilt if possible, or left as is. It sounds extremely hacky but... Yeah.

KoBeWi:
I managed to make the path available using a questionable hack 😬 Though for some reason it's not always reliable (it seems to depend on some cache).
I also adjusted the message:
image
For now the instance is completely discarded.

In total there are 3 paths where loading scene can fail: instances, placeholders and inheritance. The former 2 are handled by this PR, the last one I think is not recoverable.

reduz (approved):

akien-mga:
Will need a rebase.

KoBeWi:
Rebased. Seems like someone added a description for `recording_properties` in the meantime. I replaced it with mine, as it provides more information.

reduz:
give me a second before merging this, I want to check something

reduz:
Alright, I think its good to merge, but I think what we still need to have an option to recursively fix dependencies on subresources that don't load because of this.

doc/classes/MissingNode.xml
```diff
@@ -13,8 +13,11 @@
 		<member name="original_class" type="String" setter="set_original_class" getter="get_original_class">
 			The name of the class this node was supposed to be (see [method Object.get_class]).
 		</member>
+		<member name="original_scene" type="String" setter="set_original_scene" getter="get_original_scene">
+			Returns the path of the scene this node was instance of originally.
+		</member>
 		<member name="recording_properties" type="bool" setter="set_recording_properties" getter="is_recording_properties">
-			If set to [code]true[/code], allows new properties to be added on top of the existing ones with [method Object.set].
+			If [code]true[/code], setting a non-existent property will store its value in the node's internal HashMap. If [code]false[/code], only existing properties' values will be accepted.
```
Mickeon, line 20: Hii, it was me that wrote the prior description. There's no point mentioning the internal HashMap here. It's not something the user can access directly. I would personally revert the prior description a bit (but keep the mention of what happens if `false`).
```suggestion
			If [code]true[/code], allows new properties to be set along with existing ones. If [code]false[/code], only existing properties' values can be set, and new properties cannot be added.
```

Bike-shedding, I know.

akien-mga:
Thanks!

Flavelius:
I tried to address this a while ago aswell, and while it was more of a hackfix (https://github.com/godotengine/godot/pull/75103), i think not losing nodes is still the actual desire, as some sources might just have been renamed or moved; where it shouldn't be assumed they are actually missing.

scene/main/missing_node.cpp
```diff
@@ -77,20 +85,31 @@ bool MissingNode::is_recording_properties() const {
...
+	}
 	return ret;
 }
 
 void MissingNode::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_original_class", "name"), &MissingNode::set_original_class);
 	ClassDB::bind_method(D_METHOD("get_original_class"), &MissingNode::get_original_class);
 
+	ClassDB::bind_method(D_METHOD("set_original_scene", "name"), &MissingNode::set_original_class);
+	ClassDB::bind_method(D_METHOD("get_original_scene"), &MissingNode::get_original_class);
```
raulsntos, line 105: Looks like these are bound to the wrong methods. Opened https://github.com/godotengine/godot/pull/108531.

PR #61227, opened 2022-05-20
[3.x] Fix signed distance field font rendering

Calinou:
This fix works in both GLES3 and GLES2. Thanks @filipworksdev for the fix :slightly_smiling_face: 

The rendering formula in the shader was adjusted to further improve the sharpness/antialiasing quality balance.

Improvements to rendering quality could be made by adjusting SDF smoothing depending on font size (relative to screen size) or performing supersampling. However, the `screen_pixel_size` uniform is only available when `SCREEN_UV` is used in the shader, and I need that to adjust SDF smoothing based on resolution.

This closes https://github.com/godotengine/godot/issues/8022.

## Preview

*The SDF texture is about 40 pixels tall.*

## TODO

- [x] Fix SDF font rendering when batching is enabled and non-SDF nodes are rendered. This bug affects both GLES3 and GLES2.

PR #86786, opened 2024-01-04
[3.x] Fix signed distance field font rendering

### LINKED ISSUE #8022: Distance field fonts are blurry
Tested on Godot 2.1.2
Windows 10 64 bits

I created a theme with a custom font, and it looks fine with Labels. However, every other control gets blurry text instead. I expected the same result as the label, because I see the text appears to be the same size.

image

[ItemListFontBlurry.zip](https://github.com/godotengine/godot/files/839877/ItemListFontBlurry.zip)

Related QA thread I investigated from: https://godotengine.org/qa/13127/itemlist-font-is-scaled-and-blurry

lawnjelly:
This fix works in both GLES3 and GLES2.

The rendering formula in the shader was adjusted to further improve the sharpness/antialiasing quality balance.

Fixes #8022 .
Supersedes #61227 .

Refer to original PR for full info.

## Notes
* I can't _completely_ vouch for this as it is @Calinou 's work, I've just added some rudimentary state tracking to prevent calling `VisualServer` to set SDF state for every draw.
* I've opted to just store a bool per `canvas_item` for SDF state and wrap the changes with a `select_font()` function. I did look at putting it into `get_font()` but `get_font()` is also used when not drawing.
* There's the possibility this whole SDF on / off per canvas item could get out of sync if the user decides to draw fonts / mix fonts, and cause some visual anomalies, but it will hopefully be rare (and only affect users who are concerned with SDF fonts).
* Of course better suggestions for managing the state without compromising performance are always welcome, there's quite a few ways of managing this.
* I haven't spent a long time on this evaluating etc, it is probably pretty benign and should do the job. Can always be improved later if necessary (bear in mind how many are still using 3.x).

### DIFF
--- a/drivers/gles3/shaders/canvas.glsl
+++ b/drivers/gles3/shaders/canvas.glsl
@@ -581,18 +581,17 @@ void main() {
 
 #endif
 
-#if !defined(COLOR_USED)
-	//default behavior, texture by color
-
 #ifdef USE_DISTANCE_FIELD
-	const float smoothing = 1.0 / 32.0;
-	float distance = textureLod(color_texture, uv, 0.0).a;
-	color.a = smoothstep(0.5 - smoothing, 0.5 + smoothing, distance) * color.a;
+	// Higher is smoother, but also more blurry. Lower is crisper, but also more aliased.
+	// TODO: Adjust automatically based on screen resolution/font size ratio.
+	const float smoothing = 0.125;
+	float dist = texture(color_texture, uv, 0.0).a;
+	color.a = smoothstep(0.5 - smoothing, 0.5 + smoothing, dist);
 #else
+#if !defined(COLOR_USED)
+	// Default behavior, texture by color.
 	color *= texture(color_texture, uv);
-
 #endif
-
 #endif
 
 	vec3 normal;
--- a/drivers/gles2/rasterizer_canvas_base_gles2.cpp
+++ b/drivers/gles2/rasterizer_canvas_base_gles2.cpp
@@ -58,6 +58,7 @@ void RasterizerCanvasBaseGLES2::canvas_begin() {
 	state.using_large_vertex = false;
 	state.using_modulate = false;
 
+	state.canvas_shader.set_conditional(CanvasShaderGLES2::USE_DISTANCE_FIELD, false);
 	state.canvas_shader.set_conditional(CanvasShaderGLES2::USE_ATTRIB_LIGHT_ANGLE, false);
 	state.canvas_shader.set_conditional(CanvasShaderGLES2::USE_ATTRIB_MODULATE, false);
 	state.canvas_shader.set_conditional(CanvasShaderGLES2::USE_ATTRIB_LARGE_VERTEX, false);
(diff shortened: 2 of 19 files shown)


drivers/gles3/shaders/canvas.glsl
```diff
@@ -583,16 +583,15 @@ void main() {
...
-
 #endif
 
+#ifdef USE_DISTANCE_FIELD
+	// Higher is smoother, but also more blurry. Lower is crisper, but also more aliased.
+	// TODO: Adjust automatically based on screen resolution/font size ratio.
+	const float smoothing = 0.125;
+	float dist = texture(color_texture, uv, 0.0).a;
+	color.a = smoothstep(0.5 - smoothing, 0.5 + smoothing, dist);
 #endif
```
clayjohn, line 595: I'm a bit skeptical of the shader change here.  This logic results in 2 changes of behaviour that seem unintended:
1. If `COLOR` is used, then the font texture is still sampled which changes behavior (although seems like an improvement to me)
2. The RGB color from the distance field texture is used now even when using distance field rendering. It seems like this works because the texture is always white, but its probably best to just not have the texture sample at all


In 4.x the MSDF rendering is also an if/else combined with the texture sample.

lawnjelly: @Calinou as this is his code. I'll try and take a look too.

lawnjelly: I must admit I'm slightly lost and not so familiar with this part of the shader, and defer to you and Calinou (and the situation where COLOR is in use etc) but would this be any better? (And presumably the same for GLES2)

I was interpreting your post above as "when `color_texture` is being used for SDF, only the alpha is used, and the RGB is all white, so no need to read them." Whether this makes a difference in practice I have no idea :grin:  (as the read is probably always RGBA).

```
#ifdef USE_DISTANCE_FIELD
	// Higher is smoother, but also more blurry. Lower is crisper, but also more aliased.
	// TODO: Adjust automatically based on screen resolution/font size ratio.
	const float smoothing = 0.125;
	float dist = texture(color_texture, uv, 0.0).a;
	color.a = smoothstep(0.5 - smoothing, 0.5 + smoothing, dist);
#else
#if !defined(COLOR_USED)
	//default behavior, texture by color
	color *= texture(color_texture, uv);
#endif
#endif
```
Here the SDF would be applied to color even if `COLOR_USED`, and the RGB is not applied when using distance field.

clayjohn: That makes more sense to me and is much closer to what we do in 4.x

lawnjelly:
Fixed up the shader to the amended version discussed in the comments.

Calinou (approved):
Tested locally, it works as expected (Linux, GeForce RTX 4090 with NVIDIA 545.29.06).

**Testing project:** https://github.com/godotengine/godot-demo-projects/tree/3.x/gui/sdf_font

## Before

### GLES3

https://github.com/godotengine/godot/assets/180032/f3dd7218-3273-4824-8dcf-a74e7143b069

### GLES2

https://github.com/godotengine/godot/assets/180032/482834a8-8f9e-45fb-88cb-7edb313d54b4

## After *(this PR)*

### GLES3

https://github.com/godotengine/godot/assets/180032/1587e3b3-2e89-49ed-aeeb-f22f38f649ae

### GLES2

https://github.com/godotengine/godot/assets/180032/93c2c7d8-fded-4cb0-ab62-570208fa915d

akien-mga:
Thanks!

PR #86787, opened 2024-01-04
Fix various spelling errors across engine solution 

ryevdokimov:
Found a couple spelling errors while skimming through code, then decided to utilize Visual Studio's built in spell checker across the solution since this is a semi new feature and I figured someone would eventually do this anyways. Fixed some obvious spelling errors and ignored minor grammatical things to avoid being pedantic.

### DIFF
--- a/modules/navigation/nav_agent.cpp
+++ b/modules/navigation/nav_agent.cpp
@@ -326,7 +326,7 @@ const Dictionary NavAgent::get_avoidance_data() const {
 		_avoidance_data["new_velocity"] = Vector3(rvo_agent_3d.newVelocity_.x(), rvo_agent_3d.newVelocity_.y(), rvo_agent_3d.newVelocity_.z());
 		_avoidance_data["velocity"] = Vector3(rvo_agent_3d.velocity_.x(), rvo_agent_3d.velocity_.y(), rvo_agent_3d.velocity_.z());
 		_avoidance_data["position"] = Vector3(rvo_agent_3d.position_.x(), rvo_agent_3d.position_.y(), rvo_agent_3d.position_.z());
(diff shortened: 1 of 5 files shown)


modules/navigation/nav_agent.cpp
```diff
@@ -326,7 +326,7 @@ const Dictionary NavAgent::get_avoidance_data() const {
 		_avoidance_data["new_velocity"] = Vector3(rvo_agent_3d.newVelocity_.x(), rvo_agent_3d.newVelocity_.y(), rvo_agent_3d.newVelocity_.z());
 		_avoidance_data["velocity"] = Vector3(rvo_agent_3d.velocity_.x(), rvo_agent_3d.velocity_.y(), rvo_agent_3d.velocity_.z());
 		_avoidance_data["position"] = Vector3(rvo_agent_3d.position_.x(), rvo_agent_3d.position_.y(), rvo_agent_3d.position_.z());
-		_avoidance_data["prefered_velocity"] = Vector3(rvo_agent_3d.prefVelocity_.x(), rvo_agent_3d.prefVelocity_.y(), rvo_agent_3d.prefVelocity_.z());
+		_avoidance_data["preferred_velocity"] = Vector3(rvo_agent_3d.prefVelocity_.x(), rvo_agent_3d.prefVelocity_.y(), rvo_agent_3d.prefVelocity_.z());
```
akien-mga, line 329: CC @smix8 to make sure this doesn't break compat.

I only see this being written too and never read from though, which is weird.

akien-mga (approved):

modules/navigation/nav_agent.cpp
smix8: That is only debug print data not used for anything else so a rename is not a problem.

akien-mga:
Thanks!

Issue #84559, opened 2023-11-07
C# Android SSL Crash using System.Net.Http.HttpClient


jpetersen23:
### Godot version

4.2-beta4

### System information

Godot v4.2.beta4.mono - macOS 13.6.0 - Vulkan (Mobile) - integrated Apple M2 Max - Apple M2 Max (12 Threads)

### Issue description

Making a network request using System.Net.Http.HttpClient to "https://www.example.com" cause a crash, calling "http://www.example.com" does not crash. If I use Godot's node HttpRequest it does not crash.

The logcat output for the crash looks like:

11-06 17:44:57.813  3989  4048 I godot   : Pressed button SystemSsl
11-06 17:44:57.856  3989  4048 I godot   : send request to https://www.example.com
11-06 17:44:57.902  4405  4405 I SKBD    : and isTosAccept false
11-06 17:44:58.071  1061  1209 D DnsProxyListener: DNSDBG::dns addrinfo af 0
11-06 17:44:58.103  4405  4405 I SKBD    : and isTosAccept false
11-06 17:44:58.282  3989  4201 E godot   : USER ERROR: BUG: Unreferenced static string to 0: Physics2DConstraintSolveIslands
11-06 17:44:58.282  3989  4201 E godot   :    at: unref (core/string/string_name.cpp:127)
11-06 17:44:58.282  3989  4201 E godot   : USER ERROR: BUG: Unreferenced static string to 0: Physics2DConstraintSetup
11-06 17:44:58.282  3989  4201 E godot   :    at: unref (core/string/string_name.cpp:127)
--------- beginning of crash
11-06 17:44:58.284  3989  4201 F libc    : FORTIFY: pthread_mutex_lock called on a destroyed mutex (0x6f51b6400c)

### Steps to reproduce

This is the code that crashes in its simplest form:

	private void Send(string url)
	{
		var client = new System.Net.Http.HttpClient();
		var request = new HttpRequestMessage(HttpMethod.Get, url);
		GD.Print($"send request to {url}");
		var response = client.Send(request); // <-- crashes on this line.
		GD.Print(response.ToString()); // this does not get reached.
		var stream = response.Content.ReadAsStream();
		StreamReader reader = new StreamReader(stream, System.Text.Encoding.UTF8);
		var text = reader.ReadToEnd();
		GD.Print(text);
	}

	public override void _Process(double delta)
	{
		Send("https://www.example.com");
	}

### Minimal reproduction project

[...]

PR #86791, opened 2024-01-04
Change the value of DotNetOS.Android to "linux-bionic"

scgm0:
Currently DotNetOS.Android value is "android" and the platform RID is "android-arm64" when exported, but "android-arm64" does not support NativeAOT yet.
According to https://learn.microsoft.com/en-us/dotnet/core/rid-catalog, "linux-bionic-arm64" is a c library based on Android, it can be used for Android and supports NativeAOT.
I've only tested \<PublishAot\>true\<\/PublishAot\> and \<PublishAot\>false\<\/PublishAot\> under .Net8, and it's not clear if there are any other issues.
(There are a few other steps required to export Android using NativeAOT, but for now this is a necessary step)

### DIFF
--- a/modules/mono/editor/GodotTools/GodotTools/Utils/OS.cs
+++ b/modules/mono/editor/GodotTools/GodotTools/Utils/OS.cs
@@ -54,7 +54,7 @@ namespace GodotTools.Utils
             public const string OSX = "osx";
             public const string Linux = "linux";
             public const string Win10 = "win10";
-            public const string Android = "android";
+            public const string Android = "linux-bionic";
             public const string iOS = "ios";
             public const string iOSSimulator = "iossimulator";
             public const string Browser = "browser";


modules/mono/editor/GodotTools/GodotTools/Utils/OS.cs
```diff
@@ -39,7 +39,7 @@ public static class Platforms
             public const string Windows = "windows";
             public const string MacOS = "macos";
             public const string LinuxBSD = "linuxbsd";
-            public const string Android = "android";
+            public const string Android = "linux-bionic";
```
raulsntos, line 42: These are the Godot platform identifiers, not the .NET RIDs, those are in the `DotNetOS` static class which is just below this one.

raulsntos (changes requested):
We are already using the linux bionic runtime when exporting to Android which is why we only support arm64 and x64, and why some API is unavailable (see https://github.com/godotengine/godot/issues/84559).

Also, instead of changing the Android RID, I think it'd be better to add a new constant to `DotNetOS` for linux bionic if we need it. And decide which RID to use depending on which runtime the project needs.

modules/mono/editor/GodotTools/GodotTools/Utils/OS.cs
scgm0: > These are the Godot platform identifiers, not the .NET RIDs, those are in the `DotNetOS` static class which is just below this one.

I'm sorry

raulsntos:
Superseded by https://github.com/godotengine/godot/pull/97908. Thanks for the contribution nonetheless.

PR #86792, opened 2024-01-04
Fix Android dynamic library loading

m4gr3d:
Revert https://github.com/godotengine/godot/pull/86682 for the Android platform as it breaks dynamic library loading

### DIFF
--- a/platform/android/os_android.cpp
+++ b/platform/android/os_android.cpp
@@ -170,8 +170,6 @@ Error OS_Android::open_dynamic_library(const String p_path, void *&p_library_han
 		so_file_exists = false;
 	}
 
-	ERR_FAIL_COND_V(!FileAccess::exists(path), ERR_FILE_NOT_FOUND);
-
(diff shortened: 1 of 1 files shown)


dsnopek:
@m4gr3d What about something like PR https://github.com/godotengine/godot/pull/86794 which still gives a different error code if the file can't be found? (I'm not 100% sure it's correct, because I haven't had a chance to test it.)

m4gr3d:
cc @dsnopek @akien-mga We should include this fix in the dev2 build (or revert the original PR) to avoid introducing a regression on Android for dev2.

dsnopek (approved):
Per our discussion on PR https://github.com/godotengine/godot/pull/86794, there may not be a way to reliably distinguish between "not found" and "cannot load". So, let's just go with reverting for now.

But it would definitely provide a better developer experience for GDExtension if we were able to come up with a way to distinguish between those, at least in the most common cases.

akien-mga:
Thanks!

PR #86793, opened 2024-01-04
Add left/right shift/ctrl/alt/meta key enum to provides a way to distinguish between left and right shift/ctrl/alt/meta

### LINKED ISSUE #1387: Add a way to differentiate Left and Right Shift, Alt, and Control keys
A pinball-style game with left and right flippers.

I would like the option of using the left and right Shift keys as left and right paddles for users with tenkeyless or shorter keyboards, however, the Input Map treats these keys both as "Shift", and the `InputEventKey` `scancode` is `16777237` for both keys.

If there was a way to differentiate left and right Shift keys, it would be possible to use these keys for separate functions.

I can imagine two ways to solve this issue:
1. Add `KEY_RIGHT_SHIFT`, `KEY_RIGHT_ALT`, `KEY_RIGHT_CONTROL` to the KeyList enum.

```
func _input(ev):
  if ev is InputEventKey:
    match ev.scancode:
      KEY_SHIFT:
        print("left shift")
      KEY_RIGHT_SHIFT:
        print("right shift")
```

2. Add a `location` field to `InputEventKey` with a new enum for left/right. This would preserve the current functionality of `KEY_SHIFT`, etc., but probably would require more changes to get working with the Input Map menu.
```
func _input(ev):
  if ev is InputEventKey and ev.scancode == KEY_SHIFT:
    match ev.location:
      MULTIKEY_LOCATION_LEFT:
        print("left shift")
      MULTIKEY_LOCATION_RIGHT:
        print("right shift")
```

I'm not aware of any possible way to work around this issue in GDScript.

This feature would require adding to the GlobalScope KeyList enum or adding an enum to InputEventKey.

L2750558108:
The solution 1 of https://github.com/godotengine/godot-proposals/issues/1387
I think this is better than solution 2 because it doesn't add complexity to the code and it works without modifying the editor

Input.is_physical_key_pressed function now only allow the key enum of shift/ctrl/alt/meta with left/right ,or it will return false
Input.is_key_pressed function allow both shift/ctrl/alt/meta with left/right and shift/ctrl/alt/meta without left/right. In fact, the specific use of which key is determined by the virtual key provided by the operating system, in most cases without left/right (At least I didn't test out the case with left/right)

The value of InputEventKey.physical_keycode is the same as the allow range of Input.is_physical_key_pressed
The value of InputEventKey.keycode is the same as the allow range of Input.is_key_pressed

Now only avaliable on Windows, but could be ported to other platforms.

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/1387*


AThousandShips:
See also:
* https://github.com/godotengine/godot/pull/80231

For another PR solving this

core/os/keyboard.cpp
```diff
@@ -60,19 +60,29 @@ static const _KeyCodeText _keycodes[] = {
 	{Key::PAGEUP                ,"PageUp"},
 	{Key::PAGEDOWN              ,"PageDown"},
 	{Key::SHIFT                 ,"Shift"},
+	{Key::LEFT_SHIFT                 ,"LeftShift"},
```
AThousandShips, line 63:
```suggestion
	{Key::LEFT_SHIFT                 ,"LeftShift"},
```
Same for all

AThousandShips:
Your commit message has to be in English btw 🙂

L2750558108:
> Your commit message has to be in English btw 🙂

Sorry I forgot. I've changed it now

AThousandShips:
Later when finishing this up if it gets accepted I'd say to change the message to the title of the PR, which is more clear 🙂, but no need now

AThousandShips:
Superseded by:
* https://github.com/godotengine/godot/pull/80231

Thank you for your contribution nonetheless!

PR #86795, opened 2024-01-04
Fix negative light rendering in global illumination

### LINKED ISSUE #114819: Negative lights bake as if they were positive
### Tested versions

- Reproducible in 4.5.1.stable.mono

### System information

Godot v4.5.1.stable.mono - Nobara Linux 43 (KDE Plasma Desktop Edition) on Wayland - X11 display driver, Multi-window, 1 monitor - Vulkan (Forward+) - integrated AMD Radeon 780M Graphics (RADV PHOENIX) - AMD Ryzen 7 7840U w/ Radeon  780M Graphics (16 threads) - 30.65 GiB memory

### Issue description

the `Negative` flag is ignored by `LightmapGI` when baking lightmaps. static negative lights are treated as if they were regular (non-negative) lights.

Image
left-to-right, top-to-bottom: dynamic light, dynamic negative light, baked light, baked negative light.

### Steps to reproduce

- add a `LightmapGI`, a `MeshInstance3D`, and a light (any) to a 3D scene.
- set the light's `Negative` flag to `true`.
- set the light's `Bake Mode` to `Static`.
- unwrap the `MeshInstance3D`'s UV2 for lightmap/AO.
- bake lightmaps.

[baked_negative_lights.zip](https://github.com/user-attachments/files/24538292/baked_negative_lights.zip)

Calinou:
Negative lights no longer emit positive indirect light.

This affects VoxelGI, SDFGI and LightmapGI.

- This closes https://github.com/godotengine/godot/issues/114819, and closes https://github.com/godotengine/godot/pull/114933 by superseding it.

**Testing project:** [test_light_negative.zip](https://github.com/godotengine/godot/files/13834537/test_light_negative.zip)

## Preview

*All lights have **Indirect Energy** set to `10`, and the yellow light on the left has its **Negative** property enabled.*

### VoxelGI

Before | After *(this PR)*
-|-
Screenshot_20240109_184429 | Screenshot_20240109_185218

### LightmapGI

Before | After *(this PR)*
-|-
Screenshot_20240104_201430 | Screenshot_20240104_201118

### SDFGI

Before | After *(this PR)*
-|-
Screenshot_20240104_201513 | Screenshot_20240104_201353


Ansraer:
In the first comparison, why did the white light to the right side of the wall suddenly get so much brighter with the PR?

Calinou:
> In the first comparison, why did the white light to the right side of the wall suddenly get so much brighter with the PR?

This appears to be due to different properties being used between screenshots. I've retested the PR on latest `master` and have updated the VoxelGI screenshots accordingly :slightly_smiling_face:

Calinou:
Rebased and tested again, it works as expected on all renderers.

modules/lightmapper_rd/lightmapper_rd.h
```diff
@@ -299,9 +299,9 @@ class LightmapperRD : public Lightmapper {
...
-	virtual void add_omni_light(const String &p_name, bool p_static, const Vector3 &p_position, const Color &p_color, float p_energy, float p_indirect_energy, float p_range, float p_attenuation, float p_size, float p_shadow_blur) override;
-	virtual void add_spot_light(const String &p_name, bool p_static, const Vector3 &p_position, const Vector3 p_direction, const Color &p_color, float p_energy, float p_indirect_energy, float p_range, float p_attenuation, float p_spot_angle, float p_spot_attenuation, float p_size, float p_shadow_blur) override;
+	virtual void add_directional_light(const String &p_name, bool p_static, const Vector3 &p_direction, const Color &p_color, float p_energy, float p_indirect_energy, float p_angular_distance, float p_shadow_blur, bool p_negative = false) override;
+	virtual void add_omni_light(const String &p_name, bool p_static, const Vector3 &p_position, const Color &p_color, float p_energy, float p_indirect_energy, float p_range, float p_attenuation, float p_size, float p_shadow_blur, bool p_negative = false) override;
```
allenwp, line 312:
```suggestion
	virtual void add_omni_light(const String &p_name, bool p_static, const Vector3 &p_position, const Color &p_color, float p_energy, float p_indirect_energy, float p_range, float p_attenuation, float p_size, float p_shadow_blur, bool p_negative) override;
```

I'm not sure that default parameters are a good idea here. I don't see the issue with explicitly asking for whether a light is negative; by having no default it prompts the caller of the function to think about whether lights may be negative, which may prevent future bugs relating to this uncommon non-physical feature of negative lights.

allenwp (approved):
I see that `render_scene_cull.cpp` also has a number of these lines that don't apply the negative light sign:
``` c++
cache->energy = RSG::light_storage->light_get_param(instance->base, RS::LIGHT_PARAM_ENERGY);
```

Do these need to be considered as well? I've never looked at this cache before, so I'm not sure if it should be holding unmodified light energy that will later be modified by a negative light sign or if it should have the negative light sign baked into it when `light_get_param(instance->base, RS::LIGHT_PARAM_ENERGY)` is read.

One this and my other comment have been addressed, I'd say this PR looks good to me, so I'll give it an approval now. I have noticed that the changes introduced by this PR simply align these global illumination functions to have the same code style as the rest of the light code in Godot.

Calinou:
I've pushed an update to address both comments.

> I should also mention: It may be desirable to clip the energy of negative lights in global illumination. So this would mean that energy would be `max(0.0, energy_after_applying_sign)`. I'm not familiar enough with these GI algorithms, but the general rule in the industry is "avoid negative lights for any lighting simulation because they are non-physical and cause lighting simulations to break down"... So... I really don't know what the right answer is here.

This makes sense to me, but I'll wait for @clayjohn to weigh in just in case.

modules/lightmapper_rd/lightmapper_rd.h
allenwp: I've unresolved this comment because I see a thumbs up, which suggests agreement... but I also see the functions still have the default parameter? Just wanting to double check something didn't go funny with force pushing.

blueskythlikesclouds:
> It may be desirable to clip the energy of negative lights in global illumination

Would this not cause it to simply act as a light that does not exist? I think the point of negative lights is to kill the light energy like a black hole, isn't it?

allenwp:
> Would this not cause it to simply act as a light that does not exist? I think the point of negative lights is to kill the light energy like a black hole, isn't it?

Yeah, that's a good point. So long as the math works out well for luminance/radiance simulation in GI, I'd say it's a good approach to store negative; I just really don't know how the GI algorithms work, so I can't say what's best.

blueskythlikesclouds (commented):
This seems to work as I expect, but negative static lights do not show up at all in lightmaps. It appears to be due to this early quit: https://github.com/godotengine/godot/blob/master/modules/lightmapper_rd/lm_compute.glsl#L525

Changing it to `if (abs(attenuation * light_data.energy <= 0.0001))` appears to fix it but I'm not sure about the implications.

Calinou:
Following https://github.com/godotengine/godot-proposals/discussions/15413, I've rebased and tested again with all light types (including AreaLight3D), it works as expected. I've applied https://github.com/godotengine/godot/pull/86795#pullrequestreview-4599751330 and LightmapGI baking now takes negative light into account.

PR #86798, opened 2024-01-04
Add autocompletion for RenderingServer's global shader methods & `has_os_feature`

Mickeon:
Related to https://github.com/godotengine/godot/pull/86753, https://github.com/godotengine/godot/pull/86747, https://github.com/godotengine/godot/pull/86758,  https://github.com/godotengine/godot/pull/86764, and https://github.com/godotengine/godot/pull/86777

This PR adds autocompletion to all of **RenderingServer**'s methods that are about global shaders, allowing you to see the available ones right away, instead of having to check the Project Settings every time.

image

It also features... `has_os_feature()` because it's really nice to have a general clue as to what to put in there.

image

### DIFF
--- a/servers/rendering_server.cpp
+++ b/servers/rendering_server.cpp
@@ -2204,6 +2204,25 @@ void RenderingServer::fix_surface_compatibility(SurfaceData &p_surface, const St
 }
 #endif
 
+#ifdef TOOLS_ENABLED
+void RenderingServer::get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const {
+	String pf = p_function;
+	if (p_idx == 0) {
+		if (pf == "global_shader_parameter_set" || pf == "global_shader_parameter_set_override" ||
+				pf == "global_shader_parameter_get" || pf == "global_shader_parameter_get_type" || pf == "global_shader_parameter_remove") {
+			for (StringName E : global_shader_parameter_get_list()) {
+				r_options->push_back(E.operator String().quote());
+			}
+		} else if (pf == "has_os_feature") {
+			for (String E : { "\"rgtc\"", "\"s3tc\"", "\"bptc\"", "\"etc\"", "\"etc2\"", "\"astc\"" }) {
+				r_options->push_back(E);
+			}
+		}
+	}
+	Object::get_argument_options(p_function, p_idx, r_options);
+}
+#endif
+
 void RenderingServer::_bind_methods() {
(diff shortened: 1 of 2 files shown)


servers/rendering_server.cpp
```diff
@@ -2204,6 +2204,23 @@ void RenderingServer::fix_surface_compatibility(SurfaceData &p_surface, const St
 }
 #endif
 
+void RenderingServer::get_argument_options(const StringName &p_function, int p_idx, List<String> *r_options) const {
```
clayjohn, line 2207: Is there any danger to wrapping this in an EDITOR macro? In general data should never be read from the RenderingServer, but particularly, ``global_shader_parameter_get_list()`` should never be called outside the editor (it will fail if called from outside the editor).

Mickeon: There's no way for `get_argument_options` to be called outside of the editor. In fact, one can easily argue that all of `get_argument_options`'s overrides should be surrounded with #ifdef TOOLS_ENABLED, but it's not been done yet

servers/rendering_server.cpp
```diff
@@ -2204,6 +2204,23 @@ void RenderingServer::fix_surface_compatibility(SurfaceData &p_surface, const St
...
+	String pf = p_function;
+	if (p_idx == 0) {
+		if (pf == "global_shader_parameter_set" || pf == "global_shader_parameter_set_override" ||
+				pf == "global_shader_parameter_get" || pf == "global_shader_parameter_get_type" || pf == "global_shader_parameter_remove") {
+			for (StringName E : global_shader_parameter_get_list()) {
+				r_options->push_back(E.operator String().quote());
+			}
+		} else if (pf == "has_os_feature") {
+			for (String E : { "rgtc", "s3tc", "bptc", "etc", "etc2", "astc" }) {
```
AThousandShips, line 2216: Quote the constants here as well IMO

Mickeon: Quoting them may waste a few bytes in build size but that's splitting hairs

Mickeon:
Updated this PR to cleanse @clayjohn's worries by surrounding the method with TOOLS_ENABLED.

clayjohn (approved):
Seems fine to me. I am not familiar with the editor-side of things, but from the rendering-side it looks fine

akien-mga (approved):

akien-mga:
Thanks!

PR #86799, opened 2024-01-04
Add autocompletion for AnimationLibrary & AnimationMixer's methods

Mickeon:
Complements https://github.com/godotengine/godot/pull/86754

image

This PR autocompletes the **AnimationLibrary** methods with the names of every existing animation, if they can be detected. Admittedly, this is fairly niche.
But, AnimationMixer's methods are much less so.


fire (approved):
I am approving this from the usability perspective that it is a good change for bettering quality of life.

akien-mga:
Thanks!

PR #86804, opened 2024-01-04
Allow all editor modes to select nodes in the viewport

### LINKED ISSUE #8563: Allow for all editor modes to be able to select nodes in the viewport

image

Title

I feel like it would improve workflow to allow the other modes to select objects in the viewport, because currently nothing happens, which I believe is counter-intuitive to what someone would expect. It seems more organic to just interpret a user clicking on a viewport object when nothing else is selected as their desire to select said object. I can't think of any downsides to allowing this.  

All modes allow selection of nodes in the viewport, but the gizmo maintains the current modes functionality. 

Not that I'm aware of.

This is a pretty fundamental change to the editor's core.

ryevdokimov:
Closes [godotengine/godot-proposals#8563](https://github.com/godotengine/godot-proposals/issues/8563)

All behavior once a node is selected is the same, this only allows you to select a node with any editor mode when no node is selected.

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -1839,17 +1839,17 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 
 					clicked = ObjectID();
 
-					if ((spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_SELECT && b->is_command_or_control_pressed()) || spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_ROTATE) {
+					if (can_select_gizmos && ((spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_SELECT && b->is_command_or_control_pressed()) || spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_ROTATE)) {
 						begin_transform(TRANSFORM_ROTATE, false);
 						break;
 					}
 
-					if (spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_MOVE) {
+					if (can_select_gizmos && spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_MOVE) {
 						begin_transform(TRANSFORM_TRANSLATE, false);
 						break;
 					}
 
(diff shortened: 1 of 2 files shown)


KoBeWi:
This breaks Ruler Mode. Since it doesn't interact with nodes, it should not be capable of selecting. Same with Pan Mode.

ryevdokimov:
> This breaks Ruler Mode. Since it doesn't interact with nodes, it should not be capable of selecting. Same with Pan Mode.

Good catch, it's fixed. As well I allowed for the move tool to allow dragging in the same click as selecting, just like the select tool.

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -1839,17 +1839,17 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 
 					clicked = ObjectID();
 
-					if ((spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_SELECT && b->is_command_or_control_pressed()) || spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_ROTATE) {
+					if (can_select_gizmos && ((spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_SELECT && b->is_command_or_control_pressed()) || spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_ROTATE)) {
```
KoBeWi, line 1842: What is this change for?

ryevdokimov: The code below that section is where selection the selection process for 3D takes place. The modified code adds a check to make sure that a gizmo can be selected (implying something was selected, and one exists) before entering the body of the if statements, where transforms begin, and the encompassing case is broken from, preventing selection.

KoBeWi: Ah, for some reason I thought this already exists in 3D and the PR is 2D only 🙃

KoBeWi (approved):

editor/plugins/node_3d_editor_plugin.cpp
ryevdokimov: All good. I could have separated the PR but I figured it would by jarring if 2D did one thing and 3D did another.

PR #86805, opened 2024-01-04
Allow `ui_cancel` to unselect all nodes in all edit modes in 2D, matching 3D behavior

ryevdokimov:
Supplementary to #86804, but not dependent. 

I couldn't find a reason why this shouldn't be the case as this is the behavior in 3D.

### DIFF
--- a/editor/plugins/canvas_item_editor_plugin.cpp
+++ b/editor/plugins/canvas_item_editor_plugin.cpp
@@ -2463,7 +2463,7 @@ bool CanvasItemEditor::_gui_input_select(const Ref<InputEvent> &p_event) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


KoBeWi (approved):
> Supplementary to https://github.com/godotengine/godot/pull/86804, but not dependent.

When you press Escape to clear selection, there is no way to select back without switching tool. That could've been the reason behind original condition.

ryevdokimov:
> > Supplementary to #86804, but not dependent.
> 
> When you press Escape to clear selection, there is no way to select back without switching tool. That could've been the reason behind original condition.

Ah, so supplementary **and** dependent would have been more accurate. 😅

Kind of odd that this is the behavior in 3D then, I must have just been desensitized to it. 

Both PRs should unify the behavior of 2D and 3D and allow for a less janky workflow then IMO.

akien-mga:
Thanks!

PR #86807, opened 2024-01-05
Add static method as equivalent for HTTPClient's `query_string_from_dict`

### LINKED ISSUE #8801: Make query_string_from_dict static in HTTPClient.

A project that uses HTTPRequest to send requests to an API.

Creating an HTTPClient object every time I need to use query_string_from_dict in HTTPRequest.

The problem can be solved by making the query_string_from_dict method of HTTPClient static. I don't see any inconvenience in making this change.

In a single line, you can perform the conversion of a dictionary of parameters to a string.

If you want to send a request with the parameters q="a" and category="all" using a HTTPRequest object named http_request, you simply do:

```gdscript
http_request.request("http://www.example.com/search?" + HTTPClient.query_string_from_dict({
	"q": "a",
	"category": "all",
})
```

```gdscript
var http_client := HTTPClient.new()
http_request.request("http://www.example.com/search?" + http_client.query_string_from_dict({
	"q": "a",
	"category": "all",
})
```

I think it's a method that should be readily available in the HTTPRequest class. I believe handling it this way shouldn't pose a major issue.

Mickeon:
Closes https://github.com/godotengine/godot-proposals/issues/8801

35c7390ad9347c20297e56942f5ac668

### DIFF
--- a/core/io/http_client.cpp
+++ b/core/io/http_client.cpp
@@ -68,7 +68,7 @@ Error HTTPClient::_request(Method p_method, const String &p_url, const Vector<St
 	return request(p_method, p_url, p_headers, size > 0 ? (const uint8_t *)body_utf8.get_data() : nullptr, size);
 }
 
-String HTTPClient::query_string_from_dict(const Dictionary &p_dict) {
+String HTTPClient::query_string_from(const Dictionary &p_dict) {
 	String query = "";
 	Array keys = p_dict.keys();
 	for (int i = 0; i < keys.size(); ++i) {
@@ -95,6 +95,10 @@ String HTTPClient::query_string_from_dict(const Dictionary &p_dict) {
 		}
 	}
 	return query.substr(1);
+}
+
+String HTTPClient::query_string_from_dict(const Dictionary &p_dict) {
+	return HTTPClient::query_string_from(p_dict);
 }
 
 Error HTTPClient::verify_headers(const Vector<String> &p_headers) {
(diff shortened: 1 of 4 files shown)


akien-mga:
This breaks compatibility. I'm not sure if this is something we can resolve with compatibility bindings for both GDExtension and C#.

CC @dsnopek @raulsntos

dsnopek:
I think it'd be possible to register a compatibility method for the non-static method, which calls the static method. I've never tried that, but the method binds (both static and non-static) are all on the same list together, so I think it should work in preserving binary compatibility. Someone just needs to try it :-)

This would definitely break source compatibility for a lot of bindings, but that's true of many changes where we provide compatibility methods. And we're usually OK with that, since developers will only encounter it when they update their bindings.

raulsntos:
C# can't have overloads where the only thing that changes is the `static`-ness of the method. So we wouldn't be able to generate the compat method since it would be invalid C# syntax, and therefore it will break binary compatibility. Can we instead add a new static method with a different name and deprecate the instance method?

Mickeon:
Sounds reasonable. I have a hard time coming up with a new name, though, unless it's just... `query_string_from`, `make_query_string_from`, `query_from_dict`...

Worst case scenario, this is just not doable until the next major release, and having it is not exactly urgent.

Mickeon:
This PR has been modified to add `query_string_from` (better naming suggestions are welcome). The existing method still exists, but is marked as deprecated, maintaining compatibility.

PR #61834, opened 2022-06-09
Jitter shadow map dithering pattern across frames when TAA is enabled

Calinou:
This improves shadow quality by reducing the visibility of the noisy pattern caused by dithering.

This jittering also applies when FSR2 is enabled, as it provides its own form of temporal antialiasing.

This closes https://github.com/godotengine/godot-proposals/issues/4179 and closes https://github.com/godotengine/godot/issues/53534.

**Testing project:** [test_shadow_jitter.zip](https://github.com/godotengine/godot/files/11579988/test_shadow_jitter.zip)

## Preview

*Click to view at full size (this is required for accurate preview). TAA is enabled on all screenshots.*

### DirectionalLight3D with non-PCSS shadows

| Before | After
|-|-|
|  |  |

### OmniLight3D with PCSS shadows

| Before | After
|-|-|
|  |  |

<details>
<summary>Older revision of this PR</summary>

### DirectionalLight3D with non-PCSS shadows

| Before | After
|-|-|
|  |  |

### OmniLight3D with PCSS shadows

| Before | After
|-|-|
|  |  |
</details>

PR #86809, opened 2024-01-05
Tune TAA disocclusion scale to avoid rejecting all samples during motion.

clayjohn:
Noticed while reviewing https://github.com/godotengine/godot/pull/61834

The values we use for the disocclusion scale were a bit problematic. Basically they mean the following:
`disocclusion_threshold`: If the difference in velocity of this pixel between this frame and last frame is less than this amount, we can happily use the reprojected color. If the difference is greater we weight this frame more heavily (basically we assume there is a good chance that this pixel is the result of disocclusion)
`disocclusion_scale`: This scales the disocclusion amount (which is measured in pixels `(difference_in_velocity - disocclusion_threshold)`) to arrive at a final "disocclusion" value.

The final disocclusion value is used to weight towards the current sample. A disocclusion value of 1 means "throw away all previous samples". 0 means, this isn't a disoccluded pixel, blend it as normal. 

As a result of having such high values, we pretty much always had a disocclusion value clamped to 1 if there was any movement at all. 

https://www.elopezr.com/temporal-aa-and-the-quest-for-the-holy-trail/ has a nice breakdown of this type of disocclusion factor. The author uses 0.001 for threshold and 10 for scale. But they measure disocclusion in uv-space and we measure in texel-space. Accordingly, I increased/decreased the values by a factor of 1,000 and the results look a lot better. 

They are most noticable in shadows with https://github.com/godotengine/godot/pull/61834 applied. 

_In this screenshot the camera is moving slowly. On the left is master, on the right is this PR_
Screenshot from 2024-01-04 12-16-50

I tested this PR in a couple of scenes, both with and without https://github.com/godotengine/godot/pull/61834 and cannot see any signs of new ghosting or other artifacts that can come from improper disocclusion handling.

### DIFF
--- a/servers/rendering/renderer_rd/effects/taa.cpp
+++ b/servers/rendering/renderer_rd/effects/taa.cpp
@@ -62,8 +62,8 @@ void TAA::resolve(RID p_frame, RID p_temp, RID p_depth, RID p_velocity, RID p_pr
 	memset(&push_constant, 0, sizeof(TAAResolvePushConstant));
 	push_constant.resolution_width = p_resolution.width;
 	push_constant.resolution_height = p_resolution.height;
-	push_constant.disocclusion_threshold = 0.025f;
-	push_constant.disocclusion_scale = 10.0f;
+	push_constant.disocclusion_threshold = 2.5f; // If velocity changes by less than this amount of texels we can retain the accumulation buffer.
+	push_constant.disocclusion_scale = 0.01f; // Scale the weight of this pixel calculated as (change in velocity - threshold) * scale.
 
 	RD::ComputeListID compute_list = RD::get_singleton()->compute_list_begin();
 	RD::get_singleton()->compute_list_bind_compute_pipeline(compute_list, pipeline);
--- a/servers/rendering/renderer_rd/shaders/effects/taa_resolve.glsl
+++ b/servers/rendering/renderer_rd/shaders/effects/taa_resolve.glsl
@@ -307,6 +307,8 @@ float luminance(vec3 color) {
 	return max(dot(color, lumCoeff), 0.0001f);
 }
 
+// This is "velocity disocclusion" as described by https://www.elopezr.com/temporal-aa-and-the-quest-for-the-holy-trail/.
+// We use texel space, so our scale and threshold differ.
 float get_factor_disocclusion(vec2 uv_reprojected, vec2 velocity) {
 	vec2 velocity_previous = imageLoad(last_velocity_buffer, ivec2(uv_reprojected * params.resolution)).xy;
 	vec2 velocity_texels = velocity * params.resolution;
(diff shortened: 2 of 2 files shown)


mrjustaguy:
I think 2.5 pixels is a bit too much, as that'll allow pixels to bleed into neighbors up to 3 pixels away.

The optimal value is likely sqrt(2)/2 pixels as that is the distance between center of the pixel and corner of the pixel, and there's no point in going outside of the pixel's grid for reprojected samples, as they should be relevant for different pixels, but not the current one.

clayjohn:
> I think 2.5 pixels is a bit too much, as that'll allow pixels to bleed into neighbors up to 3 pixels away.
> 
> The optimal value is likely sqrt(2)/2 pixels as that is the distance between center of the pixel and corner of the pixel, and there's no point in going outside of the pixel's grid for reprojected samples, as they should be relevant for different pixels, but not the current one.

This isn't a pixel distance. Its a texel difference in velocities. Basically it says, if the velocity from last frame changes by less than this amount, treat it as if its the same pixel. IMO 2.5 is a really small amount and it would be safe to go up even higher

mrjustaguy:
ah, I misunderstood the above in that case.

Calinou:
Tested locally in https://github.com/Calinou/godot-reflection (CLI arguments: --write-movie output.avi --quit-after 600 -- --benchmark`).

Video showcasing `master` (left), PR (right):

https://github.com/godotengine/godot/assets/180032/175d764f-9dd3-47c7-9d2d-b924aeed9737

TAA coverage seems a lot better with this PR, but it's blurrier and ghosting is a lot more common. Some stills showing ghosting:

### On top of the staircase in the middle

taa1

### Behind the ladder in the top-right corner

taa2

After fast camera movement, the PR has much better TAA coverage but it's also blurrier if you look at the grass pot in the distance:

taa3 webp

mrjustaguy:
IIRC the adaptive box size is quite problematic, too high values and it'll ghost like crazy, too low values and it's a jittery mess.. Figuring out the proper function for calculating it would improve the ghosting seen in the stairs example, but figuring that out is a very big ask IMO

clayjohn:
> TAA coverage seems a lot better with this PR, but it's blurrier and ghosting is a lot more common. Some stills showing ghosting:

As mrjustaguy says solving this problem will be pretty challenging. The reason we didn't have any ghosting before is because we had no reprojection at all. Same thing with the perceived blurriness in motion, there was no blurriness before because there was no TAA

Ansraer:
Hmm, looking at the blurrier textures seen in Calinou's last screenshot, I am wondering if this is caused by the camera jitter (which we need for TAA to work).

If so, we might be able to alleviate the problem by adjusting the texture coordinates in the fragment shader. The idea would be to move it in the opposite direction of the camera jitter to ensure that an object's textures are stable across several frames, which would reduce the blurring when TAA looks at past frames. I think I read about this technique somewhere a while ago, but can't remember where it was right now.

BastiaanOlij (approved):
LGTM, seeing these values are sourced from the original implementation. From reading the reactions, it feels like we're just seeing TAA in action.

I suggest we merge this and see how people react to the visual change.

akien-mga:
Thanks!

PR #32949, opened 2019-10-20
Implement debug memory marking

RandomShaper:
This consists in marking memory regions with certain specific values when they are (re)allocated or
freed in order to ease debugging of certain bugs, catching them as early as possible, detect
uninitialized variables, etc.

Some C++ implementations may already do this in debug mode, but the idea of this is to make it work
exactly the same regardless the platform/compiler/library.

```
ALLOCATED
+-------------------------+
| F1 F1 F1 F1 F1 F1 F1 F1 |
+-------------------------+

REALLOCATED (new size > old size)

/----------------- new size -----------------/
/------- old size -------/---- size diff ----/

+------------------------+===================+
|   current contents     | F2 F2 F2 F2 F2 F2 |
+------------------------+===================+

REALLOCATED (new size < old size)

/----------------- old size -----------------/
/------- new size -------/---- size diff ----/

+------------------------+···················+
|   kept contents        | F3 F3 F3 F3 F3 F3 :
+------------------------+···················+

FREED
+·························+
: F4 F4 F4 F4 F4 F4 F4 F4 :
+·························+
```

PR #86818, opened 2024-01-05
Junk fill freed memory on dev builds to help find use after frees.

tdaven:
While investigating #86427, not all systems would crash.  My desktops with Fedora 39 crash consistently but other systems like mingw windows build did not.  This makes it hard for others to confirm the issue or investigate when they can't reproduce the issue.

Filling memory with junk on free helps give deterministic behavior.  Its currently guarded by DEV_ENABLED where there are already lots of slower checks being done.  There is nothing magic about the 0xab value I'm filling with.  It was picked as I've used it before.

With junk filling, the mingw build mentioned above consistently crashes the same as my Fedora 39 systems.

Since it was helpful for confirming behavior looking into #86427, I figured make a PR for it.  I could easily see something like this not being desired (and it fairly easy to do when needed).

### DIFF
--- a/core/os/memory.h
+++ b/core/os/memory.h
@@ -35,6 +35,7 @@
 #include "core/templates/safe_refcount.h"
 
 #include <stddef.h>
+#include <cstring>
 #include <new>
 #include <type_traits>
 
@@ -100,6 +101,12 @@ _ALWAYS_INLINE_ bool predelete_handler(void *) {
 	return true;
 }
 
+#ifdef DEV_ENABLED
+#define junk_fill(p_mem, p_size) memset(static_cast<void *>(p_mem), 0xab, p_size)
+#else
+#define junk_fill(p_mem, p_size)
+#endif
+
 template <class T>
 void memdelete(T *p_class) {
 	if (!predelete_handler(p_class)) {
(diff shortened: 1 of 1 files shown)


RandomShaper:
I'm generally in favor of these measures. In fact, I PR'd something similar time ago (#32949), but considering it was rejected at the time, maybe we need to re-discuss it.

Ansraer:
Sounds like a sensible change to me unless it has too much of a performance cost (dev builds are already quite slow).
Though are you sure that 0xAB is a good number? I think that 0xBAD or 0xBADC0DE would be better choices due to uhm... reasons. 😁

tdaven:
Using 0xab is just an easy thing.  We could do some multi-byte value if desired.  That just means we have to calculate how many times we can copy that value into the range (unless we are comfortable falling back on alignment requirements).  You just want something that is likely to cause issues if accessed and be obvious.  Zero or 0xff are less obvious, a big block of repeating value of 0xab or something else is pretty easy to spot.

For reference, [jemalloc](https://jemalloc.net/jemalloc.3.html) defaults to 0x5a (search for opt.junk in the manual) when enabled.

PR #86822, opened 2024-01-05
Feature/4.x cli add export files argument

### LINKED ISSUE #8748: Godot CLI for --export-pack needs a option about preset option override

I want to implement an incremental update system for my project. I don't want to pack all the resources of the entire project into one large PCK file, instead I plan to pack each file into a separate PCK file and perform incremental updates on a file-by-file basis. "

However, the export preset feature only supports pre-setting the resources to be packed into a PCK, and cannot pack each resource into a separate PCK file. So, I tried using Godot's CLI (--export-pack) to pack each resource into a PCK file. However, I found that --export-pack also uses the preset to pack, and it does not support overriding the export_files configuration item to dynamically specify the resources. As a result, I had to modify the export_presets.cfg file continuously using a Python script to achieve PCK export for each resource.

I hope to add a feature to override the preset for --export-pack, which would allow me to have more flexibility in controlling the incremental packing of PCK files, and even achieve automated patching.

Thie new option argment is just like below.

```
godot --headless --export-pack -p export_files=PackedStringArray("res://myscene.tscn")  "Windows Desktop" ./dist/window
```

This difficult. I can use python too modified export_presets.cfg for each file export, but it's very slow Because there only one export_presets.cfg，can't do some concurrent.

This is about CI/CD, good thing to builtin.

kaluluosi:
This PR is about this proposal:
 [Add support for exporting specific resources using the --export-pack command line argument #8748](https://github.com/godotengine/godot-proposals/issues/8748#issue-2060560960)

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/8748#issue-2060560960*


main/main.cpp
```diff
@@ -510,6 +510,7 @@ void Main::print_help(const char *p_binary) {
 	OS::get_singleton()->print("                                    The target directory must exist.\n");
 	OS::get_singleton()->print("  --export-debug <preset> <path>    Export the project in debug mode using the given preset and output path. See --export-release description for other considerations.\n");
 	OS::get_singleton()->print("  --export-pack <preset> <path>     Export the project data only using the given preset and output path. The <path> extension determines whether it will be in PCK or ZIP format.\n");
+	OS::get_singleton()->print("  --export-files <file1,file2,...>  Specific files to export. It overrides the 'Resources to export' option of preset.\n");
```
AThousandShips, line 513: This should probably be made more clear, making it clear that the files have to be one uninterrupted string with no spaces

kaluluosi: > Specific files to export.A string of filenames separated by comma(no spacees interrupted). It overrides the 'Resources to export' option of preset.

Do you think this is okay?

AThousandShips: Sounds good I think, though the suggestion on the proposal was to allow files and directories, and to have separated by spaces, would need to be discussed 🙂

kaluluosi: I just override the  the presets.selected_files. Didn't modified anything else. So this --export-files indeed  supports both directory path and file path.
image

Or I can just use path not file to confuse people.
> --export-paths <path1,path2,...>  Specific paths(starts  with res://) to export. A string of paths separated by comma(no spacees interrupted). It overrides the 'Resources to export' option of preset.\n"

AThousandShips: But the files are files only, not directories, many places check specifically for files

kaluluosi: The proposal  I write does not talk about support both file and directory, just file.
But It exactly a problem that I didn't check if path is file or directory.
I think we should make it pure. `--export-files` just files.

AThousandShips: I'm talking about this: https://github.com/godotengine/godot-proposals/issues/8748#issuecomment-1873329157

I think just files is a bit too restrictive, but it'll be something to discuss

kaluluosi: Yeb. It's not hard to support directory path. Just find out the direcotry path string and recurse it. I will take a try.

kaluluosi: I did it. Add support for both file and directory and test it locally.

The cli help I change to this.
```shell
--export-files <path1,path2,...>  Specific a comma-separated list of files and folders to export(with 'res://'). It overrides the 'Resources to export' option of preset.
                                    WARN: It must be used with --export-release\debug\pack.It doesn't export but just override preset. 
```

kaluluosi: And the 'expot files without preset' is too difficult for me. We can't export without preset as it targets platforms. I don't know if it is good to create default preset to insteal `p_preset`. That is export workflow thing. I'm noob I can't handle this. 
So this PR just contains --export-files argument.

PR #82264, opened 2023-09-24
Core: Allow methods of built-in `Variant` types to be used as Callables

dalexeev:
* Closes #82251.
* Reverts #78274.

Makes the following code work:

```gdscript
var array: Array = [1, 2, 3]
print(array) # [1, 2, 3]
var callable: Callable = array.clear
callable.call()
print(array) # []
```

~~Note that signals can still only be connected to `Object`-based callables.~~ Fixed by #82695.

PR #86823, opened 2024-01-05
GDScript: Allow utility functions to be used as `Callable`

dalexeev:
* This PR aims to improve GDScript consistency. After #82264, it makes sense to allow utility functions to be used as `Callable`s too.

```gdscript
prints.callv([1, 2, 3])
print(mini.call(1, 2))
timer.timeout.connect(print.bind("Completed!"))
```

Also, this makes it possible to call vararg utility functions (such as `print()`) with `Callable.callv()` without boilerplate.

### DIFF
--- a/modules/gdscript/gdscript_utility_callable.h
+++ b/modules/gdscript/gdscript_utility_callable.h
@@ -1,0 +1,65 @@
+/**************************************************************************/
+/*  gdscript_utility_callable.h                                           */
+/**************************************************************************/
+/*                         This file is part of:                          */
+/*                             GODOT ENGINE                               */
+/*                        https://godotengine.org                         */
+/**************************************************************************/
+/* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+/* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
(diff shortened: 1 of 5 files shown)


modules/gdscript/gdscript_utility_callable.h
```diff
@@ -0,0 +1,65 @@
...
+class GDScriptUtilityCallable : public CallableCustom {
+	StringName function_name;
+	enum Type {
+		TYPE_INVALID,
+		TYPE_GLOBAL,
+		TYPE_GDSCRIPT,
+	};
+	Type type = TYPE_INVALID;
+	GDScriptUtilityFunctions::FunctionPtr gdscript_function = nullptr;
+	uint32_t h = 0;
```
adamscott, line 47:
```suggestion
	uint32_t _hash = 0;
```
Just to be a little bit less cryptic than a standalone `h`.

adamscott (approved):
Reviewed during the GDScript meeting. Looks good! ~~With the rename of `h` to `_hash`~~ (edit: postponed after the merge), it seems that everything is good to merge now.

modules/gdscript/gdscript_utility_callable.h
dalexeev: The name is copied from other `CustomCallable`s, in the core and in the GDScript module. Should I fix this in other `CustomCallable`s too?

adamscott: It's ok then. We could do another PR to rename those variables all together.

AThousandShips (approved):

akien-mga:
Thanks!

PR #86824, opened 2024-01-05
Fix playback position label update in Audio Stream Importer

### LINKED ISSUE #86550: Audio Stream Importer playback position is not updated unless playhead is visible
### Tested versions

- Reproducible in: v4.2.1.stable.official [b09f793f5] and v4.3.dev1.official [9d1cbab1c]

### System information

Godot v4.2.1.stable - Windows 10.0.22621

### Issue description

When imported audio is playing in Audio Stream Importer, unless the playhead is visible, current playback position is not updated.

### Steps to reproduce

- Import audio file into project (tested with >3 minute long .mp3)
- Double-click on imported file in File System tab to open Audio Stream Importer
- Press play in bottom-left corner, notice that current playback position in the bottom-right corner is updating
- Make playhead leave your screen by zooming in and etc., notice that current playback position is no longer updated

N/A

alessandrofama:
The current playback position label doesn't get updated if the playhead is not visible. This can be visualized in this example:

[audio-stream-importer-1.webm](https://github.com/godotengine/godot/assets/26153311/886c74e7-32ce-4e34-be46-a87eb6920032)

The label is updated in `_draw_indicator()` at the end of the function, but we return early if the calculated offset falls outside the valid range. This fix moves the label updating above this check, ensuring that the label always gets updated regardless of the visibility of the playhead:

[audio-stream-importer-2.webm](https://github.com/godotengine/godot/assets/26153311/6e7460c8-af47-4fc4-946e-9ceca8d49c43)

Fixes #86550.

### DIFF
--- a/editor/import/audio_stream_import_settings.cpp
+++ b/editor/import/audio_stream_import_settings.cpp
@@ -277,6 +277,8 @@ void AudioStreamImportSettingsDialog::_draw_indicator() {
 		rect.size.height -= y_ofs;
 	}
 
+	_current_label->set_text(String::num(_current, 2).pad_decimals(2) + " /");
+
 	float ofs_x = (_current - zoom_bar->get_value()) * rect.size.width / zoom_bar->get_page();
 	if (ofs_x < 0 || ofs_x >= rect.size.width) {
 		return;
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #86827, opened 2024-01-05
[Pseudolocalization] Prevent out of bounds reads.

### LINKED ISSUE #86825: "Unicode parsing error" with double vowels pseudolocalization enabled
### Tested versions

Reproducible on master

### System information

Godot v4.2.dev (691634969) - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2070 SUPER (NVIDIA; 31.0.15.4584) - AMD Ryzen 7 5800X3D 8-Core Processor (16 Threads)

### Issue description

Enabling "Double Vowels" in the pseudolocalization settings is spamming a bunch of `Unicode parsing error, some characters were replaced with � (U+FFFD): NUL character`.

I tracked it down to `TranslationServer::double_vowels()` in `core/string/translation.cpp`, which seems to loop one too many times. The line `res += p_message[i]` in the middle calls `String::operator+=(char32_t)` which checks for 0 and logs the error.

```
String TranslationServer::double_vowels(String &p_message) const {
	String res;
	for (int i = 0; i < p_message.size(); i++) {
		if (pseudolocalization_skip_placeholders_enabled && is_placeholder(p_message, i)) {
			res += p_message[i];
			res += p_message[i + 1];
			i++;
			continue;
		}
		res += p_message[i];
		if (p_message[i] == 'a' || p_message[i] == 'e' || p_message[i] == 'i' || p_message[i] == 'o' || p_message[i] == 'u' ||
				p_message[i] == 'A' || p_message[i] == 'E' || p_message[i] == 'I' || p_message[i] == 'O' || p_message[i] == 'U') {
			res += p_message[i];
		}
	}
	return res;
};
```
I'm not familiar with the internals of Godot, but shouldn't `p_message.size()` be `p_message.length()`? It looks like `String::size()` returns `_cowdata.size()`, which includes the null terminator:

```
_FORCE_INLINE_ int size() const { return _cowdata.size(); }
```

whereas `String::length()` returns the number of characters, excluding the null terminator:

```
_FORCE_INLINE_ int length() const {
	int s = size();
	return s ? (s - 1) : 0; // length does not include zero
}
```

If that's the case, there's a bunch more `size()` calls in that file.

### Steps to reproduce

I'm using godot-cpp, but just this seems to reproduce it:

```
godot::String s("123");
godot::TranslationServer::get_singleton()->pseudolocalize(s);
```

N/A

bruvzg:
Fixes https://github.com/godotengine/godot/issues/86825

### DIFF
--- a/core/string/translation.cpp
+++ b/core/string/translation.cpp
@@ -881,7 +881,7 @@ StringName TranslationServer::tool_pseudolocalize(const StringName &p_message) c
 
 String TranslationServer::get_override_string(String &p_message) const {
 	String res;
-	for (int i = 0; i < p_message.size(); i++) {
+	for (int i = 0; i < p_message.length(); i++) {
 		if (pseudolocalization_skip_placeholders_enabled && is_placeholder(p_message, i)) {
 			res += p_message[i];
 			res += p_message[i + 1];
@@ -895,7 +895,7 @@ String TranslationServer::get_override_string(String &p_message) const {
 
 String TranslationServer::double_vowels(String &p_message) const {
 	String res;
-	for (int i = 0; i < p_message.size(); i++) {
+	for (int i = 0; i < p_message.length(); i++) {
 		if (pseudolocalization_skip_placeholders_enabled && is_placeholder(p_message, i)) {
 			res += p_message[i];
 			res += p_message[i + 1];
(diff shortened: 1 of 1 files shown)


akien-mga (approved):
Looks good.

Are there risks of this misuse of `size` vs `length` being more widespread in the codebase? Seems dangerous.

bruvzg:
> Are there risks of this misuse of size vs length being more widespread in the codebase? Seems dangerous.

Should not be too dangerous. There's always a terminating zero, so it won't crash. In most cases, it will just unnecessarily process or copy an extra zero char.

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #77712, opened 2023-05-31
Fix lossless formats in PortableCompressedTexture2D

nklbdev:
This is a right copy of my [previous wrong pull-request](https://github.com/godotengine/godot/pull/77047)

Sorry for my bad English(((

`PortableCompressedTexture2D` have a problem with creating from PNG images and some other little problems.
I want to use this resource type as the way to import animations from Aseprite in my [Aseprite importers plugin bundle](https://github.com/nklbdev/godot-4-aseprite-importers)

For now I am creating one PNG image near every `*.aseprite`-file that imported. Then call some methods to update filesystem and import the image. Then plugin is continuing it's work and embedding new image in an imported resource.

`PortableCompressedTexture2D` is a beautiful way to embed PNG data silently in `SpriteFrames` or other resources. But it is not usable now. Method `create_from_image` cannot load PNG image and shows errors in console:

```
WARNING: Not a PNG file
     at: PNGDriverCommon::check_error (drivers\png\png_driver_common.cpp:56)
ERROR: Condition "!success" is true. Returning: ERR_FILE_CORRUPT
   at: PNGDriverCommon::png_to_image (drivers\png\png_driver_common.cpp:69)
ERROR: Condition "err" is true. Returning: Ref<Image>()
   at: ImageLoaderPNG::load_mem_png (drivers\png\image_loader_png.cpp:64)
ERROR: It's not a reference to a valid Image object.
   at: (.\core/io/image.h:426)
ERROR: Condition "err" is true. Returning: Ref<Image>()
   at: _jpegd_mem_loader_func (modules\jpg\image_loader_jpegd.cpp:131)
ERROR: It's not a reference to a valid Image object.
   at: (.\core/io/image.h:426)
ERROR: Condition "p_width <= 0 || p_width > 16384" is true.
   at: GLES3::TextureStorage::texture_set_size_override (drivers\gles3\storage\texture_storage.cpp:1053)
```

This is because PNG image have not only PNG prefix `‰PNG`, but also [Godot's own `PNG ` prefix](https://github.com/godotengine/godot/blob/ffd32a244b43ff58c13819c2debf8cf3b58ecbdc/drivers/png/image_loader_png.cpp#L80)

I understand that this solution is not ideal.
Theoretically, there could be such a sequence of bytes PNG  in this memory location for some other reason. Not because Godot added it there. I don't know how to be completely sure of this. If you know such a criteria, please tell me and I will correct the code.

[...]

PR #86835, opened 2024-01-05
Fix data format storage method for PortableCompressedTexture2D

nklbdev:
This is a fix after previous merged PR: https://github.com/godotengine/godot/pull/77712

It fixes backwards compatibility after this PR:
The main idea is:
- Remove the problem enum value
- Encode data format as `data_format + 1` to distinguish it from files, generated by previous versions of Godot.
- Subtract 1 while `data_format` decoding if `data_format_code` is greater than 0

Why i cannot just place the `DATA_FORMAT_UNDEFINED` in back of the enum or make it's value so big:
The `DATA_FORMAT_UNDEFINED` is the format that all the old images are in. That's why I can't put it at the end of the enum values list. It must be equal 0. Therefore, it conflicts with the `DATA_FORMAT_IMAGE`. That's why he had to be removed.

Alternative way:
Store the boolean “new data format” sign in a separate byte. And store `data_format` as is (without adding 1)

### DIFF
--- a/scene/resources/compressed_texture.h
+++ b/scene/resources/compressed_texture.h
@@ -40,7 +40,6 @@ class CompressedTexture2D : public Texture2D {
 
 public:
 	enum DataFormat {
-		DATA_FORMAT_UNDEFINED,
 		DATA_FORMAT_IMAGE,
 		DATA_FORMAT_PNG,
 		DATA_FORMAT_WEBP,
--- a/scene/resources/portable_compressed_texture.cpp
+++ b/scene/resources/portable_compressed_texture.cpp
@@ -33,7 +33,6 @@
 #include "core/config/project_settings.h"
 #include "core/io/marshalls.h"
 #include "scene/resources/bit_map.h"
-#include "scene/resources/compressed_texture.h"
 
 void PortableCompressedTexture2D::_set_data(const Vector<uint8_t> &p_data) {
 	if (p_data.size() == 0) {
@@ -44,7 +43,7 @@ void PortableCompressedTexture2D::_set_data(const Vector<uint8_t> &p_data) {
 	uint32_t data_size = p_data.size();
 	ERR_FAIL_COND(data_size < 20);
 	compression_mode = CompressionMode(decode_uint16(data));
-	CompressedTexture2D::DataFormat data_format = CompressedTexture2D::DataFormat(decode_uint16(data + 2));
+	DataFormat data_format = DataFormat(decode_uint16(data + 2));
 	format = Image::Format(decode_uint32(data + 4));
 	uint32_t mipmap_count = decode_uint32(data + 8);
 	size.width = decode_uint32(data + 12);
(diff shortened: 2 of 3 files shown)


AThousandShips:
When you say old data? Is this data from before #77712? Where was that handling added?

nklbdev:
@AThousandShips, yes, I mean the resources created by the version of Godot before https://github.com/godotengine/godot/pull/77712. When the user downloads the new version of the engine and tries to open his old project, he should succeed.

nklbdev:
I'm afraid I didn't make myself clear enough.

The new version of the code will not save `DATE_FORMAT_UNDEFINED` ever. This value was needed so that when reading the resource bytes of the old version, we could understand that for this image [we do not need to use the `loader_func`](https://github.com/godotengine/godot/blob/89cc635c0554cb2e518c830969ca4c5eedda0f4e/scene/resources/portable_compressed_texture.cpp#L63C32-L63C32), as in the old version - this is for backward compatibility.

Therefore, this enum value must be removed. Instead, we need to store the mark of the new data format in some other way.

AThousandShips:
Then that works fine, but no need to encode the old format assuming the value is always 0 in those bytes for old encoded formats, assuming it's only used when loading old files, loaded data shouldn't have it when in memory, assuming the reading from file is correct

nklbdev:
If we will read this byte from old resource (0) with this enum contents, we will get `DATA_FORMAT_IMAGE`. This is wrong, because this will select `loader_func == Image::_png_mem_unpacker_func`:
```
	enum DataFormat {
		DATA_FORMAT_IMAGE, // <- this value will be read incorrectly instead of DATA_FORMAT_UNDEFINED
		DATA_FORMAT_PNG,
		DATA_FORMAT_WEBP,
		DATA_FORMAT_BASIS_UNIVERSAL,
		DATA_FORMAT_UNDEFINED, // <- this value at this place is useless
	};
```

But this order of values is breaking compatibility too
```
	enum DataFormat {
	        DATA_FORMAT_UNDEFINED,
		DATA_FORMAT_IMAGE,
		DATA_FORMAT_PNG,
		DATA_FORMAT_WEBP,
		DATA_FORMAT_BASIS_UNIVERSAL,
	};
```

Therefore, the `DATA_FORMAT_UNDEFINED` value will have to be removed.

AThousandShips:
Yes, that's why offsetting it solves it, that's what I said about the value being offset, I agreed that the value has to be removed from the enumeration, but you don't need to indicate that it's new or old format except by it either having or not having a specified format, the problem isn't there but that the compressed texture data is broken, not the portable one

AThousandShips:
No, because you offset it by 1 🙂, just like your code here, you don't need to store it being in the old format any other way:
> Then that works fine, but no need to encode the old format assuming the value is always 0 in those bytes for old encoded formats

nklbdev:
Then I probably don’t understand what exactly broken.
Do you have any ideas on how to fix this?

I don't want problems for the development of the engine. Therefore, you can always roll back my previous pull request. I will not be against.

AThousandShips:
Just do this way that you already are, using `0` to mean "this is an old file, update it" and the rest to be explicit, the problem is that `CompressedTexture` uses the same `DataFormat`, so files saved before your PR stored `0` to mean `DATA_FORMAT_IMAGE`, but now that one isn't valid, and so on, that's why changing enums is dangerous 🙂

nklbdev:
@AThousandShips, tell me please, does the solution I present in the current PR solve this problem? It seems that it should.

AThousandShips:
I think so yes 🙂 sorry said so several times I'm sorry if I'm unclear, others will have to review it but to my eyes this should solve it, the issue was the added entry in the enum really, the other data is unchanged from your previous PR, it just has a manual addition and subtraction to make it work the same way

AThousandShips:
No problem! I understand 🙂

AThousandShips:
I'll come back to look later today but I realized that you should probably add a new enum to `PortableCompressedTexture` to handle this, to prevent future problems, as you might not assume that the one in `CompressedTexture` is used in an unrelated class, just copy the one that's there, with the undefined one intact, and leave the code you originally added as it was, but restore the code in `CompressedTexture`

AThousandShips:
You didn't remove it from `CompressedTexture2D`, please fix that 🙂

nklbdev:
@AThousandShips, thank you, you are a genius!
I replaced my commit with your suggestions.
I was too timid. I was afraid to influence too much a codebase that is not mine.
And it didn’t even occur to me that I could introduce a new enumeration into the code!

AThousandShips (approved):
This solves the regression, haven't tested the portable format changes, but they are unchanged from the code in the previous PR

akien-mga:
Thanks!

PR #86845, opened 2024-01-05
Avoid regressing in progress reporting in resource load

RandomShaper:
Given how the dependency tree is discovered (on the go), the way the load progress is computed can make it to get lower. This PR simply avoids it by not allowing to report a lower value than the maximum of those already given.


YuriSizov (approved):
Is this just about how the import looks visually, i.e. the behavior of the progress bar?

Code makes sense if we want to make sure the completeness never regresses.

akien-mga:
Thanks!

RandomShaper:
> Is this just about how the import looks visually, i.e. the behavior of the progress bar?

Yes. For a proper approach that doesn't need this band-aid in the first place, resource loading should happen quite differently.

akien-mga:
Cherry-picked for 4.2.2.

PR #86847, opened 2024-01-05
Fix `TileMap` quadrant canvas item position not being local

### LINKED ISSUE #86812: TileSet Vertex Shader unexpected behaviour- rendering_quadrant_size affects VERTEX value
### Tested versions

- Found in: v4.2.1.stable.official [b09f793f5]

### System information

Godot v4.2.1.stable - Windows 10.0.22621 - Vulkan (Forward+) - dedicated GeForce GT 730 () - Intel(R) Core(TM) i3-10100F CPU @ 3.60GHz (8 Threads)

### Issue description

When the TileMap is drawn, at VERTEX position x:256 or y:256 (or both), it causes the vertex shadow to produce unexpected behaviour.

imagen_2024-01-04_193831252

### Steps to reproduce

My tile size is 16*16.

the code required for this shader is the following at it must be used in a TileSet Tile using the TileSet editor.
```
// the UV values are not used here because they are relative to the atlas texture. therefore the UV.x and UV.y do not represent values from 0.0 to 1.0

varying vec2 pos;
void vertex() {
	pos = VERTEX / 32.0; //32.0 is later referred as "division value"
}
void fragment() {
	vec2 lpos = mod(pos,1.0);
	COLOR.rgb = vec3(lpos.x, lpos.y, 0.5);
}
```
The code allows to pass to fragment a value that represent the coordinates of the tile (`vec2 pos`).
Then it converts it to a value between `0.0; 0.5` and `0.5; 1.0` (since the position is divided by 32 instead of 16).

*this allows to create gradients conntecting 4 tiles*

The error appears in every tile that is in x:256 or y:256, in order to provide example, take tile at `(x: 256, y: 0)`: the VERTEX value will have the values: 

| (240.0, 00.0) | (256.0, 00.0) |
|----|----|
| (240.0, 16.0) | (256.0, 16.0) |

Since is not posible to debug the shader, this values were obtain comparing the VERTEX value of the affected tiles with their neighbors using conditionals inside the fragment shader. This revealed that:

these are the same values of the tile located at `(x:240, y:0)`

However, this error **will not occur** if  rendering_quadrant_size is set to 32 or the division value is set to 16.

[bugreport.zip](https://github.com/godotengine/godot/files/13836833/bugreport.zip)

edit: i corrected the table values (the VERTEX values for each vertex for tile `(256, 0)`), values were misplaced

kleonc:
Should fix #86812, see https://github.com/godotengine/godot/issues/86812#issuecomment-1879145463 for details.

### DIFF
--- a/scene/2d/tile_map_layer.cpp
+++ b/scene/2d/tile_map_layer.cpp
@@ -500,7 +500,7 @@ void TileMapLayer::_rendering_quadrants_update_cell(CellData &r_cell_data, SelfL
 			quadrant_coords = Vector2i(
 					coords.x > 0 ? coords.x / quad_size : (coords.x - (quad_size - 1)) / quad_size,
 					coords.y > 0 ? coords.y / quad_size : (coords.y - (quad_size - 1)) / quad_size);
-			canvas_items_position = quad_size * quadrant_coords;
+			canvas_items_position = tile_map_node->map_to_local(quad_size * quadrant_coords);
 		}
 
 		Ref<RenderingQuadrant> rendering_quadrant;


groud:
Thanks for the contribution. 

I am not sure this makes sense, and not sure the original issue either. I think this will break anyway as soon as you have tiles that are bigger than a cell, with either big tiles, or a texture region size configured in the atlas bigger than the tile size. 

The "top left corner" of the CanvasItem cannot be determined unless you know which tile is in the top left cell, which is a problem as all vertex coordinates would change depending on it. 

For me, the original behavior is expected and preferable.

kleonc:
First of all I had a bug in this PR as `0.5 * tile_set->get_tile_size()` does `int * Vector2i` aka result is always zero... :upside_down_face: Pushed corrected version with a cast.


>I am not sure this makes sense, and not sure the original issue either. I think this will break anyway as soon as you have tiles that are bigger than a cell, with either big tiles, or a texture region size configured in the atlas bigger than the tile size.
>
> The "top left corner" of the CanvasItem cannot be determined unless you know which tile is in the top left cell, which is a problem as all vertex coordinates would change depending on it.

Oh, I meant the top left of the bounding rect of the quadrant's "first cell" (cell at `quad_size * quadrant_coords` map coords). AFAICT this depends only on the TileSet settings + TileMap quadrant size, such quadrant origin doesn't depend on whether/what tiles are rendered (texture size doesn't affect cell counds).

Like for isometric or hexagonal shapes using `VERTEX` in the shader seems kinda pointless anyway since all tiles are rendered as quads regardless of the shape and thus for hex/iso such quads overlap the neighbors / their vertices are outside of the actual cell area.


The most important part is that when opening this PR I haven't remembered about `render_mode world_vertex_coords` ([`clayjohn` reminded about it right when I was opening this PR](https://github.com/godotengine/godot/issues/86812#issuecomment-1879147530)), I've only remembered about `render_mode  skip_vertex_transform` at that time. :upside_down_face:
With `render_mode world_vertex_coords` available probably most use cases are already covered. In case someone wants local coords in the shader instead of the global ones then it's possible to pass `tilemap.global_transform.affine_inverse()` as a shader uniform and use it for converting global to local. So also not a problem. And I dont' see why ever in-quadrant coords would be needed over local coords (but if needed they could also be manually calculated within the shader from the local coords).

I still think the current positions set for quadrant canvas items are strange / confusing though. But with this PR I see users still being confused, just differently. :smile:

|| v4.3.dev1.official [9d1cbab1c] | this PR |
|--------|--------|--------|
|`fract(VERTEX / 128.0)`|5A1Gir0Q6y|qNiVAoU1ca|


So in case you still disagree with this PR then please close / give a closing note in #86812 as in such case there's nothing to be done about it.

And of course thanks for the feedback!

groud:
> Oh, I meant the top left of the bounding rect of the quadrant's "first cell" (cell at quad_size * quadrant_coords map coords). AFAICT this depends only on the TileSet settings + TileMap quadrant size, such quadrant origin doesn't depend on whether/what tiles are rendered (texture size doesn't affect cell counds).

What I mean is that this value does not mean a lot. If you have a texture region size bigger than the grid, then all quadrants will end up unaligned with you textures anyway, and the origin won't be a the "correct" position.

To make it clear, my point is that there's not a lot more sense to align the origin of canvas items to the top-left corner of the cell than there is to align it to the center of the top-left cell. The center of the cell is in general consistent with the way the TileMap node deals with cells. So yeah, it might be a bit weird at first, but all offsets/positions in the tile configurations are done according to the center of the cell. So I think we should keep things consistent there.

> tilemap.global_transform.affine_inverse() as a shader uniform and use it for converting global to local

I wonder if we could provide some built-in shader variables for the TileMap node. That could be a better approach IMO.

Zireael07:
> I wonder if we could provide some built-in shader variables for the TileMap node.

This would be a good idea in general

kleonc:
> What I mean is that this value does not mean a lot. If you have a texture region size bigger than the grid, then all quadrants will end up unaligned with you textures anyway, and the origin won't be a the "correct" position.
> 
> To make it clear, my point is that there's not a lot more sense to align the origin of canvas items to the top-left corner of the cell than there is to align it to the center of the top-left cell. The center of the cell is in general consistent with the way the TileMap node deals with cells. So yeah, it might be a bit weird at first, but all offsets/positions in the tile configurations are done according to the center of the cell. So I think we should keep things consistent there.

Yeah, I agree. Pretty much the only advantage I can see of the top-left over the center is that it seems nicer for the most basic case of a rectangular grid. For any other setup there will always be something off. For such other setups the center indeed seems like a more reliable reference point. And consistency is a good point in favor of the center.

Just to be clear: currently (before this PR) it's neither the top-left corner nor the center of the cell, it's the map-coords of the cell (see https://github.com/godotengine/godot/issues/86812#issuecomment-1879145463).
Initially you've said "the original behavior is expected and preferable", now you've said the center is better. I'm not sure if you meant only "center is better than top-left" or "center is better than both top-left and current behavior".

Anyway, I agree in general center makes more sense than top-left, so I've pushed a change so it's center now in this PR (if needed it's easy to offset manually by the half cell size anyway).

> > tilemap.global_transform.affine_inverse() as a shader uniform and use it for converting global to local
> 
> I wonder if we could provide some built-in shader variables for the TileMap node. That could be a better approach IMO.

AFAICT from the rendering POV a TileMap is just an arbitrary canvas item in the middle of some canvas item parent-children hierarchy. And if I get it right you're suggesting to somehow provide some partial transforms to such specific canvas item (TileMap) to be available in a shader run for a different (descendant) canvas item (quadrant)? I have no idea if this is feasible / how would it meant to be done.

Of course I agree having built-in transforms like (quadrant to TileMap / TileMap to global / their inverses) would be nice from the user POV. But yeah, I don't see the "how". :upside_down_face:

groud:
> Just to be clear: currently (before this PR) it's neither the top-left corner nor the center of the cell, it's the map-coords of the cell (see https://github.com/godotengine/godot/issues/86812#issuecomment-1879145463).

Oh damn, you are right. Indeed that was the original intent for that code. I just focused on the "half a cell" offset but not on that. using `map_to_local` is indeed a correct fix.

> AFAICT from the rendering POV a TileMap is just an arbitrary canvas item in the middle of some canvas item parent-children hierarchy. And if I get it right you're suggesting to somehow provide some partial transforms to such specific canvas item (TileMap) to be available in a shader run for a different (descendant) canvas item (quadrant)? I have no idea if this is feasible / how would it meant to be done.

Yeah I'm not sure either. It's unclear to me what could be done, I was suggesting this as a general idea if it could help.
But yeah, I guess it would require a proposal.

Any way, in its current state the PR looks good. Approved.

groud (approved):

akien-mga:
Thanks!

PR #86848, opened 2024-01-05
Specify how CanvasTexture does not work in 3D

### LINKED ISSUE #75342: Document which texture types can be used in a Sprite3D
### Godot version

4.0

### System information

PopOS, Forward+, Nvidia 960M(525.85.05), Forward+(Vulkan)

### Issue description

Using a Canvas Texture on a Sprite3D results in a white image when it should just display the sprite.

whitesprite

### Steps to reproduce

Create Canvas Texture using a sprite sheet. Use that Canvas Texture with a Sprite3D

[test-project.zip](https://github.com/godotengine/godot/files/11070591/test-project.zip)

Mickeon:
Closes https://github.com/godotengine/godot/issues/75342

I am finding this difficult to solve in the class reference. **CanvasTexture** is the odd one among all `Texture2D`s, because it explicitly does not work outside of CanvasItems, and that is already documented.

The most that can be done is being more explicit.

Can be cherrypicked for prior versions.

### DIFF
--- a/doc/classes/CanvasTexture.xml
+++ b/doc/classes/CanvasTexture.xml
@@ -5,7 +5,7 @@
 	</brief_description>
 	<description>
 		[CanvasTexture] is an alternative to [ImageTexture] for 2D rendering. It allows using normal maps and specular maps in any node that inherits from [CanvasItem]. [CanvasTexture] also allows overriding the texture's filter and repeat mode independently of the node's properties (or the project settings).
-		[b]Note:[/b] [CanvasTexture] cannot be used in 3D rendering. For physically-based materials in 3D, use [BaseMaterial3D] instead.
+		[b]Note:[/b] [CanvasTexture] cannot be used in 3D. It will not display correctly when applied to any [VisualInstance3D], such as [Sprite3D] or [Decal]. For physically-based materials in 3D, use [BaseMaterial3D] instead.
 	</description>
 	<tutorials>
 		<link title="2D Lights and Shadows">$DOCS_URL/tutorials/2d/2d_lights_and_shadows.html</link>


doc/classes/CanvasTexture.xml
```diff
@@ -5,7 +5,7 @@
 	</brief_description>
 	<description>
 		[CanvasTexture] is an alternative to [ImageTexture] for 2D rendering. It allows using normal maps and specular maps in any node that inherits from [CanvasItem]. [CanvasTexture] also allows overriding the texture's filter and repeat mode independently of the node's properties (or the project settings).
-		[b]Note:[/b] [CanvasTexture] cannot be used in 3D rendering. For physically-based materials in 3D, use [BaseMaterial3D] instead.
+		[b]Note:[/b] [CanvasTexture] should not be used in 3D. It will not display correctly when applied to any [VisualInstance3D], such as [Sprite3D] or [Decal]. For physically-based materials in 3D, use [BaseMaterial3D] instead.
```
clayjohn, line 8:
```suggestion
		[b]Note:[/b] [CanvasTexture] cannot be used in 3D. It will not display correctly when applied to any [VisualInstance3D], such as [Sprite3D] or [Decal]. For physically-based materials in 3D, use [BaseMaterial3D] instead.
```

Looks good. I would go back to "cannot" as it is stronger. Should not implies that it might work but you shouldn't do it. In this case it just doesn't work, its the same as not applying a texture

clayjohn (approved):
Looks good!

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

PR #83409, opened 2023-10-15
[3.x] Add project settings for `AVAudioSessionCategory` on iOS

jfriv:
Adds project settings for AVAudioSessionCategory on iOS, duplicate of https://github.com/godotengine/godot/pull/81196 for 3.x

PR #81196, opened 2023-08-31
Add project settings for `AVAudioSessionCategory` on iOS

georgwacker:
Adds project settings for AVAudioSessionCategory on iOS, related to #50638 (does not fix the error log).

session_category

For the setting path I tried to stay consistent with the setting structure in display/window/ios.

PR #86850, opened 2024-01-05
[3.x] iOS: Support TTS on iOS 16 on real devices on iOS 16+

novalis:
See e.g. https://stackoverflow.com/questions/76776262/how-to-fix-av-speech-synthesizer-error-unable-to-list-voice-folder-even-after

I haven't tested TTS on 4.x as I am still finishing up my last 3.x project.  So I don't know if this patch needs to be forward-ported to `master`.  

I can say on ios 16.6.1, on a 10th generation iPad (real not simulator), this makes the difference between TTS working (that is, producing audible speech) and not (that is, not producing audible speech).

### DIFF
--- a/platform/iphone/tts_ios.h
+++ b/platform/iphone/tts_ios.h
@@ -31,6 +31,9 @@
 #ifndef TTS_IOS_H
 #define TTS_IOS_H
 
+#if __has_include(<AVFAudio/AVAudioSession.h>)
+#import <AVFAudio/AVAudioSession.h>
+#endif
 #if __has_include(<AVFAudio/AVSpeechSynthesis.h>)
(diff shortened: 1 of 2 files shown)


bruvzg:
In 4.x AV session category is configurable, in 3.x it's hard-coded to Ambient. Instead of overwriting it in the other place, it's better to back port https://github.com/godotengine/godot/pull/81196 to allow configuration, not every app require setting it to Playback.

bruvzg:
> it's better to back port https://github.com/godotengine/godot/pull/81196 to allow configuration, not every app require setting it to Playback.

There's already PR for it - https://github.com/godotengine/godot/pull/83409

novalis:
Awesome, thanks.

Issue #68293, opened 2022-11-05
RenderingDevice uses index offsets incorrectly


pkdawson:
### Godot version

v4.0.beta.mono.custom_build [8245e052c]

### System information

Windows 11, GTX 1060 (526.47)

### Issue description

`RenderingDevice::index_array_create` takes a `p_index_offset` and checks it as if it’s a number of indices, but `draw_list_bind_index_array` passes this value to `vkCmdBindIndexBuffer` which expects an offset in bytes.

https://github.com/godotengine/godot/blob/6882890a34528e056f13020798f4a473046990bb/drivers/vulkan/rendering_device_vulkan.cpp#L7754

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdBindIndexBuffer.html

However, this same value is also passed (correctly) to `vkCmdDrawIndexed` by `draw_list_draw`.

https://github.com/godotengine/godot/blob/6882890a34528e056f13020798f4a473046990bb/drivers/vulkan/rendering_device_vulkan.cpp#L7860

So Godot effectively has two conflicting implementations of the index offset, and one of those implementations is broken. My suggested solution is to set the buffer offset to 0 for `vkCmdBindIndexBuffer`.

That fixes it for me, with draw commands from Dear ImGui. Here’s the before and after:
![index_offsets](https://user-images.githubusercontent.com/132955/200126223-2cd50857-24bb-493f-8059-62319591b340.png)

### Steps to reproduce

Try drawing anything with a non-zero index offset

clayjohn:
This certainly looks like a problem that comes from always using an offset of 0. 

I agree with your proposed solution, do you mind turning it into a PR?

PR #86852, opened 2024-01-05
Fix usage of index offsets in RenderingDevice

### LINKED ISSUE #86838: RenderingDevice: Index offset incorrectly applied twice
### Tested versions

Reproducible in 4.3.dev1 and later.
Not reproducible in 4.2.1 and earlier.

### System information

Godot v4.3.dev.mono (89cc635c0) - Windows 10.0.22631 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4070 Ti (NVIDIA; 31.0.15.4629) - 13th Gen Intel(R) Core(TM) i5-13600K (20 Threads)

### Issue description

The bug described in #68293 has crept back in during the RenderingDevice refactor, affecting both Vulkan and Direct3D 12. The index array is bound with an offset, but then that offset is applied again when drawing, giving the wrong result.

https://github.com/godotengine/godot/blob/89cc635c0554cb2e518c830969ca4c5eedda0f4e/servers/rendering/rendering_device.cpp#L4260-L4262

https://github.com/godotengine/godot/blob/89cc635c0554cb2e518c830969ca4c5eedda0f4e/servers/rendering/rendering_device.cpp#L4373

The same solution (#68303) fixes this. The `p_offset` parameter in `command_render_bind_index_buffer` could just be removed, but maybe it's needed for console graphics APIs, I dunno.

### Steps to reproduce

Run the MRP, check whether there's a nice red-green-blue triangle centered in the screen.

index-bug-42

index-bug-43

[index-bug.zip](https://github.com/godotengine/godot/files/13844176/index-bug.zip)

pkdawson:
Fixes #86838

Removed duplicate usage of index offset. I tested Direct3D on Windows 11, it looks like it works the same way as Vulkan in this respect, so it also required a fix to render correctly.

### DIFF
--- a/servers/rendering/rendering_device.cpp
+++ b/servers/rendering/rendering_device.cpp
@@ -3557,10 +3557,10 @@ void RenderingDevice::draw_list_bind_index_array(DrawListID p_list, RID p_index_
 #ifdef DEBUG_ENABLED
 	dl->validation.index_array_max_index = index_array->max_index;
 #endif
-	dl->validation.index_array_size = index_array->indices;
-	dl->validation.index_array_offset = index_array->offset;
+	dl->validation.index_array_count = index_array->indices;
 
-	draw_graph.add_draw_list_bind_index_buffer(index_array->driver_id, index_array->format, index_array->offset);
+	const uint64_t offset_bytes = index_array->offset * (index_array->format == INDEX_BUFFER_FORMAT_UINT16 ? sizeof(uint16_t) : sizeof(uint32_t));
+	draw_graph.add_draw_list_bind_index_buffer(index_array->driver_id, index_array->format, offset_bytes);
 
 	if (index_array->draw_tracker != nullptr) {
 		draw_graph.add_draw_list_usage(index_array->draw_tracker, RDG::RESOURCE_USAGE_INDEX_BUFFER_READ);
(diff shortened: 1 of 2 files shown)


clayjohn (commented):
The change makes sense to me. But I would like @RandomShaper to confirm that its fine to merge

RandomShaper:
I acknowledge the issue boils down to the offset being applied twice. However, I'd like to state a different take on the solution.

To me, the drivers are doing the right thing. I mean, at their level of abstraction, they have no knowledge about what the offset is for, so they must just apply it. The problem lies at the `RenderingDevice` level, because that's the one doing the double-application. It binds with offset and then asks to render with the offset again. My suggested fix would be then to stop `RenderingDevice` from re-applying the offset at draw time.

This (untested) patch is what I'd do: 
[p.txt](https://github.com/godotengine/godot/files/13859423/p.txt)

pkdawson:
As described previously in #68293, `RenderingDevice::index_array_create` takes the offset as a number of indices, but the drivers want the offset in bytes, so that needed an adjustment. Confirmed that this solution works too.

RandomShaper (approved):

clayjohn (approved):
Looks great! Sorry for breaking this twice now :grimacing:

akien-mga:
This has conflicts following the merge of #84976, could you rebase (and I would suggest re-testing, as #84976 changed things significantly).

pkdawson:
Rebased. Confirmed that this works with my MRP and my more complex test case of Dear ImGui meshes (https://github.com/pkdawson/imgui-godot), with no other obvious bugs.

akien-mga:
Thanks!

PR #86853, opened 2024-01-05
Update uk.po

somerandomdough:
взули -> вузли (nodes)
Explanation: misspelling
Fix: switch positions of two letters


AThousandShips:

However this is not where these files are edited, instead you edit them on the webplate, see [here](https://docs.godotengine.org/en/stable/contributing/documentation/editor_and_docs_localization.html)

Closing as this is not a valid change

PR #86858, opened 2024-01-06
Include time delta in `SceneTree` signals `process_frame` and `physics_frame`

Rubonnek:
Unsure if this is wanted upstream but I find this patch particularly useful with non-`Node` objects.

In particular I've implemented a `RefCounted` ringbuffer that keeps a series of screenshots of the last five seconds just so that I can save those in case something interesting happens. With this patch I can avoid implementing leaky abstraction to get the delta through `Node.get_process_delta_time()` or `Node.get_physics_process_delta_time()` or passing the ringbuffer to a `Node` and have it pass the delta within its `_process` or `_physics_process` call. I'm aware the ringbuffer could be turned into a `Node` but it doesn't need all that functionality.

### DIFF
--- a/doc/classes/SceneTree.xml
+++ b/doc/classes/SceneTree.xml
@@ -297,11 +297,13 @@
 			</description>
 		</signal>
 		<signal name="physics_frame">
+			<param index="0" name="delta" type="float" />
 			<description>
 				Emitted immediately before [method Node._physics_process] is called on every node in the [SceneTree].
 			</description>
 		</signal>
 		<signal name="process_frame">
+			<param index="0" name="delta" type="float" />
 			<description>
 				Emitted immediately before [method Node._process] is called on every node in the [SceneTree].
 			</description>
--- a/scene/main/scene_tree.cpp
+++ b/scene/main/scene_tree.cpp
@@ -461,7 +461,7 @@ bool SceneTree::physics_process(double p_time) {
 	}
 	physics_process_time = p_time;
 
-	emit_signal(SNAME("physics_frame"));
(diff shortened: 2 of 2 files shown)


AThousandShips:
I believe adding new arguments to signals breaks compatibility, as signals won't ignore them and will fail to call:
* https://github.com/godotengine/godot-proposals/issues/7258

Rubonnek:
> I believe adding new arguments to signals breaks compatibility, as signals won't ignore them and will fail to call:
> 
>     * [Discard additional arguments in a connected signal's callback function godot-proposals#7258](https://github.com/godotengine/godot-proposals/issues/7258)

Thanks for the info -- I didn't think this patch would break compatibility. I just realized this issue and that one also have workarounds using decorators. I'll leave a comment there in a couple of days since I'm short of time today.

I don't know why I thought an `autoload` wasn't accessible in an `Object._init()` -- that was why I opened this PR and I didn't think to test that.

Turns out adding a `Node._process`-like function to any `Object` within its `_init()` function is possible with any `autoload` -- the callback just needs to be decorated using a signal connection helper similar to the following to add the `delta` parameter:

```gdscript
func subscribe_to_process(p_callable : Callable) -> void:
	var decorator : Callable = Callable(self, &"__decorate_process_frame")
	decorator = decorator.bind(p_callable)
	var _ignore : int = MyAutoload.get_tree().connect(&"process_frame", decorator)

func __decorate_process_frame(p_callable : Callable) -> void:
	var delta : float = MyAutoload.get_process_delta_time()
	p_callable.call(delta)
```

The current functionality is enough for my needs. This PR is not needed.

PR #86859, opened 2024-01-06
Add input signals to `SceneTree`

Rubonnek:
Unsure if this is wanted upstream, but I find it useful with non-`Node` objects.

In particular I wanted to implement a Konami code detector but the input was not propagating to my target node due to calling `Viewport.set_input_as_handled()` somewhere else. Figured it was best to get the `InputEvent` and inspect it before any `Node` input calls were made.

This patch also makes it easier to implement the Konami-code detector as a `RefCounted` object since now it only needs a reference to the SceneTree and there's no need to couple its code with a `Node`.

### DIFF
--- a/doc/classes/SceneTree.xml
+++ b/doc/classes/SceneTree.xml
@@ -272,6 +272,12 @@
 		</member>
 	</members>
 	<signals>
+		<signal name="input">
+			<param index="0" name="event" type="InputEvent" />
+			<description>
+				Emitted immediately before [method Node._input] is called on every node in the [SceneTree].
+			</description>
+		</signal>
 		<signal name="node_added">
 			<param index="0" name="node" type="Node" />
 			<description>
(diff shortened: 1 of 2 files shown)


EIREXE:
This is also useful for reading ANY input, regardless of what happens or who consumes it, which could be useful for integrations with UI frameworks like imgui or for input systems that change the glyphs when a button press is detected.

Sauermann:
Take the following scene tree
```
- Root Window
  - SubViewportContainer
    - SubViewport
```
and lets assume, that no node sets the input events as handled.

With the current changes of the PR the scene tree will emit each of the new signals twice.
- first when the Input Event is handled in the Root window
- second when the Input Event is handled in the SubViewport

I believe that a better approach would be to emit each signal only once.

Mickeon:
I know it sounds useful, and I'm sure it is, but it's an awful lot of signals to be emitted, potentially...

Sauermann:
> In particular I wanted to implement a Konami code detector but the input was not propagating to my target node due to calling `Viewport.set_input_as_handled()` somewhere else

In order to access the InputEvent guaranteed without interference from any `set_input_as_handled()` calls, you can listen to the signal `window_input` in the root window. So this is currently already possible.

Rubonnek:
> So this is currently already possible.

I missed that! Thank you!

PR #81628, opened 2023-09-13
GDScript: Fix lambda hot reloading

rune-scape:
fixes https://github.com/godotengine/godot/issues/62769

swaps out the function pointers used in LambdaCallables when a script is reloaded
if it cant find a good replacement it invalidates the callable, so no more 'Bad Address Index' and crashes, it tells you its called an invalid lambda
the logic it uses to swap lambdas is very simple right now, swapping only if the argument count vaguely matches, and no lambdas were added or removed, evaluated for each function the lambda was in
also should be safe across threads, because lambdas can be created on any thread. i took care to make sure it wouldnt slow lambda creation down

it may be a good idea to add some more logic to how old lamdas are matched to new ones, the data is there but i dont know what would be intuitive, what would break logic, im all ears for heuristics

**MRP:**
[81628.zip](https://github.com/godotengine/godot/files/12830903/81628.zip)

_Production edit: added the MRP_

PR #86860, opened 2024-01-06
[4.2] GDScript: Lambda hotswap fixes

rune-scape:
cherrypicked https://github.com/godotengine/godot/pull/86569 for 4.2 :)

### DIFF
--- a/modules/gdscript/gdscript.cpp
+++ b/modules/gdscript/gdscript.cpp
@@ -1391,51 +1391,43 @@ String GDScript::debug_get_script_name(const Ref<Script> &p_script) {
 }
 #endif
 
-GDScript::UpdatableFuncPtr GDScript::func_ptrs_to_update_main_thread;
-thread_local GDScript::UpdatableFuncPtr *GDScript::func_ptrs_to_update_thread_local = nullptr;
-
-GDScript::UpdatableFuncPtrElement GDScript::_add_func_ptr_to_update(GDScriptFunction **p_func_ptr_ptr) {
-	UpdatableFuncPtrElement result = {};
-
-	{
-		MutexLock lock(func_ptrs_to_update_thread_local->mutex);
-		result.element = func_ptrs_to_update_thread_local->ptrs.push_back(p_func_ptr_ptr);
-		result.func_ptr = func_ptrs_to_update_thread_local;
-
-		if (likely(func_ptrs_to_update_thread_local->initialized)) {
-			return result;
-		}
-
-		func_ptrs_to_update_thread_local->initialized = true;
+GDScript::UpdatableFuncPtr::UpdatableFuncPtr(GDScriptFunction *p_function) {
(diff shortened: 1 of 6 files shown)


AThousandShips (dismissed):

AThousandShips:
Normally when something is easily done with `git cherry-pick` we leave it to the production team to handle, unsure if this was trivial

Will need a decision from the gdscript team and see if it doesn't cause any unforeseen issues before merging but code looks good

rune-scape:
ah 👍 ya it was trivial, not much of the surrounding code changed ,  didn't know 👍

AThousandShips:
Then future suggestion would be to ask on the PR if it can be cheery picked

akien-mga:
@rune-scape Could you help us evaluate how risky this PR is, i.e. does it have potential to introduce regressions and are the issues it fixes worth it?

For stable branches we're usually quite conservative with what we cherry-pick, and this code specifically had a number of critical regressions in the phase leading to 4.2-stable. It's not clear whether users are affected by the issues #86569 fixes in the wild (no bug report), so if it has potential to instead break some functionality, I'd opt for not cherry-picking the PR (and let it get tested thoroughly in the 4.3 dev phase).

akien-mga:
CC @adamscott @RandomShaper

RandomShaper:
Given that I won't be very aware of how this turns out in the wild, all I can say is I agree with not cherry-picking at least until 4.3 has been thoroughly tested and also it's maybe not worth the risk if there are no reported issues in 4.2, anyway. I wouldn't oppose, though, regardless no reports, but, yes, let's see if the current approach crystalizes into stability in the wild.

rune-scape:
technically not a critical bug, but the way that https://github.com/godotengine/godot/pull/81628 works is fundementally broken, the the PRs that followed didnt fix the fact that it invalidates basically all lambdas unless theres only 1 script with lambdas, making it safer than nothing ig
i can make an issue demonstrating this for 4.2,
the original code is convoluted and has 2 mutexes to lock in order to add a func ptr, which could deadlock if not done properly,,, thats what most of the following PRs helped with, but they also made the code even harder to follow and inspect
this code only uses 1 mutex very simply, i think its much safer in that regard

akien-mga:
The thing is, if it's breaking hot reloading, it's not a big deal, it's just an editor time inconvenience. What it addressed however was runtime crashes which we're critical. I want to make sure we don't re-introduce runtime issues for the sake of improving editor lambda editing usability.

rune-scape:
ok i understand, was giving my input and maybe was distracted from how important its stability is,
i do think the way this code is written is much safer than what is in 4.2 fwiw, but not tested nearly as much

akien-mga:
Let's give it a go then and see what happens with 4.2.2 RC 1 tests.

akien-mga (approved):

akien-mga:
Thanks!

PR #79185, opened 2023-07-08
Disable registering 3D physics types when 3D is disabled

aaronfranke:
This PR organizes the 2D and 3D physics code in `register_server_types.cpp` and allows the 3D physics type registration code to be disabled by wrapping it in `#ifndef _3D_DISABLED`.

In the current master, a macOS arm64 template_release build is 66.5 MB (66503407 bytes). Using `disable_3d=yes` brings this down to 60.2 MB (60224255 bytes), a reduction of 9.4%.

With this PR, the `disable_3d=yes` build is 58.8 MB (58848751 bytes), a reduction of 2.3% compared to master with this option, and brings the total savings from 9.4% to 11.5% compared to 3D enabled, a reduction of 2.1%.

PR #86874, opened 2024-01-06
Fix some build errors with `disable_3d=yes`

AThousandShips:
* Some tests were incorrectly included
* SCU builds with animation
* Animation `switch`

I'm unsure if this is the best way with the scu, but it solves the issue, could also wrap the code in `#ifndef` but picked this way for right now, can switch if desired

I'm not sure why the linking fails with the SCU builds *if* the file is along with the others, so worth investigating separately

Edit: Swapped to the plain `#ifndef` version, felt more guaranteed

~The issues with the SCU is link errors, which *might* be MSVC specific, haven't confirmed yet~
Not MSVC specific, happens on mingw as well


scene/animation/animation_mixer.cpp
AThousandShips, line 1: This parts errors regardless, a `switch` with only a `default`

scene/animation/animation_mixer.cpp
```diff
@@ -954,15 +954,15 @@ Variant AnimationMixer::post_process_key_value(const Ref<Animation> &p_anim, int
...
 				const Skeleton3D *skel = Object::cast_to<Skeleton3D>(p_object);
 				return Vector3(p_value) * skel->get_motion_scale();
 			}
 			return p_value;
-		} break;
 #endif // _3D_DISABLED
+		} break;
 		default: {
 		} break;
 	}
```
aaronfranke, line 968: Considering we are using only one of the track type values, this function might make more sense if we just rewrite it to not use a switch statement.

```cpp
Variant AnimationMixer::_post_process_key_value(const Ref<Animation> &p_anim, int p_track, Variant p_value, const Object *p_object, int p_object_idx) {
#ifndef _3D_DISABLED
	if (p_anim->track_get_type(p_track) == Animation::TYPE_POSITION_3D && p_object_idx >= 0) {
		const Skeleton3D *skel = Object::cast_to<Skeleton3D>(p_object);
		return Vector3(p_value) * skel->get_motion_scale();
	}
#endif // _3D_DISABLED
	return p_value;
}
```

aaronfranke (commented):
See also #79185

aaronfranke (approved):
Looks good to me.

akien-mga (approved):

akien-mga:
Thanks!

AThousandShips:
Thank you!

YuriSizov:
Cherry-picked for 4.2.2.

PR #82193, opened 2023-09-23
Add repeat option to `SpriteBase3D`

imaskillissue:
Fixes #82044

Added to SpiteBase3D to Flags Repeat.

PR #86875, opened 2024-01-06
Fix texture artifacts around Sprite3D with linear filtering

### LINKED ISSUE #13044: Expose texture_repeat on SpriteBase3D

A 3D game utilizing Sprite3Ds for a 2D style. Not pixel art, therefore the usage of linear filtering is preferred.

On 3d Sprites the texture repeat seems to be enabled by default, with no direct way to edit it. This means you can't really use textures with linear filtering like you can with 2d sprites, as artifacts can appear at the edges from the repeating.

being able to manually change the texture repeat would bring 3d sprites on par with their 2d counterpart. And this similarity is the main reason to use them over a meshinstance3d.

The property would have to be added to the abstract SpriteBase3D-class.
Internally Sprite3D seems to call a method called `BaseMaterial3D.get_material_for_2d()` to generate a material for the given parameters and flags. this method would also need to include an argument for texture repeat, setting the flag for it along with the other ones.

the documentation suggests using a texture with transparent padding at the edges to prevent the artifacts from occurring with linear filtering. But that would require adding the padding to every texture, so it isn't a really easy workaround. The other option would be to use a Material and disable repeat there. But using a material on a Sprite3d would also overwrite all other settings of the node, making it not really different from a MeshInstance3d.
And the whole point of Sprite3d is that it doesn't require adding a material and allows easy change of each parameter without affecting other instances that use the same material.

Given that it concerns the inner workings of a Node and its usage of a method that can only be used internally, an Addon isn't really an option.

### LINKED ISSUE #73739: Gradients on a sprite 3d have artifacts.
### Godot version

4.0.rc2 EDIT: I have tested, and it appears on 4.0.rc3 as well.

### System information

Windows 10, NVIDIA GeForce GTX 1650 SUPER

### Issue description

the problem
Sprite 3Ds with a gradient texture assigned to them have a line on the edge that should not be present. 

This still occurs even if using a Quad mesh with a gradient texture.

### Steps to reproduce

1. Spawn in a Sprite3d.
2. Give it a GradientTexture2D (A GradientTexture1D works too.)
3. Observe the artifact on the edge.

If you want to do this with a Quad instead:
1. Spawn in a MeshInstance3D
2. Give it a QuadMesh.
3. Give the QuadMesh a StandardMaterial3D
4. Set the material's texture (under albedo) to a GradientTexture2D (A GradientTexture1D works too.)
5. (Optional) Set the material's Shading_Mode to unshaded to better see the artifacts.

Included in the project below is a Camera3D set to view a Sprite3D with a GradientTexture2D assigned to it, and a Quad with an unshaded material with a GradientTexture2D assigned to it.

[GradientArtifactsProject.zip](https://github.com/godotengine/godot/files/10800764/GradientArtifactsProject.zip)

### LINKED ISSUE #82044: Sprite3D texture wrapping causing pixel lines on border
### Godot version

### System information

Godot v4.1.1.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2080 Ti (NVIDIA; 31.0.15.3734) - Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz (8 Threads)

### Issue description

Sprite3D has pixel lines which appear to be the texture wrapping. There are no options on Sprite3D to disable wrapping.

image

Some troubleshooting for this issue has been carried out in my [Reddit thread.](https://old.reddit.com/r/godot/comments/16nobpu/new_to_godot_and_loving_it_any_idea_how_i_can/)

Tested on 4.1.1 and 4.2dev5

### Steps to reproduce

1. Create Sprite3D node
2. Import texture with lossless compression (or any compression, nothing I tried made a difference)
3. Apply texture to Sprite3D

[Sprite3D_bug_21-09-23.zip](https://github.com/godotengine/godot/files/12684098/Sprite3D_bug_21-09-23.zip)

Mickeon:
Closes https://github.com/godotengine/godot/issues/73739
Closes https://github.com/godotengine/godot-proposals/issues/13044

This PR fully fixes the original issue described, by disabling the texture repeat on the material generated by **Sprite3D**, preventing the color to leak around the edges outright.

| Before | After
| --- | --- |
| image| image

This issue also exists in **Label3D**, but you'd never ever see it in practice, as it resizes with a bit of padding around the edges to compensate.

_Bugsquad edit: Fixes https://github.com/godotengine/godot/issues/82044_

### DIFF
--- a/scene/resources/material.cpp
+++ b/scene/resources/material.cpp
@@ -2406,6 +2406,7 @@ Ref<Material> BaseMaterial3D::get_material_for_2d(bool p_shaded, Transparency p_
 	material->set_flag(FLAG_ALBEDO_TEXTURE_MSDF, p_msdf);
 	material->set_flag(FLAG_DISABLE_DEPTH_TEST, p_no_depth);
 	material->set_flag(FLAG_FIXED_SIZE, p_fixed_size);
+	material->set_flag(FLAG_USE_TEXTURE_REPEAT, false); // Undesired.
 	material->set_alpha_antialiasing(p_alpha_antialiasing_mode);
 	material->set_texture_filter(p_filter);
 	if (p_billboard || p_billboard_y) {


clayjohn:
Supersedes https://github.com/godotengine/godot/pull/82193/files

Mickeon:
Interesting. This PR started with that very same new bool property, but then I realised how useless it would be in practice and scaled it back to always be `false`.

clayjohn (commented):
Some more context in https://github.com/godotengine/godot/issues/82044

In particular it sounds like users sometimes use the region property of the sprite to make it tile within the Sprite3D. So we probably need to have an option to enable repeat like what was added in https://github.com/godotengine/godot/pull/82193

Mickeon:
No, I do believe it. In hindsight it makes sense and can actually be used that way. My assumption was wrong.

Regardless of the way it is done, I'm not sure about it being the "default" as suggested. The repeating texture repeating issues around the border is not obvious.

clayjohn:
> No, I do believe it. In hindsight it makes sense and can actually be used that way. My assumption was wrong.
> 
> Regardless of the way it is done, I'm not sure about it being the "default" as suggested. The repeating texture repeating issues around the border is not obvious.

I agree. I think repeat could be an option, but shouldn't be the default

398utubzyt:
@clayjohn Is this PR intended to replace #82193? It seems like #82193 already provides an option to enable/disable sprite texture wrapping as discussed previously.

imaskillissue:
> > No, I do believe it. In hindsight it makes sense and can actually be used that way. My assumption was wrong.
> > 
> > Regardless of the way it is done, I'm not sure about it being the "default" as suggested. The repeating texture repeating issues around the border is not obvious.
> 
> I agree. I think repeat could be an option, but shouldn't be the default

In my Pull Request(#82193) it's already by default disabled. I thought its most of the cases  a bug then a feature and people who wanna use it most likely find how to enable it.

anniryynanen:
Is there something others (me, maybe) can do to help this PR move forward?

akien-mga:
> Is there something others (me, maybe) can do to help this PR move forward?

It needs a decision from the rendering team on whether this approach is preferred, or exposing the repeat option like done in #82193 or #90202. So one thing you can do is check all three PRs and bring them up for decision making in a rendering team meeting.

clayjohn:
> > Is there something others (me, maybe) can do to help this PR move forward?
> 
> It needs a decision from the rendering team on whether this approach is preferred, or exposing the repeat option like done in #82193 or #90202. So one thing you can do is check all three PRs and bring them up for decision making in a rendering team meeting.

The rendering team discussed this in March and explained how it should be implemented here: https://github.com/godotengine/godot/pull/82193#issuecomment-1996597424

StrayEddy (approved):
The FLAG_USE_TEXTURE_REPEAT is put to false, which would fix the issue

Mickeon:
Please read the above comment and [associated link](https://github.com/godotengine/godot/pull/82193#issuecomment-1996597424)

Calinou:
Is this superseded by https://github.com/godotengine/godot/pull/98122?

clayjohn:
Superseded by https://github.com/godotengine/godot/pull/98122

PR #86878, opened 2024-01-06
Parse the names of children of `TabContainer`s on POT generation

YeldhamDev:
`TabContainer` makes use of its children's names to title its tabs, so those should be translated as well. If a child has `auto_translate` disabled, it will be skipped, as well as all children if the `TabContainer` itself disabled it.


KoBeWi:
This can generate false positives if `set_tab_title()` is used.

YeldhamDev:
@KoBeWi The the user can toggle `auto_translate` off in the respective node (or the container) itself, as explained above.

KoBeWi (approved):

akien-mga:
Thanks!

Issue #86883, opened 2024-01-06
Comment node missing in Visual Shaders


kubaofc123:
### Tested versions

- Reproducible in: 4.2.1.stable - Not reproducible in 4.1.1.stable

### System information

Godot v4.2.1.stable - Windows 10.0.19044 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3080 (NVIDIA; 31.0.15.3129) - AMD Ryzen 9 5950X 16-Core Processor (32 Threads)

### Issue description

In Godot 4.2, the comment node in visual shaders is missing and cannot be added to graph, class VisualShaderNodeComment. Class still exists but is not usable
Below are images of version 4.1.1 and 4.2.1.

![image](https://github.com/godotengine/godot/assets/17231482/caa7497a-ea15-4ae6-bbf6-2892438b30c6)
![image](https://github.com/godotengine/godot/assets/17231482/b36f2784-f98a-40ed-9d5c-93c0cc6199a8)

### Steps to reproduce

Open Visual Shader graph and try adding node "Comment"

### Minimal reproduction project (MRP)

N/A

jsjtxietian:
See https://github.com/godotengine/godot/pull/79307

kubaofc123:
Understood, closing down

PR #79307, opened 2023-07-10
Remove GraphNode's comment property and related functionality

Geometror:
(this functionality will be reintroduced in a new and more flexible way)

Part of the graph editor refactoring process.
Comment nodes in the current form aren't really working well, especially with several of them on top of each other.
Improved, special Nodes for organizing the graph will be reintroduced at a later time. 
(There are already some experiments, but a satisfying solution hasn't been found yet)

PR #86890, opened 2024-01-06
Update doc to mention that visual comment node is currently not visible

jsjtxietian:
Related to https://github.com/godotengine/godot/issues/86883

It looks like many users are won't aware of the rework and due to the fact that the replacement is not ready yet, IMO we can at least update the doc now as suggested in https://github.com/godotengine/godot/pull/79307#issuecomment-1867647049

The doc itself maybe not needed in 4.3 as the rework might be ready in 4.3, but can be cherry picked for 4.2 at least now.


Geometror:
Comment nodes were removed completely and unfortunately won't be restored. 
A replacement was added with #88014 that should solve all the problems we had with the old comment nodes, but it works a bit differently. I think this PR can be closed.

Mickeon:
I feel like this is not necessary anymore, even for 4.2, as the original issues have been closed. So I'm closing this PR, but thank you for your contribution.

If you feel like this is still necessary for a potential 4.2 version, open a PR with a different note, as the current one is not exactly fitting now.

PR #86891, opened 2024-01-06
Add autocompletion to `get_node_or_null`

Mickeon:
Someone may have forgot to, years after. Oops! Either that, or they thought "Well, the Node not always guaranteed to be there, so we may as well not autocomplete the results" which, I will sincerely disregard.


AThousandShips (approved):
LGTM

Mickeon:
Can be backported to prior 4.x versions

AThousandShips:
I'll leave that decision to the production team as it's an enhancement 🙂, I default to only add that to bug fixes and documentation as it's a bigger decision than I am comfortable taking responsibility for

akien-mga:
Thanks!

Issue #86869, opened 2024-01-06
Calling `EditorInterface.open_scene_from_path()` in a plugin's `_enter_tree()/ready()` functions crashes Godot


razcore-rad:
### Tested versions

- Reproducible in: 4.1.2.stable.official.399c9dc39, 4.1.3.stable.official.f06b6836a, 4.2.stable.official.46dc27791, 4.2.1.stable.official.b09f793f5, 4.3.dev1.official.9d1cbab1c

### System information

Ubuntu 23.10, Vulkan, NVIDIA GeForce RTX™ 2060 / NVIDIA GeForce RTX™ 2060

### Issue description

Calling `get_editor_interface().open_scene_from_path("res://dbg1.tscn")` in a plugin's `_enter_tree()`/`_ready()` function when previously saved the project with a different scene will lead to a crash, e.g.:

```sh
❯ ~/Downloads/Godot_v4.1.2-stable_linux.x86_64 -e
Godot Engine v4.1.2.stable.official.399c9dc39 - https://godotengine.org
Vulkan API 1.3.242 - Forward+ - Using Vulkan Device #0: NVIDIA - NVIDIA GeForce RTX 2060
 

================================================================
handle_crash: Program crashed with signal 11
Engine version: Godot Engine v4.1.2.stable.official (399c9dc393f6f84c0b4e4d4117906c70c048ecf2)
Dumping the backtrace. Please include this when reporting the bug to the project developer.
[1] /lib/x86_64-linux-gnu/libc.so.6(+0x42910) [0x7fabc7042910] (??:0)
[2] /home/razcore-rad/Downloads/Godot_v4.1.2-stable_linux.x86_64() [0x1df00fd] (??:0)
[3] /home/razcore-rad/Downloads/Godot_v4.1.2-stable_linux.x86_64() [0x4ecb862] (??:0)
[4] /home/razcore-rad/Downloads/Godot_v4.1.2-stable_linux.x86_64() [0x12d7b3e] (??:0)
[5] /home/razcore-rad/Downloads/Godot_v4.1.2-stable_linux.x86_64() [0x129a148] (??:0)
[6] /home/razcore-rad/Downloads/Godot_v4.1.2-stable_linux.x86_64() [0x118df68] (??:0)
[7] /home/razcore-rad/Downloads/Godot_v4.1.2-stable_linux.x86_64() [0x2ce47c1] (??:0)
[...]

PR #86892, opened 2024-01-06
Add nullcheck for `editor_plugin_screen` to prevent crash on startup

jsjtxietian:
Prevents the crash in https://github.com/godotengine/godot/issues/86869

From the crash call stack, the problem looks like `open_scene_from_path` in `_enter_tree` is called to early, before `editor_plugin_screen` is inited and thus the crash. There maybe other ways of preventing this crash (something like delay any call on the call chain ?),  but I think at least add some null check is harmless.


KoBeWi:
I can't reproduce the issue anymore. Is this still relevant?

PR #86894, opened 2024-01-06
In `merge_with` also merge some default fields of theme

### LINKED ISSUE #86857: Theme.merge_with does not seem to merge the default font size
### Tested versions

Tested on v4.2.stable.official [46dc27791]

### System information

Godot v4.2.stable - Windows 10.0.19045 - Vulkan (Mobile) - dedicated NVIDIA GeForce GTX 1060 (NVIDIA; 31.0.15.4633) - Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz (8 Threads)

### Issue description

The `theme.merge_with(other)` [method ](https://docs.godotengine.org/en/stable/classes/class_theme.html#class-theme-method-merge-with)does not seem to merge the default font size from `other` into `theme`.  Not too familiar with the source code for Godot, but reading the [merge_with C++ code](https://github.com/godotengine/godot/blob/89cc635c0554cb2e518c830969ca4c5eedda0f4e/scene/resources/theme.cpp#L1548) it looks like it isn't copying over default properties (and maybe other 'top-level' and default properties if there are any)?

### Steps to reproduce

1. Open attached minimal reproduction project then run it.
2. Notice that the theme that has been merged with didn't get its default font size updated

Snippet of code:
```
	var other_theme := preload("res://other_theme.tres")
	print("Before merging, default font size is %d and font color is %s" % [theme.default_font_size, theme.get_color("font_color", "Label").to_html()])
	print("Other theme's default font size is %d and font color is %s" % [other_theme.default_font_size, other_theme.get_color("font_color", "Label").to_html()])
	theme.merge_with(other_theme)
	print("After merging, default font size is %d and font color is %s" % [theme.default_font_size, theme.get_color("font_color", "Label").to_html()])
```
After merging, the Label's font color was correctly updated but the default font size was not.

[MergeWithBug.zip](https://github.com/godotengine/godot/files/13847461/MergeWithBug.zip)

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/86857

I'm not sure whether `merge_with`  should take care of these default fields, feel free to correct me.

### DIFF
--- a/scene/resources/theme.cpp
+++ b/scene/resources/theme.cpp
@@ -1613,6 +1613,17 @@ void Theme::merge_with(const Ref<Theme> &p_other) {
 		}
 	}
 
+	// Defaults.
+	if (p_other->has_default_font()) {
+		set_default_font(p_other->default_font);
+	}
+	if (p_other->has_default_font_size()) {
+		set_default_font_size(p_other->default_font_size);
+	}
+	if (p_other->has_default_base_scale()) {
+		set_default_base_scale(p_other->default_base_scale);
+	}
+
 	_unfreeze_and_propagate_changes();
 }
 


scene/resources/theme.cpp
```diff
@@ -1613,6 +1613,11 @@ void Theme::merge_with(const Ref<Theme> &p_other) {
 		}
 	}
 
+	// Defaults.
+	set_default_font(p_other->default_font);
```
AThousandShips, line 1617: This should probably be optional, at minimum it should only assign if the other has these set, so:
```suggestion
	if (p_other->has_default_font()) {
		set_default_font(p_other->default_font);
	}
```
And similar for the rest

jsjtxietian: Yes you are right, thanks !

YuriSizov (approved):
Yeah, that makes sense.

akien-mga:
Thanks!

PR #86907, opened 2024-01-07
Change AudioStreamPlayer autoplay and GLTFBufferView getters to be const

aaronfranke:
I wanted to use these functions with `const AudioStreamPlayer *p_` etc parameters, but I found that these methods were not marked as const, so I could not use them. This PR fixes this and adds compatibility methods.

### DIFF
--- a/modules/gltf/structures/gltf_buffer_view.compat.inc
+++ b/modules/gltf/structures/gltf_buffer_view.compat.inc
@@ -1,5 +1,5 @@
 /**************************************************************************/
-/*  gltf_buffer_view.h                                                    */
+/*  gltf_buffer_view.compat.inc                                           */
 /**************************************************************************/
 /*                         This file is part of:                          */
 /*                             GODOT ENGINE                               */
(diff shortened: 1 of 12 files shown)


modules/gltf/structures/gltf_buffer_view.compat.inc
```diff
@@ -0,0 +1,61 @@
...
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef DISABLE_DEPRECATED
+
+GLTFBufferIndex GLTFBufferView::_get_buffer_bind_compat_86907() {
+	return buffer;
```
AThousandShips, line 34: Normally we call the original method to make sure there's no issues with the changes in that method, I'd say that would be appropriate here too as it's valid to call
```suggestion
	return get_buffer();
```

aaronfranke: Makes sense, and makes it very clear that this is a trivial binding-only method with no special behavior.

fire (approved):
The theory is that add const breaks nothing. So it's ok.

akien-mga:
Thanks!

PR #85870, opened 2023-12-06
Fix overriding `CollisionObject3D::_mouse_enter()` and `_mouse_exit()` from GDExtension

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1325

Currently, this code is explicitly calling these virtual methods on a script. This needs to be switched to `GDVIRTUAL_CALL()` in order to work for both scripts and GDExtension.

PR #86908, opened 2024-01-07
Fix virtual calls for GDExtension in `CollisionObject2D`

AdSkipper1337:
Hello, this is my first PR for Godot engine, 

if I am doing something wrong please kindly point me in the right direction.
This fix is similar in nature to https://github.com/godotengine/godot/pull/85870
Same bug, same fix but 2D instead of 3D respectively. I will comment on the issue of that bug since its so similar and link the PR there.
Hopefully that is in line with the way you usually handle a case like this arround here, if you have a diffferent process with making a separate bug report first, let me know!

### DIFF
--- a/scene/2d/collision_object_2d.cpp
+++ b/scene/2d/collision_object_2d.cpp
@@ -523,30 +523,22 @@ void CollisionObject2D::_input_event_call(Viewport *p_viewport, const Ref<InputE
 }
 
 void CollisionObject2D::_mouse_enter() {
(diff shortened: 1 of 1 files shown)


rburing (approved):

rburing:
Thanks for the PR! This looks good to me.

akien-mga:

akien-mga:
Cherry-picked for 4.2.2.

PR #86920, opened 2024-01-07
Fix build warning with memset value being too large

### LINKED ISSUE #83342: GLES3: Texture's memset exceeds maximum object size
### Godot version

4.2.beta [a574c02]

### System information

Arch Linux 6.5.6-zen2-1-zen (64-bit)

### Issue description

When the compiler tries to compiles `drivers/gles3/texture_storage.cpp`, it only gets one warning saying that `memset` specified bound will be exceeds maximum object size

### Steps to reproduce

1. Clone (if you haven't)
2. Compile it
3. Look out for something like `warning: 'void* memset(void*, int, size_t)' specified bound between 18446744065119617024 and 18446744073709551612 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]`
3.5. If isn't then, try different `target` like `target=template_release` and look again, repeat until either it's there or somehow gone.

N/A

griffinkh:
Should resolve issue #83342

* *Bugsquad edit, closes: #83342*

### DIFF
--- a/drivers/gles3/storage/texture_storage.cpp
+++ b/drivers/gles3/storage/texture_storage.cpp
@@ -1475,7 +1475,7 @@ void TextureStorage::update_texture_atlas() {
 		//generate atlas
 		Vector<TextureAtlas::SortItem> itemsv;
 		itemsv.resize(texture_atlas.textures.size());
-		int base_size = 8;
+		uint32_t base_size = 8;
 
 		int idx = 0;
 
@@ -1488,7 +1488,7 @@ void TextureStorage::update_texture_atlas() {
 			si.size.height = (src_tex->height / border) + 1;
 			si.pixel_size = Size2i(src_tex->width, src_tex->height);
 
-			if (base_size < si.size.width) {
+			if (base_size < (uint32_t)si.size.width) {
 				base_size = nearest_power_of_2_templated(si.size.width);
 			}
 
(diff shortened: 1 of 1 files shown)


drivers/gles3/storage/texture_storage.cpp
```diff
@@ -1475,7 +1475,7 @@ void TextureStorage::update_texture_atlas() {
 		//generate atlas
 		Vector<TextureAtlas::SortItem> itemsv;
 		itemsv.resize(texture_atlas.textures.size());
-		int base_size = 8;
+		unsigned int base_size = 8;
```
clayjohn, line 1478: We use `uint32_t` instead of `unsigned int`
```suggestion
		uint32_t base_size = 8;
```

clayjohn (changes requested):
Instead of using `unsigned int` this should use `uint32_t`, otherwise this seems great!

drivers/gles3/storage/texture_storage.cpp
griffinkh: thanks! fixed

griffinkh:
change is made, thanks!

clayjohn (approved):
Looks great! Thanks for the quick update

akien-mga:

akien-mga:
Cherry-picked for `4.2`, `4.1`, and `4.0`.

PR #86923, opened 2024-01-07
Allow disabling `TileMapLayer` by `TileMap` process mode

### LINKED ISSUE #86527: TileMap Physics layer won't disable when process set to disable while every other physics object will
### Tested versions

v4.2.1.stable.flathub [b09f793f5]

### System information

Godot v4.2.1.stable (b09f793f5) - Freedesktop SDK 23.08 (Flatpak runtime) - X11 - GLES3 (Compatibility) - NVIDIA GeForce RTX 3070 Ti (nvidia) - AMD Ryzen 5 5600X 6-Core Processor (12 Threads)

### Issue description

If we put a RayCast2D and make it intersect with any disabled CollisionObject2D, it won't detect it;
However, a disabled TileMap WILL be detected by Area2d and RayCast2d.

(The rest is just rant, I know it's not appropriate but I'm so upset)
I've been trying to work around this for a month now, I've wasted weeks on converting TileMap data into arrays and painting the map on codes with more failure.
It used to work with simple hiding and showing, and apparently it was a bug, after an update now invisible object can also block your way!! Horray!!
But now I've notice there's inconsistency between TileMap and other physics object, I hereby claim my code has no issue, I'll put this part of the project on halt until it's fixed.

### Steps to reproduce

1.Set up any CollisionObject2D inherited class except PhysicalBone2D, give them CollisionShape2D to make them functional
2.Also create a TileMap with Physics layer property
3.Create RayCast2D or Area2D, print raycast.get_collider() or area.get_overlapping_bodies()/get_overlapping_areas() to check if all of the above are detected
4.Set everything except detector's process mode to disable
5.Print again, you should see TileMap is still there while every other stuffs are undetectable 

[ScriptTest.zip](https://github.com/godotengine/godot/files/13772349/ScriptTest.zip)
1.Press "Detect" and see how much it detect
2.Press "Set Disable" to disable everything aside detector
3.Press "Detect" again to see how the TileMap stands his ground to block my project progression
(Note: I originally had another "Set Enable" button, but somehow only CharacterBody agrees to turn back on, might be another issue but I'm not bothering)

kaluluosi:
* fixes https://github.com/godotengine/godot/issues/86527


groud:
Hey, thanks for the contribution.

This is however, not the right way to implement the feature. Please have a look the implementation in `CollisionObject2D`. There something like that that is done:

```
void CollisionObject2D::_notification(int p_what) {
		...
		case NOTIFICATION_DISABLED: {
			_apply_disabled();
		} break;

		case NOTIFICATION_ENABLED: {
			_apply_enabled();
		} break;
}
```
You should not have to add anything to `node.cpp`

kaluluosi:
> Hey, thanks for the contribution.
> 
> This is however, not the right way to implement the feature. Please have a look the implementation in `CollisionObject2D`. There something like that that is done:
> 
> ```
> void CollisionObject2D::_notification(int p_what) {
> 		...
> 		case NOTIFICATION_DISABLED: {
> 			_apply_disabled();
> 		} break;
> 
> 		case NOTIFICATION_ENABLED: {
> 			_apply_enabled();
> 		} break;
> }
> ```
> 
> You should not have to add anything to `node.cpp`

Thanks for remind. I didn't notic the notification.

scene/2d/tile_map_layer.cpp
```diff
@@ -677,7 +677,7 @@ void TileMapLayer::_physics_update() {
 	const Ref<TileSet> &tile_set = get_effective_tile_set();
 
 	// Check if we should cleanup everything.
-	bool forced_cleanup = in_destructor || !enabled || !is_inside_tree() || !tile_set.is_valid();
+	bool forced_cleanup = in_destructor || !enabled || is_inside_tree() || !tile_set.is_valid() || !tile_map_node->is_enabled();
```
AThousandShips, line 680:
```suggestion
	bool forced_cleanup = in_destructor || !enabled || !is_inside_tree() || !tile_set.is_valid() || !tile_map_node->is_enabled();
```
Unless I'm mistaken this is necessary, cc @groud

scene/2d/tile_map.cpp
```diff
@@ -259,6 +259,12 @@ void TileMap::_notification(int p_what) {
 				_update_notify_local_transform();
 			}
 		} break;
+		case NOTIFICATION_DISABLED: {
```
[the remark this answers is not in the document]
kaluluosi: 哦，对。谢谢提醒。这里可以更简略。

scene/2d/tile_map_layer.cpp
kaluluosi: sorry. my fail.

PR #86925, opened 2024-01-07
Avoid calling `set_scale` instead of avoiding calling `set_rotation`

### LINKED ISSUE #86889: RemoteTransform2D set to update position and rotation only, is slightly affecting scale.
### Tested versions

- Discovered in v4.1.3.stable.official [f06b6836a]
- Reproducible in v4.2.1.stable.official [b09f793f5] and v4.3.dev1.official [9d1cbab1c]
- Could not reproduce in v3.5.3.stable.official [6c814135b]

### System information

Godot v4.1.3.stable - Ubuntu 22.04.3 LTS 22.04 - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce GT 730 (nvidia; 470.223.02) - AMD FX(tm)-4100 Quad-Core Processor (4 Threads)

### Issue description

I'm using a RemoteTransform2D node to set the position and rotation of a Sprite2D. The scale of the Sprite2D should not be affected.
The parent of the RemoteTransform2D has a scale that is not {1,1}, and the RemoteTransform2D node is set to use global co-ordinates.
The scale of the target Sprite2D is changing slightly each time the position and rotation are updated. This change is only around 0.0000001 each frame, but it is accumulating over a long-playing animation, to the point of becoming noticeable after an hour or so.
I suspect that there is a floating-point error introduced when calculating the new transform matrix.

### Steps to reproduce

MRP attached, open the main scene and view the output generated by the tool script. This will show that the icon scale is not exactly {1,1} as expected, but is approximately so.
Run the animation in the player for a few minutes to see the scale drift over time.

[RemoteTransformScaleBug.zip](https://github.com/godotengine/godot/files/13850354/RemoteTransformScaleBug.zip)

Rindbee:
Setting the rotation will correct the scale, but not vice versa (And it will cause the cumulative effect of floating point errors to cause offset).

This does not completely eliminate floating point errors in some cases (initial scale is `(1.1, 1.1)`), but performs better in some cases (initial scale is `(1, 1)`). 

Fix #86889.

### DIFF
--- a/scene/2d/remote_transform_2d.cpp
+++ b/scene/2d/remote_transform_2d.cpp
@@ -74,14 +74,13 @@ void RemoteTransform2D::_update_remote() {
 		Transform2D n_trans = n->get_global_transform();
 		Transform2D our_trans = get_global_transform();
 
-		// There are more steps in the operation of set_rotation, so avoid calling it.
-		Transform2D trans = update_remote_rotation ? our_trans : n_trans;
+		Transform2D trans = update_remote_scale ? our_trans : n_trans;
 
-		if (update_remote_rotation ^ update_remote_position) {
+		if (update_remote_scale ^ update_remote_position) {
 			trans.set_origin(update_remote_position ? our_trans.get_origin() : n_trans.get_origin());
 		}
-		if (update_remote_rotation ^ update_remote_scale) {
-			trans.set_scale(update_remote_scale ? our_trans.get_scale() : n_trans.get_scale());
+		if (update_remote_scale ^ update_remote_rotation) {
+			trans.set_rotation(update_remote_rotation ? our_trans.get_rotation() : n_trans.get_rotation());
 		}
 
 		n->set_global_transform(trans);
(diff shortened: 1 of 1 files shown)


kleonc:
This makes so the property not being updated by the RemoteTransform2D effectively can't be set in the target Node2D because the RemoteTransform2D will overwrite it with the cached value. The cache would need to be updated after any transform change of such target Node2D I guess? :thinking:

Like in the MRP from #86889:
```gdscript
# Icon.gd
extends Sprite2D

func _ready():
	scale = Vector2.ONE

func _process(delta: float) -> void:
	var div := 300.0
	var frame = Engine.get_process_frames()
	var s = scale
	scale = Vector2(frame / div, frame / div)
	print("%7.3v %7.3v" % [s, scale])
```
Outputs for this PR:
``` 
(  1.000,   1.000) (  0.000,   0.000)
(  1.000,   1.000) (  0.003,   0.003)
(  1.000,   1.000) (  0.007,   0.007)
(  1.000,   1.000) (  0.010,   0.010)
(  1.000,   1.000) (  0.013,   0.013)
(  1.000,   1.000) (  0.017,   0.017)
(  1.000,   1.000) (  0.020,   0.020)
(  1.000,   1.000) (  0.023,   0.023)
(  1.000,   1.000) (  0.027,   0.027)
(  1.000,   1.000) (  0.030,   0.030)
(  1.000,   1.000) (  0.033,   0.033)
(  1.000,   1.000) (  0.037,   0.037)
...
```
So the scale keeps being reseted to the cached value although it's not meant to be updated/affected by the RemoteTransform2D.
QPEFnXAfhQ

Rindbee:
If there is no other way to change the node's transform, it may be better to use cached values to correct for floating point errors.

If there are other ways to change the node's transform, the transform set there will become the standard for correcting the result (the floating point error can be ignored).

Using floating intermediate values does not guarantee that the final result will not deviate from the original value.

Rindbee:
This PR cannot fundamentally solve the linked issue.

PR #86928, opened 2024-01-07
Update file_access.cpp

LagowiecDev:

### DIFF
--- a/core/io/file_access.cpp
+++ b/core/io/file_access.cpp
@@ -803,20 +803,23 @@ String FileAccess::get_multiple_md5(const Vector<String> &p_file) {
 	ctx.start();
 
 	for (int i = 0; i < p_file.size(); i++) {
-		Ref<FileAccess> f = FileAccess::open(p_file[i], READ);
-		ERR_CONTINUE(f.is_null());
-
-		unsigned char step[32768];
-
-		while (true) {
-			uint64_t br = f->get_buffer(step, 32768);
-			if (br > 0) {
-				ctx.update(step, br);
-			}
-			if (br < 4096) {
-				break;
-			}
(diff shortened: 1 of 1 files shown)


core/io/file_access.cpp
```diff
@@ -803,20 +803,23 @@ String FileAccess::get_multiple_md5(const Vector<String> &p_file) {
...
-			if (br > 0) {
-				ctx.update(step, br);
-			}
-			if (br < 4096) {
-				break;
-			}
-		}
+	    Ref<FileAccess> f = FileAccess::open(p_file[i], READ);
+	    if (f.is_null()) {
+	        ERR_FAIL_V_MSG(f.is_null(), "Error opening file: " + p_file[i] + ".");
```
AThousandShips, line 808: This is invalid, this will not continue but exit, and also this macro takes more arguments

What are you trying to achieve?

AThousandShips (changes requested):

core/io/file_access.cpp
LagowiecDev: i getting:  core/io/file_access.cpp:807 - Condition "f.is_null()" is true. Continuing.

core/io/file_access.cpp
AThousandShips, line 1: You have also changed tabs into spaces, that's not correct

core/io/file_access.cpp
AThousandShips: Instead you can change the line with `ERR_CONTINUE(f.is_null());` with:
```cpp
ERR_CONTINUE_MSG(f.is_null(), vformat("Error opening file: '%s'.", p_file[i]));
```

PR #86929, opened 2024-01-07
Deprecate usage of nested Arrays as key-value pairs in `String.format()`

### LINKED ISSUE #8753: Deprecate/discourage usage of nested Arrays in `String.format()`

It's not relevant for this proposal.

The String's _format()_ method supports both **Array** and **Dictionary**. This is excellent and perfectly functional.

##### Example taken from the String documentation
```GDScript
# Prints "Waiting for Godot is a play by Samuel Beckett, and Godot Engine is named after it."
var use_array_values = "Waiting for {0} is a play by {1}, and {0} Engine is named after it."
print(use_array_values.format(["Godot", "Samuel Beckett"]))

# Prints "User 42 is Godot."
print("User {id} is {name}.".format({"id": 42, "name": "Godot"}))
```

But, many people may not even realised you can use nested arrays:

```GDScript
print("User {id} is {name}.".format([["id", 42], ["name", "Godot"]]))
```

The problem is, [the documentation about this has always been dry](https://github.com/godotengine/godot/issues/69253). And admittedly, explaining it in detail without confusing it with the normal **Array** behavior would lead to... very little gain (<sub>bad sign</sub>)...
If the nested **Array** is meant to be interpreted as a key-value pair, why not just using a **Dictionary**? It's much more straightforward. To the average user, if there even is _any_ performance gain, it would not matter.

Discourage the usage of a nested array for String's exposed `format()` method, essentially deprecating it. Encourage using a **Dictionary**, instead, and warn about a possible removal in next major release.

```GDScript
# Deprecate:
print("User {id} is {name}.".format([["id", 42], ["name", "Godot"]]))
# In favour of:
print("User {id} is {name}.".format({"id": 42, "name": "Godot"}))
```

It is extremely easy to gloss over, yes. 

String itself is core.

Mickeon:
Closes https://github.com/godotengine/godot-proposals/issues/8753.

For **String**'s `format()`, This PR deprecates using nested arrays in an **Array** as key-value pairs, recommending dictionaries for that purpose, instead. The functionality still exists, but is discouraged with a warning. The existing behavior with Arrays is left unchanged.

This is an obscure behavior that is barely worth documenting. See the proposal for further reasoning.

### DIFF
--- a/core/string/ustring.cpp
+++ b/core/string/ustring.cpp
@@ -3552,9 +3552,11 @@ String String::format(const Variant &p_values, const String &p_placeholder) cons
 		Array values_arr = p_values;
 
 		for (int i = 0; i < values_arr.size(); i++) {
-			if (values_arr[i].get_type() == Variant::ARRAY) { //Array in Array structure [["name","RobotGuy"],[0,"godot"],["strength",9000.91]]
+#ifndef DISABLE_DEPRECATED
+			if (values_arr[i].get_type() == Variant::ARRAY) { // Array in Array structure [["name","RobotGuy"], [0,"godot"], ["strength",9000.91]].
 				Array value_arr = values_arr[i];
 
+				WARN_DEPRECATED_MSG("In String.format(), Arrays inside another Array are treated as key-value pairs. This behavior is deprecated. Consider using a Dictionary instead.");
 				if (value_arr.size() == 2) {
 					String key = value_arr[0];
 					String val = value_arr[1];
@@ -3563,14 +3565,17 @@ String String::format(const Variant &p_values, const String &p_placeholder) cons
 				} else {
 					ERR_PRINT(vformat("Invalid format: the inner Array at index %d needs to contain only 2 elements, as a key-value pair.", i).ascii().get_data());
 				}
-			} else { //Array structure ["RobotGuy","Logis","rookie"]
-				String val = values_arr[i];
+				continue;
+			}
+#endif // DISABLE_DEPRECATED
 
-				if (p_placeholder.contains_char('_')) {
-					new_string = new_string.replace(p_placeholder.replace("_", String::num_int64(i)), val);
-				} else {
-					new_string = new_string.replace_first(p_placeholder, val);
-				}
+			// Array structure ["RobotGuy", "Logis", "rookie"].
+			String val = values_arr[i];
+
+			if (p_placeholder.contains_char('_')) {
+				new_string = new_string.replace(p_placeholder.replace("_", String::num_int64(i)), val);
+			} else {
+				new_string = new_string.replace_first(p_placeholder, val);
 			}
 		}
 	} else if (p_values.get_type() == Variant::DICTIONARY) {


core/string/ustring.cpp
```diff
@@ -3524,10 +3524,16 @@ String String::format(const Variant &values, const String &placeholder) const {
...
 			String i_as_str = String::num_int64(i);
 
-			if (values_arr[i].get_type() == Variant::ARRAY) { //Array in Array structure [["name","RobotGuy"],[0,"godot"],["strength",9000.91]]
+#ifndef DISABLE_DEPRECATED
+			bool already_warned = false;
+			if (values_arr[i].get_type() == Variant::ARRAY) { // Array in Array structure [["name","RobotGuy"], [0,"godot"], ["strength",9000.91]].
 				Array value_arr = values_arr[i];
 
 				if (value_arr.size() == 2) {
+					if (!already_warned) {
```
AThousandShips, line 3533: Unnecessary, this is only shown once:
```cpp
#define WARN_DEPRECATED_MSG(m_msg)                                                                                                                                       \
	if (true) {                                                                                                                                                          \
		static std::atomic<bool> warning_shown;                                                                                                                          \
		if (!warning_shown.load()) {                                                                                                                                     \
			_err_print_error(FUNCTION_STR, __FILE__, __LINE__, "This method has been deprecated and will be removed in the future.", m_msg, false, ERR_HANDLER_WARNING); \
			warning_shown.store(true);                                                                                                                                   \
		}                                                                                                                                                                \
	} else                                                                                                                                                               \
		((void)0)
```

Mickeon: That's cool, they were way ahead of me.

core/string/ustring.cpp
```diff
@@ -3524,10 +3524,12 @@ String String::format(const Variant &values, const String &placeholder) const {
 		for (int i = 0; i < values_arr.size(); i++) {
 			String i_as_str = String::num_int64(i);
 
-			if (values_arr[i].get_type() == Variant::ARRAY) { //Array in Array structure [["name","RobotGuy"],[0,"godot"],["strength",9000.91]]
+#ifndef DISABLE_DEPRECATED
+			if (values_arr[i].get_type() == Variant::ARRAY) { // Array in Array structure [["name","RobotGuy"], [0,"godot"], ["strength",9000.91]].
 				Array value_arr = values_arr[i];
 
 				if (value_arr.size() == 2) {
+					WARN_DEPRECATED_MSG("Arrays structured as key-value pairs in String.format() are deprecated. Consider using a Dictionary, instead.");
```
Maran23, line 3532: `Consider using a Dictionary instead.` sounds better. I don't think the comma is needed here.

Mickeon:
I don't see the value in doing this later than sooner. If agreed upon, it would be nice to merge ASAP so that if there's any sort of minor backlash this "feature" can be kept.

core/string/ustring.cpp
```diff
@@ -3538,15 +3540,18 @@ String String::format(const Variant &values, const String &placeholder) const {
 				} else {
 					ERR_PRINT(String("STRING.format Inner Array size != 2 ").ascii().get_data());
 				}
-			} else { //Array structure ["RobotGuy","Logis","rookie"]
-				Variant v_val = values_arr[i];
-				String val = v_val;
+				continue;
```
growingbrain, line 3543: Should this be a 'break' or can we use both 'Array' and 'nested Array' in the same format call?

core/string/ustring.cpp
```diff
@@ -3538,15 +3540,18 @@ String String::format(const Variant &values, const String &placeholder) const {
...
-					new_string = new_string.replace_first(placeholder, val);
-				}
+			// Array structure ["RobotGuy", "Logis", "rookie"].
+			Variant v_val = values_arr[i];
+			String val = v_val;
+
+			if (placeholder.find("_") > -1) {
+				new_string = new_string.replace(placeholder.replace("_", i_as_str), val);
+			} else {
+				new_string = new_string.replace_first(placeholder, val);
```
growingbrain, line 3554: Should we move this code to be the first condition.
if (values_arr[i].get_type() == Variant::STRING) {
...
}
#ifndef DISABLE_DEPRECATED
			if (values_arr[i].get_type() == Variant::ARRAY) {
}

growingbrain (commented):
Maybe some re-organization to put the non-warning Array first and the case of nested array last in the if/else if.

Mickeon:
Not sure how to feel about the suggestions. I tried removing deprecating this in a way that would not impact the original code too drastically.

core/string/ustring.cpp
Mickeon: You always could mix the two (_for some reason_).

growingbrain:
> Not sure how to feel about the suggestions. I tried removing deprecating this in a way that would not impact the original code too drastically.

No worries, no need to make any changes based on my comments.  Thanks for the work!

Mickeon:
Rebased after a0dbdcc3abbd3e6307c6e68d0e60f8c0fa31d576

Mickeon:
Rebased once more

core/string/ustring.cpp
```diff
@@ -3942,10 +3942,12 @@ String String::format(const Variant &values, const String &placeholder) const {
 		for (int i = 0; i < values_arr.size(); i++) {
 			String i_as_str = String::num_int64(i);
 
-			if (values_arr[i].get_type() == Variant::ARRAY) { //Array in Array structure [["name","RobotGuy"],[0,"godot"],["strength",9000.91]]
+#ifndef DISABLE_DEPRECATED
+			if (values_arr[i].get_type() == Variant::ARRAY) { // Array in Array structure [["name","RobotGuy"], [0,"godot"], ["strength",9000.91]].
 				Array value_arr = values_arr[i];
 
 				if (value_arr.size() == 2) {
+					WARN_DEPRECATED_MSG("Arrays structured as key-value pairs in String.format() are deprecated. Consider using a Dictionary instead.");
```
Ivorforce, line 3950: I think it makes sense to warn on any array size. I'd also clarify what is happening:
 
```suggestion
				if (value_arr.size() == 2) {
					WARN_DEPRECATED_MSG("When using Arrays in Arrays, they are interpreted as key-value pairs and must be of length 2. This behavior is deprecated. Consider formatting with a Dictionary instead.");
```

core/string/ustring.cpp
```diff
@@ -3956,15 +3958,18 @@ String String::format(const Variant &values, const String &placeholder) const {
...
+#endif // DISABLE_DEPRECATED
 
-				if (placeholder.contains_char('_')) {
-					new_string = new_string.replace(placeholder.replace("_", i_as_str), val);
-				} else {
-					new_string = new_string.replace_first(placeholder, val);
-				}
+			// Array structure ["RobotGuy", "Logis", "rookie"].
+			Variant v_val = values_arr[i];
+			String val = v_val;
```
Ivorforce, line 3967: While you're at it, why not remove the unnecessary copy?

```suggestion
			String val = values_arr[i];
```

Ivorforce (approved):
I'm all for this. This is weird, inconsistent behavior, and should be removed (see https://github.com/godotengine/godot-proposals/issues/8753#issuecomment-2873309359).

Deprecating it seems appropriate.

core/string/ustring.cpp
Mickeon: I'll see what I can do if I can rebase this, because Git doesn't quite catch that most of what has been done in the PR is indent code

core/string/ustring.cpp
Mickeon: I will try moving this upwards but I'd be wary of clarifying too much because sometimes lots of words are more confusing than fewer.

hpvb (approved):
We discussed this in the core meeting, nobody seems to want to keep the old behavior. We can merge this.

Repiteo:
Thanks!

PR #86933, opened 2024-01-07
Fix 2D viewport texture drop issue with global history registration

### LINKED ISSUE #86826: Dropping a texture in empty scene creates a global action that can erase other scenes
### Tested versions

4.3 dev, but probably exists since 4.0

### System information

### Issue description

When you drop a texture to 2D viewport and the scene has no root, the newly created node will be set as root, but the action is registered in global history. If you undo it and redo *on another scene*, the newly created node will replace the root of current scene:

https://github.com/godotengine/godot/assets/2223172/c4676c44-6e50-4a4f-945a-1c1127ac7f97

Trying to undo this operation will result in empty scene. You need to revert it to restore the original content.

### Steps to reproduce

1. Open a non-empty scene
2. Create new scene
3. In the new scene, drop a texture file into 2D view
4. Undo
5. Go back to the first scene
6. Redo

N/A

alessandrofama:
This fix addresses an issue where dropping a texture onto the 2D viewport in a scene without a root node incorrectly registers the action in the global history. Subsequently undoing and redoing this action in another scene results in the newly created node replacing the root of the current scene. Fixes #86826.

Fix video:

https://github.com/godotengine/godot/assets/26153311/886dbb96-3017-4702-bf8d-fee70b668b72

### DIFF
--- a/editor/plugins/canvas_item_editor_plugin.cpp
+++ b/editor/plugins/canvas_item_editor_plugin.cpp
@@ -5842,7 +5842,7 @@ void CanvasItemEditorViewport::_perform_drop_data() {
 	Vector<String> error_files;
 
 	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
-	undo_redo->create_action(TTR("Create Node"));
+	undo_redo->create_action_for_history(TTR("Create Node"), EditorNode::get_editor_data().get_current_edited_scene_history_id());
 
 	for (int i = 0; i < selected_files.size(); i++) {
 		String path = selected_files[i];


editor/plugins/canvas_item_editor_plugin.cpp
```diff
@@ -5842,7 +5842,7 @@ void CanvasItemEditorViewport::_perform_drop_data() {
 	Vector<String> error_files;
 
 	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
-	undo_redo->create_action(TTR("Create Node"));
+	undo_redo->create_action_for_history(TTR("Create Node"), EditorNode::get_singleton()->get_editor_data().get_current_edited_scene_history_id());
```
KoBeWi, line 5845:
```suggestion
	undo_redo->create_action_for_history(TTR("Create Node"), EditorNode::get_editor_data().get_current_edited_scene_history_id());
```
This method is static.

KoBeWi (approved):

editor/plugins/canvas_item_editor_plugin.cpp
alessandrofama: Fixed 👍

PR #86934, opened 2024-01-07
[macOS export] Fix RLE icon generation.

### LINKED ISSUE #86922: macOS export generated icns icons corrupted for 32 and 16 px size
### Tested versions

4.2.1.stable

### System information

macOS 14.2.1 and macOS 10.15.7

### Issue description

If you just specify a project icon in the project settings `application/config/icon`Godot will automatically generate platform specific icns files for macOS (and for other platforms I guess). However, it seems there's an issue with how these are generated, since the two lowest sizes in the icns file 32x and 16x seems to be corrupted (looks like the channels are misaligned) 

Screenshot 2024-01-07 at 14 20 43

Searching for clues I found this old "fix" https://github.com/godotengine/godot/pull/23989/ to what sounds like the same issue. However, I'm not sure if it actually fixed the issue or if this fix is what is causing the current issue. I'm sure there was a reason for it at the time, but it seems odd to me that it specifically encodes the two 32px and 16px versions of the icons in a different format than the rest of the sizes, and these now display incorrectly. 

I've also tested the icon display in both on the current macOS 14.2.1 and an old 10.15.7 installation and they display incorrectly on both. So it doesn't appear lika a format change in macOS icns. Only you probably won't notice it on newer macs since they're all retina and don't use the lower res versions of the icons in display.

### Steps to reproduce

- Set up a project with a png project icon.
- Export the project for macOS and observe generated .icns file in the exported package

Here's the generated icns icons file from above
[icon.icns.zip](https://github.com/godotengine/godot/files/13853613/icon.icns.zip)

This is an export icon generation issue, and I don't think it's specific to a project.

bruvzg:
Fixes https://github.com/godotengine/godot/issues/86922

There's no official documentation for the format, but seems like it's not exactly PackBits RLE (while it works fine with most images), so I have replaced it with [this algorithm](https://github.com/nyteshade/ByteRunLengthCoder/blob/main/Encode%20Pseudocode.md) which seems to be more compatible.

### DIFF
--- a/platform/macos/export/export_plugin.cpp
+++ b/platform/macos/export/export_plugin.cpp
@@ -494,65 +494,49 @@ void _rgba8_to_packbits_encode(int p_ch, int p_size, Vector<uint8_t> &p_source, 
 	int src_len = p_size * p_size;
 
 	Vector<uint8_t> result;
-	result.resize(src_len * 1.25); //temp vector for rle encoded data, make it 25% larger for worst case scenario
-	int res_size = 0;
-
-	uint8_t buf[128];
-	int buf_size = 0;
 
 	int i = 0;
+	const uint8_t *src = p_source.ptr();
 	while (i < src_len) {
-		uint8_t cur = p_source.ptr()[i * 4 + p_ch];
+		Vector<uint8_t> seq;
 
-		if (i < src_len - 2) {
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #62735, opened 2022-07-05
Allows parsing of invalid UTF-16 surrogates and some non-standard UTF-8 variants, makes Unicode parse errors more verbose.

bruvzg:
- Allows parsing and storing of invalid (unpaired) UTF-16 surrogates (can be encountered in Windows filenames).
- Allows parsing of UTF-8 overlong encoding and non-standard characters (outside Unicode range, unpaired surrogates).
- Disallow `NUL` characters in the middle of `String`.
- Instead of aborting parse on the first error, invalid parts are replaced with spaces (`0x20`).
- Makes Unicode parse errors more verbose.
- Changes `parse_utfXX` return type to `Error`, instead of misleading `== true` for error. Possible return values:
  - `ERR_CANT_OPEN` - `nullptr` as input;
  - `FAILED` - some parts can't be decoded and replaced with spaces.
  - `ERR_PARSE_ERROR` - some parts are non Unicode compliant, but no data is lost.
  - `OK` - string is fully valid.

PR #86936, opened 2024-01-07
Attempt parsing environment variables as UTF-8

akx:
## What?

This PR makes the Unix `get_environment()` implementation attempt to decode values as UTF-8, and if that fails, return the value as-is, as has been the previous behavior. (Since the UNIX specification does not specify an encoding for environment variables, this sounds like the safe bet.)

This matches what the Windows `get_environment()` implementation does in spirit; it uses `GetEnvironmentVariableW()` and `String::utf16`.

## Why?

I have a project which renders texts in short movie sequences based on inputs read from environment variables (as that seemed the easiest way to get values into the Godot process), and there was an audible groan when I saw the all-too-familiar UTF-8 [mojibake](https://en.wikipedia.org/wiki/Mojibake) for `ä` being interpreted as Latin-1:

With vanilla `4.2.1.stable.official.b09f793f5` on macOS:

pre

With this PR:

post

### DIFF
--- a/drivers/unix/os_unix.cpp
+++ b/drivers/unix/os_unix.cpp
@@ -703,10 +703,15 @@ bool OS_Unix::has_environment(const String &p_var) const {
 }
 
 String OS_Unix::get_environment(const String &p_var) const {
-	if (getenv(p_var.utf8().get_data())) {
-		return getenv(p_var.utf8().get_data());
+	const char *val = getenv(p_var.utf8().get_data());
+	if (val == nullptr) { // Not set; return empty string
+		return "";
 	}
-	return "";
+	String s;
+	if (s.parse_utf8(val) == OK) {
+		return s;
+	}
+	return String(val); // Not valid UTF-8, so return as-is
 }
 
 void OS_Unix::set_environment(const String &p_var, const String &p_value) const {
--- a/tests/core/os/test_os.h
+++ b/tests/core/os/test_os.h
@@ -47,11 +47,33 @@ TEST_CASE("[OS] Environment variables") {
 			OS::get_singleton()->has_environment("HOME"),
 			"The HOME environment variable should be present.");
 #endif
+}
 
-	OS::get_singleton()->set_environment("HELLO", "world");
+TEST_CASE("[OS] UTF-8 environment variables") {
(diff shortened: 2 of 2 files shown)


AThousandShips:
Since this applies to all these platforms we need to confirm this also works on all these platforms

The POSIX standard specifies the portable character set for values, for portability, unsure if this is a strict standard

bruvzg:
> The POSIX standard specifies the portable character set for values, for portability, unsure if this is a strict standard

Portable character set is ASCII without control chars, so it's compatible with UTF-8.

AThousandShips:
Unsure what happens if it's encoded differently but that'll just be differently garbled, so not a serious issue

drivers/unix/os_unix.cpp
```diff
@@ -703,10 +703,15 @@ bool OS_Unix::has_environment(const String &p_var) const {
...
+	const char *val = getenv(p_var.utf8().get_data());
+	if (val == nullptr) { // Not set; return empty string
+		return "";
 	}
-	return "";
+	String s;
+	if (s.parse_utf8(val) == OK) {
+		return s;
+	}
+	return String(val); // Not valid UTF-8, so return as-is
```
bruvzg, line 714:
```suggestion
	return String::utf8(val);
```

I'm not sure if this check is necessary, `String(val)` will work correctly only if string is Latin-1 (ASCII will be handled by UTF-8 parser), and this is an arbitrary assumption.

A proper way to check encoding is likely reading the value of `LC_CTYPE` variable. But for all modern systems it should be UTF-8.

akx: I copied this idea from https://github.com/godotengine/godot/blob/c8c483cf57a768110fce57e509f9b855e69d34b7/drivers/unix/os_unix.cpp#L515-L519, introduced in https://github.com/godotengine/godot/pull/62735.

If you want me to change that to the suggested, sure, but this seemed more bullet-proof.

bruvzg: It's OK as is. It both cases, we only guarantee to parse UTF-8 and ASCII.

akx:
So... how would I get this merged? I'm currently stuck forward-porting this patch locally :)

bruvzg (approved):

akien-mga:
Thanks!

Issue #56341, opened 2021-12-30
Godot 4.0 editor GUI scaling doesn't react to moving window to different monitor


hhyyrylainen:
### Godot version

v4.0.dev.custom_build.91b97dac0

### System information

Linux (Fedora 35), Wayland, Vulkan API 1.2.189 - Using Vulkan Device #0: AMD - AMD RADV SIENNA_CICHLID (6900 XT)

### Issue description

When moving a Godot editor window from my 4K screen to my smaller screen, the GUI scaling stays the same, meaning that all the GUI elements take up a much larger proportion of the Godot window than they probably should. At least I can see much less content in the file and node trees, also the placeholder text is even cut off on the search bar. Instead Godot should scale down the GUI when moved to a monitor with a different DPI / resolution.

Here's how the really large scaling looks in 4.0:
![Kuvakaappaus - 2021-12-30 10-13-58](https://user-images.githubusercontent.com/3796411/147734423-8937aa1b-3d52-4a5e-a669-3106abb869a3.png)

And for reference here's 3.4 (showing a different project) where I can actually see much more useful content:
![Kuvakaappaus - 2021-12-30 10-19-13](https://user-images.githubusercontent.com/3796411/147734437-c55336a4-c93c-4a16-bd9a-67ba98a13b69.png)

### Steps to reproduce

- Checkout the latest master code
- Compile with `scons target=release_debug -j 30 tools=yes`
- Run editor from terminal: `bin/godot.linuxbsd.opt.tools.64`
- Observe how the project manager window is well scaled on the 4K monitor it opens on, but when I move it over to my primary monitor that is 2560x1440 all the GUI elements are very large
- Open any project and observe again, when moved to the 2560x1440 monitor (I opened a separate issue about things opening on the wrong monitor: https://github.com/godotengine/godot/issues/56340), how the Godot editor GUI elements are scaled too big, with the usability being pretty severely impacted as very little content can fit on screen.

### Minimal reproduction project

problem is with the Godot editor itself, a new project / any project can be opened to see more of the issue

Calinou:
Run-time editor scale changes would need to be supported to allow for this, but this would require a lot of work. https://github.com/godotengine/godot/pull/52170 might make this easier, but it's still far from trivial.

alvinhochun:
Thanks @Calinou for kindly pointing me to this issue from IRC, and because of that I thought I'd put my two cents here...

[...]

Issue #84948, opened 2023-11-15
Resolution issue with dual monitors with different DPI


Jules5:
### Godot version

4.1.2.stable

### System information

Vulkan API 1.3.224 - Forward+ - Using Vulkan Device #0: NVIDIA - NVIDIA GeForce GTX 1060 with Max-Q Design

### Issue description

I have two screens: 
- my laptop's one, 1920x1080, 120 DPI, 125% Windows scale 
- and an external one, 1920x1080, 96 DPI, 100% Windows scale

I work on a 2D pixel perfect game. When I display it fullscreen on my laptop everything is nice and sharp, but not on the external screen where I have scaling issues (the pixel art is altered). I logged the screen resolution according to Godot and I got this:
- laptop: 1920x1080
- external: 2400x1350 (which is 125% of 1920x1080...)

I tried to disable the "allow_hidpi" settings, then I got this: 
- laptop: 1536x864 (125% of this gives 1920x1080...)
- external: 1920x1080

Finally I put back the "allow_hidpi" settings and disabled the 125% scaling in windows settings: 
- laptop: 1920x1080
- external: 1920x1080

As I understand these results I'd say that Godot is handling well the Windows scaling through the "allow_hidpi" setting, but is failign to handle correctly the case where the two screens have a different scaling value in the Windows settings. I expected Godot to use a 1920x1080 resolution for the two screens in this case.

### Steps to reproduce

- Have two screens with a same resolution
- Give screen 1 a 125% scaling in Windows settings
- Give screen 2 a 100% scaling in Windows settings
- Launch any Godot scene and print "DisplayServer.screen_get_size()"
- Move the window to a screen to another and notice the displayed value changes.

### Minimal reproduction project

N/A

Calinou:
Godot doesn't use the per-monitor DPI awareness API yet, as per https://github.com/godotengine/godot/issues/56341.

Please upload a minimal reproduction project[^1] to make this easier to troubleshoot.

[^1]: A small Godot project which reproduces the issue, with no unnecessary files included. Be sure to not include the `.godot` folder in the archive (but keep `project.godot`).<br><br>Drag and drop a ZIP archive to upload it. **Do not select another field until the project is done uploading.**<br><br>**Note for C# users:** If your issue is *not* Mono-specific, please upload a minimal reproduction project written in GDScript or VisualScript. This will make it easier for contributors to reproduce the issue locally as not everyone has a Mono setup available.

Jules5:
[...]

PR #86943, opened 2024-01-07
Use `PER_MONITOR_DPI_AWARE` option if it's supported by OS

100gold:
Fixes some problems described in #56341.
It is not an implementation of automatic scaling when a window moves from one monitor to another, but at least the PER_MONITOR_DPI_AWARE option can fix problems with unnecessary scaling for secondary monitors.

A new project setting "display/window/dpi/hidpi_awareness" has been introduced, which replaces "display/window/dpi/allow_hidpi". New setting has 3 options:
    "No" - application not ready for high-resolution displays.
    "System wide" - application can work with high-resolution displays, but is not ready for different dpi on different monitors.
    "Per monitor" - application can work with high-resolution displays, also can handle situations when different displays use their own dpi value.

### DIFF
--- a/core/os/os.h
+++ b/core/os/os.h
@@ -47,7 +47,7 @@ public:
 	enum HidpiAwareness {
 		NO_AWARENESS,
 		SYSTEM_WIDE_AWARENESS,
-		PER_MONITOR_AWARENESS
+		PER_MONITOR_AWARENESS,
 	};
 
 private:
--- a/core/os/os.h
+++ b/core/os/os.h
@@ -43,6 +43,14 @@
 class MainLoop;
 
 class OS {
+public:
+	enum HidpiAwareness {
+		NO_AWARENESS,
+		SYSTEM_WIDE_AWARENESS,
+		PER_MONITOR_AWARENESS
+	};
+
+private:
 	static OS *singleton;
 	static uint64_t target_ticks;
 	String _execpath;
@@ -57,7 +65,7 @@ class OS {
 	String _local_clipboard;
 	// Assume success by default, all failure cases need to set EXIT_FAILURE explicitly.
 	int _exit_code = EXIT_SUCCESS;
-	bool _allow_hidpi = false;
+	HidpiAwareness _hidpi_awareness = HidpiAwareness::NO_AWARENESS;
 	bool _allow_layered = false;
 	bool _stdout_enabled = true;
 	bool _stderr_enabled = true;
@@ -223,7 +231,7 @@ public:
 	virtual String get_model_name() const;
 
 	bool is_layered_allowed() const { return _allow_layered; }
-	bool is_hidpi_allowed() const { return _allow_hidpi; }
+	HidpiAwareness get_hidpi_awareness() const { return _hidpi_awareness; }
 
 	void ensure_user_data_dir();
 
--- a/doc/classes/ProjectSettings.xml
+++ b/doc/classes/ProjectSettings.xml
@@ -949,10 +949,10 @@
 		</member>
 		<member name="display/window/dpi/hidpi_awareness" type="int" setter="" getter="" default="1">
 			This option controls how the application handles high-resolution displays.
-			[code]No[/code] - application not ready for high-resolution displays.
-			[code]System wide[/code] - application can work with high-resolution displays, but is not ready for different dpi on different monitors.
-			[code]Per monitor[/code] - application can work with high-resolution displays, also can handle situations when different displays use their own DPI value.
-			If "No" used, the platform's low-DPI fallback will be used on HiDPI displays, which causes the window to be displayed in a blurry or pixelated manner (and can cause various window management bugs). Therefore, it is recommended to make your project scale to [url=$DOCS_URL/tutorials/rendering/multiple_resolutions.html]multiple resolutions[/url] instead of disabling this setting.
+			[code]No[/code] - the application is not ready for high-resolution displays.
+			[code]System wide[/code] - the application can work with high-resolution displays, but is not ready for different DPI values on different monitors.
+			[code]Per monitor[/code] - the application can work with high-resolution displays, it can also handle situations when different displays use their own DPI value.
+			If [b]No[/b] is used, the platform's low-DPI fallback will be used on HiDPI displays, which causes the window to be displayed in a blurry or pixelated manner (and can cause various window management bugs). Therefore, it is recommended to make your project scale to [url=$DOCS_URL/tutorials/rendering/multiple_resolutions.html]multiple resolutions[/url] instead of disabling this setting.
 			[b]Note:[/b] This setting has no effect on Linux as DPI-awareness fallbacks are not supported there.
 		</member>
 		<member name="display/window/energy_saving/keep_screen_on" type="bool" setter="" getter="" default="true">
--- a/doc/classes/ProjectSettings.xml
+++ b/doc/classes/ProjectSettings.xml
@@ -947,8 +947,12 @@
 		<member name="display/mouse_cursor/tooltip_position_offset" type="Vector2" setter="" getter="" default="Vector2(10, 10)">
 			Position offset for tooltips, relative to the mouse cursor's hotspot.
 		</member>
-		<member name="display/window/dpi/allow_hidpi" type="bool" setter="" getter="" default="true">
-			If [code]true[/code], allows HiDPI display on Windows, macOS, Android, iOS and Web. If [code]false[/code], the platform's low-DPI fallback will be used on HiDPI displays, which causes the window to be displayed in a blurry or pixelated manner (and can cause various window management bugs). Therefore, it is recommended to make your project scale to [url=$DOCS_URL/tutorials/rendering/multiple_resolutions.html]multiple resolutions[/url] instead of disabling this setting.
+		<member name="display/window/dpi/hidpi_awareness" type="int" setter="" getter="" default="1">
+			This option controls how the application handles high-resolution displays.
+			[code]No[/code] - application not ready for high-resolution displays.
+			[code]System wide[/code] - application can work with high-resolution displays, but is not ready for different dpi on different monitors.
+			[code]Per monitor[/code] - application can work with high-resolution displays, also can handle situations when different displays use their own DPI value.
+			If "No" used, the platform's low-DPI fallback will be used on HiDPI displays, which causes the window to be displayed in a blurry or pixelated manner (and can cause various window management bugs). Therefore, it is recommended to make your project scale to [url=$DOCS_URL/tutorials/rendering/multiple_resolutions.html]multiple resolutions[/url] instead of disabling this setting.
 			[b]Note:[/b] This setting has no effect on Linux as DPI-awareness fallbacks are not supported there.
 		</member>
 		<member name="display/window/energy_saving/keep_screen_on" type="bool" setter="" getter="" default="true">
--- a/editor/run/editor_run.cpp
+++ b/editor/run/editor_run.cpp
@@ -286,17 +286,24 @@ EditorRun::WindowPlacement EditorRun::get_window_placement() {
 	int window_placement = EDITOR_GET("run/window_placement/rect");
 	if (screen_rect != Rect2()) {
 		if (DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_HIDPI)) {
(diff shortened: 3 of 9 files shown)


100gold:
This #84948 issue matches better for PR, but discussion about PROCESS_PER_MONITOR_DPI_AWARE was in #56341.

bruvzg:
Simply changing DPI awareness will make the UI unusable, since Godot will not auto-scale to match it.

See https://github.com/godotengine/godot/pull/86022 for the auto-scaling implementation (should be working, but still need some work to address differences in macOS and Windows API).

100gold:
Well... yes. This code works only for godot application. If the application can scale well for different screen resolution. When the editor window moves to a display with a different scale factor, things get worse.

bruvzg:
> When the editor window moves to a display with a different scale factor, things get worse.

That's exactly why SHC_PROCESS_SYSTEM_DPI_AWARE was chosen.

Calinou:
> Simply changing DPI awareness will make the UI unusable, since Godot will not auto-scale to match it.

We can probably use this outside of the editor at least, as it affects games that use the `canvas_items` or `viewport` stretch mode (and therefore automatically scale according to the window size) aexpect to look pixel-perfect.

bruvzg:
I guess it can be an option in project settings, without proper DPI scaling it's bad for the editor, but if you are setting it for the running project and aware that you should handle scaling manually it's OK (the same can be done for macOS as well).

100gold:
I was think about it, but I don't understand some details.
There is a short plan to make new option:
- The "Allow hiDPI" checkbox becomes a choices list of 3 items "no hidpi", "system wide hidpi", "per monitor hidpi". The naming should be more adequate)
- Somehow i need to convert old project options to new ones. I guess there is some standard way to make such a converter?
- Documentation update. Should this be done in this patch? All languages?
- Patch contains only Windows code. Looks like there is similar issues for MacOS. Is any other OS should be involved?

platform/windows/display_server_windows.cpp
```diff
@@ -4658,7 +4658,11 @@ DisplayServerWindows::DisplayServerWindows(const String &p_rendering_driver, Win
 			SetProcessDpiAwareness_t SetProcessDpiAwareness = (SetProcessDpiAwareness_t)GetProcAddress(Shcore, "SetProcessDpiAwareness");
 
 			if (SetProcessDpiAwareness) {
-				SetProcessDpiAwareness(SHC_PROCESS_SYSTEM_DPI_AWARE);
+				if ((os_ver.dwMajorVersion > 6) || ((os_ver.dwMajorVersion == 6) && (os_ver.dwMajorVersion >= 3))) {
```
alvinhochun, line 4661: The version check should be unnecessary. `SetProcessDpiAwareness` is only available since Windows 8.1, so this path will practically always be taken.

alvinhochun:
I experimented locally for a bit. It turns out that enabling per-monitor DPI for only the game and not the editor causes a minor annoyance with debugging/running the game from the editor. When the editor runs system DPI aware and launches a game that runs per-monitor DPI aware, the initial position of the game window can be way off. This is because the `--position` argument that the editor uses to position the game uses monitor sizes that has been DPI-virtualized. The argument gets interpreted differently by the game, which gets monitor sizes that are *not* DPI-virtualized. So far I haven't been able to think of a way to fix this nicely.

alvinhochun:
> I was think about it, but I don't understand some details. There is a short plan to make new option:
> 
> * The "Allow hiDPI" checkbox becomes a choices list of 3 items "no hidpi", "system wide hidpi", "per monitor hidpi". The naming should be more adequate)
> * Somehow i need to convert old project options to new ones. I guess there is some standard way to make such a converter?

Should we have a completely new option for this instead, so that we don't need to change the meaning and type of an existing option? Maybe it could be something like `display/window/dpi/prefer_native_pixel_display`?

> 
> * Documentation update. Should this be done in this patch? All languages?

Translations will be handled separately on Weblate.

> 
> * Patch contains only Windows code. Looks like there is similar issues for MacOS. Is any other OS should be involved?

Honestly I have no idea how it currently works on other platforms, *especially* macOS. (I don't think you *need* to implement it for all OSes yourself though.)

Calinou:
> EDIT 2: Using SetThreadDpiAwarenessContext to temporarily switch to per-monitor DPI aware might also be an option, but only on Windows 10.

Windows versions below 10 are supported on a best effort basis, so it's fine to call this function as long as it has a no-op on older Windows versions.

100gold:
> Should we have a completely new option for this instead, so that we don't need to change the meaning and type of an existing option? Maybe it could be something like `display/window/dpi/prefer_native_pixel_display`?

This create complicated relations between two checkboxes. When one is checked, state of the other is ignored. Additionally, there is a big PR( #86022 ) which also affects the issue . That PR adds another option about dpi. However, that option can also be part of those choices list.

ghuser404:
There is an updated API for this called `DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2`.

ghuser404:
There is no real benefit in giving users option to use `SHC_PROCESS_SYSTEM_DPI_AWARE` -- because it results in blurry renders. Instead, Godot should always try to use `DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2` and then conditionally resize window (you shoulв be able to out of this for pixel-perfect games) in WM_DPICHANGED message.
```
case WM_DPICHANGED: {
	WindowData &window = windows[window_id];
	if (window.minimized || window.maximized || window.fullscreen)
		break;

	const float xscale = HIWORD(wParam) / (float)USER_DEFAULT_SCREEN_DPI;
	const float yscale = LOWORD(wParam) / (float)USER_DEFAULT_SCREEN_DPI;

	RECT* suggested = (RECT*)lParam;
	SetWindowPos(hWnd, HWND_TOP,
		suggested->left,
		suggested->top,
		suggested->right - suggested->left,
		suggested->bottom - suggested->top,
		SWP_NOACTIVATE | SWP_NOZORDER);
} break;
```

100gold:
> There is no real benefit in giving users option to use `SHC_PROCESS_SYSTEM_DPI_AWARE` -- because it results in blurry renders.

This setting is used by godot editor itself, so i think it should be kept for now.

> There is an updated API for this called `DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2`.

There are some differences between _per_monitor_aware and _per_monitor_aware_v2, but it is look likes godot doesn't use things which is affected by _V2. I don't sure what aware_version should be used.

ghuser404:
> This setting is used by godot editor itself, so i think it should be kept for now.

Yeah, and that's really bad. Literally makes everything blurry on a scaled display. Per-monitor will work just fine, because it applies natural window scaling, instead of bitmap scaling.

Shan-AsymptoteTech:
Any progress? For multi-monitor setups,  this merge is required for things to work properly.

jolexxa:
@Shan-AsymptoteTech if you are using c# or can use a little bit of it in your project, you can make use of [Platform](https://github.com/chickensoft-games/Platform) which uses native Windows API's to turn on per monitor awareness temporarily and capture the true scale factor. It also captures actual physical screen pixel dimensions on macOS to disambiguate the true scale factor there as well. The higher level [GameTools](https://github.com/chickensoft-games/GameTools) package makes use of Platform to offer easy to use scaling solutions for your game.

I wrote all this up in a [Display Scaling](https://chickensoft.games/blog/display-scaling) blog, if it's helpful. Obviously, supporting this in-engine is preferable to a workaround like this, but at least there's something.

core/os/os.h
```diff
@@ -43,6 +43,14 @@
 class MainLoop;
 
 class OS {
+public:
+	enum HidpiAwareness {
+		NO_AWARENESS,
+		SYSTEM_WIDE_AWARENESS,
+		PER_MONITOR_AWARENESS
```
AThousandShips:
```suggestion
		PER_MONITOR_AWARENESS,
```

editor/run/editor_run.cpp
```diff
@@ -286,17 +286,24 @@ EditorRun::WindowPlacement EditorRun::get_window_placement() {
...
-			bool hidpi_proj = GET_CONFIG_WITH_OVERRIDE("display", "window/dpi/allow_hidpi");
+			OS::HidpiAwareness hidpi_proj = GET_CONFIG_WITH_OVERRIDE("display", "window/dpi/hidpi_awareness");
+			OS::HidpiAwareness hidpi_editor = OS::get_singleton()->get_hidpi_awareness();
 			int display_scale = 1;
-
-			if (OS::get_singleton()->is_hidpi_allowed()) {
-				if (hidpi_proj) {
+			// There are some problems
+			// - display_scale is int, but in one case it's assigned a float value.
+			// - Is it necessary to somehow handle the situation when the project uses OS::HidpiAwareness::PER_MONITOR_AWARENESS ?
```
AThousandShips:
```suggestion
			// - Is it necessary to somehow handle the situation when the project uses OS::HidpiAwareness::PER_MONITOR_AWARENESS?
```

editor/run/editor_run.cpp
```diff
@@ -286,17 +286,24 @@ EditorRun::WindowPlacement EditorRun::get_window_placement() {
...
+			//   Looks like project is more aware of that 'dpi' than editor and should be able handle it on it's own.
```
AThousandShips:
```suggestion
			//   Looks like project is more aware of that 'dpi' than editor and should be able handle it on its own.
```

doc/classes/ProjectSettings.xml
```diff
@@ -947,8 +947,12 @@
 		<member name="display/mouse_cursor/tooltip_position_offset" type="Vector2" setter="" getter="" default="Vector2(10, 10)">
 			Position offset for tooltips, relative to the mouse cursor's hotspot.
 		</member>
-		<member name="display/window/dpi/allow_hidpi" type="bool" setter="" getter="" default="true">
-			If [code]true[/code], allows HiDPI display on Windows, macOS, Android, iOS and Web. If [code]false[/code], the platform's low-DPI fallback will be used on HiDPI displays, which causes the window to be displayed in a blurry or pixelated manner (and can cause various window management bugs). Therefore, it is recommended to make your project scale to [url=$DOCS_URL/tutorials/rendering/multiple_resolutions.html]multiple resolutions[/url] instead of disabling this setting.
+		<member name="display/window/dpi/hidpi_awareness" type="int" setter="" getter="" default="1">
+			This option controls how the application handles high-resolution displays.
+			[code]No[/code] - application not ready for high-resolution displays.
```
AThousandShips:
```suggestion
			[code]No[/code] - the application is not ready for high-resolution displays.
```

doc/classes/ProjectSettings.xml
```diff
@@ -947,8 +947,12 @@
...
+			[code]System wide[/code] - application can work with high-resolution displays, but is not ready for different dpi on different monitors.
```
AThousandShips:
```suggestion
			[code]System wide[/code] - the application can work with high-resolution displays, but is not ready for different DPI values on different monitors.
```

doc/classes/ProjectSettings.xml
```diff
@@ -947,8 +947,12 @@
...
+			[code]Per monitor[/code] - application can work with high-resolution displays, also can handle situations when different displays use their own DPI value.
```
AThousandShips:
```suggestion
			[code]Per monitor[/code] - the application can work with high-resolution displays, it can also handle situations when different displays use their own DPI value.
```

doc/classes/ProjectSettings.xml
```diff
@@ -947,8 +947,12 @@
...
+			If "No" used, the platform's low-DPI fallback will be used on HiDPI displays, which causes the window to be displayed in a blurry or pixelated manner (and can cause various window management bugs). Therefore, it is recommended to make your project scale to [url=$DOCS_URL/tutorials/rendering/multiple_resolutions.html]multiple resolutions[/url] instead of disabling this setting.
```
AThousandShips:
```suggestion
			If [b]No[/b] is used, the platform's low-DPI fallback will be used on HiDPI displays, which causes the window to be displayed in a blurry or pixelated manner (and can cause various window management bugs). Therefore, it is recommended to make your project scale to [url=$DOCS_URL/tutorials/rendering/multiple_resolutions.html]multiple resolutions[/url] instead of disabling this setting.
```

Mickeon: Uhm, `No` actually. For all of these suggestions we should be using `[b]` tags because these are the options' names, not the underlying values.

PR #86947, opened 2024-01-08
Made tooltip on `TileSet` Paint Button give more information when disabled

### LINKED ISSUE #78162: TileSet "Paint" button for painting properties stops working and becomes greyed for an unknown reason
### Godot version

4.1.beta1

### System information

Godot v4.1.beta1 - Windows 10.0.22621 - Vulkan (Forward+) - dedicated AMD Radeon RX 5700 XT (Advanced Micro Devices, Inc.; 31.0.14057.2005) - AMD Ryzen 7 5800X 8-Core Processor (16 Threads)

### Issue description

I have no idea what caused this, so not sure how this could be tracked down. But, figured it is worth noting since I couldn't find an issue opened already.

I have a rather simple project with a few nodes and Tilemap. After a long time of working on this project today, I went back to paint a new physics layer using the Tileset > Paint feature. However it was greyed out, and no amount of switching tabs, moving windows, or flicking through the inspector or scene tabs did anything.

Screenshot 2023-06-12 192030

This picture shows the paint button not working.

I did nothing but close Godot and reopen it, and it began working again.

Screenshot 2023-06-12 192112

Is this intended, and did I somehow activate something that disabled the ability to Paint, or is this a bug?

### Steps to reproduce

No detailed steps, unfortunately. It just randomly became greyed out after not opening the TileSet or TileMap windows for a long time and would not become active until restarting Godot.

N/A because I couldn't tell you what caused this

TheSofox:
Fixes #78162

When `TileSet` becomes read only (such as multiple scenes using it at the same time), the Paint Button becomes disabled. Apparently this behaviour is not properly communicated to users. This PR changes the tooltip of the Paint Button when enabled/disabled to communicate this behaviour better to the user. When enabled the tooltip is the normal "Paint properties.", however now when disabled the tooltip is "TileSet is in read-only mode. Make the resource unique to edit TileSet properties."

### DIFF
--- a/editor/plugins/tiles/tile_set_atlas_source_editor.cpp
+++ b/editor/plugins/tiles/tile_set_atlas_source_editor.cpp
@@ -1046,6 +1046,13 @@ void TileSetAtlasSourceEditor::_update_toolbar() {
 	}
 }
 
+void TileSetAtlasSourceEditor::_update_buttons() {
+	tool_paint_button->set_disabled(read_only);
+	tool_paint_button->set_tooltip_text(read_only ? TTR("TileSet is in read-only mode. Make the resource unique to edit TileSet properties.") : TTR("Paint properties."));
+	tools_settings_erase_button->set_disabled(read_only);
+	tool_advanced_menu_button->set_disabled(read_only);
+}
+
 void TileSetAtlasSourceEditor::_tile_atlas_control_mouse_exited() {
 	hovered_base_tile_coords = TileSetSource::INVALID_ATLAS_COORDS;
 	tile_atlas_control->queue_redraw();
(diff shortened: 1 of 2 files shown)


groud:
Hey, thanks for the contribution!

I think the idea is quite good, and the implementation is acceptable, but it is a bit unusual to define strings as variables like that.  It makes things a bit more difficult to track IMO. I think I would suggest, instead doing it that way, to create some sort of an `update_readonly()` method, and update buttons/tooltips/etc... there.
That would avoid having to define those strings elsewhere.

TheSofox:
Good feedback, I've implemented it by adding a `_update_buttons_enabled()` function and removing the string variables.

TheSofox:
Renamed function `_update_buttons()` since it just seemed neater.
The function also contains other bits of repeated code, so I think it cleans up the code overall.

groud (approved):

akien-mga:
Thanks!

Issue #86784, opened 2024-01-04
PortableCompressedTexture2D.create_from_image fails in various ways.


API-Beast:
### Tested versions

- Reproducible in: 4.2.stable, 4.1.stable

### System information

Godot v4.2.stable.mono - Windows 10.0.19044 - GLES3 (Compatibility) - NVIDIA GeForce GTX 1060 6GB (NVIDIA; 31.0.15.3742) - Intel(R) Core(TM) i5-4670K CPU @ 3.40GHz (4 Threads)

### Issue description

PortableCompressedTexture2D.create_from_image generates errors in basically every configuration.

- LOSSLESS, source image has no mip maps: ⚠ Console shows errors
- LOSSLESS, source image has mip maps: ❌ Texture is unusable
- LOSSY, source image has no mip maps: ⚠ Console shows errors
- LOSSY, source image has mip maps: ❌ Texture is unusable
- BASIS_UNIVERSAL: ⚠ Console shows errors

---------------

## create_from_image LOSSLESS without mip-maps.

⚠ Resulting PortableCompressedTexture2D is valid but multiple errors are generated.

```swift
var img = Image.load_from_file(sheet_path)
var sheet = PortableCompressedTexture2D.new()
sheet.create_from_image(img, PortableCompressedTexture2D.COMPRESSION_MODE_LOSSLESS)
```

> WARNING: Not a PNG file
>      at: check_error (drivers/png/png_driver_common.cpp:56)
> ERROR: Condition "!success" is true. Returning: ERR_FILE_CORRUPT
>    at: png_to_image (drivers/png/png_driver_common.cpp:69)
> ERROR: Condition "err" is true. Returning: Ref<Image>()
>    at: load_mem_png (drivers/png/image_loader_png.cpp:64)
> ERROR: It's not a reference to a valid Image object.
>    at: (./core/io/image.h:433)
> ERROR: Condition "err" is true. Returning: Ref<Image>()
>    at: _jpegd_mem_loader_func (modules/jpg/image_loader_jpegd.cpp:132)
> ERROR: It's not a reference to a valid Image object.
>    at: (./core/io/image.h:433)
> ERROR: Condition "p_width <= 0 || p_width > 16384" is true.
>    at: texture_set_size_override (drivers/gles3/storage/texture_storage.cpp:1117)

## create_from_image LOSSLESS with mip-maps.

❌ Resulting PortableCompressedTexture2D is invalid. 

```swift
[...]

PR #86950, opened 2024-01-08
Fix wrong fail condition in compressed texture's `_set_data`

jsjtxietian:
Fixes "Condition "mipsize < data_size" is true." mentioned in https://github.com/godotengine/godot/issues/86784

Along with some clean up.


akien-mga (approved):
Makes sense.

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #86952, opened 2024-01-08
Fix selecting popup menu items on mouse release (reverted)

ryevdokimov:
I'm not sure this ever worked, since I don't think InputEventMouseButton registers in windows on mouse up and messing with the input code is probably a bad idea unless there is a very good reason.

This PR should fix it though. Regular mouse click behavior is maintained (keep menu open). 

image
image

### DIFF
--- a/scene/gui/popup_menu.cpp
+++ b/scene/gui/popup_menu.cpp
@@ -438,6 +438,9 @@ void PopupMenu::_input_from_window(const Ref<InputEvent> &p_event) {
 }
 
 void PopupMenu::_input_from_window_internal(const Ref<InputEvent> &p_event) {
+	Ref<InputEventMouseButton> b = p_event;
+	Ref<InputEventMouseMotion> m = p_event;
+
 	if (!items.is_empty()) {
 		Input *input = Input::get_singleton();
 		Ref<InputEventJoypadMotion> joypadmotion_event = p_event;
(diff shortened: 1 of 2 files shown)


ryevdokimov:
Something messed up with selecting checkboxes, looking into it.

ryevdokimov:
Ok, fixed and works with these checkboxes now.

image

scene/gui/popup_menu.cpp
```diff
@@ -566,52 +569,50 @@ void PopupMenu::_input_from_window_internal(const Ref<InputEvent> &p_event) {
...
 
-	if (b.is_valid()) {
-		if (!item_clickable_area.has_point(b->get_position())) {
+	if ((b.is_valid() && b->is_pressed()) || (!mouse_is_pressed && drag_to_press)) {
+		drag_to_press = false;
+
+		// Disable clicks under a time threshold to avoid selection right when opening the popup.
+		uint64_t now = OS::get_singleton()->get_ticks_msec();
+		uint64_t diff = now - popup_time_msec;
+		if (diff < 250) {
```
KoBeWi, line 589: This should take effect only when dragging. If you click twice very fast, the second click is not registered, which makes the menu feel bugged.

ryevdokimov: Really, I'd like to get rid of this entirely since it's typically bad UI design to have the cursor over an item when the menu is expanded, but there are instances where that exist in the editor (like language selection in editor settings, which I personally think should include a scroll container). The reason it was increased from 150 to 250 is because slightly dragging can accidently select items in these menus, and the difference in 100ms is not noticeable in normal circumstances. 

While it exists, I'll make it trigger on dragging. I agree.

ryevdokimov: "Speed Run" users assuming they don't just hotkey everything should be able to drag select things as fast as humanly possible, without the risk of affecting "Click-Click" users, is how I see it, but I think this PR is at least worth testing with the general public until other issues are resolved.

KoBeWi:
Bug: mouse wheel will select options, making it impossible to scroll long lists.

ryevdokimov:
Good catch, it's fixed now.

KoBeWi (approved):

akien-mga:
Thanks!

Issue #80346, opened 2023-08-06
Right Mouse Button Behavior is Inconsistent with Standard Windows Behavior


jtwine:
### Godot version

v4.1.1.stable.official [bd6af8e0e]  

### System information

Godot v4.1.1.stable - Windows 10.0.22621 - Vulkan (Compatibility) - NVIDIA GeForce GTX 1060 (NVIDIA; 30.0.14.7311) - Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz (12 Threads)

### Issue description

In Windows, mouse clicks that **activate** something like:
- Pressing a button
- Selecting a menu item
- Raising a context menu

All act on the _release_ of the button, not on the _press down_ of it.  You can test this yourself in any normal Windows application with a menu bar.  Activate one of the top-level menu items, and in the popup menu, _mouse down and hold_ on one of the items. You will not activate that menu item until you release the mouse button.  Additionally, you can mouse down on one item but then move to another one and then release the mouse button and you will activate the menu item the mouse is over at the time of the release.

Buttons have similar behavior - you can press and hold them, but they will not activate until you release the mouse button, and even then, only if you release the button while over the button.  This is what allows you to back out of an action if you accidently press the wrong button: all you have to do is move off of it before releasing the button and you will have effectively canceled the activation of that button.  (You may have even done this unconsciously yourself.)  Godot seems to have the behavior correct, BTW.

However, using the Right Mouse Button (RMB) to raise a context menu causes the menu to appear on the _down_ of the mouse button instead of the _release_ of it.  It also does the same thing with the context menu key on an extended keyboard.  The context menu should not appear until the mouse button is **released**, and the same goes for the content menu key.

Also, pressing the RMB during a Drag-and-Drop (D&D) operation does not work as expected, either.  The context menu appears but does not enter a standard _menu modal message loop_ so releasing the Left Mouse Button **commits** the D&D operation, but right clicking to activate a menu item causes the menu to reappear and breaks the connection between the mouse movement indicating the drop location in the editor as the caret location will update, but the drop location will be the caret's location when the context menu first appeared.

Pressing the RMB during a D&D operation should either _cancel_ the D&D operation, or it should be _ignored_ allowing the user to complete the D&D operation normally.  The latter is the standard behavior of most text editors in Windows, but for other D&D operations, using the _other mouse button_ will **cancel** the D&D operation.  You can verify this in apps like Windows Explorer and Outlook for example.

### Steps to reproduce

[...]

PR #86953, opened 2024-01-08
Remove the ability to open the top popup menus with right-click

ryevdokimov:
Related to #86952

This is kind of a weird behavior and is pretty none-standard across software I've seen. It also leads to weird bugs, where you can right click select check boxes, but only when the popup menu was opened with a right-click.

image

### DIFF
--- a/scene/gui/menu_bar.cpp
+++ b/scene/gui/menu_bar.cpp
@@ -102,7 +102,7 @@ void MenuBar::gui_input(const Ref<InputEvent> &p_event) {
 
 	Ref<InputEventMouseButton> mb = p_event;
 	if (mb.is_valid()) {
-		if (mb->is_pressed() && (mb->get_button_index() == MouseButton::LEFT || mb->get_button_index() == MouseButton::RIGHT)) {
+		if (mb->is_pressed() && (mb->get_button_index() == MouseButton::LEFT)) {
 			int index = _get_index_at_point(mb->get_position());
 			if (index >= 0) {
 				_open_popup(index);


lostminds:
On macOS you can open native top bar menus (and select items) with either left or right mouse button. While most users probably use left button I don't see the benefit of not allowing using the right mouse button for opening menus if someone prefers that. However, when I test it in the 4.2.1 editor the embedded godot menus there only seems to support opening with left mouse already?

If you'd like to block opening menus with right-click, perhaps something like the [BaseButton button_mask](https://docs.godotengine.org/en/stable/classes/class_basebutton.html#class-basebutton-property-button-mask) property could be a better solution, that way the user can configure what mouse buttons should be supported in a consistent way.

AThousandShips:
This should probably be discussed more in depth on a [proposal](https://github.com/godotengine/godot-proposals) about what way to go and the interest in such a change, it does break compatibility IMO as it removes existing features that people might use or want

ryevdokimov:
> This should probably be discussed more in depth on a [proposal](https://github.com/godotengine/godot-proposals) about what way to go and the interest in such a change, it does break compatibility IMO as it removes existing features that people might use or want

Hmm, this is technically a bug (or several) as well then. What I'm understanding and know is:

On mac the top menus can **only** open with left click, but mac users expect to be able to open and select with either? 

On windows you **can** open with a right click, but it's janky as it allows you to subsequently select an item with right click, but only once.

On windows as well, you **can't** open these popup menus with a right-click, so the behavior isn't consistent in the editor node. 

So, at least this PR would make the behavior consistent across the editor node, across platforms (I think?), and prevent some side effects. Not sure, it's a weird situation.

AThousandShips:
The menu bars in the engine doesn't have to match the system ones, and those other menus are not the same kind, so that's not really relevant to changing these, as in treating this as a bug, lack of consistency isn't a bug necessarily, that's why this needs to be discussed, this might not be the solution to this inconsistency, or a solution might not be desired at all

ryevdokimov:
> The menu bars in the engine doesn't have to match the system ones, and those other menus are not the same kind, so that's not really relevant to changing these, as in treating this as a bug, lack of consistency isn't a bug necessarily, that's why this needs to be discussed, this might not be the solution to this inconsistency, or a solution might not be desired at all
> 
> Also this _isn't_ just about the editor, so we have to consider all uses of the menu bar node

Good points. I'll think about this more.

Sauermann:
Related to https://github.com/godotengine/godot/issues/80346 and https://github.com/godotengine/godot-proposals/discussions/7463

PR #86957, opened 2024-01-08
A couple of enhancements to user-visible threading semantics

RandomShaper:
Kept in two commits because, regardless both related to threading, the changes are heavily unrelated otherwise.

I'd appreciate especially careful reviewing of my `GDVIRTUAL` changes, as this is the first time I deal with that.

### DIFF
--- a/doc/classes/EditorImportPlugin.xml
+++ b/doc/classes/EditorImportPlugin.xml
@@ -120,6 +120,13 @@
 		<link title="Import plugins">$DOCS_URL/tutorials/plugins/editor/import_plugins.html</link>
 	</tutorials>
 	<methods>
+		<method name="_can_import_threaded" qualifiers="virtual const">
+			<return type="bool" />
+			<description>
+				Tells whether this importer can be run in parallel on threads, or, on the contrary, it's only safe for the editor to call it from the main thread, for one file at a time.
+				If this method is not overridden, it will return [code]true[/code] by default (i.e., safe for parallel importing).
+			</description>
+		</method>
 		<method name="_get_import_options" qualifiers="virtual const">
 			<return type="Dictionary[]" />
 			<param index="0" name="path" type="String" />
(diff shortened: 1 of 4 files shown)


RandomShaper:
CC @SlugFiller

SlugFiller:
Cool! Nice to see things are happening quickly. Here's hoping it gets merged soon.

YuriSizov:
> Here's hoping it gets merged soon.

Did you test it, do you approve it? 🙃

SlugFiller:
With just the first commit, I hit the same deadlock in running `ResourceSaver.save` in parallel.

With both commits, and `_can_import_threaded` set to return `false`, it works perfectly. However, if I don't implement `_can_import_threaded`, Godot poofs on import. So something is definitely not right about the `GDVIRTUAL_CALL`.

RandomShaper:
@SlugFiller, thank you for the feedback. I'm wondering, what happens if you override but return `true`?

SlugFiller:
When returning `true` it works with single, or deadlocks on `ResourceSaver.save` if done in parallel. If I don't override it, it poofs even on single.

RandomShaper:
I've fixed a very silly mistake causing infinite recursion if the method is not overridden. Please test again with `false`. The problem with parallel saving will have to be investigated and fixed separately as it's not in the scope of this PR.

SlugFiller:
Poofing is gone, and returning `false` works for importing 3 scenes at once. Seems great.

doc/classes/EditorImportPlugin.xml
```diff
@@ -120,6 +120,13 @@
 		<link title="Import plugins">$DOCS_URL/tutorials/plugins/editor/import_plugins.html</link>
 	</tutorials>
 	<methods>
+		<method name="_can_import_threaded" qualifiers="virtual const">
+			<return type="bool" />
+			<description>
+				Tells whether this importer can be run in parallel on threads, or, on the contrary, it's only safe for the editor to call it from the main thread, for one file at a time.
+				If this method it not overridden, it will return [code]true[/code] by default (i.e., safe for parallel importing).
```
akien-mga, line 127:
```suggestion
				If this method is not overridden, it will return [code]true[/code] by default (i.e., safe for parallel importing).
```

akien-mga (approved):

Issue #86935, opened 2024-01-07
AudioStream's internal `get_stream_name` seems to be entirely unused, rendering `_get_stream_name` pointless.


Mickeon:
### Tested versions

Starting from 4.0

### System information

Windows 10

### Issue description

Opening this issue before doing something hasty.

In the source code, it seems like **AudioStream**'s `get_stream_name` is overridden properly for almost every inherited **AudioStream** type. However, outside of the tests, this method is not ever actually called.

This also makes the exposed, overridable method `_get_stream_name` entirely pointless, as the returned name is not ever used for anything.

The simplest solution to this issue is to expose `get_stream_name`

### Steps to reproduce

This requires a look at the source code. Find every reference of `get_stream_name`. Every mention is an appropriate override of the original, but it is never called for any actual purpose.

### Minimal reproduction project (MRP)

No need.

PR #86927, opened 2024-01-07
Fill the documentation of a few overlooked classes

Mickeon:
This PR completely fills the documentation of **AudioEffect**, **AudioEffectInstance**, **MissingNode** and **MissingResource**. 

For **AudioEffect** and **AudioEffectInstance**, the description has been overhauled with appropriate links to each other, AudioServer and the oddly missing tutorial section, and so maaaaany missing details
<details>
<summary>Hmm...</summary>

</details>

For MissingNode and MissingResource, it adds missing details in their behavior.

100% completion, here we go.

PR #86958, opened 2024-01-08
Add missing documentation for AudioStream & AudioStreamPlayback

Mickeon:
Related to https://github.com/godotengine/godot/pull/86927

This PR completes the documentation of **AudioServer**, **AudioStream** and **AudioStreamPlayback**, filling all of the undocumented methods. 
The existing descriptions where mostly left untouched <sub>even though they're quite poor</sub>.

Some doubts are still there, but it's still way better than leaving them as the "mystery meat" methods. Actually writing for them has you realise how [flawed some of these are](https://github.com/godotengine/godot/issues/86935).

### DIFF
--- a/doc/classes/AudioStream.xml
+++ b/doc/classes/AudioStream.xml
@@ -16,31 +16,39 @@
 		<method name="_get_beat_count" qualifiers="virtual const">
 			<return type="int" />
 			<description>
+				Overridable method. Should return the total number of beats of this audio stream. Used by the engine to determine the position of every beat.
+				Ideally, the returned value should be based off the stream's sample rate ([member AudioStreamWAV.mix_rate], for example).
 			</description>
 		</method>
 		<method name="_get_bpm" qualifiers="virtual const">
 			<return type="float" />
 			<description>
+				Overridable method. Should return the tempo of this audio stream, in beats per minute (BPM). Used by the engine to determine the position of every beat.
+				Ideally, the returned value should be based off the stream's sample rate ([member AudioStreamWAV.mix_rate], for example).
 			</description>
 		</method>
 		<method name="_get_length" qualifiers="virtual const">
 			<return type="float" />
 			<description>
+				Override this method to customize the returned value of [method get_length]. Should return the length of this audio stream, in seconds.
 			</description>
 		</method>
 		<method name="_get_stream_name" qualifiers="virtual const">
 			<return type="String" />
 			<description>
+				Override this method to customize the name assigned to this audio stream. Unused by the engine.
 			</description>
 		</method>
 		<method name="_instantiate_playback" qualifiers="virtual const">
 			<return type="AudioStreamPlayback" />
 			<description>
+				Override this method to customize the returned value of [method instantiate_playback]. Should returned a new [AudioStreamPlayback] created when the stream is played (such as by an [AudioStreamPlayer])..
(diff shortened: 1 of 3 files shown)


doc/classes/AudioStreamPlayback.xml
```diff
@@ -31,28 +34,34 @@
 			<param index="1" name="rate_scale" type="float" />
 			<param index="2" name="frames" type="int" />
 			<description>
+				Override this method to customize how the audio stream is mixed while the playback is active.
```
RedMser, line 37:
```suggestion
				Override this method to customize how the audio stream is mixed.
```
Mixing seems to happen even if `is_playing()` returns false.

doc/classes/AudioStreamPlayback.xml
```diff
@@ -31,28 +34,34 @@
...
 		<method name="_stop" qualifiers="virtual">
 			<return type="void" />
 			<description>
+				Override this method to customize what happens when the playback is stopped, such as by calling [method AudioStreamPlayer.stop].
 			</description>
 		</method>
 		<method name="_tag_used_streams" qualifiers="virtual">
 			<return type="void" />
 			<description>
+				Called if [method AudioServer.set_enable_tagging_used_audio_streams] has been set to [code]true[/code].
```
RedMser, line 64:
```suggestion
				Called whenever the audio stream is mixed if [method AudioServer.set_enable_tagging_used_audio_streams] has been set to [code]true[/code] and the playback is active.
```

Include info about it being called after every mix step, and only if playback is active.

RedMser (commented):
From a quick look into `AudioServer::_mix_step` there were some things that did not seem right to me.

Mickeon:
Updated this PR to address the above, and filled the descriptions of both `_tag_used_streams` and `set_enable_tagging_used_audio_streams` based on further insight https://chat.godotengine.org/channel/audio?msg=nh7hTna84fNH7vcPJ

doc/classes/AudioStream.xml
```diff
@@ -16,31 +16,39 @@
 		<method name="_get_beat_count" qualifiers="virtual const">
 			<return type="int" />
 			<description>
+				Should return the total number of beats of this audio stream. Used by the engine to determine the position of every beat.
```
YuriSizov, line 19: I think we should always explicitly tell the user that virtual methods need to be implemented by them. Without this clarification a description starting with "Should return" looks concerning, as if we don't know exactly if this method returns what we think it returns.

Same below.

Mickeon: The usual "_Override this method to[...]_" cannot really work here otherwise it would be too verbose.

I would introduce them with "_Overridable method._" personally. I don't like to say "_Virtual method to be implemented by the user_" because it feels very redundant, but that's obviously just an opinion. `virtual` is right there, you can see what it does, and all virtual methods begin with an underscore. Users should be used by this, especially if they're trying to make their own audio streams.

Mickeon:
Updated the PR to add "Overridable method" or similar to the methods that don't explicitly start with "Override" or similar.

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

PR #86960, opened 2024-01-08
Implement Unique Node IDs

reduz:
* Scene instantiation and inheritance now use unique IDs for referencing.
* Ensures that if a node is moved or renamed, **the scene that inherits keeps functioning by still referencing that node properly (which it does by ID, no longer by path)**.
* Eventually will be extended to Asset IO

**WARNING**: THIS NEEDS A LOT OF TESTING AND REVIEW - DO NOT USE IN YOUR PROJECT, IT WILL BREAK IT.

Note: This bumps the binary format version, files saved with this will no longer work in earlier versions of Godot.

Note: ResourceFormatText::save_as_binary is broken with this, but this function is no longer in use, should it be removed as part of this PR?

Note: I am allowing negative numbers in there for IDs (only 0 is unassigned), I may as well not, let me know if you want me to change this.

### DIFF
--- a/core/io/resource_uid.cpp
+++ b/core/io/resource_uid.cpp
@@ -242,6 +242,16 @@ void ResourceUID::clear() {
 	unique_ids.clear();
 	changed = false;
 }
+
+uint32_t ResourceUID::generate_random_u32() const {
+	uint32_t num;
+	Error err = ((CryptoCore::RandomGenerator *)crypto)->get_random_bytes((uint8_t *)&num, sizeof(uint32_t));
+	if (err != OK) {
+		num = 0;
+	}
+	return num;
+}
+
 void ResourceUID::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("id_to_text", "id"), &ResourceUID::id_to_text);
 	ClassDB::bind_method(D_METHOD("text_to_id", "text_id"), &ResourceUID::text_to_id);
--- a/scene/resources/packed_scene.cpp
+++ b/scene/resources/packed_scene.cpp
@@ -33,6 +33,7 @@
 #include "core/config/engine.h"
 #include "core/config/project_settings.h"
 #include "core/core_string_names.h"
+#include "core/crypto/crypto_core.h"
 #include "core/io/missing_resource.h"
 #include "core/io/resource_loader.h"
 #include "core/templates/local_vector.h"
@@ -43,7 +44,9 @@
 #include "scene/main/missing_node.h"
 #include "scene/property_utils.h"
 
-#define PACKED_SCENE_VERSION 3
+#define PACKED_SCENE_VERSION 4
+
+#define MIN_UNIQUE_ID_PACKED_SCENE_VERSION 4
 
 #ifdef TOOLS_ENABLED
 SceneState::InstantiationWarningNotify SceneState::instantiation_warn_notify = nullptr;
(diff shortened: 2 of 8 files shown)


scene/resources/packed_scene.cpp
```diff
@@ -1589,7 +1667,94 @@ Vector<StringName> SceneState::get_node_groups(int p_idx) const {
...
+				int32_t subindex_id = String::to_int(path_names[i].operator String().ptr() + 2);
+				int node_index = -1;
+
+				while (node_index == -1) {
+					node_index = base_state->_find_node_by_id(subindex_id);
+					if (base_state->base_scene_idx != -1) {
+						// There is inheritance at play, try with parent.
+						Ref<PackedScene> parent = base_state->variants[base_state->base_scene_idx];
+						if (parent.is_valid()) {
+							base_state = parent->get_state().ptr();
```
reduz, line 1727: This part here is pretty odd. It means if it did not find the node in the local scene, it will try with the scene that it inherits from. I am not sure if this is 100% correct, but It's the only way I can think of that works with inheritance.

reduz:
I think what I may need to change is that if you instantiate a scene (not inherit it), the ID of the instance should probably be different than that of the scene when opened separatedly. I already change it if instantiated more than once, but may need to make an exception here by forcing the node ID to unassigned when called with GEN_EDIT_STATE_INSTANCE.

core/io/resource_uid.cpp
```diff
@@ -242,6 +242,16 @@ void ResourceUID::clear() {
 	unique_ids.clear();
 	changed = false;
 }
+
+uint32_t ResourceUID::generate_random_u32() const {
+	uint32_t num;
+	Error err = ((CryptoCore::RandomGenerator *)crypto)->get_random_bytes((uint8_t *)&num, sizeof(uint32_t));
```
fire, line 248: Will this ever change to a int64? I've seen other systems like https://en.wikipedia.org/wiki/Snowflake_ID use 64 bits.

Mickeon: I have a feeling 4,294,967,296 nodes in a single PackedScene is not an amount that will be reached any time soon.

fire: I didn't know we're incrementing. We are using random bytes here.

Mickeon: Oh, then, it may actually be an issue, but unsure.

lawnjelly: It looks like where it is called, it checks for duplicates and calls again until a unique is found, so collisions shouldn't be a problem (aside from the time spent checking for duplicates).

reduz: Probably no, this is only for a single scene so no need for so much.

KoBeWi:
I opened a scene (tscn), it added a bunch of IDs to nodes, but only that. I'd expect nodes to use these IDs instead of parent path, but that's not the case. I checked some simple inheritance case (added node in inherited scene and moved its parent in the base scene) and it still breaks.

reduz:
@KoBeWi it won't use paths within the local scene because it does not make much sense (and path is more readable). IDs are used mostly when you do scene inheritance and place nodes as children of nodes of other scenes. They are also used for signal connections.

KoBeWi:
Ok I tested again and indeed, parent is always a path, even if the node is under inherited node or editable child. Both in tscn and scn. The only case that works correctly is binary inherited scene.

reduz:
@KoBeWi Maybe I broke it, thanks still for giving it a spin, will see if I can fix it today!

akien-mga:
Superseded by #106837.

PR #86961, opened 2024-01-08
Allow specifying a scene in completion tests

HolonProduction:
Followup to #85178.

- Adds the option to specify a scene path in the test cfg file. This will come in handy when performing many tests against the same scene (e.g. testing everything related to `get_node`).
- Fixes an oversight in the original PR, which meant that scene files would not correctly work in sub folders.
- Solve inconsistencies in key naming for the cfg file. (They were different between code and example test).
- Also documents the format for completion tests inside the test README


vnen (approved):

YuriSizov:
Thanks!

PR #51156, opened 2021-08-01
Add const lvalue ref to container parameters

Muller-Castro:
Adds const lvalue reference to the core container types specified in the [documentation](https://docs.godotengine.org/en/stable/development/cpp/core_types.html) that are declared in function parameters.

`core/*`: https://github.com/godotengine/godot/pull/86966
`editor/*`: https://github.com/godotengine/godot/pull/88368
`modules/*`: https://github.com/godotengine/godot/pull/88915
`platform/*`: https://github.com/godotengine/godot/pull/88971
`servers/*`: https://github.com/godotengine/godot/pull/88972
`scene/*`: https://github.com/godotengine/godot/pull/88974
others: https://github.com/godotengine/godot/pull/88975

PR #86966, opened 2024-01-08
Add const lvalue ref to `core/*` container parameters

Muller-Castro:
Adds const lvalue reference to the core container types specified in the [documentation](https://docs.godotengine.org/en/stable/development/cpp/core_types.html) that are declared in function parameters in `core/*` files

initial: https://github.com/godotengine/godot/pull/51156
`editor/*`: https://github.com/godotengine/godot/pull/88368
`modules/*`: https://github.com/godotengine/godot/pull/88915
`platform/*`: https://github.com/godotengine/godot/pull/88971
`servers/*`: https://github.com/godotengine/godot/pull/88972
`scene/*`: https://github.com/godotengine/godot/pull/88974
others: https://github.com/godotengine/godot/pull/88975

### DIFF
--- a/core/core_bind.cpp
+++ b/core/core_bind.cpp
@@ -257,7 +257,7 @@ String OS::get_executable_path() const {
 	return ::OS::get_singleton()->get_executable_path();
 }
 
-Error OS::shell_open(String p_uri) {
+Error OS::shell_open(const String &p_uri) {
 	if (p_uri.begins_with("res://")) {
 		WARN_PRINT("Attempting to open an URL with the \"res://\" protocol. Use `ProjectSettings.globalize_path()` to convert a Godot-specific path to a system path before opening it with `OS.shell_open()`.");
 	} else if (p_uri.begins_with("user://")) {
@@ -266,7 +266,7 @@ Error OS::shell_open(String p_uri) {
 	return ::OS::get_singleton()->shell_open(p_uri);
 }
 
-Error OS::shell_show_in_file_manager(String p_path, bool p_open_folder) {
+Error OS::shell_show_in_file_manager(const String &p_path, bool p_open_folder) {
 	if (p_path.begins_with("res://")) {
 		WARN_PRINT("Attempting to explore file path with the \"res://\" protocol. Use `ProjectSettings.globalize_path()` to convert a Godot-specific path to a system path before opening it with `OS.shell_show_in_file_manager()`.");
 	} else if (p_path.begins_with("user://")) {
--- a/core/io/dir_access.cpp
+++ b/core/io/dir_access.cpp
@@ -131,7 +131,7 @@ Error DirAccess::erase_contents_recursive() {
 	return _erase_recursive(this);
 }
 
-Error DirAccess::make_dir_recursive(String p_dir) {
+Error DirAccess::make_dir_recursive(const String &p_dir) {
 	if (p_dir.length() < 1) {
 		return OK;
 	}
@@ -188,7 +188,7 @@ DirAccess::AccessType DirAccess::get_access_type() const {
 	return _access_type;
 }
 
-String DirAccess::fix_path(String p_path) const {
+String DirAccess::fix_path(const String &p_path) const {
 	switch (_access_type) {
 		case ACCESS_RESOURCES: {
 			if (ProjectSettings::get_singleton()) {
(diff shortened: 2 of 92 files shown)


AThousandShips:
This doesn't just cover core, you need to split it down further place, let's keep it to just `core/`

core/core_bind.cpp
```diff
@@ -52,11 +52,13 @@ Error ResourceLoader::load_threaded_request(const String &p_path, const String &
 	return ::ResourceLoader::load_threaded_request(p_path, p_type_hint, p_use_sub_threads, ResourceFormatLoader::CacheMode(p_cache_mode));
 }
 
-ResourceLoader::ThreadLoadStatus ResourceLoader::load_threaded_get_status(const String &p_path, Array r_progress) {
+ResourceLoader::ThreadLoadStatus ResourceLoader::load_threaded_get_status(const String &p_path, const Array &r_progress) {
```
AThousandShips, line 55: Not valid, this should be:
```suggestion
ResourceLoader::ThreadLoadStatus ResourceLoader::load_threaded_get_status(const String &p_path, Array &r_progress) {
```
If changed at all

core/core_bind.cpp
```diff
@@ -279,15 +281,17 @@ String OS::read_string_from_stdin() {
 	return ::OS::get_singleton()->get_stdin_string();
 }
 
-int OS::execute(const String &p_path, const Vector<String> &p_arguments, Array r_output, bool p_read_stderr, bool p_open_console) {
+int OS::execute(const String &p_path, const Vector<String> &p_arguments, const Array &r_output, bool p_read_stderr, bool p_open_console) {
```
AThousandShips, line 284:
```suggestion
int OS::execute(const String &p_path, const Vector<String> &p_arguments, Array &r_output, bool p_read_stderr, bool p_open_console) {
```

AThousandShips:
Please limit it to **just** core, this is still too big to do a proper review of for these kinds of changes 🙂 no harm in multiple PRs, especially when it makes it possible to review

Muller-Castro:
I tried to just upload the `core/*` files but it was issuing build errors because of `core/*` virtual methods overridden across multiple parts of the engine

AThousandShips:
Then those should be handled separately in their own PR because this still has over 300 files, far too big for a review like this, for those specific methods you can do a dedicated PR

Muller-Castro:
Ok I'll leave the virtual methods that are issuing build errors out of this PR and put then in another PR

AThousandShips:
Or at minimum remove all unrelated changes from other parts, because they make up a lot of it 🙂

Muller-Castro:
You mean remove all folders except `core` and core's virtual methods that aren't compiling?

core/io/dir_access.cpp
```diff
@@ -460,28 +460,30 @@ Error DirAccess::_copy_dir(Ref<DirAccess> &p_target_da, String p_to, int p_chmod
 	return OK;
 }
 
-Error DirAccess::copy_dir(String p_from, String p_to, int p_chmod_flags, bool p_copy_links) {
+Error DirAccess::copy_dir(const String &p_from, const String &p_to, int p_chmod_flags, bool p_copy_links) {
```
AThousandShips, line 463: For now I'd remove any case that requires copying in the body, let's keep this to just const changes, and possibly changing the iterators if needed

AThousandShips (commented):
Initial glance, will go over properly when the scope has been reduced 🙂

core/io/dir_access.cpp
Muller-Castro: Got it!

Muller-Castro:
Those non-`core/*` are inevitable for all `core/*` files to build

AThousandShips:
Will take a look, thank you and sorry for the inconvenience (and you'll get a bonus of more PRs!)

core/variant/method_ptrcall.h
AThousandShips, line 1: I need a second eye on this, as it's very core and specific, unsure about possible side effects, they're my only question mark as it were, should be fine but it's so central that possible side effects could be major

AThousandShips (approved):
Other than the note below this looks good and safe to me

AThousandShips:
You have a merge commit in your history, please use rebasing instead, do this by:
```
git reset --hard HEAD^
git rebase -i master
git push --force
```

akien-mga:
Sorry for the delay. Looks good to me overall, needs a rebase and then this should be ready to merge if nobody has any outstanding concerns about the change.

AThousandShips:
I'd like an eye by a core maintainer familiar with the `method_ptrcall.h` file, just to ensure there's no weirdness with that, otherwise I think this looks good

CC @reduz perhaps

dsnopek:
The changes in `method_ptrcall.h` seem fine to me! The `const` is added to `Vector`'s that we're only reading from.

akien-mga:
Thanks!

PR #86968, opened 2024-01-08
Add class reference documentation for GDExtension & GDExtensionManager

Mickeon:
This PR fills in the class reference of **GDExtension** and **GDExtensionManager**. It also links to the existing tutorials.

I don't know what else to say.

More feedback and useful insight is necessary from actual GDExtension developers. A few of these descriptions are unfortunately still barren.

### DIFF
--- a/doc/classes/GDExtension.xml
+++ b/doc/classes/GDExtension.xml
@@ -1,31 +1,42 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <class name="GDExtension" inherits="Resource" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
 	<brief_description>
+		A native library for GDExtension.
 	</brief_description>
 	<description>
+		The [GDExtension] resource type represents a [url=https://en.wikipedia.org/wiki/Shared_library]shared library[/url] which can expand the functionality of the engine. The [GDExtensionManager] singleton is responsible for loading, reloading, and unloading [GDExtension] resources.
+		[b]Note:[/b] GDExtension itself is not a scripting language and has no relation to [GDScript] resources.
 	</description>
 	<tutorials>
+		<link title="GDExtension overview">$DOCS_URL/tutorials/scripting/gdextension/what_is_gdextension.html</link>
+		<link title="GDExtension example in C++">$DOCS_URL/tutorials/scripting/gdextension/gdextension_cpp_example.html</link>
 	</tutorials>
 	<methods>
 		<method name="close_library">
 			<return type="void" />
 			<description>
+				Closes the current library.
+				[b]Note:[/b] You normally should not call this method directly. This is handled automatically by [method GDExtensionManager.unload_extension].
 			</description>
 		</method>
 		<method name="get_minimum_library_initialization_level" qualifiers="const">
 			<return type="int" enum="GDExtension.InitializationLevel" />
 			<description>
+				Returns the lowest level required for this extension to be properly initialized (see the [enum InitializationLevel] enum).
 			</description>
 		</method>
 		<method name="initialize_library">
 			<return type="void" />
 			<param index="0" name="level" type="int" enum="GDExtension.InitializationLevel" />
 			<description>
+				Initializes the library bound to this GDextension at the given initialization [param level].
+				[b]Note:[/b] You normally should not call this method directly. This is handled automatically by [method GDExtensionManager.load_extension].
 			</description>
 		</method>
 		<method name="is_library_open" qualifiers="const">
 			<return type="bool" />
 			<description>
+				Returns [code]true[/code] if this extension's library has been opened.
 			</description>
 		</method>
 		<method name="open_library">
@@ -33,17 +44,23 @@
 			<param index="0" name="path" type="String" />
 			<param index="1" name="entry_symbol" type="String" />
 			<description>
+				Opens the library at the specified [param path].
+				[b]Note:[/b] You normally should not call this method directly. This is handled automatically by [method GDExtensionManager.load_extension].
 			</description>
 		</method>
 	</methods>
 	<constants>
 		<constant name="INITIALIZATION_LEVEL_CORE" value="0" enum="InitializationLevel">
+			The library is initialized at the same time as the core features of the engine.
 		</constant>
 		<constant name="INITIALIZATION_LEVEL_SERVERS" value="1" enum="InitializationLevel">
+			The library is initialized at the same time as the engine's servers (such as [RenderingServer] or [PhysicsServer3D]).
 		</constant>
 		<constant name="INITIALIZATION_LEVEL_SCENE" value="2" enum="InitializationLevel">
+			The library is initialized at the same time as the engine's scene-related classes.
 		</constant>
 		<constant name="INITIALIZATION_LEVEL_EDITOR" value="3" enum="InitializationLevel">
+			The library is initialized at the same time as the engine's editor classes. Only happens when loading the GDExtension in the editor.
 		</constant>
 	</constants>
 </class>
--- a/doc/classes/GDExtensionManager.xml
+++ b/doc/classes/GDExtensionManager.xml
@@ -1,65 +1,82 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <class name="GDExtensionManager" inherits="Object" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
 	<brief_description>
+		Provides access to GDExtension functionality.
 	</brief_description>
 	<description>
+		The GDExtensionManager loads, initializes, and keeps track of all available [GDExtension] libraries in the project.
+		[b]Note:[/b] Do not worry about GDExtension unless you know what you are doing.
 	</description>
 	<tutorials>
+		<link title="GDExtension overview">$DOCS_URL/tutorials/scripting/gdextension/what_is_gdextension.html</link>
+		<link title="GDExtension example in C++">$DOCS_URL/tutorials/scripting/gdextension/gdextension_cpp_example.html</link>
 	</tutorials>
 	<methods>
 		<method name="get_extension">
 			<return type="GDExtension" />
 			<param index="0" name="path" type="String" />
 			<description>
+				Returns the [GDExtension] at the given file [param path], or [code]null[/code] if it has not been loaded or does not exist.
 			</description>
 		</method>
 		<method name="get_loaded_extensions" qualifiers="const">
 			<return type="PackedStringArray" />
 			<description>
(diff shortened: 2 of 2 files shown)


doc/classes/GDExtension.xml
```diff
@@ -1,49 +1,64 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <class name="GDExtension" inherits="Resource" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
 	<brief_description>
+		A shared native library for GDExtension.
```
dsnopek, line 4: Making these brief descriptions are is so hard :-)

While this is accurate, I feel like using fewer adjectives makes this a little easier to understand, especially for less technical folks:

```suggestion
		A native library for GDExtension.
```

doc/classes/GDExtension.xml
```diff
@@ -1,49 +1,64 @@
...
 	</brief_description>
 	<description>
+		The [GDExtension] resource type represents a set of libraries written for GDExtension. These libraries expand the existing functionality of the engine. They are handled by the [GDExtensionManager] singleton.
```
dsnopek, line 7: While the `.gdextension` file will list multiple libraries, a `GDExtension` resource will only ever represent a single one. Maybe:

```suggestion
		The [GDExtension] resource type represents a libraries written for GDExtension, which can expand the functionality of the engine. The [GDExtensionManager] singleton is responsible for loading, reloading and unloading [GDExtension] resources.
```

doc/classes/GDExtension.xml
```diff
@@ -1,49 +1,64 @@
...
 		<method name="close_library">
 			<return type="void" />
 			<description>
+				Closes the current library.
 			</description>
 		</method>
 		<method name="get_minimum_library_initialization_level" qualifiers="const">
 			<return type="int" enum="GDExtension.InitializationLevel" />
 			<description>
+				Returns the lowest level required for this extension to be properly initialized. Ideally, if this value is lower than the current level, this [GDExtension] should be restarted to ensure it is properly applied.
```
dsnopek, line 25: I don't know that we need the extra detail here:

```suggestion
				Returns the lowest level required for this extension to be properly initialized.
```

doc/classes/GDExtension.xml
```diff
@@ -1,49 +1,64 @@
...
 			</description>
 		</method>
 		<method name="initialize_library">
 			<return type="void" />
 			<param index="0" name="level" type="int" enum="GDExtension.InitializationLevel" />
 			<description>
+				Initializes the library bound to this GDextension.
```
dsnopek, line 32: Maybe:

```suggestion
				Initializes this GDExtension at the given initialization level.
```

This should probably also have a note saying that developers shouldn't call this under normal circumstance, it will be handled by `GDExtensionManager`.

I'm actually not entirely sure why we've exposed this at all? It's especially weird considering we haven't exposed `deinitialize_library()`, its paired method.

doc/classes/GDExtension.xml
```diff
@@ -1,49 +1,64 @@
...
 	</description>
 	<tutorials>
+		<link title="GDExtension overview">$DOCS_URL/tutorials/scripting/gdextension/what_is_gdextension.html</link>
+		<link title="GDExtension example in C++">$DOCS_URL/tutorials/scripting/gdextension/gdextension_cpp_example.html</link>
 	</tutorials>
 	<methods>
 		<method name="close_library">
 			<return type="void" />
 			<description>
+				Closes the current library.
```
dsnopek, line 19: This should probably have a note saying that developers shouldn't call this under normal circumstances, and instead use `GDExtensionManager.unload_extension()`.

doc/classes/GDExtension.xml
```diff
@@ -1,49 +1,64 @@
...
 	</methods>
 	<constants>
 		<constant name="INITIALIZATION_LEVEL_CORE" value="0" enum="InitializationLevel">
+			The library is initialized at the same time as core features of the engine.
 		</constant>
 		<constant name="INITIALIZATION_LEVEL_SERVERS" value="1" enum="InitializationLevel">
+			The library is initialized at the same time as the engine's servers (such as [RenderingServer] or [PhysicsServer3D]).
 		</constant>
 		<constant name="INITIALIZATION_LEVEL_SCENE" value="2" enum="InitializationLevel">
+			The library is initialized after the currently open scenes are loaded, if they exists.
```
dsnopek, line 58: This isn't accurate. Maybe:

```suggestion
			The library is initialized at the same time as the engine's scene-related classes.
```

doc/classes/GDExtension.xml
```diff
@@ -1,49 +1,64 @@
...
 		</constant>
 		<constant name="INITIALIZATION_LEVEL_EDITOR" value="3" enum="InitializationLevel">
+			The library is initialized after the editor has fully finished loading. Only available in the editor.
```
dsnopek, line 61: Also not accurate, maybe:

```suggestion
			The library is initialized at the same time as the engine's editor classes. Only happens when loading the GDExtension in the editor.
```

doc/classes/GDExtensionManager.xml
```diff
@@ -1,65 +1,82 @@
...
 			<param index="0" name="path" type="String" />
 			<description>
+				Returns [code]true[/code] if the extension at the given file [param path] has already been loaded successfully. See also [method get_loaded_extensions].
 			</description>
 		</method>
 		<method name="load_extension">
 			<return type="int" enum="GDExtensionManager.LoadStatus" />
 			<param index="0" name="path" type="String" />
 			<description>
+				Loads an extension by absolute file path. The [param path] needs to point to a valid [GDExtension]. Return [constant LOAD_STATUS_OK] if successful.
```
dsnopek, line 39:
```suggestion
				Loads an extension by absolute file path. The [param path] needs to point to a valid [GDExtension]. Returns [constant LOAD_STATUS_OK] if successful.
```

dsnopek (commented):
Thanks for being the first person brave enough to attempt to document these classes :-)

doc/classes/GDExtension.xml
Mickeon: Unfortunately, the extra detail may need to be mentioned somehow, since the user can use this method combined with the others about library initialization are exposed.

doc/classes/GDExtension.xml
Mickeon: > I'm actually not entirely sure why we've exposed this at all? It's especially weird considering we haven't exposed `deinitialize_library()`, its paired method.

It may be worth discussing this with the rest of the GDExtension team. There have been times before where Reduz & Co. accidentally exposed methods to the documentation & scripting (hence the name of the PR).
After a lot of time has passed, only then you can tell there's a flaw in the API when struggling to write documentation for these.

Mickeon:
Addressed all of the above feedback and suggestions. Thank you very much for the insight

doc/classes/GDExtension.xml
```diff
@@ -1,49 +1,68 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <class name="GDExtension" inherits="Resource" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
 	<brief_description>
+		A native library for GDExtension.
 	</brief_description>
 	<description>
+		The [GDExtension] resource type represents a set of libraries written for GDExtension. These libraries expand the existing functionality of the engine. They are handled by the [GDExtensionManager] singleton.
```
AThousandShips, line 7:
```suggestion
```
Accidentally left the original text in

doc/classes/GDExtensionManager.xml
```diff
@@ -1,65 +1,82 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <class name="GDExtensionManager" inherits="Object" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
 	<brief_description>
+		Provides access to GDExtension functionality.
 	</brief_description>
 	<description>
+		The GDExtensionManager loads, initializes and keeps track of all available [GDExtension] libraries in the project.
```
AThousandShips, line 7:
```suggestion
		The GDExtensionManager loads, initializes, and keeps track of all available [GDExtension] libraries in the project.
```

doc/classes/GDExtension.xml
```diff
@@ -1,49 +1,68 @@
...
+		The [GDExtension] resource type represents a [url=https://en.wikipedia.org/wiki/Shared_library]shared library[/url] written for GDExtension, which can expand the functionality of the engine. The [GDExtensionManager] singleton is responsible for loading, reloading and unloading [GDExtension] resources.
```
AThousandShips, line 8:
```suggestion
		The [GDExtension] resource type represents a [url=https://en.wikipedia.org/wiki/Shared_library]shared library[/url] written for GDExtension, which can expand the functionality of the engine. The [GDExtensionManager] singleton is responsible for loading, reloading, and unloading [GDExtension] resources.
```

AThousandShips (approved):
Looks good otherwise

Mickeon:
Thank you, I'm very happy to have this as soon as possible.

doc/classes/GDExtension.xml
dsnopek: I've added an item to the agenda of tonight's GDExtension meeting, to discuss if this method should even be bound in the first place

dsnopek (approved):
Thanks!

While I personally think we should probably unbind `GDExtension::initialize_library()`, and hence don't need docs for it, that doesn't necessarily need to hold up this PR.

This looks great!

doc/classes/GDExtension.xml
```diff
@@ -1,49 +1,66 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <class name="GDExtension" inherits="Resource" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
 	<brief_description>
+		A native library for GDExtension.
 	</brief_description>
 	<description>
+		The [GDExtension] resource type represents a [url=https://en.wikipedia.org/wiki/Shared_library]shared library[/url] written for GDExtension, which can expand the functionality of the engine. The [GDExtensionManager] singleton is responsible for loading, reloading, and unloading [GDExtension] resources.
```
paddy-exe, line 7:
```suggestion
		The [GDExtension] resource type represents a [url=https://en.wikipedia.org/wiki/Shared_library]shared library[/url] which can expand the functionality of the engine. The [GDExtensionManager] singleton is responsible for loading, reloading, and unloading [GDExtension] resources.
```

paddy-exe: I would just say a GDExtension resource expands the functionality as GDExtension is the underlying system enabling this. It is not necessarily written **for** it.

Mickeon:
Accepted the above suggestion

PR #86969, opened 2024-01-08
[GDScript] Fix lifetime of call base object

### LINKED ISSUE #86788: Player crashes randomly in Godot 4.2 on macOS
### Tested versions

- Reproducible in 4.2.1 stable

### System information

Godot v4.2.1.stable - macOS 14.1.0 - Vulkan (Forward+) - integrated Apple M3 Max - Apple M3 Max (14 Threads)

### Issue description

The editor player crashes randomly on this simple project. Sometimes I have no logs at all, sometimes I have the following error with no backtrace at all, and once I managed to have the backtrace printed.
EDIT: also crashes when running export build on macOS.

```
  Engine version: Godot Engine v4.2.1.stable.official (b09f793f564a6c95dc76acc654b390e68441bd01)
  [1] 1   libsystem_platform.dylib            0x0000000180e37a24 _sigtramp + 56
  [2] 2   ???                                 0xffff800180e840a8 0x0 + 18446603342678868136
  [3] 3   CoreFoundation                      0x0000000180e840a8 __CFBasicHashDrain + 356
  [4] 4   CoreFoundation                      0x0000000180fb7e28 _CFRelease + 292
  [5] 5   CoreFoundation                      0x0000000180eb82a4 __CFBinaryPlistWriteOrPresize + 308
  [6] 6   Foundation                          0x0000000181fda6e8 -[NSKeyedArchiver finishEncoding] + 640
  [7] 7   AppKit                              0x000000018481e634 __62-[NSPersistentUIArchiver finishEncodingWithCompletionHandler:]_block_invoke + 64
  [8] 8   AppKit                              0x000000018541cbb8 ___NSPersistentUIDispatchQueueAsync_block_invoke + 76
  [9] 9   libdispatch.dylib                   0x0000000180c59cb8 _dispatch_call_block_and_release + 32
  [10] 10  libdispatch.dylib                   0x0000000180c5b910 _dispatch_client_callout + 20
  [11] 11  libdispatch.dylib                   0x0000000180c62ea4 _dispatch_lane_serial_drain + 748
  [12] 12  libdispatch.dylib                   0x0000000180c639d4 _dispatch_lane_invoke + 380
  [13] 13  libdispatch.dylib                   0x0000000180c6e61c _dispatch_root_queue_drain_deferred_wlh + 288
  [14] 14  libdispatch.dylib                   0x0000000180c6de90 _dispatch_workloop_worker_thread + 404
  [15] 15  libsystem_pthread.dylib             0x0000000180e05114 _pthread_wqthread + 288
```

### Steps to reproduce

Run the main scene and move around with WASD.
Sometimes it will crash within a second, sometimes after a while.

[DemoProject.zip](https://github.com/godotengine/godot/files/13831974/DemoProject.zip)

AThousandShips:
Ensures that the lifetime of an `Object` outlasts calls to it

Also adds a check to `Object::~Object()` to check for locked state

This solves the immediate problem but might not be elegant, needs deeper investigation, but this should solve the problems

This makes member variables used as the base for a call persist until the end of the call, as a temporary

* Fixes: https://github.com/godotengine/godot/issues/86788 (needs confirmation)

### DIFF
--- a/core/object/object.cpp
+++ b/core/object/object.cpp
@@ -2050,6 +2050,10 @@ void Object::detach_from_objectdb() {
 }
 
 Object::~Object() {
+#ifdef DEBUG_ENABLED
+	CRASH_COND_MSG(_lock_index.get() != 1, "TODO message for object.");
+#endif // DEBUG_ENABLED
+
 	if (script_instance) {
 		memdelete(script_instance);
 	}
--- a/modules/gdscript/gdscript_compiler.cpp
+++ b/modules/gdscript/gdscript_compiler.cpp
@@ -708,11 +708,27 @@ GDScriptCodeGenerator::Address GDScriptCompiler::_parse_expression(CodeGen &code
 											gen->write_call_method_bind(result, base, method, arguments);
 										}
 									} else {
+										// Make temporary if member variable and ref-counted.
+										if (ClassDB::is_parent_class(class_name, "RefCounted") && base.mode == GDScriptCodeGenerator::Address::MEMBER) {
+											GDScriptCodeGenerator::Address temp = codegen.add_temporary(base.type);
+											gen->write_assign(temp, base);
+
+											base = temp;
+										}
+
 										gen->write_call(result, base, call->function_name, arguments);
 									}
 								} else if (base.type.has_type && base.type.kind == GDScriptDataType::BUILTIN) {
 									gen->write_call_builtin_type(result, base, base.type.builtin_type, call->function_name, arguments);
 								} else {
+									// Make temporary if member variable.
+									if (base.mode == GDScriptCodeGenerator::Address::MEMBER) {
+										GDScriptCodeGenerator::Address temp = codegen.add_temporary(base.type);
+										gen->write_assign(temp, base);
+
+										base = temp;
+									}
+
 									gen->write_call(result, base, call->function_name, arguments);
 								}
 								if (base.mode == GDScriptCodeGenerator::Address::TEMPORARY) {
--- a/modules/gdscript/tests/scripts/runtime/features/ref_counted_lifetime.gd
+++ b/modules/gdscript/tests/scripts/runtime/features/ref_counted_lifetime.gd
@@ -1,0 +1,26 @@
+# https://github.com/godotengine/godot/issues/86788
+
+extends Node
+
+class MyClassTyped:
+	func enter(outer):
+		outer.my_class_typed = MyClassTyped.new()
+		exit()
+
+	func exit():
+		pass
+
+class MyClassUnTyped:
+	func enter(outer):
+		outer.my_class_untyped = MyClassUnTyped.new()
+		exit()
+
+	func exit():
+		pass
+
+var my_class_typed := MyClassTyped.new()
+var my_class_untyped = MyClassUnTyped.new()
+
+func test():
+	my_class_typed.enter(self)
+	my_class_untyped.enter(self)
--- a/modules/gdscript/tests/scripts/runtime/features/ref_counted_lifetime.out
+++ b/modules/gdscript/tests/scripts/runtime/features/ref_counted_lifetime.out
@@ -1,0 +1,1 @@
+GDTEST_OK


core/object/object.cpp
```diff
@@ -1971,6 +1971,10 @@ void Object::detach_from_objectdb() {
 }
 
 Object::~Object() {
+#ifdef DEBUG_ENABLED
+	CRASH_COND_MSG(_lock_index.get() != 1, "TODO message for object.");
```
AThousandShips, line 1975: Any suggestions for the text here are welcome 🙂

modules/gdscript/gdscript_compiler.cpp
```diff
@@ -698,11 +698,27 @@ GDScriptCodeGenerator::Address GDScriptCompiler::_parse_expression(CodeGen &code
 											gen->write_call_method_bind(result, base, method, arguments);
 										}
 									} else {
+										// Make temporary if member variable.
```
AThousandShips, line 701: Unsure where to place this, if it's relevant for a greater part of this block, and if it should be more selective (or if we should make a helper function for it), etc.

modules/gdscript/gdscript_compiler.cpp
```diff
@@ -698,11 +698,27 @@ GDScriptCodeGenerator::Address GDScriptCompiler::_parse_expression(CodeGen &code
...
+										if (base.mode == GDScriptCodeGenerator::Address::MEMBER) {
```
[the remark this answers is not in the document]
AThousandShips: Added this for now

modules/gdscript/tests/scripts/runtime/features/ref_counted_lifetime.gd
AThousandShips, line 1: Added a test, this just confirms that it no longer crashes (it crashes with the check in `Object` and without the remaining code)

modules/gdscript/tests/scripts/runtime/features/ref_counted_lifetime.gd
```diff
@@ -0,0 +1,26 @@
+# GH-86788
```
AThousandShips, line 1:
```suggestion
# https://github.com/godotengine/godot/issues/86788
```
Will use this syntax when I go back over

AThousandShips:
I'd say the issue is not in variant as it handles the reference counting correctly, the issue is that the call base isn't copied but instead used directly for the call, making any change to it affect that, and in this case freeing it, `Variant` is oblivious to any of this as the call isn't on it directly, but on the pointer, at least as far as I can tell in my assessment

To me it looks like this issue is that when calling it does it specifically on the member variable, not the data *in* the member variable, which isn't correct in my opinion, there are some other cases like this:
* https://github.com/godotengine/godot/pull/88905

AThousandShips:
While we could add some way to mess with refcounting in the call operation itself I'd say that'd be a bit hacky and error prone, I'd say it's the responsibility of the caller to not delete the reference, assuming they can, in this case the compiler fails to do so properly, by not creating a copy to guarantee it

RandomShaper:
It's true that Godot tries to use pointers to `Variant` wherever it can, for performance. And that indeed can cause unexpected behaviors. The one you are mentioning about signals, etc. indeed seems to require taking the optimization away in the implementation to provide the right user-visible behavior. However, in this case, provided there's no notion of call-with-base-by-value in user land, we have further freedom to fix the issue at the implementation level, maybe without performance compromises.

Once we narrow it down to whether to frame GDScript or `Variant` as the flawed piece, the more I think about it, the less sure I am about which one.
- `Variant`: If we assume its contract with `RefCounted` involves letting them survive calls, we'd try to do apply, maybe somewhat hacky, but very performant and focused, solution there. It would benefit every user of `Variant`.
- GDScript: If we formulate a rule by which a `RefCounted` called through `Variant` can't be freed during the call (non-`RefCounted` wouldn't be an issue, would they?), then GDScript (as well as any other user of `Variant`) has to ensure proper management.

The latter is looking better to me the more I think about it. That's the rationale behind this PR, isn't it? If that's the idea, I wouldn't oppose to it. If anything, I'd also add assertions to `Variant` so it's more explicit in the code.

AThousandShips:
I'd say that unless you're careless in c++ land you won't encounter this, generally the assumption is that the caller ensures the lifetime of any pointer it sends, so with the guarantee in c++ of caller lifetime on values, i.e. `Variant(...).call(...)` is always alive until the end of that call, and assuming any `Variant *` passed to a method is not killed, it'll always be safe in c++ land

So to me the issue with option one there is that while it's performant, it isn't free, and the cost comes regardless if you handle things safely or not

So I'd say that generally, assuming you don't do things like passing pointers with uncertain lifetime from c++, it's safe here, and adding ref-count safety measures in `Variant` (or I'd say, in `RefCounted` to make it fully secure) would IMO create an overhead that's not needed if you're safe about things

There might be some even more elegant way to do it in GDScript land, perhaps something directly in the VM, but I'd say it's pretty cheap already, it just creates a temporary, and perhaps we could make even more filtering so it's only done when absolutely necessary (it isn't needed, for example, if the member is accessed by a getter, but this PR doesn't make that distinction as far as I recall, I'll go over the code again and see)

But good points and thank you for your analysis

tl;dr; I'd say option one would be useful, but it's mostly unnecessary assuming the use of `Variant` in c++ land is done safely, which applies generally not just to `Variant`, and you'd probably have to be a bit badly abusing the types to make it be a problem in c++

RandomShaper:
> I'd say that unless you're careless in c++ land you won't encounter this, generally the assumption is that the caller ensures the lifetime of any pointer it sends, so with the guarantee in c++ of caller lifetime on values, i.e. `Variant(...).call(...)` is always alive until the end of that call, and assuming any `Variant *` passed to a method is not killed, it'll always be safe in c++ land

Just to be sure I'm following the root issue here, the problem is not that a `Variant` was destroyed (as in it's memory being reassigned or its memory page unmapped), but that it was reassigned (i.e., `operator=`). So, the lifetime in question is that of the object it was pointing to.

In fact, and this is a piece of fresh news, I've just debugged the issue and it seems that the problem is `OBJ_DEBUG_LOCK` in `Object::callp()`. I mean, it may be the case that the code for object calls is designed so that the call itself is the last access to the object. If that's true, it's OK from the engine design standpoint (or it may have been the original idea, hard to tell) that an object is released during a call to it, only that in debug builds, `OBJ_DEBUG_LOCK` still needs to access it.

AThousandShips:
The issue is that we're calling on a `Variant` that was reassigned, and the underlying `RefCounted` was freed, the lifetime is that of the `Variant`, namely the member variable on the stack

The lock isn't the problem here, that just works as a check to ensure things work right, it doesn't prevent anything

RandomShaper:
Oh, well, I called the debug lock a "problem," but I didn't mean to imply it is bad or wrong. It was not the most accurate word for what I was trying to say. My point was about what the original intent could have been. I just wanted to share an interesting possibility with you.

Commenting out `OBJ_DEBUG_LOCK` in `Object::callp()` there are no longer crashes nor _asan_ complains, which is the same I was experiencing with the MRP in the original issue.

Back to the PR itself, regardless other potential approaches that could be explored, I believe it's good.

AThousandShips:
> Commenting out OBJ_DEBUG_LOCK in Object::callp() there are no longer crashes nor asan complains, which is the same I was experiencing with the MRP in the original issue.

Are you on Mac? The crash seems to happen there specifically because of how Mac handles memory

RandomShaper:
> Are you on Mac? The crash seems to happen there specifically because of how Mac handles memory

Nope, I'm on Windows. To be more precise, I'm generally getting no crashes so I have to use asan to detect the issue.

AThousandShips:
Likely superseded by:
* https://github.com/godotengine/godot/pull/96856

But awaiting confirmation of the original issue

AThousandShips:
Superseded by:
* https://github.com/godotengine/godot/pull/96856

Issue #72328, opened 2023-01-29
No autocompletion for enums in match


Pan-Horacy:
### Godot version

v4.0.beta16.official [518b9e580]

### System information

Windows 10, Forward+

### Issue description

Missing autocompletion when using enums inside match statements.
![image](https://user-images.githubusercontent.com/86495127/215349313-2d98183f-2513-4b35-8bd5-e207b6b8440e.png)
The same situation in current 3.5 for comparison:
![image](https://user-images.githubusercontent.com/86495127/215349362-fb2be06f-9e16-446d-8741-f1d94d2485e7.png)

### Steps to reproduce

1. Create Enum
2. Create match statement that uses enum
3. Try to write case for it

### Minimal reproduction project

[minimal_project.zip](https://github.com/godotengine/godot/files/10530584/minimal_project.zip)

hd-genius:
I'd like to be assigned to work on this issue.

Zireael07:
@hd-genius Godot does not practice assigning - just make a PR

hd-genius:
Okay, thanks for letting me know!

Eliwi:
Having the same issue in Godot 4.0.2

cdrch:
Just reporting that the issue is still present in Godot v4.1.stable.official [970459615]. Enum autocompletion also still works outside of `match`, and all other forms of autocompletion that I tested worked within `match`.

DarkBiquette666:
Still not working in 4.3

yadavvishal1:
Same problem:
Godot v4.3.stable - Windows 10.0.22631 - Vulkan (Mobile) - dedicated NVIDIA GeForce RTX 2060 SUPER (NVIDIA; 32.0.15.6109) - AMD Ryzen 7 3700X 8-Core Processor (16 Threads)

![Image](https://github.com/user-attachments/assets/8fdb5dbc-08be-4ea3-ade5-94b5e42618eb)

Lexyth:
Just to mention it, autocomplete does work when the colon is already present, at least in v4.3.stable.official [77dcf97d8].
![Image](https://github.com/user-attachments/assets/84cb3594-638e-4f6f-a6f5-b8e307d7cf9e)

notroid5:
Can confirm this works in 4.5 beta 3

PR #86973, opened 2024-01-08
Add tests for `get_node` autocompletion

HolonProduction:
Requires #86961

### DIFF
--- a/modules/gdscript/tests/test_completion.h
+++ b/modules/gdscript/tests/test_completion.h
@@ -113,6 +113,8 @@ static void test_directory(const String &p_dir) {
 			// Require pointer sentinel char in scripts.
 			CHECK(code.find_char(0xFFFF) != -1);
 
+			print_line("Testing completion for: ", next);
+
 			ConfigFile conf;
 			if (conf.load(path.path_join(next.get_basename() + ".cfg")) != OK) {
 				FAIL("No config file found.");
@@ -153,6 +155,7 @@ static void test_directory(const String &p_dir) {
 			GDScriptLanguage::get_singleton()->complete_code(code, path.path_join(next), owner, &options, forced, call_hint);
 			String contains_excluded;
 			for (ScriptLanguage::CodeCompletionOption &option : options) {
+				print_line(option.display);
 				for (const Dictionary &E : exclude) {
 					if (match_option(E, option)) {
 						contains_excluded = option.display;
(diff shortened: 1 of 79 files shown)


Mickeon:
A *whooping* 100 files added. Are they all truly necessary for the tests?

HolonProduction:
> A _whooping_ 100 files added. Are they all truly necessary for the tests?

Yeah the file count kinda exploded :sweat_smile:. But every idea I have to reduce it seems worse.

- Each test needs at least 2 Files. That makes 50 test, but some of them are duplicated in different call situations (yes this is needed, see for example [#72328](https://github.com/godotengine/godot/issues/72328) and there also was another one that i cant find rn :disappointed:). So we are at ~30 test, and I plan to do more, not reduce them. For something as complex as this, reducing tests doesn't seem like a good idea.
- I often have test which require the same output for multiple files. But specifying script paths inside of the cfg files, would make it harder to find the right cfg. And I don't wan't to have multiple scripts inside of the same file, since it would introduce yet another "magic" character to separate them.
- We could reduce each test into one file, but this would either require a sort of preface for the GDScript and parsing that would be annoying. Or we could embed the GDScript inside the cfg, but that would make it impossible to test something manually by opening the script in the editor.

If you have any other ideas to reduce it, I'm all in for it, that being said many test files aren't that big of a problem to justify any of the tradeoffs that I listed IMO.

modules/gdscript/tests/test_completion.h
```diff
@@ -31,6 +31,8 @@
 #ifndef TEST_COMPLETION_H
 #define TEST_COMPLETION_H
 
+#include "core/io/resource_loader.h"
+#include "core/io/resource_uid.h"
```
AThousandShips, line 35: I'd put these among the rest, none of the tests are run without tools anyway so no need for includes

HolonProduction:
Waiting for #87759 in order to fix the CI.

HolonProduction:
Ready for review

vnen (approved):
On a cursory look it is good to me. Thanks for working on this!

akien-mga:
Thanks!

HolonProduction:
Oh shit, I just noticed I forgot to squash the commits and now you merged it :sweat_smile:. I guess it can't be changed now?

akien-mga:
Yeah I missed that. It can't be changed once merged.

PR #86974, opened 2024-01-08
C#: Integrate script/assembly generation into SCons

Repiteo:
QOL for C# engine development workflow. Now SCons is capable of generating C# glue & building .NET assemblies via dedicated SCons builders. This behavior is enabled by default if the mono module is enabled. If the old generation method is preferred or you simply don't want to generate these files, this can be disabled with `dotnet_build=no`

### DIFF
--- a/.github/workflows/linux_builds.yml
+++ b/.github/workflows/linux_builds.yml
@@ -128,16 +128,6 @@ jobs:
           target: ${{ matrix.target }}
           tests: ${{ matrix.tests }}
 
-      - name: Generate C# glue
-        if: ${{ matrix.build-mono }}
-        run: |
-          ${{ matrix.bin }} --headless --generate-mono-glue ./modules/mono/glue
-
-      - name: Build .NET solutions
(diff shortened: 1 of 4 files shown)


raulsntos:
Seems like this has the same goal as https://github.com/godotengine/godot/pull/76542? If so I think the approach in https://github.com/godotengine/godot/pull/76542 is better.

Repiteo:
That does seem quite similar! It seems to take a much more involved approach, whereas this is more about appending the commands to an existing system. I'm not opposed to a more involved approach, but if that were to happen I'd rather have some sort of overhaul to the buildsystem itself to make msbuild do a lot more heavy lifting. I'll try looking into it while taking inspiration from that PR, particularly with seeing how they handled postbuild timing

Repiteo:
Revamped the implementation to use a similar format to the above pr. Specifically: calls are now handled via build systems instead of calling `AddPostAction`. My version remains comparatively simple, leaving the existing scripts almost entirely unchanged

Issue #75245, opened 2023-03-23
Custom C# Resource types made through `add_custom_type` are missing from inspector create resource menu


Atlinx:
### Godot version

4.0.1.stable.mono.official

### System information

Windows 11

### Issue description

I have a plugin that registers C# resources with `EditorPlugin.AddCustomType` as a workaround to https://github.com/godotengine/godot/issues/27470. However the registered resources now don't show up in the create resource menu within the inspector

![broken](https://user-images.githubusercontent.com/25368491/227078893-21c07025-e01d-46f8-97d3-0068b160c46d.gif)

I suspect it might have something to do with how GDScript resources are getting first class support while C# resources still have yet to receive it.

### Steps to reproduce

1) Create C# plugin and a custom C# resource
2) Add the custom C# resource type using `EditorPlugin.AddCustomType`
3) Create a C# node script and export a "Resource".
4) Rebuild the solution.
5) Add the node script to a node
6) Try to create a resource within the exported field by clicking on the empty slot.

### Minimal reproduction project

https://github.com/Atlinx/Godot-Mono-CustomResourceRegistry

RedworkDE:
This issue basically gets obsoleted by https://github.com/godotengine/godot/pull/72619.

raulsntos:
The `add_custom_type` API and global classes are 2 different APIs.

The `add_custom_type` API is kind of _deprecated_ and could be replaced with global classes, it was kept as a fallback for languages that don't implement global class support yet (such as C#). I don't think GDScript types registered with this API work either, so it's not an issue exclusive to C#.

Zylann:
This still happens in 4.1.1, and in GDScript too.

`add_custom_type` alone makes the resource appear in the New Resource dialog, but doesn't appear in dropdowns.
`class_name` alone makes the resource appear in dropdown, but not in the New Resource dialog.

So I end up using both at once...

raulsntos:
@Zylann `class_name` resources should be appearing in both:

MyResource dropdown | Create New Resource
-|-
![MyResource dropdown](https://github.com/godotengine/godot/assets/3903059/72c64420-06d4-48ea-80e1-1c4728ae9842) | ![MyResource Create New Resource](https://github.com/godotengine/godot/assets/3903059/091043f7-03b4-4167-887c-7a585f24cb05)

[...]

PR #86979, opened 2024-01-08
Fix missing icons for custom resources in EditorResourcePicker

### LINKED ISSUE #86072: Custom resource (ArrayMesh) icons are not displayed in resource selector menu and breaks the mesh preview.
### Tested versions

Reproducible in v4.2.stable.arch_linux

### System information

Arch Linux - Godot v4.2.stable.arch_linux

### Issue description

I am working on my custom ArrayMesh class (IcosphereMesh).
After adding `@icon("res://icon.svg")` and restarting the project I figured out that icon is displayed only when resource already set:
image

But in selector menu it has standart resource icon:
image

Also as you can see on the first picture - custom ArrayMesh icon BLOCKS small preview of the mesh 
(preview window is empty) The same picture but with no icon `# @icon("res://icon.svg")` :
image

It is not very important bug but it make me feel quite uncomfortable...

Problem related to Resource icons also was mentioned in this issue: https://github.com/godotengine/godot/issues/82624

### Steps to reproduce

1. Create new project
2. Create 3d scene
3. Add MeshInstance3d to main scene node
4. Create custom class that `extends ArrayMesh`
6. Create a simple surface to be shown in mesh preview display windows
7. Set `MeshInstance3d.mesh` parameter to your custom mesh class
8. Put `@icon("res://icon.svg")`  at the start of class script
9. Reload project
10. Open main scene
11. Open Inspector of MeshInstance3d
12. Unwrap `MeshInstance3d.mesh` property in MeshInstance3d inspector
13. Right click `MeshInstance3d.mesh` to open mesh selector menu.

Results: 
1. Small preview window will be empty,it will show only setted icon and no more. (Bug not affects big preview window in `MeshInstance3d.mesh` inspector)
2. In menu selector your custom array mesh class will have default resource icon that is not equal to "res://icon.svg"

[project.zip](https://github.com/godotengine/godot/files/13648057/project.zip)

Thank you for your attention and sorry for my bad English. Good luck in engine development.

alessandrofama:
The icons of custom resources created with `class_name` and annotated with `@icon` or GDExtension classes that have an icon specified in the .gdextension file are not appearing in the `EditorResourcePicker`.  The problem is that the `EditorResourcePicker` retrieves the editor theme icon for the resource type and defaults to the `Object` icon if the type wasn't found. This will apply both to `class_name` and GDExtension resources.

This solution addresses the issue by replacing the usage of `Control::get_editor_theme_icon` with `EditorNode::get_class_icon` to ensure the correct icon is retrieved for the resource. Additionally, this fix removes the `custom_resources` lookup above that call, as these resources, added through `EditorPlugin::add_custom_type`, were not being included in the allowed types within `_add_allowed_type` in the `EditorResoucePicker`. Currently, these particular custom resources are never displayed in the picker. The related issue is logged here: #75245.

Fixes #86072.

Example with class_name & GDExtension resources extending ArrayMesh, both with default Godot icon.svg icon, exported ArrayMesh variable:

Before:
Screenshot 2024-01-08 234150

After:
Screenshot 2024-01-08 234332

### DIFF
--- a/editor/editor_resource_picker.cpp
+++ b/editor/editor_resource_picker.cpp
@@ -487,38 +487,16 @@ void EditorResourcePicker::set_create_options(Object *p_menu_node) {
 		HashSet<StringName> allowed_types;
 		_get_allowed_types(false, &allowed_types);
 
-		Vector<EditorData::CustomType> custom_resources;
-		if (EditorNode::get_editor_data().get_custom_types().has("Resource")) {
-			custom_resources = EditorNode::get_editor_data().get_custom_types()["Resource"];
-		}
-
 		for (const StringName &E : allowed_types) {
 			const String &t = E;
 
-			bool is_custom_resource = false;
-			Ref<Texture2D> icon;
-			if (!custom_resources.is_empty()) {
-				for (int j = 0; j < custom_resources.size(); j++) {
-					if (custom_resources[j].name == t) {
-						is_custom_resource = true;
-						if (custom_resources[j].icon.is_valid()) {
-							icon = custom_resources[j].icon;
-						}
-						break;
-					}
-				}
-			}
-
-			if (!is_custom_resource && !ClassDB::can_instantiate(t)) {
+			if (!ClassDB::can_instantiate(t)) {
 				continue;
 			}
 
 			inheritors_array.push_back(t);
 
-			if (!icon.is_valid()) {
-				icon = get_editor_theme_icon(has_theme_icon(t, EditorStringName(EditorIcons)) ? t : String("Object"));
-			}
-
+			Ref<Texture2D> icon = EditorNode::get_singleton()->get_class_icon(t, "Object");
 			int id = TYPE_BASE_ID + idx;
 			edit_menu->add_icon_item(icon, vformat(TTR("New %s"), t), id);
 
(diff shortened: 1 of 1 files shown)


YuriSizov (approved):
Looks good and works as expected, thanks!

There is a slight concern that the new lookup path can be slower when you have many custom resources. But that's already a problem across the editor and the solution is to improve caching.

I can also confirm that old style custom types are not working in the resource picker at the moment. The removal of related code is justified IMO, and if we ever fix those custom types, icon lookup should be re-implemented as a part of the `get_class_icon` method.

editor/editor_resource_picker.cpp
```diff
@@ -943,6 +926,7 @@ void EditorResourcePicker::_ensure_resource_menu() {
 	add_child(edit_menu);
 	edit_menu->connect("id_pressed", callable_mp(this, &EditorResourcePicker::_edit_menu_cbk));
 	edit_menu->connect("popup_hide", callable_mp((BaseButton *)edit_button, &BaseButton::set_pressed).bind(false));
+	edit_menu->add_theme_constant_override("icon_max_width", get_theme_constant(SNAME("class_icon_size"), EditorStringName(Editor)));
```
YuriSizov, line 929: We would generally put all configuration lines before `add_child` (except signal connects, which we usually put after `add_child`, like it's done here). But this isn't critical.

alessandrofama: Ah that is good to know! Fixed & I will keep this in mind for next time. Thank you!

PR #86980, opened 2024-01-09
OpenXR: Update to version 1.0.33

BastiaanOlij:
Update the OpenXR library to 1.0.33 so we can use some of the new features.


BastiaanOlij:
@kisg @m4gr3d as far as I'm aware this shouldn't directly influence us but OpenXR has seen a number of enhancements around the Khronos loader on Android. Most focus I believe on the layer system. It is likely more vendors will start adopting it soon.
We might want to look into this at some point.

akien-mga (approved):

akien-mga:
Thanks!

PR #86982, opened 2024-01-09
OpenXR: Add Meta touch plus interaction profile

BastiaanOlij:
This PR adds support for Metas touch plus interaction profile.

It only adds the needed meta data in order to use this profile, it does **not** add this to the default action map. Any touch controller is already supported by the base touch interaction profile.

This new profile adds Quest Pro and Quest 3 specific controller inputs that can be used if developers choose to add support. For this the developer should create their own action map, as would be best practice in any case.

Depends on #86980

### DIFF
--- a/modules/openxr/extensions/openxr_meta_controller_extension.cpp
+++ b/modules/openxr/extensions/openxr_meta_controller_extension.cpp
@@ -1,0 +1,184 @@
+/**************************************************************************/
+/*  openxr_meta_controller_extension.cpp                                  */
+/**************************************************************************/
+/*                         This file is part of:                          */
+/*                             GODOT ENGINE                               */
+/*                        https://godotengine.org                         */
+/**************************************************************************/
+/* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+/* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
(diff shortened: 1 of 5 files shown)


akien-mga:
Just for my information, as I maintain the Godot package on Fedora and Mageia where it builds against system-provided OpenXR: What's the minimal OpenXR version to build Godot after this PR, 1.0.32 or 1.0.33?

modules/openxr/extensions/openxr_meta_controller_extension.cpp
```diff
@@ -0,0 +1,106 @@
...
+	return available;
+}
+
+void OpenXRMetaControllerExtension::on_register_metadata() {
+	OpenXRInteractionProfileMetadata *metadata = OpenXRInteractionProfileMetadata::get_singleton();
+	ERR_FAIL_NULL(metadata);
+
+	// Normal touch controller is part of the core spec
+
+	// Touch controller plus (Quest Pro/3)
```
dsnopek, line 53: This is the comment. This is only for the Quest 3, not Quest Pro.

dsnopek (commented):
Thanks!

This needs to be rebased on `master`, now that both PR https://github.com/godotengine/godot/pull/86980 and https://github.com/godotengine/godot/pull/86792 are merged.

I tested this on both the Quest 3 and Quest Pro, by creating some new actions and assigning them on a new "Touch controller plus" profile:

Selection_100

And then printing out some messages when signals on `XRController3D` were emitted.

This worked great on the Quest 3! However, on the Quest Pro, it picked up the "Touch controller" profile instead. I think this actually only applies to the Quest 3 controllers, not Quest Pro (which are technically called "Touch Pro" not "Touch Plus"). This is only represented in a comment though.

Otherwise, the code looks great!

BastiaanOlij:
@dsnopek good catch! For this we actually also need to add [XR_FB_touch_controller_pro](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_touch_controller_pro)

So I guess we should add that here as well. Can't say I'm too happy about Metas approach here but alas...

@akien-mga honestly, I don't know. I'll ask Rylie if we keep track of that somewhere. My gut feeling says the pro controller was available for awhile, but the plus controller probably was added to 33

BastiaanOlij:
Rebased and now has support for both Pro and Plus controllers. @dsnopek if you don't mind giving it another try?

dsnopek (approved):
With the latest updates, after creating a new profile for "Touch controller pro", it's now working for me on both Quest 3 and Quest Pro :-)

dsnopek:
Or, should we try to sneak in support for [this one](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_touch_controller_proximity) too? It adds some new paths for the "Touch controller" profile.

We could do that in a follow-up too, though.

BastiaanOlij:
> Or, should we try to sneak in support for [this one](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_touch_controller_proximity) too? It adds some new paths for the "Touch controller" profile.

Why not, might as well do the whole lot. It's in there but the DDOS seems to be causing some CI issues :)

akien-mga:
Thanks!

PR #86983, opened 2024-01-09
Make `get_id_path` return empty when first point is disabled

### LINKED ISSUE #86866: AStar2D get_id_path() returns a path when the first point is disabled
### Tested versions

### System information

Godot v4.2.1.stable - Windows 10.0.22621 - Vulkan (Forward+) - integrated Intel(R) UHD Graphics (Intel Corporation; 27.20.100.9415) - Intel(R) Core(TM) i5-1035G1 CPU @ 1.00GHz (8 Threads)

### Issue description

AStar2D's get_id_path() function returns a path when the first point is disabled using set_point_disabled(). It shouldn't.

get_path_id() does not return a path when any other point is disabled using set_point_disabled().

### Steps to reproduce

See code snippet:

```
var astar_test = AStar2D.new()
astar_test.add_point(0,Vector2(0,0))
astar_test.add_point(1,Vector2(0,1))
astar_test.add_point(2,Vector2(0,2))
astar_test.add_point(3,Vector2(0,3))

astar_test.connect_points(0,1)
astar_test.connect_points(1,2)
astar_test.connect_points(2,3)

print(astar_test.get_id_path(0,3)) #Finds a path. Expected behaviour.

astar_test.set_point_disabled(0,true) #Disable the first point.

print(astar_test.get_id_path(0,3)) #Finds a path, despite the first point being disabled. Unexpected behaviour.

astar_test.set_point_disabled(0,false) #Re-enable the first point

astar_test.set_point_disabled(3,true) #Disable the last point

print(astar_test.get_id_path(0,3)) #Does not find a path. Expected behaviour.

```

[astar2d_bug.zip](https://github.com/godotengine/godot/files/13848493/astar2d_bug.zip)

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/86866

This should be discussed more, at least this pr breaks compatibility and might not suit for other's understanding of how A* should work. If this change is not good, I will drop this commit and update the doc instead.

Also I find that AStarGrid2D doesn't have similiar methods for enable and disable single point, I guess it is intended because it deals with region rather than point?

### DIFF
--- a/core/math/a_star.cpp
+++ b/core/math/a_star.cpp
@@ -481,6 +481,10 @@ Vector<int64_t> AStar3D::get_id_path(int64_t p_from_id, int64_t p_to_id, bool p_
 		return ret;
 	}
 
+	if (!a->enabled) {
+		return Vector<int64_t>();
+	}
+
 	Point *begin_point = a;
 	Point *end_point = b;
 
@@ -762,6 +766,10 @@ Vector<int64_t> AStar2D::get_id_path(int64_t p_from_id, int64_t p_to_id, bool p_
 		Vector<int64_t> ret;
 		ret.push_back(a->id);
 		return ret;
+	}
+
+	if (!a->enabled) {
+		return Vector<int64_t>();
 	}
 
 	AStar3D::Point *begin_point = a;


AThousandShips:
Will leave it without the label of compatibility breakage for now, the documentation makes it clear the disabled points are not used for pathfinding, so this is an unintended oversight in my opinion, and not something people should have relied on

marcosdon:
I'm glad other people had the same problem... I thought I was going crazy :) In my project, I stumbled on a bug because I assumed that the first position was checked. At least the documentation should be clear about this behavior, so other people don't fall in the false assumption.

smix8 (approved):

smix8:
If the point is disabled it can't be part of a path, everything else is a bug.

akien-mga:
Thanks!

PR #86991, opened 2024-01-09
Shaders: Allow trailing commas in function calls

### LINKED ISSUE #8840: Allow trailing commas in shader language

A shader-heavy video sequence generator

Long shading expressions like
```glsl
vec3 fadeInTex = clamp(mix(-1.0, 2.0, fadeIn) + texture(fade_in_bias, UV).rgb, 0.0, 1.0);
```
would be nicer formatted as
```glsl
vec3 fadeInTex = clamp(
  mix(-1.0, 2.0, fadeIn) + texture(fade_in_bias, UV).rgb, 
  0.0, 
  1.0,
);
```
but the presence of a trailing comma in the argument list (as allowed in GDScript in 4.0+, https://github.com/godotengine/godot/issues/41066) yields

> error(69): Expected expression, found: 'PARENTHESIS_CLOSE'

For context, function definition argument lists already accept a single trailing comma; both

```glsl
vec3 neonChevrons(vec2 uv,) {
```
and
```glsl
vec3 neonChevrons(
	vec2 uv,
) {
```
parse fine.

This is somewhat tautological, but:

Fixing the shading language parser to allow trailing commas in function call argument lists in the shader language parser would allow one to use trailing commas in function call argument lists.

When the user reformats their shader code to a form where a function call has a trailing comma in its argument list, the comma token would be swallowed, and no error for a missing expression would be emitted.

This can not be worked around with script, because it's part of the shading language parser: https://github.com/godotengine/godot/blob/84e205b5a17bfe7ace96b78c410ec10aa52241d2/servers/rendering/shader_language.cpp#L5739-L5743

The shading language parser is part of Godot core.

akx:
This PR implements https://github.com/godotengine/godot-proposals/issues/8840: allowing trailing commas in shader call function call argument lists like
```glsl
vec3 fadeInTex = clamp(
  mix(-1.0, 2.0, fadeIn) + texture(fade_in_bias, UV).rgb, 
  0.0, 
  1.0,
);
```

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/8840*

### DIFF
--- a/servers/rendering/shader_language.cpp
+++ b/servers/rendering/shader_language.cpp
@@ -3176,7 +3176,7 @@ bool ShaderLanguage::_validate_function_call(BlockNode *p_block, const FunctionI
 				}
 
 				if (!fail) {
-					// When running tests, we might not have a proper rendering server to check low-endness with
+					// When running tests, we might not have a proper rendering server to check low-endness with.
 					RenderingServer *rendering_server = RenderingServer::get_singleton();
 					if (rendering_server != nullptr && rendering_server->is_low_end()) {
 						if (builtin_func_defs[idx].high_end) {
@@ -3639,7 +3639,7 @@ bool ShaderLanguage::_parse_function_arguments(BlockNode *p_block, const Functio
 		// if so, it must be a argument-list-trailing comma.
 		if (_peek_token().type == TK_PARENTHESIS_CLOSE) {
 			// Swallow the closing parenthesis and return.
-			ERR_FAIL_COND_V(_get_token().type != TK_PARENTHESIS_CLOSE, false);
+			_get_token();
 			break;
 		}
 	}
--- a/servers/rendering/shader_language.cpp
+++ b/servers/rendering/shader_language.cpp
@@ -883,7 +883,7 @@ ShaderLanguage::Token ShaderLanguage::_get_token() {
 #undef GETCHAR
 }
 
-bool ShaderLanguage::_lookup_next(Token &r_tk) {
+bool ShaderLanguage::_lookup_next_for_completion(Token &r_tk) {
 	TkPos pre_pos = _get_tkpos();
 	int line = pre_pos.tk_line;
 	_get_token();
@@ -6939,7 +6939,7 @@ Error ShaderLanguage::_parse_block(BlockNode *p_block, const FunctionInfo &p_fun
 			}
 
 #ifdef DEBUG_ENABLED
-			if (is_const && _lookup_next(next)) {
+			if (is_const && _lookup_next_for_completion(next)) {
 				if (is_token_precision(next.type)) {
 					keyword_completion_context = CF_UNSPECIFIED;
 				}
@@ -8337,7 +8337,7 @@ Error ShaderLanguage::_parse_shader(const HashMap<StringName, FunctionInfo> &p_f
 			case TK_GLOBAL: {
 #ifdef DEBUG_ENABLED
 				keyword_completion_context = CF_UNIFORM_KEYWORD;
-				if (_lookup_next(next)) {
+				if (_lookup_next_for_completion(next)) {
 					if (next.type == TK_UNIFORM) {
 						keyword_completion_context ^= CF_UNIFORM_KEYWORD;
 					}
@@ -8355,7 +8355,7 @@ Error ShaderLanguage::_parse_shader(const HashMap<StringName, FunctionInfo> &p_f
 				if (tk.type == TK_INSTANCE) {
 #ifdef DEBUG_ENABLED
 					keyword_completion_context = CF_UNIFORM_KEYWORD;
-					if (_lookup_next(next)) {
+					if (_lookup_next_for_completion(next)) {
 						if (next.type == TK_UNIFORM) {
 							keyword_completion_context ^= CF_UNIFORM_KEYWORD;
 						}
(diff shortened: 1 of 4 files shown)


servers/rendering/shader_language.cpp
```diff
@@ -394,6 +394,9 @@ const ShaderLanguage::KeyWord ShaderLanguage::keyword_list[] = {
 	{ TK_ERROR, nullptr, CF_UNSPECIFIED, {}, {} }
 };
 
+/**
+ * @brief Parse the next token from the code.
```
AThousandShips, line 398: We don't use these forms of documentation comments, should just be a normal comment or removed 🙂

tests/servers/rendering/test_shader_language.h
```diff
@@ -0,0 +1,66 @@
+/**************************************************************************/
+/*  test_shader_language.h                                            */
```
AThousandShips, line 2:
```suggestion
/*  test_shader_language.h                                                */
```

servers/rendering/shader_language.h
```diff
@@ -1011,7 +1011,8 @@ class ShaderLanguage {
 
 	Token _make_token(TokenType p_type, const StringName &p_text = StringName());
 	Token _get_token();
-	bool _lookup_next(Token &r_tk);
+	Token _peek_token();
+	bool _lookup_next_for_completion(Token &r_tk);
```
AThousandShips, line 1015: Why this rename?

akx: Because it's not clear from the original name, or the (lack of 😁) documentation comment for that function that it peeks two tokens and only returns `true` (and the second token) if it's on the same physical line where the parser had been before calling it.

Naively, I tried `_lookup_next()` first before I realized it's not semantically `_peek_token()` at all.

AThousandShips:
See also:
* https://github.com/godotengine/godot/pull/87075

servers/rendering/shader_language.cpp
```diff
@@ -394,6 +394,9 @@ const ShaderLanguage::KeyWord ShaderLanguage::keyword_list[] = {
 	{ TK_ERROR, nullptr, CF_UNSPECIFIED, {}, {} }
 };
 
+/**
+ * Parse the next token from the code and advance the cursor.
+ */
```
AThousandShips, line 399:
```suggestion
// Parse the next token from the code and advance the cursor.
```

servers/rendering/shader_language.cpp
```diff
@@ -896,6 +899,17 @@ bool ShaderLanguage::_lookup_next(Token &r_tk) {
 	return false;
 }
 
+/**
+ * Parse the next token, then rewind the tokenizer state
+ * so it looks as if the token was never parsed.
+ */
```
AThousandShips, line 905:
```suggestion
// Parse the next token, then rewind the tokenizer state so it looks as if the token was never parsed.
```

tests/servers/rendering/test_shader_language.h
```diff
@@ -0,0 +1,66 @@
...
+#ifndef TEST_SHADER_LANGUAGE_H
+#define TEST_SHADER_LANGUAGE_H
+
+#include "servers/rendering/shader_language.h"
+
+#include "tests/test_macros.h"
+
+#include <servers/rendering/shader_types.h>
+
+#include <cctype>
```
AThousandShips, line 40:
```suggestion
```
Unnecessary include

servers/rendering/shader_language.cpp
```diff
@@ -3617,6 +3635,13 @@ bool ShaderLanguage::_parse_function_arguments(BlockNode *p_block, const Functio
 			_set_error(RTR("Expected ',' or ')' after argument."));
 			return false;
 		}
+		// Peek at whether there's a closing parenthesis after the last comma;
+		// if so, it must be a argument-list-trailing comma.
+		if (_peek_token().type == TK_PARENTHESIS_CLOSE) {
+			// Swallow the closing parenthesis and return.
+			ERR_FAIL_COND_V(_get_token().type != TK_PARENTHESIS_CLOSE, false);
+			break;
+		}
```
AThousandShips, line 3644: I'm not sure this sequence makes sense, is it actually possible for the peeked token to not end up being the correct token? If `_peek_token` can result in a different result than `_get_token` I'd say something is wrong

AThousandShips:
Haven't tested but doesn't look like this fixes:
```
MyStruct my_struct = MyStruct(foo, bar, );
```
Or
```
uniform float my_uniform : hint_range(0, 1, );
```

If we allow it explicitly in function calls these should also be allowed IMO, or it makes things more inconsistent

akx:
> Haven't tested but doesn't look like this fixes [struct initializers or uniform hint syntax]

No, it doesn't. Those are parsed by an entirely different part of code.

> If we allow it explicitly in function calls these should also be allowed IMO, or it makes things more inconsistent

Sure, I can do that. I'm just waiting for an actual decision on https://github.com/godotengine/godot-proposals/issues/8840 before doing more work that might just end up being thrown away if someone decides trailing commas should not be allowed.

servers/rendering/shader_language.cpp
akx: There is precedent around the codebase to do e.g.

```
/** This method uses a signed argument for better error reporting as it's used from the scripting API. */
```

so I'm not sure if there is a consistent style guide recommendation to do this instead.

servers/rendering/shader_language.cpp
akx: Ditto here.

servers/rendering/shader_language.cpp
akx: We can do that. Just feels incorrect to get the token and not do anything about it.

servers/rendering/shader_language.cpp
AThousandShips: We essentially never use this style, and it's not really a doc-comment, but it's permitted in the comment style

akx:
Closing since there has been no movement on https://github.com/godotengine/godot-proposals/issues/8840 and to clean up my open PRs list.

PR #86999, opened 2024-01-09
Improve error message when a GDScript instance fails to be constructed

### LINKED ISSUE #5899: Inform the users that non-default-constructible nodes are a bad thing

2D strategy game

When I created a new script that inherited from Node and put in into the scene, which I tried to later instantiate I got the following, cryptic error:

```
E 0:00:00:0302   _create_instance: Error constructing a GDScriptInstance.
  <C++ Error>    Method/function failed. Returning: nullptr
  <C++ Source>   modules/gdscript/gdscript.cpp:183 @ _create_instance()
```

Worst of all, is that you can ignore the error and everything seems to work fine, until you access the object and find out that only the base class was constructed...

Currently, the above error is the only thing that is telling us that we have a type which cannot be default constructed and we try to instantiate the scene or duplicate it.

*NDC* = *N*on *D*efault *C*onstructible

There are several ways we can help user. From the cheapest, but least useful, to more intrusive ones.

1. Be more explicit in the documentation about this issue. The only mention about this issue is present ... in the documentation of `Object` which is a place that users won't go into, I guess. Text there states as follows:
"Note: If [_init](https://docs.godotengine.org/en/latest/classes/class_object.html#class-object-method-init) is defined with required parameters, the Object with script may only be created directly. If any other means (such as [PackedScene.instantiate](https://docs.godotengine.org/en/latest/classes/class_packedscene.html#class-packedscene-method-instantiate) or [Node.duplicate](https://docs.godotengine.org/en/latest/classes/class_node.html#class-node-method-duplicate)) are used, the script's initialization will fail."

We should at least add this information where we introduce users to Nodes or GDScript language.

As the biggest offender of this rule are custom `Node`-derived classes, subsequent proposals are aimed towards them.

2. In text editor when a NDC class is detected, a code warning can be added, so that the user knows that he/she limits the usage of this Node to dynamic instantiation.

3. In the Editor when a node is added which is NDC add a configuration warning (originating from the Node I guess) that this won't go well when trying to do `instantiate`.

4. When doing a `pack()` of the scene (or during construction of `PackedScene`) raise an error specifying that trying to continue will fail in the long run.

I believe 1, 3 & 4 would be perfect. Documentation wouldn't be harmed. You could continue writing your `func _init(arg) -> void` `Node`s without interruptions. When trying to instantiate them you could get some friendly warnings and finally, when you are stubborn, you would get a flat error (and no scene).

Described above.

It cannot be worked around - it's just a gotcha that every new user of Godot engine has to go through. Needlessly, might I add.

No asset library is concerned with creation of your custom types.

### LINKED ISSUE #8488: Improve error message when Godot is unable to construct GDScriptInstance

AThousandShips:
I think this wording works well, but could take some inspiration from https://github.com/godotengine/godot/pull/86635, would be up for either way

* Salvaged from: #76632

* Closes: https://github.com/godotengine/godot-proposals/issues/5899
* Closes: https://github.com/godotengine/godot-proposals/issues/8488

### DIFF
--- a/modules/gdscript/gdscript.cpp
+++ b/modules/gdscript/gdscript.cpp
@@ -162,13 +162,14 @@ GDScriptInstance *GDScript::_create_instance(const Variant **p_args, int p_argco
 
 	_super_implicit_constructor(this, instance, r_error);
 	if (r_error.error != Callable::CallError::CALL_OK) {
+		String error_text = Variant::get_call_error_text(instance->owner, "@implicit_new", nullptr, 0, r_error);
 		instance->script = Ref<GDScript>();
 		instance->owner->set_script_instance(nullptr);
 		{
 			MutexLock lock(GDScriptLanguage::singleton->mutex);
 			instances.erase(p_owner);
 		}
-		ERR_FAIL_V_MSG(nullptr, "Error constructing a GDScriptInstance.");
+		ERR_FAIL_V_MSG(nullptr, "Error constructing a GDScriptInstance: " + error_text);
 	}
 
 	if (p_argcount < 0) {
@@ -179,13 +180,14 @@ GDScriptInstance *GDScript::_create_instance(const Variant **p_args, int p_argco
 	if (initializer != nullptr) {
 		initializer->call(instance, p_args, p_argcount, r_error);
 		if (r_error.error != Callable::CallError::CALL_OK) {
+			String error_text = Variant::get_call_error_text(instance->owner, "_init", p_args, p_argcount, r_error);
 			instance->script = Ref<GDScript>();
 			instance->owner->set_script_instance(nullptr);
 			{
 				MutexLock lock(GDScriptLanguage::singleton->mutex);
 				instances.erase(p_owner);
 			}
-			ERR_FAIL_V_MSG(nullptr, "Error constructing a GDScriptInstance.");
+			ERR_FAIL_V_MSG(nullptr, "Error constructing a GDScriptInstance: " + error_text);
 		}
 	}
 	//@TODO make thread safe


modules/gdscript/gdscript.cpp
```diff
@@ -179,13 +180,14 @@ GDScriptInstance *GDScript::_create_instance(const Variant **p_args, int p_argco
...
 		if (r_error.error != Callable::CallError::CALL_OK) {
+			String error_text = Variant::get_call_error_text(instance->owner, "_init", p_args, p_argcount, r_error);
 			instance->script = Ref<GDScript>();
 			instance->owner->set_script_instance(nullptr);
 			{
 				MutexLock lock(GDScriptLanguage::singleton->mutex);
 				instances.erase(p_owner);
 			}
-			ERR_FAIL_V_MSG(nullptr, "Error constructing a GDScriptInstance.");
+			ERR_FAIL_V_MSG(nullptr, "Error constructing a GDScriptInstance: " + error_text);
```
AThousandShips, line 190: Could do something like:
```suggestion
			if (r_error.error == Callable::CallError::CALL_ERROR_TOO_FEW_ARGUMENTS) {
				ERR_FAIL_V_MSG(nullptr, "Error constructing a GDScriptInstance: _init(...) requires parameters. Consider adding default values.");
			} else {
			ERR_FAIL_V_MSG(nullptr, "Error constructing a GDScriptInstance: " + error_text);
```
Or similar, unsure if it should be added to the above case as well, or how to word it

vnen: I'm not sure if this is the best place to do it, as there are many ways a constructor can be called. This message would also trigger if the constructor accepts 2 arguments and was called with 1, which could be an actual user mistake when calling it.

AThousandShips: Agreed, keeping it simple then :)

vnen (approved):

AThousandShips:
Thank *you*!

PR #87006, opened 2024-01-09
Fix member names of `AudioFrame` to match extension

### LINKED ISSUE #86986: AudioFrame parameter name mismatch
### Tested versions

last code ： [84e205b](https://github.com/godotengine/godot/commit/84e205b5a17bfe7ace96b78c410ec10aa52241d2)

### System information

windows 11

### Issue description

I think AudioFrame parameter name mismatch,

in this code: https://github.com/godotengine/godot/blob/84e205b5a17bfe7ace96b78c410ec10aa52241d2/core/math/audio_frame.h#L55C1-L55C1
```
struct AudioFrame {
	//left and right samples
	float l = 0.f, r = 0.f;
}
```

the param name is l and r, 

but in this code:
https://github.com/godotengine/godot/blob/84e205b5a17bfe7ace96b78c410ec10aa52241d2/core/register_core_types.cpp#L295
```
GDREGISTER_NATIVE_STRUCT(AudioFrame, "float left;float right");
```
the param name is left and right, 

I think you need to modify one of them? If necessary, I can provide PR

### Steps to reproduce

null

null

AThousandShips:
Ran without the compatibility cases and it didn't complain about any unused cases so seems that all the old uses are fixed, would need someone to confirm though

Ran CI without the old names and didn't have any problems on those platforms so should be covered

* Closes: https://github.com/godotengine/godot/issues/86986

### DIFF
--- a/core/math/audio_frame.h
+++ b/core/math/audio_frame.h
@@ -51,105 +51,123 @@ static const float AUDIO_PEAK_OFFSET = 0.0000000001f;
 static const float AUDIO_MIN_PEAK_DB = -200.0f; // linear_to_db(AUDIO_PEAK_OFFSET)
 
 struct AudioFrame {
-	//left and right samples
-	float l = 0.f, r = 0.f;
+	// Left and right samples.
+	union {
+		struct {
+			float left;
+			float right;
+		};
+#ifndef DISABLE_DEPRECATED
+		struct {
+			float l;
+			float r;
+		};
+#endif
+		float levels[2] = { 0.0 };
+	};
 
-	_ALWAYS_INLINE_ const float &operator[](int idx) const { return idx == 0 ? l : r; }
-	_ALWAYS_INLINE_ float &operator[](int idx) { return idx == 0 ? l : r; }
+	_ALWAYS_INLINE_ const float &operator[](int p_idx) const {
+		DEV_ASSERT((unsigned int)p_idx < 2);
+		return levels[p_idx];
+	}
+	_ALWAYS_INLINE_ float &operator[](int p_idx) {
+		DEV_ASSERT((unsigned int)p_idx < 2);
+		return levels[p_idx];
+	}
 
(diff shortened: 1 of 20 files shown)


core/math/audio_frame.h
```diff
@@ -51,105 +51,123 @@ static const float AUDIO_PEAK_OFFSET = 0.0000000001f;
...
+			float l;
+			float r;
+		};
+#endif
+		float levels[2] = { 0.0 };
+	};
+
+	_ALWAYS_INLINE_ const float &operator[](int p_idx) const {
+		DEV_ASSERT((unsigned int)p_idx < 2);
+		return levels[p_idx];
```
AThousandShips, line 71: Could keep the old syntax here but nicer on release IMO and matches how we do in `Vector2/3/4`

core/math/audio_frame.h
```diff
@@ -51,105 +51,123 @@ static const float AUDIO_PEAK_OFFSET = 0.0000000001f;
...
+		res.right += (p_t * (p_b.right - right));
 
 		return res;
 	}
 
 	_ALWAYS_INLINE_ AudioFrame(float p_l, float p_r) {
-		l = p_l;
-		r = p_r;
+		left = p_l;
+		right = p_r;
```
vnen, line 138: Not entirely needed but the parameters could be renamed here too.

AThousandShips: Avoided it to reduce the impact but can do it there too

AThousandShips:
@YuriSizov only tangentially related to extensions, the alternative would be to rename the extension ones to `l/r` possibly breaking compatibility, and being less readable, this doesn't affect extensions at all otherwise

YuriSizov:
@AThousandShips This is only an issue if we consider extensions to begin with, AFAIU. Which makes the GDExtension team a natural stakeholder in the solution, whether it touches extension code or not 🙂

akien-mga (approved):
Might need a rebase for good measure.

AThousandShips:
Thank *you*!

PR #87010, opened 2024-01-09
Fix Dummy audio driver initialization issue on WASAPI output device initialization failure

### LINKED ISSUE #71640: AudioStreamPlayer "finished" signal is never emitted if there is no audio output device
### Godot version

3.5.1.stable

### System information

Windows 10, GLES3, NVIDIA GTX 970, driver version 31.0.15.1694

### Issue description

The AudioStreamPlayer node's "finished" signal is never emitted if the system has no audio output device. I tested this by unplugging all headphones and line out devices from my machine, and then starting my project. My intuition is that having no output device should functionally equivalent to muting the game, however this is not the case. If the game is muted (but still has an output device) the "finished" signal is still emitted.

If you don't know that the lack of an audio output device causes this signal to never be emitted, and you `yield` on the signal in code, this can easily lead to severe game-breaking bugs.

I guess it makes some kind of sense that no audio can be played without an audio output device, however the `is_playing` method on AudioStreamPlayer _does_ return true even when there is no output device.

I have not tested AudioStreamPlayer3D or AudioStreamPlayer2D, but they are probably affected too.

### Steps to reproduce

Sample code:

```GDScript
$AudioStreamPlayer.stream = preload("some audio file.wav")
$AudioStreamPlayer.play()
yield($AudioStreamPlayer, "finished")
print("This line is never printed")
```

Unplug all your headphones and speakers, and then run the code. The `print` statement is never executed.

N/A

alessandrofama:
`AudioDriverWASAPI::init` consistently returns `Error::OK`, even when encountering a failure during the initialization of the output device. This behaviour blocks the dummy driver from initializing in `AudioDriverManager::initialize`. Fixes #71640.

### DIFF
--- a/drivers/wasapi/audio_driver_wasapi.cpp
+++ b/drivers/wasapi/audio_driver_wasapi.cpp
@@ -557,12 +557,10 @@ Error AudioDriverWASAPI::init() {
 
 	target_latency_ms = Engine::get_singleton()->get_audio_output_latency();
 
-	Error err = init_output_device();
-	if (err != OK) {
-		ERR_PRINT("WASAPI: init_output_device error");
-	}
-
 	exit_thread.clear();
+
+	Error err = init_output_device();
+	ERR_FAIL_COND_V_MSG(err != OK, err, "WASAPI: init_output_device error.");
 
 	thread.start(thread_func, this);
 


drivers/wasapi/audio_driver_wasapi.cpp
```diff
@@ -559,7 +559,7 @@ Error AudioDriverWASAPI::init() {
 
 	Error err = init_output_device();
 	if (err != OK) {
-		ERR_PRINT("WASAPI: init_output_device error");
+		ERR_FAIL_V_MSG(ERR_CANT_OPEN, "WASAPI: init_output_device error");
```
AThousandShips, line 562:
```suggestion
		ERR_FAIL_V_MSG(ERR_CANT_OPEN, "WASAPI: init_output_device error.");
```
While we're at it let's correct the format 🙂

AThousandShips: Or rather replace the whole block with:
```cpp
ERR_FAIL_COND_V_MSG(err != OK, ERR_CANT_OPEN, "WASAPI: init_output_device error.");
```
Or better yet:
```cpp
ERR_FAIL_COND_V_MSG(err != OK, err, "WASAPI: init_output_device error.");
```

alessandrofama: makes sense! Thank you!

drivers/wasapi/audio_driver_wasapi.cpp
```diff
@@ -558,9 +558,7 @@ Error AudioDriverWASAPI::init() {
 	target_latency_ms = Engine::get_singleton()->get_audio_output_latency();
 
 	Error err = init_output_device();
-	if (err != OK) {
-		ERR_PRINT("WASAPI: init_output_device error");
-	}
+	ERR_FAIL_COND_V_MSG(err != OK, err, "WASAPI: init_output_device error.");
 
 	exit_thread.clear();
```
AThousandShips, line 563: Unsure if this should be moved up, it is above in the alsa code and could cause threading problems if not fired with this change, it's put at the top in some cases, I don't know what the exact structure should be for that

AThousandShips (approved):
Makes sense otherwise, and other drivers exit this way as well

drivers/wasapi/audio_driver_wasapi.cpp
alessandrofama: I can see it's above in the output initialisation of the alsa & pulseaudio drivers. The flag is only set in the thread entry point and in `AudioDriverWASAPI::finish`. Not sure about the structure too. Could be moved up for consistency.

alessandrofama: moved it up

akien-mga:
Thanks!

akien-mga:
The original bug report was against 3.5.1, but this PR doesn't seem applicable to the 3.x code, which currently looks like this:
```cpp
Error AudioDriverWASAPI::init() {
        mix_rate = GLOBAL_GET("audio/mix_rate");

        Error err = init_render_device();
        if (err != OK) {
                ERR_PRINT("WASAPI: init_render_device error");
        }

        exit_thread.clear();

        thread.start(thread_func, this);

        return OK;
}
```

@alessandrofama Would you be up for checking whether the issue is still reproducible in `3.x` and thus how to solve it there?

alessandrofama:
> @alessandrofama Would you be up for checking whether the issue is still reproducible in `3.x` and thus how to solve it there?

Yes, it was reproducible. Created a PR :)

YuriSizov:
Cherry-picked for 4.2.2.

PR #87012, opened 2024-01-09
Fix `LineEdit` caret rendering/positioning on Undo.

### LINKED ISSUE #67169: Using redo/undo in LineEdit may break caret rendering
### Godot version

880a0177d

### System information

Linux Mint 21, Vulkan API 1.2.0, NVIDIA GeForce GTX 1050 Ti

### Issue description

This may be a drawing issue, related to #27420, #27421.

### Steps to reproduce

1. Add a long text;
2. Delete from back to front;
3. Undo the previous deletion.

_No response_

TheSofox:
Fixes  #67169.

This fixes the circumstance where undoing deleted text in `LineEdit` could lead to caret appearing outside of the box. (This bug has become even easier to reproduce with recent PR https://github.com/godotengine/godot/pull/86732 , though repo steps are the same )

The reason for this appears to be that in the `undo()` function, the `set_caret_column()` was called before the `_shape()` function. However, it seems the `_shape()` function needs to be called first (as it is in other parts of the `LineEdit.cpp`) as otherwise the `get_caret_pixel_pos()` that `set_caret_column()` uses gives an incorrect value (that the caret is at 0, 0).

I've reordered these functions in both `undo()` and `redo()` and the bug now appears fixed with the caret behaving appropriately when undoing.

### DIFF
--- a/scene/gui/line_edit.cpp
+++ b/scene/gui/line_edit.cpp
@@ -1228,9 +1228,10 @@ void LineEdit::undo() {
 	TextOperation op = undo_stack_pos->get();
 	text = op.text;
 	scroll_offset = op.scroll_offset;
-	set_caret_column(op.caret_column);
 
 	_shape();
+	set_caret_column(op.caret_column);
+
 	_emit_text_change();
 }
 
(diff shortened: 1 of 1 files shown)


bruvzg (approved):

akien-mga:
Thanks!

PR #87018, opened 2024-01-09
Add documentation to PhysicsServer2DExtension

Mickeon:
Oh goodness me.

This PR... This PR fills in the class reference of **PhysicsServer2DExtension**.

A grand majority of these methods link to the **PhysicsServer2D** equivalents. The reasons for this are as follow:
- Curating whole new descriptions with the assumption that that they are overridable now is not something I would wish for to my dearest enemy. It could be my job and I wouldn't do it, at least not all at once.
- Copying the descriptions between the two word-by-word would mean having to maintain **two** copies of the same identical description.
- This documentation is tailored towards advanced users, even more "advanced" than those that would read **PhysicsServer2D**'s class reference. They should roughly know what they're doing already.
     - Most of this documentation basically behaves like a quality of life "click here to read more". 

Not all of them point back to **PhysicsServer2D**. Some have been described from scratch, because there's no equivalent of them exposed to the scripting languages.

Could use the feedback of physics-savvy maintainers, but I somehow have the confidence that the new descriptions are accurate enough.

### DIFF
--- a/doc/classes/PhysicsServer2DExtension.xml
+++ b/doc/classes/PhysicsServer2DExtension.xml
@@ -17,6 +17,7 @@
 			<param index="2" name="transform" type="Transform2D" />
 			<param index="3" name="disabled" type="bool" />
 			<description>
+				Overridable version of [method PhysicsServer2D.area_add_shape].
 			</description>
 		</method>
 		<method name="_area_attach_canvas_instance_id" qualifiers="virtual">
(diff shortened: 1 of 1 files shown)


Mickeon:
This and https://github.com/godotengine/godot/pull/87030 could still use a small look

mhilbrunner (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #87021, opened 2024-01-09
Fix export window size getting epic proportions

### LINKED ISSUE #85551: [4.2] Export Window Sized outside of screen, cannot be resized when export template is selected

hakro:
Fixes https://github.com/godotengine/godot/issues/85551

Not sure why, but the smart autowrap on the seemingly harmless label `server_strip_message` gives the export window a crazy height when`Export Mode` is set to `Export as dedicated server` , making it impossible to access the buttons to export your project when in `Single Window Mode`.
That label is only supposed to hold some information text, listing the resources that will be stripped from the server export.

For some reason changing the autowrap mode to something else fixes the issue. So I switched it to `AUTOWRAP_WORD` here, which should look quite similar to what's intended.

Might be a `TextServer` issue, but I have not looked into that.

### DIFF
--- a/editor/export/project_export.cpp
+++ b/editor/export/project_export.cpp
@@ -1276,7 +1276,7 @@ ProjectExportDialog::ProjectExportDialog() {
 
 	server_strip_message = memnew(Label);
 	server_strip_message->set_visible(false);
-	server_strip_message->set_autowrap_mode(TextServer::AUTOWRAP_WORD_SMART);
+	server_strip_message->set_autowrap_mode(TextServer::AUTOWRAP_WORD);
 	resources_vb->add_child(server_strip_message);
 
 	{


dsnopek (approved):
Thanks so much for digging into this!

I tested this PR and it fixed the issue for me. And, as the original author of the line you changing (sorry!), I can say that we don't need it to be `AUTOWRAP_WORD_SMART` for any particular reason - using `AUTOWRAP_WORD` should be fine.

Calinou:
There is another fix here: https://github.com/godotengine/godot/pull/86145

MewPurPur:
I'm more interested in what's causing WORD_SMART to cause this travesty. I wouldn't like it causing any nasty bugs in my projects.

YuriSizov:
Thanks for opening a PR! The one linked by Calinou, https://github.com/godotengine/godot/pull/86145, is the correct way to address this for now.

Changing to another autowrapping behavior is not going to fix the issue, ultimately. The problem is with the autowrapping itself, not with this specific mode. Problem with autowrapping is that without any sensible default size the control will try to shrink to its minimum allowed size. Autowrapping makes it possible for the control to shrink to the size of its smallest text unit horizontally, because it can wrap around the remainder. This causes it to initially compute itself to have its smallest width, but as a result it has its biggest height.

A different wrapping mode is just changing the size of that smallest width, but not the rest of the behavior. Of course if your smallest unit is a word instead of a character, the vertical expansion won't be as dramatic. But the underlying problem will still be present. So for now we must give labels with wrapping a reasonable width, like https://github.com/godotengine/godot/pull/86145 does.

So I'm going to close this PR, but thanks nevertheless!

PR #87022, opened 2024-01-09
Fix incorrect mapping of initial action as clear region continue to clear.

DarioSamo:
Fixes the regression caused by #84976 on canvas renderer where the 2D lights and shadows demo showed broken behavior.

INITIAL_ACTION_CLEAR_REGION_CONTINUE implied a clear on the region defined by the draw list. This was incorrectly assigned to "LOAD" in the new initial action redesign, leading to the framebuffer not being cleared as intended.

Before (broken):
image

After:
image

No existing code uses INITIAL_ACTION_CLEAR_REGION_CONTINUE anymore, so if there's any other potential places that should be reviewed, it's whatever used the action before the PR was merged, which was limited to a very small amount of effects (notably, omni-lights in 3D used them but they're not broken, so their action mappings seem correct).


clayjohn (approved):
Looks great!

DarioSamo:
Small doc update for CI.

misc/extension_api_validation/4.2-stable.expected
```diff
@@ -35,7 +35,7 @@ GH-84976
 Validate extension JSON: Error: Field 'classes/RenderingDevice/enums/FinalAction/values/FINAL_ACTION_CONTINUE': value changed value in new API, from 2.0 to 0.
 Validate extension JSON: Error: Field 'classes/RenderingDevice/enums/FinalAction/values/FINAL_ACTION_MAX': value changed value in new API, from 3.0 to 2.
 Validate extension JSON: Error: Field 'classes/RenderingDevice/enums/InitialAction/values/INITIAL_ACTION_CLEAR': value changed value in new API, from 0.0 to 1.
-Validate extension JSON: Error: Field 'classes/RenderingDevice/enums/InitialAction/values/INITIAL_ACTION_CLEAR_REGION_CONTINUE': value changed value in new API, from 2.0 to 0.
+Validate extension JSON: Error: Field 'classes/RenderingDevice/enums/InitialAction/values/INITIAL_ACTION_CLEAR_REGION_CONTINUE': value changed value in new API, from 2.0 to 1.
```
AThousandShips, line 38: You need to add a new entry for this not edit an old one 🙂

akien-mga: This is a changelog between 4.2-stable and current `master`, so it shouldn't include iterations on the API within the dev cycle. I.e. changing this in place seems correct.

DarioSamo: Yes, @akien-mga's reasoning is pretty much why I did it in place, as it's just a fix for an incorrect mapping that the PR introduced.

akien-mga:
Thanks!

PR #87025, opened 2024-01-10
[Button] Add autowrap feature.

bruvzg:
Implements https://github.com/godotengine/godot-proposals/issues/8823

### DIFF
--- a/scene/gui/button.cpp
+++ b/scene/gui/button.cpp
@@ -128,6 +128,15 @@ void Button::_notification(int p_what) {
 			queue_redraw();
 		} break;
 
+		case NOTIFICATION_RESIZED: {
+			if (autowrap_mode != TextServer::AUTOWRAP_OFF) {
+				_shape();
+
+				update_minimum_size();
+				queue_redraw();
+			}
+		} break;
+
 		case NOTIFICATION_DRAW: {
 			const RID ci = get_canvas_item();
 			const Size2 size = get_size();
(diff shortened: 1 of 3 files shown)


hemu0069 (approved):
good

hemu0069 (commented):
this one good

scene/gui/button.cpp
```diff
@@ -470,6 +496,19 @@ String Button::get_text() const {
 	return text;
 }
 
+void Button::set_autowrap_mode(TextServer::AutowrapMode p_mode) {
+	if (autowrap_mode != p_mode) {
+		autowrap_mode = p_mode;
+		_shape();
+		queue_redraw();
+		update_minimum_size();
+	}
```
AThousandShips, line 505:
```suggestion
	if (autowrap_mode == p_mode) {
		return;
	}
	autowrap_mode = p_mode;
	_shape();
	queue_redraw();
	update_minimum_size();
```
Use early exit

bruvzg: The rest of the setters in the `LineEdit` use this style, so I have used it as well for consistency.
This probably should be changed everywhere in the separate PR, since different parts of code use both styles.

AThousandShips: I'd say that the early exit style is the preferred even with consistency, but I'll leave it to you 🙂

klootas:
Thanks for doing this PR, @bruvzg - hope it gets accepted soon.

While I agree that the label's autowrap property is a glaring omission from Button, it's just one of potentially hundreds of properties we need to access regarding the label. I suggest exposing the label in the buttons API, so all the properties can be accessed at least through code. Currently, even subclassing Button doesn't give you access to the label.

YuriSizov:
@klootas Buttons don't use labels internally. They draw the text directly.

klootas:
> @klootas Buttons don't use labels internally. They draw the text directly.

Oh, I see - thanks for the clarification. Then I think this PR fixes a pretty big shortcoming in a really good way.👍

akien-mga (approved):
Looks good to me. Needs a rebase.

akien-mga:
> Needs a rebase.

Poke :)

akien-mga:
Thanks!

HunterWilkins:
Was this added at all in Godot 4.2.2 .NET? Not seeing anything equating what this PR is surrounding in my editor.
godot 4 no button text wrap
There's only the "Text Overrun Behavior," which only allows for trimming, not wrapping.

PR #87026, opened 2024-01-10
Refactor parsing of integers

odriansu:
Ensure more consistent behavior across these three methods, including improved overflow and underflow detection.

This PR addresses:

In some cases, error messages are not handled properly.
Large numbers produce excessive error message output. This reduces output to just one.

In methods `to_int(const char ...)` and `to_int(const wchar_t ...)`, there is an issue with handling characters that include a '+' sign and a decimal point, such as +12.34, resulting in 0 instead of 12. However, -12.34 does yield the expected result of -12.

The method `to_int(const char32_t ...)` may experience out-of-range access in *p_str when the parameter p_len doesn't match the size. Additionally, when p_clamp is set to true, the function encounters an issue near max or min values where the sign may be changed. For instance, 9223372036854775809 results in -9223372036854775807 instead of 9223372036854775807.

### DIFF
--- a/core/string/ustring.cpp
+++ b/core/string/ustring.cpp
@@ -2372,66 +2372,94 @@ int64_t String::to_int() const {
 	return integer * sign;
 }
 
-int64_t String::to_int(const char *p_str, int p_len) {
-	int to = 0;
-	if (p_len >= 0) {
-		to = p_len;
-	} else {
-		while (p_str[to] != 0 && p_str[to] != '.') {
-			to++;
-		}
+template <class C>
+int64_t parse_integer(const C *p_str, int p_len, bool p_clamp) {
+	int str_index = 0;
+	C src_char = p_str[str_index];
+
+	// String is empty no conversion can be performed.
+	if (p_len == 0 || src_char == '\0') {
+		return 0;
 	}
 
-	int64_t integer = 0;
+	// Determine if the entire digit sequence should be read.
+
+	int end_pos = 0;
+	bool complete_seq = true;
+	if (p_len > 0) {
+		end_pos = p_len; // Parsing will stop at the provided position.
+		complete_seq = false;
+	}
+
+	// Check the sign at the beginning of the string.
+
 	int64_t sign = 1;
+	if (src_char == '-') {
+		sign = -1;
+		str_index++;
+	} else if (src_char == '+') {
+		str_index++;
+	}
+	src_char = p_str[str_index];
 
-	for (int i = 0; i < to; i++) {
-		char c = p_str[i];
-		if (is_digit(c)) {
-			bool overflow = (integer > INT64_MAX / 10) || (integer == INT64_MAX / 10 && ((sign == 1 && c > '7') || (sign == -1 && c > '8')));
-			ERR_FAIL_COND_V_MSG(overflow, sign == 1 ? INT64_MAX : INT64_MIN, "Cannot represent " + String(p_str).substr(0, to) + " as a 64-bit signed integer, since the value is " + (sign == 1 ? "too large." : "too small."));
-			integer *= 10;
-			integer += c - '0';
+	const int64_t max_div_10 = INT64_MAX / 10;
+	const int64_t min_div_10 = INT64_MIN / 10;
+	int64_t digit = 0;
+	int64_t integer = 0;
+	bool is_out_of_range = false;
 
-		} else if (c == '-' && integer == 0) {
-			sign = -sign;
-		} else if (c != ' ') {
+	// Parsing stops upon detecting overflow/underflow, reaching the specified end,
+	// or after reading the entire digit sequence.
+	while (is_digit(src_char)) {
+		digit = (src_char - '0') * sign;
+
+		if (str_index == end_pos && !complete_seq) {
 			break;
 		}
+
+		// Detect integer overflow/underflow
+		if (integer > max_div_10 || integer < min_div_10) {
+			is_out_of_range = true;
+			break;
+		}
+
+		// Handle edge cases for overflow/underflow
+		if ((integer == max_div_10 || integer == min_div_10) && (digit > 7 || digit < -8)) {
+			is_out_of_range = true;
+			break;
+		}
+
+		integer = integer * 10 + digit;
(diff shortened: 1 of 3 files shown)


core/string/ustring.cpp
```diff
@@ -2372,66 +2372,94 @@ int64_t String::to_int() const {
...
+		digit = (src_char - '0') * sign;
+
+		if (str_index == end_pos && !complete_seq) {
 			break;
 		}
-	}
 
-	return integer * sign;
-}
+		// Detect integer overflow/underflow
```
AThousandShips, line 2420:
```suggestion
		// Detect integer overflow/underflow.
```

core/string/ustring.cpp
```diff
@@ -2372,66 +2372,94 @@ int64_t String::to_int() const {
...
+		}
 
-int64_t String::to_int(const wchar_t *p_str, int p_len) {
-	int to = 0;
-	if (p_len >= 0) {
-		to = p_len;
-	} else {
-		while (p_str[to] != 0 && p_str[to] != '.') {
-			to++;
+		// Handle edge cases for overflow/underflow
```
AThousandShips, line 2426:
```suggestion
		// Handle edge cases for overflow/underflow.
```

core/string/ustring.h
```diff
@@ -332,10 +332,15 @@ class String {
...
 	int64_t bin_to_int() const;
+
+	// Removes any non-numeric characters and stops at the first occurrence of a decimal point.
 	int64_t to_int() const;
 
-	static int64_t to_int(const char *p_str, int p_len = -1);
-	static int64_t to_int(const wchar_t *p_str, int p_len = -1);
+	// Parsing stops at the first non-digit character. If p_len is negative,
+	// parsing continues until the end of the digit sequence. Additionally, when p_clamp is true,
+	// the integer is clamped to the range of int64_t (+9223372036854775807, -9223372036854775808)
```
AThousandShips, line 341:
```suggestion
	// the integer is clamped to the range of int64_t (+9223372036854775807, -9223372036854775808).
```

tests/core/string/test_string.h
```diff
@@ -469,6 +471,28 @@ TEST_CASE("[String] String to integer") {
 	CHECK(String("999999999999999999999999999999999999999999999999999999999").to_int() == INT64_MAX); // Too large, largest possible is returned.
 	CHECK(String("-999999999999999999999999999999999999999999999999999999999").to_int() == INT64_MIN); // Too small, smallest possible is returned.
 	ERR_PRINT_ON
+
+	// This method stops at the first non-digit character.
+
+	static const char *nums_char[14] = { "1237461283", "-22", "0", "-1123412", "", "10_000_000", "-1_2_3_4", "-9223372036854775808", "+12.34 ", "-0", "007", "-+45", "-46/46/46", "-7-2" };
+	static const wchar_t *nums_wchar[14] = { L"1237461283", L"-22", L"0", L"-1123412", L"", L"10_000_000", L"-1_2_3_4", L"-9223372036854775808", L"+12.34 ", L"-0", L"007", L"-+45", L"-46/46/46", L"-7-2" };
+	static const char32_t *nums_char32[14] = { U"1237461283", U"-22", U"0", U"-1123412", U"", U"10_000_000", U"-1_2_3_4", U"-9223372036854775808", U"+12.34 ", U"-0", U"007", U"-+45", U"-46/46/46", U"-7-2" };
+	static const int64_t integers[14] = { 1237461283, -22, 0, -1123412, 0, 10, -1, -9223372036854775808, 12, 0, 7, 0, -46, -7 };
```
AThousandShips, line 480:
```suggestion
	static const int64_t integers[14] = { 1237461283, -22, 0, -1123412, 0, 10, -1, -9223372036854775808LL, 12, 0, 7, 0, -46, -7 };
```

tests/core/string/test_string.h
```diff
@@ -469,6 +471,28 @@ TEST_CASE("[String] String to integer") {
 	CHECK(String("999999999999999999999999999999999999999999999999999999999").to_int() == INT64_MAX); // Too large, largest possible is returned.
 	CHECK(String("-999999999999999999999999999999999999999999999999999999999").to_int() == INT64_MIN); // Too small, smallest possible is returned.
 	ERR_PRINT_ON
+
+	// This method stops at the first non-digit character.
+
+	static const char *nums_char[14] = { "1237461283", "-22", "0", "-1123412", "", "10_000_000", "-1_2_3_4", "-9223372036854775808", "+12.34 ", "-0", "007", "-+45", "-46/46/46", "-7-2" };
+	static const wchar_t *nums_wchar[14] = { L"1237461283", L"-22", L"0", L"-1123412", L"", L"10_000_000", L"-1_2_3_4", L"-9223372036854775808", L"+12.34 ", L"-0", L"007", L"-+45", L"-46/46/46", L"-7-2" };
+	static const char32_t *nums_char32[14] = { U"1237461283", U"-22", U"0", U"-1123412", U"", U"10_000_000", U"-1_2_3_4", U"-9223372036854775808", U"+12.34 ", U"-0", U"007", U"-+45", U"-46/46/46", U"-7-2" };
+	static const int64_t integers[14] = { 1237461283, -22, 0, -1123412, 0, 10, -1, -9223372036854775808LL, 12, 0, 7, 0, -46, -7 };
```
AThousandShips, line 480:
```suggestion
	static const int64_t integers[14] = { 1237461283, -22, 0, -1123412, 0, 10, -1, -9223372036854775808uLL, 12, 0, 7, 0, -46, -7 };
```
Possible that the negation is unary and not part of the constant, but strange

odriansu: > Possible that the negation is unary and not part of the constant, but strange

i think should be (-9223372036854775807LL -1) as it contains literals that are within the range.

core/string/ustring.h
AThousandShips, line 341:
```suggestion
	// the integer is clamped to the range of int64_t [-9223372036854775808, +9223372036854775807, ].
```
To not be ambiguous, and order from start to end (with general interval syntax this would be excluding the ends of that range)

PR #87029, opened 2024-01-10
[macOS/iOS export] Add option to set custom Info.plist data.

### LINKED ISSUE #8848: Add custom Info.plist key additions field to iOS and macOS export settings

An editor, exporting to multiple formats including an internal file format I want to associate with the editor app.

On both iOS and macOS [many application meta-properties](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Introduction/Introduction.html) are controlled by the Info.plist file in the exported application bundle. This is generated by Godot at export based on the export template, and populated by many of the values you can fill in in the export settings.

I've opened a number of related issues previously (like https://github.com/godotengine/godot/issues/74154, https://github.com/godotengine/godot-proposals/issues/8289, https://github.com/godotengine/godot-proposals/issues/6360) that I'm happy to see have been implemented improving export for iOS and macOS. But now I find myself once again needing to make some additions to the Info.plist. 

However, this time the change is a bit more specific and experimental and I don't think these specific keys will be widely used, so I don't think it warranted to add specific fields for these specific properties.  (I'm attempting to add declarations for <key>UTExportedTypeDeclarations</key> and <key>CFBundleDocumentTypes</key> to define file formats that should be associated with the exported app as the owner). But this means I have to export the project unsigned, manually modify the Info.plist and then manually sign, package and notarize the project each time. 

So, what I'm proposing is to instead add a new "Custom Info.plist keys" field in the export options for iOS and macOS, for example at the end of the XCode section. This field would allow advanced users to simply add additional keys content they want to be added at the end of the Info.plist file during the export process, but before the bundle is signed, packaged, notarized etc. 

For example you could add something like the document type:
```
<key>CFBundleDocumentTypes</key>
<array>
	<dict>
		<key>CFBundleTypeExtensions</key>
		<array>
			<string>ext</string>
		</array>
		<key>CFBundleTypeIconFile</key>
		<string>DocIcon.icns</string>
		<key>CFBundleTypeRole</key>
		<string>Editor</string>
		<key>LSHandlerRank</key>
		<string>Owner</string>
		...
	</dict>
</array>
```
Or something like the [iOS device required performance capabilities](https://developer.apple.com/documentation/bundleresources/information_property_list/uirequireddevicecapabilities?language=objc)

This would allow users to make seldom used additions like the cases above without need for additional specific fields in the export settings. And it would also make it much easier to test and fix any future issues like the ones linked in the introduction above related to missing plist keys. 

The basic implementation of this would be very simple, just the addition of another field (sorry @bruvzg) in the export settings for iOS and macOS. Possibly with an expanding/scrolling text field in the UI to fit longer additions. 
Then when generating the Info.plist during export all it would require to get the basic functionality would be to insert the custom Info.plist additions string at the end before the closing </dict></plist>-tags.

If you want to make it more advanced you could do some basic format checking to warn if the addition isn't properly formatted as closed plist content tags (would probably break the exported app). Or, if you want to be really advanced, allow the content of the custom plist content to overwrite and replace existing keys with the same name, letting the user use this custom field to also replace tags Godot and the export template generates. Basically allowing advanced users to replace any key in the entire Info.plist file if they want to using this field.

No, but as described above it can be worked around by adding several complicated manual steps to the export process.

Exporting is a core part of the engine

bruvzg:
Fixes https://github.com/godotengine/godot-proposals/issues/8848

### DIFF
--- a/platform/ios/doc_classes/EditorExportPlatformIOS.xml
+++ b/platform/ios/doc_classes/EditorExportPlatformIOS.xml
@@ -10,6 +10,13 @@
 		<link title="iOS plugins documentation index">$DOCS_URL/tutorials/platform/ios/index.html</link>
 	</tutorials>
 	<members>
+		<member name="application/additional_plist_content" type="String" setter="" getter="">
+			Additional data added to the root [code]&lt;dict&gt;[/code] section of the [url=https://developer.apple.com/documentation/bundleresources/information_property_list]Info.plist[/url] file. The value should be an XML section with pairs of key-value elements, e.g.:
+			[codeblock]
+			&lt;key&gt;key_name&lt;/key&gt;
+			&lt;string&gt;value&lt;/string&gt;
+			[/codeblock]
+		</member>
 		<member name="application/app_store_team_id" type="String" setter="" getter="">
 			Apple Team ID, unique 10-character string. To locate your Team ID check "Membership details" section in your Apple developer account dashboard, or "Organizational Unit" of your code signing certificate. See [url=https://developer.apple.com/help/account/manage-your-team/locate-your-team-id]Locate your Team ID[/url].
 		</member>
--- a/platform/macos/doc_classes/EditorExportPlatformMacOS.xml
+++ b/platform/macos/doc_classes/EditorExportPlatformMacOS.xml
@@ -10,6 +10,13 @@
 		<link title="Running Godot apps on macOS">$DOCS_URL/tutorials//export/running_on_macos.html</link>
 	</tutorials>
 	<members>
+		<member name="application/additional_plist_content" type="String" setter="" getter="">
+			Additional data added to the root [code]&lt;dict&gt;[/code] section of the [url=https://developer.apple.com/documentation/bundleresources/information_property_list]Info.plist[/url] file. The value should be an XML section with pairs of key-value elements, e.g.:
+			[codeblock]
+			&lt;key&gt;key_name&lt;/key&gt;
+			&lt;string&gt;value&lt;/string&gt;
+			[/codeblock]
+		</member>
 		<member name="application/app_category" type="String" setter="" getter="">
 			Application category for the App Store.
 		</member>
--- a/core/io/plist.cpp
+++ b/core/io/plist.cpp
@@ -448,7 +448,9 @@ PList::PList() {
 }
 
 PList::PList(const String &p_string) {
-	load_string(p_string);
+	String err_str;
+	bool ok = load_string(p_string, err_str);
+	ERR_FAIL_COND_MSG(!ok, "PList: " + err_str);
 }
 
 uint64_t PList::read_bplist_var_size_int(Ref<FileAccess> p_file, uint8_t p_size) {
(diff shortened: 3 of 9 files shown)


akien-mga:
It should probably mention what format is expected (maybe simply linking to the relevant Apple docs).

bruvzg:
Added link to the Apple docs, and example.

platform/macos/doc_classes/EditorExportPlatformMacOS.xml
```diff
@@ -10,6 +10,13 @@
 		<link title="Running Godot apps on macOS">$DOCS_URL/tutorials//export/running_on_macos.html</link>
 	</tutorials>
 	<members>
+		<member name="application/additional_plist_content" type="String" setter="" getter="">
+			Additional data added to the root [code]&lt;dict&gt;[/code] section of the [url=https://developer.apple.com/documentation/bundleresources/information_property_list]Info.plist[/url] file. Value should be XML section with pairs of key-value elements, e.g:
```
akien-mga, line 14:
```suggestion
			Additional data added to the root [code]&lt;dict&gt;[/code] section of the [url=https://developer.apple.com/documentation/bundleresources/information_property_list]Info.plist[/url] file. The value should be an XML section with pairs of key-value elements, e.g.:
```

platform/ios/doc_classes/EditorExportPlatformIOS.xml
```diff
@@ -10,6 +10,13 @@
 		<link title="iOS plugins documentation index">$DOCS_URL/tutorials/platform/ios/index.html</link>
 	</tutorials>
 	<members>
+		<member name="application/additional_plist_content" type="String" setter="" getter="">
+			Additional data added to the root [code]&lt;dict&gt;[/code] section of the [url=https://developer.apple.com/documentation/bundleresources/information_property_list]Info.plist[/url] file. Value should be XML section with pairs of key-value elements, e.g:
```
akien-mga, line 14:
```suggestion
			Additional data added to the root [code]&lt;dict&gt;[/code] section of the [url=https://developer.apple.com/documentation/bundleresources/information_property_list]Info.plist[/url] file. The value should be an XML section with pairs of key-value elements, e.g.:
```

akien-mga (approved):

akien-mga:
How does it behave on export if the user puts invalid data / syntax errors? Do we export gracefully and then the .app would fail running? Is the error then clear enough?

bruvzg:
> How does it behave on export if the user puts invalid data / syntax errors? Do we export gracefully and then the .app would fail running?

If it's completely invalid data, the app will run, but won't show info or have an icon. If it's valid fomatting but unknown keys/values, it will be ignored.

bruvzg:
> we could likely check that XMLParser is happy with the syntax at least. 

It won't be enough, valid XML data is not necessarily valid part of `<dict>` element. Proper validation should check if it's matching http://www.apple.com/DTDs/PropertyList-1.0.dtd schema, but I do not think we have any code for this, and it probably is not worth adding for such minor feature.

lostminds:
> It won't be enough

You're right that it won't be enough to ensure it's a valid addition. But hopefully checking that it's valid xml should at least ensure the additions are correctly closed tags so the addition doesn't risk messing up the parsing of the entire Info.plist. If we can be reasonably sure about that it's a lesser problem if the content in the addition is just ignored if it's formatted incorrectly.

akien-mga:
Yeah if it doesn't add significant complexity to the code, I think just validating that the syntax is valid XML would be enough.

bruvzg:
We actually have a dedicated PList parser for ad-hoc code signing (which can give more verbose error output than XMLParser), so I guess it probably better to use it.

bruvzg:
Added validation using PList parser from ad-hoc code signer, most likely it won't catch every error, since it's not designed to for validation, but should reject completely invalid data.

akien-mga (approved):

akien-mga:
Thanks!

KeiMuriKoe:
Where application/additional_plist_content option in iOS export?

image
image

PR #87030, opened 2024-01-10
Add documentation to PhysicsDirectBodyState2DExtension

Mickeon:
Related to https://github.com/godotengine/godot/pull/87018

This..
This PR fills in all of **PhysicsDirectBodyState2DExtension**'s class reference. What a mouthful. All of the descriptions end up just pointing back to the original **PhysicsDirectBodyState2D**. See https://github.com/godotengine/godot/pull/87018 for reasoning.

If both this PR and the prior one are acceptable, I will make a future PR for the 3D equivalents. Otherwise... I guess writing tailored documentation for these wouldn't be nearly as bad, given there's much fewer methods to talk about?

### DIFF
--- a/doc/classes/PhysicsDirectBodyState2DExtension.xml
+++ b/doc/classes/PhysicsDirectBodyState2DExtension.xml
@@ -14,6 +14,7 @@
 			<return type="void" />
 			<param index="0" name="force" type="Vector2" />
 			<description>
+				Overridable version of [method PhysicsDirectBodyState2D.add_constant_central_force].
 			</description>
 		</method>
 		<method name="_add_constant_force" qualifiers="virtual">
@@ -21,24 +22,28 @@
 			<param index="0" name="force" type="Vector2" />
 			<param index="1" name="position" type="Vector2" />
 			<description>
+				Overridable version of [method PhysicsDirectBodyState2D.add_constant_force].
 			</description>
 		</method>
 		<method name="_add_constant_torque" qualifiers="virtual">
 			<return type="void" />
 			<param index="0" name="torque" type="float" />
 			<description>
+				Overridable version of [method PhysicsDirectBodyState2D.add_constant_torque].
 			</description>
 		</method>
 		<method name="_apply_central_force" qualifiers="virtual">
 			<return type="void" />
 			<param index="0" name="force" type="Vector2" />
 			<description>
+				Overridable version of [method PhysicsDirectBodyState2D.apply_central_force].
 			</description>
 		</method>
 		<method name="_apply_central_impulse" qualifiers="virtual">
 			<return type="void" />
 			<param index="0" name="impulse" type="Vector2" />
 			<description>
+				Overridable version of [method PhysicsDirectBodyState2D.apply_central_impulse].
 			</description>
 		</method>
 		<method name="_apply_force" qualifiers="virtual">
@@ -46,6 +51,7 @@
 			<param index="0" name="force" type="Vector2" />
 			<param index="1" name="position" type="Vector2" />
 			<description>
+				Overridable version of [method PhysicsDirectBodyState2D.apply_force].
 			</description>
 		</method>
 		<method name="_apply_impulse" qualifiers="virtual">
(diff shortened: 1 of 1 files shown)


Mickeon:
I tested this prior to making this PR but the test does not seem to agree.
image
Directly pointing to setters and getters in the documentation does work correctly, even if they are grouped with their property, at least in the built-in documentation. I can only assume that the online documentation does not in fact support this.

doc/classes/PhysicsDirectBodyState2DExtension.xml
```diff
@@ -14,250 +14,294 @@
...
 			<return type="Vector2" />
 			<param index="0" name="contact_idx" type="int" />
 			<description>
+				Overridable version of [method PhysicsDirectBodyState2D.get_contact_local_velocity_at_position].
 			</description>
 		</method>
 		<method name="_get_inverse_inertia" qualifiers="virtual const">
 			<return type="float" />
 			<description>
+				Overridable version of [method PhysicsDirectBodyState2D.get_inverse_inertia].
```
AThousandShips, line 195: These aren't exposed as distinct methods but as properties, so should probably be:
```suggestion
				Overridable version of getter for [member PhysicsDirectBodyState2D.inverse_inertia].
```

Mickeon: Yeah, I became aware of it. The thing is, even if they're not distinct, the built-in documentation will recognise and point to the corresponding property just fine. Hence this caught me off-guard.

AThousandShips: I'd make a note of it and open an issue to investigate improvements to the docgen

Mickeon: Would it be on godot or godot-docs?

AThousandShips: Here I'd say as the script is here

YuriSizov: I think the description can be more detailed to begin with. It can say that it overrides the behavior of such and such property and its respective getter, for example. Can also suggest why this would be needed.

Mickeon:
Applied the new description to all methods that are setters/getters of a property in **PhysicsDirectBodyState2D**.

YuriSizov:
For the rest "Overridable version of ..." still stands out to me. So I did some lookup, and what we have is:

---

Virtual method to be implemented by the user. ...
Virtual method to be overridden by the user. ...
Virtual method which can be overridden to ...
Virtual method to override ...
Virtual method which can be overridden to ...

Override this method to ...

Implement to ...
Implement this in GDExtension to ...

---

Looks like my suggestion above wasn't the most consistent option. The most popular unique variation seems to be "Override this method to".

Mickeon:
> Looks like my suggestion above wasn't the most consistent option. The most popular unique variation seems to be "Override this method to". And "Implement this in GDExtension to" is used specifically by another extension class.

Indeed, "Override this method to" is the most consistent and also my favourite, I always try to word it that way.
But for all of these *Extension classes, which already begin their description with at least something along the lines of:

> "This class extends [PhysicsDirectBodyState2D] by providing additional virtual methods that can be overridden. When these methods are overridden, they will be called instead of the internal methods of the physics server."

One could also argue that introducing the method that way is needlessly verbose, too.

YuriSizov:
> One could also argue that introducing the method that way is needlessly verbose, too.

True, I would say that we almost don't even need descriptions, we just need a link to the non-extension class to explain what each methods connects to. Which can probably be automated somehow.

I would add that calling it an overridable version is a bit of a misnomer, since that's not exactly how virtual methods work (at least not in every case). But eh, we can live with this as is for now, we shouldn't really be using this PR to harmonize all these descriptions.

Mickeon:
I'm okay with coming up with a standard description as a good starting point, so that it can be applied to https://github.com/godotengine/godot/pull/87018, as well. If it has to be with "that" description, so be it.

Mickeon:
How do we feel about this? Should the description be reverted back to what it was for consistency, at least?

YuriSizov:
@Mickeon Ultimately it doesn't matter much as we're all over the place with these descriptions. So if content-wise this is fine, then we should just merge it as is.

mhilbrunner (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

akien-mga:
Cherry-picked for 4.1.4.

Issue #32996, opened 2019-10-22
TextEdit always clips content


somnivore:
**Godot version:**
3564547208030264a28cc55232992aab3cbdad94

**OS/device including version:**
ArcoLinux

**Issue description:**
TextEdit forcibly clips all content, regardless of whether the `Clip Content` box is checked.

![image](https://user-images.githubusercontent.com/28724859/67325878-6faaf200-f4db-11e9-8bde-2ea76b8f37fa.png)
![image](https://user-images.githubusercontent.com/28724859/67325902-78032d00-f4db-11e9-8e1a-2bc4a4a0d35e.png)

YeldhamDev:
I'm quite sure `TextEdit` wouldn't even work properly without clipping.

codecustard:
Clip Content should probably default to true in `TextEdit` and the following line should be changed to reflect the clip content property:
https://github.com/godotengine/godot/blob/72dfa67dd3508f20592959d306e8f2ef159eea5f/scene/gui/text_edit.cpp#L725

girng:
Should probably use a different node or at least put that TextureRect inside another node and container it

mitsuan:
can I contribute to this project please.

ondesic:
I agree that it should be set to TRUE be default, but allow the user to turn off clipping.  As a side not, LineEdit allows no clipping, and it looks great.

Calinou:
I can confirm this on commit 4.0.dev.custom_build [b5b633ff5]. In particular, I need this to display a focus outline that's slightly *outside* the TextEdit. Removing the clipping function call in the TextEdit C++ code works, but it likely has unintended side-effects :slightly_smiling_face:

KoBeWi:
This happens when you disable clipping:
![image](https://user-images.githubusercontent.com/2223172/146410412-36a08b33-4565-405b-a5d7-a89d0514b76c.png)
Clipping is global for the CanvasItem, you can't disable it only for some parts. So to make this work properly, we'd need to move all text rendering to a separate canvas item (server-side). Similar thing is done for quadrants in TileMap.

kitbdev:
- Clip contents can be disabled as of https://github.com/godotengine/godot/pull/89993

PR #87033, opened 2024-01-10
Enable content clipping for LineEdit, add TextServer option to clamp glyph offsets.

### LINKED ISSUE #8860: Clip LineEdit overflow characters instead of trimming for improved usability

An editor where the user can edit parameter values in line edit fields

When a string is too long to fit in a LineEdit the overflow characters are trimmed (as in the whole character is removed if any part of it extends outside the field) instead of clipped (as in the last character is cut off). This is an issue because it makes it appear as the field has another content, with no hint to the user that there is overflow that doesn't fit in the text.

https://github.com/godotengine/godot-proposals/assets/17763524/8ef4db8b-2161-4ac8-92e5-a445bfef63b5

It is especially problematic when displaying values with trailing decimals that are trimmed, making 10.01 look like the value is 10.0 (or even 0.01) based on the last caret position and alignment as you can see above.

Switch to instead clipping the line edit content, so that the user can see cut off overflow characters as an indication that there is text that doesn't fit in the field. 
Add a setting "trim characters" matching that of the current Label trim parameter and set that to Trim Characters by default (at least for importing older versions) to avoid compatibility breakage.

I assume that internally the LineEdit uses something like a Label node for rendering the text content, so the change should just be a matter of using the existing Label trim settings of this label component to clip instead of trim the text in the LineEdit.

Well, I guess you could just add a separate clipped Label node on top of the LineEdit node to do the rendering, and have that display the text instead with the clipping, setting the line edit to render the text transparent. But this would be complicated to keep synced with the caret moving, selection and scrolling in the field. And it feels very much like a hack for something that should be a standard feature of the LineEdit.

The LineEdit is a core component

### LINKED ISSUE #87016: Accent marks render outside of LineEdit's `text` parameter in the inspector when many are added
### Tested versions

### System information

Ubuntu 20.04 LTS

### Issue description

In the inspector. I held down `Q` for a while, so there are many accent marks.
arabicbug_godot

The 2D scene looks correct though
arabicbug_2dscene

For more comparison...
My browser URL bar (10 accent marks; it seems to stack them in-place)
arabicbug_urlbar

VIM (10 accent marks; it seems to only show the first two)
arabicbug_vim

### Steps to reproduce

Open the MRP, open `line_edit.tscn`, click on the `LineEdit` node, and look at the inspector.

The steps I took to create the MRP:
Follow these instructions at your own risk since they involve installing and switching to a new key layout.
1. Ensure you can input Arabic.  On Ubuntu 20 Settings -> Region & Language -> Manage Installed Languages -> Install/Remove Languages... -> Choose Arabic and let it install.  Once it is done, close and reopen Settings.  Then go to Region & Language -> Click on the + sign -> Choose Arabic.  In your status bar, you should see `en`, click on it to switch to `ar`.
2. Create a LineEdit node
3. Type any character.  For this example I typed single-quote (`'`) which produced `ط` in Arabic.
4. Type `Q`, this will add an accent above it, like this: `طَ`
5. Hold down `Q`, and this will continuously add more accents.  Watch it clip out of `LineEdit`.

Notice that the rendered text in the 2D scene looks correct, which may mean this is an issue with the editor's font.

[ArabicBug.zip](https://github.com/godotengine/godot/files/13878296/ArabicBug.zip)

bruvzg:
Fixes https://github.com/godotengine/godot/issues/87016
Fixes https://github.com/godotengine/godot-proposals/issues/8860

- Enable content clipping for `LineEdit` (via separate canvas item for the text).
- Add `TextServer` option to clamp glyph offsets (global default, editor setting, and API for overriding it in the custom drawing).

| Before | After (no clamping) | After (with clamping set to `1.0`) |
| --- | --- | --- |
| Screenshot 2024-01-10 at 10 42 53 | Screenshot 2024-01-10 at 11 16 59 | Screenshot 2024-01-10 at 11 34 11 |

### DIFF
--- a/scene/gui/line_edit.cpp
+++ b/scene/gui/line_edit.cpp
@@ -819,6 +819,12 @@ void LineEdit::_notification(int p_what) {
 				style->draw(ci, Rect2(Point2(), size));
 			}
 
+			Rect2 clip_rect = Rect2(Point2(style->get_margin(SIDE_LEFT), style->get_margin(SIDE_TOP)), size - Vector2(style->get_margin(SIDE_LEFT) + style->get_margin(SIDE_RIGHT), style->get_margin(SIDE_TOP) + style->get_margin(SIDE_BOTTOM)));
+			RenderingServer::get_singleton()->canvas_item_clear(canvas_item_text);
+			RenderingServer::get_singleton()->canvas_item_set_custom_rect(canvas_item_text, !is_visibility_clip_disabled(), clip_rect);
+			RenderingServer::get_singleton()->canvas_item_set_clip(canvas_item_text, true);
+			RenderingServer::get_singleton()->canvas_item_set_visibility_layer(canvas_item_text, get_visibility_layer());
+
 			if (has_focus()) {
 				theme_cache.focus->draw(ci, Rect2(Point2(), size));
 			}
@@ -927,9 +933,9 @@ void LineEdit::_notification(int p_what) {
 				Vector2 oofs = ofs;
 				for (int i = 0; i < gl_size; i++) {
 					for (int j = 0; j < glyphs[i].repeat; j++) {
-						if (ceil(oofs.x) >= x_ofs && (oofs.x + glyphs[i].advance) <= ofs_max) {
+						if (ceil(oofs.x + glyphs[i].advance) >= x_ofs && (oofs.x) <= ofs_max) {
 							if (glyphs[i].font_rid != RID()) {
-								TS->font_draw_glyph_outline(glyphs[i].font_rid, ci, glyphs[i].font_size, outline_size, oofs + Vector2(glyphs[i].x_off, glyphs[i].y_off), glyphs[i].index, font_outline_color);
+								TS->font_draw_glyph_outline(glyphs[i].font_rid, canvas_item_text, glyphs[i].font_size, outline_size, oofs + Vector2(glyphs[i].x_off, glyphs[i].y_off), glyphs[i].index, font_outline_color);
 							}
 						}
 						oofs.x += glyphs[i].advance;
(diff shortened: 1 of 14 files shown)


doc/classes/EditorSettings.xml
```diff
@@ -576,6 +576,9 @@
 		<member name="interface/editor/font_antialiasing" type="int" setter="" getter="">
 			FreeType's font anti-aliasing mode used to render the editor fonts. Most fonts are not designed to look good with anti-aliasing disabled, so it's recommended to leave this enabled unless you're using a pixel art font.
 		</member>
+		<member name="interface/editor/font_clamp_glyph_offset" type="float" setter="" getter="">
+			Sets default value for glyph offset clamping. Value is a fraction of font height. If set to zero or negative value, no clamping is done.
```
Calinou, line 580: Out of curiosity, why does this need to be an editor setting (as opposed to say, a project setting)?

scene/gui/line_edit.cpp
```diff
@@ -810,6 +810,8 @@ void LineEdit::_notification(int p_what) {
 
 			RID ci = get_canvas_item();
 
+			RenderingServer::get_singleton()->canvas_item_set_clip(ci, true);
```
Calinou, line 813: Note that this may negatively affect focus StyleBox drawing, in a manner similar to TextEdit: https://github.com/godotengine/godot/issues/32996

doc/classes/EditorSettings.xml
bruvzg: I guess project setting also should be added.

scene/gui/line_edit.cpp
Kiisu-Master: Not only the focus stylebox but also children of the node.

lostminds:
After looking at the code a little closer I see now that LineEdit does its own rendering of text, would it not be possible to have the LineEdit maintain an internal RichTextLabel child node for this purpose? A regular Label would probably be enough in most cases, but a RichTextLabel would allow rendering separate selection color etc. This label node could then be clipped separately, solving this issue and https://github.com/godotengine/godot-proposals/issues/8860 without needing to make the entire LineEdit control clip children with the issues that would entail as pointed out above? It could also reduce some code duplication as the text rendering code would no longer be needed in LineEdit.

bruvzg:
Adding a full Label is overkill, but adding a separate canvas item to draw should probably work.

doc/classes/EditorSettings.xml
bruvzg: Moved to project settings.

bruvzg:
Added a separate canvas item for the text, now it should clip only text. Also changed clipping check to implement https://github.com/godotengine/godot-proposals/issues/8860 as well.

dhoverml:
I tried cherry-picking this to master (f317cc713a), but got two merge conflicts in `modules/text_server_adv/text_server_adv.cpp` (I believe with 85df221610e)

Issue #83762, opened 2023-10-22
C# Game crashes randomly on changing scenes with USER ERROR: FATAL: Condition "gchandle.is_released()" is true


MattSayer:
### Godot version

v4.1.1.stable.mono.official.bd6af8e0e

### System information

Godot v4.1.1.stable.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3080 (NVIDIA; 31.0.15.3623) - Intel(R) Core(TM) i7-10700KF CPU @ 3.80GHz (16 Threads)

### Issue description

When switching scenes in my C# project, randomly the game will crash with the following stack trace:

![crashlog](https://github.com/godotengine/godot/assets/112981870/7f6ff8c3-bc94-42d9-91a6-abbfb106826c)

USER ERROR: FATAL: Condition "gchandle.is_released()" is true.
   at: mono_object_disposed_baseref (modules/mono/csharp_script.cpp:1835)

In the console output, this occurs as well:

Cannot open file '/root/godot/modules/mono/glue/GodotSharp/GodotSharp/Godot.SourceGenerators.Internal/Godot.SourceGenerators.Internal.UnmanagedCallbacksGenerator/Godot.NativeInterop.NativeFuncs.generated.cs'.
  Failed to read file: '/root/godot/modules/mono/glue/GodotSharp/GodotSharp/Godot.SourceGenerators.Internal/Godot.SourceGenerators.Internal.UnmanagedCallbacksGenerator/Godot.NativeInterop.NativeFuncs.generated.cs'.
  Cannot load C# script file '/root/godot/modules/mono/glue/GodotSharp/GodotSharp/Godot.SourceGenerators.Internal/Godot.SourceGenerators.Internal.UnmanagedCallbacksGenerator/Godot.NativeInterop.NativeFuncs.generated.cs'.
  Failed loading resource: /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Godot.SourceGenerators.Internal/Godot.SourceGenerators.Internal.UnmanagedCallbacksGenerator/Godot.NativeInterop.NativeFuncs.generated.cs. Make sure resources have been imported by opening the project in the editor at least once.

[...]

PR #87045, opened 2024-01-10
C#: Fix debug crashes due to check released pointers over weak ones

dsh0416:
fixes debug crashes described in https://github.com/godotengine/godot/issues/83762#issuecomment-1885147082, which `CRASH_COND` check released pointers over weak ones that might be released by CLR.

### DIFF
--- a/modules/mono/csharp_script.cpp
+++ b/modules/mono/csharp_script.cpp
@@ -1856,10 +1856,8 @@ void CSharpInstance::mono_object_disposed(GCHandleIntPtr p_gchandle_to_free) {
 }
 
 void CSharpInstance::mono_object_disposed_baseref(GCHandleIntPtr p_gchandle_to_free, bool p_is_finalizer, bool &r_delete_owner, bool &r_remove_script_instance) {
-#ifdef DEBUG_ENABLED
-	CRASH_COND(!base_ref_counted);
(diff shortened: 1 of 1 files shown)


AThousandShips:
I'd suggest replacing the whole section here like so:
```cpp
#ifdef DEBUG_ENABLED
	CRASH_COND(!base_ref_counted);
	CRASH_COND(!gchandle.is_weak() && gchandle.is_released());
#endif
```
With:
```cpp
	DEV_ASSERT(base_ref_counted);
	DEV_ASSERT(gchandle.is_weak() || !gchandle.is_released());
```

AThousandShips:
My suggestions were incorrect, so should be reverted, that's why I deleted them, sorry for not pointing that out

dsh0416:
> My suggestions were incorrect, so should be reverted, that's why I deleted them, sorry for not pointing that out

reverted.

rburing:
To align with our workflow the commits need to be squashed so there is just one commit with the intended change (then force push to your branch).

medzernik:
Is there any progress on this?

paulloz:
> Is there any progress on this?

We believe this does not fix the core issue (although it hides the symptoms by bypassing the crash condition). So the issue needs to be investigated further.

akien-mga:
Closing as per above comment.

PR #87047, opened 2024-01-10
Add missing descriptions to remaining signals

Mickeon:
This PR adds descriptions to signals which were previously missing one.
These are:
- **GraphEdit**'s `node_deselected` signal.
- **Skeleton3D**'s `bone_enabled_changed`, `pose_updated` and `show_rest_only_changed` signals.
     - And by extension, the `show_rest_only` property and the `NOTIFICATION_UPDATE_SKELETON` constant.

It also tweaks the descriptions of surrounding signals very slightly.

This also would make all signals in the class reference documented. 100%.

### DIFF
--- a/doc/classes/GraphEdit.xml
+++ b/doc/classes/GraphEdit.xml
@@ -301,7 +301,7 @@
 	<signals>
 		<signal name="begin_node_move">
 			<description>
-				Emitted at the beginning of a GraphNode movement.
+				Emitted at the beginning of a [GraphElement]'s movement.
 			</description>
 		</signal>
 		<signal name="connection_drag_ended">
@@ -344,13 +344,14 @@
 		</signal>
 		<signal name="copy_nodes_request">
 			<description>
-				Emitted when the user presses [kbd]Ctrl + C[/kbd].
+				Emitted when this [GraphEdit] captures a [code]ui_copy[/code] action ([kbd]Ctrl + C[/kbd] by default). In general, this signal indicates that the selected [GraphElement]s should be copied.
 			</description>
 		</signal>
 		<signal name="delete_nodes_request">
 			<param index="0" name="nodes" type="StringName[]" />
 			<description>
-				Emitted when attempting to remove a GraphNode from the GraphEdit. Provides a list of node names to be removed (all selected nodes, excluding nodes without closing button).
+				Emitted when this [GraphEdit] captures a [code]ui_graph_delete[/code] action ([kbd]Delete[/kbd] by default).
+				[param nodes] is an array of node names that should be removed. These usually include all selected nodes.
 			</description>
 		</signal>
 		<signal name="disconnection_request">
@@ -364,28 +365,29 @@
 		</signal>
 		<signal name="duplicate_nodes_request">
 			<description>
-				Emitted when a GraphNode is attempted to be duplicated in the GraphEdit.
+				Emitted when this [GraphEdit] captures a [code]ui_graph_duplicate[/code] action ([kbd]Ctrl + D[/kbd] by default). In general, this signal indicates that the selected [GraphElement]s should be duplicated.
 			</description>
 		</signal>
 		<signal name="end_node_move">
 			<description>
-				Emitted at the end of a GraphNode movement.
+				Emitted at the end of a [GraphElement]'s movement.
 			</description>
 		</signal>
 		<signal name="node_deselected">
 			<param index="0" name="node" type="Node" />
 			<description>
+				Emitted when the given [GraphElement] node is deselected.
 			</description>
 		</signal>
 		<signal name="node_selected">
 			<param index="0" name="node" type="Node" />
 			<description>
-				Emitted when a GraphNode is selected.
+				Emitted when the given [GraphElement] node is selected.
 			</description>
 		</signal>
 		<signal name="paste_nodes_request">
 			<description>
-				Emitted when the user presses [kbd]Ctrl + V[/kbd].
+				Emitted when this [GraphEdit] captures a [code]ui_paste[/code] action ([kbd]Ctrl + V[/kbd] by default). In general, this signal indicates that previously copied [GraphElement]s should be pasted.
 			</description>
 		</signal>
 		<signal name="popup_request">
--- a/doc/classes/Skeleton3D.xml
+++ b/doc/classes/Skeleton3D.xml
@@ -310,12 +310,14 @@
 			[b]Note:[/b] Unless this value is [code]1.0[/code], the key value in animation will not match the actual position value.
 		</member>
 		<member name="show_rest_only" type="bool" setter="set_show_rest_only" getter="is_show_rest_only" default="false">
+			If [code]true[/code], forces the bones in their default rest pose, regardless of their values. In the editor, this also prevents the bones from being edited.
 		</member>
 	</members>
 	<signals>
 		<signal name="bone_enabled_changed">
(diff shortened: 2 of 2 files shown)


doc/classes/GraphEdit.xml
```diff
@@ -375,12 +375,13 @@
...
 			<param index="0" name="node" type="Node" />
 			<description>
+				Emitted when the GraphNode [param node] is deselected.
 			</description>
 		</signal>
 		<signal name="node_selected">
 			<param index="0" name="node" type="Node" />
 			<description>
-				Emitted when a GraphNode is selected.
+				Emitted when the GraphNode [param node] is selected.
```
YuriSizov, line 384: Both of these should reference `GraphElement`, not `GraphNode`. Same is true for `begin_node_move`/`end_node_move`.

`copy_nodes_request`, `paste_nodes_request`, and `duplicate_nodes_request` don't have any type checks and are triggered on input actions, so their descriptions should also refer to `GraphElement`, I think. (Also `duplicate_nodes_request`'s description is awkwardly phrased).

`delete_nodes_request` is also triggered by an action, but it explicitly passes a collection of nodes for some reason, and those indeed are `GraphNode` and not `GraphElement`. Not sure if this makes sense (and this is something we can still break compatibility for). But for now the description makes sense there, but it should be reworded, as it refers to a single node and to having a closing button.

YuriSizov: All connection related signals are `GraphNode` specific, so those are probably all fine.

Mickeon: Always amusing to notice poorly worded descriptions that are so poorly worded they're actually misleading.
I should only adjust a few without going out of scope for the PR, probably.

YuriSizov: I think adjusting descriptions of those would be in scope (you can basically rely on the same template for all of them). But doing any changes to the graph nodes themselves to address inconsistencies is of course out of scope, just me musing.

Mickeon:
I gave my best shot at updating the descriptions of those few surrounding methods too, to give them slightly more detail. But, I have little knowledge on them at the moment, and they may be factually incorrect.

doc/classes/GraphEdit.xml
```diff
@@ -344,13 +344,14 @@
 		</signal>
 		<signal name="copy_nodes_request">
 			<description>
-				Emitted when the user presses [kbd]Ctrl + C[/kbd].
+				Emitted when this [GraphEdit] requests [GraphElement] nodes to be copied, such as by pressing [kbd]Ctrl + C[/kbd] (the [code]ui_copy[/code] action).
```
AThousandShips, line 347:
```suggestion
				Emitted when this [GraphEdit] requests [GraphElement] nodes to be copied, for example by pressing [kbd]Ctrl + C[/kbd] (the [code]ui_copy[/code] action).
```
"Such as" feels odd here

Mickeon:
Updated the PR to replace those "such as" with "for example".

doc/classes/GraphEdit.xml
```diff
@@ -344,13 +344,14 @@
 		</signal>
 		<signal name="copy_nodes_request">
 			<description>
-				Emitted when the user presses [kbd]Ctrl + C[/kbd].
+				Emitted when this [GraphEdit] requests [GraphElement] nodes to be copied, for example by pressing [kbd]Ctrl + C[/kbd] (the [code]ui_copy[/code] action).
```
YuriSizov, line 347: Okay, I think this is a bit of a weird phrasing, because what it says is that a node is requesting something by pressing keys on the keyboard. 🙃 How about

```suggestion
				Emitted when a [code]ui_copy[/code] action is captured by this [GraphEdit] ([kbd]Ctrl + C[/kbd] by default). In general, you are expected to handle this signal only for selected [GraphElement]s.
```

doc/classes/GraphEdit.xml
```diff
@@ -344,13 +344,14 @@
...
 			</description>
 		</signal>
 		<signal name="delete_nodes_request">
 			<param index="0" name="nodes" type="StringName[]" />
 			<description>
-				Emitted when attempting to remove a GraphNode from the GraphEdit. Provides a list of node names to be removed (all selected nodes, excluding nodes without closing button).
+				Emitted when this [GraphEdit] requests [GraphNode] nodes to be deleted, for example by pressing [kbd]Delete[/kbd] (the [code]ui_graph_delete[/code] action).
```
YuriSizov, line 353: Same as above, but without the second sentence (it's handled by the following paragraph).

doc/classes/GraphEdit.xml
```diff
@@ -364,28 +365,29 @@
 		</signal>
 		<signal name="duplicate_nodes_request">
 			<description>
-				Emitted when a GraphNode is attempted to be duplicated in the GraphEdit.
+				Emitted when this [GraphEdit] requests [GraphElement] nodes to be duplicated, for example by pressing [kbd]Ctrl + D[/kbd] (the [code]ui_graph_duplicate[/code] action).
```
YuriSizov, line 368: Same here, both sentences.

doc/classes/GraphEdit.xml
```diff
@@ -364,28 +365,29 @@
...
 		<signal name="end_node_move">
 			<description>
-				Emitted at the end of a GraphNode movement.
+				Emitted at the end of a [GraphElement]'s movement.
 			</description>
 		</signal>
 		<signal name="node_deselected">
 			<param index="0" name="node" type="Node" />
 			<description>
+				Emitted when the [GraphElement] [param node] is deselected.
```
YuriSizov, line 379:
```suggestion
				Emitted when the given [GraphElement] node is deselected.
```
I think this avoids punning.

doc/classes/GraphEdit.xml
```diff
@@ -364,28 +365,29 @@
...
 			</description>
 		</signal>
 		<signal name="node_selected">
 			<param index="0" name="node" type="Node" />
 			<description>
-				Emitted when a GraphNode is selected.
+				Emitted when the [GraphElement] [param node] is selected.
```
YuriSizov, line 385: Same here.

doc/classes/GraphEdit.xml
```diff
@@ -364,28 +365,29 @@
...
 			</description>
 		</signal>
 		<signal name="paste_nodes_request">
 			<description>
-				Emitted when the user presses [kbd]Ctrl + V[/kbd].
+				Emitted when this [GraphEdit] requests [GraphElement] nodes to be pasted, for example by pressing [kbd]Ctrl + V[/kbd] (the [code]ui_paste[/code] action).
```
YuriSizov, line 390: Same as with previous such signals.

YuriSizov (changes requested):
Sorry for not reviewing sooner and thanks for looking into the remaining descriptions. I recommend doing some adjustments, but otherwise it's good.

Mickeon:
Updated the PR accounting for the above feedback

YuriSizov:
Amazing! One last thing, you've updated `end_node_move` to reference `GraphElement`, but you forgot to update `begin_node_move`. Once this is done, we can merge 👍

Mickeon:
Should be all good soon

YuriSizov (approved):

AThousandShips (approved):

YuriSizov:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

PR #87048, opened 2024-01-10
Use SceneStringNames in Skeleton3DEditorPlugin

Mickeon:
I noticed while making https://github.com/godotengine/godot/pull/87047 that SceneStringNames definitions of these signal names exist, but are only used in **Skeleton3D**. This PR also uses them for the equivalent editor plugin.
They're... longer, more verbose definitions, as SceneStringNames do not have a shorthand, but it does make sense?


AThousandShips (approved):
LGTM, makes sense to use these when we can, don't think this hurts readability

akien-mga:
Thanks!

PR #87050, opened 2024-01-10
Skeleton3D: Convert 3.x skeletons

nikitalita:
This implements conversion of the old 3.x Skeleton format to 4.x upon load (similar to old Mesh format conversion).

Skeletons had a `Transform pose` property in 3.x, which was broken out in 4.x to 
- `Vector3 pose_position;`
- `Quaternion pose_rotation;` 
- `Vector3 pose_scale`

Additionally, in 3.x, the pose of the skeleton was relative to the `rest` position; in 4.x, the poses are absolute.

 So, we recompute the poses to be absolute and break the `Transform3D` into its various components.

### DIFF
--- a/scene/3d/skeleton_3d.cpp
+++ b/scene/3d/skeleton_3d.cpp
@@ -95,6 +95,19 @@ bool Skeleton3D::_set(const StringName &p_path, const Variant &p_value) {
 		set_bone_pose_rotation(which, p_value);
 	} else if (what == "scale") {
 		set_bone_pose_scale(which, p_value);
+#ifndef DISABLE_DEPRECATED
+	} else if (what == "pose") {
+		// Kept for compatibility from 3.x to 4.x.
+		WARN_DEPRECATED_MSG("Skeleton uses old pose format, which is deprecated (and loads slower). Consider re-importing or re-saving the scene." +
+				(is_inside_tree() ? vformat(" Path: \"%s\"", get_path()) : String()));
+		// Old Skeleton poses were relative to rest, new ones are absolute, so we need to recompute the pose.
+		// Skeleton3D nodes were always written with rest before pose, so this *SHOULD* work...
+		Transform3D rest = get_bone_rest(which);
+		Transform3D pose = rest * (Transform3D)p_value;
+		set_bone_pose_position(which, pose.origin);
+		set_bone_pose_rotation(which, pose.basis.get_rotation_quaternion());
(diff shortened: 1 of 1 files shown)


fire (approved):
I think this is good. @TokageItLab @lyuma Can you check?

This was how it was described to me by Tokage YEARS ago.

TokageItLab (approved):
This should be fine for the bone properties.

However, there is no animation compatibility, so if you want full compatibility with 3.x Skeleton, you will need to consider that as well; There is never full compatibility possible with animations. Bone animation is now absolute, not relative, and scale has been isolated so that transforms should be different when combined with rotation and scale.

scene/3d/skeleton_3d.cpp
```diff
@@ -95,6 +95,20 @@ bool Skeleton3D::_set(const StringName &p_path, const Variant &p_value) {
 		set_bone_pose_rotation(which, p_value);
 	} else if (what == "scale") {
 		set_bone_pose_scale(which, p_value);
+#ifndef DISABLE_DEPRECATED
+	} else if (what == "pose") {
+		// Kept for compatibility from 3.x to 4.x.
+		WARN_DEPRECATED_MSG(vformat(
+				"Skeleton uses old pose format, which is deprecated (and loads slower). Consider re-importing or re-saving the scene. Path: \"%s\"",
```
AThousandShips, line 102: I'd suggest making the path part optional, otherwise it just prints an empty and confusing path, something like:
```cpp
		WARN_DEPRECATED_MSG(
				"Skeleton uses old pose format, which is deprecated (and loads slower). Consider re-importing or re-saving the scene." + 
				is_inside_tree() ? vformat(Path: \"%s\"", get_path()) : String());
```

nikitalita:
> This should be fine for the bone properties.
> 
> However, there is no animation compatibility, so if you want full compatibility with 3.x Skeleton, you will need to consider that as well; There is never full compatibility possible with animations. Bone animation is now absolute, not relative, and scale has been isolated so that transforms should be different when combined with rotation and scale.

Yes, this is fixed in my next PR. https://github.com/godotengine/godot/pull/87106

scene/3d/skeleton_3d.cpp
akien-mga: Makes sense, could you amend the commit with this tiny edit @nikitalita?

nikitalita: Done.

YuriSizov:
Thanks!

PR #87057, opened 2024-01-10
Default to 32-bit depth map on Forward+ renderer and 24-bit only on Mobile renderer

clayjohn:
Discussed here: https://github.com/godotengine/godot-proposals/issues/7366

Most desktop devices support `DATA_FORMAT_D32_SFLOAT_S8_UINT` (Windows: 99.88%; Linux: 98.66%; Mac: 100%) and there is very little or no performance hit for using it (for AMD it might even be faster since D24 is emulated). 

For mobile devices we want to use `DATA_FORMAT_D24_UNORM_S8_UINT` wherever possible as the practical limit is 32 bits before performance drops off. Its also much better supported on Android (98.76%). iOS only supports `DATA_FORMAT_D32_SFLOAT_S8_UINT` so the format check will fail and always fall back to `DATA_FORMAT_D32_SFLOAT_S8_UINT`.

An extra note. In most cases the current check was sufficient as a significant number of desktop devices don't support `DATA_FORMAT_D24_UNORM_S8_UINT` while a significant number of android devices don't support `DATA_FORMAT_D32_SFLOAT_S8_UINT`. But on a lot of desktop devices we were unnecessarily using `DATA_FORMAT_D24_UNORM_S8_UINT`.

### DIFF
--- a/servers/rendering/renderer_rd/storage_rd/render_scene_buffers_rd.cpp
+++ b/servers/rendering/renderer_rd/storage_rd/render_scene_buffers_rd.cpp
@@ -151,6 +151,14 @@ void RenderSceneBuffersRD::configure(const RenderSceneBuffersConfiguration *p_co
 	// cleanout any old buffers we had.
 	cleanup();
 
+	// At least one of these is required to be supported.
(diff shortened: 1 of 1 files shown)


Calinou (approved):

akien-mga:
Thanks!

PR #46657, opened 2021-03-04
Revert backport of 2D transform and camera snapping options

lawnjelly:
As discussed in https://github.com/godotengine/godot/pull/46615#issuecomment-790647994 I think there are a number of problem with `transform_snapping` currently and we should leave this until 3.2.5.

We can either mark the existing functions as experimental or revert, personally I'm in favour of reverting for 3.2.4, because I'm not keen on having broken functionality in stable release. Some might argue that the existing functions were helpful but we have to weigh that up against user confusion and expectations.

PR #87058, opened 2024-01-10
Remove 2D transform snapping to rely instead on the GPU

### LINKED ISSUE #56793: Vulkan: 2D Pixel Snap not working
### Godot version

4.0dev (ce6de71)

### System information

Windows 10, NVIDIA GeForce GTX 960

### Issue description

I've seen other issues like this but they're all closed as fixed. I'm still having this issue. Pixels are not snapping correctly with either or both pixel snap options.
1 pixel gap in between collisions.

### Steps to reproduce

Create a project with a floor and characterbody2d with basic movement functions, gravity, etc.
Set display stretch mode to `viewport` and `keep`.
Enable `Snap 2d Transforms to Pixel` or `Snap 2d Vertices to Pixel` or both.

image

[PixelSnapBug.zip](https://github.com/godotengine/godot/files/7872520/PixelSnapBug.zip)

### LINKED ISSUE #84632: Sprite "crunch" when placed in certain positions on screen

adamscott:
More or less applies #46657 to the `master` branch (4.x). It fixed issues on 3.x, but wasn't forward-ported in 4.x.

This PR removes the need of rounding brought by #84380. As the GPU is doing the rounding, we don't have to floor the transform values on the DisplayServer side.

### Merlin: Scale of the Magic
Using the open-source pixel perfect game [Merlin: Scale of the Magic](https://github.com/nicolasbize/Scalazard) by @nicolasbize.

#### w/o this PR
https://github.com/godotengine/godot/assets/270928/426c4009-8b45-476e-9a47-d5080baa9aaa

#### w/ this PR
https://github.com/godotengine/godot/assets/270928/7b62d896-ff5f-4a72-ade2-eb84135d5d90

#### w/o this PR
image

#### w/ this PR
image

## Relates to
Supersedes #84380.
Fixes #56793.
Fixes #84632.

### DIFF
--- a/servers/rendering/renderer_canvas_cull.cpp
+++ b/servers/rendering/renderer_canvas_cull.cpp
@@ -248,9 +248,6 @@ void RendererCanvasCull::_cull_canvas_item(Item *p_canvas_item, const Transform2
 	}
 
 	Transform2D xform = ci->xform;
-	if (snapping_2d_transforms_to_pixel) {
-		xform.columns[2] = xform.columns[2].floor();
-	}
 	xform = p_transform * xform;
 
 	Rect2 global_rect = xform.xform(rect);
--- a/servers/rendering/renderer_viewport.cpp
+++ b/servers/rendering/renderer_viewport.cpp
@@ -43,19 +43,11 @@ static Transform2D _canvas_get_transform(RendererViewport::Viewport *p_viewport,
 	float scale = 1.0;
 	if (p_viewport->canvas_map.has(p_canvas->parent)) {
 		Transform2D c_xform = p_viewport->canvas_map[p_canvas->parent].transform;
-		if (p_viewport->snap_2d_transforms_to_pixel) {
-			c_xform.columns[2] = c_xform.columns[2].floor();
-		}
 		xf = xf * c_xform;
 		scale = p_canvas->parent_scale;
 	}
 
 	Transform2D c_xform = p_canvas_data->transform;
-
-	if (p_viewport->snap_2d_transforms_to_pixel) {
-		c_xform.columns[2] = c_xform.columns[2].floor();
-	}
-
 	xf = xf * c_xform;
 
 	if (scale != 1.0 && !RSG::canvas->disable_scale) {


markdibarry:
Seeing a lot of issues using the pixel perfect test project provided in #84380.

https://github.com/godotengine/godot/assets/21325943/b0cd947d-5ec3-460b-ac33-3fae34efc954

adamscott:
@markdibarry @KeyboardDanni I'll try to port [PixelPerfectTest.zip](https://github.com/godotengine/godot/files/13244300/PixelPerfectTest.zip) to Godot 3.x, if there's issues there too.

KeyboardDanni:
I looked at your commits and it looks like it's just removing the snapping that's applied to all 2D transforms (leaving in the snapping that's done on Sprites). Feels like something is missing here. I would expect changes in the shader code as well if we are to rely on GPU snapping. My understanding is that the GPU snap was only done on the 3.x branch.

clayjohn:
> I looked at your commits and it looks like it's just removing the snapping that's applied to all 2D transforms (leaving in the snapping that's done on Sprites). Feels like something is missing here. I would expect changes in the shader code as well if we are to rely on GPU snapping. My understanding is that the GPU snap was only done on the 3.x branch.

GPU snap is here and its implemented the same as it was in 3.x
https://github.com/godotengine/godot/blob/9b522ac1a85cab1a7a867b7a9f3bb102d9376ac2/servers/rendering/renderer_rd/shaders/canvas.glsl#L201-L206

The difference is that 3.x removed the transform snapping which is what this PR does.

KeyboardDanni:
That particular code doesn't seem to be doing much snapping, unless something got configured wrong. The third video looks functionally identical to disabling snapping (with the exception of Sprites which are still being snapped). Which is why it seems like something is missing here.

adamscott:
@clayjohn @KeyboardDanni We were doing this PR thinking that Godot 3.x was doing fine. But I ported your [PixelPerfectTest.zip](https://github.com/godotengine/godot/files/13244300/PixelPerfectTest.zip) test to Godot 3.x ([PixelPerfectTest3.zip](https://github.com/godotengine/godot/files/13896525/PixelPerfectTest3.zip)). And it seems that Godot 3.x code is failing your test too.

https://github.com/godotengine/godot/assets/270928/31eb3be9-8ca0-4c1c-8c22-39012f2de106

KeyboardDanni:
Looking at this, I'm wondering if the *camera* isn't getting snapped to pixel. It looks like the vertices of individual 2D elements are snapped. But you're still going to get jitter if the camera isn't being snapped as well. Technically you could work around this on the script side, but I feel like it'd be better to just have this happen in the shader as part of the vertex snap, so that everything would "just work".

I also feel that the way the snap options are presented in settings is a bit confusing. Users might think that you'd need to enable both snap settings (transform and vertex), when really you should only be using one, as having both enabled might cause unforeseen issues.

adamscott:
> Looking at this, I'm wondering if the _camera_ isn't getting snapped to pixel. It looks like the vertices of individual 2D elements are snapped. But you're still going to get jitter if the camera isn't being snapped as well. Technically you could work around this on the script side, but I feel like it'd be better to just have this happen in the shader as part of the vertex snap, so that everything would "just work".

Oh my, I didn't enable the 2D snapping in my Godot 3.x test. I'll capture stuff and come back with a video.

KeyboardDanni:
I'd also like to add that if we're going to remove 2D transform snapping, it should be done thoroughly. Remove the project setting, transform snap variables, etc. Replacing the transform snapping mode with a no-op would only make things more confusing for users.

KeyboardDanni:
cc @lawnjelly since they've worked on this stuff for 3.x and might have some insight.

adamscott:
Here's Godot 3.5.2 running PixelPerfectTest3 using GPU snap (failing the test too).

https://github.com/godotengine/godot/assets/270928/76ddd638-518a-40d3-9529-dfc61ffbae0e

adamscott:
@KeyboardDanni @markdibarry I've sent you private chat messages on https://chat.godotengine.org/. I'm currently organizing a meeting next week to talk about pixel perfect rendering issues. I think you'd be welcome to join and share your insight.

akien-mga:
Superseded by #87297.

Issue #82698, opened 2023-10-03
Soundtrack and animation not in sync when using autostart (for heavy scenes)


teadrinker:
### Godot version

v4.1.1.stable.mono.official [bd6af8e0e]

### System information

Godot v4.1.1.stable.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2070 Super (NVIDIA; 31.0.15.2754) - Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz (12 Threads)

### Issue description

when using the autostart feature with a heavy scene, the audio is out of sync with the animation.
when calling play from a script, sync works well in the same scene.

### Steps to reproduce

open project in synctest.zip and run project.
Note:
 * The drums at 0s and at 16s are out of sync (using autostart)
 * The drums at 32s and at 52s, sync works as expected (because script plays animation at 20s)

### Minimal reproduction project

[SyncTest.zip](https://github.com/godotengine/godot/files/12787331/SyncTest.zip)

MJacred:
This is not really a bug, but a limitation of the AnimationPlayer's autoplay on load: https://docs.godotengine.org/en/stable/classes/class_animationplayer.html#class-animationplayer-property-autoplay.

It's triggered when the scene is loaded. This does not mean the scene is rendered. See also my comment here: https://github.com/godotengine/godot/issues/81549#issuecomment-1871398751

In your case, this will do the trick: first, deactivate the `autoplay on load` for your track "new_animation". And in a script attached e.g. to your scene root node, have this code:
```gdscript
func _ready():
	RenderingServer.connect("frame_post_draw", play_music)

func play_music():
	RenderingServer.disconnect("frame_post_draw", play_music)
	$AnimationPlayer.play("new_animation")
```

PR #87061, opened 2024-01-10
Add AudioStreamPlayerInternal to unify stream players

KoBeWi:
We have 3 AudioStreamPlayer classes and they share a lot of code. Code duplication, especially this big, is bad and that doesn't need explaining. Every fix or common feature has to be implemented 3 times and there is risk of code differences that result in bugs.

This PR tries to unify the code under a single class - AudioStreamPlayerInternal. It's an internal class used by the 3 audio players (not by inheritance). I delegated all common code and fields to the new class. The best indicator how big was the code duplication is that this PR has more deleted lines despite I added a whole new class. I also cleaned up some includes.

I tried to avoid introducing changes, but there are small differences from the original code, mostly in the order of some lines (some things were ordered differently in different players and common methods will obviously unify that). Probably needs testing to check whether all features are still working and there is no change in behavior. I did some basic tests and the audio still plays at least.

In the meantime I also came up with an alternate approach of taking advantage of C++ multiple inheritance. Maybe I'll try that too, as it should save some more code and possibly make it more readable.

### DIFF
--- a/scene/2d/audio_stream_player_2d.cpp
+++ b/scene/2d/audio_stream_player_2d.cpp
@@ -33,86 +33,48 @@
 #include "core/config/project_settings.h"
 #include "scene/2d/area_2d.h"
 #include "scene/2d/audio_listener_2d.h"
-#include "scene/main/window.h"
+#include "scene/audio/audio_stream_player_internal.h"
+#include "scene/main/viewport.h"
 #include "scene/resources/world_2d.h"
-
-#define PARAM_PREFIX "parameters/"
+#include "scene/scene_string_names.h"
+#include "servers/audio/audio_stream.h"
+#include "servers/audio_server.h"
 
 void AudioStreamPlayer2D::_notification(int p_what) {
+	internal->notification(p_what);
+
 	switch (p_what) {
 		case NOTIFICATION_ENTER_TREE: {
 			AudioServer::get_singleton()->add_listener_changed_callback(_listener_changed_cb, this);
-			if (autoplay && !Engine::get_singleton()->is_editor_hint()) {
-				play();
-			}
-			set_stream_paused(!can_process());
 		} break;
 
 		case NOTIFICATION_EXIT_TREE: {
-			set_stream_paused(true);
 			AudioServer::get_singleton()->remove_listener_changed_callback(_listener_changed_cb, this);
-		} break;
-
-		case NOTIFICATION_PREDELETE: {
-			stop();
-		} break;
-
-		case NOTIFICATION_PAUSED: {
-			if (!can_process()) {
-				// Node can't process so we start fading out to silence.
-				set_stream_paused(true);
-			}
-		} break;
-
-		case NOTIFICATION_UNPAUSED: {
-			set_stream_paused(false);
 		} break;
 
 		case NOTIFICATION_INTERNAL_PHYSICS_PROCESS: {
 			// Update anything related to position first, if possible of course.
-			if (setplay.get() > 0 || (active.is_set() && last_mix_count != AudioServer::get_singleton()->get_mix_count()) || force_update_panning) {
+			if (setplay.get() > 0 || (internal->active.is_set() && last_mix_count != AudioServer::get_singleton()->get_mix_count()) || force_update_panning) {
 				force_update_panning = false;
 				_update_panning();
 			}
 
 			if (setplayback.is_valid() && setplay.get() >= 0) {
-				active.set();
-				AudioServer::get_singleton()->start_playback_stream(setplayback, _get_actual_bus(), volume_vector, setplay.get(), pitch_scale);
+				internal->active.set();
+				AudioServer::get_singleton()->start_playback_stream(setplayback, _get_actual_bus(), volume_vector, setplay.get(), internal->pitch_scale);
 				setplayback.unref();
 				setplay.set(-1);
 			}
 
-			if (!stream_playbacks.is_empty() && active.is_set()) {
-				// Stop playing if no longer active.
-				Vector<Ref<AudioStreamPlayback>> playbacks_to_remove;
-				for (Ref<AudioStreamPlayback> &playback : stream_playbacks) {
-					if (playback.is_valid() && !AudioServer::get_singleton()->is_playback_active(playback) && !AudioServer::get_singleton()->is_playback_paused(playback)) {
-						playbacks_to_remove.push_back(playback);
-					}
-				}
-				// Now go through and remove playbacks that have finished. Removing elements from a Vector in a range based for is asking for trouble.
-				for (Ref<AudioStreamPlayback> &playback : playbacks_to_remove) {
-					stream_playbacks.erase(playback);
-				}
-				if (!playbacks_to_remove.is_empty() && stream_playbacks.is_empty()) {
-					// This node is no longer actively playing audio.
-					active.clear();
-					set_physics_process_internal(false);
-				}
-				if (!playbacks_to_remove.is_empty()) {
-					emit_signal(SNAME("finished"));
-				}
+			if (!internal->stream_playbacks.is_empty() && internal->active.is_set()) {
+				internal->process();
 			}
-
-			while (stream_playbacks.size() > max_polyphony) {
-				AudioServer::get_singleton()->stop_playback_stream(stream_playbacks[0]);
-				stream_playbacks.remove_at(0);
-			}
+			internal->ensure_playback_limit();
 		} break;
 	}
 }
 
-// Interacts with PhysicsServer2D, so can only be called during _physics_process
+// Interacts with PhysicsServer2D, so can only be called during _physics_process.
 StringName AudioStreamPlayer2D::_get_actual_bus() {
 	Vector2 global_pos = get_global_position();
 
@@ -144,12 +106,12 @@ StringName AudioStreamPlayer2D::_get_actual_bus() {
 
 		return area2d->get_audio_bus_name();
 	}
-	return default_bus;
+	return internal->bus;
 }
 
 // Interacts with PhysicsServer2D, so can only be called during _physics_process
 void AudioStreamPlayer2D::_update_panning() {
-	if (!active.is_set() || stream.is_null()) {
+	if (!internal->active.is_set() || internal->stream.is_null()) {
 		return;
 	}
 
@@ -194,7 +156,7 @@ void AudioStreamPlayer2D::_update_panning() {
 		}
 
 		float multiplier = Math::pow(1.0f - dist / max_distance, attenuation);
-		multiplier *= Math::db_to_linear(volume_db); // Also apply player volume!
+		multiplier *= Math::db_to_linear(internal->volume_db); // Also apply player volume!
 
 		float pan = relative_to_listener.x / screen_size.x;
 		// Don't let the panning effect extend (too far) beyond the screen.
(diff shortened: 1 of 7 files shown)


scene/2d/audio_stream_player_2d.cpp
```diff
@@ -33,84 +33,46 @@
...
-				}
-			}
-
-			while (stream_playbacks.size() > max_polyphony) {
-				AudioServer::get_singleton()->stop_playback_stream(stream_playbacks[0]);
-				stream_playbacks.remove_at(0);
+			if (!internal.stream_playbacks.is_empty() && internal.active.is_set()) {
+				internal.process();
 			}
+			internal.ensure_playback_limit();
```
KoBeWi, line 70: The playback limit is applied each frame, unlike in AudioStreamPlayer, which does it only on `play()`. Maybe other players could do that too?

scene/audio/audio_stream_player.cpp
```diff
@@ -220,43 +126,19 @@ AudioStreamPlayer::MixTarget AudioStreamPlayer::get_mix_target() const {
...
 void AudioStreamPlayer::_set_playing(bool p_enable) {
-	if (p_enable) {
-		play();
-	} else {
-		stop();
-	}
+	internal.set_playing(p_enable);
 }
 
 bool AudioStreamPlayer::_is_active() const {
```
KoBeWi, line 132: This method code was the same as `is_playing()`. I unified it to do `active.is_set()` as other players, idk if it makes difference.

scene/3d/audio_stream_player_3d.cpp
```diff
@@ -595,83 +543,52 @@ void AudioStreamPlayer3D::seek(float p_seconds) {
...
-		return AudioServer::get_singleton()->get_playback_position(stream_playbacks[stream_playbacks.size() - 1]);
-	}
-	return 0;
+	return internal.get_playback_position();
 }
 
 void AudioStreamPlayer3D::set_bus(const StringName &p_bus) {
-	//if audio is active, must lock this
+	// If audio is active, must lock this.
 	AudioServer::get_singleton()->lock();
```
KoBeWi, line 565: The lock does not exist in other players. Should it be removed?

Mickeon:
> In the meantime I also came up with an alternate approach of taking advantage of C++ multiple inheritance. Maybe I'll try that too, as it should save some more code and possibly make it more readable.

I'm very interested. Readability seems to be the must. But, is something like the suggestion ever used across  the rest of the codebase?

Whatever path is taken to unite the AudioStreamPlayers, it would be nice _in the future_ to do the same for other classes with a lot of duplicated code.

ellenhp:
An important discussion point is whether this may interfere with people creating custom streams with gdextension. I think the answer is no because they can just inherit audiostream and continue duplicating all the code we're consolidating here. But that's a use case I want us to continue supporting so I want to bring it up :)

Mickeon:
Compatibility is retained by doing this. From the outsider's perspective nothing should change.

KoBeWi:
> But, is something like the suggestion ever used across the rest of the codebase?

No, AFAIK every class is single inheritance.

I tried that approach and seems like you can't bind methods from another class (in that case AudioStreamPlayerInternal would be the secondary parent class), which makes it useless. Maybe there is a way to untangle the inheritance, but it's not that easy, so meh. The solution in this PR is better than what we have anyway.

ellenhp:
> Compatibility is retained by doing this. From the outsider's perspective nothing should change.

Yeah makes sense, I'm just trying to be thorough. I also think we should come up with a better name because we have a couple audio streams that won't end up subclassing this new internal class because they're dynamic (microphone, generator, probably others).

Mickeon:
It could be called something like "Core". I had it for a bit and it's the only thing that comes to mind.

KoBeWi:
> I also think we should come up with a better name because we have a couple audio streams that won't end up subclassing this new internal class

The new class is for audio stream players (it's AudioStream**Player**Internal), not audio streams. I think the name is fine.

ellenhp:
Yeah sorry, neither of my comments make any sense for this. I got a covid vaccine about 36 hours ago and I'm still in the living dead phase of recovery. Didn't help that I hadn't looked at the code yet. I do want to refactor streams at some point too but there's a lot less we can do there without breaking stuff. As you noted, you can't create a new player in GDExtension yet so we really don't have to worry about breaking compat, and I'm not even aware of any custom modules that implement their own players without going through an entirely different middleware.

scene/3d/audio_stream_player_3d.cpp
ellenhp: Good catch. It should be unnecessary. We handle thread contention between mixing and modification of playback params like this in AudioServer with SafeList.

scene/audio/audio_stream_player.cpp
MJacred: Hm… for AudioStreamPlayer, the meaning changes from "AudioServer has my playback registered and setup, and started playing" to "I started being processed and the AudioServer is about to start setting up my playback".

But no idea if this has any tangible repercussions. `_is_active` is a Node thing afaik

scene/audio/audio_stream_player_internal.cpp
```diff
@@ -0,0 +1,253 @@
...
+	_set_process(false);
+}
+
+void AudioStreamPlayerInternal::set_node(Node *p_node, const Callable &p_play_callable, bool p_physical) {
+	node = p_node;
+	play_callable = p_play_callable;
+	physical = p_physical;
+
+	AudioServer::get_singleton()->connect("bus_layout_changed", callable_mp((Object *)node, &Object::notify_property_list_changed));
+	AudioServer::get_singleton()->connect("bus_renamed", callable_mp((Object *)this, &Object::notify_property_list_changed).unbind(3));
```
MJacred, line 185: I'm not sure about this, but shouldn't the second connect also be to `node`?

```suggestion
	AudioServer::get_singleton()->connect("bus_renamed", callable_mp((Object *)node, &Object::notify_property_list_changed).unbind(3));
```

scene/2d/audio_stream_player_2d.h
```diff
@@ -64,12 +63,6 @@ class AudioStreamPlayer2D : public Node2D {
...
 	bool force_update_panning = false;
 
-	float volume_db = 0.0;
-	float pitch_scale = 1.0;
-	bool autoplay = false;
-	StringName default_bus = SceneStringNames::get_singleton()->Master;
-	int max_polyphony = 1;
-
 	void _set_playing(bool p_enable);
 	bool _is_active() const;
```
MJacred, line 68: You forgot to remove a few lines below:

```c++
	void _on_bus_layout_changed();
	void _on_bus_renamed(int p_bus_index, const StringName &p_old_name, const StringName &p_new_name);
```

scene/3d/audio_stream_player_3d.cpp
```diff
@@ -595,83 +543,49 @@ void AudioStreamPlayer3D::seek(float p_seconds) {
...
-	return 0;
+	return internal.get_playback_position();
 }
 
 void AudioStreamPlayer3D::set_bus(const StringName &p_bus) {
-	//if audio is active, must lock this
-	AudioServer::get_singleton()->lock();
-	bus = p_bus;
-	AudioServer::get_singleton()->unlock();
+	internal.bus = p_bus;
```
MJacred, line 564:
```suggestion
	internal.bus = p_bus; // This will be pushed to the audio server during the next physics timestep, which is fast enough.
```

And could you add this info to the AudioStreamPlayer2D/3D docs, please?

MJacred (changes requested):
The rest should be fine, I'd say. Some processing order was changed, e.g. in `_notification`, but should make no difference (I think).

scene/audio/audio_stream_player.cpp
MJacred: Not sure if relevant, but does animation use this for syncing the start? Audio already starts earlier than other things. For reference: https://github.com/godotengine/godot/issues/82698#issuecomment-1871426859

scene/3d/audio_stream_player_3d.cpp
Mickeon: Somewhat out of scope of the PR since it's always been the case, pretty much

MJacred: The comment was copied from same func from AudioStreamPlayer2D.

The docs update is just for info for users to not be confused about possible delays if their physics process is stuttering

scene/3d/audio_stream_player_3d.cpp
```diff
@@ -242,32 +243,12 @@ void AudioStreamPlayer3D::_notification(int p_what) {
...
 		} break;
 
 		case NOTIFICATION_EXIT_TREE: {
-			set_stream_paused(true);
 			AudioServer::get_singleton()->remove_listener_changed_callback(_listener_changed_cb, this);
 		} break;
 
-		case NOTIFICATION_PREDELETE: {
-			stop();
-		} break;
```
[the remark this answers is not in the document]
KoBeWi: This is handled by the internal class, which also receives this notification.

scene/2d/audio_stream_player_2d.cpp
```diff
@@ -33,84 +33,47 @@
...
 		} break;
 
 		case NOTIFICATION_EXIT_TREE: {
-			set_stream_paused(true);
 			AudioServer::get_singleton()->remove_listener_changed_callback(_listener_changed_cb, this);
 		} break;
 
-		case NOTIFICATION_PREDELETE: {
-			stop();
-		} break;
```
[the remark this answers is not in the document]
KoBeWi: See below.

scene/3d/audio_stream_player_3d.cpp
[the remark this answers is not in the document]
MJacred: How?
The relevant internal class (see code below) cannot reset `setplay`. That variable only exists in AudioStreamPlayer2D and *3D.

```
void AudioStreamPlayerInternal::notification(int p_what) {
...
            case Node::NOTIFICATION_PREDELETE: {
                       for (Ref<AudioStreamPlayback> &playback : stream_playbacks) {
                            AudioServer::get_singleton()->stop_playback_stream(playback);
                        }
                        stream_playbacks.clear();
            } break;
...
}
```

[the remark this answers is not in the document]
KoBeWi: Ok fixed (not pushed yet).

scene/audio/audio_stream_player_internal.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+		case Node::NOTIFICATION_INTERNAL_PROCESS: {
+			process();
+		} break;
+
+		case Node::NOTIFICATION_PREDELETE: {
+			for (Ref<AudioStreamPlayback> &playback : stream_playbacks) {
+				AudioServer::get_singleton()->stop_playback_stream(playback);
+			}
+			stream_playbacks.clear();
```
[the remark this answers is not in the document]
KoBeWi: This is code from original AudioStreamPlayer, which did only this. I'd call `stop()` if I wanted to do more, but unified it into this part. The flag is irrelevant at that point.

tbh this made me realize that `setsetplay` is irrelevant too.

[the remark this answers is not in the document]
MJacred: > This is code from original AudioStreamPlayer, which did only this

True. But AudioStreamPlayer2D & *3D cleared `active`, because they called `stop()`.

Whether the flag is relevant or not. No idea. At least, afaik, the audio system does not use that.

Similar with `setplay`.

I don't know how `NOTIFICATION_PREDELETE`, `AudioStreamPlayerInternal::_set_process` and `AudioStreamPlayerInternal::process` exactly interact with each other. The order of execution. And what happens if the 2d/3d audio player node is called with `play()` and then immediately deleted. Or similar edge cases.

The feedback regarding `active` and `setplay` is about preserving the original behaviour - and make it consistent among the audio players.

No idea if they were a "just in case" measure or if they actively prevented edge cases. So, if you know they don't matter, then it's completely fine with me to not reset these variables.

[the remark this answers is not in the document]
KoBeWi: `NOTIFICATION_PREDELETE` is received just before the object is freed. The object does not call any method anymore (except destructor), even deferred calls.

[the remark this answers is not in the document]
MJacred: Ah, then as long as `NOTIFICATION_INTERNAL_PHYSICS_PROCESS` is not executed afterwards, then it sounds safe. Thanks!

scene/audio/audio_stream_player_internal.cpp
KoBeWi: Yes, I already fixed it locally.

scene/3d/audio_stream_player_3d.cpp
KoBeWi: I'd wait until a confused user opens an issue. So far it hasn't been a problem to anyone, likely because changing buses dynamically is a very uncommon action.
I copied the comment though.

KoBeWi:
Rebased and updated. I had to change the AudioStreamPlayerInternal to Object, to handle the newly added `parameter_list_changed` signal. I also fixed the `null` issue from #86473.

reduz (approved):
Looks good, I was hopeful it would save more code, but I guess its still a lot less error prone and easier to maintain. Great work!

akien-mga:
Failing test:
```
./tests/core/object/test_class_db.h:833:
TEST SUITE: [ClassDB]
TEST CASE:  [ClassDB] Add exposed classes, builtin types, and global enums
  [ClassDB] Validate exposed classes

./tests/core/object/test_class_db.h:376: ERROR: CHECK_FALSE( (p_arg.name.is_empty() || p_arg.name.begins_with("_unnamed_arg")) ) is NOT correct!
  values: CHECK_FALSE( true )
  logged: Unnamed argument in position 0 of method 'AudioStreamPlayer2D.set_pitch_scale'.
```

scene/audio/audio_stream_player_internal.h
```diff
@@ -0,0 +1,105 @@
...
+
+#include "core/object/ref_counted.h"
+#include "core/templates/safe_refcount.h"
+
+class AudioStream;
+class AudioStreamPlayback;
+class Node;
+
+class AudioStreamPlayerInternal : public Object {
+	GDCLASS(AudioStreamPlayerInternal, Object);
```
YuriSizov, line 42: Do we actually need this to be a Godot Object? Can it be just a class?

KoBeWi: It needs to be Object to connect the signal.

YuriSizov: Ah, now I see it! Okay then.

YuriSizov (approved):
Gave it a quick look, and it looks good!

akien-mga:
Thanks!

PR #87063, opened 2024-01-10
Misc changes to the GLTF module before audio PR

aaronfranke:
This PR includes a few changes I made while working on my upcoming GLTF audio PR, but without any audio stuff.

* Fully document GLTFBufferView.
* Partially document GLTFAccessor.
* Improve some documentation in GLTFDocumentExtension.
* Add a method to GLTFState to append a PackedByteArray to the buffers / bufferViews, with optional deduplication.
* Add a method to GLTFBufferView to read the data from this bufferView as a PackedByteArray.
* Allow the node JSON to be empty when calling `import_node` on extensions. This is required to handle the case of Godot scene -> in-memory GLTF data -> Godot scene, so like a half-export then half-import.
* Allow the scene node to be `null` when calling `export_node` on extensions. This is required to handle the case of GLTF file -> in-memory GLTF data -> GLTF file, so like a half-import then half-export. This is an actual use case I have that needs to be supported, and anyway, this is good for code cleanliness to ensure the steps are fully separated.

### DIFF
--- a/modules/gltf/doc_classes/GLTFBufferView.xml
+++ b/modules/gltf/doc_classes/GLTFBufferView.xml
@@ -1,22 +1,40 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <class name="GLTFBufferView" inherits="Resource" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../../doc/class.xsd">
 	<brief_description>
+		Represents a GLTF buffer view.
 	</brief_description>
 	<description>
+		GLTFBufferView is a data structure representing GLTF a [code]bufferView[/code] that would be found in the [code]"bufferViews"[/code] array. A buffer is a blob of binary data. A buffer view is a slice of a buffer that can be used to identify and extract data from the buffer.
+		Most custom uses of buffers only need to use the [member buffer], [member byte_length], and [member byte_offset]. The [member byte_stride] and [member indices] properties are for more advanced use cases such as interleaved mesh data encoded for the GPU.
 	</description>
 	<tutorials>
+		<link title="Buffers, BufferViews, and Accessors in Khronos glTF specification">https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_005_BuffersBufferViewsAccessors.md</link>
 		<link title="Runtime file loading and saving">$DOCS_URL/tutorials/io/runtime_file_loading_and_saving.html</link>
 	</tutorials>
+	<methods>
+		<method name="load_buffer_view_data" qualifiers="const">
+			<return type="PackedByteArray" />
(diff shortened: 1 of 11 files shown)


fire:
Can you write an example how to use handle the case of GLTF file -> in-memory GLTF data -> GLTF file, so like a half-import then half-export.

I think I know what you mean, but what do you mean? It would help the general godot developer if we were able to put a sample here.

Otherwise I didn't see any obvious problems, but need a testplan -- like project examples of the use cases you mentioned.

aaronfranke:
@fire Here is an example use case from The Mirror, packing a GLTF into a GLB:

```gdscript
## Converts a GLTF document (including all its external dependencies) to a GLB byte array.
## GLTF files can have dependencies on other files. This would make distribution very complex.
## In order to resolve this, we can convert the GLTF to GLB data at runtime to pack it into one file.
static func convert_gltf_files_to_glb_data(path: String) -> PackedByteArray:
	var state: GLTFState = GLTFState.new()
	var doc: GLTFDocument = GLTFDocument.new()
	var err = doc.append_from_file(path, state)
	if err:
		push_error(str(err))
		return PackedByteArray()
	return doc.generate_buffer(state)
```

modules/gltf/doc_classes/GLTFBufferView.xml
```diff
@@ -1,22 +1,40 @@
...
 		</member>
 		<member name="byte_offset" type="int" setter="set_byte_offset" getter="get_byte_offset" default="0">
+			The offset, in bytes, from the start of the buffer to the start of this buffer view.
 		</member>
 		<member name="byte_stride" type="int" setter="set_byte_stride" getter="get_byte_stride" default="-1">
+			The stride, in bytes, between interleaved data. If [code]-1[/code], this buffer view is not interleaved.
 		</member>
 		<member name="indices" type="bool" setter="set_indices" getter="get_indices" default="false">
+			True if the GLTFBufferView's OpenGL GPU buffer type is an [code]ELEMENT_ARRAY_BUFFER[/code] used for vertex indices (integer constant [code]34963[/code]). False if the buffer type is [code]ARRAY_BUFFER[/code] used for vertex attributes (integer constant [code]34962[/code]) or when any other value. See [url=https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_005_BuffersBufferViewsAccessors.md]Buffers, BufferViews, and Accessors[/url] for possible values. This property is set but never used, setting this property will do nothing.
```
fire, line 37: I don't think this is opengl specific.

aaronfranke: GLTF is derived from OpenGL. These constants were originally defined in OpenGL even though they are now a part of GLTF.

fire (approved):

akien-mga:
Thanks!

PR #87064, opened 2024-01-10
Add some more comments to AudioServer and remove playback fade-in

### LINKED ISSUE #66452: Godot 4 - Audio clips have their beginning cut off
### Godot version

v4.0.beta1.official [20d667284]

### System information

Windows 10, Vulkan, NVidia RTX 2070 Super

### Issue description

In Godot 4, some audio clips sound incorrect in engine for me; it sounds like the start is being cut off. I've tried both .ogg and .wav files and it happens with both, I've tried tweaking the import settings on the sound to no effect, and I've tried an identical setup in Godot 3.4.2 and the issue does not occur.

Here's a video comparing the original sound file, the sound in Godot 4, and the sound in Godot 3. You can hear that in Godot 4 the sound file sounds less "sharp" at the beginning as it does in the original file and Godot 3. Both these projects were created with no changes made to any settings.

https://user-images.githubusercontent.com/3895119/192319165-d8c13f8f-4ce7-4a0c-a10f-9e494e10ac65.mp4

### Steps to reproduce

1. Create a new project in Godot 4
2. Import a short sound like the one I used (It's most noticeable on very short 8-bit style sounds)
3. Play through the preview or through an AudioStreamPlayer and compare with the original sound file

Godot 4.0 beta project: 
[Soundtest.zip](https://github.com/godotengine/godot/files/9647648/Soundtest.zip)

Godot 3.4.2 project:
[Soundtest3.zip](https://github.com/godotengine/godot/files/9647650/Soundtest3.zip)

ellenhp:
Should fix #66452 but I haven't tested yet. This also adds a ton of docs to hopefully make the AudioServer more approachable to newcomers.

There might be a better way to fix this bug though because I suspect this change will cause pops if you change the bus of a running player.

### DIFF
--- a/servers/audio_server.cpp
+++ b/servers/audio_server.cpp
@@ -343,32 +343,38 @@ void AudioServer::_mix_step() {
 			bus->soloed = false;
 		}
 	}
+	// This is legacy code from 3.x that allows video players and other audio sources that do not implement AudioStreamPlayback to output audio.
 	for (CallbackItem *ci : mix_callback_list) {
 		ci->callback(ci->userdata);
 	}
 
+	// Main mixing loop for audio streams.
+	// The basic idea here is to copy the samples returned by the AudioStreamPlayback's mix function into the audio buffers, while always maintaining a lookahead buffer of size LOOKAHEAD_BUFFER_SIZE to allow fade-outs for sudden stoppages.
 	for (AudioStreamPlaybackListNode *playback : playback_list) {
 		// Paused streams are no-ops. Don't even mix audio from the stream playback.
 		if (playback->state.load() == AudioStreamPlaybackListNode::PAUSED) {
 			continue;
 		}
 
+		// If `fading_out` is true, we're in the process of fading out the stream playback.
+		// Currently this sets the volume of the stream to 0 which creates a linear interpolation between its previous volume and silence. A more punchy option for fading out could be to just use the lookahead buffer.
 		bool fading_out = playback->state.load() == AudioStreamPlaybackListNode::FADE_OUT_TO_DELETION || playback->state.load() == AudioStreamPlaybackListNode::FADE_OUT_TO_PAUSE;
 
 		AudioFrame *buf = mix_buffer.ptrw();
 
-		// Copy the lookeahead buffer into the mix buffer.
+		// Copy the old contents of the lookeahead buffer into the beginning of the mix buffer.
 		for (int i = 0; i < LOOKAHEAD_BUFFER_SIZE; i++) {
 			buf[i] = playback->lookahead[i];
 		}
 
-		// Mix the audio stream
+		// Mix the audio stream.
 		unsigned int mixed_frames = playback->stream_playback->mix(&buf[LOOKAHEAD_BUFFER_SIZE], playback->pitch_scale.get(), buffer_size);
 
 		if (tag_used_audio_streams && playback->stream_playback->is_playing()) {
 			playback->stream_playback->tag_used_streams();
 		}
 
+		// Check to see if the stream has run out of samples.
 		if (mixed_frames != buffer_size) {
 			// We know we have at least the size of our lookahead buffer for fade-out purposes.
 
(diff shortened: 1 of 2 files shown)


servers/audio_server.cpp
```diff
@@ -343,32 +343,38 @@ void AudioServer::_mix_step() {
...
 		}
 
+		// If `fading_out` is true, we're in the process of fading out the stream playback.
+		// Currently this sets the volume of the stream to 0 which creates a linear interpolation between its previous volume and silence. A more punchy option for fading out could be to just use the lookahead buffer.
 		bool fading_out = playback->state.load() == AudioStreamPlaybackListNode::FADE_OUT_TO_DELETION || playback->state.load() == AudioStreamPlaybackListNode::FADE_OUT_TO_PAUSE;
 
 		AudioFrame *buf = mix_buffer.ptrw();
 
-		// Copy the lookeahead buffer into the mix buffer.
+		// Copy the old contents of the lookeahead buffer into the beginning of the mix buffer.
```
RedMser, line 365:
```suggestion
		// Copy the old contents of the lookahead buffer into the beginning of the mix buffer.
```

servers/audio_server.cpp
```diff
@@ -1361,7 +1375,7 @@ void AudioServer::init_channels_and_buffers() {
 void AudioServer::init() {
 	channel_disable_threshold_db = GLOBAL_DEF_RST("audio/buses/channel_disable_threshold_db", -60.0);
 	channel_disable_frames = float(GLOBAL_DEF_RST(PropertyInfo(Variant::FLOAT, "audio/buses/channel_disable_time", PROPERTY_HINT_RANGE, "0,5,0.01,or_greater"), 2.0)) * get_mix_rate();
-	buffer_size = 512; //hardcoded for now
+	buffer_size = 512; // Hardcoded for now. This would be really nice to have as a project setting because currently it limits audio latency to an absolute minimum of 11ms with default mix rate, but there's some additional work required to make that happen. See TODOs in `_mix_step_for_channel`.
```
reduz, line 1378: In reality most professional audio software I know uses 256 (~6ms at 44100) as the absolute minimum block size because this is considered the limit that you can tell apart if say, you are creating a synthesizer and will be playing notes received via Midi. Vastly most people can't really tell the difference between 12 and 6ms.

For game content, 12ms latency should be way more than enough.

That said, I think our problem here is different and its that we should make it customizable not as a block size but as a minimum latency size in ms. This way we can have a default of say, 10ms, and then the audio server converts this to audio frames based on the sampling rate and it finds the nearest power of 2 immediately above.

This should be ideal because it is very common that some devices will open audio in stuff like 192khz, for which 512 audio frames block is an absolute waste.

reduz (commented):
Looks great to me!

servers/audio_server.cpp
```diff
@@ -343,32 +343,38 @@ void AudioServer::_mix_step() {
...
+		// If `fading_out` is true, we're in the process of fading out the stream playback.
+		// Currently this sets the volume of the stream to 0 which creates a linear interpolation between its previous volume and silence. A more punchy option for fading out could be to just use the lookahead buffer.
 		bool fading_out = playback->state.load() == AudioStreamPlaybackListNode::FADE_OUT_TO_DELETION || playback->state.load() == AudioStreamPlaybackListNode::FADE_OUT_TO_PAUSE;
 
 		AudioFrame *buf = mix_buffer.ptrw();
 
-		// Copy the lookeahead buffer into the mix buffer.
+		// Copy the old contents of the lookeahead buffer into the beginning of the mix buffer.
 		for (int i = 0; i < LOOKAHEAD_BUFFER_SIZE; i++) {
 			buf[i] = playback->lookahead[i];
```
MJacred, line 367: proposal: add a `fadeout_`/`FADEOUT_` prefix to the members `lookahead` and `LOOKAHEAD_BUFFER_SIZE` to point out its purpose explicitly

ellenhp:
If someone wants to salvage this feel free! Since I've mostly taken a step back from the project I'd like to see this mixing loop documented so others can maintain it. I think it's better than where we were with 3.x from but right now it is very intimidating in its minimally-commented state. I just don't have bandwidth to get PRs merged lately. The bar for approval being high is great for the project but I don't have energy to meet it lately.

KeyboardDanni:
Salvaged as #98131.

PR #87075, opened 2024-01-11
[Shaders] Disallow trailing commas in function declarations

AThousandShips:
Might not be the most elegant fix, but not very familiar with this code, if someone has a cleaner idea for doing this I'd appreciate suggestions. Added it with deprecation to avoid breaking existing shaders, though I'm not sure that's necessary

I was considering enclosing the whole `while` loop in a block to check closing paren at the start, which would make catching this easier, but want to avoid noise in the blame if I can, having to indent the whole thing.

For reasoning, see the discussion in:
* https://github.com/godotengine/godot-proposals/issues/8840

Alternative to:
* https://github.com/godotengine/godot/pull/86991

### DIFF
--- a/servers/rendering/shader_language.cpp
+++ b/servers/rendering/shader_language.cpp
@@ -10881,7 +10881,13 @@ Error ShaderLanguage::_parse_shader(const HashMap<StringName, FunctionInfo> &p_f
 
 					if (tk.type == TK_COMMA) {
 						tk = _get_token();
-						//do none and go on
+#ifdef DISABLE_DEPRECATED
+						// Disallow trailing comma.
+						if (tk.type == TK_PARENTHESIS_CLOSE) {
+							_set_error(RTR("Expected a valid data type for argument. Trailing commas are not allowed."));
+							return ERR_PARSE_ERROR;
+						}
+#endif
 					} else if (tk.type != TK_PARENTHESIS_CLOSE) {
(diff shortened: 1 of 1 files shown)


servers/rendering/shader_language.cpp
```diff
@@ -9733,7 +9733,13 @@ Error ShaderLanguage::_parse_shader(const HashMap<StringName, FunctionInfo> &p_f
 
 					if (tk.type == TK_COMMA) {
 						tk = _get_token();
-						//do none and go on
+#ifdef DISABLE_DEPRECATED
+						// Disallow trailing comma.
+						if (tk.type == TK_PARENTHESIS_CLOSE) {
+							_set_error(RTR("Expected a valid data type for argument."));
```
AThousandShips, line 9739: Picked this error as it matches what happens if you do `void foo(,)`

AThousandShips:
Could add a warning on non-disabled cases (i.e. default builds) if desired

servers/rendering/shader_language.cpp
akx: Would a more user-friendly error be better, though? E.g. "(Remove the trailing comma?)"

AThousandShips: Could add 🙂 considering a warning as well so working on the wording, thank you!

PR #87078, opened 2024-01-11
Add auto fit timeline and bezier scale on animation editor

### LINKED ISSUE #7667: Add a button/shortcut to fit zoom to duration in the animation editor

Arcade style pixel game with lots of fast animations.

I use AnimationPlayer a lot for various things in the project, and I also create a lot of test AnimationPlayer that are used only for testing and later removed from the project, so I often have to do lots of zooming in / out to fit the used timeline on the AnimationPlayer panel.

The problem is that zooming in/out is a bit cumbersome because (due to the nature of what happens when the numbers grow) the zooming in is a bit slow in the beginning when I just start zooming in, and then the zooming factor just grows and is hard to control as it gets bigger. So I have to be unnecessarily careful for such a simple operation as zooming.

So it's hard to zoom in to the exact duration that I'm working with in a quick and easy way.

One solution would be to simply have a way to "zoom to fit" the used part of the timeline (the duration of the animation).

For example, middle mouse click on the zoom slider's head or something super simple like that:
image

Let's assume I create a new animation, which would be shown as let's say 4 seconds long, out of which the first 1 is used.
I'd type 0.5 as the duration of my animation.
At this point the duration marking on the timeline changes to 0.5, so obviously this is way too zoomed out for what I'll be working with.

Now I can just MMB click on the slider's head and the first 0.5 seconds stretch to cover the majority of the timeline length.

Even better (not sure how feasible that'd be to implement), pressing "Shift+Enter" (or something similar) instead of only "Enter" after typing in the duration of the animation could have the same effect, but the first solution is more universal (it can be used even if I wasn't in the process of entering new duration value).

N\A

It can't be an add-on in the asset library…

### LINKED ISSUE #85826: Make it more intuitive to zoom/scroll vertically in AnimationPlayer Bézier curve editor
### Tested versions

Found in 4.2.stable

### System information

Godot v4.2.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1060 with Max-Q Design (NVIDIA; 30.0.15.1272) - Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz (12 Threads)

### Issue description

I need to animate movement in pixels on a 1920x1080 canvas, but the curve editor won't let me scroll up past 350 or zoom the editor vertically.
image

In the minimal reproduction project, it's even worse:
image

### Steps to reproduce

1. create new project
2. add AnimationPlayer and a node to move
3. create a new animation, in which the node's position is animated with a Bézier track
4. add keyframes with both values that are near 0 and those that are far from 0
5. open the curve editor

[fourtest.zip](https://github.com/godotengine/godot/files/13578438/fourtest.zip)

Hilderin:
Add a button at the bottom of the animation editor that change the zoom based on the animation length and the bezier scale based on the values and handles of the displayed tracks. The icon and the tooltip of the button change depending if the bezier editor is displayed or not.

Some refactor was made in animation_track_editor.cpp to remove code duplication with the visibility check of the tracks.

This should help with the issue #85826

The objective was to help animation editor, above all the bezier editor which always starts with a strange scale. For small values, we always need to ctrl+alt+mouse wheel to see the cursves correctly. The auto fit helps a lot with that.

Auto fit button (outside bezier editor):
image

If you change the animation length, the zoom of the timeline can now be automatically updated based on the new length:
image

The size of the editor, the tracks names, etc... are taken into account exactly how it's done in the drawing of the timeline.

When i start Godot and open the Bezier editor for the first time, it looks like that:
image

The button is now updated with a different icon and tooltip for bezier:
image

After I pressed the new button:
image

It uses the min/max values of the keys and the handle positions. Only the visible tracks (not hidden and filtered) are used. That's why i refactored a little bit the code where the visibility of the tracks was made.

Tested with version 4.3.dev on Windows 11

- *Production edit: This closes https://github.com/godotengine/godot-proposals/issues/7667.*
- Closes #85826


editor/animation_track_editor.h
```diff
@@ -189,13 +191,14 @@ class AnimationTimelineEdit : public Range {
 	int get_name_limit() const;
 	int get_buttons_width() const;
 
-	float get_zoom_scale() const;
+	float get_zoom_scale() const;
```
AThousandShips, line 194:
```suggestion
	float get_zoom_scale() const;
```
Trailing whitespace

editor/animation_track_editor.cpp
```diff
@@ -6405,6 +6459,20 @@ bool AnimationTrackEditor::is_grouping_tracks() {
...
+}
+
+void AnimationTrackEditor::_auto_fit_bezier() {
+	timeline->auto_fit();
+
+	if (bezier_edit->is_visible()) {
+		bezier_edit->auto_fit_vertically();
+	}
+}
+
```
AThousandShips, line 6474:
```suggestion
```

editor/animation_track_editor.cpp
```diff
@@ -6405,6 +6459,20 @@ bool AnimationTrackEditor::is_grouping_tracks() {
 	return !view_group->is_pressed();
 }
 
+
```
AThousandShips, line 6462:
```suggestion
```

editor/animation_track_editor.cpp
```diff
@@ -1632,6 +1637,45 @@ void AnimationTimelineEdit::set_zoom(Range *p_zoom) {
 	zoom->connect("value_changed", callable_mp(this, &AnimationTimelineEdit::_zoom_changed));
 }
 
+
+void AnimationTimelineEdit::auto_fit() {
+
```
AThousandShips, line 1642:
```suggestion
void AnimationTimelineEdit::auto_fit() {
```

editor/animation_track_editor.cpp
```diff
@@ -1632,6 +1637,45 @@ void AnimationTimelineEdit::set_zoom(Range *p_zoom) {
...
+	if (!animation.is_valid()) {
+		return;
+	}
+
+	float anim_length = animation->get_length();
+
+	int timeline_width_pixels = get_size().width - get_buttons_width() - get_name_limit();
+
+	//I want a little buffer at the end... (5% looks nice and we should keep some space for the bezier handles)
```
AThousandShips, line 1651:
```suggestion
	// I want a little buffer at the end... (5% looks nice and we should keep some space for the bezier handles).
```

editor/animation_track_editor.cpp
```diff
@@ -1632,6 +1637,45 @@ void AnimationTimelineEdit::set_zoom(Range *p_zoom) {
...
+	timeline_width_pixels *= 0.95;
+		
+	//The technique is to reuse the _get_zoom_scale function directly to be sure that the auto_fit is always calculated
+	//the same way as the zoom slider. It's a little bit more calculation then doing the inverse of get_zoom_scale but
+	//it's really easier to inderstand and should always be accurate.
+	float new_zoom = zoom->get_max();
+	for (;;) {
+
```
AThousandShips, line 1659:
```suggestion
	while (true) {
```

editor/animation_bezier_editor.cpp
```diff
@@ -739,6 +748,60 @@ void AnimationBezierTrackEdit::set_filtered(bool p_filtered) {
...
+		if (maximum_value - minimum_value > CMP_EPSILON) {
+			timeline_v_zoom = (maximum_value - minimum_value) / ((get_size().height - timeline->get_size().height) * 0.9);
+		} else {
+			timeline_v_zoom = CMP_EPSILON;
+		}
+		queue_redraw();
+	}
+
+
+
```
AThousandShips, line 802:
```suggestion
```

editor/animation_bezier_editor.cpp
```diff
@@ -646,6 +617,44 @@ void AnimationBezierTrackEdit::_notification(int p_what) {
...
+			}
+		}
+	}
+
+	return true;
+}
+
+// Check if the curves for a track are displayed in the editor (not hidden). Includes the check on the track visibility.
+bool AnimationBezierTrackEdit::_is_track_curves_displayed(int track_index) {
+
```
AThousandShips, line 644:
```suggestion
```

editor/animation_bezier_editor.h
```diff
@@ -98,6 +98,8 @@ class AnimationBezierTrackEdit : public Control {
 	void _menu_selected(int p_index);
 
 	void _play_position_draw();
+	bool _is_track_displayed(int track_index);
+	bool _is_track_curves_displayed(int track_index);
```
AThousandShips, line 102:
```suggestion
	bool _is_track_displayed(int p_track_index);
	bool _is_track_curves_displayed(int p_track_index);
```
Prefix arguments with `p_`

editor/animation_track_editor.h
```diff
@@ -181,6 +181,8 @@ class AnimationTimelineEdit : public Range {
 	virtual void gui_input(const Ref<InputEvent> &p_event) override;
 	void _track_added(int p_track);
 
+	float _get_zoom_scale(double zoom_value) const;
```
AThousandShips, line 184:
```suggestion
	float _get_zoom_scale(double p_zoom_value) const;
```

editor/animation_track_editor.cpp
```diff
@@ -1305,7 +1305,12 @@ void AnimationTimelineEdit::_zoom_changed(double) {
 }
 
 float AnimationTimelineEdit::get_zoom_scale() const {
-	float zv = zoom->get_max() - zoom->get_value();
+	return _get_zoom_scale(zoom->get_value());
+}
+
+
+float AnimationTimelineEdit::_get_zoom_scale(double zoom_value) const {
```
AThousandShips, line 1312:
```suggestion
float AnimationTimelineEdit::_get_zoom_scale(double p_zoom_value) const {
```

AThousandShips:
Thank you for your contribution! I'd suggest you read [this](https://docs.godotengine.org/en/stable/contributing/development/code_style_guidelines.html) to see some of the style details for the code 🙂

editor/animation_track_editor.cpp
```diff
@@ -1305,7 +1305,12 @@ void AnimationTimelineEdit::_zoom_changed(double) {
 }
 
 float AnimationTimelineEdit::get_zoom_scale() const {
-	float zv = zoom->get_max() - zoom->get_value();
+	return _get_zoom_scale(zoom->get_value());
+}
+
+
```
AThousandShips, line 1311:
```suggestion
```

editor/animation_track_editor.cpp
```diff
@@ -1632,6 +1637,44 @@ void AnimationTimelineEdit::set_zoom(Range *p_zoom) {
 	zoom->connect("value_changed", callable_mp(this, &AnimationTimelineEdit::_zoom_changed));
 }
 
+void AnimationTimelineEdit::auto_fit() {
+
```
AThousandShips, line 1641:
```suggestion
```

editor/animation_track_editor.cpp
```diff
@@ -1632,6 +1637,44 @@ void AnimationTimelineEdit::set_zoom(Range *p_zoom) {
...
+	//I want a little buffer at the end... (5% looks nice and we should keep some space for the bezier handles)
+	timeline_width_pixels *= 0.95;
+		
+	//The technique is to reuse the _get_zoom_scale function directly to be sure that the auto_fit is always calculated
+	//the same way as the zoom slider. It's a little bit more calculation then doing the inverse of get_zoom_scale but
+	//it's really easier to inderstand and should always be accurate.
+	float new_zoom = zoom->get_max();
+	while(true) {
+
```
AThousandShips, line 1658:
```suggestion
```

editor/animation_track_editor.h
Hilderin: Adjusted in the amended commit.

editor/animation_track_editor.cpp
Hilderin: Adjusted in the amended commit.

editor/animation_track_editor.cpp
Hilderin: Adjusted in the amended commit.

editor/animation_track_editor.cpp
Hilderin: Adjusted in the amended commit.

AThousandShips:
Please use the format tool, see the link I sent 🙂

Hilderin:
Oh that's the point I missed, sorry, I'm really new to Open source project. I'm downling LLVM right now and I'll try it. Thanks!

TokageItLab (changes requested):
In the TrackEditor, this should be fine.

However, in the BezierEditor, this is similar in behavior to "focus" (Ctrl + F), so you can make a method with the range as an argument and "focus" and "auto fit" should share it.

Also, it performs a huge zoom when the vertical range is 0, so you must avoid the case as an exception, the same as with "focus". This should be avoidable by sharing methods.

CookieBadger:
@Hilderin are you still planning to finish this? Otherwise, I might help :)

Hilderin:
Yes, I was planning to work on it in the week to come. I was busy last week. Thanks for the offer @CookieBadger .

CookieBadger:
@Hilderin great to hear, and I really dig your PR :)

Hilderin:
I created a function _zoom_vertically in animation_bezier_editor and used it for the focus and auto_fit and I placed there a minimum height at 0.01 total (-0.005 to 0.005). I was not sure what should be a good minimum height value.
image

I think i messed up the PR with my attempt to merge from master and committing my last changes with a amend commit, amending the merge commit. You have a suggestion to fix that?

CookieBadger:
@Hilderin about the commits, as long as you did not accidentally delete anything that you weren't supposed to, squashing your commits should fix it, afaik

TokageItLab (changes requested):

Calinou (commented):

However, when clicking the autofit button, the timeline horizontal scroll isn't put at the beginning. This means that if you zoom in towards the end of an animation, the animation will still be scrolled after clicking the autofit button. You need to manually scroll towards the left to see the entire animation. I'd expect this scrolling to happen automatically.

Hilderin:
Wierd, I added this in AnimationTimelineEdit::auto_fit() to scroll the timeline to the left with my latest changes:
```
hscroll->set_value(0);
```

Do you means like that:
Before:
image
Alter:
image

Maybe you can show me with a screenshot?

CookieBadger:
Something that you should take into account, is that there can be keyframes in the negative time before the animation starts, and keyframes placed after the animation ends. This is for example useful when you want the animation to start or end in the middle of a non-linear easing. If you press the zoom button, it only adjusts to the animation length, not to the keyframes. Since in the Bezier editor, the vertical zoom also adjusts to the keyframes, I think it would be logical, if the button either adjusted 
a. for the time from the first to the last keyframe (+ buffer time in the end, as it is currently)
b. for the time from the first keyframe to the animation end or the last keyframe, whichever is longer (+ buffer in the end)
image

Hilderin:
Very interesting suggestion... I did not thought about that!
Just to be sure I understand what you suggest, do you think the following should work:
1. Set begin to zero and end at the animation end (animation lenght)
2. Loop for all keyframes:
3. If < begin, set begin to keyframe
4. If > end, set end to keyframe
5. Adjust horizontal zoom using (end - begin) + buffer in the end
6. Scroll horizontally to begin

This should work in Bezier Editor or not.

b. for the time from the first keyframe to the animation end or the last keyframe, whichever is longer (+ buffer in the end)

CookieBadger:
@Hilderin yes, that would work. Instead of scrolling to a `begin` timestamp you could also scroll to the minimum value of the scroll to `get_min()` of the HScrollBar. This way, you don't have to check where the first keyframe is.

Hilderin:
I thought about it and realized that I did not add shortcut for the new button. The text I used for the tooltip was too long and unclear so I renamed it for "Fit to panel", like in a text editor (ex: Google Docs, zoom: Fit). I choose Ctrl-Shift-F for the default keys, it does not seems to be in conflict with other shortcuts and the usage of the F key that is used for Focus seems appropriate.

image
image
image

What do you think?

CookieBadger:
I love the fact that it has a shortcut, I was actually going to suggest that. Maybe Ctrl+F or Shift+F would be more accessible?

Hilderin:
I changed the default shortcut for Ctrl-F, it's easier to use and still no conflict with others shortcuts (I think).

I'm just waiting on the answer of @TokageItLab about the minimum vertical height (CMP_EPSILON or something else).

editor/animation_bezier_editor.cpp
```diff
@@ -741,6 +749,63 @@ void AnimationBezierTrackEdit::set_filtered(bool p_filtered) {
...
+	}
+}
+
+void AnimationBezierTrackEdit::_zoom_vertically(real_t minimum_value, real_t maximum_value) {
+	timeline_v_scroll = (maximum_value + minimum_value) / 2.0;
+
+	real_t target_height = maximum_value - minimum_value;
+	if (target_height < 0.01) {
+		target_height = 0.01;
+	}
```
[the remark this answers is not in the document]
TokageItLab: 0.01 seems too large to me personally. Also, it would be preferable to skip it as a process when the value is too small as keeping the previous behavior.

editor/animation_track_editor.cpp
```diff
@@ -7179,6 +7267,19 @@ AnimationTrackEditor::AnimationTrackEditor() {
 	bottom_hb->add_child(zoom);
 	timeline->set_zoom(zoom);
 
+	auto_fit = memnew(Button);
+	auto_fit->set_flat(true);
+	auto_fit->connect("pressed", callable_mp(this, &AnimationTrackEditor::_auto_fit));
+	auto_fit->set_shortcut(ED_SHORTCUT("animation_editor/auto_fit", TTR("Fit to panel"), KeyModifierMask::CMD_OR_CTRL | Key::F));
```
TokageItLab, line 7273: `Ctrl + F` conflicts with search, so I would recommend `Alt + F`.

TokageItLab (changes requested):

TokageItLab (approved):
LGTM! Please squash the PR into one commit as follow [Pull request workflow](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase).

editor/animation_bezier_editor.h
```diff
@@ -188,6 +190,7 @@ class AnimationBezierTrackEdit : public Control {
 	void _draw_track(int p_track, const Color &p_color);
 
 	float _bezier_h_to_pixel(float p_h);
+	void _zoom_vertically(real_t minimum_value, real_t maximum_value);
```
AThousandShips, line 193:
```suggestion
	void _zoom_vertically(real_t p_minimum_value, real_t p_maximum_value);
```

editor/animation_bezier_editor.cpp
```diff
@@ -741,6 +749,60 @@ void AnimationBezierTrackEdit::set_filtered(bool p_filtered) {
...
+
+		int key_count = animation->track_get_key_count(i);
+
+		for (int j = 0; j < key_count; ++j) {
+			real_t value = animation->bezier_track_get_key_value(i, j);
+
+			minimum_value = MIN(value, minimum_value);
+			maximum_value = MAX(value, maximum_value);
+
+			//We also want to includes the handles...
```
AThousandShips, line 771:
```suggestion
			// We also want to includes the handles...
```

For all your comments

Hilderin:
I was able to squash my commits and made corrections on my comments.

akien-mga:

Hilderin:
Thanks!! And thanks all of you for your help and support :)

Proggle:
A much-needed UI improvement, great job :)

PR #84018, opened 2023-10-26
Add the ability to expose nodes for direct access in instantiated scenes

yahkr:
# (Updated 4/13/2026)
Implements Exposed Nodes, a tool that allows scene authors to explicitly define which child nodes are visible and editable by default when a scene is instanced.

- Closes [https://github.com/godotengine/godot-proposals/issues/8266](https://github.com/godotengine/godot-proposals/issues/8266)

Exposed nodes:
- Appear at the instance root in the Scene Tree
- Allow property modification
- Propagate one instancing level
- Can be re-exposed
- Can be defined via Advanced Import Settings

Pasted image 20260410134402

Related:
- Closes https://github.com/godotengine/godot-proposals/issues/3248
- Closes https://github.com/godotengine/godot-proposals/issues/4265
- Closes https://github.com/godotengine/godot-proposals/issues/9660
- Closes https://github.com/godotengine/godot-proposals/issues/5475
- Closes (?) https://github.com/godotengine/godot-proposals/issues/7803
- Closes (?) https://github.com/godotengine/godot-proposals/issues/9536
# Examples

 Sample Project  [mrp.zip](https://github.com/user-attachments/files/17603057/mrp.zip)

 Example by @ptlthg https://github.com/godotengine/godot/pull/84018#issuecomment-3267406194
# Technical Overview

## Serialization
Stored within the PackedScene:
```
[exposed path="VBoxContainer/PanelContainer/MarginContainer1/HBoxContainer/TitleLabel"]  
[exposed path="VBoxContainer/MarginContainer2/Content"]

Similar to how Editable Children is saved:
[editable path="..."]
```

## Metadata (defined in Node.h)
During `PackedScene::instantiate`:
- The stored exposed NodePaths are resolved.
- The corresponding nodes are marked with metadata.

- `META_EXPOSED_IN_OWNER`  
	- Marks that the node is declared exposed in the scene that owns it.
-  `META_MARKED_FOR_EXPOSURE`
	- Marked within the current scene that we want to expose or re-expose the node.
- `META_EXPOSED_IN_INSTANCE`  
	- Stores a reference to the highest-level instanced scene that re/exposes the node.
	- This encodes provenance and allows the editor to determine:    
	    1. Whether the stored instance's owner matches the currently edited scene.
	    2. If not, the node is not surfaced.
- `META_CONTAINS_EXPOSED_NODES`  
    Stored on the owner as a fast-path optimization to avoid unnecessary subtree scans.

[...]

PR #30623, opened 2019-07-16
Prevent editing properties managed by parent container

Awkor:
Addresses #28718 by making the properties that are managed by parent container not editable inside the inspector.

Currently the properties that get disabled are:
* Anchor
* Margin
* Position
* Scale
* Size

This is achieved by disabling mouse interactions and preventing focus on the aforementioned inspector properties.

This is how it looks when properties get disabled:
capture

*Bugsquad edit:* Fixes #28718 and fixes https://github.com/godotengine/godot-proposals/issues/2202.

Issue #19068, opened 2018-05-20
Scaled Controls don't align properly


Essojadojef:
**Godot version:** 3.0

**OS/device including version:**

**Issue description:**
Margins ignore the scale of the control.
The top-left corner is set as if the node was not scaled.

Scale: 1x1, Anchor: center right
![selezione_025](https://user-images.githubusercontent.com/33575693/40283718-055fa90a-5c83-11e8-839b-d2deba659c69.png)

Scale: 2x2, Anchor: center right (the transparent panel is also scaled 2x2, but the textureRect didn't align to it's center right)
![selezione_026](https://user-images.githubusercontent.com/33575693/40283742-55711b0e-5c83-11e8-86a5-5bacb5e62c01.png)

What was expected: (with Anchor: center left)
![selezione_028](https://user-images.githubusercontent.com/33575693/40283810-c30328a0-5c84-11e8-8952-2985e5345dc5.png)

**Steps to reproduce:**
Add a Control to a parent Control.
Change the child's scale.
Change the child's anchor from the Layout menu (to a layout different from top left).

**Minimal reproduction project:**
[align-scaled-issue.zip](https://github.com/godotengine/godot/files/2020719/align-scaled-issue.zip)

AndrewMorsillo:
Is there a workaround for this? From what I can tell this issue makes some (responsive) UIs impossible to create.

KoBeWi:
Well, the workaround is to avoid using `rect_scale`. I did numerous GUIs in Godot and never needed it.

> From what I can tell this issue makes some (responsive) UIs impossible to create.

Any examples?

Calinou:
> From what I can tell this issue makes some (responsive) UIs impossible to create.

See [Multiple resolutions](https://docs.godotengine.org/en/latest/tutorials/viewports/multiple_resolutions.html) in the documentation.

Klaider:
@Callinou This documentation link is broken. Archive: http://web.archive.org/web/20181105034839/https://docs.godotengine.org/en/latest/tutorials/viewports/multiple_resolutions.html

Klaider:
This is not a workaround for the issue though.

Calinou:
> @Callinou This documentation link is broken. Archive: http://web.archive.org/web/20181105034839/https://docs.godotengine.org/en/latest/tutorials/viewports/multiple_resolutions.html

[...]

PR #87081, opened 2024-01-11
Add transform offset to Control nodes

timoschwarzer:
This is an implementation of my proposal outlined in this discussion: https://github.com/godotengine/godot-proposals/issues/7053#issuecomment-1883816733

**See the most recent state of this PR here: https://github.com/godotengine/godot/pull/87081#issuecomment-3904492472**

tl;dr: Currently it's hard and complicated to animate Control nodes without breaking the layout, especially if the Control nodes are in deeply nested layout structures which is often the case. This PR adds new properties to Control nodes that specify a render transform that does not affect and is not affected by layouting (similar to CSS' `transform`). These make UI animations much less pain to work with

- `render_offset`: translation
- `render_scale`: scale
- `render_rotation`, rotation
- `render_transform_pivot`, pivot for scaling and rotating

`render_offset` and `render_transform_pivot` are relative to the size of the Control node by default (I think it's a very common usecase to rotate around and scale from the center) but it can be toggled to use absolute pixel values instead.

[Screencast from 2024-01-11 17-10-13.webm](https://github.com/godotengine/godot/assets/8849554/28f2f723-6b9f-4687-b7de-0e6c06695837)

## Considerations

~~Input is currently *not* affected by the render transform. That means if you e.g. apply a render offset to a button, the button is still clickable at its original position but drawn somewhere else. This might need some discussion whether that's desirable or not.~~ There is a toggle for this now.

### DIFF
--- a/doc/classes/Control.xml
+++ b/doc/classes/Control.xml
@@ -1098,6 +1098,41 @@
 			Distance between the node's top edge and its parent control, based on [member anchor_top].
 			Offsets are often controlled by one or multiple parent [Container] nodes, so you should not modify them manually if your node is a direct child of a [Container]. Offsets update automatically when you move or resize the node.
 		</member>
+		<member name="offset_transform_enabled" type="bool" setter="set_offset_transform_enabled" getter="is_offset_transform_enabled" default="false">
+			If [code]true[/code], applies all offset transform properties. Otherwise, no offset transform is applied and the properties have no effect.
+		</member>
+		<member name="offset_transform_pivot" type="Vector2" setter="set_offset_transform_pivot" getter="get_offset_transform_pivot" default="Vector2(0, 0)">
+			Pivot used by [member offset_transform_rotation] and [member offset_transform_scale] in absolute units.
+			The final pivot position is the combined value of this property and [member offset_transform_pivot_ratio].
+			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
+		</member>
+		<member name="offset_transform_pivot_ratio" type="Vector2" setter="set_offset_transform_pivot_ratio" getter="get_offset_transform_pivot_ratio" default="Vector2(0.5, 0.5)">
+			Same as [member offset_transform_pivot] but expressed in units relative to the [Control] [member size] where [code]Vector2(0, 0)[/code] is the top-left corner of this control, and [code]Vector2(1, 1)[/code] is its bottom-right corner.
+			The final pivot position is the combined value of this property and [member offset_transform_pivot].
+			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
+		</member>
+		<member name="offset_transform_position" type="Vector2" setter="set_offset_transform_position" getter="get_offset_transform_position" default="Vector2(0, 0)">
+			Position offset in absolute units. The final offset is the combined value of this property and [member offset_transform_position_ratio].
+			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
+		</member>
+		<member name="offset_transform_position_ratio" type="Vector2" setter="set_offset_transform_position_ratio" getter="get_offset_transform_position_ratio" default="Vector2(0, 0)">
+			Same as [member offset_transform_position] but expressed in units relative to the [Control] [member size] where [code]Vector2(0, 0)[/code] is the top-left corner of this control, and [code]Vector2(1, 1)[/code] is its bottom-right corner.
+			The final offset is the combined value of this property and [member offset_transform_position].
+			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
+		</member>
+		<member name="offset_transform_rotation" type="float" setter="set_offset_transform_rotation" getter="get_offset_transform_rotation" default="0.0">
+			Rotation offset. The rotation pivot is defined by [member offset_transform_pivot] and [member offset_transform_pivot_ratio].
+			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
+		</member>
+		<member name="offset_transform_scale" type="Vector2" setter="set_offset_transform_scale" getter="get_offset_transform_scale" default="Vector2(1, 1)">
+			Scale offset. The scale pivot is defined by [member offset_transform_pivot] and [member offset_transform_pivot_ratio].
+			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
+		</member>
+		<member name="offset_transform_visual_only" type="bool" setter="set_offset_transform_visual_only" getter="is_offset_transform_visual_only" default="true">
+			If [code]true[/code], the offset transforms is only applied visually and does not affect input. In other words, this Control will still receive input events at its original location before the offset transform is applied.
+			If [code]false[/code], the entire transform of this Control is affected and input events will register where the Control is visually.
+			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
+		</member>
 		<member name="physics_interpolation_mode" type="int" setter="set_physics_interpolation_mode" getter="get_physics_interpolation_mode" overrides="Node" enum="Node.PhysicsInterpolationMode" default="2" />
 		<member name="pivot_offset" type="Vector2" setter="set_pivot_offset" getter="get_pivot_offset" default="Vector2(0, 0)">
 			By default, the node's pivot is its top-left corner. When you change its [member rotation] or [member scale], it will rotate or scale around this pivot.
--- a/scene/gui/control.cpp
+++ b/scene/gui/control.cpp
@@ -528,6 +528,8 @@ void Control::_validate_property(PropertyInfo &p_property) const {
 	}
 	// Validate which positioning properties should be displayed depending on the parent and the layout mode.
 	Control *parent_control = get_parent_control();
+	bool is_anchor_offset_property_name = p_property.name.begins_with("offset_") && !p_property.name.begins_with("offset_transform_");
+
 	if (Engine::get_singleton()->is_editor_hint() && !parent_control) {
 		// If there is no parent control, display both anchor and container options.
 
@@ -539,12 +541,12 @@ void Control::_validate_property(PropertyInfo &p_property) const {
 
 		// Use the layout mode to display or hide advanced anchoring properties.
 		bool use_custom_anchors = _get_anchors_layout_preset() == -1; // Custom "preset".
-		if (!use_custom_anchors && (p_property.name.begins_with("anchor_") || p_property.name.begins_with("offset_") || p_property.name.begins_with("grow_"))) {
+		if (!use_custom_anchors && (p_property.name.begins_with("anchor_") || is_anchor_offset_property_name || p_property.name.begins_with("grow_"))) {
 			p_property.usage ^= PROPERTY_USAGE_EDITOR;
 		}
 	} else if (Object::cast_to<Container>(parent_control)) {
 		// If the parent is a container, display only container-related properties.
-		if (p_property.name.begins_with("anchor_") || p_property.name.begins_with("offset_") || p_property.name.begins_with("grow_") || p_property.name == "anchors_preset") {
+		if (p_property.name.begins_with("anchor_") || is_anchor_offset_property_name || p_property.name.begins_with("grow_") || p_property.name == "anchors_preset") {
 			p_property.usage ^= PROPERTY_USAGE_DEFAULT;
 		} else if (p_property.name == "position" || p_property.name == "rotation" || p_property.name == "scale" || p_property.name == "size" || p_property.name == "pivot_offset" || p_property.name == "pivot_offset_ratio") {
 			p_property.usage = PROPERTY_USAGE_EDITOR | PROPERTY_USAGE_READ_ONLY;
@@ -610,7 +612,7 @@ void Control::_validate_property(PropertyInfo &p_property) const {
 			p_property.usage ^= PROPERTY_USAGE_EDITOR;
 		}
 		bool use_custom_anchors = use_anchors && _get_anchors_layout_preset() == -1; // Custom "preset".
-		if (!use_custom_anchors && (p_property.name.begins_with("anchor_") || p_property.name.begins_with("offset_") || p_property.name.begins_with("grow_"))) {
+		if (!use_custom_anchors && (p_property.name.begins_with("anchor_") || is_anchor_offset_property_name || p_property.name.begins_with("grow_"))) {
 			p_property.usage ^= PROPERTY_USAGE_EDITOR;
 		}
 	}
@@ -721,6 +723,11 @@ Transform2D Control::_get_internal_transform() const {
 	// T(pivot_offset) * R(rotation) * S(scale) * T(-pivot_offset)
 	Transform2D xform(data.rotation, data.scale, 0.0f, get_combined_pivot_offset());
 	xform.translate_local(-get_combined_pivot_offset());
+
+	if (is_offset_transform_enabled() && !data.offset_transform->visual_only) {
+		xform *= get_offset_transform();
+	}
+
 	return xform;
 }
 
@@ -731,6 +738,10 @@ void Control::_update_canvas_item_transform() {
 	// We use a little workaround to avoid flickering when moving the pivot with _edit_set_pivot()
 	if (is_inside_tree() && Math::abs(Math::sin(data.rotation * 4.0f)) < 0.00001f && get_viewport()->is_snap_controls_to_pixels_enabled()) {
 		xform[2] = (xform[2] + Vector2(0.5, 0.5)).floor();
+	}
+
+	if (is_offset_transform_enabled() && data.offset_transform->visual_only) {
+		xform *= get_offset_transform();
 	}
 
 	RenderingServer::get_singleton()->canvas_item_set_transform(get_canvas_item(), xform);
@@ -1879,6 +1890,218 @@ void Control::set_stretch_ratio(real_t p_ratio) {
 real_t Control::get_stretch_ratio() const {
 	ERR_READ_THREAD_GUARD_V(0);
 	return data.expand;
+}
+
+// Offset transform.
+
+void Control::set_offset_transform_enabled(bool p_enabled) {
+	if (is_offset_transform_enabled() == p_enabled) {
+		return;
+	}
+
+	if (p_enabled) {
+		_ensure_allocated_offset_transform();
+		data.offset_transform->enabled = true;
+	} else {
+		// Never deallocate to not lose previously set values
+		data.offset_transform->enabled = false;
+	}
+
+	queue_redraw();
+	_notify_transform();
+	queue_accessibility_update();
+}
+
+bool Control::is_offset_transform_enabled() const {
+	return data.offset_transform != nullptr && data.offset_transform->enabled;
+}
+
+void Control::set_offset_transform_position(const Vector2 &p_offset) {
+	if (get_offset_transform_position() == p_offset) {
+		return;
+	}
+
+	_ensure_allocated_offset_transform();
+	data.offset_transform->translation_absolute = p_offset;
+
+	if (!data.offset_transform->enabled) {
+		return;
+	}
+
+	queue_redraw();
+	_notify_transform();
+	queue_accessibility_update();
+}
+
+Vector2 Control::get_offset_transform_position() const {
+	if (data.offset_transform == nullptr) {
+		return Data::OffsetTransform::DEFAULT_TRANSLATION_ABSOLUTE;
+	}
+
+	return data.offset_transform->translation_absolute;
+}
+
+void Control::set_offset_transform_position_ratio(const Vector2 &p_offset) {
+	if (get_offset_transform_position_ratio() == p_offset) {
+		return;
+	}
+
+	_ensure_allocated_offset_transform();
+	data.offset_transform->translation_relative = p_offset;
+
+	if (!data.offset_transform->enabled) {
+		return;
+	}
+
+	queue_redraw();
+	_notify_transform();
+	queue_accessibility_update();
+}
+
+Vector2 Control::get_offset_transform_position_ratio() const {
+	if (data.offset_transform == nullptr) {
+		return Data::OffsetTransform::DEFAULT_TRANSLATION_RELATIVE;
+	}
+
+	return data.offset_transform->translation_relative;
+}
+
+void Control::set_offset_transform_scale(const Vector2 &p_scale) {
+	if (get_offset_transform_scale() == p_scale) {
+		return;
+	}
+
+	_ensure_allocated_offset_transform();
+	data.offset_transform->scale = p_scale;
+
+	if (!data.offset_transform->enabled) {
+		return;
+	}
+
+	queue_redraw();
+	_notify_transform();
+	queue_accessibility_update();
+}
+
+Vector2 Control::get_offset_transform_scale() const {
+	if (data.offset_transform == nullptr) {
+		return Data::OffsetTransform::DEFAULT_SCALE;
+	}
+
+	return data.offset_transform->scale;
+}
+
+void Control::set_offset_transform_rotation(real_t p_rotation) {
+	if (get_offset_transform_rotation() == p_rotation) {
+		return;
+	}
+
+	_ensure_allocated_offset_transform();
+	data.offset_transform->rotation = p_rotation;
+
+	if (!data.offset_transform->enabled) {
+		return;
+	}
+
+	queue_redraw();
+	_notify_transform();
+	queue_accessibility_update();
+}
+
+real_t Control::get_offset_transform_rotation() const {
+	if (data.offset_transform == nullptr) {
+		return Data::OffsetTransform::DEFAULT_ROTATION;
+	}
+
+	return data.offset_transform->rotation;
+}
+
+void Control::set_offset_transform_pivot(const Vector2 &p_pivot) {
+	if (get_offset_transform_pivot() == p_pivot) {
+		return;
+	}
+
+	_ensure_allocated_offset_transform();
+	data.offset_transform->pivot_absolute = p_pivot;
+
+	if (!data.offset_transform->enabled) {
+		return;
+	}
+
+	queue_redraw();
+	_notify_transform();
+	queue_accessibility_update();
+}
+
+Vector2 Control::get_offset_transform_pivot() const {
+	if (data.offset_transform == nullptr) {
+		return Data::OffsetTransform::DEFAULT_PIVOT_ABSOLUTE;
+	}
+
+	return data.offset_transform->pivot_absolute;
+}
+
+void Control::set_offset_transform_pivot_ratio(const Vector2 &p_pivot) {
+	if (get_offset_transform_pivot_ratio() == p_pivot) {
+		return;
+	}
+
+	_ensure_allocated_offset_transform();
+	data.offset_transform->pivot_relative = p_pivot;
+
+	if (!data.offset_transform->enabled) {
+		return;
+	}
+
+	queue_redraw();
+	_notify_transform();
+	queue_accessibility_update();
+}
+
+Vector2 Control::get_offset_transform_pivot_ratio() const {
+	if (data.offset_transform == nullptr) {
+		return Data::OffsetTransform::DEFAULT_PIVOT_RELATIVE;
+	}
+
+	return data.offset_transform->pivot_relative;
+}
+
+void Control::set_offset_transform_visual_only(bool p_enabled) {
+	if (is_offset_transform_visual_only() == p_enabled) {
+		return;
+	}
+
+	_ensure_allocated_offset_transform();
+	data.offset_transform->visual_only = p_enabled;
+
+	if (!data.offset_transform->enabled) {
+		return;
+	}
+
+	queue_redraw();
+	_notify_transform();
+	queue_accessibility_update();
+}
+
+bool Control::is_offset_transform_visual_only() const {
+	if (data.offset_transform == nullptr) {
+		return Data::OffsetTransform::DEFAULT_VISUAL_ONLY;
+	}
+
+	return data.offset_transform->visual_only;
+}
+
+Transform2D Control::get_offset_transform() const {
+	if (!is_offset_transform_enabled()) {
+		return Transform2D();
+	}
+
+	Vector2 combined_translation = data.offset_transform->translation_absolute + data.offset_transform->translation_relative * get_size();
+	Vector2 combined_pivot = data.offset_transform->pivot_absolute + data.offset_transform->pivot_relative * get_size();
+
+	Transform2D offset_xform(data.offset_transform->rotation, data.offset_transform->scale, 0.0f, combined_pivot + combined_translation);
+	offset_xform.translate_local(-combined_pivot);
+	return offset_xform;
 }
 
 // Input events.
@@ -3701,6 +3924,14 @@ Control *Control::make_custom_tooltip(const String &p_text) const {
 	return Object::cast_to<Control>(ret);
 }
 
+void Control::_ensure_allocated_offset_transform() {
+	if (data.offset_transform != nullptr) {
+		return;
+	}
+
+	data.offset_transform = memnew(Data::OffsetTransform);
+}
+
 // Base object overrides.
 
 void Control::_accessibility_action_foucs(const Variant &p_data) {
@@ -4070,6 +4301,23 @@ void Control::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_v_size_flags", "flags"), &Control::set_v_size_flags);
 	ClassDB::bind_method(D_METHOD("get_v_size_flags"), &Control::get_v_size_flags);
 
+	ClassDB::bind_method(D_METHOD("set_offset_transform_enabled", "enabled"), &Control::set_offset_transform_enabled);
+	ClassDB::bind_method(D_METHOD("is_offset_transform_enabled"), &Control::is_offset_transform_enabled);
+	ClassDB::bind_method(D_METHOD("set_offset_transform_position", "offset"), &Control::set_offset_transform_position);
+	ClassDB::bind_method(D_METHOD("get_offset_transform_position"), &Control::get_offset_transform_position);
+	ClassDB::bind_method(D_METHOD("set_offset_transform_position_ratio", "offset"), &Control::set_offset_transform_position_ratio);
+	ClassDB::bind_method(D_METHOD("get_offset_transform_position_ratio"), &Control::get_offset_transform_position_ratio);
+	ClassDB::bind_method(D_METHOD("set_offset_transform_scale", "scale"), &Control::set_offset_transform_scale);
+	ClassDB::bind_method(D_METHOD("get_offset_transform_scale"), &Control::get_offset_transform_scale);
+	ClassDB::bind_method(D_METHOD("set_offset_transform_rotation", "rotation"), &Control::set_offset_transform_rotation);
+	ClassDB::bind_method(D_METHOD("get_offset_transform_rotation"), &Control::get_offset_transform_rotation);
+	ClassDB::bind_method(D_METHOD("set_offset_transform_pivot", "pivot"), &Control::set_offset_transform_pivot);
+	ClassDB::bind_method(D_METHOD("get_offset_transform_pivot"), &Control::get_offset_transform_pivot);
+	ClassDB::bind_method(D_METHOD("set_offset_transform_pivot_ratio", "pivot"), &Control::set_offset_transform_pivot_ratio);
+	ClassDB::bind_method(D_METHOD("get_offset_transform_pivot_ratio"), &Control::get_offset_transform_pivot_ratio);
+	ClassDB::bind_method(D_METHOD("set_offset_transform_visual_only", "enabled"), &Control::set_offset_transform_visual_only);
+	ClassDB::bind_method(D_METHOD("is_offset_transform_visual_only"), &Control::is_offset_transform_visual_only);
+
 	ClassDB::bind_method(D_METHOD("set_theme", "theme"), &Control::set_theme);
 	ClassDB::bind_method(D_METHOD("get_theme"), &Control::get_theme);
 
@@ -4277,6 +4525,16 @@ void Control::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "size_flags_vertical", PROPERTY_HINT_FLAGS, "Fill:1,Expand:2,Shrink Center:4,Shrink End:8"), "set_v_size_flags", "get_v_size_flags");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "size_flags_stretch_ratio", PROPERTY_HINT_RANGE, "0,20,0.01,or_greater"), "set_stretch_ratio", "get_stretch_ratio");
 
+	ADD_GROUP("Offset Transform", "offset_transform_");
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "offset_transform_enabled", PROPERTY_HINT_GROUP_ENABLE), "set_offset_transform_enabled", "is_offset_transform_enabled");
+	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "offset_transform_position", PROPERTY_HINT_NONE, "suffix:px"), "set_offset_transform_position", "get_offset_transform_position");
+	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "offset_transform_position_ratio", PROPERTY_HINT_NONE), "set_offset_transform_position_ratio", "get_offset_transform_position_ratio");
+	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "offset_transform_scale", PROPERTY_HINT_LINK), "set_offset_transform_scale", "get_offset_transform_scale");
+	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "offset_transform_rotation", PROPERTY_HINT_RANGE, "-360,360,0.1,or_less,or_greater,radians_as_degrees"), "set_offset_transform_rotation", "get_offset_transform_rotation");
+	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "offset_transform_pivot", PROPERTY_HINT_NONE, "suffix:px"), "set_offset_transform_pivot", "get_offset_transform_pivot");
+	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "offset_transform_pivot_ratio", PROPERTY_HINT_NONE), "set_offset_transform_pivot_ratio", "get_offset_transform_pivot_ratio");
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "offset_transform_visual_only", PROPERTY_HINT_NONE), "set_offset_transform_visual_only", "is_offset_transform_visual_only");
+
 	ADD_GROUP("Localization", "");
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "localize_numeral_system"), "set_localize_numeral_system", "is_localizing_numeral_system");
 
@@ -4454,6 +4712,10 @@ Control::Control() {
 
 Control::~Control() {
 	memdelete(data.theme_owner);
+
+	if (data.offset_transform != nullptr) {
+		memdelete(data.offset_transform);
+	}
 
 	// Resources need to be disconnected.
 	for (KeyValue<StringName, Ref<Texture2D>> &E : data.theme_icon_override) {
--- a/scene/gui/control.h
+++ b/scene/gui/control.h
@@ -183,6 +183,25 @@ private:
 
 	// This Data struct is to avoid namespace pollution in derived classes.
 	struct Data {
+		struct OffsetTransform {
+			static constexpr Vector2 DEFAULT_TRANSLATION_ABSOLUTE = Vector2();
+			static constexpr Vector2 DEFAULT_TRANSLATION_RELATIVE = Vector2();
+			static constexpr Vector2 DEFAULT_SCALE = Vector2(1, 1);
+			static constexpr real_t DEFAULT_ROTATION = 0.0;
+			static constexpr Vector2 DEFAULT_PIVOT_ABSOLUTE = Vector2();
+			static constexpr Vector2 DEFAULT_PIVOT_RELATIVE = Vector2(0.5, 0.5);
+			static constexpr bool DEFAULT_VISUAL_ONLY = true;
+
+			bool enabled = false;
+			Vector2 translation_absolute = DEFAULT_TRANSLATION_ABSOLUTE;
+			Vector2 translation_relative = DEFAULT_TRANSLATION_RELATIVE;
+			Vector2 scale = DEFAULT_SCALE;
+			real_t rotation = DEFAULT_ROTATION;
+			Vector2 pivot_absolute = DEFAULT_PIVOT_ABSOLUTE;
+			Vector2 pivot_relative = DEFAULT_PIVOT_RELATIVE;
+			bool visual_only = DEFAULT_VISUAL_ONLY;
+		};
+
 		bool initialized = false;
 
 		// Global relations.
@@ -213,6 +232,8 @@ private:
 		Vector2 scale = Vector2(1, 1);
 		Vector2 pivot_offset;
 		Vector2 pivot_offset_ratio;
+
+		OffsetTransform *offset_transform = nullptr;
 
 		Point2 pos_cache;
 		Size2 size_cache;
@@ -361,6 +382,8 @@ private:
 	void _theme_changed();
 	void _notify_theme_override_changed();
 	void _invalidate_theme_cache();
+
+	void _ensure_allocated_offset_transform();
 
 	// Extra properties.
 
@@ -560,6 +583,26 @@ public:
 	BitField<SizeFlags> get_v_size_flags() const;
 	void set_stretch_ratio(real_t p_ratio);
 	real_t get_stretch_ratio() const;
+
+	// Offset transform.
+
+	void set_offset_transform_enabled(bool p_enabled);
+	bool is_offset_transform_enabled() const;
+	void set_offset_transform_position(const Vector2 &p_offset);
+	Vector2 get_offset_transform_position() const;
+	void set_offset_transform_position_ratio(const Vector2 &p_offset);
+	Vector2 get_offset_transform_position_ratio() const;
+	void set_offset_transform_scale(const Vector2 &p_scale);
+	Vector2 get_offset_transform_scale() const;
+	void set_offset_transform_rotation(real_t p_rotation);
+	real_t get_offset_transform_rotation() const;
+	void set_offset_transform_pivot(const Vector2 &p_pivot);
+	Vector2 get_offset_transform_pivot() const;
+	void set_offset_transform_pivot_ratio(const Vector2 &p_pivot);
+	Vector2 get_offset_transform_pivot_ratio() const;
+	void set_offset_transform_visual_only(bool p_enabled);
+	bool is_offset_transform_visual_only() const;
+	Transform2D get_offset_transform() const;
 
 	// Input events.
 
--- a/editor/scene/gui/control_editor_plugin.cpp
+++ b/editor/scene/gui/control_editor_plugin.cpp
@@ -1155,10 +1155,78 @@ ControlEditorToolbar *ControlEditorToolbar::singleton = nullptr;
 
 // Editor plugin.
 
+void ControlOffsetTransformPreview::edit(Control *p_control) {
+	if (p_control == selected_control) {
+		return;
+	}
+
+	const Callable update_overlays = callable_mp(plugin, &EditorPlugin::update_overlays);
+
+	if (selected_control) {
+		selected_control->disconnect(SceneStringName(draw), update_overlays);
+	}
+
+	selected_control = p_control;
+
+	if (selected_control) {
+		selected_control->connect(SceneStringName(draw), update_overlays);
+	}
+
+	plugin->update_overlays();
+}
+
+void ControlOffsetTransformPreview::forward_canvas_draw_over_viewport(Control *p_overlay) const {
+	if (!selected_control || !selected_control->is_offset_transform_enabled() || selected_control->is_offset_transform_visual_only()) {
+		return;
+	}
+
+	Point2 top_left = Point2();
+	Point2 bottom_right = selected_control->get_size();
+	Point2 top_right = Point2(bottom_right.x, top_left.y);
+	Point2 bottom_left = Point2(top_left.x, bottom_right.y);
+
+	Transform2D control_transform_without_offset = selected_control->get_global_transform() * selected_control->get_offset_transform().affine_inverse();
+	top_left = control_transform_without_offset.xform(top_left);
+	bottom_right = control_transform_without_offset.xform(bottom_right);
+	top_right = control_transform_without_offset.xform(top_right);
+	bottom_left = control_transform_without_offset.xform(bottom_left);
+
+	Transform2D canvas_transform = CanvasItemEditor::get_singleton()->get_canvas_transform();
+	top_left = canvas_transform.xform(top_left);
+	bottom_right = canvas_transform.xform(bottom_right);
+	top_right = canvas_transform.xform(top_right);
+	bottom_left = canvas_transform.xform(bottom_left);
+
+	Color color = Color(0.5, 0.5, 0.5, 0.75);
+	p_overlay->draw_dashed_line(top_left, top_right, color, 5.0, 4.0);
+	p_overlay->draw_dashed_line(top_right, bottom_right, color, 5.0, 4.0);
+	p_overlay->draw_dashed_line(bottom_right, bottom_left, color, 5.0, 4.0);
+	p_overlay->draw_dashed_line(bottom_left, top_left, color, 5.0, 4.0);
+}
+
+ControlOffsetTransformPreview::ControlOffsetTransformPreview(EditorPlugin *p_plugin) {
+	plugin = p_plugin;
+}
+
+void ControlEditorPlugin::edit(Object *p_object) {
+	offset_transform_preview->edit(Object::cast_to<Control>(p_object));
+}
+
+bool ControlEditorPlugin::handles(Object *p_object) const {
+	return Object::cast_to<Control>(p_object) != nullptr;
+}
+
+void ControlEditorPlugin::forward_canvas_draw_over_viewport(Control *p_overlay) {
+	offset_transform_preview->forward_canvas_draw_over_viewport(p_overlay);
+}
+
 ControlEditorPlugin::ControlEditorPlugin() {
 	toolbar = memnew(ControlEditorToolbar);
 	toolbar->hide();
 	add_control_to_container(CONTAINER_CANVAS_EDITOR_MENU, toolbar);
+
+	offset_transform_preview = memnew(ControlOffsetTransformPreview(this));
+	EditorNode::get_singleton()->get_gui_base()->add_child(offset_transform_preview);
 
 	Ref<EditorInspectorPluginControl> plugin;
 	plugin.instantiate();
--- a/editor/scene/gui/control_editor_plugin.h
+++ b/editor/scene/gui/control_editor_plugin.h
@@ -244,14 +244,36 @@ public:
 	ControlEditorToolbar();
 };
 
+class ControlOffsetTransformPreview : public Control {
+	GDCLASS(ControlOffsetTransformPreview, Control);
+
+	EditorPlugin *plugin = nullptr;
+	Control *selected_control = nullptr;
+
+	friend class ControlEditorPlugin;
+
+public:
+	void edit(Control *p_control);
+
+	void forward_canvas_draw_over_viewport(Control *p_overlay) const;
+
+	ControlOffsetTransformPreview(EditorPlugin *p_plugin);
+};
+
 // Editor plugin.
 class ControlEditorPlugin : public EditorPlugin {
 	GDCLASS(ControlEditorPlugin, EditorPlugin);
 
 	ControlEditorToolbar *toolbar = nullptr;
+	ControlOffsetTransformPreview *offset_transform_preview = nullptr;
 
 public:
 	virtual String get_plugin_name() const override { return "Control"; }
+
+	virtual void edit(Object *p_object) override;
+	virtual bool handles(Object *p_object) const override;
+
+	virtual void forward_canvas_draw_over_viewport(Control *p_overlay) override;
 
 	ControlEditorPlugin();
 };


scene/gui/control.h
```diff
@@ -190,6 +190,13 @@ class Control : public CanvasItem {
 		Vector2 scale = Vector2(1, 1);
 		Vector2 pivot_offset;
 
+		Vector2 render_offset;
+		bool render_offset_relative_to_size = true;
+		Vector2 render_scale = Vector2(1, 1);
+		real_t render_rotation = 0.0;
+		Vector2 render_transform_pivot = Vector2(0, 0);
```
AThousandShips, line 197:
```suggestion
		Vector2 render_transform_pivot;
```
Unnecessary default value

scene/gui/control.cpp
```diff
@@ -3415,6 +3493,19 @@ void Control::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("find_next_valid_focus"), &Control::find_next_valid_focus);
 	ClassDB::bind_method(D_METHOD("find_valid_focus_neighbor", "side"), &Control::find_valid_focus_neighbor);
 
+	ClassDB::bind_method(D_METHOD("set_render_offset"), &Control::set_render_offset);
```
AThousandShips, line 3496:
```suggestion
	ClassDB::bind_method(D_METHOD("set_render_offset", "offset"), &Control::set_render_offset);
```

scene/gui/control.cpp
```diff
@@ -3415,6 +3493,19 @@ void Control::_bind_methods() {
...
+	ClassDB::bind_method(D_METHOD("get_render_offset"), &Control::get_render_offset);
+	ClassDB::bind_method(D_METHOD("set_render_offset_relative_to_size"), &Control::set_render_offset_relative_to_size);
```
AThousandShips, line 3498:
```suggestion
	ClassDB::bind_method(D_METHOD("set_render_offset_relative_to_size", "relative"), &Control::set_render_offset_relative_to_size);
```

scene/gui/control.cpp
```diff
@@ -3415,6 +3493,19 @@ void Control::_bind_methods() {
...
+	ClassDB::bind_method(D_METHOD("get_render_offset_relative_to_size"), &Control::get_render_offset_relative_to_size);
+	ClassDB::bind_method(D_METHOD("set_render_scale"), &Control::set_render_scale);
```
AThousandShips, line 3500:
```suggestion
	ClassDB::bind_method(D_METHOD("set_render_scale", "scale"), &Control::set_render_scale);
```

scene/gui/control.cpp
```diff
@@ -3415,6 +3493,19 @@ void Control::_bind_methods() {
...
+	ClassDB::bind_method(D_METHOD("get_render_scale"), &Control::get_render_scale);
+	ClassDB::bind_method(D_METHOD("set_render_rotation"), &Control::set_render_rotation);
```
AThousandShips, line 3502:
```suggestion
	ClassDB::bind_method(D_METHOD("set_render_rotation", "rotation"), &Control::set_render_rotation);
```

scene/gui/control.cpp
```diff
@@ -3415,6 +3493,19 @@ void Control::_bind_methods() {
...
+	ClassDB::bind_method(D_METHOD("get_render_rotation"), &Control::get_render_rotation);
+	ClassDB::bind_method(D_METHOD("set_render_transform_pivot"), &Control::set_render_transform_pivot);
```
AThousandShips, line 3504:
```suggestion
	ClassDB::bind_method(D_METHOD("set_render_transform_pivot", "pivot"), &Control::set_render_transform_pivot);
```

scene/gui/control.cpp
```diff
@@ -3415,6 +3493,19 @@ void Control::_bind_methods() {
...
+	ClassDB::bind_method(D_METHOD("get_render_transform_pivot"), &Control::get_render_transform_pivot);
+	ClassDB::bind_method(D_METHOD("set_render_transform_pivot_relative_to_size"), &Control::set_render_transform_pivot_relative_to_size);
```
AThousandShips, line 3506:
```suggestion
	ClassDB::bind_method(D_METHOD("set_render_transform_pivot_relative_to_size", "relative"), &Control::set_render_transform_pivot_relative_to_size);
```

scene/gui/control.cpp
```diff
@@ -1667,6 +1685,66 @@ Size2 Control::get_custom_minimum_size() const {
...
+	queue_redraw();
+	_notify_transform();
+}
+
+Vector2 Control::get_render_transform_pivot() const {
+	return data.render_transform_pivot;
+}
+
+void Control::set_render_transform_pivot_relative_to_size(bool p_relative) {
+	data.render_transform_pivot_relative_to_size = p_relative;
```
AThousandShips, line 1739:
```suggestion
	if (data.render_transform_pivot_relative_to_size == p_relative) {
		return;
	}
	data.render_transform_pivot_relative_to_size = p_relative;
```
Use early exit here

scene/gui/control.cpp
```diff
@@ -1667,6 +1685,66 @@ Size2 Control::get_custom_minimum_size() const {
...
+	queue_redraw();
+	_notify_transform();
+}
+
+Vector2 Control::get_render_offset() const {
+	return data.render_offset;
+}
+
+void Control::set_render_offset_relative_to_size(bool p_relative) {
+	data.render_offset_relative_to_size = p_relative;
```
AThousandShips, line 1699:
```suggestion
	if (data.render_offset_relative_to_size == p_relative) {
		return;
	}
	data.render_offset_relative_to_size = p_relative;
```

fire:
I remember Node2D has properties.

Vector2 | position | Vector2(0, 0)
float | rotation | 0.0
Vector2 | scale | Vector2(1, 1)

How are these different?

Mickeon:
One could argue that both `render_rotation` and `render_scale` are not that useful since (_unless I'm severely mistaken_) Controls already ignore both rotation and scale within containers and when moving anchors/margins.

timoschwarzer:
@Mickeon this is currently considered a bug: https://github.com/godotengine/godot/issues/19068

That's why I added rotation and scale render transforms separately although they have the same result currently :)

timoschwarzer:
@Mickeon yeah, that's why I added render_rotation. So if the bug with rotation gets fixed eventually, people who used it for animation can switch to render_rotation and people who want rotated controls to actually use more space can finally do that. Same for scale and render_scale.

djrain:
How this would affect children of the animated node, if at all? In our games we're usually animating whole control-based scenes or Containers, like a playing card or something, frequently containing both Controls and Node2D, so the whole thing needs to be animated as a unit. I currently achieve that by applying an animation transform to the node but only after any parent Container does its layout.

timoschwarzer:
@djrain the current implementation sets the CanvasItem transform of the Control node, so it affects all children as well.

timoschwarzer:
@AThousandShips thanks for the review :raised_hands: applied everything and fixed code style issues

scene/gui/control.cpp
```diff
@@ -3576,6 +3675,14 @@ void Control::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "size_flags_vertical", PROPERTY_HINT_FLAGS, "Fill:1,Expand:2,Shrink Center:4,Shrink End:8"), "set_v_size_flags", "get_v_size_flags");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "size_flags_stretch_ratio", PROPERTY_HINT_RANGE, "0,20,0.01,or_greater"), "set_stretch_ratio", "get_stretch_ratio");
 
+	ADD_GROUP("Render Transform", "render_");
+	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "render_offset", PROPERTY_HINT_NONE), "set_render_offset", "get_render_offset");
```
AThousandShips, line 3679:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "render_offset", PROPERTY_HINT_NONE, "suffix:px), "set_render_offset", "get_render_offset");
```

scene/gui/control.cpp
```diff
@@ -3576,6 +3675,14 @@ void Control::_bind_methods() {
...
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "render_offset_relative_to_size", PROPERTY_HINT_NONE), "set_render_offset_relative_to_size", "get_render_offset_relative_to_size");
```
AThousandShips, line 3680:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "render_offset_relative_to_size"), "set_render_offset_relative_to_size", "get_render_offset_relative_to_size");
```

scene/gui/control.cpp
```diff
@@ -3576,6 +3675,14 @@ void Control::_bind_methods() {
...
+	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "render_scale", PROPERTY_HINT_NONE), "set_render_scale", "get_render_scale");
```
AThousandShips, line 3681:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "render_scale"), "set_render_scale", "get_render_scale");
```

scene/gui/control.cpp
```diff
@@ -3576,6 +3675,14 @@ void Control::_bind_methods() {
...
+	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "render_rotation", PROPERTY_HINT_NONE, "-360,360,0.1,or_less,or_greater,radians_as_degrees"), "set_render_rotation", "get_render_rotation");
+	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "render_transform_pivot", PROPERTY_HINT_NONE), "set_render_transform_pivot", "get_render_transform_pivot");
```
AThousandShips, line 3683:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "render_transform_pivot", PROPERTY_HINT_NONE, "suffix:pc"), "set_render_transform_pivot", "get_render_transform_pivot");
```

scene/gui/control.cpp
```diff
@@ -3576,6 +3675,14 @@ void Control::_bind_methods() {
...
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "render_transform_pivot_relative_to_size", PROPERTY_HINT_NONE), "set_render_transform_pivot_relative_to_size", "get_render_transform_pivot_relative_to_size");
```
AThousandShips, line 3684:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "render_transform_pivot_relative_to_size"), "set_render_transform_pivot_relative_to_size", "get_render_transform_pivot_relative_to_size");
```

scene/gui/control.cpp
AThousandShips: Still unresolved

scene/gui/control.cpp
timoschwarzer: I had this first but when using relative-to-size units it's not pixels anymore. Is there a way to dynamically set the suffix based on another property?

scene/gui/control.cpp
timoschwarzer: see above

scene/gui/control.cpp
timoschwarzer: Whoops!

scene/gui/control.cpp
AThousandShips: Yes, using `_validate_property`

timoschwarzer:
@AThousandShips now it should have everything except the px suffix

scene/gui/control.cpp
AThousandShips: But otherwise the `PROPERTY_HINT_NONE` is unnecessary 🙂

timoschwarzer: I implemented it but I couldn't figure out yet how to make it re-evaluate the properties when `render_offset_relative_to_size` or `render_transform_pivot_relative_to_size` change. `notify_property_list_changed()` was freezing for a second and resetting scroll in the inspector so that's suboptimal :sweat_smile:

AThousandShips: That's the one, there's something else wrong if it does that

timoschwarzer: Okay, added it. Thanks! :raised_hands:

Mickeon:
Two things:
- You should add these lines (and ideally roughly document them) in `Control.xml` otherwise [that Linux build](https://github.com/godotengine/godot/actions/runs/8020071560/job/21909161475?pr=87081) will never compile
    
- When the PR is in "Draft" state it means it is not ready for review. I don't know if this is intentional.

timoschwarzer:
@Mickeon I added the docs and removed the WIP status so the PR should be ready for review now :slightly_smiling_face:

Mickeon:
That's good, I... hmm... This is not a review but some thoughts:
- This PR could've used a separate proposal to further discuss potential solutions to the issue.
    - I probably should've recommended this much sooner, but everyone has their own things to worry about.
- The PR as is adds a LOT of properties to all Controls, which is most certainly hefty. It does introduce overhead, yet a grand majority of Control nodes in your standard application are not going to make use of them, or at least not all the time.
    - Because of it, in my opinion, it's worth being careful if this PR is even considered as is. Of course as you can see everyone is desiring _something_ like this, but having it for _every_ Control may be excessive. There may be multiple solutions to the same issue worth evaluating.

Calinou (commented):
Tested locally, it works as expected. There's the question of whether we should shift input coordinates as well, but this isn't always desired if you're using this for animation purposes. For instance, when hovering an UI element and shifting it upwards, if you held your mouse cursor on the bottom edge of the element, you could end up in a situation where the UI element is being hovered and unhovered continuously in a loop. Not shifting input coordinates prevents this kind of situation, at the cost of less a precise idea of where you can actually click the button (particularly for rotated buttons).

https://github.com/godotengine/godot-proposals/issues/4338 would be a particularly good pairing with this PR, so that you can easily extend the clickable area of your buttons to cater for the space they can take during an animation.

**Testing project:** [test_pr_87081.zip](https://github.com/godotengine/godot/files/14488602/test_pr_87081.zip)

https://github.com/godotengine/godot/assets/180032/7ef07896-224a-4c09-9243-d86487adeea1

timoschwarzer:
@Mickeon I can only speak for myself, but I've had more use for the newly added properties in contrast to existing `position`, `rotation` and `scale` properties. In addition to that, the point of this PR is to make *all* Controls easily animatable without messing with layouting. I don't see why only some Controls should have the render transform properties and some shouldn't (where would you draw the line here?).

I didn't measure it but I can't imagine the overhead being even noticable. The final transform is not calculated every frame but is cached iirc.

conde2:
First of all, amazing pull request, indeed very useful.

I would only suggest to check if you're not setting the same value in the setters. To avoid a wrong notify and a queue draw.

markdibarry:
Also, rotation already has limited use with controls since `clip_contents` is already not compatible.
image

filkata:
Hoping that this is still going to get merged.

Calinou:
@filkata Please don't bump issues without contributing significant new information. Use the :+1: reaction button on the first post instead.

scene/gui/control.cpp
```diff
@@ -483,6 +483,18 @@ void Control::_validate_property(PropertyInfo &p_property) const {
...
+		p_property.hint_string = data.render_offset_relative_to_size
+				? ""
+				: "suffix:px";
+	}
+
+	if (p_property.name == "render_transform_pivot") {
+		p_property.hint_string = data.render_transform_pivot_relative_to_size
+				? ""
+				: "suffix:px";
+	}
```
Mickeon, line 496: Can't this be shortened to the following?
```suggestion
	if (p_property.name == "render_offset" && !data.render_offset_relative_to_size) {
		p_property.hint_string = "suffix:px";
	}

	if (p_property.name == "render_transform_pivot" && !data.render_transform_pivot_relative_to_size) {
		p_property.hint_string = "suffix:px";
	}
```

scene/gui/control.h
```diff
@@ -493,6 +500,20 @@ class Control : public CanvasItem {
 	void set_custom_minimum_size(const Size2 &p_custom);
 	Size2 get_custom_minimum_size() const;
 
+	// Render transform
```
Mickeon, line 503:
```suggestion
	// Render transform.
```

Mickeon (commented):
I will leave it to someone else to speak about the implementation.

Not a fan of "_layouting_" at all, or "absolute pixel values", though I cannot think of any other way of wording it at the moment.

doc/classes/Control.xml
```diff
@@ -1019,6 +1019,24 @@
 		<member name="position" type="Vector2" setter="_set_position" getter="get_position" default="Vector2(0, 0)">
 			The node's position, relative to its containing node. It corresponds to the rectangle's top-left corner. The property is not affected by [member pivot_offset].
 		</member>
+		<member name="render_offset" type="Vector2" setter="set_render_offset" getter="get_render_offset" default="Vector2(0, 0)">
+			Position offset applied after layouting. If [member render_offset_relative_to_size] is true, the offset is relative to the Control's size, otherwise it represents absolute pixel values.
+		</member>
+		<member name="render_offset_relative_to_size" type="bool" setter="set_render_offset_relative_to_size" getter="get_render_offset_relative_to_size" default="true">
+			If true, [member render_offset] is relative to the Control size, otherwise it is absolute pixel values.
```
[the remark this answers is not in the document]
timoschwarzer: The size

scene/gui/control.cpp
timoschwarzer: Wouldn't that prevent it from being set back to "" if one of the `*_relative_to_size` properties is set back to false?

Mickeon: Like `_get_property_list()`, `_validate_property()` is called every time properties need to be refreshed (on `notify_property_list_changed()`).
I may be completely misunderstanding it, but that would also mean the `hint_string` is reset back to what is defined in the **PROPERTY_INFO** macro (in this case, nothing).

Still, no deal breaker. Perhaps the current code is even better as is (although the ternary could be in one line)

fire:
The general conversation form https://github.com/godotengine/godot-proposals/issues/7053 seems to be the maintainers are divided on adding this. 

When there's a clear agreement pull requests get merged quickly, but with disagreement enhancements get stuck.

Mickeon:
I do really appreciate that @timoschwarzer has been keeping this PR rebased and up-to-date as much as possible. I'm not part of the team of interest, but I _am_ one of many Godot users.[My worry is the same as it used to be from the start](https://github.com/godotengine/godot/pull/87081). That is, nodes in general are not exactly light in memory and the additional overhead may prove problematic, given that most **Control** nodes do not need this.

So the question to me always remains: is the trade-off worth it? Regardless, the feature is obviously very, very desired.

timoschwarzer:
@Mickeon would it help to benchmark the actual impact this change would have, especially in UI heavy projects such as the Godot editor itself? I think I could do that.

Mickeon:
I can't say benchmarking would do a lot for the people of interest, because there's still the question of whether **Control** nodes should be capable of this in the first place. 
But it may confirm or deny my own suspicious about the "unnecessary" overhead. I do repeat that there _may_ be other ways to accomplish this that could be more satisfactory.

timoschwarzer:
I will be busy for two weeks now but will try to get some meaningful data then.

> I do repeat that there *may* be other ways to accomplish this that could be more satisfactory

I would of course like to see more sophisticated ways to accomplish animations for Control nodes but so far the other solutions I have seen came with limitations and aren't as flexible while I do see and have had use cases to animate both Control and Container nodes. When you say "more satisfactory", do you tie that directly to "not having render transform properties on all Control nodes"?

timoschwarzer:
@Mickeon I finally did some benchmarks:


## Benchmark

Sample 1: Master branch at 1fc82087658066935bed9e1350d62e334c7e0309
Sample 2: Same as Sample 1 plus this commit

For each sample, I launched the editor into an empty project, let everything load and then closed the editor again. Then I launch the same editor executable with a profiler, let it open and close it again. Each sample was profiled three times, the results per sample were roughly the same.

System:
```
Godot v4.4.dev (48bcad089) - Arch Linux #1 ZEN SMP PREEMPT_DYNAMIC Thu, 12 Sep 2024 17:17:51 +0000 on Wayland - X11 display driver, Multi-window, 2 monitors - Vulkan (Forward+) - dedicated Intel(R) Arc(tm) A770 Graphics (DG2) - Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz (16 threads)
```

Compiled with
```sh
scons platform=linuxbsd target=editor production=yes linker=mold lto=none debug_symbols=yes
```

Profiled with:
```sh
valgrind --tool=massif ./godot.linuxbsd.editor.x86_64 -e --path /path/to/empty/godot/project
```


## Results

*Sample 1 (master branch):*
Peak Heap allocation: 574.13380 MiB - 574.17510 MiB

*Sample 2 (this PR):*
Peak Heap allocation: 575.23056 MiB - 575.23404 MiB

*Difference*:
1.05894 MiB - 1.09676 MiB
→ **This PR causes the editor to use approximately 0.18% more heap memory.**

KoBeWi:
I think this is the only way to reliably animate Control nodes inside Container with AnimationPlayer when their size is undetermined. Which is nice, but I think such strict requirement is not that common. There are multiple workarounds and if all fails, you can use Tweens.

That said, I think there is a way to have this feature without much overhead (currently, aside from memory, it also makes transforming more complex). The `render_*` stuff can be moved to a separate struct, which is null by default. Once any render property is set for the first time, allocate the struct. This limits the additional memory usage to single pointer for most Controls and allows to opt-out from the transform operations when the struct is null. Though not sure if doing this is a good idea 🤔

Other than that, since this is purely visual, should these properties be serialized in the scene? Especially with the input problem you mentioned? The new properties should have their purpose documented in Control's class description (i.e. that they are meant only for animation). If the usage is optional like I described above, and it's properly documented, I think it *might* be fine to have this feature.

timoschwarzer:
*Warning: personal opinion and experiences.* :slightly_smiling_face: 


> I think this is the only way to reliably animate Control nodes inside Container with AnimationPlayer when their size is undetermined. Which is nice, but I think such strict requirement is not that common.

I must completely disagree with your statement about this not being a common requirement. In every project so far that involves UI, a feature to apply a render transform to Control nodes of different kinds would have made things much easier to create. And it already helped a lot since I use a fork with this PR applied. Yes, I could have used Tweens for many of the things that I animated, but next to not being able to react to transform changes in parent nodes, Tweens just cannot hold up to the iteration speed and quality that AnimationPlayer gives. Being able to change animation easing curves and previewing them with one click - or even using Bezier curves - is a huge reduction in effort that I need to spend to get the same results in comparison to using Tweens. Looking at the reactions to this PR and the proposal, I don't seem to be alone with this opinion.


> That said, I think there is a way to have this feature without much overhead (currently, aside from memory, it also makes transforming more complex).

0.18% more memory usage in the most UI heavy Godot project that I know of (the Editor) still does not sound like much overhead to me, but a solution like you pointed out sounds like a good compromise, if it's really a concern.


> Also in a scene with 10 levels of nesting, one extra node is irrelevant.

I would agree immediately if Godot had some kind of slot mechanism that lets you define where children should be inserted for instanced scenes. Too often have I been at a point where I got screwed by this to be able to agree here right now. I have had dozens of such cases where this deep nesting of controls prevented modularization of logical Control "components" into separate scenes. I cannot tell immediately if this would be a problem if render transformation is achieved using Containers.

KoBeWi:
> I would agree immediately if Godot had some kind of slot mechanism that lets you define where children should be inserted for instanced scenes.

Soon: #84018

torcado194:
Just to give my two cents: I think this feature is invaluable, to the point that I had already implemented this myself locally (though just for offset, not rotation). I've been using this PR for several weeks and have found no issues with it.
I very much prefer having these properties on Control, regardless of the memory overhead. I think a dedicated node for render offset is a fine alternative, but I am also of the opinion that I would rather reduce the necessary nesting in the scene tree, at least for something as general as a visible offset.

My only feedback is about making the "relative to size" toggles default disabled, to keep them in-line with how transforms work everywhere else (that being in absolute pixels), and because positioning in absolute units is much more common in my experience. I agree that having a rotation pivot at the center is the most common case, but at the very least I don't think that decision should effect the transform offset (and should probably also include similar changes to the basic pivot offset unit).

Mickeon:
For the ones interested in this feature, see and give feedback to the following alternative as well:
- https://github.com/godotengine/godot/pull/98955

timoschwarzer:
@dog-on-moon I agree that the inspector properties should be more intuitive to use, like with transform properties.

> And to the extent of supporting parity, I'm tempted to favor removing "offset_relative_to_size" and "transform_pivot_relative_to_size" as well -- primarily because "offset_relative_to_size" can already be false in most cases, and we already expect developers to manually set the Control's pivot offset to size / 2 in its property documentation for scaling about its center. It's a kind of complex option that I would prefer to be reimplemented via scripting or done manually in most cases.

I'm not so sure about this one. I feel like rotating and scaling around the center is a very common use case for animations (which this PR mainly targets), if not even the most common use case. Having to manually update the pivot every time the control size changes sounds like a lot of boilerplate that I'd rather not want to be needing to do. Dynamically sized controls are quite common, e.g. for Labels or Buttons with translated strings.

It wouldn't be an issue though if there was something that could keep the pivot at a certain relative position of the control.

dog-on-moon:
> I feel like rotating and scaling around the center is a very common use case for animations (which this PR mainly targets), if not even the most common use case. Having to manually update the pivot every time the control size changes sounds like a lot of boilerplate that I'd rather not want to be needing to do for every single control. Dynamically sized controls are quite common, e.g. for Labels or Buttons with translated strings.
> 
> It wouldn't be an issue though if there was something that could keep the pivot at a certain relative position of the control.

Yea.. I do completely agree with this being the most common use case. I think I have nitpicky concerns either way:
1. if we have pivot offset be in absolute pixels, I think it is more intuitive from the inspector and it shares parity with the existing Control field, but I agree with you that it would be a very inconvenient default field for the common use cases.
2. if we have pivot offset be relative to scale, I think it would be a tad less intuitive than absolute pixels, especially right now where the center offset is Vector2(0.5, 0.5).

To try and make point 2 more intuitive, I tried exporting a `Vector2(50, 50)` with the hint string `"suffix:%"`. Maybe this is more tacky? Just trying to come up with a inspector view for it that would help express it to users.

image

After some thinking I believe I am in favor for breaking parity here and keeping the pivot offset as relative to scale. But I could still go either way, I would like for a maintainer to make the call here.

timoschwarzer:
Another thing where we have relative values similar to these here are anchors, but there we don't present the value as percentages in the inspector.
Either way I think these two at least should match up. If the relative pivot offset is presented as a percentage, anchors should, too.

Mickeon:
Regarding the percentage, we should stick to using values between `0.0` and `1.0`. However that is displayed in the Inspector is worth tackling separately. These are not the only properties affected by this.

djrain:
Please please get this in 4.5, it's long overdue. And I agree pivot offset should be relative to scale for sure.

popcar2:
I gave some feedback on the PR in the original proposal thread: https://github.com/godotengine/godot-proposals/issues/7053#issuecomment-3470974058

TL;DR: Overall I think this PR would be perfect if the render transform variables were hidden behind a checkbox rather than a regular group (ie https://github.com/godotengine/godot/pull/105272). I also think there should be a variable in code such as `use_render_transform` that can be used to apply or disable all the render transform variables. This would also be really important in code to quickly tell if something is being offset or not, rather than having to check all the render transform variables individually.

scene/gui/control.cpp
```diff
@@ -4236,6 +4350,14 @@ void Control::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "size_flags_vertical", PROPERTY_HINT_FLAGS, "Fill:1,Expand:2,Shrink Center:4,Shrink End:8"), "set_v_size_flags", "get_v_size_flags");
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "size_flags_stretch_ratio", PROPERTY_HINT_RANGE, "0,20,0.01,or_greater"), "set_stretch_ratio", "get_stretch_ratio");
 
+	ADD_GROUP("Render Transform", "render_");
+	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "render_offset", PROPERTY_HINT_NONE), "set_render_offset", "get_render_offset");
```
AThousandShips:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::VECTOR2, "render_offset"), "set_render_offset", "get_render_offset");
```
As per the old suggestion, `PROPERTY_HINT_NONE` is the default here

paulovbettio:
I've made the changes to the PR based on @popcar2 proposal and other suggestions made here

**Changes**
- Added a bool `enable_render_transform` for toggling render transforms. False as default
- Added PROPERTY_HINT_LINK to the render_scale
- render_offset_relative_to_size is set to false as default (it is confusing to have a relative value as default since the default position does not works this way. I'm also not sure if this is necessary, but it can be useful)


https://github.com/user-attachments/assets/406e7347-28fd-44a5-a028-a07b8a4d4ede

I can make a new pull request with this changes if you don't mind, here is the patch:

[render_transform_toggle.patch](https://github.com/user-attachments/files/23434493/control_transform_renders.patch)

I advocate for (0.5, 0.5) being the default pivot for render transforms, since scaling from the center is much more common than scaling from the sides, afaik

paulovbettio:
What's preventing this from being merged? I'm willing to do the necessary work to make this PR part of the engine. Forgive my enthusiasm, but I'm really hyped about this feature for a long time now :)

Mickeon:
TL:DR: Memory and performance overhead.
Albeit it was tested to be small for today's standards, in the most ideal scenario there should be none at all when the new render properties are not used, which is in the grand majority of cases.
The [proposed **TransformContainer**](https://github.com/godotengine/godot/pull/98955) addresses this problem because:
- it is opt-in (you _have_ to choose where to use it), meaning there's only overhead where necessary
- using a container node is the most fitting for Godot to have this kind of feature, according to some maintainers
	- > A container does seem more like the Godot-way of doing things
	- > Wrapping containers is how SwiftUI does it 

However, as other users have brought up with their own experience, the container workflow becomes quite inconvenient and tedious for anything more complex than a few nodes. Most notably, everything that _may_ be animated in an UI needs to be wrapped inside one, which makes accessing the underlying node(s) very annoying, even in code.
See also a [small recent discussion on RocketChat](https://chat.godotengine.org/channel/new-contributors?msg=StLjMYKLaxNbe9Lnt).

What was roughly suggested to address this, and I'm quoting other maintainers here [from a past discussion](https://chat.godotengine.org/channel/core?msg=vmcKidsLgtSGfWjso):
> The way to solve the memory issue is to store a pointer instead of storing the actual data
> The pattern we use in other places in the engine is to wrap the properties in a struct. Then create an instance of that struct when the object is created. The memory for the struct will be stored on its own and the only overhead in the actual object will be the pointer

> I'd be surprised if it was an inherent property in many UI frameworks. Animatable properties are usually only used temporarily, so creating a permanent performance cost for them isn't great

However, it's really worth noting that memory-wise this may also do not much:
> You would need to profile the code to find out. Memory alignment + size issues are extremely difficult to figure out in advance without profiling

timoschwarzer:
- Rebased
- Added `render_transform_enabled` to toggle whether render transform is applied altogether or not for a `Control` node (thanks @popcar2 !)
- Set default pivot to 0.5/0.5

paulovbettio:
I really don't know how to benchmark things well on the engine, so please correct me if something is wrong with my method.

I've compiled both versions of the engine using the default values (scons -j16)

Godot 4.5.1-stable
Captura de tela de 2025-11-10 00-17-39

Godot 4.6-dev (with render transforms)
Captura de tela de 2025-11-10 00-18-09

The benchmark was done by instancing 512x512 ColorRects, changing its color and adding them as children. After the generation, it updates the label to show the time it took to do all this process. It makes use of threads to create the ColorRects.

I've run this test some times and it does not seem to add a noticeable performance overhead to controls. Maybe there is room for improvement for memory based on the comments of other maintainers.

[control_render_transforms_2025-11-10_00-22-55.zip](https://github.com/user-attachments/files/23445296/control_render_transforms_2025-11-10_00-22-55.zip)

About the TransformContainer PR, I agree that TransformContainers is the most Godot way to do it. However, render transforms are here to fill in the gaps of a much bigger and breaking problem on Godot Controls: [the lack of transitions and animations](https://github.com/godotengine/godot-proposals/issues/8992).

Animations and transitions should be part of Themes. I’m not sure yet how that would be like or implemented, but Themes need motion, one way or another. And while this does not happen, this PR becomes more of a necessity to make animating controls easier, with a better user experience and a less bloated scene tree.

djrain:
So what needs to happen to get this in 4.6? feature freeze is coming up, and this amazing PR is about to be 2 years old. The situation is getting a bit ridiculous :)

GreenCrowDev:
Even if this doesn't get merged in 4.6, it's still a feature that Godot should have at some point.
To me it's not about choosing to implement render transform or either `TransformContainer`, we should have both. I understand that Godot sometimes want to have things done in its opinionated way: nodes. But imho they may serve different purposes, and even work differently as far as I understand.
As @paulovbettio stated, at some point we want to have animations in Theme.

> Animations and transitions should be part of Themes. I’m not sure yet how that would be like or implemented, but Themes need motion, one way or another. And while this does not happen, this PR becomes more of a necessity to make animating controls easier, with a better user experience and a less bloated scene tree.

But in the meantime, render transform is the safest way to implement this (featurewise), considering that it's a something common in other middleware UI frameworks (like NoesisGUI).

paulovbettio:
Feature freeze is here!

It would be nice to have the Godot maintainers to give their verdict on both PRs, we need to close this issue with a good implementation

AdriaandeJongh (commented):
Like @RPicster I'm in favor of this solution over the separate node. This will give developers all the tools to make great UI animations of Controls that are in Containers.

I built & tested this PR as part of the usability team. I haven't looked at the code. The core functionality is generally great! A few notes, of which I do think we need to address the first:

**Relative to Size**
I am not a fan of how the 'Relative to Size' toggles between the two methods of absolute and relative offset right now. If the vector of the offset or pivot is beyond [1.0, 1.0] chances are that the toggle makes the rendered control disappear (because a relative offset of 80 pixels makes it move waaaaaay outside the control's size). to improve the UX of this, I would suggest the following:
1. get rid of the boolean and simply have two vector inputs in the inspector **that are additive**, eg. `render_offset_absolute` and `render_offset_relative` as well as `render_pivot_absolute` and `render_pivot_relative`. this doesn't introduce any states or modes, and thus no 'I forgot to check the box'-kind of situations.
2. keep the vector + bool properties as is, but **convert** the absolute numbers to relative when hitting the toggle and vise versa. this feels like patch on top of the existing system which addresses the most annoying behavior, but the confusion about the differences with the modes remain.

**Naming & documentation**
- I'm pretty strongly against using the words 'Relative to Size'. If we decide to _not_ go with my number 1 suggestion above (because that already clarifies a lot), I'd simply use 'Relative' here, making the full name of the bool 'Offset is Relative', and mention in the docs / tooltip the example that a relative offset of [1.0, 1.0] means that the Control's rendering will be offset'ed one full width and one full height.
- I'm not sure about 'Render Transform' as a category name. I think "Render Offset" is more accessible, even though technically you can adjust more than positional translation... but as a category it applies and it is more friendly / less tech-y.

timoschwarzer:
Thanks for the review!

I'd favor option 2. Option 1 comes with some new factors of ambiguity for me. When I set both an absolute and a relative offset, is the relative offset applied after the absolute offset? If yes and the render scale is set, does it apply the relative offset before or after the render scale etc.?

I agree that the term 'Relative' is enough for the option/checkbox.

I do not care too strongly about the category name, but "offset" for me implies mainly positional, maybe rotational offsets. Edit: maybe "Offset Transform" could work?

When there's a decision or general direction where this PR is going, I'll happily implement the changes necessary for it to get merged.

clayjohn:
Just jumping in here to highlight some of the discussions that have happened between a few of us. 

From what I can tell, the goal is to allow animating the transform + pivot of control nodes to add a bit of juice (which differs from taking absolute control over the transform and ignoring the parent container). 

In Godot 3.x this was possible by animating the rotation, scale, position, or pivot directly. 

In Godot 4.0 we lost that ability. We removed access to the transform + pivot in https://github.com/godotengine/godot/pull/30623 since users were modifying the properties directly in editor, and then losing the modifications when the parent changed layout (due to running the scene, or modifying the parent in editor). It seems that the people involved with this change didn't realize how valuable the existing behaviour was for adding juice and ended up throwing out the baby with the bathwater. 

Now we have two competing PRs to get that behaviour back. Ironically, we can restore the 3.x behaviour with a few lines of code https://github.com/clayjohn/godot/commit/eb0d4459b5736604506acaf30acdfca412cdc1dd (this was provided to me by @BlooRabbit who has been using it in production for awhile). 

That is to say we actually have 3 competing implementations to choose between:

The big factors are:
1. Needs to support animation
2. Performance impact
3. Complexity of implementation
4. Editor UX/workflow
5. Whether input is affected

**1. Introduce a new transform node**

Pros:
- Builds on the concept of parent/child containers (argued by some to be the "Godot way" of doing things)
- Avoids complexity in Control
- Transform survives parent container changes

Cons:
- More complicated workflow (need to add a parent TransformContainer to every Node you want to animate)
- Performance, introduces several transformations on parent re-sort

This approach only impacts rendering, input does not move with the object

**2. Add render_transform properties (this PR)**

Pros:
- Simple UX, simplest workflow
- Transform survives parent container changes

Cons:
- Adds complexity to Control compared to the other approaches (but doesn't add much complexity overall)
- Performance, introduces several transformations each time you edit the transform

This approach impacts rendering, input does not move with the object

**3. animate transform directly (Godot 3.x approach)**

Pros:
- Simplest code change, no added complexity to Control
- Simple UX, simplest workflow
- Lowest performance cost

Cons:
- Affected by parent container layout changes. Modifications will be lost when parent changes

This approach impacts both input and rendering. You are actually moving the control node itself

### Summary

Based on the feedback here and on https://github.com/godotengine/godot/pull/98955 it is clear that this is the preferred approach between the two. I agree that this PR is preferable. 

However, I also see some outstanding disagreement on whether the input area should change in response to the animation. When comparing this PR to the 3.x approach, the biggest differences are in:
1. Whether the animated changes to transform are lost when the parent container changes
2. Whether the input rect is transformed as well

So a final question for those involved. **Which behaviour is more important? Is it more important to have the input rect move along with the rendered rect, or is it more important that the animation changes survive any changes to the parent container?**

_edit: As a final final thought. I believe that we could probably get both behaviours if we wanted to by doing something that is part way in between this PR and the Godot 3.x approach. The important thing for us to understand and agree on is how important it is for input to be affected by the animation transform_

paulovbettio:
> On the other hand "randomly" losing transforms because the parent container did re-compute its layout would break some things. 

It would break a lot of things. Following the GD3 behavior, if we want to update a HBoxContainer of a deck of cards while the player is hovering one of them, the hovering state position would be discarded. And this would happen to a lot of other similar systems.

One of the reasons Godot 3 UI was confusing for me at the start was because the containers were not authoritative/explicit, since it is odd for the Controls to reset their position on sibling update. Godot 4 made containers easy to understand and making this behavior return is a huge step back imo.

torcado194:
Chiming in with agreements with all of the recent responses.  
I definitely do not think re-implementing the previously removed 3.x behavior is desirable, I think it's good that those properties are not editable for many reasons (primarily clarity).

I want to echo the opinions about the input rect not changing. It's absolutely desirable to have a way to change the (visual) position of something without affecting its input rect. I think this behavior is desirable regardless of whether it comes from this PR or otherwise, simply having a way to separate those aspects of a Control node would give a lot of control to the user. It's a bonus that that behavior would be bundled with this feature. Buttons flickering on hover because its rect moves is a very common UI gotcha (in web dev, app dev, anywhere), this would be a nice safety net for a lot of users.

I also want to highlight another piece of the performance viewpoint that's slightly tangential. A common way to offset control nodes is using a MarginContainer. Notably, animating the properties of this node incurs the same relayouting penalties as moving the position normally, which would also be the case if the 3.x behavior was reverted. This was a big bottleneck for me, and one that was not easy to benchmark with the builtin tools. Having a way to change the visual position of a control node is inherently a performance boost because there's no simple method to do so that doesn't cause relayouting.

scene/gui/control.cpp
```diff
@@ -725,6 +733,26 @@ void Control::_update_canvas_item_transform() {
...
+		if (data.render_transform_pivot_relative_to_size) {
+			absolute_render_transform_pivot *= get_size();
+		}
+
+		xform *= Transform2D()
+						 .translated(-absolute_render_transform_pivot)
+						 .rotated(data.render_rotation)
+						 .scaled(data.render_scale)
+						 .translated(absolute_render_transform_pivot)
+						 .translated(absolute_render_offset);
```
kleonc: Rotation/scale have wrong order here (different than everywhere else; try testing it with non-uniform scale, unlikely the behavior is desired):
```suggestion
		xform *= Transform2D()
						 .translated(-absolute_render_transform_pivot)
						 .scaled(data.render_scale)
						 .rotated(data.render_rotation)
						 .translated(absolute_render_transform_pivot)
						 .translated(absolute_render_offset);
```
And this should be equivalent to the above but more performant:
```suggestion
		Transform2D render_xform(data.render_rotation, data.render_scale, 0.0f, absolute_render_transform_pivot + absolute_render_offset);
		render_xform.translate_local(-absolute_render_transform_pivot);
		xform *= render_xform;
```

AdriaandeJongh:
+1 “input should not be affected by the render transform”. it seems like there is consensus on that here, so i think we should continue the conversation on how to improve this PR.

regarding my UX suggestion on the toggle, here’s a more specific example of what i think is desired behavior **to maintain the position of the rendered offset when toggling**. for a control with a size of 400x400, where relative is enabled and the `render_offset` is `0.2,0.2`, upon toggling to disable relative the `render_offset` should change to `80.0,80.0`. upon toggling relative again, `render_offset` converts back to `0.2,0.2`. this would mean no more disappearing of the control’s graphics. a much better experience for the user.

BlooRabbit:
Tbh I do not see how the fact of players not being able to click a button where they see it could NOT be a problem. A button's purpose is to be clicked (and hovered). IMO this issue is not trivial.

In Godot 3 you could just animate any property and it worked fine if you understood what you were doing. Changing this created issues (hence this PR) and, rather than reverting 10 lines of code, we are considering the addition of new layers of complexity creating new problems.

timoschwarzer:
> In Godot 3 you could just animate any property for UI juice (typically rotation and scale) and it worked fine if you understood what you were doing. Changing this created issues (hence this PR) and, rather than reverting 10 lines of code, we are considering the addition of new layers of complexity creating new problems.

That I made this PR because the transform properties on Controls were hidden is *not* true. In fact, I used a very similar change in personal projects with Godot 3. The core point here is that the iteration speed on UI animations was and is very slow, regardless of whether "you know what you are doing" or not. It requires endless amounts of boilerplate code. Creating transitions without or with only a little amount of code while using containers for layout is currently not possible.

> Godot 3 you could just animate any property for UI juice (typically rotation and scale)

In fact, I use render transforms the most for positional offsets (swipe/slide in/out animations), rather than rotation and scale. Rotation and scale in Godot 3 (and 4) are also a bit special since they do not affect the layout step of the parent container and it's currently not very clear to me whether this is considered a bug or not, but it does not seem intended. (See https://github.com/godotengine/godot/issues/19068 and all linked issues within that issue)

markdibarry:
After re-reading all the comments, it sounds like having both the visual and the input lined up is desired as an *option* but majority of actual use-cases want them separate. IMO, this PR is the way forward then, since 3.x's behavior would not satisfy the majority of needs, along with the parent container drawbacks making it a non-starter.

Decisions eventually need made and not everyone will be happy, but I think the most effective solution is to see if any of the performance issues can be mitigated for when this feature is not in use so this isn't seen as a regression to those who don't benefit from it. Then, merging it is a net-positive, and we can explore seeing about having input follow as an option if the demand is there.

kitbdev:
I think the bigger issue with input transform is that currently it doesn't even update the mouse state when a Control moves normally:
- https://github.com/godotengine/godot/issues/87203

So it needs a mouse move or press to update the hover state.
I think there was performance concerns with updating it every time any control moved.

timoschwarzer:
Not sure whether I completely understand the suggestion but if this PR implements a toggle to enable/disable input transformation as well, then it should be strictly better than the Godot 3.x approach because the only difference that I see then would be that this PR is resistant to parent containers messing with the transforms.

BlooRabbit:
Indeed, the new render_transform properties + an option checkbox to make input area follow rendering, would provide a better final result than the more basic Godot 3 approach.  And even if the new properties will probably be confusing for some, they will provide a solution to a substantial problem that has been out there for a long time now.

(one thing we could still do is unlock the ability for users to modify pivot and pivot_ratio directly in the main properties panel, because that's unrelated to preserving parent container layout and you don't change that kind of property by mistake.)

scene/gui/control.cpp
```diff
@@ -3701,6 +3840,23 @@ Control *Control::make_custom_tooltip(const String &p_text) const {
...
 
+void Control::_ensure_allocated_offset_transform() {
+	if (data.offset_transform != nullptr) {
+		return;
+	}
+
+	data.offset_transform = memnew(Data::OffsetTransform);
+}
+
+void Control::_apply_offset_transform(Transform2D &xform) const {
```
AThousandShips:
```suggestion
void Control::_apply_offset_transform(Transform2D &p_xform) const {
```

BlooRabbit:
> Re-reading @AdriaandeJongh's [proposal about relative offset/scale](https://github.com/godotengine/godot/pull/87081#pullrequestreview-3797956199) I now think that option 1 is indeed better.
> 
> ~What do we think about "Visual Transform" as category/feature name?~ Edit: With input transformation that name does not make sense anymore.

Do we really need the pivot offsets? Pivot and pivot_ratio could be set by users in the main property panel. Unless I am mistaken the pivot has no reason to be changed by the parent container. That way we reduce the amount of added properties

doc/classes/Control.xml
```diff
@@ -983,6 +983,10 @@
 		</method>
 	</methods>
 	<members>
+		<member name="absolute_position_offset" type="Vector2" setter="set_absolute_position_offset" getter="get_absolute_position_offset" default="Vector2(0, 0)">
+			Position offset in absolute units. The actual offset is the combined value of this property and [member relative_position_offset].
+			Has no effect unless [member offset_transform_enabled] is true.
```
AThousandShips:
```suggestion
			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
```

doc/classes/Control.xml
```diff
@@ -1098,6 +1102,25 @@
 			Distance between the node's top edge and its parent control, based on [member anchor_top].
 			Offsets are often controlled by one or multiple parent [Container] nodes, so you should not modify them manually if your node is a direct child of a [Container]. Offsets update automatically when you move or resize the node.
 		</member>
+		<member name="offset_transform_absolute_pivot" type="Vector2" setter="set_offset_transform_absolute_pivot" getter="get_offset_transform_absolute_pivot" default="Vector2(0, 0)">
+			Pivot used for [member rotation_offset] and [member scale_offset] in absolute units.
+			The actual pivot position is the combined value of this property and [member offset_transform_relative_pivot].
+			Has no effect unless [member offset_transform_enabled] is true.
```
AThousandShips:
```suggestion
			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
```

doc/classes/Control.xml
```diff
@@ -1098,6 +1102,25 @@
...
+		</member>
+		<member name="offset_transform_enabled" type="bool" setter="set_offset_transform_enabled" getter="is_offset_transform_enabled" default="false">
+			If [code]true[/code], applies all offset transform properties. Otherwise, no offset transform is applied and the properties have no effect.
+		</member>
+		<member name="offset_transform_relative_pivot" type="Vector2" setter="set_offset_transform_relative_pivot" getter="get_offset_transform_relative_pivot" default="Vector2(0.5, 0.5)">
+			Same as [member offset_transform_absolute_pivot] but expressed in units relative to the [Control] [member size] where [code]Vector2(0, 0)[/code] is the top-left corner of this control, and [code]Vector2(1, 1)[/code] is its bottom-right corner.
+			The actual pivot position is the combined value of this property and [member offset_transform_absolute_pivot].
+			Has no effect unless [member offset_transform_enabled] is true.
```
AThousandShips:
```suggestion
			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
```

doc/classes/Control.xml
```diff
@@ -1098,6 +1102,25 @@
...
+		</member>
+		<member name="offset_transform_visual_only" type="bool" setter="set_offset_transform_visual_only" getter="is_offset_transform_visual_only" default="true">
+			If [code]true[/code], applies all offset transforms only visually without affecting input. Buttons for example would still be pressable at their
+			location before offset transforms are applied.
+			If [code]false[/code], the entire transform of this Control is affected so inputs will be registered at where the Control is visually.
+			Has no effect unless [member offset_transform_enabled] is true.
```
AThousandShips:
```suggestion
			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
```

doc/classes/Control.xml
```diff
@@ -1110,19 +1133,32 @@
 		<member name="position" type="Vector2" setter="_set_position" getter="get_position" default="Vector2(0, 0)">
 			The node's position, relative to its containing node. It corresponds to the rectangle's top-left corner. The property is not affected by [member pivot_offset].
 		</member>
+		<member name="relative_position_offset" type="Vector2" setter="set_relative_position_offset" getter="get_relative_position_offset" default="Vector2(0, 0)">
+			Same as [member absolute_position_offset] but expressed in units relative to the [Control] [member size] where [code]Vector2(0, 0)[/code] is the top-left corner of this control, and [code]Vector2(1, 1)[/code] is its bottom-right corner.
+			The actual offset is the combined value of this property and [member absolute_position_offset].
+			Has no effect unless [member offset_transform_enabled] is true.
```
AThousandShips:
```suggestion
			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
```

doc/classes/Control.xml
```diff
@@ -1110,19 +1133,32 @@
...
 		<member name="rotation" type="float" setter="set_rotation" getter="get_rotation" default="0.0">
 			The node's rotation around its pivot, in radians. See [member pivot_offset] to change the pivot's position.
 			[b]Note:[/b] This property is edited in the inspector in degrees. If you want to use degrees in a script, use [member rotation_degrees].
 		</member>
 		<member name="rotation_degrees" type="float" setter="set_rotation_degrees" getter="get_rotation_degrees">
 			Helper property to access [member rotation] in degrees instead of radians.
 		</member>
+		<member name="rotation_offset" type="float" setter="set_rotation_offset" getter="get_rotation_offset" default="0.0">
+			Rotation offset. The rotation pivot is defined by [member offset_transform_absolute_pivot] and [member offset_transform_relative_pivot].
+			Has no effect unless [member offset_transform_enabled] is true.
```
AThousandShips:
```suggestion
			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
```

doc/classes/Control.xml
```diff
@@ -1110,19 +1133,32 @@
...
 			[b]Note:[/b] If the Control node is a child of a [Container] node, the scale will be reset to [code]Vector2(1, 1)[/code] when the scene is instantiated. To set the Control's scale when it's instantiated, wait for one frame using [code]await get_tree().process_frame[/code] then set its [member scale] property.
 		</member>
+		<member name="scale_offset" type="Vector2" setter="set_scale_offset" getter="get_scale_offset" default="Vector2(1, 1)">
+			Scale offset. The scale pivot is defined by [member offset_transform_absolute_pivot] and [member offset_transform_relative_pivot].
+			Has no effect unless [member offset_transform_enabled] is true.
```
AThousandShips:
```suggestion
			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
```

doc/classes/Control.xml
```diff
@@ -1098,6 +1102,25 @@
...
+			If [code]true[/code], applies all offset transform properties. Otherwise, no offset transform is applied and the properties have no effect.
+		</member>
+		<member name="offset_transform_relative_pivot" type="Vector2" setter="set_offset_transform_relative_pivot" getter="get_offset_transform_relative_pivot" default="Vector2(0.5, 0.5)">
+			Same as [member offset_transform_absolute_pivot] but expressed in units relative to the [Control] [member size] where [code]Vector2(0, 0)[/code] is the top-left corner of this control, and [code]Vector2(1, 1)[/code] is its bottom-right corner.
+			The actual pivot position is the combined value of this property and [member offset_transform_absolute_pivot].
+			Has no effect unless [member offset_transform_enabled] is true.
+		</member>
+		<member name="offset_transform_visual_only" type="bool" setter="set_offset_transform_visual_only" getter="is_offset_transform_visual_only" default="true">
+			If [code]true[/code], applies all offset transforms only visually without affecting input. Buttons for example would still be pressable at their
+			location before offset transforms are applied.
```
AThousandShips:
```suggestion
			If [code]true[/code], only applies offset transforms visually without affecting input. Buttons for example would still be pressable at their locations before offset transforms are applied.
```

doc/classes/Control.xml
```diff
@@ -1098,6 +1102,25 @@
...
+			If [code]false[/code], the entire transform of this Control is affected so inputs will be registered at where the Control is visually.
```
AThousandShips:
```suggestion
			If [code]false[/code], the entire transform of this Control is affected so inputs will be registered where the Control is visually.
```
I'd say

scene/gui/control.h
```diff
@@ -362,6 +383,9 @@ class Control : public CanvasItem {
 	void _notify_theme_override_changed();
 	void _invalidate_theme_cache();
 
+	void _ensure_allocated_offset_transform();
+	void _apply_offset_transform(Transform2D &xform) const;
```
AThousandShips:
```suggestion
	void _apply_offset_transform(Transform2D &p_xform) const;
```

scene/gui/control.cpp
AThousandShips:
```suggestion
void Control::_apply_offset_transform(Transform2D &p_xform) const {
```

scene/gui/control.cpp
```diff
@@ -1881,6 +1892,134 @@ real_t Control::get_stretch_ratio() const {
...
+	if (p_enabled) {
+		_ensure_allocated_offset_transform();
+		data.offset_transform->enabled = true;
+	} else {
+		// Never deallocate to not lose previously set values
+		data.offset_transform->enabled = false;
+	}
+
+	queue_redraw();
+	_notify_transform();
```
bruvzg, line 1911: Since return value of `get_internal_transform()` is changing, it (and other setters) should also call `queue_accessibility_update();`.


```suggestion
	_notify_transform();
	queue_accessibility_update();
```

scene/gui/control.cpp
```diff
@@ -1881,6 +1892,134 @@ real_t Control::get_stretch_ratio() const {
...
+}
+
+bool Control::is_offset_transform_enabled() const {
+	return data.offset_transform != nullptr && data.offset_transform->enabled;
+}
+
+void Control::set_absolute_position_offset(const Vector2 &p_offset) {
+	_ensure_allocated_offset_transform();
+	data.offset_transform->translation_absolute = p_offset;
```
bruvzg, line 1925: Any reason why it (and all other properties) is calling `queue_redraw` when noting changed? Probably should have early return case:


```suggestion
	if (data.offset_transform->translation_absolute == p_offset) {
		return;
	}
	data.offset_transform->translation_absolute = p_offset;
```

timoschwarzer: Implemented this by comparing to the getters, which prevents allocation when you set it to a default value.

timoschwarzer:
What if I want to make a container that aligns children in an arc like this? (excuse my paint drawing)

image

Then the children control rotation and pivot are managed by the parent container. I know it's a niche use case but I want to make the difference between layout transform and render/offset transform clear. One is managed by the parent container for putting the control where *the container wants it to be* and the render/offset transform is applied on top for e.g. animations that transition to or from that target transform and is completely isolated from the container.

If your suggestion is to opaquely map the layout transform properties (the ones that existed before this PR) to the offset transform properties when a control is inside a container then I'm very much against that. Not only messes this up the Layout/Offset Transform distinction further but also makes it more confusing for users what propertly they are currently changing. What happens when I change one of these properties while the control is inside a container and then move it out of a container? Are the offset properties stuck at their current values then and I need to add it to a container again for me to be able to set them?

Nolkaloid:
Okay I see your point. Makes sense!

In that case, If I could suggest something for the render transforms: I think it would be nice to have the transformed rect drawn in addition to the layout rect, perhaps in a light orange color or something like that.
I think it would add a bit of clarity.

dog-on-moon:
> perhaps you can make a case for the use of this offset transform outside of a container / the layout system? because I don’t see any reason not to prefer simply using the control system and transforms over this offset system in those cases.

Offset transforms are extremely useful for relative movement to an existing layout, and not all layouts depend on a container parent.

1. A custom textbox rendered in 3d space. The MarginContainer uses render transform as a relative offset from the origin. The parent of the MarginContainer is a custom zero-size control that calculates an onscreen position from a Node3D. Setting the position of the MarginContainer would be more difficult, as its position is dynamic based on the size of the text label.

https://github.com/user-attachments/assets/1dad7811-cc58-4601-b7de-1fe7ac53622e

2. A scene where I'm using Controls a bit more artistically, and not necessarily constraining them to a rigid layout -- but still taking advantage of panels and text labels.

godot windows editor x86_64_UlrPIoMqpp

3. The root node of a Control scene is being moved here as a spawn effect.

https://github.com/user-attachments/assets/e8fddca2-a770-4a93-93ad-94cab32ef2f0

My team has been using this PR since November 2024. Searching our codebase, the term "render_offset" appears 46 times throughout different scripts, and 114 times throughout different scenes, so I'm confident I could find more cases if need be.

doc/classes/Control.xml
```diff
@@ -1098,6 +1102,24 @@
...
+			Pivot used for [member rotation_offset] and [member scale_offset] in absolute units.
+			The actual pivot position is the combined value of this property and [member offset_transform_relative_pivot].
+			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
+		</member>
+		<member name="offset_transform_enabled" type="bool" setter="set_offset_transform_enabled" getter="is_offset_transform_enabled" default="false">
+			If [code]true[/code], applies all offset transform properties. Otherwise, no offset transform is applied and the properties have no effect.
+		</member>
+		<member name="offset_transform_relative_pivot" type="Vector2" setter="set_offset_transform_relative_pivot" getter="get_offset_transform_relative_pivot" default="Vector2(0.5, 0.5)">
+			Same as [member offset_transform_absolute_pivot] but expressed in units relative to the [Control] [member size] where [code]Vector2(0, 0)[/code] is the top-left corner of this control, and [code]Vector2(1, 1)[/code] is its bottom-right corner.
+			The actual pivot position is the combined value of this property and [member offset_transform_absolute_pivot].
```
AdriaandeJongh:
```suggestion
			The final pivot position is the combined value of this property and [member offset_transform_absolute_pivot].
```

doc/classes/Control.xml
```diff
@@ -1098,6 +1102,24 @@
 			Distance between the node's top edge and its parent control, based on [member anchor_top].
 			Offsets are often controlled by one or multiple parent [Container] nodes, so you should not modify them manually if your node is a direct child of a [Container]. Offsets update automatically when you move or resize the node.
 		</member>
+		<member name="offset_transform_absolute_pivot" type="Vector2" setter="set_offset_transform_absolute_pivot" getter="get_offset_transform_absolute_pivot" default="Vector2(0, 0)">
+			Pivot used for [member rotation_offset] and [member scale_offset] in absolute units.
+			The actual pivot position is the combined value of this property and [member offset_transform_relative_pivot].
```
AdriaandeJongh:
```suggestion
			The final pivot position is the combined value of this property and [member offset_transform_relative_pivot].
```

doc/classes/Control.xml
```diff
@@ -983,6 +983,10 @@
 		</method>
 	</methods>
 	<members>
+		<member name="absolute_position_offset" type="Vector2" setter="set_absolute_position_offset" getter="get_absolute_position_offset" default="Vector2(0, 0)">
+			Position offset in absolute units. The actual offset is the combined value of this property and [member relative_position_offset].
```
AdriaandeJongh:
```suggestion
			Position offset in absolute units. The final offset is the combined value of this property and [member relative_position_offset].
```

doc/classes/Control.xml
```diff
@@ -1098,6 +1102,24 @@
 			Distance between the node's top edge and its parent control, based on [member anchor_top].
 			Offsets are often controlled by one or multiple parent [Container] nodes, so you should not modify them manually if your node is a direct child of a [Container]. Offsets update automatically when you move or resize the node.
 		</member>
+		<member name="offset_transform_absolute_pivot" type="Vector2" setter="set_offset_transform_absolute_pivot" getter="get_offset_transform_absolute_pivot" default="Vector2(0, 0)">
+			Pivot used for [member rotation_offset] and [member scale_offset] in absolute units.
```
AdriaandeJongh:
```suggestion
			Pivot used by [member rotation_offset] and [member scale_offset] in absolute units.
```

doc/classes/Control.xml
```diff
@@ -1098,6 +1102,24 @@
...
+			If [code]true[/code], applies all offset transform properties. Otherwise, no offset transform is applied and the properties have no effect.
+		</member>
+		<member name="offset_transform_relative_pivot" type="Vector2" setter="set_offset_transform_relative_pivot" getter="get_offset_transform_relative_pivot" default="Vector2(0.5, 0.5)">
+			Same as [member offset_transform_absolute_pivot] but expressed in units relative to the [Control] [member size] where [code]Vector2(0, 0)[/code] is the top-left corner of this control, and [code]Vector2(1, 1)[/code] is its bottom-right corner.
+			The actual pivot position is the combined value of this property and [member offset_transform_absolute_pivot].
+			Has no effect unless [member offset_transform_enabled] is [code]true[/code].
+		</member>
+		<member name="offset_transform_visual_only" type="bool" setter="set_offset_transform_visual_only" getter="is_offset_transform_visual_only" default="true">
+			If [code]true[/code], only applies offset transforms visually without affecting input. Buttons for example would still be pressable at their locations before offset transforms are applied.
```
AdriaandeJongh:
```suggestion
			If [code]true[/code], the offset transforms is only applied visually and does not affect input. In other words, this Control will still receive input events at its original location before the offset transform is applied.
```

doc/classes/Control.xml
```diff
@@ -1098,6 +1102,24 @@
...
+			If [code]false[/code], the entire transform of this Control is affected so inputs will be registered where the Control is visually.
```
AdriaandeJongh:
```suggestion
			If [code]false[/code], the entire transform of this Control is affected and input events will register where the Control is visually.
```

doc/classes/Control.xml
```diff
@@ -1110,19 +1132,32 @@
 		<member name="position" type="Vector2" setter="_set_position" getter="get_position" default="Vector2(0, 0)">
 			The node's position, relative to its containing node. It corresponds to the rectangle's top-left corner. The property is not affected by [member pivot_offset].
 		</member>
+		<member name="relative_position_offset" type="Vector2" setter="set_relative_position_offset" getter="get_relative_position_offset" default="Vector2(0, 0)">
+			Same as [member absolute_position_offset] but expressed in units relative to the [Control] [member size] where [code]Vector2(0, 0)[/code] is the top-left corner of this control, and [code]Vector2(1, 1)[/code] is its bottom-right corner.
+			The actual offset is the combined value of this property and [member absolute_position_offset].
```
AdriaandeJongh:
```suggestion
			The final offset is the combined value of this property and [member absolute_position_offset].
```

AdriaandeJongh (commented):
Bunch of small nitpicks on the docs.

AdriaandeJongh (approved):
1. I didn't review the code.
 
2. Why isn't there a size offset, actually? A size property would put it on par with the features of the regular Transform. It would be in absolute units and relative to the original size, and would work side by side with `Scale Offset` which is kind of its sister property using relative sizing instead of absolute.

3. If `offset_transform_visual_only` is `false`, the orange hitbox of the Control moves to the offsetted location. I think that's expected, as that's where the input transform also moves. But it would still be useful to know where its original transform is located because the properties added by this PR are relative to that and to be able to see them gives the user a point of reference. Perhaps a gray-ish bounding box for the transform untouched by the offset transform? Like so:

Image

4. I think the properties would be best named using the template 'main characteristic + Offset + variation' so that it's easier to read the list of properties at first glance. For example: `Position Offset Absolute`, `Position Offset Relative`, `Scale Offset` (current), `Rotation Offset` (current), `Pivot Offset Absolute`, and `Pivot Offset Relative`. However, we can't ignore that the original transform already has two properties named `Pivot Offset` and `Pivot Offset Ratio` making all the potential naming options a bit limited and suboptimal, but... I'd still like to propose renaming (and ordering) the properties to the following:
  - `Position Offset`
  - `Position Offset Ratio`
  - `Size Offset`
  - `Scale Offset`
  - `Rotation Offset`
  - `Pivot Transform Offset`
  - `Pivot Transform Offset Ratio`
  - `Offset Visual Transform Only`

Other than that, I played around with it more, tried a few dumb animations like this one, and look forward to using this in one of our future projects :)

https://github.com/user-attachments/assets/78ab079d-87c1-42a4-952d-26922947a632

scene/gui/control.cpp
```diff
@@ -1882,6 +1892,170 @@ real_t Control::get_stretch_ratio() const {
...
+void Control::set_absolute_position_offset(const Vector2 &p_offset) {
+	if (get_absolute_position_offset() == p_offset) {
+		return;
+	}
+
+	_ensure_allocated_offset_transform();
+	data.offset_transform->translation_absolute = p_offset;
+	queue_redraw();
+	_notify_transform();
+	queue_accessibility_update();
```
KoBeWi, line 1933: This should only happen when offset transform is enabled.

scene/gui/control.h
```diff
@@ -183,6 +183,25 @@ class Control : public CanvasItem {
 
 	// This Data struct is to avoid namespace pollution in derived classes.
 	struct Data {
+		struct OffsetTransform {
+			static constexpr Vector2 DEFAULT_TRANSLATION_ABSOLUTE = Vector2();
+			static constexpr Vector2 DEFAULT_TRANSLATION_RELATIVE = Vector2();
```
KoBeWi, line 188: Vector2 does not need explicit value if it's zero.
```suggestion
			static constexpr Vector2 DEFAULT_TRANSLATION_ABSOLUTE;
			static constexpr Vector2 DEFAULT_TRANSLATION_RELATIVE;
```

KoBeWi (approved):
The code looks fine.

felinusfish:
So excited to see this! Rendering offsets is awesome, definitely will make animation so much easier. Thank you so much for making this PR!!

bruvzg (approved):
Code looks good.

scene/gui/control.h
timoschwarzer: Compiler disagrees: `error: 'constexpr' static data member 'DEFAULT_TRANSLATION_ABSOLUTE' must have an initializer`

KoBeWi: Huh.

Cautami:
~~On the bindings side, are the following method names intentional? 
`IsOffsetTransformEnabled`
`IsOffsetTransformVisualOnly`
I primarily work with C# and the naming immediately stuck out to me, the set methods are appropriately named `SetBlahBlah` so I was wondering.~~

Ignore me, it seems "Is" is apparently used in other places, such as `IsVisible` I'm not sure how I didn't notice this.

AdriaandeJongh (commented):
Tested it again and everything looks good.

I will admit that leaving 'Visuals Only' enabled for a parent that offsets itself and its children leads to very confusing bounding boxes, both the grey and the orange ones. Example:

https://github.com/user-attachments/assets/88e3780d-879d-4d2a-b69b-7ec38e5da998

It makes me wonder whether this is a bug, actually? Shouldn't the bounding boxes (both the grey and the orange) already incorporate the offsets from their parents?

Edit: I realized it's not a bug because input will still be inside the orange rectangle of course, regardless of the offsets of the parent. So... if people make these inheriting offsets, it'll be confusion all around.

AdriaandeJongh (approved):
Tested this in depth and found it to work as expected. The latest additions of the grey bounding box and the updated property names look good.

BlooRabbit:
> > they are not controlled or reset by the parent container anyway
> 
> The fact that they are not reset sounds like a bug to me. Other properties, like rotation or scale, are reset, even if the container does not use them.

Not sure why the parent container should set or reset a pivot of its children? Setting the pivot of the child node simply means that you consider this node should rotate or scale (for example) around that pivot. Why would that be controlled by a parent container, which only aims at organizing the child nodes in a certain way (horizontally, vertically, etc) ? And that's normal - not all properties are controlled  by the parent, thankfully (e.g. visibility).  Or am I missing something?

kleonc:
In the current form what's lacking is the way to obtain the resulting `offset_transform` Transform2D. The user should be able to obtain it without checking out the source code and manually combining the `offset_transform_*` properties in the same way.

However `get_offset_transform() -> Transform2D` would still be kinda problematic given there's `offset_transform_visual_only` toggle, because such toggle makes branching necessary in the user code.
E.g. let's say the user wants to calculate global[^1] visual position of some local point, then it would need to be done like:
```gdscript
var control: Control = ...
var point_local: Vector2 = ...

var visual_local_to_global: Transform2D = control.get_global_transform()
# if `offset_transform_visual_only == false` then `offset_transform` is already incorporated into `get_global_transform()`
if control.offset_transform_visual_only:
	visual_local_to_global *= control.get_offset_transform()

var point_global: Vector2 = visual_local_to_global * point_local
```

And yeah, just writing the snippet above I can already see it will be _super easy_ to explain to the users the difference between the "local" and "visual local" coordinates (not sure what's the exact nomenclature we should use here). But relevant 2D transforms related manuals in the docs should be checked/updated to account for this PR. E.g.:
- https://github.com/godotengine/godot-docs/blob/3aea0830480af422b5dab3b2d9762980cb9f505b/tutorials/2d/canvas_layers.rst
- https://github.com/godotengine/godot-docs/blob/3aea0830480af422b5dab3b2d9762980cb9f505b/tutorials/2d/2d_transforms.rst
- https://github.com/godotengine/godot-docs/blob/3aea0830480af422b5dab3b2d9762980cb9f505b/tutorials/2d/custom_drawing_in_2d.rst
- https://github.com/godotengine/godot-docs/blob/3aea0830480af422b5dab3b2d9762980cb9f505b/engine_details/architecture/2d_coordinate_systems.rst

Also existing Control's docs might need updating, e.g. for `get_transform()`.

[^1]: Aka in the containing canvas layer's coordiante system.

gnat:
Curious @timoschwarzer, how reasonable would it be to get this on all `Node2D`? Seems like it would solve the 10+ year issue of AnimationPlayer not supporting relative transforms https://github.com/godotengine/godot-proposals/issues/1191#issuecomment-4559836307

PS: Transform Offset is my favorite addition to 4.7, thanks for all of your hard work!

timoschwarzer:
@gnat I think a proper implementation of relative animations would be better. Implementing relative transforms for all Node2Ds would only fix the issue for transforms, but what if I want to animate something else than a transform relatively?

gnat:
@timoschwarzer IMHO it may be worth keeping an open mind about it because:

* Offset/relative transforms are definitely the main pain point of using `AnimationPlayer`
* There has been zero actionable plan or direction on it forever. (@Calinou is of the opinion that dedicated relative animation will never happen because it's too complicated)
* I think it'd be very convenient to have these on all `Node2D` anyway. Why not have both?

Thankfully `Control` no longer suffers from the issue, which is nice.

PR #87084, opened 2024-01-11
[Shaders] Allow trailing comma in initializer lists

AThousandShips:
GLSL allows trailing comma in array constructors, and so does many languages, this brings Godot shaders to parity with that

I believe I have caught every case, but might have missed some, as this is quite complex code, it works for both declarations in global, function, and argument scope, so believe I've caught everything

Edit: To clarify this is the curly braces style initialization by initializer list, as supported from version 4.2, not the parenthesis form, i.e. `float my_arr[3] = { 1.0, 2.0, 3.0, };` not `float my_array[3] = float[3](1.0, 2.0, 3.0);`

### DIFF
--- a/servers/rendering/shader_language.cpp
+++ b/servers/rendering/shader_language.cpp
@@ -4876,6 +4876,15 @@ ShaderLanguage::Node *ShaderLanguage::_parse_array_constructor(BlockNode *p_bloc
 			tk = _get_token();
 			if (tk.type == TK_COMMA) {
 				an->initializer.push_back(n);
+
+				if (auto_size) {
+					TkPos prev_pos = _get_tkpos();
+					tk = _get_token();
+					if (tk.type == TK_CURLY_BRACKET_CLOSE) {
+						break;
+					}
+					_set_tkpos(prev_pos);
+				}
 			} else if (!auto_size && tk.type == TK_PARENTHESIS_CLOSE) {
 				an->initializer.push_back(n);
 				break;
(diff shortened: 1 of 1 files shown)


akx:
By my reading of the GLSL grammar (transliterated from the [spec PDF](https://registry.khronos.org/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf) into text [here](https://gist.github.com/akx/57ec0d4a2bf4a5d3d37775710e3d8421)), array initializers _shouldn't_ allow a trailing comma, but maybe I've read things wrong.

AThousandShips:
The parser we are using, glslang, does, I've confirmed it in the parsed code

akx:
> The parser we are using, glslang

Maybe `ShaderLanguage` should use glslang's parser all in all, then?

> this is GLSL 4.2 stuff

Good to know. Should the documentation for the shading language be updated to reflect this? It's saying it's based on GLSL ES 3.0.

https://github.com/godotengine/godot-docs/blob/f6d53d141edad708f2106f77723895c66c65b674/tutorials/shaders/shader_reference/shading_language.rst#L9-L11

AThousandShips:
It would need to, initializer style arrays are not, AFAIK, supported in GLSL ES 3.0

clayjohn:
> It would need to, initializer style arrays are not, AFAIK, supported in GLSL ES 3.0
> 
> > Maybe ShaderLanguage should use glslang's parser all in all, then?
> 
> Not sure how feasible that would be, it's GLSL specific, also glslang is only used when not using the compatibility backend, so it'd add an unnecessary dependency, might be possible to leverage some of that code though

Keep in mind the compatibility backend is GL ES 3.0 based. So we mostly limit our syntax to what is supported by GLSL ES 300 / GLSL 330.

AThousandShips:
Yeah it seems the initializer syntax is the one major divergence from that

AThousandShips:
If it works in the compatibility backend to use initializer lists then it doesn't pass it like that

But the discussions aren't moot by that reason though

clayjohn:
To update on this. From my perspective, we shouldn't be cherrypicking advanced features from newer versions of OpenGL without good reason. The shader compiler is already super complex and IMO this kind of change adds complexity without providing a real benefit to users. 

Further, without adding significant compatibility code ourselves, we risk breaking the engine on older devices where this sort of behaviour is not guaranteed by the spec and whether it works is up to the driver

AThousandShips:
Closing based on the above for now, was mainly a change to improve compatibility with external code as it's listed as supported in the relevant versions of the language, but it's just a minor convenience 

I'll reopen it if desired in the future

PR #87085, opened 2024-01-11
Refactor editor theme generation and add spacing presets

YuriSizov:
_I'm going to invalidate so many PRs with this one..._

This is a complete rework of the editor theme generation logic. For the longest time our theme generator has been an unmaintainable spaghetti mess with a lot of cross-borrowing between various definitions without a clear structure.

This naturally made maintaining generated editor themes much harder than it needs to be. We also didn't have a good picture of which parameters can be exposed to users and what effect such exposure would have. So in this mega rework I try to bring some order into this.

----

This PR contains 3 commits. 2 of them are about organization and cleanup, and the last one introduces a long awaited feature (but in its first iteration, with more work expected in future).

## Reorganize code related to editor theming

* This change introduces a new `EditorThemeManager` class to abstract theme generation and its subroutines.
* Logic related to `EditorTheme`, `EditorColorMap`, and editor icons has been extracted into their respective files with includes cleaned up.
* All related files have been moved to a separate folder to better scope them in the project. This includes relevant generated files as well.

This commit acts as a starting point for the changes that follow it. It's pretty extensive in the way it touches across the codebase, and it introduces new files and build system boilerplate changes. Pretty much any PR that is currently touching the old `editor_themes.h/cpp` needs to be rebased because of this. But such is the price of a better code structure.

Some of the changes here are made with further plans to optimize theme generation and make it more granular and avoid regenerating themes in their entirety if possible. I also plan to base editor theme caching on this, which should speed up editor startup by about half a second (on my laptop, at least).

## Split theme generation logic into several subroutines

* This change splits the theme generation process into more manageable subroutines, encapsulating one part of the logic.
* To better control how parameters and shared definitions are reused across the theme, a new theme configuration struct has been added.
* Everything not passed and not explicitly shared — is scoped. This is done to prevent it from being automatically accessible throughout the process. This should ensure that the decision to share styles is a conscious one.

In the future we should try to reduce the number of unique definitions and share most of what we have. Simplicity should make it easier for us to define guidelines for existing and future editor UI/UX features. This PR is a stepping stone on this path.

This change also includes some effort separating redefinitions of default theme items and custom types introduced only by the editor. In a few cases where editor-specific definitions need to reference default definitions we simply fetch them from the theme. It's not ideal and hides the dependency a bit, but hopefully these cases will be abstracted properly in due time.

This is another change that invalidates existing PRs targeting `editor_themes.h/cpp`. 

## Introduce editor theme spacing presets

This is the eye candy of the PR, although, to be completely honest, the work only begins here.

The third commit adds new editor settings to give users more control over compactness/spacing of the editor theme. It comes with 2 numerical properties and 3 presets (Compact, Default, Spacious). Numerical properties can be completely customized to achieve a more desirable look if presets aren't satisfactory.

Limited effort was applied to make sure both Compact and Spacious presets work and look fine, but further tuning
and adjustments are totally expected. Some controls will require layout changes or more granular fixes to their constants to make the most of it.

This change supersedes https://github.com/godotengine/godot/pull/86363.
*Bugsquad edit:* Implements https://github.com/godotengine/godot-proposals/issues/8505.

----

Note that adjusting spacing alone may not be enough, and changing fonts and font sizes is another thing you want to do for a better compact/spacious experience.

For the rough comparison, here's how these presets look compared to the current master:

### Current master

godot windows editor dev x86_64_2024-01-11_18-49-08

### This PR — Default

godot windows editor dev x86_64_2024-01-11_18-12-18

### This PR — Compact

godot windows editor dev x86_64_2024-01-11_18-11-48

### This PR — Spacious

godot windows editor dev x86_64_2024-01-11_18-12-49


YuriSizov:
Notes for testing:

* The first and the second commit should not result in any discernible visual changes. I would suggest you check the new code as if it was new, because comparing to the old version would be quite a waste of time. You'd have to trust me that nothing was lost in the translation.

* Currently when you change theme settings and the theme is regenerated, there are errors in the output log related to benchmarks. This is a problem in master, and I will fix it in a follow up.

passivestar:
Awesome work!

2 pieces of feedback:

- Compact trees items have the same spacing as Default, are you planning to decrease that?

- Very glad to see you got rid of overall margins in docks in Compact because it makes the contents of docks _visually_ separated by a single line instead of 3 lines. But some content needs those margins fixed locally now, e.g the scene tree visibility icons now have 0px separation on the right, which means trees need their _own_ margins to fix that. Is that something you're planning to address in the Compact mode?

akien-mga:
> Is this really necessary? 145 files changed with just that line.

It isn't strictly speaking necessary to allow configuring a more compact spacing look. But it's necessary to improve the quality of the current editor theme code, which makes it very difficult to allow the level of customization that users keep asking for. See the OP for a thorough description of the purpose of the first two commits. It's a lot of changes, but it's something that will significantly increase the maintainability and flexibility of the theme presets.

As mentioned in the OP, it also paves the way for future work on optimization, caching, and likely more customization possibilities (both for presets / via editor settings, and for actual editor themes that can be shared with the community).

It's something that has been discussed among maintainers for a long time (since before 4.0), and Yuri now had the time to finally do this much needed refactoring for 4.3. We're in good time in the dev cycle to allow doing both this refactoring, and then adding new features on top like a Compact spacing preset.

> Isn't this pr #86363 much simpler and solves the fundamental problem better?

It's much simpler for sure, but as discussed there, we don't think that it solves the fundamental problem better. We don't want users to have to toggle 3 individual settings to achieve the highly requested margin-less look. And there would likely end up being more in the future, as #86363 is just a first step but there's a lot more changes needed to really allow making things compact and user-friendly (fonts, inner margins, icon spacing, etc.).

For users interested in how this addresses https://github.com/godotengine/godot-proposals/issues/8505, you should only look at the third commit, the first two are internal refactoring for maintainability, scalability, and future improvements.

coppolaemilio:
> Isn't this pr #86363 much simpler and solves the fundamental problem better?

While the other solution might be simpler, this one offers a better default for everyone, and a possibility for us to make a properly spaced custom compact mode. The idea behind making changes here is decreasing the gap between all the elements (which most people already agree are too big). The other PR was able to fix it with an input box that will very likely only serve the purpose of being changed to the minimum available.

editor/themes/editor_theme_manager.h
```diff
@@ -31,11 +31,111 @@
...
+		Ref<StyleBoxFlat> content_panel_style;
+		Ref<StyleBoxFlat> tree_panel_style;
+
+		Vector2 widget_margin;
+	};
 
 	static Ref<Theme> _create_base_theme(const Ref<Theme> &p_old_theme = nullptr);
+	static ThemeConfiguration _create_theme_config(const Ref<Theme> &p_theme);
+
+	static void _create_shared_styles(const Ref<Theme> &p_theme, ThemeConfiguration &p_config);
```
akien-mga, line 135: Would it make sense to use `const ThemeConfiguration &`?

akien-mga (commented):
So far I've reviewed the two refactoring commits, they look good to me. I agree it makes sense to group all editor theme-related code in a subfolder, add an EditorThemeManager to tie it all up, and make it logic clearer with subroutines.

My only nitpick is whether the plural is justified in `editor/themes/` or if it should just be `editor/theme/` (like `scene/theme/`). The plural makes me think I'll find separate files for different themes, which doesn't seem to be the case, instead it's theme (singular) related classes (one of which, the Manager, defines multiple theme presets, plural). Buuut that's super nitpicky and I don't feel strongly about this. You maintain this code, you decide :)

---

For the actual new presets being added, I haven't reviewed yet. I'll do some testing and share my feedback, but I think others will also have more qualified input on the UX than I do.

I'll just say I fully support adding these presets, and changing the default spacing, as there's a very high demand for it, and there seems to be consensus in the community that it would be a better default.

YuriSizov:
@passivestar 

> Awesome work!

Thanks! 💙

> Compact trees items have the same spacing as Default, are you planning to decrease that?

If you mean vertical space, which you mentioned in your proposal, then yes, this needs to be looked into. If you mean horizontal space, I'm not sure how much it can be reduced. Our default spacing already barely leaves any space for the relationship lines and the collapse button.

> Some content needs those margins fixed locally now, e.g the scene tree visibility icons now have 0px separation on the right, which means trees need their _own_ margins to fix that. Is that something you're planning to address in the Compact mode?

Yep, this is exactly what I meant by further work. I did the first pass to fix some issues caused by reduced margins, but only barely. (In other words, if you take the current master and change the old default margin size property, you won't get the same result. But I definitely do not consider the current state as final and fully polished, I just want to get this thing out first.)

Currently everything is driven by 2 properties, for base and additional spacing, but they aren't always used in a consistent manner (or at all, some values are just hardcoded). This needs to be addressed with a careful combing through the definitions and experimentations.

I hope we can keep everything simple and control all the compactness aspects with the same two properties through clever relations between them and theme definitions, but we will have to see if it's possible. The goal is to make sure that "Compact" or "Spacious" don't do anything you cannot tweak with manual properties. I.e. using "Custom" and adjusting other properties should give you exactly the same result as using a preset (otherwise it's not a preset at all, if it has some inherent magic unachievable otherwise).

I want to get in touch with you to learn what you had to adjust for your custom compact theme and see if we can do something similar.

YuriSizov:
@akien-mga 

> My only nitpick is whether the plural is justified in `editor/themes/` or if it should just be `editor/theme/` (like `scene/theme/`). The plural makes me think I'll find separate files for different themes, which doesn't seem to be the case, instead it's theme (singular) related classes (one of which, the Manager, defines multiple theme presets, plural).

Yeah, I went back and forth on this, but ended up with `themes` for a couple of reasons. The old file was already called this way, even though it was just a generator with multiple presets and custom options possible as its output. It irked me a bit, but at the same time it really wasn't one singular theme like it is in `scene/theme/`. I also didn't want to cause any more confusion than necessary between `editor/theme(s)` and `editor/theme(s)/editor_theme.h/cpp`.

So I settled on `editor/themes/` and thus `editor/themes/editor_theme.h/cpp`. I think it makes it a nice callback to the old name, if you feel particularly nostalgic too :) But ultimately, I also don't have a strong preference about it.

editor/themes/editor_theme_manager.h
YuriSizov: This particular method still writes to the passed config, can it still be demarked as const in this case? Methods below only read, so I think it's definitely fine there.

akien-mga: Right, I'm always unsure if it's the actual data or only the pointer/reference which are const. I checked and indeed `const SomeStruct &` would prevent modifying struct members (I thought it would just prevent overwriting the struct with a new one).

`f3` fails here:
```cpp
#include <iostream>

struct MyData {
	int a = 42;
	int b = -10;
};

void f1(MyData d) {
	d.a = 5;
	d.b = 7;
}

void f2(MyData &d) {
	d.a = 5;
	d.b = 7;
}

void f3(const MyData &d) {
	d.a = 5;
	d.b = 7;
}

int main() {
	MyData d1;
	f1(d1);
	std::cout << d1.a << " " << d1.b << std::endl;

	MyData d2;
	f2(d2);
    std::cout << d2.a << " " << d2.b << std::endl;

	MyData d3;
	f3(d3);
	std::cout << d3.a << " " << d3.b << std::endl;

	return 0;
}
```

YuriSizov:
> But as I understand it's irrelevant for this project because everything will be dynamically computed and you're more interested in what props rather than what values.

Yes, exactly! Thanks, I will investigate your edits for clues and message you if I have any further questions.

tokengamedev:
Awesome Work ❤️ 
It is exactly what I was asking for too.

Couple of things, that can be looked into.
- The font sizes can be played around in spacious and compact as in spacious the font looks small
- Default and compact looks the same. I assumed Default should be as the editor currently is (may be it is the screenshot)
- Icon sizes may be looked into.

For me @passivestar image above looks almost ideal to have,

YuriSizov:
@tokengamedev 

> The font sizes can be played around in spacious and compact as in spacious the font looks small

Because font sizes are configured on another settings page, I didn't want to automatically adjust them because it may not be expected by users. But as mentioned in the OP, I agree with the idea that for the best experience you might want to adjust them as well.

> Default and compact looks the same. I assumed Default should be as the editor currently is (may be it is the screenshot)

They are definitely not the same. Some elements have similar spacing, but that's because they are already very compact. But a lot of extra spacing is removed.

Compare [Default](https://private-user-images.githubusercontent.com/11782833/296005264-c4db35f8-abf9-410b-9122-2cc7fad15c72.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDUwNjUwNDgsIm5iZiI6MTcwNTA2NDc0OCwicGF0aCI6Ii8xMTc4MjgzMy8yOTYwMDUyNjQtYzRkYjM1ZjgtYWJmOS00MTBiLTkxMjItMmNjN2ZhZDE1YzcyLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAxMTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMTEyVDEzMDU0OFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTIyNWVjYTg1ZDA0YTYxN2M3NmJlNmU5OGM2MjkxOGIzNjFiYjIxMjE2NWFiOGNjMjJkZTRhZjViMjY1YmJjN2QmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.PfxpgDQd7-1zBSnLOVAnxlcKHsuIYKNXu2Qrl-9EgWs) to [Compact](https://private-user-images.githubusercontent.com/11782833/296005221-96125139-dcf3-4904-8557-3ead7033a3b9.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDUwNjUwNDgsIm5iZiI6MTcwNTA2NDc0OCwicGF0aCI6Ii8xMTc4MjgzMy8yOTYwMDUyMjEtOTYxMjUxMzktZGNmMy00OTA0LTg1NTctM2VhZDcwMzNhM2I5LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAxMTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMTEyVDEzMDU0OFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWQxMTBjNGM4OGJlOGU5ZGExN2RhZTQ3YjA5Nzk3NWNiYjUwMDdlMjZjYjRjOTBjZDA2YjZkNGQxOWQ1ZTJkYjEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.MlLeM3ntxu6WfCxsw6iMQ4n6nhVoHfLR3fKPybjtpKM). And then to the [Current master](https://private-user-images.githubusercontent.com/11782833/296005750-0a7dee1c-741d-43e2-aaa8-2d3d9011e8df.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDUwNjUwNDgsIm5iZiI6MTcwNTA2NDc0OCwicGF0aCI6Ii8xMTc4MjgzMy8yOTYwMDU3NTAtMGE3ZGVlMWMtNzQxZC00M2UyLWFhYTgtMmQzZDkwMTFlOGRmLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAxMTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMTEyVDEzMDU0OFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWM0MmE1MjRjNzUyOGI1NzllMjFlY2U5YjlkZTU2NTE4NTYzYzU1YjhiMmVhYWQ5NTEzOGIzZThmMjYyOWZhOGQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.o1m1_eeKIPuvYcAfdQ1fI0A37qNDqURtQTQ7yCkCCVY). Note that Default is not exactly the same as the current master because we've discussed reducing the spacing even for the default layout, which this does. But Compact is still more... compact.

> Icon sizes may be looked into.

Yeah, that's a good point. Currently all icons are scale solely based on the editor scale factor, but we may want to introduce a separate icon scale and use it here, especially for the Spacious mode.

Calinou:
The UI looks great on the screenshots, although Spacious feels a little excessive. It probably needs testing on a tablet to see if it really needs to be *this* spacious, or whether a middleground would be more suited.

editor/themes/editor_theme_manager.cpp
```diff
@@ -0,0 +1,2232 @@
...
+
+			config.accent_color = preset_accent_color;
+			config.base_color = preset_base_color;
+			config.contrast = preset_contrast;
+			config.draw_extra_borders = preset_draw_extra_borders;
+
+			EditorSettings::get_singleton()->set_initial_value("interface/theme/accent_color", config.accent_color);
+			EditorSettings::get_singleton()->set_initial_value("interface/theme/base_color", config.base_color);
+			EditorSettings::get_singleton()->set_initial_value("interface/theme/contrast", config.contrast);
+			EditorSettings::get_singleton()->set_initial_value("interface/theme/draw_extra_borders", config.draw_extra_borders);
```
adamscott, line 235: I would put this into a function.

editor/themes/editor_theme_manager.cpp
```diff
@@ -0,0 +1,2232 @@
...
+				preset_accent_color = Color(0.45, 0.75, 1.0);
+				preset_base_color = Color(0, 0, 0);
+				// The contrast rate value is irrelevant on a fully black theme.
+				preset_contrast = 0.0;
+				preset_draw_extra_borders = true;
+			} else { // Default
+				preset_accent_color = Color(0.44, 0.73, 0.98);
+				preset_base_color = Color(0.21, 0.24, 0.29);
+				preset_contrast = config.default_contrast;
+			}
```
adamscott, line 225: Is there a reason for this not being in a switch statement?

editor/themes/editor_theme_manager.cpp
```diff
@@ -0,0 +1,2232 @@
...
+			} else { // Default
+				preset_base_spacing = 4;
+				preset_extra_spacing = 0;
+			}
+
+			config.base_spacing = preset_base_spacing;
+			config.extra_spacing = preset_extra_spacing;
+
+			EditorSettings::get_singleton()->set_initial_value("interface/theme/base_spacing", config.base_spacing);
+			EditorSettings::get_singleton()->set_initial_value("interface/theme/additional_spacing", config.extra_spacing);
```
adamscott, line 267: Like the previous comment, could be inside a function.

editor/themes/editor_theme_manager.cpp
```diff
@@ -0,0 +1,2232 @@
...
+		if (custom_theme.is_valid()) {
+			theme->merge_with(custom_theme);
+		}
+	}
+
+	OS::get_singleton()->benchmark_end_measure("EditorTheme", "Generate Theme");
+	return theme;
+}
+
+bool EditorThemeManager::must_regenerate_theme() {
```
adamscott, line 2206: Is this method used to check if a theme regeneration is needed?

If so, I would name it something less ambiguous, like `is_theme_regeneration_needed()`. At first glance, I thought that `must_regenerate_theme()` had side-effects.

editor/themes/editor_theme.cpp
```diff
@@ -0,0 +1,131 @@
...
+// Keep in sync with Theme::get_color.
+Color EditorTheme::get_color(const StringName &p_name, const StringName &p_theme_type) const {
+	if (color_map.has(p_theme_type) && color_map[p_theme_type].has(p_name)) {
+		return color_map[p_theme_type][p_name];
+	} else {
+		if (editor_theme_types.has(p_theme_type)) {
+			WARN_PRINT(vformat("Trying to access a non-existing editor theme color '%s' in '%s'.", p_name, p_theme_type));
+		}
+		return Color();
+	}
```
adamscott, line 49:
```suggestion
	}
	
	if (editor_theme_types.has(p_theme_type)) {
		WARN_PRINT(vformat("Trying to access a non-existing editor theme color '%s' in '%s'.", p_name, p_theme_type));
	}
	return Color();
```

The else here is not needed. But it's a nitpick, feel free to ignore it.

adamscott (approved):
Few nitpicks, but it's great overall! Great job @YuriSizov !

editor/themes/editor_theme.cpp
YuriSizov: I'll ignore it for now because I made no changes to this class, only extracted it. 🙃 It will need some cleanup in future, as outlined by the comment at the top, so code style can be improved then.

editor/themes/editor_theme_manager.cpp
YuriSizov: What would be the benefit compared to it just being scoped?

editor/themes/editor_theme_manager.cpp
YuriSizov: I don't know, I didn't change it. I guess it could be!

editor/themes/editor_theme_manager.cpp
adamscott: It's just that I find that it's a long if statement. But yeah, you're right.

Listwon:
Removing margins/paddings in labels here (Import As, Compress) hurts readability, it shouldn't stick to the edge of the window/screen. If we can afford large indentations for properties, we can afford subtle margins for labels too.
image

editor/themes/editor_theme_manager.cpp
Joseph-DiGiovanni: > Is there a reason for this not being in a switch statement?

You can't use a String in a switch statement as far as I'm aware.

YuriSizov:
@Listwon Thanks, this is known. This is why this is only the first draft of the compact mode and why further tuning is expected. I may look into some tweaks before we merge this, but it's not really blocking — for us to get the compact mode right with this PR, and follow-ups will address these issues.

editor/themes/editor_theme_manager.cpp
YuriSizov: Renamed it to `is_generated_theme_outdated()` to be as explicit as possible.

editor/themes/editor_theme_manager.cpp
YuriSizov: Ah, you are correct, of course! Only integers and enums are supported by C++ switch statement. This could, of course, be an enum, but we only ever need it in one place, so it's probably not worth it.

YuriSizov:
I addressed the feedback by @adamscott (the new method is renamed, formatting in some affected python files improved).

I also adjusted Trees in the compact mode, which should now be close to what @passivestar outlined in his proposal. I also made the defaults for the spacious mode a little less spacious. Here's quick comparison:

## Default preset

godot windows editor dev x86_64_2024-01-15_13-38-38

## Compact preset

godot windows editor dev x86_64_2024-01-15_13-38-51

## Spacious preset

godot windows editor dev x86_64_2024-01-15_13-39-07

YuriSizov:
With these updates I'd like for this to be considered for merging. This way others will be able to contribute to the compact mode math, and we can also start a tracker with the findings from early testers.

akien-mga (approved):
Looks good to me!

YuriSizov:
Rebased on `master` after https://github.com/godotengine/godot/pull/84193.

akien-mga:
Thanks!

editor/themes/editor_theme_manager.cpp
```diff
@@ -0,0 +1,2232 @@
...
+		p_theme->set_constant("resize_margin", "Window", 4 * EDSCALE);
+		p_theme->set_font("title_font", "Window", p_theme->get_font(SNAME("title"), EditorStringName(EditorFonts)));
+		p_theme->set_font_size("title_font_size", "Window", p_theme->get_font_size(SNAME("title_size"), EditorStringName(EditorFonts)));
+
+		// AcceptDialog.
+		p_theme->set_stylebox("panel", "AcceptDialog", p_config.dialog_style);
+		p_theme->set_constant("buttons_separation", "AcceptDialog", 8 * EDSCALE);
+
+		// FileDialog.
+		p_theme->set_icon("folder", "FileDialog", p_theme->get_icon(SNAME("Folder"), EditorStringName(EditorIcons)));
```
KoBeWi, line 1165: Inconsistent use of SNAME. This line has 4 StringNames, but 2 of them use raw strings. Though we have such cases everywhere in the codebase :/

YuriSizov: Yeah, we can do a cleanup here. I also wanted to add `get_editor_icon` to `EditorTheme` to avoid all the repetition.

PR #87090, opened 2024-01-11
C#: Consolidate `MSBuild` actions

Repiteo:
The majority of `build_asemblies.py` actions are now handled by a single solution file. This removes a handful of redundant rebuilds that previously occurred, and helps eliminate output bloat (particularly for warnings, which won't output warnings whatsoever during the `GodotSharp` second pass). Also, all actions that previously redirected the output directory will instead copy the relevant outputs to a designated directory. I'm unsure if there's ramifications for these changes outside of the buildsystem, but within those constraints everything works as expected

### DIFF
--- a/modules/mono/build_scripts/build_assemblies.py
+++ b/modules/mono/build_scripts/build_assemblies.py
@@ -193,69 +193,6 @@ def run_msbuild(tools: ToolsLocation, sln: str, msbuild_args: Optional[List[str]
     return subprocess.call(args, env=msbuild_env)
 
 
-def build_godot_api(msbuild_tool, module_dir, output_dir, push_nupkgs_local, precision):
-    target_filenames = [
-        "GodotSharp.dll",
-        "GodotSharp.pdb",
-        "GodotSharp.xml",
-        "GodotSharpEditor.dll",
-        "GodotSharpEditor.pdb",
-        "GodotSharpEditor.xml",
-        "GodotPlugins.dll",
-        "GodotPlugins.pdb",
-        "GodotPlugins.runtimeconfig.json",
-    ]
-
-    for build_config in ["Debug", "Release"]:
-        editor_api_dir = os.path.join(output_dir, "GodotSharp", "Api", build_config)
-
-        targets = [os.path.join(editor_api_dir, filename) for filename in target_filenames]
-
-        args = ["/restore", "/t:Build", "/p:Configuration=" + build_config, "/p:NoWarn=1591"]
-        if push_nupkgs_local:
-            args += ["/p:ClearNuGetLocalCache=true", "/p:PushNuGetToLocalSource=" + push_nupkgs_local]
-        if precision == "double":
-            args += ["/p:GodotFloat64=true"]
-
-        sln = os.path.join(module_dir, "glue/GodotSharp/GodotSharp.sln")
-        exit_code = run_msbuild(
-            msbuild_tool,
-            sln=sln,
-            msbuild_args=args,
-        )
-        if exit_code != 0:
-            return exit_code
-
-        # Copy targets
-
-        core_src_dir = os.path.abspath(os.path.join(sln, os.pardir, "GodotSharp", "bin", build_config))
-        editor_src_dir = os.path.abspath(os.path.join(sln, os.pardir, "GodotSharpEditor", "bin", build_config))
-        plugins_src_dir = os.path.abspath(os.path.join(sln, os.pardir, "GodotPlugins", "bin", build_config, "net6.0"))
-
-        if not os.path.isdir(editor_api_dir):
-            assert not os.path.isfile(editor_api_dir)
-            os.makedirs(editor_api_dir)
-
-        def copy_target(target_path):
-            from shutil import copy
-
-            filename = os.path.basename(target_path)
-
-            src_path = os.path.join(core_src_dir, filename)
-            if not os.path.isfile(src_path):
-                src_path = os.path.join(editor_src_dir, filename)
-            if not os.path.isfile(src_path):
-                src_path = os.path.join(plugins_src_dir, filename)
-
-            print(f"Copying assembly to {target_path}...")
-            copy(src_path, target_path)
-
-        for scons_target in targets:
-            copy_target(scons_target)
-
-    return 0
-
-
 def generate_sdk_package_versions():
     # I can't believe importing files in Python is so convoluted when not
     # following the golden standard for packages/modules.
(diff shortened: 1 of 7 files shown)


paulloz:
I like this a lot. Haven't looked too closely at it yet, but having a single centralized solution is the biggest usability boon here, IMHO. Streamlining the currently convoluted build process would also be a good thing. In general, MSBuild is a very versatile tool that we should use more of.

modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj
```diff
@@ -61,4 +60,18 @@
...
+      <SecondPassProjects Include="..\..\..\glue\GodotSharp\GodotPlugins\GodotPlugins.csproj"/>
+      <SecondPassProperties Include="WarningLevel=0"/> <!-- Any relevant warnings would've already been logged in first pass -->
+      <SecondPassProperties Condition="'$(Configuration)' == 'Debug'" Include="Configuration=Release"/>
+      <SecondPassProperties Condition="'$(Configuration)' == 'Release'" Include="Configuration=Debug"/>
+      <SecondPassProperties Condition="'$(GodotPlatform)' != ''" Include="GodotPlatform=$(GodotPlatform)"/>
+      <SecondPassProperties Condition="'$(GodotFloat64)' != ''" Include="GodotFloat64=$(GodotFloat64)"/>
+      <SecondPassProperties Condition="'$(GodotNoDeprecated)' != ''" Include="GodotNoDeprecated=$(GodotNoDeprecated)"/>
+    </ItemGroup>
+    <MSBuild Projects="@(SecondPassProjects)" Properties="@(SecondPassProperties)"/>
+  </Target>
```
raulsntos, line 76: Is this just to ensure the `GodotSharp` and `GodotSharpEditor` projects are built before `GodotTools`? If so, the "SecondPass" name seems misleading as this would happen _before_ the build and not _after_.

Also, the `Configuration` properties are swapped.

However, I think it would be better to rely on MSBuild to understand which projects need to be built first when resolving the dependency graph. Did you try changing the `<Reference Include="GodotSharp">` into a proper `<ProjectReference>` instead of using a custom target?

modules/mono/glue/GodotSharp/GodotPlugins/GodotPlugins.csproj
```diff
@@ -15,4 +15,7 @@
     <ProjectReference Include="..\GodotSharp\GodotSharp.csproj" />
   </ItemGroup>
 
+  <Target Name="CopyFiles" AfterTargets="PostBuildEvent">
+    <Copy SourceFiles="$(OutputPath)\$(AssemblyName).dll;$(OutputPath)\$(AssemblyName).pdb;$(OutputPath)\$(AssemblyName).runtimeconfig.json" DestinationFolder="$(SolutionDir)\..\..\bin\GodotSharp\Api\$(Configuration)"/>
+  </Target>
```
raulsntos, line 20: I don't really like this kind of thing because we have to explicitly specify which files to copy, so if the build output changes we may end up missing files. But it's probably fine for now.

modules/mono/build_scripts/build_assemblies.py
```diff
@@ -363,48 +300,29 @@ def build_all(
...
+    return run_msbuild(msbuild_tool, sln=sln, chdir_to=module_dir, msbuild_args=args)
 
 
 def main():
     import argparse
     import sys
 
     parser = argparse.ArgumentParser(description="Builds all Godot .NET solutions")
-    parser.add_argument("--godot-output-dir", type=str, required=True)
+    parser.add_argument("--godot-output-dir", type=str, default="", help="Deprecated")
```
raulsntos, line 325: I'm fine with deprecating this, I don't think I've ever used it with anything other than `bin`, but I wonder if this will break someone's workflow.

raulsntos (changes requested):
There's a couple of issues with the custom "SecondPass" target that I think we can resolve by using the standard MSBuild dependency graph instead. Other than that, the changes look fine and I like consolidating our multiple solution files into one.

neikeq:
I would like to review this more deeply soon, but as a quick observation, while having a single solution is probably the correct decision, it's actually desirable not to load all projects when you're just working on a single area. Specially, considering that `GodotSharp` is huge because of the generated classes. And having to unload projects every time you open the solution or want to change areas is too burdensome.
Fortunately, this is precisely what [solution filters](https://learn.microsoft.com/en-us/visualstudio/ide/filtered-solutions?view=visualstudio) are for. So instead of removing the current solutions, just change them to solution filters.

modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj
```diff
@@ -61,4 +60,18 @@
     <ProjectReference Include="..\GodotTools.Core\GodotTools.Core.csproj" />
   </ItemGroup>
 
+  <Target Name="SecondPass" BeforeTargets="PreBuildEvent">
+    <ItemGroup>
+      <SecondPassProjects Include="..\..\..\glue\GodotSharp\GodotSharp\GodotSharp.csproj"/>
+      <SecondPassProjects Include="..\..\..\glue\GodotSharp\GodotSharpEditor\GodotSharpEditor.csproj"/>
+      <SecondPassProjects Include="..\..\..\glue\GodotSharp\GodotPlugins\GodotPlugins.csproj"/>
+      <SecondPassProperties Include="WarningLevel=0"/> <!-- Any relevant warnings would've already been logged in first pass -->
```
neikeq: I don't understand why this is needed. When building the entire solution at once, does MSBuild print warnings twice for a project if that project is referenced by another one within the solution?

modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj
Repiteo: > Also, the `Configuration` properties are swapped.

That's the entire point. The current `build_assemblies.py` does multiple passes when building the Godot API, unlike every other project which only builds one of `Release` or `Debug`. This "second pass" emulates that behavior.

It's *probably* a mistake to have multiple passes instead of just letting `build_assemblies.py` build both `Debug` and `Release` all at once, but I wanted this implementation to emulate the original as much as possible

neikeq:
The old code builds both `Debug` and `Release` configurations of `GodotSharp` and copies them to `${GodotSourceRoot}/bin/GodotSharp/Api/${Configuration}`. The new code no longer does that, and only builds one.

To be fair, I don't remember why we build both, because I don't imagine . Perhaps one is completely unused and this is just a vestige from the Godot 3.x times?

Another observation: the release builds (`release_debug`) of the Godot editor look for the `Debug` configuration of the `GodotSharp` assembly.

modules/mono/editor/GodotTools/GodotTools/GodotTools.csproj
raulsntos: I see, but when I looked at the `.binlog`, this actually ends up building `GodotSharp` 3 times. I think it would be better to only have MSBuild care about one build = one configuration, and then we can invoke MSBuild multiple times from `build_assemblies.py` for each of the configurations we need to build.

That said, I'm pretty sure we don't need `build_assemblies.py` to build more than one configuration. As @neikeq said in https://github.com/godotengine/godot/pull/87090#issuecomment-5799568796, I think this is a vestige from when we used to reference the DLL directly in Godot 3.x, instead of using NuGet packages. Nowadays, you would only need to build for one configuration and push the packages to NuGet.org or your local NuGet feed. For local builds you may want to build `Debug`, but `godot-build-scripts` only cares about `Release`.

Repiteo: Alright, I'll remove it outright

PR #83014, opened 2023-10-08
Add deprecation notice to a duplicate method of class `Window`

Chubercik:
- Fixes #82936.
- Closes #87093.

This pull request adds a deprecation warning to the `Window::move_to_foreground()` method, as internally it contains the same code as the `Window::grab_focus()` method.

In a future version of Godot (*4.3*? *4.4*? ***5.x***?) said method ought to be removed, however, to maintain compatibility a deprecation warning should suffice for now.

**Note**: If this is not the right way to purge duplicates from the codebase I am more than willing to re-do it once I've been shown the proper way.

PR #87093, opened 2024-01-11
Remove duplication of methods' code of class `Window` using an alias

### LINKED ISSUE #82936: Duplicated Code in Window::move_to_foreground() and Window::grab_focus()
### Godot version

4.1.2.stable.mono

### System information

Desktop

### Issue description

The Window::move_to_foreground() and Window::grab_focus() methods in the Window class contain duplicated code that serves the same purpose.

```cpp
void Window::move_to_foreground() {
	ERR_MAIN_THREAD_GUARD;
	if (embedder) {
		embedder->_sub_window_grab_focus(this);
	} else if (window_id != DisplayServer::INVALID_WINDOW_ID) {
		DisplayServer::get_singleton()->window_move_to_foreground(window_id);
	}
}

void Window::grab_focus() {
	ERR_MAIN_THREAD_GUARD;
	if (embedder) {
		embedder->_sub_window_grab_focus(this);
	} else if (window_id != DisplayServer::INVALID_WINDOW_ID) {
		DisplayServer::get_singleton()->window_move_to_foreground(window_id);
	}
}
```

### Steps to reproduce

Open [scene/main/window.cpp](https://github.com/godotengine/godot/blob/master/scene/main/window.cpp)

not needed

Chubercik:
- Fixes #82936.
- Closes #83014.

This PR is an alternative fix to the issue linked above; what it does is it makes the `Window::move_to_foreground()` method an alias for the `Window::grab_focus()` method.

**Note**: I tried many different ways of making an alias in C++ - this version compiles, but I'm not sure if this is the best approach: ideally we'd like a situation where calling one method is internally a call of the other one.

Any suggestions for improvements are welcome.

### DIFF
--- a/scene/main/window.cpp
+++ b/scene/main/window.cpp
@@ -521,13 +521,7 @@ void Window::request_attention() {
 }
 
 void Window::move_to_foreground() {
-	ERR_MAIN_THREAD_GUARD;
-	if (embedder) {
-		embedder->_sub_window_grab_focus(this);
-
-	} else if (window_id != DisplayServer::INVALID_WINDOW_ID) {
-		DisplayServer::get_singleton()->window_move_to_foreground(window_id);
-	}
+	Window::grab_focus();
 }
 
 bool Window::can_draw() const {
--- a/scene/main/window.h
+++ b/scene/main/window.h
@@ -301,7 +301,7 @@ public:
 	bool is_maximize_allowed() const;
 
 	void request_attention();
-	void move_to_foreground();
+	_FORCE_INLINE_ void move_to_foreground();
 
 	virtual void set_visible(bool p_visible);
 	bool is_visible() const;
--- a/doc/classes/Window.xml
+++ b/doc/classes/Window.xml
@@ -211,6 +211,7 @@
 			<return type="void" />
 			<description>
 				Causes the window to grab focus, allowing it to receive user input.
+				[b]Note:[/b] Does the same thing as [method Window.move_to_foreground].
 			</description>
 		</method>
 		<method name="has_focus" qualifiers="const">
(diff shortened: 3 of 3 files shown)


bruvzg (commented):
I think https://github.com/godotengine/godot/pull/83014 is a better way to handle it.

scene/main/window.cpp
```diff
@@ -521,13 +521,7 @@ void Window::request_attention() {
...
 
 void Window::move_to_foreground() {
-	ERR_MAIN_THREAD_GUARD;
-	if (embedder) {
-		embedder->_sub_window_grab_focus(this);
-
-	} else if (window_id != DisplayServer::INVALID_WINDOW_ID) {
-		DisplayServer::get_singleton()->window_move_to_foreground(window_id);
-	}
+	Window::grab_focus();
```
AThousandShips, line 524:
```suggestion
	grab_focus();
```
Unnecessary

scene/main/window.h
```diff
@@ -301,7 +301,7 @@ class Window : public Viewport {
 	bool is_maximize_allowed() const;
 
 	void request_attention();
-	void move_to_foreground();
+	_FORCE_INLINE_ void move_to_foreground();
```
AThousandShips, line 304: This won't work AFAIK, inlining requires the definition, I'd instead suggest:
```suggestion
	_FORCE_INLINE_ void move_to_foreground() { grab_focus(); }
```
And removing from the source file

akien-mga:
Let's go with https://github.com/godotengine/godot/pull/83014 then, sorry for having you make a redundant PR, I wasn't sure which approach would be best.

Chubercik:
No worries, I created this one mainly to make it easier to make a decision, having both options to compare side by side :)

PR #87095, opened 2024-01-12
Add `CanvasItem::get_canvas_layer_node()`

### LINKED ISSUE #8863: Expose `CanvasItem.get_canvas_layer()` to GDScript/C#

An alternate UI solution implemented with scripts based on Node2D.

While handling inputs, I need to calculate the order of my custom control roots in the viewport to properly propagate input events from Viewport to my custom control roots.
Because `CanvasItem.get_canvas_layer()` is not exposed to GDScript/C#, It's impossible to do that like [how Godot Controls do that](https://github.com/godotengine/godot/blob/352434668923978f54f2236f20116fc96ebc9173/scene/gui/control.h#L155).

Expose `get_canvas_layer()` to GDScript.
It's also great to have a method to get which CanvasLayer node the CanvasItem node is in, just like you can get Viewport node using `CanvasItem.get_viewport()`.

```gds
extends Node2D
func foo():
	var layer: int = get_canvas_layer()
	var layerNode: CanvasLayer = get_canvas_layer_node()
```

As far as I know, there's no way to get the CanvasLayer from CanvasItem via script, since it's not exposed.
The only thing you can get is canvas RIDs, but you can't pull viewport canvas stacking layer data from RenderingServer.

It's basic Viewport/CanvasLayer functions in CanvasItem nodes.

TML233:
* Added `CanvasLayer::get_canvas_layer_node`, which returns the CanvasLayer node the CanvasItem is in. Returns `nullptr` when the CanvasItem is not in a CanvasLayer node.

Fix godotengine/godot-proposals#8863

### DIFF
--- a/doc/classes/CanvasItem.xml
+++ b/doc/classes/CanvasItem.xml
@@ -411,6 +411,12 @@
 				Returns the canvas item RID used by [RenderingServer] for this item.
 			</description>
 		</method>
+		<method name="get_canvas_layer_node" qualifiers="const">
+			<return type="CanvasLayer" />
+			<description>
+				Returns the [CanvasLayer] that contains this node, or [code]null[/code] if the node is not in any [CanvasLayer].
+			</description>
+		</method>
 		<method name="get_canvas_transform" qualifiers="const">
 			<return type="Transform2D" />
 			<description>
--- a/scene/main/canvas_item.cpp
+++ b/scene/main/canvas_item.cpp
@@ -1169,6 +1169,7 @@ void CanvasItem::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_local_mouse_position"), &CanvasItem::get_local_mouse_position);
 	ClassDB::bind_method(D_METHOD("get_global_mouse_position"), &CanvasItem::get_global_mouse_position);
 	ClassDB::bind_method(D_METHOD("get_canvas"), &CanvasItem::get_canvas);
+	ClassDB::bind_method(D_METHOD("get_canvas_layer_node"), &CanvasItem::get_canvas_layer_node);
 	ClassDB::bind_method(D_METHOD("get_world_2d"), &CanvasItem::get_world_2d);
 	//ClassDB::bind_method(D_METHOD("get_viewport"),&CanvasItem::get_viewport);
 
@@ -1323,6 +1324,11 @@ int CanvasItem::get_canvas_layer() const {
 	} else {
 		return 0;
 	}
+}
+
+CanvasLayer *CanvasItem::get_canvas_layer_node() const {
+	ERR_READ_THREAD_GUARD_V(nullptr);
+	return canvas_layer;
 }
 
 void CanvasItem::set_visibility_layer(uint32_t p_visibility_layer) {
--- a/scene/main/canvas_item.h
+++ b/scene/main/canvas_item.h
@@ -363,6 +363,7 @@ public:
 	virtual Rect2 get_anchorable_rect() const { return Rect2(0, 0, 0, 0); };
 
 	int get_canvas_layer() const;
+	CanvasLayer *get_canvas_layer_node() const;
 
 	CanvasItem();
 	~CanvasItem();


Sauermann:
Please edit in your opening post the last line and change it to: 
fix godotengine/godot-proposals#8863

This change will link this PR to that proposal and causes github to automatically close the proposal, when this PR gets merged.

doc/classes/CanvasItem.xml
```diff
@@ -405,6 +405,18 @@
 				Returns the [RID] of the [World2D] canvas where this item is in.
 			</description>
 		</method>
+		<method name="get_canvas_layer" qualifiers="const">
+			<return type="int" />
+			<description>
+				Returns the layer number of the [CanvasLayer] where this node is in.
```
Sauermann, line 411: This description should also mention the special case that the CanvasItem can be in the viewports default canvas layer.

TML233: > This description should also mention what happens, when there is no `CanvasLayer` or alternatively replace "[CanvasLayer]" by "canvas layer".

Thank you! I've already changed it and mentioned that `0` will be returned if there's no `CanvasLayer`.

TML233:
> Please edit in your opening post the last line and change it to: fix [godotengine/godot-proposals#8863](https://github.com/godotengine/godot-proposals/issues/8863)
> 
> This change will link this PR to that proposal and causes github to automatically close the proposal, when this PR gets merged. These keywords are explained in https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests.

Thank you! I've fixed both issues you mentioned.

doc/classes/CanvasItem.xml
```diff
@@ -405,6 +405,19 @@
...
+			<return type="int" />
+			<description>
+				Returns the layer number of the [CanvasLayer] where the node is in.
+				If the node is not in any [CanvasLayer], the layer number will be `0`.
+			</description>
+		</method>
+		<method name="get_canvas_layer_node" qualifiers="const">
+			<return type="CanvasLayer" />
+			<description>
+				Returns the [CanvasLayer] where the node is in, or null if the node is not in any [CanvasLayer].
```
AThousandShips, line 418:
```suggestion
				Returns the [CanvasLayer] that this node is in, or [code]null[/code] if the node is not in any [CanvasLayer].
```

Mickeon: We tend to avoid prepositions after the subject even if it's perfectly correct English because it's harder to read for non-native speakers _(speaking from experience writing these, not actually a maintainer)_, so the latter may be more appropriate.

If you want convey the relation between the two more, instead of using "of" which is pretty generic, you could say for example "Returns the [CanvasLayer] that contains this node".

Sauermann (approved):
I'm not familiar with `ERR_READ_THREAD_GUARD_V`, but otherwise the PR changes look correct.

groud:
I think I am ok with exposing the node, as there seem to be no way to retrieve it easily otherwise. 

I'd, however, avoid exposing `get_canvas_layer`. It bloating the API for something that could easily be done manually. See the function:
```
int CanvasItem::get_canvas_layer() const {
	ERR_READ_THREAD_GUARD_V(0);
	if (canvas_layer) {
		return canvas_layer->get_layer();
	} else {
		return 0;
	}
}
```
Also, I think the use case is not common enough to justify exposing `get_canvas_layer`.

TML233:
> I think I am ok with exposing the node, as there seem to be no way to retrieve it easily otherwise. 
> 
> 
> 
> I'd, however, avoid exposing `get_canvas_layer`. It bloating the API for something that could easily be done manually. See the function:
> 
> ```
> 
> int CanvasItem::get_canvas_layer() const {
> 
> 	ERR_READ_THREAD_GUARD_V(0);
> 
> 	if (canvas_layer) {
> 
> 		return canvas_layer->get_layer();
> 
> 	} else {
> 
> 		return 0;
> 
> 	}
> 
> }
> 
> ```
> 
> Also, I think the use case is not common enough to justify exposing `get_canvas_layer`.
> 
> 

Okay, I'll de-expose `get_canvas_layer` but keep `get_canvas_layer_node`.

doc/classes/CanvasItem.xml
```diff
@@ -411,6 +411,18 @@
...
+		<method name="get_canvas_layer" qualifiers="const">
+			<return type="int" />
+			<description>
+				Returns the layer index of the [CanvasLayer] that this node is in. Returns [code]0[/code] if this node is not in any [CanvasLayer].
+			</description>
+		</method>
+		<method name="get_canvas_layer_node" qualifiers="const">
+			<return type="CanvasLayer" />
+			<description>
+				Returns the [CanvasLayer] that this node is in, or [code]null[/code] if the node is not in any [CanvasLayer].
```
Mickeon, line 423: I previously commented on this https://github.com/godotengine/godot/pull/87095#discussion_r1450273251 but it went unseen, probably.

TML233: Sorry it might went unseen while i was updating the code. I'll change this into "Returns the [CanvasLayer] that contains this node."

Thanks!

Mickeon:
The commit could use a different name since `get_canvas_layer` is no longer exposed.

Mickeon (approved):
The documentation is completely fine, though

akien-mga:

PR #87096, opened 2024-01-12
Add `shader_cache_dir_valid` check to `_save_to_cache`

### LINKED ISSUE #87079: Disabling shader cache gives _save_to_cache error in editor
### Tested versions

Reproducible in: 4.2.1

I also think it will happen in 4.2, 4.1.3, 4.1.2, 4.1.1, 4.1 but haven't tested there.

### System information

Windows 11 64-bit, Ryzen 5 5500U, 8GB RAM, Nvidia GTX 1650

### Issue description

Disabling shader cache from `Project -> Project Settings -> Rendering -> Shader Compiler -> Shader Cache -> Enabled` gives the error:

`_save_to_cache: Condition "f.is_null()" is true.`

Though, the game runs fine, but I keep getting this error.

The reason I'm disabling shader cache is because of the issue I had with 3D models not being rendered on certain Android devices. The fix to this issue was to disable shader cache, and doing this fixed the rendering problem. I had opened a issue (resolved now) for this earlier here - https://github.com/godotengine/godot/issues/86956

### Steps to reproduce

1. Download the MRP provided.
2. Extract the apk/project files to your desired location.
3. Open the project files using Godot.
4. Run `main.tscn`
5. Upon running, you'll see `_save_to_cache` error message shown 4 times.

[Android 3D Rendering Issue.zip](https://github.com/godotengine/godot/files/13905039/Android.3D.Rendering.Issue.zip)

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/87079

Alternatively one could argue that we should add the check to the call site instead of inside the function, but I think add a protection inside it is harmless and prevent future mistakes.

### DIFF
--- a/drivers/gles3/shader_gles3.cpp
+++ b/drivers/gles3/shader_gles3.cpp
@@ -611,6 +611,7 @@ void ShaderGLES3::_save_to_cache(Version *p_version) {
 #ifdef WEB_ENABLED // not supported in webgl
 	return;
 #else
+	ERR_FAIL_COND(!shader_cache_dir_valid);
 #if !defined(ANDROID_ENABLED) && !defined(IOS_ENABLED)
 	if (RasterizerGLES3::is_gles_over_gl() && (glGetProgramBinary == NULL)) { // ARB_get_program_binary extension not available.
 		return;
--- a/drivers/gles3/shader_gles3.h
+++ b/drivers/gles3/shader_gles3.h
@@ -209,7 +209,9 @@ protected:
 				_compile_specialization(s, p_variant, version, p_specialization);
 				version->variants[p_variant].insert(p_specialization, s);
 				spec = version->variants[p_variant].lookup_ptr(p_specialization);
-				_save_to_cache(version);
+				if (shader_cache_dir_valid) {
+					_save_to_cache(version);
+				}
 			}
 		} else if (spec->build_queued) {
 			// Still queued, wait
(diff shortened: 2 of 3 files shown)


akien-mga:
> Alternatively one could argue that we should add the check to the call site instead of inside the function, but I think add a protection inside it is harmless and prevent future mistakes.

I think it might need to do both.

Usually we would use `ERR_FAIL_COND` if the contract of a function is not respected. Here, I think `_save_to_cache` should never be called if there's no valid shader cache dir, so it should raise an error.

And then the call site where it's not properly doing the check should be changed. `_save_to_cache` is called in three files, and two of them properly check:
```cpp
if (shader_cache_dir_valid) {
_save_to_cache(p_version);
}
```

But that check is missing in `shader_gles3.h`:
```cpp
Version::Specialization *spec = version->variants[p_variant].lookup_ptr(p_specialization);
if (!spec) {
	if (false) {
		// Queue load this specialization and use defaults in the meantime (TODO)

		spec = version->variants[p_variant].lookup_ptr(specialization_default_mask);
	} else {
		// Compile on the spot
		Version::Specialization s;
		_compile_specialization(s, p_variant, version, p_specialization);
		version->variants[p_variant].insert(p_specialization, s);
		spec = version->variants[p_variant].lookup_ptr(p_specialization);
		_save_to_cache(version);
	}
```

jsjtxietian:
> But that check is missing in `shader_gles3.h`

Yes I noticed that too, you are right, it need to do both, an `if` guard at the call site and ERR_FAIL_COND in the function .

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

SinaZK:
Could we see this on 4.1.5 too? We have a released mobile game using 4.1 and we need to get rid of many annoying errors when shader cache is off in the project settings.

PR #87099, opened 2024-01-12
Implement `trim_final_newlines` setting and functionality

### LINKED ISSUE #8867: Add an editor setting to trim final newlines when saving a script

The relevant aspect of the project is that the built-in script editor is the primary method of interacting with scripts. 

Often when working on complex code, stray newlines will make their way into a script at the end of the file. Sometimes these lines go uncaught and get committed into the repo. 

- Adding `trim_final_newlines_on_save` setting will help to clean up any stray newlines. 

This helps to ensure the code cleanliness while interacting with the built-in editor. This would help ensure parity with other editors like Visual Studio Code ([trimFinalNewlines](https://code.visualstudio.com/docs/getstarted/settings#_default-settings)). 

- If `trim_final_newlines_on_save` is enabled and a user saves a script, it will remove any newlines after the final newline of the file. 

This can be worked around with an editor plugin, but it is a general quality of life improvement to the script editor.

It fits with the current `trim_whitespace_on_save` setting that is already present.

bitwise-aiden:
Closes: https://github.com/godotengine/godot-proposals/issues/8867

**Currently in a draft state as I need to figure out the best method for collapsing carets that fall outside the removed area of text.** 

This PR implements the `trim_final_newlines` setting to reduce the newlines at the end of file to the expected 1 newline. This puts the built-in script editor on parity with other editors like Visual Studio Code and helps reduce stray lines making their way into user repos. 

### Implementation

I've opted to remove all bar the last empty lines from the file, checking simply for length 0. It intentionally runs after the `trim_trailing_whitespace` as to consume lines that only contain whitespace characters as well. In the case that `trim_trailing_whitespace_on_save` is not enabled it will then trim all lines back to the line with whitespace, leaving that untouched with a single newline following. 

This relies on the functionality added in https://github.com/godotengine/godot/pull/86978 for caret and selection support.

### DIFF
--- a/doc/classes/EditorSettings.xml
+++ b/doc/classes/EditorSettings.xml
@@ -1012,6 +1012,9 @@
 		<member name="text_editor/behavior/files/restore_scripts_on_load" type="bool" setter="" getter="">
 			If [code]true[/code], reopens scripts that were opened in the last session when the editor is reopened on a given project.
 		</member>
+		<member name="text_editor/behavior/files/trim_final_newlines_on_save" type="bool" setter="" getter="">
+			If [code]true[/code], trims all empty newlines after the final newline when saving a script. Final newlines refer to the empty newlines found at the end of files. Since these serve no practical purpose, they can and should be removed to make version control diffs less noisy.
+		</member>
 		<member name="text_editor/behavior/files/trim_trailing_whitespace_on_save" type="bool" setter="" getter="">
 			If [code]true[/code], trims trailing whitespace when saving a script. Trailing whitespace refers to tab and space characters placed at the end of lines. Since these serve no practical purpose, they can and should be removed to make version control diffs less noisy.
 		</member>
--- a/editor/code_editor.cpp
+++ b/editor/code_editor.cpp
@@ -1123,6 +1123,31 @@ void CodeTextEditor::trim_trailing_whitespace() {
 	}
 }
 
+void CodeTextEditor::trim_final_newlines() {
+	int final_line = text_editor->get_line_count() - 1;
+	int check_line = final_line;
+
+	String line = text_editor->get_line(check_line);
+
+	while (line.is_empty() && check_line > -1) {
+		--check_line;
+
+		line = text_editor->get_line(check_line);
+	}
+
+	++check_line;
+
+	if (check_line < final_line) {
+		text_editor->begin_complex_operation();
+
+		text_editor->remove_text(check_line, 0, final_line, 0);
+
+		text_editor->merge_overlapping_carets();
+		text_editor->end_complex_operation();
+		text_editor->queue_redraw();
+	}
+}
+
 void CodeTextEditor::insert_final_newline() {
 	int final_line = text_editor->get_line_count() - 1;
 	String line = text_editor->get_line(final_line);
--- a/editor/plugins/script_text_editor.cpp
+++ b/editor/plugins/script_text_editor.cpp
@@ -431,6 +431,10 @@ void ScriptTextEditor::trim_trailing_whitespace() {
 	code_editor->trim_trailing_whitespace();
 }
 
+void ScriptTextEditor::trim_final_newlines() {
+	code_editor->trim_final_newlines();
+}
+
 void ScriptTextEditor::insert_final_newline() {
 	code_editor->insert_final_newline();
 }
@@ -1427,6 +1431,9 @@ void ScriptTextEditor::_edit_option(int p_op) {
 		case EDIT_TRIM_TRAILING_WHITESAPCE: {
 			trim_trailing_whitespace();
 		} break;
+		case EDIT_TRIM_FINAL_NEWLINES: {
+			trim_final_newlines();
+		} break;
 		case EDIT_CONVERT_INDENT_TO_SPACES: {
 			code_editor->set_indent_using_spaces(true);
 			convert_indent();
@@ -2299,6 +2306,7 @@ void ScriptTextEditor::_enable_code_editor() {
 	edit_menu->get_popup()->add_separator();
 	edit_menu->get_popup()->add_shortcut(ED_GET_SHORTCUT("ui_text_completion_query"), EDIT_COMPLETE);
 	edit_menu->get_popup()->add_shortcut(ED_GET_SHORTCUT("script_text_editor/trim_trailing_whitespace"), EDIT_TRIM_TRAILING_WHITESAPCE);
+	edit_menu->get_popup()->add_shortcut(ED_GET_SHORTCUT("script_text_editor/trim_final_newlines"), EDIT_TRIM_FINAL_NEWLINES);
 	{
 		PopupMenu *sub_menu = memnew(PopupMenu);
 		sub_menu->add_shortcut(ED_GET_SHORTCUT("script_text_editor/convert_indent_to_spaces"), EDIT_CONVERT_INDENT_TO_SPACES);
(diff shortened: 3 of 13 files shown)


editor/plugins/script_text_editor.cpp
```diff
@@ -2462,6 +2470,7 @@ void ScriptTextEditor::register_editor() {
 	ED_SHORTCUT("script_text_editor/evaluate_selection", TTR("Evaluate Selection"), KeyModifierMask::CMD_OR_CTRL | KeyModifierMask::SHIFT | Key::E);
 	ED_SHORTCUT("script_text_editor/toggle_word_wrap", TTR("Toggle Word Wrap"), KeyModifierMask::ALT | Key::Z);
 	ED_SHORTCUT("script_text_editor/trim_trailing_whitespace", TTR("Trim Trailing Whitespace"), KeyModifierMask::CMD_OR_CTRL | KeyModifierMask::ALT | Key::T);
+	ED_SHORTCUT("script_text_editor/trim_final_newlines", TTR("Trim Final Newlines"), KeyModifierMask::CMD_OR_CTRL | KeyModifierMask::SHIFT | Key::T);
```
bitwise-aiden, line 2473: I've added this to have parity with `Trim Tailing Whitespace`, but I'm not sure that it has the same utility. I could see removing this, or possibly defaulting to no keybind.

editor/code_editor.cpp
```diff
@@ -1129,6 +1129,53 @@ void CodeTextEditor::trim_trailing_whitespace() {
...
+
+			int loc = text_editor->get_caret_line(c);
+			if (loc > check_line) {
+				if (i == 0) {
+					text_editor->set_caret_line(check_line);
+				} else {
+					text_editor->remove_caret(c);
+				}
+			}
+		}
```
bitwise-aiden, line 1172: This is my first pass at implementing the caret and selection handling for when either are outside of the area that is now going to be deleted. 

I have gotten it down to only a small fraction of my manual tests resulting in errors, but would love to know if this is heading in the right direction in case I am completely off base. 

Known passing:
- Caret inside section to be deleted
- Caret outside section to be deleted
- Selection inside section to be deleted
- Selection outside section to be deleted
- Selection overlapping section to be deleted when caret is at the bottom of selection

Known failing: 
- Selection overlapping section to be deleted when caret is at the top of selection
  - Note: This failure occurs when the mouse clicks somewhere else, exiting the selection mode. It appears that something is still referencing the previous to line despite calling `select` for the caret. 
- Multiple carets inside section to be deleted
  - Note: There is one error that the caret index is out of range in `get_caret_line` and then one error per caret for `get_line_wrap_index_at_column` with line out of range.
  
Any suggestions on this would be appreciated.

bitwise-aiden: > Multiple carets inside section to be deleted

Fixed by iterating backwards over the cursors. Previously I was deleting then trying to access.

kitbdev:
Yeah, the selection code is more complex than it needs to be at the moment.
Try using `TextEdit::set_selection_mode(TextEdit::SelectionMode::SELECTION_MODE_NONE, check_line, 0, c)` when changing the selection, so that the caret's selecting_line gets set to a not invalid line.

Or if #86978 is merged first, you wouldn't have to worry about carets or selections at all, the `remove_text()` would handle it.

bitwise-aiden:
> Or if https://github.com/godotengine/godot/pull/86978 is merged first, you wouldn't have to worry about carets or selections at all, the remove_text() would handle it.

That's awesome context @kitbdev! Thanks :D 
I don't have a great deal of bandwidth to push this forward at the moment so I'll remove the caret editing part, test against your branch, and leave it until you've pushed.

bitwise-aiden:
@kitbdev I've just tested against this change, it all looks good, the errors have gone away and any caret issues seem to be resolved too. Thank you for taking all that on 🙌🏻

editor/code_editor.cpp
bitwise-aiden: No longer needed due to @kitbdev's comment + PR.

editor/plugins/script_text_editor.cpp
bitwise-aiden: While waiting for the cursor PR to be merged, @AThousandShips, do you have any preference on this addition?

AThousandShips: No opinions here on my part

doc/classes/EditorSettings.xml
```diff
@@ -838,6 +838,9 @@
 		<member name="text_editor/behavior/files/restore_scripts_on_load" type="bool" setter="" getter="">
 			If [code]true[/code], reopens scripts that were opened in the last session when the editor is reopened on a given project.
 		</member>
+		<member name="text_editor/behavior/files/trim_final_newlines_on_save" type="bool" setter="" getter="">
+			If [code]true[/code], trims all newlines after the final newline when saving a script. Final newlines refer to the newlines found at the end of files. Since these serve no practical purpose, they can and should be removed to make version control diffs less noisy.
```
AThousandShips, line 842: I'd suggest using "empty lines" in the wording here

editor/code_editor.cpp
```diff
@@ -1129,6 +1129,27 @@ void CodeTextEditor::trim_trailing_whitespace() {
...
+	int final_line = text_editor->get_line_count() - 1;
+	int check_line = final_line;
+
+	String line = text_editor->get_line(final_line);
+
+	while (line.is_empty() && check_line > -1) {
+		line = text_editor->get_line(--check_line);
+	}
+
+	if (check_line++ < final_line) {
```
AThousandShips, line 1142:
```suggestion
	if (check_line < final_line) {
		++check_line;
```
More readable IMO, it doesn't matter outside of the block anyway

editor/code_editor.cpp
```diff
@@ -1129,6 +1129,27 @@ void CodeTextEditor::trim_trailing_whitespace() {
 	}
 }
 
+void CodeTextEditor::trim_final_newlines() {
+	int final_line = text_editor->get_line_count() - 1;
+	int check_line = final_line;
+
+	String line = text_editor->get_line(final_line);
```
AThousandShips, line 1136:
```suggestion
	String line = text_editor->get_line(check_line);
```
Clearer intent IMO

editor/plugins/script_text_editor.cpp
bitwise-aiden: Cool, I think that I will add it as a potential keybind with no default since as @Calinou suggested this feature will be on by default.

editor/code_editor.cpp
bitwise-aiden: That was supposed to be an `if(++check_line < final_line)` to avoid triggering removal when there is zero or one empty lines at the end of the file. 

Is the preference increment/decrement outside of operations, or is this and the `line = text_editor->get_line(--check_line)` fine?

AThousandShips: I'd say generally yes, it's much better for readability

bitwise-aiden: Awesome, I'll update this tonight then we are just waiting on the caret/selection PR to be merged.

editor/plugins/text_shader_editor.cpp
```diff
@@ -1225,6 +1234,9 @@ TextShaderEditor::TextShaderEditor() {
 	disk_changed->add_button(TTR("Resave"), !DisplayServer::get_singleton()->get_swap_cancel_ok(), "resave");
 	disk_changed->connect("custom_action", callable_mp(this, &TextShaderEditor::save_external_data));
 
+	trim_trailing_whitespace_on_save = EDITOR_GET("text_editor/behavior/files/trim_trailing_whitespace_on_save");
```
bitwise-aiden, line 1237: This line looks like it may have been mistakenly removed in [this PR](https://github.com/godotengine/godot/pull/88474/files#diff-cd71ad8157d2fd2599a3fd5d1944dc81bec04203ba522ba8dbc89a01bdb2064a). Without the initializer getting these values it will mean opening the shader editor won't respect the users settings.

bitwise-aiden:
Since @kitbdev's PR has been merged https://github.com/godotengine/godot/pull/86978, this is now ready for final review. 

I've rebased and tested the cases locally where this was errorring earlier (multicaret / multi line selection) and everything seems to be working well.

editor/plugins/text_shader_editor.cpp
bitwise-aiden: I tested it some more and I must have had the settings window open which would have captured the ctrl + S meaning that the trim didn't occur. 

Everything looks good now!

bitwise-aiden:
Updated with Kitbdev's last comment. Ready for review.

akien-mga (approved):
Looks good to me.

AThousandShips (approved):

akien-mga:
Thanks!

Issue #49781, opened 2021-06-20
Sprite with shader that modifies COLOR ignores modulate property


KoBeWi:
### Godot version

d696bd1

### System information

W10

### Issue description

When you add a simple shader that changes `COLOR`, modulate property is ignored. It was fine in 3.x

### Steps to reproduce

1. Add Sprite2D node
2. Apply this shader
```
shader_type canvas_item;

void fragment() {
	COLOR = texture(TEXTURE, UV);
}
```
3. Try changing modulate property

insomniacUNDERSCORElemon:
Modulation also seems to not work with polygon2d and line2d nodes, in most/all configurations that I can tell. For instance no change when textured (the tooltip says modulate applies to textures), untextured, or when using vertex colors.

Though the most sense of having this (aside from textured polygons) is for vertex colors due to their complexity, which would really benefit here for instance if you wanted something grayscale like a cloud to be recolored for different effects (potions, clouds, team colors, having a library to re-use simple art in different situations etc) which does work in 3.3.2.

So hopefully this isn't intended, at least without something to replace it (like a palette system so you can alter colors and they change all vertices with that index, especially if there was a way to add+control color mixing). With `clip children` (when fixed+no blending, see 49198) it *might* be something that could be wrangled into something dynamic as-needed by someone clever... but that isn't quite as quick and easy as modulate is.

----------------------

Somewhat different, but issue #25120 is related. Because in previous versions (3.1 beta 1 and before) you could have vertex colors along with vertices that still use the color property instead. Which was useful in a similar-but-different way. Not that I think it was possible to have much control over *which* vertices had vertex color and which ones used the color property (but if your colors were vaguely in the same area, it might work).

groud:
> Modulation also seems to not work with polygon2d and line2d nodes, in most/all configurations that I can tell. For instance no change when textured (the tooltip says modulate applies to textures), untextured, or when using vertex colors.

https://github.com/godotengine/godot/pull/49863 will likely fix this issue.

But the problem @KoBeWi has here is different, it is unlikely to be fixed by this PR.

h0lley:
Still occurring in Godot 4 alpha 3.

In Godot 3 multiplication with modulation color happened implicitly unless `MODULATE` was used explicitly in shader code.
meanwhile, `MODULATE` was removed from the fragment built-ins entirely - presumably intentionally as the docs already reflect the change.
[...]

Issue #63550, opened 2022-07-27
GD4 Corrupting some existing GD3 arraymeshes 


tinmanjuggernaut:
### Godot version

4.0alpha12

### System information

Win 10/64 NVIDIA GeForce GTX 1060/PCIe/SSE2

### Issue description

As I'm converting over my game and hundreds of assets to GD4, some of my meshes are strange in the way they respond to light. Most are fine. Maybe 5% so far are messed up. They previously looked the same as all the other pieces in GD3. I have a village pack from the same artist where nearly every asset uses the same wood material. Every piece rendered the same in GD3. Now some of the faces on some of the objects look like their normals are inverted. Yet blender doesn't think so. Nor does GD4 if I reimport the object from the original file.

I also have other objects that flicker and invert depending on camera angle, though it's the same wood material that everything in the village uses. I'm still testing that, it may be another issue.

We imported these meshes with the default settings in GD3 and saved them in scene files in the native arraymesh format.
They were converted to GD4 using aaronfranke's script https://gist.github.com/aaronfranke/79b424226475d277d0035b7835b09c5f, which basically just renames poolarray to packedarray and some other renames. The array data should have remained the same.

## Foot bridge
Here's a bridge with an empty standard material, albedo at 50V. You can see the sun is at X=-30 and the bridge is lit up normally. 

![image](https://user-images.githubusercontent.com/632766/181307267-55cf156b-604b-4be9-abc0-7769f93f3114.png)

Now the sun is at X=-150 (30 degrees off of 0). The top of the bridge should be lit up exactly the same as above. Look at the sphere and the ground.

![image](https://user-images.githubusercontent.com/632766/181307571-4e651731-1bee-456d-a14a-b8153e8c1330.png)

I have to rotate it this far (-120 or 50 degrees off 0) to get the top of the bridge to light up when the rotation is on that side.
![image](https://user-images.githubusercontent.com/632766/181307802-dd9f364c-9124-4a7a-81fa-b0d3800c2c98.png)

## Beams

These two objects are using the exact same material. You can see the light wood on the top faces of the left beam. But the side faces are all dark. They should be as light a the right piece. They are from the same asset pack and were imported the same way.

![image](https://user-images.githubusercontent.com/632766/181312721-229b38db-706b-4e08-89b1-26d4be13001b.png)

Standard white material

![image](https://user-images.githubusercontent.com/632766/181312916-798a854e-798b-4692-aae5-4015765e43ca.png)

[...]

PR #87106, opened 2024-01-12
Convert 3.x ESCN skeletons, animations, and shaders upon import

nikitalita:
Depends on #88009, #88078

## The Problem

ESCN was a discrete scene format that was created by the Godot team for ease-of-export from Blender to Godot. The plugin was originally written for Godot 3.x and used its scene format, and has not since been updated.

However, while Godot still has support for importing this format, ESCN importing has been broken since Godot 4.0 was released due significant changes in the scene format that conversion was not implemented for:

-  Skeletons had a `Transform pose` property which was relative to bone rest in 3.x , which was broken out to `Vector3 pose_position; Quaternion pose_rotation; Vector3 pose_scale` that are absolute in 4.x. 
-  Animations used to have a transform track type with transforms relative to bone rest in 3.x, which was similarly broken out into position_3d, rotation_3d, and scale types that are absolute in 4.x
-  The shaders that were included in the ESCN export were of the 3.x format, which is not compatible with 4.x.

## Solution

This adds the following features to the ESCN editor importer:

1) Converts 3.x Skeletons to use the new pose properties and recomputes them to be absolute rather than relative to the bone rest.
2) Converts 3.x Animations with `transform` tracks into position, rotation, and scale tracks that are absolute rather than relative to the bone rest.
3) Converts old shader code to 4.x-compliant shader code

## Design Considerations

The reason this is a draft right now is because of the following issues that I need advice on. This will likely be broken up into seperate PRs based on the feedback.

### ResourceLoader modification to allow special handlers for a single load

I had issues with converting Animations and Shaders; basically, we can't handle converting them in either of the `_set()` functions for these classes like we can for `Skeleton3D` because:

- Animation tracks are stored ini style:

  image

  This means that, upon load, [each of those _indexed_ properties gets set, one at a time](https://github.com/godotengine/godot/blob/master/scene/resources/animation.cpp#L37). We need to split the transformation track up into position, rotation, and scale tracks, and we can't arbitrarily add tracks in the middle of a load, since the index of the tracks will change.

- Shaders compile their code upon setting the `code` property; if the code fails to compile, `code` won't be set. Additionally, it will result in a very, very long error message that will obliterate the error logs and significantly confuse the user.

- Both of those class names are the same from 3.x to 4.x, so I couldn't implement another class for either of those to handle the loading either.

So, to solve these issues, I modified the ResourceLoader to allow the addition of special handlers. Basically, we can add a handler for a specified `Resource` class that takes in a `MissingResource` and returns a `Resource`. If we try to load a resource, and the resource loader detects that it has a handler for this resource class, it will instead load the resource as a `MissingResource` and, after the resource and its properties are finished loading, it will hand it off to the handler to instantiate the resource and set its properties.

I had considered modifying the ResourceLoader classes to just add the property to `missing_resource_properties` if it fails to be set and then just getting those properties out of the metadata, but as mentioned above, this would result in a bunch of error messages (1000+ lines in total length) every time we tried to import an ESCN, which is not ideal.

The way I have it now, I'm not really sure if this is the correct design; this sort of special handling needs to be only for a single load rather than a global handler, so I decided to only add it to the instantiated `ResourceLoader` classes rather than the static `ResourceFormatLoader` classes. This breaks the pattern a bit, as it looks like it's intended that the user doesn't use the instantiated `ResourceLoader` class and instead uses the `ResourceFormatLoader` class. I'd appreciate some advice on how to implement this such that it can be used for a single load and doesn't break the pattern here.

### ShaderLanguage token stream emitting

Converting Shaders from 3.x to 4.x is a bit more complicated than how it's currently handled in the 3.x to 4.x converter. In order to do a proper conversion of a 3.x to 4.x, we need to do the following:

- Replace everything in RenamesMap3To4::shaders_renames
- `SCREEN_TEXTURE`, `DEPTH_TEXTURE`, and `NORMAL_ROUGHNESS_TEXTURE` were removed, so their usage necessitates adding a uniform declaration at the top of the file
- If shader_type is "particles", we need to rename the function `void vertex()` to `void process()`
- `CLEARCOAT_GLOSS` was changed to `CLEARCOAT_ROUGHNESS` and its usage was inverted (i.e. ascending values now increase the roughness rather than decreasing it). So, we need to invert all usages of CLEARCOAT_GLOSS:
  - invert all lefthand assignments:
    - `CLEARCOAT_GLOSS = 5.0 / foo;`
      - becomes: `CLEARCOAT_ROUGHNESS = (1.0 - (5.0 / foo));`,
    - `CLEARCOAT_GLOSS *= 1.1;`
      - becomes `CLEARCOAT_ROUGHNESS = (1.0 - ((1.0 - CLEARCOAT_ROUGHNESS) * 1.1));`
  - invert all righthand usages
    - `foo = CLEARCOAT_GLOSS;`
      - becomes: `foo = (1.0 - CLEARCOAT_ROUGHNESS);`
- `async_visible` and `async_hidden` were removed, so these render modes need to be removed
-  `specular_blinn` and `specular_phong` render modes were removed, and we can't handle these, so we just throw an error
-  `MODULATE` [is not supported (yet)](https://github.com/godotengine/godot/issues/49781) in 4.x, so we have to throw an error.

As you can see, this is a bit more complicated than a simple regex search and replace would allow for, in particular the `CLEARCOAT_GLOSS` replacement. We at least need a token stream in order to properly detect for these conditions and modify the code.

So, in order to do this, I modified `ShaderLanguage` to emit a token stream:

- The function `token_debug_stream()` was added to emit a parsed token stream for the entire file.
- Token was modified to add `position` and `length` members so that we can emit code based on the token stream; this is neccesary primarily because the token parser can modify both identifiers and constant values before creating the token.
- The tokens `TK_TAB`, `TK_CR` ,`TK_SPACE`, `TK_NEWLINE`, `TK_BLOCK_COMMENT`, `TK_LINE_COMMENT`, and `TK_PREPROC_DIRECTIVE` were added and `_get_token()` was modified to emit them, but only if we are doing a debug parse. The reason for adding these was so that we can easily emit the new code from the token stream without obliterating the original formatting and comments

With these modifications, we're able to easily detect for the above conditions and insert and remove tokens from the token stream, and then emit code based on that token stream.

I'm a bit more confident about how I modified ShaderLanguage here but I'd still like to get feedback on how I modified ShaderLanguage here, and if there are any other 3.x to 4.x differences that I need to handle that I missed.

### DIFF
--- a/editor/import/3d/resource_importer_scene.cpp
+++ b/editor/import/3d/resource_importer_scene.cpp
@@ -3528,11 +3528,43 @@ void EditorSceneFormatImporterESCN::get_extensions(List<String> *r_extensions) c
 	r_extensions->push_back("escn");
 }
 
+int get_text_format_version(String p_path) {
+	Error error;
+	Ref<FileAccess> f = FileAccess::open(p_path, FileAccess::READ, &error);
+	ERR_FAIL_COND_V_MSG(error != OK || f.is_null(), -1, "Cannot open file '" + p_path + "'.");
+	String line = f->get_line().strip_edges();
+	// skip empty lines and comments
+	while (line.is_empty() || line.begins_with(";")) {
+		line = f->get_line().strip_edges();
+		if (f->eof_reached()) {
+			break;
+		}
+	}
+	int format_index = line.find("format");
+	ERR_FAIL_COND_V_MSG(format_index == -1, -1, "No format specifier in file '" + p_path + "'.");
+	String format_str = line.substr(format_index).get_slicec('=', 1).strip_edges();
+	ERR_FAIL_COND_V_MSG(!format_str.substr(0, 1).is_numeric(), -1, "Invalid format in file '" + p_path + "'.");
+	int format = format_str.to_int();
+	return format;
+}
+
 Node *EditorSceneFormatImporterESCN::import_scene(const String &p_path, uint32_t p_flags, const HashMap<StringName, Variant> &p_options, List<String> *r_missing_deps, Error *r_err) {
 	Error error;
+	int format = get_text_format_version(p_path);
+	ERR_FAIL_COND_V(format == -1, nullptr);
 	Ref<PackedScene> ps = ResourceFormatLoaderText::singleton->load(p_path, p_path, &error);
 	ERR_FAIL_COND_V_MSG(ps.is_null(), nullptr, "Cannot load scene as text resource from path '" + p_path + "'.");
 	Node *scene = ps->instantiate();
+	ERR_FAIL_COND_V(!scene, nullptr);
+	if (format == 2) {
+		TypedArray<Node> skel_nodes = scene->find_children("*", "AnimationPlayer");
+		for (int32_t node_i = 0; node_i < skel_nodes.size(); node_i++) {
+			// Force re-compute animation tracks.
+			AnimationPlayer *player = cast_to<AnimationPlayer>(skel_nodes[node_i]);
+			ERR_CONTINUE(!player);
+			player->advance(0);
+		}
+	}
 	TypedArray<Node> nodes = scene->find_children("*", "MeshInstance3D");
 	for (int32_t node_i = 0; node_i < nodes.size(); node_i++) {
 		MeshInstance3D *mesh_3d = cast_to<MeshInstance3D>(nodes[node_i]);
@@ -3541,9 +3573,22 @@ Node *EditorSceneFormatImporterESCN::import_scene(const String &p_path, uint32_t
 		// Ignore the aabb, it will be recomputed.
 		ImporterMeshInstance3D *importer_mesh_3d = memnew(ImporterMeshInstance3D);
 		importer_mesh_3d->set_name(mesh_3d->get_name());
-		importer_mesh_3d->set_transform(mesh_3d->get_relative_transform(mesh_3d->get_parent()));
-		importer_mesh_3d->set_skin(mesh_3d->get_skin());
+		Node *parent = mesh_3d->get_parent();
+		Transform3D rel_transform = mesh_3d->get_relative_transform(parent);
+		if (rel_transform == Transform3D() && parent && parent != mesh_3d) {
+			// If we're here, we probably got a "data.parent is null" error
+			// Node3D.data.parent hasn't been set yet but Node.data.parent has, so we need to get the transform manually
+			Node3D *parent_3d = mesh_3d->get_parent_node_3d();
+			if (parent == parent_3d) {
+				rel_transform = mesh_3d->get_transform();
+			} else if (parent_3d) {
+				rel_transform = parent_3d->get_relative_transform(parent) * mesh_3d->get_transform();
+			} // Otherwise, parent isn't a Node3D.
+		}
+		importer_mesh_3d->set_transform(rel_transform);
+		Ref<Skin> skin = mesh_3d->get_skin();
 		importer_mesh_3d->set_skeleton_path(mesh_3d->get_skeleton_path());
+		importer_mesh_3d->set_skin(skin);
 		Ref<ArrayMesh> array_mesh_3d_mesh = mesh_3d->get_mesh();
 		if (array_mesh_3d_mesh.is_valid()) {
 			// For the MeshInstance3D nodes, we need to convert the ArrayMesh to an ImporterMesh specially.
@@ -3583,8 +3628,6 @@ Node *EditorSceneFormatImporterESCN::import_scene(const String &p_path, uint32_t
 			continue;
 		}
 	}
-
-	ERR_FAIL_NULL_V(scene, nullptr);
 
 	return scene;
 }
--- a/editor/shader/text_shader_editor.cpp
+++ b/editor/shader/text_shader_editor.cpp
@@ -54,6 +54,9 @@
 #include "servers/rendering/shader_types.h"
 
 #include "modules/regex/regex.h"
+#ifndef DISABLE_DEPRECATED
+#include "servers/rendering/shader_converter.h"
+#endif
 
 /*** SHADER SYNTAX HIGHLIGHTER ****/
 
@@ -1353,6 +1356,28 @@ void TextShaderEditor::_menu_option(int p_option) {
 		case EDIT_COMPLETE: {
 			tx->request_code_completion();
 		} break;
+#ifndef DISABLE_DEPRECATED
+		case EDIT_CONVERT: {
+			if (shader.is_null()) {
+				return;
+			}
+			String code = code_editor->get_text_editor()->get_text();
+			if (code.is_empty()) {
+				return;
+			}
+			ShaderDeprecatedConverter converter;
+			if (!converter.is_code_deprecated(code)) {
+				if (converter.get_error_text() != String()) {
+					shader_convert_error_dialog->set_text(vformat(RTR("Line %d: %s"), converter.get_error_line(), converter.get_error_text()));
+					shader_convert_error_dialog->popup_centered();
+					ERR_PRINT("Shader conversion failed: " + converter.get_error_text());
+				}
+				confirm_convert_shader->popup_centered();
+				return;
+			}
+			_convert_shader();
+		} break;
+#endif
 		case SEARCH_FIND: {
 			code_editor->get_find_replace_bar()->popup_search();
 		} break;
@@ -1507,6 +1532,43 @@ void TextShaderEditor::_notification(int p_what) {
 		} break;
 	}
 }
+#ifndef DISABLE_DEPRECATED
+void TextShaderEditor::_convert_shader() {
+	if (shader.is_null()) {
+		return;
+	}
+	String code = code_editor->get_text_editor()->get_text();
+	if (code.is_empty()) {
+		return;
+	}
+	ShaderDeprecatedConverter converter;
+	if (!converter.convert_code(code)) {
+		String err_text = converter.get_error_text();
+		if (err_text.is_empty()) {
+			err_text = TTR("Unknown error occurred while converting the shader.");
+		} else if (converter.get_error_line() > 0) {
+			err_text = vformat("%s (line %d)", err_text, converter.get_error_line());
+		}
+
+		shader_convert_error_dialog->set_text(err_text);
+		shader_convert_error_dialog->popup_centered();
+		ERR_PRINT("Shader conversion failed: " + err_text);
+		return;
+	}
+	String new_code = converter.emit_code();
+
+#ifdef DEBUG_ENABLED
+	print_line(converter.get_report());
+#endif
+	if (new_code == code) {
+		return;
+	}
+	// Ensure undoable.
+	code_editor->get_text_editor()->set_text(new_code);
+	code_editor->get_text_editor()->tag_saved_version();
+	code_editor->_validate_script();
+}
+#endif
 
 void TextShaderEditor::_apply_editor_settings() {
 	code_editor->update_editor_settings();
@@ -2022,6 +2084,11 @@ TextShaderEditor::TextShaderEditor() {
 	edit_menu->get_popup()->add_shortcut(ED_GET_SHORTCUT("ui_text_completion_query"), EDIT_COMPLETE);
 	edit_menu->get_popup()->connect(SceneStringName(id_pressed), callable_mp(this, &TextShaderEditor::_menu_option));
 
+#ifndef DISABLE_DEPRECATED
+	edit_menu->get_popup()->add_separator();
+	edit_menu->get_popup()->add_item(TTR("Convert 3.x Shader"), EDIT_CONVERT);
+#endif
+
 	search_menu = memnew(MenuButton);
 	search_menu->set_flat(false);
 	search_menu->set_theme_type_variation("FlatMenuButton");
@@ -2155,6 +2222,20 @@ TextShaderEditor::TextShaderEditor() {
 	preview_timer->set_wait_time(0.001);
 	preview_timer->connect("timeout", callable_mp(code_editor, &ShaderTextEditor::redraw_preview_lines));
 
+#ifndef DISABLE_DEPRECATED
+	shader_convert_error_dialog = memnew(AcceptDialog);
+	shader_convert_error_dialog->set_title(TTR("Error converting shader"));
+	shader_convert_error_dialog->set_hide_on_ok(true);
+	add_child(shader_convert_error_dialog);
+
+	confirm_convert_shader = memnew(ConfirmationDialog);
+	confirm_convert_shader->set_title(TTR("Confirm Convert 3.x Shader"));
+	confirm_convert_shader->set_text(TTR("This shader does not appear to be a 3.x shader.\nAre you sure you want to convert it?"));
+	confirm_convert_shader->get_ok_button()->set_text(TTR("Convert"));
+	confirm_convert_shader->get_cancel_button()->set_text(TTR("Cancel"));
+	confirm_convert_shader->connect(SceneStringName(confirmed), callable_mp(this, &TextShaderEditor::_convert_shader));
+	add_child(confirm_convert_shader);
+#endif
 	_apply_editor_settings();
 	code_editor->show_toggle_files_button(); // TODO: Disabled for now, because it doesn't work properly.
 }
--- a/scene/3d/mesh_instance_3d.cpp
+++ b/scene/3d/mesh_instance_3d.cpp
@@ -70,8 +70,19 @@ bool MeshInstance3D::_set(const StringName &p_name, const Variant &p_value) {
 		return true;
 	}
 
-	if (p_name.operator String().begins_with("surface_material_override/")) {
-		int idx = p_name.operator String().get_slicec('/', 1).to_int();
+	String name = p_name;
+
+#ifndef DISABLE_DEPRECATED
+	if (name.begins_with("material/")) {
+		WARN_DEPRECATED_MSG("This mesh uses an old deprecated parameter name. Consider re-saving this scene in order for it to continue working in future versions." +
+				(is_inside_tree() ? vformat(" Path: \"%s\"", get_path()) : (get_name().is_empty() ? String() : vformat(" Name: \"%s\"", get_name()))));
+	}
+	if (name.begins_with("surface_material_override/") || name.begins_with("material/"))
+#else
+	if (name.begins_with("surface_material_override/"))
+#endif
+	{
+		int idx = name.get_slicec('/', 1).to_int();
 
 		if (idx >= surface_override_materials.size() || idx < 0) {
 			return false;
--- a/scene/animation/animation_mixer.cpp
+++ b/scene/animation/animation_mixer.cpp
@@ -257,6 +257,101 @@ TypedArray<StringName> AnimationMixer::_get_animation_library_list() const {
 	return ret;
 }
 
+#if !defined(_3D_DISABLED) && !defined(DISABLE_DEPRECATED)
+bool AnimationMixer::_recalc_animation(const Ref<Animation> &p_anim) {
+	HashMap<int, Vector<real_t>> new_track_values_map;
+	Node *parent = get_node_or_null(root_node);
+	if (!parent) {
+		return false;
+	}
+
+	for (int i = 0; i < p_anim->get_track_count(); i++) {
+		int track_type = p_anim->track_get_type(i);
+		if (!p_anim->track_is_relative_to_rest(i)) {
+			continue;
+		}
+		if (track_type == Animation::TYPE_POSITION_3D || track_type == Animation::TYPE_ROTATION_3D || track_type == Animation::TYPE_SCALE_3D) {
+			NodePath path = p_anim->track_get_path(i);
+			Node *node = parent->get_node(path);
+			ERR_FAIL_NULL_V(node, false);
+			Skeleton3D *skel = Object::cast_to<Skeleton3D>(node);
+			if (!skel) { // transforming non-skeleton node, not relative to rest
+				continue;
+			}
+			StringName bone = path.get_subname(0);
+			int bone_idx = skel->find_bone(bone);
+			if (bone_idx == -1) {
+				continue;
+			}
+			Transform3D rest = skel->get_bone_rest(bone_idx);
+			new_track_values_map[i] = Vector<real_t>();
+			const int32_t POSITION_TRACK_SIZE = 5;
+			const int32_t ROTATION_TRACK_SIZE = 6;
+			const int32_t SCALE_TRACK_SIZE = 5;
+			int32_t track_size = POSITION_TRACK_SIZE;
+			if (track_type == Animation::TYPE_ROTATION_3D) {
+				track_size = ROTATION_TRACK_SIZE;
+			}
+			new_track_values_map[i].resize(track_size * p_anim->track_get_key_count(i));
+			real_t *r = new_track_values_map[i].ptrw();
+			for (int j = 0; j < p_anim->track_get_key_count(i); j++) {
+				real_t time = p_anim->track_get_key_time(i, j);
+				real_t transition = p_anim->track_get_key_transition(i, j);
+				if (track_type == Animation::TYPE_POSITION_3D) {
+					Vector3 a_pos = p_anim->track_get_key_value(i, j);
+					Transform3D t;
+					t.set_origin(a_pos);
+					Vector3 new_a_pos = (rest * t).origin;
+
+					real_t *ofs = &r[j * POSITION_TRACK_SIZE];
+					ofs[0] = time;
+					ofs[1] = transition;
+					ofs[2] = new_a_pos.x;
+					ofs[3] = new_a_pos.y;
+					ofs[4] = new_a_pos.z;
+				} else if (track_type == Animation::TYPE_ROTATION_3D) {
+					Quaternion q = p_anim->track_get_key_value(i, j);
+					Transform3D t;
+					t.basis.rotate(q);
+					Quaternion new_q = (rest * t).basis.get_rotation_quaternion();
+					real_t *ofs = &r[j * ROTATION_TRACK_SIZE];
+					ofs[0] = time;
+					ofs[1] = transition;
+					ofs[2] = new_q.x;
+					ofs[3] = new_q.y;
+					ofs[4] = new_q.z;
+					ofs[5] = new_q.w;
+				} else if (track_type == Animation::TYPE_SCALE_3D) {
+					Vector3 v = p_anim->track_get_key_value(i, j);
+					Transform3D t;
+					t.scale(v);
+					Vector3 new_v = (rest * t).basis.get_scale();
+
+					real_t *ofs = &r[j * SCALE_TRACK_SIZE];
+					ofs[0] = time;
+					ofs[1] = transition;
+					ofs[2] = new_v.x;
+					ofs[3] = new_v.y;
+					ofs[4] = new_v.z;
+				}
+			}
+		}
+	}
+	if (new_track_values_map.is_empty()) {
+		return false;
+	}
+	for (int i = 0; i < p_anim->get_track_count(); i++) {
+		if (!new_track_values_map.has(i)) {
+			continue;
+		}
+		p_anim->set("tracks/" + itos(i) + "/keys", new_track_values_map[i]);
+		p_anim->set("tracks/" + itos(i) + "/relative_to_rest", false);
+	}
+	p_anim->emit_changed();
+	return true;
+}
+#endif // !defined(_3D_DISABLED) || !defined(DISABLE_DEPRECATED)
+
 void AnimationMixer::get_animation_library_list(LocalVector<StringName> *p_libraries) const {
 	for (const AnimationLibraryData &lib : animation_libraries) {
 		p_libraries->push_back(lib.name);
@@ -691,6 +786,11 @@ bool AnimationMixer::_update_caches() {
 	}
 	for (const StringName &E : sname_list) {
 		const Ref<Animation> &anim = get_animation(E);
+#if !defined(_3D_DISABLED) && !defined(DISABLE_DEPRECATED)
+		if (anim->has_tracks_relative_to_rest()) {
+			_recalc_animation(anim);
+		}
+#endif
 		for (int i = 0; i < anim->get_track_count(); i++) {
 			if (!anim->track_is_enabled(i)) {
 				continue;
--- a/scene/resources/animation.cpp
+++ b/scene/resources/animation.cpp
@@ -106,6 +106,18 @@ bool Animation::_set(const StringName &p_name, const Variant &p_value) {
 			} else if (type == "animation") {
 				add_track(TYPE_ANIMATION);
 			} else {
+#ifndef DISABLE_DEPRECATED
+				// for compatibility with 3.x animations
+				if (_started_load) { // Only do compatibility conversions if we are loading a resource.
+					if (type == "transform") {
+						WARN_DEPRECATED_MSG("Animation uses old 'transform' track types, which is deprecated (and loads slower). Consider re-importing or re-saving the resource.");
+						_transform_track_data[track] = { { { "type", "transform" } } };
+						// Add an empty track that will get replaced after we are finished loading, so we don't throw off the track indices.
+						add_track(TYPE_ANIMATION, track);
+						return true;
+					}
+				}
+#endif // DISABLE_DEPRECATED
 				return false;
 			}
 
@@ -113,6 +125,38 @@ bool Animation::_set(const StringName &p_name, const Variant &p_value) {
 		}
 
 		ERR_FAIL_UNSIGNED_INDEX_V((uint32_t)track, tracks.size(), false);
+
+#ifndef DISABLE_DEPRECATED
+		if (what == "relative_to_rest") {
+			Track *t = tracks[track];
+			switch (t->type) {
+				case TYPE_POSITION_3D: {
+					PositionTrack *tt = static_cast<PositionTrack *>(t);
+					tt->relative_to_rest = p_value;
+				} break;
+				case TYPE_ROTATION_3D: {
+					RotationTrack *rt = static_cast<RotationTrack *>(t);
+					rt->relative_to_rest = p_value;
+				} break;
+				case TYPE_SCALE_3D: {
+					ScaleTrack *st = static_cast<ScaleTrack *>(t);
+					st->relative_to_rest = p_value;
+				} break;
+				default: {
+					return false;
+				}
+			}
+			return true;
+		}
+		// If we have a transform track, we need to store the data in the metadata to be able to convert it to the new format after the load is finished.
+		if (_started_load) { // Only do this on resource loads, not on editor changes
+			// check the metadata to see if this track is a transform track that we are holding on to
+			if (_transform_track_data.has(track)) {
+				_transform_track_data[track][what] = p_value;
+				return true;
+			}
+		}
+#endif // DISABLE_DEPRECATED
 
 		if (what == "path") {
 			track_set_path(track, p_value);
@@ -882,6 +926,11 @@ void Animation::_get_property_list(List<PropertyInfo> *p_list) const {
 			p_list->push_back(PropertyInfo(Variant::INT, "tracks/" + itos(i) + "/interp", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_NO_EDITOR | PROPERTY_USAGE_INTERNAL));
 			p_list->push_back(PropertyInfo(Variant::BOOL, "tracks/" + itos(i) + "/loop_wrap", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_NO_EDITOR | PROPERTY_USAGE_INTERNAL));
 			p_list->push_back(PropertyInfo(Variant::ARRAY, "tracks/" + itos(i) + "/keys", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_NO_EDITOR | PROPERTY_USAGE_INTERNAL));
+#ifndef DISABLE_DEPRECATED
+			if (track_is_relative_to_rest(i)) {
+				p_list->push_back(PropertyInfo(Variant::ARRAY, "tracks/" + itos(i) + "/relative_to_rest", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_NO_EDITOR | PROPERTY_USAGE_INTERNAL));
+			}
+#endif // DISABLE_DEPRECATED
 		}
 		if (track_get_type(i) == TYPE_AUDIO) {
 			p_list->push_back(PropertyInfo(Variant::BOOL, "tracks/" + itos(i) + "/use_blend", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_NO_EDITOR | PROPERTY_USAGE_INTERNAL));
@@ -1135,6 +1184,63 @@ int Animation::_marker_insert(double p_time, LocalVector<MarkerKey> &p_keys, con
 }
 
 ////
+#ifndef DISABLE_DEPRECATED
+bool Animation::has_tracks_relative_to_rest() const {
+	for (uint32_t i = 0; i < tracks.size(); i++) {
+		if (track_is_relative_to_rest(i)) {
+			return true;
+		}
+	}
+	return false;
+}
+
+bool Animation::track_is_relative_to_rest(int p_track) const {
+	ERR_FAIL_UNSIGNED_INDEX_V((uint32_t)p_track, tracks.size(), false);
+	Track *t = tracks[p_track];
+
+	switch (t->type) {
+		case TYPE_POSITION_3D: {
+			PositionTrack *tt = static_cast<PositionTrack *>(t);
+			return tt->relative_to_rest;
+		} break;
+		case TYPE_ROTATION_3D: {
+			RotationTrack *rt = static_cast<RotationTrack *>(t);
+			return rt->relative_to_rest;
+		} break;
+		case TYPE_SCALE_3D: {
+			ScaleTrack *st = static_cast<ScaleTrack *>(t);
+			return st->relative_to_rest;
+		} break;
+		default: {
+			return false; // Animation does not really use transitions.
+		} break;
+	}
+}
+
+void Animation::track_set_relative_to_rest(int p_track, bool p_relative_to_rest) {
+	ERR_FAIL_UNSIGNED_INDEX((uint32_t)p_track, tracks.size());
+	Track *t = tracks[p_track];
+
+	switch (t->type) {
+		case TYPE_POSITION_3D: {
+			PositionTrack *tt = static_cast<PositionTrack *>(t);
+			tt->relative_to_rest = p_relative_to_rest;
+		} break;
+		case TYPE_ROTATION_3D: {
+			RotationTrack *rt = static_cast<RotationTrack *>(t);
+			rt->relative_to_rest = p_relative_to_rest;
+		} break;
+		case TYPE_SCALE_3D: {
+			ScaleTrack *st = static_cast<ScaleTrack *>(t);
+			st->relative_to_rest = p_relative_to_rest;
+		} break;
+		default: {
+			return; // Animation does not really use transitions.
+		} break;
+	}
+	emit_changed();
+}
+#endif // DISABLE_DEPRECATED
 
 int Animation::position_track_insert_key(int p_track, double p_time, const Vector3 &p_position) {
 	ERR_FAIL_UNSIGNED_INDEX_V((uint32_t)p_track, tracks.size(), -1);
@@ -5262,6 +5368,112 @@ void Animation::compress(uint32_t p_page_size, uint32_t p_fps, float p_split_tol
 
 	print_line("Original size: " + itos(orig_size) + " - Compressed size: " + itos(new_size) + " " + String::num(float(new_size) / float(orig_size) * 100, 2) + "% pages: " + itos(compression.pages.size()));
 #endif
+}
+
+void Animation::_start_load(const StringName &p_res_format_type, int p_res_format_version) {
+#ifndef DISABLE_DEPRECATED
+	_started_load = true;
+#endif
+}
+
+void Animation::_finish_load(const StringName &p_res_format_type, int p_res_format_version) {
+#ifndef DISABLE_DEPRECATED // 3.x compatibility, convert transform tracks to separate tracks
+	if (!_started_load) {
+		return;
+	}
+	_started_load = false;
+	int offset = 0;
+	for (const KeyValue<int, Dictionary> &E : _transform_track_data) {
+		int track_idx = E.key;
+		// now that we have all the tracks, we need to split the transform tracks into separate tracks
+		// this is because the current animation player doesn't support transform tracks
+		// so we need to split them into separate position, rotation, and scale tracks
+		// No need to worry about compression here; this was added in 4.x and wasn't backported to 3.x.
+		const Dictionary &track_data = E.value;
+		if (track_data.is_empty()) {
+			continue;
+		}
+		// split the transform track into separate tracks
+
+		// Old scene format only used 32-bit floats, did not have configurable real_t.
+		Vector<float> keys = track_data["keys"];
+		int vcount = keys.size();
+		int tcount = vcount / 12;
+		ERR_CONTINUE_MSG((vcount % 12) != 0, "Failed to convert transform track: invalid number of key frames.");
+
+		Vector<real_t> position_keys;
+		Vector<real_t> rotation_keys;
+		Vector<real_t> scale_keys;
+		position_keys.resize(tcount * 5); // time + transition + xyz
+		rotation_keys.resize(tcount * 6); // time + transition + xyzw
+		scale_keys.resize(tcount * 5); // time + transition + xyz
+		// split the keys into separate tracks
+		for (int j = 0; j < tcount; j++) {
+			// it's position (Vector3, xyz), then rotation (Quaternion, xyzw), then scale (Vector3, xyz)
+			// each track has time and transition values, so get those
+			const float *ofs = &(keys.ptr()[j * 12]);
+			float time = ofs[0];
+			float transition = ofs[1];
+
+			position_keys.write[j * 5 + 0] = time;
+			position_keys.write[j * 5 + 1] = transition;
+			position_keys.write[j * 5 + 2] = ofs[2]; // x
+			position_keys.write[j * 5 + 3] = ofs[3]; // y
+			position_keys.write[j * 5 + 4] = ofs[4]; // z
+
+			rotation_keys.write[j * 6 + 0] = time;
+			rotation_keys.write[j * 6 + 1] = transition;
+			rotation_keys.write[j * 6 + 2] = ofs[5]; // x
+			rotation_keys.write[j * 6 + 3] = ofs[6]; // y
+			rotation_keys.write[j * 6 + 4] = ofs[7]; // z
+			rotation_keys.write[j * 6 + 5] = ofs[8]; // w
+
+			scale_keys.write[j * 5 + 0] = time;
+			scale_keys.write[j * 5 + 1] = transition;
+			scale_keys.write[j * 5 + 2] = ofs[9]; // x
+			scale_keys.write[j * 5 + 3] = ofs[10]; // y
+			scale_keys.write[j * 5 + 4] = ofs[11]; // z
+		}
+		Array c_track_keys = track_data.keys();
+		c_track_keys.erase("type");
+		c_track_keys.erase("keys");
+		remove_track(track_idx + offset); // remove dummy track
+		NodePath node_path;
+		add_track(TYPE_POSITION_3D, track_idx + offset);
+		for (int j = 0; j < c_track_keys.size(); j++) {
+			String key = c_track_keys[j];
+			const Variant &value = track_data[key];
+			if (key == "path") {
+				node_path = value;
+			}
+			_set("tracks/" + itos(track_idx + offset) + "/" + key, value);
+		}
+		bool is_bone_transform = node_path.get_subname_count() != 0;
+		_set("tracks/" + itos(track_idx + offset) + "/keys", position_keys);
+		_set("tracks/" + itos(track_idx + offset) + "/relative_to_rest", is_bone_transform);
+		offset++;
+
+		add_track(TYPE_ROTATION_3D, track_idx + offset);
+		for (int j = 0; j < c_track_keys.size(); j++) {
+			String key = c_track_keys[j];
+			_set("tracks/" + itos(track_idx + offset) + "/" + key, track_data[key]);
+		}
+		_set("tracks/" + itos(track_idx + offset) + "/keys", rotation_keys);
+		_set("tracks/" + itos(track_idx + offset) + "/relative_to_rest", is_bone_transform);
+		offset++;
+		add_track(TYPE_SCALE_3D, track_idx + offset);
+		for (int j = 0; j < c_track_keys.size(); j++) {
+			String key = c_track_keys[j];
+			_set("tracks/" + itos(track_idx + offset) + "/" + key, track_data[key]);
+		}
+		_set("tracks/" + itos(track_idx + offset) + "/keys", scale_keys);
+		_set("tracks/" + itos(track_idx + offset) + "/relative_to_rest", is_bone_transform);
+		offset++;
+		offset--; // subtract 1 because we removed the dummy track
+	}
+	// erase the track list
+	_transform_track_data.clear();
+#endif // DISABLE_DEPRECATED
 }
 
 bool Animation::_rotation_interpolate_compressed(uint32_t p_compressed_track, double p_time, Quaternion &r_ret) const {
--- a/scene/resources/animation.h
+++ b/scene/resources/animation.h
@@ -144,6 +144,9 @@ private:
 	struct PositionTrack : public Track {
 		LocalVector<TKey<Vector3>> positions;
 		int32_t compressed_track = -1;
+#ifndef DISABLE_DEPRECATED
+		bool relative_to_rest = false;
+#endif
 		PositionTrack() { type = TYPE_POSITION_3D; }
 	};
 
@@ -152,6 +155,9 @@ private:
 	struct RotationTrack : public Track {
 		LocalVector<TKey<Quaternion>> rotations;
 		int32_t compressed_track = -1;
+#ifndef DISABLE_DEPRECATED
+		bool relative_to_rest = false;
+#endif
 		RotationTrack() { type = TYPE_ROTATION_3D; }
 	};
 
@@ -160,6 +166,9 @@ private:
 	struct ScaleTrack : public Track {
 		LocalVector<TKey<Vector3>> scales;
 		int32_t compressed_track = -1;
+#ifndef DISABLE_DEPRECATED
+		bool relative_to_rest = false;
+#endif
 		ScaleTrack() { type = TYPE_SCALE_3D; }
 	};
 
@@ -293,6 +302,13 @@ private:
 	LoopMode loop_mode = LOOP_NONE;
 	bool capture_included = false;
 	void _check_capture_included();
+
+	void _track_update_hash(int p_track);
+
+#ifndef DISABLE_DEPRECATED
+	bool _started_load = false;
+	HashMap<int, Dictionary> _transform_track_data;
+#endif
 
 	/* Animation compression page format (version 1):
 	 *
@@ -453,7 +469,11 @@ public:
 	double track_get_key_time(int p_track, int p_key_idx) const;
 	real_t track_get_key_transition(int p_track, int p_key_idx) const;
 	bool track_is_compressed(int p_track) const;
-
+#ifndef DISABLE_DEPRECATED
+	bool has_tracks_relative_to_rest() const;
+	bool track_is_relative_to_rest(int p_track) const;
+	void track_set_relative_to_rest(int p_track, bool p_relative_to_rest);
+#endif
 	int position_track_insert_key(int p_track, double p_time, const Vector3 &p_position);
 	Error position_track_get_key(int p_track, int p_key, Vector3 *r_position) const;
 	Error try_position_track_interpolate(int p_track, double p_time, Vector3 *r_interpolation, bool p_backward = false) const;
@@ -561,6 +581,9 @@ public:
 		}
 	}
 #endif // TOOLS_ENABLED
+
+	virtual void _start_load(const StringName &p_res_format_type, int p_res_format_version) override;
+	virtual void _finish_load(const StringName &p_res_format_type, int p_res_format_version) override;
 
 	// Helper functions for Rotation.
 	static double interpolate_via_rest(double p_from, double p_to, double p_weight, double p_rest = 0.0); // Deterministic slerp to prevent to cross the inverted rest axis.
--- a/servers/rendering/shader_converter.cpp
+++ b/servers/rendering/shader_converter.cpp
@@ -1,0 +1,3045 @@
+/**************************************************************************/
+/*  shader_converter.cpp                                                  */
+/**************************************************************************/
+/*                         This file is part of:                          */
+/*                             GODOT ENGINE                               */
+/*                        https://godotengine.org                         */
+/**************************************************************************/
+/* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+/* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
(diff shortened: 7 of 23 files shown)


AThousandShips:
These are several unrelated improvements, please make a PR for each feature

nikitalita:
> These are several unrelated improvements, please make a PR for each feature

That's the plan, they're not going to stay in this PR. This is just my working branch so I can get feedback on specific design questions.

AThousandShips:
Thats not really how you should do things, each issue or feature should be discussed separately, and you shouldn't open a PR for your working branch but work on each feature on one branch unless they're dependent on each other 🙂, this just takes up processing and duplicates things

nikitalita:
Ok, I have removed the extraneous commits and specified which PRs this depends on.

editor/import/3d/resource_importer_scene.cpp
```diff
@@ -2734,6 +2740,998 @@ void ResourceImporterScene::get_scene_importer_extensions(List<String> *p_extens
...
+		}
+
+		return new_code;
+	}
+
+	TokenStreamManip() = delete;
+	TokenStreamManip(const String &p_code) :
+			old_code(p_code) {
+		ShaderLanguage sl;
+		sl.token_debug_stream(old_code, code_tokens, true);
```
nikitalita, line 3415: token_debug_stream

lyuma:
I think this adds too much complexity to the import system which is not really the point of the importer.

If the goal is to add back compat with 3.x resources, we should do it one of two ways:
1. expose them with a conversion utility (and perhaps the escn importer could invoke this conversion utility code automatically, but this way it isn't adding complexity to the import system itself). This code will have a maintenance cost to maintain.
2. fix data in the resource loader or the property setters.

My personal opinion is that if this is to be done automatically, it should be done in the resources itself, which you attempted to explain why it wasn't possible, but I think it should be made to be possible. For example, we have discussed a compatibility system. We know which Godot version the resources originated from, so perhaps we can add the missing parts to the resource loader so that we can adapt the properties. These would be a good usecase for such a compatibility system.

> Animation tracks are stored ini style:

> This means that, upon load, [each of those indexed properties gets set, one at a time](https://github.com/godotengine/godot/blob/master/scene/resources/animation.cpp#L37). We need to split the transformation track up into position, rotation, and scale tracks, and we can't arbitrarily add tracks in the middle of a load, since the index of the tracks will change.

Yes, but you will know how much the index changes: you can store an offset that starts at 0 and increases by 2 for every transform track. In principle, the indices could be offset as the properties are set. It certainly is challenging and may be impossible without some changes to the core resource loader, but I do think we should be willing to look outside of the box rather than adding a lot of complexity to the import system.


> Shaders compile their code upon setting the code property; if the code fails to compile, code won't be set. Additionally, it will result in a very, very long error message that will obliterate the error logs and significantly confuse the user.

What if when the code fails to compile, it could attempt to run it through the above converter and use that converted code if it succeeds?

> Both of those class names are the same from 3.x to 4.x, so I couldn't implement another class for either of those to handle the loading either.

Yeah I don't think we want to add compatibility code especially at runtime. Load-time or editor-time conversion could be supported only if TOOLS_ENABLED, for example.

Do note also that if we do go this route, we need to make sure mesh compatibility issues are solved .

related issues:
* #87050 will fix skeletons upgraded on 3.x
* #63550 meshes are broken on upgrade from 3.x - I think this would be a blocker. Do we need to solve this as well?

nikitalita:
So what we talked about in the rocketchat was that the best course of action for modifying the resource loading to accommodate this would be to instead add `void _start_load(int format, bool binary)` and `void _finish_load(int format, bool binary)` virtual methods (that are NOPs by default) to `Resource` to accommodate this. the binary and text resource loaders would call `_start_load()` when the sub/main resource is instantiated, before any properties are loaded, and `_finish_load()` after all the properties have been read and set on the instanced resource.

For `Animation`, the main problem is that the `transform` tracks need to be split into seperate tracks, and all the properties for the tracks are set one at a time.

So for `Animation`, the steps would be:

1) if `_start_load()` is called, set a flag in the metadata
2) in `_set()` in Animation.cpp, if the `_start_load()` flag is set and if we read in a track with a `transform` type, store it and all of its subsequent properties in the resource metadata
3) once `_finish_load()` is called, get those tracks out of the metadata, and then split them into `position_3d`, `rotation_3d`, and `scale_3d` tracks, and then clear the metadata of the flag and the `transform` track data.

This would still entail some steps in the scene importer; we would still have to recompute the animation frames to be absolute rather than relative to the skeleton rest, but we already do plenty of that in the scene importer for other scene types.

For Shader, the main issues are:
 1. the code is compiled as soon as the `code` property is set, which does not allow for fixing it after the fact,  
 2. failed compilations will spit out 1000+ line error messages
 3. We don't want to auto-convert any shaders that aren't actually from Godot 3.x. 

For `Shader`, the steps would be:

1) If `_start_load()` is called, AND the `format` value is binary format 3 or text format 2 (i.e. the resource format versions used for all Godot 3.x releases), then we set a flag in the metadata
2) in `set_code()` in Shader.cpp, if the flag is set:
   1) instead of compiling right away, first run the code through the ShaderLanguage parser (this is done to avoid printing the huge error messages that accompany failed shader compiles)
   2) If the ShaderLanguage parser fails, THEN we do some additional tests to see if it is in fact a 3.x shader (i.e. checking for old identifiers)
   3) If our tests indiciate that it is a 3.x shader, we run it through the code converter 
   4) If the code conversion succeeds, then we compile and set the code property.

So, these steps would only be run on Shaders that are embeded in other resources (i.e. would not automatically be run for standalone `gdshader` files), and would only be run if the resource is in a deprecated format. This will prevent the "Unity shader auto-upgrade" shenanigans that @lyuma expressed their wishes in avoiding.

servers/rendering/shader_converter.h
```diff
@@ -0,0 +1,249 @@
...
+	bool fail_on_unported = true;
+
+	bool function_pass_failed = false;
+	bool var_pass_failed = false;
+	String err_str;
+	int err_line = 0;
+	Token eof_token{ ShaderLanguage::TK_EOF, {}, 0, 0, 0, 0 };
+	static RS::ShaderMode get_shader_mode_from_string(const String &p_mode);
+
+	String get_token_literal_text(const Token &tk) const;
```
AThousandShips, line 212:
```suggestion
	String get_token_literal_text(const Token &p_tk) const;
```

servers/rendering/shader_converter.h
```diff
@@ -0,0 +1,249 @@
...
+	static Token mkTok(TokenType p_type, const StringName &p_text = StringName(), double constant = 0, uint16_t p_line = 0);
+
+	void reset();
+	bool preprocess_code();
+	List<Token>::Element *get_next_token();
+	List<Token>::Element *get_prev_token();
+	List<Token>::Element *remove_cur_and_get_next();
+	TokenType peek_next_tk_type(uint32_t count = 1) const;
+	TokenType peek_prev_tk_type(uint32_t count = 1) const;
```
AThousandShips, line 221:
```suggestion
	TokenType peek_next_tk_type(uint32_t p_count = 1) const;
	TokenType peek_prev_tk_type(uint32_t p_count = 1) const;
```

servers/rendering/shader_converter.h
```diff
@@ -0,0 +1,249 @@
...
+	List<Token>::Element *get_pos() const;
+	bool reset_to(List<Token>::Element *p_pos);
+	bool insert_after(const Vector<Token> &token_list, List<Token>::Element *p_pos);
+	bool insert_before(const Vector<Token> &token_list, List<Token>::Element *p_pos);
+	bool insert_after(const Token &token, List<Token>::Element *p_pos);
+	bool insert_before(const Token &token, List<Token>::Element *p_pos);
+	List<Token>::Element *replace_curr(const Token &token);
+	List<Token>::Element *_get_next_token_ptr(List<Token>::Element *_curr_ptr) const;
+	List<Token>::Element *_get_prev_token_ptr(List<Token>::Element *_curr_ptr) const;
+	TokenType _peek_tk_type(int64_t count, List<Token>::Element **r_pos = nullptr) const;
```
AThousandShips, line 231:
```suggestion
	bool insert_after(const Vector<Token> &p_token_list, List<Token>::Element *p_pos);
	bool insert_before(const Vector<Token> &p_token_list, List<Token>::Element *p_pos);
	bool insert_after(const Token &p_token, List<Token>::Element *p_pos);
	bool insert_before(const Token &p_token, List<Token>::Element *p_pos);
	List<Token>::Element *replace_curr(const Token &p_token);
	List<Token>::Element *_get_next_token_ptr(List<Token>::Element *p_curr_ptr) const;
	List<Token>::Element *_get_prev_token_ptr(List<Token>::Element *p_curr_ptr) const;
	TokenType _peek_tk_type(int64_t p_count, List<Token>::Element **r_pos = nullptr) const;
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2231 @@
...
+	}
+	return {};
+}
+
+bool ShaderDeprecatedConverter::has_builtin_rename(RS::ShaderMode p_mode, const String &p_name, const String &p_function) {
+	for (int i = 0; renamed_builtins[i].name != nullptr; i++) {
+		if (renamed_builtins[i].name == p_name) {
+			for (int j = 0; j < renamed_builtins[i].mode_functions.size(); j++) {
+				if (renamed_builtins[i].mode_functions[j].first == p_mode) {
+					if (p_function == "") { // Empty function means don't check function.
```
AThousandShips, line 222:
```suggestion
					if (p_function.is_empty()) { // Empty function means don't check function.
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2231 @@
...
+	return SL::TK_EMPTY;
+}
+
+Vector<SL::TokenType> ShaderDeprecatedConverter::get_removed_builtin_hints(const String &p_name) {
+	for (int i = 0; removed_builtins[i].name != nullptr; i++) {
+		if (removed_builtins[i].name == p_name) {
+			return removed_builtins[i].hints;
+		}
+	}
+	return {};
```
AThousandShips, line 252:
```suggestion
	return Vector<SL::TokenType>();
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2231 @@
...
+	return SL::TK_MAX;
+}
+
+String ShaderDeprecatedConverter::get_renamed_function(const String &p_name) {
+	for (int i = 0; renamed_functions[i].name != nullptr; i++) {
+		if (renamed_functions[i].name == p_name) {
+			return renamed_functions[i].replacement;
+		}
+	}
+	return {};
```
AThousandShips, line 418:
```suggestion
	return String();
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2231 @@
...
+		case ShaderLanguage::TK_NEWLINE:
+		case ShaderLanguage::TK_BLOCK_COMMENT:
+		case ShaderLanguage::TK_LINE_COMMENT:
+		case ShaderLanguage::TK_PREPROC_DIRECTIVE:
+			return true;
+		default:
+			break;
+	}
+	return false;
+};
```
AThousandShips, line 612:
```suggestion
}
```
For all these

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2231 @@
...
+	return false;
+#define INSERT_FAIL(insert_ret)                          \
+	if (unlikely(!insert_ret)) {                         \
+		err_str = "Internal error: Token insert failed"; \
+		return false;                                    \
+	}
+
+#define EOF_FAIL(tokE)                                            \
+	COND_MSG_FAIL(tokE == nullptr, RTR("Unexpected end of file")) \
+	COND_LINE_MSG_FAIL(tokE->get().type == TT::TK_EOF || tokE->get().type == TT::TK_ERROR, tokE->get().line, tokE->get().type == TT::TK_ERROR ? vformat(RTR("Parser Error (%s) ", tokE->get().text)) : RTR("Unexpected end of file"))
```
AThousandShips, line 1141:
```suggestion
#define COND_MSG_FAIL(m_cond, m_msg) \
	if (unlikely(m_cond)) {          \
		err_str = m_msg;             \
		return false;                \
	}
#define COND_LINE_MSG_FAIL(m_cond, m_line, m_msg) \
	if (unlikely(m_cond)) {                       \
		err_line = line + 1;                      \
		err_str = m_msg;                          \
		return false;                             \
	}
#define LINE_MSG_FAIL(m_line, m_msg) \
	err_line = m_line + 1;           \
	err_str = m_msg;                 \
	return false;
#define MSG_FAIL(m_msg)    \
	err_line = line + 1;   \
	err_str = m_msg;       \
	return false;
#define INSERT_FAIL(m_insert_ret)                          \
	if (unlikely(!m_insert_ret)) {                         \
		err_str = "Internal error: Token insert failed";   \
		return false;                                      \
	}

#define EOF_FAIL(m_tok_E)                                             \
	COND_MSG_FAIL(m_tok_E == nullptr, RTR("Unexpected end of file")); \
	COND_LINE_MSG_FAIL(m_tok_E->get().type == TT::TK_EOF || m_tok_E->get().type == TT::TK_ERROR, m_tok_E->get().line, m_tok_E->get().type == TT::TK_ERROR ? vformat(RTR("Parser Error (%s) ", m_tok_E->get().text)) : RTR("Unexpected end of file"));
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2231 @@
...
+	}
+	after_type_decl = get_pos();
+
+	/***
+	 * The first pass gets the uniform declarations; we require this is to ensure idempotency for inserting new uniforms and replacing type hints.
+	 * The second pass gets the function declarations; these are used for determining if a renamed built-in is valid in the current scope
+	 * The third pass gets the variable declarations; These are used for determining if renamed built-ins have been previously declared, and for detecting new keywords used as identifiers.
+	 */
+	// first pass; we just get the Uniform declarations
+	auto skip_array_size = [&]() -> bool {
```
AThousandShips, line 1190: We don't use lambdas unless necessary

scene/3d/mesh_instance_3d.cpp
```diff
@@ -49,8 +49,19 @@ bool MeshInstance3D::_set(const StringName &p_name, const Variant &p_value) {
...
+#ifndef DISABLE_DEPRECATED
+	if (name.begins_with("material/")) {
+		WARN_DEPRECATED_MSG("This mesh uses an old deprecated parameter name. Consider re-saving this scene in order for it to continue working in future versions." +
+				(is_inside_tree() ? vformat(" Path: \"%s\"", get_path()) : String()));
+	}
+	if (name.begins_with("surface_material_override/") || name.begins_with("material/"))
+#else
+	if (name.begins_with("surface_material_override/"))
+#endif
+	{
```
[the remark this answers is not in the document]
nikitalita: I intentionally did not do it this way because this screws up code folding and formatting in VS Code and other IDEs.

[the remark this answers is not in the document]
AThousandShips: That's fair :) Might need to make that clear though, or make the if statement partially part of the deprecated part, like so:
```suggestion
#endif
	if (name.begins_with("surface_material_override/")
#ifndef DISABLE_DEPRECATED
			|| name.begins_with("material/")
#endif
		) {
```
Which is done elsewhere

nikitalita:
depends on #88078

Zireael07:
I have a project that contains ESCN, that needs this PR to make it in....

nikitalita:
grab one of the editor artifacts from the build jobs, import it, and then save it as a godot 4 scene :)

fire (commented):
I'm under the weather, but I want some progress on this.

Some of the core io apis changed and I can't figure if they're necessary.

The changes are a bit invasive.

@clayjohn probably has opinions about the shader_converter.cpp

fire:
There are some build errors.

nikitalita:
sorry, I fatfingered the delete button when I rebased

Zireael07:
Bump. This still hasn't made it in and I still have a project that has ESCN (downloaded, not created by me so I don't have the sources)

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2591 @@
...
+	return false;
+}
+
+String ShaderDeprecatedConverter::get_render_mode_rename(const String &p_name) {
+	for (int i = 0; renamed_render_modes[i].name != nullptr; i++) {
+		if (renamed_render_modes[i].name == p_name) {
+			return renamed_render_modes[i].replacement;
+		}
+	}
+	return {};
```
AThousandShips:
```suggestion
	return String();
```
Please be explicit about return types

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2591 @@
...
+		case ShaderLanguage::TK_PREPROC_DIRECTIVE:
+			return true;
+		default:
+			break;
+	}
+	return false;
+}
+
+List<SL::Token>::Element *ShaderDeprecatedConverter::_get_next_token_ptr(List<Token>::Element *p_curr_ptr) const {
+	ERR_FAIL_COND_V(p_curr_ptr == nullptr, p_curr_ptr);
```
AThousandShips:
```suggestion
	ERR_FAIL_NULL_V(p_curr_ptr, p_curr_ptr);
```
For all these

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2591 @@
...
+		return RS::SHADER_SPATIAL;
+	} else if (p_mode == "canvas_item") {
+		return RS::SHADER_CANVAS_ITEM;
+	} else if (p_mode == "particles") {
+		return RS::SHADER_PARTICLES;
+	} else { // 3.x didn't support anything else.
+		return RS::SHADER_MAX;
+	}
+}
+// Remove from the current token to end (exclsusive) and return the new current token.
```
AThousandShips:
```suggestion

// Remove from the current token to end (exclsusive) and return the new current token.
```
Please leave an empty line between methods

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2591 @@
...
+#define LINE_MSG_FAIL(m_line, m_msg) \
+	err_line = m_line + 1;           \
+	err_str = m_msg;                 \
+	return false;
+#define MSG_FAIL(m_msg) \
+	err_str = m_msg;    \
+	return false;
+
+#define EOF_FAIL(m_tok_E)                                             \
+	COND_MSG_FAIL(m_tok_E == nullptr, RTR("Unexpected end of file")); \
```
AThousandShips:
```suggestion
	COND_MSG_FAIL(m_tok_E == nullptr, RTR("Unexpected end of file.")); \
```
For all the custom errors, errors should be complete sentences and end with a period

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2591 @@
...
+
+ShaderDeprecatedConverter::TokenE *ShaderDeprecatedConverter::_rename_keyword_id(TokenE *pos, bool p_detected_3x, const HashMap<TokenType, String> &p_new_reserved_word_renames) {
+	String rename = p_new_reserved_word_renames[pos->get().type];
+	reset_to(pos);
+	return replace_curr(mkTok(TT::TK_IDENTIFIER, rename), "Identifier '%s' is a reserved word in this version of Godot, renamed to '%s'.");
+}
+
+bool ShaderDeprecatedConverter::_handle_new_keyword_rename(TokenE *pos, bool p_detected_3x, HashMap<TokenType, String> &p_new_reserved_word_renames) {
+	TokenType tk_type = pos->get().type;
+	String p_name = get_token_literal_text(pos->get());
```
AThousandShips:
```suggestion
	String name = get_token_literal_text(pos->get());
```
Or something similar, `p_` is reserved for arguments

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2591 @@
...
+		}
+	}
+	return true;
+}
+
+String ShaderDeprecatedConverter::emit_code() const {
+	if (code_tokens.size() == 0) {
+		return "";
+	}
+	String new_code = "";
```
AThousandShips:
```suggestion
	String new_code;
```

fire:
Do you have a series of 3.x ESCNs so I can make a test suite for testing this pr?

nikitalita:
[Character.escn.zip](https://github.com/user-attachments/files/23401513/Character.escn.zip)
Here are a few that should be a good start

fire (approved):
I am not able to quality assurance test or do a technical review. I honestly believe that @nikitalita should get this work merged before Godot Engine 3 usage is minimized as we're almost 4.6.

My focus is trying to get the animation team work and 3d pipeline team work into 4.6 which the has a freeze date by the end of November.

@nikitalita is the maintainer of https://github.com/GDRETools/gdsdecomp so there is an expectation that she can help fix reported bugs with the software.

Cheers when this gets merged.

fire:
Needs a rebase.

scene/animation/animation_mixer.cpp
```diff
@@ -244,6 +244,97 @@ TypedArray<StringName> AnimationMixer::_get_animation_library_list() const {
...
+				track_size = ROTATION_TRACK_SIZE;
+			}
+			new_track_values_map[i].resize(track_size * anim->track_get_key_count(i));
+			real_t *r = new_track_values_map[i].ptrw();
+			for (int j = 0; j < anim->track_get_key_count(i); j++) {
+				real_t time = anim->track_get_key_time(i, j);
+				real_t transition = anim->track_get_key_transition(i, j);
+				if (track_type == Animation::TYPE_POSITION_3D) {
+					Vector3 a_pos = anim->track_get_key_value(i, j);
+					Transform3D t = Transform3D();
```
AThousandShips:
```suggestion
					Transform3D t;
```
Unnecessary default

scene/animation/animation_mixer.cpp
```diff
@@ -244,6 +244,97 @@ TypedArray<StringName> AnimationMixer::_get_animation_library_list() const {
...
+
+					real_t *ofs = &r[j * POSITION_TRACK_SIZE];
+					ofs[0] = time;
+					ofs[1] = transition;
+					ofs[2] = new_a_pos.x;
+					ofs[3] = new_a_pos.y;
+					ofs[4] = new_a_pos.z;
+				} else if (track_type == Animation::TYPE_ROTATION_3D) {
+					Quaternion q = anim->track_get_key_value(i, j);
+					Transform3D t = Transform3D();
```
AThousandShips:
```suggestion
					Transform3D t;
```

scene/animation/animation_mixer.cpp
```diff
@@ -244,6 +244,97 @@ TypedArray<StringName> AnimationMixer::_get_animation_library_list() const {
...
+					real_t *ofs = &r[j * ROTATION_TRACK_SIZE];
+					ofs[0] = time;
+					ofs[1] = transition;
+					ofs[2] = new_q.x;
+					ofs[3] = new_q.y;
+					ofs[4] = new_q.z;
+					ofs[5] = new_q.w;
+				} else if (track_type == Animation::TYPE_SCALE_3D) {
+					Vector3 v = anim->track_get_key_value(i, j);
+					Transform3D t = Transform3D();
```
AThousandShips:
```suggestion
					Transform3D t;
```

servers/rendering/shader_converter.h
```diff
@@ -0,0 +1,310 @@
...
+	int err_line = 0;
+
+	Token eof_token{ ShaderLanguage::TK_EOF, {}, 0, 0, 0, 0 };
+
+	HashMap<int, Vector<String>> report;
+
+	static RS::ShaderMode get_shader_mode_from_string(const String &p_mode);
+
+	String get_token_literal_text(const Token &p_tk) const;
+	static Token mkTok(TokenType p_type, const StringName &p_text = StringName(), double constant = 0, uint16_t p_line = 0);
```
AThousandShips:
```suggestion
	static Token mk_tok(TokenType p_type, const StringName &p_text = StringName(), double p_constant = 0.0, uint16_t p_line = 0);
```

servers/rendering/shader_converter.h
```diff
@@ -0,0 +1,310 @@
...
+	List<Token>::Element *remove_cur_and_get_next();
+	TokenType peek_next_tk_type(uint32_t p_count = 1) const;
+	TokenType peek_prev_tk_type(uint32_t p_count = 1) const;
+	List<Token>::Element *get_pos() const;
+	bool reset_to(List<Token>::Element *p_pos);
+	bool insert_after(const Vector<Token> &p_token_list, List<Token>::Element *p_pos);
+	bool insert_before(const Vector<Token> &p_token_list, List<Token>::Element *p_pos);
+	bool insert_after(const Token &p_token, List<Token>::Element *p_pos);
+	bool insert_before(const Token &p_token, List<Token>::Element *p_pos);
+	List<Token>::Element *replace_curr(const Token &p_token, String comment_prefix = String());
```
AThousandShips:
```suggestion
	List<Token>::Element *replace_curr(const Token &p_token, const String &p_comment_prefix = String());
```

servers/rendering/shader_converter.h
```diff
@@ -0,0 +1,310 @@
...
+	List<Token>::Element *_get_next_token_ptr(List<Token>::Element *p_curr_ptr) const;
+	List<Token>::Element *_get_prev_token_ptr(List<Token>::Element *p_curr_ptr) const;
+	TokenType _peek_tk_type(int64_t p_count, List<Token>::Element **r_pos = nullptr) const;
+
+	bool scope_has_decl(const String &p_scope, const String &p_name) const;
+	bool _handle_new_keyword_rename(TokenE *pos, bool p_detected_3x, HashMap<TokenType, String> &p_func_renames);
+	TokenE *_rename_keyword_id(TokenE *pos, bool p_detected_3x, const HashMap<TokenType, String> &p_func_renames);
```
AThousandShips:
```suggestion
	bool _handle_new_keyword_rename(TokenE *p_pos, bool p_detected_3x, HashMap<TokenType, String> &p_func_renames);
	TokenE *_rename_keyword_id(TokenE *p_pos, bool p_detected_3x, const HashMap<TokenType, String> &p_func_renames);
```

servers/rendering/shader_converter.h
```diff
@@ -0,0 +1,310 @@
...
+	bool _tok_is_start_of_decl(const Token &p_tk);
+	bool _skip_uniform();
+	bool _parse_uniforms();
+	bool _parse_structs();
+	bool _skip_array_size();
+	bool _parse_struct();
+	bool _skip_struct();
+	bool _check_deprecated_type(TokenE *p_type_tok);
+	bool _add_to_report(int p_line, const String &p_msg, int level = 0);
+	bool _add_comment_before(const String &p_comment, List<Token>::Element *p_pos, bool warning = true);
```
AThousandShips:
```suggestion
	bool _add_to_report(int p_line, const String &p_msg, int p_level = 0);
	bool _add_comment_before(const String &p_comment, List<Token>::Element *p_pos, bool p_warning = true);
```

servers/rendering/shader_converter.h
```diff
@@ -0,0 +1,310 @@
...
+	bool _add_comment_at_eol(const String &p_comment, List<Token>::Element *p_pos);
+	bool _process_func_decl_statement(TokenE *p_start_tok, TokenE *p_type_tok, bool p_second_pass = false);
+	bool _process_decl_statement(TokenE *p_start_tok, TokenE *p_type_tok, const String &p_scope = "<global>", bool p_func_args = false);
+	bool _parse_decls(bool p_first_pass);
+	bool _process_decl_if_exist(String p_curr_func, bool p_first_pass);
```
AThousandShips:
```suggestion
	bool _process_decl_if_exist(const String &p_curr_func, bool p_first_pass);
```

tests/servers/rendering/test_shader_converter.h
```diff
@@ -0,0 +1,936 @@
...
+
+	SUBCASE("All renamed built-ins are not currently built-in") {
+		for (RS::ShaderMode mode : modes) {
+			ShaderLanguage::ShaderCompileInfo info;
+			get_compile_info(info, mode);
+			for (const String &builtin : builtins) {
+				// Now get the funcs applicable for this mode and built-in.
+				for (const String &func : rename_func_map[mode][builtin]) {
+					// The built-in should not be present in the built-ins list.
+					auto &finfo = info.functions[func];
```
AThousandShips: `auto` is [not allowed](https://contributing.godotengine.org/en/latest/engine/guidelines/cpp_usage_guidelines.html#auto-keyword)

tests/servers/rendering/test_shader_converter.h
```diff
@@ -0,0 +1,936 @@
...
+}
+
+// TODO: Remove this when the MODULATE built-in PR lands.
+// If this fails, remove the MODULATE entry from ShaderDeprecatedConverter::removed_builtins, then remove this test and the following test.
+TEST_CASE("[ShaderDeprecatedConverter] MODULATE is not a built-in") {
+	ShaderLanguage::ShaderCompileInfo info;
+	get_compile_info(info, RS::ShaderMode::SHADER_CANVAS_ITEM);
+	SUBCASE("MODULATE is not a built-in") {
+		for (const String &func : Vector<String>{ "vertex", "fragment", "light" }) {
+			auto &finfo = info.functions[func];
```
AThousandShips: Same

tests/servers/rendering/test_shader_converter.h
```diff
@@ -0,0 +1,936 @@
...
+	HashSet<String> render_modes_set;
+	for (const String &render_mode : render_modes) {
+		render_modes_set.insert(render_mode);
+	}
+	// static const char *code_template = "shader_type %s; render_mode %s;";
+	static const char *code_template = "shader_type %s; render_mode blend_mix, %s, depth_draw_always;";
+	SUBCASE("Renamed render modes are not currently valid render modes") {
+		ShaderLanguage::ShaderCompileInfo info;
+		get_compile_info(info, RS::SHADER_SPATIAL);
+		for (const auto &render_mode : info.render_modes) {
```
AThousandShips: Same

tests/servers/rendering/test_shader_converter.h
```diff
@@ -0,0 +1,936 @@
...
+	List<String> removed_render_modes;
+	ShaderDeprecatedConverter::_get_render_mode_removals_list(&removed_render_modes);
+	HashSet<String> render_modes_set;
+	for (const String &render_mode : removed_render_modes) {
+		render_modes_set.insert(render_mode);
+	}
+	SUBCASE("Removed render modes are not currently valid render modes") {
+		ShaderLanguage::ShaderCompileInfo info;
+		get_compile_info(info, RS::SHADER_SPATIAL);
+		for (const auto &render_mode : info.render_modes) {
```
AThousandShips: Same

tests/servers/rendering/test_shader_language.h
```diff
@@ -0,0 +1,144 @@
...
+		"shader_type %s;\nvoid vertex() {float %s;}\n",
+		"shader_type %s;\nuniform sampler2D %s;\n",
+		"shader_type %s;\nconst float %s = 1.0;\n",
+		nullptr
+	};
+	static const char *varying_template = "shader_type %s;\nvarying float %s;\n";
+	Vector<String> non_varying_types = { "particles", "sky", "fog" };
+
+	auto shader_types_to_test = ShaderTypes::get_singleton()->get_types();
+	for (auto shader_type : shader_types_to_test) {
```
AThousandShips: Same

servers/rendering/shader_language.cpp
```diff
@@ -956,22 +998,46 @@ ShaderLanguage::Token ShaderLanguage::_peek() {
...
-
 	Token tk = _get_token();
+
 	while (tk.type != TK_EOF && tk.type != TK_ERROR) {
-		output += itos(tk_line) + ": " + get_token_text(tk) + "\n";
+		String literal_text = p_code.substr(tk.pos, tk.length);
+		output += itos(tk_line) + " (" + itos(tk.pos) + ":" + itos(tk.pos + tk.length) + "): " + get_token_text(tk) + "  [" + literal_text;
+		String suffix = "]\n";
+		// add error string if invalid float constant
+		if (debug_parse && tk.type == TK_FLOAT_CONSTANT && literal_text.ends_with("f") && !literal_text.contains(".") && !literal_text.contains("e")) {
```
AThousandShips:
```suggestion
		if (debug_parse && tk.type == TK_FLOAT_CONSTANT && literal_text.ends_with("f") && !literal_text.contains_char('.') && !literal_text.contains_char('e')) {
```

scene/animation/animation_mixer.cpp
```diff
@@ -244,6 +244,97 @@ TypedArray<StringName> AnimationMixer::_get_animation_library_list() const {
 	return ret;
 }
 
+#if !defined(_3D_DISABLED) || !defined(DISABLE_DEPRECATED)
+bool AnimationMixer::_recalc_animation(Ref<Animation> &anim) {
```
AThousandShips:
```suggestion
bool AnimationMixer::_recalc_animation(Ref<Animation> &p_anim) {
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "core/error/error_macros.h"
+#include "core/templates/pair.h"
+#include "core/typedefs.h"
```
AThousandShips:
```suggestion
```
Already included

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+	List<String> current_builtin_funcs;
+	ShaderLanguage::get_builtin_funcs(&current_builtin_funcs);
+	HashSet<String> old_funcs;
+	for (int i = 0; old_builtin_funcs[i] != nullptr; i++) {
+		old_funcs.insert(old_builtin_funcs[i]);
+	}
+	HashSet<String> new_funcs;
+	for (List<String>::Element *E = current_builtin_funcs.front(); E; E = E->next()) {
+		if (!old_funcs.has(E->get())) {
+			new_funcs.insert(E->get());
```
AThousandShips:
```suggestion
	for (const String &E : current_builtin_funcs) {
		if (!old_funcs.has(E)) {
			new_funcs.insert(E);
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+		case TT::TK_INT_CONSTANT:
+		case TT::TK_FLOAT_CONSTANT:
+		case TT::TK_UINT_CONSTANT: {
+			if (p_tk.pos == NEW_IDENT) {
+				// Fix for 3.x float constants not having a decimal point.
+				if (!p_tk.is_integer_constant() && p_tk.text != "") {
+					return p_tk.text;
+				}
+				String const_str = rtos(p_tk.constant);
+				if (!p_tk.is_integer_constant() && !const_str.contains(".")) {
```
AThousandShips:
```suggestion
				if (!p_tk.is_integer_constant() && !const_str.contains_char('.')) {
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+	String report_str;
+	for (const KeyValue<int, Vector<String>> &p : report) {
+		for (const String &v : p.value) {
+			report_str += vformat(RTR("Line %d: %s\n"), p.key, v);
+		}
+	}
+	return report_str;
+}
+
+bool ShaderDeprecatedConverter::_add_to_report(int p_line, const String &p_msg, int level) {
```
AThousandShips:
```suggestion
bool ShaderDeprecatedConverter::_add_to_report(int p_line, const String &p_msg, int p_level) {
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+		report[p_line] = Vector<String>();
+	}
+	if (report[p_line].has(p_msg)) {
+		return false;
+	}
+	report[p_line].push_back(p_msg);
+	return true;
+}
+
+bool ShaderDeprecatedConverter::_add_comment_before(const String &p_comment, List<Token>::Element *p_pos, bool warning) {
```
AThousandShips:
```suggestion
bool ShaderDeprecatedConverter::_add_comment_before(const String &p_comment, List<Token>::Element *p_pos, bool p_warning) {
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+				return false;
+			}
+			continue;
+		}
+
+		switch (cur_tok->get().type) {
+			case TT::TK_FLOAT_CONSTANT: {
+				String const_str = get_token_literal_text(cur_tok->get()).to_lower();
+				// 3.x float constants allowed for a value without a decimal point if it ended in `f` (e.g. `1f`).
+				if (const_str.ends_with("f") && const_str.find(".") == -1 && const_str.find("e") == -1) {
```
AThousandShips:
```suggestion
				if (const_str.ends_with("f") && !const_str.contains_char('.') && !const_str.contains_char('e')) {
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+					}
+					if (next_tk->get().type == TT::TK_SEMICOLON) {
+						break;
+					}
+				}
+			} break;
+			case TT::TK_IDENTIFIER: {
+				String id = get_token_literal_text(cur_tok->get());
+				if (has_builtin_rename(shader_mode, id, curr_func) || is_removed_builtin(shader_mode, id, curr_func)) {
+					if ((!scope_has_decl(curr_func, id) && !function_decls.has(id))) {
```
AThousandShips:
```suggestion
					if (!scope_has_decl(curr_func, id) && !function_decls.has(id)) {
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+bool ShaderDeprecatedConverter::_check_deprecated_type(TokenE *p_type_pos) {
+	if (p_type_pos->get().type == TT::TK_IDENTIFIER && has_removed_type(get_token_literal_text(p_type_pos->get()))) {
+		const String i_err_msg = vformat(RTR("Deprecated type '%s' is not supported by this version of Godot."), get_token_literal_text(p_type_pos->get()));
+		COND_LINE_MSG_FAIL(fail_on_unported, p_type_pos->get().line, i_err_msg);
+		_add_comment_before(i_err_msg, p_type_pos);
+	}
+	return true;
+}
+
+ShaderDeprecatedConverter::TokenE *ShaderDeprecatedConverter::_rename_keyword_id(TokenE *pos, bool p_detected_3x, const HashMap<TokenType, String> &p_new_reserved_word_renames) {
```
AThousandShips:
```suggestion
ShaderDeprecatedConverter::TokenE *ShaderDeprecatedConverter::_rename_keyword_id(TokenE *p_pos, bool p_detected_3x, const HashMap<TokenType, String> &p_new_reserved_word_renames) {
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+	String rename = p_new_reserved_word_renames[pos->get().type];
+	reset_to(pos);
+	return replace_curr(mkTok(TT::TK_IDENTIFIER, rename), "Identifier '%s' is a reserved word in this version of Godot, renamed to '%s'.");
+}
+
+bool ShaderDeprecatedConverter::_handle_new_keyword_rename(TokenE *pos, bool p_detected_3x, HashMap<TokenType, String> &p_new_reserved_word_renames) {
```
AThousandShips:
```suggestion
bool ShaderDeprecatedConverter::_handle_new_keyword_rename(TokenE *p_pos, bool p_detected_3x, HashMap<TokenType, String> &p_new_reserved_word_renames) {
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+			} else if (all_hints.has(cur_tok)) {
+				continue; // Hint, don't replace it.
+			}
+			cur_tok = _rename_keyword_id(cur_tok, detected_3x, new_reserved_word_renames);
+			continue;
+		}
+		switch (cur_tok->get().type) {
+			case TT::TK_FLOAT_CONSTANT: {
+				// Earlier versions of Godot 3.x (< 3.5) allowed the use of the `f` sigil with float constants without a decimal place.
+				if (cur_tok_text.ends_with("f") && !cur_tok_text.contains(".") && !cur_tok_text.contains("e")) {
```
AThousandShips:
```suggestion
				if (cur_tok_text.ends_with("f") && !cur_tok_text.contains_char('.') && !cur_tok_text.contains_char('e')) {
```

editor/shader/text_shader_editor.cpp
```diff
@@ -1319,7 +1386,20 @@ TextShaderEditor::TextShaderEditor() {
...
+	shader_convert_error_dialog->set_title(TTR("Error converting shader"));
+	shader_convert_error_dialog->set_hide_on_ok(true);
+	add_child(shader_convert_error_dialog);
+
+	confirm_convert_shader = memnew(ConfirmationDialog);
+	confirm_convert_shader->set_title(TTR("Confirm Convert 3.x Shader"));
+	confirm_convert_shader->set_text(TTR("This shader does not appear to be a 3.x shader.\nAre you sure you want to convert it?"));
+	confirm_convert_shader->get_ok_button()->set_text(TTR("Convert"));
+	confirm_convert_shader->get_cancel_button()->set_text(TTR("Cancel"));
+	confirm_convert_shader->connect("confirmed", callable_mp(this, &TextShaderEditor::_convert_shader));
```
AThousandShips:
```suggestion
	confirm_convert_shader->connect(SceneStringName(confirmed), callable_mp(this, &TextShaderEditor::_convert_shader));
```

servers/rendering/shader_converter.h
```diff
@@ -0,0 +1,307 @@
...
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#pragma once
+
+#ifndef DISABLE_DEPRECATED
+
+#include "core/templates/pair.h"
```
AThousandShips:
```suggestion
```

servers/rendering/shader_converter.cpp
AThousandShips:
```suggestion
	ERR_FAIL_NULL_V(p_curr_ptr, p_curr_ptr);
```
For all these

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2591 @@
...
+	return curr_ptr;
+}
+
+List<SL::Token>::Element *ShaderDeprecatedConverter::get_prev_token() {
+	curr_ptr = _get_prev_token_ptr(curr_ptr);
+	return curr_ptr;
+}
+
+List<SL::Token>::Element *ShaderDeprecatedConverter::remove_cur_and_get_next() {
+	ERR_FAIL_COND_V(!curr_ptr, nullptr);
```
AThousandShips:
```suggestion
	ERR_FAIL_NULL_V(curr_ptr, nullptr);
```
For all these

servers/rendering/shader_converter.cpp
AThousandShips:
```suggestion
	String name = get_token_literal_text(pos->get());
```
Or something similar, `p_` is reserved for arguments

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2591 @@
...
+	TokenType tk_type = pos->get().type;
+	String p_name = get_token_literal_text(pos->get());
+	if (tokentype_is_new_reserved_keyword(tk_type)) {
+		if (!p_detected_3x) {
+			// If we're not sure it's a 3.x shader, just add a comment.
+			_add_comment_before(vformat(RTR("Identifier '%s' is a reserved word in this version of Godot."), p_name), pos);
+			return false;
+		}
+		if (!p_new_reserved_word_renames.has(tk_type)) {
+			String rename = p_name + String("_");
```
AThousandShips:
```suggestion
			String rename = p_name + "_";
```
Redundant, for all these

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2591 @@
...
+			String type = get_token_literal_text(var.type_pos->get());
+			if (struct_renames.has(type)) {
+				reset_to(var.type_pos);
+				var.type_pos = replace_curr(mkTok(TT::TK_IDENTIFIER, struct_renames[type]), "Struct type '%s' renamed to '%s'.");
+				reset_to(after_shader_decl);
+			}
+		}
+
+		if (_handle_new_keyword_rename(E.value[0].name_pos, detected_3x, new_reserved_word_renames)) {
+			for (VarDecl &varDecl : E.value) {
```
AThousandShips:
```suggestion
			for (VarDecl &var_decl : E.value) {
```
Please use `snake_case`

scene/animation/animation_mixer.cpp
```diff
@@ -238,6 +238,101 @@ TypedArray<StringName> AnimationMixer::_get_animation_library_list() const {
...
+
+	for (int i = 0; i < anim->get_track_count(); i++) {
+		int track_type = anim->track_get_type(i);
+		if (!anim->track_is_relative_to_rest(i)) {
+			continue;
+		}
+		if (track_type == Animation::TYPE_POSITION_3D || track_type == Animation::TYPE_ROTATION_3D || track_type == Animation::TYPE_SCALE_3D) {
+			NodePath path = anim->track_get_path(i);
+			Node *node = parent->get_node(path);
+			ERR_FAIL_COND_V(!node, false);
```
AThousandShips:
```suggestion
			ERR_FAIL_NULL_V(node, false);
```

AThousandShips (commented):
I forgot this one earlier

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+bool ShaderDeprecatedConverter::insert_after(const Token &p_token, List<Token>::Element *p_pos) {
+	ERR_FAIL_COND_V(p_pos == nullptr, false);
+	Token new_token = p_token;
+	new_token.pos = NEW_IDENT;
+	code_tokens.insert_after(p_pos, new_token);
+	return true;
+}
+
+bool ShaderDeprecatedConverter::insert_before(const Token &p_token, List<Token>::Element *p_pos) {
+	ERR_FAIL_COND_V(p_pos == nullptr, false);
```
AThousandShips:
```suggestion
	ERR_FAIL_NULL_V(p_pos, false);
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+			// replace the function name
+			String rename = get_main_function_rename(name);
+			reset_to(var.name_pos);
+			var.name_pos = replace_curr(mk_tok(TT::TK_IDENTIFIER, rename), "Function '%s' renamed to '%s' in this version of Godot.");
+			reset_to(after_shader_decl);
+			func_renames[name] = rename;
+			// Only doing this because "process" is a common word and we don't want to clobber an existing function/global named that.
+			// Won't clobber a pre-exising "process" function that has the correct main signature because of the check before.
+			if (function_decls.has(rename) || scope_has_decl("<global>", rename)) {
+				String rerename = rename + String("_");
```
AThousandShips:
```suggestion
				String rerename = rename + "_";
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+					for (int i = 0; i < var_decls[rename].size(); i++) {
+						VarDecl &rere_var = var_decls[rename].write[i];
+						reset_to(rere_var.name_pos);
+						rere_var.name_pos = replace_curr(mk_tok(TT::TK_IDENTIFIER, rerename), conflict_comment_prefix);
+						reset_to(after_shader_decl);
+					}
+				}
+			}
+		} else if (id_is_new_builtin_func(name)) {
+			String rename = name + String("_");
```
AThousandShips:
```suggestion
			String rename = name + "_";
```

Ivorforce (commented):
Can someone explain to me why this is adding compatibility handling code to setters and getters? The `get_meta` call isn't cheap, so checking it on call might incur a noticeable runtime cost. 
I would expect compatibility code to be handled in code where it is conceivable that we might be dealing with old resources (e.g. import), and not in code that could be called on a frame-by-frame basis (e.g. the setters / getters).
cc @lyuma, I'm reading above you have asked specifically to pull the compat code outside the importer?

nikitalita:
> Can someone explain to me why this is adding compatibility handling code to setters and getters? The `get_meta` call isn't cheap, so checking it on call might incur a noticeable runtime cost. 

I can replace it with member variables, for some reason I was gunshy about adding additional members when I originally wrote it

Ivorforce:
That might be better (especially if done as a bit field). 
But by separation of concerns, I would still much prefer if compat handling code was done outside runtime code. Compat code is something we only need to deal with very rarely, so it shouldn't increase the complexity of the runtime code, which we need to fully understand and be prepared to actively maintain.

nikitalita:
> That might be better (especially if done as a bit field). But by separation of concerns, I would still much prefer if compat handling code was done outside runtime code. Compat code is something we only need to deal with very rarely, so it shouldn't increase the complexity of the runtime code, which we need to fully understand and be prepared to actively maintain.

This is how we handle compatibility in nearly all other instances though; for example, look at the ArrayMesh code.

Ivorforce:
> This is how we handle compatibility in nearly all other instances though; for example, look at the ArrayMesh code.

I'm having a look, and yea, I do see a similar pattern. I'm not convinced this is the right approach to compatibility handling, but I haven't been there when that code was written.
I'll think about it a bit and ask other maintainers for their opinions.

scene/animation/animation_mixer.cpp
```diff
@@ -685,6 +780,11 @@ bool AnimationMixer::_update_caches() {
 	}
 	for (const StringName &E : sname_list) {
 		Ref<Animation> anim = get_animation(E);
+#if !defined(_3D_DISABLED) && !defined(DISABLE_DEPRECATED)
+		if (anim->has_tracks_relative_to_rest()) {
+			_recalc_animation(anim);
+		}
```
lyuma, line 792: This is adding code in runtime which I am not happy with.

It's also doing conversions in such a way that we now have an additional `relative_to_rest` property internally on Animations which we will have to continue to support across the entirety of the Animation system and tooling indefinitely.

Clearly @TokageItLab would need to give the final call on this, but given the complexity the animation system already has, I'm uncertain if this is a good idea. I really don't like that this introduces a strange interaction between the track cache and the Animation keyframes but only if a legacy flag is in use, and one that can never be removed.

When Godot 5.0 comes out, no doubt there will still be Animation resources floating around with this flag that will have the same problem. It also means that the animation track editor might be broken for these animations, since the actual tracks in the resource are rest-relative. How would we even know which Animation's are running this track conversion during every playback, and which aren't?

I understand the problem though... this conversion cannot be done during import because the skeleton rest is not known until runtime, and in fact the Animation resource could be loaded independently of the tscn with the Skeleton3D and only attached later on, so Godot would have no way to adapt the Animation resource for the associated Skeleton3D.

That said, if we have a way of knowing which Skeleton3D corresponds to the given Animation, then I think I would prefer if the escn importer should do the work of converting the legacy animation tracks to not be rest-relative.

editor/import/3d/resource_importer_scene.cpp
```diff
@@ -3425,11 +3425,43 @@ void EditorSceneFormatImporterESCN::get_extensions(List<String> *r_extensions) c
...
 	ERR_FAIL_COND_V_MSG(ps.is_null(), nullptr, "Cannot load scene as text resource from path '" + p_path + "'.");
 	Node *scene = ps->instantiate();
+	ERR_FAIL_COND_V(!scene, nullptr);
+	if (format == 2) {
+		TypedArray<Node> skel_nodes = scene->find_children("*", "AnimationPlayer");
+		for (int32_t node_i = 0; node_i < skel_nodes.size(); node_i++) {
+			// Force re-compute animation tracks.
+			AnimationPlayer *player = cast_to<AnimationPlayer>(skel_nodes[node_i]);
+			ERR_CONTINUE(!player);
+			player->advance(0);
```
lyuma, line 3565: if we can do the track conversion here using an explicit function call instead of `player->advance(0)` and remove all the code about caches, then I'd feel a lot less bad about the runtime cost issue.

But it doesn't change that this is bleeding importer complexity into the animation system.

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,2589 @@
...
+				} else if (has_removed_type(id) && peek_next_tk_type() == TT::TK_IDENTIFIER) {
+					// Declaration with unported 3.x type.
+					return true;
+				}
+			} break;
+			default:
+				break;
+		}
+	}
+	return is_3x;
```
lyuma, line 2354: Unused variable `is_3x`

It's not my area, but I want to say that given we're already at line 2030, the Shader converter is extremely complex, and I am skeptical if this amount of code is justifiable especially in runtime builds. I did not expect thousands of lines of code to be added to every build of Godot editor and runtime. My guess is this will create too great a maintainability burden to include in the core engine.

The ultimate call here goes to the rendering maintainers since they will shoulder the cost of maintaining this compatibility code.

lyuma (changes requested):
Sorry that I made a mistake with my comment 20 months ago. I genuinely forgot about this PR, and I didn't hear anything for a long time. Given that 4.x has been out a long time at this point, I question the value of introducing conversion for 3.x assets at this point.

> Can someone explain to me why this is adding compatibility handling code to setters and getters? The get_meta call isn't cheap, so checking it on call might incur a noticeable runtime cost.

Yeah I wasn't expecting runtime overhead here. My comment was building off the other PR #87050 which was a few lines of conversion code in a standalone property setter, which would not get run except if loading an old asset. If you see how that PR is written, that is the extent to which I expected code to be added to property setters.

Also, I was making the assumption that the property name had changed between 3.x and 4.x and that the conversion could be contained within the property setter... I did not want or expect detection code to be necessary in order to facilitate performing conversions in the property setter.

My expectation was that we would ultimately need some conversion tool, since this mimics the project conversion tool we used for things that changed between 3.x and 4.x.

I feel really bad because this is a huge amount of work, but I did not anticipate thousands of lines of code being added to facilitate shader migration, nor that there would be a flag on animation tracks to convert them at runtime... even if it may be the only option, it's not something I was thinking about when I made that comment.

If we do move forward with this, it might be good to gate any complex conversion code that could have a runtime cost behind TOOLS_ENABLED

If you want to chat more, I'd be happy to talk about options, insofar as I am actually responsible for the areas involved (of course, Animation and Rendering changes would have to be approved by their respective teams)

Also, a lot of the animation and asset maintainers are having a bi-weekly PR review meeting tomorrow in just under 12 hours. The information to join will be in the #animation channel in rocket chat if you happen to be available.

nikitalita:
To be honest with you, I’m leaning towards closing this PR. Nobody here is very enthusiastic about what this enables and most have concerns about runtime and maintenance costs, which is understandable, but which would be an impediment to merging without a large amount of rewriting. I’m currently using it in production on my fork, it runs fine, and I don’t really have a problem with just rebasing on top of master for the foreseeable future. Unless anyone here really wants this, I’m just going to close it.

Ivorforce:
Please don't close the PR! We are definitely interested in features that help developers migrate from 3.x to 4.x, and we really appreciate your work on the ESCN migration. I'm sorry if reviews or reactions didn't come off this way.

We're currently discussing among maintainers how we want to handle backwards compatibility code-wise. This seems not to be trivial, but we'll get back to you soon. Hopefully, we'll find a solution that addresses these concerns (reducing runtime code complexity) without creating much additional work for you.

_(PS. if you do decide to close the PR, that would also be OK. It's possible that someone else will pick it up and finalize it, and credit will be given to you on merge)_

nikitalita:
@Ivorforce thank you for saying so! I’m willing to continue working on it as long as maintainers are interested in it.

scene/animation/animation_mixer.cpp
nikitalita: >That said, if we have a way of knowing which Skeleton3D corresponds to the given Animation, then I think I would prefer if the escn importer should do the work of converting the legacy animation tracks to not be rest-relative.

This is originally how I had this setup, but I think I misconstrued your original comment a while back.

While this method has the added benefit of supporting older TSCNs as well, I understand the concerns about runtime complexity. So, would it be enough to simply remove the `ADD_PROPERTY` macro (and the property declaration in `_get_property_list`), and move the conversion code from the mixer to `animation.cpp` and only call it from the escn importer?

servers/rendering/shader_converter.cpp
nikitalita: It was originally a lot simpler. The reason for the complexity is that I wanted this to be a declarative as possible so that it could become a model for writing future conversion tools for future API breaks. Additionally, I knew that this might be seen as a maintenance burden, so I added a bunch of tests for it that were written such so that if features changed in a way that would break the converter, they'd fail and point to an easy way to fix it, thus reducing the maintenance burden.

scene/animation/animation_mixer.cpp
TokageItLab: I believe adding this as compatibility code is not acceptable. Relative animation was [removed in 4.0](https://godotengine.org/article/animation-data-redesign-40/).

In the past, after discussing with reduz, the alternative provided at that time was `_post_process_key_value()`, and it is possible to make relative animation work in a [custom module](https://github.com/TokageItLab/realtime_retarget) using this.

Supporting this would be akin to shipping that custom module in core, but considering the circumstances under which it was provided as a custom module, this would likely be rejected.

Please note that what can be provided as compatibility code is resource conversion, such as converting animation keys from relative to absolute animation. Runtime-based solutions are almost never permitted.

nikitalita:
To sum up what we discussed during the meeting:
- move code off hot path in update_caches in AnimationMixer to something cold
- Ensure that that code is moved to animation_mixer.compat.inc to reduce maintenance burden
- see if we can somehow ensure that the animation does get recalced whenever an old 3.x animation is loaded so that we don’t have to serialize the `relative_to_rest` property

lyuma:
Took some notes from the animation meeting.
[EDIT: I see you also took notes, @nikitalita . Thanks! I guess we can keep both sets here in case one of us missed something]

Something we all agree on is we need to move conversion code out of Animation and AnimationMixer. We do not want the relative to rest functionality to be reintroduced.

Need to work with core maintainers to come up with an acceptable place to perform offline conversion during import of escn or converting 3.x assets. Because the animation code needs access to the Skeleton3D node to determine rest poses, this cannot be done inside a property setter alone. Maybe can be done in some compatibility conversion code outside of the animation code itself.

There was discussion of maybe putting conversion code in .compat.inc. but how exactly this is done needs consensus with core maintainers.
We also need to make sure we surround any code we add with TOOLS_ENABLED and is_editor_hint checks.

nikitalita:
I've rebased this and updated the animation code to not use the metadata when updating transform tracks. I have not yet updated it to not perform animation caching on the hot path; please let me know when you have an answer for where to put this.

Also, I think it might be best if this got split into multiple PRs since this covers two separate areas, shader and animation code. Would that be preferable?

scene/resources/animation.h
```diff
@@ -153,6 +159,9 @@ class Animation : public Resource {
 	struct ScaleTrack : public Track {
 		LocalVector<TKey<Vector3>> scales;
 		int32_t compressed_track = -1;
+#ifndef DISABLE_DEPRECATED
+		bool relative_to_rest = false;
+#endif
```
TokageItLab:
```suggestion
```

scene/resources/animation.h
```diff
@@ -445,7 +459,11 @@ class Animation : public Resource {
 	double track_get_key_time(int p_track, int p_key_idx) const;
 	real_t track_get_key_transition(int p_track, int p_key_idx) const;
 	bool track_is_compressed(int p_track) const;
-
+#ifndef DISABLE_DEPRECATED
+	bool has_tracks_relative_to_rest() const;
+	bool track_is_relative_to_rest(int p_track) const;
+	void track_set_relative_to_rest(int p_track, bool p_relative_to_rest);
+#endif
```
TokageItLab:
```suggestion
```

TokageItLab:
Since this PR still contains many lines that should be removed, I suggest creating a new PR. In that PR, you should only modify files under `editor/project_upgrade/` and avoid writing migration functionality within each class except for `_set()` and `_get()`.

scene/resources/animation.cpp
```diff
@@ -5268,6 +5374,112 @@ void Animation::compress(uint32_t p_page_size, uint32_t p_fps, float p_split_tol
...
+			scale_keys.write[j * 5 + 4] = ofs[11]; // z
+		}
+		Array c_track_keys = track_data.keys();
+		c_track_keys.erase("type");
+		c_track_keys.erase("keys");
+		remove_track(track_idx + offset); // remove dummy track
+		NodePath node_path;
+		add_track(TYPE_POSITION_3D, track_idx + offset);
+		for (int j = 0; j < c_track_keys.size(); j++) {
+			String key = c_track_keys[j];
```
AThousandShips:
```suggestion
		for (String key : c_track_keys) {
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,3041 @@
...
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef DISABLE_DEPRECATED
+
+#include "shader_converter.h"
+#include "shader_types.h"
```
AThousandShips:
```suggestion

#include "servers/rendering/shader_types.h"
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,3041 @@
...
+			} break;
+			default:
+				break;
+		}
+	}
+	return true;
+}
+
+bool ShaderDeprecatedConverter::_preprocess_code() {
+	COND_MSG_FAIL(code_tokens.size() == 0, RTR("Empty shader file"));
```
AThousandShips:
```suggestion
	COND_MSG_FAIL(code_tokens.is_empty(), RTR("Empty shader file"));
```

servers/rendering/shader_converter.cpp
```diff
@@ -0,0 +1,3041 @@
...
+			} break;
+			default:
+				break;
+		}
+	}
+	return true;
+}
+
+String ShaderDeprecatedConverter::emit_code() const {
+	if (code_tokens.size() == 0) {
```
AThousandShips:
```suggestion
	if (code_tokens.is_empty()) {
```

tests/servers/rendering/test_shader_converter.h
```diff
@@ -0,0 +1,1007 @@
...
+		for (RS::ShaderMode mode : modes) {
+			for (const String &builtin : builtins) {
+				// now get the funcs applicable for this mode and builtins
+				ShaderLanguage::TokenType type = ShaderDeprecatedConverter::get_removed_builtin_uniform_type(builtin);
+				if (type == ShaderDeprecatedConverter::TokenType::TK_ERROR) {
+					continue;
+				}
+				Vector<ShaderLanguage::TokenType> hints = ShaderDeprecatedConverter::get_removed_builtin_hints(builtin);
+				Vector<String> funcs = ShaderDeprecatedConverter::_get_funcs_builtin_removal(mode, builtin);
+				String hint_string = "";
```
AThousandShips:
```suggestion
				String hint_string;
```

Zireael07:
Any updates? I have a project that's stuck on 3.x due to using an ESCN asset...

akien-mga:
Well if that PR works for your use case, you can build it locally and convert your game with it.

PR #87108, opened 2024-01-12
Capitalize OIDN in editor settings and properties

YuriSizov:
Fixes the name of OIDN not being properly capitalized in the editor settings:

godot windows editor dev x86_64_2024-01-12_13-17-01

godot windows editor dev x86_64_2024-01-12_13-45-57

----

While looking for other easy pickings I also noticed that we have a problem with `path3d` being turned into `Path 3d`. This is because the `3d` to `3D` conversion is only done on individual bits in names, but the conversion of the node name into two words is a separate code path.

godot windows editor dev x86_64_2024-01-12_13-44-33

We could add `path3d` explicitly to the list, but this feels counter-productive. I didn't risk it trying to make a quick fix after looking at it for a bit, but it's something we should eventually correct.


akien-mga (approved):

akien-mga:
Thanks!

PR #87115, opened 2024-01-12
Make TileMapLayers extend Node2D and work as children of TileMap

groud:
The main goal of this PR is to start the work of moving TileMapLayers to individual nodes. For now, those are implemented as TileMap internal nodes, but I aim to move them as independent nodes in later PRs.

A few notes:
- the main goal of the PR was to remove some layer-specific properties that were redundant with the Node2D properties.
- the TileMapLayer node is still significantly dependent on the parent TileMap node:
   - The TileMap node is the one responsible to queue tilemap internal updates. This update queuing will need to be moved to the internals of the TileMapLayer
   - Some properties are defined at the TileMap level (`collision_animatable`, `rendering_quandrant_size`, `collision_visibility_mode`, etc...). The plan here is to "duplicate" those properties at the layer-level, then make the TileMap node update all its children layers at once.
- This should be 99% backward-compatible for now, though there might be some unexpected side-effects here and there as moving to a child node needed some tweaks here and there.
- For now, TileMapLayer properties/API is not exposed, it should be one of the last step. Basically, I plan a similar API to TileMap but without the need to specify a layer.

Once this is merged, I plan in the short term to remove the remaining dependencies from TileMapLayer to TileMap, then to split TileMap into two:
- TileMapGroup: a simpler node that will group layers together, allowing children TileMapLayers to share a TileSet, get highlighted/dimmed depending on the selected layer (and, maybe, get their position constrained). The API should be fairly simple.
- TileMap (extending TileMapGroup) : for backward-compatibility

Related to: https://github.com/godotengine/godot-proposals/issues/7122

### DIFF
--- a/misc/extension_api_validation/4.2-stable.expected
+++ b/misc/extension_api_validation/4.2-stable.expected
@@ -51,3 +51,11 @@ Validate extension JSON: Error: Field 'classes/RenderingDevice/methods/texture_u
 Barrier arguments have been removed from all relevant functions as they're no longer required.
 Draw and compute list overlap no longer needs to be specified.
 Initial and final actions have been simplified into fewer options.
+
+
+GH-87115
+--------
+Validate extension JSON: Error: Field 'classes/TileMap/methods/get_collision_visibility_mode': is_const changed value in new API, from false to true.
+Validate extension JSON: Error: Field 'classes/TileMap/methods/get_navigation_visibility_mode': is_const changed value in new API, from false to true.
+
+Two TileMap getters were made const. No adjustments should be necessary.
--- a/scene/2d/tile_map_layer.cpp
+++ b/scene/2d/tile_map_layer.cpp
@@ -40,6 +40,18 @@
 #include "servers/navigation_server_3d.h"
 #endif // DEBUG_ENABLED
 
+TileMap *TileMapLayer::_fetch_tilemap() const {
+	return TileMap::cast_to<TileMap>(get_parent());
+}
+
+Ref<TileSet> TileMapLayer::_fetch_tileset() const {
+	TileMap *tile_map_node = _fetch_tilemap();
+	if (!tile_map_node) {
+		return Ref<TileSet>();
+	}
+	return tile_map_node->get_tileset();
+}
+
 #ifdef DEBUG_ENABLED
 /////////////////////////////// Debug //////////////////////////////////////////
 constexpr int TILE_MAP_DEBUG_QUADRANT_SIZE = 16;
@@ -51,11 +63,11 @@ Vector2i TileMapLayer::_coords_to_debug_quadrant_coords(const Vector2i &p_coords
 }
 
 void TileMapLayer::_debug_update() {
-	const Ref<TileSet> &tile_set = tile_map_node->get_tileset();
+	const Ref<TileSet> &tile_set = _fetch_tileset();
 	RenderingServer *rs = RenderingServer::get_singleton();
 
 	// Check if we should cleanup everything.
-	bool forced_cleanup = in_destructor || !enabled || !tile_map_node->is_inside_tree() || !tile_set.is_valid() || !tile_map_node->is_visible_in_tree();
+	bool forced_cleanup = in_destructor || !enabled || !tile_set.is_valid() || !is_visible_in_tree();
 
 	if (forced_cleanup) {
 		for (KeyValue<Vector2i, Ref<DebugQuadrant>> &kv : debug_quadrant_map) {
@@ -120,10 +132,10 @@ void TileMapLayer::_debug_update() {
 			} else {
 				ci = rs->canvas_item_create();
 				rs->canvas_item_set_z_index(ci, RS::CANVAS_ITEM_Z_MAX - 1);
-				rs->canvas_item_set_parent(ci, tile_map_node->get_canvas_item());
+				rs->canvas_item_set_parent(ci, get_canvas_item());
 			}
 
-			const Vector2 quadrant_pos = tile_map_node->map_to_local(debug_quadrant.quadrant_coords * TILE_MAP_DEBUG_QUADRANT_SIZE);
+			const Vector2 quadrant_pos = tile_set->map_to_local(debug_quadrant.quadrant_coords * TILE_MAP_DEBUG_QUADRANT_SIZE);
 			Transform2D xform(0, quadrant_pos);
 			rs->canvas_item_set_transform(ci, xform);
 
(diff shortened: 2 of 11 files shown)


misc/extension_api_validation/4.2-stable.expected
```diff
@@ -51,3 +51,12 @@ Validate extension JSON: Error: Field 'classes/RenderingDevice/methods/texture_u
 Barrier arguments have been removed from all relevant functions as they're no longer required.
 Draw and compute list overlap no longer needs to be specified.
 Initial and final actions have been simplified into fewer options.
+
+GH-19564
```
AThousandShips, line 55:
```suggestion

GH-19564
```

misc/extension_api_validation/4.2-stable.expected
```diff
@@ -51,3 +51,12 @@ Validate extension JSON: Error: Field 'classes/RenderingDevice/methods/texture_u
...
+--------
+Validate extension JSON: Error: Field 'classes/TileMap/methods/get_collision_visibility_mode': is_const changed value in new API, from false to true.
+Validate extension JSON: Error: Field 'classes/TileMap/methods/get_navigation_visibility_mode': is_const changed value in new API, from false to true.
+Validate extension JSON: Error: Hash changed for 'classes/TileMap/methods/get_collision_visibility_mode', from 78C71571 to 6526718C. This means that the function has changed and no compatibility function was provided.
```
AThousandShips, line 59: I think a non-const compatibility method needs to be bound for this

CC @dsnopek

dsnopek: Yeah, I think `const`-ness is part of the hash. (And we should never be ignoring any of the "hash changed" messages)

KoBeWi:
Layer names need some default:
image

I wonder how the editing will look once TileMapLayers are nodes. I assume your plan is to allow the TileMap editor work per layer, so you can select node and edit it directly. But will the current workflow be preserved? Once the TileMapGroup is a thing, will I be able to select it and edit any layer without switching nodes?

Also name TileMapGroup sounds confusing, because it suggests grouping TileMaps, while it groups layers.

groud:
> Layer names need some default:

Ah yeah, I'll fix that.

> I wonder how the editing will look once TileMapLayers are nodes. I assume your plan is to allow the TileMap editor work per layer, so you can select node and edit it directly. 

Yes, that's the plan. Basically you would select a layer node in the scene tree instead of selecting the layer in the bottom dropdown. 

> But will the current workflow be preserved? Once the TileMapGroup is a thing, will I be able to select it and edit any layer without switching nodes?

Yes, for backward compatibility, but for the TileMap nodes. I might keep the TileMap editor as-is and provide a converter from a TileMap node to TileMapGroup + TileMapLayers. I thought about removing the TileMap editor, and only provide the converter, but I think forcing people to change all their API calls might be a bit too early right now.

Basically the plan is:
- TileMapGroup + TileMapLayer = new workflow, editor on TIleMapLayer but use the parents TileMapGroup for storing some things.
- TileMap: deprecated. Maybe keeps the current editor (if it's not too much work). Provides a converter to the new workflow.

> Also name TileMapGroup sounds confusing, because it suggests grouping TileMaps, while it groups layers.

I kind of agree. Ideally I would have preferred the TileMapLayer node to be a TileMap one (basically removing the possibility to have layers on the TileMap node), but that would have made the backward compatibility a lot more complicated. We could rename TileMapGroup to TileMapLayerGroup, but that felt a bit long to me.

KoBeWi:
So this is a problem. The current workflow allows to quickly switch between layers with a shortcut and it's very efficient. This ability should be preserved with the new workflow. However, even if we are able to quickly switch between layers, it will involve inspector updates, which are often expensive. Hence I think the Group node should allow editing all layers at once, like currently. It doesn't need any duplicate code, the editor can handle 2 node types and the group would be an extended case.

arkology:
> I would have preferred the TileMapLayer node to be a TileMap one (basically removing the possibility to have layers on the TileMap node), but that would have made the backward compatibility a lot more complicated. We could rename TileMapGroup to TileMapLayerGroup, but that felt a bit long to me.

I almost sure this was already discussed, but is it possible (reasonable) to integrate new layer system into old one using concept "use new system if has TileMapLayer child node, use old system if there are no TileMapLayer children"? Of course this will be complicated in terms of code maintinability, but solves "easy to use" goal.

BTW please link proposal issue to first post.

groud:
> So this is a problem. The current workflow allows to quickly switch between layers with a shortcut and it's very efficient. This ability should be preserved with the new workflow

That's a good point, but I think it's probably easily doable.

>  However, even if we are able to quickly switch between layers, it will involve inspector updates, which are often expensive. Hence I think the Group node should allow editing all layers at once, like currently. It doesn't need any duplicate code, the editor can handle 2 node types and the group would be an extended case.

Hmm, that could maybe be doable but I am not sure. The main problem is that the TileMap node relies on the layers being strictly indexed, so right now, in this PR, there's even an array to list all children layers. So, like, there's kind of an incompatibility between the TileMap editor (with the dropdown) and the TileMapLayer workflow. That probably means that I would need a special editor for TileMapGroup, and would not be able to reuse parts of the TileMap editor.  But maybe I can figure out a solution later on.

However, this kind of sounds like putting a band-aid on a wooden leg. If switching node is slow due to inspector updates, we should maybe improve the inspector for that (like, threading its update or something).

> I almost sure this was already discussed, but is it possible (reasonable) to integrate new layer system into old one using concept "use new system if has TileMapLayer child node, use old system if there are no TileMapLayer children"? Of course this will be complicated in terms of code maintinability, but solves "easy to use" goal.

I'd rather avoid it TBH, this makes things more confusing than they need to be IMO.

misc/extension_api_validation/4.2-stable.expected
groud: Should be fixed now.

KoBeWi:
I tested the changes and found some weird bug with collisions.
My game scenes each have TileMap. The scenes are changed when you go out of screen and the new scene has global position next to the previous one.
The bug is that when I change scenes, some TileMap layers will have colliders shifted. What appears in game is different from how it collides. It only applies to some layers and they are offseted by some amount (not sure how much, given their real collider is invisible).

Here I go from one scene to another, and the other scene has tiles with no collision and invisible tiles. The bricks are on a different layer (0) than the sand (2).

https://github.com/godotengine/godot/assets/2223172/fca3582e-0312-4b71-b1bd-ef8088d0ce4f

What's worse, when I return to the first scene, my `move_and_collide()` will report a collision with collider being null, when I collide with the TileMap.

scene/2d/tile_map_layer.cpp
```diff
@@ -40,6 +40,16 @@
 #include "servers/navigation_server_3d.h"
 #endif // DEBUG_ENABLED
 
+TileMap *TileMapLayer::_fetch_tilemap() const {
+	return TileMap::cast_to<TileMap>(get_parent());
+}
+
+Ref<TileSet> TileMapLayer::_fetch_tileset() const {
+	TileMap *tile_map_node = _fetch_tilemap();
+	ERR_FAIL_COND_V(!tile_map_node, Ref<TileSet>());
```
KoBeWi, line 49:
```suggestion
	ERR_FAIL_NULL_V(tile_map_node, Ref<TileSet>());
```

scene/2d/tile_map_layer.cpp
```diff
@@ -51,11 +61,11 @@ Vector2i TileMapLayer::_coords_to_debug_quadrant_coords(const Vector2i &p_coords
 }
 
 void TileMapLayer::_debug_update() {
-	const Ref<TileSet> &tile_set = tile_map_node->get_tileset();
+	const Ref<TileSet> &tile_set = _fetch_tileset();
 	RenderingServer *rs = RenderingServer::get_singleton();
 
 	// Check if we should cleanup everything.
-	bool forced_cleanup = in_destructor || !enabled || !tile_map_node->is_inside_tree() || !tile_set.is_valid() || !tile_map_node->is_visible_in_tree();
+	bool forced_cleanup = in_destructor || !enabled || !is_inside_tree() || !tile_set.is_valid() || !is_visible_in_tree();
```
KoBeWi, line 68: `is_inside_tree()` is technically redundant, because `is_visible_in_tree()` already depends on it.

scene/main/canvas_item.h
```diff
@@ -237,7 +237,7 @@ class CanvasItem : public Node {
 	Color get_modulate() const;
 	Color get_modulate_in_tree() const;
 
-	void set_self_modulate(const Color &p_self_modulate);
+	virtual void set_self_modulate(const Color &p_self_modulate);
```
KoBeWi, line 240: Seems like it supersedes #75320? (and likely fixes the same issue)

KoBeWi:
The problem does not exist on master.

scene/main/canvas_item.h
groud: The point of making this virtual is for the TileMapLayer to trigger an update. It is needed because the modulation for the selected layer depends on this (basically, it overrides self_modulate on the layer canvas item. So if you modify this value, we need to darken it before applying it again).
 
It does not fix the issue, but I am not sure it is worth fixing TBH. The way to solve it would be to multiply the self-modulate value of the layer by the parent TileMap's one, but only if it is an internal node. That's kind of additional glue code that will not be required anyway with the new workflow. So, well, not sure how much we want to keep the deprecated workflow alive. 

In any case, the whole point of the PR is to make TileMapLayer less dependent on TileMap for now, so I'd rather fix that later on.

groud:
Alright, I've just updated the PR. The bugs seems fixed now.
This PR includes the changes in https://github.com/godotengine/godot/pull/86841, it is required for the bug reported by @KoBeWi to be fixed.

KoBeWi:
The collision issue is fixed, but the null collider bug still exists (i.e. `move_and_collide()` returning a collision where `get_collider()` returns null).

groud:
> The collision issue is fixed, but the null collider bug still exists (i.e. move_and_collide() returning a collision where get_collider() returns null).

Hmm, that's quite weird. It seems attached to the tilemap node correctly: `ps->body_attach_object_instance_id(body, tile_map_node->get_instance_id());` I'll investigate.

KoBeWi:
It does not happen consistently. In my case it happens when going to another scene and back (refer to my comment/video above).

KoBeWi:
In my case I don't truly change/reload scene, just swap the map. The player node exists between scenes. Seems like this is the required condition.

Here's some simple repro scene:
[NoColliderBug.tscn.txt](https://github.com/godotengine/godot/files/13953648/NoColliderBug.tscn.txt)
Just press any key and you will start getting nulls. The problem does not occur in 4.3 dev2.

groud:
> The collision issue is fixed, but the null collider bug still exists (i.e. move_and_collide() returning a collision where get_collider() returns null).

The destructor for layers was simply not cleaning things up, including physics bodies. So you would end up with bodies attached to a freed node. I updated the PR and it should be fixed now.

scene/main/canvas_item.h
KoBeWi: I mean TileMapLayer `self_modulate` has no effect. This at least should be fixed.
Also if it's only used for darkening then it's pointless right now, because you can't change `self_modulate` inside editor without using hacks to make the node appear in scene.

groud: > I mean TileMapLayer self_modulate has no effect. This at least should be fixed.

Well, I don't know. Something like the ColorPicker node, which has internal sub-nodes, has a self_modulate property that does nothing too. Solving it kind of requires the implementing something like `TileMapLayer::set_layer_as_internal(...)` so that the TileMapLayer behaves differently depending on whether it is an internal node of the TileMap or not. 

I feel it would be a bit hacky, even if I agree it seems like more practical to have self-modulate doing something. So well, I guess it's probably worth it.

>  Also if it's only used for darkening then it's pointless right now, because you can't change self_modulate inside editor without using hacks to make the node appear in scene. (also DIRTY_FLAGS_LAYER_MODULATE looks unused)

Ah right. but it will be useful in the new workflow. Let's say it's future-proof then. :grin:

KoBeWi: Why being internal matters here? When layer will be visible in the editor, users will expect that it's `self_modulate` is functional.
And again, this has nothing to do with the parent TileMap node.

groud: I've just pushed an update, modifying `TileMapLayer`'s self_modulate in the remote tree should work now.

groud:
I also added @KoBeWi 's icon. It's a bit weird, but that's a good placeholder for now: 
Screenshot_20240117_160529

KoBeWi (approved):

YuriSizov:
Thanks!

groud:
Ah damn, the canvas_item.cpp changes was wrong, I was expecting the other PR to be merged before and to rebase that one. 

Sorry, I should have said it. I'll open a PR to fix that tomorrow.

naturally-intelligent:
Will this change make an idea like selecting multiple layers, as described in this proposal (https://github.com/godotengine/godot-proposals/issues/9296) harder or easier?

KoBeWi:
Selecting multiple layers will be easier, but it doesn't affect editing them.

PR #87117, opened 2024-01-12
Add renaming of PDB files to avoid blocking them

### LINKED ISSUE #82536: Locked PDB file blocks GDExtension compilation when Editor is launched within debugging
### Godot version

v4.2.dev (4c3dc2636)

### System information

Godot v4.2.dev (4c3dc2636) - Windows 10.0.19042 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1660 Ti (NVIDIA; 31.0.15.3713) - Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz (12 Threads)

### Issue description

When Godot Editor is launched within debugging session, it locks PDB file, preventing GDextension compilation

Very similar issue with DLL blocking was fixed recently in this PR: https://github.com/godotengine/godot/pull/80188. That works fine when Editor launched without debugger, but with debugger it also blocks PDB. Could we copy it the same way?

### Steps to reproduce

1. checkout godot master branch
2. compile it with `scons dev_build=yes debug_symbols=yes platform=windows compiledb=yes -j8`
3. get some gdextension. dodge-the-creeps example from gdext is fine, then `git clone https://github.com/godot-rust/gdext.git && cd gdext/examples/dodge-the-creeps/rust`
4. compile this gdextension `cargo build --package dodge-the-creeps --lib`
5. start editor with attached debugger (I used CLion) `path_to_godot_working_copy\bin\godot.windows.editor.x86_64.exe --editor --path ../godot` (that's path from `rust` dir to `gdext/examples/dodge-the-creeps/godot`)
6. wait for scene loading. it should load fine.
6. open `./src/player.rs` (`gdext/examples/dodge-the-creeps/rust/src/player.rs`) in any texty editor and append following line to the end: `impl Player { fn hello() {} }`
7. again compile cargo build --package dodge-the-creeps --lib --features custom-godot
8. get error: `note: LINK : fatal error LNK1201: error writing to program database 'C:\dev\reload_lab\gdext\target\debug\deps\dodge_the_creeps.pdb'; check for insufficient disk space, invalid path, or insufficient privilege`

N/A

DmitriySalnikov:
Fixes #82536
Supersedes https://github.com/godotengine/godot-cpp/pull/1331

In this implementation, I patch the copied ~DLL by replacing the PDB path in it with a new non-blocked path. Actually, all that is left of the PDB path is the file name, so that the new name almost never goes beyond the existing `CODEVIEW` entry.

Strangely, while I was testing this implementation the debugger did not block more than one PDB file, whereas in python/scons implementation PDBs were blocked until the debugger was restarted. If PDBs will be blocked until the debugger is restarted, 1000 copies of the PDB can be created.
I also noticed that `vscode` and `vs` load debug symbols with no problem, but `vs` doesn't update the breakpoints state and they have to be manually recreated.

If the DLL is renamed, then the PDB too, and previously created copies of it will not be deleted automatically.
Also, if a version without debug symbols is built, past previous will not be deleted.

Tested only with `godot-cpp` and `msvc`.

### DIFF
--- a/core/extension/gdextension.cpp
+++ b/core/extension/gdextension.cpp
@@ -729,52 +729,17 @@ GDExtensionInterfaceFunctionPtr GDExtension::get_interface_function(const String
 
 Error GDExtension::open_library(const String &p_path, const String &p_entry_symbol) {
 	String abs_path = ProjectSettings::get_singleton()->globalize_path(p_path);
-#if defined(WINDOWS_ENABLED) && defined(TOOLS_ENABLED)
-	// If running on the editor on Windows, we copy the library and open the copy.
-	// This is so the original file isn't locked and can be updated by a compiler.
-	bool library_copied = false;
-	if (Engine::get_singleton()->is_editor_hint()) {
-		if (!FileAccess::exists(abs_path)) {
-			ERR_PRINT("GDExtension library not found: " + abs_path);
-			return ERR_FILE_NOT_FOUND;
-		}
+	String actual_lib_path;
+	Error err = OS::get_singleton()->open_dynamic_library(abs_path, library, true, &actual_lib_path, Engine::get_singleton()->is_editor_hint());
 
-		// Copy the file to the same directory as the original with a prefix in the name.
-		// This is so relative path to dependencies are satisfied.
-		String copy_path = abs_path.get_base_dir().path_join("~" + abs_path.get_file());
-
-		// If there's a left-over copy (possibly from a crash) then delete it first.
-		if (FileAccess::exists(copy_path)) {
-			DirAccess::remove_absolute(copy_path);
-		}
-
-		Error copy_err = DirAccess::copy_absolute(abs_path, copy_path);
-		if (copy_err) {
-			ERR_PRINT("Error copying GDExtension library: " + abs_path);
-			return ERR_CANT_CREATE;
-		}
-		FileAccess::set_hidden_attribute(copy_path, true);
-		library_copied = true;
-
-		// Save the copied path so it can be deleted later.
-		temp_lib_path = copy_path;
-
-		// Use the copy to open the library.
-		abs_path = copy_path;
+	if (actual_lib_path.get_file() != abs_path.get_file()) {
+		// If temporary files are generated, let's change the library path to point at the original,
+		// because that's what we want to check to see if it's changed.
+		library_path = actual_lib_path.get_base_dir().path_join(p_path.get_file());
 	}
-#endif
 
-	Error err = OS::get_singleton()->open_dynamic_library(abs_path, library, true, &library_path);
 	ERR_FAIL_COND_V_MSG(err == ERR_FILE_NOT_FOUND, err, "GDExtension dynamic library not found: " + abs_path);
 	ERR_FAIL_COND_V_MSG(err != OK, err, "Can't open GDExtension dynamic library: " + abs_path);
-
-#if defined(WINDOWS_ENABLED) && defined(TOOLS_ENABLED)
-	// If we copied the file, let's change the library path to point at the original,
-	// because that's what we want to check to see if it's changed.
-	if (library_copied) {
-		library_path = library_path.get_base_dir() + "\\" + p_path.get_file();
-	}
-#endif
 
 	void *entry_funcptr = nullptr;
 
@@ -802,13 +767,6 @@ Error GDExtension::open_library(const String &p_path, const String &p_entry_symb
 void GDExtension::close_library() {
 	ERR_FAIL_NULL(library);
 	OS::get_singleton()->close_dynamic_library(library);
-
-#if defined(TOOLS_ENABLED) && defined(WINDOWS_ENABLED)
-	// Delete temporary copy of library if it exists.
-	if (!temp_lib_path.is_empty() && Engine::get_singleton()->is_editor_hint()) {
-		DirAccess::remove_absolute(temp_lib_path);
-	}
-#endif
 
 	library = nullptr;
 	class_icon_paths.clear();
@@ -1014,13 +972,6 @@ Error GDExtensionResourceLoader::load_gdextension_resource(const String &p_path,
 
 	err = p_extension->open_library(is_static_library ? String() : library_path, entry_symbol);
 	if (err != OK) {
-#if defined(WINDOWS_ENABLED) && defined(TOOLS_ENABLED)
-		// If the DLL fails to load, make sure that temporary DLL copies are cleaned up.
-		if (Engine::get_singleton()->is_editor_hint()) {
-			DirAccess::remove_absolute(p_extension->get_temp_library_path());
-		}
-#endif
-
 		// Unreference the extension so that this loading can be considered a failure.
 		p_extension.unref();
 
--- a/core/extension/gdextension.h
+++ b/core/extension/gdextension.h
@@ -47,9 +47,6 @@ class GDExtension : public Resource {
 
 	void *library = nullptr; // pointer if valid,
 	String library_path;
-#if defined(WINDOWS_ENABLED) && defined(TOOLS_ENABLED)
-	String temp_lib_path;
-#endif
 	bool reloadable = false;
 
 	struct Extension {
@@ -129,10 +126,6 @@ public:
 
 	Error open_library(const String &p_path, const String &p_entry_symbol);
 	void close_library();
-
-#if defined(WINDOWS_ENABLED) && defined(TOOLS_ENABLED)
-	String get_temp_library_path() const { return temp_lib_path; }
-#endif
 
 	enum InitializationLevel {
 		INITIALIZATION_LEVEL_CORE = GDEXTENSION_INITIALIZATION_CORE,
--- a/core/os/os.h
+++ b/core/os/os.h
@@ -153,7 +153,7 @@ public:
 
 	virtual void alert(const String &p_alert, const String &p_title = "ALERT!");
 
-	virtual Error open_dynamic_library(const String &p_path, void *&p_library_handle, bool p_also_set_library_path = false, String *r_resolved_path = nullptr) { return ERR_UNAVAILABLE; }
+	virtual Error open_dynamic_library(const String &p_path, void *&p_library_handle, bool p_also_set_library_path = false, String *r_resolved_path = nullptr, bool p_generate_temp_files = false) { return ERR_UNAVAILABLE; }
 	virtual Error close_dynamic_library(void *p_library_handle) { return ERR_UNAVAILABLE; }
 	virtual Error get_dynamic_library_symbol_handle(void *p_library_handle, const String &p_name, void *&p_symbol_handle, bool p_optional = false) { return ERR_UNAVAILABLE; }
 
--- a/drivers/unix/os_unix.cpp
+++ b/drivers/unix/os_unix.cpp
@@ -741,7 +741,7 @@ String OS_Unix::get_locale() const {
 	return locale;
 }
 
-Error OS_Unix::open_dynamic_library(const String &p_path, void *&p_library_handle, bool p_also_set_library_path, String *r_resolved_path) {
+Error OS_Unix::open_dynamic_library(const String &p_path, void *&p_library_handle, bool p_also_set_library_path, String *r_resolved_path, bool p_generate_temp_files) {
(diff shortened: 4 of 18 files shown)


platform/windows/editor_windows_utils.h
```diff
@@ -0,0 +1,45 @@
...
+#ifndef EDITOR_WINDOWS_UTILS_H
+#define EDITOR_WINDOWS_UTILS_H
+
+#if defined(WINDOWS_ENABLED) && defined(TOOLS_ENABLED)
+
+#include "core/string/ustring.h"
+
+class EditorWindowsUtils {
+public:
+	static Error copy_and_rename_pdb(const String p_dll_path);
```
AThousandShips, line 40:
```suggestion
	static Error copy_and_rename_pdb(const String &p_dll_path);
```

platform/windows/editor_windows_utils.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include "core/io/file_access.h"
+
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#undef FAILED // overrides Error::FAILED
+
+#include <dbghelp.h>
+#include <winnt.h>
+
+Error EditorWindowsUtils::copy_and_rename_pdb(const String p_dll_path) {
```
AThousandShips, line 46:
```suggestion
Error EditorWindowsUtils::copy_and_rename_pdb(const String &p_dll_path) {
```

platform/windows/editor_windows_utils.cpp
```diff
@@ -0,0 +1,233 @@
...
+	// Open and read the PDB information if available.
+	{
+		ULONG dbg_info_size = 0;
+		DWORD dbg_info_position = 0;
+
+		{
+			// The custom LoadLibraryExW is used instead of open_dynamic_library to avoid loading the original PDB into the debugger.
+			HMODULE library_ptr = LoadLibraryExW((LPCWSTR)(p_dll_path.utf16().get_data()), NULL, LOAD_LIBRARY_AS_DATAFILE_EXCLUSIVE);
+
+			ERR_FAIL_COND_V_MSG(!library_ptr, ERR_FILE_CANT_OPEN, vformat("Failed to load library '%s'.", p_dll_path));
```
AThousandShips, line 65:
```suggestion
			ERR_FAIL_NULL_V_MSG(library_ptr, ERR_FILE_CANT_OPEN, vformat("Failed to load library '%s'.", p_dll_path));
```

platform/windows/editor_windows_utils.cpp
```diff
@@ -0,0 +1,233 @@
...
+			file->seek(dbg_info_position);
+			dll_data = file->get_buffer(dbg_info_size);
+			ERR_FAIL_COND_V_MSG(file->get_error() != OK, file->get_error(), vformat("Failed to read data from library '%s'.", p_dll_path));
+		}
+
+		const char *raw_pdb_path = nullptr;
+		int raw_pdb_offset = 0;
+		DWORD *pdb_info_signature = (DWORD *)dll_data.ptr();
+
+		auto get_info = [&raw_pdb_path, &raw_pdb_offset](auto pdb) {
```
AThousandShips, line 120: [`auto` should be avoided](https://docs.godotengine.org/en/stable/contributing/development/cpp_usage_guidelines.html#lambdas) if it's not absolutely necessary

In this case it isn't necessary, you can make it call depending on the case below IMO

platform/windows/editor_windows_utils.cpp
```diff
@@ -0,0 +1,233 @@
...
+			BYTE Guid[16];
+			DWORD Age;
+			BYTE PdbFileName[1];
+		};
+
+		Vector<uint8_t> dll_data;
+
+		{
+			Error err = OK;
+			auto file = FileAccess::open(p_dll_path, FileAccess::READ, &err);
```
AThousandShips, line 108:
```suggestion
			Ref<FileAccess> file = FileAccess::open(p_dll_path, FileAccess::READ, &err);
```
`auto` is [not allowed](https://docs.godotengine.org/en/stable/contributing/development/cpp_usage_guidelines.html#auto-keyword)

platform/windows/editor_windows_utils.cpp
DmitriySalnikov: I used a lambda because I didn't find `offsetof()` from `stddef.h` at first (didn't want to use `cstddef`). Now I made it more compact.

platform/windows/editor_windows_utils.cpp
DmitriySalnikov: Yeah, I know, I forgot to replace it.

DmitriySalnikov:
After a strange bug with breakpoints in `Visual Studio`, I decided to try other debuggers:
* `remedybg` 3.9.6+ loads the PDB and updates breakpoints out of the box. That said, it doesn't even lock the PDB and this patch makes no sense for it 😐.
* `x64dbg` also loads updated PDBs without problems and does not block them.
* `raddbg` updates breakpoints without problems and does not lock the PDB (and creates its own locked files). But at the moment has problems with [relative paths](https://github.com/EpicGames/raddebugger/issues/32).

<details>
<summary>Visual Studio bug, not related to PR</summary>
In my testing only <code>Visual Studio</code> has the breakpoints bug. Of course, there are other IDEs and debuggers, but I have not tested them.
</br>
Example of a bug:

https://github.com/godotengine/godot/assets/7782218/8ec85d50-7503-4dbb-8e81-f7a3598a7666

</details>

dsnopek:
Thanks!

Overall, I think pursuing the approach of copying the PDB file is the right way to go. However, I'm not sure about a couple of things:

- I don't know that it makes sense for something under `core/` to directly include something from `platform/`. In other cases where we do this, it's much more generic, and it's done with special headers like `platform_thread.h`, `platform_gl.h`, and `platform_config.h`. I'll defer to the production team (cc @akien-mga @YuriSizov) on this, but it may be better to keep the Windows code for this in `core/extension/`
- With the DLL copies, we're only making one copy and we clean it up at shutdown. This PR appears to allow making up to 1000 copies of the PDB and it appears to clean them up only when running the editor again, but not at editor shutdown. I'd prefer we handle the PDB in the same way we handle the DLL, where we just make one copy with a predictable name, and clean it up afterwards - unless there is a compelling technical reason not to do it that way

YuriSizov:
> * I don't know that it makes sense for something under `core/` to directly include something from `platform/`. In other cases where we do this, it's much more generic, and it's done with special headers like `platform_thread.h`, `platform_gl.h`, and `platform_config.h`. I'll defer to the production team (cc @akien-mga @YuriSizov) on this, but it may be better to keep the Windows code for this in `core/extension/`

As a general rule platform code should not be bundled into anything outside of the each platform's respective folder. This allows for platforms to remain modular and for new platforms to be able to use all of the existing framework without patching the engine.

I can see in this case that there is already a bunch of platform-specific code in the modified file. So it's tempting to say that this is okay as is. However, your example of `platform_thread.h` and friends looks more appealing to me. It offers a hook for platforms to integrate into the process without an explicit dependency. That would be the way to do it, IMO. Possibly to move all platform-dependent hacks out of this core file this way.

dsnopek:
> This won't be a problem for some debuggers, but Visual Studio will block the removal of the PDB (at least one).

I think one debugger (even a popular one) having issues with cleaning up the PDB when the editor shuts down is probably acceptable. We'll always need code to clean-up or overwrite a left over PDB when the editor starts up again, because the editor can always crash before it's clean up can run. But I think attempting to clean up (and only making one copy in the first place) is important to try to do.

DmitriySalnikov:
Added HashMap for storing temporary PDB files.
Added `remove_temp_temp_pdbs` to try to delete temporary PDB files when closing the extension.
Fixed renaming of PDB files when running a project.

DmitriySalnikov:
Do I need to make any other edits?

DmitriySalnikov:
Done.

dsnopek:
Thanks! This is looking really great to me :-) 

I did some testing (although, I'm not the best person to test Windows and MSVC stuff). When building my extension with gcc (which doesn't use a PDB file), it correctly did nothing with no complaints. :-) Then with MSVC, it seems to be working correctly: the PDB file got renamed, and it still stopped on my breakpoints. I can confirm the behavior where the PDB file doesn't get cleaned up at exit when debugging with MSVC, but I don't think there's any more we can do about that.

The only remaining open question is if it's OK that `core/extension/gdextension.cpp` is directly including `platform/windows/editor_windows_utils.h`? Usually, we don't allow stuff in `core/` to directly include stuff in `platform/`.

We have a couple of workarounds to this where all platforms have a specific header, like `platform_config.h` and `platform_gl.h` and we include that, and the current platform's version gets pulled in. We could add a `platform_extension.h` or something like that? Although, I don't know that we'll ever have other platforms that need special code aside from Windows.

@akien-mga What do you think? Is the direct `core/` -> `platform/` include OK in this case? Or, should we try to come up with some way to make it indirect? Or, move the Windows util code into core?

akien-mga:
> @akien-mga What do you think? Is the direct `core/` -> `platform/` include OK in this case? Or, should we try to come up with some way to make it indirect? Or, move the Windows util code into core?

Yeah this needs to be changed, `core` shouldn't include anything from outside `core`.

dsnopek:
I haven't thought about this too deeply, but I wonder if we could put the PDB copying logic (and the existing DLL copying logic) into `OS_Windows::open_dynamic_library()` and `OS_Windows::close_dynamic_library()`?

Android has a bunch of custom copying logic in its `open_dynamic_library()` and this wouldn't be too dissimilar.

DmitriySalnikov:
So you want me to move all `#ifdef WINDOWS_ENABLED ... #endif` to `os_windows.cpp`? I think it looks doable, but somehow it needs to be specified in `open_dynamic_library` that temporary files need to be created, any ideas?

Leave the `editor_windows_utils.cpp ` or also merge with `os_windows.cpp `?

dsnopek:
> So you want me to move all `#ifdef WINDOWS_ENABLED ... #endif` to `os_windows.cpp`?

Well, so far it's just an idea - I haven't thought through it far enough to know if it's a good idea or not. :-) If you take a look at it and it doesn't seem like there'd be any problems, then give it a try.

> I think it looks doable, but somehow it needs to be specified in `open_dynamic_library` that temporary files need to be created, any ideas?

Right now, we make the temporary copies when `Engine::get_singleton()->is_editor_hint()` is true. The `Engine` singleton is part of core, so I think it should be fine to keep using that?

> Leave the `editor_windows_utils.cpp ` or also merge with `os_windows.cpp `?

I think either way would fine. Within `platform/windows/` it can have any many helper classes as makes sense.

DmitriySalnikov:
> Right now, we make the temporary copies when `Engine::get_singleton()->is_editor_hint()` is true.

I found the use of `open_dynamic_library` in the `mono` module. I don't think `is_editor_hint` is enough.

I'm thinking of adding another argument to `open_dynamic_library`. I will have to update all OS implementations, but then it will be possible to activate the generation of temporary files, for example for FFI (#85741).

DmitriySalnikov:
* Moved Windows code from `gdextension.cpp ` to `os_windows.cpp `.
* Added the `p_generate_temp_files` argument to `open_dynamic_library`.
* The maximum number of prints of `Failed to remove temp PDB` is limited.

dsnopek:
Discussed at the GDExtension meeting, and we think moving the Windows-specific code into `OS_Windows` makes sense! Someone will still need to review the code and re-test, but at a high-level this seems like the right direction.

dsnopek (approved):
Sorry for taking so long to come back to this yet again!

However, I reviewed the code in-depth and it looks great to me. I also re-did the same Windows testing I did earlier, and everything still seems to be working fine. :-)

This needs to be re-based (there are conflicts now), but after that, I think this is ready to go!

DmitriySalnikov:
In #89677, the paths in `ERR_PRINT` were fixed, but in my PR, these prints were moved to `os_windows.cpp` and I fixed the paths there a while ago.
There was also a conflict in `SCsub`, where I just added my cpp file.

DmitriySalnikov:
Is there a chance of merging before 4.3 is released?

bruvzg (approved):

akien-mga:
Thanks!

PR #57469, opened 2022-01-31
Fix incorrect Rect2i calculations: intersects and encloses

Sauermann:
Correction of some border cases for Rect2i calculations.
Clarify expand documentation.
Added Unit tests for these cases.

resolves #57468

PR #32478, opened 2019-10-02
Make Rect2.encloses return true on same size

AlexHolly:
fix #26615

```
func _ready():
	assert(Rect2(0,0,0,0).encloses(Rect2(0,0,0,0)))
	assert(Rect2(0,0,2,2).encloses(Rect2(0,0,2,2)))
	assert(Rect2(0,0,2,2).encloses(Rect2(1,1,1,1)))
	assert(Rect2(0,0,2,2).encloses(Rect2(1,1,0.9,0.9)))
```

PR #87118, opened 2024-01-12
[Core] Fix `AABB.encloses` failing on shared upper bound

AThousandShips:
This differs from `Rect2(i)` and was fixed for those classes in the past

See:
* https://github.com/godotengine/godot/pull/32478
* https://github.com/godotengine/godot/pull/57469

Which fixed this for those types, makes sense to me to have parity between the two, and it doesn't make sense to only allow it on one side, or fail to enclose yourself. I'd also say that the following should be equivalent: `a.encloses(b)` and `a.intersection(b) == b`

Essentially: Either it should be true only when *no* sides touch, or it should allow any side to touch, an alternative would be to add an `encloses_inclusive` or some other wording that does what this does, and change all the checks for the `encloses` to be strict

(Will open a godot-cpp equivalent if this gets approved)

### DIFF
--- a/core/math/aabb.h
+++ b/core/math/aabb.h
@@ -200,11 +200,11 @@ inline bool AABB::encloses(const AABB &p_aabb) const {
 
 	return (
 			(src_min.x <= dst_min.x) &&
-			(src_max.x > dst_max.x) &&
+			(src_max.x >= dst_max.x) &&
 			(src_min.y <= dst_min.y) &&
-			(src_max.y > dst_max.y) &&
+			(src_max.y >= dst_max.y) &&
 			(src_min.z <= dst_min.z) &&
-			(src_max.z > dst_max.z));
+			(src_max.z >= dst_max.z));
 }
 
 Vector3 AABB::get_support(const Vector3 &p_normal) const {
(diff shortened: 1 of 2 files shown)


Sauermann (approved):

aaronfranke (approved):
Yup, this is good for consistency and I'd say matches user expectations. Approved.

akien-mga:
Would like @lawnjelly's opinion on this, I remember *a lot* of bikeshedding last time we attempted changing such conventions.

Sauermann:
The bikeshedding in #57469 was mostly about `expand_to` with the result, that it works as intended.
My understanding was, that the change to `a1.encloses(a1) == true` was the consensus. So this PR simply fixes a bug in my opinion.

lawnjelly:
Yes this is fine imo. The difficult version is the integer bound where you can have exclusive or inclusive definition at upper border. 👍

AThousandShips:
Will add some extra checks like these to the `Rect2(i)` in a separate PR to ensure compliance

akien-mga:
Thanks!

AThousandShips:
Thank you!

YuriSizov:
Cherry-picked for 4.2.2.

PR #87125, opened 2024-01-12
Clear drag preview nodes on `NOTIFICATION_DRAG_END`

ryevdokimov:
This didn't really work before. It took seconds for the preview node to actually delete when 'ui_cancel' was pressed before dropping, and there are certain situations where you could keep the preview node in the viewport indefinitely, such as panning around.

2D Before:

https://github.com/godotengine/godot/assets/105675984/55fd767b-a051-4190-949f-d15553afc8f0

2D After: 

https://github.com/godotengine/godot/assets/105675984/c69ace89-0fa1-465e-8339-bff237f06469

3D Before: 

https://github.com/godotengine/godot/assets/105675984/36629cb5-7632-4bfb-97ef-bd810efa1ae9

3D After: 

https://github.com/godotengine/godot/assets/105675984/28172689-d56c-47a1-9ace-eca004a5dc75

### DIFF
--- a/editor/plugins/canvas_item_editor_plugin.cpp
+++ b/editor/plugins/canvas_item_editor_plugin.cpp
@@ -6027,6 +6027,10 @@ void CanvasItemEditorViewport::_notification(int p_what) {
 		case NOTIFICATION_EXIT_TREE: {
 			disconnect("mouse_exited", callable_mp(this, &CanvasItemEditorViewport::_on_mouse_exit));
 		} break;
+
+		case NOTIFICATION_DRAG_END: {
+			_remove_preview();
+		} break;
 	}
 }
 
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -3017,6 +3017,8 @@ void Node3DEditorViewport::_notification(int p_what) {
 			// Clear preview material when dropped outside applicable object.
 			if (spatial_editor->get_preview_material().is_valid() && !is_drag_successful()) {
 				_remove_preview_material();
+			} else {
+				_remove_preview_node();
 			}
 		} break;
 	}


Sauermann (dismissed):
In its current form, this PR adds an additional call to the Controls `can_drop_data()` just to check within this call, if `ui_cancel` is pressed. This adds unnecessary `can_drop_data()` calls just to get notified, that dragging was cancelled.
When dragging ends, there is a `NOTIFICATION_DRAG_END` sent to the nodes. It would be better to listen for that notification and do something along the idea of the following lines:

``` diff
diff --git a/editor/plugins/canvas_item_editor_plugin.cpp b/editor/plugins/canvas_item_editor_plugin.cpp
index f16aff555c..f8f2f0da7d 100644
--- a/editor/plugins/canvas_item_editor_plugin.cpp
+++ b/editor/plugins/canvas_item_editor_plugin.cpp
@@ -6027,6 +6027,10 @@ void CanvasItemEditorViewport::_notification(int p_what) {
                        disconnect("mouse_exited", callable_mp(this, &CanvasItemEditorViewport::_on_mouse_exit));
                } break;
+
+               case NOTIFICATION_DRAG_END: {
+                       _remove_preview();
+               }
        }
 }
 ```

It is likely that with this approach other calls to hide the label could be removed from the code.

This approach has the advantage, that it contains no changes to the viewport core class and aligns with the ["Prefer local solutions" best practice](https://docs.godotengine.org/en/latest/contributing/development/best_practices_for_engine_contributors.html#prefer-local-solutions).

ryevdokimov:
That makes a lot of sense, thank you. Changes applied.

Sauermann (commented):
It looks right, but a review from someone with more expertise regarding editor plugins would help more, since I'm not familiar with them.

KoBeWi (approved):
Commit message/title probably needs a tweak.

ryevdokimov:
Could be. I'm open to suggestions, how can I improve it?

KoBeWi:
You could use the title of my PR - "Clear editor drag preview when drag ends".

akien-mga:
Thanks!

PR #84438, opened 2023-11-04
Offset Node3D position when drag and dropping

warent:
Fix https://github.com/godotengine/godot/issues/84424

PR #81892, opened 2023-09-19
Add a warning for when the scene root node is transformed

aaronfranke:
This is a low priority PR, but I think it's a good idea.

This PR adds a configuration warning for when the root node has a transform. This is usually a mistake when saving a branch as a scene, or having the wrong node selected when doing transform operations. Instances of the scene will set their own transform, so it's best practice to not have the root be transformed.

Screenshot 2023-09-18 at 5 31 02 PM

PR #87126, opened 2024-01-12
Fix position and basis of 3D Scenes instantiated via drag and drop

### LINKED ISSUE #84424: 3D objects dropped into the viewport do not respect original positioning and reset it to cursor position.
### Godot version

4.2 dev 4

### System information

Windows 11 - Vulkan - Nvidia RTX 4070 - intel i5 13600KF

### Issue description

Just as in the video, if you drag and drop a 3D scene into the 3D viewport the object's position is reset to the cursor position and the original offset in the saved scene is ignored after drag&drop is released (it is kept in while placing the scene in the viewport though).

- In 2D the offset is kept as you would expect so 2D and 3D are behaving here differently too.

Video: https://youtu.be/wNZeh-Km6QA

### Steps to reproduce

Drag and drop "drop_mesh.tscn" 3D scenes files into the viewport, move it around to see that the offset is kept properly then release the mouse button and observe how the object is reset to the cursor position.

What I'd expect is the same behaviour as in 3D that the object's offset would be kept.

[TestSceneDragAndDrop.zip](https://github.com/godotengine/godot/files/13253831/TestSceneDragAndDrop.zip)

ryevdokimov:
Fixes #84424

Supersedes #84438

https://github.com/godotengine/godot/assets/105675984/2029b533-d1c1-4088-a2c2-f85fbdb4e89c

image

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -4371,7 +4371,8 @@ bool Node3DEditorViewport::_create_instance(Node *parent, String &path, const Po
 		}
 
 		Transform3D new_tf = node3d->get_transform();
-		new_tf.origin = parent_tf.affine_inverse().xform(preview_node_pos);
+		new_tf.origin = parent_tf.affine_inverse().xform(preview_node_pos + node3d->get_position());
+		new_tf.basis = parent_tf.affine_inverse().basis * new_tf.basis;
 
 		undo_redo->add_do_method(instantiated_scene, "set_transform", new_tf);
 	}


aaronfranke:
See also https://github.com/godotengine/godot/pull/81892

ryevdokimov:
Don't achieve. I will reopen this, I just screwed up.

akien-mga (approved):
Tested locally, this seems to behave as I would expect it, similar to how it does with 2D nodes.

akien-mga:
Doesn't cherry-pick easily on `4.2`, would need a dedicated backport PR.

PR #87130, opened 2024-01-12
C#: Enable default items for `GodotSharp` & `GodotSharpEditor`

Repiteo:
`bindings_generator.cpp` will now account for potentially unwanted, older versions of files by clearing the generated folder (and subfolders) prior to building glue. This cleans up the `GodotSharp` and `GodotSharpEditor` project files by quite a lot, as they can now safely auto-include their contents like the rest of the repo. `GeneratedIncludes.props` is no longer required and has consequently been removed from the generation process; as such, `dotnet_format.sh` was adjusted to no longer create them as dummy files.

### DIFF
--- a/misc/scripts/dotnet_format.sh
+++ b/misc/scripts/dotnet_format.sh
@@ -5,12 +5,8 @@
 
 set -uo pipefail
 
-# Create dummy generated files.
+# Create dummy generated file.
 echo "<Project />" > modules/mono/SdkPackageVersions.props
-mkdir -p modules/mono/glue/GodotSharp/GodotSharp/Generated
-echo "<Project />" > modules/mono/glue/GodotSharp/GodotSharp/Generated/GeneratedIncludes.props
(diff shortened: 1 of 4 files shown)


raulsntos (commented):
The reason why we had this disabled is to prevent accidentally using old .cs files in the compilation, as explained by the comment:

https://github.com/godotengine/godot/blob/cae7599498070226fecba29d45dffaa36ecf20b4/modules/mono/glue/GodotSharp/GodotSharp/GodotSharp.csproj#L142-L143

I suppose clearing the directory may take care of that, but it feels like the previous system was more robust.

It's a nice cleanup though, so I think we should go for it. But I'd like to know what other C# contributors think first.

paulloz:
I don't particularly have a strong opinion here. I'm kind of on the "if it ain't broke, don't fix it" side. But I see the appeal of having simplified project files.

PR #87131, opened 2024-01-12
Properly change GridMap floors while selecting

### LINKED ISSUE #87107: Gridmap selection hotkeys no longer working since 4.2
### Tested versions

Reproducible in 4.2 onwards. it was working in Godot 3, and it was working up until 4.1.3

### System information

Godot v4.2.stable - Windows 10.0.22621 - Vulkan (Forward+) - integrated AMD Radeon(TM) Graphics (Advanced Micro Devices, Inc.; 31.0.12044.56000) - AMD Ryzen 7 5700U with Radeon Graphics (16 Threads)

### Issue description

You can hold shift and hold the left mouse button to select cells in a gridmap. While holding shift, you used to be able to be able to select multiple floors by also pressing the "e" or "q" keys, as seem in the video below:

https://www.youtube.com/watch?v=xond8D3mf2w&feature=youtu.be

But since 4.2, this stopped working, and now nothing happens, as seem in the video below:

https://www.youtube.com/watch?v=S2RFTIe3XoQ&feature=youtu.be

I am pressing "q" and "e" in the video, but nothing happens

### Steps to reproduce

- Create a Scene with a GridMap
- Create a MeshLibrary and put it in the GridMap
- Try to select multiple floors by holding shift, holding the left mouse button and pressing either "q" or "e"

N/A

KoBeWi:
Fixes #87107

https://github.com/godotengine/godot/assets/2223172/0f4074c5-1d22-4453-8f88-2f649e12d6a5

Note that behavior is still different than in the previous versions. You need to release Shift to use E and Q, also it changes the current selection position, instead of the origin like it did previously.

### DIFF
--- a/modules/gridmap/editor/grid_map_editor_plugin.cpp
+++ b/modules/gridmap/editor/grid_map_editor_plugin.cpp
@@ -60,10 +60,18 @@ void GridMapEditor::_menu_option(int p_option) {
 	switch (p_option) {
 		case MENU_OPTION_PREV_LEVEL: {
 			floor->set_value(floor->get_value() - 1);
+			if (selection.active && input_action == INPUT_SELECT) {
+				selection.current[edit_axis]--;
+				_validate_selection();
+			}
 		} break;
 
 		case MENU_OPTION_NEXT_LEVEL: {
 			floor->set_value(floor->get_value() + 1);
+			if (selection.active && input_action == INPUT_SELECT) {
+				selection.current[edit_axis]++;
+				_validate_selection();
+			}
 		} break;
 
 		case MENU_OPTION_X_AXIS:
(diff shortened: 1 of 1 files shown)


Calinou (commented):
This doesn't seem to work on my end:

https://github.com/godotengine/godot/assets/180032/3c78cc14-f3c6-45dd-9f5b-3a489a88269d

I double-checked my shortcuts and they're set to their default values. I've also released <kbd>Shift</kbd> while pressing <kbd>Q</kbd> and <kbd>E</kbd>.

KoBeWi:
It only works while still dragging. I guess that's another difference, as previously you were able to Shift+Q/E after releasing mouse button (though it feels like a bug).

I opened this PR as a quick bugfix, and to see if such behavior is acceptable. I can try improving it if it's not good enough.

akien-mga:
So should we go ahead with the current partial fix, or spend time to figure out how to solve the issue when not dragging?

BrentonWildes:
> So should we go ahead with the current partial fix, or spend time to figure out how to solve the issue when not dragging?

Please move it forward! This is blocking the Lingo team from updating beyond 4.1.

akien-mga (approved):
Approving as per above discussion.

PR #87135, opened 2024-01-13
Clear drag preview material on `NOTIFICATION_DRAG_END`

ryevdokimov:
Related to: #87125

Currently when dragging a material onto a target it will still apply even after the action 'ui_cancel'. This PR will prevent that.


KoBeWi (approved):

akien-mga:
Cherry-picked for 4.3.1.

PR #83117, opened 2023-10-10
C#: Enable nullable environment for GodotSharp

Repiteo:
This PR showcases a GodotSharp environment where nullability is enabled. While [initially conceived](https://github.com/godotengine/godot/pull/82983#issuecomment-1752047518) as a means to centralize changes which can be branched off into standalone PRs (eg: #82980, #82983), it also showcases the consequences of trying to retrofit the concept without first undergoing significant refactoring

The biggest culprit is `GodotObject`, and how it's uniquely a reference type that Godot allows for nullability. While this draft explores the idea of returning GodotObject fields as nullable, I'm not convinced that makes for the best approach, because it floods almost every existing method/field with an added nullability check. It might be worth considering an approach similar to [Unity Objects](https://github.com/Unity-Technologies/UnityCsReference/blob/master/Runtime/Export/Scripting/UnityEngineObject.bindings.cs#L103), where there's operator overloads & behind-the-scenes null checks to better handle `null` in the context of that environment. Or it could even be as simple as throwing in a bunch of `[AllowNull]/[MaybeNull]` annotations in the generation for object files

While the majority of areas have warnings accounted for, there's still a handful that have them. This is mainly those that I'm not entirely sure yet how to handle, or areas that feel particularly sensitive to adjustment. It's also entirely possible that many of these warnings/cases simply wouldn't be a thing with future core additions (particularly, some means of knowing if an object return value *can* be null). Either way, this is provided as-is for discussion and so potential areas can be cherrypicked for implementation as needed

PR #87137, opened 2024-01-13
C#: Enable nullable environment for `GodotTools`

Repiteo:
Enables nullability in majority of `GodotTools` projects. At first I didn't think this would be possible without `GodotSharp` getting nullable first (see #83117), but it turns out there was quite a lot that could be setup as-is! The only real area that didn't get expanded upon were certain `GodotObject` fields (mirroring the selectively disabled nullability seen on a handful of files beforehand), which is likely to be more feasible when `GodotSharp` gets full-project nullability.

### DIFF
--- a/modules/mono/editor/GodotTools/GodotTools.BuildLogger/GodotBuildLogger.cs
+++ b/modules/mono/editor/GodotTools/GodotTools.BuildLogger/GodotBuildLogger.cs
@@ -7,11 +7,11 @@ namespace GodotTools.BuildLogger
 {
     public class GodotBuildLogger : ILogger
     {
-        public string Parameters { get; set; }
+        public string? Parameters { get; set; }
         public LoggerVerbosity Verbosity { get; set; }
 
-        private StreamWriter _logStreamWriter;
-        private StreamWriter _issuesStreamWriter;
+        private StreamWriter _logStreamWriter = StreamWriter.Null;
+        private StreamWriter _issuesStreamWriter = StreamWriter.Null;
         private int _indent;
 
         public void Initialize(IEventSource eventSource)
--- a/modules/mono/editor/GodotTools/GodotTools.Core/StringExtensions.cs
+++ b/modules/mono/editor/GodotTools/GodotTools.Core/StringExtensions.cs
@@ -7,7 +7,7 @@ namespace GodotTools.Core
 {
     public static class StringExtensions
     {
-        private static readonly string _driveRoot = Path.GetPathRoot(Environment.CurrentDirectory);
+        private static readonly string _driveRoot = Path.GetPathRoot(Environment.CurrentDirectory)!;
 
         public static string RelativeToPath(this string path, string dir)
         {
@@ -26,9 +26,6 @@ namespace GodotTools.Core
 
         public static string NormalizePath(this string path)
         {
-            if (string.IsNullOrEmpty(path))
-                return path;
-
             bool rooted = path.IsAbsolutePath();
 
             path = path.Replace('\\', '/');
--- a/modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Client.cs
+++ b/modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Client.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Collections.Generic;
 using System.IO;
 using System.Net;
@@ -27,7 +28,7 @@ namespace GodotTools.IdeMessaging
 
         private readonly IMessageHandler messageHandler;
 
-        private Peer peer;
+        private Peer? peer;
         private readonly SemaphoreSlim connectionSem = new SemaphoreSlim(1);
 
         private readonly Queue<NotifyAwaiter<bool>> clientConnectedAwaiters = new Queue<NotifyAwaiter<bool>>();
@@ -53,6 +54,7 @@ namespace GodotTools.IdeMessaging
         public bool IsDisposed { get; private set; }
 
         // ReSharper disable once MemberCanBePrivate.Global
+        [MemberNotNullWhen(true, "peer")]
         public bool IsConnected => peer != null && !peer.IsDisposed && peer.IsTcpClientConnected;
 
         // ReSharper disable once EventNeverSubscribedTo.Global
@@ -111,7 +113,7 @@ namespace GodotTools.IdeMessaging
             if (disposing)
             {
                 peer?.Dispose();
-                fsWatcher?.Dispose();
+                fsWatcher.Dispose();
             }
         }
 
@@ -215,12 +217,12 @@ namespace GodotTools.IdeMessaging
             using (var fileStream = new FileStream(MetaFilePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
             using (var reader = new StreamReader(fileStream))
             {
-                string portStr = reader.ReadLine();
+                string? portStr = reader.ReadLine();
 
                 if (portStr == null)
                     return null;
 
-                string editorExecutablePath = reader.ReadLine();
+                string? editorExecutablePath = reader.ReadLine();
 
                 if (editorExecutablePath == null)
                     return null;
@@ -333,7 +335,7 @@ namespace GodotTools.IdeMessaging
             }
         }
 
-        public async Task<TResponse> SendRequest<TResponse>(Request request)
+        public async Task<TResponse?> SendRequest<TResponse>(Request request)
             where TResponse : Response, new()
         {
             if (!IsConnected)
@@ -346,7 +348,7 @@ namespace GodotTools.IdeMessaging
             return await peer.SendRequest<TResponse>(request.Id, body);
         }
 
-        public async Task<TResponse> SendRequest<TResponse>(string id, string body)
+        public async Task<TResponse?> SendRequest<TResponse>(string id, string body)
             where TResponse : Response, new()
         {
             if (!IsConnected)
--- a/modules/mono/editor/GodotTools/GodotTools.IdeMessaging/ClientHandshake.cs
+++ b/modules/mono/editor/GodotTools/GodotTools.IdeMessaging/ClientHandshake.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics.CodeAnalysis;
 using System.Text.RegularExpressions;
 
 namespace GodotTools.IdeMessaging
@@ -9,7 +10,7 @@ namespace GodotTools.IdeMessaging
 
         public string GetHandshakeLine(string identity) => $"{ClientHandshakeBase},{identity}";
 
-        public bool IsValidPeerHandshake(string handshake, out string identity, ILogger logger)
+        public bool IsValidPeerHandshake(string handshake, [NotNullWhen(true)] out string? identity, ILogger logger)
         {
             identity = null;
 
--- a/modules/mono/editor/GodotTools/GodotTools.IdeMessaging/GodotIdeMetadata.cs
+++ b/modules/mono/editor/GodotTools/GodotTools.IdeMessaging/GodotIdeMetadata.cs
@@ -1,3 +1,5 @@
+using System.Diagnostics.CodeAnalysis;
+
 namespace GodotTools.IdeMessaging
 {
     public readonly struct GodotIdeMetadata
@@ -23,7 +25,7 @@ namespace GodotTools.IdeMessaging
             return !(a == b);
         }
 
-        public override bool Equals(object obj)
+        public override bool Equals([NotNullWhen(true)] object? obj)
         {
             return obj is GodotIdeMetadata metadata && metadata == this;
         }
--- a/modules/mono/editor/GodotTools/GodotTools.IdeMessaging/IHandshake.cs
+++ b/modules/mono/editor/GodotTools/GodotTools.IdeMessaging/IHandshake.cs
@@ -1,8 +1,10 @@
+using System.Diagnostics.CodeAnalysis;
+
 namespace GodotTools.IdeMessaging
 {
     public interface IHandshake
     {
         string GetHandshakeLine(string identity);
-        bool IsValidPeerHandshake(string handshake, out string identity, ILogger logger);
+        bool IsValidPeerHandshake(string handshake, [NotNullWhen(true)] out string? identity, ILogger logger);
     }
 }
--- a/modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Peer.cs
+++ b/modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Peer.cs
@@ -46,11 +46,11 @@ namespace GodotTools.IdeMessaging
 
         private readonly SemaphoreSlim writeSem = new SemaphoreSlim(1);
 
-        private string remoteIdentity = string.Empty;
-        public string RemoteIdentity => remoteIdentity;
+        private string? remoteIdentity;
+        public string RemoteIdentity => remoteIdentity ??= string.Empty;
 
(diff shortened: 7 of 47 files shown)


Repiteo:
Updated to include `GodotTools.IdeMessaging` as well; while it's still a touchy file, I realized that it was perfectly safe to upgrade to C# 9 without issue, which does include nullability!

YuriSizov:

Repiteo:
Done and done!

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/ClientHandshake.cs
```diff
@@ -9,7 +9,7 @@ public class ClientHandshake : IHandshake
 
         public string GetHandshakeLine(string identity) => $"{ClientHandshakeBase},{identity}";
 
-        public bool IsValidPeerHandshake(string handshake, out string identity, ILogger logger)
+        public bool IsValidPeerHandshake(string handshake, out string? identity, ILogger logger)
```
raulsntos, line 12: Seems like this parameter is missing the nullability attribute.

```suggestion
        public bool IsValidPeerHandshake(string handshake, [NotNullWhen(true)] out string? identity, ILogger logger)
```

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/GodotIdeMetadata.cs
```diff
@@ -23,7 +23,7 @@ public GodotIdeMetadata(int port, string editorExecutablePath)
             return !(a == b);
         }
 
-        public override bool Equals(object obj)
+        public override bool Equals(object? obj)
```
raulsntos, line 26:
```suggestion
        public override bool Equals([NotNullWhen(true)] object? obj)
```

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/IHandshake.cs
```diff
@@ -3,6 +3,6 @@ namespace GodotTools.IdeMessaging
     public interface IHandshake
     {
         string GetHandshakeLine(string identity);
-        bool IsValidPeerHandshake(string handshake, out string identity, ILogger logger);
+        bool IsValidPeerHandshake(string handshake, out string? identity, ILogger logger);
```
raulsntos, line 6:
```suggestion
        bool IsValidPeerHandshake(string handshake, [NotNullWhen(true)] out string? identity, ILogger logger);
```

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Client.cs
```diff
@@ -343,10 +343,10 @@ public async Task<TResponse> SendRequest<TResponse>(Request request)
             }
 
             string body = JsonConvert.SerializeObject(request);
-            return await peer.SendRequest<TResponse>(request.Id, body);
+            return await peer!.SendRequest<TResponse>(request.Id, body);
```
raulsntos, line 346: `IsConnected` implies that `peer` is not null so we should annotate `IsConnected` instead of using the null forgiveness operator here.

```csharp
[MemberNotNull(nameof(peer))]
public bool IsConnected => ...
```

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Utils/NotifyAwaiter.cs
```diff
@@ -25,7 +28,7 @@ public void OnCompleted(Action continuation)
             this.continuation = continuation;
         }
 
-        public void SetResult(T result)
+        public void SetResult(T? result)
```
raulsntos, line 31: I don't think this should be nullable.

```suggestion
        public void SetResult(T result)
```

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Utils/NotifyAwaiter.cs
```diff
@@ -1,17 +1,20 @@
...
-        private Exception exception;
-        private T result;
+        private Action? continuation;
+        private Exception? exception;
+        private T? result;
 
         public bool IsCompleted { get; private set; }
 
-        public T GetResult()
+        public T? GetResult()
```
raulsntos, line 17: This should probably not be nullable.

```suggestion
        public T GetResult()
```

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Peer.cs
```diff
@@ -238,12 +238,12 @@ public async Task<TResponse> SendRequest<TResponse>(string id, string body)
                 queue.Enqueue(responseAwaiter);
             }
 
-            return (TResponse)await responseAwaiter;
+            return (TResponse?)await responseAwaiter;
```
raulsntos, line 241: It doesn't look like `ResponseAwaiter` ever sets the result to null.

raulsntos (commented):
Thanks for taking the time to work on this. I think it might've been easier to split the PR by csproj but what's done is done.

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/IHandshake.cs
Repiteo: Surprisingly, this attribute isn't available in .NET Standard 2.0

modules/mono/editor/GodotTools/GodotTools/Export/XcodeHelper.cs
```diff
@@ -40,20 +40,20 @@ private static string FindSelectedXcode()
...
                     return selectedXcode;
 
                 // The path already pointed to Contents/Developer
                 var dirInfo = new DirectoryInfo(selectedXcode);
-                if (dirInfo.Name != "Developer" || dirInfo.Parent.Name != "Contents")
+                if (dirInfo.Name != "Developer" || dirInfo.Parent?.Name != "Contents")
                 {
                     Console.WriteLine(Path.GetDirectoryName(selectedXcode));
-                    Console.WriteLine(System.IO.Directory.GetParent(selectedXcode).Name);
+                    Console.WriteLine(System.IO.Directory.GetParent(selectedXcode)?.Name);
```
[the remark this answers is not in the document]
Repiteo: I don't believe we'd want this to throw an exception regardless; the error that gets printed already showcases an unrecognized path, so this just ensures that gets through

modules/mono/editor/GodotTools/GodotTools/Export/XcodeHelper.cs
```diff
@@ -40,20 +40,20 @@ private static string FindSelectedXcode()
...
+            string? selectedXcode = FindSelectedXcode();
             if (selectedXcode != null)
             {
                 if (Directory.Exists(Path.Combine(selectedXcode, "Contents", "Developer")))
                     return selectedXcode;
 
                 // The path already pointed to Contents/Developer
                 var dirInfo = new DirectoryInfo(selectedXcode);
-                if (dirInfo.Name != "Developer" || dirInfo.Parent.Name != "Contents")
+                if (dirInfo.Name != "Developer" || dirInfo.Parent?.Name != "Contents")
```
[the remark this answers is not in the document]
Repiteo: Had no clue this syntax existed, neat!

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/IHandshake.cs
raulsntos: Well, that's not really surprising since these attributes didn't exist when .NET Standard 2.0 came out. But yeah I missed that this project is targeting `netstandard2.0`. We can, however, add the attribute ourselves since the compiler only cares about the shape.

See https://github.com/godotengine/godot/blob/0cce6eb1505f7dd9c0cece2a825b1509cf23f158/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/CodeAnalysisAttributes.cs

modules/mono/editor/GodotTools/GodotTools/Export/XcodeHelper.cs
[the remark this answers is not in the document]
raulsntos: It's somewhat new and I haven't really checked if this project is on a new enough version of C#. A slightly older version would be:

```csharp
if (dirInfo is not { Parent: { Name: "Contents" }, Name: "Developer" })
```

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Peer.cs
Repiteo: It doesn't, but it stops the IDE from whining about converting to a possible null value, despite recognizing that `responseAwaiter` isn't null here

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/IHandshake.cs
Repiteo: Huh! Well that makes matters more convenient

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Peer.cs
raulsntos: It could be because you changed `GetResult` in `NotifyAwaiter` which `ResponseAwaiter` derives from: https://github.com/godotengine/godot/pull/87137#discussion_r1471668129

modules/mono/editor/GodotTools/GodotTools/Ides/GodotIdeManager.cs
```diff
@@ -59,7 +59,7 @@ private string GetExternalEditorIdentity(ExternalEditorId editorId)
             switch (editorId)
             {
                 case ExternalEditorId.None:
-                    return null;
+                    return string.Empty;
```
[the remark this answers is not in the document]
Repiteo: Converting to `string.Empty` felt more appropriate for the enum-to-string usecase, especially when it would eventually need to be converted regardless to work as a dictionary key

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Peer.cs
Repiteo: That's exactly what's happening, yeah. I assumed that we wanted `GetResult` to be nullable though, as the result itself could be null and we don't have a safeguard against that. Would it be preferable to make it non-nullable & throw an exception if `result` is null?

Repiteo: So something like:
```cs
public T GetResult() => result ?? throw exception ?? new ArgumentNullException(nameof(result));
```

modules/mono/editor/GodotTools/GodotTools/GodotSharpEditor.cs
```diff
@@ -43,14 +44,17 @@ public static class Settings
 
         private Button _bottomPanelBtn;
         private Button _toolBarBuildButton;
+#nullable enable
 
         // TODO Use WeakReference once we have proper serialization.
-        private WeakRef _exportPluginWeak;
-        private WeakRef _inspectorPluginWeak;
+        private WeakRef? _exportPluginWeak;
+        private WeakRef? _inspectorPluginWeak;
```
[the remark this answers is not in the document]
Repiteo: These should absolutely be part of the `#nullable disable` group, don't know how I missed that

modules/mono/editor/GodotTools/GodotTools.IdeMessaging/Peer.cs
raulsntos: As I understand it, result will never be null if it was set from `SetResult`, only if the await completes with `SetException`. In this case, since exception is not null it will be thrown in `GetResult` already, so I'd say using the null forgiveness operator may be correct here.

```csharp
public T GetResult()
{
    if (exception != null)
        throw exception;
    return result!;
}
```

modules/mono/editor/GodotTools/GodotTools.BuildLogger/GodotBuildLogger.cs
```diff
@@ -7,11 +7,11 @@ namespace GodotTools.BuildLogger
 {
     public class GodotBuildLogger : ILogger
     {
-        public string Parameters { get; set; }
+        public string? Parameters { get; set; }
```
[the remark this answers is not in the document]
Repiteo: Nullability is disabled for that linked file overall, but the description of the property is: "The parameter string (can be null)". This certainly makes it seem like making it nullable is the correct call in a null-aware context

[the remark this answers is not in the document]
raulsntos: I agree. Like I said in https://github.com/godotengine/godot/pull/87137#discussion_r1475452039 I'm not aware of any guidelines to follow in this case so I was only bringing attention to it, but it's probably fine to leave it as-is.

modules/mono/editor/GodotTools/GodotTools.Core/StringExtensions.cs
```diff
@@ -7,7 +7,7 @@ namespace GodotTools.Core
 {
     public static class StringExtensions
```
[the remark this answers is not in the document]
Repiteo: The check for an empty string is still relevant though, as a later part of the script relies on `path[path.Length - 1]`, which would break if the string was empty. Though after looking at some of the native `Path` functions, it seems that exceptions are thrown if the path is null or empty, so perhaps those exceptions can simply be added to the `StringExtensions` functions?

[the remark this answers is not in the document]
Repiteo: For now, I'll make that check return `string.Empty` to guarantee the return value isn't `null`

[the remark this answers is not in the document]
raulsntos: I'm not sure I understand, the method `NormalizePath` has a non-nullable string parameter so callers shouldn't be passing null, and since this is not public API we can guarantee that will never happen because we are always the caller. Are you saying we sometimes use null with this method?

About changing the return in the check to `string.Empty`, I'm not sure why that makes a difference since you are returning so you never reach the later part of the script that uses `path[path.Length - 1]`.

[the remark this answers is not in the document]
Repiteo: We don't use `null` with that method, the `IsNullOrEmpty` really only checks for the empty string in practice. As such, it could be trimmed down to just `if (path == string.Empty)`

The return of `string.Empty` would only really be needed in the freak scenario where `null` *was* passed, which would make it return an empty string as if it were simply passed an empty string. So it's for the sake of ensuring a valid return if it were to return early; passing that first check means it will always have a length greater than zero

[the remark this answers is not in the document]
raulsntos: But is `string.Empty` ever passed? I feel like we always pass a non-empty string to that method but if we want to still check for this I would probably prefer to keep the `string.IsNullOrEmpty`.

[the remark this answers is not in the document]
Repiteo: It doesn't seem like it. I suppose then that the check can be removed altogether, which has precedence with several of our other path-related functions lacking such a check. It'd probably be worth having explicit exception thrown in several of these functions, but that's outside the scope of this PR

raulsntos (approved):
LGTM. Thanks!

akien-mga:
Thanks!

PR #80706, opened 2023-08-17
Rework `update_property` for array

ajreckof:
Fixes #75124

This is  a huge rework of how array property editor is updated. Previously it would delete every editor for each element and recreate them. Now it first check whether the property editor will be the same. 

This is not 100% tested but I couldn't find other bugs so decided to share to see if other people could find some bugs while I'm finishing testing otr if they had any suggestion on the code

PR #87140, opened 2024-01-13
Fix nested array editor not being updated on type change.

### LINKED ISSUE #87139: Nested Array editors do not update automatically
### Tested versions

most likely a regression from #80706

### System information

### Issue description

When changing the type of a nested Array nothing gets updated until you close and open it again. 

### Steps to reproduce

- create a script and export an array
- create a nested array
- try changing the type nothing happens close and re-open the array editor change will be shown 

N/A

ajreckof:
Fix #87139

### DIFF
--- a/editor/editor_properties_array_dict.cpp
+++ b/editor/editor_properties_array_dict.cpp
@@ -207,6 +207,9 @@ void EditorPropertyArray::_property_changed(const String &p_property, Variant p_
 	Variant array = object->get_array().duplicate();
 	array.set(index, p_value);
 	emit_changed(get_edited_property(), array, p_name, p_changing);
+	if (!p_changing) {
+		update_property();
+	}
 }
 
 void EditorPropertyArray::_change_type(Object *p_button, int p_slot_index) {
@@ -230,7 +233,7 @@ void EditorPropertyArray::_change_type_menu(int p_index) {
 	Variant array = object->get_array().duplicate();
 	array.set(changing_type_index, value);
 
-	emit_changed(get_edited_property(), array, "", true);
+	emit_changed(get_edited_property(), array);
 	update_property();
 }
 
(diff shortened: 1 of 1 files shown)


editor/editor_properties_array_dict.cpp
```diff
@@ -230,8 +230,7 @@ void EditorPropertyArray::_change_type_menu(int p_index) {
 	Variant array = object->get_array().duplicate();
 	array.set(changing_type_index, value);
 
-	emit_changed(get_edited_property(), array, "", true);
-	update_property();
+	emit_changed(get_edited_property(), array);
```
AThousandShips, line 233: This changes a relevant part of the original code, was the removal of the arguments intentional? It's not the same as the rest, unsure if the `true` for `p_changing` is relevant here

ajreckof: This is the main fix the argument here was preventing the updates from being from fired from the root Inspector which was the main reasons of the bug here. The rest is mainly preventing multiple updates and removing default arguments from function call ("" and false are the default values). I just realized with your comment that there was only here that it was to true and that therefore only changing type was not working.

ajreckof:
#80706 was not marked as cherrypick (maybe wrongly) and this fix a regression introduced there so either this should not be cherrypicked or the other should too.

editor/editor_properties_array_dict.cpp
AThousandShips: Do you mean the removal of `update_property`? Or did you just also remove the default arguments at the same time?

ajreckof: the removal of update property is to prevent multiple updates. 
Removing the default argument ("" and false) does nothing and is just a by-product of how I made my search and replace.

ajreckof:
Discovered the double update is needed until the same is done for Dictionnaries as it breaks without it for now when you nest an array under a dictionnary.

akien-mga:
Superseded by #88231.

PR #84913, opened 2023-11-14
Validate texture format selection when exporting to Windows or Linux

clayjohn:
Fixes: https://github.com/godotengine/godot/issues/81234
Closes: https://github.com/godotengine/godot-proposals/issues/7230

In https://github.com/godotengine/godot/pull/73829 we made s3tc + bptc the default export options. But we exposed users to the risk that, if they disable one or the other, their games would not load. 

In Godot 4.x, it is not possible to use S3TC without BPTC (or vice versa). Nor is it possible to use ETC2 without ASTC (or vice versa). 

This change validates that S3TC and BPTC have the same value (either both on or both off) and either S3TC/BPTC or ETC2/ASTC are used at export time. 

Finally it updates the docs for the settings and removes "etc" from the feature list (as ETC textures are no longer used by the engine).

PR #87144, opened 2024-01-13
Use ETC2 compression for grayscale images

BlueCube3310:
Sets the default compression type for grayscale textures to the higher-quality ETC2 format instead of ETC1.

Textures can still be explicitly compressed as ETC1 by using `Image`'s `compress` function, although [this PR](https://github.com/godotengine/godot/pull/84913#issue-1993603194) mentions that the ETC1 format is no longer used by Godot, so should compression support for it be removed entirely?

MRP for testing: [etc1grayscale.zip](https://github.com/godotengine/godot/files/13927998/etc1grayscale.zip)

### DIFF
--- a/modules/etcpak/image_compress_etcpak.cpp
+++ b/modules/etcpak/image_compress_etcpak.cpp
@@ -40,7 +40,7 @@
 EtcpakType _determine_etc_type(Image::UsedChannels p_channels) {
 	switch (p_channels) {
 		case Image::USED_CHANNELS_L:
-			return EtcpakType::ETCPAK_TYPE_ETC1;
+			return EtcpakType::ETCPAK_TYPE_ETC2;
 		case Image::USED_CHANNELS_LA:
 			return EtcpakType::ETCPAK_TYPE_ETC2_ALPHA;
 		case Image::USED_CHANNELS_R:


fire:
I don’t think this changes the quality you’d need to go to the bptc or astc formats to get better qualify

Calinou:
> I don’t think this changes the quality you’d need to go to the bptc or astc formats to get better qualify

ETC2 has much better quality than ETC1. It's close to S3TC but not exactly matching it either.

fire:
Are you sure for a single channel? I wasn’t able to explore this. But we should be able to get a visualization as png.

BlueCube3310:
Here is a comparison between etc1 and etc2, both compressed with `etcpak`:

|source|etc1|etc2|diff|
|-------|----|----|----|
|road-1072823_1280|in1|in|diff|

Image by [Valiphotos](https://pixabay.com/users/valiphotos-1720744/) (https://pixabay.com/photos/road-forest-fall-autumn-season-1072823/)

fire (approved):
I can visually tell when ETC1 is used vs ETC2 is used in the png comparison so it should be ok.

Edited:

It was hard to compare ETC2 vs original. So that's good.

akien-mga:
Thanks!

Issue #72246, opened 2023-01-28
[Build System] compling with MSVC (14.3) is verbose


kuhnchris:
### Godot version

master

### System information

Win 11

### Issue description

Debug output, incase it's specific to a certain MSVC version:
```
c:\Dev\godot>scons separate_debug_symbols=true debug_symbols=true
scons: Reading SConscript files ...
Automatically detected platform: windows
Auto-detected 12 CPU cores available for build parallelism. Using 11 cores by default. You can override it with the -j argument.
Found MSVC version 14.3, arch x86_64
Building for platform "windows", architecture "x86_64", target "editor".
Checking for C header file mntent.h... no
scons: done reading SConscript files.
scons: Building targets ...
[  0%] Compiling platform\windows\console_wrapper_windows.cpp ...
[  0%] console_wrapper_windows.cpp
[  3%] Compiling main\main.cpp ...
[  3%] Compiling main\main_timer_sync.cpp ...
[  3%] Compiling main\performance.cpp ...
[  3%] main.cpp
main_timer_sync.cpp
performance.cpp
```
(weirdly the logs shows it in order, the real output looks like this:
![image](https://user-images.githubusercontent.com/761911/215266047-fb06f36f-2556-4323-a478-65640948cc8d.png)

)

the lines without the `[ x% ]` are the output of MSVC, apparently.
Doesn't happen when using MinGW/gcc/linux.
Using "debug_symbols=true" or "separate_debug_symbols=true" is irrelevant, it happens even without them.
Low priority as it's just a "optics" thing (and possibly increasing build log size)

### Steps to reproduce

* Have MSVC
* Compile godot (master) from source.

### Minimal reproduction project

N/A

Kabiirk:
Hi,

I'm still going through the Godot source code, as per my understanding, In the ```SConstruct``` file, the below code is executed :
https://github.com/godotengine/godot/blob/218bef90af2091afde3b1be816c87286c194a2a8/SConstruct#L955-L960 

which calls the below code
https://github.com/godotengine/godot/blob/218bef90af2091afde3b1be816c87286c194a2a8/methods.py#L1034-L1047

and somewhere, these variables are being called as well
https://github.com/godotengine/godot/blob/218bef90af2091afde3b1be816c87286c194a2a8/methods.py#L533-L556

After going through the file, could it have something to do with the ```flush()``` in ```methods.py``` ?

[...]

PR #85153, opened 2023-11-20
SCons: Add two new `COMSTR` environment variables

Repiteo:
This PR aims to make the output logs of a non-verbose build more uniform by adding two new environment variables: `RCCOMSTR` and `GENCOMSTR`. `RCCOMSTR` is an existing keyword for SCons, used for compiling `.rc` files into `.res`/`.o`, so adding it eliminates the last remaining verbose output (on Windows anyway). `GENCOMSTR` is a bit more involved, as it effectively replaces the `methods.Run` "short_message" argument. However, because it's now formatted to automatically reference the generated file, the original messages are no longer needed

PR #87154, opened 2024-01-13
SCons: Silence redundant MSVC output

Repiteo:
"Closes" #72246. Technically, that issue is already closed & marked as "not planned", as this is actually an msvc bug[^1]. Nonetheless, I've figured out a workaround for the majority of the issues caused by this: redirecting the `stdout` to `NULL` where supported. This works like a charm when compiling objects and resources, effectively reducing output bloat by half! ~~I haven't found a workaround for the linker, but there's only ever less than a handful of files linked in a given build, so it's much less of a concern.~~ In practice, the output logs are significantly cleaner & much more closely resemble `gcc`/`clang`, and preserving `stderr` allows potential errors to still be caught and logged as before

Before:
```
[  9%] Compiling thirdparty\libvorbis\sharedbook.c ...
mapping0.c
[  9%] Compiling thirdparty\libvorbis\smallft.c ...
mdct.c
[  9%] Compiling thirdparty\libvorbis\synthesis.c ...
psy.c
registry.c
sharedbook.c
res0.c
smallft.c
synthesis.c
[  9%] Compiling modules\msdfgen\register_types.cpp ...
[  9%] Compiling thirdparty\libvorbis\vorbisfile.c ...
[  9%] Compiling thirdparty\libvorbis\window.c ...
```

After:
```
[  9%] Compiling thirdparty\libvorbis\res0.c ...
[  9%] Compiling thirdparty\libvorbis\sharedbook.c ...
[  9%] Compiling thirdparty\libvorbis\smallft.c ...
[  9%] Compiling thirdparty\libvorbis\synthesis.c ...
[  9%] Compiling modules\theora\video_stream_theora.cpp ...
[  9%] Compiling modules\theora\register_types.cpp ...
[  9%] Compiling modules\msdfgen\register_types.cpp ...
[  9%] Compiling thirdparty\libvorbis\vorbisfile.c ...
[  9%] Compiling thirdparty\libvorbis\window.c ...
[ 10%] Compiling thirdparty\astcenc\astcenc_averages_and_directions.cpp ...
[ 10%] Compiling thirdparty\astcenc\astcenc_block_sizes.cpp ...
[ 10%] Compiling thirdparty\astcenc\astcenc_color_quantize.cpp ...
[ 10%] Compiling thirdparty\astcenc\astcenc_color_unquantize.cpp ...
[ 10%] Compiling thirdparty\astcenc\astcenc_compress_symbolic.cpp ...
```

[^1]: https://developercommunity.visualstudio.com/t/allow-having-clexe-not-print-the-compiled-source-f/717761

### DIFF
--- a/platform/windows/detect.py
+++ b/platform/windows/detect.py
@@ -186,6 +186,7 @@ def get_opts():
         BoolVariable("use_asan", "Use address sanitizer (ASAN)", False),
         BoolVariable("debug_crt", "Compile with MSVC's debug CRT (/MDd)", False),
         BoolVariable("incremental_link", "Use MSVC incremental linking. May increase or decrease build times.", False),
+        BoolVariable("silence_msvc", "Silence MSVC's stdout. Decreases output log bloat by roughly half.", True),
         ("angle_libs", "Path to the ANGLE static libraries", ""),
         # Direct3D 12 support.
         ("mesa_libs", "Path to the MESA/NIR static libraries (required for D3D12)", ""),
@@ -357,6 +358,20 @@ def configure_msvc(env, vcvars_msvc_config):
         env.AppendUnique(CPPDEFINES=["WINDOWS_SUBSYSTEM_CONSOLE"])
 
     ## Compile/link flags
+
+    env["MAXLINELENGTH"] = 8192  # Windows Vista and beyond, so always applicable.
+
+    if env["silence_msvc"]:
+        env.Prepend(CCFLAGS=[">", "NUL"])
+        env.Prepend(LINKFLAGS=[">", "NUL"])
+
+        # "> NUL" fails if using a tempfile, circumvent by removing the argument altogether.
+        old_esc_func = env["TEMPFILEARGESCFUNC"]
+
+        def trim_nul(arg):
+            return "" if arg in [">", "NUL"] else old_esc_func(arg)
+
+        env["TEMPFILEARGESCFUNC"] = trim_nul
 
     if env["debug_crt"]:
         # Always use dynamic runtime, static debug CRT breaks thread_local.


Repiteo:
Hmm, looks like the build actions have some discrepancy to locally building, causing the output to fail. Looking into this now

Repiteo:
Alright, figured it out! The method worked *if* the command wasn't redirected to a temp file; Otherwise, the `> NUL` command failed to process as expected. This would also explain why linking didn't seem to work, as it was always calling to a tempfile as well. This is now dealt with in two ways:

1. Increased the `MAXLINELENGTH` environment variable to 8192. It's normally set to 2048 for Windows, but it can be quadrupled to 8192 for Vista and later, which will always be applicable for us. It seems that there's not a situation where that increased line length is exceeded for objects/links.

2. Remove `> NUL` from the arguments if taken to a tempfile. In the offchance that a commandline *does* exceed the above limit, we'll just not bother at all and revert back to the old behavior, ensuring that the build will process & succeed no matter what.

Unfortunately not the silver-bullet solution I was hoping for, as there's now a scenario where an obscenely long commandline will have the old behavior. But in practice, it shouldn't affect us at all, and as a bonus we now have linker output suppressed as well!

Calinou (approved):
Tested locally (with `progress=no` and rebased on top of `master` https://github.com/godotengine/godot/commit/dad6c774b019ef8c5dccb4a1955c6a77b41a323e), it works as expected on Windows 11 23H2 + Visual Studio 2022.

image

It also works when applied at the same time as https://github.com/godotengine/godot/pull/85153.

akien-mga:
Thanks!

Issue #87138, opened 2024-01-13
Tilemap editor adds extra pixel for odd numbered tile resolutions


IndieJoJo:
### Tested versions

Reproducible in Godot v4.2.1

### System information

Godot v4.2.1.stable - Windows 10.0.19045 - Vulkan (Mobile) - dedicated AMD Radeon RX 6700 XT (Advanced Micro Devices, Inc.; 31.0.23013.1023) - AMD Ryzen 5 5600 6-Core Processor (12 Threads)

### Issue description

When adding a tilemap and setting layers with Reset to default tile shape, odd numbered tile resolutions like (9x9 and 5x5) have an additional pixel added which can mess with physics and navigation.
![image](https://github.com/godotengine/godot/assets/136153464/d318d84b-a25a-4d1d-99e7-3b281dd77468)

### Steps to reproduce

Try adding a layer and pressing reset to default tile shape for any odd numbered resolution.

### Minimal reproduction project (MRP)

[TilemapTest - Copy.zip](https://github.com/godotengine/godot/files/13927140/TilemapTest.-.Copy.zip)

kleonc:
There are indeed truncations of floats to ints in few places resulting in incorrect offsets, PR fixing these incoming. AFAICT these affect only the debug drawing though.

---

What's a bigger problem is that `NavigationServer2D::bake_from_source_geometry_data` seems to produce incorrect navigation polygon.

Specifically here:
https://github.com/godotengine/godot/blob/26b1fd0d842fa3c2f090ead47e8ea7cd2d6515e1/editor/plugins/tiles/tile_data_editors.cpp#L2831-L2843
for a single outline `[(-2.5, -2.5), (2.5, -2.5), (2.5, 2.5), (-2.5, 2.5)]` it produces a navigation polygon with vertices `[(3.0, 3.0), (-3.0, 3.0), (-3.0, -3.0), (3.0, -3.0)]` (a single polygon with indices `[0, 1, 2, 3]`).

More specifically the rounding happens here when converting `Vector2` to `Point64`:
https://github.com/godotengine/godot/blob/26b1fd0d842fa3c2f090ead47e8ea7cd2d6515e1/modules/navigation/nav_mesh_generator_2d.cpp#L742-L745
https://github.com/godotengine/godot/blob/26b1fd0d842fa3c2f090ead47e8ea7cd2d6515e1/thirdparty/clipper2/include/clipper2/clipper.core.h#L165-L166
[...]

PR #87160, opened 2024-01-14
Fix some `TileMap` debug drawing truncating to ints

kleonc:
Fixes part of #87138 (besides the incorrect generation of the navigation polygon (see https://github.com/godotengine/godot/issues/87138#issuecomment-1890810824)).

- `quadrant_pos` was correctly calculated as `Vector2` but it was converted to `Vector2i` when passing to `_X_draw_cell_debug` methods:
https://github.com/godotengine/godot/blob/26b1fd0d842fa3c2f090ead47e8ea7cd2d6515e1/scene/2d/tile_map_layer.cpp#L126-L136

- `Rect2i::get_center` returning `Vector2i` instead of `Vector2` strikes again. This time making the offset for the texture region center being calculated incorrectly.
Still no idea why this method works like that. For me it's like if `Vector2i::length` would be returning `int` instead of `float` etc. :upside_down_face: Maybe we could/should add something like `Rect2i::get_centerf`? :thinking: (assuming `Rect2i::get_center` can't be changed because of compat)

|debug drawing| Before<br>(v4.3.dev1.official [9d1cbab1c])| After<br>(this PR) |
|-|--------|--------|
|Physics<br>- rendering offsets fixed|Godot_v4 3-dev1_win64_m0yNeSm9AH|AUAudD4WTr|
|Navigation<br>- rendering offsets fixed<br>- still incorrect polygon generated|TJOynU8MBc|WapiwxvbWN|

Haven't tested this thoroughly so not sure if maybe something else got broken/revealed incorrect offset calculation or something like that.

### DIFF
--- a/editor/plugins/tiles/tile_set_atlas_source_editor.cpp
+++ b/editor/plugins/tiles/tile_set_atlas_source_editor.cpp
@@ -1869,7 +1869,7 @@ void TileSetAtlasSourceEditor::_tile_atlas_control_unscaled_draw() {
 		for (int i = 0; i < tile_set_atlas_source->get_tiles_count(); i++) {
 			Vector2i coords = tile_set_atlas_source->get_tile_id(i);
 			Rect2i texture_region = tile_set_atlas_source->get_tile_texture_region(coords);
-			Vector2i position = texture_region.get_center() + tile_set_atlas_source->get_tile_data(coords, 0)->get_texture_origin();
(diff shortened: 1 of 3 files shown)


JekSun97:
I hope the Godot team will no longer rewrite the entire engine :D

groud (approved):

akien-mga:
Thanks!

PR #87161, opened 2024-01-14
Do not commit transforms and handles if no changes were made

ryevdokimov:
Currently left clicking without moving the mouse will commit a transform and handles unnecessarily flooding the output and history with transform messages. This PR checks that the mouse was moved, meaning a changed was made, before committing. 

image

image

image

### DIFF
--- a/editor/plugins/abstract_polygon_2d_editor.cpp
+++ b/editor/plugins/abstract_polygon_2d_editor.cpp
@@ -291,6 +291,7 @@ bool AbstractPolygon2DEditor::forward_gui_input(const Ref<InputEvent> &p_event) 
 
 					const PosVertex closest = closest_point(gpoint);
 					if (closest.valid()) {
+						original_mouse_pos = gpoint;
 						pre_move_edit = _get_polygon(closest.polygon);
 						edited_point = PosVertex(closest, xform.affine_inverse().xform(closest.pos));
 						selected_point = closest;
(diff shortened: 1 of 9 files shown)


fire:
@KoBeWi

Thoughts on avoiding the commit of transformations when no actual change to the transform values has occurred? This could improve the editor's performance by avoiding redundant actions, but I am not familiar with this code.

KoBeWi:
Makes sense.
Seems like the same problem exists in 2D (e.g. in CollisionShape2D), not sure if it's as easy to fix.

KoBeWi:
CollisionShape2D was an example, there are more 2D editors - Path2D, Polygon2D, Line2D, Raycast2D and possibly others. Hence it might not be as easy, because the code is not unified like in 3D.

ryevdokimov:
Yeah, I know it was a premature push, my bad. I'm working on fixing all of them.

ryevdokimov:
> CollisionShape2D was an example, there are more 2D editors - Path2D, Polygon2D, Line2D, Raycast2D and possibly others. Hence it might not be as easy, because the code is not unified like in 3D.

A lot of these end up inheriting the same class, so it wasn't too bad. I went through all the 2D nodes to see if I missed anything, but I am admittedly not as familiar with 2D as I am with 3D, so let me know if I missed any.

KoBeWi (approved):

akien-mga:
Thanks!

PR #87162, opened 2024-01-14
Modified undo_redo.h

### LINKED ISSUE #87101: Using undo shortcut doesn't work on Script Editor.
### Tested versions

Reproducible in : 352434668923978f54f2236f20116fc96ebc9173
not reproducible in : 4.2

### System information

### Issue description

When pressing the shortcut it shows nothing to Undo even though there is clearly something as shown when doing undo through menu. 

https://github.com/godotengine/godot/assets/66184050/73675e7a-d2a5-4582-be80-94498016175a

### Steps to reproduce

- Open Script Editor
- Write something
- Press Ctrl + Z or whatever your undo shortcut
- it prints Nothing to undo
- go in menu Edit > Undo and it will properly Unod

N/A

blitzboah:
Fixes #87101

### DIFF
--- a/core/object/undo_redo.h
+++ b/core/object/undo_redo.h
@@ -103,7 +103,7 @@ protected:
 	static void _bind_methods();
 
 public:
-	void create_action(const String &p_name = "", MergeMode p_mode = MERGE_DISABLE, bool p_backward_undo_ops = false);
+	void create_action(const String &p_name = "", MergeMode p_mode = MERGE_ALL, bool p_backward_undo_ops = false);
 
 	void add_do_method(const Callable &p_callable);
 	void add_undo_method(const Callable &p_callable);


AThousandShips (changes requested):
I don't believe this is the proper solution, this will affect every use of this, with various possible risks, instead the place that uses it should be fixed instead, [local solutions to local problems](https://docs.godotengine.org/en/stable/contributing/development/best_practices_for_engine_contributors.html#prefer-local-solutions)

Further the script editor uses the editor undo system and it handles the checks afaik

victorbstan:
Undo is available in editor if triggering manually from the editor menu, I think the issue is about which undo queue is called when using the key shortcut, IE: if the script editor is focused it should undo from the script editor queue, if the 3d/2d editor is focused, it should undo from the editor queue.

ajreckof:
as stated in the bug report this bug is a regression first step to solve it should be to find which PR introduced the bug too find out where to start from. I agree with AThousandShips that this solution is very likely to have a lot of unwanted side effects.

akien-mga:
Closing based on the discussion above.

PR #87164, opened 2024-01-14
Rename camera `near` and `far` private members to avoid conflict with `Windows.h` defines

0x0ACB:
Sadly `Windows.h` defines macros for both `near` and `far` so naming variables like this, especially in header files is causing a lot of issues. As soon as any file includes `Windows.h` before `camera_3d.h` your build will simply fail. While `camera_3d.h` does already include undefs for mingw builds this is insufficient as those macros are used in other parts of the windows headers as well.

This PR simply renames `Camera3D::far` and `Camera3D::near` to `Camera3D::camera_far` and `Camera3D::camera_near` to avoid this issue alltogether. Since all public methods stay the same this shouldn't cause any downstream issues.

### DIFF
--- a/scene/3d/camera_3d.h
+++ b/scene/3d/camera_3d.h
@@ -36,11 +36,6 @@
 #include "scene/resources/camera_attributes.h"
 #include "scene/resources/environment.h"
 
-#ifdef MINGW_ENABLED
-#undef near
-#undef far
-#endif
-
 class Camera3D : public Node3D {
 	GDCLASS(Camera3D, Node3D);
 
@@ -72,8 +67,9 @@ private:
 	real_t fov = 75.0;
 	real_t size = 1.0;
 	Vector2 frustum_offset;
-	real_t near = 0.05;
-	real_t far = 4000.0;
+	// _ prefix to avoid conflict with Windows defines.
+	real_t _near = 0.05;
+	real_t _far = 4000.0;
 	real_t v_offset = 0.0;
 	real_t h_offset = 0.0;
 	KeepAspect keep_aspect = KEEP_HEIGHT;
--- a/scene/3d/camera_3d.cpp
+++ b/scene/3d/camera_3d.cpp
@@ -45,14 +45,14 @@ void Camera3D::_update_camera_mode() {
 	force_change = true;
 	switch (mode) {
 		case PROJECTION_PERSPECTIVE: {
-			set_perspective(fov, near, far);
+			set_perspective(fov, _near, _far);
 
 		} break;
 		case PROJECTION_ORTHOGONAL: {
-			set_orthogonal(size, near, far);
+			set_orthogonal(size, _near, _far);
 		} break;
 		case PROJECTION_FRUSTUM: {
-			set_frustum(size, frustum_offset, near, far);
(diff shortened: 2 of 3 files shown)


AThousandShips:
These could alternatively be undefined with other things in `typedefs.h` where a lot of things windows badly defines are undefined, like `min/max`

0x0ACB:
That won't work. While you can remove the min/max defs from Windows.h with `WIN32_LEAN_AND_MEAN` or `NOMINMAX` undefining `far` and `near` has a lot of issues as these macros are used in a lot of additional headers that then won't compile at all. Renaming them is the only solution to this issue that I got to work. A lot of function definitions in windows headers are using the near and far macros.

0x0ACB:
Don't think these can simply be removed in other locations. Atleast the couple where I found this macro there were local variables called near and far . So it would require a lot more refactoring to get rid of all of those

scene/3d/camera_3d.h
```diff
@@ -72,8 +67,8 @@ class Camera3D : public Node3D {
 	real_t fov = 75.0;
 	real_t size = 1.0;
 	Vector2 frustum_offset;
-	real_t near = 0.05;
-	real_t far = 4000.0;
+	real_t camera_near = 0.05;
+	real_t camera_far = 4000.0;
```
akien-mga, line 71: I'd suggest something like this since adding "camera_" as prefix doesn't really add meaningful information. And we have some classes where private members use `_` as a prefix, though it's not a convention we use explicitly.
```suggestion
	// _ prefix to avoid conflict with Windows defines.
	real_t _near = 0.05;
	real_t _far = 4000.0;
```

Alternatively, `near_distance`/`far_distance` if we want to be explicit at the cost of verbosity.

akien-mga (approved):
Makes sense to me. Those are private members so it doesn't break compat for C++ modules.

akien-mga:
Could you amend the commit message to be more explicit, e.g. like I renamed the PR? This would prevent confusion about this possibly renaming the actual exposed `near` and `far` properties. And we like commit messages starting with a capital letter ;)

> Don't think these can simply be removed in other locations. Atleast the couple where I found this macro there were local variables called near and far . So it would require a lot more refactoring to get rid of all of those

Would be nice to do as a follow-up PR if you're interested.

scene/3d/camera_3d.h
0x0ACB: I added the _ prefix

fire (changes requested):
@akien-mga This might require compatibility methods so we don't change the existing api for Camera3D. I am not very familiar with compatibility methods to guide.

AThousandShips:
These are private member variables in the c++ class, not exposed properties

fire (approved):
If they're private apis, that's fine.

akien-mga:
Thanks!

Jimmio92:
This is ridiculous.

```
#include <windows.h>
#undef min
#undef max
```

they're completely unused for most Win32 stuff and are macros. undef them.

0x0ACB:
You alright mate?

This is not even about `min` and `max` macros, which by the way you can turn of via `#define NOMINMAX`.

PR #87165, opened 2024-01-14
Fix Windows editor reload when a GDExtension lacks `FreeLibrary` support

31:
* For https://github.com/godotengine/godot-proposals/issues/8772
* Implements the idea I mentioned in [a comment](https://github.com/godotengine/godot-proposals/issues/8772#issuecomment-1890778324), but with a bit more polish.

As far as I can tell, this functionality is only useful on Windows, and only to address that issue/proposal. For now though, I didn't write the PR to go out of its way to make it not work on other OSes or in other situations.

About this:

> 	// Kill the previous editor process and wait for it to stop. The OS::kill
>	// method sends the kill signal without waiting, so it can't be used here.

I could change `kill` on Windows to wait for the process to terminate, but for compatibility reasons and the potential for a perf hit when it isn't necessary to wait, I decided not to do that in this PR for now.

I've only tested this on my local Windows 10 machine with a trivial GDExtension written in Go (cgo).

### DIFF
--- a/core/os/os.cpp
+++ b/core/os/os.cpp
@@ -536,6 +536,16 @@ List<String> OS::get_restart_on_exit_arguments() const {
 	return restart_commandline;
 }
 
+Error OS::kill_previous_editor_and_wait(const ProcessID &p_pid) {
+	// On most platforms, kill waits for the process to exit. On Windows, it
+	// doesn't, and needs to be overridden for this case specifically.
+	return kill(p_pid);
+}
+
+bool OS::is_restart_responsible_for_exit() const {
+	return false;
+}
+
 PackedStringArray OS::get_connected_midi_inputs() {
 	if (MIDIDriver::get_singleton()) {
 		return MIDIDriver::get_singleton()->get_connected_inputs();
--- a/core/os/os.h
+++ b/core/os/os.h
@@ -95,6 +95,8 @@ public:
 		RENDER_SEPARATE_THREAD
 	};
 
+	typedef int64_t ProcessID;
+
 protected:
 	friend class Main;
 	// Needed by tests to setup command-line args.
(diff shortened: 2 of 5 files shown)


31:
I wasn't happy with my original commit summary and reworded it to focus on what it fixes rather than how. I was hesitant before to say "fix" because I don't think this is something Godot was intending to support before, but it seems like the best way to phrase it concisely.

dsnopek:
Just copying [my comment](https://github.com/godotengine/godot-proposals/issues/8772#issuecomment-1894061640) from the proposal:

> I don't like the idea of killing the process. It really doesn't solve the problem you describe, that is: a new unrelated process can still start with the same PID, but now we're killing it, which seems like a worse outcome.
> 
> What if it waited for the process, but had a reasonable timeout, where it would give up waiting if it took longer than, say, 5 seconds?

31:
> a new unrelated process can still start with the same PID, but now we're killing it, which seems like a worse outcome.

Can you describe how this would happen in more detail? The mechanism in this PR is specifically designed to avoid that. The old process waits for 10 seconds for the new process to show up and kill it. Since the old process doesn't terminate, its PID is still being used and can't be used by a new process.

I think increasing that timeout from 10s to infinite would also be reasonable, on second thought, to avoid a timing issue on any machine no matter how slow. This changes the error behavior a bit, but any timeout is a compromise.

31:
After a conversation at https://chat.godotengine.org/channel/gdextension/thread/BoGSKEcYPDqALL8bq, the main concern is about using `TerminateProcess` at all: if there are external consequences of using that function (will a crash handler care?), and if this might cause some important Godot editor cleanup code not to run. I'm not enough of a Windows and Godot internals expert to really be sure.

I'm still interested in going after a non-racing solution rather than a workaround, so if someone is able to confirm how this will work, I'm still interested in finishing this PR.

If we could signal to the old process that it should quit in some other way than `TerminateProcess`, that might also help.

PR #87167, opened 2024-01-14
Make CharacterBody properties public

0x0ACB:
For some reason the get and set methods for most of the properties in `CharacterBody3D` are marked as private in C++ even though they are all exposed to GDScript. This PR simply marks those methods public in C++ as well so its possible to call them.

### DIFF
--- a/scene/2d/physics_body_2d.h
+++ b/scene/2d/physics_body_2d.h
@@ -362,44 +362,6 @@ public:
 	int get_slide_collision_count() const;
 	PhysicsServer2D::MotionResult get_slide_collision(int p_bounce) const;
 
-	CharacterBody2D();
-	~CharacterBody2D();
-
-private:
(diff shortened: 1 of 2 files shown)


AThousandShips:
Same should be done to the 2D case if it's done here

0x0ACB:
Well I would say access should be consistent in general. I only noticed these methods were marked private in `CharacterBody3D` because I was trying to use them in C++ and couldn't. No idea where else in the codebase this is an issue.

AThousandShips:
In this case the 2D and 3D should match as much as possible, general consistency is also good but there's no reason to just do it for one largely identical class

0x0ACB:
Updated CharacterBody2D

KoBeWi (approved):

akien-mga:
Thanks!

Issue #75998, opened 2023-04-12
Run current scene (F6) often launches unexpected scene


h0lley:
### Godot version

v4.0.2.stable.official [7a0977ce2]

### System information

Ubuntu

### Issue description

I would expect F6 to always launch the scene that is currently selected from the scene tab bar. This also seems to have been the case up until recently.

However, currently it seems that if test play is already running, and F6 is used to not just launch a new test play process but also terminate the previous one, then it always loads the same scene the previous test play session has been started with - regardless of the scene tab that is actually selected in the editor.

So a workaround appears to be to always manually stop test play (F8) before starting it again, but this messes a lot with the UX and workflow of rapid prototyping.

### Steps to reproduce

1. Start any project with at least two scenes
2. Launch into scene A via F6, keep it running in the background
3. Then in the editor, open scene B
4. Press F6 again to launch into scene B
5. Note that test play unexpectedly starts with scene A again

### Minimal reproduction project

should be trivial to reproduce in any project, let me know if not

Calinou:
I remember seeing this reported in another issue, but I can't find it right now.

h0lley:
> This may actually be intended behavior though – pressing F6 without stopping the project first will restart the project, not run the current scene.

that is not the behavior of Godot 3, and I am pretty sure it changed in 4 only a couple of weeks or perhaps roughly two months ago at most (that is roughly when I started noticing and being annoyed by it, but only today it triggered me sufficiently to make me look into reproduction 😅).

plus I don't see how that would make any intuitive sense nor is it being communicated in the tooltip. F6 should have only one function and not a hidden second one. restarting the project can still be done by pressing F5 or F6 depending on with which scene you want to start with the original behavior.

AThousandShips:
Nope this changed some 8 months ago

h0lley:
ah, interesting.
I actually never realized that when starting test play, the `Run current scene (F6): Play the edited scene` button turns into a `Run current scene (F6): Reload the played scene` button.

I think this is confusing and don't see the usability in it.
the behavior of Godot 3 makes is what I'd expect intuitively and has greater usability.

|Behavior|Action|Use case|
|Godot 3|Pressing F5||
|Godot 3|Pressing F5 when the game already runs||
|Godot 4|Pressing F6||

h0lley:
ah, interesting.
[...]

PR #58665, opened 2022-03-01
Improve scene playing and reloading

KoBeWi:
This PR improves the scene reloading behavior in the editor. When you play a scene, the play button you pressed turns into reload arrow. In case of play custom, it reloaded again the scene you picked, but for play current it always played the currently opened the scene, despite the button also changes. I also added a tooltip to better indicate this behavior and refactored code a bit to decrease duplication.
image

Fixes #31515

PR #87170, opened 2024-01-14
Fix reloading current scene forgetting path

### LINKED ISSUE #87163: "Reload the played scene" (F6) is launching the currently focused tab, not reloading the previously selected scene
### Tested versions

Reproducible in 4.2.1 Stable
Not Reproducible in 4.0.4

### System information

Godot v4.2.1.stable - Windows 10

### Issue description

I recently upgraded from 4.0.4 to 4.2.1 and suddenly the reload button is launching the currently focused scene tab instead of the scene previously selected to run.

This behavior is different than before when it would reload the previously selected scene.

It is important for this button to work correctly so I can deliberately test a (not main) scene while using the editor tabs to update and fix other scenes that are part of it.
For example, the main scene of my project is a "Menu" scene but I am developing an "Arena" scene with a "Character" and a new "Obstacle".
I select the Arena and press the "Run Current Scene (F6)" button so I can test the interaction between the Character and the Obstacle
I then use the editor tabs to inspect and make changes to the child scenes.
Previously I would then press the "Reload the Played Scene" button to re-launch the Arena and test my updates, but instead it will launch the currently focused tab, be it the Character or Obstacle, which is not helpful for testing interaction.

Is there a setting somewhere that I've overlooked? Thanks

### Steps to reproduce

Pick a scene, launch it.
Switch a tab.
Try to relaunch the scene you picked.
Notice that it launches the currently focused tab instead.

-

AThousandShips:
Scene was stopped before running which cleared the last run scene, forcing it to run the currently edited scene instead

* Fixes: https://github.com/godotengine/godot/issues/87163

### DIFF
--- a/editor/gui/editor_run_bar.cpp
+++ b/editor/gui/editor_run_bar.cpp
@@ -273,12 +273,13 @@ void EditorRunBar::play_main_scene(bool p_from_native) {
 }
 
 void EditorRunBar::play_current_scene(bool p_reload) {
+	String last_current_scene = run_current_filename; // This is necessary to have a copy of the string.
+
 	EditorNode::get_singleton()->save_default_environment();
 	stop_playing();
 
 	current_mode = RunMode::RUN_CURRENT;
 	if (p_reload) {
-		String last_current_scene = run_current_filename; // This is necessary to have a copy of the string.
 		_run_scene(last_current_scene);
 	} else {
 		_run_scene();


KoBeWi:
The Play Current Scene button changes icon and tooltip when the project is running.
image
The current behavior is intended (see #58665)

AThousandShips:
So the current behavior is against the tooltip?

editor/gui/editor_run_bar.cpp
```diff
@@ -273,12 +273,13 @@ void EditorRunBar::play_main_scene(bool p_from_native) {
 }
 
 void EditorRunBar::play_current_scene(bool p_reload) {
+	String last_current_scene = run_current_filename; // This is necessary to have a copy of the string.
+
 	EditorNode::get_singleton()->save_default_environment();
 	stop_playing();
```
AThousandShips, line 279: This clears the `run_current_filename` value, see below

AThousandShips:
So to confirm, the intended behavior is to run the scene open in the editor tab, not the scene that is running as the button is played?

KoBeWi:
Ok I misunderstood 🙃
Sounds like it's indeed a bug and it should reload the running scene. Though this behavior was discussed after my change and it possibly has changed since then.

AThousandShips:
It seems to be an oversight, as the line saves the variable, even though it would always be empty at that point

KoBeWi (approved):

akien-mga:
I think there's been confusion around what should be the behavior, and if the "expected" behavior is not currently working, I'm worried it might have become the de facto expectation from users.

When this was changed (apparently it was working initially), users complained about the change of behavior compared to Godot 3:
- https://github.com/godotengine/godot/issues/75998
- https://github.com/godotengine/godot-proposals/issues/6550

So I'm worried that this bugfix might be unpopular.

KoBeWi:
We could either move the reload behavior to another button (e.g. Run Project) or add a new button for reloading scenes.

AThousandShips:
At least the tooltip would need changing, the description is "Reload the played scene" which cannot be interpreted in any other way IMO, how is loading a new scene "reloading"?

This is a regression from relocating the run bar, it worked before, and the code in place is for this behavior, otherwise that code makes no sense

So this restores intended behavior, and fixes an oversight in previous PRs, at least if the current behavior is kept the tooltip must be updated

editor/gui/editor_run_bar.cpp
AThousandShips: The code in `stop_playing` clearing `run_current_filename` was introducted in #76026, and wasn't in the original code

KoBeWi:
> at least if the current behavior is kept the tooltip must be updated

And the icon too, because the button changes into reload.

akien-mga:
> So this restores intended behavior, and fixes an oversight in previous PRs, at least if the current behavior is kept the tooltip must be updated

My point is, the intended behavior was opposed by some users shortly after it was implemented, and then eventually an unintended regression "fixed" the change to restore the 3.x behavior by error.

So this to me is a strong indication that we need to properly assess what users want, and not what was approve back in 2022 and led to this tooltip saying this :)

AThousandShips:
Well I'd say to fix this now and discuss it, or change the tooltip and icon now, and discuss it, but currently it causes confusion (evidently from the report)

AThousandShips:
Further I'd argue that reverting that improvement by effectively accepting this bug (which several code changes and design changes) worsens workflow, possible senario:
* Using a main scene, like a testing ground scene, to test various components, such as a player, an enemy, etc.
* Run the main scene, then do some testing, go to the player scene, do some editing, reload running scene

If you now want the 3.x behavior you just press the "stop" button, and then the "run current scene" button, trivial

However if we revert that improvement now you have to move back and forth between scenes in the editor to re-run the scene you're using as a base, this is *much* more disruptive than having to press two buttons instead of one IMO

I'd say this can be discussed further, but I don't think leaving a regression in place even with fixes to fix the confusing things, is a good workflow, and at least this needs to be fixed in one way or another, because the current state is confusing

YuriSizov (approved):
Seems like I made a mistake, sorry. I agree that the assumed workflow makes a lot of sense in practice. You usually have complex scenes with instantiated packed scenes in them, and may want to look into one of those while still debugging the same starting scene. It's a nuisance to do this if the button launches the currently edited scene.

YuriSizov:
If we worry about breaking expectations of people who are currently used to the status quo, let's avoid cherry-picking this fix and introduce it in 4.3. We can also make it an option, if the opinions are so evenly split we can't decide on the most optimal solution.

AThousandShips:
I agree on the cherry pick, let's discuss it for the backwards cases and optionality etc.

akien-mga:
Thanks!

AThousandShips:
Thank you!

PR #87173, opened 2024-01-14
Add Sphere Mask and Particles Turbulence Nodes and Posterize ColorFunc

### LINKED ISSUE #5106: [TRACKER] Visual Shader nodes wishlist

Various VFX

Some nontrivial VS nodes are missing.

Adding a couple of nodes to reach feature parity with what spatial shaders and canvas shaders can offer out-of-the-box.

Spatial:
- RotationMatrix node (supply axis and angle, return transform and/or transformed vertex)
- Proximity fade
- Camera fade
- Rand from seed (copy-pasted from particles)

It can, but it's about improving VS UX.

It can, but it's about improving VS UX.

paddy-exe:
Adds Posterize ColorFunc Mode to CanvasItem and Spatial Visual Shader Nodes 
Adds Particle Turbulence to Particle Visual Shader Nodes

Closes https://github.com/godotengine/godot-proposals/issues/5106
Part of https://github.com/godotengine/godot-proposals/issues/7666

### Posterize node
grafik

### ParticleTurbulence

https://github.com/godotengine/godot/assets/38077837/23ea529b-ff6b-4d2c-90b9-c6edbd64f858

_Production edit: closes godotengine/godot-roadmap#7_


scene/resources/visual_shader_particle_nodes.cpp
```diff
@@ -1162,6 +1162,205 @@ VisualShaderNodeParticleAccelerator::VisualShaderNodeParticleAccelerator() {
...
+		case 8:
+			return "influence over life";
+	}
+	return String();
+}
+
+String VisualShaderNodeParticleTurbulence::generate_global_per_node(Shader::Mode p_mode, int p_id) const {
+	String code;
+
+	code += "vec4 turb_grad(vec4 p) {\n";
```
AThousandShips, line 1254: I'd suggest doing this with R strings

paddy-exe: Could you give me an example or point me to a file where it is implemented with R strings? Haven't used R strings yet

scene/resources/visual_shader_nodes.cpp
```diff
@@ -3179,15 +3179,46 @@ String VisualShaderNodeColorFunc::get_caption() const {
...
+					return "";
+				case 1:
+					return "Steps";
+				default:
+					break;
+			}
+			break;
+		default:
+			return "";
+	}
```
AThousandShips, line 3221:
```suggestion
					break;
			}
			break;
		default:
			break;
	}
	return "";
```

scene/resources/visual_shader_nodes.cpp
```diff
@@ -3179,15 +3179,46 @@ String VisualShaderNodeColorFunc::get_caption() const {
...
+					return PORT_TYPE_VECTOR_3D;
+				case 1:
+					return PORT_TYPE_SCALAR_INT;
+				default:
+					break;
+			}
+			break;
+		default:
+			return PORT_TYPE_VECTOR_3D;
+	}
```
AThousandShips, line 3204:
```suggestion
			break;
	}
	return PORT_TYPE_VECTOR_3D;
```

scene/resources/visual_shader_particle_nodes.cpp
AThousandShips: See `sky_material` for an example 🙂

paddy-exe: Wait, you mean for `generate_global_per_node`? Why there? If anywhere I would assume `generate_code`

AThousandShips: Because it's a long block of strings being added and it's much cleaner if it's in one block of string instead

paddy-exe: Should be done now :D

doc/classes/VisualShaderNodeParticleTurbulence.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="VisualShaderNodeParticleTurbulence" inherits="VisualShaderNode" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
+	<brief_description>
+		Visual shader node for randomized particle movement.
```
Geometror, line 4: For consistency :)
```suggestion
		A visual shader node for randomized particle movement.
```

scene/resources/visual_shader_particle_nodes.cpp
```diff
@@ -1162,6 +1162,206 @@ VisualShaderNodeParticleAccelerator::VisualShaderNodeParticleAccelerator() {
...
+}
+
+float get_turb_influence(uint __noise_seed, float __turb_influence_min, float __turb_influence_max, float __turb_influence_over_life) {
+	float __turb_influence_range = mix(__turb_influence_min, __turb_influence_max, __rand_from_seed(__noise_seed));
+	float __turb_influence = clamp(__turb_influence_range * __turb_influence_over_life, 0.0, 1.0);
+	return __turb_influence;
+}
+	)");
+
+	return code;
```
Geometror, line 1340: Do you think it would be feasible/make sense to share this with `ParticleProcessMaterial` to avoid duplication?

doc/classes/VisualShaderNodeParticleTurbulence.xml
```diff
@@ -0,0 +1,11 @@
...
+	</brief_description>
+	<description>
+		The Turbulence Node will output a noise direction and turbulence influence values based on the given noise parameters which can be used to influence the give particles random velocity.
```
Geometror, line 7:
```suggestion
		The Turbulence Node generates a noise direction and turbulence influence value based on the given noise parameters.
		These can be used to add a turbulent motion to the particle's velocity, which allows for creating organic and natural-looking particle effects.
```

doc/classes/VisualShaderNodeSphereMask.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="VisualShaderNodeSphereMask" inherits="VisualShaderNode" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
+	<brief_description>
+		A visual shader node that smoothly masks via multiplication.
+	</brief_description>
+	<description>
+		The SphereMask Node offers circular masking options with optional harshness and center specifications. It can be used to blend out the outer edges of a sprite.
```
Geometror, line 7: Did you mean "hardness"? :)

Geometror (commented):
Posterize needs its own AddOption (`visual_shader_editor_plugin.cpp` ~line 6210) so that you can add it directly like the other color functions.

I played a bit with every node and it looks like everything is working fine.

Ideally this PR would be split up since (I think) these nodes are unrelated to each other, but nonetheless great work!

scene/resources/visual_shader_particle_nodes.cpp
Geometror: Allowed myself to mark this as resolved :)

scene/resources/visual_shader_particle_nodes.cpp
paddy-exe: I don't think it makes sense here to combine these because I had to change quite a few parts of the shader code to make it usable for visual shaders due to the separation of the `start` and `process` functions in the particle shader. Furthermore, the `ParticleProcessMaterial` can't be converted to a visual shader anyway so I currently see no reason to do so.

paddy-exe:
@Geometror Thanks for the detailed feedback! I have amended the changes you suggested

Ansraer (changes requested):
The sphere mask node sounds like a useful node to have (I know that I created something similar it manually multiple times in the past) but I really think the name should be changed.
"Mask" kind of implies for me that this can only be used to mask an existing texture, which is only one usecase of it. In addition the node itself doesn't mask anything, needing an additional Multiplication OP node for that purpose.

IMO "SphereShape" would be a far better name that better communicates what this node actually does.

scene/resources/visual_shader_particle_nodes.cpp
```diff
@@ -1162,6 +1162,206 @@ VisualShaderNodeParticleAccelerator::VisualShaderNodeParticleAccelerator() {
...
+}
+
+VisualShaderNodeParticleTurbulence::VisualShaderNodeParticleTurbulence() {
+	set_input_port_default_value(2, 1.0); // noise strength
+	set_input_port_default_value(3, 9.0); // noise scale
+	set_input_port_default_value(4, Vector3(0.0, 0.0, 0.0)); // noise speed
+	set_input_port_default_value(5, 0.2); // noise speed random
+	set_input_port_default_value(6, 0.1); // influence min
+	set_input_port_default_value(7, 0.1); // influence max
+	set_input_port_default_value(8, 1.0); // influence over life
```
Geometror, line 1362: I know this is the way we currently do it for almost all nodes, but we should really start to use enums to describe the ports/slots. (not a blocker, just for reference)

Geometror (commented):
@Ansraer I have to disagree here, in my opinion "mask" is a better term to describe what it does (and is likely the most common usecase, not just to mask textures, but whole parts of shaders). However, I'm fine with "shape" too.
But I think it should be called CircleMask since it's 2D.

Geometror:
Besides that, I think it would be of value to implement it as a general Shape/Mask node (which allows selecting a shape using a dropdown menu). Example from material maker:
grafik
Well, I just noticed it's called "Shape" there, one point for Ansraer :)

paddy-exe:
Updated the PR and removed the SphereMask node for now. I have the changes locally still.

PR #87114, opened 2024-01-12
Overhaul AABB's documentation

Mickeon:
Very much related to https://github.com/godotengine/godot/pull/69816

This PR almost completely overhauls **AABB**'s class reference documentation, and makes it match the current state of **Rect2** and **Rect2i**.

Some of the documentation is ported 1:1 from **Rect2**'s, but there are notable differences from Rect2:
- Say "bounding box" instead of "rectangle";
    - The current documentation likes to say "AABB" a lot but its quite confusing to imagine.
- Reduce the scale of some examples;
    - With **AABB** we're usually talking about meters, which can be a lot larger than pixels. And besides, using 2 digits for every value, out of 6...? It's a lot to follow.
- Avoid mentions of "_bottom-left-back_", "_top-right-forward_" corners.
     - With 3 dimensions envisioning and writing "_bottom-left and back_" corner is quite difficult to understand.
- Change parameter names around:
    - For some reason, **Rect2** and **AABB** have different parameter names for some methods (e.g. "_b_" becomes "_with_"). Descriptions had to be adjusted accordingly, and they may or may not sound more awkward as a result.
     
Not everything was copied word-for-word, however.
- Clarify `intersect_ray()` and `intersect_segment()`, `get_endpoints()`, etc.;
- Clarify `has_volume`:
     - _"Returns [code]true[/code] if the [AABB] has a volume"_ thanks, I guess...
- Aadd an example for all `*_axis_*` methods;
- Fix misinformation in `has_point`:
    - "_Points on the faces of the AABB are considered included, though float-point precision errors may impact the accuracy of such checks._". Oh the "inaccuracy" excuse.
    https://github.com/godotengine/godot/blob/26b1fd0d842fa3c2f090ead47e8ea7cd2d6515e1/core/math/aabb.h#L320-L328
    If that statement were true, these would be `>=`, not `>`.

    

------
Feedback is very, very welcome. Thank you for the support in advance.

PR #84756, opened 2023-11-11
Doc: Mention C++ uses `Basis::get_column`

Jackerty:
I have face-planted now a couple of times to this.

For GDScript users `Basis` has an `x`, `y`, and  `z` for accessing the columns of the basis matrix. These can also be accessed in GDScript by array index.

When using C++ GDExtension, `Basis` does not have `x`, `y`, or `z`, so I have used array indexing twice to access the columns. Except, this does not work! Array index in C++ gives the rows of the basis matrix!

To access columns undocumented `get_column`/`set_column` methods have to be used.

As such, I would like to suggest that documentation would have a note for C++ developers. This is one of the rare instances where a C++ developer has to have arcane knowledge to follow a GDScrip tutorial.

PR #87175, opened 2024-01-14
Overhaul Basis' documentation

Mickeon:
Continuation of https://github.com/godotengine/godot/pull/86664

_Bugsquad edited:_
Related to https://github.com/godotengine/godot/pull/87114, and ~[several other past efforts](https://github.com/godotengine/godot/pulls?q=is%3Apr+author%3A%40me+overhaul+)~ [several other past efforts](https://github.com/godotengine/godot/pulls?q=is%3Apr+author%3AMickeon+overhaul+).

This PR completely overhauls Basis' documentation to the point where it is almost unrecognisable

### Why?

The problems with the current documentation are summarised fairly easily:
- It is written under the assumption that the reader has a PhD in linear algebra, not for a developer.
- It explains the inner workings of Basis too much.
- It is just far, far too verbose sometimes, with no actual substance.
    - "_Assuming that the matrix is a proper rotation matrix, slerp performs a spherical-linear interpolation with another rotation matrix._"
    - "This operator multiplies all components of the Basis, which scales it uniformly."
    - "_This can be used to validate if the basis is non-distorted, which is important for physics and other use cases._"
    - "_Assuming that the matrix is the combination of a rotation and scaling, return the absolute value of scaling factors along each axis._" <sub>Guys, it's the [length](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-method-length) of every axis. That's it!</sub>

### How does this PR fix it?

On the other hand, this PR does so **many** changes that, explaining the reasoning behind every single one of them is pretty difficult.
As more notes may be edited latter, let's get these out of the way:
- Avoid advanced math terms at all costs:
   - "_Orthonormal_", "_orthogonal_", "_transformation_". Avoid mentioning them unless it's very appropriate.
   - Exception to terms like "_normalized_" can stay. They have a clear and common purpose in Godot.
- Say "_basis_" consistently, avoid "_basis matrix_" or "_matrix_" alone.
   - Most users are reading this class reference for the **Basis** aspect of the... **Basis**. Mentioning its matrix a lot can potentially be confusing, unless strictly necessary for a given method. 
   Advanced users know fully well it is a 3x3 matrix and will do as they please.
- Elaborate on **oh-so-many** methods!
    - How should I structure these parameters? How is the returned value like? What does the operation I'm doing even mean?
    - In some cases, this is just not feasible, as it would not just be too complicated, but it would bloat the documentation with information most users wouldn't care about.
    This may be frowned upon, but sometimes a link to an external article is a must.

More specific notes:

- Give full details on what Euler angles are:
    - What does each axis represent? What does the default EulerOrder do?
    - At first I was reluctant to link to [Wikipedia article](https://en.wikipedia.org/wiki/Euler_angles), but I was sold by the simple-to-follow visual examples.
- Explain how the UNIFORM basis is structured;
- Add a much-needed example for `slerp()`;
- Expose the formula for `tdotx`, `tdoty`, and `tdotz`;
- Explain the difference between scripting and C++ source code's `[]` operator;
    - Should supersede https://github.com/godotengine/godot/pull/84756.
- ...
--------

Feedback is basically welcome.

### Sources:
- https://datacadamia.com/geometry/skewing
- https://en.wikipedia.org/wiki/Diagonal_matrix
- https://en.wikipedia.org/wiki/Euler_angles

### DIFF
--- a/doc/classes/Basis.xml
+++ b/doc/classes/Basis.xml
@@ -4,10 +4,12 @@
 		A 3×3 matrix for representing 3D rotation and scale.
 	</brief_description>
 	<description>
-		A 3×3 matrix used for representing 3D rotation and scale. Usually used as an orthogonal basis for a [Transform3D].
-		Contains 3 vector fields X, Y and Z as its columns, which are typically interpreted as the local basis vectors of a transformation. For such use, it is composed of a scaling and a rotation matrix, in that order (M = R.S).
-		Basis can also be accessed as an array of 3D vectors. These vectors are usually orthogonal to each other, but are not necessarily normalized (due to scaling).
+		The [Basis] built-in [Variant] type is a 3x3 [url=https://en.wikipedia.org/wiki/Matrix_(mathematics)]matrix[/url] used to represent 3D rotation, scale, and shear. It is frequently used within a [Transform3D].
+		A [Basis] is composed by 3 axis vectors, each representing a column of the matrix: [member x], [member y], and [member z]. The length of each axis ([method Vector3.length]) influences the basis's scale, while the direction of all axes influence the rotation. Usually, these axes are perpendicular to one another. However, when you rotate any axis individually, the basis becomes sheared. Applying a sheared basis to a 3D model will make the model appear distorted.
+		A [Basis] is [b]orthogonal[/b] if its axes are perpendicular to each other. A basis is [b]normalized[/b] if the length of every axis is [code]1[/code]. A basis is [b]uniform[/b] if all axes share the same length (see [method get_scale]). A basis is [b]orthonormal[/b] if it is both orthogonal and normalized, which allows it to only represent rotations. A basis is [b]conformal[/b] if it is both orthogonal and uniform, which ensures it is not distorted.
 		For a general introduction, see the [url=$DOCS_URL/tutorials/math/matrices_and_transforms.html]Matrices and transforms[/url] tutorial.
+		[b]Note:[/b] Godot uses a [url=https://en.wikipedia.org/wiki/Right-hand_rule]right-handed coordinate system[/url], which is a common standard. For directions, the convention for built-in types like [Camera3D] is for -Z to point forward (+X is right, +Y is up, and +Z is back). Other objects may use different direction conventions. For more information, see the [url=$DOCS_URL/tutorials/assets_pipeline/importing_scenes.html#d-asset-direction-conventions]Importing 3D Scenes[/url] tutorial.
+		[b]Note:[/b] The basis matrices are exposed as [url=https://www.mindcontrol.org/~hplus/graphics/matrix-layout.html]column-major[/url] order, which is the same as OpenGL. However, they are stored internally in row-major order, which is the same as DirectX.
 	</description>
 	<tutorials>
 		<link title="Math documentation index">$DOCS_URL/tutorials/math/index.html</link>
@@ -22,7 +24,7 @@
 		<constructor name="Basis">
 			<return type="Basis" />
 			<description>
-				Constructs a default-initialized [Basis] set to [constant IDENTITY].
+				Constructs a [Basis] identical to the [constant IDENTITY].
 			</description>
 		</constructor>
 		<constructor name="Basis">
@@ -37,14 +39,16 @@
 			<param index="0" name="axis" type="Vector3" />
 			<param index="1" name="angle" type="float" />
 			<description>
-				Constructs a pure rotation basis matrix, rotated around the given [param axis] by [param angle] (in radians). The axis must be a normalized vector.
+				Constructs a [Basis] that only represents rotation, rotated around the [param axis] by the given [param angle], in radians. The axis must be a normalized vector.
+				[b]Note:[/b] This is the same as using [method rotated] on the [constant IDENTITY] basis. With more than one angle consider using [method from_euler], instead.
 			</description>
 		</constructor>
 		<constructor name="Basis">
 			<return type="Basis" />
 			<param index="0" name="from" type="Quaternion" />
 			<description>
-				Constructs a pure rotation basis matrix from the given quaternion.
+				Constructs a [Basis] that only represents rotation from the given [Quaternion].
+				[b]Note:[/b] Quaternions [i]only[/i] store rotation, not scale. Because of this, conversions from [Basis] to [Quaternion] cannot always be reversed.
 			</description>
 		</constructor>
 		<constructor name="Basis">
@@ -53,7 +57,7 @@
 			<param index="1" name="y_axis" type="Vector3" />
 			<param index="2" name="z_axis" type="Vector3" />
 			<description>
-				Constructs a basis matrix from 3 axis vectors (matrix columns).
+				Constructs a [Basis] from 3 axis vectors. These are the columns of the basis matrix.
 			</description>
 		</constructor>
 	</constructors>
@@ -61,8 +65,10 @@
 		<method name="determinant" qualifiers="const">
 			<return type="float" />
 			<description>
-				Returns the determinant of the basis matrix. If the basis is uniformly scaled, its determinant is the square of the scale.
-				A negative determinant means the basis has a negative scale. A zero determinant means the basis isn't invertible, and is usually considered invalid.
+				Returns the [url=https://en.wikipedia.org/wiki/Determinant]determinant[/url] of this basis's matrix. For advanced math, this number can be used to determine a few attributes:
+				- If the determinant is exactly [code]0[/code], the basis is not invertible (see [method inverse]).
+				- If the determinant is a negative number, the basis represents a negative scale.
+				[b]Note:[/b] If the basis's scale is the same for every axis, its determinant is always that scale by the power of 2.
 			</description>
 		</method>
 		<method name="from_euler" qualifiers="static">
@@ -70,7 +76,10 @@
 			<param index="0" name="euler" type="Vector3" />
 			<param index="1" name="order" type="int" default="2" />
 			<description>
-				Constructs a pure rotation Basis matrix from Euler angles in the specified Euler rotation order. By default, use YXZ order (most common). See the [enum EulerOrder] enum for possible values.
+				Constructs a new [Basis] that only represents rotation from the given [Vector3] of [url=https://en.wikipedia.org/wiki/Euler_angles]Euler angles[/url], in radians.
+				- The [member Vector3.x] should contain the angle around the [member x] axis (pitch).
+				- The [member Vector3.y] should contain the angle around the [member y] axis (yaw).
+				- The [member Vector3.z] should contain the angle around the [member z] axis (roll).
 				[codeblocks]
 				[gdscript]
 				# Creates a Basis whose z axis points down.
@@ -85,13 +94,14 @@
 				GD.Print(myBasis.Z); // Prints (0, -1, 0).
 				[/csharp]
 				[/codeblocks]
+				The order of each consecutive rotation can be changed with [param order] (see [enum EulerOrder] constants). By default, the YXZ convention is used ([constant EULER_ORDER_YXZ]): the basis rotates first around the Y axis (yaw), then X (pitch), and lastly Z (roll). When using the opposite method [method get_euler], this order is reversed.
 			</description>
 		</method>
 		<method name="from_scale" qualifiers="static">
 			<return type="Basis" />
 			<param index="0" name="scale" type="Vector3" />
 			<description>
-				Constructs a pure scale basis matrix with no rotation or shearing. The scale values are set as the diagonal of the matrix, and the other parts of the matrix are zero.
+				Constructs a new [Basis] that only represents scale, with no rotation or shear, from the given [param scale] vector.
 				[codeblocks]
 				[gdscript]
 				var my_basis = Basis.from_scale(Vector3(2, 4, 8))
@@ -108,26 +118,33 @@
 				GD.Print(myBasis.Z); // Prints (0, 0, 8).
 				[/csharp]
 				[/codeblocks]
+				[b]Note:[/b] In linear algebra, the matrix of this basis is also known as a [url=https://en.wikipedia.org/wiki/Diagonal_matrix]diagonal matrix[/url].
 			</description>
 		</method>
 		<method name="get_euler" qualifiers="const">
 			<return type="Vector3" />
 			<param index="0" name="order" type="int" default="2" />
 			<description>
-				Returns the basis's rotation in the form of Euler angles. The Euler order depends on the [param order] parameter, by default it uses the YXZ convention: when decomposing, first Z, then X, and Y last. The returned vector contains the rotation angles in the format (X angle, Y angle, Z angle).
-				Consider using the [method get_rotation_quaternion] method instead, which returns a [Quaternion] quaternion instead of Euler angles.
+				Returns this basis's rotation as a [Vector3] of [url=https://en.wikipedia.org/wiki/Euler_angles]Euler angles[/url], in radians.
+				- The [member Vector3.x] contains the angle around the [member x] axis (pitch);
+				- The [member Vector3.y] contains the angle around the [member y] axis (yaw);
+				- The [member Vector3.z] contains the angle around the [member z] axis (roll).
+				The order of each consecutive rotation can be changed with [param order] (see [enum EulerOrder] constants). By default, the YXZ convention is used ([constant EULER_ORDER_YXZ]): Z (roll) is calculated first, then X (pitch), and lastly Y (yaw). When using the opposite method [method from_euler], this order is reversed.
+				[b]Note:[/b] Euler angles are much more intuitive but are not suitable for 3D math. Because of this, consider using the [method get_rotation_quaternion] method instead, which returns a [Quaternion].
+				[b]Note:[/b] In the Inspector dock, a basis's rotation is often displayed in Euler angles (in degrees), as is the case with the [member Node3D.rotation] property.
 			</description>
 		</method>
 		<method name="get_rotation_quaternion" qualifiers="const">
 			<return type="Quaternion" />
 			<description>
-				Returns the basis's rotation in the form of a quaternion. See [method get_euler] if you need Euler angles, but keep in mind quaternions should generally be preferred to Euler angles.
+				Returns this basis's rotation as a [Quaternion].
+				[b]Note:[/b] Quatenions are much more suitable for 3D math but are less intuitive. For user interfaces, consider using the [method get_euler] method, which returns Euler angles.
 			</description>
 		</method>
 		<method name="get_scale" qualifiers="const">
 			<return type="Vector3" />
 			<description>
-				Assuming that the matrix is the combination of a rotation and scaling, return the absolute value of scaling factors along each axis.
+				Returns the length of each axis of this basis, as a [Vector3]. If the basis is not sheared, this is the scaling factor. It is not affected by rotation.
 				[codeblocks]
 				[gdscript]
 				var my_basis = Basis(
@@ -154,18 +171,19 @@
 				GD.Print(myBasis.Scale); // Prints (2, 4, 8).
 				[/csharp]
 				[/codeblocks]
+				[b]Note:[/b] If the value returned by [method determinant] is negative, the scale is also negative.
 			</description>
 		</method>
 		<method name="inverse" qualifiers="const">
 			<return type="Basis" />
 			<description>
-				Returns the inverse of the matrix.
+				Returns the [url=https://en.wikipedia.org/wiki/Invertible_matrix]inverse of this basis's matrix[/url].
 			</description>
 		</method>
 		<method name="is_conformal" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns [code]true[/code] if the basis is conformal, meaning it preserves angles and distance ratios, and may only be composed of rotation and uniform scale. Returns [code]false[/code] if the basis has non-uniform scale or shear/skew. This can be used to validate if the basis is non-distorted, which is important for physics and other use cases.
+				Returns [code]true[/code] if this basis is conformal. A conformal basis is both [i]orthogonal[/i] (the axes are perpendicular to each other) and [i]uniform[/i] (the axes share the same length). This method can be especially useful during physics calculations.
 			</description>
 		</method>
 		<method name="is_equal_approx" qualifiers="const">
@@ -187,15 +205,16 @@
 			<param index="1" name="up" type="Vector3" default="Vector3(0, 1, 0)" />
 			<param index="2" name="use_model_front" type="bool" default="false" />
 			<description>
-				Creates a Basis with a rotation such that the forward axis (-Z) points towards the [param target] position.
-				The up axis (+Y) points as close to the [param up] vector as possible while staying perpendicular to the forward axis. The resulting Basis is orthonormalized. The [param target] and [param up] vectors cannot be zero, and cannot be parallel to each other.
-				If [param use_model_front] is [code]true[/code], the +Z axis (asset front) is treated as forward (implies +X is left) and points toward the [param target] position. By default, the -Z axis (camera forward) is treated as forward (implies +X is right).
+				Creates a new [Basis] with a rotation such that the forward axis (-Z) points towards the [param target] position.
+				By default, the -Z axis (camera forward) is treated as forward (implies +X is right). If [param use_model_front] is [code]true[/code], the +Z axis (asset front) is treated as forward (implies +X is left) and points toward the [param target] position.
+				The up axis (+Y) points as close to the [param up] vector as possible while staying perpendicular to the forward axis. The returned basis is orthonormalized (see [method orthonormalized]). The [param target] and [param up] vectors cannot be [constant Vector3.ZERO], and cannot be parallel to each other.
 			</description>
 		</method>
 		<method name="orthonormalized" qualifiers="const">
 			<return type="Basis" />
 			<description>
-				Returns the orthonormalized version of the matrix (useful to call from time to time to avoid rounding error for orthogonal matrices). This performs a Gram-Schmidt orthonormalization on the basis of the matrix.
+				Returns the orthonormalized version of this basis. An orthonormal basis is both [i]orthogonal[/i] (the axes are perpendicular to each other) and [i]normalized[/i] (the axes have a length of [code]1[/code]), which also means it can only represent rotation.
+				It is often useful to call this method to avoid rounding errors on a rotating basis:
 				[codeblocks]
 				[gdscript]
 				# Rotate this Node3D every frame.
@@ -222,7 +241,8 @@
 			<param index="0" name="axis" type="Vector3" />
 			<param index="1" name="angle" type="float" />
 			<description>
-				Introduce an additional rotation around the given axis by [param angle] (in radians). The axis must be a normalized vector.
+				Returns this basis rotated around the given [param axis] by [param angle] (in radians). The [param axis] must be a normalized vector (see [method Vector3.normalized]).
+				Positive values rotate this basis clockwise around the axis, while negative values rotate it counterclockwise.
 				[codeblocks]
 				[gdscript]
 				var my_basis = Basis.IDENTITY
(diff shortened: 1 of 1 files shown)


doc/classes/Basis.xml
```diff
@@ -22,7 +22,7 @@
 		<constructor name="Basis">
 			<return type="Basis" />
 			<description>
-				Constructs a default-initialized [Basis] set to [constant IDENTITY].
+				Constructs a [Basis] identical to [constant IDENTITY].
```
fire, line 25:
```suggestion
				Constructs a [Basis] identical to the [constant IDENTITY].
```

doc/classes/Basis.xml
```diff
@@ -187,15 +201,17 @@
 			<param index="1" name="up" type="Vector3" default="Vector3(0, 1, 0)" />
 			<param index="2" name="use_model_front" type="bool" default="false" />
 			<description>
-				Creates a Basis with a rotation such that the forward axis (-Z) points towards the [param target] position.
-				The up axis (+Y) points as close to the [param up] vector as possible while staying perpendicular to the forward axis. The resulting Basis is orthonormalized. The [param target] and [param up] vectors cannot be zero, and cannot be parallel to each other.
-				If [param use_model_front] is [code]true[/code], the +Z axis (asset front) is treated as forward (implies +X is left) and points toward the [param target] position. By default, the -Z axis (camera forward) is treated as forward (implies +X is right).
+				?? Creates a new [Basis] with a rotation such that the forward axis (-Z) points towards the [param target] position.
+				?? Creates a new [Basis] rotated towards the [param target] position.
```
fire, line 205: `??` Problem again.

doc/classes/Basis.xml
```diff
@@ -340,94 +371,122 @@
...
-				This operator divides all components of the [Basis], which inversely scales it uniformly.
+				Divides all components of the [Basis] by the given [int]. This affects the basis' scale proportionately.
 			</description>
 		</operator>
 		<operator name="operator ==">
 			<return type="bool" />
 			<param index="0" name="right" type="Basis" />
 			<description>
-				Returns [code]true[/code] if the [Basis] matrices are exactly equal.
+				Returns [code]true[/code] if the conponents of both [Basis] matrices are exactly equal.
```
fire, line 480:
```suggestion
				Returns [code]true[/code] if the components of both [Basis] matrices are exactly equal.
```

fire, line 480:
```suggestion
				Returns [code]true[/code] if the components of both [Basis] matrices are exactly equal.
```

doc/classes/Basis.xml
```diff
@@ -4,9 +4,9 @@
...
 	</brief_description>
 	<description>
-		A 3×3 matrix used for representing 3D rotation and scale. Usually used as an orthogonal basis for a [Transform3D].
-		Contains 3 vector fields X, Y and Z as its columns, which are typically interpreted as the local basis vectors of a transformation. For such use, it is composed of a scaling and a rotation matrix, in that order (M = R.S).
-		Basis can also be accessed as an array of 3D vectors. These vectors are usually orthogonal to each other, but are not necessarily normalized (due to scaling).
+		The [Basis] built-in [Variant] type represents a 3x3 [url=https://en.wikipedia.org/wiki/Matrix_(mathematics)]matrix[/url] used for 3D rotation, scale, and shearing. It is frequently used within a [Transform3D].
+		?? A [Basis] is composed by 3 axis vectors: [member x], [member y], and [member z]. The length of each axis ([method Vector3.length]) influences the basis' scale. Usually, these axes are perpendicular to one another. If this is not the case, the basis is sheared and is thus not comformal (see [method is_conformal]). Applying this basis to a 3D model will result in the model appearing distorted.
```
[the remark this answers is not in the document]
Mickeon: > ##### This PR may contain multiple versions of the same line I was not sure about, denoted with "??". Reviewing of these lines is heavily appreciated.

I'll bring it right on top of the PR's description next time.
Which ones do you prefer or do you have more suggestions?

TokageItLab:
> Avoid advanced math terms at all costs:
> "Orthonormal", "orthogonal", "transformation". Avoid mentioning them unless it's very appropriate.
Exception to terms like "normalized" can stay. They have a clear and common purpose in Godot.

I suggest that these terms be left as needed. This is because explaining things like "whether the angles of all basis vectors are perpendicular" or "whether the scales are all 1's" every time would be redundant.

Instead, I reccomend you write an explanation of the term in [Matrices and transforms](https://docs.godotengine.org/en/stable/tutorials/math/matrices_and_transforms.html) and link to it.

Mickeon:
> I suggest that these terms be left as needed. This is because explaining things like "whether the angles of all basis vectors are perpendicular" or "whether the scales are all 1's" every time would be redundant.

I am not sure how to feel about this, because the class reference as is in this PR does read just fine even without them.

Perhaps one solution is to add a note that *later* explains what those terms mean, in relation to the current method?

doc/classes/Basis.xml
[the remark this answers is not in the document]
TokageItLab: When noting the length of the vector of basis matrices, it is necessary to specify that the right-hand orthogonal system is positive. This is because focusing only on length can lead to negative scales.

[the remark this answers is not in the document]
Mickeon: @TokageItLab If you have more insight like this on where it is absolutely necessary to mention these terms please do highlight them to make this PR as right as it can be

Mickeon:
I updated the PR with an additional line about the meaning of a few terms, as I wait from more precise feedback from @TokageItLab

TokageItLab:
I simply point out that there is no definition of what the negative of the basis vector means in visually.

Actually, that definition may not be necessary until rendering, but if you are trying to draw/link a figure, you will need to annotate that the coordinate system of the Godot and the figure are same.

Also, it would be necessary to mention Row-major order and Column-major order. For those moving from other game engines, this can be confusing. See also https://www.mindcontrol.org/~hplus/graphics/matrix-layout.html.

Mickeon:
> Also, it would be necessary to mention Row-major order and Column-major order. For those moving from other game engines, this can be confusing easily (and I have seen many such people). See also https://www.mindcontrol.org/~hplus/graphics/matrix-layout.html.

I suppose this can be mentioned, yeah... It does sound important.

> Actually, that definition may not be necessary until rendering, but if you are trying to draw/link a figure, you will need to annotate that the coordinate system of the Godot and the figure are same.

But I am not sure about. Would it not be for granted that both coordinate systems are the same?

TokageItLab:
As far as I know, mathematics basically uses the right-hand coordinate system, but in cases such as when a diagram created by someone working in DirectX is referenced, the left-hand coordinate system may be used and the direction of rotation, etc. may be flipped.

In order to counter such a claim from a DirectX user that the rotation is reversed, it is better to specify that Godot is in the right-hand coordinate system.

@aaronfranke previously wrote about the coordinate system for the asset pipeline https://docs.godotengine.org/en/4.1/tutorials/assets_pipeline/importing_scenes.html#d-asset-direction-conventions, but I am not sure if it is appropriate to link to that. Perhaps you can find a more appropriate link.

doc/classes/Basis.xml
```diff
@@ -37,14 +37,15 @@
...
+				Constructs a [Basis] that only represents rotation, rotated around the [param axis] by the given [param angle], in radians. The axis must be a normalized vector.
+				[b]Note:[/b] This is the same as using [method rotated] on the [constant IDENTITY] basis. With more than one angle consider using [method from_euler], instead.
 			</description>
 		</constructor>
 		<constructor name="Basis">
 			<return type="Basis" />
 			<param index="0" name="from" type="Quaternion" />
 			<description>
-				Constructs a pure rotation basis matrix from the given quaternion.
+				Constructs a [Basis] that only represents rotation from the given [Quaternion].
```
TokageItLab, line 48: It might be worth mentioning that the conversion from Quaternion->Basis->Quaternion is reversible, but the conversion from Basis->Quaternion->Basis is irreversible because Quaternion does not properly account for scale.

Currently Godot sometimes causes unexpected transforming when converting from Basis to Quaternion with negative scales, but I haven't investigated if there is a way to handle this well.

Mickeon:
Linking to more advanced "lessons" does sound good to me. The goal of this PR was to make the Basis class reference more accessible, so talking too much about advanced topics is a bit out of scope. Even my head is spinning about this...

> As far as I know, mathematics basically uses the right-hand coordinate system, but in cases such as when a diagram created by someone working in DirectX is referenced, the left-hand coordinate system may be used and the direction of rotation, etc. may be flipped.

This is true... Different programs all use different coordinate systems.
As far as I Godot only uses two: one for 2D and another for 3D. In this PR, you may notice I tried to make the Basis' axes easier to imagine (_The IDENTITY constant, x, y, and z descriptions_). This looked good enough and in my opinion, this kind of information is MUCH more useful to be included in **Transform3D**'s class reference _(coming soon)_.

But if it indeed is not enough, I could pull my favourite reference sheet as a link for users to see: https://bevy-cheatbook.github.io/fundamentals/coords.html

doc/classes/Basis.xml
```diff
@@ -4,9 +4,9 @@
...
-		A 3×3 matrix used for representing 3D rotation and scale. Usually used as an orthogonal basis for a [Transform3D].
-		Contains 3 vector fields X, Y and Z as its columns, which are typically interpreted as the local basis vectors of a transformation. For such use, it is composed of a scaling and a rotation matrix, in that order (M = R.S).
-		Basis can also be accessed as an array of 3D vectors. These vectors are usually orthogonal to each other, but are not necessarily normalized (due to scaling).
+		The [Basis] built-in [Variant] type represents a 3x3 [url=https://en.wikipedia.org/wiki/Matrix_(mathematics)]matrix[/url] used for 3D rotation, scale, and shearing. It is frequently used within a [Transform3D].
+		A [Basis] is composed by 3 axis vectors, each representing a column of the matrix: [member x], [member y], and [member z]. The length of each axis ([method Vector3.length]) influences the basis' scale, while the direction of all axes influence the rotation. Usually, these axes are perpendicular to one another. However, when you rotate any axis individually, the basis becomes sheared. Applying a sheared basis to a 3D model will make the model appear distorted.
```
aaronfranke, line 8: You could make the argument that "distorted" applies to anything that doesn't pass the `is_conformal` check, including non-uniform scale. I'm not sure how specifically to rephrase it.

doc/classes/Basis.xml
```diff
@@ -159,13 +173,13 @@
...
 			<description>
-				Returns the inverse of the matrix.
+				Returns the [url=https://en.wikipedia.org/wiki/Invertible_matrix]inverse of this basis' matrix[/url].
 			</description>
 		</method>
 		<method name="is_conformal" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns [code]true[/code] if the basis is conformal, meaning it preserves angles and distance ratios, and may only be composed of rotation and uniform scale. Returns [code]false[/code] if the basis has non-uniform scale or shear/skew. This can be used to validate if the basis is non-distorted, which is important for physics and other use cases.
+				Returns [code]true[/code] if this basis is conformal. A conformal basis only represents rotation or scale, with no shearing, and thus it is not distorted. This method can be especially useful during physics calculations.
```
aaronfranke, line 182: This is not correct. Conformal requires uniform scale. If the scale is not uniform, it is not conformal.

doc/classes/Basis.xml
```diff
@@ -340,94 +370,124 @@
 	</methods>
 	<members>
 		<member name="x" type="Vector3" setter="" getter="" default="Vector3(1, 0, 0)">
-			The basis matrix's X vector (column 0). Equivalent to array index [code]0[/code].
+			The basis' X axis, and the column [code]0[/code] of the matrix.
```
aaronfranke, line 373: This is not correct grammar, `s'` is used if the word is both plural and has ownership, but `basis` is not plural.

doc/classes/Basis.xml
```diff
@@ -281,34 +297,48 @@
...
-				Transposed dot product with the Z axis of the matrix.
+				Returns the transposed dot product between [param with] and the [member z] axis (see [method transposed]).
+				This is equivalent to [code]basis.x.z * vector.x + basis.y.z * vector.y + basis.z.z * vector.z[/code].
 			</description>
 		</method>
 		<method name="transposed" qualifiers="const">
 			<return type="Basis" />
 			<description>
-				Returns the transposed version of the matrix.
+				Returns the transposed version of this basis. This turns the basis matrix' columns into rows, and its rows into columns.
```
aaronfranke, line 341: Typo here.

doc/classes/Basis.xml
```diff
@@ -340,94 +370,124 @@
...
+			On an uniform basis, this vector points to the back ([constant Vector3.BACK]).
 		</member>
 	</members>
 	<constants>
 		<constant name="IDENTITY" value="Basis(1, 0, 0, 0, 1, 0, 0, 0, 1)">
-			The identity basis, with no rotation or scaling applied.
+			The identity basis. This is a basis with no rotation, no shearing, and its scale being [code]1[/code]. This means that:
+			- The [member x] points to the right ([constant Vector3.RIGHT]);
+			- The [member y] points to the top ([constant Vector3.UP]);
+			- The [member z] points to the back ([constant Vector3.BACK]).
```
aaronfranke, line 390: You are mixing directions and sides here. Top is a side, back is a direction. We should use directions here, so top should be changed to up. If we were talking about sides then back would be changed to rear.

doc/classes/Basis.xml
```diff
@@ -340,94 +370,124 @@
...
-				This operator multiplies all components of the [Basis], which scales it uniformly.
+				Multiplies all components of the [Basis] by the given [float]. This affects the basis' scale proportionately.
 			</description>
 		</operator>
 		<operator name="operator *">
 			<return type="Basis" />
 			<param index="0" name="right" type="int" />
 			<description>
-				This operator multiplies all components of the [Basis], which scales it uniformly.
+				Multiplies all components of the [Basis] by the given [int]. This affects the basis' scale proportionately.
```
aaronfranke, line 460: "Uniformly" is *the* term for this operation, not "proportionately". "basis uniform scale" on Google gives useful search results, "basis proportionate scale" gives no useful results.

aaronfranke (changes requested):

doc/classes/Basis.xml
Mickeon: One draft of the same sentence was in fact pointing `is_comformal` but I couldn't have it sounding nice or at the very least overbearing. It needs to be mentioned though, so I need to find a way.

doc/classes/Basis.xml
Mickeon: The "Basis to Quaternion" quirk should be written in Quartenion's constructor. I 'll keep that in mind.

Mickeon:
Thank you @TokageItLab and @aaronfranke very much for your time, here comes a second revision.

doc/classes/Basis.xml
```diff
@@ -4,10 +4,12 @@
...
+		A [Basis] is [b]orthogonal[/b] if its axes are perpendicular from each other, which allows it to represent rotations. A basis is [b]normalized[/b] if the length of every axis is [code]1[/code]. A basis is [b]uniform[/b] if all axes share the same length (see [method get_scale]).
 		For a general introduction, see the [url=$DOCS_URL/tutorials/math/matrices_and_transforms.html]Matrices and transforms[/url] tutorial.
+		[b]Note:[/b] Godot uses the [b]right-hand coordinate system[/b] (+X is right, +Y is up, and +Z is back). This may be different from the coordinate system you are used to. For example, by convention, 3D assets are rotated towards the camera (+X is [b]left[/b], +Y is up, and +Z is [b]front[/b]). For more information, see the [url=$DOCS_URL/tutorials/assets_pipeline/importing_scenes.html#d-asset-direction-conventions]Importing 3D Scenes[/url] tutorial.
+		[b]Note:[/b] The basis matrices are stored in [url=https://www.mindcontrol.org/~hplus/graphics/matrix-layout.html]column-major[/url] order, which is the same as OpenGL. In comparison, DirectX uses row-major order.
```
aaronfranke, line 12: This is not quite accurate. The C++ code stores them in row-major. It would be more accurate to say they are *exposed* as column-major, not *stored* as column-major.

doc/classes/Basis.xml
```diff
@@ -4,10 +4,12 @@
...
+		A [Basis] is [b]orthogonal[/b] if its axes are perpendicular from each other, which allows it to represent rotations. A basis is [b]normalized[/b] if the length of every axis is [code]1[/code]. A basis is [b]uniform[/b] if all axes share the same length (see [method get_scale]).
 		For a general introduction, see the [url=$DOCS_URL/tutorials/math/matrices_and_transforms.html]Matrices and transforms[/url] tutorial.
+		[b]Note:[/b] Godot uses the [b]right-hand coordinate system[/b] (+X is right, +Y is up, and +Z is back). This may be different from the coordinate system you are used to. For example, by convention, 3D assets are rotated towards the camera (+X is [b]left[/b], +Y is up, and +Z is [b]front[/b]). For more information, see the [url=$DOCS_URL/tutorials/assets_pipeline/importing_scenes.html#d-asset-direction-conventions]Importing 3D Scenes[/url] tutorial.
```
aaronfranke, line 11: This is misleading, coordinate system handedness has nothing to do with the conventions of how assets are oriented. They are separate things.

To make an analogy, it's like saying "The USA uses US dollars. This may be different from what you are used to, for example, they use feet and inches instead of meters". That statement would imply that US dollars imply feet and inches. No, they are completely separate things.

Mickeon: What could be said in this case, then?

aaronfranke: Actually, looking at this closer, Basis actually has no inherent handedness. I'm pretty sure that all of the same math methods will work exactly as-is if you used the same Basis code with a left-handed coordinate system. What makes Godot right-handed is its interpretation of the data we store in Basis.

The only places where conventions are encoded into Basis are 1) In the default Euler angle order YXZ which works best for Y-vertical, X-sides, Z-forward/back systems, and 2) The directions used by the `looking_at` method.

Mickeon: So the note about the coordinate system would be best suited for Transform3D, instead? Omitting it here may conflict with what @TokageItLab requested to mention above.

TokageItLab: > Actually, that definition may not be necessary until rendering, but if you are trying to draw/link a figure, you will need to annotate that the coordinate system of the Godot and the figure are same.

As I said above, no coordinate system definition is required until rendering. However, since you seem to have links to some figures, so it would be better to indicate whether the figure and the result rendered by Godot match or not.

aaronfranke: To clarify, the preamble I wrote in the asset direction conventions section that Tokage linked is prerequisite information to explaining the asset direction conventions in terms of that coordinate system. So it's "Godot is X, and in terms of X, Godot is Y" not "Godot is X, and because of X, Godot is Y". An immediate minor improvement would be to remove the "For example" prefix to that sentence.

In terms of the documentation of the Basis structure, we need to make it clear specifically what documentation applies to Basis itself, what applies to its interpretation, and what applies to a specific method.

It may also be useful to mention specifically what engines don't use right-handed, instead of a vague notion of not being what you're used to. Right-handed is the dominant system, it's just a few outliers like Unity and Unreal that use left-handed.

I recommend that the text be changed to this:

```
Godot uses a [url=https://en.wikipedia.org/wiki/Right-hand_rule]right-handed coordinate system[/url]. Most software is right-handed, but this is different from the Unity and Unreal engines. Built-in types like [Camera3D] use the convention of +X is right, +Y is up, and +Z is back. Other objects within this coordinate system may use different direction conventions. For more information, see the [url=$DOCS_URL/tutorials/assets_pipeline/importing_scenes.html#d-asset-direction-conventions]Importing 3D Scenes[/url] tutorial.
```

Feel free to tweak if needed. Any further elaboration of direction conventions should be localized to the description of the `looking_at` method IMO.

Mickeon:
I updated the PR to try and address the feedback above. Thank you very much to the both of you for the very detailed explanations and patience

I also slightly changed the first sentence of the description ever so slightly. Thought it was noting.

Mickeon:
I updated the PR to try and address the feedback above. Thank you very much to the both of you for the very detailed explanations and patience

I also slightly changed the first sentence of the description ever so slightly. Thought it was noting.

doc/classes/Basis.xml
```diff
@@ -281,34 +301,48 @@
...
+				[/codeblock]
 			</description>
 		</method>
 		<method name="tdotx" qualifiers="const">
 			<return type="float" />
 			<param index="0" name="with" type="Vector3" />
 			<description>
-				Transposed dot product with the X axis of the matrix.
+				Returns the transposed dot product between [param with] and the [member x] axis (see [method transposed]).
+				This is equivalent to [code]basis.x.x * vector.x + basis.y.x * vector.y + basis.z.x * vector.z[/code].
```
aaronfranke, line 323: This is the opposite of what it should be. I think you got tripped up by the C++ code using rows. `rows[1][0]` in C++ is `basis.x.y` in GDScript.

```suggestion
				This is equivalent to [code]basis.x.dot(vector)[/code].
```

Honestly, we should consider deprecating this method. It doesn't seem very useful. `basis.tdotx(vector)` is the same amount of characters as `basis.x.dot(vector)`, and the engine itself never uses this method outside of internal calculations in Basis itself.

doc/classes/Basis.xml
```diff
@@ -187,15 +205,16 @@
...
+				The up axis (+Y) points as close to the [param up] vector as possible while staying perpendicular to the forward axis. The returned basis is orthonormalized (see [method orthonormalized]). The [param target] and [param up] vectors cannot be [constant Vector3.ZERO], and cannot be parallel to each other.
 			</description>
 		</method>
 		<method name="orthonormalized" qualifiers="const">
 			<return type="Basis" />
 			<description>
-				Returns the orthonormalized version of the matrix (useful to call from time to time to avoid rounding error for orthogonal matrices). This performs a Gram-Schmidt orthonormalization on the basis of the matrix.
+				Returns the orthonormalized version of this basis. An orthonormal basis is both [i]orthogonal[/i] (the axes are perpendicular to each other) and [i]normalized[/i] (the axes have a length of [code]1[/code]), which also means it can only represent rotation.
+				It is often useful to call this method to avoid rounding errors on a rotating basis;
```
aaronfranke, line 217: You wrote a semicolon here instead of a colon.

doc/classes/Basis.xml
```diff
@@ -53,24 +57,29 @@
...
 		</method>
 		<method name="from_euler" qualifiers="static">
 			<return type="Basis" />
 			<param index="0" name="euler" type="Vector3" />
 			<param index="1" name="order" type="int" default="2" />
 			<description>
-				Constructs a pure rotation Basis matrix from Euler angles in the specified Euler rotation order. By default, use YXZ order (most common). See the [enum EulerOrder] enum for possible values.
+				Constructs a new [Basis] that only represents rotation from the given [Vector3] of [url=https://en.wikipedia.org/wiki/Euler_angles]Euler angles[/url], in radians.
+				- The [member Vector3.x] should contain the angle around the [member x] axis (pitch);
+				- The [member Vector3.y] should contain the angle around the [member y] axis (yaw);
```
aaronfranke, line 81: We don't need semicolons here, these can just be periods.

doc/classes/Basis.xml
Mickeon: Thank you for finding this back to me. I saw this error was somewhere in my PRs but I forgot which one it was.

doc/classes/Basis.xml
Mickeon: I'd be almost all up for it. I don't see the point of this method for most users, although it may run faster.

doc/classes/Basis.xml
```diff
@@ -340,94 +374,125 @@
...
+			# | 1 | 0 | 0
+			# | 0 | 1 | 0
+			# | 0 | 0 | 1
+			[/codeblock]
 			This is identical to creating [constructor Basis] without any parameters. This constant can be used to make your code clearer, and for consistency with C#.
 		</constant>
 		<constant name="FLIP_X" value="Basis(-1, 0, 0, 0, 1, 0, 0, 0, 1)">
-			The basis that will flip something along the X axis when used in a transformation.
+			?? The basis that will flip another basis along its [member x] axis when transformed (multiplied) by it.
+			?? When any basis is transformed (multiplied) by this basis, it will be flipped along its [member x] axis.
```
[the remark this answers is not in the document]
Mickeon: Do check this part again to make sure I understood this right. A lot of concepts are extremely hard to summarize, and it's difficult to pick the most appropriate ones. I'm not sure where to fit.:
> This applies a scale of -1 on the global X axis (where global in this context can also mean "parent-relative").

doc/classes/Basis.xml
```diff
@@ -340,94 +374,127 @@
...
+			# | 1 | 0 | 0
+			# | 0 | 1 | 0
+			# | 0 | 0 | 1
+			[/codeblock]
 			This is identical to creating [constructor Basis] without any parameters. This constant can be used to make your code clearer, and for consistency with C#.
 		</constant>
 		<constant name="FLIP_X" value="Basis(-1, 0, 0, 0, 1, 0, 0, 0, 1)">
-			The basis that will flip something along the X axis when used in a transformation.
+			When any basis is transformed (multiplied) by [constant FLIP_X], it will be flipped along its [member x] axis. This negates all [member x]'s components.
+			If [constant FLIP_X] is transformed by any other basis, the [member Vector3.x] component of all axes will be negated, instead.
```
aaronfranke, line 411: The wording here seems a bit weird, conceptually speaking. When we multiply the order is `parent * child`. We probably would say that child is transformed by parent, not parent is transformed by child, but really the result is a composition of both matrices, so perhaps we can just say multiplied.

The statement "it will be flipped along its [member x] axis" is not correct. It doesn't flip the whole basis along the axis, it just flips the axis. Proof:

```gdscript
func _ready():
	var b := Basis.IDENTITY
	b.y = Vector3(1, 1, 0)
	print(b * Basis.FLIP_X)
	# Output: [X: (-1, 0, 0), Y: (1, 1, 0), Z: (0, 0, 1)]
```

```suggestion
			When any basis is multiplied by [constant FLIP_X], it negates all components of the [member x] axis (the X column).
			When [constant FLIP_X] is multiplied by any basis, it negates the [member Vector3.x] component of all axes (the X row).
```

Mickeon:
Also would like to note, I have purposely omitted `looking_at` from this PR because I believe it would be better to address the description of it and all other similar methods in a separate, focused PR.

doc/classes/Basis.xml
```diff
@@ -247,7 +267,7 @@
 			<return type="Basis" />
 			<param index="0" name="scale" type="Vector3" />
 			<description>
-				Introduce an additional scaling specified by the given 3D scaling factor.
+				Returns this basis with each axis' components scaled by the given [param scale]'s components.
```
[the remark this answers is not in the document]
Mickeon: We're getting back to the territory of "I'm not sure most users need this information". Those that are heavily interested in the Basis's rows would have already inferred this and have their doubts "quenched" by the example below (but perhaps it doesn't hurt to repeat it).

Also, "global or parent-relative scale" does not feel like a concept that applies to a standalone Basis, and because of this I'm finding difficult to integrate it nicely.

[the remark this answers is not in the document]
aaronfranke: Basically by "global or parent-relative scale" I am trying to convey the concept of a transform being applied above/before/from the left/etc, not below/after/from the right/etc as a child of the basis.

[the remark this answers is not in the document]
Mickeon: Wouldn't this be more appropriate to describe in **Transform3D**'s class reference, instead? I will keep it in mind still, because **Transform3D**'s docs make similar claims.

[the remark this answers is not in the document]
aaronfranke: Well, a Basis is still a linear transformation, just with only the rotation/scale/shear/etc and no position.

aaronfranke (approved):
This is great! I don't see any more problems with this, it's a great improvement.

fire (approved):

AThousandShips (approved):

akien-mga:
Thanks!

Mickeon:
The late-night merge!!!!! 🌙

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

PR #87179, opened 2024-01-14
Add support for extra spacing relative to the font size.

bruvzg:


scene/resources/font.compat.inc
```diff
@@ -31,11 +31,16 @@
 #ifndef DISABLE_DEPRECATED
 
 RID Font::_find_variation_compat_80954(const Dictionary &p_variation_coordinates, int p_face_index, float p_strength, Transform2D p_transform) const {
-	return find_variation(p_variation_coordinates, p_face_index, p_strength, p_transform, 0, 0, 0, 0);
+	return find_variation(p_variation_coordinates, p_face_index, p_strength, p_transform, Vector4i(), Vector4());
 }
 
+RID Font::_find_variation_compat_XXXXX(const Dictionary &p_variation_coordinates, int p_face_index, float p_strength, Transform2D p_transform, int p_spacing_top, int p_spacing_bottom, int p_spacing_space, int p_spacing_glyph) const {
+	return find_variation(p_variation_coordinates, p_face_index, p_strength, p_transform, Vector4i(p_spacing_top, p_spacing_bottom, p_spacing_space, p_spacing_glyph), Vector4());
+};
```
AThousandShips, line 39:
```suggestion
}
```

scene/resources/font.h
```diff
@@ -137,6 +138,7 @@ class Font : public Resource {
 	virtual int get_font_stretch() const;
 
 	virtual int get_spacing(TextServer::SpacingType p_spacing) const { return 0; };
+	virtual double get_spacing_relative(TextServer::SpacingType p_spacing) const { return 0; };
```
AThousandShips, line 141:
```suggestion
	virtual double get_spacing_relative(TextServer::SpacingType p_spacing) const { return 0; }
```

scene/resources/font.h
```diff
@@ -118,7 +119,7 @@ class Font : public Resource {
 	virtual TypedArray<Font> get_fallbacks() const;
 
 	// Output.
-	virtual RID find_variation(const Dictionary &p_variation_coordinates, int p_face_index = 0, float p_strength = 0.0, Transform2D p_transform = Transform2D(), int p_spacing_top = 0, int p_spacing_bottom = 0, int p_spacing_space = 0, int p_spacing_glyph = 0) const { return RID(); };
+	virtual RID find_variation(const Dictionary &p_variation_coordinates, int p_face_index = 0, float p_strength = 0.0, Transform2D p_transform = Transform2D(), const Vector4i &p_spacing = Vector4i(), const Vector4 &p_spacing_relative = Vector4()) const { return RID(); };
```
AThousandShips, line 122:
```suggestion
	virtual RID find_variation(const Dictionary &p_variation_coordinates, int p_face_index = 0, float p_strength = 0.0, Transform2D p_transform = Transform2D(), const Vector4i &p_spacing = Vector4i(), const Vector4 &p_spacing_relative = Vector4()) const { return RID(); }
```

misc/extension_api_validation/4.1-stable_4.2-stable.expected
```diff
@@ -91,7 +91,7 @@ Added optional argument. Compatibility methods registered.
 
 GH-80954
 --------
-Validate extension JSON: Error: Field 'classes/Font/methods/find_variation/arguments': size changed value in new API, from 4 to 8.
+Validate extension JSON: Error: Field 'classes/Font/methods/find_variation/arguments': size changed value in new API, from 4 to 6.
```
AThousandShips, line 94: Should not be updated here, afaik, it is for 4.2 specially

bruvzg: I do not think there's any way to avoid it, since validator is always comparing it to the latest version, not a compatibility method.

In this case, argument count changed in the following way: `4` (in 4.1) ⇾ `8` (in 4.2) ⇾ `6` (in this PR).

AThousandShips: But there's a compatibility version registered, did this error out when not adding this change?

bruvzg: Yes, without this change it's giving an error in `Compatibility to 4.1-stable is broken in the following ways:` part of the script. There are compatibility methods for both `4` and `8` argument versions.

AThousandShips: CC @dsnopek any idea here?

abcjjy:
When will this be merged?

PR #87180, opened 2024-01-14
Include `animation.length` in Animation example

31:
The length of an Animation isn't automatically set by adding keys, and it must be set manually. The existing example only has keys up to 0.5s, so the default value of 1.0s may be acceptable. However, this results in unexpected behavior for anyone who makes an animation longer than 1.0s. Include animation.length in the example because it's important.

Also, increase the key position to 2.0s so it won't be confusing that the assignment is somewhat redundant.

### DIFF
--- a/doc/classes/Animation.xml
+++ b/doc/classes/Animation.xml
@@ -8,21 +8,23 @@
 		[codeblocks]
 		[gdscript]
 		# This creates an animation that makes the node "Enemy" move to the right by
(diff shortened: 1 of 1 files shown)


Calinou (approved):

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.1.4.

PR #87183, opened 2024-01-14
Add tooltips to the export buttons

### LINKED ISSUE #8870: Add tooltips to project export buttons

Anything

When exporting a project from Project > Export, the user is shown three buttons at the bottom for exporting, namely Export All, Export Project and Export PCK/ZIP. Of these buttons, it's difficult for a beginner to know what the difference is and which one is needed. I would've guessed I needed to Export Project, however this did not do what I expected.

You can read docs, but even these are vague about the differences between these buttons (perhaps docs also need further details?) https://docs.godotengine.org/en/stable/tutorials/export/exporting_projects.html

Currently, only the Export All button has a tooltip when hovered, this is helpful for explaining what the button does. Add tooltips for the other two export buttons to explain what these do.

Tooltips are used around the editor so users are familiar with hovering for a second over something to see more info. However this is unavailable for the Export Project and Export PCK/ZIP buttons, despite their names not clearly indicating their use.

No, and would be used frequently

Exporting projects is core

hakro:
Closes https://github.com/godotengine/godot-proposals/issues/8870

### DIFF
--- a/editor/export/project_export.cpp
+++ b/editor/export/project_export.cpp
@@ -1395,12 +1395,14 @@ ProjectExportDialog::ProjectExportDialog() {
 
 	set_cancel_button_text(TTR("Close"));
 	set_ok_button_text(TTR("Export PCK/ZIP..."));
+	get_ok_button()->set_tooltip_text(TTR("Export the project resources as a PCK or ZIP package. This is not a playable build, only the project data without a Godot executable."));
 	get_ok_button()->set_disabled(true);
 #ifdef ANDROID_ENABLED
 	export_button = memnew(Button);
 	export_button->hide();
 #else
 	export_button = add_button(TTR("Export Project..."), !DisplayServer::get_singleton()->get_swap_cancel_ok(), "export");
+	export_button->set_tooltip_text(TTR("Export the project as a playable build (Godot executable and project data) for the selected preset."));
 #endif
 	export_button->connect("pressed", callable_mp(this, &ProjectExportDialog::_export_project));
 	// Disable initially before we select a valid preset


editor/export/project_export.cpp
```diff
@@ -1395,12 +1395,14 @@ ProjectExportDialog::ProjectExportDialog() {
 
 	set_cancel_button_text(TTR("Close"));
 	set_ok_button_text(TTR("Export PCK/ZIP..."));
+	get_ok_button()->set_tooltip_text("Export all the project files as a PCK or ZIP package.");
```
akien-mga, line 1398: Might be getting a bit verbose, but this might help address @JUSTCAMH's suggestion. And softens the "all the project files" as one might expect this means every single file is exported, while it only exports what Godot considers a resource (or what's configured in the Resources tab).
```suggestion
	get_ok_button()->set_tooltip_text("Export the project resources as a PCK or ZIP package. This is not a playable build, only the project data without a Godot executable.");
```

editor/export/project_export.cpp
```diff
@@ -1395,12 +1395,14 @@ ProjectExportDialog::ProjectExportDialog() {
...
 	get_ok_button()->set_disabled(true);
 #ifdef ANDROID_ENABLED
 	export_button = memnew(Button);
 	export_button->hide();
 #else
 	export_button = add_button(TTR("Export Project..."), !DisplayServer::get_singleton()->get_swap_cancel_ok(), "export");
+	export_button->set_tooltip_text(TTR("Export the project as a playable game for the selected preset only."));
```
akien-mga, line 1405:
```suggestion
	export_button->set_tooltip_text(TTR("Export the project as a playable build (Godot executable and project data) for the selected preset."));
```

hakro:
@JUSTCAMH the text has been enriched.
We can probably also add a paragraph detailing the differences in the documentation, so that the tooltips don't get too cluttered in the editor.

@akien-mga I squashed the commits, unfortunately that made it loose track of you co-authorship :(

editor/export/project_export.cpp
```diff
@@ -1395,12 +1395,14 @@ ProjectExportDialog::ProjectExportDialog() {
 
 	set_cancel_button_text(TTR("Close"));
 	set_ok_button_text(TTR("Export PCK/ZIP..."));
+	get_ok_button()->set_tooltip_text("Export the project resources as a PCK or ZIP package. This is not a playable build, only the project data without a Godot executable.");
```
akien-mga, line 1398: Forgot this before (same for the other tooltip), this should use `TTR` to be extracted for translation.
```suggestion
	get_ok_button()->set_tooltip_text(TTR("Export the project resources as a PCK or ZIP package. This is not a playable build, only the project data without a Godot executable."));
```

hakro: Good catch. Sorry for the miss.
On the bright side, you got your co-authorship back in the latest commit ^^
Thx

akien-mga (approved):

akien-mga:
Thanks!

PR #87186, opened 2024-01-14
[MP] Fix remote net ID cleanup

### LINKED ISSUE #86146: Is "SceneReplicationInterface::on_sync_receive: Ignoring sync data" an error?
### Tested versions

4.2 and HEAD @ aa5b6ed13e4644633baf2a8a1384c82e91c533a1

### System information

Windows 10

### Issue description

I have two MultiplayerSpawner that spawn scenes ("Skeleton" and "Knife" respectively).

The multiplayer authority for all the spawners and spawned scenes is the server.

I call queue_free() in the server in order to remove spawned scenes.

Everything seems to work well gameplay-wise, but the debug log gets one of those:
"Ignoring sync data from non-authority or for missing node."
sometimes on the client from here:
https://github.com/godotengine/godot/blob/aa5b6ed13e4644633baf2a8a1384c82e91c533a1/modules/multiplayer/scene_replication_interface.cpp#L879

I have added significant logging in the Godot source and I can see that when this happens:
* the notification for exit_tree is received by MultiplayerSynchronizer in the client, as a result
* _stop() is called on the MultiplayerSynchronizer in the client, which sets root_path_cache to ObjectID() (0).
* then ~1ms later the sync message arrives, root_path_cache.is_valid() is false as _stop() has set it to zero, as a result so sync.get_root_node() returns null and the message above is printed.

Here is a sample screenshot with my debug output:
image

I am not sure if I am doing something wrong or if it's just expected that somehow the "client" gets syncs shortly after being de-spawned and those cause this error.

If it's the latter, can we make this not appear in the debug log?

### Steps to reproduce

Create a MultiplayerSpawner, spawn children including a MultiplayerSynchronizer from the server, remove them by calling queue_free() in the server

I don't have a minimum reproduction project as of now. If the behavior described above is not expected, I can try to reproduce in a smaller form, but it may be hard given that it requires multiplayer.

Faless:
Synchronizers for spawned nodes were not correctly keeping track of the net ID assigned by the remote, preventing the replication from performing the proper cleanup.

This resulted in errors being thrown when sync messages were received after despawn (which is possible due to their unreliable nature).

Fixes #86146

### DIFF
--- a/modules/multiplayer/multiplayer_synchronizer.cpp
+++ b/modules/multiplayer/multiplayer_synchronizer.cpp
@@ -49,11 +49,11 @@ void MultiplayerSynchronizer::_stop() {
 	}
 #endif
 	root_node_cache = ObjectID();
-	reset();
 	Node *node = is_inside_tree() ? get_node_or_null(root_path) : nullptr;
 	if (node) {
 		get_multiplayer()->object_configuration_remove(node, this);
 	}
+	reset();
 }
 
 void MultiplayerSynchronizer::_start() {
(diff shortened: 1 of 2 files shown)


mhilbrunner (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #87190, opened 2024-01-14
[MP] Handle cleanup of "scene cache" nodes

Faless:
Make sure we delete the relevant ObjectID from the cache when the nodes are removed from tree.

### DIFF
--- a/modules/multiplayer/scene_cache_interface.cpp
+++ b/modules/multiplayer/scene_cache_interface.cpp
@@ -35,25 +35,61 @@
 #include "core/io/marshalls.h"
 #include "scene/main/node.h"
 #include "scene/main/window.h"
+#include "scene/scene_string_names.h"
+
+SceneCacheInterface::NodeCache &SceneCacheInterface::_track(Node *p_node) {
+	const ObjectID oid = p_node->get_instance_id();
+	NodeCache *nc = nodes_cache.getptr(oid);
+	if (!nc) {
+		nodes_cache[oid] = NodeCache();
+		p_node->connect(SceneStringNames::get_singleton()->tree_exited, callable_mp(this, &SceneCacheInterface::_remove_node_cache).bind(oid), Object::CONNECT_ONE_SHOT);
+	}
+	return nodes_cache[oid];
+}
+
+void SceneCacheInterface::_remove_node_cache(ObjectID p_oid) {
+	NodeCache *nc = nodes_cache.getptr(p_oid);
+	if (!nc) {
+		return;
+	}
+	for (KeyValue<int, int> &E : nc->recv_ids) {
+		PeerInfo *pinfo = peers_info.getptr(E.key);
+		ERR_CONTINUE(!pinfo);
+		pinfo->recv_nodes.erase(E.value);
+	}
+	for (KeyValue<int, bool> &E : nc->confirmed_peers) {
+		PeerInfo *pinfo = peers_info.getptr(E.key);
+		ERR_CONTINUE(!pinfo);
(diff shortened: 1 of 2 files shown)


Faless:
This is also a candidate for cherry pick, but it might be better to test it a bit more in 4.3 devs

modules/multiplayer/scene_cache_interface.cpp
```diff
@@ -244,22 +271,22 @@ bool SceneCacheInterface::send_object_cache(Object *p_obj, int p_peer_id, int &r
 Object *SceneCacheInterface::get_cached_object(int p_from, uint32_t p_cache_id) {
 	Node *root_node = SceneTree::get_singleton()->get_root()->get_node(multiplayer->get_root_path());
 	ERR_FAIL_NULL_V(root_node, nullptr);
-	HashMap<int, PathGetCache>::Iterator E = path_get_cache.find(p_from);
-	ERR_FAIL_COND_V_MSG(!E, nullptr, vformat("No cache found for peer %d.", p_from));
-
-	HashMap<int, PathGetCache::NodeInfo>::Iterator F = E->value.nodes.find(p_cache_id);
-	ERR_FAIL_COND_V_MSG(!F, nullptr, vformat("ID %d not found in cache of peer %d.", p_cache_id, p_from));
+	ERR_FAIL_COND_V(!peers_info.has(p_from), nullptr);
```
YuriSizov, line 274: The old way it was written with `find()` was more optimal, as it avoided double lookup. Is this no longer a concern here?

modules/multiplayer/scene_cache_interface.cpp
```diff
@@ -35,25 +35,61 @@
...
 #include "scene/main/node.h"
 #include "scene/main/window.h"
+#include "scene/scene_string_names.h"
+
+SceneCacheInterface::NodeCache &SceneCacheInterface::_track(Node *p_node) {
+	const ObjectID oid = p_node->get_instance_id();
+	NodeCache *nc = nodes_cache.getptr(oid);
+	if (!nc) {
+		nodes_cache[oid] = NodeCache();
+		p_node->connect(SceneStringNames::get_singleton()->tree_exited, callable_mp(this, &SceneCacheInterface::_remove_node_cache).bind(p_node->get_instance_id()), Object::CONNECT_ONE_SHOT);
```
YuriSizov, line 45:
```suggestion
		p_node->connect(SceneStringNames::get_singleton()->tree_exited, callable_mp(this, &SceneCacheInterface::_remove_node_cache).bind(oid), Object::CONNECT_ONE_SHOT);
```

YuriSizov (approved):
Code seems fine otherwise.

modules/multiplayer/scene_cache_interface.cpp
Faless: Thanks, fixed!

akien-mga:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

ysxsj:
When executing QueueFree() on a node containing MultiplayerSynchronizer, or when executing QueueFree() on a node containing [Rpc] in a mounting script, an error message will be reported: _remove_mode_cache: Condition "! Pinfo" is true Continuing. In modules/multiplayer/scend_cache_interface. cpp: 57 @ _remove_mode_cache()

akien-mga:
> Cherry-picked for 4.2.2.

Cherry-picked reverted for 4.2.3 as it introduced a regression (fixed in 4.3 by a partial revert #94984).

Issue #50930, opened 2021-07-27
TabContainer resets visibility to first tab in editor


chucklepie:
### Godot version

3.3 stable

### System information

linux mint

### Issue description

You're working on a single tab and need to switch to another scene or code. When you go back, whatever tab(s) you had selected is/are removed and the first one is selected.

![image](https://user-images.githubusercontent.com/12863685/127138119-f0d1acf2-9804-46e5-80af-993d0b0510c0.png)

This is really very annoying and forces you to constantly update visibility in the editor or temporarily move your tab (if you are editing just one) to the very top.

btw, It wouldn't be 'as bad' if you could drag items down (but still a pain) but Godot stops you from moving items to the bottom of the list because when at the bottom it eithier changes level to one before or one after, it will not let you move a sibling to be the last one.

### Steps to reproduce

1. create a tab container with multiple tabs
2. select one tab
3. change to another open scene then go back

### Minimal reproduction project

[game.zip](https://github.com/godotengine/godot/files/6888588/game.zip)

pycbouh:
Can you demonstrate what you do with a video? Those steps are not enough to reproduce the issue. Changes to the visibility and selection in the scene dock are preserved when you switch between the scenes.

chucklepie:
Sure. I didn't because I presumed it was simple and in case it was already a case. Give me a few minutes.

chucklepie:
@pycbouh 
In the video, TabDemo is selected and visible, switch scene, switch back and it's now at the first. I repeat with multiple and the same happens.

https://user-images.githubusercontent.com/12863685/127145523-6283d110-ed46-4162-9717-a3278213d05b.mp4

timothyqiu:
The supposed way of switching tabs is changing the `Current Tab` property of `TabContainer` I think.

The editor is trying to re-sync the scene with `Current Tab`.

Calinou:
> btw, It wouldn't be 'as bad' if you could drag items down (but still a pain) but Godot stops you from moving items to the bottom of the list because when at the bottom it eithier changes level to one before or one after, it will not let you move a sibling to be the last one.

You can select a node and use <kbd>Ctrl + Up</kbd>/<kbd>Ctrl + Down</kbd> to move it in the scene tree. This is more precise than using drag-and-drop for this kind of stuff.

timothyqiu:
[...]

PR #87194, opened 2024-01-14
Allow no tabs to be selected in TabBar and TabContainer

### LINKED ISSUE #8233: Allow no tabs to be selected in `TabBar` and `TabContainer`

Godot Editor

In an effort to get the bottom panel to have more functionality like regular docks do, for #4565 and #7156.
I was trying to make the Bottom Panel use a `TabBar` instead of an `HBoxContainer` of buttons, but I realized the bottom panel needs to be collapsible and be able to have no tabs selected.

The plan is to first make it use `TabBar` since there is a plugin API change for `add_bottom_panel_item`.
Then it will be changed to use a `TabContainer` once it gets some other features like #2250 and https://github.com/godotengine/godot/pull/82468.

However, if we don't want Bottom Panel to use `TabBar` or break the current API, then the needed functionality can be implemented in different ways instead.

`TabBar` and `TabContainer` would have an optional setting that allows no tabs to be selected. When a selected tab is clicked, it deselects.

Only the tab header would be shown when there are no tabs selected.

This will let the Bottom Panel shrink with `clip_tabs` for #7156 and situations with lots of bottom panel items. It will also allow the tabs to be rearranged with `drag_to_rearrange_enabled` (only within Bottom Panel at first).

`TabBar` would have a `allow_deselection` (or `toggle_tabs`?) field that when enabled allows no tabs to be selected.
When clicking on the current tab, `current_tab` gets set to -1.
The `tab_changed` signal should be emitted when this happens.

For `TabContainer`, when `current_tab` is -1 hide all child's content.

Not really, you would have to re-implement `TabBar`.

It's for the editor.

### LINKED ISSUE #21223: [feature] TabContainer: switching tabs while editing by clicking on the tab
Minor enhancement.
When editing controls inside a TabContainer, it's annoying to have to click the TabContainer and changing the Current Tab property to switch the tab. It should be possible to directly click on the tabs and make the Current Tab property automatically change. I've seen this feature in some GUI widgets editors.

### LINKED ISSUE #73785: Godot 4 rc3: TabBar child node set anchor not working in the editor.
### Godot version

v4.0.rc3.official (7e79aea)

### System information

Mac Air M1 2020

### Issue description

TabBar childs anchor not working in editor.
Ảnh màn hình 2023-02-23 lúc 04 28 01

### Steps to reproduce

Create a TabContainer node with some Tabbar node inside it. Then create some UI node inside Tabbar node. Set any anchor to child node, nothing happen. The child still in Vector2.ZERO.
Note: only child on first Tabbar node works.

[Uploading GodotProject.zip…]()

### LINKED ISSUE #74653: Editor preview of the contents of a TabContainer is incorrect for every tab except the first one
### Godot version

### System information

Linux

### Issue description

Creating a tabbed interface for options, the Y position for elements inside the Audio tab (second tab) appear to be incorrect in comparison to the Graphics tab (first tab).

Compare:
01-Graphics_tab_in_editor
02-Audio_tab_in_editor

That said, at runtime these appear to be in the correct positions.

Compare:
03-Graphics_tab_runtime
04-Audio_tab_runtime

### Steps to reproduce

A minimum scene tree:
```
Control
    TabContainer
        Control
            Label
        Control
            Label
```

Where the labels have the same position according to Layout/Transform.

[74653_demonstrator.zip](https://github.com/godotengine/godot/files/10931078/74653_demonstrator.zip) should be a minimal reproduction.

kitbdev:
- closes https://github.com/godotengine/godot-proposals/issues/8233
- related #21223
- fixes #74653
- fixes #73785
- fixes `remove_tab()` and `move_tabs()` not updating previous tab properly.
- related: https://github.com/godotengine/godot/issues/50930#issuecomment-1782019598

When TabContainer children visibility changes, that tab is marked as current and other tabs are deselected.

Adds deselect_enabled property to TabBar and TabContainer so that no tabs can be selected (disabled by default).
current_tab can now be -1 if this is enabled or there are no valid tabs.

Considerations:
- If there is only one child of a TabContainer and deselect_enabled is false, then the child cannot be hidden.
- This also prevents multiple TabContainer children from being visible at once.
- Changing visibility of TabContainer children changes the current tab, so when editing you need to make sure to switch to the tab you want to be the first selected tab before saving the scene.

### DIFF
--- a/doc/classes/TabBar.xml
+++ b/doc/classes/TabBar.xml
@@ -229,8 +229,11 @@
 		<member name="clip_tabs" type="bool" setter="set_clip_tabs" getter="get_clip_tabs" default="true">
 			If [code]true[/code], tabs overflowing this node's width will be hidden, displaying two navigation buttons instead. Otherwise, this node's minimum size is updated so that all tabs are visible.
 		</member>
-		<member name="current_tab" type="int" setter="set_current_tab" getter="get_current_tab" default="0">
-			Select tab at index [code]tab_idx[/code].
+		<member name="current_tab" type="int" setter="set_current_tab" getter="get_current_tab" default="-1">
+			The index of the current selected tab. A value of [code]-1[/code] means that no tab is selected and can only be set when [member deselect_enabled] is [code]true[/code] or if all tabs are hidden or disabled.
+		</member>
+		<member name="deselect_enabled" type="bool" setter="set_deselect_enabled" getter="get_deselect_enabled" default="false">
+			If [code]true[/code], all tabs can be deselected so that no tab is selected. Click on the current tab to deselect it.
 		</member>
 		<member name="drag_to_rearrange_enabled" type="bool" setter="set_drag_to_rearrange_enabled" getter="get_drag_to_rearrange_enabled" default="false">
 			If [code]true[/code], tabs can be rearranged with mouse drag.
--- a/scene/gui/tab_bar.cpp
+++ b/scene/gui/tab_bar.cpp
@@ -278,7 +278,11 @@ void TabBar::gui_input(const Ref<InputEvent> &p_event) {
 			}
 
 			if (found != -1) {
-				set_current_tab(found);
+				if (deselect_enabled && get_current_tab() == found) {
+					set_current_tab(-1);
+				} else {
+					set_current_tab(found);
+				}
 
 				if (mb->get_button_index() == MouseButton::RIGHT) {
 					// Right mouse button clicked.
@@ -616,8 +620,8 @@ void TabBar::set_tab_count(int p_count) {
 	if (p_count == 0) {
 		offset = 0;
 		max_drawn_tab = 0;
-		current = 0;
-		previous = 0;
+		current = -1;
+		previous = -1;
 	} else {
 		offset = MIN(offset, p_count - 1);
 		max_drawn_tab = MIN(max_drawn_tab, p_count - 1);
@@ -640,7 +644,12 @@ int TabBar::get_tab_count() const {
 }
 
 void TabBar::set_current_tab(int p_current) {
-	ERR_FAIL_INDEX(p_current, get_tab_count());
+	if (p_current == -1) {
+		// An index of -1 is only valid if deselecting is enabled or there are no valid tabs.
+		ERR_FAIL_COND_MSG(!_can_deselect(), "Cannot deselect tabs, deselection is not enabled.");
+	} else {
+		ERR_FAIL_INDEX(p_current, get_tab_count());
+	}
 
 	previous = current;
 	current = p_current;
@@ -1069,8 +1078,13 @@ void TabBar::add_tab(const String &p_str, const Ref<Texture2D> &p_icon) {
 	queue_redraw();
 	update_minimum_size();
 
-	if (tabs.size() == 1 && is_inside_tree()) {
-		emit_signal(SNAME("tab_changed"), 0);
+	if (tabs.size() == 1) {
+		if (is_inside_tree()) {
+			set_current_tab(0);
+		} else {
+			current = 0;
+			previous = -1;
+		}
 	}
 }
 
@@ -1082,8 +1096,8 @@ void TabBar::clear_tabs() {
 	tabs.clear();
 	offset = 0;
 	max_drawn_tab = 0;
-	current = 0;
-	previous = 0;
+	current = -1;
+	previous = -1;
 
 	queue_redraw();
 	update_minimum_size();
@@ -1094,34 +1108,43 @@ void TabBar::remove_tab(int p_idx) {
 	ERR_FAIL_INDEX(p_idx, tabs.size());
 	tabs.remove_at(p_idx);
 
-	bool is_tab_changing = current == p_idx && !tabs.is_empty();
+	bool is_tab_changing = current == p_idx;
 
 	if (current >= p_idx && current > 0) {
 		current--;
+	}
+	if (previous >= p_idx && previous > 0) {
+		previous--;
 	}
 
 	if (tabs.is_empty()) {
 		offset = 0;
 		max_drawn_tab = 0;
-		previous = 0;
+		current = -1;
+		previous = -1;
 	} else {
-		// Try to change to a valid tab if possible (without firing the `tab_selected` signal).
-		for (int i = current; i < tabs.size(); i++) {
-			if (!is_tab_disabled(i) && !is_tab_hidden(i)) {
-				current = i;
-				break;
-			}
-		}
-		// If nothing, try backwards.
-		if (is_tab_disabled(current) || is_tab_hidden(current)) {
-			for (int i = current - 1; i >= 0; i--) {
+		if (current != -1) {
+			// Try to change to a valid tab if possible (without firing the `tab_selected` signal).
+			for (int i = current; i < tabs.size(); i++) {
 				if (!is_tab_disabled(i) && !is_tab_hidden(i)) {
 					current = i;
 					break;
 				}
 			}
+			// If nothing, try backwards.
+			if (is_tab_disabled(current) || is_tab_hidden(current)) {
+				for (int i = current - 1; i >= 0; i--) {
+					if (!is_tab_disabled(i) && !is_tab_hidden(i)) {
+						current = i;
+						break;
+					}
+				}
+			}
+			// If still no valid tab, deselect.
+			if (is_tab_disabled(current) || is_tab_hidden(current)) {
+				current = -1;
+			}
 		}
-
 		offset = MIN(offset, tabs.size() - 1);
 		max_drawn_tab = MIN(max_drawn_tab, tabs.size() - 1);
 
@@ -1301,17 +1324,9 @@ void TabBar::_move_tab_from(TabBar *p_from_tabbar, int p_from_index, int p_to_in
 
 	if (!is_tab_disabled(p_to_index)) {
 		set_current_tab(p_to_index);
-		if (tabs.size() == 1) {
-			_update_cache();
-			queue_redraw();
-			emit_signal(SNAME("tab_changed"), 0);
-		}
 	} else {
 		_update_cache();
 		queue_redraw();
-		if (tabs.size() == 1) {
-			emit_signal(SNAME("tab_changed"), 0);
-		}
 	}
 
 	update_minimum_size();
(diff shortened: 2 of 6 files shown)


KoBeWi:
How does this fix https://github.com/godotengine/godot/issues/21223?

Also there is some bug when adding a first child to a newly created TabContainer. The child is invisible and you can't make it appear by changing tabs.

kitbdev:
Oh, I misread #21223, this only address the 'bonus' part of it, where clicking the eye to change visibility sets the current tab and the others are hidden. Changed to just be related.

And I think the invisible child issue is partly because I changed how the tab_changed signal fires in edge cases, but maybe I shouldn't change it for compatibility.
When adding the first child to a TabContainer, should the tab_changed signal be emitted?

AThousandShips (commented):
The change to the default means that users who previously selected the first tab will have no tabs selected, this might cause unexpected behaviour

doc/classes/TabBar.xml
```diff
@@ -229,8 +229,11 @@
 		<member name="clip_tabs" type="bool" setter="set_clip_tabs" getter="get_clip_tabs" default="true">
 			If [code]true[/code], tabs overflowing this node's width will be hidden, displaying two navigation buttons instead. Otherwise, this node's minimum size is updated so that all tabs are visible.
 		</member>
-		<member name="current_tab" type="int" setter="set_current_tab" getter="get_current_tab" default="0">
-			Select tab at index [code]tab_idx[/code].
+		<member name="current_tab" type="int" setter="set_current_tab" getter="get_current_tab" default="-1">
+			The index of the current selected tab. A value of [code]-1[/code] means that no tab is selected.
+		</member>
+		<member name="deselect_enabled" type="bool" setter="set_deselect_enabled" getter="get_deselect_enabled" default="false">
+			If [code]true[/code], all tabs can be deselected so that no tab is selected. Click on the [member current_tab] to deselect it.
```
KoBeWi, line 236:
```suggestion
			If [code]true[/code], all tabs can be deselected so that no tab is selected. Click on the current tab to deselect it.
```
Referencing the property is confusing here.

doc/classes/TabBar.xml
```diff
@@ -229,8 +229,11 @@
 		<member name="clip_tabs" type="bool" setter="set_clip_tabs" getter="get_clip_tabs" default="true">
 			If [code]true[/code], tabs overflowing this node's width will be hidden, displaying two navigation buttons instead. Otherwise, this node's minimum size is updated so that all tabs are visible.
 		</member>
-		<member name="current_tab" type="int" setter="set_current_tab" getter="get_current_tab" default="0">
-			Select tab at index [code]tab_idx[/code].
+		<member name="current_tab" type="int" setter="set_current_tab" getter="get_current_tab" default="-1">
+			The index of the current selected tab. A value of [code]-1[/code] means that no tab is selected.
```
KoBeWi, line 233: Should mention that -1 can be set only with `deselect_enabled` active.

scene/gui/tab_bar.cpp
```diff
@@ -640,7 +644,19 @@ int TabBar::get_tab_count() const {
 }
 
 void TabBar::set_current_tab(int p_current) {
-	ERR_FAIL_INDEX(p_current, get_tab_count());
+	if (p_current == -1) {
+		// An index of -1 is only valid if deselecting is enabled or there are no valid tabs.
+		bool no_valid_tabs = true;
+		for (int i = 0; i < get_tab_count(); i++) {
```
KoBeWi, line 650: This check is pointless when `deselect_enabled` is enabled.

scene/gui/tab_bar.cpp
```diff
@@ -640,7 +644,19 @@ int TabBar::get_tab_count() const {
...
+			if (!is_tab_disabled(i) && !is_tab_hidden(i)) {
+				no_valid_tabs = false;
+				break;
+			}
+		}
+		ERR_FAIL_COND_MSG(!deselect_enabled && !tabs.is_empty() && !no_valid_tabs, "Cannot deselect tabs, deselection is not enabled.");
```
KoBeWi, line 656:
```suggestion
		ERR_FAIL_COND_MSG(!deselect_enabled && !no_valid_tabs, "Cannot deselect tabs, deselection is not enabled.");
```
`no_valid_tabs` already handles it.

KoBeWi:
The logic for setting -1 can be improved. Pseudocode:
```
can_deselect = deselect_enabled
if (!can_deselect):
	for each tab:
		if tab.invisible or tab.disabled:
			can_deselect = true
			break

ERR_FAIL_COND_MSG(!can_deselect)
```

scene/gui/tab_bar.cpp
```diff
@@ -1520,6 +1550,9 @@ void TabBar::ensure_tab_visible(int p_idx) {
 	if (!is_inside_tree() || !buttons_visible) {
 		return;
 	}
+	if (p_idx == -1) {
```
KoBeWi, line 1553: Shouldn't this check for `deselect_enabled`?
Also might be worth mentioning it in the docs, as this is rather unexpected.

doc/classes/TabBar.xml
kitbdev: It can be set to -1 if deselect is active or if no tabs can be selected, I'll mention it.

scene/gui/tab_bar.cpp
```diff
@@ -1662,6 +1695,20 @@ bool TabBar::get_select_with_rmb() const {
 	return select_with_rmb;
 }
 
+void TabBar::set_deselect_enabled(bool p_enabled) {
+	if (deselect_enabled == p_enabled) {
+		return;
+	}
+	deselect_enabled = p_enabled;
+	if (!deselect_enabled && !tabs.is_empty()) {
+		select_next_available();
```
KoBeWi, line 1704: This should only happen if current tab is -1.

scene/gui/tab_bar.cpp
kitbdev: I don't want it to error when -1 is passed whether `deselect_enabled` is true or false, since it is a value that current_tab can be.
I'm not sure what is unexpected about it, it wouldn't do anything otherwise but error.

kitbdev:
> The logic for setting -1 can be improved. Pseudocode:

This would change the logic. All tabs must be hidden or disabled for deselection, not any.
I can add the condition and change it to a for each loop though.

scene/gui/tab_bar.cpp
KoBeWi: > since it is a value that current_tab can be.

But it can be -1 only when `deselect_enabled` or there are no tabs, right? When -1 `current_tab` results in an error, so should this method I think.

KoBeWi:
Right, I updated it.

scene/gui/tab_bar.cpp
kitbdev: Ok, I'll have to have another check for no valid tabs then.

KoBeWi: Maybe it can be a method, since you need it a few times. Something like `can_deselect()`.

kitbdev:
Fixed issue when removing all but a disabled tab would deselect that tab instead of none.
Fixed multiple tab_changed signals when moving tab.

scene/gui/tab_bar.cpp
```diff
@@ -1516,10 +1531,28 @@ void TabBar::_ensure_no_over_offset() {
...
 
+bool TabBar::_can_deselect() const {
+	if (deselect_enabled) {
+		return true;
+	}
+	// All tabs must be disabled or hidden.
+	bool can_deselect = true;
+	for (const Tab &tab : tabs) {
+		if (!tab.disabled && !tab.hidden) {
+			can_deselect = false;
```
KoBeWi, line 1542: Technically you can return `false` here and return `true` after the loop.

KoBeWi (approved):
I checked the linked issues and they aren't really reproducible, so they could be closed even without this PR I think.

kitbdev:
I added an MRP to https://github.com/godotengine/godot/issues/74653 with steps.
https://github.com/godotengine/godot/issues/73785 is a duplicate of it.

KoBeWi:
So your PR adds tab changing based on visibility 🤔
Well, can't say if it's expected/desirable, but at least it works correctly.

YeldhamDev:
When no tab is selected, the panel itself isn't draw. I don't think this should be how it works.

kitbdev:
> When no tab is selected, the panel itself isn't draw. This shouldn't be how it works.

This is intended, because I am planning to make the Editor's Bottom Dock into a TabContainer, and the Panel will need to hide.
I could add an option for it if we want?

YeldhamDev:
Then the "panel" theme style should be overridden with an empty one instead.

YuriSizov:
Why would we want to draw the panel if there is no content to display?

YeldhamDev:
- It's already the default behavior to always show the panel, even if there's no tabs available.
- It's visually displeasing to have the panel completely vanish while the tabs are still hanging in there.
- The panel can still be used to display some sort of content by the user, depending on their needs.

scene/gui/tab_container.cpp
```diff
@@ -460,6 +465,43 @@ void TabContainer::_on_active_tab_rearranged(int p_tab) {
...
+void TabContainer::_on_tab_visibility_changed(Control *p_child) {
+	if (updating_visibility) {
+		return;
+	}
+	int tab_index = get_tab_idx_from_control(p_child);
+	if (tab_index == -1) {
+		return;
+	}
+	// Only allow one tab to be visible.
+	bool made_visibile = p_child->is_visible();
```
YuriSizov, line 477: Typo :)

YuriSizov:
> * It's already the default behavior to always show the panel, even if there's no tabs available.
> * It's visually displeasing to have the panel completely vanish while the tabs are still hanging in there.

This is rather arbitrary. The default behavior doesn't change anyway, and if you enable the feature to allow closing all the tabs, then you want to be able to close all tabs, not just hide their content.

> * The panel can still be used to display some sort of content by the user, depending on their needs.

This is hard to control, so it would be limited in its usefulness. However, it got me thinking. We have `use_hidden_tabs_for_min_size`, which is not respected in this case. But it should. We should still draw the panel at least if we enable this option.

For the rest I'm not convinced it's a blocking issue.

YuriSizov (approved):
I think this is overall good, from reading the code.

But the question above is up in the air.

kitbdev:
Since I'm not using `use_hidden_tabs_for_min_size`, the panel will collapse away for me, and if it was then I could set the panel style to empty, so drawing the panel works for me.

YeldhamDev (approved):

PR #87205, opened 2024-01-15
Fix MultiMesh resource loading when using CACHE_MODE_REPLACE

### LINKED ISSUE #87166: MultiMesh3D throws errors and can corrupt scenes, Condition "instance_count > 0" is true.
### Tested versions

4.2.1

### System information

Windows 11 - Vulkan - Nvidia RTX 4070 - intel i5 13600KF

### Issue description

For some reason MultiMesh3D now throws an error like this:

image

and it can also lead to scene corruption - I don't know the conditions for that, sometimes my scenes just get corrupted when I add the multimesh, I have no idea why, it's not any particular mesh itself I tried built-in ones too.

### Steps to reproduce

With MRP:
Open the project if you don't see the error on the first try (no .godot folder yet so it doesn't show) then close the project and open it again, you should see it now or after opening the scene with the MultiMesh3D (called planet.tscn) - you might need to have both scenes open when you close the project so they reopen.
(sometimes under no idea what circumstances you can get a corrupted scene error and the scene will no longer open though I have no idea why or how to reproduce this)

Without MRP:
1. Make a scene with MultiMesh3D and populate a surface with some MeshInstance3D (for testing I used the Cylinder mesh) and save it as a scene.
2. Instantiate that scene into your main scene.
3. Run your project.
4. Close the project and open it again
5. You should see the error now or after opening the scene with the MultiMesh3D.

[MultiMeshMRP.zip](https://github.com/godotengine/godot/files/13930789/MultiMeshMRP.zip)

bitsawer:
* Fixes https://github.com/godotengine/godot/issues/87166, or at least the loading error part. I couldn't repro the scene corruption, so this might either fix that or it is an unrelated issue caused by something else.

MultiMesh is a bit unusual resource as the properties must updated in specific order for things to work. That is, property `instance_count` must be updated after some others. This is [documented](https://docs.godotengine.org/en/latest/classes/class_multimesh.html#class-multimesh-property-instance-count), but a bit clumsy. This calling order applies when loading a saved MultiMesh resource, too. Things work most of the time, but when the editor loads the resource using `ResourceFormatLoader::CACHE_MODE_REPLACE` the error checks in some property updates get in the way because the property setter calls are now coming in "wrong" order because the previous load already updated some properties. This can result in an incorrectly configured MultiMesh.

This PR simply removes the error macros from some property setters. Things work perfectly fine without them, although the error messages can give some useful hints to check the class documentation when accessing the API programmatically. For the editor UI, I added `_validate_property()` to disable some fields in the editor if `instance_count` is not 0 (previously, if you tried to edit these properties the error macros triggered, now the fields are disabled). The overall behavior of MultiMesh could be improved in the future, but this fixes the loading issue.

### DIFF
--- a/scene/resources/multimesh.cpp
+++ b/scene/resources/multimesh.cpp
@@ -219,6 +219,8 @@ void MultiMesh::set_instance_count(int p_count) {
 	ERR_FAIL_COND(p_count < 0);
 	RenderingServer::get_singleton()->multimesh_allocate_data(multimesh, p_count, RS::MultimeshTransformFormat(transform_format), use_colors, use_custom_data);
 	instance_count = p_count;
+
+	notify_property_list_changed();
 }
 
 int MultiMesh::get_instance_count() const {
@@ -278,7 +280,6 @@ RID MultiMesh::get_rid() const {
 }
 
 void MultiMesh::set_use_colors(bool p_enable) {
-	ERR_FAIL_COND(instance_count > 0);
 	use_colors = p_enable;
 }
 
@@ -287,7 +288,6 @@ bool MultiMesh::is_using_colors() const {
 }
 
 void MultiMesh::set_use_custom_data(bool p_enable) {
-	ERR_FAIL_COND(instance_count > 0);
 	use_custom_data = p_enable;
 }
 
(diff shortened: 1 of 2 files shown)


KoBeWi:
I think the reason why there was an error when setting these properties with existing meshes is that it won't cause the existing meshes to update, which without a clear error will appear as if setting the property had no effect.

A better fix would be probably adding a method like `_queue_mesh_update()`, which would update meshes at the end of the current frame, so that it doesn't matter in which order you set the properties.

That said, this might also point at some faulty logic in CACHE_MODE_REPLACE. Looks like it loads the resource anew, but it reuses already loaded sub-resources, which I think is unexpected.

PR #87207, opened 2024-01-15
Add missing check when insert function `normal_roughness_compatibility`

### LINKED ISSUE #87202: Shader compilation error when calling `textureSize` on the normal roughness texture
### Tested versions

- Reproduced with Godot v4.3.dev2.official [352434668]
- Cannot reproduce with v4.2.1.stable.official [b09f793f5]

### System information

Godot v4.3.dev2 - Pop!_OS 22.04 LTS - X11 - Vulkan (Forward+) - integrated Intel(R) UHD Graphics 620 (KBL GT2) () - Intel(R) Core(TM) i5-8350U CPU @ 1.70GHz (4 Threads)

### Issue description

When creating a shader of type `spatial`, calling `textureSize` on the normal roughness texture will cause a shader error.

An example might be:
```glsl
shader_type spatial;

uniform sampler2D normal_roughness_buffer: hint_normal_roughness_texture;

void fragment() {
	ivec2 normal_roughness_size = textureSize(normal_roughness_buffer, 0);
	ALBEDO = vec3(1.0, 0.0, 0.0);
}
```

image

The error output is quite long (>10000 lines) but, the initial text indicates:
```
servers/rendering/renderer_rd/shader_rd.cpp:470 - Condition "f.is_null()" is true.
  (3) servers/rendering/renderer_rd/shader_rd.cpp:483 - Parameter "p_version->variants" is null.
  servers/rendering/renderer_rd/forward_clustered/scene_shader_forward_clustered.cpp:181 - Condition "!shader_singleton->shader.version_is_valid(version)" is true.
  servers/rendering/renderer_rd/shader_rd.cpp:293 - Error compiling Fragment shader, variant #0 (
  #define MODE_RENDER_DEPTH
  ).
  servers/rendering/renderer_rd/shader_rd.cpp:294 - Failed parse:
  WARNING: 0:73: '' : all default precisions are highp; use precision statements to quiet warning, e.g.:
           "precision mediump int; precision highp float;" 
  ERROR: 0:2210: 'normal_roughness_compatibility' : no matching overloaded function found 
  ERROR: 0:2210: '=' :  cannot convert from ' const float' to ' temp highp 2-component vector of int'
  ERROR: 0:2210: '' : compilation terminated 
  ERROR: 3 compilation errors.  No code generated.
```

### Steps to reproduce

1. Create an empty 3D scene
2. Create a `MeshInstance` with a `ShaderMaterial`
3. In the mesh's material's shader, call `textureSize` on the normal roughness texture
4. Note the errors compiling shaders

[normal_roughness.zip](https://github.com/godotengine/godot/files/13933891/normal_roughness.zip)

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/87202

Without this check, the shader `ivec2 normal_roughness_size = textureSize(normal_roughness_buffer, 0);` compiles to 
`ivec2 m_normal_roughness_size=normal_roughness_compatibility(textureSize(sampler2D(normal_roughness_buffer, SAMPLER_LINEAR_WITH_MIPMAPS_REPEAT), 0));` and thus type mismatch and compile error.

Introduced in https://github.com/godotengine/godot/pull/86316/ , cc @clayjohn Any suggestions?

### DIFF
--- a/servers/rendering/shader_compiler.cpp
+++ b/servers/rendering/shader_compiler.cpp
@@ -1340,7 +1340,7 @@ String ShaderCompiler::_dump_node_code(const SL::Node *p_node, int p_level, Gene
 					if (is_screen_texture && !texture_func_returns_data && actions.apply_luminance_multiplier) {
 						code = "(" + code + " * vec4(vec3(sc_luminance_multiplier), 1.0))";
 					}
-					if (is_normal_roughness_texture) {
+					if (is_normal_roughness_texture && !texture_func_returns_data) {
 						code = "normal_roughness_compatibility(" + code + ")";
 					}
 				} break;


danbolt:
Hi, thanks for the quick turnaround!

I just compiled Godot at 26b1fd0d8 with this change and it addressed the issue I was encountering with #87202. :+1:

clayjohn (approved):
Good catch!

This is the correct fix. Great work

akien-mga:
Thanks!

Issue #79336, opened 2023-07-11
Window.size_changed and NOTIFICATION_WM_SIZE_CHANGED are not triggered when "window.content_scale_mode == Window.ContentScaleMode.CONTENT_SCALE_MODE_VIEWPORT"


Gnumaru:
### Godot version

v4.1.stable.official [970459615]

### System information

Godot v4.1.stable - Windows 10.0.19044 - Vulkan (Compatibility) - NVIDIA GeForce GTX 1050 Ti (NVIDIA; 31.0.15.2647) - Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz (8 Threads)

### Issue description

When using a ContentScaleMode equal to CONTENT_SCALE_MODE_VIEWPORT, the size_changed signal of the root window is not emited regardless of changing the size via code or resizing the window manually. not only the signal is not emited, but the notification is also not triggered. Look at output log:

![image](https://github.com/godotengine/godot/assets/5783414/682fe574-2012-449a-ad88-48d468cd201b)

### Steps to reproduce

On a new project, on a new scene, create a script in the scene root, paste this script and run it
```gdscript
extends Node2D
var mold_size:Vector2i
func _ready()->void:
	print('==================== BEGIN')
	var vwindow:Window = get_window()
	print('_ready(): connecting size_changed to on_size_changed')
	vwindow.size_changed.connect(on_size_changed)
	print('\n_ready(): changing scale to CONTENT_SCALE_MODE_DISABLED')
	vwindow.content_scale_mode = Window.ContentScaleMode.CONTENT_SCALE_MODE_DISABLED
	print('_ready(): changing to 1')
	vwindow.size = Vector2i.ONE
	print('_ready(): changing to 128')
	vwindow.size = Vector2i.ONE * 128
	print('\n_ready(): changing scale to CONTENT_SCALE_MODE_CANVAS_ITEMS')
	vwindow.content_scale_mode = Window.ContentScaleMode.CONTENT_SCALE_MODE_CANVAS_ITEMS
	print('_ready(): changing to 1')
	vwindow.size = Vector2i.ONE
	print('_ready(): changing to 128')
	vwindow.size = Vector2i.ONE * 128
	print('\n_ready(): changing scale to CONTENT_SCALE_MODE_VIEWPORT')
	vwindow.content_scale_mode = Window.ContentScaleMode.CONTENT_SCALE_MODE_VIEWPORT
	print('_ready(): changing to 1')
[...]

PR #87210, opened 2024-01-15
Add `window_size_changed` signal

### LINKED ISSUE #8788: Make `NOTIFICATION_WM_SIZE_CHANGED´ available to child nodes of the window

Godot

Accessing the notification, when the size of the root window changes is very unintuitive.
This problem has been discussed in the bug report https://github.com/godotengine/godot/issues/79336 which was solved by improving the documentation. However the unintuitive way could be improved.

The problem itself can be solved by making `NOTIFICATION_WM_SIZE_CHANGED` available to child nodes of the window. That allows users to easily access when the root window is changed.
Since this change could have a negative performance impact, the performance impact needs to be evaluated beforehand.

Implementation is likely straightforward.
Whenever a `Window` receives a `NOTIFICATION_WM_SIZE_CHANGED`, it automatically sends the notification recursively to child nodes with the exception of other `Window`-nodes.

In order to access the notifcation in the root window, one has to add a GDScript to the root node, that listens to notifications. A demo project of how to do that is available in https://github.com/godotengine/godot/issues/79336#issuecomment-1631627181.

This is a core behavior

kubasobon:
Towards https://github.com/godotengine/godot-proposals/issues/8788

This PR adds a new signal to the `Window` class called `window_size_changed`. The signal is emitted whenever the window size is changed, either in code or using UI.

The PR is inspired by https://github.com/godotengine/godot/issues/79336 and my issues with `Viewport.size_changed` signal while using either `Window.ContentScaleMode.CONTENT_SCALE_MODE_VIEWPORT` or `Window.ContentScaleMode.CONTENT_SCALE_MODE_CANVAS_ITEMS`. The `Viewport.size_changed` signal, which `Window` class inherits, is not emitted with the scaling mode set. It works as expected when scaling is disabled. This is not indicated in the engine documentation. There is no alternative to the signal, leaving users with only the `NOTIFICATION_WM_SIZE_CHANGED` to subscribe to, and in a pretty obscure way as showcased in <https://github.com/godotengine/godot/issues/79336#issuecomment-1631627181> (patching the root node script).

To better explain:

| enum | `ContentScaleMode` | `Window.size_changed` emitted | `Window.window_size_changed` emitted - added by this PR |
|------|--------------------|-------------------------------|---|
|    0 | CONTENT_SCALE_MODE_DISABLED | yes 💚  | yes 💚  |
|    1 | CONTENT_SCALE_MODE_CANVAS_ITEMS | no ❌   | yes 💚  |
|    2 | CONTENT_SCALE_MODE_VIEWPORT | no ❌ |  yes 💚  |

*Bugsquad edit:*
- Closes https://github.com/godotengine/godot-proposals/issues/8788

### DIFF
--- a/doc/classes/Window.xml
+++ b/doc/classes/Window.xml
@@ -765,6 +765,11 @@
 			<description>
 				Emitted when the [Window] is currently focused and receives any input, passing the received event as an argument. The event's position, if present, is in the embedder's coordinate system.
 			</description>
+		<signal name="window_size_changed">
+			<description>
+				Emitted when [Window] size is changed.
+			</description>
+		</signal>
 		</signal>
 	</signals>
 	<constants>
--- a/doc/classes/Window.xml
+++ b/doc/classes/Window.xml
@@ -765,11 +765,11 @@
 			<description>
 				Emitted when the [Window] is currently focused and receives any input, passing the received event as an argument. The event's position, if present, is in the embedder's coordinate system.
 			</description>
+		</signal>
 		<signal name="window_size_changed">
 			<description>
 				Emitted when [Window] size is changed.
 			</description>
-		</signal>
 		</signal>
 	</signals>
 	<constants>
--- a/scene/main/window.cpp
+++ b/scene/main/window.cpp
@@ -1163,6 +1163,7 @@ void Window::_update_viewport_size() {
 	if (old_size != size) {
 		old_size = size;
 		notification(NOTIFICATION_WM_SIZE_CHANGED);
+		emit_signal(SNAME("window_size_changed"));
 	}
 
 	if (embedder) {
@@ -2926,6 +2927,7 @@ void Window::_bind_methods() {
 	ADD_SIGNAL(MethodInfo("theme_changed"));
 	ADD_SIGNAL(MethodInfo("dpi_changed"));
 	ADD_SIGNAL(MethodInfo("titlebar_changed"));
+	ADD_SIGNAL(MethodInfo("window_size_changed"));
 
 	BIND_CONSTANT(NOTIFICATION_VISIBILITY_CHANGED);
 	BIND_CONSTANT(NOTIFICATION_THEME_CHANGED);


kubasobon:
cc: @Gnumaru This PR aims at solving your original complaint in https://github.com/godotengine/godot/issues/79336. I'd love for you to weigh in.

Gnumaru:
> cc: @Gnumaru This PR aims at solving your original complaint in #79336. I'd love for you to weigh in.

Hi.

Regarding performance, I have absolutely no idea what is more performant: many (say, a thousand) nodes connecting to a signal or the same nodes nodes implementing _notification in gdscript just to receive a size changed notification. I have a hunch that implementing _notification performs the poorest since _notification is called for a lot of things and a size changed signal would be called only when changing the window size.

Regarding usability, it is undoubtedly better to listen to signals than implementing _notification and matching an int to a constant

Regarding names I think window_size_changed is better but please read to the end. I think the original size changed signal (and name) name is already misleading. In my own personal preference, what I reeeally would like is that we had two separate signals with specific purpose and names for 'real' and 'virtual' sizes, I mean, the actual window size as the operating system sees it and the underlying viewport size as godot renders it.

My original real life problem when I opened https://github.com/godotengine/godot/issues/79336 is that I wanted to force the window size (the actual window, handled by the OS, not the underlying viewport) to an integer multiple of my game's base resolution. Nowadays godot have  Window.ContentScaleStretch.CONTENT_SCALE_STRETCH_INTEGER but it is still not exactly what I would like since the actual window still can have a variable size, only the underlying viewport gets rendered to the integer multiple, centered in the window and surrounded with black bars.

If we had a signal specific for notifying the value change of the actual window size (named window_size_changed while the other could be named viewport_size_changed) then I could simply connect to window_size_changed, check if the window is not maximized neither full screen (in which case I would leave as is) and, if the window size is not a multiple of my base resolution, I would resize the window again to the nearest multiple, like I'm already doing, but I'm attaching a script to the root window just for that, and I think that attaching scripts for engine objects like the root window or the scene tree is something that should really not be needed except for the most exceptional cases.

AThousandShips:
Signals are not free and unlike notifications they are sent no matter if anyone listens to them, or rather the overhead doesn't go away, especially with a duplicate and potentially confusing signal

kubasobon:
@AThousandShips thank you for explaining this. My main goal for this PR is to solve the case where `size_changed` signal does not get emitted if any kind of Content Scaling is enabled. Currently there is no other way of knowing the window has been resized, than subscribing to the notification. This might be confusing to other users, as it was for me, since `size_changed` signal just stops working.

If there is a better way of getting notified every time a window is resized, please let me know.

Gnumaru:
> Renaming `Viewport.size_changed` to `Viewport.viewport_size_changed` would make it clearer, but constitutes a **big** breaking change IMHO. Would improving documentation and elaborating on when to use which signal help?

Indeed, renaming the size_changed signal is too big of a compat breaking change to happen out of the blue. Is it possible to create a "signal alias" or something like that? or maybe having two redundant signals (size_changed and viewport_size_changed) and mark size_changed as deprecated in the docs, saying that it will be removed in favor of viewport_size_changed in a future (but yet undetermined) version?

The docs could be further improved but I think that if the cost is not high (both performance cost and man-hour coding cost), the best would be to have signals (or notifications) that handle the specific separate needs of knowing if the actual operating system window size changed OR the underlying viewport size changed (which would never happen when the scale mode is viewport unless it is possible to change it via script).

Please note that I'm not asking for increasing the scope of this PR. I'm just throwing ideas. Having another signal (like the one in this PR) or having another notification that gets triggered in any condition is already the most  beneficial change to the current state of things, regardless if the name of the signal is good or not. But I think it is good name since it would allow a future refactoring (godot 5?) renaming size_changed to viewport_size_changed and NOTIFICATION_WM_SIZE_CHANGED  to NOTIFICATION_VIEWPORT_SIZE_CHANGED

kubasobon:
@Gnumaru thank you for your remarks. 

My main concern and vision for the scope of the PR is either fixing `Window.size_changed` with content scaling enabled (difficult, since it is inherited from Viewport, which in fact does not change size with scaling) or introducing a solution which enables users to track window size changes with similar effort/complexity - e.g. a signal.

wgetJane:
i doubt that there is a valid performance concern with emitting one signal every time a window is resized (which happens very occasionally anyway)

Gnumaru:
> i doubt that there is a valid performance concern with emitting one signal every time a window is resized (which happens very occasionally anyway)

Depending on the operating system and the window manager, while you resize a window by dragging the corners, the resize signal could be emitted every frame. But I do admit that very seldomly someone would be resizing a game window by dragging the borders

Gnumaru:
Indeed. Both the notification and signal should either be triggered every frame or one single time when the user let go the mouse click. On windows (since windows XP up to windows 11), there is a performance setting whose name I cannot remember right now that disables or enables drawing the window content when dragging or resizing the window. If this is set to render the window content (which is the default) then the notification and signal should trigger every frame while clicks and hold the window border in order to resize the window, or at least while holding the mouse button down and moving the mouse (thus actually resizing the window).

PR #87206, opened 2024-01-15
Fix DX12 build on older MSVC versions, fix missing PIX include.

bruvzg:
Fixes https://www.reddit.com/r/godot/comments/1949mh4/comment/khfagrg/?utm_source=reddit&utm_medium=web2x&context=3

PR #87212, opened 2024-01-15
Fix build failing with PIX enabled

RandomShaper:
After the refactor to RDD, the driver needs to see the PIX headers (in addition to the contex).

@bruvzg, any MinGW implications?

### DIFF
--- a/drivers/d3d12/d3d12_context.cpp
+++ b/drivers/d3d12/d3d12_context.cpp
@@ -65,17 +65,6 @@ extern "C" {
 char godot_nir_arch_name[32];
 }
 
-#ifdef PIX_ENABLED
-#if defined(__GNUC__)
-#define _MSC_VER 1800
-#endif
-#define USE_PIX
-#include "WinPixEventRuntime/pix3.h"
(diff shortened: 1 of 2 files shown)


bruvzg:
Isn't it the same as adding the same include to the driver `.cpp`, which is added in https://github.com/godotengine/godot/pull/87206

But if it's needed for something else, it should be OK to move it the header.

RandomShaper:
It is. It was just to avoid duplication, but duplication has its own advantages. I don't have a strong preference.

akien-mga:
I'd say if we expect this might be used in more than two files in the future, or in all files that use `d3d12_context.h`, it makes sense to include in the header. Otherwise if it's going to stay limited to 2 files, it should be fine with #8720 only.

RandomShaper:
Oh, for some reason I forgot #87206 already had the fix, even if I had just checked it. Duplicating is fine to me given it's only two cases and avoids cluttering the headers. Also, ongoing work on the rendering architecture, by @DarioSamo, may allow that only the RDD needs to include it eventually.

Closing as superseded by #87206.

PR #87215, opened 2024-01-15
Fix Atlas regions when using size limit

### LINKED ISSUE #82781: Atlas regions aren't adapted after Size Limit is set on atlas Texture2D
### Godot version

4.1.2.stable.mono (also 4.2.dev-6.mono)

### System information

Godot v4.1.2.stable.mono - macOS 14.0.0 - Vulkan (Mobile) - integrated Apple M1 - Apple M1 (8 Threads)

### Issue description

After creating an atlas and reimporting it I noticed the file was over 5k in height. As the target are android devices I tried limiting the size of the atlas to 4096 pixels. After reimporting the atlas `Texture2D` was correctly scaled, but the regions generated for sprites seem to be still targeting the not limited atlas texture.

You can see the results in the screenshots below. First one shows the original, not limited, atlas with correctly displayed `Sprite2D`s. Second one shows the atlas with `Size Limit` set and broken sprites.

image

image

### Steps to reproduce

1. Create atlas containing sprites
2. Add `Sprite2D`s using textures used to create the atlas to the scene
3. Try limiting size of the atlas using the Import settings (`Size Limit`) 

N/A

onurtnrkl:
Fixes #82781


AThousandShips:
See also:
* https://github.com/godotengine/godot/pull/87145

editor/import/resource_importer_texture_atlas.cpp
```diff
@@ -276,9 +277,17 @@ Error ResourceImporterTextureAtlas::import_group_file(const String &p_group_file
...
+	group_file_config.instantiate();
+	Error err = group_file_config->load(p_group_file + ".import");
+	ERR_FAIL_COND_V(err != OK, ERR_FILE_CORRUPT);
+	ERR_FAIL_COND_V(!group_file_config->has_section_key("params", "process/size_limit"), ERR_FILE_CORRUPT);
+	const int atlas_size_limit = group_file_config->get_value("params", "process/size_limit");
+
 	//pack the charts
 	int atlas_width, atlas_height;
-	EditorAtlasPacker::chart_pack(charts, atlas_width, atlas_height);
+	EditorAtlasPacker::chart_pack(charts, atlas_width, atlas_height, atlas_size_limit);
```
AThousandShips, line 290: This breaks if `size_limit` is `0`, which is the default, which means no size limit

editor/import/resource_importer_texture_atlas.cpp
```diff
@@ -276,9 +277,17 @@ Error ResourceImporterTextureAtlas::import_group_file(const String &p_group_file
 		idx++;
 	}
 
+	//get atlas size limit
+	Ref<ConfigFile> group_file_config;
+	group_file_config.instantiate();
+	Error err = group_file_config->load(p_group_file + ".import");
```
AThousandShips, line 283: Is this file guaranteed to exist *before* first import?

AThousandShips (dismissed):

AThousandShips:
The size limit isn't really designed for these files, it's for *scaling* the resulting image, it should probably be disabled for these atlas files as it doesn't work with them correctly, I'm not sure this is a bug

editor/import/resource_importer_texture_atlas.cpp
AThousandShips: That's most likely a bug due to the implementation not expecting it, what was the resulting size?

AThousandShips: Please try with that MRP and see what the results are and how safe it becomes

onurtnrkl: I got it know. Thanks for the explanation.

editor/import/resource_importer_texture_atlas.cpp
AThousandShips: It is not, if you don't have the file already imported it will not import it and fails with a lot of errors, now the file can't ever be imported again until you manually add the import file

AThousandShips:
Thank you for your work but this needs more work and some discussion and evaluation on the way to go forward

I don't think this needs to be changed at all, the size limit doesn't work as this issue expects it to, and this change, even with the default value and other problems being fixed, is in my opinion only likely to introduce further confusion in users

AThousandShips:
The better solution would probably be to fetch the size limit and then scale based on the result size, but that's quite complicated, I'll see if I can write some tests later today to help with that

It would depend on the specific way the size limit is implemented though which risks making this code break if that code changes

AThousandShips:
Can't test much myself at the moment but try applying this change, on top of `master` without your other changes:
```diff
diff --git a/editor/import/resource_importer_texture_atlas.cpp b/editor/import/resource_importer_texture_atlas.cpp
index d437f23740..7a2f7101d2 100644
--- a/editor/import/resource_importer_texture_atlas.cpp
+++ b/editor/import/resource_importer_texture_atlas.cpp
@@ -313,6 +313,24 @@ Error ResourceImporterTextureAtlas::import_group_file(const String &p_group_file
                cache = res_cache;
        }

+       float size_scale = 1.0f;
+
+       if (FileAccess::exists(p_group_file + ".import")) {
+               Ref<ConfigFile> group_file_config;
+               group_file_config.instantiate();
+               Error err = group_file_config->load(p_group_file + ".import");
+               if (err = OK) {
+                       const int atlas_size_limit = group_file_config->get_value("params", "process/size_limit");
+                       if (atlas_size_limit > 0 && (atlas_width > atlas_size_limit || atlas_height > atlas_size_limit)) {
+                               if (atlas_width > atlas_height) {
+                                       size_scale = atlas_width / (float)atlas_size_limit;
+                               } else {
+                                       size_scale = atlas_height / (float)atlas_size_limit;
+                               }
+                       }
+               }
+       }
+
        //save the images
        idx = 0;
        for (const KeyValue<String, HashMap<StringName, Variant>> &E : p_source_file_options) {
@@ -327,10 +345,10 @@ Error ResourceImporterTextureAtlas::import_group_file(const String &p_group_file
                        Ref<AtlasTexture> atlas_texture;
                        atlas_texture.instantiate();
                        atlas_texture->set_atlas(cache);
-                       atlas_texture->set_region(Rect2(offset, pack_data.region.size));
+                       atlas_texture->set_region(Rect2(offset, pack_data.region.size) * size_scale);

                        if (!pack_data.is_cropped) {
-                               atlas_texture->set_margin(Rect2(pack_data.region.position, pack_data.image->get_size() - pack_data.region.size));
+                               atlas_texture->set_margin(Rect2(pack_data.region.position, pack_data.image->get_size() - pack_data.region.size) * size_scale);
                        }

                        texture = atlas_texture;
@@ -356,7 +374,7 @@ Error ResourceImporterTextureAtlas::import_group_file(const String &p_group_file

                                        for (int j = 0; j < vc; j++) {
                                                vw[j] = chart.vertices[j];
-                                               Vector2 uv = chart.vertices[j];
+                                               Vector2 uv = chart.vertices[j] * size_scale;
                                                if (chart.transposed) {
                                                        SWAP(uv.x, uv.y);
                                                }
```

AThousandShips:
Tested it a bit and it works, partially, can do some further testing later but you try it out too to verify

AThousandShips:
You accidentally wiped your branch resetting it, you need to push again with your changes applied

onurtnrkl:
> Tested it a bit and it works, partially, can do some further testing later but you try it out too to verify
> 
> An immediately obvious problem however that makes any "fix" to this likely to fail anyway is that `AtlasTexture` can't be scaled after the fact, so the scaled down textures won't work correctly in the end, this makes me feel this should be solved by an error, and or a check to fail if the size limit is set
> 
> Edit: Did some more revisions and the above code now works reasonably well as far as I can tell, needs some further testing though, and some decisions

Region positions works but this change broke my tile system. I should do same scale calculation in my tile renderer in this case.
Screenshot 2024-01-15 at 20 30 28

AThousandShips:
Please don't update with merges, push with `git push --force`

onurtnrkl:
> > Tested it a bit and it works, partially, can do some further testing later but you try it out too to verify
> > An immediately obvious problem however that makes any "fix" to this likely to fail anyway is that `AtlasTexture` can't be scaled after the fact, so the scaled down textures won't work correctly in the end, this makes me feel this should be solved by an error, and or a check to fail if the size limit is set
> > Edit: Did some more revisions and the above code now works reasonably well as far as I can tell, needs some further testing though, and some decisions
> 
> Region positions works but this change broke my tile system. I should do same scale calculation in my tile renderer in this case. Screenshot 2024-01-15 at 20 30 28

Maybe `scale_size` property can be added to `Texture2D` for similar cases in the future.

AThousandShips:
See:
* https://github.com/godotengine/godot-proposals/discussions/8660

Issue #87196, opened 2024-01-15
Enabling the compatibility renderer leaks `Image` objects when using the Editor


Rubonnek:
### Tested versions

26b1fd0d842fa3c2f090ead47e8ea7cd2d6515e1 - reproducible
4.2.1 - Not reproducible

### System information

Godot v4.3.dev (e3b9f08ce) - Arch Linux #1 SMP PREEMPT_DYNAMIC Fri, 05 Jan 2024 16:20:41 +0000 - Tty - GLES3 (Compatibility) - AMD Radeon RX 6900 XT (radeonsi, navi21, LLVM 16.0.6, DRM 3.54, 6.6.10-arch1-1) () - AMD Ryzen 9 5950X 16-Core Processor (32 Threads)

### Issue description

Enabling the compatibility renderer will leak `Image` objects. This issue does not happen with the `Forward+` nor the `Mobile` rendering methods.

I think this issue was introduced within the last two months.

### Steps to reproduce

Open the Editor against the MRP through the command line (i.e. for example `godot -e --verbose` if on path).

Close the editor after a few seconds and you should see a similar message:

```
WARNING: ObjectDB instances leaked at exit (run with --verbose for details).
     at: cleanup (core/object/object.cpp:2204)
Leaked instance: Image:-9170885200035390510 - Resource path: 
Leaked instance: Image:-9170880252233060709 - Resource path: 
Leaked instance: Image:-9170864309314457951 - Resource path: 
Leaked instance: Image:-9170863759558644050 - Resource path: 
Leaked instance: Image:-9170858811756319055 - Resource path: 
Leaked instance: Image:-9170858262000505166 - Resource path: 
Leaked instance: Image:-9170853863953994059 - Resource path: 
Leaked instance: Image:-9170853314198180170 - Resource path: 
Leaked instance: Image:-9170848366395855175 - Resource path: 
Leaked instance: Image:-9170847816640041286 - Resource path: 
```

### Minimal reproduction project (MRP)

[glow.zip](https://github.com/godotengine/godot/files/13933608/glow.zip) - from the [demo projects](https://github.com/godotengine/godot-demo-projects/tree/4a4b46cbe1f5105b664b99f325984096151b75a1/2d/glow) with the Compatibility renderer enabled by default.

Rubonnek:
The last commit where this issue wasn't happening is 9444d297ed0b1dbc7c05fa0bf2e06241335f5057.

Issue introduced by #86587.

CC @RandomShaper

PR #87222, opened 2024-01-15
Fix inaccuracies in the report of leaked objects

RandomShaper:
Discovered while investigating #87196.

The reported object id was being reported wrongly, for a couple of reasons:
1. The id was being computed wrongly. We could get the id from the object, but I've decided to keep the computation, only fixed.
2. The id was printed as a signed integer, which was wrong for reference types, which have the MSB set.


akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #87223, opened 2024-01-15
Replace some "uncommon" words in class reference

Mickeon:
Over time some very "exotic" words have seeped through the class reference. This is not very good because the assumption is that non-native English speakers should be able to read the documentation, as well.

When it can be done, it's better to be consistent. Broad terms and less common words relative to the rest of the class reference should be avoided.

This PR may seem big, but don't be fooled. Most of these words can only be found twice at most across the class reference, anyway. Some typos on the way are fixed, as well.

The most notable replacements include:
- "Untouched" -> "Unaffected", "Unchanged";
- "standalone" -> "independent";
- "Fine" -> "Good", "Precise" (depending on context);
- "Notwithstanding" -> "Regardless of"; <sub>what the heck, guys? I had to look up this one, too!</sub>
- "Disregards" -> "Ignores";
- "Neglects to" -> "Fails to";


doc/classes/Array.xml
```diff
@@ -479,7 +479,7 @@
 			<return type="Variant" />
 			<param index="0" name="position" type="int" />
 			<description>
-				Removes and returns the element of the array at index [param position]. If negative, [param position] is considered relative to the end of the array. Leaves the array untouched and returns [code]null[/code] if the array is empty or if it's accessed out of bounds. An error message is printed when the array is accessed out of bounds, but not when the array is empty.
+				Removes and returns the element of the array at index [param position]. If negative, [param position] is considered relative to the end of the array. Leaves the array unaffected and returns [code]null[/code] if the array is empty or if it's accessed out of bounds. An error message is printed when the array is accessed out of bounds, but not when the array is empty.
```
AThousandShips, line 482:
```suggestion
				Removes and returns the element of the array at index [param position]. If negative, [param position] is considered relative to the end of the array. Leaves the array unchanged and returns [code]null[/code] if the array is empty or if it's accessed out of bounds. An error message is printed when the array is accessed out of bounds, but not when the array is empty.
```

AThousandShips (approved):
Looks good otherwise

doc/classes/Callable.xml
```diff
@@ -4,7 +4,7 @@
 		A built-in type representing a method or a standalone function.
 	</brief_description>
 	<description>
-		[Callable] is a built-in [Variant] type that represents a function. It can either be a method within an [Object] instance, or a standalone function not related to any object, like a lambda function. Like all [Variant] types, it can be stored in variables and passed to other functions. It is most commonly used for signal callbacks.
+		[Callable] is a built-in [Variant] type that represents a function. It can either be a method within an [Object] instance, or an independent function not related to any object, like a lambda function. Like all [Variant] types, it can be stored in variables and passed to other functions. It is most commonly used for signal callbacks.
```
YuriSizov, line 7:
```suggestion
		[Callable] is a built-in [Variant] type that represents a function. It can either be a method within an [Object] instance, or an independent function not related to any object, like a lambda function. Like all [Variant] types, it can be stored in variables and passed to other functions. It is most commonly used for signal callbacks.
```

An independent function doesn't sound like anything to me. IMO "standalone" is a valid term for this problem space and should not be avoided. Although I agree that in the example of "STATE_MACHINE_TYPE_GROUPED" its used out of place.

doc/classes/GeometryInstance3D.xml
```diff
@@ -34,7 +34,7 @@
...
+			Overrides the bounding box of this node with a custom one. This can be used to avoid the expensive [AABB] recalculation that happens when a skeleton is used with a [MeshInstance3D] or to have precise control over the [MeshInstance3D]'s bounding box. To use the default AABB, set value to an [AABB] with all fields set to [code]0.0[/code]. To avoid frustum culling, set [member custom_aabb] to a very large AABB that covers your entire game world such as [code]AABB(-10000, -10000, -10000, 20000, 20000, 20000)[/code]. To disable all forms of culling (including occlusion culling), call [method RenderingServer.instance_set_ignore_culling] on the [GeometryInstance3D]'s [RID].
```
YuriSizov, line 37: This change isn't needed, but it doesn't hurt.

doc/classes/ProjectSettings.xml
```diff
@@ -2360,7 +2360,7 @@
 			[b]Note:[/b] This property is only read when the project starts. To set the subsurface scattering scale at runtime, call [method RenderingServer.sub_surface_scattering_set_scale] instead.
 		</member>
 		<member name="rendering/environment/volumetric_fog/use_filter" type="int" setter="" getter="" default="1">
-			Enables filtering of the volumetric fog effect prior to integration. This substantially blurs the fog which reduces fine details but also smooths out harsh edges and aliasing artifacts. Disable when more detail is required.
+			Enables filtering of the volumetric fog effect prior to integration. This substantially blurs the fog which reduces sharper details but also smooths out harsh edges and aliasing artifacts. Disable when more detail is required.
```
YuriSizov, line 2363: "fine details" is a common phrase. "sharper details" is not.

YuriSizov:
Other than that, seems like a nice cleanup!

doc/classes/GeometryInstance3D.xml
Mickeon: To be a little more transparent on my thought process, I looked up every single reference of "fine" and realised it is extremely rarely used, and each one of them has a different meaning. If the only remaining mentions of "fine" are... as in "fine details", that's good. At least the meaning is consistent.

Mickeon:
Thank you @AThousandShips and @YuriSizov for taking your time looking at this!

doc/classes/GeometryInstance3D.xml
YuriSizov: I mean, it's a generic word, it has many meanings and is a part of many established phrases too 🙃

doc/classes/ProjectSettings.xml
```diff
@@ -2210,7 +2210,7 @@
...
 		</member>
 		<member name="physics/common/physics_jitter_fix" type="float" setter="" getter="" default="0.5">
-			Controls how much physics ticks are synchronized with real time. For 0 or less, the ticks are synchronized. Such values are recommended for network games, where clock synchronization matters. Higher values cause higher deviation of in-game clock and real clock, but allows smoothing out framerate jitters. The default value of 0.5 should be fine for most; values above 2 could cause the game to react to dropped frames with a noticeable delay and are not recommended.
+			Controls how much physics ticks are synchronized with real time. For 0 or less, the ticks are synchronized. Such values are recommended for network games, where clock synchronization matters. Higher values cause higher deviation of in-game clock and real clock, but allows smoothing out framerate jitters. The default value of 0.5 should be good for most; values above 2 could cause the game to react to dropped frames with a noticeable delay and are not recommended.
```
YuriSizov, line 2213: Here it should also be "good enough", btw :)

YuriSizov (approved):

akien-mga:
Thanks!

PR #77004, opened 2023-05-12
Use WorkerThreadPool for Server threads

reduz:
* Servers now use WorkerThreadPool for background computation.
* This helps keep the number of threads used fixed at all times.
* It also ensures everything works on HTML5 with threads.
* And makes it easier to support disabling threads for also HTML5.

The only downside of this is that GLES3 threads likely no longer work because every time the render thread is called, the thread context is different. This needs to be researched how to fix. Maybe if GLES3 is used, threaded rendering server should be disabled for the time being?

PR #87229, opened 2024-01-15
Run resource previewer on the main thread if using GL compatibility

### LINKED ISSUE #87196: Enabling the compatibility renderer leaks `Image` objects when using the Editor

RandomShaper:
I don't know why the issue was surfaced by #86587, but to my understanding, it was already there previously. The issue is that with OpenGL (Compatibility rendering method)  you can't deal with textures from arbitrary threads. The rendering server tries to do the right thing by sending the tasks to a queue, but the resource importer will blindly continue doing its job without waiting for them to be created. The worst consequence is that, given the relevant texture-related functions take references to resource references (`Ref<>&`), by the time the previewer thinks it's done with some images it can decrease their refcount to zero. Because of that, there are dangling references by the time the queue is flushed later.

This PR fixes the ill situations by just letting the main thread flush any outstanding resource importing tasks every idle frame. That ensures that any previewer, either one that directly deals with rendering or one that doesn't, because in the end all of them have to create the preview images, don't incurr in sync issues.

This will affect the responsiveness of the editor while there are queued preview tasks. The code tries to let the editor breathe a bit to avoid looking frozen.

#77004 should fix the issue at some point, so this PR may be reverted or the approach re-assessed. Some care will be needed anyway in the resource previewing code so that lifespans, etc. are properly handled. Pre-4.3 there's no other fix than this.

Finally, some extra testing would be appreciated.

Fixes #87196.

### DIFF
--- a/editor/editor_resource_preview.cpp
+++ b/editor/editor_resource_preview.cpp
@@ -41,6 +41,7 @@
 #include "editor/editor_settings.h"
 #include "editor/editor_string_names.h"
 #include "scene/resources/image_texture.h"
+#include "servers/rendering/rendering_server_default.h"
 
 bool EditorResourcePreviewGenerator::handles(const String &p_type) const {
 	bool success = false;
@@ -338,6 +339,20 @@ void EditorResourcePreview::_thread() {
 	exited.set();
 }
 
+void EditorResourcePreview::_idle_callback() {
+	if (!singleton) {
+		// Just in case the shutdown of the editor involves the deletion of the singleton
+		// happening while additional idle callbacks can happen.
+		return;
+	}
+
+	// Process preview tasks, trying to leave a little bit of responsiveness worst case.
+	uint64_t start = OS::get_singleton()->get_ticks_msec();
+	while (!singleton->queue.is_empty() && OS::get_singleton()->get_ticks_msec() - start < 100) {
+		singleton->_iterate();
+	}
+}
+
 void EditorResourcePreview::_update_thumbnail_sizes() {
 	if (small_thumbnail_size == -1) {
 		// Kind of a workaround to retrieve the default icon size.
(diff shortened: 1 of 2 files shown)


YuriSizov:
`TilesEditorUtils` is a very close copy of the entire `EditorResourcePreview` class. I assume it needs the same fix?

RandomShaper:
> `TilesEditorUtils` is a very close copy of the entire `EditorResourcePreview` class. I assume it needs the same fix?

Good question. I've checked the code and it seems it's fine already. The code there is already safe by using several sync mechanisms to let the main thread deal with the viewports and eventually, only when it's known the texture is ready, claiming the results.

By the way, `EditorResourcePreview` may want to do something similar, which would still allow some work to happen in parallel on another thread, but such an approach would involve adding the needed "sync points" across both that class and all the preview plugins. I don't think it's very worth the work, when an upcoming PR will fix the underlying issue for good, unless the approach in this PR turned out to be too impractical in practice (he-he, what an expression).

editor/editor_resource_preview.cpp
```diff
@@ -441,27 +456,38 @@ void EditorResourcePreview::check_for_invalidation(const String &p_path) {
...
+	if (DisplayServer::get_singleton()->get_name() == "headless") {
+		return;
+	}
+
+	if (RSG::texture_storage->can_create_resources_async()) {
 		ERR_FAIL_COND_MSG(thread.is_started(), "Thread already started.");
 		thread.start(_thread_func, this);
+	} else {
+		SceneTree *st = Object::cast_to<SceneTree>(OS::get_singleton()->get_main_loop());
+		if (st) {
```
akien-mga, line 468: Is there any realistic scenario where the editor main loop wouldn't be SceneTree?
If not, we might want to put a `ERR_FAIL_NULL(st);` there for good measure instead of `if (st)`.

akien-mga (approved):
Looks good to me.

editor/editor_resource_preview.cpp
RandomShaper: Good point. Changed.

akien-mga:
Tested and confirmed this fixes https://github.com/godotengine/godot/issues/87196.

akien-mga:
Thanks!

Issue #76774, opened 2023-05-06
GeometryInstance3D transparency does not work in mobile or compatibility renderers


fracteed:
### Godot version

4.0.2 stable

### System information

win10 960m + win11 3050

### Issue description

The transparency parameter on a geometry instance 3d object only works  in the forward+ renderer. It does not work with the vulkan mobile or opengl compatibility renderers.

@clayjohn has confirmed that has not been added to these renderers.

### Steps to reproduce

- add a cube mesh
- move the transparency slider

### Minimal reproduction project

Not really necessary.

Calinou:
This also applies to visibility range distance fade using the **Self** or **Dependencies** fade mode. Both act the same as **Disabled** when using the Forward Mobile or Compatibility rendering methods (https://github.com/godotengine/godot/issues/69877).

eswartz:
I spent a few hours before discovering this issue. I might recommend documenting this limitation . The GeometryInstance3D.transparency (etc) docs don't mention that these are only available in Forward+.

jbromberg:
Just ran into this. Should at the very least be documented, but ideally support could be added.

Calinou:
> BTW, should this ticket be replicated into #55871? Currently it's only on the OpenGL/compatibility tracker.

Done.

PR #87231, opened 2024-01-15
Show warning for GeometryInstance3D transparency in Mobile/Compatibility

Calinou:
This feature is currently only supported when using Forward+.

- See https://github.com/godotengine/godot/issues/76774.


eswartz:
I'd like to see this in 4.3 please :) I ran into this yet again the other day (in the guise of fade distance).

clayjohn (approved):
I think it would be quite easy to add this to both renderers. The code can be mostly copy-pasted from the Forward+ renderer, but let's add this for now to avoid confusing people

AThousandShips (approved):

akien-mga:
Thanks!

PR #87233, opened 2024-01-15
Fix a possible crash when unreferencing an empty CowData<T>

### LINKED ISSUE #87076: Godot crashes when importing OGG file or renaming folder / OGG Vorbis (release builds only)
### Tested versions

Godot Engine v4.2.1.stable.mono.official

### System information

Windows 10 - Godot v4.2.1.stable.mono.official (Forward+)

### Issue description

Godot Editor crashes (v4.2.1),
- if you try to import an OGG file 
- if you rename the folder in which the OGG file is located. Reimport causes crash

If you remove the OGG file from the project folder, the project is loaded successfully.

**Better description needed**
I would also like the error text to be better defined. For Lain it is absolutely meaningless.
See also my proposal: https://github.com/godotengine/godot-proposals/discussions/8808
My commit in the pull request: https://github.com/godotengine/godot/pull/84723

### Vide open OGG file project

https://github.com/godotengine/godot/assets/44207756/50e305ad-8095-4fef-8bc1-6d092966d13a

### Steps to reproduce

N/A

[OGG TestProject.zip](https://github.com/godotengine/godot/files/13943427/OGG.TestProject.zip)

bs-mwoerner:
Fixes #87076.

`CowData<T>::_unref()` was implemented like this:
```c++
template <class T>
void CowData<T>::_unref(void *p_data) {
	if (!p_data) {
		return;
	}

	SafeNumeric<uint32_t> *refc = _get_refcount();

	if (refc->decrement() > 0) {
		return; // still in use
	}
	// clean up

	if (!std::is_trivially_destructible<T>::value) {
		uint32_t *count = _get_size();
		T *data = (T *)(count + 1);

		for (uint32_t i = 0; i < *count; ++i) {
			// call destructors
			data[i].~T();
		}
	}

	// free mem
	Memory::free_static((uint8_t *)p_data, true);
}

_FORCE_INLINE_ SafeNumeric<uint32_t> *_get_refcount() const {
	if (!_ptr) {
		return nullptr;
	}

	return reinterpret_cast<SafeNumeric<uint32_t> *>(_ptr) - 2;
}
```

On the surface, `if (!p_data)` doesn't do anything because `Memory::free_static()` uses its own null check and the parameter is otherwise unused, so in the release builds of Godot 4.2.1 and 4.3 dev 1 (and most likely others), there is at least one occasion where this check is optimized out by the compiler. However, the method assumes that `p_data == this->_ptr`, so that `if (!p_data)` is an implicit check for `if (this->_ptr != nullptr)`. If the compiler, unaware of this assumption, removes the `if (!p_data)`, then `if (refc->decrement() > 0)` dereferences a null pointer whenever `this->_ptr` is null.

The `p_data` parameter makes it possible to decrement the reference count of the internal pointer but then free a different pointer should that reference count reach zero. I'm not sure what the original use case for this was, especially since the method may crash should you pass in anything other than `this->ptr`. In the current code, the method is only ever called as `_unref(_ptr)`.

I removed the `p_data` parameter and made the method explicitly operate on `this->_ptr`. This should prevent the compiler from optimizing the check out although I don't know how to create an official release build, so I can't check the generated code.

Since `CowData<T>` is the internal data structure used in `String` and `Vector<T>`, this change affects basically the whole engine. Hopefully in a good way.

### DIFF
--- a/core/templates/cowdata.h
+++ b/core/templates/cowdata.h
@@ -113,7 +113,7 @@ private:
 		return *out;
 	}
 
-	void _unref(void *p_data);
+	void _unref();
 	void _ref(const CowData *p_from);
 	void _ref(const CowData &p_from);
 	uint32_t _copy_on_write();
@@ -195,8 +195,8 @@ public:
 };
 
 template <class T>
-void CowData<T>::_unref(void *p_data) {
-	if (!p_data) {
+void CowData<T>::_unref() {
+	if (!_ptr) {
 		return;
 	}
 
@@ -218,7 +218,7 @@ void CowData<T>::_unref(void *p_data) {
 	}
 
 	// free mem
-	Memory::free_static((uint8_t *)p_data, true);
+	Memory::free_static((uint8_t *)_ptr, true);
 }
 
 template <class T>
@@ -251,7 +251,7 @@ uint32_t CowData<T>::_copy_on_write() {
 			}
 		}
 
-		_unref(_ptr);
+		_unref();
 		_ptr = _data;
 
 		rc = 1;
@@ -272,7 +272,7 @@ Error CowData<T>::resize(int p_size) {
 
 	if (p_size == 0) {
 		// wants to clean up
-		_unref(_ptr);
+		_unref();
 		_ptr = nullptr;
 		return OK;
 	}
@@ -398,7 +398,7 @@ void CowData<T>::_ref(const CowData &p_from) {
 		return; // self assign, do nothing.
 	}
 
-	_unref(_ptr);
+	_unref();
 	_ptr = nullptr;
 
 	if (!p_from._ptr) {
@@ -412,7 +412,7 @@ void CowData<T>::_ref(const CowData &p_from) {
 
 template <class T>
 CowData<T>::~CowData() {
-	_unref(_ptr);
+	_unref();
 }
 
 #if defined(__GNUC__) && !defined(__clang__)


bs-mwoerner:
Setting this to draft for now because of a report that the crash is still there in a release build that includes this change and because I don't quite agree with the logic of my explanation for why the compiler might remove the null check. I'd welcome any comments and insights.

This is the relevant code that triggers the crash in the linked issue:

```c++
RBMap<uint64_t, Vector<Vector<uint8_t>>> sorted_packets;
while (true)
{
    // [...]
    PackedByteArray data;
    data.resize(packet.bytes);
    memcpy(data.ptrw(), packet.packet, packet.bytes);
    sorted_packets[granule_pos].push_back(data);
    packet_count++;
}
```

From what I gather from the assembly, this is what the compiler does when the scope ends (pseudocode):
```
packet_count = packet_count + 1;
if (tempdata._cowdata._ptr != nullptr)
{
    decrement the reference count at tempdata._cowdata._ptr - 8
    if the reference count is now 0, free tempdata._cowdata._ptr

    if (data._cowdata._ptr != nullptr)
        refcptr = data._cowdata._ptr - 8
    else
        refcptr = nullptr
}
else
{
    // This branch is taken just before the crash
    if (r12 != nullptr)
        refcptr = r12 - 8
    else
        refcptr = nullptr
}
decrement the reference count at refcptr // This crashes on the final loop because r12 and thus refcptr is null
if the reference count is now 0, free r12
goto top_of_while
```
`r12` is the first parameter that was passed to `memcpy(data.ptrw(), packet.packet, packet.bytes)`, which is the same value as `data._cowdata._ptr` and `tempdata._cowdata._ptr` (`tempdata` is probably some temporary copy created when copy-constructing the Vector in the map). The compiler seems to be certain that a null check is not necessary when doing `data._cowdata._unref(data._cowdata._ptr)`, but I can't see why.

bs-mwoerner:
So, after noticing that my own production builds with Visual Studio don't crash, realizing that Godot's Windows binaries are built with gcc, and reading up on gcc quirks, I learned that gcc indeed considers a pointer to be proven non-null if it has ever been passed to `memcpy`, even if that memcpy was a zero-byte nop and didn't touch or check the pointer. So there's actually nothing wrong with `CowData<T>`. The cause of the linked crash is likely that when compiling with gcc, the zero-byte `memcpy` silently removes the null check when `data` is eventually cleaned up after going out of scope. Therefore, `ResourceImporterOggVorbis` must make sure that the data is not zero-length before attempting to "copy" it into the array.

Sorry about the turmoil. I didn't expect the compiler to be so bold in its optimizations and didn't even consider that as a possible side effect. I'll open a new pull request for the fix in `ResourceImporterOggVorbis`.

PR #87240, opened 2024-01-16
Add bone update option to OpenXRHand to allow preserving original hand scale

Malcolmnixon:
This pull request adds an option to OpenXRHand to select the type of bone updates to perform.
image

The supported options are:
- Full (default) which works the same as today applying both position and rotation of the tracked hand
- Rotation Only which only rotates the bones preserving their relative position and length

The Rotation mode is useful if the hand model is not weighted to support re-scaling to the users hands; or the hand is scaled to non-human sizes.

With "Full" the following hand model gets damaged (the fingers get pushed into the palm):
image

Switching to "Rotation Only" preserves the bone lengths - and the mesh integrity:
image


BastiaanOlij:
I would rename 'All' to 'Full' and 'Rotation' to 'Rotation only'.

Malcolmnixon:
> I would rename 'All' to 'Full' and 'Rotation' to 'Rotation only'.

Done
image

dsnopek (approved):
I haven't tested it, but the code looks good to me!

m4gr3d (approved):

YuriSizov:
Thanks!

PR #87243, opened 2024-01-16
Add support for fractional font sizes.

### LINKED ISSUE #9559: Make theme constant and font size use float instead of int

Currently building a solution to theme animations in godot

Turns out all theme constant and font size types are integers, so animating their values over time can't be smooth

Floats would allow smooth transitions

`theme_constant` and `theme_font_size` should use float instead. For any theme constant that may actually need an integer, just floor or round the float

No. Since theme constants are the only way to configure nodes like HBoxContainer, such nodes can't have float constants

Theme constant and theme font size handling are core

### LINKED ISSUE #36815: Support for theme item booleans
**Godot version:**
da502f7

**Issue description:**
AFAIK theme properties don't support boolean values, but there's few of them that fake booleans by using 0 and 1 constants. Example is `shadow_as_outline` in (RichText)Label or `autohide` in H/VSplitContainer. As briefly discussed in #36773, we should make these properties proper booleans.

bruvzg:
Changes `font_size` and `outline_size` (theme constants) from `int` to `float` for better font scaling control.

- *Production edit: This closes https://github.com/godotengine/godot/issues/36815 and closes https://github.com/godotengine/godot-proposals/issues/9559 (see https://github.com/godotengine/godot/pull/87243#issuecomment-1894266187).*


YuriSizov:
We can't change theme constants to floats, it breaks compatibility and hinders their usability as booleans.

There is already a PR by @Calinou for that, but we haven't merged it for those reasons.

bruvzg:
As a POC, switched theme constants to `Variant` with type annotation (it the constant declarations), and validation (on set/override, warning if set to wrong type value, error if value is not convertible to correct type).

Should also use correct property editor fields:
Screenshot 2024-01-16 at 20 11 08
Screenshot 2024-01-16 at 20 11 33

AdriaandeJongh:
As my proposal [here](https://github.com/godotengine/godot-proposals/issues/9323) is strongly related to this PR, could / should glyph spacing and space glyph spacing also be included in this PR?

CodeCrafterLog:
I really think that this PR should merge, because when I wanted to use an animation on a [margin container](http://docs.godotengine.org/en/stable/classes/class_margincontainer.html) I couldn't make it smooth while changing it's constand margins in theme overrides. As a solution for those who have simular problems with theme overrides and animations I recommend to use a simple tool script like as follow:

```
@tool
class_name SmoothMargin extends MarginContainer

@export var margin_left: float = 0:
	set(v):
		margin_left = v
		add_theme_constant_override("margin_left", floori(v))
	get:
		return get_theme_constant("margin_left")
@export var margin_top: float = 0:
	set(v):
		margin_top = v
		add_theme_constant_override("margin_top", floori(v))
	get:
		return get_theme_constant("margin_top")
@export var margin_right: float = 0:
	set(v):
		margin_right = v
		add_theme_constant_override("margin_right", floori(v))
	get:
		return get_theme_constant("margin_right")
@export var margin_bottom: float = 0:
	set(v):
		margin_bottom = v
		add_theme_constant_override("margin_bottom", floori(v))
	get:
		return get_theme_constant("margin_bottom")
```

Just use the values of your own theme overrides

Calinou:
> What about interpolating the changes internally instead?
Adding animation checkbox that will display animation setting for the constants.

AnimationPlayer can already interpolate integer values between each other, but intermediate steps will never be used as the underlying type is still `int`. This means font size changes would look visibly jittery.

In general, changing font size continuously with an animation is not recommended unless the font uses MSDF rendering (or font oversampling is disabled in the Import dock). With standard dynamic font rendering, the font must be re-rasterized for each size. With floating-point sizes, this means the font must be re-rasterized every frame as it'll (almost) always pick a new, unique font size (e.g. `11.5` requires different rasterization than `11.6`).

Calinou:
>  the text gets blur at some point.

This doesn't happen if the font is imported with MSDF enabled in the Import dock. MSDF fonts are always crisp, although they don't look as good as dynamic fonts at small sizes.

PR #76575, opened 2023-04-29
Fix renaming a node to the name of its siblings causing exported `NodePath` to point to the wrong node

nongvantinh:
Once text editing is complete, the method `void SceneTreeEditor::_renamed()` is triggered, followed by `void SceneTreeDock::_node_prerenamed(Node *p_node, const String &p_new_name)`. Afterward, the method `void SceneTreeDock::perform_node_renames(Node *p_base, HashMap<Node *, NodePath> *p_renames, HashMap<Ref<Animation>, HashSet<int>> *r_rem_anims)` is called, which ultimately leads to line 1707 where the method `p_base->set(propertyname, updated_variant);` modifies the previously set value with the new Node's name. Next, the method `void SceneTreeEditor::_rename_node(ObjectID p_node, const String &p_name)` is triggered to update the TreeItem in SceneTreeEditor and the Node's name. However, the problem arises in line 962: `n->set_name(p_name);` when the Node's name is validated using `data.parent->_validate_child_name(this, true);`. It is then discovered that the new name conflicts with a sibling's name, prompting `_generate_serial_child_name(p_child, name);` to append a number to the renaming Node's name. Unfortunately, the value of the exported variable is not updated, causing the bug.

Fixes: #76192
Fixes: #76680

PR #87252, opened 2024-01-16
Fix renaming a node to the name of its siblings breaking NodePath

### LINKED ISSUE #76192: Renaming a name present in a NodePath to an invalid name will wrongly update the NodePath
### Godot version

4.1.dev (4d5f10fc3a)

### System information

MacOS 13.3 M1 Pro

### Issue description

If you have a NodePath pointing to a node and you rename a node on this path to a value that is already used by one of his siblings, the NodePath will be updated with the value you set while the node will be updated with another unused value created by adding a number after the value you set.

### Steps to reproduce

1. create a root and two children
2. On the root export a NodePath and set it to one of the children
3. Rename the referenced child to the name of the other child
4. now the NodePath points to the other child

N/A

ajreckof:
rebase and updated version of #76575
Fix #76192
I also fix another issue where there was cases where node name was not visually updated because name was changed to an invalid node name which was automatically changed to a valid one which happens to be the previous name of the node which cancels renaming.
This issue was already present with invalid character and this PR added the same for renaming to the name of a sibling.

### DIFF
--- a/scene/main/node.h
+++ b/scene/main/node.h
@@ -614,6 +614,7 @@ public:
 
 #ifdef TOOLS_ENABLED
 	String validate_child_name(Node *p_child);
+	String prevalidate_child_name(Node *p_child, StringName p_name);
 #endif
 	static String adjust_name_casing(const String &p_name);
 
--- a/editor/gui/scene_tree_editor.cpp
+++ b/editor/gui/scene_tree_editor.cpp
@@ -1051,10 +1051,9 @@ void SceneTreeEditor::_rename_node(Node *p_node, const String &p_name) {
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


scene/main/node.h
```diff
@@ -614,6 +614,7 @@ class Node : public Object {
 
 #ifdef TOOLS_ENABLED
 	String validate_child_name(Node *p_child);
+	String prevalidate_child_name(Node *p_child, StringName p_name);
```
AThousandShips, line 617:
```suggestion
	String prevalidate_child_name(Node *p_child, const StringName &p_name);
```

KoBeWi (approved):

YuriSizov:
Thanks!

PR #80343, opened 2023-08-06
C#: Add analyzer release tracking

raulsntos:
Fixes diagnostic [RS2008](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md#rs2008-enable-analyzer-release-tracking).

PR #64899, opened 2022-08-25
Make C# source generators incremental

raulsntos:
Switch to incremental source generators to increase their performance as they are executed in steps and avoid re-doing work.

### TODO
- [ ] I am not convinced that the way I'm reading the analyzer properties is the best way to do it since the syntax step is still going to be executed regardless of the properties' values.
- [ ] I haven't tested them thoroughly and I'd like to make sure I didn't introduce a regression before marking the PR out of draft.

_Production edit: closes godotengine/godot-roadmap#50_

PR #87253, opened 2024-01-16
Provide a roslyn analyzers corresponding to the GD0001 and GD0002

### LINKED ISSUE #81: GD0001: partial modifier static analysis for Godot 4
Godot 4 now requires anything inheriting from Godot.Object to be partial. The Plugin doesn't recognise this, yet

example code: 

```
public class Main : Node3D
{
	public override void _Ready()
	{
		GD.Print("_Ready");
	}
}
```

This looks fine in Rider, but when building in Godot it yields:

>[...]\Main.cs(4,1): Error GD0001 : Missing partial modifier on declaration of type 'Main' which is a subclass of 'Godot.Object'

It would be good to have a static analysis of Rider to show that error in the editor before building.

van800:
Fixes https://github.com/JetBrains/godot-support/issues/81

Discussion:
https://chat.godotengine.org/channel/dotnet?msg=dtyXNTdDNij3S7Jvr

I have also added a code-fix 
image
Tried it in Rider and VScode

Benefit for vscode users is that report from analyzer is much nicer then multiple reports from generators.
Old behaviour:
image
New behaviour:
image

### DIFF
--- a/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators.Tests/ClassPartialModifierAnalyzerTest.cs
+++ b/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators.Tests/ClassPartialModifierAnalyzerTest.cs
@@ -1,0 +1,20 @@
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Godot.SourceGenerators.Tests;
+
+public class ClassPartialModifierTest
+{
+    [Fact]
+    public async Task ClassPartialModifierCodeFixTest()
+    {
+        await CSharpCodeFixVerifier<ClassPartialModifierCodeFixProvider, ClassPartialModifierAnalyzer>
+            .Verify("ClassPartialModifier.GD0001.cs", "ClassPartialModifier.GD0001.fixed.cs");
+    }
+
+    [Fact]
+    public async void OuterClassPartialModifierAnalyzerTest()
+    {
+        await CSharpAnalyzerVerifier<ClassPartialModifierAnalyzer>.Verify("OuterClassPartialModifierAnalyzer.GD0002.cs");
+    }
+}
--- a/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators.Tests/TestData/Sources/ClassPartialModifier.GD0001.cs
+++ b/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators.Tests/TestData/Sources/ClassPartialModifier.GD0001.cs
@@ -1,0 +1,6 @@
+using Godot;
+
+public class {|GD0001:ClassPartialModifier|} : Node
+{
+    
+}
--- a/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
+++ b/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
@@ -1,0 +1,112 @@
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Godot.SourceGenerators
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class ClassPartialModifierAnalyzer : DiagnosticAnalyzer
+    {
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
+            ImmutableArray.Create(Common.ClassPartialModifierRule, Common.OuterClassPartialModifierRule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+            context.RegisterSyntaxNodeAction(AnalyzeNode, SyntaxKind.ClassDeclaration);
+        }
+
+        private void AnalyzeNode(SyntaxNodeAnalysisContext context)
+        {
+            if (context.Node is not ClassDeclarationSyntax classDeclaration)
+                return;
+
+            if (context.ContainingSymbol is not INamedTypeSymbol typeSymbol)
+                return;
+
+            if (!typeSymbol.InheritsFrom("GodotSharp", GodotClasses.GodotObject))
+                return;
+
+            if (!classDeclaration.IsPartial())
+                context.ReportDiagnostic(Diagnostic.Create(
+                    Common.ClassPartialModifierRule,
+                    classDeclaration.Identifier.GetLocation(),
+                    typeSymbol.ToDisplayString()));
+
+            var outerClassDeclaration = context.Node.Parent as ClassDeclarationSyntax;
+            while (outerClassDeclaration is not null)
+            {
+                var outerClassTypeSymbol = context.SemanticModel.GetDeclaredSymbol(outerClassDeclaration);
+                if (outerClassTypeSymbol == null)
+                    return;
+
+                if (!outerClassDeclaration.IsPartial())
+                    context.ReportDiagnostic(Diagnostic.Create(
+                        Common.OuterClassPartialModifierRule,
+                        outerClassDeclaration.Identifier.GetLocation(),
+                        outerClassTypeSymbol.ToDisplayString()));
+
+                outerClassDeclaration = outerClassDeclaration.Parent as ClassDeclarationSyntax;
+            }
+        }
+    }
+
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public sealed class ClassPartialModifierCodeFixProvider : CodeFixProvider
+    {
+        public override ImmutableArray<string> FixableDiagnosticIds =>
+            ImmutableArray.Create(Common.ClassPartialModifierRule.Id);
+
+        public override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            // Get the syntax root of the document.
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            // Get the diagnostic to fix.
+            var diagnostic = context.Diagnostics.First();
+
+            // Get the location of code issue.
+            var diagnosticSpan = diagnostic.Location.SourceSpan;
+
+            // Use that location to find the containing class declaration.
+            var classDeclaration = root?.FindToken(diagnosticSpan.Start)
+                .Parent?
+                .AncestorsAndSelf()
+                .OfType<ClassDeclarationSyntax>()
+                .First();
+
+            if (classDeclaration == null)
+                return;
+
+            context.RegisterCodeFix(
+                CodeAction.Create(
+                    "Add partial modifier",
+                    cancellationToken => AddPartialModifierAsync(context.Document, classDeclaration, cancellationToken),
+                    classDeclaration.ToFullString()),
+                context.Diagnostics);
+        }
+
+        private static async Task<Document> AddPartialModifierAsync(Document document,
+            ClassDeclarationSyntax classDeclaration, CancellationToken cancellationToken)
+        {
+            // Create a new partial modifier.
+            var partialModifier = SyntaxFactory.Token(SyntaxKind.PartialKeyword);
+            var modifiedClassDeclaration = classDeclaration.AddModifiers(partialModifier);
+            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            // Replace the old class declaration with the modified one in the syntax root.
+            var newRoot = root!.ReplaceNode(classDeclaration, modifiedClassDeclaration);
+            var newDocument = document.WithSyntaxRoot(newRoot);
+            return newDocument;
+        }
+    }
+}
--- a/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/Common.cs
+++ b/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/Common.cs
@@ -7,63 +7,25 @@ namespace Godot.SourceGenerators
     {
         private static readonly string _helpLinkFormat = $"{VersionDocsUrl}/tutorials/scripting/c_sharp/diagnostics/{{0}}.html";
 
-        public static void ReportNonPartialGodotScriptClass(
-            GeneratorExecutionContext context,
-            ClassDeclarationSyntax cds, INamedTypeSymbol symbol
-        )
-        {
-            string message =
-                "Missing partial modifier on declaration of type '" +
-                $"{symbol.FullQualifiedNameOmitGlobal()}' that derives from '{GodotClasses.GodotObject}'";
+        internal static readonly DiagnosticDescriptor ClassPartialModifierRule =
+            new DiagnosticDescriptor(id: "GD0001",
+                title: $"Missing partial modifier on declaration of type that derives from '{GodotClasses.GodotObject}'",
+                messageFormat: $"Missing partial modifier on declaration of type '{{0}}' that derives from '{GodotClasses.GodotObject}'",
+                category: "Usage",
+                DiagnosticSeverity.Error,
+                isEnabledByDefault: true,
+                $"Classes that derive from '{GodotClasses.GodotObject}' must be declared with the partial modifier.",
+                helpLinkUri: string.Format(_helpLinkFormat, "GD0001"));
 
-            string description = $"{message}. Classes that derive from '{GodotClasses.GodotObject}' " +
-                                 "must be declared with the partial modifier.";
-
-            context.ReportDiagnostic(Diagnostic.Create(
-                new DiagnosticDescriptor(id: "GD0001",
-                    title: message,
-                    messageFormat: message,
-                    category: "Usage",
-                    DiagnosticSeverity.Error,
-                    isEnabledByDefault: true,
-                    description,
-                    helpLinkUri: string.Format(_helpLinkFormat, "GD0001")),
-                cds.GetLocation(),
-                cds.SyntaxTree.FilePath));
-        }
-
-        public static void ReportNonPartialGodotScriptOuterClass(
-            GeneratorExecutionContext context,
-            TypeDeclarationSyntax outerTypeDeclSyntax
-        )
-        {
-            var outerSymbol = context.Compilation
-                .GetSemanticModel(outerTypeDeclSyntax.SyntaxTree)
-                .GetDeclaredSymbol(outerTypeDeclSyntax);
-
-            string fullQualifiedName = outerSymbol is INamedTypeSymbol namedTypeSymbol ?
-                namedTypeSymbol.FullQualifiedNameOmitGlobal() :
-                "type not found";
-
-            string message =
-                $"Missing partial modifier on declaration of type '{fullQualifiedName}', " +
-                $"which contains nested classes that derive from '{GodotClasses.GodotObject}'";
-
-            string description = $"{message}. Classes that derive from '{GodotClasses.GodotObject}' and their " +
-                                 "containing types must be declared with the partial modifier.";
-
-            context.ReportDiagnostic(Diagnostic.Create(
-                new DiagnosticDescriptor(id: "GD0002",
-                    title: message,
-                    messageFormat: message,
-                    category: "Usage",
-                    DiagnosticSeverity.Error,
-                    isEnabledByDefault: true,
-                    description,
-                    helpLinkUri: string.Format(_helpLinkFormat, "GD0002")),
-                outerTypeDeclSyntax.GetLocation(),
-                outerTypeDeclSyntax.SyntaxTree.FilePath));
-        }
+        internal static readonly DiagnosticDescriptor OuterClassPartialModifierRule =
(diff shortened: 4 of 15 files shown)


van800:
@raulsntos I have cherry-picked your change from https://github.com/godotengine/godot/pull/87342

paulloz (commented):
As discussed previously, this moves the diagnostics we were reporting from within the generators to its own analyzer, right? Essentially doing the job twice (i.e. the generator is already visiting all those classes, and now the analyzer does too).  

I think we could probably report only from `ScriptPathAttributeGenerator`, and fix the source location to not underline the entire class.

paulloz:
Actually, the more I think about it, the more I'm back to thinking that moving all of our diagnostics to their own analyzers is a good idea. At the same time, we could clean the mess that is `Common.cs` at the moment (e.g., we're defining diagnostics as `static readonly`, but not using them depending on from where it is reported, see [here](https://github.com/godotengine/godot/blob/5034478611697358f4e135c69f364ef6d8cd54dc/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/Common.cs#L464-L493)).

It'd also reduce the work done by incremental generators once #64899 is a thing. Not to mention that the eventual shift to a GDExtension implementation, as far as I understand, will mean the generated content will be vastly different (the generators are going to change), while the public API should stay compatible (the analyzers/diagnostics should stay the same).

paulloz (commented):
I have a CSC error about this analyzer, both when compiling `Godot.SourceGenerators`, and when opening Godot afterwards. Not entirely sure why.
```
CSC : warning CS8032: An instance of analyzer Godot.SourceGenerators.ClassPartialModifierAnalyzer cannot be created from E:\Godot Worktree\godot\modules\mono\editor\G
odot.NET.Sdk\Godot.SourceGenerators\bin\Release\netstandard2.0\Godot.SourceGenerators.dll : Could not load type 'Godot.SourceGenerators.ClassPartialModifierAnalyzer' 
from assembly 'Godot.SourceGenerators, Version=4.3.0.0, Culture=neutral, PublicKeyToken=null'.. [E:\Godot Worktree\godot\modules\mono\editor\Godot.NET.Sdk\Godot.Sourc 
eGenerators.Sample\Godot.SourceGenerators.Sample.csproj]
```

Apart from that, it seems fine.  

The only thing I'm a bit scared about is users will now be able to compile broken projects by suppressing `GD0001`. We say that it shouldn't be suppressed, but still... I wonder if we should find a way to still block compilation here (with something additionally reporting at `Location.None`, or IDK). 

Would be nice for #88295 to be merged before, to prevent regressions. And maybe #80343 too.

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+            messageFormat: MessageFormat,
+            category: "Usage",
+            DiagnosticSeverity.Error,
+            isEnabledByDefault: true,
+            Description,
+            helpLinkUri: string.Format(Common.HelpLinkFormat, RuleId));
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
+            => ImmutableArray.Create(
+                Rule);
```
raulsntos, line 39: This line doesn't seem to be long enough to require a line-break here.

```suggestion
            => ImmutableArray.Create(Rule);
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+                .AncestorsAndSelf()
+                .OfType<ClassDeclarationSyntax>()
+                .First();
+
+            if (classDeclaration == null)
+                return;
+
+            context.RegisterCodeFix(
+                CodeAction.Create(
+                    "Make class partial",
```
raulsntos, line 95: I think the usual wording for these kinds of fixes in the BCL is something like this:

```suggestion
                    "Add 'partial' modifier",
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+            if (!typeSymbol.InheritsFrom("GodotSharp", GodotClasses.GodotObject))
+                return;
+
+            if (!classDeclaration.IsPartial())
+                context.ReportDiagnostic(Diagnostic.Create(Rule, classDeclaration.Identifier.GetLocation(), typeSymbol.Name, GodotClasses.GodotObject));
+        }
+    }
+
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public class ClassPartialModifierAnalyzerFix : CodeFixProvider
```
raulsntos, line 66: I would avoid the word 'analyzer' in the class name. Also this class can be sealed instead of sealing all the overridden methods.

```suggestion
    public sealed class ClassPartialModifierCodeFixProvider : CodeFixProvider
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Document = Microsoft.CodeAnalysis.Document;
+
+namespace Godot.SourceGenerators
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class ClassPartialModifierAnalyzer : DiagnosticAnalyzer
```
raulsntos, line 16: We can also make this class sealed.

```suggestion
    public sealed class ClassPartialModifierAnalyzer : DiagnosticAnalyzer
```

raulsntos:
> The only thing I'm a bit scared about is users will now be able to compile broken projects by suppressing `GD0001`

The project should still not compile if we are generating partials of a class not marked as partial.

paulloz:
But we are skipping non partial classes, so we won't generate the partials.

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators.Tests/ClassPartialModifierAnalyzerTest.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Godot.SourceGenerators.Tests;
+
+public class ClassPartialModifierTest
+{
+    const string nodeWithoutPartialModifier = "ClassPartialModifierAnalyzer.GD0001.cs";
+    [Fact]
+    public async void Test()
+    {
+        await CSharpAnalyzerVerifier<ClassPartialModifierAnalyzer>.Verify(nodeWithoutPartialModifier);
```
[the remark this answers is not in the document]
paulloz: That's interesting, I'll note it down 📝

raulsntos:
We probably shouldn't, otherwise that means we're still checking if the class is partial in the generator and we said we were moving the check to the analyzer. 

That said, it's probably fine to keep the skipping behavior in this PR, changing it is tricky because it's kind of mixed with GD0002 which would increase the scope of this PR. GD0002 should probably be included in the GD0001 analyzer anyway, but I didn't want to delay this PR further.

And to be honest, I don't expect users to disable this diagnostic, doing so would be ignoring an error in your code so whatever happens after that can be considered undefined behavior.

van800:
Thank you for all the suggestions, I have also cleaned up redundant sealed modifiers and did a cosmetic fix, which bugged me: trimmed the repeating part of Description. Unfortunately the practice of preparing just one commit pull request makes it harder for you to follow, what was changed. For the future, should I keep multiple commits and squash them after the review?

AThousandShips:
When this is ready to merge you will need to squash :) you can do it now but there will be some code style review once this has been approved

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,115 @@
...
+                context.Diagnostics);
+        }
+
+        private static async Task<Document> AddPartialModifierAsync(Document document,
+            ClassDeclarationSyntax classDeclaration, CancellationToken cancellationToken)
+        {
+            // Create a new partial modifier
+            var partialModifier = SyntaxFactory.Token(SyntaxKind.PartialKeyword);
+            var modifiedClassDeclaration = classDeclaration.AddModifiers(partialModifier);
+            var root = await document.GetSyntaxRootAsync(cancellationToken);
```
raulsntos, line 108: Should add `.ConfigureAwait(false)`.

```suggestion
            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,115 @@
...
+            // Replace the old class declaration with the modified one in the syntax root
+            var newRoot = root.ReplaceNode(classDeclaration, modifiedClassDeclaration);
```
raulsntos, line 110: I guess `root` is never going to be null in practice?

```suggestion
            var newRoot = root!.ReplaceNode(classDeclaration, modifiedClassDeclaration);
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators.Tests/TestData/Sources/ClassPartialModifier.GD0001.cs
```diff
@@ -0,0 +1,6 @@
+using Godot;
+
+public class {|GD0001:ClassPartialModifier|} : Node
+{
+    
+}
```
raulsntos, line 6:
```suggestion
}

```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/Godot.SourceGenerators.csproj
```diff
@@ -21,8 +21,7 @@
     <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.8.0" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.3" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.8.0" PrivateAssets="all" />
```
raulsntos, line 24: Can you explain why this change is needed? I suspect this is causing the `CS8032` diagnostic that @paulloz is getting.

van800: Thankfully not an expert in that.
I was thinking that, there are high-level packages like "Microsoft.CodeAnalysis.CSharp.Workspaces" or "Microsoft.CodeAnalysis.CSharp.Features". Any of those two references all the required smaller packages.
 
 When implementing a code-fix, I was looking at https://github.com/microsoft/Microsoft.Unity.Analyzers/blob/9159bac481bfcfe01913c4ac9594b1fdcb139734/src/Microsoft.Unity.Analyzers/Microsoft.Unity.Analyzers.csproj#L12 as a reference.

van800:
I have also rewritten the GD0002.
Unfortunately, I can't answer the question about which nuget package to prefer.
Locally I have tried "Microsoft.CodeAnalysis.CSharp.Features" instead of "Microsoft.CodeAnalysis.CSharp.Workspaces".
It also works fine for me.

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,95 @@
...
+            // Return if not a type symbol.
+            if (context.ContainingSymbol is not INamedTypeSymbol typeSymbol)
+                return;
+
+            if (!typeSymbol.InheritsFrom("GodotSharp", GodotClasses.GodotObject))
+                return;
+
+            if (!classDeclaration.IsPartial())
+                context.ReportDiagnostic(Diagnostic.Create(Common.ClassPartialModifierRule, classDeclaration.Identifier.GetLocation(),
+                    typeSymbol.Name));
```
raulsntos, line 41: It looks like we use `Symbol.ToDisplayString()` instead of `Symbol.Name` in all the other diagnostics. Also, we usually split `ReportDiagnostic` like this:

```suggestion
                context.ReportDiagnostic(Diagnostic.Create(
                    Common.ClassPartialModifierRule,
                    classDeclaration.Identifier.GetLocation(),
                    typeSymbol.ToDisplayString()));
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/Godot.SourceGenerators.csproj
van800: downgrading the version seem to be fine. pushed that.

paulloz:
> I have a CSC error about this analyzer, both when compiling `Godot.SourceGenerators`, and when opening Godot afterwards. Not entirely sure why.
> 
> ```
> CSC : warning CS8032: An instance of analyzer Godot.SourceGenerators.ClassPartialModifierAnalyzer cannot be created from E:\Godot Worktree\godot\modules\mono\editor\G
> odot.NET.Sdk\Godot.SourceGenerators\bin\Release\netstandard2.0\Godot.SourceGenerators.dll : Could not load type 'Godot.SourceGenerators.ClassPartialModifierAnalyzer' 
> from assembly 'Godot.SourceGenerators, Version=4.3.0.0, Culture=neutral, PublicKeyToken=null'.. [E:\Godot Worktree\godot\modules\mono\editor\Godot.NET.Sdk\Godot.Sourc 
> eGenerators.Sample\Godot.SourceGenerators.Sample.csproj]
> ```

This isn't happening any more 👍

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,125 @@
...
+                return;
+
+            if (!classDeclaration.IsPartial())
+                context.ReportDiagnostic(Diagnostic.Create(
+                    Common.ClassPartialModifierRule,
+                    classDeclaration.Identifier.GetLocation(),
+                    typeSymbol.ToDisplayString()));
+        }
+
+        private void AnalyzeNodeOuterClass(SyntaxNodeAnalysisContext context)
```
raulsntos, line 47: I think `AnalyzeNodeOuterClass` could be combined with `AnalyzeNode`.

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,125 @@
...
+                return;
+
+            if (!typeSymbol.InheritsFrom("GodotSharp", GodotClasses.GodotObject))
+                return;
+
+            var outerClassDeclaration = context.Node.Parent as ClassDeclarationSyntax;
+            while (outerClassDeclaration is not null)
+            {
+                var outerClassTypeSymbol = context.SemanticModel.GetDeclaredSymbol(outerClassDeclaration);
+                if (outerClassTypeSymbol == null) return;
```
raulsntos, line 62: To follow our code-style, the `return` should go on its own line.

```suggestion
                if (outerClassTypeSymbol == null)
                    return;
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/Common.cs
```diff
@@ -7,63 +7,25 @@ public static partial class Common
...
+
+        internal static readonly DiagnosticDescriptor OuterClassPartialModifierRule = new(
+            id: "GD0002",
+            title: $"Missing partial modifier on declaration of type which contains nested classes that derive from '{GodotClasses.GodotObject}'",
+            messageFormat: $"Missing partial modifier on declaration of type '{{0}}' which contains nested classes that derive from '{GodotClasses.GodotObject}'",
+            category: "Usage",
+            DiagnosticSeverity.Error,
+            isEnabledByDefault: true,
+            $"Classes that derive from '{GodotClasses.GodotObject}' and their containing types must be declared with the partial modifier.",
+            helpLinkUri: string.Format(_helpLinkFormat, "GD0002"));
```
raulsntos, line 28: To be consistent with the other rules and because RS2003 doesn't work with target-typed new.

```suggestion
        internal static readonly DiagnosticDescriptor ClassPartialModifierRule =
            new DiagnosticDescriptor(id: "GD0001",
                title: $"Missing partial modifier on declaration of type that derives from '{GodotClasses.GodotObject}'",
                messageFormat: $"Missing partial modifier on declaration of type '{{0}}' that derives from '{GodotClasses.GodotObject}'",
                category: "Usage",
                DiagnosticSeverity.Error,
                isEnabledByDefault: true,
                $"Classes that derive from '{GodotClasses.GodotObject}' must be declared with the partial modifier.",
                helpLinkUri: string.Format(_helpLinkFormat, "GD0001"));

        internal static readonly DiagnosticDescriptor OuterClassPartialModifierRule =
            new DiagnosticDescriptor(id: "GD0002",
                title: $"Missing partial modifier on declaration of type which contains nested classes that derive from '{GodotClasses.GodotObject}'",
                messageFormat: $"Missing partial modifier on declaration of type '{{0}}' which contains nested classes that derive from '{GodotClasses.GodotObject}'",
                category: "Usage",
                DiagnosticSeverity.Error,
                isEnabledByDefault: true,
                $"Classes that derive from '{GodotClasses.GodotObject}' and their containing types must be declared with the partial modifier.",
                helpLinkUri: string.Format(_helpLinkFormat, "GD0002"));
```

raulsntos (approved):
Looks good to me. Thanks!

akien-mga:

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+            ImmutableArray.Create(Common.ClassPartialModifierRule.Id);
+
+        public override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            // Get the syntax root of the document.
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            // Get the diagnostic to fix
```
AThousandShips, line 75:
```suggestion
            // Get the diagnostic to fix.
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+            var diagnostic = context.Diagnostics.First();
+
+            // Get the location of code issue
```
AThousandShips, line 78:
```suggestion
            // Get the location of code issue.
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+            var diagnosticSpan = diagnostic.Location.SourceSpan;
+
+            // Use that location to find the containing class declaration
```
AThousandShips, line 81:
```suggestion
            // Use that location to find the containing class declaration.
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+                    "Add partial modifier",
+                    cancellationToken => AddPartialModifierAsync(context.Document, classDeclaration, cancellationToken),
+                    classDeclaration.ToFullString()),
+                context.Diagnostics);
+        }
+
+        private static async Task<Document> AddPartialModifierAsync(Document document,
+            ClassDeclarationSyntax classDeclaration, CancellationToken cancellationToken)
+        {
+            // Create a new partial modifier
```
AThousandShips, line 102:
```suggestion
            // Create a new partial modifier.
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/ClassPartialModifierAnalyzer.cs
```diff
@@ -0,0 +1,112 @@
...
+            var partialModifier = SyntaxFactory.Token(SyntaxKind.PartialKeyword);
+            var modifiedClassDeclaration = classDeclaration.AddModifiers(partialModifier);
+            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            // Replace the old class declaration with the modified one in the syntax root
```
AThousandShips, line 106:
```suggestion
            // Replace the old class declaration with the modified one in the syntax root.
```

AThousandShips (approved):

akien-mga:
Thanks!

PR #84939, opened 2023-11-15
Update control xml and color rect xml regarding mouse filter defaults

Daniel-The-Fox:
Please see issue https://github.com/godotengine/godot/issues/84796 which describes and discusses the confusion the default values for property `mouse_filter` can cause. 

I'm therefore suggesting to update the docs for `Control` and `ColorRect` to warn users while the default of `mouse_filter` for `ColorRect` is set to `MOUSE_FILTER_STOP`. 🤯😢 

Changing the default value of `mouse_filter` for `ColorRect` should be in scope of another PR, once the discussion in issue https://github.com/godotengine/godot/issues/84796 confirmed to do so. I would highly encourage it... 😉

Issue #84796, opened 2023-11-12
Event Handling Inconsistency with ColorRect and Sprite with Area2D


p10tr3k:
### Godot version

v4.1.3.stable.official [f06b6836a]

### System information

MacOs 14.1 (intel)

### Issue description

I've encountered a potential issue with mouse event handling when using a Sprite with Area2D and CollisionShape2D in conjunction with a CanvasLayer and ColorRect. The setup is as follows:

- A scene containing a Sprite with an attached Area2D and CollisionShape2D.
- The Area2D has a connected input_event signal.
- A CanvasLayer with a ColorRect, where the CanvasLayer has an index of -1, placing the ColorRect behind the sprite.

**Expected behavior:**
The input_event on the Area2D should trigger for all mouse events, regardless of the presence of the ColorRect in the CanvasLayer.

**Current behavior:**
When the ColorRect is added to the scene:

- Left and right mouse button click events are not registered by the Area2D input_event, seemingly blocked or absorbed by the ColorRect.
- However, mouse wheel events are still registered as expected.

**Workaround:**
Setting the Mouse Filter property of the ColorRect to Pass allows the Area2D input_event to receive left and right mouse button events.

### Steps to reproduce

Create a scene with a Sprite that includes Area2D and CollisionShape2D.
Connect an input_event signal on the Area2D.
Add a CanvasLayer with an index of -1.
Place a ColorRect in the CanvasLayer.
Observe that left and right mouse click events on the Sprite do not trigger the input_event, while mouse wheel events do.

### Minimal reproduction project

[area2dButtonEvent.zip](https://github.com/godotengine/godot/files/13328340/area2dButtonEvent.zip)

KoBeWi:
It's because of `mouse_force_pass_scroll_events` property.

p10tr3k:
You're correct, however, this behavior seems somewhat counterintuitive. If the CanvasLayer's index is set to a lower value, one would logically expect that the ColorRect shouldn't act as if it's positioned above the Area2D. This inconsistency in layering and event handling merits further investigation.

Deozaan:
Sorry, I'm not currently able to check at my computer, but I'm pretty sure I've seen a warning in the inspector recently when editing something's z index, saying it only affects render order, not input/mouse event order. So I think this is expected behavior.

Daniel-The-Fox:
[...]

PR #87255, opened 2024-01-16
Add color in built-in documentation for overridden properties

Mickeon:
Should mitigate https://github.com/godotengine/godot/issues/84796 to an extent, and perhaps supersedes https://github.com/godotengine/godot/pull/84939?

The name is pretty self-explanatory.
It uses the warning color. This is used pretty consistently for other "inheritance"-related concepts in the editor, so I think it makes sense.

| . | . |
| --- |-- |
| image | image

I was also asked for a bit of padding between overridden and normal properties, although... I still don't like the way it looks, but hey, it works.

### DIFF
--- a/editor/editor_help.cpp
+++ b/editor/editor_help.cpp
@@ -146,6 +146,7 @@ void EditorHelp::_update_theme_item_cache() {
 	theme_cache.value_color = get_theme_color(SNAME("value_color"), SNAME("EditorHelp"));
 	theme_cache.qualifier_color = get_theme_color(SNAME("qualifier_color"), SNAME("EditorHelp"));
 	theme_cache.type_color = get_theme_color(SNAME("type_color"), SNAME("EditorHelp"));
+	theme_cache.override_color = get_theme_color(SNAME("override_color"), SNAME("EditorHelp"));
 
 	theme_cache.doc_font = get_theme_font(SNAME("doc"), EditorStringName(EditorFonts));
 	theme_cache.doc_bold_font = get_theme_font(SNAME("doc_bold"), EditorStringName(EditorFonts));
@@ -995,10 +996,29 @@ void EditorHelp::_update_doc() {
 		class_desc->push_table(4);
 		class_desc->set_table_column_expand(1, true);
 
+		cd.properties.sort_custom<PropertyCompare>();
+
+		bool is_generating_overridden_properties = true; // Set to false as soon as we encounter a non-overridden property.
+		bool overridden_property_exists = false;
+
 		for (int i = 0; i < cd.properties.size(); i++) {
 			// Ignore undocumented private.
 			if (cd.properties[i].name.begins_with("_") && cd.properties[i].description.strip_edges().is_empty()) {
 				continue;
+			}
+			if (is_generating_overridden_properties && !cd.properties[i].overridden) {
+				is_generating_overridden_properties = false;
+				// No need for the extra spacing when there's no overridden property.
+				if (overridden_property_exists) {
+					class_desc->push_cell();
+					class_desc->pop();
+					class_desc->push_cell();
+					class_desc->pop();
+					class_desc->push_cell();
+					class_desc->pop();
+					class_desc->push_cell();
+					class_desc->pop();
+				}
 			}
 			property_line[cd.properties[i].name] = class_desc->get_paragraph_count() - 2; //gets overridden if description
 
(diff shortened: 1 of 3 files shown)


Mickeon:
The test failed because the generated class reference would need to be updated to move the overridden properties above all else. In my opinion, the sorting of this PR does makes sense, but that's _a lot_ of multiple, small changes that would need to be done, and I'm not sure if this PR is desired just yet as is.

AThousandShips:
I think this should only be on the display side, unless they're put in a separate tag I don't think we should do complicated sorting like this

YuriSizov:
> I think this should only be on the display side, unless they're put in a separate tag I don't think we should do complicated sorting like this, you should be able to iterate over them twice when displaying, slightly unnecessary processing but the most trivial method IMO

Yeah, I agree with this. This change shouldn't require precise ordering in the XML files. It should just be a visual enhancement.

Mickeon:
Yyeah, the thing that peeved me most was the "unnecessary" second sorting pass, but if that is tolerable, so be it. Done.

Mickeon:
Yyeah, the thing that peeved me most was the "unnecessary" second sorting pass, but if that is tolerable, so be it. Done.

Mickeon:
Rebased after minor conflict.

editor/editor_help.cpp
```diff
@@ -1054,25 +1073,27 @@ void EditorHelp::_update_doc() {
 			_push_code_font();
 
 			if (!cd.properties[i].default_value.is_empty()) {
-				class_desc->push_color(theme_cache.symbol_color);
 				if (cd.properties[i].overridden) {
+					class_desc->push_color(get_theme_color(SNAME("warning_color"), EditorStringName(Editor)));
```
AThousandShips, line 1077: I'd suggest adding this as a separate theme cache entry defaulting to the same color

editor/editor_help.h
```diff
@@ -198,6 +198,16 @@ class EditorHelp : public VBoxContainer {
 	static void _gen_extensions_docs();
 	static void _compute_doc_version_hash();
 
+	struct PropertyCompare {
+		_FORCE_INLINE_ bool operator()(const DocData::PropertyDoc &l, const DocData::PropertyDoc &r) const {
+			// Sort overridden properties above all else.
+			if (l.overridden == r.overridden) {
+				return l.name.naturalcasecmp_to(r.name) < 0;
+			}
+			return l.overridden > r.overridden;
```
AThousandShips, line 207:
```suggestion
		_FORCE_INLINE_ bool operator()(const DocData::PropertyDoc &p_l, const DocData::PropertyDoc &p_r) const {
			// Sort overridden properties above all else.
			if (p_l.overridden == p_r.overridden) {
				return p_l.name.naturalcasecmp_to(p_r.name) < 0;
			}
			return p_l.overridden;
```
I'd say just checking the value is sufficient (and less confusing)

AThousandShips (approved):
LGTM, with some suggestions

editor/editor_help.cpp
Mickeon: Overridden properties rarely come into play, so I felt like doing it was a bit overkill. But it can be done.

AThousandShips: I'm just not sure the specific color here, using `warning_color` is fully reliably readable on different themes

editor/editor_help.h
Mickeon: So long as it achieves the same thing, sure, absolutely

AThousandShips: If both are `true` or `false` it enters the first part, otherwise it enters the second, and then it should be sorted first if `p_l` is overridden

editor/editor_help.cpp
Mickeon: I'm not sure I understand what you are suggestions. There have been some issues with the light theme and the warning color, but they're unrelated to the PR itself.

AThousandShips: I mean that like the `value_color` this should have a dedicated configurable color, specific for this

Also think it doesn't work quite well on the light theme:
image

Mickeon: Yyyyeah. For light theme, this would inherently be fixed by https://github.com/godotengine/godot/pull/88058

Meanwhile, I updated the PR to use a customizable `override_color`.

KoBeWi (approved):
Implementation looks ok.

PR #56420, opened 2022-01-02
Automatically create folder in project manager create and import

nathanfranke:
### Edit:

~~2023-09-07 - I am busy, but once I have time this will be my top priority PR.~~
~~2023-10-17 - I am working on this slowly :slightly_smiling_face:~~
2023-10-18 - Rebased and happy, but should wait for 4.3 :smiley: 

---

Supersedes https://github.com/godotengine/godot/pull/49602
Fixes https://github.com/godotengine/godot/issues/60332
Fixes https://github.com/godotengine/godot/issues/74678
Resolves https://github.com/godotengine/godot-proposals/issues/4749
Resolves https://github.com/godotengine/godot-proposals/issues/2754
Partly r3solves https://github.com/godotengine/godot-proposals/issues/7978

### Behavior:

- When create_dir is enabled, browse selects the containing/parent directory, project name is appended.
- When create_dir is disabled, whether before or after browsing, browse selects the exact project folder.
- Directory naming convention is configurable. `kebab-case` (default), `snake_case`, etc.

### Potential improvements:

- ~~Behavior of `capitalize()` means `test5` turns into `test-5/`.~~ Fixed, now `test5` --> `test5/`
- Persist/remember if `create_dir` is pressed. Would we need an editor setting?
- Maybe bug: horizontal scroll position of target path resets on change

Video:

https://user-images.githubusercontent.com/14253836/219931962-029b1d8f-21aa-4a68-91ba-320601b0f499.mp4

Creation:
image

Import:
image

PR #87266, opened 2024-01-16
Reorganize project manager code

YuriSizov:
### Highlights

- Extract individual components to their own files.
- Improve order of declarations and definitions within those files.
- Some insignificant naming cleanup.

Note that `ProjectDialog` is only extracted without any further touchup, as there is an upcoming PR touching on it, https://github.com/godotengine/godot/pull/56420. I wanted to wait for @nathanfranke to finish his first, but it's been a few weeks, so I settled for making the changes as simple to adapt to as possible.

There are going to be some more exciting PRs based on this one in the near future.

### DIFF
--- a/editor/SCsub
+++ b/editor/SCsub
@@ -113,6 +113,7 @@ if env.editor_build:
     SConscript("icons/SCsub")
     SConscript("import/SCsub")
     SConscript("plugins/SCsub")
+    SConscript("project_manager/SCsub")
     SConscript("themes/SCsub")
 
     lib = env.add_library("editor", env.editor_sources)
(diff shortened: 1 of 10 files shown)


akien-mga (approved):
Looks great!

In a follow-up, we could consider moving `project_manager.cpp` to the subfolder, even though the repeated name is a bit awkward.

On the other hand, we have this pattern in other platforms of having a subfolder with a similar name as a top-level class, e.g. `servers/rendering_server.h`/`servers/rendering/*`, so maybe it's fine as is.

YuriSizov:
> In a follow-up, we could consider moving `project_manager.cpp` to the subfolder, even though the repeated name is a bit awkward.

I did consider it, but since it's the entry point, same as `editor_node.h/cpp`, I felt it made more sense at the `editor/` level. Though this can change if/when I do the entrypoint reorganization that we've discussed. So maybe then!

PR #71578, opened 2023-01-17
Fix recursive resource local to scene (array/dictionary properties on resources)

RedMser:
Related to #71243
Fixes #71406
Fixes #85614

Any resource that contains arrays or dictionaries with local to scene resources will now be duplicated and configured correctly. This also includes built-in nodes, such as when marking a mesh's surface materials as local to scene.

A side effect of this change is that any array and dictionary property is now duplicated for local-to-scene instances.

What **isn't handled** in this PR is the case where a **scene's node** has an exported array/dictionary property containing local to scene resources. That code has a lot more special cases that I don't fully understand yet, but the solution would look largely similar.

---

NOTE: Using resources as dictionary keys does not work with the inspector (they get saved as `null` into the scene file), so this code branch was not tested.

PR #87268, opened 2024-01-16
Fix `resource_local_to_scene` in arrays and dictionaries

### LINKED ISSUE #71243: Resource local to scene is ignored for arrays and dictionaries (including surface materials)
### Godot version

Tested again on 4.2 beta 2

### System information

Windows 10

### Issue description

Exporting a resource that is marked as local to scene works as expected, both when exposed on a node and on a sub-resource.

But when the resource is a part of an array or dictionary, local to scene no longer works. The resource instances are shared.

### Steps to reproduce

- Create a node or resource which exports an array or dictionary which contains local-to-scene resources.
- Instantiate the scene multiple times.
- On the resource instances, observe `get_local_scene() == null` and modifications are shared between instances, despite `resource_local_to_scene == true`.

[CustomLocalToScene.zip](https://github.com/godotengine/godot/files/10439505/CustomLocalToScene.zip)

Two instances of a scene are included in the root. On the second instance, all field values are `*= 2`. All resources that are in arrays and dictionaries are modified on both instances, while the modifications should all only affect one of the two instances.

Following is the current output log showing `resource_local_to_scene`, the field's integer value and `get_local_scene()`:
<details>
<summary>Output Log</summary>

```
pre-modified first ---------
simple:
true	1	LocalToSceneTests1:<Node#26491224610>
array:
true	2	<Object#null>
true	3	<Object#null>
dict:
true	4	<Object#null>
sub:
true	5	LocalToSceneTests1:<Node#26491224610>
sub array:
true	6	<Object#null>
sub dict:
true	20	<Object#null>
check first --------- (should be same as above)
simple:
true	1	LocalToSceneTests1:<Node#26491224610>
array:
true	4	<Object#null>
true	6	<Object#null>
dict:
true	8	<Object#null>
sub:
true	5	LocalToSceneTests1:<Node#26491224610>
sub array:
true	12	<Object#null>
sub dict:
true	40	<Object#null>
check second --------- (should be *= 2 on each field)
simple:
true	2	LocalToSceneTests2:<Node#26575111315>
array:
true	4	<Object#null>
true	6	<Object#null>
dict:
true	8	<Object#null>
sub:
true	10	LocalToSceneTests2:<Node#26575111315>
sub array:
true	12	<Object#null>
sub dict:
true	40	<Object#null>
```
</details>

Wyxaldir:
Attempts to fix resource_local_to_scene in arrays, see #71243.

*Bugsquad edit:*
- Fixes #71243

### DIFF
--- a/scene/resources/packed_scene.cpp
+++ b/scene/resources/packed_scene.cpp
@@ -327,35 +327,16 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
 							//handle resources that are local to scene by duplicating them if needed
 							Ref<Resource> res = value;
 							if (res.is_valid()) {
-								if (res->is_local_to_scene()) {
-									if (n.instance >= 0) { // For the root node of a sub-scene, treat it as part of the sub-scene.
-										value = get_remap_resource(res, resources_local_to_sub_scene, node->get(snames[nprops[j].name]), node);
-									} else {
-										HashMap<Ref<Resource>, Ref<Resource>>::Iterator E = resources_local_to_scene.find(res);
-										Node *base = i == 0 ? node : ret_nodes[0];
-										if (E) {
-											value = E->value;
-										} else {
-											if (p_edit_state == GEN_EDIT_STATE_MAIN) {
-												//for the main scene, use the resource as is
-												res->configure_for_local_scene(base, resources_local_to_scene);
-												resources_local_to_scene[res] = res;
-											} else {
-												//for instances, a copy must be made
-												Ref<Resource> local_dupe = res->duplicate_for_local_scene(base, resources_local_to_scene);
-												resources_local_to_scene[res] = local_dupe;
-												value = local_dupe;
-											}
-										}
-									}
-									//must make a copy, because this res is local to scene
-								}
+								value = make_local_resource(value, n, resources_local_to_sub_scene, node, snames[nprops[j].name], resources_local_to_scene, i, ret_nodes, p_edit_state);
 							}
 						}
 						if (value.get_type() == Variant::ARRAY) {
 							Array set_array = value;
+							value = setup_resources_in_array(set_array, n, resources_local_to_sub_scene, node, snames[nprops[j].name], resources_local_to_scene, i, ret_nodes, p_edit_state);
+
 							bool is_get_valid = false;
 							Variant get_value = node->get(snames[nprops[j].name], &is_get_valid);
+
 							if (is_get_valid && get_value.get_type() == Variant::ARRAY) {
 								Array get_array = get_value;
 								if (!set_array.is_same_typed(get_array)) {
@@ -363,8 +344,26 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
 								}
 							}
 						}
-						if (p_edit_state == GEN_EDIT_STATE_INSTANCE && value.get_type() != Variant::OBJECT) {
-							value = value.duplicate(true); // Duplicate arrays and dictionaries for the editor
+						if (value.get_type() == Variant::DICTIONARY) {
+							Dictionary dictionary = value;
+							const Array keys = dictionary.keys();
+							const Array values = dictionary.values();
+
+							if (has_local_resource(values) || has_local_resource(keys)) {
+								Array duplicated_keys = keys.duplicate(true);
+								Array duplicated_values = values.duplicate(true);
+
+								duplicated_keys = setup_resources_in_array(duplicated_keys, n, resources_local_to_sub_scene, node, snames[nprops[j].name], resources_local_to_scene, i, ret_nodes, p_edit_state);
+								duplicated_values = setup_resources_in_array(duplicated_values, n, resources_local_to_sub_scene, node, snames[nprops[j].name], resources_local_to_scene, i, ret_nodes, p_edit_state);
+
+								dictionary.clear();
+
+								for (int dictionary_index = 0; dictionary_index < keys.size(); dictionary_index++) {
+									dictionary[duplicated_keys[dictionary_index]] = duplicated_values[dictionary_index];
+								}
+
+								value = dictionary;
+							}
 						}
 
 						bool set_valid = true;
@@ -378,6 +377,9 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
 
 						if (set_valid) {
 							node->set(snames[nprops[j].name], value, &valid);
+						}
+						if (p_edit_state == GEN_EDIT_STATE_INSTANCE && value.get_type() != Variant::OBJECT) {
+							value = value.duplicate(true); // Duplicate arrays and dictionaries for the editor.
 						}
 					}
 				}
@@ -566,6 +568,50 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
 	return ret_nodes[0];
 }
 
+Variant SceneState::make_local_resource(Variant &p_value, const SceneState::NodeData &p_node_data, HashMap<Ref<Resource>, Ref<Resource>> &p_resources_local_to_sub_scene, Node *p_node, const StringName p_sname, HashMap<Ref<Resource>, Ref<Resource>> &p_resources_local_to_scene, int p_i, Node **p_ret_nodes, SceneState::GenEditState p_edit_state) const {
+	Ref<Resource> res = p_value;
+	if (res.is_null() || !res->is_local_to_scene()) {
+		return p_value;
+	}
+
+	if (p_node_data.instance >= 0) { // For the root node of a sub-scene, treat it as part of the sub-scene.
+		return get_remap_resource(res, p_resources_local_to_sub_scene, p_node->get(p_sname), p_node);
+	} else {
+		HashMap<Ref<Resource>, Ref<Resource>>::Iterator E = p_resources_local_to_scene.find(res);
+		Node *base = p_i == 0 ? p_node : p_ret_nodes[0];
+		if (E) {
+			return E->value;
+		} else if (p_edit_state == GEN_EDIT_STATE_MAIN) { // For the main scene, use the resource as is
+			res->configure_for_local_scene(base, p_resources_local_to_scene);
+			p_resources_local_to_scene[res] = res;
+			return res;
+		} else { // For instances, a copy must be made.
+			Ref<Resource> local_dupe = res->duplicate_for_local_scene(base, p_resources_local_to_scene);
+			p_resources_local_to_scene[res] = local_dupe;
+			return local_dupe;
+		}
+	}
+}
+
+Array SceneState::setup_resources_in_array(Array &p_array_to_scan, const SceneState::NodeData &p_n, HashMap<Ref<Resource>, Ref<Resource>> &p_resources_local_to_sub_scene, Node *p_node, const StringName p_sname, HashMap<Ref<Resource>, Ref<Resource>> &p_resources_local_to_scene, int p_i, Node **p_ret_nodes, SceneState::GenEditState p_edit_state) const {
+	for (int i = 0; i < p_array_to_scan.size(); i++) {
+		if (p_array_to_scan[i].get_type() == Variant::OBJECT) {
+			p_array_to_scan[i] = make_local_resource(p_array_to_scan[i], p_n, p_resources_local_to_sub_scene, p_node, p_sname, p_resources_local_to_scene, p_i, p_ret_nodes, p_edit_state);
+		}
+	}
+	return p_array_to_scan;
+}
+
+bool SceneState::has_local_resource(const Array &p_array) const {
+	for (int i = 0; i < p_array.size(); i++) {
+		Ref<Resource> res = p_array[i];
+		if (res.is_valid() && res->is_local_to_scene()) {
+			return true;
+		}
+	}
+	return false;
+}
+
 static int _nm_get_string(const String &p_string, HashMap<StringName, int> &name_map) {
 	if (name_map.has(p_string)) {
 		return name_map[p_string];
@@ -730,8 +776,18 @@ Error SceneState::_parse_node(Node *p_owner, Node *p_node, int p_parent_idx, Has
 		if (!pinned_props.has(name)) {
 			bool is_valid_default = false;
 			Variant default_value = PropertyUtils::get_property_default_value(p_node, name, &is_valid_default, &states_stack, true);
+
 			if (is_valid_default && !PropertyUtils::is_property_value_different(value, default_value)) {
-				continue;
+				if (value.get_type() == Variant::ARRAY && has_local_resource(value)) {
+					// Save anyway
+				} else if (value.get_type() == Variant::DICTIONARY) {
+					Dictionary dictionary = value;
+					if (!has_local_resource(dictionary.values()) && !has_local_resource(dictionary.keys())) {
+						continue;
+					}
+				} else {
+					continue;
+				}
 			}
 		}
 
--- a/scene/resources/packed_scene.h
+++ b/scene/resources/packed_scene.h
@@ -157,6 +157,10 @@ public:
 	bool can_instantiate() const;
 	Node *instantiate(GenEditState p_edit_state) const;
 
+	Array setup_resources_in_array(Array &array_to_scan, const SceneState::NodeData &n, HashMap<Ref<Resource>, Ref<Resource>> &resources_local_to_sub_scene, Node *node, const StringName sname, HashMap<Ref<Resource>, Ref<Resource>> &resources_local_to_scene, int i, Node **ret_nodes, SceneState::GenEditState p_edit_state) const;
+	Variant make_local_resource(Variant &value, const SceneState::NodeData &p_node_data, HashMap<Ref<Resource>, Ref<Resource>> &p_resources_local_to_sub_scene, Node *p_node, const StringName p_sname, HashMap<Ref<Resource>, Ref<Resource>> &p_resources_local_to_scene, int p_i, Node **p_ret_nodes, SceneState::GenEditState p_edit_state) const;
+	bool has_local_resource(const Array &p_array) const;
+
 	Ref<SceneState> get_base_scene_state() const;
 
 	void update_instance_resource(String p_path, Ref<PackedScene> p_packed_scene);


scene/resources/packed_scene.cpp
```diff
@@ -356,6 +356,38 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
 							Array set_array = value;
 							bool is_get_valid = false;
 							Variant get_value = node->get(snames[nprops[j].name], &is_get_valid);
+
+							// For every item in the array, we'll check if it's a resource and if it needs to be duplicated.
+							// The copying is just a copy of the code above (see if (value.get_type() == Variant::OBJECT)).
+							for (int ai = 0; ai < set_array.size(); ai++)
+							{
+								Ref<Resource> res = set_array[ai];
+								if (!res.is_valid()) continue;
```
AThousandShips, line 365:
```suggestion
								if (!res.is_valid()) {
									continue;
								}
```
Always use brackets

scene/resources/packed_scene.cpp
```diff
@@ -356,6 +356,38 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
...
+								if (!res->is_local_to_scene()) continue;
```
AThousandShips, line 366:
```suggestion
								if (!res->is_local_to_scene()) {
									continue;
								}
```

scene/resources/packed_scene.cpp
```diff
@@ -356,6 +356,38 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
 							Array set_array = value;
 							bool is_get_valid = false;
 							Variant get_value = node->get(snames[nprops[j].name], &is_get_valid);
+
+							// For every item in the array, we'll check if it's a resource and if it needs to be duplicated.
+							// The copying is just a copy of the code above (see if (value.get_type() == Variant::OBJECT)).
+							for (int ai = 0; ai < set_array.size(); ai++)
```
AThousandShips, line 362:
```suggestion
							for (int array_i = 0; array_i < set_array.size(); array_i++)
```
More descriptive

scene/resources/packed_scene.cpp
```diff
@@ -356,6 +356,38 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
...
+							{
```
AThousandShips, line 363: Should always be with the `for`

scene/resources/packed_scene.cpp
```diff
@@ -356,6 +356,38 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
...
+								Ref<Resource> res = set_array[ai];
+								if (!res.is_valid()) continue;
+								if (!res->is_local_to_scene()) continue;
+								if (n.instance >= 0) { // For the root node of a sub-scene, treat it as part of the sub-scene.
+									set_array[ai] = get_remap_resource(res, resources_local_to_sub_scene, node->get(snames[nprops[j].name]), node);
+								}
+								else {
```
AThousandShips, line 370: Should always be on the same line as `}`, automatic tests will complain about this

Wyxaldir: Sorry about all that, thanks for the reviews though!

AThousandShips: No worries! It takes time to learn 🙂

RedMser:
Thanks for tackling this! ❤️ 

However, it doesn't seem to fix the issue in the current state. Running the MRP with your PR applied, there still are unduplicated resources, leaving the debug output the same as in the issue description (highlighted):

image

Your new code seems to be called and duplicating properly, so I'm not sure what's going wrong.

---

Also, you should likely remove the `fixes #...` from the description, since it does not handle dictionaries. The issue would automatically close on merge if you keep the wording like this.

RedMser:
Debugging...

I tried adding `print_line(vformat("%s from %s to %s", node->get_instance_id(), res->get_instance_id(), local_dupe->get_instance_id()));` below the line `Ref<Resource> local_dupe`

I then changed the MRP to log `get_instance_id()` and gave each a unique letter in the output:

```
// [node] from [res] to [local_dupe]
[X] from [A] to [B]
[X] from [C] to [D]
[Y] from [B] to [E]
[Y] from [D] to [F]
pre-modified first ---------
simple:
true	1	[G]
array:
true	2	[E]
true	3	[F]
dict:
true	4	[H]
sub:
true	5	[I]
sub array:
true	6	[J]
sub dict:
true	20	[K]
check first --------- (should be same as above)
simple:
true	1	[G]
array:
true	4	[E]
true	6	[F]
dict:
true	8	[H]
sub:
true	5	[I]
sub array:
true	6	[J]
sub dict:
true	20	[K]
check second --------- (should be *= 2 on each field)
simple:
true	2	[L]
array:
true	4	[E]
true	6	[F]
dict:
true	8	[H]
sub:
true	10	[M]
sub array:
true	12	[N]
sub dict:
true	40	[O]
```

Note how [Y]'s scene duplicates seem to re-use the resources of [X].

Wyxaldir:
That's strange, as I thought it was working here. I have another test project where this fixed my issues, and the logs for the MRP seem right on my end.

``` 
pre-modified first ---------
simple:
true	1	LocalToSceneTests1:<Node#26457670861>
array:
true	2	LocalToSceneTests1:<Node#26457670861>
true	3	LocalToSceneTests1:<Node#26457670861>
dict:
true	4	<Object#null>
sub:
true	5	LocalToSceneTests1:<Node#26457670861>
sub array:
true	6	LocalToSceneTests1:<Node#26457670861>
sub dict:
true	20	LocalToSceneTests1:<Node#26457670861>
check first --------- (should be same as above)
simple:
true	1	LocalToSceneTests1:<Node#26457670861>
array:
true	2	LocalToSceneTests1:<Node#26457670861>
true	3	LocalToSceneTests1:<Node#26457670861>
dict:
true	8	<Object#null>
sub:
true	5	LocalToSceneTests1:<Node#26457670861>
sub array:
true	6	LocalToSceneTests1:<Node#26457670861>
sub dict:
true	20	LocalToSceneTests1:<Node#26457670861>
check second --------- (should be *= 2 on each field)
simple:
true	2	LocalToSceneTests2:<Node#26608665818>
array:
true	4	LocalToSceneTests2:<Node#26608665818>
true	6	LocalToSceneTests2:<Node#26608665818>
dict:
true	8	<Object#null>
sub:
true	10	LocalToSceneTests2:<Node#26608665818>
sub array:
true	12	LocalToSceneTests2:<Node#26608665818>
sub dict:
true	40	LocalToSceneTests2:<Node#26608665818>```

**Comment by RedMser** [comment]
Hmm I redownloaded the MRP and it indeed works as expected now! :tada:
Checked the diff between the two and this change stands out to me:

image

(left is working, right is broken)

Not sure what happened, or if it's something I unknowingly did 😅

**Comment by akien-mga** [comment]

File: scene/resources/packed_scene.cpp [hunk 6]
```diff
@@ -356,6 +356,38 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
...
+											//for the main scene, use the resource as is
+											res->configure_for_local_scene(base, resources_local_to_scene);
+											resources_local_to_scene[res] = res;
+										} else {
+											//for instances, a copy must be made
+											Ref<Resource> local_dupe = res->duplicate_for_local_scene(base, resources_local_to_scene);
+											resources_local_to_scene[res] = local_dupe;
+											set_array[ai] = local_dupe;
+										}
+									}
```
AThousandShips, line 388:
```suggestion
									} else if (p_edit_state == GEN_EDIT_STATE_MAIN) {
										// For the main scene, use the resource as is.
										res->configure_for_local_scene(base, resources_local_to_scene);
										resources_local_to_scene[res] = res;
									} else {
										// For instances, a copy must be made.
										Ref<Resource> local_dupe = res->duplicate_for_local_scene(base, resources_local_to_scene);
										resources_local_to_scene[res] = local_dupe;
										set_array[ai] = local_dupe;
									}
```
Unnecessary nesting

scene/resources/packed_scene.cpp
```diff
@@ -356,6 +356,36 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
 							Array set_array = value;
 							bool is_get_valid = false;
 							Variant get_value = node->get(snames[nprops[j].name], &is_get_valid);
+
+							// For every item in the array, we'll check if it's a resource and if it needs to be duplicated.
+							// The copying is just a copy of the code above (see if (value.get_type() == Variant::OBJECT)).
+							for (int ai = 0; ai < set_array.size(); ai++) {
+								Ref<Resource> res = set_array[ai];
+								if (!res.is_valid()) {
```
KoBeWi, line 364:
```suggestion
								if (res.is_null()) {
```

KoBeWi:
I'm getting a deja vu here.
This only handles resources in arrays, but arrays can be nested. Then there are also Dictionaries, which can use Resources both for keys and values. And they can be nested too. Also I'm not sure if nested resources are handled or if it's for nodes only.
Ultimately we'd end up with a unified method that just scans everything recursively. I remember we have at least 3 of them.

Although that's probably all corner cases. The issue mentions also Dictionaries, so I think this PR could also handle them. It doesn't need to be recursive for now, but it would be nice if the copied code could be moved to some method.

scene/resources/packed_scene.cpp
```diff
@@ -356,6 +356,36 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
 							Array set_array = value;
 							bool is_get_valid = false;
 							Variant get_value = node->get(snames[nprops[j].name], &is_get_valid);
+
+							// For every item in the array, we'll check if it's a resource and if it needs to be duplicated.
+							// The copying is just a copy of the code above (see if (value.get_type() == Variant::OBJECT)).
+							for (int ai = 0; ai < set_array.size(); ai++) {
+								Ref<Resource> res = set_array[ai];
```
KoBeWi, line 363: It's safer to check if type is OBJECT, like above.

Wyxaldir:
> btw I also made a simpler reproduction project: [CustomLocalToScene.zip](https://github.com/godotengine/godot/files/14044234/CustomLocalToScene.zip) 
If all shapes have different size it means the bug is fixed.

This zip is missing the script I think.

I'll try to implement the method you outlined. As for dictionaries, I'm not certain how to tackle it, should I check for object and then dictionary? (pseudo code below) 
```
if (value.get_type() == Variant::OBJECT) {
    if (value.get_type() == Variant::DICTIONARY) {
        // Check for resources in keys and values
    }
}
```

And yeah, the nesting is certainly an issue, I'll see if I can deal with it too while I'm at it.

RedMser:
> As for dictionaries, I'm not certain how to tackle it, should I check for object and then dictionary?

You can get inspired by my PR that tackled the other half of the issue: https://github.com/godotengine/godot/pull/71578
There's a switch-case which handles all cases (including nesting, arrays, dictionary keys and values).

KoBeWi:
> This zip is missing the script I think.

Sorry, here's a fixed one:
[CustomLocalTooScene.zip](https://github.com/godotengine/godot/files/14052632/CustomLocalTooScene.zip)

>  As for dictionaries, I'm not certain how to tackle it, should I check for object and then dictionary?

Dictionary is not Object.

Also I wouldn't really bother for recursive checking, as it often caused performance problems. For now handling only top level properties is fine I think.

Wyxaldir:
The issue so far is that the resource is always duplicated now, even if it already was. What this does (to my understanding) is that the values get reset every time you open the scene.
For example, I open a scene and set the radius of a CircleShape2D to 125. It is independent from the parent scene, great! When I reopen the scene however it is duplicated again and the radius is reset back to the original value from the parent scene.

```cpp
void SceneState::setup_local_resource(Ref<Resource>& res, Array& origin_array, const int origin_array_index, const SceneState::NodeData& n, HashMap<Ref<Resource>, Ref<Resource>>& resources_local_to_sub_scene, Node* node, const StringName sname, HashMap<Ref<Resource>, Ref<Resource>>& resources_local_to_scene, int i, Node** ret_nodes, SceneState::GenEditState p_edit_state) const
{
	if (!res->is_local_to_scene()) {
		printf("Resource was not local.\n");
		return;
	}
	if (n.instance >= 0) { // For the root node of a sub-scene, treat it as part of the sub-scene.
		//origin_array[origin_array_index] = get_remap_resource(res, resources_local_to_sub_scene, node->get(snames[nprops[j].name]), node);
		origin_array[origin_array_index] = get_remap_resource(res, resources_local_to_sub_scene, node->get(sname), node);
		printf("Remapping.\n");
	} else {
		HashMap<Ref<Resource>, Ref<Resource>>::Iterator E = resources_local_to_scene.find(res);
		Node* base = i == 0 ? node : ret_nodes[0];
		if (E) {
			origin_array[origin_array_index] = E->value;
			printf("Found original, no change.\n");
		} else if (p_edit_state == GEN_EDIT_STATE_MAIN) {
			// For the main scene, use the resource as is.
			//res->configure_for_local_scene(base, resources_local_to_scene);
			res->duplicate_for_local_scene(base, resources_local_to_scene);
			resources_local_to_scene[res] = res;
			printf("Main scene.\n");
		} else {
			// For instances, a copy must be made.
			Ref<Resource> local_dupe = res->duplicate_for_local_scene(base, resources_local_to_scene);
			resources_local_to_scene[res] = local_dupe;
			origin_array[origin_array_index] = local_dupe;
			printf("Duplicate for local scene.\n");
		}
	}
}
```

My guess is that somehow it's not finding the resource when it runs this for some reason.
```cpp
HashMap<Ref<Resource>, Ref<Resource>>::Iterator E = resources_local_to_scene.find(res);
Node* base = i == 0 ? node : ret_nodes[0];
```

Your test scene works btw, when the nodes are modified at runtime they are properly duplicated, so that's good.

Wyxaldir:
I'm so close, but I can't put together the final line(s) needed. The reason @KoBeWi 's method doesn't work is because the final variant needs to be assigned to value, but we were just assigning it to the Ref<Resource> previously. I made a version that returns a Ref<Resource> and assigns it to value and it works now. 

I imagine it's the same thing with the array, but despite attempting to assign value to the modified array I still run into the same issues as above. I'm not sure if Arrays have a special way of being assigned though that I might be missing however.

This is my version of your method now btw: 
```cpp
Ref<Resource> SceneState::make_local_resource(Variant& r_value, const SceneState::NodeData& p_node_data, HashMap<Ref<Resource>, Ref<Resource>>& p_resources_local_to_sub_scene, Node* p_node, const StringName p_sname, HashMap<Ref<Resource>, Ref<Resource>>& p_resources_local_to_scene, int p_i, Node** p_ret_nodes, SceneState::GenEditState p_edit_state) const
{
	Ref<Resource> res = r_value;
	if (res.is_null() || !res->is_local_to_scene()) {
		return r_value;
	}

	if (p_node_data.instance >= 0) { // For the root node of a sub-scene, treat it as part of the sub-scene.
		return get_remap_resource(res, p_resources_local_to_sub_scene, p_node->get(p_sname), p_node);
	} else {
		HashMap<Ref<Resource>, Ref<Resource>>::Iterator E = p_resources_local_to_scene.find(res);
		Node* base = p_i == 0 ? p_node : p_ret_nodes[0];
		if (E) {
			return E->value;
		} else if (p_edit_state == GEN_EDIT_STATE_MAIN) { // For the main scene, use the resource as is
			res->configure_for_local_scene(base, p_resources_local_to_scene);
			p_resources_local_to_scene[res] = res;
			return res;
		} else { // For instances, a copy must be made.
			Ref<Resource> local_dupe = res->duplicate_for_local_scene(base, p_resources_local_to_scene);
			p_resources_local_to_scene[res] = local_dupe;
			return local_dupe;
		}
	}
}
```

Wyxaldir:
Getting more specific: the issue seems to be that the default value for the array is not properly referencing the value from the parent scene:
```cpp
if (!pinned_props.has(name)) {
    bool is_valid_default = false;
    // FIXME: the default_value is the current value rather than the value in the parent scene
    Variant default_value = PropertyUtils::get_property_default_value(p_node, name, &is_valid_default, &states_stack, true);
        if (is_valid_default && !PropertyUtils::is_property_value_different(value, default_value)) { 
            continue;
        }
}
```

I _believe_ the error is in the PropertyUtils::get_node_states_stack method, but this is getting way over my head unfortunately.

AThousandShips:
Is this intended to be a work in progress currently? If so you might want to convert it to a draft 🙂

Wyxaldir:
Sorry for needing the workflow to be reset so often, I made sure to check my warnings this time so hopefully the checks pass this time!

KoBeWi:
Looks better now, but something is wrong with Dictionaries. When you have both key and value local to scene, they double when loading the scene

Also you should remove the commented out code.

scene/resources/packed_scene.cpp
```diff
@@ -327,44 +327,38 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
...
 								if (!set_array.is_same_typed(get_array)) {
 									value = Array(set_array, get_array.get_typed_builtin(), get_array.get_typed_class_name(), get_array.get_typed_script());
 								}
 							}
 						}
-						if (p_edit_state == GEN_EDIT_STATE_INSTANCE && value.get_type() != Variant::OBJECT) {
-							value = value.duplicate(true); // Duplicate arrays and dictionaries for the editor
+						if (value.get_type() == Variant::DICTIONARY) {
+							Dictionary dictionary = value;
+							if (has_local_resource(dictionary.keys()) || has_local_resource(dictionary.values())) {
```
KoBeWi, line 349: Given how `keys()` and `values()` are implemented, it's better to store them in a variable, to avoid calling these methods more than once.

scene/resources/packed_scene.cpp
```diff
@@ -730,8 +819,23 @@ Error SceneState::_parse_node(Node *p_owner, Node *p_node, int p_parent_idx, Has
...
+
 			if (is_valid_default && !PropertyUtils::is_property_value_different(value, default_value)) {
-				continue;
+				// FIXME: This saves local_to_scene resources NO MATTER WHAT, even if their value hasn't changed.
+				// This can cause .tscn files to get larger than needed, so the default value check should take into account
+				// local resources somehow.
+
+				// If the array/dictionary contains a resource which is local_to_scene, we need to save it anyway.
+				if (value.get_type() == Variant::ARRAY && has_local_resource(value)) {
+					// Save anyway
```
KoBeWi, line 830:
```suggestion
					// Save anyway.
```

scene/resources/packed_scene.cpp
```diff
@@ -730,8 +819,23 @@ Error SceneState::_parse_node(Node *p_owner, Node *p_node, int p_parent_idx, Has
 		if (!pinned_props.has(name)) {
 			bool is_valid_default = false;
 			Variant default_value = PropertyUtils::get_property_default_value(p_node, name, &is_valid_default, &states_stack, true);
+
 			if (is_valid_default && !PropertyUtils::is_property_value_different(value, default_value)) {
-				continue;
+				// FIXME: This saves local_to_scene resources NO MATTER WHAT, even if their value hasn't changed.
+				// This can cause .tscn files to get larger than needed, so the default value check should take into account
+				// local resources somehow.
```
KoBeWi, line 826: This is expected tbh. If a resource is local to scene, it gets duplicated, which makes the value modified.

scene/resources/packed_scene.cpp
```diff
@@ -327,44 +327,38 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
...
+						if (value.get_type() == Variant::DICTIONARY) {
+							Dictionary dictionary = value;
+							if (has_local_resource(dictionary.keys()) || has_local_resource(dictionary.values())) {
+								Array keys = dictionary.keys().duplicate(true);
+								Array values = dictionary.values().duplicate(true);
+
+								keys = setup_resources_in_array(keys, n, resources_local_to_sub_scene, node, snames[nprops[j].name], resources_local_to_scene, i, ret_nodes, p_edit_state);
+								values = setup_resources_in_array(values, n, resources_local_to_sub_scene, node, snames[nprops[j].name], resources_local_to_scene, i, ret_nodes, p_edit_state);
+
+								for (int dictionary_index = 0; dictionary_index < keys.size(); dictionary_index++) {
```
KoBeWi, line 356: You need to clear the Dictionary before re-assigning keys. This is likely what caused the double content bug I mentioned (I only tested with local resources, so it duplicated everything).

scene/resources/packed_scene.cpp
```diff
@@ -327,44 +327,43 @@ Node *SceneState::instantiate(GenEditState p_edit_state) const {
...
 									value = Array(set_array, get_array.get_typed_builtin(), get_array.get_typed_class_name(), get_array.get_typed_script());
 								}
 							}
 						}
-						if (p_edit_state == GEN_EDIT_STATE_INSTANCE && value.get_type() != Variant::OBJECT) {
-							value = value.duplicate(true); // Duplicate arrays and dictionaries for the editor
+						if (value.get_type() == Variant::DICTIONARY) {
+							Dictionary dictionary = value;
+							Array keys = dictionary.keys();
+							Array values = dictionary.values();
```
KoBeWi, line 350:
```suggestion
							const Array keys = dictionary.keys();
							const Array values = dictionary.values();
```

KoBeWi (approved):
Works correctly now. You still need to remove the commented out code as I mentioned.

akien-mga:

PR #87270, opened 2024-01-16
Test global transform when adding detached tree

Sauermann:
Global transform should be correct after attaching a detached tree.

Add unittest for pr #86841.


akien-mga (approved):

akien-mga:
Thanks!

PR #63368, opened 2022-07-23
Code quality: Fix header guards consistency

akien-mga:
Adds `header_guards.sh` bash script, used in CI to validate future
changes. Can be run locally to fix invalid header guards.

Finally took the time to address this pet peeve of mine with a proper script.

The expected style is, for a file named `some_class.h`:
```cpp
// License block.

#ifndef SOME_CLASS_H
#define SOME_CLASS_H

// Includes.

// Class declaration(s).

#endif // SOME_CLASS_H
```

See https://docs.godotengine.org/en/latest/community/contributing/code_style_guidelines.html#header-includes

There are some exceptions for files which exist multiple time in the codebase where either a prefix or a suffix is added:
- Module `register_types.h`
- Platform `export/` and `api/` headers
- Renderer implementation for effects

Check `misc/scripts/header_guards.sh` for details.

Renames `servers/audio/effects/{eq,reverb}.*` to have a `_filter` suffix, as the names were a bit confusing next to the `audio_effect_{eq,reverb}.*` that use them.

Renames `modules/gdscript/language_server/lsp.hpp` to `godot_lsp.h`, this was the only Godot C++ header using `.hpp` extension outside thirdparty code.

PR #87271, opened 2024-01-16
[3.x] Move COREAUDIO_ENABLED check inside AUDIO_DRIVER_COREAUDIO_H block

halotroop2288:


drivers/coreaudio/audio_driver_coreaudio.h
```diff
@@ -120,6 +120,6 @@ class AudioDriverCoreAudio : public AudioDriver {
 	~AudioDriverCoreAudio();
 };
 
-#endif
-
 #endif // AUDIO_DRIVER_COREAUDIO_H
+
+#endif // COREAUDIO_ENABLED
```
AThousandShips, line 125: This should really be inside the header guard and not outside IMO, i.e. the ifdef should be moved inside instead

halotroop2288:
OK we can do it your way. It's done.

akien-mga:
Thanks for the contribution!

As I pointed out on https://github.com/godotengine/godot/pull/87272#issuecomment-1895379562, for the `3.x` branch we prefer not to do this kind of style fixes unless they actually solve a problem. It took a lot of effort to improve the code style consistency of the whole codebase in the `master` branch, and redoing all that work for `3.x` wouldn't be worth the effort at this stage.

The change in this commit is correct, but I know there are many other places with similar style inconsistencies which have no incidence on the functionality of the code, as I've fixed them all in the `master` branch. For this case specifically, you can have a look at #63368. As you can see, more than a thousand lines of code were impacted, so doing it file by file as started in this PR wouldn't be a good option. I wouldn't suggest trying to backport this PR to `3.x`, as again this would be more effort than worth (not only to do it, but for maintainers to review it and make sure there's effectively no change in behavior).

I included this change in #87272 because why not, but otherwise I'd advise against trying to tackle other such minor header guard name/comment inconsistencies if they're not actual bugs that change the code logic.

PR #87272, opened 2024-01-16
[3.x] Fix `#if *_ENABLED` inconsistencies, should check if defined

halotroop2288:
As discussed on RocketChat, these checks are probably written incorrectly.

### DIFF
--- a/drivers/coreaudio/audio_driver_coreaudio.h
+++ b/drivers/coreaudio/audio_driver_coreaudio.h
@@ -28,10 +28,10 @@
 /* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
 /**************************************************************************/
 
-#ifdef COREAUDIO_ENABLED
-
 #ifndef AUDIO_DRIVER_COREAUDIO_H
 #define AUDIO_DRIVER_COREAUDIO_H
+
+#ifdef COREAUDIO_ENABLED
 
 #include "servers/audio_server.h"
 
@@ -120,6 +120,6 @@ public:
 	~AudioDriverCoreAudio();
 };
 
-#endif
+#endif // COREAUDIO_ENABLED
 
 #endif // AUDIO_DRIVER_COREAUDIO_H
--- a/drivers/gles2/rasterizer_scene_gles2.cpp
+++ b/drivers/gles2/rasterizer_scene_gles2.cpp
@@ -2840,7 +2840,7 @@ void RasterizerSceneGLES2::_post_process(Environment *env, const CameraMatrix &p
 
 		glBindFramebuffer(GL_READ_FRAMEBUFFER, 0);
 		glBindFramebuffer(GL_DRAW_FRAMEBUFFER, 0);
-#elif IPHONE_ENABLED
+#elif defined(IPHONE_ENABLED)
 
 		glBindFramebuffer(GL_READ_FRAMEBUFFER, storage->frame.current_rt->multisample_fbo);
 		glBindFramebuffer(GL_DRAW_FRAMEBUFFER, next_buffer);
@@ -2848,7 +2848,7 @@ void RasterizerSceneGLES2::_post_process(Environment *env, const CameraMatrix &p
 
 		glBindFramebuffer(GL_READ_FRAMEBUFFER, 0);
 		glBindFramebuffer(GL_DRAW_FRAMEBUFFER, 0);
-#elif ANDROID_ENABLED
+#elif defined(ANDROID_ENABLED)
 
 		// In GLES2 Android Blit is not available, so just copy color texture manually
 		_copy_texture_to_buffer(storage->frame.current_rt->multisample_color, next_buffer);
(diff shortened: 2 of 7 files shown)


AThousandShips:

halotroop2288:
> Please squash your commits into one

There's a button for that. You don't have to make web contributors do all the work.

Squash and merge

AThousandShips:
We don't do that for various reasons, it's part of the workflow 🙂, see that link, we don't squash commits because it:
* Doesn't allow it *and* merge commits, and we want merge commits
* It resets the author verification and muddles the history

We're well aware of this feature and there's very good reasons we don't use it

halotroop2288:
It's squashed. Do you need anything else?

AThousandShips:
No it looks good otherwise 🙂

lawnjelly:
I didn't write this, but a quick test in godbolt suggests in the existing format:

If e.g. `IPHONE_ENABLED` is defined as 1 or true, then the code underneath is compiled.
If it is false, then the code underneath does not compile.

If it is undefined, the compiler will flag an error:

```
error: #elif with no expression
```

So by adding the `defined` the PR may have changed the logic (and maybe introduced bugs?), it will compile the code underneath whether or not the `ENABLED` is defined as true or false.

So this may need a little double checking if to be changed. Afaik the `ENABLED` seem to be set in the scons scripts, so it should not be set, but I wonder why it is compiling at all if e.g. `IPHONE_ENABLED` is undefined. :thinking: 

I have limited time to check this, but just wanted to flag this and make sure it had been tested.

akien-mga:
There's a difference between passing a define on the command line with `-DSOME_DEFINE` and defining it without value with `#define SOME_DEFINE`. The former defines it as `1`, and is equivalent to `-DSOME_DEFINE=1` or `#define SOME_DEFINE 1`. The latter defines it as empty and `#if SOME_DEFINE` will fail with `error: #if with no expression`.

Notably, this also applies to names which have _not_ been defined on the command line. If `SOME_DEFINE` doesn't exist, `#if SOME_DEFINE` actually compiles and is equivalent to `#if 0` (but `SOME_DEFINE` itself isn't `0`, not sure how the preprocessor sees it exactly).

That's why this code "works" e.g. on Android where only `ANDROID_ENABLED` is defined, without throwing a compile-time error on `#elif IPHONE_ENABLED` which comes before checking `#elif ANDROID_ENABLED`. If we defined `#define IPHONE_ENABLED` (without value), then it would fail.

Test script to confirm that:

<details>

```cpp
// Compile with: g++ main.cpp -DD -DE=1 -DF=0

#include <cstdio>

#define A
#define B 1
#define C 0
// D passed on command line with -DD
// E passed on command line with -DE=1
// F passed on command line with -DF=0
// no G

#define _STR(m_x) #m_x
#define _MKSTR(m_x) _STR(m_x)

int main() {
	printf("A: %s\n", _MKSTR(A));
	printf("B: %s\n", _MKSTR(B));
	printf("C: %s\n", _MKSTR(C));
	printf("D: %s\n", _MKSTR(D));
	printf("E: %s\n", _MKSTR(E));
	printf("F: %s\n", _MKSTR(F));
	printf("G: %s\n", _MKSTR(G));

/*
// Fails with: error: #if with no expression
#if A
	printf("A is true.\n");
#endif
*/
#if B
	printf("B is true.\n");
#endif
#if C
	printf("C is true.\n");
#endif
#if D
	printf("D is true.\n");
#endif
#if E
	printf("E is true.\n");
#endif
#if F
	printf("F is true.\n");
#endif
#if G
	printf("G is true.\n");
#endif
	return 0;
}
```

Prints:
```
A: 
B: 1
C: 0
D: 1
E: 1
F: 0
G: G
B is true.
D is true.
E is true.
```

</details>

All `..._ENABLED` defines are passed on the command line and thus pass both `#if ..._ENABLED` and `#if defined(..._ENABLED)` checks. Still, we shouldn't count on this and the de facto convention is that we consider them true *if they are defined*, precisely to avoid having these discussions over and over again :P

Numbers in the codebase speak for themselves (`master` branch):

```
$ rg -g'!thirdparty' "#if defined.*ENABLED" | wc -l
261
$ rg -g'!thirdparty' "#ifdef .*ENABLED" | wc -l
2432
$ rg -g'!thirdparty' "#elif defined.*ENABLED" | wc -l
14
```

```
rg -g'!thirdparty' "#if [^d\!].*ENABLED" | wc -l
8
$ rg -g'!thirdparty' "#elif [^d\!].*ENABLED" | wc -l
4
```

For that last one it's 7 occurrences in the `3.x` branch, which are the ones addressed by this PR. To me the code as is is clearly wrong, so this PR goes in the right direction.

akien-mga:
This PR should likely also assess the cases where `#if [^d\!].*ENABLED` and not just `#elif`, which have the same problem. That would fully solve the inconsistency.

And this all needs to be done for the `master` branch in priority over `3.x`. Code style fixes are mainly relevant for the branch which we aim to maintain for the next decade, which is the `master` branch. There are lots of style inconsistencies in the `3.x` branch which we opted not to resolve because it's a lot of busywork for a branch which is predominantly in maintenance support.

I'll make a `master` PR.

akien-mga:
> This PR should likely also assess the cases where `#if [^d\!].*ENABLED` and not just `#elif`, which have the same problem. That would fully solve the inconsistency.

I amended the commit with the `#if` inconsistencies resolved too.

> I'll make a `master` PR.

Done: #87286.

akien-mga (approved):
Should be good to go.

Also included the minor style fix from #87271 in this PR as it's also just fixing a pre-processor macro inconsistency.

akien-mga:

PR #87275, opened 2024-01-16
Add support for default arguments in utility functions

LeoBelda:
Facilitate godotengine/godot-proposals#4425

Made use of helpers from `binder_common.h` to support default arguments in utility functions, the same way it is done in `variant_call.cpp` for static functions. 
Updated doc_tools, api_dump, and tests related to utility functions.

Tested locally by adding `tolerance` as a default argument of `is_zero_approx()`, (see https://github.com/godotengine/godot/commit/c8d6367a9cd8a1d96266fc20777638d1166ba80e), and validated that:
- using `is_zero_approx()` in GDscript with/without the default argument was giving the expected result
- the documentation was correctly generated  
image
- the hints were correctly displaying default arguments 
image


Chaosus:
@LeoBelda This needs to be rebased. Just a reminder.

LeoBelda:
@Chaosus Sorry I couldn't be present for a while, thanks for updating and fixing that code in the other PR, I'm closing this one

PR #87277, opened 2024-01-16
[macOS] Handle Trackpad / Magic Mouse swipe events and simulate XBUTTON1 / XBUTTON2 buttons.

### LINKED ISSUE #86393: Godot doesn't capture the physical 'back' and 'forward' buttons of my mouse
### Tested versions

4.2.1

### System information

Godot v4.2.1.stable - macOS 14.2.0, using a Logitech MX Master 3S mouse with the Steermouse system settings plugin instead of Logitech's own garbage configuration software

### Issue description

The editor doesn't capture the physical 'back' and 'forward' buttons of my mouse, meaning I can't conveniently go back and forth in the script editor. And before we go blaming the Steermouse system settings plugin: the back and forward buttons work in all other applications where it makes sense (eg. browser, mail, text editors, other IDE's, etc) so Godot is specifically not tracking these buttons.

I have `interface/editor/mouse_extra_buttons_navigate_history` enabled.

Perhaps this has something to do with the fact that I'm on macOS and the input enums for the back and forward button have a comment saying those buttons are windows only, meaning that it never worked on macOS in the first place: https://github.com/godotengine/godot/blob/9d1cbab1c432b6f1d66ec939445bec68b6af519e/core/input/input_enums.h#L120

### Steps to reproduce

Get yourself a mouse with back and forward buttons! Then put this in a script in an empty project in Godot:
```
func _input(event):
	print(event)
```
and then run from the editor on Mac. Press the back and forward buttons on the mouse: notice that nothing related to the back and forward buttons is in the output.

N/A

bruvzg:
Back/forward physical buttons on the mouse (at least on Logitech MX Master) also trigger the same swipe events (not mouse events).

Fixes https://github.com/godotengine/godot/issues/86393

### DIFF
--- a/platform/macos/godot_content_view.h
+++ b/platform/macos/godot_content_view.h
@@ -72,7 +72,7 @@
 
 - (void)processScrollEvent:(NSEvent *)event button:(MouseButton)button factor:(double)factor;
 - (void)processPanEvent:(NSEvent *)event dx:(double)dx dy:(double)dy;
-- (void)processMouseEvent:(NSEvent *)event index:(MouseButton)index pressed:(bool)pressed;
+- (void)processMouseEvent:(NSEvent *)event index:(MouseButton)index pressed:(bool)pressed outofstream:(bool)outofstream;
 - (void)setWindowID:(DisplayServer::WindowID)wid;
 - (void)updateLayerDelegate;
 - (void)cancelComposition;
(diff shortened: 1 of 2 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #85583, opened 2023-12-01
Fix `InputEventMouseMotion` is sent to the wrong `Popup` on Windows

Maran23:
This fixes that a `Popup` may do not get mouse motion events, although it should.
 The consequences were that e.g. a tooltip for a popup item blocked highlighting. (https://github.com/godotengine/godot/issues/70361)
Another issue is, that opening a sub menu blocks the main menu. (https://github.com/godotengine/godot/issues/77246)

The logic now always uses the window that was chosen by Windows itself, which is always the right thing.
The popup behaviour is now the same as on linux and feels much more crisp (tested on a SteamDeck).

I also tested and made sure that bugs like https://github.com/godotengine/godot/issues/69368 do not reappear, since the code written was partly to fix this issue.

Note for testers: `display/window/subwindows/embed_subwindows` needs to be disabled to test this in games.

Fixes: https://github.com/godotengine/godot/issues/70361
Fixes: - https://github.com/godotengine/godot/issues/77246 (only on Windows, not on MacOS)
Fixes: https://github.com/godotengine/godot/issues/66273
Fixes: https://github.com/godotengine/godot/issues/73926

PR #87280, opened 2024-01-16
Display path in tooltips of sub-resource list

KoBeWi:
List of scene's sub-resources lists them using either names or "unnamed". Sometimes it might be desirable to know the path of resource (mostly when you have an error in built-in script and it can't be opened directly), so this PR adds it in the tooltip:
image
Though seems like PopupMenu and tooltips don't work that well together due to some hover issues 🤔

### DIFF
--- a/editor/scene_tree_dock.cpp
+++ b/editor/scene_tree_dock.cpp
@@ -3998,6 +3998,7 @@ void SceneTreeDock::_list_all_subresources(PopupMenu *p_menu) {
 			}
 
 			p_menu->add_item(display_text);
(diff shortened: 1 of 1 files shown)


Maran23:
I created a PR to deal with the popup mouse hover tooltip problems you mentioned: https://github.com/godotengine/godot/pull/85583
Feel free to test if this fixes the problem in this case. It did for the menu.

YuriSizov (approved):

KoBeWi:
@Maran23 Yeah, your PR makes it much more usable.

YuriSizov:
Thanks!

PR #2796, opened 2015-11-16
Improved gamepad support on windows and linux

27thLiz:
Uses SDL2 gamepad mappings, supports hot-swapping on windows and fixes some issues like MS$ keyboards being recognized as joysticks on linux. 
It should now also be easy to add support for force-feedback/rumble .

If your gamepad is not in the built-in list, you can add it's mapping via the env variable "SDL_GAMECONTROLLERCONFIG"
Based of punto- 's work in #1464 , thanks for that ;)

Sorry for unnecessary commits, I'm new to git and didn't know about rebase :/

PR #87286, opened 2024-01-17
Fix `#if *_ENABLED` inconsistencies, should check if defined

akien-mga:
Redo and extension of #87272 for the `master` branch.
See https://github.com/godotengine/godot/pull/87272#issuecomment-1895372574 for rationale.

This shouldn't change behavior, but makes things consistent with the rest of the codebase and less error prone.

### DIFF
--- a/core/input/input_builders.py
+++ b/core/input/input_builders.py
@@ -45,10 +45,10 @@ def make_default_controller_mappings(target, source, env):
                 platform_mappings[current_platform][guid] = line
 
     platform_variables = {
-        "Linux": "#if LINUXBSD_ENABLED",
+        "Linux": "#ifdef LINUXBSD_ENABLED",
         "Windows": "#ifdef WINDOWS_ENABLED",
         "Mac OS X": "#ifdef MACOS_ENABLED",
-        "Android": "#if defined(__ANDROID__)",
+        "Android": "#ifdef ANDROID_ENABLED",
         "iOS": "#ifdef IOS_ENABLED",
         "Web": "#ifdef WEB_ENABLED",
     }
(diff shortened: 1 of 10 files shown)


YuriSizov (approved):

core/input/input_builders.py
```diff
@@ -45,10 +45,10 @@ def make_default_controller_mappings(target, source, env):
                 platform_mappings[current_platform][guid] = line
 
     platform_variables = {
-        "Linux": "#if LINUXBSD_ENABLED",
+        "Linux": "#ifdef LINUXBSD_ENABLED",
         "Windows": "#ifdef WINDOWS_ENABLED",
         "Mac OS X": "#ifdef MACOS_ENABLED",
-        "Android": "#if defined(__ANDROID__)",
+        "Android": "#ifdef ANDROID_ENABLED",
```
akien-mga, line 51: I added this, which is the only functional change in this PR.
I couldn't see a reason why this one used a compiler-provided macro instead of our own like other platforms.

I guess at the time it was authored we might not have had an `ANDROID_ENABLED` define yet? Or it was just an oversight.

akien-mga: Did a quick blame trip, seems like this was added in 40b0c5583493ca34e5a1f3623cb122b98d74cc52 as is, but that was copied from the defines used at the time in `main/input_default.cpp`. And those date back to af633c794179d29a4bf0c01983c8b08d14553596, which already had the `ANDROID_ENABLED` define, so I think this was just an oversight.

YuriSizov: > which already had the `ANDROID_ENABLED` define

Well, we had it since GODOT IS OPEN SOURCE, so I think your assessment is correct. It's strange, because it's inconsistent, but you'd have to ask Ariel about it, because it originates from https://github.com/godotengine/godot/pull/2796/commits/0d807264a1d6cf6c6055171021f7e984bdce5b1d.

YuriSizov:
Thanks!

PR #87293, opened 2024-01-17
Improve editor theme generation after the refactor

YuriSizov:
Follow-up to https://github.com/godotengine/godot/pull/87085.

- Fixes a bug with the color mode of icons and fonts not respecting the color preset.
- Fixes a bug with the benchmark by moving benchmarking code to the main class and tracking the number of runs.
- As a cleanup step, moves `is_dark_theme()` from `EditorSettings` to `EditorThemeManager`.
- Adds a hashing mechanism to easily compare editor settings affecting the theme generation.
- Updates the icon generation to use this mechanism, and adds notes for future improvements to fonts and styles.
- Properly tracks the `subresource_hue_tint` setting and accounts for it when deciding to regenerate the theme.
- Removes `EDITOR_GET`s from `editor_icons.cpp` to better consolidate them in the main file.
  - This also needs to be done in some way to `editor_fonts.cpp`, but that's a more complex task for now. 
- Some less notable cleanup.

### DIFF
--- a/editor/debugger/editor_performance_profiler.cpp
+++ b/editor/debugger/editor_performance_profiler.cpp
@@ -31,9 +31,9 @@
 #include "editor_performance_profiler.h"
 
 #include "editor/editor_property_name_processor.h"
-#include "editor/editor_settings.h"
 #include "editor/editor_string_names.h"
 #include "editor/themes/editor_scale.h"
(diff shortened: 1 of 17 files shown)


akien-mga (approved):
Looks good to me from a cursory review.

PR #87294, opened 2024-01-17
Allow `free()` to be used as Callable

### LINKED ISSUE #86706: Can't use `free()` as Callable
### Tested versions

Confirmed in 4.2, but likely applies to all current 4.x versions

### System information

### Issue description

When trying to use `free()` as a Callable (e.g. for connecting signal or using `call_deferred()`), it fails with `Parser Error: Identifier not found: free`, but only at runtime.

### Steps to reproduce

1.
```GDScript
func _ready():
	free.call_deferred()
```
2. Run.

N/A

vnen:
This method is registered in a special way so ClassDB doesn't naturally know about its existence. Here it is hardcoded if any other option fail to check if it is about the "free()" method and, if so, say it exists and return a Callable.

Fix #86706

### DIFF
--- a/core/object/class_db.cpp
+++ b/core/object/class_db.cpp
@@ -31,6 +31,7 @@
 #include "class_db.h"
 
 #include "core/config/engine.h"
+#include "core/core_string_names.h"
 #include "core/io/resource_loader.h"
 #include "core/object/script_language.h"
 #include "core/os/mutex.h"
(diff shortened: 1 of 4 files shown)


akien-mga:
It seems to make the GDScript test suite crash:
```
https://github.com/godotengine/godot/actions/runs/7557248007/job/20576060316?pr=87294
```

vnen:
Yeah, the problem is some things still tries to get the MethodBind, which does not exist for `free()`. I'll change back the `has_method` to avoid this issue and add a more localized solution in GDScript.

dalexeev (approved):
Looks good to me. Probably worth adding a test.

```gdscript
var node := Node.new()
var callable: Callable = node.free
callable.call()
print(node)

node = Node.new()
callable = node["free"]
callable.call()
print(node)
```

vnen:
Added a test.

akien-mga:
Thanks!

PR #87295, opened 2024-01-17
Fix one-click deploy label

adamscott:
Making consistent the label [with the documentation](https://docs.godotengine.org/en/latest/tutorials/export/one-click_deploy.html) (for once!)

### Before

image

### After

image

### DIFF
--- a/editor/editor_run_native.cpp
+++ b/editor/editor_run_native.cpp
@@ -67,10 +67,10 @@ void EditorRunNative::_notification(int p_what) {
 				}
 				if (popup->get_item_count() == 0) {
 					remote_debug->set_disabled(true);
-					remote_debug->set_tooltip_text(TTR("No Remote Debug export presets configured."));
+					remote_debug->set_tooltip_text(TTR("No one-click deploy export presets configured."));
 				} else {
 					remote_debug->set_disabled(false);
-					remote_debug->set_tooltip_text(TTR("Remote Debug"));
+					remote_debug->set_tooltip_text(TTR("One-click Deploy"));
 				}
 
 				first = false;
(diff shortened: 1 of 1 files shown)


bruvzg (commented):
I would do the opposite and rename the docs page. "Remote debug" better represent what's this feature doing.

Also, there are properties and built-in docs parts that are referring to it in both ways, so these should be updated as well.

Mickeon:
Renaming the doc page sounds a bit daunting. It's not just the name of the page itself, rather the whole explanation of the feature that would need to be readjusted. It has also been the name of the feature for quite a while.

mhilbrunner:
I'm fine either way. "Remote debug" would seem to need less explanation than "One-click deploy" to me, coming at it without any knowledge. We can rename the docs page and do redirects. Otherwise, this is fine. Not sure how to best decide this?

arkology:
It's kinda strange to see that PR has `cherrypick` labels and at the same time targets 4.x. Something went wrong?

Mickeon:
It was potentially thought that this PR would've been merged sooner, but as you can see it got moved to the `4.x` milestone.

akien-mga:
> I would do the opposite and rename the docs page. "Remote debug" better represent what's this feature doing.
> 
> Also, there are properties and built-in docs parts that are referring to it in both ways, so these should be updated as well.

I agree with this, I think "Remote Debug" is the better name.

"One-click Deploy" is a marketing name and it's useful for that purpose - make people understand that they can get their game running on a device in a single click. That's good for discoverability. But for actual users, it's then more important to understand that this isn't just a one-click export thing, it actually hooks up a debugger to their exported project and lets them debug (even live edit!) remotely.

Calinou:
> "One-click Deploy" is a marketing name and it's useful for that purpose -

It's also two clicks if you have more than one target device available, as it switches to a MenuButton in that case :slightly_smiling_face:

Mickeon:
Should we as well... just bite the bullet with a full rename? I am not against it.

Repiteo:
Superseded by #122858

PR #81605, opened 2023-09-13
GDScript: Fix some lambda bugs

dalexeev:
* Closes #69491.
* Closes #79707.
* Closes #81572.

As stated in #81572, `GDScriptFunction::_get_call_error()` can crash the engine.

https://github.com/godotengine/godot/blob/3ed4497113fa10611b90290ce22a751fb9d26e2e/modules/gdscript/gdscript_vm.cpp#L118-L121

https://github.com/godotengine/godot/blob/3ed4497113fa10611b90290ce22a751fb9d26e2e/modules/gdscript/gdscript_vm.cpp#L1773-L1773

I decided to check what `err.argument` was equal to, and it turned out to be `-1`. This is because lambda captures are hidden arguments of lambda function.

https://github.com/godotengine/godot/blob/3ed4497113fa10611b90290ce22a751fb9d26e2e/modules/gdscript/gdscript_lambda_callable.cpp#L77-L88

If for some reason a capture is invalid (for example, due to a bug in the compiler or if the object has been freed at this point), then this leads to a negative argument index and a crash.

https://github.com/godotengine/godot/blob/3ed4497113fa10611b90290ce22a751fb9d26e2e/modules/gdscript/gdscript_vm.cpp#L545-L548

https://github.com/godotengine/godot/blob/3ed4497113fa10611b90290ce22a751fb9d26e2e/modules/gdscript/gdscript_function.h#L109-L113

* (See also #80844.)

**Status**

* #62769
  * :no_entry_sign: This bug is difficult to reproduce. It seems to be related to hot reloading.
* #74607 
  * :no_entry_sign: Some crashes were prevented, but the hot reloading bug remained (which can also cause a crash in itself).
* #69491
  * :white_check_mark: Fixed.
* A bug with freed capture.
  * #79707
  * :white_check_mark: Fixed.
* #81572
  * :white_check_mark: Fixed ([comment](https://github.com/godotengine/godot/pull/81605#issuecomment-1717325026)). 

```gdscript
@tool
extends EditorScript

func _run() -> void:
    var x: Node = Node.new()
    x.free()
    (func (): print(x)).call()
```

PR #87296, opened 2024-01-17
GDScript: Assume freed object to be of any Object type

### LINKED ISSUE #86609: Providing static typing to an argument throws "Cannot convert argument 2 from Object to Object"
### Tested versions

- Reproducible in v4.2.1.stable.arch_linux

### System information

Godot v4.2.1.stable unknown - EndeavourOS #1 SMP PREEMPT_DYNAMIC Thu, 14 Dec 2023 03:45:42 +0000 - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1060 3GB (nvidia; 545.29.06) - Intel(R) Core(TM) i5-4670K CPU @ 3.40GHz (4 Threads)

### Issue description

I am attaching a function to a timer:
```gdscript
node.timer.connect("timeout", tower_shoot.bind(node, target))

# ...

func tower_shoot(node: Tower, target: CharacterBody3D) -> void:
	if !is_instance_valid(target):
		return
	print(target.get_class()) # CharacterBody3D
```
When "target" is freed, an error appears:
```
E 0:00:10:0716   emit_signalp: Error calling from signal 'timeout' to callable: 'Node3D(game.gd)::tower_shoot': Cannot convert argument 2 from Object to Object.
  <C++ Source>   core/object/object.cpp:1140 @ emit_signalp()
```
Strangely, the error goes away when *not* providing a type to `target`, and the first `if` statement handles the rest.
```gdscript
func tower_shoot(node: Tower, target) -> void: # <- removed type for "target"
	if !is_instance_valid(target):
		return # no more errors because this runs
	print(target.get_class())
```

I'm forcing static typing within GoDot, so I can't let the "target" be untyped. Is this a bug, or am I missing something?

P.S. simpler example:
```gdscript
extends Node2D

func _ready() -> void:
	$t1/Timer.connect("timeout", foo.bind($t1, $t2))

	await get_tree().create_timer(5).timeout
	$t2.queue_free() # will cause the error, but if we would've let t2 without a type, this wouldn't happen

func foo(t1: Node2D, t2: Node2D) -> void:
	print(t1, t2)
```

### Steps to reproduce

```gdscript
@tool
extends EditorScript

func _run() -> void:
    var obj: Object = Object.new()
    obj.free()
    var obj2: Object = obj # No error.
    foo(obj) # Error.

func foo(obj: Object) -> void:
    pass```
Removing `: Object` from `foo` solves the issue.

[godot_bug.zip](https://github.com/godotengine/godot/files/13794554/godot_bug.zip)

### LINKED ISSUE #101423: Static typing for for loop variable errors on invalid references
### Tested versions

v4.3.stable.official [77dcf97d8]

### System information

Godot v4.3.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2060 (NVIDIA; 32.0.15.6636) - AMD Ryzen 5 7600X 6-Core Processor (12 Threads)

### Issue description

v4.3.stable.official [77dcf97d8]

The recently added static typing for for loop variables here https://github.com/godotengine/godot/pull/80247 has a weird quirk that I feel like I may be considered a bug, or at least strange behavior. 

If you are looping over an array that contains node references, and one of those nodes gets freed, `<Freed Object>` remains in its index. While normally you can use `is_instance_valid()` for validating your node references, trying to do this in a typed for loop will fail with the error "Trying to assign invalid previously freed instance" at the "for" line, before it even gets to where you can validate it. However, it's possible to get around this by removing the type from the loop variable and checking it after the instance is validated, but that then makes typing the loop variable questionable in the first place. Since I think that may be unclear, here's some example code of what I mean:

```gdscript
# Tested with an Array with some generic Node2Ds
var objects_to_be_freed: Array = get_tree().get_nodes_in_group("object_freeing")

# After any of those nodes is freed...

# Runs without error
for object in objects_to_be_freed:
	if is_instance_valid(object):
		print("Valid!")
	else:
		print("Invalid!")

# Will cause error: "Trying to assign invalid previously freed instance"
for object: Node2D in objects_to_be_freed: # Errors at this line
	if is_instance_valid(object):
		print("Valid!")
	else:
		print("Invalid!")

# Runs without error and effectively keeps static typing as far as I can tell
for object in objects_to_be_freed:
	if is_instance_valid(object) and object is Node2D:
		print("Valid!")
	else:
		print("Invalid!")
```
This makes me wonder if this behavior is intended, and if it is intended, why? The 3rd block (the workaround) seems like it would be the best way to handle this in all cases, and if that's the case, then why isn't the first block just interpreted the same way by the engine? 

I would expect the first block to work without error as written, simply printing "Invalid!" if the object was freed.

### Steps to reproduce

Make an array, add node references to it, free any of the nodes without clearing the reference, create a for loop to iterate over that array, and make the for loop variable typed to those nodes. Tested this with the code snippet in the issue description.

N/A

vnen:
Since we can't check the actual type, we assume the type is correct if the expected type is Object-derived, the same as what happen with null values.

* Fix #86609

Edit: added an exception for `is` to also allow it on freed objects, but in this case it's always `false`. Also added a few test cases.

* _Bugsquad edit:_ Fixes #101423.

### DIFF
--- a/modules/gdscript/gdscript_function.h
+++ b/modules/gdscript/gdscript_function.h
@@ -106,10 +106,10 @@ public:
 					return false;
 				}
 
-				bool was_freed = false;
-				Object *obj = p_variant.get_validated_object_with_check(was_freed);
+				Object *obj = p_variant.get_validated_object();
 				if (!obj) {
-					return !was_freed;
+					// Assume null or freed to be Object.
+					return true;
 				}
 
 				if (!ClassDB::is_parent_class(obj->get_class_name(), native_type)) {
@@ -126,10 +126,10 @@ public:
 					return false;
 				}
 
-				bool was_freed = false;
-				Object *obj = p_variant.get_validated_object_with_check(was_freed);
+				Object *obj = p_variant.get_validated_object();
 				if (!obj) {
-					return !was_freed;
+					// Assume null or freed to be Object.
+					return true;
 				}
 
 				Ref<Script> base = obj && obj->get_script_instance() ? obj->get_script_instance()->get_script() : nullptr;
--- a/modules/gdscript/gdscript_lambda_callable.cpp
+++ b/modules/gdscript/gdscript_lambda_callable.cpp
@@ -101,15 +101,6 @@ void GDScriptLambdaCallable::call(const Variant **p_arguments, int p_argcount, V
 		args.resize(p_argcount + captures_amount);
 		for (int i = 0; i < captures_amount; i++) {
 			args.write[i] = &captures[i];
-			if (captures[i].get_type() == Variant::OBJECT) {
-				bool was_freed = false;
-				captures[i].get_validated_object_with_check(was_freed);
-				if (was_freed) {
-					ERR_PRINT(vformat(R"(Lambda capture at index %d was freed. Passed "null" instead.)", i));
-					static Variant nil;
-					args.write[i] = &nil;
-				}
-			}
 		}
 		for (int i = 0; i < p_argcount; i++) {
 			args.write[i + captures_amount] = p_arguments[i];
@@ -229,15 +220,6 @@ void GDScriptLambdaSelfCallable::call(const Variant **p_arguments, int p_argcoun
 		args.resize(p_argcount + captures_amount);
 		for (int i = 0; i < captures_amount; i++) {
 			args.write[i] = &captures[i];
-			if (captures[i].get_type() == Variant::OBJECT) {
-				bool was_freed = false;
-				captures[i].get_validated_object_with_check(was_freed);
-				if (was_freed) {
-					ERR_PRINT(vformat(R"(Lambda capture at index %d was freed. Passed "null" instead.)", i));
-					static Variant nil;
-					args.write[i] = &nil;
-				}
-			}
 		}
 		for (int i = 0; i < p_argcount; i++) {
 			args.write[i + captures_amount] = p_arguments[i];
--- a/modules/gdscript/gdscript_vm.cpp
+++ b/modules/gdscript/gdscript_vm.cpp
@@ -826,13 +826,7 @@ Variant GDScriptFunction::call(GDScriptInstance *p_instance, const Variant **p_a
 				GD_ERR_BREAK(native_type_idx < 0 || native_type_idx >= _global_names_count);
 				const StringName native_type = _global_names_ptr[native_type_idx];
 
-				bool was_freed = false;
-				Object *object = value->get_validated_object_with_check(was_freed);
-				if (was_freed) {
-					err_text = "Left operand of 'is' is a previously freed instance.";
-					OPCODE_BREAK;
-				}
-
+				Object *object = value->get_validated_object();
 				*dst = object && ClassDB::is_parent_class(object->get_class_name(), native_type);
 				ip += 4;
 			}
@@ -848,12 +842,7 @@ Variant GDScriptFunction::call(GDScriptInstance *p_instance, const Variant **p_a
 				Script *script_type = Object::cast_to<Script>(type->operator Object *());
 				GD_ERR_BREAK(!script_type);
 
-				bool was_freed = false;
-				Object *object = value->get_validated_object_with_check(was_freed);
-				if (was_freed) {
-					err_text = "Left operand of 'is' is a previously freed instance.";
-					OPCODE_BREAK;
-				}
+				Object *object = value->get_validated_object();
 
 				bool result = false;
 				if (object && object->get_script_instance()) {
--- a/modules/gdscript/tests/scripts/runtime/features/freed_treated_as_object_for_argument.gd
+++ b/modules/gdscript/tests/scripts/runtime/features/freed_treated_as_object_for_argument.gd
@@ -1,0 +1,7 @@
+func typed(node: Node) -> void:
+	print(node)
+
+func test():
+	var node := Node.new()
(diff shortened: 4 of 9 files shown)


dalexeev:
Makes a lot of sense to me, but we should check other places as well. For example, the following hack is probably no longer needed, at least replacing `Freed Object` with `null` (see #81605 for details):

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_lambda_callable.cpp#L95-L101

The old behavior (error when passing `Freed Object` as a parameter) was convenient, perhaps people rely on it:



It seems that the useful error will not appear with this PR, since it does not appear now if you remove the type hint from the function parameter. Probably a similar check should be done separately, since we check this in many cases:

<details>
<summary>Incomplete list</summary>

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_utility_functions.cpp#L550

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_utility_functions.cpp#L567

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_vm.cpp#L825

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_vm.cpp#L847

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_vm.cpp#L1362

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_vm.cpp#L1399

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_vm.cpp#L2249

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_vm.cpp#L2560

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_vm.cpp#L2602

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_vm.cpp#L2970

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_vm.cpp#L3301

</details>

By the way, I noticed an error, `was_freed` is not used:

https://github.com/godotengine/godot/blob/107f2961ccfac179af7682eb5f6e7ea91e80040c/modules/gdscript/gdscript_vm.cpp#L1721-L1730

vnen:
What I think is that is if a value is immediately used it should give an error. For instance, passing it to a utility function should raise the error because the user does not have control of its contents, but passing to a user-made function (like in this PR) should not trigger it automatically. Because sometimes the user don't have a lot of control of the source of data. For most cases being freed or `null` is pretty much the same, so while the error is somewhat useful in some situations, it does not really exclude argument validation.

I do agree that it should be changed in a few other places to be consistent with this if needed.

vnen:
Updated this to remove the check in lambda captures, since that is not needed anymore. I also added a change to allow `is` to be performed on a freed object, which will always return `false` since the type cannot be checked.

I considered doing the same for casting (`as`) but I think it's better to keep it an error. If you're casting a freed object you're most likely doing something wrong.

Also added a few test cases to validate the behavior.

Macksaur:
I disagree with `as` staying an error. `is`/`as` are supposed to be non-throwing and safe operators to write type-safe code, not randomly crash to the editor operators. I was very surprised they worked like this.

> If you're casting a freed object you're most likely doing something wrong.

Can you describe which scenarios this helps?

Assuming you deliberately have a freed object you cannot store/access/move it safely elsewhere or type-check with the `is`/`as` operators since it will break, despite the logic being functionally correct otherwise. With the current design you are burdened to micromanage every resource intricately. It basically makes freed objects radioactive, when they could be graceful.

I think that *all* of the VM breaks should be removed and replaced with gentle linting (that can be toggled off) rather than hard errors that prevent the user from progressing with simpler code.

Shadows-of-Fire:
The behavior of `<freed> as Type` should match the behavior of `null as Type`. Granted, I don't know what the behavior of `null as Type` is off the top of my head.

From a language perspective, I would expect a previously freed object to behave the same as `null` in all contexts except error reporting, when we can state that we have a known previously-freed object rather than an explicit `null`.

romgerman:
> I disagree with `as` staying an error. `is`/`as` are supposed to be non-throwing and safe operators to write type-safe code, not randomly crash to the editor operators. I was very surprised they worked like this.

I agree with that. In my project (4.4) I must do an object validation check before `is` operator otherwise I get an error.
```gdscript
if is_instance_valid(dmg.source) and dmg.source is Player:
	_play_hit_sfx()
```
This can happen if the `dmg.source` value is a node (as it should be in my project) and the node is freed. I expect the `is` operator to work as a guard to check that because logically if an object is freed then its `null` and `null` is not of any class type. (Correct me if I'm wrong. I assume you cannot use freed objects to access their properties and methods. But maybe it's a wrong assumption)

PR #87301, opened 2024-01-17
Replace built-in docs codeblock's leading spaces with tabs

Mickeon:
This PR replaces the spaces used in the codeblocks of the class references with actual tabs.
If your Editor Setting's preference is using spaces, nothing changes.
With one GIF, this means the following (notice the tabs):

| Before | After |
| --- | --- |
|  | 

However, hm... It comes at a cost. For the tab to display in a way that actually looks good, the **RichTextLabel**'s `tab_size` has to be set to `8`, otherwise the tab's indent would be way, way too small, for some reason. Because of this, there's the side effect of increasing the indentation across the whole built-in documentation:

Take a closer look:

| Before | After |
| --- | --- |
| image | image |
| image |image | 
| image | image |
| image | image

Some may say this is a happy accident, some may absolutely not. So keep that in mind.

### DIFF
--- a/editor/editor_help.cpp
+++ b/editor/editor_help.cpp
@@ -35,6 +35,7 @@
 #include "core/input/input.h"
 #include "core/object/script_language.h"
 #include "core/os/keyboard.h"
+#include "core/string/string_builder.h"
 #include "core/version.h"
 #include "editor/doc_data_compressed.gen.h"
 #include "editor/editor_node.h"
@@ -2384,6 +2385,7 @@ static void _add_text_to_rt(const String &p_bbcode, RichTextLabel *p_rt, Control
 	List<String> tag_stack;
 	bool code_tag = false;
 	bool codeblock_tag = false;
+	const bool using_tab_indent = int(EDITOR_GET("text_editor/behavior/indent/type")) == 0;
 
 	int pos = 0;
 	while (pos < bbcode.length()) {
@@ -2395,6 +2397,25 @@ static void _add_text_to_rt(const String &p_bbcode, RichTextLabel *p_rt, Control
 
 		if (brk_pos > pos) {
 			String text = bbcode.substr(pos, brk_pos - pos);
+			if (codeblock_tag && using_tab_indent) {
+				// Replace the code block's space indentation with tabs.
+				StringBuilder builder;
+				PackedStringArray text_lines = text.split("\n");
+				for (const String &line : text_lines) {
+					String stripped_line = line.dedent();
+					int space_count = line.length() - stripped_line.length();
+
+					if (builder.num_strings_appended() > 0) {
+						builder.append("\n");
+					}
+					if (space_count > 0) {
+						builder.append(String("\t").repeat(MAX(space_count / 4, 1)) + stripped_line);
+					} else {
+						builder.append(line);
+					}
+				}
+				text = builder.as_string();
+			}
 			if (!code_tag && !codeblock_tag) {
 				text = text.replace("\n", "\n\n");
 			}
@@ -2934,6 +2955,7 @@ EditorHelp::EditorHelp() {
 	EDITOR_DEF("text_editor/help/sort_functions_alphabetically", true);
 
 	class_desc = memnew(RichTextLabel);
+	class_desc->set_tab_size(8);
 	add_child(class_desc);
 	class_desc->set_threaded(true);
 	class_desc->set_v_size_flags(SIZE_EXPAND_FILL);


YuriSizov:

Mickeon:
Done just that.

MewPurPur:
I don't like the increased indent but I think it's worth it. I'm just confused why we'd need an 8 tab size for what's supposed to be monospace. I would've expected the mono tabs to be as wide as my code editor theme (which is at 3 for me).

Mickeon:
> I'm just confused why we'd need an 8 tab size for what's supposed to be monospace.

Honestly, I am not sure why **RichTextLabel** acts like this. I tried a value of `4` which you would think would be appropriate, but it resulted in the padding being **way too** short for some reason.

It originally was also based on the "indent size" Editor Setting, but because it would've affected the entire **RichTextLabel** I was not sure about it.

editor/editor_help.cpp
```diff
@@ -2061,6 +2062,24 @@ static void _add_text_to_rt(const String &p_bbcode, RichTextLabel *p_rt, Control
 
 		if (brk_pos > pos) {
 			String text = bbcode.substr(pos, brk_pos - pos);
+			if (codeblock_tag && code_indent_type == 0) {
+				// Replace the code block's space indentation with tabulation.
+				PackedStringArray text_lines = text.split("\n");
+				text = String();
+				for (String line : text_lines) {
```
AThousandShips, line 2069:
```suggestion
				for (const String &line : text_lines) {
```

editor/editor_help.cpp
```diff
@@ -2061,6 +2062,24 @@ static void _add_text_to_rt(const String &p_bbcode, RichTextLabel *p_rt, Control
 
 		if (brk_pos > pos) {
 			String text = bbcode.substr(pos, brk_pos - pos);
+			if (codeblock_tag && code_indent_type == 0) {
+				// Replace the code block's space indentation with tabulation.
```
AThousandShips, line 2066:
```suggestion
				// Replace the code block's space indentation with tabs.
```
"tabulation" means creating a list

AThousandShips (approved):
LGTM

Calinou (dismissed):
Tested locally, there are issues with code samples that contain multiple tabs on the same line.

For instance, the HTTPRequest class description has a code sample where the `push_error()` line has two leading tabs, but none are displayed and copied here:

image

If you select the block and paste it in a script, you get this:

image

Mickeon:
You will never guess who the idiot that swapped divisor and dividend around is.

Mickeon:
You will never guess who the idiot that completely f<sub>u</sub><sup>c</sup><sub>k</sub><sup>e</sup>d up the formula is.

Should work fine now:

editor/editor_help.cpp
```diff
@@ -2384,6 +2384,7 @@ static void _add_text_to_rt(const String &p_bbcode, RichTextLabel *p_rt, Control
 	List<String> tag_stack;
 	bool code_tag = false;
 	bool codeblock_tag = false;
+	const int code_indent_type = EDITOR_GET("text_editor/behavior/indent/type");
```
KoBeWi, line 2387: Changing this setting won't update currently opened help pages.
Not sure if it's worth handling though.

editor/editor_help.cpp
```diff
@@ -2395,6 +2396,24 @@ static void _add_text_to_rt(const String &p_bbcode, RichTextLabel *p_rt, Control
 
 		if (brk_pos > pos) {
 			String text = bbcode.substr(pos, brk_pos - pos);
+			if (codeblock_tag && code_indent_type == 0) {
```
KoBeWi, line 2399: Could use a comment on where the 0 comes from.

editor/editor_help.cpp
```diff
@@ -2395,6 +2396,24 @@ static void _add_text_to_rt(const String &p_bbcode, RichTextLabel *p_rt, Control
...
+					String stripped_line = line.dedent();
+					int space_count = line.length() - stripped_line.length();
+
+					if (!text.is_empty()) {
+						text += "\n";
+					}
+					if (space_count > 0) {
+						text += String("\t").repeat(MAX(space_count / 4, 1)) + stripped_line;
+					} else {
+						text += line;
```
KoBeWi, line 2413: Looks like a job for StringBuilder.

editor/editor_help.cpp
Mickeon: I'd rather modify this part here to use a bool.

In a prior version of this PR I also handled spaces and adjusted the padding depending on `text_editor/behavior/indent/size`, but in practice it looked very bad.

KoBeWi (approved):

akien-mga:
Thanks!

Mickeon:
And from here on out begins the built-in reference revolution

PR #87313, opened 2024-01-17
Refinement of DirAccess logic

### LINKED ISSUE #87274: After loading a pck file (including zip), you can only access the directories and files in the pck
### Tested versions

v4.2.1.stable.arch_linux

### System information

Godot v4.2.1.stable unknown - Arch Linux #1 ZEN SMP PREEMPT_DYNAMIC Mon, 15 Jan 2024 01:11:17 +0000 - Wayland - GLES3 (Compatibility) - Mesa Intel(R) UHD Graphics 630 (CFL GT2) () - Intel(R) Core(TM) i5-9300H CPU @ 2.40GHz (8 Threads)

### Issue description

As the title suggests
图片
I'm using the pck system for mod functions, and it's getting in the way of some of my plans because I still need to access directories and files in my own "res" after loading the player-supplied pck.
By looking at the source code, this seems to be intentional? Is it possible to add a parameter to "load_resource_pack" to determine whether file system access starts from the loaded pck?
图片

### Steps to reproduce

Running Projects

[test_pck.zip](https://github.com/godotengine/godot/files/13955871/test_pck.zip)

scgm0:
Fix #87274 
Referred to the FileAccess related code, I roughly tested linux and android, may need more tests

### DIFF
--- a/core/io/dir_access.cpp
+++ b/core/io/dir_access.cpp
@@ -31,7 +31,7 @@
 #include "dir_access.h"
 
 #include "core/config/project_settings.h"
-#include "core/io/file_access.h"
+#include "core/io/file_access_pack.h"
 #include "core/os/memory.h"
 #include "core/os/os.h"
 #include "core/templates/local_vector.h"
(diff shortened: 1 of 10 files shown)


core/io/dir_access.cpp
```diff
@@ -222,7 +222,7 @@ String DirAccess::fix_path(String p_path) const {
 	return p_path;
 }
 
-DirAccess::CreateFunc DirAccess::create_func[ACCESS_MAX] = { nullptr, nullptr, nullptr };
+DirAccess::CreateFunc DirAccess::create_func[ACCESS_MAX] = { nullptr, nullptr };
```
AThousandShips, line 225: I'm not sure about this change, why two entries for an array with three entries? I'd say it's wrong in `FileAccess`, not here

scgm0:
I ignored the mixed directory traversal case, which was more complicated than I thought, so I turned this PR into a draft first.

core/io/dir_access.cpp
scgm0: > I'm not sure about this change, why two entries for an array with three entries? I'd say it's wrong in `FileAccess`, not here

So I should change FileAccess instead of here?

AThousandShips: I'd say yes, not sure this causes any problems, but it's probably not good either to have this 2/3 thing

scgm0: > I'd say yes, not sure this causes any problems, but it's probably not good either to have this 2/3 thing

Done

scgm0:
Closing this pr for now because I can't think of an elegant solution, which requires more discussion.

PR #87316, opened 2024-01-18
Keep 2D items global transform the same when set to top level

### LINKED ISSUE #76205: Toggling top-level does not update 2D editor viewport
### Godot version

4.0.2

### System information

Windows 10 x64

### Issue description

https://user-images.githubusercontent.com/2223172/232775581-ad40bacc-b051-4d71-87fb-ffad3f60be2f.mp4

### Steps to reproduce

1. Add Sprite2D
2. Toggle its `top_level`
3. Notice how the gizmo rectangle stays in the old position until you hover 2D editor

N/A

ryevdokimov:
Fixes: #76205

This solution might be trying to kill the fly with a sledgehammer, but it seemed the most intuitive and keeps in line with what 3D does.

Otherwise, depending on the transform of the parent, the child node being set to top level could result in it appearing and being positioned in an unexpectedly different way. I can't think of any scenarios where this is the desired result. 

Before: 

https://github.com/godotengine/godot/assets/105675984/53cca64f-5da9-40a2-8138-6e81457bfbf9

After:

https://github.com/godotengine/godot/assets/105675984/c4685034-bcba-48db-a39c-00ba701488c4


YuriSizov:
Do we still need this, or was the other PR sufficient?

ryevdokimov:
It’s up to the Godot team, the other PR does solve the problem while keeping the functionality the same, but I personally think the functionality is strange. 

In 3D the world coordinates and transforms don’t change when you set top level. In 2D it does which seems undesirable. This PR makes the behavior between 3D and 2D the same.

lawnjelly:
The current 2D behaviour is probably the same in 3.x. It makes sense, by setting to toplevel, you are removing the parent transform, so you would _expect_ it to move.

If anything, the behaviour in 3D (if as you say it stays in position) would be more "unexpected", but if there is a difference it likely occurs because in 3D the global transform is set and used directly in the server (and presumably is dirty in the server), whereas for 2D the global transform is recalculated through the hierarchy in the server on each render.

ryevdokimov:
Just to play devil's advocate, should I expect it to move if I reparent a child node of a child of root, to root? If I right, click the child node and select 'Reparent'. It gives me an option change the global transform, which is the default (at least on my editor where I've never touched this setting). Additionally, if you drag to reparent, you aren't even prompted, maintaining the global transform is forced. 

This can't both be the expected behavior, right? 

https://github.com/godotengine/godot/assets/105675984/1d9ea812-85aa-4986-8f4d-1a1b84f5f4d9

AThousandShips:
The related issue is closed, what remains here to fix? What issue is this now fixing?

ryevdokimov:
> > Additionally, if you drag to reparent, you aren't even prompted, maintaining the global transform is forced.
> 
> You can hold Shift to bypass that.

Oh cool, I did know that, thanks. 😅

But my point still stands, in both circumstances maintaining the global transform is the "default" behavior. If you all accept that changing a node to top-level doesn't have to be consistent with the default of reparenting that's fine with me, and I can close this.

@AThousandShips, this PR does what the title says as alternative solution to the original issue.

AThousandShips:
(I missed the conversation above, hence deleted my comment)

ryevdokimov:
I suppose what you could also do is just flip the default behavior of reparenting. Let me know if that's something that makes sense to do.

PR #87318, opened 2024-01-18
Add indeterminate mode to ProgressBar

### LINKED ISSUE #8877: Add an "indeterminate" mode to the ProgressBar control

A game with networking or anything that has the user waiting for something with an unknown progress.

Some actions with unknown progress should have a visible indication that something is happening, in my case this is establishing a network connection and initial sync and then loading from a resource, the second part can be easily shown using a ProgressBar, but not the first, considering it's all part of the same "progress" of loading into a game map, it'd be nice if it could be solved with just one control.

It's easy enough to indicate progress through other means but it'd be good for ProgressBar to provide a simple checkbox toggle for "indeterminate" mode that could be switched on an off when the exact progress of the task is unknown but you still want to make clear to the user that something is happening. 

We can add a boolean "indeterminate" property to ProgressBar, when checked this would supersede the current rendering (no percentage shown and no current value shown in the bar since these wouldn't make sense) and we'd get something like this:

https://github.com/godotengine/godot-proposals/assets/138269/894ee6a5-42a2-4241-b81d-07d5ba4d6d40

This is a POC I already have implemented, should be easy to polish it up for a PR.

You can't really achieve the same look just scripting the ProgressBar since it always starts from either the left or the right, you can always make a custom control but if you want it to follow the application theme it's a bit bit more complex.

This would make the ProgressBar node more useful in a broader set of scenarios.

ckaiser:
Implements proposal [#8877](https://github.com/godotengine/godot-proposals/issues/8877), supports all the normal modes of operation for progress bars, RTL stuff and everything I could think of, but I'm still new to the codebase so I might've missed something.

For their speed and size, I tried to make it something sensible for both small bars that might go in editor stuff to big ones that could go across an entire screen in a game, so they have a minimum animation speed and then it scales based on either height or width, whatever is largest to account for top/down stuff. I've been staring at them for a while so any feedback would be great.

Video:

https://github.com/godotengine/godot/assets/138269/79d1cb25-1d73-45b3-a03f-31a4c5a2c776

*Bugsquad edit:*
- Closes https://github.com/godotengine/godot-proposals/issues/8877

### DIFF
--- a/doc/classes/ProgressBar.xml
+++ b/doc/classes/ProgressBar.xml
@@ -9,8 +9,14 @@
 	<tutorials>
 	</tutorials>
 	<members>
+		<member name="editor_preview_indeterminate" type="bool" setter="set_editor_preview_indeterminate" getter="is_editor_preview_indeterminate_enabled">
+			If [code]false[/code], the [member indeterminate] animation will be paused in the editor.
+		</member>
 		<member name="fill_mode" type="int" setter="set_fill_mode" getter="get_fill_mode" default="0">
 			The fill direction. See [enum FillMode] for possible values.
+		</member>
+		<member name="indeterminate" type="bool" setter="set_indeterminate" getter="is_indeterminate" default="false">
+			When set to [code]true[/code], the progress bar indicates that something is happening with an animation, but does not show the fill percentage or value.
 		</member>
 		<member name="show_percentage" type="bool" setter="set_show_percentage" getter="is_percentage_shown" default="true">
 			If [code]true[/code], the fill percentage is displayed on the bar.
--- a/scene/gui/progress_bar.cpp
+++ b/scene/gui/progress_bar.cpp
@@ -50,8 +50,57 @@ Size2 ProgressBar::get_minimum_size() const {
 
 void ProgressBar::_notification(int p_what) {
 	switch (p_what) {
+		case NOTIFICATION_INTERNAL_PROCESS: {
+			if (is_visible_in_tree()) {
+				_inderminate_fill_progress += get_process_delta_time() * MAX(indeterminate_min_speed, MAX(get_size().width, get_size().height) / 2);
+				queue_redraw();
+			}
+		} break;
 		case NOTIFICATION_DRAW: {
 			draw_style_box(theme_cache.background_style, Rect2(Point2(), get_size()));
+
+			if (indeterminate) {
+				Size2 size = get_size();
+				real_t fill_size = MIN(size.width, size.height) * 2;
+
+				if (Engine::get_singleton()->is_editor_hint() && !editor_preview_indeterminate) {
+					// Center the filled bar when we're not previewing the animation.
+					_inderminate_fill_progress = (MAX(size.width, size.height) / 2) + (fill_size / 2);
+				}
+
+				switch (mode) {
+					case FILL_END_TO_BEGIN:
+					case FILL_BEGIN_TO_END: {
+						// Follow the RTL layout with the animation to match how the bar would fill.
+						bool right_to_left = mode == (is_layout_rtl() ? FILL_BEGIN_TO_END : FILL_END_TO_BEGIN);
+
+						if (_inderminate_fill_progress > size.width + fill_size) {
+							_inderminate_fill_progress = right_to_left ? -fill_size : 0;
+						}
+
+						real_t x = right_to_left ? size.width - _inderminate_fill_progress : _inderminate_fill_progress - fill_size;
+						draw_style_box(theme_cache.fill_style, Rect2(x, 0, fill_size, size.height).intersection(Rect2(Point2(), size)));
+					} break;
+					case FILL_TOP_TO_BOTTOM: {
+						if (_inderminate_fill_progress > size.height + fill_size) {
+							_inderminate_fill_progress = 0;
+						}
+
+						draw_style_box(theme_cache.fill_style, Rect2(0, _inderminate_fill_progress - fill_size, size.width, fill_size).intersection(Rect2(Point2(), size)));
+					} break;
+					case FILL_BOTTOM_TO_TOP: {
+						if (_inderminate_fill_progress > size.height + fill_size) {
+							_inderminate_fill_progress = -fill_size;
+						}
+
+						draw_style_box(theme_cache.fill_style, Rect2(0, size.height - _inderminate_fill_progress, size.width, fill_size).intersection(Rect2(Point2(), size)));
+					} break;
+					case FILL_MODE_MAX:
+						break;
+				}
+
+				return;
+			}
 
 			float r = get_as_ratio();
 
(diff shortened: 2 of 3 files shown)


scene/gui/progress_bar.h
```diff
@@ -63,15 +66,20 @@ class ProgressBar : public Range {
 	};
 
 	void set_fill_mode(int p_fill);
-	int get_fill_mode();
+	int get_fill_mode() const;
```
AThousandShips, line 69: You need to add a compatibility method for this if you change it here, I'd leave it as is

KoBeWi:
The ProgressBar will cause redrawing even when 2D editor is currently hidden (e.g. when using Script Editor). Though I don't know how to prevent that, as the scene is technically "visible", because it's part of a viewport. Some nodes, like particles, will respect editor visibility correctly, but it's probably some rendering trick.

ckaiser:
I've done the property updates for show percentage when it's in indeterminate mode, and I've been trying to figure out how to really know which editor is visible, but it didn't seem possible without trying to delve into some hackery.

My proposal is the following: Adding a new `get_visible_editor()` function to `EditorNode` that returns a value from the `EditorTable` enum. Since we can have extra editors thanks to plugins, I've added a new member to the enum to identify this, and tested it to check that it doesn't cause any issues.

How does that sound? IDK how we feel about adding new methods to a core class but I think it can definitely be useful for other controls that may have any kind of animation (I was thinking of proposing a "spinner" control for example)

KoBeWi:
That doesn't look like a proper solution. You shouldn't include editor code in a non-editor class.

I think for now the simplest solution is making the progress bar not process in the editor, unless you enable some property for editor preview.

ckaiser:
Made the adjustments and added the editor preview property, here's how it looks with the preview disabled:
image

doc/classes/ProgressBar.xml
```diff
@@ -9,9 +9,15 @@
 	<tutorials>
 	</tutorials>
 	<members>
+		<member name="editor_preview_indeterminate" type="bool" setter="set_editor_preview_indeterminate" getter="is_editor_preview_indeterminate_enabled">
+			Preview the [member indeterminate] animation in the editor when set to [code]true[/code]
```
KoBeWi, line 13:
```suggestion
			If [code]false[/code], the [member indeterminate] animation will be paused in the editor.
```

scene/gui/progress_bar.cpp
```diff
@@ -130,9 +179,20 @@ void ProgressBar::_notification(int p_what) {
 	}
 }
 
+void ProgressBar::_validate_property(PropertyInfo &p_property) const {
+	Control::_validate_property(p_property);
```
KoBeWi, line 183:
```suggestion
```
This method is multilevel, so no need to call parent class.

scene/gui/progress_bar.cpp
```diff
@@ -153,14 +213,53 @@ bool ProgressBar::is_percentage_shown() const {
...
+}
+
+void ProgressBar::set_editor_preview_indeterminate(bool p_preview_indeterminate) {
+	if (editor_preview_indeterminate == p_preview_indeterminate) {
+		return;
+	}
+	editor_preview_indeterminate = p_preview_indeterminate;
+	_inderminate_fill_progress = 0;
+
+	notify_property_list_changed();
```
KoBeWi, line 240:
```suggestion
```
It does not affect any property, so this should not be called.

scene/gui/progress_bar.cpp
```diff
@@ -153,14 +213,53 @@ bool ProgressBar::is_percentage_shown() const {
...
+bool ProgressBar::is_indeterminate() const {
+	return indeterminate;
+}
+
+void ProgressBar::set_editor_preview_indeterminate(bool p_preview_indeterminate) {
+	if (editor_preview_indeterminate == p_preview_indeterminate) {
+		return;
+	}
+	editor_preview_indeterminate = p_preview_indeterminate;
+	_inderminate_fill_progress = 0;
```
KoBeWi, line 238: This will reset the value even at runtime. Maybe you should add a check for editor hint.
Or even
```suggestion
	if (!Engine::get_singleton()->is_editor_hint()) {
		return
	}
```
Switching the flag is ok, but that method should have no other effect at runtime.

scene/gui/progress_bar.cpp
```diff
@@ -50,9 +50,58 @@ Size2 ProgressBar::get_minimum_size() const {
 
 void ProgressBar::_notification(int p_what) {
 	switch (p_what) {
+		case NOTIFICATION_INTERNAL_PHYSICS_PROCESS: {
+			if (is_visible_in_tree() && (!Engine::get_singleton()->is_editor_hint() || editor_preview_indeterminate)) {
```
KoBeWi, line 54: It would be best if the processing wasn't enabled at all. That would make the editor part of this check unnecessary.

Calinou (commented):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/9adb7c7d130c6d4eb0e80b92d6eebd71fac3384d), it works as expected.

The progress bar should be updated in `NOTIFICATION_PROCESS`, not `NOTIFICATION_PHYSICS_PROCESS`. Otherwise, it looks choppy if **Physics Ticks Per Second** is decreased to a low value (here, `20`):

https://github.com/godotengine/godot/assets/180032/6d1fc8ee-ded8-4d0c-a843-e740986926d6

This is also noticeable on high refresh rate monitors with the default value (`60`).

I noticed **Show Percentage** becomes read-only in the inspector when **Indeterminate** is enabled, but typically, we hide incompatible properties instead of making them read-only:

image

Also, when you disable **Preview Indeterminate** after enabling it, it should probably go back to its default state (with the indeterminate marker halfway through the progress bar) so you don't have to save and reload the scene to put the marker where it was previously.

ckaiser:
Changed it to internal process and tweaked the numbers a little to get speeds right.

My justification for the read only state is that I considered the indeterminate state as an "override" of sorts, usually something that will precede an actual loading state, so I figured it'd be nice to be able to see if the percentage is going to be enabled when the initial indeterminate state is changed by a script in future, but I can make the adjustment with no issue.

Calinou (approved):
Works great now :slightly_smiling_face: 

I've tested pausing the game and it works as expected too (the animation pauses unless its process mode is set to **Always**). Time scale also affects the animation speed, which I think is expected. To get more control over that behavior, we'd need https://github.com/godotengine/godot-proposals/issues/7775 to be implemented.

AThousandShips (approved):
Looks good otherwise! 🙂

KoBeWi (approved):

scene/gui/progress_bar.cpp
```diff
@@ -50,9 +50,58 @@ Size2 ProgressBar::get_minimum_size() const {
...
+				if (Engine::get_singleton()->is_editor_hint() && !editor_preview_indeterminate) {
+					// Center the filled bar when we're not previewing the animation
+					_inderminate_fill_progress = (MAX(size.width, size.height) / 2) + (fill_size / 2);
+				}
+
+				switch (mode) {
+					case FILL_END_TO_BEGIN:
+					case FILL_BEGIN_TO_END: {
+						// Follow the RTL layout with the animation to match how the bar would fill
+						bool right_to_left = is_layout_rtl() ? (mode == FILL_BEGIN_TO_END) : (mode == FILL_END_TO_BEGIN);
```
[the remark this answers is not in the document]
ckaiser: Agreed, changed the line in the normal drawing code to match this one as well

akien-mga:
Thanks!

PR #55387, opened 2021-11-27
Turbulence/Noise for ParticleMaterial (2D / 3D)

RPicster:
It's a feature I miss a lot and brings much more life and interesting patterns to almost any Particle System.

I uploaded a longer demo video showing the features and parameters in action:
https://www.youtube.com/watch?v=oTsEaKpzseE

**Here are some small examples:**

https://user-images.githubusercontent.com/9423774/161746807-e6ada9fb-2381-45cd-95d0-717283b7ed3a.mp4

https://user-images.githubusercontent.com/9423774/161746822-3a38f111-6bc1-4df4-a0bf-5b399a324ed4.mp4

https://user-images.githubusercontent.com/9423774/161746837-5c6121e2-f5aa-41f0-9e6e-436919daff65.mp4

https://user-images.githubusercontent.com/9423774/161746903-785a67c8-c566-4c04-99dc-b053bbcf6bde.mp4

https://user-images.githubusercontent.com/9423774/161746914-aff4973a-729e-412b-8a28-e7e9ee0644b8.mp4

**It comes with the following parameters**
- Turbulence Mode: Control if turbulence will be used and the turbulence style.
- Turbulence Noise Contrast: Controls the contrast of the turbulence noise.
- Turbulence Noise Scale: Controls the overall scale of the turbulence noise pattern
- Turbulence Noise Speed: Controls the movement speed of the turbulence noise pattern (vec3)
- Turbulence Noise Speed Random: Randomizes the Noise Speed - this helps to break up visual patterns if they are visible.
- Turbulence Influence over life : This controls how much each particle will be influenced by the turbulence over it's lifetime
- Position Influence : How much the turbulence influences each particles position (translation)
- Velocity Influence : How much the turbulence influences each particles velocity
- Color Ramp : Offers the option to multiply each particles color with this ramp depending on the turbulence direction

turbulence_options

(The original pull request also had rotation and scale influece, those were removed to reduce complexity)

I also took care that **collisions** look good and happen in a way that makes sense (no further turbulence gets applied after a collision).

The newly added code snippets for the noise in the shader have been checked with the original authors and both gave their (very friendly) approval to be included in the engine code.

**TO DO:**
- [x] Find out if licensed code parts are ok to use (Got feeback from Inigo Quilez that it's no problem)
[...]

PR #87320, opened 2024-01-18
Fix collided particles getting stuck with zero velocity.

### LINKED ISSUE #87221: GPU Particle Attractor has no effect on stationary particles (GPUParticle3D)
### Tested versions

### System information

4.2 stable, Windows 10 - Vulkan Forward+ - Nvidia GeForce RTX 2070 Super

### Issue description

I'm trying to use a particle attractor to shoot away particles that have landed (collided) on a surface. The attractor works fine as long as they are moving, but when they land on the surface and become still the attractor don't have any effect on the particles.

I have set the Collision to Rigid and Friction to 0. I've tried Bounce from 0 and above (with a bounce of 1 the attractor work as the particle has bounced off the surface and is in the air again).

### Steps to reproduce

Create a GPU Particle 3D
Create a GPU Particle Collision Box
Create a GPU Particle Attractor Box 3D

Assign a sphere shape to the Particle and turn Collision on, set to Rigid and Friction to 0 and Bounce to 0
Have the particles land on the Collision box and become still
Move the Attractor over the particles to have them move. They are not affected by the Attractor

[ParticleTestProjectX.zip](https://github.com/godotengine/godot/files/13946986/ParticleTestProjectX.zip)

### LINKED ISSUE #91346: Particles which only have a velocity but no acceleration don't collide properly
### Tested versions

v4.2.2.stable.mono.official [15073afe3]

### System information

Godot v4.2.2.stable.mono - Windows 10.0.22631 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4080 (NVIDIA; 31.0.15.5222) - AMD Ryzen 9 7950X3D 16-Core Processor (32 Threads)

### Issue description

When a particle system has no set acceleration then the collision does not work unless particle speeds are high. See this video:

https://github.com/godotengine/godot/assets/327257/fc2a90f9-32ba-46cc-9e88-5bb2279ff6d7

The left particle system has no initial velocity and only set gravity to -5. The right particle system has an initial velocity of 3.  In both particle systems the particles have visually similar speeds, yet their collision behaviour is totally different. With a velocity of 3 the particles will just stick on the surface. Only if I increase the velocity the particles will actually behave similarly to the particle system with acceleration.

### Steps to reproduce

Check the `particles.tscn` in the attached example project. Both particle systems are set up exactly the same, except that the left one uses acceleration while the right one uses velocity.

[collision_reproducer.zip](https://github.com/godotengine/godot/files/15159121/collision_reproducer.zip)

Rudolph-B:
Resolves https://github.com/godotengine/godot/issues/87221 

Changes the behavior of slow collided particles from being set to completely stationary to only removing the part of the velocity that will move the particle into the obstruction.

Project showcasing the change:
[CollisionTest.zip](https://github.com/godotengine/godot/files/13971341/CollisionTest.zip)

It is basically GPUParticles3D falling on a GPUParticlesCollisionBox3D with a GPUParticlesAttractorBox3D trying to push the particles of.

The issue can be seen [here](https://github.com/godotengine/godot/assets/52322745/35111322-061f-4f43-a03c-57c08282e337). The particles get stuck even though this is supposed to be a friction less environment.

How the changes fix the issue can be seen [here](https://github.com/godotengine/godot/assets/52322745/f91ef0f4-32e2-4ea4-8eaf-bf0aec126826). The particles glide on top of the obstruction. And with the the GPUParticlesAttractorBox3D disabled particles still fall normally and become stationary as can be seen [here](https://github.com/godotengine/godot/assets/52322745/14b3e369-5344-4cdf-b00e-1afc2effc722)

An other possible fix was to adjust the if statement found in [particle_process_material.cpp](https://github.com/godotengine/godot/blob/master/scene/resources/particle_process_material.cpp):991
```
		code += "		if (length(VELOCITY) > 3.0) {\n";
		code += "			TRANSFORM[3].xyz += COLLISION_NORMAL * COLLISION_DEPTH;\n";
		code += "			VELOCITY -= COLLISION_NORMAL * dot(COLLISION_NORMAL, VELOCITY) * (1.0 + collision_bounce);\n";
		code += "			VELOCITY = mix(VELOCITY,vec3(0.0),clamp(collision_friction, 0.0, 1.0));\n";
		code += "		} else {\n";
		code += "			VELOCITY = vec3(0.0);\n";
		code += "		}\n";
```
To be less strict with `length(VELOCITY) > 1.0` or something similar. This produced excessively bouncy particles though.

_bugsquad edit: Fixes https://github.com/godotengine/godot/issues/91346_

### DIFF
--- a/scene/resources/particle_process_material.cpp
+++ b/scene/resources/particle_process_material.cpp
@@ -995,13 +995,15 @@ void ParticleProcessMaterial::_update_shader() {
 	code += "	\n";
 	if (collision_mode == COLLISION_RIGID) {
 		code += "	if (COLLIDED) {\n";
-		code += "		if (length(VELOCITY) > 3.0) {\n";
-		code += "			TRANSFORM[3].xyz += COLLISION_NORMAL * COLLISION_DEPTH;\n";
-		code += "			VELOCITY -= COLLISION_NORMAL * dot(COLLISION_NORMAL, VELOCITY) * (1.0 + collision_bounce);\n";
-		code += "			VELOCITY = mix(VELOCITY,vec3(0.0),clamp(collision_friction, 0.0, 1.0));\n";
-		code += "		} else {\n";
-		code += "			VELOCITY = vec3(0.0);\n";
-		code += "		}\n";
+		code += "		float collision_response = dot(COLLISION_NORMAL, VELOCITY);\n";
+		code += "		float slide_to_bounce_trigger = step(2.0/clamp(collision_bounce + 1.0, 1.0, 2.0), abs(collision_response));\n";
+		code += "		TRANSFORM[3].xyz += COLLISION_NORMAL * COLLISION_DEPTH;\n";
+		code += "		// Remove all components of VELOCITY that is not tangent to COLLISION_NORMAL\n";
+		code += "		VELOCITY -= COLLISION_NORMAL * collision_response;\n";
+		code += "		// Apply friction only to VELOCITY across the surface (Effectively decouples friction and bounce behavior).\n";
+		code += "		VELOCITY = mix(VELOCITY,vec3(0.0),clamp(collision_friction, 0.0, 1.0));\n";
+		code += "		// Add bounce velocity to VELOCITY\n";
+		code += "		VELOCITY -= COLLISION_NORMAL * collision_response * (collision_bounce * slide_to_bounce_trigger);\n";
 		code += "	}\n";
 	} else if (collision_mode == COLLISION_HIDE_ON_CONTACT) {
 		code += "	if (COLLIDED) {\n";


clayjohn:
It seems that this if statement comes from https://github.com/godotengine/godot/pull/55387

@RPicster Mentions the reason for zeroing the velocity as:
> I also took care that collisions look good and happen in a way that makes sense (no further turbulence gets applied after a collision).

This PR will reintroduce that issue. I have a feeling that we will need to do something more intrusive. Right now the logic is just suppressing any movement at all if there is a collision and the velocity is low. But that has two problems:
1. Its essentially an infinite amount of friction for slow moving particles
2. The cutoff is arbitrary

The bounciness you see from decreasing the VELOCITY threshold comes from the fact that the particles are set to be a bit bouncy by default and this logic applies the bounce every frame (i.e. bounce amount isn't effectively scaled by velocity appropriately). I think the issue is that only the magnitude of VELOCITY is taken into account when calculating the bounce. So a large horizontal velocity translates into a large movement along the normal, which doesn't make a lot of sense.

As a follow up I would check a few things:
1. Is it possible to suppress the noise without setting VELOCITY to 0? If so, then we can remove this if branch completely and just suppress the noise when the magnitude of VELOCITY is less than 0
2. Can we modify the bounce term to take direction of VELOCITY into account (more than just as a scale factor) (perhaps multiplying by the absolute direction of VELOCITY would be enough?)

RPicster:
Please make sure to test your intended behavior with all kinds of settings and velocities/accelerations. Turbulence including.

Changing things in particle behavior may work under certain circumstances (eg for your use case), but may break others.

Rudolph-B:
I've made updates to the code: I replaced the if statement with a step function to handle the transition from bouncing to sliding in particles. The step function uses the particle's velocity along the collision normal to determine how long it bounces before it starts sliding. Additionally, particles defined with higher bounciness will also bounce longer before sliding.

I have adjusted my test project to include some more test cases:
[CollisionTest.zip](https://github.com/godotengine/godot/files/15164388/CollisionTest.zip)

Everything looks good to me but I am not really sure how the interaction with turbulence is supposed to work.

scene/resources/particle_process_material.cpp
```diff
@@ -988,13 +988,11 @@ void ParticleProcessMaterial::_update_shader() {
...
-		code += "			VELOCITY -= COLLISION_NORMAL * dot(COLLISION_NORMAL, VELOCITY) * (1.0 + collision_bounce);\n";
-		code += "			VELOCITY = mix(VELOCITY,vec3(0.0),clamp(collision_friction, 0.0, 1.0));\n";
-		code += "		} else {\n";
-		code += "			VELOCITY = vec3(0.0);\n";
-		code += "		}\n";
+		code += "		float collision_response = dot(COLLISION_NORMAL, VELOCITY);\n";
+		code += "		float collision_trigger = step(2, length(collision_response)/clamp(collision_bounce + 1.0, 1.0, 2.0));\n";
+		code += "		TRANSFORM[3].xyz += COLLISION_NORMAL * COLLISION_DEPTH;\n";
+		code += "		VELOCITY -= COLLISION_NORMAL * collision_response * (1.0 + collision_bounce * collision_trigger);\n";
+		code += "		VELOCITY = mix(VELOCITY,vec3(0.0),clamp(collision_friction, 0.0, 1.0));\n";
```
Ansraer, line 995: I am not happy with the fact that friction is always applied. IMO the collision friction should depend on the direction the particle is trying to move, so that when the particle is trying to move e.g. straight away from the surface friction should be 0.

scene/resources/particle_process_material.cpp
```diff
@@ -988,13 +988,11 @@ void ParticleProcessMaterial::_update_shader() {
...
 		code += "	if (COLLIDED) {\n";
-		code += "		if (length(VELOCITY) > 3.0) {\n";
-		code += "			TRANSFORM[3].xyz += COLLISION_NORMAL * COLLISION_DEPTH;\n";
-		code += "			VELOCITY -= COLLISION_NORMAL * dot(COLLISION_NORMAL, VELOCITY) * (1.0 + collision_bounce);\n";
-		code += "			VELOCITY = mix(VELOCITY,vec3(0.0),clamp(collision_friction, 0.0, 1.0));\n";
-		code += "		} else {\n";
-		code += "			VELOCITY = vec3(0.0);\n";
-		code += "		}\n";
+		code += "		float collision_response = dot(COLLISION_NORMAL, VELOCITY);\n";
+		code += "		float collision_trigger = step(2, length(collision_response)/clamp(collision_bounce + 1.0, 1.0, 2.0));\n";
```
Ansraer, line 992: Since collision_response is a float you should probably use abs instead of length.

Ansraer (changes requested):

clayjohn:
Testing out the new MRP, I can confirm that, with this PR, turbulence is free to drag particles across the surface of the collider. I can also confirm that adding a bit of friction stops that. To me, that is what is intuitive.

@RPicster When you check this out, please pay special attention to distinguish between particles sliding across the surface of collisions due to turbulence, and particles jittering on the surface of collisions due to turbulence. I suspect that the bouncing/jittering is the behaviour that was most annoying before as the sliding can easily be addressed by adding friction.

RPicster:
> Testing out the new MRP, I can confirm that, with this PR, turbulence is free to drag particles across the surface of the collider. I can also confirm that adding a bit of friction stops that. To me, that is what is intuitive.
> 
> @RPicster When you check this out, please pay special attention to distinguish between particles sliding across the surface of collisions due to turbulence, and particles jittering on the surface of collisions due to turbulence. I suspect that the bouncing/jittering is the behaviour that was most annoying before as the sliding can easily be addressed by adding friction.

I tested this PR in different scenarios and I find the behaviour very nice. It works like Clay described it. Increasing friction will give control over the behaviour.

Get's a 👍 from my side.

One thing I noticed and it was mentioned by Clay in the Rendering Meeting was the "distance" from the particle to the surface after collision - so after turning up the collision size of the particles I observed bouncing particles on the surface.

I also tested this in 4.3dev6 and this PR and it seems that this behaviour is not great in this PR.
Not saying it was "great" before, but the problem that was already there is now much more visible and visually impactful.

I think it is related to interpolation/fixed FPS.


This is 4.3dev6 - you can observe the particle is just stopped on collision, sometimes at the wrong distance:
https://github.com/godotengine/godot/assets/9423774/af4448bf-77e3-4e6d-ad44-a7ec49a188f2


This PR - the particle starts bouncing in and out of the surface. Increasing fixed FPS makes it much less visible.
https://github.com/godotengine/godot/assets/9423774/b8e3765f-96b2-47a2-a25d-659c9de44477

I'd say it would be great if this could be fixed 🥇

scene/resources/particle_process_material.cpp
Rudolph-B: I have changed it to use abs()

scene/resources/particle_process_material.cpp
Rudolph-B: I have done this by basically splitting the VELOCITY into 3 steps:
 1. Remove the component of the velocity pushing the particle further into the surface.
 2. Apply friction
 3. Add bounce component of velocity

Rudolph-B:
I have made the adjustments requested by @Ansraer.

I am still looking into the problem @RPicster noted.

Rudolph-B:
@RPicster Can you maybe share a MRP. I am struggling to clearly see the issue.

RPicster:
> @RPicster Can you maybe share a MRP. I am struggling to clearly see the issue.
[bouncy.zip](https://github.com/godotengine/godot/files/15196792/bouncy.zip)
Here is a scene where the issue is easy to spot 👍

Rudolph-B:
I believe I have fixed the issue. What happened was, after a particle collided, it was move to just touch the surface of the collider. When determining whether a particle is collided this particle then often didn't count as collided. In cases where the particle then instantly started moving again in the direction of the collider it would cause a jarring snap like affect. Especially at lower FPS. 

The solution I implemented was to add an EPSILON term to all 4 collision calculations in `servers/rendering/renderer_rd/shaders/particles.glsl`. The goal being that particles stay collided until they move away from the collider.

The below project has a scene name "bouncy" showcasing the behavior for all 4 3D collision shapes.
[CollisionTest.zip](https://github.com/godotengine/godot/files/15210291/CollisionTest.zip)

@RPicster Can you please test this again.

I have not yet checked other collision system if they have similar issues.

scene/resources/particle_process_material.cpp
```diff
@@ -995,13 +995,12 @@ void ParticleProcessMaterial::_update_shader() {
...
-		code += "			TRANSFORM[3].xyz += COLLISION_NORMAL * COLLISION_DEPTH;\n";
-		code += "			VELOCITY -= COLLISION_NORMAL * dot(COLLISION_NORMAL, VELOCITY) * (1.0 + collision_bounce);\n";
-		code += "			VELOCITY = mix(VELOCITY,vec3(0.0),clamp(collision_friction, 0.0, 1.0));\n";
-		code += "		} else {\n";
-		code += "			VELOCITY = vec3(0.0);\n";
-		code += "		}\n";
+		code += "		float collision_response = dot(COLLISION_NORMAL, VELOCITY);\n";
+		code += "		float slide_to_bounce_trigger = step(2.0/clamp(collision_bounce + 1.0, 1.0, 2.0), abs(collision_response));\n";
+		code += "		TRANSFORM[3].xyz += COLLISION_NORMAL * COLLISION_DEPTH;\n";
+		code += "		VELOCITY -= COLLISION_NORMAL * collision_response;\n";
```
clayjohn, line 1001:
```suggestion
		// First de-penetrate the particle.
		code += "		VELOCITY -= COLLISION_NORMAL * collision_response;\n";
```

scene/resources/particle_process_material.cpp
```diff
@@ -995,13 +995,12 @@ void ParticleProcessMaterial::_update_shader() {
...
+		code += "		VELOCITY = mix(VELOCITY,vec3(0.0),clamp(collision_friction, 0.0, 1.0));\n";
```
clayjohn, line 1002:
```suggestion
		// Then apply friction across the surface.
		code += "		VELOCITY = mix(VELOCITY,vec3(0.0),clamp(collision_friction, 0.0, 1.0));\n";
```

scene/resources/particle_process_material.cpp
```diff
@@ -995,13 +995,12 @@ void ParticleProcessMaterial::_update_shader() {
...
+		code += "		VELOCITY -= COLLISION_NORMAL * collision_response * (collision_bounce * slide_to_bounce_trigger);\n";
```
clayjohn, line 1003:
```suggestion
		// Finally apply the bounce.
		code += "		VELOCITY -= COLLISION_NORMAL * collision_response * (collision_bounce * slide_to_bounce_trigger);\n";
```

clayjohn (commented):
As discussed in the rendering meeting. Just separate the particles.glsl stuff to a new PR (add it to OpenGL at the same time) and add some comments about the new order of operations for friction. Then should be good to merge!

Rudolph-B:
@clayjohn Added the comments and removed the particles.glsl stuff. Will make the new PR for particles.glsl a bit later I would first like to do some tests and see if 2D particle collisions have a similar issue.

clayjohn (approved):
Looks great to me! Great work

akien-mga:
Thanks!

RPicster:
It doesn't really add anything to this PR - But I just happen to come across the original problem this PR solves and I could apply the changed shader code in a "real-world" scenario.

It looks really perfect - so thank you again @Rudolph-B !

PR #87327, opened 2024-01-18
[Editor] Add pan and zoom sensitivity settings.

### LINKED ISSUE #8559: Increase the sensitivity of touchpad gestures on Mac

Anything 3D

Touchpad sensitivity is very low. This video shows what happens when I swipe across the entire surface of a 16-inch macbook touchpad (it's a really big touchpad) in Blender vs Godot:

https://github.com/godotengine/godot-proposals/assets/60579014/0c562650-7c20-4298-8170-1a44c205896e

Blender does a complete 360, Godot barely moves. The results are similar when trying to pan with shift+swipe. Interestingly, zoom with pinch is much better, zoom doesn't need any adjustments. There's a discussion mentioning the issue here https://github.com/godotengine/godot-proposals/discussions/8532, but I thought this deserves a proposal as it's probably not hard to fix

Increase the sensitivity of orbit and pan gestures

If orbit and pan behave similarly to Blender that would be great

N/A

N/A

bruvzg:
Default setting is too slow for Mac trackpad (but OK for mouse). And there were complaints about it being both too high and too low, so it should be user configurable.

Fixes https://github.com/godotengine/godot-proposals/issues/8559

### DIFF
--- a/doc/classes/EditorSettings.xml
+++ b/doc/classes/EditorSettings.xml
@@ -289,7 +289,7 @@
 			See also [member editors/3d/navigation/navigation_scheme].
 		</member>
 		<member name="editors/3d/freelook/freelook_sensitivity" type="float" setter="" getter="">
-			The mouse sensitivity to use while freelook mode is active in the 3D editor. See also [member editors/3d/navigation_feel/orbit_sensitivity].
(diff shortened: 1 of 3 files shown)


passivestar:
I tested it and it works, but unfortunately this doesn't solve my problem. My mouse sensitivity is fine, the problem is that the touchpad is slow relative to mouse. Touchpad needs separate sensitivity sliders, otherwise I'll have to change sensitivity every time I'm not using my mouse

ryevdokimov (commented):
Shouldn't this also affect scroll zooming?

For example (not sure if this works as intended, this is just for reference.)

image

PR #87328, opened 2024-01-18
Remove unnecessary `_set_global_invalid` added by mistake

groud:
This mistake was introduced in https://github.com/godotengine/godot/pull/87115 but the issue was already solved in https://github.com/godotengine/godot/pull/86841.

The call to `_set_global_invalid` is indeed already done at line 319.

(I marked it as a bug, but I don't think it has any impact. I marked it a "bug" because it's solving a mistake. If that makes any sense :sweat_smile: )


akien-mga (approved):

Sauermann (approved):

akien-mga:
Thanks!

PR #87331, opened 2024-01-18
Simplify RefCounted check in `free()`

KoBeWi:
Use the `is_ref_counted()` method, which does a simple bool check, instead of casting.

### DIFF
--- a/core/object/object.cpp
+++ b/core/object/object.cpp
@@ -728,7 +728,7 @@ Variant Object::callp(const StringName &p_method, const Variant **p_args, int p_
 			r_error.expected = 0;
 			return Variant();
 		}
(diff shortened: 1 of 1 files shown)


KoBeWi:
It's initialized directly in the RefCounted constructor:
https://github.com/godotengine/godot/blob/1952f64b07b2a0d63d5ba66902fd88190b0dcf08/core/object/ref_counted.cpp#L96-L97

https://github.com/godotengine/godot/blob/1952f64b07b2a0d63d5ba66902fd88190b0dcf08/core/object/object.cpp#L1958-L1963

No other class overrides it, so it's safe to rely on.

akien-mga (approved):

akien-mga:
Thanks!

PR #87333, opened 2024-01-18
[Editor] Fix theme switch icon and text color update issue

alula:
Fixes a minor issue I just spotted where text remains color that matches previously set theme, eg. text remains dark if you switched from light theme.


alula:
seems to be handled by #87293

PR #63563, opened 2022-07-28
Update export dialog to handle many architectures

aaronfranke:
This PR has two commits. The first commit adds `const` to a few places, to make it clear that these methods do not modify anything in the class.

The second commit is the bulk of this PR, it updates the export dialog to handle many architectures. This is fully compatible with the new export template names in https://github.com/godotengine/godot/pull/63096

* For Windows, Linux, and UWP, they get a dropdown menu called `binary_format/architecture`.
  * This means that `binary_format/64_bits` is now gone! (not to be confused with `bits=` in the build system)
  * UWP already had a dropdown before, but this PR makes it more consistent with the others.
  * Windows, Linux, and UWP all default to x86_64.
  * The only options for Windows are x86_64 and x86_32. When it's possible to build Godot for Windows for ARM, we'll add arm64 (and maybe arm32?) to the list.
* For macOS, it will use "universal", which is both x86_64 and arm64 at the same time.
* For Android and iOS, I did not touch them, except I added a comment for Android. These platforms already have checkboxes for which architectures to include in the app bundle.
* For the last platform we support, web, I did not touch it. When 64-bit WebAssembly inevitably starts existing and we want to support it, we will need to modify the web platform's code.

I will note that most users will not need to change the architecture, since most desktops are x86_64, and for macOS the universal architecture will work on all supported systems. But it's good to have this "properly" supported.

Here is a screenshot of what the dropdown looks like for Linux (which has the most options by *far*, and to be clear, it's not at all expected that the Godot organization provides all of these officially):

PR #87338, opened 2024-01-18
[3.x] Linux: Add support for arm64 and arm32 export templates

akien-mga:
This is done in a hacky way, mostly to keep it simple and avoid having to do a refactoring of the `EditorExportPlatform` interface.

Only Windows and Linux use `EditorExportPlatformPC`, and thus to handle the new architectures for Linux, we simply do a few checks here and there with a couple new methods to register the export template names for Linux arm64 and arm32.

We also don't change the name of the x86_64 and x86_32 templates, so that users who have custom build pipelines can keep using them without major disruption. ~The only compat breakage is the removal of `binary_format/64_bits`, so users with an export preset where this was *off* would need to reconfigure it to select `x86_32` as `binary_format/architecture`. _Might_ be possible to add compat code to handle this gracefully, I haven't looked yet.~

For Godot 4.0, we did refactor everything in #63563 to allow exporting binaries for different architectures cleanly. For 3.6, which is likely the last feature release for the 3.x branch, I tend to cut corners as these improvements will be shorter lived and thus new tech debt isn't as big a concern.

### DIFF
--- a/editor/editor_export.cpp
+++ b/editor/editor_export.cpp
@@ -69,6 +69,17 @@ bool EditorExportPreset::_set(const StringName &p_name, const Variant &p_value) 
 		}
 		return true;
 	}
+#ifndef DISABLE_DEPRECATED
+	// Compatibility with new `binary_format/architecture` for Linux in 3.6+.
+	else if (p_name == "binary_format/64_bits" && values.has("binary_format/architecture")) {
+		values["binary_format/architecture"] = (bool)p_value ? "x86_64" : "x86_32";
+		EditorExport::singleton->save_presets();
+		if (update_visibility["binary_format/architecture"]) {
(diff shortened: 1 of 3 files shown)


YuriSizov (approved):
I haven't tested it, but the general approach makes sense to me.

akien-mga:
> The only compat breakage is the removal of `binary_format/64_bits`, so users with an export preset where this was _off_ would need to reconfigure it to select `x86_32` as `binary_format/architecture`. _Might_ be possible to add compat code to handle this gracefully, I haven't looked yet.

This was easy to do, so I implemented it. So existing `export_presets.cfg` configs for Linux x86_32 (`binary_format/64_bits` off) will still work and be converted to `binary_format/architecture = "x86_32"`.

Issue #85547, opened 2023-11-30
[4.2] Windows 11 window decoration shadows flicker constantly while Godot is running


HybridEidolon:
### Godot version

v4.2.stable.official [46dc27791]

### System information

Godot v4.2.stable - Windows 10.0.22621 - Vulkan (Mobile) - dedicated AMD Radeon RX 6800 XT (Advanced Micro Devices, Inc.; 31.0.22023.1014) - 13th Gen Intel(R) Core(TM) i7-13700KF (24 Threads)

### Issue description

In these conditions:

1. Running Windows 11 with AMD driver
2. on a display with Variable Refresh Rate/"Freesync" enabled (possibly unrelated)
3. Using the Forward+ or Mobile renderers (not Compatibility)

the shadow window decoration on _every_ window (not just Godot's) flickers _constantly,_ even when the game or editor itself is out of focus.

### Steps to reproduce

1. Make a scene in the Forward+ or Mobile renderers
2. Run the game

### Minimal reproduction project

[shadowflicker.zip](https://github.com/godotengine/godot/files/13514642/shadowflicker.zip)

HybridEidolon:
Issue is still present in 4.3 dev 1.

HybridEidolon:
This is almost entirely resolved by 3be3d50 (#87340). Not sure if I should close and create a new one. The shadows only flicker very rarely when dragging the window around, and more frequently when resizing it.

akien-mga:
Thanks for testing!

Opening a new issue for the remaining symptoms would be good.

Let's close this one as fixed by #87340.

PR #80566, opened 2023-08-12
Fix Vulkan CPU/GPU synchronization

darksylinc:
The original code had two bugs:

1. It assumed Swapchain count can be used for proper CPU/GPU synchronization (it's not)
2. It used a double buffer scheme (FRAME_LAG = 2) but the vkWaitForFences/vkResetFences calls were in the wrong order and in the wrong place
3. The swapchain bug forced a quad-buffer scheme with roughly 1 or 2 frames of wait; which hid all the issues

This commit gets rid of `RenderingDeviceVulkan::frame_count` & `RenderingDeviceVulkan::frame` which should've never existed and makes everything depend on `VulkanContext::frame_index` &
`VulkanContext::FRAME_LAG`

See https://github.com/godotengine/godot/issues/80550 for details

Fix #80550
Likely fixes https://github.com/godotengine/godot/issues/81670.

## Specific things that need testing:

Aside from overall stability and ensure there are no regressions

- [ ] See if there is an improvement for DEVICE_LOST users. #71929
- [ ]  See if users who reported higher latency/lag vs OpenGL renderer report improvements #71795
- [x] Test if performance changes (for better or worse)
- [x] Test FRAME_LAG = 3. The buggy behavior had the latency disadvantages of triple buffering with the performance disadvantages of double buffering. Perhaps we can get a speed boost with the same latency as before by using FRAME_LAG = 3

**Note:** This PR may uncover [bugs that were previously hidden](https://github.com/godotengine/godot/issues/80570) or that were rare to reproduce. This may give the illusion that the code has become more unstable when in reality it's just uncovering hidden flaws.

Issue #87309, opened 2024-01-17
Godot editor (or an exported application) doesn't respond after being minimized in the dock and restored with a mouse click on MacOS


OKGougou:
### Tested versions

- Reproducible in: 4.2.1-stable, 4.2-stable
- Not reproducible in: 4.1.3-stable

### System information

Godot v4.2.1.stable - macOS 13.6.3 - Vulkan (Forward+) - integrated Apple M1 Pro - Apple M1 Pro (8 Threads)

### Issue description

After being minimized in the dock, the restored editor window becomes unresponsive if restored with a mouse click in the dock window list.

### Steps to reproduce

Open a Godot 4.2 or 4.2.1 project on MacOS.
Minimize Godot in the dock.
Click on the minimized window in the dock window list.
Try interacting with Godot.

Expected behavior:
Godot should respond to mouse and keyboard interactions.

Actual behavior:
The user interface does not respond to mouse and keyboard interaction.

The problem does not arise if the window is restored from the Godot application icon in the dock, nor if the window restore keyboard shortcut is used.
(To use the keyboard shortcut, select Godot with Command+Tab, hold down Command, press Option+Command, release Command while pressing Option).
If the unresponsive window is minimized in the dock and restored by a mouse-click on the Godot application icon in the dock or by a keyboard shortcut, the restored Godot window becomes active again and displays the changes it received when it was unresponsive.

### Minimal reproduction project (MRP)

No required project.

lostminds:
Tested and confirmed the same issue is still present in Godot 4.3.dev2 on macOS 14.2.1. 

Using the newly added Window menu (https://github.com/godotengine/godot/pull/83987) items to expand the minimized window restores the interactivity as expected. But just clicking the minimized window in the Dock still leads to it being opened and activated, but not reacting to any input events (mouse or keyboard). 

However, when experimenting a little it seems that it is actually processing the input events, it's just not redrawing the window to reflect the changes, so it appears as if it's not reacting.

PR #87340, opened 2024-01-18
Finish splitting functionality of the `RenderingDevice` backends into `RenderingDeviceDriver`.

### LINKED ISSUE #87532: Baking lightmaps always prints a warning message about PSO caching
### Tested versions

4.3.dev 0bcc0e92b

### System information

Godot v4.3.dev (0bcc0e92b) - Fedora Linux 39 (KDE Plasma) - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4090 (nvidia; 545.29.06) - 13th Gen Intel(R) Core(TM) i9-13900K (32 Threads)

### Issue description

When baking LightmapGI, I always get a warning message:

```
WARNING: There's already a RenderingDeviceDriverVulkan instance doing PSO caching. Only one can at the same time. This one won't.
     at: pipeline_cache_create (drivers/vulkan/rendering_device_driver_vulkan.cpp:2198)
```

Assuming this is a limitation we can't lift (and not a bug), I suggest silencing this particular instance when baking lightmaps since it's not relevant information for end users here. This could be done by moving it to `print_verbose()` or having a property to disable the warning in local RenderingDevices.

PS: The warning message could also mention which instance of the RenderingDevice it is (e.g. with an object ID), so you can distinguish it when multiple local RenderingDevices are printing it.

cc @DarioSamo @RandomShaper 

### Steps to reproduce

- Bake a LightmapGI node.

[test_lightmap_preview_bake_4.x.zip](https://github.com/godotengine/godot/files/14040292/test_lightmap_preview_bake_4.x.zip)

DarioSamo:
This is a follow up PR to @RandomShaper's PR https://github.com/godotengine/godot/pull/83452 with the intention of cleaning up the loose ends that weren't put behind an abstraction yet. This also includes the changes identified by @darksylinc's PR https://github.com/godotengine/godot/pull/80566 as well as the new project settings that allow configuration of the swap chain image count and the buffered frame count of RenderingDevice.

The main reason behind pushing for this change is that the additional flexibility provided by the PR is essential to providing new asynchronous transfer queues, which will lead to [massive improvements in loading times](https://twitter.com/dariosamo/status/1735363059202375819) (not included in this PR). This will also enable to provide additional multi-queue functionality in the future for the Acyclic Command Graph that was recently merged. However, the current design makes it impossible to implement this without making awkward changes to all backends and maintaining all of the backends at once.

Another important reason for this change is that this PR will finally enable the possibility of the RenderingDevice's local device working independently of the renderer that was chosen, **which means the GPU lightmapper can finally work in Compatibility mode!** @clayjohn has proven this to be possible in a separate experiment (not part of this PR but based on top of it) where he added the required UV2 rendering to the compatibility renderer.

The overall net result means a lot of the logic is now shared across backends and the only thing the backend needs to worry about is exposing the following functionality:
- Command Queues
- CPU/GPU Fences
- GPU/GPU Semaphores
- Swap Chains

The resulting code is also a lot easier to follow and it should be much simpler to fix some of the [strange bugs that have been plaguing](https://github.com/godotengine/godot/issues/71929) the backends for a while. However, I do not expect my first attempt at this to have been perfect, since swap chain logic can be particularly hard to handle correctly, so we'll need some extensive testing to confirm that this doesn't introduce any regressions.

NOTE: No performance improvements or regressions are expected from this PR, it is purely an organizational change and functionality wise it should be identical in the default case.

Confirm the changes work as intended in all the affected platforms.
- [x] Windows
- [x] Linux (X11 Systems)
- [x] Linux (Wayland)
- [x] Android
- [x] macOS
- [x] iOS
- [x] OpenXR (Desktop)
- [x] OpenXR (Quest)

TODO:
- [x] CmdBeginDebugUtilsLabelEXT and related functions should be fetched like the older versions.
- [x] Rendering Device should print the API name instead of "API".
- [x] There should be no need to print the information about the picked device for RDs that aren't the Singleton.
- [x] Fix all CI issues so it builds successfully in all platforms.
- [x] Provide compatibility methods if necessary for RenderingDevice on the functions that have been changed.
- [x] Finish documentation changes.

___

- *Production edit: This closes https://github.com/godotengine/godot/issues/87532.*


doc/classes/ProjectSettings.xml
```diff
@@ -2656,6 +2656,21 @@
...
+			1 may give you the lowest lag/latency but at the high cost of no parallelism between CPU &amp; GPU.
+			Try the [url=https://darksylinc.github.io/vsync_simulator/]V-Sync Simulator[/url], an interactive interface that simulates presentation to better understand how it is affected by different variables under various conditions.
+			[b]Note:[/b] This property is only read when the project starts. There is currently no way to change this value at run-time.
+		</member>
+		<member name="rendering/rendering_device/vsync/swapchain_image_count" type="int" setter="" getter="" default="3">
+			The number of images the swapchain will consist of (back buffers + front buffer).
+			[code]2[/code] corresponds to double-buffering and [code]3[/code] to triple-buffering. A value of [code]1[/code] is not recommended.
+			Double-buffering may give you the lowest lag/latency but if V-Sync is on and the system can't render at 60 fps, the framerate will go down in multiples of it (e.g. 30 fps, 15, 7.5, etc ). Triple buffering gives you higher framerate (specially if the system can't reach a constant 60 fps) at the cost of up to 1 frame of latency, with [constant DisplayServer.VSYNC_ENABLED] (FIFO).
```
AThousandShips, line 2668:
```suggestion
			Double-buffering may give you the lowest lag/latency but if V-Sync is on and the system can't render at 60 fps, the framerate will go down in multiples of it (e.g. 30 fps, 15, 7.5, etc.). Triple buffering gives you higher framerate (specially if the system can't reach a constant 60 fps) at the cost of up to 1 frame of latency, with [constant DisplayServer.VSYNC_ENABLED] (FIFO).
```

drivers/d3d12/rendering_context_driver_d3d12.cpp
```diff
@@ -0,0 +1,313 @@
...
+	LocalVector<IDXGIAdapter1 *> adapters;
+	IDXGIAdapter1 *adapter = nullptr;
+	do {
+		adapter = create_adapter(adapters.size());
+		if (adapter != nullptr) {
+			adapters.push_back(adapter);
+		}
+	} while (adapter != nullptr);
+
+	ERR_FAIL_COND_V_MSG(adapters.size() == 0, ERR_CANT_CREATE, "Adapters enumeration reported zero accessible devices.");
```
AThousandShips, line 144:
```suggestion
	ERR_FAIL_COND_V_MSG(adapters.is_empty(), ERR_CANT_CREATE, "Adapters enumeration reported zero accessible devices.");
```

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -1772,25 +1861,173 @@ void RenderingDeviceDriverD3D12::command_pipeline_barrier(
...
+/**** FENCES ****/
+/****************/
+
+RDD::FenceID RenderingDeviceDriverD3D12::fence_create() {
+	ComPtr<ID3D12Fence> d3d_fence;
+	HRESULT res = device->CreateFence(0, D3D12_FENCE_FLAG_NONE, IID_PPV_ARGS(d3d_fence.GetAddressOf()));
+	ERR_FAIL_COND_V(!SUCCEEDED(res), FenceID());
+
+	HANDLE event_handle = CreateEvent(nullptr, FALSE, FALSE, nullptr);
+	ERR_FAIL_COND_V(event_handle == nullptr, FenceID());
```
AThousandShips, line 1874:
```suggestion
	ERR_FAIL_NULL_V(event_handle, FenceID());
```

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -5408,109 +5828,394 @@ uint64_t RenderingDeviceDriverD3D12::api_trait_get(ApiTrait p_trait) {
...
+	err = create_command_signature(device.Get(), D3D12_INDIRECT_ARGUMENT_TYPE_DISPATCH, sizeof(D3D12_DISPATCH_ARGUMENTS), &indirect_cmd_signatures.dispatch);
+	ERR_FAIL_COND_V(err != OK, ERR_CANT_CREATE);
+
+	return OK;
+}
+
+Error RenderingDeviceDriverD3D12::initialize(uint32_t p_device_index, uint32_t p_frame_count) {
+	context_device = context_driver->device_get(p_device_index);
+	adapter = context_driver->create_adapter(p_device_index);
+	ERR_FAIL_COND_V(adapter == nullptr, ERR_CANT_CREATE);
```
AThousandShips, line 6191:
```suggestion
	ERR_FAIL_NULL_V(adapter, ERR_CANT_CREATE);
```

drivers/vulkan/rendering_context_driver_vulkan.cpp
```diff
@@ -0,0 +1,690 @@
...
+	String objects_string;
+	if (p_callback_data->objectCount > 0) {
+		objects_string = "\n\tObjects - " + String::num_int64(p_callback_data->objectCount);
+		for (uint32_t object = 0; object < p_callback_data->objectCount; ++object) {
+			objects_string +=
+					"\n\t\tObject[" + String::num_int64(object) + "]" +
+					" - " + string_VkObjectType(p_callback_data->pObjects[object].objectType) +
+					", Handle " + String::num_int64(p_callback_data->pObjects[object].objectHandle);
+
+			if (nullptr != p_callback_data->pObjects[object].pObjectName && strlen(p_callback_data->pObjects[object].pObjectName) > 0) {
```
AThousandShips, line 264:
```suggestion
			if (p_callback_data->pObjects[object].pObjectName != nullptr && strlen(p_callback_data->pObjects[object].pObjectName) > 0) {
```

modules/lightmapper_rd/lightmapper_rd.cpp
```diff
@@ -1017,7 +1021,37 @@ LightmapperRD::BakeError LightmapperRD::bake(BakeQuality p_quality, bool p_use_d
...
+				memdelete(rd);
+				memdelete(rcd);
+				rd = nullptr;
+				rcd = nullptr;
+			}
+		}
+	}
+
+	if (rd == nullptr) {
+		ERR_FAIL_V(BAKE_ERROR_LIGHTMAP_CANT_PRE_BAKE_MESHES);
```
AThousandShips, line 1053: This should be a `ERR_FAIL_NULL_V`

darksylinc:
I've been reviewing it today, the Vulkan & generic parts look alright (I did not review D3D12).

As I already told Dario personally, the only thing that makes me nervous is this:

```
RenderingDeviceGraph::frame
RenderingDevice::frame
```

They're both cloned versions and there's no guarantee(*) they're in sync. But on the other hand, if we'd unify them into just `RenderingDevice`, we still have no guarantee `RenderingDeviceGraph` will notice when `RenderingDevice::frame` advances.

And what's missing would be to aggressively check:
```
DEV_ASSERT( RenderingDevice::frame == RenderingDeviceGraph::frame );
```

Nonetheless, as Dario explained to me and I've confirmed, `RenderingDeviceGraph::frame` isn't currently in use because that's only used for secondary command buffers, which had to be disabled because they were broken on NVIDIA (either because of NV errors, or because of Godot's faults).


(*) By "no guarantee" I mean if someone in the future touches this code and inadvertently introduces subtle bugs. Even if the current code is 100% bug free; checks need to be done to ensure it doesn't get broken tomorrow.

BastiaanOlij:
Haven't had time to dive deep into this but making sure OpenXR still works I can confirm that:
- PCVR works which suggests our OpenXR overrules are still working
- Running on Quest this currently hard crashes but I haven't been able to capture good crash info. Needs to be tested whether this is an Android issue or specific to Android XR devices.

DarioSamo:
It was indeed crashing in Android due to an implementation error on my part. Now I've checked and it seems to be working. Weirdly enough, it also seems to fix a bug where using the controls in the editor would stretch the contents of the viewports to be twice its scale.

Calinou (approved):
Tested locally on Linux + NVIDIA, it works as expected.

However, while this resolves https://github.com/godotengine/godot/issues/87532, every time I bake lightmaps, I now see this message in the Output panel:

```
API 1.3.260 - Forward+ - Using Device #0: NVIDIA - NVIDIA GeForce RTX 4090
```

It's missing "Vulkan" before "API", and I think there should be a way to silence it as it may be excessive for certain local RenderingDevices.

DarioSamo:
> However, while this resolves #87532, every time I bake lightmaps, I now see this message in the Output panel:
> 
> ```
> API 1.3.260 - Forward+ - Using Device #0: NVIDIA - NVIDIA GeForce RTX 4090
> ```
> 
> It's missing "Vulkan" before "API", and I think there should be a way to silence it as it may be excessive for certain local RenderingDevices.

Noted, I'll implement these suggestions.

reduz:
Fantastic work on this!!

DarioSamo:
Opening this up for review, I'll need some help in testing out the other platforms remaining in the TODO list as I don't own the hardware to verify this.

AThousandShips (approved):
Some style details but looks good otherwise for style and macro usage

drivers/d3d12/rendering_context_driver_d3d12.h
```diff
@@ -0,0 +1,120 @@
...
+	SurfaceID surface_create(const void *p_platform_data) override;
+	void surface_set_size(SurfaceID p_surface, uint32_t p_width, uint32_t p_height) override final;
+	void surface_set_vsync_mode(SurfaceID p_surface, DisplayServer::VSyncMode p_vsync_mode) override final;
+	DisplayServer::VSyncMode surface_get_vsync_mode(SurfaceID p_surface) const override final;
+	uint32_t surface_get_width(SurfaceID p_surface) const override final;
+	uint32_t surface_get_height(SurfaceID p_surface) const override final;
+	void surface_set_needs_resize(SurfaceID p_surface, bool p_needs_resize) override final;
+	bool surface_get_needs_resize(SurfaceID p_surface) const override final;
+	void surface_destroy(SurfaceID p_surface) override final;
+	bool is_debug_utils_enabled() const override final;
```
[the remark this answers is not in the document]
DarioSamo: Fair enough, also taking out some of these finals as well. They were mostly copied from the other style of files but it'll be quite problematic in the future for a platform that just needs to inherit and change a few things of these drivers.

doc/classes/ProjectSettings.xml
```diff
@@ -2658,6 +2658,21 @@
 		</member>
 		<member name="rendering/rendering_device/staging_buffer/texture_upload_region_size_px" type="int" setter="" getter="" default="64">
 		</member>
+		<member name="rendering/rendering_device/vsync/frame_queue_size" type="int" setter="" getter="" default="2">
+			The number of frames to track on the CPU side before stalling to wait for the GPU.
+			1 may give you the lowest lag/latency but at the high cost of no parallelism between CPU &amp; GPU.
```
Mickeon, line 2663: Could also start with "A value of", but honestly the subject is fairly clear here.
The ampersand is used extremely rarely in the class reference, too, but I wouldn't split hairs over it, if you feel like it looks better.
```suggestion
			[code]1[/code] may give you the lowest lag/latency but at the high cost of no parallelism between CPU and GPU.
```

doc/classes/ProjectSettings.xml
```diff
@@ -2658,6 +2658,21 @@
...
+			Try the [url=https://darksylinc.github.io/vsync_simulator/]V-Sync Simulator[/url], an interactive interface that simulates presentation to better understand how it is affected by different variables under various conditions.
+			[b]Note:[/b] This property is only read when the project starts. There is currently no way to change this value at run-time.
+		</member>
+		<member name="rendering/rendering_device/vsync/swapchain_image_count" type="int" setter="" getter="" default="3">
+			The number of images the swapchain will consist of (back buffers + front buffer).
+			[code]2[/code] corresponds to double-buffering and [code]3[/code] to triple-buffering. A value of [code]1[/code] is not recommended.
```
Mickeon, line 2669: > A value of [code]1[/code] is not recommended.

Why? If it can be summarised with "It just does not look good" feel free to mention it. If the explanation is too long I'm okay with omitting it.

doc/classes/ProjectSettings.xml
```diff
@@ -2658,6 +2658,21 @@
...
+			Double-buffering may give you the lowest lag/latency but if V-Sync is on and the system can't render at 60 fps, the framerate will go down in multiples of it (e.g. 30 fps, 15, 7.5, etc.). Triple buffering gives you higher framerate (specially if the system can't reach a constant 60 fps) at the cost of up to 1 frame of latency, with [constant DisplayServer.VSYNC_ENABLED] (FIFO).
+			Use double-buffering with [constant DisplayServer.VSYNC_ENABLED]. Triple-buffering is a must if you plan on using [constant DisplayServer.VSYNC_MAILBOX] mode.
+			Try the [url=https://darksylinc.github.io/vsync_simulator/]V-Sync Simulator[/url], an interactive interface that simulates presentation to better understand how it is affected by different variables under various conditions.
+			[b]Note:[/b] This property is only read when the project starts. There is currently no way to change this value at run-time.
+			[b]Note:[/b] Some platforms may restrict the actual value.
```
Mickeon, line 2674: This note is a bit too up to interpretation. It would be lovely to mention a few platforms that do this, by name.

Mickeon (commented):
Some thoughts about the class reference (which I like a lot by the way)

DarioSamo:
@Mickeon Those were taken from @darksylinc's other PR with his permission so I'll have to refer to him to see if he agrees with the changes, as I did not write these originally.

doc/classes/ProjectSettings.xml
darksylinc: The whole block of text has gone through multiple revisions.

The problem is that this setting is related to the following 3 topics:

 - V-Sync
 - Input to Display Latency
 - CPU & GPU synchronization

One could write a 100-page (or more) essay of these 3. There is only so much that can be squashed into a tooltip. They are all advanced topics.

This is why the tooltip refers to the [Simulator](https://darksylinc.github.io/vsync_simulator) which contains an interactive simulator and [links](https://askubuntu.com/a/1420575/274245) to more [bibliography](https://www.gdcvault.com/play/1026327/Controller-to-Display-Latency-in) about it that go in depth.

Regarding this particular question. A value of 1is not recommended, is because the CPU will always have to wait for the GPU, which can hurt performance a lot, but might be important for specific scenarios where latency must be the lowest (real time medical imagining _might_ need this), or RAM is tight and synchronization is not necessary or rare (AI and blockchain mining comes to mind).

However I saw that this PR from @DarioSamo always performs a `actual_count = max( 2u, swapchain_image_count )` unless local device is used. I don't know exactly why he did this, but it may make this comment irrelevant and maybe `swapchain_image_count` should have 2 as min value.

doc/classes/ProjectSettings.xml
darksylinc: Since the platforms and API backends are evolving over time, it's hard to say.

I know for sure these values are ignored in OpenGL (compatibility). But the other backends (D3D12, Metal) are not finished yet. It's also unclear what happens in macOS via MoltenVK.

Mickeon: In that case, if the compatibility renderer is guaranteed to stay like this, that's worth mentioning (it also encompasses other platforms like Web).

HybridEidolon:
As currently written, this seems to partially fix #85547. The shadow flickering issue now only happens when:

- Resizing the window
- Holding Alt for the Alt-Tab widget

Maximizing and minimizing on W11 with the AMD driver work great. Resizing works fine with aformentioned issue.

doc/classes/ProjectSettings.xml
DarioSamo: We consulted with @reduz and we should just have two as the safe minimum, so the note will be pretty irrelevant.

doc/classes/ProjectSettings.xml
DarioSamo: Same reason as the other value, these are clamped to 2 as well at minimum to avoid introducing "footguns" the users could misconfigure.

doc/classes/ProjectSettings.xml
DarioSamo: Rendering Device is not used by the compatibility renderer, which is where this setting lives on in the configuration panel. It doesn't sound like mentioning Compatibility renderer is relevant in that case to me.

The actual values that might be restricted by platforms is a pretty mixed bag to give clear examples and seems to change every now and then. I'd suggest a note that just says the value is a suggestion to the OS but that it might not match what the user actually requests at the low level.

DarioSamo:
Rebased to integrate the latest changes to D3D12.

Still looking for confirmation for Apple devices from anyone who owns them!

HybridEidolon:
@DarioSamo I built it on my M1 Macbook Air (only for x86_64, for some reason arm64 isn't able to find mvk symbols) and it works fine in a basic scene with SDFGI enabled. Resizing, moving, fullscreen, all of that works great. Just shader compilation slowness.

Calinou:
> Edit: Spoke too soon; minimizing to tray does appear to cause issues with re-renders, but I'm not sure if that's new to this branch or something that's always existed on macOS.

This is likely https://github.com/godotengine/godot/issues/87309.

DarioSamo:
> Edit: Spoke too soon; minimizing to tray does appear to cause issues with re-renders, but I'm not sure if that's new to this branch or something that's always existed on macOS.

If you can confirm it's not a regression from `master` then that's probably fine.

BastiaanOlij:
Still having problems running this on Quest:
image

It looks like the injection of OpenXR code to initialise Vulkan is not happening correctly

DarioSamo:
Thanks to @BastiaanOlij we tracked down that it is not necessary to request the transfer bit for the main queue, as apparently it is optional to do so and the Quest does not define it despite requesting the graphics and compute bits. This confirms that OpenXR does work in Quest and so all of the platforms we were looking to verify are working.

BastiaanOlij:
> Thanks to @BastiaanOlij we tracked down that it is not necessary to request the transfer bit for the main queue, as apparently it is optional to do so and the Quest does not define it despite requesting the graphics and compute bits. This confirms that OpenXR does work in Quest and so all of the platforms we were looking to verify are working.

Such a weird problem, but glad we figured it out. This all works great imho

servers/rendering/rendering_device_graph.cpp
```diff
@@ -40,15 +40,7 @@ RenderingDeviceGraph::RenderingDeviceGraph() {
...
-	_wait_for_secondary_command_buffer_tasks();
-
-	for (Frame &f : frames) {
-		for (SecondaryCommandBuffer &secondary : f.secondary_command_buffers) {
-			if (secondary.command_pool.id != 0) {
-				driver->command_pool_free(secondary.command_pool);
-			}
-		}
-	}
+	// Default destruction.
```
RandomShaper, line 43:
```suggestion
```

drivers/d3d12/rendering_context_driver_d3d12.cpp
```diff
@@ -0,0 +1,314 @@
...
+#endif
+#endif
+
+RenderingContextDriverD3D12::RenderingContextDriverD3D12() {
+	CharString cs = Engine::get_singleton()->get_architecture_name().ascii();
+	memcpy(godot_nir_arch_name, (const char *)cs.get_data(), cs.size());
+}
+
+RenderingContextDriverD3D12::~RenderingContextDriverD3D12() {
+	// Does nothing.
```
RandomShaper, line 86:
```suggestion
```

drivers/d3d12/rendering_context_driver_d3d12.cpp
```diff
@@ -0,0 +1,314 @@
...
+		}
+		sdk_config->Release();
+	}
+
+	return OK;
+}
+
+Error RenderingContextDriverD3D12::_initialize_debug_layers() {
+	ComPtr<ID3D12Debug> debug_controller;
+	HRESULT res;
```
RandomShaper, line 119:
```suggestion
	HRESULT res = OK;
```

drivers/d3d12/rendering_context_driver_d3d12.h
```diff
@@ -0,0 +1,120 @@
...
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic pop
+#endif
+
+using Microsoft::WRL::ComPtr;
+
+#define ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
+
+class RenderingContextDriverD3D12 : public RenderingContextDriver {
+private:
```
RandomShaper, line 69:
```suggestion
```

drivers/d3d12/rendering_context_driver_d3d12.h
```diff
@@ -0,0 +1,120 @@
...
+	ComPtr<IDXGIFactory2> dxgi_factory;
+	TightLocalVector<Device> driver_devices;
+	bool tearing_supported = false;
+
+	Error _init_device_factory();
+	Error _initialize_debug_layers();
+	Error _initialize_devices();
+
+public:
+	RenderingContextDriverD3D12();
```
RandomShaper, line 80: I'm too late for `RenderingDeviceGraph`, but Godot codebase has a standard of putting constructors and destructors as the last members of classes. (This also applies to other offending cases in this PR.)

drivers/d3d12/rendering_context_driver_d3d12.h
```diff
@@ -0,0 +1,120 @@
...
+	virtual uint32_t surface_get_width(SurfaceID p_surface) const override;
+	virtual uint32_t surface_get_height(SurfaceID p_surface) const override;
+	virtual void surface_set_needs_resize(SurfaceID p_surface, bool p_needs_resize) override;
+	virtual bool surface_get_needs_resize(SurfaceID p_surface) const override;
+	virtual void surface_destroy(SurfaceID p_surface) override;
+	virtual bool is_debug_utils_enabled() const override;
+
+	// Platform-specific data for the Windows embedded in this driver.
+	struct WindowPlatformData {
+		HWND window;
```
RandomShaper, line 101:
```suggestion
	HWND window = nullpr;
```

(Well, or `NULL`, as you did below, which is more Win32-ish, although in Godot `nullptr` is generally favored even when interacting with C APIs.)

drivers/d3d12/rendering_device_driver_d3d12.h
```diff
@@ -64,15 +64,13 @@ using Microsoft::WRL::ComPtr;
 
 #define D3D12_BITCODE_OFFSETS_NUM_STAGES 3
 
+// Forward declarations.
```
RandomShaper, line 67:
```suggestion
```

drivers/d3d12/rendering_device_driver_d3d12.h
```diff
@@ -86,8 +84,58 @@ class RenderingDeviceDriverD3D12 : public RenderingDeviceDriver {
...
+		uint64_t max_srvs_per_shader_stage = 0;
+		uint64_t max_cbvs_per_shader_stage = 0;
+		uint64_t max_samplers_across_all_stages = 0;
+		uint64_t max_uavs_across_all_stages = 0;
+		uint64_t timestamp_frequency = 0;
+	};
+
+	struct SubgroupCapabilities {
+		uint32_t size;
+		bool wave_ops_supported;
```
RandomShaper, line 97:
```suggestion
		uint32_t size = 0;
		bool wave_ops_supported = false;
```

drivers/d3d12/rendering_device_driver_d3d12.h
```diff
@@ -86,8 +84,58 @@ class RenderingDeviceDriverD3D12 : public RenderingDeviceDriver {
...
+		uint32_t supported_operations_flags_rd() const;
+	};
+
+	struct VRSCapabilities {
+		bool draw_call_supported; // We can specify our fragment rate on a draw call level.
+		bool primitive_supported; // We can specify our fragment rate on each drawcall.
+		bool primitive_in_multiviewport;
+		bool ss_image_supported; // We can provide a density map attachment on our framebuffer.
+		uint32_t ss_image_tile_size;
+		bool additional_rates_supported;
```
RandomShaper, line 108:
```suggestion
		bool draw_call_supported = false; // We can specify our fragment rate on a draw call level.
		bool primitive_supported = false; // We can specify our fragment rate on each drawcall.
		bool primitive_in_multiviewport = false;
		bool ss_image_supported = false; // We can provide a density map attachment on our framebuffer.
		uint32_t ss_image_tile_size = 0;
		bool additional_rates_supported = false;
```

drivers/d3d12/rendering_device_driver_d3d12.h
```diff
@@ -86,8 +84,58 @@ class RenderingDeviceDriverD3D12 : public RenderingDeviceDriver {
...
+	};
+
+	struct ShaderCapabilities {
+		D3D_SHADER_MODEL shader_model;
+		bool native_16bit_ops;
```
RandomShaper, line 113:
```suggestion
		D3D_SHADER_MODEL shader_model = (D3D_SHADER_MODEL)0;
		bool native_16bit_ops = false;
```

drivers/d3d12/rendering_device_driver_d3d12.h
```diff
@@ -86,8 +84,58 @@ class RenderingDeviceDriverD3D12 : public RenderingDeviceDriver {
...
+	};
+
+	struct StorageBufferCapabilities {
+		bool storage_buffer_16_bit_access_is_supported;
```
RandomShaper, line 117:
```suggestion
		bool storage_buffer_16_bit_access_is_supported = false;
```

drivers/d3d12/rendering_device_driver_d3d12.h
```diff
@@ -86,8 +84,58 @@ class RenderingDeviceDriverD3D12 : public RenderingDeviceDriver {
...
+	};
+
+	struct FormatCapabilities {
+		bool relaxed_casting_supported;
```
RandomShaper, line 121:
```suggestion
		bool relaxed_casting_supported = false;
```

servers/rendering/rendering_context_driver.cpp
```diff
@@ -0,0 +1,86 @@
...
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "rendering_context_driver.h"
+
+RenderingContextDriver::~RenderingContextDriver() {
+	// Does nothing.
```
RandomShaper, line 34:
```suggestion
```

RandomShaper (changes requested):
This is in such a good level of quality that I haven't been able but to insult it with nitpickery. 😀

drivers/d3d12/rendering_context_driver_d3d12.h
DarioSamo: I'd stick with NULL if using Win32.

drivers/d3d12/rendering_context_driver_d3d12.cpp
DarioSamo: What's the point of the default initialization if it's pretty much guaranteed it'll have a value from either branch of the code? The sanitization can easily detect something went wrong if the code ahead is affected, while a default initialization might hide the issue. That's pretty much the reasoning in case you were wondering.

doc/classes/RenderingDevice.xml
```diff
@@ -590,6 +590,7 @@
 		</method>
 		<method name="screen_get_framebuffer_format" qualifiers="const">
 			<return type="int" />
+			<param index="0" name="screen" type="int" default="0" />
 			<description>
 				Returns the screen's framebuffer format.
```
Mickeon, line 595: Since you added a screen parameter it may be worth being more explicit.
```suggestion
				Returns the framebuffer format of the given screen.
```

DarioSamo: Agreed.

DarioSamo:
Seems like I'll have to re-do my changes for Wayland, testing would be appreciated when that is done as well.

drivers/d3d12/rendering_context_driver_d3d12.h
DarioSamo: Done.

drivers/d3d12/rendering_context_driver_d3d12.h
DarioSamo: I just remembered we can't do this. This is part of a union, we can't have default initialization on the struct.

RandomShaper: Oh, indeed.

DarioSamo:
> Oof some recent merges just conflicted this PR. I'm still testing this, but this already needs a rebase xD

You mean like in the middle of the hour I was just working on this? 😱 

Let me re-do it.

DarioSamo:
@Riteo Alright, conflicts should be fixed now.

DarioSamo:
> I tried this branch briefly with the Wayland backend on my crappy laptop and I couldn't notice any particular artifacts on the [Garage 42 demo](https://godotengine.org/asset-library/asset/2379) (I'll test properly on my desktop later).
> 
> There's a segfault on exit though:

I think that's a good enough confirmation for me TBH, the segfault sounds unrelated to the platform and something I might not just be cleaning up properly, so I'll review it. Thanks for the test!

drivers/d3d12/rendering_context_driver_d3d12.cpp
DarioSamo: Should we mark this as resolved?

DarioSamo:
> There's a segfault on exit though:
> 
> ```
> Thread 1 "godot.linuxbsd." received signal SIGILL, Illegal instruction.
> 0x000055555f098bae in LocalVector<RenderingDevice::Frame, unsigned int, false, true>::operator[] (this=0x7fffef3aac68, p_index=1) at ./core/templates/local_vector.h:161
> 161			CRASH_BAD_UNSIGNED_INDEX(p_index, count);
> (gdb) bt
> #0  0x000055555f098bae in LocalVector<RenderingDevice::Frame, unsigned int, false, true>::operator[](unsigned int) (this=0x7fffef3aac68, p_index=1) at ./core/templates/local_vector.h:161
> #1  0x000055555efb3265 in RenderingDevice::_end_frame() (this=0x7fffef3aa020) at ./servers/rendering/rendering_device.cpp:4887
> #2  0x000055555efb3a6c in RenderingDevice::_flush_and_stall_for_all_frames() (this=0x7fffef3aa020) at ./servers/rendering/rendering_device.cpp:4919
> #3  0x000055555efb70e7 in RenderingDevice::finalize() (this=0x7fffef3aa020) at ./servers/rendering/rendering_device.cpp:5323
> #4  0x000055555efd5746 in RenderingDevice::~RenderingDevice() (this=0x7fffef3aa020, __in_chrg=<optimized out>) at ./servers/rendering/rendering_device.cpp:6123
> #5  0x000055555afc4b34 in memdelete<RenderingDevice>(RenderingDevice*) (p_class=0x7fffef3aa020) at ./core/os/memory.h:109
> #6  0x000055555aff72ee in DisplayServerWayland::~DisplayServerWayland() (this=0x7ffff716b9a0, __in_chrg=<optimized out>) at platform/linuxbsd/wayland/display_server_wayland.cpp:1386
> #7  0x000055555aff923c in memdelete<DisplayServer>(DisplayServer*) (p_class=0x7ffff716b9a0) at ./core/os/memory.h:109
> #8  0x000055555b0257bc in finalize_display() () at main/main.cpp:325
> #9  0x000055555b04232f in Main::cleanup(bool) (p_force=false) at main/main.cpp:4056
> #10 0x000055555af802da in main(int, char**) (argc=3, argv=0x7fffffffe2e8) at platform/linuxbsd/godot_linuxbsd.cpp:76
> ```

Turns out this is actually related to the platform, but only because I added a call to "finalize()" by mistake twice to RenderingDevice when no other platform was doing it. My bad. Either way I've fixed RD to be resilient against finalizing twice and also made it so it's not called twice on Wayland.

clayjohn (approved):
I've reviewed the code and tested locally. This PR has also seen extensive revision and testing from others. 

I am satisfied at this point that it is ready to merge

It just needs a quick rebase

DarioSamo:
Rebased to fixed conflicts.

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -1858,22 +2090,219 @@ void RenderingDeviceDriverD3D12::command_buffer_end(CommandBufferID p_cmd_buffer
...
+		swap_chain->render_targets.push_back(render_target);
+
+		// Create texture information for the framebuffer to reference the resource. Since the states pointer must
+		// reference an address of the element itself, we must insert it first and then modify it.
+		swap_chain->render_targets_info.push_back(TextureInfo());
+		TextureInfo &texture_info = swap_chain->render_targets_info[i];
+		texture_info.owner_info.states.subresource_states.push_back(D3D12_RESOURCE_STATE_PRESENT);
+		texture_info.states_ptr = &texture_info.owner_info.states;
+		texture_info.format = swap_chain->data_format;
+		texture_info.desc = CD3DX12_RESOURCE_DESC(render_target->GetDesc());
```
bruvzg, line 2256: This breaks MinGW build.

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -5224,17 +5652,9 @@ void RenderingDeviceDriverD3D12::begin_segment(CommandBufferID p_cmd_buffer, uin
 	frames[frame_idx].desc_heap_walkers.aux.rewind();
 	frames[frame_idx].desc_heap_walkers.rtv.rewind();
 	frames[frame_idx].desc_heaps_exhausted_reported = {};
-	frames[frame_idx].null_rtv_handle = { 0 };
+	frames[frame_idx].null_rtv_handle = {};
```
bruvzg, line 5655: Also, breaks MinGW build.

drivers/d3d12/rendering_device_driver_d3d12.cpp
DarioSamo: I checked your other PR, how strange that the headers literally define it differently. Now it's a bit late but maybe we have to make a wrapper around that function that works for both compilers without putting the compiler paths inside the caller methods.

bruvzg:
Seems like some change in this PR broke multi-threaded renderer (both compatibility and Vulkan) on macOS (see https://github.com/godotengine/godot/pull/90268#discussion_r1554267653), but I'm not what's exactly wrong.

Alex2782:
**RenderPass2KHR** - `Compatibility fallback` does not work properly on Android LG Nexus 5X, Adreno 418

Also same crash on MacOS without:
``` cpp
_register_requested_device_extension(VK_KHR_CREATE_RENDERPASS_2_EXTENSION_NAME, false);
```



--------------------------

https://github.com/godotengine/godot/blob/780e1a50408360cf0cf93c0b55b59e9d7b4ad0b1/drivers/vulkan/rendering_device_driver_vulkan.cpp#L979

### **crash at:**
https://github.com/godotengine/godot/blob/780e1a50408360cf0cf93c0b55b59e9d7b4ad0b1/drivers/vulkan/rendering_device_driver_vulkan.cpp#L1011

https://github.com/godotengine/godot/blob/780e1a50408360cf0cf93c0b55b59e9d7b4ad0b1/drivers/vulkan/rendering_device_driver_vulkan.cpp#L970

-------------------------------
If I **deactivate the extension** (`VK_KHR_CREATE_RENDERPASS_2_EXTENSION_NAME`), I can also reproduce the crash on **MacMini M1.**

https://github.com/godotengine/godot/blob/780e1a50408360cf0cf93c0b55b59e9d7b4ad0b1/drivers/vulkan/rendering_device_driver_vulkan.cpp#L435
```cpp
Error RenderingDeviceDriverVulkan::_initialize_device_extensions() {
	enabled_device_extension_names.clear();

	_register_requested_device_extension(VK_KHR_SWAPCHAIN_EXTENSION_NAME, true);
	_register_requested_device_extension(VK_KHR_MULTIVIEW_EXTENSION_NAME, false);
	_register_requested_device_extension(VK_KHR_FRAGMENT_SHADING_RATE_EXTENSION_NAME, false);
	//_register_requested_device_extension(VK_KHR_CREATE_RENDERPASS_2_EXTENSION_NAME, false);
	_register_requested_device_extension(VK_KHR_SHADER_FLOAT16_INT8_EXTENSION_NAME, false);
```








Bildschirmfoto 2024-04-25 um 17 53 05

DarioSamo:
@Alex2782 Can you confirm if https://github.com/godotengine/godot/pull/91169 fixes it? Thank you very much for the report by the way.

PR #78092, opened 2023-06-10
Preserve selection when focusing SpinBox

KoBeWi:
Fixes #78079

PR #87341, opened 2024-01-18
[3.x] Preserve selection when focusing SpinBox

KoBeWi:
Backport of #78092

I copied the code and replaced function calls with their equivalents, but I didn't test it.


akien-mga (approved):

akien-mga:
Thanks!

PR #87344, opened 2024-01-18
[Core] Add way to check if a signal has any connections

### LINKED ISSUE #8912: Add 'is_connected()' function without parameters to check if signal is connected

2D MMORPG

When creating signals, it is difficult to follow all of the possible combinations of signals I used in my game.

So I want to force check that my signal is connected first, and if not, the game crashes.

Is there a way to force signals to be connected beforehand? I think its a good idea. For example, if you have a system with more than 100+ signals, how can you follow each one?
What happens when a single connection is broken? How would you trouble shoot it?

I think we should also add 'final' keyword or something to signals in GDScript, so that the game engine checks that all the connections are fine.

Its quality of life feature, not really neccesery

You could check with 

```
get_connections().size() > 0
```

But we should instead have more inline function:

```
is_connection()
```

It can also be indicated with
```
has_connection()
```

Simple check in the core of the engine

Yes

As I said before its core of signals

AThousandShips:
Added to `Object` and `Signal`

~Went with `any_connected` because it felt clear and direct as a description, could use `has_connections` too if desired~

Will see about adding some unit tests for this but we don't have unit tests for `is_connected` so unsure how to approach that

Unsure if I need to do anything special on the C# side, looked at the `Signal` class and didn't see anything to add

* Closes: https://github.com/godotengine/godot-proposals/issues/8912

### DIFF
--- a/core/object/object.cpp
+++ b/core/object/object.cpp
@@ -1457,6 +1457,24 @@ bool Object::is_connected(const StringName &p_signal, const Callable &p_callable
 	return s->slot_map.has(*p_callable.get_base_comparator());
 }
 
+bool Object::has_connections(const StringName &p_signal) const {
+	const SignalData *s = signal_map.getptr(p_signal);
+	if (!s) {
+		bool signal_is_valid = ClassDB::has_signal(get_class_name(), p_signal);
+		if (signal_is_valid) {
+			return false;
+		}
+
+		if (!script.is_null() && Ref<Script>(script)->has_script_signal(p_signal)) {
+			return false;
+		}
+
+		ERR_FAIL_V_MSG(false, "Nonexistent signal: " + p_signal + ".");
+	}
+
+	return !s->slot_map.is_empty();
+}
+
 void Object::disconnect(const StringName &p_signal, const Callable &p_callable) {
 	_disconnect(p_signal, p_callable);
 }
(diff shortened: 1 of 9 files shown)


Mickeon:
Unsure about `any_connected` because it is more... "distorted" when reading through it in your own codebase, compared to `has_connections`. To me it does look fine as a Signal method (e.g. `$Button.pressed.any_connected()`), but not so much for Object (`$Button.any_connected("pressed")`)

AThousandShips:
I'd prefer the name to be uniform on both, but I think there's no difference in reading `is_connected` on `Object`

aaronfranke:
I think `has_connections` is clearer than `any_connected`. The latter seems like it would accept an array of Callables or a vararg of Callables.

AThousandShips:
I'd say that both could be seen that way, but I'll change the naming from the consensus, I've got no strong feelings either way

AThousandShips:
Might go with `is_any_connected` instead later, but we'll see

ShlomiRex:
Suggestion:

Change current function `is_connected(name)` to `is_connected_to(name)`

Now you can also give the function `any_connected()` change to `is_connected()`

for more elegant naming

Mickeon:
No changing existing function names, it would break compatibility.

Mickeon (approved):
Documentation is completely fine.

akien-mga (approved):

akien-mga:
Thanks!

AThousandShips:
Thank *you*!

PR #87349, opened 2024-01-18
Fix memory leak from not clearing the buffer barrier vector properly on the render graph.

DarioSamo:
Fixes a small memory leak that #84976 introduced from the last minute change to buffer barriers, which were not being cleared properly every frame leading to a small but consistent leak every frame.


akien-mga (approved):

clayjohn (approved):

akien-mga:
Thanks!

PR #87350, opened 2024-01-18
Add support for material-based debanding in the Compatibility renderer

### LINKED ISSUE #4193: Add material-based debanding as an alternative (or replacement) to screen-space debanding

The Godot editor :slightly_smiling_face: 

Godot currently offers screen-space debanding in the GLES3 (`3.x`) and Vulkan Clustered (`master`) backends. Debanding currently doesn't work in GLES2 because a LDR rendering pipeline is used there. Vulkan Mobile backend uses an HDR rendering pipeline but with lower precision than Vulkan Clustered (with [debanding also not working there](https://github.com/godotengine/godot/issues/55879)).

Add a debanding filter that works at a *material* level instead of doing so at screen-space level.

Performance seems to be roughly the same from my testing, although it may be slower in complex scenes when the depth prepass is disabled (as is the case in GLES2 and Vulkan Mobile).
On the bright side, material-level debanding will not operate on the sky, saving GPU resources in outdoor scenes. (Instead, the sky shader itself should have debanding, like PhysicalSkyMaterial already does.)

Depending on how well this material-level filter behave, we may consider removing the screen-space approach for 4.0. If there are still valid reasons to keep it, then material-level debanding should be provided as an additional option (ideally in all rendering backends for consistency, even GLES3 and Vulkan Clustered).

I have WIP implementations for the Vulkan Mobile backend and GLES2:

- `master`: https://github.com/Calinou/godot/tree/vulkan-mobile-add-material-debanding
- `3.x`: https://github.com/Calinou/godot/tree/add-material-debanding-3.x

## GLES2

*Click to view at full size.*

| Without debanding | With debanding |
|-|-|
|  |  |

## Vulkan Mobile

| Without debanding | With debanding |
|-|-|
| exposure_1_without_debanding | exposure_1_with_debanding |

This can be worked around with a custom shader already: https://github.com/fractilegames/godot-gles2-debanding-material

Since games with a lowpoly/stylized aesthetic are often developed with Godot (and I don't see that changing soon), I think it makes sense to provide this as a built-in option. In any case, debanding would still be disabled by default to avoid making screenshots larger when it's not needed.

### LINKED ISSUE #7336: Dither colors in any intermediate write to color framebuffer when using the mobile renderer

Godot

Users often complain about the excessive banding in the mobile renderer. The banding is a result of using an RGB10_A2 framebuffer. Uses such limited bit depth is necessary to achieve full frame rates on mobile devices, however, it can lead to unavoidable banding. 

Using a post-process de-banding solution isn't enough as the banding is ingrained during the main scene pass

The solution is to deband during every intermediate write to the color buffer. I.e. during the scene shader and sky shader pass. 

This can be controlled with a project setting so that users do not have to take on the cost of debanding if they do not want to

Add a line to deband at the end of the scene shader similar to what we have now in the sky shader
https://github.com/godotengine/godot/blob/0c2144da908a8223e188d27ed1d31d8248056c78/servers/rendering/renderer_rd/shaders/environment/sky.glsl#L286

In the sky shader always enable debanding

Comparison:
_current master_
https://user-images.githubusercontent.com/16521339/235844386-8e3c7d04-6a53-478a-bb39-c810fe216ba4.png

_With debanding_
https://user-images.githubusercontent.com/16521339/235844406-f6e84c47-11d6-4fe6-a422-12a53e0a25ce.png

This can't be worked around as debanding has to happen after lighting

It is necessary for all 3D games using the mobile renderer

Calinou:
This works in a fashion similar to the Mobile renderer, with some adjustments to better account for the additive lighting approach used for lights with shadows enabled.

- This closes https://github.com/godotengine/godot-proposals/issues/4193
and closes https://github.com/godotengine/godot-proposals/issues/7336.

**Testing project:** [test_material_debanding.zip](https://github.com/godotengine/godot/files/13980617/test_material_debanding.zip)

## Preview

*All screenshots use the Compatibility renderer.*

### Real-time lights

No debanding | With debanding
-|-
No debanding | With debanding

### Lightmaps

No debanding | With debanding
-|-
No debanding | With debanding

<details>
<summary>Old version</summary>

*Click to view at full size (images **must** be compared at their original size).*

### Forward+

No debanding | Viewport-based debanding | Material-based debanding | Both debanding
-|-|-|-
Screenshot_20240118_191618_forward_plus_no_debanding webp | Screenshot_20240118_191651_forward_plus_viewport_debanding webp | Screenshot_20240118_191730_forward_plus_material_debanding webp | Screenshot_20240118_191739_forward_plus_viewport_material_debanding webp

### Mobile

No debanding | Viewport-based debanding | Material-based debanding | Both debanding
-|-|-|-
Screenshot_20240118_192121_mobile_no_debanding webp | Screenshot_20240118_192136_mobile_viewport_debanding webp | Screenshot_20240118_192137_mobile_material_debanding webp | Screenshot_20240118_192137_mobile_viewport_material_debanding webp

<details>
<summary>With debanding applied after the division by `sc_luminance_multiplier` (looks too strong)</summary>

No debanding | Viewport-based debanding | Material-based debanding | Both debanding
-|-|-|-
Screenshot_20240118_192121_mobile_no_debanding webp | Screenshot_20240118_192136_mobile_viewport_debanding webp | Screenshot_20240118_192649_mobile_lum_material_debanding webp | Screenshot_20240118_192601_mobile_lum_viewport_material_debanding webp | 
</details>

### Compatibility

No debanding | Viewport-based debanding | Material-based debanding | Both debanding
-|-|-|-
Screenshot_20240118_192346_compatibility_no_debanding webp | Not available | Screenshot_20240118_192338_compatibility_material_debanding webp | Not available
</details>

### DIFF
--- a/doc/classes/RenderingServer.xml
+++ b/doc/classes/RenderingServer.xml
@@ -2391,8 +2391,8 @@
 			<return type="void" />
 			<param index="0" name="enable" type="bool" />
 			<description>
-				When using the Mobile renderer, [method material_set_use_debanding] can be used to enable or disable the debanding feature of 3D materials ([BaseMaterial3D] and [ShaderMaterial]).
-				[method material_set_use_debanding] has no effect when using the Compatibility or Forward+ renderer. In Forward+, [Viewport] debanding can be used instead.
+				When using the Mobile and Compatibility renderers, [method material_set_use_debanding] can be used to enable or disable the debanding feature of 3D materials ([BaseMaterial3D] and [ShaderMaterial]).
+				[method material_set_use_debanding] has no effect when using the Forward+ renderer. In Forward+, [Viewport] debanding can be used instead.
 				See also [member ProjectSettings.rendering/anti_aliasing/quality/use_debanding] and [method RenderingServer.viewport_set_use_debanding].
 			</description>
 		</method>
--- a/drivers/gles3/rasterizer_scene_gles3.cpp
+++ b/drivers/gles3/rasterizer_scene_gles3.cpp
@@ -3480,6 +3480,10 @@ void RasterizerSceneGLES3::_render_list_template(RenderListParameters *p_params,
 					}
 				}
 
+				if (material_use_debanding) {
+					spec_constants |= SceneShaderGLES3::USE_MATERIAL_DEBANDING;
+				}
+
 				if (uses_additive_lighting) {
 					spec_constants |= SceneShaderGLES3::USE_ADDITIVE_LIGHTING;
 
@@ -3525,6 +3529,11 @@ void RasterizerSceneGLES3::_render_list_template(RenderListParameters *p_params,
 						} else if (scene_state.directional_shadow_quality >= RS::SHADOW_QUALITY_SOFT_LOW) {
 							spec_constants |= SceneShaderGLES3::SHADOW_MODE_PCF_5;
 						}
+					}
+
+					if (material_use_debanding) {
+						// Additive lighting also needs to use debanding.
+						spec_constants |= SceneShaderGLES3::USE_MATERIAL_DEBANDING;
 					}
 				}
 			}
@@ -4366,7 +4375,7 @@ void RasterizerSceneGLES3::lightmaps_set_bicubic_filter(bool p_enable) {
 }
 
 void RasterizerSceneGLES3::material_set_use_debanding(bool p_enable) {
-	// Material debanding not yet implemented.
+	material_use_debanding = p_enable;
 }
 
 RasterizerSceneGLES3::RasterizerSceneGLES3() {
@@ -4383,6 +4392,7 @@ RasterizerSceneGLES3::RasterizerSceneGLES3() {
 	positional_soft_shadow_filter_set_quality((RS::ShadowQuality)(int)GLOBAL_GET("rendering/lights_and_shadows/positional_shadow/soft_shadow_filter_quality"));
 	directional_soft_shadow_filter_set_quality((RS::ShadowQuality)(int)GLOBAL_GET("rendering/lights_and_shadows/directional_shadow/soft_shadow_filter_quality"));
 	lightmaps_set_bicubic_filter(GLOBAL_GET("rendering/lightmapping/lightmap_gi/use_bicubic_filter"));
+	material_set_use_debanding(GLOBAL_GET("rendering/anti_aliasing/quality/use_debanding"));
 
 	{
 		// Setup Lights
--- a/drivers/gles3/rasterizer_scene_gles3.h
+++ b/drivers/gles3/rasterizer_scene_gles3.h
@@ -765,6 +765,7 @@ protected:
 	bool glow_bicubic_upscale = false;
 
 	bool lightmap_bicubic_upscale = false;
+	bool material_use_debanding = false;
 
 	/* Sky */
 
--- a/drivers/gles3/shaders/scene.glsl
+++ b/drivers/gles3/shaders/scene.glsl
@@ -39,6 +39,7 @@ RENDER_MATERIAL = false
 SECOND_REFLECTION_PROBE = false
 LIGHTMAP_BICUBIC_FILTER = false
 RENDER_MOTION_VECTORS = false
+USE_MATERIAL_DEBANDING = false
 
 
 #[vertex]
@@ -1997,6 +1998,20 @@ vec4 textureArray_bicubic(sampler2DArray tex, vec3 uv, vec2 texture_size) {
 #endif //LIGHTMAP_BICUBIC_FILTER
 #endif // RENDER_MOTION_VECTORS
 
+#ifdef USE_MATERIAL_DEBANDING
+// From https://alex.vlachos.com/graphics/Alex_Vlachos_Advanced_VR_Rendering_GDC2015.pdf
+// and https://www.shadertoy.com/view/MslGR8 (5th one starting from the bottom)
+// NOTE: `frag_coord` is in pixels (i.e. not normalized UV).
+vec3 screen_space_dither(vec2 frag_coord) {
+	// Iestyn's RGB dither (7 asm instructions) from Portal 2 X360, slightly modified for VR.
+	vec3 dither = vec3(dot(vec2(171.0, 231.0), frag_coord));
+	dither.rgb = fract(dither.rgb / vec3(103.0, 71.0, 97.0));
+
+	// Subtract 0.5 to avoid slightly brightening the whole viewport.
+	return (dither.rgb - 0.5) / 255.0;
+}
+#endif // USE_MATERIAL_DEBANDING
+
 void main() {
 #ifndef RENDER_MOTION_VECTORS
 	//lay out everything, whatever is unused is optimized away anyway
@@ -2560,6 +2575,11 @@ void main() {
 
 	// Tonemap before writing as we are writing to an sRGB framebuffer
 	frag_color.rgb *= exposure;
+
+#ifdef USE_MATERIAL_DEBANDING
+	frag_color.rgb += screen_space_dither(gl_FragCoord.xy);
+#endif // USE_MATERIAL_DEBANDING
+
 #ifdef APPLY_TONEMAPPING
 	frag_color.rgb = apply_tonemapping(frag_color.rgb);
 #endif
@@ -2829,6 +2849,19 @@ void main() {
 
 	additive_light_color *= (1.0 - fog.a);
 #endif // !FOG_DISABLED
+
+#ifdef USE_MATERIAL_DEBANDING
+	// Apply a per-light offset to avoid making the dithering pattern too noticeable
+	// when using multiple lights with additive blending.
+#if defined(ADDITIVE_OMNI)
+	vec2 offset = vec2(float(omni_light_index) * 5.0, float(omni_light_index) * 2.0);
+#elif defined(ADDITIVE_SPOT)
+	vec2 offset = vec2(float(spot_light_index) * 5.0, float(spot_light_index) * 2.0);
+#else // Directional
+	vec2 offset = vec2(0.0);
+#endif
+	frag_color.rgb += screen_space_dither(gl_FragCoord.xy + offset);
+#endif // USE_MATERIAL_DEBANDING
 
 	// Tonemap before writing as we are writing to an sRGB framebuffer
 	additive_light_color *= exposure;


BastiaanOlij (approved):
Code wise this looks good. From the screenshots it does seem to look like material based debanding works way better than the viewport debanding.

It would be good to get some ideas about the impact on performance, and it would be good to have a recording (though compression likely will make it impossible to judge whether the debanding has a good effect).

Actually, that brings an interesting topic to mind, how much will debanding effect game streaming..

Calinou:
> Actually, that brings an interesting topic to mind, how much will debanding effect game streaming..

Lossy compression will generally destroy any form of dithering pattern (especially subtle ones), so streaming shouldn't be noticeably affected by debanding.

clayjohn (commented):
The implementation looks fine to me. 

What worries me about this PR is that it requires users to toggle a setting on every single material/shader in their game. Its easy for small games to have hundreds of materials and for larger games to have many thousand. 

Is there a use-case for toggling debanding on a single material but not others? 

IMO this should probably be a project setting that enables a specialization constant unless we have a reason to only selectively enable it on certain meshes.

Calinou:
> Is there a use-case for toggling debanding on a single material but not others?

Yes, flat-color materials (or materials with textures that are used with "UV keying" techniques) typically need it more than materials with detailed textures.

clayjohn:
> Yes, flat-color materials typically need it more than materials with detailed textures. However, since textures are sometimes used for materials that _look_ flat-color (but are still using a texture), we can't enable debanding only on untextured materials.

Would there be harm in having debanding enabled on textured materials even when not _needed_? Is your concern for the performance overhead?

I'm trying to figure out if there is a use-case for a user to go and enable this project setting on materials individually. From what I can tell it makes more sense as an all-or-nothing setting where it applies to all materials or none.

Calinou:
> Is your concern for the performance overhead?

Yes, it may be non-negligible if you have overdraw in your scene when using the mobile renderer. If this isn't important for your target hardware, the project setting to force debanding on all materials would address this.

Calinou:
I've looked into converting this into a project setting and have pushed an additional commit. However, it doesn't appear to work and am not sure how to get it working in Compatibility as well. I see the GLES3 scene shader has a `spec_constant` emulation still.

However, should we be using a different shader version or specialization constant here? It's a setting that won't be toggled during gameplay after all, so I'm not sure if we should use up space in the specialization constant map for this (in Forward+/Mobile, that is).

clayjohn:
> However, should we be using a different shader version or specialization constant here? It's a setting that won't be toggled during gameplay after all, so I'm not sure if we should use up space in the specialization constant map for this (in Forward+/Mobile, that is).

In the RD renderers we have a set of spec constants that don't change frequently. I would use spec constants and then group it with those. 

https://github.com/godotengine/godot/blob/17e7f85c06366b427e5068c5b3e2940e27ff5f1d/servers/rendering/renderer_rd/forward_clustered/render_forward_clustered.cpp#L4083-L4084

The reason this is preferable is so that users can change the setting for all materials without restarting the editor. 

https://github.com/godotengine/godot/blob/17e7f85c06366b427e5068c5b3e2940e27ff5f1d/servers/rendering/renderer_rd/forward_clustered/scene_shader_forward_clustered.cpp#L834

If you were to make a shader version then you wouldn't have an easy way to invalidate all shaders and update (as you only set the version at shader compile time).

roalyr:
Here is a stress test. Please consider changing to `dither.rgb - 0.8` for better results, because otherwise when using `blend_add` materials the dithering patter starts to stand out on overlapping meshes.

Original scene, no dither:
dither_disabled

Different subtraction values (I suggest `dither.rgb - 0.8`):
Dithering

Currently, with `dither.rgb - 0.5` the pattern is visible even on absolutely black overlapping materials:
image

clayjohn:
> Here is a stress test. Please consider changing to `dither.rgb - 0.8` for better results, because otherwise when using `blend_add` materials the dithering patter starts to stand out on overlapping meshes.

We should probably disable the dither altogether for non-opaque surfaces. Shifting the dither by 0-8 will introduce stronger dithering artifacts for brighter scenes

roalyr:
> > Here is a stress test. Please consider changing to `dither.rgb - 0.8` for better results, because otherwise when using `blend_add` materials the dithering patter starts to stand out on overlapping meshes.
> 
> We should probably disable the dither altogether for non-opaque surfaces. Shifting the dither by 0-8 will introduce stronger dithering artifacts for brighter scenes

Is that the case for `blend_add`, when `ALPHA` is not involved? What kind of solution the above scene would benefit from if dithering is disabled?
And also, can this sort of dithering be done in .gdshader on demand then?

clayjohn:
> Is that the case for `blend_add`, when `ALPHA` is not involved? What kind of solution the above scene would benefit from if dithering is disabled? ~And also, can this sort of dithering be done in .gdshader on demand then?~ Edit: never mind, I can just implement it in .gdshader then.

Anything with `blend_add` is non-opaque. So yes, it would apply even when ALPHA is not involved. The problem is the dithering takes place in screen space. So if you add a material multiple times, then the dither is applied the same way to the same pixel and makes it much stronger. 

The other alternative would be to rotate the dithering samples per material to ensure that they don't line up. That might be possible, but it would make the approach much more complex.

Calinou:
> The other alternative would be to rotate the dithering samples per material to ensure that they don't line up. That might be possible, but it would make the approach much more complex.

This is something I toyed with for distance fade using `INSTANCE_ID`. If it's a simple translation offset, it's cheap but still quite effective in my experience.

roalyr:
Also, just in case, I suggest to tweak debanding constants in the project shader for the best psttern look across all colors. For instance: https://github.com/roalyr/GD4Lancer/blob/main/Assets/Shaders/Clouds/Cloud_cull_front.gdshader

RadiantUwU:
image
Material debanding fails on refractive materials.

Calinou:
@RadiantUwU The banding doesn't come from a lack of color buffer precision, but because bilinear scaling is used instead of something more elaborate like bicubic scaling. See https://github.com/godotengine/godot/pull/86047.

RadiantUwU:
> @RadiantUwU That "banding" in refractions doesn't come from a lack of color buffer precision, but because bilinear scaling is used instead of something more elaborate like bicubic scaling. See #86047.

The issue was the material somehow disappeared after it re-imported a glb file. It's resolved.

michaelharmonart:
@Calinou this is really big for us standalone VR devs, since banding in darker scenes is pretty bad in the compatibility renderer without this. 

Any blockers for this currently? I think this would be a great improvement.

michaelharmonart:
@Calinou I've realized one of the main culprits for banding in my app is the depth fog in one of the night scenes. Is the approach you're implementing here something that would work correctly with exponential/depth fog?

Calinou:
> @Calinou I've realized one of the main culprits for banding in my app is the depth fog in one of the night scenes. Is the approach you're implementing here something that would work correctly with exponential/depth fog?

Yes, that's what the testing project showcases :slightly_smiling_face:

allenwp:
I would like to spend some time deeply reviewing this PR, but I won't have time until I finish up work on the HDR output PR. That said, a couple of things come to mind at an initial glance at the code changes that I would like to quickly point out sooner rather than later:

1. This screen space dithering function was designed to be applied as the final step before quantization from floating point values to integer values. The `/ 255.0` part of the function is specific to an 8-bit buffer and it should likely be changed to `/ 1023.0` when the quantization target is a 10-bit buffer. (This probably means `/ 255.0` for Compatibility and `/ 1023.0` for Mobile.)
2. Because of point 1, I expect that having this applied in the Forward+ rendering method will produce the same issue described in #107730, but I haven't tested it yet. I believe the solution is to simply disable material dithering for the Forward+ rendering method and only have this dithering applied as the final step before quantization (as in, only for the Compatibility and Mobile rendering methods).
3. I wrote some more notes on the subject [here](https://github.com/godotengine/godot/pull/107782#issuecomment-2998573030) and [here](https://github.com/godotengine/godot/pull/107782#issuecomment-3001414299). Specifically, it's possible to implement a debanding dither that can be applied to a linear encoding, like described by Timothy Lottes, but I suspect that this still must be applied in the specific order of 1) special linear dither, 2) inverse EOTF, 3) quantization (??? I haven't deeply looked into this approach, so I might be wrong???). I don't think these exact steps apply here...
4. The original dither function from Valve was designed to be applied with a time parameter and only applied with a 37.5% strength. The strength used in Godot's debanding has historically been 100% with no time parameter. These choices should probably be revisited (maybe a different debanding strength makes sense). For example, to apply a 37.5% debanding strength, you simply replace `/ 255.0` with `* 0.375 / 255.0` or for this scenario of a 10-bit quantization replace `/ 1023.0` with `* 0.375 / 1023.0`. Maybe an even larger dither strength above `1.0` could be considered.
5. I am adding a number of comments to the existing dither function in the codebase as a part of my #107782 bugfix PR. These comments should probably be proliferated to reduce accidental misuse of the dithering function.

Again, sorry, I haven't actually tested this out, but I wanted to get the theory behind this concept voiced sooner rather than later. I'll revisit this in the future weeks... (feel free to mention me if you don't hear back!)

(Edit: I removed a possibly erroneous comment about Compatibility)

allenwp:
I'm currently [working on improving this debanding PR](https://github.com/allenwp/godot/tree/add-spatial-shader-debanding), using some of the ideas from my previous comment.

The effect is subtle and best viewed at full resolution in a dark surround, but here's what the before and after looks for Mobile (notice how material debanding with 10-bit aligned dither on mobile is still effective at debanding while producing less noise):

| 8-bit debanding (this PR) | 10 bit debanding
| --- | ---
| mobile-calinou-debanding | mobile-allenwp-debanding-material-and-viewport

The Compatibility debanding is proving to be hard to improve on for the case of additive lights. I also think I need to get a better understanding of the buffer formats that come into play...

Calinou:
> The effect is subtle and best viewed at full resolution in a dark surround, but here's what the before and after looks for Mobile (notice how material debanding with 10-bit aligned dither on mobile is still effective at debanding while producing less noise):

When viewing images at 100% zoom in a new tab, I spot some "rings" around the lights in the 10-bit debanding that aren't there in the 8-bit debanding. Is that normal, or something due to my monitor's calibration?

allenwp:
> > The effect is subtle and best viewed at full resolution in a dark surround, but here's what the before and after looks for Mobile (notice how material debanding with 10-bit aligned dither on mobile is still effective at debanding while producing less noise):
> 
> When viewing images at 100% zoom in a new tab, I spot some "rings" around the lights in the 10-bit debanding that aren't there in the 8-bit debanding. Is that normal, or something due to my monitor's calibration?

Ah, yes, the banding is a bit more visible around the yellow light with the 10-bit based debanding, while the 8-bit is more noisy in the shadows. Good catch. I'll look a bit more into it. Personally, I prefer having a the 10-bit image that has less noise overall, but maybe others would prefer having more noise and even less banding... (Both images are a massive improvement over no debanding, of course!)

Calinou:
> Personally, I prefer having a the 10-bit image that has less noise overall, but maybe others would prefer having more noise and even less banding... (Both images are a massive improvement over no debanding, of course!)

I tend to be more bothered by ringing than noise personally, especially on a display with higher pixel density where the noise becomes less noticeable, but not the ringing.

michaelharmonart:
> I'm currently [working on improving this debanding PR](https://github.com/allenwp/godot/tree/add-spatial-shader-debanding), using some of the ideas from my previous comment.
> 
> The effect is subtle and best viewed at full resolution in a dark surround, but here's what the before and after looks for Mobile (notice how material debanding with 10-bit aligned dither on mobile is still effective at debanding while producing less noise):
> 8-bit debanding (this PR) 	10 bit debanding
> mobile-calinou-debanding 	mobile-allenwp-debanding-material-and-viewport
> 
> The Compatibility debanding is proving to be hard to improve on for the case of additive lights. I also think I need to get a better understanding of the buffer formats that come into play...

10 bit looks to have less noise and less banding to me. Both look good

allenwp:
> Ah, yes, the banding is a bit more visible around the yellow light with the 10-bit based debanding, while the 8-bit is more noisy in the shadows. Good catch. I'll look a bit more into it. Personally, I prefer having a the 10-bit image that has less noise overall, but maybe others would prefer having more noise and even less banding... (Both images are a massive improvement over no debanding, of course!)

Whoops, I was looking at the images in my web browser at 100%, which is actually 125% because my OS has everything scaled to 125%. When actually view these images at 100%, I see more banding with the "8-bit debanding" image. I see banding in both, but it's noticeably worse to me in the "8-bit". This is my attempt at highlighting the sort of banding I see with the "8-bit debanding" style of the current state of this PR:

image

It might actually not be banding that I'm seeing, but instead the dither pattern, as it seems to cover the most of the image.

...So I looked into this and indeed, it seems to be the dithering pattern that I'm seeing:

image

In the screenshot I posted, it's using both material and viewport debanding, so I [made a tool for finding an offset](https://github.com/user-attachments/files/21354623/dither-offset-test.zip)) that works well with the base dithering pattern. In this image the top and bottom bars are the original colour, the left is the base dither pattern, the right is the offset dither pattern, and the middle is the left and right added to each other (in linear space):

image

(The dithered parts look darker in this example, but this isn't an issue for viewport debanding because the "darkening" pixels have more perceptual weight than the "brightening" pixels since I'm doing this blend in linear space. It's possible some brightening is happening with the linear space debanding of the material debanding... But this would be very minor for most values except the darkest.)

End result is something that has less noise than anything else I've seen for this scenario of Mobile material + viewport debanding:

image

At this point, I don't think I'm even fighting with debanding anymore... Instead, I think the main issues are relating to a [Moiré pattern](https://en.wikipedia.org/wiki/Moir%C3%A9_pattern) created by the two dithers, but even then, it's a huge strain on my eyes to even see this stuff with my monitor. 😵‍💫

All that to say, I'm concerned about increasing the dither strength any more because the dither pattern isn't a perfect silver bullet and, at least to me, I can see the dither pattern's diagonal lines appear on my two displays when it is applied on the incorrect space with only an 8-bit alignment instead of 10...

I'll keep chipping away at this to see if I can get some better behaviour from the Compatibility rendering method, but at least what this PR currently has for Compatibility debanding works a whole lot better than nothing!

allenwp:
I've mostly finished work on an second iteration of this PR that reduces what I believe is unnecessary noise by aligning the debanding to final float-to-integer quantization: https://github.com/allenwp/godot/tree/add-spatial-shader-debanding

In my branch, I did the following:

- Rebased to current `master`
- Corrected Mobile material debanding to be applied as the last step before quantization from floating point to integer values and corrected the scale to align to 10-bit integer values
- Disabled material debanding entirely when using the Forward+ rendering method
  - Because Forward+ uses floating point buffers for rendering materials, there is never a need for material debanding(??) Please correct me if I'm wrong and you know of a case where Forward+ needs this sort of material debanding.
- Consolidated the project setting for `use_debanding` so that it affects both viewport and material debanding
  - Viewport and material debanding can still be individually controlled through scripting
  - I don't see any issue with combining material and viewport debanding on the Mobile rendering method. I think most users would want this behaviour and if they don't, they can turn it off at run-time.
- Updated documentation, including a simplified project setting description that simply states that the dithering filter is applied at different steps of the rendering process depending on the rendering method and HDR 2D setting. Details about when viewport/material debanding are available are included in their documentation.

And in a second commit I did this:
 - Corrected Compatibility material debanding to be applied as the last step before quantization from floating point to integer values`*`

After all of this, I think the result is a very simple, easy to use feature that, from a layperson's perspective, simply "fixes the bug where debanding doesn't work on the Mobile and Compatibility rendering methods".

`*` There is one remaining item that could be improved: the Compatibility material debanding method adds brightness to the scene after the change in my second commit. In this way, my proposed approach for Compatibility simply doesn't work. I think it's good to merge with Calinou's approach and improve Compatibility material debanding at some point in the future. When I try and understand this PR's current approach, I think the idea is to add a lot of extra noise in the first base pass and then add only add smaller amounts of noise with each additive light pass. This way, the extra lights don't make the debanding increase the brightness of the scene. I think the reason my version increases brightness is because the specific offsets for each additive pass results in a dithering pattern that adds brightness overall, but I'm not too sure just yet.

Anyway, I've kept my failed attempt for Compatibility as a separate commit in my branch so the rest can be merged into this PR if desired. @Calinou, when you have a chance, you're welcome to review and merge my version (without the failed Compatibility commit). Or I can make a new PR with you as a co-author if that's preferable!

I expect I won't have time to work on the Compatibility material debanding for the next while, so don't hold up on me figuring out a better solution for that one.

Calinou:
> @Calinou, when you have a chance, you're welcome to review and merge my version (without the failed Compatibility commit). Or I can make a new PR with you as a co-author if that's preferable!

Feel free to open a new PR with me as a co-author. I think it's better to proceed this way, so you can make further changes more easily.

allenwp:
I have opened a [new PR](https://github.com/godotengine/godot/pull/109084/) that is a simplified and trimmed down version of the material debanding introduced by this PR for the Mobile renderer. Please give my new PR a test, as I hope it will be merged in Godot 4.6: #109084 (Download the editor build artifact from the [checks page](https://github.com/godotengine/godot/pull/109084/checks).)

@roalyr Would you be interested in testing my new PR #109084 to confirm that there aren't any substantial issues with transparency that remain? I can't easily reproduce the issue that you described, so I get the feeling this was fixed at some point.

roalyr:
@allenwp I don't use Godot4 for a while, switched to Godot3 + GLES2 for my project, so I don't really have a suitable test subject anymore.

Calinou:
Rebased and tested again, it works as expected. See OP for updated screenshots.

This PR now only features debanding support in Compatibility, since Mobile debanding was implemented in https://github.com/godotengine/godot/pull/109084/.

I've tested various scenarios and they all work: lights without shadows, lights with shadows, lightmaps. The noise level is kept to a minimum, although it will be more noticeable when you have multiple overlapping lights with shadows enabled (due to the use of multipass rendering). That said, if you have a lot of overlapping lights with shadows enabled, you're likely to run into performance issues in the first place given the multipass approach used in Compatibility for these lights.

allenwp:
Thanks for the rebase!

This debanding approach should never be applied when the shader is writing to a floating point buffer; it should only be applied when writing to an integer buffer (with `/255` for an 8-bit buffer and `/1023` for a 10-bit buffer).

With the current state of this PR, enabling HDR 2D switches the buffer formats to floating point buffers, but debanding is still applied. In this scenario debanding should not be applied. This should be handled per-viewport, given that different viewports can have different HDR 2D settings.

clayjohn:
I'm not sure it is worth doing in the compatibility renderer. No one has complained about banding. In fact, everyone who has complained has pointed out that the compatibility renderer works great. 

So ultimately, I'm not sure the extra complexity is worth adding

michaelharmonart:
I would die to have this in compatibility 

for VR banding is extremely distracting, and right now in dark scenes it's pretty atrocious

dsnopek:
> I'm not sure it is worth doing in the compatibility renderer. No one has complained about banding. In fact, everyone who has complained has pointed out that the compatibility renderer works great.

I would second the idea that banding isn't a big problem in compatibility. I use compatibility in 95% of my projects, and it has never been a problem for me (even in VR). It's the times I've tested switching my projects to the mobile renderer that I've noticed serious banding issues.

That's not to say there isn't any banding in compatibility, or that we shouldn't do this - having more options is nice :-) - just that it's never struck me as a big problem

allenwp:
> I would die to have this in compatibility 
> 
> 
> 
> for VR banding is extremely distracting, and right now in dark scenes it's pretty atrocious

Actually, this is an important thing to consider; when creating dark scenes in some VR headsets, debanding is very necessary. We needed this for Keep Talking and Nobody Explodes for Gear VR when the lights went out in our virtual scene; on those old HMDs, the brightness of the display was quite high, so huge amounts of banding was visible when writing to an 8-bit buffer.

That said, the solution here might be to add Viewport debanding for Compatibility HDR 2D. I believe this would resolve this issue without the need for material debanding. Viewport debanding would be added to `post.glsl` and would only be applied when HDR 2D is enabled.

Does this sound like it would solve your issue @michaelharmonart? Or is there a strong reason that it is impossible for you to adapt your game to use HDR 2D with the Compatibility renderer?

Calinou:
> No one has complained about banding. 

While the issue isn't as bad as in Mobile, it's still present in Compatibility to an extent *at least* similar to Forward+ when debanding is off (usually greater). While I didn't create an issue for it (as it wasn't a bug per se), I personally found it to be a big enough issue to be worth resolving.

It largely depends on art style; the moment you have large flat color planes (especially with fog), debanding becomes a must-have in my book. In some pixel art games, it can also be important if the texel density is low enough that the texels end up forming large enough flat planes when viewed up close. A lot of games that use Compatibility go for a simplified art style (sometimes with vertex shading to boot), where banding is more likely to be seen.

> Or is there a strong reason that it is impossible for you to adapt your game to use HDR 2D with the Compatibility renderer?

The performance cost of HDR 2D in standalone VR might be prohibitive, given the memory bandwidth constraint. It's probably fine in PCVR though.

That said, if Compatibility with HDR 2D uses a RGBAH buffer, it makes sense to implement Viewport debanding for consistency with the Mobile renderer.

michaelharmonart:
> Does this sound like it would solve your issue @michaelharmonart? Or is there a strong reason that it is impossible for you to adapt your game to use HDR 2D with the Compatibility renderer?

edit:

Looks like I was probably using viewport debanding with a 10 bit integer buffer last time I tested and that's why I wasn't impressed.

I'll have to give HDR 2D a try (I didn't know it was an option for 3D) and see what the performance is like.

roalyr (approved):

pyhfen:
I personally use `return (dither.rgb / 255.0) * 0.15;` & glow breaks it. I use a lower value since i only want debanding. I dont want debanding to add a noise problem.

Here are some comparisons.
comparedeband

`return (dither.rgb / 255.0) * 0.15;` & `return (dither.rgb - 0.85) / 255.0;` give the same result in maths. I dont understand graphics maths so i dont know why it darkens the image.

comparedeband2

Increasing the bloom value gives me an idea as to why glow breaks debanding. I think glow IS banding. In all the screen shots bloom was set to 0.
glowISbanding

Side by side with `return (dither.rgb / 255.0) * 0.15;`.
glowbloombanding

Calinou:
The glow pass itself likely needs debanding to avoid this problem. This should be tackled separately (if it can be done at all in Compatibility, given the performance constraints).

pyhfen:
I dont think glow pass needs debanding. Debanding works fine but not at lower values. At higher values theres too much noise.

EDIT:
Ive removed 2 lines of code from `scene.glsl`, currently i cant even see the noise pattern & theres no banding.
This is with `return (dither.rgb - 0.5) / 255.0;`
nobanding

The problem with this one was the excessive noise but its not there anymore. Simply remove `frag_color.rgb = linear_to_srgb(frag_color.rgb);` & `additive_light_color = linear_to_srgb(additive_light_color);`. I ve removed them because these 2 lines cause shadowed & non shadowed lights to look different & for some reason it effects noise from this PR.

PR #87353, opened 2024-01-19
Implement maximum undo steps in UndoRedo

### LINKED ISSUE #3575: Reimplement maximum undo steps in UndoRedo

A 3D painting application.

While adding undo/redo support to mesh painting, I realized it would be useful to be able to limit the undo actions to save on memory/disc space, as there are a lot of textures to be saved per brush stroke. This is currently not possible in Godot's native UndoRedo class.

Readd the `get_max_steps` / `set_max_steps` methods removed in godotengine/godot#15460. Steps that are older than `max_steps` would be freed from memory, and couldn't be reached by calling `undo()`.

```gdscript
var ur := UndoRedo()
ur.max_steps = 1

ur.create_action()
ur.add_do_property(self, "trade", Trade.new("apples", "eggs"))
ur.commit()

# Could undo the last action.

ur.create_action()
ur.add_do_property(self, "trade", Trade.new("bananas", "apples"))
ur.commit()

# The first action isn't available anymore and the trade is freed from memory.
```

It can be worked around with the native UndoRedo, but then you could also just reimplement it completely.

It's common undo functionality, so it can be added in core.

OverloadedOrama:
Implements https://github.com/godotengine/godot-proposals/issues/3575 and brings back the removed code from #15494, with added documentation.

For the future, I think adding a limit for the maximum memory usage as suggested in https://github.com/godotengine/godot-proposals/issues/3575#issuecomment-974671845 is still worth it, if max steps and maximum memory can even work together, but I don't know how to implement that.

Tested with [Pixelorama](https://github.com/Orama-Interactive/Pixelorama), it seems to be working as intended.

_Production edit: Closes https://github.com/godotengine/godot-proposals/issues/3575_

### DIFF
--- a/core/object/undo_redo.cpp
+++ b/core/object/undo_redo.cpp
@@ -316,6 +316,14 @@ void UndoRedo::commit_action(bool p_execute) {
 	_redo(p_execute); // perform action
 	committing--;
 
+	if (max_steps > 0) {
+		// Clear early steps.
+
+		while (actions.size() > max_steps) {
+			_pop_history_tail();
+		}
+	}
+
 	if (add_message && callback && actions.size() > 0) {
 		callback(callback_ud, actions[actions.size() - 1].name);
 	}
@@ -473,6 +481,14 @@ uint64_t UndoRedo::get_version() const {
 	return version;
 }
 
+void UndoRedo::set_max_steps(int p_max_steps) {
+	max_steps = p_max_steps;
+}
+
+int UndoRedo::get_max_steps() const {
+	return max_steps;
+}
+
 void UndoRedo::set_commit_notify_callback(CommitNotifyCallback p_callback, void *p_ud) {
 	callback = p_callback;
 	callback_ud = p_ud;
@@ -517,8 +533,12 @@ void UndoRedo::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("has_undo"), &UndoRedo::has_undo);
 	ClassDB::bind_method(D_METHOD("has_redo"), &UndoRedo::has_redo);
 	ClassDB::bind_method(D_METHOD("get_version"), &UndoRedo::get_version);
+	ClassDB::bind_method(D_METHOD("set_max_steps", "max_steps"), &UndoRedo::set_max_steps);
+	ClassDB::bind_method(D_METHOD("get_max_steps"), &UndoRedo::get_max_steps);
 	ClassDB::bind_method(D_METHOD("redo"), &UndoRedo::redo);
 	ClassDB::bind_method(D_METHOD("undo"), &UndoRedo::undo);
+
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "max_steps", PROPERTY_HINT_RANGE, "0,50,1,or_greater"), "set_max_steps", "get_max_steps");
 
 	ADD_SIGNAL(MethodInfo("version_changed"));
 
--- a/doc/classes/UndoRedo.xml
+++ b/doc/classes/UndoRedo.xml
@@ -255,6 +255,11 @@
 			</description>
 		</method>
 	</methods>
+	<members>
+		<member name="max_steps" type="int" setter="set_max_steps" getter="get_max_steps" default="0">
+			The maximum number of steps that can be stored in the undo/redo history. If the number of stored steps exceeds this limit, older steps are removed from history and can no longer be reached by calling [method undo]. A value of [code]0[/code] or lower means no limit.
+		</member>
+	</members>
 	<signals>
 		<signal name="version_changed">
 			<description>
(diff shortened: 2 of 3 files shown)


core/object/undo_redo.cpp
```diff
@@ -316,6 +316,14 @@ void UndoRedo::commit_action(bool p_execute) {
 	_redo(p_execute); // perform action
 	committing--;
 
+	if (max_steps > 0 && actions.size() > max_steps) {
+		// Clear early steps
```
akien-mga, line 320:
```suggestion
		// Clear early steps.
```

core/object/undo_redo.cpp
```diff
@@ -517,6 +533,9 @@ void UndoRedo::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("has_undo"), &UndoRedo::has_undo);
 	ClassDB::bind_method(D_METHOD("has_redo"), &UndoRedo::has_redo);
 	ClassDB::bind_method(D_METHOD("get_version"), &UndoRedo::get_version);
+	ClassDB::bind_method(D_METHOD("set_max_steps", "max_steps"), &UndoRedo::set_max_steps);
+	ClassDB::bind_method(D_METHOD("get_max_steps"), &UndoRedo::get_max_steps);
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "max_steps"), "set_max_steps", "get_max_steps");
```
AThousandShips, line 538: Should be moved down between methods and signals

Mickeon: I'd say around 50 is a pretty sensible limit. By putting "or_greater" users can go above it, anyway.

AThousandShips: Agreed, the limit is just to scale the property editor when relevant

core/object/undo_redo.cpp
```diff
@@ -316,6 +316,14 @@ void UndoRedo::commit_action(bool p_execute) {
 	_redo(p_execute); // perform action
 	committing--;
 
+	if (max_steps > 0 && actions.size() > max_steps) {
```
KoBeWi, line 319:
```suggestion
	if (max_steps > 0) {
```
The second check is already handled by `while`.

core/object/undo_redo.cpp
```diff
@@ -517,8 +533,11 @@ void UndoRedo::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("has_undo"), &UndoRedo::has_undo);
 	ClassDB::bind_method(D_METHOD("has_redo"), &UndoRedo::has_redo);
 	ClassDB::bind_method(D_METHOD("get_version"), &UndoRedo::get_version);
+	ClassDB::bind_method(D_METHOD("set_max_steps", "max_steps"), &UndoRedo::set_max_steps);
+	ClassDB::bind_method(D_METHOD("get_max_steps"), &UndoRedo::get_max_steps);
 	ClassDB::bind_method(D_METHOD("redo"), &UndoRedo::redo);
 	ClassDB::bind_method(D_METHOD("undo"), &UndoRedo::undo);
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "max_steps", PROPERTY_HINT_RANGE, "0,50,1,or_greater"), "set_max_steps", "get_max_steps");
```
AThousandShips, line 540:
```suggestion

	ADD_PROPERTY(PropertyInfo(Variant::INT, "max_steps", PROPERTY_HINT_RANGE, "0,50,1,or_greater"), "set_max_steps", "get_max_steps");
```

KoBeWi:
The history is trimmed too late. When you commit an action that goes over maximum step count, the previous action will be only removed after this one finishes committing, so the actual step count is over the limit. If you e.g. use `get_history_count()` inside a redo operation, it will return a wrong value.

KoBeWi (approved):

doc/classes/UndoRedo.xml
```diff
@@ -255,6 +255,11 @@
 			</description>
 		</method>
 	</methods>
+	<members>
+		<member name="max_steps" type="int" setter="set_max_steps" getter="get_max_steps" default="0">
+			The maximum value of steps that can be stored in the undo/redo history. Steps older than this value are being freed from memory and can't be reached by calling [method undo]. A value of [code]0[/code] or lower means no limit.
```
YuriSizov, line 260:
```suggestion
			The maximum number of steps that can be stored in the undo/redo history. If the number of stored steps exceeds this limit, older steps are removed from history and can no longer be reached by calling [method undo]. A value of [code]0[/code] or lower means no limit.
```

YuriSizov (approved):
Docs needs some adjustments, but looks good otherwise.

YuriSizov:
Thanks!

PR #87358, opened 2024-01-19
[iOS] Update linker flags for Xcode 15.2.

bruvzg:
Xcode 15.2 fixed the linker bug, so `ld_classic` is necessary only for older versions of Xcode 15.


akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

akien-mga:
Cherry-picked for 4.1.4.

Issue #57015, opened 2022-01-20
Luminance calculation formula appears to be wrong


Ansraer:
### Godot version

4.0

### System information

Windows 10 Vulkan

### Issue description

I am currently reading Godot's shader code and was a bit stumped when I stumbled across how the luminance was calculated. Godot currently uses `luminance = max(color.r, max(color.g, color.b));`.

While this estimation should work okay-ish for most use cases it is not technically correct, since the human eye is far more sensitive to green colors.

According to e.g. Fillament or Wikipedia more correct formulas would be:
`L = 0.299*R + 0.587*G + 0.114*B` and `L = 0.2126*R + 0.7152*G + 0.0722*B`  for gamma corrected content.
(in practise this should obviously be calculated as a dot product)

This means that in a worst-case scenario (e.g. when the color is `{0,0,1}`) the currently calculated luminance value could be many times higher than it should be.

I did find a comment in luminance_reduce_raster.glsl#L59 that mentions the correct formula but implies that using it overexposes the result. This might imply that there are some other problems with how Godot handles luminance.

### Steps to reproduce

1. Open shaders/copy.glsl
2. Scroll to line 189
3. Be confused

### Minimal reproduction project

Should be present in all Godot projects.

### Effected shaders I found so far:
copy.glsl#L189
effect_blur.glsl#L285
blur_raster.glsl#L126
luminance_reduce_raster.glsl#L57
luminance_reduce.glsl#L47

advent94:
@Calinou this change was merged: https://github.com/godotengine/godot/blob/master/core/math/color.h

PR #87360, opened 2024-01-19
Implement glow/bloom on compatibility renderer

### LINKED ISSUE #66455: OpenGL: Environment glow not implemented yet
*Related to https://github.com/godotengine/godot/issues/66457.*

### Godot version

4.0.beta (53d2a9acd)

### System information

Fedora 36, GeForce GTX 1080 (NVIDIA 515.65.01)

### Issue description

Glow has no effect on the displayed scene when enabled in OpenGL, regardless of the value of **Bloom** or **HDR Threshold**.

Both glow and adjustments are enabled on the example below.

### Steps to reproduce

- Switch to the OpenGL driver in the Project Settings.
  - Make sure to start the editor with single-window mode enabled afterwards. This can be done by enabling **Single Window Mode** in the Editor Settings before changing the driver, or by opening the project in the editor directly with the `--single-window` [command line argument](https://docs.godotengine.org/en/latest/tutorials/editor/command_line_tutorial.html).
- Add a WorldEnvironment node, a Camera3D node and a MeshInstance3D in front of the camera.
- Enable glow and adjustments. Change brightness/contrast/saturation to make a visible difference in the scene.

[test_opengl_glow_adjustments_editor.zip](https://github.com/godotengine/godot/files/9648499/test_opengl_glow_adjustments_editor.zip)

BastiaanOlij:
This is a fairly straight forward and basic implementation of glow. For now we've left out some of the bells and whistles to keep this performant on mobile devices.

This PR also enabled/abuses the `2D HDR` setting on viewports to cause the viewport to use an RGBA16F buffer allowing for high dynamic range for both 2D and 3D output. Colors are still output in sRGB color space so it's more an approximation but it works.

The glow works using "Dual filtering" to blur the source image which is more limited to what we're doing in the Vulkan renderer (but much more bandwidth efficient). Right now we're also using simple blending to apply the glow to the final image, functional and fast, but not as pretty as what the Vulkan render does.

image

See [Marius Bjørge presentation at Siggraph 2015 "Bandwidth-Efficient Rendering"](https://community.arm.com/cfs-file/__key/communityserver-blogs-components-weblogfiles/00-00-00-20-66/siggraph2015_2D00_mmg_2D00_marius_2D00_slides.pdf) for more info on the blur effect being used.

Without HDR enabled we need to apply our luminance multiplier trick that we also use on Vulkan Mobile (but now in sRGB space which in theory should have less banding issues) where we use an RGB10A2 buffer.

> Note: Glow on a viewport with a transparent background **only** works when HDR is enabled. We can not use the luminance multiplier trick in this case due to using RGBA8 buffers.
> If glow is used with stereo rendering, this disables the benefits of foveated rendering as we are using an intermediate buffer.

- *Production edit: This closes https://github.com/godotengine/godot/issues/66455.*
- *Production edit: This closes godotengine/internal-team-priorities#16*

### DIFF
--- a/drivers/gles3/shaders/effects/glow.glsl
+++ b/drivers/gles3/shaders/effects/glow.glsl
@@ -1,0 +1,113 @@
+/* clang-format off */
+#[modes]
+
+// Based on Dual filtering glow as explained in Marius Bjørge presentation at Siggraph 2015 "Bandwidth-Efficient Rendering"
+
+mode_filter = #define MODE_FILTER
+mode_downsample = #define MODE_DOWNSAMPLE
+mode_upsample = #define MODE_UPSAMPLE
+
+#[specializations]
+
+USE_MULTIVIEW = false
+
+#[vertex]
+layout(location = 0) in vec2 vertex_attrib;
+
+/* clang-format on */
+
+out vec2 uv_interp;
+
+void main() {
+	uv_interp = vertex_attrib * 0.5 + 0.5;
+	gl_Position = vec4(vertex_attrib, 1.0, 1.0);
+}
+
+/* clang-format off */
+#[fragment]
+/* clang-format on */
+
+#ifdef MODE_FILTER
+#ifdef USE_MULTIVIEW
+uniform sampler2DArray source_color; // texunit:0
+#else
+uniform sampler2D source_color; // texunit:0
+#endif // USE_MULTIVIEW
+uniform float view;
+uniform vec2 pixel_size;
+uniform float luminance_multiplier;
+uniform float glow_bloom;
+uniform float glow_hdr_threshold;
+uniform float glow_hdr_scale;
+uniform float glow_luminance_cap;
+#endif // MODE_FILTER
+
+#ifdef MODE_DOWNSAMPLE
+uniform sampler2D source_color; // texunit:0
+uniform vec2 pixel_size;
+#endif // MODE_DOWNSAMPLE
+
+#ifdef MODE_UPSAMPLE
+uniform sampler2D source_color; // texunit:0
+uniform vec2 pixel_size;
+#endif // MODE_UPSAMPLE
+
+in vec2 uv_interp;
+
+layout(location = 0) out vec4 frag_color;
+
+void main() {
+#ifdef MODE_FILTER
+	// Note, we read from an image with double resolution, so we average those out
+#ifdef USE_MULTIVIEW
+	vec2 half_pixel = pixel_size * 0.5;
+	vec3 uv = vec3(uv_interp, view);
+	vec3 color = textureLod(source_color, uv, 0.0).rgb * 4.0;
+	color += textureLod(source_color, uv - vec3(half_pixel, 0.0), 0.0).rgb;
+	color += textureLod(source_color, uv + vec3(half_pixel, 0.0), 0.0).rgb;
+	color += textureLod(source_color, uv - vec3(half_pixel.x, -half_pixel.y, 0.0), 0.0).rgb;
+	color += textureLod(source_color, uv + vec3(half_pixel.x, -half_pixel.y, 0.0), 0.0).rgb;
+#else
+	vec2 half_pixel = pixel_size * 0.5;
+	vec2 uv = uv_interp;
+	vec3 color = textureLod(source_color, uv, 0.0).rgb * 4.0;
+	color += textureLod(source_color, uv - half_pixel, 0.0).rgb;
+	color += textureLod(source_color, uv + half_pixel, 0.0).rgb;
+	color += textureLod(source_color, uv - vec2(half_pixel.x, -half_pixel.y), 0.0).rgb;
+	color += textureLod(source_color, uv + vec2(half_pixel.x, -half_pixel.y), 0.0).rgb;
+#endif // USE_MULTIVIEW
+	color /= luminance_multiplier * 8.0;
+
+	float luminance = dot(color, vec3(0.2126, 0.7152, 0.0722));
+	float feedback = max(smoothstep(glow_hdr_threshold, glow_hdr_threshold + glow_hdr_scale, luminance), glow_bloom);
+
+	color = min(color * feedback, vec3(glow_luminance_cap));
+
+	frag_color = vec4(luminance_multiplier * color, 1.0);
+#endif // MODE_FILTER
+
+#ifdef MODE_DOWNSAMPLE
+	vec2 half_pixel = pixel_size * 0.5;
+	vec4 color = textureLod(source_color, uv_interp, 0.0) * 4.0;
+	color += textureLod(source_color, uv_interp - half_pixel, 0.0);
+	color += textureLod(source_color, uv_interp + half_pixel, 0.0);
+	color += textureLod(source_color, uv_interp - vec2(half_pixel.x, -half_pixel.y), 0.0);
+	color += textureLod(source_color, uv_interp + vec2(half_pixel.x, -half_pixel.y), 0.0);
+	frag_color = color / 8.0;
+#endif // MODE_DOWNSAMPLE
+
+#ifdef MODE_UPSAMPLE
+	vec2 half_pixel = pixel_size * 0.5;
+
+	vec4 color = textureLod(source_color, uv_interp + vec2(-half_pixel.x * 2.0, 0.0), 0.0);
+	color += textureLod(source_color, uv_interp + vec2(-half_pixel.x, half_pixel.y), 0.0) * 2.0;
+	color += textureLod(source_color, uv_interp + vec2(0.0, half_pixel.y * 2.0), 0.0);
+	color += textureLod(source_color, uv_interp + vec2(half_pixel.x, half_pixel.y), 0.0) * 2.0;
+	color += textureLod(source_color, uv_interp + vec2(half_pixel.x * 2.0, 0.0), 0.0);
+	color += textureLod(source_color, uv_interp + vec2(half_pixel.x, -half_pixel.y), 0.0) * 2.0;
+	color += textureLod(source_color, uv_interp + vec2(0.0, -half_pixel.y * 2.0), 0.0);
+	color += textureLod(source_color, uv_interp + vec2(-half_pixel.x, -half_pixel.y), 0.0) * 2.0;
+
+	frag_color = color / 12.0;
+#endif // MODE_UPSAMPLE
+}
--- a/drivers/gles3/shaders/effects/post.glsl
+++ b/drivers/gles3/shaders/effects/post.glsl
@@ -1,0 +1,96 @@
+/* clang-format off */
+#[modes]
+mode_default = #define MODE_DEFAULT
+// mode_glow = #define MODE_GLOW
+
+#[specializations]
+
+USE_MULTIVIEW = false
+USE_GLOW = false
+USE_LUMINANCE_MULTIPLIER = false
+
+#[vertex]
+layout(location = 0) in vec2 vertex_attrib;
+
+/* clang-format on */
+
+out vec2 uv_interp;
+
+void main() {
+	uv_interp = vertex_attrib * 0.5 + 0.5;
+	gl_Position = vec4(vertex_attrib, 1.0, 1.0);
+}
+
+/* clang-format off */
+#[fragment]
+/* clang-format on */
+
+#include "../tonemap_inc.glsl"
+
+#ifdef USE_MULTIVIEW
+uniform sampler2DArray source_color; // texunit:0
+#else
+uniform sampler2D source_color; // texunit:0
+#endif // USE_MULTIVIEW
+
+uniform float view;
+uniform float luminance_multiplier;
+
+#ifdef USE_GLOW
+uniform sampler2D glow_color; // texunit:1
+uniform vec2 pixel_size;
+uniform float glow_intensity;
+
+vec4 get_glow_color(vec2 uv) {
+	vec2 half_pixel = pixel_size * 0.5;
+
+	vec4 color = textureLod(glow_color, uv + vec2(-half_pixel.x * 2.0, 0.0), 0.0);
+	color += textureLod(glow_color, uv + vec2(-half_pixel.x, half_pixel.y), 0.0) * 2.0;
+	color += textureLod(glow_color, uv + vec2(0.0, half_pixel.y * 2.0), 0.0);
+	color += textureLod(glow_color, uv + vec2(half_pixel.x, half_pixel.y), 0.0) * 2.0;
+	color += textureLod(glow_color, uv + vec2(half_pixel.x * 2.0, 0.0), 0.0);
+	color += textureLod(glow_color, uv + vec2(half_pixel.x, -half_pixel.y), 0.0) * 2.0;
+	color += textureLod(glow_color, uv + vec2(0.0, -half_pixel.y * 2.0), 0.0);
+	color += textureLod(glow_color, uv + vec2(-half_pixel.x, -half_pixel.y), 0.0) * 2.0;
+
+	return color / 12.0;
+}
+#endif // USE_GLOW
+
+in vec2 uv_interp;
+
+layout(location = 0) out vec4 frag_color;
+
+void main() {
+#ifdef USE_MULTIVIEW
+	vec4 color = texture(source_color, vec3(uv_interp, view));
+#else
+	vec4 color = texture(source_color, uv_interp);
+#endif
+
+#ifdef USE_GLOW
+	vec4 glow = get_glow_color(uv_interp) * glow_intensity;
+
+	// Just use softlight...
+	glow.rgb = clamp(glow.rgb, vec3(0.0f), vec3(1.0f));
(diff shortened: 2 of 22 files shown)


BastiaanOlij:
This is now working both in HDR and normal mode. In normal mode we render to RGB10A2 as before, but if glow is enabled we apply a multiplier. Needs testing to see if this leads to unwanted banding. We could play with the multiplier to see if we can find a better balance if needed.

Also still need to test in stereo

BastiaanOlij:
@clayjohn would be good if you could have a first look at this and give some feedback on what else we should do here.

clayjohn:
We discussed hiding the "level" properties and the "normalized" property.

You can hide them from the editor here (note this only hides it from the editor, it doesn't change saving/loading):
https://github.com/godotengine/godot/blob/51991e20143a39e9ef0107163eaf283ca0a761ea/scene/resources/environment.cpp#L1065-L1071


Then you just have to remember to call `notify_property_list_changed()` here:
https://github.com/godotengine/godot/blob/51991e20143a39e9ef0107163eaf283ca0a761ea/scene/resources/environment.cpp#L602-L620

BastiaanOlij:
> We discussed hiding the "level" properties and the "normalized" property.

I've hidden all glow properties we currently don't support. There might be one or two that we can add support for. 
Clang format was being a bit annoying though.

I wasn't sure if `notify_property_list_changed` was required, it's called on `set_glow_enabled` which is the only thing that impacts visibility here.

Calinou (commented):
Tested locally, it works as expected in 3D.

**Testing project:** [test_glow_compatibility.zip](https://github.com/godotengine/godot/files/14101582/test_glow_compatibility.zip)

One issue however is that since the preview environment has glow enabled by default, it looks very focused and very strong in Compatibility:

Forward+ | Mobile | Compatibility
-|-|-
Screenshot_20240130_174402 | Screenshot_20240130_174416 | Screenshot_20240130_174349

To avoid this, I suggest disabling glow in the preview environment (at least when Compatibility is used), which will also help improve performance on integrated graphics.

In 2D, setting background mode to **Canvas** then enabling glow results in a black viewport:

Glow disabled | Glow enabled
-|-
image | image

This occurs regardless of whether **HDR 2D** is enabled in the project settings.

Also, I'm surprised only the **Intensity** setting is available. Shouldn't there be **Bloom** and **HDR Threshold** properties at least, so you get more control over where glow appears?

BastiaanOlij:
@Calinou I added support for bloom and the thresholds, turned out to be easier than I thought, that makes it work pretty close to the original.

I'll have a look at the canvas mode, I think that might be a result of 2D stuff not being copied to the new intermediate buffer.

BastiaanOlij:
Ok, fixed the canvas mode as well. Do note that 2D rendering caps the values so glow requires some bloom or a lower hdr threshold. Hopefully that is ok enough.

drivers/gles3/shaders/effects/glow.glsl
```diff
@@ -0,0 +1,113 @@
...
+	vec2 uv = uv_interp;
+	vec3 color = textureLod(source_color, uv, 0).rgb * 4.0;
+	color += textureLod(source_color, uv - half_pixel, 0).rgb;
+	color += textureLod(source_color, uv + half_pixel, 0).rgb;
+	color += textureLod(source_color, uv - vec2(half_pixel.x, -half_pixel.y), 0).rgb;
+	color += textureLod(source_color, uv + vec2(half_pixel.x, -half_pixel.y), 0).rgb;
+#endif // USE_MULTIVIEW
+	color /= luminance_multiplier * 8.0;
+
+	float luminance = max(color.r, max(color.g, color.b));
```
Ansraer, line 81: No no no no no no no
Not again, see #57015

Ansraer (changes requested):

DanielSnd:
I was testing this PR and disabling Depth Prepass in project settings seems to completely break rendering for me. Can reproduce the issue in @Calinou 's test project.

`driver/depth_prepass/enable=false`

drivers/gles3/shaders/effects/glow.glsl
BastiaanOlij: I'm not sure it does a very good job here, it basically only creates a glow effect if you overblow everything. With that formula the luminance value stays way too low unless you get near pure whites. I have a feeling that our emission color gets clamped somewhere even after our energy multiplier is applied.

BastiaanOlij: Ahhhhh, @clayjohn I think I might have found a snag, unless this is just the way it is. The problem here lies in tonemapping that blunts our color output. If I turn off tonemapping it behaves beautifully.  Normally we perform glow before tonemapping.

I'm wondering, seeing we're doing a post pass for glow anyway, should we disable the early tonemapping in our scene shader and do it as part of our post process shader?

BastiaanOlij:
> I was testing this PR and disabling Depth Prepass in project settings seems to completely break rendering for me. Can reproduce the issue in @Calinou 's test project.
> 
> `driver/depth_prepass/enable=false`

Good catch Daniel. Actually possibly not broken by this PR but already broken, writing to depth was disabled when clearing the depth buffer. Moved some code forward to fix.

drivers/gles3/shaders/effects/glow.glsl
BastiaanOlij: Aaaaaaand done! AND ITS SO MUCH BETTER!

Tonemapping in now applied in post IF glow is enabled (or 3D scaling is enabled). And it looks sooo much nicer :)

DanielSnd:
> > I was testing this PR and disabling Depth Prepass in project settings seems to completely break rendering for me. Can reproduce the issue in @Calinou 's test project.
> > `driver/depth_prepass/enable=false`
> 
> Good catch Daniel. Actually possibly not broken by this PR but already broken, writing to depth was disabled when clearing the depth buffer. Moved some code forward to fix.

Found a new issue with it, if there's a shader that's has the render modes `cull_front` and also `depth_draw_never` in the scene it either freezes the scene view or makes it really slow (depending on the complexity of the mesh I believe).

```
shader_type spatial;

render_mode cull_front,depth_draw_never;
```

Repro steps:
```
1 - Create a spatial shader with render_mode cull_front,depth_draw_never;
2 - Assign to a mesh in the scene (more complex meshes seem to make the problem worse).
3 - Turn off glow on world environment.
4 - Turn on glow on world environment.
5 - Try to rotate the scene view around, it'll be frozen or rotating with very delayed iterations.
```

drivers/gles3/shaders/effects/post.glsl
```diff
@@ -0,0 +1,96 @@
...
+
+#ifdef USE_GLOW
+uniform sampler2D glow_color; // texunit:1
+uniform vec2 pixel_size;
+uniform float glow_intensity;
+
+vec4 get_glow_color(vec2 uv) {
+	vec2 half_pixel = pixel_size * 0.5;
+
+	vec4 color = textureLod(glow_color, uv + vec2(-half_pixel.x * 2.0, 0.0), 0);
```
DanielSnd, line 47: When trying to run it on android I got an error about the textureLod not finding the correct overload. Instead of `, 0);` at the end it should be 0.0 so it's interpreted as a float instead of an int.

Updated it becomes:
```glsl
vec4 color = textureLod(glow_color, uv + vec2(-half_pixel.x * 2.0, 0.0), 0.0);
	color += textureLod(glow_color, uv + vec2(-half_pixel.x, half_pixel.y), 0.0) * 2.0;
	color += textureLod(glow_color, uv + vec2(0.0, half_pixel.y * 2.0), 0.0);
	color += textureLod(glow_color, uv + vec2(half_pixel.x, half_pixel.y), 0.0) * 2.0;
	color += textureLod(glow_color, uv + vec2(half_pixel.x * 2.0, 0.0), 0.0);
	color += textureLod(glow_color, uv + vec2(half_pixel.x, -half_pixel.y), 0.0) * 2.0;
	color += textureLod(glow_color, uv + vec2(0.0, -half_pixel.y * 2.0), 0.0);
	color += textureLod(glow_color, uv + vec2(-half_pixel.x, -half_pixel.y), 0.0) * 2.0;
```

drivers/gles3/shaders/effects/glow.glsl
```diff
@@ -0,0 +1,113 @@
...
+	color += textureLod(source_color, uv - vec3(half_pixel.x, -half_pixel.y, 0.0), 0).rgb;
+	color += textureLod(source_color, uv + vec3(half_pixel.x, -half_pixel.y, 0.0), 0).rgb;
+#else
+	vec2 half_pixel = pixel_size * 0.5;
+	vec2 uv = uv_interp;
+	vec3 color = textureLod(source_color, uv, 0).rgb * 4.0;
+	color += textureLod(source_color, uv - half_pixel, 0).rgb;
+	color += textureLod(source_color, uv + half_pixel, 0).rgb;
+	color += textureLod(source_color, uv - vec2(half_pixel.x, -half_pixel.y), 0).rgb;
+	color += textureLod(source_color, uv + vec2(half_pixel.x, -half_pixel.y), 0).rgb;
```
DanielSnd, line 77: When trying to run it on android I got an error about the textureLod not finding the correct overload. Instead of , 0); at the end it should be 0.0 so it's interpreted as a float instead of an int.

Updated it becomes:

```glsl
vec3 color = textureLod(source_color, uv, 0.0).rgb * 4.0;
	color += textureLod(source_color, uv - vec3(half_pixel, 0.0), 0.0).rgb;
	color += textureLod(source_color, uv + vec3(half_pixel, 0.0), 0.0).rgb;
	color += textureLod(source_color, uv - vec3(half_pixel.x, -half_pixel.y, 0.0), 0.0).rgb;
	color += textureLod(source_color, uv + vec3(half_pixel.x, -half_pixel.y, 0.0), 0.0).rgb;
#else
	vec2 half_pixel = pixel_size * 0.5;
	vec2 uv = uv_interp;
	vec3 color = textureLod(source_color, uv, 0.0).rgb * 4.0;
	color += textureLod(source_color, uv - half_pixel, 0.0).rgb;
	color += textureLod(source_color, uv + half_pixel, 0.0).rgb;
	color += textureLod(source_color, uv - vec2(half_pixel.x, -half_pixel.y), 0.0).rgb;
	color += textureLod(source_color, uv + vec2(half_pixel.x, -half_pixel.y), 0.0).rgb;	
```

Changing from 0 to 0.0 works and no longer fails the compilation on android.

DanielSnd (changes requested):
Found some issues when trying to run with this PR on android, it failed the shader compilation.

```c++
E 0:00:14:0887   _display_error_with_code: PostShaderGLES3: Fragment shader compilation failed:
ERROR: 0:250: 'textureLod' : no matching overloaded function found 
ERROR: 0:250: '=' :  cannot convert from 'const float' to '4-component vector of float'
ERROR: 0:251: 'textureLod' : no matching overloaded function found 
ERROR: 0:252: 'textureLod' : no matching overloaded function found 
ERROR: 0:253: 'textureLod' : no matching overloaded function found 
ERROR: 0:254: 'textureLod' : no matching overloaded function found 
ERROR: 0:255: 'textureLod' : no matching overloaded function found 
ERROR: 0:256: 'textureLod' : no matching overloaded function found 
ERROR: 0:257: 'textureLod' : no matching overloaded function found 
ERROR: 9 compilation errors.  No code generated.
```

I found out that it's because on the textureLod calls it's using 0 instead of 0.0 so it's being interpreted as an integer. Changing it to 0.0 in all the textureLod calls fixed the issue.

DanielSnd:
Not sure if this is the right place to put this, but I've been trying to check how the glow would run on my quest 2. After the fixes I pointed out above I managed to get it to run, however if I have both Glow and the MSAA enabled (in compatibility mode with the merged PR that introduces MSAA to compatibility mode) the game freezes on load and I have to force-restart the device.

If I disable MSAA the game starts just fine and I can see the glow effect, no errors, no problems.

If I disable glow and leave MSAA enabled the game starts just fine and I can see the MSAA effect, no glow but no errors no problems.

Having both enabled at the same time causes the freeze on load. (Black screen frozen)

I also tried to cheese the device by starting with MSAA on and glow off, and added a timer to a GDScript that enabled glow after 5 seconds. It still froze the device but in a different way. It took me to passthrough, and didn't recognize controllers or anything. Pressing power twice took me back to the game, I could look around for a moment, I didn't see any glow, but then it took me right back to passthrough.

BastiaanOlij:
> Found a new issue with it, if there's a shader that's has the render modes `cull_front` and also `depth_draw_never` in the scene it either freezes the scene view or makes it really slow (depending on the complexity of the mesh I believe).

Haven't been able to reproduce this so I think there is more going on in this example. Could use an MRP for this

BastiaanOlij:
> Having both enabled at the same time causes the freeze on load. (Black screen frozen)

I'm able to reproduce this but it has me stumped. There is nothing in the error log, renderdoc just shows that we're skipping rendering all together. Android is such a PITA to debug.

When using glow we're rendering to an intermediate buffer. When using MSAA on devices that have this capability (Quest does, PC does not), we're using a special MSAA GLES extension that does MSAA in tile memory and then directly resolves and writes to texture. So we're using that here on the intermediate buffer. 
My guess is that this is working correctly with our normal framebuffer, but failing on the intermediate buffer.

drivers/gles3/shaders/effects/glow.glsl
BastiaanOlij: This is fixed, also fixed issue where I broke sky radiance map rendering due to these changes...

DanielSnd:
> > Found a new issue with it, if there's a shader that's has the render modes `cull_front` and also `depth_draw_never` in the scene it either freezes the scene view or makes it really slow (depending on the complexity of the mesh I believe).
> 
> Haven't been able to reproduce this so I think there is more going on in this example. Could use an MRP for this

Makes sense, I tested on the new push and it's still happening, MRP attached:
[ReproGlowIssue.zip](https://github.com/godotengine/godot/files/14313780/ReproGlowIssue.zip)

BastiaanOlij:
I found out a big problem with our code, and this is probably not even introduced by glow but it came to the surface, is our handling of global state in OpenGL. We often leave stuff as is after a process finishes with the assumption that the next process doesn't mind.

One such thing is leaving shaders bound, we are finding out that more and more functions that seemingly do nothing with shaders, can still fail if the current bound shader does not match the current framebuffer setup. In our case here I got a bunch of error spam from a `glClear` in the canvas renderer, because of the last shader used during 3D rendering.

The other issue had to do with setting and retaining the last culling state and the last depth buffer state, both causing issues with the execution of our glow shader causing the UI to seemingly freeze (it was just not processing the glow shader). 

I've thus did a LOT of cleanup on the way we handle global state in the 3D rendering part. There is more here to do but I've done the basics. I don't like having this logic where it is however, it should be more central. One thing @clayjohn and I have been discussing in the past is to rename our `config` class to an more general purpose `opengl(_wrapper)` class. This class is already a mix of OpenGL status info and additional function handling, moving this into the `drivers/gles3` folder and implementing it to hold and modify global state, and do stuff such as tracking resources, seems like a good idea. Out of scope for this PR however.

clayjohn (approved):
I think there are a few more improvements we can make to this. But seeing how long the process has been, I'm happy to merge this for now and then make the improvements in follow up PRs.

Particularly, we should make follow up PRs to:
1. Improve the luminance calculation. Using the current method means that blue objects don't glow
2. start with a 1/4 res downsample instead of 1/2 res. This will expand the size of the glow a bit, but it will also reduce the cost of glow.

BastiaanOlij:
> 1. Improve the luminance calculation. Using the current method means that blue objects don't glow

cc @Ansraer :)

> 2. start with a 1/4 res downsample instead of 1/2 res. This will expand the size of the glow a bit, but it will also reduce the cost of glow.

I think this should be a setting, and we'll need to see how we can modify the downsample and upsample as the logic from the original dual filtering spec is purely 1/2 res downstep/upstep. 
I also think that in this case we should have the final upstep go to a glow buffer, right now I'm doing the final upstep as part of writing into the render buffer to save an extra step. But I don't think we can efficiently do that especially when combined with scaling.

It would definitely be nice to do these as follow up PRs.

akien-mga:
Thanks! Amazing work that should make a lot of happy users :tada:

theromis:
@BastiaanOlij thank you for great improvement, trying to use it with my project, I have unshaded shaders:
```
shader_type spatial;
render_mode unshaded;

uniform float glow_power : hint_range(1.0, 100.0) = 30.0;
uniform vec3 color : source_color = vec3(0.0, 0.686, 0.757);

void fragment() {
    ALBEDO = color * glow_power;
}
```

simple and stupid, but it works with mobile and forward+, but doesn't work with current main build, is it possible to make it work or maybe some workaround?

BastiaanOlij:
@theromis worth raising an issue for so it's not overlooked. In theory this should work, I don't see anything that stands out in the scene shader that would limit the output.

theromis:
@BastiaanOlij finally found env setup to make this shader work, sorry, false alarm, palette looks different but at least it works.

Screenshot 2024-03-11 at 10 47 42

Issue #81287, opened 2023-09-03
`mouse_entered`/`mouse_exited` signals emitted for physics bodies even when hidden behind GUI overlay.


GlyphTheWolf:
### Godot version

v4.1.beta2.official [a2575cba4] and later up to 4.1.1 stable

### System information

Windows 10.0.19045 - Vulkan (Compatibility) - NVIDIA GeForce GTX 1050 (NVIDIA; 31.0.15.3713) - Intel(R) Core(TM) i5-6500 CPU @ 3.20GHz (4 Threads)

### Issue description

My project use signals `mouse_entered` and `mouse_exited` of `CollisionObject3D` to show some custom tooltips for 3D objects player may interact with.

The issue is that with latest Godot version behavior of that signals has changed and now they are triggered for my 3D scene objects even when user is hovering over GUI overlay (that is `Controls` with mouse filter set to `MOUSE_FILTER_STOP` for background, my GUI windows are just `Controls` not popups or real Godot windows)  and object is hidden behind. That cause tooltips for 3D objects are shown on my GUI windows what is of course not desired.

I think it is a bug/regression, because same thing was working fine on previous Godot versions and also I would expect old behavior based on input handling documentation which states that physics object picking is done as last step if not blocked by GUI (and some other possible processing) first.

From what I checked behavior has changed between:
v4.1.beta1.official [828ec2c5d] <- Here behavior is OK as I would expect
v4.1.beta2.official [a2575cba4] <- From now on behavior is WRONG

Here you can see issue in my project (run with v4.1.1.stable.official [bd6af8e0e]):

![tooltip_bug](https://github.com/godotengine/godot/assets/97635482/026c5666-8c35-4f24-bffa-efef9658f03b)

Tooltip for "World Map" is displayed since world map object hidden under GUI overlay received `mouse_entered` signal even I believe it shouldn't in this case.

### Steps to reproduce

In my minimal reproduction project there is red sphere in the middle as 3D object for picking and blue rectangle covering half of the viewport as overlay GUI. In the upper right corner there is debug label showing if the red sphere is hovered or not (simply based on `mouse_entered`/`mouse_exited` signals from `CollisionObject3D`).

Expected behaviour is that red spehere should be hovered only when mouse is outside of the blue rectangle (and hovering visible part of sphere of course).

[...]

PR #61088, opened 2022-05-16
Keep input event as unhandled if they go through a control set to MOU…

groud:
I had that issue where I wanted to be able to drag&drop things onto a game view. To do so, I tried put a control set to MOUSE_FILTER_PASS on top of my world, expecting that all unhandled mouse event would end up being processed by `_unhandled_event`, turned out it wasn't the case. Basically, as soon as an event is given to a Control somewhere, the event would be considered as handled if it went through a control with MOUSE_FILTER_PASS or MOUSE_FILTER_STOP.

However, I think it is not expected that an event which went through Controls set to MOUSE_FILTER_PASS ends up as handled if it wasn't explicitly set as handled.

As a consequence, what this PR does is simply set a mouse event as handled only if it it was stopped by a control. If the event reaches the root going only through controls set to MOUSE_FILTER_PASS or MOUSE_FILTER_IGNORE (and is not set as handled explicitly), then the event is now passed to the `_unhandled_event` processing.

<details>
  <summary>Obsolete edit</summary>
while investigating this issue, I ended up with other problems that I tried to fix too:

1) There was a mistake in this line of `_gui_call_input(...)` : 
```
bool ismouse = ev.is_valid() || Object::cast_to<InputEventMouseMotion>(*p_input) != nullptr;
``` 
~~The `ev.is_valid()` part was always true, which caused all events, even keyboard ones, to be stopped while climbing up the tree if the Control had the filter set to `MOUSE_FILTER_STOP`. This does not seemed like an intended behavior.~~

2) ~~Scroll wheel events were unstoppable. It was not much of an issue when those event were not forwarded to `_unhandled_input`, but with this PR, it became quite annoying. So I removed this special case (as ScrollContainer is already set to `MOUSE_FILTER_PASS` by default anyway). It is debatable whether or not this change is welcome, as it means that we would need all containers in a hierarchy to be set as to use the `MOUSE_FILTER_IGNORE` or `MOUSE_FILTER_PASS` modes so that the event can climb up the tree. Maybe an alternative would be to add another mode, set by default, that would allow those specific events to pass? Like `MOUSE_FILTER_STOP_BUT_SCROLLS` (bad name, I have to admit ^^)? 
</details>

Alright, I had to modify a little bit the PR:
[...]

Issue #55432, opened 2021-11-29
Control node "Pass" mousefilter is not passing input


golddotasksquestions:
### Godot version

3.4.1 rc1

### System information

Win64 Nvidea 765M

### Issue description

Control node "Pass" mouse filter behaves like "Stop".

The white Control has mouse_enter, mouse_exit and input_event signals connected. When the mouse hovers over it, it should turn green. However if there is a Control node above with a "Pass" mouse filter, it will stop the input, instead of passing it through.

https://user-images.githubusercontent.com/47016402/143825790-5d9b7234-adbf-472e-a22a-8c37b9c775e1.mp4

### Steps to reproduce

1. Add a Control Node and connect input mouse input signals such as mouse_enter, mouse_exit, input_event to a script.
2. In the script print something or change color when event occurs.
3. Add another Control node, overlapping the first and set it's mouse filter to "Pass"
4. Observe the top Control consuming the input and not passing it on despite being set to "Pass".

### Minimal reproduction project

[control_not_passing_input_issue.zip](https://github.com/godotengine/godot/files/7616299/control_not_passing_input_issue.zip)

KoBeWi:
We had similar issue recently #55288
This s not a bug, it's expected, documented behavior (try putting `ColorRect_pass` as child of `drop_target`). But apparently documenting it is not enough. I wonder if we can do more about it 🤔

golddotasksquestions:
>it's expected

Why would this be expected?

The purpose of the "Pass" mouse filter compared to "Stop" is to handle the input and pass it on. That's it's only purpose.

Why would it be expected behaviour to have "Pass" behave like "Stop" in any circumstance? That would not make any sense whatsoever. **We already have "Stop" for the stop behaviour!**

reduz:
The behavior is expected, the problem is that the naming of the mouse filter modes suck and the documentation should be better on what is the premise of how is works. We were debating changing the names for the filters and adding a new skip one that is more akin to what users expect in this situation.

golddotasksquestions:
As a user who does not understand this is intended behaviour, can anyone explain why it has been implemented like this?

@reduz it sounds like you want to keep this behaviour and solve this issue with renaming the flag. Would you mind to help me understand why this would be preferred over implementing the behaviour properly? (after your holiday is fine)

reduz:
As promised, here is a proposal that should shed more detail.
https://github.com/godotengine/godot-proposals/issues/3613

[...]

Issue #54529, opened 2021-11-03
Control consumes CollisionObject events for Area2D despite mouse_filter set to "pass"


golddotasksquestions:
### Godot version

3.3.4 stable, 3.4rc3, Godot v4.0.dev.20211015.official [f113dc986]

### System information

Win 64, Nvidea 765M

### Issue description

This one has cost me a lot of frustration and debugging time. Tested and confirmed in all current versions: 3.3.4, 3.4, 4.0.

If there is a Control overlapping an Area2D, the Control will consume the input event and mouse_entered and mouse_exited signal, despite mouse_event being set to "pass".

**This makes it impossible to use Area2D nodes and Control nodes in the same project when both overlap and need the mouse events**

https://user-images.githubusercontent.com/47016402/139970970-37d8ca75-53e1-4302-9a17-7787a740ded8.mp4

And  mouse_entered and mouse_exited signals of the Area2D won't fire either when the Area2D is overlapping with the Control:

https://user-images.githubusercontent.com/47016402/139999079-ddfc4d05-cc7e-422e-ac79-401143ad4eb5.mp4

### Steps to reproduce

1. Add a Control node, set it's mouse_filter to "pass" and connect the gui_input(event) signal to a script
2. Add a Area2D node and a CollisonShape2D, and connect the input_event(event) signal to a script
3. Make sure the Area2D is overlapping the Control node
4. Observe the Area input_event(event) signal not firing any more

### Minimal reproduction project

Godot 3.X minimal project: 
[area_over_control_bug_godot3.zip](https://github.com/godotengine/godot/files/7464746/area_over_control_bug_godot3.zip)

Godot 4.0 minimal project:
[area_over_control_bug_godot4.zip](https://github.com/godotengine/godot/files/7464745/area_over_control_bug_godot4.zip)

BimDav:
I think that the first case respects the documentation of MOUSE_FILTER_PASS: "Even if no control handled it at all, the event will still be handled automatically, so unhandled input will not be fired." Since the event is handled, it will not collide with Areas.

The second case (mouse_entered and mouse_exited signals) is probably due to the same reason, but documentation should explain that events must not be handled for these signals to be emitted.

That being said, maybe the event should not be handled automatically in the Control, since your usercase seems common.

golddotasksquestions:
Yes, but it should not be handled! That's the whole point of this issue.

BimDav:
[...]

PR #87362, opened 2024-01-19
Prevent `PhysicsDirectSpaceState3D::instersect_ray` when mouse is over non-ignorable gui

### LINKED ISSUE #87322: Mouse Stop Filter unexpectedly triggers Area3D's mouse_enter signal after triggering expected mouse_exit signal
### Tested versions

- Reproducible in v4.2.1.stable.official [b09f793f5]

### System information

Godot v4.2.1.stable - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4060 (NVIDIA; 31.0.15.3734) - Intel(R) Core(TM) i5-14600K (20 Threads)

## Overview
The issue is that an `Area3D`'s `mouse_enter` signal is unexpectedly emitting immediately after its `mouse_exit` signal emits.  I don't think this is the expected behavior because in my **MRP** I can reproduce this only seconds after demonstrating that it does not always occur.

#### Setup
The **MRP** has a `MeshInstance` (a sphere with radius=1) and a `Button` in the center of the screen so that it partially overlaps the `MeshInstance`.  The `Button`'s mouse filter is set to **stop**, and the `MeshInstance` has an `Area3D` (with a collision body that matches the mesh) which triggers a print statement when the mouse enters or exits the `Area3D`.
#### Issue
What I expect to see, and what happens some of the time, is that when the mouse hovers over the `Area3D` the scene should print "MOUSE: ENTER", and when the mouse continues to move so that is hovers over the button (which overlaps the `Area3D`) the scene should print "MOUSE: EXIT".  In some case, however, there is an additional (and unexpected) "MOUSE: ENTER" that gets printed right after the expected "MOUSE: EXIT"

### Steps to reproduce

Using the MRP:
1. Launch the app
2. Hover over the sphere
    - See that the console prints "MOUSE ENTER" 
3. Hover off of the sphere (not going over the button)
    - See that the console prints "MOUSE EXIT"
4.  Hover over the sphere, but quickly continue to move the mouse so you also hover over the button.
    - This step doesn't always produce the unexpected behavior.  I don't know exactly what causes the unexpected behavior, but if you keep trying I think you'll be able to see it (see the attached recording)

**Observed behavior:**
After step 4, I sometimes see 3 new print statements:
```
MOUSE: ENTER
MOUSE: EXIT
MOUSE: ENTER
```

**Expected behavior:**
I expect to always see only 2 new print statements
```
MOUSE: ENTER
MOUSE: EXIT
```

### Screen Recording (with console print statements)

https://github.com/godotengine/godot/assets/8028662/156120e9-372f-49fc-a2a6-75844a009ac1

[minimal_mouse_hove_example.zip](https://github.com/godotengine/godot/files/13971902/minimal_mouse_hove_example.zip)

stokatyan:
## Overview
This PR is intended to fix https://github.com/godotengine/godot/issues/87322, where a non-ignorable gui element was not blocking a `CollisionShape3D`'s `_mouse_enter` from being emitted.

## Issue
The cause of this bug is that `_drop_physics_mouseover` (in viewport.cpp) would get called when the mouse would move over gui, and that would cause a `_mouse_exit` to occur, and it would also result in the following if statement condition to be true: `new_collider != physics_object_over` because the `physics_object_over` would be null.  That shouldn't be an issue, except that it is possible for `_process_picking` to be invoked on the same exact frame as the previous `_drop_physics_mouseover`.

I'm not 100% certain about the last sentence above, but that was my best understanding based on my debugging.  I look forward to hearing from others that understand the engine better than I do.

## Fix
I think the fix is to only check for 3D objects in `process_picking` if `!gui.mouse_over`.  This does not break the Stop, Ignore, and Pass settings (as far as I can tell), but I have only tested on the minimum example project attached to this PR, and in the MRPs created based on comments from others.

## Minimum Fix Project
[minimal_mouse_hove_example.zip](https://github.com/godotengine/godot/files/13986723/minimal_mouse_hove_example.zip)
 
## Screen Recording of Fix

https://github.com/godotengine/godot/assets/8028662/1376f503-4ac5-4221-86e1-830cd44fbc28

When the mouse hovers over the 3D shape (it looks 2D because there is no lighting), the shape scales up use _mouse_enter.  When the mouse hovers off the 3D shape it scales down to its original scale.  I can explain the logs further if people have questions, but I want to keep this description short so I will hold off on that.

### DIFF
--- a/scene/main/viewport.cpp
+++ b/scene/main/viewport.cpp
@@ -881,7 +881,7 @@ void Viewport::_process_picking() {
 					}
 				}
 			}
-		} else {
+		} else if (!gui.mouse_over || gui.mouse_over->get_mouse_filter() != Control::MOUSE_FILTER_STOP) {
 			if (camera_3d) {
 				Vector3 from = camera_3d->project_ray_origin(pos);
 				Vector3 dir = camera_3d->project_ray_normal(pos);
@@ -3172,6 +3172,7 @@ void Viewport::_update_mouse_over(Vector2 p_pos) {
 		// Find the common ancestor of `gui.mouse_over` and `over`.
 		Control *common_ancestor = nullptr;
 		LocalVector<Control *> over_ancestors;
+		bool ancestor_is_stop_filter = false;
 
 		if (over) {
 			// Get all ancestors that the mouse is currently over and need an enter signal.
@@ -3189,6 +3190,7 @@ void Viewport::_update_mouse_over(Vector2 p_pos) {
 					}
 					if (ancestor_control->get_mouse_filter() == Control::MOUSE_FILTER_STOP) {
 						// MOUSE_FILTER_STOP breaks the propagation chain.
+						ancestor_is_stop_filter = true;
 						break;
 					}
 				}
@@ -3203,7 +3205,7 @@ void Viewport::_update_mouse_over(Vector2 p_pos) {
 		if (gui.mouse_over || !gui.mouse_over_hierarchy.is_empty()) {
 			// Send Mouse Exit Self and Mouse Exit notifications.
 			_drop_mouse_over(common_ancestor);
-		} else {
+		} else if (ancestor_is_stop_filter) {
 			_drop_physics_mouseover();
 		}
 


scene/main/viewport.cpp
```diff
@@ -881,7 +881,7 @@ void Viewport::_process_picking() {
 					}
 				}
 			}
-		} else {
+		} else if (!gui.mouse_over) {
```
Sauermann, line 884: In #61088 the following behavior got introduced:
- When an `InputEventMouseMotion` is over a Control node and the Control node doesn't mark the event as handled, then the input event can be used for physics picking. See

https://github.com/godotengine/godot/blob/7827c8e37010241b376748b22db85847c8614dbd/scene/main/viewport.cpp#L2001-L2004

With the changes of this PR in its current form, physics picking will no longer work for these input events in 3D, because for these input events, `gui.mouse_over == true`.

stokatyan: Thanks for the explanation, I understand and I will take another look.  Also, I will not include the .godot folder going forward.

stokatyan:
@Sauermann I am moving this out of draft mode.  My latest commit fixes the bug I reported, and the [issue ](https://github.com/godotengine/godot/pull/87362#discussion_r1459445070) you found.  Please let me know what you think.

## Latest Fix
2c6ba092265c9a1b0f5d7adac9dcaa575e33aed7 Fixes the issues by blocking `PhysicsDirectSpaceState3D::instersect_ray` when the mouse is over a `Control` with mouse filter set to `MOUSE_FILTER_STOP`.  

This commit also prevents an unneeded `_mouse_exit` from occurring when initially entering a `Control` with mouse a filter set to `MOUSE_FILTER_PASS` (by only allowing `_drop_physics_mouseover` to be called when `gui.mouse_over`'s filter is `MOUSE_FILTER_STOP`).  The unneeded `_mouse_exit` was triggering an extra `_mouse_enter` -- similar to the original bug.

## Latest MRP
[minimal_mouse_hover_example_02.zip](https://github.com/godotengine/godot/files/13996476/minimal_mouse_hover_example_02.zip) (no .godot folder)

## Screen Recording

https://github.com/godotengine/godot/assets/8028662/3224c5e2-3b6e-4de5-ab2a-c8a2e7203df2

The screen recording shows that there are no extra calls to the 3D collision object's mouse enter/exit, and the PASS-Control does not prevent print statements from being executed when mouse is moved over it.

## Question
I noticed that the print statement for mouse movement still get printed even over the section of the PASS-Control that overlaps the STOP-Control (~9s into the screen recording).  That might be another bug, but if it is, I think it is separate from this PR, and might be less of a big deal than the current bug since devs can make easier workarounds for it.  

My question is: does the engine keep track of every Control hierarchy that it is over, or does it only keep track of the top-most Control hierarchy?

kitbdev:
related: 
- https://github.com/godotengine/godot/issues/54529

There is an issue when there are multiple Controls in the mouse over hierarchy and the top one is STOP.
In this Gif:
- ta0 is the red one and has mouse filter STOP
- ta1 is the child of ta0 and has mouse filter PASS
godot mouse over issue 87362
When entering the PASS Control from the Area2D, it keeps mouse over and does not drop it when the mouse is no longer overlapping it. When entering the Area2D from the PASS Control, it does not receive mouse over at all.
Area2D and Area3D are the same here.

> does the engine keep track of every Control hierarchy that it is over, or does it only keep track of the top-most Control hierarchy?

Yes, it only keeps track of the the top most since it is intended that the mouse can only be over one chain of Control nodes at a time.

The mouse filter MOUSE_FILTER_PASS actually sends events up to its parent with event bubbling, it is not intended to let events behind it.
See these issues for more about it:
- https://github.com/godotengine/godot-proposals/issues/3613
- https://github.com/godotengine/godot/issues/55432
- https://github.com/godotengine/godot-proposals/issues/7167
- https://github.com/godotengine/godot-proposals/issues/1775

Since PASS sends events up and not behind, I feel that PASS and STOP should work the same in this case and not let mouse events through to the physics picking at all, but others have different opinions about it (#54529).

stokatyan:
@kitbdev I'm not sure I completely understand your comment.  I see the issue, but are you suggesting that this PR is the cause of the issue you linked?  Is there something I should change in my PR or are you just pointing out that there is a separate, but possibly related, issue?

kitbdev:
The linked issues are only related.

The issue in the Gif is a result of this PR, since you changed the behavior (`_drop_physics_mouseover()` affects 2d and 3d)
Not that the original behavior was completely functional, since the original issue you fixed also exists in 2D as well.

stokatyan:
@kitbdev can you share the MRP of the gif you shared?  I see the issue in your gif, but I'd like to be sure I test against the same setup.

I think I know the issue -- my changes don't take into account the gui.mouse_over_hierarchy.  I will take another look.

kitbdev:
Sure, here's the MRP:
[mouseover_87362_2D.zip](https://github.com/godotengine/godot/files/13999431/mouseover_87362_2D.zip)

stokatyan:
@kitbdev https://github.com/godotengine/godot/pull/87362/commits/d6efa634ee4fbf17bd149f8ad479120369a5c2c9 fixes the issue exposed by your MRP, and fixes the issue exposed in my MRP.

I did not expect it, but on `master` a PASS-Control that is a child of a STOP-Control must result in an invocation of `_drop_physics_mouseover` even if the portion of the PASS-Control that mouse is over is not overlapping the STOP-Control.  This commit keeps that behavior by checking if any ancestor of the PASS-Control is a STOP-Control.

scene/main/viewport.cpp
```diff
@@ -3172,6 +3172,7 @@ void Viewport::_update_mouse_over(Vector2 p_pos) {
 		// Find the common ancestor of `gui.mouse_over` and `over`.
 		Control *common_ancestor = nullptr;
 		LocalVector<Control *> over_ancestors;
+		bool ancenstor_is_stop_filter = false;
```
AThousandShips, line 3175:
```suggestion
		bool ancestor_is_stop_filter = false;
```

stokatyan: nice catch, fixed: 3fc735f0b80da9cbe640c2c3112262523d21d7d3

mournguard:
This issue is breaking a lot of my interactions and it took me a while to figure out what was happening, I basically couldn't believe the events could be **this** wrong and thought it was my code for hours.

I think there's many more problems with this, all these issues seem related: #81287, #108240, #104724, #101903, and maybe more.

Many of these are about specific nodes or cases, but it might be a bit more general ; The original issue was about `Area3D`s, but I stumbled on that problem with a `StaticBody3D` on the latest 4.5.rc1.

Nevertheless, if this PR still fixes it, even if only for 3d nodes, that's a still a good thing to merge in considering how glaring the problem is.

SpruceBruce64:
@stokatyan If you're still interested in this PR I tested this fix by taking your commit and applying it over the latest master and found an edge case where if your control node tree has a margin container with the mouse filtering set to pass the original issue persists. If you change the filter to stop then the fix works. I think there's more cases like this but I wasn't able to recreate it often enough to pin point the cause. 

I made a minimum project to demonstrate it with a margin container so you can test out changing the mouse filter
[bug-test.zip](https://github.com/user-attachments/files/22758539/bug-test.zip)

stokatyan:
@SpruceBruce64 I see your comment, but won't be able to get back into this PR until later next week.  Thanks for sharing your comment, I will share an update soon.

stokatyan:
@SpruceBruce64 is the edge case introduced by this PR, or is it an existing issue?  If it is an existing issue I think it is better to merge this instead of expanding the scope of this PR.

PR #87371, opened 2024-01-19
Replace error checks against `size` with `is_empty`

AThousandShips:
Replacing just the `size` -> `is_empty` here because those are the cases I feel improves readability, also fixes some unusual cases like `< 1` and `<= 0`


akien-mga (approved):

AThousandShips:
For reference these were searched for with `ERR_FAIL_COND(_V)?(_MSG)?\(\w+\.size\(\) (== 0|< 1)` and ``ERR_FAIL_COND(_V)?(_MSG)?\(!\w+\.size\(\)` as far as I recall, and verified these were all when rebasing now, so should be complete

akien-mga:
Thanks!

AThousandShips:
Thank *you*!

PR #87373, opened 2024-01-19
Add `--log-file` command line argument to write output log to a file

### LINKED ISSUE #8803: Ability to write godot console output to log file

Creating a systemd service to run dedicated server

Not getting any log output neither `syslog` nor `file` or `append` type

https://github.com/godotengine/godot/issues/86785

Implementing --log /path/to/log.file in need of file logging states we working with file rather than redirect detection.
For logrotate important to handle SIGHUP to reopen log file.

I think as a future improvement same as predefined --headless parameter might be a good idea to have:

* --background (fork after start)
* --pid /path/to/pid.file (store process ID to be able to use `Type=forked` in systemd unit file it will help to restart process if it dies)

By specifying logging ability we must flush on every line of code

Unfortunately currently impossible to run godot in background and get to know what heppening

This is pretty basic feature and it needed for dedicated servers

Calinou:
This allows using a custom output path for the log file, with log rotation disabled. This works even if file logging is disabled in the project settings, or for the editor/project manager.

`--log-file`'s value can be an absolute path or relative to the project directory (similar to existing arguments like `--write-movie`).

Note that flushing stdout on print isn't forced in release builds (it still depends on the project setting's value). We may need a way to enforce this, but we already have a lot of CLI arguments exposed.

- This closes https://github.com/godotengine/godot-proposals/issues/8803.

### DIFF
--- a/doc/classes/ProjectSettings.xml
+++ b/doc/classes/ProjectSettings.xml
@@ -433,16 +433,18 @@
 			If canvas item redraw debugging is active, this will be the time the flash will last each time they redraw.
 		</member>
 		<member name="debug/file_logging/enable_file_logging" type="bool" setter="" getter="" default="false">
-			If [code]true[/code], logs all output to files.
+			If [code]true[/code], logs all output and error messages to files. See also [member debug/file_logging/log_path], [member debug/file_logging/max_log_files], and [member application/run/flush_stdout_on_print].
 		</member>
 		<member name="debug/file_logging/enable_file_logging.pc" type="bool" setter="" getter="" default="true">
 			Desktop override for [member debug/file_logging/enable_file_logging], as log files are not readily accessible on mobile/Web platforms.
 		</member>
 		<member name="debug/file_logging/log_path" type="String" setter="" getter="" default="&quot;user://logs/godot.log&quot;">
 			Path at which to store log files for the project. Using a path under [code]user://[/code] is recommended.
+			This can be specified manually on the command line using the [code]--log-file &lt;file&gt;[/code] [url=$DOCS_URL/tutorials/editor/command_line_tutorial.html]command line argument[/url]. If this command line argument is specified, log rotation is automatically disabled (see [member debug/file_logging/max_log_files]).
 		</member>
 		<member name="debug/file_logging/max_log_files" type="int" setter="" getter="" default="5">
-			Specifies the maximum number of log files allowed (used for rotation).
+			Specifies the maximum number of log files allowed (used for rotation). Set to [code]1[/code] to disable log file rotation.
+			If the [code]--log-file &lt;file&gt;[/code] [url=$DOCS_URL/tutorials/editor/command_line_tutorial.html]command line argument[/url] is used, log rotation is always disabled.
 		</member>
 		<member name="debug/gdscript/warnings/assert_always_false" type="int" setter="" getter="" default="1">
 			When set to [code]warn[/code] or [code]error[/code], produces a warning or an error respectively when an [code]assert[/code] call always evaluates to false.
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -172,6 +172,7 @@ static bool editor = false;
 static bool project_manager = false;
 static bool cmdline_tool = false;
 static String locale;
+static String log_file;
 static bool show_help = false;
 static uint64_t quit_after = 0;
 static OS::ProcessID editor_pid = 0;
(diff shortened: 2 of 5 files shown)


main/main.cpp
```diff
@@ -1689,12 +1701,22 @@ Error Main::setup(const char *execpath, int argc, char *argv[], bool p_second_ph
 	GLOBAL_DEF("debug/file_logging/log_path", "user://logs/godot.log");
 	GLOBAL_DEF(PropertyInfo(Variant::INT, "debug/file_logging/max_log_files", PROPERTY_HINT_RANGE, "0,20,1,or_greater"), 5);
 
-	if (!project_manager && !editor && FileAccess::get_create_func(FileAccess::ACCESS_USERDATA) &&
-			GLOBAL_GET("debug/file_logging/enable_file_logging")) {
+	// If `--log-file` is used to override the log path, allow creating logs for the project manager or editor
+	// and even if file logging is disabled in the Project Settings.
+	if (FileAccess::get_create_func(FileAccess::ACCESS_USERDATA) &&
+			(!log_file.is_empty() || (!project_manager && !editor && GLOBAL_GET("debug/file_logging/enable_file_logging")))) {
```
akien-mga, line 1707: `--log-file` typically wouldn't be used to write to `user://`, but to a pass relative to the working directory, so I'm not sure the check for `ACCESS_USERDATA` makes sense / is enough.

misc/dist/shell/godot.fish
```diff
@@ -67,6 +67,7 @@ complete -c godot -l gpu-index -d "Use a specific GPU (run with --verbose to get
 complete -c godot -l text-driver -d "Set the text driver" -x
 complete -c godot -l tablet-driver -d "Set the pen tablet input driver" -x
 complete -c godot -l headless -d "Enable headless mode (--display-driver headless --audio-driver Dummy). Useful for servers and with --script"
+complete -c godot -l log-file -d "Write output/error log to the specified path instead of the default location defined by the project" -x
 complete -c godot -l write-movie -d "Writes a video to the specified path (usually with .avi or .png extension). --fixed-fps is forced when enabled" -x
```
akien-mga, line 71: To be consistent with other changes.
```suggestion
complete -c godot -l write-movie -d "Write a video to the specified path (usually with .avi or .png extension). --fixed-fps is forced when enabled" -x
```

akien-mga (approved):
Sounds like a good option to have, especially for when users struggling getting actual console output to debug issues on Windows. Would be worth checking to confirm that it works well for both the `.console.exe` wrapper and the normal `.exe`.

doc/classes/ProjectSettings.xml
```diff
@@ -433,16 +433,18 @@
 			If canvas item redraw debugging is active, this will be the time the flash will last each time they redraw.
 		</member>
 		<member name="debug/file_logging/enable_file_logging" type="bool" setter="" getter="" default="false">
-			If [code]true[/code], logs all output to files.
+			If [code]true[/code], logs all output and error messages to files. See also [member debug/file_logging/log_path], [member debug/file_logging/max_log_files] and [member application/run/flush_stdout_on_print].
```
AThousandShips, line 436:
```suggestion
			If [code]true[/code], logs all output and error messages to files. See also [member debug/file_logging/log_path], [member debug/file_logging/max_log_files], and [member application/run/flush_stdout_on_print].
```

AThousandShips (approved):

main/main.cpp
Calinou: I've tested an absolute path located outside `user://` and it worked fine, but I guess we can check for `ACCESS_FILESYSTEM` when the log path isn't empty.

YuriSizov (approved):
Seems to work fine on Windows with either `console.exe` or regular `exe`.

YuriSizov:
Thanks!

theromis:
@Calinou thank you for quick development, is it good idea to create different proposal for optional running godot in background, reopening log file on signal and having ability to have a PID file to run it as a service. Or better just leave it as is?

Calinou:
> @Calinou thank you for quick development, is it good idea to create different proposal for optional running godot in background, reopening log file on signal and having ability to have a PID file to run it as a service. Or better just leave it as is?

Yes, each of these items should be tracked in its *own* proposal. Please don't cram several unrelated items into a single proposal, as this makes it difficult to track them individually.

theromis:
Thank you, created
https://github.com/godotengine/godot-proposals/issues/8957
https://github.com/godotengine/godot-proposals/issues/8959
https://github.com/godotengine/godot-proposals/issues/8960

PR #87376, opened 2024-01-19
SCons: Add `num_jobs` as an explicit option so it can be set from other sources

shana:
The `-j` flag is only settable via the command line, which makes it hard to configure when running builds from places like VS, where the flag isn't easily exposed or configurable.

This adds a `num_jobs` option, which lets users configure the number of jobs to be used by default (in a custom.py file or other similar mechanism) if `-j` isn't specified, instead of always defaulting to number of cores - 1. 

*Contributed by W4 Games :heart:*

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -234,6 +234,11 @@ opts.Add(BoolVariable("tests", "Build the unit tests", False))
 opts.Add(BoolVariable("fast_unsafe", "Enable unsafe options for faster rebuilds", False))
 opts.Add(BoolVariable("ninja", "Use the ninja backend for faster rebuilds", False))
 opts.Add(BoolVariable("compiledb", "Generate compilation DB (`compile_commands.json`) for external tools", False))
+opts.Add(
+    "num_jobs",
+    "Use up to N jobs when compiling (equivalent to `-j N`). Defaults to max jobs - 1. Ignored if -j is used.",
+    "",
+)
 opts.Add(BoolVariable("verbose", "Enable verbose output for the compilation", False))
 opts.Add(BoolVariable("progress", "Show a progress indicator during compilation", True))
 opts.Add(EnumVariable("warnings", "Level of compilation warnings", "all", ("extra", "all", "moderate", "no")))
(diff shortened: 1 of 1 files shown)


shana:
Fixed a small issue found after more testing, this is ready to go 👍

SConstruct
```diff
@@ -258,6 +258,7 @@ opts.Add(BoolVariable("scu_build", "Use single compilation unit build", False))
 opts.Add("scu_limit", "Max includes per SCU file when using scu_build (determines RAM use)", "0")
 opts.Add(BoolVariable("engine_update_check", "Enable engine update checks in the Project Manager", True))
 opts.Add(BoolVariable("steamapi", "Enable minimal SteamAPI integration for usage time tracking (editor only)", False))
+opts.Add("num_jobs", "Allow N jobs at once", "")
```
akien-mga, line 261:
```suggestion
opts.Add("num_jobs", "Use up to N jobs when compiling (equivalent to `-j N`). Defaults to max jobs - 1.", "")
```

And I'd move it further up, e.g. between `compiledb` and `verbose`. Nitpicky, but semantically it's more related to these early options than later ones like `steamapi` or `engine_update_check` which impact the actual code being compiled.

SConstruct
```diff
@@ -537,16 +538,22 @@ initial_num_jobs = env.GetOption("num_jobs")
...
-        env.SetOption("num_jobs", safer_cpu_count)
+        cpu_count = os.cpu_count()
+        if cpu_count is None:
+            print_warning(
+                "Couldn't auto-detect CPU count to configure build parallelism. Specify it with the -j argument."
+            )
+        else:
+            safer_cpu_count = cpu_count if cpu_count <= 4 else cpu_count - 1
+            print(
+                "Auto-detected %d CPU cores available for build parallelism. Using %d cores by default. You can override it with the -j argument."
```
akien-mga, line 553:
```suggestion
                "Couldn't auto-detect CPU count to configure build parallelism. Specify it with the `-j` or `num_jobs` arguments."
            )
        else:
            safer_cpu_count = cpu_count if cpu_count <= 4 else cpu_count - 1
            print(
                "Auto-detected %d CPU cores available for build parallelism. Using %d cores by default. You can override it with the `-j` or `num_jobs` arguments."
```

SConstruct
shana: Gotcha! I think I've been moving it down over and over as more vars have been added, haha. I'll fix that up asap.

akien-mga (approved):
Tested, it seems to work well.

Worth noting that if both `num_jobs` and `-j` are used, `-j` takes precedence, regardless of the order.
`num_jobs` in `custom.py` seems to work fine too.

shana:
> Tested, it seems to work well.
> 
> Worth noting that if both `num_jobs` and `-j` are used, `-j` takes precedence, regardless of the order. `num_jobs` in `custom.py` seems to work fine too.

That's a good point. Where should that note be? Maybe

`Use up to N jobs when compiling (equivalent to -j N). Defaults to max jobs - 1. Ignored if -j is specified.`

?

shana:
I tweaked the message to note the precedent of `-j`.

Calinou (approved):
Tested locally, it works as expected. Code looks good to me.

akien-mga:
Thanks!

PR #87378, opened 2024-01-19
Add NavigationMesh `border_size` property for tile baking

smix8:
Adds NavigationMesh `border_size` property for tile baking.

This allows users to partition larger game worlds into regions that have navigation meshes with easy to align tile edges. This tile edges can be efficiently merged by the navigation map using edgekeys. It also avoids problems with the agent_radius offset affecting the tile edges.

The difference between baking AABB and border size is, that the AABB limits the entire source geometry used in the baking as a bounding area while the border size limits the surface of the resulting navigation mesh.

This means in conjunction they can be used to limit the final navigation mesh without having unwanted offsets by e.g. agent radius at tile edges.

nav_mesh_border_size

When users do the setup the border_size should be larger than the agent_radius and the AABB should be grown by the exact amount of the border_size.

The border size that ReCast uses internally is in voxel units, so the Godot border size float is ceiled to the cell size.

To bake a "perfect" aligned tile navigation mesh additional properties need to be considered. E.g. the edge_max_error should be at or blow 1.0.

### DIFF
--- a/doc/classes/NavigationMesh.xml
+++ b/doc/classes/NavigationMesh.xml
@@ -96,6 +96,11 @@
 			The distance to erode/shrink the walkable area of the heightfield away from obstructions.
 			[b]Note:[/b] While baking, this value will be rounded up to the nearest multiple of [member cell_size].
 		</member>
+		<member name="border_size" type="float" setter="set_border_size" getter="get_border_size" default="0.0">
+			The size of the non-navigable border around the bake bounding area.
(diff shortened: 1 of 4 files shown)


akien-mga:
We merged the 2D counterpart, is this one also good to go? CC @Scony

Scony (approved):

Scony:
All good @akien-mga

AThousandShips (approved):
Style and docs look good

akien-mga:
Thanks!

PR #87379, opened 2024-01-19
Change TileMapEditor to TileMapLayerEditor

groud:
This is the continuation of my work on moving TileMap layers to individual nodes.

This PR does several things:
- Implement TileMapGroup class, that holds the TileSet and a few editor things for TileMapLayers
- Replace the TileMapEditor by a TileMapLayerEditor, basically changing the logic to act at the layer-level instead of the TileMap one.

Things should work more or less the same for now. The only difference should be in the display of the grid, as it will consider only the currently selected layer instead of the whole TileMap.

### DIFF
--- a/doc/classes/TileMapLayerGroup.xml
+++ b/doc/classes/TileMapLayerGroup.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="TileMapLayerGroup" inherits="Node2D" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
+	<brief_description>
+		Groups a set of tile map layers together, allowing them to share a provided [TileSet].
+	</brief_description>
+	<description>
+		Groups together tile map layers as part or the same map, replacing the [TileMap] node. Child layers will use this node's [member tile_set].
+		The editor also uses [TileMapLayerGroup] as a way to store which layers are selected in a given group. This allows highlighting the currently selected layers.
+	</description>
+	<tutorials>
+	</tutorials>
+	<members>
+		<member name="tile_set" type="TileSet" setter="set_tileset" getter="get_tileset">
+			The assigned [TileSet]. This TileSet will be applied to all child layers.
+		</member>
+	</members>
+</class>
--- a/editor/plugins/tiles/tile_map_layer_editor.cpp
+++ b/editor/plugins/tiles/tile_map_layer_editor.cpp
@@ -1,5 +1,5 @@
 /**************************************************************************/
-/*  tile_map_editor.cpp                                                   */
+/*  tile_map_layer_editor.cpp                                             */
 /**************************************************************************/
 /*                         This file is part of:                          */
 /*                             GODOT ENGINE                               */
@@ -28,7 +28,7 @@
 /* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
 /**************************************************************************/
 
-#include "tile_map_editor.h"
+#include "tile_map_layer_editor.h"
 
 #include "tiles_editor_plugin.h"
 
@@ -40,6 +40,7 @@
 #include "editor/themes/editor_scale.h"
 
 #include "scene/2d/camera_2d.h"
+#include "scene/2d/tile_map_layer.h"
 #include "scene/gui/center_container.h"
 #include "scene/gui/split_container.h"
 
@@ -47,22 +48,26 @@
 #include "core/math/geometry_2d.h"
 #include "core/os/keyboard.h"
 
-void TileMapEditorTilesPlugin::tile_set_changed() {
+TileMapLayer *TileMapLayerSubEditorPlugin::_get_edited_layer() const {
+	return Object::cast_to<TileMapLayer>(ObjectDB::get_instance(edited_tile_map_layer_id));
+}
+
+void TileMapLayerEditorTilesPlugin::tile_set_changed() {
 	_update_fix_selected_and_hovered();
 	_update_tile_set_sources_list();
 	_update_source_display();
 	_update_patterns_list();
 }
 
-void TileMapEditorTilesPlugin::_on_random_tile_checkbox_toggled(bool p_pressed) {
+void TileMapLayerEditorTilesPlugin::_on_random_tile_checkbox_toggled(bool p_pressed) {
 	scatter_controls_container->set_visible(p_pressed);
 }
 
-void TileMapEditorTilesPlugin::_on_scattering_spinbox_changed(double p_value) {
+void TileMapLayerEditorTilesPlugin::_on_scattering_spinbox_changed(double p_value) {
 	scattering = p_value;
 }
 
-void TileMapEditorTilesPlugin::_update_toolbar() {
+void TileMapLayerEditorTilesPlugin::_update_toolbar() {
 	// Stop draggig if needed.
 	_stop_dragging();
 
@@ -96,13 +101,13 @@ void TileMapEditorTilesPlugin::_update_toolbar() {
 	}
 }
 
-void TileMapEditorTilesPlugin::_update_transform_buttons() {
-	TileMap *tile_map = Object::cast_to<TileMap>(ObjectDB::get_instance(tile_map_id));
-	if (!tile_map) {
+void TileMapLayerEditorTilesPlugin::_update_transform_buttons() {
+	TileMapLayer *edited_layer = _get_edited_layer();
+	if (!edited_layer) {
 		return;
 	}
 
-	Ref<TileSet> tile_set = tile_map->get_tileset();
+	Ref<TileSet> tile_set = edited_layer->get_effective_tile_set();
 	if (tile_set.is_null() || selection_pattern.is_null()) {
 		return;
 	}
@@ -126,7 +131,7 @@ void TileMapEditorTilesPlugin::_update_transform_buttons() {
 	}
 }
 
-void TileMapEditorTilesPlugin::_set_transform_buttons_state(const Vector<Button *> &p_enabled_buttons, const Vector<Button *> &p_disabled_buttons, const String &p_why_disabled) {
+void TileMapLayerEditorTilesPlugin::_set_transform_buttons_state(const Vector<Button *> &p_enabled_buttons, const Vector<Button *> &p_disabled_buttons, const String &p_why_disabled) {
 	for (Button *button : p_enabled_buttons) {
 		button->set_disabled(false);
 		button->set_tooltip_text("");
@@ -137,14 +142,14 @@ void TileMapEditorTilesPlugin::_set_transform_buttons_state(const Vector<Button 
 	}
 }
 
-Vector<TileMapSubEditorPlugin::TabData> TileMapEditorTilesPlugin::get_tabs() const {
-	Vector<TileMapSubEditorPlugin::TabData> tabs;
+Vector<TileMapLayerSubEditorPlugin::TabData> TileMapLayerEditorTilesPlugin::get_tabs() const {
+	Vector<TileMapLayerSubEditorPlugin::TabData> tabs;
 	tabs.push_back({ toolbar, tiles_bottom_panel });
 	tabs.push_back({ toolbar, patterns_bottom_panel });
 	return tabs;
 }
 
-void TileMapEditorTilesPlugin::_tab_changed() {
+void TileMapLayerEditorTilesPlugin::_tab_changed() {
 	if (tiles_bottom_panel->is_visible_in_tree()) {
 		_update_selection_pattern_from_tileset_tiles_selection();
 	} else if (patterns_bottom_panel->is_visible_in_tree()) {
@@ -152,7 +157,7 @@ void TileMapEditorTilesPlugin::_tab_changed() {
 	}
 }
 
-void TileMapEditorTilesPlugin::_update_tile_set_sources_list() {
+void TileMapLayerEditorTilesPlugin::_update_tile_set_sources_list() {
 	// Update the sources.
 	int old_current = sources_list->get_current();
 	int old_source = -1;
@@ -161,13 +166,13 @@ void TileMapEditorTilesPlugin::_update_tile_set_sources_list() {
 	}
 	sources_list->clear();
 
-	TileMap *tile_map = Object::cast_to<TileMap>(ObjectDB::get_instance(tile_map_id));
-	if (!tile_map) {
+	TileMapLayer *edited_layer = _get_edited_layer();
+	if (!edited_layer) {
 		return;
 	}
 
-	Ref<TileSet> tile_set = tile_map->get_tileset();
-	if (!tile_set.is_valid()) {
+	Ref<TileSet> tile_set = edited_layer->get_effective_tile_set();
+	if (tile_set.is_null()) {
 		return;
 	}
 
@@ -244,15 +249,15 @@ void TileMapEditorTilesPlugin::_update_tile_set_sources_list() {
 	TilesEditorUtils::get_singleton()->set_sources_lists_current(sources_list->get_current());
 }
 
-void TileMapEditorTilesPlugin::_update_source_display() {
+void TileMapLayerEditorTilesPlugin::_update_source_display() {
 	// Update the atlas display.
-	TileMap *tile_map = Object::cast_to<TileMap>(ObjectDB::get_instance(tile_map_id));
-	if (!tile_map) {
+	TileMapLayer *edited_layer = _get_edited_layer();
+	if (!edited_layer) {
 		return;
 	}
 
-	Ref<TileSet> tile_set = tile_map->get_tileset();
-	if (!tile_set.is_valid()) {
+	Ref<TileSet> tile_set = edited_layer->get_effective_tile_set();
+	if (tile_set.is_null()) {
 		return;
 	}
 
@@ -291,14 +296,14 @@ void TileMapEditorTilesPlugin::_update_source_display() {
 	}
 }
 
-void TileMapEditorTilesPlugin::_patterns_item_list_gui_input(const Ref<InputEvent> &p_event) {
-	TileMap *tile_map = Object::cast_to<TileMap>(ObjectDB::get_instance(tile_map_id));
-	if (!tile_map) {
+void TileMapLayerEditorTilesPlugin::_patterns_item_list_gui_input(const Ref<InputEvent> &p_event) {
+	TileMapLayer *edited_layer = _get_edited_layer();
+	if (!edited_layer) {
 		return;
 	}
 
-	Ref<TileSet> tile_set = tile_map->get_tileset();
-	if (!tile_set.is_valid() || EditorNode::get_singleton()->is_resource_read_only(tile_set)) {
+	Ref<TileSet> tile_set = edited_layer->get_effective_tile_set();
+	if (tile_set.is_null()) {
 		return;
 	}
 
@@ -327,7 +332,7 @@ void TileMapEditorTilesPlugin::_patterns_item_list_gui_input(const Ref<InputEven
 	}
 }
 
-void TileMapEditorTilesPlugin::_pattern_preview_done(Ref<TileMapPattern> p_pattern, Ref<Texture2D> p_texture) {
+void TileMapLayerEditorTilesPlugin::_pattern_preview_done(Ref<TileMapPattern> p_pattern, Ref<Texture2D> p_texture) {
 	// TODO optimize ?
 	for (int i = 0; i < patterns_item_list->get_item_count(); i++) {
 		if (patterns_item_list->get_item_metadata(i) == p_pattern) {
@@ -337,14 +342,14 @@ void TileMapEditorTilesPlugin::_pattern_preview_done(Ref<TileMapPattern> p_patte
 	}
 }
 
-void TileMapEditorTilesPlugin::_update_patterns_list() {
-	TileMap *tile_map = Object::cast_to<TileMap>(ObjectDB::get_instance(tile_map_id));
-	if (!tile_map) {
+void TileMapLayerEditorTilesPlugin::_update_patterns_list() {
+	TileMapLayer *edited_layer = _get_edited_layer();
+	if (!edited_layer) {
 		return;
 	}
 
-	Ref<TileSet> tile_set = tile_map->get_tileset();
-	if (!tile_set.is_valid()) {
+	Ref<TileSet> tile_set = edited_layer->get_effective_tile_set();
+	if (tile_set.is_null()) {
 		return;
 	}
 
@@ -354,7 +359,7 @@ void TileMapEditorTilesPlugin::_update_patterns_list() {
 		int id = patterns_item_list->add_item("");
 		patterns_item_list->set_item_metadata(id, tile_set->get_pattern(i));
 		patterns_item_list->set_item_tooltip(id, vformat(TTR("Index: %d"), i));
-		TilesEditorUtils::get_singleton()->queue_pattern_preview(tile_set, tile_set->get_pattern(i), callable_mp(this, &TileMapEditorTilesPlugin::_pattern_preview_done));
+		TilesEditorUtils::get_singleton()->queue_pattern_preview(tile_set, tile_set->get_pattern(i), callable_mp(this, &TileMapLayerEditorTilesPlugin::_pattern_preview_done));
 	}
 
 	// Update the label visibility.
@@ -369,14 +374,14 @@ void TileMapEditorTilesPlugin::_update_patterns_list() {
 	select_last_pattern = false;
 }
 
-void TileMapEditorTilesPlugin::_update_atlas_view() {
-	TileMap *tile_map = Object::cast_to<TileMap>(ObjectDB::get_instance(tile_map_id));
-	if (!tile_map) {
+void TileMapLayerEditorTilesPlugin::_update_atlas_view() {
+	TileMapLayer *edited_layer = _get_edited_layer();
+	if (!edited_layer) {
 		return;
 	}
 
-	Ref<TileSet> tile_set = tile_map->get_tileset();
-	if (!tile_set.is_valid()) {
+	Ref<TileSet> tile_set = edited_layer->get_effective_tile_set();
+	if (tile_set.is_null()) {
 		return;
 	}
 
@@ -385,19 +390,19 @@ void TileMapEditorTilesPlugin::_update_atlas_view() {
 	TileSetAtlasSource *atlas_source = Object::cast_to<TileSetAtlasSource>(source);
 	ERR_FAIL_NULL(atlas_source);
 
-	tile_atlas_view->set_atlas_source(*tile_map->get_tileset(), atlas_source, source_id);
+	tile_atlas_view->set_atlas_source(*tile_set, atlas_source, source_id);
 	TilesEditorUtils::get_singleton()->synchronize_atlas_view(tile_atlas_view);
 	tile_atlas_control->queue_redraw();
 }
 
-void TileMapEditorTilesPlugin::_update_scenes_collection_view() {
-	TileMap *tile_map = Object::cast_to<TileMap>(ObjectDB::get_instance(tile_map_id));
-	if (!tile_map) {
+void TileMapLayerEditorTilesPlugin::_update_scenes_collection_view() {
+	TileMapLayer *edited_layer = _get_edited_layer();
+	if (!edited_layer) {
 		return;
 	}
 
-	Ref<TileSet> tile_set = tile_map->get_tileset();
-	if (!tile_set.is_valid()) {
+	Ref<TileSet> tile_set = edited_layer->get_effective_tile_set();
+	if (tile_set.is_null()) {
 		return;
 	}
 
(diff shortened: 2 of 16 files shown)


AThousandShips:
Are we okay with major compatibility breakage for tilemap at this point? I'd say changing the inheritance even by inserting a different class between a class and it's parent is breaking compatibility

AThousandShips:
I'll tag it as such for now, as per the general concept of it, and we'll see if it's desirable, unsure what areas it might affect elsewhere

editor/plugins/tiles/tiles_editor_plugin.cpp
```diff
@@ -324,62 +324,117 @@ void TileMapEditorPlugin::_tile_map_changed() {
...
+
+	Node *parent = edited_layer->get_parent();
+	ERR_FAIL_NULL(parent);
+
+	TileMapLayer *new_layer = Object::cast_to<TileMapLayer>(parent->get_node_or_null(String(p_name)));
+	edit(new_layer);
+}
+
+void TileMapEditorPlugin::_edit_tile_map_layer(TileMapLayer *p_tile_map_layer) {
+	ERR_FAIL_COND(!p_tile_map_layer);
```
AThousandShips, line 355:
```suggestion
	ERR_FAIL_NULL(p_tile_map_layer);
```

editor/plugins/tiles/tiles_editor_plugin.cpp
```diff
@@ -324,62 +324,117 @@ void TileMapEditorPlugin::_tile_map_changed() {
...
+		tile_set_plugin_singleton->make_visible(true);
+		tile_set_id = tile_set->get_instance_id();
+	} else {
+		tile_set_plugin_singleton->edit(nullptr);
+		tile_set_plugin_singleton->make_visible(false);
+	}
+}
+
+void TileMapEditorPlugin::_edit_tile_map_group(TileMapGroup *p_tile_map_group) {
+	ERR_FAIL_COND(!p_tile_map_group);
```
AThousandShips, line 384:
```suggestion
	ERR_FAIL_NULL(p_tile_map_group);
```

editor/plugins/tiles/tiles_editor_plugin.cpp
```diff
@@ -324,62 +324,117 @@ void TileMapEditorPlugin::_tile_map_changed() {
...
+
+	Vector<StringName> selected_layers = p_tile_map_group->get_selected_layers();
+
+	TileMapLayer *selected_layer = nullptr;
+	if (selected_layers.size() >= 1) {
```
AThousandShips, line 389:
```suggestion
	if (selected_layers.size() > 0) {
```

KoBeWi:
When you deselect TileMap, the layer highlight will stay.

KoBeWi:
Editor crash when changing TileMap node type to TileMapGroup:
```
Engine version: Godot Engine v4.3.dev.custom_build (0bcc0e92b3f0ac57d4c4650722f347593a258572)
[0] TileMap::is_collision_animatable (C:\godot_source\scene\2d\tile_map.cpp:342)
[1] TileMapLayer::_update_notify_local_transform (C:\godot_source\scene\2d\tile_map_layer.cpp:1610)
[2] TileMapLayer::_notification (C:\godot_source\scene\2d\tile_map_layer.cpp:1687)
[3] TileMapLayer::_notificationv (C:\godot_source\scene\2d\tile_map_layer.h:218)
[4] Object::notification (C:\godot_source\core\object\object.cpp:840)
[5] Node::_propagate_enter_tree (C:\godot_source\scene\main\node.cpp:264)
[6] Node::_set_tree (C:\godot_source\scene\main\node.cpp:2996)
[7] Node::_add_child_nocheck (C:\godot_source\scene\main\node.cpp:1404)
[8] Node::add_child (C:\godot_source\scene\main\node.cpp:1432)
[9] Node::replace_by (C:\godot_source\scene\main\node.cpp:2880)
[10] SceneTreeDock::_replace_node (C:\godot_source\editor\scene_tree_dock.cpp:2790)
[11] SceneTreeDock::replace_node (C:\godot_source\editor\scene_tree_dock.cpp:2730)
[12] SceneTreeDock::_create (C:\godot_source\editor\scene_tree_dock.cpp:2665)
[13] call_with_variant_args_helper<SceneTreeDock> (C:\godot_source\core\variant\binder_common.h:308)
[14] call_with_variant_args<SceneTreeDock> (C:\godot_source\core\variant\binder_common.h:418)
[15] CallableCustomMethodPointer<SceneTreeDock>::call (C:\godot_source\core\object\callable_method_pointer.h:99)
[16] Callable::callp (C:\godot_source\core\variant\callable.cpp:57)
[17] Object::emit_signalp (C:\godot_source\core\object\object.cpp:1128)
[18] Node::emit_signalp (C:\godot_source\scene\main\node.cpp:3676)
[19] Object::emit_signal<> (C:\godot_source\core\object\object.h:922)
[20] CreateDialog::_confirmed (C:\godot_source\editor\create_dialog.cpp:434)
[21] call_with_variant_args_helper<CreateDialog> (C:\godot_source\core\variant\binder_common.h:308)
[22] call_with_variant_args<CreateDialog> (C:\godot_source\core\variant\binder_common.h:418)
[23] CallableCustomMethodPointer<CreateDialog>::call (C:\godot_source\core\object\callable_method_pointer.h:99)
[24] Callable::callp (C:\godot_source\core\variant\callable.cpp:57)
[25] Object::emit_signalp (C:\godot_source\core\object\object.cpp:1128)
[26] Node::emit_signalp (C:\godot_source\scene\main\node.cpp:3676)
[27] Object::emit_signal<> (C:\godot_source\core\object\object.h:922)
[28] Tree::gui_input (C:\godot_source\scene\gui\tree.cpp:3844)
[29] Control::_call_gui_input (C:\godot_source\scene\gui\control.cpp:1816)
[30] Viewport::_gui_call_input (C:\godot_source\scene\main\viewport.cpp:1595)
[31] Viewport::_gui_input_event (C:\godot_source\scene\main\viewport.cpp:1824)
[32] Viewport::push_input (C:\godot_source\scene\main\viewport.cpp:3351)
[33] Window::_window_input (C:\godot_source\scene\main\window.cpp:1616)
[34] call_with_variant_args_helper<Window,Ref<InputEvent> const &,0> (C:\godot_source\core\variant\binder_common.h:303)
[35] call_with_variant_args<Window,Ref<InputEvent> const &> (C:\godot_source\core\variant\binder_common.h:418)
[36] CallableCustomMethodPointer<Window,Ref<InputEvent> const &>::call (C:\godot_source\core\object\callable_method_pointer.h:99)
[37] Callable::callp (C:\godot_source\core\variant\callable.cpp:57)
[38] Callable::call<Ref<InputEvent> > (C:\godot_source\core\variant\variant.h:863)
[39] DisplayServerWindows::_dispatch_input_event (C:\godot_source\platform\windows\display_server_windows.cpp:2751)
[40] DisplayServerWindows::_dispatch_input_events (C:\godot_source\platform\windows\display_server_windows.cpp:2721)
[41] Input::_parse_input_event_impl (C:\godot_source\core\input\input.cpp:760)
[42] Input::flush_buffered_events (C:\godot_source\core\input\input.cpp:1026)
[43] DisplayServerWindows::process_events (C:\godot_source\platform\windows\display_server_windows.cpp:2437)
[44] OS_Windows::run (C:\godot_source\platform\windows\os_windows.cpp:1476)
[45] widechar_main (C:\godot_source\platform\windows\godot_windows.cpp:182)
[46] _main (C:\godot_source\platform\windows\godot_windows.cpp:204)
[47] main (C:\godot_source\platform\windows\godot_windows.cpp:218)
[48] WinMain (C:\godot_source\platform\windows\godot_windows.cpp:232)
[49] __scrt_common_main_seh (D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288)
[50] <couldn't map PC to fn name>
-- END OF BACKTRACE --
```
Requires a sufficiently complex TileMap to happen.

KoBeWi:
Unlike the previous PR, this one introduces an unimplemented functionality in the form of TileMapGroup. You can create it, but it's completely useless and has no documentation. It's not really a problem at this stage, but it puts more pressure to have the new workflow finished in time.

I wonder if there is a way to do what you are planning without adding a new node. From what I understand, the only difference between TileMap and TileMapGroup is that the former doesn't expose its layer nodes, so you have to edit them in the inspector. But since it will create the layer internally anyway, it should work fine if it happens when the user opens an old scene for the first time, no? It would require a bit of compatibility code, but the TileMap can be converted automatically to TileMapGroup, thus the new node is not really needed.

groud:
> Unlike the previous PR, this one introduces an unimplemented functionality in the form of TileMapGroup. You can create it, but it's completely useless and has no documentation. It's not really a problem at this stage, but it puts more pressure to have the new workflow finished in time.

I could make it an abstract class I guess, that would work.

> I wonder if there is a way to do what you are planning without adding a new node. From what I understand, the only difference between TileMap and TileMapGroup is that the former doesn't expose its layer nodes, so you have to edit them in the inspector. But since it will create the layer internally anyway, it should work fine if it happens when the user opens an old scene for the first time, no? It would require a bit of compatibility code, but the TileMap can be converted automatically to TileMapGroup, thus the new node is not really needed.

Hmm. That is good question. There are several problem to solve, but I am not 100% sure of what is doable. Is see several things to solve:
- if we keep only one node, that means we will have to keep the TileMap API as is, but deprecate a lot of its methods. I guess that's acceptable, but it might be a bit more confusing than deprecating the node. 
- Some parts of the to-be-deprecated API rely on the child nodes having values controlled by the parent TileMap. For example, the `collision_animatable` property requires the children TileMapLayer nodes to be set as `use_kinematic_body`. Also, I moved some settings to individual layers, like the `rendering_quadrant_size`. This means that some of those will have no effect anymore on the children nodes (or it could have, but it might be weird in the editor that a parent node modifies its children properties)
- The deprecated API will be slowed down as every call will have to iterate over children nodes to find the layer corresponding to a given index (in case you have nodes as child which are not TileMapLayers). That might be acceptable, idk.
- In general, that would keep a lot of bloat in the TileMap API, so not sure what is worth. This might be solvable by keeping the TileMapGroup we have now, but not expose it to users though. That would simplify the work needed the day be remove the deprecated part basically.

KoBeWi:
idk then. The day when we remove deprecated stuff is very far away, so when redesigning the API we need to keep in mind that this will stay for a long while. Also another concern is what we want to be the default name of the node. I think TileMap is more intuitive/familiar than TileMapGroup and making the latter default would be a bit odd. It would be best if we could rename the old node to e.g. TileMapOld, but it's not possible unfortunately :/

If we go the route of adding a new class (which can be abstract for now), I think at least it should have a better name.
LayeredTileMap? TileManager? TileLayerArray? TileLayerMap? No idea tbh.

editor/plugins/tiles/tiles_editor_plugin.cpp
AThousandShips: I mean it's up to you but `>= 1` feels kinda weird, for an integer it's the same and it's less clear than `> 0`

groud: I don't know, it's the same to me... 

Anyway, it's changed on my local fork.

doc/classes/TileMapLayerGroup.xml
```diff
@@ -0,0 +1,14 @@
...
+<class name="TileMapLayerGroup" inherits="Node2D" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
+	<brief_description>
+	</brief_description>
+	<description>
+	</description>
+	<tutorials>
+	</tutorials>
+	<members>
+		<member name="tile_set" type="TileSet" setter="set_tileset" getter="get_tileset">
+			The assigned [TileSet]. This TileSet will be applied to all children layers.
```
AThousandShips, line 11:
```suggestion
			The assigned [TileSet]. This TileSet will be applied to all child layers.
```

groud:
Alright. According to @dsnopek, adding the detected API changes in the `.expected` file should work.
I also renamed `TileMapGroup` to `TileMapLayerGroup` and made it abstract.

groud:
> You seem to have generated your documentation on an out of date branch and deleted some content

It should be fixed now, the PR should be ready.

doc/classes/TileMapLayerGroup.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="TileMapLayerGroup" inherits="Node2D" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
+	<brief_description>
+		Groups a set of tile map layers together, allowing them to share a provided [TileSet].
+	</brief_description>
+	<description>
+	</description>
```
KoBeWi, line 7: Could use a description too, event explaining that the class is work in progress.
Though not sure how long it would be useful.

scene/resources/tile_set.h
```diff
@@ -527,15 +528,17 @@ class TileSet : public Resource {
...
-	void draw_tile_shape(CanvasItem *p_canvas_item, Transform2D p_transform, Color p_color, bool p_filled = false, Ref<Texture2D> p_texture = Ref<Texture2D>());
+	Vector<Vector2> get_tile_shape_polygon() const;
+	void draw_tile_shape(CanvasItem *p_canvas_item, Transform2D p_transform, Color p_color, bool p_filled = false, Ref<Texture2D> p_texture = Ref<Texture2D>()) const;
 
 	// Used by TileMap/TileMapLayer
 	Vector2 map_to_local(const Vector2i &p_pos) const;
 	Vector2i local_to_map(const Vector2 &p_pos) const;
 	bool is_existing_neighbor(TileSet::CellNeighbor p_cell_neighbor) const;
 	Vector2i get_neighbor_cell(const Vector2i &p_coords, TileSet::CellNeighbor p_cell_neighbor) const;
+	TypedArray<Vector2i> get_surrounding_cells(const Vector2i &p_coords);
```
KoBeWi, line 539:
```suggestion
	TypedArray<Vector2i> get_surrounding_cells(const Vector2i &p_coords) const;
```
Why this isn't const?

doc/classes/TileMapLayerGroup.xml
groud: I added a few lines for now, I guess it should be enough until we expose it properly.

scene/2d/tile_map_layer_group.cpp
```diff
@@ -0,0 +1,139 @@
...
+/**************************************************************************/
+
+#include "tile_map_layer_group.h"
+
+#include "core/core_string_names.h"
+#include "scene/2d/tile_map_layer.h"
+
+void TileMapLayerGroup::_cleanup_selected_layers() {
+	for (int i = 0; i < selected_layers.size(); i++) {
+		const StringName &name = selected_layers[i];
```
KoBeWi, line 38:
```suggestion
		const String name = selected_layers[i];
```
You don't use the StringName, so the cast can be performed here. (although it doesn't matter much)

scene/2d/tile_map_layer.h
```diff
@@ -232,13 +232,15 @@ class TileMapLayer : public Node2D {
 		DIRTY_FLAGS_LAYER_NAVIGATION_ENABLED,
 		DIRTY_FLAGS_LAYER_INDEX_IN_TILE_MAP_NODE,
 
-		DIRTY_FLAGS_TILE_MAP_SELECTED_LAYER,
+		DIRTY_FLAGS_TILE_MAP_GROUP_SELECTED_LAYERS,
```
KoBeWi, line 235: Shouldn't this be DIRTY_FLAGS_TILE_MAP_LAYER_GROUP? For consistency.

scene/2d/tile_map_layer.h
```diff
@@ -338,8 +341,14 @@ class TileMapLayer : public Node2D {
 	void _renamed();
 	void _update_notify_local_transform();
 
+	// Internal updates
```
KoBeWi, line 344:
```suggestion
	// Internal updates.
```

editor/plugins/tiles/tile_map_layer_editor.cpp
```diff
@@ -96,14 +97,14 @@ void TileMapEditorTilesPlugin::_update_toolbar() {
...
+void TileMapLayerEditorTilesPlugin::_update_transform_buttons() {
+	TileMapLayer *edited_layer = Object::cast_to<TileMapLayer>(ObjectDB::get_instance(edited_tile_map_layer_id));
+	if (!edited_layer) {
 		return;
 	}
 
-	Ref<TileSet> tile_set = tile_map->get_tileset();
-	if (tile_set.is_null() || selection_pattern.is_null()) {
+	Ref<TileSet> tile_set = edited_layer->get_effective_tile_set();
+	if (!tile_set.is_valid() || selection_pattern.is_null()) {
```
KoBeWi, line 107: Why did you change it?
```suggestion
	if (tile_set.is_null() || selection_pattern.is_null()) {
```

editor/plugins/tiles/tile_map_layer_editor.cpp
```diff
@@ -291,14 +292,14 @@ void TileMapEditorTilesPlugin::_update_source_display() {
 	}
 }
 
-void TileMapEditorTilesPlugin::_patterns_item_list_gui_input(const Ref<InputEvent> &p_event) {
-	TileMap *tile_map = Object::cast_to<TileMap>(ObjectDB::get_instance(tile_map_id));
-	if (!tile_map) {
+void TileMapLayerEditorTilesPlugin::_patterns_item_list_gui_input(const Ref<InputEvent> &p_event) {
+	TileMapLayer *edited_layer = Object::cast_to<TileMapLayer>(ObjectDB::get_instance(edited_tile_map_layer_id));
```
KoBeWi, line 296: I think this deserves a helper method, because it's done everywhere.

editor/plugins/tiles/tile_map_layer_editor.cpp
```diff
@@ -3826,119 +3786,124 @@ Vector<Vector2i> TileMapEditor::get_line(TileMap *p_tile_map, Vector2i p_from_ce
...
+		for (int i = 0; i < tile_map_layer_group->get_child_count(); i++) {
+			const TileMapLayer *layer = Object::cast_to<TileMapLayer>(tile_map_layer_group->get_child(i));
+			if (layer) {
+				int index = layers_selection_button->get_item_count();
+				layers_selection_button->add_item(layer->get_name());
+				layers_selection_button->set_item_disabled(index, !layer->is_enabled());
+				layers_selection_button->set_item_metadata(index, layer->get_name());
+				if (edited_tile_map_layer_id.is_valid()) {
+					const TileMapLayer *selected_layer = Object::cast_to<TileMapLayer>(ObjectDB::get_instance(edited_tile_map_layer_id));
+					ERR_FAIL_COND(!selected_layer);
```
KoBeWi, line 3875: I think ERR_CONTINUE would be better.

editor/plugins/tiles/tile_map_layer_editor.cpp
groud: You mean this part  `Object::cast_to<TileMapLayer>(ObjectDB::get_instance(edited_tile_map_layer_id));` ?
I know this this snippet is used multiple time but I am not sure I can squash it in a helper while keeping things readable:
```
	TileMapLayer *edited_layer = Object::cast_to<TileMapLayer>(ObjectDB::get_instance(edited_tile_map_layer_id));
	if (!edited_layer) {
		return;
	}

	Ref<TileSet> tile_set = edited_layer->get_effective_tile_set();
	if (tile_set.is_null()) {
		return;
	}
```

scene/2d/tile_map_layer_group.h
```diff
@@ -0,0 +1,67 @@
...
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef TILE_MAP_LAYER_GROUP_H
+#define TILE_MAP_LAYER_GROUP_H
+
+#include "scene/2d/node_2d.h"
+#include "scene/resources/tile_set.h"
```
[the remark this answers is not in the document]
groud: Had to add it in the TileMap header anyway though. But well, it's forward-declared here now.

editor/plugins/tiles/tile_map_layer_editor.cpp
KoBeWi: Well it could be something like this
```
TileMapLayer *edited_layer = get_edited_layer();
```
Not sure how it can negatively impact readability.

I'm not referring to the TileSet part, as it's not used everywhere.

scene/2d/tile_map_layer_group.h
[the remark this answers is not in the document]
KoBeWi: > Had to add it in the TileMap header anyway though.

I don't see why though. That file doesn't even refer to TileSet at all.

KoBeWi (approved):

editor/plugins/tiles/tile_map_layer_editor.cpp
groud: > Not sure how it can negatively impact readability.

I meant if I had to compact the whole snippet. But just for the edited_layer that makes sense. I'll do that then.

scene/2d/tile_map_layer_group.h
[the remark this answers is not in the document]
groud: > I don't see why though. That file doesn't even refer to TileSet at all.

I mean I had to keep it in `tile_map.h`. It makes use of `TileSet::TerrainsPattern` which requires TileSet to be imported.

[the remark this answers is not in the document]
KoBeWi: Ah right, I only looked at the changes and missed that there is other usage in the file.

YuriSizov:
We should merge it soon (but probably not for dev3, which we plan to have tomorrow). But before we do, I wanted someone to take a look at the changes so we have two approvals. It can be me or Remi, but volunteers are also welcome!

editor/plugins/tiles/tiles_editor_plugin.cpp
LiBooks: > I don't know, it's the same to me...
> 
> Anyway, it's changed on my local fork.

I agree with your opinion. His changes are meaningless, and this is his meaningless speculative behavior

akien-mga: @LiBooks Please refrain from personally attacking contributors like this. This is a clear breach of our Code of Conduct: https://godotengine.org/code-of-conduct/

I'll remove your post. Next occurrence of such inappropriate behavior would result in a ban.

groud: > I agree with your opinion. His changes are meaningless, and this is his meaningless speculative behavior

Well, thanks, but please avoid reopening that kind of threads on such minor details. It's not super constructive. (no harm done, just an advice for the future :) )
Thanks for having a look to the PR anyway!

editor/plugins/tiles/tile_map_layer_editor.cpp
```diff
@@ -40,29 +40,34 @@
...
 #include "scene/gui/split_container.h"
 
 #include "core/input/input.h"
 #include "core/math/geometry_2d.h"
 #include "core/os/keyboard.h"
 
-void TileMapEditorTilesPlugin::tile_set_changed() {
+TileMapLayer *TileMapLayerSubEditorPlugin::_get_edited_layer() const {
+	return Object::cast_to<TileMapLayer>(ObjectDB::get_instance(edited_tile_map_layer_id));
+};
```
AThousandShips, line 53:
```suggestion
}
```

editor/plugins/tiles/tile_map_layer_editor.cpp
```diff
@@ -3654,10 +3620,14 @@ TileMapEditorTerrainsPlugin::TileMapEditorTerrainsPlugin() {
...
 }
 
-TileMapEditorTerrainsPlugin::~TileMapEditorTerrainsPlugin() {
+TileMapLayerEditorTerrainsPlugin::~TileMapLayerEditorTerrainsPlugin() {
 }
 
-void TileMapEditor::_notification(int p_what) {
+TileMapLayer *TileMapLayerEditor::_get_edited_layer() const {
+	return Object::cast_to<TileMapLayer>(ObjectDB::get_instance(edited_tile_map_layer_id));
+};
```
AThousandShips, line 3628:
```suggestion
}
```

AThousandShips (approved):
The code looks good, did some basic testing but haven't done extensive use of it yet, but the changes to the map itself seems reasonable, and most of the editor side is renaming

Edit: this might worsen threading issues due to making the queuing unique to each layer, but the threading should be simple to fix, see:
* #87478

scene/2d/tile_map_layer_group.h
```diff
@@ -0,0 +1,68 @@
...
+	void _tile_set_changed();
+
+protected:
+	Ref<TileSet> tile_set;
+
+	void _notification(int p_what);
+	static void _bind_methods();
+
+public:
+	// For editor use.
```
kleonc, line 55: Shouldn't it be within `#ifdef TOOLS_ENABLED` or something like that then? :thinking:

scene/2d/tile_map_layer_group.cpp
```diff
@@ -0,0 +1,140 @@
...
+		if (layer) {
+			layer->notify_tile_map_change(TileMapLayer::DIRTY_FLAGS_LAYER_GROUP_SELECTED_LAYERS);
+		}
+	}
+}
+
+void TileMapLayerGroup::_notification(int p_what) {
+	if (p_what == NOTIFICATION_CHILD_ORDER_CHANGED) {
+		_cleanup_selected_layers();
+	}
```
kleonc, line 75: Needs to be done only on child removed? If that's the case then overriding `remove_child_notify` instead probably makes more sense.

scene/2d/tile_map_layer.cpp
```diff
@@ -2618,14 +2643,23 @@ Vector2i TileMapLayer::get_coords_for_body_rid(RID p_physics_body) const {
...
 }
 
+Ref<TileSet> TileMapLayer::get_effective_tile_set() const {
+	TileMapLayerGroup *tile_map_layer_group = Object::cast_to<TileMapLayerGroup>(get_parent());
+	if (tile_map_layer_group) {
+		return tile_map_layer_group->get_tileset();
+	} else {
+		return Ref<TileSet>();
+	}
+}
```
kleonc, line 2653: Note there's still `TileMapLayer::_fetch_tileset` which does the same but assumming `TileMap` instead of `TileMapGroup`. Probably should be removed and all its usages replaced by this? :thinking:

kleonc (commented):
Skimmed though the code, mostly looks fine to me. :+1:

editor/plugins/tiles/tile_map_layer_editor.cpp
```diff
@@ -983,10 +978,7 @@ void TileMapEditorTilesPlugin::forward_canvas_draw_over_viewport(Control *p_over
 							}
 
 							// Get the tile modulation.
-							Color modulate = tile_data->get_modulate();
-							Color self_modulate = tile_map->get_modulate_in_tree() * tile_map->get_self_modulate();
-							modulate *= self_modulate;
-							modulate *= tile_map->get_layer_modulate(tile_map_layer);
+							Color modulate = tile_data->get_modulate() * edited_layer->get_modulate_in_tree();
```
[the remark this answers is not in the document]
groud: Right, I assumed `get_modulate_in_tree` would apply it, but no. Thanks!

scene/2d/tile_map_layer.cpp
groud: Ah right, good idea.

scene/2d/tile_map_layer_group.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+bool TileMapLayerGroup::is_highlighting_selected_layer() const {
+	return highlight_selected_layer;
+}
+
+#endif // TOOLS_ENABLED
+
+void TileMapLayerGroup::remove_child_notify(Node *p_child) {
+	_cleanup_selected_layers();
```
AThousandShips, line 104:
```suggestion
#ifdef TOOLS_ENABLED
	_cleanup_selected_layers();
#endif
```

AThousandShips:
Unsure if this is exactly what's intended here but applied this to fix the compilation issues on my end:
```diff
diff --git a/scene/2d/tile_map_layer.cpp b/scene/2d/tile_map_layer.cpp
index 5e351293b3c42..e866d82c03564 100644
--- a/scene/2d/tile_map_layer.cpp
+++ b/scene/2d/tile_map_layer.cpp
@@ -198,6 +198,7 @@ void TileMapLayer::_rendering_update() {
 
 		// Modulate the layer.
 		Color layer_modulate = get_modulate();
+#ifdef TOOLS_ENABLED
 		const Vector<StringName> selected_layers = tile_map_node->get_selected_layers();
 		if (tile_map_node->is_highlighting_selected_layer() && selected_layers.size() == 1 && get_name() != selected_layers[0]) {
 			TileMapLayer *selected_layer = Object::cast_to<TileMapLayer>(tile_map_node->get_node_or_null(String(selected_layers[0])));
@@ -212,6 +213,7 @@ void TileMapLayer::_rendering_update() {
 				}
 			}
 		}
+#endif // TOOLS_ENABLED
 		rs->canvas_item_set_modulate(get_canvas_item(), layer_modulate);
 	}
 
@@ -2965,4 +2967,4 @@ TerrainConstraint::TerrainConstraint(Ref<TileSet> p_tile_set, const Vector2i &p_
 		}
 	}
 	terrain = p_terrain;
-}
\ No newline at end of file
+}
diff --git a/scene/2d/tile_map_layer_group.cpp b/scene/2d/tile_map_layer_group.cpp
index be522e2aa9c0d..7ddafc12104ca 100644
--- a/scene/2d/tile_map_layer_group.cpp
+++ b/scene/2d/tile_map_layer_group.cpp
@@ -101,7 +101,9 @@ bool TileMapLayerGroup::is_highlighting_selected_layer() const {
 #endif // TOOLS_ENABLED
 
 void TileMapLayerGroup::remove_child_notify(Node *p_child) {
+#ifdef TOOLS_ENABLED
 	_cleanup_selected_layers();
+#endif
 }
 
 void TileMapLayerGroup::_bind_methods() {
 ```
 
 Would be good to get that out of the way and this merged

akien-mga:
Thanks!

PR #87380, opened 2024-01-19
Allow an optional editor SCsub in platforms

RandomShaper:
Each platform port have `detect.py` and `SCsub` scripts, but they are only relevant when building precisely for that platform. However, their editor part ~`export/`~ can't be further customized and have to be happy with the simple logic the editor build script features for them. This PR lets platforms have an optional `SCsub` to allow further customizations, like adding extra files (e.g., and this is my current real-world need, having third-party code scoped to the editor portion of a given platform).

**UPDATE:** Added also a way to extend modules with per-plaform `SCsub`. Moving to draft until this is discussed further.

**UPDATE:** Removed the addition above, as it turned out to be a bad approach.

### DIFF
--- a/editor/SCsub
+++ b/editor/SCsub
@@ -45,6 +45,11 @@ if env.editor_build:
 
     methods.write_file_if_needed("register_exporters.gen.cpp", reg_exporters_inc + reg_exporters)
 
+    for e in env.platform_exporters:
+        scsub_path = os.path.join(Dir("#platform/%s/editor" % e).abspath, "SCsub")
+        if os.path.exists(scsub_path):
+            SConscript(scsub_path)
+
     # Core API documentation.
     docs = []
     docs += Glob("#doc/classes/*.xml")


editor/SCsub
```diff
@@ -40,6 +40,11 @@ if env.editor_build:
         reg_exporters += "\tregister_" + e + "_exporter_types();\n"
     reg_exporters += "}\n"
 
+    for e in env.platform_exporters:
+        scsub_path = "../platform/" + e + "/export/SCsub"
+        if os.path.exists(scsub_path):
```
akien-mga, line 45: Adding support for this makes sense, but shouldn't it then run *instead* of adding all the `.cpp` files in the folder? Typically a `SCsub` is responsible for adding the files in its folder, and is used when what a folder needs is more than just plainly globbing its files.

If we do this, that means making line 33 the `else` branch of the SCsub check.

RandomShaper: Makes sense, but I also have a little concern with that. If the SCsub is meant to add the files, that can be taken as freedom not to add them. And that wouldn't work because the build script assumes that there's an export plugin there with the proper name. And it wouldn't make sense otherwise, given the name of the subdir. So, I'm wondering now if it wouldn't be better to keep `export/` uncustomizable and instead add support for an optional `platform/???/editor/` directory, where the extra files and SCsub would be. Or maybe that would still be a bit confusing, since the export plugin is part of the editor, too. However, reorganizing the structure so it can represent and support everything clearly* would be too much of a change for such a marginal use case. Thoughts?

RandomShaper:
@shana is working on something similar to this for a broader scope. We've agreed that these extra build scripts should follow the root dir structure. So, I've updated this PR in a way that the extra file has to be `platform/.../editor/SCsub`.

editor/SCsub
shana: > Typically a `SCsub` is responsible for adding the files in its folder, and is used when what a folder needs is more than just plainly globbing its files.

As @RandomShaper mentioned, I'm also planning this type of change for all artifact-generating SCsubs. The goal of these platform-specific, optional, SCsub files, would be to augment/modify the original caller environment, by adding files or defines or other things, or even taking the existing source files that already registered and calling `add_library` to do a specialized build just for that platform, for example. 

This lets us split responsibilities and customize the build without having to maintain a ton of patches everywhere. The caller knows what files go into building a thing and what all the default settings are in general, and the platform knows what tweaks are needed on top in order to compile for its specific environment. Wherever we have "if env.platform == windows", for example, that's a sign that the platform needs to customize a particular part of the build. If a specialized SCsub inside the platform folder got called at that point to customize the build, we wouldn't need any of those types of checks anywhere, the platforms would just take care of it.

Issue #87374, opened 2024-01-19
`Control.LayoutMode` enum needs to be exposed for the `layout_mode` property


geekros:
### Tested versions

v4.2.1.stable.official [b09f793f5]

### System information

Window 11 v4.2.1.stable.official [b09f793f5]

### Issue description

When using item_label.layout_mode = 1, a warning message appears:

```
W 0:00:01:0356   Integer used when an enum value is expected. If this is intended cast the integer to the enum type.
```

but I can't find the corresponding available enumeration, such as LAYOUT_MODE_POSITION.

### Steps to reproduce

```
var item_label:Label = Label.new()
item_label.layout_mode = LAYOUT_MODE_POSITION
```

### Minimal reproduction project (MRP)

```
var item_label:Label = Label.new()
item_label.layout_mode = LAYOUT_MODE_POSITION
```

geekros:
I don't want to see the warning messages, and I need to know where you have hidden its enumeration?

akien-mga:
It seems like we forgot to bind the `LayoutMode` enum from the `Control` class, so indeed `LAYOUT_MODE_POSITION` isn't exposed in the API / docs.

YuriSizov:
> It seems like we forgot to bind the LayoutMode enum from the Control class

No, we have not. You are not supposed to use these properties and these enums from scripting. They are internal.

akien-mga:
Users keep trying to use those internal properties though, so I think we need to ask ourselves "why". Just telling them they're not supposed to use them doesn't seem sufficient.

- Why can internal properties be used at all? It defeats the purpose of hiding them if we need to still support users using them.
- Why do users keep trying to use them? How do they find them, and what do they allow? Do they work?
- Why do we want to prevent users from using them?

AThousandShips:
Related:
* https://github.com/godotengine/godot/issues/80209

AThousandShips:
The problem here is that the property is exposed with `PROPERTY_USAGE_INTERNAL`, which AFAIK prevents them from being exposed to extensions and C#, but due to as far as I know an oversight this wasn't applied for GDScript

AThousandShips:
It has an enum type because it's still stored, the proper solution is to not expose it at all as it is internal and doesn't work as a property, it only works when loading a scene in the editor

dalexeev:
[...]

PR #87381, opened 2024-01-19
Better hide internal properties from users

### LINKED ISSUE #25651: Properties tagged PROPERTY_USAGE_INTERNAL are usable from GDscript
I started out reporting that properties tagged PROPERTY_USAGE_INTERNAL were not showing in the docs, but then I saw [this](https://github.com/godotengine/godot/issues/13334) PR and realized that they were not even supposed to be usable in GDScript. 

I noticed with Multimesh which has three arrays that are tagged PROPERTY_USAGE_INTERNAL, but can all be used from GDScript.
https://github.com/godotengine/godot/blob/c19840dad73774dd9ae10e99399e1d3f7284f75b/scene/resources/multimesh.cpp#L267-L269

They don't appear in the docs (which I guess is good?) but they do autocomplete in GDScript and work fine in GDScript.

@akien-mga You may want to look at this.

YuriSizov:
This addresses some of pain points raised by https://github.com/godotengine/godot/issues/87374 and other similar tickets. Internal properties are not meant to be used in scripting, they aren't documented. So this PR does a couple of things.

- Adds a custom tooltip for internal properties to the inspector dock. I'm open to suggestions for what it should say, but so far it's this:

godot windows editor dev x86_64_2024-01-19_18-43-00

- Removes internal properties from GDScript code suggestions and makes them unresolvable for symbol lookup (i.e. ctrl-clicking).

godot windows editor dev x86_64_2024-01-19_18-50-10

This doesn't really solve the root of the issue, nor does it help C# or GDExtension users. But it should help reduce the confusion.

----

Edit: Since the issue is specific to GDScript, this PR would effectively "fix" and close https://github.com/godotengine/godot/issues/25651.

### DIFF
--- a/editor/editor_help.cpp
+++ b/editor/editor_help.cpp
@@ -2862,9 +2862,17 @@ void EditorHelpTooltip::parse_tooltip(const String &p_text) {
 	const String &property_name = slices[2];
 	const String &property_args = slices[3];
 
+	String formatted_text;
+
+	// Exclude internal properties, they are not documented.
+	if (type == "internal_property") {
+		formatted_text = "[i]" + TTR("This property can only be set in the Inspector.") + "[/i]";
+		set_text(formatted_text);
+		return;
+	}
+
 	String title;
 	String description;
-	String formatted_text;
 
 	if (type == "class") {
 		title = class_name;
--- a/editor/editor_inspector.cpp
+++ b/editor/editor_inspector.cpp
@@ -436,6 +436,10 @@ void EditorProperty::set_doc_path(const String &p_doc_path) {
 	doc_path = p_doc_path;
 }
 
+void EditorProperty::set_internal(bool p_internal) {
+	internal = p_internal;
+}
+
 void EditorProperty::update_property() {
 	GDVIRTUAL_CALL(_update_property);
 }
@@ -748,10 +752,10 @@ void EditorProperty::shortcut_input(const Ref<InputEvent> &p_event) {
 		if (ED_IS_SHORTCUT("property_editor/copy_value", p_event)) {
 			menu_option(MENU_COPY_VALUE);
 			accept_event();
-		} else if (ED_IS_SHORTCUT("property_editor/paste_value", p_event) && !is_read_only()) {
+		} else if (!is_read_only() && ED_IS_SHORTCUT("property_editor/paste_value", p_event)) {
 			menu_option(MENU_PASTE_VALUE);
 			accept_event();
-		} else if (ED_IS_SHORTCUT("property_editor/copy_property_path", p_event)) {
+		} else if (!internal && ED_IS_SHORTCUT("property_editor/copy_property_path", p_event)) {
 			menu_option(MENU_COPY_PROPERTY_PATH);
 			accept_event();
 		}
@@ -1036,6 +1040,8 @@ void EditorProperty::_update_popup() {
 	menu->add_icon_shortcut(get_editor_theme_icon(SNAME("ActionPaste")), ED_GET_SHORTCUT("property_editor/paste_value"), MENU_PASTE_VALUE);
 	menu->add_icon_shortcut(get_editor_theme_icon(SNAME("CopyNodePath")), ED_GET_SHORTCUT("property_editor/copy_property_path"), MENU_COPY_PROPERTY_PATH);
 	menu->set_item_disabled(MENU_PASTE_VALUE, is_read_only());
+	menu->set_item_disabled(MENU_COPY_PROPERTY_PATH, internal);
+
 	if (!pin_hidden) {
 		menu->add_separator();
 		if (can_pin) {
@@ -3329,7 +3335,11 @@ void EditorInspector::update_tree() {
 				if (use_doc_hints) {
 					// `|` separator used in `EditorHelpTooltip` for formatting.
 					if (theme_item_name.is_empty()) {
-						ep->set_tooltip_text("property|" + classname + "|" + property_prefix + p.name + "|");
+						if (p.usage & PROPERTY_USAGE_INTERNAL) {
+							ep->set_tooltip_text("internal_property|" + classname + "|" + property_prefix + p.name + "|");
+						} else {
+							ep->set_tooltip_text("property|" + classname + "|" + property_prefix + p.name + "|");
+						}
 					} else {
 						ep->set_tooltip_text("theme_item|" + classname + "|" + theme_item_name + "|");
 					}
@@ -3337,6 +3347,8 @@ void EditorInspector::update_tree() {
 				}
 
 				ep->set_doc_path(doc_path);
+				ep->set_internal(p.usage & PROPERTY_USAGE_INTERNAL);
+
 				ep->update_property();
 				ep->_update_pin_flags();
 				ep->update_editor_property_status();
--- a/modules/gdscript/gdscript_editor.cpp
+++ b/modules/gdscript/gdscript_editor.cpp
@@ -1135,7 +1135,7 @@ static void _find_identifiers_in_base(const GDScriptCompletionIdentifier &p_base
 							List<PropertyInfo> members;
 							scr->get_script_property_list(&members);
 							for (const PropertyInfo &E : members) {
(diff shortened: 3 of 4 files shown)


Mickeon:
We could also take advantage of the the RichTextLabel and have it also display in yellow. If the users opening issues about this gave you nightmares add a warning icon, too.

AdriaandeJongh:
I presume users primarily look at the tooltip to find out what a property does and it seems like this takes that away? I would perhaps instead show the 'only in the inspector' notice **in addition to** the description as opposed to _replace_ the description.

AdriaandeJongh:
Oh, and to make it even better and truly solve this issue, if a property is only found in the inspector, the description could mention how else users should change this property in code, eg. what other / exposed properties they _can_ access.

AThousandShips:
There's no description to replace as it's undocumented because it's internal

YuriSizov:
I agree that a description would be useful even if it only explains what exactly does it do in the inspector. But the problem is, as explained above, that these properties by their nature don't have a description. There is nothing we can show for them.

AdriaandeJongh:
I see. Hmm. Well, then, adding descriptions to those properties would be a great new feature that would clarify their use in the editor a lot. I presume that's out of the scope of this PR?

AThousandShips:
That's not possible unless we add a separate system to handle it, they can't and *shouldn't* be documented in the conventional sense, further that would make them very difficult to translate and track that, it'd be a whole separate system from the documentation system which is what we use to get the descriptions

Mickeon:
Out of scope indeed. As is currently, for a property to have a description it needs to be written in the class reference. But that would mean it would have to be exposed to the documentation generation, which by itself omits internal properties by design.

Assuming you make an exception for all of these properties, you'd still have to hide it from the built-in documentation, while still displaying it when generating a tooltip.

KoBeWi:
It's possible to add descriptions for things not exposed in class reference, see #85868

YuriSizov:
> It's possible to add descriptions for things not exposed in class reference, see #85868

Right, that can be useful here if/when we merge it.

editor/editor_inspector.cpp
```diff
@@ -3329,7 +3329,11 @@ void EditorInspector::update_tree() {
 				if (use_doc_hints) {
 					// `|` separator used in `EditorHelpTooltip` for formatting.
 					if (theme_item_name.is_empty()) {
-						ep->set_tooltip_text("property|" + classname + "|" + property_prefix + p.name + "|");
+						if (p.usage & PROPERTY_USAGE_INTERNAL) {
+							ep->set_tooltip_text("internal_property|" + classname + "|" + property_prefix + p.name + "|");
```
nlupugla, line 3333: Should the tool tip read "internal property" instead of "internal_property" because it's a user-facing message?

YuriSizov: It's not a user facing message.

nlupugla: In that case, totally fine :)
It seemed like it was setting a tool tip somewhere that the user would see.

YuriSizov: No, same as it is with the surrounding code, this text is parsed by the EditorHelpTooltip control and the actual content is different.

editor/editor_help.cpp
```diff
@@ -2862,9 +2862,17 @@ void EditorHelpTooltip::parse_tooltip(const String &p_text) {
 	const String &property_name = slices[2];
 	const String &property_args = slices[3];
 
+	String formatted_text;
+
+	// Exclude internal properties, they are not documented.
+	if (type == "internal_property") {
+		formatted_text = "[i]" + TTR("This property can only be set in the Inspector.") + "[/i]";
```
nlupugla, line 2869: I do think "This property can only be set in the Inspector" is confusing. To me, it just suggests that the user shouldn't change it with a script, but they should feel free to change it in the inspector. If that's what is intended, fine, but I thought the intention was that the user should not touch this property at all. If that's the intention, maybe something like "This is an internal property. It should not be modified by the user."

YuriSizov: > To me, it just suggests that the user shouldn't change it with a script, but they should feel free to change it in the inspector.

That's exactly the intention. Why would we want to have a property visible and editable, but tell people to not touch it?

nlupugla: I thought maybe it was a quirk of the engine that certain properties just needed to be exposed for stuff to work correctly 🤷 

If that's the case, the tooltip should provide some description of what the property actually does no? It seems like either the user is not supposed to interact with the property, so no description is required OR the user is encouraged to edit the property from the inspector, in which case a description seems absolutely necessary.

YuriSizov: > If that's the case, the tooltip should provide some description of what the property actually does no?

Yes, ideally, but as discussed above in the comments, we currently don't have any mechanism to document internal properties because by design they are not supposed to be documented.

nlupugla: I see. Are all internal properties like `anchors_preset` in that there is a `set_anchors_preset` method that the user is supposed to use instead of setting the property itself? If so, maybe there's a way to inject the documentation for the appropriate setter method into the tooltip, or at least a message like "see `set_anchors_preset` method for details."

In any case, if that's beyond the scope of this PR, what you're doing now looks good to me and the code seems fine. As you can tell from my comments, I don't know a lot about this TIWAGOS :)

YuriSizov: `set_anchors_preset` is unrelated to the `anchors_preset` property, they just have a name overlap. Functionally they are related, of course, since both deal with presets for anchors.

That said, the potential description for `anchors_preset` would include references to all existing `set_*_preset` methods.

KoBeWi:
There's also Copy Property Path button in the context menu.
image
Should it be hidden too?

modules/gdscript/gdscript_editor.cpp
```diff
@@ -1083,7 +1083,7 @@ static void _find_identifiers_in_base(const GDScriptCompletionIdentifier &p_base
 							List<PropertyInfo> members;
 							scr->get_script_property_list(&members);
 							for (const PropertyInfo &E : members) {
-								if (E.usage & (PROPERTY_USAGE_CATEGORY | PROPERTY_USAGE_GROUP | PROPERTY_USAGE_SUBGROUP)) {
+								if (E.usage & (PROPERTY_USAGE_CATEGORY | PROPERTY_USAGE_GROUP | PROPERTY_USAGE_SUBGROUP | PROPERTY_USAGE_INTERNAL)) {
```
KoBeWi, line 1086: These flags are repeated a couple of times. It could be constant.

YuriSizov:
> Should it be hidden too?

Ah yes. I only checked this context menu for "Open Documentation" (which is already excluded by other checks), but not for this.

modules/gdscript/gdscript_editor.cpp
YuriSizov: We should definitely define one for `PROPERTY_USAGE_CATEGORY | PROPERTY_USAGE_GROUP | PROPERTY_USAGE_SUBGROUP`. I'll try to look into it some other time though.

YuriSizov:
Rebased (first push) and disabled "Copy Property Path" (second push)

godot windows editor dev x86_64_2024-01-26_20-53-00

AThousandShips (approved):
Tested and works well, I think the specifics can be improved on in the future but I think this is a good baseline and start

editor/editor_inspector.cpp
```diff
@@ -751,7 +755,7 @@ void EditorProperty::shortcut_input(const Ref<InputEvent> &p_event) {
 		} else if (ED_IS_SHORTCUT("property_editor/paste_value", p_event) && !is_read_only()) {
 			menu_option(MENU_PASTE_VALUE);
 			accept_event();
-		} else if (ED_IS_SHORTCUT("property_editor/copy_property_path", p_event)) {
+		} else if (ED_IS_SHORTCUT("property_editor/copy_property_path", p_event) && !internal) {
```
KoBeWi, line 758: Could be reversed (i.e. `internal` check first).

KoBeWi (approved):

akien-mga:
Thanks!

PR #87384, opened 2024-01-19
Add method to get "base" system UI color and system theme change callback.

bruvzg:
Implements `DisplayServer.get_base_color` to get default control background color (on macOS and Windows).
Implements `DisplayServer.set_system_theme_change_callback` to set theme change callback (all platforms, except web, tested on Windows 11, macOS 14.2.1, PopOS, Android 14 and iOS 17.2).

```gdscript
func _theme_upd():
	$ColorRectBack.color = DisplayServer.get_base_color()
	$ColorRectFront.color = DisplayServer.get_accent_color()

func _ready():
	DisplayServer.set_system_theme_change_callback(Callable(self, "_theme_upd"))
	_theme_upd()
```

Note: `theme_change` callback is triggered by any OS UI style related change, it's not necessarily mean that dark/light mode was switched or accent/base color changed.

### DIFF
--- a/doc/classes/DisplayServer.xml
+++ b/doc/classes/DisplayServer.xml
@@ -136,7 +136,7 @@
 			<description>
 				Displays OS native dialog for selecting files or directories in the file system.
 				Callbacks have the following arguments: [code]status: bool, selected_paths: PackedStringArray, selected_filter_index: int[/code].
-				[b]Note:[/b] This method is implemented if the display server has the [constant FEATURE_NATIVE_DIALOG] feature. Supported platforms include Linux (X11 and Wayland), Windows, and macOS.
+				[b]Note:[/b] This method is implemented if the display server has the [constant FEATURE_NATIVE_DIALOG] feature. Supported platforms include Linux (X11/Wayland), Windows, and macOS.
 				[b]Note:[/b] [param current_directory] might be ignored.
 				[b]Note:[/b] On Linux, [param show_hidden] is ignored.
 				[b]Note:[/b] On macOS, native file dialogs have no title.
@@ -161,7 +161,7 @@
 				- [code]"values"[/code] - [PackedStringArray] of values. If empty, boolean option (check box) is used.
 				- [code]"default"[/code] - default selected option index ([int]) or default boolean value ([bool]).
 				Callbacks have the following arguments: [code]status: bool, selected_paths: PackedStringArray, selected_filter_index: int, selected_option: Dictionary[/code].
-				[b]Note:[/b] This method is implemented if the display server has the [constant FEATURE_NATIVE_DIALOG] feature. Supported platforms include Linux (X11 and Wayland), Windows, and macOS.
+				[b]Note:[/b] This method is implemented if the display server has the [constant FEATURE_NATIVE_DIALOG] feature. Supported platforms include Linux (X11/Wayland), Windows, and macOS.
 				[b]Note:[/b] [param current_directory] might be ignored.
 				[b]Note:[/b] On Linux (X11), [param show_hidden] is ignored.
 				[b]Note:[/b] On macOS, native file dialogs have no title.
@@ -179,6 +179,13 @@
 			<return type="Color" />
 			<description>
 				Returns OS theme accent color. Returns [code]Color(0, 0, 0, 0)[/code], if accent color is unknown.
+				[b]Note:[/b] This method is implemented on macOS and Windows.
+			</description>
+		</method>
+		<method name="get_base_color" qualifiers="const">
+			<return type="Color" />
+			<description>
+				Returns the OS theme base color (default control background). Returns [code]Color(0, 0, 0, 0)[/code] if the base color is unknown.
 				[b]Note:[/b] This method is implemented on macOS and Windows.
 			</description>
 		</method>
@@ -1134,6 +1141,14 @@
 			<param index="0" name="filename" type="String" />
 			<description>
 				Sets the window icon (usually displayed in the top-left corner) in the operating system's [i]native[/i] format. The file at [param filename] must be in [code].ico[/code] format on Windows or [code].icns[/code] on macOS. By using specially crafted [code].ico[/code] or [code].icns[/code] icons, [method set_native_icon] allows specifying different icons depending on the size the icon is displayed at. This size is determined by the operating system and user preferences (including the display scale factor). To use icons in other formats, use [method set_icon] instead.
+			</description>
+		</method>
+		<method name="set_system_theme_change_callback">
+			<return type="void" />
+			<param index="0" name="callable" type="Callable" />
+			<description>
+				Sets the [param callable] that should be called when system theme settings are changed. Callback method should have zero arguments.
+				[b]Note:[/b] This method is implemented on Android, iOS, macOS, Windows, and Linux (X11/Wayland).
 			</description>
 		</method>
 		<method name="status_indicator_set_callback">
--- a/platform/android/java/lib/src/org/godotengine/godot/Godot.kt
+++ b/platform/android/java/lib/src/org/godotengine/godot/Godot.kt
@@ -149,6 +149,7 @@ class Godot(private val context: Context) : SensorEventListener {
 	private var useApkExpansion = false
 	private var useImmersive = false
 	private var useDebugOpengl = false
+	private var darkMode = false;
 
 	private var containerLayout: FrameLayout? = null
 	var renderView: GodotRenderView? = null
@@ -183,6 +184,8 @@ class Godot(private val context: Context) : SensorEventListener {
 			Log.d(TAG, "OnCreate already invoked")
 			return
 		}
+
+		darkMode = context.resources?.configuration?.uiMode?.and(Configuration.UI_MODE_NIGHT_MASK) == Configuration.UI_MODE_NIGHT_YES
 
 		beginBenchmarkMeasure("Startup", "Godot::onCreate")
 		try {
@@ -560,6 +563,17 @@ class Godot(private val context: Context) : SensorEventListener {
 	}
 
 	/**
+	 * Configuration change callback
+	*/
+	fun onConfigurationChanged(newConfig: Configuration) {
+		var newDarkMode = newConfig.uiMode?.and(Configuration.UI_MODE_NIGHT_MASK) == Configuration.UI_MODE_NIGHT_YES
+		if (darkMode != newDarkMode) {
+			darkMode = newDarkMode
+			GodotLib.onNightModeChanged()
+		}
+	}
+
+	/**
 	 * Activity result callback
 	 */
 	fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
@@ -731,7 +745,7 @@ class Godot(private val context: Context) : SensorEventListener {
 	 */
 	@Keep
 	private fun isDarkModeSupported(): Boolean {
-		return Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q
+		return context.resources?.configuration?.uiMode?.and(Configuration.UI_MODE_NIGHT_MASK) != Configuration.UI_MODE_NIGHT_UNDEFINED
 	}
 
 	/**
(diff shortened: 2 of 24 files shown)


doc/classes/DisplayServer.xml
```diff
@@ -142,6 +142,13 @@
 				[b]Note:[/b] This method is implemented on macOS and Windows.
 			</description>
 		</method>
+		<method name="get_base_color" qualifiers="const">
+			<return type="Color" />
+			<description>
+				Returns OS theme base color (default control background). Returns [code]Color(0, 0, 0, 0)[/code], if base color is unknown.
```
AThousandShips, line 148:
```suggestion
				Returns the OS theme base color (default control background). Returns [code]Color(0, 0, 0, 0)[/code], if the base color is unknown.
```

bruvzg:
Added theme change callback support for the rest of platforms (except web), tested Linux/X11 code on PopOS, have not tested it on Android and iOS yet.

platform/linuxbsd/freedesktop_portal_desktop.cpp
```diff
@@ -488,8 +503,45 @@ void FreeDesktopPortalDesktop::_thread_file_dialog_monitor(void *p_ud) {
 				}
 			}
 		}
+		if (bus) {
+			DBusMessage *message = dbus_message_new_method_call(
```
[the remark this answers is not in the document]
bruvzg: Changed it to use signal (but I'm not sure if there's any difference, we still need to continuously probe for signal message the same way).

platform/android/java/lib/src/org/godotengine/godot/Godot.kt
```diff
@@ -559,6 +559,13 @@ class Godot(private val context: Context) : SensorEventListener {
 		}
 	}
 
+	/**
+	 * Configuration change callback
+	*/
+	fun onConfigurationChanged(newConfig: Configuration) {
+		GodotLib.themeChanged()
```
m4gr3d, line 566: `onConfigurationChanged` is fired for every config changes we override, and we override [quite a few](https://github.com/godotengine/godot/blob/0bcc0e92b3f0ac57d4c4650722f347593a258572/platform/android/java/app/AndroidManifest.xml#L42) so this approach is not sufficient as it'll create a lot of invalid `themeChanged` callbacks.

To report changes for the theme only, we'll need to store the night mode flags on startup (e.g: [in `Godot::onCreate(...)` after the activity is retrieved](https://github.com/godotengine/godot/blob/0bcc0e92b3f0ac57d4c4650722f347593a258572/platform/android/java/lib/src/org/godotengine/godot/Godot.kt#L190), and compare the stored value with the new value retrieved from the new `Configuration` instance provided by `onConfigurationChanged`.

[Here](https://stackoverflow.com/a/44170179) is an example from SO.

platform/android/java/lib/src/org/godotengine/godot/GodotLib.java
```diff
@@ -219,6 +219,11 @@ public static native boolean initialize(Activity activity,
 	 */
 	public static native void requestPermissionResult(String p_permission, boolean p_result);
 
+	/**
+	 * Invoked on the theme change.
+	 */
+	public static native void themeChanged();
+
```
m4gr3d, line 226: For the Android implementation, let's be specific in the naming and description since this only reports changes from Dark to Light mode and vice-versa (e.g: can switch to `onNightModeChanged()`).

m4gr3d (changes requested):

platform/android/java/lib/src/org/godotengine/godot/Godot.kt
bruvzg: Done.

doc/classes/DisplayServer.xml
```diff
@@ -1095,6 +1102,14 @@
 				Sets the window icon (usually displayed in the top-left corner) in the operating system's [i]native[/i] format. The file at [param filename] must be in [code].ico[/code] format on Windows or [code].icns[/code] on macOS. By using specially crafted [code].ico[/code] or [code].icns[/code] icons, [method set_native_icon] allows specifying different icons depending on the size the icon is displayed at. This size is determined by the operating system and user preferences (including the display scale factor). To use icons in other formats, use [method set_icon] instead.
 			</description>
 		</method>
+		<method name="set_theme_change_callback">
+			<return type="void" />
+			<param index="0" name="callable" type="Callable" />
+			<description>
+				Sets the [param callable] that should be called when system theme settings are changed.
+				[b]Note:[/b] This method is implemented on Android, iOS, macOS and Windows, and Linux (X11).
```
AThousandShips, line 1110:
```suggestion
				[b]Note:[/b] This method is implemented on Android, iOS, macOS, Windows, and Linux (X11).
```

YuriSizov:
By the way, what about the accent color (on systems that support it) or the theme name or something (on systems that support _that_)? Windows has accent colors, configurable by user, and some Linux DEs have named themes which we actually have matching presets for.

bruvzg:
> There's an accent-color XDG Portal setting: https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Settings.html

This seems to be a new addition, so I'm not sure if any real DE support it, but since it's D-Bus we probably should try reading it anyway.

platform/android/java/lib/src/org/godotengine/godot/Godot.kt
```diff
@@ -184,6 +185,10 @@ class Godot(private val context: Context) : SensorEventListener {
 			return
 		}
 
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```
m4gr3d, line 188: I don't think you need to check against Android Q since according to the documentation, the night mode apis have been available since api level 8.

platform/android/java/lib/src/org/godotengine/godot/Godot.kt
```diff
@@ -559,6 +564,19 @@ class Godot(private val context: Context) : SensorEventListener {
 		}
 	}
 
+	/**
+	 * Configuration change callback
+	*/
+	fun onConfigurationChanged(newConfig: Configuration) {
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```
m4gr3d, line 571: Similar comment here, you should be able to remove the Android version check.

m4gr3d (changes requested):

platform/android/java/lib/src/org/godotengine/godot/Godot.kt
bruvzg: Removed.

platform/android/java/lib/src/org/godotengine/godot/Godot.kt
```diff
@@ -739,10 +753,7 @@ class Godot(private val context: Context) : SensorEventListener {
 	 */
 	@Keep
 	private fun isDarkMode(): Boolean {
-		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-			return context.resources?.configuration?.uiMode?.and(Configuration.UI_MODE_NIGHT_MASK) == Configuration.UI_MODE_NIGHT_YES
-		}
-		return false
+		return context.resources?.configuration?.uiMode?.and(Configuration.UI_MODE_NIGHT_MASK) == Configuration.UI_MODE_NIGHT_YES
```
m4gr3d, line 756: This should just return the `darkMode` variable; given the logic in `onConfigurationChanged`, that value will always be up-to-date.

platform/android/java/lib/src/org/godotengine/godot/Godot.kt
```diff
@@ -739,10 +753,7 @@ class Godot(private val context: Context) : SensorEventListener {
 	 */
```
m4gr3d, line 753: This is not part of your original changes, but to be consistent for the `isDarkModeSupported()` method above, we should just return:

```
return context.resources?.configuration?.uiMode?.and(Configuration.UI_MODE_NIGHT_MASK) != Configuration.UI_MODE_NIGHT_UNDEFINED
```

m4gr3d (changes requested):
Some minor feedback to address and the Android logic should be good to go!

m4gr3d (approved):
The Android section looks good!

doc/classes/DisplayServer.xml
```diff
@@ -142,6 +142,13 @@
 				[b]Note:[/b] This method is implemented on macOS and Windows.
 			</description>
 		</method>
+		<method name="get_base_color" qualifiers="const">
+			<return type="Color" />
+			<description>
+				Returns the OS theme base color (default control background). Returns [code]Color(0, 0, 0, 0)[/code], if the base color is unknown.
```
akien-mga, line 148:
```suggestion
				Returns the OS theme base color (default control background). Returns [code]Color(0, 0, 0, 0)[/code] if the base color is unknown.
```

platform/linuxbsd/x11/display_server_x11.cpp
```diff
@@ -364,6 +364,10 @@ bool DisplayServerX11::is_dark_mode() const {
 	}
 }
 
+void DisplayServerX11::set_theme_change_callback(const Callable &p_callable) {
+	portal_desktop->set_theme_change_callback(p_callable);
+}
```
akien-mga, line 369: Can the same be added for Wayland (and thus update the docs accordingly)?

akien-mga (approved):
Looks good to me!

I wonder if the "theme_change" methods/signals should be "system_theme_change" to make it clear that it's about the OS theme settings and not the Godot editor/project theme?

doc/classes/DisplayServer.xml
```diff
@@ -1120,6 +1127,14 @@
 				Sets the window icon (usually displayed in the top-left corner) in the operating system's [i]native[/i] format. The file at [param filename] must be in [code].ico[/code] format on Windows or [code].icns[/code] on macOS. By using specially crafted [code].ico[/code] or [code].icns[/code] icons, [method set_native_icon] allows specifying different icons depending on the size the icon is displayed at. This size is determined by the operating system and user preferences (including the display scale factor). To use icons in other formats, use [method set_icon] instead.
 			</description>
 		</method>
+		<method name="set_system_theme_change_callback">
+			<return type="void" />
+			<param index="0" name="callable" type="Callable" />
+			<description>
+				Sets the [param callable] that should be called when system theme settings are changed. Callback method should have zero arguments.
+				[b]Note:[/b] This method is implemented on Android, iOS, macOS, Windows, and Linux (X11).
```
akien-mga, line 1135:
```suggestion
				[b]Note:[/b] This method is implemented on Android, iOS, macOS, Windows, and Linux (X11 and Wayland).
```

akien-mga: Actually should be `(X11/Wayland)`, there's only 2 occurrences of `(X11 and Wayland)` which would also be worth changing to keep it consistent.
```
$ rg "Linux \(X11/Wayland\)" | wc -l
35
$ rg "Linux \(X11 and Wayland\)" | wc -l
2
```

akien-mga (approved):

akien-mga:
Needs a rebase.

Calinou (commented):
Tested locally on Fedora 39 KDE X11, this returns a fully transparent color for both base and accent color (so nothing is visible):

```gdscript
extends Node2D

func _theme_upd():
	print("Updating theme")
	$ColorRectBack.color = DisplayServer.get_base_color()
	print($ColorRectBack.color)
	$ColorRectFront.color = DisplayServer.get_accent_color()
	print($ColorRectFront.color)

func _ready():
	DisplayServer.set_system_theme_change_callback(Callable(self, "_theme_upd"))
	_theme_upd()
```

I'm using the Breeze Dark theme. The theme switch callback works (it's called if I switch system themes), but colors are still fully transparent.

bruvzg:
> Tested locally on Fedora 39 KDE X11

Colors are supported only on macOS and Windows, there's no standard Linux API for this.

akien-mga:
Thanks!

PR #87385, opened 2024-01-19
Fix SubViewport physics picking

### LINKED ISSUE #84258: Mouse input events not being detected from Area2D inside a SubViewport unless a key is pressed

Sauermann:
Apply the logic of `handle_input_locally` for physics picking.

resolve #84258
cherry-pick of #85665 from 4.2 to master.

I see this as an interim solution until #77926 gets reviewed.
I believe, that it is unlikely that #77926 gets reviews anytime soon, but since #84258 is quite nasty, I want to provide a bugfix-PR, that is easier to review and that already received some testing in 4.2.

### DIFF
--- a/scene/main/viewport.cpp
+++ b/scene/main/viewport.cpp
@@ -738,6 +738,14 @@ void Viewport::_process_picking() {
 
 	while (physics_picking_events.size()) {
 		local_input_handled = false;
+		if (!handle_input_locally) {
+			Viewport *vp = this;
+			while (!Object::cast_to<Window>(vp) && vp->get_parent()) {
+				vp = vp->get_parent()->get_viewport();
+			}
+			vp->local_input_handled = false;
+		}
+
 		Ref<InputEvent> ev = physics_picking_events.front()->get();
 		physics_picking_events.pop_front();
 


scene/main/viewport.cpp
```diff
@@ -738,6 +738,14 @@ void Viewport::_process_picking() {
 
 	while (physics_picking_events.size()) {
 		local_input_handled = false;
+		if (!handle_input_locally) {
+			Viewport *vp = this;
+			while (!Object::cast_to<Window>(vp) && vp->get_parent()) {
```
AThousandShips, line 743: Assuming it's guaranteed this will always be non-null?

Sauermann: Physics picking happens only when the viewport is inside the scene-tree:

https://github.com/godotengine/godot/blob/74c32faa78b54863f8f25c538083907c2bf71791/scene/main/viewport.cpp#L676-L679

So the returned viewport is also inside the scene-tree.

The same method is used in preexisting code:

https://github.com/godotengine/godot/blame/d76c1d0e516fedc535a2e394ab780cac79203477/scene/main/viewport.cpp#L3657-L3673

scene/main/viewport.cpp
```diff
@@ -738,6 +738,14 @@ void Viewport::_process_picking() {
...
+				vp = vp->get_parent()->get_viewport();
+			}
+			vp->local_input_handled = false;
+		}
+
```
adamscott, line 748: Just to be sure, you seem to skip every viewport (even ones with `local_input_handled == true` until the last one, then you set `local_input_handled` to `false`.
Is this what you want?

Sauermann: This PR intends to treat `handle_input_locally` for physics picking in the same way as it is treated in other places. That includes skipping viewports with `local_input_handled == true`.
The fact that currently `handle_input_locally` is treated differently in different parts of the code-base is the reason for the linked issue.

akien-mga (approved):

YuriSizov:
Thanks!

PR #87386, opened 2024-01-19
Add GLES3 infrastructure for lightmap baking in the compatibility backend

clayjohn:
Combined with https://github.com/godotengine/godot/pull/87340 this allows baking lightmaps from the compatibility backend as long as the device supports creating an RD context. 

This was needed to verify that https://github.com/godotengine/godot/pull/87340 works as intended. 

This PR can still be tested without merging https://github.com/godotengine/godot/pull/87340 as the functions it implements are all available through the RenderingServer API directly.

Skies are slightly brighter in the compatibility backend right now and this has an impact on lightmap brightness. The results when baking from the compatibility backend are slightly brighter than in the other backends, but the results are very close.

### DIFF
--- a/doc/classes/LightmapGI.xml
+++ b/doc/classes/LightmapGI.xml
@@ -9,7 +9,7 @@
 		[b]Performance:[/b] [LightmapGI] provides the best possible run-time performance for global illumination. It is suitable for low-end hardware including integrated graphics and mobile devices.
 		[b]Note:[/b] Due to how lightmaps work, most properties only have a visible effect once lightmaps are baked again.
 		[b]Note:[/b] Lightmap baking on [CSGShape3D]s and [PrimitiveMesh]es is not supported, as these cannot store UV2 data required for baking.
-		[b]Note:[/b] If no custom lightmappers are installed, [LightmapGI] can only be baked when using the Vulkan backend (Forward+ or Mobile), not OpenGL.
+		[b]Note:[/b] If no custom lightmappers are installed, [LightmapGI] can only be baked from devices that support the Forward+ or Mobile rendering backends.
 	</description>
 	<tutorials>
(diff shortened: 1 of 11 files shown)


Calinou (approved):
Tested locally (rebased on top of `master` + https://github.com/godotengine/godot/pull/87340), it works as expected (including when the editor is temporarily overridden to use OpenGL).

<details open>

- **OS:** Linux (Fedora 39)
</details>

___

The lightmap texture indeed looks different if baked while in Compatibility compared to Forward+:

- Baked while editor is in Forward+: [test_small.zip](https://github.com/godotengine/godot/files/14045106/test_small.zip)
- Baked while editor is in Compatibility: [test_compatibility_small.zip](https://github.com/godotengine/godot/files/14045107/test_compatibility_small.zip)

___

Baking lightmaps then exiting the editor will result in a message being printed in the terminal:

```
ERROR: 2 RID allocations of type 'N5GLES37TextureE' were leaked at exit.
```

This doesn't occur if you open a scene that has lightmaps already baked then close the editor.

___

The LightmapGI node configuration warning can be removed now:

image

The class reference note in `doc/classes/LightmapGI.xml` about Vulkan being required can also be removed.

That said, we may want to detect whether Vulkan can be run at all when creating a LightmapGI node in the editor so we can display a node configuration warning if Vulkan can't be run.

___

PS: This is really amazing work by the way – I didn't even know this was technically possible without starting a separate Godot process :slightly_smiling_face:

clayjohn:
> The LightmapGI node configuration warning can be removed now:

For now, I have just updated the warning as lots of people use Godot on devices that don't support Vulkan. That's why the warning now says that it requires a device that supports the RD backends rather than just saying its unsupported

clayjohn:
Pushed an update with the documentation update and to fix the leaking textures.

akien-mga:
Thanks!

PR #83343, opened 2023-10-14
Fix export variable of type Node pointing to a wrong child node when duplicating

warriormaster12:
Took a while to figure that this was a two part issue. 

1. properties were considered as overridden since `get_property_value`  method would return a variant that was read from a packed scene file (NodePath). However, when calling `instantiate` the variants would be converted to a proper type (Object) if they were in the [deferred_properties list](https://github.com/godotengine/godot/blob/master/scene/resource/packed_scene.cpp#L487).
2. When duplicating a node, variants of type Object would not update their path in that  process, which is fine for nodes that aren't children of the duplicated node.

before pr: 

https://github.com/godotengine/godot/assets/33091666/2d5312ab-112a-4b20-a73b-a6b6948cf4c6

pr:

https://github.com/godotengine/godot/assets/33091666/7e880a6b-8e21-4aab-a128-cbe25dcf4588

Fixes: https://github.com/godotengine/godot/issues/82670
Supersedes: https://github.com/godotengine/godot/pull/83006

PR #87387, opened 2024-01-19
Fix property of type Node duplication when script is attached to a child node

warriormaster12:
This is a continuation of a last PR where I seemed to have only partially solved the problem with node duplication as pointed out by one of the comments [after the PR got merged](https://github.com/godotengine/godot/pull/83343#issuecomment-1872988602). 

This PR should hopefully solve the edge case when a script is attached to a child node but the user duplicates that node's parent.

### DIFF
--- a/scene/main/node.cpp
+++ b/scene/main/node.cpp
@@ -2737,25 +2737,9 @@ Node *Node::_duplicate(int p_flags, HashMap<const Node *, Node *> *r_duplimap) c
 				}
 
 			} else {
-				// If property points to a node which is owned by a node we are duplicating, update its path.
-				if (value.get_type() == Variant::OBJECT) {
-					Node *property_node = Object::cast_to<Node>(value);
-					if (property_node && is_ancestor_of(property_node)) {
-						value = current_node->get_node_or_null(get_path_to(property_node));
-					}
-				} else if (value.get_type() == Variant::ARRAY) {
-					Array arr = value;
-					if (arr.get_typed_builtin() == Variant::OBJECT) {
-						for (int i = 0; i < arr.size(); i++) {
-							Node *property_node = Object::cast_to<Node>(arr[i]);
-							if (property_node && is_ancestor_of(property_node)) {
-								arr[i] = current_node->get_node_or_null(get_path_to(property_node));
-							}
-						}
-						value = arr;
-					}
+				if (value.get_type() != Variant::OBJECT && (value.get_type() != Variant::ARRAY || static_cast<Array>(value).get_typed_builtin() != Variant::OBJECT)) {
+					current_node->set(name, value);
 				}
-				current_node->set(name, value);
 			}
 		}
 	}
@@ -2770,6 +2754,8 @@ Node *Node::duplicate(int p_flags) const {
 	if (dupe && (p_flags & DUPLICATE_SIGNALS)) {
 		_duplicate_signals(this, dupe);
 	}
+
+	_duplicate_properties_node(this, this, dupe);
 
 	return dupe;
 }
@@ -2791,6 +2777,8 @@ Node *Node::duplicate_from_editor(HashMap<const Node *, Node *> &r_duplimap, con
 	// because re-targeting of connections from some descendant to another is not possible
 	// if the emitter node comes later in tree order than the receiver
 	_duplicate_signals(this, dupe);
+
+	_duplicate_properties_node(this, this, dupe);
 
 	return dupe;
 }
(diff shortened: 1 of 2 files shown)


scene/main/node.cpp
```diff
@@ -2620,7 +2619,6 @@ Node *Node::_duplicate(int p_flags, HashMap<const Node *, Node *> *r_duplimap) c
 			parent->move_child(dup, pos);
 		}
 	}
-
 	for (List<const Node *>::Element *N = node_tree.front(); N; N = N->next()) {
```
AThousandShips, line 2622:
```suggestion

	for (List<const Node *>::Element *N = node_tree.front(); N; N = N->next()) {
```
Unnecessary and unrelated whitespace change

scene/main/node.cpp
```diff
@@ -2559,7 +2559,6 @@ Node *Node::_duplicate(int p_flags, HashMap<const Node *, Node *> *r_duplimap) c
 	if (get_name() != String()) {
 		node->set_name(get_name());
 	}
-
 #ifdef TOOLS_ENABLED
```
AThousandShips, line 2562:
```suggestion

#ifdef TOOLS_ENABLED
```

scene/main/node.h
```diff
@@ -237,6 +237,7 @@ class Node : public Object {
 	void _propagate_groups_dirty();
 	Array _get_node_and_resource(const NodePath &p_path);
 
+	void _duplicate_deferred_properties(const Node *p_root, const Node *p_original, Node *p_copy) const;
```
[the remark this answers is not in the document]
warriormaster12: Not sure if it’s even a correct term to be used but I used property_utils.cpp as an example https://github.com/godotengine/godot/blob/0bcc0e92b3f0ac57d4c4650722f347593a258572/scene/property_utils.cpp#L77 

Maybe it should be called ```_duplicate_proprerties_node``` instead 🤔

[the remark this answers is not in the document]
AThousandShips: That one refers to properties that are deferred afaik, specifically, unsure what to name this one myself

warriormaster12:
@AThousandShips is it possible to get this cherry-picked for 4.2?

AThousandShips:
I don't know the area well enough to make that call

warriormaster12:
Can you confirm that this doesn’t happen with a version of godot before my PRs?

scene/main/node.cpp
```diff
@@ -2654,25 +2654,9 @@ Node *Node::_duplicate(int p_flags, HashMap<const Node *, Node *> *r_duplimap) c
...
-					if (arr.get_typed_builtin() == Variant::OBJECT) {
-						for (int i = 0; i < arr.size(); i++) {
-							Node *property_node = Object::cast_to<Node>(arr[i]);
-							if (property_node && is_ancestor_of(property_node)) {
-								arr[i] = current_node->get_node_or_null(get_path_to(property_node));
-							}
-						}
-						value = arr;
-					}
+				if (value.get_type() != Variant::OBJECT && !(value.get_type() == Variant::ARRAY && static_cast<Array>(value).get_typed_builtin() == Variant::OBJECT)) {
```
AThousandShips, line 2657:
```suggestion
				if (value.get_type() != Variant::OBJECT && (value.get_type() != Variant::ARRAY || static_cast<Array>(value).get_typed_builtin() != Variant::OBJECT)) {
```

scene/main/node.cpp
```diff
@@ -2761,6 +2749,48 @@ void Node::remap_nested_resources(Ref<Resource> p_resource, const HashMap<Ref<Re
 }
 #endif
 
+// Duplicates properties that store a Node.
+// This has to be called after nodes have been duplicated since
+// we only then get a full picture of how the duplicated node tree looks like
```
AThousandShips, line 2754:
```suggestion
// only then do we get a full picture of how the duplicated node tree looks like.
```

scene/main/node.cpp
```diff
@@ -2761,6 +2749,48 @@ void Node::remap_nested_resources(Ref<Resource> p_resource, const HashMap<Ref<Re
...
+			if (property_node && (p_root->is_ancestor_of(property_node) || property_node->get_path_to(p_root) == ".")) {
+				value = p_copy->get_node_or_null(p_original->get_path_to(property_node));
+				p_copy->set(name, value);
+			}
+		} else if (value.get_type() == Variant::ARRAY) {
+			Array arr = value;
+			if (arr.get_typed_builtin() == Variant::OBJECT) {
+				for (int i = 0; i < arr.size(); i++) {
+					Node *property_node = Object::cast_to<Node>(arr[i]);
+					if (property_node && (p_root->is_ancestor_of(property_node) || property_node->get_path_to(p_root) == ".")) {
```
AThousandShips, line 2779:
```suggestion
					if (property_node && (p_root == property_node || p_root->is_ancestor_of(property_node))) {
```

scene/main/node.cpp
```diff
@@ -2761,6 +2749,48 @@ void Node::remap_nested_resources(Ref<Resource> p_resource, const HashMap<Ref<Re
...
+			continue;
+		}
+		String name = E.name;
+		if (name == script_property_name) {
+			continue;
+		}
+		Variant value = p_original->get(name);
+		if (value.get_type() == Variant::OBJECT) {
+			Node *property_node = Object::cast_to<Node>(value);
+			if (property_node && (p_root->is_ancestor_of(property_node) || property_node->get_path_to(p_root) == ".")) {
```
AThousandShips, line 2770:
```suggestion
			if (property_node && (p_root == property_node || p_root->is_ancestor_of(property_node))) {
```

scene/main/node.cpp
```diff
@@ -2844,6 +2832,48 @@ void Node::remap_nested_resources(Ref<Resource> p_resource, const HashMap<Ref<Re
...
+void Node::_duplicate_properties_node(const Node *p_root, const Node *p_original, Node *p_copy) const {
+	List<PropertyInfo> props;
+	p_copy->get_property_list(&props);
+	StringName script_property_name = CoreStringNames::get_singleton()->_script;
+	for (const PropertyInfo &E : props) {
+		if (!(E.usage & PROPERTY_USAGE_STORAGE)) {
+			continue;
+		}
+		String name = E.name;
+		if (name == script_property_name) {
```
KoBeWi, line 2847: I think this check is irrelevant, because you are casting to Node, so a Script won't pass anyway.

KoBeWi:
There is some bug where if you duplicate a node and the duplicated node has its properties updated, the original will also have them changed.

https://github.com/godotengine/godot/assets/2223172/634e896f-e959-442e-a468-110e2b91f663

warriormaster12:
@KoBeWi Try again, value now is deeply duplicated so the bug shouldn't happen.

scene/main/node.cpp
warriormaster12: Removed the if statement

KoBeWi: Now `script_property_name` is unused. You should remove it too.

KoBeWi (approved):

scene/main/node.cpp
warriormaster12: Done, how silly of me

akien-mga:
Thanks!

PR #87388, opened 2024-01-19
Implement `LIMIT_MAX_COMPUTE_SHARED_MEMORY_SIZE` to `limit_get` in the Vulkan backend

clayjohn:
Also add a more helpful warning that is only displayed in dev builds.

Bug reported by PaperMartin on rocketchat

### DIFF
--- a/drivers/d3d12/rendering_device_driver_d3d12.cpp
+++ b/drivers/d3d12/rendering_device_driver_d3d12.cpp
@@ -6218,6 +6218,8 @@ uint64_t RenderingDeviceDriverD3D12::limit_get(Limit p_limit) {
 			return D3D12_CS_THREAD_GROUP_MAX_Y;
 		case LIMIT_MAX_COMPUTE_WORKGROUP_SIZE_Z:
 			return D3D12_CS_THREAD_GROUP_MAX_Z;
+		case LIMIT_MAX_COMPUTE_SHARED_MEMORY_SIZE:
+			return D3D12_CS_TGSM_REGISTER_COUNT * sizeof(float);
 		case LIMIT_SUBGROUP_SIZE:
 		// Note in min/max. Shader model 6.6 supports it (see https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_WaveSize.html),
 		// but at this time I don't know the implications on the transpilation to DXIL, etc.
--- a/drivers/metal/metal_device_properties.h
+++ b/drivers/metal/metal_device_properties.h
@@ -125,6 +125,7 @@ struct MetalLimits {
 	uint32_t maxVertexInputBindingStride;
 	uint32_t maxDrawIndexedIndexValue;
 	uint32_t maxShaderVaryings;
+	uint32_t maxThreadGroupMemoryAllocation;
 
 	double temporalScalerInputContentMinScale;
 	double temporalScalerInputContentMaxScale;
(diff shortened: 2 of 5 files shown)


AThousandShips:
This seems to also not be available in d3d12, unsure what limit to return though

RandomShaper:
In Direct3D 12 it's mostly fine to let the `default` kick in and return a huge value to lie on the optimistic side if there's no known way of checking the limit. Speaking of which, this PR adds a pessimistic-by-default approach to Vulkan. I believe both should do the same. Options:
a) Keep Vulkan RDD not having a `default`. When a new limit is added to the enum, the compiler would (should) complain and we can re-assess in case Vulkan has no way of checking. Otherwise, we'll be always fine.
b) Have the same `default` code in both drivers (optimistic).
c) Refactor to the base implementation is responsible for returning the optimistic value in case the specific RDD can't handle it.

clayjohn:
> In Direct3D 12 it's mostly fine to let the `default` kick in and return a huge value to lie on the optimistic side if there's no known way of checking the limit. Speaking of which, this PR adds a pessimistic-by-default approach to Vulkan. I believe both should do the same. Options: a) Keep Vulkan RDD not having a `default`. When a new limit is added to the enum, the compiler would (should) complain and we can re-assess in case Vulkan has no way of checking. Otherwise, we'll be always fine. b) Have the same `default` code in both drivers (optimistic). c) Refactor to the base implementation is responsible for returning the optimistic value in case the specific RDD can't handle it.

To be clear, the Vulkan backend has always returned `0` by default. This PR doesn't change that, it just demotes the error to a warning and only uses it in dev builds. 

If we want to maintain compatibility in the Vulkan backend we need to keep it at `0` by default. We could make the D3D12 driver do the same. 

We could also agree to break compatibility and unify the two to some other value. 

I don't think its a big deal to break compatibility for an undefined return value. So I'm happy to do either. 

I'm not sure `(1 << 30)` is a safe default value for many of the enum values though. As a lot of the limits exposed are memory limits that tend to cap out at `(1 << 14)` or  `(1 << 16)`

sakrel:
This is what Mesa's Vulkan over D3D12 translation layer maps `maxComputeSharedMemorySize` to:
`.maxComputeSharedMemorySize = D3D12_CS_TGSM_REGISTER_COUNT * sizeof(float)`
https://github.com/Mesa3D/mesa/blob/957009978ef6d7121fc0d710d03bc20097d4d46b/src/microsoft/vulkan/dzn_device.c#L1680

RandomShaper:
Considering all the above:
1. Can we then make both drivers return a safe semi-optimistic value? Say, `(1 << 14)`, or maybe `INT16_MAX`?
2. Let's apply @sakrel's findings to D3D12.

clayjohn:
Updated to also include support for Metal. I standardized all three on using `(1 << 30)` since that didn't seem to cause issues in the D3D12 backend and indeed providing a number that is too big is better than providing a 0 in most cases

Calinou (approved):
Tested locally on Linux/Vulkan, it works as expected. Code looks good to me.

```gdscript
func _ready() -> void:
	print(RenderingServer.get_video_adapter_name())
	var rd = RenderingServer.create_local_rendering_device()
	print(rd.limit_get(RenderingDevice.LIMIT_MAX_COMPUTE_SHARED_MEMORY_SIZE))

	# NVIDIA GeForce RTX 4090
	# 49152
	# The value reported matches what's shown here:
	# https://vulkan.gpuinfo.org/displayreport.php?id=37298#properties
```

For reference, `maxComputeSharedMemorySize` is always 16384 or more on Vulkan: https://vulkan.gpuinfo.org/displaydevicelimit.php?name=maxComputeSharedMemorySize

It's almost always 32768 or more on desktop platforms too, but a significant amount of Android/iOS devices are limited to 16384.

clayjohn:
> It's almost always 32768 or more on desktop platforms too, but a significant amount of Android/iOS devices are limited to 16384.

Since D3D12 requires 32768, anything that supports D3D12 should report at least 32768 in Vulkan too.

Repiteo:
Thanks!

PR #87389, opened 2024-01-19
Explicitly initialize all of `FileAccess::create_func[ACCESS_MAX]`

scgm0:
It looks like it should have three entries.

### DIFF
--- a/core/io/file_access.cpp
+++ b/core/io/file_access.cpp
@@ -38,7 +38,7 @@
 #include "core/io/marshalls.h"
 #include "core/os/os.h"
 
-FileAccess::CreateFunc FileAccess::create_func[ACCESS_MAX] = { nullptr, nullptr };
+FileAccess::CreateFunc FileAccess::create_func[ACCESS_MAX] = {};
 
 FileAccess::FileCloseFailNotify FileAccess::close_fail_notify = nullptr;
 


RandomShaper:
According to the curent style, this is correct.

Question for language lawyers: would this initialization of two items when there should be three let the third one be garbage or will it be guaranteed to be null anyway?

Suggestion (for Prod team or for discussion): get rid of all the zero-initializations of `static`/`thread_local` variables, since they are already zero-initialized implicitly.

RandomShaper (approved):
I'm approving, but Production has to decide if this should be squashed with maybe other small fixes by the author.

core/io/file_access.cpp
```diff
@@ -38,7 +38,7 @@
 #include "core/io/marshalls.h"
 #include "core/os/os.h"
 
-FileAccess::CreateFunc FileAccess::create_func[ACCESS_MAX] = { nullptr, nullptr };
+FileAccess::CreateFunc FileAccess::create_func[ACCESS_MAX] = { nullptr, nullptr, nullptr };
```
akien-mga, line 41: Could be initialized like this to be future proof.
```suggestion
FileAccess::CreateFunc FileAccess::create_func[ACCESS_MAX] = {};
```

akien-mga:
> Suggestion (for Prod team or for discussion): get rid of all the zero-initializations of `static`/`thread_local` variables, since they are already zero-initialized implicitly.

I wasn't aware that `static` or `thread_local` variables are zero-initialized implicitly. Is that guaranteed by the standard?
Is there a cost to initializing them explicitly anyway? Otherwise I tend to err on the side of explicitness, so nobody is taken aback if they don't know the specifics of `static`/`thread_local` default initialization, compared to other regular member variables which we try to systematically initialize.

RandomShaper:
> Is that guaranteed by the standard?

Yes.

> Is there a cost to initializing them explicitly anyway?

Not at the machine level.

> Otherwise I tend to err on the side of explicitness, so nobody is taken aback if they don't know the specifics of `static`/`thread_local` default initialization, compared to other regular member variables which we try to systematically initialize.

As always, I'll happily follow any standard in the written and unwritten Godot style guides. However, I'd want to give a different perspective on this. When one is declaring something `static`, let alone `thread_local`, they are already expected to know the implications. If they can't understand such data is placed somewhere else and, in general, is subject to different laws, they are already in trouble. Knowing that default-initialization (boiling down to zero-initializtion as needed) is a thing in them is just another little piece of the same puzzle and any decision about where to draw the boundary runs the risk of being arbitrary.

core/io/file_access.cpp
akien-mga: Amended the commit with this.

akien-mga (approved):
As mentioned by @RandomShaper, we may want to go the other way and remove explicit zero-initializations for static members. But for now let's follow the existing style and make the init explicit.

akien-mga:
Thanks!

PR #87390, opened 2024-01-20
Add basic multimesh data needed for headless export to the Dummy rendering server

### LINKED ISSUE #86396: MultiMesh buffers are sometimes reset to 0 on export
### Tested versions

Found in v4.2.1.stable.official, v4.1.2.stable.official and v4.1.1.stable.official

### System information

Godot v4.2.1.stable - Manjaro Linux - X11 - Vulkan (Forward+) - integrated AMD Radeon Vega 10 Graphics (RADV RAVEN) - AMD Ryzen 7 3700U with Radeon Vega Mobile Gfx (8 Threads)

### Issue description

I have a project where some levels contain multiple large-ish MultiMeshes (> 10000 instances), which were populated by a tool script and saved to the level's scene file.  These MultiMeshes always render correctly when I run the game from the editor, however, they sometimes don't render correctly when I try to export the game.  After some investigation I found that all the values in the `buffer` for said MultiMeshes had been reset to 0 in the exported build, so I think it is something to do with how the MultiMesh data is getting packed on export.

Unfortunately I haven't been able to reproduce this in an MRP (although I have included one that contains the offending MultiMeshes in case anyone else can).  What makes it worse is that it seems to happen fairly randomly and is not specific to one version of Godot - in fact, after first triggering the bug I was able to consistently reproduce it in a previous version of my game (that had exported with no issues just a few months prior) using the same version of Godot that had been used to export said version of the game (4.1.1).

I was able to work-around the issue with a simple helper script that saves the instance transforms to a custom file which re-applies said transforms when the scene is loaded.

### Steps to reproduce

1. Generate several MultiMeshes with a large number of instances (10000+) (I'm currently unsure how many are needed to trigger the bug)
2. Run the project from the editor
3. Export the project

### Expected behaviour
The MultiMeshes render correctly both when running from the editor and in the exported build of the project.  See the grass in the screenshot below is visible:
working

### Actual behaviour
The MultiMeshes render correctly when running from the editor, but sometimes after exporting the project the MultiMeshes disappear as a result of having zero-ed/null data in the buffer.  See the grass in the screenshot below is now missing:
not working

[Multimesh Buffer Bug.zip](https://github.com/godotengine/godot/files/13743566/Multimesh.Buffer.Bug.zip)

clayjohn:
Fixes: https://github.com/godotengine/godot/issues/86396

Multimesh relies on storing and retrieving the buffer in the rendering server to function. So we need to have at least that information in the RenderingServer. 

This PR adds the bare minimum amount of data to store and retrieve the multimesh buffer for export.

### DIFF
--- a/servers/rendering/dummy/storage/mesh_storage.cpp
+++ b/servers/rendering/dummy/storage/mesh_storage.cpp
@@ -63,3 +63,33 @@ void MeshStorage::mesh_clear(RID p_mesh) {
 
 	m->surfaces.clear();
 }
+
+RID MeshStorage::multimesh_allocate() {
+	return multimesh_owner.allocate_rid();
+}
+
+void MeshStorage::multimesh_initialize(RID p_rid) {
+	multimesh_owner.initialize_rid(p_rid, DummyMultiMesh());
+}
+
+void MeshStorage::multimesh_free(RID p_rid) {
+	DummyMultiMesh *multimesh = multimesh_owner.get_or_null(p_rid);
+	ERR_FAIL_NULL(multimesh);
+
+	multimesh_owner.free(p_rid);
+}
+
+void MeshStorage::multimesh_set_buffer(RID p_multimesh, const Vector<float> &p_buffer) {
+	DummyMultiMesh *multimesh = multimesh_owner.get_or_null(p_multimesh);
+	ERR_FAIL_NULL(multimesh);
+	multimesh->buffer.resize(p_buffer.size());
+	float *cache_data = multimesh->buffer.ptrw();
+	memcpy(cache_data, p_buffer.ptr(), p_buffer.size() * sizeof(float));
+}
+
+Vector<float> MeshStorage::multimesh_get_buffer(RID p_multimesh) const {
+	DummyMultiMesh *multimesh = multimesh_owner.get_or_null(p_multimesh);
+	ERR_FAIL_NULL_V(multimesh, Vector<float>());
+
+	return multimesh->buffer;
+}
--- a/servers/rendering/dummy/storage/mesh_storage.h
+++ b/servers/rendering/dummy/storage/mesh_storage.h
@@ -50,6 +50,12 @@ private:
 
 	mutable RID_Owner<DummyMesh> mesh_owner;
 
+	struct DummyMultiMesh {
+		PackedFloat32Array buffer;
+	};
+
+	mutable RID_Owner<DummyMultiMesh> multimesh_owner;
+
 public:
 	static MeshStorage *get_singleton() {
 		return singleton;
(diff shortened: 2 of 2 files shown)


BastiaanOlij (approved):
lgtm but I do think we need to think if we want to do more here. Yes this fixes the issue with data being lost while we want to keep it because we're exporting, but it is an awful lot of data to potentially keep in memory when the headless server is used to host a multiplayer game (depending on the architecture used off course).

Possibly we could add a project setting for the dummy renderer "keep mesh data" that defaults to true, and can be turned off?

BastiaanOlij:
Also slightly off topic, but the editor actually has a few places where we also retrieve mesh data from the rendering server. As we're getting data from the GPU that produces a stall point as we need to wait for the GPU to finish everything scheduled and wait for the data to be retrieved.

I've often wondered if it doesn't make sense to have an option to keep the data on the CPU instead of tossing it after transfering the data to the GPU. That should be an opt-in thing, maybe even on mesh level, but it could be worth thinking about.

clayjohn:
> Also slightly off topic, but the editor actually has a few places where we also retrieve mesh data from the rendering server. As we're getting data from the GPU that produces a stall point as we need to wait for the GPU to finish everything scheduled and wait for the data to be retrieved.
> 
> I've often wondered if it doesn't make sense to have an option to keep the data on the CPU instead of tossing it after transfering the data to the GPU. That should be an opt-in thing, maybe even on mesh level, but it could be worth thinking about.

We cache it in the DummyRenderingServer right now (which is totally on the CPU) https://github.com/godotengine/godot/blob/74c32faa78b54863f8f25c538083907c2bf71791/servers/rendering/dummy/storage/mesh_storage.h#L72

We've been discussing with smix8 and reduz about having optional caching for when the mesh needs to be read-back regularly (I cced you on the thread)

clayjohn:
> lgtm but I do think we need to think if we want to do more here. Yes this fixes the issue with data being lost while we want to keep it because we're exporting, but it is an awful lot of data to potentially keep in memory when the headless server is used to host a multiplayer game (depending on the architecture used off course).
> 
> Possibly we could add a project setting for the dummy renderer "keep mesh data" that defaults to true, and can be turned off?

That sounds like a good idea to explore in general for the dummy renderer as mesh data is going to be the biggest culprit by far and it is currently always cached as well

YuriSizov:
Thanks!

YuriSizov:
> That sounds like a good idea to explore in general for the dummy renderer as mesh data is going to be the biggest culprit by far and it is currently always cached as well

I interpreted this as "let's merge this for now, but look into the suggestion in future". 🙃

akien-mga:
Cherry-picked for 4.2.2.

PR #76121, opened 2023-04-16
Fix missing shader params in headless export 

cgsdev0:
Fixes https://github.com/godotengine/godot/issues/66842

> When exporting a project using --headless --export-pack, any shader parameters/uniforms specified in the inspector are not included in the exported project. All shader parameters are reverted to their default values (or null).

I'm not confident this change is 100% correct; I did some testing with my project where I first observed the behavior, and the change did resolve the issue.

This is the function we used to call previously:
https://github.com/godotengine/godot/blob/a7276f1ce0c2911216a2c4718efddab98ddffd8f/scene/resources/shader.cpp#L117-L141

I don't really understand what this group logic is doing, and I think my change potentially loses that information. Definitely open to feedback, this is my first time exploring the Godot codebase 😄 

(This change isn't needed for 3.x)

PR #87392, opened 2024-01-20
Add DummyShader handling to Dummy RenderingServer to ensure shader parameters are saved in headless export

### LINKED ISSUE #66842: Shader parameters specified in the inspector are not included when exporting a project in headless mode
### Godot version

4.0.beta2.official.f8745f2f7

### System information

Void linux

### Issue description

When exporting a project using `--headless --export-pack`, any shader parameters/uniforms specified in the inspector are not included in the exported project. All shader parameters are reverted to their default values (or null).

### Steps to reproduce

Project setup:
1. Create a cube mesh.
2. Add a shader material to the cube with the following simple shader:
```
shader_type spatial;
render_mode unshaded;

uniform vec4 color: source_color = vec4(0.0, 1.0, 0.0, 0.0); // Green

void fragment() {
        ALBEDO = color.rgb;
}
```
3. In the inspector, set the 'color' shader parameter/uniform to be a red color.
4. When running the project from the editor, the cube is now red.

Export without `--headless` option:
```
rm -rf .godot
./Godot_v4.0-beta2_linux.x86_64 --editor --export-pack Linux/X11 export.zip
./Godot_v4.0-beta2_linux.x86_64 --main-pack export.zip
# Cube is red (correct behavior)
```

Export with `--headless` option:
```
rm -rf .godot
./Godot_v4.0-beta2_linux.x86_64 --editor --headless --export-pack Linux/X11 export_headless.zip
./Godot_v4.0-beta2_linux.x86_64 --main-pack export_headless.zip
# Cube is green (incorrect behavior)
```

The behavior can be reproduced on Godot 4.0 alpha 16 through Godot 4.0 beta 2.

[ShaderParameterHeadless.zip](https://github.com/godotengine/godot/files/9700232/ShaderParameterHeadless.zip)

clayjohn:
Supersedes: https://github.com/godotengine/godot/pull/76121

Fixes https://github.com/godotengine/godot/issues/66842 (including the MRP in https://github.com/godotengine/godot/pull/76121#issuecomment-1721927547)

This introduces the minimal set of logic needed by `get_shader_parameter_list()` which is used by materials to figure out what shader parameters need to be exported. 

We can get away with a single shader compiler for all shader types since we don't care about the generated code, we just need the uniforms.

Also, we need to validate the existence of `RendererSceneOcclusionCull::get_singleton()` in `RendererSceneCull::free()` now since that gets called to free certain shaders.

### DIFF
--- a/servers/rendering/dummy/storage/material_storage.h
+++ b/servers/rendering/dummy/storage/material_storage.h
@@ -31,13 +31,31 @@
 #ifndef MATERIAL_STORAGE_DUMMY_H
 #define MATERIAL_STORAGE_DUMMY_H
 
+#include "servers/rendering/shader_compiler.h"
+#include "servers/rendering/shader_language.h"
 #include "servers/rendering/storage/material_storage.h"
 #include "servers/rendering/storage/utilities.h"
 
 namespace RendererDummy {
 
 class MaterialStorage : public RendererMaterialStorage {
+private:
+	static MaterialStorage *singleton;
+
+	struct DummyShader {
+		HashMap<StringName, ShaderLanguage::ShaderNode::Uniform> uniforms;
+	};
+
+	mutable RID_Owner<DummyShader> shader_owner;
+
+	ShaderCompiler dummy_compiler;
+
 public:
+	static MaterialStorage *get_singleton() { return singleton; }
+
+	MaterialStorage();
+	~MaterialStorage();
+
 	/* GLOBAL SHADER UNIFORM API */
 
 	virtual void global_shader_parameter_add(const StringName &p_name, RS::GlobalShaderParameterType p_type, const Variant &p_value) override {}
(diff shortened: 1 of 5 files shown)


servers/rendering/dummy/storage/material_storage.h
```diff
@@ -31,13 +31,33 @@
...
+	};
+
+	mutable RID_Owner<DummyShader> shader_owner;
+
+	ShaderCompiler dummy_compiler;
+
 public:
+	static MaterialStorage *get_singleton() {
+		return singleton;
+	};
```
AThousandShips, line 56:
```suggestion
	}
```

clayjohn: Hmmm, this is just copy-pasted from MeshStorage. It seems that all the Dummy rendering servers use this style. I can just fix all three at once I guess

BastiaanOlij (approved):
LGTM, interesting use case, I think we would have run into this in the XR team as we're working on using CI to build APKs for standalone VR devices.

YuriSizov:
Thanks!

nikitalita:
The dummy renderer sets `low_end` to true. Won't this fail to compile the shaders if it uses high-end built-in functions and such?

nikitalita:
Yup:

```
SHADER ERROR: Built-in function "findLSB(int)" is only supported on high-end platforms.
          at: (null) (:26)
ERROR: Shader compilation failed.
   at: shader_set_code (servers/rendering/dummy/storage/material_storage.cpp:90)
```

And when it fails to compile, it obliterates the shader parameters, too:

image

akien-mga:
Cherry-picked for 4.2.2.

PR #87393, opened 2024-01-20
Use `os_unfair_lock` on Apple platforms

stuartcarnie:
Specialise the `SpinLock` implementation on Apple platforms to use an `os_unfair_lock`, which is more efficient and often recommended by Apple for games and other realtime applications. `os_unfair_lock` is particularly important for Apple Silicon, as it ensures that performance (P) cores are not tied up in busy loops and handles priority inversion issues between threads.

Apple discusses their use in the context of game engines and real-time apps regularly in their videos and tech talks:

* [priority inversion](https://developer.apple.com/videos/play/tech-talks/110147?time=1782)
* [handling input events for visionOS](https://developer.apple.com/wwdc23/10089?time=1172)
* [busy loops stealing P cores](https://developer.apple.com/wwdc20/10214?time=1637)

    > On Apple Silicon Macs, busy-waiting can have the effect of pointlessly occupying the P cores, causing an overall delay in completion of the entire work. You should prefer a synchronization primitives that block when they can't make progress, NSLock, os_unfair_lock, pthread mutexes, are all examples of blocking locks.

`os_unfair_lock` also appears regularly in the context of Metal frameworks, such as this call stack:

### DIFF
--- a/core/os/spin_lock.h
+++ b/core/os/spin_lock.h
@@ -33,6 +33,25 @@
 
 #include "core/typedefs.h"
 
+#if defined(__APPLE__)
+
+#include <os/lock.h>
+
+class SpinLock {
+	mutable os_unfair_lock _lock = OS_UNFAIR_LOCK_INIT;
+
+public:
+	_ALWAYS_INLINE_ void lock() const {
+		os_unfair_lock_lock(&_lock);
+	}
+
+	_ALWAYS_INLINE_ void unlock() const {
+		os_unfair_lock_unlock(&_lock);
+	}
+};
+
+#else
+
 #include <atomic>
 
 class SpinLock {
@@ -48,5 +67,7 @@ public:
 		locked.clear(std::memory_order_release);
 	}
 };
+
+#endif // __APPLE__
 
 #endif // SPIN_LOCK_H


core/os/spin_lock.h
```diff
@@ -49,4 +68,6 @@ class SpinLock {
 	}
 };
 
+#endif
```
AThousandShips, line 71:
```suggestion
#endif // __APPLE__
```

stuartcarnie: Done 👍🏻

RandomShaper:
I'm curious about something: do you know (or could you find out) how `std::mutex` is implemented on Apple platforms?

stuartcarnie:
I will see if I can learn anything.

stuartcarnie:
> I'm curious about something: do you know (or could you find out) how `std::mutex` is implemented on Apple platforms?

Looks like they are implemented using pthread's locking APIs. This call stack:



steps into

RandomShaper:
Interesting. Thanks for checking it!

RandomShaper (approved):
Despite Godot was diverging from having platform-specific implementations of threading stuff, the`SpinLock` is still custom enough and if there's a superior implementation for certain platforms, I believe it's worth having it.

To be honest, I don't know for sure how superior it is to the current one we have (especially, considering there's an in-progress PR that will improve it), but in this case it's very unlikely this PR won't be beneficial.

PR #87414, opened 2024-01-20
RayCast2D and RayCast3D: warn to check `is_colliding` before `get_collision_point` and `get_collision_normal`

### LINKED ISSUE #84642: RayCast3D doesn't update collision point when calling force_raycast_update()
### Godot version

4.1.3.stable

### System information

Windows 11, Intel Core i7

### Issue description

When updating a RayCast3D node's position via script, getting the collision point after calling force_raycast_update() returns the same point as prior to moving the node. 

### Steps to reproduce

Run sample project
Press Space
Results are printed to console.

[RayCast Bug.zip](https://github.com/godotengine/godot/files/13302785/RayCast.Bug.zip)

rburing:
Adds warnings to check `is_colliding` first, in the documentation of `get_collision_point` and `get_collision_normal`.

The warning is phrased in the positive, making no guarantees about what happens if `is_colliding` returns `false`. In that case, if using Godot Physics, the vectors will be zero if the ray never collided, or they will contain the values from the previous time when there was a collision, but I think this implementation detail is not worth documenting. (It may be worth adding an error for that case in the future.)

Also moved some adjacent information from a note to the main text, since the main text was just incomplete without it.

- Closes https://github.com/godotengine/godot/issues/84642

### DIFF
--- a/doc/classes/RayCast2D.xml
+++ b/doc/classes/RayCast2D.xml
@@ -69,13 +69,14 @@
 			<return type="Vector2" />
 			<description>
 				Returns the normal of the intersecting object's shape at the collision point, or [code]Vector2(0, 0)[/code] if the ray starts inside the shape and [member hit_from_inside] is [code]true[/code].
+				[b]Note:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned normal is valid and up-to-date.
 			</description>
 		</method>
 		<method name="get_collision_point" qualifiers="const">
 			<return type="Vector2" />
 			<description>
-				Returns the collision point at which the ray intersects the closest object. If [member hit_from_inside] is [code]true[/code] and the ray starts inside of a collision shape, this function will return the origin point of the ray.
-				[b]Note:[/b] This point is in the [b]global[/b] coordinate system.
+				Returns the collision point at which the ray intersects the closest object, in the global coordinate system. If [member hit_from_inside] is [code]true[/code] and the ray starts inside of a collision shape, this function will return the origin point of the ray.
+				[b]Note:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned point is valid and up-to-date.
 			</description>
 		</method>
 		<method name="is_colliding" qualifiers="const">
--- a/doc/classes/RayCast3D.xml
+++ b/doc/classes/RayCast3D.xml
@@ -76,13 +76,14 @@
 			<return type="Vector3" />
 			<description>
 				Returns the normal of the intersecting object's shape at the collision point, or [code]Vector3(0, 0, 0)[/code] if the ray starts inside the shape and [member hit_from_inside] is [code]true[/code].
+				[b]Note:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned normal is valid and up-to-date.
 			</description>
 		</method>
 		<method name="get_collision_point" qualifiers="const">
 			<return type="Vector3" />
 			<description>
-				Returns the collision point at which the ray intersects the closest object. If [member hit_from_inside] is [code]true[/code] and the ray starts inside of a collision shape, this function will return the origin point of the ray.
-				[b]Note:[/b] This point is in the [b]global[/b] coordinate system.
+				Returns the collision point at which the ray intersects the closest object, in the global coordinate system. If [member hit_from_inside] is [code]true[/code] and the ray starts inside of a collision shape, this function will return the origin point of the ray.
+				[b]Note:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned point is valid and up-to-date.
 			</description>
 		</method>
 		<method name="is_colliding" qualifiers="const">


doc/classes/RayCast2D.xml
```diff
@@ -69,13 +69,14 @@
...
 				Returns the normal of the intersecting object's shape at the collision point, or [code]Vector2(0, 0)[/code] if the ray starts inside the shape and [member hit_from_inside] is [code]true[/code].
+				[b]Warning:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned normal is meaningful and up-to-date.
 			</description>
 		</method>
 		<method name="get_collision_point" qualifiers="const">
 			<return type="Vector2" />
 			<description>
-				Returns the collision point at which the ray intersects the closest object.
-				[b]Note:[/b] This point is in the [b]global[/b] coordinate system.
+				Returns the collision point at which the ray intersects the closest object, in the global coordinate system.
```
Sauermann, line 78:
```suggestion
				Returns the collision point at which the ray intersects the closest object, in the canvas layer coordinate system.
```
As discussed in https://github.com/godotengine/godot-proposals/issues/3866, "global" can be confusing in 2D. Since this PR changes this line, I believe it best to make this more precise.

rburing: I agree that would be more precise, but it is more confusing if the `RayCast2D` is not under a `CanvasLayer` node, so I think the explanation should cover that (very common) case as well. I know it's layer 0 internally or something like that, but I'm not convinced the user (whose `RayCast2D` is probably not under a `CanvasLayer` node) will know. I didn't plan to go into this here, but if you want you can make a suggestion addressing this.

Sauermann: Unfortunately I don't have a better idea - it's a complex topic.
If this isn't feasible for now, it can easily be deferred to a later point in time. I would have considered this just nice to have in the context of this PR.

rburing: So I agree it could be improved, but I would prefer to leave it for another PR.

doc/classes/RayCast2D.xml
```diff
@@ -69,13 +69,14 @@
 			<return type="Vector2" />
 			<description>
 				Returns the normal of the intersecting object's shape at the collision point, or [code]Vector2(0, 0)[/code] if the ray starts inside the shape and [member hit_from_inside] is [code]true[/code].
+				[b]Warning:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned normal is meaningful and up-to-date.
```
Mickeon, line 72: Warnings are reserved for more "dangerous" operations. Saying both "_meaningful_" and "_up-to-date_" is a bit redundant, and the former is too up to interpretation, so it could be changed.
```suggestion
				[b]Note:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned normal is valid and up-to-date.
```

doc/classes/RayCast2D.xml
```diff
@@ -69,13 +69,14 @@
...
 			</description>
 		</method>
 		<method name="get_collision_point" qualifiers="const">
 			<return type="Vector2" />
 			<description>
-				Returns the collision point at which the ray intersects the closest object.
-				[b]Note:[/b] This point is in the [b]global[/b] coordinate system.
+				Returns the collision point at which the ray intersects the closest object, in the global coordinate system.
+				[b]Warning:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned point is meaningful and up-to-date.
```
Mickeon, line 79: Ditto
```suggestion
				[b]Note:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned point is valid and up-to-date.
```

doc/classes/RayCast3D.xml
```diff
@@ -76,13 +76,14 @@
 			<return type="Vector3" />
 			<description>
 				Returns the normal of the intersecting object's shape at the collision point, or [code]Vector3(0, 0, 0)[/code] if the ray starts inside the shape and [member hit_from_inside] is [code]true[/code].
+				[b]Warning:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned normal is meaningful and up-to-date.
```
Mickeon, line 79: Ditto
```suggestion
				[b]Note:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned normal is valid and up-to-date.
```

doc/classes/RayCast3D.xml
```diff
@@ -76,13 +76,14 @@
...
 			</description>
 		</method>
 		<method name="get_collision_point" qualifiers="const">
 			<return type="Vector3" />
 			<description>
-				Returns the collision point at which the ray intersects the closest object.
-				[b]Note:[/b] This point is in the [b]global[/b] coordinate system.
+				Returns the collision point at which the ray intersects the closest object, in the global coordinate system.
+				[b]Warning:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned point is meaningful and up-to-date.
```
Mickeon, line 86: Ditto
```suggestion
				[b]Note:[/b] Check that [method is_colliding] returns [code]true[/code] before calling this method to ensure the returned point is valid and up-to-date.
```

Mickeon (changes requested):

Mickeon:
Also needs to be rebased. The part about global coordinates is handled more nicely in this PR, by being part of the same paragraph instead of a separate note.

Mickeon (approved):

akien-mga:
Thanks!

PR #87417, opened 2024-01-20
Use callable for `TreeItem` custom draw

AThousandShips:
Old behavior is deprecated

### DIFF
--- a/doc/classes/TreeItem.xml
+++ b/doc/classes/TreeItem.xml
@@ -160,6 +160,13 @@
 				Returns the custom color of column [param column].
 			</description>
 		</method>
+		<method name="get_custom_draw_callback" qualifiers="const">
+			<return type="Callable" />
+			<param index="0" name="column" type="int" />
+			<description>
+				Returns the custom callback of column [param column].
+			</description>
+		</method>
 		<method name="get_custom_font" qualifiers="const">
 			<return type="Font" />
 			<param index="0" name="column" type="int" />
@@ -553,13 +560,23 @@
 				Sets the given column's custom color.
 			</description>
 		</method>
-		<method name="set_custom_draw">
+		<method name="set_custom_draw" is_deprecated="true">
 			<return type="void" />
 			<param index="0" name="column" type="int" />
 			<param index="1" name="object" type="Object" />
 			<param index="2" name="callback" type="StringName" />
 			<description>
 				Sets the given column's custom draw callback to [param callback] method on [param object].
+				The [param callback] should accept two arguments: the [TreeItem] that is drawn and its position and size as a [Rect2].
+				[i]Deprecated.[/i] Use [method TreeItem.set_custom_draw_callback] instead.
+			</description>
+		</method>
+		<method name="set_custom_draw_callback">
+			<return type="void" />
+			<param index="0" name="column" type="int" />
+			<param index="1" name="callback" type="Callable" />
+			<description>
+				Sets the given column's custom draw callback. Use an empty [Callable] ([code skip-lint]Callable()[/code]) to clear the custom callback.
 				The [param callback] should accept two arguments: the [TreeItem] that is drawn and its position and size as a [Rect2].
 			</description>
 		</method>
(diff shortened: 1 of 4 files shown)


AThousandShips:
Could also add `get_custom_draw_object` and `get_custom_draw_callback` to fetch the callback from the method, but starting out like this

KoBeWi:
tbh custom draw should be a Callable, but it's too late to change it now I guess.

AThousandShips:
Agreed, we can add a new method for it and deprecate this one, it would still work fine with the new feature as it can still use the old method, but I think adding this won't hurt and does allow for easier clearing up

AThousandShips:
Gonna look at reworking this for a callable as well as allowing clearing

doc/classes/TreeItem.xml
```diff
@@ -553,14 +560,24 @@
...
 				The [param callback] should accept two arguments: the [TreeItem] that is drawn and its position and size as a [Rect2].
+				[i]Deprecated.[/i] Use [method TreeItem.set_custom_draw_callback] instead.
+			</description>
+		</method>
+		<method name="set_custom_draw_callback">
+			<return type="void" />
+			<param index="0" name="column" type="int" />
+			<param index="1" name="callback" type="Callable" />
+			<description>
+				Sets the given column's custom draw callback. Call with [code skip-lint]Callable()[/code] to clear the custom callback.
```
AThousandShips, line 579: Unsure about the exact wording here but think relevant to point out you can clear it and how to do so

KoBeWi: Maybe
```suggestion
				Sets the given column's custom draw callback. Use empty [Callable] ([code skip-lint]Callable()[/code]) to clear the custom callback.
```

scene/gui/tree.cpp
```diff
@@ -2297,11 +2319,8 @@ int Tree::draw_item(const Point2i &p_pos, const Point2 &p_draw_ofs, const Size2
 
 				} break;
 				case TreeItem::CELL_MODE_CUSTOM: {
-					if (p_item->cells[i].custom_draw_obj.is_valid()) {
-						Object *cdo = ObjectDB::get_instance(p_item->cells[i].custom_draw_obj);
-						if (cdo) {
-							cdo->call(p_item->cells[i].custom_draw_callback, p_item, Rect2(item_rect));
-						}
+					if (p_item->cells[i].custom_draw_callback.is_valid()) {
+						p_item->cells[i].custom_draw_callback.call(p_item, Rect2(item_rect));
```
KoBeWi, line 2323: If the provided callback can't be called (due to argument mismatch etc.), this will fail silently. We could print an error when that happens.

Though it's a pre-existing problem, so it can be fixed later.

AThousandShips: I'll take a look at fixing it here 🙂

AThousandShips: Attempted an improvement of this part

doc/classes/TreeItem.xml
AThousandShips: Will include in my next push

scene/gui/tree.cpp
```diff
@@ -2297,10 +2319,15 @@ int Tree::draw_item(const Point2i &p_pos, const Point2 &p_draw_ofs, const Size2
...
-							cdo->call(p_item->cells[i].custom_draw_callback, p_item, Rect2(item_rect));
+					if (p_item->cells[i].custom_draw_callback.is_valid()) {
+						Variant args[] = { p_item, Rect2(item_rect) };
+						const Variant *argptrs[] = { &args[0], &args[1] };
+
+						Callable::CallError ce;
+						Variant ret;
+						p_item->cells[i].custom_draw_callback.callp(argptrs, 2, ret, ce);
+						if (ce.error != Callable::CallError::CALL_OK) {
+							ERR_PRINT("Error calling custom draw method: " + Variant::get_callable_error_text(p_item->cells[i].custom_draw_callback, argptrs, 2, ce) + ".");
```
AThousandShips, line 2330: Haven't worked much with this specific method for callbacks so unsure if this is the cleanest way to do it

KoBeWi (approved):

YuriSizov (approved):

YuriSizov:
Thanks!

PR #87419, opened 2024-01-20
Fix shortcut name consistency in SceneTreeDock

KoBeWi:
The new "Show in FileSystem" in SceneTreeDock uses different casing than others, making it recognized as unique translation string.


ryevdokimov (commented):
You can find a few of these using regex. Here are a couple more I found just in the same file. 

https://github.com/godotengine/godot/blob/2bbbccb17bb9498b4a8ce1a4773a2076b47f17c7/editor/scene_tree_dock.cpp#L663-L666

I checked this particular usage, and it follows the convention of minor words starting lower-case, so it would probably be good to throw in this PR too. 

image

ryevdokimov (approved):

akien-mga (approved):

akien-mga:
Thanks!

PR #87421, opened 2024-01-20
GDScript: Use autoload singleton name in `GDScriptDocGen`

### LINKED ISSUE #8441: Facilitate class documentation of project singletons (or, allow hyperlinks to class docs using arbitrary text)

A solar system simulation @ivoyager that is distributed as a set of editor plugins.

Links to project or plugin singletons in class documentation are awkward. My plugin has an autoload singleton 'IVGlobal'. Unfortunately, I can't link it in class documentation using `[IVGlobal]`. Instead, I have to link it throughout class documentation using `["addons/ivoyager_core/singletons/global.gd"]`. Or I can give it `class_name IVGlobalSingleton` and then link it with that name. But that's still clunky. It's especially problematic for _this_ class, as I need to reference it very often in other class docs.

Compare with core singletons, which can be linked, e.g., `[PhysicsServer3D]`. [Edit: You can't do this in GDScript because you can't assign class_name with the same value as the autoload name.]

It should be possible (by some means) to generate a link in class docs that has the name of the autoload and links to its class doc. In particular, this would improve documentation for plugins that use singletons.

Option 1: Allow a singleton to have class_name that is the same as its autoload name. This currently throws error: "Class XXXX hides an autoload singleton". (I have no clue whether this is technically feasible.)

Option 2: Add tag `[autoload <AutoloadName>]` to the docs system. This would resolve as a link <AutoloadName> in the docs (with the autoload name, not the class name) and point to the autoload class doc (whatever its actual class name).

Option 3 (most general): Analogous to the `url` tag, which does work in class docs, we could have a tag that allows linking to class docs using arbitrary text. This could be: `[class=<ClassName>]arbitrary text[/class]`.

My personal preference would be both 2 and 3. Option 2 is "cleaner" for solving this specific problem. However, option 3 could be used in many other cases (well, hyperlinks are like that).

No.

It's core function, not accessible to GDScript.

dalexeev:
* Closes godotengine/godot-proposals#8441.

You can't use the autoload singleton name as a class name, so I think it's safe to use the autoload name instead of the script path in the docs. If the autoload script also has a global name, then it will be used first.

[test-autoload-doc.zip](https://github.com/godotengine/godot/files/13999143/test-autoload-doc.zip)

### DIFF
--- a/modules/gdscript/editor/gdscript_docgen.cpp
+++ b/modules/gdscript/editor/gdscript_docgen.cpp
@@ -32,21 +32,29 @@
 
 #include "../gdscript.h"
 
-String GDScriptDocGen::_get_script_path(const String &p_path) {
+#include "core/config/project_settings.h"
+
+HashMap<String, String> GDScriptDocGen::singletons;
+
+String GDScriptDocGen::_get_script_name(const String &p_path) {
+	const HashMap<String, String>::ConstIterator E = singletons.find(p_path);
+	if (E) {
+		return E->value;
+	}
 	return p_path.trim_prefix("res://").quote();
 }
 
 String GDScriptDocGen::_get_class_name(const GDP::ClassNode &p_class) {
 	const GDP::ClassNode *curr_class = &p_class;
 	if (!curr_class->identifier) { // All inner classes have an identifier, so this is the outer class.
-		return _get_script_path(curr_class->fqcn);
+		return _get_script_name(curr_class->fqcn);
 	}
 
 	String full_name = curr_class->identifier->name;
 	while (curr_class->outer) {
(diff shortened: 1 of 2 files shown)


adamscott (approved):
Reviewed during the GDScript meeting. The solution is somewhat not ideal, but it's the best that can be done.

All resolves to the fact that you cannot name the autoload the same name as a `class_name`, which is bad because if you name then your class `AutoloadSingleton`, all the docs would refer to (i.e.) `AutoloadSingleton` instead of the actual way to use the class, which is (i.e.) `Autoload`.

Thanks @dalexeev !

akien-mga:
Thanks!

PR #87181, opened 2024-01-14
Overhaul Quaternion documentation

Mickeon:
Closes https://github.com/godotengine/godot-docs/issues/6271

This PR aims to completely overhaul the **Quaternion** type.
The reasoning and changelog are extremely similar to the "[Overhaul Basis Documentation](https://github.com/godotengine/godot/pull/87175)" PR, so I'm not sure I should bother repeating them twice.

#### More specific notes:
- Specify that the quaternions in Godot use the **Hamilton** convention, not JPL;
    - Closes https://github.com/godotengine/godot-docs/issues/6271. I still am very, very unsure on what it means, but for users that do care, they can now know.
- Mention https://quaternions.online/ and https://github.com/iwatake2222/rotation_master as point of reference;
    - _I hope these websites use the **Hamilton** convention, it would be a shame if they don't_.
- Mention that quaternions are 32-bit by default.
- Mention "Basis to Quaternion" conversion caveats.
- Made the extra line about not messing with the quaternions' components a note
- ...

-------------------

I started this ASAP after seeing the "overwhelmingly" positive reception of the prior PR after a few mere hours. Feedback is very, very welcome. Quaternions are a very complicated concept to digest and comprehend.

###### This PR may contain multiple versions of the same line I was not sure about, denoted with "??". Reviewing of these lines is heavily appreciated.

Sources:
- https://quaternions.online/
- https://fzheng.me/2017/11/12/quaternion_conventions_en/

PR #87422, opened 2024-01-20
Quaternion Documentation: Describe edge cases

Bobert-Brockson:
Working on the [Rust GDExt](https://github.com/godot-rust/gdext) library I became aware of how the edge cases for quaternions are handled. I believe exposing this may help future library authors/maintainers work with the engine.


AThousandShips:
See also:
* https://github.com/godotengine/godot/pull/87181

doc/classes/Quaternion.xml
```diff
@@ -93,7 +94,7 @@
 			<return type="float" />
 			<description>
 				Returns the angle of the rotation represented by this quaternion.
-				[b]Note:[/b] The quaternion must be normalized.
+				[b]Note:[/b] The quaternion must be normalized. Returns [code]0[/code] if the quaternion has a length and is not normalized. Returns [code]PI[code] if the quaternion has no length and is not normalized.
```
AThousandShips, line 97:
```suggestion
				[b]Note:[/b] The quaternion must be normalized. Returns [code]0[/code] if the quaternion has a length and is not normalized. Returns [code]PI[/code] if the quaternion has no length and is not normalized.
```

0awful:
Having become aware of #87181. I'm not seeing any specific discrepancies between the two. However one is a significantly larger changeset. I believe this PR, being the smaller one, should bear the responsibility of handling merge conflicts.

doc/classes/Quaternion.xml
```diff
@@ -39,7 +39,8 @@
 			<param index="0" name="axis" type="Vector3" />
 			<param index="1" name="angle" type="float" />
 			<description>
-				Constructs a quaternion that will rotate around the given axis by the specified angle. The axis must be a normalized vector.
+				Constructs a quaternion that will rotate around the given axis by the specified angle.
+				[b]Note:[/b] The axis must be a normalized vector. Returns [code](0, 0, 0, 1)[/code] if your vector is not normalized in Debug mode. Returns the quaternion that would have been generated by normalizing the axis in release mode. If the axis has length [code]0[/code] and therefore cannot be normalized returns [code](0, 0, 0, 0)[/code]
```
Mickeon, line 43: I do trust that this is all correct, as it is way out of my expertise. I **heavily** appreciate the extra detail.

However, the way this entire note and others are structured in this PR is extremely inconsistent and kind of "not good" for several reasons:
- `(0, 0, 0, 1)` is the **IDENTITY** Quaternion and should be referred to as such (`Returns [constant IDENTITY]`) to point users in the right direction, as it is not an arbitrary, magic value.
- "Debug" in "debug mode" is never capitalized across the entire class reference. Saying "debug builds" is also more common in this context.
- The whole note is extremely verbose. "Debug build" and "release build" are mutually exclusive, so there's almost no need to define both.
    - It also feels like the behavior in debug builds is the exception and not the norm, as the following sentence implies nothing wrong really happens when the axis is not normalized:
    _"Returns the quaternion that would have been generated by normalizing the axis in release mode."_
    This means that it's only worth noting the more "unusual" behavior in debug builds.
    - There's also little point in saying "_and therefore cannot be normalized_" because users interacting with a Quaternion should already have a decent understanding of vector math.

0awful: I appreciate the feedback. I'll make those modifications. Great work on your documentation improvements by the way.

doc/classes/Quaternion.xml
```diff
@@ -164,7 +166,7 @@
 			<param index="1" name="weight" type="float" />
 			<description>
 				Returns the result of the spherical linear interpolation between this quaternion and [param to] by amount [param weight].
-				[b]Note:[/b] Both quaternions must be normalized.
+				[b]Note:[/b] Both quaternions must be normalized. Returns [code](0, 0, 0, 1)[/code] if any quaternion is not normalized in debug mode. Returns [code]self[/code] if any quaternion is not normalized in release mode.
```
Mickeon, line 169: `self` is a GDScript-exclusive context and it only applies to Object, because all other Variant types cannot be extended.

In this situation the class reference likes to say, for example "Returns this quaternion unchanged" or "Returns the quaternion without changes" or similar.

Mickeon:
https://github.com/godotengine/godot/pull/87181 has been merged a while ago. Is this PR still necessary? Could it be rebased?

doc/classes/Quaternion.xml
mhilbrunner, line 43:
```suggestion
				[b]Note:[/b] The axis must be a normalized vector. Returns [code](0, 0, 0, 1)[/code] if your vector is not normalized in Debug mode. Returns the quaternion that would have been generated by normalizing the axis in release mode. If the axis has length [code]0[/code] and therefore cannot be normalized returns [code](0, 0, 0, 0)[/code].
```

mhilbrunner:
@0awful Any chance of a rebase and adding the missing "."? 🙏

Bobert-Brockson:
I stepped away from the internet for a while. Having returned I've resolved the merge conflicts. But due to the age these peculiarities that occurred may no longer be true. Similarly the requested changes have not been performed.

Updated to this state for anyone with desire to carry this further. I cannot commit to revalidating my discoveries at this time, as such I am updating this to draft.

PR #84840, opened 2023-11-13
Simplify `Control` internal transform calculation

kleonc:
```
T // translation
R // rotation
S // scaling

Control._get_internal_transform() = T(pivot_offset) * R(rotation) * S(scale) * T(-pivot_offset)
```

Previously `T(pivot_offset) * R(rotation) * S(scale) * T(-pivot_offset)` was calculated by unnecessarily doing 1 affine inverse ($`T(pivot\_offset) = T^{-1}(-pivot\_offset)`$) and 2 full Transform2D multiplications.

Now `T(pivot_offset) * R(rotation) * S(scale)` is constructed directly, and `T(-pivot_offset)` is applied locally with a dedicated method (not performing full Transform2D multiplication).

Precision of the result is not improved / the same as before because even though before a full inverse were calculated and more multiplications / additions were done, it was just some superfluous multiplications by `1.0` / additions of `0.0 * something`, hence not introducing precision errors.

PR #87432, opened 2024-01-21
Fix issue where `set_global_position(global_position)` in `Control` resulted in a different result than `global_position`

### LINKED ISSUE #87354: Control.set_global_position(global_position) doesn't work after rotated around custom pivot
### Tested versions

Godot v4.2.stable

### System information

Godot v4.2.stable - macOS 13.6.1 - GLES3 (Compatibility) - ANGLE (AMD, ANGLE Metal Renderer: AMD Radeon Pro 560X, 版本13.6.1（版号22G313）) - Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz (12 Threads)

### Issue description

When a Control's pivot offset is set to somewhere other than (0, 0) and rotated, set_global_position(global_position) will move the Control far away. 

### Steps to reproduce

```GDScript
extends ColorRect

func _ready():
    pivot_offset = size
    rotation_degrees = 90
    for i in 36:
        set_global_position(global_position)
        await get_tree().process_frame
        print('%s %s %s'%[position, global_position, size])

```

N/A

Rindbee:
Previously, the case where the pivot point was not at the origin was ignored.

Fix #87354.

### DIFF
--- a/scene/gui/control.cpp
+++ b/scene/gui/control.cpp
@@ -1419,13 +1419,15 @@ void Control::_set_global_position(const Point2 &p_point) {
 
 void Control::set_global_position(const Point2 &p_point, bool p_keep_offsets) {
 	ERR_MAIN_THREAD_GUARD;
-	Transform2D inv;
 
-	if (data.parent_canvas_item) {
-		inv = data.parent_canvas_item->get_global_transform().affine_inverse();
+	Transform2D global_transform_cache = get_global_transform();
+	if (p_point == global_transform_cache.get_origin()) {
+		return; // Edge case, but avoids calculation.
 	}
 
-	set_position(inv.xform(p_point), p_keep_offsets);
+	Point2 internal_position = global_transform_cache.affine_inverse().xform(p_point);
+
+	set_position(internal_position + data.pos_cache, p_keep_offsets);
 }
 
 Point2 Control::get_global_position() const {


scene/gui/control.cpp
```diff
@@ -1425,7 +1425,18 @@ void Control::set_global_position(const Point2 &p_point, bool p_keep_offsets) {
 		inv = data.parent_canvas_item->get_global_transform().affine_inverse();
 	}
 
-	set_position(inv.xform(p_point), p_keep_offsets);
+	Point2 new_position;
+	if (data.pivot_offset.is_zero_approx() || (data.scale.is_equal_approx(Vector2(1.0f, 1.0f)) && Math::is_zero_approx(Math::fmod((double)data.rotation, Math_TAU)))) {
```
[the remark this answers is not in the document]
Rindbee: Thanks for your suggestion.

scene/gui/control.cpp
```diff
@@ -1419,13 +1419,29 @@ void Control::_set_global_position(const Point2 &p_point) {
...
 	}
 
-	set_position(inv.xform(p_point), p_keep_offsets);
+	Point2 new_position;
+	if (data.pivot_offset.is_zero_approx() || (data.scale.is_equal_approx(Vector2(1.0, 1.0)) && Math::is_zero_approx(Math::fmod(data.rotation, (real_t)Math_TAU)))) {
+		// The pivot point is at the origin or simply translated.
+		new_position = inv.xform(p_point);
+	} else {
+		inv = Transform2D(0.0, -get_position()) * inv; // Translation transformation in parent space.
+		inv = _get_internal_transform().affine_inverse() * inv; // Internal transformation relative to pivot point.
```
kleonc, line 1439: This is equivalent to `inv = get_transform().affine_inverse() * inv;` as:
```
   _get_internal_transform().affine_inverse() * Transform2D(0.0f, -get_position())
== _get_internal_transform().affine_inverse() * Transform2D(0.0f, get_position()).affine_inverse()
== (Transform2D(0.0f, get_position()) * _get_internal_transform()).affine_inverse()
== get_transform().affine_inverse()
```
(see #84840)

Further, given `inv` is initially the inverse of the parent global transform:
```
   get_transform().affine_inverse() * parent_global_transform.affine_inverse()
== (parent_global_transform * get_transform()).affine_inverse()
== get_global_transform().affine_inverse()
```

So the whole calculation could be simplified like:
```diff
diff --git a/scene/gui/control.cpp b/scene/gui/control.cpp
index 349102fafb..c352dd4a14 100644
--- a/scene/gui/control.cpp
+++ b/scene/gui/control.cpp
@@ -1419,13 +1419,8 @@ void Control::_set_global_position(const Point2 &p_point) {
 
 void Control::set_global_position(const Point2 &p_point, bool p_keep_offsets) {
 	ERR_MAIN_THREAD_GUARD;
-	Transform2D inv;
-
-	if (data.parent_canvas_item) {
-		inv = data.parent_canvas_item->get_global_transform().affine_inverse();
-	}
-
-	set_position(inv.xform(p_point), p_keep_offsets);
+	Point2 new_position = get_global_transform().affine_inverse().xform(p_point) + get_position();
+	set_position(new_position, p_keep_offsets);
 }
 
 Point2 Control::get_global_position() const {
```

Not sure if it needs a separate case for no pivot offset / (no rotation + unit scale).

scene/gui/control.cpp
```diff
@@ -1419,13 +1419,29 @@ void Control::_set_global_position(const Point2 &p_point) {
...
+
 	Transform2D inv;
 
 	if (data.parent_canvas_item) {
 		inv = data.parent_canvas_item->get_global_transform().affine_inverse();
 	}
 
-	set_position(inv.xform(p_point), p_keep_offsets);
+	Point2 new_position;
+	if (data.pivot_offset.is_zero_approx() || (data.scale.is_equal_approx(Vector2(1.0, 1.0)) && Math::is_zero_approx(Math::fmod(data.rotation, (real_t)Math_TAU)))) {
```
kleonc, line 1434: BTW such check for the rotation would fail in case it's almost `TAU` etc. like `359.999999` degrees (it could be close to no rotation but not near-zero after `fmod`ding it).

Alternative could be e.g. `wrap`ping it into `[-PI, PI]` instead (and then checking only for near-zero). Or just adding here an additional check for near-`TAU`. Not sure if/what is preferred.

scene/gui/control.cpp
Rindbee: > Not sure if it needs a separate case for no pivot offset / (no rotation + unit scale).

If using its own `get_global_transform()`, no distinction is needed. The `global_transform` might already be cached.

scene/gui/control.cpp
```diff
@@ -1419,13 +1419,15 @@ void Control::_set_global_position(const Point2 &p_point) {
 
 void Control::set_global_position(const Point2 &p_point, bool p_keep_offsets) {
 	ERR_MAIN_THREAD_GUARD;
-	Transform2D inv;
 
-	if (data.parent_canvas_item) {
-		inv = data.parent_canvas_item->get_global_transform().affine_inverse();
+	Transform2D global_transform = get_global_transform();
```
YuriSizov, line 1423: You need to rename it to avoid shadowing. Use `xform` for example.

kleonc (approved):
AFAICT it works fine now (I've done some testing).

kleonc:
Shouldn't be cherry-picked, it introduced a regression: #89497.

PR #87334, opened 2024-01-18
Overhaul Transform3D documentation

Mickeon:
Related to https://github.com/godotengine/godot/pull/87181

This PR aims to completely overhaul the **Transform3D** type's documentation.
The reasoning and changelog are extremely similar to the "[Overhaul Basis Documentation](https://github.com/godotengine/godot/pull/87175)" PR, so I'm not sure I should bother repeating them twice.

#### More specific notes: 
- Elaborate on `orthonormalized`, `inverse`, and `affline_inverse` a lot more.
- Use global scope `lerp`'s description as a base for `interpolate_with`.
- ...

PR #87440, opened 2024-01-21
Overhaul Node3D documentation

### LINKED ISSUE #87896: Global position losing axis value when assigning axis values separately
### Tested versions

-Reproducible in Godot 4.2.1

### System information

Godot v4.2.1.stable - Fedora Linux 39 (Workstation Edition) - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4060 Ti (nvidia; 535.154.05) - AMD Ryzen 7 5800X3D 8-Core Processor (16 Threads)

### Issue description

I tried to generate a grid of boxes. Assigning the z position immediately after assigning x caused the x value to default to 0, same is true for the inverted setup or assigning z -> y -> x. All values except x are 0.

### Steps to reproduce

```GDScript
extends Node3D

@export var width: int = 10
@export var depth: int = 10

var box = preload("res://Scenes/Box.tscn")

func _ready():
	for z in depth:
		for x in width:
			var boxInst: Node = box.instantiate()
			boxInst.global_position.x = x-(width>>1)
			boxInst.global_position.z = z-(depth>>1)
			add_child(boxInst)
```

[MRP.zip](https://github.com/godotengine/godot/files/14151255/MRP.zip)

A workaround is to do something in between assigning the values. Example:
```GDScript
extends Node3D

@export var width: int = 10
@export var depth: int = 10

var box = preload("res://Scenes/Box.tscn")

func _ready():
	for z in depth:
		for x in width:
			var boxInst: Node = box.instantiate()
			boxInst.global_position.x = x-(width>>1)
			add_child(boxInst)
			boxInst.global_position.z = z-(depth>>1)
```

### LINKED ISSUE #93651: Node3D objects will not render when changing scale
### Tested versions

v4.3.beta2.official [b75f0485b]

### System information

Godot v4.3.beta2 - macOS 14.5.0 - Vulkan (Mobile) - integrated Apple M1 - Apple M1 (8 Threads)

### Issue description

MeshInstance3D objects will not render when changing SecondNode3D scale.
Code when hidden
```
mesh_instance.scale = Vector3(0.6, 0.6, 0)
```
But when I use the following code, it will render
```
mesh_instance.scale = Vector3(0.6, 0.6, 0.6)
```

I'm not sure if this is correct

### Steps to reproduce

```
extends Node3D

@onready var mesh_instance = $Camera3D/Node3D

func _ready() -> void:
	change_size()

func change_size():
	mesh_instance.scale = Vector3(0.6, 0.6, 0)
```

Scene tree structure:

Mickeon:
Closes https://github.com/godotengine/godot/issues/87896
Closes https://github.com/godotengine/godot/issues/93651
May address https://github.com/godotengine/godot-docs/issues/9930 (?)

Related to [several other past efforts](https://github.com/godotengine/godot/pulls?q=is%3Apr+author%3AMickeon+overhaul+), but especially the [Basis](https://github.com/godotengine/godot/pull/87175) and [Transform3D](https://github.com/godotengine/godot/pull/87334) ones.

This PR aims to completely overhaul **Node3D**'s class reference in an attempt to make it more easy to digest, as well as bring on par with the rest of the documentation.

Why? It generally suffers from the same issues highlighted in the past. The most notable of which being that a lot of descriptions are dull, succinct, or put the focus on the wrong subject:
- _"Set subgizmo selection for this node in the editor."_
- _"Updates all the [**Node3D**] gizmos attached to this node."_
- _"Transforms [param global_point] from world space to this node's local space."_
- _"Rotates the local transformation around the X axis by angle in radians."_

**More specific notes:**

- Remove any leftover references to "Spatial" nodes from the class reference.
- Modified `to_local` and `to_global`'s descriptions to imply they **do** return something.
- Elaborate on when visibility_changed is emitted.
    - The sentence is almost word-for-word from the way some NOTIFICATION constants are written in the **Node** class. 
- Revamp the descriptions of **NOTIFICATION_TRANSFORM_CHANGED** & similar.
    - _"In order for [constant NOTIFICATION_TRANSFORM_CHANGED] to work, users first need to ask for it, with [method set_notify_local_transform]."_
    a grand majority of **Node3D**s will receive NOTIFICATION_TRANSFORM_CHANGED without needing for the "user" to opt-in. They need this otherwise they would not work.

--------------

This PR may contain commented XML. I use it as a point of reference or to symbolize alternatives of the same sentence. Even if this is a draft, feedback on things worth mentioning is still very welcome.

### DIFF
--- a/doc/classes/Node3D.xml
+++ b/doc/classes/Node3D.xml
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<class name="Node3D" inherits="Node" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
+<class name="Node3D" inherits="Node" keywords="spatial" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
 	<brief_description>
-		Most basic 3D game object, parent of all 3D-related nodes.
+		Base object in 3D space, inherited by all 3D nodes.
 	</brief_description>
 	<description>
-		Most basic 3D game object, with a [Transform3D] and visibility settings. All other 3D game objects inherit from [Node3D]. Use [Node3D] as a parent node to move, scale, rotate and show/hide children in a 3D project.
-		Affine operations (rotate, scale, translate) happen in parent's local coordinate system, unless the [Node3D] object is set as top-level. Affine operations in this coordinate system correspond to direct affine operations on the [Node3D]'s transform. The word local below refers to this coordinate system. The coordinate system that is attached to the [Node3D] object itself is referred to as object-local coordinate system.
-		[b]Note:[/b] Unless otherwise specified, all methods that have angle parameters must have angles specified as [i]radians[/i]. To convert degrees to radians, use [method @GlobalScope.deg_to_rad].
-		[b]Note:[/b] Be aware that "Spatial" nodes are now called "Node3D" starting with Godot 4. Any Godot 3.x references to "Spatial" nodes refer to "Node3D" in Godot 4.
+		The [Node3D] node is the base representation of a node in 3D space. All other 3D nodes inherit from this class.
+		Affine operations (translation, rotation, scale) are calculated in the coordinate system relative to the parent, unless the [Node3D]'s [member top_level] is [code]true[/code]. In this coordinate system, affine operations correspond to direct affine operations on the [Node3D]'s [member transform]. The term [i]parent space[/i] refers to this coordinate system. The coordinate system that is attached to the [Node3D] itself is referred to as object-local coordinate system, or [i]local space[/i].
+		[b]Note:[/b] Unless otherwise specified, all methods that need angle parameters must receive angles in [i]radians[/i]. To convert degrees to radians, use [method @GlobalScope.deg_to_rad].
+		[b]Note:[/b] In Godot 3 and older, [Node3D] was named [i]Spatial[/i].
 	</description>
 	<tutorials>
 		<link title="Introduction to 3D">$DOCS_URL/tutorials/3d/introduction_to_3d.html</link>
@@ -18,32 +18,33 @@
 			<return type="void" />
 			<param index="0" name="gizmo" type="Node3DGizmo" />
 			<description>
-				Attach an editor gizmo to this [Node3D].
-				[b]Note:[/b] The gizmo object would typically be an instance of [EditorNode3DGizmo], but the argument type is kept generic to avoid creating a dependency on editor classes in [Node3D].
+				Attaches the given [param gizmo] to this node. Only works in the editor.
+				[b]Note:[/b] [param gizmo] should be an [EditorNode3DGizmo]. The argument type is [Node3DGizmo] to avoid depending on editor classes in [Node3D].
 			</description>
 		</method>
 		<method name="clear_gizmos">
 			<return type="void" />
 			<description>
-				Clear all gizmos attached to this [Node3D].
+				Clears all [EditorNode3DGizmo] objects attached to this node. Only works in the editor.
 			</description>
 		</method>
 		<method name="clear_subgizmo_selection">
 			<return type="void" />
 			<description>
-				Clears subgizmo selection for this node in the editor. Useful when subgizmo IDs become invalid after a property change.
+				Deselects all subgizmos for this node. Useful to call when the selected subgizmo may no longer exist after a property change. Only works in the editor.
 			</description>
 		</method>
 		<method name="force_update_transform">
 			<return type="void" />
 			<description>
-				Forces the transform to update. Transform changes in physics are not instant for performance reasons. Transforms are accumulated and then set. Use this if you need an up-to-date transform when doing physics operations.
+				Forces the node's [member global_transform] to update, by sending [constant NOTIFICATION_TRANSFORM_CHANGED]. Fails if the node is not inside the tree.
+				[b]Note:[/b] For performance reasons, transform changes are usually accumulated and applied [i]once[/i] at the end of the frame. The update propagates through [Node3D] children, as well. Therefore, use this method only when you need an up-to-date transform (such as during physics operations).
 			</description>
 		</method>
 		<method name="get_gizmos" qualifiers="const">
 			<return type="Node3DGizmo[]" />
 			<description>
-				Returns all the gizmos attached to this [Node3D].
+				Returns all the [EditorNode3DGizmo] objects attached to this node. Only works in the editor.
 			</description>
 		</method>
 		<method name="get_global_transform_interpolated">
@@ -57,14 +58,15 @@
 		<method name="get_parent_node_3d" qualifiers="const">
 			<return type="Node3D" />
 			<description>
-				Returns the parent [Node3D], or [code]null[/code] if no parent exists, the parent is not of type [Node3D], or [member top_level] is [code]true[/code].
-				[b]Note:[/b] Calling this method is not equivalent to [code]get_parent() as Node3D[/code], which does not take [member top_level] into account.
+				Returns the parent [Node3D] that directly affects this node's [member global_transform]. Returns [code]null[/code] if no parent exists, the parent is not a [Node3D], or [member top_level] is [code]true[/code].
+				[b]Note:[/b] This method is not always equivalent to [method Node.get_parent], which does not take [member top_level] into account.
 			</description>
 		</method>
 		<method name="get_world_3d" qualifiers="const">
 			<return type="World3D" />
 			<description>
-				Returns the current [World3D] resource this [Node3D] node is registered to.
+				Returns the [World3D] this node is registered to.
+				Usually, this is the same as the world used by this node's viewport (see [method Node.get_viewport] and [method Viewport.find_world_3d]).
 			</description>
 		</method>
 		<method name="global_rotate">
@@ -72,53 +74,54 @@
 			<param index="0" name="axis" type="Vector3" />
 			<param index="1" name="angle" type="float" />
 			<description>
-				Rotates the global (world) transformation around axis, a unit [Vector3], by specified angle in radians. The rotation axis is in global coordinate system.
+				Rotates this node's [member global_basis] around the global [param axis] by the given [param angle], in radians. This operation is calculated in global space (relative to the world) and preserves the [member global_position].
 			</description>
 		</method>
 		<method name="global_scale">
 			<return type="void" />
 			<param index="0" name="scale" type="Vector3" />
 			<description>
-				Scales the global (world) transformation by the given [Vector3] scale factors.
+				Scales this node's [member global_basis] by the given [param scale] factor. This operation is calculated in global space (relative to the world) and preserves the [member global_position].
+				[b]Note:[/b] This method is not to be confused with the [member scale] property.
 			</description>
 		</method>
 		<method name="global_translate">
 			<return type="void" />
 			<param index="0" name="offset" type="Vector3" />
 			<description>
-				Moves the global (world) transformation by [Vector3] offset. The offset is in global coordinate system.
+				Adds the given translation [param offset] to the node's [member global_position] in global space (relative to the world).
 			</description>
 		</method>
 		<method name="hide">
 			<return type="void" />
 			<description>
-				Disables rendering of this node. Changes [member visible] to [code]false[/code].
+				Prevents this node from being rendered. Equivalent to setting [member visible] to [code]false[/code]. This is the opposite of [method show].
 			</description>
 		</method>
 		<method name="is_local_transform_notification_enabled" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns whether node notifies about its local transformation changes. [Node3D] will not propagate this by default.
+				Returns [code]true[/code] if the node receives [constant NOTIFICATION_LOCAL_TRANSFORM_CHANGED] whenever [member transform] changes. This is enabled with [method set_notify_local_transform].
 			</description>
 		</method>
 		<method name="is_scale_disabled" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns whether this node uses a scale of [code](1, 1, 1)[/code] or its local transformation scale.
+				Returns [code]true[/code] if this node's [member global_transform] is automatically orthonormalized. This results in this node not appearing distorted, as if its global scale were set to [constant Vector3.ONE] (or its negative counterpart). See also [method set_disable_scale] and [method orthonormalize].
+				[b]Note:[/b] [member transform] is not affected by this setting.
 			</description>
 		</method>
 		<method name="is_transform_notification_enabled" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns whether the node notifies about its global and local transformation changes. [Node3D] will not propagate this by default.
+				Returns [code]true[/code] if the node receives [constant NOTIFICATION_TRANSFORM_CHANGED] whenever [member global_transform] changes. This is enabled with [method set_notify_transform].
 			</description>
 		</method>
 		<method name="is_visible_in_tree" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns [code]true[/code] if the node is present in the [SceneTree], its [member visible] property is [code]true[/code] and all its ancestors are also visible. If any ancestor is hidden, this node will not be visible in the scene tree.
-				Visibility is checked only in parent nodes that inherit from [Node3D]. If the parent is of any other type (such as [Node], [AnimationPlayer], or [Node2D]), it is assumed to be visible.
-				[b]Note:[/b] This method does not take [member VisualInstance3D.layers] into account, so even if this method returns [code]true[/code], the node might end up not being rendered.
+				Returns [code]true[/code] if this node is inside the scene tree and the [member visible] property is [code]true[/code] for this node and all of its [Node3D] ancestors [i]in sequence[/i]. An ancestor of any other type (such as [Node] or [Node2D]) breaks the sequence. See also [method Node.get_parent].
+				[b]Note:[/b] This method cannot take [member VisualInstance3D.layers] into account, so even if this method returns [code]true[/code], the node may not be rendered.
 			</description>
 		</method>
 		<method name="look_at">
@@ -127,12 +130,11 @@
 			<param index="1" name="up" type="Vector3" default="Vector3(0, 1, 0)" />
 			<param index="2" name="use_model_front" type="bool" default="false" />
 			<description>
-				Rotates the node so that the local forward axis (-Z, [constant Vector3.FORWARD]) points toward the [param target] position.
+				Rotates the node so that the local forward axis (-Z, [constant Vector3.FORWARD]) points toward the [param target] position. This operation is calculated in global space (relative to the world).
 				The local up axis (+Y) points as close to the [param up] vector as possible while staying perpendicular to the local forward axis. The resulting transform is orthogonal, and the scale is preserved. Non-uniform scaling may not work correctly.
-				The [param target] position cannot be the same as the node's position, the [param up] vector cannot be zero.
-				The [param target] and the [param up] cannot be [constant Vector3.ZERO], and shouldn't be colinear to avoid unintended rotation around local Z axis.
-				Operations take place in global space, which means that the node must be in the scene tree.
+				The [param target] position cannot be the same as the node's position, the [param up] vector cannot be [constant Vector3.ZERO]. Furthermore, the direction from the node's position to the [param target] position cannot be parallel to the [param up] vector, to avoid unintended rotation around the local Z axis.
 				If [param use_model_front] is [code]true[/code], the +Z axis (asset front) is treated as forward (implies +X is left) and points toward the [param target] position. By default, the -Z axis (camera forward) is treated as forward (implies +X is right).
+				[b]Note:[/b] This method fails if the node is not in the scene tree. If necessary, use [method look_at_from_position] instead.
 			</description>
 		</method>
 		<method name="look_at_from_position">
@@ -142,13 +144,13 @@
 			<param index="2" name="up" type="Vector3" default="Vector3(0, 1, 0)" />
 			<param index="3" name="use_model_front" type="bool" default="false" />
 			<description>
-				Moves the node to the specified [param position], and then rotates the node to point toward the [param target] as per [method look_at]. Operations take place in global space.
+				Moves the node to the specified [param position], then rotates the node to point toward the [param target] position, similar to [method look_at]. This operation is calculated in global space (relative to the world).
 			</description>
 		</method>
 		<method name="orthonormalize">
 			<return type="void" />
 			<description>
-				Resets this node's transformations (like scale, skew and taper) preserving its rotation and translation by performing Gram-Schmidt orthonormalization on this node's [Transform3D].
+				Orthonormalizes this node's [member basis]. This method sets this node's [member scale] to [constant Vector3.ONE] (or its negative counterpart), but preserves the [member position] and [member rotation]. See also [method Transform3D.orthonormalized].
 			</description>
 		</method>
 		<method name="rotate">
@@ -156,7 +158,7 @@
 			<param index="0" name="axis" type="Vector3" />
 			<param index="1" name="angle" type="float" />
 			<description>
-				Rotates the local transformation around axis, a unit [Vector3], by specified angle in radians.
+				Rotates this node's [member basis] around the [param axis] by the given [param angle], in radians. This operation is calculated in parent space (relative to the parent) and preserves the [member position].
 			</description>
 		</method>
 		<method name="rotate_object_local">
@@ -164,69 +166,74 @@
 			<param index="0" name="axis" type="Vector3" />
 			<param index="1" name="angle" type="float" />
 			<description>
-				Rotates the local transformation around axis, a unit [Vector3], by specified angle in radians. The rotation axis is in object-local coordinate system.
+				Rotates this node's [member basis] around the [param axis] by the given [param angle], in radians. This operation is calculated in local space (relative to this node) and preserves the [member position].
 			</description>
 		</method>
 		<method name="rotate_x">
 			<return type="void" />
 			<param index="0" name="angle" type="float" />
 			<description>
-				Rotates the local transformation around the X axis by angle in radians.
+				Rotates this node's [member basis] around the X axis by the given [param angle], in radians. This operation is calculated in parent space (relative to the parent) and preserves the [member position].
 			</description>
 		</method>
 		<method name="rotate_y">
 			<return type="void" />
 			<param index="0" name="angle" type="float" />
 			<description>
-				Rotates the local transformation around the Y axis by angle in radians.
+				Rotates this node's [member basis] around the Y axis by the given [param angle], in radians. This operation is calculated in parent space (relative to the parent) and preserves the [member position].
 			</description>
 		</method>
 		<method name="rotate_z">
 			<return type="void" />
 			<param index="0" name="angle" type="float" />
 			<description>
-				Rotates the local transformation around the Z axis by angle in radians.
+				Rotates this node's [member basis] around the Z axis by the given [param angle], in radians. This operation is calculated in parent space (relative to the parent) and preserves the [member position].
 			</description>
 		</method>
 		<method name="scale_object_local">
 			<return type="void" />
 			<param index="0" name="scale" type="Vector3" />
 			<description>
-				Scales the local transformation by given 3D scale factors in object-local coordinate system.
+				Scales this node's [member basis] by the given [param scale] factor. This operation is calculated in local space (relative to this node) and preserves the [member position].
 			</description>
 		</method>
 		<method name="set_disable_scale">
 			<return type="void" />
 			<param index="0" name="disable" type="bool" />
 			<description>
-				Sets whether the node uses a scale of [code](1, 1, 1)[/code] or its local transformation scale. Changes to the local transformation scale are preserved.
+				If [code]true[/code], this node's [member global_transform] is automatically orthonormalized. This results in this node not appearing distorted, as if its global scale were set to [constant Vector3.ONE] (or its negative counterpart). See also [method is_scale_disabled] and [method orthonormalize].
+				[b]Note:[/b] [member transform] is not affected by this setting.
 			</description>
 		</method>
 		<method name="set_identity">
 			<return type="void" />
 			<description>
-				Reset all transformations for this node (sets its [Transform3D] to the identity matrix).
+				Sets this node's [member transform] to [constant Transform3D.IDENTITY], which resets all transformations in parent space ([member position], [member rotation], and [member scale]).
 			</description>
 		</method>
 		<method name="set_ignore_transform_notification">
 			<return type="void" />
 			<param index="0" name="enabled" type="bool" />
 			<description>
-				Sets whether the node ignores notification that its transformation (global or local) changed.
+				If [code]true[/code], the node will not receive [constant NOTIFICATION_TRANSFORM_CHANGED] or [constant NOTIFICATION_LOCAL_TRANSFORM_CHANGED].
+				It may useful to call this method when handling these notifications to prevent infinite recursion.
 			</description>
 		</method>
 		<method name="set_notify_local_transform">
 			<return type="void" />
 			<param index="0" name="enable" type="bool" />
 			<description>
-				Sets whether the node notifies about its local transformation changes. [Node3D] will not propagate this by default.
+				If [code]true[/code], the node will receive [constant NOTIFICATION_LOCAL_TRANSFORM_CHANGED] whenever [member transform] changes.
+				[b]Note:[/b] Some 3D nodes such as [CSGShape3D] or [CollisionShape3D] automatically enable this to function correctly.
 			</description>
 		</method>
-		<method name="set_notify_transform">
+		<method name="set_notify_transform" keywords="set_notify_global_transform">
 			<return type="void" />
 			<param index="0" name="enable" type="bool" />
 			<description>
-				Sets whether the node notifies about its global and local transformation changes. [Node3D] will not propagate this by default, unless it is in the editor context and it has a valid gizmo.
+				If [code]true[/code], the node will receive [constant NOTIFICATION_TRANSFORM_CHANGED] whenever [member global_transform] changes.
+				[b]Note:[/b] Most 3D nodes such as [VisualInstance3D] or [CollisionObject3D] automatically enable this to function correctly.
+				[b]Note:[/b] In the editor, nodes will propagate this notification to their children if a gizmo is attached (see [method add_gizmo]).
 			</description>
 		</method>
 		<method name="set_subgizmo_selection">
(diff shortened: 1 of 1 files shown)


doc/classes/Node3D.xml
```diff
@@ -223,134 +236,145 @@
...
-				Set subgizmo selection for this node in the editor.
+				Selects the [param gizmo]'s subgizmo with the given [param id] and sets its transform. Only works in the editor.
 				[b]Note:[/b] The gizmo object would typically be an instance of [EditorNode3DGizmo], but the argument type is kept generic to avoid creating a dependency on editor classes in [Node3D].
 			</description>
 		</method>
 		<method name="show">
 			<return type="void" />
 			<description>
-				Enables rendering of this node. Changes [member visible] to [code]true[/code].
+				Allows this node to be rendered. Equivalent to setting [member visible] to [code]true[/code]. This is the opposite of [method hide]
```
AThousandShips, line 246:
```suggestion
				Allows this node to be rendered. Equivalent to setting [member visible] to [code]true[/code]. This is the opposite of [method hide].
```

doc/classes/Node3D.xml
```diff
@@ -136,85 +143,91 @@
...
+				Orthonormalizes this node's [member basis]. This resets the [member scale] or any other distortion to the default, but preserves the [member position] and [member rotation]. See also [method Transform3D.orthonormalized].
 			</description>
 		</method>
 		<method name="rotate">
 			<return type="void" />
 			<param index="0" name="axis" type="Vector3" />
 			<param index="1" name="angle" type="float" />
 			<description>
-				Rotates the local transformation around axis, a unit [Vector3], by specified angle in radians.
+				Rotates this node's local [member transform] around the [param axis] by the given [param angle], in radians. The operation happens in local space.
```
RedMser, line 154: From how I understand it, `rotate` and `rotate_[xyz]` specify the axis in global space (parent space? not sure about proper terminology), but both modify the local space `transform`. Whereas `rotate_object_local` also transforms the rotation axis into local space. Right now, both of these methods have the same description.

doc/classes/Node3D.xml
```diff
@@ -223,134 +236,145 @@
...
 			</description>
 		</method>
 		<method name="translate">
 			<return type="void" />
 			<param index="0" name="offset" type="Vector3" />
 			<description>
-				Changes the node's position by the given offset [Vector3].
-				Note that the translation [param offset] is affected by the node's scale, so if scaled by e.g. [code](10, 1, 1)[/code], a translation by an offset of [code](2, 0, 0)[/code] would actually add 20 ([code]2 * 10[/code]) to the X coordinate.
+				<!-- Changes the node's position by the given offset [Vector3]. -->
+				Adds the given translation [param offset] to the node's [member position] in local space.
```
RedMser, line 268: From how I understand it, `translate` specifies the offset in global space (parent space? not sure about proper terminology), but modifies the local space `transform`. Whereas `translate_object_local` also transforms the offset into local space. Right now, both of these methods have the same description.

RedMser (commented):
One bit of the Node3D methods (and corresponding docs) always confused me as a user. I've added comments there, but am unsure about proper terminology.

doc/classes/Node3D.xml
Mickeon: I have kept these in mind for further testing. These methods have also always confused me so it would be nice to finally write something easier to understand.

lostminds:
Node2D has a lot of the same properties as the 3D version, and [often with very short descriptions](https://docs.godotengine.org/en/latest/classes/class_node2d.html) in the docs. So if you add improved descriptions of the concepts in the Node3D versions perhaps some of them can be applied to Node2D as well? Like for example what the global_position and global_transform properties do and how they work.

Mickeon:
That's the idea! One thing at the time. I decided to tackle Node3D first because it's fresh on the mind after both https://github.com/godotengine/godot/pull/87175 and https://github.com/godotengine/godot/pull/87334.

doc/classes/Node3D.xml
```diff
@@ -18,97 +18,102 @@
...
-				Clears subgizmo selection for this node in the editor. Useful when subgizmo IDs become invalid after a property change.
+				Deselects all subgizmos for this node. Useful to call when the selected subgizmo may no longer exist after a property change. Only works in the editor.
 			</description>
 		</method>
 		<method name="force_update_transform">
 			<return type="void" />
 			<description>
-				Forces the transform to update. Transform changes in physics are not instant for performance reasons. Transforms are accumulated and then set. Use this if you need an up-to-date transform when doing physics operations.
+				Forces the node's [member transform] to update. Fails if the node is not inside the tree.
+				[b]Note:[/b] For performance reasons, transform changes are usually accumulated and applied [i]once[/i] at the end of the frame. The update propagates through children 3D nodes, as well. Therefore, use this method only when you need an up-to-date transform (such as during physics operations).
```
AThousandShips, line 41:
```suggestion
				[b]Note:[/b] For performance reasons, transform changes are usually accumulated and applied [i]once[/i] at the end of the frame. The update propagates through 3D node children as well. Therefore, only use this method when you need an up-to-date transform (such as during physics operations).
```

doc/classes/Node3D.xml
```diff
@@ -137,85 +142,91 @@
 		<method name="orthonormalize">
 			<return type="void" />
 			<description>
-				Resets this node's transformations (like scale, skew and taper) preserving its rotation and translation by performing Gram-Schmidt orthonormalization on this node's [Transform3D].
+				Orthonormalizes this node's [member basis]. This resets the [member scale] or any other distortion to the default, but preserves the [member position] and [member rotation]. See also [method Transform3D.orthonormalized].
```
AThousandShips, line 145: Don't think "distortion" is an appropriate term here, unsure what to use though

doc/classes/Node3D.xml
Mickeon: If it needs to say "3D node children" it may as well say "Node3D children"

AThousandShips: Can also be "child 3D nodes" but I know you prefer the former :) it just shouldn't be "children ... Nodes"

Mickeon: Aaah, right. The PR of last time!

Mickeon:
I have updated the PR with just... more. I am still not fully happy on what is here, as there's a constant fear of losing important information through all of this. So this is still gonna be WIP for a bit more time until I come back with a fresher mind.

doc/classes/Node3D.xml
```diff
@@ -18,97 +19,101 @@
...
+				Returns all the [EditorNode3DGizmo] objects attached to this node. Only works in the editor.
 			</description>
 		</method>
 		<method name="get_parent_node_3d" qualifiers="const">
 			<return type="Node3D" />
 			<description>
-				Returns the parent [Node3D], or [code]null[/code] if no parent exists, the parent is not of type [Node3D], or [member top_level] is [code]true[/code].
-				[b]Note:[/b] Calling this method is not equivalent to [code]get_parent() as Node3D[/code], which does not take [member top_level] into account.
+				<!-- Returns the parent [Node3D], or [code]null[/code] if no parent exists, the parent is not a [Node3D], or [member top_level] is [code]true[/code]. -->
+				Returns the parent [Node3D] that directly affects this node's [member global_transform]. Returns [code]null[/code] if no parent exists, the parent is not a [Node3D], or [member top_level] is [code]true[/code].
```
Mickeon, line 55: HEAVY indecision between these two. Above is entirely correct, but does not explain why this method exists.

doc/classes/Node3D.xml
```diff
@@ -1,13 +1,14 @@
...
-		[b]Note:[/b] Unless otherwise specified, all methods that have angle parameters must have angles specified as [i]radians[/i]. To convert degrees to radians, use [method @GlobalScope.deg_to_rad].
-		[b]Note:[/b] Be aware that "Spatial" nodes are now called "Node3D" starting with Godot 4. Any Godot 3.x references to "Spatial" nodes refer to "Node3D" in Godot 4.
+		<!-- Most basic 3D game object, with a [Transform3D] and visibility settings. All other 3D game objects inherit from [Node3D]. Use [Node3D] as a parent node to move, scale, rotate and show/hide children in a 3D project. -->
+		The [Node3D] node is the base representation of a node in 3D space. All other 3D nodes inherit from this class.
+		Affine operations (rotate, scale, translate) happen in parent's local coordinate system, unless the [Node3D]'s [member top_level] is [code]true[/code]. Affine operations in this coordinate system correspond to direct affine operations on the [Node3D]'s transform. The word local below refers to this coordinate system. The coordinate system that is attached to the [Node3D] object itself is referred to as object-local coordinate system.
```
Mickeon, line 9: Affine, direct Affine operations... 😖

doc/classes/Node3D.xml
```diff
@@ -224,143 +233,155 @@
...
-			Defines the visibility range parent for this node and its subtree. The visibility parent must be a GeometryInstance3D. Any visual instance will only be visible if the visibility parent (and all of its visibility ancestors) is hidden by being closer to the camera than its own [member GeometryInstance3D.visibility_range_begin]. Nodes hidden via the [member Node3D.visible] property are essentially removed from the visibility dependency tree, so dependent instances will not take the hidden node or its ancestors into account.
+			Path to the visibility range parent for this node and its descendants. The visibility parent must be a [GeometryInstance3D].
+			Any visual instance will only be visible if the visibility parent (and all of its visibility ancestors) is hidden by being closer to the camera than its own [member GeometryInstance3D.visibility_range_begin]. Nodes hidden via the [member Node3D.visible] property are essentially removed from the visibility dependency tree, so dependent instances will not take the hidden node or its descendants into account.
```
Mickeon, line 343: I still find a bit difficult to understand what this is for, despite my best efforts.

Mickeon:
Marking this as ready to be lynched by anyone more savvy than me, no beating around the bush.

doc/classes/Node3D.xml
```diff
@@ -131,91 +134,95 @@
...
+				Orthonormalizes this node's [member basis]. This resets this node's [member scale] to the default, but preserves the [member position] and [member rotation]. See also [method Transform3D.orthonormalized].
 			</description>
 		</method>
 		<method name="rotate">
 			<return type="void" />
 			<param index="0" name="axis" type="Vector3" />
 			<param index="1" name="angle" type="float" />
 			<description>
-				Rotates the local transformation around axis, a unit [Vector3], by specified angle in radians.
+				Rotates this node's local [member transform] around the [param axis] by the given [param angle], in radians. The operation happens in local space (relative to the parent).
```
kleonc, line 151: This does rotate so-called local transform (which represents the transformation from the local to the parent coordinate system), around an axis specified in the **parent** coordinate system. So the operation does not happen in the local/object space, but in the parent space (as axis is defined in the parent space)!

> [!caution]
> "local space" != "parent space"/"relative to parent"

Please don't mix them.
If some existing docs refers to the parent space as "local", I would consider this as a bug. Are there such docs? :thinking:

I hope you haven't got merged any PRs introducing similar mixing of local/parent spaces? :upside_down_face:

---

`rotate_{x/y/z}(angle)` also rotate around parent space axis (+X/+Y/+Z), they're equivalent to `rotate(Vector3(1, 0, 0), angle)`/`rotate(Vector3(0, 1, 0), angle)`/`rotate(Vector3(0, 0, 1), angle)`.

---

BTW I often use local/global space etc. but I think it might be better if in the docs we'd refer to them explicitly as "local/global coordinate system", it should be clearer and more precise at the same time.
(What is space? Baby don't hurt me...)

doc/classes/Node3D.xml
```diff
@@ -131,91 +134,95 @@
...
 			</description>
 		</method>
 		<method name="rotate_object_local">
 			<return type="void" />
 			<param index="0" name="axis" type="Vector3" />
 			<param index="1" name="angle" type="float" />
 			<description>
-				Rotates the local transformation around axis, a unit [Vector3], by specified angle in radians. The rotation axis is in object-local coordinate system.
+				Rotates this node's local [member transform] around the [param axis] by the given [param angle], in radians. The operation happens in local space (relative to the parent).
```
kleonc, line 159: This does rotate so-called local transform (which transforms from local to parent coordinate system), around an axis specified in the **local** coordinate system. So this does happen in the local/object space, not in the parent space.
(this method name is super clear about it as it explicitly states `object_local`)

doc/classes/Node3D.xml
```diff
@@ -224,143 +231,155 @@
...
 			<description>
-				Updates all the [Node3D] gizmos attached to this node.
+				Updates all the [EditorNode3DGizmo] objects attached to this node. Only works in the editor.
 			</description>
 		</method>
 	</methods>
 	<members>
 		<member name="basis" type="Basis" setter="set_basis" getter="get_basis">
-			Basis of the [member transform] property. Represents the rotation, scale, and shear of this node.
+			Basis of the [member transform] property. Represents the rotation, scale, and shear of this node in local space (relative to the parent node).
```
kleonc, line 281: Node3D in its own local/object coordinate system is always not rotated/scaled, its Basis would be `Basis.IDENTITY`. This `basis` is in parent space / transforms from local to parent space.

doc/classes/Node3D.xml
```diff
@@ -224,143 +231,155 @@
...
 		</member>
 		<member name="global_transform" type="Transform3D" setter="set_global_transform" getter="get_global_transform">
-			World3D space (global) [Transform3D] of this node.
+			The transformations of this node in global space (relative to the world). Contains and represents this node's [member global_position], [member global_rotation], and global scale.
+			[b]Note:[/b] Getting this property while the node is not inside the tree will fail and return [constant Transform3D.IDENTITY].
 		</member>
-		<member name="position" type="Vector3" setter="set_position" getter="get_position" default="Vector3(0, 0, 0)">
-			Local position or translation of this node relative to the parent. This is equivalent to [code]transform.origin[/code].
+		<member name="position" type="Vector3" setter="set_position" getter="get_position" default="Vector3(0, 0, 0)" keywords="translation">
+			Position (translation) of this node in local space (relative to the parent node). This is equivalent to the [member transform]'s [member Transform3D.origin].
```
kleonc, line 304: Node3D's position in its own local/object coordinate space is always zero (`Vector3.ZERO`). This `position` is in the parent coordinate system.

And so on for other _"in local space (relative to the parent node)"_...

kleonc (commented):
Only skimmed through, added some brief notes. Please fix these _"local space (relative to the parent)"_.

AFAIK "local" in Godot's properties/method names always refer to object-local, not to parent-local. Docs should be consistent with this.
If there are some exceptions that I'm not aware of, please point to them. :slightly_smiling_face:

doc/classes/Node3D.xml
Mickeon: I am taking this to heart. 

> I hope you haven't got merged any PRs introducing similar mixing of local/parent spaces? 🙃

I am not sure, as aaronfranke scrutinised https://github.com/godotengine/godot/pull/87175 which was accounted for when making  https://github.com/godotengine/godot/pull/87334 , but a second look from you cannot hurt anything other than my self-esteem.

doc/classes/Node3D.xml
```diff
@@ -224,143 +231,155 @@
...
-			[b]Note:[/b] This property is edited in the inspector in degrees. If you want to use degrees in a script, use [member rotation_degrees].
+			Rotation of this node as [url=https://en.wikipedia.org/wiki/Euler_angles]Euler angles[/url], in radians and in local space (relative to the parent node). This value is obtained from [member basis]'s rotation.
+			- The [member Vector3.x] is the angle around the local X axis (pitch);
+			- The [member Vector3.y] is the angle around the local Y axis (yaw);
+			- The [member Vector3.z] is the angle around the local Z axis (roll).
+			The order of each consecutive rotation can be changed with [member rotation_order] (see [enum EulerOrder] constants). By default, the YXZ convention is used ([constant EULER_ORDER_YXZ]): Z (roll) is calculated first, then X (pitch), and lastly Y (yaw).
```
aaronfranke, line 315: Which is calculated first depends on whether you are composing or decomposing and whether you are considering intrinsic or extrinsic angles. I'm not sure if this note after the colon is helpful, it may be better to just have users click on `EULER_ORDER_YXZ` to read more there.

Mickeon:
Given your recent review I'll revisit. The "local" distinction is bit overwhelming to think about, but I must pursue for the sake of a nicer class reference.

doc/classes/Node3D.xml
```diff
@@ -18,97 +18,100 @@
...
+				Returns all the [EditorNode3DGizmo] objects attached to this node. Only works in the editor.
 			</description>
 		</method>
 		<method name="get_parent_node_3d" qualifiers="const">
 			<return type="Node3D" />
 			<description>
-				Returns the parent [Node3D], or [code]null[/code] if no parent exists, the parent is not of type [Node3D], or [member top_level] is [code]true[/code].
-				[b]Note:[/b] Calling this method is not equivalent to [code]get_parent() as Node3D[/code], which does not take [member top_level] into account.
+				Returns the parent [Node3D] that directly affects this node's [member global_transform]. Returns [code]null[/code] if no parent exists, the parent is not a [Node3D], or [member top_level] is [code]true[/code].
+				[b]Note:[/b] This method is not always equivalent to [method Node.get_parent], which does not take [member top_level] into account.
```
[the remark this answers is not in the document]
Mickeon: Yes. I believe directly linking to `get_parent` is more important than using `[code]` tags for an example that only works in GDScript.

Mickeon:
I have rebased. My general confusion stems from the original documentation's definition of "local" which has been irritatingly misleading.

> BTW I often use local/global space etc. but I think it might be better if in the docs we'd refer to them explicitly as "local/global coordinate system", it should be clearer and more precise at the same time.

> In general, Godot's properties are parent-relative by default unless further qualified, where the word global qualifies it as fully global relative to the world space, and the word local qualifies it as relative to within the node's own transform. For the docs where we need to clarify this, let's use "parent-relative", not "local".

Thank you @kleonc and @aaronfranke for writing detailed explanations on the distinction between these 3 "spaces"/"coordinate systems", but we have to discuss this. We need a consistent way to refer to these in order to reduce ambiguity as much as possible.
It doesn't help that in some other places, different terms are used to refer to the same concept [^1][^2]. I myself understand now that both "_object-local_" and "_parent-local_" spaces are, in fact, "_local_", but a problem arises when both are summarised with that single word.

Right now, in this PR, I am using the term "parent space". This term is never mentioned anywhere in the class reference. Even though a lot of wording can still be improved, this has been done for sake of accuracy first.

[^1]: https://docs.blender.org/manual/en/latest/animation/constraints/interface/common.html#space-types
[^2]: https://help.autodesk.com/view/MAYAUL/2024/ENU/?guid=GUID-A63AC5C8-8822-42AC-827E-164B5266DA03

doc/classes/Node3D.xml
```diff
@@ -1,13 +1,13 @@
...
-		[b]Note:[/b] Unless otherwise specified, all methods that have angle parameters must have angles specified as [i]radians[/i]. To convert degrees to radians, use [method @GlobalScope.deg_to_rad].
-		[b]Note:[/b] Be aware that "Spatial" nodes are now called "Node3D" starting with Godot 4. Any Godot 3.x references to "Spatial" nodes refer to "Node3D" in Godot 4.
+		The [Node3D] node is the base representation of a node in 3D space. All other 3D nodes inherit from this class.
+		Affine operations (translation, rotation, scale) are calculated in the coordinate system relative to the parent, unless the [Node3D]'s [member top_level] is [code]true[/code]. In this coordinate system, affine operations correspond to direct affine operations on the [Node3D]'s [member transform]. The term "[i]parent space[/i]" refers to this coordinate system. The coordinate system that is attached to the [Node3D] itself is referred to as object-local coordinate system, or "[i]local space[/i]".
```
tetrapod00, line 8: What's the reasoning for the double emphasis for new technical terms? IMO either italics or quotes is sufficient. Both seems overkill.

doc/classes/Node3D.xml
```diff
@@ -1,13 +1,13 @@
...
+		[b]Note:[/b] Unless otherwise specified, all methods that need angle parameters must receive angles in [i]radians[/i]. To convert degrees to radians, use [method @GlobalScope.deg_to_rad].
+		[b]Note:[/b] In Godot 4, "[i]Spatial[/i]" nodes are named "[i]Node3D[/i]". Some parts of the documentation may still mention Godot 3.x's "[i]Spatial[/i]" nodes.
```
tetrapod00, line 10:
```suggestion
		[b]Note:[/b] In Godot 3, [Node3D] was called "Spatial".
```

I don't like keeping notes like this around, nearly 2 years and 4 minor versions into Godot 4's lifespan. At the very least, we should remove the part about parts of the documentation being incorrect, and we should treat Godot 4 as the norm rather than Godot 3.

It looks like all(?) the remaining references to "spatial" in the class reference were actually recently fixed in one of your PRs. So, good job on that and we *definitely* should not say that our docs are wrong, in our docs.

Mickeon: I'd remove this line too but, the n.1 fear of these kinds of PRs is the loss of any possible information that may have solved some old, perhaps overly-specific issue opened in the past.

Mickeon: I think adding `spatial` as a keyword in Node3D would actually make this note even more redundant.

tetrapod00: I think keeping the note around in general is justified (a rename of a core class in a major version is a big change), but we should not explicitly mention that our docs still use the wrong terms. The issue seems generally fixed in the class reference. I haven't checked the manual yet, but in my experience I haven't seen usage of spatial, and if there are still some usages I can make a PR to correct them.

doc/classes/Node3D.xml
Mickeon: Not much reasoning other than personal preference. In this particular case it was to make these words stand out even more, because they are referenced in the rest of the class reference without this explanation, but it's not... necessary.

doc/classes/Node3D.xml
Mickeon: Just to give closure, `rotate` is indeed in parent space (same as the `transform` property). If the `position` was anything but `Vector3(0, 0, 0)`, the node would rotate _and_ move around the parent.

doc/classes/Node3D.xml
tetrapod00: This is your call since the class reference is your domain, but I personally think it's too much emphasis. As one more point of data, I don't believe it's currently used. Searching for `"[i]` and `[i]"` finds 0 and 1 current instances (and the second one is using italics to reference an error message, not for new term introduction as you are doing here).

Mickeon: Nah, you're right, it's not a good precedent.

doc/classes/Node3D.xml
```diff
@@ -232,143 +239,159 @@
...
 		</member>
 		<member name="scale" type="Vector3" setter="set_scale" getter="get_scale" default="Vector3(1, 1, 1)">
-			Scale part of the local transformation.
-			[b]Note:[/b] Mixed negative scales in 3D are not decomposable from the transformation matrix. Due to the way scale is represented with transformation matrices in Godot, the scale values will either be all positive or all negative.
-			[b]Note:[/b] Not all nodes are visually scaled by the [member scale] property. For example, [Light3D]s are not visually affected by [member scale].
+			Scale of this node in parent space (relative to the parent node). This value is obtained from [member basis]'s scale.
+			[b]Note:[/b] The behavior of some 3D node types is not affected by this property. These include [Light3D], [Camera3D], [AudioStreamPlayer3D], etc.
+			[b]Warning:[/b] The scale's components must either be all positive or all negative, and [b]not[/b] exactly [code]0.0[/code]. Otherwise, it won't be possible to obtain the scale from the [member basis]. This may cause the intended scale to be lost when reloaded from disk, and potentially other unstable behavior.
```
Mickeon, line 343: Alternatives to the same note. This is the same as the original, but switched around:
```suggestion
			[b]Warning:[/b] Due to the way scale is represented with transformation matrices in Godot, the scale's components must either be all positive, all negative, or all [code]0.0[/code]. Mixed negative scales in 3D cannot be decomposed from the [member transform].
```
This... I don't know
```suggestion
			[b]Warning:[/b] The scale's components must either be all positive or all negative, and [b]not[/b] exactly [code]0.0[/code]. Otherwise, this may cause unstable behavior, as it won't be possible to obtain the intended scale from the [member basis].
```

tetrapod00: Doesn't the first alternative invert the meaning...? I thought that scale *shouldn't* be all `0.0`, as per this note. You can check this trivially in the editor, you get `Condition "det == 0" is true.` errors.

doc/classes/Node3D.xml
```diff
@@ -232,143 +239,159 @@
...
+			Path to the visibility range parent for this node and its descendants. The visibility parent must be a [GeometryInstance3D].
+			Any visual instance will only be visible if the visibility parent (and all of its visibility ancestors) is hidden by being closer to the camera than its own [member GeometryInstance3D.visibility_range_begin]. Nodes hidden via the [member Node3D.visible] property are essentially removed from the visibility dependency tree, so dependent instances will not take the hidden node or its descendants into account.
 		</member>
 		<member name="visible" type="bool" setter="set_visible" getter="is_visible" default="true">
-			If [code]true[/code], this node is drawn. The node is only visible if all of its ancestors are visible as well (in other words, [method is_visible_in_tree] must return [code]true[/code]).
+			If [code]true[/code], this node can be visible. The node is only rendered when all of its ancestors can be visible, as well. That means [method is_visible_in_tree] must return [code]true[/code].
```
aaronfranke, line 356: "can be" makes sense for the property itself, but for the latter part it doesn't make sense for a condition to depend on a condition.

```suggestion
			If [code]true[/code], this node can be visible. The node is only rendered when all of its ancestors are visible as well. That means [method is_visible_in_tree] must return [code]true[/code].
```

doc/classes/Node3D.xml
```diff
@@ -232,143 +239,159 @@
...
 		</member>
 		<member name="top_level" type="bool" setter="set_as_top_level" getter="is_set_as_top_level" default="false">
-			If [code]true[/code], the node will not inherit its transformations from its parent. Node transformations are only in global space.
+			If [code]true[/code], the node does not inherit its transformations from its parent. As such, node transformations will only be in global space, which also means that [member global_transform] and [member transform] will be identical.
 		</member>
 		<member name="transform" type="Transform3D" setter="set_transform" getter="get_transform" default="Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0)">
-			Local space [Transform3D] of this node, with respect to the parent node.
+			The transformation of this node, in parent space (relative to the parent node). Contains and represents this node's [member position], [member rotation], and [member scale].
```
aaronfranke, line 349: This drops the `[Transform3D]`.

For the transform itself, I think describing it as local space is both accurate and helpful, because people new to Godot need to immediately know that this isn't global.

There are 2 sets of transforms (all parents, and this node's transform), which result in 3 spaces (global with neither, parent-relative with just all parents, and object-local with both). For other properties and methods, it makes sense to clarify parent-relative or object-local because we need to clarify "Is it local in the same way that the node's transform is local, or is it local within the node's transform, even more local than the local transform?" but for the transform itself, there is no confusion, local can only mean one thing.

```suggestion
			Local space [Transform3D] of this node, relative to the parent node. Contains and represents this node's [member position], [member rotation], and [member scale].
```

doc/classes/Node3D.xml
```diff
@@ -232,143 +239,159 @@
...
 		</member>
 		<member name="rotation_order" type="int" setter="set_rotation_order" getter="get_rotation_order" enum="EulerOrder" default="2">
-			Specify the axis rotation order of the [member rotation] property. The final orientation is constructed by rotating the Euler angles in the order specified by this property.
+			The axis rotation order of the [member rotation] property. The final orientation is calculated by rotating around the X, Y, and Z axis in this order.
 		</member>
 		<member name="scale" type="Vector3" setter="set_scale" getter="get_scale" default="Vector3(1, 1, 1)">
-			Scale part of the local transformation.
-			[b]Note:[/b] Mixed negative scales in 3D are not decomposable from the transformation matrix. Due to the way scale is represented with transformation matrices in Godot, the scale values will either be all positive or all negative.
-			[b]Note:[/b] Not all nodes are visually scaled by the [member scale] property. For example, [Light3D]s are not visually affected by [member scale].
+			Scale of this node in parent space (relative to the parent node). This value is obtained from [member basis]'s scale.
```
aaronfranke, line 341: I am begging you, please do not make changes to the documentation if you don't take the time to actually verify the information is correct. The scale is applied after the rotation, which means the scale is *not* relative to the parent node. For example, here is a box scaled by 3 on the local Z axis, rotated by 90 degrees around the Y axis, and therefore it gets bigger on the global X axis.

Screenshot 2024-11-24 at 11 25 05 AM

doc/classes/Node3D.xml
Mickeon: I am considering somewhat copying the description from CanvasItem, which was updated while this PR was in limbo:

> If [code]true[/code], this [CanvasItem] may be drawn. Whether this [CanvasItem] is actually drawn depends on the visibility of all of its [CanvasItem] ancestors. In other words: this [CanvasItem] will be drawn when [method is_visible_in_tree] returns [code]true[/code] and all [CanvasItem] ancestors share at least one [member visibility_layer] with this [CanvasItem].

doc/classes/Node3D.xml
Mickeon: So it's in object-local space!

doc/classes/Node3D.xml
Mickeon: > This drops the `[Transform3D]`.

It's intentional. There's often no need to specify the type when it is already defined by the property header.

doc/classes/Node3D.xml
```diff
@@ -232,143 +239,159 @@
...
-			Specify the axis rotation order of the [member rotation] property. The final orientation is constructed by rotating the Euler angles in the order specified by this property.
+			The axis rotation order of the [member rotation] property. The final orientation is calculated by rotating around the X, Y, and Z axis in this order.
 		</member>
 		<member name="scale" type="Vector3" setter="set_scale" getter="get_scale" default="Vector3(1, 1, 1)">
-			Scale part of the local transformation.
-			[b]Note:[/b] Mixed negative scales in 3D are not decomposable from the transformation matrix. Due to the way scale is represented with transformation matrices in Godot, the scale values will either be all positive or all negative.
-			[b]Note:[/b] Not all nodes are visually scaled by the [member scale] property. For example, [Light3D]s are not visually affected by [member scale].
+			Scale of this node in local space (relative to this node). This value is obtained from [member basis]'s scale.
+			[b]Note:[/b] The behavior of some 3D node types is not affected by this property. These include [Light3D], [Camera3D], [AudioStreamPlayer3D], etc.
```
aaronfranke, line 342: "etc" implies other elements in the list are obvious (like "and so forth").

```suggestion
			[b]Note:[/b] The behavior of some 3D node types is not affected by this property. These include [Light3D], [Camera3D], [AudioStreamPlayer3D], and more.
```

doc/classes/Node3D.xml
```diff
@@ -232,143 +239,159 @@
...
-				Note that the translation [param offset] is affected by the node's scale, so if scaled by e.g. [code](10, 1, 1)[/code], a translation by an offset of [code](2, 0, 0)[/code] would actually add 20 ([code]2 * 10[/code]) to the X coordinate.
+				Adds the given translation [param offset] to the node's [member position], in parent space (relative to the parent). As such, the final translation is multiplied by the node's [member scale], unlike [method translate_object_local].
 			</description>
 		</method>
 		<method name="translate_object_local">
 			<return type="void" />
 			<param index="0" name="offset" type="Vector3" />
 			<description>
-				Changes the node's position by the given offset [Vector3] in local space.
+				Adds the given translation [param offset] to the node's position, in local space (relative to this node). As such, the final translation is relative to the node's facing direction.
```
aaronfranke, line 277:
```suggestion
				Adds the given translation [param offset] to the node's position, in object-local space (relative to this node). As such, the final translation is relative to the node's facing direction.
```

doc/classes/Node3D.xml
```diff
@@ -139,91 +142,95 @@
...
-				Rotates the local transformation around axis, a unit [Vector3], by specified angle in radians. The rotation axis is in object-local coordinate system.
+				Rotates this node's local transformation around the [param axis] by the given [param angle], in radians. This operation is calculated in local space (relative to this node), and as such does not affect the [member position].
 			</description>
 		</method>
 		<method name="rotate_x">
 			<return type="void" />
 			<param index="0" name="angle" type="float" />
 			<description>
-				Rotates the local transformation around the X axis by angle in radians.
+				Rotates this node's local transformation around its X axis by the given [param angle], in radians. This changes the pitch, without affecting the [member position].
```
aaronfranke, line 174: I checked the code and tested, it actually doesn't use the node's local axis, it's parent-relative.

```suggestion
				Rotates this node's local transformation around the parent's X axis by the given [param angle], in radians. This changes the pitch, without affecting the [member position].
```

doc/classes/Node3D.xml
```diff
@@ -139,91 +142,95 @@
 			<param index="2" name="up" type="Vector3" default="Vector3(0, 1, 0)" />
 			<param index="3" name="use_model_front" type="bool" default="false" />
 			<description>
-				Moves the node to the specified [param position], and then rotates the node to point toward the [param target] as per [method look_at]. Operations take place in global space.
+				Moves the node to the specified [param position], and then rotates the node to point toward the [param target] as per [method look_at]. This operation is calculated in global space (relative to the world).
```
aaronfranke, line 145: You change this part from "Operations take place" to "This operation is calculated", but the one in "look_at" still says "Operations take place". We should pick one wording and be consistent.

doc/classes/Node3D.xml
```diff
@@ -57,66 +58,68 @@
...
+				Returns the parent [Node3D] that directly affects this node's [member global_transform]. Returns [code]null[/code] if no parent exists, the parent is not a [Node3D], or [member top_level] is [code]true[/code].
+				[b]Note:[/b] This method is not always equivalent to [method Node.get_parent], which does not take [member top_level] into account.
 			</description>
 		</method>
 		<method name="get_world_3d" qualifiers="const">
 			<return type="World3D" />
 			<description>
-				Returns the current [World3D] resource this [Node3D] node is registered to.
+				Returns the [World3D] this node is registered to.
+				Usually, this is the same as this node's viewport (see [method Node.get_viewport] and [method Viewport.find_world_3d]).
```
aaronfranke, line 69:
```suggestion
				Usually, this is the same as the world used by this node's viewport (see [method Node.get_viewport] and [method Viewport.find_world_3d]).
```

doc/classes/Node3D.xml
Mickeon: I gather this means every mention of it should say "object-local" instead

kleonc (commented):
I've commented some on the methods' descriptions changes (skipped properties).

@Mickeon please don't apply my suggestions before @aaronfranke would comment on some, as I think I'm suggesting some changes to what he already suggested before.

---

BTW I think Euler angles explanations will need some further overhaul, they're not fully clear as is (see [this forum post](https://forum.godotengine.org/t/is-the-name-euler-angles-used-in-godot-correct/52328), and [my response](https://forum.godotengine.org/t/is-the-name-euler-angles-used-in-godot-correct/52328/2) with the info which should be somehow contained in the docs (not fully sure where exactly)). But this possibly could be done in a separate PR later.

doc/classes/Node3D.xml
```diff
@@ -57,66 +58,68 @@
...
 			<description>
-				Returns whether node notifies about its local transformation changes. [Node3D] will not propagate this by default.
+				Returns [code]true[/code] if the node receives [constant NOTIFICATION_LOCAL_TRANSFORM_CHANGED] whenever [member transform] changes. This is enabled with [method set_notify_local_transform].
 			</description>
 		</method>
 		<method name="is_scale_disabled" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns whether this node uses a scale of [code](1, 1, 1)[/code] or its local transformation scale.
+				Returns [code]true[/code] if this node's [member scale] is ignored. See also [method set_disable_scale].
```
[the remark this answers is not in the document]
Mickeon: From an outside perspective, in most scenarios, it does just affect the node's scale. I don't think the suggestion makes that clear enough and makes it look more complicated than it is.

Your suggestion is very accurate and worth noting some way still...
Something roughly similar to... (have not tested yet)?
```suggestion
				Returns [code]true[/code] if this node's [member global_transform] is automatically orthonormalized. This results in the node appearing undistorted, as if its [member global_scale] was set to [constant Vector3.ONE]. See also [method set_disable_scale] and [method Transform3D.orthonormalized].
```

Mickeon:
Pushed an update to address all of the above. But there's just a bit more to do.
- I did not go with the shortened version of each suggestion, as I feel like the descriptions can afford to be explicit here.
- I am not sure it's necessary to state that, for example, `rotate_x` is equivalent to using an axis of `Vector3(1, 0, 0` in `rotate`. I am not dying on that hill, though.
- I like saying "basis" instead of "local transformation". That's actually genius. The local transformation's origin is always void, so this simplifies things a lot to me.
    - I think mentioning that the position is unaffected could still be warranted, though. Just to make it clearer.
    - Perhaps in `global_basis` and `basis`'s own description these are **the** object-local transformations?
- I have handled similar descriptions as suggested, but the idea of saying _"around the global [param axis] going through this node's origin"_ interests me. It sounds a whole lot easier to imagine the whole operation that way.
- I have written a note encouraging to [use `translate_object_local` instead of `translate`](https://github.com/godotengine/godot/pull/87440#discussion_r1860914892)

doc/classes/Node3D.xml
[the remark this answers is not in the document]
kleonc: "as the last step of calculating it" indeed is not needed, just "automatically orthonormalized" sounds the same but simpler. :+1:

> "as if its [member global_scale] were set to [constant Vector3.ONE]"

In theory an orthonormal basis could have scale factor being any combination of `-1`/`+1`. After quick testing, in Godot `basis.orthonormalized().get_scale()` can return either `(1, 1, 1)` or `(-1, -1, -1)` (approx, you can get e.g. `( 1.0000000000,  0.9999998212,  0.9999999404)`).

```gdscript
@tool
extends EditorScript

func _run() -> void:
	var counts := {}
	for i in 1_000_000:
		var b := Basis(
			Vector3(randf_range(-2, 2), randf_range(-2, 2), randf_range(-2, 2)),
			Vector3(randf_range(-2, 2), randf_range(-2, 2), randf_range(-2, 2)),
			Vector3(randf_range(-2, 2), randf_range(-2, 2), randf_range(-2, 2))
		)
		#var s := b.orthonormalized().get_scale()
		var s := b.orthonormalized().get_scale().round()
		counts[s] = counts.get(s, 0) + 1
	for s in counts:
		#print("%13.10v %d" % [s, counts[s]])
		print("%2.0v %d" % [s, counts[s]]))
```
```
( 1,  1,  1) 500005
(-1, -1, -1) 499995
```

But `(-1, -1, -1)` should be only if the basis involves some negative scaling, like e.g. for `Basis(Vector3(-1, 0, 0), Vector3(0, 1, 0), Vector3(0, 0, 1))`.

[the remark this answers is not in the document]
Mickeon: I am unsure on how to word this, in that case. I have a feeling I'll come back to this in another PR, for now I'm trying to stick to accuracy.
> If [code]true[/code], this node's [member global_transform] is automatically orthonormalized. This results in this node not appearing distorted, as if its global scale were set to the default. See also [method is_scale_disabled] and [method orthonormalize].

kleonc:
> * I think mentioning that the position is unaffected could still be warranted, though. Just to make it clearer.

Nothing against making things clearer. 

> * Perhaps in `global_basis` and `basis`'s own description these are **the** object-local transformations?

If by "object-local transformation" you mean non-translating / keeping the position of the given object/node unchanged then sure you could say so. But I wouldn't add anything like that, this will be confusing. There's already `Node3D::translate_object_local` method. If "object-local" would mean "not changing position" then how could you ever "translate-object-local"? Makes no sense. Let's not add potential additional confusion.

The "object-local" being used means the same as what we already call just "local" in the context of the given node/object. Like:
- parent == parent-local == (parent-object)-local
- local == (self/this)-local == (self/this-object)-local == object-local

In fact I was wondering whether we should maybe use e.g. parent-local/self-local instead of just parent/local. Local is always relative to something, so it's potentially easy to misinterpret. Object-local is meant to be more clear, as it specifies that the "local" refers to the given object, not its parent etc. But I think that in context of e.g. Node3D it's already clear that "local" refers to this specific node/object? :thinking:

Anyway, if we'd want to use e.g. "object-local" instead of just "local" then we should probably do so consequently across the whole docs. AFAICT currently we use just "local" to refer to the given object/node, so I'm saying let's keep using it.

> * I have handled similar descriptions as suggested, but the idea of saying _"around the global [param axis] going through this node's origin"_ interests me. It sounds a whole lot easier to imagine the whole operation that way.

Again, nothing against clarifications. The purpose of the docs is not to be as concise as possible (while being correct). So if the descriptions can be made a little longer but easier to grasp for "a regular user" then sure it's welcomed (we'll try to keep them correct).

> * I have written a note encouraging to [use `translate_object_local` instead of `translate`](https://github.com/godotengine/godot/pull/87440#discussion_r1860914892)

If we want to encourage using `translate_object_local` then maybe we should make `translate`'s description only redirect to `translate_object_local`? Something like: _"Same as [method translate_object_local], for compatibility reasons. Prefer using it instead."_ This way the user would have to check it out and hence more likely use it? :thinking: Just throwing an idea, I don't have a strong opinion on that, it's fine as is / up to you.

But might be a good to add additional note/warning to `translate` like _"(For compatibility reasons, )this operation is not calculated in the parent space. To translate in parent space use `node_3d.position += offset`."_.

Mickeon:
I have updated to add "_and preserves the position_" where appropriate, as well as update `translate` and `translate_object_local` again.


> Anyway, if we'd want to use e.g. "object-local" instead of just "local" then we should probably do so consequently across the whole docs. AFAICT currently we use just "local" to refer to the given object/node, so I'm saying let's keep using it.

Mhm. Hence my reluctance to say **object-local** space. If it needs to be clarified it needs to be done in one huge swoop across the docs.


> > * I have handled similar descriptions as suggested, but the idea of saying _"around the global [param axis] going through this node's origin"_ interests me. It sounds a whole lot easier to imagine the whole operation that way.
> 
> Again, nothing against clarifications. The purpose of the docs is not to be as concise as possible (while being correct). So if the descriptions can be made a little longer but easier to grasp for "a regular user" then sure it's welcomed (we'll try to keep them correct).

Problem here is that I'd have to give up on directly mentioning the `basis` otherwise the sentence doesn't make sense, and for consistency that may need to escalate for the surrounding descriptions as well.

> Same as [method translate_object_local], for compatibility reasons. Prefer using it instead.

This sounds **exactly** like a deprecation message, but the method is not deprecated until we decide that separately. So it's a bit odd to do this.

Mickeon:
I have updated to add "_and preserves the position_" where appropriate, as well as update `translate` and `translate_object_local` again.


> Anyway, if we'd want to use e.g. "object-local" instead of just "local" then we should probably do so consequently across the whole docs. AFAICT currently we use just "local" to refer to the given object/node, so I'm saying let's keep using it.

Mhm. Hence my reluctance to say **object-local** space. If it needs to be clarified it needs to be done in one huge swoop across the docs.


> > * I have handled similar descriptions as suggested, but the idea of saying _"around the global [param axis] going through this node's origin"_ interests me. It sounds a whole lot easier to imagine the whole operation that way.
> 
> Again, nothing against clarifications. The purpose of the docs is not to be as concise as possible (while being correct). So if the descriptions can be made a little longer but easier to grasp for "a regular user" then sure it's welcomed (we'll try to keep them correct).

Problem here is that I'd have to give up on directly mentioning the `basis` otherwise the sentence doesn't make sense, and for consistency that may need to escalate for the surrounding descriptions as well.

> Same as [method translate_object_local], for compatibility reasons. Prefer using it instead.

This sounds **exactly** like a deprecation message, but the method is not deprecated until we decide that separately. So it's a bit odd to do this.

aaronfranke:
@Mickeon Can you rebase this PR? There's a lot of good here, we should try to get this merged sometime soon.

Mickeon:
Rebased after 8af50726522773a9f5d280054c9517a694d7ff65 and d0c421976c0822351a1a0401548677f3e5caa399 . Wew, what a trip.

The descriptions affected were `is_visible_in_tree` and `look_at`. In particular, see https://github.com/godotengine/godot/pull/100991#issuecomment-2726533175 too.

Mickeon:
Rebased after 8af50726522773a9f5d280054c9517a694d7ff65 and d0c421976c0822351a1a0401548677f3e5caa399 . Wew, what a trip.

The descriptions affected were `is_visible_in_tree` and `look_at`. In particular, see https://github.com/godotengine/godot/pull/100991#issuecomment-2726533175 too.

doc/classes/Node3D.xml
aaronfranke: When I wrote the original text, one of the main points was to inform the reader that the scale isn't stored directly, but rather is calculated from the Basis.

A scale of 0 on one axis is expected to give `Condition "det == 0" is true.` errors when rendering, but it should work with getting the scale and having only one axis be 0. If that doesn't work, it seems like a bug. Is this a problem with Basis, or does it only happen with Node3D?

doc/classes/Node3D.xml
```diff
@@ -142,91 +145,96 @@
...
 			<return type="void" />
 			<param index="0" name="enable" type="bool" />
 			<description>
-				Sets whether the node notifies about its local transformation changes. [Node3D] will not propagate this by default.
+				If [code]true[/code], the node will receive [constant NOTIFICATION_LOCAL_TRANSFORM_CHANGED] whenever [member transform] changes.
+				[b]Note:[/b] Some 3D nodes such as [CSGShape3D] or [CollisionShape3D] automatically enable this to function correctly.
 			</description>
 		</method>
-		<method name="set_notify_transform">
+		<method name="set_notify_transform" keywords="set_notify_global_transform">
```
aaronfranke, line 231: This is out-of-scope for a docs PR but I think we should really rename this to `set_notify_global_transform` and then keep `set_notify_transform` as an alias.

doc/classes/Node3D.xml
```diff
@@ -142,91 +145,96 @@
...
-				Scales the local transformation by given 3D scale factors in object-local coordinate system.
+				Scales this node's [member basis] by the given [param scale] factor. This operation is calculated in local space (relative to this node) and preserves the [member position].
 			</description>
 		</method>
 		<method name="set_disable_scale">
 			<return type="void" />
 			<param index="0" name="disable" type="bool" />
 			<description>
-				Sets whether the node uses a scale of [code](1, 1, 1)[/code] or its local transformation scale. Changes to the local transformation scale are preserved.
+				If [code]true[/code], this node's [member global_transform] is automatically orthonormalized. This results in this node not appearing distorted, as if its global scale were set to the default. See also [method is_scale_disabled] and [method orthonormalize].
```
aaronfranke, line 205: "the default" isn't ideal here. A user could grab a scene with a node at a scale of (2, 3, 4), and then if instancing that and using editable children, the reset button in the inspector will set to a "default" of (2, 3, 4).

```suggestion
				If [code]true[/code], this node's [member global_transform] is automatically orthonormalized. This results in this node not appearing distorted, as if its global scale were set to [code](1, 1, 1)[/code]. See also [method is_scale_disabled] and [method orthonormalize].
```

doc/classes/Node3D.xml
```diff
@@ -127,12 +131,11 @@
...
+				Rotates the node so that the local forward axis (-Z, [constant Vector3.FORWARD]) points toward the [param target] position. This operation is calculated in global space (relative to the world).
 				The local up axis (+Y) points as close to the [param up] vector as possible while staying perpendicular to the local forward axis. The resulting transform is orthogonal, and the scale is preserved. Non-uniform scaling may not work correctly.
-				The [param target] position cannot be the same as the node's position, the [param up] vector cannot be zero.
-				The [param target] and the [param up] cannot be [constant Vector3.ZERO], and shouldn't be colinear to avoid unintended rotation around local Z axis.
-				Operations take place in global space, which means that the node must be in the scene tree.
+				The [param target] position cannot be the same as the node's position, the [param up] vector cannot be [constant Vector3.ZERO]. Furthermore, the direction from the node's position to the [param target] position cannot be parallel to the [param up] vector, to avoid unintended rotation around local Z axis.
```
aaronfranke, line 136:
```suggestion
				The [param target] position cannot be the same as the node's position, the [param up] vector cannot be [constant Vector3.ZERO]. Furthermore, the direction from the node's position to the [param target] position cannot be parallel to the [param up] vector, to avoid unintended rotation around the local Z axis.
```

doc/classes/Node3D.xml
```diff
@@ -18,32 +18,33 @@
...
 			<description>
-				Clears subgizmo selection for this node in the editor. Useful when subgizmo IDs become invalid after a property change.
+				Deselects all subgizmos for this node. Useful to call when the selected subgizmo may no longer exist after a property change. Only works in the editor.
 			</description>
 		</method>
 		<method name="force_update_transform">
 			<return type="void" />
 			<description>
-				Forces the transform to update. Transform changes in physics are not instant for performance reasons. Transforms are accumulated and then set. Use this if you need an up-to-date transform when doing physics operations.
+				Forces the node's [member transform] to update. Fails if the node is not inside the tree.
```
aaronfranke, line 40: I took a peek at the implementation, I think this would be more accurate.

```suggestion
				Forces the [SceneTree]'s list of pending transform notifications to include this node, which will be accumulated and applied [i]once[/i] at the end of the frame. Fails if the node is not inside the tree.
```

I did copy some text from the line below, I'm not exactly sure what's the best phrasing for what goes in a note and what goes here, but anything is fine to me.

doc/classes/Node3D.xml
```diff
@@ -1,13 +1,13 @@
...
-		[b]Note:[/b] Be aware that "Spatial" nodes are now called "Node3D" starting with Godot 4. Any Godot 3.x references to "Spatial" nodes refer to "Node3D" in Godot 4.
+		The [Node3D] node is the base representation of a node in 3D space. All other 3D nodes inherit from this class.
+		Affine operations (translation, rotation, scale) are calculated in the coordinate system relative to the parent, unless the [Node3D]'s [member top_level] is [code]true[/code]. In this coordinate system, affine operations correspond to direct affine operations on the [Node3D]'s [member transform]. The term [i]parent space[/i] refers to this coordinate system. The coordinate system that is attached to the [Node3D] itself is referred to as object-local coordinate system, or [i]local space[/i].
+		[b]Note:[/b] Unless otherwise specified, all methods that need angle parameters must receive angles in [i]radians[/i]. To convert degrees to radians, use [method @GlobalScope.deg_to_rad].
+		[b]Note:[/b] In Godot 3, [Node3D] was named [i]Spatial[/i].
```
aaronfranke, line 10: The wording here implies that getting renamed to "Spatial" was an event, I think this is clearer:

```suggestion
		[b]Note:[/b] In Godot 3 and older, [Node3D] was named [i]Spatial[/i].
```

doc/classes/Node3D.xml
Mickeon: I'd rather reference the `Vector3.ONE` constant for a bit more maintainability. Maybe also saying "it's global scale were set to *the default of* [...]" but I'm not sure.

doc/classes/Node3D.xml
Mickeon: Agree. This one is nasty but at least addressable any time unlike the `global_scale` method.

doc/classes/Node3D.xml
```diff
@@ -235,143 +243,161 @@
...
+				[b]Note:[/b] Prefer using [method translate_object_local], instead, as this method may be changed in a future release.
+				[b]Note:[/b] Despite the naming convention, this operation is [b]not[/b] calculated in parent space for compatibility reasons. To translate in parent space, add [param offset] to the [member position] ([code]node_3d.position += offset[/code]).
 			</description>
 		</method>
 		<method name="translate_object_local">
 			<return type="void" />
 			<param index="0" name="offset" type="Vector3" />
 			<description>
-				Changes the node's position by the given offset [Vector3] in local space.
+				Adds the given translation [param offset] to the node's position, in local space (relative to this node). As such, the final translation is relative to the node's facing direction, that being its local forward axis (-Z, [constant Vector3.FORWARD]).
```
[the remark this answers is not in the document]
Mickeon: Good point. It would still be nice to further "hammer in" that it has _something_ to do with the node's local direction, because it makes it much easier to imagine. The node's "right" is not always the same as the global "right".
Maybe a code example would help, but not in this PR.

doc/classes/Node3D.xml
Mickeon: Mentioning the scene tree's inner workings on this is a bit much, but the behavior detail (possibly akin to `queue_redraw()`) may be very important.
But I'm not sure if it's going to be misunderstood or not. `force_update_transform` implies the updating happens in the moment the method is called, but it doesn't seem true, then?

doc/classes/Node3D.xml
Mickeon: > A scale of 0 on one axis is expected to give `Condition "det == 0" is true.` errors when rendering, but it should work with getting the scale and having only one axis be 0.

It does indeed work. In Node3D, the scale is also stored separately in memory (`data.scale`) probably for this reason.
Basis is not affected by this because the scale always gets calculated on the spot. It's a Node3D quirk.

There's a lot of things wrong that happen when any axis is `0.0`, so when I rewrote this I thought it's best to have a catch-all warning. Among them, as you may imagine, the rotation is not stored correctly either.

doc/classes/Node3D.xml
Mickeon: Okay, hold on.

> Forces the [SceneTree]'s list of pending transform notifications to include this node, which will be accumulated and applied [i]once[/i] at the end of the frame.

https://github.com/godotengine/godot/blob/44c55ad59b01dbf360777b1a9a18a229aeb1f4f7/scene/3d/node_3d.cpp#L828-L836

What you described is the usual behaviour. The method seems to do exactly the opposite. That is, remove them from the list of "pending notifications" and send NOTIFICATION_TRANSFORM_CHANGED right away, which makes more sense for its namesake.

aaronfranke: Oh, you're right.

aaronfranke (approved):

doc/classes/Node3D.xml
```diff
@@ -235,143 +243,161 @@
...
+				Returns the [param global_point] converted from global space to this node's local space. This is the opposite of [method to_global].
 			</description>
 		</method>
 		<method name="translate">
 			<return type="void" />
 			<param index="0" name="offset" type="Vector3" />
 			<description>
-				Changes the node's position by the given offset [Vector3].
-				Note that the translation [param offset] is affected by the node's scale, so if scaled by e.g. [code](10, 1, 1)[/code], a translation by an offset of [code](2, 0, 0)[/code] would actually add 20 ([code]2 * 10[/code]) to the X coordinate.
+				Adds the given translation [param offset] to the node's position, in local space (relative to this node). As such, the final translation is relative to the node's facing direction, that being its local forward axis (-Z, [constant Vector3.FORWARD]).
```
Mickeon, line 274: Should this be trimmed as well?
```suggestion
				Adds the given translation [param offset] to the node's position, in local space (relative to this node).
```

Mickeon:
Pushed to address incorrect constants in the `global_*` properties

doc/classes/Node3D.xml
fire: can related to hlod lod systems. Let me know if you need more keywords

Mickeon: It would certainly be nice to get more information about it. But changing this description can also be done in a separate PR.

doc/classes/Node3D.xml
```diff
@@ -142,91 +145,96 @@
...
 			<description>
-				Moves the node to the specified [param position], and then rotates the node to point toward the [param target] as per [method look_at]. Operations take place in global space.
+				Moves the node to the specified [param position], then rotates the node to point toward the [param target] position, similar to [method look_at]. This operation is calculated in global space (relative to the world).
 			</description>
 		</method>
 		<method name="orthonormalize">
 			<return type="void" />
 			<description>
-				Resets this node's transformations (like scale, skew and taper) preserving its rotation and translation by performing Gram-Schmidt orthonormalization on this node's [Transform3D].
+				Orthonormalizes this node's [member basis]. This resets this node's [member scale] to the default, but preserves the [member position] and [member rotation]. See also [method Transform3D.orthonormalized].
```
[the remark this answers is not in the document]
Mickeon: One take I was thinking is as follows:
```suggestion
				Orthonormalizes this node's [member basis]. This method normalizes this node's [member scale] (setting its components to be all [code]1.0[/code] or [code]-1.0[/code]), but preserves the [member position] and [member rotation]. See also [method Transform3D.orthonormalized].
```
But I may be going with what follows, for simplicity sake(??):
```suggestion
				Orthonormalizes this node's [member basis]. This method sets this node's [member scale] to [constant Vector3.ONE] (or its negative equivalent), but preserves the [member position] and [member rotation]. See also [method Transform3D.orthonormalized].
```

doc/classes/Node3D.xml
```diff
@@ -142,91 +145,96 @@
...
-				Scales the local transformation by given 3D scale factors in object-local coordinate system.
+				Scales this node's [member basis] by the given [param scale] factor. This operation is calculated in local space (relative to this node) and preserves the [member position].
 			</description>
 		</method>
 		<method name="set_disable_scale">
 			<return type="void" />
 			<param index="0" name="disable" type="bool" />
 			<description>
-				Sets whether the node uses a scale of [code](1, 1, 1)[/code] or its local transformation scale. Changes to the local transformation scale are preserved.
+				If [code]true[/code], this node's [member global_transform] is automatically orthonormalized. This results in this node not appearing distorted, as if its global scale were set to [constant Vector3.ONE]. See also [method is_scale_disabled] and [method orthonormalize].
```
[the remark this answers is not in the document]
Mickeon: Same comment as https://github.com/godotengine/godot/pull/87440#discussion_r1998308448

doc/classes/Node3D.xml
```diff
@@ -235,143 +242,161 @@
...
 	</methods>
 	<members>
 		<member name="basis" type="Basis" setter="set_basis" getter="get_basis">
-			Basis of the [member transform] property. Represents the rotation, scale, and shear of this node.
+			Basis of the [member transform] property. Represents the rotation, scale, and shear of this node in parent space (relative to the parent node).
 		</member>
 		<member name="global_basis" type="Basis" setter="set_global_basis" getter="get_global_basis">
-			Global basis of this node. This is equivalent to [code]global_transform.basis[/code].
+			Basis of the [member global_transform] property. Represents the rotation, scale, and shear of this node in global space (relative to the world).
+			[b]Note:[/b] If the node is not inside the tree, getting this property fails and return [constant Basis.IDENTITY].
```
kleonc, line 298: Sorry but this was a typo in my previous suggestion:
```suggestion
			[b]Note:[/b] If the node is not inside the tree, getting this property fails and returns [constant Basis.IDENTITY].
```

doc/classes/Node3D.xml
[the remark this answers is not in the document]
kleonc: `scale` is a Vector3 and normalizing a vector makes its length be `1.0`, not each of its coordinates be `+1.0`/`-1.0`. Meaning the first suggestion is at least confusing/misleading.

I like the second suggestion, seems correct and simple. :+1:

doc/classes/Node3D.xml
```diff
@@ -57,68 +58,71 @@
...
 			</description>
 		</method>
 		<method name="is_visible_in_tree" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns [code]true[/code] if the node is present in the [SceneTree], its [member visible] property is [code]true[/code] and all its ancestors are also visible. If any ancestor is hidden, this node will not be visible in the scene tree.
-				Visibility is checked only in parent nodes that inherit from [Node3D]. If the parent is of any other type (such as [Node], [AnimationPlayer], or [Node2D]), it is assumed to be visible.
-				[b]Note:[/b] This method does not take [member VisualInstance3D.layers] into account, so even if this method returns [code]true[/code], the node might end up not being rendered.
+				Returns [code]true[/code] if the node is inside the scene tree, its [member visible] property is [code]true[/code] and all its [Node3D] ancestors are also visible. If any [Node3D] ancestor is hidden, this node is not visible and will not be rendered. See also [method get_parent_node_3d].
+				[b]Note:[/b] If an ancestor is of any other type (such as [Node] or [Node2D]), it is assumed to be visible.
```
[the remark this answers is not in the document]
kleonc: The "sequence" sounds kinda strange/weird, but seems acceptable to me. Can't really come up with better wording myself. What's important is it seems correct now. :+1:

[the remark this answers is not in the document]
Mickeon: I have a feeling I'll have to revisit this again with a whole thesis someday, anyhow.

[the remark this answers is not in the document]
kleonc: > That's funny 'cause the description in this PR is _also_ taken almost verbatim from Node2D's equivalent method.

And BTW description for `CanvasItem.is_visible_in_tree` is similarily incorrect indeed, as generally the behavior seems kinda the same (`top_level` ignored etc.). But I think I see some bugs/inconsistencies in propagation there for edge cases (breaking the parent-chain with non-CanvasItem). I'll investigate it further later and try to report / address them later.

kleonc (approved):

Repiteo:
Thanks!

PR #82212, opened 2023-09-23
Allow editing editor settings from project manager

KoBeWi:
Closes https://github.com/godotengine/godot-proposals/issues/1602

https://github.com/user-attachments/assets/c576f515-fb21-4296-8466-ef3267bd05d9

I had to add a couple random ifs to decouple EditorSettingsDialog and related controls from EditorNode. Also the dialog requires a few singletons. They are initialized when opening it for the first time, so there is a short freeze.

PR #75916, opened 2023-04-11
Add automatic checking for engine updates

KoBeWi:
Closes https://github.com/godotengine/godot-proposals/issues/1132

The PR adds update checks in a form of a label on the bottom of Project Manager and editor. See https://github.com/godotengine/godot/pull/75916#issuecomment-1964669768 for more images.

The update checker will check for newer version and tell you when one is available. Clicking the button will open the download page for this specific version. The checker can check updates in 3 modes: "Newest Unstable", which checks for newest official version, "Newest Stable", which checks for newest stable version and "Newest Patch", which checks for the newest patch of the current minor version. "Newest Stable" is the default for stable versions, "Newest Unstable" is default for preview versions.

Note that updates are disabled by default due to network mode being offline by default. There is also a setting to disable updates separately from network mode.

For version checking I'm parsing the [version YAML](https://raw.githubusercontent.com/godotengine/godot-website/master/_data/versions.yml) from Godot website. The editor will make many assumptions about structure of the file (such as it's sorted from newest to oldest version), so hopefully it won't change 😅

You can test the PR by modifying `version.py` file. Compiling with `allow_automatic_update=false` scons option will disable the dialog completely.

PR #69751, opened 2022-12-08
[Editor Help] Add support for embedded and remote images in the class reference pages.

bruvzg:
Following up https://github.com/godotengine/godot/pull/69712#issuecomment-1340790970

- Add method to update RTL images without changing the text.
- Add support for `[img]http(s)://image_url[/img]` in the documentation, to load remote images (downloaded and cached).

<details>
 <summary>Usage Example</summary>

```xml
<brief_description>
	Axis-Aligned Bounding Box.
	[img]https://godotengine.org/storage/app/uploads/public/638/8d9/a60/6388d9a60ac95864295672.jpg[/img]
</brief_description>
<description>
	[img]editor://Logo[/img]
	[AABB] consists of a position, a size, and several utility functions. It is typically used for fast overlap tests.
	It uses floating-point coordinates. The 2D counterpart to [AABB] is [Rect2].
	Negative values for [member size] are not supported and will not work for most methods. Use [method abs] to get an AABB with a positive size.
	[b]Note:[/b] Unlike [Rect2], [AABB] does not have a variant that uses integer coordinates.
	Loading (bad url): [img]http://192.168.3.210:34/f4f/dopefish.jpg[/img]
</description>
```

https://user-images.githubusercontent.com/7645683/206395062-45c9369b-87fd-40bf-b5c4-5f8266387d60.mov

</details>

PR #87443, opened 2024-01-21
Improve layout and UX of the project manager

YuriSizov:
NVIDIA_Share_2024-01-21_14-18-20

## Introduction

For a few months now I have been thinking about how the project manager feels unpolished and is not a very good representation of Godot's editor UI. I considered that we already have established a nice basic layout with the editor itself, and it's would fit very well into the project manager. I also considered that this would be the least controversial way to update the project manager right now, since such a change would make it resemble the editor more, rather than try to push for something completely new.

With a redesign it's important to keep in mind not only what we want to provide right now, but also what we want to achieve in the future. There has long been talks about such features as **an update checker** or **a news feed** from the official website. At the same time, we wanted to improve the way how we handle privacy within the editor and its online features. These two goals are naturally at odds with each other, so it was also my task here to address them together and provide you with a full picture.

And finally, the project manager is currently lacking in the configuration department. You can only change a language, but not other settings which may be important when you first launch Godot. Such settings can be the editor UI theme and the display scale. These not being configurable from the getgo negatively affect user experience. I wanted to address this as well.

So I started with some rework weeks ago, and was able to finally finish it now.

### A refreshed look based on the "main view" design of the Godot editor itself.

The tab container has been replaced with the familiar view switches, which grab your attention much better. This should help highlight the asset library much better than before, where a tab for it would be smooshed and lost. Not much is different about the layout itself within each tab, since recent additions (as of 4.2) already make it more organized.

_We currently have some issues with keyboard navigation, which this PR doesn't fix. But it's something that we should definitely look into later._

NVIDIA_Share_2024-01-21_14-18-20

The top bar is split into three parts, which tie the overall look together. In the middle you have the tabs to switch between the project list, the asset library, and potentially more views. As a bait I've added a yet disabled "News" tab to give you all a better idea of how it might fit in. I am fully committed to implement it, if we agree on having it. Or it can be done by someone else too. Not in this PR, but as a follow-up, and, again, only if we agree. I can remove it from the final version of this PR if we don't agree or if we decide that leaving it disabled like that is not a good idea.

On the left I've added a Godot logo, just as a stylistic choice. I think it adds some character to the overall look, but I'm open to change it to something else. We can switch to a different logo (say, only text without the cute robot), or remove it too. But I'd like to insist on keeping it, or at least coming up with an alternative filler for this space. The logo acts as a replacement for the old "About" button.

On the right the old language selector is gone in favor of the new Quick Settings dialog button. More about it below.

The version string is now at the bottom. @KoBeWi's update checker (https://github.com/godotengine/godot/pull/75916) can be moved there as well, to alert users in a non-intrusive way. The new footer bar does take away from the vertical space, but in a recent PR I have actually fixed the size application in the project manager. So in the 4.3 branch it actually has more vertical space to work with by default than in 4.2 and before. That was actually a bug I discovered while working on this, and I think that combined these two changes result in plenty of space left for the project list.

Finally, I'm also trying something with the welcome dialog that you see if you have no projects. I think it's a good direction, but I'm not sure if I like the result insofar. But we can always improve on it later.

NVIDIA_Share_2024-01-21_14-51-22

### A new Quick Settings dialog to adjust more than just the language.

While there is a PR (https://github.com/godotengine/godot/pull/82212) to bring the entire editor settings dialog into the project manager, I find this solution to be overwhelming for the majority of use cases where you need to configure something before launching the editor.

Instead I propose that we provide a small and local dialog which allows customizing only some of the settings which may be immediately useful, especially when you start Godot for the first time. I've decided to stop at the language, scale, theme, and network mode (more about that one below). It's a bit of a custom UI, but it's easily to implement and maintain, and it allows users to quickly edit the most useful visual settings which directly affect their experience with the project manager.

NVIDIA_Share_2024-01-21_14-19-00

If we still want to provide the full editor settings dialog in the project manager, I think it can be added on top of this Quick Settings dialog. We can add a button to the Quick Settings dialog which would bring the full dialog up, and I think that would be sufficient compromise for those who need to tinker with some deeper laid settings.

There is, of course, an argument for having a "basic" toggle in the editor settings. But from my experience with the project settings dialog this toggle doesn't significantly help with the user experience. We reduce the number of settings (while also hiding some useful ones, arbitrarily), but the navigation still sucks. It's not a tailored menu, it's still a verbose list, which is not too clear to navigate. So my proposal is to add a tailored menu first, and then maybe add a shortcut for the full menu on top of it.

### The network mode setting to bring the editor offline.

With this PR we will start in the offline mode by default. This means that the asset library won't be making any web requests, not until you let the editor go online. The same will apply to future possible features like the update checker and the news section. The same should probably apply to the dialog which allows you to download export templates, and other editor tools with online activity, but pardon me, I haven't implemented it in this PR for simplicity's sake.

The idea is that we don't want to break privacy for users who are concerned about it. And we also want this to be as small of an obstacle as possible for the rest. So besides a new editor settings all UIs which need online activity should offer a way to switch to the online mode with one click.

NVIDIA_Share_2024-01-21_14-19-19

Added to the "Network" group of settings "Network Mode" currently has two options, "Offline" and "Online". I consciously made it an option list because I can imagine there being a need for a less binary configuration, with some partial mode to allow some connectivity (e.g. only from the Godot-hosted sources).

This feature should make it less controversial to add the update checker and the news tab, or other such features, in the future. It should also replace the adhoc flag from the update checker PR (https://github.com/godotengine/godot/pull/75916), as it provides a more universal solution for the editor to go completely offline.

-----

I've kept commits separate as I fully expect some bikeshedding to happen, some rejections to occur, and some ideas to be cut or to be asked to be split out. Each commit is self sufficient and should make it easier to review, but what I should do with them next is up to the consensus of you all.

To close things off, here's a short demo of how the first time experience might look like now.

https://github.com/godotengine/godot/assets/11782833/2b43711a-3bca-4882-bedd-9bfc89ed8438


ettiSurreal:
If quick settings are added, `filesystem/directories/autoscan_project_path` and `filesystem/directories/default_project_path` (maybe have a "remember path" checkbox in the create project dialog?) could be added to it, they seem more convenient to have in the project manager.
Though to be fair new projects shouldn't be placed directly in the documents folder by default to begin with (have a dedicated "Godot Projects" folder in there, but that's for another discussion.

GuyUnger:
I like it, the only thing that feels off is that on the right panel all options are about editing one of the projects except for "remove missing", feels like it doesnt belong there

AThousandShips:
I'd suggest separating the "remove missing" button by putting it at the bottom instead

YuriSizov:
> I like it, the only thing that feels off is that on the right panel all options are about editing one of the projects except for "remove missing", feels like it doesnt belong there

I think there was a discussion about it recently. It belongs to the top button row semantically, but it's also a feature that is not needed often. So I think someone mentioned putting it in some three-dot menu in the top row. Or we could do as @AThousandShips suggests.

theMotionblur:
I like it, though I'm not sure I about the First Popup Dialog box prompting the User to Make a choice about going online or cnaceling out of something. Feels a little unfriendly to begin with.
Would it be possible to maybe integrate the question about going online and trying an official demo projects as a text in the empty project layout? Or maybe have it as a selection option when the users wants to add an actual project.

YuriSizov:
> Would it be possible to maybe integrate the question about going online and trying an official demo projects as a text in the empty project layout?

Yeah, I thought about replacing the dialog with a similar UI placed on top of an empty project list. That's could be nice!

ettiSurreal:
Got an idea, so here's pretty hasty mockup (also I'm not a pro UI designer so don't judge). 
What if the menu on the right was just removed outright and replaced with buttons on the projects themselves?

image
(Note because I can't show it here, tags would be removed by hovering over one, then an X would appear on it, pressing which would remove it).
Could probably add an action popup by right clicking, which shows what was formerly on the right menu and more.

Always found it sort of weird how opening in file explorer and favoriting are on the projects themselves, At first i thought the folder icon wasn't a button and was just a fancy icon for the file path.

Doubt you'll go with this but I hope it at least gives some inspiration.

AThousandShips:
Icons on the entries themselves has been discussed in the past but this removes a lot of information, the icons aren't always obvious (and tooltips should not contain fundamental information, only additional information), I also find this much busier and less easy to navigate

ettiSurreal:
I agree with your points, though the main things I like from it is it feeling more "balanced" in a way, due to the tabs being centered now, and the simplified tag management (would need a better icon for adding tags when there are none though).
An additional idea i had was that, maybe when you hover over the project icon, the play (or edit) button appears over it, but ended up not adding it to the mockup.

AdrielCodeops:
Both Unreal and Unity display a tooltip when you right click on the project entry or on the "settings" icon in the project entry. I don't find it intuitive to have to select the project entry in one panel and have to switch to another unrelated panel to perform actions on the project. Another reason is that this design is very cumbersome to use on a smartphone where you expect to hold down on the project entry or click on some extra icon on the entry itself.

Maybe I'm wrong but I feel like 99% of the time you are opening a project or searching for one, not removing/adding/modifying anything about it in the project manager. I want to see more projects and less buttons that i don't touch most of the time.

In fact, I would move everything (other than new/add/search) to a side panel. That way you have WAY more vertical space to show you what matters: your projects. And if you make it hideable you'd have a ui that works well on both desktop and mobile.

AThousandShips:
> I want to see more projects and less buttons that i don't touch most of the time.

You won't see more projects without the sidebar than otherwise

AdrielCodeops:
Every mobile application allows you to maintain an entry to display options specific to that entry. None works the way godot works.

And yes, removing the entire top of the UI in a right side panel will effectively show more projects.

AThousandShips:
Less than the space for one project is taken up by the top, and just by making it wider we won't have more lines, that would just make things more crowded, most projects don't have names that are long enough to matter that there's a sidebar of not

YuriSizov:
I don't have a goal to make the editor more usable on a smartphone with this PR. For a decent user experience in such a constrained environment we need to design a bespoke interface.

So let's forgo this part of the discission for this time.

tokengamedev:
I have few questions/issues related to existing and updated design; It will be great if these changes can be taken up too.

1. Why the new button is so small, can it be named **New Project** to make it bigger. It is one of the primary actions in the Project Manager.
2. Can the project icon be smaller with padding on all sides. When you open the project manager, the icons hit like jack hammer to the eyes.
3. Remove Missing is not part of the group as it does not require project selection. it should be in the top header, but it can be ignored if don't want to change.
4. The new settings are actually customizations. Ideally it should be named **Customize**. 
5. If the warning symbol could be displayed just after project name instead of bottom right corner, it will be prominent.
6. Instead of **Godot Engine - Project Manager** in the title (which sounds robotic), it can be better like **Welcome to Godot Engine**(Which feels much better and warm)

In my view, the project manager looks crowded with functionality. I am not sure how many functionalities are really used that regularly. A gut check (or poll) of the usage has to be done and adjust the UI accordingly.

YuriSizov:
Based on feedback from @theMotionblur and @hmans adjusted the message that appears when the project list is empty. Instead of being put into a popup, it is now always visible in the body of the empty list. You now have two clear actions, one to create a new project, and one to go to the asset library. The note about online connections only appears if you are in the offline mode.

I also renamed the "New" button to "Create" to give it a bit more significance, visually.

image

PS. I've pushed it as a new commit. Once we are done with iterations, I will clean up commits into one or several but isolated commits, depending on how the discussion goes.

YuriSizov:
@tokengamedev 

> 1. Why the new button is so small, can it be named **New Project** to make it bigger. It is one of the primary actions in the Project Manager.

Yep, this was noted in other feedback and I've made some changes. I agree it can be "New Project" or "Create Project" too, if we prefer.

> 2. Can the project icon be smaller with padding on all sides. When you open the project manager, the icons hit like jack hammer to the eyes.

Are you talking about Godot logo, or about project icons? Project icons seem to fit the needed height right now, but I don't particularly have an opinion about what you suggest. This can be done separately though.

> 3. Remove Missing is not part of the group as it does not require project selection. it should be in the top header, but it can be ignored if don't want to change.

As discussed above there are some options. For now I've moved it lower in the sidebar as suggested above.

> 4. The new settings are actually customizations. Ideally it should be named **Customize**.

I don't see much point in this distinction. It's just a handpicked subset of settings, and it can be changed in future to include other options, not just what I've initially decided to include. I'd argue that the network mode is already falling out of the "Customize" grouping.

> 5. If the warning symbol could be displayed just after project name instead of bottom right corner, it will be prominent.

I don't have an opinion about that, and it's outside of the scope for this PR.

> 6. Instead of **Godot Engine - Project Manager** in the title (which sounds robotic), it can be better like **Welcome to Godot Engine**(Which feels much better and warm)

Sure, but this may make it harder to find in the task bar. This can also be done separately anyway, it's a minor change.

JoNax97:
Very nice work! I had to open the 4.2 Project Manager to check the differences, the new UI feels like how it was always meant to be!
I shared this on Twitter but I figured I'd use the proper channel as well. I have a small revision for the blank/no projects UI. I think button contrast and information hierarchization can be improved, and I'd personally add the option to open a project, which I think deserves the same attention as creating a new one. 

Don't mind the exact details like the button borders or the vertical spacing between text and buttons. It's just a quick and dirty mockup done in paint lol.

image

tokengamedev:
@YuriSizov , thanks for the detailed reply.

> Yep, this was noted in other feedback and I made some changes, see above. I agree it can be "New Project" or "Create Project" too, if we prefer.

I don't mind either way, my suggestion was to improve the prominence of the button as it is a primary action. It should stand out from the other array of actions as a UX designer would say.

> Are you talking about Godot logo, or about project icons? Project icons seem to fit the needed height right now, but I don't particularly have an opinion about the changes that you suggest. This can be done separately though.

yes, talking about the project Icons and not the Godot Logo(which in my opinion is awesome ❤️) . Just give a try to reduce the icons size to 2/3 or 1/2 but centered vertically. if it does not feel good, ignore this comment.

> I don't see much point in this distinction. It's just a handpicked subset of settings, and it can be changed in future to include other options, not just what I've initially decided to include. I'd argue that the network mode is already falling out of the "Customize" grouping.

Technically you are right. but, I looked into other editors like android studio, and VS code, they have customize. Moreover, if you add the word customize, the debate over what settings to be added in project manager is out. Talking about network mode: honestly, it is not necessary, as when you switch to asset library or news, it should automatically turn on or if you insist, follow the @JoNax97 approach above, although buttons should be link buttons

> I also entertained the idea of not having any text in the top-right corner at all. Just a button with an icon. This may make the UI less busy. 

👍 . I like the gear(⚙️) icon if settings, palette(🎨) icon  if customize


Anyway, it is awesome work. Keep it up.

YuriSizov:
> Talking about network mode: honestly, it is not necessary, as when you switch to asset library or news

The point is that it won't automatically turn on. Not unless you actively choose to go online as a user. In the current implementation if you click the "Open Asset Library" button from the empty list suggestion, it will auto-enable the online mode (which is why in the older iteration it spelled "Go Online and ..."). It should be acceptable because we have a note right next to it, explaining this.

But when simply switching to the asset library tab it won't go online by itself. If it would, there would be no point in the feature to limit online connectivity.

YuriSizov:
Adjusted formatting of the empty message a bit and added the import button too. I think it should be good enough for the purposes of this PR.

image

gomaproi:
small nitpick: it would be nice if the buttons had icons to indicate that they behave the same as the buttons at the top 🙃

image

theMotionblur:
> Adjusted formatting of the empty message a bit and added the import button too. I think it should be good enough for the purposes of this PR.
> 
> image

Yeah! I find that a lot nicer and more welcoming. :)
Also agree with @gomaproi - the icons could make it more coherent with the rest of the design.

JoNax97:
Part of my proposed design aimed to avoid the "pancake stack" effect that happens when multiple center-aligned elements are stacked. I think this hampers readability. If nothing else, I'd recommend turning the 2 paragraph into a single one 👍

YuriSizov:
> avoid the "pancake stack" effect

Could you explain why you think this negatively impacts readability?

JoNax97:
I find that you have to go back and forth to find where each block of text (or buttons) starts, since they all have the same separation and there's no easy queue to distinguish where each element starts.

In left-aligned text, your eyes can easily find the shorter length of the last line of each block (not to mention paragraph indentation).

In other words, I find that one can't make out the composition of the text at a glance. You have to slowly parse it instead.

YuriSizov:
Thanks for the explanation, and it makes sense. Although I would say that for a couple of lines it's not an issue that we should worry about.

octanejohn:
what if the sort goes inside the last edited label and alighs with the vertical collum

AThousandShips (approved):
Haven't tested it yet (will do tomorrow) but I think the design changes are a major improvement both for style and usability

I also especially like the default offline mode, it inspires confidence to me to have the editor not expect network by default and I think others will agree from a privacy and comfort level

rick551a:
Some suggestions:

"Asset Library" button:
Maybe call it the "Online Asset Library" , self-explanatory & dosen't require explaining that the button will take you online.


"You don't have any projects yet"
Maybe give an indication of what directory Godot is looking for Projects in.
e.g. "You do not have any projects found at C:/Godotdir/projects"

Icon:
A little icon overlay on the project icon to indicate what version of godot the project is using. e.g. 
|               |	
|icon.png |
|               |
[        [4.1]]

Sometimes I have 4.0 & 4.3beta projects & am unsure which is which.

AThousandShips:
> Maybe give an indication of what directory Godot is looking for Projects in.

That's just the default directory, it's not mandatory

LeulMulugeta:
small suggestion: Maybe the Note at the bottom should be shorter
55c840

doc/classes/EditorSettings.xml
```diff
@@ -722,6 +722,9 @@
 			Specify the multiplier to apply to the scale for the editor gizmo handles to improve usability on touchscreen devices.
 			[b]Note:[/b] Defaults to [code]1[/code] on non-touchscreen devices.
 		</member>
+		<member name="network/connection/network_mode" type="int" setter="" getter="">
+			Determines whether online features are enabled in the editor, such as the Asset Library. Setting this property to "Offline" is recommended to limit editor's internet activity, especially if privacy is a concern.
```
bruvzg, line 726: Is it disabling export template download as well? If it's not, it's probably should be named more specific, like `network/connection/online_asset_library`. Since export template dialog will download mirror list without asking, so it's not "offline".

AThousandShips:
The part about sending data is relevant and shouldn't be removed, it's not obvious

doc/classes/EditorSettings.xml
AThousandShips: Should probably be extended to cover that part too, or a separate setting

YuriSizov: I addressed this in the OP:

> The same should probably apply to the dialog which allows you to download export templates, and other editor tools with online activity, but pardon me, I haven't implemented it in this PR for simplicity's sake.

It is a generic setting, and everything that wants to go online in the editor must respect it. I just didn't go through everything to make it so for this PR.

KoBeWi:
When you go online in asset lib and then go offline, you end up with this:
image
The online info only appears after you try to interact with something. All buttons are clickable, but they do nothing. Should the interface disappear when you go offline?

KoBeWi:
It's nice to have this note: 
image
But maybe it should explicitly say that clicking Open Asset Library makes you go online. It automatically changes the setting after all.

YuriSizov:
> The online info only appears after you try to interact with something. All buttons are clickable, but they do nothing. Should the interface disappear when you go offline?

Right now the locked state of the UI is tied to whether or not we have a "config" downloaded. I was back and forth about disabling and clearing the UI if you go offline, but decided to keep it as is for now. It's an edge case. New requests won't go out, but you also already have some data loaded, so why bother removing it? You can only reach this state if you were online to begin with.

----

The force-push is just a rebase. I'll work on the feedback, but not yet.

AThousandShips (approved):
(Just confirming again after compiling and testing, looks and feels great)

Calinou (approved):

### Default

*Focus outline appears around the Godot logo because I last clicked on it.*

image

### Light

image


### Black (OLED)

image

YuriSizov:
@Riteo What do you mean by making the eyes full? Removing irises? The logo is exactly the same in both light and dark modes, it's just color adjusted like any other icon.

atirut-w:
> What do you mean by making the eyes full?

The eyes are circles and not dots.

image

Baiter160:
PhotoGridLite_1706358684855

Baiter160:
Put a "About Godot" with the white logo is much more appealing!

AThousandShips:
The two lines would get too cramped, also it'd be confusing as you can't translate the text

YuriSizov:
@Baiter160 We prefer not to use this version of the monochrome logo.

skyace65:
It seems kinda odd to have the Godot logo directly below where the Godot Icon and the words "Godot Engine - Project Manager" are. Does it look different on other operating systems? (I'm on Win 10)

atirut-w:
> Does it look different on other operating systems?

On Linux, specifically GNOME and KDE, the window titles are in the middle by default IIRC.

YuriSizov:
> Why not making it like this in the first place?

No reason, it was quickly put together. It probably needs some optimizations still, for example. I did ask Kenney on Twitter, when I had a chance, specifically out of concern about smaller details, and he mentioned that the logo looks great as is.

I'm not against trying it with a simpler form, though. I can even adjust it a bit to better work with the pixel grid on a smaller scale. Was just clarifying that light and dark use the same file, so any difference is up to our SVG rasterizer being non-deterministic and editor display scale.

deralmas (approved):
Amazing work!

timothyqiu:
I think the note about Asset Library is not necessary since the button says "**Go Online** and Open Asset Library". It could be shown on the Asset Library page when in offline mode.

The description text also feels unnecessary, it's just combing the three buttons' text into a sentence :P

The three buttons feels floating now (don't belong to the list area) compared to those shown in https://github.com/godotengine/godot/pull/87443#issuecomment-1904002935.


| Now | Then |
|---|---|
| now | then |

editor/plugins/asset_library_editor_plugin.cpp
```diff
@@ -1432,6 +1440,39 @@ void EditorAssetLibrary::_update_asset_items_columns() {
...
+	library_message_action = p_action;
+	library_message_button->connect("pressed", library_message_action);
+	library_message_button->show();
+
+	library_message_box->show();
+}
+
+void EditorAssetLibrary::_force_online_mode() {
+	EditorSettings::get_singleton()->set_setting("network/connection/network_mode", EditorSettings::NETWORK_ONLINE);
+	EditorSettings::get_singleton()->notify_changes();
```
KoBeWi, line 1472: Is the `notify_changes()` necessary? We use it only in the settings dialog (and there are more places where settings are changed).

ettiSurreal:
I'd suggest removing the "Note:" from the asset library comment, it makes it look like a dev/code comment.

editor/plugins/asset_library_editor_plugin.cpp
YuriSizov: I think without that the notification isn't sent, and I rely on it elsewhere.

KoBeWi: I remember simply modifying the settings would cause propagation and saving automatically, but maybe it has changed.

editor/project_manager.cpp
```diff
@@ -1123,88 +1211,163 @@ ProjectManager::ProjectManager() {
...
+	// Asset library view.
 	if (AssetLibraryEditorPlugin::is_available()) {
 		asset_library = memnew(EditorAssetLibrary(true));
-		asset_library->set_name(TTR("Asset Library Projects"));
-		tabs->add_child(asset_library);
+		asset_library->set_name("AssetLibraryTab");
+		_add_main_view(MAIN_VIEW_ASSETLIB, TTR("Asset Library"), Ref<Texture2D>(), asset_library);
 		asset_library->connect("install_asset", callable_mp(this, &ProjectManager::_install_project));
 	} else {
 		print_verbose("Asset Library not available (due to using Web editor, or SSL support disabled).");
```
KoBeWi, line 1334: Wouldn't it be better to make a tooltip like in news?

editor/plugins/asset_library_editor_plugin.cpp
YuriSizov: Supposedly doing `set`/`set_setting` should emit a signal which the editor settings dialog connects to and triggers the notification and the save on 1.5 second delay. I'll double check.

editor/project_manager.cpp
YuriSizov: Yeah, I'm in favor. I didn't want to change the status quo here, as I didn't exactly know why it was done like that. But since we have the disabled state we might just as well use it.

KoBeWi (approved):

editor/plugins/asset_library_editor_plugin.cpp
YuriSizov: Okay, so it works. But the delay makes it awkward. You click the button, and for a couple of seconds nothing happens. So I'll keep it for the buttons that are supposed to force the setting immediately.

akien-mga:
Thanks! Great rework both design and feature wise :1st_place_medal: 

Let's see how the community likes it :)

Clonkex:
I generally like the redesign (although the logo feels tacked on somehow), but I think it's crazy to set defaults that are useful to only a tiny fraction of users. Seriously, how many people actually care about the privacy implications of a program checking for updates or downloading the latest news posts? It has to be in the 0.01% range _at most_, surely. Why wouldn't you make it a CLI option for those people and leave sensible defaults for the rest of us?

Think about it this way. If checking for updates automatically is a problem for your privacy, you're already well and truly screwed by a thousand other pieces of software. You couldn't use the internet at all.

AThousandShips:
So your logic is "you've already got tons of software tracking you, what's one more?" How is that helpful?

Also plenty of people care, despite it not being something you can fathom

Clonkex:
No, my logic is "Godot isn't tracking you so why would we make offline mode the default when 99.99% of people will want online mode?" Browsers and websites are the real privacy issue. Setting Godot (of all things) to offline mode by default literally solves nothing.

I get caring about privacy, but this isn't about privacy, this is about setting defaults that apply to only the smallest number of users. There's no way to justify that that I can see. You say it's not a hard thing to enable once? True! Why don't we just disable mouse input by default and give keyboard-only users the default they've always wanted? You only have to enable the setting once, right, so who cares that the default is only useful to a tiny fraction of users? (And yes, this is a massive exaggeration, but you get where I'm coming from, right? It just seems crazy to me.)

Truly caring about privacy is a _huge_ amount of work that very, _very_ few people are willing to put in. People just want convenience... even game devs using open source engines.

(This rant is courtesy of a long-time desire to see Linux become mainstream combined with frustration at Linux devs for their total and complete lack of understanding of and focus on what normal people actually want.)

AThousandShips:
You do understand that there's more reasons for not automatically going online right? Like not doing random downloads on a slow or metered connection...

But since you're making complete strawmen like:
> Why don't we just disable mouse input by default and give keyboard-only users the default they've always wanted?

I'll bid you a good day...

Clonkex:
> You do understand that there's more reasons for not automatically going online right? Like not doing random downloads on a slow or metered connection

More than most, yes! I've had slow, metered connections for many years and experienced the frustration of not being able to disable automatic updates. However, even on my most limited internet I never thought a program should start in some fully offline mode, only that it should have the option of limiting how much it downloads (mostly, I cared about not auto-updating). But that's entirely beside the point! Godot isn't wanting to do automatic updates, only automatic _checking_ for updates. Godot will never automatically update because it can't; it's a game engine and upgrading your project is often a major undertaking. And yet again, as I've said multiple times now, I'm _not_ suggesting we should remove the offline mode, only that it shouldn't be the default.

> But since you're making complete strawmen like

That's not a strawman. That's just getting across how ridiculous the choice seems to me. If you want to ignore my real arguments, that's cool with me. I've said my piece and I'm happy to leave it at that.

Calinou:
Remember that Godot is backed by a foundation based in EU, so we have to be GDPR-compliant at the end of the day. One of the points of GDPR is that you need to ask for explicit consent to do anything that deals with personal data somewhere in the pipeline.

Clonkex:
I'm not trying to be contrary, but that's a stretch. While IP addresses are typically considered PII, sending your IP address to a host without explicit consent is not automatically a breach of GDPR, especially if the IP isn't logged. You _don't_ need user consent so long as you have a "legitimate interest". Even if you're logging IPs, it probably counts as "legitimate interest" so long as you're doing it for a good reason (and if you're not logging anything it definitely counts). IANAL but see [article 6.1.f](https://gdpr.eu/article-6-how-to-process-personal-data-legally).

Either way, a Network Mode switch doesn't constitute user consent to processing of PII, so if that's your concern, this is not an appropriate way to resolve it.

Faless:
@Clonkex 

This change is only in part about privacy, and more about security.

For privacy, we probably want to link/display the [privacy policy](https://godotengine.org/privacy-policy/) before going online (so this PR is at least a step in the correct direction).

The security concern is about how certain functionalities of the asset library (and soon Editor Help - WIP in #69751), depends on content hosted on third party websites, over which we have no control.

In general, the Godot Asset Library does not store any content, which is instead hosted on github, gitlab, or other third party site.

This means we have no control over e.g. the images that are used as icons, or screenshots when displaying the asset list in the editor.

Since there have been in the past multiple remote code execution vulnerabilities in image handling libraries, we believe it's safer to not force the risk on the user immediately after selecting the  "Asset Library" tab (which can still be used to install local assets).

PR #87444, opened 2024-01-21
Tweak XROrigin3D documentation

Mickeon:
This PR slightly tweaks the documentation of **XROrigin3D**, linking to the other associated XR classes, rewording `world_scale` and-

I don't know, you tell me!

### DIFF
--- a/doc/classes/XROrigin3D.xml
+++ b/doc/classes/XROrigin3D.xml
@@ -5,20 +5,19 @@
 	</brief_description>
 	<description>
 		This is a special node within the AR/VR system that maps the physical location of the center of our tracking space to the virtual location within our game world.
-		There should be only one of these nodes in your scene and you must have one. All the XRCamera3D, XRController3D and XRAnchor3D nodes should be direct children of this node for spatial tracking to work correctly.
+		Multiple origin points can be added to the scene tree, but only one can used at a time. All the [XRCamera3D], [XRController3D], and [XRAnchor3D] nodes should be direct children of this node for spatial tracking to work correctly.
 		It is the position of this node that you update when your character needs to move through your game world while we're not moving in the real world. Movement in the real world is always in relation to this origin point.
-		For example, if your character is driving a car, the XROrigin3D node should be a child node of this car. Or, if you're implementing a teleport system to move your character, you should change the position of this node.
+		For example, if your character is driving a car, the [XROrigin3D] node should be a child node of this car. Or, if you're implementing a teleport system to move your character, you should change the position of this node.
 	</description>
 	<tutorials>
 		<link title="XR documentation index">$DOCS_URL/tutorials/xr/index.html</link>
 	</tutorials>
 	<members>
 		<member name="current" type="bool" setter="set_current" getter="is_current" default="false">
-			Is this XROrigin3D node the current origin used by the [XRServer]?
+			If [code]true[/code], this origin node is currently being used by the [XRServer]. Only one origin point can be used at a time.
 		</member>
 		<member name="world_scale" type="float" setter="set_world_scale" getter="get_world_scale" default="1.0">
-			Allows you to adjust the scale to your game's units. Most AR/VR platforms assume a scale of 1 game world unit = 1 real world meter.
-			[b]Note:[/b] This method is a passthrough to the [XRServer] itself.
+			The scale of the game world compared to the real world. This is the same as [member XRServer.world_scale]. By default, most AR/VR platforms assume that 1 game unit corresponds to 1 real world meter.
 		</member>
 	</members>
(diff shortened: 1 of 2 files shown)


doc/classes/XROrigin3D.xml
```diff
@@ -5,20 +5,19 @@
 	</brief_description>
 	<description>
 		This is a special node within the AR/VR system that maps the physical location of the center of our tracking space to the virtual location within our game world.
-		There should be only one of these nodes in your scene and you must have one. All the XRCamera3D, XRController3D and XRAnchor3D nodes should be direct children of this node for spatial tracking to work correctly.
+		There must be only one of these nodes in the scene tree at the time. All the [XRCamera3D], [XRController3D] and [XRAnchor3D] nodes should be direct children of this node for spatial tracking to work correctly.
```
AThousandShips, line 8:
```suggestion
		There must be only one of these nodes in the scene tree at the time. All the [XRCamera3D], [XRController3D], and [XRAnchor3D] nodes should be direct children of this node for spatial tracking to work correctly.
```

doc/classes/XROrigin3D.xml
```diff
@@ -5,20 +5,19 @@
 	</brief_description>
 	<description>
 		This is a special node within the AR/VR system that maps the physical location of the center of our tracking space to the virtual location within our game world.
-		There should be only one of these nodes in your scene and you must have one. All the XRCamera3D, XRController3D and XRAnchor3D nodes should be direct children of this node for spatial tracking to work correctly.
+		There must be only one of these nodes in the scene tree at the time. All the [XRCamera3D], [XRController3D], and [XRAnchor3D] nodes should be direct children of this node for spatial tracking to work correctly.
```
dsnopek, line 8: This isn't exactly right. Similar to `Camera3D`, there should only be one `XROrigin3D` that has `current` set to `true`, but multiple can exist in the scene tree at once.

One pattern is to load a new scene with its own `XROrigin` and related nodes, then when it's ready, set the new `XROrigin` to current, and then unload the old scene.

doc/classes/XROrigin3D.xml
```diff
@@ -5,20 +5,19 @@
 	</brief_description>
 	<description>
 		This is a special node within the AR/VR system that maps the physical location of the center of our tracking space to the virtual location within our game world.
-		There should be only one of these nodes in your scene and you must have one. All the XRCamera3D, XRController3D and XRAnchor3D nodes should be direct children of this node for spatial tracking to work correctly.
+		Multiple origin points can be added to the scene tree, but only one can used at the same time. All the [XRCamera3D], [XRController3D], and [XRAnchor3D] nodes should be direct children of this node for spatial tracking to work correctly.
```
dsnopek, line 8:
```suggestion
		Multiple origin points can be added to the scene tree, but only one can used at a time. All the [XRCamera3D], [XRController3D], and [XRAnchor3D] nodes should be direct children of this node for spatial tracking to work correctly.
```

Mickeon: "at the same time" in the class reference is more common than "at a time", but either way, it's fine.

dsnopek: To me this is a singular vs plural thing. Two (or more things) happen "at the same time", but one thing happens "at a time". But I'm from the US so there could be differences in other dialects

doc/classes/XROrigin3D.xml
```diff
@@ -5,20 +5,19 @@
...
-		For example, if your character is driving a car, the XROrigin3D node should be a child node of this car. Or, if you're implementing a teleport system to move your character, you should change the position of this node.
+		For example, if your character is driving a car, the [XROrigin3D] node should be a child node of this car. Or, if you're implementing a teleport system to move your character, you should change the position of this node.
 	</description>
 	<tutorials>
 		<link title="XR documentation index">$DOCS_URL/tutorials/xr/index.html</link>
 	</tutorials>
 	<members>
 		<member name="current" type="bool" setter="set_current" getter="is_current" default="false">
-			Is this XROrigin3D node the current origin used by the [XRServer]?
+			If [code]true[/code], this origin node is currently being used by the [XRServer]. Only one origin point can be used at the same time.
```
dsnopek, line 17:
```suggestion
			If [code]true[/code], this origin node is currently being used by the [XRServer]. Only one origin point can be used at a time.
```

dsnopek (approved):
Looks good to me. Thanks!

m4gr3d (approved):
Looks good!

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

PR #87446, opened 2024-01-21
Tweak Joint2D and Joint3D documentation for consistency

Mickeon:
Small tweaks to both **Joint2D** and **Joint3D** to make the wording consistent between the two. Also makes the description of `node_a` and `node_b` less misleading.

### DIFF
--- a/doc/classes/Joint2D.xml
+++ b/doc/classes/Joint2D.xml
@@ -4,7 +4,7 @@
 		Abstract base class for all 2D physics joints.
 	</brief_description>
 	<description>
-		Abstract base class for all joints in 2D physics. 2D joints bind together two physics bodies and apply a constraint.
+		Abstract base class for all joints in 2D physics. 2D joints bind together two physics bodies ([member node_a] and [member node_b]) and apply a constraint.
 	</description>
 	<tutorials>
 	</tutorials>
@@ -12,7 +12,7 @@
 		<method name="get_rid" qualifiers="const">
 			<return type="RID" />
 			<description>
-				Returns the joint's [RID].
+				Returns the joint's internal [RID] from the [PhysicsServer2D].
 			</description>
 		</method>
 	</methods>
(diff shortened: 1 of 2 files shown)


doc/classes/Joint3D.xml
```diff
@@ -13,19 +13,19 @@
...
 		</method>
 	</methods>
 	<members>
 		<member name="exclude_nodes_from_collision" type="bool" setter="set_exclude_nodes_from_collision" getter="get_exclude_nodes_from_collision" default="true">
-			If [code]true[/code], the two bodies of the nodes are not able to collide with each other.
+			If [code]true[/code], the two bodies bound together cannot collide with each other.
 		</member>
 		<member name="node_a" type="NodePath" setter="set_node_a" getter="get_node_a" default="NodePath(&quot;&quot;)">
-			The node attached to the first side (A) of the joint.
+			Path to the first node (A) attached to the joint. The node must inherit [PhysicsBody3D].
```
rburing, line 25: One of the node paths (A or B) is allowed to be empty, which has the effect of attaching the other body to an invisible static body without collision shapes, at the point given by the joint's position. (For example, that way you can easily make a pendulum with a PinJoint3D.) That should be phrased somehow nicely, maybe without the technical detail of the invisible static body without collision shapes. This also affects the description at the top of the file.

Mickeon: Talk about obscure behavior. I have yet to use joints in any of my projects and I would have never known about this, at least at a glance. You may check again.

Mickeon:
K-kindly removed the request from the documentation team. I LITERALLY AM THE DOCUMENTATION TEAM

akien-mga:
Well you'll have a hard time getting reviews without asking for reviews. The documentation team has more members than one ;)

mhilbrunner (approved):

doc/classes/Joint2D.xml
```diff
@@ -22,13 +22,15 @@
 			When set to [code]0[/code], the default value from [member ProjectSettings.physics/2d/solver/default_constraint_bias] is used.
 		</member>
 		<member name="disable_collision" type="bool" setter="set_exclude_nodes_from_collision" getter="get_exclude_nodes_from_collision" default="true">
-			If [code]true[/code], [member node_a] and [member node_b] can not collide.
+			If [code]true[/code], the two bodies bound together cannot collide with each other.
 		</member>
 		<member name="node_a" type="NodePath" setter="set_node_a" getter="get_node_a" default="NodePath(&quot;&quot;)">
-			The first body attached to the joint. Must derive from [PhysicsBody2D].
+			Path to the first body (A) attached to the joint. The node must inherit [PhysicsBody2D].
+			If left empty and [member node_b] is set, the body hangs as if attached to an invisible [StaticBody2D].
```
rburing, line 29: This is currently only true for `PinJoint2D`. See the `GodotPhysicsServer2D::joint_make_*` methods (in contrast to the 3D equivalents). I think we should make the implementation consistent with 3D (so this documentation description becomes true), if it makes sense.

doc/classes/Joint3D.xml
```diff
@@ -13,19 +13,21 @@
...
+			If left empty and [member node_b] is set, the body hangs as if attached to an invisible [StaticBody3D].
```
rburing, line 26: "Hanging" is not correct in all cases, e.g. when the joint has a motor enabled or if there is no gravity, so it should just be "is attached" or "is connected" or something generic like that. Also just "invisible" leaves open what/where this secret body is, so I think it should be e.g. "an invisible fixed [StaticBody3D] without collision shapes".

doc/classes/Joint2D.xml
Mickeon: Definitely. For now I can strip this detail from Joint2D but it should be an actual feature.

Mickeon:
Updated the PR to address the above issues, you may look again.

akien-mga (approved):

akien-mga:
Thanks!

PR #87458, opened 2024-01-22
Prevent tilemap editor hang and crash by mutex guarding tile_map_layer dirty.cell_list.

### LINKED ISSUE #86226: Crash when editing/opening a huge tileset atlas
### Tested versions

4.2.1

### System information

Windows 11 Godot 4.2.1

### Issue description

there seems to be a bug where it crashes when editing and opening a tile set with a huge set of textures

### Steps to reproduce

Create a new tileset -> Drag and drop the given atlas for editting = either going to be laggy, or crash

13_School_32x32

none

baptr:
TilesEditorUtils calls tile_map->set_pattern from a thread (to refresh tile pattern preview images). set_pattern sets each of the cells into the temporary map, marking them dirty via set_cell.

This can race against the tree invoking the TileMapLayer's internal_update, which proceses dirty.cell_list and eventually clears it. If a cell's shared dirty entry ends up losing the race, clear can get stuck in an infinite loop where remove refuses to change it because the entry doesn't think it's a member of the list, but the list head still points to it. This spams

```
ERROR: Condition "p_elem->_root != this" is true.
   at: remove (./core/templates/self_list.h:80)
```

And eventually leads to a crash when the error message buffer fills.

Guarding the list mutations avoids the race that sets up for this failure, fixes #86226.

* *Bugsquad edit, alternative to: https://github.com/godotengine/godot/pull/87470*

### DIFF
--- a/scene/2d/tile_map_layer.cpp
+++ b/scene/2d/tile_map_layer.cpp
@@ -2111,6 +2111,8 @@ void TileMapLayer::notify_tile_map_change(DirtyFlags p_what) {
 }
 
 void TileMapLayer::internal_update() {
+	MutexLock lock(dirty.mutex);
+
 	// Find TileData that need a runtime modification.
 	// This may add cells to the dirty list is a runtime modification has been notified.
 	_build_runtime_update_tile_data();
@@ -2187,6 +2189,7 @@ void TileMapLayer::set_cell(const Vector2i &p_coords, int p_source_id, const Vec
 	c.alternative_tile = alternative_tile;
 
 	// Make the given cell dirty.
+	MutexLock lock(dirty.mutex);
 	if (!E->value.dirty_list_element.in_list()) {
 		dirty.cell_list.add(&(E->value.dirty_list_element));
 	}
--- a/scene/2d/tile_map_layer.h
+++ b/scene/2d/tile_map_layer.h
@@ -262,6 +262,7 @@ private:
 
 	// Dirty flag. Allows knowing what was modified since the last update.
 	struct {
+		Mutex mutex;
 		bool flags[DIRTY_FLAGS_MAX] = { false };
 		SelfList<CellData>::List cell_list;
 	} dirty;


scene/2d/tile_map_layer.h
```diff
@@ -31,6 +31,7 @@
 #ifndef TILE_MAP_LAYER_H
 #define TILE_MAP_LAYER_H
 
+#include "core/os/mutex.h"
```
AThousandShips, line 34:
```suggestion
```
Already included

baptr: Only indirectly. I guess the "requires" in https://docs.godotengine.org/en/stable/contributing/development/code_style_guidelines.html#:~:text=Godot%20and%20third%2Dparty%20headers%20should%20be%20included%20in%20the%20file%20that%20requires%20them%2C%20i.e.%20in%20the%20.h%20header%20if%20used%20in%20the%20declarative%20code%20or%20in%20the%20.cpp%20if%20used%20only%20in%20the%20imperative%20code. really means it, and you don't include-what-you-use?

Done

AThousandShips: Yep, we avoid it as it clutters a lot

scene/2d/tile_map_layer.cpp
```diff
@@ -2111,6 +2111,8 @@ void TileMapLayer::notify_tile_map_change(DirtyFlags p_what) {
 }
 
 void TileMapLayer::internal_update() {
+	MutexLock lock(dirty.mutex);
```
AThousandShips, line 2114: This lock might be too strict and cause performance problems, would need to be evaluated, it locks quite a large block of code, might be better to lock more specifically to the critical sections

CC @RandomShaper

baptr: It basically just prevents set_cell from running during internal_update, which seemed pretty reasonable. ~all of the _ helper functions here can touch the dirty cell list, so it seemed pretty fairly scoped. I'm open to other ideas though.

AThousandShips: A lot of the stuff in those methods do things unrelated to it though, so unless those are also critical theras potential stuff there

Also unsure if this is the only affected list, or if some other solution is better here, i.e. preventing multiple threads from touching this at all, I'm not even sure how that is even happening here

scene/2d/tile_map_layer.cpp
```diff
@@ -2187,6 +2189,7 @@ void TileMapLayer::set_cell(const Vector2i &p_coords, int p_source_id, const Vec
 	c.alternative_tile = alternative_tile;
 
 	// Make the given cell dirty.
+	MutexLock lock(dirty.mutex);
 	if (!E->value.dirty_list_element.in_list()) {
 		dirty.cell_list.add(&(E->value.dirty_list_element));
 	}
```
AThousandShips, line 2195:
```suggestion
	{
		MutexLock lock(dirty.mutex);
		if (!E->value.dirty_list_element.in_list()) {
			dirty.cell_list.add(&(E->value.dirty_list_element));
		}
	}
```
I'd at least reduce this one, the rest should be looked into

AThousandShips: Though there *is* a possible race condition between this whole method and the update one, as it looks up an iterator in `tile_map` above which is touched by the update, I think this whole class needs some thread safety evaluation

AThousandShips: There *is* threading for the `TileMap` editor and specifically the pattern generation, I suspect this is what's causing this and might be the area to look into, rather than adding it to the `TileMap` itself, unsure how to handle that part specifically though

AThousandShips: I suspect this code is the problem:
https://github.com/godotengine/godot/blob/0bcc0e92b3f0ac57d4c4650722f347593a258572/editor/plugins/tiles/tiles_editor_plugin.cpp#L70-L144

But unsure how to approach it exactly, might work by adding a `update_internals()` after setting the pattern, or doing some other deferred operation here, will check if I can replicate the crash later today and see about this as a solution

AThousandShips:
Opened an alternative PR that handles this by making the editor preview thread safe instead, this solves the crash and generates the previews properly

We might want to look into general thread safety of this class as well, but I think this issue should be solved by making the editor management of this thread safe instead

YuriSizov:
Superseded by https://github.com/godotengine/godot/pull/87470. Thanks for your contribution nevertheless!

PR #87462, opened 2024-01-22
Fix PopupMenu doesn't respect its ScrollContainer's margins

### LINKED ISSUE #87461: PopupMenu doesn't respect the content margins of it's ScrollContainer's StyleBox
### Tested versions

4.2.1

### System information

Windows 10

### Issue description

The PopupMenu doesn't respect it's child node ScrollContainer's Panel stylebox minimum size.

### Steps to reproduce

create a MenuButton
create a theme for the MenuButton and change the margins of the ScrollContainer panel StyleBox

[PopupMenuTest.zip](https://github.com/godotengine/godot/files/14005422/PopupMenuTest.zip)

WhalesState:
Fixes https://github.com/godotengine/godot/issues/87461

Removed the `MarginContainer` and applied the `PopupMenu` panel style to the `ScrollContainer`.

This results in the same old behavior, instead of applying the panel's content margins to the margin container and drawing the panel style manually.

Removing the include of the margin container header file has resulted in errors in 8 other files which was depending on it, so I have included the header in each one of them.

### DIFF
--- a/scene/gui/popup_menu.cpp
+++ b/scene/gui/popup_menu.cpp
@@ -213,8 +213,8 @@ Size2 PopupMenu::_get_item_icon_size(int p_idx) const {
 }
 
 Size2 PopupMenu::_get_contents_minimum_size() const {
-	Size2 minsize = theme_cache.panel_style->get_minimum_size(); // Accounts for margin in the margin container
-	minsize.x += scroll_container->get_v_scroll_bar()->get_size().width * 2; // Adds a buffer so that the scrollbar does not render over the top of content
+	Size2 minsize = theme_cache.panel_style->get_minimum_size();
+	minsize.width += scroll_container->get_v_scroll_bar()->get_size().width;
 
 	float max_w = 0.0;
 	float icon_w = 0.0;
@@ -304,16 +304,11 @@ int PopupMenu::_get_items_total_height() const {
 }
 
 int PopupMenu::_get_mouse_over(const Point2 &p_over) const {
-	if (p_over.x < 0 || p_over.x >= get_size().width) {
+	if (p_over.x < 0 || p_over.x >= get_size().width || p_over.y < theme_cache.panel_style->get_margin(Side::SIDE_TOP)) {
 		return -1;
 	}
 
-	// Accounts for margin in the margin container
-	Point2 ofs = theme_cache.panel_style->get_offset();
-
-	if (ofs.y > p_over.y) {
-		return -1;
-	}
+	Point2 ofs;
 
 	for (int i = 0; i < items.size(); i++) {
 		ofs.y += theme_cache.v_separation;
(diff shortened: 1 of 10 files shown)


WhalesState:
> or (better), the internal ScrollContainer should have an override for its panel style to force-disable it, because it's redundant

I'm trying to do this without creating a new `StyleBoxEmpty` in every new instance. but I don't know how. and if we use the same PopupMenu panel style from theme cache it will be drawn twice.

WhalesState:
I have removed the include of the `margin_container.cpp` after testing the new changes, so those errors needs to be fixed in the `color_picker.cpp`.

scene/gui/popup_menu.cpp
```diff
@@ -694,10 +694,6 @@ void PopupMenu::_draw_items() {
 	control->set_custom_minimum_size(Size2(0, _get_items_total_height()));
 	RID ci = control->get_canvas_item();
 
-	Size2 margin_size;
```
WhalesState, line 697: `margin_size` was calculated but not used.

scene/gui/popup_menu.cpp
```diff
@@ -1015,6 +1006,8 @@ void PopupMenu::_notification(int p_what) {
 		case NOTIFICATION_THEME_CHANGED:
 		case Control::NOTIFICATION_LAYOUT_DIRECTION_CHANGED:
 		case NOTIFICATION_TRANSLATION_CHANGED: {
+			scroll_container->add_theme_style_override("panel", theme_cache.panel_style);
```
WhalesState, line 1009: I just wanted to set the `ScrollContainer` panel style in NOTIFICATION_THEME_CHANGED, but the three of them are linked, and it will result in creating and extra function or duplicating code.

editor/plugins/font_config_plugin.h
```diff
@@ -35,6 +35,7 @@
 #include "editor/editor_plugin.h"
 #include "editor/editor_properties.h"
 #include "editor/editor_properties_array_dict.h"
+#include "scene/gui/margin_container.h"
```
KoBeWi, line 38: This can be in cpp file.

scene/gui/color_picker.h
```diff
@@ -38,6 +38,7 @@
 #include "scene/gui/grid_container.h"
 #include "scene/gui/label.h"
 #include "scene/gui/line_edit.h"
+#include "scene/gui/margin_container.h"
```
KoBeWi, line 41: This can also be in cpp file.

scene/gui/popup_menu.cpp
KoBeWi: You can use `[[falthrough]]`.

KoBeWi (dismissed):
Left some comments, otherwise looks fine.

WhalesState:
while testing i have found some bugs related to items and size and it took me a long time to make sure that the popup menu will work fine when changing the item's `v_separation` and the `panel_style`'s content margins.

I have noticed that the ScrollContainer is no longer drawing the scroll over it's children, and it do the margins automatically for the Scroll Bar, so I have removed the extra space used for the Vertical Scroll Bar since it will be added automatically when it's visible.

After applying the stylebox directly to the `ScrollContainer` and testing with large `v_separation` value to force the scroll to appear, i have noticed some issues and i have changed the calculactions to take into account the new items size with `v_separation` and the correct mouse hover while taking into account the panel's top margin.

KoBeWi:
Thanks for reminding, I actually forgot to test the scrollbar 🤦‍♂️
And it made me find 2 regressions:
- the ScrollBar can't be clicked. Clicking it will activate an item under it
- when opening menu at the edge of the screen, the scrollbar goes outside the screen

WhalesState:
I have found that this issue presents in all github windows builds, but it doesn't exist when building locally without `d3d12=yes` so it may be an issue related to d3d 12:
- the ScrollBar can't be clicked. Clicking it will activate an item under it (can be tested with language list in editor settings)

Fixed:
- when opening menu at the edge of the screen, the scrollbar goes outside the screen (can be tested with resource list in the inspector)

Another issue:
- when dragging the menu slider and releasing the mouse over an item, the item gets selected and the menu hides. on the next menu popup the slider dragging continues after popup because it doesn't recieve the mouse release event.

WhalesState:
> * the ScrollBar can't be clicked. Clicking it will activate an item under it (can be tested with language list in editor settings)

This issue appeard after rebasing the branch, i think it's related to this https://github.com/godotengine/godot/pull/86952 pull request. I will try to fix it.

WhalesState:
please have a look on this https://github.com/godotengine/godot/pull/88392#issuecomment-1953497134

KoBeWi:
Needs rebase after #86952 was reverted.

KoBeWi (approved):

akien-mga:
Thanks!

PR #87464, opened 2024-01-22
[Linux] Add support for PCK embedding section with non GNU-ld linkers.

### LINKED ISSUE #87438: PCK section missing from compiled templates
### Tested versions

4.3.dev HEAD

### System information

Arch Linux

### Issue description

When exporting with embedded PCK an error is produced saying the PCK section is missing. Probing the resultant executable file with objdump -x you can indeed see there is no pck section in the ELF file.

### Steps to reproduce

This is with the MYSQL extension installed. To get it to compile on Arch I had to use LLVM and LLD, which may be the issue?

So... compile the templates with LLVM and LDD and examine the results to see if the pck section exists in them.

No need for a project. Just compile the templates from the source tree with LLVM and LLD.

bruvzg:
Fixes https://github.com/godotengine/godot/issues/87438

Replaces linker scripts with section data definition in the code (same as used on Windows), this seems to work with `mold` and `lld` (including LTO builds).


akien-mga (approved):
Looks good to me.

platform/linuxbsd/godot_linuxbsd.cpp
```diff
@@ -41,6 +41,18 @@
...
+// For export templates, add a section; the exporter will patch it to enclose
+// the data appended to the executable (bundled PCK)
+#if !defined(TOOLS_ENABLED) && defined(__GNUC__)
+static const char dummy[8] __attribute__((section("pck"), used)) = { 0 };
+
+// Dummy function to prevent LTO from discarding "pck" section.
+extern "C" const char *pck_section_dummy_call() __attribute__((used));
+extern "C" const char *pck_section_dummy_call() {
+	return &dummy[0];
+};
```
AThousandShips, line 53:
```suggestion
}
```

deralmas (approved):
I don't have much experience with linking so TIWAGOS but this looks like a win overall.

FTR, mold has some reasoning behind the absence of linker script support:

> Mold supports a very limited set of linker script features, which is just sufficient to read /usr/lib/x86_64-linux-gnu/libc.so on Linux systems (on Linux, that file is contrary to its name not a shared library but an ASCII linker script that loads a real libc.so file.)
>
> Beyond that, we have no plan to support any additional linker script features. The linker script is an ad-hoc, over-designed, complex language which we believe needs to be replaced by a simpler mechanism. We have a plan to add a replacement for the linker script to mold instead.

platform/linuxbsd/godot_linuxbsd.cpp
```diff
@@ -41,6 +41,18 @@
 #include <sys/resource.h>
 #endif
 
+// For export templates, add a section; the exporter will patch it to enclose
+// the data appended to the executable (bundled PCK)
```
akien-mga, line 45: I didn't point it out initially because there was the same issue in the Windows code, but if we're doing style fixes there, let's also handle this one:
```suggestion
// For export templates, add a section; the exporter will patch it to enclose
// the data appended to the executable (bundled PCK).
```

akien-mga:
Cherry-picked for 3.6.

PR #87470, opened 2024-01-22
[Editor] Fix threading problems with `TileMap` preview

### LINKED ISSUE #86226: Crash when editing/opening a huge tileset atlas

AThousandShips:
As investigated in #87458 this is due to threading problems in the editor, specifically the preview generation, this solves this by handling deferred calls on a separate queue, rather than adding thread safety mechanisms to `TileMapLayer` which I'd say aren't necessary otherwise due to thread safety guarantees, I also suspect there are other possible race conditions here that won't be solved by adding those thread safety features, instead I suggest removing the thread safety from the equation entirely by controlling the timing

The general topic of thread safety of nodes not on the tree is a problem and other nodes suffer from problems due to deferred calls AFAIK, so a broader investigation of that and how `TileMap` and `TileMapLayer` relates to this should be done as well, but I think the pressing concern here is the editor related crash

* Fixes: https://github.com/godotengine/godot/issues/86226

Alternative to:
* https://github.com/godotengine/godot/pull/87458

### DIFF
--- a/editor/plugins/tiles/tiles_editor_plugin.cpp
+++ b/editor/plugins/tiles/tiles_editor_plugin.cpp
@@ -68,6 +68,9 @@ void TilesEditorUtils::_thread_func(void *ud) {
 }
 
 void TilesEditorUtils::_thread() {
+	CallQueue queue;
+	MessageQueue::set_thread_singleton_override(&queue);
+
 	pattern_thread_exited.clear();
 	while (!pattern_thread_exit.is_set()) {
 		pattern_preview_sem.wait();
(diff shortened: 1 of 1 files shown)


editor/plugins/tiles/tiles_editor_plugin.cpp
```diff
@@ -127,6 +130,8 @@ void TilesEditorUtils::_thread() {
 				// Add the viewport at the last moment to avoid rendering too early.
 				callable_mp((Node *)EditorNode::get_singleton(), &Node::add_child).call_deferred(viewport, false, Node::INTERNAL_MODE_DISABLED);
 
+				MessageQueue::get_singleton()->flush();
```
AThousandShips, line 133: Unsure if the better style is this or:
```suggestion
				queue.flush();
```

RandomShaper: An interesting question. I'd favor the cumbersome syntax, because, even if in this case you have the pseudo-singleton variable at hand, a more explicit syntax is more consistent with the typical usage of message queues.

RandomShaper (approved):
Notwithstanding the idea of further investigation, and with the disclaimer that I haven't checked the details, this looks like the right thing to do.

YuriSizov:
Thanks!

AThousandShips:
Thank you!

YuriSizov:
Cherry-picked for 4.2.2.

PR #87476, opened 2024-01-22
Fix selection regression in `EditorHelpSearch`

### LINKED ISSUE #87472: Open button gets disabled for some help classes when searching
### Tested versions

4.3 dev 0bcc0e9
Does not seem to happen in 4.3 dev2

### System information

### Issue description

https://github.com/godotengine/godot/assets/2223172/4d71ec7a-1749-4227-bb6e-644ada6fc937

Seems to affect abstract classes. You need to select another class and back to enable opening.

### Steps to reproduce

1. Press F1
2. Type "StyleBox"
3. Press Open

N/A

AThousandShips:
Investigated clearing the selection details on removing a `TreeItem` but think that's a bit of a larger investigation so keeping this restricted

The bug was caused by the removed item being currently selected, and then failing on re-select here:
https://github.com/godotengine/godot/blob/0bcc0e92b3f0ac57d4c4650722f347593a258572/scene/gui/tree.cpp#L2615-L2617

* Fixes: https://github.com/godotengine/godot/issues/87472

### DIFF
--- a/editor/editor_help_search.cpp
+++ b/editor/editor_help_search.cpp
@@ -445,6 +445,9 @@ bool EditorHelpSearch::Runner::_phase_match_classes() {
 }
 
 void EditorHelpSearch::Runner::_populate_cache() {
+	// Deselect to prevent re-selection issues.
+	results_tree->deselect_all();
+
 	root_item = results_tree->get_root();
 
 	if (root_item) {


AThousandShips:
My alternative fix was:
```diff
diff --git a/scene/gui/tree.cpp b/scene/gui/tree.cpp
index c67c3cd98d..fa22313d14 100644
--- a/scene/gui/tree.cpp
+++ b/scene/gui/tree.cpp
@@ -121,6 +121,9 @@ void TreeItem::_change_tree(Tree *p_tree) {

                if (tree->selected_item == this) {
                        tree->selected_item = nullptr;
+                       for (int i = 0; i < p_item->cells.size(); i++) {
+                               cells.write[i].selected = false;
+                       }
                }

                if (tree->drop_mode_over == this) {
```
Which also solves this, unsure which is best but as that one has some potentially wider implications I kept it simple here

YuriSizov:
> Which also solves this, unsure which is best but as that one has some potentially wider implications I kept it simple here

I prefer the local solution, so I think your PR is fine. As long as it doesn't produce any noticeable flickering.

AThousandShips:
It happens as it clears the tree so I don't think it will, didn't notice any myself (other than the existing rendering issues with open popups)

YuriSizov (approved):

YuriSizov:
Thanks!

AThousandShips:
Thank you!

PR #87478, opened 2024-01-22
Prevent threading problems in `TileMap`

### LINKED ISSUE #87402: Instantiate big Tilemap on thread dont work.
### Tested versions

Godot 4.1 and 4.2

### System information

Macbook Pro M2

### Issue description

When instantiate a big Tilemap on a separate thread and try to return it to the main thread, issues occur. I've encounter various bugs while trying to find a working solution.  It works fine if i call Thread::wait_to_finish() on the main thread. But passing the built scene with call_deferred, problems occur. 

On the attached project, the game freezes on 4.2. Ive also encountered crashes (mostly on 4.1) as shownhere:

```
Engine version: Godot Engine v4.2.1.stable.official (b09f793f564a6c95dc76acc654b390e68441bd01)
[1] 1   libsystem_platform.dylib            0x00000001834f2a84 _sigtramp + 56
[2] SpriteFrames::get_frame_duration(StringName const&, int) const
[3] NavigationObstacle2D::get_vertices() const
[4] NavigationObstacle2D::get_vertices() const
[5] NavigationObstacle2D::get_vertices() const
[6] VariantInitializer<Vector<Color>>::init(Variant*)
[7] Object* ClassDB::creator<ENetConnection>()
[8] Object::get_instance_id() const
[9] JSON::get_data() const
[10] VariantInitializer<Vector<Color>>::init(Variant*)
[11] Object* ClassDB::creator<ENetConnection>()
[12] Object::get_instance_id() const
[13] JSON::get_data() const
[14] VariantInitializer<Vector<Color>>::init(Variant*)
[15] Object* ClassDB::creator<ENetConnection>()
[16] MultiplayerAPI::is_server()
[17] AnimationPlayerEditor::unpin()
[18] CallableCustomExtension::default_compare_less(CallableCustom const*, CallableCustom const*)
[19] SceneTree::get_root() const
[20] Node::get_index(bool) const
[21] Node::get_index(bool) const
[22] RendererCompositorRD::_create_current()
[23] void rx::mtl::GetMatrixUniformMetal<unsigned int>(unsigned int, unsigned int*, unsigned int const*, bool)
[24] RendererCompositorRD::_create_current()
[25] 25  dyld                                0x000000018316bf28 start + 2236
```

What is going on, is this a bug? Or how do i solve it. One must be able to instantiate a bigger scene on a separate thread.

EDIT:

On the current attached minimal project i get the following error multiple times in the log-file:

```
USER ERROR: Condition "p_elem->_root != this" is true.
   at: remove (./core/templates/self_list.h:80)
```

It does however work with smaller tilemaps

### Steps to reproduce

Create a big TileMap and try to loading it and instantiate it using a Thread.

[playground.zip](https://github.com/godotengine/godot/files/13997588/playground.zip)

AThousandShips:
Still needing to rework the places this is used and do some deeper testing but just to do some testing with this base and for some input

Will also update documentation etc. to clarify that it doesn't update automatically when not in the tree but that manual updates are required

Follow-up/related to (which it reverts on `master` as its no longer needed):
* https://github.com/godotengine/godot/pull/87470

See also:
* https://github.com/godotengine/godot/issues/86226

**Note**: This does *not* make `TileMap` thread safe, like all nodes they aren't safe to manipulate from multiple threads, or from a thread when in the scene tree

What it does is make it possible to use it in a thread safe fashion, by preventing it from being manipulated on the main thread due to deferred updates

---

* Fixes: https://github.com/godotengine/godot/issues/87402

### DIFF
--- a/doc/classes/TileMap.xml
+++ b/doc/classes/TileMap.xml
@@ -5,7 +5,7 @@
 	</brief_description>
 	<description>
 		Node for 2D tile-based maps. Tilemaps use a [TileSet] which contain a list of tiles which are used to create grid-based maps. A TileMap may have several layers, layouting tiles on top of each other.
-		For performance reasons, all TileMap updates are batched at the end of a frame. Notably, this means that scene tiles from a [TileSetScenesCollectionSource] may be initialized after their parent.
+		For performance reasons, all TileMap updates are batched at the end of a frame. Notably, this means that scene tiles from a [TileSetScenesCollectionSource] may be initialized after their parent. This is only queued when inside the scene tree.
 		To force an update earlier on, call [method update_internals].
 	</description>
 	<tutorials>
--- a/scene/2d/tile_map_layer.cpp
+++ b/scene/2d/tile_map_layer.cpp
@@ -1625,8 +1625,11 @@ void TileMapLayer::_queue_internal_update() {
 	if (pending_update) {
 		return;
 	}
-	pending_update = true;
-	callable_mp(this, &TileMapLayer::_deferred_internal_update).call_deferred();
+	// Don't update when outside the tree, it doesn't do anything useful, and causes threading problems.
+	if (is_inside_tree()) {
+		pending_update = true;
+		callable_mp(this, &TileMapLayer::_deferred_internal_update).call_deferred();
+	}
 }
 
 void TileMapLayer::_deferred_internal_update() {
@@ -1695,7 +1698,8 @@ void TileMapLayer::_notification(int p_what) {
 
 		case NOTIFICATION_EXIT_TREE: {
 			dirty.flags[DIRTY_FLAGS_LAYER_IN_TREE] = true;
-			_queue_internal_update();
+			// Update immediately on exiting.
+			update_internals();
 		} break;
 
 		case TileMap::NOTIFICATION_ENTER_CANVAS: {
@@ -1705,7 +1709,8 @@ void TileMapLayer::_notification(int p_what) {
 
 		case TileMap::NOTIFICATION_EXIT_CANVAS: {
 			dirty.flags[DIRTY_FLAGS_LAYER_IN_CANVAS] = true;
-			_queue_internal_update();
+			// Update immediately on exiting.
+			update_internals();
 		} break;
 
 		case TileMap::NOTIFICATION_VISIBILITY_CHANGED: {
--- a/editor/plugins/tiles/tiles_editor_plugin.cpp
+++ b/editor/plugins/tiles/tiles_editor_plugin.cpp
@@ -69,9 +69,6 @@ void TilesEditorUtils::_thread_func(void *ud) {
 }
 
 void TilesEditorUtils::_thread() {
-	CallQueue queue;
-	MessageQueue::set_thread_singleton_override(&queue);
-
 	pattern_thread_exited.clear();
 	while (!pattern_thread_exit.is_set()) {
 		pattern_preview_sem.wait();
@@ -131,8 +128,6 @@ void TilesEditorUtils::_thread() {
 				// Add the viewport at the last moment to avoid rendering too early.
 				callable_mp((Node *)EditorNode::get_singleton(), &Node::add_child).call_deferred(viewport, false, Node::INTERNAL_MODE_DISABLED);
 
-				MessageQueue::get_singleton()->flush();
-
 				RS::get_singleton()->connect(SNAME("frame_pre_draw"), callable_mp(const_cast<TilesEditorUtils *>(this), &TilesEditorUtils::_preview_frame_started), Object::CONNECT_ONE_SHOT);
 
 				pattern_preview_done.wait();
@@ -145,11 +140,7 @@ void TilesEditorUtils::_thread() {
 				viewport->queue_free();
 			}
 		}
-
-		MessageQueue::get_singleton()->flush();
 	}
-
-	MessageQueue::get_singleton()->flush();
 	pattern_thread_exited.set();
 }
 


AThousandShips:
Due to the various effects of this I'm not marking it for cherry-picking and keeping https://github.com/godotengine/godot/pull/87470 as the main fix for the editor crashes

I'd say the possible drawbacks of not being able to have the map update when not in the tree are worth the thread safety improvements (without having to make the whole class thread safe), the main problem is that even when not in the tree it isn't thread safe as the update is deferred and can collide with threaded processing, especially when adding a lot of tiles for example adding a pattern

AThousandShips:
(reverted the changes from #87470 for this as they are no longer needed if this works, and it just takes up some performance and memory, but as that fix is cherry-pickable and I don't feel confident cherry-picking this one at least not without extensive testing, and it does change behavior)

AThousandShips:
Not fully ready by a bit yet but marking ready for review to get some eyes on it, still needs some testing and cleaning up and decisions

This changes behavior in cases where it is already thread safe, i.e. when modified from the main thread, when not in the scene tree

In these cases, where, for example, you update a tile map incrementally but on the main thread, outside of the scene tree, the updated internals won't be available even if you do it, for example, in `_process`, this *was* the case normally, and *does* work fully safely

I'm not feeling confident that this change won't cause confusion in these specific cases, or in "partially" thread safe use, i.e. when you're lucky and it doesn't crash (which is largely the case when you don't do large batch updates), as you now have to explicitly call `update_internals`

But I'm not familiar enough with all the details to know what if any of that data is relevant outside of the tree, if it isn't I'd say to also cherry pick this one

scene/2d/tile_map_layer.cpp
```diff
@@ -1630,7 +1630,8 @@ void TileMapLayer::_notification(int p_what) {
 		case NOTIFICATION_EXIT_TREE: {
 			dirty.flags[DIRTY_FLAGS_LAYER_IN_TREE] = true;
 			TileMap *tile_map_node = _fetch_tilemap();
-			tile_map_node->queue_internal_update();
+			// Update immediately on exiting, alternatively we can simply skip this update and let it be done on entering or on request.
+			tile_map_node->update_internals();
```
AThousandShips, line 1634: This part is a point where I'm unsure about the exact solution

scene/2d/tile_map.cpp
```diff
@@ -229,8 +229,10 @@ void TileMap::queue_internal_update() {
 	if (pending_update) {
 		return;
 	}
-	pending_update = true;
-	callable_mp(this, &TileMap::_internal_update).call_deferred();
+	if (is_inside_tree()) {
```
AThousandShips, line 232: Could move this to `if (pending_update) {` as `if (pending_update || !is_inside_tree()) {`

AThousandShips:
At least temporarily rebased on top of:
* #87379

For easier review based on those changes

scene/2d/tile_map_layer.cpp
```diff
@@ -1623,32 +1687,29 @@ void TileMapLayer::_notification(int p_what) {
...
-			tile_map_node->queue_internal_update();
+			_queue_internal_update();
 		} break;
 
 		case NOTIFICATION_EXIT_TREE: {
 			dirty.flags[DIRTY_FLAGS_LAYER_IN_TREE] = true;
-			TileMap *tile_map_node = _fetch_tilemap();
-			tile_map_node->queue_internal_update();
+			// Update immediately on exiting, alternatively we can simply skip this update and let it be done on entering or on request.
+			update_internals();
```
AThousandShips, line 1696: This part is a point where I'm unsure about the exact solution

scene/2d/tile_map_layer.cpp
```diff
@@ -1614,6 +1616,68 @@ void TileMapLayer::_update_notify_local_transform() {
...
 }
 
+void TileMapLayer::_queue_internal_update() {
+	if (pending_update) {
+		return;
+	}
+	if (is_inside_tree()) {
+		pending_update = true;
+		callable_mp(this, &TileMapLayer::_deferred_internal_update).call_deferred();
+	}
```
AThousandShips, line 1626: Could move this to `if (pending_update) {` as `if (pending_update || !is_inside_tree()) {`

akien-mga:
CC @groud

doc/classes/TileMap.xml
```diff
@@ -5,8 +5,8 @@
 	</brief_description>
 	<description>
 		Node for 2D tile-based maps. Tilemaps use a [TileSet] which contain a list of tiles which are used to create grid-based maps. A TileMap may have several layers, layouting tiles on top of each other.
-		For performance reasons, all TileMap updates are batched at the end of a frame. Notably, this means that scene tiles from a [TileSetScenesCollectionSource] may be initialized after their parent.
-		To force an update earlier on, call [method update_internals].
+		For performance reasons, all TileMap updates are batched at the end of a frame. Notably, this means that scene tiles from a [TileSetScenesCollectionSource] may be initialized after their parent. This is only queued when inside the scene tree.
+		To force an update earlier on, or when not in the scene tree, call [method update_internals].
```
groud, line 9: Forced updates are useless when a TileMapLayer is outside the tree, so I don't think I would mention it.

scene/2d/tile_map_layer.cpp
groud: No that is the right solution. You need to trigger an update when leaving the tree to free the CanvasItems, physics shapes, etc... anyway. Users would complain if removing a TileMap from the tree would still have CanvasItems or collisions shapes visible/colliding.

doc/classes/TileMap.xml
AThousandShips: So the data updated is not relevant except for the physics/rendering/navigation/etc.? Not for the data accessible/usable by users? If so then this should be sufficient to work things out, and might not need forcing any updates on tree exit

scene/2d/tile_map_layer.cpp
groud: I think it is clear enough that way. I would probably add a comment about why we don't update while not being in the tree. Basically: 1) it's useless and 2) it's creating issues.

doc/classes/TileMap.xml
groud: Yes. These updates are for runtime only, when the TileMap is in the tree. All subsystems updates just clear everything when TileMapLayer isn't in the tree. 

So those:
```
	_build_runtime_update_tile_data();

	// Update all subsystems.
	_rendering_update();
	_physics_update();
	_navigation_update();
	_scenes_update();
#ifdef DEBUG_ENABLED
	_debug_update();
#endif // DEBUG_ENABLED

	_clear_runtime_update_tile_data();	
```

Are forced to cleanup everything using something like:
```
	bool forced_cleanup = in_destructor || !enabled || !tile_set.is_valid() || !is_visible_in_tree();
```
or similar. (here, is_visible_in_tree ensure we clean up if the node is not in the tree)

scene/2d/tile_map_layer.cpp
groud: Ah and we probably want to remove the comment here. At least the "alternatively part" I guess :sweat_smile:

AThousandShips:
There, removed the additional note in the documentation, cleaned up the comments, and added a comment to the blocked update

scene/2d/tile_map_layer.cpp
AThousandShips: Already done :)

groud (approved):

akien-mga:
Thanks!

AThousandShips:
Thank *you*!

grusad:
Hey. Was just testing this one out from Godot 4.3 Dev 4 build and it didn't seem to solve the issue i have regarding instantiating big Tilemaps on separate thread. [#87402](https://github.com/godotengine/godot/issues/87402)

I still get the logs spammed with this message:
USER ERROR: Condition "p_elem->_root != this" is true.
   at: remove (./core/templates/self_list.h:80)

From what i saw, this PR is in the Dev 4 build.

AThousandShips:
Left a comment on that issue, I'm not sure it's entirely a bug as [resources are not thread safe](https://docs.godotengine.org/en/stable/tutorials/performance/thread_safe_apis.html#resources), but worth investigating

Issue #83795, opened 2023-10-22
Evaluate Expression shortcut is no longer working in 4.2 and also breaks the script editor


MewPurPur:
### Godot version

Godot v4.2.beta2 - Pop!_OS 22.04 LTS - X11 - Vulkan (Forward+) - integrated Intel(R) Graphics (ADL GT2) () - 12th Gen Intel(R) Core(TM) i5-1235U (12 Threads)

### System information

Godot v4.2.beta2 - Pop!_OS 22.04 LTS - X11 - Vulkan (Forward+) - integrated Intel(R) Graphics (ADL GT2) () - 12th Gen Intel(R) Core(TM) i5-1235U (12 Threads)

### Issue description

Evaluate Expression in the script editor is no longer working in latest 4.2. It's still working from the context menu, but not when I do Ctrl+Shift+E.

Not only this, but it breaks the script editor and you can only navigate through it after that, like it's read-only.

### Steps to reproduce

Type `PI/2` in the script editor and do `Ctrl+Shift+E`.

Expectation: `1.57`

Reality: "(2) Caller thread can't call this function in this node (/root). Use call_deferred() or call_thread_group() instead."

### Minimal reproduction project

N/A

KoBeWi:
Can't reproduce in beta3.

MewPurPur:
I can still reproduce this in beta3.

KoBeWi:
https://github.com/godotengine/godot/assets/2223172/999d8711-8334-4d28-b1a4-1f9ed077ce65

Then maybe it requires some extra steps?

MewPurPur:
It's exactly as I described and I can reproduce it consistently in every project. Maybe my machine has something to do with this?

akien-mga:
I also can't reproduce on Linux (Mageia 9, KDE on X11, using default editor settings).

Could you test with default editor settings / in self-contained mode, to see if that's caused by an editor setting?

MewPurPur:
I didn't do that, but I think I dug up what might have been the cause: https://askubuntu.com/questions/1083913/what-does-ctrl-shift-e-do-while-typing-text

I don't get the errors I mentioned in my original post, and I don't get easily locked out, and the underlined "e" appears as described in the linked post, though the functionality doesn't work as described.

When in this state, I can still find ways to lock myself out of the code editor, although again, not as easily as before. A lot of the same issues also happen when I press Ctrl+Shift+U.

Example of a softlock:

[...]

Issue #86863, opened 2024-01-06
TextEdit multicaret and selection issues


kitbdev:
### Tested versions

Godot v4.2, v4.3.dev (179dfdc8d)

### System information

Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2070 (NVIDIA; 31.0.15.3598) - Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz (12 Threads)

### Issue description

Some of these issues are only relevent to `CodeEdit`, `CodeEditor`, or `ScriptTextEditor`, but most stem from `TextEdit`.

related issues:
- #73202
- #71217
- #72410
- #83589
- #81535
- #83410 
- #72797
- https://github.com/godotengine/godot/issues/80378

related PRs:
- #61902
- #72671
- #69132
- #73704
- #69617

I am working on a PR to fix all of these issues.

API <a id="api" href="#api">#</a>
- Lots of methods don't redraw.
	- `add_caret`, `remove_caret`, `set_caret_line`, `set_caret_col`, `set_line`, `remove_text`, etc.
	- This is noticeable when `caret_blink` is off.
- Passing out of bounds `caret_index` to various functions doesn't error for indexes at carets.size() or below -1. 
	- For `insert_text_at_caret`, `select_word_under_caret`, `select_word_under_caret`, `delete_selection`, `_handle_unicode_input_internal`, `_backspace_internal`, `_cut_internal`, `_copy_internal`, `_paste_internal`, `_paste_primary_clipboard_internal`, `get_word_under_caret`, `has_selection`, `get_selected_text`.
- `remove_text()` doesn't adjust carets. It is mentioned in the documentation, but it is not consistent with all other methods.
- `set_line()` doesn't properly adjust carets. selection from is ignored and passing in multiple lines doesn't adjust carets below.
- Adding multiple lines with `insert_line_at` doesn't offset carets after the inserted line correctly.
	- When `insert_line_at` is called with the subsequent line, they are merged together so undo will undo both calls.
- When the TextEdit is out of the tree, editing multiple carets multiple times does not work as expected and can show errors. This is because `caret_index_edit_dirty` is only set when in the tree, and after the first edit, getting the caret order won't be up to date.

Backspace <a id="backspace" href="#backspace">#</a>
[...]

PR #87479, opened 2024-01-22
Fix TextEdit IME issues

kitbdev:
- Fixes IME issues in https://github.com/godotengine/godot/issues/86863
- Extracted from #86978

IME text is now applied before most actions, so editing text and changing carets won't cause issues.

IME is applied before the right click context menu appears, before using the edit menu for any operation, and when clicking elsewhere in the TextEdit.

Scrolling with the mousewheel is no longer blocked when the IME is open.
Updating the IME will adjust the view to the caret and update the IME window position.

### DIFF
--- a/doc/classes/TextEdit.xml
+++ b/doc/classes/TextEdit.xml
@@ -101,6 +101,12 @@
 				Adjust the viewport so the caret is visible.
 			</description>
 		</method>
+		<method name="apply_ime">
+			<return type="void" />
+			<description>
+				Applies text from the [url=https://en.wikipedia.org/wiki/Input_method]Input Method Editor[/url] (IME) to each caret and closes the IME if it is open.
+			</description>
+		</method>
 		<method name="backspace">
 			<return type="void" />
 			<param index="0" name="caret_index" type="int" default="-1" />
@@ -112,6 +118,12 @@
 			<return type="void" />
 			<description>
 				Starts a multipart edit. All edits will be treated as one action until [method end_complex_operation] is called.
+			</description>
+		</method>
+		<method name="cancel_ime">
+			<return type="void" />
+			<description>
+				Closes the [url=https://en.wikipedia.org/wiki/Input_method]Input Method Editor[/url] (IME) if it is open. Any text in the IME will be lost.
 			</description>
 		</method>
 		<method name="center_viewport_to_caret">
@@ -611,7 +623,7 @@
 		<method name="has_ime_text" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns if the user has IME text.
+				Returns [code]true[/code] if the user has text in the [url=https://en.wikipedia.org/wiki/Input_method]Input Method Editor[/url] (IME).
 			</description>
 		</method>
 		<method name="has_redo" qualifiers="const">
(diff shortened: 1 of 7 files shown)


MewPurPur:
Likely fixes #83795. I tested the big PR and found that the issue I deducted here has been fixed.

scene/gui/text_edit.cpp
```diff
@@ -3568,16 +3598,8 @@ void TextEdit::insert_text_at_caret(const String &p_text, int p_caret) {
...
-		for (int i = 0; i < carets.size(); i++) {
-			String t;
-			if (get_caret_column(i) >= 0) {
-				t = text[get_caret_line(i)].substr(0, get_caret_column(i)) + ime_text + text[get_caret_line(i)].substr(get_caret_column(i), text[get_caret_line(i)].length());
-			} else {
-				t = ime_text;
-			}
-			text.invalidate_cache(get_caret_line(i), get_caret_column(i), true, t, structured_text_parser(st_parser, st_args, t));
-		}
+	if (has_ime_text()) {
```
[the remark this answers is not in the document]
kitbdev: If I don't check for it, it would run the other branch in `_update_ime_text()` and invalidates the text cache unnecessarily.

doc/classes/TextEdit.xml
```diff
@@ -101,6 +101,12 @@
 				Adjust the viewport so the caret is visible.
 			</description>
 		</method>
+		<method name="apply_ime">
+			<return type="void" />
+			<description>
+				Applies text from the IME to each caret and closes the IME if it is open.
```
Mickeon, line 107: It's not really explained what IME stands for.
```suggestion
				Applies text from the Input Method Editor (IME) to each caret and closes the IME if it is open.
```
Or, should we have a link?
```suggestion
				Applies text from the [url=https://en.wikipedia.org/wiki/Input_method]Input Method Editor[/url] (IME) to each caret and closes the IME if it is open.
```

doc/classes/TextEdit.xml
```diff
@@ -114,6 +120,12 @@
 				Starts a multipart edit. All edits will be treated as one action until [method end_complex_operation] is called.
 			</description>
 		</method>
+		<method name="cancel_ime">
+			<return type="void" />
+			<description>
+				Closes the IME if it is open. Any text in the IME will be lost.
```
Mickeon, line 126: Ditto
```suggestion
				Closes the [url=https://en.wikipedia.org/wiki/Input_method]Input Method Editor[/url] (IME) if it is open. Any text in the IME will be lost.
```

doc/classes/TextEdit.xml
```diff
@@ -611,7 +623,7 @@
 		<method name="has_ime_text" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns if the user has IME text.
+				Returns [code]true[/code] if the user has IME text.
```
Mickeon, line 626: Ditto
```suggestion
				Returns [code]true[/code] if the user has text in the [url=https://en.wikipedia.org/wiki/Input_method]Input Method Editor[/url] (IME).
```

Mickeon (commented):
I cannot speak on whether this is a good bunch of changes or not, but a few thoughts on the docs:

doc/classes/TextEdit.xml
kitbdev: I'll use the link so it matches the DisplayServer docs.

akien-mga (approved):
Looks good to me overall. Let's get this merged to unblock the next batch of fixes :)

akien-mga:
Thanks!

PR #87482, opened 2024-01-22
Direct3D 12: Let NIR runtime data and push constants use the same register scheme as bindings

RandomShaper:
Before this PR, the fixed space-register used for the NIR runtime data and root constants were not using the same multipliers the regular bindings have applied to. This seemed not to be a problem in practice, but was not as safe for "namespacing" the inputs of the shader as applying the same scheme to everything. This PR makes register usage consisten again.

### DIFF
--- a/drivers/d3d12/rendering_device_driver_d3d12.cpp
+++ b/drivers/d3d12/rendering_device_driver_d3d12.cpp
@@ -91,10 +91,8 @@ extern "C" {
 
 static const D3D12_RANGE VOID_RANGE = {};
 
-static const uint32_t ROOT_CONSTANT_SPACE = RDD::MAX_UNIFORM_SETS + 1;
-static const uint32_t ROOT_CONSTANT_REGISTER = 0;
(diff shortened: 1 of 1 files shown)


YuriSizov:
@RandomShaper Seems like we never added the D3D12 driver's folder to [codeowners](https://github.com/godotengine/godot/blob/master/.github/CODEOWNERS#L38). Would you mind adding this change to this PR while you're at it?

RandomShaper:
> @RandomShaper Seems like we never added the D3D12 driver's folder to [codeowners](https://github.com/godotengine/godot/blob/master/.github/CODEOWNERS#L38). Would you mind adding this change to this PR while you're at it?

Submitted separately (#87574). I would've just added another commit to this PR, but I'm making thousands of D3D12 PRs today, so one more is no big deal.

clayjohn (approved):
Looks fine to me. I don't know how I could test and confirm that it works, so I will just trust you :)

akien-mga:
Thanks!

Issue #87418, opened 2024-01-20
Script UID improvements


KoBeWi:
### Tested versions

4.3

### System information

Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1060 (NVIDIA; 30.0.15.1403) - Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz (8 Threads)

### Issue description

GDScript has a basic UID support now, but it has some missing features:
- [x] Allow extending script by its UID (`extends "uid://acdc"`)
- [x] Allow using script preloaded with UID as a type (`const Type = preload("uid://acdc")`)
- [x] Add "Copy Script UID" option to Script Editor, in addition to Copy Script Path
![image](https://github.com/godotengine/godot/assets/2223172/aaee32ad-2446-4e5b-8ba7-935a722db24c)

The first two result in an error currently, the last one does not exist (though you can copy UID using FileSystem context menu).

### Steps to reproduce

Try using any of the above.

### Minimal reproduction project (MRP)

N/A

dalexeev:
> Allow using script preloaded with UID as a type (`const Type = preload("uid://acdc")`)

Preloading with only a UID is less readable and inconsistent with how other resources (it is assumed that UIDs can be safely removed/changed). I would expect an optional `uid` argument for `load`/`preload`:

```gdscript
const Type = preload("res://script.gd", "uid://acdc")
```

The same for `extends`, but the syntax probably looks worse due to the lack of parentheses:

```gdscript
extends "res://script.gd", "uid://acdc"
```

Also note that you can use relative paths with `preload` and `extends`:

```gdscript
extends "./script.gd"

const Type = preload("./other_script.gd")
```

For `load`, we can also add support for relative paths (see https://github.com/godotengine/godot/pull/66550#issuecomment-1648038763) or add "magic" constants like `__FILE__`, `__DIR__` (but we should distinguish between absolute and `res://` paths).

AThousandShips:
This would also allow us to have a warning if the plain path is out of date, possibly with an option to update it in the editor

KoBeWi:
> Preloading with only a UID is less readable and inconsistent with other resources (it is assumed that UIDs can be safely removed/changed).

No, the whole purpose of UID is that it's *not* supposed to change, unlike paths. The engine ensures that. Other resources already support load and preload, it only doesn't work for scripts.

The readability can be always improved by adding a tooltip when hovering UID or even displaying it in a special way.

[...]

PR #87488, opened 2024-01-22
Allow using script preloaded with UID as a type

brno32:
Addresses second item from https://github.com/godotengine/godot/issues/87418

Again, I'd like to add a test but since the tests currently only load one .gd file at a time, the test setup would need some extensive work


dalexeev:
Preloading with only a UID is less readable and inconsistent with other resources. I would expect an optional `uid` argument for `load`/`preload` (`preload("res://script.gd", "uid://uid")`). See https://github.com/godotengine/godot/issues/87418#issuecomment-1907516677 for details.

Issue #81730, opened 2023-09-16
Clip Children with semitransparent child has unwanted drawings 


tokengamedev:
### Godot version

4.2.dev.4

### System information

Windows 11

### Issue description

When a `PanelContainer`/`Panel` with children have clip children set to clip only and when children have alpha less than 1 (Either in modulate or use alpha in color of `ColorRect`.
Unwanted draw happens for controls if there are adjacent controls like Panel in a `HBoxContainer`.
![image](https://github.com/godotengine/godot/assets/84369217/2b6295d1-f885-478e-9776-bb3efa36541e)

Note: `PanelContainer` is used to have corner radius for the overall look and feel using StyleBox

### Steps to reproduce

Create a project using the Node Hierarchy as follows:
```
+ Control (AnchorPreset = FullRect)
|
+-- HBoxContainer (AnchorPreset = FullRect)
|    |
|    +--Panel (Horizontal Expand = True) 
|    |
|    +-- PanelContainer (Clip Children = "Clip Only")
|    |   |
|    |   +--ColorRect (color.alpha < 1.0)
```

when this project is run you can see the left panel is drawn on top of PanelContainer

### Minimal reproduction project

[ClipChildrenTest.zip](https://github.com/godotengine/godot/files/12640281/ClipChildrenTest.zip)

Removed the .godot folder, if facing issue opening the project I will add the .godot folder
P.S: Why this .godot folder in 4.2?

tokengamedev:
Did some more testing  in 4.2.2 rc1 and 4.2.1 stable and 
This may be an issue with clipping and panel and panel container types.

![image](https://github.com/godotengine/godot/assets/84369217/1d969e09-4394-47c8-a005-3f62877feb64)
_Check the image above. the last one is MarginContainer with clipping enabled but no issues._

Even If you reorder the panels, the bands are only displayed in 2nd panel and onwards

tokengamedev:
This is working as designed so closing it

PR #87489, opened 2024-01-22
Disable scissor test after rendering batches in compatibility renderer

### LINKED ISSUE #87396: Screen texture weird interaction with ScrollContainer
### Tested versions

- Reproducible in 4.2.stable, 4.2.1.stable

### System information

Windows 10 - Godot v4.2.1.stable - Vulkan (GL Compatibility) - dedicated

### Issue description

Screen texture becomes empty (black) outside of ScrollContainer.
屏幕截图(559)

### Steps to reproduce

In MRP, run the project and the display should be same as issue screenshot.

Other possibly important detail, if ScrollContainer is moved to the same level as ColorRect filter, the problem disappears. If just using `COLOR = texture(SCREEN_TEXTURE, SCREEN_UV);` in the filter shader, problem disappears. 

If a new Label node is added as sibling of ScrollContainer but below it in the node structure, problem disappears. But if it isn't visible or is a Control node for example, problem persists.

[screen_texture_ui.zip](https://github.com/godotengine/godot/files/13997040/screen_texture_ui.zip)

clayjohn:
Fixes: https://github.com/godotengine/godot/issues/87396

The scissor test was accidentally left enabled. Already, it is disabled before rendering the batches, but we don't want the scissor state to leak into other areas. So its best to disable it after as well.

### DIFF
--- a/drivers/gles3/rasterizer_canvas_gles3.cpp
+++ b/drivers/gles3/rasterizer_canvas_gles3.cpp
@@ -775,6 +775,7 @@ void RasterizerCanvasGLES3::_render_items(RID p_to_render_target, int p_item_cou
 		_render_batch(p_lights, i, r_render_info);
 	}
 
+	glDisable(GL_SCISSOR_TEST);
 	state.current_batch_index = 0;
 	state.canvas_instance_batches.clear();
 	state.last_item_index += index;


lawnjelly (approved):

tokengamedev:
Not sure, but most probably this will fix https://github.com/godotengine/godot/issues/81730 too.

Assuming based on the workaround provided by the issue https://github.com/godotengine/godot/issues/87396, which works on the issue.

clayjohn:
> Not sure, but most probably this will fix #81730 too.
> 
> Assuming based on the workaround provided by the issue #87396, which works on the issue.

Unfortunately, I don't think that this will fix https://github.com/godotengine/godot/issues/81730. Your MRP in https://github.com/godotengine/godot/issues/81730 is using the mobile renderer and this change only touches the compatibility renderer

YuriSizov:
Thanks!

YuriSizov:
Cherry-picked for 4.2.2.

YuriSizov:
Cherry-picked for 4.1.4.

PR #87490, opened 2024-01-22
Add 'Copy Script UID' option to Script Editor

brno32:
Addresses the third item from https://github.com/godotengine/godot/issues/87418


KoBeWi:
Script UIDs are back, but built-in scripts don't have them. The option should be disabled in that case, otherwise you end up with `uid://<invalid>`.

PR #87495, opened 2024-01-23
Clean up outdated `USE_LIGHT_SHADER_CODE` usage

### LINKED ISSUE #85370: ALPHA and discard in light() have no effect in Compatibility or Forward+ renderers
### Godot version

v4.2.rc2.official [1ba920fad]

### System information

Godot v4.2.rc2 - Windows 10.0.19045 - GLES3 (Compatibility) - NVIDIA GeForce GTX 1660 Ti (NVIDIA; 31.0.15.2802) - Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (12 Threads)

### Issue description

Between 4.1.3 and 4.2, something changed so that writing to `ALPHA` in the `light()` function of a Spatial shader has no effect, if using the OpenGL Compatibility renderer. Writing to `ALPHA` still works in the `fragment()` function, or when using a Vulkan renderer.

EDIT: I failed to check for this before, but `discard` also similarly does nothing.

Here is a comparison of my project in 4.1.3 and 4.2 rc2 (I've already fixed it now by moving things to `fragment()`):

Image1

### Steps to reproduce

1. Write a Spatial shader that sets `ALPHA` to a value lower than `1.0` in the `light()` function
2. Nothing happens to `ALPHA`

[min-repro-11-25-23.zip](https://github.com/godotengine/godot/files/13465818/min-repro-11-25-23.zip)

jsjtxietian:
Fixes https://github.com/godotengine/godot/issues/85370

Now all three renderers work the same, but still when there is no light in the scene, light shader won't be called.

### DIFF
--- a/drivers/gles3/storage/material_storage.cpp
+++ b/drivers/gles3/storage/material_storage.cpp
@@ -1178,7 +1178,6 @@ MaterialStorage::MaterialStorage() {
 		actions.usage_defines["SCREEN_PIXEL_SIZE"] = "@SCREEN_UV";
 		actions.usage_defines["NORMAL"] = "#define NORMAL_USED\n";
 		actions.usage_defines["NORMAL_MAP"] = "#define NORMAL_MAP_USED\n";
-		actions.usage_defines["LIGHT"] = "#define LIGHT_SHADER_CODE_USED\n";
 		actions.usage_defines["SPECULAR_SHININESS"] = "#define SPECULAR_SHININESS_USED\n";
 
 		actions.render_mode_defines["skip_vertex_transform"] = "#define SKIP_TRANSFORM_USED\n";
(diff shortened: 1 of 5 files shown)


drivers/gles3/storage/material_storage.cpp
```diff
@@ -1178,7 +1178,7 @@ MaterialStorage::MaterialStorage() {
 		actions.usage_defines["SCREEN_PIXEL_SIZE"] = "@SCREEN_UV";
 		actions.usage_defines["NORMAL"] = "#define NORMAL_USED\n";
 		actions.usage_defines["NORMAL_MAP"] = "#define NORMAL_MAP_USED\n";
-		actions.usage_defines["LIGHT"] = "#define LIGHT_SHADER_CODE_USED\n";
+		actions.usage_defines["LIGHT"] = "#define LIGHT_CODE_USED\n";
```
clayjohn, line 1181: This should be removed. `LIGHT_CODE_USED` is already added by the shader compiler https://github.com/godotengine/godot/blob/0bcc0e92b3f0ac57d4c4650722f347593a258572/drivers/gles3/shader_gles3.cpp#L179

jsjtxietian: Thanks！

clayjohn (approved):
Looks great!

YuriSizov:
Thanks!

PR #87498, opened 2024-01-23
[macOS] Fix missing modifier + space key down events.

### LINKED ISSUE #87491: (macOS) Space bar being pressed is not detected when the Control or Command keys are held down
### Tested versions

Reproducible in 4.3.dev2 [352434668], v4.2.1.stable [b09f793f5], and v4.0.alpha15 [432b25d36]

Not reproducible in v3.6.beta3 [21ab700f2]

### System information

Godot v4.2.1.stable - macOS 14.2.1 - Vulkan (Forward+) - dedicated AMD Radeon Pro 570X - Intel(R) Core(TM) i5-8500 CPU @ 3.00GHz (6 Threads)

### Issue description

Godot does not detect the space bar being pressed at all if Control or Command are held down.

The _input() method of nodes is not called, and neither `Input.is_key_pressed()` nor `Input.is_action_pressed()` will reflect the true state of the space bar.
The release event for space bar always works as expected, whether control is held down or not.

I briefly tested it on windows and this bug does not appear to be present there.### 

### Steps to reproduce

Print when the space bar is pressed down:
```gdscript
extends Node2D

func _input(event: InputEvent):
	if event is InputEventKey:
		if event.keycode == KEY_SPACE and event.is_pressed():
			print("spacebar")
```

On windows it will still print if control is held, but on macOS it will not.

N/A

bruvzg:
Fixes https://github.com/godotengine/godot/issues/87491

Note: by default `Command` + `Space` is a spotlight shortcut, unless it's disabled in the system setting it's normal to not receive any events for it).

### DIFF
--- a/platform/macos/godot_content_view.mm
+++ b/platform/macos/godot_content_view.mm
@@ -576,21 +576,23 @@
 			String u32text;
 			u32text.parse_utf16(text.ptr(), text.length());
 
+			DisplayServerMacOS::KeyEvent ke;
+			ke.window_id = window_id;
+			ke.macos_state = [event modifierFlags];
+			ke.pressed = true;
+			ke.echo = [event isARepeat];
+			ke.keycode = KeyMappingMacOS::remap_key([event keyCode], [event modifierFlags], false);
+			ke.physical_keycode = KeyMappingMacOS::translate_key([event keyCode]);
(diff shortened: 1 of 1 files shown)


YuriSizov (approved):

YuriSizov:
Thanks!

PR #87504, opened 2024-01-23
Auto-bake edited NavigationPolygons in the Editor on a timer

smix8:
Auto-bakes edited NavigationPolygons in the Editor on a timer.

The pre-4.2 NavigationRegion2D did automatically rebake the entire NavigationPolygon everytime it was changed even a little.

This caused severe performance problems while editing more complex polygons.

Baking a navigation mesh is not exactly a quick process. It was very noticeable while developing the 2D navigation mesh baking with the Editor freezing all the time while dragging outline points. That was the reason why with 4.2. this auto-bake was removed from the Editor. It was expected that now with the 2D navigation mesh very complex polygons will be far more common then before and users will complain about this performance problem.

That is also what happened (the complex polygon part) but now some users started to miss the convenience of the old feature and want it back. If they only have simple projects it was convenient to not hit the bake button all the time to see how things turn out when baked. It is also regularly confusing new users that do not understand why there is no polygon, not having discovered the bake button yet.

So this pr adds back the auto-bake on a timer that refreshes on every change before firing the bake. By default 1.5 seconds which gives enough time to drag around multiple polygon outline points without triggering the bake. The timer can be adjusted in the EditorSettings. By turning it down to 0.0 seconds the old instant-bake is back. By going negativ <0 with the timer it gets disabled completely again. Slower users that need more time to arrange things can increase the timer near infinite, so the feature stays accesible.

I think this solves the issue for all the 3 major user groups.

### DIFF
--- a/editor/plugins/navigation_polygon_editor_plugin.cpp
+++ b/editor/plugins/navigation_polygon_editor_plugin.cpp
@@ -31,6 +31,7 @@
 #include "navigation_polygon_editor_plugin.h"
 
 #include "editor/editor_node.h"
+#include "editor/editor_settings.h"
 #include "editor/editor_undo_redo_manager.h"
 #include "scene/2d/navigation_region_2d.h"
 #include "scene/gui/dialogs.h"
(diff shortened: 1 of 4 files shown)


fire (approved):
I don't usually review the navigation pull requests, but this appears to be a timer for delayed baking of navigation. It seems fine.

editor/plugins/navigation_polygon_editor_plugin.h
```diff
@@ -56,6 +56,10 @@ class NavigationPolygonEditor : public AbstractPolygon2DEditor {
 	Button *button_reset = nullptr;
 	Label *bake_info = nullptr;
 
+	Timer *rebake_timer;
```
AThousandShips, line 59:
```suggestion
	Timer *rebake_timer = nullptr;
```

editor/plugins/navigation_polygon_editor_plugin.cpp
```diff
@@ -193,6 +240,19 @@ void NavigationPolygonEditor::_update_polygon_editing_state() {
 	}
 }
 
+void NavigationPolygonEditor::_rebake_timer_timeout() {
+	if (!node) {
+		return;
+	}
+	Ref<NavigationPolygon> navigation_polygon = node->get_navigation_polygon();
+	if (!navigation_polygon.is_valid()) {
```
AThousandShips, line 248:
```suggestion
	if (navigation_polygon.is_null()) {
```

AThousandShips (approved):

akien-mga:
Thanks!

PR #87512, opened 2024-01-23
Fix validation errors by improving stage and slice tracking behavior of RenderingDeviceGraph.

DarioSamo:
Makes some improvements to the tracking behavior of RenderingDeviceGraph to fix some errors that were spotted in other scenes.

- Stage bits could be missed between operations to be inserted correctly due to the pipeline stage changing between subsequent operations that don't imply transition barriers. The stage bits have been split off into their own bitfield and are added to the destination as well when detecting adjacency on the graph.
- The graph could potentially reorder operations that normalize a texture's slices to be performed after the slices were already transitioned due to not tracking the entire area of the resource as a write operation when this happens. Whenever normalization happens now, the area of whatever is normalized is merged and considered for searching previous commands for intersections.
- The graph could potentially cause a validation error by missing the dependency between a command that only reads from a resource (without causing a transition) and a command that writes to a resource slice. This has been fixed by introducing separate lists for reading resources completely and only reading slices. Both lists can be cleared separately depending on whether the resource is modifying only a slice or the entire resource: if it's the entire resource, it'll always delete both lists. If it's only a slice, it won't delete them at all from the full read list, but it'll delete them if it completely encloses the area of a previous read operation. It'll also only add a dependency if there's an intersection between both areas.

This has been verified to fix various issues that could crop up when using certain effects. The https://github.com/godotengine/godot/issues/75440 showed a few of these errors as well as using a heavy amount of omni lights and various effects in other projects. It might not be possible to identify visual differences depending on the nature of the errors, so checking with validation layers is recommended.

The PR also adds a few optional print commands behind a macro to help aid future debugging, as it can be very troublesome to identify these issues without the information from the driver backend that matches the IDs reported by the Vulkan layers.

### DIFF
--- a/servers/rendering/rendering_device_graph.h
+++ b/servers/rendering/rendering_device_graph.h
@@ -112,8 +112,9 @@ public:
 		int32_t buffer_barrier_count = 0;
 #endif
 		int32_t label_index = -1;
-		BitField<RDD::PipelineStageBits> src_stages;
-		BitField<RDD::PipelineStageBits> dst_stages;
+		BitField<RDD::PipelineStageBits> previous_stages;
+		BitField<RDD::PipelineStageBits> next_stages;
+		BitField<RDD::PipelineStageBits> self_stages;
 	};
 
 	struct RecordedBufferCopy {
@@ -150,7 +151,8 @@ public:
 	struct ResourceTracker {
 		uint32_t reference_count = 0;
 		int64_t command_frame = -1;
-		int32_t read_command_list_index = -1;
+		int32_t read_full_command_list_index = -1;
+		int32_t read_slice_command_list_index = -1;
 		int32_t write_command_or_list_index = -1;
 		int32_t draw_list_index = -1;
 		int32_t compute_list_index = -1;
@@ -171,7 +173,8 @@ public:
 			if (new_command_frame != command_frame) {
 				usage_access.clear();
 				command_frame = new_command_frame;
-				read_command_list_index = -1;
+				read_full_command_list_index = -1;
+				read_slice_command_list_index = -1;
 				write_command_or_list_index = -1;
 				draw_list_index = -1;
 				compute_list_index = -1;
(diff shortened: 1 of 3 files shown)


servers/rendering/rendering_device_graph.h
```diff
@@ -590,6 +594,7 @@ class RenderingDeviceGraph {
 	static RDD::BarrierAccessBits _usage_to_access_bits(ResourceUsage p_usage);
 	int32_t _add_to_command_list(int32_t p_command_index, int32_t p_list_index);
 	void _add_adjacent_command(int32_t p_previous_command_index, int32_t p_command_index, RecordedCommand *r_command);
+	int32_t _add_to_slice_read_list(int32_t p_command_index, Rect2i suberesources, int32_t p_list_index);
```
AThousandShips, line 597:
```suggestion
	int32_t _add_to_slice_read_list(int32_t p_command_index, Rect2i p_subresources, int32_t p_list_index);
```

servers/rendering/rendering_device_graph.h
```diff
@@ -590,6 +594,7 @@ class RenderingDeviceGraph {
...
 	int32_t _add_to_write_list(int32_t p_command_index, Rect2i suberesources, int32_t p_list_index);
```
AThousandShips, line 598:
```suggestion
	int32_t _add_to_write_list(int32_t p_command_index, Rect2i p_subresources, int32_t p_list_index);
```
Let's also fix this typo

DarioSamo:
Added an extra fix to this where resource trackers would not get reset correctly if their data is outdated if they were first encountered while traversing the dirty list of a texture. This error could be spotted when loading the TPS demo.

DarioSamo:
Added a fix for another edge case that could show up: access bits should be assumed to be the same as the parent when a new slice is being tracked. Resulted in another small validation error that only happened once during setup in the scene being tested in mobile.

clayjohn (approved):
Makes sense to me. 

I tested this locally on a couple different test scenes in both renders as well as two test scenes on an Android device (both using the mobile renderer).

akien-mga:
Thanks!

PR #65532, opened 2022-09-08
Various fixes to C# warnings

raulsntos:
- Suppress CS0108 caused by shadowing members in derived classes.
	- Fixes https://github.com/godotengine/godot/issues/15763
- Suppress CA1716 caused by parameter names matching C# keywords.
	- Follow up to https://github.com/godotengine/godot/pull/87518.

<details>
<summary><h3>Extracted to different PRs</h3></summary>

- Suppress REFL045 in a false positive scenario.
	- Extracted to https://github.com/godotengine/godot/pull/88570.
- Suppress CA1710 caused by Array not having the Collection suffix.
	- Extracted to https://github.com/godotengine/godot/pull/87518.
- Suppress CA1716 caused by parameter names matching C# keywords.
	- Extracted to https://github.com/godotengine/godot/pull/87518.
- Use SymbolEqualityComparer in ScriptPathAttributeGenerator to fix RS1024.
	- Extracted to https://github.com/godotengine/godot/pull/87342.
- Suppress dependencies of SourceGenerators package to fix NU5128.
	- Extracted to https://github.com/godotengine/godot/pull/79501.
- Suppress nullability warnings in source generators because we already have a Debug.Assert (but it's missing the nullability annotations in netstandard2.0).
	- Extracted to https://github.com/godotengine/godot/pull/87518.
- Implement disposable pattern in `GodotSynchronizationContext` (and make it sealed).
	- Extracted to https://github.com/godotengine/godot/pull/72053.
- Fix reference to `NotificationSceneInstantiated` in the documentation.
	- Follow up to https://github.com/godotengine/godot/pull/64410.
	- Extracted to https://github.com/godotengine/godot/pull/72462.
- Fix RS2008: Enable analyzer release tracking.
	- Implemented by https://github.com/godotengine/godot/pull/80343.

</details>

### TODO
- [ ] Fix CA1001: Type owns disposable field but is not disposable.
	- [ ] `Array` and `Dictionary`: This was already discussed a bit in https://github.com/godotengine/godot/pull/64089#discussion_r940589373 but I think it's not as straightforward as the rest of the fixes and it should probably be handled in a separate PR.
	- [ ] `Variant`: It already implements `IDisposable` so I don't know why it complains.
		- Seems like it's a bug in the analyzer: https://github.com/dotnet/roslyn-analyzers/issues/6151.

PR #87518, opened 2024-01-23
Clean a number of C# warnings

paulloz:
Trying to reduce the number of warnings to a manageable amount to see things more clearly. Not perfect, but a step forward.

- `[Obsolete]` tag generated in the middle of documentation comments
- Potential `null` values in generators
- Obsolete call to `GetEditorInterface()`
- We don't want `Godot.Collections.Array` to end with `Collection`
- A few culture specifications and use of `AsSpan` instead of `SubString` in `StringExtensions`
- Disable CA1716 in GodotSharp

### DIFF
--- a/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/Helper.cs
+++ b/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/Helper.cs
@@ -1,0 +1,15 @@
+using System;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Godot.SourceGenerators
+{
+    public static class Helper
+    {
+        [Conditional("DEBUG")]
+        public static void ThrowIfNull([NotNull] object? value)
+        {
+            _ = value ?? throw new ArgumentNullException();
+        }
+    }
+}
--- a/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/MustBeVariantAnalyzer.cs
+++ b/modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/MustBeVariantAnalyzer.cs
@@ -1,5 +1,4 @@
 using System.Collections.Immutable;
-using System.Diagnostics;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -34,7 +33,7 @@ namespace Godot.SourceGenerators
 
             // Method invocation or variable declaration that contained the type arguments
             var parentSyntax = context.Node.Parent;
-            Debug.Assert(parentSyntax != null);
+            Helper.ThrowIfNull(parentSyntax);
 
             var sm = context.SemanticModel;
 
@@ -49,9 +48,10 @@ namespace Godot.SourceGenerators
                     continue;
 
                 var typeSymbol = sm.GetSymbolInfo(typeSyntax).Symbol as ITypeSymbol;
-                Debug.Assert(typeSymbol != null);
+                Helper.ThrowIfNull(typeSymbol);
 
                 var parentSymbol = sm.GetSymbolInfo(parentSyntax).Symbol;
+                Helper.ThrowIfNull(parentSymbol);
 
                 if (!ShouldCheckTypeArgument(context, parentSyntax, parentSymbol, typeSyntax, typeSymbol, i))
                 {
@@ -69,7 +69,7 @@ namespace Godot.SourceGenerators
 
                 var marshalType = MarshalUtils.ConvertManagedTypeToMarshalType(typeSymbol, typeCache);
 
-                if (marshalType == null)
+                if (marshalType is null)
                 {
                     Common.ReportGenericTypeArgumentMustBeVariant(context, typeSyntax, typeSymbol);
                     continue;
--- a/modules/mono/glue/GodotSharp/GodotSharp/Core/Array.cs
+++ b/modules/mono/glue/GodotSharp/GodotSharp/Core/Array.cs
@@ -16,7 +16,9 @@ namespace Godot.Collections
     /// interfacing with the engine. Otherwise prefer .NET collections
     /// such as <see cref="System.Array"/> or <see cref="List{T}"/>.
     /// </summary>
+#pragma warning disable CA1710 // Identifiers should have correct suffix
     public sealed class Array :
+#pragma warning restore CA1710
         IList<Variant>,
         IReadOnlyList<Variant>,
         ICollection,
--- a/modules/mono/glue/GodotSharp/GodotSharp/Core/StringExtensions.cs
+++ b/modules/mono/glue/GodotSharp/GodotSharp/Core/StringExtensions.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
-using System.IO;
 using System.Security;
 using System.Security.Cryptography;
 using System.Text;
@@ -220,7 +219,7 @@ namespace Godot
                 {
                     if (hasText)
                     {
-                        sb.Append(instance.Substring(indentStop, i - indentStop));
+                        sb.Append(instance.AsSpan(indentStop, i - indentStop));
                     }
                     sb.Append('\n');
                     hasText = false;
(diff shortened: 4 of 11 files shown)


modules/mono/glue/GodotSharp/GodotSharp/Core/Array.cs
```diff
@@ -16,7 +16,9 @@ namespace Godot.Collections
     /// interfacing with the engine. Otherwise prefer .NET collections
     /// such as <see cref="System.Array"/> or <see cref="List{T}"/>.
     /// </summary>
+#pragma warning disable CA1710
     public sealed class Array :
+#pragma warning restore CA1710
```
raulsntos, line 21: This has been previously discussed in https://github.com/godotengine/godot/pull/65532#discussion_r967427306 where neikeq suggested to add it to `.editorconfig` directly.

I personally don't have a strong preference either way, but if we're going to keep the pragmas I like to always include the message in a comment so it's clear what the diagnostic is about without having to lookup the Id.

```suggestion
#pragma warning disable CA1710 // Identifiers should have correct suffix
    public sealed class Array :
#pragma warning restore CA1710
```

modules/mono/glue/GodotSharp/GodotSharp/Core/StringExtensions.cs
```diff
@@ -323,7 +322,7 @@ public static string Capitalize(this string instance)
                 string slice = aux.GetSliceCharacter(' ', i);
                 if (slice.Length > 0)
                 {
-                    slice = char.ToUpper(slice[0]) + slice.Substring(1);
+                    slice = char.ToUpper(slice[0], CultureInfo.InvariantCulture) + slice.Substring(1);
```
raulsntos, line 325:
```suggestion
                    slice = char.ToUpperInvariant(slice[0]) + slice.Substring(1);
```

modules/mono/glue/GodotSharp/GodotSharp/Core/StringExtensions.cs
```diff
@@ -408,13 +407,13 @@ private static string CamelcaseToUnderscore(this string instance, bool lowerCase
...
                 {
-                    newString += instance.Substring(startIndex, i - startIndex) + "_";
+                    newString += string.Concat(instance.AsSpan(startIndex, i - startIndex), "_");
                     startIndex = i;
                 }
             }
 
             newString += instance.Substring(startIndex, instance.Length - startIndex);
-            return lowerCase ? newString.ToLower() : newString;
+            return lowerCase ? newString.ToLower(CultureInfo.InvariantCulture) : newString;
```
raulsntos, line 416:
```suggestion
            return lowerCase ? newString.ToLowerInvariant() : newString;
```

modules/mono/glue/GodotSharp/GodotSharp/Core/StringExtensions.cs
```diff
@@ -479,9 +478,9 @@ public static int CompareTo(this string instance, string to, bool caseSensitive
                         return -1; // If this is empty, and the other one is not, then we're less... I think?
                     if (to[toIndex] == 0)
                         return 1; // Otherwise the other one is smaller..
-                    if (char.ToUpper(instance[instanceIndex]) < char.ToUpper(to[toIndex])) // More than
+                    if (char.ToUpper(instance[instanceIndex], CultureInfo.InvariantCulture) < char.ToUpper(to[toIndex], CultureInfo.InvariantCulture)) // More than
```
raulsntos, line 481:
```suggestion
                    if (char.ToUpperInvariant(instance[instanceIndex]) < char.ToUpperInvariant(to[toIndex])) // More than
```

modules/mono/glue/GodotSharp/GodotSharp/Core/StringExtensions.cs
```diff
@@ -479,9 +478,9 @@ public static int CompareTo(this string instance, string to, bool caseSensitive
...
                         return -1;
-                    if (char.ToUpper(instance[instanceIndex]) > char.ToUpper(to[toIndex])) // Less than
+                    if (char.ToUpper(instance[instanceIndex], CultureInfo.InvariantCulture) > char.ToUpper(to[toIndex], CultureInfo.InvariantCulture)) // Less than
```
raulsntos, line 483:
```suggestion
                    if (char.ToUpperInvariant(instance[instanceIndex]) > char.ToUpperInvariant(to[toIndex])) // Less than
```

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/MustBeVariantAnalyzer.cs
```diff
@@ -51,9 +51,10 @@ private void AnalyzeNode(SyntaxNodeAnalysisContext context)
                 var typeSymbol = sm.GetSymbolInfo(typeSyntax).Symbol as ITypeSymbol;
                 Debug.Assert(typeSymbol != null);
 
-                var parentSymbol = sm.GetSymbolInfo(parentSyntax).Symbol;
+                var parentSymbol = sm.GetSymbolInfo(parentSyntax!).Symbol;
+                Debug.Assert(parentSymbol != null);
 
-                if (!ShouldCheckTypeArgument(context, parentSyntax, parentSymbol, typeSyntax, typeSymbol, i))
+                if (!ShouldCheckTypeArgument(context, parentSyntax!, parentSymbol!, typeSyntax, typeSymbol!, i))
```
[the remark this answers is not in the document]
paulloz: Yes, the fact `Debug.Assert` doesn't trigger nullability in standard is a bummer. Another solution would be to make our own wrapper to `Debug.Assert` and use `[NotNullWhen]` 🤔

modules/mono/glue/GodotSharp/GodotSharp/Core/Array.cs
paulloz: I don't have a strong opinion either, I used `#pragma` here to be the less disruptive possible. But if most people agree, I'm also down to suppress it entirely from the `.editorconfig`.

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/Helper.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Godot.SourceGenerators
+{
+    public static class Helper
+    {
+        [Conditional("DEBUG")]
+        public static void ThrowWhenNull([NotNull] object? value)
```
raulsntos, line 10: I would name it `ThrowIfNull` to match .NET throw helpers (e.g.: [`ArgumentNullException.ThrowIfNull`](https://github.com/dotnet/runtime/blob/205ef031e0fe5152dede0bd9f99d0f6f9e7f1e45/src/libraries/System.Private.CoreLib/src/System/ArgumentNullException.cs#L62))

modules/mono/editor/Godot.NET.Sdk/Godot.SourceGenerators/MustBeVariantAnalyzer.cs
```diff
@@ -49,9 +48,10 @@ private void AnalyzeNode(SyntaxNodeAnalysisContext context)
                     continue;
 
                 var typeSymbol = sm.GetSymbolInfo(typeSyntax).Symbol as ITypeSymbol;
-                Debug.Assert(typeSymbol != null);
+                Helper.ThrowWhenNull(typeSymbol);
 
-                var parentSymbol = sm.GetSymbolInfo(parentSyntax).Symbol;
+                var parentSymbol = sm.GetSymbolInfo(parentSyntax!).Symbol;
```
raulsntos, line 53: Shouldn't need this null forgiveness operator since we're checking the nullability of `parentSyntax` earlier.

```suggestion
                var parentSymbol = sm.GetSymbolInfo(parentSyntax).Symbol;
```

modules/mono/glue/GodotSharp/GodotSharp/Core/Array.cs
raulsntos: > I used `#pragma` here to be the less disruptive possible.

That was also what I was thinking in https://github.com/godotengine/godot/pull/65532. Let's keep the pragma and add the comment then, we can always change our minds in the future.

raulsntos (approved):
Looks good to me, thanks! Don't forget to squash the commits.

akien-mga:
Thanks!

PR #87522, opened 2024-01-24
Add OpenXR XR_EXTX_overlay extension support

CheesecakeCG:
I took a shot at adding support for the [XR_EXTX_overlay](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_EXTX_overlay) extension based on https://github.com/GodotVR/godot_openxr/pull/101. This should allow Godot to be used to create overlays for Monado (and maybe other OpenXR runtimes?).

~~I was not able to get it working, Monado still treats Godot games as normal apps when using this.~~ This also still needs:

- [ ] The ability to detect when the application becomes invisible
- [ ] To be exposed to GDScript
- [X] To be exposed the project settings. 

This is my first time adding a feature to Godot or really writing C++, so I've probably missed something else too.


modules/openxr/extensions/openxr_overlay_extension.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+	return request_extensions;
+}
+
+void *OpenXROverlayExtension::set_session_create_and_get_next_pointer(void *p_next_pointer) {
+	XrSessionCreateInfoOverlayEXTX *sessionCreateInfoOverlay = new XrSessionCreateInfoOverlayEXTX();
+	sessionCreateInfoOverlay->next = p_next_pointer;
+	sessionCreateInfoOverlay->sessionLayersPlacement = session_layers_placement;
+	return sessionCreateInfoOverlay;
+}
```
AThousandShips, line 47:
```suggestion
}

```
Missing trailing linefeed

modules/openxr/extensions/openxr_overlay_extension.h
```diff
@@ -0,0 +1,49 @@
...
+
+	virtual HashMap<String, bool *> get_requested_extensions() override;
+
+	virtual void *set_session_create_and_get_next_pointer(void *p_next_pointer) override;
+
+private:
+	// SessionLayersPlacement session_layers_placement = IN_FRONT;
+	uint32_t session_layers_placement = UINT32_MAX;
+};
+#endif // OPENXR_OVERLAY_EXTENSION_H
```
AThousandShips, line 49:
```suggestion

#endif // OPENXR_OVERLAY_EXTENSION_H

```

modules/openxr/extensions/openxr_overlay_extension.h
```diff
@@ -0,0 +1,49 @@
...
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+#ifndef OPENXR_OVERLAY_EXTENSION_H
+#define OPENXR_OVERLAY_EXTENSION_H
+#include "openxr_extension_wrapper.h"
```
AThousandShips, line 32:
```suggestion

#include "openxr_extension_wrapper.h"
```

modules/openxr/extensions/openxr_overlay_extension.h
```diff
@@ -0,0 +1,49 @@
...
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+#ifndef OPENXR_OVERLAY_EXTENSION_H
```
AThousandShips, line 30:
```suggestion

#ifndef OPENXR_OVERLAY_EXTENSION_H
```

modules/openxr/extensions/openxr_overlay_extension.cpp
```diff
@@ -0,0 +1,47 @@
...
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "openxr_overlay_extension.h"
+// TODO: Make this a singleton
```
AThousandShips, line 32:
```suggestion

// TODO: Make this a singleton
```

modules/openxr/extensions/openxr_overlay_extension.cpp
```diff
@@ -0,0 +1,47 @@
...
+// TODO: Add project options for this
+HashMap<String, bool *> OpenXROverlayExtension::get_requested_extensions() {
+	HashMap<String, bool *> request_extensions;
+
+	request_extensions[XR_EXTX_OVERLAY_EXTENSION_NAME] = new bool(true);
```
AThousandShips, line 37: This should be a variable in the class, as the others do, or it'll leak memory, if always available you should use:
```suggestion
	request_extensions[XR_EXTX_OVERLAY_EXTENSION_NAME] = nullptr;
```
Otherwise
```suggestion
	request_extensions[XR_EXTX_OVERLAY_EXTENSION_NAME] = &available;
```
And add it to the header

modules/openxr/extensions/openxr_overlay_extension.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+	return request_extensions;
+}
+
+void *OpenXROverlayExtension::set_session_create_and_get_next_pointer(void *p_next_pointer) {
+	XrSessionCreateInfoOverlayEXTX *sessionCreateInfoOverlay = new XrSessionCreateInfoOverlayEXTX();
```
AThousandShips, line 43:
```suggestion
	XrSessionCreateInfoOverlayEXTX *session_create_info_overlay = new XrSessionCreateInfoOverlayEXTX();
```
`snake_case`

CheesecakeCG:
For the layer placement setting, I'm still not sure if it's better to have two separate options here with presets or just show the one int? I think most of the time it makes the most sense to just always set it to the maximum or minimum (Listed as Front & Back), but there are definitely cases where setting it to a specific value does make more sense (starting multiple related overlays).
image

SpookySkeletons:
Setting to zero places it on the same layer as the base in the compositor, whatever comes active lastly shows on top if two values are the same.

LOVR engine is already handling it as a single int, anything above 0 counts as overlay so that behavior is fine.

Really good time to specify because of how the layering works, it's more likely than not you want the application end user to be able to configure this int as it can very easily not be set to a preferable setting by the developer.

BastiaanOlij:
Cool! thanks for working on this @CheesecakeCG , I need to find some time to properly dive into it but looks like a good start.

beniwtv:
I also had planned to implement this (but life got in the way), gonna have a look at this.

beniwtv:
So, I now had some time to look at this, for now I also haven't been able to show the actual overlay, will try some more though.

CheesecakeCG:
I've pushed some of the changes I made a few weeks ago, and rebased it so I could check if it still works with OpenXR composition layers enabled at the same time (it does). I'm still figuring out how to restart the OpenXR interface with a different layer placement, but I haven't had much time to work on this.

CheesecakeCG:
> So, I now had some time to look at this, for now I also haven't been able to show the actual overlay, will try some more though.

Make sure you're using an OpenXR runtime that implements the extension, so far I think Monado is the only one that does natively, but there's an API Layer for it:
https://github.com/LunarG/OpenXR-OverlayLayer

In my test project, I set the environment set to have a transparent sky and  I have script that runs:
```gdscript
get_viewport().transparent_bg = true
get_viewport().use_xr = true
```

PR #75434, opened 2023-03-28
C#: Fix memory leak when passing RefCounted objects to C#

RedworkDE:
- Fixes https://github.com/godotengine/godot/issues/71081

If objects are passed to C# a managed wrapper is created for them.
Previously this wrapper was tracked from native by a strong GC handle, keeping it alive but the managed instance also increments the native refcount keeping the native object alive, allowing neither to be freed unless the managed object is manually disposed.

This changes the link to a weak reference so that the managed object can be garbage collected, when no longer references. Unlike scripts, the object the wrapper can be cheaply recreated, so there is no need (or possibility) to do the whole dance of holding a strong reference only when there also is a managed reference to the object.

Also sometimes there was an additional strong handle to the managed wrapper that just got discarded without being released that kept the wrappers alive.

Initially the strong handle is necessary, as during creating there is a moment where that handle is the only reference to the newly created wrapper.

This fix is kinda hacky, but the whole binding handling is in need of a bigger cleanup anyways, so I left it at a minimal fix.

MRP for the issue is actually just calling any method that returns a new RefCounted object a bunch and not disposing the result, eg:
```cs
for (int i = 0; i < 10_000; i++)
	DirAccess.Open("res://");
```

PR #87526, opened 2024-01-24
C#: Fix incorrect GC handle for non-instantiable types.

### LINKED ISSUE #71081: SceneTree.CreateTimer() does not free the timer after its time elapses
### Godot version

4.0.beta10.mono.official [d0398f62f]

### System information

Windows 11 Home 64-bit (Build 22621.963), Intel Core i5 11400F, NVIDIA GeForce RTX 3070

### Issue description

While the [document](https://docs.godotengine.org/en/latest/classes/class_scenetree.html#class-scenetree-method-create-timer) says "The timer will be automatically freed after its time elapses.", the Debugger says the opposite.

The timer created by `SceneTree.CreateTimer()` is not freed automatically after its time elapses. The Object count and Memory usage in the Debugger tab keep increasing.

### Steps to reproduce

1. Create new project.
2. Create C# script `CreateTimerTester.cs` with the following contents:
```C#
using Godot;
public partial class CreateTimerTester : Node2D
{
    public override void _Ready()
    {
        base._Ready();
        TestCreateTimer();
    }

    private async void TestCreateTimer()
    {
        var tree = GetTree();
        do
        {
            await ToSignal(tree.CreateTimer(0.1, true, false, true), "timeout");
        }
        while (true);
    }
}
```
3. Create new scene. Create a Node2D. Attach the script to the node.
4. Play the scene. Observe the Object count and Memory usage in the Debugger tab.

(Optional) 5. If you call `TestCreateTimer();` multiple times with a for-loop, it can be more effective. Here is a screenshot of calling it 5000 times and then stopping the loops by a cancellation token after 10 seconds:
Screenshot 2023-01-08 174024
The full code for step 5 is in the reproduction project.

[TestCreateTimer.zip](https://github.com/godotengine/godot/files/10369124/TestCreateTimer.zip)

### LINKED ISSUE #87515: C#: Non-instantiable GodotObject can't be GC-collected.
### Tested versions

4.x

### System information

Win11

### Issue description

Non-instantiable classes including `SceneTreeTimer`, `DirAccess`, and `PhysicsDirectSpaceState2D`, etc.

These types of managed instance cannot be collected by GC (even if they are disposed).

### Steps to reproduce

Use MRP below. I add a button to manually `GC.Collect()`.

### Case 1:

```csharp
for (var i = 0; i < 1000; i++)
{
    var atlas = new AtlasTexture(); // instantiable RefCounted
    var fa = DirAccess.Open("res://"); // non-instantiable RefCounted
}
```

After forcing `GC.Collect()`, only `AtlasTexture`s are collected (and freed). You can check this on Profiler tab in the editor.

### Case 2:

```csharp
for (var i = 0; i< 1000;i++)
{
    using var node = GetNode("Button"); // instantiable GodotObject
    using var state = GetWorld2D().DirectSpaceState; // PhysicsDirectSpaceState2D, non-instantiable GodotObject
}
```
This actually creates 1000*2 temp instances referring to one same native object, and they should be collected on next GC.

However, after `GC.Collect()`, only the instances of node are collected. The 1000 instances of state are still alive.

You can check this using memory inspector like dotMemory or Visual Studio:

image
_Memory capture after forcing GC, PhysicsDirectSpaceState2D (and DirAccess in case 1) are still alive._

[gc_collect_test.zip](https://github.com/godotengine/godot/files/14029719/gc_collect_test.zip)

zaevi:
- Fix #87515.

As #87515 explained, the binding generator didn't consider `_ConstructAndInitialize` for non-instantiable types.

This PR modifies its generation.
From:
```csharp
internal DirAccess() {}
```
To:
```csharp
internal DirAccess() : this(true)
{
    unsafe
    {
        _ConstructAndInitialize(null, NativeName, CachedType, refCounted: true);
    }
}
```

This will pass correct native type (`CachedType` above) and avoid accident GC handle.

- Help to close #71081.
- May alternative to #75434.

### DIFF
--- a/modules/mono/glue/GodotSharp/GodotSharp/Core/GodotObject.base.cs
+++ b/modules/mono/glue/GodotSharp/GodotSharp/Core/GodotObject.base.cs
@@ -38,6 +38,8 @@ namespace Godot
         {
             if (NativePtr == IntPtr.Zero)
             {
+                Debug.Assert(nativeCtor != null);
+
                 NativePtr = nativeCtor();
 
                 InteropUtils.TieManagedToUnmanaged(this, NativePtr,
--- a/modules/mono/editor/bindings_generator.cpp
+++ b/modules/mono/editor/bindings_generator.cpp
@@ -1621,9 +1621,13 @@ Error BindingsGenerator::_generate_cs_type(const TypeInterface &itype, const Str
 					   << CLOSE_BLOCK_L2 CLOSE_BLOCK_L1;
 			} else {
 				// Hide the constructor
-				output.append(MEMBER_BEGIN "internal ");
-				output.append(itype.proxy_name);
-				output.append("() {}\n");
+				output << MEMBER_BEGIN "internal " << itype.proxy_name << "() : this("
+					   << (itype.memory_own ? "true" : "false") << ")\n" OPEN_BLOCK_L1
+					   << INDENT2 "unsafe\n" INDENT2 OPEN_BLOCK
+					   << INDENT3 "_ConstructAndInitialize(null, "
+					   << BINDINGS_NATIVE_NAME_FIELD ", CachedType, refCounted: "
+					   << (itype.is_ref_counted ? "true" : "false") << ");\n"
+					   << CLOSE_BLOCK_L2 CLOSE_BLOCK_L1;
 			}
 
 			// Add.. em.. trick constructor. Sort of.


modules/mono/glue/GodotSharp/GodotSharp/Core/GodotObject.base.cs
```diff
@@ -38,6 +38,9 @@ bool refCounted
         {
             if (NativePtr == IntPtr.Zero)
             {
+                if (nativeCtor == null)
+                    throw new NativeConstructorNotFoundException(nativeName);
```
zaevi, line 42: Add null checking for `nativeCtor`. However, those non-instantiable objects are created from the engine and have `NativePtr` already, so this would never be thrown.

kkjinping:
good job

modules/mono/glue/GodotSharp/GodotSharp/Core/GodotObject.base.cs
raulsntos: I'm not sure if I like this check, specially if it's never going to be true. What do you think about using `Debug.Assert` instead?

zaevi: I don't like it either :>

raulsntos (approved):
Looks great, thanks! Since this looks like it fixes the same issue as #75434, I think it should supersede it.

akien-mga:
Thanks!

PR #81458, opened 2023-09-08
Documentation: Add support for deprecated/experimental messages

dalexeev:
Add support for deprecated/experimental messages to core and GDScript. See:

* godotengine/godot-proposals#6559
* godotengine/godot-proposals#7622

```gdscript
## my_func desc.
## @deprecated: Use [method another] instead.
func my_func():
    pass
```

* *Bugsquad edit, fixes: https://github.com/godotengine/godot-docs/issues/7900*

* _Production edit: closes godotengine/godot-roadmap#71_

PR #75012, opened 2023-03-17
Fix auto-translations in editor

KoBeWi:
This PR enables auto-translation of Windows and Controls inside the editor, if they are not part of the edited scene.
Fixes #50650
Technically `TTR()` can now be retired in favor of `TTRC()`, probably.

I use `get_viewport() != get_tree()->get_root()` to determine whether a node is part of the scene tree, but this might yield some false-positives. A more reliable condition is
```C++
if (Engine::get_singleton()->is_editor_hint() && get_tree()->get_edited_scene_root() &&
		(get_tree()->get_edited_scene_root()->get_viewport() == get_viewport() || get_tree()->get_edited_scene_root()->get_viewport()->is_ancestor_of(get_viewport())))
```
and for Window
```C++
if (Engine::get_singleton()->is_editor_hint() && get_tree()->get_edited_scene_root() &&
		(get_tree()->get_edited_scene_root()->get_viewport() == get_parent_viewport() || get_tree()->get_edited_scene_root()->get_viewport()->is_ancestor_of(get_parent_viewport()))) {
```
but not sure if it's necessary 🤔

PR #87530, opened 2024-01-24
Rework the auto translation system

### LINKED ISSUE #8897: Make `auto_translate` state inheritable

The Godot Engine.

`auto_translate` doesn't work correctly on some complex built-in `Control`s (e. g. `FileDialog`), due to the children being unaware if its parent's `auto_translate` is enabled or not.

Add a new `auto_translate_mode` member, that would work like `process_mode`, having values like `INHERIT`, `ALWAYS`, and `DISABLED`. This would not only fix the issue, but allow for some more advanced scenarios for users (e. g. having an entire portion of the UI not being translated).

To preserve compatibility, it would work as follows:

- `auto_translated_mode` would be added with just the `INHERIT` (default) and `ALWAYS` values.
- If `auto_translate` is disabled, it would be the same as setting `auto_translated_mode` to `DISABLED`.
- Once we enter a phase were we can break compatibility again (likely Godot 5), `auto_translate` is removed and `DISABLED` is added to `auto_translate_mode`.

Not really, since this is for built-in nodes.

See above.

YeldhamDev:
Closes https://github.com/godotengine/godot-proposals/issues/8897.

This PR adds the `auto_translate_mode` member to `Node`, deprecating `auto_translate` in `Control` and `Window`. This not only removes the need for duplicate code between those two, but also makes it easier to add it to other node types in the future as well (such as `Label3D`). `atr()` was also moved to `Node` for the same reasons above.

Here's a list of the available modes:
- `AUTO_TRANSLATE_MODE_ALWAYS` - Makes `auto_translate` work exactly as it does now.
- `AUTO_TRANSLATE_MODE_INHERIT` - The new default, inherits the `auto_translate` value of its parent.
- `AUTO_TRANSLATE_MODE_DISABLED` - The same as if `auto_translate` was toggled off.

The new default slight breaks compatibility, but I think it's for the better.

### DIFF
--- a/doc/classes/Control.xml
+++ b/doc/classes/Control.xml
@@ -937,9 +937,8 @@
 		<member name="anchor_top" type="float" setter="_set_anchor" getter="get_anchor" default="0.0">
 			Anchors the top edge of the node to the origin, the center or the end of its parent control. It changes how the top offset updates when the node moves or changes size. You can use one of the [enum Anchor] constants for convenience.
 		</member>
-		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true">
+		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true" deprecated="Use [member Node.auto_translate_mode] instead.">
 			Toggles if any text should automatically change to its translated version depending on the current locale.
-			Also decides if the node's strings should be parsed for POT generation.
 		</member>
 		<member name="clip_contents" type="bool" setter="set_clip_contents" getter="is_clipping_contents" default="false">
 			Enables whether rendering of [CanvasItem] based children should be clipped to this control's rectangle. If [code]true[/code], parts of a child which would be visibly outside of this control's rectangle will not be rendered and won't receive input.
--- a/doc/classes/Node.xml
+++ b/doc/classes/Node.xml
@@ -186,6 +186,32 @@
 				[b]Note:[/b] [SceneTree]'s group methods will [i]not[/i] work on this node if not inside the tree (see [method is_inside_tree]).
 			</description>
 		</method>
+		<method name="atr" qualifiers="const">
+			<return type="String" />
+			<param index="0" name="message" type="String" />
+			<param index="1" name="context" type="StringName" default="&quot;&quot;" />
+			<description>
+				Translates a [param message], using the translation catalogs configured in the Project Settings. Further [param context] can be specified to help with the translation.
+				This method works the same as [method Object.tr], with the addition of respecting the [member auto_translate_mode] state.
+				If [method Object.can_translate_messages] is [code]false[/code], or no translation is available, this method returns the [param message] without changes. See [method Object.set_message_translation].
+				For detailed examples, see [url=$DOCS_URL/tutorials/i18n/internationalizing_games.html]Internationalizing games[/url].
+			</description>
+		</method>
+		<method name="atr_n" qualifiers="const">
+			<return type="String" />
+			<param index="0" name="message" type="String" />
+			<param index="1" name="plural_message" type="StringName" />
+			<param index="2" name="n" type="int" />
+			<param index="3" name="context" type="StringName" default="&quot;&quot;" />
+			<description>
+				Translates a [param message] or [param plural_message], using the translation catalogs configured in the Project Settings. Further [param context] can be specified to help with the translation.
+				This method works the same as [method Object.tr_n], with the addition of respecting the [member auto_translate_mode] state.
+				If [method Object.can_translate_messages] is [code]false[/code], or no translation is available, this method returns [param message] or [param plural_message], without changes. See [method Object.set_message_translation].
+				The [param n] is the number, or amount, of the message's subject. It is used by the translation system to fetch the correct plural form for the current language.
+				For detailed examples, see [url=$DOCS_URL/tutorials/i18n/localization_using_gettext.html]Localization using gettext[/url].
+				[b]Note:[/b] Negative and [float] numbers may not properly apply to some countable subjects. It's recommended to handle these cases with [method atr].
+			</description>
+		</method>
 		<method name="call_deferred_thread_group" qualifiers="vararg">
 			<return type="Variant" />
 			<param index="0" name="method" type="StringName" />
@@ -203,7 +229,7 @@
 		<method name="can_process" qualifiers="const">
 			<return type="bool" />
 			<description>
-				Returns [code]true[/code] if the node can receive processing notifications and input callbacks ([constant NOTIFICATION_PROCESS], [method _input], etc) from the [SceneTree] and [Viewport]. The returned value depends on [member process_mode]:
+				Returns [code]true[/code] if the node can receive processing notifications and input callbacks ([constant NOTIFICATION_PROCESS], [method _input], etc.) from the [SceneTree] and [Viewport]. The returned value depends on [member process_mode]:
 				- If set to [constant PROCESS_MODE_PAUSABLE], returns [code]true[/code] when the game is processing, i.e. [member SceneTree.paused] is [code]false[/code];
 				- If set to [constant PROCESS_MODE_WHEN_PAUSED], returns [code]true[/code] when the game is paused, i.e. [member SceneTree.paused] is [code]true[/code];
 				- If set to [constant PROCESS_MODE_ALWAYS], always returns [code]true[/code];
@@ -923,6 +949,10 @@
 		</method>
 	</methods>
 	<members>
+		<member name="auto_translate_mode" type="int" setter="set_auto_translate_mode" getter="get_auto_translate_mode" enum="Node.AutoTranslateMode" default="0">
+			Defines if any text should automatically change to its translated version depending on the current locale (for nodes such as [Label], [RichTextLabel], [Window], etc.). See [enum AutoTranslateMode].
+			Also decides if the node's strings should be parsed for POT generation.
+		</member>
 		<member name="editor_description" type="String" setter="set_editor_description" getter="get_editor_description" default="&quot;&quot;">
 			An optional description to the node. It will be displayed as a tooltip when hovering over the node in the editor's Scene dock.
 		</member>
@@ -1180,10 +1210,10 @@
 			Notification received when the [TextServer] is changed.
 		</constant>
 		<constant name="PROCESS_MODE_INHERIT" value="0" enum="ProcessMode">
-			Inherits [member process_mode] from the node's parent. For the root node, it is equivalent to [constant PROCESS_MODE_PAUSABLE]. This is the default for any newly created node.
+			Inherits [member process_mode] from the node's parent. This is the default for any newly created node.
 		</constant>
 		<constant name="PROCESS_MODE_PAUSABLE" value="1" enum="ProcessMode">
-			Stops processing when [member SceneTree.paused] is [code]true[/code]. This is the inverse of [constant PROCESS_MODE_WHEN_PAUSED].
+			Stops processing when [member SceneTree.paused] is [code]true[/code]. This is the inverse of [constant PROCESS_MODE_WHEN_PAUSED], and the default for the root node.
 		</constant>
 		<constant name="PROCESS_MODE_WHEN_PAUSED" value="2" enum="ProcessMode">
 			Process [b]only[/b] when [member SceneTree.paused] is [code]true[/code]. This is the inverse of [constant PROCESS_MODE_PAUSABLE].
@@ -1232,6 +1262,16 @@
 		</constant>
 		<constant name="INTERNAL_MODE_BACK" value="2" enum="InternalMode">
 			The node will be placed at the end of the parent's children list, after any non-internal sibling.
+		</constant>
+		<constant name="AUTO_TRANSLATE_MODE_INHERIT" value="0" enum="AutoTranslateMode">
+			Inherits [member auto_translate_mode] from the node's parent. This is the default for any newly created node.
+		</constant>
+		<constant name="AUTO_TRANSLATE_MODE_ALWAYS" value="1" enum="AutoTranslateMode">
+			Always automatically translate. This is the inverse of [constant AUTO_TRANSLATE_MODE_DISABLED], and the default for the root node.
+		</constant>
+		<constant name="AUTO_TRANSLATE_MODE_DISABLED" value="2" enum="AutoTranslateMode">
+			Never automatically translate. This is the inverse of [constant AUTO_TRANSLATE_MODE_ALWAYS].
+			String parsing for POT generation will be skipped for this node and children that are set to [constant AUTO_TRANSLATE_MODE_INHERIT].
 		</constant>
 	</constants>
 </class>
--- a/doc/classes/Window.xml
+++ b/doc/classes/Window.xml
@@ -557,7 +557,7 @@
 		<member name="always_on_top" type="bool" setter="set_flag" getter="get_flag" default="false">
 			If [code]true[/code], the window will be on top of all other windows. Does not work if [member transient] is enabled.
 		</member>
-		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true">
+		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true" deprecated="Use [member Node.auto_translate_mode] instead.">
 			Toggles if any text should automatically change to its translated version depending on the current locale.
 		</member>
 		<member name="borderless" type="bool" setter="set_flag" getter="get_flag" default="false">
--- a/editor/plugins/packed_scene_translation_parser_plugin.cpp
+++ b/editor/plugins/packed_scene_translation_parser_plugin.cpp
@@ -52,32 +52,58 @@ Error PackedSceneEditorTranslationParserPlugin::parse_file(const String &p_path,
 	Ref<SceneState> state = Ref<PackedScene>(loaded_res)->get_state();
 
 	Vector<String> parsed_strings;
-	List<String> tabcontainer_paths;
+	Vector<Pair<NodePath, bool>> atr_owners;
+	Vector<String> tabcontainer_paths;
 	for (int i = 0; i < state->get_node_count(); i++) {
 		String node_type = state->get_node_type(i);
-		bool is_control = ClassDB::is_parent_class(node_type, "Control");
-		if (!is_control && !ClassDB::is_parent_class(node_type, "Window")) {
-			continue;
+		String parent_path = state->get_node_path(i, true);
+
+		// Find the `auto_translate_mode` property.
+		bool auto_translating = true;
+		bool auto_translate_mode_found = false;
+		for (int j = 0; j < state->get_node_property_count(i); j++) {
+			if (state->get_node_property_name(i, j) != "auto_translate_mode") {
+				continue;
+			}
+
+			auto_translate_mode_found = true;
+
+			int idx_last = atr_owners.size() - 1;
+			if (idx_last > 0 && !parent_path.begins_with(atr_owners[idx_last].first)) {
+				// Switch to the previous auto translation owner this was nested in, if that was the case.
+				atr_owners.remove_at(idx_last);
+				idx_last -= 1;
+			}
+
+			int auto_translate_mode = (int)state->get_node_property_value(i, j);
+			if (auto_translate_mode == Node::AUTO_TRANSLATE_MODE_DISABLED) {
+				auto_translating = false;
+			}
+
+			atr_owners.push_back(Pair(state->get_node_path(i), auto_translating));
+
+			break;
 		}
 
-		// Find the `auto_translate` property, and abort the string parsing of the node if disabled.
-		bool auto_translating = true;
-		for (int j = 0; j < state->get_node_property_count(i); j++) {
-			if (state->get_node_property_name(i, j) == "auto_translate" && (bool)state->get_node_property_value(i, j) == false) {
-				auto_translating = false;
-				break;
+		// If `auto_translate_mode` wasn't found, that means it is set to its default value (`AUTO_TRANSLATE_MODE_INHERIT`).
+		if (!auto_translate_mode_found) {
+			int idx_last = atr_owners.size() - 1;
+			if (idx_last > 0 && atr_owners[idx_last].first == parent_path) {
+				auto_translating = atr_owners[idx_last].second;
+			} else {
+				atr_owners.push_back(Pair(state->get_node_path(i), true));
 			}
 		}
 
 		// Parse the names of children of `TabContainer`s, as they are used for tab titles.
 		if (!tabcontainer_paths.is_empty()) {
-			String parent_path = state->get_node_path(i, true);
 			if (!parent_path.begins_with(tabcontainer_paths[tabcontainer_paths.size() - 1])) {
 				// Switch to the previous `TabContainer` this was nested in, if that was the case.
-				tabcontainer_paths.pop_back();
+				tabcontainer_paths.remove_at(tabcontainer_paths.size() - 1);
 			}
 
-			if (is_control && auto_translating && !tabcontainer_paths.is_empty() && parent_path == tabcontainer_paths[tabcontainer_paths.size() - 1]) {
+			if (auto_translating && !tabcontainer_paths.is_empty() && ClassDB::is_parent_class(node_type, "Control") &&
+					parent_path == tabcontainer_paths[tabcontainer_paths.size() - 1]) {
 				parsed_strings.push_back(state->get_node_name(i));
 			}
 		}
--- a/scene/gui/control.cpp
+++ b/scene/gui/control.cpp
@@ -268,6 +268,7 @@ String Control::properties_managed_by_container[] = {
 bool Control::_set(const StringName &p_name, const Variant &p_value) {
 	ERR_MAIN_THREAD_GUARD_V(false);
 	String name = p_name;
+
 	if (!name.begins_with("theme_override")) {
 		return false;
 	}
@@ -309,7 +310,6 @@ bool Control::_set(const StringName &p_name, const Variant &p_value) {
 		} else {
 			return false;
 		}
-
 	} else {
 		if (name.begins_with("theme_override_icons/")) {
 			String dname = name.get_slicec('/', 1);
@@ -333,12 +333,14 @@ bool Control::_set(const StringName &p_name, const Variant &p_value) {
 			return false;
 		}
 	}
+
 	return true;
 }
 
 bool Control::_get(const StringName &p_name, Variant &r_ret) const {
 	ERR_MAIN_THREAD_GUARD_V(false);
 	String sname = p_name;
+
 	if (!sname.begins_with("theme_override")) {
 		return false;
 	}
@@ -2996,7 +2998,6 @@ void Control::set_layout_direction(Control::LayoutDirection p_direction) {
 	ERR_FAIL_INDEX((int)p_direction, 4);
 
 	data.layout_dir = p_direction;
-	data.is_rtl_dirty = true;
 
 	propagate_notification(NOTIFICATION_LAYOUT_DIRECTION_CHANGED);
 }
@@ -3097,21 +3098,17 @@ bool Control::is_localizing_numeral_system() const {
 	return data.localize_numeral_system;
 }
 
+#ifndef DISABLE_DEPRECATED
 void Control::set_auto_translate(bool p_enable) {
 	ERR_MAIN_THREAD_GUARD;
-	if (p_enable == data.auto_translate) {
-		return;
-	}
-
-	data.auto_translate = p_enable;
-
-	notification(MainLoop::NOTIFICATION_TRANSLATION_CHANGED);
+	set_auto_translate_mode(p_enable ? AUTO_TRANSLATE_MODE_ALWAYS : AUTO_TRANSLATE_MODE_DISABLED);
 }
 
 bool Control::is_auto_translating() const {
 	ERR_READ_THREAD_GUARD_V(false);
-	return data.auto_translate;
+	return can_auto_translate();
 }
+#endif
 
 // Extra properties.
 
@@ -3172,14 +3169,6 @@ void Control::_notification(int p_notification) {
 		} break;
 
 		case NOTIFICATION_ENTER_TREE: {
-#ifdef TOOLS_ENABLED
-			if (is_part_of_edited_scene()) {
-				// Don't translate Controls on scene when inside editor.
-				set_message_translation(false);
-				notification(NOTIFICATION_TRANSLATION_CHANGED);
-			}
-#endif
-
 			// Emits NOTIFICATION_THEME_CHANGED internally.
 			set_theme_context(ThemeDB::get_singleton()->get_nearest_theme_context(this));
 		} break;
@@ -3192,6 +3181,7 @@ void Control::_notification(int p_notification) {
 
 		case NOTIFICATION_EXIT_TREE: {
 			set_theme_context(nullptr, false);
+
 			release_focus();
 			get_viewport()->_gui_remove_control(this);
 		} break;
@@ -3506,8 +3496,10 @@ void Control::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_layout_direction"), &Control::get_layout_direction);
 	ClassDB::bind_method(D_METHOD("is_layout_rtl"), &Control::is_layout_rtl);
 
+#ifndef DISABLE_DEPRECATED
 	ClassDB::bind_method(D_METHOD("set_auto_translate", "enable"), &Control::set_auto_translate);
 	ClassDB::bind_method(D_METHOD("is_auto_translating"), &Control::is_auto_translating);
+#endif
 
 	ClassDB::bind_method(D_METHOD("set_localize_numeral_system", "enable"), &Control::set_localize_numeral_system);
 	ClassDB::bind_method(D_METHOD("is_localizing_numeral_system"), &Control::is_localizing_numeral_system);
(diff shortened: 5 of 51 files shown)


scene/gui/control.cpp
```diff
@@ -3687,6 +3777,10 @@ void Control::_bind_methods() {
 	BIND_ENUM_CONSTANT(TEXT_DIRECTION_LTR);
 	BIND_ENUM_CONSTANT(TEXT_DIRECTION_RTL);
 
+	BIND_ENUM_CONSTANT(AUTO_TRANSLATE_MODE_INHERIT);
+	BIND_ENUM_CONSTANT(AUTO_TRANSLATE_MODE_ALWAYS);
+	// BIND_ENUM_CONSTANT(AUTO_TRANSLATE_MODE_DISABLED);
```
AThousandShips, line 3782:
```suggestion
	//BIND_ENUM_CONSTANT(AUTO_TRANSLATE_MODE_DISABLED);
```

scene/gui/control.h
```diff
@@ -149,6 +149,12 @@ class Control : public CanvasItem {
 		TEXT_DIRECTION_INHERITED = TextServer::DIRECTION_INHERITED,
 	};
 
+	enum AutoTranslateMode {
+		AUTO_TRANSLATE_MODE_INHERIT, // Same as parent `Control`/`Window`.
+		AUTO_TRANSLATE_MODE_ALWAYS, // Always auto translate.
+		// AUTO_TRANSLATE_MODE_DISABLED, // Never auto translate. Commented until compatibility breakage is allowed, where it will replace the `auto_translate` property.
```
AThousandShips, line 155:
```suggestion
		//AUTO_TRANSLATE_MODE_DISABLED, // Never auto translate. Commented until compatibility breakage is allowed, where it will replace the `auto_translate` property.
```

scene/gui/control.h
```diff
@@ -624,11 +634,17 @@ class Control : public CanvasItem {
...
+	void propagate_auto_translate_owner(Node *p_owner);
+	Node *get_auto_translate_owner() const;
+
 	void set_auto_translate(bool p_enable);
 	bool is_auto_translating() const;
-	_FORCE_INLINE_ String atr(const String p_string) const {
-		return is_auto_translating() ? tr(p_string) : p_string;
-	};
+
+	_FORCE_INLINE_ String atr(const String p_string) const { return can_auto_translate() ? tr(p_string) : p_string; };
```
AThousandShips, line 647:
```suggestion
	_FORCE_INLINE_ String atr(const String p_string) const { return can_auto_translate() ? tr(p_string) : p_string; }
```

scene/main/window.h
```diff
@@ -389,15 +393,25 @@ class Window : public Viewport {
...
+
+	void propagate_auto_translate_owner(Node *p_owner);
+	Node *get_auto_translate_owner() const;
+
 	void set_auto_translate(bool p_enable);
 	bool is_auto_translating() const;
-	_FORCE_INLINE_ String atr(const String p_string) const { return is_auto_translating() ? tr(p_string) : p_string; };
 
-	Rect2i get_usable_parent_rect() const;
+	_FORCE_INLINE_ String atr(const String p_string) const { return can_auto_translate() ? tr(p_string) : p_string; };
```
AThousandShips, line 414:
```suggestion
	_FORCE_INLINE_ String atr(const String p_string) const { return can_auto_translate() ? tr(p_string) : p_string; }
```

YeldhamDev:
@AThousandShips Changes made.

Mickeon:
Where's the "new default" in the current version of this PR? I don't think I see it.

timothyqiu:
Can we deprecate `auto_translate`, and make `auto_translate_mode` one of `DISABLED`, `INHERIT`, and `ALWAYS`?

To keep things compatible, make `auto_translate` backed by `auto_translate_mode` under the hood.

- If existing code sets `auto_translate`, set `auto_translate_mode` to `DISABLED` / `ALWAYS` accordingly.
- Hide `auto_translate` in the Inspector and don't save it when saving the scene.
   - If exsiting scene enabled it, `auto_translate_mode` will be updated when opening the scene.

akien-mga:
> Can we deprecate `auto_translate`, and make `auto_translate_mode` one of `DISABLED`, `INHERIT`, and `ALWAYS`?
> 
> To keep things compatible, make `auto_translate` backed by `auto_translate_mode` under the hood.
> 
>     * If existing code sets `auto_translate`, set `auto_translate_mode` to `DISABLED` / `ALWAYS` accordingly.
> 
>     * Hide `auto_translate` in the Inspector and don't save it when saving the scene.
>       
>       * If exsiting scene enabled it, `auto_translate_mode` will be updated when opening the scene.

Yes, this can be done by handling `auto_translate` is `_set`/`_get`, it doesn't even need a property registered formally (so there's no need to hide it).

scene/gui/control.h
```diff
@@ -149,6 +149,12 @@ class Control : public CanvasItem {
 		TEXT_DIRECTION_INHERITED = TextServer::DIRECTION_INHERITED,
 	};
 
+	enum AutoTranslateMode {
+		AUTO_TRANSLATE_MODE_INHERIT, // Same as parent `Control`/`Window`.
+		AUTO_TRANSLATE_MODE_ALWAYS, // Always auto translate.
+		//AUTO_TRANSLATE_MODE_DISABLED, // Never auto translate. Commented until compatibility breakage is allowed, where it will replace the `auto_translate` property.
```
Mickeon, line 155: Along with the suggestions to outright replace `auto_translate`, it would make more sense to shift **AUTO_TRANSLATE_MODE_DISABLED** as the first enum constant (0), in my opinion.

YeldhamDev: There doesn't seem to be an actual pattern with this sort of stuff in the codebase, even in the file itself. I just followed how `Node` does it with `process_mode` (where `DISABLED` is last).

YeldhamDev:
@timothyqiu @akien-mga The methods themselves would still need to be registered, right? To maintain GDNative compatibility I mean.

akien-mga:
> @timothyqiu @akien-mga The methods themselves would still need to be registered, right? To maintain GDNative compatibility I mean.

Yeah, but they can be registered as compat bindings. See e.g. `scene/2d/tile_map.compat.inc` for an example of how it's done.

Alternatively, we can keep the methods registered as normal methods and the property registered too, and mark them deprecated. Both options should be fine.

YeldhamDev:
Alright, with the increasing number of nodes that need auto translation that don't inherit `Control`, I think moving the auto translation properties back to `Node` is the only sane way forward.

editor/plugins/packed_scene_translation_parser_plugin.cpp
```diff
@@ -52,32 +52,53 @@ Error PackedSceneEditorTranslationParserPlugin::parse_file(const String &p_path,
...
 		// Parse the names of children of `TabContainer`s, as they are used for tab titles.
 		if (!tabcontainer_paths.is_empty()) {
-			String parent_path = state->get_node_path(i, true);
 			if (!parent_path.begins_with(tabcontainer_paths[tabcontainer_paths.size() - 1])) {
 				// Switch to the previous `TabContainer` this was nested in, if that was the case.
 				tabcontainer_paths.pop_back();
 			}
 
-			if (is_control && auto_translating && !tabcontainer_paths.is_empty() && parent_path == tabcontainer_paths[tabcontainer_paths.size() - 1]) {
+			if (auto_translating && ClassDB::is_parent_class(node_type, "Control") && !tabcontainer_paths.is_empty() &&
```
AThousandShips, line 100:
```suggestion
			if (auto_translating && !tabcontainer_paths.is_empty() && ClassDB::is_parent_class(node_type, "Control") &&
```
Let's put a cheaper condition first

scene/main/node.cpp
```diff
@@ -1149,6 +1164,49 @@ bool Node::is_processing_unhandled_key_input() const {
...
+	return data.auto_translate_mode;
+}
+
+bool Node::can_auto_translate() const {
+	ERR_READ_THREAD_GUARD_V(false);
+	if (!data.is_auto_translate_dirty || data.auto_translate_mode != AUTO_TRANSLATE_MODE_INHERIT) {
+		return data.is_auto_translating;
+	}
+
+	const_cast<Node *>(this)->data.is_auto_translate_dirty = false;
```
AThousandShips, line 1194: Maybe make this mutable instead, there's already `path_cache`

scene/main/node.cpp
```diff
@@ -1149,6 +1164,49 @@ bool Node::is_processing_unhandled_key_input() const {
...
+
+	Node *parent = data.parent;
+	while (parent) {
+		if (parent->data.auto_translate_mode == AUTO_TRANSLATE_MODE_INHERIT) {
+			parent = parent->data.parent;
+			continue;
+		}
+
+		const_cast<Node *>(this)->data.is_auto_translating = parent->data.auto_translate_mode == AUTO_TRANSLATE_MODE_ALWAYS;
```
AThousandShips, line 1203: Ditto

YeldhamDev:
@AThousandShips Changes made.

doc/classes/Control.xml
```diff
@@ -937,9 +937,8 @@
 		<member name="anchor_top" type="float" setter="_set_anchor" getter="get_anchor" default="0.0">
 			Anchors the top edge of the node to the origin, the center or the end of its parent control. It changes how the top offset updates when the node moves or changes size. You can use one of the [enum Anchor] constants for convenience.
 		</member>
-		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true">
-			Toggles if any text should automatically change to its translated version depending on the current locale.
-			Also decides if the node's strings should be parsed for POT generation.
+		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="false" is_deprecated="true">
+			Toggles if any text should automatically change to its translated version depending on the current locale. Use [member Node.auto_translate_mode] instead.
```
Mickeon, line 941: The deprecated message is more like this.
```suggestion
			Toggles if any text should automatically change to its translated version depending on the current locale.
			[i]Deprecated.[/i] Use [member Node.auto_translate_mode] instead.
```

doc/classes/Node.xml
```diff
@@ -905,6 +905,10 @@
 		</method>
 	</methods>
 	<members>
+		<member name="auto_translate_mode" type="int" setter="set_auto_translate_mode" getter="get_auto_translate_mode" enum="Node.AutoTranslateMode" default="0">
+			Defines if any text should automatically change to its translated version depending on the current locale (see [enum AutoTranslateMode]).
+			Also decides if the node's strings should be parsed for POT generation.
```
Mickeon, line 910: May be worth noting which built-in Nodes actually make use of this.

YeldhamDev:
@Mickeon Changes made.

doc/classes/Node.xml
```diff
@@ -905,6 +905,10 @@
 		</method>
 	</methods>
 	<members>
+		<member name="auto_translate_mode" type="int" setter="set_auto_translate_mode" getter="get_auto_translate_mode" enum="Node.AutoTranslateMode" default="0">
+			Defines if any text should automatically change to its translated version depending on the current locale (for nodes such as [Label], [RichTextLabel], [Window], etc). See [enum AutoTranslateMode].
```
Mickeon, line 909: The voice in my head called Yuri told me to add the period to "etc" last time.
```suggestion
			Defines if any text should automatically change to its translated version depending on the current locale (for nodes such as [Label], [RichTextLabel], [Window], etc.). See [enum AutoTranslateMode].
```

doc/classes/Node.xml
```diff
@@ -1161,10 +1165,10 @@
 			Notification received when the [TextServer] is changed.
 		</constant>
 		<constant name="PROCESS_MODE_INHERIT" value="0" enum="ProcessMode">
-			Inherits [member process_mode] from the node's parent. For the root node, it is equivalent to [constant PROCESS_MODE_PAUSABLE]. This is the default for any newly created node.
+			Inherits [member process_mode] from the node's parent. This is the default for any newly created node.
 		</constant>
 		<constant name="PROCESS_MODE_PAUSABLE" value="1" enum="ProcessMode">
-			Stops processing when [member SceneTree.paused] is [code]true[/code]. This is the inverse of [constant PROCESS_MODE_WHEN_PAUSED].
+			Stops processing when [member SceneTree.paused] is [code]true[/code]. This is the inverse of [constant PROCESS_MODE_WHEN_PAUSED], and the default for the root node.
```
Mickeon, line 1171: Out of scope of the PR, but sure! You did this for consistency with the other enum, that's okay. I mean, it is technically a lie, but that's how it behaves.

Mickeon (approved):
A single nitpick but 👍

doc/classes/Node.xml
YeldhamDev: Still better than "For the root node, it is equivalent to `PROCESS_MODE_PAUSABLE`." in `PROCESS_MODE_INHERIT` before.

doc/classes/Node.xml
YeldhamDev: There are plenty of instances of non-punctuated "etc" in the docs (including this very file), but sure.

Mickeon (approved):
Docs are just fine now, I cannot speak for the implementation itself.

RandomShaper (commented):
I think you can play it simple by using only main-thread guards. After all, this is not a functionality you need to play with from threaded node processing. They are very general properties that can be limited to the main thread just fine.

RandomShaper:
> @RandomShaper Is this thread guard necessary at all: https://github.com/godotengine/godot/pull/87530/files#diff-87dda4f114b485518869c93afac19b35800f98c7f10e4681bdbba1a2ec4f1928R1168
> 
> Since mainly it just does call `propagate_notification()` which already has a guard.

I think it is becase without it there's some data already being written to in a thread-unsafe manner before the other function eventually fails and, regardless, it's good to fail as early as the code realizes something incorrect.

KoBeWi:
Some notes:
- Object has a built-in mechanism for enabling/disabling translations - `can_translate_messages`, which affects `tr()`. It could be possibly used internally instead of adding `atr()`
- While the PR tries to keep compatibility, if you put an `auto_translate` node under a non-`auto_translate` node, the inherit will make the child non-`auto_translate` after upgrade. But since `auto_translate` is default and we don't save it in scene, I guess there is no way around that (also it's fairly rare case).

doc/classes/Node.xml
```diff
@@ -1219,5 +1223,14 @@
...
 			The node will be placed at the end of the parent's children list, after any non-internal sibling.
 		</constant>
+		<constant name="AUTO_TRANSLATE_MODE_INHERIT" value="0" enum="AutoTranslateMode">
+			Inherits [member auto_translate_mode] from the node's parent. This is the default for any newly created node.
+		</constant>
+		<constant name="AUTO_TRANSLATE_MODE_ALWAYS" value="1" enum="AutoTranslateMode">
+			Always automatically translate. This is the inverse of [constant AUTO_TRANSLATE_MODE_DISABLED], and the default for the root node.
+		</constant>
+		<constant name="AUTO_TRANSLATE_MODE_DISABLED" value="2" enum="AutoTranslateMode">
+			Never automatically translate. This is the inverse of [constant AUTO_TRANSLATE_MODE_ALWAYS].
```
KoBeWi, line 1233: Should probably mention that makes the node skipped from POT generation (the property already mentions it, but doesn't clarify).

KoBeWi: Well, I did not know what POT stands for until now, so I think it's not important. The editor always refers to it as POT.
Explaining it in more detail is something for a tutorial documentation page.

scene/main/node.cpp
```diff
@@ -136,17 +143,19 @@ void Node::_notification(int p_notification) {
...
 			}
-			// Remove the process group
+			// Remove the process group.
 			if (data.process_thread_group_owner == this) {
 				_remove_process_group();
 			}
 			data.process_thread_group_owner = nullptr;
 			data.process_owner = nullptr;
 
+			data.is_auto_translate_dirty = true;
```
KoBeWi, line 157: Why is it set here if the value is irrelevant outside tree?

scene/main/node.cpp
```diff
@@ -1149,6 +1164,49 @@ bool Node::is_processing_unhandled_key_input() const {
 	return data.unhandled_key_input;
 }
 
+void Node::set_auto_translate_mode(AutoTranslateMode p_mode) {
+	ERR_THREAD_GUARD
+	if (data.auto_translate_mode == p_mode) {
+		return;
+	}
+
+	if (p_mode == AUTO_TRANSLATE_MODE_INHERIT && is_inside_tree() && !data.parent) {
```
KoBeWi, line 1173:
```suggestion
	if (p_mode == AUTO_TRANSLATE_MODE_INHERIT && data.inside_tree && !data.parent) {
```

scene/gui/button.cpp
```diff
@@ -109,6 +109,13 @@ Ref<StyleBox> Button::_get_current_stylebox() const {
 
 void Button::_notification(int p_what) {
 	switch (p_what) {
+		case NOTIFICATION_ENTER_TREE: {
+			// Update text with the current translation state.
+			if (!text.is_empty() && get_auto_translate_mode() != AUTO_TRANSLATE_MODE_DISABLED) {
+				notification(NOTIFICATION_TRANSLATION_CHANGED);
```
KoBeWi, line 115: This looks like something that should be sent automatically, like `NOTIFICATION_THEME_CHANGED`.

editor/plugins/packed_scene_translation_parser_plugin.cpp
```diff
@@ -52,32 +52,53 @@ Error PackedSceneEditorTranslationParserPlugin::parse_file(const String &p_path,
...
 		for (int j = 0; j < state->get_node_property_count(i); j++) {
-			if (state->get_node_property_name(i, j) == "auto_translate" && (bool)state->get_node_property_value(i, j) == false) {
-				auto_translating = false;
-				break;
+			if (state->get_node_property_name(i, j) != "auto_translate_mode") {
+				continue;
 			}
+
+			int idx_last = atr_owners.size() - 1;
+			if (idx_last > 0 && !parent_path.begins_with(atr_owners[idx_last].first)) {
```
KoBeWi, line 69: Ooof, accessing list elements using index. I think Vector or LocalVector would be better suited here.

scene/gui/control.cpp
```diff
@@ -3506,8 +3502,10 @@ void Control::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_layout_direction"), &Control::get_layout_direction);
 	ClassDB::bind_method(D_METHOD("is_layout_rtl"), &Control::is_layout_rtl);
 
+#ifndef DISABLE_DEPRECATED
```
KoBeWi, line 3505: Technically the C++ methods should also be wrapped, but if we use `DISABLE_DEPRECATED` only to keep track of things, that would be unnecessary.

scene/main/window.h
```diff
@@ -31,10 +31,10 @@
 #ifndef WINDOW_H
 #define WINDOW_H
 
+#include "scene/gui/control.h"
```
KoBeWi, line 34: Why was this added?

scene/main/node.cpp
YeldhamDev: That was more from my first go at this PR, before placing the new member in `Node`. In `Control`, the `is_rtl_dirty` variable is also set to `true` on exiting (on canvas exiting, in that case).

scene/gui/button.cpp
YeldhamDev: Wouldn't that mean that every UI node with text would redraw on tree entering, even if it didn't need to?

YeldhamDev:
@KoBeWi

> Object has a built-in mechanism for enabling/disabling translations - can_translate_messages, which affects tr(). It could be possibly used internally instead of adding atr()

I don't know how that would work, since the check is done when a translation is requested with `atr()`, and `Object` isn't aware of `auto_translate_mode`.

> Seems like POT generator does not properly handle inherited disabled auto-translation. Such nodes are still included in the POT.

Fixed. That was happening because since `AUTO_TRANSLATE_MODE_INHERIT` is the default value, it wasn't being saved to the scene state at all, so now if `auto_translate_mode` is missing, it's assumed that value is the default one.

scene/main/window.h
YeldhamDev: Residue from the previous implementation, removed.

KoBeWi:
> I don't know how that would work, since the check is done when a translation is requested with atr(), and Object isn't aware of auto_translate_mode.

You could call `set_message_translation()` internally based on `auto_translate_mode`. Since the message translation is only exposed as simple setter/getter, the translate mode would be a higher level wrapper for that.

YeldhamDev:
How would that work with `AUTO_TRANSLATE_MODE_INHERIT`?

KoBeWi:
I guess it wouldn't 🤔 Since the current solution already works and is rather simple then it's fine as is.

doc/classes/Node.xml
```diff
@@ -1219,5 +1223,15 @@
...
+			String parsing for POT generation will be skipped for this and children that are set to [constant AUTO_TRANSLATE_MODE_INHERIT].
```
KoBeWi, line 1234:
```suggestion
			String parsing for POT generation will be skipped for this node and children that are set to [constant AUTO_TRANSLATE_MODE_INHERIT].
```
idk if this was intended.

scene/main/node.cpp
KoBeWi: Then `is_rtl_dirty` is probably also set unnecessarily.
I mean, the node won't display any text and you do the same when entering tree, so this line is effectively useless.

scene/main/node.cpp
```diff
@@ -217,6 +226,12 @@ void Node::_notification(int p_notification) {
 				memdelete(child);
 			}
 		} break;
+
+		case NOTIFICATION_TRANSLATION_CHANGED: {
+			if (is_inside_tree()) {
```
KoBeWi, line 231:
```suggestion
			if (data.inside_tree) {
```
Mostly for consistency (though it's slightly more performant if the compiler does not inline it automatically).

scene/gui/button.cpp
KoBeWi: Well, every CanvasItem already redraws on tree entering, so it doesn't make difference.

YeldhamDev: Alright then, gonna need to know why this method in your old PR is here though: https://github.com/godotengine/godot/pull/75012/files#diff-2b7ba034670859ced66e442b9ec59f7351febc3b4352e535bc85da804bcca266R2918

KoBeWi: The comment explains it 🙃
Nodes on the scene shouldn't be translated, because it would make editing them more difficult.

YeldhamDev: ...I meant the combination of `set_message_translation(false)` and `notification(NOTIFICATION_TRANSLATION_CHANGED)`. Doesn't those two cancel each other out?

KoBeWi: I think the TRANSLATION_CHANGED was to force the texts to update, as some of them can be set before entering tree. I don't remember if it actually fixed something, but I'd assume it did.

If the translations are now part of Node, this code should be moved.

YeldhamDev:
Changes finished.

KoBeWi (approved):
Looks fine now.
I only wonder about `atr()`, the name is rather obscure and could use a better one (now that it's exposed).

doc/classes/Node.xml
```diff
@@ -182,6 +182,32 @@
...
 			</description>
 		</method>
+		<method name="atr" qualifiers="const">
+			<return type="String" />
+			<param index="0" name="message" type="String" />
+			<param index="1" name="context" type="StringName" default="&quot;&quot;" />
+			<description>
+				Translates a [param message], using the translation catalogs configured in the Project Settings. Further [param context] can be specified to help with the translation.
+				If [method Object.can_translate_messages] is [code]false[/code], or no translation is available, this method returns the [param message] without changes. See [method Object.set_message_translation].
+				This method works the same as [method Object.tr], with the addition of respecting the [member auto_translate_mode] state.
```
Mickeon, line 192: This paragraph says the most important part. The fact that it automatically translates depending on the `auto_translate_mode` should appear much sooner in the description. Otherwise it's very easy to (accidentally) mix it around with `tr`.

(Same goes for `atr_n`)

Mickeon:
I agree `atr` may not be the best name for this, but perhaps it's the only way. The problem I have with it is that by "uncompressing" the shorthand you get `auto_translate`, but the "auto" in this context is extremely vague. `tr` also **automatically** translates stuff without it, people assume that's the whole point of it.


But part of me is also wondering, is exposing `atr` and `atr_n` truly necessary right now?
Excluding checking for inheritance, could the following snippet of code not achieve the same thing?
```GDScript
var cool_string = tr("Congratulations!") if auto_translate_mode == AUTO_TRANSLATE_MODE_ALWAYS else "cool_string"
```
I am struggling to think of occasions where you want to continuously change the `auto_translate_mode` to the point of needing something NOT to be translated at times.

Perhaps, judging by the snippet above, what is actually necessary is checking if the Node can automatically translate or not (akin to `can_process`), but even then it should need a bit of user demand.

doc/classes/Node.xml
YeldhamDev: Done.

KoBeWi:
> Excluding checking for inheritance, could the following snippet of code not achieve the same thing?

It could, but the point is inheritance. If a user makes a custom Control that displays text, they'd want the text to be translated properly like in other nodes. Without `atr()` exposed (or the `can_auto_translate()` method) that's not possible, unless you reimplement `can_auto_translate()` yourself.

YeldhamDev:
This PR is ready.

Mickeon:
I don't have anything against the documentation. I'm still questioning the usefulness of `atr` and `atr_n`. I'm in the opinion that a method called `can_auto_translate()` would be more generally useful and less confusing, but I'm not going to cry in my sleep about it.

doc/classes/Node.xml
```diff
@@ -186,6 +186,32 @@
...
+		</method>
+		<method name="atr_n" qualifiers="const">
+			<return type="String" />
+			<param index="0" name="message" type="String" />
+			<param index="1" name="plural_message" type="StringName" />
+			<param index="2" name="n" type="int" />
+			<param index="3" name="context" type="StringName" default="&quot;&quot;" />
+			<description>
+				Translates a [param message] or [param plural_message], using the translation catalogs configured in the Project Settings. Further [param context] can be specified to help with the translation.
+				This method works the same as [method Object.tr], with the addition of respecting the [member auto_translate_mode] state.
```
akien-mga, line 208:
```suggestion
				This method works the same as [method Object.tr_n], with the addition of respecting the [member auto_translate_mode] state.
```

doc/classes/Node.xml
```diff
@@ -186,6 +186,32 @@
...
+				If [method Object.can_translate_messages] is [code]false[/code], or no translation is available, this method returns [param message] or [param plural_message], without changes. See [method Object.set_message_translation].
+				The [param n] is the number, or amount, of the message's subject. It is used by the translation system to fetch the correct plural form for the current language.
+				For detailed examples, see [url=$DOCS_URL/tutorials/i18n/localization_using_gettext.html]Localization using gettext[/url].
+				[b]Note:[/b] Negative and [float] numbers may not properly apply to some countable subjects. It's recommended handling these cases with [method atr].
```
akien-mga, line 212:
```suggestion
				[b]Note:[/b] Negative and [float] numbers may not properly apply to some countable subjects. It's recommended to handle these cases with [method atr].
```

scene/gui/control.cpp
```diff
@@ -3558,9 +3550,12 @@ void Control::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "size_flags_stretch_ratio", PROPERTY_HINT_RANGE, "0,20,0.01,or_greater"), "set_stretch_ratio", "get_stretch_ratio");
 
 	ADD_GROUP("Localization", "");
-	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "auto_translate"), "set_auto_translate", "is_auto_translating");
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "localize_numeral_system"), "set_localize_numeral_system", "is_localizing_numeral_system");
```
akien-mga, line 3553: Just food for thought - is this something that could be streamlined together with the added `auto_translate_mode`, maybe adding enum values that enable both translation + numbers localization?

May not be a good idea, but just leaving this here for consideration.

scene/main/node.cpp
```diff
@@ -1149,6 +1173,49 @@ bool Node::is_processing_unhandled_key_input() const {
...
+	}
+
+	data.auto_translate_mode = p_mode;
+	data.is_auto_translating = p_mode != AUTO_TRANSLATE_MODE_DISABLED;
+	data.is_auto_translate_dirty = true;
+
+	propagate_notification(NOTIFICATION_TRANSLATION_CHANGED);
+}
+
+Node::AutoTranslateMode Node::get_auto_translate_mode() const {
```
akien-mga, line 1193: I don't think we include the enum prefix in the same class that defines it usually (might be wrong though, didn't check).
```suggestion
AutoTranslateMode Node::get_auto_translate_mode() const {
```

scene/main/node.h
```diff
@@ -211,6 +217,10 @@ class Node : public Object {
 		bool display_folded = false;
 		bool editable_instance = false;
 
+		AutoTranslateMode auto_translate_mode = AUTO_TRANSLATE_MODE_INHERIT;
+		mutable bool is_auto_translating = true;
+		mutable bool is_auto_translate_dirty = true;
```
akien-mga, line 222: This might be a silly question, but what makes these bools require `mutable` compared to other in the struct? Thread safety?

scene/main/window.cpp
```diff
@@ -2649,21 +2640,17 @@ bool Window::is_layout_rtl() const {
...
 void Window::set_auto_translate(bool p_enable) {
 	ERR_MAIN_THREAD_GUARD;
-	if (p_enable == auto_translate) {
-		return;
-	}
-
-	auto_translate = p_enable;
-
-	notification(MainLoop::NOTIFICATION_TRANSLATION_CHANGED);
+	set_auto_translate_mode(p_enable ? Control::AUTO_TRANSLATE_MODE_ALWAYS : Control::AUTO_TRANSLATE_MODE_DISABLED);
```
akien-mga, line 2646:
```suggestion
	set_auto_translate_mode(p_enable ? Node::AUTO_TRANSLATE_MODE_ALWAYS : Node::AUTO_TRANSLATE_MODE_DISABLED);
```

akien-mga (commented):
Quick code review. Looks good!
Still have to actually test.

akien-mga:
I tested it with this scene setup:

image

With Auto Translate *disabled* on FileDialog, and after setting the project test locale to `fr` and importing `editor/translations/editor/fr.po` as translations, it seems like the internal nodes of the dialog still get translated automatically. Only the dialog title seems to stay in English (and gets translated with Auto Translate *enabled*).

With Inherit/Always:
image

With Disabled:
image

MRP: [autotranslate_filedialog.zip](https://github.com/godotengine/godot/files/14297872/autotranslate_filedialog.zip)

scene/main/node.h
YeldhamDev: See https://github.com/godotengine/godot/pull/87530#discussion_r1469625184.

scene/main/node.cpp
YeldhamDev: We do. In fact, it raises an error if we don't. :slightly_smiling_face:

scene/gui/control.cpp
YeldhamDev: As much as I like the idea, how many extra options that would add? I feel like it would make it a little convoluted.

akien-mga: Yeah I don't really see how it would work to be honest :D

doc/classes/Window.xml
```diff
@@ -558,8 +558,9 @@
 		<member name="always_on_top" type="bool" setter="set_flag" getter="get_flag" default="false">
 			If [code]true[/code], the window will be on top of all other windows. Does not work if [member transient] is enabled.
 		</member>
-		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true">
+		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true" is_deprecated="true">
```
YeldhamDev, line 561: The checks are complaining that `is_deprecated="true"` should be `deprecated=""`, but no other doc file has it like that.

akien-mga: See #81458.

scene/gui/control.cpp
Mickeon: Clamping it all together in a BitField of some sort sounds interesting...

doc/classes/Control.xml
```diff
@@ -937,9 +937,9 @@
 		<member name="anchor_top" type="float" setter="_set_anchor" getter="get_anchor" default="0.0">
 			Anchors the top edge of the node to the origin, the center or the end of its parent control. It changes how the top offset updates when the node moves or changes size. You can use one of the [enum Anchor] constants for convenience.
 		</member>
-		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true">
+		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true" deprecated="Use [member Node.auto_translate_mode] instead.">
 			Toggles if any text should automatically change to its translated version depending on the current locale.
-			Also decides if the node's strings should be parsed for POT generation.
+			[i]Deprecated.[/i] Use [member Node.auto_translate_mode] instead.
```
dalexeev, line 942:
```suggestion
```

YeldhamDev:
Changes made.

akien-mga (approved):

akien-mga:
Thanks!

PR #87534, opened 2024-01-24
Cherry-picks for the 4.2 branch (future 4.2.2) - 1st batch

YuriSizov:
The first batch only contains PRs which also target 4.1 with their cherry-picking. There is going to be a second batch after that with PRs which are 4.2-only.


akien-mga (approved):
Looks good!

YuriSizov:
Thanks for the review!

PR #87541, opened 2024-01-24
Avoid saving the `texture_rd_rid` property of TextureRD resources

clayjohn:
This fixes an unreported bug I noticed while working on a more advanced demo using Texture2DRDs. 

The bug arises when using `@tool` scripts that modify Texture2DRds. The engine was saving the `texture_rd_rid` property to disk and then trying to load and initialize the texture using the old RID. That doesn't work because RIDs don't persist between runs. 

Also this PR removes the `texture_rd_rid` from the editor as you can't set the value from editor. It has to be set at run time after allocating an RD texture and obtaining its RID.

CC @BastiaanOlij

### DIFF
--- a/doc/classes/Texture2DRD.xml
+++ b/doc/classes/Texture2DRD.xml
@@ -10,7 +10,7 @@
 	</tutorials>
 	<members>
 		<member name="resource_local_to_scene" type="bool" setter="set_local_to_scene" getter="is_local_to_scene" overrides="Resource" default="false" />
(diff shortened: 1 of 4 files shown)


AThousandShips:
(remember to quote annotations so you don't tag accounts)

AThousandShips (approved):
Makes sense to me, did some related work previously

The editor part also makes perfect sense as they're not repeatable

BastiaanOlij (approved):
Good find! Looks good to me :)

YuriSizov:
Thanks!

PR #87542, opened 2024-01-24
Make the Script Editor's parser execute sooner on edit after error was found

### LINKED ISSUE #8945: Make the Script Editor's parser execute sooner on edit after error was found

Generally contributing to the engine.

I write code and I'm sure a lot of people reading this do too. I can make mistakes for a variety of reasons and the code parser is generally pretty good at telling me.

But it would also be nice if the **Script Editor** told me I fixed them right.
Suppose you misspell the name of a method. As you're writing the rest of the code an error will pop up, appropriately. You will then "fix" the code, but have to wait a bit to make sure that you did it correctly.

This wait time does add up. It's especially more noticeable for beginners, as they're likely to make more mistakes. It's overall a bit clunky to not see the errors disappear sooner.

Make the parser execute way sooner **if an error is known to exist**.

The time should depend on the user's `text_editor/completion/idle_parse_delay`, of course.
It could either be a hardcoded number of "half the time than usual", or exposed as a separate **Editor Setting**.

Unsure what to put here.

You could decrease the `text_editor/completion/idle_parse_delay` Editor Setting to something more comfortable for your needs. But it's not always nice. For some users, myself included, it makes the editor feel very "petty" or otherwise annoying to deal with.

_"Yes, I know! Of course there's an error, I wasn't even done writing the full thing!"_

The Script Editor is core and as far as I know there is no way to trigger the code analyser early.

Mickeon:
Closes https://github.com/godotengine/godot-proposals/issues/8945

This PR is fairly self-explanatory. And dang, does it feel good.

##### Delay is scaled way down for demonstration:

A new Editor Setting exists, ~~as a multiplier to tweak this behavior~~. I am using a default value of `0.5` and personally, I do find it great. but that may be personal preference.
Naming suggestions for this one would be nice.

Oh and, just for clarity, this does not apply to the Shader editor. Unsure why, honestly. It may be hardcoded. Or maybe this PR would break something about it?

### DIFF
--- a/doc/classes/EditorSettings.xml
+++ b/doc/classes/EditorSettings.xml
@@ -1292,6 +1292,9 @@
 		<member name="text_editor/completion/idle_parse_delay" type="float" setter="" getter="">
 			The delay in seconds after which the script editor should check for errors when the user stops typing.
 		</member>
+		<member name="text_editor/completion/idle_parse_delay_with_errors_found" type="float" setter="" getter="">
+			The delay used instead of [member text_editor/completion/idle_parse_delay], when the parser has found errors. A lower value should feel more responsive while fixing code, but may cause notable stuttering and increase CPU usage.
+		</member>
 		<member name="text_editor/completion/put_callhint_tooltip_below_current_line" type="bool" setter="" getter="">
 			If [code]true[/code], the code completion tooltip will appear below the current line unless there is no space on screen below the current line. If [code]false[/code], the code completion tooltip will appear above the current line.
 		</member>
--- a/editor/code_editor.cpp
+++ b/editor/code_editor.cpp
@@ -1119,7 +1119,12 @@ void CodeTextEditor::update_editor_settings() {
 	text_editor->set_code_hint_draw_below(EDITOR_GET("text_editor/completion/put_callhint_tooltip_below_current_line"));
 	code_complete_enabled = EDITOR_GET("text_editor/completion/code_complete_enabled");
 	code_complete_timer->set_wait_time(EDITOR_GET("text_editor/completion/code_complete_delay"));
-	idle->set_wait_time(EDITOR_GET("text_editor/completion/idle_parse_delay"));
+	bool first_time = idle_time == 0.0;
+	idle_time = EDITOR_GET("text_editor/completion/idle_parse_delay");
+	idle_time_with_errors = EDITOR_GET("text_editor/completion/idle_parse_delay_with_errors_found");
+	if (first_time) {
+		idle->set_wait_time(idle_time);
+	}
 
 	// Appearance: Guidelines
 	if (EDITOR_GET("text_editor/appearance/guidelines/show_line_length_guidelines")) {
@@ -1624,8 +1629,11 @@ void CodeTextEditor::_notification(int p_what) {
 void CodeTextEditor::set_error_count(int p_error_count) {
 	error_button->set_text(itos(p_error_count));
 	error_button->set_visible(p_error_count > 0);
-	if (!p_error_count) {
+	if (p_error_count > 0) {
 		_set_show_errors_panel(false);
+		idle->set_wait_time(idle_time_with_errors); // Parsing should happen sooner.
+	} else {
+		idle->set_wait_time(idle_time);
 	}
 }
 
@@ -1797,7 +1805,6 @@ CodeTextEditor::CodeTextEditor() {
 	add_child(status_bar);
 	status_bar->set_h_size_flags(SIZE_EXPAND_FILL);
 	status_bar->set_custom_minimum_size(Size2(0, 24 * EDSCALE)); // Adjust for the height of the warning icon.
-
 	idle = memnew(Timer);
 	add_child(idle);
 	idle->set_one_shot(true);
--- a/editor/code_editor.h
+++ b/editor/code_editor.h
@@ -174,6 +174,8 @@ class CodeTextEditor : public VBoxContainer {
 
 	Label *info = nullptr;
 	Timer *idle = nullptr;
+	float idle_time = 0.0f;
+	float idle_time_with_errors = 0.0f;
 	bool code_complete_enabled = true;
 	Timer *code_complete_timer = nullptr;
 	int code_complete_timer_line = 0;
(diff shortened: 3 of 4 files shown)


AThousandShips:
Can workshop it further but I think the name needed a bit of clarity that it "editing after an error" as opposed to "if we already (somehow) know an error exists before parsing"

MewPurPur:
I don't quite understand; so this PR makes it so if the parser is currently in its error state, it will parse sooner to check if there are still issues?

Mickeon:
I'm struggling to come up with a better name. `idle_parse_edit_error_scale`? Unconvinced.

MewPurPur:
`idle_parse_delay_in_error_state` would've been a good time if it weren't a multiplier.

Mickeon:
I really don't want to have it a separate value because it _sounds_ unreasonable not to clamp it. But if there is no other way, so be it, I suppose.

kitbdev:
I feel like it should be a separate value since it would be easier to use and there would be more control over it.
For instance, if I wanted to maximize the idle parse delay to 10 seconds, then the minimum I could set the error state parse delay is 1 second, where otherwise it would be 0.1 seconds. 

If it does need to be clamped, it could be done in code and mentioned in the description.

Mickeon:
It doesn't **need** to be clamped, per say, but any value greater than the usual delay is outstandingly counter-productive.

Mickeon:
Updated the PR to have the additional setting override the default delay.

> If it does need to be clamped, it could be done in code and mentioned in the description.

It's not worth enforcing it and writing a note about it. I left it up to the user to be silly.

Mickeon:
Updated the PR to have the additional setting override the default delay.

> If it does need to be clamped, it could be done in code and mentioned in the description.

It's not worth enforcing it and writing a note about it. I left it up to the user to be silly.

doc/classes/EditorSettings.xml
```diff
@@ -1292,6 +1292,9 @@
 		<member name="text_editor/completion/idle_parse_delay" type="float" setter="" getter="">
 			The delay in seconds after which the script editor should check for errors when the user stops typing.
 		</member>
+		<member name="text_editor/completion/idle_parse_delay_with_errors_found" type="float" setter="" getter="">
+			The delay used instead of [member text_editor/completion/idle_parse_delay], when the parser has found errors. A lower value should feel more responsive while fixing code.
```
KoBeWi, line 1296: Depends on what do you mean by "responsive". Lower value means the script is parsed more often, which can result in delays.

editor/code_editor.h
```diff
@@ -174,6 +174,8 @@ class CodeTextEditor : public VBoxContainer {
 
 	Label *info = nullptr;
 	Timer *idle = nullptr;
+	float idle_time;
+	float idle_time_with_errors;
```
KoBeWi, line 178:
```suggestion
	float idle_time = 0;
	float idle_time_with_errors = 0;
```

doc/classes/EditorSettings.xml
Mickeon: Then it may be worth mentioning something about processing cost here...

editor/code_editor.cpp
```diff
@@ -1798,9 +1802,12 @@ CodeTextEditor::CodeTextEditor() {
 	status_bar->set_h_size_flags(SIZE_EXPAND_FILL);
 	status_bar->set_custom_minimum_size(Size2(0, 24 * EDSCALE)); // Adjust for the height of the warning icon.
 
+	idle_time = EDITOR_GET("text_editor/completion/idle_parse_delay");
+	idle_time_with_errors = EDITOR_GET("text_editor/completion/idle_parse_delay_in_error_state");
```
KoBeWi, line 1806: Not needed, because `update_editor_settings()` is called automatically when the editor is created.

editor/code_editor.cpp
```diff
@@ -1798,9 +1802,12 @@ CodeTextEditor::CodeTextEditor() {
...
 	idle = memnew(Timer);
 	add_child(idle);
 	idle->set_one_shot(true);
+	idle->set_wait_time(idle_time);
```
KoBeWi, line 1810: But then this should be somewhere else I suppose.

Mickeon: But where?

KoBeWi: You can call it in `update_editor_settings()` if the time was 0.

Mickeon: In hindsight it's not that big of a deal to reset the `idle` Timer when editor settings are modified, but I will still do the change.

editor/code_editor.cpp
```diff
@@ -1119,7 +1119,11 @@ void CodeTextEditor::update_editor_settings() {
 	text_editor->set_code_hint_draw_below(EDITOR_GET("text_editor/completion/put_callhint_tooltip_below_current_line"));
 	code_complete_enabled = EDITOR_GET("text_editor/completion/code_complete_enabled");
 	code_complete_timer->set_wait_time(EDITOR_GET("text_editor/completion/code_complete_delay"));
-	idle->set_wait_time(EDITOR_GET("text_editor/completion/idle_parse_delay"));
+	if (idle_time == 0.0) {
+		idle->set_wait_time(idle_time);
```
KoBeWi, line 1123: This will initialize it to 0.
You can do `bool first_time = idle_time == 0.0` and move the condition after the setting is loaded.

Mickeon: There we go

KoBeWi (approved):

akien-mga:
Thanks!

editor/code_editor.h
```diff
@@ -174,6 +174,8 @@ class CodeTextEditor : public VBoxContainer {
 
 	Label *info = nullptr;
 	Timer *idle = nullptr;
+	float idle_time = 0.0f;
+	float idle_time_with_errors = 0.0f;
```
akien-mga, line 178: These should probably have sensible defaults, and/or we need to make sure to read the editor settings early enough, as in current `master` this PR causes:
```
ERROR: Time should be greater than zero.
   at: set_wait_time (./scene/main/timer.cpp:83)
```

PR #87543, opened 2024-01-24
Remove now unused `year` member from `version.py`

akien-mga:
We changed copyright to use "present" for the current year, so we no longer need to hardcode this and (like now) forget to bump it yearly.

_Technically_ it breaks compat by removing `year` from the `Engine.get_version_info()` dictionary.
I think it's in the scope of acceptable changes for 4.3, but shouldn't be cherry-picked to 4.2.x (instead, we'll just bump the `year` to 2024).

Non-compat breaking version for `3.x`: #87579

### DIFF
--- a/core/version.h
+++ b/core/version.h
@@ -33,6 +33,12 @@
 
 #include "core/version_generated.gen.h"
 
+// Copied from typedefs.h to stay lean.
+#ifndef _STR
+#define _STR(m_x) #m_x
+#define _MKSTR(m_x) _STR(m_x)
+#endif
+
 // Godot versions are of the form <major>.<minor> for the initial release,
 // and then <major>.<minor>.<patch> for subsequent bugfix releases where <patch> != 0
 // That's arbitrary, but we find it pretty and it's the current policy.
--- a/platform/windows/godot_res.rc
+++ b/platform/windows/godot_res.rc
@@ -1,16 +1,12 @@
 #include "core/version.h"
-#ifndef _STR
-#define _STR(m_x) #m_x
-#define _MKSTR(m_x) _STR(m_x)
-#endif
 
 GODOT_ICON ICON platform/windows/godot.ico
 
 1 VERSIONINFO
-FILEVERSION    	VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
-PRODUCTVERSION 	VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
-FILEOS         	4
-FILETYPE       	1
+FILEVERSION     VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
+PRODUCTVERSION  VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
+FILEOS          4
+FILETYPE        1
 BEGIN
     BLOCK "StringFileInfo"
     BEGIN
@@ -21,7 +17,7 @@ BEGIN
             VALUE "FileVersion",            VERSION_NUMBER
             VALUE "ProductName",            VERSION_NAME
             VALUE "Licence",                "MIT"
-            VALUE "LegalCopyright",         "Copyright (c) 2007-" _MKSTR(VERSION_YEAR) " Juan Linietsky, Ariel Manzur and contributors"
+            VALUE "LegalCopyright",         "(c) 2007-present Juan Linietsky, Ariel Manzur and Godot Engine contributors"
             VALUE "Info",                   "https://godotengine.org"
             VALUE "ProductVersion",         VERSION_FULL_BUILD
         END
--- a/platform/windows/godot_res_wrap.rc
+++ b/platform/windows/godot_res_wrap.rc
@@ -1,16 +1,12 @@
 #include "core/version.h"
-#ifndef _STR
-#define _STR(m_x) #m_x
-#define _MKSTR(m_x) _STR(m_x)
-#endif
 
 GODOT_ICON ICON platform/windows/godot_console.ico
 
 1 VERSIONINFO
-FILEVERSION    	VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
-PRODUCTVERSION 	VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
-FILEOS         	4
-FILETYPE       	1
+FILEVERSION     VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
+PRODUCTVERSION  VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
+FILEOS          4
+FILETYPE        1
 BEGIN
     BLOCK "StringFileInfo"
     BEGIN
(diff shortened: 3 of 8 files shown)


platform/windows/godot_res.rc
```diff
@@ -1,16 +1,12 @@
...
 
 1 VERSIONINFO
-FILEVERSION    	VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
-PRODUCTVERSION 	VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
-FILEOS         	4
-FILETYPE       	1
+FILEVERSION     VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
+PRODUCTVERSION  VERSION_MAJOR,VERSION_MINOR,VERSION_PATCH,0
+FILEOS          4
+FILETYPE        1
```
akien-mga, line 9: I just removed weird tabs mixed with spaces and replaced them by a space.

platform/windows/godot_res.rc
```diff
@@ -21,7 +17,7 @@ BEGIN
             VALUE "FileVersion",            VERSION_NUMBER
             VALUE "ProductName",            VERSION_NAME
             VALUE "Licence",                "MIT"
-            VALUE "LegalCopyright",         "Copyright (c) 2007-" _MKSTR(VERSION_YEAR) " Juan Linietsky, Ariel Manzur and contributors"
+            VALUE "LegalCopyright",         "Copyright (c) 2007-present Juan Linietsky, Ariel Manzur and Godot Engine contributors"
```
akien-mga, line 20: I wondered how the copyright line appears on Windows, and whether the "Copyright (c)" part would be redundant with Windows' own labels, or whether it needs to be kept short. Could be good if someone can check on a Windows build.

bruvzg: It is redundant, but it's common way to do it.

yyy

akien-mga: I shortened it a bit by removing the "Copyright" word, but keeping "(c)".

core/version.h
```diff
@@ -31,6 +31,7 @@
 #ifndef VERSION_H
 #define VERSION_H
 
+#include "core/typedefs.h"
```
akien-mga, line 34: Needed for `_MKSTR` used in that file. It just happened to always compile in other files as `typedefs.h` is virtually included everywhere (but in the `.rc` files).

thirdparty/miniupnpc/src/miniupnpcstrings.h
```diff
@@ -1,9 +1,7 @@
 #ifndef MINIUPNPCSTRINGS_H_INCLUDED
 #define MINIUPNPCSTRINGS_H_INCLUDED
 
-#include "core/version.h"
-
-#define OS_STRING VERSION_NAME "/1.0"
+#define OS_STRING "Godot Engine/1.0"
```
akien-mga, line 4: CC @mhilbrunner @Faless 

The previous code was a good idea in general, and allows tweaking this in Godot forks with a different name. But it was the only thirdparty code reaching into `core/version.h`, and actually `core/version_generated.gen.h`, which is recreated for each commit (includes Git hash) so that means this header and the miniupnpc files using it would always get recompiled. Doesn't seem worth it.

akien-mga: BTW if we want to keep it dynamic, we could instead parse `version.py` from `modules/upnp/SCsub` and pass `OS_STRING` on the command line. But I don't think it's worth the trouble, at least until someone asks for it.

akien-mga: Well it's not as bad as I mentioned actually, `version_generated.gen.h` doesn't contain the commit hash, this one is split to `version_hash.gen.h` which is not included by default in `version.h`.

`version_generated.gen.h` shouldn't change too often:
```
/* THIS FILE IS GENERATED DO NOT EDIT */
#ifndef VERSION_GENERATED_GEN_H
#define VERSION_GENERATED_GEN_H
#define VERSION_SHORT_NAME "godot"
#define VERSION_NAME "Godot Engine"
#define VERSION_MAJOR 4
#define VERSION_MINOR 3
#define VERSION_PATCH 0
#define VERSION_STATUS "dev"
#define VERSION_BUILD "custom_build"
#define VERSION_MODULE_CONFIG ""
#define VERSION_WEBSITE "https://godotengine.org"
#define VERSION_DOCS_BRANCH "latest"
#define VERSION_DOCS_URL "https://docs.godotengine.org/en/" VERSION_DOCS_BRANCH
#endif // VERSION_GENERATED_GEN_H
```

So I could restore this if we want, though I still feel it's cleaner not to have thirdparty code reach out to our `core` if we can avoid it.

mhilbrunner: No I agree, changing this seems fine.

YuriSizov (approved):
I support the removal, and I think this covers everything that needed to be updated.

platform/windows/godot_res_wrap.rc
```diff
@@ -21,7 +17,7 @@ BEGIN
             VALUE "FileVersion",            VERSION_NUMBER
             VALUE "ProductName",            VERSION_NAME " (Console)"
             VALUE "Licence",                "MIT"
-            VALUE "LegalCopyright",         "Copyright (c) 2007-" _MKSTR(VERSION_YEAR) " Juan Linietsky, Ariel Manzur and contributors"
+            VALUE "LegalCopyright",         "(c) 2007-present Juan Liniestky, Ariel Manzur and Godot Engine contributors"
```
AThousandShips, line 20:
```suggestion
            VALUE "LegalCopyright",         "(c) 2007-present Juan Linietsky, Ariel Manzur and Godot Engine contributors"
```
Unless it's spelled wrong everywhere else?

platform/windows/godot_res.rc
```diff
@@ -21,7 +17,7 @@ BEGIN
             VALUE "FileVersion",            VERSION_NUMBER
             VALUE "ProductName",            VERSION_NAME
             VALUE "Licence",                "MIT"
-            VALUE "LegalCopyright",         "Copyright (c) 2007-" _MKSTR(VERSION_YEAR) " Juan Linietsky, Ariel Manzur and contributors"
+            VALUE "LegalCopyright",         "(c) 2007-present Juan Liniestky, Ariel Manzur and Godot Engine contributors"
```
AThousandShips, line 20:
```suggestion
            VALUE "LegalCopyright",         "(c) 2007-present Juan Linietsky, Ariel Manzur and Godot Engine contributors"
```

platform/windows/godot_res_wrap.rc
akien-mga: Oops.

akien-mga: Fixed, thanks for catching that.

AThousandShips (approved):

platform/windows/godot_res_wrap.rc
akien-mga: I toyed with the idea of just having `(c) 2007-present Godot Engine contributors` to keep it short, but then changed my mind and typed the names back :)

mhilbrunner (approved):

YuriSizov:
Thanks!

PR #87546, opened 2024-01-24
OpenXR: Make it possible to implement vendor extensions to hand tracking from GDExtension

dsnopek:
There are a number of OpenXR vendor extensions for hand tracking that need to be able to add structures to the `next` pointer chain on the `XrHandJointLocationEXT` structure, including:

- [XR_FB_hand_tracking_aim](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_hand_tracking_aim)
- [XR_FB_hand_tracking_capsule](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_hand_tracking_capsules)
- [XR_FB_hand_tracking_mesh](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_hand_tracking_mesh)

As vendor extensions, these should be implemented in [godot_openxr_vendors](https://github.com/GodotVR/godot_openxr_vendors) via GDExtension, however, there isn't presently a way to add something to this particular `next` pointer chain from GDExtension.

This PR aims to fix that!

The API here is a little awkward, in that it adds a `_set_hand_joint_locations_and_get_next_pointer()` method to `OpenXRExtensionWrapper`, where hand tracking is itself implemented as an `OpenXRExtensionWrapper`. However, I'm not sure there's a way to do it that doesn't have worse downsides, especially for GDExtension. So, in my opinion, this is an acceptable awkwardness. :-)

Also, don't worry about having to loop over all the extension wrappers. That bit is only run when setting up each "hand tracker" for the first time. Once setup, it'll keep writing data to the structures we provide.

Anyway, please let me know what you think!

### DIFF
--- a/modules/openxr/doc_classes/OpenXRExtensionWrapperExtension.xml
+++ b/modules/openxr/doc_classes/OpenXRExtensionWrapperExtension.xml
@@ -123,6 +123,14 @@
 				Called when the OpenXR session state is changed to visible. This means OpenXR is now ready to receive frames.
 			</description>
 		</method>
+		<method name="_set_hand_joint_locations_and_get_next_pointer" qualifiers="virtual">
+			<return type="int" />
+			<param index="0" name="hand_index" type="int" />
+			<param index="1" name="next_pointer" type="void*" />
+			<description>
+				Adds additional data structures when each hand tracker is created.
+			</description>
+		</method>
 		<method name="_set_instance_create_info_and_get_next_pointer" qualifiers="virtual">
 			<return type="int" />
 			<param index="0" name="next_pointer" type="void*" />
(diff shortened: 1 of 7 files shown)


BastiaanOlij (approved):
LGTM

My only remark seeing we're getting a lot of these is whether we should introduce a more generic method that takes a type enum?

Though I think there is a lot to say about writing this out fully as we do now seeing we can properly detail the use case in the class help.

This is a discussion for later though :)

BastiaanOlij:
> The API here is a little awkward, in that it adds a `_set_hand_joint_locations_and_get_next_pointer()` method to `OpenXRExtensionWrapper`, where hand tracking is itself implemented as an `OpenXRExtensionWrapper`. However, I'm not sure there's a way to do it that doesn't have worse downsides, especially for GDExtension. So, in my opinion, this is an acceptable awkwardness. :-)

Hand tracking is a core extension, so I don't see a big problem with this. The extension wrappers are so we can organise the code and not end up with a big blob of code in a single API class. I indeed can't think of a nicer way to do this.

akien-mga:
Thanks!

PR #87550, opened 2024-01-24
C#: Fix sorting for generic types when reloading assemblies.

### LINKED ISSUE #78513: .NET: Failed to unload assemblies. Please check <this issue> for more information.
### Godot version

Any 4.x version

### System information

any

### Issue description

Assembly reloading can fail for various reasons, usually because a library used in tools code is not compatible with assembly unloading.

After unloading has failed, C# scripts will be unavailable until the editor is restarted (in rare cases it may be possible to complete the unloading by re-building assemblies after some time).

If assembly unloading fails for your project check [Microsoft's troubleshooting instructions](https://learn.microsoft.com/en-us/dotnet/standard/assembly/unloadability#troubleshoot-unloadability-issues) and ensure that you are not using one of the libraries known to be incompatible:
- https://github.com/JamesNK/Newtonsoft.Json/issues/2414

If you know of additional libraries that cause issues, please leave a comment.
If your code doesn't use any libraries, doesn't violate any guidelines and you believe unloading is blocked by godot, please open a new issue.

### Steps to reproduce

various

```csharp
using Godot;
using System;

[Tool]
public partial class UnloadingIssuesSample : Node
{
	public override void _Ready()
	{
		// block unloading with a strong handle
		var handle = System.Runtime.InteropServices.GCHandle.Alloc(this);

		// register cleanup code to prevent unloading issues
		System.Runtime.Loader.AssemblyLoadContext.GetLoadContext(System.Reflection.Assembly.GetExecutingAssembly()).Unloading += alc =>
		{
			// handle.Free();
		};
	}
}
```

### LINKED ISSUE #79519: C# Script Registration that use Generics can error out in the ScriptManagerBridge

zaevi:
Fixes (partially) #78513 that caused by generic type.

When reloading assemblies, base types are reloaded first, followed by inherited types. But with generic types, the sorting function doesn't work properly, resulting in the inherited type being loaded before the base type, which creates duplicate CSharpScript by accident.

The sorting algorithm (`CSharpScriptDepSort`) uses `CSharpScript::get_base_script`, but the script for generic type do not have a `resource_path`, so it returns null:

https://github.com/godotengine/godot/blob/4b6ad349886288405890b07d4a8da425eb3c97ec/modules/mono/csharp_script.cpp#L2779-L2781

This PR makes minor changes to fix fail reloading by generics, and developers can use generic types in editor basically. To completely support it, there is still work to do, like adding virtual path for generic `CSharpScript`.

- Close #79519 
- Supersede #83204 
- Supersede #86305 

_Thanks to @Delsin-Yu for helping with debugging._

### DIFF
--- a/modules/mono/csharp_script.cpp
+++ b/modules/mono/csharp_script.cpp
@@ -697,18 +697,24 @@ struct CSharpScriptDepSort {
 			// Shouldn't happen but just in case...
 			return false;
 		}
-		const Script *I = B->get_base_script().ptr();
+		const CSharpScript *I = get_base_script(B.ptr()).ptr();
 		while (I) {
 			if (I == A.ptr()) {
 				// A is a base of B
 				return true;
 			}
 
-			I = I->get_base_script().ptr();
+			I = get_base_script(I).ptr();
 		}
 
 		// A isn't a base of B
 		return false;
+	}
+
+	// Special fix for constructed generic types.
+	Ref<CSharpScript> get_base_script(const CSharpScript *p_script) const {
+		Ref<CSharpScript> base_script = p_script->base_script;
+		return base_script.is_valid() && !base_script->class_name.is_empty() ? base_script : nullptr;
 	}
 };
 
--- a/modules/mono/csharp_script.h
+++ b/modules/mono/csharp_script.h
@@ -60,6 +60,7 @@ class CSharpScript : public Script {
 
 	friend class CSharpInstance;
 	friend class CSharpLanguage;
+	friend struct CSharpScriptDepSort;
 
 	bool tool = false;
 	bool global_class = false;


zaevi:
MRP: [GenericTest.zip](https://github.com/godotengine/godot/files/14043903/GenericTest.zip)

Steps to reproduce: Open the project, modify any C# script, and build.

Delsin-Yu:
After successfully performing an Assembly Unloading, it is crucial for the Native side to reload the list of unloaded scripts in a specific order.

> For example: a `ChildClass` must have its `BaseClass` reloaded before getting passed into the C# reloading method, otherwise, since after unloading, there is no way for the C# side to recognize this new `BaseClass`, it will create and tie a new instance of `BaseClass` back to the Native side, this will eventually causing `Duplicate Key Exception` when the actual pending `BaseClass` is reloading.

[CSharpScriptDepSort ](https://github.com/godotengine/godot/blob/master/modules/mono/csharp_script.cpp#L693) contains the algorithm for sorting these unload-pending scripts, this method relies on [CSharpScript::get_base_script()](https://github.com/godotengine/godot/blob/master/modules/mono/csharp_script.cpp#L2779) for obtaining the base script of a CSharp script.
Unfortunately, for generic inheritances, the base type for a `GenericChildClass : GenericBaseClass<int>` is not GenericBaseClass<T> which we have a path for, instead the actual base type is GenericBaseClass<int> that never has a script_path, so the `get_base_script()` method always returns `null` for `GenericChildClass` in this case, hence it breaks the sorting method, causing unpredictable script reloading order, and we eventually get `Duplicate Key Exception` on the Managed side if that random order happens to be incorrect.

This PR adds a new `get_base_script` dedicated to this sorting algorithm, which has everything the original `algorithm` does, except it replaces the script_path empty checking with class_name empty checking, it should fix the script ordering issue.

However, `GenericBaseClass<int>` is a valid type, so the public API `get_base_script` should return the actual CSharpScript instance, instead of a null value, but assigning path to these in-between types proves challenging, we cannot duplicate the script_path property of the original script `GenericBaseClass<T>` since `Godot resources don't allow multiple resources sharing the same path`, so a separate issue, pr is required on implementing support for some sorts of virtual paths, such as `res://path/to/GenericBaseClass.cs: Int32`, but that is out of the scope of the current PR.

Delsin-Yu:
This doesn't look correct.
image

raulsntos (approved):
LGTM. Thank you so much to everyone involved for helping diagnostic and fix the issue!

akien-mga:
Thanks!

Issue #86731, opened 2024-01-03
shader compilation very slow in Compatibility renderer


djrain:
### Tested versions

- Reproducible in: 4.2.1 stable

### System information

macOS Ventura, Android 13 - compatibility

### Issue description

Showing/loading CanvasItem shaders for the first time is much slower when using Compatibility renderer compared to Forward+ or Mobile. It occurs on macOS (2020 Mac Mini M1) and Android (Google Pixel 4 XL), but not on my iPhone 14 Pro Max (compatibility actually seems to be faster there). Unable to test Windows.

Here it is on mobile mode on Mac, a short stutter

https://github.com/godotengine/godot/assets/33777501/e2f09a1e-58a6-497a-b4ce-455d0a856278

and compatibility, much longer stutter

https://github.com/godotengine/godot/assets/33777501/bec8d6ba-69fa-4755-8f85-cea6f71c3f7c

### Steps to reproduce

Run MRP on Mac or Android. Shaders will load and show after 1 second

### Minimal reproduction project (MRP)

[ShaderCompilationIssue.zip](https://github.com/godotengine/godot/files/13814880/ShaderCompilationIssue.zip)

djrain:
Not sure if it's closely related, but we were just testing our game on iPhone, and in compatibility mode the performance was just worse in general (like 70 fps) and just switching to mobile magically brought it back up to 120. Also having any GPUParticles2D just crashes the game, which I know was happening on mac, but I don't recall this being a problem on iOS before...

djrain:
@clayjohn no rush, but any leads on this? We're launching our mobile game very soon, and we'd really prefer to use Compatibility, but the load times currently are not workable...

clayjohn:
Sorry, no updates from my end. This is a challenging problem as it stems from the difference in drivers on the different platforms. MacOS has very good Metal drivers and bad OpenGL drivers. OpenGL shader compilation on MacOS tends to be slow and there isn't much room to work around it. We have discussed some big changes to the shaders that might help, but it will be significant amount of work, and may not do much. 

Android can be in a similar boat. Some devices have really good OpenGL support and others don't. So from a sample size of 1, its tough to say whether the problem is with Godot or with drivers again. Unfortunately, even if the problem is Godot, the solution might not be the same for every device. Making a change to speed up compiling on one device may slow down compiling on another device. So the process to make fixes to this has to, by nature, be quite slow and careful. 

[...]

PR #87553, opened 2024-01-24
Significantly improve the speed of shader compilation in compatibility backend

clayjohn:
Partially addresses: https://github.com/godotengine/godot/issues/86731

When we first create shaders in the compatibility backend we compile all shader "modes" with the default set of specializations. The canvas shader has 5 "modes". The specialization with lights enabled is about 60x larger than when lights are disabled. Specializations (other than the default) are only compiled when used. So by disabling lights by default we can compile only the simple version by default and then only compile the expensive shaders when needed. 

On my laptop (Intel® Core™ i7-1165G7, Mesa Intel® Xe Graphics) the shader cache for each canvasitem shader in the MRP totaled about 2.9 mb, with this PR they are ~50 kb and when a light is used in the scene it increases to ~650. 

It is hard to gather accurate measurements, but in this case the performance difference is obvious. It goes from having a noticeable stall to no stall at all. 

**Testing the MRP from https://github.com/godotengine/godot/issues/86731**
No caching:
Before: 374 ms
After: 162 ms (debug build, so might be slightly faster in practice)

Caching:
Before: 359 ms
After: 46 ms (debug build, so might be slightly faster in practice)

On my MacBook Pro (early 2015, intel 15 dual core 2.7 GHz) I also see a significant improvement from about 7ms to about 2.5 ms in the same MRP

**Note: To fully take advantage of this PR, you must clear your shader cache. This approach doesn't break the shader cache so its totally safe to use and backport, but the full benefits are only seen with a clean cache. On that note, non-cached runs are a bit faster with this change.**

CC @djrain

I'd like to leave https://github.com/godotengine/godot/issues/86731 open as I think we can do quite a bit more to bring shader compilation time down. I'm making this PR in the meantime as it makes a huge difference and is super safe.

### DIFF
--- a/drivers/gles3/shaders/canvas.glsl
+++ b/drivers/gles3/shaders/canvas.glsl
@@ -9,7 +9,7 @@ mode_instanced = #define USE_ATTRIBUTES \n#define USE_INSTANCING
 
 #[specializations]
 
-DISABLE_LIGHTING = false
+DISABLE_LIGHTING = true
 USE_RGBA_SHADOWS = false
 SINGLE_INSTANCE = false
 


Calinou:
I've benchmarked starting then exiting project manager with `--quit`. I've also tried the same with a 2D project featuring only a Sprite2D, but couldn't notice a measurable difference. I've removed the shader cache at `~/.cache/nvidia/GLCache/` between runs.

*Project manager always uses OpenGL unless otherwise specified on the command line.*

<details open>

- **OS:** Linux (Fedora 39)
</details>

#### Before

```
  Time (mean ± σ):      2.817 s ±  0.089 s    [User: 1.538 s, System: 0.227 s]
  Range (min … max):    2.703 s …  2.967 s    10 runs
```

#### After

```
  Time (mean ± σ):      2.720 s ±  0.323 s    [User: 1.542 s, System: 0.220 s]
  Range (min … max):    1.827 s …  2.969 s    10 runs
```

clayjohn:
The project manager only uses one shader so I'm not surprised you aren't seeing much of a benefit. 

The MRP I used compiles 30 shaders in a frame which makes it a little easier to see the difference

Alex2782:
[/gles3/shaders/scene.glsl](https://github.com/godotengine/godot/blob/5034478611697358f4e135c69f364ef6d8cd54dc/drivers/gles3/shaders/scene.glsl#L11) There are 5 `DISABLE_` defines in the file, can they also be set to `true` without hesitation?

On `moto e5 play` and `Huawei Mediapad T3 AGS-W09` (both Adreno (TM) 308) there is a strange `glLinkProgram` shader freeze bug. 2D scenes work as soon as the flags are set to `true`. 3D scenes unfortunately not yet.  

----------------------
I performed this measurement on an Android device, for `VERTEX` and `FRAGMENT`
It is a simple 2D/3D project. 4 vertex and 4 fragment shaders are compiled.
(DEBUG, Samsung Tab S7, no cache and no optimizations). 

```cpp
		uint64_t start = OS::get_singleton()->get_ticks_msec();
		glShaderSource(spec.vert_id, 1, &cstr, nullptr);
		glCompileShader(spec.vert_id);
		uint64_t end = OS::get_singleton()->get_ticks_msec();
		print_line("glCompileShader: ", (end-start), " msec, name: ", name, " VERTEX");
```

### CanvasShaderGLES3



- **VERTEX** is compiled up to 5x faster
- **FRAGMENT** is compiled up to 10x faster

<details><summary>Logs CanvasShaderGLES3</summary>

-----------------------------------------

**DISABLE_LIGHTING = true**
```
glCompileShader:  3  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  4  msec, name:  CanvasShaderGLES3  FRAGMENT
glCompileShader:  4  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  5  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  3  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  3  msec, name:  CanvasShaderGLES3  FRAGMENT
glCompileShader:  3  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  4  msec, name:  CanvasShaderGLES3  FRAGMENT
glCompileShader:  3  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  4  msec, name:  CanvasShaderGLES3  FRAGMENT
```

------------------

**DISABLE_LIGHTING = false** 
```
glCompileShader:  15  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  31  msec, name:  CanvasShaderGLES3  FRAGMENT
glCompileShader:  15  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  31  msec, name:  CanvasShaderGLES3  FRAGMENT
glCompileShader:  15  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  31  msec, name:  CanvasShaderGLES3  FRAGMENT
glCompileShader:  14  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  30  msec, name:  CanvasShaderGLES3  FRAGMENT
glCompileShader:  14  msec, name:  CanvasShaderGLES3  VERTEX
glCompileShader:  30  msec, name:  CanvasShaderGLES3  FRAGMENT
```
</details>

-----------------------------------------

### SceneShaderGLES3

 sometimes **up to 6x faster**, I have lowered a lot of settings, with default values it is perhaps even faster.

<details><summary>Logs SceneShaderGLES3 </summary>



-------------------------------------

**DISABLE_ * = true**
```
glCompileShader:  6  msec, name:  SceneShaderGLES3  VERTEX
glCompileShader:  8  msec, name:  SceneShaderGLES3  FRAGMENT
glCompileShader:  5  msec, name:  SceneShaderGLES3  VERTEX
glCompileShader:  8  msec, name:  SceneShaderGLES3  FRAGMENT
glCompileShader:  3  msec, name:  SceneShaderGLES3  VERTEX
glCompileShader:  4  msec, name:  SceneShaderGLES3  FRAGMENT
glCompileShader:  4  msec, name:  SceneShaderGLES3  VERTEX
glCompileShader:  5  msec, name:  SceneShaderGLES3  FRAGMENT
```

**DISABLE_ * = false**
```
glCompileShader:  9  msec, name:  SceneShaderGLES3  VERTEX
glCompileShader:  36  msec, name:  SceneShaderGLES3  FRAGMENT
glCompileShader:  7  msec, name:  SceneShaderGLES3  VERTEX
glCompileShader:  23  msec, name:  SceneShaderGLES3  FRAGMENT
glCompileShader:  5  msec, name:  SceneShaderGLES3  VERTEX
glCompileShader:  5  msec, name:  SceneShaderGLES3  FRAGMENT
glCompileShader:  5  msec, name:  SceneShaderGLES3  VERTEX
glCompileShader:  5  msec, name:  SceneShaderGLES3  FRAGMENT
```
</details>

image

clayjohn:
> [/gles3/shaders/scene.glsl](https://github.com/godotengine/godot/blob/5034478611697358f4e135c69f364ef6d8cd54dc/drivers/gles3/shaders/scene.glsl#L11) There are 5 `DISABLE_` defines in the file, can they also be set to `true` without hesitation?

Potentially, yes. I would like to explore the situation for scene.glsl a little bit more before committing to a decision. The tradeoffs are slightly different than with canvas.glsl. 

The biggest different is that a typical 3D scene is more likely to actually use multiple render modes whereas in canvas.glsl, users often only need one of the render modes. So saving the compile cost for the unused render modes is significant. 

The other difference is that in 3D, lighting is much more likely to be used. So if we disable lights by default, then we compile two things, the unlit version (at load time) and the lit version at run time. Since the lit version is so much more expensive to compile we are more likely to notice stutters during gameplay.

lawnjelly (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

PR #87555, opened 2024-01-24
Replace internal usage of ImageTexture in VideoStreamPlayer for Texture2D

EIREXE:
Currently Godot uses an ImageTexture internally for the video player, this is problematic in any case where an ImageTexture isn't returned by the VideoStreamPlayback (it does indeed look like an oversight, since that returns a normal Texture2D).

I stumbled upon this while working on a compute shader based implementation of YUV conversions using Texture2DRD.

### DIFF
--- a/scene/gui/video_stream_player.cpp
+++ b/scene/gui/video_stream_player.cpp
@@ -31,7 +31,6 @@
 #include "video_stream_player.h"
 
 #include "core/os/os.h"
-#include "scene/resources/image_texture.h"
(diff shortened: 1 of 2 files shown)


KoBeWi:
There's also a forward declaration of ImageTexture in h file.

EIREXE:
> There's also a forward declaration of ImageTexture in h file.

my b

KoBeWi (approved):

akien-mga:
Thanks!

PR #87559, opened 2024-01-25
Add FoldableContainer Node

WhalesState:


scene/gui/foldable_container.cpp
```diff
@@ -0,0 +1,301 @@
...
+			Ref<Texture2D> icon = expanded ? theme_cache.arrow : theme_cache.arrow_collapsed;
+			Ref<StyleBox> title_style;
+			if (is_hovering) {
+				title_style = expanded ? theme_cache.title_hover_style : theme_cache.title_collapsed_hover_style;
+			} else {
+				title_style = expanded ? theme_cache.title_style : theme_cache.title_collapsed_style;
+			}
+
+			if (title_style.is_valid()) {
+				title_size = Point2(get_size().width, title_style->get_minimum_size().height);
```
WhalesState, line 169: Need to initialize and set `title_size = Point2(get_size().width, 0)` before checking if the title_style is valid. I just need to know the case when a `StyleBox` will be invalid! because it's hard to make it null except from code.

scene/gui/foldable_container.cpp
```diff
@@ -0,0 +1,301 @@
...
+				title_pos = title_style->get_offset();
+				title_width = get_size().width - title_style->get_minimum_size().width;
+			}
+
+			if (theme_cache.title_font.is_valid()) {
+				title_pos.y += theme_cache.title_font->get_ascent(theme_cache.title_font_size);
+			}
+
+			if (icon.is_valid()) {
+				int h_separation = CLAMP(theme_cache.h_separation, 0, title_width - icon->get_width() - 1);
```
WhalesState, line 196: we need to make sure that h_separation never fall below 0, title width is `container width - (icon.width + title_panel.min_size + h_separation)`, so we need to set `Container.min_size.width` to `h_separation` + `icon.width if icon is valid` + `title_panel.min_size.width if title_panel is valid`.

scene/gui/foldable_container.h
```diff
@@ -0,0 +1,92 @@
...
+class FoldableContainer : public Container {
+	GDCLASS(FoldableContainer, Container);
+
+private:
+	bool expanded = true;
+	String title;
+	HorizontalAlignment title_alignment = HORIZONTAL_ALIGNMENT_LEFT;
+
+	bool is_hovering = false;
+	Point2 title_size;
```
WhalesState, line 45: this is the title panel size, should be changed to int `title_panel_height`, to prevent confusion with `title_width` which is the title text width which is used in the draw_string funtion in the draw notification.

scene/gui/foldable_container.cpp
```diff
@@ -0,0 +1,301 @@
...
+				}
+				icon->draw(ci, icon_pos);
+			}
+
+			if (theme_cache.title_font.is_valid()) {
+				Color font_color = expanded ? theme_cache.title_font_color : theme_cache.title_collapsed_font_color;
+				if (is_hovering) {
+					font_color = theme_cache.title_hover_font_color;
+				}
+				draw_string(theme_cache.title_font, title_pos, title, title_alignment, title_width, theme_cache.title_font_size, font_color);
```
WhalesState, line 218: `draw_string()` is used temporarily, it should be replaced later after adding RTL text orientation and translation.

scene/gui/foldable_container.cpp
```diff
@@ -0,0 +1,301 @@
...
+	ClassDB::bind_method(D_METHOD("set_expanded", "expanded"), &FoldableContainer::set_expanded);
+	ClassDB::bind_method(D_METHOD("is_expanded"), &FoldableContainer::is_expanded);
+	ClassDB::bind_method(D_METHOD("set_title", "title"), &FoldableContainer::set_title);
+	ClassDB::bind_method(D_METHOD("get_title"), &FoldableContainer::get_title);
+	ClassDB::bind_method(D_METHOD("set_title_alignment", "alignment"), &FoldableContainer::set_title_alignment);
+	ClassDB::bind_method(D_METHOD("get_title_alignment"), &FoldableContainer::get_title_alignment);
+
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "expanded"), "set_expanded", "is_expanded");
+	ADD_PROPERTY(PropertyInfo(Variant::STRING, "title"), "set_title", "get_title");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "title_alignment", PROPERTY_HINT_ENUM, "Left,Center,Right"), "set_title_alignment", "get_title_alignment");
```
WhalesState, line 274: I have ignored fill alignment, since the text already fills the available space automatically, also text clipping happens automatically, but we should follow the `LineEdit` text styling for more advanced text render options.

scene/gui/foldable_container.cpp
```diff
@@ -0,0 +1,301 @@
...
+				} else {
+					title_size.height = MAX(title_size.height, icon->get_height());
+				}
+			}
+
+			Point2 title_pos;
+			int title_width = 0;
+			if (title_style.is_valid()) {
+				title_style->draw(ci, Rect2(Point2(), title_size));
+				set_custom_minimum_size(Size2(0, title_size.height));
```
WhalesState, line 186: we need to find a way to set the custom minimum size outside the draw notification, because it prevents setting it in the editor.

scene/gui/foldable_container.cpp
```diff
@@ -0,0 +1,301 @@
...
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "foldable_container.h"
+
+#include "scene/theme/theme_db.h"
+
+Size2 FoldableContainer::get_minimum_size() const {
+	if (!expanded) {
+		return Size2();
```
WhalesState, line 37: I think returning the `Size2(0, title_size.height)` returns Size2(0, 0), we need to define the node minimum size correctly.

WhalesState:
I have tried to override the theme stylebox with a null value for `TabBar`, and it throws this error only without pausing the game!
` Condition "!p_style.is_valid()" is true.
  <C++ Source>   scene\gui\control.cpp:2833 @ Control::add_theme_style_override()
`
and in `TabBar` it's only checking if the `Tab` icon is valid!
is it safe to use `theme_cache.styleboxes` directly without checking if they are valid?

YuriSizov:
This needs a proposal.

WhalesState:
> This needs a proposal.

Yes, but it's still a work in progress since I'm learning more about `TextLine` class now, after finishing it we can discuss it's usability in a feature proposal, since it can be helpful for game menus and applications UI and the editor.

YuriSizov:
Well, we might reject the proposal because a foldable container can br easily done with a script, for example. So I would rather you not wasted your time unless you know this is desired.

Remember, proposal comes before a PR.

WhalesState:
> Well, we might reject the proposal because a foldable container can be easily done with a script, for example. So I would rather you not wasted your time unless you know this is desired.
> 
> Remember, proposal comes before a PR.

it will have a limitation with themes, this is why i have started it, i do have the same Container written in gdscript, and to customize it you need to create a custom_type, and if the project doesn't have a project_theme in project setting it will be hard to find the theme resource to add the custom_type to it, also you will have to keep the theme styleboxes, constants, colors, fonts as variables, and without a theme changing a stylebox property such as color will not load since they don't have signals to emit so we call queue_redraw() to redraw the styleboxes.

KoBeWi:
> without a theme changing a stylebox property such as color it will not load since they don't have signals to emit to call queue_redraw() to redraw the styleboxes.

Pretty sure you can use `NOTIFICATION_THEME_CHANGED` and connect to the stylebox `changed` signal if the notification alone doesn't suffice.

Mickeon:
If you could _roughly_ fill in the documentation it would be heavily appreciated. Since you are the one that developed the Node in the first place, it should be easier for you to document and note down any additional quirk, as well.
I can take a look at it for additional polish at a later time.

scene/gui/foldable_container.cpp
```diff
@@ -0,0 +1,410 @@
...
+	ClassDB::bind_method(D_METHOD("set_title", "title"), &FoldableContainer::set_title);
+	ClassDB::bind_method(D_METHOD("get_title"), &FoldableContainer::get_title);
+	ClassDB::bind_method(D_METHOD("set_title_alignment", "alignment"), &FoldableContainer::set_title_alignment);
+	ClassDB::bind_method(D_METHOD("get_title_alignment"), &FoldableContainer::get_title_alignment);
+	ClassDB::bind_method(D_METHOD("set_language", "language"), &FoldableContainer::set_language);
+	ClassDB::bind_method(D_METHOD("get_language"), &FoldableContainer::get_language);
+	ClassDB::bind_method(D_METHOD("set_text_direction", "text_direction"), &FoldableContainer::set_text_direction);
+	ClassDB::bind_method(D_METHOD("get_text_direction"), &FoldableContainer::get_text_direction);
+
+	ADD_SIGNAL(MethodInfo("expanded", PropertyInfo(Variant::BOOL, "is_expanded")));
```
AThousandShips, line 372: Can't name the signal and the property the same

WhalesState: ok, I also was wondering about that error. Thank you all.

WhalesState:
Sorry about the noise again, sometimes I forget to commit the changes before doing a force push,

PR #87571, opened 2024-01-25
Fix `FileAccessMemory` off by one error in `eof_reached`

0x0ACB:
A small off by one error in `FileAccessMemory`. With the current implementation `eof_reached` is never true when using eg `get_buffer`. Noticed this when trying to load *.tres files from memory which triggered the warning in `get_buffer` indefinitely.


RandomShaper (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #87575, opened 2024-01-25
CI: Update `mymindstorm/setup-emsdk` to v14, should fix cache folder conflicts

akien-mga:
https://github.com/mymindstorm/setup-emsdk/releases/tag/v14

### DIFF
--- a/.github/workflows/web_builds.yml
+++ b/.github/workflows/web_builds.yml
@@ -40,10 +40,11 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Emscripten latest
-        uses: mymindstorm/setup-emsdk@v12
+        uses: mymindstorm/setup-emsdk@v14
         with:
           version: ${{env.EM_VERSION}}
           actions-cache-folder: ${{env.EM_CACHE_FOLDER}}
+          cache-key: emsdk-${{ matrix.cache-name }}-${{env.GODOT_BASE_BRANCH}}-${{github.ref}}-${{github.sha}}
 
       - name: Verify Emscripten setup
         run: |


YuriSizov:
Oh, that's awesome!

YuriSizov (changes requested):
So this doesn't help. They simply added the workflow name as a prefix to the old naming scheme. But our CI has one entry point workflow. So both builds still have the same name for the cache:

```
🔗 GHA-3.1.39-linux-x64
```

The second-to-run build still throws an error, but for some reason it's just not caught as a problem anymore:

```
Failed to save: Unable to reserve cache with key 🔗 GHA-3.1.39-linux-x64, another job may be creating this cache. More details: Cache already exists. Scope: refs/pull/87575/merge, Key: 🔗 GHA-3.1.39-linux-x64, Version: bb287b8d239dae54b2b69d026bd1284e06aff2037346bb5491e7f80dd3ba6b57
```

So it seems we still need to do something manually to resolve this. Seems like they've added `cache-key` to the configuration options. We should populate it with something fitting. See [here](https://github.com/mymindstorm/setup-emsdk/commit/74881103d0c783af64976423564ba36dd99e8fc1#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6).

.github/workflows/web_builds.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Emscripten latest
-        uses: mymindstorm/setup-emsdk@v12
+        uses: mymindstorm/setup-emsdk@v14
         with:
           version: ${{env.EM_VERSION}}
           actions-cache-folder: ${{env.EM_CACHE_FOLDER}}
```
YuriSizov, line 46:
```suggestion
          actions-cache-folder: ${{env.EM_CACHE_FOLDER}}
          cache-key: ${{ matrix.cache-name }}-emsdk
```

Something like this, perhaps?

.github/workflows/web_builds.yml
```diff
@@ -40,10 +40,11 @@ jobs:
...
+          cache-key: ${{ matrix.cache-name }}-emsdk
```
YuriSizov, line 47: Oops, I forgot this variable only contains a prefix, and we actually add more to it when creating the actual cache name. So this works, but it's still not unique enough for multiple runs:

image

This should match our main cache naming scheme.

```suggestion
          cache-key: ${{ matrix.cache-name }}-${{env.GODOT_BASE_BRANCH}}-${{github.ref}}-${{github.sha}}-emsdk
```

akien-mga: Thanks, changed.

Overall, I wonder why caching is so convoluted. What we'd want typically is to just cache based on the Emscripten version, and that's it. But somehow it keeps trying to recreate the cache each time it's using it, instead of just using the version that's already there?

akien-mga: Hm should the `-emsdk` part come earlier in the cache key? IIRC substrings are removed from right to left to find the closest cache hit. It should maybe be `emsdk-${{ matrix.cache-name }}-${{env.GODOT_BASE_BRANCH}}-${{github.ref}}-${{github.sha}}` instead?

YuriSizov: That's a good point. I guess if it's added at the end, we can accidentally download it when trying to find cache for scons itself.

It's not that substrings are removed, we do it manually with `restore-keys` (and `setup-emsdk` doesn't, AFAICT). But we do indeed risk a collision because the first part is the same. So yeah, should probably add it somewhere earlier, or at the very start. Good catch!

YuriSizov (approved):
LGTM!

YuriSizov:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

PR #68169, opened 2022-11-02
Add user-facing C# unhandled exception reporting.

zorbathut:
I wanted a way to catch unhandled exceptions, because I want to display those in-game and also be able to report them to online bug reporting services. This does part of the job! It doesn't do all the job; this is more of a proof-of-concept that I am happy to parlay into good code.

This allows me to do this during my startup:

```cs
Godot.Logging.UnhandledExceptionReporter += (e) =>
{
    ErrorCollector.GatherEx(e);
};
```

and it works great for my current needs.

It is not perfect.

* This catches C# exceptions. It does not catch engine warnings or errors. I'm going to do that too eventually, just not today.
* This does not provide any mechanism by which errors (or warnings) can be suppressed. Sometimes that's useful!
* It's unclear how to extend this API to provide those; most likely the interface will need significant changes to add those, and, well, breaking changes kinda suck.

I am, to reiterate, happy to work on this more if there's specific guidance on how it should be designed; I'm also happy to put together engine-error and engine-warning reporting too. Let me know if this is interesting and if it's worth continuing!

PR #56014, opened 2021-12-17
Add Breakpad for crash dump generation

hhyyrylainen:
I started working on a simple Breakpad crash reporting integration to Godot related to this proposal: https://github.com/godotengine/godot-proposals/issues/1896

Some current caveats:
- ~~I have not tested this with Godot 4, I originally made this against the 3.4 stable label and cherry picked to this branch (that other branch is here: https://github.com/Revolutionary-Games/godot/tree/crash_dumper_3.x). I only did a very quick compile fix and checked that Godot starts on this branch. There didn't seem to be a working sample project available in the asset library(?) so I couldn't test with a project.~~
- ~~There's some cleanup left to do regarding which files to include and which need to be compiled~~
- ~~My local install of clang-format doesn't work with the options in this repo so I couldn't format the files with it (clang-format failed with code 1: /home/hhyyrylainen/Projects/godot/.clang-format:151:1: error: unknown key 'SpacesInLineCommentPrefix')~~ update to Fedora 35 fixed this by having a newer clang version
- ~~I noticed that Windows crash reporter on destruction doesn't disable itself. seems like nothing else either disables it on shutdown. Is that intended? It's not consistent with Linux. For my own testing I added the line to disable it in the destructor, but I didn't commit that here~~ I added this code
- ~~I haven't yet tested whether the created crashdumps can be decoded with stackwalk (I'm going to work on my build scripts next to make this work, I'm not confident enough with the Godot build process yet to make a PR to the build scripts repo). Also Windows is less tested than Linux~~ Windows crash dumps only work with the special mingw supporting stackwalk, but no special version is required in the Godot repo when building Godot, so this is just a minor inconvenience.
- ~~I don't have a mac to develop on, so~~ support for that needs to be done later

PR #87576, opened 2024-01-25
Add user-facing log interception callback.

### LINKED ISSUE #8964: Add user-facing log interception callback

A mod-heavy exploration/building game.

Error messages cannot be reported to a central server. Mod developers can't see error messages easily. Testers (both third-party and mod) can't see error messages easily. In addition, it's easy to entirely miss an error message.

Allow the user to provide a callback which is called on every log message, after which they can do whatever processing they like on it, including "display it in the main game pane".

Importantly, this should buffer messages on the first frame so that the callback can detect super-early load issues.

I actually have a working implementation at https://github.com/godotengine/godot/pull/87576 so I'm just going to point over there. But the short answer is "a singleton that lets you attach a callback".

No; there's currently no way to retrieve log messages from in-game without engine modifications.

As far as I know, this is impossible to implement as an add-on. It *might* be possible in a C++ module as long as you don't care about messages before you apply the hooks, but I do - intrinsically, if you want to record messages before you apply a callback, you need engine support for that.

zorbathut:
I wanted a way to catch all Godot messages, including errors, so they can be displayed to the user and/or handled in code. (If you're saying "why do you want that", read the next comment!) Not only should this handle all messages *after* the callback is applied, I wanted to handle all messages *before* the callback is applied so that startup warnings or errors are reported as well.

This provides a LogManager singleton that can have a callback attached. It saves messages until Frame 1, meaning that if you apply your callback during Frame 0, it quickly replays its entire saved log for you.

This used to be https://github.com/godotengine/godot/pull/68169 but takes a completely different approach at solving the same underlying issue. For my purposes, this is far better, and it's also supported on all scripting languages.

Personal concerns with this solution:

* ~~It imposes a bit of CPU overhead and memory churn on every log message. I don't think this is a big issue because there's already a ton of that, but this does impose further costs.~~ Essentially solved; it still imposes a function call, std::atomic lookup, and conditional, but I don't think that's a big problem.
* If a project spews a ton of log messages out before the first frame, then never registers a log handler, all the work done to handle that is wasted. It's unclear how this can be handled; it's also probably not a big deal because projects generally shouldn't be outputting that much on the first frame.
* ~~The callback *must* be threadsafe, but will generally work pretty much fine even if it's not threadsafe. This is a perfect recipe for introducing subtle threading issues in people's code. I'm not sure how to fix this - providing the message immediately is important to me, and I haven't come up with a way to coherently single-thread that process without introducing performance issues and long message delays.~~ Solved by splitting the callback into Nonthreadsafe and Buffered versions.
* ~~There is currently no way to attach more than one callback, nor is there a way to remove a callback.~~ Solved.
* ~~If multiple callbacks were supported, the current design would either mean that the *second* callback hook didn't get the full debug log replay, or it would require some further engineering work to support that.~~ Solved.
* If the error is caused by a C# exception, and you're setting a callback in C#, then this doesn't provide access to the actual C# exception object. It could! I don't personally need this functionality but it's a pretty obvious thing to add later if there's demand.
* The whole "pass data as a dictionary of variants" approach is pretty gnarly. It's something used in a lot of places already - see [PhysicsDirectSpaceState3D.get_rest_info](https://docs.godotengine.org/en/stable/classes/class_physicsdirectspacestate3d.html#class-physicsdirectspacestate3d-method-get-rest-info) - but I still don't *like* it. Still, when in Rome.

Things that do in fact need to be fixed before commit:

* ~~The documentation is lacking. (I'll fix this if the design is approved in general.)~~ Solved.

Edit: Related feature proposals:

* https://github.com/godotengine/godot-proposals/issues/1896
* https://github.com/godotengine/godot-proposals/issues/3062
* https://github.com/godotengine/godot-proposals/issues/8964

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/8964*


zorbathut:
> Why do you want to record unhandled exceptions in-game, anyway?

A bunch of answers here.

First, my general experience is that in *every engine*, errors regularly get unnoticed by people. People don't look at the error log unless they already know there's something posted, and people never expect an error. I've seen this happen in Unity, Unreal, Godot, and custom engines; it's just a universal truth and I've given up convincing people to look at the error log on their own, they just don't do it.

So this is being used for a system that pops up an error dialog when an error shows up. It pops up *in the game viewport*, which makes it impossible to ignore (the other effective solution I've seen is a modal popup.) Example:

image

The funny part is that when I originally implemented this, I realized this particular error had been happening for a few days and I hadn't even noticed it *while working on the error reporting system*.

Second, testers need to be aware of errors also, and will often be testing on the fully-built game (otherwise you're just begging for release-only errors, which *will* happen.) They need some visible error log, and that means I need to be able to reliably record errors to show ingame.

Third, stuff like Sentry is very convenient for reporting errors to a central server. But you still need a way to *get* those errors, and right now there's no way besides modifying the engine code. This is a general solution for retrieving all messages so you can do whatever you like with them.

Fourth, I'm setting this system up for mod support. Modders don't have the full game engine tools available either, and they still need access to errors. In addition, *their players* need access to errors. Note that all of this is modeled off Rimworld's Developer Mode, which is considered a critical feature for mod developers:

image

The end result of all of this is that in-game error reporting is really *really* handy.

AThousandShips:
Please open a [feature proposal](https://github.com/godotengine/godot-proposals) to track this feature and the demand/ideas 🙂

core/core_bind.h
```diff
@@ -567,6 +567,28 @@ class EngineDebugger : public Object {
...
+	LogManager();
+	~LogManager();
+
+	static LogManager *get_singleton() { return singleton; }
+
+	void register_log_capture(const Callable &p_callable);
+
+private:
+	friend class ::UserLogManagerLogger;
+	void process(const Dictionary &&p_message);
```
AThousandShips, line 589:
```suggestion
	void process(const Dictionary &p_message);
```
Const move doesn't make sense to me, and `Dictionary` isn't designed with move capabilities in mind, we don't use move like this in the engine, and in any case `Dictionary` is ref-counted so it doesn't need to be moved

core/io/logger.h
```diff
@@ -106,4 +110,26 @@ class CompositeLogger : public Logger {
...
+	~UserLogManagerLogger();
+
+	static UserLogManagerLogger *get_singleton() { return singleton; }
+
+	virtual void logv(const char *p_format, va_list p_list, bool p_err) override _PRINTF_FORMAT_ATTRIBUTE_2_0;
+	virtual void log_error(const char *p_function, const char *p_file, int p_line, const char *p_code, const char *p_rationale, bool p_editor_notify, ErrorType p_type = ERR_ERROR) override;
+
+private:
+	friend class core_bind::LogManager;
+	void process(const Dictionary &&p_message);
```
AThousandShips, line 131:
```suggestion
	void process(const Dictionary &p_message);
```

farfalk:
@zorbathut thank you for the ping! I agree with everything you wrote. Can't speak for the code unfortunately. Will try to understand how to test this PR, it's a first for me, but i think this is **very** important, so I'll try to put in the effort


@AThousandShips 
> Please open a [feature proposal](https://github.com/godotengine/godot-proposals) to track this feature and the demand/ideas 🙂

There are already some existing proposals afaik: a possible candidate is https://github.com/godotengine/godot-proposals/issues/1896 or maybe https://github.com/godotengine/godot-proposals/issues/3062

AThousandShips:
Would have noticed them if they were listed in the original PR 🙂, or if OP had linked to them instead

farfalk:
> Would have noticed them if they were listed in the original PR 🙂, or if OP had linked to them instead

No worries! Luckily we're a nice community, so we can compensate for each other's shortcomings 💪

zorbathut:
I did look through the proposals for anything that matched but didn't see it through skimming - unfortunately there aren't great keywords, "log" gives a huge number of results and I guess I didn't choose any better keywords. But thanks for the links, though, at least *someone* managed to find them :D

core/core_bind.h
zorbathut: Fixed!

core/io/logger.h
zorbathut: Fixed!

farfalk:
> I did look through the proposals for anything that matched but didn't see it through skimming - unfortunately there aren't great keywords, "log" gives a huge number of results and I guess I didn't choose any better keywords. But thanks for the links, though, at least *someone* managed to find them :D

Yeah well, i've been lurking around this specific logging issue for a while, so I remember the most promising proposals and PRs by heart 😅

Maybe could be useful to cross-reference this with some or all the proposals I linked above, with the old PR and with some other related famous ones, like https://github.com/godotengine/godot/pull/56014

zorbathut:
Proposals linked in the intro :)

As much as I'd love to be able to handle stuff like Breakpad, this kinda doesn't - it just handles errors, not crashes, and true crash handling is tough because the environment is - sort of by definition - in a broken state. Trying to enter the .NET runtime might just crash it further. I think practically that would have to be a C++ interface just to minimize the cross-contamination with the rest of the code.

This is also something I might do someday but it may take a while :)

Delsin-Yu:
Just want to add to this, by having an in-game log pad / in-game message capture mechanism allows developers to early-report these errors, before they actually find some part of the functionality is broken and it is too late to pause for debugging the remote scene tree.

zorbathut:
Just to update on some brainstorming . . .

> The callback must be threadsafe, but will generally work pretty much fine even if it's not threadsafe. This is a perfect recipe for introducing subtle threading issues in people's code. I'm not sure how to fix this - providing the message immediately is important to me, and I haven't come up with a way to coherently single-thread that process without introducing performance issues and long message delays.

. . . my current best solution is to provide two ways of registering a callback, one Buffered and one Nonthreadsafe. Either as two separate functions or as one function with a two-value enum as a parameter. If it's registered as Buffered, it just stores up messages until they're dispatched all together in the main thread, if it's registered as Nonthreadsafe it has the current behavior.

In theory, this means people will look at the two options and say "oh, a decision! I should understand this decision" instead of either finding themselves annoyed by buffered-by-default or ending up with bizarre crashes thanks to unintentionally choosing nonthreadsafe-by-default.

Better suggestions appreciated, but this seems like a solution that is at least neither actively dangerous nor actively insufficient.

zorbathut:
It will not surprise me at all if I'm missing stuff :)

> add_error_handler

This one seems to function as a fallback for printing errors if the OS singleton doesn't yet exist. Right now I'm starting right after the OS singleton is initialized, so it's not necessary, but by using this I might be able to buffer even earlier.

I'm tempted to look into this one, but it's also going to be a pain because with the current API I'll have to detach the error_handler event when I attach to the OS singleton in order to avoid getting double-warnings.

> add_print_handler

This actually does route through the OS singleton and would crash if the OS singleton wasn't initialized, so I can be certain I'm not losing anything by not registering this one. However, it's *also* set up to silence itself if `application/run/disable_stdout` is enabled. It's unclear to me if it should - it's arguable that, with that enabled, we should be disabling the stdout but not the part that records logs for internal usage. This would all be a pretty big architectural change, plus require some user intent decisions, and is something I'd probably prefer to leave up to a later change (I'm going to solve this for myself by just not turning that flag on :V) 

> There is a separate mechanism for [the debugger](https://github.com/godotengine/godot/blob/5034478611697358f4e135c69f364ef6d8cd54dc/core/debugger/engine_debugger.h#L119-L120), although this is probably less relevant since it is only applicable in debug builds. Finally there is [EditorToaster](https://github.com/godotengine/godot/blob/5034478611697358f4e135c69f364ef6d8cd54dc/editor/gui/editor_toaster.h#L114-L116) (there is currently no way to send a toast in scripting, nor to add a handler). It is not necessary to add all at once, but it is worth considering in terms of compatibility/consistency for follow-up PRs.

Yep, that sounds reasonable to me too. I kinda-sorta feel like some more unified Single Way To Do Logging might be the right approach, but that's a much heavier-weight refactor than I am currently up for.

That might be the appropriate time to solve the `add_error_handler` issue, though.

One nice part about the design I currently have is that all of this would work without API changes; `type` is just a string, so if we end up adding `"debugger"` and `"editorToast"` types, that's totally fine. And extending the buffer earlier won't cause problems, it's all just juggling stuff behind the scenes.

AThousandShips:
I second the callback thing, was going to suggest this myself, at least to explore the option of adding support for registering a custom callback to the existing logging output, it catches everything that currently exists with various data provided

zorbathut:
Alright, so it looks like the TODO list is:

* Better documentation
* Make the log system shut down more thoroughly if nothing is hooked and it's after frame 0
* Split the callback into Nonthreadsafe and Buffered, implement Buffered

A lot of the other potential improvements (start recording earlier, record more types of message) wouldn't involve breaking backwards compatibility and can be done later.

Any way I can attract someone with the authority to say "yes, assuming you do this competently we would be happy to integrate it"? Or to provide a different set of changes that need to be made?

zorbathut:
Big update time.

I've made the proposed changes and fleshed out the documentation. I also rebased the entire thing because the history was frankly misleading; if anyone wants to see the original, it's in https://github.com/zorbathut/godot/commits/pr_loghook_old/ .

The threading work is finicky, because threading work is always finicky. I've done my best to document everything, especially the subtle things.

As far as I am concerned, this is ready to go (pending review). Note that I considered the last two revisions *not* ready to go. This revision is ready to go!

. . . admittedly I rebased it without redoing the tests, though, so we'll see if I regret that.

Swarkin:
> If a project spews a ton of log messages out before the first frame, then never registers a log handler, all the work done to handle that is wasted. It's unclear how this can be handled; it's also probably not a big deal because projects generally shouldn't be outputting that much on the first frame.

How noticeable is this if at all? Will it impact `--headless` or Server apps?

core/io/logger.cpp
```diff
@@ -266,3 +267,354 @@ CompositeLogger::~CompositeLogger() {
...
+		SWAP(buffered_logs_mirror, buffered_logs);
+	}
+
+	// Dispatch to all the buffered callables, in a threadsafe manner
+	// Any buffered callables that disable themselves stop getting messages ASAP
+	// any buffered callables that get attached might start getting messages midway through
+	// we're ok with that, it's still a chronologically coherent block
+	for (int i = 0; i < buffered_logs_mirror.size(); ++i) {
+		// This is the same index-as-iterator-to-avoid-lock-issues dance we do in register_log_capture_nonthreadsafe()
+		int indexToSendTo = 0;
```
Calinou, line 472: Please use snake_case naming for variables as per the [code style guidelines](https://docs.godotengine.org/en/stable/contributing/development/code_style_guidelines.html).

core/io/logger.cpp
```diff
@@ -266,3 +267,354 @@ CompositeLogger::~CompositeLogger() {
...
+		if (state == STATE_BUFFERING) {
+			buffered_logs.append(p_message);
+		}
+		captures_nonthreadsafe_mirror = captures_nonthreadsafe;
+	}
+
+	// Dispatch to all the nonthreadsafe callables at the moment we added to the buffer
+	// We actually don't have to care about cutesy threadsafety for once because we're working off a local copy of the captures list
+
+	for (int i = 0; i < captures_nonthreadsafe_mirror.size(); ++i) {
```
Calinou, line 524: You may be able to use a for range loop here (`for (... : ...)`). Also, we tend to prefer post-incrementation to pre-incrementation when we can't use `for` range loops.

core/core_bind.cpp
```diff
@@ -1900,4 +1900,61 @@ void EngineDebugger::_bind_methods() {
...
 }
 
+////// LogManager //////
+// This is the client-facing part of the user log hooking system.
+// It's basically just a shunt that passes callbacks to the persistent UserLogManagerLogger singleton.
+
+LogManager *LogManager::singleton = nullptr;
+
+LogManager::LogManager() {
+	ERR_FAIL_COND_MSG(singleton != nullptr, "Somehow created two LogManagers");
```
Calinou, line 1910:
```suggestion
	ERR_FAIL_COND_MSG(singleton != nullptr, "Only one LogManager instance can be created at a time.");
```

core/core_bind.cpp
```diff
@@ -1900,4 +1900,61 @@ void EngineDebugger::_bind_methods() {
...
+
+	singleton = this;
+}
+
+LogManager::~LogManager() {
+	ERR_FAIL_COND_MSG(singleton != this, "LogManager::singleton not correct on exit");
```
Calinou, line 1916:
```suggestion
	ERR_FAIL_COND_MSG(singleton != this, "LogManager::singleton is not correct on exit.");
```

core/core_bind.cpp
```diff
@@ -1900,4 +1900,61 @@ void EngineDebugger::_bind_methods() {
...
+	if (log_manager != nullptr) {
+		log_manager->unregister_log_capture_buffered(p_callable);
+	}
+}
+
+void LogManager::_bind_methods() {
+	ClassDB::bind_method(D_METHOD("register_log_capture_nonthreadsafe", "callable"), &LogManager::register_log_capture_nonthreadsafe);
+	ClassDB::bind_method(D_METHOD("unregister_log_capture_nonthreadsafe", "callable"), &LogManager::unregister_log_capture_nonthreadsafe);
+	ClassDB::bind_method(D_METHOD("register_log_capture_buffered", "callable"), &LogManager::register_log_capture_buffered);
+	ClassDB::bind_method(D_METHOD("unregister_log_capture_buffered", "callable"), &LogManager::unregister_log_capture_buffered);
```
Calinou, line 1957: The name `nonthreadsafe` looks out of place to me. I suggest `non_thread_safe` as we already use `thread_safe` in many other places throughout the engine.

Calinou (changes requested):
Tested locally (rebased on top of `master` fb10e67fefb85af3b8f5abd30db561588d8456c6), it works as expected.

**Testing project:** [test_pr_87576.zip](https://github.com/godotengine/godot/files/14337336/test_pr_87576.zip)

Note that whether `\n` is included in `message.text` will depend on the message type, as this happens automatically with the `info` type but not with `error` and `warning`. This is probably a good idea as this lets you distinguish between `print()` and `printraw()` if it lacks a newline at the end.

Screenshot_20240219_231901

Using `await get_tree().process_frame` **twice** can be used to prevent engine initialization messages from showing up in the log callback if you're calling this in a main scene's `_ready()` function or an autoload.

There are some codestyle changes I suggested but the functionality itself seems to be working pretty well. It's something I wanted for a long time, so I'm glad seeing it finally being implemented in a robust manner :slightly_smiling_face: 

We should test this on mobile/web platforms too to make sure there's nothing unexpected popping up, especially as the web platform can run without support for multiple threads now (to avoid the need for cross-origin isolation headers).

core/io/logger.cpp
zorbathut: Done! I'm honestly surprised the style checks don't catch this one.

core/io/logger.cpp
zorbathut: Done!

core/core_bind.cpp
zorbathut: Done!

zorbathut:
Changes made! Let me know if there's anything  else needed.

> We should test this on mobile/web platforms too to make sure there's nothing unexpected popping up, especially as the web platform can run without support for multiple threads now (to avoid the need for cross-origin isolation headers).

I'm not going to have access to mobile dev for a week, but I'll see if I can get web working.

zorbathut:
Works on web, at least!

If someone else can tackle mobile it'd be appreciated, otherwise, probably Sunday or Monday.

core/io/logger.cpp
zorbathut: It might be https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html - I may take a look at this once I'm back home.

. . . entirely because I know I'm going to screw this up *all the time* if I don't have guardrails.

zorbathut:
Just for the record, I currently have no pending tasks I'm aware of on this; all requested changes and testing have been completed. I'm basically just waiting for reviewers to have the spare time to check it out.

If reviewers are waiting on me to do something, let me know what :)

core/core_bind.cpp
```diff
@@ -1900,4 +1900,61 @@ void EngineDebugger::_bind_methods() {
...
+	ERR_FAIL_COND_MSG(log_manager == nullptr, "log_manager not yet initialized; this shouldn't be possible");
+	if (log_manager != nullptr) {
+		log_manager->unregister_log_capture_non_thread_safe(p_callable);
+	}
+}
+
+void LogManager::register_log_capture_buffered(const Callable &p_callable) {
+	UserLogManagerLogger *log_manager = UserLogManagerLogger::get_singleton();
+	ERR_FAIL_COND_MSG(log_manager == nullptr, "log_manager not yet initialized; this shouldn't be possible");
+	if (log_manager != nullptr) {
```
AThousandShips, line 1940: Not needed

doc/classes/LogManager.xml
```diff
@@ -0,0 +1,58 @@
...
+				Your callback will be called with a single Variant containing a dictionary with fields. It is guaranteed to include:
+				[code]type[/code]: The type of message. Currently "error", "warning", 'info", "script", and "shader".
+				[code]text[/code]: The message text.
+				Optionally, the dictionary may contain:
+				[code]file[/code]: The source file the message originated from.
+				[code]line[/code]: The line number the message originated from.
+				[code]function[/code]: The function the message originated from.
+				[code]rationale[/code]: Further detailed information about the message.
+				If the callback is registered before the first frame, it will immediately replay all messages from the start of the game, in order.
+				This function may be called on the main thread only.
```
AThousandShips, line 27:
```suggestion
				[b]Note:[/b] This function may be called on the main thread only.
```
For all these

core/io/logger.cpp
zorbathut: (gonna close this myself just to clean it up, re-open if there's info you wanted in here)

zorbathut:
(should've done this update a little differently, sorry for the clutter, will have it solved shortly)

AThousandShips:
If you like I can add as suggestions to make it easier :)

core/io/logger.cpp
```diff
@@ -266,3 +267,354 @@ CompositeLogger::~CompositeLogger() {
...
+	message["text"] = buf;
+	message["type"] = "info";
+
+	process(message);
+
+	Memory::free_static(buf);
+}
+
+void UserLogManagerLogger::log_error(const char *p_function, const char *p_file, int p_line, const char *p_code, const char *p_rationale, bool p_editor_notify, ErrorType p_type) {
+	if (state == STATE_OFF) {
```
AThousandShips, line 327:
```suggestion
	if (state.get() == STATE_OFF) {
```

core/io/logger.cpp
```diff
@@ -266,3 +267,354 @@ CompositeLogger::~CompositeLogger() {
...
+// This is the internal user log hooking system, which does all the hard work.
+
+UserLogManagerLogger *UserLogManagerLogger::singleton = nullptr;
+
+UserLogManagerLogger::UserLogManagerLogger() {
+	ERR_FAIL_COND_MSG(singleton != nullptr, "Somehow created two UserLogManagerLoggers.");
+
+	singleton = this;
+
+	prebuffering = true;
```
AThousandShips, line 281:
```suggestion
	prebuffering.set();
```

core/io/logger.cpp
```diff
@@ -266,3 +267,354 @@ CompositeLogger::~CompositeLogger() {
...
+
+	// this is about to get overwritten by `recalculate_state`
+	state = STATE_OFF;
```
AThousandShips, line 284:
```suggestion
	state.set(STATE_OFF);
```

core/io/logger.cpp
```diff
@@ -266,3 +267,354 @@ CompositeLogger::~CompositeLogger() {
...
+	}
+
+	process(message);
+}
+
+void UserLogManagerLogger::register_log_capture_non_thread_safe(const Callable &p_callable) {
+	// It gets *extremely* hard to guarantee the proper semantics if you're allowed to call this from other threads.
+	ERR_FAIL_COND_MSG(!::Thread::is_main_thread(), "This call is forbidden outside the main thread.");
+
+	if (prebuffering) {
```
AThousandShips, line 368:
```suggestion
	if (prebuffering.is_set()) {
```

AThousandShips (commented):
Just some pointers :)

doc/classes/LogManager.xml
zorbathut: Done for all three.

core/core_bind.cpp
zorbathut: Fixed.

doc/classes/LogManager.xml
```diff
@@ -0,0 +1,58 @@
...
+			</description>
+		</method>
+		<method name="register_log_capture_non_thread_safe">
+			<return type="void" />
+			<param index="0" name="callable" type="Callable" />
+			<description>
+				Registers a non_thread_safe callback for the log capture system.
+				Your callback will be called whenever a log message is output. This may happen in any thread or in any situation. [b]Your callback must be fully threadsafe, and it must not block.[/b] It may be called from any thread, including the main thread, even if another callback is currently being called. It may be called twice in parallel. It may be called in scenarios where you are disallowed from modifying Node values or creating Nodes. If you do anything to create a log message from this function, it may be called recursively, and this can potentially result in a stack overflow. Any or all of these things may happen before this register call returns. [b]Treat this function with extreme care.[/b]
```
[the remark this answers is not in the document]
zorbathut: Fixed, along with a few other instances of `threadsafe` in this change.

core/io/logger.cpp
```diff
@@ -266,3 +267,354 @@ CompositeLogger::~CompositeLogger() {
...
+		}
+	} else {
+		// it's not Frame 0, therefore we don't have to mess around with the buffer at all
+		// just lock, register ourselves as a callable, and recalculate the state so we can start capturing messages
+		MutexLock lock(mutex);
+		register_callable(captures_non_thread_safe, p_callable);
+		recalculate_state();
+	}
+}
+void UserLogManagerLogger::unregister_log_capture_non_thread_safe(const Callable &p_callable) {
```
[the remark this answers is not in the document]
zorbathut: Fixed.

core/core_bind.cpp
```diff
@@ -1900,4 +1900,61 @@ void EngineDebugger::_bind_methods() {
...
+		log_manager->register_log_capture_buffered(p_callable);
+	}
+}
+
+void LogManager::unregister_log_capture_buffered(const Callable &p_callable) {
+	UserLogManagerLogger *log_manager = UserLogManagerLogger::get_singleton();
+	ERR_FAIL_COND_MSG(log_manager == nullptr, "log_manager not yet initialized; this shouldn't be possible");
+	if (log_manager != nullptr) {
```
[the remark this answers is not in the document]
zorbathut: Fixed.

core/core_bind.cpp
```diff
@@ -1900,4 +1900,61 @@ void EngineDebugger::_bind_methods() {
...
+LogManager::~LogManager() {
+	ERR_FAIL_COND_MSG(singleton != this, "LogManager::singleton not correct on exit");
+
+	singleton = nullptr;
+}
+
+void LogManager::register_log_capture_non_thread_safe(const Callable &p_callable) {
+	UserLogManagerLogger *log_manager = UserLogManagerLogger::get_singleton();
+	ERR_FAIL_COND_MSG(log_manager == nullptr, "log_manager not yet initialized; this shouldn't be possible");
+	if (log_manager != nullptr) {
```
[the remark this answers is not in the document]
zorbathut: Fixed.

core/core_bind.cpp
```diff
@@ -1900,4 +1900,61 @@ void EngineDebugger::_bind_methods() {
...
+		log_manager->register_log_capture_non_thread_safe(p_callable);
+	}
+}
+
+void LogManager::unregister_log_capture_non_thread_safe(const Callable &p_callable) {
+	UserLogManagerLogger *log_manager = UserLogManagerLogger::get_singleton();
+	ERR_FAIL_COND_MSG(log_manager == nullptr, "log_manager not yet initialized; this shouldn't be possible");
+	if (log_manager != nullptr) {
```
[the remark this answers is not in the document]
zorbathut: Fixed.

zorbathut:
Alright, think that's all the pending change requests again :) Not sure why the automated tests haven't kicked off but I'd guess that's either on Github or on your end.

AThousandShips:
Probably because you made changes in quick succession, unsure, will look

zorbathut:
Status update: I again have no pending tasks and am waiting entirely on reviewers. Ball's in your court! Let me know if it ends up back in mine :)

maxrun-games:
Would love to see this included in the upcoming release, looking forward to it, it plugs a major gap in Godot 4.

Thanks for the work @zorbathut

conde2:
Tested and works great in windows platform. Waiting for it to be in next release.

farfalk:
Is this going to be considered in time for 4.3 stable?

If not, what's blocking it? Are there other logging-related issues or PRs that need to be managed together with this one before it gets the green light?

Just trying to assess some kind of general timeline for the availability of the feature

zorbathut:
> @zorbathut great work. Would be possible to share your code for the Debug Log window?

I looked into extracting my code cleanly and determined that it would unfortunately be really hard to do well.

So I did it badly, it's [available over here](https://github.com/zorbathut/godottools/tree/dev). You *will* need to do some cleanup in order to make it work; if you do that, please send a pull request, thanks :D

zorbathut:
Also, just to reiterate, as far as I know I am waiting entirely on the Godot team. I personally would not expect this to end up in 4.3 at this point, but I'm not the one who gets to decide this.

I'm not sure if there's any way I can solve @Calinou 's requested changes; as far as I know, those are taken care of now?

Calinou (approved):
I took a glance at the code and it looks good to me now. Thanks for thoroughly commenting it :slightly_smiling_face:

Calinou:
> Is this going to be considered in time for 4.3 stable?

We can't guarantee this will make it in for 4.3, although it would be ideal since we also got https://github.com/godotengine/godot/pull/89206 for 4.3.

> But Github is still complaining about them.

GitHub will keep the last review state until the review is superseded by another review (like I just did here), or another organization member dismisses the review.

AThousandShips (approved):

core/io/logger.h
```diff
@@ -37,6 +37,10 @@
 
 #include <stdarg.h>
 
+namespace core_bind {
+class LogManager;
```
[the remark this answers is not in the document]
zorbathut: I think this was a relic from an old version of the code, honestly. Fixed.

core/io/logger.cpp
```diff
@@ -31,6 +31,7 @@
 #include "logger.h"
 
 #include "core/config/project_settings.h"
+#include "core/core_bind.h"
```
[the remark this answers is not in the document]
zorbathut: Another relic from an old version of the code, although I did have to add `thread.h` manually after removing this one.

doc/classes/LogManager.xml
```diff
@@ -0,0 +1,58 @@
...
+	</tutorials>
+	<methods>
+		<method name="register_log_capture_buffered">
+			<return type="void" />
+			<param index="0" name="callable" type="Callable" />
+			<description>
+				Registers a buffered callback for the log capture system.
+				Your callback will occasionally be called when log messages have been output. This is guaranteed to happen on the main thread and preserve message order; however, there may be some delay (at most, one frame) between the message being sent and the callback being called.
+				Your callback will be called with a single Variant containing a dictionary with fields. It is guaranteed to include:
+				[code]type[/code]: The type of message. Currently "error", "warning", 'info", "script", and "shader".
```
AThousandShips, line 19:
```suggestion
				[code]type[/code]: The type of message. Currently "error", "warning", "info", "script", and "shader".
```

farfalk:
@zorbathut the folks at Sentry are pretty pumped about this PR and potential followups. One of their developers should appear in followups, at least as commenters, If I understood correctly.

About this specific PR, quoting a message by Bruno Garcia in https://github.com/getsentry/sentry-godot/issues/1 :

 > _"If the error is caused by a C# exception, and you're setting a callback in C#, then this doesn't provide access to the actual C# exception object. It could! I don't personally need this functionality but it's a pretty obvious thing to add later if there's demand."_
 > 
> **Sentry to be able to integrate properly would need the instance of the exception. So we can, for example, get the right metadata in order to server-side, find debug symbols and look up line numbers and paths. For example.**

I think at this point the "providing access to the actual C# exception object" will be considered in followups? Or is it something feasible in the current PR?

Asking without any expectation or urgency, I really can't estimate the load of the feature (maybe one day I'll study the Godot source code... but not today)

In the meantime, thanks a lot for your work! Can't wait to log everything!

zorbathut:
> I think at this point the "providing access to the actual C# exception object" will be considered in followups? Or is it something feasible in the current PR?

I do not personally want to add more stuff to the PR, partly because I don't want to slow down the process, partly because of *entirely selfish reasons*, specifically that I want this code in place and I don't care about the exception object :V

I don't remember offhand how the exception catching works - I think I had an early version that did this in https://github.com/godotengine/godot/pull/68169 - and it might be necessary to add a parameter to Godot error messages for a LanguageSpecificErrorData or something along those lines, that we could wrap the C# Exception in. Then when passed to Godot we look for the existence of that and slap it into a new field in the error handler.

Happy to talk to @bruno-garcia about this, I will *probably* get to it on my own eventually but uh don't hold your breath on that one I've got a TODO list the size of a giraffe and available free time more comparable to an underweight squirrel.

reduz (changes requested):
The general approach I think is OK, what I would change is that, instead of registering callables which is kinda unintuitive what you get ( or adding information via dictionaries) we do it via Extension API.

your Logger binding is OK, but instead, register objects like this rather than callables.

Here is some mock-up of how I would do it:
```cpp
 // IN H
 class LoggerExtension : public Reference {
 	GDCLASS(LoggerExtension, Reference)
  protected:
  
  
  	enum ErrorType { //need to bind this to script, can be done here.
  	
  	};
  
 	GDVIRTUAL7( _log_error, String, String, int, String, String, bool, ErrorType)
 	
 public:
 
 	virtual void log_error(const char *p_function, const char *p_file, int p_line, const char *p_code, const char *p_rationale, bool p_editor_notify = false, ::ErrorType p_type = ERR_ERROR);

 };
 
 VARIANT_ENUM_CAST( LoggerExtension::ErrorType) 
 // in CPP
 
 void LoggerExtension::log_error(const char *p_function, const char *p_file, int p_line, const char *p_code, const char *p_rationale, bool p_editor_notify = false, ::ErrorType p_type = ::ERR_ERROR) {
 
 	GDVIRTUAL_CALL(_log_error,p_function,p_file,p_line,p_code,p_rationale, p_editor_notify,ErrorType(p_type));
}

 void LoggerExtension::_bind_methods() {
	GDVIRTUAL_BIND(_log_error, "function", "file", "line", "code", "rationale", "editor_notify", "error_type")
	BIND_ENUM_CONSTANT( error types ) 
 }
  

```

zorbathut:
> The general approach I think is OK, what I would change is that, instead of registering callables which is kinda unintuitive what you get ( or adding information via dictionaries) we do it via Extension API.

So, a few big downsides here.

First, this makes it impossible to provide optional parameters. We don't actually have all these parameters in all cases, so either we need to provide default values ("line will be filled if it exists, otherwise it will be -1") or we need to change them to some kind of Optional<> type. I'm not sure there *is* an Optional right now (there probably should be!)

Second, this makes it considerably harder to add new parameters in the future. Even in this thread we're talking about adding an Exception object; here, we either have to break everyone's code to add a new parameter, or add an entire new set of callbacks.

Third, I admit I don't see the advantage. The Callable infrastructure already allows a callable on a member of an object; why force an entire object derivation? In C# I can just do this:

```cs
Godot.LogManager.RegisterLogCaptureNonthreadsafe(Callable.From((Variant var) => {
    // code here
}));
```

and I admit I don't see a benefit to making it bigger than that.

Fourth, *the entire point of Callable is to use for callbacks*, [which is exactly what this is](https://docs.godotengine.org/en/stable/classes/class_callable.html):

> Callable is a built-in [Variant](https://docs.godotengine.org/en/stable/classes/class_variant.html#class-variant) type that represents a function. It can either be a method within an [Object](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object) instance, or a standalone function not related to any object, like a lambda function. Like all [Variant](https://docs.godotengine.org/en/stable/classes/class_variant.html#class-variant) types, it can be stored in variables and passed to other functions. It is most commonly used for signal callbacks.

----

I *do* agree the dictionary solution is sorta crummy, but the solution here isn't mandatory inheritance, it's proper `struct` support, and this isn't the first example in Godot of a dictionary being used as an ad-hoc struct.

reduz:
@zorbathut I prefer a class is used like this because its friendlier to statically typed languages (extension languages, C#) and its more intuitive to receive the arguments. Additionally, performance wise its far more efficient to do it like this.

If more parameters are added, this function is left as is and hidden and another one is added, so it will continue working.

zorbathut:
> If more parameters are added, this function is left as is and hidden and another one is added, so it will continue working.

How does this work from the user perspective? Doesn't this either mean "the program stops compiling", "you stop getting some events", or "we're now dispatching events to multiple functions, some of which don't have implementations"?

core/io/logger.cpp
```diff
@@ -266,3 +267,357 @@ CompositeLogger::~CompositeLogger() {
...
+	for (const Callable &callable : captures_non_thread_safe_mirror) {
+		dispatch_message(p_message, callable);
+	}
+}
+
+void UserLogManagerLogger::dispatch_message(const Dictionary &p_message, const Callable &p_callable) {
+	// ideally we should verify that the mutex is *not* held by this thread, but the current API provides no way to do that
+	// (it's possible it's already been scooped up by another thread, that's fine)
+
+	if (p_callable.is_null()) {
```
AThousandShips, line 536:
```suggestion
	if (!p_callable.is_valid()) {
```

AThousandShips (dismissed):
Missed this one earlier, assuming we'd still go with callables, as `is_null` and `is_valid` are not complementary, so this doesn't guarantee it's valid

core/io/logger.cpp
AThousandShips: No, it doesn't check if it's been deleted, unless you mean specifically set to `Callable()`, if the callable has an object that was freed for example 

It's not significantly slower, and ensures things won't crash, it's not correct to check this unless you're certain the object isn't freed

zorbathut: Yes, I mean specifically set to `Callable()`. That's what `unregister_callable()` does.

I'm surprised `callp()` doesn't protect against this, but, sure, I'll change that.

AThousandShips: It does in debug builds, but not otherwise, so it's important to use the correct check

core/io/logger.cpp
```diff
@@ -266,3 +267,362 @@ CompositeLogger::~CompositeLogger() {
...
+
+void UserLogManagerLogger::dispatch_message(const Dictionary &p_message, const Callable &p_callable) {
+	// ideally we should verify that the mutex is *not* held by this thread, but the current API provides no way to do that
+	// (it's possible it's already been scooped up by another thread, that's fine)
+
+	if (!p_callable.is_valid()) {
+		// in most cases, this is going to be a deleted callable, so ignore it;
+		// is_null() would be enough for that, we intentionally invalidate it to null in register_callable
+		//
+		// it's apparently possible for callables to be yanked out from under us, though, and if we call them, we crash
```
AThousandShips, line 540: This comment isn't necessary

zorbathut: I do specifically want to call out "it is expected and normal that it's null, because our removal does that"; it's not otherwise clear that this is standard behavior. Do you have another phrasing you'd recommend?

AThousandShips: No phrasing needed, this is the universal way to check this, it doesn't need explaining here, it's not unusual or unexpected, there's almost no checks against `Callable.is_null` in the codebase, and most remaining ones are invalid and will hopefully be removed soon

This isn't obvious to new users at times, but this isn't the place to clarify this

zorbathut: No, again, the thing I want to clarify isn't "we're checking if the callable is valid here", it's "normal functionality of the removal function replaces callables with null callables and this will happen much more often than you would normally expect callable invalidation to happen".

I'll take the entire comment out if you insist but I think that's removing important information.

AThousandShips: This class is already pretty over annotated, this makes it even more cluttered

But if a comment is necessary just add "Check the callable is valid and hasn't been disconnected", anything more is unnecessary, the code explains itself, it isn't *just* to prevent calling it if it's null, that'll be handled automatically, it's to prevent crashing, but that's something that you don't need to explain here

So anyone reading the code will see that removing the callable sets it to empty, so no confusion is going to happen

Comments should explain things that *aren't* clear from the code, and the clearing is already explained in the other method

AThousandShips: Generally, once approved, this code needs a comment pass to clean up unnecessary comments and their style as well

zorbathut: > This class is already pretty over annotated, this makes it even more cluttered

Intentionally so; it's delicate multithreading work and there's a lot of implicit stuff going on.

> Comments should explain things that aren't clear from the code, and the clearing is already explained in the other method

I think the problem is locality. Sure, the clearing is already explained in the other method, but I can *absolutely* see someone (for example) changing `callp()` to do the `is_valid()` check itself, then looking at every place `callp()` is used and saying "aha, don't need this check anymore!"

If you're looking at A(), and there's an important bit of behavior that you need to know in order to properly modify A(), and that bit of behavior is located in B(), and there's no comment in A() directing you to B(), and there's no other indication that you should go look at B(), then this is - in my opinion - a bug waiting to happen.

> Generally, once approved, this code needs a comment pass to clean up unnecessary comments and their style as well

Is there a way that the reviewer can just submit whatever changes they want? If you want to change the code, you're welcome to just do it.

Again, I think this is a mistake, if you insist I'll change it to whatever you want, but at this point it might just be easier if you made the change instead of me; it'll take less time for you to do it than for you to explain exactly what you want.

AThousandShips: You don't have to change it, it's just my suggestion, but I'll go over this code once it's approved fully and review the comments, and you can apply those changes then

> but I can absolutely see someone (for example) changing callp() to do the is_valid() check itself

That kind of sweeping change won't happen in such a careless way, and anyone doing so would go over the code carefully, no worries, also this wouldn't make any difference here as the check wouldn't be needed in this case either, so it doesn't really apply here

Delsin-Yu:
As a C# developer, I'm fairly concerned about the dictionaries we are passing on here. 
This log hook we are now creating can certainly get frequently called by the engine, passing a dictionary means we will have to allocate managed memory every time there is a new log message created (if the developer hooks it). 
However, I don't see any better alternatives, maybe this dictionary can be optimized after we have proper struct support.

zorbathut:
On some level I'm just not convinced that optimizing this for top performance is really relevant. Compare the cost of allocating a Dictionary or a RefCounted class against the cost of generating the stacktrace; my suspicion is that generating the stacktrace is already going to be *considerably* slower. Compare that against the cost of displaying the text in the Godot editor in debug mode; compare that against the cost of sending the text to VS Code or Rider in debug mode; I'm deeply suspicious that these allocations just aren't that relevant.

It's possible there's going to be someone who needs to output a million *individual* log lines per second for some reason, and I think my answer to them would be "look, you have weird requirements, you should probably not be using the standard debug output system, especially because it's going to be generating a stack trace on every single line, this is designed for debug not for high-throughput data output".

ynot01:
> This actually does route through the OS singleton and would crash if the OS singleton wasn't initialized, so I can be certain I'm not losing anything by not registering this one. However, it's _also_ set up to silence itself if `application/run/disable_stdout` is enabled. It's unclear to me if it should - it's arguable that, with that enabled, we should be disabling the stdout but not the part that records logs for internal usage. This would all be a pretty big architectural change, plus require some user intent decisions, and is something I'd probably prefer to leave up to a later change (I'm going to solve this for myself by just not turning that flag on :V)

Disabling `application/run/disable_stdout` seems like a pretty destructive solution to reduce print overhead for a user that doesn't wish to use this feature, should it not have its own setting?

zorbathut:
The log interception feature is essentially zero-overhead if it's not being used; it's about the same performance hit as adding a config variable would be. The way to disable it is "don't use it".

In an ideal world with infinite refactor time available, I think we'd have a `disable_stdout` flag that disables literally stdout, and a `disable_logging` flag that disables all logging (and therefore also skips stacktraces and formatting), but no need for a specific `disable_log_interception` flag.

(honestly in an *ideal* world we might just have `disable_stdout` and it would implicitly disable all logging if there's no stdout and no log interception, though that might be an excess of cutesy autodetection)

reduz:
Hey! In the end as I am just rehashing large part of the notifications to add also script backtraces, I added custom logging support in my PR (https://github.com/godotengine/godot/pull/91006). My approach is simpler, does not add any extra logging by default and thread safety is up to the implementer of the logger (so we don't need to do this built-in into the engine, which is not normally needed in a lot of logging APIs as they handle their own thread safety).

Let me know if you also need anything in mine that may be missing that this one provides.

zorbathut:
> Let me know if you also need anything in mine that may be missing that this one provides.

Two problems:

* This doesn't report on any problems before the hook is applied, and there are actual error messages that are posted that early. This would skip those entirely, and fixing that hole was one of the major motivations for doing it this way.
* It's nice to say "it's the caller's responsibility", but the vast majority of people writing GDScript are not going to be up to date with threadsafety behavior, and it's only a matter of time before something running in a separate thread posts an error and their script totally blows up. I don't think moving this responsibility to the user is a good idea.

It is simpler, but it's simpler by virtue of not doing things that are actually really important; virtually all the complexity of this solution comes from handling the tough cases.

reduz:
@zorbathut 
> This doesn't report on any problems before the hook is applied, and there are actual error messages that are posted that early. This would skip those entirely, and fixing that hole was one of the major motivations for doing it this way. This does not do what I need.

This is intended, if you want to get such early reporting, write an extension. Those can be hooked into the logger at the very beginning of the engine initialization. I am not going to just save all messages internally "just in case" someone adds a custom logger because 99.999% of times It's not going to happen.

> It's nice to say "it's the caller's responsibility", but the vast majority of people writing GDScript are not going to be up to date with threadsafety behavior.

That's fine, this is not a feature intended for every user, It's instead meant for extending core engine functionality in the most efficient way possible.

zorbathut:
> I am not going to just save all messages internally "just in case" someone adds a custom logger because 99.999% of times It's not going to happen.

That's not what my patch does, however.

> That's fine, this is not a feature intended for every user, It's instead meant for extending core engine functionality in the most efficient way possible.

I strongly feel that exposing a big subtle threadsafety vulnerability directly to users is not a good idea. This *will* cause hard-to-track-down instability in people's games.

Can't stop you, obviously, but I think it's a mistake.

Repiteo:
Despite #91006 being merged, I don't believe that it supersedes what this PR is trying to achieve, so I'm keeping this open

nonchip:
@Repiteo it seems to expose `OS.add_logger`, how does that differ from the callback suggested here?

KerekesDavid:
@nonchip It's probably a bit off topic to discuss this in the PR and not in the proposal, if you have questions about the rationale you should probably ask follow up stuff there.

Regardless, this PR would allow for grabbing logs/warnings/errors all the way from engine startup. It would make the feature a lot more robust for logging errors from testers' machines for example. Right now, you can't reliably log errors that happen before the callbacks are registered.

Although, now that OS.add_logger is already available, I could see a reduced version of this PR that adds the startup buffer to those callbacks instead, and leaves the thread unsafe part alone (but I think it's a nice qol not to have to think about that stuff most of the time).

zorbathut:
For what it's worth, I've been using this in my own project now for about a year and a half, and it's absolutely found issues for me. Highly recommended! I've got a [full vendor framework including my current patches](https://github.com/zorbathut/dieselhorse_godot_framework) available for anyone who wants it, though it's pretty undocumented.

PR #87579, opened 2024-01-25
[3.x] Update `year` property in `version.py` to 2024

akien-mga:
Seems like we missed this one when changing the copyright statements to use `present` instead of the hardcoded `year`.

And backport other minor improvements from #87543.


YuriSizov (approved):

akien-mga:
Cherry-picked for 3.5.4.

PR #87583, opened 2024-01-25
Modify display and documentation of attenuation for Light3D

milkiq:
When using `OmniLight3D` or `SpotLight3D` in a 3D scene, adjusting the `attenuation` property doesn't behave as expected in the editing panel. The default value of `1.0` doesn't result in linear changes in light intensity over distance, and the modes selected through right-clicking on the property, such as `Zero` and `Ease In-Out`, don't influence the light intensity decay as their names suggest.

This description can be confusing for users. For example, with `Attenuation` set to 1, `Range` set to 5, and the light source positioned at (0, 2.5, 0), the color values at the center point of a plane with no lighting (color: rgb(0.5,0.5,0.5)) and a default white plane at (0,0,0) (color: rgb(1,1,1)) are not the same. It is evident that the decay is not linear.

PixPin_2024-01-25_20-08-36

Upon inspecting the source code, found that Godot uses the following formula to calculate light attenuation in Forward+ mode:

```glsl
float get_omni_attenuation(float distance, float inv_range, float decay) {
    float nd = distance * inv_range;
    nd *= nd;
    nd *= nd; // nd^4
    nd = max(1.0 - nd, 0.0);
    nd *= nd; // nd^2
    return nd * pow(max(distance, 0.0001), -decay);
}
```

Here, `distance` is the distance from the light to the vertex, `inv_range` is the reciprocal of the light's `range` property, and `decay` is the light's `attenuation` property.

The formula can be expressed as:

$$
attenuation = \frac{(1 - (\frac{distance}{range})^4)^2}{distance^{decay}}
$$

When `attenuation` is 1 and `range` is 5, the distance decay function produces the following curve:

image-20240125201324384

Clearly, it is not linear within the range [0, 5].

Conversely, when `attenuation` is 0.5 and `range` is 5, it behaves more like linear decay:

image-20240125201912582

In the editor, the `attenuation` property is displayed using an ease curve:

```c++
ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "omni_attenuation", PROPERTY_HINT_EXP_EASING, "attenuation"), "set_param", "get_param", PARAM_ATTENUATION);
```

The curve code is:

```c++
double Math::ease(double p_x, double p_c) {
	if (p_x < 0) {
		p_x = 0;
	} else if (p_x > 1.0) {
		p_x = 1.0;
	}
	if (p_c > 0) {
		if (p_c < 1.0) {
			return 1.0 - Math::pow(1.0 - p_x, 1.0 / p_c);
		} else {
			return Math::pow(p_x, p_c);
		}
	} else if (p_c < 0) {
		//inout ease

		if (p_x < 0.5) {
			return Math::pow(p_x * 2.0, -p_c) * 0.5;
		} else {
			return (1.0 - Math::pow(1.0 - (p_x - 0.5) * 2.0, -p_c)) * 0.5 + 0.5;
		}
	} else {
		return 0; // no ease (raw)
	}
}
```

This curve code clearly differs from the actual distance attenuation function. Therefore, it is suggested to replace `PROPERTY_HINT_EXP_EASING` in the editor with `PROPERTY_HINT_RANGE` and set the default value to 0. When the default value is 0, the light attenuation is more uniform within the range, and adjusting the range from 5 to 10 significantly increases the coverage:

image-20240125202637904

When set to 1, adjusting the range has almost no effect on the function, and beyond half the range, the light's impact on objects is minimal (even with a range set to 4096, objects beyond 20 units from the light receive less than 1/20th of the illumination), which might not align with the developer's expectation when adjusting the range to cover a larger area:

image-20240125202817845

The issue with `SpotLight3D` is similar to `OmniLight3D` and should be addressed accordingly.

### DIFF
--- a/doc/classes/OmniLight3D.xml
+++ b/doc/classes/OmniLight3D.xml
@@ -14,8 +14,10 @@
 	</tutorials>
 	<members>
 		<member name="omni_attenuation" type="float" setter="set_param" getter="get_param" default="1.0">
-			The light's attenuation (drop-off) curve. A number of presets are available in the [b]Inspector[/b] by right-clicking the curve. Zero and negative values are allowed but can produce unusual effects.
-			[b]Note:[/b] Very high [member omni_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member omni_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member omni_attenuation] value possible for the visuals you're trying to achieve.
+			Controls the distance attenuation function for omnilights.
+			A value of [code]0.0[/code] smoothly attenuates light at the edge of the range. A value of [code]1.0[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
+			[b]Note:[/b] Setting it to [code]1.0[/code] may result in distant objects receiving minimal light, even within range. For example, with a range of [code]4096[/code], an object at [code]100[/code] units receives less than [code]0.1[/code] energy.
+			[b]Note:[/b] Using negative or values higher than [code]10.0[/code] may lead to unexpected results.
 		</member>
 		<member name="omni_range" type="float" setter="set_param" getter="get_param" default="5.0">
 			The light's radius. Note that the effectively lit area may appear to be smaller depending on the [member omni_attenuation] in use. No matter the [member omni_attenuation] in use, the light will never reach anything outside this radius.
--- a/doc/classes/SpotLight3D.xml
+++ b/doc/classes/SpotLight3D.xml
@@ -24,8 +24,10 @@
 			The spotlight's [i]angular[/i] attenuation curve. See also [member spot_attenuation].
 		</member>
 		<member name="spot_attenuation" type="float" setter="set_param" getter="get_param" default="1.0">
-			The spotlight's light energy (drop-off) attenuation curve. A number of presets are available in the [b]Inspector[/b] by right-clicking the curve. Zero and negative values are allowed but can produce unusual effects. See also [member spot_angle_attenuation].
-			[b]Note:[/b] Very high [member spot_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member spot_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member spot_attenuation] value possible for the visuals you're trying to achieve.
+			Controls the distance attenuation function for spotlights.
+			A value of [code]0.0[/code] smoothly attenuates light at the edge of the range. A value of [code]1.0[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
+			[b]Note:[/b] Setting it to [code]1.0[/code] may result in distant objects receiving minimal light, even within range. For example, with a range of [code]4096[/code], an object at [code]100[/code] units receives less than [code]0.1[/code] energy.
+			[b]Note:[/b] Using negative or values higher than [code]10.0[/code] may lead to unexpected results.
 		</member>
 		<member name="spot_range" type="float" setter="set_param" getter="get_param" default="5.0">
 			The maximal range that can be reached by the spotlight. Note that the effectively lit area may appear to be smaller depending on the [member spot_attenuation] in use. No matter the [member spot_attenuation] in use, the light will never reach anything outside this range.
--- a/scene/3d/light_3d.cpp
+++ b/scene/3d/light_3d.cpp
@@ -616,7 +616,7 @@ void OmniLight3D::_bind_methods() {
 
 	ADD_GROUP("Omni", "omni_");
 	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "omni_range", PROPERTY_HINT_RANGE, "0,4096,0.001,or_greater,exp"), "set_param", "get_param", PARAM_RANGE);
-	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "omni_attenuation", PROPERTY_HINT_EXP_EASING, "attenuation"), "set_param", "get_param", PARAM_ATTENUATION);
+	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "omni_attenuation", PROPERTY_HINT_RANGE, "-10,10,0.001,or_greater,or_less"), "set_param", "get_param", PARAM_ATTENUATION);
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "omni_shadow_mode", PROPERTY_HINT_ENUM, "Dual Paraboloid,Cube"), "set_shadow_mode", "get_shadow_mode");
 
 	BIND_ENUM_CONSTANT(SHADOW_DUAL_PARABOLOID);
@@ -649,7 +649,7 @@ PackedStringArray SpotLight3D::get_configuration_warnings() const {
 void SpotLight3D::_bind_methods() {
 	ADD_GROUP("Spot", "spot_");
 	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "spot_range", PROPERTY_HINT_RANGE, "0,4096,0.001,or_greater,exp,suffix:m"), "set_param", "get_param", PARAM_RANGE);
-	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "spot_attenuation", PROPERTY_HINT_EXP_EASING, "attenuation"), "set_param", "get_param", PARAM_ATTENUATION);
+	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "spot_attenuation", PROPERTY_HINT_RANGE, "-10,10,0.01,or_greater,or_less"), "set_param", "get_param", PARAM_ATTENUATION);
 	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "spot_angle", PROPERTY_HINT_RANGE, "0,180,0.01,degrees"), "set_param", "get_param", PARAM_SPOT_ANGLE);
 	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "spot_angle_attenuation", PROPERTY_HINT_EXP_EASING, "attenuation"), "set_param", "get_param", PARAM_SPOT_ATTENUATION);
 }


AThousandShips:
Changing the default value breaks compatibility, it shouldn't be changed unless it fixes some bug (can't see this from the description, if the documentation is wrong *it* should be changed, not the default, as people who pick this original value still want it, and it'll now be different in all their projects)

LiBooks:
> Changing the default value breaks compatibility, it shouldn't be changed unless it fixes some bug (can't see this from the description, if the documentation is wrong _it_ should be changed, not the default, as people who pick this original value still want it, and it'll now be different in all their projects)
> 
> Default values are _not_ stored in scenes, so anyone who picked `1.0` will now suddenly have `0.0`, drastically different

But I believe that it is very necessary to change this default value in the future. The value provided by the default should be applicable to common operations, but under the current default value, even if the range of a point light source is set to 4000 meters, its effective illumination range is less than 20 meters. This is a serious error when combined with the incorrect ease curve provided in the editor. It has misled the development of many engine functions and 3D demos in the past few years.

AThousandShips:
That's for the future though, and needs to be discussed in a proposal I'd say, currently it breaks compatibility, it being ill suited isn't a reason to break compatibility

> It has misled the development of many engine functions and 3D demos in the past few years.

Can you provide a source for this claim?

LiBooks:
> That's for the future though, and needs to be discussed in a proposal I'd say, currently it breaks compatibility, it being ill suited isn't a reason to break compatibility
> 
> > It has misled the development of many engine functions and 3D demos in the past few years.
> 
> Can you provide a source for this claim?

You can test it yourself by creating a point light source and a 3D mesh object. Then drag the light range, and you will find the situation I mentioned. In the past year, about three game developers have mentioned to me the problem of point light source anomalies when creating indoor scenes. They developed based on the attenuation curve displayed in the editor during development, but they could not create an effective point light source that illuminates a wide indoor area. The curve is seriously inconsistent with the actual lighting situation. And when the default value is 1, the light will decay rapidly with distance, which is also inconsistent with the development ideas of most game developers.

AThousandShips:
That's hearsay and anecdotal evidence though 🙃 changing the default value is only really valid when the old value is completely useless and it's certain essentially no one is using it or that no one should be using it

Changing this value needs more than that, I'd suggest opening a [proposal](https://github.com/godotengine/godot-proposals) to discuss this and keep it out of this PR

LiBooks:
> 尽管这是道听途说和轶事证据，但只有当旧值完全无用并且可以肯定基本上没有人在使用它或者没有人应该使用它时，更改默认值才真正有效
> 
> 改变这个值需要的不仅仅是这个，我建议打开一个[提案](https://github.com/godotengine/godot-proposals)来讨论这个并将其排除在这个 PR 之外

ok

milkiq:
So, in this PR, we just modify its display, replacing the ease curve with a range value, and updated the documentation to demonstrate how adjusting attenuation can better suit the user's project. Correct?

> Changing this value needs more than that, I'd suggest opening a [proposal](https://github.com/godotengine/godot-proposals) to discuss this and keep it out of this PR

By the way, if we open a proposal to discuss the distance attenuation function, it might be beneficial to compare how other game engines control light attenuation. In previous attenuation functions, attempting to fit physical lighting effects may not be the best approach. It is more important to enable users to understand and achieve the desired lighting effects conveniently.

AThousandShips:
> So, in this PR, we just modify its display, replacing the ease curve with a range value, and updated the documentation to demonstrate how adjusting attenuation can better suit the user's project. Correct?

I'd say so yes

milkiq:
Changed commit.

clayjohn:
Just jumping in to add some context here. We use a physically accurate inverse square falloff in Godot with a modification to make the intensity drop off to 0 at the end of the range. The `attenuation` property allows you to tweak the falloff if you do not want physically accurate lighting. 

An `attenuation` value of `1` does not and should not indicate a linear falloff. It indicates that no change is made to the default inverse square falloff. 

I think its a good idea to focus our conversation more on what users will find intuitive for editing. 

> By the way, if we open a proposal to discuss the distance attenuation function, it might be beneficial to compare how other game engines control light attenuation. In previous attenuation functions, attempting to fit physical lighting effects may not be the best approach. It is more important to enable users to understand and achieve the desired lighting effects conveniently.

In Godot 3.x, we didn't use the physically-based attenuation by default. So the default curve was the linear falloff (attenuation = 1.0) and it was really confusing for users. They had a extremely hard time to make lights look good and they complained a lot. We changed the default in Godot 4.0 and until now, nobody has complained or even noticed, instead we have a lot of people exclaiming how good lights look by default now. 

As far as I know, all the other major engines have been using a formula similar to ours for many years now.

I think you are probably right and what is confusing is the fact that attenuation is displayed as a curve which seems to indicate that it is the only thing used to calculate the falloff of the light source.

doc/classes/OmniLight3D.xml
```diff
@@ -14,8 +14,8 @@
...
-			The light's attenuation (drop-off) curve. A number of presets are available in the [b]Inspector[/b] by right-clicking the curve. Zero and negative values are allowed but can produce unusual effects.
-			[b]Note:[/b] Very high [member omni_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member omni_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member omni_attenuation] value possible for the visuals you're trying to achieve.
+			The parameter controls the distance attenuation function for omnilights. A value of 0 smoothly attenuates light at the range edges. Setting it to 1 approaches a physical lighting model but may result in distant objects receiving minimal light, even within the range. For instance, with a range of 4096, an object at 100 units receives less than 0.1 energy. A value of 0.5 approximates linear attenuation. Using negative or higher values may lead to unpredictable outcomes.
```
Mickeon, line 17: I cannot speak on factual accuracy but for anything else I may:
- Surround the values with `[code]`;
- No need to start with "The paramenter", plus it's not consistent.
- "For instance" -> "For example".
```suggestion
			Controls the distance attenuation function for omnilights. A value of [code]0[/code] smoothly attenuates light at the range edges. Setting it to [code]1[/code] approaches a physical lighting model but may result in distant objects receiving minimal light, even within range. For example, with a range of [code]4096[/code], an object at [code]100[/code] units receives less than [code]0.1[/code] energy. A value of [code]0.5[/code] approximates linear attenuation. Using negative or higher values may lead to unpredictable outcomes.
```
As for the overall structure of the description, I'm not sure. It is pretty lengthy and could either be shortened or at the very least split onto a new line.

doc/classes/SpotLight3D.xml
```diff
@@ -24,8 +24,8 @@
...
-			[b]Note:[/b] Very high [member spot_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member spot_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member spot_attenuation] value possible for the visuals you're trying to achieve.
+			The parameter controls the distance attenuation function for spotlights. A value of 0 smoothly attenuates light at the range edges. Setting it to 1 approaches a physical lighting model but may result in distant objects receiving minimal light, even within the range. For instance, with a range of 4096, an object at 100 units receives less than 0.1 energy. A value of 0.5 approximates linear attenuation. Using negative or higher values may lead to unpredictable outcomes.
```
Mickeon, line 27: Ditto (in hindsight I should've clicked the "Review" button)

doc/classes/OmniLight3D.xml
AThousandShips: (congrats on, if I'm not mistaken, first "official" review)

milkiq:
I've revised the documentation based on your suggestion. @Mickeon

LiBooks:
> Just jumping in to add some context here. We use a physically accurate inverse square falloff in Godot with a modification to make the intensity drop off to 0 at the end of the range. The `attenuation` property allows you to tweak the falloff if you do not want physically accurate lighting.
> 
> An `attenuation` value of `1` does not and should not indicate a linear falloff. It indicates that no change is made to the default inverse square falloff.
> 
> I think its a good idea to focus our conversation more on what users will find intuitive for editing.
> 
> > By the way, if we open a proposal to discuss the distance attenuation function, it might be beneficial to compare how other game engines control light attenuation. In previous attenuation functions, attempting to fit physical lighting effects may not be the best approach. It is more important to enable users to understand and achieve the desired lighting effects conveniently.
> 
> In Godot 3.x, we didn't use the physically-based attenuation by default. So the default curve was the linear falloff (attenuation = 1.0) and it was really confusing for users. They had a extremely hard time to make lights look good and they complained a lot. We changed the default in Godot 4.0 and until now, nobody has complained or even noticed, instead we have a lot of people exclaiming how good lights look by default now.
> 
> As far as I know, all the other major engines have been using a formula similar to ours for many years now.
> 
> I think you are probably right and what is confusing is the fact that attenuation is displayed as a curve which seems to indicate that it is the only thing used to calculate the falloff of the light source.

Thank you very much for your response. In addition, I would like to mention that currently Godot uses a physical formula for attenuation by default, but the light intensity is measured in non-physical units by default. Therefore, I suggest that the formula used and the units of measurement be unified by default. Thank you again for your response.

doc/classes/OmniLight3D.xml
```diff
@@ -14,8 +14,11 @@
 	</tutorials>
 	<members>
 		<member name="omni_attenuation" type="float" setter="set_param" getter="get_param" default="1.0">
-			The light's attenuation (drop-off) curve. A number of presets are available in the [b]Inspector[/b] by right-clicking the curve. Zero and negative values are allowed but can produce unusual effects.
-			[b]Note:[/b] Very high [member omni_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member omni_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member omni_attenuation] value possible for the visuals you're trying to achieve.
+			Controls the distance attenuation function for omnilights.
+			[b]Note:[/b] A value of [code]0[/code] smoothly attenuates light at the range edges. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```
AThousandShips, line 18:
```suggestion
			[b]Note:[/b] A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```
The edges here would be at 0 and max, as it's circular, so range singular is more correct imo

doc/classes/SpotLight3D.xml
```diff
@@ -24,8 +24,11 @@
...
 		</member>
 		<member name="spot_attenuation" type="float" setter="set_param" getter="get_param" default="1.0">
-			The spotlight's light energy (drop-off) attenuation curve. A number of presets are available in the [b]Inspector[/b] by right-clicking the curve. Zero and negative values are allowed but can produce unusual effects. See also [member spot_angle_attenuation].
-			[b]Note:[/b] Very high [member spot_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member spot_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member spot_attenuation] value possible for the visuals you're trying to achieve.
+			Controls the distance attenuation function for spotlights.
+			[b]Note:[/b] A value of [code]0[/code] smoothly attenuates light at the range edges. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```
AThousandShips, line 28:
```suggestion
			[b]Note:[/b] A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```

milkiq:
You are right👍, I resolved them. @AThousandShips

Mickeon:
I feel bad to nitpick but I feel the need to again, even though I have already gave my two cents on this PR.

milkiq:
@Mickeon @AThousandShips What more can I do to improve this PR?

doc/classes/OmniLight3D.xml
```diff
@@ -14,8 +14,11 @@
 	</tutorials>
 	<members>
 		<member name="omni_attenuation" type="float" setter="set_param" getter="get_param" default="1.0">
-			The light's attenuation (drop-off) curve. A number of presets are available in the [b]Inspector[/b] by right-clicking the curve. Zero and negative values are allowed but can produce unusual effects.
-			[b]Note:[/b] Very high [member omni_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member omni_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member omni_attenuation] value possible for the visuals you're trying to achieve.
+			Controls the distance attenuation function for omnilights.
+			[b]Note:[/b] A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```
Mickeon, line 18: Does not need to be a note, it is describing exactly what this property does, not really a quirk.
```suggestion
			A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```

doc/classes/OmniLight3D.xml
```diff
@@ -14,8 +14,11 @@
...
+			[b]Note:[/b] Setting it to [code]1[/code] may result in distant objects receiving minimal light, even within range. For example, with a range of [code]4096[/code], an object at [code]100[/code] units receives less than [code]0.1[/code] energy.
+			[b]Note:[/b] If your lighting range exceeds 200, it's recommended to use 0; for ranges below 200, consider using 0.5.
+			[b]Note:[/b] Using negative or higher values may lead to unexpected results.
```
Mickeon, line 21: The line below is continuing the train of thought of the above line. They could be one in the same, perhaps. Let's surround
```suggestion
			[b]Note:[/b] If your lighting range exceeds [code]200[/code], it's recommended to use [code]0[/code]. Otherwise, consider using [code]0.5[/code]. Using negative or higher values may lead to unexpected results.
```

doc/classes/SpotLight3D.xml
```diff
@@ -24,8 +24,11 @@
...
+			Controls the distance attenuation function for spotlights.
+			[b]Note:[/b] A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
+			[b]Note:[/b] Setting it to [code]1[/code] may result in distant objects receiving minimal light, even within range. For example, with a range of [code]4096[/code], an object at [code]100[/code] units receives less than [code]0.1[/code] energy.
+			[b]Note:[/b] If your lighting range exceeds 200, it's recommended to use 0; for ranges below 200, consider using 0.5.
+			[b]Note:[/b] Using negative or higher values may lead to unexpected results.
```
Mickeon, line 31: Ditto
```suggestion
			[b]Note:[/b] If your lighting range exceeds [code]200[/code], it's recommended to use [code]0[/code]. Otherwise, consider using [code]0.5[/code]. Using negative or higher values may lead to unexpected results.
```

doc/classes/SpotLight3D.xml
```diff
@@ -24,8 +24,11 @@
...
 		</member>
 		<member name="spot_attenuation" type="float" setter="set_param" getter="get_param" default="1.0">
-			The spotlight's light energy (drop-off) attenuation curve. A number of presets are available in the [b]Inspector[/b] by right-clicking the curve. Zero and negative values are allowed but can produce unusual effects. See also [member spot_angle_attenuation].
-			[b]Note:[/b] Very high [member spot_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member spot_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member spot_attenuation] value possible for the visuals you're trying to achieve.
+			Controls the distance attenuation function for spotlights.
+			[b]Note:[/b] A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```
Mickeon, line 28: And ditto.
```suggestion
			A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```

Mickeon (changes requested):
Thanks for mentioning me. Good stuff, it just may need a few more tweaks.

doc/classes/OmniLight3D.xml
AThousandShips: I'm a bit unsure about "lighting range", I'd say "light range" reads better

doc/classes/SpotLight3D.xml
AThousandShips: Same here

Mickeon (approved):
Documentation looks great, I cannot speak for the PROPERTY_HINT change

scene/3d/light_3d.cpp
```diff
@@ -616,7 +616,7 @@ void OmniLight3D::_bind_methods() {
 
 	ADD_GROUP("Omni", "omni_");
 	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "omni_range", PROPERTY_HINT_RANGE, "0,4096,0.001,or_greater,exp"), "set_param", "get_param", PARAM_RANGE);
-	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "omni_attenuation", PROPERTY_HINT_EXP_EASING, "attenuation"), "set_param", "get_param", PARAM_ATTENUATION);
+	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "omni_attenuation", PROPERTY_HINT_RANGE, "-10,10,0.001"), "set_param", "get_param", PARAM_ATTENUATION);
```
Calinou, line 619: Allow lesser/greater values to preserve existing behavior (and avoid clamping values in existing scenes):

```suggestion
	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "omni_attenuation", PROPERTY_HINT_RANGE, "-10,10,0.001,or_lesser,or_greater"), "set_param", "get_param", PARAM_ATTENUATION);
```

scene/3d/light_3d.cpp
```diff
@@ -649,7 +649,7 @@ PackedStringArray SpotLight3D::get_configuration_warnings() const {
 void SpotLight3D::_bind_methods() {
 	ADD_GROUP("Spot", "spot_");
 	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "spot_range", PROPERTY_HINT_RANGE, "0,4096,0.001,or_greater,exp,suffix:m"), "set_param", "get_param", PARAM_RANGE);
-	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "spot_attenuation", PROPERTY_HINT_EXP_EASING, "attenuation"), "set_param", "get_param", PARAM_ATTENUATION);
+	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "spot_attenuation", PROPERTY_HINT_RANGE, "-10,10,0.01"), "set_param", "get_param", PARAM_ATTENUATION);
```
Calinou, line 652: Allow less/greater values to preserve existing behavior (and avoid clamping values in existing scenes):

```suggestion
	ADD_PROPERTYI(PropertyInfo(Variant::FLOAT, "spot_attenuation", PROPERTY_HINT_RANGE, "-10,10,0.01,or_less,or_greater"), "set_param", "get_param", PARAM_ATTENUATION);
```

doc/classes/OmniLight3D.xml
```diff
@@ -14,8 +14,10 @@
...
+			Controls the distance attenuation function for omnilights.
+			A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
+			[b]Note:[/b] Setting it to [code]1[/code] may result in distant objects receiving minimal light, even within range. For example, with a range of [code]4096[/code], an object at [code]100[/code] units receives less than [code]0.1[/code] energy.
+			[b]Note:[/b] If your light range exceeds [code]200[/code], it's recommended to use [code]0[/code]. Otherwise, consider using [code]0.5[/code]. Using negative or higher values may lead to unexpected results.
```
clayjohn, line 20: I wouldn't include this recommendation. Attenuation is a very style-specific choice. For realistic looking lighting you always want to leave it at 1. Telling users to set it to 0.5 without a reason why is not helpful. 

The above paragraphs are very helpful and IMO communicate everything that we need

doc/classes/SpotLight3D.xml
```diff
@@ -24,8 +24,10 @@
...
+			Controls the distance attenuation function for spotlights.
+			A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
+			[b]Note:[/b] Setting it to [code]1[/code] may result in distant objects receiving minimal light, even within range. For example, with a range of [code]4096[/code], an object at [code]100[/code] units receives less than [code]0.1[/code] energy.
+			[b]Note:[/b] If your light range exceeds [code]200[/code], it's recommended to use [code]0[/code]. Otherwise, consider using [code]0.5[/code]. Using negative or higher values may lead to unexpected results.
```
clayjohn, line 30: Same comment as above

milkiq:
OK. Resolved all.

milkiq:
@AThousandShips @clayjohn @Calinou Is there anything else I need to improve for this PR?

Calinou (approved):
Tested locally, it works as expected. As a bonus, this also makes it easier to select zero/negative attenuation values without having to right-click the property.

**Testing project:** [test_pr_87583.zip](https://github.com/godotengine/godot/files/14173027/test_pr_87583.zip)

### Before

https://github.com/godotengine/godot/assets/180032/7a94c935-4855-4d47-8817-60453b3f43bf

https://github.com/godotengine/godot/assets/180032/387d8da9-c863-43e3-86d4-0ea84fd4661e

### After

https://github.com/godotengine/godot/assets/180032/8adb08be-c3bb-4466-b265-74b55f61171a

https://github.com/godotengine/godot/assets/180032/b4c9166c-e62f-4b1b-a75a-70c6d2e8d9c1

doc/classes/SpotLight3D.xml
```diff
@@ -24,8 +24,10 @@
...
-			[b]Note:[/b] Very high [member spot_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member spot_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member spot_attenuation] value possible for the visuals you're trying to achieve.
+			Controls the distance attenuation function for spotlights.
+			A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
+			[b]Note:[/b] Setting it to [code]1[/code] may result in distant objects receiving minimal light, even within range. For example, with a range of [code]4096[/code], an object at [code]100[/code] units receives less than [code]0.1[/code] energy.
+			[b]Note:[/b] Using negative or higher values may lead to unexpected results.
```
Calinou, line 30:
```suggestion
			[b]Note:[/b] Using negative or values higher than [code]10.0[/code] may lead to unexpected results.
```

doc/classes/OmniLight3D.xml
```diff
@@ -14,8 +14,10 @@
...
-			[b]Note:[/b] Very high [member omni_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member omni_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member omni_attenuation] value possible for the visuals you're trying to achieve.
+			Controls the distance attenuation function for omnilights.
+			A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
+			[b]Note:[/b] Setting it to [code]1[/code] may result in distant objects receiving minimal light, even within range. For example, with a range of [code]4096[/code], an object at [code]100[/code] units receives less than [code]0.1[/code] energy.
+			[b]Note:[/b] Using negative or higher values may lead to unexpected results.
```
Calinou, line 20:
```suggestion
			[b]Note:[/b] Using negative or values higher than [code]10.0[/code] may lead to unexpected results.
```

doc/classes/OmniLight3D.xml
```diff
@@ -14,8 +14,10 @@
 	</tutorials>
 	<members>
 		<member name="omni_attenuation" type="float" setter="set_param" getter="get_param" default="1.0">
-			The light's attenuation (drop-off) curve. A number of presets are available in the [b]Inspector[/b] by right-clicking the curve. Zero and negative values are allowed but can produce unusual effects.
-			[b]Note:[/b] Very high [member omni_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member omni_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member omni_attenuation] value possible for the visuals you're trying to achieve.
+			Controls the distance attenuation function for omnilights.
+			A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```
Calinou, line 18:
```suggestion
			A value of [code]0.0[/code] smoothly attenuates light at the edge of the range. A value of [code]1.0[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```

doc/classes/SpotLight3D.xml
```diff
@@ -24,8 +24,10 @@
...
 		</member>
 		<member name="spot_attenuation" type="float" setter="set_param" getter="get_param" default="1.0">
-			The spotlight's light energy (drop-off) attenuation curve. A number of presets are available in the [b]Inspector[/b] by right-clicking the curve. Zero and negative values are allowed but can produce unusual effects. See also [member spot_angle_attenuation].
-			[b]Note:[/b] Very high [member spot_attenuation] values (typically above 10) can impact performance negatively if the light is made to use a larger [member spot_range] to compensate. This is because culling opportunities will become less common and shading costs will be increased (as the light will cover more pixels on screen while resulting in the same amount of brightness). To improve performance, use the lowest [member spot_attenuation] value possible for the visuals you're trying to achieve.
+			Controls the distance attenuation function for spotlights.
+			A value of [code]0[/code] smoothly attenuates light at the edge of the range. A value of [code]1[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```
Calinou, line 28:
```suggestion
			A value of [code]0.0[/code] smoothly attenuates light at the edge of the range. A value of [code]1.0[/code] approaches a physical lighting model. A value of [code]0.5[/code] approximates linear attenuation.
```

doc/classes/OmniLight3D.xml
milkiq: OK👌

akien-mga:

PR #87584, opened 2024-01-25
Retain meta data set on importer nodes

tamask:
During the import process, ImporterMeshInstance3D nodes are replaced with MeshInstance3D nodes. This change copies over any meta data that might have been set during the import process, for example, in a GLTFDocumentExtension.

This is a resolution for https://github.com/godotengine/godot-proposals/discussions/6586.

### DIFF
--- a/editor/import/3d/resource_importer_scene.cpp
+++ b/editor/import/3d/resource_importer_scene.cpp
@@ -762,6 +762,7 @@ Node *ResourceImporterScene::_pre_fix_node(Node *p_node, Node *p_root, HashMap<R
 					StaticBody3D *col = memnew(StaticBody3D);
 					col->set_transform(mi->get_transform());
 					col->set_name(fixed_name);
+					_copy_meta(p_node, col);
 					p_node->replace_by(col);
 					p_node->set_owner(nullptr);
 					memdelete(p_node);
@@ -776,6 +777,7 @@ Node *ResourceImporterScene::_pre_fix_node(Node *p_node, Node *p_root, HashMap<R
 			StaticBody3D *sb = memnew(StaticBody3D);
 			sb->set_name(fixed_name);
 			Object::cast_to<Node3D>(sb)->set_transform(Object::cast_to<Node3D>(p_node)->get_transform());
+			_copy_meta(p_node, sb);
 			p_node->replace_by(sb);
 			p_node->set_owner(nullptr);
 			memdelete(p_node);
@@ -820,6 +822,7 @@ Node *ResourceImporterScene::_pre_fix_node(Node *p_node, Node *p_root, HashMap<R
 
 			RigidBody3D *rigid_body = memnew(RigidBody3D);
 			rigid_body->set_name(_fixstr(name, "rigid_body"));
+			_copy_meta(p_node, rigid_body);
 			p_node->replace_by(rigid_body);
 			rigid_body->set_transform(mi->get_transform());
 			p_node = rigid_body;
@@ -884,6 +887,7 @@ Node *ResourceImporterScene::_pre_fix_node(Node *p_node, Node *p_root, HashMap<R
 		Ref<NavigationMesh> nmesh = mesh->create_navigation_mesh();
 		nmi->set_navigation_mesh(nmesh);
 		Object::cast_to<Node3D>(nmi)->set_transform(mi->get_transform());
+		_copy_meta(p_node, nmi);
 		p_node->replace_by(nmi);
 		p_node->set_owner(nullptr);
 		memdelete(p_node);
@@ -924,6 +928,7 @@ Node *ResourceImporterScene::_pre_fix_node(Node *p_node, Node *p_root, HashMap<R
 		VehicleBody3D *bv = memnew(VehicleBody3D);
 		String n = _fixstr(p_node->get_name(), "vehicle");
 		bv->set_name(n);
+		_copy_meta(p_node, bv);
 		p_node->replace_by(bv);
 		p_node->set_name(n);
 		bv->add_child(p_node);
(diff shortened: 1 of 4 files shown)


editor/import/3d/resource_importer_scene.cpp
```diff
@@ -2168,6 +2168,16 @@ Node *ResourceImporterScene::_generate_meshes(Node *p_node, const Dictionary &p_
 			} break;
 		}
 
+		{
+			List<StringName> meta_list;
+			p_node->get_meta_list(&meta_list);
+			for (List<StringName>::Element *element = meta_list.front(); element; element = element->next()) {
```
AThousandShips, line 2174:
```suggestion
			for (List<StringName>::Element *E = meta_list.front(); E; E = E->next()) {
```
We normally name it so

tamask: Ah ok, great!

editor/import/3d/resource_importer_scene.cpp
```diff
@@ -2168,6 +2168,16 @@ Node *ResourceImporterScene::_generate_meshes(Node *p_node, const Dictionary &p_
 			} break;
 		}
 
+		{
+			List<StringName> meta_list;
+			p_node->get_meta_list(&meta_list);
+			for (List<StringName>::Element *E = meta_list.front(); E; E = E->next()) {
+				StringName meta_key = element->get();
```
AThousandShips, line 2175:
```suggestion
				StringName meta_key = E->get();
```
This one too, sorry wasn't by my computer or I'd have made it a single suggestion but forgot

tamask:
I've updated this pull request to be more thorough with copying over meta data on various imported nodes, not just ImporterMeshInstance3D.

editor/import/3d/resource_importer_scene.cpp
```diff
@@ -2199,6 +2211,16 @@ void ResourceImporterScene::_add_shapes(Node *p_node, const Vector<Ref<Shape3D>>
 	}
 }
 
+void ResourceImporterScene::_copy_meta(Node *p_src_node, Node *p_dst_node) {
+	List<StringName> meta_list;
+	p_src_node->get_meta_list(&meta_list);
+	for (List<StringName>::Element *E = meta_list.front(); E; E = E->next()) {
+		StringName meta_key = E->get();
```
AThousandShips, line 2218:
```suggestion
	for (const StringName &meta_key : meta_list) {
```
More convenient approach thanks to iterators

tamask: ah, nice, thanks!

tamask:
I've updated the pull request to also copy over metadata from ImporterMesh to ArrayMesh. In addition to ImporterMeshInstance3D, ImporterMesh is the other intermediary class used where metadata set in a GLTFDocumentExtension would be lost. 

Here's an example use case in a GLTFDocumentExtension that copies over the `extras` from the gltf json so that it's available in the packed scene in godot:

```gdscript
func _apply_extras_to_meshes(state: GLTFState) -> void:
    for index: int in range(state.meshes.size()):
        var extras: Dictionary = state.json.meshes[index].get("extras", {})
        var mesh: GLTFMesh = state.meshes[index]
        _apply_extras_to_meta(state, mesh.mesh, extras)
```

This way, custom properties set in blender on the mesh data can be conveyed over to the godot side via gltf.

lyuma:
Looks good and easily digestible so I think it should be safe in 4.3 - Would it be possible to implement the same change in `modules/gltf/extensions/gltf_document_extension_convert_importer_mesh.cpp` ?
Basically, the GLTFDocumentExtension is the corresponding process for runtime GLTFDocument import, since the editor import pipeline code is not enabled at runtime.

https://github.com/godotengine/godot/blob/master/modules/gltf/extensions/gltf_document_extension_convert_importer_mesh.cpp
It's only in one place, so duplicating the contents of `_copy_meta` is acceptable in this case.

tamask:
> Looks good and easily digestible so I think it should be safe in 4.3 - Would it be possible to implement the same change in `modules/gltf/extensions/gltf_document_extension_convert_importer_mesh.cpp` ? Basically, the GLTFDocumentExtension is the corresponding process for runtime GLTFDocument import, since the editor import pipeline code is not enabled at runtime.
> 
> https://github.com/godotengine/godot/blob/master/modules/gltf/extensions/gltf_document_extension_convert_importer_mesh.cpp It's only in one place, so duplicating the contents of `_copy_meta` is acceptable in this case.

Absolutely, I've updated the pull request source branch to reflect this.

lyuma (approved):
This will fix a source of confusion when working with import plugins. Great work!

AThousandShips (approved):

editor/import/3d/resource_importer_scene.cpp
```diff
@@ -2553,6 +2563,7 @@ Node *ResourceImporterScene::_generate_meshes(Node *p_node, const Dictionary &p_
 			}
 
 			if (mesh.is_valid()) {
+				_copy_meta(*(src_mesh_node->get_mesh()), *mesh);
```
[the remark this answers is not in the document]
lyuma: This suggestion seems like it is mixing two separate conventions. Would it not logically follow that all conversions from Ref<Foo> to Foo* should use .ptr(), not only the first one?
```suggestion
				_copy_meta(src_mesh_node->get_mesh().ptr(), mesh.ptr());
```
Actually come to think of it, usually in C++, operator* returns Foo&, but it seems Ref<Foo>::operator* returns Foo*?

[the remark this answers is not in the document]
tamask: I'll update both arguments to use the `.ptr()` convention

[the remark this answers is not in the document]
tamask: Updated the source branch to reflect this suggestion:

https://github.com/cardboardcomputer/godot/blob/e19f0042caddb7d0ceb260e9ba9f5e5ee08a9f44/editor/import/3d/resource_importer_scene.cpp#L2566

https://github.com/cardboardcomputer/godot/blob/e19f0042caddb7d0ceb260e9ba9f5e5ee08a9f44/modules/gltf/extensions/gltf_document_extension_convert_importer_mesh.cpp#L71

akien-mga:

tamask:

Wonderful, and thank you!

PR #87586, opened 2024-01-25
Add custom paths support (custom `res://`/`user://`-like paths)

### LINKED ISSUE #6307: Custom resource path support

Godot

There are several cases where the resource file paths offered by Godot are not flexible enough, and community has complained about this.
* A way to have editor-only assets (for global plugins).
* A way to stream assets from the web or other locations.
* A way to just create custom resource paths for users to interact with sandboxed game data.

Godot currently hardcodes `res://` and `user://`, but users often require more flexibility here.

This feature would allow to add custom resource path providers to the filesystem. Users would need to pass a class and some configuration information. Probably an API like this:

```GDScript
# Create the game:// path using an existing filesystem path
FileAccess.add_resource_path("game://","/home/user/Documents/somepath") 
# Create the assets:// path using an URL
FileAccess.add_resource_path_url("assets://","https://mygame.com/assets")
# Create the database:// path using a custom class that inherits FileAccess and DirAccess.
FileAccess.add_resource_path_class("database://","CustomFileAccessClass","CustomDirAccessClass",userdata)
```
This would allow, as an example, to have global editor plugins. As the editor could register an `editor://` path where the global plugins are saved.

Currently the 'res://' and 'user://' paths are hardcoded, and a lot of places int he codebase check for "res://' manualy to validate resource files. This would need to be cleaned up to look up into a FileAccess function.

Implementing this is not specifically hard, since what determines the base path and the File/DirAccess to use is _mostly_ on `open()`.

N/A

N/A

adamscott:
Capture d’écran du 2024-01-25 09-45-02_2

# tl;dr
Enables custom resource paths that users can setup and use themselves.

```gdscript
# Maps a path to a custom path
FileAccess.map_path_to_custom_path("home", "/home/myself")
# (with or without the ://, it removes it automatically)
FileAccess.map_path_to_custom_path("home://", "/home/myself")
# Map classes (custom `FileAccess` and `DirAccess`) to custom path
FileAccess.map_classes_to_custom_path("cloud://", "CloudFileAccess", "CloudDirAccess")
# Removes a custom path
FileAccess.remove_custom_path("home")
# Checks if the prefix is supported
FileAccess.is_custom_path("home")
# Returns a dictionary containing the definition of the custom path
FileAccess.get_custom_path_data("home")
# Returns an array containing all the custom path prefixes
FileAccess.get_custom_paths()
```

# Todo
- [ ] Write the documentation

# Fixes
- Fix godotengine/godot-proposals#6307
- Closes godotengine/internal-team-priorities#92

### DIFF
--- a/core/config/project_settings.cpp
+++ b/core/config/project_settings.cpp
@@ -268,6 +268,15 @@ String ProjectSettings::globalize_path(const String &p_path) const {
 		return p_path.replace("user://", "");
 	}
 
+	// Custom resource paths
+	Dictionary resource_paths = FileAccess::get_resource_paths();
+	for (int i = 0; i < resource_paths.size(); i++) {
+		String key = resource_paths.get_key_at_index(i);
+		if (p_path.begins_with(key + "://")) {
+			return p_path.replace(key + ":/", resource_paths[key]);
+		}
+	}
+
 	return p_path;
 }
 
--- a/core/config/project_settings.cpp
+++ b/core/config/project_settings.cpp
@@ -270,8 +270,10 @@ String ProjectSettings::globalize_path(const String &p_path) const {
 
 	// Custom resource paths
 	Dictionary resource_paths = FileAccess::get_resource_paths();
-	for (int i = 0; i < resource_paths.size(); i++) {
-		String key = resource_paths.get_key_at_index(i);
+	List<Variant> resource_keys;
+	resource_paths.get_key_list(&resource_keys);
+	for (const Variant &v : resource_keys) {
+		String key = v;
 		if (p_path.begins_with(key + "://")) {
 			return p_path.replace(key + ":/", resource_paths[key]);
 		}
--- a/core/config/project_settings.cpp
+++ b/core/config/project_settings.cpp
@@ -268,17 +268,20 @@ String ProjectSettings::globalize_path(const String &p_path) const {
 		return p_path.replace("user://", "");
 	}
 
-	// Custom resource paths
-	Dictionary resource_paths = FileAccess::get_resource_paths();
-	List<Variant> resource_keys;
-	resource_paths.get_key_list(&resource_keys);
-	for (const Variant &v : resource_keys) {
-		String key = v;
-		if (p_path.begins_with(key + "://")) {
-			return p_path.replace(key + ":/", resource_paths[key]);
+	// Custom paths
+	Vector<String> custom_paths = FileAccess::get_custom_paths();
+	for (const String &prefix : custom_paths) {
+		if (!p_path.begins_with(prefix + "://")) {
+			continue;
+		}
+		Dictionary data = FileAccess::get_custom_path_data(prefix);
+		if (data["type"] == "path") {
+			return p_path.replace(prefix + ":/", data["path"]);
+		} else {
+			// `data["type"]` is of type `"class"`, so no path is available.
+			return p_path;
 		}
 	}
-
 	return p_path;
 }
 
--- a/core/io/file_access.cpp
+++ b/core/io/file_access.cpp
@@ -900,6 +900,9 @@ void FileAccess::add_resource_path(const String &p_protocol, const String &p_pat
 	ERR_FAIL_COND_MSG(protocol == "res" || protocol == "user", vformat(R"(Protocol "%s://" is built-in.)", protocol));
 	ERR_FAIL_COND_MSG(resource_paths.has(protocol), vformat(R"(Protocol "%s://" is already registered as a resource path.)", protocol));
 	ERR_FAIL_COND_MSG(resource_paths_class.has(protocol), vformat(R"(Protocol "%s://" is already registered as a resource path class.)", protocol));
+	ERR_FAIL_COND_MSG(!Engine::get_singleton()->is_editor_hint() && protocol == "editor", R"(Protocol "editor://" cannot be registered when the editor is not running.)");
+
+	print_line(vformat("adding %s as a protocol for %s", p_protocol, p_path));
 
 	resource_paths[protocol] = p_path;
 }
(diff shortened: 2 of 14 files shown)


arkology:
This is big step (even not directly) towards global addons 🥳

adamscott:
> This is big step (even not directly) towards global addons 🥳

That's the goal!

core/io/file_access.cpp
```diff
@@ -852,6 +858,70 @@ String FileAccess::get_sha256(const String &p_file) {
...
 
+void FileAccess::add_resource_path(const String &p_protocol, const String &p_path) {
+	ERR_FAIL_COND_MSG(p_protocol.length() == 0, "Protocol parameter is empty.");
+
+	String protocol = p_protocol;
+	if (protocol.ends_with("://")) {
+		protocol = protocol.left(protocol.length() - 3);
+	}
+
+	ERR_FAIL_COND_MSG(protocol == "res" || protocol == "user", vformat("Protocol \"%s://\" is built-in.", protocol));
```
AThousandShips, line 869:
```suggestion
	ERR_FAIL_COND_MSG(protocol == "res" || protocol == "user", vformat(R"(Protocol "%s://" is built-in.)", protocol));
```
I'd use R strings over escaping here, for all

RedMser:
Why do `is_resource_path` and `get_resource_path(s)` fail for `res` and `user`? I understand that adding/removing them is dangerous, so those erroring makes sense.
But the current handling of the getter methods are just exposing an implementation detail to the user. They should instead be a consistent API for "every virtual path `*://` managed by Godot"...

adamscott:
> But the current handling of the getter methods are just exposing an implementation detail to the user.

No, `user://` and `res://` are special cases handled differently. This PR is not about exposing an implementation detail, it's a feature in itself.

core/io/file_access.cpp
```diff
@@ -852,6 +858,70 @@ String FileAccess::get_sha256(const String &p_file) {
 	return String::hex_encode_buffer(hash, 32);
 }
 
+void FileAccess::add_resource_path(const String &p_protocol, const String &p_path) {
+	ERR_FAIL_COND_MSG(p_protocol.length() == 0, "Protocol parameter is empty.");
```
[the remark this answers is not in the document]
adamscott: Oops! You're right!

core/io/file_access_extension.h
```diff
@@ -0,0 +1,159 @@
...
+#include "core/object/gdvirtual.gen.inc"
+#include "core/object/object.h"
+
+class FileAccessExtension : public FileAccess {
+	GDCLASS(FileAccessExtension, FileAccess);
+
+protected:
+	static void _bind_methods();
+
+	GDVIRTUAL2R(Error, _open_internal, String, int);
```
AThousandShips, line 44:
```suggestion
	GDVIRTUAL2R(Error, _open_internal, const String &, int);
```
I'd say they should preferably match their normal method signatures

core/io/file_access_extension.cpp
```diff
@@ -0,0 +1,355 @@
...
+}
+
+bool FileAccessExtension::is_open() const {
+	bool is_open = false;
+	GDVIRTUAL_CALL(_is_open, is_open);
+	return is_open;
+}
+
+String FileAccessExtension::get_path() const {
+	String path = "";
```
AThousandShips, line 54:
```suggestion
	String path;
```
Unnecessary defaults (all non simple types)

core/config/project_settings.cpp
```diff
@@ -268,6 +268,15 @@ String ProjectSettings::globalize_path(const String &p_path) const {
...
 	}
 
+	// Custom resource paths
+	Dictionary resource_paths = FileAccess::get_resource_paths();
+	for (int i = 0; i < resource_paths.size(); i++) {
+		String key = resource_paths.get_key_at_index(i);
+		if (p_path.begins_with(key + "://")) {
+			return p_path.replace(key + ":/", resource_paths[key]);
+		}
+	}
```
AThousandShips, line 278:
```suggestion
	List<Variant> resource_keys;
	resource_paths.get_key_list(&resource_keys);
	for (const Variant &v : resource_keys) {
		String key = v;
		if (p_path.begins_with(key + "://")) {
			return p_path.replace(key + ":/", resource_paths[key]);
		}
	}
```
Preferred (and faster) syntax, the `get_key_at_index` is unnecessarily slow

core/io/file_access_extension.cpp
adamscott: image

The compiler is screaming at me. I'll initialize the values.

akien-mga: `String` is default initialized, so the suggestion is correct.

Primitive types lime `bool`, `int`, or enums the like `Error`, do need an explicit initialization.

adamscott: Oh, I understand better now @AThousandShips's comment. Thanks @akien-mga!

zedrun00:
the pr is good.
i used the pr, i think will need add prefix_dir options in "EditorFileDialog" and "FileDialog"

CycloneRing:
@adamscott Do you think it's possible to add memory based file system to this PR? A simple VFS so files can be used directly from memory like "memory://my_scene.tscn"

adamscott:
> @adamscott Do you think it's possible to add memory based file system to this PR? A simple VFS so files can be used directly from memory like "memory://my_scene.tscn"

This clearly isn't in the scope of my PR.

scgm0:
> @adamscott Do you think it's possible to add memory based file system to this PR? A simple VFS so files can be used directly from memory like "memory://my_scene.tscn"

This? #98287

arkology:
Is there anything we can help with to bring the moment of merge of this PR closer?

Also if I'm correct this PR and https://github.com/godotengine/godot/pull/98544 are related.

scgm0:
> Is there anything we can help with to bring the moment of merge of this PR closer?
> 
> Also if I'm correct this PR and #98544 are related.

They are not related; they simply implement the same proposal (https://github.com/godotengine/godot-proposals/issues/6307). This PR is based on the existing FileAccess & DirAccess, while #98544 opts to refactor the I/O system, replacing the current FileAccess & DirAccess.

markushonok:
I think it would be renamed to custom schemes. `scheme://path`
See also https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

PR #87587, opened 2024-01-25
[RTL] Connect image update signals.

### LINKED ISSUE #87585: RichTextLabel does not listen to an image's `changed` signal to redraw content
### Tested versions

- Reproducible in: 4.3.dev (https://github.com/godotengine/godot/commit/4b6ad349886288405890b07d4a8da425eb3c97ec), 4.2.1.stable

### System information

Godot v4.2.1.stable unknown - Arch Linux #1 SMP PREEMPT_DYNAMIC Sat, 13 Jan 2024 14:37:14 +0000 - X11 - Vulkan (Forward+) - dedicated AMD Radeon RX 6500 XT (RADV NAVI24) () - AMD Ryzen 5 5600G with Radeon Graphics (12 Threads)

### Issue description

The `RichTextLabel` node allows embedding images through BBCode, which internally loads any `Texture2D` resource on the project path. This can be used with custom `Texture2D` implementations as well.

However, unlike many other nodes such as `TextureRect`, `Sprite2D`, `Sprite3D` and `Button`, the label does not listen to the resource's `changed` signal. Thus, in a situation where the resource has changed and has emitted such signal, the label will not update and redraw it's content, unless forced to do so (by calling `queue_redraw` manually or something else that forces a redraw, like changing the window size).

https://github.com/godotengine/godot/assets/6501975/da50d07f-ffa0-4c7d-9b83-dd6a6d146a97

This can occur in the scenario where a custom `Texture2D` is created which has logic to swap texture files internally, and thus uses `emit_changed` to ensure a redraw is done:

### Steps to reproduce

- Create a custom resource inheriting ``Texture2D``
```gdscript
extends Texture2D
class_name CustomTexture2D

var tex : Texture2D
static var flip := false

func _init():
	choose_texture()

func choose_texture():
	if flip:
		tex = preload("res://icon.png")
	else:
		tex = preload("res://icon_2.png")
	flip = not flip
	emit_changed()

# 2D nodes
func _draw(to_canvas_item, pos, modulate, transpose):
	tex.draw(to_canvas_item, pos, modulate, transpose)

func _draw_rect(to_canvas_item, rect, tile, modulate, transpose):
	tex.draw_rect(to_canvas_item, rect, tile, modulate, transpose)

...
```
- Create a resource and store it in the project files
image
- Create a `RichTextLabel` that uses this new resource
image
- Make the resource swap textures on some other object, like a `Button`:
```gdscript
extends Button

@export var resource : CustomTexture2D

func _on_pressed():
	if resource:
		resource.choose_texture()
```

[ReproRichTextBug.zip](https://github.com/godotengine/godot/files/14053938/ReproRichTextBug.zip)

bruvzg:
Fixes https://github.com/godotengine/godot/issues/87585

https://github.com/godotengine/godot/assets/7645683/e2320ddd-3aa7-4995-bdc6-7b4e1348b2c1

### DIFF
--- a/scene/gui/rich_text_label.cpp
+++ b/scene/gui/rich_text_label.cpp
@@ -239,7 +239,8 @@ void RichTextLabel::_update_line_font(ItemFrame *p_frame, int p_line, const Ref<
 	RID t = l.text_buf->get_rid();
 	int spans = TS->shaped_get_span_count(t);
 	for (int i = 0; i < spans; i++) {
-		ItemText *it = reinterpret_cast<ItemText *>((uint64_t)TS->shaped_get_span_meta(t, i));
+		Item *it_span = items.get_or_null(TS->shaped_get_span_meta(t, i));
+		ItemText *it = reinterpret_cast<ItemText *>(it_span);
 		if (it) {
 			Ref<Font> font = p_base_font;
 			int font_size = p_base_font_size;
@@ -307,7 +308,7 @@ float RichTextLabel::_resize_line(ItemFrame *p_frame, int p_line, const Ref<Font
 				Size2 img_size = img->size;
 				if (img->size_in_percent) {
 					img_size = _get_image_size(img->image, p_width * img->rq_size.width / 100.f, p_width * img->rq_size.height / 100.f, img->region);
-					l.text_buf->resize_object((uint64_t)it, img_size, img->inline_align);
+					l.text_buf->resize_object(it->rid, img_size, img->inline_align);
 				}
 			} break;
 			case ITEM_TABLE: {
@@ -455,9 +456,9 @@ float RichTextLabel::_resize_line(ItemFrame *p_frame, int p_line, const Ref<Font
 				}
 				int row_idx = (table->align_to_row < 0) ? table->rows_baseline.size() - 1 : table->align_to_row;
 				if (table->rows_baseline.size() != 0 && row_idx < (int)table->rows_baseline.size() - 1) {
-					l.text_buf->resize_object((uint64_t)it, Size2(table->total_width, table->total_height), table->inline_align, Math::round(table->rows_baseline[row_idx]));
+					l.text_buf->resize_object(it->rid, Size2(table->total_width, table->total_height), table->inline_align, Math::round(table->rows_baseline[row_idx]));
 				} else {
-					l.text_buf->resize_object((uint64_t)it, Size2(table->total_width, table->total_height), table->inline_align);
+					l.text_buf->resize_object(it->rid, Size2(table->total_width, table->total_height), table->inline_align);
 				}
 			} break;
 			default:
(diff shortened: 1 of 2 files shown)


AThousandShips (dismissed):
LGTM implementation wise

scene/gui/rich_text_label.cpp
```diff
@@ -3250,6 +3254,9 @@ void RichTextLabel::add_image(const Ref<Texture2D> &p_image, int p_width, int p_
 	item->key = p_key;
 	item->tooltip = p_tooltip;
 
+	item->owner = this;
+	item->image->connect_changed(callable_mp(this, &RichTextLabel::_texture_changed), CONNECT_REFERENCE_COUNTED);
```
KoBeWi, line 3258: Why it's not connected to `queue_redraw()` directly?

rsubtil:
There's another issue I didn't notice; the size of an `ItemImage` remains the same, so any size changes do not get updated. The `queue_redraw` is not enough, it needs some text changes to occur:

https://github.com/godotengine/godot/assets/6501975/433bbd88-8f07-4e56-93b0-4f8beddb881c

scene/gui/rich_text_label.cpp
bruvzg: This would make sense with the original version, but now it's doing more stuff.

bruvzg:
Updated to fix size changes as well. Also replaced some nasty "cast raw item pointer to `uint64_t`" code with safer RIDs.

rsubtil:
Can confirm it now works like a charm! :tada:

scene/gui/rich_text_label.cpp
```diff
@@ -932,7 +933,8 @@ int RichTextLabel::_draw_line(ItemFrame *p_frame, int p_line, const Vector2 &p_o
 		// Draw inlined objects.
 		Array objects = TS->shaped_text_get_objects(rid);
 		for (int i = 0; i < objects.size(); i++) {
-			Item *it = reinterpret_cast<Item *>((uint64_t)objects[i]);
+			Item *it_span = items.get_or_null(objects[i]);
+			Item *it = reinterpret_cast<Item *>(it_span);
```
AThousandShips, line 937: Unnecessary I think, just collapse to
```cpp
Item *it_span = items.get_or_null(objects[i]);
```

scene/gui/rich_text_label.cpp
```diff
@@ -1601,7 +1603,8 @@ float RichTextLabel::_find_click_in_line(ItemFrame *p_frame, int p_line, const V
 
 		Array objects = TS->shaped_text_get_objects(rid);
 		for (int i = 0; i < objects.size(); i++) {
-			Item *it = reinterpret_cast<Item *>((uint64_t)objects[i]);
+			Item *it_span = items.get_or_null(objects[i]);
+			Item *it = reinterpret_cast<Item *>(it_span);
```
AThousandShips, line 1607: Ditto

AThousandShips (approved):
Looks good otherwise

akien-mga (approved):

akien-mga:
Thanks!

PR #87590, opened 2024-01-25
Add transfer queue support to RenderingDevice and enable multithreaded resource loading

DarioSamo:
A few major bottlenecks were identified in Godot that this PR addresses.
- Access to RenderingDevice is pretty much entirely blocked using a single thread safety mutex. This means that any thread that wishes to create buffers or textures from different threads needs to wait for the others or even the main thread to free its access.
- When Vsync is enabled, waiting is performed while inside the mutex itself, so any arbitrary waiting caused by the presentation rate of the screen potentially stalls all efforts of loading resources in background threads. This can result in major increases in loading times in most games that choose to draw a loading screen and have Vsync enabled.
- While the hardware provides dedicated queues for transfer operations that are much faster for the process and can be run independently from the main application loop, Godot chooses to use the same staging buffer logic that it uses for resources uploading during each frame that it uses for resources uploaded from background threads.

The PR solves all these bottlenecks and also simplifies the separate setup command buffer logic that was in place in RenderingDevice. RenderingDevice is now able to create "Transfer Workers" on the fly that can use their own staging buffer and command buffer to record the upload of all resources queued by the thread. Once it reaches a certain saturation point, it'll submit all the work to the dedicated transfer queue. The main loop will wait if necessary for these transfer workers if it's been determined that one of the resources that was uploaded with a transfer worker was used in the frame. This design allows to effectively upload resources in the background without stalling the main thread at all until the resource needs to be used.

The PR also changes the flexibility of the thread safety of RenderingDevice: whatever is allowed to be called from different threads is now allowed with no mutex involved, and whatever is not allowed now has a thread guard in place instead. A thread guard is a much cheaper option than a mutex and it also verifies that RenderingDevice is being used with its intended behavior: drawing from the main loop and not from other threads that could potentially cause issues. This was mostly done following @reduz's advice as noted [here](https://gist.github.com/reduz/950a9c9dd2878e56cd545d790b3f6a9b).

So far we lack a public project that allows to showcase loading a large amount of resources, but one of W4's internal projects that uses a large amount of textures and meshes had some good results:
- `master` with Vsync enabled (60 HZ Monitor): ~25 seconds
- `master` with Vsync disabled: ~2.5 seconds (10x faster!)
- `transfer_queues` with both Vsync enabled and disabled: ~2.1 seconds

We can see one of the major benefits of this approach is that it makes the behavior much more consistent regardless of the user's vsync settings and refresh rate.

Projects with lots of resources that rely on background resource loading during loading screens or even during gameplay are the ideal candidates for testing out this PR.

TODO:
- [ ] Decide on how to configure the max amount of transfer workers (right now it is just derived from the CPU count since it matches the logic of WorkerThreadPool).
- [x] Experiment with taking out the thread-safety of RID Owner by adding the logic from @reduz's PR (https://github.com/godotengine/godot/pull/86333).

### DIFF
--- a/servers/rendering/rendering_device.cpp
+++ b/servers/rendering/rendering_device.cpp
@@ -36,7 +36,14 @@
 #include "core/config/project_settings.h"
 #include "core/io/dir_access.h"
 
+// TODO: Thread safety
+// - Roll back thread safe attribute for RID_Owner members after the read-only/atomic update scheme is implemented.
+
 #define FORCE_SEPARATE_PRESENT_QUEUE 0
+
+#define ERR_RENDER_THREAD_MSG String("This function (") + String(__func__) + String(") can only be called from the render thread. ")
+#define ERR_RENDER_THREAD_GUARD() ERR_FAIL_COND_MSG(render_thread_id != Thread::get_caller_id(), ERR_RENDER_THREAD_MSG);
+#define ERR_RENDER_THREAD_GUARD_V(m_ret) ERR_FAIL_COND_V_MSG(render_thread_id != Thread::get_caller_id(), (m_ret), ERR_RENDER_THREAD_MSG);
 
 /**************************/
 /**** HELPER FUNCTIONS ****/
@@ -137,6 +144,8 @@ RenderingDevice::ShaderSPIRVGetCacheKeyFunction RenderingDevice::get_spirv_cache
 /***************************/
 
 void RenderingDevice::_add_dependency(RID p_id, RID p_depends_on) {
+	_THREAD_SAFE_METHOD_
+
 	if (!dependency_map.has(p_depends_on)) {
 		dependency_map[p_depends_on] = HashSet<RID>();
 	}
@@ -151,6 +160,8 @@ void RenderingDevice::_add_dependency(RID p_id, RID p_depends_on) {
 }
 
 void RenderingDevice::_free_dependencies(RID p_id) {
+	_THREAD_SAFE_METHOD_
+
 	// Direct dependencies must be freed.
 
 	HashMap<RID, HashSet<RID>>::Iterator E = dependency_map.find(p_id);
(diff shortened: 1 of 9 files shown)


Saul2022:
I know big projects i needed but the most i could find was the lumberyard one so tested on my igpu (radeon vega 3)  Results

Screenshot_2024-01-26-11-27-09-34_cc0c40aae00121c8e1b1866ef91e05c7
Screenshot_2024-01-26-11-23-54-98_cc0c40aae00121c8e1b1866ef91e05c7

DarioSamo:
> I know big projects i needed but the most i could find was the lumberyard one so tested on my igpu (radeon vega 3) Results
the 4.3 is the 1 minute and 4.2 is 1.04 , not much difference but it an igpu vs gb's.

On the editor or games? The benefits mostly apply to using threaded resource loading, it shouldn't really matter that much on situations where it's loading but stuck without drawing anything.

Saul2022:
Well now added loading screen and nice results from (though vsync dissabled, it enabled was 15 secs ) And while they may not be the most accurate, it works ( still figuring out how to make it in editor)

Screenshot_2024-02-02-15-07-00-95_1eeb8c37998c08f2a00dbbf329757a78
Screenshot_2024-02-02-15-09-32-82_1eeb8c37998c08f2a00dbbf329757a78

Calinou (approved):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/10e111477db68fe65776a1d68fb1ffccaf6520fc + https://github.com/godotengine/godot/pull/86333), it works as expected on Linux + NVIDIA.

On https://github.com/godotengine/tps-demo, it loads a full second faster than on `master` with neither PR applied.

servers/rendering/rendering_device.cpp
```diff
@@ -5094,6 +5614,9 @@ Error RenderingDevice::initialize(RenderingContextDriver *p_context, DisplayServ
 		ERR_CONTINUE(err != OK);
 	}
 
+	// TODO: How should this max size be determined?
+	transfer_worker_pool_max_size = OS::get_singleton()->get_processor_count();
```
RandomShaper, line 5618: I haven't checked this very deeply yet. So far, I got the impression that this code is not spawning threads but having parallel bookkeeping for thr threads already existing that may be loading resources.

If that's true, a beginning would be `WorkerThreadPool::get_thread_count()`. Even if currently resource loading is only allowed on low priority threads, that can mean the whole thread pool being used because there's the potential of more low-prio tasks in flight than that number and also because the are no pre-assigned low-prio threads, so any thread from the pool can be running a low-prio task at a given moment.

That said, the main thead as well as arbitrary threads can be loading resources at a given point. So, I'd at least add one to that count. If user creates extra threads while a WTP-backed load is in progress, well, it's only fair that they have to wait for a transfer worker to be available.

DarioSamo: That's correct, it doesn't spawn threads on its own nor does it pre-create them, it just allows a maximum number of concurrent workers. Using WTP as a reference does seem like a good start.

servers/rendering/rendering_device.cpp
```diff
@@ -385,32 +425,85 @@ void RenderingDevice::_staging_buffer_execute_required_action(StagingRequiredAct
...
+
+	ERR_FAIL_COND_V_MSG(draw_list, ERR_INVALID_PARAMETER,
+			"Copying buffers is forbidden during creation of a draw list");
+	ERR_FAIL_COND_V_MSG(compute_list, ERR_INVALID_PARAMETER,
+			"Copying buffers is forbidden during creation of a compute list");
+
+	Buffer *src_buffer = _get_buffer_from_owner(p_src_buffer);
+	if (!src_buffer) {
+		ERR_FAIL_V_MSG(ERR_INVALID_PARAMETER, "Source buffer argument is not a valid buffer of any type.");
+	}
```
AThousandShips, line 439:
```suggestion
	ERR_FAIL_NULL_V_MSG(src_buffer, ERR_INVALID_PARAMETER, "Source buffer argument is not a valid buffer of any type.");
```

servers/rendering/rendering_device.cpp
```diff
@@ -385,32 +425,85 @@ void RenderingDevice::_staging_buffer_execute_required_action(StagingRequiredAct
...
+
+	Buffer *dst_buffer = _get_buffer_from_owner(p_dst_buffer);
+	if (!dst_buffer) {
+		ERR_FAIL_V_MSG(ERR_INVALID_PARAMETER, "Destination buffer argument is not a valid buffer of any type.");
+	}
```
AThousandShips, line 444:
```suggestion
	ERR_FAIL_NULL_V_MSG(dst_buffer, ERR_INVALID_PARAMETER, "Destination buffer argument is not a valid buffer of any type.");
```

AThousandShips (approved):

DarioSamo:
I'll close this one as its changes are included in #90400.

PR #87595, opened 2024-01-25
C#: Fix loading plugin to support debug symbols

### LINKED ISSUE #43: Support for Godot 4 + mono

Linux Manjaro KDE

**Issue description:**

I'm trying to use Vscode with Godot 4 + mono, but the options for C# doesnt appear as described in the README.MD

granitrocky:
Fixes https://github.com/godotengine/godot-csharp-vscode/issues/43

This change replaces `LoadFromStream` with a direct `LoadFromAssemblyPath` to allow the C# module to interface with `netcoredbg`

This solution comes from this issue:
https://github.com/Samsung/netcoredbg/issues/91#issuecomment-1189690292

### DIFF
--- a/modules/mono/glue/GodotSharp/GodotPlugins/PluginLoadContext.cs
+++ b/modules/mono/glue/GodotSharp/GodotPlugins/PluginLoadContext.cs
@@ -54,19 +54,9 @@ namespace GodotPlugins
             string? assemblyPath = _resolver.ResolveAssemblyToPath(assemblyName);
             if (assemblyPath != null)
             {
+                // Loading directly from the path lets `netcoredbg` see the debug symbols
                 AssemblyLoadedPath = assemblyPath;
-
-                // Load in memory to prevent locking the file
-                using var assemblyFile = File.Open(assemblyPath, FileMode.Open, FileAccess.Read, FileShare.Read);
-                string pdbPath = Path.ChangeExtension(assemblyPath, ".pdb");
-
-                if (File.Exists(pdbPath))
-                {
-                    using var pdbFile = File.Open(pdbPath, FileMode.Open, FileAccess.Read, FileShare.Read);
-                    return LoadFromStream(assemblyFile, pdbFile);
-                }
-
-                return LoadFromStream(assemblyFile);
+                return LoadFromAssemblyPath(assemblyPath);
             }
 
             return null;


raulsntos:
This sounds like a bug on netcoredbg. `AssemblyLoadContext.LoadFromStream` should work as it does in Microsoft's debugger.

As I understand it from reading the issue you linked, netcoredbg is unable to find the PDB file when using the `AssemblyLoadContext.LoadFromStream` overload that only takes the assembly stream (the DLL). In this case, they try to find the PDB file name from the DLL and if it doesn't contain it then they can't find it.

I haven't checked but I'd imagine the DLL built from Godot **contains** the PDB file name, so they should be able to find it. But even if it didn't contain it, we're not using that overload. We're using the overload that takes the assembly stream AND the assembly symbols stream (the PDB), so we're already providing the PDB. This case (our case) should work, and if it doesn't then it's a bug on netcoredbg.

There's also a linked PR (https://github.com/Samsung/netcoredbg/pull/108) that seems like it fixes the netcoredbg bug, so I suggest waiting for that instead. You could try that branch and see if it fixes the issue for you.

---

As for why we are using `AssemblyLoadContext.LoadFromStream`, if you read the comment: we're loading in memory to prevent locking the file.

https://github.com/godotengine/godot/blob/5034478611697358f4e135c69f364ef6d8cd54dc/modules/mono/glue/GodotSharp/GodotPlugins/PluginLoadContext.cs#L59

granitrocky:
I agree that the netcoredbg branch is probably the better option if it gets resolved, but reading through it, I do see that it's pretty stale.

Are we sure that `LoadFromAssemblyPath` locks the file? I couldn't find anything that says it does. Even took a quick glance through some of the code in the dotnet repo and nothing jumped out at me.

raulsntos:
AFAIK it's a Windows thing but I don't use Windows so I can't check. It seems GDExtensions have that issue right now:
- https://github.com/godotengine/godot/issues/82536

thanikTheDev:
I just created a new PR(Samsung/netcoredbg#164) which should fix this issue. I do believe the netcoredbg route because it was unable to read in-memory module symbols. msdbg does not.

thanikTheDev:
@granitrocky The changes have been implemented so you can try the latest version of netcoredbg.

granitrocky:
@thanikTheDev Thanks! I'll try it out and if it fixes the issue I can close this PR.

raulsntos:
@granitrocky Can you confirm that this was fixed upstream?

granitrocky:
@raulsntos I can confirm that the latest `netcoredbg` works. I will close this issue.

granitrocky:
I used the `netcore` emacs package along with `dap-mode`. I did have to manually change the command line that was invoked to include a path to godot on the netcoredbg invocation. I imagine there's a better way to do that, though.

Novack:
> I used the `netcore` emacs package along with `dap-mode`. I did have to manually change the command line that was invoked to include a path to godot on the netcoredbg invocation. I imagine there's a better way to do that, though.
> 
> This does seem to be a VSCodium related problem and not netcoredbg or Godot
> 
> [VSCodium/vscodium#82](https://github.com/VSCodium/vscodium/issues/82)

I think either you or me misreading something :)

Apart from the "easy way" not working anymore, that post you linked seems to indicate precisely how to make it work. In fact is the current [recommended approach](https://github.com/VSCodium/vscodium/blob/master/docs/index.md#proprietary-debugging-tools) by vscodium team. Its only in godot that we are having issues to make it work.

PR #87599, opened 2024-01-26
Added Label3D auto localization

DeltaW0x:
Added support for automatic key-matched string localization directly from the inspector, like present in Label (2D)

### DIFF
--- a/doc/classes/Label3D.xml
+++ b/doc/classes/Label3D.xml
@@ -48,6 +48,10 @@
 		<member name="alpha_scissor_threshold" type="float" setter="set_alpha_scissor_threshold" getter="get_alpha_scissor_threshold" default="0.5">
 			Threshold at which the alpha scissor will discard values.
 		</member>
+		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true">
+			Toggles if any text should automatically change to its translated version depending on the current locale.
+			Also decides if the node's strings should be parsed for POT generation.
+		</member>
 		<member name="autowrap_mode" type="int" setter="set_autowrap_mode" getter="get_autowrap_mode" enum="TextServer.AutowrapMode" default="0">
 			If set to something other than [constant TextServer.AUTOWRAP_OFF], the text gets wrapped inside the node's bounding rectangle. If you resize the node, it will change its height automatically to show all the text. To see how each mode behaves, see [enum TextServer.AutowrapMode].
 		</member>
--- a/scene/3d/label_3d.cpp
+++ b/scene/3d/label_3d.cpp
@@ -57,6 +57,9 @@ void Label3D::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_language", "language"), &Label3D::set_language);
 	ClassDB::bind_method(D_METHOD("get_language"), &Label3D::get_language);
 
+	ClassDB::bind_method(D_METHOD("set_auto_translate", "enable"), &Label3D::set_auto_translate);
+	ClassDB::bind_method(D_METHOD("is_auto_translating"), &Label3D::is_auto_translating);
+
 	ClassDB::bind_method(D_METHOD("set_structured_text_bidi_override", "parser"), &Label3D::set_structured_text_bidi_override);
 	ClassDB::bind_method(D_METHOD("get_structured_text_bidi_override"), &Label3D::get_structured_text_bidi_override);
 
@@ -164,6 +167,9 @@ void Label3D::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::INT, "structured_text_bidi_override", PROPERTY_HINT_ENUM, "Default,URI,File,Email,List,None,Custom"), "set_structured_text_bidi_override", "get_structured_text_bidi_override");
 	ADD_PROPERTY(PropertyInfo(Variant::ARRAY, "structured_text_bidi_override_options"), "set_structured_text_bidi_override_options", "get_structured_text_bidi_override_options");
 
+	ADD_GROUP("Localization", "");
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "auto_translate"), "set_auto_translate", "is_auto_translating");
+
 	BIND_ENUM_CONSTANT(FLAG_SHADED);
 	BIND_ENUM_CONSTANT(FLAG_DOUBLE_SIDED);
 	BIND_ENUM_CONSTANT(FLAG_DISABLE_DEPTH_TEST);
@@ -208,7 +214,7 @@ void Label3D::_notification(int p_what) {
 			viewport->disconnect("size_changed", callable_mp(this, &Label3D::_font_changed));
 		} break;
 		case NOTIFICATION_TRANSLATION_CHANGED: {
-			String new_text = tr(text);
+			String new_text = atr(text);
 			if (new_text == xl_text) {
 				return; // Nothing new.
 			}
@@ -635,8 +641,11 @@ void Label3D::_shape() {
 }
 
 void Label3D::set_text(const String &p_string) {
+	if (text == p_string) {
+		return;
+	}
 	text = p_string;
-	xl_text = tr(p_string);
+	xl_text = atr(p_string);
 	dirty_text = true;
 	_queue_update();
 }
@@ -683,6 +692,20 @@ void Label3D::set_text_direction(TextServer::Direction p_text_direction) {
 
 TextServer::Direction Label3D::get_text_direction() const {
 	return text_direction;
+}
+
+void Label3D::set_auto_translate(bool p_enable) {
+	if (p_enable == auto_translate) {
+		return;
+	}
+
+	auto_translate = p_enable;
+
+	notification(MainLoop::NOTIFICATION_TRANSLATION_CHANGED);
+}
+
+bool Label3D::is_auto_translating() const {
+	return auto_translate;
 }
 
 void Label3D::set_language(const String &p_language) {
--- a/scene/3d/label_3d.h
+++ b/scene/3d/label_3d.h
@@ -133,6 +133,8 @@ private:
 	TextServer::StructuredTextParser st_parser = TextServer::STRUCTURED_TEXT_DEFAULT;
 	Array st_args;
 
+	bool auto_translate = true;
+
 	RID text_rid;
 	Vector<RID> lines_rid;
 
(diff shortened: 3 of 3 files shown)


akien-mga:
Thanks for the contribution!

Did you test whether the changes work? I might be missing something, but the code doesn't look like it will actually do anything to enable translations. It's just the interface for it copied over from `control.h`, but there's no code in `label_3d.cpp` that will actually make use of the added properties to trigger a translation from what I can see.

DeltaW0x:
The code was tested and it works as it should, as far as I know.I have built it and both the matching and the ability to change language at runtime work, the label updates . Sorry but this is my first pull request ever.

akien-mga:
Are you sure you committed all your local changes? In `label_3d.cpp`, the values for `auto_translate` can set and read, but it's never used by any internal API (again, unless I'm missing something, which may be possible).

I would expect that it would need to be handled in `NOTIFICATION_TRANSLATION_CHANGED` in `_notification` to actually decide whether to translate or not.

DeltaW0x:
I'm sorry, maybe I'm misunderstanding something, the code that handles the translation was copied directly from `contro.h`, which is how the Label node can handle the automatic key-matching directly from the Label inspector text field, without tr(). It is exaclty the same, just move from control to Label3D. There is a nicer way to do this without copying a bunch of code probably, but this seems to work fine for now, better than having nothing or to have to translate the strings from script using tr(). 
The translation changed notification is pushed like is pushed in `contro.h`

```
Label3D::set_auto_translate(bool p_enable) {
	ERR_MAIN_THREAD_GUARD;
	if (p_enable == auto_translate) {
		return;
	}

	auto_translate = p_enable;

	notification(MainLoop::NOTIFICATION_TRANSLATION_CHANGED);
}
```

akien-mga:
Well the code in `label.cpp` works because of this part:
```cpp
void Label::_notification(int p_what) {
	switch (p_what) {
		case NOTIFICATION_TRANSLATION_CHANGED: {
			String new_text = atr(text);
			if (new_text == xl_text) {
				return; // Nothing new.
			}
			xl_text = new_text;
			if (visible_ratio < 1) {
				visible_chars = get_total_character_count() * visible_ratio;
			}
			dirty = true;

			queue_redraw();
			update_configuration_warnings();
		} break;
```

So I'm really puzzled it would work in `label_3d.cpp` without similar code. But I guess I should test the PR to see :D

akien-mga:
Ah I see `label_3d.cpp` does have this code already:
```cpp
		case NOTIFICATION_TRANSLATION_CHANGED: {
			String new_text = tr(text);
			if (new_text == xl_text) {
				return; // Nothing new.
			}
			xl_text = new_text;
			dirty_text = true;
			_queue_update();
		} break;
```

So indeed what was missing was just the code to trigger this notification.

DeltaW0x:
Oh please don't worry it's ok, I hadn't seen it either, it just worked so i thought about pushing the repo '^^, I'm just inexperienced so I thought I had missed something

scene/3d/label_3d.cpp
```diff
@@ -685,6 +695,38 @@ TextServer::Direction Label3D::get_text_direction() const {
...
+	notification(MainLoop::NOTIFICATION_TRANSLATION_CHANGED);
+}
+
+bool Label3D::is_localizing_numeral_system() const {
+	ERR_READ_THREAD_GUARD_V(false);
+	return localize_numeral_system;
+}
+
+void Label3D::set_auto_translate(bool p_enable) {
+	ERR_MAIN_THREAD_GUARD;
```
akien-mga, line 715: Threads guards aren't used in `Label3D` currently, so this one should likely be removed (same in the getter). I don't know if it means that the class is thread-safe or unsafe, but at least it operates under different assumptions than Controls.

scene/3d/label_3d.h
```diff
@@ -182,6 +185,16 @@ class Label3D : public GeometryInstance3D {
...
 
+	void set_localize_numeral_system(bool p_enable);
+	bool is_localizing_numeral_system() const;
+
+	void set_auto_translate(bool p_enable);
+	bool is_auto_translating() const;
+
+	_FORCE_INLINE_ String atr(const String p_string) const {
+		return is_auto_translating() ? tr(p_string) : p_string;
+	};
```
akien-mga, line 196: I know it was copied from `control.h` as is but it's redundant there too.
```suggestion
	}
```

akien-mga (commented):
So, now I understand better why it seems to work, and my initial confusion.

The key thing which makes auto translation work is the `atr()` method, which handles checking whether auto translation is enabled or not. You added the implementation in `label_3d.h`, but you're not making use of it in `label_3d.cpp`.

You need to change this for it to work like in `label.cpp`:
```diff
diff --git a/scene/3d/label_3d.cpp b/scene/3d/label_3d.cpp
index 3fb472335e..ff8bd32866 100644
--- a/scene/3d/label_3d.cpp
+++ b/scene/3d/label_3d.cpp
@@ -208,7 +208,7 @@ void Label3D::_notification(int p_what) {
 			viewport->disconnect("size_changed", callable_mp(this, &Label3D::_font_changed));
 		} break;
 		case NOTIFICATION_TRANSLATION_CHANGED: {
-			String new_text = tr(text);
+			String new_text = atr(text);
 			if (new_text == xl_text) {
 				return; // Nothing new.
 			}
@@ -635,8 +635,11 @@ void Label3D::_shape() {
 }
 
 void Label3D::set_text(const String &p_string) {
+	if (text == p_string) {
+		return;
+	}
 	text = p_string;
-	xl_text = tr(p_string);
+	xl_text = atr(p_string);
 	dirty_text = true;
 	_queue_update();
 }
```

If you test the current PR, turning *off* auto translation shouldn't work, it will keep translating.

DeltaW0x:
Fixed the issues you pointed out. Now the automatic translation indeed stops when it's disabled

akien-mga:

Then from my perspective the PR looks ready. I'll still aim to actually test it tomorrow but the code makes sense to me.

CC @YeldhamDev who's our auto translation expert now ;)

akien-mga:
Ah one more thing actually, you need to add documentation for the new property:
```diff
diff --git a/doc/classes/Label3D.xml b/doc/classes/Label3D.xml
index 977fb00402..c4c9c58543 100644
--- a/doc/classes/Label3D.xml
+++ b/doc/classes/Label3D.xml
@@ -48,6 +48,10 @@
 		<member name="alpha_scissor_threshold" type="float" setter="set_alpha_scissor_threshold" getter="get_alpha_scissor_threshold" default="0.5">
 			Threshold at which the alpha scissor will discard values.
 		</member>
+                <member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true">
+                        Toggles if any text should automatically change to its translated version depending on the current locale.
+                        Also decides if the node's strings should be parsed for POT generation.
+                </member>
 		<member name="autowrap_mode" type="int" setter="set_autowrap_mode" getter="get_autowrap_mode" enum="TextServer.AutowrapMode" default="0">
 			If set to something other than [constant TextServer.AUTOWRAP_OFF], the text gets wrapped inside the node's bounding rectangle. If you resize the node, it will change its height automatically to show all the text. To see how each mode behaves, see [enum TextServer.AutowrapMode].
 		</member>
```

Not sure if the part about POT generation is accurate (I copied the description from Control), Yeldham can maybe clarify.

scene/3d/label_3d.h
```diff
@@ -133,6 +133,9 @@ class Label3D : public GeometryInstance3D {
 	TextServer::StructuredTextParser st_parser = TextServer::STRUCTURED_TEXT_DEFAULT;
 	Array st_args;
 
+	bool auto_translate = true;
+	bool localize_numeral_system = true;
```
akien-mga, line 137: I missed this one, should be removed too.

DeltaW0x: You mean only `localize_numeral_system` right? isn't `auto_translate ` is necessary?

akien-mga: Yes `localize_numeral_system` is now unused and should be removed. The other one is needed.

YeldhamDev:
This PR kinda came in a bad time right now, since I'm re-doing how auto translation works. To make matters worse, having yet another type of node requiring it makes the system even more convoluted, since both `Control` and `Window` already share a lot of duplicate code.

DeltaW0x:
Of course, my perfect timing as always. Anyways, i've squashed the commits. It should be good for merging but at this point it's kinda useless i think, if there's a new system in development

akien-mga (approved):
I'll approve from my side I think the implementation is straightforward and correct. Great work!

But before merging, let's see first how @YeldhamDev's refactor in #87530 goes, and how to include Label3D one way or another.

doc/classes/Label3D.xml
```diff
@@ -48,6 +48,10 @@
 		<member name="alpha_scissor_threshold" type="float" setter="set_alpha_scissor_threshold" getter="get_alpha_scissor_threshold" default="0.5">
 			Threshold at which the alpha scissor will discard values.
 		</member>
+		<member name="auto_translate" type="bool" setter="set_auto_translate" getter="is_auto_translating" default="true">
++                        Toggles if any text should automatically change to its translated version depending on the current locale.
++                        Also decides if the node's strings should be parsed for POT generation.
++       </member>
```
bruvzg, line 54:
```suggestion
			Toggles if any text should automatically change to its translated version depending on the current locale.
			Also decides if the node's strings should be parsed for POT generation.
		</member>
```

It should use tabs not spaces (also remove `+` at the line start).

bruvzg (approved):

bruvzg:
The same change can be done to the `TextMesh` as well.

DeltaW0x:
If there's more than one node that needs autotranslation maybe it's just better to wait for the general refractor, I don't think it's a good idea to have the same exact code just copied and pasted in multiple spaces

InfernalWAVE:
This is great! Any idea what version we could expect to see this implemented?

Mickeon:
The implementation depends on https://github.com/godotengine/godot/pull/87530 (TL:DR Every Node having `auto_translate_mode`). Once that's merged this PR should be rebased as it would basically come "for free", without code duplication.

Mickeon:
https://github.com/godotengine/godot/pull/87530 has been merged. This is the go. Time to rebase this to use the new `auto_translate_mode`.

DeltaW0x:
Wait, I thought there's no need for this with the new localization system?

akien-mga:
I believe the changes to `Label3D::set_text()` would still be needed to actually switch to using `atr()` instead of `tr()` (and likewise for `TextMesh`).

akien-mga:
> I believe the changes to `Label3D::set_text()` would still be needed to actually switch to using `atr()` instead of `tr()` (and likewise for `TextMesh`).

A task for you @YeldhamDev ;)

YeldhamDev:
Done: #89056

PR #87602, opened 2024-01-26
Update `XRPose.get_adjusted_transform` documentation

DanielKinsman:
The basis is not adjusted by this function, make this explicit in the documentation.

https://github.com/godotengine/godot/blob/master/servers/xr/xr_pose.cpp#L96

Can decline if you think it is already obvious enough without needing further explanation.

### DIFF
--- a/doc/classes/XRPose.xml
+++ b/doc/classes/XRPose.xml
@@ -14,7 +14,7 @@
 		<method name="get_adjusted_transform" qualifiers="const">
 			<return type="Transform3D" />
 			<description>
-				Returns the [member transform] with world scale and our reference frame applied. This is the transform used to position [XRNode3D] objects.
+				Returns the [member transform] with world scale and our reference frame applied. This is the transform used to position [XRNode3D] objects. Only the origin is adjusted, not the basis.
 			</description>
 		</method>
 	</methods>


doc/classes/XRPose.xml
```diff
@@ -14,7 +14,7 @@
 		<method name="get_adjusted_transform" qualifiers="const">
 			<return type="Transform3D" />
 			<description>
-				Returns the [member transform] with world scale and our reference frame applied. This is the transform used to position [XRNode3D] objects.
+				Returns the [member transform] with world scale and our reference frame applied. This is the transform used to position [XRNode3D] objects. Only the origin is adjusted, not the basis.
```
AThousandShips, line 17: This doesn't seem to be true from the code though:
https://github.com/godotengine/godot/blob/5034478611697358f4e135c69f364ef6d8cd54dc/servers/xr/xr_pose.cpp#L99

It transforms the result by the reference frame

Mickeon: If OP is confident on this there may be a bug either in Godot itself or their hardware, I wonder.

AThousandShips: Might be that their testing conditions has a specific reference frame case that doesn't showcase all outcomes, but this code clearly modifies the full transform

DanielKinsman:
I think I just wasn't familiar with the conditions under which the adjusted transform differs from the regular transform. Calling `XRServer.center_on_hmd()` does indeed change the adjusted basis. Pressing the center button on my headset (pico 4) doesn't affect it though and everything ends up screwy. Will have to see if there is a way to hook into that.

Sorry for taking up your time.

PR #86717, opened 2024-01-02
Add a python script to install Direct3D 12 SDK components.

mrbbbaixue:
Supersedes #86123, This is the python version for the previous PR.

- Now the default install path is `%LOCALAPPDATA%\Godot\build_deps\`
- The script will try to remove previous failed download files to avoid error.
- Will output the install folder after installation is completed.
- Now CI windows build will use this script to download SDK components.

This is how it works:
<details><summary>Details</summary>
<p>

```
[1/4] DirectX Shader Compiler
Downloading DirectX Shader Compiler dxc_2023_08_14.zip ...
Removing existing local DirectX Shader Compiler installation in C:\Users\USERNAME\AppData\Local\Godot\build_deps\dxc ...
Extracting DirectX Shader Compiler dxc_2023_08_14.zip to C:\Users\USERNAME\AppData\Local\Godot\build_deps\dxc ...
DirectX Shader Compiler dxc_2023_08_14.zip installed successfully.

[2/4] Mesa NIR
Downloading Mesa NIR godot-nir-23.1.0-1-devel.zip ...
Removing existing local Mesa NIR installation in C:\Users\USERNAME\AppData\Local\Godot\build_deps\mesa ...
Extracting Mesa NIR godot-nir-23.1.0-1-devel.zip to C:\Users\USERNAME\AppData\Local\Godot\build_deps\mesa ...
Mesa NIR godot-nir-23.1.0-1-devel.zip installed successfully.

[3/4] WinPixEventRuntime
Downloading WinPixEventRuntime 1.0.231030001 ...
Removing existing local WinPixEventRuntime installation in C:\Users\USERNAME\AppData\Local\Godot\build_deps\pix ...
Extracting WinPixEventRuntime 1.0.231030001 to C:\Users\USERNAME\AppData\Local\Godot\build_deps\pix ...
WinPixEventRuntime 1.0.231030001 installed successfully.

[4/4] DirectX 12 Agility SDK
Downloading DirectX 12 Agility SDK 1.610.4 ...
Removing existing local DirectX 12 Agility SDK installation in C:\Users\USERNAME\AppData\Local\Godot\build_deps\agility_sdk ...
Extracting DirectX 12 Agility SDK 1.610.4 to C:\Users\USERNAME\AppData\Local\Godot\build_deps\agility_sdk ...
DirectX 12 Agility SDK 1.610.4 installed successfully.

[...]

PR #87609, opened 2024-01-26
SCons: Fix Windows cross-compilation from Linux after #86717

akien-mga:
#86717 broke building Godot on Linux (for any platform) as long as mingw is installed, as the Windows `detect.py` would fail parsing its options due to `os.getenv("LOCALAPPDATA")` being `None` and not a string.

So I fixed it, adding support for actually installing deps on Linux too (and likely macOS). I chose to put them in `bin/build_deps`, since this is a more niche use case so I don't think they need to be cached system wide.

### DIFF
--- a/platform/windows/detect.py
+++ b/platform/windows/detect.py
@@ -164,8 +164,21 @@ def get_opts():
 
     mingw = os.getenv("MINGW_PREFIX", "")
 
-    # Direct3D 12 SDK dependencies folder
-    d3d12_deps_folder = os.path.join(os.getenv("LOCALAPPDATA"), "Godot", "build_deps")
+    # Direct3D 12 SDK dependencies folder.
+    d3d12_deps_folder = os.getenv("LOCALAPPDATA")
+    if d3d12_deps_folder:
+        d3d12_deps_folder = os.path.join(d3d12_deps_folder, "Godot", "build_deps")
+    else:
+        # Cross-compiling, the deps install script puts things in `bin`.
+        # Getting an absolute path to it is a bit hacky in Python.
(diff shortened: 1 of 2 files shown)


platform/windows/detect.py
```diff
@@ -164,8 +164,21 @@ def get_opts():
 
     mingw = os.getenv("MINGW_PREFIX", "")
 
-    # Direct3D 12 SDK dependencies folder
-    d3d12_deps_folder = os.path.join(os.getenv("LOCALAPPDATA"), "Godot", "build_deps")
+    # Direct3D 12 SDK dependencies folder.
+    d3d12_deps_folder = os.getenv("LOCALAPPDATA")
+    if d3d12_deps_folder:
+        d3d12_deps_folder = os.path.join(deps_folder, "Godot", "build_deps")
```
YuriSizov, line 170:
```suggestion
        d3d12_deps_folder = os.path.join(d3d12_deps_folder, "Godot", "build_deps")
```

Copy-paste error?

akien-mga: Yep.

akien-mga:
I've tested this successfully on Linux, cross-compiling to Windows with `d3d12=yes` enabled. The install script works, and the resulting binary also seems functional. So great, we properly support mingw for d3d12 builds :D

Would be good to have this PR checked briefly on Windows, at least making sure it starts building fine with `d3d12=yes`.

YuriSizov (approved):
Worked fine for me on Windows where I had no prerequisites installed prior to this. Both the new script and the `d3d12=yes` build with default params.

PR #87610, opened 2024-01-26
Mention Multiplayer.`get_remote_sender_id`'s 0 after await

### LINKED ISSUE #86719: multiplayer.get_remote_sender_id() returns 0 in an rpc after 'await'
### Tested versions

v4.2.1.stable.official.b09f793f5

### System information

Godot v4.2.1.stable - Ubuntu 22.04.3 LTS 22.04 - X11 - GLES3 (Compatibility) - Mesa Intel(R) Graphics (RKL GT1) () - 11th Gen Intel(R) Core(TM) i7-11700 @ 2.50GHz (16 Threads)

### Issue description

In multiplayer, during a remote procedure call (RPC), I noticed that multiplayer.get_remote_sender_id() gives inconsistent results after a line with the 'await' keyword (or at least results that are unexpected to me.) 
After a line with the await keyword, the result of multiplayer.get_remote_sender_id() is 0, instead of the actual remote sender ID. Before the line with the await keyword, it gives the sender ID just fine. It seems that after await, if forgets that it's in the rpc?

```gdscript
@rpc("any_peer", "call_remote", "reliable")
func example_rpc():
	print("1. Received from: ", multiplayer.get_remote_sender_id())  # An ID like: 134295262
	await get_tree().create_timer(0.1).timeout
	print("2. Received from: ", multiplayer.get_remote_sender_id())  # Always 0
```

Maybe this is the expected behavior, but in that case it should maybe be added to the documentation?

### Steps to reproduce

Start two instances of the MRP (below). In the first instance, click the server button (which starts a server). In the second, click the client button (which starts a client, connects to the server, and executes an RPC on the server). The RPC is also shown above. Before the line with the await keyword (which awaits a timer) the correct ID is shown, but after the await keyword the ID is incorrectly given as 0.

[MRP.zip](https://github.com/godotengine/godot/files/13812110/MRP.zip)

Mickeon:
Closes https://github.com/godotengine/godot/issues/86719

It does not hurt to be more explicit. I considered adding a visual example but this looks good enough.

### DIFF
--- a/doc/classes/MultiplayerAPI.xml
+++ b/doc/classes/MultiplayerAPI.xml
@@ -34,7 +34,7 @@
 			<return type="int" />
 			<description>
 				Returns the sender's peer ID for the RPC currently being executed.
-				[b]Note:[/b] If not inside an RPC this method will return 0.
+				[b]Note:[/b] This method returns [code]0[/code] when called outside of an RPC. As such, the original peer ID may be lost when code execution is delayed (such as with GDScript's [code]await[/code] keyword).
 			</description>
 		</method>
 		<method name="get_unique_id">


doc/classes/MultiplayerAPI.xml
```diff
@@ -34,7 +34,7 @@
 			<return type="int" />
 			<description>
 				Returns the sender's peer ID for the RPC currently being executed.
-				[b]Note:[/b] If not inside an RPC this method will return 0.
+				[b]Note:[/b] This method returns [code]0[/code] when called outside of an RPC. As such, the original peer ID may be lost when code execution is delayed (such as with GDScript's [code]await[/code] keyword).
```
Mickeon, line 37: Alternative take of the same line:
```suggestion
				[b]Note:[/b] This method returns [code]0[/code] when called outside of an RPC. Delaying code execution (such as with GDScript's [code]await[/code] keyword) may bring the RPC out of context.
```

mhilbrunner (approved):

doc/classes/MultiplayerAPI.xml
mhilbrunner: I prefer the current one.

akien-mga:
Thanks!

Issue #1405, opened 2015-02-20
"in" operator doesn't match int/float


Spooner:
1 in [1.0] => false
1.0 in [1] => false
They are both true in Python (and that is what I expect!).

reduz:
that's interesting, maybe something related to type conversion

PR #87613, opened 2024-01-26
GDExtension method calls should not pass nullptr args

### LINKED ISSUE #1056: Passing null to a virtual function call causes a null pointer dereference (GDExtension)
I've got a project that overrides `ScriptLanguageExtension::_complete_code`, one of whose parameters (the third one) is an `Object*` pointer. Godot passes null to this argument, which causes a fault as `Object::get_instance_binding` is ultimately called on this pointer:

```
// Object::get_instance_binding locks a mutex, which is in the zero-page as the object is a null pointer
#0  0x00007ffff7d53b04 in pthread_mutex_lock () from /usr/lib/libc.so.6
#1  0x0000555555a51e2f in __gthread_mutex_lock (__mutex=0x148) at /usr/include/c++/12.2.1/x86_64-pc-linux-gnu/bits/gthr-default.h:749
#2  std::mutex::lock (this=this@entry=0x148) at /usr/include/c++/12.2.1/bits/std_mutex.h:100
// Note this=0x0 - this is called on a null object
#3  0x00005555596d8f29 in Object::get_instance_binding (this=0x0, p_token=0x55555c70c2f0, p_callbacks=0x7ffff49f1690 <godot::Object::___binding_callbacks>) at core/object/object.cpp:1700
// The interface between godot-cpp and the engine, converting the object pointer
#4  0x00005555596ba202 in gdextension_object_get_instance_binding (p_object=<optimized out>, p_token=<optimized out>, p_callbacks=<optimized out>) at core/extension/gdextension_interface.cpp:928
#5  0x00007ffff4904b2b in godot::PtrToArg<godot::Object*>::convert (p_ptr=0x7ffffffface0) at /home/znix/static/programming/c++/gdwren/godot-cpp/include/godot_cpp/core/method_ptrcall.hpp:171
// The result of BIND_VIRTUAL_METHOD and it's associated calls
#6  godot::call_with_ptr_args_retc_helper<WrenLanguage, godot::Dictionary, godot::String const&, godot::String const&, godot::Object*, 0ul, 1ul, 2ul> (p_instance=0x55555fd52380, p_method=&virtual table offset 224, 
    p_args=0x7fffffffad40, r_ret=0x7ffffffface8) at /home/znix/static/programming/c++/gdwren/godot-cpp/include/godot_cpp/core/binder_common.hpp:209
#7  0x00007ffff4901112 in godot::call_with_ptr_args<WrenLanguage, godot::Dictionary, godot::String const&, godot::String const&, godot::Object*> (p_instance=0x55555fd52380, p_method=&virtual table offset 224, p_args=0x7fffffffad40, 
    r_ret=0x7ffffffface8) at /home/znix/static/programming/c++/gdwren/godot-cpp/include/godot_cpp/core/binder_common.hpp:229
#8  0x00007ffff48fb64d in godot::ScriptLanguageExtension::register_virtuals<WrenLanguage, godot::ScriptLanguageExtension>()::{lambda(void*, void const* const*, void*)#24}::operator()(void*, void const* const*, void*) const (
    __closure=0x0, p_instance=0x55555fd52380, p_args=0x7fffffffad40, p_ret=0x7ffffffface8) at /home/znix/static/programming/c++/gdwren/cmake-build-debug/godot-cpp/gen/include/godot_cpp/classes/script_language_extension.hpp:231
#9  0x00007ffff48fb680 in godot::ScriptLanguageExtension::register_virtuals<WrenLanguage, godot::ScriptLanguageExtension>()::{lambda(void*, void const* const*, void*)#24}::_FUN(void*, void const* const*, void*) ()
    at /home/znix/static/programming/c++/gdwren/cmake-build-debug/godot-cpp/gen/include/godot_cpp/classes/script_language_extension.hpp:231
// The in-engine caller
#10 0x00005555591ffc2e in ScriptLanguageExtension::_gdvirtual__complete_code_call<true> (r_ret=..., arg3=0x0, arg2=..., arg1=..., this=0x55555fd523c0) at ./core/object/script_language_extension.h:351
#11 ScriptLanguageExtension::complete_code (this=this@entry=0x55555fd523c0, p_code=..., p_path=..., p_owner=p_owner@entry=0x0, r_options=r_options@entry=0x7fffffffaea0, r_force=@0x7fffffffae9f: false, r_call_hint=...)
    at ./core/object/script_language_extension.h:355
```

The following diff did fix it, but it feels fairly hacky - I'm not sure if this is the right place for checking for null-ness, or if null is even supposed to be a valid value:

```diff
diff --git a/include/godot_cpp/core/method_ptrcall.hpp b/include/godot_cpp/core/method_ptrcall.hpp
index 6b092bb..e917925 100644
--- a/include/godot_cpp/core/method_ptrcall.hpp
+++ b/include/godot_cpp/core/method_ptrcall.hpp
@@ -168,7 +168,10 @@ MAKE_PTRARG_BY_REFERENCE(Variant);
 template <class T>
 struct PtrToArg<T *> {
        _FORCE_INLINE_ static T *convert(const void *p_ptr) {
-               return reinterpret_cast<T *>(godot::internal::gde_interface->object_get_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr)), godot::internal::token, &T::___binding_callbacks));
+               GDExtensionObjectPtr objPtr = *reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr));
+               if (objPtr == nullptr)
+                       return nullptr;
+               return reinterpret_cast<T *>(godot::internal::gde_interface->object_get_instance_binding(objPtr, godot::internal::token, &T::___binding_callbacks));
        }
        typedef Object *EncodeT;
        _FORCE_INLINE_ static void encode(T *p_var, void *p_ptr) {
@@ -179,7 +182,10 @@ struct PtrToArg<T *> {
 template <class T>
 struct PtrToArg<const T *> {
        _FORCE_INLINE_ static const T *convert(const void *p_ptr) {
-               return reinterpret_cast<const T *>(godot::internal::gde_interface->object_get_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr)), godot::internal::token, &T::___binding_callbacks));
+               GDExtensionObjectPtr objPtr = *reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr));
+               if (objPtr == nullptr)
+                       return nullptr;
+               return reinterpret_cast<const T *>(godot::internal::gde_interface->object_get_instance_binding(objPtr, godot::internal::token, &T::___binding_callbacks));
        }
        typedef const Object *EncodeT;
        _FORCE_INLINE_ static void encode(T *p_var, void *p_ptr) {
```

### LINKED ISSUE #86478: [GDExtension][Godot4.2] Passing typed variable with null value from GDScript to GDExtension's created function lead engine to silent crash
### Tested versions

- Reproducible in 4.2.1.stable

### System information

Godot v4.2.1.stable - Windows 10.0.19045 - Vulkan (Mobile) - dedicated NVIDIA GeForce GTX 650 (NVIDIA; 30.0.14.7464) - 13th Gen Intel(R) Core(TM) i5-13400 (16 Threads)

### Issue description

If variable was typed and null value was assigned to it, passing this variable to GDExtension created function lead engine to silent crash. This not happened if variable was not typed, or method created not by GDExtension.

GDExtension:
```
#include <gdextension_interface.h>
#include <godot_cpp/core/defs.hpp>
#include <godot_cpp/core/class_db.hpp>
#include <godot_cpp/godot.hpp>

namespace godot {
	class ExampleClass : public Object {
		GDCLASS(ExampleClass, Object);
		
	protected:
		static void _bind_methods() {
			ClassDB::bind_method(D_METHOD("test_method"), &ExampleClass::test_method);
		}
		
	public:
		void test_method(ExampleClass* pointer) {}
	};
	
	void extension_init(ModuleInitializationLevel p_level) {
		if (p_level != MODULE_INITIALIZATION_LEVEL_SCENE) { return; }
		ClassDB::register_class<ExampleClass>();
	}
	void extension_uninit(ModuleInitializationLevel p_level) {
		if (p_level != MODULE_INITIALIZATION_LEVEL_SCENE) { return; }
	}
	extern "C" GDExtensionBool GDE_EXPORT library_init(GDExtensionInterfaceGetProcAddress p_get_proc_address, const GDExtensionClassLibraryPtr p_library, GDExtensionInitialization *r_initialization) {
		GDExtensionBinding::InitObject init_obj(p_get_proc_address, p_library, r_initialization);

		init_obj.register_initializer(extension_init);
		init_obj.register_terminator(extension_uninit);
		init_obj.set_minimum_library_initialization_level(MODULE_INITIALIZATION_LEVEL_SCENE);

		return init_obj.init();
	}
};
```
GDScript:
```
extends Node

func _ready():
	var some : ExampleClass = ExampleClass.new()
	var target : ExampleClass = null
	some.test_method(target)

```

Crash not happening if var declared as:
`var target = null`

### Steps to reproduce

1. Create or import example project.

[project.zip](https://github.com/godotengine/godot/files/13760330/project.zip)

rune-scape:
fix for a nasty crash that happens when passing a null variant to any gdextension function parameter expecting any kind of Object *.
the GDExtensionMethodBind  expects a pointer to an `Object *` but `get_opaque_pointer` returns a nullptr which it tries to dereference..

could also be fixed in godot-cpp with a nullptr check, but this seemed more in line with what the binding code expects 

this or another fix should rlly be applied to 4.2, 4.1, nd 4.0
it can be easily cherrypicked for 4.2 but 4.1 and 4.0 have different binding code, ill make a PR for 4.1 when its apropriate 

fixes https://github.com/godotengine/godot/issues/86478
fixes https://github.com/godotengine/godot-cpp/issues/1056

### DIFF
--- a/core/extension/gdextension.cpp
+++ b/core/extension/gdextension.cpp
@@ -230,12 +230,14 @@ public:
 			const void **argptrs = (const void **)alloca(argument_count * sizeof(void *));
 			for (uint32_t i = 0; i < argument_count; i++) {
 				argptrs[i] = VariantInternal::get_opaque_pointer(p_args[i]);
+				ERR_FAIL_NULL(argptrs[i]);
 			}
 
 			void *ret_opaque = nullptr;
 			if (r_ret) {
 				VariantInternal::initialize(r_ret, return_value_info.type);
-				ret_opaque = r_ret->get_type() == Variant::NIL ? r_ret : VariantInternal::get_opaque_pointer(r_ret);
+				ret_opaque = VariantInternal::get_opaque_pointer(r_ret);
+				ERR_FAIL_NULL(ret_opaque);
 			}
 
 			ptrcall(p_object, argptrs, ret_opaque);
--- a/core/variant/variant_internal.h
+++ b/core/variant/variant_internal.h
@@ -341,10 +341,11 @@ public:
 		}
 	}
 
+	inline static void *static_nil_ptr = nullptr;
 	_FORCE_INLINE_ static void *get_opaque_pointer(Variant *v) {
 		switch (v->type) {
 			case Variant::NIL:
-				return nullptr;
+				return &static_nil_ptr;
 			case Variant::BOOL:
 				return get_bool(v);
 			case Variant::INT:
@@ -428,7 +429,7 @@ public:
 	_FORCE_INLINE_ static const void *get_opaque_pointer(const Variant *v) {
 		switch (v->type) {
 			case Variant::NIL:
-				return nullptr;
+				return &static_nil_ptr;
 			case Variant::BOOL:
 				return get_bool(v);
 			case Variant::INT:


dsnopek:
Hm. I'll do some more "historical" investigation when I have a chance, but if we've really been encoding null objects this way in 4.0, 4.1, and 4.2, I'd be reluctant to change this in Godot, because if it's been consistent, this sounds more like a godot-cpp bug.

rune-scape:
pretty sure a bug in godot itsself,, i mean it doesnt even reach the gdextension dll,,
it seems like a clash between what the VariantInternal stuff used to do and what the GDExtensionMethodBind expects when making a __validated call__,, in 4.1 the crash happens in the extension dll instead
i guess the ptr to nullptr could be a ptr to the variant itsself like it is here:
https://github.com/godotengine/godot/blob/8fb744ac230a3896bf354e3a82760a94dccde024/core/extension/gdextension.cpp#L239
but only if a null variant is guaranteed  to be zeroed out (  ^ as long as ret_opaque points to valid memory it shouldnt really matter what it is bc its memory for a null return type, not getting read or written to, just maybe ^ )
and i cant look thru every file to check..
so this solution seemed pretty good

rune-scape:
hmmm, the failed check is a cancelled ios build,,,

dsnopek:
Sorry for taking so long to get back to this!

I didn't do any testing, but I skimmed the code from 4.0, 4.1 and 4.2, and it does appear that we've been encoding null `Object *` as `nullptr` in ptrcalls through all of those versions.

So, I'm personally against making this change in Godot. I think consistency with previous versions is important unless there is a clear advantage to changing the encoding, and I don't think there is in this case (in fact, I think the current encoding is slightly better because we don't have to dereference the `Object **` to check for `nullptr` - it's just `nullptr`). While this change is unlikely to break anything, it's likely that all bindings other than godot-cpp are already doing a correct null check, and changing the encoding is going to make the extra check that they're doing redundant, but they probably won't remove it in order to retain compatibility with older Godot versions.

I think the crash should be fixed in godot-cpp instead

rune-scape:
> it does appear that we've been encoding null `Object *` as `nullptr` in ptrcalls through all of those versions.
> So, I'm personally against making this change in Godot. I think consistency with previous versions is important unless there is a clear advantage to changing the encoding

i dont think this comment makes sense, this is a crash that happens when a validated call is made and a `Nil` value is encoded as an `Object **` , a `Nil` value can only be passed to an `Object *` parameter, builtin types cannot be null  ,,  its been consistently broken in all of gdextension this has to be fixed in Every version ,, there is no consistency to follow , that doesnt mean it was intentional, and i dont think this is 'changing the encoding' i think its following it more closely

> While this change is unlikely to break anything, it's likely that all bindings other than godot-cpp are already doing a correct null check, and changing the encoding is going to make the extra check that they're doing redundant, but they probably won't remove it in order to retain compatibility with older Godot versions.

it actually seems like D bindings Dont make this check,
https://github.com/godot-dlang/godot-dlang/blob/ff421674d9f775f9cc88e57a6ab625b172b97b4f/src/godot/api/wrap.d#L356
but godot-rust does,,, if im interpreting this correctly
https://github.com/godot-rust/gdext/blob/88d8ed0719eb70d2a5af370e9b0fa672acea032a/godot-core/src/obj/raw.rs#L464C12-L464C15

since its a validated call, i doubt a null check should be done ,, it would need to get added in a couple places, it seems unlikely that someone just forgot that, i was interpreting the `GDExtensionObjectPtr *` or `Object **` as 'a non null pointer to a godot `Object *`' ,, like a reference, but is a C api, because all other builtin type pointers dont have a null check ,, none of the validated arg to pointer conversion stuff have a null check ,, but its not written out anywhere ,, just gotta ask whoever made this i guess ,,, vnen and/or reduz

rune-scape:
@vnen i see your name and reduz in the commit log for gdextension validated ptrcall ,, do you have insight about this?
are `GDExtensionObjectPtr *`s meant to be nullable when making a validated call?
should this change be in godot-cpp instead?

dsnopek:
Here's alternative PR https://github.com/godotengine/godot-cpp/pull/1405 which fixes the issue in godot-cpp rather than Godot itself.

I've also added discussion of this issue to the agenda of the next GDExtension meeting. (If anyone is interested in participating, please stop by #gdextension on chat.godotengine.org for details about the GDExtension meetings.)

dsnopek:
We discussed this at the GDExtension meeting, and felt that it would be great if @Bromeon (or other maintainers of GDExtension bindings) could weigh in on the current encoding of `Object *`. It would also be nice to hear from @vnen or @reduz on what the original intention of the encoding was. Mainly, it's a question of if the current encoding is "incorrect" and should be fixed, or if the encoding is OK (since it has at least been consistent) and godot-cpp should be updated to correctly handle it.

lilizoey:
In godot rust, null objects in argument calls are passed as pointers to null object pointers in all version of godot except 4.0. In 4.0 we pass them directly as null pointers.

However we also dont use null objects directly all that frequently (im unsure if we ever do actually) when calling gdextension methods since we have an [outstanding bug](https://github.com/godot-rust/gdext/issues/156) where direct method calls only allow for non-null objects. The current workaround there is to use `Object.call` instead which means we then pass null variants instead of null objects. So it may be that we're doing it wrong and just haven't noticed because we actually dont ever do it.

I dont think it would really impact much if we end up needing to switch to encoding them as null pointers instead.

dsnopek:
_(Adding this comment just to differentiate my personal opinion from what I wrote in my [previous comment](https://github.com/godotengine/godot/pull/87613#issuecomment-1982160546), which was the result of that GDExtension meeting, and isn't the same as my personal view.)_

As I've stated previously, it's my opinion that we don't really gain anything by changing the encoding as proposed in this PR: both the current encoding and the proposed encoding are capable of representing a null `Object *` just fine.

The only advantage to changing the encoding is theoretical "correctness", if the proposed encoding more closely matches the original intention for the encoding (however, it's not totally clear what the original intention was).

My personal vote is still for keeping the current encoding. Regardless of the intention, it's the encoding we've been using in all versions of Godot 4.x, and it's capable of representing what we need it to, so I see no compelling reason to change it.

But we certainly could improve the overall situation by documenting the encoding, so that it's clearer to the implementors of GDExtension bindings.

Bromeon:
> As I've stated previously, it's my opinion that we don't really gain anything by changing the encoding as proposed in this PR: both the current encoding and the proposed encoding are capable of representing a null `Object *` just fine.
> 

> My personal vote is still for keeping the current encoding. Regardless of the intention, it's the encoding we've been using in all versions of Godot 4.x, and it's capable of representing what we need it to, so I see no compelling reason to change it.

That is reasonable -- it would also prevent bindings from possibly needing to deal with different behavior across multiple API versions. As such, I'm with you in light of "don't fix what's not broken".

reduz (dismissed):
Seems good, great catch.

reduz:
Sorry I though this was a fix, so adding to the discussion: Regarding the intention of the encoding, I think (if I am not mistaken) it was that it passed a pointer to a pointer for returning, and a single pointer for argument passing. The problem is that this caused a lot of issues so I think at some point it was changed?

I honestly would use what work best for you, since you are the consumers of the API at this point.

reduz:
I checked #1405 and this looks like a better solution I guess, since each platform can handle it as preferred?

dsnopek:
Superseded by PR https://github.com/godotengine/godot-cpp/pull/1405

PR #87615, opened 2024-01-26
[Editor] Allow selecting instanced scenes with select list

### LINKED ISSUE #87607: Alt+RMB in Select mode does not show anything
### Tested versions

Tested in v4.2.stable.official [46dc27791]

### System information

Windows 10

### Issue description

When using the Select Mode (via Q) and pressing Alt+RMB in a 3D scene, the promised function "Show list of all nodes at position clicked" does not work - no list is shown. It works in 2D.

Screenshot_7

### Steps to reproduce

See issue description. If the function is not supposed to work in 3D environments, then the description should reflect that.

N/A

AThousandShips:
Picks the owning instance as is done in other selection methods

Unsure if this change is desired, and not familiar with the exact code here, so need some eyes on it to verify nothing is wrong, but it works correctly in my testing, only selects the parent once if there's multiple nodes to pick inside it

* Closes: https://github.com/godotengine/godot/issues/87607

### DIFF
--- a/editor/scene/3d/node_3d_editor_plugin.cpp
+++ b/editor/scene/3d/node_3d_editor_plugin.cpp
@@ -928,16 +928,27 @@ void Node3DEditorViewport::_find_items_at_pos(const Point2 &p_pos, Vector<_RayRe
 
 	HashSet<Node3D *> found_nodes;
 
+	Node *edited_scene = get_tree()->get_edited_scene_root();
+
 	for (Node3D *spat : nodes_with_gizmos) {
 		if (!spat) {
 			continue;
 		}
 
-		if (found_nodes.has(spat)) {
+		Node3D *item = spat;
+		if (item != edited_scene) {
+			item = Object::cast_to<Node3D>(edited_scene->get_deepest_editable_node(item));
+		}
+
+		if (!item) {
 			continue;
 		}
 
-		if (!p_include_locked_nodes && _is_node_locked(spat)) {
+		if (found_nodes.has(item)) {
+			continue;
+		}
+
+		if (!p_include_locked_nodes && _is_node_locked(item)) {
 			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -1549,10 +1560,9 @@ void Node3DEditorViewport::_list_select(Ref<InputEventMouseButton> b) {
 			if (_is_node_locked(spat)) {
 				locked = 1;
 			} else {
-				Node *ed_scene = EditorNode::get_singleton()->get_edited_scene();
```
AThousandShips, line 1552: This was already fetched above so removed unnecessary duplicate

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -6500,20 +6510,20 @@ void vertex() {
 	// Points are already in world space, so no need for MODEL_MATRIX anymore.
 	vec4 clip_a = PROJECTION_MATRIX * (VIEW_MATRIX * vec4(point_a, 1.0));
 	vec4 clip_b = PROJECTION_MATRIX * (VIEW_MATRIX * vec4(point_b, 1.0));
-	
+
```
AThousandShips, line 6513: Automatic trailing space removal, can revert if desired but had difficulties changing my editor to prevent it

AThousandShips:
Unsure if it should be considered a fix or an enhancement, so marking as the latter for now, for parity with the other methods of selection it would be an exception, and in that case a bug, it's also confusing and counter-intuitive

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -831,18 +831,29 @@ void Node3DEditorViewport::_find_items_at_pos(const Point2 &p_pos, Vector<_RayRe
...
 
 		if (!spat) {
 			continue;
 		}
 
-		if (found_nodes.has(spat)) {
+		Node3D *item = spat;
+		if (item != edited_scene) {
+			item = Object::cast_to<Node3D>(edited_scene->get_deepest_editable_node(item));
+		}
```
AThousandShips, line 846: Some of the other methods handle group stuff here, but `_list_select` has its own group processing code so not touching it here, unsure what's best here, leaving as is for now

PR #87616, opened 2024-01-26
Add cancel transformation shortcut to 2D to match 3D

ryevdokimov:
Currently the only way to cancel a transformation in 2D is with a right-click. This PR adds a shortcut 'Cancel Transformation' to allow users to use the escape key by default to cancel a transformation and to change it if they desire.  

image


KoBeWi (approved):
Looks fine consistency-wise.

akien-mga:
Thanks!

PR #87621, opened 2024-01-26
Add transform aborted message to 2D matching 3D

ryevdokimov:
Related to: #87616

It's in 3D and I think it's useful.

https://github.com/godotengine/godot/assets/105675984/5b06f562-68f3-4bc4-980c-862159e988d9


KoBeWi:
This adds a lot of code for a trivial feature that IMO isn't that useful.
I think it needs a proposal.

ryevdokimov:
It probably could be reduced if some of the functionalities were hard-coded, but I tried to make the code similar to how it was written in 3D and to keep it flexible in case more messages on a timer in the future were required. Fair enough though. 

Added: [https://github.com/godotengine/godot-proposals/issues/8977](https://github.com/godotengine/godot-proposals/issues/8977)

PR #87622, opened 2024-01-26
Add a scan_importer_support check that was missing

mxaddict:
This PR is to move the changes to scan_import_support from https://github.com/godotengine/godot/pull/85448

This PR fixes a use case where you open a project that contains .blend or .fbx files in it, but had import support off.

So it will check for import support for those assets before attempting import.

If you have don't have import support for those assets configured and attempt to import it will start the wizard to auto detect the path for either/or blend/fbx files.

### DIFF
--- a/editor/editor_file_system.cpp
+++ b/editor/editor_file_system.cpp
@@ -524,7 +524,7 @@ bool EditorFileSystem::_test_for_reimport(const String &p_path, bool p_only_impo
 	return false; //nothing changed
 }
 
-bool EditorFileSystem::_scan_import_support(Vector<String> reimports) {
+bool EditorFileSystem::scan_import_support(Vector<String> p_reimports) {
 	if (import_support_queries.size() == 0) {
 		return false;
 	}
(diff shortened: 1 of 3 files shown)


editor/editor_file_system.h
```diff
@@ -316,6 +315,7 @@ class EditorFileSystem : public Node {
 	void scan();
 	void scan_changes();
 	void update_file(const String &p_file);
+	bool scan_import_support(Vector<String> reimports);
```
AThousandShips, line 318:
```suggestion
	bool scan_import_support(Vector<String> p_reimports);
```

editor/editor_file_system.cpp
```diff
@@ -524,7 +524,7 @@ bool EditorFileSystem::_test_for_reimport(const String &p_path, bool p_only_impo
 	return false; //nothing changed
 }
 
-bool EditorFileSystem::_scan_import_support(Vector<String> reimports) {
+bool EditorFileSystem::scan_import_support(Vector<String> reimports) {
```
AThousandShips, line 527:
```suggestion
bool EditorFileSystem::scan_import_support(Vector<String> p_reimports) {
```

fire:
I haven't ran this `scan_importer_support` check, but it doesn't break --headless mode right?

mxaddict:
> I haven't ran this `scan_importer_support` check, but it doesn't break --headless mode right?

Not sure, how do I run a headless mode to check?

I've not used it before.

mxaddict:
> I haven't ran this `scan_importer_support` check, but it doesn't break --headless mode right?

Could you point me in the direction of how to check this? 😄

mxaddict:
I'm not sure what the next step for this PR is, so gonna close for now :)

PR #51011, opened 2021-07-29
Make drag and drop into viewport add to root node by default

foxydevloper:
Closes: https://github.com/godotengine/godot-proposals/issues/2381
Implements the behavior suggested in https://github.com/godotengine/godot-proposals/issues/2381#issuecomment-787539659

This PR makes dragging and dropping a texture, mesh, or scene from the FileSystem into the 2D or 3D viewport add the node as a child of the current scene's root node by default. Holding Ctrl when dropping will add the node as a child of the selected node instead.

This also improves the tooltips in the 2D editor that show the modifier keys when dragging in a new node.
It also makes the tooltips appear when dragging in a scene in the 2D editor, not just when dragging in an image.
image

PR #87623, opened 2024-01-26
Change scene drag-and-drop modifier behavior in the 2D and 3D editors

### LINKED ISSUE #80795: Dragging an object does not respect selection.
### Godot version

3.6 Beta 3

### System information

Ubuntu 22

### Issue description

When I drag an object from the file system of the engine, the selected object is ignored, it is added to the scene end.

In Godot 3.5.2, everything works correctly, when you drag an object, it becomes a child of the selected object.

This is how it worked before (3.5.2)

https://github.com/godotengine/godot/assets/13809475/fc951a11-33f4-4c08-b02b-f595cbb4375d

This is how it works now (3.6)

https://github.com/godotengine/godot/assets/13809475/fb88e5c1-289e-4fa9-987c-e6c1acfcacd2

### Steps to reproduce

Drag an object (in 3D, *.mesh) into the workspace from the engine's file system.

Does not trumpet, because works in any project.

Calinou:
Behavior is now consistent across 2D and 3D editors.

- By default, drag-and-dropping adds the scene as a child of the selected node. Hold <kbd>Alt</kbd> when releasing the mouse to add the scene as a child of the root node, or <kbd>Shift</kbd> to add the scene as a sibling of the selected node.
- To choose a different node type in the 2D editor when drag-and-dropping a texture resource, hold <kbd>Alt + Shift</kbd> (instead of just <kbd>Alt</kbd>).
- If multiple nodes are selected, only the first one is taken into account. This was previously the behavior in 2D, but not in 3D (an error dialog appeared instead). This makes the UI more forgiving.

<kbd>Ctrl</kbd> is no longer used in shortcuts because it interferes with the grid snapping toggle. Using <kbd>Alt</kbd> in shortcuts here is fine, because it's checked when the mouse is *released*, not when it's pressed. Therefore, it'll work correctly on Linux with window manager defaults.

___

- This closes https://github.com/godotengine/godot/issues/80795.

## Preview

*I hold <kbd>Alt</kbd> when performing the last drag-and-drop operation, which is why the last node is added to the root node instead.*

https://github.com/godotengine/godot/assets/180032/0223ef5b-6264-4e4c-8358-d863a4b7f036

### DIFF
--- a/editor/plugins/canvas_item_editor_plugin.cpp
+++ b/editor/plugins/canvas_item_editor_plugin.cpp
@@ -5659,7 +5659,7 @@ void CanvasItemEditorViewport::_create_preview(const Vector<String> &files) cons
 		Ref<PackedScene> scene = Ref<PackedScene>(Object::cast_to<PackedScene>(*res));
 		if (texture != nullptr || scene != nullptr) {
 			bool root_node_selected = EditorNode::get_singleton()->get_editor_selection()->is_selected(EditorNode::get_singleton()->get_edited_scene());
-			String desc = TTR("Drag and drop to add as child of current scene's root node.") + "\n" + vformat(TTR("Hold %s when dropping to add as child of selected node."), keycode_get_string((Key)KeyModifierMask::CMD_OR_CTRL));
+			String desc = TTR("Drag and drop to add as child of selected node.") + "\n" + TTR("Hold Alt when dropping to add as child of root node.");
 			if (!root_node_selected) {
 				desc += "\n" + TTR("Hold Shift when dropping to add as sibling of selected node.");
 			}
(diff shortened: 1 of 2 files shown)


ryevdokimov:
#87126 is related.

ryevdokimov:
Tested every scenario in the description. Seems to work.

ryevdokimov (approved):

KoBeWi:
For 2D, the current behavior was implemented in #51011
Although before that there was no way to control where the node is added, so changing the default is fine.

KoBeWi:
Using Alt+Shift for changing type makes it impossible to change type and add as sibling.

KoBeWi (approved):
Looks fine overall.

The Shift+Alt conflict I mentioned is a corner-case, I'm not sure what modifier we could use to deconflict it. Maybe it's fine this way.

akien-mga:
Thanks!

PR #87625, opened 2024-01-26
Generate script resource preview without parsing

YuriSizov:
During the last days of development of Godot 4.2 we were fighting a lot with deadlocks during the editor startup.

One of the sources of these deadlocks was the editor resource preview generator. Specifically, the one implementation responsible for generating previews for script files. Specifically specifically, GDScript files. The cause for this issue was in the way GDScript files resolve their dependencies when loaded.

Essentially, loading a script containing references to other resources, such as preloads of packed scenes, will trigger a load for those resources as well. And to generate a preview of a script we would load it completely and fully. Which means attempting to generate a preview of a script may trigger a load of a packed scene, which may need a preview of its own, which will deadlock the generator.

The sad part is that we don't need a fully loaded script. We only take the source code of the script (plus a reference to its `ScriptLanguage`). This is something we can do by simply reading the file as text.

Which is what this PR does. It makes sure that instead of loading a script resource we just read it as text and use the file's extension to get the corresponding `ScriptLanguage`, cheap and simple. The rest is the same as before.

There is no noticeable performance difference, unfortunately. At least not with an example I could fabricate to test this with. YMMV.

```
Before:

Generated 'res://OneBadScript.gd' preview in 5148 usec
Generated 'res://OneBadScript.gd' preview in 4171 usec
Generated 'res://OneBadScript.gd' preview in 5058 usec
Generated 'res://OneBadScript.gd' preview in 4574 usec

After:

Generated 'res://OneBadScript.gd' preview in 5484 usec
Generated 'res://OneBadScript.gd' preview in 4343 usec
Generated 'res://OneBadScript.gd' preview in 4933 usec
Generated 'res://OneBadScript.gd' preview in 4096 usec

```
_Some deviations are possible depending on how busy the editor is in each particular moment._

There is no visible difference in the generated file either:

**Before**
before
**After**
after

----

I also added a small benchmark for the generator to test the timings. I decided to keep it for this PR, but as a separate commit and as verbose prints. If this is excessive I can drop it. Note that I couldn't use `print_verbose` here because `print_verbose` is undefined by the included `variant_utility.h` header. Probably something worth addressing?

### DIFF
--- a/editor/editor_resource_preview.cpp
+++ b/editor/editor_resource_preview.cpp
@@ -129,6 +129,8 @@ void EditorResourcePreview::_preview_ready(const String &p_path, int p_hash, con
 void EditorResourcePreview::_generate_preview(Ref<ImageTexture> &r_texture, Ref<ImageTexture> &r_small_texture, const QueueItem &p_item, const String &cache_base, Dictionary &p_metadata) {
 	String type;
 
+	uint64_t started_at = OS::get_singleton()->get_ticks_usec();
+
 	if (p_item.resource.is_valid()) {
 		type = p_item.resource->get_class();
 	} else {
@@ -138,6 +140,10 @@ void EditorResourcePreview::_generate_preview(Ref<ImageTexture> &r_texture, Ref<
 	if (type.is_empty()) {
 		r_texture = Ref<ImageTexture>();
 		r_small_texture = Ref<ImageTexture>();
+
+		if (is_print_verbose_enabled()) {
+			print_line(vformat("Generated '%s' preview in %d usec", p_item.path, OS::get_singleton()->get_ticks_usec() - started_at));
+		}
 		return; //could not guess type
 	}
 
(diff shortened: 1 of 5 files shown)


editor/editor_resource_preview.cpp
```diff
@@ -138,6 +140,10 @@ void EditorResourcePreview::_generate_preview(Ref<ImageTexture> &r_texture, Ref<
 	if (type.is_empty()) {
 		r_texture = Ref<ImageTexture>();
 		r_small_texture = Ref<ImageTexture>();
+
+		if (is_print_verbose_enabled()) {
+			print_line(vformat("Generated '%s' preview in %d usec", p_item.path, OS::get_singleton()->get_ticks_usec() - started_at));
+		}
```
akien-mga, line 146: We have `print_verbose` for this:
```suggestion
		print_verbose(vformat("Generated '%s' preview in %d usec", p_item.path, OS::get_singleton()->get_ticks_usec() - started_at));
```

editor/editor_resource_preview.cpp
```diff
@@ -196,6 +202,10 @@ void EditorResourcePreview::_generate_preview(Ref<ImageTexture> &r_texture, Ref<
 			_write_preview_cache(f, thumbnail_size, has_small_texture, FileAccess::get_modified_time(p_item.path), FileAccess::get_md5(p_item.path), p_metadata);
 		}
 	}
+
+	if (is_print_verbose_enabled()) {
+		print_line(vformat("Generated '%s' preview in %d usec", p_item.path, OS::get_singleton()->get_ticks_usec() - started_at));
+	}
```
akien-mga, line 208:
```suggestion
	print_verbose(vformat("Generated '%s' preview in %d usec", p_item.path, OS::get_singleton()->get_ticks_usec() - started_at));
```

akien-mga (approved):
Looks great to me. A much welcome improvement!

editor/editor_resource_preview.cpp
akien-mga: > Note that I couldn't use `print_verbose` here because `print_verbose` is undefined by the included `variant_utility.h` header. Probably something worth addressing?

I had missed this comment. So indeed this can't be done for now.

akien-mga:
Thanks!

PR #87627, opened 2024-01-26
Undefine yet another macro from Windows headers

RandomShaper:
For some reason, `MONO_FONT` from Win32 has started conflicting with `rich_text_label.h`'s `DefaultFont::MONO_FONT` for me.

So I'd like to get rid of the issue.

### DIFF
--- a/core/typedefs.h
+++ b/core/typedefs.h
@@ -93,6 +93,7 @@
 #undef OK
 #undef CONNECT_DEFERRED // override from Windows SDK, clashes with Object enum
 #undef MemoryBarrier
+#undef MONO_FONT
 #endif
 
 // Make room for our constexpr's below by overriding potential system-specific macros.


akien-mga:
Someone please go back to 1985 and force MS to use namespaces for all their generic and completely random defines...

akien-mga (approved):

fire (approved):

lawnjelly:
Just for information, many projects wrap windows headers to prevent this kind of namespace collision (and increased compile time), so the `include <windows.h>` is confined to a single cpp file.

Sometimes this is not possible, but 99% of the common defines can be safely redefined locally (`FILE` is a common example if I remember right, it's a while since I did any windows).

akien-mga:
> Just for information, many projects wrap windows headers to prevent this kind of namespace collision (and increased compile time), so the `include <windows.h>` is confined to a single cpp file.

The main difficulty with this approach is that we're not the only users of the code, and folks who have custom forks or C++ modules which consume our API may not have the same discipline. So as far as possible, we need to strive to make our code compile despite too early includes of `windows.h`.

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

PR #87628, opened 2024-01-26
Improve error reporting in the asset library and in related types

YuriSizov:
Right now opening asset library, both in the project manager and in the editor, may result in a barrage of error messages. All of them are related to various problems with assets listed on the frontend. But that's not really relevant to end users. And for people who may find some relevance in these messages, they remain obscure and and confusing; require debugging the engine to actually understand.

So this PR tries to improve related error messages, adds early checks to avoid invalid calls to internal APIs, and hides some of the messages behind the `--verbose` flag.

This is what the output can look like:

```
## PM

editor\plugins\asset_library_editor_plugin.cpp:872 - Asset Library: Error getting image from 'https://raw.githubusercontent.com/ShatReal/drag-drop-inventory/main/icon.png' for asset # 2026.
WARNING: Asset Library: Error getting image from 'https://raw.githubusercontent.com/ShatReal/drag-drop-inventory/main/icon.png' for asset # 2026.
     at: EditorAssetLibrary::_image_request_completed (editor\plugins\asset_library_editor_plugin.cpp:872)
WARNING: Asset Library: Error getting image from 'https://raw.githubusercontent.com/ShatReal/Infinite-Save-System-4/master/icon.png' for asset # 2380.
     at: EditorAssetLibrary::_image_request_completed (editor\plugins\asset_library_editor_plugin.cpp:872)
editor\plugins\asset_library_editor_plugin.cpp:872 - Asset Library: Error getting image from 'https://raw.githubusercontent.com/ShatReal/Infinite-Save-System-4/master/icon.png' for asset # 2380.
ERROR: ImageLoaderSVG: Failed to create SVG from buffer, error code 30.
   at: (modules\svg\image_loader_svg.cpp:75)
ImageLoaderSVG: Failed to create SVG from buffer, error code 30.
ERROR: Asset Library: Invalid image downloaded from 'https://gitee.com/L_HM/backpack_grid/blob/master/icon.png' for asset # 2363
   at: EditorAssetLibrary::_image_update (editor\plugins\asset_library_editor_plugin.cpp:799)
editor\plugins\asset_library_editor_plugin.cpp:799 - Asset Library: Invalid image downloaded from 'https://gitee.com/L_HM/backpack_grid/blob/master/icon.png' for asset # 2363

## Editor

editor\plugins\asset_library_editor_plugin.cpp:927 - Asset Library: Badly formatted image URL '	https://raw.githubusercontent.com/Daylily-Zeleen/GDScript-Formatter/main/icon.png' for asset # 2540.
```

The latter for example currently manifests as

```
ERROR: Invalid URL scheme: 	https://.
   at: (scene\main\http_request.cpp:57)
Invalid URL scheme: 	https://.
```

-----

Some of the touched up error messages are outside of the scope for the asset library specifically, but the improvements should still be welcome. Namely, we seem to have this beauty across the codebase: "It's not a reference to a valid ... object." And despite the claim of the commit message that added it, it's not actually an obvious error, especially if repeated 8 times verbatim.

PS. Changes to `EditorAssetLibrary::_image_update` include whitespace and nesting removal. They aren't as scary as they look, use the whitespace off mode when reviewing 🙃

### DIFF
--- a/editor/plugins/asset_library_editor_plugin.cpp
+++ b/editor/plugins/asset_library_editor_plugin.cpp
@@ -745,86 +745,97 @@ void EditorAssetLibrary::_select_asset(int p_id) {
 	_api_request("asset/" + itos(p_id), REQUESTING_ASSET);
 }
 
-void EditorAssetLibrary::_image_update(bool use_cache, bool final, const PackedByteArray &p_data, int p_queue_id) {
+void EditorAssetLibrary::_image_update(bool p_use_cache, bool p_final, const PackedByteArray &p_data, int p_queue_id) {
 	Object *obj = ObjectDB::get_instance(image_queue[p_queue_id].target);
+	if (!obj) {
+		return;
+	}
 
-	if (obj) {
-		bool image_set = false;
-		PackedByteArray image_data = p_data;
+	bool image_set = false;
+	PackedByteArray image_data = p_data;
 
-		if (use_cache) {
-			String cache_filename_base = EditorPaths::get_singleton()->get_cache_dir().path_join("assetimage_" + image_queue[p_queue_id].image_url.md5_text());
+	if (p_use_cache) {
+		String cache_filename_base = EditorPaths::get_singleton()->get_cache_dir().path_join("assetimage_" + image_queue[p_queue_id].image_url.md5_text());
 
-			Ref<FileAccess> file = FileAccess::open(cache_filename_base + ".data", FileAccess::READ);
-			if (file.is_valid()) {
-				PackedByteArray cached_data;
-				int len = file->get_32();
-				cached_data.resize(len);
+		Ref<FileAccess> file = FileAccess::open(cache_filename_base + ".data", FileAccess::READ);
+		if (file.is_valid()) {
+			PackedByteArray cached_data;
+			int len = file->get_32();
+			cached_data.resize(len);
 
-				uint8_t *w = cached_data.ptrw();
-				file->get_buffer(w, len);
+			uint8_t *w = cached_data.ptrw();
+			file->get_buffer(w, len);
 
-				image_data = cached_data;
+			image_data = cached_data;
+		}
+	}
+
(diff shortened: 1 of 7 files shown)


editor/plugins/asset_library_editor_plugin.cpp
```diff
@@ -857,7 +868,10 @@ void EditorAssetLibrary::_image_request_completed(int p_status, int p_code, cons
 		_image_update(p_code == HTTPClient::RESPONSE_NOT_MODIFIED, true, p_data, p_queue_id);
 
 	} else {
-		WARN_PRINT("Error getting image file from URL: " + image_queue[p_queue_id].image_url);
+		if (is_print_verbose_enabled()) {
+			WARN_PRINT(vformat("Asset Library: Error getting image from '%s' for asset # %d.", image_queue[p_queue_id].image_url, image_queue[p_queue_id].asset_id));
+		}
```
AThousandShips, line 873: Makes me think we want to add a dedicated macro for this, a `WARN_PRINT_VERBOSE` or something

YuriSizov: So far this is very situational, and normally we want all errors and warnings to be always printed. But yes, I was thinking about this as well.

AThousandShips: We have a `WARN_PRINT_ONCE` so a macro for `WARN_PRINT_VERBOSE` could be adapted to that too, but good to discuss in the future

YuriSizov: 👍

The reason why I mentioned a dedicated local method is because I want `WARN_PRINT_ONCE` to be called once for all warnings from all other methods. Since `WARN_PRINT_ONCE` works based on a scope, I need to create a funnel for all messages.

akien-mga (approved):
Looks great!

akien-mga:
Thanks!

PR #87630, opened 2024-01-26
OpenXR: Allow moving vendor passthrough extensions to GDExtension

dsnopek:
This PR is implementing an idea from @BastiaanOlij in [this comment](https://github.com/godotengine/godot/pull/86407#issuecomment-1880397139):

> One of the options I was thinking off is to make this purely work the way OpenXR intended. Eg:
> 
> * we deprecate the [is_passthrough_enabled](https://docs.godotengine.org/en/stable/classes/class_xrinterface.html#class-xrinterface-method-is-passthrough-enabled), [is_passthrough_supported](https://docs.godotengine.org/en/stable/classes/class_xrinterface.html#class-xrinterface-method-is-passthrough-supported), [start_passthrough](https://docs.godotengine.org/en/stable/classes/class_xrinterface.html#class-xrinterface-method-start-passthrough) and [stop_passthrough](https://docs.godotengine.org/en/stable/classes/class_xrinterface.html#class-xrinterface-method-stop-passthrough) methods in XRInterface
> * we make it clear in the documentation that [get_supported_environment_blend_modes](https://docs.godotengine.org/en/stable/classes/class_xrinterface.html#class-xrinterface-method-get-supported-environment-blend-modes) returns the available modes and what they mean
> * we make it clear in the documentation that [set_environment_blend_mode](https://docs.godotengine.org/en/stable/classes/class_xrinterface.html#class-xrinterface-method-set-environment-blend-mode) should be used for passthrough/AR
> 
> Now in the vendor plugin, for any vendor that requires additional logic, we check what environment blend mode is set to and perform the additional logic.
> 
> We could add the ability to register a callback or emit a signal when set_environment_blend_mode changes the blend mode to make starting and stopping passthrough easier for those runtimes that require that.
> 
> To keep backwards compatibility we can:
> 
> * make sure that `is_passthrough_enabled` and `is_passthrough_supported` return true if `XR_ENV_BLEND_MODE_ALPHA_BLEND` is returned by `get_supported_environment_blend_modes`
> * have `start_passthrough` set our environment blend mode to `XR_ENV_BLEND_MODE_ALPHA_BLEND`
> * have `stop_passthrough` set our environment blend mode to `XR_ENV_BLEND_MODE_OPAQUE`

[This PR](https://github.com/GodotVR/godot_openxr_vendors/pull/79) moves `XR_FB_passthrough` to 'godot_openxr_vendors', which allows this PR to be tested on the Meta Quest (it's working for me!)

The one thing that's a little messy in this PR is all the stuff with `emulate_environment_blend_mode_alpha_blend`, which I needed to add, because the Meta Quest (which I'm testing on) doesn't actually support `XR_ENV_BLEND_MODE_ALPHA_BLEND`, so if we try to set it to that, it'll just refuse. I may try to think of a better way to implement this bit - I'm open to suggestions as well!

Please let me know what you think :-)

**note to merge team** this should not be merged until both this PR, and the vendor plugins PR mentioned up above are both finished and approved. They should be merged in unison.


modules/openxr/openxr_api.cpp
```diff
@@ -3095,11 +3094,21 @@ bool OpenXRAPI::is_environment_blend_mode_supported(XrEnvironmentBlendMode p_ble
 }
 
 bool OpenXRAPI::set_environment_blend_mode(XrEnvironmentBlendMode p_blend_mode) {
+	if (emulate_environment_blend_mode_alpha_blend && p_blend_mode == XR_ENVIRONMENT_BLEND_MODE_ALPHA_BLEND) {
```
BastiaanOlij, line 3097: `get_supported_environment_blend_modes` needs to be altered too, so that if `XR_ENVIRONMENT_BLEND_MODE_ALPHA_BLEND` is not in the returned list, but `emulate_environment_blend_mode_alpha_blend` is true, it should be added. It will be `get_supported_environment_blend_modes` with which the user determines passthrough is available. (the `is_passthrough_supported` should be marked deprecated if you haven't already)

modules/openxr/openxr_interface.cpp
```diff
@@ -1198,6 +1196,11 @@ Array OpenXRInterface::get_supported_environment_blend_modes() {
 				WARN_PRINT("Unsupported blend mode found: " + String::num_int64(int64_t(env_blend_modes[i])));
 		}
 	}
+
+	if (openxr_api->is_emulating_environment_blend_mode_alpha_blend() && modes.find(XR_ENV_BLEND_MODE_ALPHA_BLEND) == -1) {
```
BastiaanOlij, line 1200: Ah, you added it here... I think it should be moved into `openxr_api`, but its a bit of a nitpicking thing. The hope is that eventually vendors like Meta, Pico and HTC jump onboard and actually embed this logic on the runtime side, and we'll be able to remove the logic we're adding here from `openxr_api`.

From my discussions I do get the impression that this was always the intent, but we may have gone down too far the rabbit hole to turn back.

BastiaanOlij (approved):
Other then some small nitpicking issues, this is perfect David, thanks for adding this.

BastiaanOlij:
Just for some extra background information for other reviewers as what David posted in the OP doesn't contain information on what got us here.

In the OpenXR spec it was always the intension to use the blend modes on the main layer to both query the capabilities of the device used and to set the mode of the device. So:

- `XR_ENVIRONMENT_BLEND_MODE_OPAQUE` indicates the rendered image is shown as is, no blending is performed, ergo the device is a VR only device or in VR mode (camera off in passthrough, blacked out visor, etc)
- `XR_ENVIRONMENT_BLEND_MODE_ADDITIVE` indicates the rendered image is "added" to the background, this really refers to traditional AR where an image is projected in front of the users eye (Hololens, Northstar, etc), these devices do not have the ability to fully block out the natural image the user sees.
- `XR_ENVIRONMENT_BLEND_MODE_ALPHA_BLEND` indicates the rendered image is alpha blended with the background. This refers to modern AR that use display technology that can obscure the natural image. This currently applies to waveguide based devices (MagicLeap) and passthrough based devices (Quest, Lynx R1, HTC Elite, etc.)

The great thing about this approach is that to the developer, it is easy to recognise these 3 modes and what they mean, to query the device which modes are supported and easy to select the desired mode without worrying how the hardware achieves that goal.

The problem that was introduced, especially in relation to passthrough, was two fold.
First, it wasn't clear from the spec that this was the desired behaviour sought after for passthrough, hence currently only Lynx R1 implementing it as designed.
Second, Meta had a need to quickly provide support for an already developer passthrough implementation when they embraced OpenXR, hence adding a vendor extension that introduced passthrough implemented similarly to their VrAPI.

We at Godot, having already implemented the VrAPI flavour of passthrough, followed suit and implemented Metas vendor extension but basing the passthrough design of Godot on this.

Fast forward to today, where some vendors are adopting the OpenXR way of doing passthrough AR while others copying Metas approach, we are faced with a confusing situation of implementing multiple different passthrough implementations and presenting them to the end user.

After much discussion internally, and asking the OpenXR WG about the original intent of the spec, we decided to present AR as designed by OpenXR to the user, and solve the platform differences internally by utilising our vendor plugin approach. Hence this PR.

modules/openxr/openxr_interface.cpp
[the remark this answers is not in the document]
BastiaanOlij: Does that work? I though when a function returns integers, `!value` translates to `value == 0`? Or is it `value != 1` ? I could be totally wrong here

BastiaanOlij:
Also almost forgot, we have to make sure that once this is merged, we update https://docs.godotengine.org/en/stable/tutorials/xr/openxr_passthrough.html to describe using the environment blend modes.

We can then have a chapter at the end that details that for passthrough to work on Quest and HTC, users must update to the latest vendors plugin. Here we can also mention about the start/end passthrough methods being deprecated. 

I might work on a draft for this.

modules/openxr/openxr_interface.cpp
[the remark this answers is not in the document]
akien-mga: It's changing from `find()`, which returns an integer to `has()`, which returns a boolean. `has() = (find() != -1)`.

modules/openxr/extensions/openxr_extension_wrapper_extension.h
```diff
@@ -39,7 +39,7 @@
 #include "core/os/thread_safe.h"
 #include "core/variant/native_ptr.h"
 
-class OpenXRExtensionWrapperExtension : public Object, OpenXRExtensionWrapper {
+class OpenXRExtensionWrapperExtension : public Object, public OpenXRExtensionWrapper, public OpenXRCompositionLayerProvider {
```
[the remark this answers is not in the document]
dsnopek: We could make a new extension class, however, nothing special is done with an `OpenXRCompositionLayerProvider` until you call `openxr_api->register_composition_layer_provider()`, so I think it's safe for them to always be `OpenXRCompositionLayerProvider`, since only the extensions that really need to do anything for that will register themselves.

modules/openxr/openxr_interface.cpp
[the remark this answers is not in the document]
dsnopek: Thanks, updated in my latest push!

Mickeon (commented):
I cannot speak one bit about how accurate the docs are, but a few notes.

modules/openxr/openxr_api.h
```diff
@@ -104,8 +104,10 @@ class OpenXRAPI {
 
 	// blend mode
 	XrEnvironmentBlendMode environment_blend_mode = XR_ENVIRONMENT_BLEND_MODE_OPAQUE;
+	XrEnvironmentBlendMode requested_environment_blend_mode = XR_ENVIRONMENT_BLEND_MODE_OPAQUE;
```
m4gr3d, line 107: I wonder if we should start using a `struct` variable instead for this pattern (we used the same approach for the `reference_space`) to ensure that updates to this logic does not inadvertently forget to account for the other field.

```
struct OpenXRBlendMode {
    XrEnvironmentBlendMode environment_blend_mode = XR_ENVIRONMENT_BLEND_MODE_OPAQUE;
    XrEnvironmentBlendMode requested_environment_blend_mode = XR_ENVIRONMENT_BLEND_MODE_OPAQUE;
    bool emulate_environment_blend_mode_alpha_blend = false;
}

OpenXRBlendMode blend_mode;
```

We should also document the differences between the two values (should also do the same for `reference_space`) to make it clear how they are being used.

m4gr3d (approved):
Some minor comments, but otherwise it looks good!

modules/openxr/doc_classes/OpenXRAPIExtension.xml
```diff
@@ -94,13 +100,34 @@
...
+			<param index="0" name="extension" type="OpenXRExtensionWrapperExtension" />
+			<description>
+				Registers the given extension as a composition layer provider.
+			</description>
+		</method>
+		<method name="set_emulate_environment_blend_mode_alpha_blend">
+			<return type="void" />
+			<param index="0" name="enabled" type="bool" />
+			<description>
+				Sets whether or not an OpenXR extension is loaded which is capable of emulating the [constant XRInterface.XR_ENV_BLEND_MODE_ALPHA_BLEND] blend mode.
```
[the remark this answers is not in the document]
BastiaanOlij: No, if we support alpha blend we're not emulating it.

modules/openxr/openxr_api.h
BastiaanOlij: I like this, all for clearly grouping things that belong together.

dsnopek: Are you envisioning this including these two as well?

```c++
	uint32_t num_supported_environment_blend_modes = 0;
	XrEnvironmentBlendMode *supported_environment_blend_modes = nullptr;
```

Part of the reason I ask is that `emulate_environment_blend_mode_alpha_blend` field is more a part of the list of supported modes, rather than the current/requested mode.

Or, maybe we should have two structs, like:

```c++
struct OpenXRBlendMode {
    XrEnvironmentBlendMode current_mode = XR_ENVIRONMENT_BLEND_MODE_OPAQUE;
    XrEnvironmentBlendMode requested_mode = XR_ENVIRONMENT_BLEND_MODE_OPAQUE;
};
OpenXRBlendMode blend_mode;

struct OpenXRBlendModeList {
    uint32_t num_modes = 0;
    XrEnvironmentBlendMode *modes = nullptr;
    bool emulate_alpha_blend_mode = false;
};
OpenXRBlendModeList supported_blend_modes;
```

That would also fit the pattern used for reference spaces pretty well.

In any case, either of these refactors will end up changing a whole bunch of code unrelated to passthrough. Could we do that as part of a dedicated PR?

doc/classes/XRInterface.xml
```diff
@@ -102,16 +102,18 @@
 				Is [code]true[/code] if this interface has been initialized.
 			</description>
 		</method>
-		<method name="is_passthrough_enabled">
+		<method name="is_passthrough_enabled" is_deprecated="true">
 			<return type="bool" />
 			<description>
 				Is [code]true[/code] if passthrough is enabled.
+				[i]Deprecated.[/i] Instead, check if [member environment_blend_mode] is [constant XRInterface.XR_ENV_BLEND_MODE_ALPHA_BLEND].
```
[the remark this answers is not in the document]
dsnopek: Fixed in my latest push

modules/openxr/doc_classes/OpenXRAPIExtension.xml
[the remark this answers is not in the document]
dsnopek: Fixed in my latest push

modules/openxr/openxr_api.h
m4gr3d: The two structs approach looks better. I'd imagine that `OpenXRBlendModeList supported_blend_modes` is initialized once on startup?

And yeah, totally fine to do it in a separate PR.

dsnopek:
My latest code changes adds a new method to `OpenXRAPIExtension` that was needed to allow the godot-cpp PR to detect that alpha blend mode isn't _really_ supported, to prevent it trying to emulate it when not necessary.

Anyway, at this point, I think all the review that directly pertains to this PR (as opposed to refactoring for follow-ups) has been addressed, either in a comment or with a code change.

modules/openxr/doc_classes/OpenXRAPIExtension.xml
```diff
@@ -94,13 +106,34 @@
...
+				Registers the given extension as a composition layer provider.
+			</description>
+		</method>
+		<method name="set_emulate_environment_blend_mode_alpha_blend">
+			<return type="void" />
+			<param index="0" name="enabled" type="bool" />
+			<description>
+				If set to [code]true[/code], an OpenXR extension is loaded which is capable of emulating the [constant XRInterface.XR_ENV_BLEND_MODE_ALPHA_BLEND] blend mode.
+			</description>
+		</method>
```
m4gr3d, line 122: We can do this automatically on our end instead of letting the gdextension dictates whether this should be set or not, by going through the list of gdextensions and checking their return values to `is_environment_blend_mode_alpha_blend_supported()`.

This would mitigate the issue where multiple gdextensions can be loaded each with different level of support. For example, one gdextension may have full support while the other may only be able to emulate support. The latter would set this flag which would cause the logic to assume that emulation is required when it's not.

modules/openxr/doc_classes/OpenXRAPIExtension.xml
```diff
@@ -75,12 +75,24 @@
...
 			<description>
 				Returns [code]true[/code] if OpenXR is initialized.
 			</description>
 		</method>
+		<method name="is_real_environment_blend_mode_alpha_blend_supported">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if the OpenXR runtime really supports [constant XRInterface.XR_ENV_BLEND_MODE_ALPHA_BLEND].
+			</description>
+		</method>
```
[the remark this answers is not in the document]
dsnopek: Sure, makes sense!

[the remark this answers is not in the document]
dsnopek: Added in my latest push!

dsnopek:
I've consolidated `OpenXRAPIExtension::is_emulating_environment_blend_mode_alpha_blend()` and `OpenXRAPIExtension::is_real_environment_blend_mode_alpha_blend_supported()` in my latest push (per the suggestion above)

akien-mga:
Thanks!

BastiaanOlij:
Oops, this was supposed to wait until we merge the counter part on the vendors plugin :P No worries, we'll deal with it.
@dsnopek we need to make sure we prioritise this next week :)

modules/openxr/doc_classes/OpenXRAPIExtension.xml
```diff
@@ -75,6 +75,12 @@
 				Returns the id of the system, which is a [url=https://registry.khronos.org/OpenXR/specs/1.0/man/html/XrSystemId.html]XrSystemId[/url] cast to an integer.
 			</description>
 		</method>
+		<method name="is_environment_blend_mode_alpha_supported">
+			<return type="int" enum="OpenXRAPIExtension.OpenXRAlphaBlendModeSupport" />
+			<description>
+				Returns [enum OpenXRAPIExtension.OpenXRAlphaBlendModeSupport] denoting if [constant XRInterface.XR_ENV_BLEND_MODE_ALPHA_BLEND] is really support, emulated or not supported at all.
```
m4gr3d, line 81: nit: should be `is really supported`

modules/openxr/openxr_api_extension.h
```diff
@@ -70,7 +72,21 @@ class OpenXRAPIExtension : public RefCounted {
...
 	bool can_render();
 
+	void register_composition_layer_provider(OpenXRExtensionWrapperExtension *p_extension);
+	void unregister_composition_layer_provider(OpenXRExtensionWrapperExtension *p_extension);
+
+	enum OpenXRAlphaBlendModeSupport {
+		OPENXR_ALPHA_BLEND_MODE_SUPPORT_NONE = 0,
+		OPENXR_ALPHA_BLEND_MODE_SUPPORT_REAL = 1,
+		OPENXR_ALPHA_BLEND_MODE_SUPPORT_EMULATING = 2,
+	};
```
m4gr3d, line 82: Out of curiosity, is it possible to reuse the same enum declaration here and in `openxr_api.h` instead of redeclaring it in both places?

dsnopek: I don't think so. I tried using both a `typedef` and `using` to pull the `enum` in from `OpenXRAPI` but the binding layer wouldn't recognize it correctly as being part of the class. Maybe someone else knows a way to do it?

modules/openxr/doc_classes/OpenXRAPIExtension.xml
dsnopek: Ah, thanks! Fixed in my latest push

dsnopek: Er, oops, this was already merged :-)

dsnopek: Ok, here's a PR to fix the type-o: https://github.com/godotengine/godot/pull/88298

PR #87631, opened 2024-01-26
Make viewport message dependent on framerate not physics step

ryevdokimov:
Related to #87621

How long this message appear is currently depending on the framerate when less than 60fps because it under the process notification. At 30 frames per second the message will appear twice as long, because the notification is being called half the time of each physics tick. 

image

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -2836,7 +2836,7 @@ void Node3DEditorViewport::_notification(int p_what) {
 					last_message = message;
 				}
 
-				message_time -= get_physics_process_delta_time();
+				message_time -= get_process_delta_time();
 				if (message_time < 0) {
 					surface->queue_redraw();
 				}


editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -2830,18 +2830,6 @@ void Node3DEditorViewport::_notification(int p_what) {
 				spatial_editor->update_transform_gizmo();
 			}
 
-			if (message_time > 0) {
-				if (message != last_message) {
-					surface->queue_redraw();
-					last_message = message;
-				}
-
-				message_time -= get_physics_process_delta_time();
```
Calinou, line 2839: Isn't it a matter of replacing `get_physics_process_delta_time()` with `get_process_delta_time()` here, without moving the code around?

ryevdokimov: It could be, but I wanted to keep it consistent with #87621, since canvas_item_editor_plugin.cpp doesn't have a notification for process. I thought it would be strange to make one just for this functionality, when either/or would work.

ryevdokimov: Although this does beg the question of **why** the stuff in canvas plugin is in physics process and not just process?

ryevdokimov: Yeah, I'll need to dig into this whole issue. It seems like it has the potential to affect several things. For now, I've applied your suggestion make it makes a lot more sense given what I said above.

Calinou (approved):

editor/plugins/node_3d_editor_plugin.cpp
Calinou: I've opened a separate issue for this: https://github.com/godotengine/godot/issues/87885

akien-mga:
Thanks!

PR #87633, opened 2024-01-26
Expose OpenGL ES 3.0 API

reduz:
An experimental PR that exposes the whole OpenGL ES 3.0 API to the binding layer.

It allows to, eventually, write custom rendering code when using the compatibility renderer, allowing to customize rendering.

The feature would only be useful after implementing a compositor on the GLES3 side (if this ever happens).

This is a draft made just to see if its possible and what would it take, likely needs to be completed and cleaned up at some point if there is intention to merge it.

Screens:

image
[..]
image

How does code look:

image

### DIFF
--- a/drivers/gles3/gles3.cpp
+++ b/drivers/gles3/gles3.cpp
@@ -1,0 +1,924 @@
+/**************************************************************************/
+/*  gles3.cpp                                                             */
+/**************************************************************************/
+/*                         This file is part of:                          */
+/*                             GODOT ENGINE                               */
+/*                        https://godotengine.org                         */
+/**************************************************************************/
+/* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+/* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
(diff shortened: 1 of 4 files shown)


Calinou:
The PascalCase casing looks unnatural to me, especially since OpenGL uses camelCase and not PascalCase. Either way, I think it'd be better to transform it to snake_case for a GDScript binding (and have C# convert it to PascalCase automatically).

BastiaanOlij:
I don't think we should expose the OpenGL API 1:1 directly, it bypasses our resource tracking and it will cause issues when dependencies are created between objects created in externals but used by core (which is the user case in a number of VR scenarios that we're already having problems with right now).

I had already started to make plans to take the logic in config.h (which contains some extension logic) and utilities.h (does resource tracking) and move that into an OpenGL class that we then complete with access to all OpenGL methods but with cherry picked ones extended with tracking and, where needed, dependency logic.

BastiaanOlij:
Owh and yes I would LOVE to make rendering effects work on GLES with this, but the same remark applies, without a bit of extra sauce around resource and dependency tracking, we're going to get into serious problems.

JekSun97:
This is something many have been asking for and will also allow you to go beyond just rendering, giving us all more options!

drivers/gles3/gles3.h
```diff
@@ -0,0 +1,2127 @@
...
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef GLES3_H
+#define GLES3_H
+
+#ifdef GLES3_ENABLED
+
+#include "core/object/class_db.h"
+#include "core/object/object.h"
```
AThousandShips, line 37:
```suggestion
```
Redundant

drivers/gles3/gles3.h
```diff
@@ -0,0 +1,2127 @@
...
+#include "platform_gl.h"
```
AThousandShips, line 38:
```suggestion

#include "platform_gl.h"
```
I'd separate these as they're from two different scopes

drivers/gles3/gles3.h
```diff
@@ -0,0 +1,2127 @@
...
+
+class GLES3 : public Object {
+	GDCLASS(GLES3, Object)
+protected:
+	static void _bind_methods();
+
+public:
+	enum GLEnum : unsigned int {
```
AThousandShips, line 46: Might want to prefix all these with `GL_` or something

drivers/gles3/gles3.cpp
```diff
@@ -0,0 +1,924 @@
...
+	{ GLES3::DEPTH_RANGE, 2, 0 },
+	{ GLES3::MAX_VIEWPORT_DIMS, 2, 0 },
+	{ GLES3::PROGRAM_BINARY_FORMATS, 0, GLES3::NUM_PROGRAM_BINARY_FORMATS },
+	{ GLES3::MAX_VIEWPORT_DIMS, 4, 0 },
+	{ GLES3::SHADER_BINARY_FORMATS, 0, GLES3::NUM_SHADER_BINARY_FORMATS },
+	{ GLES3::MAX_VIEWPORT_DIMS, 4, 0 },
+	{ 0, 0, 0 }
+};
+
+#endif
```
AThousandShips, line 924:
```suggestion
#endif

```

PR #87634, opened 2024-01-26
GDScript: Reintroduce binary tokenization on export

vnen:
This adds back a function available in 3.x: exporting the GDScript files in a binary form by converting the tokens recognized by the tokenizer into a data format.

It is enabled by default on export but can be manually disabled. The format helps with loading times since, the tokens are easily reconstructed, and with hiding the source code, since recovering it would require a specialized tool. Code comments are not stored in this format.

The `--test` command can also include a `--use-binary-tokens` flag which will run the GDScript tests with the binary format instead of the regular source code by converting them in-memory before the test runs.

Besides the regular option to export GDScript as binary tokens, this also includes a compression option on top of it. The binary format needs to encode some information which generally makes it bigger than the source text. This option reduces that difference by using Zstandard compression on the buffer.

- *Production edit: See https://github.com/godotengine/godot-proposals/issues/8605.*
- *Production edit: removed line breaks*
- *Production edit: closes godotengine/godot-roadmap#83*

*Bugsquad edit: `.gdc` files*


arkology:
> Besides the regular option to export GDScript as binary tokens, this
also includes a compression option on top of it. The binary format
needs to encode some information which generally makes it bigger than
the source text. This option reduces that difference by using Zstandard
compression on the buffer. 

Would be great to see exported projects size comparison

modules/gdscript/gdscript_parser.cpp
```diff
@@ -372,23 +379,56 @@ Error GDScriptParser::parse(const String &p_source_code, const String &p_script_
 	}
 }
 
+Error GDScriptParser::parse_binary(const Vector<uint8_t> &p_binary, const String &p_script_path) {
+	GDScriptTokenizerBuffer *buffer_tokenizer = memnew(GDScriptTokenizerBuffer);
+	Error err = buffer_tokenizer->set_code_buffer(p_binary);
+
+	if (err) {
+		return err;
```
AThousandShips, line 387:
```suggestion
		memdelete(buffer_tokenizer);
		return err;
```
Memory leak

modules/gdscript/gdscript_parser.cpp
```diff
@@ -359,6 +364,8 @@ Error GDScriptParser::parse(const String &p_source_code, const String &p_script_
 	parse_program();
 	pop_multiline();
 
+	memdelete(text_tokenizer);
```
AThousandShips, line 367:
```suggestion
	memdelete(text_tokenizer);
	tokenizer = nullptr;
```
To avoid dangling pointers

modules/gdscript/gdscript_parser.cpp
```diff
@@ -372,23 +379,56 @@ Error GDScriptParser::parse(const String &p_source_code, const String &p_script_
...
+			push_error(current.literal);
+		}
+		current = tokenizer->scan();
+	}
+
+	push_multiline(false); // Keep one for the whole parsing.
+	parse_program();
+	pop_multiline();
+
+	memdelete(buffer_tokenizer);
```
AThousandShips, line 406:
```suggestion
	memdelete(buffer_tokenizer);
	tokenizer = nullptr;
```

vnen:
To be clear, this is unrelated to the IR proposal. The proposal itself will take some time to solidify because it needs approval from the maintainers and not everyone think it's feasible.

This PR is a palliative to avoid having plain source code export while we wait for that or figure out something else, because a lot of people has been complaining about this issue.

AThousandShips:
> But returning tokenization does not solve this problem because scripts are still easily opened with gdsdecomp tool.

That won't change in any real way with any solution, any compiled/parsed/processed format will be vulnerable to that, it's just a matter of degrees, the biggest improvement will be this over plain source to the vast majority of "attacks"

AThousandShips:
I'd suggest reading up on this and getting a realistic perspective on this, if you genuinely believe that decompilation isn't a realistic way of reverse engineering code then you are not going to have realistic expectations of the outcomes of these things

> but after almost two months it turned out that no work had even begun

Please be realistic about the time it takes to undertake such a major project, and the explanations for the expectations on it above by vnen...

(Also if you're going to block me you will not react to my comments please, that's just petty, I don't want to have to block you back...)

fire:
Reducing loading times and applying zstd compression is good reason to reintroduce binary tokenization on its own merits.

editor/export/editor_export_preset.h
```diff
@@ -84,6 +90,7 @@ class EditorExportPreset : public RefCounted {
 	bool enc_directory = false;
 
 	String script_key;
+	int script_mode = MODE_SCRIPT_BINARY_TOKENS;
```
reduz, line 93: I think compressed is probably better default.

reduz (approved):
Looks good on my end.

reduz:
@MichaelWengren you can decompile and modify anything, be it .net or anything else. 
Eventually instead of an IR, what will most likely happen is GDScript files optionally compiled to binary (C) per platform.

vnen:
Updated to set the compressed mode as the default.

editor/export/project_export.cpp
```diff
@@ -1373,6 +1389,20 @@ ProjectExportDialog::ProjectExportDialog() {
...
+	script_vb->set_name(TTR("Scripts"));
+
+	script_mode = memnew(OptionButton);
+	script_vb->add_margin_child(TTR("GDScript Export Mode:"), script_mode);
+	script_mode->add_item(TTR("Text (easier debugging)"), (int)EditorExportPreset::MODE_SCRIPT_TEXT);
+	script_mode->add_item(TTR("Binary tokens (faster loading)"), (int)EditorExportPreset::MODE_SCRIPT_BINARY_TOKENS);
+	script_mode->add_item(TTR("Compressed binary tokens (smaller files)"), (int)EditorExportPreset::MODE_SCRIPT_BINARY_TOKENS_COMPRESSED);
+	script_mode->connect("item_selected", callable_mp(this, &ProjectExportDialog::_script_export_mode_changed));
+
+	sections->add_child(script_vb);
```
akien-mga, line 1404: Technically this is GDScript specific code spilling into the main editor, which wouldn't be relevant if the GDScript module is disabled. So it's slightly breaking encapsulation.

But it's not a first, and I'm not sure it's worth adding a complex abstraction for this so it can be handled in the module for each script language individually. Your call if you think it's worth looking into.

akien-mga (approved):

AThousandShips (changes requested):
Some accidental reverts, see above

editor/export/project_export.cpp
vnen: I though about this but I'm not sure of what would be the best solution. This is the same way it is handled in 3.x.

vnen:
> Some accidental reverts, see above

I thought `git push --force-with-lease` would prevent this. I'll have to be more careful from now on.

AThousandShips (approved):

akien-mga:
Thanks!

ogapo:
Just wanted to say I really appreciate this commit. I don't see obscuring the code as a security feature, but making the files smaller and faster is great. Most importantly, having comments stripped is very much a value add for a lot of developers. It's easy to leak things accidentally there since many devs used to compiled languages would assume comments are not user-visible.

azur-wolf:
about source safety (reverse-engineering protection, obfuscation, etc) there's any thread to follow?

dalexeev:
@azur-wolf See:

* godotengine/godot-proposals#4220

There are also proposals that can make decompilation more difficult, although they are not directly intended for this:

* godotengine/godot-proposals#8605
* godotengine/godot-proposals#6031
* godotengine/godot-proposals#3069

jason-cahill:
Hi @vnen : thank you for all the hard work you do. Can I ask if this work is replacing variable names? While classes and function names would also be great to obfuscate, the number one thing to minify would be variable names. Many software organizations don’t write much in the way of comments, and symbols go a long way towards being the documentation. Having the tokenizer scramble variable names would really take a huge step towards making the intent of a function opaque.

Calinou:
> Can I ask if this work is replacing variable names?

No, it keeps identifiers as-is. Local variables could technically be renamed, but this is nontrivial to implement. Member variables can't be renamed without breaking the API that other classes may attempt to use, especially since there is no way to mark member variables as private yet.

PR #87635, opened 2024-01-26
Manage time singleton in `register_core_types`

rune-scape:
the time singleton is used rlly early on, and `Time::get_singleton()` was always called before it was initialized, creating a memory leak when it `memnew`ed, then ObjectDB caught the stray instance
this registers it in a more specific place right after StringName and ObjectDB are initialized,

### DIFF
--- a/core/register_core_types.cpp
+++ b/core/register_core_types.cpp
@@ -102,6 +102,7 @@ static core_bind::Marshalls *_marshalls = nullptr;
 static core_bind::EngineDebugger *_engine_debugger = nullptr;
 
 static IP *ip = nullptr;
+static Time *_time = nullptr;
 
 static core_bind::Geometry2D *_geometry_2d = nullptr;
 static core_bind::Geometry3D *_geometry_3d = nullptr;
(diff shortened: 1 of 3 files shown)


core/register_core_types.cpp
```diff
@@ -96,6 +96,7 @@ static Ref<ResourceFormatLoaderJSON> resource_loader_json;
 static core_bind::ResourceLoader *_resource_loader = nullptr;
 static core_bind::ResourceSaver *_resource_saver = nullptr;
 static core_bind::OS *_os = nullptr;
+static Time *_time = nullptr;
```
aaronfranke, line 99: I would prefer this to be in a different order, so we can keep all the `core_bind` lines together. You could perhaps put this right below `static IP *ip` (which might need to be `_ip`?).

aaronfranke (commented):
Makes sense to me, we want to unify where this is created and have it be as early as possible.

aaronfranke (approved):
If it runs it should be good.

Issue #85085, opened 2023-11-19
Mesh Library Export silently ignores some meshes


danielbarla:
### Godot version

4.1.2.stable.official [399c9dc39]

### System information

Godot v4.1.2.stable - Windows 10.0.22621 - Vulkan (Compatibility) - AMD Radeon RX 6800 (Advanced Micro Devices, Inc.; 31.0.14037.17019) - AMD Ryzen 7 5800X3D 8-Core Processor (16 Threads)

### Description

When using the "Scene -> Export As -> MeshLibrary" functionality, some meshes seem to be consistently ignored, while others are successfully exported. The editor does not seem to give any indication of any issues encountered, leading the user to think the export was perfect. In some cases, the resulting `.tres` or `.meshlib` file is mostly empty (containing only some boilerplate).

### What doesn't work

The export itself is incomplete. In one case, when exporting a scene with 20 assets, I end up with a mesh lib with 12. In the repro case below, I have a scene with 1 asset which results in a mesh lib with 0 meshes.

### How do you expect it to work instead?

- The assets I've encountered this with were very simple low-poly meshes from Kenney.nl; my expectation would be for the export to work in general.
- If there's any kind of failure during the export, I'd like to detect it at the earliest possible moment. Having a warning message show up in the debugger would be a good low-effort way to achieve this. Having an easy way to drill into the root cause of the failure (e.g. some inconsistency in the mesh data, etc), perhaps via a warning message containing further information about how to enable debug diagnostics for the export feature would be great.

### Scale and scope

Unsure, but all GLB assets from Kenney's [City Kit](https://www.kenney.nl/assets/city-kit-commercial) seem to suffer from this issue, while none of the GLB assets from their [City Builder](https://github.com/KenneyNL/Starter-Kit-City-Builder) starter project seem to have this issue. Both the difficulty with the export itself and the silent way it fails are issues, though they are actually separate.

### Existing / similar issues

The issue seems most similar to the following unanswered topic on ask.godotengine: https://ask.godotengine.org/152903/anyone-have-issues-exporting-meshlibraries-before . The main difference seems to be that in that question, the nodes have an icon indicating they are of type MeshInstance3d, while in my case (both in situations where the export works and where it doesn't), they are simply Node3D.

### Steps to reproduce

(shortcut: use the attached repro project which has the two assets mentioned below)

### Unhappy path

- Create a new Godot project
[...]

PR #87651, opened 2024-01-27
MeshLibrary scene import: recurse 4 nodes deep

WebFreak001:
before we only recursed 2 nodes deep, now we just do 4 nodes deep - this should be sufficient until there is a new system for exporting nodes with more control.

Fixes importing certain Kenney assets, like described in #85085 or shown in https://dev.to/robpc/why-godot-drops-objects-when-creating-a-meshlibrary-14bg

We still only pick at most 1 MeshInstance3D from every node of only the immediate scene children.

Only tested this locally on my specific use-case where it fixed the issue (with Kenney's assets, different pack but same root issue as in the link)


akien-mga:
Superseded by #87923.

Thanks for the contribution nevertheless!

PR #87654, opened 2024-01-27
Close editor inspector popup on scroll

### LINKED ISSUE #65191: Scrolling with mouse wheel does not close popups
### Godot version

0c221f0

### System information

Windows 10 x64

### Issue description

godot windows tools x86_64_E7RepJv48o

### Steps to reproduce

1. Right-click to open context menu in some scrollable area
2. Scroll

_No response_

ryevdokimov:
Fixes: #65191

I'm not sure if the scroll wheel should close all popups because this may break some functionality when used in games, so I limited it to just the editor inspectors. 

I tried to find a more straightforward and more efficient way to do this, but this was the simplest that didn't seem to have any side effects or require major surgery in the code.

Before: 

https://github.com/godotengine/godot/assets/105675984/a23d90ab-6aae-4811-9e3a-4e9e5ab014bd

After:

https://github.com/godotengine/godot/assets/105675984/d7d82f00-99e2-4644-b447-d14de0b5ac8e

### DIFF
--- a/editor/editor_inspector.cpp
+++ b/editor/editor_inspector.cpp
@@ -4059,6 +4059,8 @@ void EditorInspector::_vscroll_changed(double p_offset) {
 		return;
 	}
 
+	propagate_notification(NOTIFICATION_WM_CLOSE_REQUEST);
+
 	if (object) {
 		scroll_cache[object->get_instance_id()] = p_offset;
 	}


ryevdokimov:
I realized that the popup menus that are already visible in the inspector (without right clicking) are also broken. I pushed a change that calls a close request any time you're scrolling regardless of what's open. It's not as efficient, but I think the performance penalties are minimal.

KoBeWi:
That's a very hacky solution .-.
I wonder if it's possible to close the popups directly. Also the issue technically isn't about inspector only, the same problem occurs in scene dock, file dock etc.

ryevdokimov:
Single line hacks are the best hacks though 😂

I take your point, however. I couldn't find an easy direct way to do it at the time. I'll try to revisit the problem in the future.

KoBeWi:
The problem with this hack is that it's abusing a notification sent when window is being closed. Something may already handle it, leading to invalid behavior.

ryevdokimov:
I think you're right, I didn't consider looking upstream and see what propagated the notification itself, so I see now that is definitely is hackier that I thought. Maybe the solution is to just build something similar with a new notification?

KoBeWi:
You could also add a signal. It requires connecting it in every EditorProperty, but it's probably better than propagating a notification to every node in the inspector.

Issue #62599, opened 2022-07-01
Setting control node to pass and overriding _gui_input does not work.


Nukiloco:
### Godot version

v4.0.alpha.calinou [caa9ec8ac]

### System information

Windows 10, Nvidia GTX 750 TI, Nvidia GTX 1050 TI

### Issue description

From what I understand from the documentation on gui inputs:

``Stop`` captures the mouse input events and doesn't allow them to pass on.
``Pass`` captures the mouse input events but if its not accepted then it passes it on.
``Ignore`` ignores the mouse input events and pass them on.

However in the reproduction project I have below, this doesn't seem to be happening with the ``Pass`` event.

### Steps to reproduce

Download the reproduction project below, press F5 and run the project. Hover your mouse over the button and notice how its not being highlighted and that you can't click on it.

The main scene in this project is setup where all the control nodes are set to ``Ignore`` except...
``Button`` which is set to ``Stop``
``Panel`` which is set to ``Pass``

You will see that the ``Panel`` node has a script attached to it. In this script you can see that ``_gui_input`` is overridden but does not call ``accept_input`` at all. From my understanding, this should allow the input to pass into the Button.

### Minimal reproduction project

[GuiInputPassIssue.zip](https://github.com/godotengine/godot/files/9030639/GuiInputPassIssue.zip)

Sauermann:
In your MRP, you want to pass the event to siblings and not to parents. This is currently not possible and what you experience is expected behavior.
This proposal summarizes the current situation: godotengine/godot-proposals#3613

PR #87656, opened 2024-01-27
Rename Mouse Filter Pass to Propagate Up for clarity

kitbdev:
- related (closes?) https://github.com/godotengine/godot/issues/55432
- related (closes?) https://github.com/godotengine/godot/issues/62599
- related https://github.com/godotengine/godot-proposals/issues/3613
- related https://github.com/godotengine/godot-proposals/issues/1775
- related https://github.com/godotengine/godot-proposals/issues/7167

`MOUSE_FILTER_PASS` is replaced by `MOUSE_FILTER_PROPAGATE_UP`.
The functionality is the same and this should not break compatibility.
In the inspector everything that was set to `Pass` should now show as `Propagate Up`.
The enum constant `MOUSE_FILTER_PASS` has the same value but is deprecated.

Lots of people get confused by `MOUSE_FILTER_PASS`, thinking it means that events should be passed to Controls behind it, not to its parent Controls. The new name should make it clear it is not intended to do that.

Edit: Previously was 'Bubble'.


Mickeon:
I'm not necessarily against this but is **MOUSE_FILTER_BUBBLE** really a better name for it? I don't personally find it to be clearer or more explicit than **MOUSE_FILTER_PASS** for most users, although from a web developer's perspective the name would make some sense.

doc/classes/Control.xml
```diff
@@ -1325,15 +1325,18 @@
...
+		<constant name="MOUSE_FILTER_BUBBLE" value="1" enum="MouseFilter">
+			The control will receive mouse movement input events and mouse button input events if clicked on through [method _gui_input]. The control will also receive the [signal mouse_entered] and [signal mouse_exited] signals. If this control does not handle the event, the event will propagate up to its parent control if it has one. The event is bubbled up the Node hierarchy until it reaches a non-[CanvasItem], a control with [constant MOUSE_FILTER_STOP], or a CanvasItem with [member CanvasItem.top_level]. This will allow signals to fire in all controls it reaches. If no control handled it, the event will be passed to [method Node._shortcut_input] for further processing.
```
Mickeon, line 1331: This is a huge paragraph that could be split onto a newline, to maintain a small summary of the constant. Surround **CanvasItem** with square brackets, lowercase "node" in this context makes more sense.
```suggestion
			The control will receive mouse movement input events and mouse button input events if clicked on through [method _gui_input]. The control will also receive the [signal mouse_entered] and [signal mouse_exited] signals.
			If this control does not handle the event, the event will propagate up to its parent control if it has one. The event is bubbled up the node hierarchy until it reaches a non-[CanvasItem], a control with [constant MOUSE_FILTER_STOP], or a [CanvasItem] with [member CanvasItem.top_level]. This will allow signals to fire in all controls it reaches. If no control handled it, the event will be passed to [method Node._shortcut_input] for further processing.
```

ryevdokimov:
> I'm not necessarily against this but is **MOUSE_FILTER_BUBBLE** really a better name for it? I don't personally find it to be clearer or more explicit than **MOUSE_FILTER_PASS** for most users, although from a web developer's perspective the name would make some sense.

Having done web development, I think the name is a lot clearer per its description, and I think it makes sense to follow popular convention since many of these platforms can overlap.

Mickeon:
I'm a web developer myself and I'm more-so concerned from the perspective of users that come to engine to... make games. I agree on the sentiment to use some form of the convention, it's just...

The improved description of this PR is great, and I feel like it would reduce a lot of confusion about this, just by itself. That, combined with a link mentioning event bubbling, and there may be less need to deprecate the current name.

kitbdev:
> is MOUSE_FILTER_BUBBLE really a better name for it?

I was also thinking about `MOUSE_FILTER_BUBBLE_UP` or `MOUSE_FILTER_PROPAGATE_UP`.

Here are some of the suggested alternatives I didn't use: 
- `MOUSE_FILTER_HANDLE_WITH_PARENT` is pretty long and needs `MOUSE_FILTER_STOP` to also be renamed.
- `EVENT_PROPAGATION_PARENT` needs mouse filters to be renamed entirely. (Which makes sense, but I wanted to deprecate as little as needed)
- `MOUSE_FILTER_PARENT` isn't very clear.

I'm not sure if there is another term for event bubbling, but if there is a better name let me know.

For non-web development, I know Unity also uses [Event Bubbling](https://docs.unity3d.com/Manual/UIE-Events-Dispatching.html).

>That, combined with a link mentioning event bubbling

I could add a link to the Wikipedia page. It's focused on web development but I think most resources about event bubbling are.
https://en.wikipedia.org/wiki/Event_bubbling

> and there may be less need to deprecate the current name.

I think that when users see 'Pass', they may assume they know what it does without checking the docs, so I feel it is still important to change the name.

ryevdokimov:
I understand the want for not deprecating the current name, just putting out my 2 cents. 

In my opinion the name of the various data members and functions are arguably more important than its description.  Self-documenting code and all that. The name pass to me suggests some kind of pass(through) functionality which is where the confusion stems from. Even with the documentation, when I read the code there will always be that hiccup. 

Whether it's game of web development the concepts are the same, and I think you'd be doing all developers a favor by reinforcing that knowledge every time they read the code.

AThousandShips:
I don't think it's acceptable to rename this enum as it's reasonably correct, and renaming it is a major breaking change that'll break any code using this value, the documentation should be changed for clarity instead, we're at a point in 4.x where breaking this level of compatibility IMO is completely unacceptable for anything that isn't a broken behavior

Adding an alias for the name would be reasonably acceptable with the old name deprecated, but I still think that's too drastic a change, there's nothing wrong with the current name as such, so better and clearer documentation is the way to go, especially as "Bubble" isn't any more clear without insight than "Pass" is IMO, either name requires you to know what's meant

Mickeon:
Could a proposal be made for this to gather some further feedback?

AThousandShips:
A proposal sounds like the way forward indeed, keep in mind possible confusion from 3.x though as well, this PR should also add a conversion from 3.x to rename the enum used to help with this

kitbdev:
I figured the 5 issues I linked would be enough, especially https://github.com/godotengine/godot-proposals/issues/3613 which has ongoing discussion of changing the name.
If I opened another proposal it feels like it would be a duplicate.

golddotasksquestions:
While `mouse_filter_bubble_up` would probably be a better name for the _current_ behavior, a name change would not fix any of the issues it has. 

Users want a `mouse_filter_pass`, not a `mouse_filter_bubble_up` behavior.

kitbdev:
The current behavior works as intended, and I think that people mostly want it to be changed _because_ the name causes misleading expectations.
Changing it to have pass-through behavior would be a breaking change, so its not going to happen. Adding a separate pass-through mouse filter may be possible, but then it is even more important that this current behavior is renamed.

ajreckof:
As someone not coming from web-devalopment buble means really nothing to me. On the other end pass means immediatly something to me and I know the pass means it is passed to the parent for it to handle but sometime I think the other way around I need to do something ie let it pass through the node to node behind and I'm like this seems so basic as this is something a lot could want that their should be one that does it. Since it can't be the other two it should be this one. Sometime I'm even remembering it pass to the parent and I'm twisting like maybe then the parent will check if the nex child wants to handle it. Obviously it does not work like this and I'm always feeling dumb. But what I wanted to say is I think the problem stems more from the missing functionality then the miss-leading name. As I think most people using the engine don't know about event bubbling I don't feel it would help them as they would still assume b elimination it should be the last one they don't understand. Obviously they might not complain about it as much but they might just start complaining that their is none doing what they want which in the end was what they wanted from the start. Just adding a PASSTHROUGH would make clear PASS is not PASSTHROUGH. 

On another not I feel like MOUSE_FILTER_PROPAGATE_UP is way clearer then current and MOUSE_FILTER_BUBBLE

kitbdev:
Updated to use Propagate Up instead of Bubble.
I thought conversions were only for things that need to be changed, but I guess not. I added a conversion from 3 to 4 in `editor\renames_map_3_to_4.cpp`, let me know if I didn't do it correctly.

image

Mickeon:
I'll still recommend to open a detailed proposal about this because it is unlikely this will be accepted due to lack of consensus.

Calinou:
I'd just rename the inspector hint from `Pass` to `Pass (Propagate Up)`, so we don't have to break compatibility.

kitbdev:
This PR doesn't break compat, but that is a good idea.
- opened a new PR for it https://github.com/godotengine/godot/pull/93363

Issue #65803, opened 2022-09-14
"2D" and "3D" on node's names change when added to a scene


FHEK789:
### Godot version

4.0.alpha17.mono

### System information

Kubuntu 22.04

### Issue description

The "2D" or "3D" part of nodes' names change to lowercase when added to a scene, so `CharacterBody2D` becomes `CharacterBody2d`.

### Steps to reproduce

1. Open "Add Child Node" dialog
2. See that the 2D/3D part of nodes' names are uppercase
3. Add some node and see that in the Scene dock the 2D/3D part is now 2d/3d

dalexeev:
This is caused by the combination of #64570 and #63902. Previously, the Pascal Case variant actually meant Auto.

TheYellowArchitect:
Related for docs to upgrade consistently: https://github.com/godotengine/godot-docs/issues/6245

PR #87659, opened 2024-01-27
Respect `adjust_name_casing` setting more consistently

### LINKED ISSUE #87652: editor/naming/node_name_casing not respected for default Create Root Node buttons
### Tested versions

- Reproducible in: v4.3.dev2.official [352434668]

### System information

Windows 10 & Debian GNU/Linux 11 (bullseye)

### Issue description

The 3 default "Create Root Node" buttons:
- `2D Scene`
- `3D Scene`
- `User Interface`
all generate a root node that doesn't respect the `editor/naming/node_name_casing` project setting

Screenshot from 2024-01-27 15-08-47

### Steps to reproduce

1. change the `editor/naming/node_name_casing` project setting to something other than PascalCase
2. create a new scene
3. use any of the default 3 "Create Root Node" buttons
4. see that the `Node2D`, `Node3D`, or `Control` nodes do not follow the case setting

this can be done with a fresh, empty Godot project. Here's one with `naming/node_name_casing=2` (snake_case) 
[casing.zip](https://github.com/godotengine/godot/files/14073752/casing.zip)

SylvanSign:
- attempts to fix:
   - https://github.com/godotengine/godot/issues/87652
   - https://github.com/godotengine/godot/issues/87653
- I'm new to this repo, so not sure whether these are the best ways to fix those issues

* *Bugsquad edit, fixes: https://github.com/godotengine/godot/issues/87652*
* *Bugsquad edit, fixes: https://github.com/godotengine/godot/issues/87653*

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -7597,9 +7597,11 @@ void Node3DEditor::_add_sun_to_scene(bool p_already_added_environment) {
 		// Create a root node so we can add child nodes to it.
 		SceneTreeDock::get_singleton()->add_root_node(memnew(Node3D));
 		base = get_tree()->get_edited_scene_root();
+		base->set_name_to_case_adjusted_class();
 	}
 	ERR_FAIL_NULL(base);
 	Node *new_sun = preview_sun->duplicate();
+	new_sun->set_name_to_case_adjusted_class();
 
 	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
 	undo_redo->create_action(TTR("Add Preview Sun to Scene"));
@@ -7626,10 +7628,12 @@ void Node3DEditor::_add_environment_to_scene(bool p_already_added_sun) {
 		// Create a root node so we can add child nodes to it.
 		SceneTreeDock::get_singleton()->add_root_node(memnew(Node3D));
 		base = get_tree()->get_edited_scene_root();
+		base->set_name_to_case_adjusted_class();
 	}
 	ERR_FAIL_NULL(base);
 
 	WorldEnvironment *new_env = memnew(WorldEnvironment);
+	new_env->set_name_to_case_adjusted_class();
 	new_env->set_environment(preview_environment->get_environment()->duplicate(true));
 	if (GLOBAL_GET("rendering/lights_and_shadows/use_physical_light_units")) {
 		new_env->set_camera_attributes(preview_environment->get_camera_attributes()->duplicate(true));
--- a/scene/main/node.h
+++ b/scene/main/node.h
@@ -395,6 +395,8 @@ public:
 	String get_description() const;
 	void set_name(const String &p_name);
 
+	_FORCE_INLINE_ void set_name_to_case_adjusted_class() { set_name(adjust_name_casing(get_class())); }
+
 	InternalMode get_internal_mode() const;
 
 	void add_child(Node *p_child, bool p_force_readable_name = false, InternalMode p_internal = INTERNAL_MODE_DISABLED);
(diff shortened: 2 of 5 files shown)


scene/main/node.cpp
```diff
@@ -1315,7 +1315,7 @@ void Node::_generate_serial_child_name(const Node *p_child, StringName &name) co
 	if (name == StringName()) {
 		// No name and a new name is needed, create one.
 
-		name = p_child->get_class();
+		name = adjust_name_casing(p_child->get_class());
```
Mickeon, line 1318: Hold on here. This also adjusts the name when `force_readable_name` is true in Node's `add_child`.

It's worth keeping in mind if this is desired (in my opinion, yes).

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -7561,6 +7561,7 @@ void Node3DEditor::_add_sun_to_scene(bool p_already_added_environment) {
 	}
 	ERR_FAIL_NULL(base);
 	Node *new_sun = preview_sun->duplicate();
+	new_sun->set_name(adjust_name_casing(new_sun->get_name()));
```
Mickeon, line 7564: Did you not mean to fix the same issue for the Add Environment to Scene button? (It's created on the method right below this one).

FantasmaGolicon: I am also just a first contributor, but putting my insight since I also looked at the issue without being aware there was a PR up already (And thought about doing this same solution):

Would it be preferred to fix the name for the WorldEnvironment on the calling point instead of modifying the Node implementation directly? I see that it is critical code, and it may not be worth the larger impact for an Editor-only fix. Along with the potential perf considerations.

Mickeon: My thoughts exactly. As [I commented above](https://github.com/godotengine/godot/pull/87659#discussion_r1468838556), while the above solution may be more encompassing it would also affect code outside of the editor (even at run-time), which may have some unintended consequences. In this case it may be better to be "safe" and implement the same solution as the Sun node.

Mickeon (dismissed):
It's all good now, [could you squash your commits](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase)?

AThousandShips:
Your commit seems not to be linked to your GitHub account. See: [Why are my commits linked to the wrong user?](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user) for more info.

Mickeon:
The current solution would work fine and the PR should be kept as is, probably. But perhaps, to avoid repeating oneself, it could be delegated to a separate function or macro.

scene/main/node.h
```diff
@@ -757,6 +757,8 @@ Error Node::rpc_id(int p_peer_id, const StringName &p_method, VarArgs... p_args)
 #define ERR_READ_THREAD_GUARD_V(m_ret)
 #endif
 
+#define SET_NAME_TO_CASE_ADJUSTED_CLASS(nodePtr) nodePtr->set_name(adjust_name_casing(nodePtr->get_class()))
```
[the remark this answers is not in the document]
dalexeev: I would prefer a method instead of a macro:

```c++
_FORCE_INLINE_ void set_name_to_case_adjusted_class() { set_name(adjust_name_casing(get_class())); }
```

It might make sense to name the method `reset_name()`, although this is too vague a name and might confuse someone into thinking the name will reset to something like `@Node@123`.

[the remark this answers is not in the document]
AThousandShips: In either case it should be:
```suggestion
#define SET_NAME_TO_CASE_ADJUSTED_CLASS(m_node_ptr) m_node_ptr->set_name(adjust_name_casing(m_node_ptr->get_class()))
```

KoBeWi:
Forcing `adjust_name_casing()` with pascal case breaks names:
image
See this old issue: #65803

KoBeWi:
At least for `adjust_name_casing()`, we could add an argument whether it should skip pascal case. It's default almost everywhere, so it would save some checks.

Issue #25328, opened 2019-01-25
Profiler should list functions in inverse order.


Ranoller:
**Godot version:**

Master

**OS/device including version:**

All

**Issue description:**

Profiler shows script functions in time-ascendent way. Usual propose of profiler is to know what parts of code are more expensive to the processor, so this order is nonsense, **should list functions from more expensive ones to less**. Autoscroll doesn´t help with that because allways rise the list to the top and you can´t  follow the bottom functions (the more expensives ones).... I think this change is not new functionality, not difficult to achieve and can be adressed before 3.1 launch because (not compatibility breaking or functionality change).

Zylann:
Also mentionned in https://github.com/godotengine/godot/issues/5682

akien-mga:
Fixed by #25354.

PR #87661, opened 2024-01-28
Fix editor profiler script function sort order

baptr:
In 4.2.1 stable, and at HEAD, a busy frame in a gdscript-only game may show next to no time in Script Functions (note 70ms in Process, 0.00ms in Scripts):
image

However, looking at a similar frame with the CLI `--profiling` flag shows much more useful output:

```
FRAME: total: 0.662813 script: 0.612449/92 %
0:res://autoload/skill_manager.gd::61::restore_skill
        total: 0.679129/102 %   self: 0.000074/0 % tcalls: 8
1:res://spawners/components/spawn_placer_component.gd::28::SpawnPlacerComponent._place_loop
        total: 0.611002/92 %    self: 0.000103/0 % tcalls: 1
2:res://actor/actor.gd::28::Actor.run
        total: 0.610062/92 %    self: 0.000166/0 % tcalls: 1
3:res://components/behavior_component.gd::57::BehaviorComponent.run
        total: 0.60987/92 %     self: 0.000004/0 % tcalls: 1
4:res://autoload/skill_manager.gd::132::restore_behavior
        total: 0.609817/92 %    self: 0.00001/0 % tcalls: 1
5:res://autoload/skill_manager.gd::54::restore_rule
        total: 0.60979/92 %     self: 0.00003/0 % tcalls: 2
6:res://autoload/skill_manager.gd::0::Resource.duplicate
        total: 0.609644/91 %    self: 0.609644/91 % tcalls: 9
```

After this PR, we see the more expensive functions in the editor list as well: 

image

---

The engine internally limits the number of functions reported back (to 16 by default). To this point, it's been sort the profiling info in *ascending* order of time spent, then trimming the list. This meant we may only see the *best* (fastest) functions, instead of the worst, which you probably want when profiling.

Now the servers sort more closely matches the local_debugger, which worked fine.

Cleans up #25328

### DIFF
--- a/editor/debugger/editor_profiler.cpp
+++ b/editor/debugger/editor_profiler.cpp
@@ -350,7 +350,7 @@ void EditorProfiler::_update_frame() {
 			category->set_custom_color(0, _get_color_from_signature(m.categories[i].signature));
 		}
 
-		for (int j = m.categories[i].items.size() - 1; j >= 0; j--) {
+		for (int j = 0; j < m.categories[i].items.size(); j++) {
 			const Metric::Category::Item &it = m.categories[i].items[j];
 
 			if (it.internal == it.total && !display_internal_profiles->is_pressed() && m.categories[i].name == "Script Functions") {
--- a/servers/debugger/servers_debugger.cpp
+++ b/servers/debugger/servers_debugger.cpp
@@ -198,7 +198,7 @@ class ServersDebugger::ScriptsProfiler : public EngineProfiler {
 	typedef ServersDebugger::ScriptFunctionInfo FunctionInfo;
 	struct ProfileInfoSort {
 		bool operator()(ScriptLanguage::ProfilingInfo *A, ScriptLanguage::ProfilingInfo *B) const {
-			return A->total_time < B->total_time;
+			return A->total_time > B->total_time;
 		}
 	};
 	Vector<ScriptLanguage::ProfilingInfo> info;


servers/debugger/servers_debugger.cpp
```diff
@@ -198,7 +198,7 @@ class ServersDebugger::ScriptsProfiler : public EngineProfiler {
 	typedef ServersDebugger::ScriptFunctionInfo FunctionInfo;
 	struct ProfileInfoSort {
 		bool operator()(ScriptLanguage::ProfilingInfo *A, ScriptLanguage::ProfilingInfo *B) const {
-			return A->total_time < B->total_time;
+			return A->total_time > B->total_time;
```
AThousandShips, line 201: Is this only used in that one place? If not any other possible use of this part needs to be updated for the order change

baptr: This ProfileInfoSort is only used [once, locally](https://github.com/godotengine/godot/blob/17e7f85c06366b427e5068c5b3e2940e27ff5f1d/servers/debugger/servers_debugger.cpp#L243)

The result is visible ~2 places:
1. In the order that `servers:function_signature` is [emitted](https://github.com/godotengine/godot/blob/17e7f85c06366b427e5068c5b3e2940e27ff5f1d/servers/debugger/servers_debugger.cpp#L256), but it is only consumed to put in a map [here](https://github.com/godotengine/godot/blob/17e7f85c06366b427e5068c5b3e2940e27ff5f1d/editor/debugger/script_editor_debugger.cpp#L682)
1. In the frame data [emitted](https://github.com/godotengine/godot/blob/17e7f85c06366b427e5068c5b3e2940e27ff5f1d/servers/debugger/servers_debugger.cpp#L319-L321) with `servers:profile_total` or `server:profile_frame`, [stored](https://github.com/godotengine/godot/blob/17e7f85c06366b427e5068c5b3e2940e27ff5f1d/editor/debugger/script_editor_debugger.cpp#L787) in the script funcs category.items, [added](https://github.com/godotengine/godot/blob/17e7f85c06366b427e5068c5b3e2940e27ff5f1d/editor/debugger/script_editor_debugger.cpp#L793-L795) to the UI, and consumed only [here](https://github.com/godotengine/godot/blob/90a5f23e798df8192e903fc5785e29fa093e5caa/editor/debugger/editor_profiler.cpp#L353), which this makes handle correctly.

So I think it's all good.

editor/debugger/editor_profiler.cpp
```diff
@@ -350,7 +350,7 @@ void EditorProfiler::_update_frame() {
 			category->set_custom_color(0, _get_color_from_signature(m.categories[i].signature));
 		}
 
-		for (int j = m.categories[i].items.size() - 1; j >= 0; j--) {
+		for (int j = 0; j < m.categories[i].items.size(); j++) {
```
adamscott, line 353: This change is not really needed, doesn't it?

akien-mga (approved):
Nice catch!

akien-mga:
Cherry-picked for 4.2.2.

Issue #78079, opened 2023-06-10
Impossible to start text selection on unfocused SpinBox


floppyhammer:
### Godot version

v4.1.beta.custom_build [d3651ea45]

### System information

Windows 11

### Issue description

![1](https://github.com/godotengine/godot/assets/28705694/15c14c93-7ad7-425a-8fb0-bef841b75527)

It's not possible to start text selection on a SpinBox when it's unfocused.

Note: After a bit of digging in the code, I found that clicking on a SpintBox doesn't trigger `SpinBox::gui_input` no matter if it's focused or not. However, releasing mouse left button does trigger the function.

### Steps to reproduce

See the gif.

### Minimal reproduction project

It's simple to reproduce.

Sauermann:
`SpinBox::_line_edit_focus_enter` gets called when the LineEdit of the SpinBox receives focus and during that the drag-selection gets canceled.
Bisecting lead to #64529

floppyhammer:
I confirm SpinBox in 3.x has the same issue, probably due to the cherry-pick of https://github.com/godotengine/godot/pull/64529.

Calinou:
cc @KoBeWi

KoBeWi:
You can select all with double click. There is also `select_all_on_focus` property.

The "regression" was mentioned in the PRs caveats 🙃 I'll see if it can be improved.

PR #87674, opened 2024-01-28
Prevent `LineEdit` focus loss when text is submitted or rejected and allow selecting without editing with arrow keys.

### LINKED ISSUE #87673: LineEdit and SpinBox focus issue
### Tested versions

4.2.1

### System information

win 10

### Issue description

1 - when it's focused with keyboard, it immediately switch to edit mode instead of just selecting the line edit, so the only way to select the next focusable `CanvasItem` is by pressing Tab.
2 - if the `LineEdit` is focused and escape button is pressed, the focus is lost! and it's impossible to focus any `CanvasItem` again with the keyboard until we focus one of them with mouse click . [ Editor and Game ]

### Steps to reproduce

N/A

N/A

WhalesState:
Fixes: https://github.com/godotengine/godot/issues/87673

- Prevent focus loss when text is rejected by pressing `Escape`.
- [Editor] Prevent focus loss when a text property is submitted by pressing `Enter`.
- Allow exiting edit mode by pressing `Escape` to select other focusable `Control` nodes.
- Allow entering focus mode only without editing by entering focus with arrow keys, pressing `Enter` will toggle between select/edit modes and pressing `Escape` while editing will return back to select mode.

### DIFF
--- a/doc/classes/LineEdit.xml
+++ b/doc/classes/LineEdit.xml
@@ -4,7 +4,14 @@
 		An input field for single-line text.
 	</brief_description>
 	<description>
-		[LineEdit] provides an input field for editing a single line of text. It features many built-in shortcuts that are always available ([kbd]Ctrl[/kbd] here maps to [kbd]Cmd[/kbd] on macOS):
+		[LineEdit] provides an input field for editing a single line of text.
+		- When the [LineEdit] control is focused using the keyboard arrow keys, it will only gain focus and not enter edit mode.
+		- To enter edit mode, click on the control with the mouse or press the "ui_text_submit" action (default: [kbd]Enter[/kbd] or [kbd]Kp Enter[/kbd]).
+		- To exit edit mode, press "ui_text_submit" or "ui_cancel" (default: [kbd]Escape[/kbd]) actions.
+		- Check [method is_editing] and [signal editing_toggled] for more information.
+		[b]Important:[/b]
+		- Focusing the [LineEdit] with "ui_focus_next" (default: [kbd]Tab[/kbd]) or "ui_focus_prev" (default: [kbd]Shift + Tab[/kbd]) or [method Control.grab_focus] still enters edit mode (for compatibility).
+		[LineEdit] features many built-in shortcuts that are always available ([kbd]Ctrl[/kbd] here maps to [kbd]Cmd[/kbd] on macOS):
 		- [kbd]Ctrl + C[/kbd]: Copy
 		- [kbd]Ctrl + X[/kbd]: Cut
 		- [kbd]Ctrl + V[/kbd] or [kbd]Ctrl + Y[/kbd]: Paste/"yank"
@@ -137,6 +144,12 @@
 			<param index="0" name="text" type="String" />
 			<description>
 				Inserts [param text] at the caret. If the resulting value is longer than [member max_length], nothing happens.
+			</description>
+		</method>
+		<method name="is_editing" qualifiers="const">
+			<return type="bool" />
+			<description>
+				Returns whether the [LineEdit] is being edited.
 			</description>
 		</method>
 		<method name="is_menu_visible" qualifiers="const">
@@ -301,6 +314,12 @@
 		</member>
 	</members>
 	<signals>
+		<signal name="editing_toggled">
+			<param index="0" name="toggled_on" type="bool" />
+			<description>
+				Emitted when the [LineEdit] switches in or out of edit mode.
+			</description>
+		</signal>
 		<signal name="text_change_rejected">
 			<param index="0" name="rejected_substring" type="String" />
 			<description>
--- a/scene/gui/line_edit.cpp
+++ b/scene/gui/line_edit.cpp
@@ -45,6 +45,70 @@
 #include "editor/editor_settings.h"
 #endif
 
+void LineEdit::_edit() {
+	if (!is_inside_tree()) {
+		return;
+	}
+
+	if (!has_focus()) {
+		grab_focus();
+	}
+
+	if (!editable || editing) {
+		return;
+	}
+
+	editing = true;
+	_validate_caret_can_draw();
+
+	DisplayServer::WindowID wid = get_window() ? get_window()->get_window_id() : DisplayServer::INVALID_WINDOW_ID;
+	if (wid != DisplayServer::INVALID_WINDOW_ID && DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_IME)) {
+		DisplayServer::get_singleton()->window_set_ime_active(true, wid);
+		Point2 pos = Point2(get_caret_pixel_pos().x, (get_size().y + theme_cache.font->get_height(theme_cache.font_size)) / 2) + get_global_position();
+		if (get_window()->get_embedder()) {
+			pos += get_viewport()->get_popup_base_transform().get_origin();
+		}
+		DisplayServer::get_singleton()->window_set_ime_position(pos, wid);
+	}
+
+	show_virtual_keyboard();
+	queue_redraw();
+	emit_signal(SNAME("editing_toggled"), true);
+}
+
+void LineEdit::_unedit() {
+	if (!editing) {
+		return;
+	}
+
+	editing = false;
+	_validate_caret_can_draw();
+
+	DisplayServer::WindowID wid = get_window() ? get_window()->get_window_id() : DisplayServer::INVALID_WINDOW_ID;
+	if (wid != DisplayServer::INVALID_WINDOW_ID && DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_IME)) {
+		DisplayServer::get_singleton()->window_set_ime_position(Point2(), wid);
+		DisplayServer::get_singleton()->window_set_ime_active(false, wid);
+	}
+	ime_text = "";
+	ime_selection = Point2();
+	_shape();
+	set_caret_column(caret_column); // Update scroll_offset.
+
+	if (DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_VIRTUAL_KEYBOARD) && virtual_keyboard_enabled) {
+		DisplayServer::get_singleton()->virtual_keyboard_hide();
+	}
+
+	if (deselect_on_focus_loss_enabled && !selection.drag_attempt) {
+		deselect();
+	}
+
+	emit_signal(SNAME("editing_toggled"), false);
+}
+
+bool LineEdit::is_editing() const {
+	return editing;
+}
+
 void LineEdit::_swap_current_input_direction() {
 	if (input_direction == TEXT_DIRECTION_LTR) {
 		input_direction = TEXT_DIRECTION_RTL;
@@ -52,7 +116,6 @@ void LineEdit::_swap_current_input_direction() {
 		input_direction = TEXT_DIRECTION_LTR;
 	}
 	set_caret_column(get_caret_column());
-	queue_redraw();
 }
 
 void LineEdit::_move_caret_left(bool p_select, bool p_move_by_word) {
@@ -240,6 +303,11 @@ void LineEdit::_delete(bool p_word, bool p_all_to_right) {
 }
 
 void LineEdit::unhandled_key_input(const Ref<InputEvent> &p_event) {
+	// Return to prevent editing if just focused.
+	if (!editing) {
+		return;
+	}
+
 	Ref<InputEventKey> k = p_event;
 
 	if (k.is_valid()) {
@@ -265,26 +333,38 @@ void LineEdit::gui_input(const Ref<InputEvent> &p_event) {
 
 	Ref<InputEventMouseButton> b = p_event;
 
-	if (b.is_valid()) {
-		if (ime_text.length() != 0) {
-			// Ignore mouse clicks in IME input mode.
-			return;
-		}
-		if (b->is_pressed() && b->get_button_index() == MouseButton::RIGHT && context_menu_enabled) {
-			_update_context_menu();
-			menu->set_position(get_screen_position() + get_local_mouse_position());
-			menu->reset_size();
-			menu->popup();
-			grab_focus();
+	// Ignore mouse clicks in IME input mode.
+	if (b.is_valid() && ime_text.is_empty()) {
+		if (b->is_pressed() && b->get_button_index() == MouseButton::RIGHT) {
+			if (editable && !selection.enabled) {
+				set_caret_at_pixel_pos(b->get_position().x);
+			}
+
+			if (context_menu_enabled) {
+				_update_context_menu();
+				menu->set_position(get_screen_position() + get_local_mouse_position());
+				menu->reset_size();
+				menu->popup();
+			}
+
+			if (editable && !editing) {
+				_edit();
+			}
+
 			accept_event();
 			return;
 		}
 
-		if (is_middle_mouse_paste_enabled() && b->is_pressed() && b->get_button_index() == MouseButton::MIDDLE && is_editable() && DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_CLIPBOARD_PRIMARY)) {
+		if (editable && is_middle_mouse_paste_enabled() && b->is_pressed() && b->get_button_index() == MouseButton::MIDDLE && DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_CLIPBOARD_PRIMARY)) {
 			String paste_buffer = DisplayServer::get_singleton()->clipboard_get_primary().strip_escapes();
 
 			deselect();
 			set_caret_at_pixel_pos(b->get_position().x);
+
+			if (!editing) {
+				_edit();
+			}
+
 			if (!paste_buffer.is_empty()) {
 				insert_text_at_caret(paste_buffer);
 
@@ -295,7 +375,6 @@ void LineEdit::gui_input(const Ref<InputEvent> &p_event) {
 					text_changed_dirty = true;
 				}
 			}
-			grab_focus();
 			accept_event();
 			return;
 		}
@@ -304,10 +383,13 @@ void LineEdit::gui_input(const Ref<InputEvent> &p_event) {
 			return;
 		}
 
-		_reset_caret_blink_timer();
+		if (editing) {
+			_reset_caret_blink_timer();
+		}
+
 		if (b->is_pressed()) {
 			accept_event(); // Don't pass event further when clicked on text field.
-			if (!text.is_empty() && is_editable() && _is_over_clear_button(b->get_position())) {
+			if (editable && !text.is_empty() && _is_over_clear_button(b->get_position())) {
 				clear_button_status.press_attempt = true;
 				clear_button_status.pressing_inside = true;
 				queue_redraw();
@@ -330,7 +412,7 @@ void LineEdit::gui_input(const Ref<InputEvent> &p_event) {
 					const int triple_click_tolerance = 5;
 					const bool is_triple_click = !b->is_double_click() && (OS::get_singleton()->get_ticks_msec() - last_dblclk) < triple_click_timeout && b->get_position().distance_to(last_dblclk_pos) < triple_click_tolerance;
 
-					if (is_triple_click && text.length()) {
+					if (is_triple_click && !text.is_empty()) {
 						// Triple-click select all.
 						selection.enabled = true;
 						selection.begin = 0;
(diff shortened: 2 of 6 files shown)


WhalesState:
maybe we will need to add `grab_focus_edit()` to focus the `LineEdit` in edit mode directly.

AThousandShips:
You keep wiping your branch and closing the PR, I'd suggest making your changes and check the branch locally before pushing and causing such noise in the history 🙂

editor/animation_track_editor.cpp
```diff
@@ -2942,8 +2942,7 @@ void AnimationTrackEdit::gui_input(const Ref<InputEvent> &p_event) {
 		path_popup->set_position(get_screen_position() + path_rect.position - theme_ofs);
 		path_popup->set_size(path_rect.size);
 		path_popup->popup();
-		path->grab_focus();
-		path->set_caret_column(path->get_text().length());
+		path->grab_focus_edit();
```
WhalesState, line 2945: I forgot to add this line, and I don't want to make more noise! 😄
`path->set_caret_column(path->get_text().length());`

editor/gui/editor_file_dialog.cpp
```diff
@@ -64,8 +64,9 @@ void EditorFileDialog::popup_file_dialog() {
 void EditorFileDialog::_focus_file_text() {
 	int lp = file->get_text().rfind(".");
 	if (lp != -1) {
+		file->grab_focus_edit();
 		file->select(0, lp);
-		file->grab_focus();
+		file->set_caret_column(lp);
```
WhalesState, line 69: need to keep the caret at the start of the `LineEdit` too, because pressing left button to move the caret will move it to the start of the selection and will remove the selection.

scene/gui/file_dialog.cpp
```diff
@@ -52,10 +52,11 @@ void FileDialog::popup_file_dialog() {
 void FileDialog::_focus_file_text() {
 	int lp = file->get_text().rfind(".");
 	if (lp != -1) {
-		file->select(0, lp);
 		if (file->is_inside_tree() && !get_tree()->is_node_being_edited(file)) {
-			file->grab_focus();
+			file->grab_focus_edit();
 		}
+		file->select(0, lp);
+		file->set_caret_column(lp);
```
WhalesState, line 59: need to remove this line too.

editor/project_manager.cpp
```diff
@@ -843,7 +843,7 @@ void ProjectManager::shortcut_input(const Ref<InputEvent> &p_ev) {
 			} break;
 			case Key::F: {
 				if (k->is_command_or_control_pressed()) {
-					search_box->grab_focus();
+					this->search_box->grab_focus_edit();
```
KoBeWi, line 846: What's the `this` for?

doc/classes/LineEdit.xml
```diff
@@ -126,6 +126,14 @@
 				Returns the selection end column.
 			</description>
 		</method>
+		<method name="grab_focus_edit">
+			<return type="void" />
+			<param index="0" name="select_all" type="bool" default="false" />
+			<description>
+				Focus the [LineEdit] in edit mode.
+				If [param select_all] is [code]true[/code], it will select all [member text] and move the caret to it's length.
```
KoBeWi, line 134:
```suggestion
				If [param select_all] is [code]true[/code], it will select all [member text] and move the caret to its length.
```

doc/classes/LineEdit.xml
```diff
@@ -180,6 +194,13 @@
 				Selects the whole [String].
 			</description>
 		</method>
+		<method name="set_focus_edit">
+			<return type="void" />
+			<param index="0" name="focus_edit" type="bool" />
+			<description>
+				Toggles focus edit mode, if [member editable] is [code]false[/code] it will just focus the [LineEdit].
```
KoBeWi, line 201: This description sounds misleading. It should more clearly say that it affects behavior of `grab_focus()`.

editor/find_in_files.cpp
```diff
@@ -465,8 +465,7 @@ void FindInFilesDialog::_notification(int p_what) {
 		case NOTIFICATION_VISIBILITY_CHANGED: {
 			if (is_visible()) {
 				// Doesn't work more than once if not deferred...
-				callable_mp((Control *)_search_text_line_edit, &Control::grab_focus).call_deferred();
-				_search_text_line_edit->select_all();
+				callable_mp(_search_text_line_edit, &LineEdit::grab_focus_edit).bind(true).call_deferred();
```
KoBeWi, line 468:
```suggestion
				callable_mp(_search_text_line_edit, &LineEdit::grab_focus_edit).call_deferred(true);
```

editor/gui/editor_spin_slider.cpp
```diff
@@ -640,8 +640,7 @@ void EditorSpinSlider::_focus_entered() {
 	value_input->set_focus_next(find_next_valid_focus()->get_path());
 	value_input->set_focus_previous(find_prev_valid_focus()->get_path());
 	callable_mp((CanvasItem *)value_input_popup, &CanvasItem::show).call_deferred();
-	callable_mp((Control *)value_input, &Control::grab_focus).call_deferred();
-	callable_mp(value_input, &LineEdit ::select_all).call_deferred();
+	callable_mp(value_input, &LineEdit::grab_focus_edit).bind(true).call_deferred();
```
KoBeWi, line 643:
```suggestion
	callable_mp(value_input, &LineEdit::grab_focus_edit).call_deferred(true);
```

editor/create_dialog.cpp
```diff
@@ -474,8 +474,7 @@ void CreateDialog::_notification(int p_what) {
 
 		case NOTIFICATION_VISIBILITY_CHANGED: {
 			if (is_visible()) {
-				callable_mp((Control *)search_box, &Control::grab_focus).call_deferred(); // Still not visible.
-				search_box->select_all();
+				callable_mp(search_box, &LineEdit::grab_focus_edit).bind(true).call_deferred(); // Still not visible.
```
KoBeWi, line 477:
```suggestion
				callable_mp(search_box, &LineEdit::grab_focus_edit).call_deferred(true); // Still not visible.
```

editor/plugins/visual_shader_editor_plugin.cpp
```diff
@@ -4251,8 +4251,7 @@ void VisualShaderEditor::_show_members_dialog(bool at_mouse_pos, VisualShaderNod
 	Vector2 difference = (dialog_rect.get_end() - window_rect.get_end()).max(Vector2());
 	members_dialog->set_position(members_dialog->get_position() - difference);
 
-	callable_mp((Control *)node_filter, &Control::grab_focus).call_deferred(); // Still not visible.
-	node_filter->select_all();
+	callable_mp(node_filter, &LineEdit::grab_focus_edit).bind(true).call_deferred(); // Still not visible.
```
KoBeWi, line 4254:
```suggestion
	callable_mp(node_filter, &LineEdit::grab_focus_edit).call_deferred(true); // Still not visible.
```

scene/gui/line_edit.cpp
```diff
@@ -45,6 +45,84 @@
...
+	if (focus_edit == p_focus_edit) {
+		return;
+	}
+
+	if (p_focus_edit && !editable) {
+		return;
+	}
+
+	focus_edit = p_focus_edit;
+	emit_signal(SNAME("focus_edit_toggled"), focus_edit);
```
KoBeWi, line 78: This signal is emitted when calling the method, which is something we usually try to avoid. I think it can be handled by connecting to `focus_entered` and checking whether the node is being edited.
Though if it simplifies things, maybe it's justified in this case.

scene/gui/line_edit.cpp
```diff
@@ -45,6 +45,84 @@
...
 
+void LineEdit::grab_focus_edit(bool p_select_all) {
+	if (!is_inside_tree()) {
+		return;
+	}
+	set_focus_edit(true);
+
+	if (p_select_all) {
+		select_all();
+		set_caret_column(text.length());
```
KoBeWi, line 56: Doesn't `select_all()` already set the caret?

scene/gui/line_edit.cpp
```diff
@@ -244,6 +327,23 @@ void LineEdit::gui_input(const Ref<InputEvent> &p_event) {
 
 	Ref<InputEventMouseButton> b = p_event;
 
+	if (!focus_edit) {
+		Ref<InputEventKey> k = p_event;
```
KoBeWi, line 331: This should be after the `return` block.

KoBeWi:
The SpinBox changes brought back #78079.

editor/project_manager.cpp
WhalesState: Maybe it was added by mistake, because i have copied the changes while working on 4.2.1 stable branch.

doc/classes/LineEdit.xml
WhalesState: `grab_focus_edit()` is an alias for 
```cpp
LineEdit::grab_focus();
LineEdit::set_focus_edit(true);
LineEdit::select_all(); // optionally.
```
This is similar to the old behavior of `grab_focus()`, and yes `set_focus_edit(true)` is the same as `grab_focus_edit(false)` so maybe we can only expose grab_focus_edit() because it's more easier to notice when we type `grab_focus`. and maybe we can add release_focus_edit() to set `focus_edit` to false.

scene/gui/line_edit.cpp
WhalesState: i think it sets the caret to the start of the selection by default, since pressing the right arrow will move the caret to the end of the selection and remove the selection, and I got used to the old behavior so it may be confusing to change it.

doc/classes/LineEdit.xml
KoBeWi: Yeah that sounds like what I suggested below, but with different names.

KoBeWi:
```
C:\godot_source\core/object/message_queue.cpp:223 - Error calling deferred method: 'LineEdit::LineEdit::edit': Method expected 1 arguments, but called with 0.
```
Since `edit()` takes an argument now, every `call_deferred()` needs to be fixed, because `callable_mp()` does not support default arguments.

scene/gui/spin_box.h
```diff
@@ -65,8 +65,7 @@ class SpinBox : public Range {
 		double diff_y = 0.0;
 	} drag;
 
-	void _line_edit_focus_enter();
-	void _line_edit_focus_exit();
+	void _line_edit_editing_toggled(bool toggled_on);
```
KoBeWi, line 68:
```suggestion
	void _line_edit_editing_toggled(bool p_toggled_on);
```

WhalesState:
> Which made me realize that this breaks compatibility. I ignored it before, because you can just press Enter and still edit the LineEdit, but I think users who used `grab_edit()` to edit their controls will now be confused and might not realize that they need to press Enter. Not sure if it's worth handing, it would require a property to make the new behavior optional. Though defaulting to the old behavior would make the new one less discoverable, so idk.

We can optionally return back the old behavior, but the main issue was the focus loss when pressing `escape`, also we can allow force editing when focused but this is also useless since the new changes will force editing when focused with `tab` or `shift + tab` by default but not the arrow keys. The best solution is to note the new changes in docs, using `edit()` will also grab focus if it's unfocused or not editable, all they need to do is to change `grab_focus()` with `edit()` so we can also not this in the `grab_focus()` and `release_focus()` discriptions.

WhalesState:
may we can enter edit mode directly for the first time when it's not focused with keyboard, and we exit back to focus mode when the escape key is pressed to avoid focus loss and be able to select another GUI with keyboard.

KoBeWi:
Makes sense I think. You can still use Tab to change focus in edit mode.

KoBeWi:
@WhalesState Looks like you erased your changes by mistake (I assume).

WhalesState:
There was an issue while undoing the commit, it was reverting the branch publishing instead of the last commit.

Now with this changes, LineEdit will enter focus mode only when it's focused with arrow keys, using `grab_focus()` this will only prevent focus loss, and will alow users to exist edit mode by pressing `Enter` or `Escape` to submit or reject the text.

WhalesState:
@KoBeWi, Please let me know if any changes are required.


https://github.com/user-attachments/assets/79e8ae51-db91-476b-ac52-7c8ebb29f612

doc/classes/LineEdit.xml
```diff
@@ -301,6 +307,12 @@
 		</member>
 	</members>
 	<signals>
+		<signal name="editing_toggled">
+			<param index="0" name="toggled_on" type="bool" />
+			<description>
+				Emitted when the [LineEdit] switches between focus and edit modes.
```
WhalesState, line 313: This isn't clear, since edit mode insures focus, maybe `select and edit modes` or `focus and focus-edit modes` is better.

KoBeWi: `Emitted when the [LineEdit] switches in or out of edit mode.`?
I'd also explain edit mode in the class' description.

scene/gui/line_edit.cpp
```diff
@@ -45,6 +45,70 @@
...
+	}
+
+	if (!editable || editing) {
+		return;
+	}
+
+	editing = true;
+	_validate_caret_can_draw();
+
+	if (get_viewport()->get_window_id() != DisplayServer::INVALID_WINDOW_ID && DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_IME)) {
```
KoBeWi, line 64: Not sure if this is an equivalent of the old code with window ID.

scene/gui/line_edit.cpp
```diff
@@ -45,6 +45,70 @@
...
+
+	if (DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_VIRTUAL_KEYBOARD) && virtual_keyboard_enabled) {
+		DisplayServer::get_singleton()->virtual_keyboard_hide();
+	}
+
+	if (deselect_on_focus_loss_enabled) {
+		// deselect calls queue_redraw().
+		deselect();
+	} else {
+		queue_redraw();
```
KoBeWi, line 102: `set_caret_column()` also calls `queue_redraw()`.

scene/gui/line_edit.cpp
```diff
@@ -265,26 +334,37 @@ void LineEdit::gui_input(const Ref<InputEvent> &p_event) {
 
 	Ref<InputEventMouseButton> b = p_event;
 
-	if (b.is_valid()) {
-		if (ime_text.length() != 0) {
-			// Ignore mouse clicks in IME input mode.
-			return;
```
KoBeWi, line 271: What happened to this part?

scene/gui/line_edit.cpp
```diff
@@ -1132,43 +1235,19 @@ void LineEdit::_notification(int p_what) {
...
 
-			DisplayServer::WindowID wid = get_window() ? get_window()->get_window_id() : DisplayServer::INVALID_WINDOW_ID;
-			if (wid != DisplayServer::INVALID_WINDOW_ID && DisplayServer::get_singleton()->has_feature(DisplayServer::FEATURE_IME)) {
-				DisplayServer::get_singleton()->window_set_ime_active(true, wid);
-				Point2 pos = Point2(get_caret_pixel_pos().x, (get_size().y + theme_cache.font->get_height(theme_cache.font_size)) / 2) + get_global_position();
-				if (get_window()->get_embedder()) {
-					pos += get_viewport()->get_popup_base_transform().get_origin();
-				}
-				DisplayServer::get_singleton()->window_set_ime_position(pos, wid);
+			if (!(Input::get_singleton()->is_action_pressed("ui_up") || Input::get_singleton()->is_action_pressed("ui_down") || Input::get_singleton()->is_action_pressed("ui_left") || Input::get_singleton()->is_action_pressed("ui_right"))) {
```
KoBeWi, line 1238: You could add a comment to explain purpose of this condition.

scene/gui/line_edit.cpp
```diff
@@ -1661,7 +1740,7 @@ void LineEdit::clear() {
 	_text_changed();
 
 	// This should reset virtual keyboard state if needed.
-	if (has_focus()) {
+	if (has_focus() && editing) {
```
KoBeWi, line 1743: Same: can `editing` be true without focus?

KoBeWi (approved):
Tested basic functionality and it works great.
I noticed some IME change that I'm not sure about; they need to be assessed.

> This also can be applied to TextEdit, but to only exit editing mode by pressing escape and to not enter editing mode when it's focused with arrows.

Yeah this would be useful for TextEdit too.

scene/gui/line_edit.cpp
WhalesState: `_edit` and `_unedit` was missing the new changes related to some bug fixes, they should follow the old focus entered/exited code. Thank you.

scene/gui/line_edit.cpp
WhalesState: Combined both checks `if (b.is_valid() && ime_text.is_empty())` in one statement and returned back the comment.

WhalesState:
> Tested basic functionality and it works great. I noticed some IME change that I'm not sure about; they need to be assessed.

We have to make sure that the old code is still the same.

Changes in this PR:
- `NOTIFICATION_FOCUS_ENTER` old code was moved to `_edit()`. 
- `NOTIFICATION_FOCUS_EXIT` old code was moved to `_unedit()`.
- `String.length() == 0` and `String.length() != 0` was changed to `String.is_empty()` and `!String.is_empty()` for readability.
- Removed redundant `queue_redraw()` calls.

scene/gui/spin_box.cpp
```diff
@@ -602,8 +594,7 @@ SpinBox::SpinBox() {
 	line_edit->set_horizontal_alignment(HORIZONTAL_ALIGNMENT_LEFT);
 
 	line_edit->connect("text_submitted", callable_mp(this, &SpinBox::_text_submitted), CONNECT_DEFERRED);
-	line_edit->connect(SceneStringName(focus_entered), callable_mp(this, &SpinBox::_line_edit_focus_enter), CONNECT_DEFERRED);
-	line_edit->connect(SceneStringName(focus_exited), callable_mp(this, &SpinBox::_line_edit_focus_exit), CONNECT_DEFERRED);
+	line_edit->connect("editing_toggled", callable_mp(this, &SpinBox::_line_edit_editing_toggled), CONNECT_DEFERRED);
```
WhalesState, line 597: Is it better use `SNAME("editing_toggled")` and `SNAME("text_submitted")` because those `StringName`s are already defined in `LineEdit`.

doc/classes/LineEdit.xml
```diff
@@ -4,7 +4,15 @@
 		An input field for single-line text.
 	</brief_description>
 	<description>
-		[LineEdit] provides an input field for editing a single line of text. It features many built-in shortcuts that are always available ([kbd]Ctrl[/kbd] here maps to [kbd]Cmd[/kbd] on macOS):
+		[LineEdit] provides an input field for editing a single line of text.
+		[b]New behavior:[/b]
+		- When the [LineEdit] control is focused using keyboard arrow keys, it will only gain focus and not enter edit mode.
+		- To enter edit mode, click on the control with the mouse or press the "ui_text_submit" action (default: Enter or Kp Enter keys).
```
AThousandShips, line 10:
```suggestion
		- To enter edit mode, click on the control with the mouse or press the "ui_text_submit" action (default: [kbd]Enter[/kbd] or [kbd]Kp Enter[/kbd]).
```

doc/classes/LineEdit.xml
```diff
@@ -4,7 +4,15 @@
...
+		- To exit edit mode, press "ui_text_submit" or "ui_cancel" (default: Escape key) actions.
+		- Check [method is_editing] and [signal editing_toggled] for more information.
+		[b]Important:[/b]
+		- Focusing the [LineEdit] with "ui_focus_next" (default: Tab key) or "ui_focus_prev" (default: Shift+Tab key) or [method Control.grab_focus] still enters edit mode (for compatibility).
```
AThousandShips, line 14:
```suggestion
		- Focusing the [LineEdit] with "ui_focus_next" (default: [kbd]Tab[/kbd]) or "ui_focus_prev" (default: [kbd]Shift + Tab[/kbd]) or [method Control.grab_focus] still enters edit mode (for compatibility).
```

AThousandShips (approved):

doc/classes/LineEdit.xml
```diff
@@ -4,7 +4,15 @@
 		An input field for single-line text.
 	</brief_description>
 	<description>
-		[LineEdit] provides an input field for editing a single line of text. It features many built-in shortcuts that are always available ([kbd]Ctrl[/kbd] here maps to [kbd]Cmd[/kbd] on macOS):
+		[LineEdit] provides an input field for editing a single line of text.
+		[b]New behavior:[/b]
```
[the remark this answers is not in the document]
WhalesState: Which line? I Just wanted to move the new changes after the short description and before the shortcuts list.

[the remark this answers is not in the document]
WhalesState: Got it, `[b]New behavior:[/b]` line was removed, i thought it's about the ordering.

akien-mga:
Thanks!

PR #80418, opened 2023-08-08
Core: Add parameter `no_inheritance` to `Script.get_script_*_list()` methods

dalexeev:
This PR adds optional `no_inheritance` parameter to the following methods:

* `Script.get_script_method_list()`;
* `Script.get_script_property_list()`;
* `Script.get_script_signal_list()`.

Passing `true` returns info about the members of the current class only, no base classes. The `no_inheritance` parameter is already in the corresponding `ClassDB` methods. Also in GDScript there are private methods with the inverted `p_include_base` parameter.

I needed this functionality while working on #80411 ([code](https://github.com/godotengine/godot/blob/206be5dec72a3e67b8bcd758eb86d60b67d46fa2/editor/connections_dialog.cpp#L1286-L1300)). Also I think it can be needed in user code (like plugins) as the alternative is to get the full list for each base class and exclude the items of the parent class.

* Also closes godotengine/godot-proposals#10402.

PR #87680, opened 2024-01-28
Add methods to get argument count of methods

### LINKED ISSUE #8596: Add a method to get a Callable's argument count

Editor plugins

Callable.unbind() allows unbinding a number of arguments, but there is no way to dynamically know how many arguments are allowed/should be unbound

A new method on Callable, called get_argument_count() or get_allowed_arguments_count() that returns the number of arguments allowed by the callable 

Here are 2 example usages:

To unbind some unknown Callable for a signal:
```
signal.connect(callable.unbind(callable.get_bound_arguments_count() - callable.get_allowed_arguments_count()))
```

Example of a reusable, JavaScript style, array mapper. Only possible if this is implemented (optional index and array)
```
class_name Utility 
static func map(arr: Array, callable: Callable):
    var result := []
    for (i = 0; i < arr.size(); i += 1):
        result.push_back(
            callable.callv(
                [arr[i], i, arr].slice(0, callable.get_allowed_argument_count())
            )
        )
    return result 
```
Which could be used like so:
```
Utility.map(
    [1, 2, 3],
    func(num): return num * 2
)
Utility.map(
    [0, 5, 10],
    func(num, i): return str(i) + ": " str(num)
)
Utility.map(
    [8, 15, 11],
    func(num, i, arr): return num / arr.size()
)
```

It's only possible with known Callables, but it's not currently possible to implement something like the above utility function

I don't think an add-on can do this

AThousandShips:
Added to:
* `Callable`s
* `Object`
* `ClassDB`
* `Script(Instance)`s

Todo:
- [x] Document
- [x] Add tests
- [x] Implement for C# (needs confirmation)
- [x] Implement for GDExtension (confirmed)
- [x] Implement for godot-cpp (needs confirmation)

Help with how to do this in C# would be very welcome, I have some ideas but I'm not very familiar with the C# internals

To confirm the godot-cpp side, see:
* https://github.com/godotengine/godot-cpp/pull/1375

Might split off the extension part into a separate PR if desired

---

* Closes: https://github.com/godotengine/godot-proposals/issues/8596


core/object/object.h
```diff
@@ -856,6 +856,7 @@ class Object {
 	Variant property_get_revert(const StringName &p_name) const;
 
 	bool has_method(const StringName &p_method) const;
+	int get_method_argument_count(const StringName &p_method) const;
```
AThousandShips, line 859: Unsure if I should add an equivalent method to `ClassDB`

AThousandShips:
Changing so that when there's no defined number, i.e. when the Callable is invalid, or when there's no such method, it returns -1 to distinguish, will push soon

AThousandShips:
Unsure if the various methods should error instead of silently returning `0`, but leaving as is right now

modules/mono/managed_callable.cpp
AThousandShips, line 1: Unfamiliar with C# and if this needs any specific treatment here or if I should just leave it alone

modules/mono/signal_awaiter_utils.cpp
AThousandShips, line 1: Same here

core/variant/callable.h
```diff
@@ -153,6 +154,7 @@ class CallableCustom {
 	virtual void call(const Variant **p_arguments, int p_argcount, Variant &r_return_value, Callable::CallError &r_call_error) const = 0;
 	virtual Error rpc(int p_peer_id, const Variant **p_arguments, int p_argcount, Callable::CallError &r_call_error) const;
 	virtual const Callable *get_base_comparator() const;
+	virtual int get_argument_count() const = 0;
```
AThousandShips, line 157: Keeping this virtual for now, but once I've identified the use cases I will probably make it non-virtual, just wanting to make sure I don't miss anything anywhere

AThousandShips:
Tested the Extension side in godot-cpp, will open a PR for that too but can confirm it works with the relevant changes

AThousandShips:
I believe the GDExtension part is ready, but would appreciate some feedback on that

CC @godotengine/gdextension 

Ditto for C# side

CC @godotengine/dotnet

AThousandShips:
Lot of details to work out but ready for review by the relevant teams, will add tests maybe tomorrow, at least in GDScript

Mickeon:
This may be the first time function arguments are acknowledged with method names. It has me pondering if this should be the standard across the whole class reference ("parameter" vs "argument")

doc/classes/Object.xml
```diff
@@ -682,6 +682,14 @@
 				Returns the object's metadata entry names as a [PackedStringArray].
 			</description>
 		</method>
+		<method name="get_method_argument_count" qualifiers="const">
+			<return type="int" />
+			<param index="0" name="method" type="StringName" />
+			<description>
+				Returns the number of arguments of the given [param method].
```
Mickeon, line 689: Consistency with `has_method`.
```suggestion
				Returns the number of arguments of the given [param method] by name.
```

AThousandShips:
Not the first time, see `get_bound_arguments_count` etc.

doc/classes/ClassDB.xml
```diff
@@ -65,6 +65,15 @@
 				Returns an array with the names all the integer constants of [param class] or its ancestry.
 			</description>
 		</method>
+		<method name="class_get_method_argument_count" qualifiers="const">
+			<return type="int" />
+			<param index="0" name="class" type="StringName" />
+			<param index="1" name="method" type="StringName" />
+			<param index="2" name="no_inheritance" type="bool" default="false" />
+			<description>
+				Returns the number of arguments of the method [param method] of [param class] or its ancestry if [param no_inheritance] is [code]false[/code].
```
Mickeon, line 74: So many "of"s here. Perhaps replacing one would still make sense?
```suggestion
				Returns the number of arguments for the method [param method] of [param class] or its ancestry if [param no_inheritance] is [code]false[/code].
```

doc/classes/Callable.xml
```diff
@@ -112,6 +112,12 @@
 				Calls the method represented by this [Callable]. Unlike [method call], this method expects all arguments to be contained inside the [param arguments] [Array].
 			</description>
 		</method>
+		<method name="get_argument_count" qualifiers="const">
+			<return type="int" />
+			<description>
+				Returns the total number of arguments of this [Callable], accounting for any bound (or unbound) arguments. The number of bound arguments is subtracted from the total, and the number of unbound arguments are added.
```
Mickeon, line 118: I understand what this is trying to say but it is a bit confusing when worded like this. I have no suggestions at the moment.

AThousandShips: Me too, was trying to figure out how to say it

doc/classes/ClassDB.xml
AThousandShips: I'm not so sure `for` is grammatical here, it is the arguments *of* the method, but will see what I can do with the wording

doc/classes/Callable.xml
Mickeon: Okay, I'll help you think. From the outsider developer's perspective, what changes? What needs to be known? What will this method be used for?

Judging by the code, I would say the important part to get here is that what we mean by "arguments" is "bunch of values that _can_ (oftentimes _should_) be passed into the Callable"
```suggestion
				Returns the total number of arguments this [Callable] should take, including optional arguments.
```
The part about `bind` and `unbind` is a consequence of this definition.

AThousandShips: I like that, I'd keep the bind/unbind explicit as it can be very, *very* confusing (I got it backwards the first time I wrote the code and added instead of subtracted), but I'll take that suggestion as a base 🙂 thank you!

Mickeon: Oh by all means, it needs to be specified to remove any doubts. It's just that -- and I repeat myself -- by rewording the definition this way it makes the behavior of `bind` and `unbind` look like a natural consequence of it, instead of a potentially confusing exception to the norm.

core/extension/gdextension_interface.h
```diff
@@ -420,6 +422,8 @@ typedef struct {
 	GDExtensionCallableCustomLessThan less_than_func;
 
 	GDExtensionCallableCustomToString to_string_func;
+
+	GDExtensionCallableCustomGetArgumentCount get_argument_count_func;
 } GDExtensionCallableCustomInfo;
```
Bromeon, line 427: I think this is a breaking change:

If someone running a 4.2 extension declares the struct on the binding side, it won't have that last field. If this extension is then run in Godot 4.3, Godot will try to access `get_argument_count_func`, causing UB.

Unfortunately it might need a `GDExtensionCallableCustomInfo2` iteration, but this also impacts other functions 😕 ideas?

AThousandShips: Then I'll leave the the extension side to future implementation perhaps, wasn't sure hence asking, will look at those details but will split it off to keep it simple on this side 🙂 thank you

Bromeon: If there are other similar additions in the future (e.g. advanced argument queries for type/default-ness), it could make sense to batch some changes in a minor release, to avoid proliferation of struct iterations.

AThousandShips: Indeed, will open a new PR to discuss this more coherently and make a new structure with the related changes there and update the godot-cpp PR, will link here when done

AThousandShips:
Re-added the GDExtension code, with new compatibility details, works well in my testing (but haven't tested the ScriptInstance extension part as godot-cpp doesn't utilize it)

core/extension/gdextension_interface.h
```diff
@@ -2312,6 +2392,7 @@ typedef GDExtensionScriptInstancePtr (*GDExtensionInterfaceScriptInstanceCreate)
 /**
  * @name script_instance_create2
  * @since 4.2
+ * @deprecated in Godot 4.3. Use `script_instance_create3` instead.
```
AThousandShips, line 2395: Forgot to update the line in the `script_instance_create` part, will do later

core/object/script_language_extension.h
```diff
@@ -101,6 +101,8 @@ class ScriptExtension : public Script {
 	EXBIND1RC(bool, has_method, const StringName &)
 	EXBIND1RC(bool, has_static_method, const StringName &)
 
+	EXBIND1RC(int, get_method_argument_count, const StringName &)
```
AThousandShips, line 104: Could remove this part from the extension and rely on `get_method_info` but it might not provide all the necessary details, but it would make extending easier and only affect the `Callable` part

Just as an option to reduce the extent of extension changes

modules/mono/managed_callable.cpp
```diff
@@ -85,6 +85,11 @@ ObjectID ManagedCallable::get_object() const {
 	return CSharpLanguage::get_singleton()->get_managed_callable_middleman()->get_instance_id();
 }
 
+int ManagedCallable::get_argument_count() const {
+	// TODO: I'm not sure if this is needed as I don't know where this is used
+	return 0;
```
raulsntos, line 90: This would be called when you call `get_argument_count` on a `Callable` created from a C# delegate/lambda. You'd need to create a method in `DelegateUtils.cs` like the `DelegateEquals` and `DelegateHash` methods because we can only retrieve the delegate information from C#.

AThousandShips:
Added basic tests both for GDScript and in tests

Can look at adding more general callable tests in the future but out of scope for this

modules/mono/managed_callable.cpp
AThousandShips: Good, then I will see how that can work out, should be workable

modules/mono/signal_awaiter_utils.cpp
```diff
@@ -90,6 +90,11 @@ ObjectID SignalAwaiterCallable::get_object() const {
 	return target_id;
 }
 
+int SignalAwaiterCallable::get_argument_count() const {
+	// TODO: I'm not sure if this is needed as I don't know where this is used
+	return 0;
```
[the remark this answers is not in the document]
AThousandShips: Then I'll just have them defaulted 🙂 Thank you!

modules/mono/managed_callable.cpp
AThousandShips: I've solved this one, and discovered that the signal awaiters are accessible from the GDScript side with `get_connections` on a `Signal` so trying out some code, will push soon

AThousandShips:
The C# side should be complete now, I'll leave the event signal handlers alone as internal details

Will work some more on the documentation tomorrow or the day after

Waiting for some more feedback from C# and GDExtension on the specifics here

modules/mono/signal_awaiter_utils.h
AThousandShips, line 1: Might make these warn when calling them and indicate they're undefined, alternatively add a warning to `CallableCustom` noting it's not implemented for this case

core/extension/gdextension_interface.h
Bromeon: I see you added a `GDExtensionScriptInstanceInfo3`, but is this single niche feature really worth another iteration of the struct -- which means extra compat code on both Godot + binding side?

I would really consider holding out until we can pack more into this; otherwise we'll end up with `GDExtensionScriptInstanceInfo7` soon 😉

AThousandShips: Thinking a bit further I might restrict it to the script side and not on the instance side, we'll see!

AThousandShips:
Reorganized the code, added validity return data for all the checks in `Object`, `ClassDB`, `Script`, and `ScriptInstance`, added a fallback method to the instances and separated the instance extension into its own commit for ease of integration

doc/classes/ScriptExtension.xml
```diff
@@ -80,6 +80,12 @@
 			<description>
 			</description>
 		</method>
+		<method name="_get_script_method_argument_count" qualifiers="virtual const">
```
AThousandShips, line 83: This now expects either an `int` if valid, or `null` if not, will see about documenting but this file is sparse already so will see

modules/mono/csharp_script.cpp
```diff
@@ -1792,6 +1792,34 @@ bool CSharpInstance::has_method(const StringName &p_method) const {
 			gchandle.get_intptr(), &p_method);
 }
 
+int CSharpInstance::get_method_argument_count(const StringName &p_method, bool *r_is_valid) const {
```
AThousandShips, line 1795: Will look at writing an instance bridge method for this like the `has_method` part, will see if needed

core/object/object.cpp
```diff
@@ -679,6 +679,58 @@ bool Object::has_method(const StringName &p_method) const {
...
+			if (r_is_valid) {
+				*r_is_valid = true;
+			}
+			return ret;
+		}
+	}
+
+	{
+		bool valid = false;
+		int ret = ClassDB::get_method_argument_count(get_class_name(), p_method, &valid);
```
AThousandShips, line 707: Now delegates to the `ClassDB` case instead, keeping the code in one place

core/object/script_instance.cpp
```diff
@@ -32,6 +32,28 @@
 
 #include "core/object/script_language.h"
 
+int ScriptInstance::get_method_argument_count(const StringName &p_method, bool *r_is_valid) const {
```
AThousandShips, line 35: Simple delegation here, very much like how it's done in GDScript but without the dedicated features used there, allows for using this feature without extending `ScriptInstance`, though it might not catch all cases, and might not be as optimal

core/object/script_language.cpp
```diff
@@ -102,6 +102,22 @@ Dictionary Script::_get_script_constant_map() {
 	return ret;
 }
 
+int Script::get_script_method_argument_count(const StringName &p_method, bool *r_is_valid) const {
```
AThousandShips, line 105: Fallback in the same way, this means that as long as an extended script has `get_method_info` with appropriate return information you can utilize the whole chain of functionality, so this should work out of the box for existing extensions that provide method information

modules/mono/glue/GodotSharp/GodotSharp/Core/DelegateUtils.cs
```diff
@@ -44,6 +44,34 @@ internal static int DelegateHash(IntPtr delegateGCHandle)
...
+                    return 0;
+                }
+                var @method = @delegate.Method;
+                if (@method is null)
+                {
+                    *outIsValid = godot_bool.False;
+                    return 0;
+                }
+                *outIsValid = godot_bool.True;
+                return @method.GetParameters().Length;
```
raulsntos, line 65: Just use the null conditional operator and we don't have to repeat the null check return.

```suggestion
                var @delegate = (Delegate?)GCHandle.FromIntPtr(delegateGCHandle).Target;
                int? argCount = @delegate?.Method?.GetParameters().Length;
                if (argCount is null)
                {
                    *outIsValid = godot_bool.False;
                    return 0;
                }
                *outIsValid = godot_bool.True;
                return argCount.Value;
```

AThousandShips: Thank you! Never used these features (and haven't used C# to any real degree in over 10 years)

Will do

AThousandShips:
Will look at more tests later, but there's no real tests for a lot of the methods of `ClassDB` and `Object` so unsure how to go about that, so keeping it to tests for the `Callable`s for now

core/extension/gdextension_interface.h
dsnopek: I have an unrelated change that I've been wanting to make to `GDExtensionScriptInstanceInfo2` - I can try to make a PR for it for 4.3 so this isn't the only change to that struct :-)

AThousandShips: Then I'll keep this in 🙂

core/object/script_language_extension.h
```diff
@@ -101,6 +101,19 @@ class ScriptExtension : public Script {
...
+	virtual int get_script_method_argument_count(const StringName &p_method, bool *r_is_valid = nullptr) const override {
+		Variant ret;
+		if (GDVIRTUAL_CALL(_get_script_method_argument_count, p_method, ret)) {
+			if (r_is_valid) {
+				*r_is_valid = ret.get_type() == Variant::INT;
+			}
+			return (ret.get_type() == Variant::INT) ? ret.operator int() : 0;
+		}
+		// Fallback to default.
+		return Script::get_script_method_argument_count(p_method, r_is_valid);
```
AThousandShips, line 114: Unsure if it should fall back when the method doesn't return a valid value here instead, like:
```cpp
		Variant ret;
		if (GDVIRTUAL_CALL(_get_script_method_argument_count, p_method, ret) && ret.get_type() == Variant::INT) {
			if (r_is_valid) {
				*r_is_valid = true;
			}
			return ret.operator int();
		}
		// Fallback to default.
		return Script::get_script_method_argument_count(p_method, r_is_valid);
```
It'd allow falling back, but also custom behaviour, but unsure

core/variant/callable.h
```diff
@@ -153,6 +154,7 @@ class CallableCustom {
 	virtual void call(const Variant **p_arguments, int p_argcount, Variant &r_return_value, Callable::CallError &r_call_error) const = 0;
 	virtual Error rpc(int p_peer_id, const Variant **p_arguments, int p_argcount, Callable::CallError &r_call_error) const;
 	virtual const Callable *get_base_comparator() const;
+	virtual int get_argument_count(bool &r_is_valid) const;
```
AThousandShips, line 157: Using reference here to avoid unnecessary null checks

core/object/script_language.h
```diff
@@ -150,6 +150,8 @@ class Script : public Resource {
 	virtual bool has_method(const StringName &p_method) const = 0;
 	virtual bool has_static_method(const StringName &p_method) const { return false; }
 
+	virtual int get_script_method_argument_count(const StringName &p_method, bool *r_is_valid = nullptr) const;
```
AThousandShips, line 153: Named it this way as per the note for `has_method` above

core/object/script_language_extension.h
dsnopek: I'm not sure I understand. The alternative in your comment is to allow returning values that aren't integers but that could be converted to integers?

AThousandShips: No it's to fall back to the default if it doesn't return an `INT`

modules/gdscript/gdscript_utility_callable.cpp
```diff
@@ -80,6 +80,20 @@ ObjectID GDScriptUtilityCallable::get_object() const {
...
+	switch (type) {
+		case TYPE_GLOBAL:
+			r_is_valid = true;
+			return Variant::get_utility_function_argument_count(function_name);
+		case TYPE_GDSCRIPT:
+			r_is_valid = true;
+			return GDScriptUtilityFunctions::get_function_argument_count(function_name);
+		default:
+			r_is_valid = false;
+			return 0;
```
dalexeev, line 93: I would prefer an explicit `case TYPE_INVALID` instead of `default` for consistency. If we potentially add a new type and forget to add the new case here, the compiler will produce a warning that is treated as an error.

dsnopek (approved):
The GDExtension changes look good to me! I tested with godot-cpp, and everything seems to be working fine.

(I haven't reviewed the core, GDScript or C# changes - only GDExtension)

core/object/script_language_extension.h
AThousandShips: Then I'll add it to my next push

core/object/object.cpp
```diff
@@ -679,6 +679,58 @@ bool Object::has_method(const StringName &p_method) const {
...
+	if (scr != nullptr) {
+		bool valid = false;
+		int ret = scr->get_script_method_argument_count(p_method, &valid);
+		if (valid) {
+			if (r_is_valid) {
+				*r_is_valid = true;
+			}
+			return ret;
+		}
+	}
```
dalexeev, line 726: Shouldn't inheritance be checked here since `Script::get_script_method_argument_count()` doesn't take into account base classes, unlike `ScriptInstance::get_method_argument_count()`? (This is quite confusing. See also #80418.)

AThousandShips: Good point, I based this on how methods like `get_method_info` and `has_method` work in GDScript and Mono for the Script and ScriptInstance cases respectively, but open to other suggestions, can make this one iterate without trouble 🙂

AThousandShips: Also realized this is called on the class itself, got it confused with the static part in the case above, so should probably use `script` instead of casting itself, unsure though

Will add both for comparison

modules/gdscript/gdscript_utility_callable.cpp
AThousandShips: Will need a default path with a warning to accomplish that unfortunately, but will add that, it complains of non-returning path with just the three cases

AThousandShips:
Fixed some of the above pointers, squashed the basic extension things into the first commit (keeping the instance side separate for easier reworking) and did some other minor rework

modules/gdscript/gdscript_utility_callable.cpp
```diff
@@ -80,6 +80,22 @@ ObjectID GDScriptUtilityCallable::get_object() const {
...
+		case TYPE_GLOBAL:
+			r_is_valid = true;
+			return Variant::get_utility_function_argument_count(function_name);
+		case TYPE_GDSCRIPT:
+			r_is_valid = true;
+			return GDScriptUtilityFunctions::get_function_argument_count(function_name);
+		default:
+			ERR_FAIL_V_MSG(0, "Invalid type.");
+	}
```
dalexeev, line 96:
```suggestion
	}
	ERR_FAIL_V_MSG(0, "Invalid type.");
```

This will compile ans not suppress the warning if a case is accidentally missed.

core/object/object.cpp
```diff
@@ -679,6 +679,59 @@ bool Object::has_method(const StringName &p_method) const {
...
+		bool valid = false;
+		int ret = scr->get_script_method_argument_count(p_method, &valid);
+		if (valid) {
+			if (r_is_valid) {
+				*r_is_valid = true;
+			}
+			return ret;
+		}
+		scr = scr->get_base_script().ptr();
+	}
```
AThousandShips, line 727: Just calling it on the script itself and letting `ScriptInstance` handle the attached script as is done elsewhere

core/variant/callable.h
AThousandShips: Might go back to having this be virtual, but define a default behaviour to refer back to, that way new code doesn't forget to add it

akien-mga:
So this adds `GDExtensionScriptInstanceInfo3` for GDExtension. I'm not opposed to it, but it's worth considering whether we have other changes in the pipeline (implemented or planned) that would also need to change that struct, so we can bundle them all in the same release and avoid adding new structs all the time, making GDExtension full of backward compatibility structs. CC @godotengine/gdextension

core/extension/gdextension_interface.h
akien-mga: Let's try to get @dsnopek's change in with this one for 4.3 then, or keep both (and potential future additions) for 4.4 maybe?

AThousandShips: Agreed, both at the same time, keeping the extension code in its separate commit until we decide and alternatively splitting it off

AThousandShips:
Went ahead and split off the script instance part and will open a dedicated PR for it to make the review easier so we can merge this safely without timing it

core/extension/gdextension_interface.h
dsnopek: @akien-mga @AThousandShips Here's a PR with the change I had in mind: https://github.com/godotengine/godot/pull/89055

akien-mga:
I think this could now be squashed and merged?

AThousandShips:
As long as the C# side looks good I think we're good, I'll squash tomorrow if there's no questions

AThousandShips:
I'll squash it up soon and we can give it a spin, in my testing the c# code worked and was based on suggestions by raulsntos, so I feel confident in going forward, if there's any questions I can always split the C# off and leave the placeholders

raulsntos (approved):
The C# changes look good to me, thanks!

doc/classes/ScriptExtension.xml
KoBeWi: `Return the expected argument count for the given [param method], or [code]null[/code] if it can't be determined.`

I've seen new methods added here many times, but they never get a description because "the class is not documented anyway".

AThousandShips: I'll document it in the extension case, it doesn't have it doesn't actually handle `null` special rn, will look though

AThousandShips: I'll add it tomorrow, or in the follow up PR, someone (not me) should document all of these though

KoBeWi (approved):
The implementation looks fine overall.

akien-mga:
Thanks!

AThousandShips:
Thank you! Will improve the docs and fixup the followup PR tomorrow

PR #87686, opened 2024-01-28
[TreeItem] Add `get_button_color()`

radzo73:
### What?
Adds `TreeItem.get_button_color(column, id)`, a function that returns the color of a button in a `TreeItem`, typically set with `set_button_color`. Returns `Color()` if the specified button column-ID pair is not found in the `TreeItem`.

### Why?
Currently working on something that uses buttons in a tree to open up a color picker. I want to set the starting color of the picker to whatever color the button is, which I couldn't really do before this PR (meaning the starting color of the picker would either be the previously selected color or some default color like `Color.WHITE`).

BASED ON: master
CONFLICTS WITH: Nothing, as far as I know
CLOSES: Something, probably

### DIFF
--- a/doc/classes/TreeItem.xml
+++ b/doc/classes/TreeItem.xml
@@ -96,6 +96,14 @@
 				Returns the button index if there is a button with ID [param id] in column [param column], otherwise returns -1.
 			</description>
 		</method>
+		<method name="get_button_color" qualifiers="const">
+			<return type="Color" />
+			<param index="0" name="column" type="int" />
+			<param index="1" name="id" type="int" />
+			<description>
+				Returns the color of the button with ID [param id] in column [param column]. If the specified button does not exist, returns [constant Color.BLACK].
+			</description>
+		</method>
 		<method name="get_button_count" qualifiers="const">
 			<return type="int" />
 			<param index="0" name="column" type="int" />
--- a/scene/gui/tree.cpp
+++ b/scene/gui/tree.cpp
@@ -1222,6 +1222,12 @@ int TreeItem::get_button_by_id(int p_column, int p_id) const {
 	return -1;
 }
 
+Color TreeItem::get_button_color(int p_column, int p_index) const {
+	ERR_FAIL_INDEX_V(p_column, cells.size(), Color());
+	ERR_FAIL_INDEX_V(p_index, cells[p_column].buttons.size(), Color());
+	return cells[p_column].buttons[p_index].color;
+}
+
 void TreeItem::set_button_tooltip_text(int p_column, int p_index, const String &p_tooltip) {
 	ERR_FAIL_INDEX(p_column, cells.size());
 	ERR_FAIL_INDEX(p_index, cells[p_column].buttons.size());
@@ -1662,6 +1668,7 @@ void TreeItem::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_button_tooltip_text", "column", "button_index"), &TreeItem::get_button_tooltip_text);
 	ClassDB::bind_method(D_METHOD("get_button_id", "column", "button_index"), &TreeItem::get_button_id);
 	ClassDB::bind_method(D_METHOD("get_button_by_id", "column", "id"), &TreeItem::get_button_by_id);
+	ClassDB::bind_method(D_METHOD("get_button_color", "column", "id"), &TreeItem::get_button_color);
 	ClassDB::bind_method(D_METHOD("get_button", "column", "button_index"), &TreeItem::get_button);
 	ClassDB::bind_method(D_METHOD("set_button_tooltip_text", "column", "button_index", "tooltip"), &TreeItem::set_button_tooltip_text);
 	ClassDB::bind_method(D_METHOD("set_button", "column", "button_index", "button"), &TreeItem::set_button);
--- a/scene/gui/tree.h
+++ b/scene/gui/tree.h
@@ -266,6 +266,7 @@ public:
 	int get_button_id(int p_column, int p_index) const;
 	void erase_button(int p_column, int p_index);
 	int get_button_by_id(int p_column, int p_id) const;
+	Color get_button_color(int p_column, int p_index) const;
 	void set_button_tooltip_text(int p_column, int p_index, const String &p_tooltip);
 	void set_button(int p_column, int p_index, const Ref<Texture2D> &p_button);
 	void set_button_color(int p_column, int p_index, const Color &p_color);


doc/classes/TreeItem.xml
```diff
@@ -96,6 +96,14 @@
 				Returns the button index if there is a button with ID [param id] in column [param column], otherwise returns -1.
 			</description>
 		</method>
+		<method name="get_button_color" qualifiers="const">
+			<return type="Color" />
+			<param index="0" name="column" type="int" />
+			<param index="1" name="id" type="int" />
+			<description>
+				Returns the color of the button with ID [param id] in column [param column]. If the specified button does not exists, returns [Color].
```
AThousandShips, line 104:
```suggestion
				Returns the color of the button with ID [param id] in column [param column]. If the specified button does not exists, returns [code]Color()[/code].
```

scene/gui/tree.cpp
```diff
@@ -1222,6 +1222,12 @@ int TreeItem::get_button_by_id(int p_column, int p_id) const {
 	return -1;
 }
 
+Color TreeItem::get_button_color(int p_column, int p_index) const{
```
AThousandShips, line 1225:
```suggestion
Color TreeItem::get_button_color(int p_column, int p_index) const {
```

doc/classes/TreeItem.xml
```diff
@@ -96,6 +96,14 @@
 				Returns the button index if there is a button with ID [param id] in column [param column], otherwise returns -1.
 			</description>
 		</method>
+		<method name="get_button_color" qualifiers="const">
+			<return type="Color" />
+			<param index="0" name="column" type="int" />
+			<param index="1" name="id" type="int" />
+			<description>
+				Returns the color of the button with ID [param id] in column [param column]. If the specified button does not exists, returns [code]Color()[/code].
```
AThousandShips, line 104:
```suggestion
				Returns the color of the button with ID [param id] in column [param column]. If the specified button does not exists, returns [code skip-lint]Color()[/code].
```

doc/classes/TreeItem.xml
```diff
@@ -96,6 +96,14 @@
 				Returns the button index if there is a button with ID [param id] in column [param column], otherwise returns -1.
 			</description>
 		</method>
+		<method name="get_button_color" qualifiers="const">
+			<return type="Color" />
+			<param index="0" name="column" type="int" />
+			<param index="1" name="id" type="int" />
+			<description>
+				Returns the color of the button with ID [param id] in column [param column]. If the specified button does not exists, returns [code skip-lint]Color()[/code].
```
Mickeon, line 104: > Currently, there's no way to link to constuctors with BBCode style tags. I suppose I could have this method return Color.TRANSPARENT, because you are able to reference constants in the docs

There's no much point to reference constructors in the class reference, yet. Even for this example, you can replace it and it would be more understandable in my opinion. "Clearer", if you will.
```suggestion
				Returns the color of the button with ID [param id] in column [param column]. If the specified button does not exists, returns [constant Color.TRANSPARENT].
```

Mickeon:
Regardless, the other button "getters" do exist so this feels like an accidental omission, possibly because it was never needed by the editor.

Mickeon:
Could you also squash those commits? See https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html?highlight=squash%20commits#the-interactive-rebase

radzo73:
Commits should be squashed?

Mickeon:
Yes, please see the article I linked above. In short, all of the commits you've made need to be merged into one.

scene/gui/tree.cpp
```diff
@@ -1222,6 +1222,12 @@ int TreeItem::get_button_by_id(int p_column, int p_id) const {
 	return -1;
 }
 
+Color TreeItem::get_button_color(int p_column, int p_index) const {
+	ERR_FAIL_INDEX_V(p_column, cells.size(), Color::named("TRANSPARENT"));
+	ERR_FAIL_INDEX_V(p_index, cells[p_column].buttons.size(), Color::named("TRANSPARENT"));
```
Mickeon, line 1227: No reason to do this. The documentation can state the constant, but internally it can still be an empty Color constructor (as it should be).

radzo73: Except the constructor `Color()` returns the equivalent of `Color.BLACK` (or `Color(0, 0, 0, 0)` in C#)
so do I return a transparent color or update the docs to reference `Color()`?

AThousandShips: No it returns `Color(0, 0, 0, 1)`

Mickeon: Alright, it's settled then. `Color()` internally, and the documentation should refer to `Color.BLACK`. Forgive me, the original statement of the PR confused me.

radzo73:
-Docs no longer mention fail case. This makes it consistent with other documented functions such as `Gradient.get_color` and `GraphNode.get_input_port_color`
- `TreeItem.get_button_color` now returns `Color()` in its fail cases.
Squashed commits!

Mickeon:
Do mention the return value in case of failure, please. It was a nice touch. If anything the rest of the descriptions are doing it wrong by omitting this (but it needs to be addressed another time)

doc/classes/TreeItem.xml
```diff
@@ -96,6 +96,14 @@
 				Returns the button index if there is a button with ID [param id] in column [param column], otherwise returns -1.
 			</description>
 		</method>
+		<method name="get_button_color" qualifiers="const">
+			<return type="Color" />
+			<param index="0" name="column" type="int" />
+			<param index="1" name="id" type="int" />
+			<description>
+				Returns the color of the button with ID [param id] in column [param column]. If the specified button does not exist, returns [code skip-lint]Color()[/code].
```
Mickeon, line 104: Almost there! We agreed to refer to the constant in the description.
```suggestion
				Returns the color of the button with ID [param id] in column [param column]. If the specified button does not exist, returns [constant Color.BLACK].
```

radzo73:
- Docs mention fail case again. It does use `[code skip-lint]Color()[/code]` because we can't link to constructors using BBCode, sadly.
- Squashed commits

Mickeon (approved):
That aside, this makes sense to me and it has been an odd omission for a while.

doc/classes/TreeItem.xml
radzo73: The problem with doing that is that `Color()` doesn't always return the equivalent of `Color.BLACK`. In the C#, `Color()` instead returns the equivalent of `Color(0, 0, 0, 0)`.

So what should I do?

Mickeon: That's *even more* of a reason to use the constant to refer to it. The source code of the engine will not behave any differently, no matter what scripting language you use.

In C++ the color constructor returns `Color(0, 0, 0, 1)`, which corresponds to **Color.BLACK**. If you call this function in C# and it fails, the returned value is still `Color(0, 0, 0, 1)`.

radzo73:
Updated and squashed (again) (again)

akien-mga (approved):

akien-mga:
Thanks!

PR #87669, opened 2024-01-28
Fix possible deadlock when creating scripts during a background garbage collection

bs-mwoerner:
Fixes #87405.

I'd really appreciate someone else looking over this, but since holding on to the mutex seems like an oversight (compare with virtually the same code in the constructur) and this change is very localized, this might be a suitable quick fix for a problem that's rare but deadlocks the game and is hard to diagnose if it materializes.

PR #87688, opened 2024-01-28
Remove unnecessary `this->` expressions

AThousandShips:
Unnecessary and hurts readability IMO, also creates confusion over *why* someone would use this as it's redundant by default so when you look at the code you wonder why this was done, to me for example it makes it seem like it's accessing something in some other object, as you'd obviously never use `->` for the current object

Because we have shadowing warnings now any of these are absolutely redundant, as there can't be a method local variable with the same name

I suspect several of these were for previous shadowings without `p_` prefixes

### DIFF
--- a/modules/gdscript/gdscript_parser.h
+++ b/modules/gdscript/gdscript_parser.h
@@ -223,7 +223,7 @@ public:
 		}
 
 		bool operator!=(const DataType &p_other) const {
-			return !(this->operator==(p_other));
+			return !(*this == p_other);
 		}
 
 		void operator=(const DataType &p_other) {
--- a/core/math/convex_hull.cpp
+++ b/core/math/convex_hull.cpp
@@ -344,31 +344,31 @@ public:
 		Rational128(int64_t p_value) {
 			if (p_value > 0) {
 				sign = 1;
-				this->numerator = p_value;
+				numerator = p_value;
 			} else if (p_value < 0) {
 				sign = -1;
-				this->numerator = -p_value;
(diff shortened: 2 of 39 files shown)


modules/gdscript/gdscript_parser.h
```diff
@@ -223,7 +223,7 @@ class GDScriptParser {
 		}
 
 		bool operator!=(const DataType &p_other) const {
-			return !(this->operator==(p_other));
+			return !(operator==(p_other));
```
AThousandShips, line 226:
```suggestion
			return !(*this == p_other);
```
Already added but will update after CI has run

fire (approved):
The pr is removing unnecessary `this->` and so it shouldn't break stuff. I scanned the code and it seemed to be entirely removing `this->`. Looks good to me.

AThousandShips:
That's all the cases I could find by a search for it, save for the two fixed by:
* https://github.com/godotengine/godot/pull/87669

Also excluded cases like `p_this->` and `this->*` which are all valid

lawnjelly:
To hazard a guess, some of these (e.g. `Rational128`) may have been copy pasted from templates, where the `this->` is necessary to compile on some compilers afaik.

AThousandShips:
Will restore those just to be safe 🙂

AThousandShips:
Added a few commented cases just for completeness

akien-mga (approved):

akien-mga:
Thanks!

AThousandShips:
Thank you!

PR #87695, opened 2024-01-29
Adding astar update_point and removing table lookup from add_point

gladelynch:
I removed the Point table look up in every add_point call, which is there in order to give the method a dual purpose of updating points that already exist. Realistically add_point is called to create new Points much more frequently than to update existing points, and the performance overhead in the current implementation can be substantial as the grid size scales. Especially in 3d.

I separated it into 2 methods:
update_point will look for an existing Point in the grid and update its weight and position.
add_point will create new Points at the given id position, as it currently does.

Main concern here for me is that users who call add_point to update their Points instead of update_point could create some memory leakage due to creating a new Point every time. I'm not sure if there's a good way to prevent this without nullifying the point of this change. I'm not an experienced c++ programmer so others may have opinions.

### DIFF
--- a/core/math/a_star.cpp
+++ b/core/math/a_star.cpp
@@ -45,18 +45,6 @@ int64_t AStar3D::get_available_point_id() const {
 	return last_free_id;
 }
 
-void AStar3D::update_point(int64_t p_id, const Vector3 &p_pos, real_t p_weight_scale) {
-	ERR_FAIL_COND_MSG(p_id < 0, vformat("Can't update a point with negative id: %d.", p_id));
-	ERR_FAIL_COND_MSG(p_weight_scale < 0.0, vformat("Can't update a point with weight scale less than 0.0: %f.", p_weight_scale));
-
-	Point *found_pt;
-	bool p_exists = points.lookup(p_id, found_pt);
-	ERR_FAIL_COND_V_MSG(!p_exists, Vector3(), vformat("Can't update point's position. Point with id: %d doesn't exist.", p_id));
-
-	found_pt->pos = p_pos;
-	found_pt->weight_scale = p_weight_scale;
-}
-
 void AStar3D::add_point(int64_t p_id, const Vector3 &p_pos, real_t p_weight_scale) {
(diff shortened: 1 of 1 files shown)


AThousandShips (changes requested):
This will leak memory when adding a point when it already exists, and changes behavior as it also ignores data of existing points when adding them when already existing, the original `add_point` should not be changed IMO

Adding a check in `add_point` to fail when adding a point that already exists would break compatibility

Edit: Further, this breaks compatibilty, as the current behavior is documented and therefore part of the official API:
> If there already exists a point for the given id, its position and weight scale are updated to the given values.

The memory leak is unacceptable, but if the behaviour is changed the documentation has to be changed to match

Edit: this also breaks neighbours of existing points, so it breaks a lot of places

core/math/a_star.cpp
```diff
@@ -49,23 +49,27 @@ void AStar3D::add_point(int64_t p_id, const Vector3 &p_pos, real_t p_weight_scal
...
+	pt->pos = p_pos;
+	pt->weight_scale = p_weight_scale;
+	pt->prev_point = nullptr;
+	pt->open_pass = 0;
+	pt->closed_pass = 0;
+	pt->enabled = true;
+	points.set(p_id, pt);
+}
+
+void AStar3D::update_point(int64_t p_id, const Vector3 &p_pos, real_t p_weight_scale) {
```
AThousandShips, line 63: You haven't added these to the header

blooblahguy:
You're right. Besides the fact that my PR is _very_ incomplete, I can understand that the memory leak is unacceptable. But on a sidenote this shouldn't break neighbor connections, as it does exactly what it did before just as 2 different methods instead.
Either way - I'll close this out! Thanks

AThousandShips:
It will break neighbor connections, see:
https://github.com/godotengine/godot/blob/fa48a51183567934984b381ad8ec281cb24d66ba/core/math/a_star.h#L46-L63

As you can see the neighbors will now point to a dangling pointer 🙂

Do discuss this in a [proposal](https://github.com/godotengine/godot-proposals) if you want to add something like this, and than you for your contribution nonetheless!

PR #87696, opened 2024-01-29
Implement support for PCK signing.

bruvzg:
- Adds support for PCK signing. Works in a similar manner to the PCK encryption:
  - SHA256 file hashes added to the PCK directory structure.
  - "require validation" file flags added to the PCK directory structure (hashes of the files w/o flag are not validated to save loading time).
  - PCK directory is signed using provided keypair.
  - If enabled (executable compiled with the public key), it will refuse to load unsigned PCKs.
- Adds buttons to generate keys from the exporter UI.
- Improve PCK encryption and signing filters handling to automatically include imported files and remaps.

| | |
| -- | -- |
| Screenshot 2024-01-29 at 07 53 33 | Screenshot 2024-01-29 at 07 54 06 |

As with encryption, custom build is required to enable validation (using environment variables a build time):

Screenshot 2024-01-29 at 07 55 29


core/io/pck_packer.h
```diff
@@ -53,14 +53,23 @@ class PCKPacker : public RefCounted {
...
 		bool encrypted = false;
+		bool require_verification = false;
 		Vector<uint8_t> md5;
+		Vector<uint8_t> sha256;
 	};
 	Vector<File> files;
 
+protected:
+#ifndef DISABLE_DEPRECATED
+	Error _add_file_compat_87696(const String &p_file, const String &p_src, bool p_encrypt);
```
AThousandShips, line 64:
```suggestion
	Error _add_file_bind_compat_87696(const String &p_file, const String &p_src, bool p_encrypt);
```

core/io/pck_packer_compat.inc
AThousandShips, line 1: Should be named `pck_packer.compat.inc`

editor/export/editor_export_platform.h
```diff
@@ -53,7 +59,7 @@ class EditorExportPlatform : public RefCounted {
 	static void _bind_methods();
 
 public:
-	typedef Error (*EditorExportSaveFunction)(void *p_userdata, const String &p_path, const Vector<uint8_t> &p_data, int p_file, int p_total, const Vector<String> &p_enc_in_filters, const Vector<String> &p_enc_ex_filters, const Vector<uint8_t> &p_key);
+	typedef Error (*EditorExportSaveFunction)(void *p_userdata, const String &p_source_path, const String &p_path, const Vector<uint8_t> &p_data, int p_file, int p_total, const Vector<String> &p_enc_in_filters, const Vector<String> &p_enc_ex_filters, const Vector<String> &p_sign_in_filters, const Vector<String> &p_sign_ex_filters, const Vector<uint8_t> &p_key, bool p_is_signed);
```
akien-mga, line 62: Likewise for this, I'm thinking a public struct might be something each export plugin can reuse for this.

akien-mga (commented):
Looks like a pretty thorough implementation. I have yet to test it to see how the flow is like. This will require extensive user documentation so they know how to use it (we're currently also lacking in documentation for the script signing, we have a page about compiling custom templates but no real mention of it in the Exporting section of the docs).

I'd like @Faless to review the crypto related changes.

And overall this requires some discussion with maintainers and users to see how desirable this feature would be. It adds a bit of complexity, so we need to make sure it's something users would actually use.

There's significant demand for better obfuscation / protection of PCKs, so this seems to go in the right direction for this use case, but we need to make sure those users would be happy with this solution.

reduz:
This looks extremely good. In fact, I am not sure if we should also just deprecate the old encryption menu and keep signing with private key only.

MBCX:
Hey, I'd love to test this out. However, it seems that this branch is quite behind master, and the encryption mechanism has changed a lot since, making this branch harder to test out.

bruvzg:
> and the mechanism for embedding the encryption key has changed a lot since, making this branch harder to test out.

Key embedding mechanism did not change (there were few optimizations in the build scripts), but crypto API seems to be a bit different now. I have rebased it, but have not fully tested after it.

bruvzg:
> I have rebased it, but have not fully tested after it.

Tested key generation, export with signing and signature validation, seem to work fine.

MBCX:
Finally got the time to build and test this out, and as you said, it all works as expected. I do have a suggestion though.

It is explained that the binary compiled with the public key will refuse to load unsigned packs (seems to be intentional design). However, I don't think it should be *this* extreme, because now I'm forced to **always** sign the PCK, even if I want to share a demo of a game (for example).

This differs from the encryption. If my binary has the encryption key embed, but I decide to **not** encrypt the pck (continuing the game demo example), I can switch off the encryption and everything works fine. If I then later decide **to** encrypt (so, the final, release version of the game), I can switch on the encryption and everything still works. This mechanism should also apply for signing.

bruvzg:
> However, I don't think it should be this extreme

The point of signing is to prevent PCK modification (unlike encryption, it won't prevent PCK extraction in any form), so if you allow loading unsigned PCKs it's essentially 100% useless. If you want to modify PCK, you'll just strip the signature and do whatever you want.

MBCX:
Makes sense, thanks for explaining. I'll admit that my question comes from my lack of knowledge on this stuff, so my bad if it was a bit dumb 😅.

MBCX:
Welp, looks like 4.4 broke this feature again xs.

Looks like #90782 is the culprit. I tried rebasing on my own but then I get error when building.
```bash
Compiling editor/export/editor_export.cpp ...
BUILD: Elapsed time [00:11:20]
Compiling editor/export/editor_export_platform.cpp ...
BUILD: Elapsed time [00:11:20]
Compiling editor/export/editor_export_platform_extension.cpp ...
BUILD: Elapsed time [00:11:22]
Compiling editor/export/editor_export_platform_pc.cpp ...
editor/export/editor_export_platform.cpp:971:49: error: cannot initialize a parameter of type 'EditorExportSaveFunction' (aka 'Error (*)(void *, const EditorExportPlatform::ExportFileData &, const Vector<unsigned char> &)') with an lvalue of type 'Error (void *, const String &, const Vector<uint8_t> &, int, int, const Vector<String> &, const Vector<String> &, const Vector<uint8_t> &)' (aka 'Error (void *, const String &, const Vector<unsigned char> &, int, int, const Vector<String> &, const Vector<String> &, const Vector<unsigned char> &)'): different number of parameters (3 vs 8)
  971 |         return export_project_files(p_preset, p_debug, _script_save_file, &data, _script_add_shared_object);
      |                                                        ^~~~~~~~~~~~~~~~~
editor/export/editor_export_platform.h:305:109: note: passing argument to parameter 'p_func' here
  305 |         Error export_project_files(const Ref<EditorExportPreset> &p_preset, bool p_debug, EditorExportSaveFunction p_func, void *p_udata, EditorExportSaveSharedObject p_so_func = nullptr);
      |                                                                                                                    ^
BUILD: Elapsed time [00:11:24]
Compiling editor/export/editor_export_plugin.cpp ...
````

File: thirdparty/mbedtls/include/godot_core_ecdsa_mbedtls_config.h [hunk 3]
```diff
@@ -0,0 +1,75 @@
...
+
+#define MBEDTLS_AES_C
+#define MBEDTLS_BASE64_C
+#define MBEDTLS_CTR_DRBG_C
+#define MBEDTLS_ENTROPY_C
+#define MBEDTLS_MD5_C
+#define MBEDTLS_SHA1_C
+#define MBEDTLS_SHA256_C
+#define MBEDTLS_PLATFORM_ZEROIZE_ALT
+#define MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES
```
akien-mga, line 51: To avoid going out of sync with the base mbedtls config, maybe you could just include `godot_core_mbedtls_config.h`, and then add the extra ECP defines?

akien-mga: CC @Faless to review the extra defines too.

akien-mga:
I've had in mind to review this for a while - would be worth a try to include in 4.4 beta 1, so it could use a rebase.

core/crypto/crypto_core.cpp
```diff
@@ -32,6 +32,7 @@
 
 #include "core/os/os.h"
 
+#define MBEDTLS_ALLOW_PRIVATE_ACCESS
```
Faless, line 35: Why is this needed?

thirdparty/mbedtls/include/godot_core_ecdsa_mbedtls_config.h
Faless: I think we need to decide if we want to add those in core or not. Having yet another code path for core is too hard to maintain IMO.

And if we decide to add signing and ECDSA in core, we should evaluate if it's worth to allow disabling mbedtls at all (i.e. how much size do we save in having this VS the whole mbedtls module enabled?)

Faless:
I think this PR should be split in two (at least 2 commits), one adding ECDSA support (see my comment above about adding it to core as a separate option VS always adding it VS implementing it in the module), and the other implementing signature verification.

In fact, the signature part does not need to use ECDSA (i.e. it could use RSA which is already implemented in the module).

core/crypto/crypto_core.cpp
bruvzg: To have access to `mbedtls_ecp_keypair` struct members.

bruvzg:
> In fact, the signature part does not need to use ECDSA (i.e. it could use RSA which is already implemented in the module).

This will add core ↔ module dependency which is also not nice, but I guess it make sense to reuse existing RSA code, will split it into two PRs.

bruvzg:
> One thing I noticed, is that this PR seems to sign and verify every added file

It's only sign and verify PCK directory (list of files and file hashes), and check hashes for specific files (usually there's no reason to check heavy files like textures and models).

Faless:
> > 
> 
> It's only sign and verify PCK directory (list of files and file hashes), and check hashes for specific files (usually there's no reason to check heavy files like textures and models).

I'm not sure that's the best course of action tbh... it seems very error prone this way, and unlike encryption, hashing is meant to be fast and the heavy operations are only done over the final hash.

bruvzg:
> Actually it's not necessary to use module directly, it can use generic CryptoKey andCrypto directly for RSA.

Nop, no gonna work with module at all, PCK is loaded before loading modules.

bruvzg:
> I'm not sure that's the best course of action tbh... it seems very error prone this way, and unlike encryption, hashing is meant to be fast and the heavy operations are only done over the final hash.

I'll benchmark it, but I suspect hashing the whole pack won't be fast enough to be completely seamless (especially on older mobile platforms and with large packs).

bruvzg:
> I'll benchmark it, but I suspect hashing the whole pack won't be fast enough to be completely seamless (especially on older mobile platforms and with large packs).

Tested it on M4 Pro mac, it's about 200 MiB/sec for debug build and about 400 MiB/sec for optimized release build. Wikipedia also mentions 196 MiB/sec for old AMD CPUs, so I guess we can assume 200 MiB/sec as a reasonable assumption. With 1-2 GiB PCK it's 5-10 sec delay, so hashing the whole pack is not a reasonable option.

Faless:
> so hashing the whole pack is not a reasonable option.

I honestly disagree (keep in mind, on those old systems, even reading the PCK from disk will be way slower than the 200 MB/s), but maybe I'm missing the point of having signed/files pck? What is the end goal of the feature? And how is supposed to be used (i.e. how would the user achieve the goal)?

bruvzg:
> keep in mind, on those old systems, even reading the PCK from disk will be way slower than the 200 MB/s

It does not matter, PCK is not read fully at once, so unlike checking hashing it at load won't cause a big stutter (same for hashing individual parts of the PCK).

> What is the end goal of the feature?

Preventing modification of the critical parts of the exported PCK (scripts, for example). As with PCK encryption, it assumes some external measures are taken to prevent executable modification (DRM, platform which is validating executable signatures), otherwise it's pretty simple to workaround.

bruvzg:
Changed PR to use RSA from `mbedtls` module, without adding ECDSA code to the core (but as I mentioned before it adds extra init level to make sure `mbedtls` module is ready before main PCK is loaded).

Faless:
> It does not matter, PCK is not read fully at once

> Preventing modification of the critical parts of the exported PCK (scripts, for example).

Right, then it make sense to only sign the metadata (which includes file hashes), but we still need to **verify the hashes for all files** (when we load them I guess), or specific files could be tampered, and scripts could be added to them (almost all files, including resource types, can contain scripts).

We also don't need to add both md5 and sha256, we should probably just use sha256 when the PCK (metadata) is signed.

bruvzg:
> Right, then it make sense to only sign the metadata (which includes file hashes), but we still need to verify the hashes for all files (when we load them I guess), or specific files could be tampered, and scripts could be added to them (almost all files, including resource types, can contain scripts).

That's what this PR do, file hashes are verified (only SHA256 hash), in this case speed is less relevant since only a specific file is hashed when it is first accessed, so overhead will be spread out.

> We also don't need to add both md5 and sha256, we should probably just use sha256 when the PCK (metadata) is signed.

I guess we can remove MD5 hash, not sure why it was there in the first place, it's not used for anything.

bruvzg:
> I guess we can remove MD5 hash, not sure why it was there in the first place, it's not used for anything.

Actually, it seems to be exposed via `get_file_hash` and used by to handle patches, but it can be switched to SHA256 as well.

Faless:
> That's what this PR do, file hashes are verified (only SHA256 hash), in this case speed is less relevant since only a specific file is hashed when it is first accessed, so overhead will be spread out.

I was referring to:

> * "require validation" file flags added to the PCK directory structure (hashes of the files w/o flag are not validated to save loading time).

And indeed the code seems to only verify files with the given flag.

I understand one may not care about verifying e.g. textures (though again, the speed difference will be minimal), but this is, as I said, very error prone, since a lot of resource files can still contain scripts.

And a PCK signed and distributed with an improper flag (e.g. excluding a `.sample` audio file), will result in all binaries using that key being susceptible to load malicious PCKs (where a `.sample` file is tampered to include a script), which is quite in contrast with what one would expect from digital signatures

bruvzg:
> I understand one may not care about verifying e.g. textures (though again, the speed difference will be minimal), but this is, as I said, very error prone, since a lot of resource files can still contain scripts.

It can be error-prone, but it defaults to be applied to all files when signing is enabled, so it will be different only if filters are deliberately modified.

bruvzg:
Removed MD5 hashes.

Faless:
> It can be error-prone, but it defaults to be applied to all files when signing is enabled, so it will be different only if filters are deliberately modified.

My point is that if you really don't want to modify the filters ever, then there's no reason to have the filters in the first place.

The moment you allow them, people will use them, and that will result in allowing unsafe PCKs which defeats the purpose of the feature.

bruvzg:
> The moment you allow them, people will use them, and that will result in allowing unsafe PCKs which defeats the purpose of the feature.

I guess this make sense, and this is not as important as with extremely slow encryption, I have removed the filters.

PR #87701, opened 2024-01-29
Allow `Node.create_tween()` outside SceneTree

KoBeWi:
`Node.create_tween()` does not work when the node is outside tree. This was never an intended limitation, I just forgot that we have global access to SceneTree.

### DIFF
--- a/doc/classes/Node.xml
+++ b/doc/classes/Node.xml
@@ -205,7 +205,7 @@
 		<method name="create_tween">
 			<return type="Tween" />
 			<description>
-				Creates a new [Tween] and binds it to this node. Fails if the node is not inside the tree.
+				Creates a new [Tween] and binds it to this node.
 				This is the equivalent of doing:
 				[codeblocks]
 				[gdscript]
@@ -215,7 +215,8 @@
 				GetTree().CreateTween().BindNode(this);
 				[/csharp]
 				[/codeblocks]
-				The Tween will start automatically on the next process frame or physics frame (depending on [enum Tween.TweenProcessMode]).
+				The Tween will start automatically on the next process frame or physics frame (depending on [enum Tween.TweenProcessMode]). See [method Tween.bind_node] for more info on Tweens bound to nodes.
+				[b]Note:[/b] The method can still be used when the node is not inside [SceneTree]. It can fail in an unlikely case of using a custom [MainLoop].
 			</description>
 		</method>
 		<method name="duplicate" qualifiers="const">
--- a/scene/main/node.cpp
+++ b/scene/main/node.cpp
@@ -2326,8 +2326,14 @@ void Node::_propagate_replace_owner(Node *p_owner, Node *p_by_owner) {
 
 Ref<Tween> Node::create_tween() {
 	ERR_THREAD_GUARD_V(Ref<Tween>());
-	ERR_FAIL_NULL_V_MSG(data.tree, nullptr, "Can't create Tween when not inside scene tree.");
-	Ref<Tween> tween = get_tree()->create_tween();
+
+	SceneTree *tree = data.tree;
+	if (!tree) {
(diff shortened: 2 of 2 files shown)


JekSun97:
Don't forget about 3.x :)

scene/main/node.cpp
```diff
@@ -2326,8 +2326,14 @@ void Node::_propagate_replace_owner(Node *p_owner, Node *p_by_owner) {
...
 Ref<Tween> Node::create_tween() {
 	ERR_THREAD_GUARD_V(Ref<Tween>());
-	ERR_FAIL_NULL_V_MSG(data.tree, nullptr, "Can't create Tween when not inside scene tree.");
-	Ref<Tween> tween = get_tree()->create_tween();
+
+	SceneTree *tree = data.tree;
+	if (!tree) {
+		tree = SceneTree::get_singleton();
+	}
+	ERR_FAIL_NULL_V_MSG(tree, nullptr, "No available SceneTree to create the Tween.");
```
AThousandShips, line 2334:
```suggestion
	ERR_FAIL_NULL_V_MSG(tree, Ref<Tween>(), "No available SceneTree to create the Tween.");
```
For uniformity and clarity IMO

AThousandShips (approved):
LGTM otherwise, looked at the tween processing code and it doesn't depend on the node being in the tree

It's worth noting that the default tween mode will not allow it to process if it's not in the tree, the documentation isn't entirely clear on this

KoBeWi:
Right, I forgot to update the docs .-.

AThousandShips:
You could still achieve the same situation with the existing functionality, haven't tested but this should do it:
```gdscript
var my_tween = Engine.get_main_loop().create_tween()
my_tween.bind_node(self)
```
It's just much more convenient to get a tween now, but the caveat is that it won't process off the tree by default, but I suspect that's largely a desired behavior

So just a note about the pause mode `PAUSE_MODE_BIND` also requiring the node to be attached should be enough

KoBeWi:
Updated the docs.

Note that when you bind a node, the Tween will not process if the node is not inside scene tree, regardless of the pause mode. I linked the `bind_node()` method in the description, because it explains it in detail.

doc/classes/Node.xml
```diff
@@ -205,7 +205,7 @@
 		<method name="create_tween">
 			<return type="Tween" />
 			<description>
-				Creates a new [Tween] and binds it to this node. Fails if the node is not inside the tree.
+				Creates a new [Tween] and binds it to this node.
 				This is the equivalent of doing:
```
AThousandShips, line 209: It should be obvious, but might want to clarify here
```suggestion
				This is the equivalent of doing (if inside the tree):
```
Or similar

AThousandShips (approved):
Had missed the details in `bind_node`

LGTM again, just one possible improvement

doc/classes/Node.xml
KoBeWi: Not sure about that. If I add this, it would be worth to mention that the method also works outside the tree, adding unnecessary details. Previously the description just stated the method does not work outside scene tree, now it always works (unless you use custom MainLoop), so I just removed the note. No people complained about that limitation, so if someone assumes from the example that it's impossible to call the method outside the tree, it's not really a problem.

Mickeon:
I believe it is worth adding a tiny note about `create_tween` works outside of the tree. Two reasons:
- Returning users need some way of discovering this without religiously looking at patchnotes;
- The assumption is that there's only one **MainLoop** and that it is a **SceneTree**, and for 99% of projects that's true. But for the few projects with custom MainLoops, this Quality-of-Life handling may not work as _they_ intend.

Mickeon (approved):
👍

akien-mga:
Thanks!

PR #87706, opened 2024-01-29
Make script UIDs optional

KoBeWi:
This PR removes the UID code from GDScript's `save()` and instead the ScriptCreateDialog is now responsible for adding UIDs, which means that script no longer force adding UIDs.
image
The new checkbox is enabled by default and it's state is remembered per-project. The script is created as before and the UID is added afterwards by `ResourceSaver.set_uid()`.

The script editor got a new File menu option to add UID to existing files:
image
Deleting a UID will now unlink it from the script, so you'll need to re-add it manually or generate a new one. I also added a method to ScriptLanguage to check whether the language supports UIDs.

There are still some bugs, so putting as draft.

### DIFF
--- a/core/object/script_language.h
+++ b/core/object/script_language.h
@@ -253,6 +253,7 @@ public:
 #endif
 	virtual bool supports_builtin_mode() const = 0;
 	virtual bool supports_documentation() const { return false; }
+	virtual bool supports_uid() const { return false; }
 	virtual bool can_inherit_from_file() const { return false; }
 	virtual int find_function(const String &p_function, const String &p_code) const = 0;
 	virtual String make_function(const String &p_class, const String &p_name, const PackedStringArray &p_args) const = 0;
--- a/editor/plugins/script_editor_plugin.cpp
+++ b/editor/plugins/script_editor_plugin.cpp
@@ -1429,6 +1429,30 @@ void ScriptEditor::_menu_option(int p_option) {
 				es->run();
 			} break;
 
+			case FILE_INSERT_UID: {
+				Ref<Script> scr = current->get_edited_resource();
+				if (scr.is_null()) {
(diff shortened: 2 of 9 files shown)


arkology:
> The script editor got a new File menu option to add UID to existing files

Also it would be nice to have ability generating uids for all scripts in project. Last time I tested, uid was generated only when saving with existing changes (saving file without making any changes did not generate uid), so algorithm was "open every script, make changes, save script"

KoBeWi:
> Also it would be nice to have ability generating uids for all scripts in project.

I'm not sure about adding this option. It's only useful for old projects and you need to use it once. It can be solved with a simple editor script:
```GDScript
@tool
extends EditorScript

func _run() -> void:
	scan("res://")

func scan(dir: String):
	for file in DirAccess.get_files_at(dir):
		if file.get_extension() != "gd":
			continue
		
		var path := dir.path_join(file)
		if ResourceLoader.get_resource_uid(path) != ResourceUID.INVALID_ID:
			continue
		
		ResourceSaver.set_uid(path, ResourceUID.create_id())
	
	for dir2 in DirAccess.get_directories_at(dir):
		if dir2.begins_with("."):
			continue
		
		scan(dir.path_join(dir2))
```
...except `set_uid()` is not exposed. I'll fix that.

KoBeWi:
Updated. I fixed some weird error with modified times and added proper warnings when inserting UID. Also inserting UID if a script already has one is not allowed.

There is a weird bug where inserting UID twice will result in 2 UIDs generated; seems like `set_uid()` does not properly register it? Not sure how to do that.

Also UIDs were reverted, so this branch can't be rebased until they are re-added .-.

arkology:
> It can be solved with a simple editor script
> except set_uid() is not exposed. I'll fix that.

I think its enough, thanks

KoBeWi:
#97352 adds alternate script UID implementation that makes this unnecessary.

PR #74405, opened 2023-03-05
Fix multi-threaded resource loading

RandomShaper:
I've reworked the resource loader so all the kinds of loads (current thread, another thread, multiple threads) cooperate better among them regardless which are in flight. This should get rid of some or all of the issues that users are encountering in this area. ~(I've come across some crashes due to some of the resources being loaded being of classes which are not thread-safe, but those are separate issues.)~ **UPDATE 3: Fixed as well.**

**The exposed API is the same**, but the inner workings of the loaders are changed so that they have more common code paths.

**Progress reporting also works** now! It has room for improvement, but at least it goes from 0 to 1 with growing intermediate values.

~TODO: Progress reporting is still not done.~

There are plans to maybe re-rewrite this based on a worker thread pool, but this should serve as a better base than the former implementation.

Supersedes #74120.
Supersedes #74108.

Fixes #56882.
May fix #75815.
May fix #76928.
May fix #75841.

Now resource loading passes thread sanitization tests from #73825 with only a few minor data race issues that are not a real issue in practice (see https://github.com/RandomShaper/godot/actions/runs/4337516127/jobs/7573570235).

**UPDATE 1:**
Cleanup of load tasks is pretty robust now (or so I think). ~When closing the engine, a cascade of cancels happens so tasks don't leak resources. The slight issue with that is that resource loaders print errors about non-satisfied dependencies, but that could be made silent if a big deal.~

**UPDATE 2:**
~This is synergistic with #75937.~ (Already merged.)

**UPDATE 3:**
Added one commit which also avoids the issues (i.e., crashes) that certain resource types can cause. Specifically, this is for any resource type that calls some sort of update/refresh on itself upon a property change, if being loaded from a non-main thread, can have such a deferred call made in the main thread while the resource is still not ready!

**UPDATE 4:**
~Now includes #76754 and #76755 because it's better to test them all together.~ (Already merged.)

**UPDATE 5:**
- Fixed a few general issues.
- Added a new commit about a very specific kind of issue, similar to the one described in _update 3_, but corresponding to materials and their auto-generated shaders. These use a different mechanism than the message queue and, therefore, need a different fix.

  This can be seen as a follow-up of #52466. The fix done in that PR is good, but there are still race conditions if the material has its properties changed in a thread beyond the constructor. That was an issue for loading materials in threads. Therefore this commit adds a fix that leverages the mechanism introduced in _update 3_, to have even those cases fixed, but only in the scope of resource loading.

**UPDATE 6 (wow!):**
[...]

PR #87711, opened 2024-01-29
Fix `ResourceLoader.load_threaded_get_status` returning `[0]` constantly in exported projects.

### LINKED ISSUE #56882: ResourceLoader load_threaded_get_status  progress: Array = [] is always [0]
___

***Bugsquad note:** This issue has been confirmed several times already. No need to confirm it further.*

___

### Godot version

4.0 dev

### System information

Linux Ubuntu 20.04.3 LTS 

### Issue description

From Godot Documentation:
● ThreadLoadStatus load_threaded_get_status(path: String, progress: Array = [])

Returns the status of a threaded loading operation started with load_threaded_request() for the resource at path. See ThreadLoadStatus for possible return values.

An array variable can optionally be passed via progress, and will return a one-element array containing the percentage of completion of the threaded loading.

But progress value is always  [0]

### Steps to reproduce

It is simple to reproduce, run this Project: https://github.com/VP-GAMES/Godot4InteractiveSceneChanger

Add print(_progress) in func  _update_progress() [Line 53] in this script: https://github.com/VP-GAMES/Godot4InteractiveSceneChanger/blob/main/addons/interactive_scene_changer/InteractiveSceneChanger.gd

Output
[0]
[0]
[0]
[0]
.
.
n

_No response_

matthew1006:
This PR fixes #56882 (which wasn't completely fixed by #74405) by pushing `p_original_path` instead of `p_path` into `load_paths_stack` and `sub_tasks`.

I believe the issue is that `thread_load_tasks` always uses the original path of to the resource in `res://` but in exported projects `p_path` is the path to the imported file in the `.godot` folder so `load_paths_stack` doesn't store the correct path.
This causes the lookup at [line 245](https://github.com/godotengine/godot/blob/master/core/io/resource_loader.cpp#L245) to fail, which causes `sub_tasks` to remain empty, which causes `dep_count` at [line 512](https://github.com/godotengine/godot/blob/master/core/io/resource_loader.cpp#L512) to be 0, causing `_dependency_get_progress` to return 0, etc.

Using `p_original_path` fixes this issue.

### DIFF
--- a/core/io/resource_loader.cpp
+++ b/core/io/resource_loader.cpp
@@ -244,11 +244,11 @@ Ref<Resource> ResourceLoader::_load(const String &p_path, const String &p_origin
 		thread_load_mutex.lock();
 		HashMap<String, ThreadLoadTask>::Iterator E = thread_load_tasks.find(load_paths_stack->get(load_paths_stack->size() - 1));
 		if (E) {
-			E->value.sub_tasks.insert(p_path);
+			E->value.sub_tasks.insert(p_original_path);
 		}
 		thread_load_mutex.unlock();
 	}
-	load_paths_stack->push_back(p_path);
+	load_paths_stack->push_back(p_original_path);
 
 	// Try all loaders and pick the first match for the type hint
 	bool found = false;


RandomShaper (approved):
Sadly, I don't have time to test this, but the rationale and the changes look good to me. So, I'm leaving my approval, but I'd need help from either the OP of the issues or the Prod team as extra testers of this.

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

akien-mga:
> Cherry-picked for 4.2.2.

Reverted for 4.2.2 due to #90076.

#90091 may fix the regression in `master`, but it still needs review.

PR #87713, opened 2024-01-29
[3.x] Portals - include in bound and special cases in start room

### LINKED ISSUE #87685: Godot 3 portals that are parallel but not coplanar create a gap in room hulls
### Tested versions

Reproducible in godot 3.5.3, godot 3.5.2. Godot 4.2.1 does not have the portal system and is unaffected.

### System information

Linux - Godot 3.5.3 - GLES3 - Mesa Intel(R) Graphics (ADL GT2)

### Issue description

If you are editing complicated room geometry and you end up with 2 planes that are parallel but not coplanar they may cause a gap between the convex hull of their common room.

image

Ideally this would create an overlap between `Hall` and `Room3` rather than creating a gap between `Hall2` and `Room3`.

If that's not possible then a warning would be nice.

#### Workaround

To work around this issue ensure all parallel portals that connect to the same face of a room are also coplanar, or rotate the portals to allow creation of a convex hull.

### Steps to reproduce

Create a scene like the  MRP (or download the MRP)
Open the main scene and click 'Convert all rooms' for portal culling. Notice that Room3 and Hall2 have a gap, which causes everything to disappear when the camera is in the gap between rooms after exiting the portal in Hall2

[portal_room_offset_godot3.zip](https://github.com/godotengine/godot/files/14077488/portal_room_offset_godot3.zip)

lawnjelly:
* Re-introduces a property for portals to decide whether they are included in room bounds during room conversion.
* Adds a special case for portals that extend into the start room, which may be caused by level design inaccuracies.

Fixes #87685

## Explanation
The special case deals with the case where the user has placed a portal inside a room but not coplanar to the bound. This previously had the effect of shrinking the bound, leading to a "no man's land" where it wasn't sure what to render, and you could get drop outs where objects were incorrectly culled as the camera moved between two rooms (see #87685  for example).

In order to get around this problem firstly I re-added the ability to exclude a portal from the room bound. I must have removed this ability inadvertently in the initial PR and not realised. This enables the rooms to not be "shrunk" by portals that are slightly misplaced.

The second special case deals with the problems where a portal in the start room is not co-planar to the bound _exactly_. Previously, providing the portal was pretty close, you were unlikely to see drop outs. These are caused when the camera moves _in front_ of the portal, but still _within_ the start room. Previously when the camera was in front of a portal, the logic determined it could not see through it.

The solution in this PR was to simply disable rejection of portals in the start room if you move in front of them. Theoretically, this should not be possible anyway with a perfectly built room. If you do move in front, it is likely you would expect to be able to see through the portal anyway, so the tracing traverses directly into the linked room, and does not create any extra culling planes for the portal (these are undefined in this "fudge" scenario).

## Notes
* A lot of the changes here are just comments (in the end I fixed up the capitalization of existing comments in the trace function, so they were consistent with the new standard).
* The special case treatment for the start room is disabled for internal portals, in order to work well with internal rooms.
* I've tested this as best I can with regular rooms and internal rooms. It should be more robust than before.
* The case where in front of a portal plane in the start room when the camera is "to the side" of the portal and thus shouldn't be able to really see through it is sidestepped for now. This could be made more efficient (in terms of less rendered), but if anything it will currently over-render in this rare situation - i.e. there should be no falsely occluded objects and the viewer shouldn't see any artifacts.

### DIFF
--- a/doc/classes/Portal.xml
+++ b/doc/classes/Portal.xml
@@ -23,6 +23,9 @@
 		</method>
 	</methods>
 	<members>
+		<member name="include_in_bound" type="bool" setter="set_include_in_bound" getter="get_include_in_bound" default="false">
+			When a manual bound has not been explicitly specified for a [Room], the convex hull bound will be estimated from the geometry of the objects within the room. This setting determines whether the portal geometry is included in this estimate of the room bound.
+		</member>
 		<member name="linked_room" type="NodePath" setter="set_linked_room" getter="get_linked_room" default="NodePath(&quot;&quot;)">
 			This is a shortcut for setting the linked [Room] in the name of the [Portal] (the name is used during conversion).
 		</member>
--- a/scene/3d/portal.cpp
+++ b/scene/3d/portal.cpp
@@ -46,6 +46,7 @@ Portal::Portal() {
 
 	_settings_active = true;
 	_settings_two_way = true;
+	_settings_include_in_bound = false;
 	_internal = false;
 	_linkedroom_ID[0] = -1;
 	_linkedroom_ID[1] = -1;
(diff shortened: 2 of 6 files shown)


Calinou (approved):
Not tested on my end as I don't use portals in 3.x, but the code looks good to me.

akien-mga:
Thanks!

PR #87714, opened 2024-01-29
Expose `set_uid()`

### LINKED ISSUE #11971: Expose `ResourceSaver.set_uid()` to scripting.

EditorPlugin which handles all kinds of files in the EditorFileSystem, and possibly outside of it.

Specifically It has a function to copy arbitary files, regardless of their types, be it Resources or not.

It uses `DirAccess.copy_absolute()` to copy the files.

There seems to be no way to manually set UID to a resource (at least without loading and resaving it, which is probably more intensive anyways).

For example, if I copy a **Resource** file using:

```gdscript
DirAccess.copy_absolute(input_file, output_file)
```
And then ALT+TAB out and back to the editor, I'll get this warning, and it will update the resource file to have a new UID.
```
  WARNING: editor/editor_file_system.cpp:911 - Duplicate UID detected for Resource at "res://test_copy.tscn".
  WARNING: Old Resource path: "res://test.tscn". The new file UID was changed automatically.
```

If I instead manually update the editor file system::
```gdscript
DirAccess.copy_absolute(input_file, output_file)
EditorInterface.get_resource_filesystem().update_file(output_file) # or scan() etc.
```
I do not get the warning after I ALT+TAB out and back in, however I get this warning after reloading the current project, as the resource file still has old UID:
```
  WARNING: editor/editor_file_system.cpp:1352 - UID duplicate detected between res://test_copy.tscn and res://test.tscn.
```

So there is a mechanic to update the UID in some cases (but not all, related https://github.com/godotengine/godot/issues/102490), but they also result in warnings to the console, what I would like to avoid seeing.

Expose `ResourceSaver.set_uid()` to scripting.

It would allow for a easy way to update a resources UID without loading and resaving a resource. 

Although I am not familiar with the inner workings of `set_uid()` (as each `ResourceFormatSaver` does something different), but it might be slower to load and save a resource just to get a new UID than if using `ResourceSaver.set_uid()` (especially if the resource is massive).

The above proposed  warning is generated in `editor_file_system.cpp` at method `_update_scan_actions()`, and it internally uses `ResourceSaver.set_uid()`

```c++
const String new_file_path = ia.dir->get_file_path(idx);
const ResourceUID::ID existing_id = ResourceLoader::get_resource_uid(new_file_path);
if (existing_id != ResourceUID::INVALID_ID) {
    const String old_path = ResourceUID::get_singleton()->get_id_path(existing_id);
    if (old_path != new_file_path && FileAccess::exists(old_path)) {
        const ResourceUID::ID new_id = ResourceUID::get_singleton()->create_id();
        ResourceUID::get_singleton()->add_id(new_id, new_file_path);
        ResourceSaver::set_uid(new_file_path, new_id);
        WARN_PRINT(vformat("Duplicate UID detected for Resource at \"%s\".\nOld Resource path: \"%s\". The new file UID was changed automatically.", new_file_path, old_path));
    } else {
        // Re-assign the UID to file, just in case it was pulled from cache.
        ResourceSaver::set_uid(new_file_path, existing_id);
    }
}
```
https://github.com/godotengine/godot/blob/49cc57a75dfd262bd49698910595becf3aae597b/editor/editor_file_system.cpp#L903-916

If the function was exposed, it would allow setting the UID similiarly:

```gdscript
DirAccess.copy_absolute(input_file, output_file)
if ResourceLoader.exists(output_file):
    var new_uid = ResourceUID.create_id()
    ResourceUID.add_id(new_uid, output_file) # if necessary
    ResourceLoader.set_uid(new_uid, output_file)
```

It is possible to load a resource and then resave it immediately:
```gdscript
var res = ResourceLoader.load(input_file)
out.error = ResourceSaver.save(res, output_file)
```
Which results in a file with a unique UID, but might be slow on big files, and also introduces small amount of complexity to my use case, where I also want to handle non Resource files.

Editor functionality.

KoBeWi:
Exposes `set_uid()` method. Given the exposed methods in ResourceUID singleton, exposing this method too is only logical and has some valid use-cases (like https://github.com/godotengine/godot/pull/87706#issuecomment-1915441490).

Closes https://github.com/godotengine/godot-proposals/issues/11971

### DIFF
--- a/doc/classes/ResourceSaver.xml
+++ b/doc/classes/ResourceSaver.xml
@@ -53,6 +53,15 @@
 				[b]Note:[/b] When the project is running, any generated UID associated with the resource will not be saved as the required code is only executed in editor mode.
 			</description>
 		</method>
+		<method name="set_uid">
+			<return type="int" enum="Error" />
+			<param index="0" name="resource" type="String" />
+			<param index="1" name="uid" type="int" />
+			<description>
+				Sets the UID of the given [param resource] path to [param uid]. You can generate a new UID using [method ResourceUID.create_id].
+				Since resources will normally get a UID automatically, this method is only useful in very specific cases.
+			</description>
+		</method>
 	</methods>
 	<constants>
 		<constant name="FLAG_NONE" value="0" enum="SaverFlags" is_bitfield="true">
--- a/core/core_bind.cpp
+++ b/core/core_bind.cpp
@@ -169,6 +169,10 @@ Error ResourceSaver::save(const Ref<Resource> &p_resource, const String &p_path,
 	return ::ResourceSaver::save(p_resource, p_path, p_flags);
 }
 
+Error ResourceSaver::set_uid(const String &p_path, ResourceUID::ID p_uid) {
+	return ::ResourceSaver::set_uid(p_path, p_uid);
+}
+
 Vector<String> ResourceSaver::get_recognized_extensions(const Ref<Resource> &p_resource) {
 	List<String> exts;
 	::ResourceSaver::get_recognized_extensions(p_resource, &exts);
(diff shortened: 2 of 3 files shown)


doc/classes/ResourceSaver.xml
```diff
@@ -45,6 +45,15 @@
 				[b]Note:[/b] When the project is running, any generated UID associated with the resource will not be saved as the required code is only executed in editor mode.
 			</description>
 		</method>
+		<method name="set_uid">
+			<return type="int" enum="Error" />
+			<param index="0" name="resource" type="String" />
+			<param index="1" name="uid" type="int" />
+			<description>
+				Sets the UID of the [param resource] to [param uid]. You can generate a new UID using [ResourceUID.create_id].
+				Since resources will normally get a UID automatically, this method is useful in very specific cases.
```
Mickeon, line 54: Minor nitpicks
```suggestion
				Sets the UID of the given [param resource] path to [param uid]. You can generate a new UID using [ResourceUID.create_id].
				Since resources will normally get a UID automatically, this method is only useful in very specific cases.
```

Mickeon (approved):
Docs are just fine

fire:
See also https://github.com/godotengine/godot/pull/88111

Repiteo (approved):

Repiteo:
Needs a rebase; just to ensure that the jobs still run successfully

Repiteo:
Thanks!

PR #87721, opened 2024-01-29
Enforce calling RenderingDevice code from rendering thread in TextureRD classes

clayjohn:
Needed for https://github.com/godotengine/godot/pull/87590

This fixes an unreported bug in the TextureRD classes where RenderingDevice functions could get called from any thread. Currently we allow doing this as we guard every RD function that cares about threads behind a mutex. However in https://github.com/godotengine/godot/pull/87590 we are moving to a much more efficient approach which is to have thread guards on single-threaded functions and to have nothing on the functions that can be called from any thread. 

Accordingly, certain functions will need to be called from the rendering thread using `call_on_render_thread()`. The overall impact of this change is minimal.

### DIFF
--- a/scene/resources/texture_rd.cpp
+++ b/scene/resources/texture_rd.cpp
@@ -74,27 +74,7 @@ void Texture2DRD::set_texture_rd_rid(RID p_texture_rd_rid) {
 	ERR_FAIL_NULL(RS::get_singleton());
 
 	if (p_texture_rd_rid.is_valid()) {
-		ERR_FAIL_NULL(RD::get_singleton());
-		ERR_FAIL_COND(!RD::get_singleton()->texture_is_valid(p_texture_rd_rid));
-
-		RD::TextureFormat tf = RD::get_singleton()->texture_get_format(p_texture_rd_rid);
-		ERR_FAIL_COND(tf.texture_type != RD::TEXTURE_TYPE_2D);
-		ERR_FAIL_COND(tf.depth > 1);
-		ERR_FAIL_COND(tf.array_layers > 1);
-
-		size.width = tf.width;
-		size.height = tf.height;
-
-		texture_rd_rid = p_texture_rd_rid;
-
-		if (texture_rid.is_valid()) {
-			RS::get_singleton()->texture_replace(texture_rid, RS::get_singleton()->texture_rd_create(p_texture_rd_rid));
-		} else {
-			texture_rid = RS::get_singleton()->texture_rd_create(p_texture_rd_rid);
-		}
-
-		notify_property_list_changed();
-		emit_changed();
+		RS::get_singleton()->call_on_render_thread(callable_mp(this, &Texture2DRD::_set_texture_rd_rid).bind(p_texture_rd_rid));
 	} else if (texture_rid.is_valid()) {
 		RS::get_singleton()->free(texture_rid);
 		texture_rid = RID();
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
Good catch. I would change this slightly because this means the property doesn't change at all until the rendering thread can process this. Which means that I can call `set_texture_rd_rid`, but still get the old rid afterwards when I call `get_texture_rd_rid`.

The same with the texture object, it will now be created late, while we likely need it earlier when it is being assigned to materials etc. This is why `texture_rd_create` creates the RID regardless of the thread, even if it doesn't do anything else.

So I think we should assign the `texture_rd_rid` variable, and call `texture_rd_create` if `!texture.is_valid` (and probably the emit changes thing) on the main thread, and then do the rest on the rendering thread. This could mean that we're assigning an invalid rid but in that case the texture will just not be fully setup.

clayjohn:
> Good catch. I would change this slightly because this means the property doesn't change at all until the rendering thread can process this. Which means that I can call `set_texture_rd_rid`, but still get the old rid afterwards when I call `get_texture_rd_rid`.
> 
> The same with the texture object, it will now be created late, while we likely need it earlier when it is being assigned to materials etc. This is why `texture_rd_create` creates the RID regardless of the thread, even if it doesn't do anything else.

We can't call `texture_rd_create()` if `texture_rd_rid` is invalid. And we can't check the validity of `texture_rd_rid` on the main thread. 

That being said, I'm not too worried about creating the texture object. We create a placeholder if the user calls `get_rid()` before this code runs.  

This has made me realize that ``texture_rd_create()`` is actually a bit of a problem. It is defined using `FUNCRIDTEX2`:

https://github.com/godotengine/godot/blob/4577dfdb6747600438ac4b86c29a4d6c12ec9b0a/servers/rendering/rendering_server_default.h#L159-L168

Which will create the texture on whatever thread it is called from. But it _has_ to be called from the rendering thread because it immediately calls ``texture_is_shared()`` which has to be called from the rendering thread. 

https://github.com/godotengine/godot/blob/4577dfdb6747600438ac4b86c29a4d6c12ec9b0a/servers/rendering/renderer_rd/storage_rd/texture_storage.cpp#L1469-L1474

> So I think we should assign the `texture_rd_rid` variable, and call `texture_rd_create` if `!texture.is_valid` (and probably the emit changes thing) on the main thread, and then do the rest on the rendering thread. This could mean that we're assigning an invalid rid but in that case the texture will just not be fully setup.

So I think the correct solution here is going to be a mix of things

Let's assign `texture_rd_rid` right away and then call `texture_rd_create` on the rendering thread. We can call `notify_property_list_changed` right away, but we should wait to emit changed just in case `texture_rid` isn't assigned yet. 

Then, we need to also stop using `FUNCRIDTEX2` for texture_rd_create to ensure that it gets called from the rendering thread.

BastiaanOlij:
@clayjohn it only calls `texture_allocate` on the thread its called from. It will call `texture_rd_initialize` on the render thread, thats what the
```
if (Thread::get_caller_id() == server_thread || RSG::texture_storage->can_create_resources_async()) {    \ 
	RSG::texture_storage->m_type##_initialize(ret, p1, p2);                                              \ 
} else {                                                                                                 \ 
	command_queue.push(RSG::texture_storage, &RendererTextureStorage::m_type##_initialize, ret, p1, p2); \ 
}       
```
does?

BastiaanOlij:
Or is it the `can_create_resources_async` that causes issues when that returns true?

clayjohn:
> Or is it the `can_create_resources_async` that causes issues when that returns true?

Ya, in the RD renderers `can_create_resources_async` is true so the function gets called immediately from the calling thread. That's why we need a different macro for it

RandomShaper (approved):

akien-mga:
Thanks!

PR #87722, opened 2024-01-29
Restore previous script open behavior for EditorPlugins

### LINKED ISSUE #87383: Scene Dock script button unnecessarily restrictive and breaks backward compatibility
### Tested versions

4.3.dev2

### System information

Windows 11

### Issue description

In Godot 4.2 and earlier, we relied on the `EditorPlugin::_edit` function call to open our Visual Scripting script attached to nodes within the scene in our GDExtension Plug-In.

In Godot 4.3, the change https://github.com/godotengine/godot/pull/84284 broke this functionality.  The change unnecessarily restricts implementations to having to register a `ScriptEditorBase` implementation; however, for GDExtensions, there is currently not an exposed way to contribute these, and so this breaks our user's ability to interact with the SceneTree Dock and to quickly jump from a node to a visual script that is attached.

### Steps to reproduce

1. Use Godot 4.3.dev2
2. Install Orchestrator Plugin https://github.com/Vahera/godot-orchestrator/releases (2.0.dev1)
3. Create an orchestration script, attaching it to a scene node in any scene.
4. Attempt to click on the script button next to the scene node, plugin does not react and show the graph window.

N/A

Naros:
Fixes #87383

### DIFF
--- a/editor/scene_tree_dock.cpp
+++ b/editor/scene_tree_dock.cpp
@@ -1578,6 +1578,15 @@ void SceneTreeDock::_load_request(const String &p_path) {
 }
 
 void SceneTreeDock::_script_open_request(const Ref<Script> &p_script) {
+	int plugin_count = EditorNode::get_singleton()->get_editor_data().get_editor_plugin_count();
+	for (int i = 0; i < plugin_count; i++) {
+		EditorPlugin *plugin = EditorNode::get_singleton()->get_editor_data().get_editor_plugin(i);
+		ScriptEditorPlugin *script_editor_plugin = Object::cast_to<ScriptEditorPlugin>(plugin);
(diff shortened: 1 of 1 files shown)


KoBeWi:
This looks like a hack to me. And tbh the original PR was also kind of a hack and spawned more regressions (#88061). Maybe a proper fix would be some way to edit a resource without opening it in the inspector.

Naros:
I agree @KoBeWi, but I also think that would likely be a much more intrusive change, and at that point I'd argue maybe the best course would be to simply revert the original change and restore the old behavior.  Then consider reimplementing the fix targeting 4.4+ at this point so that we can fully validate a working solution without regressions.

akien-mga:
Superseded by #88288. Thanks for the contribution!

PR #87723, opened 2024-01-30
Allow `SO_REUSEADDR` on Windows UDP sockets.

Malcolmnixon:
UPDATED: This pull request allows SO_REUSEADDR to be used on Windows UDP sockets. This keeps the TCP protection in place; but allows multiple UDPServer instances to listen for packets on the same port. This should make Windows and Linux behave identical with regards to UDPServer instances.

The following video demonstrates a MoCap application sending UDP skeleton packets to both a Godot application, and the Godot editor tool-script.
https://github.com/godotengine/godot/assets/1863707/7a0b6b92-65f3-41c1-b7df-054a9d384c1b

NOTE: Due to the MoCap packets being unicast, about 50% were going to the Godot editor, and 50% to the Godot application.

### DIFF
--- a/drivers/unix/net_socket_posix.cpp
+++ b/drivers/unix/net_socket_posix.cpp
@@ -714,12 +714,16 @@ void NetSocketPosix::set_reuse_address_enabled(bool p_enabled) {
 
 // On Windows, enabling SO_REUSEADDR actually would also enable reuse port, very bad on TCP. Denying...
 // Windows does not have this option, SO_REUSEADDR in this magical world means SO_REUSEPORT
-#ifndef WINDOWS_ENABLED
+#ifdef WINDOWS_ENABLED
+	if (_is_stream) {
+		return;
+	}
+#endif
+
 	int par = p_enabled ? 1 : 0;
 	if (setsockopt(_sock, SOL_SOCKET, SO_REUSEADDR, SOCK_CBUF(&par), sizeof(int)) < 0) {
 		WARN_PRINT("Unable to set socket REUSEADDR option!");
 	}
-#endif
 }
 
 void NetSocketPosix::set_reuse_port_enabled(bool p_enabled) {


Faless:
> This pull request allows SO_REUSEADDR to be used on Windows UDP sockets. This keeps the TCP protection in place; but allows multiple UDPServer instances to receive the same packets. This should make Windows and Linux behave identical with regards to UDPServer instances.

`SO_REUSEADDR` doesn't really work that way on *nix system.

It does not allow routing of a regular single packet to multiple bound sockets.

You can test this with the following script:

```
extends Control

var udp_server1 := UDPServer.new()
var udp_server2 := UDPServer.new()

var udp_send := PacketPeerUDP.new()

func _ready():
	print(udp_server1.listen(4343))
	print(udp_server2.listen(4343))

	udp_send.connect_to_host("127.0.0.1", 4343)
	udp_send.put_packet(PackedByteArray([1,2,3,4]))

	udp_server1.poll()
	udp_server2.poll()
	print("Server 1 ", udp_server1.is_connection_available())
	print("Server 2 ", udp_server2.is_connection_available())
```

Will print:
```
0
0
Server 1 false
Server 2 true
```

As you can see, the packet was routed to the last bound socket, not all of them.

As far as I understand this is also the behavior on Windows **without** `SO_REUSEADDR` which instead has a completely different meaning on Windows.

Faless:
> My prior experience with this has been combining SO_REUSEADDR with broadcast packets (network discovery) rather than unicast, and in those cases all applications bound using SO_REUSEADDR receive the same broadcast UDP packet.

Indeed, as far as I know that is the behavior on Linux for broadcast and multicast packets, but not for unicast packets

Malcolmnixon:
All right, this PR does **NOT** do what I had intended, but still may be useful - indeed I had incorrectly applied my experience with broadcast packets inappropriately to unicast packets, and numerous documentation/discussions make it clear that the sockets API cannot be used to implement multiple applications receiving unicast packets on the same computer.

Instead this change allows Windows to support multiple UDPServers listening to the same broadcast packets just as is supported on Linux today.

Consider the following minor modification to the previous code from @Faless:
```gdscript
extends Control

var udp_server1 := UDPServer.new()
var udp_server2 := UDPServer.new()

var udp_send := PacketPeerUDP.new()

func _ready():
	print("Operating System: ", OS.get_distribution_name())
	print("Server 1 listen ", udp_server1.listen(4343))
	print("Server 2 listen ", udp_server2.listen(4343))

	udp_send.set_broadcast_enabled(true)
	udp_send.set_dest_address("255.255.255.255", 4343)
	udp_send.put_packet(PackedByteArray([1,2,3,4]))

	udp_server1.poll()
	udp_server2.poll()
	print("Server 1 ", udp_server1.is_connection_available())
	print("Server 2 ", udp_server2.is_connection_available())
```

With Godot 4.2.1 on Linux we get both UDPServer instances receiving the same broadcast packet:
```
Operating System: Ubuntu 23.10
Server 1 listen 0
Server 2 listen 0
Server 1 true
Server 2 true
```

With Godot 4.2.1 on Windows we get a failure to start listening on the second UDPServer because the port is busy:
```
Operating System: Windows
Server 1 listen 0
Server 2 listen 2
Server 1 true
Server 2 false
```

With this patch on Windows we now get both UDPServer starting correctly and receiving the same broadcast packet:
```
Operating System: Windows
Server 1 listen 0
Server 2 listen 0
Server 1 true
Server 2 true
```

It doesn't solve the original problem I was trying to solve; however it does appear to meet the updated PR description (thanks @akien-mga) and normalizes some network behavior bringing Windows behavior in line with Linux for broadcast packets.

PR #87731, opened 2024-01-30
Use more reliable method to `get_window_at_screen_position` in Windows

0x0ACB:
`WindowFromPoint` is not reliable when using transparent windows. `ChildWindowFromPoint(HWND_DESKTOP, p)` does practically the same as `WindowFromPoint` with the only difference that it will not ignore transparent windows.

I had issues with physics picking because `gui.mouse_in_viewport` was false in `Viewport::_process_picking` since `WindowFromPoint` would return the Window below mine which was not a godot window.


Zireael07:
Why close?

0x0ACB:
Ill reopen in a bit. But my branch kinda got corrupted

PR #87738, opened 2024-01-30
mbedtls: Update to upstream version 2.28.7

akien-mga:
https://github.com/Mbed-TLS/mbedtls/releases/tag/v2.28.7

### DIFF
--- a/thirdparty/README.md
+++ b/thirdparty/README.md
@@ -519,7 +519,7 @@ in the MSVC debugger.
 ## mbedtls
 
 - Upstream: https://github.com/Mbed-TLS/mbedtls
-- Version: 2.28.5 (47e8cc9db2e469d902b0e3093ae9e482c3d87188, 2023)
+- Version: 2.28.7 (555f84735aecdbd76a566cf087ec8425dfb0c8ab, 2024)
 - License: Apache 2.0
 
 File extracted from upstream release tarball:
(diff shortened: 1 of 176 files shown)


thirdparty/README.md
```diff
@@ -528,7 +528,7 @@ File extracted from upstream release tarball:
   except `config_psa.h` and `psa_util.h`
 - All `.c` and `.h` from `library/` to `thirdparty/mbedtls/library/` except
   those starting with `psa_*`
-- The `LICENSE` file
+- The `LICENSE` file (edited to keep only the Apache 2.0 variant)
```
akien-mga, line 531: I think mbedTLS used to be dual licensed Apache-2.0 OR GPL-2.0, then they removed the GPL-2.0 which was just forcing them to generate two sets of source tarballs for each license, and now they re-added the dual licensing...? Not sure what GPL-2.0 brings as an option next to Apache-2.0, but well.

We're obviously staying on the Apache-2.0 terms, so to prevent confusion, I kept the older `LICENSE` file which doesn't have the full text of the GPL-2.0 as an option.

Faless: They added the GPLv2 because the FSF keeps saying that Apache 2.0 is not compatible with the GPLv2 (though they say it's compatible with the GPLv3).
There's some details here: https://www.apache.org/licenses/GPL-compatibility.html
And (briefly) in the release notes: https://github.com/Mbed-TLS/mbedtls/releases/tag/v2.28.6

Faless (approved):
LGTM

akien-mga:
Cherry-picked for 3.6.
Cherry-picked for 3.5.4.

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

PR #77699, opened 2023-05-31
Preserve vector ratio when dragging through 0

KoBeWi:
Fixes #77687

PR #87740, opened 2024-01-30
Fix editor properties vector ratio breaking when an element is set to zero

ryevdokimov:
A ratio should be maintained even when an element in a vector is set to zero. 

Before: 

https://github.com/godotengine/godot/assets/105675984/71707513-9aa7-4c37-b2a1-9df3188417c1

After: 

https://github.com/godotengine/godot/assets/105675984/32a7e56f-1096-416f-b25e-125824bd9ee6


KoBeWi:
Nice fix, it partially supersedes #77699. We could probably revert the part about grabbing, because it's no longer necessary.

KoBeWi (approved):

akien-mga:
Thanks!

Issue #39221, opened 2020-06-01
Timers run inside scene drag preview


KoBeWi:
**Godot version:**

3.2.2 beta3

**Issue description:**

This is more funny bug than critical, but when you drag a scene over viewport, any Timer in that scene will run inside the preview. I discovered it when my timer connected to AnimationPlayer made the preview animated (lol).
![15jfhsY9sp](https://user-images.githubusercontent.com/2223172/83408886-a94dbf00-a413-11ea-8671-a6a94240424a.gif)
(the cursor seems to be off on the GIF for some reason, ignore it)

Luckily connecting the Timer to `queue_free()` doesn't crash the editor (just makes the preview disappear xd), so it's a rather minor bug.

**Minimal reproduction project:**

[ReproductionProject.zip](https://github.com/godotengine/godot/files/4711224/ReproductionProject.zip)

lightspot21:
Can reproduce on Linux (Arch) 3.2.2 beta3. However, `master` doesn't exhibit this behavior (has the preview functionality been merged yet?)

Calinou:
@lightspot21 `master` has had its windowing system rewritten from scratch, so some features may be missing.

akien-mga:
As discussed in #39232, let's shelve it for now, and revisit if we find it to actually be a problem worth fixing.

PR #87743, opened 2024-01-30
Fix `Camera2D` is enabled when dragging scene files to the `CanvasItemEditor`

### LINKED ISSUE #87704: Weird camera bug
### Tested versions

4.2.1.stable

### System information

Godot v4.2.1.stable - Windows 10.0.22621 - GLES3 (Compatibility) - NVIDIA GeForce GT 1030 (NVIDIA; 31.0.15.4633) - AMD Ryzen 3 1300X Quad-Core Processor (4 Threads)

### Issue description

In the 2d viewport when i drag a node that has a camera viewport camera changes to that nodes camera.

https://github.com/godotengine/godot/assets/78092286/f4c67ace-59a5-4409-a803-6e2df1f5a520

### Steps to reproduce

-Open Godot and create a project
-Create 2 nodes and one of them should have camera 2d node
-Drag node that has camera 2d to other node

N/A

WhalesState:
Fixes: https://github.com/godotengine/godot/issues/87704

### DIFF
--- a/scene/main/node.cpp
+++ b/scene/main/node.cpp
@@ -2437,7 +2437,7 @@ StringName Node::get_property_store_alias(const StringName &p_property) const {
 
 bool Node::is_part_of_edited_scene() const {
 	return Engine::get_singleton()->is_editor_hint() && is_inside_tree() && get_tree()->get_edited_scene_root() &&
-			(get_tree()->get_edited_scene_root() == this || get_tree()->get_edited_scene_root()->is_ancestor_of(this));
+			get_tree()->get_edited_scene_root()->get_parent()->is_ancestor_of(this);
 }
 #endif
 


scene/2d/camera_2d.cpp
```diff
@@ -34,11 +34,11 @@
 #include "scene/main/window.h"
 
 bool Camera2D::_is_editing_in_editor() const {
-#ifdef TOOLS_ENABLED
-	return is_part_of_edited_scene();
-#else
+	if (Engine::get_singleton()->is_editor_hint() && is_inside_tree() && get_tree()->get_edited_scene_root()) {
+		return get_tree()->get_edited_scene_root()->get_viewport() == get_viewport() ||
+				get_tree()->get_edited_scene_root() == this || get_tree()->get_edited_scene_root()->is_ancestor_of(this);
```
WhalesState, line 39: `get_tree()->get_edited_scene_root() == this` this check is useless, if the Camera2D is the edited scene root then it's viewport will be the edited scene viewport too. need to remove it.

KoBeWi:
This bug potentially affects all nodes that rely on `is_part_of_edited_scene()` and can be part of a drag preview, so the change should be done in `is_part_of_edited_scene()` I think.

WhalesState:
> This bug potentially affects all nodes that rely on `is_part_of_edited_scene()` and can be part of a drag preview, so the change should be done in `is_part_of_edited_scene()` I think.

Yes, but the main issue is that the dragged scenes is instantiated and is added to the Editor root while dragging, this behavior will result in more issues like running the tool scripts while dragging, especially if it's a heavy scene.

`is_part_of_edited_scene()` works as expected, but while dragging it will return false because a scene instance will be added to the editor node root before dropping it to the scene.

KoBeWi:
Which is why I suggested that the change can be done in `is_part_of_edited_scene()`, so that it checks whether the node is under scene root viewport, not the scene itself. While that would make it technically incorrect, the only nodes under the root viewport are the scene and the drag previews, so it shouldn't have any negative impact.

WhalesState:
Is it fine to add this `get_tree()->get_edited_scene_root()->get_viewport() == get_viewport()` ?

```cpp
bool Node::is_part_of_edited_scene() const {
	return Engine::get_singleton()->is_editor_hint() && is_inside_tree() && get_tree()->get_edited_scene_root() && 
			(get_tree()->get_edited_scene_root() == this || get_tree()->get_edited_scene_root()->get_viewport() == get_viewport() || get_tree()->get_edited_scene_root()->is_ancestor_of(this));
}
```

KoBeWi:
It reminded me of this old issue: https://github.com/godotengine/godot/issues/39221

The best condition would be: `Engine::get_singleton()->is_editor_hint() && is_inside_tree() && EditorNode::get_singleton()->get_scene_root()->is_ancestor_of(this)`, but since this is Node code, the EditorNode can't be used. The problem with your check is that if a scene is empty, the issue will still be happening. Though it's still less likely to appear.

I'd suggest this for now:
```C++
bool Node::is_part_of_edited_scene() const {
	return Engine::get_singleton()->is_editor_hint() && is_inside_tree() && get_tree()->get_edited_scene_root() && 
			get_tree()->get_edited_scene_root()->get_parent()->is_ancestor_of(this);
}
```

KoBeWi (approved):

akien-mga:
Thanks!

PR #87745, opened 2024-01-30
OpenXR: Allow changing play area mode during active session

dsnopek:
Currently, we don't allow changing the OpenXR reference space type when there is already an active session running. This was done because we didn't know the use case, and didn't want to include something that developers could misused.

However, in [this comment](https://github.com/godotengine/godot/pull/87235#discussion_r1469898831), @m4gr3d mentioned a use case for it in the (still in-progress) XR editor.

So, this PR allows changing it during an active session!

In my testing, changing the reference space can be a little jarring when going into or out of the `STAGE` reference space, depending on the size and shape of your guardian area and where you happen to be standing, since it uses the center of that space as it's reference point. However, switching between `LOCAL` and `LOCAL_FLOOR` isn't so bad, because they have the same X and Z coordinates (it's only the Y that is different).

I'm not entirely sure how to document this, but I've made an attempt! Feedback welcome :-)

### DIFF
--- a/doc/classes/XRInterface.xml
+++ b/doc/classes/XRInterface.xml
@@ -141,6 +141,7 @@
 			<param index="0" name="mode" type="int" enum="XRInterface.PlayAreaMode" />
 			<description>
 				Sets the active play area mode, will return [code]false[/code] if the mode can't be used with this interface.
+				[b]Note:[/b] Changing this after the interface has already been initialized can be jarring for the player, so it's recommended to recenter on the HMD with [method XRServer.center_on_hmd] (if switching to [constant XRInterface.XR_PLAY_AREA_STAGE]) or make the switch during a scene change.
 			</description>
 		</method>
 		<method name="start_passthrough">
--- a/modules/openxr/openxr_api.cpp
+++ b/modules/openxr/openxr_api.cpp
@@ -680,71 +680,85 @@ bool OpenXRAPI::is_reference_space_supported(XrReferenceSpaceType p_reference_sp
 	return false;
 }
 
-bool OpenXRAPI::setup_spaces() {
-	XrResult result;
+bool OpenXRAPI::setup_play_space() {
+	ERR_FAIL_COND_V(session == XR_NULL_HANDLE, false);
 
 	XrPosef identityPose = {
 		{ 0.0, 0.0, 0.0, 1.0 },
 		{ 0.0, 0.0, 0.0 }
 	};
 
-	ERR_FAIL_COND_V(session == XR_NULL_HANDLE, false);
+	XrReferenceSpaceType new_reference_space;
+	XrSpace new_play_space = XR_NULL_HANDLE;
+	bool will_emulate_local_floor = false;
 
-	// create play space
-	{
-		emulating_local_floor = false;
+	if (is_reference_space_supported(requested_reference_space)) {
+		new_reference_space = requested_reference_space;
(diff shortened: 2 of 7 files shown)


m4gr3d (approved):
Looks great!

modules/openxr/openxr_api.cpp
```diff
@@ -1262,10 +1276,14 @@ void OpenXRAPI::set_view_configuration(XrViewConfigurationType p_view_configurat
...
 }
 
-void OpenXRAPI::set_requested_reference_space(XrReferenceSpaceType p_requested_reference_space) {
-	ERR_FAIL_COND(is_initialized());
-
+bool OpenXRAPI::set_requested_reference_space(XrReferenceSpaceType p_requested_reference_space) {
 	requested_reference_space = p_requested_reference_space;
+
+	if (is_initialized()) {
+		return setup_play_space();
```
BastiaanOlij, line 1283: One thing we should add here is calling into the XRServer and resetting our reference frame transform. A user shouldn't be using `center_on_hmd` in combination with `local` and `local floor` reference spaces but just in case they did...

Similarly we should probably do that as well when the user trigger a recenter just before we emit the pose recenter signal, but only if the reference space isn't `stage`.

BastiaanOlij (approved):
LGTM. 

Made me realise there is a missing feature in reference spaces though, the user can trigger a recenter and will be positioned accordingly. But I don't think we can.
Like you mention in your intro, we probably want to trigger a recenter as part of changing the reference space (though I would leave this up to the user so they have choice).

I'll raise this with the WG, to see if there is a function to trigger a recenter from the game side, or if this is something that needs to be added.

modules/openxr/openxr_api.cpp
dsnopek: > One thing we should add here is calling into the XRServer and resetting our reference frame transform. A user shouldn't be using `center_on_hmd` in combination with `local` and `local floor` reference spaces but just in case they did...

Ah, good point! If they were in `STAGE`, used `center_on_hmd()`, then switched to `LOCAL` or `LOCAL_FLOOR`, they'd be stuck with the old reference frame.

In my latest push I've added an `XRServer::clear_reference_frame()` function which is called when changing the reference space. I've also exposed it, because I could imagine scenarios where a developer would want to clear any previous recentering, for example, when switching from VR to MR mode.

> Similarly we should probably do that as well when the user trigger a recenter just before we emit the pose recenter signal, but only if the reference space isn't `stage`.

I'm less sure about this one. If we clear it when switching to `LOCAL` or `LOCAL_FLOOR`, and we give the developer the ability to clear it when they need to, I'm not sure this is necessary.

doc/classes/XRInterface.xml
```diff
@@ -141,6 +141,7 @@
 			<param index="0" name="mode" type="int" enum="XRInterface.PlayAreaMode" />
 			<description>
 				Sets the active play area mode, will return [code]false[/code] if the mode can't be used with this interface.
+				Changing this after the interface has already been initialized can be jarring for the player, so it's recommended to recenter on the HMD or make the switch during a scene change.
```
Mickeon, line 144: This kind of statement usually fits snuggly in a note. I also assume that "recenter" here refers to a method.
```suggestion
				[b]Note:[/b] Changing this after the interface has already been initialized can be jarring for the player, so it's recommended to recenter on the HMD with [method XRServer.center_on_hmd] or make the switch during a scene change.
```

dsnopek: Thanks! Included in my latest push (with a little more detail about when it makes sense to recenter on the HMD)

akien-mga:
Thanks!

PR #87750, opened 2024-01-30
Wayland: Suspend window after frame timeout or suspend state

### LINKED ISSUE #87963: Wayland: Shader `TIME` is unreliable (slow shaders) due to frame pacing method
### Tested versions

- Reproducible in 4.3.dev [b4e2a24c1]
- Not reproducible in anything earlier than 4.3.dev [7e0f7d3ab] (specific to the Wayland backend)

### System information

Godot v4.3.dev (b4e2a24c1) - Arch Linux #1 SMP PREEMPT_DYNAMIC Thu, 01 Feb 2024 10:30:35 +0000 - Wayland - Vulkan (Forward+) - dedicated AMD Radeon RX 6650 XT (RADV NAVI23) () - AMD Ryzen 5 2600 Six-Core Processor (12 Threads)

### Issue description

On Wayland, ideally, the compositor tells us when to draw. The driver handles vertical synchronization by waiting indefinitely until we get a "frame event". 

As a render-bound application, we can't rely on this feature, as we'd stall the main loop indefinitely if the window is hidden or just occluded, so we handle those events manually by using a bunch of hacks to constantly tick at a high fixed rate (2000 Hz) and trigger the renderer only exactly when we get a frame event.

Unfortunately, there seems to be an edge case where the shaders' `TIME` variable is step in 2000ths-of-a-second increments instead of the actual render rate, slowing them _considerably_. The FPS are also being reported incorrectly to 2000 FPS but that's not intrinsic to the problem.

In other words, we're ticking at a very high fixed rate and the renderer is getting confused, slowing shaders down.

This approach is pretty overkill and, although it could definitely be improved (such as by dynamically changing the tick rate), I think some different approach such as the one described in #87750 would be a better option, all in all, which would also fix this issue.

For more info see [this post by emersion](https://emersion.fr/blog/2018/wayland-rendering-loop/) which describes the problem in more detail, although we can't use their same approach.

See also #82222 for a deeper analysis and tentative fix of this general main loop edge case.

### Steps to reproduce

1. Edit the MRP with a Wayland-enabled build (requires commit e0f7d3abd4a02f423a5b95616840e028a9af7fa or later).
2. If needed change the `target_refresh_rate` variable of the `mrp.gd` script to match the refresh rate of the screen you're going to run the MRP on.
3. Run the project and notice how the shader-modulated icon on the left is going extra slow compared to the process equivalent.
4. With the aid of tools such as MangoHud or mesa's debug layers, notice how the reported FPS of 2000 is not correct (it's the tick rate, _not_ the framerate).
5. Press `T` to limit the FPS and ticking rate to the refresh rate set.
6. Notice how the shader is now running properly.

[graphics-delta-mrp.zip](https://github.com/godotengine/godot/files/14159484/graphics-delta-mrp.zip)

### LINKED ISSUE #88248: GPU Particles have very slow framerate on Wayland
### Tested versions

- Reproducible in: Godot v4.3.dev3

### System information

Godot v4.3.dev3.mono - Fedora Linux 39 (Workstation Edition) - Wayland - Vulkan (Forward+) - dedicated AMD Radeon RX 6700 XT (RADV NAVI22) () - Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz (8 Threads)

### Issue description

GPU particles seem to run at something like 1/100 speed.

https://github.com/godotengine/godot/assets/15041324/7f55ec30-684d-4095-b2eb-878529415295

### Steps to reproduce

1. Open Godot with  --display-driver wayland.
2. Open MRP
3. Observe Particle systems

[mrpwp.zip](https://github.com/godotengine/godot/files/14239404/mrpwp.zip)

deralmas:
Depends on #88374.

Fixes #87963.

This is a pretty popular approach that took a while for me to wrap my head around and which only recently got "official" support through an update (xdg_shell version 6), so I think that this is all-in-all a better option than the overkill 2000Hz ticking we have now :P

Basically, we wait for a frame event and, if either too much time passes or we get the new `suspended` state, we consider the window as "hidden" and stop drawing, ticking by the low usage rate.

This should work great for KDE and Mutter, which support the new state, but not yet for sway, which is still stuck at a very old xdg_shell version and thus falls back to the timeout approach.

Be aware that if we rely on timing out the engine will have to stall for the whole timeout, which _could_ be problematic but doensn't seem like it. Further testing is needed.

Special thanks go to the guys over at #wayland on OFTC, who very patiently explained me this approach way too many times.

---

FTR, the current timeout is 1 second, but I think that we could go as low as 200ms without any big consequence. Further testing is needed. **Edit:** Seems to not cause big issues, I think we can keep it like this for now.

~~Note that the new xdg-shell extension is not implemented in this PR, as I can't really test is as sway doesn't implement it yet.~~ **Edit:** It is implemented now! Sway still doesn't support it though :(

*Bugsquad edit:*
- Also fixes #88248.

### DIFF
--- a/platform/linuxbsd/os_linuxbsd.cpp
+++ b/platform/linuxbsd/os_linuxbsd.cpp
@@ -127,14 +127,6 @@ void OS_LinuxBSD::alert(const String &p_alert, const String &p_title) {
 	}
 }
 
-int OS_LinuxBSD::get_low_processor_usage_mode_sleep_usec() const {
-	if (DisplayServer::get_singleton() == nullptr || DisplayServer::get_singleton()->get_name() != "Wayland" || is_in_low_processor_usage_mode()) {
-		return OS::get_low_processor_usage_mode_sleep_usec();
-	}
-
-	return 500; // Roughly 2000 FPS, improves frame time when emulating VSync.
-}
-
 void OS_LinuxBSD::initialize() {
 	crash_handler.initialize();
 
--- a/platform/linuxbsd/os_linuxbsd.h
+++ b/platform/linuxbsd/os_linuxbsd.h
@@ -127,8 +127,6 @@ public:
 
 	virtual void alert(const String &p_alert, const String &p_title = "ALERT!") override;
 
-	virtual int get_low_processor_usage_mode_sleep_usec() const override;
-
 	virtual bool _check_internal_feature_support(const String &p_feature) override;
 
 	void run();
--- a/platform/linuxbsd/wayland/display_server_wayland.cpp
+++ b/platform/linuxbsd/wayland/display_server_wayland.cpp
@@ -867,11 +867,11 @@ bool DisplayServerWayland::window_is_focused(WindowID p_window_id) const {
 }
 
 bool DisplayServerWayland::window_can_draw(DisplayServer::WindowID p_window_id) const {
-	return frame;
+	return !suspended;
 }
 
 bool DisplayServerWayland::can_any_window_draw() const {
-	return frame;
+	return !suspended;
 }
 
 void DisplayServerWayland::window_set_ime_active(const bool p_active, DisplayServer::WindowID p_window_id) {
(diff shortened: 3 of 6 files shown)


akien-mga (approved):
Tested on Fedora 39 KDE Wayland, works well. I can confirm the bug with the MRP from #87963 and this PR fixes it.

@pracedru also confirmed this PR fixes #88248.

akien-mga:
> @akien-mga does the 1sec timeout cause issues when a window gets toggled in/out of view, especially in regards to audio? I couldn't test it yet.
> 
> Edit: TBC, even if it did it would probably still cause less problems than ticking at 2000hz, but I wonder if this is something we should be aware of.

I didn't notice any issue, even after doing some more testing now. In verbose mode, I never saw `Frame timeout, suspended.`, so I'm not sure if I'm doing what's needed to trigger it.

I tried:
- Running a project from the editor (which turns the editor to low process usage mode while the game window is focused), and having a looping sound play in the editor at the same time (no sound problem, even when toggling between both windows).
- Running a project from command line and making the window lose focus, both with and without low processor usage mode.

akien-mga:
> You don't trigger it by losing focus, instead by suspending (e.g. hiding) the window. Try to minimize the project for a few seconds or put it behind another window.

I still don't seem to trigger that logic when I minimize the window or make it fully covered by another one.

> Basically, as per the comments, the whole engine will stall for one second (we could go as low as 200ms as per OP) once it gets suspended and only _then_ enable low usage mode. _That's_ the dirty heuristic.

Well we could give 200ms a try and see if anyone reports issues, WDYT? Otherwise I'm fine merging as is and see if anyone runs into issues with that delay.

akien-mga:
> > I still don't seem to trigger that logic when I minimize the window or make it fully covered by another one.
> 
> A window is considered suspended here when it doesn't get a frame event for one second. A quick search in a `WAYLAND_DEBUG=1` log for `frame` should clear any doubts.
> 
> I'll test ASAP to confirm whether this works, but I have a feeling that KDE is still ticking minimized windows.

Indeed I still see `frame` ticks when minimized with `WAYLAND_DEBUG=1`.

akien-mga (approved):
Tested again briefly on KDE, seems to work fine!

akien-mga:
Thanks!

PR #87756, opened 2024-01-30
Allow all tool modes to select

### LINKED ISSUE #87608: Involuntary object transformations, due to freeform mouse drag handling in viewport
### Tested versions

I'm using v4.2.stable.official [46dc27791]

### System information

Windows 10; only tested in 3D, might be the same in 2D

### Issue description

I often accidentally move or rotate nodes a tiny bit around. I often notice that much later. This is especially annoying with objects that are supposed to be in precise locations, like parts of build kits.

### Steps to reproduce

The four modes Select, Move, Rotateand Scale Mode allow moving/rotating/scaling the currently selected object(s) even when the cursor is not near the gizmo or the object is actually seen on screen. I understand the upside: this allows transforming objects based on the current viewport space, instead of world/object space. But this also makes it very easy to accidentally transform things with one or more misplaced clicks.

Another minor gripe: in Move/Rotate/Scale Mode, even if you just click without actually dragging (mostly because I always forget that I can't select another object if I'm not in Select Mode), it creates an entry in the undo history.

So I guess the main question is if there is a way to block Godot from moving 3D objects in Select or any of the Translation Modes if I click/drag the cursor in the scene but not hovering the gizmo? In any case I would rather propose to extend the "Use Local Space (T)" button to allow switching between global/local/viewport, and get rid of the input recognition outside of the gizmo.

N/A

ryevdokimov:
Implements: [https://github.com/godotengine/godot-proposals/issues/9007](https://github.com/godotengine/godot-proposals/issues/9007)

Would fix: #87608

Would benefit: https://github.com/godotengine/godot/pull/102176

It would also make basically all tools work like "Select Mode", which would probably need to be renamed, and is therefore also an extension of #86804

Example: 

https://github.com/godotengine/godot/assets/105675984/bc2cd78e-207a-4d2f-90a3-d65da845e896

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -1950,33 +1950,37 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 
 					bool node_selected = get_selected_count() > 0;
 
-					if (node_selected && ((spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_SELECT && b->is_command_or_control_pressed()) || spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_ROTATE)) {
-						begin_transform(TRANSFORM_ROTATE, false);
-						break;
-					}
-
-					if (node_selected && spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_MOVE) {
-						begin_transform(TRANSFORM_TRANSLATE, false);
-						break;
-					}
-
-					if (node_selected && spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_SCALE) {
-						begin_transform(TRANSFORM_SCALE, false);
-						break;
-					}
-
-					if (after != EditorPlugin::AFTER_GUI_INPUT_CUSTOM) {
+					if (after != EditorPlugin::AFTER_GUI_INPUT_CUSTOM && !b->is_alt_pressed()) {
 						// Single item selection.
 						clicked = _select_ray(b->get_position());
 
-						selection_in_progress = true;
+						if (clicked.is_valid() && !editor_selection->is_selected(ObjectDB::get_instance<Node>(clicked))) {
+							selection_in_progress = true;
+							break;
+						}
 
 						if (clicked.is_null()) {
 							// Default to region select.
 							cursor.region_select = true;
 							cursor.region_begin = b->get_position();
 							cursor.region_end = b->get_position();
+							break;
 						}
+					}
+
+					if (!clicked_wants_append && node_selected && ((spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_SELECT && b->is_command_or_control_pressed()) || spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_ROTATE)) {
+						begin_transform(TRANSFORM_ROTATE, false);
+						break;
+					}
+
+					if (!clicked_wants_append && node_selected && spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_MOVE) {
+						begin_transform(TRANSFORM_TRANSLATE, false);
+						break;
+					}
+
+					if (!clicked_wants_append && node_selected && spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_SCALE) {
+						begin_transform(TRANSFORM_SCALE, false);
+						break;
 					}
 
 					surface->queue_redraw();
@@ -2034,7 +2038,6 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 						set_message("");
 						spatial_editor->update_transform_gizmo();
 					}
-					surface->queue_redraw();
 				}
 
 			} break;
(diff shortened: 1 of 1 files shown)


ratkingsminion:
I'm not sure if this is some on my side, but with this PR I have the following behaviour (when having a 3D object selected and dragging the cursor (LMB) over another object):

Select Mode: Selected object gets translated in viewport space
Select Mode, with Alt pressed: Nothing

Move Mode: Selected object gets translated in viewport space
Move Mode, with Alt pressed: The same

Rotate Mode: Selected object gets translated(!) in viewport space
Rotate Mode, with Alt pressed: Selected object gets rotated in viewport space

Scale Mode: Selected object gets translated(!) in viewport space
Scale Mode, with Alt pressed: Selected object gets scaled (local space)

What I would have expected:

Select Mode: Nothing
Select Mode, with Alt pressed: Selected object gets translated in viewport space

Move Mode: Nothing
Move Mode, with Alt pressed: Selected object gets translated in viewport space

Rotate Mode: Nothing
Rotate Mode, with Alt pressed: Selected object gets rotated in viewport space

Scale Mode: Nothing
Scale Mode, with Alt pressed: Selected object gets scaled (local space)

Interesting enough when I do Alt+RMB, the other objects gets selected.

ryevdokimov:
I just pushed a change that should resolve that. It looks like even in 4.2 stable in select mode when you drag over some kind of geometry it will translate the node(s), which I don't fully understand why that would be the case. The changes I just made make it so it always defaults to region select.

ratkingsminion:
Indeed, now everything behaves like I'd expect it to, at least in my limited test scene.

ryevdokimov:
Good to hear. Assuming there isn't any more bugs, the only thing left to finish this draft is to update the tooltips, but I'm not sure that's worth doing unless this gets more traction and because of a different potential issue I found: #87924

AThousandShips:
For future reference I'd suggest just rebasing your branches on top of some other fix if you want to include it, makes it much easier to rebase after the other PR has been merged, now you need to potentially handle merge conflicts

ryevdokimov:
My heart is in the right place, but you're always there to make it right. Thanks again.

editor/editor_settings.cpp
```diff
@@ -668,7 +668,7 @@ void EditorSettings::_load_defaults(Ref<ConfigFile> p_extra_config) {
 	// 3D: Navigation
 	_initial_set("editors/3d/navigation/invert_x_axis", false);
 	_initial_set("editors/3d/navigation/invert_y_axis", false);
-	EDITOR_SETTING(Variant::INT, PROPERTY_HINT_ENUM, "editors/3d/navigation/navigation_scheme", 0, "Godot,Maya,Modo")
+	EDITOR_SETTING(Variant::INT, PROPERTY_HINT_ENUM, "editors/3d/navigation/navigation_scheme", 0, "Godot,Maya,Modo,Temp")
```
ryevdokimov, line 671: I don't know what I would call this.

Calinou (commented):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/c2ba0a864610e826da4d2993140dedf14752da68), it works as expected.

However, I'm not sure why we need an editor setting for this (and why this is disabled by default either). Is this done to cater for 3D editor navigation schemes where you hold <kbd>Alt</kbd> to perform various actions?

ryevdokimov:
I would agree to having this behavior be the default, I just didn't know how much we were willing to break off from the original behavior, where clicking outside the selected node still performs the action of the current tool mode. 

Having all tool modes be able to select, and having that original behavior go away seems like a big change, but personally due to issues like https://github.com/godotengine/godot/issues/87608, and with the introduction of https://github.com/godotengine/godot/pull/102176, I think the pros outweigh the cons, so I'd be happy to adjust this PR if there is consensus on this.

ryevdokimov:
I'll just go ahead and adjust the PR have this behavior be the default, so it can be judged by its own merits, and then if we need to do something that accommodates the old behavior we can discuss adding it.

ryevdokimov:
Fixed an issue, where trying to switch the active selected node causes micro-transforms, and shift-click-dragging regardless of tool mode will region select, regardless of the tool mode similar to Blender.

Calinou (approved):
Tested locally, it works as expected. Code looks good to me.

I think this makes sense usability-wise. Being able to box select regardless of the existing selection under the mouse cursor is appreciated too (previously, this only worked in Select mode).

Note that <kbd>Shift</kbd> for box selection does not take priority if you are currently hovering the manipulator gizmo in any mode, but this isn't a new issue. (This behavior could also get in the way if you're using precise grid snapping with <kbd>Ctrl + Shift</kbd>, so it should be discussed in a proposal first.)

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -1950,35 +1950,39 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
...
-						break;
-					}
-
-					if (after != EditorPlugin::AFTER_GUI_INPUT_CUSTOM) {
+					if (after != EditorPlugin::AFTER_GUI_INPUT_CUSTOM && !b->is_alt_pressed()) {
 						// Single item selection.
 						clicked = _select_ray(b->get_position());
 
-						selection_in_progress = true;
+						if (clicked.is_valid() && !editor_selection->is_selected(Object::cast_to<Node>(ObjectDB::get_instance(clicked)))) {
```
AThousandShips, line 1957:
```suggestion
						if (clicked.is_valid() && !editor_selection->is_selected(ObjectDB::get_instance<Node>(clicked))) {
```
New preferred syntax

AThousandShips (approved):

Repiteo:
Thanks!

Nodragem:
Hello! 
 
I am in Modo navigation mode, which is best for pen tablet users (e.g. artists?), and similar to Blender's "emulate 3 button mouse". In this mode, to hold `alt` + `left-click` rotates the view.

In 4.5dev3, when I left-click and drag I have a selection box whatever the transformation mode I am in (i.e. selection, move, rotate, scale). However, when I press `alt` and `left-click`, I am rotating the viewport as usual and it is not affecting the selected node. 

Does the `alt` transform has to be activated in the Editor Settings? in any case it will conflict with the Modo and Maya 3D navigation scheme.  

I would recommend to avoid as much as possible to rely on modifier keys (such as `alt`), as they are quite hard to maintain in terms of conflicts, and it makes the functionality hard to discover.

If I understand well, the main goals of this PR is to NOT use the whole screen as a button for viewport aligned translation, rotation and all direction scaling, to 1) avoid accidental transform, and 2) allow box selection in all transform mode. 
Rather than using `alt`, I would suggest to add new visual elements to the gizmo as in Blender:
- a circle in the center of the translate gizmo for viewport aligned translation
- a circle around the rotation gizmo for viewport aligned rotation
- a yellow box in the center of the scale gizmo for all direction scaling

It would indeed be less easy to implement, but this way, these operations are still discoverable and they won't conflict with the 3D navigation schemes.

PR #87757, opened 2024-01-30
Do not reflect the origin lines in a mirror

### LINKED ISSUE #87684: 3D editor grid origin lines appear in ReflectionProbes
### Tested versions

- Reproducible in: 4.3.dev 0bcc0e92b
- Not reproducible in: 4.2.1.stable

### System information

Godot v4.3.dev (0bcc0e92b) - Fedora Linux 39 (KDE Plasma) - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4090 (nvidia; 545.29.06) - 13th Gen Intel(R) Core(TM) i9-13900K (32 Threads)

### Issue description

3D editor grid origin lines appear in ReflectionProbes.

This is a regression from https://github.com/godotengine/godot/pull/83895. cc @clayjohn

### Origin lines shown (default)

Screenshot_20240128_191806

### Origin lines hidden in the View menu

*Remember to hide and show the ReflectionProbe to force it to update.*

Screenshot_20240128_191819

### Steps to reproduce

- Add a ReflectionProbe node near the origin, with a reflective object nearby.
- Notice how the red/green/blue origin lines appear on the reflection.

[test_reflection_probe_origin_lines.zip](https://github.com/godotengine/godot/files/14077415/test_reflection_probe_origin_lines.zip)

hakro:
Fixes #87684

Do not ignore culling on the origin lines, in order for their layer mask to be taken into account.

(+ a quick spacing change for consistency with the other shaders in the file)

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -6490,7 +6490,7 @@ void Node3DEditor::_init_indicators() {
 // 3D editor origin line shader.
 
 shader_type spatial;
-render_mode blend_mix,cull_disabled,unshaded, fog_disabled;
+render_mode blend_mix, cull_disabled, unshaded, fog_disabled;
 
 void vertex() {
 	vec3 point_a = MODEL_MATRIX[3].xyz;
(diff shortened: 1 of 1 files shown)


Calinou:
This may have been done intentionally so that lines don't appear if you're far away from the world origin.

cc @clayjohn

clayjohn (commented):
Looks like we will have to add a custom AABB instead using `instance_set_custom_aabb()`. The old origin lines went 1048576 units in each direction. So to match that, we should use AABB((-1048576, -1048576, -1048576), (1048576 * 2, 1048576 * 2, 1048576 * 2))

hakro:
Thank you both for the feedback.

@clayjohn  I've run some tests both before and after this PR, and with an added AABB.

They all yielded the same result. The lines die off at coordinate 1'000'000 in every direction anyways.

As an example, by adding this line with the values you suggested : 
```
RS::get_singleton()->instance_set_custom_aabb(origin_instance, AABB(Vector3(-1048576,-1048576,-1048576), Vector3(1048576 * 2, 1048576 * 2, 1048576 * 2)));
```
the line only gets to 1M. Here is a screen shot of that on the `x` axis

Screenshot_20240131_150254

So unless I'm not understanding this correctly (which might very well be the case), the AABB is not needed?

clayjohn (approved):
@hakro Thanks for checking. 

Looking at the surrounding code I am reminded that, in the end, I went with a MultiMesh approach for this. The MultiMesh calculates its own AABB that should cover the extents of the lines. So indeed, adding the custom AABB is not necessary.

Calinou (approved):

image

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #87758, opened 2024-01-30
Allow GDExtensions to register virtual methods and call them on scripts

### LINKED ISSUE #82267: BIND_VIRTUAL_METHOD in GDextensions doesn't create a virtual function in GDscript
### Godot version

4.1.1

### System information

Pop!_OS 22.04

### Issue description

Using `BIND_VIRTUAL_METHOD` script in `_bind_methods()` doesn't create a virtual function in GDScript.

### Steps to reproduce

1. Create simple GDextension
2. Build it
3. Add to the project

```C++
//MyClass.h
#ifndef MYCLASS_H
#define MYCLASS_H

#include <godot_cpp/classes/node.hpp>

namespace godot{
class MyClass : public Node{
	GDCLASS(MyClass, Node)

protected:
	static void _bind_methods();

	void foo();

};
}

#endif

//MyClass.cpp
#include "my_class.h"

#include <godot_cpp/core/class_db.hpp>
#include <godot_cpp/variant/variant.hpp>

using namespace godot;


void MyClass::_bind_methods(){
	BIND_VIRTUAL_METHOD(MyClass, foo);
}


void MyClass::foo(){

}
```
After building the extension, you can `MyClass` in the editor. However, you can't override `foo()` function as GDscript doesn't see it and it isn't shown in documentation section.

N/A

dsnopek:
Currently, GDExtension classes can implement virtual methods registered in Godot. However, GDExtensions classes can't register their own virtual methods, which scripts attached to them can implement.

This PR adds the necessary changes to the GDExtension interface to allow this!

PR https://github.com/godotengine/godot-cpp/pull/1377 adds support for this to godot-cpp.

Fixes https://github.com/godotengine/godot/issues/82267

_Production edit: closes godotengine/internal-team-priorities#37_


paddy-exe (approved):
I checked the code to the best of my understanding. Reading the source code the changes makes sense to me. I haven't tested it though.

core/extension/gdextension_interface.h
```diff
@@ -381,6 +381,18 @@ typedef struct {
...
 
+typedef struct {
+	GDExtensionStringNamePtr name;
+	uint32_t method_flags; // Bitfield of `GDExtensionClassMethodFlags`.
+
+	GDExtensionPropertyInfo return_value;
+	GDExtensionClassMethodArgumentMetadata return_value_metadata;
+
+	uint32_t argument_count;
+	GDExtensionPropertyInfo *arguments_info;
```
Bromeon, line 392: I would name this `arguments` -- would be consistent with `return_value` and also `GDExtensionMethodInfo::arguments`: 

https://github.com/godotengine/godot/blob/a0106607e889fb0ed4ab5bf84a4b2707c3ac5598/core/extension/gdextension_interface.h#L242-L255

core/extension/gdextension_interface.h
```diff
@@ -2268,6 +2280,34 @@ typedef GDExtensionObjectPtr (*GDExtensionInterfaceObjectGetInstanceFromId)(GDOb
...
+ * @since 4.3
+ *
+ * Checks if this object has a script with the given method.
+ *
+ * @param p_object A pointer to the Object.
+ * @param p_method A pointer to a StringName identifying the method.
+ *
+ * @returns true if the method exists.
+ */
+typedef GDExtensionBool (*GDExtensionInterfaceObjectHasScriptMethod)(GDExtensionConstObjectPtr p_object, GDExtensionConstStringNamePtr p_method);
```
Bromeon, line 2294: > Checks if this object has a script with the given method.

My understanding from this documentation is:
- if no script is attached, this always returns `false`
- if a user defines _any_ method in the script with that name -- whether it overrides a virtual method or not -- it returns true
- unclear whether it applies to `_ready`, `_process` and so on?

Is this correct? Maybe we can clarify these semantics a bit?
Also if/how it relates to [`ClassDB::has_method()`](https://docs.godotengine.org/en/stable/classes/class_classdb.html#class-classdb-method-class-has-method) 🤔

core/extension/gdextension_interface.cpp
```diff
@@ -1194,6 +1194,33 @@ static GDObjectInstanceID gdextension_object_get_instance_id(GDExtensionConstObj
...
+	if (script_instance) {
+		return script_instance->has_method(method);
+	}
+	return false;
+}
+
+static void gdextension_object_call_script_method(GDExtensionObjectPtr p_object, GDExtensionConstStringNamePtr p_method, const GDExtensionConstVariantPtr *p_args, GDExtensionInt p_argument_count, GDExtensionUninitializedVariantPtr r_return, GDExtensionCallError *r_error) {
+	Object *o = (Object *)p_object;
+	const StringName method = *reinterpret_cast<const StringName *>(p_method);
+	const Variant **args = (const Variant **)p_args;
```
Bromeon, line 1211: I know this happens a lot in existing code, but is there any guideline regarding when to choose `reinterpret_cast` and when C-style casts?

Bromeon (commented):
Very cool feature! 😊

core/extension/gdextension_interface.h
dsnopek: I used `arguments_info` to be consistent with `GDExtensionClassMethodInfo::arguments_info` which is (in my opinion) a more similar struct to this new one than `GDExtensionMethodInfo`. It's unfortunate that we already have these inconsistencies in the API. :-/ Given that, I'm not sure what the right choice is here.

core/extension/gdextension_interface.h
dsnopek: That is correct! Regarding:

> * unclear whether it applies to `_ready`, `_process` and so on?

It applies to those as well. If the script has any method by the given name, it returns `true`.

> Also if/how it relates to [`ClassDB::has_method()`](https://docs.godotengine.org/en/stable/classes/class_classdb.html#class-classdb-method-class-has-method) 🤔

`ClassDB::has_method()` is looking for method binds, aka native methods, not script methods.

Anyway, I'll see if I can update the docs to clarify this a bit!

core/extension/gdextension_interface.cpp
dsnopek: Unfortunately, I don't think we have any guidelines. This looks as it does because I copy-pasted it from `gdextension_variant_call()`. :-) I could change the 2nd cast to `reinterpret_cast`, but that doesn't really improve the overall situation. We could save fixing this for an overhaul of all the casts used in GDExtension, where we could establish some sort of guidelines?

core/extension/gdextension_interface.h
Bromeon: Ah, I see. Well, then we have to pick our poison 😀 I'm fine with keeping it more consistent with the other one.

I'm not sure if this is somehow worth documenting, in case the question comes up again? Maybe not 🤔

dsnopek: Ok, let's stick with this name in this case. Probably the best place to document this is on the big list of "things to change when we can completely break compatibility" :-)

core/extension/gdextension_interface.cpp
Bromeon: Yes, let's evaluate it outside this PR then 🙂

core/extension/gdextension_interface.h
Bromeon: Do we have such a list yet? 😀

core/extension/gdextension_interface.cpp
dsnopek: I vote for addressing the whole casting situation in a dedicated PR, and leaving it out-of-scope for this PR. The changes here match what we've already got.

dsnopek:
@Bromeon Thanks for the review! I've attempted to address your points in my latest push :-)

core/extension/gdextension_interface.h
Bromeon: Looks good! About `_ready` etc, I assume it only returns true if they are actually _overridden_ in the script, not just if the type offers such a method -- correct?

That's the only unclarity I could see remaining, but if you think it's clear enough as-is, that's also fine with me.

Bromeon:
Thanks, responded! Will also try to test this with godot-rust this or next week.

core/extension/gdextension_interface.h
dsnopek: > About `_ready` etc, I assume it only returns true if they are actually _overridden_ in the script, not just if the type offers such a method -- correct?

Yes, it only returns true if that specific script has an implementation for the named method. It doesn't care if there is a virtual method registered for that type (which I think we don't even track in release builds anyway)

> That's the only unclarity I could see remaining, but if you think it's clear enough as-is, that's also fine with me.

Personally, I think it's clear enough, so I'll mark as resolved. But we can always add more to the docs if we end up getting questions about this :-)

Bromeon (commented):
General question: I assume there is no way to _require_ GDScript to implement a certain method, right? As in abstract/pure virtual methods.

To implement such semantics, user code should probably resort to a default implementation that causes an error?

dsnopek:
> General question: I assume there is no way to _require_ GDScript to implement a certain method, right? As in abstract/pure virtual methods.

Not via this API. However, the godot-cpp PR adds a `GDVIRTUAL_REQUIRED_CALL()` (which is based on a macro of the same name in Godot) which will print an error if the virtual method isn't implemented. This is in contrast to the `GDVIRTUAL_CALL()` macro which just evaluate to `false` if the method isn't implemented. So, I'd expect other bindings to do something similar, generating errors when needed.

> To implement such semantics, user code should probably resort to a default implementation that causes an error?

Yep, something like that! In godot-cpp we're mimicking how Godot does it with macros and templates, but other bindings can do whatever makes sense for that particular binding.

core/extension/gdextension_interface.h
```diff
@@ -381,6 +381,18 @@ typedef struct {
 	GDExtensionVariantPtr *default_arguments;
 } GDExtensionClassMethodInfo;
 
+typedef struct {
+	GDExtensionStringNamePtr name;
+	uint32_t method_flags; // Bitfield of `GDExtensionClassMethodFlags`.
+
+	GDExtensionPropertyInfo return_value;
+	GDExtensionClassMethodArgumentMetadata return_value_metadata;
```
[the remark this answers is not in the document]
dsnopek: For a method without a return value, you set `return_value.type` to `GDEXTENSION_VARIANT_TYPE_NIL`.

I suppose we could add a `has_return_value` property for consistency with `GDExtensionClassMethodInfo`?

But [on the Godot side](https://github.com/godotengine/godot/pull/87758/files#diff-685d5eb25bcc42c360e9dbdce28a401900c320f7e5b6c75c27a4d43c7292e09cR1624), we need to have property info for the return, so we'd just be setting the type to nil if that bool was set.

I think there's other parts of the API that use nil to indicate a `void` method as well. For example, I think that's how `GDExtensionMethodInfo` works (as opposed to `GDExtensionClassMethodInfo` which has the boolean). With all the inconsistencies, it's hard to pick which thing to be consistent with. :-)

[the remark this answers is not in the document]
Bromeon: I see. Maybe we should decide on a "ideal convention" and add TODOs in the places that deviate? I.e. directly next to the fields that have the "wrong" one.

That would ensure that if we ever have a v2 of a certain struct, we can make it right, so the inconsistencies would balance out over time.

What do you think? And could we make a call on what should be "ideal", I don't have any favorite 😀

Bromeon:
So, I implemented this in Rust, seems to work well! 🦀 

Two questions that came up during implementation -- please let me know if that's by design, or I missed something:

1. It doesn't seem possible to register both a regular (early-binding) and a virtual function with the same name. So while the binding itself _can_ initiate a dynamic dispatch (calling the script method or a C++/Rust fallback), the Godot engine _cannot_ (via `Object.call`, or from GDScript code). It will always call the script method, or fail if there is no script.
   - The workaround would be to have a separate method that calls the dynamic-dispatch mechanism from the binding. This method could then be registered with Godot, making the dispatch available to GDScript.

2. Overridden virtual functions in scripts cannot call their `super` implementation, i.e. C++/Rust code (again because it doesn't exist, from the Godot perspective).
   - Also here, this could be addressed by having a separate method that contains only the fallback code.

Bromeon (approved):
The only thing left might be a short `TODO` on the conventions (see [here](https://github.com/godotengine/godot/pull/87758#discussion_r1485603148)), so we avoid proliferating even more styles 😉 

None of the points in my last post (`super` calls, script-side dynamic dispatch) is a blocker in my opinion -- even in case we eventually plan to change anything in that regard, this PR is a great foundation either way 🙂 Once merged, bindings can follow up with their implementation, too 🚀

dsnopek:
@Bromeon:

> So, I implemented this in Rust, seems to work well! 🦀

Great, thanks!

> 1. It doesn't seem possible to register both a regular (early-binding) and a virtual function with the same name. So while the binding itself _can_ initiate a dynamic dispatch (calling the script method or a C++/Rust fallback), the Godot engine _cannot_ (via `Object.call`, or from GDScript code). It will always call the script method, or fail if there is no script.

This is just how virtual methods in Godot work (including in C++ modules): you can't really have a default implementation of a virtual method.

The usual pattern is to have a non-virtual method named _without_ an underscore which calls a virtual method named _with_ an underscore. Then if the class wants to do some default behavior if the virtual method isn't implemented (checked via `GDVIRTUAL_IS_OVERRIDDEN()`), or process the return value of the virtual method in some way, it's done in the non-virtual method. And it's the non-virtual method that's the API that folks are expected to call directly, not the virtual method.

This is, of course, different than how virtual methods work in C++, but it's how the binding layer does them.

> 2\. Overridden virtual functions in scripts cannot call their `super` implementation, i.e. C++/Rust code (again because it doesn't exist, from the Godot perspective).

Yes, but this sort of makes sense in the context of virtual methods not having default implementations.

However, I do think we should add the ability to call `super` implementations (I have the start of a branch that does this - I probably won't finish it in time for Godot 4.3, though), but not to call the default implementation (because there isn't one). Instead, this is needed for when a GDExtension implements a virtual method from a parent class, and the script also wants to implement that virtual method, but it doesn't want to completely wipe out the GDExtensions implementation. I think this will be especially important when we get around to implementing the ability for a GDExtension class to descend from a parent class in another GDExtension.

However, we don't need to account for that in this PR - that's for a different PR in the future. :-)

core/extension/gdextension_interface.h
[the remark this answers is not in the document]
dsnopek: Alright, I spent some time thinking about this, and it feels like `GDExtensionClassMethodInfo` is the odd one out.

So, in `GDExtensionClassVirtualMethod`, I've renamed `arguments_info` to just `arguments` (to match `GDExtensionMethodInfo`) and added some `@todo`'s by the inconsistent bits of `GDExtensionClassMethodInfo` that we can consider changing for consistency in a future version of that struct.

dsnopek:
Thanks!

> The only thing left might be a short `TODO` on the conventions (see [here](https://github.com/godotengine/godot/pull/87758#discussion_r1485603148)), so we avoid proliferating even more styles 😉

I've added the TODO's in my latest push :-)

Bromeon:
> This is just how virtual methods in Godot work (including in C++ modules): you can't really have a default implementation of a virtual method.

Thanks for the elaboration, that makes sense. I will then document this accordingly in Rust and see if we can also establish some convention.

---

> However, I do think we should add the ability to call `super` implementations [...], but not to call the default implementation (because there isn't one). Instead, this is needed for when a GDExtension implements a virtual method from a parent class, and the script also wants to implement that virtual method, but it doesn't want to completely wipe out the GDExtensions implementation.

I see, so this would work for extensions overriding Godot virtual methods (which are then further overridden by scripts), but not for virtual methods declared by the extension. This _might_ (?) cause some confusion but if it's well documented (including the patterns to work around it), should be OK.


---

> I've added the TODO's in my latest push :-)

Looks great, thanks a lot! From my side, this PR is good to go 🙂

akien-mga:
Thanks!

limbonaut:
Thanks! :star_struck:

Bromeon:
@limbonaut The GDExtension API generally targets the next minor Godot version (4.3 here), binary compatibility gets very complex otherwise. So this won't be backported.

You can however either build Godot yourself or wait [for the next 4.3-dev release](https://godotengine.org/download/archive/).

OctoD:
Hello! Thanks for this PR, it will solve a lot of problems with my extension. Before opening an issue, I would report that defining a virtual method like the one below causes godot to crash on startup (without outputting an error). This is caused by the `... Args` of `PropertyInfo` passed to the `MethodInfo` struct.

```cpp
ClassDB::add_virtual_method(get_class_static(), MethodInfo(Variant::BOOL, "_method_name", PropertyInfo(Variant::STRING, "value")));
```

Issue #82830, opened 2023-10-05
ResourceLoader.load() inconsistently ignores/replaces cache when loading PackedScenes


dandeliondino:
### Godot version

4.1.1-stable, 4.2-dev6

### System information

Windows 11

### Issue description

I am working on a 2D game. Different levels are assigned to different environments (like overworld, underground, underwater). Most of the components are the same for each level, but the Sprite2Ds, AnimatedSprite2Ds and TileMaps all have slightly different artwork for each environment.

Instead of creating different versions of all the level components for every environment, I planned to swap the textures at runtime by using `take_over_path()` prior to loading each level.

However, this yielded inconsistent results, so I created this MRP to figure out what was happening.

There appear to be at least 3 inconsistencies:
(1) When the texture is assigned to a property that is *not* in a subresource, using ResourceLoader.load() with `CACHE_MODE_REPLACE` does not change textures already on-screen. When the texture is in a subresource, it will.
(2) If the texture is in a subresource and its node is nested inside a saved branch of the scene being instantiated, it *won't* ignore or replace the cache. If the texture is *not* in a subresource, it will.
(3) When the node is an AnimatedSprite2D and its SpriteFrames resource is saved externally, it *won't* ignore or replace the cache. When the node is a TileMap and the TileSet resource is saved externally, it *will* ignore or replace the cache.

I don't see other issues describing *exactly* the same things for Godot 4, but #59669, which originated in Godot 3, might be related. And, while #59686 and #62408 are for Godot 3.x, the root cause (the cache parameter not being passed consistently to other functions) seems like it could be relevant here.

I obtained the same results with Godot 4.1.1-stable and Godot 4.2-dev6, and in with both Compatibility and the Forward+ renderers.

### Steps to reproduce

This project contains a script that instantiates Sprite2Ds, AnimatedSprite2Ds and TileMaps before and after using take_over_path() to swap their shared texture "frog.png" with "snail.png". For the Sprite2D, the texture is assigned to the texture property. For AnimatedSprite2D and TileMap, the texture is assigned inside of its subresource.

The script on main.tscn has the following options:

![options](https://github.com/godotengine/godot/assets/68911895/2ce8b90c-5c87-4568-93f7-7ee37bf2cb1f)

"Instantiate after only" does not instantiate any scenes until *after* take_over_path() is called (preventing "frog.png" from being cached).

"Scene types" are described below.

"Cache mode" is the CacheMode enum value passed to ResourceLoader.load().

[...]

PR #87759, opened 2024-01-30
Respect cache mode when loading external dependencies.

HolonProduction:
This line is part of #86960, but this seems to have quite a way ahead of it and I need it to fix the CI for #86973.

Also this might fix
https://github.com/godotengine/godot/issues/82830

### DIFF
--- a/core/io/resource_format_binary.cpp
+++ b/core/io/resource_format_binary.cpp
@@ -430,7 +430,7 @@ Error ResourceLoaderBinary::parse_variant(Variant &r_v) {
 						path = remaps[path];
 					}
 
-					Ref<Resource> res = ResourceLoader::load(path, exttype);
+					Ref<Resource> res = ResourceLoader::load(path, exttype, cache_mode);
 
 					if (res.is_null()) {
 						WARN_PRINT(String("Couldn't load resource: " + path).utf8().get_data());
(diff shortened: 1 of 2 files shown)


AThousandShips:
This should also be done in the binary format if done here, the same code is used there with the reuse mode

HolonProduction:
Done

RandomShaper:
I think this change makes sense. I'd like more people looking into it, though.

One thing I'd like to point out is that, maybe, the editor is relying on the current behavior not to have to reload external resources when loading a scene:
https://github.com/godotengine/godot/blob/fb10e67fefb85af3b8f5abd30db561588d8456c6/editor/editor_node.cpp#L3857

If that were the case, we may need to add a mode, such as `CACHE_MODE_REPLACE_ROOT_REUSE_OTHERS`, but with a more compact name.

RandomShaper:
> One thing I'd like to point out is that, maybe, the editor is relying on the current behavior not to have to reload external resources when loading a scene:
> 
> https://github.com/godotengine/godot/blob/fb10e67fefb85af3b8f5abd30db561588d8456c6/editor/editor_node.cpp#L3857
> 
> If that were the case, we may need to add a mode, such as `CACHE_MODE_REPLACE_ROOT_REUSE_OTHERS`, but with a more compact name.

On an _n_-th thought, if _replace_ is meant to change the way subresources are loaded, I does make sense that the mode is jsut propagated.

HolonProduction:
The change seems to be far more critical than I thought it would be. Closing in favor of #88664.

PR #87760, opened 2024-01-30
Extract BottomPanel from EditorNode

kitbdev:
- related: #84193

Extracts all functionality related to the bottom panel from EditorNode.
No functionality should change.

Details
- Added `editor/gui/editor_bottom_panel.h/cpp`
- Removed `bottom_hb`, since it was never used anywhere.
- Added `bool p_visible` param to `make_item_visible()` (`make_bottom_panel_item_visible()`) since EditorNode needed it.
- Added `save_layout_to_config()` and `load_layout_from_config()`.
- Renamed variables and methods to be more clear.
- It needs `set_process(true)` as part of #72420, this can be removed in the future when a better fix is made for it.

### DIFF
--- a/editor/debugger/editor_debugger_node.cpp
+++ b/editor/debugger/editor_debugger_node.cpp
@@ -38,6 +38,7 @@
 #include "editor/editor_settings.h"
 #include "editor/editor_string_names.h"
 #include "editor/editor_undo_redo_manager.h"
+#include "editor/gui/editor_bottom_panel.h"
 #include "editor/gui/editor_run_bar.h"
 #include "editor/inspector_dock.h"
 #include "editor/plugins/editor_debugger_plugin.h"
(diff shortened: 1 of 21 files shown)


editor/debugger/editor_debugger_node.cpp
```diff
@@ -38,6 +38,7 @@
 #include "editor/editor_settings.h"
 #include "editor/editor_string_names.h"
 #include "editor/editor_undo_redo_manager.h"
+#include "editor/gui/editor_bottom_panel.h"
```
KoBeWi, line 41: Worth checking whether `editor_node.h` can be removed (in all files).

editor/editor_node.cpp
```diff
@@ -7094,61 +6923,13 @@ EditorNode::EditorNode() {
 
 	// Bottom panels.
 
-	bottom_panel = memnew(PanelContainer);
+	bottom_panel = memnew(EditorBottomPanel);
 	bottom_panel->add_theme_style_override("panel", theme->get_stylebox(SNAME("BottomPanel"), EditorStringName(EditorStyles)));
```
KoBeWi, line 6927: This override should be moved too.

editor/gui/editor_bottom_panel.cpp
```diff
@@ -0,0 +1,278 @@
...
+static const String META_TEXT_TO_COPY = "text_to_copy";
+
+void EditorBottomPanel::_notification(int p_what) {
+	switch (p_what) {
+		case NOTIFICATION_THEME_CHANGED: {
+			expand_button->set_icon(get_editor_theme_icon(SNAME("ExpandBottomDock")));
+		} break;
+
+		case NOTIFICATION_PROCESS: {
+			is_switching = false;
```
KoBeWi, line 55: This can be handled either by deferred call or by one-shot connecting to `process_frame` signal of SceneTree. It's more efficient than enabling processing for this node.

kitbdev: Originally, I didn't want to change the behavior since this is just for extraction and I don't know the edge cases for it.

I tried removing `is_switching` without `set_pressed_no_signal()` but I couldn't reproduce the issues from #72420, it seems that some other changes since then have fixed the issue.

KoBeWi: `set_pressed_no_signal()` also suppresses `toggled` signal.

KoBeWi (approved):

akien-mga:
Thanks!

PR #87764, opened 2024-01-31
Wayland: Implement `window_get_native_handle`

deralmas:
Related to a discussion regarding OpenXR done in https://github.com/Riteo/godot/pull/4#discussion_r1467363946

This will be the most useful for stuff like OpenXR, although we'd need a way to eventually also expose the EGL handles.

### DIFF
--- a/platform/linuxbsd/wayland/display_server_wayland.cpp
+++ b/platform/linuxbsd/wayland/display_server_wayland.cpp
@@ -557,6 +557,37 @@ Vector<DisplayServer::WindowID> DisplayServerWayland::get_window_list() const {
 	return ret;
 }
 
+int64_t DisplayServerWayland::window_get_native_handle(HandleType p_handle_type, WindowID p_window) const {
+	MutexLock mutex_lock(wayland_thread.mutex);
+
+	switch (p_handle_type) {
+		case DISPLAY_HANDLE: {
+			return (int64_t)wayland_thread.get_wl_display();
+		} break;
+
+		case WINDOW_HANDLE: {
+			return (int64_t)wayland_thread.window_get_wl_surface(p_window);
+		} break;
+
+		case WINDOW_VIEW: {
+			return 0; // Not supported.
+		} break;
+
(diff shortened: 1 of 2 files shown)


platform/linuxbsd/wayland/display_server_wayland.cpp
```diff
@@ -557,6 +557,35 @@ Vector<DisplayServer::WindowID> DisplayServerWayland::get_window_list() const {
...
+		case WINDOW_VIEW: {
+			return 0; // Not supported.
+		} break;
+
+		case OPENGL_CONTEXT: {
+			if (egl_manager) {
+				return (int64_t)egl_manager->get_context(p_window);
+			}
+			return 0;
+		} break;
```
bruvzg, line 581:
```suggestion
#ifdef GLES3_ENABLED
		case OPENGL_CONTEXT: {
			if (egl_manager) {
				return (int64_t)egl_manager->get_context(p_window);
			}
			return 0;
		} break;
#endif // GLES3_ENABLED
```

platform/linuxbsd/wayland/display_server_wayland.cpp
```diff
@@ -557,6 +557,35 @@ Vector<DisplayServer::WindowID> DisplayServerWayland::get_window_list() const {
...
+int64_t DisplayServerWayland::window_get_native_handle(HandleType p_handle_type, WindowID p_window) const {
+	MutexLock mutex_lock(wayland_thread.mutex);
+
+	switch (p_handle_type) {
+		case DISPLAY_HANDLE: {
+			return (int64_t)wayland_thread.get_wl_display();
+		} break;
+
+		case WINDOW_HANDLE: {
+			return (int64_t)wayland_thread.window_get_wl_surface(p_window);
```
bruvzg, line 569: Also, should it be raw surface handle or exported handle from `zxdg_exporter_v1`?

bruvzg (approved):

akien-mga:
Thanks!

PR #87765, opened 2024-01-31
Wayland: Disable backend at build-time if wayland-scanner is missing

### LINKED ISSUE #87762: Building on X11 now requires explicit `wayland=no` parameter
### Tested versions

- Reproducible in v4.3.dev.custom_build [313f623b9]
- Not reproducible in v4.3.dev.custom_build [17e7f85c0]
- Likely introduced in #86180

### System information

Debian GNU/Linux 12 (bookworm) 12 - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2070 (nvidia)

### Issue description

The following command no longer works for compiling godot in a X11 non-wayland setup:
```
pyston-scons use_llvm=yes linker=mold platform=linuxbsd target=editor dev_mode=yes dev_build=yes scu_build=yes '-j 5'
```
because it fails with the following error:
```
sh: 1: wayland-scanner: not found
wayland-scanner not found. Aborting.
```

I would expect, that I can compile Godot in a non-wayland-system without wayland-scanner installed

I was able to workaround this by two methods
- Install `libwayland-bin` on Debian (not an ideal solution, because on a non-wayland system it should not be necessary to install wayland-packages)
- use for compiling
  - pyston-scons **wayland=no** use_llvm=yes linker=mold platform=linuxbsd target=editor dev_mode=yes dev_build=yes scu_build=yes '-j 5'
  - (not ideal, because the buildsystem should work with default values).

### Steps to reproduce

Build Godot on a non-wayland X11 system with
```
pyston-scons use_llvm=yes linker=mold platform=linuxbsd target=editor dev_mode=yes dev_build=yes scu_build=yes '-j 5'
```

Problem happens without any MRP.

deralmas:
Fixes #87762

This allows previous X11-only setups to still build Godot with default settings. Note that compilation will still abort if wayland-scanner is present but not the various Wayland libraries.

---

BTW, I wonder if we should disable either backend depending if their libraries are missing instead of straight up aborting, now that we have two of them. This should cover 99% of issues though, as if one has `wayland-scanner` they probably have everything anyways.

### DIFF
--- a/platform/linuxbsd/detect.py
+++ b/platform/linuxbsd/detect.py
@@ -208,8 +208,8 @@ def configure(env: "Environment"):
 
     if env["wayland"]:
         if os.system("wayland-scanner -v") != 0:
-            print("wayland-scanner not found. Aborting.")
-            sys.exit(255)
+            print("wayland-scanner not found. Disabling wayland support.")
+            env["wayland"] = False
 
     if env["touch"]:
         env.Append(CPPDEFINES=["TOUCH_ENABLED"])


Rindbee:
Why not merge the options `x11`, `wayland` into one option?

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
> BTW, I wonder if we should disable either backend depending if their libraries are missing instead of straight up aborting, now that we have two of them. This should cover 99% of issues though, as if one has `wayland-scanner` they probably have everything anyways.

We could, but I think for now this is good enough as is. Once we make Wayland enabled by default, we might want to ensure there's a bit more intention there from the user, and not just making assumptions based on what they had installed in their host machine (e.g. maybe the user only cares about X11 for their local builds, but they want the export templates they build to support all configs their users may have, which includes Wayland). So they need to be clearly aware if they compile X11-only export templates.

PR #87769, opened 2024-01-31
Fix a regression that breaks gizmo transforming when 'View Gizmos' is off

ryevdokimov:
#86804 introduces a regression. This PR fixes it by separating the logic from the gizmo entirely, and just looks if a node is selected.


akien-mga (approved):

akien-mga:
Thanks!

PR #79140, opened 2023-07-07
Make Vector<T>::duplicate() const

aaronfranke:
I tried to duplicate a `const PackedStringArray` and I noticed that it was impossible, but it should be possible.

PR #87778, opened 2024-01-31
Make Callable `bind` method const

aaronfranke:
This allows using `bind` from a `const` context. `bind` is an internal method exposed through a custom binding, so we can change it. In the future we should make this same change to `bindv` but that would break compatibility and there is no way to bind compat methods for Variant yet (see the discussion in #79140).


akien-mga (approved):

akien-mga:
Thanks!

PR #87781, opened 2024-01-31
Fix crash when previewing a scene with a mesh as the root node

aaronfranke:
This PR fixes a crash that happens during an edge case where the imported scene's root node is a mesh. The code replaces each ImporterMeshInstance3D node with a MeshInstance3D, but in the case of the scene root, the reference to the scene was being left as a dangling pointer. Now it sets the scene to the new root node. Also I added a comment.

I have used `unlikely` here because this check will only be true at most once for each file and usually 0 times.

Test file: [crashlight.glb.zip](https://github.com/godotengine/godot/files/14111598/crashlight.glb.zip)

In master: *crash*

With this PR: 
Screenshot 2024-01-31 at 6 19 00 AM


reduz (approved):

akien-mga:
Thanks!

PR #87783, opened 2024-01-31
Make `GDScriptUtilityCallable` return call error when method is invalid

AThousandShips:
Follow-up to:
* #86823

For completeness, all other callables appear to have proper error reporting when possible


modules/gdscript/gdscript_utility_callable.cpp
```diff
@@ -84,6 +84,9 @@ void GDScriptUtilityCallable::call(const Variant **p_arguments, int p_argcount,
 	switch (type) {
 		case TYPE_INVALID:
 			ERR_PRINT(vformat(R"(Trying to call invalid utility function "%s".)", function_name));
```
dalexeev, line 86:
```suggestion
			r_return_value = vformat(R"(Trying to call invalid utility function "%s".)", function_name);
```

dalexeev (approved):
Looks good to me. The error shouldn't happen (there shouldn't be a way to obtain an invalid utility callable), but anyway thank you for fixing it!

akien-mga:
Thanks!

AThousandShips:
Thank *you*!

PR #87787, opened 2024-01-31
ResourceImporterScene: Replace animation bool with an import type string enum

aaronfranke:
The current code for ResourceImporterScene has a boolean for importing an animation library, and if false, it imports as a scene. This is not ideal, a boolean does not make much sense because there are more than 2 conceivable possibilities, we don't want to hard-code "animation or scene". This PR replaces this bool with an enum.

In the future we can add more possibilities to this enum, like importing a mesh resource. This PR is part of https://github.com/godotengine/godot-proposals/issues/7494 I am working on it in small pieces to avoid huge unreviewable PRs.

### DIFF
--- a/editor/import/3d/resource_importer_scene.cpp
+++ b/editor/import/3d/resource_importer_scene.cpp
@@ -96,9 +96,10 @@ void EditorSceneFormatImporter::get_import_options(const String &p_path, List<Re
 	GDVIRTUAL_CALL(_get_import_options, p_path);
 }
 
-Variant EditorSceneFormatImporter::get_option_visibility(const String &p_path, bool p_for_animation, const String &p_option, const HashMap<StringName, Variant> &p_options) {
+Variant EditorSceneFormatImporter::get_option_visibility(const String &p_path, const String &p_scene_import_type, const String &p_option, const HashMap<StringName, Variant> &p_options) {
 	Variant ret;
-	GDVIRTUAL_CALL(_get_option_visibility, p_path, p_for_animation, p_option, ret);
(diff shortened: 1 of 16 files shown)


aaronfranke:
@lyuma My initial plan was for this only to be extendable from within ResourceImporterScene itself, in which case you can add to the enum. However a String or StringName would also work well, I can change it to that. I'm not totally sure about the requirements to allow this class to be extended outside of Godot but I'm all good with leaving the option open.

aaronfranke:
@lyuma I made it use a String, and @fire requested that I request your review.

editor/import/3d/resource_importer_scene.cpp
```diff
@@ -245,51 +249,68 @@ void EditorScenePostImportPlugin::_bind_methods() {
...
 
 String ResourceImporterScene::get_visible_name() const {
-	return animation_importer ? "Animation Library" : "Scene";
+	// This is displayed on the UI. Friendly names here are nice but not vital, so fall back to the type.
+	if (_scene_import_type == "PackedScene") {
+		return "Scene";
+	} else if (_scene_import_type == "AnimationLibrary") {
+		return "Animation Library";
+	}
+	return _scene_import_type;
```
lyuma, line 269: there's a function that is used elsewhere that converts identifiers to human readable (like AnimationLibrary -> Animation Library). we should call that instead of hardcoding

lyuma (approved):
I realize this is only changing the internal api which means plugins don't get to use the new types, but we are going to rework the public apis in another change in the future.

it makes sense to start with this internal change as is.

akien-mga:
Thanks!

PR #87789, opened 2024-01-31
Add OS::get_cwd virtual

L2750558108:
It seems that we get current working directory by creating a DirAccess without path init and use get_current_dir(). It doesn't look intuitive.
In fact, the DirAccess finally calls the cwd getting function of specific platform. This PR move them to the OS::get_cwd virtual.

### DIFF
--- a/core/os/os.h
+++ b/core/os/os.h
@@ -179,6 +179,7 @@ public:
 	virtual Error shell_open(String p_uri);
 	virtual Error shell_show_in_file_manager(String p_path, bool p_open_folder = true);
 	virtual Error set_cwd(const String &p_cwd);
+	virtual String get_cwd() const;
 
 	virtual bool has_environment(const String &p_var) const = 0;
 	virtual String get_environment(const String &p_var) const = 0;
--- a/drivers/unix/os_unix.cpp
+++ b/drivers/unix/os_unix.cpp
@@ -700,6 +700,17 @@ Error OS_Unix::set_cwd(const String &p_cwd) {
 	return OK;
 }
 
+String OS_Unix::get_cwd() const {
+	String cwd;
+	char ret[PATH_MAX];
+	char *temp = getcwd(ret, PATH_MAX);
+	ERR_FAIL_NULL_V(temp, "");
+	if (cwd.parse_utf8(ret) != OK) {
+		cwd = ret; // No utf8, maybe latin1?
+	}
+	return cwd;
+}
+
 bool OS_Unix::has_environment(const String &p_var) const {
 	return getenv(p_var.utf8().get_data()) != nullptr;
 }
--- a/platform/windows/os_windows.cpp
+++ b/platform/windows/os_windows.cpp
@@ -914,6 +914,21 @@ Error OS_Windows::set_cwd(const String &p_cwd) {
 	return OK;
 }
 
+String OS_Windows::get_cwd() const {
+	const DWORD size = GetCurrentDirectoryW(0, nullptr);
+
+	Char16String buffer;
+	buffer.resize((int)size);
+	if (GetCurrentDirectoryW(size, (wchar_t *)buffer.ptrw()) == 0) {
+		return OS::get_cwd();
+	}
+
+	String result;
+	result.parse_utf16(buffer.ptr());
+
+	return result;
+}
+
 Vector<String> OS_Windows::get_system_fonts() const {
 	if (!dwrite_init) {
 		return Vector<String>();
--- a/core/os/os.cpp
+++ b/core/os/os.cpp
@@ -327,6 +327,10 @@ Error OS::set_cwd(const String &p_cwd) {
 	return ERR_CANT_OPEN;
 }
 
+String OS::get_cwd() const {
+	return get_executable_path();
+}
+
 Dictionary OS::get_memory_info() const {
 	Dictionary meminfo;
 
--- a/drivers/unix/os_unix.h
+++ b/drivers/unix/os_unix.h
@@ -60,6 +60,7 @@ public:
 	virtual Error get_dynamic_library_symbol_handle(void *p_library_handle, const String p_name, void *&p_symbol_handle, bool p_optional = false) override;
 
 	virtual Error set_cwd(const String &p_cwd) override;
+	virtual String get_cwd() const override;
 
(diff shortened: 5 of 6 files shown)


Calinou:
Please generate documentation by running the compiled editor with `--doctool`, then fill in the changes made to `doc/classes/OS.xml`.

drivers/unix/os_unix.cpp
```diff
@@ -699,6 +699,16 @@ Error OS_Unix::set_cwd(const String &p_cwd) {
 
 	return OK;
 }
+String OS_Unix::get_cwd()
+{
+	String cwd = "";
+	char ret[2048];
+	ERR_FAIL_NULL_V(getcwd(ret, 2048), ERR_BUG);
```
bruvzg, line 706:
```suggestion
	char ret[PATH_MAX];
	ERR_FAIL_NULL_V(getcwd(ret, PATH_MAX), ERR_BUG);
```

On some systems, `PATH_MAX` can be 4K.

platform/windows/os_windows.cpp
```diff
@@ -914,6 +916,14 @@ Error OS_Windows::set_cwd(const String &p_cwd) {
 	return OK;
 }
 
+String OS_Windows::get_cwd()
+{
+	WCHAR ret[2048];
+	GetCurrentDirectoryW(2048, ret);
+	String cwd = String::utf16((const char16_t *)ret);
+	return cwd;
```
bruvzg, line 924: I think we are using 2048 buffer for the same function, but some places of the Windows API mention 32K as max. path length. I'm not sure if it's relevant in this case, but it might be more robust to use "preflighting" syntax and allocate specific buffer size:

```c++
	const DWORD size = GetCurrentDirectoryW(0, nullptr);

	Char16String buffer;
	buffer.resize((int)size);
	if (GetCurrentDirectoryW(size, (wchar_t *)buffer.ptrw()) == 0) {
		return OS::get_cwd();
	}

	String result;
	result.parse_utf16(buffer.ptr());

	return result;
```

platform/windows/os_windows.cpp
```diff
@@ -914,6 +916,14 @@ Error OS_Windows::set_cwd(const String &p_cwd) {
 	return OK;
 }
 
+String OS_Windows::get_cwd()
+{
```
bruvzg:
```suggestion
String OS_Windows::get_cwd() {
```

drivers/unix/os_unix.cpp
```diff
@@ -699,6 +699,16 @@ Error OS_Unix::set_cwd(const String &p_cwd) {
 
 	return OK;
 }
+String OS_Unix::get_cwd()
+{
```
bruvzg:
```suggestion
String OS_Unix::get_cwd() {
```

core/os/os.h
```diff
@@ -179,6 +179,7 @@ class OS {
 	virtual Error shell_open(String p_uri);
 	virtual Error shell_show_in_file_manager(String p_path, bool p_open_folder = true);
 	virtual Error set_cwd(const String &p_cwd);
+	virtual String get_cwd() const;
```
bruvzg, line 182: Note: neither `set_cwd`, not newly added `get_cwd` are bound, and therefore it will be accessible only from the engine C++ code (not from GDScript or GDExtension). If you want't it to be accessible, a wrapper should be added to the `CoreBind` (`OS` is a special case and its methods can't be bound directly).

Here's links to one of such wrappers, for example (`OS.get_name`):

Wrapper definition:
https://github.com/godotengine/godot/blob/9adb7c7d130c6d4eb0e80b92d6eebd71fac3384d/core/core_bind.h#L175

Wrapper implementation:
https://github.com/godotengine/godot/blob/9adb7c7d130c6d4eb0e80b92d6eebd71fac3384d/core/core_bind.cpp#L351-L353

Wrapper binding:
https://github.com/godotengine/godot/blob/9adb7c7d130c6d4eb0e80b92d6eebd71fac3384d/core/core_bind.cpp#L601

L2750558108:
> Please generate documentation by running the compiled editor with `--doctool`, then fill in the changes made to `doc/classes/OS.xml`.

Do functions that are not registered in the bind_method also need it?

bruvzg (commented):
Also take a look at our code style [guidelines](https://docs.godotengine.org/en/stable/contributing/development/cpp_usage_guidelines.html), there are some issues with it.

core/os/os.h
L2750558108: I noticed that set_cwd is not bound, so I don't plan to get_cwd the corresponding one.

drivers/unix/os_unix.cpp
```diff
@@ -699,6 +699,15 @@ Error OS_Unix::set_cwd(const String &p_cwd) {
 
 	return OK;
 }
+String OS_Unix::get_cwd() const {
```
AThousandShips, line 702:
```suggestion

String OS_Unix::get_cwd() const {
```

core/os/os.h
L2750558108: I hope it will provide an intuitive way to get the current working directory in future engine development, and it will correspond to set_cwd.
Maybe it could also be used to gradually replace the original code that used DirAccess to get the working path?

platform/windows/os_windows.cpp
```diff
@@ -52,11 +52,13 @@
 #include <direct.h>
 #include <knownfolders.h>
 #include <process.h>
+#include <processenv.h>
 #include <psapi.h>
 #include <regstr.h>
 #include <shlobj.h>
 #include <wbemcli.h>
 #include <wincrypt.h>
+#include <winnt.h>
```
AThousandShips, line 61: Why were these two includes added? They're not needed for the change, were they added automatically by your editor? If they're not needed please remove them

Delsin-Yu:
Can `OS::get_current_working_directory` be a better alternative naming-wise?

L2750558108:
> Can `OS::get_current_working_directory` be a better alternative naming-wise? Or is `cwd` a common knowledge

The name comes from another set_cwd method that already exists. I'm not sure if this should be changed

Delsin-Yu:
> I'm not sure if this should be changed

Oh, then no.

PR #87794, opened 2024-01-31
Separate and unify iterators from `Vector` and `LocalVector`

dimkKy:
Both `LocalVector` and `Vector` now use the same `VectorIterator` defined in STL-algorithms-compatible `vector_iterator.h` ([#86573](https://github.com/godotengine/godot/pull/86573) included)

`LocalVector` `if(tight)` checks are now [compile-time](https://godbolt.org/z/W1YcrK1aY), and `remove_at()` and `insert()` now use `std::memmove()` for _trivially copyable_ types to shift data.

### DIFF
--- a/core/templates/local_vector.h
+++ b/core/templates/local_vector.h
@@ -31,11 +31,14 @@
 #ifndef LOCAL_VECTOR_H
 #define LOCAL_VECTOR_H
 
+#include "core/error/error_list.h"
 #include "core/error/error_macros.h"
 #include "core/os/memory.h"
 #include "core/templates/sort_array.h"
 #include "core/templates/vector.h"
+#include "core/templates/vector_iterator.h"
 
+#include <cstring>
 #include <initializer_list>
 #include <type_traits>
 
@@ -46,38 +49,64 @@ class LocalVector {
 private:
 	U count = 0;
 	U capacity = 0;
-	T *data = nullptr;
+	T *data = nullptr; 
+
+protected:
+	_FORCE_INLINE_ void realloc_more(U p_size) {
+		CRASH_COND_MSG(p_size <= capacity, "Precondition for realloc_more not met");
+		
+		T *p_data = (T *)memrealloc(data, p_size * sizeof(T));
+		if (likely(p_data)) {
+			data = p_data;
+			capacity = p_size;
+		}
+		else {
+			CRASH_NOW_MSG(error_names[ERR_OUT_OF_MEMORY])
+		}
+	}
 
 public:
-	T *ptr() {
-		return data;
-	}
+	using Iterator = VectorIterator<T>;
+	using ConstIterator = ConstVectorIterator<T>;
 
-	const T *ptr() const {
-		return data;
-	}
+	T *ptr() { return data; }
+	const T *ptr() const { return data; }
 
 	_FORCE_INLINE_ void push_back(T p_elem) {
 		if (unlikely(count == capacity)) {
-			capacity = tight ? (capacity + 1) : MAX((U)1, capacity << 1);
-			data = (T *)memrealloc(data, capacity * sizeof(T));
-			CRASH_COND_MSG(!data, "Out of memory");
+			constexpr auto new_cap = [](U p_cap) noexcept {
+				if constexpr (tight) {
+					return p_cap + 1;
+				}
+				else {
+					return = MAX((U)1, p_cap << 1);
+				}
+			};
+
+			realloc_more(new_cap(capacity));
 		}
 
-		if constexpr (!std::is_trivially_constructible<T>::value && !force_trivial) {
+		if constexpr (!std::is_trivially_constructible_v<T> && !force_trivial) {
 			memnew_placement(&data[count++], T(p_elem));
-		} else {
+		}
+		else {
 			data[count++] = p_elem;
 		}
 	}
 
 	void remove_at(U p_index) {
 		ERR_FAIL_UNSIGNED_INDEX(p_index, count);
-		count--;
-		for (U i = p_index; i < count; i++) {
-			data[i] = data[i + 1];
+		--count;
+		if constexpr (std::is_trivially_copyable_v<T>/* && !force_trivial ?*/) {
+			std::memmove(&data[p_index], &data[p_index + 1], 
+				(count - p_index) * sizeof(T));
 		}
-		if constexpr (!std::is_trivially_destructible<T>::value && !force_trivial) {
+		else {
+			for (U i = p_index; i < count - 1; ++i) {
+				data[i] = data[i + 1];
+			}
+		}
+		if constexpr (!std::is_trivially_destructible_v<T> && !force_trivial) {
 			data[count].~T();
 		}
 	}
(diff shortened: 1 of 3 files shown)


AThousandShips (dismissed):
Using aspects of the STL library isn't allowed, this includes the `<iterator>` library, we don't need these iterators to be compliant

core/templates/local_vector.h
```diff
@@ -46,38 +49,64 @@ class LocalVector {
...
-	}
+	T *ptr() { return data; }
+	const T *ptr() const { return data; }
 
 	_FORCE_INLINE_ void push_back(T p_elem) {
 		if (unlikely(count == capacity)) {
-			capacity = tight ? (capacity + 1) : MAX((U)1, capacity << 1);
-			data = (T *)memrealloc(data, capacity * sizeof(T));
-			CRASH_COND_MSG(!data, "Out of memory");
+			constexpr auto new_cap = [](U p_cap) noexcept {
```
AThousandShips, line 77: Don't use lambda unless absolutely necessary, see [here](https://docs.godotengine.org/en/latest/contributing/development/cpp_usage_guidelines.html#lambdas)

core/templates/local_vector.h
```diff
@@ -122,37 +151,42 @@ class LocalVector {
 	_FORCE_INLINE_ bool is_empty() const { return count == 0; }
 	_FORCE_INLINE_ U get_capacity() const { return capacity; }
 	_FORCE_INLINE_ void reserve(U p_size) {
-		p_size = tight ? p_size : nearest_power_of_2_templated(p_size);
+		constexpr auto new_size = [](U _p_size) noexcept {
```
AThousandShips, line 154: Same here

core/templates/local_vector.h
```diff
@@ -46,38 +49,64 @@ class LocalVector {
...
 	}
 
 	void remove_at(U p_index) {
 		ERR_FAIL_UNSIGNED_INDEX(p_index, count);
-		count--;
-		for (U i = p_index; i < count; i++) {
-			data[i] = data[i + 1];
+		--count;
+		if constexpr (std::is_trivially_copyable_v<T>/* && !force_trivial ?*/) {
+			std::memmove(&data[p_index], &data[p_index + 1],
```
AThousandShips, line 101: I'm not sure this is a good change, and again using the standard library

core/templates/local_vector.h
```diff
@@ -162,85 +196,34 @@ class LocalVector {
...
-			return *this;
-		}
-		_FORCE_INLINE_ ConstIterator &operator--() {
-			elem_ptr--;
-			return *this;
-		}
-
-		_FORCE_INLINE_ bool operator==(const ConstIterator &b) const { return elem_ptr == b.elem_ptr; }
-		_FORCE_INLINE_ bool operator!=(const ConstIterator &b) const { return elem_ptr != b.elem_ptr; }
+	_FORCE_INLINE_ Iterator begin() { return { data }; }
```
AThousandShips, line 199:
```suggestion
	_FORCE_INLINE_ Iterator begin() { return data; }
```
Unnecessary and hurts readability, same for all

core/templates/local_vector.h
```diff
@@ -162,85 +196,34 @@ class LocalVector {
...
+		if (p_index == count) {
 			push_back(p_val);
-		} else {
+		}
+		else {
 			resize(count + 1);
-			for (U i = count - 1; i > p_pos; i--) {
-				data[i] = data[i - 1];
+			if constexpr (std::is_trivially_copyable_v<T>/* && !force_trivial ?*/) {
+				std::memmove(&data[p_index + 1], &data[p_index],
```
AThousandShips, line 213: Same here about STL

core/templates/local_vector.h
```diff
@@ -162,85 +196,34 @@ class LocalVector {
...
+					data[i] = data[i - 1];
+				}
+			}	
+			data[p_index] = p_val;
 		}
 	}
 
 	int64_t find(const T &p_val, U p_from = 0) const {
-		for (U i = p_from; i < count; i++) {
+		for (U i = p_from; i < count; ++i) {
```
AThousandShips, line 226: Revert these, unnecessary changes 🙂

core/templates/vector_iterator.h
```diff
@@ -0,0 +1,171 @@
...
+
+template <class T>
+struct ConstVectorIterator {
+	using iterator_category = std::random_access_iterator_tag;
+	using value_type = T;
+	using difference_type = ptrdiff_t;
+	using pointer = const value_type*;
+	using reference = const value_type&;
+	
+	reference operator*() const noexcept { return *elem_ptr; }
```
AThousandShips, line 49: No need for all the `noexcept` as we don't even use exceptions
```suggestion
	reference operator*() const { return *elem_ptr; }
```

AThousandShips:
Thank you for your contribution! This looks interesting, but needs quite a bit of change to follow our style and use rules, if you need more help please ask 🙂

For style see [here](https://docs.godotengine.org/en/latest/contributing/development/code_style_guidelines.html)

core/templates/local_vector.h
AThousandShips: Then I'd say to use the non-prefix version for clarity 🙂, then it's fine

dimkKy:
I have applied suggested changes, thank you for the quick response. However I do think `iterator_tag` should stay

dimkKy:
Looks like formatter broke some tabs, will fix it later

AThousandShips:
That's fine, I'll keep my "changes requested" for now, if someone on the core team approves this as it is I'll remove that, but I still think that the addition of the `<iterator>` part is unnecessary and does nothing for the engine, as it is only useful for components that we don't use

AThousandShips:
Will do a deeper dive later but my first pointers are:
* I'm unsure about the consequences of adding new templates, I worry that adding two new template declarations to pollute the name system for each pair of `Vector/LocalVector`, I'm not sure but I think it might be worse than just the names inside those types, like `Vector<T>::Iterator`
* I'm not sure about the inheritance of the two iterator types, I'd prefer them as two independent classes

One idea I just realized that might resolve all this but leave everything unchanged but improving code unification is this:
Create a macro instead, make a macro for the whole iterator and then use them in the two classes

That effectively keeps things as they were for the most part, but reduces duplication

And finally I'd suggest splitting this into two PRs, the other changes are unrelated to the iterators, and is best reviewed as a separate improvement, so please separate the iterator improvements from the allocation and memory rearrangements, this is a general preference for improvements to the engine 🙂

And thank you for your work on this in any case, it's very interesting and I think at least some of the memory related changes might well be included, even if the iterators aren't (another reason to split it up)

AThousandShips:
Not sure about the force trivial part

What I mean with the namespace is that you now add `VectorIterator<T>` as a name in the global namespace, which is not always desired, they make debug data larger

dimkKy:
Splitted into [#88420](https://github.com/godotengine/godot/pull/88420) and [#87794](https://github.com/godotengine/godot/pull/88422).

PR #87795, opened 2024-01-31
Direct3D 12: Avoid terrible leak related to command allocators

RandomShaper:
Otherwise, Godot keeps eating and eating memory just by being run.


clayjohn (commented):
Seems fine to me. But deferring to @DarioSamo for final approval

DarioSamo:
This matches what I've done on my own elsewhere so it makes sense, although I'm surprised I didn't spot it sooner either.

DarioSamo (approved):

akien-mga:
Thanks!

PR #87796, opened 2024-01-31
Namespace shader cache files by graphics API

RandomShaper:
Namely, cache files now end with `.d3d12.cache` or `.vulkan.cache` instead of just `.cache`. This avoids the whole shader cache from being invalidated (and thus regenerated) when switching rendering drivers.


clayjohn (approved):

akien-mga:
Thanks!

PR #87823, opened 2024-02-01
Don't invoke adb with no runnable Android preset

KoBeWi:
Android export plugin will periodically call adb.exe to check for available devices. It does this constantly while the editor is running, regardless if the project is targeted for Android or not. Performance implications aside, this check will sometimes fail and spit an error that reminds you that it runs in the background.

This PR tweaks this behavior. The adb will be called only if the project has an export preset for Android and it's marked runnable (AFAIK it's a requirement to run something on a device). I added a new signal that gets emitted when adding/removing preset and toggling Runnable.

### DIFF
--- a/editor/export/editor_export.cpp
+++ b/editor/export/editor_export.cpp
@@ -110,8 +110,13 @@ void EditorExport::save_presets() {
 	save_timer->start();
 }
 
+void EditorExport::emit_presets_runnable_changed() {
+	emit_signal(_export_presets_runnable_updated);
+}
+
 void EditorExport::_bind_methods() {
-	ADD_SIGNAL(MethodInfo("export_presets_updated"));
+	ADD_SIGNAL(MethodInfo(_export_presets_updated));
+	ADD_SIGNAL(MethodInfo(_export_presets_runnable_updated));
 }
 
 void EditorExport::add_export_platform(const Ref<EditorExportPlatform> &p_platform) {
@@ -135,6 +140,7 @@ void EditorExport::add_export_preset(const Ref<EditorExportPreset> &p_preset, in
 	} else {
 		export_presets.insert(p_at_pos, p_preset);
 	}
+	emit_presets_runnable_changed();
 }
 
 int EditorExport::get_export_preset_count() const {
@@ -149,6 +155,7 @@ Ref<EditorExportPreset> EditorExport::get_export_preset(int p_idx) {
 void EditorExport::remove_export_preset(int p_idx) {
 	export_presets.remove_at(p_idx);
 	save_presets();
+	emit_presets_runnable_changed();
 }
 
 void EditorExport::add_export_plugin(const Ref<EditorExportPlugin> &p_plugin) {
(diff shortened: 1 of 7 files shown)


platform/android/export/export_plugin.cpp
```diff
@@ -423,6 +423,26 @@ void EditorExportPlatformAndroid::_check_for_changes_poll_thread(void *ud) {
 		OS::get_singleton()->execute(adb, args);
 	}
 }
+
+void EditorExportPlatformAndroid::_update_preset_status() {
+	print_line("updating");
```
akien-mga, line 428: Debug line.

editor/export/editor_export.cpp
```diff
@@ -110,8 +110,13 @@ void EditorExport::save_presets() {
...
 }
 
+void EditorExport::emit_presets_changed() {
+	emit_signal(_export_presets_changed);
+}
+
 void EditorExport::_bind_methods() {
-	ADD_SIGNAL(MethodInfo("export_presets_updated"));
+	ADD_SIGNAL(MethodInfo(_export_presets_updated));
+	ADD_SIGNAL(MethodInfo(_export_presets_changed));
```
akien-mga, line 119: The names are quite similar, it's a bit hard to know at a glance which one means what.

KoBeWi: I wasn't sure how to name the new signal. It could be `export_preset_list_changed`, but that doesn't cover runnable change.

bruvzg:
The same is true for iOS, so it should have the same change as well.

editor/export/editor_export.cpp
m4gr3d: Can we just use `_export_presets_updated` to signal both an addition to the list and an update to the `runnable` state?

I don't think we need to introduce a new signal.

m4gr3d (changes requested):

editor/export/editor_export.cpp
KoBeWi: I didn't want to change what this signal does, but sure.

KoBeWi: Wait no, the reason why I didn't use the old signal is that it's emitted too often for my purposes. There is no reason to run the update code on unrelated changes.

KoBeWi:
Did similar change to iOS, but I didn't test it.

editor/export/editor_export.cpp
m4gr3d: > Wait no, the reason why I didn't use the old signal is that it's emitted too often for my purposes. There is no reason to run the update code on unrelated changes.

I think it may be a compromise worth making; after all, it'll end up being called an order of magnitude less than it's being invoked in the current codebase.

m4gr3d: Or you could create a new signal specifically for the `runnable` update.

KoBeWi:
Renamed the signal to `export_presets_runnable_updated`.

editor/export/editor_export.cpp
```diff
@@ -110,8 +110,13 @@ void EditorExport::save_presets() {
 	save_timer->start();
 }
 
+void EditorExport::emit_presets_changed() {
```
m4gr3d, line 113: The method should be renamed to match the signal name.

m4gr3d (approved):

akien-mga:
Thanks!

PR #87826, opened 2024-02-01
Expose Read Only and Write Only storage buffers

Kimau:
Enable flags so you can create buffers using the read only or write only flags which translate to BUFFER_USAGE_TRANSFER_FROM_BIT and BUFFER_USAGE_TRANSFER_TO_BIT

### DIFF
--- a/doc/classes/RenderingDevice.xml
+++ b/doc/classes/RenderingDevice.xml
@@ -1873,7 +1873,7 @@
 			Buffer can only be read by the shader using it.
 		</constant>
 		<constant name="STORAGE_BUFFER_USAGE_WRITEONLY" value="4" enum="StorageBufferUsage" is_bitfield="true">
-			Buffer can only be written to by the shader using it, usually used by an outpupt buffer to be copied back to the CPU later.
+			Buffer can only be written to by the shader using it, usually used by an output buffer to be copied back to the CPU later.
 		</constant>
 		<constant name="UNIFORM_TYPE_SAMPLER" value="0" enum="UniformType">
 			Sampler uniform.
--- a/doc/classes/RenderingDevice.xml
+++ b/doc/classes/RenderingDevice.xml
@@ -1868,6 +1868,8 @@
 			Index buffer in 32-bit unsigned integer format. This limits the maximum index that can be specified to [code]4294967295[/code].
 		</constant>
 		<constant name="STORAGE_BUFFER_USAGE_DISPATCH_INDIRECT" value="1" enum="StorageBufferUsage" is_bitfield="true">
+		<constant name="STORAGE_BUFFER_USAGE_READONLY" value="2" enum="StorageBufferUsage" is_bitfield="true">
+		<constant name="STORAGE_BUFFER_USAGE_WRITEONLY" value="4" enum="StorageBufferUsage" is_bitfield="true">
 		</constant>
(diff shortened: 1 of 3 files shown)


doc/classes/RenderingDevice.xml
```diff
@@ -1868,6 +1868,8 @@
 			Index buffer in 32-bit unsigned integer format. This limits the maximum index that can be specified to [code]4294967295[/code].
 		</constant>
 		<constant name="STORAGE_BUFFER_USAGE_DISPATCH_INDIRECT" value="1" enum="StorageBufferUsage" is_bitfield="true">
+		<constant name="STORAGE_BUFFER_USAGE_READONLY" value="2" enum="StorageBufferUsage" is_bitfield="true">
```
AThousandShips, line 1871:
```suggestion
		</constant>
		<constant name="STORAGE_BUFFER_USAGE_READONLY" value="2" enum="StorageBufferUsage" is_bitfield="true">
		</constant>
```

adamscott: It took me a little while to understand. @Kimau, you inserted the constant entry inside the other entry, XML-wise.

AThousandShips: Yep, why I suggested here because this can be cryptic to figure out error wise from the CI

Kimau: Whoops sorry, will fix

doc/classes/RenderingDevice.xml
```diff
@@ -1869,6 +1869,12 @@
 		</constant>
 		<constant name="STORAGE_BUFFER_USAGE_DISPATCH_INDIRECT" value="1" enum="StorageBufferUsage" is_bitfield="true">
 		</constant>
+		<constant name="STORAGE_BUFFER_USAGE_READONLY" value="2" enum="StorageBufferUsage" is_bitfield="true">
+			Buffer can only be read by the shader using it.
+		</constant>
+		<constant name="STORAGE_BUFFER_USAGE_WRITEONLY" value="4" enum="StorageBufferUsage" is_bitfield="true">
+			Buffer can only be written to by the shader using it, usually used by an outpupt buffer to be copied back to CPU later.
```
AThousandShips, line 1876:
```suggestion
			Buffer can only be written to by the shader using it, usually used by an outpupt buffer to be copied back to the CPU later.
```

servers/rendering/rendering_device.h
```diff
@@ -850,6 +850,8 @@ class RenderingDevice : public RenderingDeviceCommons {
 
 	enum StorageBufferUsage {
 		STORAGE_BUFFER_USAGE_DISPATCH_INDIRECT = 1,
+		STORAGE_BUFFER_USAGE_READONLY = 2,
+		STORAGE_BUFFER_USAGE_WRITEONLY = 4
```
AThousandShips, line 854:
```suggestion
		STORAGE_BUFFER_USAGE_WRITEONLY = 4,
```
To make future extensions easier (like the first value)

Kimau:
Digging though and after looking at the code more and chatting to @clayjohn not even sure this is useful.
The storage buffer stuff is always creating with GPU alloc which doesn't hit the relevant code so this PR is not effective as it used to be. Might come back to this as I'm poking at trying to solve some problems in this area for my own purposes.

Leaving open for dicussion, and if I get a better solution.

PR #87829, opened 2024-02-01
Throw exception when solution file is missing during exporting

### LINKED ISSUE #86591: Export fails, but doesn't error, when solution file is missing
### Tested versions

4.2.1.stable (mono)
4.2.2-rc (mono) ([d5064a7](https://github.com/godotengine/godot/commit/d5064a7d443d4b335169c0d02e822ee720abcc5d))

### System information

Godot v4.2.1.stable.mono - Windows 10.0.19045 - Vulkan (Forward+)

### Issue description

When exporting a project and the `.sln` file is not where expected, nothing is reported to the user.
The export finishes, seemingly successfully, but the necessary .NET binaries are not included in the export, making it crash on startup. It seems that `dotnet publish` is never invoked.

It would be helpful if the export fails and also reports a descriptive error. The editor can not handle a missing solution file. Currently it is practically impossible to figure out something went wrong.

Furthermore, when using the `--headless --export` settings, a zero exit code is reported when this happens. This is probably already tracked in #83042 though.

### Steps to reproduce

1. Create project
2. Create solution (Project -> Tools -> C#)
3. Create a new scene, and make it the main scene
4. Attach a C# script to the scene
5. Rename the solution file
6. Create an export profile that relies on `dotnet publish`, e.g. Windows Desktop
7. Run the export in an empty directory
8. Start the newly exported game

The game binary crashes on startup with exception `0xc0000005` (access violation). This is because the .NET publish is missing. 

[SlnProject.zip](https://github.com/godotengine/godot/files/13789846/SlnProject.zip)

The MRP already has the necessary setup (scene, script, export profile, missing solution file). The only thing left is to run the export using the profile.

nongvantinh:
Fixes: #86591

### DIFF
--- a/modules/mono/editor/GodotTools/GodotTools/Export/ExportPlugin.cs
+++ b/modules/mono/editor/GodotTools/GodotTools/Export/ExportPlugin.cs
@@ -90,6 +90,13 @@ namespace GodotTools.Export
                     $"Resource of type {Internal.CSharpLanguageType} has an invalid file extension: {path}",
                     nameof(path));
 
+            if (!ProjectContainsDotNet())
+            {
+                _maybeLastExportError = $"This project contains C# files but no solution file was found at the following path: {GodotSharpDirs.ProjectSlnPath}\n" +
+                    "A solution file is required for projects with C# files. Please ensure that the solution file exists in the specified location and try again.";
+                throw new InvalidOperationException($"{path} is a C# file but no solution file exists.");
+            }
+
             // TODO: What if the source file is not part of the game's C# project?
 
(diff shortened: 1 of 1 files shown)


raulsntos:
We can't do this because it would prevent non-C# projects from being exported. The C# version of Godot should still allow creating and exporting GDScript-only games, this becomes more important as we try to unify the two versions of the Godot editor.

akien-mga (approved):
The new commit seems to address @raulsntos' concern and makes sense to me.

I'll let you double check Raul.

raulsntos (approved):
LGTM. Thanks!

akien-mga:
Thanks!

PR #87830, opened 2024-02-01
Fix wrong process method in AudioStreamPlayer

### LINKED ISSUE #87802: AudioStreamPlayer does not call "finished" signal
### Tested versions

Reproducible on 313f623b9d102cc8c411ae7cab9518f98c2f87f2
Introduced by https://github.com/godotengine/godot/pull/87061

### System information

Godot v4.3.dev.mono (313f623b9) - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3070 Laptop GPU (NVIDIA; 31.0.15.3742) - 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz (16 Threads)

### Issue description

AudioStreamPlayer never calls "finished" and emits no notifications, it also does not call AudioStreamPlayerInternal::process the impact beyond no notifications is not fully understood.

The cause is that AudioStreamPlayerInternal::play_basic now calls:
  _set_process(true);
Instead of:
  set_process_internal(true);

In the Node flow this triggers a different notification type but AudioStreamPlayerInternal::notification still uses
  case Node::NOTIFICATION_INTERNAL_PROCESS: {
  process();
  } break;

The 2D + 3D versions are fine as they use the internal physics loop notification to drive calls to process.

### Steps to reproduce

If you open the attached MRP in 4.2 stable it will repeat a small sound immediately as the main scene loads and log a message saying everything is looking good.

If you run a local build of 4.3+ and open the MRP it will play the sound once everytime you reopen the scene and you wont see any log message occuring.

[MRP_AudioFinished.zip](https://github.com/godotengine/godot/files/14118721/MRP_AudioFinished.zip)

KoBeWi:
Fixes #87802

### DIFF
--- a/scene/audio/audio_stream_player_internal.cpp
+++ b/scene/audio/audio_stream_player_internal.cpp
@@ -38,7 +38,7 @@ void AudioStreamPlayerInternal::_set_process(bool p_enabled) {
 	if (physical) {
 		node->set_physics_process_internal(p_enabled);
 	} else {
-		node->set_process(p_enabled);
+		node->set_process_internal(p_enabled);
 	}
 }
 


MJacred (approved):

RPicster:
Tested this fix - works!
Would be great if it can be merged for dev4.
I can't test 4.3dev3 on my project with the bug in place as it breaks progression and I don't want to build a work around 😅

wpbirney:
Confirmed this fixes for me as well

akien-mga (approved):

akien-mga:
Thanks!

amarraff:
I'm still experiencing a problem with this in Godot v4.3.stable.official [77dcf97d8].

I'm using the base AudioStreamPlayer (not 2D or 3D). I run "audio_player.play()" with its stream set to an .ogg file; the .ogg file's "loop" property is set to "true" in its import settings.

When I run "await audio_player.finished" the same problem as in #87802 and #88264 happens - the code after the await statement never runs. **This problem is fixed** by changing the .ogg file's "loop" property to "false". This is, of course, not a viable solution when I need audio stream to loop.

I noticed that AudioStreamPlayer now has "Parameters -> Looping" but I cannot figure out how to set this via code, it's not mentioned in the docs. I tried "audio_player.looping = true", "audio_player.looking.on = true", and "audio_player.set_parameter("looping",true)". But, all of these return errors.

If this warrants a new issue, please let me know and I can make one.

RPicster:
I think this would be better as a new issue.

amarraff:
> I think this would be better as a new issue.

Will do!

PR #87831, opened 2024-02-01
Core: Implement a XOR operator for BitField

deralmas:
At least the Wayland backend had to do a horrible backwards-and-forwards with casts to achieve a XOR operation, in our case used to get a button difference.

I'm not aware of other uses, but the core chat over RC seemed favourable at a simple operator overload.

I've never worked with core types so tell me if this is a good idea and if so, if this is implemented correctly :D

### DIFF
--- a/core/variant/type_info.h
+++ b/core/variant/type_info.h
@@ -300,6 +300,7 @@ public:
 	_FORCE_INLINE_ constexpr BitField(T p_value) { value = (int64_t)p_value; }
 	_FORCE_INLINE_ operator int64_t() const { return value; }
 	_FORCE_INLINE_ operator Variant() const { return value; }
+	_FORCE_INLINE_ BitField<T> operator^(const BitField<T> &p_b) const { return BitField<T>(value ^ p_b.value); }
 };
 
 #define TEMPL_MAKE_BITFIELD_TYPE_INFO(m_enum, m_impl)                                                                                            \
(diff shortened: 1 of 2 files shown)


core/variant/type_info.h
```diff
@@ -300,6 +300,7 @@ class BitField {
 	_FORCE_INLINE_ constexpr BitField(T p_value) { value = (int64_t)p_value; }
 	_FORCE_INLINE_ operator int64_t() const { return value; }
 	_FORCE_INLINE_ operator Variant() const { return value; }
+	_FORCE_INLINE_ BitField<T> operator*(const BitField<T> &p_b) const { return BitField<T>(value ^ p_b.value); }
```
AThousandShips, line 303:
```suggestion
	_FORCE_INLINE_ BitField<T> operator^(const BitField<T> &p_b) const { return BitField<T>(value ^ p_b.value); }
```
I'd suggest the xor operator unless unavailable to not be confusing

AThousandShips (approved):
LGTM otherwise, and makes a lot of sense

I'd maybe add more operators for completeness, but this one is specifically relevant

core/variant/type_info.h
AThousandShips: Might not have reached that point as the one use is in the platform code

Also I make these all the time

akien-mga (approved):

akien-mga:
Thanks!

PR #87833, opened 2024-02-01
Clear 2D editor drag preview when drag ends

KoBeWi:
When you drag a resource into 2D editor, its preview will sometimes remain:

https://github.com/godotengine/godot/assets/2223172/a3146b15-7215-4280-8420-2426efb3ef61

The easiest way to reproduce is by pressing Escape to cancel dragging, but it sometimes happens upon dropping in the inspector.

This PR fixes it.


Sauermann:
partial duplicate of #87125

ryevdokimov:
> partial duplicate of #87125

The mentioned PR above solves the same problem the same way, and also applies to 3D as well.

Issue #76425, opened 2023-04-25
Touch input causes FPS drop and stuttering on some iOS devices (Pro Motion?)


djrain:
### Godot version

3.6 beta 1

### System information

iOS 16, GLES3

### Issue description

Discussion started in #32139 but now seems like a separate issue. 

I believe #69200 fixed that issue, but now I'm still seeing stutter and FPS drop from touch input on iPhone 14 Pro Max.
The issue is not reproducible on iPhone 12 or iPhone 6s, so I guess maybe this has to do with Pro Motion (high refresh rate).

Here are some screen recordings comparing a near-empty scene on iPhone 14 Pro Max and iPhone 6s. Godot icons indicate tapping input.

iPhone 6s - performance unaffected by touch input, as it should be:

https://user-images.githubusercontent.com/33777501/234172363-56198b63-d709-4b8d-b4c9-4abd48f8d573.MP4

----
---
iPhone 14 Pro Max (high refresh rate DISABLED in Godot) - notice severe FPS drop:

https://user-images.githubusercontent.com/33777501/234172378-4ac52bd0-435e-485f-b2b1-6e6ecbaa4975.mov

---
---
iPhone 14 Pro Max (high refresh rate enabled in Godot) - again, significant FPS drop:

https://user-images.githubusercontent.com/33777501/234172394-a19cbd91-be93-4950-a3dc-60f20d76a2c0.mov

iPhone 14 is of course many times more powerful than the 6s, so this makes no sense at all. Especially the case where pro motion is not even enabled in Godot and the 14 is struggling.

Our game requires a lot of tapping, and these stutters are unfortunately making the difference between a shippable game, and unplayable garbage :(  so we'd be super grateful for any help on this.

### Steps to reproduce

Export and run MRP main scene on a recent iOS device (may require Pro Motion)

### Minimal reproduction project

[iOSTouchStutter.zip](https://github.com/godotengine/godot/files/11317628/iOSTouchStutter.zip)

djrain:
@akien-mga could be worth a shot, thanks -  but the really odd thing is that the issue persists even if I disable high refresh rate in the phone settings...

tbveralrud:
> That's just a hunch but maybe with the high refresh rate you're getting too many inputs processed and running potentially expensive logic for each event?

I believe the attached project doesn't doesn't have any logic run on input events yet still causes a stutter.

[...]

PR #87836, opened 2024-02-01
macOS: Use autorelease pools

stuartcarnie:
Autorelease pools ensure that autoreleased objects are freed regularly. These changes also reduce the number of allocated objects and memory usage considerably. This is typical of any custom event loop in macOS (and iOS), such as SDLs pump events:

https://github.com/libsdl-org/SDL/blob/68a4bb01e0d549bfa1f2418dd9463f62980eb210/src/video/cocoa/SDL_cocoaevents.m#L564-L569

### DIFF
--- a/platform/macos/display_server_macos.mm
+++ b/platform/macos/display_server_macos.mm
@@ -160,6 +160,7 @@ DisplayServerMacOS::WindowID DisplayServerMacOS::_create_window(WindowMode p_mod
 							  defer:NO];
 		ERR_FAIL_NULL_V_MSG(wd.window_object, INVALID_WINDOW_ID, "Can't create a window");
 		[wd.window_object setWindowID:window_id_counter];
+		[wd.window_object setReleasedWhenClosed:NO];
 
 		wd.window_view = [[GodotContentView alloc] init];
 		ERR_FAIL_NULL_V_MSG(wd.window_view, INVALID_WINDOW_ID, "Can't create a window view");
--- a/platform/macos/os_macos.mm
+++ b/platform/macos/os_macos.mm
@@ -755,21 +755,25 @@ void OS_MacOS::run() {
 		return;
 	}
 
-	main_loop->initialize();
+	@autoreleasepool {
+		main_loop->initialize();
+	}
 
 	bool quit = false;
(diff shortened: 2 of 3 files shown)


platform/macos/os_macos.mm
```diff
@@ -755,21 +755,25 @@
 		return;
 	}
 
-	main_loop->initialize();
+	@autoreleasepool {
+		main_loop->initialize();
+	}
```
stuartcarnie, line 760: Lots of garbage is cleaned up here with a scoped autorelease pool

clayjohn:
CC @RevoluPowered It would be good to see if this helps with the issue you guys are facing

platform/macos/display_server_macos.mm
```diff
@@ -160,6 +160,7 @@
 							  defer:NO];
 		ERR_FAIL_NULL_V_MSG(wd.window_object, INVALID_WINDOW_ID, "Can't create a window");
 		[wd.window_object setWindowID:window_id_counter];
+		[wd.window_object setReleasedWhenClosed:NO];
```
stuartcarnie, line 163: With autorelease pools, objects are getting freed properly, and this results in a crash if not set.

bruvzg (approved):
A dedicated pool for event loop makes sense. Not sure if it will have any impact for `Main::` calls, these should not do much Obj-C related stuff, but I guess won't hurt.

bruvzg:
I'm pretty surprised we did not have a dedicated pool set up in the event-loop. That's probably my fault, and I forgot to re-add it after removing old `NSAutoreleasePool` syntax (still used in 3.x) when we were switching to ARC.

Calinou:
It might be worth checking if an iOS version of this would impact https://github.com/godotengine/godot/issues/76425.

stuartcarnie:
> A dedicated pool for event loop makes sense. Not sure if it will have any impact for `Main::` calls, these should not do much Obj-C related stuff, but I guess won't hurt.

I wouldn't have thought so either, but I checked the number of objects in the autorelease pool before and after those setup calls, and it was not insignificant. If there is no pool, they never get released, so it does help.

PR #61906, opened 2022-06-10
Show a GUI alert when crashing on desktop platforms

Calinou:
This makes it possible to know that Godot has crashed without looking at the console (which may not be visible to the user).

- Change the window title while dumping the backtrace, as this can take a while to finish.
- If file logging is enabled, print path to log file to standard error and the GUI alert.
- Clicking OK will open the log file in the default program associated with `.log` files. Notepad has native associations for `.log` on Windows, so this will work on Windows too (once this PR is ported to Windows).

___

- This closes https://github.com/godotengine/godot-proposals/issues/7917.

### Linux

https://user-images.githubusercontent.com/180032/173110079-8756207b-cb3d-4b31-af61-6ccf0d690129.mp4

### Windows

image

### macOS

Screenshot_20240206_195358 | Screenshot_20240206_195417
-|-

## TODO

- [ ] Consider adding a headless handler for `OS::alert()`, so we don't have to hack around it here. Make this handler print the string with some special formatting around it, but without the blocking behavior (as this behavior is undesired in headless mode).
- [x] Port changes to Windows. Everything seems to work there 🙂
- [x] Port changes to macOS. This is now functional thanks to @bruvzg~~, but https://github.com/godotengine/godot/pull/87842 needs to be merged if we want it to work with `OS.crash()` (the method used for testing this crash handler).~~

_Production edit: closes godotengine/godot-roadmap#5_

PR #87842, opened 2024-02-01
[macOS] Add SIGTRAP to the crash handler.

### LINKED ISSUE #87718: Crash handler on macOS isn't triggered, even in official editor builds
### Tested versions

- Reproducible in: 3.5.3.stable, 4.1.3.stable, 4.2.1.stable, 4.3.dev 9ab5cedef6de32826b9184e960b3880df3888a5a

### System information

Godot v4.2.1.stable - macOS 13.4.1 - Vulkan (Forward+) - integrated Apple M1 - Apple M1 (8 Threads)

### Issue description

The crash handler on macOS isn't triggered, even in official editor builds. This also occurs with self-compiled builds not put in an `.app` bundle.

In comparison, the crash handler works on Linux and Windows (when compiled in MSVC and in debug builds only).

### Steps to reproduce

Run the editor from a terminal. Create a project with the following script attached on the root node, then run it:

```gdscript
func _ready():
    OS.crash("message")
```

Instead of seeing a crash backtrace (even if mangled due to a lack of debug symbols), you'll only see the following:

image

[test_crash_handler.zip](https://github.com/godotengine/godot/files/14090186/test_crash_handler.zip)

bruvzg:
Fixes https://github.com/godotengine/godot/issues/87718

It's unnecessary for other platforms and not triggered by real crashes, but useful to get trace from `OS.crash` call.

### DIFF
--- a/platform/macos/crash_handler_macos.mm
+++ b/platform/macos/crash_handler_macos.mm
@@ -75,6 +75,7 @@ static void handle_crash(int sig) {
 	signal(SIGSEGV, SIG_DFL);
 	signal(SIGFPE, SIG_DFL);
 	signal(SIGILL, SIG_DFL);
+	signal(SIGTRAP, SIG_DFL);
 
 	if (OS::get_singleton() == nullptr) {
 		abort();
@@ -193,6 +194,7 @@ void CrashHandler::disable() {
 	signal(SIGSEGV, SIG_DFL);
 	signal(SIGFPE, SIG_DFL);
 	signal(SIGILL, SIG_DFL);
+	signal(SIGTRAP, SIG_DFL);
 #endif
 
 	disabled = true;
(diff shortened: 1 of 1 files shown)


Calinou (commented):
Tested locally, it mostly works as expected.

I can see https://github.com/godotengine/godot/pull/61906 partially working: window title is changed and crash backtrace is printed to stdout (and log file), but the window never exits and I don't see the alert appearing like I do on Windows and Linux.

That said, this may just be due to `OS::alert()` blocking the handler in a different way than on Windows and Linux. I may have to skip `OS::alert()` on macOS and just open the log file directly with the default program.

image

bruvzg:
> That said, this may just be due to OS::alert() blocking the handler in a different way than on Windows and Linux. I may have to skip OS::alert() on macOS and just open the log file directly with the default program.

In case of macOS, `OS::alert` require event loop to run, otherwise it won't show (or will be unresponsive).

Calinou (approved):
This is fully functional with the latest revision of https://github.com/godotengine/godot/pull/61906, so this should be good to go.

akien-mga:
Thanks!

PR #87848, opened 2024-02-02
Fix `godot.natvis` after CowData 64-bit promotion

nikitalita:
Since the natvis can't use the class functions to determine size, it has to get the size manually via casting the cowdata and indexing. CowData was recently promoted to 64-bits (#86730), which broke this. So, I have fixed that by changing the casts to `long long` instead of `int`.


fire (approved):
I think this is correct.

akien-mga:
Thanks!

PR #87850, opened 2024-02-02
Proof of concept fix for Async GPU Readbacks

granitrocky:
I have an implementation and a basic demo to show how to achieve Asynchronous Compute Shader execution with Vulkan Timeline Semaphores.

Currently Vulkan specific just to get the ball rolling

And here's a repurposed ray tracing project that demonstrates how my changes can avoid locking up the CPU while waiting for Compute Shaders to finish execution:
https://github.com/granitrocky/Raytracing_Godot4

* *Bugsquad edit, related: https://github.com/godotengine/godot-proposals/issues/7886*

### DIFF
--- a/drivers/vulkan/vulkan_context.cpp
+++ b/drivers/vulkan/vulkan_context.cpp
@@ -42,6 +42,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <cstdint>
 
 #define ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
 #define APP_SHORT_NAME "GodotEngine"
@@ -1794,6 +1795,22 @@ Error VulkanContext::_create_semaphores() {
 		/*pNext*/ nullptr,
 		/*flags*/ 0,
 	};
+	// Create compute semaphore that can be polled to determine whether
+	// a given compute block has finished.
+    // Every time a compute call finishes, we'll increment the value
+    // of this semaphore by one, so we can check the status
+    VkSemaphoreTypeCreateInfo timelineCreateInfo = {
+        /*sType*/ VK_STRUCTURE_TYPE_SEMAPHORE_TYPE_CREATE_INFO,
+        /*pNext*/ nullptr,
+        /*semaphoreType*/ VK_SEMAPHORE_TYPE_TIMELINE,
+        /*initialValue*/ 0,
+    };
+
+	VkSemaphoreCreateInfo computeSemaphoreCreateInfo = {
+		/*sType*/ VK_STRUCTURE_TYPE_SEMAPHORE_CREATE_INFO,
+		/*pNext*/ &timelineCreateInfo,
+		/*flags*/ 0,
+	};
 
 	// Create fences that we can use to throttle if we get too far
 	// ahead of the image presents.
@@ -1815,6 +1832,9 @@ Error VulkanContext::_create_semaphores() {
 		}
 	}
 	frame_index = 0;
+
+    err = vkCreateSemaphore(device, &computeSemaphoreCreateInfo, nullptr, &compute_semaphore[0]);
+		ERR_FAIL_COND_V(err, ERR_CANT_CREATE);
 
 	// Get Memory information and properties.
 	vkGetPhysicalDeviceMemoryProperties(gpu, &memory_properties);
@@ -2781,16 +2801,26 @@ void VulkanContext::local_device_push_command_buffers(RID p_local_device, const 
 	LocalDevice *ld = local_device_owner.get_or_null(p_local_device);
 	ERR_FAIL_COND(ld->waiting);
 
+    compute_semaphore_signal_value += 1;
+
+    VkTimelineSemaphoreSubmitInfo signal_info;
+    signal_info.sType = VK_STRUCTURE_TYPE_TIMELINE_SEMAPHORE_SUBMIT_INFO;
+    signal_info.pNext = nullptr;
+    signal_info.waitSemaphoreValueCount = 0;
+    signal_info.signalSemaphoreValueCount = 1;
+    signal_info.pSignalSemaphoreValues = &compute_semaphore_signal_value;
+
 	VkSubmitInfo submit_info;
 	submit_info.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
-	submit_info.pNext = nullptr;
+	submit_info.pNext = &signal_info;
 	submit_info.pWaitDstStageMask = nullptr;
 	submit_info.waitSemaphoreCount = 0;
 	submit_info.pWaitSemaphores = nullptr;
 	submit_info.commandBufferCount = p_count;
 	submit_info.pCommandBuffers = (const VkCommandBuffer *)p_buffers;
-	submit_info.signalSemaphoreCount = 0;
-	submit_info.pSignalSemaphores = nullptr;
+	submit_info.signalSemaphoreCount = 1;
+	submit_info.pSignalSemaphores = &compute_semaphore[0];
+
 
 	VkResult err = vkQueueSubmit(ld->queue, 1, &submit_info, VK_NULL_HANDLE);
 	if (err == VK_ERROR_OUT_OF_HOST_MEMORY) {
@@ -2813,7 +2843,17 @@ void VulkanContext::local_device_sync(RID p_local_device) {
 
 	vkDeviceWaitIdle(ld->device);
 	ld->waiting = false;
+
 }
+
+bool VulkanContext::local_device_check_status(RID p_local_device) {
+	LocalDevice *ld = local_device_owner.get_or_null(p_local_device);
+	ERR_FAIL_COND_V(!ld->waiting, false);
+    uint64_t out = 0;
+    vkGetSemaphoreCounterValue(ld->device, compute_semaphore[0], &out);
+    return out >= compute_semaphore_signal_value;
+}
+
 
 void VulkanContext::local_device_free(RID p_local_device) {
 	LocalDevice *ld = local_device_owner.get_or_null(p_local_device);
@@ -2901,6 +2941,7 @@ VulkanContext::~VulkanContext() {
 				vkDestroySemaphore(device, image_ownership_semaphores[i], nullptr);
 			}
 		}
+        vkDestroySemaphore(device, compute_semaphore[0], nullptr);
 		if (inst_initialized && is_instance_extension_enabled(VK_EXT_DEBUG_UTILS_EXTENSION_NAME)) {
 			DestroyDebugUtilsMessengerEXT(inst, dbg_messenger, nullptr);
 		}
--- a/drivers/vulkan/vulkan_context.h
+++ b/drivers/vulkan/vulkan_context.h
@@ -40,6 +40,7 @@
 #include "rendering_device_driver_vulkan.h"
 #include "servers/display_server.h"
 #include "servers/rendering/renderer_rd/api_context_rd.h"
+#include <cstdint>
 
 #ifdef USE_VOLK
 #include <volk.h>
@@ -133,6 +134,8 @@ private:
 	VkFormat format;
 	VkSemaphore draw_complete_semaphores[FRAME_LAG];
 	VkSemaphore image_ownership_semaphores[FRAME_LAG];
+	VkSemaphore compute_semaphore[1];
+    uint64_t compute_semaphore_signal_value = 0;
 	int frame_index = 0;
 	VkFence fences[FRAME_LAG];
 	VkPhysicalDeviceMemoryProperties memory_properties;
@@ -313,6 +316,7 @@ public:
 	virtual RID local_device_create() override final;
 	virtual void local_device_push_command_buffers(RID p_local_device, const RDD::CommandBufferID *p_buffers, int p_count) override final;
 	virtual void local_device_sync(RID p_local_device) override final;
+    virtual bool local_device_check_status(RID p_local_device) override final;
 	virtual void local_device_free(RID p_local_device) override final;
 
 	VkFormat get_screen_format() const;
--- a/servers/rendering/renderer_rd/api_context_rd.h
+++ b/servers/rendering/renderer_rd/api_context_rd.h
@@ -54,6 +54,7 @@ public:
 	virtual RID local_device_create() = 0;
 	virtual void local_device_push_command_buffers(RID p_local_device, const RDD::CommandBufferID *p_buffers, int p_count) = 0;
 	virtual void local_device_sync(RID p_local_device) = 0;
+    virtual bool local_device_check_status(RID p_local_device) = 0;
 	virtual void local_device_free(RID p_local_device) = 0;
 
 	virtual void set_setup_buffer(RDD::CommandBufferID p_command_buffer) = 0;
--- a/servers/rendering/rendering_device.cpp
+++ b/servers/rendering/rendering_device.cpp
@@ -29,6 +29,7 @@
 /**************************************************************************/
 
 #include "rendering_device.h"
+#include "core/os/thread_safe.h"
 #include "rendering_device.compat.inc"
 
 #include "rendering_device_binds.h"
@@ -4659,6 +4660,12 @@ void RenderingDevice::sync() {
 	local_device_processing = false;
 }
 
+bool RenderingDevice::check_status(){
+	_THREAD_SAFE_METHOD_
+
+        return context->local_device_check_status(local_device);
+}
+
 void RenderingDevice::_free_pending_resources(int p_frame) {
 	// Free in dependency usage order, so nothing weird happens.
 	// Pipelines.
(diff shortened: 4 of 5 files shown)


drivers/vulkan/vulkan_context.cpp
```diff
@@ -42,6 +42,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <cstdint>
```
AThousandShips, line 45:
```suggestion
```
Already included (even in the header, only include what's necessary)

drivers/vulkan/vulkan_context.h
```diff
@@ -40,6 +40,7 @@
 #include "rendering_device_driver_vulkan.h"
 #include "servers/display_server.h"
 #include "servers/rendering/renderer_rd/api_context_rd.h"
+#include <cstdint>
```
AThousandShips, line 43:
```suggestion
```

servers/rendering/rendering_device.cpp
```diff
@@ -29,6 +29,7 @@
 /**************************************************************************/
 
 #include "rendering_device.h"
+#include "core/os/thread_safe.h"
```
AThousandShips, line 32:
```suggestion
```

servers/rendering/rendering_device.cpp
```diff
@@ -4659,6 +4660,12 @@ void RenderingDevice::sync() {
 	local_device_processing = false;
 }
 
+bool RenderingDevice::check_status(){
+	_THREAD_SAFE_METHOD_
+
+        return context->local_device_check_status(local_device);
```
AThousandShips, line 4666:
```suggestion
	return context->local_device_check_status(local_device);
```

servers/rendering/rendering_device.h
```diff
@@ -1333,6 +1333,7 @@ class RenderingDevice : public RenderingDeviceCommons {
 
 	void submit();
 	void sync();
+    bool check_status();
```
AThousandShips, line 1336:
```suggestion
	bool check_status();
```

servers/rendering/renderer_rd/api_context_rd.h
```diff
@@ -54,6 +54,7 @@ class ApiContextRD {
 	virtual RID local_device_create() = 0;
 	virtual void local_device_push_command_buffers(RID p_local_device, const RDD::CommandBufferID *p_buffers, int p_count) = 0;
 	virtual void local_device_sync(RID p_local_device) = 0;
+    virtual bool local_device_check_status(RID p_local_device) = 0;
```
AThousandShips, line 57:
```suggestion
	virtual bool local_device_check_status(RID p_local_device) = 0;
```

drivers/vulkan/vulkan_context.h
```diff
@@ -313,6 +316,7 @@ class VulkanContext : public ApiContextRD {
 	virtual RID local_device_create() override final;
 	virtual void local_device_push_command_buffers(RID p_local_device, const RDD::CommandBufferID *p_buffers, int p_count) override final;
 	virtual void local_device_sync(RID p_local_device) override final;
+    virtual bool local_device_check_status(RID p_local_device) override final;
```
AThousandShips, line 319:
```suggestion
	virtual bool local_device_check_status(RID p_local_device) override final;
```

drivers/vulkan/vulkan_context.cpp
```diff
@@ -2901,6 +2941,7 @@ VulkanContext::~VulkanContext() {
 				vkDestroySemaphore(device, image_ownership_semaphores[i], nullptr);
 			}
 		}
+        vkDestroySemaphore(device, compute_semaphore[0], nullptr);
```
AThousandShips, line 2944:
```suggestion
		vkDestroySemaphore(device, compute_semaphore[0], nullptr);
```

drivers/vulkan/vulkan_context.cpp
```diff
@@ -2781,16 +2801,26 @@ void VulkanContext::local_device_push_command_buffers(RID p_local_device, const
...
 	ERR_FAIL_COND(ld->waiting);
 
+    compute_semaphore_signal_value += 1;
+
+    VkTimelineSemaphoreSubmitInfo signal_info;
+    signal_info.sType = VK_STRUCTURE_TYPE_TIMELINE_SEMAPHORE_SUBMIT_INFO;
+    signal_info.pNext = nullptr;
+    signal_info.waitSemaphoreValueCount = 0;
+    signal_info.signalSemaphoreValueCount = 1;
+    signal_info.pSignalSemaphoreValues = &compute_semaphore_signal_value;
```
AThousandShips, line 2811:
```suggestion
	compute_semaphore_signal_value += 1;

	VkTimelineSemaphoreSubmitInfo signal_info;
	signal_info.sType = VK_STRUCTURE_TYPE_TIMELINE_SEMAPHORE_SUBMIT_INFO;
	signal_info.pNext = nullptr;
	signal_info.waitSemaphoreValueCount = 0;
	signal_info.signalSemaphoreValueCount = 1;
	signal_info.pSignalSemaphoreValues = &compute_semaphore_signal_value;
```

drivers/vulkan/vulkan_context.cpp
```diff
@@ -2813,8 +2843,18 @@ void VulkanContext::local_device_sync(RID p_local_device) {
...
 }
 
+bool VulkanContext::local_device_check_status(RID p_local_device) {
+	LocalDevice *ld = local_device_owner.get_or_null(p_local_device);
+	ERR_FAIL_COND_V(!ld->waiting, false);
+    uint64_t out = 0;
+    vkGetSemaphoreCounterValue(ld->device, compute_semaphore[0], &out);
+    return out >= compute_semaphore_signal_value;
+}
+
```
AThousandShips, line 2856:
```suggestion
	uint64_t out = 0;
	vkGetSemaphoreCounterValue(ld->device, compute_semaphore[0], &out);
	return out >= compute_semaphore_signal_value;
}
```

drivers/vulkan/vulkan_context.h
```diff
@@ -133,6 +134,8 @@ class VulkanContext : public ApiContextRD {
 	VkFormat format;
 	VkSemaphore draw_complete_semaphores[FRAME_LAG];
 	VkSemaphore image_ownership_semaphores[FRAME_LAG];
+	VkSemaphore compute_semaphore[1];
+    uint64_t compute_semaphore_signal_value = 0;
```
AThousandShips, line 138:
```suggestion
	uint64_t compute_semaphore_signal_value = 0;
```

drivers/vulkan/vulkan_context.cpp
```diff
@@ -2781,16 +2801,26 @@ void VulkanContext::local_device_push_command_buffers(RID p_local_device, const
...
 	submit_info.pWaitDstStageMask = nullptr;
 	submit_info.waitSemaphoreCount = 0;
 	submit_info.pWaitSemaphores = nullptr;
 	submit_info.commandBufferCount = p_count;
 	submit_info.pCommandBuffers = (const VkCommandBuffer *)p_buffers;
-	submit_info.signalSemaphoreCount = 0;
-	submit_info.pSignalSemaphores = nullptr;
+	submit_info.signalSemaphoreCount = 1;
+	submit_info.pSignalSemaphores = &compute_semaphore[0];
+
```
AThousandShips, line 2823:
```suggestion
```

drivers/vulkan/vulkan_context.cpp
```diff
@@ -1816,6 +1833,9 @@ Error VulkanContext::_create_semaphores() {
 	}
 	frame_index = 0;
 
+    err = vkCreateSemaphore(device, &computeSemaphoreCreateInfo, nullptr, &compute_semaphore[0]);
+		ERR_FAIL_COND_V(err, ERR_CANT_CREATE);
```
AThousandShips, line 1837:
```suggestion
	err = vkCreateSemaphore(device, &computeSemaphoreCreateInfo, nullptr, &compute_semaphore[0]);
	ERR_FAIL_COND_V(err, ERR_CANT_CREATE);
```

drivers/vulkan/vulkan_context.cpp
```diff
@@ -1794,6 +1795,22 @@ Error VulkanContext::_create_semaphores() {
...
+	// Create compute semaphore that can be polled to determine whether
+	// a given compute block has finished.
+    // Every time a compute call finishes, we'll increment the value
+    // of this semaphore by one, so we can check the status
+    VkSemaphoreTypeCreateInfo timelineCreateInfo = {
+        /*sType*/ VK_STRUCTURE_TYPE_SEMAPHORE_TYPE_CREATE_INFO,
+        /*pNext*/ nullptr,
+        /*semaphoreType*/ VK_SEMAPHORE_TYPE_TIMELINE,
+        /*initialValue*/ 0,
+    };
```
AThousandShips, line 1807:
```suggestion
	// Every time a compute call finishes, we'll increment the value
	// of this semaphore by one, so we can check the status.
	VkSemaphoreTypeCreateInfo timelineCreateInfo = {
		/*sType*/ VK_STRUCTURE_TYPE_SEMAPHORE_TYPE_CREATE_INFO,
		/*pNext*/ nullptr,
		/*semaphoreType*/ VK_SEMAPHORE_TYPE_TIMELINE,
		/*initialValue*/ 0,
	};
```

drivers/vulkan/vulkan_context.cpp
```diff
@@ -1794,6 +1795,22 @@ Error VulkanContext::_create_semaphores() {
...
+
+	VkSemaphoreCreateInfo computeSemaphoreCreateInfo = {
+		/*sType*/ VK_STRUCTURE_TYPE_SEMAPHORE_CREATE_INFO,
+		/*pNext*/ &timelineCreateInfo,
```
AThousandShips, line 1811:
```suggestion
	VkSemaphoreCreateInfo compute_semaphore_create_info = {
		/*sType*/ VK_STRUCTURE_TYPE_SEMAPHORE_CREATE_INFO,
		/*pNext*/ &timeline_create_info,
```

servers/rendering/rendering_device.cpp
```diff
@@ -4659,6 +4660,12 @@ void RenderingDevice::sync() {
 	local_device_processing = false;
 }
 
+bool RenderingDevice::check_status(){
```
AThousandShips, line 4663:
```suggestion
bool RenderingDevice::check_status() {
```

drivers/vulkan/vulkan_context.cpp
```diff
@@ -2813,8 +2843,18 @@ void VulkanContext::local_device_sync(RID p_local_device) {
 
 	vkDeviceWaitIdle(ld->device);
 	ld->waiting = false;
+
```
AThousandShips, line 2846:
```suggestion
```

DarioSamo:
This is an interesting PR but we might run into conflicts since we intend to merge https://github.com/godotengine/godot/pull/87340 and that changes the entire design around RD being able to create semaphores and fences freely.

That said I find the use of timeline semaphores interesting. These are not guaranteed to be supported so you must check for support I'm afraid, so they can't just be created without checking for that. But judging by the way they're being used in the PR, I get the feeling they might not be strictly necessary.

If you feel the use of timeline semaphores will be required in the future we could probably take that into account in the redesign of the PR I mentioned (as they're guaranteed in D3D12 but in Vulkan they must be queried for support). That's why the RHI goes for the lowest common denominator. Do you think it might be possible to achieve the effect you're looking for without using them?

granitrocky:
I tried to find a way of doing this with a simple boolean semaphore, but from what I could read in the API, there's no non-blocking way of reading a boolean semaphore value.

If there is, a lot of this complexity goes away, so I'd be happy for that.

granitrocky:
@DarioSamo 

I'm going to take some time and see if I can get this to sit on top of https://github.com/godotengine/godot/pull/87340 

Hopefully they play nicely together.

DarioSamo:
> I tried to find a way of doing this with a simple boolean semaphore, but from what I could read in the API, there's no non-blocking way of reading a boolean semaphore value.
> 
> If there is, a lot of this complexity goes away, so I'd be happy for that.

[Fences actually allow a non-blocking (although not free) way to check their status](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetFenceStatus.html), so if you only want to check if a command buffer in particular is finished, that could be a pretty cheap way to go about it. Considering the fact a local device does not allow more than one simultaneous command buffer, if you have to wait for a fence you can assume all work assigned to the previous buffer must be complete already, so you wouldn't need a counter for it either.

Keep in mind this is not me being strict for the sake of it, it's just that since Godot aims to be as compatible as possible, you might find some resistance in using features newer than 1.0 as they might not be supported in the devices it aims to support. Therefore anything that needs a particular version supported must be explicitly checked and a suitable alternative path must be taken if support is not present (if possible). For something as core as this, if we can find a way to work around it the better.

> I'm going to take some time and see if I can get this to sit on top of https://github.com/godotengine/godot/pull/87340
> Hopefully they play nicely together.
Thanks! My suggestion mostly comes from the fact if it sits on top of the RHI, you won't need to re-implement it for all backends (since D3D12 is already in master and Metal is also coming at a later point).

granitrocky:
@DarioSamo I totally understand. This was just a first pass attempt to see what gets the wheels spinning in people's heads since I almost certainly don't have the best approach.

Here's a bit of an architecture question. Are compute shaders run in the same command buffer as the render/swap chains? Because the use case for this PR is a compute shader that runs over multiple frames needing to be synchronized only after its work has completed.

Calinou:
> Keep in mind this is not me being strict for the sake of it, it's just that since Godot aims to be as compatible as possible, you might find some resistance in using features newer than 1.0 as they might not be available in the devices it aims to support.

To clarify, it's more of an issue of vendor support (check gpuinfo.org) than Vulkan version. Pretty much all devices where Vulkan *is worth using* support the 1.2 specification, but they don't always support the extensions you wish to use.

DarioSamo:
> Here's a bit of an architecture question. Are compute shaders run in the same command buffer as the render/swap chains? Because the use case for this PR is a compute shader that runs over multiple frames needing to be synchronized only after its work has completed.

That is currently the case. There's no architecture for allowing compute lists to run in parallel at the moment as resource tracking and the render graph are only guaranteed to work correctly when called from the main thread. Any work queued up from those is run in the command buffer used for rendering and is expected to finish before presenting to the swap chain.

The use case you describe can be implemented very easily on top of the PR I posted (and in fact https://github.com/godotengine/godot/pull/87590 takes an approach like that to implement asynchronous transfer queues that run in parallel to the main ones). I can't say however what the ideal interface for accessing this would be yet (we have a lot of thread restrictions to enforce and resource tracking would need to be kept local to the context).

granitrocky:
@DarioSamo I think I follow. The actual dispatch of the compute payload would need to be done in a separate command buffer altogether, which your PR allows.

So since the use case here is that we don't actually want to transfer anything, just run a check that the compute workload is done, that should be able to be done atomically. In the linked project, you can see that I'm just checking every frame whether the workload is done, and if so, call `RD.sync()`

Maybe my use case is too simplistic to even really offer many benefits, and my efforts may be better spent working with PR https://github.com/godotengine/godot/pull/87590

DarioSamo:
> So since the use case here is that we don't actually want to transfer anything, just run a check that the compute workload is done, that should be able to be done atomically. In the linked project, you can see that I'm just checking every frame whether the workload is done, and if so, call `RD.sync()`

Your linked project uses local device, so it's a bit of a different story. On local devices, the command buffers are local to it and are not shared with the Rendering Device used for drawing graphics at all.

Given your particular use case now that I see it, I don't think you'd need to use timeline semaphores at all. Your current logic could be implemented if you call vkGetFenceStatus on the vkFence object for a particular 'frame'. In the case of local devices, frames only ever holds one frame, so you'd just be checking if the frame has finished executing without doing an active wait on it.

My other suggestions mostly apply if you wish to look into a solution that uses the existing RenderingDevice without creating a local one, as that comes with extra cost and limitations for sharing resources between them.

granitrocky:
Ah okay. So "local device" here means that I'm creating my own local `RenderingDevice`? As opposed to using the singleton version?

DarioSamo:
> Ah okay. So "local device" here means that I'm creating my own local RenderingDevice? As opposed to using the singleton version?

Yup. Therefore all resources you create there can't be shared with the Singleton version. That may or may not be useful to you (we use this on the offline lightmapper for example).

granitrocky:
Okay cool. I appreciate your notes. I'll try and see whether https://github.com/godotengine/godot/pull/87590 exposes what I'm looking for and whether I have anything to contribute.

granitrocky:
@QuinnLeavitt Unfortunately I haven't had time to see how much work it would be to get this to work with https://github.com/godotengine/godot/pull/87590 .

I would love to take another crack at this when I have some time.

Issue #53295, opened 2021-10-01
Invalid cross includes that do not respect area encapsulation rules


akien-mga:
### Godot version

4.0.dev (06c1b40b)

### System information

Any

### Issue description

The engine codebase is compartmentalized in areas/folders which have a defined dependency order to prevent both cyclic dependencies and general bad design patterns.

To ensure this encapsulation, we need to make sure that we don't include headers from high level areas in low level areas. For example `core` is the lowest level area, so it shouldn't include headers from any other folder (aside from `thirdparty`, which is independent from Godot code).

We should add CI checks to prevent adding such invalid includes, but first we need to fix the problems we already have in the codebase (#29730 is related, this issue is the same but for the whole codebase/dependency path).

The dependency order should follow the build order as defined here:
https://github.com/godotengine/godot/blob/06c1b40b84645b4460dd49d95694a7779d636f92/SConstruct#L724-L737

- core
- servers
- scene
- drivers
- editor
- modules
- tests
- main
- platform

I'll list invalid includes in comments for each folder/area.

### Steps to reproduce

n/a

akien-mga:
# Servers

```
$ rg '#include "scene/'  
audio_server.cpp
42:#include "scene/resources/audio_stream_sample.h"

display_server.cpp
34:#include "scene/resources/texture.h"

navigation_server_3d.h
40:#include "scene/3d/navigation_region_3d.h"

text_server.cpp
32:#include "scene/main/canvas_item.h"

physics_3d/soft_body_3d_sw.h
43:#include "scene/resources/mesh.h"

rendering/rasterizer_dummy.h
37:#include "scene/resources/mesh.h"

navigation_server_2d.h
40:#include "scene/2d/navigation_region_2d.h"

text_server.h
38:#include "scene/resources/texture.h"

xr/xr_positional_tracker.h
35:#include "scene/resources/mesh.h"

audio/effects/audio_effect_record.h
39:#include "scene/resources/audio_stream_sample.h"
```

```
$ rg '#include "editor/'
audio/effects/audio_effect_record.h
38:#include "editor/import/resource_importer_wav.h"
```

akien-mga:
# Scene

```
$ rg '#include "editor/' 
gui/rich_text_label.cpp
45:#include "editor/editor_scale.h"

gui/tree.cpp
45:#include "editor/editor_scale.h"

gui/graph_edit.cpp
40:#include "editor/editor_scale.h"

gui/gradient_edit.cpp
[...]

PR #87858, opened 2024-02-02
Fix `rpc` call from name based `Callable`s

### LINKED ISSUE #87841: Callable.rpc() on RPC function throws an error "Invalid call. Nonexistent function 'rpc' in base 'Callable'.
### Tested versions

- Reproducible in: Godot v4.2.1 stable win64

### System information

Windows 10 - Godot v4.2.1.stable

### Issue description

Creating a `Callable` and calling the `rpc` function throws an error stating that the call is invalid, but calling the `rpc` function on the function and not the `Callable` works.

### Steps to reproduce

Create a new GDScript in a blank project, attach it to a node, add the following to the script:
```gdscript
@rpc
func hello():
	print("RPC fn has been called!")

func _ready():
	hello.rpc() # OK
	var fn = Callable(self, "hello")
	fn.rpc() # ERR
```

[bugreport1.zip](https://github.com/godotengine/godot/files/14131316/bugreport1.zip)

AThousandShips:
Makes `rpc` calls work for the `Callable(obj, method)` form `Callable`s, works like the `GDScriptRPCCallable` call (which also does no special treatment for the error return, I'll leave it as such and leave any improvements or changes to that side to future improvements to both)

* Fixes: https://github.com/godotengine/godot/issues/87841

### DIFF
--- a/core/variant/callable.cpp
+++ b/core/variant/callable.cpp
@@ -92,10 +92,31 @@ Error Callable::rpcp(int p_id, const Variant **p_arguments, int p_argcount, Call
 		r_call_error.expected = 0;
 		return ERR_UNCONFIGURED;
 	} else if (!is_custom()) {
-		r_call_error.error = CallError::CALL_ERROR_INVALID_METHOD;
-		r_call_error.argument = 0;
-		r_call_error.expected = 0;
-		return ERR_UNCONFIGURED;
+		Object *obj = ObjectDB::get_instance(ObjectID(object));
+#ifdef DEBUG_ENABLED
+		if (!obj || !obj->is_class("Node")) {
+			r_call_error.error = CallError::CALL_ERROR_INSTANCE_IS_NULL;
+			r_call_error.argument = 0;
+			r_call_error.expected = 0;
+			return ERR_UNCONFIGURED;
+		}
+#endif
+
+		int argcount = p_argcount + 2;
+		const Variant **argptrs = (const Variant **)alloca(sizeof(Variant *) * argcount);
+		const Variant args[2] = { p_id, method };
+
+		argptrs[0] = &args[0];
+		argptrs[1] = &args[1];
+		for (int i = 0; i < p_argcount; ++i) {
+			argptrs[i + 2] = p_arguments[i];
+		}
+
+		CallError tmp;
+		Error err = (Error)obj->callp(SNAME("rpc_id"), argptrs, argcount, tmp).operator int64_t();
+
+		r_call_error.error = Callable::CallError::CALL_OK;
+		return err;
 	} else {
 		return custom->rpc(p_id, p_arguments, p_argcount, r_call_error);
 	}


core/variant/callable.cpp
```diff
@@ -91,10 +92,17 @@ Error Callable::rpcp(int p_id, const Variant **p_arguments, int p_argcount, Call
 		r_call_error.expected = 0;
 		return ERR_UNCONFIGURED;
 	} else if (!is_custom()) {
-		r_call_error.error = CallError::CALL_ERROR_INVALID_METHOD;
-		r_call_error.argument = 0;
-		r_call_error.expected = 0;
-		return ERR_UNCONFIGURED;
+		Node *node = Object::cast_to<Node>(ObjectDB::get_instance(ObjectID(object)));
+#ifdef DEBUG_ENABLED
```
AThousandShips, line 96: Using the same debug only check here

core/variant/callable.cpp
```diff
@@ -35,6 +35,7 @@
 #include "core/object/script_language.h"
 #include "core/variant/callable_bind.h"
 #include "core/variant/variant_callable.h"
+#include "scene/main/node.h"
```
akien-mga, line 38: This is problematic, `core` shouldn't depend on `scene`.
There are a couple exceptions currently that we need to fix, but in theory `core` shouldn't even know about Node.

AThousandShips: That's true, but RPC calls simply require `Node`, so unsure how to solve otherwise

akien-mga: See #53295 for context.

akien-mga: You can likely work it around with strings:
```cpp
Object *obj = ObjectDB::get_instance(ObjectID(object));
#ifdef DEBUG_ENABLED
if (!obj || !obj->is_class("Node")) {
	r_call_error.error = CallError::CALL_ERROR_INSTANCE_IS_NULL;
	r_call_error.argument = 0;
	r_call_error.expected = 0;
	return ERR_UNCONFIGURED;
}
#endif
r_call_error.error = CallError::CALL_OK;
return obj->call("rpcp", p_id, method, p_arguments, p_argcount);
```

AThousandShips: Sounds good, will look at that!

core/variant/callable.cpp
```diff
@@ -92,10 +92,24 @@ Error Callable::rpcp(int p_id, const Variant **p_arguments, int p_argcount, Call
...
+#endif
+		LocalVector<const Variant *> argptrs;
+		argptrs.resize(p_argcount + 2);
+		Variant args[2] = { p_id, method };
+		argptrs[0] = &args[0];
+		argptrs[1] = &args[1];
+		for (int i = 0; i < p_argcount; ++i) {
+			argptrs[i + 2] = p_arguments[i];
+		}
+		return (Error)obj->callp("rpc_id", argptrs.ptr(), p_argcount + 2, r_call_error).operator int64_t();
```
AThousandShips, line 112:
```suggestion
		int argcount = p_argcount + 2;
		const Variant **argptrs = (const Variant **)alloca(sizeof(Variant *) * argcount);
		Variant args[2] = { p_id, method };
		argptrs[0] = &args[0];
		argptrs[1] = &args[1];
		for (int i = 0; i < p_argcount; ++i) {
			argptrs[i + 2] = p_arguments[i];
		}
		return (Error)obj->callp("rpc_id", argptrs, argcount, r_call_error).operator int64_t();
```
Alternatively

core/variant/callable.cpp
```diff
@@ -92,10 +92,31 @@ Error Callable::rpcp(int p_id, const Variant **p_arguments, int p_argcount, Call
...
+		const Variant args[2] = { p_id, method };
+
+		argptrs[0] = &args[0];
+		argptrs[1] = &args[1];
+		for (int i = 0; i < p_argcount; ++i) {
+			argptrs[i + 2] = p_arguments[i];
+		}
+
+		CallError tmp;
+		Error err = (Error)obj->callp("rpc_id", argptrs, argcount, tmp).operator int64_t();
```
KoBeWi, line 116:
```suggestion
		Error err = (Error)obj->callp(SNAME("rpc_id"), argptrs, argcount, tmp).operator int64_t();
```

KoBeWi (approved):
This fixes the error, but not sure how to confirm whether the method is actually called.
The code looks fine though.

Faless (approved):
This looks great to me!

I remember @reduz wanted this to be implemented at the language level using custom callables, but I think it make sense to also have the option of using `Callable(obj, method)` as an alternative if the language doesn't implement it.

akien-mga:
Thanks!

AThousandShips:
Thank *you*!

PR #87860, opened 2024-02-02
Don't round `loop_offset` on import + Use double for `loop_offset` consistently

### LINKED ISSUE #87427: AudioStreamImportSettingsDialog: loop offset field only allows values with up to three decimals, causing a loop point imprecision that sometimes leads to an audible pop when looping

MJacred:
Not sure if this PR should close https://github.com/godotengine/godot/issues/87427, but it would cover their requirement of a floating point value with the precision of 12 decimal places, as Godot's `float` provides for `double` `14 reliable decimal digits of precision`. **Except the value will be rounded/cut off again on saving to a file?**

I removed the `set_step` call to prevent rounding, as we don't know how the user will use their given precision.

Side note: ogg vorbis was already using `double` consistently.

### DIFF
--- a/editor/import/audio_stream_import_settings.cpp
+++ b/editor/import/audio_stream_import_settings.cpp
@@ -437,7 +437,7 @@ void AudioStreamImportSettingsDialog::edit(const String &p_path, const String &p
 			beats_edit->set_value(beats);
 			beats_enabled->set_pressed(beats > 0);
 			loop->set_pressed(config_file->get_value("params", "loop", false));
-			loop_offset->set_value(config_file->get_value("params", "loop_offset", 0));
+			loop_offset->set_value(double(config_file->get_value("params", "loop_offset", 0)));
 			bar_beats_edit->set_value(config_file->get_value("params", "bar_beats", 4));
 
 			List<String> keys;
@@ -517,7 +517,7 @@ void AudioStreamImportSettingsDialog::_settings_changed() {
 
 void AudioStreamImportSettingsDialog::_reimport() {
 	params["loop"] = loop->is_pressed();
-	params["loop_offset"] = loop_offset->get_value();
+	params["loop_offset"] = double(loop_offset->get_value());
 	params["bpm"] = bpm_enabled->is_pressed() ? double(bpm_edit->get_value()) : double(0);
 	params["beat_count"] = (bpm_enabled->is_pressed() && beats_enabled->is_pressed()) ? int(beats_edit->get_value()) : int(0);
 	params["bar_beats"] = (bpm_enabled->is_pressed()) ? int(bar_beats_edit->get_value()) : int(4);
@@ -543,7 +543,7 @@ AudioStreamImportSettingsDialog::AudioStreamImportSettingsDialog() {
 	loop_hb->add_child(memnew(Label(TTR("Offset:"))));
 	loop_offset = memnew(SpinBox);
 	loop_offset->set_max(10000);
-	loop_offset->set_step(0.001);
+	loop_offset->set_disallow_step_rounding(true);
 	loop_offset->set_suffix("sec");
 	loop_offset->set_tooltip_text(TTR("Loop offset (from beginning). Note that if BPM is set, this setting will be ignored."));
 	loop_offset->connect("value_changed", callable_mp(this, &AudioStreamImportSettingsDialog::_settings_changed).unbind(1));
--- a/scene/gui/range.cpp
+++ b/scene/gui/range.cpp
@@ -98,7 +98,7 @@ void Range::_set_value_no_signal(double p_val) {
 		return;
 	}
 
-	if (shared->step > 0) {
+	if (shared->step > 0 && !shared->skip_step_rounding) {
 		p_val = Math::round((p_val - shared->min) / shared->step) * shared->step + shared->min;
 	}
 
@@ -165,6 +165,10 @@ void Range::set_step(double p_step) {
 
 	shared->step = p_step;
 	shared->emit_changed("step");
+}
+
+void Range::set_disallow_step_rounding(bool disallow) {
+	shared->skip_step_rounding = disallow;
 }
 
 void Range::set_page(double p_page) {
@@ -258,6 +262,7 @@ void Range::unshare() {
 	nshared->max = shared->max;
 	nshared->val = shared->val;
 	nshared->step = shared->step;
+	nshared->skip_step_rounding = shared->skip_step_rounding;
 	nshared->page = shared->page;
 	nshared->exp_ratio = shared->exp_ratio;
 	nshared->allow_greater = shared->allow_greater;
--- a/scene/gui/range.h
+++ b/scene/gui/range.h
@@ -41,6 +41,7 @@ class Range : public Control {
 		double min = 0.0;
 		double max = 100.0;
 		double step = 1.0;
+		bool skip_step_rounding = false;
 		double page = 0.0;
 		bool exp_ratio = false;
 		bool allow_greater = false;
@@ -78,6 +79,7 @@ public:
 	void set_min(double p_min);
 	void set_max(double p_max);
 	void set_step(double p_step);
+	void set_disallow_step_rounding(bool disallow);
 	void set_page(double p_page);
 	void set_as_ratio(double p_value);
 
--- a/modules/minimp3/audio_stream_mp3.h
+++ b/modules/minimp3/audio_stream_mp3.h
@@ -95,7 +95,7 @@ class AudioStreamMP3 : public AudioStream {
 	int channels = 1;
 	float length = 0.0;
 	bool loop = false;
-	float loop_offset = 0.0;
+	double loop_offset = 0.0;
 	void clear_data();
 
 	double bpm = 0;
--- a/modules/minimp3/resource_importer_mp3.cpp
+++ b/modules/minimp3/resource_importer_mp3.cpp
@@ -117,7 +117,7 @@ Ref<AudioStreamMP3> ResourceImporterMP3::import_mp3(const String &p_path) {
 
 Error ResourceImporterMP3::import(const String &p_source_file, const String &p_save_path, const HashMap<StringName, Variant> &p_options, List<String> *r_platform_variants, List<String> *r_gen_files, Variant *r_metadata) {
 	bool loop = p_options["loop"];
-	float loop_offset = p_options["loop_offset"];
+	double loop_offset = p_options["loop_offset"];
 	double bpm = p_options["bpm"];
 	float beat_count = p_options["beat_count"];
 	float bar_beats = p_options["bar_beats"];


MJacred:
Remark:  
I think casting the result of `get_value` to double is unnecessary (as it returns `double`), but this line by Juan made me somewhat uncertain?

https://github.com/godotengine/godot/blob/e01fb324e5a6c8e5401b362a1954b9f692972cac/editor/import/audio_stream_import_settings.cpp#L521

editor/import/audio_stream_import_settings.cpp
```diff
@@ -471,7 +471,7 @@ void AudioStreamImportSettingsDialog::_settings_changed() {
 
 	updating_settings = true;
 	stream->call("set_loop", loop->is_pressed());
-	stream->call("set_loop_offset", loop_offset->get_value());
+	stream->call("set_loop_offset", double(loop_offset->get_value()));
```
bs-mwoerner, line 474: `Range::get_value()` already returns a `double`, though, so this cast doesn't do anything, does it?

editor/import/audio_stream_import_settings.cpp
```diff
@@ -437,7 +437,7 @@ void AudioStreamImportSettingsDialog::edit(const String &p_path, const String &p
 			beats_edit->set_value(beats);
 			beats_enabled->set_pressed(beats > 0);
 			loop->set_pressed(config_file->get_value("params", "loop", false));
-			loop_offset->set_value(config_file->get_value("params", "loop_offset", 0));
+			loop_offset->set_value(double(config_file->get_value("params", "loop_offset", 0)));
```
bs-mwoerner, line 440: Similarly, `Range::set_value()` takes a `double`, so this cast doesn't do anything that didn't happen before.

editor/import/audio_stream_import_settings.cpp
MJacred: I think so too, but I adjusted it because of what I said in https://github.com/godotengine/godot/pull/87860#issuecomment-1923494238.  
I don't trust floating point values in computer systems, so I copied what someone more experienced did.

editor/import/audio_stream_import_settings.cpp
MJacred: See above

editor/import/audio_stream_import_settings.cpp
MJacred: But… yeah. I think this one should be safe. Though the one below is reading from VARIANT::FLOAT afiak (I know that it's a double under the hood… but there should still be in implicit conversion going on that we can make an explicit cast)

editor/import/audio_stream_import_settings.cpp
MJacred: _(copied from previous conversation that was marked resolved)_

Here it is reading from a VARIANT::FLOAT afiak (I know that it's a double under the hood… but there should still be in implicit conversion going on that we can make an explicit cast)

editor/import/audio_stream_import_settings.cpp
```diff
@@ -543,7 +543,6 @@ AudioStreamImportSettingsDialog::AudioStreamImportSettingsDialog() {
 	loop_hb->add_child(memnew(Label(TTR("Offset:"))));
 	loop_offset = memnew(SpinBox);
 	loop_offset->set_max(10000);
-	loop_offset->set_step(0.001);
```
bs-mwoerner, line 546: I don't have a compiler right now to check, but isn't the default step 1.0? If so, this would make it so that you can only loop on full seconds, wouldn't it?

RustyRoboticsBV:
Unfortunately, this doesn't resolve [#87427](https://github.com/godotengine/godot/issues/87427). Removal of the step value causes the `loop_offset` field to only accept integer values.

After re-adding a step value (I used 0.00001), the old truncation problem still persists. For instance, 36.12318 becomes 36.1232 and 360.12318 becomes 360.123.

I think @bs-mwoerner is probably on to something [here](https://github.com/godotengine/godot/pull/87554#issuecomment-1923601714).

editor/import/audio_stream_import_settings.cpp
MJacred: Indeed, https://docs.godotengine.org/en/stable/classes/class_range.html#class-range-property-step says default is `0.01`. Let me check if I can do anything about this…

editor/import/audio_stream_import_settings.cpp
MJacred: Indeed

editor/import/audio_stream_import_settings.cpp
MJacred: The docs may be inaccurate, because in the code it says `double step = 1.0;` :laughing:

MJacred:
@RustyRoboticsBV: I think I got it. If you want to test again?

scene/gui/range.h
```diff
@@ -41,6 +41,7 @@ class Range : public Control {
 		double min = 0.0;
 		double max = 100.0;
 		double step = 1.0;
+		bool skip_step_rounding = false;
```
MJacred, line 44: I'm uncertain if `Shared` is the correct place or the class Range itself. But I think it should be possible (at the very least internally), to prevent rounding.

editor/import/audio_stream_import_settings.cpp
bs-mwoerner: Oh, right! That's definitely a documentation bug then. Looks like 1.0 has been the actual default value for years.

scene/gui/range.h
bs-mwoerner: Sorry for the constant nagging without doing any actual work myself, but I think this new `set_disallow_step_rounding()` method has the same effect as simply doing `set_step(0.0)`? 😅

MJacred: It does have the same effect.

With the additional bonus of breaking the SpinBox: the increment/decrement will stop working.  
With this change, it will keep working and increments/decrements in seconds (which is not ideal, tbh, but first I want to see if it actually works, then we can talk about the steps. Or even use the current steps).

If this works out, I might need to split the PR anyway, because I'm touching GUI stuff… (but first testing if it works)

bs-mwoerner: Ah, that's a good point! Yes, I didn't think about the step buttons. So this way, we could have a SpinBox that supports step increments but still allows floating point entries at full precision. Yes, I think this would be one way to work around the original problem of not being able to enter the precise "seconds" equivalent of the sample count.

MJacred: At first, I wanted to set negative steps to prevent auto-rounding. But then the buttons would be inversed :laughing:

RustyRoboticsBV: Hey, I tested the updated PR, I'm afraid the issue still there. 🙁

The spinbox itself is working fine, though it still rounds values with more than six digits after the decimal dot (but that's more than enough for the loop_offset issue).

However, on reimport the rounding issue persists.

MJacred: Bummer. Hm… I'll have a closer look over the weekend and ask in the core channel (sounds like the best fit because of the serializing)

bs-mwoerner:
I've now explored the idea of having different units (Seconds/Samples/Beats) with different step values some more and created a separate PR for that #87882. That's definitely a more complicated solution than just tweaking the input field, but it allows using round numbers for samples and beats without having to convert them to lengthy decimal time values. This now also keeps the precision across reimports (although this is unrelated to how we handle the actual input). Maybe you can have a look at that and weigh in with some thoughts on whether the additional complexity (and exposing the sampling rate from the audio streams) is worth it.

lyuma:
one reason I was pushing for this change is because it would be easier to keep compatibility with and doesn't require much consensus: it can be argued to be a bugfix, since the rounding we were doing was not a multiple of the sample rate and hence unpredictable.

Even if we do end up #87882, there might be reason to allow precise values in seconds (the user may still prefer to use seconds and copy the precise timestamp from their audio software), so it seems worth solving this in any case.

`set_disallow_step_rounding` or `set_step(0.0)` really ought to work when the native type is double, and if it's not working we should maybe see if there are bugs we can fix that are causing doubles to lose precision.

scene/gui/range.h
KoBeWi: You can set separate arrow step for SpinBox. The new property is unnecessary.

MJacred: @KoBeWi: This PR will be changed (see https://github.com/godotengine/godot/pull/87882#issuecomment-1938766979). I haven't received any response yet, so I'll change the code as stated there.

And I completely forgot that we still run into rounding issues, even if double (de-)serialization would be fully exhausted. It just happens at a later decimal place. So rounding should happen either way.

MJacred:
@lyuma: see https://github.com/godotengine/godot/pull/87860#discussion_r1494937650. And yes, better double (de-)serializing should be done. But that is core. After that, we can increase the freedom of `step` (ie. making it way smaller) for loop offset further.

editor/import/audio_stream_import_settings.cpp
```diff
@@ -472,10 +437,11 @@ void AudioStreamImportSettingsDialog::edit(const String &p_path, const String &p
 			beats_edit->set_value(beats);
 			beats_enabled->set_pressed(beats > 0);
 			loop->set_pressed(config_file->get_value("params", "loop", false));
-			loop_offset->set_value(config_file->get_value("params", "loop_offset", 0));
+			loop_offset->set_value(double(config_file->get_value("params", "loop_offset", 0)));
```
Jaytheway, line 440: This seems not necessary, there's implicit cast to `double` since `Range::set_value` takes `double`.

editor/import/audio_stream_import_settings.cpp
```diff
@@ -551,7 +517,7 @@ void AudioStreamImportSettingsDialog::_settings_changed() {
 
 void AudioStreamImportSettingsDialog::_reimport() {
 	params["loop"] = loop->is_pressed();
-	params["loop_offset"] = loop_offset->get_value();
+	params["loop_offset"] = double(loop_offset->get_value());
```
Jaytheway, line 520: `Range::get_value` returns `double`, no cast needed.

editor/import/audio_stream_import_settings.cpp
```diff
@@ -571,64 +537,60 @@ AudioStreamImportSettingsDialog::AudioStreamImportSettingsDialog() {
...
 	loop_hb->add_spacer();
 	loop_hb->add_child(memnew(Label(TTR("Offset:"))));
 	loop_offset = memnew(SpinBox);
-	loop_offset->set_accessibility_name(TTRC("Offset:"));
 	loop_offset->set_max(10000);
-	loop_offset->set_step(0.001);
-	loop_offset->set_suffix("s");
-	loop_offset->set_h_size_flags(Control::SIZE_EXPAND_FILL);
-	loop_offset->set_stretch_ratio(0.33);
+	loop_offset->set_disallow_step_rounding(true);
```
Jaytheway, line 546: `Range` seems to have changed since this PR, it now has `Range::set_use_rounded_values()` which defaults to `false` and is separate from step quantization.

Here calling `set_step(0.0)` is sufficient to disable step "rounding"/quantization.

scene/gui/range.cpp
Jaytheway, line 101: This is no longer valid with the updated `Range` class, the rounding and step quntization has moved to `Range::_calc_value`

Jaytheway, line 170: No longer valid with the updated `Range`

Jaytheway, line 265: Also not valid

scene/gui/range.h
Jaytheway, line 44: This is no longer valid with the updated `Range`

Jaytheway, line 82: And this is also no longer valid

Jaytheway (changes requested):
The `Range` class seems to have changed since this PR.

There's already `Ragne::set_use_rounded_values()`, which is separate from step quantization.

- With this, `loop_offset->set_step(0.0)` seems sufficient.
- However, the loop offset in mp3 and ogg streams should still be changed to use `double` instead of float.
- mp3 and ogg resource importers will display higher precision values in the `Advanced...` settings window if doubles are used, however, in `ResourceImporterMP3::get_import_options` (and `ResourceImporterOggVorbis` as well) `PropertyInfo` has to be updated for "loop_offset" to also allow higher precision, just to set the step to 0: `PropertyInfo(Variant::FLOAT, PROPERTY_HINT_RANGE, "0.0,0.0,0.0,or_greater", 0)`

Validated the loop offsets using `double` and `set_step(0.0)` with both, ogg and mp3, setting loop point at known odd waveform zero-crossing sample bound - both behaved correctly, looping without click in the audio import advanced editor preview, as well as playing reimported file from `AudioStreamPlayer`.

--- 

Worth noting, that if the user is trying to match sample values by fraction of a second - e.g. if trying to match zero-crossing from audio editors that display sample time as fraction of a second (e.g. Audacity), or converting sample time to seconds - it can be off by 1 tiny fraction in either direction.

In my test I had to move the loop offset of an ogg file from 0.20228 - which is what Audacity was displaying at the zero-crossing sample - to 0.20229 in Godot, where it produced no click on loop.

This might be worth adding to documentation.

PR #80284, opened 2023-08-05
Implement reloading of GDExtensions

dsnopek:
This is a ~~draft~~ PR adding "live" reloading of GDExtensions!

This would (partially) fix https://github.com/godotengine/godot/issues/66231

In order to work with godot-cpp, you need to use the companion PR https://github.com/godotengine/godot-cpp/pull/1200

## What works in my limited testing on Linux/Windows

- After re-compiling the extension, if you unfocus and then refocus the editor window, the extension will unload and then reload, replacing all the classes with their new versions
- It will reuse the method bind objects if the same method is re-added, so if GDScript or some other scripting language cached the method binds, they should keep working!
- The cached virtual method pointers will be cleared, so the new implementation will be picked up. (In my test project, I have it printing something in `_process()` and after reload it'll immediately start printing the new text!)
- Object properties will be saved and restored after the reload

## How to test

- Compile Godot using the code from this PR.
- Update your language binding to support hot-reload:
  - With godot-cpp, use this PR: https://github.com/godotengine/godot-cpp/pull/1200
  - Other language bindings need to be updated to pass `GDExtensionClassCreationInfo2::recreate_instance_func` which is used to create a new object defined by the GDExtension for a pre-existing object on the Godot side.
- Add `reloadable = true` to your .gdextension file

### Downloading pre-compiled binaries:

1. Login into github
2. Click the "checks" tab of this PR
3. On the right there's an "artifacts" button.
4. Download the appropriate build for your plaform

image

---

## TODO:

_UPDATE: I think the remaining TODO's can be done in follow-ups. What's in this PR should provide a working basis that can be built upon!_

- ~~Actual "live reloading" where the editor detects that the shared library has been updated and automatically initiates the reload. In the current PR, you have to manually call `GDExtensionManager.reload_extension(path)`~~
- ~~With the previous point, I think we probably want to allow GDExtensions to mark themselves as not being safe to reload. Or maybe each GDExtension should opt in? It's really only the developer of the GDExtension that wants live reloading - users of third-party GDExtensions probably don't want or need this, especially given all the extra accounting necessary to make it work.~~
- When methods are gone after the reload, it marks the method binds as "invalid" and if you try to call them an error is printed and it won't be called. However, if something is caching method binds, it should probably check if it's still valid, and if not, stop caching it! If GDScript or anything else in the engine is doing this, we should update it. **This could be handled in a follow-up PR - what is in this PR is at least safe.**
[...]

PR #87863, opened 2024-02-02
Expose `NOTIFICATION_EXTENSION_RELOADED` to `ClassDB`

EterDelta:
Currently, the extensions reload notification lacks of ClassDB binding in the Object class.
This prevents it from being generated in the bindings, which doesn't make sense since it was introduced to be used by extensions in the first place. (https://github.com/godotengine/godot/pull/80284#issuecomment-1732641407)

Something that stands out in that discussion is whether it should be removed, as there are no _real world use cases_.
But in fact there are! We can take a look at how **engine classes** like [Bone2d](https://github.com/godotengine/godot/blob/9adb7c7d130c6d4eb0e80b92d6eebd71fac3384d/scene/2d/skeleton_2d.cpp) cache parent pointers when entering the tree. These cached pointers are also used for **editor** tasks, like configuring warnings.
This would be imposible without the reload notification in an extension, as everytime we hot reload, the pointers become dangling, causing us to lose track of whatever we were pointing to in the editor.


dsnopek:
Yes, I agree this should be exposed!

I'd mark it as approved, but it looks like it needs a docs update to pass CI.

Mickeon:
Right off the bat, while exposing it is probably a good idea, it being among both **NOTIFICATION_POSTINITIALIZE** and **NOTIFICATION_PREDELETE** definitely stands out like a sore thumb... It is useful, but it is not useful every day, so to speak? I'm a bit wary of that.

If it is agreed that exposing the constant is fine, then I would see no reason to keep **NOTIFICATION_PREDELETE_CLEANUP** hidden, either.

dsnopek:
In this case, `NOTIFICATION_EXTENSION_RELOADED` is meant to be used by GDExtensions, but the code generation used in most (all?) GDExtension bindings won't know to generate if it's not exposed. I suppose they can still use it by referring to the integer value? In any case, the point I'm trying to make is that this was added _in order_ to be exposed to GDExtension, there's no use for it in core itself :-)

Mickeon (approved):
I see, well... In that case it's fine. I just wish we didn't have to document it in full display like this.
Autocompletion would always show it, even if it can really only be used for GDExtension. It's among a few things that contribute to bloat.

It could be defined in the GDExtension class, couldn't it? ProjectSettings does something similar.

dsnopek (approved):

dsnopek:
> It could be defined in the GDExtension class, couldn't it? ProjectSettings does something similar.

Do you have a reference to the ProjectSettings thing?

We could manually define it in godot-cpp without exposing it. But that won't help any of the other bindings, they'd all have to manually define it too (whereas exposing it the normal way it'll just get picked up by the code generation).

EterDelta:
I mean, its usage can extend beyond  _just GDExtension_. 🤔
As I wrote in the doc, derived classes also get notified, that includes **scripts**.
So, maybe tool scripts that extend extensions classes could implement some custom synchronization that needs to be notified each time the parent reloads.

Mickeon:
> Do you have a reference to the ProjectSettings thing?

Sorry, I actually meant **EditorSettings**. Important distinction

What I do mean is, just because it is received by every **Object** does not mean it needs to be defined in **Object**. **EditorSettings** notifies **NOTIFICATION_EDITOR_SETTINGS_CHANGED** to everything in the tree, and editor nodes answer accordingly. For example:
https://github.com/godotengine/godot/blob/9adb7c7d130c6d4eb0e80b92d6eebd71fac3384d/editor/project_manager.cpp#L107-L112

dsnopek:
Hm. Well, I guess we could do that. Personally, I prefer having it on `Object` because I feel like GDExtension is very much a part of `Object`, however, I'm clearly biased. :-) I'll defer to whatever the @godotengine/core team thinks is best!

akien-mga (approved):

akien-mga:
Makes sense to me that it's exposed in `Object`.

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #87865, opened 2024-02-02
Update Range.xml: fix incorrect default value for `step`

MJacred:
Unless there's a hidden change from `1.0` to `0.01`, the correct default value is `1.0`.

See https://github.com/godotengine/godot/blob/10e111477db68fe65776a1d68fb1ffccaf6520fc/scene/gui/range.h#L43

**EDIT**: If I'm not mistaken, the internal tools cannot handle this case, see https://github.com/godotengine/godot/pull/87865#issuecomment-1953095025

### DIFF
--- a/scene/gui/range.cpp
+++ b/scene/gui/range.cpp
@@ -326,6 +326,9 @@ void Range::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "allow_greater"), "set_allow_greater", "is_greater_allowed");
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "allow_lesser"), "set_allow_lesser", "is_lesser_allowed");
 
+	// Default value would be 0.01 otherwise, override for doc generation purposes.
+	ADD_PROPERTY_DEFAULT("step", 1.0);
+
 	GDVIRTUAL_BIND(_value_changed, "new_value");
 
 	ADD_LINKED_PROPERTY("min_value", "value");


MJacred:
I don't see how the change could fail the Linux w/Mono check and cause an 
```
ERROR: unimplemented base type encountered in renderer scene cull
   at: instance_set_base (servers/rendering/renderer_scene_cull.cpp:636)
```

Mickeon:
No, the default values are generated by the class reference generator and you shouldn't touch them this way. You need to override the value manually in `_bind_methods` to make it an exception to the norm.
**RandomNumberGeneration** does this:
https://github.com/godotengine/godot/blob/10e111477db68fe65776a1d68fb1ffccaf6520fc/core/math/random_number_generator.cpp#L49-L51

MJacred:
@Mickeon: the class reference guides for contribution say nothing about this. And I see only a handful uses of `ADD_PROPERTY_DEFAULT`. I also don't see where it could get the `0.01` from.  
Perhaps the class reference generator (wherever it is) cannot read that value because it's not in the class, but rather in a struct within the class...

Should I close the PR?

AThousandShips:
It seems that it might be due to some delay in the generation, as the actual value even when constructing it is `1`, but this is likely a larger, more complex issue to solve

Mickeon:
> the class reference guides for contribution say nothing about this

Yyyyyeah, it doesn't say _many_ things. It could use a refresher someday... 📝

> Should I close the PR?

If you change it to use **ADD_PROPERTY_DEFAULT** I don't see a reason to. Right now the PR does not work as you may think, because the doc generation will always want to override it.
If you do make the change, you should probably comment the reason on the side.

> There might also be other cases that change with this, so needs a deeper investigation, if necessary

True, but the same could be said for other classes that use the macro. I mentioned **RandomClassGenerator** because it is the most obvious exception.
Other ones worth noting:
https://github.com/godotengine/godot/blob/9adb7c7d130c6d4eb0e80b92d6eebd71fac3384d/scene/2d/collision_shape_2d.cpp#L286-L287
https://github.com/godotengine/godot/blob/10e111477db68fe65776a1d68fb1ffccaf6520fc/servers/audio_server.cpp#L1767-L1769

Very few are not as obvious:

https://github.com/godotengine/godot/blob/10e111477db68fe65776a1d68fb1ffccaf6520fc/modules/multiplayer/scene_multiplayer.cpp#L679
https://github.com/godotengine/godot/blob/10e111477db68fe65776a1d68fb1ffccaf6520fc/scene/2d/tile_map.cpp#L1255

AThousandShips:
I meant that this might be a bug in the doctool, which has other cases like this

Mickeon (changes requested):
See discussion above. I think an **ADD_PROPERTY_DEFAULT** for this would be fine (above it should be documented why it is inserted). At worst it can be re-evaluated later.

MJacred:
@Mickeon: I reserved some time tomorrow evening to change and test it

scene/gui/range.cpp
```diff
@@ -326,6 +326,9 @@ void Range::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "allow_greater"), "set_allow_greater", "is_greater_allowed");
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "allow_lesser"), "set_allow_lesser", "is_lesser_allowed");
 
+	// Default values are non-deterministic, override for doc generation purposes.
```
Mickeon, line 329: This is not true, that's for RNG.
```suggestion
	// Default value would be 0.01 otherwise, override for doc generation purposes.
```

MJacred:
Ok, I've updated the code to use `ADD_PROPERTY_DEFAULT`. But it does not work. I have no idea why. It works for others, but not here. `object.h` is indirectly included. Maybe it does not work for floats, or it does not reach `default_values` in class_db.h. No idea

And I only could get the doctool to run after commenting some code because this error was popping up 6 times:

```
ERROR: Condition "actions.default_repeat == ShaderLanguage::REPEAT_DEFAULT" is true. Returning: String()
   at: _get_sampler_name (servers/rendering/shader_compiler.cpp:283)
```

scene/gui/range.cpp
MJacred: Done

MJacred:
@akien-mga: you have used `ADD_PROPERTY_DEFAULT` before. Do you have any idea why it doesn't work here?

Mickeon:
Interesting that the checks have NOT failed, despite the default value being changed. There may either indeed be an underlying problem here.

> Why do I have the feeling it's because `step` is within `struct Shared`

Oh.

KoBeWi:
It's not a problem with the struct, but rather with Range itself being an abstract class. I'm trying to find where the default value is initialized. Since the class itself is not instantiated, the default is fetched from another class that inherits it (except none of them use `0.01`...).

Mickeon:
I wonder why that's the case and how many properties unintentionally use the wrong default value as a result?

KoBeWi:
Well, since the class can't be instantiated, ClassDB is not expected to fetch a default value, so doctools instead try to fetch the default from the inheriting classes (which might override the default). tbh abstract classes not being instantiable is an artificial limitation, as they are not abstract in C++ sense and creating instance of Range should be normally possible.

I'm not sure how to solve this *properly*. It's possible to make `ADD_PROPERTY_DEFAULT` work properly by tweaking the code I linked, but that still means manually checking for unexpected defaults and hard-coding them. Until abstract classes can be instantiated on their own, they will always be prone to accidentally using default from an inheriting class.

MJacred:
This is a bigger issue than just fixing one property in `Range`. Therefore, I opened a proposal [here](https://github.com/godotengine/godot-proposals/issues/9184), and I'm closing this PR.

PR #86654, opened 2023-12-30
Prevent escape key from closing Editor Settings window when filtering for shortcuts

ryevdokimov:
Fixes: #85688

Escape key previously closed the window when listening for a key, and escape is pressed. 

image

Now it doesn't, matching the behavior of   "Event Configuration" window. 

image

PR #87870, opened 2024-02-02
Remove unnecessary focus methods and directly bind to `set_close_on_escape`

ryevdokimov:
Suggested by @AThousandShips for #86654, figured I would apply it everywhere else.


akien-mga (approved):

akien-mga:
Thanks!

PR #87871, opened 2024-02-02
Use `_v` shorthand for type traits and `if constexpr` where appropriate

vittorioromeo:
This is a pure refactoring PR that shouldn't affect the behavior or Godot in any way or form. I've used the shorthand `_v` notation for type traits (which is more concise, and faster to compile on modern compilers) and promoted some `if` constructs to `if constexpr` where appropriate.

Both of the changes had a precedent in the codebase.

BTW, I'm working on some more improvements to the internals, hope this kind of PR is appreciated!

### DIFF
--- a/core/variant/type_info.h
+++ b/core/variant/type_info.h
@@ -43,14 +43,10 @@ struct EnableIf<false, T> {
 };
 
 template <typename, typename>
-struct TypesAreSame {
-	static bool const value = false;
-};
+inline constexpr bool types_are_same_v = false;
 
-template <typename A>
-struct TypesAreSame<A, A> {
-	static bool const value = true;
-};
+template <typename T>
+inline constexpr bool types_are_same_v<T, T> = true;
 
 template <typename B, typename D>
 struct TypeInherits {
@@ -60,7 +56,7 @@ struct TypeInherits {
 	static char (&test(...))[2];
 
 	static bool const value = sizeof(test(get_d())) == sizeof(char) &&
-			!TypesAreSame<B volatile const, void volatile const>::value;
+			!types_are_same_v<B volatile const, void volatile const>;
 };
 
 namespace GodotTypeInfo {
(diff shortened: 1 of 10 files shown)


core/variant/type_info.h
```diff
@@ -43,14 +43,10 @@ struct EnableIf<false, T> {
 };
 
 template <typename, typename>
-struct TypesAreSame {
-	static bool const value = false;
-};
+inline constexpr bool typesAreSame = false;
```
AThousandShips, line 46: Id not name it this, it breaks all our naming standards, I'd just have it be called as originally

vittorioromeo: What is the preferred naming scheme for `inline constexpr` variable templates?

Using `TypesAreSame` suggests it's a type, which is not.

I can see `is_vector_type` in `variant_converters.h`. There is also `_class_is_enabled` in `object.h`.

Should I use `snake_case`?

AThousandShips: Id say snake case indeed, as this naming isn't used anywhere else AFAIK

vittorioromeo: Renamed to `types_are_same_v`

Geometror (approved):
Looks good! 
Btw these kinds of PRs are definitely appreciated :)

vittorioromeo:
> Looks good! Btw these kinds of PRs are definitely appreciated :)

Thank you, I am glad to hear that!

What is the general procedure for the PRs to get merged after review is approved?

akien-mga:
> What is the general procedure for the PRs to get merged after review is approved?

See https://docs.godotengine.org/en/stable/contributing/workflow/pr_review_guidelines.html#merging-pull-requests

TL;DR: Currently I'm the one overseeing all merges, and they need to fit within relevant needs of the codebase at the time, and the release cycle (e.g. the imminent 4.3-dev3 snapshot). In this case this PR is likely fine to merge at any time as it seems to be a low risk refactoring, so it will be in the next batch.

PR #87872, opened 2024-02-02
Direct3D 12: Enhance management of texture data life cycle

RandomShaper:
1. When available, graphics memory is asked not to be zero-initialized.

   That brings better performance since in most cases those bytes will be overwritten anyway.

~2. Texture memory initialization is tracked across the limbo, garbage and done states.~

   ~- Limbo: resource just allocated, contents are dangerous unless overwritten by certain operations (like a clear), or discarded.~
   ~- Garbage: the memory has been discarded (i.e., D3D12 knows it has to ignore previous contents), but it's still not guaranteed to be black, orthogonally to the non-zero-mem feature.~
   ~- Done: the texture has known contents, either because it was updated with some image data or because it was cleared, worst case by this code, to black, to ensure garbage data is never used.~

**UPDATE:** Approach simplified. It turns out that texture with many many subresources would add a lot of strain.

2. Texture memory initialization is guaranteed by clearing in black every texture that can be a UAV or RTV.

### DIFF
--- a/drivers/d3d12/rendering_device_driver_d3d12.cpp
+++ b/drivers/d3d12/rendering_device_driver_d3d12.cpp
@@ -697,10 +697,6 @@ void RenderingDeviceDriverD3D12::_resource_transition_batch(ResourceInfo *p_reso
 		}
 	}
 
-	if (p_new_state == D3D12_RESOURCE_STATE_UNORDERED_ACCESS) {
-		res_states->last_batch_transitioned_to_uav = res_barriers_batch;
-	}
-
 #ifdef DEBUG_COUNT_BARRIERS
 	frame_barriers_cpu_time += OS::get_singleton()->get_ticks_usec() - start;
 #endif
@@ -1416,6 +1412,9 @@ RDD::TextureID RenderingDeviceDriverD3D12::texture_create(const TextureFormat &p
 			tex_info->aliasing_hack.owner_info.aliases.insert(aliases[i].first, aliases[i].second);
 		}
 	}
+	if ((p_format.usage_bits & (TEXTURE_USAGE_STORAGE_BIT | TEXTURE_USAGE_COLOR_ATTACHMENT_BIT))) {
+		textures_pending_clear.add(&tex_info->pending_clear);
+	}
 
 	return TextureID(tex_info);
 }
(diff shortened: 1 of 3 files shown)


drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -1919,6 +1977,31 @@ D3D12_RENDER_TARGET_VIEW_DESC RenderingDeviceDriverD3D12::_make_rtv_for_texture(
 	return rtv_desc;
 }
 
+D3D12_UNORDERED_ACCESS_VIEW_DESC RenderingDeviceDriverD3D12::_make_ranged_uav_for_texture(const TextureInfo *p_texture_info, uint32_t p_mipmap_offset, uint32_t p_layer_offset, uint32_t p_layers, bool p_add_bases) {
+	D3D12_UNORDERED_ACCESS_VIEW_DESC uav_desc = p_texture_info->aliasing_hack.main_uav_desc;
+
+	// Leveraging union members' compatibility.
```
DarioSamo, line 1983: I feel this trick hurts the readability and maintenance of the code. While it's true the compatibility is guaranteed as long as the headers don't change (and they won't), the semantic meaning of the code is lost since it's not accounting for the existence of every view dimension case we want to cover and filling in their corresponding structures. I'd take the time to properly fill out each case here.

RandomShaper: Sometimes I am lazy... I've reworked it (as well as other spot in the code I was abusing the union) to make the initialization explicit in all cases.

akien-mga:
Needs a rebase, and a re-review.

RandomShaper:
Rebased!

DarioSamo:
PR currently does not handle `D3D12_SRV_DIMENSION_TEXTURECUBEARRAY` in `_make_rtv_for_texture`, so it crashes in dev mode with the following call stack due to the assert.

image
image
image

I could reproduce this on the 3D Platformer project, probably due to the use of reflection probes.

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -3918,6 +4028,22 @@ void RenderingDeviceDriverD3D12::command_uniform_set_prepare_for_use(CommandBuff
...
+
+							// If about to sample a texture whose contents are still garbage, black it out.
+							// This is not the most optimal way to handle this, but should happen once in a blue moon.
+							if ((wanted_state & D3D12_RESOURCE_STATE_ALL_SHADER_RESOURCE) && tex_info->states_ptr->subresource_states[subresource].init_state != ResourceInfo::States::ResourceStateEx::INIT_STATE_DONE) {
+								TextureSubresourceRange range;
+								range.base_mipmap = j;
+								range.mipmap_count = 1;
+								range.base_layer = i;
+								range.layer_count = 1;
+								_command_clear_color_texture(p_cmd_buffer, tex_info, Color(), range, false);
```
DarioSamo, line 4040: Not necessarily a code error just yet but it seems the comment is a bit misleading as to how often this happens. Upon running a very simple project I got this path above 3000 times on startup. So much that it triggered the error where it ran out of RTV descriptors and I had no choice but to increase it.

Probably worth reviewing how often this triggers and if it's intentional.

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -4317,11 +4443,53 @@ void RenderingDeviceDriverD3D12::command_resolve_texture(CommandBufferID p_cmd_b
...
-	if ((tex_info->desc.Flags & D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET)) {
+	_command_clear_color_texture(p_cmd_buffer, tex_info, p_color, p_subresources, true);
+}
+
+void RenderingDeviceDriverD3D12::_command_clear_color_texture(CommandBufferID p_cmd_buffer, TextureInfo *p_tex_info, const Color &p_color, const TextureSubresourceRange &p_subresources, bool p_use_transition_batching) {
+	const CommandBufferInfo *cmd_buf_info = (const CommandBufferInfo *)p_cmd_buffer.id;
+	ID3D12Resource *res_to_clear = p_tex_info->main_texture ? p_tex_info->main_texture : p_tex_info->resource;
+
+	auto _transition_subresources = [&](D3D12_RESOURCE_STATES p_new_state) {
+		D3D12_RESOURCE_BARRIER *non_batched_barriers = ALLOCA_ARRAY(D3D12_RESOURCE_BARRIER, p_subresources.layer_count * p_subresources.mipmap_count);
```
DarioSamo, line 4456: It seems like these barriers are allocated regardless of whether you use transition batching or not. Shouldn't they only be allocated when it's false?

DarioSamo (changes requested):
After fixing the error I mentioned above at least everything seems to still work, but the behavior regarding the excessive amount of clears likely needs to be reviewed, as the comment indicates it shouldn't happen often but it triggers often enough to go above the default limit.

drivers/d3d12/rendering_device_driver_d3d12.cpp
RandomShaper: In the next push I'm only allocating if it's `false`. Regarding the potentially huge count of barriers, I'd wait until it's known to be a real-world issue. The `alloca` is done with the worst case in mind, but it's very likely that many can be merged. As far as I know, `alloca` won't cause a crash by itself, only if addresses beyond the stack turned out to be actually used.

drivers/d3d12/rendering_device_driver_d3d12.cpp
RandomShaper: The new approach should be free from that issue.

DarioSamo (approved):

DarioSamo:
Approved. The new approach to clearing the textures seems a lot cleaner and also requires far less tracking that can hurt performance.

akien-mga:
Thanks!

PR #37796, opened 2020-04-11
Improve consistency in file order

Gogsi:
Fixes #37721 by ensuring that files in EditorFileSystem and ScriptEditor are sorted using NaturalNoCaseComparator. Since EditorFileSystem will always be correctly sorted, no additional sorting is required in i.e. FileSystemDock

PR #87874, opened 2024-02-02
Fix sorting of files/dirs in dialogs

### LINKED ISSUE #87868: Filesystem 4.x regression : Files/fodlers starting with "_" are displayed AFTER "a"/"z" instead of before as it works in other softwares (including Godot 3.X)
### Tested versions

Reproducible in Godot v4.2.1.stable

### System information

Godot v4.2.1.stable - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1070 () - Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz (8 Threads)

### Issue description

In Godot 4.X, the filesystem now wrongly consider files/folders starting with an underscore \"_" should be displayed before files/folders starting with "a" or "z".

- This is a big issue because it prevents users to easily organize the files/folders by just adding \"_" before the files/folders we want to list first in their own folders.
- The workaround of renaming them "a\_" still doesn't work because for example "actions" would still be displayed before "a\_". So we need to do "aaaaaa_" instead, which is annoying to do and hurts readability
- An other issue with this is it means the order of files is different between Godot 4 and other softwares such as Windows Explorer, VSCode and even Godot 3 !

❌ The way it's currently displayed in Godot 4 ("\_" comes after "a" or any letter)

image

✅ The way it should be displayed (as in Godot 3, Windows Explorer, VSCode, and almost any software)

image

### Steps to reproduce

1. Create several folders (or files)
2. Name one so it starts with an underscore "_"
3. The folder starting with "_" is displayed AFTER the other one starting with letters, while it should be displayed first

[FileSystem_4.X_OrderIssue.zip](https://github.com/godotengine/godot/files/14140758/FileSystem_4.X_OrderIssue.zip)

AThousandShips:
Sorts leading `_` before all except `.`.

Optionally bound to scripting

Optionally added to `FileDialog` but can omit that if it's compatibility breaking

Added to a few places, think I've gotten all the places in the editor that handles files by ordering

The exact ordering can be further discussed, i.e. sorting `_` before just alphanumerics, etc., but keeping it relatively simple here

Based on how sorting is done in Windows file explorer (though in Powershell it sorts `_` after uppercase, so it depends, but I think it makes sense based on the reasons in the issue below)

* Fixes: https://github.com/godotengine/godot/issues/87868

### DIFF
--- a/core/string/ustring.cpp
+++ b/core/string/ustring.cpp
@@ -927,52 +927,106 @@ static _FORCE_INLINE_ signed char natural_cmp_common(const char32_t *&r_this_str
 	return 0;
 }
 
+static _FORCE_INLINE_ signed char naturalcasecmp_to_base(const char32_t *p_this_str, const char32_t *p_that_str) {
+	if (p_this_str && p_that_str) {
+		while (*p_this_str == '.' || *p_that_str == '.') {
+			if (*p_this_str++ != '.') {
+				return 1;
+			}
+			if (*p_that_str++ != '.') {
+				return -1;
+			}
+			if (!*p_that_str) {
+				return 1;
+			}
+			if (!*p_this_str) {
+				return -1;
+			}
+		}
+
+		while (*p_this_str) {
+			if (!*p_that_str) {
+				return 1;
+			} else if (is_digit(*p_this_str)) {
+				if (!is_digit(*p_that_str)) {
+					return -1;
+				}
+
+				signed char ret = natural_cmp_common(p_this_str, p_that_str);
+				if (ret) {
+					return ret;
+				}
+			} else if (is_digit(*p_that_str)) {
+				return 1;
+			} else {
+				if (*p_this_str < *p_that_str) { // If current character in this is less, we are less.
+					return -1;
+				} else if (*p_this_str > *p_that_str) { // If current character in this is greater, we are greater.
+					return 1;
+				}
+
+				p_this_str++;
+				p_that_str++;
+			}
+		}
+		if (*p_that_str) {
+			return -1;
+		}
+	}
+
+	return 0;
+}
+
 signed char String::naturalcasecmp_to(const String &p_str) const {
 	const char32_t *this_str = get_data();
 	const char32_t *that_str = p_str.get_data();
 
-	if (this_str && that_str) {
-		while (*this_str == '.' || *that_str == '.') {
-			if (*this_str++ != '.') {
+	return naturalcasecmp_to_base(this_str, that_str);
+}
+
+static _FORCE_INLINE_ signed char naturalnocasecmp_to_base(const char32_t *p_this_str, const char32_t *p_that_str) {
+	if (p_this_str && p_that_str) {
+		while (*p_this_str == '.' || *p_that_str == '.') {
+			if (*p_this_str++ != '.') {
 				return 1;
 			}
-			if (*that_str++ != '.') {
+			if (*p_that_str++ != '.') {
 				return -1;
 			}
-			if (!*that_str) {
+			if (!*p_that_str) {
 				return 1;
 			}
-			if (!*this_str) {
+			if (!*p_this_str) {
 				return -1;
 			}
 		}
 
-		while (*this_str) {
-			if (!*that_str) {
+		while (*p_this_str) {
+			if (!*p_that_str) {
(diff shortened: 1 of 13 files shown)


AThousandShips:
This was originally intentional and done in:
* https://github.com/godotengine/godot/pull/37796

Going with restoring the useful sorting here and unifying everything in this direction instead

editor/editor_file_system.h
```diff
@@ -68,14 +68,6 @@ class EditorFileSystemDirectory : public Object {
...
 	};
 
-	struct FileInfoSort {
-		bool operator()(const FileInfo *p_a, const FileInfo *p_b) const {
-			return p_a->file < p_b->file;
-		}
-	};
-
-	void sort_files();
-
```
AThousandShips, line 78: Removed this as it's not used anywhere that I could find, will restore if necessary

core/string/ustring.cpp
```diff
@@ -927,112 +927,161 @@ static _FORCE_INLINE_ signed char natural_cmp_common(const char32_t *&r_this_str
...
 }
 
-signed char String::naturalcasecmp_to(const String &p_str) const {
-	const char32_t *this_str = get_data();
-	const char32_t *that_str = p_str.get_data();
-
-	if (this_str && that_str) {
-		while (*this_str == '.' || *that_str == '.') {
-			if (*this_str++ != '.') {
+static _FORCE_INLINE_ signed char naturalcasecmp_to_base(const char32_t *p_this_str, const char32_t *p_that_str) {
```
AThousandShips, line 930: As both `file` and `natural` sorting use these I merged them to avoid code duplication

doc/classes/String.xml
```diff
@@ -184,6 +184,22 @@
 				Returns a string with [param chars] characters erased starting from [param position]. If [param chars] goes beyond the string's length given the specified [param position], fewer characters will be erased from the returned string. Returns an empty string if either [param position] or [param chars] is negative. Returns the original string unmodified if [param chars] is [code]0[/code].
 			</description>
 		</method>
+		<method name="filecasecmp_to" qualifiers="const">
+			<return type="int" />
+			<param index="0" name="to" type="String" />
+			<description>
+				Like [method naturalcasecmp_to] but sorts strings starting in [code]_[/code] before those starting in characters other than [code].[/code].
```
Mickeon, line 191: I would typically be against saying anything along the lines of "Look at this method to understand how this ones works". But in this case there's a lot to say and the difference is minimal, so okay!
"Begins with" is consistent with the method called the same way.
```suggestion
				Like [method naturalcasecmp_to] but prioritises strings that begin with periods ([code].[/code]) and underscores ([code]_[/code]) before any other character. Useful when sorting folders or file names.
```
(Similar suggestion for `filenocasecmp_to` below. Applies to StringName as well)

AThousandShips: The period part is already in `natural` comparisons, but applying this, thank you!

Mickeon (approved):
Nothing against this by the way, if it's the one way to fix this. I am not convinced on exposing this, since I can't recall anyone explicitly desiring it, but I am not severely opposed to it, either.

_Maybe my pure approval will attract the rest of the herd_.

core/string/ustring.h
```diff
@@ -265,6 +265,9 @@ class String {
 	signed char nocasecmp_to(const String &p_str) const;
 	signed char naturalcasecmp_to(const String &p_str) const;
 	signed char naturalnocasecmp_to(const String &p_str) const;
+	// Special sorting for files to put files starting in `_` before all except `.`, and does natural comparison otherwise.
```
Geometror, line 268: Small nitpick :)
```suggestion
	// Special sorting for file names. Names starting with `_` are put before all others except `.`, otherwise natural comparison is used.
```
... or something like that. At least I'd replace `in` with `with`.

Geometror (approved):
Looks good and seems to work fine (tested using the filesystem dialog). I think it makes sense to expose this even though nobody explicitly asked for it, because this is the behavior which was present in Godot 3.x.

AThousandShips:
Will fixup and squash :)

akien-mga:
Thanks!

AThousandShips:
Thank *you*!

Issue #77915, opened 2023-06-06
The frame field (the spinbox) in visual profiler shows big number.


MiaoBlade:
### Godot version

4.0.3.stable

### System information

win10 vulkan

### Issue description

![frame](https://github.com/godotengine/godot/assets/103508289/6f1ff1df-eaab-4cda-b31f-01790fbd4bc2)
the value is 18446744073709549568
sometimes output shows:
Cannot represent 18446744073709549568 as a 64-bit signed integer, since the value is too large.

### Steps to reproduce

i locate this bug.
in godot\editor\debugger\editor_visual_profiler.cpp line 70:
`cursor_metric_edit->set_max(frame_metrics[last_metric].frame_number);`
`cursor_metric_edit->set_min(MAX(frame_metrics[last_metric].frame_number - frame_metrics.size(), 0u));`
![watch](https://github.com/godotengine/godot/assets/103508289/256f4363-e508-4d63-a45f-e0bf804153f3)
unsigned substraction result in big number

### Minimal reproduction project

N/A

Calinou:
Thanks for the report! Consolidating in https://github.com/godotengine/godot/issues/62341.

PR #87876, opened 2024-02-02
Fix frame number underflow in visual profiler

### LINKED ISSUE #62341: Can't scroll Visual Profiler until Profiler Frame History Size is exceeded
### Godot version

v4.0.alpha.custom_build [d1dac8427]

### System information

Windows 10, Vulkan Clustered, Ryzen R5 3600, RTX 3070 (512.15)

### Issue description

When trying to start the Visual Profiler immediately after starting a project (even an empty one), the profiler elements fail to populate the left column for a few seconds, and the Frame Number locks at `18446744073709549568`.

image

If the profiler is stopped while this number is visible, it will be impossible to seek through the timeline, or to set the frame number manually. After waiting around 20 to 30 seconds after project start, this issue subsides.

You can still see the colored lines appear in the timeline, they just can't be seeked through.

https://user-images.githubusercontent.com/43602802/175263483-1255cb8f-65c1-468e-8c5e-6822d12f5170.mp4

I noticed that the Frame Number would always unfreeze at around ~3600 frames, which turned out to coincide with the default `Profiler Frame History Size` value of 3600; 60 seconds at 60fps. Since I have a 144hz monitor (and V-Sync is enabled by default), this seemed to explain why this problem lasts 20-ish seconds for me.

https://user-images.githubusercontent.com/43602802/175266360-7f9c8a19-6939-4a9e-b420-cf0fade70d67.mp4

Sure enough, adjusting the buffer size directly impacts the length of time that the Visual Profiler's timeline remains unresponsive. This means that the timeline cannot be seeked through until it's already full.

In addition to changing the buffer size, this duration can be increased by capping the framerate, or greatly reduced by disabling V-Sync. Reducing this hang-time also causes the profiled elements to appear much sooner than before. 

None of these issues seem to occur with the standard Profiler.

### Profiler Frame History Size = 480, 144 FPS, V-Sync ON

https://user-images.githubusercontent.com/43602802/175267005-97b7331f-388b-49f3-a24e-ff37508aedc2.mp4

### Steps to reproduce

1. Open a project.
2. Play (F5).
3. Alt-Tab back to the editor and open the Visual Profiler.
4. Press Start.

This seems to happen regardless of the project, so a Minimal Reproduction Project would literally just be a new empty project.

permelin:
Fixes #62341 and #77915

Unsigned subtraction underflow caused the frame counter to be locked at 2^64 in the visual profiler until `debugger/profiler_frame_history_size` number of frames had passed.

I'm bad at C++ but I _think_ this cast is safe, even after 2^63 frames.

### DIFF
--- a/editor/debugger/editor_visual_profiler.cpp
+++ b/editor/debugger/editor_visual_profiler.cpp
@@ -70,7 +70,7 @@ void EditorVisualProfiler::add_frame_metric(const Metric &p_metric) {
 	updating_frame = true;
 	clear_button->set_disabled(false);
 	cursor_metric_edit->set_max(frame_metrics[last_metric].frame_number);
-	cursor_metric_edit->set_min(MAX(frame_metrics[last_metric].frame_number - frame_metrics.size(), 0u));
+	cursor_metric_edit->set_min(MAX(int64_t(frame_metrics[last_metric].frame_number) - frame_metrics.size(), 0));
 
 	if (!seeking) {
 		cursor_metric_edit->set_value(frame_metrics[last_metric].frame_number);


editor/debugger/editor_visual_profiler.cpp
```diff
@@ -70,7 +70,7 @@ void EditorVisualProfiler::add_frame_metric(const Metric &p_metric) {
 	updating_frame = true;
 	clear_button->set_disabled(false);
 	cursor_metric_edit->set_max(frame_metrics[last_metric].frame_number);
-	cursor_metric_edit->set_min(MAX(frame_metrics[last_metric].frame_number - frame_metrics.size(), 0u));
+	cursor_metric_edit->set_min(MAX(static_cast<int64_t>(frame_metrics[last_metric].frame_number) - frame_metrics.size(), 0u));
```
Calinou, line 73: Wouldn't this work too?

```suggestion
	cursor_metric_edit->set_min(MAX(int64_t(frame_metrics[last_metric].frame_number) - frame_metrics.size(), 0u));
```

permelin: That works too but I thought `static_cast<>` was the preferred way to do it. As I said, I'm bad at C++.

permelin: I have updated the PR to use the style of casting that you proposed.

I searched through the codebase and `int64_t()` is much more commonly used than `static_cast<int64_t>`. I also checked some C++ reference docs and semantically both should be identical when it comes to numeric data types.

editor/debugger/editor_visual_profiler.cpp
```diff
@@ -70,7 +70,7 @@ void EditorVisualProfiler::add_frame_metric(const Metric &p_metric) {
 	updating_frame = true;
 	clear_button->set_disabled(false);
 	cursor_metric_edit->set_max(frame_metrics[last_metric].frame_number);
-	cursor_metric_edit->set_min(MAX(frame_metrics[last_metric].frame_number - frame_metrics.size(), 0u));
+	cursor_metric_edit->set_min(MAX(int64_t(frame_metrics[last_metric].frame_number) - frame_metrics.size(), 0u));
```
akien-mga, line 73: This doesn't seem to trigger a compiler warning, but technically the first argument to `MAX` is now `int64_t`, so it should probably not use the unsigned `0u` as second argument.
```suggestion
	cursor_metric_edit->set_min(MAX(int64_t(frame_metrics[last_metric].frame_number) - frame_metrics.size(), 0));
```

akien-mga (approved):

editor/debugger/editor_visual_profiler.cpp
permelin: It is `constexpr auto MAX(const T m_a, const T2 m_b)` so the arguments doesn't have to be the same type, and I assume that greater-than will work as expected between signed and unsigned. But this feels sloppy so I will push a correction.

There is no way I can just accept your suggestion and squash it right here in the Github UI, right?

akien-mga: > There is no way I can just accept your suggestion and squash it right here in the Github UI, right?

Sadly no :/

akien-mga:

akien-mga:
Cherry-picked for 4.2.2.

PR #87878, opened 2024-02-02
Add basic Emacs `.gitignore` entries

adamscott:
Adds these entries to the `.gitignore` file for the [Emacs editor](https://www.gnu.org/software/emacs/).
```
\#*#
.\#*
```
These are temporary files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,6 +132,10 @@ cppcheck-cppcheck-build-dir/
 *.pydevproject
 *.launch
 
+# Emacs
+\#*\#
+.\#*
+
 # GCOV code coverage
 *.gcda
 *.gcno


AThousandShips (approved):
Looked and confirmed these are in fact Emacs temporary files (I believe it also uses `~` variations which we already catch)

Showing off that Emacs nerd cred! /j (joke as in fake implying the existing attitudes of Emacs etc. and the editor wars, to clarify)

.gitignore
```diff
@@ -132,6 +132,10 @@ cppcheck-cppcheck-build-dir/
 *.pydevproject
 *.launch
 
+# Emacs
+\#*#
```
aaronfranke, line 136: Doesn't this need to be `\#*\#` for the second `#` to be accepted? Otherwise that's a comment.

adamscott: Oh, right!

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

akien-mga:
Cherry-picked for 3.6.
Cherry-picked for 3.5.4.

PR #87881, opened 2024-02-02
Add HeightMapShape3D functions to get min / max height

smix8:
Adds `HeightMapShape3D` functions to get `min` / `max` height.

Not sure why those were never exposed while already tracked on the shape. Knowing what is the min and max height in the heights array is very useful information to avoid looping over the entire large array manually.

Also there are no longer pool arrays in Godot 4, they are now packed arrays, but I don't see much value in mentioning this in the text when the function makes it obvious.

### DIFF
--- a/doc/classes/HeightMapShape3D.xml
+++ b/doc/classes/HeightMapShape3D.xml
@@ -9,9 +9,23 @@
 	</description>
 	<tutorials>
 	</tutorials>
+	<methods>
+		<method name="get_max_height" qualifiers="const">
+			<return type="float" />
+			<description>
+				Returns the largest height value found in [member map_data]. Recalculates only when [member map_data] changes.
+			</description>
+		</method>
+		<method name="get_min_height" qualifiers="const">
+			<return type="float" />
+			<description>
+				Returns the smallest height value found in [member map_data]. Recalculates only when [member map_data] changes.
+			</description>
+		</method>
+	</methods>
 	<members>
 		<member name="map_data" type="PackedFloat32Array" setter="set_map_data" getter="get_map_data" default="PackedFloat32Array(0, 0, 0, 0)">
-			Height map data, pool array must be of [member map_width] * [member map_depth] size.
+			Height map data. The array's size must be equal to [member map_width] multiplied by [member map_depth].
 		</member>
 		<member name="map_depth" type="int" setter="set_map_depth" getter="get_map_depth" default="2">
 			Number of vertices in the depth of the height map. Changing this will resize the [member map_data].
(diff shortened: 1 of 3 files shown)


doc/classes/HeightMapShape3D.xml
```diff
@@ -9,9 +9,23 @@
...
+			<return type="float" />
+			<description>
+				Returns the smallest height value found in [member map_data].
+			</description>
+		</method>
+	</methods>
 	<members>
 		<member name="map_data" type="PackedFloat32Array" setter="set_map_data" getter="get_map_data" default="PackedFloat32Array(0, 0, 0, 0)">
-			Height map data, pool array must be of [member map_width] * [member map_depth] size.
+			Height map data, array must be of [member map_width] * [member map_depth] size.
```
Mickeon, line 28: Since you're here:
```suggestion
			Height map data. The array's size must be equal to [member map_width] multiplied by [member map_depth].
```

doc/classes/HeightMapShape3D.xml
```diff
@@ -9,9 +9,23 @@
...
+		<method name="get_max_height" qualifiers="const">
+			<return type="float" />
+			<description>
+				Returns the largest height value found in [member map_data].
+			</description>
+		</method>
+		<method name="get_min_height" qualifiers="const">
+			<return type="float" />
+			<description>
+				Returns the smallest height value found in [member map_data].
```
kleonc, line 22: Might be a good idea to explicitly mention that these return cached value / they don't do any calculation.

smix8: But can you change the map_data to get things out-of-sync? I dont think so since it uses a packed array users need to set the array again to update anything.

kleonc: I wasn't suggesting they can be out of sync or anything like that (no idea if they actually can). I only meant that from the user POV I have no idea whether min/max heights are precalculated or whether they're calculated on the fly when I'm calling these methods. It would be nice to know that they are cheap to call (`O(1)`, not `O(map_data.size())`).
So an additional note would make sense to me, something like:
```suggestion
				Returns the largest height value found in [member map_data] (the value is precalculated, calling this method does not involve iterating over [member map_data]).
			</description>
		</method>
		<method name="get_min_height" qualifiers="const">
			<return type="float" />
			<description>
				Returns the smallest height value found in [member map_data] (the value is precalculated, calling this method does not involve iterating over [member map_data]).
```

Mickeon (approved):
Docs are nice. I can somewhat speak on implementation? These values already existed internally, this PR just exposes them.

kleonc (approved):
Exposing these makes sense to me.
(+I think I've already seen these values being calculated manually in some user code; not sure if I'm not making this up though :laughing:)

akien-mga:
Thanks!

PR #87886, opened 2024-02-03
Do not use `NOTIFICATION_PHYSICS_PROCESS` for 2D Editor

### LINKED ISSUE #87885: Certain editor operations rely on `NOTIFICATION_PHYSICS_PROCESS` without a good reason
### Tested versions

- Reproducible in: 4.2.1.stable, 4.3.dev cae759949

### System information

Godot v4.3.dev (cae759949) - Fedora Linux 39 (KDE Plasma) - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4090 (nvidia; 545.29.06) - 13th Gen Intel(R) Core(TM) i9-13900K (32 Threads)

### Issue description

Certain editor operations rely on `NOTIFICATION_PHYSICS_PROCESS` without a good reason. This is mainly seen through `canvas_item_editor_plugin.cpp` and `node_3d_editor_plugin.cpp`.

This has two consequences:

- At default physics ticks per second (60), moving around 2D elements in the editor results in the orange box being slightly behind and updating less smoothly on high refresh rate monitors.
- At custom physics ticks per second, editor behavior will differ with these elements appearing to lag behind more if the value is lower than 60.

### Steps to reproduce

See this 120 FPS video where a Sprite2D is being dragged around quickly (watch it at 0.5× speed or pause at certain moments to make the issue more noticeable):

https://github.com/godotengine/godot/assets/180032/ffc4cd67-29dc-4d41-aec7-f772d60e7ba8

If **Physics Ticks per Second** project setting is set to 10 and the editor is restarted, the issue becomes more extreme:

https://github.com/godotengine/godot/assets/180032/7bd95c3f-a865-4c9f-8c2d-e014aaed5801

N/A

ryevdokimov:
Edit: 3D descoped, see comments below.

Fixes: #87885

The editor plugins shouldn't use physics process for thing unrelated to physics for reasons explained in the attached issue.  

~~This PR also removes unnecessary method to snap to floor that was part of a physics process~~

https://github.com/godotengine/godot/assets/105675984/69824820-eaf5-4852-b857-8c0d53f416f2

### DIFF
--- a/editor/plugins/canvas_item_editor_plugin.cpp
+++ b/editor/plugins/canvas_item_editor_plugin.cpp
@@ -3917,7 +3917,7 @@ void CanvasItemEditor::_notification(int p_what) {
 			EditorRunBar::get_singleton()->connect("stop_pressed", callable_mp(this, &CanvasItemEditor::_update_override_camera_button).bind(false));
 		} break;
 
-		case NOTIFICATION_PHYSICS_PROCESS: {
+		case NOTIFICATION_PROCESS: {
 			EditorNode::get_singleton()->get_scene_root()->set_snap_controls_to_pixels(GLOBAL_GET("gui/common/snap_controls_to_pixels"));
 
 			int nb_having_pivot = 0;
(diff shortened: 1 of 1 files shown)


Calinou:
> This PR also removes unnecessary method to snap to floor that was part of a physics process.

It might have been done this way so that the method is only called once if multiple objects are selected.

ryevdokimov:
Oh you’re probably right, let me take a look at that, might have to descope it from this PR.

ryevdokimov:
I'm not seeing the difference it made with how the snap to floor object code was originally wrote, and have no issue snapping hundreds of objects to the floor. In fact it takes longer for me to select all the objects than it does for it to snap.

https://github.com/godotengine/godot/assets/105675984/593390eb-d31a-4a68-a208-1d1a04bcf53e

ryevdokimov:
Ok I think I have an explanation, after going through git blame which I should have done in the first place. 

2D being on physics ticks has been the case since the dawn of time, and I don't see any reason for it to continue to be.

image

3D on the other hand... 

#67411

Was apparently to fix the snapping issue and an instantiating issue when 'Run on Separate Thread' is used. 

image

This isn't ideal because it ties viewport updates to physics ticks, but I think this is now beyond the scope of the PR to fix, and can be looked at later.  

I'm going to descope 3D, and make this a 2D only update. I've looked through the code and found now raycast/collision related code in 2D, so I think this is a safe update.

editor/plugins/canvas_item_editor_plugin.cpp
```diff
@@ -3917,7 +3917,7 @@ void CanvasItemEditor::_notification(int p_what) {
 			EditorRunBar::get_singleton()->connect("stop_pressed", callable_mp(this, &CanvasItemEditor::_update_override_camera_button).bind(false));
 		} break;
 
-		case NOTIFICATION_PHYSICS_PROCESS: {
+		case NOTIFICATION_PROCESS: {
 			EditorNode::get_singleton()->get_scene_root()->set_snap_controls_to_pixels(GLOBAL_GET("gui/common/snap_controls_to_pixels"));
```
akien-mga, line 3921: It's a pre-existing issue, but re-setting this every frame (and thus re-querying the project settings for it) seems quite wasteful.

I would move it to the constructor, and connect to `ProjectSettings.settings_changed` to update.

akien-mga (approved):

akien-mga:
Thanks!

PR #87887, opened 2024-02-03
Fix unintentional translations in AnimationTree

### LINKED ISSUE #86033: AnimationTree has unintentional translations
### Tested versions

4.2

### System information

### Issue description

I noticed that AnimationTree will translate animation names.

AnimationPlayer:
image

AnimationTree:
image
Both the display and list are translated.

Then on another look, seems like are more unintentional translations here, like node Titles and path.
image
Here Output is fetched from the bottom tab name (which in Polish is translated as console, because literal translation would be less clear). Seems like the node has a name, which gets capitalized and then auto-translated (as evident by error messages that call it "output").

image
Here Root is translated, which I think is another case of capitalized name.

### Steps to reproduce

0. Make sure editor is not using English
1. Create AnimationPlayer
2. Add animation with a name that exists in translations
3. Add AnimationTree and set AnimationNodeBlendTree as root
4. Add Animation node

N/A

jtakakura:
Fixes #86033

### DIFF
--- a/editor/plugins/animation_blend_space_1d_editor.cpp
+++ b/editor/plugins/animation_blend_space_1d_editor.cpp
@@ -804,6 +804,7 @@ AnimationNodeBlendSpace1DEditor::AnimationNodeBlendSpace1DEditor() {
 	animations_menu = memnew(PopupMenu);
 	menu->add_child(animations_menu);
 	animations_menu->set_name("AddAnimations");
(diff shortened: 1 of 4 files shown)


akien-mga (approved):

akien-mga:
Thanks!

conde2:
🙌👏🏻 thanks

PR #87889, opened 2024-02-03
Add HeightMapShape3D update with Image data

smix8:
Adds `HeightMapShape3D` update with `Image` data.

Populating height values in a heightmap collision from an image file is a common use.
It shouldnt be required that users need to read large image data arrays with slow scripts to achieve this basic thing.

So what this pr does is add `update_map_data_from_image()` as a new function to `HeightMapShape3D`.

The function allows a single component black&white Image in format  `FORMAT_RF` (32 bit) or `FORMAT_RH`  (16 bit) or `FORMAT_R8` (8 bit) as input and to define a minimum and maximum height range. The image pixel values read in `0.0` to `1.0` range are then remapped to those values.

The code for this was actually found in part on the PhysicsServer3D as more or less dead-code not reachable by users. It also makes far more sense to have this functionality on the collision shape directly to update the `map_data`. This data array is commonly used for other related stuff, like creating procedual meshes from the heightmap data.

Full usage example to create heightmap collision from an image file:

```gdscript
extends Node3D

func _ready() -> void:
    var heightmap_texture: Texture = ResourceLoader.load("res://heightmap_image.exr")
    var heightmap_image: Image = heightmap_texture.get_image()
    heightmap_image.convert(Image.FORMAT_RF)

    var heightmap_shape: HeightMapShape3D = HeightMapShape3D.new()
    var height_min: float = 0.0
    var height_max: float = 10.0
    heightmap_shape.update_map_data_from_image(heightmap_image, height_min, height_max)

    var static_body: StaticBody3D = StaticBody3D.new()
    var collision_shape: CollisionShape3D = CollisionShape3D.new()
    collision_shape.shape = heightmap_shape
    static_body.add_child(collision_shape)
    add_child(static_body)
```

### DIFF
--- a/doc/classes/HeightMapShape3D.xml
+++ b/doc/classes/HeightMapShape3D.xml
@@ -6,6 +6,19 @@
 	<description>
 		A 3D heightmap shape, intended for use in physics. Usually used to provide a shape for a [CollisionShape3D]. This is useful for terrain, but it is limited as overhangs (such as caves) cannot be stored. Holes in a [HeightMapShape3D] are created by assigning very low values to points in the desired area.
 		[b]Performance:[/b] [HeightMapShape3D] is faster to check collisions against than [ConcavePolygonShape3D], but it is significantly slower than primitive shapes like [BoxShape3D].
+		A heightmap collision shape can also be build by using an [Image] reference:
+		[codeblocks]
+		[gdscript]
+		var heightmap_texture: Texture = ResourceLoader.load("res://heightmap_image.exr")
+		var heightmap_image: Image = heightmap_texture.get_image()
+		heightmap_image.convert(Image.FORMAT_RF)
+
+		var height_min: float = 0.0
+		var height_max: float = 10.0
+
+		update_map_data_from_image(heightmap_image, height_min, height_max)
+		[/gdscript]
+		[/codeblocks]
 	</description>
 	<tutorials>
 	</tutorials>
@@ -20,6 +33,17 @@
 			<return type="float" />
 			<description>
 				Returns the smallest height value found in [member map_data]. Recalculates only when [member map_data] changes.
+			</description>
+		</method>
+		<method name="update_map_data_from_image">
+			<return type="void" />
+			<param index="0" name="image" type="Image" />
+			<param index="1" name="height_min" type="float" />
+			<param index="2" name="height_max" type="float" />
+			<description>
+				Updates [member map_data] with data read from an [Image] reference. Automatically resizes heightmap [member map_width] and [member map_depth] to fit the full image width and height.
+				The image needs to be in either [constant Image.FORMAT_RF] (32 bit), [constant Image.FORMAT_RH] (16 bit), or [constant Image.FORMAT_R8] (8 bit).
+				Each image pixel is read in as a float on the range from [code]0.0[/code] (black pixel) to [code]1.0[/code] (white pixel). This range value gets remapped to [param height_min] and [param height_max] to form the final height value.
 			</description>
 		</method>
 	</methods>
--- a/scene/resources/3d/height_map_shape_3d.cpp
+++ b/scene/resources/3d/height_map_shape_3d.cpp
@@ -30,6 +30,7 @@
 
 #include "height_map_shape_3d.h"
 
+#include "core/io/image.h"
 #include "servers/physics_server_3d.h"
 
 Vector<Vector3> HeightMapShape3D::get_debug_mesh_lines() const {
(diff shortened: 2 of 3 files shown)


Redwarx008:
Great feature, it would be even better if the HeightMapShape3D could be updated partially from the height map image.

scene/resources/height_map_shape_3d.cpp
```diff
@@ -179,6 +180,48 @@ Vector<real_t> HeightMapShape3D::get_map_data() const {
 	return map_data;
 }
 
+void HeightMapShape3D::update_map_data_from_image(const Ref<Image> &p_image, real_t p_height_min, real_t p_height_max) {
+	ERR_FAIL_COND(p_image.is_null());
+	ERR_FAIL_COND(p_image->get_format() != Image::FORMAT_RF);
+	ERR_FAIL_COND(p_image->get_width() < 2);
+	ERR_FAIL_COND(p_image->get_height() < 2);
```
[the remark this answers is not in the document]
smix8: A heightmap width or depth of 1 is not a valid collision surface, it is just a line without collision that disappears.
Reason why a new heightmap starts with width/depth of 2 by default.
Also other physics extensions like Jolt op on a min block size of 2 for the same and other reasons.

doc/classes/HeightMapShape3D.xml
```diff
@@ -6,9 +6,35 @@
...
+	<methods>
+		<method name="update_map_data_from_image">
+			<return type="void" />
+			<param index="0" name="image" type="Image" />
+			<param index="1" name="height_min" type="float" />
+			<param index="2" name="height_max" type="float" />
+			<description>
+				Updates [member map_data] with data read from an [Image] reference in [constant Image.FORMAT_RF] format. Automatically resize heightmap [member map_width] and [member map_depth] to fit the full image width and height.
+				Each image pixel is read in as a float on the range from [code]0.0[/code] (black pixel) to [code]1.0[/code] (white pixel). This range value gets remapped to [param height_min] and [param height_max] to form the final height value.
+				[b]Performance:[/b] The image data needs to be received from the GPU, stalling the [RenderingServer] in the process. This can have a noticeable frame rate impact at runtime when the image update is used frequently or with high resolution images.
```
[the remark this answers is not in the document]
smix8: You are right and should be removed. I was just so used to the stutters because it is basically always a texture involve when I load heightmaps.

scene/resources/height_map_shape_3d.cpp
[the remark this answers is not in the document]
kleonc: You're right, it's the setters which are problematic here. But that's out of scope of this PR I guess.

scene/resources/height_map_shape_3d.cpp
```diff
@@ -187,6 +188,48 @@ real_t HeightMapShape3D::get_max_height() const {
 	return max_height;
 }
 
+void HeightMapShape3D::update_map_data_from_image(const Ref<Image> &p_image, real_t p_height_min, real_t p_height_max) {
+	ERR_FAIL_COND(p_image.is_null());
+	ERR_FAIL_COND(p_image->get_format() != Image::FORMAT_RF);
+	ERR_FAIL_COND(p_image->get_width() < 2);
+	ERR_FAIL_COND(p_image->get_height() < 2);
+	ERR_FAIL_COND(p_height_min > p_height_max);
```
kleonc, line 196: Missed earlier but these should likely be changed to `ERR_FAIL_COND_MSG` with user-friendly messages (especially given that e.g. it's not mentioned in the docs that image needs to be at least 2x2 etc.).

smix8: Agree, I will update the pr with error macros msgs for all the setters. The parameter names are also not in sync.

doc/classes/HeightMapShape3D.xml
```diff
@@ -6,6 +6,19 @@
 	<description>
 		A 3D heightmap shape, intended for use in physics. Usually used to provide a shape for a [CollisionShape3D]. This is useful for terrain, but it is limited as overhangs (such as caves) cannot be stored. Holes in a [HeightMapShape3D] are created by assigning very low values to points in the desired area.
 		[b]Performance:[/b] [HeightMapShape3D] is faster to check collisions against than [ConcavePolygonShape3D], but it is significantly slower than primitive shapes like [BoxShape3D].
+		A heightmap collision shape can also be build by using an [Image] reference:
+		[codeblocks]
+		[gdscript]
+		var heightmap_texture: Texture = ResourceLoader.load("res://heightmap_image.png")
```
Calinou, line 12: PNG precision in Godot is limited to 8 bits per channel, so you won't be able to use the full range this way. You need to load an EXR/HDR image instead.

doc/classes/HeightMapShape3D.xml
```diff
@@ -22,6 +35,16 @@
 				Returns the smallest height value found in [member map_data]. Recalculates only when [member map_data] changes.
 			</description>
 		</method>
+		<method name="update_map_data_from_image">
+			<return type="void" />
+			<param index="0" name="image" type="Image" />
+			<param index="1" name="height_min" type="float" />
+			<param index="2" name="height_max" type="float" />
+			<description>
+				Updates [member map_data] with data read from an [Image] reference in [constant Image.FORMAT_RF] format. Automatically resize heightmap [member map_width] and [member map_depth] to fit the full image width and height.
```
Calinou, line 44: Should Image.FORMAT_RH be supported as well? This could help reduce loading times and memory usage for large heightmaps if you don't need as much precision (e.g. if your terrain is relativel flat).

smix8: Well in that case if we are going for memory and load times might as well also add FORMAT_R8.

scene/resources/height_map_shape_3d.cpp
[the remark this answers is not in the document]
smix8: I will make a dedicated pr do fix all the function mold. Don't want to mix the two.

smix8:
Added error msgs and support for the lower precision formats `Image::FORMAT_RH` and `Image::FORMAT_R8`.

@Redwarx008 Partial updates (and reads) should be also made available, not only for the image but also for the normal byte array. I will see what I can do but that deserves its own pr.

doc/classes/HeightMapShape3D.xml
```diff
@@ -22,6 +35,17 @@
...
 			</description>
 		</method>
+		<method name="update_map_data_from_image">
+			<return type="void" />
+			<param index="0" name="image" type="Image" />
+			<param index="1" name="height_min" type="float" />
+			<param index="2" name="height_max" type="float" />
+			<description>
+				Updates [member map_data] with data read from an [Image] reference. Automatically resizes heightmap [member map_width] and [member map_depth] to fit the full image width and height.
+				The image needs to be in either [constant Image.FORMAT_RF] (32 bit) or [constant Image.FORMAT_RH] (16 bit) or [constant Image.FORMAT_R8] (8 bit).
```
AThousandShips, line 45:
```suggestion
				The image needs to be in either [constant Image.FORMAT_RF] (32 bit), [constant Image.FORMAT_RH] (16 bit), or [constant Image.FORMAT_R8] (8 bit).
```

fire (approved):

Calinou (approved):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/29b3d9e9e538f0aa8effc8ad8bf19a2915292a89), it works as expected.

**Testing project:** [test_pr_87889.zip](https://github.com/godotengine/godot/files/14856252/test_pr_87889.zip)

Screenshot_20240403_193452 png webp

scene/resources/3d/height_map_shape_3d.cpp
```diff
@@ -187,6 +188,105 @@ real_t HeightMapShape3D::get_max_height() const {
...
+
+				map_data_ptrw[i] = height_value;
+			}
+
+		} break;
+
+		default: {
+			// This should never happen but programming is hard.
+			CRASH_NOW();
+		}
```
Calinou, line 280: I would prefer if we don't use `CRASH_NOW()` here, but have nothing happen instead. Godot is designed so games should never crash after all.

scene/resources/3d/height_map_shape_3d.cpp
```diff
@@ -187,6 +188,105 @@ real_t HeightMapShape3D::get_max_height() const {
...
+
+				if (height_value < new_min_height) {
+					new_min_height = height_value;
+				}
+				if (height_value > new_max_height) {
+					new_max_height = height_value;
+				}
+
+				map_data_ptrw[i] = height_value;
+			}
```
Calinou, line 229: I wonder if this code can be moved below the `switch` statement, and have the `default` branch use `return` instead. This way, we can avoid repetition.

akien-mga: Not easily, the way `pixel_value` is calculated depends on the data type, and it needs to be calculate in the `for` loop for each index.

akien-mga:
Thanks!

PR #57760, opened 2022-02-07
Improve compilation time with include-what-you-use (`core` folder)

Geometror:
Part of a series of PRs to reduce the compilation time with [include-what-you-use](https://include-what-you-use.org/) together with some manual fixes due to the complexity of the codebase. [see #57734, #57759]

This PR restructures the includes in the `core` subfolder, so it should probably be merged before the other PRs.

Could help with https://github.com/godotengine/godot/issues/53295.

PR #87893, opened 2024-02-03
Extract `SafeBinaryMutex` to separate header

vittorioromeo:
This change simply extracts 'SafeBinaryMutex' from 'mutex.h' to 'safe_binary_mutex.h', in an effort to reduce the compilation speed impact of including `mutex.h`.

Similarly in nature to #87871, this is a pure refactoring PR that shouldn't affect the behavior or Godot in any way or form.

### DIFF
--- a/core/os/safe_binary_mutex.h
+++ b/core/os/safe_binary_mutex.h
@@ -1,5 +1,5 @@
 /**************************************************************************/
-/*  mutex.h                                                               */
+/*  safe_binary_mutex.h                                                   */
 /**************************************************************************/
 /*                         This file is part of:                          */
 /*                             GODOT ENGINE                               */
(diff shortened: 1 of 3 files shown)


AThousandShips:
Can you show some data to show this reduces compile times?

core/os/safe_binary_mutex.h
```diff
@@ -0,0 +1,124 @@
+/**************************************************************************/
+/*  mutex.h                                                               */
```
AThousandShips, line 2:
```suggestion
/*  safe_binary_mutex.h                                                   */
```

core/os/safe_binary_mutex.h
```diff
@@ -0,0 +1,124 @@
...
+class MutexLock<SafeBinaryMutex<Tag>> {
+	friend class ConditionVariable;
+
+	THREADING_NAMESPACE::unique_lock<THREADING_NAMESPACE::mutex> lock;
+
+public:
+	_ALWAYS_INLINE_ explicit MutexLock(const SafeBinaryMutex<Tag> &p_mutex) :
+			lock(p_mutex.mutex) {
+		SafeBinaryMutex<Tag>::count++;
+	};
```
AThousandShips, line 102:
```suggestion
	}
```

core/os/safe_binary_mutex.h
```diff
@@ -0,0 +1,124 @@
...
+	_ALWAYS_INLINE_ ~MutexLock() {
+		SafeBinaryMutex<Tag>::count--;
+	};
```
AThousandShips, line 105:
```suggestion
	}
```

core/os/safe_binary_mutex.h
```diff
@@ -0,0 +1,124 @@
...
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef SAFE_BINARY_MUTEX_H
+#define SAFE_BINARY_MUTEX_H
+
+#include "core/error/error_macros.h"
+#include "core/typedefs.h"
```
AThousandShips, line 35: These two includes are unnecessary

vittorioromeo:
> Can you show some data to show this reduces compile times?

I've benchmarked the compilation of the entire engine using `ClangBuildAnalyzer` and `-ftime-trace`. Here's the original report:

https://gist.github.com/vittorioromeo/fa153c494ce05da08c988a27130a79e3

There are a few heavy-hitters here and there, and not everything is easy to optimize, so I am starting with some low-hanging fruits. The "expensive headers" section of the report shows that `mutex.h` is included quite often and takes a considerable amount of time. 

Reducing the amount of code that `mutex.h` contains is a way to reduce the impact of including that header. Furthermore, splitting a specialized components (`SafeBinaryMutex`) that is only used in one place to its own header seems like good engineering practice to me regardless of what the impact on compilation time might be.

AThousandShips:
Can you show after this change as well?

core/os/safe_binary_mutex.h
vittorioromeo: Transitive includes are a bad software engineering practice for multiple reasons:

- https://softwareengineering.stackexchange.com/a/262020/78524

In fact, the widespread use of this practice throughout the Godot codebase has made it terribly difficult for me to improve compilation times significantly, when it would have been quite simple otherwise. 

I recommend that every header and source file should be as explicit as possible with its `#include` dependencies, so that it can be compiled/refactored in isolation and does not depend on invisible chain of transitive dependencies. 

Since `DEV_ASSERT` and `_ALWAYS_INLINE` are respectively defined in `error_macros.h` and `typedefs.h`, and are directly used in this header file, they should be included explicitly. 

I hope you see the advantages of this approach but I am happy to discuss further if you have any doubts or concerns.

AThousandShips: Well that's how we do things, we don't include what we use, the discussion of this is separate and that's out of scope for this

Please follow how things are done, and discuss changing them separately 🙂

vittorioromeo: > Well that's how we do things, we don't "include what we use", the discussion of this is separate and that's out of scope for this
> 
> Please follow how things are done, and discuss changing them separately 🙂
> 
> I'm well aware of the principle

From your own project's coding style guidelines (emphasis mine):

{5645905F-53B9-431B-BF7E-0150001E4C42}

As far as I can see, I am not breaking any rules, and I'm abiding to the last bullet point by "including what I use".

I also do not see any rule mentioning that I should rely on transitive includes.

AThousandShips: They're not written rules but they're rules we use, if also says "that requires them" which isn't the same

But trust me this is a rule, even if it's not written, let's not get bogged down in the letter of the law (though that page could do with an update)

If you want a second opinion I can provide it, but let's just remove these superfluous includes here

vittorioromeo:
> Can you show after this change as well? Showing the problem without indicating this is a solution would be missing the important part IMO, i.e. the showing it improves compile time

https://gist.github.com/vittorioromeo/f59a1f1e312beb603977256059681717

```
# BEFORE
779723 ms: core/os/mutex.h (included 1104 times, avg 706 ms), included via:

# AFTER
758503 ms: core/os/mutex.h (included 1110 times, avg 683 ms), included via:
```

core/os/safe_binary_mutex.h
vittorioromeo: As I strongly feel this is a mistake, I'd like to start a discussion/issue and get an official statement on the matter after a technical discussion.

What is the appropriate forum to start such discussion? Should I open an issue here, or should I open a thread on `godot-proposals`?

AThousandShips: A proposal on godot-proposals, as an issue and not a proposal with the relevant details 🙂

AThousandShips: Note that this was explicitly rejected before, see:
* https://github.com/godotengine/godot/pull/57760#issuecomment-1503228157

AThousandShips: But if you'll remove these two includes we can run the CI for this and see, and leave the discussion for later

vittorioromeo: @AThousandShips: created a discussion here as requested: https://github.com/godotengine/godot-proposals/discussions/9016

AThousandShips: I said an issue with larger details but a discussion is fine too

vittorioromeo: > I said an issue with larger details but a discussion is fine too

I apologize, I started writing the discussion after you said

> If you feel very strongly about "include what you use" please open a [proposal](https://github.com/godotengine/godot-proposals) if one doesn't already exist and advocate for it

and missed the new comment where you clarified that it should have been an issue.

AThousandShips: I'm sorry for taking a bit harder stance on this than I intended to come across, but I felt this was a simple and non-controversial suggestion following what I'm used to in the codebase, and didn't expect it to become a huge deal to even argue over

Will let someone else review this and hope we can deal with this maturely and professionally and not take these kinds of differences personally 🙂

Have a good day

Geometror (commented):
Looks fine to me, you just need to run clang-format or sort the includes manually by lexicographical order.

vittorioromeo:
@Geometror: should be good to go now! :)

Geometror (approved):

akien-mga (approved):
Makes sense to me.

akien-mga:
Thanks!

PR #87903, opened 2024-02-03
Fix `max_collisions` not being passed in `PhysicsBody3D::test_move`

### LINKED ISSUE #87900: test_move() ignores the max_collisions parameter.
### Tested versions

4.2.1 stable, reproducible in older versions since i can remember.

### System information

Fedora workstation

### Issue description

I'm working in my own implementation of the collide and slide algorithm. After a lot of testing and study i had something that works using move_and_collide as a core for it, however, to keep improving the algorithm and fix the common stutter issues associated with it i need to check for more than one contact point. Here is were move and collided showed it's problems for my uses case, such as continuous collision detection and better contact reporting (Move_and_collide reports a lot of coplanar planes, and the accuracy of the reported information is not enough, often reporting planes that it would not have touched if using CCD). I started re implementing it with test_move (Because a get_rest_info only provides the nearest contact normal and collide_shape only provide points, not normals.) 

The problem is that test_move will return one collision independently of the max_collisions parameter.

### Steps to reproduce

I created a reproduction project with two identical scenes. They consist on a sphere that moves down towards a 4 sided cone, both perfectly centred to make sure that once the sphere touches one point, is touching all of them.

The move_and_collide scene will print 4 collisions as expected.

The test_move scene will print one collision, whatever the actual maximum number of collisions is.

[test_move bug.zip](https://github.com/godotengine/godot/files/14151854/test_move.bug.zip)

AThousandShips:
* Fixes: https://github.com/godotengine/godot/issues/87900

### DIFF
--- a/scene/3d/physics_body_3d.cpp
+++ b/scene/3d/physics_body_3d.cpp
@@ -182,6 +182,7 @@ bool PhysicsBody3D::test_move(const Transform3D &p_from, const Vector3 &p_motion
 
 	PhysicsServer3D::MotionParameters parameters(p_from, p_motion, p_margin);
 	parameters.recovery_as_collision = p_recovery_as_collision;
+	parameters.max_collisions = p_max_collisions;
 
 	return PhysicsServer3D::get_singleton()->body_test_motion(get_rid(), parameters, r);
 }


yosoyfreeman:
Working as intended, thanks!

kleonc (approved):
The change seems trivial, likely it was overlooked when copy-pasting from 2D counterpart which has no `p_max_collisions` parameter (no idea why the discrepancy but that's irrelevant here).

akien-mga:
Thanks!

AThousandShips:
Thank you!

akien-mga:
Cherry-picked for 4.2.2.

PR #81746, opened 2023-09-16
Add ufbx for FBX importing

fire:
This restores FBX imports in Godot.

I would need people to test their models, or downloaded models. Specifically, I want to test that:

* The model imports correctly, with textures and animation
* That the process is smooth and intuitive

To test this properly, a variety of models is required, so do not hesitate to try different ones. You can find free FBX models at https://sketchfab.com/3d-models?features=downloadable (be careful, you may see NSFW models).

If you run into an issue, please post here with your OS, and as much information as you can about the error.

https://github.com/godotengine/godot/assets/32321/e1af70c5-11bc-4c2a-9bad-6645cfb6963f

This work is supported as part of [V-Sekai](https://github.com/V-Sekai/) and with help from @bqqbarbhg.

The work was approved at the Animation / 3D Pipeline meeting and was also discussed with the production team as an experimental marked feature for 4.3.

1. https://godotengine.org/fbx-import/
2. https://godotengine.org/article/fbx-importer-rewritten-for-godot-3-2-4/
3. https://godotengine.org/article/here-comes-godot-3-2/#3d-assets

- [x] Marked experimental
- [x] Added ufbx to Godot Engine
- [x] Correct uv1 and uv2 for import.
- [x] Animation clips support
- [x] Skeletal animation support
- [x] Meshless fbx files are imported with animations
- [x] Blend shapes are imported
- [x] Allow Utf8 string name imports  
- [x] Postponed camera imports
- [x] Postponed light imports
- [x] Blend shape animations are imported
- [x] Eight uv support
- [x] Remove code scaffolding from GLTFDocument
- [x] Prune extra files in thirdparty/ufbx
- [x] Need help with the disable enable scheme with FBX2GLTF
- [x] Handle scale ignore [HARD]
- [x] Transparency is enabled
- [x] HANA bug with clusters being parented to the root and failing import
- [x] Postponed Camera imports again
- [x] Postponed Light imports again
- [x] Add documentation
- [x] Adapt FBXDocumentExtension

No proposal exists because this restores FBX import with around five to eight repeated attempts.

~~See also https://github.com/V-Sekai/godot-ufbx~~

### Finalization after 4.3 window is open

~~- [ ] Fix engine bugs on 2 gigabyte fbx samples (cc-by)~~
- [x] Move ufbx into the godot/thirdparty folder
- [x] Write licensing information

### Finalization before 4.3 window closes

1. [x] Cameras
2. [x] Lights
3. [x] ~~Curve paths~~

### Known Issues

[...]

PR #87905, opened 2024-02-03
Fix crash during import bone remap.

### LINKED ISSUE #87904: Crashing during bone remap
### Tested versions

Reproducible in any version newer then:
[96a95cb9743e755f10fa6a14895ea099e5ac4c51] Add const lvalue ref to container parameters

### System information

Godot v4.3.dev (0a2ddaa6b) - Fedora Linux 39 (Workstation Edition) - Wayland - Vulkan (Forward+) - dedicated AMD Radeon RX 6700 XT (RADV NAVI22) () - 13th Gen Intel(R) Core(TM) i7-13700K (24 Threads)

### Issue description

The editor crashes during the import of a model when a bone map has been applied but not all bones are able to be mapped.

An example model can be found here: https://skfb.ly/6AnLO

This might be related to #84371 and #83041.

With commit 96a95cb9743e755f10fa6a14895ea099e5ac4c51, const was added to the `HashMap` passed into `PostImportPluginSkeletonRenamer::_internal_process`.  This had a subtle change of behavior.   

The code was using the `[]` operator([see here](https://github.com/godotengine/godot/blob/master/editor/import/3d/post_import_plugin_skeleton_renamer.cpp#L60)) on the `HashMap` which previously would insert if the item being accessed did not exist([see more here](https://github.com/godotengine/godot/blob/master/core/templates/hash_map.h#L545)).  This allowed the code to previously work.

The const version, does not insert and instead hits a `CRASH_COND` and crashes ([see here for more](https://github.com/godotengine/godot/blob/master/core/templates/hash_map.h#L538)).

### Steps to reproduce

1. Create new project.
2. Download the GLB version of the model.
3. Import into Godot
4. From the import dialog, add a new BoneMone, selecting the humanoid profile.
5. Click import

The only model I have on hand that reproduces currently is the one linked above.  It is over the max size.  Steps are sufficient to reproduce.

tdaven:
The HashMap being used by can fail to find a bone to remap which will result in a crash since the HashMap is const.

Use the find method on the hashmap to loop values and gracefully handle the failure to find the bone instead of crashing.

Fixes: #87904

### DIFF
--- a/editor/import/3d/post_import_plugin_skeleton_renamer.cpp
+++ b/editor/import/3d/post_import_plugin_skeleton_renamer.cpp
@@ -57,9 +57,9 @@ void PostImportPluginSkeletonRenamer::_internal_process(InternalImportCategory p
 	{
 		int len = skeleton->get_bone_count();
 		for (int i = 0; i < len; i++) {
-			StringName bn = p_rename_map[skeleton->get_bone_name(i)];
-			if (bn) {
-				skeleton->set_bone_name(i, bn);
+			HashMap<String, String>::ConstIterator bn_itr = p_rename_map.find(skeleton->get_bone_name(i));
+			if (bn_itr != p_rename_map.end()) {
+				skeleton->set_bone_name(i, bn_itr->value);
 			}
 		}
 	}
@@ -77,9 +77,9 @@ void PostImportPluginSkeletonRenamer::_internal_process(InternalImportCategory p
 					if (mesh_skeleton && node == skeleton) {
 						int len = skin->get_bind_count();
 						for (int i = 0; i < len; i++) {
-							StringName bn = p_rename_map[skin->get_bind_name(i)];
-							if (bn) {
-								skin->set_bind_name(i, bn);
+							HashMap<String, String>::ConstIterator bn_itr = p_rename_map.find(skin->get_bind_name(i));
+							if (bn_itr != p_rename_map.end()) {
+								skin->set_bind_name(i, bn_itr->value);
 							}
 						}
 					}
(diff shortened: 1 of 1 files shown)


AThousandShips:
While it shouldn't crash the fact that it didn't already use a permissive check makes me think this should cause an error instead and abort importing

tdaven:
I suspect that would change previous behavior where it would rename/remap the bones it couldn't match and not the others which might break peoples import workflows.

AThousandShips:
No previous behaviour would change though would it? It'd just crash before, I mean specifically when the original crash condition occurs?

tdaven:
Prior to 96a95cb9743e755f10fa6a14895ea099e5ac4c51 this worked just fine and didn't crash either.

AThousandShips:
Wasn't aware of that part, but that was a case where it shouldn't have manipulated the map anyway so no intentional behaviour was changed

The map is intentionally filled with valid entries and then previously copied, so the values weren't supposed to be added I'd say

Id suspect the previous copy instead of passing as const was an oversight, which was exposed by that change

But I'm not sure, we'll see what the relevant teams say 🙂

tdaven:
I suspect we don't want to warn.  I partially only added them to confirm it was working but hey are a bit too verbose.  Maybe only in dev builds or if at all.

tdaven:
Adding MRP to help in getting this merged at some point.

[bone-remap-mrp.zip](https://github.com/godotengine/godot/files/14313790/bone-remap-mrp.zip)

fire:
@TokageItLab if you want to review.

tdaven:
I don't believe t his PR is needed anymore.  The merging of  #81746 included the same type of changes(using the find methods on the hashmap).

akien-mga:
Thanks for the contribution!

PR #87913, opened 2024-02-03
Minor fix in DisplayServer docs to include Linux & Windows in `FEATURE_NATIVE_DIALOG`

OverloadedOrama:
Makes it consistent with the rest of the documentation, which states that native file dialogs are supported in Windows and Linux. Tested it myself on Linux X11, `DisplayServer.has_feature(DisplayServer.FEATURE_NATIVE_DIALOG)` does indeed return `true`.

Not sure if cherry-pickable to 4.2 because of lack of Wayland support. Let me know if I should open another PR targeted at the 4.2 branch that does not mention Wayland.


bruvzg (approved):

Mickeon (approved):

deralmas (approved):
Good catch!

akien-mga:
Thanks!

PR #87914, opened 2024-02-03
Fix `Label`'s shadow and outline overlapping

### LINKED ISSUE #35961: [3.2] Outlines go around shadows on (non-RichText) Labels
**Godot version:**
3.2

Windows 10
AMD FX-8320
GTX 1060 6GB

**Issue description:**
3.2 caused (non-RichText) Label nodes with an outline and a shadow to have the outline to go around the shadow, instead of the outline going around the text and the shadow going underneath the outline.

RichTextLabel nodes still have expected functionality. If you open the project in 3.1 then regular Labels also work as expected.

outline text bug

**Steps to reproduce:**
Create a (regular, non-RichText) Label with a shadow and outline.

[outline_text_bug.zip](https://github.com/godotengine/godot/files/4167682/outline_text_bug.zip)

WhalesState:
Fixes: https://github.com/godotengine/godot/issues/35961

Before.
image

After.
image

### DIFF
--- a/scene/gui/label.cpp
+++ b/scene/gui/label.cpp
@@ -316,16 +316,17 @@ inline void draw_glyph(const Glyph &p_gl, const RID &p_canvas, const Color &p_fo
 	}
 }
 
-inline void draw_glyph_outline(const Glyph &p_gl, const RID &p_canvas, const Color &p_font_color, const Color &p_font_shadow_color, const Color &p_font_outline_color, const int &p_shadow_outline_size, const int &p_outline_size, const Vector2 &p_ofs, const Vector2 &shadow_ofs) {
+inline void draw_glyph_outline(const Glyph &p_gl, const RID &p_canvas, const Color &p_font_outline_color, const int &p_outline_size, const Vector2 &p_ofs) {
 	if (p_gl.font_rid != RID()) {
-		if (p_font_shadow_color.a > 0) {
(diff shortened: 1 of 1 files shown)


scene/gui/label.cpp
```diff
@@ -431,7 +436,6 @@ void Label::_notification(int p_what) {
 			int outline_size = has_settings ? settings->get_outline_size() : theme_cache.font_outline_size;
 			int shadow_outline_size = has_settings ? settings->get_shadow_size() : theme_cache.font_shadow_outline_size;
 			bool rtl = (TS->shaped_text_get_inferred_direction(text_rid) == TextServer::DIRECTION_RTL);
-			bool rtl_layout = is_layout_rtl();
```
bruvzg, line 434: What's the purpose of removing this and `size`, these values can't change, so there's no reason to call functions multiple times.

bruvzg (commented):
Instead of having 3 nearly identical loops, it probably worth moving all the code (ellipsis and draw parts) to the helper function and calling it for the shadow, outline, and text.

WhalesState:
> Instead of having 3 nearly identical loops, it probably worth moving all the code (ellipsis and draw parts) to the helper function and calling it for the shadow, outline, and text.

I'm sorry if it's not the best, moving them to a helper function will be complicated for me, I may need some help.

scene/gui/label.cpp
WhalesState: I thought it's the same since they are just getters. if you want to revert this change I will do it.

WhalesState:
This was already fixed

PR #87922, opened 2024-02-04
Fix a regression when multiple nodes are selected and drag transform is attempted w/o gizmo

ryevdokimov:
#87769 fixes a regression but introduces another. I misunderstood `get_single_selected_node` and thought it would still return a node if multiple were selected node, but it actually returns false and ends up breaking transforms done via dragging outside the gizmo. This PR replaces that method with a check if more than 0 nodes are selected.


akien-mga (approved):

akien-mga:
Thanks!

PR #82792, opened 2023-10-04
Fix export mesh lib break due to imported gltf has a redundant Node3D

jsjtxietian:
Fixes #81850

The code is a little messy this way, but imho it's better to change the mesh lib export code than change the gltf import code. I would appreciate any feedback or alternative suggestions.

PR #87923, opened 2024-02-04
Make MeshLibrary export do recursive depth-search for MeshInstance3D nodes

### LINKED ISSUE #81850: Mesh Libraries Breaking Due To A Redundant Node3D Being Generated Upon Asset Import.
### Godot version

v4.1.1.stable.mono.official [bd6af8e0e]

### System information

Godot v4.1.1.stable.mono - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3080 (NVIDIA; 31.0.15.3713) - AMD Ryzen 9 7950X 16-Core Processor (32 Threads)

### Issue description

Mesh Libraries are breaking because of a redundant node being generated within the packed scenes created after importing any 3D assets.

Had to make assets local the scene and delete the bad node to make Mesh Libraries work.

Screenshot 2023-09-18 160736

Screenshot 2023-09-18 1542312

Screenshot 2023-09-18 1550432

Screenshot 2023-09-18 1556192

Screenshot 2023-09-18 1559102

### Steps to reproduce

Video showing off the issue and how to reproduce it. Uploaded to YouTube because file size is too large for Github.

https://youtu.be/yebIvO_z0-0

N/A

### LINKED ISSUE #85085: Mesh Library Export silently ignores some meshes

smix8:
Makes MeshLibrary export do recursive depth-search for MeshInstance3D nodes.

Fixes https://github.com/godotengine/godot/issues/85085
Fixes https://github.com/godotengine/godot/issues/81850
Likely other issues as it has been a long-standing issue with the MeshLibrary exporter.

Would supersede https://github.com/godotengine/godot/pull/82792 and https://github.com/godotengine/godot/pull/87651 that increases the number of searched child tiers to a hard-coded higher number.

The current MeshLibrary export has the limitation of only searching 2 node child tiers.

In the wild there are often convoluted files (e.g. GLTF import) with far more nested child nodes so those mesh items would not be detected and exported.

This pr now makes the export do a recursive depth-search until a MeshInstance3D is encountered. At this point it stops searching further children of that node and instead proceeds as normal. It will still process all sibling nodes for more items, just stop the search for the children of that specific MeshInstance3D node.

This means users can still add supporting nodes in their export scenes below their mesh instances that are not exported. E.g. the source geometry to bake the navigation mesh that gets exported. This also means very convoluted and exaggerated tree setups like this now can work in the MeshLibrary export:

meshlib_tree_export

As far as compatibility is concerned I think the only way that things break with this recursive search is if users had a MeshInstance3D in their scenes that was already not exported correctly. If exported now this would obviously change the id numbers of the MeshLibrary. I would assume that any dev worth their salt already fixed their export scenes if they encountered such an export bug with a missing mesh so this is imo only a hypothetical concern and they can easy fix it by moving that suspicious node at the end of the tree so it receives an unused id.

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -2026,8 +2026,11 @@ void EditorNode::_dialog_action(String p_file) {
 		} break;
 
 		case FILE_EXPORT_MESH_LIBRARY: {
+			bool merge_with_existing_library = file_export_lib_merge->is_pressed();
+			bool apply_mesh_instance_transforms = file_export_lib_apply_xforms->is_pressed();
+
 			Ref<MeshLibrary> ml;
-			if (file_export_lib_merge->is_pressed() && FileAccess::exists(p_file)) {
+			if (merge_with_existing_library && FileAccess::exists(p_file)) {
 				ml = ResourceLoader::load(p_file, "MeshLibrary");
 
 				if (ml.is_null()) {
@@ -2040,7 +2043,7 @@ void EditorNode::_dialog_action(String p_file) {
 				ml = Ref<MeshLibrary>(memnew(MeshLibrary));
 			}
 
-			MeshLibraryEditor::update_library_file(editor_data.get_edited_scene_root(), ml, true, file_export_lib_apply_xforms->is_pressed());
+			MeshLibraryEditor::update_library_file(editor_data.get_edited_scene_root(), ml, merge_with_existing_library, apply_mesh_instance_transforms);
 
 			Error err = ResourceSaver::save(ml, p_file);
 			if (err) {
--- a/editor/plugins/mesh_library_editor_plugin.cpp
+++ b/editor/plugins/mesh_library_editor_plugin.cpp
@@ -78,108 +78,7 @@ void MeshLibraryEditor::_import_scene(Node *p_scene, Ref<MeshLibrary> p_library,
 	HashMap<int, MeshInstance3D *> mesh_instances;
 
 	for (int i = 0; i < p_scene->get_child_count(); i++) {
-		Node *child = p_scene->get_child(i);
-
-		if (!Object::cast_to<MeshInstance3D>(child)) {
-			if (child->get_child_count() > 0) {
-				child = child->get_child(0);
-				if (!Object::cast_to<MeshInstance3D>(child)) {
-					continue;
-				}
-
-			} else {
-				continue;
-			}
-		}
-
-		MeshInstance3D *mi = Object::cast_to<MeshInstance3D>(child);
-		Ref<Mesh> mesh = mi->get_mesh();
-		if (mesh.is_null()) {
-			continue;
-		}
-
-		mesh = mesh->duplicate();
-		for (int j = 0; j < mesh->get_surface_count(); ++j) {
-			Ref<Material> mat = mi->get_surface_override_material(j);
-
-			if (mat.is_valid()) {
-				mesh->surface_set_material(j, mat);
-			}
-		}
-
-		int id = p_library->find_item_by_name(mi->get_name());
-		if (id < 0) {
-			id = p_library->get_last_unused_item_id();
-			p_library->create_item(id);
-			p_library->set_item_name(id, mi->get_name());
-		}
-
-		p_library->set_item_mesh(id, mesh);
-
-		if (p_apply_xforms) {
-			p_library->set_item_mesh_transform(id, mi->get_transform());
-		} else {
-			p_library->set_item_mesh_transform(id, Transform3D());
-		}
-
-		mesh_instances[id] = mi;
-
-		Vector<MeshLibrary::ShapeData> collisions;
-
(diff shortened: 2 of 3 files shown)


reduz (approved):
seems good

akien-mga:
Thanks!

WebFreak001:
how would one make a group of disabled/non-importing meshes now? My use-case: I have a large group of imported scenes that I didn't inherit (and don't want to inherit manually anyway) and use a `@tool` script to duplicate the meshes from each scene into my single MeshLibrary scene + then auto-generate trimesh colliders, so that I don't have to do it for the hundreds of them manually.

Previously I would have been able to just have all my source meshes/scenes inside 1/2 deep nodes to skip them but now that no longer seems to be possible.

Would a quick hot-fix to skip over nodes that have e.g. process_mode set to disabled be accepted for that or should I create a full proposal?

citrasusanto:
Hello, iam using Godot 4.2.2 in Pop OS 22.04 and i still experience the silent mesh not being added into the mesh libraries.

https://kenney.nl/assets/tower-defense-kit

I downloaded the assets from here. assets from (28/08/2019)
If i downloaded manually remove the node tempParent, i showed up after exported however there are some errors on the other tile such as tile with tree where the tree is being ignored so only the tile showed up after exported.

image

Sorry if the fix will be patched for Godot 4.3 above because iam not sure when this will be patched.

martinzizi:
Hello, iam using Godot 4.3 and i still experience the silent mesh not being added into the mesh libraries.

PR #87925, opened 2024-02-04
Use SDL for joypad input on Linux

### LINKED ISSUE #9000: Switch to SDL as the main gamepad input driver

Project Heartbeat, a rhythm game

The current input system isn't in the best condition, it fails to account for many exceptions and quirks that other engines have no problems handling (UE for example uses SDL for input).

Some examples include filtering out steam virtual gamepads even when ran from outside steam and better guessing of button bindings for unknown controllers.

Adding support for SDL will not only make Godot work more consistently, but will also decrease the workload on contributors, as SDL already handles those quirks and edge cases and is constantly being updated.

As we would use a very small fraction of the API, any potential breaking changes from upstream can be quickly dealt with.

(before you ask, the SDL controller db we already use isn't enough to overcome these issues).

While it might be tempting to keep godot's NIH solution, years of experience making Project Heartbeat has shown me that this isn't quite as simple as simple as one might think. SDL is supported by Valve and other industry players, so it's a stable and very well behaved system.

SDL can be initialized with all subsystems but input being disabled.

I have plans to replace Godot's input system to SDL for use in Project Heartbeat soon, so I could take care of the work.

SDL already works on switch, but for platforms where SDL is not available yet we could just continue to use the current input drivers.

No

It's a change to the fundamental input system of the engine

### LINKED ISSUE #59250: Laptop touchpads are seen as gamepads
### Godot version

4.0-alpha4

### System information

Linux - Manjaro 21.2.5

### Issue description

In Godot 4.0-alpha4, my laptop's touchpads are seen as gamepads. When no actual controllers are attached to my laptop, functions and signals such as `Input.get_connected_joypads()` and `Input.joy_connection_changed.connect` see two existing devices. When I add real controllers, they are assigned device ids from 2 and up. In contrast, this is not the case in Godot 3.4.3.

**Godot 3 output at start**
`[blank]`
**Godot 3 output when I connect a Xbox 360**
`0 Xbox 360 Controller`

**Godot 4 output at start**
`0 SYNA8008:00 06CB:CE58 Touchpad`
`1 ELAN901C:00 04F3:2C29`
**Godot 4 output when I connect a Xbox 360**
`2 Xbox 360 Controller`

Note that if the Xbox controller is connected when I launch the "game", it is seen with device id = 0.

A tool such as gamepad-tool yields the same results as Godot 3 (no additional devices detected). 

### Steps to reproduce

Create an empty project with a single Node2D and the following script attached
```
extends Node2D

func _ready():
	Input.joy_connection_changed.connect(self._on_joy_connection_changed)
	for device in Input.get_connected_joypads():
		print(device, " ", Input.get_joy_name(device))

func _on_joy_connection_changed(device, connected):
	print(device, " ", Input.get_joy_name(device))
```

_No response_

EIREXE:
SDL is loaded in the same way other Linux system libraries are loaded by using a wrapped and dlopen.

Optionally, SDL can be dynamically linked into the binary.

Currently for Linux only since that platform direly needs it, but should be easy to make work on Windows once stable.

Proposal at: godotengine/godot-proposals#9000

P.S: I've made sure to strip the files as much as possible, but unless you want to be pulling your hair out by stripping files individually this is the best we can do IMO.

* *Bugsquad edit, closes godotengine/godot-proposals#9000 and closes https://github.com/godotengine/godot/issues/59250*


EIREXE:
> Does this PR support using a more recent SDL library if it's statically linked using an environment variable? See [#86180 (comment)](https://github.com/godotengine/godot/pull/86180#issuecomment-1919004562).

This PR doesn't do static linking at all, it instead either does normal linux shared linking (without use_sowrap) or dynamic loading (if use_sowrap is enabled).

As far as I understand, SDL_DYNAMIC_API is something SDL itself takes care of, not ourselves. (assuming the SDL we are linking or loading is built with dynamic api support that is).

EIREXE:
After some testing in production for Project Heartbeat I realised I made a mistake and was polling constantly, which is no bueno for CPU usage, so I switched to waiting with a timeout (so we can do the exit condition properly when the input thread is disposed of).

Calinou (commented):
Tested locally (rebased on top of `master` f040a351c2f27c6b86c40c71a28babf99dd0b9bd) on Fedora 39 X11, it works as expected. All controllers were tested in wired mode only.

In terms of out-of-the-box functionality, this PR is a big improvement for many reasons:

- My motherboard's LED controller is no longer recognized as the first controller ID (it's properly ignored). This meant the first connected controller was previously always on device ID 1, not 0. This PR effectively fixes https://github.com/godotengine/godot/issues/59250.
- Switch Pro Controller works out of the box with correct mappings on Linux in USB-C, whereas it was previously not recognized at all.
- A DualSense controller was previously seen as 2 devices (the second device was likely for motion reporting but was ineffective in Godot). This also broke controller IDs if you had multiple controllers connected if you had at least one DualSense and it wasn't the last connected controller. This PR fixes this and makes only one DualSense appear.
  - With this PR, the controller name also changed from `PS5 Controller` to `DualSense Wireless Controller` which is more accurate.
- With this PR, Xbox One V3 controller is now called `Xbox One S Controller` instead of `Xbox One Controller` which is also more accurate.

There are 2 issues I noticed though:

- Weak and strong vibration motors seem to be inverted, at least on the Xbox One V3 controller. Setting weak vibration to 0 and keeping strong vibration to 1 will result in a weak vibration with this PR, whereas it'd result in a strong vibration in `master`.
- I can't use <kbd>Ctrl + C</kbd> in a terminal to exit the project anymore with this PR. This occurs even if no controllers are ever connected before or during the project's runtime.

Controller UUIDs reported for reference:

Controller | `master` | This PR
-|-|-
DualSense | `030000004c050000e60c000011810000` | `4475616c53656e736520576972656c65`
Xbox One V3 | `030000005e040000ea02000001030000` | `58626f78204f6e65205320436f6e7472`
Switch Pro Controller | `030000007e0500000920000011810000` | `4e696e74656e646f2053776974636820`

Binary sizes (Linux x86_64 stripped release export template with LTO):

- `master`: 64,166,088 bytes
- This PR: 64,186,568 bytes (+20 KB)

20 KB sounds pretty reasonable considering how much this improves controller usability and paves the way for gyro support. Of course, this will be higher on platforms where we can't rely on system-provided SDL if we decide to use SDL for input on other platforms as well.

jlogostini:
I am uncertain if this is intended behavior or a bug ether way if I connected my Left Joycon via Bluetooth it's eating 2 device slots causing issues with split-screen input

`Gamepad 0 = Joy-Con (L)`
`Gamepad 1 = Joy-Con (L) (IMU)`
`Gamepad 2 =`
`Gamepad 3 =`
`Gamepad 4 =`
`Gamepad 5 =`

EIREXE:
> I am uncertain if this is intended behavior or a bug ether way if I connected my Left Joycon via Bluetooth it's eating 2 device slots causing issues with split-screen input
> 
> `Gamepad 0 = Joy-Con (L)` `Gamepad 1 = Joy-Con (L) (IMU)` `Gamepad 2 =` `Gamepad 3 =` `Gamepad 4 =` `Gamepad 5 =`

What version of SDL are you running in your OS? from what I can tell that issue was fixed in SDL two years ago.

jlogostini:
I am on fully updated Gentoo Linux and I am running SDL 2.30.2

bemug:
I just ran some tests.
As expected this makes my WiiU Gamecube adapter to work, as it is supported correctly in SDL. It won't work without this PR.

I agree with the discussion and think this offers huge benefits in term of controller support and code re-usability.
This feature is important for me. I will follow this PR closely.

Thank you for your work.

EIREXE:
> I just ran some tests. As expected this makes my WiiU Gamecube adapter to work, as it is supported correctly in SDL. It won't work without this PR.
> 
> I agree with the discussion and think this offers huge benefits in term of controller support and code re-usability. This feature is important for me. I will follow this PR closely.
> 
> Thank you for your work.

I've also been shipping it in [Project Heartbeat Phoenix](https://store.steampowered.com/app/1216230/Project_Heartbeat/) for a while without any major hiccup.

EIREXE:
Used `SDL_HINT_JOYSTICK_THREAD` flag, and changed the styling for setting `SDL_HINT_NO_SIGNAL_HANDLERS`

geowarin:
I confirm that this fixes the multiple problems my dualsense edge controller has with godot.
On the current master, multiple buttons are inverted and the trigger and joystick are also mixed up, which makes it unusable.

Vibrations work correctly and I'm able to CTRL+C to quit the application normally.
So the problems raised by @Calinou seem to have been fixed?

20kb size increase is totally acceptable to me given that controller support for me or my users would be broken without it.

Is this PR considered for 4.3?

akien-mga:
I'm generally favorable to exploring using SDL for gamepad input, as we've been playing catch up with them over the years and we're lacking in many areas.

There are a couple problems with the approach in this PR though, which will require further work before a solution is mergeable (not suggesting to do it now, this requires more discussion with the team first):
- If we start using SDL for input, we should aim to do it for all platforms, not just Linux. We also have significant issues on Windows and macOS that it might help address.
- This PR uses SDL as a system library, which assumes that SDL is available on all Linux systems. That's fine for Steam as it's part of the Steam Linux runtime, but it's not a reasonable assumption for general portable Linux binaries. If we want to use SDL, we will probably need to build and link it statically together with Godot.
- If we're building SDL and linking it statically for all platforms, we should assess how much of SDL we actually need to build/include. It might be that we just need to reuse hidapi, or their input code, without pulling in the whole framework.

ettiSurreal:
> Additionally, note that this PR falls back to the traditional Godot input driver if SDL is not present.

Something to note, there should be a way to communicate that the user is missing SDL in this case, otherwise they might blame the game for poor input (especially since the built-in input driver will likely become neglected after this). Also telling people to install dependencies on things like itch pages is often a major turnaway for people.

So I'd suggest just adding a checkbox in the export settings for bundling SDL with the game.

EIREXE:
Makes sense tbh, i'll work on static linking in a bit

mubinulhaque:
What's the status for this PR? Now that 4.4 has started, I'd like this to be merged early on to catch as many errors as possible before 4.4 reaches stable

Is it just static linking that needs to be worked on? Personally, I think it should be statically linked, just for those who don't have SDL installed on their system, since SDL comes with a dynamic loading system anyway.

Calinou:
> Is it just static linking that needs to be worked on?

See https://github.com/godotengine/godot/pull/87925#issuecomment-2116845337, which is still relevant.

EIREXE:
Rebased against master, still haven't had time to work on statically linked SDL, but will do so ASAP.

EIREXE:
What do you want me to do with the library? Should it be in thirdparty like all other builtin libraries?

EIREXE:
Nevermind, it's going to be big trouble to get it to compile inside godot, so we should just make static builds like we do with nir.

KeyboardDanni:
> * If we're building SDL and linking it statically for all platforms, we should assess how much of SDL we actually need to build/include. It might be that we just need to reuse hidapi, or their input code, without pulling in the whole framework.

Worth noting that SDL's buildsystem lets you avoid building any subsystems you're not interested in using. So we could just build the input subsystem and that should result in a smaller binary.

autione:
Would this also bundle new GDScript APIs for leveraging additional features for gamepads, such as DualSense's adaptive triggers, or Switch's gyro movements? It would be a missed opportunity to not implement those given the capabilities this PR provides.

Calinou:
> Would this also bundle new GDScript APIs for leveraging additional features for gamepads, such as DualSense's adaptive triggers, or Switch's gyro movements? It would be a missed opportunity to not implement those given the capabilities this PR provides.

To my knowledge, this is not planned as part of this PR, but it can be done in future PRs. See https://github.com/godotengine/godot-proposals/issues/2829 and https://github.com/godotengine/godot/pull/88590.

QuinnLeavitt:
Not sure how much work it would take to transition this to SDL 3 but [SDL 3.2 stable has been released.](https://github.com/libsdl-org/SDL/releases/tag/release-3.2.0)

vaartis:
Can confirm that this works correctly for the DualSense (PS5) controller. The issue of having both the touchpad and the gamepad itself as two separate controllers is gone, now there is only the controller.

Nintorch:
I've noticed that a person named "xsellier" mentioned this PR in their commit about adding SDL3 input support for their fork of Godot, and I've noticed that static linking of SDL hasn't yet been done by anyone so this inspired me to do that myself (but only for Windows for now, I will try to setup a Linux VM later to make sure Linux support works too).
I've grabbed the code from both EIREXE, xsellier (I hope they wouldn't mind...) and the source code of SDL3, modified the scons files to include just the required files and compile just the required subsystems and I've managed to compile Godot with SDL and sdl input driver.
I can confirm that this SDL input driver indeed works because when I print Input.get_joy_guid(0) on Godot without SDL driver it prints "030000004c050000cc09000000000000" for my Dualshock 4 gamepad and with SDL driver it prints "50533420436f6e74726f6c6c6572", and I've also tried to enable gamepad gyroscope support inside the JoypadSDL class (via SDL_SetGamepadSensorEnabled(game_controller, SDL_SENSOR_GYRO, true);) and query the gyroscope data (via SDL_EVENT_GAMEPAD_SENSOR_UPDATE event) and it does print out values for gyro, if I rotate my controller the values change, so SDL input must be working correctly.
I also had to modify some SDL files to correctly compile with MSVC compiler but I'm sure we can later make it compile without changing the SDL source code.
I hope my work would be useful for this PR.

https://github.com/Nintorch/godot/commit/4c58541a84f4ecd12e4dff417ed5d3a18ea2134d

Nintorch:
Also, I know this PR is only about adding SDL and not about adding gyroscope, etc., but when SDL input driver is merged I would love to add support for new gamepad features SDL provides such as gyroscope, accelerometer, touchpads and lightbars

Nintorch:
Sorry for triple post, but I've managed to add Linux support back to my fork. I've managed to compile it in WSL and it runs correctly, when I print my gamepad's guid "536f6e7920496e746572616374697665" is being printed.
https://github.com/Nintorch/godot/commit/41030e24429db9abde9d895068a3f15c306d942f
I also want to add MacOS support because it was requested in this thread by Calinou. I've never used either a Mac or Docker but I will try my best to setup MacOS inside of Docker if that's possible and legal because I can't afford a Mac :D
I will keep you guys updated

akien-mga:
Superseded by #106218. Thanks for the contribution!

PR #80272, opened 2023-08-04
GLTF: Improve logic for keeping track of the real root node

aaronfranke:
This PR improves the logic for keeping track of the real root node. There is now a method called `_generate_scene_node_tree` which is the parent method that generates a single root node, and generates nodes under it (starting with the GLTF root nodes). In the current master, the GLTF importer fishes out the real root node by grabbing a GLTF root node and then getting its parent, super janky.

PR #87933, opened 2024-02-04
Fix crash when importing a GLTF file with a skeleton as the root

aaronfranke:
This is an edge case, but I ran into a situation where the root node of an imported GLTF scene was trying to be generated as a skeleton. This is crashing in the current master. This PR fixes it, now the skeleton will check if there is no root node and if so it will make itself the root node.

Note: The `GLTFDocument::_process_mesh_instances` function used to have a `Node *p_scene_root` parameter, but it was not used before. In #80272 I removed it. This was not quite the right fix, it should've been used, not removed.


fire (approved):

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #87934, opened 2024-02-04
Fix GLTF exporting invalid meshes and attempting to export gizmo meshes

aaronfranke:
This PR has 2 related but technically distinct bug fixes:

* Fix GLTF exporting invalid meshes. In GLTF it is not allowed to have an empty mesh. The export code will now check for this and fail, instead of exporting something that did not comply to the glTF specification.

* Fix GLTF exporting trying to export non-owned nodes in the editor. These are created by tool scripts, such as custom gizmo code. These nodes are temporary, usually for visualization purposes, they are not saved to the Godot scene and should not be saved in the GLTF file. In the event that somebody does want a tool script to save to the scene persistently and be exported to GLTF, you just need to set the owner to the scene root.

    * Note: This only affects the editor. User code at runtime is likely not using the owner system and therefore for runtime exports we should include all nodes like we have already been doing.

    * Note: This adds a new macro, `WARN_VERBOSE`, to only show a warning when in verbose mode. Lyuma was concerned about spamming warnings for valid workflows, which makes sense.


core/error/error_macros.h
```diff
@@ -730,6 +730,16 @@ void _err_flush_stdout();
 	} else                                                                                        \
 		((void)0)
 
+/**
+ * Warns about `m_msg` only when verbose mode is enabled.
+ */
+#define WARN_VERBOSE(m_msg)               \
```
fire, line 736: I'm ok with this, but typically need an opinion from the core team about these additions to macros.

fire:
I'll try my test case from https://github.com/V-Sekai-fire/SK_skeleton_merge_cat_fire_merge_01/blob/main/SK_merge_tool_01/scenes/black_cat_fire_base_01.tscn when I have a animation attached to a vrm

fire:
I lose the face blend shapes.

https://github.com/V-Sekai/v-sekai-other-world/blob/main/tools/TOOL_godot_ik/ik_test.tscn

editor_screenshot_2024-02-06T073927


image

I posted the project. Will try to see where it breaks and if it's related to this pr.

fire:
BUG: the animation that is in the scene isn't associated with the export.

modules/gltf/gltf_document.cpp
```diff
@@ -5413,14 +5413,13 @@ BoneAttachment3D *GLTFDocument::_generate_bone_attachment(Ref<GLTFState> p_state
...
+	ERR_FAIL_COND_V_MSG(p_mesh_instance->get_mesh().is_null(), -1, "glTF: Tried to export a MeshInstance3D node named " + p_mesh_instance->get_name() + ", but it has no mesh. This node will be exported without a mesh.");
+	Ref<Mesh> mesh_resource = p_mesh_instance->get_mesh();
+	ERR_FAIL_COND_V_MSG(mesh_resource->get_surface_count() == 0, -1, "glTF: Tried to export a MeshInstance3D node named " + p_mesh_instance->get_name() + ", but its mesh has no surfaces. This node will be exported without a mesh.");
 
-	Ref<Mesh> import_mesh = p_mesh_instance->get_mesh();
-	Ref<ImporterMesh> current_mesh = _mesh_to_importer_mesh(import_mesh);
+	Ref<ImporterMesh> current_mesh = _mesh_to_importer_mesh(mesh_resource);
 	Vector<float> blend_weights;
-	int32_t blend_count = import_mesh->get_blend_shape_count();
+	int32_t blend_count = mesh_resource->get_blend_shape_count();
```
fire, line 5422: I wonder if the loss of the blend shapes is related to this code.

lyuma:
@fire The loss of blend shapes is happening during import due to a faulty property setter in vrm_secondary.gd (I see you have addons/vrm in your project).

The property setter causes a gizmo to be added during the import process, as an internal child (INTERNAL_MODE_FRONT) which triggers a Godot bug in Node::replace_by since it keeps the internal index but does not forward data.internal_mode to add_child.
Node::replace_by is buggy and is what ultimately causes the corruption during import in this case, but I fixed godot-vrm to not attempt to add gizmos in a property setter when not is_inside_tree(). that should mitigate the import issue at least.

Anyway, the bug you hit with the face which I describe above is during import and unrelated to this export PR. It just happens to have a similar cause. So this shouldn't block merging.

fire (approved):

akien-mga:
Cherry-picked for 4.2.2.

PR #87935, opened 2024-02-04
C#: Implement proper generic type name printing for Godot Editor

Delsin-Yu:
The current Godot Editor Implementation fetches a class_name directly from the `System.Type.Name`, which does not return the full definition of a generic type (only the generic parameter count).

This PR uses an algorithm created by @ZerxZ, @TML233, and @jinsediaoying for printing the actual underlying generic type definition.
Our [tests](https://github.com/Delsin-Yu/System.Type.PrettyPrint/blob/main/GenericCSharpProgram.Test/Tests.cs) have convinced us this algorithm is capable of handling all generic use cases, see the following comparison.
This only runs under `DEBUG` (fallback to `System.Type.Name` otherwise), so it should not affect runtime performance.

|Before|After|
|:---|:---|
|image|image|

These are the actual type header definitions in the scripts
```csharp
public partial class ListItemComponent<TData> : Control { /*other*/ }

public partial class ComplexCrazyListItemComponent<T1, T2, T3, T4> : ListItemComponent<(int, float, (List<Func<List<Dictionary<List<int?>, List<int?[,,]>>>, List<(List<int>, List<int[]>)?>>>, List<int[,,][,][,,,,]>))?> { /*other*/ }

public partial class Test<T> : ComplexCrazyListItemComponent<int, float, GodotObject, List<int>> { /*other*/ }

public partial class Test2 : Test<GodotObject> { /*other*/ }
```

### DIFF
--- a/platform/windows/display_server_windows.cpp
+++ b/platform/windows/display_server_windows.cpp
@@ -760,8 +760,7 @@ Ref<Image> DisplayServerWindows::clipboard_get_image() const {
 						pba.append(rgbquad->rgbReserved);
 					}
 				}
-				image.instantiate();
-				image->create_from_data(info->biWidth, info->biHeight, false, Image::Format::FORMAT_RGBA8, pba);
+				image = Image::create_from_data(info->biWidth, info->biHeight, false, Image::Format::FORMAT_RGBA8, pba);
 
 				GlobalUnlock(mem);
 			}


paulloz:
Apart from the couple of `while (true)` that I personally am not a fan of, it looks good to me.  

Maybe this should run in `#if TOOLS` instead of `#if DEBUG`?

raulsntos:
Both `#if TOOLS` and `#if DEBUG` won't work. User projects normally retrieve the `GodotSharp` assembly from NuGet.org and that will be compiled with the Release configuration and no `TOOLS` define. Since you tested this on a custom build of the engine, that will use the package from your local NuGet repository, and I assume you built `GodotSharp` with the Debug configuration.

Performance-wise I think the best we can do is check `Engine.IsEditorHint()` and hope that the JIT optimizes this away after a couple of executions.

Delsin-Yu:
This should work by now.

AThousandShips:
You erased your changes by your push and resetting the branch, which closes the issue, to restore and get back to where you were use:
```
git reset --hard fea4f15
```
And then try again

Delsin-Yu:
Oh, I accidentally made a pushed in the middle of rebasing my PR to the latest `master`, will update shortly.

raulsntos:
- Superseded by https://github.com/godotengine/godot/pull/88363.

PR #87917, opened 2024-02-03
Speed up filesystem dock operations

KoBeWi:
Fixes https://github.com/godotengine/godot/issues/50576

https://github.com/godotengine/godot/assets/2223172/6ffa2866-5105-4b5b-a748-c3b34550c375

Moving, duplicating, renaming and creating folders is now instant.

In fact it's just a trick - the Tree is updated instantly upon action, but the rescan still runs in the background. When it finishes, it causes Tree sync with the actual filesystem. There are still some bugs like files appearing in wrong place or wrong selections, so putting as draft.

PR #83715, opened 2023-10-21
Preserve focus during rescans in FileSystem dock

VedatGunel:
Fixes #33126

Certain operations in the FileSystem dock trigger a rescan. During rescans, a progress bar is being shown in place of FileSystemTree and FileSystemList, and the container that holds them (_split_box_) is hidden, causing them to lose focus:

https://github.com/godotengine/godot/blob/6543495b49613d20f7e32f2b9d38e4a2f1d06db1/editor/filesystem_dock.cpp#L1303-L1307

So, I added some containers to display the progress bar on top of them instead of completely hiding them.

PR #87936, opened 2024-02-04
Grab focus of renamed file in file system dock

bitbrain:
## Current problem

It is rather frustrating to rename a lot of files, as renaming a file removes focus from a file and one has to manually click the row again to then use the arrow keys to navigate to the next file.

## Solution

After a file is moved or renamed, grab focus of currently selected file so one does not have to use the mouse at all but can simply use arrow keys to rename things. This makes the edit/rename experience super fast!

https://github.com/godotengine/godot/assets/822035/1a6009a5-ebf5-49b5-82c3-247b422a75ba

### DIFF
--- a/editor/filesystem_dock.cpp
+++ b/editor/filesystem_dock.cpp
@@ -1791,6 +1791,11 @@ void FileSystemDock::_rename_operation_confirm() {
 
 	print_verbose("FileSystem: calling rescan.");
 	_rescan();
+
+	if (ti) {
+		tree->set_selected(ti);
+		tree->grab_focus();
+	}
 }
 
 void FileSystemDock::_duplicate_operation_confirm() {


VedatGunel:
I have a PR that fixes the same issue: https://github.com/godotengine/godot/pull/83715

AThousandShips:
For reviewers, see also:
* https://github.com/godotengine/godot/pull/83715

bitbrain:
This pull request could be treated as a temporary short-term fix until both https://github.com/godotengine/godot/pull/87917 and https://github.com/godotengine/godot/pull/83715 are sorted out (which might still take some time), although it would make more sense long-term to let `rescan()` not throw away focus rather than the hack I applied here.

AThousandShips:
Would need to be discussed if the issue is serious enough to justify a temporary fix, we normally avoid them unless necessary, but if it's serious enough it could be justified

KoBeWi:
Looks like the file now keeps the selection, but the dock loses focus. You could make the dock grab focus after rescan (if it had it before), which should fix the remaining problem.

akien-mga:
Superseded by #103734. Thanks for the contribution!

PR #87939, opened 2024-02-04
GDScript: Add `is not` operator

### LINKED ISSUE #1319: Allow usage of `is not` in GDScript
This issue works for all projects that use the keywords "in" and "is"

When trying to use the opposite of "in" and "is", most people, especially the people who have used Python before would use "a not in b" and "a is not b" instead of "not a in b" and "not a is b". In Python, both the former and latter example works, but in GDScript only the latter is valid syntaxes.

Just allow usage of "a not in b" and "a is not b", which not only are in Python but also make sense grammatically 

`a is not b #equivalent to not a is b`
`a not in b #equivalent to not a in b`

The keywords themselves should be used quite oftenly, but yes its easy to have a work around.

It's a simple quality of life improvement that is pretty straightforward to implement and helps with the workload a lot.

dalexeev:
* Closes godotengine/godot-proposals#1319.

```gdscript
func test():
    var s: Variant = "str"
    print(await s is not int)
    print(not(await s is int))
    print(await (s is not int))
```

```
Class <unnamed> :
|   Function test(  ) :
|   |   Variable s : Variant
|   |   |   = "str"

|   |   print( (NOTAwait s IS int) )
|   |   print( (NOTAwait s IS int) )
|   |   print( Await (NOTs IS int) )
```

https://github.com/godotengine/godot/blob/b4e2a24c1f62088b3f7ce0197afc90832fc25009/modules/gdscript/gdscript_parser.h#L1397-L1402

### DIFF
--- a/modules/gdscript/gdscript_parser.cpp
+++ b/modules/gdscript/gdscript_parser.cpp
@@ -3285,6 +3285,19 @@ GDScriptParser::ExpressionNode *GDScriptParser::parse_lambda(ExpressionNode *p_p
 }
 
 GDScriptParser::ExpressionNode *GDScriptParser::parse_type_test(ExpressionNode *p_previous_operand, bool p_can_assign) {
+	// x is not int
+	// ^        ^^^ ExpressionNode, TypeNode
+	// ^^^^^^^^^^^^ TypeTestNode
+	// ^^^^^^^^^^^^ UnaryOpNode
+	UnaryOpNode *not_node = nullptr;
+	if (match(GDScriptTokenizer::Token::NOT)) {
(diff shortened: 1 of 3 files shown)


Mickeon (approved):
Reminder that @vnen did [not even realize](https://chat.godotengine.org/channel/gdscript?msg=hK88wfJc4FHSzL5NH) this was never a thing.

adamscott (approved):
Nice and clean. Thanks @dalexeev !

vnen (approved):

akien-mga:
Thanks!

PR #87941, opened 2024-02-04
[GDScript] Prevent running `String` number functions on invalid literal

### LINKED ISSUE #83521: Invalid (or incomplete) hex number "0x" floods error messages
### Godot version

v4.2.beta1.official [b1371806a]

### System information

Godot v4.2.beta1 - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1060 6GB (NVIDIA; 31.0.15.3640) - Intel(R) Core(TM) i5-3570K CPU @ 3.40GHz (4 Threads)

### Issue description

When I start writing a hex number in GDScript code, parser starts flooding error messages when the hex number is not yet valid.

First a single error appears, but very soon there will be 2 or 3 more. Every save (Ctrl-S) adds 5 more error messages.

I would expect to get only one error message.

kuva

### Steps to reproduce

Write `var i = 0` first and then add `x`. 

```gdscript
extends Node2D

func _ready() -> void:
	pass
	var i = 0x
```

Output:

```
  Invalid hexadecimal notation character "x" (U+0078) in string "0x".
  Invalid hexadecimal notation character "x" (U+0078) in string "0x".
  ...
```

N/A

AThousandShips:
Prevents printing excessive errors.

We could also just remove the error print in `hex_to_int` but it was added quite a while ago and is a good indication of what's wrong when parsing strings IMO, and at least I don't think it should be removed just to prevent an annoying print in the GDScript parser (I'd even argue we might want to *add* errors to `bin_to_int` and the others)

Literals containing decimal points will already stop consuming input before them so that won't trigger an error print on that part (other than the GDScript tokenizer error, but the annoying part is the print to console)

* Fixes: https://github.com/godotengine/godot/issues/83521

### DIFF
--- a/modules/gdscript/gdscript_tokenizer.cpp
+++ b/modules/gdscript/gdscript_tokenizer.cpp
@@ -672,6 +672,7 @@ GDScriptTokenizer::Token GDScriptTokenizer::number() {
 	bool has_decimal = false;
 	bool has_exponent = false;
 	bool has_error = false;
+	bool need_digits = false;
 	bool (*digit_check_func)(char32_t) = is_digit;
 
 	// Sign before hexadecimal or binary.
@@ -686,11 +687,13 @@ GDScriptTokenizer::Token GDScriptTokenizer::number() {
 			// Hexadecimal.
 			base = 16;
 			digit_check_func = is_hex_digit;
+			need_digits = true;
 			_advance();
 		} else if (_peek() == 'b') {
 			// Binary.
 			base = 2;
 			digit_check_func = is_binary_digit;
+			need_digits = true;
 			_advance();
 		}
 	}
@@ -717,6 +720,7 @@ GDScriptTokenizer::Token GDScriptTokenizer::number() {
 			}
 			previous_was_underscore = true;
 		} else {
+			need_digits = false;
 			previous_was_underscore = false;
 		}
 		_advance();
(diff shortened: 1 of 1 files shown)


modules/gdscript/gdscript_tokenizer.cpp
```diff
@@ -820,6 +824,19 @@ GDScriptTokenizer::Token GDScriptTokenizer::number() {
 		}
 	}
 
+	if (need_digits) {
+		// No digits in hex or bin literal.
+		Token error = make_error(vformat(R"(Expected %s digit after "0%c".)", (base == 16 ? "hexadecimal" : "binary"), (base == 16 ? 'x' : 'b')));
+		error.start_column = column;
+		error.leftmost_column = column;
+		error.end_column = column + 1;
+		error.rightmost_column = column + 1;
```
AThousandShips, line 833: Unsure if the full error info is required here, but added it for completeness, unsure what the difference here is to not using it like below

modules/gdscript/gdscript_tokenizer.cpp
```diff
@@ -820,6 +824,19 @@ GDScriptTokenizer::Token GDScriptTokenizer::number() {
...
+		push_error(error);
+
+		// Return empty literal.
+		return make_literal((int64_t)0);
```
dalexeev, line 837:
```suggestion
		return error;
```

AThousandShips: Will do, it prints different errors about expecting an expression but makes sense to error out

Could return errors more here to keep it unified, what do you think?

dalexeev: I think `push_error()` is used when we want to continue reading the token despite an error (for example inside string literals) and `return make_error()`/`return error` when the current token has most likely ended.

AThousandShips: Could swap for the others then too if desired

modules/gdscript/gdscript_tokenizer.cpp
dalexeev: I think it is right. Such errors should be reported as a cursor rather than a selection.

dalexeev (approved):
Looks good to me. `push_error()` vs `return error` can be improved later (like lines 844-845 and 848).

akien-mga:
Thanks!

AThousandShips:
Thank you!

akien-mga:
Cherry-picked for 4.2.2.

PR #87943, opened 2024-02-04
Only recurse depth wise in `Tree::_count_selected_items`

0x0ACB:
The current implementation for `Tree::_count_selected_items` recurses on both depth and width. This very quickly leads to a stackoverflow. I noticed this in the integrated `FileDialog` which would crash when opening folders with a lot of files.

The new implementation only recurses on depth and iterates on width. Ideally it probably shouldn't recurse at all but I didn't know if std::stack was permitted to use and I guess a `Tree` shouldn't ever be deep enough to lead to a stack overflow here anyways.

### DIFF
--- a/scene/gui/tree.cpp
+++ b/scene/gui/tree.cpp
@@ -2587,12 +2587,8 @@ int Tree::_count_selected_items(TreeItem *p_from) const {
 		}
 	}
 
-	if (p_from->get_first_child()) {
-		count += _count_selected_items(p_from->get_first_child());
-	}
-
-	if (p_from->get_next()) {
-		count += _count_selected_items(p_from->get_next());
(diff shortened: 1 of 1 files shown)


AThousandShips (approved):
This looks correct to me, it's only called internally (and only once) and on the root, so no nodes will be missed, using iteration over recursion for this won't change the outcome as the order isn't relevant

On the topic of using a stack or similar there's always benefits and drawbacks to either, and I believe largely trees will be wide and not deep so it's low risk, and we do a lot of other processes here recursively so any changes would best be done in a bigger rework 🙂

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.

PR #87948, opened 2024-02-04
Fix new region select with selection when over geometry

### LINKED ISSUE #85736: Click and drag on any node in 3D incorrectly moves selected node

### LINKED ISSUE #87949: Select tool translates node when dragging w/o the gizmo over geometry
### Tested versions

v4.2.1.stable.mono.official [b09f793f5]
v4.3.dev.custom_build [b4e2a24c1]

### System information

Godot v4.2.1.stable.mono - Windows 10.0.22631 - Vulkan (Forward+) - integrated AMD Radeon(TM) Graphics (Advanced Micro Devices, Inc.; 31.0.24002.92) - AMD Ryzen 7 PRO 6850U with Radeon Graphics (16 Threads)

### Issue description

When attempting to region-select over geometry while a node is selected, and the select tool is enabled, the selected node ends up being translated, which seems wrong.

https://github.com/godotengine/godot/assets/105675984/6d39ec75-21bc-40e9-8e93-8a13e56d3496

### Steps to reproduce

See video.

Select node using select tool.

Attempt to region select over some kind other geometry like another mesh. 

N/A

ryevdokimov:
Fixes: #87949 Fixes #85736

You should be able to region select with the select tool even if you already have a node selected. All other tools already have the default behavior of remotely manipulating their respective gizmo when a node is selected. 

Before:

https://github.com/godotengine/godot/assets/105675984/a605187a-6167-4626-9aaf-f955ccfa41ed

After:

https://github.com/godotengine/godot/assets/105675984/8d56ad74-fc17-4e54-ae99-96ca4ac59bb8

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -1985,9 +1985,8 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 			} else {
 				const bool movement_threshold_passed = _edit.original_mouse_pos.distance_to(_edit.mouse_pos) > 8 * EDSCALE;
 
-				// enable region-select if nothing has been selected yet or multi-select (shift key) is active
-				if (selection_in_progress && movement_threshold_passed) {
-					if (get_selected_count() == 0 || clicked_wants_append) {
+				if (selection_in_progress && movement_threshold_passed && clicked.is_valid()) {
+					if (clicked_wants_append || !editor_selection->is_selected(Object::cast_to<Node>(ObjectDB::get_instance(clicked)))) {
 						cursor.region_select = true;
 						cursor.region_begin = _edit.original_mouse_pos;
 						clicked = ObjectID();


KoBeWi:
This breaks selected node dragging:

https://github.com/godotengine/godot/assets/2223172/c93f2277-c002-45e3-b265-09ae00aca745

Dragging should still move a node when the clicked node is part of the current selection.

ryevdokimov:
Good catch. Working on it...

ryevdokimov:
Should be fixed now. Had to add code to check whether or not the node(s) trying to be dragged are the ones that are selected.

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -1985,9 +1985,19 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
...
+					const HashMap<Node *, Object *> &selection = editor_selection->get_selection();
+
+					bool node_in_selection_clicked = false;
+
+					for (const KeyValue<Node *, Object *> &E : selection) {
+						if (E.key->get_instance_id() == clicked) {
+							node_in_selection_clicked = true;
+							break;
+						}
+					}
```
AThousandShips, line 1998: I think this can be done more efficiently, instead of iterating over it, I think this should be enough: `editor_selection->is_selected(Object::cast_to<Node>(ObjectDB::get_instance(clicked)));`

ryevdokimov: Test it. It sure is.

KoBeWi (approved):

editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -1985,9 +1985,10 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 			} else {
 				const bool movement_threshold_passed = _edit.original_mouse_pos.distance_to(_edit.mouse_pos) > 8 * EDSCALE;
 
-				// enable region-select if nothing has been selected yet or multi-select (shift key) is active
-				if (selection_in_progress && movement_threshold_passed) {
-					if (get_selected_count() == 0 || clicked_wants_append) {
+				if (selection_in_progress && movement_threshold_passed && clicked.is_valid()) {
+					bool node_in_selection_clicked = editor_selection->is_selected(Object::cast_to<Node>(ObjectDB::get_instance(clicked)));
+
+					if (!node_in_selection_clicked) {
```
AThousandShips, line 1991:
```suggestion
					if (!node_in_selection_clicked || clicked_wants_append) {
```
Maybe keep the shift behaviour, haven't tested but was in the original

ryevdokimov: This is a really good point and would have been a regression, good catch.

ryevdokimov:
Suggestions applied and tested. Thanks for help.

AThousandShips (approved):

Cammymoop:
It seems that both this approach and the one in #85740 could be desirable, probably due to the select mode tool wearing too many hats. I would prefer having the old functionality from before 3.5 restored (selecting and moving a node with a single click) but I could see myself using this style too.

Either way it looks like there's conflicting expectations on how the select tool should work in this instance, aside from the agreement that the current behavior is a bug.

akien-mga:
Thanks!

@Cammymoop This might warrant opening a proposal to discuss what should be the desired behavior, or whether it needs to be made configurable. If the behavior changed in 3.5 / 4.0 and most users are now familiar with the current behavior as being the intended one, changing it back would likely be a problem for more users than the 3.5 / 4.0 change was.

PR #87952, opened 2024-02-04
Disable signal callback generation in C#

### LINKED ISSUE #18721: (MONO) New signal method declared outside of class
**Godot version:**
3.0.3 rc1 Mono Build

**Issue description:**
When using the editor to connect a signal to a C# script, the new method declaration is added to the end of the file, and outside of the class.  It should instead add the new method just before the final closing brace.

**Steps to reproduce:**
1. Create a button node. 
2. Attach a new C# script to this node.
3. Select the button in the editor, choose Node -> Signals
4. Select the "pressed()" signal and click connect.
5. Choose the same button as the "Node to Connect." Leave everything default and hit connect.
6. Notice that the new method is created outside of the class.

Example:
```cs
using Godot;
using System;

public class btn_move : Button
{
    public override void _Ready()
    {
    }
} //End of class


private void _on_Button_pressed()
{
    // Replace with function body
}
```

This is a low priority issue as the easy workaround is to just move the closing brace of the class, but this creates invalid code and could be confusing.

paulloz:
A while ago, we talked about disabling the signal callback generation altogether for C# until we have a better implementation. Reasons are, it currently generates the method outside the class (creates compilation error, obviously), and the behaviour is weirdly inconsistent depending on how you have external editors configured.

To prevent confusion like "the dialog didn't do the thing", I added a small message in the connection window, depending on the language of the script. Not sure if this is wanted, let me know (also, the message itself probably kinda sucks, feel free to suggest something better).

godot windows editor dev x86_64 mono_HnMDcxVcl2

*Bugsquad edit:*
- Works around / fixes #18721.

### DIFF
--- a/core/object/script_language.h
+++ b/core/object/script_language.h
@@ -256,6 +256,7 @@ public:
 	virtual bool can_inherit_from_file() const { return false; }
 	virtual int find_function(const String &p_function, const String &p_code) const = 0;
 	virtual String make_function(const String &p_class, const String &p_name, const PackedStringArray &p_args) const = 0;
+	virtual bool can_make_function() const { return true; }
 	virtual Error open_in_external_editor(const Ref<Script> &p_script, int p_line, int p_col) { return ERR_UNAVAILABLE; }
 	virtual bool overrides_external_editor() { return false; }
 
--- a/editor/connections_dialog.cpp
+++ b/editor/connections_dialog.cpp
@@ -178,6 +178,7 @@ void ConnectDialog::_tree_node_selected() {
 		set_dst_method(generate_method_callback_name(source, signal, current));
 	}
 	_update_method_tree();
+	_update_warning_label();
 	_update_ok_enabled();
 }
 
@@ -433,6 +434,23 @@ void ConnectDialog::_update_ok_enabled() {
 	get_ok_button()->set_disabled(false);
 }
 
+void ConnectDialog::_update_warning_label() {
+	Ref<Script> scr = source->get_node(dst_path)->get_script();
+	if (scr.is_null()) {
+		warning_label->set_visible(false);
+		return;
+	}
+
+	ScriptLanguage *language = scr->get_language();
+	if (language->can_make_function()) {
+		warning_label->set_visible(false);
+		return;
+	}
+
+	warning_label->set_text(vformat(TTR("%s: Callback code won't be generated, please add it manually."), language->get_name()));
+	warning_label->set_visible(true);
+}
+
 void ConnectDialog::_notification(int p_what) {
 	switch (p_what) {
 		case NOTIFICATION_ENTER_TREE: {
@@ -617,6 +635,7 @@ void ConnectDialog::init(const ConnectionData &p_cd, const PackedStringArray &p_
 
 void ConnectDialog::popup_dialog(const String p_for_signal) {
 	from_signal->set_text(p_for_signal);
+	warning_label->add_theme_color_override("font_color", warning_label->get_theme_color(SNAME("warning_color"), EditorStringName(Editor)));
 	error_label->add_theme_color_override("font_color", error_label->get_theme_color(SNAME("error_color"), EditorStringName(Editor)));
 	filter_nodes->clear();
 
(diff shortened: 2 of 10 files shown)


core/object/script_language.h
```diff
@@ -256,6 +256,7 @@ class ScriptLanguage : public Object {
 	virtual bool can_inherit_from_file() const { return false; }
 	virtual int find_function(const String &p_function, const String &p_code) const = 0;
 	virtual String make_function(const String &p_class, const String &p_name, const PackedStringArray &p_args) const = 0;
+	virtual bool can_make_function() { return true; }
```
AThousandShips, line 259: I'd suggest adding a virtual method for this to `ScriptLanguageExtension` so we don't forget about it

paulloz: Double-checking before I do something wrong. We're talking about adding an `EXBIND`, and a `GDVIRTUAL_BIND` in `ScriptLanguageExtension`, right?

AThousandShips: Yep! There should be good examples of `bool` returns there that can be copied

zaevi (approved):
Looks good to me👍

editor/connections_dialog.cpp
```diff
@@ -433,6 +434,23 @@ void ConnectDialog::_update_ok_enabled() {
...
+		return;
+	}
+
+	ScriptLanguage *language = scr->get_language();
+	if (language->can_make_function()) {
+		warning_label->set_visible(false);
+		return;
+	}
+
+	warning_label->set_text(vformat(TTR("%s: callback code won't be generated, please write it yourself."), language->get_name()));
```
AThousandShips, line 450:
```suggestion
	warning_label->set_text(vformat(TTR("%s: Callback code won't be generated, please add it manually."), language->get_name()));
```
Think this might feel more friendly

AThousandShips (approved):
LGTM code wise

Calinou (approved):
Until we find a better workaround for this (such as displaying the code you should copy in a read-only TextEdit), this is probably for the better.

raulsntos (commented):
I feel a bit hesitant about adding new API to ScriptLanguage. I don't know if there are Script languages other than C# that would make use of this, adding this just for C# knowing that C# will eventually move away from the ScriptLanguage API feels wrong.

But if the area maintainers are fine with this, then I do think this is an improvement over the current state for C#.

paulloz:
> I feel a bit hesitant about adding new API to ScriptLanguage. I don't know if there are Script languages other than C# that would make use of this, adding this just for C# knowing that C# will eventually move away from the ScriptLanguage API feels wrong.

I had a similar feeling. We could simply always return an empty string and not display anything to the user. It'd be a bit worse UX, but still better than the current state, IMHO. In that case, I'd still modify the `add_callback` logic to not overwrite the file if `make_function` returns a string of length zero...

modules/mono/csharp_script.cpp
```diff
@@ -516,22 +516,10 @@ static String variant_type_to_managed_name(const String &p_var_type_name) {
...
-			s += ", ";
-		}
-
-		s += variant_type_to_managed_name(arg.get_slice(":", 1)) + " " + escape_csharp_keyword(arg.get_slice(":", 0));
-	}
-	s += ")\n{\n    // Replace with function body.\n}\n";
-
-	return s;
+	// TODO
+	// Implement function generation. This has been disabled for now because the behaviour was not consistent
```
akien-mga, line 520: We use en_US spelling:
```suggestion
	// Implement function generation. This has been disabled for now because the behavior was not consistent
```

akien-mga (approved):
I think it's fine to add this method. It does feel very specific, but `find_function` and `make_function` do too and they're already part of the API. Having a boolean check to control the user of those two methods doesn't seem too far fetched.

modules/mono/csharp_script.cpp
paulloz: Noted. I'm going to re-write this comment as per @raulsntos' point [above](https://github.com/godotengine/godot/pull/87952#discussion_r1478874974) anyway.

raulsntos (approved):
LGTM. Thanks!

akien-mga:
Thanks!

dayo05:
I have a question about this: is this breaks custom signal registration by code as [introduced on docs](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_signals.html#custom-signals-as-c-events)?

paulloz:
> I have a question about this: is this breaks custom signal registration by code as [introduced on docs](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_signals.html#custom-signals-as-c-events)?

Hello! No, this only blocks the editor to write the empty method in your `.cs` file when you connect a signal to an inexistent method from Godot's UI.

PR #87954, opened 2024-02-04
[macOS] Do not show file type popup in the native file dialog if there's only one option, improve `*.*` filter handling.

### LINKED ISSUE #87867: Native FileDialog shows single option filtering popup if format filtering restrictions set
### Tested versions

4.2.1

### System information

macOS 14.3

### Issue description

FileDialog supports setting up filtering using [add_filter()](https://docs.godotengine.org/en/latest/classes/class_filedialog.html#method-descriptions) that can be used to restrict supported/allowed formats both when saving or loading. This works as expected as far as picking files, however, at least with the current implementation for native file dialogs the way this is shown in the interface is incorrect in my opinion. 

Example of a save dialog with only one allowed format:
Screenshot 2024-02-02 at 13 54 27

This should not show the selection button for picking format, as only one is allowed. In other words, it should only be shown if there's more than one possible save format to select (more than one filter?)

Example of open dialog with filter specifying allowed formats:
Screenshot 2024-02-02 at 13 53 47

This version shows a button "Show options" that when toggled shows this same one-item format picking button that allows you to pick which filter to apply. I'm not sure if there is a use-case for when you have multiple filtering options and want to allow switching between them like this, but at least for the case when there is just one filtering set for allowed file types I think this selector (and the entire options panel) should not be shown as there are no options to pick between.

While I realize this might not be a high priority issue, I think the fix would be very easy: Since the selectors are correctly hidden when there are no filters set, it should just be a matter of instead hiding them if there are less than two.
As indicated above this is just something I've seen on macOS native dialogs, I'm not sure if the behavior is the same on other platform native dialogs.

### Steps to reproduce

Open a native file dialog on macOS with filtering set. For example using:

```
var browseDialog:FileDialog = FileDialog.new()
browseDialog.title = "Select image"
browseDialog.initial_position = Window.WINDOW_INITIAL_POSITION_CENTER_MAIN_WINDOW_SCREEN
browseDialog.file_mode = FileDialog.FILE_MODE_OPEN_FILE
browseDialog.use_native_dialog = true
	
browseDialog.add_filter("*.svg,*.jpg","Supported formats")
browseDialog.access = FileDialog.ACCESS_FILESYSTEM
add_child(browseDialog)
browseDialog.popup()
```

optionally trying the save version with a single supported format:
```
browseDialog.file_mode = FileDialog.FILE_MODE_SAVE_FILE
browseDialog.add_filter("*.jpg","Supported formats")
```

And observe the native file dialog interface.

Code above should be sufficient. I do not believe this is dependent on other project settings.

bruvzg:
- Fixes https://github.com/godotengine/godot/issues/87867
- Fixes issues with `*.*` filter handling.

Note: Change for macOS only, on Linux and Windows file type selector is an integral part of dialog layout, and can't be removed even if there's nothing to select.

### DIFF
--- a/platform/macos/godot_open_save_delegate.mm
+++ b/platform/macos/godot_open_save_delegate.mm
@@ -103,7 +103,7 @@
 	}
 
 	NSMutableArray *new_allowed_types = [[NSMutableArray alloc] init];
-	bool allow_other = false;
+	bool has_type_popup = false;
 	{
 		NSTextField *label = [NSTextField labelWithString:[NSString stringWithUTF8String:RTR("Format").utf8().get_data()]];
 		if (@available(macOS 10.14, *)) {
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #87959, opened 2024-02-04
Add means for fixing navmap synchronization errors

### LINKED ISSUE #85548: "Navigation map synchronization error" caused internally (by recast?)
### Godot version

4.1.3.stable.official

### System information

Godot v4.1.3.stable - Arch Linux #1 SMP PREEMPT_DYNAMIC Tue, 10 Oct 2023 21:10:21 +0000 - X11 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 970 (nvidia; 535.113.01) - AMD Ryzen 5 2600 Six-Core Processor (12 Threads)

### Issue description

For context, I'm working on [navmesh support for the Terrain3D plugin](https://github.com/TokisanGames/Terrain3D/pull/253). To support navmeshes, the plugin generates 'source geometry data' for navigable portions of the terrain, and uses that as input to the engine's `bake_from_source_geometry_data` function.

I very frequently run into this error:

```
E 0:00:00:0665   sync: Navigation map synchronization error. Attempted to merge a navigation mesh polygon edge with another already-merged edge. This is usually caused by crossing edges, overlapping polygons, or a mismatch of the NavigationMesh / NavigationPolygon baked 'cell_size' and navigation map 'cell_size'.
  <C++ Source>   modules/navigation/nav_map.cpp:856 @ sync()
```

It's strange because I have verified that:
- `cell_size` and `cell_height` on the navmesh match the navigation map values.
- The mesh contains no crossing edges or overlapping polygons.
- The vertices on the mesh are not especially dense. They're all spaced out wider than cell_size/cell_height.

I've created a minimal reproducer that reproduces this error. The reproducer creates a parabolic U-shaped mesh and bakes a navmesh from it:
image

I _suspect_ that the issue is all these long-thin polygons from the edge of the mesh to the middle:
image

However, if I turn on wireframe, you can see that those long-thin polygons are not part of the source mesh, which is very regular:
image

Since these long polygons originate in the engine (or the recast library?) I think that this is a bug. If these long/dense edges are an error, then the engine shouldn't be creating them.

### Steps to reproduce

1. Download and open the attached project.
2. Open the `nav_bake_issue.tscn` scene. A tool script will run on _ready to generate the mesh and bake it.
3. Run the scene, and observe the logged error.

[nav_bake_issue.zip](https://github.com/godotengine/godot/files/13514658/nav_bake_issue.zip)

Scony:
Fixes https://github.com/godotengine/godot/issues/85548

This PR is based on my observation that for some valid input geometries, recast produces valid navigation meshes and yet the transformation that is done as part vertex hashing:
https://github.com/godotengine/godot/blob/b4e2a24c1f62088b3f7ce0197afc90832fc25009/modules/navigation/nav_map.cpp#L106-L108 is making those navigation meshes invalid. That is because some vertices are conflicting with others after transformation above and thus leading to synchronization errors mentioned in https://github.com/godotengine/godot/issues/85548

This PR proposes a map-level option to change the scale of internal rasterizer which allows the engine to perform the align-to-recast-grid transformation using greater precision. With that, in projects where navigation baking leads to synchronization problems, the precision of internal rasterizer can be increased to avoid errors (although potentially sacrificing the ability to connect separate regions).

I've prepared a straightforward demo ([godot-navigation-demos.zip](https://github.com/godotengine/godot/files/14232811/godot-navigation-demos.zip) - `ComplexNavmeshPaths.tscn`) application that showcases the problem - all it does is query the server for paths from random point A to random point B each and render them to show potential holes in navigation.

The demo outcome when using current Godot implementation:
2024-02-11-153336_4480x1440_scrot

The demo outcome when the Godot from this PR is used and `merge_rasterizer_cell_scale` is set to `0.001`:
2024-02-11-152839_4480x1440_scrot

### DIFF
--- a/modules/navigation/nav_map.cpp
+++ b/modules/navigation/nav_map.cpp
@@ -67,6 +67,7 @@ void NavMap::set_cell_size(real_t p_cell_size) {
 		return;
 	}
 	cell_size = p_cell_size;
+	_update_merge_rasterizer_cell_dimensions();
 	regenerate_polygons = true;
 }
 
@@ -75,6 +76,16 @@ void NavMap::set_cell_height(real_t p_cell_height) {
 		return;
 	}
 	cell_height = p_cell_height;
+	_update_merge_rasterizer_cell_dimensions();
+	regenerate_polygons = true;
+}
+
+void NavMap::set_merge_rasterizer_cell_scale(float p_value) {
+	if (merge_rasterizer_cell_scale == p_value) {
+		return;
+	}
+	merge_rasterizer_cell_scale = p_value;
+	_update_merge_rasterizer_cell_dimensions();
 	regenerate_polygons = true;
 }
 
@@ -103,9 +114,9 @@ void NavMap::set_link_connection_radius(real_t p_link_connection_radius) {
 }
 
 gd::PointKey NavMap::get_point_key(const Vector3 &p_pos) const {
-	const int x = static_cast<int>(Math::floor(p_pos.x / cell_size));
-	const int y = static_cast<int>(Math::floor(p_pos.y / cell_height));
-	const int z = static_cast<int>(Math::floor(p_pos.z / cell_size));
+	const int x = static_cast<int>(Math::floor(p_pos.x / merge_rasterizer_cell_size));
+	const int y = static_cast<int>(Math::floor(p_pos.y / merge_rasterizer_cell_height));
+	const int z = static_cast<int>(Math::floor(p_pos.z / merge_rasterizer_cell_size));
 
 	gd::PointKey p;
 	p.key = 0;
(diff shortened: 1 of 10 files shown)


modules/navigation/nav_map.cpp
```diff
@@ -103,9 +111,9 @@ void NavMap::set_link_connection_radius(real_t p_link_connection_radius) {
 }
 
 gd::PointKey NavMap::get_point_key(const Vector3 &p_pos) const {
-	const int x = static_cast<int>(Math::floor(p_pos.x / cell_size));
-	const int y = static_cast<int>(Math::floor(p_pos.y / cell_height));
-	const int z = static_cast<int>(Math::floor(p_pos.z / cell_size));
+	const int x = static_cast<int>(Math::round(p_pos.x / (cell_size * rasterizer_scale_factor)));
+	const int y = static_cast<int>(Math::round(p_pos.y / (cell_height * rasterizer_scale_factor)));
+	const int z = static_cast<int>(Math::round(p_pos.z / (cell_size * rasterizer_scale_factor)));
```
smix8, line 116: This `cell_xyz * rasterizer_scale_factor` should be calculated once and cached in some internal var imo.

Changing from floor() to round() will affect existing projects and break compatibility so should be reverted. It also does not change things in a good way as it shifts cells around less predictable when there is high vertex density. To solve merge issues with that there is already the scale factor added.

Mickeon (commented):
Docs look good enough, I just cannot speak on the feature itself.

smix8 (approved):

akien-mga:
Thanks!

PR #87961, opened 2024-02-04
Add NavigationPolygon `border_size` property for tile baking

smix8:
Adds NavigationPolygon `border_size` property for tile baking. Also adds `baking_rect` Rect2 bounds.

This is a 2D version of https://github.com/godotengine/godot/pull/87378.

This allows users to partition larger game worlds into regions that have navigation meshes with easy to align tile edges. This tile edges can be efficiently merged by the navigation map using edgekeys. It also avoids problems with the agent_radius offset affecting the tile edges.

Since 2D had no baking AABB like 3D this also adds a baking Rect2 for 2D to the NavigationPolygon that allows to limit the baked area with a rectangle. This rect also clips the outline polygons upfront before the major polygon operations so using this can help to avoid a lot of unnecessary processing by discarding all polygons parts outside this rect.

The difference between baking Rect and border size is, that the Rect limits the entire source geometry used in the baking as a bounding area while the border size limits the surface of the resulting navigation mesh.

This means in conjunction they can be used to limit the final navigation mesh without having unwanted offsets by e.g. agent radius at tile edges.

nav_mesh2_border_size

E.g. this allows to split a large Tilemap into source geometry chunks and bake chunk navigation meshes that have border edges aligned so they can be merged efficiently by the navigation map without using the edge connection margin.

So more advanced users can start to only bake navigation mesh chunks around their active actors instead of for the entire TileMap. (Re)baking a large TileMap as one big navigation mesh is basically performance impossible at runtime and also not very efficient in general.

tilebaking

This also adds a few smaller performance improvements like reserve() for the larger std::vectors for a small but still noticeable speedup on large maps.

Note that while the functions accepts a float and Rect2 value the final value used by the baking when all positions and additions are added up will be rasterized to an integer grid.

### DIFF
--- a/scene/resources/navigation_polygon.cpp
+++ b/scene/resources/navigation_polygon.cpp
@@ -347,6 +347,15 @@ real_t NavigationPolygon::get_cell_size() const {
 	return cell_size;
 }
 
+void NavigationPolygon::set_border_size(real_t p_value) {
+	ERR_FAIL_COND(p_value < 0.0);
+	border_size = p_value;
+}
+
+real_t NavigationPolygon::get_border_size() const {
+	return border_size;
+}
+
 void NavigationPolygon::set_parsed_geometry_type(ParsedGeometryType p_geometry_type) {
 	ERR_FAIL_INDEX(p_geometry_type, PARSED_GEOMETRY_MAX);
 	parsed_geometry_type = p_geometry_type;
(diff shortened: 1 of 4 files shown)


scene/resources/navigation_polygon.cpp
```diff
@@ -347,6 +347,15 @@ real_t NavigationPolygon::get_cell_size() const {
 	return cell_size;
 }
 
+void NavigationPolygon::set_border_size(real_t p_value) {
+	ERR_FAIL_COND(p_value < 0.0);
+	border_size = p_value;
```
Scony, line 352: shouldn't we `emit_changed();` here as well?

smix8: I did not add a changed signal for this property because all the other properties that do not need to update the inspector or the internal navigation mesh or some debug have a changed signal.

Scony: Sounds good then!

Scony (approved):

akien-mga (approved):
Not familiar enough with navigation to assess the feature, but the code and docs look good. Trusting smix8 and Scony on the usefulness of this new API.

akien-mga:
Thanks!

PR #87964, opened 2024-02-05
[4.1] Fix crash when importing a GLTF file with a skeleton as the root

aaronfranke:
This PR backports https://github.com/godotengine/godot/pull/87933 to Godot 4.1. This was not trivially cherry-pick-able due to conflicts, so I am opening this PR to backport it. In general I think it's worth backporting fixes like these when it is not too much work to do so, such as whenever the fix does not depend on missing refactors only present in master.


akien-mga (approved):

akien-mga:
Thanks!

PR #77594, opened 2023-05-28
Implement 2D instance shader parameters

KoBeWi:
Closes #62943

I wrote this a while ago and decided to rebase and test it again now. This seems to work the same as in 3D (no wonder, since it's a 1:1 copy-paste), i.e. the parameters appear in the inspector, they have correct default value etc.

The only problem is that changing the parameter has no effect. It always uses the default for the given type (not even the defined default). But it's probably the only missing bit, so if anyone can help, the feature would be finished.

PR #80458, opened 2023-08-09
Replace use of push constants in 2D RD renderer with uniform buffer

SlugFiller:
This removes the 128 byte limitation on push constant size, by only using the push constant as an index into the generated buffer. This allows adding more per-instance per-draw command data.

PR #87966, opened 2024-02-05
Implement 2D instance shader parameters

### LINKED ISSUE #62943: Instance uniforms are not implemented for 2D shaders
### Godot version

4.11 alpha

### System information

Windows 11

### Issue description

Instance keyword makes hints disappear and makes GDscript helpless in controlling the value in any way [It defaults to 0, as for float]; 
Delete instance keyword, and it works again.

Untitled

### Steps to reproduce

shader:
```
shader_type canvas_item;
instance uniform float opacity :hint_range(0.0, 1.0, 0.1)=0.5;

void fragment(){
	vec4 color = texture(TEXTURE,UV);
	color.a-=opacity;
	COLOR = color;
}
```
sprite with an area2d and collisionshape:
```
extends Sprite2D


func _on_area_2d_area_entered(area):
	material.set_shader_param("opacity",1.0)
```

EIREXE:
Rebased from @KoBeWi 's PR #77594 

I also fixed it, so it actually works, although I am not entirely sure the code I wrote is fine, if someone from rendering could comb over it I'd be thankful.

Original PR text:

Closes #62943

I wrote this a while ago and decided to rebase and test it again now. This seems to work the same as in 3D (no wonder, since it's a 1:1 copy-paste), i.e. the parameters appear in the inspector, they have correct default value etc.

The only problem is that changing the parameter has no effect. It always uses the default for the given type (not even the defined default). But it's probably the only missing bit, so if anyone can help, the feature would be finished.

### DIFF
--- a/scene/main/canvas_item.cpp
+++ b/scene/main/canvas_item.cpp
@@ -527,6 +527,59 @@ int CanvasItem::get_light_mask() const {
 	return light_mask;
 }
 
+const StringName *CanvasItem::_instance_shader_parameter_get_remap(const StringName p_name) const {
+	StringName *r = instance_shader_parameter_property_remap.getptr(p_name);
+	if (!r) {
+		String s = p_name;
+		if (s.begins_with("instance_shader_parameters/")) {
+			StringName name = s.replace("instance_shader_parameters/", "");
+			instance_shader_parameter_property_remap[p_name] = name;
+			return instance_shader_parameter_property_remap.getptr(p_name);
+		}
+		return nullptr;
+	}
+	return r;
+}
+
+bool CanvasItem::_set(const StringName &p_name, const Variant &p_value) {
+	const StringName *r = _instance_shader_parameter_get_remap(p_name);
+	if (r) {
+		set_instance_shader_parameter(*r, p_value);
+		return true;
+	}
+	return false;
+}
+
+bool CanvasItem::_get(const StringName &p_name, Variant &r_ret) const {
+	const StringName *r = _instance_shader_parameter_get_remap(p_name);
+	if (r) {
+		r_ret = get_instance_shader_parameter(*r);
+		return true;
+	}
+
+	return false;
+}
+
+void CanvasItem::_get_property_list(List<PropertyInfo> *p_list) const {
+	List<PropertyInfo> pinfo;
+	RS::get_singleton()->instance_item_get_shader_parameter_list(get_canvas_item(), &pinfo);
+	for (PropertyInfo &pi : pinfo) {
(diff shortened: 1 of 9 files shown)


EIREXE:
Okay, I just realised I didn't actually get it to work, I think a redesign of canvas.glsl is needed because there's no space in the push constant to fit an instance ID, if it's fine i'll take a look at making it work like scene.glsl does.

arkology:
> there's no space in the push constant to fit an instance ID

I dont know much about rendering, but you are talking about this https://github.com/godotengine/godot/pull/80458, right?

EIREXE:
> > there's no space in the push constant to fit an instance ID
> 
> I dont know much about rendering, but you are talking about this #80458, right?

Yes, exactly, lol, i'll just wait for that to get merged then

clayjohn:
This is a great first step towards adding support for instance shader parameters. 

The missing piece is mapping the instance id to the global array. Here is what that looks like in the Forward+ renderer:
https://github.com/godotengine/godot/blob/a9bb8509f2faac81bdb995c6c89a5347372f3498/servers/rendering/renderer_rd/forward_clustered/scene_shader_forward_clustered.cpp#L728

As you pointed out, there isn't room for that extra parameter in the current approach because push constants have such a limited size. That being said https://github.com/godotengine/godot/pull/80458 isn't really a full solution either as we will run into cache alignment issues if we start using a SSBO but then use 132 bytes for each instance

Invertex:
> As you pointed out, there isn't room for that extra parameter in the current approach because push constants have such a limited size. That being said #80458 isn't really a full solution either as we will run into cache alignment issues if we start using a SSBO but then use 132 bytes for each instance

What about moving something less important from the push constant to the instance buffers? 
In the case of canvas items, does `specular_shininess` need to be there? That feels like something that should be part of shaders that have specular enabled, creating a shininess per-instance variable. (and can't imagine many people are using specular on their canvases in the first place...)
With that removed, there'd be room for the instance index, keeping the struct 128 bytes.

But I don't have much of a grasp on how this part of the engine works, just wondering!

kakatoto-collab:
Hello, any news about that ? Instance uniforms are a must for anyone making a vampire survivor's like game.

mk56-spn:
https://github.com/godotengine/godot/pull/99230

Obsolete PR seemingly. 

Sidenote : is there someone to tag for these specifically ?

paddy-exe:
Superseded by https://github.com/godotengine/godot/pull/99230 and can therefore be closed. 2D shader instance parameters are coming in 4.4.

PR #87969, opened 2024-02-05
Use meshes instead of colliders for positioning drag to instantiate in 3D

ryevdokimov:
This is a proof of concept because it probably breaks compatibility, and the original solution may be more desirable in some circumstances. 

If anyone is interested in it, feel free to build off it.  

Some ideas: 

Account for the bounds for both target and preview, so it doesn't sink into the target.
Create a mechanism to select positioning based off mesh or collisions.
Use a modifier key to ignore meshes/collisions.
Use this code to snap existing meshes to other existing meshes. 

Currently the engine uses a physics-based ray to get an intersection of colliders to decide where to position the preview node. This comes with downsides such as depending on this physics tick rate, not snapping to meshes without colliders which is arguably the more intuitive behavior, and the code not being really straight forward as it needs to handle physics being ran on separate thread.   

This PR uses the same cull ray that node selections use to detect a target mesh, and then calculates the intersection using the Visual Instance AABB.

https://github.com/godotengine/godot/assets/105675984/249967aa-fcab-4c7e-abc0-c141f18341fa

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -2950,21 +2950,6 @@ void Node3DEditorViewport::_notification(int p_what) {
 			}
 		} break;
 
-		case NOTIFICATION_PHYSICS_PROCESS: {
-			if (!update_preview_node) {
-				return;
-			}
-			if (preview_node->is_inside_tree()) {
-				preview_node_pos = spatial_editor->snap_point(_get_instance_position(preview_node_viewport_pos));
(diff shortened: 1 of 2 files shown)


Calinou:
Does this mean the geometry's shape is no longer followed? This is problematic for placing objects in large complex scenes where entire rooms are a single mesh.

ryevdokimov:
Pretty much yeah, it will only snap to the AABB. It definitely isn't a replacement for the collider method in its current state, even though code is written that way. However, it does make it easier to test, and maybe a better strategy or use-case can be thought up of.

ryevdokimov:
I've found better ways to do similar things.

PR #87971, opened 2024-02-05
[TextServer / Font] Do not duplicate images to prevent unnecessary embedding.

bruvzg:
Fixes https://github.com/godotengine/godot/issues/82254

Note: this only works if images were imported and loaded using `load`/`preload`, in other cases images are still embedded (there's no other way).

Images for the imported BMFonts are embedded as well, since it can be packed and require conversion before used (we probably should add `ignore` import option, and auto set it for the BMFont images).

e.g.,:
```gdscript
@tool
extends FontFile
class_name MyFont

const FONT_SIZE = Vector2i(16, 16)

func _init():
	clear_cache()
	
	var image = load("res://SuperWhite.png")
	set_texture_image(0, FONT_SIZE, 0, image)
```

```tres
[gd_resource type="FontFile" script_class="MyFont" load_steps=3 format=3 uid="uid://df8eka62scrfs"]

[ext_resource type="Image" uid="uid://d1bu5rv3qc5oe" path="res://SuperWhite.png" id="1_q7u5v"]
[ext_resource type="Script" path="res://my_font.gd" id="2_dlquq"]

[resource]
cache/0/16/16/textures/0/offsets = PackedInt32Array()
cache/0/16/16/textures/0/image = ExtResource("1_q7u5v")
cache/0/16/0/ascent = 0.0
cache/0/16/0/descent = 0.0
cache/0/16/0/underline_position = 0.0
cache/0/16/0/underline_thickness = 0.0
cache/0/16/0/scale = 1.0
cache/0/16/0/kerning_overrides/16/16 = Vector2(0, 0)
cache/0/16/0/kerning_overrides/16/0 = Vector2(0, 0)
script = ExtResource("2_dlquq")
```

Issue #82254, opened 2023-09-24
FontFile's cached images are always duplicates of set images, and are bundled with resource when saved to disk.


Mickeon:
### Godot version

v4.1.1.stable.official [bd6af8e0e]

### System information

Windows 10

### Issue description

I am trying to make my own custom **FontFile**, by porting over an existing addon in 3.x called [Texture Fonts](https://github.com/ElectronicBlueberry/Texture-Fonts). After several days, it mostly works, surprisingly.

But there's one glaring, annoying flaw. When assigning an Image resource to the fon with `set_texture_image()`, a NEW **Image** is created, every time.
The reason as to why can be a problem is because when saving the custom **FontFile** as .tres, raw copies of the images _I already have saved on disk_ are bundled as sub-resources. Not just that, they're all saved as a **PackedByteArray** in pure, plain text. The bloat is ridiculous, and even if they were pure binary it'd be an unnecessary waste.

<img width="50%" src=https://github.com/godotengine/godot/assets/66727710/d7f2e53f-1d1c-4277-8ec2-68e31f370783/>

There is no easy way to replace this newly created **Image** with the original one. Even if you do, it will be overridden when the resource is saved again. I have tried numerous ways and gotten close at one point. If the Resource knew it was being saved, that may be doable.
This renders any way of doing what the original addon did impossible, without huge bloat.

<details>
<summary><h4>Click here for helevant piece of code(!?)</h4></summary>
https://github.com/godotengine/godot/blob/58f0cae4af47adcac121cc220749ddbf778f4a81/modules/text_server_adv/text_server_adv.cpp#L2606-L2626
</details>

It's possible that someone knowledgeable in this part of the font code may give more insight as to why this is the case.

### Steps to reproduce

The steps to reproduce this are a bit complex and I recommend taking a look at the MRP.
- Create a script extending **FontFile** with, say, the following piece of code:
    ```GDScript
    @tool
    extends FontFile
    class_name MyFont


    const FONT_SIZE = Vector2i(16, 16)

    func _init():
    	clear_cache()
	
    	var image := Image.load_from_file("res://icon.svg")
    	set_texture_image(0, FONT_SIZE, 0, image)
    ```
- Create a new **FontFile** resource on disk, and attach this script.
- Create a test scene with a **Label**, and assign this font Resource to the label.
- Open the **FontFile** resource with an external program and look inside. You may notice the `icon.svg` image has been saved, bundled with the resource, even though on the surface it's entirely unnecessary.

[...]

PR #87972, opened 2024-02-05
[Import] Add "skip file" import option to skip (and exclude from export) importable formats, auto set it for the images used by bitmap font.

bruvzg:
As mentioned in https://github.com/godotengine/godot/pull/87971 and https://github.com/godotengine/godot/issues/82254, bitmap fonts use external images, but can't use these images as external resources since in many cases image preprocessing is required.

This PR:
- Adds "skip file" import option to skip (and exclude from export) importable formats (e.g, images).
- Automatically set "skip file" import mode for the images used by bitmap font, to prevent exporting these images twice (as source Image/Texture2D end embedded processed image in the `fontdata`).

### DIFF
--- a/editor/import_dock.cpp
+++ b/editor/import_dock.cpp
@@ -48,7 +48,8 @@ public:
 	Ref<ResourceImporter> importer;
 	Vector<String> paths;
 	HashSet<StringName> checked;
-	bool checking;
+	bool checking = false;
+	bool skip = false;
 	String base_options_path;
 
 	bool _set(const StringName &p_name, const Variant &p_value) {
@@ -91,10 +92,6 @@ public:
 	void update() {
 		notify_property_list_changed();
 	}
-
-	ImportDockParameters() {
-		checking = false;
-	}
 };
 
 ImportDock *ImportDock::singleton = nullptr;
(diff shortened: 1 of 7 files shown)


editor/import_dock.cpp
```diff
@@ -153,7 +158,11 @@ void ImportDock::_add_keep_import_option(const String &p_importer_name) {
 	import_as->add_separator();
 	import_as->add_item(TTR("Keep File (No Import)"));
 	import_as->set_item_metadata(-1, "keep");
+	import_as->add_item(TTR("Skip File (No Import)"));
```
akien-mga, line 161: Should this be:
```suggestion
	import_as->add_item(TTR("Skip File (No Export)"));
```
?
Or "No Import nor Export" but it starts being quite lengthy.

akien-mga (approved):
Makes sense to me, the added "Skip File" option seems potentially useful in other cases (e.g. source `.blend` files when you already have a `.glb` for the same scene).

I see we're lacking documentation about the existing "Keep File", would be good to add details about both options in https://docs.godotengine.org/en/latest/tutorials/assets_pipeline/import_process.html#changing-import-resource-type

editor/import_dock.cpp
bruvzg: Not sure if `No Import` part is necessary, it's not adding any information. Maybe `Keep File (exported as is)` and `Skip File (not exported)`?

Zireael07:
>  (e.g. source .blend files when you already have a .glb for the same scene).

oh yeah please

editor/import_dock.cpp
bruvzg: Renamed options to: `Keep File (exported as is)` and `Skip File (not exported)`.

Also added some doc notes: https://github.com/godotengine/godot-docs/pull/9089

akien-mga:
Thanks!

PR #87977, opened 2024-02-05
wayland: Update to 1.22.0

Chubercik:
https://gitlab.freedesktop.org/wayland/wayland/-/releases/1.22.0

### DIFF
--- a/platform/linuxbsd/wayland/wayland_thread.cpp
+++ b/platform/linuxbsd/wayland/wayland_thread.cpp
@@ -986,6 +986,14 @@ void WaylandThread::_wl_surface_on_leave(void *data, struct wl_surface *wl_surfa
 	DEBUG_LOG_WAYLAND_THREAD(vformat("Window left output %x.\n", (size_t)wl_output));
 }
 
+// TODO: Add support to this event.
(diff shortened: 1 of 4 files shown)


akien-mga:
Thanks for looking into this.

Some API changes require downstream changes too: https://github.com/godotengine/godot/pull/87977/files#diff-c86c10b4f67151b187e2f364912792198ba5a3ef23346cfe43cb054b466b4c59

Chubercik:
At first I added short implementations that looked like this:

```c++
void WaylandThread::_wl_surface_on_preferred_buffer_scale(void *data, struct wl_surface *wl_surface, int32_t factor) {
	WindowState *ws = (WindowState *)data;
	ERR_FAIL_NULL(ws);

	ws->preferred_fractional_scale = factor;

	DEBUG_LOG_WAYLAND_THREAD(vformat("Window preferred buffer scale %d.\n", factor));
}
```

But given that this is way out of the scope I was prepared to tackle in this PR, I leaving them empty for the time being.

deralmas (approved):
Looks fine, thanks!

akien-mga:
Thanks!

PR #87978, opened 2024-02-05
Replace `mergeInto(LibraryManager.library...` for `addToLibrary(...`

adamscott:
[Since emscripten 3.1.45](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3145---082323), it is possible to use `addToLibrary` instead of `mergeInto` and having to pass `LibraryManager.library` as the first parameter.

> The function used to add symbols the JS library has been renamed from `mergeInto`, to the more specific `addToLibrary`.  This new function does not require the passing of `LibraryManager.library` as a first argument.  The old `mergeInto` continues to exist for backwards compat.

## Note to the merge team
As Godot currently builds with 3.1.39, this PR is to be merged **only when emscripten will be >= 3.1.45**.

### DIFF
--- a/modules/webrtc/library_godot_webrtc.js
+++ b/modules/webrtc/library_godot_webrtc.js
@@ -227,7 +227,7 @@ const GodotRTCDataChannel = {
 };
 
 autoAddDeps(GodotRTCDataChannel, '$GodotRTCDataChannel');
-mergeInto(LibraryManager.library, GodotRTCDataChannel);
+addToLibrary(GodotRTCDataChannel);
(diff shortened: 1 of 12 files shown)


akien-mga:
We usually aim to preserve support for older compilers, so I would only add this conditionally (falling back to the previous method) if there are significant gains. Otherwise it's better IMO to keep the code compatible with earlier 3.1.x releases.

adamscott:
Closing as per @akien-mga's previous comment.

PR #87979, opened 2024-02-05
Enable ALT + RMB to list select only on the select tool matching 2D and tooltips

### LINKED ISSUE #87924: Alt + RMB lets you list select in any tool mode in 3D
### Tested versions

v4.2.1.stable.mono.official [b09f793f5]
v4.3.dev.custom_build [677bd9783]

### System information

Windows 10.0.22631 - Vulkan (Forward+) - integrated AMD Radeon(TM) Graphics (Advanced Micro Devices, Inc.; 31.0.23013.1023) - AMD Ryzen 7 PRO 6850U with Radeon Graphics (16 Threads)

### Issue description

This is either a bug or needs a description for all the tools.

image

image

image

### Steps to reproduce

Select any tool that isn't the 'Select Tool' and ALT + RMB on a position with multiple nodes. 

N/A

ryevdokimov:
Fixes: #87924

### DIFF
--- a/editor/plugins/node_3d_editor_plugin.cpp
+++ b/editor/plugins/node_3d_editor_plugin.cpp
@@ -1659,7 +1659,7 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 					_edit.gizmo = Ref<EditorNode3DGizmo>();
 				}
 
-				if (_edit.mode == TRANSFORM_NONE && b->is_pressed()) {
+				if (b->is_pressed() && spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_SELECT) {
 					if (b->is_alt_pressed()) {
 						if (nav_scheme == NAVIGATION_MAYA) {
 							break;


editor/plugins/node_3d_editor_plugin.cpp
```diff
@@ -1661,7 +1661,7 @@ void Node3DEditorViewport::_sinput(const Ref<InputEvent> &p_event) {
 
 				if (_edit.mode == TRANSFORM_NONE && b->is_pressed()) {
 					if (b->is_alt_pressed()) {
-						if (nav_scheme == NAVIGATION_MAYA) {
+						if (nav_scheme == NAVIGATION_MAYA || spatial_editor->get_tool_mode() != Node3DEditor::TOOL_MODE_SELECT) {
```
AThousandShips, line 1664: Should probably be part of the check above instead of here IMO

AThousandShips:
I'd say that matching the existing (and seemingly old) behaviour with the tooltip is the wrong way to go, I'd say the tooltips should be improved instead

editor/plugins/node_3d_editor_plugin.cpp
ryevdokimov: The break needs to occur alt pressed when MAYA navigation is enabled. The break also occurring when the tool mode is not select is inconsequential.

AThousandShips: Why enter the block at all? Why not replace the `_edit.mode`?

ryevdokimov: I'm not sure I follow. Replace it with what?

AThousandShips: With the thing you're adding:
```cpp
if (spatial_editor->get_tool_mode() == Node3DEditor::TOOL_MODE_SELECT && b->is_pressed()) {
					if (b->is_alt_pressed()) {
						if (nav_scheme == NAVIGATION_MAYA) {
```

ryevdokimov: That would result in right clicking without alt opening a list select, which would be wrong.

AThousandShips: How would it do that? The alt check is still there?

ryevdokimov: I seriously need to wear glasses.

ryevdokimov: Ok yeah, but that ends up breaking the ALT + RMB list select when MAYA navigation is enabled.

ryevdokimov: Which is actually a bug outside this PR apparently.

AThousandShips: Breaking how? It's meant to not work in that mode, that's why the break is there

ryevdokimov: Yeah, you're right I edited my comments.

akien-mga:
Superseded by #87989.

PR #87981, opened 2024-02-05
Fix emscripten 3.1.51 breaking change about `*glGetProcAddress()`

adamscott:
[Since emscripten 3.1.51](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3151---121323), a link flag is now required in order to use `*glGetProcAddress()` family of functions.

This PR fixes that breaking change.

We use `*glGetProcAddress()` (relating to the web platform) in:
- `drivers/gles3/rasterizer_gles3.cpp`
- `drivers/gles3/storage/config.cpp`

A check is done to apply the link flag `-s GL_ENABLE_GET_PROC_ADDRESS` only if it detects emscripten ≥ 3.1.51.

### DIFF
--- a/platform/web/detect.py
+++ b/platform/web/detect.py
@@ -206,6 +206,11 @@ def configure(env: "Environment"):
         env.Append(LINKFLAGS=["-s", "USE_WEBGL2=1"])
         # Allow use to take control of swapping WebGL buffers.
         env.Append(LINKFLAGS=["-s", "OFFSCREEN_FRAMEBUFFER=1"])
+        # Breaking change since emscripten 3.1.51
+        # https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3151---121323
+        if cc_semver >= (3, 1, 51):
+            # Enables the use of *glGetProcAddress()
+            env.Append(LINKFLAGS=["-s", "GL_ENABLE_GET_PROC_ADDRESS=1"])
 
     if env["javascript_eval"]:
         env.Append(CPPDEFINES=["JAVASCRIPT_EVAL_ENABLED"])


akien-mga (approved):

akien-mga:
Thanks!

dsnopek:
Hm, I don't think we should be relying on `glGetProcAddress()` in our web builds? If we're not compiling without this, it probably means we're missing an `#ifdef WEB_ENABLED` somewhere

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

akien-mga:
Cherry-picked for 3.6.
Cherry-picked for 3.5.4.

dsnopek:
I came across this one again, because I'm debugging an unrelated issue concerning Emscripten 3.1.51.

I really think this PR was was the wrong way to go. As I [said previously](https://github.com/godotengine/godot/pull/87981#issuecomment-1935857897), we really shouldn't need `glGetProcAddress()` for the web, and if we are using it, I would consider that a bug since [it's very inefficient](https://emscripten.org/docs/optimizing/Optimizing-WebGL.html?highlight=glgetprocaddress):

Selection_166

If we chose _not_ to use the `-s GL_ENABLE_GET_PROC_ADDRESS` flag, then any time code tried to use it on the web, it would be a compiler error, which is a nice way to prevent the "bug" of using `glGetProcAddress()` on the web.

dsnopek:
Another note from [the Emscripten docs](https://emscripten.org/docs/tools_reference/settings_reference.html#gl-enable-get-proc-address) about how `-s GL_ENABLE_GET_PROC_ADDRESS=1` should be avoided:

Selection_167

dsnopek:
Rather than just complaining, I've started draft PR https://github.com/godotengine/godot/pull/93489 :-)

PR #87984, opened 2024-02-05
etcpak: Sync with upstream

BlueCube3310:
Follow-up to #86313 and #85842

Updates `etcpak` to the latest version and removes the Godot-specific patches.


akien-mga (approved):
Awesome :)

PR #87992, opened 2024-02-05
Add newer Variant types to `typed_array.h`

### LINKED ISSUE #87991: C++ MAKE_TYPED_ARRAY does not include newer Variant types
### Tested versions

4.2.1 tag, current master ( d3352813ea44447bfbf135efdec23acc4d1d3f89 )

### System information

N/A

### Issue description

`typed_array.h` has a number of lines that use the `MAKE_TYPED_ARRAY` macro for almost all `Variant` types. This was introduced back in 2020. But when the `VECTOR4`, `VECTOR4I`, and `PROJECTION` types were added, it seems that this list was never updated to include them.

https://github.com/godotengine/godot/blob/d3352813ea44447bfbf135efdec23acc4d1d3f89/core/variant/typed_array.h#L91-L132

This means that it is not possible to write something like this:

`TypedArray<Vector4>`

...even though it is possible to write this:

`TypedArray<Vector3>`

I also noticed that the `OBJECT` type was never included in this list. I can only presume this was an intentional omission...?

### Steps to reproduce

Try to compile `TypedArray<Vector4> my_array;`. Note that this results in this compile error: `typed_array.hpp(53): error C2039: 'get_class_static': is not a member of 'godot::Vector4'`

N/A

allenwp:
Fixes #87991

This intentionally omits `Object` for reasons mentioned in the discussion of the original issue.


fire:
If it's ok, you could leave a comment mentioning the omission intentionally

allenwp:
> If it's ok, you could leave a comment mentioning the omission intentionally with reasoning.

Good idea, I'll make that change.

Mickeon:
If the PR is a draft it cannot be considered ready for review, but the PR looks pretty "ready" to me. Are you sure this is still work in progress?

allenwp:
> If the PR is a draft it cannot be considered ready for review, but the PR looks pretty "ready" to me. Are you sure this is still work in progress?

Yes, sorry, just needed to head home for the day yesterday — wanted to actually build it and test it after adding the `INFO` lines before taking it out of draft. (I put it back into draft when I realized I forgot to update these `INFO` lines). I plan to have this back in review in about 4 or 5 hours when I’m back in the office and can build again :)

allenwp:
> No worries! There's no rush, of course. I just wanted to make sure this was not a mistake.

Thanks for the feedback! I'll make a note next time I leave something hanging -- I expect that it's not entirely unheard of for a PR to be left in a mergable state.

I don't think I've a good job testing the `MAKE_TYPED_ARRAY_INFO` lines, but they at least build fine. I'll move this back out of draft.

AThousandShips (approved):

akien-mga:
Thanks!

PR #88002, opened 2024-02-06
[3.x] Fix build on X11 following 64-bit detection changes

Calinou:
This was changed by https://github.com/godotengine/godot/commit/b0329fe8cb08a9f4cd3540b00683804d2b965a63 but an instance of `is64` was still left here. I had to make this change to get the `3.x` branch to build on my Linux x86_64 system.

~~Also, the old `is64` detection is still used in `platform/server/` – should we change it too?~~
**Edit:** Done. This also ports over the cross-compilation logic to the `server` platform, and allows Embree to be used in server tools builds on aarch64.

### DIFF
--- a/platform/server/detect.py
+++ b/platform/server/detect.py
@@ -82,9 +82,17 @@ def configure(env):
 
     ## Architecture
 
-    is64 = sys.maxsize > 2**32
+    # Cross-compilation
+    # TODO: Support cross-compilation on architectures other than x86.
+    host_is_64_bit = sys.maxsize > 2**32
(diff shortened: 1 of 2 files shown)


akien-mga:
Good catch, thanks!

> Also, the old `is64` detection is still used in `platform/server/` – should we change it too?

Yeah that would be good for consistency.

Calinou:
> Yeah that would be good for consistency.

Done :slightly_smiling_face:

akien-mga (approved):

akien-mga:
Thanks!

Issue #84125, opened 2023-10-28
Godot crashes when changing Editor Features when certain docks are floating


kitbdev:
### Godot version

master

### System information

Godot v4.2.beta (0ea75d77c) - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2070 (NVIDIA; 31.0.15.3598) - Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz (12 Threads)

### Issue description

When the import, filesystem, node, or history dock is floating, changing an editor feature crashes the engine.
Changing any setting triggers it.

This is because `EditorNode::_feature_profile_changed` assumes the docks are in a `TabContainer`, and if they are floating they aren't in one.

### Steps to reproduce

Make any of these docks floating: import, filesystem, node, or history
Open Editor Features (Editor -> Manage Editor Features...)
Make a profile and change a setting

### Minimal reproduction project

n/a

PR #84122, opened 2023-10-28
Fix dock visibility issues

kitbdev:
Fix dock visibility for Editor Features and distraction free mode.

- fixes https://github.com/godotengine/godot/issues/66834
- fixes #84125 crash when changing Editor Features with floating docks (import, filesystem, node, or history)
- fixes error when closing floating dock if there are less tabs than before
- fixes loading layouts in distraction free mode being very buggy (bottom dock disappears, side docks might open and be empty)
- fixes floating dock still opened even if it was disabled in Editor Features when loading layout

Updated `TabContainer`'s `move_tab_from_tab_container` to preserve all tab data.

The crash fix is a workaround, ideally it wouldn't have to close all floating docks. I am working on a different pr that should fix this in the future.

There were 4 separate implementations to update the dock slots visibility:
`update_dock_slots_visibility()`,  `_update_dock_containers()`, in `_load_docks_from_config()`, and `_dock_tab_changed()` (removed the lines affecting bottom docks for this one, see #39177).
I have consolidated them all in `update_dock_slots_visibility()`.

PR #88003, opened 2024-02-06
Refactor and fix issues in Editor Dock Manager

### LINKED ISSUE #74215: Dock position widget never disappears after changing position of a dock
### Godot version

4.0 stable

### System information

win 11

### Issue description

So if I click the three dots next to the docker and I choose another position, the docker moves away, but the popup stays where it is. Maybe I'm just dumb, but I thought that if I clicked on the popup again, it would move the new docker that's in the position where the old one used to be. But it doesn't work like that, it still moves the old one.

### Steps to reproduce

Click the three dots and move a docker.

N/A

### LINKED ISSUE #87234: Editor: Out of bounds error in one of TabBars when you have floating docks
### Tested versions

Reproduced with the current master, 9e65c5c0f4f8944d17fc7f5b05682206e9348d81.

### System information

Godot v4.3.dev (9e65c5c0f) - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3070 Ti Laptop GPU (NVIDIA; 31.0.15.3742) - 12th Gen Intel(R) Core(TM) i7-12700H (20 Threads)

### Issue description

If your editor layout has a floating dock, when you restart the editor you may be welcomed to this error:

```
scene\gui\tab_bar.cpp:643 - Index p_current = 0 is out of bounds (get_tab_count() = 0).
```

I tried to debug it, but this ended up being a deferred call, so I'm not entirely sure where it comes from. I quickly tried to find places that call this method deferred, but couldn't confirm any of them are responsible. So logging it here for now.

### Steps to reproduce

This is the layout that I have:

image

Make the Scene dock floating and then reload the project. Everything will restore correctly, but you will be greeted with an error in the output log.

N/a.

### LINKED ISSUE #88000: Crash on `Show in Filesystem` when FileSystem is at the bottom
### Tested versions

Godot v4.3.dev (10e111477) (since https://github.com/godotengine/godot/pull/86765)

### System information

Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 2070 (NVIDIA; 31.0.15.4665) - Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz (12 Threads)

### Issue description

- related https://github.com/godotengine/godot/pull/86765

The Editor will crash when `Show in Filesystem` is called from anywhere while the Filesystem is at the bottom panel.

`FileSystemDock::navigate_to_path()` assumes that the FileSystem is in a TabContainer if it is in the same window, but when it is at the bottom panel it is not, so the cast fails and is null when `set_current_tab()` is called on it.

I am working on a PR that will fix this issue.

### Steps to reproduce

- Move the FileSystem to the Bottom Panel
- Click on `Show in Filesystem` from the context menu on a Scene root, script in the script list, or resource in the inspector.
- The editor crashes.

N/A

kitbdev:
- Fixes https://github.com/godotengine/godot/issues/88000
- Fixes https://github.com/godotengine/godot/issues/87234
- Fixes https://github.com/godotengine/godot/issues/74215
- Fixes crash when disabling the Filesystem in Editor Features when it is at the bottom. (related: #84125)
- Fixes both `// FIXME: Find it, in a horribly inefficient way.` in `EditorDockManager::load_docks_from_config()`
- Fixes `// FIXME: Close all floating docks to avoid crash.` in `EditorNode::_feature_profile_changed()` from #84122
	- removed `close_all_floating_docks()`
- Fixes dock select popup drawing issues with scaling, sizes, and rtl.
- Fixes dock names being untranslated when moving to floating window or to bottom panel.
- Fixes clicking on the signals or group button in the Scene Tree Dock while the Node dock is floating shows errors.
- related https://github.com/godotengine/godot/pull/84193

#### Implementation Details

EditorDockManager
- Added `DockInfo` struct and a HashMap<Control *, DockInfo> of all docks to be able to easily access information on docks.
	- This replaces the `bottom_docks` Vector.
- Replaced `_edit_current()` with `_update_layout()` to handle all updating that needs to happen after any dock moves. Previously some actions like floating a dock wouldn't save the layout, so it wouldn't restore if a crash occurs.
- Docks disabled with Editor features are moved out of the dock and hidden, instead of using hidden tabs. They use `set_dock_enabled()` which sets a flag and closes or opens the dock.
- `close_dock()` will hide the dock and move it out of the tab container, so it doesn't affect anything and isn't seen in the dock menu. `open_dock()` will restore the dock to its previous location. These will also be useful for https://github.com/godotengine/godot-proposals/issues/8206
- Saving a closed disabled dock will still add it to a slot so we know where to restore it to after loading, just like floating windows.
- Saving bottom docks was changed to do the same so it is more like closed and floating docks. This means that previous layouts with a bottom dock will need to be resaved.
- Simplified and made private `_update_dock_slots_visibility()` because hidden tabs no longer need to be checked.
- Added `_move_dock()` to handle moving the dock, to make sure it is always properly handled when removed from the bottom panel or a floating window, and making sure the title is set when moving to a TabContainer.
- `focus_dock()` handles focusing on a dock no matter where it is.

Extracts the dock popup into its own class `DockContextPopup`. (https://github.com/godotengine/godot/pull/84193#discussion_r1452675069)
- It now operates on a dock Control, rather than indexes for the dock and bottom panel item.
- Added tooltips for the L/R move tab buttons.
- The docks are drawn now at the appropriate position when in rtl.
- Editor scale is now used for tab drawing.
  - Before (150%):
image
  - After (150%):
image
- Size issues when buttons were added/removed are fixed.
  - Before:
Screenshot 2024-02-05 172557
  - After:
image
- Docks at the bottom panel can now float directly (https://github.com/godotengine/godot/pull/86765#issuecomment-1904962229).
- Tabs can be right clicked to open the popup.

### DIFF
--- a/editor/editor_dock_manager.cpp
+++ b/editor/editor_dock_manager.cpp
@@ -49,8 +49,6 @@
 
 EditorDockManager *EditorDockManager::singleton = nullptr;
 
-static const char *META_TOGGLE_SHORTCUT = "_toggle_shortcut";
-
 void DockSplitContainer::_update_visibility() {
 	if (is_updating) {
 		return;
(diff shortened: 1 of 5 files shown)


kitbdev:
Changes:
- Removed the `layout_changed` signal and instead use `save_editor_layout_delayed()`. This prevents unnecessarily saving multiple times in a row. This saving is only really in case of a crash, and it doesn't need to happen immediately.
- `_dock_container_update_visibility()` replaces `_update_dock_slots_visibility()`, and it just updates the one tab container's visibility when it is needed, when a child is added or removed.
- `set_docks_visible()` handle updating the dock slots visibility itself.
- Connected dock slots to `active_tab_rearranged` to make sure layout is updated when dragging a tab within itself.
- Made sure `_update_layout()` only happens once after docks have been moved.
- Changed DockInfo `name` to `title` to be more accurate and better distinguish from when the Control's name is used instead.
- Fix Bottom Dock context menu position when not in fullscreen.
- Rebased.

Does `EditorNode::edit_current()` need to be called whenever the layout changes?
I have it there because it was being called when the layout changed in some cases, but I don't know if it is actually needed.

kitbdev:
Rebased to fix merge conflicts.
Shortcut is now saved in dock info instead of in the meta.
Also fixed to show translated dock title in floating window.

editor/editor_dock_manager.h
```diff
@@ -70,49 +74,51 @@ class EditorDockManager : public Object {
...
+
 	static EditorDockManager *singleton;
 
 	// To access splits easily by index.
 	Vector<DockSplitContainer *> vsplits;
 	Vector<DockSplitContainer *> hsplits;
 
-	Vector<WindowWrapper *> floating_docks;
-	Vector<Control *> bottom_docks;
+	Vector<WindowWrapper *> floating_windows;
```
KoBeWi, line 97: Why was this renamed? Windows are already "floating", the previous name was more precise. If anything, it could be `dock_windows`.

editor/editor_dock_manager.h
```diff
@@ -70,49 +74,51 @@ class EditorDockManager : public Object {
...
+	friend class DockContextPopup;
+
+	struct DockInfo {
+		String title;
+		bool open = false;
+		bool enabled = true;
+		bool at_bottom = false;
+		int previous_tab_index = -1;
+		bool previous_at_bottom = false;
+		WindowWrapper *floating_window = nullptr;
```
KoBeWi, line 86: Same thing about "floating window".

editor/editor_dock_manager.h
kitbdev: I changed it since it holds the Windows, not the docks themselves.
I'll change it to `dock_windows`, since that is more accurate.

KoBeWi (approved):
Looks great overall.

editor/editor_dock_manager.cpp
```diff
@@ -796,80 +675,307 @@ void EditorDockManager::register_dock_slot(DockSlot p_dock_slot, TabContainer *p
...
-	dock_vb->add_child(dock_hb);
+	closed_dock_parent = EditorNode::get_singleton()->get_gui_base();
+
+	dock_context_popup = memnew(DockContextPopup);
+	EditorNode::get_singleton()->get_gui_base()->add_child(dock_context_popup);
+}
+
+void DockContextPopup::_notification(int p_what) {
+	switch (p_what) {
+		case NOTIFICATION_THEME_CHANGED: {
```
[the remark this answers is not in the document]
kitbdev: Since `DockContextPopup` is a `PopupPanel` which is a `Window`, I don't have access to `NOTIFICATION_LAYOUT_DIRECTION_CHANGED`.

editor/editor_dock_manager.h
```diff
@@ -70,49 +74,51 @@ class EditorDockManager : public Object {
...
+	void _dock_container_update_visibility(TabContainer *p_dock_container);
+	void _update_layout();
 
 	void _dock_floating_close_request(WindowWrapper *p_wrapper);
-	void _dock_make_selected_float();
-	void _dock_make_float(Control *p_control, int p_slot_index, bool p_show_window = true);
-	void _restore_floating_dock(const Dictionary &p_dock_dump, Control *p_wrapper, int p_slot_index);
+	Control *_close_floating_window(WindowWrapper *p_wrapper);
+	void _dock_make_float(Control *p_dock, bool p_show_window = true);
+	void _restore_floating_dock(const Dictionary &p_dock_dump, Control *p_wrapper);
```
KoBeWi, line 114: Naming inconsistency.
floating window
floating dock

kitbdev: Here's what I'm thinking:

`_dock_floating_close_request` -> `_window_close_request`
`_close_floating_window` -> `_close_window`
`_dock_make_float` -> `_open_dock_in_window`
`_restore_floating_dock` -> `_restore_dock_to_saved_window` and change to only accept dock

akien-mga:
Thanks!

PR #88009, opened 2024-02-06
Fix converted 3.x `Skeleton3D` missing pose

nikitalita:
This PR fixes an issue that I missed in my previous PR (#87050), which introduced converting 3.x Skeletons.

For 3.x Skeletons, poses are relative to the rest position. If a pose is set to the rest position value, the pose will be set to default `Transform()` value. Since this is the default value for `pose`, this means that the pose property is not saved to the scene. 

Thus, when a deprecated Skeleton is loaded, the `pose` value is not set, and the skeleton position, rotation, and scale will be set to their default values and you'll get funky looking skeletons like this:

image

Fortunately, there is an easy fix. `bound_children` is a deprecated property on 3.x skeletons that we weren't doing anything with because it's not present in the current `Skeleton3D` format (bound children find their parent index automatically in 4.x). However, this property just so happens to be always set on all `Skeleton` nodes from 3.0 to 3.x branch current regardless of value, and it's always set AFTER both the pose and the rest properties. Thus, when that property is set on the Skeleton3D node, we just check to see if the position, rotation, and scale are their default values, and if so, set the pose to the rest position.

### DIFF
--- a/scene/3d/skeleton_3d.cpp
+++ b/scene/3d/skeleton_3d.cpp
@@ -95,17 +95,32 @@ bool Skeleton3D::_set(const StringName &p_path, const Variant &p_value) {
 	} else if (what == "scale") {
 		set_bone_pose_scale(which, p_value);
 #ifndef DISABLE_DEPRECATED
-	} else if (what == "pose") {
+	} else if (what == "pose" || what == "bound_children") {
 		// Kept for compatibility from 3.x to 4.x.
 		WARN_DEPRECATED_MSG("Skeleton uses old pose format, which is deprecated (and loads slower). Consider re-importing or re-saving the scene." +
 				(is_inside_tree() ? vformat(" Path: \"%s\"", get_path()) : String()));
-		// Old Skeleton poses were relative to rest, new ones are absolute, so we need to recompute the pose.
-		// Skeleton3D nodes were always written with rest before pose, so this *SHOULD* work...
-		Transform3D rest = get_bone_rest(which);
-		Transform3D pose = rest * (Transform3D)p_value;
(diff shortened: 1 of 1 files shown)


scene/3d/skeleton_3d.cpp
```diff
@@ -95,17 +95,32 @@ bool Skeleton3D::_set(const StringName &p_path, const Variant &p_value) {
 	} else if (what == "scale") {
 		set_bone_pose_scale(which, p_value);
 #ifndef DISABLE_DEPRECATED
-	} else if (what == "pose") {
+	} else if (what == "pose" || what == "bound_children") {
```
AThousandShips, line 98: I'd put this in two statements and duplicate the message instead to make it clearer and avoid duplicate check for the name

nikitalita: The reason it’s like this is because the check for “bound_children” is effectively a “pose” check (we don’t do anything with the property itself, we just use it to set the pose), and we don’t want to print the deprecated message twice for the same resource.

AThousandShips: Are they always set together? Then maybe just print it for one?

nikitalita: In some cases, but often the whole skeleton is set to the rest pose and would not have the “pose” property on any of the bones. In that case, “bound_children” is still an indicator that it has the old skeleton format with the old “pose” format, so the message would still apply.

lyuma (approved):
seems reasonable

fire (approved):

akien-mga:
Thanks!

Issue #85005, opened 2023-11-17
`GraphEdit`'s connection layer is not added as internal child


Yagich:
### Godot version

4.2b6

### System information

Linux/Manjaro

### Issue description

In my application, I add GraphNodes of a specific type. When getting all children with `include_internal` set to `false`, the `_connection_layer` is returned as the first child, which I don't think is intended; I expect the first non-internal child to be a user-added node.

https://github.com/godotengine/godot/blob/ad72de508363ca8d10c6b148be44a02cdf12be13/scene/gui/graph_edit.cpp#L2018

### Steps to reproduce

1. Add a GraphEdit
2. Add a GraphNode child to it at runtime
3. Try to access the node with `get_child(0)`

### Minimal reproduction project

[graphedit-non-internal-child.zip](https://github.com/godotengine/godot/files/13389335/graphedit-non-internal-child.zip)

YuriSizov:
> 4.0b6

You probably mean 4.2?

AThousandShips:
It was internal but seems to have been overlooked in https://github.com/godotengine/godot/pull/79308

CC @Geometror was this by design?

Yagich:
> > 4.0b6
> 
> You probably mean 4.2?

Yes, changed to 4.2b6, thanks!

Geometror:
I think I had to set it as a non-internal node when implementing `GraphFrame` (which was then postponed and removed from the refactoring PR) as frames need to be drawn below the connection layer (and we don't support internal nodes at arbitrary positions). Maybe this could be implemented utilizing the z index or another solution. Either way I agree the connection layer should be an internal node.

PR #67531, opened 2022-10-17
Enable Drag and Drop between SubViewports and Windows

Sauermann:
Make Drag and Drop an application-wide operation.
This allows to drop on Controls in other Viewports/Windows.

Also now all nodes in the SceneTree are notified about the Drag and Drop operation, with the exception of SubViewports that are not child-nodes of SubViewportContainers. This distinction seemed prudent, because these SubViewports need to be displayed manually and this requires manual setup also for Drag and Drop.

Unit-tests for Drag and Drop in different Windows & SubViewports are also added.

In order to achieve this, the following changes were made:
- `Viewport::_update_mouse_over` from #67791 is used to identify the drop-target (even within nested viewports)
- This Control is used as a basis for the Drop-operation, which replaces the previous algorithm, that was only aware of topmost Viewports.
- In order to handle DnD within non-default SubViewports (see for an example https://github.com/godotengine/godot-demo-projects/pull/925), it becomes necessary to track drag-operations in each (for the lack of a better name) section in the scene-tree.
- `is_directly_attached_to_screen` is replaced by `get_section_rootviewport`. which is of the same complexity, but additionally returns the topmost Viewport within the section
- `_propagate_viewport_notification` had to be adjusted to account for sections
- `DisplayServerX11::delete_sub_window` needed to be adjusted similar to #80142, since the event-callback is also needed on linuxbsd during Window-deletion

Relations:
- alternative to #71509 and has a broader scope
- fix #28522
- fix #44390
- fix #59622
- fix #77665
- fix no longer (will be done separately) #67186 
- fix duplicate #62384
- fix duplicate #68063
- was previously blocked by #71972
- utilizes #67791
- follow-up to #80142
- related to #64949
- related to #20881

MRP: [BugDragNDrop.zip](https://github.com/godotengine/godot/files/10468628/BugDragNDrop.zip) (updated 2023-01-20)

The Drag-Preview in the following Video is outdated, because that will be done in a different PR.

https://user-images.githubusercontent.com/6299227/196206434-2b8cba24-5993-40d2-a8dc-22749b7ccd27.mp4

Updated 2023-01-28: Fix merge conflict
Updated 2023-03-20: Fix merge conflict
Updated 2023-05-30: Linked another bug-report, that this PR fixes. Resolve problematic testcases. Fix merge conflict.
Updated 2023-06-07: Fix merge conflict
Updated 2023-06-08: Fix issue with docks
Updated 2023-06-19: Fix issue with simultaneous drag starts
[...]

Issue #59622, opened 2022-03-28
Error emitting `!get_viewport()->gui_is_dragging()` when using `set_drag_preview` from a sub-window


Chaosus:
### Godot version

Godot 4.0 / 7e5cd4bb48bda1b2f2a80065be0b3820a5346c1d

### System information

Windows 10

### Issue description

When user try to drag something from subwindow - the error emitting at the `set_drag_preview` function. It's observed in a visual shader editor:

![bug](https://user-images.githubusercontent.com/3036176/160355332-c5f4dd6d-6e3d-42ac-b55c-37facfebb06d.gif)

I think that's because `gui.dragging` boolean flag does not correctly transmit from sub-window to parent. 

### Steps to reproduce

^^

### Minimal reproduction project

Too easy to reproduce in the visual shader editor.

PR #88014, opened 2024-02-06
Implement GraphFrame and integrate it in VisualShader

### LINKED ISSUE #5271: Refactor of GraphEdit/GraphNode

Godot Engine

Compared to other software the current graph editor is rather limited in usability and flexibility (e.g. no subgraphing, comment nodes don't work well)
In addition to that the underlying code is quite convoluted and hard to maintain, making it difficult to extend the visual editors to cope with the demands/expectations of such a system.

The visual editor system including the GraphNode and GraphEdit nodes as well as the editor specific classes should be refactored and cleaned up to build a foundation for easy extensions and usability improvements.

This refactor may include:
- Split GraphNode into `GraphNodeBase`, `GraphNode` and `GraphNodeComment`(or `GraphFrame`) (open for naming suggestions) due to:
	-   Maintainability: The GraphNode class got really large and complex, so splitting it up improves code navigation and readability (removes some `is_comment()`-branching/ cleans up theme constants)
	-   Flexibility: Allows adding more features to the frame/comment graph node or extend GraphNodeBase for a completely custom node without most of the predefined connection functionality (reference image backdrops, etc.)
	-   Performance/implementation complexity: For example, a comment node does not need to keep track of any connections/slots/ports
- Improve the comment node system
	- Draw connection lines above comment nodes, but below normal graph nodes (would be easier with proper layers, but for now there is godotengine/godot#61783)
	- Nested comment nodes are automatically sorted
	- Consider automatically resizing comment nodes (similar to Blender)
- Allow for customizable title bars to include custom controls (e.g. buttons/icons)
    - This can be useful if we want to add a button to minimize a node (hiding all its properties/slots, just show the ports) or to enter a subgraph node
- Extract the menu from GraphEdit and make it its own control (if we don't want to expose this we can just do that internally as a improvement to code structure[GraphEdit is huge with around 2500 lines]) or at least add a toggle to turn it off so users can add their own
- [might require some discussion] Introduce GraphNode layers (e.g. as a new node) with layer specific properties as a structural improvement. This would flesh out the concept of layers - which is already (kind of) present internally - and improve it in terms of WYSIWYG. For example, here you have a backdrop layer for reference images, above that a comment node layer which keeps the nodes sorted based on their bounding boxes, then the layer containing normal/functional graph nodes (+ the connection lines) and on top of that you could add a layer which shows a subgraph dynamically. There are still many things to figure out and there may be some alternatives to this approach, but this is just what I currently have in mind.
- Modify GraphEdit to properly support subgraphs (maybe the layer introduction above plus a new GraphNode type is enough together with some additional tweaks or there is a need of a more fundamental modification)

Here's an example how some of the concepts are implemented in Blender:
grafik

I already opened a demo PR, which partially implements this proposal for interested people to try out: godotengine/godot#67152
This proposal should create a common ground for discussion to decide which of the above points could improve the current situation with GraphEdit/GraphNode, so we can provide a more flexible graph editor system to build upon (for shaders/animation trees/audio etc.).

-

The visual graph editors are a core part of the engine.

Geometror:
This PR introduces `GraphFrame`, a new `GraphElement` designed for organizing node graphs, and integrates it into the VisualShader editor (can be split if desired).
It aims to reimplement the functionality that was lost when comment nodes were removed. While it works a bit differently (very similar to Blender) it should be a lot more predictable and much less problematic compared to the old implementation of comment nodes.

Finally closes godotengine/godot-proposals#5271.

grafik

https://github.com/godotengine/godot/assets/50084500/d81c7cf9-6b3d-4d44-b9ae-07f7af88dc43

## Detailed changes:
- Add `GraphFrame`
	- A special graph element to which you can attach other graph elements (nodes and frames). Its rect automatically adapts to the attached elements (enclosing), if autoshrink is enabled. Only attached are moved together with the frame.
	- Can be tinted in an arbitrary color
	- GraphEdit automatically sorts the frames topologically (nested frames are supported)
	- GraphEdit automatically puts frames behind the connection layer
		- This requires to make the connection layer a non-internal child again (reintroduces #85005) since there is no better solution at the moment to keep it between background nodes, e.g. frames, and connectable nodes (just modifying the z_index of each node type would mess up the input and is just confusing)
- VisualShader integration
	- Remove `VisualShaderNodeComment` and replace it with `VisualShaderNodeFrame`
	- Right click context menu: Detach selected nodes from their "parent" frame
	- To attach nodes to a frame simply drag and drop them onto it.
	- Context menu option to enable/disable autoshrink and tint color (as well as choose the tint color)

grafik

## API
*Note for users of GraphEdit: Integrating GraphFrame in your application is a bit more complicated that the old comment nodes (where you could just check a property and it "worked" out of the box), especially for Undo-Redo, but in return it should work more reliable and be a lot more useful.*

- GraphEdit
	- Methods
		- `attach_graph_element_to_frame(element, frame)` : Attach a node to a frame (params are string names)
		- `detach_graph_element_from_frame(element)` : Detach a node from its frame (param is string name)
		- `get_frame(element)` : Get the frame a node is attached to
	- Signals
		- `frame_size_changed (frame, new_size)` : Emitted when the size of a frame changes
		- `nodes_linked_to_frame_request(elements, frame)` : Emitted when nodes are dropped onto a frame
- GraphElement
	- Signals
		- `resize_end` : Emitted after releasing the mouse button after resizing the node
- GraphFrame
	- Properties
		- `title` : The title of the frame
		- `autoshrink_enabled` : Whether the frame should automatically shrink to fit its contents (attached nodes/frames)
		- `autoshrink_margin` : The margin to add to the frame size on each side when autoshrink is enabled
		- `drag_margin` : The margin around the frame where it can be dragged
		- `tint_color_enabled` : Whether the frame should be tinted in a color
		- `tint_color` : The color to tint the frame in
	- Methods
		- `get_titlebar_hbox` : Get the HBoxContainer of the titlebar (useful for customizing the titlebar)
	- Signals
		- `autoshrink_toggled` : Emitted when the autoshrink property is toggled
- VisualShader
	- Methods
		- `attach_node_to_frame(node, frame)` : Attach a node to a frame
		- `detach_node_from_frame(node)` : Detach a node from its frame

## TODO:
- [x] Theme cache
- [x] Docs
- [x] Address remaining TODO comments

_Production edit: closes godotengine/godot-roadmap#13_


doc/classes/GraphFrame.xml
```diff
@@ -0,0 +1,69 @@
...
+			The color of the frame when [member tint_color_enabled] is [code]true[/code].
+		</member>
+		<member name="tint_color_enabled" type="bool" setter="set_tint_color_enabled" getter="is_tint_color_enabled" default="false">
+			If [code]true[/code], the tint color will be used to tint the frame.
+		</member>
+		<member name="title" type="String" setter="set_title" getter="get_title" default="&quot;&quot;">
+			Title of the frame.
+		</member>
+		<member name="title_centered" type="bool" setter="set_title_centered" getter="is_title_centered" default="true">
+			Determines whether the title is aligned to the left/right or centered.
```
Mickeon, line 42: Grrrrr- Bark, bark, bark.
```suggestion
			If [code]true[/code] the title is centered. Otherwise, the title is aligned to the left or right.
```

doc/classes/VisualShaderNodeFrame.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="VisualShaderNodeFrame" inherits="VisualShaderNodeResizableBase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
+	<brief_description>
+		A frame other visual shader nodes can be attached to for better organization.
+	</brief_description>
+	<description>
+		A rectangular frame that can be used to group visual shader nodes together to improve organization. 
+		Nodes attached to the frame will move with it when it is dragged and it can automatically resize to enclose all attached nodes.
+		It's title, description and color can be customized.
```
Mickeon, line 9:
```suggestion
		Its title, description and color can be customized.
```

doc/classes/GraphElement.xml
```diff
@@ -17,7 +17,7 @@
 		</member>
 		<member name="resizable" type="bool" setter="set_resizable" getter="is_resizable" default="false">
 			If [code]true[/code], the user can resize the GraphElement.
-			[b]Note:[/b] Dragging the handle will only emit the [signal resize_request] signal, the GraphElement needs to be resized manually.
+			[b]Note:[/b] Dragging the handle will only emit the [signal resize_request] and [signal resize_end] signal, the GraphElement needs to be resized manually.
```
AThousandShips, line 20:
```suggestion
			[b]Note:[/b] Dragging the handle will only emit the [signal resize_request] and [signal resize_end] signals, the GraphElement needs to be resized manually.
```

scene/gui/graph_edit.cpp
```diff
@@ -1926,6 +2250,62 @@ bool GraphEdit::is_valid_connection_type(int p_type, int p_with_type) const {
 	return valid_connection_types.has(ct);
 }
 
+void GraphEdit::attach_graph_element_to_frame(const StringName &p_graph_element, const StringName &p_parent_frame) {
+	GraphFrame *frame = Object::cast_to<GraphFrame>(get_node(NodePath(p_parent_frame)));
+	if (!frame) {
+		print_line("Parent frame does not exist or is not a GraphFrame.");
+	}
+	ERR_FAIL_COND_MSG(!frame, "Parent frame does not exist or is not a GraphFrame.");
```
AThousandShips, line 2258:
```suggestion
	ERR_FAIL_NULL_MSG(frame, "Parent frame does not exist or is not a GraphFrame.");
```

scene/gui/graph_edit.cpp
```diff
@@ -1926,6 +2250,62 @@ bool GraphEdit::is_valid_connection_type(int p_type, int p_with_type) const {
...
+		return;
+	}
+
+	GraphFrame *frame = Object::cast_to<GraphFrame>(get_node(NodePath(linked_parent_map[p_graph_element])));
+
+	if (!frame) {
+		print_line("Parent frame does not exist or is not a GraphFrame.");
+	}
+
+	ERR_FAIL_COND_MSG(!frame, "Parent frame does not exist or is not a GraphFrame.");
```
AThousandShips, line 2278:
```suggestion
	ERR_FAIL_NULL_MSG(frame, "Parent frame does not exist or is not a GraphFrame.");
```

doc/classes/GraphFrame.xml
```diff
@@ -0,0 +1,69 @@
...
+		</member>
+	</members>
+	<signals>
+		<signal name="autoshrink_toggled">
+			<description>
+				Emitted when the frame's autoshrink state changes.
```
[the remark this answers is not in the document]
Geometror: It should (probably) only be emitted if the autoshrink_enabled property changes.

scene/resources/visual_shader.cpp
```diff
@@ -1129,6 +1141,36 @@ bool VisualShader::is_port_types_compatible(int p_a, int p_b) const {
 	return MAX(0, p_a - (int)VisualShaderNode::PORT_TYPE_BOOLEAN) == (MAX(0, p_b - (int)VisualShaderNode::PORT_TYPE_BOOLEAN));
 }
 
+void VisualShader::attach_node_to_frame(Type p_type, int p_node, int p_frame) {
+	ERR_FAIL_INDEX(p_type, TYPE_MAX);
+	Graph *g = &graph[p_type];
+
+	ERR_FAIL_COND(!g->nodes.has(p_node));
+	ERR_FAIL_COND(p_frame < 0);
```
AThousandShips, line 1149:
```suggestion
	ERR_FAIL_COND(p_frame < 0);
	Graph *g = &graph[p_type];

	ERR_FAIL_COND(!g->nodes.has(p_node));
```
I'd say to not delay after lookup for this check

doc/classes/GraphEdit.xml
```diff
@@ -395,6 +417,20 @@
...
+			<param index="1" name="new_size" type="Vector2" />
+			<description>
+				Emitted when the size of the given [GraphFrame] changes.
+			</description>
+		</signal>
+		<signal name="graph_elements_linked_to_frame_request">
+			<param index="0" name="elements" type="Array" />
+			<param index="1" name="frame" type="StringName" />
+			<description>
+				Emitted when [GraphElement]s are dropped onto a [GraphFrame], when they were not previously attached to any other one.
```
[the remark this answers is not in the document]
Geometror: Went with 
```
Emitted when one or more [GraphElement]s are dropped onto the [GraphFrame] named [param frame], when they were not previously attached to any other one.
[param elements] is an array of [GraphElement]s to be attached
```

scene/gui/graph_edit.cpp
```diff
@@ -29,6 +29,7 @@
 /**************************************************************************/
 
 #include "graph_edit.h"
+
```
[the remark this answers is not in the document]
Geometror: This is intetional. When I remove this line, clang-format sorts it and it ends up before `#include "graph_edit.h"` which then results in a compilation error.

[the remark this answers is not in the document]
AThousandShips: That's very odd! It works everywhere else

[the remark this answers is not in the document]
Geometror: I just checked it with the other files, same for me there (using clang-format 17.0.6)
But I'm actually confused why it worked everywhere else in the first place since
```
#include "graph_edit.compat.inc"
#include "graph_edit.h"
```
is the lexicographically correct order.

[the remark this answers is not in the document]
AThousandShips: Unsure, some quirk possibly, but same with 17.0.6 and it works fine with this file specifically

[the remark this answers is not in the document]
Geometror: Mmh, I still have no idea why this doesn't work for me, but I changed it manually for now.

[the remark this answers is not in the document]
AThousandShips: Tested with 16.0.4 as well and it works as above, so probably some strange bug there, worth a look later

Geometror:
@Yagich 
1. Yes, it should. You can test it with the VS expression node which is resizable.
2. I assume you mean the connectable nodes (`GraphNode`s) with "slot children". Then yes, you would need to check whether the children are of type `GraphElement` or `GraphNode` (depending on whether you want the frame nodes too). Just stripping the first child from the children array would only work if you don't have any GraphFrames in there. What we could do however, is expose a method `get_connection_layer_index` so that you could just use that instead of type checking (but then again type checking is a bit safer and not that expensive).

Yagich:
I've tested the PR and overall it works great! I've only been able to find one issue: if you try to attach a frame to itself, the application will hang. It's not unfixable in user code, but I wonder if it should just be a no-op if `element` and `frame` are the same when calling `attach_graph_element_to_frame()`.

Yagich:
> I assume you mean the connectable nodes (`GraphNode`s) with "slot children"

Sorry, I think I got confused what issue this refers to when writing the original comment. What I meant was that since the connection layer is now not internal, every `get_children()` call on `GraphEdit` to get elements now has to slice off the first child. I suppose there really is no way around it now.

scene/gui/graph_edit.cpp
```diff
@@ -1926,6 +2249,55 @@ bool GraphEdit::is_valid_connection_type(int p_type, int p_with_type) const {
...
+	if (!linked_parent_map.has(p_attached_graph_element)) {
+		return nullptr;
+	}
+
+	Node *parent = get_node_or_null(NodePath(linked_parent_map[p_attached_graph_element]));
+
+	return Object::cast_to<GraphFrame>(parent);
+}
+
+TypedArray<StringName> GraphEdit::get_attached_noded_of_frame(const StringName &p_graph_frame) {
```
Yagich, line 2288: typo: `get_attached_noded_of_frame` -> `get_attached_nodes_of_frame`

Geometror: I also haven't exposed this method by accident. Fixed.

Geometror:
@Yagich Thanks for testing. I made both `attach_graph_element_to_frame()` and `detach_graph_element_from_frame()` more robust.

doc/classes/GraphFrame.xml
```diff
@@ -0,0 +1,69 @@
...
+		</member>
+		<member name="tint_color_enabled" type="bool" setter="set_tint_color_enabled" getter="is_tint_color_enabled" default="false">
+			If [code]true[/code], the tint color will be used to tint the frame.
+		</member>
+		<member name="title" type="String" setter="set_title" getter="get_title" default="&quot;&quot;">
+			Title of the frame.
+		</member>
+		<member name="title_centered" type="bool" setter="set_title_centered" getter="is_title_centered" default="true">
+			If [code]true[/code] the title is centered. Otherwise, the title is aligned to the left or right.
+		</member>
```
Mickeon, line 43: In https://github.com/godotengine/godot/pull/88014#discussion_r1481433307 I did agree with you and said that this property should probably not exist. If users need it they will ask for it.

Geometror: Removed it.

Geometror:
Rebased and adjusted the style a bit after #85017 has been merged.
image
image

akien-mga:
Tested briefly, I'm not familiar with Visual Shader so I'm just trying some random stuff.

I created a Visual Shader, attached to a Sprite2D (icon.svg).

Pressed "Add Node...", dragged and dropped ColorConstant to the graph, I get those two errors:
```
ERROR: Condition "!get_viewport()->gui_is_dragging()" is true.
   at: set_drag_preview (./scene/gui/control.cpp:1965)
ERROR: Graph element is not attached to any frame.
   at: detach_graph_element_from_frame (./scene/gui/graph_edit.cpp:2269)
```

The frame error gets repeated every time I do an action, e.g. toggling visibility of the ColorConstant, or connecting it to Output > Color.

I then found how to add a frame via right click/Add Node, which auto-shrunk to tiny. I had to right click it to disable auto-shrink, so I could extend it to encompass three nodes I had.

Then it seemed only the Output node within the frame was actually attached to the frame. It's not clear to me how to attach nodes to a frame. I did a bunch of random single or double clicking on nodes, moving them a bit within the frame, and they somehow ended up being attached, though there's visible feedback about it.

The end result works, but as a first time user this wasn't very intuitive.
image

The frame description also can easily overlap with nodes:
image

This was a really brief test on laptop with touchpad, so hard mode :P
Hope it's useful anyway :)

BlueCube3310:
Bug report: Duplicating a frame with elements inside of it will cause the elements inside to be attached to both the original and the duplicate.

Geometror:
@akien-mga @BlueCube3310 Thanks for testing! 
All the described (technical) issues should be solved now.

Regarding the usability problems:
I will experiment a bit with a cursor hint (when dragging a unattached node over a frame), a hint text for empty frames and maybe a small visual indicator that a node is attached (although that might be too much and if not likely part of a subsequent PR due to limited time).

I'm thinking about removing the description feature and instead add a true "comment" node which is basically a dragable label with a full transparent background (which would behave the same way as normal node inside a frame, so you could place it freely). IMO this would be much more intuitive.

> BTW, this is likely not for this PR, but do you plan to add support for duplicating frames with their contents, and/or saving frames with their contents as a reusable component? I'm not familiar with Visual Shader but this sounds like it could be quite useful.

Yes, I plan to do something similar (https://github.com/godotengine/godot-proposals/issues/7730) 
Although that is completely separate from graph frames, which are purely for visual organization.

Chaosus:
> Pressed "Add Node...", dragged and dropped ColorConstant to the graph, I get those two errors:
> 
> ERROR: Condition "!get_viewport()->gui_is_dragging()" is true.
>    at: set_drag_preview (./scene/gui/control.cpp:1965)

@akien-mga The first message is related to old bug: https://github.com/godotengine/godot/issues/59622, should be fixed by https://github.com/godotengine/godot/pull/67531

Geometror:
Small update: Refactoring on the visual shader side and a hint for empty frames:
frame drop

KoBeWi:
You can't undo adding a frame and trying to delete one results in a crash:
```
Engine version: Godot Engine v4.3.dev.custom_build (5d08c2631cadf29d80ca23c7d537e03c3e5edcc5)
[0] VisualShaderNode::is_deletable (C:\godot_source\scene\resources\visual_shader.cpp:352)
[1] VisualShaderNode::is_deletable (C:\godot_source\scene\resources\visual_shader.cpp:352)
[2] VisualShaderEditor::_delete_nodes_request (C:\godot_source\editor\plugins\visual_shader_editor_plugin.cpp:4346)
[3] call_with_variant_args_helper<VisualShaderEditor,TypedArray<StringName> const &,0> (C:\godot_source\core\variant\binder_common.h:304)
[4] call_with_variant_args<VisualShaderEditor,TypedArray<StringName> const &> (C:\godot_source\core\variant\binder_common.h:419)
[5] CallableCustomMethodPointer<VisualShaderEditor,TypedArray<StringName> const &>::call (C:\godot_source\core\object\callable_method_pointer.h:104)
[6] Callable::callp (C:\godot_source\core\variant\callable.cpp:58)
[7] Object::emit_signalp (C:\godot_source\core\object\object.cpp:1202)
[8] Node::emit_signalp (C:\godot_source\scene\main\node.cpp:3904)
[9] Object::emit_signal<TypedArray<StringName> > (C:\godot_source\core\object\object.h:933)
[10] GraphEdit::gui_input (C:\godot_source\scene\gui\graph_edit.cpp:1972)
[11] Control::_call_gui_input (C:\godot_source\scene\gui\control.cpp:1800)
[12] Viewport::_gui_input_event (C:\godot_source\scene\main\viewport.cpp:2178)
[13] Viewport::push_input (C:\godot_source\scene\main\viewport.cpp:3229)
[14] Window::_window_input (C:\godot_source\scene\main\window.cpp:1627)
[15] call_with_variant_args_helper<Window,Ref<InputEvent> const &,0> (C:\godot_source\core\variant\binder_common.h:304)
[16] call_with_variant_args<Window,Ref<InputEvent> const &> (C:\godot_source\core\variant\binder_common.h:419)
[17] CallableCustomMethodPointer<Window,Ref<InputEvent> const &>::call (C:\godot_source\core\object\callable_method_pointer.h:104)
[18] Callable::callp (C:\godot_source\core\variant\callable.cpp:58)
[19] Callable::call<Ref<InputEvent> > (C:\godot_source\core\variant\variant.h:864)
[20] DisplayServerWindows::_dispatch_input_event (C:\godot_source\platform\windows\display_server_windows.cpp:3450)
[21] DisplayServerWindows::_dispatch_input_events (C:\godot_source\platform\windows\display_server_windows.cpp:3420)
[22] Input::_parse_input_event_impl (C:\godot_source\core\input\input.cpp:773)
[23] Input::flush_buffered_events (C:\godot_source\core\input\input.cpp:1044)
[24] DisplayServerWindows::process_events (C:\godot_source\platform\windows\display_server_windows.cpp:2969)
[25] OS_Windows::run (C:\godot_source\platform\windows\os_windows.cpp:1476)
[26] widechar_main (C:\godot_source\platform\windows\godot_windows.cpp:181)
[27] _main (C:\godot_source\platform\windows\godot_windows.cpp:206)
[28] main (C:\godot_source\platform\windows\godot_windows.cpp:220)
[29] WinMain (C:\godot_source\platform\windows\godot_windows.cpp:234)
[30] __scrt_common_main_seh (D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288)
[31] <couldn't map PC to fn name>
-- END OF BACKTRACE --
```

doc/classes/GraphFrame.xml
```diff
@@ -0,0 +1,66 @@
...
+	</methods>
+	<members>
+		<member name="autoshrink_enabled" type="bool" setter="set_autoshrink_enabled" getter="is_autoshrink_enabled" default="true">
+			If [code]true[/code], the frame's rect will be adjusted automatically to enclose all attached [GraphElement]s.
+		</member>
+		<member name="autoshrink_margin" type="Vector2i" setter="set_autoshrink_margin" getter="get_autoshrink_margin" default="Vector2i(40, 40)">
+			The margin around the frame (horizontal, vertical) that is used to calculate the size of the frame when [member autoshrink_enabled] is [code]true[/code].
+		</member>
+		<member name="drag_margin" type="int" setter="set_drag_margin" getter="get_drag_margin" default="16">
+			The margin around the frame that can be used to drag the frame.
```
KoBeWi, line 29: This seems to be a margin *inside* the frame. It should be clarified.

KoBeWi:
The crashes are now fixed, but I still have the problem with detaching. I'm on Windows if that matters.

> What was your setup? Sounds like you added GraphNodes as children to a GraphFrame, which is not how it works (due to layering issues it's not possible this way -> connection lines). Nodes have to be attached to a frame using a GraphEdit method.

Yep, I just used it wrongly.

> I'm not sure what to do with the resizable property. It makes not much sense for frames, unless we would remove the autoshrink prop and use it as a replacement.

You can hide it at least, using `_validate_property()`.

doc/classes/GraphEdit.xml
```diff
@@ -179,6 +201,13 @@
 				Returns an [Array] containing the list of connections that intersect with the given [Rect2]. A connection consists in a structure of the form [code]{ from_port: 0, from_node: "GraphNode name 0", to_port: 1, to_node: "GraphNode name 1" }[/code].
 			</description>
 		</method>
+		<method name="get_frame">
```
KoBeWi, line 204:
```suggestion
		<method name="get_element_frame">
```
That would be more accurate to what it does.

Geometror: Oh, yes, I wanted to rename this but forgot :)

KoBeWi:
Changing `autoshrink_margin` with a node attached to a frame will not automatically redraw the frame.

https://github.com/godotengine/godot/assets/2223172/285e1e71-0fca-48e0-95c9-eb0a9a3d65ee

scene/gui/graph_frame.h
```diff
@@ -0,0 +1,108 @@
...
+
+class HBoxContainer;
+
+class GraphFrame : public GraphElement {
+	GDCLASS(GraphFrame, GraphElement);
+
+	struct _MinSizeCache {
+		int min_size;
+		bool will_stretch;
+		int final_size;
```
KoBeWi, line 44:
```suggestion
		int min_size = 0;
		bool will_stretch = 0;
		int final_size = 0;
```

scene/gui/graph_frame.h
```diff
@@ -0,0 +1,108 @@
...
+#define GRAPH_FRAME_H
+
+#include "scene/gui/graph_element.h"
+
+class HBoxContainer;
+
+class GraphFrame : public GraphElement {
+	GDCLASS(GraphFrame, GraphElement);
+
+	struct _MinSizeCache {
```
KoBeWi, line 41: Why is the name prefixed?

Geometror: Was copied from box_container.cpp, I thought this was a convention for private, single-use structs.

scene/gui/graph_frame.cpp
```diff
@@ -0,0 +1,351 @@
...
+				emit_signal(SNAME("resize_end"), get_size());
+				return;
+			}
+		}
+	}
+
+	Ref<InputEventMouseMotion> mm = p_ev;
+
+	// Only resize if the frame is not auto-resizing based on linked nodes.
+	if (resizing && mm.is_valid() && !autoshrink_enabled) {
```
KoBeWi, line 75:
```suggestion
	if (resizing && !autoshrink_enabled && mm.is_valid()) {
```
Though autoshrink check shouldn't be here I think, because `resizable` depends on it anyway. In unlikely situation where autoshrink changes during resizing, it will never be released.

scene/gui/graph_frame.h
KoBeWi: BoxContainer code is super old and was probably never updated.
Though I don't recall any other case of such struct.

scene/gui/graph_edit.cpp
```diff
@@ -2374,7 +2761,7 @@ GraphEdit::GraphEdit() {
 	top_layer->connect("focus_exited", callable_mp(panner.ptr(), &ViewPanner::release_pan_key));
 
 	connections_layer = memnew(Control);
-	add_child(connections_layer, false, INTERNAL_MODE_FRONT);
+	add_child(connections_layer, false);
```
KoBeWi, line 2764: Why did you remove internal mode? It makes the connection layer easily accessible by the user.

scene/gui/graph_edit.cpp
```diff
@@ -442,11 +442,42 @@ void GraphEdit::_update_scroll() {
...
+	if (!frame) {
+		graph_node->move_to_front();
+		return;
+	}
+
+	// Reorder the frames behind the connection layer.
+	List<GraphFrame *> attached_nodes_to_move;
+	attached_nodes_to_move.push_back(frame);
+
+	while (attached_nodes_to_move.size()) {
```
KoBeWi, line 460:
```suggestion
	while (!attached_nodes_to_move.is_empty()) {
```

scene/gui/graph_edit.cpp
```diff
@@ -442,11 +442,42 @@ void GraphEdit::_update_scroll() {
...
+		attached_nodes_to_move.pop_front();
+
+		// Move the frame to the front of the background node index range.
+		attached_frame->get_parent()->call_deferred("move_child", attached_frame, background_nodes_separator_idx - 1);
+
+		if (!frame_attached_nodes.has(attached_frame->get_name())) {
+			continue;
+		}
+
+		for (StringName attached_node_name : frame_attached_nodes.get(attached_frame->get_name())) {
```
KoBeWi, line 471:
```suggestion
		for (const StringName &attached_node_name : frame_attached_nodes.get(attached_frame->get_name())) {
```

scene/gui/graph_edit.h
```diff
@@ -304,12 +313,15 @@ class GraphEdit : public Control {
 
 	void _graph_element_selected(Node *p_node);
 	void _graph_element_deselected(Node *p_node);
-	void _graph_element_moved_to_front(Node *p_node);
-	void _graph_element_resized(Vector2 p_new_minsize, Node *p_node);
+	void _graph_element_resize_request(Vector2 p_new_minsize, Node *p_node);
+	void _graph_frame_shrink_toggled(Vector2 p_new_minsize, GraphFrame *p_node);
```
KoBeWi, line 317:
```suggestion
	void _graph_element_resize_request(const Vector2 &p_new_minsize, Node *p_node);
	void _graph_frame_shrink_toggled(const Vector2 &p_new_minsize, GraphFrame *p_node);
```

scene/gui/graph_edit.cpp
```diff
@@ -520,10 +602,24 @@ void GraphEdit::add_child_notify(Node *p_child) {
...
 
-		graph_element->connect("raise_request", callable_mp(this, &GraphEdit::_graph_element_moved_to_front).bind(graph_element));
-		graph_element->connect("resize_request", callable_mp(this, &GraphEdit::_graph_element_resized).bind(graph_element));
+		GraphFrame *graph_frame = Object::cast_to<GraphFrame>(graph_element);
+		if (graph_frame) {
+			background_nodes_separator_idx++;
+
+			call_deferred("move_child", graph_frame, 0);
+			call_deferred("_reorder_frame_nodes", graph_frame->get_name());
+			call_deferred("move_child", connections_layer, background_nodes_separator_idx);
```
KoBeWi, line 614:
```suggestion
			callable_mp((Node *)this, &Node::move_child).call_deferred(graph_frame, 0);
			callable_mp((Node *)this, &Node::move_child).call_deferred(connections_layer, background_nodes_separator_idx);
```
`_reorder_frame_nodes` does not exist.

scene/gui/graph_edit.cpp
```diff
@@ -565,8 +661,36 @@ void GraphEdit::remove_child_notify(Node *p_child) {
...
+		}
+
+		if (linked_parent_map.has(graph_element->get_name())) {
+			GraphFrame *parent_frame = Object::cast_to<GraphFrame>(get_node(NodePath(linked_parent_map[graph_element->get_name()])));
+			if (parent_frame) {
+				if (frame_attached_nodes.has(parent_frame->get_name())) {
+					frame_attached_nodes.get(parent_frame->get_name()).erase(graph_element->get_name());
+				}
+				linked_parent_map.erase(graph_element->get_name());
+				call_deferred("_reorder_frame_nodes", parent_frame->get_name());
```
KoBeWi, line 677:
```suggestion
```

scene/resources/visual_shader.cpp
```diff
@@ -4160,66 +4208,120 @@ VisualShaderNodeResizableBase::VisualShaderNodeResizableBase() {
...
+	ClassDB::bind_method(D_METHOD("set_title", "title"), &VisualShaderNodeFrame::set_title);
+	ClassDB::bind_method(D_METHOD("get_title"), &VisualShaderNodeFrame::get_title);
+
+	ClassDB::bind_method(D_METHOD("set_tint_color_enabled", "p_enable"), &VisualShaderNodeFrame::set_tint_color_enabled);
+	ClassDB::bind_method(D_METHOD("is_tint_color_enabled"), &VisualShaderNodeFrame::is_tint_color_enabled);
+
+	ClassDB::bind_method(D_METHOD("set_tint_color", "p_color"), &VisualShaderNodeFrame::set_tint_color);
+	ClassDB::bind_method(D_METHOD("get_tint_color"), &VisualShaderNodeFrame::get_tint_color);
+
+	ClassDB::bind_method(D_METHOD("set_autoshrink_enabled", "p_enable"), &VisualShaderNodeFrame::set_autoshrink_enabled);
```
AThousandShips, line 4308:
```suggestion
	ClassDB::bind_method(D_METHOD("set_autoshrink_enabled", "enable"), &VisualShaderNodeFrame::set_autoshrink_enabled);
```

scene/gui/graph_edit.cpp
```diff
@@ -442,11 +442,42 @@ void GraphEdit::_update_scroll() {
...
+	// Reorder the frames behind the connection layer.
+	List<GraphFrame *> attached_nodes_to_move;
+	attached_nodes_to_move.push_back(frame);
+
+	while (attached_nodes_to_move.size()) {
+		GraphFrame *attached_frame = attached_nodes_to_move.front()->get();
+		attached_nodes_to_move.pop_front();
+
+		// Move the frame to the front of the background node index range.
+		attached_frame->get_parent()->call_deferred("move_child", attached_frame, background_nodes_separator_idx - 1);
```
KoBeWi, line 465:
```suggestion
		callable_mp((Node *)attached_frame, &Node::move_child).call_deferred(attached_frame, background_nodes_separator_idx - 1);
```

scene/resources/visual_shader.cpp
```diff
@@ -4160,66 +4208,120 @@ VisualShaderNodeResizableBase::VisualShaderNodeResizableBase() {
...
+	ClassDB::bind_method(D_METHOD("is_autoshrink_enabled"), &VisualShaderNodeFrame::is_autoshrink_enabled);
 
-	ClassDB::bind_method(D_METHOD("set_description", "description"), &VisualShaderNodeComment::set_description);
-	ClassDB::bind_method(D_METHOD("get_description"), &VisualShaderNodeComment::get_description);
+	ClassDB::bind_method(D_METHOD("add_attached_node", "p_node"), &VisualShaderNodeFrame::add_attached_node);
```
AThousandShips, line 4311:
```suggestion
	ClassDB::bind_method(D_METHOD("add_attached_node", "node"), &VisualShaderNodeFrame::add_attached_node);
```

scene/resources/visual_shader.cpp
```diff
@@ -4160,66 +4208,120 @@ VisualShaderNodeResizableBase::VisualShaderNodeResizableBase() {
...
+	ClassDB::bind_method(D_METHOD("remove_attached_node", "p_node"), &VisualShaderNodeFrame::remove_attached_node);
```
AThousandShips, line 4312:
```suggestion
	ClassDB::bind_method(D_METHOD("remove_attached_node", "node"), &VisualShaderNodeFrame::remove_attached_node);
```

scene/resources/visual_shader.cpp
```diff
@@ -4160,66 +4208,120 @@ VisualShaderNodeResizableBase::VisualShaderNodeResizableBase() {
...
+}
+
+void VisualShaderNodeFrame::_bind_methods() {
+	ClassDB::bind_method(D_METHOD("set_title", "title"), &VisualShaderNodeFrame::set_title);
+	ClassDB::bind_method(D_METHOD("get_title"), &VisualShaderNodeFrame::get_title);
+
+	ClassDB::bind_method(D_METHOD("set_tint_color_enabled", "p_enable"), &VisualShaderNodeFrame::set_tint_color_enabled);
+	ClassDB::bind_method(D_METHOD("is_tint_color_enabled"), &VisualShaderNodeFrame::is_tint_color_enabled);
+
+	ClassDB::bind_method(D_METHOD("set_tint_color", "p_color"), &VisualShaderNodeFrame::set_tint_color);
```
AThousandShips, line 4305:
```suggestion
	ClassDB::bind_method(D_METHOD("set_tint_color", "color"), &VisualShaderNodeFrame::set_tint_color);
```

scene/gui/graph_edit.cpp
```diff
@@ -565,8 +661,36 @@ void GraphEdit::remove_child_notify(Node *p_child) {
...
+					frame_attached_nodes.get(parent_frame->get_name()).erase(graph_element->get_name());
+				}
+				linked_parent_map.erase(graph_element->get_name());
+				call_deferred("_reorder_frame_nodes", parent_frame->get_name());
+				_update_graph_frame(parent_frame);
+			}
+		}
+
+		if (frame_attached_nodes.has(graph_element->get_name())) {
+			for (String attached_node_name : frame_attached_nodes.get(graph_element->get_name())) {
```
KoBeWi, line 683:
```suggestion
			for (const StringName &attached_node_name : frame_attached_nodes.get(graph_element->get_name())) {
```

scene/gui/graph_frame.h
```diff
@@ -0,0 +1,108 @@
...
+	virtual void _resort() override;
+
+public:
+	void set_title(const String &p_title);
+	String get_title() const;
+
+	void set_autoshrink_enabled(bool p_enable);
+	bool is_autoshrink_enabled() const;
+
+	void set_autoshrink_margin(const int &p_margin);
```
AThousandShips, line 88:
```suggestion
	void set_autoshrink_margin(int p_margin);
```

scene/gui/graph_frame.cpp
```diff
@@ -0,0 +1,351 @@
...
+	ClassDB::bind_method(D_METHOD("set_autoshrink_margin", "autoshrink_margin"), &GraphFrame::set_autoshrink_margin);
+	ClassDB::bind_method(D_METHOD("get_autoshrink_margin"), &GraphFrame::get_autoshrink_margin);
+
+	ClassDB::bind_method(D_METHOD("set_drag_margin", "drag_margin"), &GraphFrame::set_drag_margin);
+	ClassDB::bind_method(D_METHOD("get_drag_margin"), &GraphFrame::get_drag_margin);
+
+	ClassDB::bind_method(D_METHOD("set_tint_color_enabled", "p_enable"), &GraphFrame::set_tint_color_enabled);
+	ClassDB::bind_method(D_METHOD("is_tint_color_enabled"), &GraphFrame::is_tint_color_enabled);
+
+	ClassDB::bind_method(D_METHOD("set_tint_color", "p_color"), &GraphFrame::set_tint_color);
```
AThousandShips, line 193:
```suggestion
	ClassDB::bind_method(D_METHOD("set_tint_color", "color"), &GraphFrame::set_tint_color);
```

scene/gui/graph_frame.cpp
```diff
@@ -0,0 +1,351 @@
...
+			// Used for drawing.
+			Ref<StyleBox> sb_to_draw_panel = selected ? theme_cache.panel_selected : sb_panel;
+			Ref<StyleBox> sb_to_draw_titlebar = selected ? theme_cache.titlebar_selected : sb_titlebar;
+			Ref<StyleBoxFlat> sb_panel_flat = sb_to_draw_panel;
+			Ref<StyleBoxTexture> sb_panel_texture = sb_to_draw_panel;
+
+			Rect2 titlebar_rect(Point2(), titlebar_hbox->get_size() + sb_titlebar->get_minimum_size());
+			Size2 body_size = get_size();
+			body_size.y -= titlebar_rect.size.height;
+			Rect2 body_rect(0, titlebar_rect.size.height, body_size.width, body_size.height);
```
AThousandShips, line 110:
```suggestion
			Rect2 body_rect(Point2(0, titlebar_rect.size.height), body_size);
```

scene/gui/graph_edit.cpp
```diff
@@ -628,6 +753,118 @@ void GraphEdit::_notification(int p_what) {
...
+	Vector2 min_point{ FLT_MAX, FLT_MAX };
+	Vector2 max_point{ -FLT_MAX, -FLT_MAX };
+
+	if (!frame_attached_nodes.has(p_frame->get_name())) {
+		return Rect2();
+	}
+
+	int autoshrink_margin = p_frame->get_autoshrink_margin();
+
+	for (StringName attached_node_name : frame_attached_nodes.get(p_frame->get_name())) {
```
KoBeWi, line 766:
```suggestion
	for (const StringName &attached_node_name : frame_attached_nodes.get(p_frame->get_name())) {
```

scene/gui/graph_edit.cpp
```diff
@@ -628,6 +753,118 @@ void GraphEdit::_notification(int p_what) {
...
+		}
+	}
+}
+
+void GraphEdit::_set_drag_frame_contained_nodes(GraphFrame *p_frame, bool p_drag) {
+	if (!frame_attached_nodes.has(p_frame->get_name())) {
+		return;
+	}
+
+	for (StringName attached_node_name : frame_attached_nodes.get(p_frame->get_name())) {
```
KoBeWi, line 832:
```suggestion
	for (const StringName &attached_node_name : frame_attached_nodes.get(p_frame->get_name())) {
```

scene/gui/graph_edit.cpp
```diff
@@ -628,6 +753,118 @@ void GraphEdit::_notification(int p_what) {
...
+		}
+	}
+}
+
+void GraphEdit::_set_position_of_frame_contained_nodes(GraphFrame *p_frame, Vector2 p_drag_accum) {
+	if (!frame_attached_nodes.has(p_frame->get_name())) {
+		return;
+	}
+
+	for (StringName attached_node_name : frame_attached_nodes.get(p_frame->get_name())) {
```
KoBeWi, line 848:
```suggestion
	for (const StringName &attached_node_name : frame_attached_nodes.get(p_frame->get_name())) {
```

editor/plugins/visual_shader_editor_plugin.cpp
```diff
@@ -282,6 +290,98 @@ void VisualShaderGraphPlugin::set_expression(VisualShader::Type p_type, int p_no
...
+	graph->attach_graph_element_to_frame(itos(p_node_id), itos(p_frame_id));
+}
+
+void VisualShaderGraphPlugin::detach_node_from_frame(VisualShader::Type p_type, int p_node_id) {
+	GraphEdit *graph = editor->graph;
+	if (!graph) {
+		return;
+	}
+
+	StringName node_name = itos(p_node_id);
```
KoBeWi, line 321:
```suggestion
	const StringName node_name = itos(p_node_id);
```

editor/plugins/visual_shader_editor_plugin.cpp
```diff
@@ -1154,8 +1327,14 @@ void VisualShaderGraphPlugin::add_node(VisualShader::Type p_type, int p_id, bool
...
 void VisualShaderGraphPlugin::remove_node(VisualShader::Type p_type, int p_id, bool p_just_update) {
 	if (visual_shader->get_shader_type() == p_type && links.has(p_id)) {
-		Node *graph_edit_node = links[p_id].graph_element->get_parent();
-		graph_edit_node->remove_child(links[p_id].graph_element);
+		GraphEdit *graph_edit = editor->graph;
+		if (!graph_edit) {
+			return;
+		}
+
+		StringName graph_element_name = links[p_id].graph_element->get_name();
```
KoBeWi, line 1335: This seems unused.

editor/plugins/visual_shader_editor_plugin.cpp
```diff
@@ -2532,38 +2723,30 @@ void VisualShaderEditor::_set_node_size(int p_type, int p_node, const Vector2 &p
...
+	if (!graph_element) {
+		return;
 	}
 
+	Size2 size = graph_element->get_size();
+
 	EditorUndoRedoManager *undo_redo = EditorUndoRedoManager::get_singleton();
-	undo_redo->create_action(TTR("Resize VisualShader Node"), UndoRedo::MERGE_ENDS);
-	undo_redo->add_do_method(this, "_set_node_size", p_type, p_node, new_size);
+	undo_redo->create_action(TTR("Resize VisualShader Node"), UndoRedo::MERGE_DISABLE);
```
KoBeWi, line 2748:
```suggestion
	undo_redo->create_action(TTR("Resize VisualShader Node"));
```
It's default.

editor/plugins/visual_shader_editor_plugin.cpp
```diff
@@ -2587,95 +2770,154 @@ void VisualShaderEditor::_preview_select_port(int p_node, int p_port) {
...
+void VisualShaderEditor::_frame_title_popup_focus_out() {
+	frame_title_change_popup->hide();
 }
 
-void VisualShaderEditor::_comment_title_popup_hide() {
-	ERR_FAIL_COND(!comment_title_change_popup->has_meta("id"));
-	int node_id = (int)comment_title_change_popup->get_meta("id");
+void VisualShaderEditor::_frame_title_popup_hide() {
+	ERR_FAIL_COND(!frame_title_change_popup->has_meta("id"));
+	int node_id = (int)frame_title_change_popup->get_meta("id");
```
KoBeWi, line 2803:
```suggestion
	int node_id = (int)frame_title_change_popup->get_meta("id", -1);
	ERR_FAIL_COND(node_id == -1);
```
Slightly better (assuming -1 is invalid ID).

KoBeWi (approved):
Looks fine now (except the `autoshrink_margin` issue, but it's less important).
I left some style comments, otherwise it's good to go.

scene/gui/graph_edit.cpp
Geometror: I tried several approaches (e.g. using the z-index plus some workarounds) and this was actually the only viable solution. I don't see how can possibly fix this without a major comfpat breakage later or introducing internal nodes at arbitrary positions.

scene/gui/graph_edit.h
```diff
@@ -332,6 +343,10 @@ class GraphEdit : public Control {
 	Dictionary _get_closest_connection_at_point(const Vector2 &p_point, float p_max_distance = 4.0) const;
 	TypedArray<Dictionary> _get_connections_intersecting_with_rect(const Rect2 &p_rect) const;
 
+	Rect2 _compute_shrinked_frame_rect(const GraphFrame *p_frame);
+	void _set_drag_frame_contained_nodes(GraphFrame *p_frame, bool p_drag);
+	void _set_position_of_frame_contained_nodes(GraphFrame *p_frame, Vector2 p_pos);
```
AThousandShips, line 348:
```suggestion
	void _set_position_of_frame_contained_nodes(GraphFrame *p_frame, const Vector2 &p_pos);
```

scene/gui/graph_frame.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+	emit_signal("autoshrink_toggled", get_size());
+	queue_redraw();
+}
+
+bool GraphFrame::is_autoshrink_enabled() const {
+	return autoshrink_enabled;
+}
+
+void GraphFrame::set_autoshrink_margin(const int &p_margin) {
```
AThousandShips, line 251:
```suggestion
void GraphFrame::set_autoshrink_margin(int p_margin) {
```

scene/gui/graph_edit.cpp
```diff
@@ -463,11 +494,42 @@ void GraphEdit::_graph_element_deselected(Node *p_node) {
 	emit_signal(SNAME("node_deselected"), graph_element);
 }
 
-void GraphEdit::_graph_element_resized(Vector2 p_new_minsize, Node *p_node) {
+void GraphEdit::_graph_element_resize_request(Vector2 p_new_minsize, Node *p_node) {
```
AThousandShips, line 497:
```suggestion
void GraphEdit::_graph_element_resize_request(const Vector2 &p_new_minsize, Node *p_node) {
```

scene/gui/graph_edit.cpp
```diff
@@ -463,11 +494,42 @@ void GraphEdit::_graph_element_deselected(Node *p_node) {
...
+	// Update all parent frames recursively bottom-up.
+	if (linked_parent_map.has(graph_element->get_name())) {
+		GraphFrame *parent_frame = Object::cast_to<GraphFrame>(get_node_or_null(NodePath(linked_parent_map[graph_element->get_name()])));
+		if (parent_frame) {
+			_update_graph_frame(parent_frame);
+		}
+	}
+}
+
+void GraphEdit::_graph_frame_shrink_toggled(Vector2 p_new_minsize, GraphFrame *p_frame) {
```
AThousandShips, line 526:
```suggestion
void GraphEdit::_graph_frame_shrink_toggled(const Vector2 &p_new_minsize, GraphFrame *p_frame) {
```

scene/gui/graph_edit.cpp
```diff
@@ -628,6 +751,118 @@ void GraphEdit::_notification(int p_what) {
...
+
+		Vector2 node_pos = attached_node->get_position_offset();
+		Vector2 size = attached_node->get_size();
+		min_point = min_point.min(node_pos);
+		max_point = max_point.max(node_pos + size);
+	}
+
+	// It's sufficient to check only one value here.
+	if (min_point.x == FLT_MAX) {
+		return Rect2(p_frame->get_position_offset(), Size2(0, 0));
```
AThousandShips, line 782:
```suggestion
		return Rect2(p_frame->get_position_offset(), Size2());
```

scene/gui/graph_edit.cpp
```diff
@@ -628,6 +751,118 @@ void GraphEdit::_notification(int p_what) {
...
+
+		attached_node->set_drag(p_drag);
+		GraphFrame *graph_frame = Object::cast_to<GraphFrame>(attached_node);
+		if (graph_frame) {
+			_set_drag_frame_contained_nodes(graph_frame, p_drag);
+		}
+	}
+}
+
+void GraphEdit::_set_position_of_frame_contained_nodes(GraphFrame *p_frame, Vector2 p_drag_accum) {
```
AThousandShips, line 841:
```suggestion
void GraphEdit::_set_position_of_frame_contained_nodes(GraphFrame *p_frame, const Vector2 &p_pos) {
```

AThousandShips (approved):

scene/gui/graph_frame.h
Geometror: I just did a quick search and it looks like prefixing structs defined in cpp files (or private inner structs) with _ seems to be common practice in our code base, even in newer code.

KoBeWi: Huh. Then it's used inconsistently 🤔

lyuma:
@Geometror would it be possible to resolve the conflict with extension_api_validation / squash?

(I believe the conflict was caused by an urgent bugfix with Node::replace_by in #89992)

akien-mga:
Thanks!

scene/gui/graph_frame.cpp
```diff
@@ -0,0 +1,357 @@
...
+			// Draw body stylebox.
+			if (tint_color_enabled) {
+				if (sb_panel_flat.is_valid()) {
+					Color original_border_color = sb_panel_flat->get_border_color();
+					sb_panel_flat = sb_panel_flat->duplicate();
+					sb_panel_flat->set_bg_color(tint_color);
+					sb_panel_flat->set_border_color(selected ? original_border_color : tint_color.lightened(0.3));
+					draw_style_box(sb_panel_flat, body_rect);
+				} else if (sb_panel_texture.is_valid()) {
+					sb_panel_texture = sb_panel_flat->duplicate();
```
Zylann, line 121: This code might crash, as highlighted by its failure to compile on my Linux CI: https://github.com/Zylann/godot_voxel/actions/runs/8560927646/job/23461023064

akien-mga: Reading the code, I think the two identifiers might be swapped bg mistake?

lyuma:
>Remove VisualShaderNodeComment and replace it with VisualShaderNodeFrame

@Geometror Removing this part of our API is causing problems with some GDExtension plugins: it wasn't marked experimental, so this breaks all rust plugins since they hard-link to the whole API. See #90303

I think the easiest way to solve this is to add a stub `class VisualShaderNodeComment : public VisualShaderNodeFrame` so it will work just like the new `VisualShaderNodeFrame`. If there is a way to register the GDCLASS without it showing up in the resource picker, that would be best.

Even though it extends from `VisualShaderNodeFrame`, it might also be necessary to implement `bind_compatibility_methods()` for compat versions of set/get_title and set/get_description that call the real versions of those methods.

eobet:
It would be easier to find this if when you search for "comment" it would also bring up the frame...

PR #88019, opened 2024-02-06
Fix `get_contact_count` sometimes being incorrect in `_process`

### LINKED ISSUE #88018: `get_contact_count` yields different results from `_process` and `_physics_process`
### Tested versions

Reproducible in: 4.3.dev [d3352813e]

### System information

Windows 11 (10.0.22631)

### Issue description

This is essentially the same problem as #87996, but for a different method.

Due to the current implementation of `RigidBody*D.get_contact_count` relying on `body_get_direct_state` you will end up with desync between the contact count reported by `_physics_process` and the contact count reported by `_process`, within the same physics frame.

This means that when a `RigidBody*D` makes contact with something, `_process` will pick up on the change before `_physics_process` does, while they still report the same position in the world, which can potentially lead to subtle and hard-to-track-down bugs.

The reason for this is because of how a frame is laid out in Godot, which in very simplified terms goes something like this:

- Maybe do the physics tick
    - Pull state from the physics server
    - Run `_physics_process`
    - Step the physics simulation/server
    - Maybe repeat
- Do the idle tick
    - Run `_process`
- Render, audio, input, etc.

The important thing to note here is that the physics simulation/server is stepped inbetween `_physics_process` and `_process`, which means that the physics simulation/server is always one tick ahead of `_physics_process` while `_process` is running.

This means that if you use something like `body_get_direct_state`, which as the name would suggest gives you direct access to the physics server, you will bypass the state synchronization that typically happens before `_physics_process` and access the most current (future) value in the physics simulation/server instead, leading to a desync with `_physics_process`.

### Steps to reproduce

- Run the MRP
- Note the output log when the body hits the floor

[ContactCountDesync.zip](https://github.com/godotengine/godot/files/14180458/ContactCountDesync.zip)

mihe:
Fixes #88018.

This replaces the usage of `body_get_direct_state` in `RigidBody*D.get_contact_count` with a cached value that's populated as part of `RigidBody*D._sync_body_state`, similar to other states in the physics bodies, which prevents `_process` from accessing a yet-to-be-synced future contact count.

This also fixes the inability to use `get_contact_count` from `_process` when using the `physics/*d/run_on_separate_thread` project setting.

### DIFF
--- a/scene/2d/physics_body_2d.cpp
+++ b/scene/2d/physics_body_2d.cpp
@@ -446,6 +446,8 @@ void RigidBody2D::_sync_body_state(PhysicsDirectBodyState2D *p_state) {
 	linear_velocity = p_state->get_linear_velocity();
 	angular_velocity = p_state->get_angular_velocity();
 
+	contact_count = p_state->get_contact_count();
+
 	if (sleeping != p_state->is_sleeping()) {
 		sleeping = p_state->is_sleeping();
 		emit_signal(SceneStringNames::get_singleton()->sleeping_state_changed);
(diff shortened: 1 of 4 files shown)


mihe:
Yes, that's correct. Contact counts are solely generated/modified during the simulation step.

akien-mga (approved):
Makes sense to me.

akien-mga:
Thanks!

PR #88020, opened 2024-02-06
GDScript: Set `has_type` false if it is `BUILTIN` but `Variant::NIL`

### LINKED ISSUE #87932: `match null` crashes GDScript compiler
### Tested versions

- Reproducible in: 4.2.1, 4.3 dev

### System information

Kubuntu 23.10

### Issue description

```gdscript
extends Node

func _ready() -> void:
    match null:
        _:
            pass
```

```
#0  0x0000555558ce1bab in CowData<GDScriptByteCodeGenerator::StackSlot>::get (this=0x55557aebe148, p_index=4) at ./core/templates/cowdata.h:184
#1  0x0000555558cdcdcf in Vector<GDScriptByteCodeGenerator::StackSlot>::operator[] (this=0x55557aebe140, p_index=4) at ./core/templates/vector.h:97
#2  0x0000555558cd65b6 in GDScriptByteCodeGenerator::write_call_utility (this=0x55557aebe0c0, p_target=..., p_function=..., p_arguments=...) at modules/gdscript/gdscript_byte_codegen.cpp:1098
#3  0x0000555558b654b8 in GDScriptCompiler::_parse_block (this=0x7fffffff9ab0, codegen=..., p_block=0x55557db37b00, p_add_locals=true, p_reset_locals=false) at modules/gdscript/gdscript_compiler.cpp:1904
#4  0x0000555558b67d69 in GDScriptCompiler::_parse_function (this=0x7fffffff9ab0, r_error=@0x7fffffff97a0: OK, p_script=0x55557ae6da00, p_class=0x55557d9f5d40, p_func=0x55557d273640, p_for_ready=false, p_for_lambda=false) at modules/gdscript/gdscript_compiler.cpp:2356
#5  0x0000555558b6b32e in GDScriptCompiler::_compile_class (this=0x7fffffff9ab0, p_script=0x55557ae6da00, p_class=0x55557d9f5d40, p_keep_state=true) at modules/gdscript/gdscript_compiler.cpp:2874
#6  0x0000555558b6cde3 in GDScriptCompiler::compile (this=0x7fffffff9ab0, p_parser=0x7fffffff9b40, p_script=0x55557ae6da00, p_keep_state=true) at modules/gdscript/gdscript_compiler.cpp:3198
```

https://github.com/godotengine/godot/blob/b4e2a24c1f62088b3f7ce0197afc90832fc25009/modules/gdscript/gdscript_compiler.cpp#L1896-L1904

https://github.com/godotengine/godot/blob/b4e2a24c1f62088b3f7ce0197afc90832fc25009/modules/gdscript/gdscript_byte_codegen.cpp#L1095-L1101

### Steps to reproduce

Save the script above.

N/A

emre0altan:
This change fixes #87932 by preventing edge case for **null** value when used with a **match**. Even though edge case is happening with third check in **IS_BUILTIN_TYPE()**, `m_var.type.builtin_type == m_type`, there is no other default type other than **Variant::NIL**, so this check wouldn't change. But returning false with **has_type** in the first check for **Variant::NIL** makes more sense since it is also the default value in **GDScriptDataType**.  

https://github.com/godotengine/godot/blob/b4e2a24c1f62088b3f7ce0197afc90832fc25009/modules/gdscript/gdscript_byte_codegen.cpp#L430-L434

https://github.com/godotengine/godot/blob/b4e2a24c1f62088b3f7ce0197afc90832fc25009/modules/gdscript/gdscript_function.h#L47-L65

### DIFF
--- a/modules/gdscript/gdscript_byte_codegen.cpp
+++ b/modules/gdscript/gdscript_byte_codegen.cpp
@@ -429,7 +429,7 @@ void GDScriptByteCodeGenerator::set_initial_line(int p_line) {
 	(m_var.type.has_type && m_var.type.kind == GDScriptDataType::BUILTIN)
 
 #define IS_BUILTIN_TYPE(m_var, m_type) \
-	(m_var.type.has_type && m_var.type.kind == GDScriptDataType::BUILTIN && m_var.type.builtin_type == m_type)
+	(m_var.type.has_type && m_var.type.kind == GDScriptDataType::BUILTIN && m_var.type.builtin_type == m_type && m_type != Variant::NIL)
 
 void GDScriptByteCodeGenerator::write_type_adjust(const Address &p_target, Variant::Type p_new_type) {
 	switch (p_new_type) {
--- a/modules/gdscript/tests/scripts/runtime/features/match_test_null.gd
+++ b/modules/gdscript/tests/scripts/runtime/features/match_test_null.gd
@@ -1,0 +1,6 @@
+func test():
+	match null:
+		null:
+			print('null matched')
+		_:
+			pass
--- a/modules/gdscript/tests/scripts/runtime/features/match_test_null.out
+++ b/modules/gdscript/tests/scripts/runtime/features/match_test_null.out
@@ -1,0 +1,2 @@
+GDTEST_OK
+null matched


dalexeev:
This makes sense since `BUILTIN` types cannot accept `null` value and there is no `Nil` type in GDScript. But in my opinion, this should be fixed first in the analyzer (`GDScriptAnalyzer::type_from_variant()`). In the compiler we can add a check in case of a bug in the analyzer, like here:

https://github.com/godotengine/godot/blob/2e684ef4f7e706ac3990e7431db84e30a4d81f3e/modules/gdscript/gdscript_compiler.cpp#L121-L126

dalexeev:
> There is no other type other than `Variant::NIL` that can represent `null`

Sorry, I was thinking that we could use `Variant::OBJECT`, but I forgot that it is also not suitable for `BUILTIN`, and `NATIVE`/`SCRIPT`/`CLASS` is ambiguous for `null`, it can cause unexpected consequences. So I guess we can leave it as is, just add a comment and test.

Or, if look at it differently, GDScript **has** an implicit type `Nil`, it's just that the user can't specify it. Perhaps the bug can be fixed differently, in `GDScriptByteCodeGenerator`.

emre0altan:
I think the one you referring is this one [https://docs.godotengine.org/en/3.1/classes/class_nil.html](url) and probably it's the default version of Variant with default NIL type. My gut feeling says that only difference from the other types is just `null` is more commonly used than `nil` and being used as `null`. But problem arises when `NIL` is treated the same way with the other types and also being used as the default type like in `Variant::get_utility_function_argument_type()`. Later if you used it to check something, then there will be an edge case for `NIL`.

https://github.com/godotengine/godot/blob/d3352813ea44447bfbf135efdec23acc4d1d3f89/core/variant/variant_utility.cpp#L1922-L1929

modules/gdscript/gdscript_compiler.cpp
```diff
@@ -100,6 +100,9 @@ GDScriptDataType GDScriptCompiler::_gdtype_from_datatype(const GDScriptParser::D
 		case GDScriptParser::DataType::BUILTIN: {
 			result.kind = GDScriptDataType::BUILTIN;
 			result.builtin_type = p_datatype.builtin_type;
+			// Fixes GH-87932. Variant::NIL used as default type when getting
+			// the argument type and has_type(false) will be the edge case fix
```
AThousandShips, line 104:
```suggestion
			// the argument type and has_type(false) will be the edge case fix.
```

modules/gdscript/tests/scripts/parser/features/match_test_null.gd
AThousandShips, line 1: Mixed tabs and spaces for indentation

dalexeev, line 1: This test should be placed in `runtime/`, not `parser/`.

dalexeev (commented):
I'd prefer to solve it here:

https://github.com/godotengine/godot/blob/b4e2a24c1f62088b3f7ce0197afc90832fc25009/modules/gdscript/gdscript_byte_codegen.cpp#L1086

```diff
-			if (!IS_BUILTIN_TYPE(p_arguments[i], Variant::get_utility_function_argument_type(p_function, i))) {
+			const Variant::Type arg_type = Variant::get_utility_function_argument_type(p_function, i);
+			// `Variant::get_utility_function_argument_type()` returns `Variant::NIL` for `Variant` arguments.
+			if (arg_type == Variant::NIL || !IS_BUILTIN_TYPE(p_arguments[i], arg_type)) {
```

We have some problem distinguishing between `Nil` and `Variant` (there is `PROPERTY_USAGE_NIL_IS_VARIANT`). In some cases, `Variant::NIL` is used as the default value, in others `Variant::VARIANT_MAX`.

modules/gdscript/tests/scripts/parser/features/match_test_null.gd
emre0altan: Thanks for the comments, updated now.

dalexeev (approved):
Looks good to me, but I'd appreciate a double check. We fixed the crash for now, but I'm not sure there aren't similar cases elsewhere. Perhaps we should change the `IS_BUILTIN_TYPE` macro or use `Variant::VARIANT_MAX` as the default value instead of `Variant::NIL`.

emre0altan:
Thanks, I'll check it again. I agree, it could be better to add it to `IS_BUILTIN_TYPE` and I was mistaken about where the default value comes from. It returns `NIL` but it didn't change when I changed the return value to `VARIANT_MAX` here:
https://github.com/godotengine/godot/blob/d3352813ea44447bfbf135efdec23acc4d1d3f89/core/variant/variant_utility.cpp#L1922-L1929
I tried to trace the registered function but it got complicated quickly. I'll have a second look there and try to find out if it's better to return `VARIANT_MAX`.

emre0altan:
I think that if a method has a `Variant` parameter, this parameter returns `Variant::NIL` from `get_arg_type()` and that is where it is used as a default value. But it is not only for `typeof` method, it is like this for any method with a `Variant` parameter. So I don't think we can change the default value but I think it is better to fix this issue by changing `IS_BUILTIN_TYPE` macro.
https://github.com/godotengine/godot/blob/master/doc/classes/%40GlobalScope.xml#L1409-L1425

vnen (approved):

akien-mga:

emre0altan:
Thank you🙂

PR #88011, opened 2024-02-06
vulkan: Update all components to Vulkan SDK 1.3.275.0

Chubercik:
`vulkan`:

- `Vulkan-Headers`: https://github.com/KhronosGroup/Vulkan-Headers/releases/tag/vulkan-sdk-1.3.275.0
- `Vulkan-Utility-Libraries`: https://github.com/KhronosGroup/Vulkan-Utility-Libraries/releases/tag/vulkan-sdk-1.3.275.0
  - `patches` have been regenerated
- `VulkanMemoryAllocator`: https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/commit/38627f4e37d7a9b13214fd267ec60e0e877e3997

`volk`:

- https://github.com/zeux/volk/releases/tag/vulkan-sdk-1.3.275.0

`glslang`:

- https://github.com/KhronosGroup/glslang/releases/tag/vulkan-sdk-1.3.275.0
  - `OGLCompiler` stub library has been fully removed from the build and won't be included from this point forward

`spirv-reflect`:

- https://github.com/KhronosGroup/SPIRV-Reflect/releases/tag/vulkan-sdk-1.3.275.0
  - `patches` have been regenerated

PR #88021, opened 2024-02-06
Fully initialize all members of structs `IdentifierActions`, `GeneratedCode` and `DefaultIdentifierActions`

### LINKED ISSUE #88015: Compilation in dev mode fails on a warning about using an uninitialized variable
### Tested versions

Reproducible in 4.3.dev (https://github.com/godotengine/godot/commit/d3352813ea44447bfbf135efdec23acc4d1d3f89)

### System information

Windows 11 - Godot 4.3.dev

### Issue description

When upgrading Vulkan SDK to version 1.3.275.0 I came across the following issue:

When I try to compile the `master` branch with the `use_mingw=yes platform=windows dev_mode=yes` SCons options set, it fails at around 93% and prints the following info:

```
[ 93%] =====
In file included from servers\rendering\dummy\storage\material_storage.h:34,
                 from servers\rendering\dummy\storage\material_storage.cpp:31:
In copy constructor 'ShaderCompiler::DefaultIdentifierActions::DefaultIdentifierActions(const ShaderCompiler::DefaultIdentifierActions&)',
    inlined from 'RendererDummy::MaterialStorage::MaterialStorage()' at servers\rendering\dummy\storage\material_storage.cpp:40:27:
./servers/rendering/shader_compiler.h:89:16: error: 'actions' is used uninitialized [-Werror=uninitialized]
   89 |         struct DefaultIdentifierActions {
      |                ^~~~~~~~~~~~~~~~~~~~~~~~
servers\rendering\dummy\storage\material_storage.cpp: In constructor 'RendererDummy::MaterialStorage::MaterialStorage()':
servers\rendering\dummy\storage\material_storage.cpp:39:50: note: 'actions' declared here
   39 |         ShaderCompiler::DefaultIdentifierActions actions;
      |                                                  ^~~~~~~
cc1plus.exe: all warnings being treated as errors

=====
[ 98%] scons: *** [servers\rendering\dummy\storage\material_storage.windows.editor.x86_64.o] Error 1
scons: building terminated because of errors.
[Time elapsed: 00:06:23.054]
```

Not sure if this information is relevant to this issue, but this is emitted when compilation starts:

```
py -3.9 -m SCons use_mingw=yes platform=windows dev_mode=yes
```

```
scons: Reading SConscript files ...
Auto-detected 8 CPU cores available for build parallelism. Using 7 cores by default. You can override it with the -j argument.
Using MinGW, arch x86_64
Building for platform "windows", architecture "x86_64", target "editor".
Checking for C header file mntent.h... (cached) no
```

### Steps to reproduce

- Try to build the source from the commit mentioned above with the following options set:
  - `use_mingw=yes platform=windows dev_mode=yes`

N/A

Chubercik:
Fixes https://github.com/godotengine/godot/issues/88015.

This one simple change

```diff
- ShaderCompiler::DefaultIdentifierActions actions;
+ ShaderCompiler::DefaultIdentifierActions actions = {};
```

is enough for Godot to compile on my setup, however, @akien-mga pointed out that this might be an issue in other places in the code [here](https://github.com/godotengine/godot/pull/88011#discussion_r1479644005). He proposed an alternative fix (changing the struct code to fully initialize all its members), but I'd like to wait for the rendering team to voice their opinions about this.

**EDIT**: I went ahead with the alternative fix (and then some more for completeness' sake).

### DIFF
--- a/servers/rendering/shader_compiler.h
+++ b/servers/rendering/shader_compiler.h
@@ -52,38 +52,38 @@ public:
 		HashMap<StringName, bool *> usage_flag_pointers;
 		HashMap<StringName, bool *> write_flag_pointers;
 
-		HashMap<StringName, ShaderLanguage::ShaderNode::Uniform> *uniforms;
+		HashMap<StringName, ShaderLanguage::ShaderNode::Uniform> *uniforms = nullptr;
 	};
 
 	struct GeneratedCode {
 		Vector<String> defines;
 		struct Texture {
 			StringName name;
-			ShaderLanguage::DataType type;
-			ShaderLanguage::ShaderNode::Uniform::Hint hint;
+			ShaderLanguage::DataType type = ShaderLanguage::DataType::TYPE_VOID;
+			ShaderLanguage::ShaderNode::Uniform::Hint hint = ShaderLanguage::ShaderNode::Uniform::Hint::HINT_NONE;
 			bool use_color = false;
-			ShaderLanguage::TextureFilter filter;
-			ShaderLanguage::TextureRepeat repeat;
-			bool global;
-			int array_size;
+			ShaderLanguage::TextureFilter filter = ShaderLanguage::TextureFilter::FILTER_DEFAULT;
+			ShaderLanguage::TextureRepeat repeat = ShaderLanguage::TextureRepeat::REPEAT_DEFAULT;
+			bool global = false;
+			int array_size = 0;
 		};
 
 		Vector<Texture> texture_uniforms;
 
 		Vector<uint32_t> uniform_offsets;
-		uint32_t uniform_total_size;
+		uint32_t uniform_total_size = 0;
 		String uniforms;
 		String stage_globals[STAGE_MAX];
 
 		HashMap<String, String> code;
 
-		bool uses_global_textures;
-		bool uses_fragment_time;
-		bool uses_vertex_time;
-		bool uses_screen_texture_mipmaps;
-		bool uses_screen_texture;
-		bool uses_depth_texture;
-		bool uses_normal_roughness_texture;
+		bool uses_global_textures = false;
+		bool uses_fragment_time = false;
+		bool uses_vertex_time = false;
+		bool uses_screen_texture_mipmaps = false;
+		bool uses_screen_texture = false;
+		bool uses_depth_texture = false;
+		bool uses_normal_roughness_texture = false;
 	};
 
 	struct DefaultIdentifierActions {
@@ -91,8 +91,8 @@ public:
 		HashMap<StringName, String> render_mode_defines;
 		HashMap<StringName, String> usage_defines;
 		HashMap<StringName, String> custom_samplers;
-		ShaderLanguage::TextureFilter default_filter;
-		ShaderLanguage::TextureRepeat default_repeat;
+		ShaderLanguage::TextureFilter default_filter = ShaderLanguage::TextureFilter::FILTER_DEFAULT;
+		ShaderLanguage::TextureRepeat default_repeat = ShaderLanguage::TextureRepeat::REPEAT_DEFAULT;
 		int base_texture_binding_index = 0;
 		int texture_layout_set = 0;
 		String base_uniform_string;


akien-mga:
Yeah I think you might have similar warnings in the other places if you do a clean build. At any rate, it doesn't make much sense IMO to initialize it only here and not everywhere else. So I think changing the struct to make sure it's always fully initialized would be the best option.

WDYT @clayjohn @BastiaanOlij ?

AThousandShips:
I think the issue might be the two texture modes in this struct

Chubercik:
So

```c++
ShaderLanguage::TextureFilter default_filter = ShaderLanguage::TextureFilter::FILTER_DEFAULT;
ShaderLanguage::TextureRepeat default_repeat = ShaderLanguage::TextureRepeat::REPEAT_DEFAULT;
```

should fix the issue?

AThousandShips:
I'd suggest initialising the values in the `IdentifierActions` struct as well while at it, like the pointer there

Chubercik:
Think I'm gonna need some help here - I assume `uniforms` should be initialized to `nullptr`, given that it's a pointer, but what about the other values you speak of: they're of type `HashMap`, not sure what the syntax would look like here..

AThousandShips:
Not those, the `bool` values, the maps should be untouched they have default initialization

Chubercik:
Are we talking about the same bools?

```c++
HashMap<StringName, Pair<int *, int>> render_mode_values;
HashMap<StringName, bool *> render_mode_flags;
HashMap<StringName, bool *> usage_flag_pointers;
HashMap<StringName, bool *> write_flag_pointers;
```

I considered these lines of code, but the `bool` (and one `int`) pointers are inside HashMaps and I'm not sure what would initialization of those alone look like.

```c++
bool uses_global_textures;
bool uses_fragment_time;
bool uses_vertex_time;
bool uses_screen_texture_mipmaps;
bool uses_screen_texture;
bool uses_depth_texture;
bool uses_normal_roughness_texture;
```

If you were talking about these bools, they're in a different struct 😅

AThousandShips:
Yes I was talking about the `GeneratedCode` my bad :)

AThousandShips:
I'd say yes, prefer to have everything basically intialized, `false` for `bool`, `0` for `int` and what ever default enum value applies

clayjohn:
> Yeah I think you might have similar warnings in the other places if you do a clean build. At any rate, it doesn't make much sense IMO to initialize it only here and not everywhere else. So I think changing the struct to make sure it's always fully initialized would be the best option.
> 
> WDYT @clayjohn @BastiaanOlij ?

This case is a bit unique. I think it is fully initialized in other areas property-by-property. The difference is that here we don't care about the actions so we left them all empty.

Chubercik:
Hope I picked sane defaults for enums.

BastiaanOlij:
> Yeah I think you might have similar warnings in the other places if you do a clean build. At any rate, it doesn't make much sense IMO to initialize it only here and not everywhere else. So I think changing the struct to make sure it's always fully initialized would be the best option.
> 
> WDYT @clayjohn @BastiaanOlij ?

I'm not against it, I've ran into problems before with compiler difference where some zero out structs and others keep junk data in there and caused sudden crashes. So if we're checking on that with compiler checks, might as well. I think the only issue with defining defaults in the structs is that there may be confusion when structs come from 3rd party APIs (like Vulkan and OpenXR). But we don't really have control over that.

RandomShaper:
> I'd say yes, prefer to have everything basically intialized, `false` for `bool`, `0` for `int` and what ever default enum value applies

I agree and would like to add that, when the default value for an enum type is not relevant (you just want to avoid garbage) or when it's clear (somewhat fuzzy, I know) what's the natual default, I'd also find `= {}` acceptable.

Regarding 3rd-part structs, in the rendering driver code, for instance, I've chosen to do `= {}` on every Vulkan and D3D12 structs, to ensure C-like APIs don't provide garbage data.

akien-mga (approved):
Looks good to me.

Rendering headers are some of the last ones where we don't systematically initialize every struct members, and they have *a lot* of structs. So I'd like someone from the rendering team to sign off on this before merging, just in case I'm missing good reasons to keep those uninitialized.

clayjohn (approved):

akien-mga:
Thanks!

PR #88024, opened 2024-02-06
Improve documentation on Android package unique name

Calinou:
- Document `$genname` behavior.
- Mention that changing the package name invalidates existing installations (i.e. new data folder is used, and the app is installed separately instead of updating existing installations).
- Update to match actual validation and Google Play guidelines.

See https://forum.godotengine.org/t/genname-token-in-export-presets/46280.


AThousandShips (approved):
Approving on style

m4gr3d (approved):
Thanks for the update!

akien-mga:
Thanks!

akien-mga:
Cherry-picked for 4.2.2.
Cherry-picked for 4.1.4.

PR #81906, opened 2023-09-19
Set `open_dominant_script_on_scene_change` to off by default

snibk:
Super tiny change - just changes the default for the `text_editor/behavior/files/open_dominant_script_on_scene_change` editor settings option to `false`.

This behavior is super unintuitive imo. When a user tabs into a scene, they may not necessarily want to edit the dominant script in that scene, but this functionality seems to assume they always do. This has the side effect (in my experience) of making the list of open scripts super messy as scripts I never intended to edit are opened in the background, requiring me to "Close All" on the list often if I want to maintain any clarity on what I am actually working on.

From https://github.com/godotengine/godot-proposals/issues/7434 it seems like others have the same feeling that this behavior is odd, however I think it makes sense to leave the option there for those who do find it useful.

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/8372*

PR #88026, opened 2024-02-06
Enable Add Type Hints editor setting by default

Calinou:
Now that GDScript type hints improve performance since Godot 4.0 and the community is increasingly getting used to typed GDScript, it makes sense to add type hints by default in autocompletion and script templates.

[Official demos will also be moving to type hints at some point in the future](https://github.com/godotengine/godot-demo-projects/issues/868), further increasing the relevance of enabling type hints out of the box.

Additional context: https://twitter.com/reduzio/status/1754794206206239036 (although this has been a general trend since 4.0's release, and even before)

### DIFF
--- a/doc/classes/EditorSettings.xml
+++ b/doc/classes/EditorSettings.xml
@@ -888,7 +888,7 @@
 			[b]Note:[/b] You can hold down [kbd]Alt[/kbd] while using the mouse wheel to temporarily scroll 5 times faster.
 		</member>
 		<member name="text_editor/completion/add_type_hints" type="bool" setter="" getter="">
-			If [code]true[/code], adds static typing hints such as [code]-&gt; void[/code] and [code]: int[/code] when using code autocompletion or when creating onready variables by drag and dropping nodes into the script editor while pressing the [kbd]Ctrl[/kbd] key.
+			If [code]true[/code], adds [url=$DOCS_URL/tutorials/scripting/gdscript/static_typing.html]GDScript static typing[/url] hints such as [code]-&gt; void[/code] and [code]: int[/code] when using code autocompletion or when creating onready variables by drag and dropping nodes into the script editor while pressing the [kbd]Ctrl[/kbd] key. If [code]true[/code], newly created scripts will also automatically have type hints added to their method parameters and return types.
 		</member>
 		<member name="text_editor/completion/auto_brace_complete" type="bool" setter="" getter="">
 			If [code]true[/code], automatically completes braces when making use of code completion.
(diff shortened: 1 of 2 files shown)


AThousandShips (approved):
I'm in favour as part of a broader push to prefer static typing, and phrasing is good

Unless I'm mistaken editor settings are all stored so this won't affect existing users, so anyone can just disable it if they dislike it

Mickeon (approved):
I'm fine with this. Although I have always felt like this option could use a bit more granularity to it.

Mickeon:
Since this changes a default Editor Setting, I would encourage whoever is reading this to also check out https://github.com/godotengine/godot/pull/81906

dalexeev:
I see little risk in this (unlike `UNTYPED_DECLARATION`, `INFERRED_DECLARATION`, and `UNSAFE_*` warnings). This only affects autocomplete methods like `_ready()`, users can still use untyped code. GDScript is a gradually typed language and should handle mixed code gracefully.

However, does this mean that our recommendations are shifting towards typed code, although we still use [only untyped code](https://docs.godotengine.org/en/stable/contributing/documentation/docs_writing_guidelines.html#dynamic-vs-static-typing) in the documentation? Also, we [recommend not mixing approaches](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/static_typing.html#typed-or-dynamic-stick-to-one-style), and changing the setting prevents users of untyped GDScript from sticking to one option.

Mickeon:
"Not mixing approaches" feels like a weird rule. I get it's meant to encourage consistency, but GDScript works completely fine with a mixed approach.
In fact, most users probably have some form of "mixture" in their codebase. That's especially true if they start programming untyped, and gradually use types as they gain experience (I can relate).

In my personal opinion the documentation should still aim to be accessible and avoid static typing when possible, as it makes the code look more overwhelming than it actually is in practice.
A lot of times `=` can be substituted with `:=` and experienced developers know this already.

adamscott:
cc. @gdquest @Xananax

SanderVanhove:
I'm a big fan of type hints and we use it nearly everywhere on Koira. We do have, as mentioned, a mixed approach though, because sometimes you are not 100% sure what type is coming in or you want to be able to support multiple types and switch within the function.

Although type hints can seem overwhelming, it helps beginners a lot with autocompletion and finding bugs before you run the game. Enabling this feature by default gives a nudge towards good programming practices.

Xananax:
Funny, I just fished [this old related article](https://existentialtype.wordpress.com/2011/03/19/dynamic-languages-are-static-languages/) back for a conversation I was having with a friend.

From an educator perspective, my go-to recommendation is that dynamic typing should only be handled by experts, while beginners need explicit types.

My personal anecdotal experience is that types are scary for newcomers, but once the hurdle passed, they are _easier_ on the slightly longer term[^1]. The concept of "this keywords is a filter that validates other values are acceptable to put in that box" is easy. The syntax is complicated.

For that reason, I am all for types by default, but I am _also_ in favor of leaving them out of the initial tutorials (`:=` is fine). Ideally, a switch above code boxes could add/remove types (similarly to GDScript/C# currently).

[^1]: Side note: This is what I've subjectively witnessed by teaching students IRL and online. In the literature though, there is no evidence of types being easier or harder for beginners (or for that matter, better or worse for anything in general, regardless of the metric chosen).

akien-mga:
Thanks!

PR #88027, opened 2024-02-06
Direct 3D 12: Implement proper fallback for format casting

RandomShaper:
In short, this removes the ugly aliasing hack used when relaxed casting is not available and adds the proper reinterpret-copy or copy-via-buffer approach.

In other words, this PR fixes SDFGI for those drivers not implementing relaxed casting and where the current hack was not working properly.

Includes #87872.

**UPDATE:** I've updated the code disabling relaxed casting for now. (See https://github.com/godotengine/godot/pull/88027#issuecomment-1934199704.)

### DIFF
--- a/drivers/d3d12/rendering_device_driver_d3d12.cpp
+++ b/drivers/d3d12/rendering_device_driver_d3d12.cpp
@@ -530,7 +530,7 @@ static const D3D12_RESOURCE_DIMENSION RD_TEXTURE_TYPE_TO_D3D12_RESOURCE_DIMENSIO
 	D3D12_RESOURCE_DIMENSION_TEXTURE2D,
 };
 
-void RenderingDeviceDriverD3D12::_resource_transition_batch(ResourceInfo *p_resource, uint32_t p_subresource, uint32_t p_num_planes, D3D12_RESOURCE_STATES p_new_state, ID3D12Resource *p_resource_override) {
+void RenderingDeviceDriverD3D12::_resource_transition_batch(ResourceInfo *p_resource, uint32_t p_subresource, uint32_t p_num_planes, D3D12_RESOURCE_STATES p_new_state) {
 	DEV_ASSERT(p_subresource != UINT32_MAX); // We don't support an "all-resources" command here.
 
 #ifdef DEBUG_COUNT_BARRIERS
@@ -538,9 +538,16 @@ void RenderingDeviceDriverD3D12::_resource_transition_batch(ResourceInfo *p_reso
 #endif
 
 	ResourceInfo::States *res_states = p_resource->states_ptr;
-	D3D12_RESOURCE_STATES *curr_state = &res_states->subresource_states[p_subresource];
 
-	ID3D12Resource *res_to_transition = p_resource_override ? p_resource_override : p_resource->resource;
+	if (p_new_state == D3D12_RESOURCE_STATE_UNORDERED_ACCESS) {
+		if (unlikely(!res_states->xfamily_fallback.subresources_dirty.is_empty())) {
+			uint32_t subres_qword = p_subresource / 64;
+			uint64_t subres_mask = (uint64_t(1) << (p_subresource % 64));
+			res_states->xfamily_fallback.subresources_dirty[subres_qword] |= subres_mask;
+		}
+	}
+
+	D3D12_RESOURCE_STATES *curr_state = &res_states->subresource_states[p_subresource];
 
 	// Transitions can be considered redundant if the current state has all the bits of the new state.
 	// This check does not apply to the common state however, which must resort to checking if the state is the same (0).
@@ -553,7 +560,7 @@ void RenderingDeviceDriverD3D12::_resource_transition_batch(ResourceInfo *p_reso
 			if (res_barriers.size() < res_barriers_count + 1) {
 				res_barriers.resize(res_barriers_count + 1);
 			}
-			res_barriers[res_barriers_count] = CD3DX12_RESOURCE_BARRIER::UAV(res_to_transition);
+			res_barriers[res_barriers_count] = CD3DX12_RESOURCE_BARRIER::UAV(p_resource->resource);
 			res_barriers_count++;
 			res_states->last_batch_with_uav_barrier = res_barriers_batch;
 		}
@@ -563,7 +570,7 @@ void RenderingDeviceDriverD3D12::_resource_transition_batch(ResourceInfo *p_reso
 
 		if (res_barriers_requests.has(res_states)) {
 			BarrierRequest &br = res_barriers_requests.get(res_states);
-			DEV_ASSERT(br.dx_resource == res_to_transition);
+			DEV_ASSERT(br.dx_resource == p_resource->resource);
 			DEV_ASSERT(br.subres_mask_qwords == STEPIFY(res_states->subresource_states.size(), 64) / 64);
 			DEV_ASSERT(br.planes == p_num_planes);
 
@@ -681,7 +688,7 @@ void RenderingDeviceDriverD3D12::_resource_transition_batch(ResourceInfo *p_reso
 			}
 		} else {
 			BarrierRequest &br = res_barriers_requests[res_states];
-			br.dx_resource = res_to_transition;
+			br.dx_resource = p_resource->resource;
 			br.subres_mask_qwords = STEPIFY(p_resource->states_ptr->subresource_states.size(), 64) / 64;
 			CRASH_COND(p_resource->states_ptr->subresource_states.size() > BarrierRequest::MAX_SUBRESOURCES);
 			br.planes = p_num_planes;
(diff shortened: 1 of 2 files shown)


drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -631,6 +633,33 @@ void RenderingDeviceDriverD3D12::_resource_transitions_flush(ID3D12GraphicsComma
...
+				D3D12_DISCARD_REGION dr = {};
+				dr.FirstSubresource = first_subres_to_init;
+				dr.NumSubresources = num_subres_to_init;
+				p_cmd_list->DiscardResource(br.dx_resource, &dr);
+				first_subres_to_init = p_subresource;
+				num_subres_to_init = 1;
+			}
+			res_states->subresource_states[p_subresource].init_state = ResourceInfo::States::ResourceStateEx::INIT_STATE_GARBAGE;
+		};
+		auto _flush_initialize_set = [&]() {
```
AThousandShips, line 654: This is only called once so I'd avoid a lambda

RandomShaper: My idea was to keep those very related pieces of functionality together so they can be studied more easily.

DarioSamo:
I ran into this exact issue so I should confirm later if it works.

DarioSamo:
The PR kind of works as expected and doesn't. As it is it did not fix the issue at all for me, but apparently it's because whenever this is called...

```cpp
D3D12_FEATURE_DATA_D3D12_OPTIONS12 options12 = {};
res = device->CheckFeatureSupport(D3D12_FEATURE_D3D12_OPTIONS12, &options12, sizeof(options12));
if (SUCCEEDED(res)) {
	format_capabilities.relaxed_casting_supported = options12.RelaxedFormatCastingSupported;
}
```

My system in particular (Windows 11 with the latest Windows SDK) will return that this feature is supported. However, as I mentioned to @RandomShaper before, it seems like a bit of a false positive as the view creation expecting relaxed format casting to work will fail and result in a crash afterwards. That poses a bit of a problem, as it seems either the driver or the SDK being used to build it is causing this weird issue (I tested with different SDKs and could not get the result to be different). It seems that whatever driver stack I'm stuck with will incorrectly report that the feature is supported when it's not.

Force-disabling the setting however works as expected and finally makes SDFGI work in D3D12, so I can confirm that the path does its job as long as the format capability is detected properly. I'm not sure the PR is at fault here as the format capability checking is a pretty standard implementation. But it seems likely we can't quite rely on the format casting flag to identify if it truly works or not in some particular scenarios. Perhaps we can rely on creating something and checking if it fails instead? Or address the IHVs who report this feature as true into checking what the problem is?

RandomShaper:
An option would be to disable the use of relaxed casting even if advertised as supported depending on OS and GPU vendor, but I'd rather do that in a separate PR or, in any case, consider it a separate issue that should indeed be interesting to report to Nvidia.

DarioSamo:
> An option would be to disable the use of relaxed casting even if advertised as supported depending on OS and GPU vendor, but I'd rather do that in a separate PR or, in any case, consider it a separate issue that should indeed be interesting to report to Nvidia.

It might be the safest option to indeed just not check for support until we have a solid confirmation that the feature is widely available and supported on more hardware.

RandomShaper:
Are you sure sure that a driver update won't do the trick for you?

DarioSamo:
> Are you sure sure that a driver update won't do the trick for you?

I don't have more driver updates to do. 😞

RandomShaper:
I've added temporary code to disable it. See the updated PR description. (Warning: the PR description asks to check your comment above. If one then reads the next comment, that is, this one, they will enter an infinite loop.)

drivers/d3d12/rendering_device_driver_d3d12.cpp
clayjohn: It looks like the call locations are over 100 lines of code away though. So I would also prefer elevating these to proper functions

RandomShaper: Those changes belong to the base PR included here (#87872). I've amended it with an approach that isn't either lambdas nor member functions. I believe the one one has all the advantages. I've rebased this PR on top of the amended one.

RandomShaper:
Rebased! (And retested.)

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -530,18 +530,21 @@ static const D3D12_RESOURCE_DIMENSION RD_TEXTURE_TYPE_TO_D3D12_RESOURCE_DIMENSIO
...
 #endif
 
 	ResourceInfo::States *res_states = p_resource->states_ptr;
-	ID3D12Resource *res_to_transition = p_resource_override ? p_resource_override : p_resource->resource;
 
 	if (p_new_state == D3D12_RESOURCE_STATE_COPY_DEST) {
 		res_states->subresource_states[p_subresource].init_state = ResourceInfo::States::ResourceStateEx::INIT_STATE_DONE;
+	} else if (p_new_state == D3D12_RESOURCE_STATE_UNORDERED_ACCESS) {
+		if (unlikely(!res_states->xfamily_fallback.subresources_dirty.is_empty())) {
+			res_states->xfamily_fallback.subresources_dirty[p_subresource / sizeof(uint64_t)] |= (uint64_t(1) << (p_subresource % sizeof(uint64_t)));
```
DarioSamo, line 546: It seems the intention of the code is to make a bitmask, but it's using the sizeof(uint64_t) which is in bytes. All of this seems to be effectively only using the first byte of each element of a vector of uint64_ts.

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -1208,12 +1207,13 @@ RDD::TextureID RenderingDeviceDriverD3D12::texture_create(const TextureFormat &p
...
 		// Two options here:
-		// 1. If ID3DDevice10 is present and driver reports relaxed casting is, leverage its new extended resource creation API (via D3D12MA).
-		// 2. Otherwise, fall back to an approach based on abusing aliasing, hoping for the best. [[CROSS_FAMILY_ALIASING]]
+		// 1. If the driver reports relaxed casting is, leverage its new extended resource creation API (via D3D12MA).
+		// 2. Otherwise, fall back to an approach based on having multiple versions of the resource and copying as needed. [[CROSS_FAMILY_FALLBACK]]
 		if (p_format.shareable_formats.size()) {
 			if (format_capabilities.relaxed_casting_supported) {
 				ComPtr<ID3D12Device10> device_10;
 				device->QueryInterface(device_10.GetAddressOf());
+				DEV_ASSERT(device_10);
```
DarioSamo, line 1216: The above comment was updated to reflect the device10 casting is no longer used for checking but an assert and an unused cast remain here. Perhaps we can relax this and just not check if device10 is present considering it doesn't use it anymore?

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -1499,75 +1439,59 @@ RDD::TextureID RenderingDeviceDriverD3D12::texture_create_from_extension(uint64_
...
-			if (E.key == RD_TO_D3D12_FORMAT[p_view.format].family) {
-				texture = E.value.Get();
-				break;
+		D3D12MA::ALLOCATION_DESC allocation_desc = {};
+		allocation_desc.HeapType = D3D12_HEAP_TYPE_DEFAULT;
+		allocation_desc.ExtraHeapFlags = D3D12_HEAP_FLAG_ALLOW_ONLY_NON_RT_DS_TEXTURES;
+
+		if (p_slice_type != -1) {
+#ifdef DEV_ENABLED
+			// Actual slicing is not contemplated. If ever needed, let'¡s at least realize.
```
DarioSamo, line 1470:
```suggestion
			// Actual slicing is not contemplated. If ever needed, let's at least realize.
```

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -1499,75 +1439,59 @@ RDD::TextureID RenderingDeviceDriverD3D12::texture_create_from_extension(uint64_
...
+		new_tex_resource_desc.Flags = D3D12_RESOURCE_FLAG_NONE; // Alternative formats can only be used as SRVs.
+
+		HRESULT res = allocator->CreateResource(
+				&allocation_desc,
+				&new_tex_resource_desc,
+				D3D12_RESOURCE_STATE_COPY_DEST,
+				nullptr,
+				new_allocation.GetAddressOf(),
+				IID_PPV_ARGS(new_texture.GetAddressOf()));
+		ERR_FAIL_COND_V_MSG(!SUCCEEDED(res), TextureID(), "D3D12MA::CreateResource failed with error " + vformat("0x%08ux", (uint64_t)res) + ".");
```
DarioSamo, line 1490:
```suggestion
		ERR_FAIL_COND_V_MSG(!SUCCEEDED(res), TextureID(), vformat("D3D12MA::CreateResource failed with error 0x%08ux.", (uint64_t)res));
```

drivers/d3d12/rendering_device_driver_d3d12.cpp
```diff
@@ -1583,102 +1507,162 @@ RDD::TextureID RenderingDeviceDriverD3D12::texture_create_shared_from_slice(Text
...
-	tex_info->states_ptr = owner_tex_info->states_ptr;
+	tex_info->resource = resource;
+	if (new_texture.Get()) {
+		// [[CROSS_FAMILY_FALLBACK]].
+
+		DEV_ASSERT(cross_family && !format_capabilities.relaxed_casting_supported);
+
+		uint32_t new_texture_subresorce_count = owner_tex_info->mipmaps * owner_tex_info->layers;
+#ifdef DEV_ENABLED
+		// Actual slicing is not contemplated. If ever needed, let'¡s at least realize.
```
DarioSamo, line 1604:
```suggestion
		// Actual slicing is not contemplated. If ever needed, let's at least realize.
```

DarioSamo (changes requested):
The fallback does work for me, there's just a couple of changes we might want to do like addressing the bitset that is using more bytes than it needs to. I haven't spotted anything on the implementation of the fallback itself.

drivers/d3d12/rendering_device_driver_d3d12.cpp
RandomShaper: Sounds good. I've sometimes even felt the temptation to use some bitmap class that some code in `thirdparty/` seems to provide... That said, I'll fix this by just using `64` in the expression and let's consider the bitset class an improvement for a future iteration.

RandomShaper: Good catch, by the way!

drivers/d3d12/rendering_device_driver_d3d12.cpp
RandomShaper: Makes sense!

drivers/d3d12/rendering_device_driver_d3d12.cpp
RandomShaper: The new approach doesn't do any of that anymore.

DarioSamo (approved):

akien-mga:
Can be rebased now that the first commit was merged.

RandomShaper:
> Can be rebased now that the first commit was merged.

Done. I was just now wondering why GitHub didn't just subsumed the superfluous commit. Didn't that use to be the case?

akien-mga:
> > Can be rebased now that the first commit was merged.
> 
> Done. I was just now wondering why GitHub didn't just subsume the superfluous commit. Didn't that use to be the case?

I don't think it does, but merging the PR as is might still work and not introduce an extra commit. It's just that the PR itself still looks like it includes two new commits not in `master`, which is a bit confusing so I prefer to rebase anyway.

akien-mga:
Thanks!

PR #88028, opened 2024-02-06
Silence wayland-scanner check in buildsystem

Calinou:
This prevents a wayland-scanner message from appearing every build when `wayland=yes` is used (the default). The error message when wayland-scanner is still printed as it's not printed by wayland-scanner itself.

wayland-scanner prints its version to stderr instead of stdout (at least in 1.22.0), so stderr needs to be redirected.

### DIFF
--- a/platform/linuxbsd/detect.py
+++ b/platform/linuxbsd/detect.py
@@ -207,8 +207,8 @@ def configure(env: "Environment"):
         env.Append(CPPDEFINES=["SOWRAP_ENABLED"])
 
     if env["wayland"]:
-        if os.system("wayland-scanner -v") != 0:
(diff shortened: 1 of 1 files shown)


adamscott:
Are you sure? Here's my `wayland-scanner -v` result.
image

Calinou:
> Are you sure? Here's my wayland-scanner -v result.

Yes, that's the correct output. You can print stderr and still have your app exit with code `0`. In fact, stderr is often used by apps like FFmpeg to report their progress separately from output messages.

deralmas (approved):
Yeah, this is a good idea, thanks!

akien-mga:
Thanks!

PR #88029, opened 2024-02-06
Display non finite numbers in stack trace debugger

### LINKED ISSUE #88006: Float values NAN or INF are shown as 0 in the debugger and inspector
*Bugsquad edit:* Regression from https://github.com/godotengine/godot/pull/81076

### Tested versions

- Reproducible in v4.2.1.stable.official [b09f793f5], v4.2.stable.official [46dc27791]
- NOT reproducible in v4.1.3.stable.official [f06b6836a] 

### System information

Godot v4.2.1.stable - Debian GNU/Linux trixie/sid trixie - Wayland - Vulkan (Forward+) - integrated Intel(R) Xe Graphics (TGL GT2) () - 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz (8 Threads)

### Issue description

Variables in the debugger window (stack variables) or properties in the inspector of type float will never show NAN or INF values correctly, they show up as 0
The tooltip in the script editor DOES show if a variable contains NAN or INF.

Rather nasty bug if you happen to be debugging a broken formula or something outputting an unwanted nan you are not aware of. The nan will profilerate but you will not see it at all. Nan has some special comparison properties also, which 0 does not, and this could really mess with your mind until you realize the debugger itself is bugged.

### Steps to reproduce

Run the mrp until the breakpoint hits.
Open the local variables in the debugger -> all values show as 0
Open the 'self' object in the inspector -> all property values show as 0

[mrp_no_nan_visible_in_debugger_and_inspector.zip](https://github.com/godotengine/godot/files/14175751/mrp_no_nan_visible_in_debugger_and_inspector.zip)

quirkylemon:
fixes #88006
This just makes it so setting the value with code doesn't check if a value is finite but it does still check when you set the value using the gui

### DIFF
--- a/editor/debugger/editor_debugger_inspector.cpp
+++ b/editor/debugger/editor_debugger_inspector.cpp
@@ -33,7 +33,6 @@
 #include "core/debugger/debugger_marshalls.h"
 #include "core/io/marshalls.h"
 #include "editor/editor_node.h"
-#include "editor/editor_properties.h"
 #include "scene/debugger/scene_debugger.h"
 
 bool EditorDebuggerRemoteObject::_set(const StringName &p_name, const Variant &p_value) {
@@ -258,6 +257,7 @@ void EditorDebuggerInspector::add_stack_variable(const Array &p_array) {
 	pinfo.type = v.get_type();
 	pinfo.hint = h;
 	pinfo.hint_string = hs;
+	pinfo.usage = PROPERTY_USAGE_DEBUG | PROPERTY_USAGE_DEFAULT;
 
 	variables->prop_list.push_back(pinfo);
 	variables->prop_values[type + n] = v;
@@ -285,14 +285,4 @@ String EditorDebuggerInspector::get_stack_variable(const String &p_var) {
 		}
 	}
 	return String();
-}
-
-void EditorDebuggerInspector::_update_ep(EditorProperty *ep) {
-	if (ep->get_edited_property_value().get_type() == Variant::FLOAT && ep->is_read_only()) {
-		double value = ep->get_edited_property_value();
-		if (!Math::is_finite(value)) {
-			Object::cast_to<EditorPropertyFloat>(ep)->set_allow_non_finite(true);
-		}
-	}
-	EditorInspector::_update_ep(ep);
 }
--- a/editor/debugger/editor_debugger_inspector.cpp
+++ b/editor/debugger/editor_debugger_inspector.cpp
@@ -33,6 +33,7 @@
 #include "core/debugger/debugger_marshalls.h"
 #include "core/io/marshalls.h"
 #include "editor/editor_node.h"
+#include "editor/editor_properties.h"
 #include "scene/debugger/scene_debugger.h"
 
 bool EditorDebuggerRemoteObject::_set(const StringName &p_name, const Variant &p_value) {
@@ -284,4 +285,14 @@ String EditorDebuggerInspector::get_stack_variable(const String &p_var) {
 		}
 	}
 	return String();
+}
+
+void EditorDebuggerInspector::_update_ep(EditorProperty *ep) {
+	if (ep->get_edited_property_value().get_type() == Variant::FLOAT && ep->is_read_only()) {
+		double value = ep->get_edited_property_value();
+		if (!Math::is_finite(value)) {
+			Object::cast_to<EditorPropertyFloat>(ep)->set_allow_non_finite(true);
+		}
+	}
+	EditorInspector::_update_ep(ep);
 }
--- a/scene/gui/range.cpp
+++ b/scene/gui/range.cpp
@@ -95,10 +95,14 @@ void Range::set_value(double p_val) {
 
 void Range::_set_value_no_signal(double p_val) {
 	if (!Math::is_finite(p_val)) {
-		if (allow_non_finite) {
-			shared->val = p_val;
+		if (shared->allow_non_finite) {
+			if (Math::is_nan(p_val)) {
+				shared->val = p_val;
+				return;
+			}
+		} else {
+			return;
 		}
-		return;
 	}
 
 	if (shared->step > 0) {
@@ -377,11 +381,11 @@ bool Range::is_lesser_allowed() const {
 }
 
 void Range::set_allow_non_finite(bool p_allow) {
-	allow_non_finite = p_allow;
+	shared->allow_non_finite = p_allow;
 }
 
 bool Range::is_non_finite_allowed() const {
-	return allow_non_finite;
+	return shared->allow_non_finite;
 }
 
 Range::Range() {
--- a/scene/gui/range.cpp
+++ b/scene/gui/range.cpp
@@ -93,7 +93,6 @@ void Range::set_value(double p_val) {
 	}
 }
 
-// This work exactly the same as before by default so it shouldn't break anything.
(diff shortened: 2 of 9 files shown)


quirkylemon:
sounds good.

AThousandShips:
I'd say this needs some further details at least, it might break things with using the input, like for `SpinBox`, as it still is returned from `get_value`, and you might do things like scrolling up and down

It also changes the behavior between the signal and non-signal case, I'd say to have more control over the printing and a special case for non-finite cases, because this changes some of how it works that can introduce issues and break compatibility

Some changes on the `Range` side is necessary, but I think more work is needed on the editor side

quirkylemon:
just to be clear when you say a more local fix you mean code specific to the debugger?

AThousandShips:
Yes, at least Investigate that side, unsure exactly how it would look, but would be important if it's possible without changing the gui element (which requires more testing to avoid it breaking things)

Mickeon:
What the heck happened here, this PR needs to be absolutely rebased.

quirkylemon:
Sorry I am absolutely horrendous at using git/github. 
it should be fixed now.

quirkylemon:
Ok so explanation for the changes made 
* The function `_update_ep` was made so `EditorDebuggerInspector` can override with it code that only applies to debugging. similar code shows up in other places in `EditorInspector` but I did not use the function there because it isn't necessary for `EditorDebuggerInspector` to work.
* `Range` can be set to non finite values now but you need to toggle that on so all code that already uses `Range` isn't affected. the toggle allowing non finite numbers is currently in public but it might be a good idea to move it to protected. 

also I learned what caused the goofy issue with git rebase
if you use `git rebase -i upstream/master` instead of `git rebase -i master` like ATthousandShips suggested. it puts in a ton of commits. 

also just to be safe if it says there outgoing changes after a rebase is it ok to push?

scene/gui/range.h
```diff
@@ -100,6 +100,8 @@ class Range : public Control {
 	void set_allow_lesser(bool p_allow);
 	bool is_lesser_allowed() const;
 
+	bool allow_non_finite = false;
```
AThousandShips, line 103: This shouldn't be a public member variable, instead it should have setters and getters

Mickeon:
Before requesting changes (as I would feel pretty bad wasting anyone's time), are we sure this is the way to go about fixing this bug?

AThousandShips:
I'm not sure the exact best way to approach this, this looks to be more in the right direction though

scene/gui/range.cpp
```diff
@@ -95,6 +95,9 @@ void Range::set_value(double p_val) {
 
 void Range::_set_value_no_signal(double p_val) {
 	if (!Math::is_finite(p_val)) {
+		if (allow_non_finite) {
+			shared->val = p_val;
+		}
 		return;
 	}
```
dalexeev, line 102: Just because non-finite is allowed doesn't mean you can set arbitrary values without checks. For example, `+INF` should be limited to `max_value` if `allow_greater` is false.

scene/gui/range.cpp
```diff
@@ -373,6 +376,14 @@ bool Range::is_lesser_allowed() const {
 	return shared->allow_lesser;
 }
 
+void Range::set_allow_non_finite(bool p_allow) {
+	allow_non_finite = p_allow;
+}
+
+bool Range::is_non_finite_allowed() const {
+	return allow_non_finite;
+}
```
dalexeev, line 385: Perhaps it makes sense to expose this as a property.

editor/debugger/editor_debugger_inspector.cpp
```diff
@@ -285,3 +286,13 @@ String EditorDebuggerInspector::get_stack_variable(const String &p_var) {
...
+
+void EditorDebuggerInspector::_update_ep(EditorProperty *ep) {
+	if (ep->get_edited_property_value().get_type() == Variant::FLOAT && ep->is_read_only()) {
+		double value = ep->get_edited_property_value();
+		if (!Math::is_finite(value)) {
+			Object::cast_to<EditorPropertyFloat>(ep)->set_allow_non_finite(true);
+		}
+	}
+	EditorInspector::_update_ep(ep);
+}
```
dalexeev, line 298: Is it possible to specify this parameter when initializing the control? In the case of the debugger, this should not change.

quirkylemon: probably, but I think we will still need to override a bit of `EditorInspector`

scene/gui/range.cpp
quirkylemon: do you know if there would be any uses for that?

Mickeon: Custom editor tools, much more than anything.

I am skeptical on exposing it myself, as I would rather have at least one individual ask for it, first, so as not to bloat the API. 
_Bbbbbbut_ there were users that did in fact rely on the previous behavior, as well (but only in an editor context, notably)

dalexeev: Ok, let's wait for an explicit request.

editor/debugger/editor_debugger_inspector.cpp
quirkylemon: ok ignore the last message. 
Is it ok to add to `PropertyUsageFlags`?
I found a much a better solution that doesn't require changing `EditorInspector` but it does require adding `PROPERTY_USAGE_DEBUG` to `PropertyUsageFlags`

scene/gui/range.cpp
quirkylemon: Is this ok? 
```
if (!Math::is_finite(p_val) && !allow_non_finite) {
	return;
}
```

dalexeev: I think yes, since there are currently no negative checks in the code below. All comparisons with NaN return false (except `!=`), rounding NaN returns NaN. But probably the following code would be more reliable in case we accidentally add a negative check in the future:

```c++
if (!Math::is_finite(p_val)) {
    if (allow_non_finite) {
        if (Math::is_nan(p_val)) {
            shared->val = p_val;
            return;
        }
    } else {
        return;
    }
}
```

editor/debugger/editor_debugger_inspector.cpp
dalexeev: I don't think adding a new usage flag is justified for this purpose. Most likely there are more local solutions.

scene/gui/range.h
```diff
@@ -53,6 +53,8 @@ class Range : public Control {
 
 	Shared *shared = nullptr;
 
+	bool allow_non_finite = false;
```
dalexeev, line 56: I'm guessing this should be added to the `Shared` structure to be consistent with `allow_greater` and `allow_lesser`.

editor/debugger/editor_debugger_inspector.cpp
quirkylemon: More local solutions do exist but adding a usage flag is a lot cleaner.
I pushed a commit so you can see how much cleaner it is.

quirkylemon: sorry the last message in hindsight is kinda rude. 
I should have asked what do you mean by a more local solution?

quirkylemon:
I am going to make a pr for the changes to Range and take a break or give up on this problem. 
https://github.com/godotengine/godot/pull/88354

PR #88045, opened 2024-02-07
Optimize when colliders are regenerated for imported meshes

Malcolmnixon:
This pull request fixes unnecessary regeneration of collision shapes in the import dialog. The old logic recalculated the colliders for every mesh in the model - including for meshes that did not have physics colliders enabled, and for meshes with existing colliders whose settings had not been modified.

This was not a problem for simple models; however @yythlj encountered this attempting to import the [shop_house.zip](https://github.com/godotengine/godot/files/14188673/shop_house.zip) model which consists of:
- 68513 vertices
- 56289 triangles
- 1176 discrete objects/meshes

Changing physics settings on any mesh resulted in calculating colliders for all 1176 meshes, which takes so long I gave up trying to measure it. This behavior is now modified to only calculate the collider for a mesh under the following circumstances:
1. The collider is being made visible (ensures previously-enabled colliders are calculated for preview in the import dialog).
2. The currently selected mesh collider settings have been modified triggering a regeneration with different collision settings.

For the shop_house model described above; if collision happened to be enabled for all 1176 meshes there would still be an unavoidable enormous delay on opening the import dialog so the collider meshes could be calculated for preview.

The following video demonstrates opening the import settings on the shop_house model with a few colliders enabled, then toggling the collider-generation on a few meshes and re-importing.

https://github.com/godotengine/godot/assets/1863707/1393f472-2f5c-4101-9540-d3a69ddee79d

### DIFF
--- a/editor/import/3d/scene_import_settings.cpp
+++ b/editor/import/3d/scene_import_settings.cpp
@@ -430,24 +430,27 @@ void SceneImportSettingsDialog::_update_view_gizmos() {
 		return;
 	}
 	for (const KeyValue<String, NodeData> &e : node_map) {
+		// Skip import nodes that aren't MeshInstance3D.
+		const MeshInstance3D *mesh_node = Object::cast_to<MeshInstance3D>(e.value.node);
+		if (mesh_node == nullptr || mesh_node->get_mesh().is_null()) {
+			continue;
+		}
+
+		// Determine if the mesh collider should be visible.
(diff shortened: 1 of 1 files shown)


fire (approved):
Looks good, we probably still want to profile and optimize colliders for all 1176 meshes since there is a way to automatically generate all colliders, but this is an improvement.

akien-mga:
Thanks!

PR #80410, opened 2023-08-08
[RTL] Add support for image dynamic updating, padding, tooltips and size in percent.

bruvzg:
- Adds method to update RTL images without changing the text.
- Adds support for small image padding.
- Adds support for image size in percent of control width.

RTL only part of https://github.com/godotengine/godot/pull/69751

PR #88047, opened 2024-02-07
Add a partial path return option for AStar

### LINKED ISSUE #277: Add an AStar function to get partial path if end point is unreachable
2d tile based game where character primarily walks through point and click.

If the player clicks on a tile that they cannot reach the current A* implementation will just return an empty path. I would like the player to walk as close as possible to wherever was clicked instead. There is the "get_closest_point" function but it doesnt help when the player clicks on a point that is is the AStar grid but is not currently reachable from the player's position.

**Describe how this feature / enhancement will help you overcome this problem or limitation:**
AStar will allow returning partial paths which would help the movement in my game tremendously. Right now I will probably have to replace with another AStar implementation that supports this.

image

Assume that the top-left green square is the player and the black squares are walls.
If I were to click on the bottom-right dark green square then currently AStar should just not return a path at all. Instead, I would like for it to return the path that leads the player to the purple square since that's the closest position that the player can reach.

I believe this is a matter of keeping track of the node with the lowest "estimated cost" during the `_solve` function of AStar and returning a path to that node if it doesn't find a path to the goal. This could be another function like `get_partial_point_path` or potentially a bool/enum passed into the current `get_point_path` function.

No, I would have to re-implement or use a completely different A* implementation.

This is just improved functionality on the current A* class.

alliephante:
This change should address and close godotengine/godot-proposals#277 by adding an optional parameter to `get_point_path` and `get_id_path` for `AStar2D`, `AStar3D`, and `AStarGrid2D` that when true will return the partial path instead of an empty path.

By having this be an optional parameter this should not have any compatibility breakages, and for those who rely on having an empty path returned if no complete path is available will not have their code affected. Those who want/need partial paths however can simply pass true in to get them.

**Note**: This will not return a partial path to a location not in the AStar point list, the destination point still has to be a point. This is especially helpful for games where points may become disconnected due to other points being disabled during play while still permitting a path that gets close to the destination point.

I also went ahead and updated the xml docs for the affected classes to mention this new possibility. If I understand correctly the Godot docs page uses these to auto-generated so a doc update for that shouldn't be necessarily.

The original implementation was modified from [KnightNine's branch located here](https://github.com/KnightNine/godot/tree/4.0-astar-experimental), however there were a bunch of other changes for other features in that branch and I wanted to only address the original proposal issue. They used a separate method as suggested in the proposal issue, but I felt that an optional parameter made more sense and to prevent having extra data structures to store the partial paths.

### Example
**Download**: [astar_partial_test.zip](https://github.com/godotengine/godot/files/14189050/astar_partial_test.zip)

I used the navigation2d demo project to test `AStarGrid2D`. I tested `AStar3D` using a personal project and I cannot share the code of that, but it behaved as expected. I have included the modified demo project for ease of testing. I did not have time to modify this demo project to explicitly showcase it working with AStar2/3D.

Screenshot_20240206_185730
This is the test using the attached demo project. The E tile was the destination tile. This tile is a part of the AStarGrid2D's points, but is surrounded by impassible/disabled points isolating it. The path gets as close as it can.

Screenshot_20240206_194638
This example shows the `AStar3D` test. As movement costs are not 1 for all, I added a few annotations. The infinity over the black tile with brown objects on it is a tile that is not in the map. The tiles with a 2 on them have a cost of 2. The blue tile with the red line was the tile selected. This tile is in the AStar point list, but is not connected to any points. The white arrows follow the path generated by AStar. As you can see it gets close to the target tile. as expected.

### 3.x
I am uncertain on compatibility for 3.x as I do not use it nor did I test this. My brief cursory glance shows multiple commits to the AStar classes since 3.x, so it is possible due to this it cannot be cherry-picked, but I really admit I am uncertain on how that process works.

### DIFF
--- a/core/math/a_star.compat.inc
+++ b/core/math/a_star.compat.inc
@@ -1,0 +1,59 @@
+/**************************************************************************/
+/*  a_star.compat.inc                                                     */
+/**************************************************************************/
+/*                         This file is part of:                          */
+/*                             GODOT ENGINE                               */
+/*                        https://godotengine.org                         */
+/**************************************************************************/
+/* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+/* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef DISABLE_DEPRECATED
+
+Vector<int64_t> AStar3D::_get_id_path_bind_compat_88047(int64_t p_from_id, int64_t p_to_id) {
+	return get_id_path(p_from_id, p_to_id, false);
+}
+
+Vector<Vector3> AStar3D::_get_point_path_bind_compat_88047(int64_t p_from_id, int64_t p_to_id) {
+	return get_point_path(p_from_id, p_to_id, false);
+}
+
+void AStar3D::_bind_compatibility_methods() {
+	ClassDB::bind_compatibility_method(D_METHOD("get_id_path", "from_id", "to_id"), &AStar3D::_get_id_path_bind_compat_88047);
+	ClassDB::bind_compatibility_method(D_METHOD("get_point_path", "from_id", "to_id"), &AStar3D::_get_point_path_bind_compat_88047);
+}
+
+Vector<int64_t> AStar2D::_get_id_path_bind_compat_88047(int64_t p_from_id, int64_t p_to_id) {
+	return get_id_path(p_from_id, p_to_id, false);
+}
+
+Vector<Vector2> AStar2D::_get_point_path_bind_compat_88047(int64_t p_from_id, int64_t p_to_id) {
+	return get_point_path(p_from_id, p_to_id, false);
+}
+
+void AStar2D::_bind_compatibility_methods() {
+	ClassDB::bind_compatibility_method(D_METHOD("get_id_path", "from_id", "to_id"), &AStar2D::_get_id_path_bind_compat_88047);
+	ClassDB::bind_compatibility_method(D_METHOD("get_point_path", "from_id", "to_id"), &AStar2D::_get_point_path_bind_compat_88047);
+}
+
+#endif // DISABLE_DEPRECATED
--- a/core/math/a_star.cpp
+++ b/core/math/a_star.cpp
@@ -29,6 +29,7 @@
 /**************************************************************************/
 
 #include "a_star.h"
+#include "a_star.compat.inc"
 
 #include "core/math/geometry_3d.h"
 #include "core/object/script_language.h"
@@ -319,6 +320,7 @@ Vector3 AStar3D::get_closest_position_in_segment(const Vector3 &p_point) const {
 }
 
 bool AStar3D::_solve(Point *begin_point, Point *end_point) {
+	last_closest_point = nullptr;
 	pass++;
 
 	if (!end_point->enabled) {
@@ -332,10 +334,17 @@ bool AStar3D::_solve(Point *begin_point, Point *end_point) {
 
 	begin_point->g_score = 0;
 	begin_point->f_score = _estimate_cost(begin_point->id, end_point->id);
+	begin_point->abs_g_score = 0;
+	begin_point->abs_f_score = _estimate_cost(begin_point->id, end_point->id);
 	open_list.push_back(begin_point);
 
 	while (!open_list.is_empty()) {
 		Point *p = open_list[0]; // The currently processed point.
+
+		// Find point closer to end_point, or same distance to end_point but closer to begin_point.
+		if (last_closest_point == nullptr || last_closest_point->abs_f_score > p->abs_f_score || (last_closest_point->abs_f_score >= p->abs_f_score && last_closest_point->abs_g_score > p->abs_g_score)) {
+			last_closest_point = p;
+		}
 
 		if (p == end_point) {
 			found_route = true;
@@ -368,6 +377,8 @@ bool AStar3D::_solve(Point *begin_point, Point *end_point) {
 			e->prev_point = p;
 			e->g_score = tentative_g_score;
 			e->f_score = e->g_score + _estimate_cost(e->id, end_point->id);
+			e->abs_g_score = tentative_g_score;
+			e->abs_f_score = e->f_score - e->g_score;
 
 			if (new_point) { // The position of the new points is already known.
 				sorter.push_heap(0, open_list.size() - 1, 0, e, open_list.ptr());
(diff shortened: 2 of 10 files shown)


theashtronaut:
When running the CI checks in my fork's branch, it failed the GDExtension compatibility
```
Error: Validate extension JSON: Error: Field 'classes/AStar2D/methods/get_id_path/arguments': size changed value in new API, from 2 to 3.
Validate extension JSON: Error: Field 'classes/AStar2D/methods/get_point_path/arguments': size changed value in new API, from 2 to 3.
Validate extension JSON: Error: Field 'classes/AStar3D/methods/get_id_path/arguments': size changed value in new API, from 2 to 3.
Validate extension JSON: Error: Field 'classes/AStar3D/methods/get_point_path/arguments': size changed value in new API, from 2 to 3.
Validate extension JSON: Error: Field 'classes/AStarGrid2D/methods/get_id_path/arguments': size changed value in new API, from 2 to 3.
Validate extension JSON: Error: Field 'classes/AStarGrid2D/methods/get_point_path/arguments': size changed value in new API, from 2 to 3.
Validate extension JSON: Error: Hash changed for 'classes/AStar2D/methods/get_id_path', from CAEE4B7E to BAEE9BE0. This means that the function has changed and no compatibility function was provided.
Validate extension JSON: Error: Hash changed for 'classes/AStar2D/methods/get_point_path', from 10C941DF to CC4B5A58. This means that the function has changed and no compatibility function was provided.
Validate extension JSON: Error: Hash changed for 'classes/AStar3D/methods/get_id_path', from CAEE4B7E to BAEE9BE0. This means that the function has changed and no compatibility function was provided.
Validate extension JSON: Error: Hash changed for 'classes/AStar3D/methods/get_point_path', from 34803E1E to 5D2437D3. This means that the function has changed and no compatibility function was provided.
Validate extension JSON: Error: Hash changed for 'classes/AStarGrid2D/methods/get_id_path', from 7693B298 to 72546031. This means that the function has changed and no compatibility function was provided.
Validate extension JSON: Error: Hash changed for 'classes/AStarGrid2D/methods/get_point_path', from 292643AB to 61DDCC3D. This means that the function has changed and no compatibility function was provided.
```

I am not sure how to fix this, I thought GDExtension would use the modified ClassDB bindings, but I don't really know how GDExtension works. Advice on how to resolve that would be greatly appreciated!

dalexeev:
> I am not sure how to fix this, I thought GDExtension would use the modified ClassDB bindings, but I don't really know how GDExtension works. Advice on how to resolve that would be greatly appreciated!

See #80410 for example. You need to add `.compat.inc` file, include it in `.cpp`, add some code to `.h`, and add `Validate extension JSON` to `4.2-stable.expected` file.

doc/classes/AStar2D.xml
```diff
@@ -139,8 +139,10 @@
 			<return type="PackedInt64Array" />
 			<param index="0" name="from_id" type="int" />
 			<param index="1" name="to_id" type="int" />
+			<param index="2" name="return_partial_path" type="bool" default="false" />
 			<description>
 				Returns an array with the IDs of the points that form the path found by AStar2D between the given points. The array is ordered from the starting point to the ending point of the path.
+				Will return a partial path to a point that is disconnected if [param return_partial_path] is true, otherwise it will return an empty array. This value is false by default.
```
dalexeev, line 145: Perhaps this parameter should be called `allow_partial_path`? I also think the description could be rephrased to be more clear. CC @AThousandShips

core/math/a_star.h
```diff
@@ -149,8 +154,8 @@ class AStar3D : public RefCounted {
 	int64_t get_closest_point(const Vector3 &p_point, bool p_include_disabled = false) const;
 	Vector3 get_closest_position_in_segment(const Vector3 &p_point) const;
 
-	Vector<Vector3> get_point_path(int64_t p_from_id, int64_t p_to_id);
-	Vector<int64_t> get_id_path(int64_t p_from_id, int64_t p_to_id);
+	Vector<Vector3> get_point_path(int64_t p_from_id, int64_t p_to_id, bool return_partial_path = false);
+	Vector<int64_t> get_id_path(int64_t p_from_id, int64_t p_to_id, bool return_partial_path = false);
```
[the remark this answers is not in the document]
theashtronaut: Oh yes d'oh on my part, fixed (and also renamed as previously suggested)

core/math/a_star.compat.inc
```diff
@@ -0,0 +1,59 @@
...
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef DISABLE_DEPRECATED
+
+Vector<int64_t> AStar3D::_get_id_path_bind_compat_88047(int64_t p_from_id, int64_t p_to_id) {
+  return get_id_path(p_from_id, p_to_id);
```
AThousandShips, line 34:
```suggestion
	return get_id_path(p_from_id, p_to_id, false);
```
Add the default value explicitly (and formatting)

Same for all these

theashtronaut:
Many thanks to everyone for the suggestions regarding naming/style! I've made those changes.



> > I am not sure how to fix this, I thought GDExtension would use the modified ClassDB bindings, but I don't really know how GDExtension works. Advice on how to resolve that would be greatly appreciated!
> 
> See #80410 for example. You need to add `.compat.inc` file, include it in `.cpp`, add some code to `.h`, and add `Validate extension JSON: Error: Field...` lines to `misc/extension_api_validation/4.2-stable.expected` file. `Compat.cs` no longer needs to add compatibility methods.

Thank you! I think I've done this correctly. The validation script no longer returned any errors related to these classes when I ran it locally at least.



> Premature approval for the feature as this has been requested many times.
> 
> Can't say much about the technical detail or style as the AStar classes are not really my habitat.
> 
> I would even go as far and do a small compatibility break here on this and make it the default to always return the partial path if the target can not be reached. That is what the majority of (new) users expect and is also how the NavigationServer path queries all behave. That the path returned empty when the target could not be reached always irritated and annoyed people.

If others agree I can set the default to true! I just defaulted to the most compatible option to start with.

core/math/a_star.compat.inc
AThousandShips: You should set up your editor too to use tabs :)

misc/extension_api_validation/4.2-stable.expected
```diff
@@ -73,3 +73,25 @@ GH-87668
 Validate extension JSON: Error: Field 'classes/Font/methods/find_variation/arguments': size changed value in new API, from 8 to 9.
 
 Added optional "baseline_offset" argument. Compatibility method registered.
+
+
+GH-81996
+--------
```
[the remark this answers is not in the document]
theashtronaut: Okay I think I fixed it!

Bromeon:
> I would even go as far and do a small compatibility break here on this and make it the default to always return the partial path if the target can not be reached. That is what the majority of (new) users expect and is also how the NavigationServer path queries all behave. That the path returned empty when the target could not be reached always irritated and annoyed people.

"Small compatibility break" is an understatement. This will break success checks for **all** existing `AStar{2D|3D}::get_{id|point}_path()` calls and thus affect the majority of pathfinding code using those APIs.

Checking for emptiness was the recommended approach in Godot 4.0, 4.1 and 4.2. I don't think changing this to accomodate for new users is a good idea. Besides, we don't even know how many people would be irritated/annoyed if a simple emptiness check would now become a "either empty or last path is destination" check.

misc/extension_api_validation/4.2-stable.expected
```diff
@@ -84,6 +84,13 @@ Validate extension JSON: Error: Field 'classes/Sky/properties/sky_material': typ
 Property hints reordered to improve editor usability. The types allowed are still the same as before. No adjustments should be necessary.
 
 
+GH-68420
+--------
+Validate extension JSON: Error: Field 'classes/Node/methods/_get_configuration_warnings/return_value': type changed value in new API, from "PackedStringArray" to "Array".
+
+Allow configuration warnings to refer to a property. Compatibility method registered.
+
+
```
Bromeon, line 93: This is unrelated to your PR and was reverted in the meantime, please don't re-add it.

theashtronaut: Good catch, I think at some point it picked it up with an upstream rebase. Should be removed now!

theashtronaut:
Just because I'm a little unclear on Godot dev cycles, is 4.3 "feature frozen" and this wouldn't be merged until 4.4 begins (and thus 4.3 is released)?

Mainly checking to know if I should keep clearing the conflicts with the 4.2-stable.expected file every few days to keep it ready to merge, or if the expectation that it's gonna wait until 4.4 then I can leave it for a while.

akien-mga:
There's still some time for 4.3 features, we're not in feature freeze yet.

But you don't need to rebase frequently to solve the conflicts on that compat file, first we need some maintainers familiar with that feature to approve the changes. Then you can do a final rebase to make it merge ready.

smix8 (approved):
For the navigation feature. Users ask regularly about a partial path option with the astar classes to be added. Tested and at least from my view it works as expected.

Not sure about code or compatibility detail. Others need to check that part.

AThousandShips (approved):

core/math/a_star.cpp
```diff
@@ -434,7 +445,12 @@ Vector<Vector3> AStar3D::get_point_path(int64_t p_from_id, int64_t p_to_id) {
 
 	bool found_route = _solve(begin_point, end_point);
 	if (!found_route) {
-		return Vector<Vector3>();
+		if (!p_allow_partial_path || last_closest_point == nullptr) {
+			return Vector<Vector3>();
+		}
+
+		// Use closest point instead
```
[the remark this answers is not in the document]
theashtronaut: Ah missed those periods. I really should figure out how to get the pre-hook style things to work

akien-mga:

PR #87612, opened 2024-01-26
ThorVG: update from v0.12.1 to v0.12.3

capnm:
https://github.com/thorvg/thorvg/releases/tag/v0.12.3

+ Full Changelog: https://github.com/thorvg/thorvg/compare/v0.12.1...v0.12.3

Godot-related SVG bug fixes:

+ svg_loader: Add missing transform functions skewX and skewY. 
    thorvg/thorvg#1928
+ sw_engine: Rectified dash line drawing issue. 
    thorvg/thorvg#1932

editor_screenshot_2024-01-26T120048

PR #88053, opened 2024-02-07
thorvg: Re-enable thread support, made opt-in upstream

### LINKED ISSUE #87788: Importing SVG's in 4.3.dev with commit 73589f6 is crashing
### Tested versions

- Reverting https://github.com/godotengine/godot/pull/87612 resolves the crashing
- Crash is not always present on windows but sometimes is
- Linux could repro the crash each time booting the engine provided you nuke the .godot/ folder

### System information

Linux - Fedora 38 - As up to date as possible

### Issue description

When importing SVG files in multiple threads the engine crashes with a realloc() crash, I did try to workaround this but had the same crash
image

### Steps to reproduce

- Import a pool of SVG's (doesn't seem to matter the type)

[TESTPROJECTREPRO.zip](https://github.com/godotengine/godot/files/14113916/TESTPROJECTREPRO.zip)

Images taken from another source, our project files are completely unrelated and they also crash.

### LINKED ISSUE #87962: ThorVG regression causing potentially corrupted SVGs
### Tested versions

- Master (b4e2a24c1f62088b3f7ce0197afc90832fc25009)

### System information

Arch Linux

### Issue description

When rendering SVGs using `ImageLoaderSVG::create_image_from_utf8_buffer` (I'm doing it from C++, but I can't see why it wouldn't affect godot proper too) the rendered images sometimes come out corrupted, an example:

cock_5

I believe it may be a ThorVG bug, since if I downgrade to thorvg 0.12.1 it works fine, but the issue is introduced in 0.12.2

### Steps to reproduce

- Delete .godot folder
- open the project
- svgs may be corrupt, or not, since it's probably some kind of data race it might be unreliable, but at least one of the svgs comes out corrupted for me

[mrp.zip](https://github.com/godotengine/godot/files/14159009/mrp.zip)

akien-mga:
- Fixup to #87612.
- Fixes #87788.
- Fixes #87962.

### DIFF
--- a/thirdparty/thorvg/src/common/tvgLock.h
+++ b/thirdparty/thorvg/src/common/tvgLock.h
@@ -38,10 +38,10 @@ namespace tvg {
     {
         Key* key = nullptr;
 
-        ScopedLock(Key& key)
+        ScopedLock(Key& p_key)
         {
-            key.mtx.lock();
-            this->key = &key;
+            p_key.mtx.lock();
+            key = &p_key;
         }
 
         ~ScopedLock()
--- a/thirdparty/thorvg/inc/config.h
+++ b/thirdparty/thorvg/inc/config.h
@@ -5,6 +5,7 @@
 #define THORVG_SVG_LOADER_SUPPORT
 #define THORVG_PNG_LOADER_SUPPORT
 #define THORVG_JPG_LOADER_SUPPORT
+#define THORVG_THREAD_SUPPORT
 
 // For internal debugging:
 //#define THORVG_LOG_ENABLED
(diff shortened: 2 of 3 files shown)


thirdparty/thorvg/src/common/tvgLock.h
```diff
@@ -38,10 +38,10 @@ namespace tvg {
     {
         Key* key = nullptr;
 
-        ScopedLock(Key& key)
+        ScopedLock(Key& p_key)
```
akien-mga, line 41: @hermet FYI, I was getting a `-Wshadow` warning from GCC here.
Happy to make a PR upstream if this kind of change looks acceptable, possibly with a different naming convention if there's one you favor.

akien-mga:
I'll yolo merge as I'm about to build 4.3-dev3 and would like this regression fixed in that build.

capnm:
I forgot that this get only default within the meson, sorry.

akien-mga:
Cherry-picked for 4.2.2.

PR #88057, opened 2024-02-07
Fix LineEdit behavior for deleting all the way to the left/right

### LINKED ISSUE #70793: cmd+backspace does not work in editor text inputs
### Godot version

v4.0.beta10.official [d0398f62f]

### System information

macOS 13.1 (22C65)

### Issue description

`cmd+backspace` does not actually backspace all to the left as it does elsewhere in macOS and other applications. `option+backspace` works and backspaces a word as intended.

### Steps to reproduce

backspace all to the left:

- open `Editor Settings`
- type `hello` into the `Filter Settings` input field
- hit `cmd+backspace` - nothing happens

another strange sequence:

- open `Editor Settings`
- type `hello` into the `Filter Settings` input field
- hit `ctrl+b` to move the cursor left one place (`hell|o`)
- hit `cmd+backspace` - nothing happens
- hit `backspace` - the last two characters are removed (`hel`)

should be reproducible with any Godot project on macOS

passivestar:
"Backspace all to left" and "Delete all to right" were swapped in `LineEdit`
Now they behave correctly like in `TextEdit`
Closes #70793

### DIFF
--- a/scene/gui/line_edit.cpp
+++ b/scene/gui/line_edit.cpp
@@ -140,7 +140,9 @@ void LineEdit::_backspace(bool p_word, bool p_all_to_left) {
 
 	if (p_all_to_left) {
 		deselect();
-		text = text.substr(0, caret_column);
+		text = text.substr(caret_column);
+		_shape();
+		set_caret_column(0);
 		_text_changed();
 		return;
 	}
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #88058, opened 2024-02-07
Tweak error and warning colors and fix `StatusWarning` icon visibility on light themes

### LINKED ISSUE #87987: Yellow atlas source warning exclamation mark is REALLY hard to see in light mode
### Tested versions

- Reproducible in v4.2.1.stable.official [b09f793f5], v4.3.dev2.official [352434668]

### System information

Windows 10.0.22631 - Vulkan (Mobile) - dedicated NVIDIA GeForce RTX 4080 Laptop GPU (NVIDIA; 31.0.15.4633) - 13th Gen Intel(R) Core(TM) i9-13900HX (32 Threads)

### Issue description

When there is an issue with the tiles of an atlas source, a yellow exclamation mark will warn the user about it. When hovering, it might show useful information, like "The current atlas source has tiles outside the texture" or something like that. Unfortunately, it is really hard to see in light mode:
image
While you certainly notice that it's there if you know about it, when I wondered why I'm getting so many errors I tried debugging for at least an hour before I learned of its existence online.

### Steps to reproduce

1. Set your Editor to light mode
2. Create a tile texture, a tileSet and an atlas source loading the texture.
3. Add tiles to the atlas source.
4. Edit the texture and remove some tiles.

[Exclamation.zip](https://github.com/godotengine/godot/files/14170756/Exclamation.zip)

rsubtil:
Fixes #87987

Originally this PR focused on fixing visibility of the `StatusWarning`, but after successive iterations and positive feedback, a few more relevant alterations were made:

- Removed `StatusWarning` and `StatusError` icons from the exclusion list to be able to convert its colors
- Set `error_color` to the same color used in `Node3D` (`#BF5A50` -> `#CD3838`)
- Shift `warning_color` to a more orange saturated hue (`#A69042` -> `#D18F19`)
- Tweak `EditorLogFilterButton` appearance (more comparison screenshots at https://github.com/godotengine/godot/pull/88058#issuecomment-1935833345):
    - Increased icon `normal` color opacity from 40% to 80%
    - Removed icon `hover` color modulate, making it now the original color, *aka* 100%
    - Lightened gray brackground on `pressed` state by 50%

---

<details><summary><b>Comparison screenshots</b></summary>

| Before | After |
| - | - |
|1_old |1_new |
|2_old |2_new |
|3_old |3_new |
|4_old |4_new |
|5_old |5_new |
|6_old |6_new |
|7_old |7_new |

</details>

### DIFF
--- a/editor/themes/editor_theme_manager.cpp
+++ b/editor/themes/editor_theme_manager.cpp
@@ -436,8 +436,8 @@ void EditorThemeManager::_create_shared_styles(const Ref<EditorTheme> &p_theme, 
 		if (!p_config.dark_theme) {
 			// Darken some colors to be readable on a light background.
 			p_config.success_color = p_config.success_color.lerp(p_config.mono_color, 0.35);
-			p_config.warning_color = p_config.warning_color.lerp(p_config.mono_color, 0.35);
-			p_config.error_color = p_config.error_color.lerp(p_config.mono_color, 0.25);
+			p_config.warning_color = Color(0.82, 0.56, 0.1);
+			p_config.error_color = Color(0.8, 0.22, 0.22);
 		}
 
 		p_theme->set_color("mono_color", EditorStringName(Editor), p_config.mono_color);
@@ -1901,14 +1901,20 @@ void EditorThemeManager::_populate_editor_styles(const Ref<EditorTheme> &p_theme
 			// When pressed, don't tint the icons with the accent color, just leave them normal.
 			p_theme->set_color("icon_pressed_color", "EditorLogFilterButton", p_config.icon_normal_color);
 			// When unpressed, dim the icons.
-			p_theme->set_color("icon_normal_color", "EditorLogFilterButton", p_config.icon_disabled_color);
+			Color icon_normal_color = Color(p_config.icon_normal_color, (p_config.dark_theme ? 0.4 : 0.8));
+			p_theme->set_color("icon_normal_color", "EditorLogFilterButton", icon_normal_color);
+			Color icon_hover_color = p_config.icon_normal_color * (p_config.dark_theme ? 1.15 : 1.0);
+			icon_hover_color.a = 1.0;
+			p_theme->set_color("icon_hover_color", "EditorLogFilterButton", icon_hover_color);
 
 			// When pressed, add a small bottom border to the buttons to better show their active state,
 			// similar to active tabs.
 			Ref<StyleBoxFlat> editor_log_button_pressed = style_flat_button_pressed->duplicate();
 			editor_log_button_pressed->set_border_width(SIDE_BOTTOM, 2 * EDSCALE);
 			editor_log_button_pressed->set_border_color(p_config.accent_color);
-
+			if (!p_config.dark_theme) {
+				editor_log_button_pressed->set_bg_color(flat_pressed_color.lightened(0.5));
+			}
 			p_theme->set_stylebox("normal", "EditorLogFilterButton", style_flat_button);
 			p_theme->set_stylebox("hover", "EditorLogFilterButton", style_flat_button_hover);
 			p_theme->set_stylebox("pressed", "EditorLogFilterButton", editor_log_button_pressed);
--- a/editor/themes/editor_color_map.cpp
+++ b/editor/themes/editor_color_map.cpp
@@ -169,9 +169,7 @@ void EditorColorMap::create() {
 	add_conversion_exception("ZoomReset");
 	add_conversion_exception("LockViewport");
 	add_conversion_exception("GroupViewport");
-	add_conversion_exception("StatusError");
 	add_conversion_exception("StatusSuccess");
-	add_conversion_exception("StatusWarning");
 	add_conversion_exception("OverbrightIndicator");
 	add_conversion_exception("MaterialPreviewCube");
 	add_conversion_exception("MaterialPreviewSphere");


adamscott (approved):
As simple as that! Thanks @rsubtil !

akien-mga:
On the second "After" screenshot, the error color also seems to lack contrast when the button is not toggled.

The toggle color itself seems to create contrast issues IMO. So I think we might need a bit more work than just special casing StatusWarning.

Calinou:
> The toggle color itself seems to create contrast issues IMO. So I think we might need a bit more work than just special casing StatusWarning.

The icon's opacity is decreased when the button isn't pressed – we probably need to lessen this effect, especially on light theme.

akien-mga:
> (I've also included the hover state as I noticed it looked bad for the yellow icon)

Yeah I was about to point out that this is still lacking contrast. Overall the colors for both StatusError and StatusWarning on the light theme seem too washed out / desaturated IMO.

And the pressed button color (blue-grayish) also seems too dark / dull IMO, at least from close up, maybe in context it fills more appropriate.

rsubtil:
> Yeah I was about to point out that this is still lacking contrast. Overall the colors for both StatusError and StatusWarning on the light theme seem too washed out / desaturated IMO.

- For `StatusError` we could make a color conversion for the same color being used for `Node3D`, which is a more saturated red.

- For `StatusWarning`, the color is already being converted. Increasing its saturation doesn't work, as yellow will always be problematic on white, and it affects other elements adversely. But, I liked the effect when I shifted to a more "orange" hue.

> And the pressed button color (blue-grayish) also seems too dark / dull IMO, at least from close up, maybe in context it feels more appropriate.

The effect is indeed much less pronounced on the dark theme, likely because of the already dark blue tint overall:
| Dark | Light |
| - | - |
| image | image |

I think reducing greatly the grayish tint works well for light themes, as the blue border is already a clear indicator of the button being toggled.

---

All in all, the changes are now as follows:

- Removed `StatusError` from the exclusion list to convert its colors
- Set `error_color` to the same color used in `Node3D` (`#BF5A50` -> `#CD3838`)
- Set `warning_color` to a more saturated color, and with a more orange tint (`#A69042` -> `#D18F19`)
- (On the output filter buttons only)
    - Increased icon `normal` color opacity from 40% to 80%
    - Removed icon `hover` color modulate (now original color, aka 100%)
    - Lightened gray brackground on `pressed` state by 50%


| | Normal | Hover | Pressed |
|-|-|-|-|
|Before|image |image |image |
|[1st iteration](https://github.com/godotengine/godot/pull/88058#issuecomment-1934070071)| image | image | image |
|2nd iteration| image|image |image|

>[!WARNING]
>This changes some core colors used throughout all the editor. The `error_color` is subtle, but the `warning_color` is more noticeable:
>
>| Before | After |
>|-|-|
>|image|image|

Mickeon:
I dig the orange hue for the warning color in the light theme. It's more contrasted.
I am not sure about a few other changes that _also_ affect the default theme, unless extremely small, as they may affect more places that one may think at a glance. Needs some testing.

Calinou:
The new warning color is great! I've always found the previous warning color less than ideal on a light theme, so this is a big improvement.

editor/themes/editor_theme_manager.cpp
```diff
@@ -396,8 +396,8 @@ void EditorThemeManager::_create_shared_styles(const Ref<EditorTheme> &p_theme,
 		if (!p_config.dark_theme) {
 			// Darken some colors to be readable on a light background.
 			p_config.success_color = p_config.success_color.lerp(p_config.mono_color, 0.35);
-			p_config.warning_color = p_config.warning_color.lerp(p_config.mono_color, 0.35);
-			p_config.error_color = p_config.error_color.lerp(p_config.mono_color, 0.25);
+			p_config.warning_color = Color("#d18f19");
+			p_config.error_color = Color("#cd3838");
```
Calinou, line 400: I suggest using the `Color(float, float, float)` form which avoids storing a string in the binary, and therefore reduces binary size ever so slightly. I did this on all colors in the codebase a while ago and it reduced binary size by a non-negligible amount.

`Color::hex(0xrrggbbaa)` might reduce binary size further by using 1 integer instead of 3 floats, but that should be evaluated later.

KoBeWi: `Color::hex(0xrrggbbaa)` can be shortened to `Color(0xrrggbbaa)`, no? It's a direct replacement for hex string.

Mickeon: My assumption was that the compiler was smarter than this but I suppose that these are among the many things to keep in mind. I did not expect for the Colors to be basically generated on the spot even for constant Strings.

rsubtil: I went with the 3-float approach then, but I agree reducing it in the future to a single hex int sounds interesting, especially for iterating on the colors and not having to constantly convert them to a [0, 1] float range.

Calinou: > By the way, I may be wrong but I believe the bits are reversed in pure hex: 0xaabbggrr.

`ARGB` was used in Godot 3.x, but Godot 4.0 switched to `RGBA`.

rsubtil:
Since the new changes ended up affecting the overall interface significantly, and there was very positive feedback on the warning color changes, I've edited this PR description to better describe the changes done, along with some new comparison screenshots.

akien-mga (approved):
Looks good to me.

akien-mga:
Thanks!

PR #88065, opened 2024-02-07
Handle display driver preferences consistently between editor and projects

deralmas:
The way that it was implemented before allowed for the project property to spill over to the editor if the `prefer_wayland` property is not present which, while not completely likely, is still a possibility. This specific part is thus more of a correctness change, ensuring that we always enforce the default display server when not running a project.

This commit also removes the restart requirement for the relative display driver project settings as they're independent and a restart would do nothing useful.


akien-mga (approved):

akien-mga:
Thanks!

mihe:
@akien-mga This resulted in a regression it seems: #94755

PR #88071, opened 2024-02-07
Fix NodePath autocompletion to ensure paths are quoted when required

### LINKED ISSUE #88059: Autocomplete suggestion doesn't suggest quotes for nested nodes with names that start with numeric value
### Tested versions

Reproducible in v3.5.3 and v4.2.beta5.official

### System information

Windows 10

### Issue description

Autocomplete suggestion doesn't suggest to add quotes for nested nodes, which start with numerical value. Going through with the incomplete suggestion will lead to an error.

Autocomplete knows how to suggest quotes for unnested nodes, that start with numerical value, but it would be nice to extend this functionality to work with nested nodes as well.

Example of unnested nodes working as intended:
kuva

### Steps to reproduce

For example having structure like suggested on the image will lead to an error:
kuva

Error: "Parse error: Invalid numeric constant: 0x"

[BugTest.zip](https://github.com/godotengine/godot/files/14194916/BugTest.zip)

Calinou:
Every component of the path is now checked to be a valid identifier, so that node names that start with a digit always require the full path to be quoted.

- This closes https://github.com/godotengine/godot/issues/88059.

## Preview

image

### DIFF
--- a/modules/gdscript/gdscript_editor.cpp
+++ b/modules/gdscript/gdscript_editor.cpp
@@ -3303,9 +3303,17 @@ static void _find_call_arguments(GDScriptParser::CompletionContext &p_context, c
 						opt = opt.substr(1);
 					}
 
-					// The path needs quotes if it's not a valid identifier (with an exception
-					// for "/" as path separator, which also doesn't require quotes).
-					if (!opt.replace("/", "_").is_valid_identifier()) {
+					// The path needs quotes if at least one of its components (excluding `/` separations)
+					// is not a valid identifier.
+					bool path_needs_quote = false;
+					for (const String &part : opt.split("/")) {
+						if (!part.is_valid_identifier()) {
+							path_needs_quote = true;
+							break;
+						}
+					}
+
+					if (path_needs_quote) {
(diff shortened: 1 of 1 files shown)


modules/gdscript/gdscript_editor.cpp
```diff
@@ -3303,9 +3303,17 @@ ::Error GDScriptLanguage::complete_code(const String &p_code, const String &p_pa
 						opt = opt.substr(1);
 					}
 
-					// The path needs quotes if it's not a valid identifier (with an exception
-					// for "/" as path separator, which also doesn't require quotes).
-					if (!opt.replace("/", "_").is_valid_identifier()) {
+					// The path needs quotes if at least one of its components (excluding `/` separations)
+					// is not a valid identifier.
+					bool path_needs_quote = false;
+					for (const String part : opt.split("/")) {
```
AThousandShips, line 3309:
```suggestion
					for (const String &part : opt.split("/")) {
```

AThousandShips (approved):
LGTM and solves the problem

akien-mga:
Thanks!

PR #62913, opened 2022-07-11
fix switching RAW mode to RGB mode issue

KuaiDain:
fixes #62894

```
return next_power_of_2(MAX(255, color.components[idx] * 255.0)) - 1;
```
when result of ```color.components[idx] * 255.0 ```  beyond 255, this issue will happen, so i modified this function as other mode

PR #88072, opened 2024-02-07
Fix `ColorPicker`'s RAW mode colors and values

### LINKED ISSUE #62894: Switching to RAW mode and then to RGB will change slider max value until refreshed
### Godot version

4.0 d26442e / 3.5

### System information

Windows 10 x64

### Issue description

When you change to RAW color mode, you can set a slider to some very big value. When you then switch to RGB, that slider will have a different max value. That value will stay until you change color mode again. I think it should switch back to 0-255 range as soon as you change the value to something within that range.
Z5sEZ3qfr0
(the GIF is from older build)

### Steps to reproduce

1. Have some ColorPicker
2. Change to RAW color mode
3. Set e.g. R to 100
4. Switch back to RGB
5. Set R to 0
6. You can still set R beyond 255

_No response_

WhalesState:
* Fixes: https://github.com/godotengine/godot/issues/62894

- Makes RAW sliders colored same as RGB mode.
- Fix RAW mode maximum values, from 100 to 1.0.
- Allow RGB and RAW modes to have greater values for overbright.
- Changes RAW step value to 1.0 / 255.0, to match the RGB mode values when changed.
- Prevent OKHSL mode from overriding the picker shape.
- Fix conversion between RAW and RGB modes when color is overbright.

### DIFF
--- a/scene/gui/color_mode.h
+++ b/scene/gui/color_mode.h
@@ -46,6 +46,8 @@ public:
 	virtual float get_spinbox_arrow_step() const { return get_slider_step(); }
 	virtual String get_slider_label(int idx) const = 0;
 	virtual float get_slider_max(int idx) const = 0;
+	virtual float get_spinbox_max(int idx) const = 0;
+	virtual bool can_allow_greater() const { return false; }
 	virtual float get_slider_value(int idx) const = 0;
 
 	virtual Color get_color() const = 0;
@@ -54,7 +56,6 @@ public:
 
 	virtual void slider_draw(int p_which) = 0;
 	virtual bool apply_theme() const { return false; }
-	virtual ColorPicker::PickerShapeType get_shape_override() const { return ColorPicker::SHAPE_MAX; }
 
 	ColorMode(ColorPicker *p_color_picker);
 	virtual ~ColorMode() {}
@@ -72,6 +73,7 @@ public:
 	virtual float get_slider_step() const override { return 1.0; }
 	virtual String get_slider_label(int idx) const override;
 	virtual float get_slider_max(int idx) const override;
+	virtual float get_spinbox_max(int idx) const override { return get_slider_max(idx); }
 	virtual float get_slider_value(int idx) const override;
 
 	virtual Color get_color() const override;
@@ -87,12 +89,16 @@ public:
 class ColorModeRGB : public ColorMode {
 public:
 	String labels[3] = { "R", "G", "B" };
+	float slider_max[4] = { 255, 255, 255, 255 };
+	float spinbox_max[4] = { 2550, 2550, 2550, 255 };
 
 	virtual String get_name() const override { return "RGB"; }
 
 	virtual float get_slider_step() const override { return 1; }
 	virtual String get_slider_label(int idx) const override;
 	virtual float get_slider_max(int idx) const override;
+	virtual float get_spinbox_max(int idx) const override;
+	virtual bool can_allow_greater() const override { return true; }
 	virtual float get_slider_value(int idx) const override;
 
 	virtual Color get_color() const override;
@@ -106,20 +112,21 @@ public:
 class ColorModeRAW : public ColorMode {
 public:
 	String labels[3] = { "R", "G", "B" };
-	float slider_max[4] = { 100, 100, 100, 1 };
+	float slider_max[4] = { 1, 1, 1, 1 };
+	float spinbox_max[4] = { 10, 10, 10, 255 };
 
 	virtual String get_name() const override { return "RAW"; }
 
-	virtual float get_slider_step() const override { return 0.001; }
-	virtual float get_spinbox_arrow_step() const override { return 0.01; }
+	virtual float get_slider_step() const override { return 1.0 / 255.0; }
 	virtual String get_slider_label(int idx) const override;
 	virtual float get_slider_max(int idx) const override;
+	virtual float get_spinbox_max(int idx) const override;
+	virtual bool can_allow_greater() const override { return true; }
 	virtual float get_slider_value(int idx) const override;
 
 	virtual Color get_color() const override;
 
 	virtual void slider_draw(int p_which) override;
-	virtual bool apply_theme() const override;
 
 	ColorModeRAW(ColorPicker *p_color_picker) :
 			ColorMode(p_color_picker) {}
@@ -138,6 +145,7 @@ public:
 	virtual float get_slider_step() const override { return 1.0; }
 	virtual String get_slider_label(int idx) const override;
 	virtual float get_slider_max(int idx) const override;
+	virtual float get_spinbox_max(int idx) const override { return get_slider_max(idx); }
 	virtual float get_slider_value(int idx) const override;
 
 	virtual Color get_color() const override;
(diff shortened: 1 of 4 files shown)


WhalesState:
Please add the bug label too, since it fixes three annoying issues.

AThousandShips:
Please open a bug report first :)

WhalesState:
> Remember you need to add "fixes" before _each_ issue number

Thanks ^^

AThousandShips:
See also:
* https://github.com/godotengine/godot/pull/62913

WhalesState:
> See also:
> 
> * [fix switching RAW mode to RGB mode issue #62913](https://github.com/godotengine/godot/pull/62913)

We share the same fix
from `float slider_max[4] = { 100, 100, 100, 1 };` to `float slider_max[4] = { 1, 1, 1, 1 };`

Cammymoop:
Being able to set values greater than 1.0 in raw mode is a feature, not a bug. #62894 is about not restoring the maximum value of the slider when going back to RGB mode, not changing the max value in raw mode.

WhalesState:
> Being able to set values greater than 1.0 in raw mode is a feature, not a bug. #62894 is about not restoring the maximum value of the slider when going back to RGB mode, not changing the max value in raw mode.

we can still allow this feature by making the spin box able to set values higher than 1.0 by enabling `allow_greater`, but I still don't understand the use of it, since they will not be visualized correctly, is it for glow/emission strength to make some colors more brighter that others? maybe because the glow the strength is a global variable for each environment.

Cammymoop:
Yes, it is for colors that are overbright in cases such as emissive colors with glow effects, or using modulate to brighten something.

I wouldn't suggest using allow_greater on the RAW mode sliders and limiting them to 1, because afaik the main reason raw mode is used at all is for overbright colors. RGB mode uses code specifically (see `get_maximum_value`) to set different slider max values in the case where you switched from raw and still want to edit the overbright color I guess but I'm not sure that that part is usefull, and you could just as easily switch to HSV instead (which would be more helpful to edit an overbright color) now that raw is a separate tab from RGB, but HSV does clamp the overbright color so that's pointless currently. 

A proper way to deal with overbright colors on all color modes (probably a separate intensity slider) is the only elegant fix I think, but I think making RGB mode use `allow_greater` or just making it clamp is the way to go for this bug. HSV could use `allow_greater` theoretically, but I don't think the internal conversion supports that properly, okhsl will definitely not work with unclamped lightness so don't bother with that one.

WhalesState:
ok, after allowing both RGB and RAW modes to have greater values, we are back again to this issue https://github.com/godotengine/godot/issues/62894, for some reason the slider maximum value changes after converting from RAW to RGB.

https://github.com/godotengine/godot/assets/53877170/17354eb9-b5f0-4355-a0fb-feb4b2cff1b7

MajorMcDoom:
Hi there, I was going to ask, while we're at it, is it difficult to make it possible to accept negative values as well? I know the most common use case is overbright colors, but negative colors are also used in many graphics tricks, and being able to export one would be great. The thing is that each component in `Color` can, in code, represent any float. But when exported in the inspector, the values are artificially being constrained by the _UX_ of the `ColorPicker`. I think that's quite a shame, and it would be great if we could get over that hurdle with this PR. But if it's too difficult, I think getting over the limit of 100 is already a pretty big win.

Calinou:
> Hi there, I was going to ask, while we're at it, is it difficult to make it possible to accept negative values as well? I know the most common use case is overbright colors, but negative colors are also used in many graphics tricks, and being able to export one would be great. The thing is that each component in `Color` can, in code, represent any float. But when exported in the inspector, the values are artificially being constrained by the _UX_ of the `ColorPicker`. I think that's quite a shame, and it would be great if we could get over that hurdle with this PR. But if it's too difficult, I think getting over the limit of 100 is already a pretty big win.

The issue is that many properties can't handle negative colors by design (or it'll result in unintended effects). We'd need additional property hints to denote that specific properties can't use negative colors, similar to the existing `PROPERTY_HINT_COLOR_NO_ALPHA`.

scene/gui/color_picker.h
```diff
@@ -276,7 +276,6 @@ class ColorPicker : public VBoxContainer {
 	void _copy_color_to_hsv();
 	void _copy_hsv_to_color();
 
-	PickerShapeType _get_actual_shape() const;
```
KoBeWi, line 279: Looks like unrelated change.

KoBeWi:
Still not convinced about slider max value change. While it's possible to input bigger number in the SpinBox, you lose ability to adjust overbright via dragging, and I know some people like doing that. Not to mention that pressing Enter to accept SpinBox value will close the popup.

WhalesState:
> Still not convinced about slider max value change. While it's possible to input bigger number in the SpinBox, you lose ability to adjust overbright via dragging, and I know some people like doing that. Not to mention that pressing Enter to accept SpinBox value will close the popup.

For RAW and RGB modes, We can allow dragging using the `SpinBox`, they can click and hold the up/down button and move the mouse up or down to change the value, we just need to change the maximum value for the `SpinBox` only and not the sliders to not lose the coloring for RAW mode, also a maximum value of 10 can be more usable than 100.

scene/gui/color_picker.h
WhalesState: This change is here for nearly a year now, and https://github.com/godotengine/godot/pull/99662 has merge conflict and was opened 9 months after this PR.

KoBeWi:
With all these `get()` methods in ColorMode, I wonder if we shouldn't just introduce a single ColorModeConfig struct, which would hold all these parameters (not in this PR).
`get_spinbox_max()` would solve the problem.

WhalesState:
Maybe i can use this to return the slider max value by default `virtual float get_spinbox_max(int idx) const { return get_slider_max(idx); }`.

KoBeWi:
Makes sense.

WhalesState:
Failed because Spinbox and Slider are shared, will have to update them manually for this to work.

WhalesState:
Added a new class `OverbrightSpinbox` and used it for RGB and RAW modes.

First i have tried to disconnect the `value_changed` and connects it again after updating the slider but I was afraid if it will be slow, so i just have connected the `value_changed` of the slider to `_update` which will return early to avoid calling `set_value_no_signal` twice.

Didn't test it yet since it's still building after the rebase, Just need to know if this approach looks fine.

WhalesState:
Works as expected.

Only the alpha slider shares it's value with the alpha spinbox, I ended up using the new `OverBrightSpinBox` for all the other SpinBoxes to avoid deleting/recreating spinboxes.

[Screencast from 01-26-2025 02:34:15 AM.webm](https://github.com/user-attachments/assets/b975b14e-52a5-4a5f-a305-9b358735db4e)

scene/gui/color_mode.h
```diff
@@ -106,20 +112,21 @@ class ColorModeRGB : public ColorMode {
 class ColorModeRAW : public ColorMode {
 public:
 	String labels[3] = { "R", "G", "B" };
-	float slider_max[4] = { 100, 100, 100, 1 };
+	float slider_max[4] = { 1, 1, 1, 1 };
+	float spinbox_max[4] = { 10, 10, 10, 255 };
```
WhalesState, line 116: Note: The 4th value is never used because we don't apply this to the alpha `SpinBox`. Maybe i can reduce the array size to `3`.

scene/gui/color_mode.h
```diff
@@ -106,20 +112,21 @@ class ColorModeRGB : public ColorMode {
...
 
 	virtual String get_name() const override { return "RAW"; }
 
-	virtual float get_slider_step() const override { return 0.001; }
-	virtual float get_spinbox_arrow_step() const override { return 0.01; }
+	virtual float get_slider_step() const override { return 1.0 / 255.0; }
```
WhalesState, line 120: This also seems not accurate if the value goes over 255.

PR #88074, opened 2024-02-07
Add colors to SCons error/warning messages

### LINKED ISSUE #9291: Make warnings/notes in build scripts colored

Just build custom export template and try master branch

The build process shows me blue text (when it compiles files) but the warnings have the same white colors as any other text

When you see a bunch of white text for especially when you enable SCU you can't notice if there was any warning because it scrolls out too quickly. If I saw a yellow color between white I would notice it...

Use not just `print("Note: Using `execinfo=yes` for the crash handler as required on platforms where glibc is missing.")` but something [like it](https://stackoverflow.com/questions/287871/how-do-i-print-colored-text-to-the-terminal) at least in [linux specific script](https://github.com/godotengine/godot/blob/master/platform/linuxbsd/detect.py)

It is for those who enter godot building or just want custom export template so it won't be used much but can't be worked around

It is a part of the build system

Calinou:
This also includes various improvements to the messages:

- Remove some "success" messages that were printed on every build, such as the macOS minimum supported version (which is already mentioned in the documentation. Only the "error" variants of these messages are kept.
- Use gray color for the build timer printed at the end of each build, and only display 2 decimals.
- Use non-zero exit codes when exiting after some error messages that previously exited with a zero (success) exit code.
- Improve formatting of the Direct3D 12 missing dependencies message.
- Clarify how MSVC is detected in the message printed when it's detected.
- Force colored output on CI (it's non-interactive but can still display colors).

___

- This closes https://github.com/godotengine/godot-proposals/issues/9291.

## Preview

Screenshot_20240207_202104

Screenshot_20240207_202502

Screenshot_20240207_202726

I also had automatic underlining of backticked text at some point, but I removed it for now. Let me know if you'd like me to readd it:

Screenshot_20240207_201931

### DIFF
--- a/platform/android/detect.py
+++ b/platform/android/detect.py
@@ -2,6 +2,7 @@ import os
 import sys
 import platform
 import subprocess
+from methods import print_error, print_info, print_warning
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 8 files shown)


deralmas (approved):
Pretty!

platform/android/detect.py
```diff
@@ -76,40 +77,40 @@ def get_flags():
...
-                "Please ensure ANDROID_HOME is correct and cmdline-tools are installed, or install NDK version "
+            print_error(
+                "Cannot find "
+                + sdkmanager
+                + "Please ensure ANDROID_HOME is correct and cmdline-tools are installed, or install NDK version "
                 + get_ndk_version()
                 + " manually."
             )
-            sys.exit()
+            sys.exit(200)
```
akien-mga, line 97: Why `200` here specifically? We seem to use `255` throughout otherwise.

akien-mga: Ah I see we had this:
```
platform/windows/detect.py
305:        sys.exit(200)
```
That's the only occurrence, so I'd tend to say it was a mistake and not an intentional error reporting API.

akien-mga:
Superseded by #91220.

PR #88075, opened 2024-02-07
libktx: Update to 4.3.1

Chubercik:
https://github.com/KhronosGroup/KTX-Software/releases/tag/v4.3.1

Also updated `thirdparty/README.md` to be more precise and regenerated `thirdparty/libktx/godot.patch`.

~**Edit**: Bundled `miniz` lib from upstream.~


Chubercik:
Seems like the guards employed in `miniz_wrapper.cpp` aren't working properly which results in a few multiply defined symbols - not sure how to fix that, haven't been exposed to such issues in a while..

thirdparty/README.md
```diff
@@ -426,19 +426,26 @@ Files extracted from upstream source:
...
-- Version: 4.1.0 (d7255fe73cd53b856731ceb9f2c279181d0dbbca, 2023)
+- Version: 4.3.1 (c0214158d551cfc779624b0f84130bcbbefef59a, 2024)
 - License: Apache-2.0
 
 Files extracted from upstream source:
 
 - `LICENSE.md`
 - `include/*`
-- `lib/dfdutils/{LICENSES/Apache-2.0.txt,KHR,*.c,*.h,*.inl}`
-- `lib/{basis_sgd.h,basis_transcode.cpp,checkheader.c,filestream.*,formatsize.h,gl_format.h,hashlist.c,ktxint.h,memstream.*,swap.c,texture*,uthash.h,vk_format.h,vkformat_enum.h}`
```
akien-mga, line 437: The reason for this format is that you can copy paste it:
```
cp path/to/ktx/source/lib/{basis_sgd.h,basis_transcode.cpp,checkheader.c,filestream.*,formatsize.h,gl_format.h,hashlist.c,ktxint.h,memstream.*,swap.c,texture*,uthash.h,vk_format.h,vkformat_enum.h} thirdparty/ktx/lib/
```
Splitting it all in a nicely formatted list looks better in the .md file, but it's arguably more tedious to actually copy the files.

Chubercik: Gotcha, I'll revert it later on

Chubercik:
Not sure what to do here, failed builds are the result of one `include` statement in `miniz_wrapper.cpp` - if the file is loaded as is, the linker complains about multiple definitions, but when only the header part of the file is loaded, template builds fail due to missing definitions.

I made some additional `define` statements in `miniz_wrapper.cpp`, so that `basisu_miniz.h` is loaded similarly as it is in the original `basis_universal` repository - this way at least editor builds compile.

thirdparty/libktx/lib/texture2.c
```diff
@@ -2107,21 +2239,34 @@ ktxTexture2_IterateLoadLevelFaces(ktxTexture2* This, PFNKTXITERCB iterCb,
...
             }
+
             // We don't fix up the texture's dataSize, levelIndex or
             // _requiredAlignment because after this function completes there
             // is no way to get at the texture's data.
             //nindex[level].byteOffset = levelOffset;
             //nindex[level].uncompressedByteLength = nindex[level].byteLength =
                                                                 //levelByteLength;
+        } else if (This->supercompressionScheme == KTX_SS_ZLIB) {
+            result = ktxUncompressZLIBInt(uncompressedDataBuf,
```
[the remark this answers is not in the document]
Chubercik: Yet another solution would be to write a Godot-specific file with definitions for functions `ktxCompressZLIBBounds`, `ktxCompressZLIBInt` and `ktxUncompressZLIBInt` so that we don't need to include `basisu_miniz.h`, but that seems like an unnecessary effort and there should be a better way to solve this.

fire (approved):
Please squash the commits https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html but I am generally ok with library upgrades.

Probably want to test a basisu ktx from gltf from https://github.com/KhronosGroup/glTF-Sample-Models/tree/main/2.0/StainedGlassLamp

COPYRIGHT.txt
```diff
@@ -389,6 +389,13 @@ Comment: Tangent Space Normal Maps implementation
 Copyright: 2011, Morten S. Mikkelsen
 License: Zlib
 
+Files: ./thirdparty/misc/miniz.c
+ ./thirdparty/misc/miniz.h
+Comment: miniz - lossless, high performance data compression library
```
akien-mga, line 394: This is normally just the name of the library, not its full description.
```suggestion
Comment: miniz
```

thirdparty/README.md
```diff
@@ -640,6 +642,10 @@ Collection of single-file libraries used in Godot components.
   * Upstream: https://archive.blender.org/wiki/index.php/Dev:Shading/Tangent_Space_Normal_Maps/
   * Version: 1.0 (2011)
   * License: zlib
+- `miniz.{c,h}`
+  * Upstream: https://github.com/richgel999/miniz
+  * Version: 3.0.2 (293d4db1b7d0ffee9756d035b9ac6f7431ef8492, 2023)
+  * License: MIT
```
akien-mga, line 648: If we're adding miniz as a standalone library, we should probably remove the copy from `thirdparty/basis_universal/encoder/basisu_miniz.h` so we don't have duplicate symbols and potentially mismatched versions of it.

Chubercik: I added `miniz` as a last resort after failing, time and time again, to use the `basisu_miniz` amalgamation - it's problematic, but I think I know how to get it to work now, so before doubling down on having a standalone lib I wanna explore using the `#ifdef TOOLS_ENABLED` preprocessor directive to separate inclusion of just the header part in the case of editor builds, and the whole file with definitions in the case of template builds.

Chubercik:
This compiles on my end, both editor and template build; I'm gonna hold on before squishing commits for someone to approve the PR, if everything is safe and sound I'll push a rebase.

akien-mga:
Changes look good overall!

The miniz situation can be solved in a better way with this extra diff:
```diff
diff --git a/modules/ktx/SCsub b/modules/ktx/SCsub
index 3a37ef93ee..887ffa5d14 100644
--- a/modules/ktx/SCsub
+++ b/modules/ktx/SCsub
@@ -12,11 +12,11 @@ thirdparty_obj = []
 thirdparty_dir = "#thirdparty/libktx/"
 thirdparty_sources = [
     "lib/basis_transcode.cpp",
-    "lib/miniz_wrapper.cpp",
     "lib/checkheader.c",
     "lib/filestream.c",
     "lib/hashlist.c",
     "lib/memstream.c",
+    "lib/miniz_wrapper.cpp",
     "lib/swap.c",
     "lib/texture.c",
     "lib/texture1.c",
@@ -35,7 +35,10 @@ env_ktx.Prepend(CPPPATH=[thirdparty_dir + "include"])
 env_ktx.Prepend(CPPPATH=[thirdparty_dir + "utils"])
 env_ktx.Prepend(CPPPATH=[thirdparty_dir + "lib"])
 env_ktx.Prepend(CPPPATH=[thirdparty_dir + "other_include"])
+
 env_ktx.Prepend(CPPPATH=["#thirdparty/basis_universal"])
+# We already build the miniz implementation in the basis_universal module.
+env_ktx.Append(CPPDEFINES=["MINIZ_HEADER_FILE_ONLY"])
 
 if env["vulkan"]:
     env_ktx.Prepend(CPPPATH=["#thirdparty/vulkan/include"])
diff --git a/thirdparty/libktx/lib/miniz_wrapper.cpp b/thirdparty/libktx/lib/miniz_wrapper.cpp
index 3707c1a806..cbd7da540a 100644
--- a/thirdparty/libktx/lib/miniz_wrapper.cpp
+++ b/thirdparty/libktx/lib/miniz_wrapper.cpp
@@ -22,7 +22,7 @@
 
 #include <assert.h>
 
-#ifndef TOOLS_ENABLED
+#if !KTX_FEATURE_WRITE
 // The reader does not link with the basisu components that already include a
 // definition of miniz so we include it here explicitly.
 #ifdef __GNUC__
diff --git a/thirdparty/libktx/patches/godot.patch b/thirdparty/libktx/patches/godot.patch
index 61a7997b97..28db86cf9b 100644
--- a/thirdparty/libktx/patches/godot.patch
+++ b/thirdparty/libktx/patches/godot.patch
@@ -36,18 +36,9 @@ index 85d53202a5..25c7a2c238 100644
      int channels[] = {0,1,2,3}; int bits[] = {5,5,5,1};
      return createDFDPacked(0, 4, bits, channels, s_UNORM);
 diff --git a/thirdparty/libktx/lib/miniz_wrapper.cpp b/thirdparty/libktx/lib/miniz_wrapper.cpp
-index 07920c4809..3707c1a806 100644
+index 07920c4809..cbd7da540a 100644
 --- a/thirdparty/libktx/lib/miniz_wrapper.cpp
 +++ b/thirdparty/libktx/lib/miniz_wrapper.cpp
-@@ -22,7 +22,7 @@
- 
- #include <assert.h>
- 
--#if !KTX_FEATURE_WRITE
-+#ifndef TOOLS_ENABLED
- // The reader does not link with the basisu components that already include a
- // definition of miniz so we include it here explicitly.
- #ifdef __GNUC__
 @@ -30,7 +30,7 @@
  #pragma GCC diagnostic ignored "-Wextra"
  #pragma GCC diagnostic ignored "-Wmisleading-indentation"
```

Basically passing `MINIZ_HEADER_FILE_ONLY` solves the issue - the header should only be included once without that define, to compile the implementation (equivalent to what a .cpp file should be for a normal two-file library). Yes, single header libraries sounded like a cool idea but they're dumb in practice.

Chubercik:
Seems like we're back to square one:

> Not sure what to do here, failed builds are the result of one `include` statement in `miniz_wrapper.cpp` - if the file is loaded as is, the linker complains about multiple definitions, but when only the header part of the file is loaded, template builds fail due to missing definitions.

The `#ifndef TOOLS_ENABLED` guard bypassed this issue by loading the definitions in the case of a template build.

akien-mga:
That's because we don't build the basis universal encoder in non-editor builds: https://github.com/godotengine/godot/blob/16d61427cab3a8e43f0a9a8ee724fc176b6433c6/modules/basis_universal/SCsub#L54

So miniz is not compiled.

So we should pass `MINIZ_HEADER_FILE_ONLY` conditionally (`if env.editor_build:  # Already implemented by basis encoder.`).

akien-mga (approved):

akien-mga:
Thanks!

PR #88076, opened 2024-02-07
Fix C# "out of sync" notice with external editors

paulloz:
The notice introduced in #85869 wasn't being dequeued when building from an external IDE. This PR fixes it by changing what triggers our reevaluation of the last valid build time. We used to react only to Godot's internal build process. We're now reacting to assembly reload, which is triggered both by building from Godot, and from the watcher. 

Thanks @Delsin-Yu @zaevi for pointing out the issue.


raulsntos (approved):
Works great, thanks!

akien-mga:
Thanks!

PR #88078, opened 2024-02-07
Support setting 3.x `material/#` properties on MeshInstance3D

nikitalita:
This PR introduces support for setting the old `material/#` properties on MeshInstance3D, which were renamed to `surface_material_override/#` in 4.x. This is only supported when using `set()` (i.e. during the initial load) and emits a warning when doing so.

Note that this should not conflict with ImporterMesh; This will only trigger if the property is named `material/{num}`. I have checked, and MeshInstance3D was the only resource that stored a property named `material` INI style in either 3.x or 4.x.

### DIFF
--- a/scene/3d/mesh_instance_3d.cpp
+++ b/scene/3d/mesh_instance_3d.cpp
@@ -62,8 +62,19 @@ bool MeshInstance3D::_set(const StringName &p_name, const Variant &p_value) {
 		return true;
 	}
 
-	if (p_name.operator String().begins_with("surface_material_override/")) {
-		int idx = p_name.operator String().get_slicec('/', 1).to_int();
+	String name = p_name;
+
+#ifndef DISABLE_DEPRECATED
+	if (name.begins_with("material/")) {
+		WARN_DEPRECATED_MSG("This mesh uses an old deprecated parameter name. Consider re-saving this scene in order for it to continue working in future versions." +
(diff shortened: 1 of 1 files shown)


fire:
If it's not too much hassle can you provide a test case?

nikitalita:
Here's an example I threw together: [example_meshinstance3d_mesh.zip](https://github.com/godotengine/godot/files/14230598/example_meshinstance3d_mesh.zip)

Without this patch:
image
With this patch:
image

nikitalita:
@fire same here

Calinou (commented):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/8ed125b42908d0d46d3b8967e3a3bc03f809b3af), it works as expected.

Note that in the MRP, I don't see the mesh path being printed even though the meshes in question are clearly part of the scene tree:

image

Perhaps the conversion step occurs too early for it to be considered to be part of the scene tree?

nikitalita:
> Perhaps the conversion step occurs too early for it to be considered to be part of the scene tree?

It looks like that is indeed the case; it looks like the node properties are set before it actually gets added. I have modified it such that it displays the name if it's not empty.

aaronfranke (approved):
Makes sense to me. These local variables could be made const but anyway this is a pretty straight-forward change that provides compatibility with the old names, with a warning, and only when compiling with deprecated features enabled.

PR #88081, opened 2024-02-07
Add editor shortcuts to toggle bottom panel visibility

### LINKED ISSUE #380: Add hotkeys for the editor's bottom panels [output, debugger, search results, audio, animation]

Recently started out so I don't have a project that directly relates to my proposal. In using the Godot editor, often times the output window opens automatically and has to be closed with a mouse click. This could be done quicker with the press of a hotkey. Same applies for the debugger which is needed often to view hints/error messages. A quick "open, read and close" cycle could be done with keyboard via hotkeys for it.

I often find myself in the position to have to switch between keyboard and mouse controls. This workflow for coding/testing could be smoothened by introducing keyboard hotkeys.

**Describe how this feature / enhancement will help you overcome this problem or limitation:**

Faster access to opening and closing the respective panels without having to grab the mouse.

In my humble opinion, a mocked up screenshot would not be very helpful to better understand my proposal.

Set hotkeys "Alt + [1, 2, 3, 4, 5]" as default for opening/closing [output, debugger, search results, audio, animation]. These key combinations are easily accessible with one hand.

I do not know if others would be interested in this as well. Personally I could see myself using the feature quite often. Thus I also cannot tell if a few lines of script might fix the problem because I cannot code in C++ and don't know the inner workings of the editor.

It covers core features of the engine which are used very frequently. An add-on might fix the issue but the proposal is to include it as a core feature for reasons of fundamentality.

Calinou:
Default shortcuts typically use the first or second letter of each word (with a few exceptions such as TileMap and TileSet to ensure consistency). Editor plugins can also specify shortcuts for their own bottom panels.

This also adds also a new shortcut to toggle the last opened bottom panel. On editor startup, this defaults to the first panel in the list (which is the Output panel).

All shortcuts are nearly organized in their own section of the Editor Settings Shortcuts dialog for easy access, and are also available in the command palette:

image

One issue is that the individual toggle shortcuts don't work if you're currently focusing on the Tree node in either the scene tree dock or FileSystem dock. Shortcuts will work if you're focused on their filter LineEdits though. I've tried to use `tb->set_shortcut_context(nullptr)` to no avail.

- This closes https://github.com/godotengine/godot-proposals/issues/380.

## Preview

https://github.com/godotengine/godot/assets/180032/96dbfbe7-597a-4130-af80-97d74c2a79f7

The shortcut appears in the tooltip when the button is hovered:

image

### DIFF
--- a/doc/classes/EditorPlugin.xml
+++ b/doc/classes/EditorPlugin.xml
@@ -409,9 +409,10 @@
 			<return type="Button" />
 			<param index="0" name="control" type="Control" />
 			<param index="1" name="title" type="String" />
+			<param index="2" name="shortcut" type="Shortcut" default="null" />
 			<description>
-				Adds a control to the bottom panel (together with [b]Output[/b], [b]Debug[/b], [b]Animation[/b], etc.). Returns the button added to the tab bar. It's up to you to manage the button's visibility as needed.
-				When your plugin is deactivated, make sure to remove your custom control with [method remove_control_from_bottom_panel] and free it with [method Node.queue_free].
+				Adds a control to the bottom panel (together with Output, Debug, Animation, etc). Returns a reference to the button added. It's up to you to hide/show the button when needed. When your plugin is deactivated, make sure to remove your custom control with [method remove_control_from_bottom_panel] and free it with [method Node.queue_free].
+				Optionally, you can specify a shortcut parameter. When pressed, this shortcut will toggle the bottom panel's visibility. See the default editor bottom panel shortcuts in the Editor Settings for inspiration. Per convention, they all use [kbd]Alt[/kbd] modifier.
 			</description>
 		</method>
 		<method name="add_control_to_container">
@@ -428,10 +429,12 @@
 			<return type="void" />
 			<param index="0" name="slot" type="int" enum="EditorPlugin.DockSlot" />
 			<param index="1" name="control" type="Control" />
+			<param index="2" name="shortcut" type="Shortcut" default="null" />
 			<description>
 				Adds the control to a specific dock slot (see [enum DockSlot] for options).
 				If the dock is repositioned and as long as the plugin is active, the editor will save the dock position on further sessions.
 				When your plugin is deactivated, make sure to remove your custom control with [method remove_control_from_docks] and free it with [method Node.queue_free].
+				Optionally, you can specify a shortcut parameter. When pressed, this shortcut will toggle the dock's visibility once it's moved to the bottom panel (this shortcut does not affect the dock otherwise). See the default editor bottom panel shortcuts in the Editor Settings for inspiration. Per convention, they all use [kbd]Alt[/kbd] modifier.
 			</description>
 		</method>
 		<method name="add_custom_type">
--- a/editor/editor_dock_manager.cpp
+++ b/editor/editor_dock_manager.cpp
@@ -38,6 +38,7 @@
 #include "scene/gui/tab_container.h"
 #include "scene/main/window.h"
 
+#include "editor/editor_command_palette.h"
 #include "editor/editor_node.h"
 #include "editor/editor_settings.h"
 #include "editor/editor_string_names.h"
@@ -47,6 +48,8 @@
 #include "editor/window_wrapper.h"
 
 EditorDockManager *EditorDockManager::singleton = nullptr;
+
+static const char *META_TOGGLE_SHORTCUT = "_toggle_shortcut";
 
 void DockSplitContainer::_update_visibility() {
 	if (is_updating) {
@@ -392,7 +395,10 @@ void EditorDockManager::_dock_move_selected_to_bottom() {
 	dock->call("_set_dock_horizontal", true);
 
 	bottom_docks.push_back(dock);
-	EditorNode::get_bottom_panel()->add_item(dock->get_name(), dock, true);
+
+	// Force docks moved to the bottom to appear first in the list, and give them their associated shortcut to toggle their bottom panel.
+	EditorNode::get_bottom_panel()->add_item(dock->get_name(), dock, dock->get_meta(META_TOGGLE_SHORTCUT), true);
+
 	dock_select_popup->hide();
 	update_dock_slots_visibility(true);
 	_edit_current();
@@ -663,7 +669,8 @@ void EditorDockManager::load_docks_from_config(Ref<ConfigFile> p_layout, const S
 			node->call("_set_dock_horizontal", true);
 
 			bottom_docks.push_back(node);
-			EditorNode::get_bottom_panel()->add_item(node->get_name(), node, true);
+			// Force docks moved to the bottom to appear first in the list, and give them their associated shortcut to toggle their bottom panel.
+			EditorNode::get_bottom_panel()->add_item(node->get_name(), node, node->get_meta(META_TOGGLE_SHORTCUT), true);
 		}
 	}
 
@@ -730,8 +737,9 @@ void EditorDockManager::close_all_floating_docks() {
 	}
 }
 
-void EditorDockManager::add_control_to_dock(DockSlot p_slot, Control *p_control, const String &p_name) {
+void EditorDockManager::add_control_to_dock(DockSlot p_slot, Control *p_control, const String &p_name, const Ref<Shortcut> &p_shortcut) {
 	ERR_FAIL_INDEX(p_slot, DOCK_SLOT_MAX);
+	p_control->set_meta(META_TOGGLE_SHORTCUT, p_shortcut);
 	dock_slot[p_slot]->add_child(p_control);
 	if (!p_name.is_empty()) {
 		dock_slot[p_slot]->set_tab_title(dock_slot[p_slot]->get_tab_idx_from_control(p_control), p_name);
--- a/editor/editor_plugin.compat.inc
+++ b/editor/editor_plugin.compat.inc
@@ -1,5 +1,5 @@
 /**************************************************************************/
-/*  editor_bottom_panel.h                                                 */
+/*  editor_plugin.compat.inc                                              */
 /**************************************************************************/
 /*                         This file is part of:                          */
 /*                             GODOT ENGINE                               */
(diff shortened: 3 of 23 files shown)


Mickeon:
I feel like a default shortcut set for **every single one** of those panels is overkill. Outside of the always-present 5 ones, it is also extremely unlikely that all these panels will be open at the same time

If it is later planned to use any of those Alt combinations for anything else it becomes much harder to argue with (Script Editor's `CTRL + D` is already a big example right now).

kitbdev:
Should we add them to the command palette as well? Since there are commands to switch main screens.

passivestar:
Yeah I agree with @Mickeon, it's probably best to leave all of the non-persistent buttons unbound and let the users bind what they need. For example I'd rarely need a hotkey for TileMap just because I don't make 2D games, Replication is only needed if you make multiplayer games, etc etc

Calinou:
> If it is later planned to use any of those Alt combinations for anything else it becomes much harder to argue with (Script Editor's CTRL + D is already a big example right now).

I can change the permanent bottom panels' shortcuts to take priority over the temporary ones.

ajreckof:
So first I'm not sure if this a problem per se but they are two bottom panel associated with `Alt+T` 
The second is a real problem. On mac Option + Letter will result in special symbols.
Option + N -> ~
Option + T -> †
Option + A -> æ
Option + D -> ∂
Option + O -> œ
Option + L -> ¬
Option + E -> ê
Option + R -> ®
Option + H -> Ì
Option + I -> î
Option + P -> π
Option + T -> †
Option + M -> µ
Option + S -> Ò
I think out of all of them only the R will not trigger a writing this means whenever the script main panel is opened using those shortcuts will add some random characters in the middle of your scripts.  On the other end preventing them could be catastrophic for example Option + N and Option + O are two that I use a lot when writing strings. I feel the best solution should to have different shortcuts for Mac Not 100% sure how it was handled for other Alt + Letter shortcuts that might have been added previously.

Mickeon:
There are some default shortcuts on Mac that are different for similar reasons, so they would need to either be something else or stripped out entirely.

Cammymoop:
Can we include a shortcut to toggle? Hide current panel if visible, otherwise show last open panel. Doesn't need a default mapping but I think it's worth one. Similar thing in VSCode is ctl-j not sure why J but it does seem free to map.

Calinou:
> Can we include a shortcut to toggle? Hide current panel if visible, otherwise show last open panel. Doesn't need a default mapping but I think it's worth one. Similar thing in VSCode is ctl-j not sure why J but it does seem free to map.

Good idea; I've added <kbd>Ctrl + J</kbd> to toggle the last opened bottom panel.

KoBeWi:
Needs rebase.

Calinou:
> Needs rebase.

Rebased and tested again, it works as expected.

editor/plugins/tiles/tiles_editor_plugin.cpp
```diff
@@ -562,7 +563,7 @@ TileSetEditorPlugin::TileSetEditorPlugin() {
 	editor->set_custom_minimum_size(Size2(0, 200) * EDSCALE);
 	editor->hide();
 
-	button = EditorNode::get_bottom_panel()->add_item(TTR("TileSet"), editor);
+	button = EditorNode::get_bottom_panel()->add_item(TTR("TileSet"), editor, ED_SHORTCUT_AND_COMMAND("bottom_panels/toggle_tile_set_bottom_panel", TTR("Toggle TileSet Bottom Panel"), KeyModifierMask::ALT | Key::S));
```
KoBeWi, line 566: TileMap and TileSet shortcuts should be close together I think.

KoBeWi:
> Should we add them to the command palette as well? Since there are commands to switch main screens.

The problem is that main screens are always available, unlike bottom panels. E.g. TileMap command does nothing without TileMap.

I'm thinking, maybe instead of adding separate shortcuts to every single panel, we could do something similar to browser tabs, i.e. add shortcuts to toggle the tabs based on their index? Alt+1 to Alt+8 for example.

Mickeon:
I prefer @KoBeWi 's suggestion by such a large margin I felt the need to communicate it as well.

ajreckof:
I really like KobeWi suggestion but I remember some PR wanting to not show search Results in bottom panel until there is an actual search but search result is one of the first and I think bottom panels that are always there should have a fixed way to toggle them. so either we put all permanent before all other or maybe for permanent one there is a two way to access them by their index shortcut and by their spcific shortcut.
Another idea based on browser is to have shortcut to go next previous panel.

KoBeWi:
> For panels which should not be currently available can't we just check if the panel is or isn't hidden and just do nothing when it's hidden?

Well that's what happens currently, but you can see hidden panel options in Command Palette and they just do nothing.

viksl:
Hm, maybe we could just leave it as is for the user to realize since this one is sort of realtively clear and if reports of this becoming an issue come then remove the non-permanent ones as you suggest?
But I'll be honest I don't really have a way to gauge how problematic this is/isn't :D.

doc/classes/EditorPlugin.xml
```diff
@@ -409,8 +409,10 @@
 			<return type="Button" />
 			<param index="0" name="control" type="Control" />
 			<param index="1" name="title" type="String" />
+			<param index="2" name="shortcut" type="Shortcut" default="null" />
 			<description>
 				Adds a control to the bottom panel (together with Output, Debug, Animation, etc). Returns a reference to the button added. It's up to you to hide/show the button when needed. When your plugin is deactivated, make sure to remove your custom control with [method remove_control_from_bottom_panel] and free it with [method Node.queue_free].
+				Optionally, you can specify a shortcut parameter which should use the [kbd]Alt[/kbd] modifier as per convention. When pressed, this shortcut will toggle the bottom panel's visibility. See the default editor bottom panel shortcuts in the Editor Settings for inspiration.
```
KoBeWi, line 415:
```suggestion
				Optionally, you can specify a shortcut parameter. When pressed, this shortcut will toggle the bottom panel's visibility. See the default editor bottom panel shortcuts in the Editor Settings for inspiration. Per convention, they all use [kbd]Alt[/kbd] modifier.
```
Currently it sounds like it's some requirement, while it's not.

editor/editor_dock_manager.cpp
```diff
@@ -392,7 +393,10 @@ void EditorDockManager::_dock_move_selected_to_bottom() {
 	dock->call("_set_dock_horizontal", true);
 
 	bottom_docks.push_back(dock);
-	EditorNode::get_bottom_panel()->add_item(dock->get_name(), dock, true);
+	// Force FileSystem dock to appear first in the list, and give it a shortcut.
+	// Since only the FileSystem dock can currently be moved to the bottom,
+	// the shortcut is assumed to be for the FileSystem dock.
+	EditorNode::get_bottom_panel()->add_item(dock->get_name(), dock, ED_SHORTCUT_AND_COMMAND("bottom_panels/toggle_filesystem_bottom_panel", TTR("Toggle FileSystem Bottom Panel"), KeyModifierMask::ALT | Key::F), true);
```
KoBeWi, line 399: We shouldn't make such assumptions. You could either check if `dock` is FileSystemDock or add a method to get bottom panel shortcut (can be called with `call()` like the one above).

kitbdev: Also, this shortcut can't be added here or in load since it can be called multiple times, leading to errors showing `The Command 'Toggle FileSystem Bottom Panel' already exists. Unable to add it.`
It should probably be in `add_control_to_dock` or removed for now.

editor/plugins/tiles/tiles_editor_plugin.cpp
Calinou: I've moved TileSet to <kbd>Alt + L</kbd> and Replication to <kbd>Alt + C</kbd>.

editor/editor_dock_manager.cpp
Calinou: I've pushed a fix for this as a second commit, which allows docks to define toggle shortcuts. These currently only work when the dock is moved to a bottom panel, but we could allow them to be used to hide/show docks in the future.

editor/editor_plugin.compat.inc
```diff
@@ -0,0 +1,47 @@
...
+Button *EditorPlugin::_add_control_to_bottom_panel_compat_88081(Control *p_control, const String &p_title) {
+	return add_control_to_bottom_panel(p_control, p_title, nullptr);
+}
+
+void EditorPlugin::_add_control_to_dock_compat_88081(DockSlot p_slot, Control *p_control) {
+	return add_control_to_dock(p_slot, p_control, nullptr);
+}
+
+void EditorPlugin::_bind_compatibility_methods() {
+	ClassDB::bind_compatibility_method(D_METHOD("add_control_to_bottom_panel", "control", "title"), &EditorPlugin::_add_control_to_bottom_panel_compat_88081);
```
kitbdev, line 44:
```suggestion
	ClassDB::bind_compatibility_method(D_METHOD("add_control_to_bottom_panel", "control", "title"), &EditorPlugin::_add_control_to_bottom_panel_compat_88081);
	ClassDB::bind_compatibility_method(D_METHOD("add_control_to_dock", "slot", "control"), &EditorPlugin::_add_control_to_dock_compat_88081);
```

misc/extension_api_validation/4.2-stable.expected
```diff
@@ -231,3 +231,11 @@ GH-89024
 Validate extension JSON: Error: Field 'classes/RichTextLabel/methods/push_meta/arguments': size changed value in new API, from 1 to 2.
 
 Added optional argument. Compatibility method registered.
+
+
+GH-88081
+--------
+Validate extension JSON: Error: Field 'classes/EditorPlugin/methods/add_control_to_bottom_panel/arguments': size changed value in new API, from 2 to 3.
```
kitbdev, line 238:
```suggestion
Validate extension JSON: Error: Field 'classes/EditorPlugin/methods/add_control_to_bottom_panel/arguments': size changed value in new API, from 2 to 3.
Validate extension JSON: Error: Field 'classes/EditorPlugin/methods/add_control_to_dock/arguments': size changed value in new API, from 2 to 3.
```

editor/editor_dock_manager.cpp
```diff
@@ -30,6 +30,7 @@
 
 #include "editor_dock_manager.h"
 
+#include "core/string/string_name.h"
```
KoBeWi, line 33: Isn't this included with Variant?

KoBeWi (approved):
Still not a fan of adding that many shortcuts, but it has some merits it seems. Though if shifting button positions are the problem, we could rework it to make them more stable (e.g. always add new buttons at the end and allow rearranging them).

Code-wise looks fine.

Calinou:
Rebased and squashed with the review suggestions incorporated.

akien-mga:
> I feel like a default shortcut set for **every single one** of those panels is overkill. Outside of the always-present 5 ones, it is also extremely unlikely that all these panels will be open at the same time
> 
> If it is later planned to use any of those Alt combinations for anything else it becomes much harder to argue with (Script Editor's `CTRL + D` is already a big example right now).

I want to echo that, I feel adding a default unique hotkey to every single panel, even some which are very contextual and only appear when a specific node is selected (like Replication, etc.) is overkill.

I would only set default hotkeys for the panels which are always present: Output, Debugger, Audio, Animation, Shader Editor, and now FileSystem since it can also be moved to the bottom panel.

Calinou:
> I would only set default hotkeys for the panels which are always present: Output, Debugger, Audio, Animation, Shader Editor, and now FileSystem since it can also be moved to the bottom panel.
>
> The rest can have shortcuts without a default binding, so users who need them can set their own preferred bindings (e.g. for TileMap / TileSet for users who do tiles-based 2D games, or Replication for users making multiplayer games with the high level multiplayer API).

Done.

akien-mga (approved):

KoBeWi:
I see the shortucts are still there 🤔

akien-mga:
Hm yes you forgot to push :P

AThousandShips (approved):

akien-mga:
Thanks!

njt1982:
Does https://docs.godotengine.org/en/4.3/tutorials/editor/default_key_mapping.html need updating?

Calinou:
> Does [docs.godotengine.org/en/4.3/tutorials/editor/default_key_mapping.html](https://docs.godotengine.org/en/4.3/tutorials/editor/default_key_mapping.html) need updating?

- Done in https://github.com/godotengine/godot-docs/pull/10046.

PR #88091, opened 2024-02-08
[TextServer] Add extra fallback step, to lookup for individual characters.

### LINKED ISSUE #9043: Allow multiple fonts to render one "glyph unit", like a glyph + diacritics/accents

A project with language localizations, where the system may not have system fonts (like web export).

When developing a game with multi-language localization for the web, the developer must define font fallbacks because [web export does not support loading system fonts](https://github.com/godotengine/godot/issues/77903#issuecomment-1640994134). When providing a text input field, a user may switch between languages while typing, which means that glyphs from multiple fonts might be typed in the same text input. When typing an Arabic accent from one font after a Latin character of another font, Godot will render two tofu glyphs instead, disrupting the user experience.

For example:
When I type the letter `a`, I see the letter `a`.
Then when I add an Arabic accent, I get two tofu glyphs: tofu_glyphs
I expected something similar to this (I typed `a` then `shift-q` using Arabic input in my browser's url bar): accent_url_bar
I think this is a missing feature despite looking like a bug.

Repro steps (tested with v4.3.dev.custom_build [d3352813e] on Ubuntu 20.03 LTS)
1. Compile the editor with `fontconfig=false` to prevent using system fonts in the editor
2. Launch the editor and create a `LineEdit` node
3. Click on the `LineEdit` node, and `Theme Overrides -> Fonts -> Font -> Choose FontVariation`
4. For the base font, choose AlbertSans-Regular.ttf
https://fonts.google.com/specimen/Albert+Sans?query=Sans&sort=alpha&subset=latin&noto.script=Latn
5. For a fallback, choose NotoSansArabic-Regular.ttf
https://fonts.google.com/noto/specimen/Noto+Sans+Arabic?query=noto&subset=arabic&noto.script=Arab
6. Ensure you can switch your system's input language between English and Arabic (this is easy to set up in Ubuntu: follow step 1 at your own risk in [Steps to reproduce](https://github.com/godotengine/godot/issues/87016))
7. Launch the app
8. Make sure your system's input language is `en`, click on the `LineEdit` and type `a`.
9. Switch your system's input language to `ar` and press `shift-q` to add an accent
10. Notice the two tofu glyphs

The only font I can find on my desktop to handle this scenario as a font fallback in Godot is DejaVu.
On my desktop outside of Godot, the shape of the original letter doesn't change with/without the accent so I don't think it is falling back to DejaVu (nor do I think it's using DejaVu by default, but hard to tell).
Also, I think it is normal to combine glyphs ([like this](https://graphicdesign.stackexchange.com/a/109918)).  So combining glyphs from separate fonts might be okay too especially considering it would be more legible than tofu glyphs.

See title

N/A

No

This is core functionality.

bruvzg:
Fixes https://github.com/godotengine/godot-proposals/issues/9043

### DIFF
--- a/modules/text_server_adv/text_server_adv.cpp
+++ b/modules/text_server_adv/text_server_adv.cpp
@@ -5814,7 +5814,7 @@ _FORCE_INLINE_ void TextServerAdvanced::_add_featuers(const Dictionary &p_source
 	}
 }
 
-void TextServerAdvanced::_shape_run(ShapedTextDataAdvanced *p_sd, int64_t p_start, int64_t p_end, hb_script_t p_script, hb_direction_t p_direction, TypedArray<RID> p_fonts, int64_t p_span, int64_t p_fb_index, int64_t p_prev_start, int64_t p_prev_end) {
+void TextServerAdvanced::_shape_run(ShapedTextDataAdvanced *p_sd, int64_t p_start, int64_t p_end, hb_script_t p_script, hb_direction_t p_direction, TypedArray<RID> p_fonts, int64_t p_span, int64_t p_fb_index, int64_t p_prev_start, int64_t p_prev_end, RID p_prev_font) {
 	RID f;
 	int fs = p_sd->spans[p_span].font_size;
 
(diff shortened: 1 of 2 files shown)


dhoverml:
I cherry-picked this to master (4e990cd7e5) and I no longer see tofu glyphs when I apply accents to non-Arabic characters.

However, this may be a separate bug, but I see that the cursor switches to the other side of the character when I'm typing the accents, but it fixes itself once I start typing Arabic characters again.

I typed `ض` (`q` on my keyboard), `#`, then added an accent (`Q` on my keyboard) into a LineEdit node.  The cursor remains on the left side until I type the accent mark (with `Q`), where it then teleports to the right side of the `#`
cursor_godot

The cursor remains on the left side when I repeat the same steps in my browser's URL bar (I added a few extra accents so you can see them)
cursor_browser

Also, in Godot, the non-Arabic glyph that I'm applying an accent to moves to the far right side of the current list of non-Arabic glyphs.
Alternate between typing `#`, `@`, `$`, and an accent.

Ok: symbol_move_0
Ok: symbol_move_1
Bad: symbol_move_2
Ok: symbol_move_3
Bad: symbol_move_4

Same sequence in a browser produces different results:
browser_symbols

bruvzg:
> However, this may be a separate bug, but I see that the cursor switches to the other side of the character when I'm typing the accents, but it fixes itself once I start typing Arabic characters again.

Pushed update to fix the direction, it should work correctly now.

dhoverml:
Wow that was fast!
I tested the latest patch, and it looks great!

The only problem I see is that it's switching the original glyph's font when it applies the accent:

This `3`  is AlbertSans-Regular.ttf
three_albertsansregular

When I apply an accent, the `3` is reshaped with NotoSansArabic-Regular.ttf
three_notosansarabicregular

The workaround is to put the Arabic font first in the list of fallbacks.
Also, I'm not too worried about this since I see some applications (like Github's text-input field) do the same.  Though some applications don't (like the Browser URL bar).

Anyway, this is great, thanks!

akien-mga (approved):

akien-mga:
Thanks!

PR #88105, opened 2024-02-08
Fix duplicating/copying `TileMap`

### LINKED ISSUE #88104: `TileMap` can't be duplicated or copied in editor
### Tested versions

- 4.3.dev [41564aaf7708b0bf594f745dd2448a54dd687cc5]

### System information

Godot v4.3.dev (d5481c648) - Windows 10.0.22631 - Vulkan (Forward+) - dedicated AMD Radeon RX 6750 XT (Advanced Micro Devices, Inc.; 31.0.23013.1023) - AMD Ryzen 9 7900X 12-Core Processor (24 Threads)

### Issue description

Attempting to copy or duplicate a `TileMap` with two or more layers fails with (might be in more cases but that's the case I replicated it with, will test with layers with data as well):
```
ERROR: Class 'TileMapLayer' or its base class cannot be instantiated.
   at: (core/object/class_db.cpp:361)
ERROR: Parameter "obj" is null.
   at: Node::_duplicate (scene/main/node.cpp:2523)
ERROR: Condition "!dup" is true. Continuing.
   at: SceneTreeDock::_tool_selected (editor/scene_tree_dock.cpp:755)
```

Will attempt to bisect the issue but it's likely directly related to the restructuring of layers, possibly:
* https://github.com/godotengine/godot/pull/87115

Edit: Can confirm that this is a regression from #87115, will do some further digging later today

### Steps to reproduce

Open the MRP and try to Copy or Duplicate the `TileMap`

[TileMapDuplicateMRP.zip](https://github.com/godotengine/godot/files/14210276/TileMapDuplicateMRP.zip)

AThousandShips:
Using `force_parent_owned` to prevent issues duplicating

Think this covers all cases

* Fixes: https://github.com/godotengine/godot/issues/88104

### DIFF
--- a/scene/2d/tile_map.cpp
+++ b/scene/2d/tile_map.cpp
@@ -387,6 +387,7 @@ void TileMap::add_layer(int p_to_pos) {
 	TileMapLayer *new_layer = memnew(TileMapLayer);
 	layers.insert(p_to_pos, new_layer);
 	add_child(new_layer);
+	new_layer->force_parent_owned();
 	new_layer->set_name(vformat("Layer%d", p_to_pos));
 	move_child(new_layer, p_to_pos);
 	for (uint32_t i = 0; i < layers.size(); i++) {
@@ -719,6 +720,7 @@ bool TileMap::_set(const StringName &p_name, const Variant &p_value) {
 			if (layers.size() == 0) {
 				TileMapLayer *new_layer = memnew(TileMapLayer);
 				add_child(new_layer);
+				new_layer->force_parent_owned();
 				new_layer->set_name("Layer0");
 				new_layer->set_layer_index_in_tile_map_node(0);
 				layers.push_back(new_layer);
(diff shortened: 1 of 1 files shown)


scene/register_scene_types.cpp
```diff
@@ -783,6 +784,7 @@ void register_scene_types() {
 	GDREGISTER_CLASS(TileMapPattern);
 	GDREGISTER_CLASS(TileData);
 	GDREGISTER_CLASS(TileMap);
+	GDREGISTER_INTERNAL_CLASS(TileMapLayer);
```
AThousandShips, line 787: Internal as it shouldn't be registered for normal purposes

AThousandShips:
Something with the test project needs tweaking or some exception case needs managing, will look in a bit

AThousandShips:
Appears the internal node is still accessible to scripting and therefore causes some problems, will investigate

AThousandShips:
Will need some tweaking possibly to handle the ability to instance these types, but unsure, the alternative would be to deal somehow with the duplication of the child nodes in a way that prevents breaking, but unsure how that would work best

AThousandShips:
It seems this is relatively trivial, will look at a different fix leveraging `is_owned_by_parent`

AThousandShips:
I've identified a few other places where `force_parent_owned` might be needed and will look at a PR for that too

Mickeon (approved):
I agree this solution is better. In the future, though, if/when layers are exposed, these snippets should be removed, otherwise you get into a nasty bug. Maybe adding a comment on the side can help?

AThousandShips:
They shouldn't be removed even then, the layers would be doubled because the recreation of the data from the properties

akien-mga:
Thanks!

AThousandShips:
Thank *you*!

PR #88084, opened 2024-02-08
Use indeterminate progressbars for editor downloads (export templates & assets)

ckaiser:
Now that we have [indeterminate progress bars](https://github.com/godotengine/godot/pull/87318), I've implemented their use for the most obvious case: downloads.
There might be other places that could use them but it'd involve changes to `ProgressDialog` and tasks, haven't looked through that as of yet.

Before & After:

https://github.com/godotengine/godot/assets/138269/7643d5d1-dd94-441d-8a95-30d5211db6fb

https://github.com/godotengine/godot/assets/138269/1184a474-176f-4311-b7ab-0de7bd8de632

---

https://github.com/godotengine/godot/assets/138269/5eab6a2d-7eef-43d3-8613-63170d2ea826

https://github.com/godotengine/godot/assets/138269/ddec0abd-18a9-4813-a286-e7f791e2d575

(I was connected to a slow VPN for testing I swear I'm not on dial-up or something 😅)

PR #88106, opened 2024-02-08
Show update spinner by default in `dev_build=yes` editor builds

Calinou:
This ensures contributors can see when something forces the editor to redraw constantly.

The existing boolean `true` value will be casted to `1` in the setting, so it'll switch to **Enabled** automatically if the setting was previously enabled. If the setting is disabled, it'll default to **Auto** (enabled on dev builds, disabled otherwise).

- See https://github.com/godotengine/godot/pull/88084.

### DIFF
--- a/editor/editor_node.cpp
+++ b/editor/editor_node.cpp
@@ -610,7 +610,7 @@ void EditorNode::_notification(int p_what) {
 				update_spinner_step_frame = frame + 1;
 
 				// Update the icon itself only when the spinner is visible.
-				if (EDITOR_GET("interface/editor/show_update_spinner")) {
+				if (_should_display_update_spinner()) {
 					update_spinner->set_icon(theme->get_icon("Progress" + itos(update_spinner_step + 1), EditorStringName(EditorIcons)));
 				}
 			}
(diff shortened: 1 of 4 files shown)


akien-mga (approved):
Sounds like a good idea to me.

editor/editor_node.cpp
```diff
@@ -6476,7 +6486,6 @@ EditorNode::EditorNode() {
 	register_exporters();
 
 	EDITOR_DEF("interface/editor/save_on_focus_loss", false);
-	EDITOR_DEF("interface/editor/show_update_spinner", false);
```
akien-mga, line 6479: Moving this to `editor_settings.cpp` seems to have added it properly to the docs.

So it seems all these properties may be missing from the docs and should be moved to (in a separate PR).

Mickeon:
<sub>This was my suggestion so yeah of course I do agree</sub>

Mickeon (approved):
<sub>This was my suggestion so yeah of course I do agree</sub>

editor/editor_settings.cpp
```diff
@@ -455,6 +455,11 @@ void EditorSettings::_load_defaults(Ref<ConfigFile> p_extra_config) {
 #else
 	EDITOR_SETTING(Variant::INT, PROPERTY_HINT_ENUM, "interface/editor/show_internal_errors_in_toast_notifications", 0, "Auto (Disabled),Enabled,Disabled")
 #endif
+#ifdef DEV_ENABLED
+	EDITOR_SETTING(Variant::INT, PROPERTY_HINT_ENUM, "interface/editor/show_update_spinner", 0, "Auto (Enabled),Enabled,Disabled")
+#else
+	EDITOR_SETTING(Variant::INT, PROPERTY_HINT_ENUM, "interface/editor/show_update_spinner", 0, "Auto (Disabled),Enabled,Disabled")
+#endif
```
Mickeon, line 462: Would it be nicer to merge the DEV_ENABLED code block with the one above this snippet?

PR #88107, opened 2024-02-08
Add particle y-billboard option to standard shader

### LINKED ISSUE #5591: Add a Preserve Orientation billboard property

Various VFX

There is no adequate billboard mode to orient particles along their velocity while preserving the billboard

Add a billboard mode that projects the y axis on screen coordinates and uses that as a billboard angle

See above

Need to re-implement by hand every time.

It's a common usecase and would improve usability.

Arnklit:
This adds a particle y-billboard option to Standard Material. 

image

When used with Align Y on the particles it allows you to have particles that align themselves with velocity, but face the camera.

image
image

Note I'm not sure if this should be added since you can also achieve this with the "Particles -> Drawing -> Transform Align -> Z Billboard + Y to Velocity". But these options don't work with stuff like Particle Animation. 

In general I'm not sure why there are billboarding settings both in the Particles and in the Material like this.

- *Production edit: This closes https://github.com/godotengine/godot-proposals/issues/5591.*


Calinou (approved):
Tested locally, it works as expected. This is a great addition to have for VFX like water fountains, sparks and the like :slightly_smiling_face: 

I've looked at the code and it seems good to me.

**Testing project:** [test_pr_88107.zip](https://github.com/godotengine/godot/files/14220975/test_pr_88107.zip)

https://github.com/godotengine/godot/assets/180032/9cd69613-1df2-4892-9377-e7843f5265d8

QbieShay:
This needs to be discussed in a rendering meeting but we should try to understand where to put bollboarding options for particles. I'm not sure standard material makes sense.

Calinou:
> This needs to be discussed in a rendering meeting but we should try to understand where to put bollboarding options for particles. I'm not sure standard material makes sense.

The reason we had Particle Billboard in the material is that it allows for configuring flipbook animation, so if we remove these options, we have to make particle flipbook animation depend on another checkbox within the material.

QbieShay:
More context: https://github.com/godotengine/godot-proposals/issues/8367

scene/resources/material.cpp
```diff
@@ -3741,7 +3759,7 @@ void BaseMaterial3D::_bind_methods() {
 	ADD_PROPERTYI(PropertyInfo(Variant::BOOL, "shadow_to_opacity"), "set_flag", "get_flag", FLAG_USE_SHADOW_TO_OPACITY);
 
 	ADD_GROUP("Billboard", "billboard_");
-	ADD_PROPERTY(PropertyInfo(Variant::INT, "billboard_mode", PROPERTY_HINT_ENUM, "Disabled,Enabled,Y-Billboard,Particle Billboard"), "set_billboard_mode", "get_billboard_mode");
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "billboard_mode", PROPERTY_HINT_ENUM, "Disabled,Enabled,Y-Billboard,Particle Billboard, Particle Y-Billboard"), "set_billboard_mode", "get_billboard_mode");
```
AThousandShips:
```suggestion
	ADD_PROPERTY(PropertyInfo(Variant::INT, "billboard_mode", PROPERTY_HINT_ENUM, "Disabled,Enabled,Y-Billboard,Particle Billboard,Particle Y-Billboard"), "set_billboard_mode", "get_billboard_mode");
```

Arnklit: Thanks. Fixed.

PR #88111, opened 2024-02-08
Expose scene unique id functionality in Resource

### LINKED ISSUE #9057: Expose scene unique id functionality in Resource

I am making a Multiset plugin for collaborative scene editing, and I need to keep track and update resources over the network across different Godot editor instances.

There is no current way to set the embedded resource ID before or when saving a resource. Also, the only way to get the ID is after saving the scene (id generated during save and is based on time!) and splitting the resource path to get the ID.

Not being able to set the scene_unique_id in Resource's is a massive road blocker when creating and keeping track of resources over the network. 

When the host creates a new embedded resource in a scene, I cannot keep track of it at all across clients, because the scene unique id is created at save time, so for each client the resource will actually have a different scene unique id, or not one at all. This makes it practically impossible to keep all the changes synced, and will mess up git history, as everyone's will be slightly different.

Expose scene unique id functionality in Resource

I made a pull request https://github.com/godotengine/godot/pull/88111

Cannot be worked around its core.

Its Core.

Ryan-000:
There is no current way to set the embedded resource ID before or when saving a resource. Also, the only way to get the ID is after saving the scene (id generated during save and is based on time!) and splitting the resource path to get the ID.

In my situation, I am making a Multiset plugin for collaborative scene editing, and this is a massive road blocker when creating and keeping track of resources over the network. When the host creates a new embedded resource in a scene, I cannot keep track of it at all across clients, because the scene unique id is created at save time, so for each client the resource will actually have a different scene unique id, or not one at all. Which makes it practically impossible to keep all the changes synced, and will mess up git history as everyone's will be slightly different.

I tested this, and it works well, if there are duplicate ids in a scene (but on different resources) at save time, only the first resource will get the id, and the other ones will have regenerated ids.

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/9057*

### DIFF
--- a/core/io/resource.cpp
+++ b/core/io/resource.cpp
@@ -130,6 +130,16 @@ String Resource::generate_scene_unique_id() {
 }
 
 void Resource::set_scene_unique_id(const String &p_id) {
+	bool is_valid = true;
+	for (int i = 0; i < p_id.length(); i++) {
+		if (!is_ascii_identifier_char(p_id[i])) {
+			is_valid = false;
+			scene_unique_id = Resource::generate_scene_unique_id();
+			break;
+		}
+	}
+
+	ERR_FAIL_COND_MSG(!is_valid, "The scene unique ID must contain only letters, numbers, and underscores.");
 	scene_unique_id = p_id;
 }
 
@@ -532,6 +542,10 @@ void Resource::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_local_scene"), &Resource::get_local_scene);
 	ClassDB::bind_method(D_METHOD("setup_local_to_scene"), &Resource::setup_local_to_scene);
 
+	ClassDB::bind_static_method("Resource", D_METHOD("generate_scene_unique_id"), &Resource::generate_scene_unique_id);
+	ClassDB::bind_method(D_METHOD("set_scene_unique_id", "id"), &Resource::set_scene_unique_id);
+	ClassDB::bind_method(D_METHOD("get_scene_unique_id"), &Resource::get_scene_unique_id);
+
 	ClassDB::bind_method(D_METHOD("emit_changed"), &Resource::emit_changed);
 
 	ClassDB::bind_method(D_METHOD("duplicate", "subresources"), &Resource::duplicate, DEFVAL(false));
@@ -542,6 +556,7 @@ void Resource::_bind_methods() {
 	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "resource_local_to_scene"), "set_local_to_scene", "is_local_to_scene");
 	ADD_PROPERTY(PropertyInfo(Variant::STRING, "resource_path", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_EDITOR), "set_path", "get_path");
 	ADD_PROPERTY(PropertyInfo(Variant::STRING, "resource_name"), "set_name", "get_name");
+	ADD_PROPERTY(PropertyInfo(Variant::STRING, "resource_scene_unique_id", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_NONE), "set_scene_unique_id", "get_scene_unique_id");
 
 	MethodInfo get_rid_bind("_get_rid");
 	get_rid_bind.return_val.type = Variant::RID;
--- a/doc/classes/Resource.xml
+++ b/doc/classes/Resource.xml
@@ -59,6 +59,12 @@
 				[/codeblock]
 			</description>
 		</method>
+		<method name="generate_scene_unique_id" qualifiers="static">
+			<return type="String" />
+			<description>
+				Generates a unique identifier for a resource to be contained inside a [PackedScene], based on the current date, time, and a random value. The returned string is only composed of letters ([code]a[/code] to [code]y[/code]) and numbers ([code]0[/code] to [code]8[/code]). See also [member resource_scene_unique_id].
+			</description>
+		</method>
 		<method name="get_local_scene" qualifiers="const">
 			<return type="Node" />
 			<description>
@@ -97,6 +103,12 @@
 		<member name="resource_path" type="String" setter="set_path" getter="get_path" default="&quot;&quot;">
 			The unique path to this resource. If it has been saved to disk, the value will be its filepath. If the resource is exclusively contained within a scene, the value will be the [PackedScene]'s filepath, followed by a unique identifier.
 			[b]Note:[/b] Setting this property manually may fail if a resource with the same path has already been previously loaded. If necessary, use [method take_over_path].
+		</member>
+		<member name="resource_scene_unique_id" type="String" setter="set_scene_unique_id" getter="get_scene_unique_id">
+			An unique identifier relative to the this resource's scene. If left empty, the ID is automatically generated when this resource is saved inside a [PackedScene]. If the resource is not inside a scene, this property is empty by default.
+			[b]Note:[/b] When the [PackedScene] is saved, if multiple resources in the same scene use the same ID, only the earliest resource in the scene hierarchy keeps the original ID. The other resources are assigned new IDs from [method generate_scene_unique_id].
+			[b]Note:[/b] Setting this property does not emit the [signal changed] signal.
+			[b]Warning:[/b] When setting, the ID must only consist of letters, numbers, and underscores. Otherwise, it will fail and default to a randomly generated ID.
 		</member>
 	</members>
 	<signals>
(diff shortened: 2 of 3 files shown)


AThousandShips:
Please open a [proposal](https://github.com/godotengine/godot-proposals/issues) to track the support and details of this feature :)

fire:
Is there an example I can use to check this is working?

fire:
While I'm reviewing and gather reviewers can you apply the documentation changes?

reduz (approved):

reduz:
Makes sense, but please add the documentation in doc/classes/Resource.xml
you need to run Godot with `--doctool .` argument so the xmls get updated.

akien-mga:
While adding documentation, please also squash the commits. See [PR workflow](https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#the-interactive-rebase) for instructions.

Mickeon (commented):
Skeptical myself on exposing this full in the Inspector. I have many thoughts but to make it brief:
It would be a property not many would modify among others that are arguably more useful. This adds to clutter, bloat, noise, however one may want to call it. It may sound like an exaggeration but... yeah.

But if it has to be exposed it needs better documentation.

Mickeon:
If this were a property defined in a script it would usually be assigned **PROPERTY_USAGE_DEFAULT**, so that it is not exposed to the Inspector and saved at all. For built-in properties I'm struggling to find any that actually do this, and this may be entirely by design.

In fact, for these kinds of properties what other Objects do is only keeping the setter and getter.
For example, all of **Node**'s `set_process` and `is_processing` methods. They could be interpreted as setting a `process` property that is not actually exposed.

doc/classes/Resource.xml
```diff
@@ -59,6 +59,12 @@
 				[/codeblock]
 			</description>
 		</method>
+		<method name="generate_scene_unique_id" qualifiers="static">
+			<return type="String" />
+			<description>
+				Generates a unique identifier for a scene embedded resource by combining the current date, time, and a random number. This combination is then processed to produce a simple, readable string made up of letters (a-z) and numbers (0-9).
```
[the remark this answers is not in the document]
Mickeon: No, no, no. No worries. This is not asking for it to be fixed. In fact, the last PR that tried to fix this was rejected, whatever the exact reason was.

fire:
You need to run `godot --doctool` as mentioned earlier so that the github actions pass and we can merge.

Mickeon:
> If you asked me yesterday what git squash is, I'd be thinking about butternut squash soup.

This is our fault actually. We should have linked you to the [Pull request workflow](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase) page sooner.

doc/classes/EditorSettings.xml
```diff
@@ -756,6 +756,8 @@
 		<member name="project_manager/sorting_order" type="int" setter="" getter="">
 			The sorting order to use in the project manager. When changing the sorting order in the project manager, this setting is set permanently in the editor settings.
 		</member>
+		<member name="resource_scene_unique_id" type="String" setter="set_scene_unique_id" getter="get_scene_unique_id">
+		</member>
```
Mickeon, line 760: This is quite odd. I am to assume that the doctools generated this on its own, but it should not happen...

core/io/resource.cpp
```diff
@@ -124,6 +124,9 @@ String Resource::generate_scene_unique_id() {
 	return id;
 }
 
+/// warning: there is no validation here
+/// use is_ascii_identifier_char() to validate before passing it in
+/// example: if you pass in a quote, it will corrupt the scene when saving
```
Mickeon, line 129: I must mention first that all new comments should follow the [comment style guide](https://docs.godotengine.org/en/stable/contributing/development/code_style_guidelines.html#comment-style-guide), ideally. So it should look more like this.
```suggestion
// Warning: there is no validation here.
// Use is_ascii_identifier_char() to validate before passing it in.
// Example: if you pass in a quote, it will corrupt the scene when saving.
```

But more importantly to me, the comment does summarise another reason as to why I was skeptical on exposing this. The outside developer should be warned about this:
- It should be documented in the class reference.
- It should not even be possible to do this. In this situation, a warning should appear, the invalid characters should be replaced or the setter should fail outright. Whatever is the most desired.

This would mean making a distinction between the `set_scene_unique_id` used internally (since the additional checks would be a waste), and the exposed `set_scene_unique_id`.

I believe that in this situation a lot of classes prefix the internal method with an underscore.

doc/classes/EditorSettings.xml
AThousandShips: You need to add a `_` prefix to prevent it from being documented I think, usage doesn't control that

AThousandShips: This is because it has its own processing AFAIK

Mickeon: Me no likey. Sounds like a separate bug to tackle.

AThousandShips (changes requested):
See above

doc/classes/EditorSettings.xml
AThousandShips: It turns out it's an explicit list that needs to be updated

AThousandShips: You need to remove this

AThousandShips (approved):

AThousandShips:
You can squash at the same time or just update the original commit :)

AThousandShips:
You use `fixup` by them, if you delete them you lose all the information

AThousandShips:
You need to push with `--force`, you added merge commits instead

AThousandShips:
Please do the following:
```
git reset --hard 1887211
git push --force
```

AThousandShips:
There we go :)

In the future remember to create your changes on a separate branch to avoid a lot of these problems, and to use rebase to update your branch and not merges

AThousandShips:
No worries!

core/io/resource.cpp
```diff
@@ -124,6 +124,9 @@ String Resource::generate_scene_unique_id() {
 	return id;
 }
 
+// Warning: there is no validation here.
+// Use is_ascii_identifier_char() to validate before passing it in.
+// Example: if you pass in a quote, it will corrupt the scene when saving.
 void Resource::set_scene_unique_id(const String &p_id) {
 	scene_unique_id = p_id;
```
fire, line 131: Is there an escape we can do that will let us use arbitrary strings but avoid breaking the scene?

Mickeon: I do reiterate my original concern in https://github.com/godotengine/godot/pull/88111#discussion_r1485169534. That is, the method exposed to outside developers should not be able to brick scenes this way (see comment for more).

core/io/resource.cpp
```diff
@@ -124,7 +124,18 @@ String Resource::generate_scene_unique_id() {
 	return id;
 }
 
+
```
AThousandShips, line 127:
```suggestion
```

core/io/resource.cpp
```diff
@@ -125,6 +125,16 @@ String Resource::generate_scene_unique_id() {
...
+	bool is_valid = true;
+	for (int i = 0; i < p_id.length(); i++) {
+		if (!is_ascii_identifier_char(p_id[i])) {
+			is_valid = false;
+			scene_unique_id = Resource::generate_scene_unique_id();
+			break;
+		}
+	}
+
+	ERR_FAIL_COND_MSG(!is_valid, "The scene unique ID must contain only letters, numbers and underscores.");
```
AThousandShips, line 137:
```suggestion
	ERR_FAIL_COND_MSG(!is_valid, "The scene unique ID must contain only letters, numbers, and underscores.");
```
Please use [Oxford comma](https://docs.godotengine.org/en/latest/contributing/documentation/docs_writing_guidelines.html#use-the-oxford-comma-to-enumerate-anything).

Mickeon (approved):
Thank you for your patience. I still think that there should be a distinction between internal and exposed method, because the additional check internally is extremely redundant. But this can be done later in a separate PR, this one should be fine after the above issue is fixed and (and the commits are squashed again).

AThousandShips:
Please remove the body of the commit message, it's just a lot of notes on what was changed and just clutters :)

fire (approved):

akien-mga:
> Please remove the body of the commit message, it's just a lot of notes on what was changed and just clutters :)

Force pushed an amend to do this, and rebased on latest `master`.

akien-mga:
Thanks!

PR #80585, opened 2023-08-13
Core: Add `allow_objects`/`full_objects` parameter to text serialization

dalexeev:
This PR adds the `allow_objects`/`full_objects` parameter to the `VariantParser`/`VariantWriter` internal API that is used for the `str_to_var()`/`var_to_str()` functions and `ConfigFile` methods.

I think security reason enough to break compatibility in this case. Let me know if you think compatibility is more important and we should provide users with an insecure API by default, even if it's inconsistent with binary serialization.

* Closes #80562
* See also godotengine/godot-proposals#4925.

_Production edit: closes godotengine/godot-roadmap#69_

PR #88112, opened 2024-02-08
Add Security Warnings to str_to_var and var_to_str Documentation

Ryan-000:
I'm adding warnings to str_to_var and var_to_str in the docs because I found out "str_to_var" can run code without warning when I checked the source code and tested it myself.

This wasn't clear before. The warnings are to help other developers know about this risk.


AThousandShips:
See:
* https://github.com/godotengine/godot/pull/80585

Which solves this issue, I'd say this is unnecessary with that fix

AThousandShips:
You need to rebase your branch see [here](https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#rebasing-onto-another-branch)

PR #84824, opened 2023-11-13
Fix duplicating sub-scene may get two copies of internal node

Rindbee:
Previously, internal node might be mistaken for `hidden_root` and be duplicated again. Exclude those internal nodes to avoid this case.

Fix #84803.
*Bugsquad edit:* Fixes #74406.

PR #88114, opened 2024-02-08
Prevent some internal nodes being duplicated in Controls

AThousandShips:
Caused by duplication of on-demand created children, for example:
```gdscript
var cpb := ColorPickerButton.new()
print(cpb.get_child_count(true))
print(cpb.get_popup())
print(cpb.get_child_count(true))

var cpb2 := cpb.duplicate()
print(cpb2.get_child_count(true))
print(cpb2.get_popup())
print(cpb2.get_child_count(true))
```
Will print:
```
0
@PopupPanel@...:<PopupPanel#...>
1
1
@PopupPanel@...:<PopupPanel#...>
2
```
This fixes these

There are probably some other potential cases but unsure how to apply each, but keeping these specific menu cases for now

### DIFF
--- a/scene/gui/color_picker.cpp
+++ b/scene/gui/color_picker.cpp
@@ -1741,6 +1741,7 @@ void ColorPicker::_pick_button_pressed() {
 		picker_preview_color->add_theme_style_override(SceneStringName(panel), picker_preview_style_box_color);
 
 		add_child(picker_window, false, INTERNAL_MODE_FRONT);
+		picker_window->force_parent_owned();
 	}
 	set_process_internal(true);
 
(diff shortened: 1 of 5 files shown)


KoBeWi:
I think #84824 fixes the same issue, no?

AThousandShips:
It does not, I tried it with at least some cases, will confirm but that is for instanced nodes I believe

AThousandShips:
Some internal nodes might not be automatically added though and might be data independent, unsure, but it'd also break compat to change it for internal nodes IMO as scripts may use it

Note that if they're added in the constructor they are already automatically excluded

Rindbee:
Well, you are right, the script can manage the entire tree, possibly adding additional cases.

KoBeWi (dismissed):

AThousandShips:
Haven't looked for any new cases of this since this was opened, there might be some cases will try and search for them

KoBeWi (approved):

akien-mga:
> Haven't looked for any new cases of this since this was opened, there might be some cases will try and search for them

There's at least https://github.com/godotengine/godot/issues/92880 which would be fixed by #89442, unless the internal node in GraphEdit that triggers it is one of those covered here.

Repiteo:
Thanks!

PR #88134, opened 2024-02-09
SCons: Use r-strings for wayland-scanner builder command

akien-mga:
I wonder if there's a static check we could do to spot this without this trial and error?


akien-mga:
It wasn't a problem in the past, but Python 3.12 made this a warning (and I think 3.13 will stop supporting this syntax and error out).

akien-mga:
So far it's only been needed when there's invalid escape codes, like `\.` here. `<` and `>` in these other examples seem fine. But yeah I can apply it to all the `wayland-scanner` invocations for consistency.

PR #88129, opened 2024-02-09
[Windows] Fix windows `is_path_invalid`, and apply it to directory creation.

bruvzg:
Fixes https://github.com/godotengine/godot/issues/88120 and applies validity checks to the directory creation.

PR #88144, opened 2024-02-09
Fix invalid file path handling in Windows when there is dot in the file name

GNSS-Stylist:
This basically re-adds the dot-removal removed by the previous commit (PR https://github.com/godotengine/godot/pull/88129). It looks like get_basename() (that was meant to replace the old "dot-removal" part?) doesn't actually remove the dot from the filename. Therefore "con.png" etc. becomes "con." and this still froze the editor.

Not sure if the dot-removal should be actually done in get_basename() instead(?) If so, feel free to discard this PR.

@bruvzg

### DIFF
--- a/drivers/windows/file_access_windows.cpp
+++ b/drivers/windows/file_access_windows.cpp
@@ -60,7 +60,12 @@ void FileAccessWindows::check_errors() const {
 
 bool FileAccessWindows::is_path_invalid(const String &p_path) {
 	// Check for invalid operating system file.
-	String fname = p_path.get_file().get_basename().to_lower();
+	String fname = p_path.get_file().to_lower();
+
+	int dot = fname.find(".");
+	if (dot != -1) {
+		fname = fname.substr(0, dot);
+	}
 	return invalid_files.has(fname);
 }
 


AThousandShips:
This is exactly what the `get_basename` method does:
https://github.com/godotengine/godot/blob/94dbf69f5d6b7d2fd9561692df2e71557607fddc/core/string/ustring.cpp#L4800-L4807

Are you sure there's nothing else missing here? This shouldn't be needed

bruvzg (commented):
>Therefore "con.png" etc. becomes "con." and this still froze the editor.

I can't reproduce it, it's correctly returning "con" in this case.

~In addition to being redundant, this logic is wrong since it's searching for the dot from the left instead of doing it from the right (like `get_basename` do), so it will reject a completely valid `con.con.con` file for example.~

bruvzg:
> In addition to being redundant, this logic is wrong since it's searching for the dot from the left instead of doing it from the right (like get_basename do), so it will reject a completely valid con.con.con file for example.

Actually, according to the Windows docs, `con.con.con` should not be valid, so searching from the lest might be more correct. But names like this seems to work in practice (at least on Windows 11).

> Also avoid these names followed immediately by an extension; for example, NUL.txt and NUL.tar.gz are both equivalent to NUL.

drivers/windows/file_access_windows.cpp
```diff
@@ -61,6 +61,11 @@ void FileAccessWindows::check_errors() const {
 bool FileAccessWindows::is_path_invalid(const String &p_path) {
 	// Check for invalid operating system file.
 	String fname = p_path.get_file().get_basename().to_lower();
+
+	int dot = fname.find(".");
+	if (dot != -1) {
+		fname = fname.substr(0, dot);
+	}
```
bruvzg, line 68:
```suggestion
	String fname = p_path.get_file().to_lower();

	int dot = fname.find(".");
	if (dot != -1) {
		fname = fname.substr(0, dot);
	}
```

So I guess something like this might be more correct, and avoid redundant search.

GNSS-Stylist:
So you commented faster than me, but I will post this anyway... :)

Did some more testing with debug-prints:
```
bool FileAccessWindows::is_path_invalid(const String &p_path) {
	// Check for invalid operating system file.

	WARN_PRINT("p_path: " + p_path);

	String fname = p_path.get_file().get_basename().to_lower();

	WARN_PRINT("fname after get_basename(): " + fname);

	int dot = fname.find(".");
	if (dot != -1) {
		fname = fname.substr(0, dot);
	}

	WARN_PRINT("fname after manual dot removal: " + fname);

	return invalid_files.has(fname);
}
```

Output with BBCode "[img]con.[/img]"
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con..remap
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con.
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con..import
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con.
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con.
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  Resource file not found: res://con. (expected type: Texture2D)

BBCode "[img]res://con.png[/img]":
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con.png.remap
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con.png
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con.png.import
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con.png
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con.png
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  Resource file not found: res://con.png (expected type: Texture2D)

BBCode "[img]res://con.png.tmp[/img]":
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con.png.tmp.remap
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con.png.tmp
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con.png.tmp.import
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con.png.tmp
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con.png.tmp
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con.png
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  Resource file not found: res://con.png.tmp (expected type: Texture2D)

BBCode "[img]con.png[/img]":
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con.png.remap
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con.png
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con.png.import
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con.png
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  drivers\windows\file_access_windows.cpp:64 - p_path: res://con.png
  drivers\windows\file_access_windows.cpp:68 - fname after get_basename(): con
  drivers\windows\file_access_windows.cpp:75 - fname after manual dot removal: con
  Resource file not found: res://con.png (expected type: Texture2D)

Those ".remap" and ".import" extensions confure me a bit. And I was wrong about ""con.png" etc. becomes "con."". Not sure why I came to that conclusion...

Reverted my changes and tried with some strings (same as above + "bare" con):
"[img]res://con.[/img]" -> Freeze
"[img]res://con.png[/img]" -> Freeze
"[img]res://con.png.tmp[/img]" -> Freeze
"[img]con.png[/img]" -> Freeze
"[img]con[/img]" -> Works

None of these freeze the editor with this PR. PC has windows 10.

bruvzg (approved):

akien-mga:

akien-mga:
Thanks!

Issue #10213, opened 2017-08-10
OptionButton selected value always set to 0 when run


earlroxas13:
OS: Windows 8.1
2.1.3 (Found it here first)
3 alpha (Also found here, used in my screenshot)

**Issue description:**
So, I set the selected value of the option button (say it is 2), and when run the value reset to 0, the value should be 2 right? But when I run the project the value become 0 and the item showed is the 0th item text.

**Steps to reproduce:**
Add OptionButton to scene,
Add some items,
Change the selected value,
And run the project.

**Link to minimal example project:**
Here is the screenshot, i use the 3.0 version, the same thing happen in 2.1.3 too.
In Editor:
![optionerroreditor](https://user-images.githubusercontent.com/8316851/29151458-7dffb1be-7dab-11e7-832a-c8a787ed0cbd.JPG)
When Run:
![optionerrorrun](https://user-images.githubusercontent.com/8316851/29151462-86b1386e-7dab-11e7-91f1-45c9179614ed.JPG)

Noshyaar:
Can confirm that it's present in recent master df5d01f

Scene file contains correct index.
```
[node name="OptionButton" type="OptionButton" parent="."]
selected = 2
```

Workaround: `select(idx)` in `_ready`

bojidar-bg:
Potentially, related to the fact that `popup` might be still empty when `selected` is being set at https://github.com/godotengine/godot/blob/3121b3a4f4c01744184e952d79f9a56e01bdba41/scene/gui/option_button.cpp#L210-L215

Alexander-Alekseev:
@earlroxas13,
Could you please provide a few more details: what do you mean by 'Add some items'? How do you add them?
According to your screenshot, your OptionButton control doesn't contain any child items. If you add them in a code, then the described behaviour is predictable: you can't set selected item in GUI if you don't have items at all. In this case "workaround" provided by @ghost is a solution.

earlroxas13:
@Alexander-Alekseev 
Yes, the workaround is working. But it mean that you have to attach script whenever you use OptionButton, it's not suppose to be that way right?

I tested on the 3.0 stable and this issue still there.
![ob](https://user-images.githubusercontent.com/8316851/36291943-f63046f4-1300-11e8-8345-f8e3dd251f8d.JPG)
[...]

PR #88145, opened 2024-02-09
Tweak property order in the inspector for OptionButton

### LINKED ISSUE #88141: `OptionButton` has a foldable `Items` section above properties that do not belong to it
### Tested versions

v4.3.dev.custom_build [94dbf69f5]

### System information

linux

### Issue description

image

notice the upper section that folds but doesn't contain the other properties. This section should be moved below the sectionless properties.

It used to be the same in `AudioStreamRandomiser`, but it [was fixed](https://github.com/godotengine/godot/pull/82411).

### Steps to reproduce

add an OptionButton

any

Ratamacue9112:
Fixes #88141. Moves the `items` property to below other properties.

### DIFF
--- a/scene/gui/option_button.cpp
+++ b/scene/gui/option_button.cpp
@@ -339,6 +339,13 @@ void OptionButton::set_item_count(int p_count) {
 		}
 	}
 
+	if (!initialized) {
+		if (queued_current != current) {
+			current = queued_current;
+		}
+		initialized = true;
+	}
+
 	_refresh_size_cache();
 	notify_property_list_changed();
 }
@@ -435,7 +442,13 @@ void OptionButton::_select(int p_which, bool p_emit) {
 }
 
 void OptionButton::_select_int(int p_which) {
-	if (p_which < NONE_SELECTED || p_which >= popup->get_item_count()) {
+	if (p_which < NONE_SELECTED) {
+		return;
+	}
+	if (p_which >= popup->get_item_count()) {
+		if (!initialized) {
+			queued_current = p_which;
+		}
 		return;
 	}
 	_select(p_which, false);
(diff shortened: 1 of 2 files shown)


Ratamacue9112:
There was a comment that said:
`// "selected" property must come after "item_count", otherwise GH-10213 occurs.`
However when I reordered it, no errors were thrown. I'm guessing this was fixed at some point and the comment was never removed.

AThousandShips:
There isn't an `item_count` any more so it's irrelevant

AThousandShips:
But it's not in the declarations here so it can't be reordered before or after

KoBeWi:
I think `ADD_ARRAY_COUNT` does the declaration.

AThousandShips:
That's true!

Note though @Ratamacue9112 that this isn't an error being thrown, so make sure to check the issue report to make sure it doesn't reappear:
* https://github.com/godotengine/godot/issues/10213

AThousandShips (changes requested):
This does indeed break the class, and `selected` breaks with this change, it works fine without this change

Ratamacue9112:
Ah sorry, didn't notice that. I'm trying to work out a fix now, any suggestions?

AThousandShips:
No, it'd require quite a bit of reorganization I think, you can't change any of the names of the properties, and we can't delay initializing this property, so I'm not sure how it would be possible to fix this without reworking quite a bit

KoBeWi:
The initializing *can* be delayed. You can add a new boolean e.g. `initialized` and if `selected` is out of index and `initialized` is false, store it into e.g. a new `queued_selected` variable. Then in `NOTIFICATION_POSTINTIALIZE` check if `queued_selected` was changed and if it was, assign it to `selected`. Also set `initialized` to `true`.

AThousandShips:
It's true we can delay the rest of the processing, what I meant was that we can't directly delay it with some property flag

Ratamacue9112:
Where should I put the check for `selected` being out of index? I've tried it in a few places, which don't seem to work, but I might just be doing it wrong.

KoBeWi:
It should be in the setter.

Ratamacue9112:
I've fixed it now. It only worked when I put check if `queued_current` was changed in the `set_item_count` method, instead of in `NOTIFICATION_POSTINITIALIZE`. I don't know if that affects anything else.

scene/gui/option_button.cpp
```diff
@@ -339,6 +339,13 @@ void OptionButton::set_item_count(int p_count) {
 		}
 	}
 
+	if (!initialized) {
+		if (queued_current != current) {
+			current = queued_current;
+		}
+		initialized = true;
```
AThousandShips, line 346: This should be set in `_select_int` IMO

scene/gui/option_button.cpp
```diff
@@ -435,9 +442,17 @@ void OptionButton::_select(int p_which, bool p_emit) {
 }
 
 void OptionButton::_select_int(int p_which) {
-	if (p_which < NONE_SELECTED || p_which >= popup->get_item_count()) {
+	if (p_which < NONE_SELECTED) {
 		return;
 	}
+	if (p_which >= popup->get_item_count()) {
+		if (initialized) {
+			ERR_FAIL_INDEX(current, popup->get_item_count());
```
AThousandShips, line 450: Why this addition? It checks the previously selected value?

scene/gui/option_button.cpp
```diff
@@ -435,9 +435,26 @@ void OptionButton::_select(int p_which, bool p_emit) {
...
+			queued_current = p_which;
+		}
+	}
+
+	if (!initialized) {
+		if (queued_current != current) {
+			current = queued_current;
+		}
+		initialized = true;
+	}
```
AThousandShips, line 456: My bad I deleted that suggestion as I got confused about the order, sorry to confuse there didn't realize you saw it and responded

Ratamacue9112: It does still work so should I just keep it like that?

AThousandShips: Restore it to how it was, it was correct as far as I could tell

AThousandShips:
I will test this out later today :)

AThousandShips (approved):
Tested and works correctly, thank you!

akien-mga:
Thanks!

PR #88159, opened 2024-02-09
Remove 3.x references in the pull request template

adamscott:
As 98% of the PRs target 4.x (don't quote me for this) and as these 3.x reference where something critical when we were developing 3.x and 4.x (on `master`) fully in parallel, we should remove these references as they are not relevant (for the great majority of users) anymore.

This PR was suggested in part by Nat from the comms after a review of the issues/PR workflow.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,9 +1,7 @@
 <!--
 Please target the `master` branch in priority.
-PRs can target `3.x` if the same change was done in `master`, or is not relevant there.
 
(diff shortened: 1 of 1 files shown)


Calinou (approved):
I agree with removing the 3.x notice given the development focus is increasingly on 4.x nowadays. We can manually ask for a 3.x version of the PR if it's really needed, but this will be reserved for important bug fixes now.

fire (approved):

akien-mga (approved):
I thought to do this too.

But if we're adding a first-interaction welcome message with an action, then we can probably remove the whole template. This HTML comment in your OP when you open a PR is always a bother IMO.

akien-mga:
Thanks!

PR #84635, opened 2023-11-08
Add PropertyListHelper

KoBeWi:
Follow-up to #83888 (partially)

This PR adds PropertyListHelper. It's a helper class for managing dynamic properties added with `_get_property_list()`. Initialize it with prefix, then register properties with name, default, setter and getter. Afterwards, instead of implementing `_get_property_list()`, `_set()` and `_get()`, just call a method in the helper and it handles everything for you. As a bonus it can also handle reverts.

I added it to ItemList. Property handling is now a bunch of one-liners, while the functionality has improved. Defaults no longer save in the scene and you can revert.

https://github.com/godotengine/godot/assets/2223172/39c4900e-1681-4be4-a292-34e136867eaf

I will do a follow-up to apply it to all relevant classes (TileMap, PopupMenu and probably some others).

UPDATE
In the second commit I changed how properties are registered and the 2 implementations are different. idk which one is better.

_Production edit: closes godotengine/godot-roadmap#14_

PR #88162, opened 2024-02-09
Add PropertyListHelper to PopupMenu

KoBeWi:
Follow-up to #84635

This PR adds a PropertyListHelper to PopupMenu.

However as I was implementing it, I found one property that required a new method. To avoid unnecessary binding, I decided to rethink the system a bit and now instead of Callables it uses MethodBind. It allows to add intermediate setters/getters when necessary without binding them and it's also safer, since it's now possible to use actual method pointer.

Aside from that, I moved one-liner methods which only call the helper to the header (up to discussion I guess), because they are rarely going to be modified and it's cleaner this way IMO (up to discussion I guess).

### DIFF
--- a/scene/gui/popup_menu.h
+++ b/scene/gui/popup_menu.h
@@ -35,6 +35,7 @@
 #include "scene/gui/margin_container.h"
 #include "scene/gui/popup.h"
 #include "scene/gui/scroll_container.h"
+#include "scene/property_list_helper.h"
 #include "scene/resources/text_line.h"
 
 class PopupMenu : public Popup {
@@ -59,7 +60,7 @@ class PopupMenu : public Popup {
 			CHECKABLE_TYPE_NONE,
 			CHECKABLE_TYPE_CHECK_BOX,
 			CHECKABLE_TYPE_RADIO_BUTTON,
-		} checkable_type;
+		} checkable_type = CHECKABLE_TYPE_NONE;
 		int max_states = 0;
 		int state = 0;
 		bool separator = false;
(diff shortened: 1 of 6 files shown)


akien-mga (approved):
Looks great!

akien-mga:
Thanks!

scene/gui/popup_menu.h
```diff
@@ -203,8 +211,10 @@ class PopupMenu : public Popup {
 
 	void _notification(int p_what);
 	bool _set(const StringName &p_name, const Variant &p_value);
-	bool _get(const StringName &p_name, Variant &r_ret) const;
-	void _get_property_list(List<PropertyInfo> *p_list) const;
+	bool _get(const StringName &p_name, Variant &r_ret) const { return property_helper.property_get_value(p_name, r_ret); }
+	void _get_property_list(List<PropertyInfo> *p_list) const { property_helper.get_property_list(p_list, items.size()); }
+	bool _property_can_revert(const StringName &p_name) const { return property_helper.property_can_revert(p_name); }
+	bool _property_get_revert(const StringName &p_name, Variant &r_property) const { return property_helper.property_get_revert(p_name, r_property); }
```
KoBeWi, line 217: I just realized this will be the same for pretty much every class, so it can be replaced with a macro 🤔
Though that's probably too extreme.

akien-mga: Yeah I thought about this too while reviewing the code, but I also thought it might be a bit overkill.

I guess you'll see as you keep implementing it, maybe with 5-6 repetitions is starts making sense to make it a macro. The main issue is that it hides the fact that the class overrides some common Object virtual methods.

PR #88164, opened 2024-02-10
Fix NavigationRegion errors from signal changes

smix8:
Fixes NavigationRegion errors from signal changes.

Also fixes that the debug would sometimes not update.


akien-mga (approved):

Scony (approved):

akien-mga:
Thanks!

PR #88167, opened 2024-02-10
Fix RichTextLabel bottom margin for text clipping

### LINKED ISSUE #87537: `RichTextLabel` not respecting the `content_margin_bottom` value.
### Tested versions

### System information

Windows 11 and Ubuntu 22.04, all renderers

### Issue description

When adding a style to a RichTextLabel, and setting content_margin_bot, the text does not respect these margin, which causes text being rendered on top of the margins.

Example:
image

### Steps to reproduce

- Add a RichTextLabel, fill it with text
- Add a StyleBoxFlat to it
- Set border width to 20 for all borders
- Set content margin to 30 for all borders
- Note how text overlaps with borders only in the bottom

[ThemeIssues.zip](https://github.com/godotengine/godot/files/14041284/ThemeIssues.zip)

DictyosteliumDiscoideum:
When a style with a bottom margin was added to a rich text label node, 
the bottom margin would be added to the text but the text would not be 
clipped. 
This adds a line to subtract the bottom margin from the label height 
before checking another line should be drawn.
Fixes #87537. Does not fix the additional issue raised regarding 
the right margins.

### DIFF
--- a/scene/gui/rich_text_label.cpp
+++ b/scene/gui/rich_text_label.cpp
@@ -817,6 +817,8 @@ int RichTextLabel::_draw_line(ItemFrame *p_frame, int p_line, const Vector2 &p_o
 	}
 
 	int line_count = 0;
+	// Bottom margin for text clipping.
+	float v_limit = theme_cache.normal_style->get_margin(SIDE_BOTTOM);
 	Size2 ctrl_size = get_size();
 	// Draw text.
 	for (int line = 0; line < l.text_buf->get_line_count(); line++) {
@@ -824,7 +826,7 @@ int RichTextLabel::_draw_line(ItemFrame *p_frame, int p_line, const Vector2 &p_o
 			off.y += theme_cache.line_separation;
 		}
 
-		if (p_ofs.y + off.y >= ctrl_size.height) {
+		if (p_ofs.y + off.y >= ctrl_size.height - v_limit) {
 			break;
 		}
 
@@ -1890,10 +1892,12 @@ void RichTextLabel::_notification(int p_what) {
 			visible_paragraph_count = 0;
 			visible_line_count = 0;
 
+			// Bottom margin for text clipping.
+			float v_limit = theme_cache.normal_style->get_margin(SIDE_BOTTOM);
 			// New cache draw.
 			Point2 ofs = text_rect.get_position() + Vector2(0, main->lines[from_line].offset.y - vofs);
 			int processed_glyphs = 0;
-			while (ofs.y < size.height && from_line < to_line) {
+			while (ofs.y < size.height - v_limit && from_line < to_line) {
 				MutexLock lock(main->lines[from_line].text_buf->get_mutex());
 
 				visible_paragraph_count++;


scene/gui/rich_text_label.cpp
```diff
@@ -1943,10 +1943,12 @@ void RichTextLabel::_notification(int p_what) {
 			visible_paragraph_count = 0;
 			visible_line_count = 0;
 
+			//bottom margin for text clipping
+			float vLimit=theme_cache.normal_style->get_margin(SIDE_BOTTOM);
```
AThousandShips, line 1947:
```suggestion
			float v_limit = theme_cache.normal_style->get_margin(SIDE_BOTTOM);
```
`snake_case`

scene/gui/rich_text_label.cpp
```diff
@@ -1943,10 +1943,12 @@ void RichTextLabel::_notification(int p_what) {
 			visible_paragraph_count = 0;
 			visible_line_count = 0;
 
+			//bottom margin for text clipping
+			float v_limit = theme_cache.normal_style->get_margin(SIDE_BOTTOM);
 			// New cache draw.
 			Point2 ofs = text_rect.get_position() + Vector2(0, main->lines[from_line].offset.y - vofs);
 			int processed_glyphs = 0;
-			while (ofs.y < size.height && from_line < to_line) {
+			while (ofs.y < size.height - vLimit && from_line < to_line) {
```
Calinou, line 1951: Fix compilation error:

```suggestion
			while (ofs.y < size.height - v_limit && from_line < to_line) {
```

scene/gui/rich_text_label.cpp
```diff
@@ -1943,10 +1943,12 @@ void RichTextLabel::_notification(int p_what) {
 			visible_paragraph_count = 0;
 			visible_line_count = 0;
 
+			//bottom margin for text clipping
+			float v_limit = theme_cache.normal_style->get_margin(SIDE_BOTTOM);
```
Calinou, line 1947: Follow comment style guide:

```suggestion
			// Bottom margin for text clipping.
			float v_limit = theme_cache.normal_style->get_margin(SIDE_BOTTOM);
```

Calinou (approved):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot/commit/0e9caa2d9cb20737f8dcf08b75fcf2a78d980569), it works as expected after fixing the compilation error mentioned in my above review comments.

### Before

https://github.com/user-attachments/assets/c81b74c3-c26c-49df-b70f-6e43d32fa041

### After

https://github.com/user-attachments/assets/ed287704-0c74-4c0f-b155-a2be2d209098

bruvzg (approved):
Seems fine. There's also the same check for the autowrapped lines, and likely need the same change:
https://github.com/godotengine/godot/blob/da5f39889f155658cef7f7ec3cc1abb94e17d815/scene/gui/rich_text_label.cpp#L820-L827

akien-mga:

PR #88173, opened 2024-02-10
FileSystem dock: Fix opening a Linux terminal

capnm:
The previous implementation of opening a terminal
in the FileSystem dock was causing errors due to the use
of a bash builtin command that does not work outside the shell.

This resulted in the following error messages:
```go
ERROR: Could not create child process: command
   at: execute (drivers/unix/os_unix.cpp:553)
ERROR: Could not create child process: command
   at: execute (drivers/unix/os_unix.cpp:553)
...
```
To resolve this issue,
+ the command is now executed inside a bash shell,
+ if the `terminal_emulator_flags` editor option is empty,
   the working directory for gnome-terminal and urxvt is now properly set.

**Tested only on GNOME platforms debian/ubuntu:**

`godot4.3devX -v -e`
...
Opening terminal emulator: gnome-terminal --working-directory ... /00-test-new_projects/new-project1/new_folder/test folder/

Opening terminal emulator: urxvt -cd ... /00-test-new_projects/new-project1/new_folder/test folder

Screenshot from 2024-02-10 22-05-50

### DIFF
--- a/editor/filesystem_dock.cpp
+++ b/editor/filesystem_dock.cpp
@@ -2128,7 +2128,8 @@ void FileSystemDock::_file_option(int p_option, const Vector<String> &p_selected
 				terminal_emulators.push_back(terminal_emulator_setting);
 			}
 
-			String arguments = EDITOR_GET("filesystem/external_programs/terminal_emulator_flags");
+			String flags = EDITOR_GET("filesystem/external_programs/terminal_emulator_flags");
+			String arguments = flags;
 			if (arguments.is_empty()) {
 				// NOTE: This default value is ignored further below if the terminal executable is `powershell` or `cmd`,
 				// due to these terminals requiring nonstandard syntax to start in a specified folder.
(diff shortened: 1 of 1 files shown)


editor/filesystem_dock.cpp
```diff
@@ -2134,17 +2134,15 @@ void FileSystemDock::_file_option(int p_option, const Vector<String> &p_selected
...
 				test_args.push_back(terminal_emulator);
-				// Silence command name being printed when found. (stderr is already silenced by `OS::execute()` by default.)
-				// FIXME: This doesn't appear to silence stdout.
-				test_args.push_back(">");
-				test_args.push_back("/dev/null");
-				int exit_code = 0;
-				const Error err = OS::get_singleton()->execute("command", test_args, nullptr, &exit_code);
-				if (err == OK && exit_code == EXIT_SUCCESS) {
+				const Error err = OS::get_singleton()->execute("bash", test_args, &pipe);
+				if (err == OK && pipe.find("/")) {
```
AThousandShips, line 2145:
```suggestion
				if (err == OK && pipe.contains("/")) {
```
Unless you mean to return true if `pipe` contains either no `/`, or it isn't the first character, if that's desired it should be made clear that this is what's desired

capnm: For  `bash | command -v gnome-terminal` you get the full path:
/usr/bin/gnome-terminal

and for `bash | command -v gnome-terminalBADCMD` I guess only new line or EOL...

AThousandShips: Are you trying to check for if there is a `/` in the string?

capnm: The | does not work for some reason. 
Maybe try calling the terminals with a --help option and see if the call succeeds...

capnm: The list elements are surrounded by double quotes in `os_unix.cpp`. 
So now it works as intended.

capnm:
Rebased to tip and tested KDE Plasma on Arch Linux:

Without this PR also broken:
Screenshot_20240214_131052

current tip + PR OK:

Screenshot_20240214_141207

akien-mga (approved):
I can't reproduce the bug on Fedora 39 with KDE Plasma (Wayland), but I tested the PR and it doesn't seem to break it either. So if it solves it for you, I guess it's fine.

Calinou:
I thought `command -v` was standard POSIX shell stuff that didn't require Bash (that's why I chose it in fact).

Calinou (approved):
Tested locally on Fedora 39 KDE X11 (with Konsole and Kitty), it works as expected.

akien-mga:
Thanks!

PR #88178, opened 2024-02-10
Link hash table primes externally to prevent data duplication in binary

Listwon:
I've started analyzing Godot binaries on Windows using [SizeBench](https://devblogs.microsoft.com/performance-diagnostics/sizebench-a-new-tool-for-analyzing-windows-binary-size/) and on top of the duplicate data results were `hash_table_size_primes` and `hash_table_size_primes_inv`.  
SizeBench

As visible above the tool shows around 168KB of wasted memory (tested on commit 36e943b6b20cb7a8a89bc30489c4a81c3e149d74 `scons target=template_release debug_symbols=yes`)

This PR reduces build size by 547KB ( `scons target=template_release`).

File scoped const arrays are linked internally which means each time `hashfuncs.h` are included, those two arrays are not merged into single data point (which would be fine and expected for const, read only data). ~I've moved their initialization to `hashfuncs.cpp` and marked `extern` in `hashfuncs.h`.~ Thanks to @Malcolmnixon suggestion I've made it simpler by using `inline constexpr` from C++17 standard.

I've not benchmarked it yet. In theory it can improve the CPU cache, because all of the hash_maps and hash_sets reuse the data from the same address in memory (meaning - there is higher chance that it will be still present in cache when needed).

### DIFF
--- a/core/templates/hashfuncs.h
+++ b/core/templates/hashfuncs.h
@@ -429,7 +429,7 @@ struct HashMapComparatorDefault<Vector3> {
 
 constexpr uint32_t HASH_TABLE_SIZE_MAX = 29;
 
-const uint32_t hash_table_size_primes[HASH_TABLE_SIZE_MAX] = {
+inline constexpr uint32_t hash_table_size_primes[HASH_TABLE_SIZE_MAX] = {
 	5,
 	13,
 	23,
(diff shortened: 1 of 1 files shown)


fire (approved):
I don't see how this functionally changes the code, so it seems like an decent optimization.

Malcolmnixon:
That's a good catch.

Since C++17 is now supported and the guidelines promote using constexpr, you might want to see if keeping the table in the header and marking them `constexpr` (and possibly `inline`) works as well. In theory this should also merge multiple instances of the data across different translation units, but also expose the values as constants to the compiler allowing constant-folding if accessed discretely.

Listwon:
I think I tried constexpr already and it didn't change anything in this case, but  I'll try `inline constexpr`. As you said, in theory it should work the same as the solution with `extern`.

Malcolmnixon (approved):
This looks good. Nice change!

akien-mga (approved):
Looks great! Thanks for the initiative looking into optimizing binary size and memory usage, that's very welcome :)

AThousandShips (approved):

akien-mga:
Thanks!

